PR #6308, opened 2018-07-01
replace mapping with serializable in compiler

kostya:


kostya:
travis crashed with getters:
```
in src/compiler/crystal/tools/expand.cr:122: undefined method 'expansions=' for Crystal::ExpandResult
        res.expansions = @found_nodes.map { |node| ExpandResult::Expansion.build(node) }

```

straight-shoota:
Just use getters where no setters are needed. You don't have to wait for travis if you run the specs locally.

kostya:
i don't know how it used and where, with mapping it all was as property.

straight-shoota:
Well, it's just six ivars and you already know one needs a setter. Just trial and error will get you through very quickly.

asterite:
Next we should remove mapping.

PR #6310, opened 2018-07-02
add Set#from_yaml

kostya:
Set#to_yaml already exists

### DIFF
--- a/src/yaml/from_yaml.cr
+++ b/src/yaml/from_yaml.cr
@@ -89,6 +89,22 @@ def Array.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
   end
 end
 
+def Set.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
+  ctx.read_alias(node, self) do |obj|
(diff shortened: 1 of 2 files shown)


RX14:
Please copy the `Array.new` code exactly.

src/yaml/from_yaml.cr
```diff
@@ -89,6 +89,31 @@ def Array.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
   end
 end
 
+def Set.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
+  ctx.read_alias(node, self) do |obj|
+    return obj
+  end
+
+  ary = new
```
wontruefree: Was this var copied from Array?
should it be something like `new_set` or something?

j8r:
What's the state of this PR? Sounds like mergeable.

RX14:
Sorry, missed the updated PR.

kostya:
i not sure it would work with aliases, but it can fixed separately

PR #844, opened 2015-06-23
Make .crystal directory location configurable

jhass:
* Default behaviour unchanged
* Allows to not clutter stuff up with .crystal directories wherever you happen to run `crystal eval`
* No conflicts, generated files are already namespaced by their full path
* Potential reusage of temporary files after changing the current working directory and compiling the same file again
* *Slight* speedup by setting the location to somewhere in /tmp
* If we later decide to put the default location into /tmp/crystal, ~/.crystal or whatever, if available, there's now a central place to implement that logic.

PR #6314, opened 2018-07-02
Human readable formatting for numbers

straight-shoota:
This PR adds three methods for formatting numbers in human-readable formats:

* `Number#format`: Similar to `#to_s` but allows customized separator and delimiter (useful for localization) and specifying the number of printed decimal places.
* `Number#humanize`: Formats the most significant digits of the number with a prefix indicating the magnitude. Output parameters, prefixes etc. can be fully customized.
* `Int#humanize_bytes`: Specialization of `Number#humanize` for binary units

```cr
123.4567.format(decimal_places: 3) # => "123.457"
1_200_000_000.humanize # => "1.2G"
1536.humanize_bytes # => "1.5KB"
```

`Number#humanize` is already put to use in the benchmark module.


src/number.cr
```diff
@@ -168,6 +168,241 @@ struct Number
...
+      start += group
+    end
+
+    decimal_places ||= dec_size
+
+    if decimal_places > 0
+      io << separator << decimals
+      unless only_significant
+        (decimal_places - dec_size).times do
+          io << '0'
```
r00ster91: Just `io << 0` would probably be better.

straight-shoota: Why do you think it would be better? 😁

r00ster91: I was pretty sure just a simple Int would be faster than a Char but nvm:
```Crystal
require "benchmark"

io = IO::Memory.new

Benchmark.ips do |bm|
  bm.report "char" {
    io << '0'
  }

  bm.report "int" {
    io << 0
  }
end
```
```
char  81.07M ( 12.34ns) (± 6.90%)  3 B/op        fastest
 int  76.95M (  13.0ns) (±11.22%)  4 B/op   1.05× slower
```

src/int.cr
```diff
@@ -486,6 +486,51 @@ struct Int
...
+  # ```
+  # 1.humanize_bytes                          # => "1B"
+  # 1024.humanize_bytes                       # => "1.0KB"
+  # 1536.humanize_bytes                       # => "1.5KB"
+  # 524288.humanize_bytes                     # => "512KB"
+  # 1073741824.humanize_bytes(standard: :IEC) # => "1.0GiB"
+  # ```
+  #
+  # See `Number#humanize` for more details on the behaviour and arguments.
+  def humanize_bytes(io : IO, precision : Int = 3, separator = '.', *, significant : Bool = true, standard : Symbol = :JDEC) : Nil
```
Sija: `:JDEC` -> `:JEDEC`

src/int.cr
```diff
@@ -486,6 +486,51 @@ struct Int
...
+    humanize(io, precision, separator, nil, base: 2 ** 10, significant: significant) do |magnitude|
```
Sija: `significant` -> `only_significant`?

src/int.cr
straight-shoota: Thanks 😉

src/int.cr
```diff
@@ -486,6 +486,51 @@ struct Int
...
+  # ```
+  # 1.humanize_bytes                          # => "1B"
+  # 1024.humanize_bytes                       # => "1.0KB"
+  # 1536.humanize_bytes                       # => "1.5KB"
+  # 524288.humanize_bytes                     # => "512KB"
+  # 1073741824.humanize_bytes(standard: :IEC) # => "1.0GiB"
+  # ```
+  #
+  # See `Number#humanize` for more details on the behaviour and arguments.
+  def humanize_bytes(io : IO, precision : Int = 3, separator = '.', *, significant : Bool = true, standard : Symbol = :JEDEC) : Nil
```
RX14: IEC should be the default :)

src/int.cr
```diff
@@ -486,6 +486,51 @@ struct Int
...
+    humanize(io, precision, separator, nil, base: 2 ** 10, significant: significant) do |magnitude|
```
RX14: just use `base: 1024`

src/int.cr
asterite: If only we had typsafe enums in the language...

straight-shoota: I considered that, but honestly I don't think it's worth adding an enum or this. It just distinguishes `:IEC` and anything else. A simple bool flag would do. I'm just using a symbol to be open for potential extensions (and can easily be changed to an enum).

src/int.cr
straight-shoota: IEC is still a long way from being widely adopted. People still mostly use JEDEC all over the place, so this should be the default (IMHO).

src/number.cr
```diff
@@ -168,6 +168,241 @@ struct Number
...
+
+  # ditto
+  def format(separator = '.', delimiter = ',', decimal_places : Int? = nil, *, group : Int = 3, only_significant : Bool = false) : String
+    String.build do |io|
+      format(io, separator, delimiter, decimal_places, group: group, only_significant: only_significant)
+    end
+  end
+
+  # Default SI prefixes ordered by magnitude.
+  SI_PREFIXES = {nil, 'k', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y', nil, 'y', 'z', 'a', 'f', 'p', 'n', 'µ', 'm'}
```
RX14: Using negative indexing for this is too magic, and removes the "ordered by magnitude" property.

src/number.cr
```diff
@@ -168,6 +168,241 @@ struct Number
...
+
+  # SI prefixes used by `#humanize`. Equal to `SI_PREFIXES` but prepends the
+  # prefix with a space charater.
+  SI_PREFIXES_PADDED = ->(magnitude : Int32, number : Float64) do
```
RX14: `number` is unused

src/number.cr
```diff
@@ -168,6 +168,241 @@ struct Number
...
+    index = (magnitude / 3).clamp(-max..max)
+    SI_PREFIXES[index]
+  end
+
+  # :nodoc:
+  def self.prefix_index(i, group = 3)
+    ((i - (i > 0 ? 1 : 0)) / group) * group
+  end
+
+  # Pretty prints this number as a `String` in a human-readable format.
```
RX14: significant isnt documented

src/int.cr
RX14: Every *human* understands IEC, and it makes the fact that you're using base-1024 absolutely unambiguous.

RX14: And not using IEC because it's not widely adopted is a massive self-fulfilling prophecy

straight-shoota: I agree. But still, I'm not convinced making IEC the default is the right path. 
OTOH, I really don't care much and it's easy to change.

asterite:
Can this be in a `number/humanize` require? The std keeps growing and base compile times are getting slower and slower.

RX14:
@asterite methods dont affect compile times *at all* unless they are used. This method is unused unless you use the benchmark harness then you're doing a release build anyway and you don't care.

asterite:
@RX14 Almost: the parser still has to parse it and build an AST for it. Every bit counts.

straight-shoota:
@asterite I don't think that extra tiny bit is worth the nuisance to have to require `number/humanize`. I'd like to encourage people to use these formatter methods for human-readable output and hope they'll be used in many Crystal applications 👍

straight-shoota:
@r00ster91 this is not the place to discuss `prelude.cr`.

src/number.cr
straight-shoota: done

asterite:
@straight-shoota Could you use descriptive commit messages for fixups? Just having `fixup!` with no context makes it really hard to review commits.

straight-shoota:
@asterite The first line of the commit message needs to reference the targeted commit, otherwise `!fixup` doesn't work. Descriptions are only visible in the second line.
Maybe I'm missing something but I can't find a way to prevent that.

asterite:
Is `fixup!` a convention or something?

yxhuvud:
@asterite if you use `git rebase --autosquash`, then it is magic that automatically make git rebase do the right thing.

straight-shoota:
@yxhuvud That still doesn't allow custom descriptions in the first line of the commit message, does it?

yxhuvud:
@straight-shoota Ah, no. That is a pity.

RX14:
@asterite it's not just a convention, it's an official part of git. `git commit --fixup=sha1` and then `git commit --autosquash master`.

RX14:
@asterite judging by `--stats`, parsing uses far far less than 1% of the time in any program. Even when compiling the entire spec suite
```
Parse:                             00:00:00.000042989 (   0.19MB)
```

asterite:
That's because `parse` stats just show the time for parsing the main file. When a `require` is processed, that parsing time happens in `top-level`, I think.

At one point all parsing was done in `parsing`, but since that has changed the `parsing` step has very little use now as stats info.

Regardless, it's not just parsing time, it's holding the AST nodes in memory for the entire compilation.

But yeah, let's not care. Let's wait for the magical solution that will enable incremental compilation :-)

asterite:
Sounds good.

src/int.cr
```diff
@@ -486,6 +486,51 @@ struct Int
...
+  # Values with binary measurements such as computer storage (e.g. RAM size) are
+  # typically expressed using unit prefixes based on 1024 (instead of multiples
+  # of 1000 as per SI standard). This method by default uses an extended range
+  # of the JEDEC units (`K`, `M`, `G`, `T`, `P`, `E`, `Z`, `Y`) which equals to
+  # the prefixes of the SI system except for uppercase `K` and is based on
+  # powers of 1024. The IEC standard prefixes (`ki`, `Mi`, `Gi`, `Ti`, `Pi`,
+  # `Ei`, `Zi`, `Yi`) will be used if *standard* is `:IEC`.
+  #
+  # ```
+  # 1.humanize_bytes                          # => "1B"
```
wontruefree: Why do some outputs have precision and some don't?
Seems like this should be `1.0B` or anything that is `humanize` that does not have a remainder should be `1B`

straight-shoota: You can't have a fraction of a byte, so `1.0 bytes` makes no sense. This is a special case, because number of bytes are non-negative integers.
In contrast, most other units are measurements expressed as real numbers. For example, `1.0 meters` is completely legit and semantically different from `1 meter` (reduced precision).

src/int.cr
asterite: Yes, please.

straight-shoota:
What do others think about IEC vs JEDEC? (see thread https://github.com/crystal-lang/crystal/pull/6314#discussion_r199770941)

straight-shoota:
The spec failure on MacOS seems really odd (see https://circleci.com/gh/crystal-lang/crystal/11612?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). It looks unrelated, but keeps popping up.

Can someone with a Mac run these specs locally and try to investigate this issue?

wontruefree:
@straight-shoota just pulled your branch on an OSX system and reproduced the same errors by running `TRAVIS_OS_NAME=osx ./bin/ci build`.  I merged your branch with upstream master and all the tests pass.

Here is the output.
https://gist.github.com/wontruefree/2e39da11d151bdb8718d081d38a20f9a

straight-shoota:
Thanks @wontruefree for investigating. It seems it only needed a rebase. I was under the impression I had rebased it yesterday on master, but that was not the case.

Now it's all green.

I'll squash the commits together to make reviewing easier.

src/http/server/handlers/log_handler.cr
```diff
@@ -22,12 +22,9 @@ class HTTP::LogHandler
...
-    return "#{seconds.round(2)}s" if seconds >= 1
-
-    millis = elapsed.total_milliseconds
-    return "#{millis.round(2)}ms" if millis >= 1
-
-    "#{(millis * 1000).round(2)}µs"
+    elapsed.total_seconds.humanize(precision: 2, significant: false) do |magnitude, _|
+      magnitude = Number.prefix_index(magnitude).clamp(-9..0)
+      {magnitude, magnitude == 0 ? "s " : "#{Number.si_prefix(magnitude)}s"}
+    end
```
RX14: Why does this require so much code?

RX14:
Also this still doesn't use enums

src/http/server/handlers/log_handler.cr
straight-shoota: Ooops, that's a mistake. I just copied the part from the benchmark formatter with the alignment specifics.

straight-shoota:
What is this supposed to mean?  https://circleci.com/gh/crystal-lang/crystal/13105?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
```
in spec/std/int_spec.cr:663: ambiguous call, implicit cast of :IEC matches all of Int::BinaryPrefixFormat



    it { 1024.humanize_bytes(format: :IEC).should eq "1.0kiB" }
```

I don't understand what the message means. It only happens on the master compiler, not 0.26.1.

asterite:
@straight-shoota There's a bug right now that makes automatic casting not work well with named arguments. I personally would remove the automatic casting feature, it's a failure and it's inconsistent with many things in the language. I rushed the implementation, as always.

Right now I would try to simply the language as much as possible, even if it means breaking changes. It's impossible to keep up with so many interacting features, specially when there's no one around that understands that code.

asterite:
@straight-shoota Or you can avoid using named arguments.

asterite:
@straight-shoota If I remember correctly, the problem is in these lines:

https://github.com/crystal-lang/crystal/blob/47b3feace7a64dce75f71b7e4289993020a2a404/src/compiler/crystal/semantic/call_error.cr#L149-L152

The logic is incorrect, it only works if there are no named arguments and all the defs that match the signature, ignoring type restrictions, is not well computed. Someone should fix that logic and it will start working well... I think. But I don't have time to tackle this.

z64:
Was doing a lot of number formatting the other day and thought of this PR. This looks like all points have been addressed, although I'm unsure about the most recent spec failure on `darwin`.

straight-shoota:
@z64 The CI failure looks like a random error caused by some timeout taking an unexpected amount of elapsed time.

I've triggered a rerun to ensure everything's fine.

src/int.cr
```diff
@@ -490,6 +490,63 @@ struct Int
...
+  # Prints this integer as a binary value in a human-readable format using
+  # a `BinaryPrefixFormat`.
+  #
+  # Values with binary measurements such as computer storage (e.g. RAM size) are
+  # typically expressed using unit prefixes based on 1024 (instead of multiples
+  # of 1000 as per SI standard). This method by default uses an extended range
+  # of the `BinaryPrefixFormat::JEDEC` units (`K`, `M`, `G`, `T`, `P`, `E`, `Z`,
+  # `Y`) which equals to the prefixes of the SI system except for uppercase `K`
+  # and is based on powers of 1024.
+  # The IEC standard prefixes (`ki`, `Mi`, `Gi`, `Ti`, `Pi`, `Ei`, `Zi`, `Yi`)
```
Sija:
```suggestion
  # The IEC standard prefixes (`Ki`, `Mi`, `Gi`, `Ti`, `Pi`, `Ei`, `Zi`, `Yi`)
```

straight-shoota:
Oh, it seems the CI failure is actually related. The new implementation is just significantly faster than what was expected previously.

That's really a bad spec based on execution speed of a piece of code.

sudo-nice:
Wouldn't it be good to have also a method like 
`1234567890.scatter # => 1_234_567_890`?

straight-shoota:
@sudo-nice That's essentially just `123456789.format(delimiter: '_')`. We could consider a dedicated method  for this format, but I think it should not be part of this PR.

sudo-nice:
@straight-shoota Oh, I overlooked that part, thanks!

src/int.cr
```diff
@@ -502,6 +502,63 @@ struct Int
...
+  # ```
+  # 1.humanize_bytes                        # => "1B"
+  # 1024.humanize_bytes                     # => "1.0KB"
+  # 1536.humanize_bytes                     # => "1.5KB"
+  # 524288.humanize_bytes                   # => "512KB"
+  # 1073741824.humanize_bytes(format: :IEC) # => "1.0GiB"
+  # ```
+  #
+  # See `Number#humanize` for more details on the behaviour and arguments.
+  def humanize_bytes(io : IO, precision : Int = 3, separator = '.', *, significant : Bool = true, format : BinaryPrefixFormat = :JEDEC) : Nil
```
RX14: `IEC` is still not the default.

j8r: Having `BinaryPrefixFormat::IEC`/`BinaryPrefixFormat::JEDEC` may be better for the doc

straight-shoota: Ok, fine...

RX14:
@straight-shoota you missed a few specs - CI's failing

Blacksmoke16:
Will this make it into 0.27.1?

bcardiff:
@straight-shoota what do you think about organizing all the humanize code in a separate file/folder?

RX14:
@bcardiff he's on holiday...

straight-shoota:
@bcardiff done

Sija:
Hmm, seems that lately CI bugs increased in frequency.

straight-shoota:
CI is still broken @bcardiff

bcardiff:
hmm, it's the same issue as in #7179. 

Lot's of `cc: error: ****.o: No such file or directory`

asterite:
Can those failing builds be reproduced? Are they using docker? If we can't reproduce these locally it'll be very hard to figure out what has gone wrong. And I'm almost sure it has nothing to do with a specific PR, I'm almost sure it's something about the CI (disk, memory?)

bcardiff:
The CRYSTAL_CACHE_DIR in the CI uses the `/tmp` inside docker. That was done a long [time ago]( https://github.com/crystal-lang/crystal/pull/844) maybe there is something special about `/tmp` (a cleanup could be triggered?) although its writes are performed over the container as any other folder.  

Let's see if [using the default location](https://github.com/bcardiff/crystal/commit/df0593b8e3f56331ab4ff3ed01ce9378de95adef) helps at [both](https://circleci.com/workflow-run/f5d73b49-dad5-4328-9c90-072c95204e83) [PRs](https://circleci.com/workflow-run/3a540f96-1c68-4df5-8bfb-d45f56b86c1b).

j8r:
Usually `/tmp` is mounted as tmps, which is RAM. If both the storage and the compilation use RAM, we can have issues [if the RAM is full](https://wiki.gentoo.org/wiki/Portage_TMPDIR_on_tmpfs#Troubleshooting). I haven't found anything more.

bcardiff:
It seems that the issue is not `/tmp` but somewhere when emitting the `.o.tmp` files the process stops. Leaving partial `.o.tmp` files and some missing `.o` oddly enough the missing files seem to be consistent.

Running in CircleCI `make std_spec FLAGS=--no-debug` seems to make things pass. Always.

I don't know if it is something about the debug information itself or the file size or something with the memory buffers.

That's all I could find.

asterite:
I suggest trying to run a debug build if the compiler with debug output (puts, after each .o is generated, check if the file is there later, etc.). Otherwise I don't know how can we debug this.

straight-shoota:
Rebased and ready to ship.

PR #6315, opened 2018-07-02
fix example codes (2018-06)

maiha:
Congrats 0.25.1! And fixed example codes.

#### 0.25.1 [b782738]
```
Specs 152 (117 successes, 14 failures, 21 pending)
Examples 1409 (1047 successes, 141 failures, 221 pending)
```

#### PR
```
Specs 152 (131 successes, 0 failures, 21 pending)
Examples 1409 (1153 successes, 0 failures, 256 pending)
```

Best regards,

### DIFF
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -252,7 +252,7 @@ module HTTP
   # timezone `GMT` (interpreted as `UTC`).
   #
   # ```
-  # HTTP.format_time(Time.new(2016, 2, 15)) # => "Mon, 15 Feb 2016 00:00:00 GMT"
+  # HTTP.format_time(Time.utc(2016, 2, 15)) # => "Mon, 15 Feb 2016 00:00:00 GMT"
   # ```
   #
   # Uses `Time::Format::HTTP_DATE` as formatter.
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -252,7 +252,7 @@ module HTTP
   # timezone `GMT` (interpreted as `UTC`).
   #
   # ```
-  # HTTP.format_time(Time.new(2016, 2, 15)) # => "Sun, 14 Feb 2016 21:00:00 GMT"
+  # HTTP.format_time(Time.new(2016, 2, 15)) # => "Mon, 15 Feb 2016 00:00:00 GMT"
   # ```
   #
   # Uses `Time::Format::HTTP_DATE` as formatter.
--- a/src/time.cr
+++ b/src/time.cr
@@ -873,7 +873,7 @@ struct Time
   # Format this time using the format specified by [RFC 3339](https://tools.ietf.org/html/rfc3339) ([ISO 8601](http://xml.coverpages.org/ISO-FDIS-8601.pdf) profile).
   #
   # ```
-  # Time.new(2016, 2, 15).to_rfc3339 # => "2016-02-15T00:00:00Z"
+  # Time.utc(2016, 2, 15).to_rfc3339 # => "2016-02-15T00:00:00Z"
   # ```
   #
   # ISO 8601 allows some freedom over the syntax and RFC 3339 exercises that
@@ -907,7 +907,7 @@ struct Time
   # Format this time using the format specified by [RFC 2822](https://www.ietf.org/rfc/rfc2822.txt).
   #
   # ```
-  # Time.new(2016, 2, 15).to_rfc2822 # => "Mon, 15 Feb 2016 00:00:00 +0000"
+  # Time.utc(2016, 2, 15).to_rfc2822 # => "Mon, 15 Feb 2016 00:00:00 +0000"
   # ```
   #
   # This is also compatible to [RFC 882](https://tools.ietf.org/html/rfc882) and [RFC 1123](https://tools.ietf.org/html/rfc1123#page-55).
--- a/src/time.cr
+++ b/src/time.cr
@@ -59,10 +59,10 @@ require "crystal/system/time"
 # time.second      # => 30
 # time.millisecond # => 0
 # time.nanosecond  # => 0
-# time.day_of_week # => Monday
+# time.day_of_week # => Time::DayOfWeek::Monday
 # time.day_of_year # => 46
 # time.monday?     # => true
-# time.time_of_day # => 15:10:30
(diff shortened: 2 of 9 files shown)


src/http/common.cr
```diff
@@ -246,7 +246,7 @@ module HTTP
   # timezone `GMT` (interpreted as `UTC`).
   #
   # ```
-  # HTTP.format_time(Time.new(2016, 2, 15)) # => "Sun, 14 Feb 2016 21:00:00 GMT"
+  # HTTP.format_time(Time.new(2016, 2, 15)) # => "Mon, 15 Feb 2016 00:00:00 GMT"
```
straight-shoota: The output depends on local time zone configuration. `Time.new` should better receive a specific time zone.

src/time.cr
```diff
@@ -37,12 +37,12 @@ require "crystal/system/time"
 #
 # ```
 # time = Time.utc(2016, 2, 15, 10, 20, 30)
-# time.to_s        # => 2016-02-15 10:20:30 UTC
+# time.to_s        # => "2016-02-15 10:20:30 UTC"
 # time = Time.new(2016, 2, 15, 10, 20, 30, location: Time::Location.load("Europe/Berlin"))
-# time.to_s        # => 2016-02-15 10:20:30 +01:00 Europe/Berlin
-# The time-of-day can be omitted and defaults to midnight (start of day):
+# time             # => 2016-02-15 10:20:30 +01:00 Europe/Berlin
```
straight-shoota: Please add `.to_s` and wrap result in quotes like in the other examples.

src/time.cr
```diff
@@ -907,7 +907,7 @@ struct Time
   # Format this time using the format specified by [RFC 2822](https://www.ietf.org/rfc/rfc2822.txt).
   #
   # ```
-  # Time.new(2016, 2, 15).to_rfc2822 # => "Mon, 15 Feb 2016 00:00:00 -0400"
+  # Time.new(2016, 2, 15).to_rfc2822 # => "Mon, 15 Feb 2016 00:00:00 +0000"
```
straight-shoota: Example should use a specific time zone.

src/time/format/custom/http_date.cr
```diff
@@ -12,7 +12,7 @@ require "./rfc_2822"
 # Time::Format::HTTP_DATE.parse("Sunday, 14-Feb-16 21:00:00 GMT") # => 2016-02-14 21:00:00 UTC
 # Time::Format::HTTP_DATE.parse("Sun Feb 14 21:00:00 2016")       # => 2016-02-14 21:00:00 UTC
 #
-# Time::Format::HTTP_DATE.format(Time.new(2016, 2, 15)) # => "Sun, 14 Feb 2016 21:00:00 GMT"
+# Time::Format::HTTP_DATE.format(Time.new(2016, 2, 15)) # => "Mon, 15 Feb 2016 00:00:00 GMT"
```
straight-shoota: Example should use a specific time zone.

src/time.cr
```diff
@@ -873,7 +873,7 @@ struct Time
   # Format this time using the format specified by [RFC 3339](https://tools.ietf.org/html/rfc3339) ([ISO 8601](http://xml.coverpages.org/ISO-FDIS-8601.pdf) profile).
   #
   # ```
-  # Time.new(2016, 2, 15).to_rfc3339 # => "2016-02-15T00:00:00+00:00"
+  # Time.new(2016, 2, 15).to_rfc3339 # => "2016-02-15T00:00:00Z"
```
straight-shoota: Example should use a specific time zone.

maiha:
@straight-shoota thanks. I added timezone by using `Time.utc` rather than `Time.new`.

asterite:
@maiha Nice! How long does it take to compile and run all examples? Do you generate a single big program, or is it one compilation per example?

maiha:
@asterite 

- **time**: 8 mins (Intel 3GHz)
- **spec**: one spec file per one class file, because
  - **putting all examples into 1 spec file**: failed due to name conflictions about `variables` and `classes`. And it takes a long time to run again.
  - **putting 1 example into 1 spec file**: failed because some examples uses `variables` and `classes` defined in previous examples.

For example, this is a generated spec file from `src/array.cr` . And I run it in docker for security reason.
https://github.com/maiha/crystal-examples/blob/master/gen/spec/array_spec.cr

maiha:
I forgot to run `tool format`.

asterite:
8 minutes is a lot of time, but I guess for smaller projects it will be a reasonable time.

I wonder if we should embed this feature directly in the compiler. That way we'll have `doctest` embedded in the language. And we'll be able to run this in CI to check that all examples compile/run fine. And they never get out of date.

I think the CI already takes a lot of time, so 8 minutes more isn't thaaat bad :-P

Thoughts?

j8r:
It's not needed to add on the CI, because this is only about comments at the end, not code. Having some incorrect docs doesn't prevent to have a working crystal compiler.
However running it before each Crystal release will enforce consistent docs :+1:

maiha:
**Execution time** can be reduced by caching. So I think it is not a big problem.

As @j8r is worried, it is a problem that there is a **not working code** case. Actually, there are many image codes like the following.
- https://github.com/crystal-lang/crystal/blob/master/src/array.cr#L143-L145
- https://github.com/crystal-lang/crystal/blob/master/src/process.cr#L428

It takes time manually to **judge** this. And Most of the time is spent here.😱 
I think this is a major barrier in the case of automation.

straight-shoota:
@j8r It's not strictly necessary, but really nice to have if the code examples are automatically validated. This doesn't even need to be integrated in the compiler. We could run the examples checker right now in CI.

@maiha Sure, some examples need manual intervention, but that's what this is about. Having automatic checks in CI would prevent simple errors and encourage people to write working examples.

Obviously, not every example can be written in a way that it is actually executable. But it should be quite easy to flag such examples as *ignore*. For example using a special syntax highlight declaration: ` ```crystal doctest=ignore `

sdogruyol:
@maiha this need a rebase and good to go :+1:

maiha:
rebased. I hope this will be merged before another conflict occurs.  😃

bcardiff:
Thanks @maiha

Issue #5291, opened 2017-11-14
Make crystal more beginner friendly


faustinoaq:
After [seeing this Treehouse teacher](https://youtu.be/71oXjLtXcv4?t=3988) trying to experiment with crystal I thought we can improve the way we show crystal to newcomers:

- [x] Make linking errors more clear to the user or making packages like `git`, `openssl` recommended packages on installation guide see Treehouse video at [1:16:48](https://youtu.be/71oXjLtXcv4?t=4618) See #5486
- [x] Make shards dependencies errors clear to the user (maybe a help message saying `shard install`), see Treehouse video at [1:21:46](https://youtu.be/71oXjLtXcv4?t=4906) See #5487
- [x] Make crystal available on Windows 😉 see: https://github.com/crystal-lang/crystal/issues/5430
- [x] With `crystal foo.cr --flag`, when `--flag` isn't found, mention "did you mean `crystal foo.cr -- --flag`" see [1:37:35](https://youtu.be/71oXjLtXcv4?t=5855)
- [ ] Add a simple tutorial similar to the "create a shard" tutorial for apps. Doesn't have to be too deep.
- [ ] Make the crystal package depend on git, a cc (which should depend on a ld), and some more of the libraries required for stdlib linking as recommended deps, and openssl-devel, etc as suggested deps. In addition to the linking error improvements above.

/cc @bcardiff @fridgerator

## Update (2017-11-15)

- Add ideas by @RX14 

## Update (2018-12-06)

- Change Windows issue to https://github.com/crystal-lang/crystal/issues/5430

RX14:
Presumably cc @kennethlove too :)

I'd add a few things to this list too:
- [ ] With `crystal foo.cr --flag`, when `--flag` isn't found, mention "did you mean `crystal foo.cr -- --flag`"
- [ ] Add a simple tutorial similar to the "create a shard" tutorial for apps. Doesn't have to be too deep.
- [ ] Make the crystal package depend on `git`, a `cc` (which should depend on a `ld`), and some more of the libraries required for stdlib linking as recommended deps, and openssl-devel, etc as suggested deps. In addition to the linking error improvements above.

Wulfklaue:
Several items that i see in the video are:

Missing system dependencies, for instance for Debian or Ubuntu you need:

```
Debian 9  ( Minimum server install with only SSH ): apt-get install apt-transport-https dirmngr libssl-dev gcc zlib1g-dev git
[...]

PR #6316, opened 2018-07-02
Did you mean flag

benglewis:
I can't seem to build the tests 100% on my Mac, so I'm opening this unable to 100% verify it :(

### DIFF
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -215,7 +215,8 @@ describe "OptionParser" do
   end
 
   it "raises on invalid option" do
-    expect_raises OptionParser::InvalidOption, "Invalid option: -j" do
+    suggestion = "(did you mean 'crystal foo.cr -- -j'?)".colorize.yellow.bold.to_s
+    expect_raises OptionParser::InvalidOption, "Invalid option: -j #{suggestion}" do
       OptionParser.parse(["-f", "-j"]) do |opts|
         opts.on("-f", "some flag") { }
       end
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -215,7 +215,7 @@ describe "OptionParser" do
   end
 
   it "raises on invalid option" do
-    suggestion = "(did you mean 'crystal foo.cr -- -j'?)".colorize.yellow.bold.to_s
+    suggestion = "(did you mean 'crystal foo.cr -- -j'?)".colorize.yellow.bold
     expect_raises OptionParser::InvalidOption, "Invalid option: -j #{suggestion}" do
       OptionParser.parse(["-f", "-j"]) do |opts|
         opts.on("-f", "some flag") { }
--- a/src/option_parser.cr
+++ b/src/option_parser.cr
@@ -29,17 +29,17 @@
 # destination = destination.upcase if upcase
 # puts "Hello #{destination}!"
 # ```
+
+require "colorize"
+
 class OptionParser
   class Exception < ::Exception
   end
 
   class InvalidOption < Exception
     def initialize(option)
-      super("Invalid option: #{option} #{
-        type.program.colorize(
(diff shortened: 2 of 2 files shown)


asterite:
Sorry, but this change makes no sense. What problem are you trying to solve?

benglewis:
@asterite https://github.com/crystal-lang/crystal/issues/5291
I would have liked to provide the actual file name in the error, but that information isn't available, and fixing that issue would involve a major amount of re-write that could break other people's use cases, so I've avoided it thus far. Let me know if you see any other issues here :)

asterite:
That issue is about showing suggestions for when you pass a flag to the compiler, and the flag is not found. Here you are changing option parser (which can be used in any crystal program, not just the compiler), to suggest that the flag should be passed Ina  different way. And that makes no sense at all.

straight-shoota:
@benglewis `OptionParser` is used by countless Crystal applications, not just by the compiler. You can't add a very compiler-specific behaviour to such a general class. This did-you-mean flag needs to go into the compiler's command handling.

benglewis:
Oh.

paulcsmith:
@benglewis I've run into this myself, especially when I was just starting Crystal. I would have loved to have seen a nice error like the one you are suggesting so I wouldn't have to ask for help.

@straight-shoota or @asterite could you provide a link to the file or some other guidance so that this suggestion could be shown in the correct place? That may help @benglewis or someone else make the changes to add this feature in the right place

straight-shoota:
This is the place to look: https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/command.cr#L170

asterite:
The original proposal is that if you run a program with `crystal foo.cr --flag` and that flag doesn't exist **in the program foo.cr**, then you'd get an error saying you should invoke it like `crystal foo.cr -- --flag`. To me, that seems impossible to implement. Or, well, it's this PR, but it's not correct to show that error because you can run the program as `. /foo` once it's compiled.

And I wouldn't go down the route of adding a flag that is present if you run the program through crystal, just for this error message.

There should simply be an answer to this in stackoverflow.com, problem solved. 

https://stackoverflow.com/questions/45117892/passing-cli-arguments-to-excutables-with-go-run

paulcsmith:
I didn't read the original issue so that was my mistake.

I don't think it is a high priority to, but I do think that Crystal will be more loved if more helpful errors can be shown to users. It's such a delight when languages/libraries give me a hint that saves even just a few minutes.

So maybe it would be hard, but if someone adds it, I think it would be a great addition! Especially because people may not know what to google for.

My thought would be that it only applies when using the `crystal` tool and if you give it an option it doesn't understand it shows what it shows now, plus an additional help saying something like: "If you meant to access this flag in your program use `-- #{options}`. Much like what the author did, but just for the `crystal` command and not in `OptionParser`. Maybe that's what you already imagined, but I wanted to get it written down just in case.

What do you think? If someone does implement it, would it be worth adding to save time and give users that extra bit of help so they don't break their flow?

asterite:
Also note that if you have a program that just does `ARGV[0]`, so it doesn't use OptionParser, then there's simply no way to detect this. And if you only add this knowledge to OptionParser, if you use another shard to parse options then it won't work (unless that shard also adds a similar solution).

So I don't think the original issue makes much sense. That's just my opinion, though, but again, this seems impossible to implement.

RX14:
Really, `crystal foo.cr --flag` should send `--flag` to `foo.cr`.

Options before the filename should go the compiler, options after should go to the file. As well as being less confusing, this also makes it possible to do `#!/usr/bin/env crystal` in scripts.

benglewis:
Thank you for the discussion guys.

I am in agreement with @RX14 here. I think that simply considering the order of the flags would be the most logical version to me as a user, but I do understand that that may not fit with the strategy for Crystal. If a conclusion can be reached on what to implement, I would be happy to try as a complete newbie to Crystal :)

PR #5428, opened 2017-12-21
Use shards version in crystal init tool

RX14:
Use the `shards version` command introduced by https://github.com/crystal-lang/shards/pull/148 to populate the `VERSION` constant inside generated projects with `crystal init`.

I moved the `VERSION` constant into the main "example.cr" file, instead of a `version.cr` file too. I guess that will be a divisive change, but I don't think a seperate `version.cr` gains us anything.

PR #6317, opened 2018-07-02
Flatten project structure created by crystal init

straight-shoota:
`crystal init` previously created `src/example/version.cr` file which only contained a `VERSION` constant. This is moved into the main library file. It also removes `require "./example/*"` from main file.

The old `version.cr` convention persisted from Ruby due to the need for gemspec files to require only the version constant from the library.

Extracted from #5428 to make this easier to merge with (hopfully) less discussion and we can focus there (or in a new issue) on the value of `VERSION`. @RX14 I hope you don't mind 😉 

https://github.com/crystal-lang/crystal/pull/5428#issuecomment-353474769
> In my view version.cr is much more cluttered than just adding a line in your main file. Ruby convention has it separate only because the gemspec depends on the version.cr - in Crystal we have it the other way around so we can merge it back in.

### DIFF
--- a/spec/compiler/crystal/tools/init_spec.cr
+++ b/spec/compiler/crystal/tools/init_spec.cr
@@ -144,18 +144,11 @@ dependencies:
       end
 
       describe_file "example/src/example.cr" do |example|
-        example.should eq(%{require "./example/*"
-
-# TODO: Write documentation for `Example`
+        example.should eq(%{# TODO: Write documentation for `Example`
 module Example
-  # TODO: Put your code here
-end
-})
-      end
-
-      describe_file "example/src/example/version.cr" do |version|
-        version.should eq(%{module Example
   VERSION = "0.1.0"
+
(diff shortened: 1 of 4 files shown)


wontruefree:
I cant 👍 @ysbaddaden comment but I agree.

kostya:
put everything in src produce much more require conflicts, as src is a search path in crystal.

straight-shoota:
@kostya What? `require "./shardname/foo"` results in exactly the same conflicts as `require "./foo"` if everything is just put into `src/shardname` folder.

ysbaddaden:
@kostya no, this conflict only happens in *Rubygem*, which adds each gem `lib` folder to the search path, which forces to use a `lib/foo` folder. This doesn't happen in Crystal, where only the project's `lib` folder is added, not each `src` folder of every shard. Having a `src/foo` folder or not is exactly the same.

PR #6319, opened 2018-07-02
random: Implement Random#uuid

woodruffw:
This method behaves identically to Ruby's `SecureRandom#uuid`,
generating a random v4 UUID (with only the version and variant being
constant/determined).

Please let me know if there's a way to do this that involves fewer allocations :slightly_smiling_face:

### DIFF
--- a/spec/std/random_spec.cr
+++ b/spec/std/random_spec.cr
@@ -260,4 +260,10 @@ describe "Random" do
       hex.should eq("9fd857f462831002ffffffffffffffff0000000000000000e88d3a30db4e730021b8a5e33b020000362f518e0700000062da")
     end
   end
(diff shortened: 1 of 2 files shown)


wontruefree:
Have you seen these api docs?
https://crystal-lang.org/api/0.25.1/UUID.html

ysbaddaden:
Yes, we already have `UUID.random` that doesn't allocate. Please use it :-)

woodruffw:
Not sure how I missed that. Sorry!

ysbaddaden:
Probably because it's reversed from Ruby (`SecureRandom.uuid` => `UUID.random`) since we have an UUID type.

PR #6322, opened 2018-07-03
JSON/YAML: error if a maximum nesting is reached (prevents stack overflow)

### LINKED ISSUE #6321: Invalid memory access when using JSON::Serializable
After a day playing around I just can't figure out what I'm doing wrong, it seems so much behind the scenes magic is involved in this so I'm completely lost.

I'm trying to add an option to save\load a neural network from JSON using the `JSON::Serializable`, but I keep getting an invalid memory access crash.

https://github.com/NeuraLegion/shainet/pull/87

If anyone can tell me what is going on or why it might crash like that I would be forever grateful.

asterite:
Fixes #6321

This PR adds a `max_nesting` property to both `JSON::Builder` and `YAML::Builder` with a default value of 99. When trying to generate a JSON/YAML document with a nesting (arrays/objects inside arrays/objects) greater than the maximum, an error is raised.

Other languages usually do the same, because it's super rare (I'd say unmanagable) to have a JSON/YAML document that deep. There was also a website listing how JSON is handled in each language, and checking for a maximum nesting was done and most languages did this: those that not were marked as red, because it means the language is not safe. And it's not safe because it will cause a stack overflow, which usually results in the end of a program.

The error message is copied from Ruby.

Of course there's no easy way to control this maximum nesting calling just `to_json`, `to_yaml`. However, in the rare case you need to change this maximum nesting you can always creating a builder, configure it, and pass it to `to_json`/`to_yaml`.

### DIFF
--- a/spec/std/json/builder_spec.cr
+++ b/spec/std/json/builder_spec.cr
@@ -278,4 +278,33 @@ describe JSON::Builder do
       end
     end
   end
+
+  it "errors on max nesting (array)" do
+    io = IO::Memory.new
+    builder = JSON::Builder.new(io)
+    builder.max_nesting = 3
+    builder.start_document
+    3.times do
+      builder.start_array
+    end
+
+    expect_raises(JSON::Error, "Nesting of 4 is too deep") do
+      builder.start_array
+    end
+  end
+
+  it "errors on max nesting (object)" do
+    io = IO::Memory.new
+    builder = JSON::Builder.new(io)
+    builder.max_nesting = 3
+    builder.start_document
+    3.times do
+      builder.start_object
+      builder.string "key"
+    end
+
+    expect_raises(JSON::Error, "Nesting of 4 is too deep") do
+      builder.start_object
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


asterite:
Oh, I was sure we had this behavior somewhere. It turns out we have it [when parsing](https://github.com/crystal-lang/crystal/blob/master/src/json/parser.cr#L128-L136). But it makes sense to have it also when emitting.

RX14:
I don't think we should prevent stack overflow: I think we should let it happen and fix the stack overflow error message.

As for parsing, time taken to parse the JSON isn't `O(nesting)` it's `O(bytesize)`, so the request body size is the correct way to stop a DoS via large JSON payload.

asterite:
I think, once you get stack overflow, there's not much you can do with the app. If we throw JSON::Error that's manageable, and it can be logged, the server can continue working, etc.

jhass:
The nesting attack on parsing isn't DoS through resource consumption but DoS through crashing the server repeatedly (and maybe indirect resource consumption of it restarting all the time).

RX14:
@jhass ah I see, well the fix to that is perhaps stack overflow crashing only one fiber.

PR #6323, opened 2018-07-03
Refactor and improve URI implementation

straight-shoota:
This PR applies a ton of mostly small improvements to `URI` and `URI::Parser`.

* The type of `URI#path` is changed to `String`. It is no longer nilable. The semantic is: every URI has a path, but it may be empty. This avoids ambiguity between empty path and nil path.
* The macro `URI::Parser.step` is removed. It was previously used to spec internals of `URI::Parser` but it no longer works because the struct can't be inherited. It doesn't make much sense to spec the internal API anyway, having unit tests for `URI.parse` is totally fine. Consequently, I have removed the specs for `URI::Parser` and integrated them into the specs for `URI.parse`.
* Most notable is probably the removal of the property `URI#opaque`. It is simply not necessary to have an extra ivar for this. It is replaced by `URI#path` which contains both a path in the hierarchical part as well as a non-hierarchical, opaque path. `URI#opaque?` answers if the URI is considered opaque.
* The spec suite has been reorganized and I added a bunch of new examples, borrows from gloang and akka.
* During that process I discovered and fixed a few issues in the parser and formatter:
  * `#to_s` no longer hides the port if it is the default port. A default port can now be stripped using `#normalize`.
  * `#parse` downcases the `scheme` to make it case-insensitive. This is recommended by RFC 3986.
  * `#parse` decodes escape sequences in `host`, and `#to_s` applies them.
  * `#parse` correctly parses an empty port. Previously, `scheme://host:/` would set `port` to `0`, now it's `nil`.
  * `#parse` parses a string starting with three slashes as empty host. Previously, `host` was `nil`.
  * `#to_s` inserts a slash if `path` does not start with one and follows an authority. Paths starting with double slash `//` are prefixed by `/.` if it does not follow an authority. Otherwise the first path component would be interpreted as authority.


spec/std/uri_spec.cr
```diff
@@ -1,51 +1,143 @@
...
-  assert_uri("https://%3AD:%40_%40@www.example.com", scheme: "https", host: "www.example.com", user: ":D", password: "@_@")
-  assert_uri("https://www.example.com/#top", scheme: "https", host: "www.example.com", path: "/", fragment: "top")
-  assert_uri("http://www.foo-bar.example.com", scheme: "http", host: "www.foo-bar.example.com")
-  assert_uri("/foo", path: "/foo")
-  assert_uri("/foo?q=1", path: "/foo", query: "q=1")
-  assert_uri("mailto:foo@example.org", scheme: "mailto", path: nil, opaque: "foo@example.org")
+  describe ".parse" do
+    # scheme
+    assert_uri("http:", scheme: "http")
+    it { URI.parse("HttP:").should eq(URI.new(scheme: "http")) }
```
yxhuvud: you are missing a spec of what the result of `#scheme` on a missing scheme is. Some tests below lacks scheme but none tests what `#scheme` is in that case.

yxhuvud:
The test failures make me wonder if `eq` should handle default ports.

straight-shoota:
Nah, the specs are only failing because of the previous incorrect behavior of ommitting the port if it was specified but equal to the default.

Comparing two `URI`s should be strict. If a port is specified, even if it's the default, there are chances this is on purpose. `http://example.com` and `http://example.com:80` are *not* identical. You can easily use `#normalize` to remove default ports.

spec/std/uri_spec.cr
straight-shoota: I'm not sure what you mean. There are many specs for no-scheme URIs. For example `assert_uri("/foo", path: "/foo")`. It ensures that the URI parsed from `"/foo"` has a path of `/foo` and all other properties at the default value. This means `scheme` should be `nil`.

asterite:
```
$ irb
irb(main):001:0> require "uri"
=> true
irb(main):002:0> u1 = URI.parse("http://foo.com:80")
=> #<URI::HTTP http://foo.com>
irb(main):003:0> u2 = URI.parse("http://foo.com")
=> #<URI::HTTP http://foo.com>
irb(main):004:0> u1 == u2
=> true
```

asterite:
I'm not saying we should copy Ruby... but specifying a default port when it's the default one, and keeping it around... I don't know... there's no difference, really.

straight-shoota:
@asterite Yeah, Ruby does that. But it also uses a scheme-specific type `URI::HTTP`. Maybe that can serve as an excuse...

Besides, just a few counter examples:

```go
u, err := url.Parse("http://foo.org:80")
fmt.Println(u.Port()) // => 80
```
```java
System.out.println(new URI("http://foo.com:80").getPort()); // => 80
```

straight-shoota:
> there's no difference, really.

For well-known schemes, probably not. But when you're using a custom scheme, it might not be given that the next consumer of a URI has the same default port configured.

`URI` should be a simple data container and serialize to and from string representation without any further processing. If you want to get rid of unnecessary default ports, use `#normalize`. 

`URI.parse(string).to_s == string` should hold for any well-formed URI.

asterite:
Sounds good

straight-shoota:
🏓 A second review would be great!

src/uri.cr
```diff
@@ -59,10 +59,10 @@ class URI
   # ```
   # URI.parse("http://foo.com/bar").path # => "/bar"
   # ```
-  getter path : String?
+  getter path : String
```
wontruefree: Why not use the `property` syntax?
It seems like the `getter` and `setter` are the same.

asterite: Because of documentation (setter and getter need different docs).

straight-shoota:
@bcardiff 🏓

straight-shoota:
I'd really like to get this merged for 0.28.0. It's so annoying to have to work around URI bugs I fixed six months ago.

spec/std/uri_spec.cr
```diff
@@ -111,35 +212,21 @@ describe "URI" do
...
-      URI.new("ldap", "www.example.com", 389).to_s.should eq("ldap://www.example.com")
-      URI.new("ldaps", "www.example.com", 636).to_s.should eq("ldaps://www.example.com")
-    end
+    it { URI.new("http", "www.example.com", 80, "/hello").to_s.should eq("http://www.example.com:80/hello") }
+    it { URI.new("http", "www.example.com", 80, "/hello", "a=1").to_s.should eq("http://www.example.com:80/hello?a=1") }
+    it { URI.new("mailto", path: "foo@example.com").to_s.should eq("mailto:foo@example.com") }
+    it { URI.new("file", path: "/foo.html").to_s.should eq("file:/foo.html") }
+    it { URI.new("file", path: "foo.html").to_s.should eq("file:foo.html") }
+    it { URI.new("file", host: "host", path: "foo.html").to_s.should eq("file://host/foo.html") }
+    it { URI.new(path: "//foo").to_s.should eq("/.//foo") }
```
Sija: That's seems odd, why the prefix?

spec/std/uri_spec.cr
```diff
@@ -111,35 +212,21 @@ describe "URI" do
...
+    it { URI.new("http", "www.example.com", user: "alice", password: "s3cr3t").to_s.should eq("http://alice:s3cr3t@www.example.com") }
+    it { URI.new("http", "www.example.com", user: ":D").to_s.should eq("http://%3AD@www.example.com") }
+    it { URI.new("http", "www.example.com", user: ":D", password: "@_@").to_s.should eq("http://%3AD:%40_%40@www.example.com") }
     it { URI.new("http", "www.example.com", user: "@al:ce", password: "s/cr3t").to_s.should eq("http://%40al%3Ace:s%2Fcr3t@www.example.com") }
     it { URI.new("http", "www.example.com", fragment: "top").to_s.should eq("http://www.example.com#top") }
-    it { URI.new("http", "www.example.com", 80, "/hello").to_s.should eq("http://www.example.com/hello") }
-    it { URI.new("http", "www.example.com", 80, "/hello", "a=1").to_s.should eq("http://www.example.com/hello?a=1") }
-    it { URI.new("mailto", opaque: "foo@example.com").to_s.should eq("mailto:foo@example.com") }
-
-    it "removes default port" do
```
Sija: Why not normalize ports from default?

spec/std/uri_spec.cr
straight-shoota: An URI starting with double slash is considered relative to the scheme. Thus `//foo` would resolve to a URI with authority `foo` and empty scheme and path.

Sija: Just now I've noticed it's the path being set and not a the full uri, thanks for explanation.

spec/std/uri_spec.cr
straight-shoota: There are no generic default ports in the URI specification. It depends solely on configuration which port is considered "default".

When no port is specified, the resolver is entrusted to choose whichever the port it seems fit. But when a port is explicitly specified, the resolver is asked to use that specific port, regardless of it's configuration for default ports.

Thus removing a port (even if it is considered default) mutates the URI. It looses information.
A well-formed URI should not be automatically stripped of anything unless explicitly asked (for example using `#normalize´).

straight-shoota:
@ysbaddaden I strongly disagree. The reason for the default port change is not *because RFC says so* but actually based on common sense.

When a URI is specified with a port, then it *has* a port and `URI` should not say otherwise. I expect that URI to stringify with that explicit port and don't see any reasonable explanation for purposely loosing that information.

Consider an example where a URI input is processed by `URI` type and relayed to another service. Let's assume that in `URI`, there is a default port registered for the URI's scheme, but the other service doesn't know a default port for this scheme. When `URI` decides *Yay, I know this port, let's smack it!*, the service receives a URI it can't resolve despite the port being **explicitly specified** initially.

There is obviously a use case to simplify URIs (mainly for user presentation) when the specified port is the default. This can be easily achieved using `#normalize`. But this should definitely not be applied by default.

> Another case I see i specs is printing `//www.example.com:1234` which is clearly not what was expected.

Could you clarify on that? I'd consider this very much the expected result for stringifying a protocol-relative URI with host `www.example.com` and port `1234`.

asterite:
I think it makes sense to keep the port if the URI has it set explicitly, even if it's the default port for that scheme.

src/uri.cr
```diff
@@ -189,15 +193,16 @@ class URI
...
-  def normalize
+  def normalize : self
     uri = dup
     uri.normalize!
     uri
   end
 
   # Destructive normalize.
-  def normalize!
+  def normalize! : Nil
```
Sija: Please, return `self` here.

straight-shoota:
Rebased and ready to ship.

wontruefree:
@straight-shoota should all the `parse_` methods in the `src/uri/uri_parser.cr` file be private?

wontruefree:
I just realized that might have been out of scope for this refactor but thanks for moving the parse methods to private.

src/uri.cr
```diff
@@ -135,49 +125,63 @@ class URI
   # uri = URI.parse "http://foo.com/posts?id=30&limit=5#time=1305298413"
   # uri.full_path # => "/posts?id=30&limit=5"
   # ```
-  def full_path
+  def full_path : String
```
j8r: Another method can also accept an IO.

straight-shoota: What?

straight-shoota:
@wontruefree Yeah I thought so at first. But then, what the heck... it's a simple thing and fits to the set of changes in this PR.

src/uri.cr
j8r: i mean an overload, `full_path(io : IO)`
This avoids creating an intermediary string when logging for example.

straight-shoota: I don't think that would be wise. It would be one random accessor in `URI` accepting an IO just because of how it's implemented.

Anyway, that'd be a tale for another day.

PR #6324, opened 2018-07-04
Fix: Disallow setters with multiple arguments

### LINKED ISSUE #5856: Setter method receiving multiple arguments
A setter method (ending in `=`) can be defined having multiple arguments:

```cr
def foo=(a, b)
  pp a, b
end
```

But such a method can't be called. Obviously `foo = 1, 2` won't work because it is interpreted as two expressions. Adding parenthesis around the arguments `foo=(1, 2)` raises a syntax error: `unterminated parenthesized expression` probably because `(` after `=` is not recognized as parenthesis of a call.

I'm not sure if it should be supported that setter methods can receive multiple arguments. It seems like this is too complicated and unnecessary.
But if you can't call such a method with more than one argument, it would make sense to raise a syntax error for the method definition with multiple arguments as well.

maxfierke:
Fixes #5856 by throwing a syntax error when defining a setter method that accepts more than one argument, seeing as there is no way to call it.


src/compiler/crystal/syntax/parser.cr
```diff
@@ -3344,6 +3347,10 @@ module Crystal
           index += 1
         end
 
+        if is_setter && args.size > 1
```
straight-shoota: Can't `is_setter` be replaced by `name.ends_with?('=')`?

asterite:
Also check named args, no splat, etc.

src/compiler/crystal/syntax/parser.cr
maxfierke: oh yeah! good catch

maxfierke:
@asterite added some checks for named args, splat, kwargs, and then also block args, since there's a similar issue calling those as well

src/compiler/crystal/syntax/parser.cr
```diff
@@ -3332,6 +3333,7 @@ module Crystal
           if block_arg = extras.block_arg
             compute_block_arg_yields block_arg
             check :")"
+            found_block = true
```
asterite: `!block_arg.nil? == found_block`

So the `found_block` var can be removed.

asterite: Ah, but I also see `found_splat`, etc., so it's fine to keep this variable 👍

spec/compiler/parser/parser_spec.cr
```diff
@@ -174,6 +174,13 @@ module Crystal
     assert_syntax_error "def foo!=; end", "unexpected token: !="
     assert_syntax_error "def foo?=(x); end", "unexpected token: ?"
 
+    # 5856
```
straight-shoota: `# #5856`

spec/compiler/parser/parser_spec.cr
```diff
@@ -174,6 +174,13 @@ module Crystal
...
+    assert_syntax_error "def foo=(a,b); end", "setter method 'foo=' cannot receive more than one argument"
```
straight-shoota: `s/receive/have/`. "receive" is more associated to the call site. Same for the others. Just my opinion.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -3344,6 +3346,14 @@ module Crystal
           index += 1
         end
 
+        if name.ends_with?('=') && !name.ends_with?("[]=")
```
straight-shoota: The second test can be `name == "[]="`. There can't be anything before `[]=` in a method name.

However, is there a particular reason why `[]=` can be called like any method but all other setters can't? 
If not, the same rules should apply (with ` args.size != 2`).

straight-shoota:
`yield` should also be a syntax error in a setter method. It's not part of the method signature, so it must be implemented differently, and doesn't need to be in this PR.

src/compiler/crystal/syntax/parser.cr
bew: With `[]=` you can do `var[1, 2] = 3` to implement a matrix type for example

straight-shoota: Okay, that's a great use case. You can even do var`[1, foo: 2] = 3` 👀 

But it still doesn't explain `foo.[]=(foo, bar: bar) do |baz| end` 😃

maxfierke: doh! I'll switch that to an equality check on `name`. @bew's totally right on the `[]=` with multi-arg case. There's at least one place in stdlib where `[]=` is given splat args and kwargs too.

Good catch on the block case for `[]=`. I'll address that.

spec/compiler/parser/parser_spec.cr
maxfierke: seems reasonable. `have` is a better hint at something static too.

src/compiler/crystal/syntax/parser.cr
asterite: https://github.com/crystal-lang/crystal/blob/8d28cbb1fcd8fc462425f13b79203aa30f64bf96/src/compiler/crystal/syntax/parser.cr#L649-L653

asterite: And because `(exp1, exp2)` is not a valid syntax, `f.x=(exp1, exp2)` is not a valid syntax either, so there's no way to pass multiple arguments there. It's the same as in Ruby (I copied it from Ruby, actually).

asterite: And the exception for splat mentioned there maybe should be removed (it doesn't work in Ruby, so...)

straight-shoota:
To clarify: Currently, `[]=` can be called with a block (`var.[]= {}`). I just think we shouldn't disallow the declaration without disallowing the call. Either we allow both or disallow both. I'm not sure about it... and it probably doesn't make a any difference whatsoever.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -687,6 +687,10 @@ module Crystal
 
             block = parse_block(block, stop_on_do: @stop_on_do)
             if block || block_arg
+              if name == "[]="
+                raise "setter method '[]=' cannot be called with a block"
```
bew: I think this check should go in semantic phase, not here in the parser?

maxfierke:
@bcardiff Are you referring to the [changes in src/object.cr](https://github.com/crystal-lang/crystal/pull/6324/files/9697515a55e8c08dfea59076c558c15ab2d005d5#diff-a1b7b85341f5f80709d73955200f851eR1095)?

blocks are disallowed for all setters with this PR. The special case here for `[]=` is to avoid generating block-supporting method forwards for `[]=`.

bcardiff:
Oh, I missed that it was inside a `method.id.ends_with?('=')`. So that's fine.

But I am still missing why there is an `if name == "[]="` in the changeset for `parser.cr`. Yet the specs clearly says that `foo=(&block)` is disallowed 🤔

maxfierke:
@bcardiff ahhh! I assume you're referring to [here](https://github.com/crystal-lang/crystal/pull/6324/files#diff-6723aec3601e19220d2c13db7320bb9fR690). That is to disallow calls to `[]=` with a block at the call site in response to this feedback: https://github.com/crystal-lang/crystal/pull/6324#issuecomment-404492002. This way, users get a compiler error instead of a silent failure if for some reason they're calling `[]=` with a block somewhere and expecting it to work (e.g. doing so caught the invalid generation via macro of the aforementioned block-supporting forward for `[]=` in `src/object.cr`)

bcardiff:
Ahhh! The parsing snippet is because `[]=` is parsed as `[`, _args_, `]`. `=`, etc. So the stressed path is not the same as `method=`

PR #5915, opened 2018-04-04
Fix File.join with empty path component

straight-shoota:
Fixes #5905 

It could be debated if an empty last component (`File.join("foo", "")`) should result in a path ending with a separator (`"foo/"`). That would be more complicated than this simple fix and is probably not that useful as `File.join("foo", "/")` yields the same result.

PR #5635, opened 2018-01-24
Implement Path type

straight-shoota:
This is a prototype of a Path type encapsulating filesystem paths to have a basis for further discussion in #5550 

For now I have only refactored `HTTP::StaticFileHandler` to use the Path type to see how it can be used.

PR #6328, opened 2018-07-04
Fix File.join with separator path component

icyleaf:
Simliar #5915.

It throws an exception if only separator in the middle component, (for example `File.join("foo", "/", "bar")`)

```
Unhandled exception: Index out of bounds (IndexError)
  from Slice(UInt8)@Slice(T)#pointer<Int32>:Pointer(UInt8)
  from Slice(UInt8)@Slice(T)#copy_to<Pointer(UInt8), Int32>:Pointer(UInt8)
  from String::Builder#write<Slice(UInt8)>:Nil
  from File::join<Tuple(String, String, String)>:String
  from File::join<String, String, String>:String
  from __icr_exec__:String
  from __crystal_main
  from Crystal::main_user_code<Int32, Pointer(Pointer(UInt8))>:Nil
  from Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32
  from main
```

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -305,6 +305,9 @@ describe "File" do
     File.join(["", "", "foo"]).should eq("foo")
     File.join(["foo", "", "bar"]).should eq("foo/bar")
     File.join(["foo", "", "", "bar"]).should eq("foo/bar")
+    File.join(["foo", "/", "bar"]).should eq("foo/bar")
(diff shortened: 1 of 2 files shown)


bcardiff:
Maybe we should wait on this after #5635 since the code is moving to other file.

straight-shoota:
This is already fixed in #5635, but it doesn't hurt to merge.

asterite:
It would be really nice to add comments inside the implementation. It's becoming really obscure.

PR #6330, opened 2018-07-04
Refactor codegen_binary_op

bcardiff:
This refactor is needed/handy as part of supporting overflowing operators.


bew:
So is `@last` set somewhere else? Or is it simply removed?

bcardiff:
@bew `@last` is used as the last llvm value and is set at `codegen_primitive` based on the return value of `codegen_primitive_*` methods. It was probably a left over of past refactors that the `@last` was mentioned in `codegen_primitive_binary`.

PR #5623, opened 2018-01-21
Implement File and Dir for win32

RX14:
This is fairly WIP, since it depends on #5584 which isn't finalized. It works though.

The main goal is to continue to reduce the amount of stubbing out in `file_spec` and `dir_spec`.

PR #6333, opened 2018-07-04
Extract platform-specifics from ENV to Crystal::System::Env and implement for win32

straight-shoota:
Requires #5623 (for win32).

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -582,41 +582,38 @@ describe "File" do
       File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
     end
 
-    # TODO: these specs don't compile on windows because ENV isn't ported
     # TODO: remove /\A\/\// hack after this is removed from macros
-    {% unless flag?(:win32) %}
-      it "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do
-        prev_home = home
-        begin
-          ENV["HOME"] = File.expand_path(datapath)
-          File.expand_path("~/").should eq(home)
-          File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-          File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
-          File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
-          File.expand_path("~").should eq(home)
-          File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
-          File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
-        ensure
-          ENV["HOME"] = prev_home
-        end
+    pending_win32 "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do
+      prev_home = home
+      begin
+        ENV["HOME"] = File.expand_path(datapath)
+        File.expand_path("~/").should eq(home)
+        File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
+        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
+        File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
+        File.expand_path("~").should eq(home)
+        File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
+        File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
+      ensure
+        ENV["HOME"] = prev_home
       end
+    end
 
-      it "converts a pathname to an absolute pathname, using ~ (home) as base (HOME=/)" do
-        prev_home = home
-        begin
-          ENV["HOME"] = "/"
-          File.expand_path("~/").should eq(home)
-          File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-          File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
-          File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
-          File.expand_path("~").should eq(home)
-          File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
-          File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
-        ensure
-          ENV["HOME"] = prev_home
-        end
+    pending_win32 "converts a pathname to an absolute pathname, using ~ (home) as base (HOME=/)" do
+      prev_home = home
+      begin
+        ENV["HOME"] = "/"
+        File.expand_path("~/").should eq(home)
+        File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
+        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
+        File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
+        File.expand_path("~").should eq(home)
+        File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
+        File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
+      ensure
+        ENV["HOME"] = prev_home
       end
-    {% end %}
+    end
   end
 
   describe "real_path" do
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -582,14 +582,13 @@ describe "File" do
       File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
     end
 
-    # TODO: remove /\A\/\// hack after this is removed from macros
     pending_win32 "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do
       prev_home = home
       begin
         ENV["HOME"] = File.expand_path(datapath)
         File.expand_path("~/").should eq(home)
         File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
+        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(%r{\A//}, "/"))
         File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
         File.expand_path("~").should eq(home)
         File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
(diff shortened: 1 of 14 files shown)


src/crystal/system/unix/env.cr
```diff
@@ -0,0 +1,55 @@
+require "c/stdlib"
+
+module Crystal::System::Env
+  # Sets an environment variable.
+  def self.set(key : String, value : String) : Nil
+    raise ArgumentError.new("Key contains null byte") if key.byte_index(0)
```
RX14: `key.check_no_bull_byte`

src/crystal/system/win32/env.cr
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  # Gets an environment variable.
+  def self.get(key : String) : String?
+    raise ArgumentError.new("Key contains null byte") if key.byte_index(0)
+
+    System.retry_wstr_buffer do |buffer, small_buf|
+      length = LibC.GetEnvironmentVariableW(key.to_utf16, buffer, buffer.size)
+      if 0 < length < buffer.size
+        break String.from_utf16(buffer[0, length])
```
RX14: This is probably *slightly* clearer using `return` instead of `break`.

src/env.cr
```diff
@@ -35,24 +35,15 @@ module ENV
...
-      LibC.unsetenv(key)
-    end
     value
   end
 
   # Returns `true` if the environment variable named *key* exists and `false`
   # if it doesn't.
   def self.has_key?(key : String) : Bool
-    !!LibC.getenv(key)
+    !!Crystal::System::Env.set?(key)
```
RX14: rename this to `has_key?` and remove the `!!`

src/lib_c/x86_64-windows-msvc/c/winnt.cr
```diff
@@ -6,4 +6,6 @@ lib LibC
   alias WCHAR = UInt16
   alias LPSTR = CHAR*
   alias LPWSTR = WCHAR*
+  alias LPCWSTR = WCHAR*
```
RX14: `LPCWSTR` shouldn't be an alias since it's just constant which means nothing ABI-wise.

src/lib_c/x86_64-windows-msvc/c/winbase.cr
```diff
@@ -5,6 +5,8 @@ require "c/int_safe"
 lib LibC
   fun GetLastError : DWORD
 
+  ERROR_ENVVAR_NOT_FOUND = 203_u32
```
RX14: https://github.com/crystal-lang/crystal/blob/master/src/winerror.cr#L253

src/crystal/system/unix/env.cr
straight-shoota: That's what I initially thought, too. But `check_no_null_byte` loses the information which of the arguments was invalid (it's `String contains null byte` instead of `Key contains null byte`). Doesn't make much difference, so I'm not sure about it...

src/lib_c/x86_64-windows-msvc/c/winnt.cr
straight-shoota: So... what should it be?

straight-shoota:
Rebased after #5623 was merged and ready for review 👍

src/lib_c/x86_64-windows-msvc/c/winnt.cr
RX14: `LPWSTR`...

src/crystal/system/unix/env.cr
RX14: Then refactor it to take an argument

src/string/utf16.cr
```diff
@@ -81,6 +73,41 @@ class String
...
+  # slice = Slice[104_u16, 105_u16, 0_u16, 55296_u16, 56485_u16, 0_u16]
+  # String.from_utf16(slice) # => "hi\0000𐂥"
+  # pointer = slice.to_unsafe
+  # string, pointer = String.from_utf16(pointer) # => "hi"
+  # string, pointer = String.from_utf16(pointer) # => "𐂥"
+  # ```
+  #
+  # Invalid values are encoded using the unicode replacement char with
+  # codepoint `0xfffd`.
+  def self.from_utf16(pointer : Pointer(UInt16)) : {String, Pointer(UInt16)}
```
RX14: This is a breaking change and as such needs some more discussion.

straight-shoota: Sure. When reading from a pointer it needs to advance the amount of `UInt16` read. We can't determine that afterwards from `String#bytesize` because UTF-8 and UTF-16 have different bytesizes. We could rename the method with this changed behaviour (to maybe `String.from_utf16_advancing_pointer`). Or change the signature in some way. But I don't think it makes much sense to have both a version of this method (with pointer) that returns the advanced pointer and one that doesn't. It's easy to just ignore the returned pointer value if you don't need it.

RX14:
I'm fine with this breaking change actually

src/string.cr
```diff
@@ -4244,8 +4244,11 @@ class String
   # Raises an `ArgumentError` if `self` has null bytes. Returns `self` otherwise.
   #
   # This method should sometimes be called before passing a `String` to a C function.
-  def check_no_null_byte
-    raise ArgumentError.new("String contains null byte") if byte_index(0)
+  def check_no_null_byte(name = nil)
+    if byte_index(0)
+      name = "`#{name}` "
+      raise ArgumentError.new("String #{name}contains null byte")
```
RX14: Why is this on two lines??

Also, I'd rather the entire exception message was passed in, i.e. `def check_no_null_byte(message = "String contains null byte")`.

RX14: The only thing you save by doing it the other way is a bit of typing, and the benefit of having fully customized error messages. For "Key contains null byte" vs "String \`key\` contains null byte" - the former is way better. Please, let's keep it simple and flexible.

straight-shoota: Two lines because I forgot a `if name` in the first line. To avoid empty argument to be printed.

Maybe full message is better. But actually, I don't think there is much benefit from it. Yes, it's more flexible, but I don't see any use for that. `check_no_null_byte` is only used to validate method arguments and for such it makes sense to declare the name of the argument if there are more than one. But apart from that, I don't see any need for customization.

src/string.cr
```diff
@@ -4244,8 +4244,11 @@ class String
   # Raises an `ArgumentError` if `self` has null bytes. Returns `self` otherwise.
   #
   # This method should sometimes be called before passing a `String` to a C function.
-  def check_no_null_byte
-    raise ArgumentError.new("String contains null byte") if byte_index(0)
+  def check_no_null_byte(name = nil)
+    if byte_index(0)
+      name = "`#{name}` " if name
```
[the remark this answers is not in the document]
straight-shoota: To show that this refers to a name. This formatting is used in many other compiler and stdlib error messages.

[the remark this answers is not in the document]
ysbaddaden: Weird, I never noticed that.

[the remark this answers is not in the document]
RX14: I think it's not that common

Sija:
IMHO `Key contains null byte` reads way better than `String 'key' contains null byte`.

src/string.cr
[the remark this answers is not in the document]
Sija: IMHO `Key contains null byte` reads way better than `String 'key' contains null byte`.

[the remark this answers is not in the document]
Sija: Why not just `"#{name || "String"} contains null byte"`?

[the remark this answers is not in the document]
asterite: I propose "Argument '#{name}' contains null byte".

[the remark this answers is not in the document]
straight-shoota: @asterite Do you mean `name` should be required?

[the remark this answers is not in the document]
RX14: Perhaps `check_no_null_byte(what = "String")` then just `#{what} contains null byte`?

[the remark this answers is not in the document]
asterite: I'm looking at the code, it always seems to be a check against an argument. To avoid repeating the argument name, maybe we can have:

```crystal
class String
  macro check_no_null_byte(name)
    raise ArgumentError.new("Argument '{{name}}' contains null byte") if {{name}}.byte_index(0)
    {{name}}
  end
end
```

Then it can be used like this:

```crystal
def some_method(arg)
  String.check_no_null_byte(arg)
  # do something
end
```

That way there's no string interpolation, and there's no need to repeat the argument name.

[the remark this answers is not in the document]
RX14: This is bikeshedding and overengineering. The diff is fine as-is.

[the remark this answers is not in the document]
straight-shoota: @RX14 I was writing the same thing. All are valid options and it doesn't make a difference regarding ENV.
Anyone can send a PR later to refactor `check_no_null_byte` if you care enough.

straight-shoota:
I discovered an issue on win32: `GetEnvironmentVariableW` doesn't differentiate between unset variable and empty string. Both have `length == 0` and return value `ERROR_ENVVAR_NOT_FOUND`. The latter seems incorrect in case of an empty string value. In contrast, `GetEnvironmentStringsW` contains variables with empty values.

This could be used to identify empty values if `length == 0` but will of course have a performance impact, and it's probably not really worth it for this edge case.

Unless someone has a idea how to circumvent this, I don't think there is anything to be done.

RX14:
@straight-shoota i'm pretty skeptical that `GetEnvironmentVariable` is behaving this way. Can you print the result of `GetLastError` (`pp WinError.new`) before and directly after the API call?

RX14:
This stackoverflow answer seems to show the Win32 API correctly returning the value of an empty env var: https://stackoverflow.com/questions/26993642/call-to-environment-getenvironmentvariable-affects-subsequent-calls

But it does show there's definitely weirdness around this part of the API - perhaps `GetEnvironmentVariable` is failing to clear `GetLastError` correctly as all syscalls should.

RX14:
Which is definitely a bug, because even a call to `GetLastError` sets the last error to `ERROR_SUCCESS`. Oh well, glad that's sorted.

RX14:
Oh actually, can you uncomment out the file/dir specs which were commented out due to using ENV?

straight-shoota:
As far as I can see, they're all pending for other reasons.

RX14:
@straight-shoota these should go from macroed out to just pending https://github.com/crystal-lang/crystal/blob/8d28cbb1fcd8fc462425f13b79203aa30f64bf96/spec/std/file_spec.cr#L585

spec/std/file_spec.cr
```diff
@@ -582,41 +582,38 @@ describe "File" do
       File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
     end
 
-    # TODO: these specs don't compile on windows because ENV isn't ported
     # TODO: remove /\A\/\// hack after this is removed from macros
```
RX14: please check the diff for #5623 and do this TODO too please

straight-shoota: Sorry, I'm not sure what you mean... reinstating `%r{\A//}`?

RX14: yes

RX14:
ping, needs a second review

straight-shoota:
@RX14 The commit message is a mess... it should've been squashed without all these fixups.

Issue #6226, opened 2018-06-20
to_big_i doesn't work out of the box


laginha87:
I was trying to run the following code:

```
puts "3a060dbf8d1a5ac3e67bc8f18843fc48".to_i(16)
```

This throws an error "Invalid Int32".
After some digging around in the String doc page I found the method I needed was ```to_big_i(16)```.
Like this:

```
puts "3a060dbf8d1a5ac3e67bc8f18843fc48".to_big_i(16)
```
but this throws "Undefined method to_big_i for String". I later found out I had to require the big package because that is where the string method is defined. So this is the working example: 

```
require "big"
puts "3a060dbf8d1a5ac3e67bc8f18843fc48".to_big_i(16)
```

It makes sense but looking at the docs page it is unclear that I needed to require the "big" module.

I'm sure this happens with other packages because not everything is included in the prelude.cr file.
I think the docs should inform when a method is defined in a different file/module that isn't included by default.

```
crystal -v
Crystal 0.25.0 (2018-06-15)

LLVM: 5.0.1
Default target: x86_64-apple-macosx
```

RX14:
1) would be the simplest, and possible to do right now.

laginha87:
I think one solution doesn't exclude the other, having a good code sample in my opinion makes it a lot easier for whoever's reading the docs.
2. Seems like a nice to have but a bit hard to implement.
Can I open up a PR with docs with code samples in ?

RX14:
yes, please do open a PR!

Another thing to do, "Invalid Int32" is a bit of a nondescriptive message, we should mention that the number is too large/would overflow.

Sija:
This can be closed?

asterite:
Yes, let's close this.

PR #6336, opened 2018-07-05
Improve docs of big module overloads

laginha87:
Fix for https://github.com/crystal-lang/crystal/issues/6226 by adding code examples to the docs that import the `big` module.

It was also mentioned in the issue to improve the error message in case of overflow, but from what I looked at the code it seems like there is no simple way to add that without changing the code significantly.
https://github.com/crystal-lang/crystal/blob/e806c95/src/string.cr#L407

### DIFF
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -426,6 +426,10 @@ struct Int
   include Comparable(BigDecimal)
 
   # Converts `self` to `BigDecimal`.
+  # ```
+  # require "big"
+  # 1212341515125412412412421.to_big_d
+  # ```
   def to_big_d
     BigDecimal.new(self)
   end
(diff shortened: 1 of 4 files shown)


Sija:
Convention is to add `require` statement (in the examples) only once, in the module docs. See [`Colorize`](https://crystal-lang.org/api/0.25.1/Colorize.html) for instance.

asterite:
@Sija These are methods added to `Int` and `Float`, it makes sense to show that `require "big"` is needed for those methods to work.

asterite:
@laginha87 Please rebase this PR's branch against master, CI is failing because of things that are already fixed in master.

laginha87:
@RX14, @BenDietze  thanks for looking into the PR, I made the requested change and rebased on master again, should be good to go.

PR #6341, opened 2018-07-06
Fix: missing virtual types in literal type guessing

### LINKED ISSUE #5342: Compiler error when using generic type in class inheritance
Hi, 

While working on a project I found this compilation error

> instance variable '@ary' of Second must be Array(Array(Int64)), not Array(Array(Int64))

The minimal code i found that trigger this error is

~~~crystal
class First(T) < Array(T)
end 

class Second
  def initialize(@ary = [ [ ] of Int64 ] )
  end 
end 

a = Second.new
~~~

I found these related issues, but  I'm no sure is the  same because #4753  is a compiler crash , #4758  is a invalid memory access and both are when calling the child class, and in this case is the parent.

Also this code 

~~~crystal
class First(T) < Array(T)
end 

class Second
  @ary = [ [ ] of Int64 ]
end 
~~~

triggers 

> instance variable '@ary' of Second must be Array(Array(Int64)), not (Array(Array(Int64)) | Array(Array(Int64)))

Crystal version
~~~
Crystal 0.23.1 [e2a1389] (2017-07-13) LLVM 3.8.1
~~~

Playground Link [36p5](https://play.crystal-lang.org/#/r/36p5)
Playground Link [36p9](https://play.crystal-lang.org/#/r/36p9)

Relate to generic type inheritance i found #5243, #5125 , #4602 and a compilation of issues at #2665

asterite:
Fixes #5342

A few `virtual_type` calls were missing.

### DIFF
--- a/spec/compiler/semantic/instance_var_spec.cr
+++ b/spec/compiler/semantic/instance_var_spec.cr
@@ -4928,4 +4928,118 @@ describe "Semantic: instance var" do
       ),
       "can't use Gen(T) as the type of instance variable @x of Foo, use a more specific type"
   end
+
+  it "guesses virtual array type (1) (#5342)" do
+    assert_type(%(
+      require "prelude"
+
+      class First(T) < Array(T)
+      end
+
+      class Second
+        @ary = [[1]]
+
+        def ary
+          @ary
+        end
+      end
+
+      Second.new.ary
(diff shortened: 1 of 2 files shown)


asterite:
Yeah, right now virtual types make the language really complex, so having that by defaut might be a good idea.

PR #6342, opened 2018-07-06
Don't merge tuple metaclasses through instance types

### LINKED ISSUE #5384: Bug with multiple assignments of tuple types to a variable
Building the code

```crystal
x = typeof({1})
x = typeof({"b"})
```

brings up the message

```
BUG: trying to assign Tuple(Int32 | String):Class <- Tuple(Int32):Class
```

This happens both with Crystal 0.23.1 and Crystal 0.24.0-2 on Ubuntu 16.04 LTS.

asterite:
Fixes #5384

### DIFF
--- a/spec/compiler/semantic/tuple_spec.cr
+++ b/spec/compiler/semantic/tuple_spec.cr
@@ -289,4 +289,15 @@ describe "Semantic: tuples" do
       ),
       "tuple too big"
   end
+
+  it "doesn't unify tuple metaclasses (#5384)" do
+    assert_type(%(
+      Tuple(Int32) || Tuple(String)
+      )) {
+      union_of(
+        tuple_of([int32] of Type).metaclass,
+        tuple_of([string] of Type).metaclass,
+      )
+    }
+  end
 end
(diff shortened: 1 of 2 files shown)


asterite:
To explain this a bit more...

When you have:

```crystal
x = {1} || {"a"}
```

what type does `x` has? Normally it's the union of both types, so `Tuple(Int32) | Tuple(String)`. However, for this particular case, the compiler turns it into `Tuple(Int32 | String)`. This is done for the cases where you do stuff like this:

```crystal
def foo
  if some_condition
    {:error, null}
  else
    {:ok, 1}
  end
end
```

The return type is `Tuple(Symbol, Int32?)`, and it's easier to deal with (you probably don't expect a union of tuples but a tuple of unions).

But maybe that wasn't a good idea after all (it makes the compiler's code *a lot* more complex).

Back to the original issues, when deciding what's the type of a union of metaclasses, we usually do it first for the instance types, and then obtain the metaclass from there. The bug was that if this is done for tuple metaclasses, the type is the metaclass of a tuple of unions (incorrect), instead of the metaclass of a union of tuples (correct). So for tuple metaclasses it's always just the simple union of metaclasses.

PR #6343, opened 2018-07-06
Remove leftover debug messages in raise.cr

hugoabonizio:
SSIA

### DIFF
--- a/src/raise.cr
+++ b/src/raise.cr
@@ -66,14 +66,11 @@ end
     if LibUnwind.__gnu_unwind_frame(ucb, context) != LibUnwind::ReasonCode::NO_REASON
       return LibUnwind::ReasonCode::FAILURE
     end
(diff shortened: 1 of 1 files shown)


wontruefree:
what about wrapping them in a debug conditional?

ysbaddaden:
I struggled with exceptions on ARM and damn, they helped a lot. Keep 'em in, commented, because they're useless to 99.99% of users they'd just clutter the compiler with macros for each and every compilation.

RX14:
I think this is fine as-is

hugoabonizio:
@ysbaddaden makes sense, just one question: why they are more useful on `raise.cr` than on other files?

RX14:
Because when raise breaks it's harder to debug.

hugoabonizio:
Thanks for clarifying!

Issue #6340, opened 2018-07-06
Is Crystal slower than Ruby in generating randoms?


lxxxvi:
Hey there. I've just got a question...

# Is Crystal slower than Ruby in generating randoms?

While playing around with Crystal I noticed that Ruby is quite faster than Crystal in generating randoms.

I wrote a [Crystal script](https://github.com/lxxxvi/ruby-vs-crystal-random-performance/blob/master/random_performance.cr) that performs 10 million randoms (`hex`, `base64`, `rand` and `random_bytes`) and measures the times.

I did the same in Ruby and found that the [Ruby script](https://github.com/lxxxvi/ruby-vs-crystal-random-performance/blob/master/secure_random_performance.rb) is significantly faster than Crystal. 

Is this expected? Are Crystal randoms more secure and thus more expensive?

| Generator       | Amount | Time in Crystal | Time in Ruby |
|:--------------- |-------:|----------------:|-------------:|
| `hex`           |    10M |         ~ 47.4s |       ~ 8.8s |
| `base64`        |    10M |         ~ 52.1s |      ~ 10.6s |
| `rand`          |    10M |         ~ 43.8s |       ~ 6.1s |
| `random_bytes`  |    10M |         ~ 47.8s |       ~ 5.8s |

(In Crystal I used `Random`, in Ruby I used `SecureRandom`)

```shell
$ crystal -v
Crystal 0.25.1 (2018-06-29)

LLVM: 5.0.2
Default target: x86_64-apple-macosx
```

```shell
$ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin17]
```

Environment

* macOS High Sierra 10.13.5 (17F77)
* 1.4 GHz Intel Core i7
* 8 GB 1867 MHz LPDDR3

(This description and all scripts are available at https://github.com/lxxxvi/ruby-vs-crystal-random-performance )

kostya:
did you compile it with --release? for me 0.25.1 is 10-20% faster than ruby.

ysbaddaden:
Comparing what's comparable, namely the default random and the secure random source (`/dev/urandom`):

```ruby
# ruby
require "benchmark"
require "securerandom"

N = 1_000_000
RANGE = 0...999_999

Benchmark.bm do |x|
  x.report("rand()") { N.times { rand(RANGE) } }
  x.report("SecureRandom.rand") { N.times { SecureRandom.rand(RANGE) } }
[...]

PR #6350, opened 2018-07-07
Switches Darwin to arc4random

chris-huxtable:
Based on discussions on #6340.

- Performance improvements
- Works in chroot
- No File Descriptors needed

Note: The same could also be done for FreeBSD.

### DIFF
--- a/src/lib_c/x86_64-macosx-darwin/c/stdlib.cr
+++ b/src/lib_c/x86_64-macosx-darwin/c/stdlib.cr
@@ -7,6 +7,8 @@ lib LibC
     rem : Int
   end
 
+  fun arc4random : UInt32
+  fun arc4random_buf(x0 : Void*, x1 : SizeT) : Void
   fun atof(x0 : Char*) : Double
   fun div(x0 : Int, x1 : Int) : DivT
   fun exit(x0 : Int) : NoReturn
--- a/src/crystal/system/random.cr
+++ b/src/crystal/system/random.cr
@@ -12,7 +12,7 @@ end
 
 {% if flag?(:linux) %}
   require "./unix/getrandom"
-{% elsif flag?(:openbsd) %}
+{% elsif flag?(:openbsd) || flag?(:darwin) %}
   require "./unix/arc4random"
 {% elsif flag?(:unix) %}
   require "./unix/urandom"
--- a/src/crystal/system/unix/arc4random.cr
+++ b/src/crystal/system/unix/arc4random.cr
@@ -1,4 +1,4 @@
-{% skip_file unless flag?(:openbsd) %}
+{% skip_file unless flag?(:openbsd) || flag?(:darwin) %}
 
 require "c/stdlib"
 


src/lib_c/x86_64-macosx-darwin/c/stdlib.cr
```diff
@@ -7,6 +7,8 @@ lib LibC
     rem : Int
   end
 
+  fun arc4random : UInt32
+  fun arc4random_buf(x0 : Void*, x1 : SizeT) : Void
```
asterite: Please:
- Replace `Void*` with `UInt8*`
- Remove `: Void`

chris-huxtable: Should this be the case even if it goes against the existing c definition?

```c
void
arc4random_buf(void *buf, size_t nbytes);
```

asterite: I guess that works... I thought `Void*` didn't work in Crystal, but for some reason it does...

It seems there's also a `Void` top-level type... it seems it was never removed from the language...

RX14: The `: Void` is implicit and the `void *` should be `UInt8*` in crystal. @asterite is correct.

chris-huxtable:
@RX14  - Based on my current research. 

- `arc4random` introduced in macOS 10.6 (2009)
- `arc4random_buf` introduced in macOS 10.7 (2011)

- `arc4random` is reseeded with `getentropy`, **not** `/dev/[u]random`
- The documentation explicitly states to use `arc4random` as it is safe in `chroot`. 

`man arc4random`:
> They can be called in almost all environments, including chroot(2), and their use is encouraged over all other standard library functions for random numbers.

- `arc4random` is based on an AES cipher as of macOS 10.12 (2016)

- With regards to whether you should use `arc4random` or `/dev/urandom/`. The documentation is explicit. You should always use `arc4random`.

`man random`:
> Applications requiring cryptographic quality randomness should use arc4random(3).

`man urandom`:
> The same random data is also available from getentropy(2).  Using the getentropy(2) system call interface will provide resiliency to file descriptor exhaustion, chroot, or sandboxing which can make /dev/random unavailable. Additionally, the arc4random(3) API provides a fast userspace random number generator built on the random data source and is preferred over directly accessing the system's random device.

`man getentropy`:
>  However, it should be noted that getentropy() is primarily intended for use in the construction and seeding of userspace PRNGs like arc4random(3) [...].

chris-huxtable:
@RX14 - You can trust whoever you like. I don't trust the libsodium guys. I happen to be of the perspective everyone should just duplicate OpenBSD's `arc4random` and keep it up to date. That isn't the case though. What is the case is macOS has `/dev/urandom` and their implementation of `arc4random`. 

All the official documentation says use `arc4random`, as does the preponderance of unofficial documentation/discussions. `arc4random` is a better choice then `/dev/urandom` on macOS.

RX14:
And until recently the linux manpages said using urandom wasn't cryptocraphically secure. Which was against the advice of almost every cryptographer out there.

At the end of the day, this is a *performance* problem, and if there's any conflict at all, i'd rather stick with what we know works and is secure.

konovod:
So, there are two main problems with arc4random
 - it is userspace (so some side-channel attacks can be possible, but this possibility is debatable). There are controversial opinions about it.
 - it can use RC4 (that is nonuniform and generally broken) on old systems.
Checking the source confirms, that macOS 10.12 (Sep 20, 2016) uses AES and 10.11 uses (modified but still unreliable) RC4 seeding it from /dev/random.
https://opensource.apple.com/source/Libc/Libc-1158.1.2/gen/FreeBSD/arc4random.c.auto.html
https://opensource.apple.com/source/Libc/Libc-1081.1.3/gen/FreeBSD/arc4random.c.auto.html

so even if we ignore first point, arc4random shouldn't be used on OSX10.11 and older.

On the other hand, `getentropy` was introduced in macOS 10.12 (Sep 20, 2016) and is a (strictly) better solution then `/dev/urandom` - it yields same data (not a userspace prng) and is safer because doesn't depends on file descriptors. So if a `getentropy` is available - it is a modern OSX and we can use either `arc4random` or `getentropy` (depending on the opinion about first point).

On the other other hand, it a libsodium issue cited before tells that `getentropy` uses `\dev\urandom`. I think the additional research is needed on this, `man` page isn't a proof.
From what Apple opensourced
```
if (getentropy(buf, size) == 0) {
		return;
	}
	// The above should never fail, but we'll try /dev/random anyway
```
so it in theory can fallback to reading from `/dev/random`, but does it possible in practice depends from can getentropy fail. I wasn't able to find `getentropy` source to check.

RX14:
And even if arc4random is secure on 10.12, are all the runtime checks and complexity (read: vulnerabilities) worth it? Just for some performance?

ysbaddaden:
Closing. We can't trust ourselves on security topics. We need strong guarantees from cryptographers that something is safe and correct.

Sadly, we don't have enough guarantees that `arc4random` is cryptographically correct on macOS. The only guarantee we have is that it was _very bad_ prior to macOS 10.12 :disappointed:

PR #6024, opened 2018-04-28
Don't automatically virtualize two types in the same hierarchy, unless one is deeper than the other

asterite:
Fixes #2661
Fixes #4303

What this means is that with this PR, the language will now behave like this:

```crystal
class Foo
end

class Bar < Foo
end

class Baz < Foo
end

var = rand < 0.5 ? Bar.new : Baz.new
typeof(var) #=> Bar | Baz
```

Previously the type would go up to `Foo`.

However:

```crystal
var = rand < 0.5 ? Bar.new : Foo.new
typeof(var) #=> Foo
```

because `Bar < Foo`, so that gets "upcasted" to `Foo` (or `Foo+`, if you know what I mean).

This makes it possible to have more accurate information about the type of a variable. For example in the compiler's source code we have a `TypeDeclaration` node whose `var` is of type `ASTNode`, where in reality it can only be `Var | InstanceVar | ClassVar | Global`. So the compiler has some "otherwise, raise bug" that could be removed.

Somehow, compilation times for the compiler don't get much more slower. Memory consumption goes a bit higher (most probably because now there are some union types that previously would go up to `ASTNode` but not now, so a bit more duplicated code is generated), but I don't think we should care anymore about compile times or memory consumption of the compiler (Slack alone, or Chrome, already consume 1GB from your computer and nobody seems to complain... or maybe yes, but well, "if they do it..."). Or, as usual, we can worry about such things later, or maybe a superhero will come with a compiler rewrite or some patches to improve performance 😛 

**PLEASE**, if you can, checkout this branch, compiler the compiler, and try it in your projects. Many things can happen:
1. Your code won't compile for some reason. This usually happens when you have `[foo, bar]`, you'll probably now have to write `[foo, bar] of T` for some `T` because (taking the above example), `Array(Foo)` is not compatible with `Array(Bar | Baz)`.
2. Your code will explode (after fixing possible errors in 1). This is what I'd like to know, as it probably means this PR is buggy.

---

Once and if we can go ahead with this change, I'd like to revisit #4837, as it can be another great addition for more type safey and accuracy in the language, avoiding useless `else raise "Impossible"` and generally making the code more robust to changes.

PR #6351, opened 2018-07-07
Revert: Don't automatically virtualize two types in the same hierarchy, unless one is deeper than the other

### LINKED ISSUE #6349: Compiler crashes due to an invalid memory access
As asked in bug #6195 I report here my issue.
I get an invalid memory access error on my project at https://github.com/LinCAS-lang/LinCAS when I try to compile it (using `crystal build --no-debug ./src/LinCAS.cr`).
It just gives me as a message

```
Invalid memory access (signal 11) at address 0x7fff9d800d58
[0x55580ef9c0b6] ???
[0x55580e67b8ca] ???
[0x5558103e4a19] ???
```

and nothing more.
I have currently installed the version 0.25.1. I guess it's a bug which belongs to the 0.25.x versions, since with the previous ones everything worked fine.
I've some troubles in trying to reduce the code since I'm dealing with thousands an thousands lines of interconnected code, and the error doesn't provide me much information about the cause of the segmentation fault.
Is there any way to to get more details?

P.S. I got the same error on Travis CI

asterite:
Fixes #6349

Reverts https://github.com/crystal-lang/crystal/pull/6024/commits/26635f1052b99ab5b52a8b051020a25435dba751

To make it clear:
- PR #6024 was just an experiment. I expected compile time for the compiler to go to the roof with that change but that didn't happen.
- However, it did happen for other projects, namely the one mentioned in #6349 
- The main problem is that with a big type hierarchy, say N types, all different union types inside that hierarchy is basically a combinatorial explosion (take any M from N where M <= N). The compiler potentially has to create all these different union types, and consider them for method overloads, dispatch, etc. In the given program this happened (a lot it seems), to the point where the compiler crashed
- Reverting this change makes the above project compile again: "Semantic (main)" takes about 3.3 seconds with a non-release compiler. Total compilation time takes about 31 seconds, again with a non-release compiler (that is, I didn't pass `--release `when I compiled the compiler). EDIT: "Semantic (main)" takes about 1 second with a release compiler. And it's about 15K lines of code :-O

Not unifying types in the same hierarchy was something that I, and probably others, wanted, but it's simply not possible. That's the whole reason why we unified types in the first place: to avoid this combinatorial explosion of union types.

And remember, this is not that terrible: we've been doing just fine before 0.25.0 without this "feature", and without me trying to play with this it would never have happened, so... :-)

### DIFF
--- a/samples/sdl/fire.cr
+++ b/samples/sdl/fire.cr
@@ -321,7 +321,7 @@ point_count = ARGV.size > 0 ? ARGV[0].to_i : 4
 yellow = YellowColorPattern.new
 magenta = MagentaColorPattern.new
 cyan = CyanColorPattern.new
-rainbow = RainbowColorPattern.new [yellow, magenta, cyan] of ColorPattern
+rainbow = RainbowColorPattern.new [yellow, magenta, cyan]
 
 main_points = [] of MainPoint
 main_points << MainPoint.new(50, 50, -Math::PI / 8, 1.4, yellow)
--- a/spec/compiler/codegen/class_spec.cr
+++ b/spec/compiler/codegen/class_spec.cr
@@ -504,15 +504,7 @@ describe "Code gen: class" do
 
   it "does to_s for virtual metaclass type (3)" do
     run(%(
-      class Class
-        def name : String
-          {{ @type.name.stringify }}
-        end
-
-        def to_s
-          name
-        end
-      end
+      require "prelude"
 
       class Foo; end
       class Bar < Foo; end
--- a/spec/compiler/semantic/class_spec.cr
+++ b/spec/compiler/semantic/class_spec.cr
@@ -178,7 +178,7 @@ describe "Semantic: class" do
       end
 
       a = Bar.new || Baz.new
-      ") { union_of types["Bar"], types["Baz"] }
+      ") { types["Foo"].virtual_type }
   end
 
   it "types class and subclass as one type" do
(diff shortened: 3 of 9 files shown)


ysbaddaden:
I guess it's acceptable.

Given `A`, and `B < A` and `C < A` types, is a specific `B | C` union inferred as `A` with this revert? I believe this was the actual issue we had, right? Inferring `Array(A)` from `[B.new, C.new]` is fine, and usually what we want, but a specific `[] of B | C` could be nice, sometimes.

RX14:
@ysbaddaden `B | C` always becomes `A`. Tracking whether a union was "explicit" (`.as`) or "implicit" (result of multiple dispatch, an `if`, etc.) would be far too much complexity and make things very difficult. You either allow `B | C` and it gets everywhere and causes slowness, or you always make it `A` and loose some expressive power.

ysbaddaden:
Sigh. Too bad then.

asterite:
My personal use case for not merging unions and go to the parent type was having an instance variable with a union type. For example a TypeDeclaration node can only have a Var, InstanceVar or ClassVar as targets, but these get condensed to ASTNode, so when dealing with this we have to do "else raise bug".

I'm thinking whether we can make at least that case work, so when you explicitly define a type declaration like that it keeps it. It would be a special type of union that doesn't go to the parent type when merging it with other types. I'll have to play with this idea, not sure it can work or it's easy. 

But then, maybe it'll be too confusing, because sometimes it will get merged, sometimes not, and it will be hard to track. So initially I'd like this to get merged, and then we can think alternative solutions.

PR #5765, opened 2018-03-03
Add MIME registry

straight-shoota:
This adds a `MIME` type which works as registry for mime types and allows to look up the mime type registered for each file name extension.

By default, a limited set of mime types is always registered, including those that have previously been used in `HTTP::StaticFileHandler`. Additionally, there is an implementation in `Crystal::System::MIME` that tries do load a mime database from the operating system. On unix systems the following paths are looked up:
```ini
/etc/mime.types
/etc/apache2/mime.types
/etc/apache/mime.types
# only freebsd
/usr/local/etc/mime.types
# only openbsd
/usr/share/misc/mime.types
```
For windows this is not yet implemented because we're missing access to the registry.

Applications (or shards) can also register custom mime types and extensions.

Closes #4632

PR #6354, opened 2018-07-07
HTTP: consider quoted charset

### LINKED ISSUE #6353: HTTP response with content-type `text/html; charset="utf-8"` breaks HTTP::Client
It seems Facebook's response has:

```
text/html; charset="utf-8"
```

as the content-type header. We are not removing the quotes, which seem to be valid (there are tons of RFCs for this, but they do appear in https://tools.ietf.org/html/rfc7231

Let's fix this by checking if there are quotes.

asterite:
Fixes #6353

### DIFF
--- a/spec/std/http/client/response_spec.cr
+++ b/spec/std/http/client/response_spec.cr
@@ -266,6 +266,12 @@ class HTTP::Client
       response.charset.should eq("UTF-8")
     end
 
+    it "returns content type and charset, removes quotes" do
+      response = Response.new(200, "", headers: HTTP::Headers{"Content-Type" => %(text/plain ; charset="UTF-8")})
+      response.content_type.should eq("text/plain")
+      response.charset.should eq("UTF-8")
+    end
+
     it "returns content type and no charset, other parameter (#2520)" do
       response = Response.new(200, "", headers: HTTP::Headers{"Content-Type" => "text/plain ; colenc=U"})
       response.content_type.should eq("text/plain")
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -214,6 +214,12 @@ module HTTP
           key = piece[0...eq_index].strip
           if key == "charset"
             value = piece[eq_index + 1..-1].strip
+
+            # For the case of a quoted charset like charset="utf-8"
+            if value.starts_with?('"') && value.ends_with?('"')
+              value = value[1...-1].strip
+            end
+
             return ComputedContentTypeHeader.new(content_type, value)
           end
         end


jhass:
Do we also want to allow single quotes (`'`) here?

asterite:
I'm not sure single quotes are allowed, but if others agree I can add it to this PR.

straight-shoota:
No, single quotes should not be interpreted as string delimiters, see definition of `quoted-string` in [RFC 723 §3.2.6](https://tools.ietf.org/html/rfc7230#section-3.2.6),

RX14:
Don't a lot of HTTP headers have the standard `value; optional-key="value"` format? Is there an RFC or standard of these headers - or even an unofficial standard - to implement a parser for them?

straight-shoota:
@RX14 There are a ton of RFCs about mime type syntax. The most recent one is [RFC 7321](https://tools.ietf.org/html/rfc7231).

As already mentioned, I have a working parser implemented, it just needs #5765 and a little bit of finishing.

RX14:
I'm not talking about mime types - i'm talking about the generic HTTP header value syntax.

straight-shoota:
Yeah, I know what this is about. It's not specific to any protocol, neither HTTP nor mail. I'm not sure if there's a proper name, but this is commonly described as mime type (examples: [.NET](https://msdn.microsoft.com/en-us/library/system.net.mime.contenttype.parameters(v=vs.110).aspx), [Java](https://docs.oracle.com/javase/7/docs/api/javax/activation/MimeType.html#MimeType(java.lang.String)), [Rust](https://docs.rs/mime/0.3.8/mime/struct.Mime.html), [Go](https://golang.org/pkg/mime/#TypeByExtension))

straight-shoota:
And despite any of this, it's still true that Crystal needs a proper parser implementation.

RX14:
My bad, that was mime types; i'm a bit tired. I meant that mime types, set-cookie, and content-disposition have the same approximate header syntax with one or more initial values then the same `; key=value` format.

straight-shoota:
As far as I know, this is called mime type, despite the fact that applications like `Set-Cookie` have nothing to to with mime. And yes, we should have a generic parser for this.

RX14:
I've never seen it named as such and I can't imagine it being called that since it would be so confusing. "mime type" already means a specific concrete RFC.

straight-shoota:
Sry, I meant media type. Doesn't make any difference, though. I guess it's just that this concept initially comes from mime type declarations and was adopted for other uses as well, without there being a proper name for this syntax.

Implementation wise, it is a MIME media type parser. This implementation can be applied to non-MIME contexts such as `Set-Cookie` or they can use a more specific, custom parser. Go for example uses `ParseMimeType` for `Content-Type`, `Content-Disposition` but a custom implementation which only recognizes a fixed set of attributes for `Set-Cookie`.

RX14:
I'd never find such a parser as a user if I had to think "mime". So I think we shouldn't name it that way. It should be in the generic `HTTP` module.

ysbaddaden:
Merging the fix. Discussion feels unrelated or that it could come later.

RX14:
Yep, discussion was unrelated, sorry.

PR #6355, opened 2018-07-07
Encoding: compare with case insensitive UTF-8 to avoid using iconv when not necessary

asterite:
We have `IO#set_encoding` which accepts an encoding. If it's UTF-8 there's nothing to do, because IO is assumed to be UTF-8 by default (regarding text). The current comparison is case sensitive, but in some cases the encoding can come with other cases (like in #6353). So to avoid using iconv, which should be slower then just doing it in Crystal (and also a bit redundant), I added a case insensitive comparison.

Update: includes tests.

### DIFF
--- a/spec/std/io/io_spec.cr
+++ b/spec/std/io/io_spec.cr
@@ -689,6 +689,18 @@ describe IO do
         end
       end
 
+      it "sets encoding to utf-8 and stays as UTF-8" do
+        io = SimpleIOMemory.new(Base64.decode_string("ey8qx+Tl8fwg7+Dw4Ozl8vD7IOLo5+jy4CovfQ=="))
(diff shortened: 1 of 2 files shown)


jhass:
Do we want to allow `utf8` (without the dash) too?

asterite:
@jhass Maybe, I'll later check if that's valid for iconv.

RX14:
I don't think it is - at least it's behind a `-DUSE_HPUX_ALIASES` compiler flag which is not the default.

RX14:
Oh - `glibc` iconv supports `utf8`: https://github.com/bminor/glibc/blob/master/iconv/iconv_open.c#L34

asterite:
Updated to check against "UTF-8" and "UTF8"

PR #6358, opened 2018-07-09
Restrict `exp.@var` to types in the same hierarchy

### LINKED ISSUE #6066: Avoid ivars to be read on objects without getter
Hi!

Crystal allows access to instances variables using `instance.@var`? Shouldn't this be avoided?

I think a `getter` should be required to do that

```crystal
class Foo
  @foo = 1
end

f = Foo.new
puts f.@foo # => 1
```

https://carc.in/#/r/40an Why?

asterite:
Fixes #6066

With this PR, doing `exp.@var` only works if:

1. the current type is the same as the type of `exp`, or
2. the current type inherits the type of `exp`, or
3. the type of `exp` inherits the current type, or
4. ~the current type and the type of `exp` live in the same namespace~

So, freely accessing instance vars is disallowed. Of course, one can always reopen a class and add an accessor (which is why I didn't see the current behaviour as a big problem, and why probably Ruby has `instance_variable_get`, because one can always access an instance var if you want to). But, at least we are not encouraging its use.

I was hesitant about point 4. All 4 points mean (Crystal, not Ruby) `protected`. Without 4, it's almost `protected`. At first I implemented just 1, 2 and 3, but then noticed it's used in `Hash::BaseIterator`:

https://github.com/crystal-lang/crystal/blob/4771dcb2125dfdda88da5afad7b2ed64391bb820/src/hash.cr#L958

So I added rule 4. However, later I noticed it's the **only** place where that "in the same namespace" rule is used for instance vars, at least in the standard library.

But, I think accessing instance vars like that should be mainly used for struct comparison, or accessing instance vars from types in the same hierarchy. If you really need to access an instance var from an unrelated type, even in the same namespace, you can always define a `protected` method, which gives you that access (like I did in this PR).

I was also hesitant to remove point 4, because now you can't say "`exp.@var` has protected access", but I expect a doc page explaining those three points without mentioning `protected` access, and we are done. Plus, it's a super rare thing to need, so understanding this rule and how it's different from `protected` won't be needed for most of users.

Separate notes:
- I had to add `IO#has_non_utf8_encoding?` as `:nodoc:` because `Char#to_s` uses it. The other alternative is to do `io.encoding == "UTF-8"` but that's probably much slower.
- For specs, I added accessors like `_var` to test the value of some instance vars. These are necessary because the tests check implementation details.

### DIFF
--- a/spec/compiler/semantic/read_instance_var_spec.cr
+++ b/spec/compiler/semantic/read_instance_var_spec.cr
@@ -1,0 +1,148 @@
+require "../../spec_helper"
+
+describe "Semantic: read instance var" do
+  it "can read from same class" do
+    assert_type(%(
+      class Foo
+        @x = 1
+
+        def foo(other)
+          other.@x
+        end
+      end
+
+      Foo.new.foo(Foo.new)
+      )) { int32 }
+  end
+
+  it "can read from subclass" do
+    assert_type(%(
+      class Foo
+        @x = 1
+
+        def foo(other)
+          other.@x
+        end
+      end
+
+      class Bar < Foo
+      end
+
+      Foo.new.foo(Bar.new)
+      )) { int32 }
+  end
+
+  it "can read from superclass" do
+    assert_type(%(
+      class Foo
+        @x = 1
+      end
+
+      class Bar < Foo
+        def foo(other)
+          other.@x
+        end
+      end
+
+      Bar.new.foo(Foo.new)
+      )) { int32 }
+  end
+
+  it "can read from different instantiations of a generic type" do
+    assert_type(%(
+      class Foo(T)
+        @x = 1
+
+        def read(other)
+          other.@x
+        end
+      end
+
+      Foo(Int32).new.read(Foo(Bool).new)
+      )) { int32 }
+  end
+
+  it "can't read from unrelated type in same namespace" do
+    assert_error %(
+      module Moo
+        class Foo
+          @x = 1
+
+          def foo(other)
+            other.@x
+          end
+        end
+
+        class Bar
+          @x = 2
+        end
+      end
+
+      Moo::Foo.new.foo(Moo::Bar.new)
+      ),
+      "can't access Moo::Bar.@x from unrelated type Moo::Foo"
+  end
+
+  it "can't read from unrelated type" do
+    assert_error %(
+      class Foo
+        @x = 1
+
+        def foo(other)
+          other.@x
+        end
+      end
+
+      class Bar
+        @x = 2
+      end
+
+      Foo.new.foo(Bar.new)
+      ),
+      "can't access Bar.@x from unrelated type Foo"
+  end
+
+  it "can't read from top-level" do
+    assert_error %(
+      class Foo
+        @x = 1
+      end
+
+      Foo.new.@x
+      ),
+      "can't access Foo.@x from the top-level"
+  end
+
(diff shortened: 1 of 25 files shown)


wontruefree:
Is this technically a breaking change?

wontruefree:
Would this prevent access from a sample like this?

```crystal
require "http/client"

class SampleConnection
  def self.collect_metrics
    client = HTTP::Client.new "www.example.com"
    pp! client.@socket
  end
end

SampleConnection.collect_metrics
```

asterite:
Yes, of course.

wontruefree:
Should there be a test case for not in the top level?

```crystal
it "can read from any non top-level" do
  assert_error %(
    class Foo
      @x = 1
    end

    class Bar
      def baz
        Foo.new.@x
      end
    end

    Bar.new.baz
    ),
    "can't access Foo.@x from the top-level"
end
```

Also is there a reason why top level is excluded?

asterite:
There's a test for that. And your test is accessing the instance var from Bar, not the top level.

wontruefree:
I just wanted to check.  It seems a lot of the tests are passing in params.  It is good to know that will be there.

spec/compiler/semantic/read_instance_var_spec.cr
```diff
@@ -0,0 +1,148 @@
...
+      class Bar
+        @x = 2
+      end
+
+      Foo.new.foo(Bar.new)
+      ),
+      "can't access Bar.@x from unrelated type Foo"
+  end
+
+  it "can't read from top-level" do
```
wontruefree: Is there a reason why the top-level is excluded from accessing ivars?

asterite: It's the main thing this PR changes.

wontruefree: I am sorry I missed that from the issue.  I didnt realize it was restricted like that.

straight-shoota:
I was under the impression that the tenor in #6066 was not to change accessing instance variables.

ysbaddaden:
1. stdlib is broken in some places, for example:
```
in src/char.cr:773: can't access String::Builder.@encoding from unrelated type Char
      if io.@encoding
```
2. why was this merged when discussion in #6066 was leaning towards "keep current behavior" ?!

Let's at least fix all usages in stdlib. I can't run std specs with crystal master now. But I vote to revert.

wontruefree:
Following up from earlier.  I am curious why was this merged?

bcardiff:
It doesn’t follow from discussion that this should have been merged. Let’s revert this PR before the release :-)

RX14:
Better to try it and see what it breaks in my opinion. We can always revert it in a point release if people *actually* want this reverted.

This PR makes the code cleaner at least, so please avoid doing a flat-out `git revert` and reverting the cleanups too.

straight-shoota:
@RX14 Breaking it to see if users' code breaks while we actually don't want it to break (and break again in a point release) doesn't seem like a good idea at all.

RX14:
@straight-shoota we change things to see if they break all the time, look at #6024, which was reverted.

Let's do some science.

straight-shoota:
Yeah, but that change was intended. It would have stayed if it didn't break.

This one is, as far as I am concerned, *not* intended. So why bother breaking if we don't want it in the end anyways?

PR #6360, opened 2018-07-10
Change String#includes? for multiple substrings

delef:
Example:
`"Bar".includes?('a', 'r') # => true` or `"Crystal".includes?("Cry", "al") # => true`
but
`"Crystal".includes?("Cry", "ol") # => false`

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -840,12 +840,17 @@ describe "String" do
     describe "by char" do
       it { "foo".includes?('o').should be_true }
       it { "foo".includes?('g').should be_false }
+      it { "foo".includes?('f', 'o').should be_true }
+      it { "foo".includes?('f', 'g').should be_false }
     end
 
     describe "by string" do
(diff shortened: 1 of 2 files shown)


spec/std/string_spec.cr
```diff
@@ -1020,10 +1025,10 @@ describe "String" do
   end
 
   it "reverses taking grapheme clusters into account" do
-    reversed = "noël".reverse
```
straight-shoota: Why is there a change in these lines?

straight-shoota:
I don't see a value in this overload. `string.includes?(a, b)` is by all means equal to `{a, b}.all? { |part| string.includes?(part) }`, which in contrast expresses clearly what's happening. Reading `includes?(a, b)` makes me wonder if it's using a logical conjunction or disjunction. My first guess would have been the latter.

Unless it was implemented with improved performance, I don't think a multiple string overload makes any sense.

delef:
@straight-shoota ok 🙂

PR #6362, opened 2018-07-10
Remove `ditto` and `nodoc` without colons

j8r:
We are used to choose `:nodoc:` with colons but `ditto` without.
However the other variant is also supported, `nodoc` and `:ditto:` could also be used.
Using colons shows that they are special semantics and not another comments.

Since `# :nodoc:` is always used vs. `# nodoc`, the diffs are mainly about changing `# ditto` to `# :ditto:`.

### DIFF
--- a/src/big/big_int.cr
+++ b/src/big/big_int.cr
@@ -53,7 +53,7 @@ struct BigInt < Int
     end
   end
 
-  # ditto
+  # :ditto:
   def initialize(num : Int::Unsigned)
     if num <= LibC::ULong::MAX
       LibGMP.init_set_ui(out @mpz, num)
@@ -62,22 +62,22 @@ struct BigInt < Int
     end
   end
 
-  # ditto
+  # :ditto:
   def initialize(num : Float::Primitive)
     LibGMP.init_set_d(out @mpz, num)
   end
 
-  # ditto
+  # :ditto:
   def self.new(num : BigFloat)
     num.to_big_i
   end
 
-  # ditto
+  # :ditto:
   def self.new(num : BigDecimal)
     num.to_big_i
   end
 
-  # ditto
+  # :ditto:
   def self.new(num : BigRational)
     num.to_big_i
   end
@@ -397,7 +397,7 @@ struct BigInt < Int
     String.new(to_cstr)
   end
 
-  # ditto
+  # :ditto:
   def to_s(io : IO) : Nil
     str = to_cstr
     io.write_utf8 Slice.new(str, LibC.strlen(str))
--- a/src/bit_array.cr
+++ b/src/bit_array.cr
@@ -224,7 +224,7 @@ struct BitArray
     io << ']'
   end
 
-  # ditto
+  # :ditto:
   def inspect(io : IO) : Nil
     to_s(io)
   end
--- a/src/compiler/crystal/semantic/top_level_visitor.cr
+++ b/src/compiler/crystal/semantic/top_level_visitor.cr
@@ -1056,13 +1056,18 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
     end
   end
 
-  def check_ditto(node)
+  def check_ditto(node : Def | Assign | FunDef | Const) : Nil
+    return if !@program.wants_doc?
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    if stripped_doc == ":ditto:"
       node.doc = @last_doc
+    elsif stripped_doc == "ditto"
+      # TODO: remove after 0.33.0
+      @program.warning_failures << "`ditto` is no longer supported. Use `:ditto:` instead"
+      node.doc = @last_doc
+    else
+      @last_doc = node.doc
     end
-
-    @last_doc = node.doc
   end
 
   def annotations_doc(annotations)
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -211,9 +211,14 @@ class Crystal::Doc::Generator
     toplevel_items.any? { |item| must_include? item }
   end
 
-  def nodoc?(str : String?)
-    return false unless str
-    str.starts_with?(":nodoc:") || str.starts_with?("nodoc")
+  def nodoc?(str : String?) : Bool
+    return false if !str || !@program.wants_doc?
+    # TODO: remove after 0.33.0
+    if str.starts_with?("nodoc")
+      @program.warning_failures << "`nodoc` is no longer supported. Use `:nodoc:` instead"
+      return true
+    end
+    str.starts_with?(":nodoc:")
   end
 
   def nodoc?(obj)
--- a/spec/compiler/semantic/doc_spec.cr
+++ b/spec/compiler/semantic/doc_spec.cr
@@ -82,7 +82,7 @@ describe "Semantic: doc" do
         def bar
         end
 
-        # ditto
+        # :ditto:
         def bar2
         end
       end
--- a/src/array.cr
+++ b/src/array.cr
@@ -666,7 +666,7 @@ class Array(T)
     self
   end
 
-  # ditto
+  # :ditto:
   def concat(other : Enumerable)
     left_before_resize = @capacity - @size
     len = @size
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -61,7 +61,7 @@ module Crystal::Macros
   def `(command) : MacroId
   end
 
-  # ditto
+  # :ditto:
   def system(command) : MacroId
   end
 
@@ -276,19 +276,19 @@ module Crystal::Macros
     def <(other : NumberLiteral) : BoolLiteral
     end
 
-    # ditto
+    # :ditto:
     def <=(other : NumberLiteral) : BoolLiteral
     end
 
-    # ditto
+    # :ditto:
     def >(other : NumberLiteral) : BoolLiteral
     end
 
-    # ditto
+    # :ditto:
     def >=(other : NumberLiteral) : BoolLiteral
     end
 
-    # ditto
+    # :ditto:
     def <=>(other : NumberLiteral) : NumberLiteral
     end
 
--- a/src/compiler/crystal/program.cr
+++ b/src/compiler/crystal/program.cr
@@ -318,7 +318,7 @@ module Crystal
       named_tuple_of(entries)
     end
 
-    # ditto
+    # :ditto:
     def named_tuple_of(entries : Array(NamedArgumentType))
       named_tuple.instantiate_named_args(entries)
     end
--- a/src/compiler/crystal/semantic/path_lookup.cr
+++ b/src/compiler/crystal/semantic/path_lookup.cr
@@ -38,7 +38,7 @@ module Crystal
       (path.global? ? program : self).lookup_path(path.names, lookup_in_namespace, include_private, location)
     end
 
-    # ditto
+    # :ditto:
     def lookup_path(names : Array(String), lookup_in_namespace = true, include_private = false, location = nil) : Type | ASTNode | Nil
       type = self
       names.each_with_index do |name, i|
--- a/src/complex.cr
+++ b/src/complex.cr
@@ -33,12 +33,12 @@ struct Complex
     @real == other.real && @imag == other.imag
   end
 
-  # ditto
+  # :ditto:
   def ==(other : Number)
     self == other.to_c
   end
 
-  # ditto
+  # :ditto:
   def ==(other)
     false
   end
@@ -235,7 +235,7 @@ struct Complex
     Complex.new(@real + other.real, @imag + other.imag)
   end
 
-  # ditto
+  # :ditto:
   def +(other : Number)
     Complex.new(@real + other, @imag)
   end
@@ -250,7 +250,7 @@ struct Complex
     Complex.new(@real - other.real, @imag - other.imag)
   end
 
-  # ditto
+  # :ditto:
   def -(other : Number)
     Complex.new(@real - other, @imag)
   end
@@ -260,7 +260,7 @@ struct Complex
     Complex.new(@real * other.real - @imag * other.imag, @real * other.imag + @imag * other.real)
   end
 
-  # ditto
+  # :ditto:
   def *(other : Number)
     Complex.new(@real * other, @imag * other)
   end
@@ -278,7 +278,7 @@ struct Complex
     end
   end
 
-  # ditto
+  # :ditto:
   def /(other : Number)
     Complex.new(@real / other, @imag / other)
   end
--- a/src/csv/builder.cr
+++ b/src/csv/builder.cr
@@ -68,7 +68,7 @@ class CSV::Builder
     end
   end
 
-  # ditto
+  # :ditto:
   def row(*values)
     row values
   end
@@ -119,7 +119,7 @@ class CSV::Builder
       end
     end
 
-    # ditto
+    # :ditto:
     def <<(value : Nil | Bool | Number)
       case @quoting
       when .all?
@@ -133,7 +133,7 @@ class CSV::Builder
       end
     end
 
-    # ditto
+    # :ditto:
     def <<(value)
       self << value.to_s
     end
@@ -145,7 +145,7 @@ class CSV::Builder
       end
     end
 
-    # ditto
+    # :ditto:
     def concat(*values)
       concat values
     end
--- a/src/dir/glob.cr
+++ b/src/dir/glob.cr
@@ -8,7 +8,7 @@ class Dir
     glob(patterns)
   end
 
-  # ditto
+  # :ditto:
   def self.[](patterns : Enumerable(String)) : Array(String)
     glob(patterns)
   end
@@ -24,7 +24,7 @@ class Dir
     glob(patterns, match_hidden: match_hidden)
   end
 
-  # ditto
+  # :ditto:
   def self.glob(patterns : Enumerable(String), match_hidden = false) : Array(String)
     paths = [] of String
     glob(patterns, match_hidden: match_hidden) do |path|
@@ -46,7 +46,7 @@ class Dir
     end
   end
 
-  # ditto
+  # :ditto:
   def self.glob(patterns : Enumerable(String), match_hidden = false, &block : String -> _)
     Globber.glob(patterns, match_hidden: match_hidden) do |path|
       yield path
--- a/src/humanize.cr
+++ b/src/humanize.cr
@@ -53,7 +53,7 @@ struct Number
     end
   end
 
-  # ditto
+  # :ditto:
   def format(separator = '.', delimiter = ',', decimal_places : Int? = nil, *, group : Int = 3, only_significant : Bool = false) : String
     String.build do |io|
       format(io, separator, delimiter, decimal_places, group: group, only_significant: only_significant)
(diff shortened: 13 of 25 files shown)


RX14:
My main concern is this breaking for people silently. Probably the best solution would be to compile-error "nodoc" or "ditto" comments to get people to change.

This wasn't broken before though, so I'm not sure we should really "fix" it.

j8r:
There is likely not much people outside crystal-lang contributors using this undocumented feature @RX14 .

Besides that, in the worst case, `nodoc` and `ditto` will be shown as-is in the method description on the HTML doc – not a big deal.

straight-shoota:
If the behaviour of the compiler is changed, it should raise. I'm not sure this should to be changed at all (at least not right away).

I like changing from `ditto` to `:ditto:` in the stdlib docs, but it doesn't need to be enforced.

RX14:
@j8r "ditto" is documented, and plenty of people use it.

straight-shoota:
The reference should better use `:ditto:` though. But still, it doesn't hurt to support `ditto` as well.

j8r:
I was searching in the code @RX14 - doesn't know it was here, my bad. Itmight also be in other places.

OK, for now we can print a warn about the deprecation of `ditto` and `nodoc`.

j8r:
What to do to merge - do I need to add a deprecated notice, what's the standard way?

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -163,7 +163,10 @@ class Crystal::Doc::Generator
   end
 
   def nodoc?(str : String?)
-    str == ":nodoc:" || str == "nodoc"
+    # TODO: keep only `:nodoc:` after 0.27.0                                                  
+    # TODO: replace puts with raise after 0.26.0
+    return puts "`nodoc` is deprecated, use `:nodoc:` instead`" if str == "nodoc"
```
Sija: that's wrong, should be:

```cr
if str == "nodoc"
  puts "`nodoc` is deprecated, use `:nodoc:` instead"
  return true
end
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1020,7 +1020,11 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
   def check_ditto(node)
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    # TODO: keep only `:ditto:` after 0.27.0
+    if stripped_doc == "ditto"
+      # TODO: replace puts with raise after 0.26.0
+      puts "`ditto` is deprecated, use `:ditto:` instead`"
```
Sija: that's wrong, see below.

paulcsmith:
I think that just like in other parts of Crystal, there should be one way to do things unless there is a benefit to having 2. Just like Crystal has fewer aliases than Ruby, I think making things consistent and making it only work with colons makes it easier to search for, easier to tell that it is "special", and simpler for codebases to have a common way of documenting things, since only one will work.

I like the idea of raising if the comment line contains just the word `ditto` or just the word `nodoc` without colons so that people know there was a change and can easily make that change (a find-and-replace would be really simple `# nodoc` -> `# :nodoc:`. Piece of cake)

straight-shoota:
Sounds good.

As an additional enhancement, I'd like that that a comment just has to start with `:nodoc:` but may contain other content. This allows to actually have documentation (in code) without being included in the public API.

src/compiler/crystal/semantic/top_level_visitor.cr
j8r: (fixed now)

j8r:
Rebased to include the change of https://github.com/crystal-lang/crystal/pull/6627 by @Sija, which addresses the @straight-shoota's comment above.

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -164,6 +164,12 @@ class Crystal::Doc::Generator
 
   def nodoc?(str : String?)
     return false unless str
+    # TODO: keep only `:nodoc:` after 0.28.0
+    # TODO: replace puts with raise after 0.27.0
+    if str == "nodoc"
```
Sija: `str.starts_with?("nodoc")`

j8r: Humm... I was sure to take it during the rebase

j8r: Go it

j8r:
What's remaining for merge, all's good?

j8r:
Can this PR be milestoned? This will avoid me to change the `TODO: after x.x.x` and rebase for each new release.

straight-shoota:
I'd prefer if this were two commits, one changing `ditto` to `:ditto:`, the other changing compiler behaviour. The first one could be merged right away. Maybe that would be a good idea to split the compiler change into a different PR. It feels like that part might need more discussion.

I think we all agree that using the variants with colons is the preferred way. Judging from the approving reactions, it seems also settled that support for non-colon variants should vanish.

It is however unclear if this should be silent, only a warning or a compiler error. And when it should be applied.

I don't like the current implementation, which just prints a message to STDOUT. Warnings tend to be ignored anyway. IIRC it was decided once that the compiler should either work or fail, but don't issue warnings.

I'd be fine with raising or failing silently. Raising for this seems a bit harsh, so I tend to prefer silently ignoring it. The costs of an error are pretty low: You'd have supposedly undocumented features show up in the API docs with a description `nodoc`, or some features documented as `ditto`. Both meanings can be expanded by the reader and are relatively easy to spot by maintainers. That's nothing worth raising for.

Maybe the docs generator could issue a warning and/or the formatter detect and adapt usages. The compiler would stop recognizing non-colon variants two releases later.

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1025,7 +1025,11 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
   def check_ditto(node)
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    # TODO: keep only `:ditto:` after 0.30.0
+    if stripped_doc == "ditto"
+      raise "`ditto` is deprecated, use `:ditto:` instead"
+      node.doc = @last_doc
```
Sija: this line is pointless when raising, ditto below.

j8r: What do you mean by failing silently? Don't handle the case at all?

ysbaddaden:
> It is however unclear if this should be silent, only a warning or a compiler error. And when it should be applied.

We make breaking changes regularly in Crystal (it's unstable) and we don't want annoying *warning messages* either, so... why bother and argue about a doc change? Let's drop support for `ditto` and `nodoc` and move on!

We'll handle deprecation properly when we aim for Crystal 1.0

straight-shoota:
@ysbaddaden It's not about deprecating. The question is: Should it be a syntax error to alert users of the breaking change or fail silently?

ysbaddaden:
Neither. It's just a comment. No errors, no silent failure. Just a changed expected behaviour (special -> comment).

RX14:
It's not "just a comment", it's documentation. And documentation is very important.

Useful documentation will silently be replaced with simply "ditto" and that's unacceptable. There must be a warning or compile error to notify people about this.

bew:
What about a warning/error/whatever only when generating documentation?
I can understand why we don't want warnings in general when compiling, or error on such unrelated-to-the-goal thing when all you want is to compile your program and run it, not having to care about your documentation.

Basically:
- When you 'compile' your program, all you care about is your code, so compilation errors should be related to the code.
- When you 'compile' your documentation, you care about the comments in your code, and we could have compilation error/warning for things related to the docs.

WDYT?

straight-shoota:
@ysbaddaden Yes, we don't show deprecation warnings but usually, when things break, we make it so that people notice that something is broken (by refusing to compile or failing specs). This breaking change would go completely unnoticed.

I like @bew's suggestion to raise only when generating docs. It only matters when generating docs, so this is a great solution.

RX14:
I also like @bew's suggestion. I think `crystal doc` should fail for *this* release on a doc comment of just "ditto" or "nodoc".

j8r:
What I'm supposed to do then?

bew:
If I understand correctly, make `crystal doc` fail when a doc is `ditto` or `nodoc` (without colons)

j8r:
Only when running `crystal doc`? How I'm supposed to do this?

straight-shoota:
There's a `check_ditto` method in `top_level_visitor` which handled the ditto stuff. I suppose this method could simply raise an error when `@program.wants_doc?` is true.

j8r:
Remains your approval @RX14

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1027,6 +1027,9 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
     stripped_doc = node.doc.try &.strip
     if stripped_doc == ":ditto:"
       node.doc = @last_doc
+      # TODO: remove after 0.29.0
```
bew: I'd suggest to put the TODO just above the `raise`, when we remove it it'll be obvious that wee don't need the `if` anymore

straight-shoota:
Needs a rebase.

asterite:
@bcardiff I think there's a setting in GitHub where you can choose to update a branch against another. A big bother appears in that case. That would free contributors from having to go back to a PR and rebase it. We could also speed things up. What do you think?

j8r:
Is it planned for 0.28.0, or needs more time for some reason? I'll rebase just before the merge.

src/big/big_int.cr
```diff
@@ -361,8 +361,8 @@ struct BigInt < Int
     String.new(to_cstr)
   end
 
-  # ditto
-  def to_s(io : IO) : Nil
+  # :ditto:
+  def to_s(io)
```
bcardiff:
```suggestion
  def to_s(io : IO) : Nil
```

src/bit_array.cr
```diff
@@ -226,8 +226,8 @@ struct BitArray
     io << ']'
   end
 
-  # ditto
-  def inspect(io : IO) : Nil
+  # :ditto:
+  def inspect(io : IO)
```
bcardiff:
```suggestion
  def inspect(io : IO) : Nil
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1025,8 +1025,11 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
   def check_ditto(node)
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    if stripped_doc == ":ditto:"
       node.doc = @last_doc
+    elsif @program.wants_doc? && stripped_doc == "ditto"
+      # TODO: remove after 0.29.0
+      raise "`ditto` is no longer supported. Use `:ditto:` instead"
```
bcardiff: I will say that maybe we can use warnings here :-)

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -198,7 +198,9 @@ class Crystal::Doc::Generator
 
   def nodoc?(str : String?)
     return false unless str
-    str.starts_with?(":nodoc:") || str.starts_with?("nodoc")
+    # TODO: remove after 0.29.0
+    raise "`nodoc` is no longer supported. Use `:nodoc:` instead" if @program.wants_doc? && str.starts_with?("nodoc")
```
bcardiff: And if we allow warnings to avoid breaking directly this should go

straight-shoota:
This doesn't need to be in 0.28.0. It would have been nice, but it's not an urgent matter. Whether we deprecate `ditto` in 0.28.0 or 0.29.0 doesn't matter. So I'd be fine with enhancing this with a warning feature but make it one of the first things for 0.29.0 :+1:

j8r:
There are no warnings features for the `crystal docs` command. Would they make sense anyway?

src/compiler/crystal/semantic/top_level_visitor.cr
j8r: How do you will do that @bcardiff ?

bcardiff: I would add in `codegen/warnings.cr` some method to emit the warning for a `ditto`.
Call that from `top_level_visitor.cr` instead of raising.

The code in `codegen/warnings.cr` has a method to ignore the emit based on location, that is the important thing to use.
The quirk is that the top_level_visitor does not need the codegen, so maybe the `SemanticVisitor` or `TopLevelVisitor` can be reopened in the `codegen/warnings.cr` file. So all the warnings methods are located near and the ignore based on location can be defined privately.

The `ignore_warning_due_to_location` uses `@program` which is available in codegen and semantic. And I guess the `@program.warnings_exclude` will be initialized.

I think is a good exercise to stress the warnings for deprecating language features. That is my main motivation to push it in this direction.

bcardiff:
@j8r I would say that the warnings can be emitted by the compiler alone, no need to make the docs emit them.

src/compiler/crystal/semantic/top_level_visitor.cr
j8r: I am not sure how to do this, even less properly :confused: 
Wouldn't it be better/simpler to have a generic way to print a warning without using annotations instead?
For example `Warning.warn "deprecated"` in Ruby or in Python:
```py
import warnings
warnings.warn("deprecated")
```

j8r: WDYT @bcardiff ?
I said this, because having to understand the compiler internals to print a warning is quite tedious on the user point a vue.

j8r:
I tried to implement basic warnings, it's hard for me to figure out how to do it properly with my limited knowledge of the compiler. Please be free to commit @bcardiff

j8r:
Anything remains to do? I'm not good with compiler stuff. The warnings are emitted, but without any location.

straight-shoota:
Could you debase on master?

straight-shoota:
@j8r I feel this PR has been going on for way too long. I'd like to merge this but only the change from `ditto/nodoc` to `:ditto:/:nodoc:` (did you make sure all current occurrences are replaced?). Could you remove the logic changes to `TopLevelVisitor` and `Doc::Generator` from this PR? I think it's best to tackle that in a follow up. And somebody else can take over.
We probably should have separated this long ago.

asterite:
Could we go the other way around? Use `ditto`, `nodoc`, `inherit`, etc? We copied `:nodoc:` from Ruby, but I don't think using that format is good, or has an equivalent in other places of the language. Using less symbols/noise is always better. And I can't imagine someone wanting to document their code with `nodoc`, `inherit` or `ditto` in single lines of their own without that being a directive of some sort.

j8r:
@asterite but then, how do we know this comments are a special semantic used in the doc generator?

asterite:
@j8r we document them

j8r:
Using something uncommon like colons directly show us that they are not regular comments.
Other symbols than colons could be used also.

bcardiff:
I like more the colon versions. But it's true that unless we validate/parse `:<word>:` are within the known doc special keywords both are equally (un)safe. The colon version has the change to implement that validation in the future at least.

straight-shoota:
@asterite I disagree. Using a special syntax highlights that this is a special value with special behaviour. It differentiates these values from regular text. When a method is documented by a single word, it might be confusing whether that is just a description or has some special meaning (is plain `super` a keyword, or `previous`?). When we use a specific format (such as wrapping a word in colons) we could even make any similar value that is not a defined keyword a failure.
`:inherit:` doesn't even need to be placed at the start of a doc comment, it can be surrounded by paragraphs (see #6989) and currently doesn't support the non-colon variant.

Omitting the colons might not break anything, but IMO it doesn't improve either. It might be a matter of habit, but I like the `:nodoc:` notation. I'm pretty sure we had previous discussions about this and always preferred the colon variant.

We could possibly consider using a unique format for specifying doc options (see #8353), for example a syntax inspired by JavaDoc tags could work for this. But I think I'd still prefer colon delimited keywords for `nodoc`, `ditto` and `inherit` because they don't define metadata but directly alter the documentation text.

j8r:
It may be a good idea @straight-shoota .
There are `TODO:`, `NOTE:`, we may also have `DITTO` and `NODOC`.
There is less noise, as @asterite pointed out, and there are still in highlighted from regular comments.

asterite:
Just for reference (I'm not saying it's a good or bad idea), I copied `ditto` from D: https://dlang.org/spec/ddoc.html#parsing

straight-shoota:
@asterite `ditto` alone might have been fine to use only the blank word because it is typically the only content of a doc comment. But `nodoc` can be followed by internal documentation. And finally `inherit` which can be somewhere inside the doc text. They definitely call for a special syntax to make them stand out.

bcardiff:
Let's keep using and encouraging `:nodoc:`, `:ditto:` and `:inherit:`. And eventually stop treating non-colon versions with a deprecation warning.

j8r:
I just keep what I've done with `sed -i -e 's/# ditto/# :ditto:/g' -e 's/# nodoc/# :nodoc:/g' $(find . -name "*.cr")` - replacing all `ditto`/`nodoc` with `:ditto:`/`:nodoc:`.

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -199,7 +199,12 @@ class Crystal::Doc::Generator
 
   def nodoc?(str : String?)
     return false unless str
-    str.starts_with?(":nodoc:") || str.starts_with?("nodoc")
+    # TODO: remove after 0.33.0
+    if @program.wants_doc? && str.starts_with?("nodoc")
+      @program.warning_failures << "`nodoc` is no longer supported. Use `:nodoc:` instead"
+      return
```
bcardiff:
```suggestion
      return true
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1054,8 +1054,12 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
   def check_ditto(node)
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    if stripped_doc == ":ditto:"
       node.doc = @last_doc
+    elsif @program.wants_doc? && stripped_doc == "ditto"
+      # TODO: remove after 0.33.0
+      @program.warning_failures << "`ditto` is no longer supported. Use `:ditto:` instead"
+      return
```
bcardiff:
```suggestion
      node.doc = @last_doc
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1054,7 +1054,11 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
   def check_ditto(node)
     stripped_doc = node.doc.try &.strip
-    if stripped_doc == ":ditto:" || stripped_doc == "ditto"
+    if stripped_doc == ":ditto:"
+      node.doc = @last_doc
```
straight-shoota: Both branches assign `node.doc = @last_doc` and afterwards there's `@last_doc = node.doc`. This looks like it could be simplified. The last assignment only has an effect when neither condition matches.

And, more generally we don't even care about docs when `@program.wants_doc?` is `false`. So that could be an early return for this method.

j8r:
@RX14, remains your approval.

straight-shoota:
Thanks @j8r

bcardiff:
FYI. I was checking how this was working and it neither outputs the warnings, nor the warning are build with location information to tell the user were to fix the `nodoc`.

PR #6367, opened 2018-07-11
Syntax: allow empty `case` without `when`

### LINKED ISSUE #6318: Allow empty `case` without `when`
An empty `case` without `when` is currently a syntax error:

```cr
case "foo"
end # unexpected token: end (expecting when or else)

case "foo"
else # unexpected token: else (expecting when)
end
```
(the first syntax error even says `expecting when or else`, which would mean the second example should work. But it actually expects a `when`)

In regular code it certainly doesn't make much sense and is about as meaningful as `if true`. But it also doesn't do any harm to have a `case` without any `when` branch.

On the other hand, it comes in handy when the code for `when` branches is automatically generated (for example in a macro):

```cr
case foo
{% for case in cases %}
when {{ case.stringify }}
  # ...
{% end %}
end
```
If `cases` is empty, this macro will produce invalid code. It would require extra effort to handle this - and  it's ugly and hard to read. If more logic is involved, it might not be as simple as wrapping the entire `case` in `{% unless cases.empty? %}`.

straight-shoota:
Fixes #6318

The following is no longer a syntax error:
```cr
case foo # expands to `foo`
end

case foo # expands to `foo; bar`
else
  bar
end
```

While an empty case doesn't make much sense in regular code, it is useful for automatically generated `when` branches (for example in a macro) and avoids extra effort to prevent empty case nodes.

This also fixes a parsing error for `case` without condition (`case;
when true; end`) and adds specs for this.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -494,6 +494,12 @@ describe Crystal::Formatter do
   assert_format "case 1\nwhen 1, # 1\n     2, # 2\n     3  # 3\n  1\nend"
   assert_format "a = case 1\n    when 1, # 1\n         2, # 2\n         3  # 3\n      1\n    end"
   assert_format "a = 1\ncase\nwhen 2\nelse\n  a /= 3\nend"
+  assert_format "case 1\nend"
+  assert_format "case 1\nelse\n  2\nend"
+  assert_format "case\nend"
+  assert_format "case 1\nend"
+  assert_format "case\nend"
+  assert_format "case\nelse\n  1\nend"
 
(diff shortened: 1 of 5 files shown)


r00ster91:
But there are still 2 error messages left that say "expecting when, else or end".
Heres the first one:
https://github.com/straight-shoota/crystal/blob/e14ec7a4af4453fca0ff9d0bcc0004864d01b910/src/compiler/crystal/syntax/parser.cr#L2544
They should be changed to "expecting end" since empty cases are allowed now.

straight-shoota:
@rooster After a `case` (or `when`) the parser expects either of `when, else, end`. This is completely fine and was actually incorrect before this commit, because after `case` it required a `when`.

PR #6258, opened 2018-06-25
fix time parse doc

icyleaf:
Fixed document but i still ask why not use `Time::Location.local` by default if time string was not contain timezone?

Issue #5332, opened 2017-11-28
[RFC] Supporting floating time


straight-shoota:
In the context of my proposal for adding time zone support (#5324) the question arises how Crystal should handle time instances which are not tied to a specific time zone, known as *floating time*.

## Conceptual Introduction

A short introduction to the concept from [W3C's time zone guide](https://www.w3.org/TR/timezone/#floating):

> Some observed time values are not related to a specific moment in incremental time. Instead, they need to be combined with local information to determine a range of acceptable incremental time values. We refer to these sorts of time values as "floating times" because they are not fixed to a specific incremental time value. Floating times are not attached and should never be attached to a particular time zone.
> Some examples of floating time events include a user’s birth date, a document's publication date, a list of official holidays, or the expiration date for an offer (if not tied explicitly to a time zone).

A floating time `08:30` can for example represent each point in time, where the local time shows `08:30`, independent of local time's timezone. If I set an alarm at `08:30`, I want that alarm to trigger when it is `8:30` wherever I am at that point and whatever time zone offset applies there.

## Usage in computer systems

Such floating times are a ubiquitous, yet often unaware way of how people interact in temporal contexts.
A computer system needs to handle them appropriately. This concept is present in many standard date and calendar formats, data serialization formats and databases. To my knowledge, most modern time libraries also have some support for floating time. 

It's easy for a Crystal application to encounter a floating time reference. What should happen if there is no representation for this? Instantly converting a floating time to incremental and assuming it to be in the local time zone would defeat the purpose of using floating time alltogether (W3C guide: *Floating times are not attached and should never be attached to a particular time zone.*).
Handling floating time wrongly is a serious issue in many pieces of software ([example on Martin Fowler's blog](https://martinfowler.com/bliki/TimeZoneUncertainty.html) and should be avoided by providing the right tools to work with.

## Integration in Crystal

A huge portion of software applications need to work with incremental time. On the other hand, floating time is certainly not as frequently needed, it can be considered as being more specific to applications tailored towards calendaring purposes and interacting with human representations of time. So, there is not as much an important reason to have support for floating time available in the standard library. It could be implemented by a shard which can be used only by applications that need this feature.

[...]

PR #5324, opened 2017-11-24
Add time zones support

straight-shoota:
Until now, instances of `Time` had a `kind` member which indicated, whether this instance holds a time in UTC, an unspecified "local" timezone or is entirely unspecified.

This PR adds a `Time::Location` class which represents a specific time zone. Each `Time` instance references a `Location` (i.e where this time applies). From this location's transition rules, the applicable offset at each point in time can be derived, including daylight savings time, time zone changes etc.

Storing only an offset from UTC is not enough, because from that you can't infer what timezone transition rules to apply. And offset changes inside a time zone happen quite frequently (for example daylight savings).

# Time
The `kind` ivar in `Time` is replaced by `location` which is a pointer to a `Location` instance.

```crystal
struct Time
  @seconds : Int64                    # 8 Byte
  @nanoseconds : Int32                # 4 Byte
  @kind : Time::Kind                  # 4 Byte
end

# Becomes:
struct Time
  @seconds : Int64                    # 8 Byte
  @nanoseconds : Int32                # 4 Byte
  @location : Pointer(Time::Location) # 8 Byte (64 bit architecture)
end
```

This means, each `Time` struct will require 24 bytes instead of the currently used 16 bytes (on 64-bit). This increases the size of a time struct slightly but it is inevitable to embed a reference to a time zone if proper time zone handling should be supported.

For incremental time (a reference to a a specific point in time), `seconds` and `nanoseconds` always represent time in UTC. The offset will only be added when needed (for example to access one of the date/time fields individually). This makes it easier than if it was stored *with* offset as that might lead to invalid or ambiguous instances (for example when an hour is skipped or happens twice at a DST switch). Another benefit is, that retrieving the offset (zone transitions are looked up by unix epoch) and comparing two instances requires no conversion. Converting a time between zones means just exchanging the `location`.

# Time::Location

Each location can be one of the following
* a timezone in the IANA time zone database read from a local copy of this database which is provided for by the operating system (available in most unix-like OS's) or a custom distribution (can be configured throught `ZONEINFO` environment variable)
* A zone with a fixed-offset
* Coordinated Universal Time, available as `Time::Location::UTC`
* an unspecified time zone, indicating a floating time which is not tied to a specific time zone.

The concept for loading location data from system sources is inspired by [Go's implementation](https://golang.org/pkg/time/#Location) which seems very reasonable for this.

[...]

PR #6369, opened 2018-07-12
Add location as required argument for Time.parse

straight-shoota:
This makes it more clear in the API that a default value is required instead of defaulting to `nil` which would raise in case of time zone information missing from the formatted string. `Time.parse_utc` and `Time.parse_local` are added as short cuts for the most common use cases.

See https://github.com/crystal-lang/crystal/pull/6258#issuecomment-403353910

### DIFF
--- a/spec/std/time/time_spec.cr
+++ b/spec/std/time/time_spec.cr
@@ -540,10 +540,10 @@ describe Time do
 
   it "parse fails without time zone" do
     expect_raises(Time::Format::Error, "no default location provided") do
-      Time.parse("2017-12-01 20:15:13", "%F %T", nil)
+      Time.parse!("2017-12-01 20:15:13", "%F %T")
     end
     Time.parse("2017-12-01 20:15:13", "%F %T", Time::Location.local).to_s("%F %T").should eq "2017-12-01 20:15:13"
-    Time.parse("2017-12-01 20:15:13 +01:00", "%F %T %:z", nil).to_s("%F %T %:z").should eq "2017-12-01 20:15:13 +01:00"
+    Time.parse!("2017-12-01 20:15:13 +01:00", "%F %T %:z").to_s("%F %T %:z").should eq "2017-12-01 20:15:13 +01:00"
   end
 
   it "parses" do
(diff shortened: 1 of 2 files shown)


straight-shoota:
You suggestion is essentially the same as we have currently: 
```cr
def parse(time : String, pattern : String, location : Location? = nil)
```

The issue with this solution is that the shortest overload `parse(String, String)` is the least safest. Requiring an explicit argument (or different name) makes this less dangerous. `nil` as an argument isn't very explicit about what it does, though. But it fits quite nice logically (if `location` is `nil`, you can't create a `Time` instance with it). However, it can easily lead to unintended beaviour if the value of `location` comes from a nilable variable.

So, maybe even more explicitness would be nice. This could either be a required argument (`parse_with_location`), a different name (`parse_with_location`, or maybe `parse!`??) or even a different type (for example `location: :raise` instead of `location: nil`).

jhass:
> You second suggestion is essentially the same as we have currently

Hence my first suggestion first and my personal opinion second. As I wrote in the second suggestion the separate overloads would be for nicer docs and code completion/IDE integration. I do see your point though and hence my compromise proposal stands first. Though I see now it doesn't make sense since there's no sensible behavior for passing false.

How terrible would it be to have a fake location like `Location::Unknown` or `Location::Raise`?

asterite:
Alternatively, make the current parse assume a location. Go for example assumes UTC.

But I'm not sure. In my workplace, for the current project, we have a tool where you specify a range of dates in the command line. It's a C# application. C# also seems to assume a location, and now I'm curious which one is it. This program sends the date to a server, and should probably be in UTC, or the server's time. So forcing the user to think a bit about whether they want UTC or local by default might be a good idea, even if it's a bit annoying.

But also, maybe the current behavior isn't bad, like jhass says, because you'll find out soon if there's something wrong (unless it's user input, which might not be a common case).

I can't make up my mind about this topic :-P

straight-shoota:
If we had `Location::Unknown`, I'd expect `parse` not to raise but return a `Time` with that location. That's possible but delays the problems to later usage of this instance. It would essentially be a floating time and that's what we had with `Kind::Unspecified` prior to `Time::Location` (see https://github.com/crystal-lang/crystal/pull/5324#issuecomment-348572486, #5332). It could still raise (even more so with `Location::Raise`) but it would dirty the API with unusable `Location` instances conveying special meaning. Using `nil` makes even more sense to me than that.

straight-shoota:
I tend to like using a different method name altogether. It clearly communicates a difference in behaviour and doesn't do so over special argument types or values. `parse!` would be great for this, the exclamation mark signals for more special attention compared to `parse`.

straight-shoota:
For reference, I just noticed a similar use case in [`String#encode`](https://crystal-lang.org/api/0.25.1/String.html#encode%28encoding%3AString%2Cinvalid%3ASymbol%3F%3Dnil%29%3ABytes-instance-method). The method signature is `encode(encoding : String, invalid : Symbol? = nil)`. If `invalid` is `nil`, it raises on an invalid multibyte sequence. If `:skip`, it will be skipped. That's essentially similar to the current `Time.parse` implementation.

straight-shoota:
I added a commit that changes `parse` to always require a location and `parse!` as a variant without location argument that eventually raises if there is no time zone information in the formatted string.

PR #6371, opened 2018-07-12
Add ECR.render for rendering without setting up an IO

straight-shoota:
This macro reduces boilerplate IO pattern if you just need to render a ECR file in a string to make it more user-friendly.

There is something to be said about using the same name as `ECR.embed` just without the argument `ioname`. However, I feel like `ECR.render` better describes what it does. I'm not settled, though, so any suggestions are welcome.

See kemalcr/kemal#467


sdogruyol:
This is definitely useful :+1:

Issue #271, opened 2014-11-30
Stack overflows


jhass:
I'm not sure if that's possible, but I think it would be great if we could turn stack overflows into exceptions. Currently they just cause a segfault.

To be clear what I'm talking about, a simple trigger:

``` ruby
def foo
  foo
end
foo
```

Valgrind hints:

```
==27879== Stack overflow in thread 1: can't grow stack to 0xffe801ff8
==27879== 
==27879== Process terminating with default action of signal 11 (SIGSEGV)
==27879==  Access not within mapped region at address 0xFFE801FF8
==27879==    at 0x4017D1: *foo:Nil (in /home/jhass/projects/crystal/tmp3)
==27879==  If you believe this happened as a result of a stack
==27879==  overflow in your program's main thread (unlikely but
==27879==  possible), you can try to increase the size of the
==27879==  main thread stack using the --main-stacksize= flag.
==27879==  The main thread stack size used in this run was 8388608.
```

waterlink:
Can't it be solved in a "soft" way? ie: by limiting the depth of method calls. Will probably cause performance issues, though..

waterlink:
Or I would look into this one: http://llvm.org/docs/doxygen/html/StackProtector_8cpp_source.html

waterlink:
The other "soft"/"hard" method to do that is:

- At entrypoint of program memorize somewhere current tip of stack
- At start of each function compare current tip of stack with memorized head of stack
- Jump to handling code when the difference is more than certain threshold (95% of stack size or something)

Handling code is something that is responsible for raising an exception.

This will in theory add about 5-6 asm instructions to each function call.

kirbyfan64:
My two cents:

Ignore the stack overflow. Use @waj's suggestion to trap SIGSEGV, then figure out how to recover the line number info and print it. This is what Nim does.

A stack trace would be nice, but it seems impossible to implement portably.

radarek:
+1 for printing something more usefull than "Segmentation fault: 11" :).

Rubinius prints both C++ and Ruby backtrace on segfault (here is example of this behavior https://gist.github.com/benweint/44b15ff689fd694b83a8).
[...]

PR #6376, opened 2018-07-12
Proof of Concept: detect stack overflows (x86_64-linux-gnu target only)

ysbaddaden:
I noticed that memory accesses that segfaults are near the bottom of stack, that is within the stack guard allocated by the kernel —I initially expected the guard to be *after* the stack, but it's actually within it. Checking if `addr` in the segfault handler (the faulting memory address) is within the stack of the current fiber stack is enough to detect a stack overflow.

Since the main fiber doesn't have a `@stack`, I use `getrlimit(2)` to get the default stack size and determine the main stack limit. I only added bindings for x86_64-linux-gnu, but they should be the same for all Linux libc.

I don't know how other POSIX systems (Darwin, OpenBSD, FreeBSD) handle their stack guard.

NOTE: there could a `Crystal::System.within_current_stack?(addr)` or `stack_overflow?` method.

refs #271

### DIFF
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -1,5 +1,6 @@
 require "c/signal"
 require "c/stdio"
+require "c/sys/resource"
 require "c/sys/wait"
 require "c/unistd"
 
@@ -269,10 +270,26 @@ end
 
 # :nodoc:
 fun __crystal_sigfault_handler(sig : LibC::Int, addr : Void*)
+  # Capture fault signals (SEGV, BUS) and finish the process printing a backtrace first
   Crystal.restore_blocking_state
 
-  # Capture fault signals (SEGV, BUS) and finish the process printing a backtrace first
-  LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+  stack = Fiber.current.@stack
+
+  unless stack
+    # main fiber (main thread stack):
+    stack_size = 0
+    if LibC.getrlimit(LibC::RLIMIT_STACK, out rlim) == 0
+      stack_size = rlim.rlim_cur
+    end
+    stack = Pointer(Void).new(Fiber.current.@stack_bottom.address - stack_size)
+  end
+
+  if stack <= addr < Fiber.current.@stack_top
+    LibC.dprintf 2, "Stack overflow: infinite or too deep recursion (invalid memory access at address 0x%lx)\n", addr, stack
+  else
+    LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+  end
+
   CallStack.print_backtrace
   LibC._exit(sig)
 end
--- a/src/lib_c/x86_64-linux-gnu/c/sys/resource.cr
+++ b/src/lib_c/x86_64-linux-gnu/c/sys/resource.cr
@@ -1,0 +1,12 @@
+lib LibC
+  alias RlimT = ULong
+
+  struct Rlimit
+    rlim_cur : RlimT
+    rlim_max : RlimT
+  end
+
+  fun getrlimit(Int, Rlimit*) : Int
+
+  RLIMIT_STACK = 3
+end


r00ster91:
Or why not checking if the signal is `11`? When I do this:
```crystal
def method
  method
end
method
```
then I always get signal 11.

jhass:
@r00ster91 because you also get it if you do `Pointer(String).new(0).value` https://carc.in/#/r/4i4r

straight-shoota:
@r00ster91 SIGSEGV (signal 11) can have a number of reasons, one of them can be a stack overflow. More details at https://en.wikipedia.org/wiki/Segmentation_fault

src/signal.cr
```diff
@@ -269,10 +270,26 @@ end
...
+  unless stack
+    # main fiber (main thread stack):
+    stack_size = 0
+    if LibC.getrlimit(LibC::RLIMIT_STACK, out rlim) == 0
+      stack_size = rlim.rlim_cur
+    end
+    stack = Pointer(Void).new(Fiber.current.@stack_bottom.address - stack_size)
+  end
+
+  if stack <= addr < Fiber.current.@stack_top
```
RX14: If we've come from a segfault, wouldn't `@stack_top` be out of date? What about detecting if the fault is anywhere inside `@stack_bottom - stack_size <= addr  <= @stack_bottom`?

ysbaddaden: Yup, that's a typo: it should be `stack_bottom` instead of `stack_top`. It still works thought, since `stack_top` was a valid stack pointer, but `stack_bottom` would be better, yeah.

Yet we must check with `stack` which is the far limit of the stack:

                           <- stack grows down
    +-------x--+----------+---------------------------------+
    ^       ^  ^          ^                                 ^
    |       |  |          |                                 |
    + stack |  + guard    + top                             + bottom
            |
            + invalid memory access: stack overflow

RX14:
@ysbaddaden ping, this seems like a couple characters fix then ready to merge

RX14:
@ysbaddaden are you interested in continuing this PR? It'd solve a long-standing problem in crystal and the code looks good

ysbaddaden:
If someone is willing to add the missing `sys/resource` bindings, please do :-)

damaxwell:
@RX14 , @ysbaddaden: On macOS it turns out that the SEGV always happens at an address lower than the top of the stack, but near the top of the stack.  The actual amount depends on the size of the stack frames being allocated in recursion.  So the check being done here never detects stack overflow.

Even on x86_64-linux-gnu, the SEGV isn't always within the body of the stack.  E.g., in some calls of a recursive function that allocates 4kb on the stack, sometimes I see something like

`top 0x7fff4a664000 addr 0x7fff4a665c98 bottom 0x7fff4ae64000`

where `addr` is the location of the SEGV. This triggers the check. But other times it's something like

`threshold 0x7ffc2cfd6000 addr 0x7ffc2cfd57e8 bottom 0x7ffc2d7d6000`

which does not.  

I think the thing to do is 

1) Expand the size of the check region to include a few reasonable sized stack frames (4K or so?) beyond the top of the stack.

2) If the check passes, report an error message like:

```
Invalid memory access (signal 11) at address 0x7fff4a665c98
Potential stack overflow (e.g., infinite or very deep recursion)
```

I.e. keep the original message (after all, although unlikely, it might have just been a vanilla bad pointer dereference).  But add on a clause that it looked like a stack overrun to help the diagnosis.  The aim isn't to be perfect, just helpful much of the time.

damaxwell:
About those bindings, I'd offer to help but I'm a little intimidated.  Presumably the linux versions will all be the same, and the darwin and BSD versions should be the same.  But that's a lot of hardware to test and verify on (esp. arm-linux-gnueabihf).  What have you guys done in the past?

Oh, and Windows. Windows doesn't have the POSIX `getrlimit` but has GetCurrentThreadStackLimits which might work instead.

RX14:
I'd say the last page in the stack and then the page after the stack.

ysbaddaden:
We use `mprotect` to protect the last 4KB. This can be bypassed, but it's weird that macOS is always below the stack, it should usually trigger a SIGSEGV within the last 4KB as on Linux.

RX14:
@damaxwell mentioned functions which allocate 4k on the stack, which could bypass a whole page.

damaxwell:
I took a look and I don't think there is a call to `mprotect` for the main fiber; that happens in `allocate_stack`, which isn't called. My initials tests only involved overrunning the stack on the main fiber, and even the trivial recursive program on macOS seg faults at an address below the top of the stack.

I just checked, and for spawned fibers on macOS, the seg faults are indeed within the stack for small recursive functions, just like Linux.

So the difference seems to be that the two OS's protect the end of the main stack differently, one within and one outside.

I think the easiest thing to do is just expand the check zone a page or two regardless.  This will address the macOS main thread, and the extra buffer will capture some other cases where stack allocation leads to a SIGSEGV beyond the end of the stack.

ysbaddaden:
Oh. We can't call `mprotect` on the main stack obviously. Good catch!

Since the program segfaulted, and if that happened within the stack or the next 4KB below it, but we safely say "Potential stack overflow (e.g., infinite or very deep recursion)", yes.

j8r:
@damaxwell you can try https://github.com/multiarch/alpine

straight-shoota:
With #6928 merged, this can be closed.

PR #6009, opened 2018-04-25
add json mapping extra fields

kostya:
this is extracted for json from https://github.com/crystal-lang/crystal/pull/4411

PR #6007, opened 2018-04-25
Don't leak variable name when assigning to ivar/cvar in method signature

asterite:
Fixes #5997
Fixes #4332

The idea is that this:

```crystal
def foo(@var)
end
```

expands to this:

```crystal
def foo(var __arg0)
  @var = __arg0 # the name __arg0 could be another one, you don't know it
end
```

instead of:

```crystal
def foo(var)
  @var = var
end
```

so you don't have access to that "implicit" local variable anymore.

I think this is good. It's a breaking change, but it's more intuitive. If you want it assigned to a local variable, explicitly assign it to a local variable.

Issue #3413, opened 2016-10-12
Undefined behavior when accessing an instance variable as default parameter


chenkovsky:
``` crystal
class A
  getter :a, :b

  def initialize(@a : Symbol, @b : Symbol = @a)
  end
end

a = A.new a: :xx
puts a.b
#output :skip
```

asterite:
Thank you for reporting this!

That `:skip` looks mysterious, but it's just the symbol with number 0. This proves it:

```cr
class A
  getter :a, :b

  def initialize(@a : Int32, @b : Int32 = @a)
  end
end

a = A.new a: 1
puts a.b # => 0
```

Definitely a bug, though.

asterite:
And I think the most probable solution we'll do is to disallow this (using instance variables in default values), it's pretty hard to implement.

bcardiff:
Note that using just the param `works`. I agree to disable ivars as default. Maybe a hint in case the ivar is assigned with this pattern could be added.

```crystal
class A
  getter :a, :b

  def initialize(@a : Int32, @b : Int32 = a)
  end
end

a = A.new a: 1
puts a.b # => 1
```

bew:
> And I think the most probable solution we'll do is to disallow this (using instance variables in default values), it's pretty hard to implement.

@asterite why is it hard to implement ?

Can't we just detect an ivar access, and verify that it has been set before ?
```cr
class Foo
  @a : Int32
  @b : Int32
  @c : Int32

  # This is invalid, @c is used before initialization
  def initialize(@a,   @b = @c,   @c = 42)
  end

  # This is valid, @a has been initialized before
  def initialize(@a,   @b = @a,   @c = 42)
  end
end
```

When/If #6007 is merged, we'd need to be careful and properly use the `__arg0` instead of `@a` when initializing `@b`, but that's doable I think.

vlazar:
Checked in Crystal 0.31.1 on MAcOS.

```
$ crystal --version
Crystal 0.31.1 (2019-10-02)

LLVM: 8.0.1
Default target: x86_64-apple-macosx
```

[Ary's example](https://github.com/crystal-lang/crystal/issues/3413#issuecomment-253213585) now gives a different result:

```crystal
class A
  getter :a, :b

  def initialize(@a : Int32, @b : Int32 = @a)
  end
end

a = A.new a: 1
puts a.b # => 0  (NO LONGER OUTPUTS 0)
```

Gives an error now

```
$ crystal test.cr
[...]

PR #6379, opened 2018-07-13
Parser: don't use invalid internal names for arguments like `@select` or `@@select`

### LINKED ISSUE #5997: Crystal parser `to_s` can produce invalid code
## Context

Hello, I'm playing around with the Crystal::Visitor, Parser, AST node and so.

I'm trying to implement a coverage binary which enrich the code with coverage operations. 
So far, I've made a proof of concept which is working is most of the cases (note: it's a PoC, the code is ugly) : https://github.com/anykeyh/crystal-coverage.

For that, I'm using the syntax visitor and crystal parser to regenerate code with additional instructions.

I'm also expanding all the relative `require` to generate one file only enriched with the coverage code. I assume the exported file can be passed to `crystal eval` operation.

### Example of code:

```crystal
class Test
  def foo
    n = 1
    while (n < 10)
      n += 1
      puts "foo!"
    end
  end

  def bar
    begin
      x = 1

      x = x + 1
      raise "Code below will never be covered"

      puts "Some code below"
    rescue
    end
  end
end

test = Test.new

test.foo
```

Same code enriched using the coverage binary:

```crystal
require "coverage/runtime"
::Coverage::File.new("spec/test.cr", "6df201f01e8cfb7f575a585521553d65",[23, 25, 3, 5, 4, 6, 12, 14, 15, 17])
#<loc:"spec/test.cr",0,0>
class Test
  def foo
    ::Coverage[0, 2] #<loc:"spec/test.cr",2,0>
        n = 1
    while (        ::Coverage[0, 4] #<loc:"spec/test.cr",3,0>
    n < 10
)
      ::Coverage[0, 3] #<loc:"spec/test.cr",4,0>
            n += 1
            ::Coverage[0, 5] #<loc:"spec/test.cr",5,0>
      puts("foo!")


    end

  end
  def bar
    ::Coverage[0, 6] #<loc:"spec/test.cr",11,0>
    begin
      x = 1
            ::Coverage[0, 7] #<loc:"spec/test.cr",13,0>
      x = x + 1

            ::Coverage[0, 8] #<loc:"spec/test.cr",14,0>
      raise("Code below will never be covered")

            ::Coverage[0, 9] #<loc:"spec/test.cr",16,0>
      puts("Some code below")

    rescue
    end
  end
end
::Coverage[0, 0] #<loc:"spec/test.cr",22,0>
test = Test.new

::Coverage[0, 1] #<loc:"spec/test.cr",24,0>
test.foo


::Coverage.get_results(Coverage::Outputter::Text.new)
```

### Problem

Since I depend heavily on `to_s` of the nodes, I've noticed a case where the code outputted is not valid code:

```crystal
  class Test
    def initialize(@select)
    end
  end
```

The AST `to_s` method will produce

```crystal
  class Test
    def initialize(select)
      @select = select
    end
  end
```

Since `select` is a reserved keyword, the code will not be evaluable. It obviously happens for any reserved keywords.

## How to reproduce

```crystal
require "compiler/crystal/syntax/*"

source = <<-CRYSTAL
  class Test
    def initialize(@select)
    end
  end
CRYSTAL

astree = Crystal::Parser.parse(source)
puts astree.to_s
```

asterite:
Fixes #5997

Right now this:

```crystal
def foo(@arg)
end
```

is expanded to this:

```crystal
def foo(arg)
  @arg = arg
end
```

The problem comes when we use a name that's also a keyword:

```crystal
def foo(@class)
end

# gets expanded to
def foo(class)
  @class = class
end
```

The above actually works fine, because the parser creates this and it can create a variable name named "class", it's just the user that can't do this. However, if we pass that to a macro, for example:

```crystal
macro id(x)
  {{x}}
end

id(def foo(@class)
end)
```

then it doesn't work because the code gets stringified and then output in the program, and then it's invalid:

```
Error in foo.cr:5: macro didn't expand to a valid program, it expanded to:

   1.   def foo(class)
   2.   @class = class
   3. end
Syntax error in expanded macro: id:1: cannot use 'class' as an argument name

  def foo(class)
          ^

id(def foo(@class)
^~
```

My [suggestion](https://github.com/crystal-lang/crystal/issues/5997#issuecomment-384279827) was to change the compiler to expand it using external named arguments:

```crystal
def foo(@arg)
end

# gets expanded to
def foo(arg __arg0)
  @arg = __arg0
end
```

The problem with that is that one can no longer do:

```crystal
def foo(@arg)
  puts arg # arg doesn't exist anymore, it's now called __arg0
end
```

However, if we only do this with names that are keywords, then there's no problem:

```crystal
def foo(@module)
  # this obviously doesn't compile, so it's no problem if the internal name is __arg0
  # because we have to always use @module to access the argument
  puts module
end
```

That's what this PR does.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -170,6 +170,8 @@ describe Crystal::Formatter do
   assert_format "def   foo (  @x, @y)  \n  end", "def foo(@x, @y)\nend"
   assert_format "def   foo (  @@x)  \n  end", "def foo(@@x)\nend"
   assert_format "def   foo (  &@block)  \n  end", "def foo(&@block)\nend"
+  assert_format "def   foo (  @select)  \n  end", "def foo(@select)\nend"
+  assert_format "def   foo (  @@select)  \n  end", "def foo(@@select)\nend"
   assert_format "def foo(a, &@b)\nend"
   assert_format "def   foo (  x  =   1 )  \n  end", "def foo(x = 1)\nend"
   assert_format "def   foo (  x  :  Int32 )  \n  end", "def foo(x : Int32)\nend"
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -176,7 +176,7 @@ module Crystal
     assert_syntax_error "def foo!=; end", "unexpected token: !="
     assert_syntax_error "def foo?=(x); end", "unexpected token: ?"
 
-    # #5895 & #6042
+    # #5895, #6042, #5997
     %w(
       begin nil true false yield with abstract
       def macro require case select if unless include
@@ -188,6 +188,8 @@ module Crystal
       assert_syntax_error "def foo(#{kw}); end", "cannot use '#{kw}' as an argument name", 1, 9
       assert_syntax_error "def foo(foo #{kw}); end", "cannot use '#{kw}' as an argument name", 1, 13
       it_parses "def foo(#{kw} foo); end", Def.new("foo", [Arg.new("foo", external_name: kw.to_s)])
+      it_parses "def foo(@#{kw}); end", Def.new("foo", [Arg.new("__arg0", external_name: kw.to_s)], [Assign.new("@#{kw}".instance_var, "__arg0".var)] of ASTNode)
+      it_parses "def foo(@@#{kw}); end", Def.new("foo", [Arg.new("__arg0", external_name: kw.to_s)], [Assign.new("@@#{kw}".class_var, "__arg0".var)] of ASTNode)
     end
 
     it_parses "def self.foo\n1\nend", Def.new("foo", body: 1.int32, receiver: "self".var)
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -29,7 +29,7 @@ module Crystal
       @def_nest = 0
       @type_nest = 0
       @call_args_nest = 0
-      @block_arg_count = 0
+      @temp_arg_count = 0
       @in_macro_expression = false
       @stop_on_yield = 0
       @inside_c_struct = false
@@ -1386,9 +1386,7 @@ module Crystal
       next_token_skip_space
 
       if @token.type == :"."
-        block_arg_name = "__arg#{@block_arg_count}"
-        @block_arg_count += 1
-
+        block_arg_name = temp_arg_name
         obj = Var.new(block_arg_name)
 
         @wants_regex = false
@@ -3662,6 +3660,20 @@ module Crystal
           raise "when specified, external name must be different than internal name", @token
         end
 
+        # If it's something like @select, we can't transform it to:
+        #
+        #     @select = select
+        #
+        # because if someone uses `to_s` later it will produce invalid code.
+        # So we do something like:
+        #
+        # def method(select __arg0)
+        #   @select = __arg0
+        # end
+        if !external_name && invalid_internal_name?(arg_name)
+          arg_name, external_name = temp_arg_name, arg_name
+        end
+
         ivar = InstanceVar.new(@token.value.to_s).at(location)
         var = Var.new(arg_name).at(location)
         assign = Assign.new(ivar, var).at(location)
@@ -3676,6 +3688,11 @@ module Crystal
         arg_name = @token.value.to_s[2..-1]
         if arg_name == external_name
           raise "when specified, external name must be different than internal name", @token
+        end
+
+        # Same case as :INSTANCE_VAR for things like @select
+        if !external_name && invalid_internal_name?(arg_name)
+          arg_name, external_name = temp_arg_name, arg_name
         end
 
         cvar = ClassVar.new(@token.value.to_s).at(location)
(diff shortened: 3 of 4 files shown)


jhass:
Let's please always rewrite to internal temporary locals. I would never have considered the access ivar arg as local by the same name more than an implementation detail, we should break this now and force people to do the assignment themselves.

asterite:
@jhass I'd rather not to. I tried, but [it was a mess](https://github.com/crystal-lang/crystal/pull/6007)

We must simply document this expansion, and that's it. There's also [this](https://github.com/crystal-lang/crystal/issues/5997#issuecomment-384294846).

straight-shoota:
@asterite I don't get it. Why do you think #6009 was a mess? That's how it should be IMHO. If you need to access an argument as a local variable, you simply shouldn't use the ivar argument short cut. That's it and everything should be fine.

asterite:
Like, take this line:

https://github.com/crystal-lang/crystal/pull/6007/files#diff-0e6efe33c8b047fbf52512a8cb7d2311L1697

```crystal
def initialize(@name, @args = [] of Arg, @return_type = nil, @varargs = false, @body = nil, @real_name = name)
```

That currently works, `@real_name = name` refers to `@name`. But changing that it will now refer to the method `name`. And because of a compiler bug which doesn't notice that this is accessing the instance var before it being used, it basically access a null pointer.

Now, I *could* fix that bug. But I can't, it's difficult and I don't have time. If someone wants to fix that bug and then change the way arguments work, then go ahead. But until then, I prefer to fix existing bugs (that I find easier :-P)

straight-shoota:
It would work if it was simply changed to this:
```cr
def initialize(name, @args = [] of Arg, @return_type = nil, @varargs = false, @body = nil, @real_name = name)
   @name = name
```
It's just that the compiler doesn't complain about the existing line?

asterite:
I know, that's what I did. But without doing it, it compiles and simply tries to access `name`, which access `@name`, which produces a segmentation fault. The compiler doesn't detect this, and I have no idea how to do it (that means I tried to look at the code to implement this and I found it very hard to do/detect).

straight-shoota:
Yeah, I was just checking I understand the problem.

But this issue exists regardless of `@ivar` argument aliases, doesn't it? If a method is used as an argument default value. It's equally undetectable.

The only special circumstance with ivar arguments is that it currently compiles and changing the language breaks code without raising a compiler error. I'd say, this is totally acceptable before 1.0.

asterite:
For me it's not acceptable. Segmentation faults are not fun.

But if you (community) want to go that direction, revivew my PR, approve it, and merge it (I won't do it because I'm against that change).

RX14:
The segfaults were a completely seperate bug that should be fixed. The change didn't introduce silent segfaults, it just exposed an existing bug. Please, lets fix the bug then merge the original PR.

asterite:
My reasoning is this:
- there are two existing bugs: one that generates invalid code with `to_s`, another one that produces segfault when using named args in som way
- this PR fixes the first bug
- what all of you are proposing is to not merge this PR, because we can change the behaviour of the language plus fix the bug that produces the segfault

So I see this as "No, let's keep the two existing bugs and wait to fix them at the same time instead of fixing one first, for which we already have a fix, and then the other".

The problem is that there's no guarantee we'll fix the bug, or when. This PR provides a fix for an existing bug, without changing existing behavior nor introducing segfaults. I can't see the harm of fixing this.

Then when that segfault bug is fixed and we can always used external names for `@foo` arguments, it's just about removing that `if` that decides when to use external arguments.

RX14:
yeah, you're absolutely right

asterite:
Thank you!

And note that I'm not against changing how `@foo` is implemented in arguments, it's just that the cost of fixing that without fixing the other bug is too high, in my opinion (as I said, the compiler started to segfault and it was hard to find why that happened, and I don't want to make users go through that process)

jhass:
Mmh, I wanted to log an issue for the bug, but it looks like I didn't fully get it yet, why does this correctly error out? https://carc.in/#/r/4ij0

asterite:
@jhass Interesting! I was going to point you to the issue that shows how to reproduce the bug, 3413, changing the instance variable access with a method that accesses it, but it turns out I can't reproduce #3413 anymore in master, it gives a correct error. I also tried reproducing it with a method access and it gives the same error. I don't know when it got fixed...

That's good news. I will try to see if I can apply this PR then to all instance/class variable arguments and see if I don't get a segfault anymore and the errors are easy to fix.

RX14:
@asterite a git bisect would be interesting

asterite:
@RX14 Yeah, I might try that too...

I'm implementing this PR for all instance/class vars but I'm not sure I like it (like before). Sometimes there's code like this:

```crystal
@x : Int32?

def foo(@x)
  z = x ? x.abs : 0
end
```

Basically, `@x` means receiving an argument and assigning it to the instance var. With this change that argument is lost and we only have access to the instance var. So we change it:

```crystal
@x : Int32?

def foo(@x)
  z = @x ? @x.abs : 0
end
```

and now it doesn't work, because there's no type filtering for instance vars. So we must do:

```crystal
@x : Int32?

def foo(x)
  @x = x
  z = x ? x.abs : 0
end
```

Essentially, we lost the ability to use `@x` in arguments if you also want to type filter them through the local variable that it gets assigned to first. And that local variable **always exists**, it's just that you want to hide it.

I'm not sure what's the benefit of all this... having to write more code? Also, accessing a local variable is faster than accessing an instance variable.

Someone care to tell me what are the benefits of this change? We can simply document that `@x` as an argument expands to `@x = x` inside the method, which is the current behavior.

asterite:
False alarm, I can still reproduce the bug:

```crystal
class A
  property a : Int32
  property b : Int32

  def initialize(@a, @b = a)
  end
end

a = A.new(1)
puts a.b # => 0
```

bew:
To me your example above should be written:
```cr
@x : Int32?

def foo(@x)
  z = (x = @x) ? x.abs : 0
end
```
Assigning to a local var to have a smaller type. Right?

jhass:
@asterite the point is that the behavior is entirely non-obvious as it's compiler generated code *affecting (or I'd say polluting) the local scope*. I would go as far as saying it leads to more brittle code as refactoring (renaming things) can easily to changed semantics (lvar access -> method call), as you basically experienced it yourself when trying to change this behavior.

Sometimes the longer variant (I wouldn't call it redundant) is better because it makes the code easier to read and follow. It's one less thing to keep in the back of your head when reading code, and this is a quite big one in my opinion.

straight-shoota:
Okay, I think we're discussing the general issue with ivar arguments in too many different places where it's mostly off-topic. Let's have a proper discussion reflecting the already mentioned arguments and current issues with this.

Issue #6372, opened 2018-07-12
Spec compilation failure (read before assignment to local variable 'base')


myfreeweb:
Weird compile error is happening when running specs with a Crystal that was built with LLVM **6.0.1** on FreeBSD:

```
in macro 'macro_1343766128' /usr/home/greg/crystal/spec/std/file_spec.cr:587, line 8:

   1.
   2.       it "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do                                                                                  
   3.         prev_home = home
   4.         begin
   5.           ENV["HOME"] = File.expand_path(datapath)
   6.           File.expand_path("~/").should eq(home)
   7.           File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))                                                                                       
>  8.           File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))                                                                         
          
read before assignment to local variable 'base'
```

(`base` is a private function in the file, not a local variable)

/cc #5556

felixbuenemann:
I'm also running into this on llvm 6.0.1 from Homebrew on macOS 10.13.5, will have to check with crystal compiled against llvm@5 (5.0.2).

I'll see if a full `make clean` helps, I only forced recompile of `llvm_ext.o` after switching from LLVM 5 to 6 because of patches required for 6.0.1.

asterite:
The error message happens at the semantic stage, which has nothing to do with LLVM, so I really doubt it has anything to do with the LLVM version.

felixbuenemann:
Yeah you are right, I'm also running into it when compiling with 5.0.2.

I also tried `make clean` and clearing out `.crystal` and `~/.cache/crystal` as well as using the precompiled 0.25.1 binaries for compilation and the error persisted.

felixbuenemann:
The error goes away if I remove the wrapping `{% unless flag?(:win32) %}` macro.

felixbuenemann:
Oh and compiling the file_spec standalone also works, but all_spec fails (make spec).

straight-shoota:
I presume this has been resolved.

Issue #5556, opened 2018-01-08
Support LLVM6


ysbaddaden:
LLVM 6.0 was branched, we may start investigating what got broken, and more importantly: what got added to the C API (surprise).

asterite:
Surprise?? 😮

ysbaddaden:
From an initial attempt (ubuntu trusty, using prebuilt https://apt.llvm.org packages), I got a few hiccups. For example missing support for GCC 4.8.5 (`no option -Wdate-time`), and a missing symbol (`LLVMInitializeInstCombine`) probably due to the packages, since we never call the function directly, and fixed with the `--link-static` flag, such as:

```sh
$ make clean crystal \
  LLVM_CONFIG="llvm-config-6.0 --link-static" \
  CXX=clang++-5.0 CC=clang-5.0
```

I then got actual compatibility issues:

```
/home/github/crystal/src/llvm/ext/llvm_ext.o: In function `LLVMDIBuilderCreateCompileUnit':
src/llvm/ext/llvm_ext.cc:(.text.LLVMDIBuilderCreateCompileUnit+0x12f): undefined reference to `llvm::DIBuilder::createCompileUnit(unsigned int, llvm::DIFile*, llvm::StringRef, bool, llvm::StringRef, unsigned int, llvm::StringRef, llvm::DICompileUnit::DebugEmissionKind, unsigned long, bool, bool, bool)'
/home/github/crystal/src/llvm/ext/llvm_ext.o: In function `LLVMDIBuilderCreateFunction':
src/llvm/ext/llvm_ext.cc:(.text.LLVMDIBuilderCreateFunction+0xf1): undefined reference to `llvm::DIBuilder::createFunction(llvm::DIScope*, llvm::StringRef, llvm::StringRef, llvm::DIFile*, unsigned int, llvm::DISubroutineType*, bool, bool, unsigned int, llvm::DINode::DIFlags, bool, llvm::MDTupleTypedArrayWrapper<llvm::DITemplateParameter>, llvm::DISubprogram*, llvm::MDTupleTypedArrayWrapper<llvm::DIType>)'
/home/github/crystal/src/llvm/ext/llvm_ext.o: In function `LLVMDIBuilderCreatePointerType':
[...]

PR #6380, opened 2018-07-13
Fix compilation with LLVM 6.0.1

felixbuenemann:
This fixes the following compile error on LLVM 6.0.1 (and 6.0.0):

```
Using /usr/local/opt/llvm/bin/llvm-config [version=6.0.1]
c++ -c  -o src/llvm/ext/llvm_ext.o src/llvm/ext/llvm_ext.cc -I/usr/local/Cellar/llvm/6.0.1/include  -stdlib=libc++ -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -std=c++11 -Wall -W -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wcovered-switch-default -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wstring-conversion -DNDEBUG   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
CRYSTAL_CONFIG_PATH="/Users/felix/dev/misc/crystal/src" ./bin/crystal build --mcpu=broadwell --release  -o .build/crystal src/compiler/crystal.cr -D without_openssl -D without_zlib
duplicate symbol _LLVMDIBuilderCreateCompileUnit in:
    /Users/felix/dev/misc/crystal/src/llvm/ext/llvm_ext.o
    /usr/local/Cellar/llvm/6.0.1/lib/libLLVMCore.a(DebugInfo.cpp.o)
duplicate symbol _LLVMDIBuilderFinalize in:
    /Users/felix/dev/misc/crystal/src/llvm/ext/llvm_ext.o
    /usr/local/Cellar/llvm/6.0.1/lib/libLLVMCore.a(DebugInfo.cpp.o)
duplicate symbol _LLVMDIBuilderCreateFile in:
    /Users/felix/dev/misc/crystal/src/llvm/ext/llvm_ext.o
    /usr/local/Cellar/llvm/6.0.1/lib/libLLVMCore.a(DebugInfo.cpp.o)
ld: 3 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error: execution of command failed with code: 1: `cc "${@}" -o '/Users/felix/dev/misc/crystal/.build/crystal'  -rdynamic  /Users/felix/dev/misc/crystal/src/llvm/ext/llvm_ext.o `/usr/local/opt/llvm/bin/llvm-config --libs --system-libs --ldflags 2> /dev/null` -lstdc++ -lpcre -lgc -lpthread /Users/felix/dev/misc/crystal/src/ext/libcrystal.a -levent -liconv -ldl -L/usr/lib -L/usr/local/lib`
make: *** [.build/crystal] Error 1
```

Since the functions `LLVMDIBuilderCreateCompileUnit` and `LLVMDIBuilderCreateFile` exist in LLVM 6.0, but have a much different signature, I chose to instead rename the conflicting methods by adding a "2" suffix, like already done for `LLVMSetCurrentDebugLocation`.

This is related to #5556.

I'm currently running into issue #6372 during `make spec`, but it also happens with LLVM 5.0.2.
The specs pass fine if I apply the fix described in https://github.com/crystal-lang/crystal/issues/6372#issuecomment-404872960.


RX14:
But crystal builds on LLVM6 right now, and has done for months on archlinux. What changed?

felixbuenemann:
I'm not sure – have you tested with 6.0.1 on archlinux?

I'll see if I still have 6.0.0 on my mac to compare.

RX14:
And none of the arch patches to LLVM 6.0.1 are to anything but the build system: https://git.archlinux.org/svntogit/packages.git/tree/trunk?h=packages/llvm

RX14:
Did you remember to run `make clean` after changing LLVM versions?

felixbuenemann:
I started with a clean slate, `make clean` and `rm -rf ~/.cache/crystal/*`.

felixbuenemann:
Maybe the symbols just don't clash on Linux. Linking on Darwin differs a fair bit from Linux.

felixbuenemann:
@RX14 If you like I can update the `prepare_build` step in the `bin/ci` script so that it uses "llvm" instead of "llvm@5". It's currently using whatever versions is used by the crystal formula in homebrew.

felixbuenemann:
Hmm, this is kind of a hen-egg problem. The ci script installs crystal from homebrew and then links the matching llvm version that the crystal formula depends on.

In oder to fix that I should update the homebrew formula to build against llvm 6 and they will ask if the patch was upstreamed :-)

What I could do is create a new PR in homebre-core that only updates the dependency but does not yet include the patch and wait for it to fail. Then I could point you at the homebrew-core jenkins build results and then push the patch in the homebrew formula in another commit to show if it fixes the build.

felixbuenemann:
Oh well, since @ilovezfs mentioned interest in bumping crystal's llvm dependency to 6.0 in der 0.25.0 release discussion I might get that through.

ilovezfs:
See also https://github.com/Homebrew/homebrew-core/pull/29774#issuecomment-402929787

Same failure.

felixbuenemann:
@matiasgarciaisaia do you get the same `make spec` error as described in #6372?

RX14:
> In oder to fix that I should update the homebrew formula to build against llvm 6 and they will ask if the patch was upstreamed :-)

You don't need the CI to run on LLVM 6 to patch crystal to support LLVM 6

felixbuenemann:
> You don't need the CI to run on LLVM 6 to patch crystal to support LLVM 6
No, only if you want to reproduce the compile error in the CI on macOS.

RX14:
That's not neccesary

RX14:
Okay, I've looked into this a bit further and i'm okay with this patch. These calls *can* be replaced with their official counterparts but i'd rather merge a quickfix patch then one that uses the new llvm-c versions. My only feedback is to `s/2/Crystal/` in the method names.

felixbuenemann:
Oh, and I think we should add `#include <llvm-c/DebugInfo.h>` header for LLVM 6.0, since that's where `LLVMDIBuilderFinalize` is defined and it's not included by `<llvm-c/Core.h>`.

I'll push an update for that shortly.

RX14:
@felixbuenemann it doesn't need to include `llvm-c/DebugInfo.h`. The technically correct solution would to be to bind `LLVMDIBuilderFinalize` in `LibLLVMExt` in < 6 and `LibLLVM` in > 6, but it really doesn't matter because we always require both libs or neither. Please leave this as-is.

RX14:
and yes, if we already use the `2` convention, then lets keep it.

felixbuenemann:
Too late :-(

Including the header allows us to spot conflicts like this easier, since they blow up at compile time instead of at link time. It would also have allowed to spot the error on Linux earlier where the symbols somehow didn't clash.

If you still think it should be removed let me now and I'll revert to the previous version.

RX14:
if that's the case, we should include it on every version.

felixbuenemann:
The header was added in LLVM 6.0 so that will be very difficult to achieve ;-)

In the future the bindings should probably be rewritten to use the new LLVM C DebugInfo API instead of the C++ bindings, but that would break backwards compatibility (or there would be two different implementations).

src/llvm/ext/llvm_ext.cc
```diff
@@ -56,14 +60,16 @@ LLVMDIBuilderRef LLVMNewDIBuilder(LLVMModuleRef mref) {
   return wrap(new DIBuilder(*m));
 }
 
+#if LLVM_VERSION_LE(5, 9)
```
ysbaddaden: Note: there won't be any 5.x other than 5.0. `LE(5, 0)` is enough, no need to indroduce unknown version numbers :)

ysbaddaden:
This is probably an acceptable quick fix, but since LLVM is adding C bindings for Debug Information, we should evolve our custom set of C++ to match the new API for older LLVM releases, so we don't break anything, and can move on. Hopefully dropping the custom C++ someday.

src/llvm/ext/llvm_ext.cc
felixbuenemann: Good catch, I just fixed this locally, but @RX14 merged in parallel before I pushed it.

I'll send a tiny new PR to fix it and reference it here.

felixbuenemann: @RX14 @ysbaddaden I fixed this in #6383.

PR #6383, opened 2018-07-14
Fix version check for LLVM < 6.0

felixbuenemann:
This is a minor fixup for the LLVM 6.0 support quick fix  (#6380) that changes the versions check for "below 6.0" (<= 5.9) to <= 5.0 since there wont be any new 5.x versions after 5.0.

### DIFF
--- a/src/llvm/ext/llvm_ext.cc
+++ b/src/llvm/ext/llvm_ext.cc
@@ -60,7 +60,7 @@ LLVMDIBuilderRef LLVMNewDIBuilder(LLVMModuleRef mref) {
   return wrap(new DIBuilder(*m));
 }
 
-#if LLVM_VERSION_LE(5, 9)
+#if LLVM_VERSION_LE(5, 0)
 void LLVMDIBuilderFinalize(LLVMDIBuilderRef dref) { unwrap(dref)->finalize(); }
 #endif
 


RX14:
Not sure this is worth merging.

ilovezfs:
http://blog.llvm.org/2016/12/llvms-new-versioning-scheme.html
>Patch releases will increment the patch component as before (producing versions like 5.0.1), and the minor component will stay at zero since no minor releases will be made.

RX14:
This is *either* incorrect or a purely-aesthetic change, depending on the implementation of `LLVM_VERSION_LE`. I vote close.

ilovezfs:
Why not change the 9 to a prettier number like 42?

felixbuenemann:
> This is either an incorrect PR or a purely-aesthetic change, depending on the implementation of `LLVM_VERSION_LE`.

I can assure you it is not incorrect, the previous code was.

The `LLVM_VERSION_LE` macro only checks the major and minor version of LLVM. After LLVM 5.0 there is 6.0 and the next version will be 7.0. There are patch releases like 5.0.1/5.0.2, but there will never be a 5.1 release.

The only reason I initially used 5.9 as a check, is because I wanted to write "lower than 6.0" and then noticed the macro was lower-than-or-equal so I change it from 6.0 to 5.9, which technically works, but is confusing to the reader of the code.

PR #6384, opened 2018-07-14
Fix reference links to .from_io methods

faustinoaq:
Ref: [`read_bytes`](https://crystal-lang.org/api/master/IO.html#read_bytes%28type%2Cformat%3AIO%3A%3AByteFormat%3DIO%3A%3AByteFormat%3A%3ASystemEndian%29-instance-method)

Notice the "See ..." section

image

Comparison to [`write_bytes`](https://crystal-lang.org/api/master/IO.html#write_bytes(object,format:IO::ByteFormat=IO::ByteFormat::SystemEndian)-instance-method)

image

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -877,7 +877,7 @@ abstract class IO
   # `from_io(io : IO, format : IO::ByteFormat = IO::ByteFormat::SystemEndian)`
   # method can be read in this way.
   #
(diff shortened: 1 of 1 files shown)


faustinoaq:
> That's incorrect, since they're class methods.

@Sija Oh, you're right, `#to_io` is different to `.from_io` :sweat_smile: , Any idea why this link wasn't generated here?

Issue #6357, opened 2018-07-09
[RFC] Implement Ruby's Enumerator Class


felixbuenemann:
I really like Ruby's [Enumerator](https://ruby-doc.org/core-2.5.1/Enumerator.html) class and I think it would be great to have something similar in Crystal.

The `Enumerator` class can be used to implement generators.

Here's an example from the Ruby docs:

```ruby
fib = Enumerator.new do |y|
  a = b = 1
  loop do
    y << a
    a, b = b, a + b
  end
end

p fib.take(10) # => [1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
```

The `y` block parameter in the example is called the *yielder* and the iterator returned from `new` blocks until the yielder receives a value or the block ends.

Since I was porting some Ruby code that depended on `Enumerator` to Crystal, I wen't ahead and implemented my own generic `Enumerator` class using the `Iterator` module and `Channel` to implement the concurrent behavior:

```crystal
class Enumerator(T)
  include Iterator(T)

  def initialize(&block : Yielder(T) -> Nil)
    @channel = Channel(T).new
    @block = block
    @block_called = false
    spawn_block
  end

  def next
    @channel.receive
  rescue Channel::ClosedError
    stop
  end

  def rewind
    if @block_called
      @channel = Channel(T).new
      @block_called = false
      spawn_block
    end
    self
  end

  private def spawn_block : Nil
    spawn do
      begin
        @block.call(Yielder(T).new(@channel))
      ensure
        @channel.close
        @block_called = true
      end
    end
  end

  private struct Yielder(T)
    def initialize(@channel : Channel(T))
    end

    def <<(data : T)
      @channel.send(data)
      self
    end
  end
end
```

The Fibonacci generator could now be implemented similar to the Ruby example:

```crystal
fib = Enumerator(UInt32).new do |y|
  a = b = 1_u32
  loop do
    y << a
    a, b = b, a + b
  end
end

p fib.first(10).to_a # => [1, 1, 2, 3, 5, 8, 13, 21, 34, 55]
```

**Please let me know if you think this is a useful addition to the Crystal Standard Library.**

If you do, I would be happy to polish the code up into a PR along with documentation and specs. 

I've also implemented `Enumerator#peek` locally, which allows to look at the next iterator value without moving it forward by buffering the value on peek, but left it out of the example for simplicity.

asterite:
Yup, it's been discussed in the past:

[...]

PR #6385, opened 2018-07-14
[WIP] Implement ruby-style Enumerator

felixbuenemann:
This is a work-in-progress and shouldn't be merged yet!

I'll update the description of the PR once the details are fully flushed out.

For discussion see #6357.

A big thanks to @asterite who helped in implementing most of the fiber stuff.


RX14:
If you turn `@@stack_pool` into another linked list of fibers, then you can skip the whole mark/sweep fiber collection to avoid allocating in `finalize`.

RX14:
Also, I'd really *really* rather not call it an enumerator. Given my java/C# experience `Enumerator` == `Iterator`. I'd name it `Iterator::Fiber` or `Iterator::Yielding` or something more descriptive than `Enumerator`.

felixbuenemann:
I agree that having both `Iterator` and `Enumerator` is very confusing and should be avoided.

I've spend quite some time thinking about this and I think it's the wrong level of abstraction.

* Maybe the right abstraction is having a top level `Yielder` class that can be used to wrap a suspendable block of code and just having an `Iterator.yielder(&block)` convenience class method to construct a Yielder from a block and use it as an iterator, similar to `Iterator.of`.
* Maybe this was a bad idea in the first place and should be scrapped.

I'll update #6357 with a more detailed analysis once I've fully thought it through.

asterite:
For me Enumerator is fine. It's also called like that in Ruby..

felixbuenemann:
Yes, but Ruby does not have an `Iterator` class.

asterite:
But Java does.
Iterator in Crystal is very similar to Java.
Enumerator in Crystal will be very similar to Ruby.
I personally think that's fine.

RX14:
@asterite but Enumerator in C# is very similar to Iterator in Crystal. That's the confusion. The naming doesn't really matter anyway, since we can probably make the interface just `Iterator.new(&block)` or `Iterator.yielder(&block)`.

asterite:
That's what I thought, but no, because you always have to specify the generic type. But maybe Iterator::Something(T) could work.

I'm still not 100% convinced about the implementation, I feel like something is missing.

felixbuenemann:
> I'm still not 100% convinced about the implementation, I feel like something is missing.

I'm not too happy about the changes to Fiber that the current implementation requires – it seems overkill. I think I'll explore an alternate approach of raising an exception inside the yielder block.

src/enumerator.cr
```diff
@@ -0,0 +1,146 @@
...
+  @done : Bool?
+
+  def initialize(&@block : Yielder(T) ->)
+    @current_fiber = Fiber.current
+    @stack = Deque(T).new(1)
+    @yielder = Yielder(T).new(self)
+    run
+  end
+
+  # Returns the next element yielded from the block or `Iterator::Stop::INSTANCE`
```
wontruefree: Since this is an Enumerator should the return type be an `Enumerator::Stop`?

felixbuenemann: I haven't tested it, but since the `Iterator` module is included it probably already works with both `Enumerator::Stop` and `Iterator::Stop`.

PR #6387, opened 2018-07-15
Fix missing comment inside code block example

faustinoaq:
See: `The time-of-day can be...`

image

Ref: https://crystal-lang.org/api/master/Time.html


Sija:
needs `crystal tool format`.

straight-shoota:
@faustinoaq I don't think you need to bother including a screenshot for such a trivial change 👍

sdogruyol:
Thanks @faustinoaq :+1:

Issue #6389, opened 2018-07-15
Move to_<num_type> from String to Slice


bew:
While working on a terminfo parsing library, where I manipulate Slice of bytes (`Bytes` type), I needed to convert digits to an actual number.
I realized that all the `to_<num_type>` conversion methods (like `to_i64`, `to_u8`, `to_f`, ..) are in `String`, and I needed to allocate a `String` just to convert the bytes to a number.

I think thoses methods should be available in `Slice` too. The number conversion methods don't need anything from `String` as they directly work with a pointer and a size. They can almost directly be copy/pasted to `Slice`!

I have a working [branch](https://github.com/bew/crystal/tree/to_num-on-slice) with this change, I can make a PR when it's accepted!

Note: the only problem I see here is that those methods will now exists for all kind of Slice, not only `Slice(UInt8)`.. But maybe it's ok? or maybe we can add a macro check inside the `to_<num_type>` methods that raise when `T != UInt8` ?
Another possibility would be to make `Bytes` a new type (not just an alias) that inherits from `Slice(UInt8)` so it's easier to add methods that works specifically on bytes.

WDYT?

RX14:
I agree that `String.unsafe_from_slice` is probably a better solution. 

Unfortunately, it can't be implemented because strings are layed out differently from every other class. Strings don't have a pointer to their data, they have their data directly after their class in memory, so it's impossible to turn a Bytes into a String without allocating.

bew:
Having `to_i` on `Slice` is only logic for `Slice(UInt8)`, right @asterite ? What do you think should be done for a `Slice(Person)` ?

asterite:
Give an error. We already have methods on slice that only work for byte slices. And byte slices are the most common slice type (in fact, but I would happily change Slice(T) to a non generic Bytes type) and remove the Slice type from the std, or maybe keep the two but return Bytes instead of Slice(UInt8) in all cases).

bew:
Ok, that's what I thought too. And I was also thinking about a `Bytes` type which inherits from `Slice(UInt8)` and would have Bytes-specific methods like `to_*` methods.

asterite:
Can't inherit from struct. They would be two different types with a few similar methods.

bew:
oh right, then maybe make it as a wrapper like:
```cr
struct Bytes
  def initialize(@raw : Slice(UInt8))
  end

  forward_missing_to @raw

  def to_i(...)
    # ...
  end
end
```

[...]

PR #6397, opened 2018-07-16
(WIP) Optimize String#gsub with block

straight-shoota:
When replacing characters in a string, typically only a minority of characters are changed like most most escaping or encoding use-cases. Currently, `String#gsub` loops through the string, yields each character and writes the return to an IO. However, these single-character writes are not as efficient as writing a larger slice.

The best example for this is a string with no replacement. In the previous implementation, it would be copied byte-by-byte (or rather char-by-char) to a new string. If some characters need to be replaced, the size of the slices that can be copied as whole is smaller, but it's still considerably faster than writing byte-by-byte.

An algorithm that only writes replaced chars and copies a series of unchanged characters in bulk can improve performance considerably.
This approach was previously used by a custom implementation in [`JSON::Builder#string`](https://github.com/crystal-lang/crystal/blob/466e0a2201c2e5b225cf4452993a95353c2b63d7/src/json/builder.cr#L95-L146).

In this PR, the proposed implementation of `String#gsub` yields each char and only writes the returned value if it is not `nil`. When using this method, returning `nil` signals that the current character can be copied as is and multiple `nil` returns are combined to a larger slice copy.

I also added a method variant receiving an `io` for better integration. And there is a similar `Slice#gsub` method, which is utilized for ascii-only strings (ore replacements). This method is part of the public API of `Slice` because it could be useful for any slice to replace some elements. It requires some sort of builder with an `IO` API, so currently it's only  effectively usable for `Slice(UInt8)`.

```
replace 6/49 chars: "Alle meine Entchen schwimmen auf dem Sä".gsub(' ', '-')

                   old   1.15M (865.84ns) (±12.39%)  192 B/op   1.36× slower
                   new 883.68k (  1.13µs) (±14.82%)  160 B/op   1.77× slower
new (ascii_only: true)   1.57M (637.87ns) (±10.70%)  160 B/op        fastest
```

```
replace all: "äääääääääääääääää".gsub('ä', '-')

                   old   1.03M (971.22ns) (±11.46%)  144 B/op   1.02× slower
                   new 739.62k (  1.35µs) (± 9.79%)  144 B/op   1.42× slower
new (ascii_only: true)   1.05M (948.91ns) (± 9.81%)  144 B/op        fastest
```

```
replace none: "Alle-meine-Entchen-schwimmen-auf-dem-Sä".gsub(' ', '-')

                   old   1.57M (638.41ns) (±15.57%)  176 B/op   1.60× slower
                   new   1.17M (852.01ns) (±10.92%)  160 B/op   2.13× slower
new (ascii_only: true)   2.51M (399.19ns) (±15.46%)  160 B/op        fastest
```

I also tested `JSON::Builder#string` and the new gsub is even slightly faster than the previous implementation because it doesn't need to decode multibyte chars (they won't be replaced anyway).

```
simple string
old   4.75M (210.51ns) (±20.85%)  12 B/op   1.15× slower
new   5.45M (183.33ns) (±29.58%)  13 B/op        fastest
many replacements
old 648.03k (  1.54µs) (±18.61%)  149 B/op   1.05× slower
new 683.31k (  1.46µs) (±20.73%)  203 B/op        fastest
normal replacements
old 812.58k (  1.23µs) (±25.99%)  253 B/op   1.25× slower
new   1.01M (985.81ns) (±25.16%)  282 B/op        fastest
```

Obviously, there are cases where the previous, simple implementation is equal or even more performant then the proposed batch copy. This is typically the case when there is a high amount and equal distribution of replaced chars, especially in ascii-only strings. For some use cases, a customized replace loop can achieve more performance than this generalized `String#gsub`.

In total, I expect this to be an improvement (backed by the benchmarks) because replacements are usually less frequent in most applications.

### DIFF
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -354,6 +354,26 @@ struct Slice(T)
     source.move_to(self)
   end
 
+  def gsub(buffer, &block : T -> _)
+    current_chunk_start = 0
+    each_index do |pos|
+      item = unsafe_at(pos)
+      if result = yield item
+        buffer.write Slice.new(to_unsafe + current_chunk_start, pos - current_chunk_start)
+        current_chunk_start = pos + 1
+
+        if result.is_a?(Tuple)
+          result.each { |part| buffer << part }
+        else
+          buffer << result
+        end
+      end
+    end
+    unless current_chunk_start == size
+      buffer.write Slice.new(to_unsafe + current_chunk_start, size - current_chunk_start)
+    end
+  end
+
   def inspect(io)
     to_s(io)
   end
--- a/spec/std/json/builder_spec.cr
+++ b/spec/std/json/builder_spec.cr
@@ -61,9 +61,15 @@ describe JSON::Builder do
     end
   end
 
-  it "writes string with controls and slashes " do
-    assert_built("\" \\\" \\\\ \\b \\f \\n \\r \\t \\u0019 \"") do
-      string(" \" \\ \b \f \n \r \t \u{19} ")
+  it "writes string with controls and slashes" do
+    assert_built(%(" \\\" \\\\ \\b \\f \\n \\r \\t \\u0019")) do |json|
+      string(" \" \\ \b \f \n \r \t \u{19}")
+    end
+  end
+
+  it "writes string with utf-8" do
+    assert_built(%("\u{1F609}")) do |json|
+      string("\u{1F609}")
     end
   end
 
--- a/src/csv/builder.cr
+++ b/src/csv/builder.cr
@@ -73,12 +73,10 @@ class CSV::Builder
   def quote_cell(value)
     append_cell do
       @io << @quote_char
-      value.each_byte do |byte|
-        case byte
+      value.gsub(@io, ascii_only: true) do |char|
+        case char
         when @quote_char
-          @io << @quote_char << @quote_char
-        else
-          @io.write_byte byte
+          next @quote_char, @quote_char
         end
       end
       @io << @quote_char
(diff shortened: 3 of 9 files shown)


src/slice.cr
```diff
@@ -354,6 +354,26 @@ struct Slice(T)
     source.move_to(self)
   end
 
+  def gsub(buffer, &block : T -> _)
```
RX14: This only compiles if `T` is `UInt8`. Are we sure we want this on slice instead of string?

bew: It looks like a similar problem as in #6389 where we have methods that make sense only for `Slice(UInt8)`, right?

straight-shoota: It's somewhat similar. But in contrast to #6389, this method conceptually fits for any slice. It's only that the `buffer` API is currently implemented by `IO` which is based on `UInt8`. But you can implement such a receiver for any type.

I'm not entirely sure about it's usefulnes, so maybe it's still better to move it to `String` as a private method.

ysbaddaden:
I don't understand your benchmarks: new is always slower than old ?!

RX14:
if old is always better than new in the non-ascii-only case, then why not keep the old implementation in that case?

straight-shoota:
I'm not sure why the non-ascii performance is so bad. In my initial benchmarks, it showed significant improvement. And the implementation is basically the same as for ascii-only. This needs investigation.

r00ster91:
@straight-shoota in the past I've also had benchmarking issues, but sometimes it actually helped to clear the cache: `sudo rm -rf ~/.cache/crystal/`. It can be glitchy sometimes.

asterite:
Yeah, there's a problem with how the compiler's cache works, I think in release it doesn't work quite well. I'll have to investigate, but I'll need reproducible code(s).

PR #6404, opened 2018-07-17
Add API docs for StaticArray

straight-shoota:


src/static_array.cr
```diff
@@ -1,4 +1,26 @@
 # A fixed-size, stack allocated array.
+#
+# `StaticArray` is a generic type with type argument `T` specifying the type of
+# it's elements and `N` the fixed size. For example `StaticArray(Int32, 3)`
```
r00ster91: I think "its" instead of "it's" is correct here.

src/static_array.cr
```diff
@@ -1,4 +1,26 @@
 # A fixed-size, stack allocated array.
+#
+# `StaticArray` is a generic type with type argument `T` specifying the type of
+# its elements and `N` the fixed size. For example `StaticArray(Int32, 3)`
```
vlazar: Shouldn't it be
> its elements and `N` **is** the fixed size.

straight-shoota:
I added an example for use with `uninitialized`. @RX14 @jhass could you pleas re-review?

src/static_array.cr
vlazar: @straight-shoota Thanks for explanation. TIL.

PR #6414, opened 2018-07-20
Compiler: evaluate instance var initializers at the metaclass level

### LINKED ISSUE #5876: Module validation failed: Call parameter type does not match function signature!
Looked through opened & closed issues, this doesn't seem to be reported yet, but I'm pretty unfamiliar with these kinds of errors. 
```cr
class Foo
  def bar
    puts "works"
  end

  @callbacks : Array(Proc(Nil)) = [
    ->{ bar }
  ]
end

Foo.new
```
```
Module validation failed: Call parameter type does not match function signature!
i32 138
 %Foo*  call void @"*Foo#bar:Nil"(i32 138)

???
???
???
???
???
__crystal_main
main
__libc_start_main
_start
???
```
This is all I have to test right now but is reproduced on carc.in's environment:
```
Crystal 0.24.1 (2017-12-20)                             
                                                        
LLVM: 5.0.0                                             
```
https://carc.in/#/r/3s96

asterite:
Fixes #5876

This currently compiles:

```crystal
class Foo
  @x : Int32 = bar

  def bar
    1
  end
end
```

However, it makes very little sense to run an instance var initializer at the instance level, because that's what we are initializing.

Java allows that behavior, but C# rejects it, demanding that `bar` has to be a static method. That makes much more sense to me.

In any case, you can initialize it at the instance level if you want to, in an `initialize` method:

```crystal
class Foo
  @x : Int32

  def initialize
    @x = bar
  end

  def bar
    1
  end
end
```

That makes it much more clear that this is run at the instance level, and additionally the compiler will check that the instance var isn't used before it's assigned a value, etc.

It also makes sense from a scope level. If we do:

```crystal
class Foo
  def bar
    1
  end

  bar # Error
  CONST = bar # error
  # but this works?: @x : Int32 = bar
end
```

So now it's uniform:

```crystal
class Foo
  @x : Int32 = bar # ok
  bar # ok
  CONST = bar # ok

  def self.bar
    1
  end
end
```

This is a breaking change, but I doubt there's a lot of code out there that uses this kind of initializers.

### DIFF
--- a/spec/compiler/semantic/instance_var_spec.cr
+++ b/spec/compiler/semantic/instance_var_spec.cr
@@ -5042,4 +5042,71 @@ describe "Semantic: instance var" do
       Second.new.hash
       )) { hash_of(hash_of(int32, int32).virtual_type, int32).virtual_type }
   end
+
+  it "doesn't solve instance var initializer in instance context (1) (#5876)" do
+    assert_error %(
+      class Foo
+        @x : Int32 = bar
+
+        def bar
+          1
+        end
+      end
+
+      Foo.new
+      ),
+      "undefined local variable or method 'bar'"
+  end
+
+  it "doesn't solve instance var initializer in instance context (2) (#5876)" do
+    assert_error %(
+      class Foo(T)
+        @x : T = bar
+
+        def bar
+          1
+        end
+      end
+
+      Foo(Int32).new
+      ),
+      "undefined local variable or method 'bar'"
+  end
+
+  it "doesn't solve instance var initializer in instance context (3) (#5876)" do
+    assert_error %(
+      module Moo(T)
+        @x : T = bar
+
+        def bar
+          1
+        end
+      end
+
+      class Foo
+        include Moo(Int32)
+      end
+
+      Foo.new
+      ),
+      "undefined local variable or method 'bar'"
+  end
+
+  it "solves instance var initializer in metaclass context (#5876)" do
+    assert_type(%(
+      class Foo
+        @x : Int32 = bar
+
+        def self.bar
+          1
+        end
+
+        def x
+          @x
+        end
+      end
+
+      Foo.new.x
+      )) { int32 }
+  end
 end
--- a/spec/compiler/codegen/class_spec.cr
+++ b/spec/compiler/codegen/class_spec.cr
@@ -646,7 +646,7 @@ describe "Code gen: class" do
   it "doesn't crash on instance variable assigned a proc, and never instantiated (#923)" do
     codegen(%(
       class Klass
-        def f(arg)
+        def self.f(arg)
         end
 
         @a : Proc(String, Nil) = ->f(String)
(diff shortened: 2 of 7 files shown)


straight-shoota:
In my opinion it seems logical that an ivar assignment behaves the same whether it is inside an `initialize` method or at class scope. I always figured these initializers as being in some kind of virtual `initialize` which is always called before the actual method. This can be particularly useful to make sure an ivar is always initialized even with subclass overrides of `initialize`. I'm not aware of a specific use case regarding instance methods, though.
I suppose the same sanity-checks could be applied to ivar initializers as inside an `initialize` method.

On the other hand, it's probably not that important and having a consistent rule is fine either way.

I just don't see an issue with `CONST = bar` and `@ivar : Int32 = bar` referring to different `bar` methods. The identifiers `CONST` and `@ivar`  themselves are on different scopes, so it makes perfect sense that their value expression can be differently scoped as well.

Sija:
ivar assignments are done at the class level so why should they be resolved differently?

asterite:
@Sija @straight-shoota What do you mean? There's no such thing in Ruby. When you do `@x = 1` at the class level in Ruby, it assigns an instance variable **of the class**, not of the instance. In Crystal we want a different behavior for that same line.

straight-shoota:
@Sija Even if Ruby was different... In Crystal `@x : Int32` declares a varibale in instance scope, so it would by no means be a surprise to anyone if the second part `= bar` also resolves in instance scope (it doesn't necessarily need be that way, type scope should be ok, too).

Sija:
@asterite @straight-shoota Sorry, my bad! I already forgot that in Ruby `@ivars` can be declared only inside of instance methods... Too much Crystal lately 😅 Still, with the issue at hand `@x = bar` declaration is happening inside of the _class_ scope, so IMHO it intuitively should refer to to class rather than instance (you can always use it like that inside of `initialize`).

straight-shoota:
And you can always reach class scope with `self.class` :stuck_out_tongue:

asterite:
Actually, I tried to see if I can fix a few issues regarding initializers and instance vars... but it's all a big mess, in addition to how generics work or are implemented. I don't feel like fixing this by adding more patches and hacks. As usual, I believe it's better to have a good portion of time to rewrite the compiler from scratch, in a readable, efficient, well thought-out way (though at least I won't do that, ever, but if Crystal ever gets enough money to do it, I'd advise to do it).

RX14:
@asterite I'm sure rewriting or refactoring portions of the compiler will be far more productive then throwing out the entire thing. If you're working on fixing semantics, then you probably dont want to be testing that with freshly written buggy codegen code.

spec/compiler/semantic/instance_var_spec.cr
```diff
@@ -5042,4 +5042,71 @@ describe "Semantic: instance var" do
...
+        @x : Int32 = bar
+
+        def bar
+          1
+        end
+      end
+
+      Foo.new
+      ),
+      "undefined local variable or method 'bar'"
```
RX14: can we detect this and emit a specialized "did you mean" error message? I suspect it wouldn't be easy but worth asking.

asterite: I think it's possible, if a lookup fails, see if there's a method with that name in the class scope... but I don't feel like implement that right now (someone else can try it, though)

asterite:
I don't think there's a big pressure to merge this, or do something regarding this. It can be tackled in the future.

PR #6418, opened 2018-07-20
Give proper error when doing sizeof uninstantiated generic type

### LINKED ISSUE #6415: Calling sizeof(Tuple) caused compiler error
The following line
``puts "#{sizeof(Tuple)}"``

caused the compiler to issue this error:
```
> crystal build --release -o bug bug.cr
BUG: called create_llvm_type for K (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
```
My compiler version is:
```
> crystal --version

LLVM: 4.0.0
```
Operating System:
```
> uname -a
Linux lambda 4.16.0-2-amd64 #1 SMP Debian 4.16.16-2 (2018-06-22) x86_64 GNU/Linux
```

asterite:
Fixes #6415

### DIFF
--- a/spec/compiler/codegen/sizeof_spec.cr
+++ b/spec/compiler/codegen/sizeof_spec.cr
@@ -73,7 +73,7 @@ describe "Code gen: sizeof" do
   end
 
   it "gives error if using instance_sizeof on a generic type without type vars" do
-    assert_error "instance_sizeof(Array)", "can't calculate instance_sizeof of generic class"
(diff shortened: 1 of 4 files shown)


straight-shoota:
Just to make sure, you should probably add a spec for instantiated generic types.

asterite:
@straight-shoota What about union types? Class types? Pointer types? Aliases? There are so many combinations you could test, I don't think it belongs in this PR, which is about fixing a bug.

PR #6419, opened 2018-07-20
Support exception handling on windows

RX14:
This was a tough one, but I learned a lot along the way and I'm glad I accomplished it.

### DIFF
--- a/src/compiler/crystal/codegen/codegen.cr
+++ b/src/compiler/crystal/codegen/codegen.cr
@@ -11,7 +11,6 @@ module Crystal
   MALLOC_NAME        = "__crystal_malloc64"
   MALLOC_ATOMIC_NAME = "__crystal_malloc_atomic64"
   REALLOC_NAME       = "__crystal_realloc64"
-  PERSONALITY_NAME   = "__crystal_personality"
   GET_EXCEPTION_NAME = "__crystal_get_exception"
 
   class Program
@@ -97,6 +96,7 @@ module Crystal
     getter llvm_typer : LLVMTyper
     getter alloca_block : LLVM::BasicBlock
     getter entry_block : LLVM::BasicBlock
+    getter personality_name : String
     property last : LLVM::Value
 
     class LLVMVar
@@ -129,6 +129,7 @@ module Crystal
     @argv : LLVM::Value
     @empty_md_list : LLVM::Value
     @rescue_block : LLVM::BasicBlock?
+    @catch_pad : LLVM::Value?
     @malloc_fun : LLVM::Function?
     @malloc_atomic_fun : LLVM::Function?
     @c_malloc_fun : LLVM::Function?
@@ -155,6 +156,15 @@ module Crystal
       @main_ret_type = node.type? || @program.nil_type
       ret_type = @llvm_typer.llvm_return_type(@main_ret_type)
       @main = @llvm_mod.functions.add(MAIN_NAME, [llvm_context.int32, llvm_context.void_pointer.pointer], ret_type)
+
+      if @program.has_flag? "windows"
+        @personality_name = "__CxxFrameHandler3"
+
+        personality_function = @llvm_mod.functions.add(@personality_name, [] of LLVM::Type, llvm_context.int32, true)
+        @main.personality_function = personality_function
+      else
+        @personality_name = "__crystal_personality"
+      end
 
       emit_main_def_debug_metadata(@main, "??") unless @debug.none?
 
@@ -282,7 +292,7 @@ module Crystal
 
       def visit(node : FunDef)
         case node.name
-        when MALLOC_NAME, MALLOC_ATOMIC_NAME, REALLOC_NAME, RAISE_NAME, PERSONALITY_NAME, GET_EXCEPTION_NAME
+        when MALLOC_NAME, MALLOC_ATOMIC_NAME, REALLOC_NAME, RAISE_NAME, @codegen.personality_name, GET_EXCEPTION_NAME
           @codegen.accept node
         end
         false
@@ -613,6 +623,12 @@ module Crystal
 
       if return_phi = context.return_phi
         return_phi.add @last, node_type
+      elsif catch_pad = @catch_pad
+        ret_block = new_block "catchret_ret"
+        builder.build_catch_ret catch_pad, ret_block
+
+        position_at_end ret_block
+        codegen_return node_type
       else
         codegen_return node_type
       end
@@ -1518,6 +1534,7 @@ module Crystal
       old_fun = context.fun
       old_ensure_exception_handlers = @ensure_exception_handlers
       old_rescue_block = @rescue_block
+      old_catch_pad = @catch_pad
       old_entry_block = @entry_block
       old_alloca_block = @alloca_block
       old_needs_value = @needs_value
@@ -1528,6 +1545,7 @@ module Crystal
 
       @ensure_exception_handlers = nil
       @rescue_block = nil
+      @catch_pad = nil
 
       block_value = yield
 
@@ -1539,6 +1557,7 @@ module Crystal
       @llvm_typer = old_llvm_typer
       @ensure_exception_handlers = old_ensure_exception_handlers
       @rescue_block = old_rescue_block
+      @catch_pad = old_catch_pad
       @entry_block = old_entry_block
       @alloca_block = old_alloca_block
       @needs_value = old_needs_value
--- a/src/compiler/crystal/codegen/crystal_llvm_builder.cr
+++ b/src/compiler/crystal/codegen/crystal_llvm_builder.cr
@@ -41,8 +41,14 @@ module Crystal
       value
     end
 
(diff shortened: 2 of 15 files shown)


asterite:
Wat? You are glad you accomplished it?? You are ****ing amazing, Chris!

bararchy:
My hero!
Great work :tada:

ukd1:
@RX14 💯 💯 💯 💯 💯 💯 💯

RX14:
A lot of this work was @bcardiff and @txe, I stand on the shoulders of giants.

straight-shoota:
I used a compiler with this patch to compile my local branch with Windows library bindings. It didn't break anything, but the specs don't contain any expectations about exceptions, because that didn't work until now. But I'll work on that.

I also compiled [crinja's](https://github.com/straight-shoota/crinja) spec suite. It's one of the largest Crystal projects I know and there are quite a number of exception raised and rescued. And it worked fine :+1: Not all specs passed (which was to be expected), but none fails because of an error in exception handling.

RX14:
@straight-shoota thats fantastic. I wouldn't be surprised if there's still quite a few edge cases in here though. The key hint is usually an "invalid instruction" segfault (llvm adds a `ud2` instruction at the end of catch blocks where it thinks it should be unreachable) or a module validation failure.

src/compiler/crystal/codegen/crystal_llvm_builder.cr
```diff
@@ -41,8 +41,14 @@ module Crystal
       value
     end
 
-    def printf(format, args = [] of LLVM::Value)
-      call @printf, [global_string_pointer(format)] + args
+    def printf(format, args = [] of LLVM::Value, catch_pad = nil)
```
bcardiff: call to printf do really need to be handled with the funclet? Up to know the call was never replaced by an invoke, so it was assumed to never raise AFAIK.

src/compiler/crystal/codegen/exception.cr
```diff
@@ -195,6 +203,153 @@ class Crystal::CodeGenVisitor
     false
   end
 
+  private def codegen_windows_seh(node : ExceptionHandler)
```
bcardiff: 1. I think there should be some comments here explaining how the exception data is handled. I know it comes from documentation of llvm seh and trial an error. But there are some decisions w.r.t. how exception is encoded as `void*` that would help future readers. Also some basic examples of the result in `llvm-ir` would be great for documentation IMO (l tried to do that at for [overflow here](https://github.com/bcardiff/crystal/blob/5ff7481718dd8fe8a6260d30d9647cffd8076d16/src/compiler/crystal/codegen/primitives.cr#L263-L291) in the past).

2. And I am tempted to suggest some comment or extraction of common parts with the codegen_landing_pad. The handling of multiple ensures is shared. Maybe a strategy struct can be used to encapsulate custom logic for pads vs seh, leaving the existing comments `visit(node : ExceptionHandler)`/`codegen_landing_pad` that goes into the details of the handling of the AST. The only real value of this would be to help the reader to understand separate the concepts that are together in this code.

src/raise.cr
```diff
@@ -40,17 +40,26 @@ end
...
+
   # Raises the *exception*.
   #
   # This will set the exception's callstack if it hasn't been already.
   # Re-raising a previously catched exception won't replace the callstack.
   def raise(exception : Exception) : NoReturn
-    exception.inspect_with_backtrace(STDERR)
-    LibC.exit(1)
-  end
+    {% if flag?(:debug_raise) %}
```
bcardiff: I would add a TODO note to remote the whole puts here, or just leave it commented for a while.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -912,4 +914,84 @@ class Crystal::CodeGenVisitor
 
     node.value
   end
+
+  def void_ptr_type_descriptor
```
bcardiff: I'm glad code snippets end up been useful :-). Thanks for tidying them up a bit.

src/compiler/crystal/codegen/llvm_builder_helper.cr
```diff
@@ -76,7 +76,35 @@ module Crystal
       builder.inbounds_gep ptr, index0, index1, name
     end
 
-    delegate ptr2int, int2ptr, and, or, not, call, bit_cast,
+    def call(func, name : String = "")
```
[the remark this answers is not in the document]
RX14: Because there's too many places where codegen generates a call outside of `codegen_call_or_invoke`. I originally had it in `codegen_call_or_invoke` but I didnt feel like duplicating funclet code in the 50 other places that generate calls (think codegen for constants, there's way more though)

src/compiler/crystal/codegen/codegen.cr
```diff
@@ -613,6 +623,12 @@ module Crystal
 
       if return_phi = context.return_phi
         return_phi.add @last, node_type
+      elsif catch_pad = @catch_pad
```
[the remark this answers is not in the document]
RX14: Yeah, I didn't really look at the spec suite since it's impossible to run the spec suite on windows yet because the compiler depends on a bunch of stuff thats not ported.

src/compiler/crystal/codegen/crystal_llvm_builder.cr
RX14: *every* call inside a catch pad needs a funclet. This is not about call/invoke and `@rescue_block`, it's about `@catch_pad` which is a totally different concern and semantic. This is about telling LLVM which method calls are inside the "catch block", and it applies to *every single method call*. `@rescue_block` is about codegenning `invoke` instructions, and it only matters for calls that raise.

src/raise.cr
RX14: I don't see why we can't leave the flag in indefinitely.

src/compiler/crystal/codegen/crystal_llvm_builder.cr
bcardiff: Ok. I also missed this was inside the builder, outside the codegen where the handling of the operand bundle is already handled.

src/raise.cr
bcardiff: If it's going to stay, then it needs to be supported in the other platforms.

RX14: Oh yeah, my bad.

Qwerp-Derp:
Should this also be on 0.26.0?

RX14:
Pushed an update which merges the old and new exception handling code and improves the comments.

src/compiler/crystal/codegen/exception.cr
```diff
@@ -4,197 +4,222 @@ class Crystal::CodeGenVisitor
...
+
+    # This is the block which is entered when the body raises an exception
     rescue_block = new_block "rescue"
 
     node_rescues = node.rescues
     node_ensure = node.ensure
     node_else = node.else
     rescue_ensure_block = nil
 
-    # Here we transform this:
```
bcardiff: I would say that this whole comment section should stay. Your additions inlines are great, but keeping this sections with numbers and later references would be ideal. They facilitate understanding the later code a lot. 

Other than that LGTM.

RX14: I don't like that block comment because it's very incomplete information, and once you distill it into it's core it's just what I've written below.

Once I start thinking how to improve the comment, I just end up with repeating what i've written below

bcardiff: The *once you distill* time is decreased a lot with a nice introduction like the one that is removed. I think that the intricate nature of the code transformation deserves that kind of introduction. Just restoring those line are enough.

RX14:
Added a comment @bcardiff

src/compiler/crystal/codegen/exception.cr
```diff
@@ -4,197 +4,278 @@ class Crystal::CodeGenVisitor
...
-    node_ensure = node.ensure
-    node_else = node.else
-    rescue_ensure_block = nil
-
-    # Here we transform this:
+    # In this codegen, we assume that LLVM only provides us with a basic try/catch abstraction with no
+    # type restrictions on the exception caught. The basic strategy is to codegen this
     #
-    # ```
+    # ```cr
```
Sija: `cr` here is invalid (that's gfm), it better be empty (like it was before)

straight-shoota: It doesn't hurt, though.

Sija: It doesn't help either.

RX14: who cares

Sija: Yeah, and humans know what `cr` means...

bcardiff: I agree with @Sija. It's better without the `cr`. But we can clean up that later (together with one more ```crystal that is [somewhere in the code](https://github.com/crystal-lang/crystal/blob/master/src/time.cr#L23)...)

PR #5229, opened 2017-11-01
Improved search feature for API docs

straight-shoota:
This PR greatly improves client-side search on generated HTML pages of the API docs.
It uses the program description available as JSON data from #4746 and implements a custom search algorithm in JavaScript.
The search results include types, constants, methods, macros matching the query string in one or more of the fields name, namespace, description, arguments or type.
A custom ranking algorithm sorts results based on matching quality and relevance.

The search-algorithm is actually quite simple, it uses no search index or advanced information retrieval functions. It simply traverses through all the types and tries to find matching features. This is really just fast enough.
I considered using something like lunr.js but this would have been complicated to setup to create a domain-specific search experience. With the custom algorithm it is very easy to fine tune matching and ranking behaviour to allow developers to find what they need as fast as possible.

This approach has some drawbacks with regards to fulltext search capabilities (especially concerning docstrings), but this can certainly be improved upon.

## Search features:
* prefix filter: `#` matches instance methods, `.` matches class methods and macros, `:`/`::` matches namespace
* search is generally case-insensitive but results with case-equality areranked higher.
* search results list can be navigated by key (when in focus):
  * `i`, `c`, `ArrowUp`: move upwards
  * `j`, `h`, `ArrowDown`: move downwards
  * `Enter`: load result
  * `Escape`: cancel search
  * arrows, `Enter` and `Escape` also work when the search-input is in focus
* support for URL fragment-based search query: `/#q=mysearch` will initialize a search at page load
* page-wide shortkeys to start search: `s` or `/`

## Preview
[Online Preview](https://straight-shoota.github.io/crystal/docs-search-only/)

This is intended to be alongside #4755 which makes it look even better ([Online Preview](https://straight-shoota.github.io/crystal/doc-search/)) though it can be merged separately.

Closes #1792

PR #6420, opened 2018-07-20
Fix docs navigator not scrolling into selected element on page load

soanvig:
According to @straight-shoota, this function worked before #5229.

I've created my own implementation of that function. It identifies open item by `.current` class, and selects last of it (to handle case of nested items).

Used function `.scrollIntoView()` browser compatibility: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

Function was tested locally and it works, but one may test by himself.


straight-shoota:
You can take a look at https://crystal-lang.org/api/0.24.1/ to see it working before.

It didn't actually scroll the current element to top but preserved the previous viewport. I think that's better, but it would need to work with the search results list which has a different view port.
The proposed implementation is good, though, and certainly an improvement.

soanvig:
So, is this going to be merged or not? Not sure about the process.

Issue #5155, opened 2017-10-20
[RFC] Remove recursive aliases


asterite:
Recursive aliases were introduced when we were implementing JSON parsing. We needed a type that could be `Nil`, `Bool`, `Int64`, `Float64`, `String` or an array of those same things, recursively, or a hash with values of those same things, recursively. So we decided that having a type that could reference itself in its definition was the way to go.

That works, but it has two disadvantages:
1. We can't define methods on this new type, because we can't define methods on aliases
2. **The language already has a way to express recursive types!**

For example, the current definition of `JSON::Type` is:

```crystal
module JSON
  alias Type = Nil | Bool | Int64 | Float64 | String | Array(Type) | Hash(String, Type)
end
```

We can express that same thing using a struct:

```crystal
module JSON
  struct Type
    @value : Nil | Bool | Int64 | Float64 | String | Array(Type) | Hash(String, Type)

    def initialize(@value)
    end
  end
end
```

Or simpler using the `record` macro:

```crystal
module JSON
  record Type, value : Nil | Bool | Int64 | Float64 | String | Array(Type) | Hash(String, Type)
end
```

In terms of memory representation and performance it's **exactly the same**: a recursive alias is a union, and as such it's represented as a struct.

Now, when working with a recursive alias such as `JSON::Type` we can simply cast a value to an integer using `json.as(Int64)`. With the struct approach we have to do `json.value.as(Int64)`. But since we don't like those casts, we have `JSON::Any` in the standard library **which wraps the `JSON::Type` alias**. Sounds familiar? We could have simply never used a recursive alias in the first place! Just make `JSON.parse` return a `JSON::Type` that is already a struct whose value is the union of possible types (recursively). Want to get the raw value? Call `value`. Similar to `JSON::Any`.

That said, I believe we can safely remove recursive aliases from the language. They are not an essential feature, and in fact using a struct is better because we can add methods to them. And removing them will also simplify the compiler's code.

asterite:
And in fact, this also solves another problem that's a source of confusion and complaint with `JSON::Any`. Right now we have: `JSON::Any#to_a` which assumes the raw value is an Array and returns it as such. But this Array is of `JSON::Type`, not of `JSON::Any`, so we can't invoke `to_a` on them.

Using the struct approach, the elements are themselves `JSON::Type` which will provide these methods. So much simpler and powerful!

paulcsmith:
Nice thinking. This looks great

[...]

PR #6423, opened 2018-07-21
Create an abstract Any for JSON/YAML

j8r:
Merges duplications between `YAML::Any` and `JSON::Any` to a common `Any` abstract struct, which is inherited.

External shards implementing configurations formats will also be able to inherit from it.

Partially related to https://github.com/crystal-lang/crystal/issues/5155

Note: this PR is a background work to implement this POC https://gist.github.com/j8r/2820157756ce55182c85667cb0456800 in a future PR - the ability to deal with dynamic keys:
```crystal
js = JSON.parse %({"a": {"b": "c"}, "e": [{"o": "p"}, 12, 2]})

p js[["a", "r"]]?  #=> nil
puts js[["e", 0, "o"]] = JSON::Any.new "test"
puts js        #=> {"a" => {"b" => "c"}, "e" => [{"o" => "test"}, 12_i64, 2_i64]}
js.delete ["e",0, "o"]
puts js        #=> {"a" => {"b" => "c"}, "e" => [{}, 12_i64, 2_i64]}
```

### DIFF
--- a/spec/std/json/any_spec.cr
+++ b/spec/std/json/any_spec.cr
@@ -26,11 +26,8 @@ describe JSON::Any do
 
     it "gets float" do
       JSON.parse("123.45").as_f.should eq(123.45)
-      JSON.parse("123.45").as_f32.should eq(123.45_f32)
       JSON.parse("123.45").as_f?.should eq(123.45)
-      JSON.parse("123.45").as_f32?.should eq(123.45_f32)
       JSON.parse("true").as_f?.should be_nil
-      JSON.parse("true").as_f32?.should be_nil
     end
 
     it "gets string" do
--- a/src/any.cr
+++ b/src/any.cr
@@ -1,0 +1,160 @@
+abstract struct Any
+  # Assumes the underlying value is an `Array` or `Hash` and returns its size.
+  # Raises if the underlying value is not an `Array` or `Hash`.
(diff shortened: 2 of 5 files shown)


asterite:
@j8r Sorry, but these PRs should be discussed before implemented.

I personally agree with one of Go's proverb:

> A little copying is better than a little dependency

Just because the code is similar, it shouldn't involve a dependency.

So a strong 👎 from me here.

j8r:
Note that the API and implementations between the `Any`s aren't the same:
no `YAML::Any#as_bool`, `YAML::Any#as_f32`  and `JSON::Any#as_bytes` 

In `JSON::Any`, there are `as_i if @raw.is_a?(Int)` and in `YAML::Any` `@raw.as?(Int64).try &.to_i`

ysbaddaden:
I think we need more analysis and experiments on (de)serialization before conducting such changes. For example I'd love some attempts at a Serde like solution (from Rust).

j8r:
I'll create a shard because for now dealing with dynamic mappings isn't possible in the stdlib

ysbaddaden:
Yes, please experiment and return back your results, both successes and shortcomings :)

j8r:
For information: a reposiotry is created, https://github.com/j8r/dynany

PR #6425, opened 2018-07-21
Compiler: fix overload order for unnamed splat vs double splat

### LINKED ISSUE #5328: Unnamed splat argument disturbs overloading based on named arguments
I expect a result of "1" from the call at the bottom of this code.

```crystal
# Removing * or replacing it with *args changes the result to 1.
def foo(*, x : Nil)
  1
end

# Commenting this method fixes changes the result to 1.
def foo(**options)
  :symbol
end

foo(x: nil) # => :symbol
```

This happens both with Crystal 0.23.1 and Crystal 0.24.0-2 on Ubuntu 16.04 LTS.

asterite:
Fixes #5328


bew:
Why close?

asterite:
Because CI failed and I don't have time to understand why (my fix didn't work)

PR #6426, opened 2018-07-22
Fix: fork and signal child handlers

### LINKED ISSUE #6416: Process.run() not returning in forked process
Hi,

I had an error in the following code:

```
stdout = IO::Memory.new
stderr = IO::Memory.new
Process.run("echo", args: {"hello"}, output: stdout, error: stderr)
puts stdout.to_s
puts stderr.to_s

while true
  puts "======================="
  puts "LOOP"
  puts "======================="
  fork_new_process()
  sleep 10
end

def fork_new_process()
  process = Process.fork()
  if process
    # Parent process continues
    return
  else
    # Child process
    stdout = IO::Memory.new
    stderr = IO::Memory.new
    puts "START"
    Process.run("echo", args: {"hello"}, output: stdout, error: stderr)
    puts "END"
    puts stdout.to_s
    puts stderr.to_s
    exit 0 # Exit on end
  end
end
```
The first time 'hello' is printed. In the forked process, 'hello' is never printed and the Process.run() doesn't even return since 'END' is never printed

```

LLVM: 4.0.0
```

ysbaddaden:
Fixes an issue where a parent process could receive signals directed
to the child or vice-versa.

fixes #6416


forksaber:
I am not sure if this should be handled as part of this PR, but I think there still exists a subtle race condition in the signal handler implementation. Consider the following order of events:

```markdown
Parent  : fork()
Child   : spawns
Child   : receives a signal
Child   : writes the signal to parent's pipe
Child   : Signal.after_fork is called
Parent  : processes the signal meant for child (**BUG**)
```
I haven't actually tried writing down a fix but something like this should work:
```crystal
#Pseudo code

# block all the signals for the current thread
pthread_sigmask(block all signals) 
pid =LibC.fork
if pid > 0
  #Parent
  pthread_sigmask(unblock all signals)
else
 #child
  if run_hooks       # (Process.fork)
    Signal.after_fork
    pthread_sigmask(unblock all signals)
    run_remaining_hooks
  else                    #(Process.exec)
    # reset all signal handlers  to SIG_DFL or  call Signal.after_fork  
    (signals.each { |signal| signal.reset } ) OR (Signal.after_fork) 
    pthread_sigmask(unblock all signals)
    execvp
  end
end
   
```

References:
https://www.redhat.com/archives/libvir-list/2008-August/msg00303.html
https://github.com/libvirt/libvirt/blob/e83da1990cfb75ae9a83559b6b702bb9fc4bc61e/src/util/vircommand.c#L297

ysbaddaden:
You're right again!

I added a commit to disable signals then reenable them after fork (parent and child). Now we should be fine :crossed_fingers:

I chose to continue to not reset signal handlers in the child process. Either developers don't care about signals and we'll keep the default handlers (i.e. SIGSEGV, SIGCHLD) or she cares about them and she's responsible for resetting them if needed.

bararchy:
@RX14 or @jhass 
Two approves here, LGTM?

Sija:
@bararchy Hold your horses, let's wait for conclusions regarding newest @forksaber's comment.

bararchy:
@Sija sure, I was under the impression this has already been reviewed.

Sija:
Well, it was. Before that comment ;)

src/process.cr
```diff
@@ -117,14 +117,31 @@ class Process
...
+
+    case pid = LibC.fork
     when 0
+      # child:
       pid = nil
-      Process.after_fork_child_callbacks.each(&.call) if run_hooks
+      if run_hooks
+        Process.after_fork_child_callbacks.each(&.call)
+        LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
+      end
```
ysbaddaden: In chose to not reenable signals in a fork/exec situation. I just leave the handlers and mask in place, and let `execve(2)` override that for the new process.

forksaber: I think if you don't reenable the signals, they will remain blocked across the exec. So, it is necessary to reenable the signals.

```markdown
**The new process shall inherit at least the following attributes from the calling process image**:

[XSI] [Option Start] Nice value (see nice()) [Option End]

[XSI] [Option Start] semadj values (see semop()) [Option End]

Process ID

Parent process ID

Process group ID

Session membership

Real user ID

Real group ID

Supplementary group IDs

Time left until an alarm clock signal (see alarm())

Current working directory

Root directory

File mode creation mask (see umask())

[XSI] [Option Start] File size limit (see ulimit()) [Option End]

**Process signal mask (see sigprocmask())**

Pending signal (see sigpending())

tms_utime, tms_stime, tms_cutime, and tms_cstime (see times())

[XSI] [Option Start] Resource limits [Option End]

[XSI] [Option Start] Controlling terminal [Option End]

[XSI] [Option Start] Interval timers [Option End]

The initial thread of the new process shall inherit at least the following attributes from the calling thread:

Signal mask (see sigprocmask() and pthread_sigmask())

Pending signals (see sigpending())
```

Reference:
http://pubs.opengroup.org/onlinepubs/009695399/functions/exec.html

ysbaddaden:
Sigh... Okay.

forksaber:
Sorry for being such an annoying person/reviewer :(
I can't help but point out that the race condition(fork/exec scenario) has shifted from
```crystal
# run_hooks: false
pid = LibC.fork
case pid
when 0
  # child
  # RACE here
  exec
end
```
to
```crystal
# run_hooks: false
block_signals
pid = LibC.fork
case pid
when 0
  # child
  unblock_signals
  # RACE here : since parent's pipe exists, any signal 
  # delivered here will end up in parent process
  exec
end

```
A crucial step is missing to fix this race condition is to reset the signal handlers to SIG_DFL before unblocking the signals:
```diff
case pid
when 0
  # child
+  Crystal::Signal.reset_signal_handlers!
  unblock_signals
  # RACE here : since parent's pipe exists, any signal 
  # delivered here will end up in parent process
  exec
end

```
An attempt to fix this:
```crystal
module Crystal::Signal
  def self.mutex
    @@mutex
  end

  # safe to invoke only in post-fork/pre-exec state
  def self.reset_signal_handlers!
    @@handlers.keys.each { |sig| sig.reset }
  end
end
```
```diff
+  # lock signal mutex before forking to ensure @@handlers is in
+  # a consistent state
+  mutex = Crystal::Signal.mutex
+  mutex.lock
+
   case pid = LibC.fork
   when 0
     # child:
     pid = nil
     if run_hooks
       Process.after_fork_child_callbacks.each(&.call)
+      # unlock mutex once signal pipes are reinitialized
+      mutex.unlock
       LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
     else
+      # no need to unlock mutex as we won't be registering
+      # any custom signal handlers before exec.
+      # Also unlocking mutex not safe here as scheduler is not reinitialized
+      Crystal::Signal.reset_signal_handlers!
       # sigmask is inherited: must reset it
       LibC.sigemptyset(pointerof(newmask))
       LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(newmask), nil)
     end
   when -1
+    # unlock mutex immediately after fork in parent
+    mutex.unlock
     # error:
     errno = Errno.value
     LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
     raise Errno.new("fork", errno)
   else
+    # unlock mutex immediately after fork in parent
+    mutex.unlock
     # parent:
     LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
   end
```

RX14:
@forksaber honestly we really appreciate your comments! It's much nicer to sort this out now instead of debugging an intermittent race condition in production years later.

ysbaddaden:
Yes, you're right again! Sigh me, too focused on the child state than on the child -> parent leak.

ysbaddaden:
Now I hope it's fine :crossed_fingers:

forksaber:
The patch LGTM :+1: when there exists only a single thread (which is currently the case). 

However, consider this multi-threaded case
```
Thread-1 : locks Crystal::Signal mutex

Thread-1 : is in the middle of modifying @@handlers hash....execution state could be 
           inside any method of the Hash implementation ( hash's internal state might 
           be inconsistent depending on the implementation)

Thread-2 : fork

Child : now accesses said inconsistent @@handlers hash possibly 
        causing segfault or undefined behaviour

Thread-1 : finishes modifying @@handlers
```
So, ```mutex.lock``` is needed before fork to ensure that no other thread modifies the ```@@handlers``` hash while the current thread forks.

I can understand that you may not want to include the ```mutex.lock``` since Crystal is currently single threaded. 
An alternative is to add some **TODO** comment related to this, so we can revisit this once thread-support lands.

ysbaddaden:
Let's reset all signals and be done with it now.

forksaber:
@ysbaddaden Agreed

forksaber:
**Assumption**: An average crystal program defines 3 custom signal handlers

To make this decision easier, I am adding a benchmark that compares the following approaches:

- **hash**: `@@handlers.keys` is used to find signals and reset them
- **reset all**: reset all signal handlers irrespective of whether they have non default values
- **array of atomic flags**:  used alongside `@@handlers` hash to register and query handler state change

```crystal
require "benchmark"
class AtomicFlag
  @flag = Atomic(UInt8).new(0_u8)

  def get : Bool
    @flag.get == 1_u8
  end

  def set(val : Bool)
    if val
      @flag.set(1_u8)
    else
      @flag.set(0_u8)
    end
  end
end

arr = StaticArray(AtomicFlag, 32).new { AtomicFlag.new }
handlers = Hash(Int32, Proc(Signal, Nil)).new

n = 3
(1..n).each do |i|
  arr[i].set true
  handlers[i] = ->(sig : Signal) { nil }
end

Benchmark.ips do |x|
  x.report("reset changed - hash  ") do
    handlers.each_key do |i|
      LibC.signal(i, LibC::SIG_DFL)
    end
  end

  x.report("reset changed - atomic") do
    (1..31).each do |i|
      flag = arr[i-1].get
      LibC.signal(i, LibC::SIG_DFL) if flag
    end
  end

  x.report("reset all             ") do
    (1..31).each { |i| LibC.signal(i, LibC::SIG_DFL) }
  end
end
```

Output
```
reset changed - hash     1.47M (678.12ns) (± 0.42%)  0 B/op        fastest
reset changed - atomic   1.42M (705.08ns) (± 0.29%)  0 B/op   1.04× slower
reset all              144.33k (  6.93µs) (± 0.42%)  0 B/op  10.22× slower
```
I am fine with either solution....using atomics to reset select signals or resetting all the signals.

ysbaddaden:
Thanks for the benchmark!

I maintain a `sigset_t`, because signals are tricky: they usually range 1-32, but there ain't no guarantee about that. In addition the POSIX functions are fast enough. With this patch it's only 1.07× slower than the unsafe `@@handlers.each_key`, so `Process.run` won't see a significant performance loss.

The set is only mutated within a mutex so mutation is thread safe. It's also modified *before* the actual trap and *after* the actual reset. I assume it's fine to use without the mutex in fork/exec. It's just a fixed set of longs, and either a thread is adding a signal to the set and we don't care (the signal isn't trapped, yet) or it's being removed and we'll just have a useless reset.

All cases look fine to me. I hope you agree @forksaber :crossed_fingers:

ysbaddaden:
I think the atomic is used in Go to avoid a mutex. Thread A could lock the mutex but be paused by the kernel, thus potentialy blocking other threads until the kernel resumes Thread A.

An atomic and custom sigset, based on NSIG is a good idea, yeah.

RX14:
I can't believe that OOO execution is an issue here though.

forksaber:
### **Firstly, I think the patch is good enough and we should merge it**.


Now, just for the sake of closure, I did manage to reproduce a data race....but the cause is **not** `Out of order execution (which is completely safe in this scenario)`. Instead, i think its probably caused by some race in kernel's bookkeeping while doing a `fork`.
And no, atomics don't fix this.

### How signals handlers are registered/unregistered?
A flag is set to `true` before registering a custom signal handler and set to `false` after removing the signal handler. 
```crystal
# register
sigset.add(sig)
signal(sig, handler)

#remove
signal(sig, SIG_DFL)
sigset.del(sig)
```

### Expectation
If there's no inconsistency/race/reordering, we should never be able to observe the flag being `false` and a custom signal handler being present.


But such a race/inconsistency exists:

### Race
```c
// race.c

#include <stdio.h>
#include <signal.h>
#include <stdlib.h>
#include <pthread.h>
#include <unistd.h>
#include <sys/wait.h>

int flag;

void custom_handler(int signal)
{
    signal = signal + 1;
}

/* Keeps switching signal(21)'s handler
 * between SIG_DFL and `custom_handler` in a loop
 * 
 *
 * All possible values of (flag, sa_handler) if execution
 * goes as expected.
 * (flag, sa_handler)
 * (0, 0)
 * (1, 0)
 * (1, <some pointer val>)
 * (1, 0)
 * (0, 0)
*/
static void *sigaction_loop(void *args)
{
    struct sigaction custom, dfl;
    custom.sa_handler = &custom_handler;
    dfl.sa_handler = SIG_DFL;

        //(flag, sa_handler)
    while(1) {
        // 0, 0
        flag = 1;
        // 1, 0
        sigaction(21, &custom, NULL);
        // 1, <some pointer val>
        sigaction(21, &dfl, NULL);
        // 1, 0
        flag = 0;
        // 0, 0
    }
}

static void check_flag_handler_consistency()
{
    struct sigaction sa;
    // fetches current sigaction struct
    sigaction(21, NULL, &sa);
    if (flag == 0 && (&sa)->sa_handler != 0)
        // unexpected (flag, sa_handler) value
        // (0, <some non zero pointer val>)
        printf("unexpected flag:%d val:%d\n", flag, (&sa)->sa_handler);
}

static void fork_loop()
{
    int pid;
    int status;
    for (int i = 0; i < 20000; i++) {
        pid = fork();
        if (pid == 0) {
            check_flag_handler_consistency();
            exit(0);
        } else if (pid < 0) {
            return;
        } else {
            waitpid(pid, &status, 0);
        }
    }
}

int main()
{
    pthread_t thread;
    pthread_create(&thread, NULL, sigaction_loop, NULL);
    fork_loop();
    return 0;
}
```

Build and run:
```
gcc -lpthread -o race race.c
./race
```
Output:
```
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
unexpected flag:0 val:4196055
```

### Fix
Have a mutex lock around `signal(sig, handler)` and `fork` so that no kernel structure gets modified while we are forking.
```c
pthread_mutex_t m;

pthread_mutex_lock(&m);
fork();
pthread_mutex_unlock(&m);

pthread_mutex_lock(&m);
signal(sig, &handler);   // or sigaction(sig, &sa, NULL);
pthread_mutex_unlock(&m)
```
### Non-Fixes
- Atomics: Atomics won't prevent this as kernel struct can still get modified by other threads while 
forking

**PS**: This race also exists in `golang`. I had to write `cgo` and patch `go` to actually observe this.

ysbaddaden:
It condos my understanding: Go is using an atomic to avoid a mutex that would block.

The race is quite edge case: it requires that a thread would modify a signal while another is exactly calling fork.

I naively assume that a program does setup a few signals at startup and won't change them much afterwards.

Maybe we can just close the pipe in fork/exec and add a "write byte unless closed" in the signal handler. That would avoid passing child signal to parent and exec will reset any signals we missed to SIG_DFL anyway.

forksaber:
Nice idea @ysbaddaden 
Building on your idea....something like this should work great.
```crystal
module Crystal::Signal
  @@after_fork_before_exec = false

  def self.after_fork_before_exec
    @@after_fork_before_exec = true
    ::Signal.each do |signal|
      LibC.signal(signal, LibC::SIG_DFL) if signal.set?
    end
  end

  def self.trap(signal, handler) : Nil
    @@mutex.synchronize do
      unless @@handlers[signal]?
        signal.set_add
####
        LibC.signal(signal.value, ->(value : Int32) {
          if @@after_fork_before_exec
             # signal handler was not reset due to the above race...lets reset it
             LibC.signal(value, LibC::SIG_DFL)
             # re raise the signal so that its handled with SIG_DFL semantics
             LibC.raise(value)
          else
             writer.write_bytes(value)
          end
        })
####
      end
      @@handlers[signal] = handler
    end
  end

end
```

We can even double down on this idea and forget about maintaining a `sigset`. Then, every signal will be reset dynamically if and when its encountered. 😁 

Thanks @bcardiff, I appreciate it 😃

bcardiff:
@ysbaddaden given that we should release soon, maybe either we merge as is this pr and defer future work, or we release 0.26 without this. Your call.

ysbaddaden:
I pushed a commit to close the pipe anyway, and to check if it's closed when writing the signal. I'm fixing the conflict.

ghastmask:
Hi it sounds like you guys already got everything figured out but I wanted to point you to the facebook folly Subprocess c++ implementation which is pretty well documented in case you hadn't seen it:

https://github.com/facebook/folly/blob/master/folly/Subprocess.cpp

bcardiff:
I would rather include this in 0.27. It is a bit sensible and I would prefer to minimize the risk of needing a 0.26.2 for this.

bcardiff:
@ysbaddaden would you rebase on master so we merge this without squashing?

ysbaddaden:
I just did. There was no conflicts, thought.

PR #6074, opened 2018-05-08
Explosive birth of automatic casts!

asterite:
## What

This PR lets the compiler match number literals against type restrictions that don't exactly match the expected type, but since the literal's value fits the type, it is automatically cast.

Additionally, a similar behaviour is provided to specify enum values by using a symbol.

All of this works as long as the call doesn't become ambiguous. The implementation was a bit tricky, but finally, and amazingly, it worked.

For a quick understanding of the first part (numbers), you can read what's being requested in #2995. With this PR, **all** of those snippets now compile.

But I'll write a few snippets here.

#### Numbers

```crystal
def foo(x : Int8)
end

foo(1)    # OK, same as passing 1_u8, because 1 fits Int8
foo(1234) # Error
```

```crystal
def foo(x : Float64)
end

foo(1) # OK, same as passing 1.0
```
#### Symbols -> Enums

```crystal
enum Color
  Red
  Green
  Blue
end

def paint(color : Color)
end

paint(:red) # OK, same as passing Color::Red
paint(:yellow) # Error
```
### Instance variables

```crystal
class Foo
  @x : Int8

  def initialize
    @x = 0 # OK, because @x's type is specified, and 0 can pass as 0_u8
  end 
end
```

### Class variables

```crystal
class Foo
  @@x : Int8 = 0 # OK
end
```

### Local variables

```crystal
x : Int8
x = 0 # OK
```

### Method default values, with restriction

```crystal
enum Color
  Red
  Green
  Blue
end

def paint(color : Color = :red) # OK
end

paint # OK
```

### Combined

```crystal
record Vector, x : Float64, y : Float64, z : Float64

# Works because record will use those restrictions in initialize
Vector.new(1, 2, 3)
```

#### File

```crystal
file = File.open(__FILE__)
file.seek(-10, :end) # Same as passing IO::Seek::End
```

#### Process

```crystal
Process.run("ls", output: :inherit) # Same as passing Process::Redirect::Inherit
```

#### String

```crystal
# Well, this one doesn't work, it's missing the type restriction in the method, but it's super easy to fix
"Súper".upcase(:ascii) # => "SúPER"
```

### What if it's ambiguous?

```crystal
def foo(x : Int8)
end

def foo(x : Int16)
end

foo(1) # Error: ambiguous call matches both Int8 and Int16
```

### When it doesn't work

```crystal
def foo(x : Int8)
end

foo(rand < 0.5 ? 1 : 2) # Error
```

[...]

PR #6427, opened 2018-07-22
Add :default to colorize and document ColorRGB, Color256

r00ster91:
I think it would be good when you can just write:
```crystal
"hello".colorize(:default)
```
instead of:
```crystal
"hello".colorize(Colorize::ColorANSI::Default)
```

### DIFF
--- a/src/colorize.cr
+++ b/src/colorize.cr
@@ -54,6 +54,12 @@
 # "foo".colorize(:red).toggle(false).toggle(true) # => "foo" in red
 # ```
 #
+# The color `:default` will just leave the object as it is (But it's an `Colorize::Object(String)` then).
+# That's handy in for example conditions:
+# ```
+# "foo".colorize(some_bool ? :green : :default)
+# ```
+#
 # Available colors are:
 # ```
 # :default
--- a/src/colorize.cr
+++ b/src/colorize.cr
@@ -50,7 +50,7 @@
 # With the `toggle` method you can temporarily disable adding the escape codes.
 # Settings of the instance are preserved however and can be turned back on later:
 # ```
-# "foo".colorize(:red).toggle(false) # => "foo" without color
+# "foo".colorize(:red).toggle(false)              # => "foo" without color
 # "foo".colorize(:red).toggle(false).toggle(true) # => "foo" in red
 # ```
 #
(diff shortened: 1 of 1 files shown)


Sija:
Would be nice to document what actually **is** the default color? Or even remove it altogether since without context IMHO having _default_ doesn't make much sense...

straight-shoota:
With #6074 `"hello".colorize(:default)` should just work if the argument type for `Object#colorize` is restricted to `ColorANSI`.

r00ster91:
@Sija having `:default` is definitely handy sometimes. Especially in conditions like:
```crystal
puts "hello".colorize(some_bool ? :green : :default)
```
so `:default` prevents some complexity. If it wouldn't exists you would have to do something like:
```crystal
if some_bool
  puts "hello".colorize(:green)
else
  puts "hello"
end
```
The `:default` color just leaves the string as it is (but it's an `Colorize::Object(String)` object then).

Sija:
@Qwerp-Derp problem with #6074 is that it allows only passing symbol literals, it ain't work if you use variables for instance.

Qwerp-Derp:
@Sija Why would you ever use anything other than a symbol literal in this case? If you were to use a variable, why not just use the full name of the enum?

```crystal
a = Colorize::ColorANSI::Default
"foo".colorize(a)
```

I agree something like `"foo".colorize(Colorize::ColorANSI::Default)` is ugly, but only because the line is unnecessarily long. If you split it up into a variable I personally think it's fine, but that's just my 2c on the issue.

asterite:
Yeah, using enums now should be the way to go.

DestyNova:
I think Enums would be much better, because at the moment it's very easy to get runtime errors that should have been caught at compile-time:

```
$ crystal build hello.cr
$ ./hello
Unhandled exception: Unknown color: purple (ArgumentError)
  from /usr/share/crystal/src/colorize.cr:244:5 in 'fore'
  from /usr/share/crystal/src/colorize.cr:118:5 in 'colorize'
  from hello.cr:3:8 in '__crystal_main'
```

MakeNowJust:
I forgot to add `default` to `COLORS` when added `ColorANSI`. So it is just bug.

This PR looks totally good and it should be merged ASAP. Using enum instead symbol is another and wider topic than this PR. Should talk in new issue.

@r00ster91 Thank you! And sorry for missing this.

PR #6428, opened 2018-07-22
Fix Atomic#swap with reference types

Exilor:
Previously the compiler crashed with "Module validation failed: atomicrmw operand must have integer type!".
(followed the example of ```#compare_and_set```)

### DIFF
--- a/src/atomic.cr
+++ b/src/atomic.cr
@@ -157,7 +157,12 @@ struct Atomic(T)
   # atomic.get      # => 10
   # ```
   def swap(value : T)
-    Ops.atomicrmw(:xchg, pointerof(@value), value, :sequentially_consistent, false)
+    {% if T.union? && T.union_types.all? { |t| t == Nil || t < Reference } || T < Reference %}
+      address = Ops.atomicrmw(:xchg, pointerof(@value).as(LibC::SizeT*), LibC::SizeT.new(value.as(T).object_id), :sequentially_consistent, false)
+      Pointer(T).new(address).as(T)
+    {% else %}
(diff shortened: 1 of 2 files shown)


RX14:
I guess they should check for references and error out with a nicer error message (use macro `raise`).

src/atomic.cr
```diff
@@ -157,7 +157,12 @@ struct Atomic(T)
   # atomic.get      # => 10
   # ```
   def swap(value : T)
-    Ops.atomicrmw(:xchg, pointerof(@value), value, :sequentially_consistent, false)
+    {% if T.union? && T.union_types.all? { |t| t == Nil || t < Reference } || T < Reference %}
```
RX14: Look at the `compare_and_set` method above for how to implement this. It uses different cases for nillable references and references. Currently this wont work for nilable references.

Exilor: It does work and I added a test specifically for that case. The difference from ```compare_and_set``` has to do with the  return value of the method, no the type of ```T```.

Exilor:
They probably should but that's outside the scope of this PR.

src/atomic.cr
RX14: Oh, my bad.

It looks to me then like `compare_and_set` should work without the special case for nilables as `Pointer(Void).null.as(String?)` == `nil`.

Exilor: I tried what you suggested and it appears to work but perhaps whoever implemented it knows of some edge case we do not.
Anyway, that is a matter for another issue/PR.

RX14: @Exilor yup, which is why I approved this :)

PR #6429, opened 2018-07-22
JSON: Handle non-standard UTF-8 string escapes

hinrik:
I came across `\u` escapes like this in Facebook's JSON-exported user
data. Here's another example of these in the wild:

  http://seclists.org/wireshark/2018/Jul/36

This issue is present in the standard Ruby/Python/Go/PHP parsers too.
The only parser I found that handles it is Perl's JSON module:

  https://github.com/makamaka/JSON/blob/master/lib/JSON/backportPP.pm#L811

### DIFF
--- a/src/json/lexer.cr
+++ b/src/json/lexer.cr
@@ -130,7 +130,7 @@ abstract class JSON::Lexer
       when '\0'
         raise "Unterminated string"
       when '\\'
-        consume_string_escape_sequence
+        consume_string_escape_sequence(skip: true)
       when '"'
         next_char
         break
@@ -154,7 +154,7 @@ abstract class JSON::Lexer
       when '\0'
         raise "Unterminated string"
       when '\\'
-        @buffer << consume_string_escape_sequence
+        consume_string_escape_sequence
       when '"'
         next_char
         break
@@ -173,20 +173,20 @@ abstract class JSON::Lexer
     end
   end
 
-  private def consume_string_escape_sequence
+  private def consume_string_escape_sequence(skip = false)
     case char = next_char
     when '\\', '"', '/'
-      char
+      @buffer << char if !skip
     when 'b'
-      '\b'
+      @buffer << '\b' if !skip
     when 'f'
-      '\f'
+      @buffer << '\f' if !skip
     when 'n'
-      '\n'
+      @buffer << '\n' if !skip
     when 'r'
-      '\r'
+      @buffer << '\r' if !skip
     when 't'
-      '\t'
+      @buffer << '\t' if !skip
     when 'u'
       hexnum1 = read_hex_number
       if hexnum1 > 0xD800 && hexnum1 < 0xDBFF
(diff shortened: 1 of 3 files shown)


asterite:
Maybe if the issue is present in so many languages, and it's a non-standard thing, it's not an issue? Where in the spec of JSON is this?

hinrik:
> Where in the spec of JSON is this?

The standard only specifies UTF-16 surrogate pairs: https://tools.ietf.org/html/rfc7159#section-7.

>Maybe if the issue is present in so many languages, and it's a non-standard thing, it's not an issue?

It was an issue for me as least, as I had to use Perl instead of Crystal for a text processing task. :)

Sure, this kind of output is non-standard and rare, but a major website is outputting it to its users, and it's possible to work around it as Perl has done, so why not do so? I'd say the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) should apply.

src/json/lexer.cr
```diff
@@ -154,7 +154,7 @@ abstract class JSON::Lexer
       when '\0'
         raise "Unterminated string"
       when '\\'
-        @buffer << consume_string_escape_sequence
+        consume_string_escape_sequence
```
asterite: Why this isn't using skip: true?

asterite:
Sounds good. But I don't understand the change. Why is it called skip? Could you add a few comments explaining the change, and this particular sequence?

src/json/lexer.cr
hinrik: Because it's inside `consume_string_with_buffer`, which cares about collecting the contents of the string into a buffer. Only `consume_string_skip` uses `skip: true`.

hinrik:
Yeah, I can try to make it a bit clearer. There wasn't an obvious simple change to make here because normally the string consuming functions return a `Char`, but for these UTF-8 escapes I'm adding the bytes directly to the string `@buffer`.

Any other logic would have to either rely on lookahead (which would require other changes to `json/lexer/io_based.cr`) or be smarter about the contents of the UTF-8 escape to know when we've parsed enough escapes to constitute a full character.

asterite:
@hinrik I'm sorry, but reviewing the code and seeing the spec, I don't understand the reason of this change. JSON says that to encode a char in UTF-8 you can use `\uXXXX`. That means `\u00c3` is the unicode character with codepoing c3 hexadecimal, which is 'Ã'. I can't find anywhere where it says that should be interpreted as a byte value. The example given in the spec, "\uD834\uDD1E", that should expand into a G clef, works as expected right now. That other parsers, except Perl (I don't know why) work like Crystal confirms my belief. Additionally, the character you want, the one you used in the spec, 'Ö', has a codepoint of 214, so it can be expressed in JSON with "\u00d6".

That Facebook generates invalid JSON is not out problem.

straight-shoota:
This simply doesn't seem correct. There would be no way to disambiguate if `"\u00c3\u0096"` should be interpreted as `Ã\u{96}` (codepoints 195 and 150) or `Ö` (codepoint 214).

asterite:
Exactly.

hinrik:
You're right. From what I can tell, it only worked with Perl due to some accidental re-encoding. Because for backwards compatibility reasons, Perl by default outputs latin1 to stdout. If I tell it to output UTF-8, I get the same output as in Crystal.

jhass:
Yeah you should work around this by preprocessing the JSON before even parsing it to convert the improper escapes into proper ones.

PR #6432, opened 2018-07-22
Fix: private aliases at top-level not considered private

### LINKED ISSUE #5752: Private alias in toplevel are not private
```cr
# some_module.cr
module SomeModule
  def self.foobar
    2
  end
end
```
```cr
# spec1.cr
require "./some_module"

private alias Mod = SomeModule

puts "In spec 1"
puts Mod.foobar
```
```cr
# spec2.cr
require "./some_module"

# no private alias here

puts "In spec 2"
puts Mod.foobar # Mod should not exist here!
```
```cr
# all.cr
require "./foo"
require "./bar"
```
Running `crystal all.cr` will successfully compile & run, even if `Mod` is a private alias in `spec1.cr` and is not defined in `spec2.cr`..

---

Also impacted by this, when using a `private alias` in toplevel in 2 spec files, using `crystal spec` to run all spec files will fail with `alias Mod is already defined`.

So if I change `spec2.cr` to:
```cr
# spec2.cr
require "./some_module"

private alias Mod = SomeModule

puts "In spec 2"
puts Mod.foobar
```
Running the 2 spec files individually will work ok, but running all the specs with `crystal spec` will fail with the `alias .. is already defined` error.

asterite:
Fixes #5752

As a bonus, this works now (similar to other declarations, like classes and constants):

```crystal
alias Foo::Bar = Int32
```

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -610,6 +610,7 @@ describe Crystal::Formatter do
   assert_format "class Foo\nend\nclass Bar\nend", "class Foo\nend\n\nclass Bar\nend"
 
   assert_format "alias  Foo  =   Bar", "alias Foo = Bar"
+  assert_format "alias  Foo::Bar  =   Baz", "alias Foo::Bar = Baz"
   assert_format "alias A = (B)"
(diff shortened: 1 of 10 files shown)


bew:
Awesome, thanks!

PR #6439, opened 2018-07-24
Change Foo:Class representation of Class types to Foo.class

RX14:
Using the Foo.class representation of class types in error messages matches the
type syntax used to write these types, instead of using a special and confusing
syntax unique to compiler error messages.

For example:

```
class variable '@@foo' of Bar must be Hash(Thing:Class, String), not Hash(Thing, String)
```

becomes

```
class variable '@@foo' of Bar must be Hash(Thing.class, String), not Hash(Thing, String)
```

which means the type is now represented exactly the same way in the error
messages as source code.

### DIFF
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -38,7 +38,7 @@ end
 
 class Crystal::Call
   def raise_matches_not_found(owner, def_name, arg_types, named_args_types, matches = nil, with_literals = false)
-    # Special case: Foo+:Class#new
+    # Special case: Foo+.class#new
     if owner.is_a?(VirtualMetaclassType) && def_name == "new"
       raise_matches_not_found_for_virtual_metaclass_new owner
     end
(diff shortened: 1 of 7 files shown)


src/compiler/crystal/semantic/call_error.cr
```diff
@@ -38,7 +38,7 @@ end
 
 class Crystal::Call
   def raise_matches_not_found(owner, def_name, arg_types, named_args_types, matches = nil, with_literals = false)
-    # Special case: Foo+:Class#new
+    # Special case: Foo+.class#new
```
straight-shoota: This is not exactly source code syntax, but it looks as if...

RX14: The `+` syntax is a different issue

RX14:
@straight-shoota but that would require changing the whole language, this is just about changing the error messages / string representation inside the compiler to be more consistent with the existing syntax.

straight-shoota:
I wasn't suggesting to change the language. Just an idea though. This PR is certainly an improvement :+1:

PR #6445, opened 2018-07-25
Fix Random#rand(Range(Float, Float)) to return Float

### LINKED ISSUE #6444: Random#rand with Range(Float32, Float32) doesn't work
[Random#rand](https://crystal-lang.org/api/0.25.1/Random.html#rand%28max%3AFloat%29%3AFloat64-instance-method) method with Range(Float, Float) overload is returning Float64, which makes it impossible to use it with Range(Float32, Float32) argument.

```
random = Random.new

random.rand(1..10) # ok
random.rand(1u8..10u8) # ok

random.rand(1.0..10.0) # ok
random.rand(1.0f32..10.0f32) # error: type must be Float64, not Float32
```
I don't know whether there is a reason, but this overload should probably return Float instead of Float64, as in the case of Range(Int, Int) overload.

Any idea how to work around this, beside casting to Float64 and then back to Float32 after generating random number?

Ubuntu 18.04, Crystal 0.25.1 [b782738ff] (2018-06-27)

straight-shoota:
Fixes #6444

### DIFF
--- a/src/random.cr
+++ b/src/random.cr
@@ -278,7 +278,7 @@ module Random
   # ```
   # Random.new.rand(6.2..21.768) # => 15.2989
   # ```
-  def rand(range : Range(Float, Float)) : Float64
+  def rand(range : Range(Float, Float)) : Float
     span = range.end - range.begin
     if range.excludes_end?
       unless range.begin < range.end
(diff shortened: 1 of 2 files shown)


src/random.cr
```diff
@@ -278,7 +278,7 @@ module Random
   # ```
```
Sija: Need a doc update too (2 lines above).

straight-shoota: Done, thanks.

PR #6450, opened 2018-07-26
Add Fiber#cancel

### LINKED ISSUE #3561: Gracefully stopping fibers
Hi,

While working on the `torrent` shard, I came across the need to tell fibers to stop running forever regularly. Often fixed through other means, like a `running` bool, I came across the following use-case where this is not easily doable:

1. Have a sorted list of requests to be made
2. Spawn a fixed amount of fibers
3. Each fiber takes out a request from the list. If the list is empty, stop the fiber. If all are stopped, **break**.
4. Do the requests. Each response may yield either new requests (Which are then added to the list), OR the result.
5. If a result was found, stop all fibers and return it. **Else**, each fiber goes 3.
6. After each fiber ran a request, another "control" fiber gets sent (through a Channel) the most promising request from each fiber. Now it checks if the most promising request from these requests is better than the previously known best. If it is not, **break** with no result.

The control fiber is basically doing: `fiber_count.times{ candidates.receive }`

The issue is, if the result is found, or the list is empty, the control fiber will now wait in the loop above forever, which would basically leak it. Another approach would be having yet another Channel which is monitored by the control fiber (Through `select`), and if received, halts the control fiber.

However, I'd love to have a primitive which lets me send some kind of message to a Fiber, any Fiber, to please stop itself.

A solution employed by languages like Java is to inject an exception (`java.lang.Thread.interrupt()`). This is quite nice, as properly implemented Fibers would then release resources through `ensure`/`rescue`. It would also kick a sleeping/waiting fiber out of a blocking call like `Channel#receive` or `IO::FileDescriptor#read`.

A second solution could be a `Fiber#close` method, which sets a flag which is checked in `Fiber#resume`, and then runs all `ensure` blocks.

Ideas?

straight-shoota:
This PR adds a method `Fiber#cancel` which allows to stop a fiber as described in #3561

The implementation is pretty simple: Calling `#cancel` sets a flag (`@cancel_request`) and enqueues the fiber to resume at the next rescheduling. `Fiber#resume` checks (after the stack switch) if the now resuming fiber has been requested to be cancelled. If so, it raises `Fiber::CancelledException` which unwinds the stack like any other exception until it reaches the `rescue` handler in `#run`.

Resuming the fiber-to-be-cancelled could already happen in `#cancel` itself. It is better to delay the unwinding until the next rescheduling, in order to allow cancelling multiple fibers at once without any unforeseeable context switches while unwinding the stack.

Apart from the included specs, I have also tested this with #6385 to resolve the issue described in https://github.com/crystal-lang/crystal/issues/6357#issuecomment-405095355 and a proof of concept for a structured concurrency feature (RFC will follow).

An issue worth mentioning is that `CancelledException` is just an ordinary exception and thus it can be rescued from. It would unintentionally be caught by any `rescue` without an explicit type restriction. This is bad because rescuing it somewhere on the unwinding stack can prevents the purpose of cancelling the fiber.

In order to solve this, `CancelledException` should not be matched by a `rescue` without a type restriction. I would recommend a change to the exception hierarchy, inserting a new exception type (for example `BaseException`)  as parent of both `Exception` and `CancelledException`. Untyped `rescue` would still catch all `Exception` types, but not `CancelledException`. But you can still rescue from `CancelledException` (or `BaseException`)  if explicitly declared.
This is similar to the exception hierarchies of Ruby and Python, both have a type for user exceptions (`StandardError` / `Exception`) and above that a base type (`Exception` / `BaseException`) which is also the parent type of system exceptions (for example `SystemExit` in both).

Fixes #3561

### DIFF
--- a/spec/std/fiber_spec.cr
+++ b/spec/std/fiber_spec.cr
@@ -1,0 +1,54 @@
+require "spec"
+
+private def assert_cancel(&block : ->)
+  exception = nil
+  fiber = spawn name: "fiber" do
+    block.call
+  rescue exc : Fiber::CancelledException
+    exception = exc
+  end
+
+  Fiber.yield
+  fiber.cancel
+  Fiber.yield
+
+  exception.should_not be_nil
+  exception.not_nil!.fiber.should eq fiber
+  exception.not_nil!.message.should eq "Fiber cancelled: #{fiber}"
+  cause = exception.not_nil!.cause.not_nil!
+  cause.message.should eq "Fiber cancel request"
+  cause.fiber.should eq Fiber.current
+end
+
+describe Fiber do
+  describe "#cancel" do
+    context "from other fiber" do
+      it "Fiber.yield" do
+        assert_cancel { Fiber.yield }
+      end
+      it "sleep" do
+        assert_cancel { sleep }
+      end
+      it "sleep(1)" do
+        assert_cancel { sleep(1) }
+      end
+    end
+
+    it "from same fiber" do
+      exception = nil
+      fiber = nil
+      fiber = spawn name: "fiber" do
+        fiber.not_nil!.cancel
+      rescue exc : Fiber::CancelledException
+        exception = exc
+      end
+
+      Fiber.yield
+
+      exception.should_not be_nil
+      exception.not_nil!.fiber.should eq fiber
+      exception.not_nil!.message.should eq "Fiber cancelled: #{fiber}"
+      exception.not_nil!.cause.should be_nil
+    end
+  end
+end
--- a/src/crystal/event.cr
+++ b/src/crystal/event.cr
@@ -28,6 +28,10 @@ struct Crystal::Event
     )
   end
 
+  def delete
+    LibEvent2.event_del(@event)
+  end
+
   def free
     LibEvent2.event_free(@event) unless @freed
     @freed = true
--- a/src/crystal/scheduler.cr
+++ b/src/crystal/scheduler.cr
@@ -60,6 +60,14 @@ class Crystal::Scheduler
     current, @current = @current, fiber
     GC.stack_bottom = fiber.@stack_bottom
     Fiber.swapcontext(pointerof(current.@stack_top), fiber.@stack_top)
+
+    # The stack has changed but `fiber` still references the old fiber, that's
+    # why we acquire the actual current fiber.
+    current_fiber = self.class.current_fiber
+
+    if cancel_request = current_fiber.@cancel_request
+      raise Fiber::CancelledException.new current_fiber, cancel_request
+    end
   end
 
   protected def reschedule : Nil
--- a/src/fiber.cr
+++ b/src/fiber.cr
@@ -174,6 +174,59 @@ class Fiber
     Crystal::Scheduler.resume(self)
   end
 
+  @cancel_request : CancelRequestException? = nil
+
+  # Stops this fiber from executing again and unwinds its stack.
+  #
+  # This method requests the fiber to raise a `CancelledException` the next time
+  # it is resumed and enqueues it in the scheduler. The unwinding will only take
+  # place the next time the scheduler reschedules.
+  #
+  # Immediately raises `CancelledException` if this is the current fiber.
+  def cancel
+    if Fiber.current == self
+      # In case the current fiber is to be canceled, just raise the exception directly.
+      raise CancelledException.new self
+    end
+
+    # Register a cancel request, it will be evaluated on next resume.
+    @cancel_request ||= CancelRequestException.new(CallStack.new)
+
+    # Trigger scheduling
+    @resume_event.try &.delete
+    Crystal::Scheduler.enqueue(self)
+  end
+
+  # A `CancelRequestException` holds the callstack of `CancelledException` where
+  # `Fiber#cancel` was called.
+  class CancelRequestException < Exception
+    getter fiber : Fiber
+
+    def initialize(@callstack : CallStack, message : String = "Fiber cancel request")
+      super(message)
+      @fiber = Fiber.current
+    end
+  end
+
+  # A `CancelledException` is raised when a fiber is resumed after it was cancelled.
+  # It unwinds the fiber's stack and should typically not be rescued.
+  #
+  # See `Fiber#cancel` for details.
+  #
+  # `cause` points to the fiber and callstack where the cancel request originated.
+  # If it is `nil`, the fiber was cancelled while executing.
+  class CancelledException < Exception
+    getter fiber : Fiber
+
+    def initialize(@fiber : Fiber, cause : CancelRequestException? = nil)
+      super("Fiber cancelled: #{fiber}", cause)
+    end
+
+    def cause : CancelRequestException?
+      @cause.as(CancelRequestException?)
+    end
+  end
+
   # :nodoc:
   def resume_event
     @resume_event ||= Crystal::EventLoop.create_resume_event(self)


asterite:
How is the performance of https://github.com/crystal-lang/crystal/blob/master/samples/channel_primes.cr affected, replacing 100.times with a bigger number (I think 1000 or 10000 is good for a benchmark)

straight-shoota:
This is the mean over 5 runs of `samples/channel_primes.cr` with 10k iterations each:
```
master: 33.640000   7.760000   41.400000 (  45.352300)
 patch: 35.680000   7.842500   43.522500 (  46.928800)
  diff: +1.06%     +1.01%      +1.05%    (  +1.03%   )
```

1% doesn't sound too bad considering the example is constantly switching fibers back and forth. In a real-world application it'll barely be noticeable.

The performance loss could probably be avoided with a customized `#resume` method that get's setup to be scheduled when `#cancel` is called. But I'd leave such an optimization for later.

RX14:
With a bit of work an uncatchable exception could be added. But it wouldnt be very useful since `ensure` blocks can raise, and if they raise that exception can easilly be caught. Accidentally or not.

I don't think this mechanism is robust enough to add.

straight-shoota:
I don't think this is only about avoiding deadlocks with mutexes or anything but to properly clean up a fiber's stack, which means to execute all `ensure` blocks.

> I don't think this mechanism is robust enough to add.

What do you mean with that?

RX14:
@straight-shoota what do you mean clean up the stack? You can just delete the stack to clean it up. The only issue is you can get deadlocks when you never unlock a mutex or so.

One way to do that is to only execute ensures, but you don't have to go that route.

straight-shoota:
I mean to unwind the stack and doing all the clean up in the process, such es executing `ensure` blocks. I mean, if you write an `ensure` block, you expect that to always execute (unless there is a huge error like segfault).

Cancelling a fiber is not huge error (it's not an error at all) but rather a simple way of canceling a task when it is no longer needed. But in my opinion, it should not just try to do the bare minimum to avoid deadlocks, but properly unwind and do clean up. I guess this is similar to `SIGTERM` and `SIGKILL`.

I'd like to have a way to gracefully stop a fiber (likely while it is waiting for IO) which can't be implemented in a nice way in user code. A typical use case for this would be something like the typical algorithm for establishing a TCP connection, which starts multiple connection attempts more or less simultaneously in fibers and as soon as one of them succeeds, the others can be stopped. But they should properly unwind and release their resources (for example sockets).

ysbaddaden:
2 exceptions, 2 stack unwinding (very costly) and an additional check for each and every fiber resume :-/

Isn't killing a thread considered bad practice? I don't see why canceling a fiber would be better.

There are better patterns to stop a fiber with proper cleanup than a hard cancel.

straight-shoota:
Well, the inner exception is optional. I thought it might be useful to know where a fiber was cancelled, but in practice it's probably not that important.

So it could be only one exception and unwinding. And I don't think there is an alternative to that if you want to execute `ensure` blocks.

The additional check for each `resume` adds a minimal overhead. And there should be a way to avoid that if really necessary.

> There are better patterns to stop a fiber with proper cleanup than a hard cancel.

First of all, it's not a hard cancel. The fiber can rescue from the exception if necessary.

What would be a better pattern, say for my example use case?

RX14:
if we do this we should do it with custom codegen so it's actually impossible to rescue. It shouldn't be too hard.

ysbaddaden:
What bugs me is two fold:

1. Stopping a fiber from an outside context (i.e the whole concept of cancel) when I believe there are better patterns;

2. Raising an exception in the fiber context when a similar concept, `exit` specifically doesn't and uses `at_exit` callbacks instead for cleanup.

straight-shoota:
`Fiber#cancel` should generally be followed by `Fiber.yield` or similar. The fiber to be canceled is enqueued, so it will be executed next. I'm sure this would be fine as a convention. It would also be an option to yield directly from `#cancel` by default, and provide an argument to not do this (for the purpose described in the OP).

Yes, in some cases it is better to use a channel or a simple loop condition to let the fiber stop itself. The problem is that this doesn't work when the fiber is waiting for an event like IO or a timeout. The only way this can be implemented is to insert a check for a cancel request after each expression that could include a context switch. As this is not practical, this logic should be in `Fiber#resume`.

I don't think this needs custom codegen. It is fine if a `CancelledException` can be caught if explicitly asked. It just shouldn't be caught by the default `rescue` without type restriction. This only requires a minor change in the compiler. And the feature would probably also be fine without it. A blank `rescue` should be treated with care, anyways. And in stdlib it is mostly used for code for error handling, so not a big deal. But `do_something_that_might_raise rescue nil` is a common idiom that shouldn't swallow `CancelledException`.

`exit` is called maximum once per process. And I would probably even prefer to raise a `SystemExitException`. Cancelling a fiber however, can be much more common and what happens inside a fiber can't be known by the code that starts and cancels a fiber. And IMHO that's exactly what `ensure` blocks are made for. To clean up whether the task was successful or not (including being cancelled).

RX14:
@straight-shoota the problem is that without a larger change to the compiler, an ensure block that raises an exception will raise a *standard* exception which *will* be cause by `rescue` without a restriction. This is an easy bug to create, and a pretty hard bug to catch or test, since ensure throwing is so rare.

asterite:
Actually the only change needed would be to introduce an intermediate exception and make `rescue` catch that intermediate exception by default. To catch all exceptions you would rescue from Exception, or Raisable, something like that. Then we could also raise on `exit`, I think there was an issue about that...

But I don't know about cancelling a fiber. There must be a reason no other language allows such thing.

RX14:
@asterite I just described the problem with that approach and I think it's the kind of ugly problem that will raise it's head in production and make a mess.

straight-shoota:
> There must be a reason no other language allows such thing.

This proposed behaviour is actually implemented very similar to Python's [`asyncio.Task.cancel`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) and Trio's [`cancel`](https://trio.readthedocs.io/en/latest/reference-core.html#trio.The%20cancel%20scope%20interface.cancel).

@RX14 What exactly is the problem with the solution I proposed in the original post? It shouldn't be very difficult to insert a new exception type above `Exception` and let `CancelledException` inherit from that. The only change in the compiler would be to recognice the new parent as base exception type but still only catch `Exception` by default.

asterite:
It also happens that Ruby has some exceptions that aren't caught by default, and the same is true for a Java (the Error class). In C# there's a ThreadAbortExcrption that can be caught but it's automatically rethrown.

It seems the proposed solution here is similar to other languages, I don't see a problem with that.

Maybe it would be nice if @waj could comment whether gracefully stopping a fiber is something bad.

RX14:
@straight-shoota I just described how that fails!!! Please read again

straight-shoota:
Oh, seems like I had overlooked [that comment](https://github.com/crystal-lang/crystal/pull/6450#issuecomment-408642471). I'm sorry.

I'm not sure this is really a big issue though. An `ensure` block should normally not raise. If it raises, it is most likely a severe error that won't allow the continuation of the task anyway. And a blank `rescue` without restrictions can always be a reason for bugs if it catches something you didn't expect to be thrown. As I already mentioned, it is mostly used for things like `do_something_that_might_fail rescue nil` if you don't care if there is an result. Maybe in such cases it would be better anyways to only catch specific exceptions. And for this we need a better exception hierarchy.

RX14:
@straight-shoota yes it usually is a severe error that won't allow continuation of the task. But so are most exceptions, and raising from `ensure` really isnt that uncommon. `io.close` can easily raise when it flushes buffered writes, and `io.close` is a common `ensure`.

Such a common thing shouldn't abort the whole cancelling of a fiber. That's why it'd need seperate codegen, to automatically silently ignore exceptions in ensure blocks when you were unwinding to close.

And given that complexity, I really don't like the idea of unwinding at all. I'd like to try the simple solution first, which is not to unwind and to identify the few resources which *can* deadlock.

RX14:
And not to mention, even having the *possibility* that a fiber can "fail to cancel" and keep running is confusing for the programmer, invites edge cases by it's very nature, and requires more careful design and more code.

straight-shoota:
I don't think it is an issue if a cancellation might not work out. It is after all a way to try to gracefully stop a fiber. Concurrency in Crystal is cooperative, so it always depends on how other components behave.

Being able to cancel a fiber gracefully with a chance of not succeeding is an improvement over not being able to cancel at all. It could very well be used in a way that a fiber is given some time to cancel itself and after that, it might just get hard killed (removed from the Scheduler).

Cancellation through an exception is a very simple implementation and IMHO it is sufficient for most use cases, even if it can't guarantee the `CancelledException` goes all the way to the top of the stack.

RX14:
@straight-shoota the problem is not that cancel can fail to *gracefully* stop a fiber, it's that it can actually accidentally resume a fiber instead. That's the whole point of what i'm saying, a cancel which can *fail resumed from a "paused" state* is useless. It's not an improvement on the current situation at all.

straight-shoota:
I'm not sure why this is a problem. The entire point is to not let a waiting fiber hang around until some timeout kicks in - or even indefinitely.

RX14:
If you call `#cancel`, the fiber should never resume running. It's as simple as that.

asterite:
I don't understand the problem with ensure and that exception.

straight-shoota:
@RX14 That's how Pythons cancel implementation works. And I don't think it is wrong.

> This arranges for a CancelledError to be thrown into the wrapped coroutine on the next cycle through the event loop. The coroutine then has a chance to clean up or even deny the request using try/except/finally.
> This does not guarantee that the task will be cancelled: the exception might be caught and acted upon, delaying cancellation of the task or preventing cancellation completely. The task may also return a value or raise a different exception.

If `CancelledException` arrives at the fiber's root, it is marked as `cancelled`. The caller of `cancel` could check after some grace period if the fiber was successfully cancelled. If not, it can either try again or do something else.

straight-shoota:
@vipsmaker You might want to take a look at #6468

vinipsmaker:
> an additional check for each and every fiber resume :-/

Context switch is where the fat goes anyway. You want to do progress as much as you can before doing a context-switch. Checking on a bool on context-switch seems rather cheap to me. You might enjoy reading this: https://lwn.net/Articles/763056/

> Isn't killing a thread considered bad practice? I don't see why canceling a fiber would be better.

Recently I've implemented a `fiber.interrupt()` for a multi-threaded fiber implementation I use daily at work. I've researched several solutions to stop a thread, including Windows API, POSIX threads, Go approach, discussions on Rust community, Java's list of deprecated thread features and Java's `interrupt` API, Boost.Fiber, Boost.Thread and maybe a few more that I don't remember right now. It turns out that POSIX-like thread cancellation API is _the_ cleanest solution to interrupt a thread. And the problem is so common that you probably want implement this API rather than pollute user-code with ugly never-ending cascades of handles being passed at every level of abstraction to do yet another ugly select-over-multiple-channels.

I guess the “bad practice” that you mention may refer to APIs that don't match POSIX semantics enough like Windows' and Java's.

In my case, a fiber may not always wake-up on the same thread and there were some kinds of races I had to deal with. My implementation is also more complete with features such as `pthread_cancel_state()` equivalent. So my case is much more complex.

vinipsmaker:
> If you call `#cancel`, the fiber should never resume running. It's as simple as that.

Wrong. If there is clean-up code to be executed, you're already scheduling some task to run (i.e. the `#resume`).

vinipsmaker:
> If `CancelledException` arrives at the fiber's root, it is marked as `cancelled`. The caller of `cancel` could check after some grace period if the fiber was successfully cancelled. If not, it can either try again or do something else.

The “it can either try again or do something else” part worries me. It just complicates design (like Java's state machine on interruption API).

straight-shoota:
@vinipsmaker 
>The “it can either try again or do something else” part worries me. 

What should happen in this case instead?

straight-shoota:
Thanks for your great input on this topic, btw!

rdp:
FWIW I don't love this option, it's a great tool for beginners but can introduce subtle bugs (like exceptions being raised *inside finalize handlers* so they basically only half cleanup, etc...it's a can of worms, and can cause unexpected behaviors, similar to how Timeout.timeout in ruby feels error prone  :|

straight-shoota:
@rdp A finalized can only raise if it contains a potential fiber switch. This would most likely be some kind of IO operation which could already raise for a number of other reasons.

The alternative to hard-kill a fiber doesn't even give a chance to call any finalizers.

RX14:
> A finalizer can only raise if it contains a potential fiber switch

I'm not sure what you mean?

straight-shoota:
Some IO operation, `sleep` or similar which allows the scheduler to move execution to a different fiber. If the finalizer runs without interruption in the same fiber, it can't raise.
But obviously, finalized are often used to release some resource which likely includes IO operations.

RX14:
Oh, I think my brain did an incorrect autocorrect. @rdp do you mean object finalizers or `ensure` blocks? Because the former is already a problem.

rdp:
I was referring to ensure blocks (which can be nested, call other methods,
etc.), sorry to not be clear there.

RX14:
@vinipsmaker crystal's current exception model is that everything inherits `Exception`, and you must raise an `Exception`. The default `rescue` type is `Exception`. So there's no way to raise an exception that isn't caught by default. Yet. I think probably the best way to solve this problem is to introduce a new layer to the exception hierarchy, similar to `Throwable` in Java.

The suggestion to only cancel fibers on suspention points is great!

straight-shoota:
> The suggestion to only cancel fibers on suspention points is great!

That's actually the only feasible way for cancelling fibers. There is no way to preempt a fiber, so it can't be interrupted at any time ("asynchronous" in pthread lingo). The currently scheduling model only allows for "deferred".

UlisseMini:
Regarding adding an uncatchable exception, erlang has this via the [exit function](http://erlang.org/doc/reference_manual/processes.html#process-termination) if the reason given is `kill` then it cannot be caught and cleanup cannot be ran.

Also haskell includes a function [throwTo](https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Concurrent.html#v:throwTo) that throws an exception to a green thread (the name `ThreadId` is misleading)

Personally i'd like to be able to raise an exception in a fiber, with this the author of the code would be in control rather then calling some mystic `Fiber.kill` method and not knowing the details.

Just thought i'd mention this, studying other solutions to a problem is the #1 way to get up to speed fast :)

rdp:
uncatchable exception I presume still runs ensure blocks?  In my head it's still just too much of a danger to developers, it opens a can of worms if you let people get by without having graceful stop points, FWIW...

straight-shoota:
Closing stale PR.
Discussion should be continued in #6468.

Issue #6386, opened 2018-07-14
HTTP::Handler override Proc


jwoertink:
While working on [this commit](https://github.com/luckyframework/lucky/pull/541), I noticed that [Proc](https://github.com/crystal-lang/crystal/blob/master/src/http/server/handler.cr#L32) is alias in HTTP::Handler. This caused an issue when I included this module in to a custom handler class, and I wanted to make a property of type `Proc`.

Using `property whatever : Proc(Other, Stuff, Here)` throws the error `Proc(Other, Nil) is not a generic type, it's a function`. To fix this, I have to add `::Proc(Other, Stuff, Here)`. The other thing was that the code led to [crashing](https://github.com/crystal-lang/crystal/issues/5656) the compiler.

My thought was that maybe if this alias is necessary, it could have a custom name like `HandlerProc` or something. If not, then I guess it could just be removed?

asterite:
Yeah, maybe it can be renamed to something else.

RX14:
@jwoertink just use `property whatever : Other, Stuff -> Here`.

Issue #5890, opened 2018-03-31
Possible cache race condition


Sija:
Refs:
- https://github.com/crystal-lang/crystal/pull/5447#issuecomment-354624371
- https://github.com/crystal-lang/crystal/pull/5882#issuecomment-377703015

/cc @asterite @RX14

Sija:
And again - https://travis-ci.org/crystal-lang/crystal/jobs/373199400#L540-L546:

```
C-har.o: In function `*Char#to_s:String':
Char:(.text+0xea7): undefined reference to `*Pointer(UInt8)@Pointer(T)#appender:Pointer::Appender(UInt8)'
Char:(.text+0x12c2): undefined reference to `*Pointer(UInt8)@Pointer(T)#[]=<Int64, UInt8>:UInt8'
/usr/bin/ld: /tmp/crystal/crystal-run-crystal-spec-output.tmp: No symbol version section for versioned symbol `*Pointer(UInt8)@Pointer(T)#[]=<Int64, UInt8>:UInt8'
/usr/bin/ld: final link failed: Nonrepresentable section on output
collect2: error: ld returned 1 exit status
Error: execution of command failed with code: 1: `cc "${@}" -o '/tmp/crystal/crystal-run-crystal-spec-output.tmp'  -rdynamic  -lpcre -lgc -lpthread /mnt/src/ext/libcrystal.a -levent -lrt -ldl -L/usr/lib -L/usr/local/lib`
```

RX14:
We've never ever had any reports of this outside CI, which is extremely strange considering how often it happens on CI. Very interesting.

j8r:
It seems to only occurs in Travis? Interesting, very likely a system issue.
Everything won't probably be possible on Travis, but we may try to change some system parameters of the underlying OS.

For example increasing the value of `/proc/sys/vm/max_map_count`(maximum number of memory map areas per process).
Not sure about previous errors of this type, but here it occurs at the end - maybe no more memory map, memory cache or something else?

I've also read [this](https://docs.travis-ci.com/user/common-build-problems/#Segmentation-faults-from-the-language-interpreter-(Ruby%2C-Python%2C-PHP%2C-Node.js%2C-etc.)), not sure if it's related but we may also try to clear the caches before compiling:
```sh
echo 1 >/proc/sys/vm/drop_caches
echo 2 >/proc/sys/vm/drop_caches
echo 3 >/proc/sys/vm/drop_caches
sync
```

[...]

PR #6453, opened 2018-07-27
rename alias HTTP::Handler::Proc to HTTP::Handler::HandlerProc

jwoertink:
closes: #6386 

I know that the name is a bit verbose, but naming is hard. 😂  If anyone has a suggestion for a better name, I'll make the change.

The reason for the rename as mentioned in that issue is that when you are creating a custom HTTP::Handler, and you include it in your class, you're not able to use the standard type `Proc` anymore since it's been overridden by the alias. I hope no one was using this in particular, but it should probably be noted as a breaking change just in case.


jwoertink:
I'm not sure what that travis error is.

Sija:
@jwoertink Unrelated to this PR, see #5890

jwoertink:
Ok cool. Just making sure I didn't break something lol

PR #6454, opened 2018-07-27
Improve samples

r00ster91:
This:
- Revives dead links by updating them and using the [WayBackMachine](https://web.archive.org/)
- Improves readability a bit by using `abort "message"` instead of

  ```crystal
  puts "message"
  exit 1
  ```
- Uses `Time.measure` instead of `Time.now - time`
- Improves other minor things


z64:
Indeed, all that 404s is not lost :smile: 

https://web.archive.org/web/20160527042358/http://benchmarksgame.alioth.debian.org/u64q/program.php?test=binarytrees&lang=yarv&id=1

r00ster91:
@z64 Ah yeah using the waybackmachine for the links is a good idea.

@straight-shoota But I don't really see why it's a problem to replace the `while true`s with `loop do`s. It just reads better and is the prefered way to make a loop. And maybe a beginner takes some of these samples as a reference and then uses `while true` instead of `loop do` in his code..

asterite:
Thank you for the PR, but the code as it is right now is fine.
While true is better than loop in my opinion.
Using tuple instead of Array is not always better (tuple should just be used to return multiple values, or to group values, but it makes little sense to use in constants)
I could go on but I don't have much time right now.
Please don't merge this.

hugoabonizio:
Official samples should reflect idiomatic Crystal code and its best pratices. Although disagree with some changes (I'm included), this is a legitimate attempt to improve and bring consistency to samples code. I think we should agree on some basic changes (e.g. unifying `while (true)` and `while true`, using tuples for immutable data, etc), instead of keeping as is.

asterite:
Using tuples for immutable data is not always correct. In samples all arrays have the same type. An array should be used, not a tuple. A tuple is not an immutable array.

asterite:
Unifying code syntax is not important. The language allows for many styles, all are correct.

straight-shoota:
@r00ster91 
> But I don't really see why it's a problem to replace the `while true`s with `loop do`s.

And I don't see a problem with `while true` why that should be changed.

> It just reads better and is the prefered way to make a loop. 

Says who?

> And maybe a beginner takes some of these samples as a reference and then uses `while true` instead of `loop do` in his code..

And that's perfectly fine. I'd even advice to do so.

Updating links to new locations is great. Code changes like `content_type`, too. Everything else doesn't make sense. Especially removing links and comments.

Sija:
👍 for constants and `abort`.

r00ster91:
Hmmmm okay will change it back to `while true` then.

samples/meteor.cr
```diff
@@ -1,5 +1,3 @@
-# Translated from: https://github.com/mozilla/rust/blob/master/src/test/bench/shootout-meteor.rs
```
Sija: https://gitlab.redox-os.org/redox-os/rust/blob/a59de37e99060162a2674e3ff45409ac73595c0e/src/test/bench/shootout-meteor.rs

samples/sudoku.cr
```diff
@@ -124,31 +124,31 @@ def sd_solve(mr, mc, s)
   ret
 end
 
-sudoku = "
-..............3.85..1.2.......5.7.....4...1...9.......5......73..2.1........4...9       near worst case for brute-force solver (wiki)
```
Sija: These comments should stay.

r00ster91: Ooh I didn't realise that these strange texts are actually comments for the specific lines. I thought they got there by accident when copying them out somewhere. I think I will add a `#` before every comment so you better notice that they are actually comments for the specific lines.

asterite:
Also accessing constants is slower than accessing local variables. Please, there's no point to these PRs.

r00ster91:
@asterite So I should revert the constants back to local variables? But at for example the [`SYMBOLS` constant](https://github.com/crystal-lang/crystal/pull/6454/files#diff-4bf87c04296a5262af90ebc1a9555412R74) it says in the [docs](https://crystal-lang.org/docs/guides/performance.html#avoid-creating-temporary-objects-over-and-over) that theres currently a big performance problem because on every [`Int#times`](https://crystal-lang.org/api/0.25.1/Int.html#times(&block:self-%3E):Nil-instance-method) iteration a new array is created. And this is prevented by making [`SYMBOLS`](https://github.com/crystal-lang/crystal/pull/6454/files#diff-4bf87c04296a5262af90ebc1a9555412R74) a constant. And that's also the case with [`SCENE`](https://github.com/crystal-lang/crystal/pull/6454/files#diff-9d75bfff942c10ed1bd07fb640773305R107) I think.

But you are right with the [`SUDOKU`](https://github.com/crystal-lang/crystal/pull/6454/files#diff-b60d3cea3cb466a70dc29f8725e8a219R127) string constant (I benchmarked). I'll change it.

asterite:
That's if you use an array literal inside a loop. Accessing a local variable inside a loop is fine.

r00ster91:
Alright changed it.

r00ster91:
I resolved the conflicts. Is anything left?

samples/http_server.cr
```diff
@@ -1,9 +1,10 @@
...
-  context.response.headers["Content-Type"] = "text/plain"
-  context.response.print("Hello world!")
+  context.response.content_type = "text/plain"
+  context.response.print "Hello world!"
 end
 
-puts "Listening on http://0.0.0.0:8080"
-server.listen "0.0.0.0", 8080
+puts "Listening on http://127.0.0.1:8080"
+server.listen(8080)
```
straight-shoota: I'd prefer this to be
```cr

server.listen
```

Issue #6083, opened 2018-05-09
Automatic cast breaks playground


sdogruyol:
After #6074 merged, this following works.

```
x : Int8

x = 1

pp x
```

However it fails in playground with the following error

```
ERROR -- : Instrumention bug found (session=1, tag=1).
```

asterite:
Yeah, that won't work. I think that gets re-written to:

```
x : Int8

x = instrument { 1 }
```

and of course that won't compile.

My suggestion: move the playground outside of the compiler, as a separate project. It will free us from having to track and fix these bugs here.

I really like the playground, but because of how it's currently implemented there's no way it can be **exactly** the Crystal language. There are many easy ways to break it by now.

j8r:
Having the playground in the compiler directory looks a bit weird.
Creating https://github.com/crystal-lang/playground would be good.

Jens0512:
+1 for moving it to crystal-lang/playground

RX14:
I don't see much benefit of splitting it out. It's not crystal but it's hardly icr-levels of hacks and works as a good first approximation. Removing it is just detrimental to newcomers for purism's sake.

asterite:
Who from the core team uses the playground? Or someone here that uses the playground daily?

bararchy:
@asterite I do, also @artlinkov its great for prototype and testing new ideas and types

j8r:
In my opinion, a "standard playground" has little sense, it's a separate project of the Crystal compiler. It has its own HTML `views`, has its JS libraries in `public` with also CSS. And perhaps in the future it will use some npm packages with a `package.json`, JS linters/tests frameworks...

Tere are so many awesome tools like [ameba](https://github.com/veelenga/ameba) that will fit better in the compiler tools, but they aren't integrated and it is fine.

asterite:
Not to mention that the compiler depends on `HTTP::Server` for this whole purpose, and it's why we have all that `without_openssl` mess, because we can't compile the compiler and easily statically/dynamically link openssl. It would simplify things so much.

luislavena:
IMO we should ship playground with releases, but not necessarily have it part of the compiler itself. Similar to how shards is built and _bundled_ with a new release, the same should be for playground.

My 2 cents.

Cheers.

RX14:
If we ship it with the release that would mean we still need to have `-Dwithout_openssl`.

I strongly vote keep it.

j8r:
Let's do a poll for the future of playground:
[...]

Issue #5215, opened 2017-10-30
[RFC] Shrink the standard library


asterite:
This issues is very controversial, but it was never really discussed in the past.

I've been thinking more and more about moving out pieces of the standard library to shards. Examples of these are the `YAML` and `XML` modules, though we could include more modules (`CSV`, `Zip`, `Big*`, maybe even `HTTP`).

These shards will most probably live under the `crystal-lang` organization, so they will be kind of "blessed".

The pros of doing this are:
- These shards can now evolve independently from the compiler. If `YAML` would have been a shard, we could already have the new functionality without having to wait for a next compiler release.
- These shards can have more "core" collaborators than the crystal repository. Because they are less critical, we could accept more contributors.
- These shards don't need to freeze their API in a backwards-compatible way after 1.0, unlike the standard library
- It simplifies the maintenance work for the crystal core developers
- It simplifies the release process. For example we lost a lot of time in the past release (0.24.0) because the version of `libxml2` that comes with Debian 7 doesn't work in Crystal. Specs weren't (and aren't) passing there. But nobody complains, so maybe nobody uses Debian 7 anymore, or maybe nobody uses Crystal+XML in Debian 7. If XML were a shard, we wouldn't have to worry about all of this (well, yes, but not in each release, we could solve this faster and better), and we would have 0.24.0 some weeks ago.
- These shards could provide more features than what they provide now. We tend to be very conservative with the standard library because every public method is something that needs to be maintained and kept for backwards compatibility. We will never add CSS3 style selectors because it's too much work and we wouldn't want to have to maintain that, and maybe it's not appropriate for the standard library. If XML is a shard then all of these reasons disappear, and it can evolve faster and better
- It's already known that some standard libraries didn't age well. Ruby had Tk in the standard library. Maybe it was popular before, but now it's not. Eventually they moved it out of the standard library. With Java, Python and C# the same happened, either because they became obsolete, or because an alternative, better solution could be developed, one that didn't have to live by the backwards-compatible rule.
- Should the Crystal distribution depend on libyaml and libxml2? Well, they come in the standard library, but they are only needed if you use them. But the compiler doesn't use them (this is why they were removed at one point from homebrew, then added back because they are needed just in case someone wants to use those modules). Well, if they are outside the standard library and the compiler distibution that problem goes away. You install the XML shard and there it says you need to also install libxml2, or maybe it comes with a postinstall that will install it for you.

Of course there are cons too. These are:
[...]

Issue #4613, opened 2017-06-23
Improve Markdown implementation - Or: Should we separate compiler tools?


straight-shoota:
The Markdown parser and renderer is still very basic and misses important features like for example raw html.

I suggest to implement the [CommonMark](http://commonmark.org/) specification which is very reasonable and the de-facto standard reference for Markdown implementations. There is an entire test suite [jgm/CommonMark](https://github.com/jgm/CommonMark) to validate implementations. Currently `Markdown.to_html` passes only 137 of 621 (22 %).

It would also be nice if the Markdown parser/renderer can be easily extended to add support for more specialized features like Github Flavoured Markdown (e.g #2217)

MakeNowJust:
GitHub Flavored Markdown (GFM) has [the spec](https://github.github.com/gfm/) now.

straight-shoota:
Yes, that's the CommonMark spec plus a few extensions, so GFM is essentially a superset of CommonMark.  These extensions are:
* Tables (#2217)
* Task List Items (#3459)
* Striketrough
* Extended Autolinks
* Disallows Raw HTML

As [asterite and ysbadden stated](https://github.com/crystal-lang/crystal/pull/3459#issuecomment-255587189) it is probably best if the default implementation does not include extensions.

There is already a Crystal implementation for CommonMark at [https://github.com/ujifgc/crmark](ujifgc/crmark), though it is a direct port of the JavaScript reference implementation, so it can certainly be optimized in many ways.

straight-shoota:
Instead of implementing a parser in Crystal we could also use the existing [C implementation](https://github.com/jgm/cmark). It supports custom extensions as well, though I have no idea how to incorporate them.

bcardiff:
It would be good to adopt the full CommonMark specs in the stdlib. But I think it will be better to have a crystal native implementation. That way we could toggle extensions like GFM which many probably want.

If the failing CommonMark specs can be grouped and the design of the parser the reference implementation can be ported then we might be in a better shape to improve the markdown implementation as you suggest.

The markdown parser is mainly used to generate the docs from crystal code. And it is slightly used in the playground. That is basically why the current implementation didn't need all markdown features.

straight-shoota:
[...]

PR #6455, opened 2018-07-27
Move out Playground from the compiler

j8r:
The playground will live in a separate repository, for now at https://github.com/j8r/crystal-playground
There are still minor issues to fix.

There are upsides/downsides of moving out the playground.
### Pros:
- Remove web related JS/CSS/HTML code of the compiler
- More time to focus on the Crystal language itself by removing JS/CSS/HTML issues and PRs
- Reduce the compile time and binary size of the compiler
- Remove the `HTTP::Server` dependency from the compiler
- All `without_openssl` could also be removed

### Cons:
- complicates the release cycle
 :arrow_forward:  Also remove playground from the tarball release. 
The release tarball will be also marginally smaller.

- I like `crystal play` and I want to keep using it
 :arrow_forward: You will be able to do either:
`brew install crystal-playground`
`apt install crystal-playground`
or clone `https://github.com/crystal-lang/crystal-playground` and `crystal build` it

### Some stats:
`make release=true` time afer a `make clean` on an Alpine Edge Docker container:

no playground:
```
real    9m 19.89s
user    9m 17.93s
sys     0m 4.60s
```
`crystal` binary size: 24.416240M

with playground:
```
real    9m 39.98s
user    9m 35.07s
sys     0m 5.08s
```
`crystal` binary size: 25.117896M

Relative discussion: #6083


straight-shoota:
I think this PR is too premature. There is no decision made neither in #6083 (which is originally an unrelated issue) nor anywhere else. And making a decision requires more discussion about how the compiler, compiler tools and standard library should be developed and integrated. This also touches #5215 and #4613. Until there is a plan for all of this, I don't think we need any PR.

j8r:
@straight-shoota sorry I don't quite understand what you mean. The change is only about the Playground, nothing else.
[People thoughts](https://github.com/crystal-lang/crystal/issues/6083#issuecomment-387872883) on this are quite clear, 19 👍 to move out vs 3👎 to do nothing.

This doesn't stop to have the same discussion again here of course, but I'm afraid having general discussions won't lead to any concrete actions.

Please let's talk on what the PR is about here – move out the Playground, and its [dozens of related issues](https://github.com/crystal-lang/crystal/search?q=playground&state=open&type=Issues). One thing at a time...

Sija:
@j8r I agree with @straight-shoota. Whatever the actions needed to be taken, IMHO they should be stewarded by someone from @crystal-lang core team, not outside contributors and for sure the repo shouldn't be outside of @crystal-lang org.

j8r:
BTW some core members has already expressed opinions:
[@asterite](https://github.com/crystal-lang/crystal/issues/6083#issuecomment-387729033), [@bcardiff](https://github.com/crystal-lang/crystal/issues/6083#issuecomment-395692480) and [RX14](https://github.com/crystal-lang/crystal/issues/6083#issuecomment-387799981)

The creation of  https://github.com/crystal/playground [has been asked](https://github.com/crystal-lang/crystal/issues/6083#issuecomment-392332281)

We know what most people think on this, the decision is up to @crystal-lang members now.

Sija:
> BTW some core members has already expressed opinions:
> @asterite, @bcardiff and RX14

Opinions are **not** actions.

> The creation of https://github.com/crystal/playground has been asked

Yet not answered (yet), hence 404 on your link.

> We know what most people think on this, the decision is up to @crystal-lang members now.

Amen.

RX14:
The playground is intrinsically linked to the compiler internals. Putting it inside a different repository will just make refactoring the compiler harder. This is a bad idea.

asterite:
The playground is what makes all those `if flag openssl` needed. The playground can catch up to compiler changes. It really doesn't belong in the compiler.

bcardiff:
I really appreciate that the playground is extracted in a new repo while keeping the history. 

But, this PR can't be accepted until:
1. the distribution scripts are updated to deliver the same experience (`$ crystal play`)
2. a ci in playground is setup to use nightly releases of the compiler to avoid issues on release
or define brew formulas / linux packages to distribute the playground, which would lead probably to have a compiler, playground and whole bundle packages.

Also, as stated [here]:

> I would like to see this done only and as long as the compiler is extended to let the program be instrumented in a better way as it is today. Otherwise changes to the language or the the ToSVisitor are too sensible to break the playground.

Another cons of splitting is that the bin wrapper of the compiler will need to be duplicated for the playground probably, and actually the playground will have a fully featured compiler, but is not the compiler...

I don't think we are going to invest in the required tasks mention earlier so unless those want to be tackled by someone, this leave things stalled for this PR for the moment.

I acknowledge the will to split it from the compiler, but I don't think is going to happen soon.

j8r:
It could happen in the far future, not anytime soon – closing.

PR #6458, opened 2018-07-27
Fix JSON::Serializable with 2 args, and YAML::Serializable with 3 arg…

### LINKED ISSUE #6405: JSON::Serializable causes compilation error when combined with an initialize method with exactly 2 params
When you have a class that includes JSON::Serializable and also has an initialize method that takes in 2 params a compilation error is raised: 
```crystal
instance variable '@foo' of Bar must be String, not JSON::PullParser
```
See playgrounds below for examples
https://play.crystal-lang.org/#/r/4jka - works fine (notice 3 vars in init call)
https://play.crystal-lang.org/#/r/4jkb - compilation error (notices 2 vars in init call)
https://play.crystal-lang.org/#/r/4jkc - also works fine (notice 1 var in init call)

Seems to only be occurring when init takes 2 params, 1 and >2 work as expected.
Cheers

kostya:
…s, fixed #6405

### DIFF
--- a/spec/std/json/serializable_spec.cr
+++ b/spec/std/json/serializable_spec.cr
@@ -35,6 +35,16 @@ class JSONAttrPerson
   end
 end
 
+struct JSONAttrPersonWithTwoFieldInInitialize
+  include JSON::Serializable
+
+  property name : String
+  property age : Int32
+
+  def initialize(@name, @age)
+  end
+end
+
 class StrictJSONAttrPerson
   include JSON::Serializable
   include JSON::Serializable::Strict
@@ -352,6 +362,12 @@ describe "JSON mapping" do
   it "parses array of people" do
     people = Array(JSONAttrPerson).from_json(%([{"name": "John"}, {"name": "Doe"}]))
     people.size.should eq(2)
+  end
+
+  it "works with class with two fields" do
+    person1 = JSONAttrPersonWithTwoFieldInInitialize.from_json(%({"name": "John", "age": 30}))
+    person2 = JSONAttrPersonWithTwoFieldInInitialize.new("John", 30)
+    person1.should eq person2
   end
 
   it "does to_json" do
--- a/src/json/serialization.cr
+++ b/src/json/serialization.cr
@@ -117,7 +117,7 @@ module JSON
 
       def self.new(pull : ::JSON::PullParser)
         instance = allocate
-        instance.initialize(pull: pull)
+        instance.initialize(pull_for_json_serializable: pull)
         GC.add_finalizer(instance) if instance.responds_to?(:finalize)
         instance
       end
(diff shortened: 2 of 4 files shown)


spec/std/json/serializable_spec.cr
```diff
@@ -35,6 +35,16 @@ class JSONAttrPerson
   end
 end
 
+struct JSONAttrPersonWithTwoFieldInInitialize
```
Sija: Add a test for single field too.

RX14:
Please restore the `*` forcing them to be named arguments. It'd also be nice to rename these `pull_for_json_serializable` so it's almost impossible to accidentally hit these methods. These initialize methods aren't used by anything but the macro-generated `.new` methods so they should be as hard to call as reasonable.

kostya:
i actually have some another problems with my code, after this commit, trying to understand why

RX14:
@kostya any progress?

kostya:
seems bug in my project not related to this commits, it happens on crystal master also, but hard to find why. 
```
in /Users/kostya/projects/crystal/src/json/serialization.cr:135: instance variable '@json_unmapped' of MyClass+ was not initialized in this 'initialize', rendering it nilable
```

kostya:
bisect to ee6ec507aa84b3a785d5a7a9eff9435db77e5d47, here problem with @json_unmapped appears.

src/yaml/serialization.cr
```diff
@@ -138,7 +138,9 @@ module YAML
       end
     end
 
-    def initialize(ctx : YAML::ParseContext, node : ::YAML::Nodes::Node, dummy : Nil)
+    def initialize(*, context_for_yaml_serializable : YAML::ParseContext, node_for_yaml_serializable : ::YAML::Nodes::Node)
+      ctx = context_for_yaml_serializable
```
luislavena: You could avoid the local assignations by using external/internal naming of the parameters:

```crystal
def foo(*, external_long_name name : String, external_long_age age : Int32)
  {name, age}
end

foo(external_long_name: "John", external_long_age: 30) # => {"John", 30}
```

src/json/serialization.cr
```diff
@@ -117,7 +117,7 @@ module JSON
 
       def self.new(pull : ::JSON::PullParser)
         instance = allocate
-        instance.initialize(pull, nil)
+        instance.initialize(pull_for_json_serializable: pull)
```
luislavena: Might I recommend the usage of double underscore to name the parameters? Macro variables get converted into `__temp_1`, perhaps we should use something similar to reference the private-yet-public interface?

src/yaml/serialization.cr
kostya: What? never see this before

spec/std/json/serializable_spec.cr
kostya: there is many classes with 1 arg already

src/json/serialization.cr
Sija: IMO that's too much. There's nothing like that used in stdlib ATM.

luislavena: Another option then will be to make initialize private, which can then only be called by this particular `.new` constructor 🤷‍♂️ , thus discouraging externals call to it.

asterite:
@bcardiff Maybe this would be nice to have for 0.26.1

PR #6459, opened 2018-07-27
Adds `IO::Memory.truncate()`

chris-huxtable:
Truncates the contents of IO::Memory at the current cursor position.

### DIFF
--- a/spec/std/io/memory_spec.cr
+++ b/spec/std/io/memory_spec.cr
@@ -226,6 +226,16 @@ describe IO::Memory do
     io.gets(1).should eq("d")
   end
 
+  it "can seek and truncate" do
+    io = IO::Memory.new "hello world"
+    io.seek(-6, Seek::Current)
+    io.pos = 0
+    io.gets().should eq("hello")
+    io << " friend"
+    io.pos = 0
+    io.gets().should eq("hello friend")
+  end
+
   it "clears" do
     io = IO::Memory.new
     io << "abc"
--- a/src/io/memory.cr
+++ b/src/io/memory.cr
@@ -336,6 +336,26 @@ class IO::Memory < IO
     @pos = value.to_i
   end
 
+
+  # Sets the current bytesize to be equal to the current position
+  # effectively truncating the contents at the current position.
+  #
+  # Note this does nothing if the curront position is at the end.
+  #
+  # ```
+  # io = IO::Memory.new "hello world"
+  # io.seek(-6, Seek::Current)
+  # io.truncate()
+  # io.pos = 0
+  # io.gets # => "hello"
+  # ```
+  def truncate() : Nil
+    check_open
+    check_resizeable
+    @bytesize = @pos if @bytesize != @pos
+  end
+
+
   # Yields an `IO::Memory` to read a section of this `IO`'s buffer.
   #
   # During the block duration `self` becomes read-only,


spec/std/io/memory_spec.cr
```diff
@@ -226,6 +226,16 @@ describe IO::Memory do
     io.gets(1).should eq("d")
   end
 
+  it "can seek and truncate" do
+    io = IO::Memory.new "hello world"
+    io.seek(-6, Seek::Current)
+    io.pos = 0
+    io.gets().should eq("hello")
```
Sija: Please, stop using empty parentheses (ditto below)! they're anti-pattern in Crystal...

chris-huxtable: I have tried to stop using my own style in PR's to Crystal. Sometimes I forget about some. I will correct it now.

Sija: Ruby & Crystal hate needless `()` ;)

src/io/memory.cr
```diff
@@ -336,6 +336,24 @@ class IO::Memory < IO
     @pos = value.to_i
   end
 
+  # Sets the current bytesize to be equal to the current position
+  # effectively truncating the contents at the current position.
+  #
+  # Note this does nothing if the curront position is at the end.
```
Sija: `Note` -> `NOTE: `
`curront` -> `current`

src/io/memory.cr
```diff
@@ -336,6 +336,24 @@ class IO::Memory < IO
...
+  # io = IO::Memory.new "hello world"
+  # io.seek(-6, Seek::Current)
+  # io.truncate
+  # io.pos = 0
+  # io.gets # => "hello"
+  # ```
+  def truncate : Nil
+    check_open
+    check_resizeable
+    @bytesize = @pos if @bytesize != @pos
```
straight-shoota: The `if` doesn't have any purpose...

chris-huxtable: My perspective was a comparison is faster then a set, so don't set it if there is no difference.

chris-huxtable: You know what, fair. A low cost applied every time to speed up bad implementations doesn't make sense.

spec/std/io/memory_spec.cr
```diff
@@ -226,6 +226,16 @@ describe IO::Memory do
     io.gets(1).should eq("d")
   end
 
+  it "can seek and truncate" do
+    io = IO::Memory.new "hello world"
+    io.seek(-6, Seek::Current)
```
r00ster91: CI fails because you need to use `IO::Seek::Current` here.

chris-huxtable: Already fixed just waiting for specs to finish running. There are actually a few mistakes.

ysbaddaden:
Use cases?

chris-huxtable:
I originally was using it as a monkey patch for my console singleton. Where I needed to add an IO equivalent to "\b \b" for buffering. This is actually more useful then that.

src/io/memory.cr
RX14: The comparison will bring the data into the L1d CPU cache anyway, and considering modern superscalar architectures, i'm sure they optimize a no-op memory write to L1d to have no effect on cache coherency.

I wouldn't be surprised if this is faster without the comparison.

RX14:
This is fine, but there's already a `File#truncate`, and that method takes a size and a default which is zero. It'd be good to keep these two method's signatures they same considering they're both `IO`s.

chris-huxtable:
It makes no sense to have a negative size. I will be making a negative value similar to a negative value in a range. In other words

```crystal
io =IO::Memory.new()
io << "hello world"
io.truncate(-6)
io.pos = 0
io.gets().should eq("hello")
```

src/io/memory.cr
```diff
@@ -336,6 +336,23 @@ class IO::Memory < IO
     @pos = value.to_i
   end
 
+  # Truncates the memory to the specified *size*.
+  #
+  # Note: a negative value truncates from the end of the `IO::Memory`.
```
Sija: `Note` -> `NOTE`

chris-huxtable: Its used in both ways throughout the project. I will use the caps as it doesn't really matter to me and seems more consistently applied.

RX14:
I really want `File#truncate` and `IO::Memory#truncate` to behave as similar as possible. This means being able to extend it past the current size by padding with zeroes. Also supporting negative sizes in `File#truncate`.

chris-huxtable:
I can understand the negative values, I have already implemented them for `IO::Memory` and will for `File` but I am not a fan of using `truncate` to extend the size. Perhaps a method like `extend` or `resize`?

asterite:
Note that setting `pos` of an `IO::Memory` past its size, and then writing to it, will fill the gap with `\0`. This is similar to what happens with a `File`, and with Ruby's `StringIO`. So these behave the same in Crystal:

```crystal
# Using a file
io = File.open("test.txt", "w+")
io.pos = 3
io.puts 1

io.rewind

p io.gets_to_end # => "\u0000\u0000\u00001\n"

io.close

# Using an IO::Memory
io = IO::Memory.new
io.pos = 3
io.puts 1

io.rewind

p io.gets_to_end # => "\u0000\u0000\u00001\n"

io.close
```

Truncating a file past its size (or well, past its maximum current size) fills that with `\0`:

```crystal
io = File.open("test.txt", "w+")
io.truncate(5)
io.puts 1

io.rewind

p io.gets_to_end # => "1\n\u0000\u0000\u0000"

io.close
```

and the same should happen with `IO::Memory`.

By the way, Ruby's `StringIO` implements `truncate` and also behaves like that.

chris-huxtable:
While working on this I noticed the growth factor is 2, which is arguably the worst choice. Based on [this article](https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md#memory-handling) and what [`NSMutableArray` uses internally](http://ciechanowski.me/blog/2014/03/05/exposing-nsmutablearray/) I propose 1.625.

@asterite, @ysbaddaden - Thoughts?

ysbaddaden:
No opinion, that depends on the memory allocator and growth. With my GC you'll never reuse the previous allocation anyway.

asterite:
Please revert the growth factor change. You just made your PR unmergable because we now must discuss two things instead of one. Please open a new issue or PR to discuss the growth factor change, which should also be applied to arrays, maybe.

chris-huxtable:
I will revert the growth factor and do some benchmarking. If that all works out I’ll open a new PR.

chris-huxtable:
Ready for review.

src/io/memory.cr
```diff
@@ -336,6 +336,27 @@ class IO::Memory < IO
     @pos = value.to_i
   end
 
+  # Truncates the memory to the specified *size*.
+  #
+  # NOTE: a negative value truncates from the end of the `IO::Memory`.
+  def truncate(size = 0) : Nil
```
straight-shoota: I think `size` should be restricted to Int, or better even Int32.

spec/std/io/memory_spec.cr
```diff
@@ -226,6 +226,44 @@ describe IO::Memory do
...
   end
 
+  it "can truncate to size" do
+    io = IO::Memory.new
+    io << "hello world"
+    io.truncate(5)
+    io.to_s.should eq("hello")
+  end
+
+  it "raises when truncate size is too large" do
```
straight-shoota: This description is incorrect. There is no raise.

spec/std/io/memory_spec.cr
```diff
@@ -226,6 +226,44 @@ describe IO::Memory do
...
+    io = IO::Memory.new
+    io << "hello world"
+
+    io.truncate(12)
+    io.truncate(13)
+
+    io.to_s.should eq("hello world\u0000\u0000")
```
straight-shoota: Should truncating with `size > bytesize` really fill up with zero bytes?
This is not truncating but expanding and the semantic difference is only governed by the current bytesize.
I'd rather have it fault tolerant and don't grow bytesize on truncate.

chris-huxtable:
@straight-shoota - Thanks for the review. I am absolutely slammed with work ATM. I won't be able to take a stab at this for a little while.

spec/std/io/memory_spec.cr
didactic-drunk: `bytesize` growing on truncate matches the behavior of `File`.

Issue #2968, opened 2016-07-07
Make `/` be float division


asterite:
Similar to Python 3, we could have `/` return a float, and add another operator, maybe `//` (though that means an empty regex right now, but we could alternatively use `%r()` for that).

So:

``` cr
# Now
1 / 2 # => 0

# With this change
1 / 2  # => 0.5
1 // 2 # => 0
```

I'm not 100% sure about this, mainly because:
- what type should the result be? `Float64`?
- `x /= 2` changes the type of `x`, while previously it doesn't, but one could use `x //= 2` now

I'm cc-ing @pbrusco because he originally proposed this :-)

Thoughts?

Nephos:
Hello,
Usually I take care of my number's type, so I don't think adding a new operator like this one is useful.
Furthermore, Ruby does not provide such operator. I hope crystal will stay close to Ruby and away from python ;)

pbrusco:
Of course I agree on changing, keeping an ambiguous division causes lots of headaches. Lots. I've seen people struggling with results and then finding out that in that language. I'm sure that division related bugs are everywhere and I think this is the way to avoid them. 

If performance is needed => // 
If precision is needed => something else (a method)

@Nephos, I think crystal will be better than ruby, but it needs the "right" division :P 

Here are the reasons (in the context of python): https://www.python.org/dev/peps/pep-0238/. 

@asterite, Thanks you for opening the discussion

bcardiff:
In crystal the user can have fine control of the numerical types. But when receiving just a number, might not be clear.

Having / to be floating point division, and // been the floor of the division would mean to have a unified semantic of the operation despite the numerical types. This favor more concistency IMO.

As for the type, Float64 by default sounds reasonable, unless Float32 is involved and no Float64. I think that will honor f32 operations for when the user cares.

ozra:
You mean something like:?
```
F32#/(Number) -> F32
Number#/(Number) -> F64
```

asterite:
I think the PEP has one very good reason to do this:

```cr
def average(elements)
  elements.sum / elements.size
end
```

The average should be a float, but this might give an integer back. We can solve this using `fdiv`:

```cr
def average(elements)
  elements.sum.fdiv(elements.size)
end
```

So right now we have:
* **`/` is "floor division" or "float division", depending on the type of the first argument**
* `fdiv` is always `"float division"

I think that's the main problem: that `/` has two meanings, depending on the type of the first argument. We could make it so that:

[...]

PR #6470, opened 2018-07-30
Parse integer division operators

bcardiff:
In the path to implement #2968 first the operators needs to be allowed.

This PR allows `//` and `&//` operators to be defined. 
The behavior will need to wait to the next release. 
`//` will be allowed to overflow and `&//` won't. Similar to `+` vs `&+`.

`//` will continue to mean empty regex when possible:
* `foo // bar` is equivalent to `foo.//(bar)`
* `foo //` is a syntax error (from now on)
* `foo(//)` is a call passing empty regex

The `ASTNode#to_s` will use `%r()` as empty regex always since `//` will not always work.

There are some refactors on specs and code for allowing operators to be used as methods and macro names introducing a state ivar `@wants_def_or_macro_name` that works similar to `@slash_is_regex` configuring the behavior of the lexer depending on the context.


bcardiff:
Yes @straight-shoota I mean to overflow as "raise overflow". `&//` won't raise. I've just edited for clarity.

RX14:
> `&//=`

we get closer to perl every day

RX14:
I really don't like all these new operators being proposed. `Wrapped(T)` will work fine, and integer division works fine as-is. Why change?

I really don't think we should introduce all these operators, they're just more things for a newcomer to learn to save a few characters typing (`.to_f`) in a fairly rare situation. I think it's a bad tradeoff.

bcardiff:
I would expect that from all the operators added in this PR `//` will be at far the most used. The rest are for consistency. Implementing only `//` would leave things unfinished in this topic.

The division operator `/` can be improved, this has been discussed already in #2968 and #6223. The main issue is semantic one. `/` performs different operators depending on the numeric type. The fact that we are used to an overload behaviour of `/` does not mean it can't be improved.

Adding `.to_i`/`.to_f` to force one or another will make the code use more (or less) precision that it should because `.to_i`/`.to_f` has fixed output type.

The wrapper api, can be implemented on top of this. But not the other way around. So operators are more powerful and expressive. 

For wrapper to work well with overloads you will need `Wrapper(T) < T` which is something that is not supported. Otherwise the type is used only locally without passing them around.

The discussion left (hence this PR) is not about if the language needs to add this, but if the implementation is good enough.

RX14:
`Wrapping(T)` is a discussion for another issue, but i'm yet to see a place where numbers with wrapping arithmetic performed on them are actually passed around significantly outside of a single class. I don't see a reason why ease of use of unchecked arithmetic is a big deal because it's just so rare. Adding a whole set of new operators just for something so rarely used seems to be like a gross premature optimization and a bad idea.

Back to this issue, I'm not sure what your concern is about `.to_i`/`.to_f`, because the current method clearly works, because it's the current method. I haven't seen *any* friction in that area and I've been on IRC helping newbies for a long time. I can't see the motivation.

RX14:
I just want data that this is actually a problem in the real world, and not a solution to a problem that we've invented in our heads and I can't find any.

bcardiff:
> Back to this issue, I'm not sure what your concern is about .to_i/.to_f, because the current method clearly works, because it's the current method.

Let's use an example, if `avg` method wants to be coded that will work with any kind of numbers you will end up adding a `to_f` to force non-integer division, because `/` currently is both (int and non-int). I think that that already needs to be improved because it can be more clear.

The `to_f` will make the result of the operation `Float64`, but if the input numbers are `Float32` then why the operation needs to be carried on 64 bits? Only because the lack `/` to express the floating point division in the most appropriate type given the operands.

Again, the fact that we are used to C like math operators does not make it the best option.

RX14:
@bcardiff I think issues like that are fairly rare, and `val = val.to_f unless val.is_a? Float` is a very short snippet for the small amount of methods that operate on generic `Number`s.

bcardiff:
That snippet is a workaround and harder to come up when learning the language. Having different operators is cleaner. Something new to learn today, yes. But is a more consistent set of operators. 

Try to explain that line when coding a method that manipulate numbers vs explain that `/` is floating point division and `//` is floor division always.

1. Splitting `/` and `//` was already agreed and pros/cons were exposed.
2. Leaving `//` to raise upon overflow and avoid adding `&//` seems incomplete.
3. Excluding `//=` and `&//=` when `/=` is already supported seems incomplete also.

I _could_ understand removing 3 from the PR, if and only if `/=` is also removed. But it seems arbitrary since `+=`, `&+=`, etc are allowed. So they are implemented.

RX14:
@bcardiff it's not just something new to learn today, it's something new that every newcomer from now until forever will have to learn.

At the end of the day though, I'm more concerned about the overflow stuff adding a bazillion more operators than this PR only adding two.

asterite:
- I think having / and // operators is more intuitive than just / (many people find it strange that 1/2 == 0)
- The & operators will be used in very specific circumstances, so this doesn't add a lot of complexity. Much less with division.
- I think the decision about introducing this // operator is already taken. Now it's time to review the code.

src/compiler/crystal/syntax/to_s.cr
```diff
@@ -985,14 +985,20 @@ module Crystal
...
 
     def visit(node : RegexLiteral)
-      @str << '/'
-      case exp = node.value
-      when StringLiteral
-        Regex.append_source exp.value, @str
-      when StringInterpolation
-        visit_interpolation(exp) { |s| Regex.append_source s, @str }
+      if (exp = node.value).is_a?(StringLiteral) && exp.value.empty?
+        # // is not allways an empty regex, sometimes is an operator
```
r00ster91: Typo: allways -> always

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -287,8 +288,17 @@ module Crystal
         line = @line_number
         column = @column_number
         char = next_char
-        if !@slash_is_regex && char == '='
+        if (@wants_def_or_macro_name || !@slash_is_regex) && char == '/'
+          case next_char
```
j8r: Can be replaced by a smaller `if` statement

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -618,6 +644,18 @@ module Crystal
           else
             @token.type = :"&*"
           end
+        when '/'
+          case next_char
+          when '/'
+            case next_char
```
j8r: ditto

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -470,6 +489,13 @@ module Crystal
             else
               symbol "&*"
             end
+          when '/'
+            case next_char
```
j8r: ditto

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -412,7 +426,12 @@ module Crystal
             symbol "*"
           end
         when '/'
-          next_char_and_symbol "/"
+          case next_char
```
j8r: ditto

spec/compiler/lexer/lexer_spec.cr
```diff
@@ -1,8 +1,11 @@
 require "../../support/syntax"
 
-private def it_lexes(string, type)
+private def it_lexes(string, type, *, slash_is_regex : Bool? = nil)
   it "lexes #{string.inspect}" do
     lexer = Lexer.new string
+    unless (v = slash_is_regex).nil?
```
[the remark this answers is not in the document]
straight-shoota: This is not the same. `slash_is_regex` is `Bool?` so you need to test for `.nil?`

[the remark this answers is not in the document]
bcardiff: That would be abusing that the default value of slash_is_regex is false.

src/compiler/crystal/syntax/lexer.cr
bcardiff: Using the case is more about following the pattern of the rest of the lexer.
But I don't follow which is the proposed smaller `if`.

src/compiler/crystal/syntax/lexer.cr
bcardiff: ditto ditto :-)

src/compiler/crystal/syntax/lexer.cr
j8r: This is fine - this just makes the lexer looks a bit more complex that it is, that's it :)
Here it would be
```crystal
if next_char == '/'
  next_char :"//="
else
  @token.type = :"//"
end
```

src/compiler/crystal/syntax/lexer.cr
j8r: It would be here
```crystal
if next_char == '/'
  next_char_and_symbol "//"
else
  symbol "/"
end
```

src/compiler/crystal/syntax/lexer.cr
j8r: With `if`s it would be (start at line 648, replacing the 2 `case`s):
```crystal
raise "&/ is not a valid operator. Did you mean &// ?" if next_char != '/'
if next_char == '='
  next_char :"&//="
else
  @token.type = :"&//"
end
```

RX14:
I forgot we'd already merged all the `&+` overflow operators. Since we've already - in my opinion - ruined the language with them you might as well go and merge this PR.

src/compiler/crystal/syntax/lexer.cr
j8r: And here it would be:
```crystal
if next_char == '/'
  next_char_and_symbol "&//"
else
  symbol "&/"
end
```

src/compiler/crystal/syntax/lexer.cr
j8r: At a note, the previous code was:
```crystal
case next_char
when '/'
  case next_char
  when '='
    next_char :"&//="
  else
    @token.type = :"&//"
  end
else
  raise "&/ is not a valid operator. Did you mean &// ?"
end
```

oprypin:
I would prefer `&//=` to be removed, along with `&+=`  `&-=`  `&*=`. If someone wants to be this specific, they can write `a = a &+ 1` rather than `a &+= 1`.

And if anyone else is thinking: how can integer division overflow? 🤔
@RX14 had to explain to me that it happens (only?) with `Int32::MIN / -1`

src/compiler/crystal/syntax/lexer.cr
bcardiff: I thought you ment the `if (@wants_def_or_macro_name ...`.
There are some `if` in the lexer, bust mostly `case` even single `when`-ed (?) case. I found them easier to read the lexer if all the structures are the same and not case by case based that we will end up having if your suggestions are applied.

At there is no real change the the actual code.

So `case` are kept. But thanks.

bcardiff:
I fixed the typo and rebased. :shipit: !

@oprypin I will still raise the card of completeness regarding the `&+=` and others. And they will be actually used in Hasher and others. All the `op=` are syntax sugar and not a core part of the language.

asterite:
By the way: https://stackoverflow.com/questions/30378591/overflow-division-operator-not-recognized

I would recommend not having the &/ operator in Crystal, just like in Swift. It's an extra check for something that can rarely happen.

RX14:
completeness for completeness's sake seems like a very bad idea to me. It just expands the set of operators.

I'd be *much* happier if the only operators we were adding were `&+`, `&-`, `&*`, `//` and `//=`.

bcardiff:
Since division by zero would still be raised in `&//` is not that that operator is free of exception. 
Ok, I will remove `&//` and `&//=`.

PR #6476, opened 2018-08-01
Add API docs for pp, p! and pp! macro methods

straight-shoota:
Follow up of #6374

### DIFF
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -34,12 +34,24 @@ module Crystal::Macros
   def flag?(name) : BoolLiteral
   end
 
-  # Prints an AST node at compile-time. Useful for debugging macros.
-  def puts(expression) : Nop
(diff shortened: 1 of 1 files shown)


src/compiler/crystal/macros.cr
```diff
@@ -34,12 +34,24 @@ module Crystal::Macros
   def flag?(name) : BoolLiteral
   end
 
-  # Prints an AST node at compile-time. Useful for debugging macros.
-  def puts(expression) : Nop
+  # Prints AST nodes at compile-time. Useful for debugging macros.
+  def puts(*expressions) : Nop
   end
 
   # Same as `puts`.
```
Sija: `# :ditto:`?

straight-shoota: I'm not sure that's really worth it for three short words...

Sija: umm, u r right.

image

PR #6477, opened 2018-08-02
Add closed state to HTTP::Server::Response

### LINKED ISSUE #6465: Http::Server::Response should have closed state
Calling `Server::Response#close` delegates to `Server::Response::Output#close` and it has some side effects (for example if no headers are written) but doesn't change any state. After `response.close` was called, `response.close?` should return `true`.

`Output` is a buffered wrapping of the transport IO (typically a socket). It doesn't delegate `#close` to the underlying socket, which is correct because the transport can be re-used. But the specific response should be closed after calling `close` (this also means write methods should honor closed state).

A practical use case is shown in kemalcr/kemal#475 where a response shouldn't be handled if it was already closed before.

straight-shoota:
Fixes #6465

### DIFF
--- a/src/http/server/response.cr
+++ b/src/http/server/response.cr
@@ -104,6 +104,11 @@ class HTTP::Server
       @output.close
     end
 
+    # Returns `true` if this response has been closed.
+    def closed?
+      @output.closed?
+    end
+
     # Generates an error response using *message* and *code*.
     #
     # Calls `reset` and then writes the given message.
@@ -154,6 +159,7 @@ class HTTP::Server
         @sync = false
         @flush_on_newline = false
         @chunked = false
+        @closed = false
       end
 
       private def unbuffered_read(slice : Bytes)
@@ -182,6 +188,10 @@ class HTTP::Server
         end
       end
 
+      def closed?
+        @closed
+      end
+
       def close
         unless response.wrote_headers?
           response.content_length = @out_count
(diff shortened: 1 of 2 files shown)


src/http/server/response.cr
```diff
@@ -182,6 +188,10 @@ class HTTP::Server
         end
       end
 
+      def closed?
+        @closed
+      end
+
```
jhass: I guess this should be checked in `unbuffered_write`?

straight-shoota: `IO::Buffered#write` already calls `check_open`. This is ensured with the `expect_raises` expectation.

src/http/server/response.cr
```diff
@@ -154,6 +159,7 @@ class HTTP::Server
         @sync = false
         @flush_on_newline = false
         @chunked = false
+        @closed = false
```
straight-shoota: @asterite This is in `#reset`. I'm not sure how useful this is, because when a request is closed, at least the headers has already been written to the underlying IO. This effectively can't be reset because the response is already sent.

asterite: Oh, cool. Reset is useful, can't remember now why, but response is reused for keep alive, check request processor.

straight-shoota: Oh yeah, `RequestProcessor` resets the response to avoid allocating a new instance for each iteration.

straight-shoota: I'll add a comment about this.

Issue #6011, opened 2018-04-25
Redesign HTTP::Client to implement missing features


straight-shoota:
`HTTP::Client` is missing some quite essential feature, and while there are some individual issues about that, it is better to discuss necessary design changes in a central place.

Some missing features:

* Proxy support #2963
* Connecting to Unix socket #2735
* Transparently re-use connections to different servers
* HTTP/2 #2125
* Mock connections (for testing)
* Following redirects #2721
* Session handling #3081
* Integration of custom middlewares (for example authentication)
* Timeout

The first group is about adding more flexibility to use different types of connections or enhance their usage. This needs to separate the implementation for establishing a connection from the `HTTP::Client`.
#6001 is an incomplete experiment for that.

The concept of using a configurable transport is employed by many other implementations of HTTP clients:

* Go: [RoundTripper](https://golang.org/pkg/net/http/#RoundTripper) (interface `request -> response`) and [Transport](https://golang.org/pkg/net/http/#Transport) (implementation, including proxy, reusing connections, HTTP/2 and a lot of basic HTTP workflow like redirects).
* Java: [com.google.api.client.http.HttpTransport](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/http/HttpTransport): builds a request (`method, url -> request` ) and `request.execute()` gets the response.
* Ruby: [Faraday](http://www.rubydoc.info/gems/faraday/Faraday/Adapter) implements a middleware interface similar to `HTTP::Handler`.
* Ruby: [Hurley](https://github.com/lostisland/hurley#connections) has a `connection` property which is a lambda `request -> response`

It makes great sense to have a `HTTP::Client` as a high-level interface delegate the heavy-lifting to a replaceable low-level API. User should be able to configure `HTTP::Client` to use a specific transport.

The default transport should allow to connect to any TCP server and maintain a connection pool to re-use idle sockets, even beyond the life-span of a single client instance. Other transports could connect to a proxy, a Unix socket or just mock a connection for testing purposes.

It is a question how the interface between `HTTP::Client` and transport should be designed. In the experimental PR #6001 the transport returns an IO for a specific request, but the client still writes the request and reads the response.
[...]

PR #6481, opened 2018-08-04
Set default host in HTTP::Client#exec

### LINKED ISSUE #6480: HTTP::Client#exec causing some "400 Bad Request" response when using it with HTTP::Request object.
The following code triggers a 400 response from example.com:

```crystal
require "http/client"

client = HTTP::Client.new "www.example.com"
req = HTTP::Request.new("GET", "/")
resp = client.exec HTTP::Request.new("GET", "/")
puts resp.status_code # 400
```

Note that `HTTP::Client#exec` seems to be generating an HTTP request without the `Host` header.  Adding the missing header will trigger expected 200 response.

```crystal
req2 = HTTP::Request.new("GET", "/", HTTP::Headers.new.tap{|h| h["Host"] = "example.com"})
resp2 = client.exec req2
puts resp2.status_code # 200
```

Class method is working as expected:
```crystal
puts HTTP::Client.exec("GET", "http://www.example.com/").status_code # 200
```

I also noticed that this didn't seem to be a problem on google.com
```crystal
client2 = HTTP::Client.new "www.google.com"
resp3 = client2.exec(req)
puts resp3.status_code # 200
```

The reason I ran into this problem was because I was trying to send the same `Request` object to multiple Hosts.  I expect I should be able to reuse the same request object cross multiple client / hosts.  Is this an acceptable usage pattern of the HTTP::Client API? 

```
crystal -v
Crystal 0.25.1 (2018-06-30)

LLVM: 5.0.2
```

straight-shoota:
Fixes #6480

### DIFF
--- a/spec/std/http/client/client_spec.cr
+++ b/spec/std/http/client/client_spec.cr
@@ -22,6 +22,12 @@ private class TestServer < TCPServer
   end
 end
 
+private class TestClient < HTTP::Client
+  def set_defaults(request)
+    super
+  end
+end
+
 module HTTP
   describe Client do
     typeof(Client.new("host"))
@@ -206,6 +212,19 @@ module HTTP
         client = Client.new("localhost", server.local_address.port)
         client.connect_timeout = 0.5
         client.get("/")
+      end
+    end
+
+    describe "#set_defaults" do
+      it "sets default Host header" do
+        client = TestClient.new "www.example.com"
+        request = HTTP::Request.new("GET", "/")
+        client.set_defaults(request)
+        request.host.should eq "www.example.com"
+
+        request = HTTP::Request.new("GET", "/", HTTP::Headers{"Host" => "other.example.com"})
+        client.set_defaults(request)
+        request.host.should eq "other.example.com"
       end
     end
   end
--- a/src/http/client.cr
+++ b/src/http/client.cr
@@ -562,6 +562,7 @@ class HTTP::Client
   end
 
   private def set_defaults(request)
+    request.headers["Host"] ||= host_header
     request.headers["User-Agent"] ||= "Crystal"
     {% if flag?(:without_zlib) %}
       false
@@ -651,9 +652,7 @@ class HTTP::Client
   end
 
   private def new_request(method, path, headers, body : BodyType)
-    HTTP::Request.new(method, path, headers, body).tap do |request|
-      request.headers["Host"] ||= host_header
-    end
+    HTTP::Request.new(method, path, headers, body)
   end
 
   private def socket


asterite:
Just note that even though this is a good change, the original issue is not solved: the use case was to send a request to multiple hosts by reusing the request instance. Because the host is only set if it's not set already, that won't work.

Maybe we should remove the exec methods from being public. I don't think there's a sane way to reuse request objects, unless we dup/clone them in exec, but that's of course not ideal.

straight-shoota:
Yeah, I figured as much. But the original issue can be solved by dup-ing the request when providing it to `exec`: `client.exec(request.dup)`.

Obviously, this is not ideal, but it works fine for that use case. I'd suggest to merge this fix and continue with #6011 to redesign HTTP client.

asterite:
I would recommend then documenting this behavior, mentioning that reusing an `HTTP::Request` shouldn't be done because the `Host` header is set once per request. That's why I would maybe prefer to hide the `exec` method, and so there's no way to misuse the API.

straight-shoota:
An alternative solution would be to modify `HTTP.serialize_headers_and_body` to insert default `host` argument provided as argument if it is missing from headers.

z64:
@asterite re: removing `exec(request : HTTP::Request)`

For what it's worth, I've made good use of this method in libs. It's been helpful to construct a namespace of `HTTP::Request` factory methods that my own `Client` implements in order to spec/unit test the requests nicely against API documentation without having to actually execute them,  and where building a mock API service would be much more expensive (as in time consuming) task, and this is just as likely to ensure I've written working runtime code.

Basically:
```cr
# requests.cr
module Request
  extend self

  def get_resource(id : ID)
    HTTP::Request.new("GET", "/resource?id=#{id}")
  end
end

# client.cr
class Client
  @http_client : HTTP::Client

  def exec(request : HTTP::Request)
    @http_client.exec(request)
  end

  def get_resource(id : ID)
    response = Request.get_resource(id)
    Resource.from_json(response.body)
  end
end

# request_spec.cr
it "#get_resource" do
  request = Request.get_resource(1)
  request.method.should eq "GET"
  request.path.should eq "/resource"
  request.query.should eq "id=1"
end
```
I could replace this with my own types if removed with a slightly more complicated interface, but having `exec(request : HTTP::Request)` has been helpful in this regard, where things like rspecs `expect(api_client).to receive(:get_resource).with(resource_id)` doesn't exist in crystal

Maybe mock connections via #6011 would provide a equally simple solution, but that's still an abstract bullet point on the list, so I don't know

newtonapple:
Correct me if I'm wrong, by removing `HTTP::Client#exec(request : HTTP::Request)` from the public API, it will effectively eliminate all public use cases / needs for `HTTP::Request`, correct?  I personally think having a reusable `HTTP::Request` can be quite useful a number of scenarios.

RX14:
A reusable `HTTP::Request` doesn't make sense because the body should be an `IO`, which is consumed on use.

PR #6484, opened 2018-08-04
Revert restrictict `exp.@var` to types in the same hierarchy

bcardiff:
This reverts #6358 to avoid an unwanted breaking change in the next release but keeps two of the refactors introduced in that PR.

* Refactor: Char & IO related non UTF-8 encoding check
* Refactor: Crystal::Call#check_visibility and extract type methods


bcardiff:
@asterite I just think it's easier to read

Issue #5938, opened 2018-04-10
Improve interface of Tempfile methods


straight-shoota:
`Tempfile` has two different methods for generating a temporary file/path which work differently:

* `Tempfile.new` (and `.open`) uses a platform-specific implementation (based on `LibC.mkstemps`) to create a temporary file.
* `Tempfile.tempname` is implemented entirely in Crystal and basically just builds a path from a randomized string. It does not actually create a file. That's useful if your code does not interact with the file as `IO`, and especially for creating temporary directories.

`.new` and `.open` take arguments `name`/`filename` and `extension` while `.tempname` only accepts `extension`. It would make sense if all methods had the same arguments (apart from other, purpose-specific methods such as `encoding`). And I think `.tempname` should also have a parameter for specifying a prefix to the path as it is equally purposeful for use cases where a file is not directly created.

`extension` (in all cases) is essentially a suffix, `name` and `filename` a prefix. All of them could even contain additional path components. As such, the names might be misleading. But I like `extension` because that's the most probable use case for it anyway.

`name` could perhaps be renamed to `prefix`... but then you might assume `suffix` as well? But it's still better than `name` IMO.
I don't think it should be a required argument (as with `new` and `open` currently). You might just need a temporary file/path and don't care about anything in the name. I see no need to enforce a prefix.

As for the argument ordering, I propose to stick to the natural sequence `prefix, suffix` and make both optional in all methods. That would lead to the following signatures:

* `Tempfile.new(prefix = nil, extension = nil, encoding = nil, invalid = nil)`
* `Tempfile.open(prefix = nil, extension = nil)` (shouldn't this have encoding and invalid as well?)
* `Tempfile.tempname(prefix = nil, extension = nil)`

This also affects #5937 and #4096

asterite:
Probably can be closed if we go with #5939

straight-shoota:
Both issues are related but discuss individual topics. This could still affect the implementation of #5939, even if the methods will have a different scope then. At least I'd prefer to change everything in one go. 

So it would be great to get some feedback on this issue here! 👍

straight-shoota:
Fixed by #6485

PR #4096, opened 2017-03-02
Feature/mktmpdir

bramswenson:
This is simply a rebase of #2911 to current master (03-01-2017)

PR #6485, opened 2018-08-04
Remove Tempfile

### LINKED ISSUE #5939: Remove class Tempfile
`Tempfile` is actually identical to `File` except for the constructor (which can be put in `File` with a different name), a few class methods (where this is also true), and the instance methods `delete` and `unlink` which are just aliases to `File.delete(@path)`.

I don't see a point in having a special type for this as the behaviour is identical to `File`, it's just an alternate way of initializing the instance.

The class methods/constructors could be renamed as followed:

* `Tempfile.new` -> `File.tempfile`
* `Tempfile.open` -> `File.tempfile` (yield)
* `Tempfile.dirname` -> `Dir.tempdir`
* `Tempfile.tempname` -> `File.tempname`

This issue is loosely related to #5938 about adjusting the arguments to the class methods/constructors but that's independent of the actual names.

straight-shoota:
This removes `Tempfile` and places its methods in `File` and `Dir` as proposed in #5939 and standardizes the arguments as proposed in #5938. All tempfile methods now receive `prefix`, `suffix` (and optional `dir`) arguments. There are also overload for only one argument `suffix` (and optional `dir`) because it is pretty common to create a temp file with a specific extension.

Previously, `require "tempfile"` was needed in order to use this feature. Now that it is integrated into `File` I think it makes sense to have it in corelib.

I'm not sure if `tempfile` with block should perhaps actually delete the tempfile. Or if there could be some other mechanism to simplify automatic deletion of tempfiles. This could be discussed in a follow-up issue, though.
Same goes for the addition of a method to create a temporary directory (see #4096). The only issue with this is that `Dir.tempdir` would also be a fitting name for this but it is now used to query the tempdir environment setting.
Related to that, there could also be a future improvement to allow modifying the global tempdir, probably in a way similar to `Time::Location.local`.

Fixes #5939 and #5938


src/file/tempfile.cr
```diff
@@ -0,0 +1,138 @@
...
+  # These values may contain directory separators.
+  #
+  # The file will be placed in *dir* which defaults to the standard temporary directory `Dir.tempdir`.
+  #
+  # *encoding* and *invalid* are passed to `IO#set_encoding`.
+  #
+  # It is the caller's responsibility to remove the file when no longer needed.
+  def self.tempfile(prefix : String?, suffix : String?, *, dir : String = Dir.tempdir, encoding = nil, invalid = nil)
+    fileno, path = Crystal::System::File.mktemp(prefix, suffix, dir)
+    new(path, fileno, blocking: true, encoding: encoding, invalid: invalid)
```
asterite: Why blocking true?

straight-shoota: Not sure... I just copied it form `Tempfile#initialize`.

ysbaddaden:
As said in issues about this, I don't like this breaking change. I prefer to keep concerns distinct, and really see no need to push this into corelib.

straight-shoota:
I think we established that there is little reason for `Tempfile` being a separate type with exactly the same behaviour as `File`.

An alternative option might be to keep `Tempfile` as a module and have these methods defined there, but return an ordinary `File` instance. This would reduce the breaking changes, it would just be that `Tempfile.new` returns `File` instead of `Tempfile` (etc.).

But the changes in the method signatures need to be breaking anyways in order to provide a sane API.

src/file/tempfile.cr
RX14: all files use blocking IO anyway because `O_NONBLOCK` has no effect on files.

straight-shoota:
Rebased and removed `blocking`

straight-shoota:
@ysbaddaden Would you feel more comfortable if we put the tempfile methods back in `Tempfile` namespace (as module) and just return `File`?
This would separate tempfile methods from regular `File` methods. And it wouldn't need to be in core lib.

RX14:
I see no reason why they can't be in the corelib, it's not particularly much more code (the code would be required as part of crystal/system anyway) and it's only dependency is libc. It doesn't drag in anything more to the corelib. This PR is fine as-is.

PR #6280, opened 2018-06-27
Show line numbers at link when there are filename duplicates in "Defined in:" section (docs)

r00ster91:
For example this **Defined in:** section example:

---

## Defined in:
[colorize.cr](https://github.com/crystal-lang/crystal/blob/7fb783f7afec0a199d6f22ef78aae0f45517068a/src/colorize.cr#L79)
[colorize.cr](https://github.com/crystal-lang/crystal/blob/7fb783f7afec0a199d6f22ef78aae0f45517068a/src/colorize.cr#L126)
[foo.cr](foo)
[bar.cr](bar)

---

Would change to this:

---

## Defined in:
[colorize.cr:79](https://github.com/crystal-lang/crystal/blob/7fb783f7afec0a199d6f22ef78aae0f45517068a/src/colorize.cr#L79)
[colorize.cr:126](https://github.com/crystal-lang/crystal/blob/7fb783f7afec0a199d6f22ef78aae0f45517068a/src/colorize.cr#L126)
[foo.cr](foo)
[bar.cr](bar)

---

for more readability.

And this PR also fixes identical link generations like in the **Defined in:** section of [Float32](https://crystal-lang.org/api/0.25.0/Float32.html) where the two [primitives.cr](https://github.com/crystal-lang/crystal/blob/7fb783f7afec0a199d6f22ef78aae0f45517068a/src/primitives.cr#L269) links lead to the same line and file because of macros.

PR #6489, opened 2018-08-05
Fix identical link generation prevention for the "Defined in" section (docs)

r00ster91:
Follow-up of #6280.

Preventing duplicates in types.cr doesn't seems to work. See: [Float32](https://crystal-lang.org/api/master/Float32.html).
But it works now by preventing them in generator.cr.

### DIFF
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -393,6 +393,9 @@ class Crystal::Doc::Generator
       filename = filename[1..-1] if filename.starts_with? File::SEPARATOR
       filename = filename[4..-1] if filename.starts_with? SRC_SEP
 
+      # Prevent identical link generation in the "Defined in:" section in the docs because of macros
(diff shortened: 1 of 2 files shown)


straight-shoota:
Is there an exaplanation *why* the existing solution doesn't work as expected?

straight-shoota:
I guess, in the end it doesn't really matter though. The implementation is equivalent as it applies the correct result to `Crystal::Doc::Type#location`. So, as long as it works, it's fine by me.

r00ster91:
Hmm I guess it's because at the time in the `add_location` method, the `filename` and `line_number` instance variables of the `Location` aren't yet correctly modified. But they are modified now in the `relative_location` before the `any?`. For example here's how `filename` gets modified before the `any?` iteration:
```crystal
filename = filename[1..-1] if filename.starts_with? File::SEPARATOR
filename = filename[4..-1] if filename.starts_with? SRC_SEP
```
And all these modifications of `filename` and `line_number` don't happen before the `any?` iteration in `add_location`. That's why it didn't work in `add_location` I think.

PR #6496, opened 2018-08-06
Fix docs of UUID enums

r00ster91:
The descriptions of the [UUID::Variant](https://crystal-lang.org/api/0.25.1/UUID/Variant.html) and [UUID::Version](https://crystal-lang.org/api/0.25.1/UUID/Version.html) enum members are currently in the wrong order.


Sija:
Would be nice to handle postfix comments, everywhere though, not just for enum members.

PR #6497, opened 2018-08-06
Make Spec compile on win32

RX14:
This makes the spec runner work on win32, and also fixes a few issues in the windows File port caught by the addition of exception support.

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -19,8 +19,8 @@ private def home
   home.chomp('/')
 end
 
-private def it_raises_on_null_byte(operation, &block)
-  it "errors on #{operation}" do
+private def it_raises_on_null_byte(operation, file = __FILE__, line = __LINE__, end_line = __END_LINE__, &block)
+  it "errors on #{operation}", file, line, end_line do
     expect_raises(ArgumentError, "String contains null byte") do
       block.call
     end
@@ -110,7 +110,8 @@ describe "File" do
       end
     end
 
-    it "raises an error when a component of the path is a file" do
+    # TODO: do we even want this?
+    pending_win32 "raises an error when a component of the path is a file" do
       ex = expect_raises(Errno, /Error determining size/) do
         File.empty?(datapath("test_file.txt", ""))
       end
@@ -468,7 +469,8 @@ describe "File" do
       end
     end
 
-    it "raises an error when a component of the path is a file" do
+    # TODO: do we even want this?
+    pending_win32 "raises an error when a component of the path is a file" do
       ex = expect_raises(Errno, /Error determining size/) do
         File.size(datapath("test_file.txt", ""))
       end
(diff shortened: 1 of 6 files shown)


src/spec/dsl.cr
```diff
@@ -1,6 +1,5 @@
 require "colorize"
 require "option_parser"
-require "signal"
```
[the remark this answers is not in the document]
RX14: signal is in the prelude so this require is fine to go anyway, but yeah adding a TODO to the `flag?` is good

spec/std/file_spec.cr
```diff
@@ -967,7 +979,7 @@ describe "File" do
       File.writable?("foo\0bar")
     end
 
-    it_raises_on_null_byte "executable?" do
+    it_raises_on_null_byte "executable?", pending_win32: true do
```
straight-shoota: Just duplicating the content of `it_raises_on_null_byte` wrapped in `pending_win32 do ... end` would be much simpler code than adding an extra argument for a single use.

RX14:
Sorry for the extended delay getting this PR up to date!

spec/std/file_spec.cr
```diff
@@ -967,8 +974,10 @@ describe "File" do
       File.writable?("foo\0bar")
     end
 
-    it_raises_on_null_byte "executable?" do
-      File.executable?("foo\0bar")
+    pending "errors on executable?" do
```
straight-shoota: pending_win32

RX14: Shows me not to commit at 1am...

src/spec.cr
```diff
@@ -118,6 +118,9 @@ if ENV["SPEC_VERBOSE"]? == "1"
   Spec.override_default_formatter(Spec::VerboseFormatter.new)
 end
 
-Signal::INT.trap { Spec.abort! }
+{% unless flag?(:win32) %}
+  # TODO(windows): re-enable this once Signal is ported
```
straight-shoota: I don't think there is any precedent for `TODO(windows)` notation. Is this really a good idea?

`TODO: (windows)`?  should be easier to parse.

RX14: `TODO(windows)` is easier to grep than `/TODO:?\s+(windows)/`

straight-shoota: AFAIK it is general consensus that the identifier is followed by either a colon or whitespace. Adding parenthesis in between won't match `/TODO[: ]/` or similar regular expressions which are pretty commonly used for this kind of thing.

RX14: just grep TODO, i've never seen anyone make the regex more complicated

straight-shoota: Take a look at tools like `rake notes` and similar. They expect a certain format, and often it includes the name followed by colon or whitespace.

I think it is a good idea to add context to comment annotations. But it shouldn't add additional complexity to the simple `# TODO: message` format. Just add the context tag as part of the message and everything is fine.

RX14: The format I used is common, I didn't invent it. I'm sick and tired of making nitpicky corrections, especially when they're hypotheticals about fucking todo comments.

straight-shoota: I've never encountered this format and specifically asked for precedents. You didn't mention anything about it being used elsewhere until now.

Sija: @straight-shoota I've seen `TODO(username): message` format before, see for instance docs for https://github.com/pgilad/leasot or https://github.com/asticode/go-astitodo. Usually what's put inside of brackets is a _username_ but that's just a convention. It's noticably less common than `TODO: message` but still.

ysbaddaden:
Nothing I guess? Let's merge?

RX14:
@ysbaddaden I was waiting for another review actually after modifying the PR

PR #5960, opened 2018-04-17
 Add OpenSSL::SSL::Server and HTTP::Server#bind_ssl

straight-shoota:
This adds a reusable `SSLServer` class which wraps another `Socket::Server` in an SSL layer.

With `HTTP::Server#bind_ssl` SSL context can be configured for every server socket individually instead of using the instance variable `tls` for all sockets.

Followup on #5776

PR #5776, opened 2018-03-05
Refactor `HTTP::Server` to bind to multiple addresses

straight-shoota:
Until now, a `HTTP::Server` instance binds to exactly one IP address (host + port). This address was set in the constructor alongside several options for passing one or more handlers.

This resulted in quite a few constructors already, and adding support for more options (such as using a Unix socket) would multiply the numbers. Additionally, it should be possible to have the server listen on multiple addresses, as many other HTTP server implementations.

This PR changes the API of `HTTP::Server`:

* The constructor only accepts handlers (in multiple variations, as before).
* `#bind` is used to add one ore multiple addresses the server should listen on. It accepts the same arguments previously used in the constructor (`port`, `host, port` - each with optional `reuse_port`).
* Additional overloads exist for `Socket::IPAddress` and `Socker::Server`.
* <del>You can bind to a Unix socket using `#bind(Socket::UNIXAddress)` or by supplying a `UNIXServer`</del> (<ins>will be a subsequent PR</ins>)
* <del>There is also a single-argument overload that accepts a string with either a host + port or a Unix socket (prefixed by `unix:`) and creates the appropriate server accordingly. See [#2735 (issue-comment)](https://github.com/crystal-lang/crystal/issues/2735#issuecomment-301349490) for the reasoning behind this.</del> (<ins>will be a subsequent PR</ins>)
* `#bind_unused_port` is a shortcut for `bind(0)` and returns the port. This is particularly useful for the use case of binding to an unused port and printing the port (could maybe change the return type to `Socket::IPAddress` to include the address as well). `#port` was removed because there is no single port anymore that this method could return.
* `#listen` starts accepting incoming requests, as before. However, there is also an catch-all overload which forwards all arguments to `#bind` and starts listening. This is a shortcut for the following:
```crystal
server.bind 80
server.listen
# same as
server.listen 80
```

With a server listening on multiple interfaces, I figured it would be useful to be able to determine from which connection a client request originates, so I added a nilable (for stubbing) instance variable `socket` to `HTTP::Server::Context`.

<del>This also includes a minor fix to copy the path of a UNIXServer to each socket. Otherwise you couldn't really tell them apart if (only by the file descriptor).</del> (<ins>will be a subsequent PR</ins>)

<del>Fixes one half of #2735</del> (<ins>will be a subsequent PR</ins>)

Issue #2735, opened 2016-06-03
HTTP::Client and HTTP::Server on Unix Socket


sdogruyol:
Currently HTTP::Server only listens on TCP.  

It'd be really great to listen on a Unix socket like Unicorn, Thin, Puma e.g

An API like this 

``` crystal
server = HTTP::Server.new("/tmp/server.sock") do |context|
  context.response.content_type = "text/plain"
  context.response.print "Hello world!"
end

server.listen
```

sdogruyol:
Preferably i'd like to tackle this issue but seems like https://github.com/crystal-lang/crystal/blob/master/src/socket/unix_server.cr needs to be improved. 

I've already managed to make bind https://github.com/crystal-lang/crystal/blob/master/src/http/server/server.cr#L135-L137 to accept UNIXServer but then `handle_client` needs to be updated.

Any tips and leads are welcome 😄

asterite:
My only comment is that I wouldn't accept a string for this, in HTTP::Server constructor, as it's confusing with a host String parameter. Maybe use a required named argument, or a UnixServer instance.

sdogruyol:
@asterite how about?

```crystal
HTTP::Server.new(socket: "/tmp/server.sock") {|context|}
```

jhass:
How about dropping both `host` and `port` and requiring an `URI` instead?

```cr
HTTP::Server.new("localhost")
HTTP::Server.new("localhost:443")
HTTP::Server.new("tcp://localhost:443")
HTTP::Server.new("unix://server.sock") # relative
HTTP::Server.new("unix:////run/server.sock") # absolute

**Comment by sdogruyol** [comment]
@jhass it sounds like a good idea but i guess it's not common to create a server that way (or is it?) 🤔

**Comment by jhass** [comment]
I've seen it in various places, one popular example is [unicorn](http://www.rubydoc.info/gems/unicorn/Unicorn/Configurator#listen-instance_method)

**Comment by ysbaddaden** [comment]
In puma you add listeners to a single server:

```crystal
server = Puma::Server.new
server.add_tcp_listener(host, port)
server.add_ssl_listener(host, port, ctx)
server.add_unix_listener(path)
```

ozra:
Having multiple listeners, both unix-socket and tcp, ssl seems like something one very well might want to do for a service (if not using nanomsg, ZMQ, etc.)!

sdogruyol:
[...]

PR #6500, opened 2018-08-07
Add HTTP::Server#bind(URI)

straight-shoota:
This PR was originally included in #5776 but it had been removed as a separate feature. Now that #5959 and #5960 are merged, it is ready to be reviewed. The original description can be found in https://github.com/crystal-lang/crystal/issues/2735#issuecomment-301349490

The URI can be parsed from a string and contains the protocol and address to bind to (plus optional additional configuration for SSL context). This allows for a uniform and easily exchangeable specification in config files.

```cr
server.bind "tcp://127.0.0.1:80"
server.bind "unix:///tmp/server.sock"
server.bind "ssh://127.0.0.1:449?key=private.key&cert=certificate.cert&ca=ca.crt
```

Along the way this also adds `Socket::{IP,UNIX,}Address.parse` and `OpenSSL::SSL::Context::{Client,Server}.parse` to implement the individual component parsers.


src/socket/address.cr
```diff
@@ -202,6 +260,43 @@ class Socket
       new(sockaddr.as(LibC::SockaddrUn*), addrlen.to_i)
     end
 
+    # Parses a `Socket::UNIXAddress` from an URI.
+    #
+    # It expects the URI to include `<scheme>://<path>` where `scheme` as well
+    # as any additional URI components (such as `fragment` or `query`) are ignored.
+    #
+    # If `host` is not empty, it will be prepended to `path` to form a relatve
```
r00ster91: Typo: "relatve"

src/socket/address.cr
```diff
@@ -75,6 +103,36 @@ class Socket
       end
     end
 
+    # Parses a `Socket::IPAddress` from an URI.
+    #
+    # It expects the URI to include `<scheme>://<host>:<port>` where `scheme` as
+    # well asa ny additional URI components (such as `path` or `query`) are ignored.
```
r00ster91: Typo: "asa ny"

src/http/server.cr
```diff
@@ -217,8 +243,50 @@ class HTTP::Server
...
+    case uri.scheme
+    when "tcp"
+      bind_tcp(Socket::IPAddress.parse(uri))
+    when "unix"
+      bind_unix(Socket::UNIXAddress.parse(uri))
+    when "ssl"
+      address = Socket::IPAddress.parse(uri)
+      context = OpenSSL::SSL::Context::Server.parse(HTTP::Params.parse(uri.query || ""))
+
+      bind_ssl(address, context)
```
bcardiff: Checking for `flag?(:without_openssl)` is missing here. At least with a duplication of the Unsupported socket type message or something more specific.

src/openssl/ssl/context.cr
```diff
@@ -90,10 +90,21 @@ abstract class OpenSSL::SSL::Context
...
     end
 
+    # Parses a client context configuration from a hash-like interface.
+    #
+    # ```
+    # require "openssl"
+    #
+    # context = OpenSSL::SSL::Context::Client.parse({"key" => "private.key", "cert" => "certificate.crt", "ca" => "ca.pem"})
+    # ```
+    def self.parse(params) : self
```
bcardiff: I would expect `T.parse` to be from `String -> T`.  I would suggest `T.from_hash` for this method.

src/http/server.cr
```diff
@@ -217,8 +243,50 @@ class HTTP::Server
...
+  # Returns the effective address it is bound to.
+  #
+  # ```
+  # server = HTTP::Server.new { }
+  # server.bind("tcp://localhost:8080") # => Socket::IPAddress.new("localhost, 8080")
+  def bind(uri : String) : Socket::Address
+    bind(URI.parse(uri))
+  end
+
+  # ditto
```
bcardiff: Warn: mixed ditto and :ditto: in the PR

src/socket/address.cr
```diff
@@ -75,6 +103,36 @@ class Socket
...
+    # well as any additional URI components (such as `path` or `query`) are ignored.
+    #
+    # `host` must be an IP address (v4 or v6), otherwise `Socket::Error` will be
+    # raised. Domain names will not be resolved.
+    #
+    # ```
+    # Socket::IPAddress.parse("tcp://127.0.0.1:8080") # => Socket::IPAddress.new("127.0.0.1", 8080)
+    # Socket::IPAddress.parse("udp://[::1]:8080")     # => Socket::IPAddress.new("::1", 8080)
+    # ```
+    def self.parse(uri : URI) : IPAddress
```
bcardiff: Somehow I'm ok with `T.parse: URI -> T` 😹 , but not with Hash.

straight-shoota: URI is essentially a structured string...

src/openssl/ssl/context.cr
```diff
@@ -90,10 +90,30 @@ abstract class OpenSSL::SSL::Context
...
+    #
+    # ```
+    # require "openssl"
+    #
+    # context = OpenSSL::SSL::Context::Client.from_hash({"key" => "private.key", "cert" => "certificate.crt", "ca" => "ca.pem"})
+    # ```
+    #
+    # Params:
+    #
+    # * `key` *(required*): Path to private key file. See `#private_key=`.
```
Sija: `*(required*)` -> `(*required*)`
ditto below

bcardiff:
@straight-shoota 

The sample in the PR description mention `ssh://` as a protocol, but it should be `ssl://` probably.
The port used is not the default `443`, although any port is supported using defaults seems better for lazy copy/pasting.

```
server.bind "ssl://127.0.0.1:443?key=private.key&cert=certificate.cert&ca=ca.crt
```

Or am I missing something?

straight-shoota:
You're absolute right.

jhass:
Bold proposal: Lets ditch `ssl://` and only support `tls://`

straight-shoota:
All the classes in Crystal are called SSL... so I'd at least keep support for `ssl://`. But adding `tls://` and using it as default (in examples) sounds good.

jhass:
No, they're called `OpenSSL`, because they follow the name of the library they wrap. On the other hand we already renamed `Http`s `ssl` flags to `tls` at some point.

straight-shoota:
No, I mean inside the `OpenSSL` namespace (which is obviously named after the library), there is `OpenSSL::SSL`.

But I'd suggest to discuss this further in a new dedicated issue...

jhass:
I wouldn't be too opposed to renaming that to TLS too :P

straight-shoota:
I'll work on that. Waiting for #6530 to be merged.

straight-shoota:
No, this wasn't discussed before explicitly. And I admit that this might look a bit surprising at first.

The reason is that `HTTP` and `HTTPS` are application protocols - and `HTTP::Server` naturally speaks HTTP(S) on all connections. `server.bind "http://"` would only tell to use the HTTP protocol (which is obvious since this is an HTTP server). 

The URIs used by `bind` are meant to specify the transport protocol (TCP, TCP+TLS or UNIX socket).
This is probably most clear when looking at `tls://127.0.0.1:443?key=private.key&cert=certificate.cert&ca=ca.crt`. This URI configures a TCP+TLS server and is not an HTTP resource.

These transport protocol schemes are also used by similar applications such as Puma or Nginx.

crisward:
Wasn't aware of other servers using these protocols, makes sense. Thanks for the clarification.

PR #3520, opened 2016-11-08
Raises on failure to run a process (Fixes #3517)

lbguilherme:
This changes the behavior of `Process.new` to block until `execvp` is performed and raise an exception if it fails. Previous behavior was to return immediately and set let the forked process exit with 127. See #3517.

Before:

```cr
it "returns status 127 if command could not be executed" do
  process = Process.new("foobarbaz")
  process.wait.exit_code.should eq(127)
end
```

After:

```cr
it "raises if command could not be executed" do
  expect_raises Errno do
    Process.new("foobarbaz")
  end
end
```

PR #6501, opened 2018-08-08
Add raise Errno if Process.new fails to exec

### LINKED ISSUE #3517: Process.new with error: true raises an uncatchable exception
```crystal
begin
  Process.run "nonexistent", error: true
rescue err
  puts 123
end
```

This won't actually catch the exception:

```
ryan@DevPC-LX ~/ptime $ crystal x.cr
execvp: No such file or directory (Errno)
[4367671] *CallStack::unwind:Array(Pointer(Void)) +87
[4367562] *CallStack#initialize:Array(Pointer(Void)) +10
[4367514] *CallStack::new:CallStack +42
[4340040] *raise<Errno>:NoReturn +24
[4513305] *Process::exec_internal<String, Array(Pointer(UInt8)), Nil, Bool, (Bool | IO::FileDescriptor), (Bool | IO::FileDescriptor), (Bool | IO::FileDescriptor), Nil>:Nil +281
[4510702] *Process#initialize<String, Nil, Nil, Bool, Bool, Bool, Bool, Bool, Nil>:Nil +1886
[4508801] *Process::new<String, Nil, Nil, Bool, Bool, Bool, Bool, Bool, Nil>:Process +209
[4508542] *Process::run:error<String, Bool>:Process::Status +110
[4336839] ???
[4350777] main +41
[140421849268037] __libc_start_main +245
[4335305] ???
[0] ???
ryan@DevPC-LX ~/ptime $ 
```

Just using something like `raise Errno.new 123` will work correctly.

straight-shoota:
This PR picks up the stalled PR #3520 by @lbguilherme 

I refactored `exec_internal` to return if it was successful instead of raising. This avoids unnecessarily raising and immediately rescuing `Errno`.

Also used `write_bytes` and `ByteFormat.decode` for piping the Int32. There is no non-raising `read_bytes`.

Fixes #3517

### DIFF
--- a/spec/std/process_spec.cr
+++ b/spec/std/process_spec.cr
@@ -14,9 +14,10 @@ describe Process do
     process.wait.exit_code.should eq(1)
   end
 
-  it "returns status 127 if command could not be executed" do
-    process = Process.new("foobarbaz")
-    process.wait.exit_code.should eq(127)
+  it "raises if command could not be executed" do
+    expect_raises Errno do
+      Process.new("foobarbaz")
+    end
   end
 
   it "run waits for the process" do
--- a/src/process.cr
+++ b/src/process.cr
@@ -183,7 +183,9 @@ class Process
   def self.exec(command : String, args = nil, env : Env = nil, clear_env : Bool = false, shell : Bool = false,
                 input : Stdio = Redirect::Inherit, output : Stdio = Redirect::Inherit, error : Stdio = Redirect::Inherit, chdir : String? = nil)
     command, argv = prepare_argv(command, args, shell)
-    exec_internal(command, argv, env, clear_env, input, output, error, chdir)
+    unless exec_internal(command, argv, env, clear_env, input, output, error, chdir)
+      raise Errno.new("execvp")
+    end
   end
 
   getter pid : Int32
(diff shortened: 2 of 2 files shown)


src/process.cr
```diff
@@ -240,25 +242,41 @@ class Process
...
       rescue ex
         ex.inspect_with_backtrace STDERR
       ensure
         LibC._exit 127
       end
     end
 
+    writer_pipe.close
+    bytes = uninitialized UInt8[4]
+    if reader_pipe.read(bytes.to_slice) == 4
```
RX14: Why not `read_bytes`?

straight-shoota: Because it raises when the pipe is closed in the forked process.

spec/std/process_spec.cr
```diff
@@ -14,9 +14,10 @@ describe Process do
     process.wait.exit_code.should eq(1)
   end
 
-  it "returns status 127 if command could not be executed" do
-    process = Process.new("foobarbaz")
-    process.wait.exit_code.should eq(127)
+  it "raises if command could not be executed" do
+    expect_raises Errno do
```
Sija: IMO would be good to add specs for passing correct `Errno.errno` value.

src/process.cr
asterite: Shouldn't this be `read_fully`?

asterite: (or something like that, a single `read` is not guaranteed to fill the given slice)

straight-shoota: Right, it should be `read_fully?` - or better just `read_bytes?`. I doubt there is a practical relevance though, the first four bytes from a pipe are hardly going to be split into multiple reads.

lbguilherme: Reads (or writes) on a pipe smaller than PIPE_BUF (usually 4096) are guaranteed to no be split. We are safe here.

asterite: Cool, thanks for the explanation.

PR #6510, opened 2018-08-09
Added 128bit primitives to the range definition

Zenohate:
I stumbled upon the fact 128bit primitives weren't handled by `range`, so here you go


RX14:
Is there any code that reproduces a compiler bug from this?

Zenohate:
Let me make a short one, will send tomorrow

jhass:
I forgot, do we still support Int128?  Seems like an  easy and quick fix otherwise.

HertzDevil:
Fixed by d6904bfecd5073f0c95447e6280b74a2238eac11.

PR #6512, opened 2018-08-09
Add tests to expectations methods

rodrigopinto:
# What is this PR for?

This PR is for adding tests for the expectation methods on the spec module.

## TODO

List of the methods to be covered with tests in this PR:

- [x] `be`
- [x] `be(value)`
- [x] `be_a(type)`
- [x] `be_close(expected, delta)`
- [x] `be_nil`
- [x] `be_truthy`
- [x] `be_falsey`
- [x] `be_true`
- [x] `be_false`
- [x] `contain(expected)`
- [x] `eq(value)`
- [x] `match(value)`
- [x] `expect_raises(klass)`

### DIFF
--- a/spec/std/spec/expectations_spec.cr
+++ b/spec/std/spec/expectations_spec.cr
@@ -1,7 +1,93 @@
 require "spec"
 
 describe "expectations" do
-  context "start_with" do
+  describe "be" do
+    it { 1.should be < 3 }
+    it { 2.should be <= 3 }
+    it { 3.should be <= 3 }
+    it { 3.should be >= 3 }
+    it { 4.should be >= 3 }
+    it { 5.should be > 3 }
+  end
+
+  describe "be" do
+    it { "hello".should be "hello" }
+    it do
+      array = [1]
+      array.should_not be [1]
+    end
+  end
+
+  describe "be_a" do
+    it { "Hello".should be_a(String) }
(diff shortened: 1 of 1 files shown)


spec/std/spec/expectations_spec.cr
```diff
@@ -1,6 +1,43 @@
 require "spec"
 
 describe "expectations" do
+  context "be_nil" do
```
straight-shoota: This should all be `describe`. It describes the `be_nil` matcher.

rodrigopinto:
> be_true and be_false are aliases to eq true and eq false. There is no need to test them again if we test eq.

☝️ @j8r I kinf of disagree with this because even being an alias it still to be a public API of the specs module, so, testing them gives us the minimum guarantee that it will respond as expected.

> I find them confusing, particularly regarding their close twins be true and be false ≠ be_true and be_false. Same with be nil and be_nil
>
> Do we really need them? true.should eq true – "true should equals true" isn't enough English to keep having this confusion?

I agree that it might creates confusions considering their close twins, but having the alias for the `eq(true|false)` I don't see as a pain point, besides the fact it is more piece of code to maintain.

rodrigopinto:
@RX14 I just rebased and squashed from the master. Let's wait for CI, but I think it is ready to go.

PR #6514, opened 2018-08-09
Parser: give syntax error on comma after newline in argument list

### LINKED ISSUE #6513: Formatter bug with newline before comma in method call
Trying to format this:
```crystal
def method(a, b)
end
method(1
,2)
```
gives
```
Error: couldn't format './main.cr', please report a bug including the contents of it: https://github.com/crystal-lang/crystal/issues

expecting ,, not `NEWLINE, `, at :3:9 (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???

```

asterite:
Fixes #6513

Code like this:

```crystal
foo(x
, y)
```

currently doesn't give a syntax error, but it should (like in Ruby). We can of course discuss whether it should really be a syntax error or not. My reasoning is that the comma signals that more arguments are coming, and it's confusing that the comma starts the next line. For example, if we remove the parentheses:

```crystal
foo x
, y
```

that won't compile. So code with and without parentheses should look more or less the same (maybe this is a weaker argument, but still...).

I believe in many other languages this is a syntax error too.

This of course is a breaking change, but I doubt there are many out there that write code like this.

This syntax option was available for many other things, like method arguments, but not available in some other things, like array literals. Now it's unified to give a syntax error.

### DIFF
--- a/spec/compiler/crystal/tools/playground_spec.cr
+++ b/spec/compiler/crystal/tools/playground_spec.cr
@@ -317,7 +317,7 @@ describe Playground::AgentInstrumentorTransformer do
   end
 
   it "do not instrument top level macro calls" do
-    assert_agent(<<-CR
+    assert_agent(<<-FROM, <<-TO
     macro bar
       def foo
         4
@@ -325,8 +325,7 @@ describe Playground::AgentInstrumentorTransformer do
     end
     bar
     foo
-    CR
-    , <<-CR
+    FROM
     macro bar
       def foo
         4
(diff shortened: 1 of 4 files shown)


asterite:
Hmmm... actually, Java, C# and Python allow such syntax. But Ruby and Go don't. So I guess it's a matter of preference.

Right now it's behaving in a way or another depending on the case, so it's not unified. Let's use this PR to discuss what should we do with this. Maybe this PR is fine, it hurts my eyes a bit to see that stray comma there :-P

hugoabonizio:
This syntax (leading comma) was used in [JS community](https://gist.github.com/isaacs/357981), but isn't used widely nowadays. I guess everybody realised it's ugly.

Issue #5821, opened 2018-03-13
Add crystal-expanded & crystal-normalized emit options


bew:
From time to time I want to see the generated Crystal code at several specific stages of the Crystal compiler:
- all the code after all top level macro expansions
- the code of one class after all top level macro expansions
- the code after semantic pass
- the code that will actually be compiled into binary (normalized & with all dead code removed)
- ...

For a start, I think it would be nice to have
- `--emit crystal-expanded`: dump the crystal code with all macros expanded
- `--emit crystal-normalized`: dump the crystal code after the normalization pass (which transforms all syntactic sugar to basic control structures, like `unless` -> `if`, `until` -> `while`, etc..).
  Also at this stage, the literals would already be expanded (e.g: string interpolation to `String.build ...`)
  This dump would be quite hard to read I think, but would still be Crystal code (raw llvm IR is also nice, but it's another lower level)

(or with `crystal` => `cr` to reduce size of flags)

Those options _could_ be used with `--no-codegen`, and the resulting files be compiled again into binary if needed (as they are valid Crystal code).

It's just ideas on how to inspect what's going on in the compiler, there are many things I didn't talked about that would be cool (e.g: after the type inference pass, or simply the list of symbols/unique string literals, ..)

Also in the long run it could help creating new stdlib (see what is needed), help during ports to new platforms (see what's used?), and in a way, inspecting into the compiler (via emit or tools, or compiler plugins) could allow to make #921 possible? (I think I'm going a bit too far for this issue, but I'd love all those things!!)

faustinoaq:
This feature would be very useful, use-case here: https://github.com/crystal-lang-tools/vscode-crystal-lang/issues/4

faustinoaq:
I just got another idea, I think printing typed code would be very useful, by example:

Original code: `foo.cr`

```cr
FOO_METHODS = %w(foo bar)

{% for method in FOO_METHODS %}
def {{method.id}}(baz)
end
{% end %}

foo("foo")
bar(42)
```

Macros expanded: `foo.expanded.cr`

```cr
def foo(baz)
end

def bar(baz)
end

foo("foo")
bar(42)
```

Typed code: `foo.typed.cr`

```cr
def foo(baz : String) : Nil
end

def bar(baz : Int32) : Nil
end

foo("foo")
bar(42)
```

This can be very useful for editor extensions and tools :+1: 

Maybe a flag like `--emit crystal-typed` ? :sweat_smile:

[...]

Issue #5820, opened 2018-03-13
Using --no-codegen blocks --emit options


bew:
When we do: `crystal build hello.cr --emit llvm-ir --no-codegen`
It'll not generate the `hello.ll` file even though it was asked for.

This is because the emit options are handled all-in-one-place in
https://github.com/crystal-lang/crystal/blob/80cbe6603f938fc2200b785bd3fb2edbd854a17d/src/compiler/crystal/compiler.cr#L676-L687

Which is called in the `codegen` block:
https://github.com/crystal-lang/crystal/blob/80cbe6603f938fc2200b785bd3fb2edbd854a17d/src/compiler/crystal/compiler.cr#L339-L353
At line 348.

I think it should emit what was asked for as soon as it get it, so the llvm-ir after it was built, the obj in the codegen phase, etc..

bew:
This also means that there is no way to dump llvm IR when there is a `module validation failed` error, that we would like to debug.

straight-shoota:
In order to debug #5972 I had to modify the compiler manually to disable module validation. So +1 to that!

bew:
While looking into that, I noticed that all the codegen phases (mainly crystal to LLVM IR and LLVM to BC+OBJ) are done in the `codegen` method.
I would suggest to separate the Crystal to LLVM IR codegen phase from the binary codegen (LLVM-IR to BC+OBJ), maybe named `IR codegen` ?

Also the current different emit options are in 2 categories:
- after llvm ir generation: for `llvm-ir` (should be processed even on `--no-codegen`)
- after binary codegen: for `asm`, `llvm-bc`, `obj` (not processed on `--no-codegen`)

Those are currently all handled and processed after binary codegen. By having the `IR codegen` separated, we could handle the different emit options at different time.

WDYT?

asterite:
What's so bad about not using `--no-codegen` when using `--emit`? No codegen means the codegen phase doesn't run, and this trivially means no artifacts (.ll, .s, .o, final executable) will be produced.

straight-shoota:
@asterite True, these are all different stages of codegen. But what do you do if you want the `.ll` (or `.s`) but not `.o`?

`--no-codegen --emit llvm-ir` makes sense for this: It essentially says "skip codegen, but emit LLVM-IR" which can be interpreted as "do codegen but abort after LLVM-IR is dumped".

asterite:
[...]

PR #6515, opened 2018-08-10
Use enum (flags) instead of hardcoded string values for emit kinds

bew:
I think that using enum flags is cleaner in this case.
It's better to remove handcoded emit values from obscure places in the compiler code :smiley:

I tried to keep the emit values the same (e.g: `llvm-ir`), but it would be simpler (?) to use `llvm_ir` as it's how the enum flag can be easily parsed. This could be changed though, but a breaking change.

This is the beginning to work on #5820, and maybe #5821.

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -42,8 +42,6 @@ class Crystal::Command
         --help, -h               show this help
     USAGE
 
-  VALID_EMIT_VALUES = %w(asm llvm-bc llvm-ir obj)
-
   def self.run(options = ARGV)
     new(options).run
   end
@@ -309,7 +307,10 @@ class Crystal::Command
       end
 
       unless no_codegen
-        opts.on("--emit [#{VALID_EMIT_VALUES.join('|')}]", "Comma separated list of types of output for the compiler to emit") do |emit_values|
+        valid_emit_values = {{ Compiler::EmitKinds.constants.map(&.stringify).reject { |m| %w(none all).includes?(m.downcase) } }}
+        valid_emit_values.map! { |v| v.tr("_", "-").downcase }
+
+        opts.on("--emit [#{valid_emit_values.join('|')}]", "Comma separated list of types of output for the compiler to emit") do |emit_values|
           compiler.emit = validate_emit_values(emit_values.split(',').map(&.strip))
         end
       end
@@ -514,12 +515,15 @@ class Crystal::Command
   end
 
   private def validate_emit_values(values)
+    emit_kinds = Compiler::EmitKinds::None
     values.each do |value|
-      unless VALID_EMIT_VALUES.includes?(value)
+      if kind = Compiler::EmitKinds.parse?(value.tr("-", "_"))
+        emit_kinds |= kind
+      else
         error "invalid emit value '#{value}'"
       end
     end
-    values
+    emit_kinds
   end
 
   private def error(msg, exit_code = 1)
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -307,7 +307,7 @@ class Crystal::Command
       end
 
       unless no_codegen
-        valid_emit_values = {{ Compiler::EmitKinds.constants.map(&.stringify).reject { |m| %w(none all).includes?(m.downcase) } }}
+        valid_emit_values = Compiler::EmitTarget.names
         valid_emit_values.map! { |v| v.tr("_", "-").downcase }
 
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/compiler.cr
```diff
@@ -98,13 +98,21 @@ module Crystal
     # and can later be used to generate API docs.
     property? wants_doc = false
 
-    # Can be set to an array of strings to emit other files other
+    @[Flags]
+    enum EmitKinds
```
straight-shoota: Better have the name in singular, that's used throughout the code base.
Maybe `EmitTarget` would even be a better name.

straight-shoota: And I'd suggest to have the values either all uppercase or all lower case. I prefer the former.

src/compiler/crystal/command.cr
```diff
@@ -309,7 +307,10 @@ class Crystal::Command
       end
 
       unless no_codegen
-        opts.on("--emit [#{VALID_EMIT_VALUES.join('|')}]", "Comma separated list of types of output for the compiler to emit") do |emit_values|
+        valid_emit_values = {{ Compiler::EmitKinds.constants.map(&.stringify).reject { |m| %w(none all).includes?(m.downcase) } }}
```
straight-shoota: Why use a macro instead of `Enum.names`?

src/compiler/crystal/compiler.cr
```diff
@@ -667,22 +675,18 @@ module Crystal
...
-        when "asm"
-          compiler.target_machine.emit_asm_to_file llvm_mod, "#{output_filename}.s"
-        when "llvm-bc"
-          FileUtils.cp(bc_name, "#{output_filename}.bc")
-        when "llvm-ir"
-          llvm_mod.print_to_file "#{output_filename}.ll"
-        when "obj"
-          FileUtils.cp(object_name, "#{output_filename}.o")
+      def emit(emit_kinds : EmitKinds, output_filename)
+        emit_kinds.each do |kind|
```
straight-shoota: Why use a loop? You can just have consecutive conditionals:
```cr
if emit_kinds.asm?
  # do asm
end
if emit_kinds.llvm_bc?
  # do llvm-bc
end
# etc...
```

asterite:
Thank you, but I disagree. The code became much harder to write, read and understand.

asterite:
31 additions and 23 deletions. Now we have more code. Also more macro code, which is always harder to understand. A gsub/tr call to replace dash with underscore.

The old code was just fine.

src/compiler/crystal/command.cr
bew: Wow how did I miss that (PR-ing at 3am may be why :smile: )
thanks!

bew:
@asterite How is it much harder to write/read/understand?

> Now we have more code. Also more macro code, which is always harder to understand.

The macro code has been removed as I didn't noticed `Enum.names`

> A gsub/tr call to replace dash with underscore.

I don't like it either, that's why I asked what I should do in the OP.
I'm also thinking about a simple `case .. when ..` block to match all possible emit values to their respective flags, to remove the dependence between the value for CLI, and the name of the flag in the code.

> The old code was just fine.

For the current use-case I agree, but I'd like to change _when_ some emit targets are emitted (for #5820 & #5821), and I really don't want to have hardcoded strings related to CLI input lying around when an enum flag can do the trick (and be typesafe!).

src/compiler/crystal/command.cr
```diff
@@ -309,7 +307,10 @@ class Crystal::Command
       end
 
       unless no_codegen
-        opts.on("--emit [#{VALID_EMIT_VALUES.join('|')}]", "Comma separated list of types of output for the compiler to emit") do |emit_values|
+        valid_emit_values = Compiler::EmitTarget.names
+        valid_emit_values.map! { |v| v.tr("_", "-").downcase }
```
Sija: You should use `Char`s instead of `String`s here.

bew: Yeah that's what I did originally, but there is only `tr(String, String)` not `tr(Char, Char)`

asterite: There's `gsub(Char, Char)`

straight-shoota:
> The macro code has been removed as I didn't noticed `Enum.names`

It has just been moved *into* `Enum.names`. It's still executing a macro an thus more complexity at compile time.

Under the current circumstances, @asterite is probably right that this isn't really useful as an improvement. But in the light of changes to the locations where these flags are used, it's a different story.
Maybe it would help to see where you want to go with this.

straight-shoota:
The CLI options could just hard coded into the description string to avoid the macro. It's not that they change all the time to be automatically generated.

PR #6516, opened 2018-08-10
Allow creating None enum flag with Enum.from_value

### LINKED ISSUE #5768: Cannot create an enum flag instance from value 0
```cr
@[Flags]
enum A
  A
  B
  C
end

pp A::None.value

puts A.from_value 0 # => Unknown enum A value: 0 (Exception)
```
https://carc.in/#/r/3o46

I think this should give `A::None`, not an exception.

bew:
Fixes #5768

It also adds type restrictions to `Int`, and fix one usage in `src/debug/elf.cr`

### DIFF
--- a/spec/std/enum_spec.cr
+++ b/spec/std/enum_spec.cr
@@ -117,7 +117,7 @@ describe Enum do
     end
 
     it "for flags enum" do
-      SpecEnumFlags.from_value?(0).should be_nil
+      SpecEnumFlags.from_value?(0).should eq(SpecEnumFlags::None)
       SpecEnumFlags.from_value?(1).should eq(SpecEnumFlags::One)
       SpecEnumFlags.from_value?(2).should eq(SpecEnumFlags::Two)
       SpecEnumFlags.from_value?(3).should eq(SpecEnumFlags::One | SpecEnumFlags::Two)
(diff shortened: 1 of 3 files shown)


spec/std/enum_spec.cr
```diff
@@ -134,10 +134,10 @@ describe Enum do
     end
 
     it "for flags enum" do
```
straight-shoota: Perhaps also add a spec for value `0` again (== `eq None`).

bew: yess good idea

bew:
Good to go? (one of travis check failed for another reason)

RX14:
Can this be rebased on master so we're sure there are no more spec failures?

src/debug/elf.cr
```diff
@@ -168,7 +168,7 @@ module Debug
       ei_version = @io.read_byte.not_nil!
       raise Error.new("Unsupported version number") unless ei_version == 1
 
-      ei_osabi = OSABI.from_value(@io.read_byte)
+      ei_osabi = OSABI.from_value(@io.read_byte.not_nil!)
```
bew: Should we `raise IO::EOFError.new` here instead of `not_nil!` in the unfortunate case where `read_byte` is nil?

bew: @ysbaddaden ^

RX14: Everywhere around here uses `not_nil!`, any change would need to change all of them in another PR.

bew: ah indeed, no question then

bew:
So I rebased on master, one of travis job fails for a weird reason: https://travis-ci.org/crystal-lang/crystal/jobs/472825523

RX14:
@bew thats just OOM

bew:
> that's just OOM

yeah sure, but it's still not wanted^^

RX14:
Travis builds randomly fail from OOM all the time

straight-shoota:
@bew Could you rebase on master to make sure it's still valid?

PR #6068, opened 2018-05-05
Fix: nonblocking standard file descriptors using timers

ysbaddaden:
This is an attempt at fixing #2713 and #5830. With this patch all examples in #2713 plus the failing example in #5380 are now working.

~~**WARNING:**
Proof of concept / work in progress!~~

**The issue:**
Setting O_NONBLOCK doesn't only affect the file descriptor of the application, but the whole file description which is shared with other programs. This is always the case for STDIN, STDOUT and STDERR that can be changed at any time by other processes. This makes Crystal applications a bad citizen too, since it changes those file descriptors in unexpected ways.

**Proposed solution:**
This patch introduces a new class, IO::StdFileDescriptor. It still relies on libevent to determine whether a file descriptor is ready, but doesn't set O_NONBLOCK and instead arms a timer then uses a blocking syscalls that will be interrupted by the timer (through the SIGALRM signal) which resumes to libevent, achieving a (somewhat) nonblocking call —i.e. it's blocking but not for too long.

The SIGALRM signal must be trapped using sigaction to set the SA_NODEFER and more importantly not set the SA_RESTART flag that would prevent syscalls from failing with EINTR.

All syscalls that may be interrupted must now check both EAGAIN and EINTR before deciding to retry or fail.

**Caveats:**
Relies on a magic number to setup the timer (1 microsecond); the value must be greater than the timer precision, yet small enough to not block operations for too long...

Some syscalls are now interrupted by SIGALRM which can happen at any time, and in situations unrelated to STDIN/OUT/ERR read/write. It's possible that some syscalls will now start failing with EINTR, and shall be retried. The linux `signal(7)` manpage lists the following syscalls that we use as interruptible for example:

- read(2), write(2) (as expected by this pull request);
- open(2);
- wait(2), waitpid(2);
- accept(2), connect(2), recv(2), recvfrom(2), recvmsg(2), send(2), sendto(2), sendmsg(2)
- flock(2), fcntl(2) F_SETLKW.

Hijacks SIGALRM —we already hijack SIGCHLD which is more common.

**WIP:**
- ~~Only C bindings for `x86_64-linux-gnu` have been added (for the time being);~~
- ~~some targets (darwin, openbsd) don't support POSIX timers but provide the older `setitimer(2)`;~~
- ~~must check and retry interuptible syscalls that failed with EINTR;~~
- POSIX only, the class doesn't exist on Win32 (still directly uses `IO::FileDescriptor`);
- Not isolated into `Crystal::System`.

Issue #5830, opened 2018-03-16
Bug with signal trap


anykeyh:
Code to reproduce this bug (need interactive terminal):

```crystal
system("echo call a subprocess")

Signal::INT.trap do
  puts "Canceled by user"
  exit
end

puts "Press CTRL+C"
gets
```

## Expected behavior

Print "Canceled by user" and exit the program 

## Current behavior

Do nothing when you press Ctrl+C

## Environment

Tested on MacOS with Crystal  0.24.1 / LLVM 5.0.1

## Additional notes

Happens when a child process is created through backslash quote or `system` call.

NIFR91:
Same problem 

```text
Crystal 0.24.2 [4f9ed8d03] (2018-03-08)

LLVM: 4.0.0
Default target: x86_64-unknown-linux-gnu
```
with code 

```
at_exit{ puts "Terminated"} 
Signal::INT.trap { exit } 
puts "Infinite Loop" 
loop { } 
```

Expected behavior:  
exit and print "Terminated" when sending interrupt signal with `kill -int [pid]`  or Ctrl+C

Current behavior:
block in trap is never executed, and the program continues execution.

---

The program works with 
```
at_exit{ puts "Terminated"} 
# Signal::INT.trap { exit } 
LibC.signal Signal::INT.value,->(s : Int32){ exit } 
puts "Infinite Loop" 
loop { } 
```

The program exit and print "Terminated" after `kill -int [pid]` or Ctrl+C

ysbaddaden:
@NIFR91 this is unrelated to the initial issue, and this can't work in Crystal: signals are handled through the event loop; you can't use signal handling to exit a busy loop in crystal, because the busy loop prevents the event loop to resume and handle the signal... Inserting a `sleep 0` or anything IO related inside the busy loop will do the job.

Note that you need crystal master for the `at_exit` block to be actually run (and thus print `Terminated`).

ysbaddaden:
Initial issue fixed in ff56d94.

NIFR91:
@ysbaddaden , yes, you are right i didn't think it thoroughly. Maybe in my program i never yield to the event loop, ill check it. I can confirm this issue  is fixed.

ysbaddaden:
Duplicate of #2713.

PR #6518, opened 2018-08-10
Reopen standard handles when they are a TTY

Timbus:
Reopening the TTY prevents crystal from breaking other programs that are
attached to the same TTY when it sets O_NONBLOCK on its std filehandles,
and also prevents crystal from breaking itself when spawning a new process.

Two important things: 
1. This changes the std handle numbers. Anyone expecting their programs STDIN/OUT/ERR to be 0,1, and 2, will be (sorta) wrong. This shouldn't actually affect anyone? But if it does, we can save the original FDs and swap em around.
2. There are two places where I am not dealing with potential error codes and would like to know how I'm supposed to handle that, especially if the output might not be available?

To test, I put these two files in the build directory, and ran tester.sh with bash

`tester.sh`
```bash
#!/usr/bin/env bash

./.build/crystal test.cr

read -p "Prompt: " result
printf '%s\n' "$result"
```

`test.cr`
```crystal
state = STDOUT.blocking
puts "Do we block now? #{state}"

system("echo call a subprocess")

if STDOUT.blocking == state
  puts "We cleanly spawned =)"
else
  puts "ERRAR! Child messed with handles"
end
```

This checks that the child process doesn't mess with the handle, and also ensures crystal doesn't mess with the TTY's stdin.


asterite:
The file(s) that define StdFileDescriptor are missing.

src/io/std_file_descriptor.cr
```diff
@@ -0,0 +1,26 @@
...
+
+class IO::StdFileDescriptor < IO::FileDescriptor
+  def initialize(@fd : Int32)
+    @closed = false
+
+    # If we have a TTY for stdin/out/err, it is a shared terminal.
+    # We need to reopen it to use O_NONBLOCK without causing other programs to break
+    if tty?
+      # Figure out the terminal TTY name
+      path = UInt8[256]
```
asterite:
```crystal
```

What you wrote creates a slice of size 1 with the value 256. You are lucky it doesn't somehow segfaults.

src/io/std_file_descriptor.cr
```diff
@@ -0,0 +1,26 @@
+require "./syscall"
+require "./file_descriptor"
+
+class IO::StdFileDescriptor < IO::FileDescriptor
```
asterite: I would probably avoid a subclass and just add a boolean in the constructor to do this small logic. But I'd like to know others' opinions.

asterite:
Yeah, it's a tad confusing.

src/io/std_file_descriptor.cr
Timbus: There was more code when I started out, trying to dup old FDs and stuff. But yeah this could be put straight into IO::FD I guess.

Timbus:
Well its 1am, I'll pick this back up when I wake up.
Ah, for reference this was supposed to be an easier alternative to #6068 but .. we will see.

asterite:
I hope it works! It's really much simpler.

src/io/std_file_descriptor.cr
straight-shoota: You don't need a flag, just a differently named constructor, like `FileDescriptor.standard` which calls `FileDescriptor.new` with the newly opened fd.

ysbaddaden: We don't need to test for `tty?` all the time either. A special constructor is a good idea.

Timbus:
'Kay. I think it's good now.

It's outside the scope of the original issue, but I might look into handling closed descriptors better. I *nearly* managed to handle closed FDs:
`crystal eval 'puts "bla"' >&-`

But handling descriptors being closed dynamically would be more ideal, and would also solve the issue in my last comment.

src/io/file_descriptor.cr
```diff
@@ -18,6 +18,27 @@ class IO::FileDescriptor < IO
     end
   end
 
+  def self.from_stdio(fd)
```
straight-shoota: This method should probably be `:nodoc:`, it's only internal implementation and probably not useful in user code.

src/kernel.cr
```diff
@@ -1,13 +1,13 @@
...
 {% else %}
   require "c/unistd"
 
-  STDIN  = IO::FileDescriptor.new(0, blocking: LibC.isatty(0) == 0)
-  STDOUT = (IO::FileDescriptor.new(1, blocking: LibC.isatty(1) == 0)).tap { |f| f.flush_on_newline = true }
-  STDERR = (IO::FileDescriptor.new(2, blocking: LibC.isatty(2) == 0)).tap { |f| f.flush_on_newline = true }
+  STDIN  = IO::FileDescriptor.from_stdio(0)
+  STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
```
j8r: I would rather not use tap for `win32` and repeat for `unix` (`&.` can be used btw), but put at the end
```crystal
STDOUT.flush_on_newline = true
STDERR.flush_on_newline = true
```

src/io/file_descriptor.cr
Timbus: :+1:

src/kernel.cr
Timbus: I don't know about const assignment in crystal, but is it potentially going to be run out-of-order if I do that?  For example, it looked like you could use STDIN in the `main` method, before code in kernel.cr runs. (I could be wrong)

asterite:
That's correct. tap must be used.

src/kernel.cr
j8r: Didn't know this detail about out of order execution.
Still, a style detail, `|f| f.` can be replaced by `&.` :smile: 
```crystal
{% if flag?(:win32) %}
  STDOUT = IO::FileDescriptor.new(1).tap &.flush_on_newline = true
  STDERR = IO::FileDescriptor.new(2).tap &.flush_on_newline = true
{% else %}
  require "c/unistd"
  STDOUT = IO::FileDescriptor.from_stdio(1).tap &.flush_on_newline = true
  STDERR = IO::FileDescriptor.from_stdio(2).tap &.flush_on_newline = true
{% end %}
```

src/io/file_descriptor.cr
```diff
@@ -18,6 +18,28 @@ class IO::FileDescriptor < IO
     end
   end
 
+  # :nodoc:
+  def self.from_stdio(fd)
+    # XXX: This is -supposed- to work, but something (libevent?) is changing the FD before here.
```
RX14: Is this comment still relevant? Is this PR still WIP?

RX14:
libuv dup3's the FD to replace the old one with the newly open one.

I'm not sure if that's required here, since the standard file descriptors remain open and referring to the same device.

Timbus:
I initially did that, but then you also need to dup a copy of the original FD so you can pass it to the kid process and it was a bit messy. From what I see libuv always uses new pipes instead of sharing FDs with a child process, so it can afford to dup over the originals (a safer approach, imo)

src/io/file_descriptor.cr
Timbus: The code should still be there, and the comment still stands until someone (me?) looks into the epoll setup that happens for all handles once the first one is initialized. But I'd say the PR is ready

RX14:
@Timbus why do the child processes need the original FD? Surely the only thing required is to reset `O_NONBLOCK` before calling `exec`?

Timbus:
That would be to solve the issue observed in https://github.com/crystal-lang/crystal/issues/5830 .. Blocking IO on the shared FD halts the event loop. So we have to give the child the original FD and keep ours as nonblocking.

RX14:
ohhh yeah, my bad, thanks for the correction!

Issue #3209, opened 2016-08-29
Cannot read big files using File.read (Int32 overflow)


olbat:
Hello,
I'm trying to load the content of a big file in memory and I used the [`File.read`](https://crystal-lang.org/api/0.18.7/File.html#read%28filename%2Cencoding%3D%3Cspanclass%3D%22n%22%3Enil%3C%2Fspan%3E%2Cinvalid%3D%3Cspanclass%3D%22n%22%3Enil%3C%2Fspan%3E%29%3AString-class-method) method to do so (bad habit from ruby).

Here is the error I got:

```
$ crystal eval 'p File.read("BIGFILE_2GB").size'
negative capacity (ArgumentError)
[4367495] *CallStack::unwind:Array(Pointer(Void)) +87
[4367386] *CallStack#initialize:Array(Pointer(Void)) +10
[4367338] *CallStack::new:CallStack +42
[4345512] *raise<ArgumentError>:NoReturn +24
[4377351] *String::check_capacity_in_bounds<Int32>:Nil +39
[4427979] *File::read<String, Nil, Nil>:String +171
[4427793] *File::read<String>:String +17
[4343431] ???
[4356537] main +41
[139793359363909] __libc_start_main +245
[4341993] ???
[0] ???
```

And even worst:

```
$ crystal eval 'p File.read("BIGFILE_4GB").size'
0
```

I'm working with crystal 0.18.7:

```
crystal --version
Crystal 0.18.7 [68783f1] (2016-07-03)
```

[...]

PR #6523, opened 2018-08-11
Improves Growth Factor

chris-huxtable:
Based on finding the growth coefficient of `IO::Memory` was 2 and the information referenced in [this article](https://github.com/facebook/folly/blob/master/folly/docs/FBVector.md#memory-handling). I did some testing on different coefficients. While my findings don't directly address the potential problems with memory re-use cited in the article, I have determined that there is a smaller coefficient which appears to have comparable times but wastes less then half the amount of memory at worst. 

That coefficient is 1.30568 with only 23.41% waste at worse. This is the fractal dimension of the Apollonian packing of circles. I probed both sides of this constant and it seems to be a local maxima. Which is interesting.

The result of this patch according to the the numbers would be a slightly slower growth rate. ~149.8 verses the current ~157.5 'iterations per second' as measured on my machine (macOS 10.13, X5675). With substantially less wasted memory due to over allocation; about 26.59% less. Even if this trade off isn't good enough using the same growth factor as before (2.0) would be faster with this implementation; now ~163.3 where before it was ~158.0.

### Growth Factor: 2 (`Math.pw2ceil`)
```
   1000 157.92  (  6.33ms) (± 2.07%)  4194943 B/op   1.00× slower
  10000 157.93  (  6.33ms) (± 2.30%)  4194948 B/op   1.00× slower
 100000 157.98  (  6.33ms) (± 2.41%)  4194943 B/op   1.00× slower
1000000 158.21  (  6.32ms) (± 2.16%)  4194942 B/op        fastest
1048577 158.12  (  6.32ms) (± 2.10%)  4194942 B/op   1.00× slower

Worst-case Waste: 50.0%
```

### Growth Factor: 2.0
```
   1000  163.6  (  6.11ms) (± 2.30%)  4194945 B/op        fastest
  10000 163.35  (  6.12ms) (± 2.48%)  4194953 B/op   1.00× slower
 100000 163.39  (  6.12ms) (± 2.29%)  4194939 B/op   1.00× slower
1000000 163.32  (  6.12ms) (± 2.38%)  4194938 B/op   1.00× slower
1048577  163.2  (  6.13ms) (± 2.24%)  4194946 B/op   1.00× slower

Worst-case Waste: 50.0%
```

### Growth Factor: 1.625
```
   1000 128.29  (   7.8ms) (± 2.67%)  4427877 B/op   1.00× slower
  10000 128.91  (  7.76ms) (± 2.35%)  4427875 B/op        fastest
 100000 128.52  (  7.78ms) (± 2.48%)  4427871 B/op   1.00× slower
1000000 128.16  (   7.8ms) (± 2.78%)  4427868 B/op   1.01× slower
1302410 128.47  (  7.78ms) (± 2.60%)  4427881 B/op   1.00× slower

Worst-case Waste: 38.46%
```

### Growth Factor: 1.618
```
   1000 140.09  (  7.14ms) (± 2.06%)  4029754 B/op   1.00× slower
  10000 139.98  (  7.14ms) (± 2.46%)  4029743 B/op   1.00× slower
 100000 139.56  (  7.17ms) (± 2.50%)  4029744 B/op   1.00× slower
1000000 140.22  (  7.13ms) (± 2.40%)  4029737 B/op        fastest
1149151 140.09  (  7.14ms) (± 2.49%)  4029733 B/op   1.00× slower
```
Worst-case Waste: 38.2%

### Growth Factor: 1.5
```
   1000 115.83  (  8.63ms) (± 2.45%)  4838648 B/op   1.00× slower
  10000 115.56  (  8.65ms) (± 2.65%)  4838654 B/op   1.00× slower
 100000 115.14  (  8.68ms) (± 2.66%)  4838655 B/op   1.01× slower
1000000 115.93  (  8.63ms) (± 2.66%)  4838661 B/op        fastest
1456110 115.53  (  8.66ms) (± 2.38%)  4838653 B/op   1.00× slower
```
Worst-case Waste: 33.33%

### Growth Factor: 1.4
```
   1000 108.82  (  9.19ms) (± 2.45%)  5297519 B/op   1.00× slower
  10000 108.86  (  9.19ms) (± 2.40%)  5297519 B/op        fastest
 100000 108.59  (  9.21ms) (± 2.69%)  5297525 B/op   1.00× slower
1000000 108.82  (  9.19ms) (± 2.40%)  5297533 B/op   1.00× slower
1372074 108.49  (  9.22ms) (± 2.35%)  5297524 B/op   1.00× slower
```
Worst-case Waste: 28.57%

### Growth Factor: 1.35845
```
   1000 113.64  (   8.8ms) (± 2.03%)  5766913 B/op   1.05× slower
  10000  113.6  (   8.8ms) (± 2.21%)  5766913 B/op   1.05× slower
 100000 113.35  (  8.82ms) (± 2.15%)  5766913 B/op   1.05× slower
1000000  117.9  (  8.48ms) (± 2.80%)  5766912 B/op   1.01× slower
1318282 118.97  (  8.41ms) (± 2.44%)  5766902 B/op        fastest

Worst-case Waste: 26.39%
```

### Growth Factor: 1.30568
```
   1000 149.23  (   6.7ms) (± 2.59%)  5161122 B/op   1.00× slower
  10000 149.83  (  6.67ms) (± 2.57%)  5161122 B/op   1.00× slower
 100000  149.5  (  6.69ms) (± 2.53%)  5161116 B/op   1.00× slower
1000000 149.86  (  6.67ms) (± 2.54%)  5161120 B/op        fastest
1055629 149.86  (  6.67ms) (± 2.65%)  5161119 B/op   1.00× slower

Worst-case Waste: 23.41%
```

### Growth Factor: 1.3
```
   1000 143.51  (  6.97ms) (± 2.41%)  5772912 B/op   1.00× slower
  10000 143.95  (  6.95ms) (± 2.63%)  5772928 B/op        fastest
 100000 143.52  (  6.97ms) (± 2.51%)  5772938 B/op   1.00× slower
1000000 143.34  (  6.98ms) (± 2.78%)  5772918 B/op   1.00× slower
1093951 143.52  (  6.97ms) (± 2.78%)  5772934 B/op   1.00× slower

Worst-case Waste: 23.08%
```

### Growth Factor: 1.2
```
   1000 130.29  (  7.68ms) (± 3.22%)  6962664 B/op   1.00× slower
  10000 130.62  (  7.66ms) (± 3.17%)  6962665 B/op        fastest
 100000 130.58  (  7.66ms) (± 3.06%)  6962639 B/op   1.00× slower
1000000 130.46  (  7.67ms) (± 3.25%)  6962651 B/op   1.00× slower
1048577 130.31  (  7.67ms) (± 3.13%)  6962664 B/op   1.00× slower

Worst-case Waste: 16.67%
```

### Growth Factor: 1.1
```
   1000 137.98  (  7.25ms) (± 2.09%)  10886345 B/op        fastest
  10000 137.76  (  7.26ms) (± 1.92%)  10886342 B/op   1.00× slower
 100000 137.94  (  7.25ms) (± 2.07%)  10886350 B/op   1.00× slower
1000000 137.72  (  7.26ms) (± 1.98%)  10886365 B/op   1.00× slower
1004476 132.47  (  7.55ms) (± 3.36%)  10886350 B/op   1.04× slower

Worst-case Waste: 9.09%
```

### DIFF
--- a/src/io/memory.cr
+++ b/src/io/memory.cr
@@ -3,11 +3,14 @@
 # The internal buffer can be resizeable and/or writeable depending
 # on how an `IO::Memory` is constructed.
 class IO::Memory < IO
+  GROWTH_FACTOR = 1.4
+
   # Returns the internal buffer as a `Pointer(UInt8)`.
   getter buffer : Pointer(UInt8)
 
   # Same as `size`.
   getter bytesize : Int32
+  getter capacity : Int32
 
   @capacity : Int32
 
@@ -82,7 +85,7 @@ class IO::Memory < IO
 
   # See `IO#write(slice)`. Raises if this `IO::Memory` is non-writeable,
   # or if it's non-resizeable and a resize is needed.
-  def write(slice : Bytes)
+  def write(slice : Bytes) : Nil
     check_writeable
     check_open
 
@@ -93,7 +96,7 @@ class IO::Memory < IO
     new_bytesize = @pos + count
     if new_bytesize > @capacity
       check_resizeable
-      resize_to_capacity(Math.pw2ceil(new_bytesize))
+      resize_to_fit(new_bytesize)
     end
 
     slice.copy_to(@buffer + @pos, count)
@@ -104,20 +107,18 @@ class IO::Memory < IO
 
     @pos += count
     @bytesize = @pos if @pos > @bytesize
-
-    nil
   end
 
   # See `IO#write_byte`. Raises if this `IO::Memory` is non-writeable,
   # or if it's non-resizeable and a resize is needed.
-  def write_byte(byte : UInt8)
+  def write_byte(byte : UInt8) : Nil
     check_writeable
     check_open
 
     new_bytesize = @pos + 1
     if new_bytesize > @capacity
       check_resizeable
-      resize_to_capacity(Math.pw2ceil(new_bytesize))
+      resize_to_fit(new_bytesize)
     end
 
     (@buffer + @pos).value = byte
@@ -128,8 +129,6 @@ class IO::Memory < IO
 
     @pos += 1
     @bytesize = @pos if @pos > @bytesize
-
-    nil
   end
 
   # :nodoc:
@@ -418,23 +417,34 @@ class IO::Memory < IO
     io.write(to_slice)
   end
 
-  private def check_writeable
+  private def check_writeable : Nil
     unless @writeable
       raise IO::Error.new "Read-only stream"
     end
   end
 
-  private def check_resizeable
+  private def check_resizeable : Nil
     unless @resizeable
       raise IO::Error.new "Non-resizeable stream"
     end
   end
 
-  private def check_needs_resize
-    resize_to_capacity(@capacity * 2) if @bytesize == @capacity
+  private def check_needs_resize : Nil
+    resize_to_capacity(@capacity * GROWTH_FACTOR) if @bytesize == @capacity
   end
 
-  private def resize_to_capacity(capacity)
+  private def resize_to_fit(size : Int32) : Nil
+    return if ( size <= capacity )
+
+    capacity = @capacity * GROWTH_FACTOR
+    while ( size > capacity )
+      capacity *= GROWTH_FACTOR
+    end
+
+    return resize_to_capacity(capacity.to_i)
+   end
+
+  private def resize_to_capacity(capacity : Int32) : Nil
     @capacity = capacity
     @buffer = @buffer.realloc(@capacity)
   end


chris-huxtable:
My testing code:
```crystal
require "benchmark"

title = "Growth Factor: #{IO::Memory::GROWTH_FACTOR}"
puts title
puts ("=" * title.size)

Benchmark.ips { |x|
  mem = IO::Memory.new()
  count = 1_000
  x.report(count.to_s) {
    mem = IO::Memory.new()
    count.times() { mem << 'a' }
  }

  count = 10_000
  x.report(count.to_s) {
    mem = IO::Memory.new()
    count.times() { mem << 'a' }
  }

  count = 100_000
  x.report(count.to_s) {
    mem = IO::Memory.new()
    count.times() { mem << 'a' }
  }

  count = 1_000_000
  x.report(count.to_s) {
    mem = IO::Memory.new()
    count.times() { mem << 'a' }
  }

  capacity = 1
  while ( count > capacity )
    capacity *= IO::Memory::GROWTH_FACTOR
  end
  count = (capacity + 1).to_i
  x.report(count.to_s) {
    mem = IO::Memory.new()
    count.times() { mem << 'a' }
  }
}
puts

count = 1_000_000
mem = IO::Memory.new()
count.times() { mem << 'a' }

count = mem.capacity + 1
mem = IO::Memory.new()
count.times() { mem << 'a' }
puts "Worst-case Waste: #{((1.0 - (mem.bytesize.to_f / mem.capacity.to_f))*100).round(2)}%"
```

src/io/memory.cr
```diff
@@ -418,23 +417,34 @@ class IO::Memory < IO
...
 
-  private def check_needs_resize
-    resize_to_capacity(@capacity * 2) if @bytesize == @capacity
+  private def check_needs_resize : Nil
+    resize_to_capacity(@capacity * GROWTH_FACTOR) if @bytesize == @capacity
   end
 
-  private def resize_to_capacity(capacity)
+  private def resize_to_fit(size : Int32) : Nil
+    return if ( size <= capacity )
```
Sija: Needless `()`

src/io/memory.cr
```diff
@@ -418,23 +417,34 @@ class IO::Memory < IO
...
+
+    capacity = @capacity * GROWTH_FACTOR
+    while ( size > capacity )
```
Sija: ditto

src/io/memory.cr
```diff
@@ -418,23 +417,34 @@ class IO::Memory < IO
...
+      capacity *= GROWTH_FACTOR
+    end
+
+    return resize_to_capacity(capacity.to_i)
```
Sija: needless `return`

src/io/memory.cr
chris-huxtable: Always raining on my parade :P. Two min :)

Sija:
Btw, similar optimization could be also applied to `Array`.

chris-huxtable:
Yep :)

I want to sort it out here first then apply it anywhere else it could be useful.

src/io/memory.cr
```diff
@@ -418,23 +417,34 @@ class IO::Memory < IO
...
+    capacity = @capacity * GROWTH_FACTOR
+    while size > capacity
+      capacity *= GROWTH_FACTOR
+    end
+
+    resize_to_capacity(capacity.to_i)
   end
 
-  private def resize_to_capacity(capacity)
+  private def resize_to_capacity(capacity : Int32) : Nil
```
Sija: Is it safe to assume that capacity is an `Int32` (and not bigger)?

chris-huxtable: I was wondering about that as well... I personally would set it to `UInt32` if not `UInt64`. Is there an equivalent to `size_t`?

chris-huxtable:
I am having 2 of my machines do every value 1.1 to 2.0 at steps of 0.01 just of the off chance there might be something better. The results I will put in Gists.

- (macOS 10.13)[https://gist.github.com/chris-huxtable/bb44cef3f008b3e36a465a7fc3d0f77d]
- (OpenBSD 6.3)[https://gist.github.com/chris-huxtable/0e67ee9c575b3e73e6b6f4e7886ada2e]

I will update them with the full results in the morning.

src/io/memory.cr
straight-shoota: Currently, yes, it's limited to `Int32`. See #3209.

ysbaddaden:
Don't run benchmarks but analyze bdwgc to understand how it does allocate memory, and more specifically how much more memory does it allocate than required.

Problem is, once you're tied to a GC memory model, it will fall flat with another memory allocator. For example with my Immix GC the allocation is always rounded to a word size, that is 4 or 8 bytes depending on 32 or 64-bit arch. You may be lucky and have enough room after the current allocation, but it's random chance.

chris-huxtable:
I intend to go deeper when I have time. If it’s just a constant I don’t think it would be to difficult to set it based on the GC’s memory model.

ysbaddaden:
In fact, with Immix GC I always reallocate and will always (unless the size didn't change), because the current thread allocator probably doesn't own the memory block anymore.

So, to me, the growth factor is mostly about growing enough vs wasting space in common collection use cases.

chris-huxtable:
Looking further into this issue its largely about wasting space. That being said The growth coefficient should still be less then 2.0 so that reuse can potentially happen. I know @ysbaddaden has said that IMMIX doesn't reuse its memory but as far as I have been able to ascertain boehm does as do many others. If we switch from boehm I think this should be re-evaluated.

There is an interesting discussion on growth rates [on stack overflow](https://stackoverflow.com/questions/1100311/what-is-the-ideal-growth-rate-for-a-dynamically-allocated-array#1100426) which also references the FBVector like I referenced above .

Until we switch from boehm I like the Idea of using Phi (φ) 1.618033987 as its less then 2.0 allowing reallocation, but will have an intermediate rate of growth keeping reallocations low. Also because its the golden ratio.

wontruefree:
@chris-huxtable are you still working on this I am interested if this can still be a performance improvement.

chris-huxtable:
I am not currently. But I could take another shot at it if there is interest.

wontruefree:
I mean I am interested in it.  Maybe someone from the core team could weight in.  I think some preliminary benchmarks might be compelling.

chris-huxtable:
It's kind of a difficult thing to test. If you don't inter-splice a number of mallocs as you grow the array when testing and manage to do so consistently you end up with no difference (realloc involves no copying) or inconstant results (realloc inconsistently copies.

chris-huxtable:
Beyond performance improvements I think it makes sense to go to a more conservative growth factor just for memory efficiency. 

@ysbaddaden, @straight-shoota, anyone else relevant - Would a PR which sets a more conservative growth factor of Phi (φ) 1.618033987 be accepted in principle, assuming it works and meets any other standards of quality?

straight-shoota:
Sure, performance optimizations are always welcome.

It's just optimizations can easily be added later, and I think *at the moment* we'd like to put a stronger focus on bug fixes and features. Feel free to go ahead, though.

chris-huxtable:
@straight-shoota - Sounds good. Just don't want to waste my time and end up in a 9 mouth back and forth about it.

rdp:
Are there any benchmarks showing performance improvement?

chris-huxtable:
It's hard to contrive a comparison. It's really more about efficiency.

caspiano:
I believe this can be closed in favour of #11482

straight-shoota:
They're definitely related. But this one is about `IO::Memory`, while #11482 only affected `Array`.

PR #6531, opened 2018-08-13
Fix Object#def_hash when passing Symbols

Sija:
When passing `Symbol` to `Object#def_hash` macro it'll hash that instead of resolving it, this PR fixes this behaviour.

Relevant spec - https://carc.in/#/r/4q0k


straight-shoota:
Why don't you include that spec in the PR?

PR #6533, opened 2018-08-13
Add tls scheme to HTTP::Server#bind and rename HTTP::Server#bind_ssl

straight-shoota:
This adds `tls://` scheme for `HTTP::Server#bind` and makes it default in documentation. It also renames `HTTP::Server#bind_ssl`to `HTTP::Server#bind_tls`.

Follow-up on #6500 (discussion starting at https://github.com/crystal-lang/crystal/pull/6500#issuecomment-412087373)

### DIFF
--- a/spec/std/http/server/server_spec.cr
+++ b/spec/std/http/server/server_spec.cr
@@ -351,9 +351,13 @@ module HTTP
           begin
             port = unused_port
             expect_raises(ArgumentError, "missing CA certificate") do
-              server.bind "ssl://127.0.0.1:#{port}?key=#{private_key}&cert=#{certificate}&verify_mode=force-peer"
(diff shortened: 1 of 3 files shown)


jhass:
> Do we want to make such a minimal breaking change to an API just introduced?
Yes, I think its fair. People using it should still be actively working on their projects.

straight-shoota:
Yeah, it's too bad 0.26.0 is already done.

Maybe something breaks and we need to ship 0.26.1 right away, then this could be included in the bug fix ;)

bcardiff:
Kemal already update it's api to use bind_ssl. 
We should add bind_ssl for 0.26.x so we can ship master as it and remove bind_ssl in 0.27

straight-shoota:
You mean to temporarily re-add `bind_ssl` as alias for `bind_tls`?

bcardiff:
Yes.

PR #6538, opened 2018-08-14
Fix ASTNode#to_s of non-unary operator call without argument

### LINKED ISSUE #6537: After defining a method named "%", the playground gives an error requesting to report a bug...
...so I'm reporting it. Doesn't really seem like a big deal to me, maybe the error is supposed to be caught and displayed and that isn't being done properly? I was just trying to see if the language would let me define a method named a percent literal. The error message in the terminal:

    E, [2018-08-13 20:11:03 -04:00 #28538] ERROR -- : Instrumention bug found (session=1, tag=21).

In the playground:

    Syntax error in play:50: unexpected token: %

the code that triggered the errror:

```crystal
    struct Clouds
      JSON.mapping( all: Int32 | Float32 )
      def %
        all
      end
    end
    c = Clouds.from_json(%({ "all": 123 }))
    c.%
```

MakeNowJust:
Fixed #6537 

Currently this code shows wrong result:

```crystal
class Foo
  def self.*
    42
  end
end

p! Foo.*
```

```console
$ crystal run foo.cr
*Foo # => 42
```

Of course `*Foo` is invalid Crystal code. It is the reason of #6537.

It is fixed by this PR.

```console
$ bin/crystal run foo.cr
Foo.* # => 42
```


src/compiler/crystal/syntax/to_s.cr
```diff
@@ -362,8 +362,15 @@ module Crystal
         @str << ' '
         node.args.last.accept self
       elsif node_obj && !letter_or_underscore?(node.name) && node.args.size == 0
-        @str << decorate_call(node, node.name)
-        in_parenthesis(need_parens, node_obj)
+        if node.name == "+" || node.name == "-" || node.name == "~"
```
Sija: Shouldn't `!` be checked as well?

MakeNowJust:
ping

PR #6539, opened 2018-08-14
Remove {% if properties.size > 0 %} conditions

r00ster91:
Since #6367 `{% if properties.size > 0 %}` conditions aren't needed anymore.


r00ster91:
Oh I forgot CI is still using 0.25.1.

PR #6541, opened 2018-08-15
Allow to call `~` with `foo.~` style

MakeNowJust:
Currently `~` method cannot call with `foo.~` style.

```crystal
1.~
# Syntax error in foo.cr:1: expecting any of these tokens: IDENT, CONST, +, -, *, /, %, |, &, ^, **, <<, <, <=, ==, !=, =~, !~, >>, >, >=, <=>, ===, [], []=, []?, [ (not '~')
#
# 1.~
#   ^
#
```

All operators except `~` can call with such style. So I thought it is bug and fixed this.

Thank you.


src/compiler/crystal/syntax/to_s.cr
```diff
@@ -367,12 +367,19 @@ module Crystal
       elsif node_obj && !letter_or_underscore?(node.name) && node.args.size == 1
         in_parenthesis(need_parens, node_obj)
 
-        @str << ' '
-        @str << decorate_call(node, node.name)
-        @str << ' '
-
         arg = node.args[0]
-        in_parenthesis(need_parens(arg), arg)
+        if node.name == "~"
```
RX14: Why is `~` special and needs a special branch?

MakeNowJust: `1.~(2)` is valid syntax now, but `ASTNode#to_s` does not expect this, so it returns `1 ~ 2` in this case. It is for such a case.

RX14: I guess i'm asking is what makes `1.~(2)` special instead of `1.+(2)` that it needs to be different here

MakeNowJust: `1 + 2` is valid, but `1 ~ 2` is invalid.

RX14: ah, makes sense

PR #3972, opened 2017-02-02
Fix self restriction with including generic module

MakeNowJust:
Ref: https://github.com/crystal-lang/crystal/pull/3847#issuecomment-270888186

Now, we can get a compile error with such a code:

```crystal
module Foo(T)
  def foo(x : T)
    x
  end
end

abstract struct Bar
  include Foo(self)
end

struct Baz1 < Bar
end

struct Baz2 < Bar
end

Baz1.new.foo Baz2.new # => no overload matches 'Baz1#foo' with type Baz2
```

This pull request adds `lazy_self` parameter to `lookup_type`. When `lazy_self` is `true`, `lookup_type` keeps `self` in generics type. It is used to look up type for `include` and `extend`.

PR #6549, opened 2018-08-16
Avoid using Comparable(self) in reference types

### LINKED ISSUE #6547: crystal/src/string.cr:139: can't declare variable with Self
This is a high severity bug that luckily seems not to impact most projects. Unfortunately, it is making it so Lucky will not work on 0.26 though. I'm giving a workshop on Crystal and Lucky tomorrow so I would love to fix this before then so everyone can build something 😜 but I know we're all volunteers so if the core team doesn't have time that is ok! I'll figure something out. 

### The problem

On Crystal 0.26, compiling lucky_record fails with this error:

```
Error in /Users/paulcsmith/projects/crystal-lang/crystal/src/string.cr:139: can't declare variable with Self

  include Comparable(self)
```

Note: This is the whole trace, even with `--verbose`. I'm not sure what in LuckyRecord/Crystal is causing this.

### Steps to reproduce

* Crystal 0.26
* Clone https://github.com/luckyframework/lucky_record
* `shards install`
* `crystal spec`
* See error

### Oddities

It seems that the error is in Crystal, but that particular line of code it is referencing (src/string.cr:139) hasn't changed in years.

I'm having a hard time even figuring out what is wrong since it fails so fast and there is no trace. I tried using `--verbose` and still didn't see anything. I'm not sure how to debug this

bcardiff:
Fixes #6547.

This is a workaround fix.

There seems to be an issue with type guessed ivars and generics and self that I'm trying to narrow it down. But maybe it also involves macros that lucky quite much since I wasn't able to frame it yet.

Meanwhile this PR removes `include Comparable(self)` from reference types (leaving it only in Enum and some specs).

### DIFF
--- a/src/compiler/crystal/syntax/location.cr
+++ b/src/compiler/crystal/syntax/location.cr
@@ -2,7 +2,7 @@ require "../../../partial_comparable"
 
 # A location of an `ASTnode`, including its filename, line number and column number.
 class Crystal::Location
-  include PartialComparable(self)
+  include PartialComparable(Crystal::Location)
 
   getter line_number
   getter column_number
--- a/src/pointer.cr
+++ b/src/pointer.cr
@@ -52,7 +52,7 @@ struct Pointer(T)
     end
   end
 
-  include Comparable(self)
+  include Comparable(Pointer(T))
 
   # Returns `true` if this pointer's address is zero.
   #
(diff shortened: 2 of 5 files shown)


straight-shoota:
I like this change anyway because it is more explicit.

bcardiff:
FYI, related discussion in https://github.com/crystal-lang/crystal/pull/3972#discussion_r210792069 that leads to revert the lazy evaluation of self.

bcardiff:
With #6557 merges this can be closed actually.

PR #6551, opened 2018-08-16
Restore #bind_ssl as deprecated api

bcardiff:
Restores `HTTP::Server#bind_ssl` removed by #6533 so the current master can be used to ship 0.26.1 without breaking changes.

`HTTP::Server#bind_ssl` can be removed in 0.27

### DIFF
--- a/src/http/server.cr
+++ b/src/http/server.cr
@@ -258,6 +258,11 @@ class HTTP::Server
     def bind_tls(address : Socket::IPAddress, context : OpenSSL::SSL::Context::Server) : Socket::IPAddress
       bind_tls(address.address, address.port, context)
     end
+
+    # DEPRECATED: Use `#bind_tls`.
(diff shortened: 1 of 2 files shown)


src/http/server.cr
```diff
@@ -258,6 +258,11 @@ class HTTP::Server
     def bind_tls(address : Socket::IPAddress, context : OpenSSL::SSL::Context::Server) : Socket::IPAddress
       bind_tls(address.address, address.port, context)
     end
+
+    # DEPRECATED: Use `#bind_tls`.
```
RX14: Please add an explicit `TODO: remove in 0.27.0` or similar

we should probably eventually settle on a `TODO(0.27.0)` style with a `grep -R "TODO($(crystal -v))" src spec` to fail the build.

bcardiff: done.

PR #6552, opened 2018-08-16
Add positive?/negative? to Number and Time::Span

mamantoha:
Hi! I know we don't want to copy every Ruby-ism in Crystal stdlib. And I will understand if you reject this PR.

But `Number` already has `even?`, `odd?`, and `zero?`. So I think `positive?` and `negative?` fits into that group of functions.

Vote for this pull request :+1: or :-1:

### DIFF
--- a/spec/std/complex_spec.cr
+++ b/spec/std/complex_spec.cr
@@ -180,4 +180,11 @@ describe "Complex" do
     Complex.new(1.2, 0).zero?.should eq false
     Complex.new(1.2, 3.4).zero?.should eq false
   end
+
+  it "test positive?" do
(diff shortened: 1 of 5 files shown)


straight-shoota:
Could you elaborate in actual use cases and how it is better than `Number#sign`?

j8r:
`#positive` and `#negative` are confusing, a _strict_ mention has to be added for correctness because there is a loosely interpretation possible (https://en.wikipedia.org/wiki/Strict)
Adding `strictly_` to the method name will render it uglier.

I don't see the value added, and I don't know any language that has this.

PR #6553, opened 2018-08-16
Refactor ISO 8601 parser for ordinal date

straight-shoota:
It is much easier to just add the amount of days.

### DIFF
--- a/src/time/format/custom/iso_8601.cr
+++ b/src/time/format/custom/iso_8601.cr
@@ -52,19 +52,11 @@ struct Time::Format
         else
           # ordinal date
           day_of_the_year = @month * 10 + current_char.to_i
-          @month = 0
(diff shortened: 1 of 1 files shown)


asterite:
Benchmark for before and after?

straight-shoota:
This refactoring isn't focused on performance but to make the code easier to understand. I didn't do any benchmarks. But I can try to see how it goes and maybe improve it further (`Time.new` and immediately adding days can probably be optimized).

straight-shoota:
Hm, well okay. The previous version was actually pretty fast. It seems going through `Time` constructor and then `#year_month_day_day_year` can't beat the performance. I couldn't get it faster than 50% of the original speed.

So never mind then, I guess that's not worth changing it.

PR #6555, opened 2018-08-17
Change Time::DayOfWeek to ISO ordinal numbering based on Monday = 1

### LINKED ISSUE #6554: DayOfWeek should be compatible to ordinal day numbers from ISO 8601
There exist two common schemes for numbering week days: either starting with sunday = 0 or monday = 1. The enum values of `Time::DayOfWeek` are `Sunday = 0, Monday = 1, ..., Saturday = 6`. The ordinal day numbers in [ISO 8601](http://xml.coverpages.org/ISO-FDIS-8601.pdf) however are `Monday = 1, ..., Saturday = 6, Sunday = 7`.
This conversion is pretty simple and besides sunday all numerical representations are the same in both numberings.

However, there should be a direct way to use ISO numbers without having to implement any conversion manually.

I'd even favour the internal representation to follow ISO `Monday = 1` because this is the most commonly used standard. I propose to change the enum values accordingly. Obviously, this is a breaking change.

```cr
enum DayOfWeek
  Monday    = 1
  Tuesday   = 2
  Wednesday = 3
  Thursday  = 4
  Friday    = 5
  Saturday  = 6
  Sunday    = 7
end
```

In order to also support 0-based numbering, additional methods should be added to the enum. For example `#value_sun0` (I don't have any more poetic idea) which returns `value % 7`. There should also be an equivalent constructor for 0-based numbers. Instead of an additional one, it might be better to simply alter `from_value` (and maybe `new`) to map `0` to `Sunday`. This is still unambiguous and the breaking change is reduced.

The alternative - if we don't want to change the internal values of `DayOfWeek` - would be to add `#to_iso_week_day` and `.from_iso_week_day` providing the required conversions. This avoids a breaking change.

*I hope this doesn't spark a discussion about which day is the first day of the week. There are different views and all are right. We just need to assign numbers to represent week days and besides technical reasons it shouldn't matter how.*

straight-shoota:
Fixes #6554

`DayOfWeek.from_value` also identifies `0` as `Sunday` making it work with both numbering schemes. I don't think this should pose a problem.
`DayOfWeek.new` still requires the proper value, meaning `0` is not recognized.

I did not add a method to retrieve the value based on sunday = 0 (initially proposed as `#value_sun0`) because I think it is much easier and maybe even better understandable to just use `value % 7`.

### DIFF
--- a/spec/std/time/time_spec.cr
+++ b/spec/std/time/time_spec.cr
@@ -548,6 +548,39 @@ describe Time do
     end
   end
 
+  describe Time::DayOfWeek do
+    it "#value" do
+      Time::DayOfWeek::Monday.value.should eq 1
+      Time::DayOfWeek::Tuesday.value.should eq 2
+      Time::DayOfWeek::Wednesday.value.should eq 3
+      Time::DayOfWeek::Thursday.value.should eq 4
+      Time::DayOfWeek::Friday.value.should eq 5
+      Time::DayOfWeek::Saturday.value.should eq 6
+      Time::DayOfWeek::Sunday.value.should eq 7
+    end
+
+    it ".from_value" do
+      Time::DayOfWeek.from_value(1).should eq Time::DayOfWeek::Monday
+      Time::DayOfWeek.from_value(2).should eq Time::DayOfWeek::Tuesday
+      Time::DayOfWeek.from_value(3).should eq Time::DayOfWeek::Wednesday
+      Time::DayOfWeek.from_value(4).should eq Time::DayOfWeek::Thursday
+      Time::DayOfWeek.from_value(5).should eq Time::DayOfWeek::Friday
+      Time::DayOfWeek.from_value(6).should eq Time::DayOfWeek::Saturday
+      Time::DayOfWeek.from_value(7).should eq Time::DayOfWeek::Sunday
+
+      # Special case: Identify 0 as Sunday
+      Time::DayOfWeek.from_value(0).should eq Time::DayOfWeek::Sunday
+
+      expect_raises(Exception, "Unknown enum Time::DayOfWeek value: 8") do
+        Time::DayOfWeek.from_value(8)
+      end
+    end
+
+    it ".new does not identify 0 as Sunday" do
+      Time::DayOfWeek.new(0).should_not eq Time::DayOfWeek::Sunday
+    end
+  end
+
   typeof(Time.now.year)
   typeof(1.minute.from_now.year)
   typeof(1.minute.ago.year)
--- a/src/time.cr
+++ b/src/time.cr
@@ -65,6 +65,15 @@ require "crystal/system/time"
 # time.time_of_day # => 10:20:30
 # ```
 #
+# For querying if a time is at a specific day of week, `Time` offers named
+# predicate methods, delegating to `#day_of_week`:
+#
+# ```
+# time.monday? # => true
+# # ...
+# time.sunday? # => false
+# ```
+#
 # ### Time Zones
 #
 # Each time is attached to a specific time zone, represented by a `Location`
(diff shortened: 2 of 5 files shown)


src/time.cr
```diff
@@ -254,27 +263,35 @@ struct Time
...
+    Friday    = 5
+    Saturday  = 6
+    Sunday    = 7
+
+    # Returns the day of week that has the given value, or raises if no such member exists.
+    #
+    # This method also accepts `0` to identify `Sunday` in order to be compliant
+    # with the `Sunday = 0` numbering. All other days are equal in both formats.
+    def self.from_value(value : Int32)
+      value = (value + 6) % 7 + 1
```
RX14: surely `return Sunday if value == 0` would be better here?

rodrigopinto: What should happen when the value is greater than `7`? 

The current implementation pass for the test below, is it the expected behavior?
```crystal
Time::DayOfWeek.from_value(8).should eq Time::DayOfWeek::Monday
Time::DayOfWeek.from_value(9).should eq Time::DayOfWeek::Tuesday
# and goes on...
```

spec/std/time/time_spec.cr
```diff
@@ -1027,6 +1027,35 @@ describe Time do
     end
   end
 
+  describe Time::DayOfWeek do
+    it "#value" do
```
rodrigopinto: I believe `describe` would fit better to the method definition.

spec/std/time/time_spec.cr
```diff
@@ -1027,6 +1027,35 @@ describe Time do
...
+      Time::DayOfWeek::Monday.value.should eq 1
+      Time::DayOfWeek::Tuesday.value.should eq 2
+      Time::DayOfWeek::Wednesday.value.should eq 3
+      Time::DayOfWeek::Thursday.value.should eq 4
+      Time::DayOfWeek::Friday.value.should eq 5
+      Time::DayOfWeek::Saturday.value.should eq 6
+      Time::DayOfWeek::Sunday.value.should eq 7
+    end
+
+    it ".from_value" do
```
rodrigopinto: ditto.

src/time.cr
```diff
@@ -690,10 +707,10 @@ struct Time
     Span.new(nanoseconds: NANOSECONDS_PER_SECOND * (offset_seconds % SECONDS_PER_DAY) + nanosecond)
   end
 
-  # Returns the day of the week (`Sunday..Saturday`).
+  # Returns the day of the week (`Monday..Sunday`).
   def day_of_week : Time::DayOfWeek
-    value = ((offset_seconds / SECONDS_PER_DAY) + 1) % 7
-    DayOfWeek.new value.to_i
+    days = offset_seconds / SECONDS_PER_DAY
+    DayOfWeek.new days.to_i % 7 + 1
```
rodrigopinto: Do not want sound nit-picking, but I was thinking if would make sense to break it in two methods, one private to return the weekday :

```crystal
def day_of_week : Time::DayOfWeek
  DayOfWeek.new(weekday_number)
end

private def weekday_number
  days = offset_seconds / SECONDS_PER_DAY
  days.to_i % 7 + 1
end
```

asterite:
What's the reason it works for 0? I would just leave the default Enum implementation.

src/time.cr
straight-shoota: I think either makes some kind of sense. But we can make it more strict, no problem.

spec/std/time/time_spec.cr
straight-shoota: This is a single test case including multiple expectations. Do you think each line should be it's own test case?

src/time.cr
straight-shoota: What would be the benefit of this additional method?

straight-shoota:
> What's the reason it works for 0?

`Sunday = 0` is also commonly used besides the standard numbering starting at `Monday = 1`. For example the time format directive `%w` uses this.

That why there should be a way to retrieve the week day from a 0-numbered index. As expressed in #6554, this doesn't need to be an override of `from_value`, there could also just be a separate method for this (like `from_value_sun0` or similar). I just figured it would be easier to also interpret `from_value(0)` as `Sunday`.

src/time.cr
rodrigopinto: The benefit I see is readability because we do two arithmetical operations to get the number of the day of the week. 

From the perspective who is reading the code, to read `DayOfWeek.new(weekday_number)` instead of `DayOfWeek.new(days.to_i % 7 + 1)` becomes more self explanatory.

spec/std/time/time_spec.cr
rodrigopinto: No. It makes sense to keep all of them in a single test case, I was just wondering `describe` fits better with the methods names and the`it` could be a description of the expectation, that would be a "documentation for free".

i.e: 

```crystal
describe "#value" do
  it "returns the number representation of the weekday" do
    #...
  end
end
```

what do you think?

straight-shoota: I generally appreciate expressiveness, but in this case I think it's just needlessly overly verbose. What's the benefit of repeating the description of the method which is already documented with its definition? If there are multiple test cases, yes, then it makes sense to describe them more explicitly. But this is just a simple spec that ensures `#value` behaves like it is supposed to.

src/time.cr
straight-shoota: Hm, then fitting in an appropriately named variable name would do the same result.

straight-shoota:
@asterite Sounds fair.

PR #6556, opened 2018-08-17
Standardize methods between JSON::Any/YAML::Any

j8r:
Details:
- Adds `#as_bool?` and `#as_bool` to `YAML::Any`
- Separate `Int32`/`Int64` and `Float32`/`Float64` tests in the `JSON::Any` specs
- Standardize and shorten conditions for nillable variants of cast methods

### DIFF
--- a/spec/std/yaml/any_spec.cr
+++ b/spec/std/yaml/any_spec.cr
@@ -7,6 +7,14 @@ describe YAML::Any do
       YAML.parse("").as_nil.should be_nil
     end
 
+    it "gets bool" do
+      YAML.parse("true").as_bool.should be_true
+      YAML.parse("false").as_bool.should be_false
+      YAML.parse("true").as_bool?.should be_true
+      YAML.parse("false").as_bool?.should be_false
+      YAML.parse("2").as_bool?.should be_nil
+    end
+
     it "gets string" do
       YAML.parse("hello").as_s.should eq("hello")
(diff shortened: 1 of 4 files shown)


j8r:
I'm wondering, why not having `as_b` instead of `as_bool` ?

spec/std/yaml/any_spec.cr
```diff
@@ -7,6 +7,14 @@ describe YAML::Any do
...
     end
 
+    it "gets bool" do
+      YAML.parse("true").as_bool.should be_true
+      YAML.parse("false").as_bool.should be_false
+      YAML.parse("true").as_bool?.should be_true
+      YAML.parse("false").as_bool?.should be_false
+      YAML.parse("2").as_bool?.should be_nil
+    end
+
```
vladfaust: Did you know that "yes" and "no" are considered booleans in YAML? http://yaml.org/type/bool.html

j8r: Surprising :open_mouth: 
If we stick with the YAML 1.2 spec, this isn't the case any more http://yaml.org/spec/1.2/spec.html#id2803629

vladfaust: Yeah,

> In this version of YAML, boolean values may be serialized as “true” or “false”; the empty scalar as “null”.

Totally makes sense, TIL!

PR #6560, opened 2018-08-18
Fix: check presence of named argument via external name

### LINKED ISSUE #6559: Obscure error when using named argument with reserved name
This used to work before Crystal 0.26

```crystal
class C
  def initialize(@abstract : Bool = false, *args, **kwargs)
  end
end

func = C.new(abstract: true)
```

But now it gives an error

```
Error in line 6: read before assignment to local variable '__temp_270'
```

https://carc.in/#/r/4ro0

Maybe it shouldn't have worked but this kind of error message will not suffice.

asterite:
Fixes #6559

It seems the bug was always there but it became apparent when we changed the way `def foo(@abstract)` and similar constructs with keywords changed their implementation.


j8r:
@RX14 won't be happy @ysbaddaden, labels are missing :)

PR #6570, opened 2018-08-20
Iterator could chain indetermined number of iterators now

### LINKED ISSUE #6532: Chain all iterator in array that isn't determinate in compilation-time
I have an array of iterators depends on user's input, so cannot be determinate in compilation-time, and now  I want to chain all of them together.

``` crystal
arrs = [] of Array(Int32)
gets.not_nil!.to_i.times {|a| arrs << [a] }

def chain_all1(arrs)
    arrs.each.map(&.each).reduce{|a,b| a.chain b}
end
def chain_all2(arrs)
    iter = arrs.each
    ret = iter.next
    return if ret.is_a? Iterator::Stop
    ret = ret.each
    until (a = iter.next).is_a? Iterator::Stop
        ret = ret.chain a.each
    end
    ret
end
chain_all1(arrs).each {|a| p a }
chain_all2(arrs).each {|a| p a }
```

The first one `chain_all1` triggers a compilation error : 
```
type must be Indexable::ItemIterator(Array(Int32), Int32), not (Indexable::ItemIterator(Array(Int32), Int32) | Iterator::Chain(Indexable::ItemIterator(Array(Int32), Int32), Indexable::ItemIterator(Array(Int32), Int32), Int32, Int32))
```

And for the second one `chain_all2`, compiler traps into a long time silence and ends in `generic type too nested`.

It's yet another recursive type problem, because Iterator tries to expand all code and it's impossible.

But 'chains all' is in fact practicable, by a struct stores all the iterators and implements `next` and `rewind` via a loop.

The code to implement a struct could be long, so I think it's should be implemented in std-lib, and a method `chain(iterators : Enumerable(Iterator))` may be provided.

xqyww123:
Fixes #6532


straight-shoota:
Did you consider merging `ChainsAll` with the existing `Chain` type? It could probably just be replaced with the new implementation which could be used by both `.chain` and `#chain`.

RX14:
https://carc.in/#/r/4s6t

xqyww123:
Yeah, I considered it. Although they both implement the same function, the performance is different. Type `Chain` is linear expanded, each `next` call to the sub-iterator is static bound, and many optimizing may be produced by compiler automatically. But `ChainAll` expands a loop, `next` call is virtual bound, no too much optimizing could be made. So theoretically, `Chain` would be faster than `ChainAll`.

However, I resolved the original `Chain`'s code. It's not an efficient implementation. If one tries to concat `N` iterators, `N` if-branch and `N` bool variables will be used. It's not something good.

After some thinking, I figure out the better implementation may use `Tuple` and macro, to generate   `switch-case` like branch using an index variable, then only one variable and one branch required for `N` concat.

But it will be more complex also... Why not make it simple and stupid? I'm going to merge `Chain` with `ChainAll` since current original implementation is as slow as a loop.

RX14:
Yes, i'm fine with this implementation, I'm still very concerned that this problem isn't unique to `Chain` though.

xqyww123:
Finally, I remain `#chain` in the same, for the purpose of performance, but added some comments to instruct this feature.
Maybe I had tried my best to improve linguistics, but I'm sorry I'm not good at foreign languages. If it's still not good enough, helps is really really appreciated, thanks.

src/iterator.cr
```diff
@@ -160,6 +160,11 @@ module Iterator(T)
 
   # Returns an iterator that returns elements from the original iterator until
   # it is exhausted and then returns the elements of the second iterator.
+  # Compared to `.chain`, it has better performance when the quantity of
```
RX14: > Compared to `#chain(Iterator(Iter))`

src/iterator.cr
```diff
@@ -160,6 +160,11 @@ module Iterator(T)
...
+  # iterators to chain is small (usually less than 4).
+  # One could also face a problem trying to chain several iterators in loop(like `Enumerable#reduce`),
```
RX14: > This method also cannot chain iterators in a loop, for that see `#chain(Iterator(Iter))`.

src/iterator.cr
```diff
@@ -200,6 +205,58 @@ module Iterator(T)
     end
   end
 
+  # The same as `#chain`, but have better performance when the quantity of
```
RX14: >but has better performance

src/iterator.cr
```diff
@@ -200,6 +205,58 @@ module Iterator(T)
...
+  # iterators to chain is large(usually greater than 4) or undetermined.
```
RX14: >large (usually

src/iterator.cr
```diff
@@ -200,6 +205,58 @@ module Iterator(T)
...
+  # It uses lesser type specification for more freedom to avoid the failure (at a cost of performance),
+  # and a loop rather than recursive calling to achieve better performance for numbers of iterators
+  # (but when the quantity is small, it's not really efficiency).
```
RX14: We don't tend to discuss implementation details in docs, only user-visible effects, like performance and ability to chain in a loop.

spec/std/iterator_spec.cr
```diff
@@ -59,6 +59,30 @@ describe Iterator do
       iter.rewind
       iter.to_a.should eq([1, 2, 'a', 'b'])
     end
+    describe "chain indeterminate number of iterators" do
```
asterite: Missing extra line above this line.

spec/std/iterator_spec.cr
```diff
@@ -59,6 +59,30 @@ describe Iterator do
...
+      it "chains all together" do
+        iters = [[0], [1], [2, 3], [4, 5, 6]].each.map &.each
+        iter = Iterator(Int32).chain iters
+        7.times { |i| iter.next.should eq i }
+      end
+      it "chains empty" do
```
asterite: Missing extra line above this line.

spec/std/iterator_spec.cr
```diff
@@ -59,6 +59,30 @@ describe Iterator do
...
+        arrs = [] of Array(Int32)
+        iter = Iterator(Int32).chain arrs.map(&.each)
+        iter.next.should be_a Iterator::Stop
+      end
+      it "chains array of empty" do
```
asterite: Missing extra line above this line.

spec/std/iterator_spec.cr
```diff
@@ -59,6 +59,30 @@ describe Iterator do
...
+        iters = [[0], [1], ([] of Int32), [2, 3], ([] of Int32), [4, 5, 6]].each.map &.each
+        iter = Iterator(Int32).chain iters
+        7.times { |i| iter.next.should eq i }
+      end
+      it "rewinds" do
```
asterite: Missing extra line above this line.

asterite:
Does this work when chaining iterators of different types? For example chaining all of `Iterator(Int32)`, `Iterator(String)` and `Iterator(Bool)`?

xqyww123:
Yes, it could and I pushed new commit with the spec.

asterite:
Please change the implementation of `self.chain` to:

```crystal
  def self.chain(iters : Iterator(Iter)) forall Iter
    ChainsAll(Iter, typeof(iters.first.first)).new iters
  end
```

That way you can do:

```crystal
Iterator.chain(iterators)
```

instead of:

```crystal
Iterator(YouHaveToWriteTheTypeHereWhenTheCompilerCanFigureThisOut).chain(iterators)
```

And please add that newline I requested :-)

Thank you!

xqyww123:
Yeah, I realized the problem but didn't find the solution. This way is brilliant and I updated the code.

PR #6572, opened 2018-08-20
Add Time#to_local_in

### LINKED ISSUE #5789: Add a method to transform `Time` instance to a different time zone while preserving wall clock
When transforming a `Time` instance to a different time zone (`Time::Location`), there can be two different outcomes:

* The new `Time` instance refers to the same instance, just expressed in a different time zone. This means the wall clock changes by the difference between time zone offsets.
* The new `Time` instance shows the same wall clock, but in a different time zone. This means the instant changes by the offset difference.

The first variant is implemented as `Time#in(location : Location)`. That's probably the most common usage, to express absolute instances in a local time zone.
The second is not so common but still there are valid use cases. This is useful to reference floating time instances in a local time zone. For example, the next turn of year is at `2019-01-01 00:00:00` but the exact instance depends on the time zone.

It is currently possible to construct such a `Time` instance but it's not easily possible to convert an existing instance. There are workarounds, but there should be a more direct way to do this:

```cr
local_time = Time.new(time.year, time.month, time.day, time.hour, time.minute, time.second, nanosecond: time.nanosecond, location: location)
```

I'm still missing good ideas for a method name, so this is just using a placeholder to show how it should work:

```cr
local_time = time.to_same_wall_clock(location)
```

[Java's Datetime API](`https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html) has the method names `withZoneSameLocal` and `withZoneSameInstant` which are good descriptive but very bulky.

straight-shoota:
Closes #5789

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -118,6 +118,10 @@ require "crystal/system/time"
 # time.to_local # equals time.in(Location.local)
 # ```
 #
+# `#to_local_in` allows changing the time zone while keeping
+# the same local date-time (wall clock) which results in a different instant
+# on the time line.
+#
 # ### Formatting and Parsing Time
 #
 # To make date-time instances exchangeable between different computer systems
@@ -501,6 +505,28 @@ struct Time
     seconds = UNIX_SECONDS + (milliseconds / 1_000)
     nanoseconds = (milliseconds % 1000) * NANOSECONDS_PER_MILLISECOND
     utc(seconds: seconds, nanoseconds: nanoseconds.to_i)
+  end
+
+  # Creates a new `Time` instance with the same local date-time representation
+  # (wall clock) in a different *location*.
+  #
+  # Unlike `#in`, which always preserves the same instant in time, `#to_local_in`
+  # adjusts the instant such that it results in the same local date-time
+  # representation. Both instants are apart from each other by the difference in
+  # zone offsets.
+  #
+  # ```
+  # new_year = Time.utc(2019, 1, 1, 0, 0, 0)
+  # tokyo = new_year.to_local_in(Time::Location.load("Asia/Tokyo"))
+  # new_york = new_year.to_local_in(Time::Location.load("America/New_York"))
+  # tokyo.to_s    # => 2019-01-01 00:00:00.0 +09:00 Asia/Tokyo
+  # new_york.to_s # => 2019-01-01 00:00:00.0 -05:00 America/New_York
+  # ```
+  def to_local_in(location : Location)
+    local_seconds = offset_seconds
+    local_seconds -= Time.zone_offset_at(local_seconds, location)
+
+    Time.new(seconds: local_seconds, nanoseconds: nanosecond, location: location)
   end
 
   def clone : Time
@@ -1041,6 +1067,10 @@ struct Time
   # time_de # => 2018-03-08 22:05:13 +01:00 Europe/Berlin
   # time_ar # => 2018-03-08 18:05:13 -03:00 America/Buenos_Aires
   # ```
+  #
+  # In contrast, `#to_local_in` changes to a different location while
+  # preserving the same wall time, which results in different instants on the
+  # time line.
   def in(location : Location) : Time
     return self if location == self.location
 
--- a/spec/std/time/time_spec.cr
+++ b/spec/std/time/time_spec.cr
@@ -496,6 +496,51 @@ describe Time do
     end
   end
 
+  describe "#to_local_in" do
+    it "keeps wall clock" do
+      location = Time::Location.fixed(3600)
+      location2 = Time::Location.fixed(12345)
+      time1 = Time.now(location)
+      time1.location.should eq(location)
+
+      time2 = time1.to_local_in(location2)
+      time2.location.should eq(location2)
+      time2.year.should eq time1.year
+      time2.month.should eq time1.month
+      time2.day.should eq time1.day
+      time2.hour.should eq time1.hour
+      time2.minute.should eq time1.minute
+      time2.second.should eq time1.second
+      time2.nanosecond.should eq time1.nanosecond
+    end
+
+    it "is the difference of offsets apart" do
+      location = Time::Location.fixed(3600)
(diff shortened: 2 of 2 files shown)


src/time.cr
```diff
@@ -503,6 +503,36 @@ struct Time
     utc(seconds: seconds, nanoseconds: nanoseconds.to_i)
   end
 
+  # Create a new `Time` instance that preserves the same wall clock reading than
+  # *time*  but in a different *location*.
+  #
+  # ```
+  # time = Time.now(Time::Location.new(-3600))
+  # time2 = Time.from_wall_clock(time, Time::Location.new(14400))
+  # time2.to_s("%Y-%m-%d %H:%M:%S") == time1.to_s("%Y-%m-%d %H:%M:%S") # => true
```
r00ster91: Typo: `time1`

asterite:
This looks a bit redundant: `Time.from_wall_clock(time, ...)`. It reminds me of elixir: `String.length(string)`.

Can we somehow provide an instance method instead?

straight-shoota:
It would of course also work as an instance method.
However, I purposely chose a constructor because it creates a new instance from combining the wall readings of the time argument with the given location.

This way it is not likely to be confused with `Time#in(location)`.

I'm definitely open to change it to an instance method. Suggestions are welcome!

asterite:
`time.to_wall_clock_in(location)`

I'm not sure it reads well (grammatically), so if we can't come up with an instance name  the class method is fine.

jhass:
I'm fine with the class method but could also live with something like `same_wall_clock_in`.

src/time.cr
```diff
@@ -503,6 +503,36 @@ struct Time
...
+  # time2 - time # => (time.offset - time2.offset).seconds
+  # ```
+  #
+  # The result of this method is equivalent to going through the calendrical
+  # constructor like this:
+  #
+  # ```
+  # time = Time.now
+  # time2 = Time.new(time.year, time.month, time.day, time.hour, time.minute, time.second, nanosecond: time.nanosecond, location: location)
+  # ```
```
paulcsmith: It might be helpful to talk about how this differs from `Time#in`. Right now it is difficult to tell which one to use and how they are different

src/time.cr
```diff
@@ -503,20 +507,23 @@ struct Time
...
   end
 
-  # Create a new `Time` instance that preserves the same wall clock reading than
-  # *time*  but in a different *location*.
+  # Create a new `Time` instance that preserves the same local date-time
+  # representation (wall clock) as *time* but observed in a different *location*.
   #
   # ```
   # time = Time.now(Time::Location.new(-3600))
   # time2 = Time.from_wall_clock(time, Time::Location.new(14400))
```
jhass: Maybe we can two real timezones like New York and Tokyo and then call the vars `new_york_time`, `tokyo_wall_clock_time` or so? (Admittedly not great names yet). `time`, `time2` gets confusing fast here :/

bcardiff:
If `Time` is seen as an immutable struct, like `record` maybe we can keep the recent idiom of `#copy_with` method that will allow the user to override any specific part of the value.

straight-shoota:
@bcardiff `copy_with` might be a useful addition to `Time`, but I'm not sure it would be a great here, because it doesn't tell there's a difference to `#in`. In fact, I would exspect `#cooy_with` with a location to do the same thing as `#in` because it just copies the instance with the bew location.

straight-shoota:
If there was a local date time representation (floating time), this could be expressed for example like in the Java 8 Date Time API as `time.toLocal().atZone(location)`.
Without the intermediary step, this could become `time.to_local_in(location)` in Crystal.

straight-shoota:
Rebased and renamed `Time.from_wall_clock` to `Time.to_local_in`.

src/time.cr
```diff
@@ -503,6 +507,41 @@ struct Time
...
+  # new_york - tokyo.in(location) # => (tokyo.offset - new_york.offset).seconds
+  # ```
+  #
+  # The result of this method is equivalent to going through the calendrical
+  # constructor like this:
+  #
+  # ```
+  # tokyo = Time.now(Time::Location.load("Asia/Tokyo"))
+  # new_york = Time.new(tokyo.year, tokyo.month, tokyo.day, tokyo.hour, tokyo.minute, tokyo.second, nanosecond: tokyo.nanosecond, location: location)
+  # ```
```
ysbaddaden: This is too abstract. The naming doesn't give much hint at what the method means to achieve, and the documentation doesn't really help until the last example. Maybe using fixed datetimes / string representations, instead of an abstract now and some time operations, and an actual use case would help?

For example, the following would be more helpful, and enough to understand:
> Creates a new `Time` instance with only the location changed. Unlike `#in` that will shift the date time so the new Time instance still refers to the same UTC timestamp, `#to_local_in` keeps the same date and time but for another location.
> 
> For example:
> ```crystal
> halloween = Time.utc(2018, 10, 31, 17, 30, 0)
> tokyo = halloween.to_local_in(Time::Location.load("Asia/Tokyo"))
> new_york = halloween.to_local_in(Time::Location.load("America/New_York"))
> tokyo.to_s    # => 2018-10-31 17:30:00 Asia/Tokyo
> new_york.to_s # => 2018-09-31 17:30:00 America/New_York
> ```

src/time.cr
```diff
@@ -507,33 +507,20 @@ struct Time
...
-  # ```
-  # tokyo = Time.now(Time::Location.load("Asia/Tokyo"))
-  # new_york = tokyo.to_local_in(Time::Location.load("America/New_York"))
-  # new_york - tokyo              # => (tokyo.offset - new_york.offset).seconds
-  # new_york - tokyo.in(location) # => (tokyo.offset - new_york.offset).seconds
-  # ```
-  #
-  # The result of this method is equivalent to going through the calendrical
-  # constructor like this:
+  # Unlike `#in`, which always references the same instant in time, `#to_local_in`
```
RX14: > which always preserves the same instant in time

straight-shoota:
Let me squash this before merging :+1:

PR #6574, opened 2018-08-20
Reorganize Time specs

straight-shoota:
With over 1000 LOC that have evolved a lot over time, `time_spec` has grown to be hard to manage and quite confusing in some parts.

This PR restructures some parts of it to make it easier to maintain and follow the purpose of each test. There have been no changes to the contents of the specs.

* Related specs are grouped together and consistently named after the methods they describe.
* All specs for parsing and formatting (well over 400 LOC and still incomplete) are extracted to `time/format_spec.cr`.
* Unnecessary helper methods are removed.
* Refactored expectations to use plain constructors instead of parsing a string representation.
* Refactored expectations to compare to an entire instance instead of comparing individual fields.

There are a lot of changes only affecting white space, so the diff is best viewed with adding `?w=1` (`-2` CLI flag).


straight-shoota:
Could this get a second review and merge? It's mostly cosmetical and I'm working on a PR that's based upon this.

PR #6214, opened 2018-06-18
Remove :nodoc: from Time::Format

straight-shoota:
Fixes #6206

PR #5123, opened 2017-10-15
Add custom time format implementations

straight-shoota:
The format patterns from Time::Format allow to define simple date formats but they are quite limited in flexibility. Many standardized formats include slight variations which must be taken into account for parsing such formats in a standards-compliant way.

This PR adds custom format implementations for some well-used time formats.
For a few it adds adds short-cut methods to `Time` class (like `Time.parse_iso8601`).

* `Time::Format::ISO_8601_DATETIME`: Parser for standard time format with many variations. It should support most of the standard format including week dates, ordinal dates and optional separators. Uses RFC 3339 as formatter.
* `Time::Format::RFC_3339`: More strict variation of ISO 8601, recommended for exchanging time data in computer systems.
* `Time::Format::RFC_2822`: Format compatible to RFC 822 and RFC 1123 used for example in HTTP protocols.
* `Time::Format::HTTP_DATE`: Proper implementation of `HTTP.parse_date` supporting time formats expressed as RFC 1123/RFC 2822, RFC 850 or asctime which are used by many HTTP protocols. Uses RFC_2822 as formatter.
* `Time::Format::YAML_DATE`: Refactored implementation of parser for YAML `datetime` type.

Surpasses #5084, #4729 

/cc @asterite @RX14

PR #6578, opened 2018-08-20
Fix Time::Format API doc

straight-shoota:
The API doc comment for `Time::Format` was overridden in multiple places (introduced by #5123), so #6214 was not a complete fix.


straight-shoota:
It would be great to add this to 0.26.1

PR #6581, opened 2018-08-21
Refactor Time methods 

straight-shoota:
This PR refactors `Time.year_month_day_year` and `Time.absolute_days` to simplify the code and make it easier to understand.

The changes are mostly just formal, but there are a few semantic changes such as storing the value of an expression in a local variable for re-use.

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -1000,10 +1000,6 @@ struct Time
       raise ArgumentError.new "Invalid month"
     end
 
-    unless 1 <= year <= 9999
-      raise ArgumentError.new "Invalid year"
-    end
-
     days = leap_year?(year) ? DAYS_MONTH_LEAP : DAYS_MONTH
     days[month]
   end
--- a/src/time.cr
+++ b/src/time.cr
@@ -1413,17 +1413,25 @@ struct Time
     end
   {% end %}
 
-  protected def self.absolute_days(year, month, day)
-    days = leap_year?(year) ? DAYS_MONTH_LEAP : DAYS_MONTH
+  # Returns the number of days from `0001-01-01` to the date indicated
+  # by *year*, *month*, *day* in the proleptic Gregorian calendar.
+  #
+  # The valid range for *year* is `1..9999` and for *month* `1..12`. The value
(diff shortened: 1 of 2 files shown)


src/time.cr
```diff
@@ -1216,11 +1220,12 @@ struct Time
...
-    m = 1
-
-    days = DAYS_MONTH
-    totaldays = offset_seconds / SECONDS_PER_DAY
+  # Returns the calendrical representation of this instance's date.
+  #
+  # The return value is a tuple consisting of year (`1..9999`), month (`1..12`),
+  # day (`1..31`) and ordinal day of the year (`1..366`).
+  protected def year_month_day_day_year : {Int32, Int32, Int32, Int32}
+    totaldays = (offset_seconds / SECONDS_PER_DAY).to_i
```
asterite: Maybe name it `total_days`?

straight-shoota: I had considered that, but wasn't sure about it. Will change it :+1:

straight-shoota:
`year_month_day_day_year` is about 25% faster than before.

I actually didn't expect much change in `absolute_days` but for some reason it's about 10% slower. But the measurements vary a lot, it's hard to nail it down, but that seems to be the trend.
When I change the operations back exactly how it was before, it somehow measures about 10% improvement.
This is all really weird and inconclusive. Can someone verify that?

This is my benchmark, it's running through the entire date range, so maybe a bit overkill...
```cr
require "benchmark"

struct Time
  Benchmark.ips do |bm|
    bm.report "new impl" do
      (1..9999).each do |year|
        ordinal = 0
        (1..12).each do |month|
          (1..days_in_month(year, month)).each do |day|
            ordinal += 1
            Time.absolute_days(year, month, day)
          end
        end
      end
    end
    bm.report "old impl" do
      (1..9999).each do |year|
        ordinal = 0
        (1..12).each do |month|
          (1..days_in_month(year, month)).each do |day|
            ordinal += 1
            Time.absolute_days_old(year, month, day)
          end
        end
      end
    end
  end
end
```

RX14:
@straight-shoota 10% is fine

PR #6589, opened 2018-08-22
Display spec runner options in spec command

icyleaf:
I did not find a solution to display spec runner options, but copy it to spec command.

### DIFF
--- a/etc/completion.zsh
+++ b/etc/completion.zsh
@@ -135,6 +135,14 @@ _crystal-run() {
 _crystal-spec() {
   _arguments \
     $programfile \
+    '(-e --example)'{-e=,--example=}'[Run examples whose full nested names include STRING]:' \
+    '(-l --line)'{-l=,--line=}'[Run examples whose line matches LINE]:' \
+    '(-p --profile)'{-p,--profile}'[Print the 10 slowest specs]' \
+    '(--fail-fast)--fail-fast[Abort the run on first failure]' \
+    '(--location)--location[Run example at line "line" in file "file", multiple allowed]' \
+    '(--junit_output)--junit_output[Generate JUnit XML output]' \
+    '(-v --verbose)'{-v,--verbose}'[Verbose output]' \
+    '(--tap)--tap[Generate TAP output (Test Anything Protocol)]' \
     $help_args \
     $no_color_args \
     $exec_args \
--- a/src/compiler/crystal/command/spec.cr
+++ b/src/compiler/crystal/command/spec.cr
@@ -12,7 +12,20 @@ class Crystal::Command
   private def spec
     compiler = Compiler.new
     OptionParser.parse(options) do |opts|
-      opts.banner = "Usage: crystal spec [options] [files]\n\nOptions:"
+      opts.banner = <<-'BANNER'
+      Usage: crystal spec [options] [files]
+
+      Options:
+          -e, --example STRING             Run examples whose full nested names include STRING
+          -l, --line LINE                  Run examples whose line matches LINE
+          -p, --profile                    Print the 10 slowest specs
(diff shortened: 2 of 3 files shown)


RX14:
This duplication is pretty terrible and will definitely be forgotten about...

asterite:
The problem with this is that `crystal spec` just translates to `crystal spec/**`, which compiles and runs all specs. And with that there's no specific spec framework chosen. The spec framework is chosen with a `require` someplace in all that directory, usually by requiring a `spec_helper`. So the options added in this PR are tied to the `spec` in the standard library. But if you use `minitest` or `spec2` the options might be different.

I don't know how to solve this issue, though...

RX14:
@asterite I think assuming the user is using stdlib spec is fine. It's not great, but it's fine.

asterite:
But then, you can run `crystal spec -h`, I don't know how that output will look mixed with the changes in this PR...

RX14:
@asterite `crystal spec -h` doesn't run any crystal code, it just displays the help for `crystal spec` itself (i.e. the compiler options) then exits. That's what this PR attempts to fix.

asterite:
No, I meant `crystal spec -h`. That already shows a bunch of options, and with this change to `banner` I don't know how that will look (I didn't try it).

straight-shoota:
Oh, you just mean the visual presentation created by this PR?

What about displaying these options in a separate section, labeling them as options for the default spec framework?

```
Usage: crystal spec [options] [files]

Options:
    -d, --debug                      Add full symbolic debug info
    --no-debug                       Skip any symbolic debug info
    -D FLAG, --define FLAG           Define a compile-time flag
    --error-trace                    Show full error trace
    --release                        Compile in release mode
    -s, --stats                      Enable statistics output
    -p, --progress                   Enable progress output
    -t, --time                       Enable execution time output
    -h, --help                       Show this message
    --no-color                       Disable colored output

Options for spec framework:
    -e , --example STRING            run examples whose full nested names include STRING
    -l , --line LINE                 run examples whose line matches LINE
    -p, --profile                    Print the 10 slowest specs
    --fail-fast                      abort the run on first failure
    --location file:line             run example at line 'line' in file 'file', multiple allowed
    --junit_output OUTPUT_DIR        generate JUnit XML output
    --help                           show this help
    -v, --verbose                    verbose output
    --tap                            Generate TAP output (Test Anything Protocol)
    --no-color                       Disable colored output
```

straight-shoota:
> This duplication is pretty terrible and will definitely be forgotten about...

Adding references at both places in the code should help.

bcardiff:
Regarding how to deal with different spec frameworks I would propose that `crystal spec -h` could: 

1. require `spec/**`
2. define the main to can an assumed `spec_formatter_options` existing function (or similar) that should be defined by the included spec framework in `spec/**`.
3. compile and run to get the output to append.

Compiling all the specs could be a bit too much time, but `crystal spec -h` could show the top part of the banner fast and delay the output of the bottom part.

The same could be used for autocompletion but it will require some caching in order to make it useful.

---

Another path to go is to have fixed options, but then it would be good to allow specs frameworks to receive this parameters without parsing them again. So there would be an API to host spec frameworks in `crystal spec` runner.

RX14:
@bcardiff I think that's overengineering, if we really wanted to do that we'd just run and execute the specs with `-h` as an argument and append that to the help output.

I'd much rather just make it possible to require `spec/options` in the compiler and the compiler just prints out a standard spec suite options. No, it won't make sense for non-standard spec runners but it's better than what we have now and it's simple.

icyleaf:
I forgot the third party spec framework, close this PR for now, but if most of core developer will agree with @straight-shoota suggest, i can make it up.

Issue #834, opened 2015-06-20
HTTP::Server: performance drop + listening socket gets closed


ysbaddaden:
While testing artanis for #788 (fixed) with wrk, I noticed 2 problems:

- a huge performance drop: from 70k/s requests to merely 23k/s.
- the listening TCP socket being closed :fearful: 

The app won't stop or fail or raise, but the listening socket gets closed. wrk can't connect anymore and lsof won't list the socket after the first wrk run. I can reproduce this every time on my ubuntu trusty laptop with current master. I cherry-picked the patch for #788 over 0.7.3 but everything is fine then: memory leak is fixed, 70k/s and the listening socket is kept opened.

I tried the HTTP::Server example against the current master, and got a bunch of `Unhandled exception: Error writing file: Broken pipe` when wrk finishes, with the same behavior: slow requests (37k/s) and the listening socket being closed.

waj:
Can you try with a889171? It should fix the closed socket. The performance dropdown is due to the socket being not buffered by default after the IO refactor. It should be fixed for 0.7.4 anyway.

waj:
I reopen until we do more testing

waj:
@ysbaddaden can you try again? a9b00cc should fix the closed socket issue and 82e63bc should improve the performance

ysbaddaden:
I confirm that both are fixed. The socket is kept open, and I even get better benchmarks (80k requests/s)! And memory is kept low & reclaimed. HTTP::Server rocks :-)

waj:
Thanks @ysbaddaden! And we still have more improvements coming, yay! ;-)

Codcore:
@waj nice work, thanks!

PR #6304, opened 2018-06-30
Let IO#sync only affect writes, introduce `IO#read_buffering?` and use it in urandom

asterite:
Fixes #6303

Being able to disable read buffering for `IO::Buffered` is good. However, having `sync` affect both reads and writes makes it easier to enable just buffering for one direction: for sockets it's better to not have write buffering by default, so whatever you write goes directly to the socket (you can always set `sync = false` and manually flush, which is what we do in a few places), but reads should always be buffered for performance reasons.

With this PR we could release 0.25.2 as a bugfix release for this performance regression.

Then for 0.26.0 we _could_ rename `sync` to `write_buffering`, or just leave it as `sync`. Maybe `write_buffering` is more explicit, even though it's a breaking change. But then, disabling read buffering is a very uncommon use case (urandom... what else?), so maybe having `sync` and `read_buffering` as separate, asymmetric properties isn't that bad.

(I know I could have used `getter` or `setter` for this, but given that `sync` doesn't use them I decided to follow that style. If someone wants to do a follow up PR to "improve" that (it's not really an improvement, because macros are slower to compile and it's just one/two line differences), then please go ahead).

Issue #3093, opened 2016-08-04
Compiler should be smarter about immutable closured variables


asterite:
I couldn't find a way to reduce the title's size, but basically:

``` cr
def capture(&block)
  block
end

x = rand < 0.5 ? 1 : nil
if x
  capture do
    x + 1 # undefined method '+' for Nil (compile-time type is (Int32 | Nil))
  end
end
```

Even though we check that `x` is not `nil`, and it's never modified again, the compiler complains because right now any closured variable has all types all the time (they are not affected by type flow). This should be improved by noticing that `x` doesn't change. It's not trivial because if inside a `while` or another `block` it gets tricky, but should be doable.

asterite:
This should also work:

```crystal
x = 1
x = "hello"

capture do
  x # here x is known to always be a String, never an Int32
end
```

PR #6590, opened 2018-08-22
Fix HTTP::Server crash with self-signed certificate

### LINKED ISSUE #6577: HTTP::Server crash with self-signed certificate
It is possible in the latest version to crash without rescuable warnings an HTTP::Server that uses openssl

Example code:

```crystal
require "openssl"
require "http/server"

end

context = OpenSSL::SSL::Context::Server.new
context.certificate_chain = "cert.pem"
context.private_key = "key.pem"
server.bind_ssl "127.0.0.1", 443, context

server.listen
```

Using a `curl https://127.0.0.1` will crash the application.

When using a self-signed certificate, or when the connecting client will generate an OpenSSL "SSL certificate problem: unable to get local issuer certificate" the ssl server process will exit.

This is problematic as basically every exposed crystal server that uses SSL can be crashed using a simple ssl request that has an empty cert store.

bcardiff:
Fixes #6577

* Refactor how `socket.accept?` is called. `handle_client` is spawned after the `io` is returned.
  * It is assumed that `socket.accept?` wont raise and a STDOUT message is logged if that does not hold.
* `OpenSSL::SSL::Server#accept?` is changed to not raise
* This was optional but `OpenSSL::SSL::Socket::Server` will now call `LibSSL.ssl_shutdown` if the accept failed.

While reviewing the current state of `HTTP::Server` and recent PRs I have some doubts:

1. #6304 split the behaviour of `#sync`. I am not sure if that change should have changed https://github.com/crystal-lang/crystal/blob/0.26.0/src/http/server.cr#L379 to also set `read_buffering = false`. I am not sure if forcing `sync = false` was for reading or writing. That `sync = false` was first introduced to fix https://github.com/crystal-lang/crystal/issues/834 I am not sure if the performance improvement came from reads or writes changes. From #6304 description it seems that just unbuffered writes are wanted, so there is nothing to change.

2. The `sync = false` is, since 0.26.0 applied _after_ the handshake in the SSL, but in 0.25 it was applied _before_. I am not sure it that could be an issue or not. If so the following snippet might be required (assuming also that `read_buffering = false` is wanted)

```patch
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -50,6 +50,11 @@ abstract class OpenSSL::SSL::Socket < IO
     def initialize(io, context : Context::Server = Context::Server.new, sync_close : Bool = false)
       super(io, context, sync_close)
 
+      if io.is_a?(IO::Buffered)
+        io.sync = false
+        io.read_buffering = false
+      end
+
       ret = LibSSL.ssl_accept(@ssl)
       unless ret == 1
         self.close
```

3. If (2) is not needed, there is still an issue. The `OpenSSL::SSL::Socket` implementation, it is a `IO::Buffered` that wraps and `IO`, and does not forward the `sync`/`read_buffering` values to the wrapped `io`. Since the default tcp socket is buffered, the ssl socket might be unbuffered due to `HTTP::Server#handle_client`, but the wrapped tcp socket is still buffered. It also make little sense to have a `IO::Buffered` wrapped by another `IO::Buffered`, so I think (2) make sense since it turns off the buffering of the inner `io`.

So, I think the only think pending should be to apply (2) as is in this PR. But I wanted to double check my assumptions of (1) mainly and, of course, the rest of the PR.

### DIFF
--- a/spec/std/http/server/server_spec.cr
+++ b/spec/std/http/server/server_spec.cr
@@ -59,6 +59,11 @@ private def unused_port
   end
 end
 
+private class SilentErrorHTTPServer < HTTP::Server
+  private def handle_exception(e)
+  end
+end
+
 module HTTP
   class Server
     describe Response do
@@ -504,6 +509,39 @@ module HTTP
         end
       end
     {% end %}
+
+    it "handles exception during SSL handshake (#6577)" do
+      server = SilentErrorHTTPServer.new do |context|
+        context.response.print "ok"
+        context.response.close
+      end
+
+      server_context, client_context = ssl_context_pair
+      address = server.bind_tls "localhost", server_context
+
+      server_done = false
+      spawn do
+        server.listen
+        server_done = true
+      end
+
+      3.times do
+        # Perform multiple wrong calls together and check
+        # that the server is still able to respond.
+        3.times do
+          empty_context = OpenSSL::SSL::Context::Client.new
+          socket = TCPSocket.new(address.address, address.port)
+          expect_raises(OpenSSL::SSL::Error) do
+            OpenSSL::SSL::Socket::Client.new(socket, empty_context)
+          end
+        end
+        HTTP::Client.get("https://#{address}/", tls: client_context).body.should eq "ok"
+      end
+
+      Fiber.yield
+
+      server_done.should be_false
+    end
   end
 
   describe HTTP::Server::RequestProcessor do
--- a/src/http/server.cr
+++ b/src/http/server.cr
@@ -355,7 +355,18 @@ class HTTP::Server
     @sockets.each do |socket|
       spawn do
         until closed?
-          spawn handle_client(socket.accept? || break)
+          io = begin
+            socket.accept?
+          rescue e
+            handle_exception(e)
+            nil
+          end
+
+          if io
+            # a non nillable version of the closured io
+            _io = io
+            spawn handle_client(_io)
+          end
         end
       ensure
         done.send nil
@@ -389,6 +400,11 @@ class HTTP::Server
     end
 
     @processor.process(io, io)
+  end
+
+  private def handle_exception(e : Exception)
+    e.inspect_with_backtrace STDERR
+    STDERR.flush
   end
 
   # Builds all handlers as the middleware for `HTTP::Server`.
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -83,6 +83,14 @@ abstract class OpenSSL::SSL::Socket < IO
     unless @ssl
       raise OpenSSL::Error.new("SSL_new")
     end
+
+    # Since OpenSSL::SSL::Socket is buffered it makes no
+    # sense to wrap a IO::Buffered with buffering activated.
+    if io.is_a?(IO::Buffered)
+      io.sync = false
+      io.read_buffering = false
+    end
+
     @bio = BIO.new(io)
     LibSSL.ssl_set_bio(@ssl, @bio, @bio)
   end


src/http/server.cr
```diff
@@ -355,7 +355,19 @@ class HTTP::Server
...
+            socket.accept?
+          rescue e
+            ::puts "#{socket.class}#accept? raised but it shouldn't"
+            ::puts e
+            nil
+          end
+
+          if io
+            # a non nillable version of io is passed to spawn
+            _io = io
```
Sija: Shoudn't `io` here be already non-nillable? IIRC `if` branch filters out `Nil`.

bcardiff: You recall correctly but using `spawn` affects that rule.

asterite: `io` is closured, so it always retains all types assigned to it. Basically https://github.com/crystal-lang/crystal/issues/3093

src/http/server.cr
```diff
@@ -355,7 +355,19 @@ class HTTP::Server
     @sockets.each do |socket|
       spawn do
         until closed?
-          spawn handle_client(socket.accept? || break)
+          io = begin
+            socket.accept?
+          rescue e
+            ::puts "#{socket.class}#accept? raised but it shouldn't"
```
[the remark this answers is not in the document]
bcardiff: I've been puts debugging inside sockets 😅

src/http/server.cr
bcardiff: The ` || break` before did the trick because the `nil` was transformed to a `NoReturn`.
That stop working in a version were I used `spawn` with block, but `|| break` could return probably since there is no block involved in `spawn`, but it was a bit cryptic.

bcardiff:
@straight-shoota when a better logging will come that will change and the private method will not be needed.

Do you have any input regarding (1)/(2)/(3)?

straight-shoota:
Better logging is not here yet. And logging won't allow to treat the actual exception.

Additionally, a private method would help adding a spec for this. Something like:

```cr
private class CustomHTTPServer < HTTP::Server
  getter last_exception : Exception?
  private def handle_exception(@last_exception)
  end
end

it "handles exception during SSL handshake" do
  server = CustomHTTPServer.new {}
  server_context, _ = ssl_context_pair
  address = server.bind_tls "localhost", server_context

  server_done = false
  spawn do
    server.listen
    server_done = true
  end

  context = OpenSSL::SSL::Context::Client.new
  socket = TCPSocket.new(address.address, address.port)
  expect_raises(OpenSSL::SSL::Error, "certificate verify failed") do
    OpenSSL::SSL::Socket::Client.new(socket, context)
  end
  Fiber.yield

  server_done.should be_false
  server.last_exception.should_not be_nil
  server.last_exception.not_nil!.message.should contain("alert unknown ca")
end
```

straight-shoota:
Regarding sync behaviour, I'm not sure about all the details... (side note: I think this entire API with `sync` and `read_buffering` needs to be overhauled to be more clearly understandable).

I suppose adding (2) is fine, but not sure if it is required.

However, I'd rather add it in `SSL::Server#accept` and `#accept?` instead of `SSL::Socket::Server#initialize`. The server is the place where an incoming connection needs to be setup. The SSL socket should not have to worry about that.

src/http/server.cr
```diff
@@ -355,7 +355,19 @@ class HTTP::Server
...
+            ::puts e
```
RX14: `e.inspect_with_backtrace(STDOUT)`.

Sija: `STDOUT` -> `STDERR`

RX14: then the other line needs to be `STDERR.puts`

bcardiff:
I applied the suggestions and turn off the buffering in wrapped `IO::Buffered` `IO`s for all `OpenSSL::SSL::Socket`.

It's ready for another round of review.

src/openssl/ssl/socket.cr
```diff
@@ -83,6 +83,14 @@ abstract class OpenSSL::SSL::Socket < IO
     unless @ssl
       raise OpenSSL::Error.new("SSL_new")
     end
+
+    # Since OpenSSL::SSL::Socket is buffered there make no
+    # sense to wrap a IO::Buffered with buffereing activated.
```
r00ster91: Typo: buffereing

spec/std/http/server/server_spec.cr
```diff
@@ -504,6 +511,29 @@ module HTTP
...
+
+      server_done = false
+      spawn do
+        server.listen
+        server_done = true
+      end
+
+      context = OpenSSL::SSL::Context::Client.new
+      socket = TCPSocket.new(address.address, address.port)
+      expect_raises(OpenSSL::SSL::Error) do
```
straight-shoota: Is there a specific reason why you removed the specific message matcher `certificate verify failed`?

src/http/server.cr
```diff
@@ -355,7 +355,18 @@ class HTTP::Server
     @sockets.each do |socket|
       spawn do
         until closed?
-          spawn handle_client(socket.accept? || break)
+          io = begin
+            socket.accept?
+          rescue e
+            self.handle_exception(e)
```
straight-shoota: Why `self.`?

src/openssl/ssl/socket.cr
```diff
@@ -83,6 +83,14 @@ abstract class OpenSSL::SSL::Socket < IO
     unless @ssl
       raise OpenSSL::Error.new("SSL_new")
     end
+
+    # Since OpenSSL::SSL::Socket is buffered there make no
```
straight-shoota: `it makes no sense` or `there is no sense`

spec/std/http/server/server_spec.cr
bcardiff: because the spec was failing in linux, but it turns out that the exception is not even raised. Need to double check in a vm...

straight-shoota: The spec I posted worked on my ubuntu 16.04 on WSL.

Perhaps try adding a few more `Fiber.yield` to make sure the error handling code is *really* executed.

straight-shoota:
It seems the spec as in 5008b48 randomly fails in about 30% of runs due to no apparent reason.

spec/std/http/server/server_spec.cr
```diff
@@ -504,6 +509,37 @@ module HTTP
...
+      address = server.bind_tls "localhost", server_context
+
+      server_done = false
+      spawn do
+        server.listen
+        server_done = true
+      end
+
+      3.times do
+        3.times do
```
j8r: Why `3.times`'ing `3.times`, `HTTP::Client.get` can't be included in a `9.times`?

bcardiff:
The spec was failing because in the openssl 1.0.2 installed in linux the client triggering the socket in the server. I change the spec to test the overall behavior of not taking down the server and keep able to response to requests.

I also rollbacked the change of calling SSL_shutdown if init failed because it was wrong. Ref:  https://github.com/openssl/openssl/issues/710

src/http/server.cr
```diff
@@ -355,7 +355,18 @@ class HTTP::Server
...
+            nil
+          end
+
+          if io
+            # a non nillable version of the closured io
```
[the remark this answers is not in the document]
bcardiff: nillable is fine, it's not a proper word but the concept is clear.

spec/std/http/server/server_spec.cr
bcardiff: Because I wanted multiple wrong calls together.

bcardiff: done

ysbaddaden:
I believe OpenSSL has its own buffer for SSL sockets, so we should avoid to have one in Crystal neither read/writes on the underlying socket (set before handshake), and only sync=true on the SSL socket wrapper.

bcardiff:
Good catch @ysbaddaden. But that is not a blocker issue and can be deferred. The `BIO_set_write_buf_size` is not even binded and the ssl were made buffered at 32e4dd51b6699986b8cdeba0cec487171e18846a to fix an issue it seems.

Issue #6522, opened 2018-08-11
Time + Time::MonthSpan inaccurate around DST change


HCLarsen:
When adding or subtracting a Span or MonthSpan to a Time object that isn't UTC, the result is off by an hour. It appears that it's adding a specific number of seconds instead of days, weeks, months or years, as you would expect.

```
time = Time.utc(1997, 10, 21, 9, 0, 0)
time + 1.weeks  #=> 1997-10-28 09:00:00.0 UTC

time = Time.new(1997, 10, 21, 9, 0, 0)
time + 1.weeks  #=> 1997-10-28 08:00:00.0 -05:00 Local

time = Time.new(1997, 10, 25, 9, 0, 0, location: Time::Location.load("America/Toronto"))
time + 1.days  #=> 1997-10-26 08:00:00.0 -05:00 Local
```

asterite:
I won't comment on whether this is good or not (it does look good to me, though), but yes, time math goes through seconds, always.

asterite:
Actually, for months it doesn't go through seconds. So:

```crystal
time = Time.new(1997, 10, 25, 9, 0, 0, location: Time::Location.load("America/Toronto"))
p time           # => 1997-10-25 09:00:00.0 -04:00 America/Toronto
p time + 1.month # => 1997-11-25 09:00:00.0 -05:00 America/Toronto
```

So yes, one of the behaviors should _maybe_ be fixed, though I can imagine it being really hard to implement, specially when right now you can combine units (except months).

HCLarsen:
Think about implementation though. When you're thinking of 1 week from today, you expect that to be the same hour, not an hour earlier because you set your clocks forward in the meantime.

HCLarsen:
The workaround I implemented in my project is only 7 lines long. I used a comparison of DST on the receiver and the argument.

asterite:
I don't know. Let's say I say "in one hour" and dst starts having effect. Then your 7 lines diff will change the hour again. Then in real life, an hour wouldn't have passed, but maybe zero or two.

Maybe with days and weeks it's different, because they are usually tied to a date.

How does Ruby's activrsupport behave regarding this?

HCLarsen:
No, my workaround doesn't change the hour if the span is 1.hour.

I've been trying to figure out how Ruby's Active Support does this, but I'm not sure yet.

HCLarsen:
Active Support does exactly what one would expect. A week after 9am on October 21st is 9am on October 28th.

[...]

Issue #4556, opened 2017-06-12
Having a Great Time


straight-shoota:
Time in stdlib is found to be lacking in different ways and there have been some mentions of possible improvements

* nanosecond resolution #4552 #3827
* JSON mapping for timespan #4446
* represent time with zone offset #2490, #3307 #2481#3827
* higher granularity for time span (e.g. properly calculate age in years)

I've done some research on time and date solutions in other languages and as far as I can tell, the [Joda Time](http://www.joda.org/joda-time/) based [Date-Time API in Java 8](https://docs.oracle.com/javase/tutorial/datetime/iso/overview.html) looks like a good source of inspiration.

A few key concepts (as far as I understand it):
* differentiate between `Time` and `Date`, with `DateTime` being a composite. Internally, the values are represented as individual fields (`hour, minute, second : Int8, nano : Int32`, `year : Int32, month, day : Int16`)
* basic blocks are `Local(Date/Time)` (no timezone) and `LocalDateTime` is wrapped in `ZonedDateTime` to add a timezone.
* Time spans can be represented as `Duration` (accurate, time based) or `Period` (conceptual, date based).

I'm not familiar enough with Crystal internals and don't know if and how these ideas could be useful in Crystal, but this could be a good place to discuss plans for improvements of Crystal's time API.

Other notable resources (other items might be appended):
* [chrono](https://github.com/chronotope/chrono) (Rust)
* [Lib-datetime Draft](https://github.com/rust-lang/rust-wiki-backup/blob/master/Lib-datetime.md) with further references
* ISO 8601: [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601), [PDF](http://xml.coverpages.org/ISO-FDIS-8601.pdf)
* Elixir [timex](https://hexdocs.pm/timex/getting-started.html)
* [Proposal: Monotonic Elapsed Time Measurements in Go](https://golang.org/design/12914-monotonic)
* W3C: [Working with Time Zones](https://www.w3.org/TR/timezone/)
* [unix4lyfe.org/time](https://unix4lyfe.org/time/) ([Hacker News thread](https://news.ycombinator.com/item?id=2725015))

akzhan:
Theoretically Time spans should have xxseconds and months. It cover all known (for me) cases. 

See http://www.w3resource.com/sql/data-type.php#INTERVAL for details.

straight-shoota:
It won't cover adding a conceptual day over switching of DST.

[...]

PR #6598, opened 2018-08-24
Add Time#shift for changing a time instance by calendar units

straight-shoota:
Based on #6574, only the last two commits are specific to this PR.

The main part of this PR renames `Time#add_months` to `Time#shift` and adds arguments for additional calendrical units.
This is copied from [Timex.shift/2](https://hexdocs.pm/timex/Timex.html#shift/2). An also fitting alternative would be `#advance` (from Rails' [Time.advance](https://api.rubyonrails.org/classes/Time.html#method-i-advance)).
This method can be used for adding arbitrary amounts of (conceptual) time units and forms a basis for an upcoming `Time::DateSpan` type (see #6522).

This is partly discussed in #6522 and #4556

Regarding performance, the new `#shift` is about twice as fast as `#add_months` for shifting by months/years. The comparison is not really fair though, because `#add_months` called `year`, `month` and `day` individually which means the rather expensive calculations in `year_month_day_day_year` are executed three times instead of once. With this simple optimization applied, `#shift` is still 50% faster than `#add_months` (probably because it does't go through the calendrical constructor).

`Time#add_span` is also renamed to `Time#shift` which is a more fitting name as well and means there are two implementations of `Time#shift`, one based on elapsed time in seconds, the other on calendrical units.
So now there are two ways for adding a time span to an instance of `Time`:
`time + 2.hours` advances by the duration of two hours on the instant time-line (elapsed time) and `time.shift hours: 2` advances by two hours in local time (wall clock).

### DIFF
--- a/spec/std/time/time_spec.cr
+++ b/spec/std/time/time_spec.cr
@@ -270,6 +270,33 @@ describe Time do
       time.shift(0, 0).should eq time
     end
 
+    describe "irregular calendrical unit ratios" do
+      it "shifts by a week if one day is left out" do
+        # The week from 2011-12-25 to 2012-01-01 for example lasted only 6 days in Samoa,
+        # because it skipped 2011-12-28 due to changing time zone from -11:00 to +13:00.
+        with_zoneinfo do
+          samoa = Time::Location.load("Pacific/Apia")
+          start = Time.new(2011, 12, 25, 0, 0, 0, location: samoa)
+
+          plus_one_week = start.shift days: 7
+          plus_one_week.should eq start + 6.days
+
+          plus_one_year = start.shift years: 1
+          plus_one_year.should eq start + 365.days # 2012 is a leap year so it should've been 366 days, but 2011-12-28 was skipped
+        end
+      end
+
+      it "shifts by conceptual hour even if elapsed time is less" do
+        # Venezuela switched from -4:30 to -4:00 on 2016-05-01, the hour between 2:00 and 3:00 lasted only 30 minutes
+        with_zoneinfo do
+          venezuela = Time::Location.load("America/Caracas")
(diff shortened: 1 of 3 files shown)


bew:
> `time + 2.hours` advances by the duration of two hours on the instant time-line (elapsed time) and `time.shift hours: 2` advances by two hours in local time (wall clock)

I don't understand the difference between the 2 methods (if there is one), could you give an example showing the difference? Ty

straight-shoota:
If the wall clock advances by two hours, it shows a time two hours later, but that doesn't mean the elapsed time between both instants also equals to two hours. For example when there is a date switch in between, they might be apart on the instant time line by the duration of only one hour or even three hours.

robacarp:
@straight-shoota first off, let me say thank you for tackling so much of the nonsense of time and date management. This is one of the things that can create a quality legacy in a language/stdlib and it can be so helpful. 

I believe you're trying to capture the essence of what it means to modify a timestamp by measured time, instead of simply incrementing the digits timestamp as if it was simple math, which I believe is what `time + delta` does. 
 
If that is the case, the word `shift` is opaque to me. I think I prefer `advance`, but in the case where a program is "advancing backwards", I'm less certain. 

Either way, thanks for this.

luislavena:
Had a joke/suggestion to rename it to `Time#travel` instead 🤣

straight-shoota:
@robacarp Yes, `Time#advance` is a potential alternative. I think I prefer `#shift` because it fit's with either direction.

straight-shoota:
@luislavena I think the time travel metaphor fits better for manipulating the clock (such as [timecop.cr](https://github.com/TobiasGSmollett/timecop.cr)).

straight-shoota:
ping @bcardiff

straight-shoota:
Rebased and ready to ship.

PR #6602, opened 2018-08-25
Use Menlo instead of Monaco in generated documentation

slice:
## Details/Motivation

Monaco is specified second to last in the font stack, but is never used because macOS has the Courier New and Courier fonts, which always gets used instead and looks pretty bad.

This patch adds Menlo to the front of the font stack and moves Monaco to be after it.

This shouldn't affect font displays on platforms other than macOS.

## Comparison

Before:

image

After:

image

### DIFF
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -235,7 +235,7 @@ body {
 }
 
 .entry-const {
-  font-family: Consolas, 'Courier New', Courier, Monaco, monospace;
+  font-family: Menlo, Monaco, Consolas, 'Courier New', Courier, monospace;
 }
 
 .entry-const code {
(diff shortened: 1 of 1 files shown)


straight-shoota:
What's the reason for using Menlo?

j8r:
People finds the `Menlo` font more readable than `Courier` fonts @straight-shoota . Not a big deal, but quite agree.

kirbyfan64:
I mean, it really doesn't hurt to fix this, since it seems there's at least a few people who like this.

slice:
While this isn't any kind of long term solution, I figured it would be a nice short term improvement to the look of the docs. 😄

bcardiff:
I love fonts. There is a long term design pending story. Actually we do have a nice design but never implemented. The community built the current one, but we do expect to match the general style of the homepage some day.

That been font ordering that this PR propose make sense.
Thanks @slice !

PR #6606, opened 2018-08-25
[WIP] Colorize PrettyPrint output

makenowjust:
This PR adds `PrettyPrint#color` method to use colorized output for `pretty_print`.

Example:

```crystal
class Foo
  def initialize(@foo : Int32)
  end
end

struct Bar
  def initialize(@bar : Int32)
  end
end

pp({
  string: "hello world",
  symbol: :foo,
  bool:   true,
  number: 123i64,
  nil:    nil,
  regexp: /foo/,
  time:   Time.now,
  class:  Foo.new(123),
  struct: Bar.new(456),
})
```

TODO:

  - [ ] Add tests for `PrettyPrint#color`
  - [ ] Add documents for `PrettyPrint#.color_enabled` and `#.palette`


asterite:
Even though this is really cool, I think it belongs in a shard.

Also, colors might work for one terminal setting and not for another, making it problematic for us (maintainers) to deal with this problem. Moving the problem to a shard, where the colors can somehow be configured, seems a better way. I'd also name it something else... maybe `ap` for "awesome print"? :-)

j8r:
Its fine, we have already colors enabled by default for compilation errors and specs.
This feature should support the existing `--no-color` option.

j8r:
Why it wouldn't work in Windows? It now supports [VT100 escape sequences](https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences).
This escape sequences are really wide spread, nearly all OSes have terminals that support them.

asterite:
Cool!

Still, there's no reason to `require "colorize"` in every program.

straight-shoota:
I think it could be useful to add the context information for colors to the pretty print API but by default it is not used. When the program includes `pretty_print/color` (for example), this could hook into `PrettyPrint` and use the context information to actually add colors.
This could either be a shard or in stdlib, but doesn't need to be in core lib (and shouldn't).

RX14:
I don't think this should be a shard, but it should be an additional `require "colorize/pretty_print"`.

MakeNowJust:
I think this should be enabled by default because it is cool, excellent and useful for debugging surely. If it needs to `require "colorize/pretty_print"`, in future the book like "Effective Crystal" says "You should add `require "colorize/pretty_print"` in every program entry-point files for colorized `pp`. It helps your development." Is it the future you hope?

RX14:
If that's the cost of keeping colorize out of the corelib, then yes that's fine.

PR #6609, opened 2018-08-26
Fix docs of Array#sort and Array#sort!

### LINKED ISSUE #6608: Array#sort requires < and <= methods, not <=>
According to the docs for [Array#sort](https://crystal-lang.org/api/0.26.0/Array.html#sort%3AArray%28T%29-instance-method), it:

> Returns a new array with all elements sorted based on the return value of their comparison method #<=>

However, `Array#sort` only uses the `#<` and `#<=` methods; it does not use the `#<=>` method.

## With <=>

```crystal
class Foo
  property :number

  def initialize(@number : Int32)
  end

  def <=>(other : Foo)
    self.number <=> other.number
  end
end

arr = [Foo.new(10), Foo.new(20), Foo.new(5), Foo.new(15)]
arr.sort!
pp arr.map(&.number)
```

Compiling yields:

>in /usr/share/crystal/src/array.cr:1872: undefined method '<' for Foo
>
>      c -= 1 if a[c] < a[c - 1]
>                     ^

## With < and <=

```crystal
class Foo
  property :number

  def initialize(@number : Int32)
  end
  
  def <(other : Foo) 
    self.number < other.number
  end

  def <=(other : Foo)
    self.number <= other.number
  end
end

arr = [Foo.new(10), Foo.new(20), Foo.new(5), Foo.new(15)]
arr.sort!
pp arr.map(&.number)
```

Running this code yields:

> [5, 10, 15, 20]

## System Information

Operating System
> Ubuntu 18.04.1 LTS

crystal -v
> Crystal 0.26.0 [eeb53c506] (2018-08-09)
> 
> LLVM: 4.0.0
> Default target: x86_64-unknown-linux-gnu

r00ster91:
Fixes #6608 and adds missing \` around #<=> in the docs of [`Array#sort_by`](https://crystal-lang.org/api/0.26.0/Array.html#sort_by%28%26block%3AT-%3E_%29%3AArray%28T%29-instance-method) and [`Array#sort_by!`](https://crystal-lang.org/api/0.26.0/Array.html#sort_by%21%28%26block%3AT-%3E_%29%3AArray%28T%29-instance-method).

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -1536,7 +1536,7 @@ class Array(T)
   end
 
   # Returns a new array with all elements sorted based on the return value of
-  # their comparison method `#<=>`
+  # their comparison methods `#<` and `#<=`.
   #
   # ```
   # a = [3, 1, 2]
@@ -1566,7 +1566,7 @@ class Array(T)
   end
 
   # Modifies `self` by sorting all elements based on the return value of their
-  # comparison method `#<=>`
+  # comparison methods `#<` and `#<=`.
   #
   # ```
   # a = [3, 1, 2]
@@ -1597,7 +1597,7 @@ class Array(T)
   end
 
   # Returns a new array with all elements sorted. The given block is called for
-  # each element, then the comparison method #<=> is called on the object
+  # each element, then the comparison method `#<=>` is called on the object
   # returned from the block to determine sort order.
   #
   # ```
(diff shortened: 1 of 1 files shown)


asterite:
This is not a good fix. The fix is to use `<=>` in array sort.

Additionally, we should support returning `nil` in the spaceship operator to error on sort. For example float nan isn't comparable with other numbers (ruby raises in this case).

Issue #4650, opened 2017-07-01
HTTP::Client hangs


nobilik:
I don't have big experience as issue writer, but I'll try

When I use HTTP::Client with concurrency >= 5 (for sure, maybe less) my app completely hangs with no output. From time to time it can pass but it happens seldom. 

`TOP` shows:
```
PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND     
13360 user   20   0  357836  22860   7580 R 100,0  0,1   1:26.57 insta_app 
```
There is no any output with `strace` also.

Crystal 0.23.0 [3c3d3e2] (2017-06-28) LLVM 3.8.1
Ubuntu 16.04.2 LTS

```
require "http/client"
API_URI = URI.parse("https://www.instagram.com")
loop do
   get_comments
   sleep 10
end
   def get_comments
    concurrency = 30
    request_qtt = 3000
    channel = Channel(Nil).new(concurrency)
    concurrency.times do
      spawn make_request(request_qtt/concurrency, channel)
    end
    concurrency.times { channel.receive }
  end

  private def make_request(instances, channel)
    client = HTTP::Client.new(API_URI)
    instances.times do
      result = client.get("/cocacola/media/")
      puts result.status_code
    end
  ensure
    client.close if client
    channel.send(nil)
  end
```

kostya:
can you give full example to reproduce? this is not compiled

nobilik:
Sorry, I fix it. Try now

nobilik:
Hi. There is one suggestion, I think the problem appear when I upgrade Crystal from 0.21 to 0.22. And in 0.22 have been added OpenSSL 1.1.0 support. I'm not sure, but maybe it's here. Because before I saw match more errors with SSL in my app. 
By the way, is there any way to downgrade Crystal version from 0.23 to 0.21? Thanks for help.

kostya:
if your try site without https, it hangs?

nobilik:
I try to make it with http and there was no issues for 50 loops. With https it happens every 5 loops or less

oprypin:
http probably just serves a redirect to https so it may not be a good indicator

nobilik:
@oprypin I've tested it on site without redirects, so I think the problem is somewhere between http client and openssl

ysbaddaden:
Which OpenSSL version is being linked and used?

ysbaddaden:
[...]

PR #6610, opened 2018-08-26
Fix zlib missing handling of buffer error

### LINKED ISSUE #6575: Zlib::Reader: Reading to EOF deadlocks application
`Zlib::Reader` can deadlock the application if you read to EOF:
```cr
require "zlib"

spawn do
  i = 0
  loop do
    puts "[#{i}] looping"
    i += 1
    sleep 1
  end
end

# Wait 3 sec to start
sleep 3

File.open("data.bin", "r") do |file|
  Zlib::Reader.new(file).gets_to_end
end
```
```
$ time cr run.cr
[0] looping                       
[1] looping                                      
[2] looping                                
^C                                                  
real    0m9.508s                      
user    0m3.162s                      
sys     0m3.205s
```

I've created https://github.com/z64/zlib-freeze which contains a sample data blob and the above code that demonstrates the deadlock, as well as an excerpt from `lldb` in `README.md`.

## Background

I found this while implementing business logic of a websocket protocol that sends chunks of zlib-compressed data over multiple frames, where you buffer each frame received until a "end of flush" binary suffix is read; so you must use a single instance of `Zlib::Reader` throughout your connections lifetime over multiple frames. Trying to inflate a single blob (shown in the above repo) deadlocked the app.

I suspect the correct behavior here should be to raise an exception; as a workaround I'm infalting into a slice (`reader.read(slice)`) which will not exhibit this deadlock behavior.

asterite:
Fixes #6575

The first commit fixes #6575: translating the [code that reproduces the bug](https://github.com/z64/zlib-freeze) to ruby raises "buffer error", and we were missing handling that error. I additionally added a check for other potential errors too.

The next commits refactor and improve zilb, fixing a few bugs too, where a function would return `Int` and it was expected to return an `Error` (an enum). This was easily detected by using the question methods of enums instead of comparing them with enum values.

Finally, peeking some Ruby code that deals with zlib, I found out about the `zError` function to get a nice error message in case zlib's stream doesn't provide one.

It might be nice to include this fix in 0.26.1 /cc @bcardiff

### DIFF
--- a/spec/std/zlib/reader_spec.cr
+++ b/spec/std/zlib/reader_spec.cr
@@ -66,7 +66,7 @@ module Zlib
       reader.read(slice).should eq(0)
     end
 
-    it "should raise bufer error on error (#6575)" do
+    it "should raise buffer error on error (#6575)" do
       io = IO::Memory.new("x\x9C4\xC9\xD1\n@@\u0010\u0005\xD0\u007F\xB9ϻeEj~E\xD2`B\xAD\xA55H\x9B\u007F\xE7\xC5۩\x93\xA0\xA0pxo\xB0\xFFX7\x90\xCB\f\u0006P\xC2$\u001C\xB5\u0013\xD6v\u000E*\xF1d\u000F*\\^~\xDFj\xE4^@5FV\xB9\xF8\xB6[\u001C\xEC\xC2s\xB0\x99\xD3\n\xCD\xF3\xBC\u0000\u0000\u0000\xFF\xFF")
 
       reader = Reader.new(io)
--- a/spec/std/zlib/reader_spec.cr
+++ b/spec/std/zlib/reader_spec.cr
@@ -71,7 +71,7 @@ module Zlib
 
       reader = Reader.new(io)
 
-      expect_raises(Flate::Error) do
+      expect_raises(Flate::Error, "flate: buffer error") do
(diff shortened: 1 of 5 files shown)


asterite:
Probably fixes #4650 too...

spec/std/zlib/reader_spec.cr
```diff
@@ -65,5 +65,15 @@ module Zlib
       slice = Bytes.empty
       reader.read(slice).should eq(0)
     end
+
+    it "should raise bufer error on error (#6575)" do
```
straight-shoota: buffer

asterite: Fixed in 5c314e018447ff08a8768321d82f06309b9a4737

bcardiff:
@asterite do you want do a rebase and squash that last commit or you prefer me to squash all together?

bcardiff:
I ment that the last commit (typo) could be squashed into one of the previous ones.
It's fine, I will merge directly this.

rdp:
I like this fix, it at allows fibers to at least not hang now, sweet!  I had some question marks about if it should retry on Z_BUF_ERROR if there's more input to be had but it can be addressed later, if it's even a problem, thanks! ref: http://zlib.net/zlib_how.html

PR #3514, opened 2016-11-07
Add intro sort to Array

c910335:
For most of the developers, performance is one of the most reasons why they choose a programming language.
But how do they know the performance of a programming language?
Usually, they test it by sorting an array because sorting is the most basic, classic and the most important algorithm in the world.
That is what I did these days.
After several tests, I found that Array#sort is usually faster than qsort in C but sometimes slower than std::sort in C++.
However, I believe that Crystal can do better.

Therefore, let me introduce "introspective sort" a.k.a. introsort, the fastest sorting algorithm as far as I know.
[Introsort - Wikipedia](https://en.wikipedia.org/wiki/Introsort)
According to Wikipedia, Introsort is a hybrid sorting algorithm that consists of quicksort, heapsort and insertion sort.
It combines the good parts of the three algorithms, with fast average performance (quicksort), worst-case O(n log n) runtime (heapsort) and fastest practical performance on an almost sorted array (insertion sort).
[Benchmarks](https://gist.github.com/c910335/4c9eb7f0b1c33a9d46e7e4b5059ced88)

Issue #6576, opened 2018-08-20
Resolve Generics in macros


vladfaust:
Original question: https://stackoverflow.com/questions/51903596/how-to-resolve-generics-in-macros

My usecase is getting full paths of reference properties to use it outside of the scope, e.g:

```crystal
class MyProject::Models::Post
  propery user : User
end

class MyProject::Models::User
  property posts : Array(Post) # This
end
```

The property `posts` could be used later outside `MyProject::Models` namespace (e.g., to use in Query builders in ORMs), therefore, a full path is needed to be known at the compile time. Despite of `user : User` resolving normally, `posts : Array(Post)` cannot be resolved because it's a `Generic`. It's not a consistent behavior IMO.

I currently use such monster in [core](https://github.com/vladfaust/core):

```crystal
type: (if _type.is_a?(Generic)
   if ("#{_type.name.resolve}" == "Array(T)")
     _type.name.resolve.stringify.gsub(/\([\w, ]+\)/, "(" + _type.type_vars.map do |v|
       v.resolve.stringify
     end.join(", ") + ")").id
   else
     _type
   end
 else
   _type.resolve # Why not just using .resolve on Generics? :(
 end)
```

But it could only resolve one level arrays and it's awful.

asterite:
Why do you need to resolve it?

vladfaust:
Okay, I haven't commented yet because I wanted to start rewriting Core, which needs custom resolve hacks for `Core::Query` to work properly at the moment, but may not need after the rewrite. However, I haven't started it yet, so the issue isn't relevant right now.

But. Even if there is no hurry in `Generic.resolve` for now for me, it feels like an unconsistent behaviour. I, as a developer, expect any type to be resolvable. If a type isn't resolvable, which means a full path to it cannot be found, that's weird.

PR #6611, opened 2018-08-26
Let Array#sort only use `<=>`, and let `<=>` return `nil` for partial comparability

### LINKED ISSUE #6608: Array#sort requires < and <= methods, not <=>

asterite:
Fixes #6608 

This PR does a few things:
- `Array#sort` and related methods used to use `<` and `<=` instead of `<=>`, as documented. This is fixed to only use `<=>`.
- Merge `Comparable` and `PartialComparable` into a single type. `<=>` can now return `nil`, meaning that the two objects aren't comparable. This can be used when two objects of the same type can be compared, but not always (partial order). One such example are floats: they are usually comparable, except for NaN values.
- Let `Array#sort` and related methods handle the case when `<=>` return `nil`, and raise an exception in that case
- Let `Number <=> Number` return `nil` for the case of NaN values, but provide an optimized implementation for `Int <=> Int` that doesn't return `nil`

I ran the benchmarks provided in [the PR that introduced `intro` sort](https://github.com/crystal-lang/crystal/pull/3514), but only checking for `sort` and `sort` with block, and compared them to the current implementation. The performance is the same, except when `<=>` can return `nil`, in which the sorting is a bit less than 3 times slower. But that makes sense, because there's a `nil` check for every comparison now. I personally prefer to give a correct result than to silently fail at runtime. But this performance penalty only applies when `<=>` returns `nil`, so it's not a very common case.

The reason for removing `PartialComparable` and merging it with `Comparable` is that it's simpler to have just one module, and let it automatically indicate partial order if `nil` is returned. If `nil` isn't returned, compiling with `--release` should give the same performance, because the check for `nil` will be optimized out.

This PR also fixes the fact that even though we had `PartialComparable`, you couldn't `sort` such types (`Array#sort` didn't handle the possibility of `nil` from `<=>`).

Note: I replaced an `Int32?` restriction with `U forall U` to allow for `sort`'s block to be of type `Int32` or `Int32?`, for performance reasons. Ideally we should be able to do `U < Int32?` or some other condition over a free variable, but that's currently not supported. Eventually we can improve that code.


src/array.cr
```diff
@@ -2022,14 +2038,26 @@ class Array(T)
...
         l.value = p.value
         l, p = p, p - 1
       end
       l.value = v
     end
   end
 
+  protected def self.cmp(v1, v2)
+    v = v1 <=> v2
+    raise ArgumentError.new("comparison of #{v1} and #{v2} failed") if v.nil?
```
Sija: In exception messages title-case if preferred. Ditto below.

straight-shoota:
I'd suggest to add the explanation about the performance regarding `Nil` to the API docs.

asterite:
@straight-shoota What do you suggest? Something like "You can return `nil` if your need partial ordering. However, that will result in a slower performance when sorting such objects. In that case, you might considering not modeling your program correctly (as you should) just so that you can gain a small performance improvement that will probably never be the bottleneck of your program"?

src/array.cr
asterite: I wish we hand't change that. This is how it looks right now:

```
Unhandled exception: Index out of bounds (IndexError)
  from src/indexable.cr:596:8 in 'at'
  from src/indexable.cr:73:5 in '[]'
  from baz.cr:2:1 in '__crystal_main'
```

It doesn't look good to have an uppercase letter after `:`. This is Ruby:

```
baz.cr:1:in `sort': comparison of Integer with 2 failed (ArgumentError)
	from baz.cr:1:in `<main>'
```

I'll capitalize it, but I'd like to discuss reverting this capitalization.

asterite:
@straight-shoota Or, well, I could add "Note that if  there is a chance that`<=>` returns `nil`, `Array#sort` will perform a bit slower".  But I'd rather not to, because then many would probably try to prematurely optimize their program at the cost of a worse modeling.

straight-shoota:
No offense taken. But I think it should be worth mentioning that there is a difference in performance depending on whether `Nil` type is included or not. If written positively, like for completely comparable types the comparator value is not nilable and the nil check will be optimized out.
Obviously it is not a good advice to apply an incorrect comparison model just to be faster. But it is good to know that when you have a completely comparable type, that it wont waste performance on nil checks.

This was actually my most important concern when reading the PR (and your comments in #6608), how performance would play out for sorting completely comparable types vs. partially comparable types. And  I think the implementation is really great.

asterite:
@straight-shoota Sounds good. Where should I add such comment? In `Comparable`, in `Comparable#<=>` or in one of the `Array#sort` methods?

straight-shoota:
I think I'd put it in `Comparable` following the description of `nil` return value.

asterite:
@straight-shoota Added the performance note.

src/array.cr
```diff
@@ -1561,7 +1561,15 @@ class Array(T)
   # b # => [3, 2, 1]
   # a # => [3, 1, 2]
   # ```
-  def sort(&block : T, T -> Int32) : Array(T)
+  def sort(&block : T, T -> U) : Array(T) forall U
+    # TODO: use a better way to check U < Int32?
+    {% begin %}
+      {% block_type = U.union? ? U.union_types.first { |t| t != Nil } : U %}
```
RX14: Can we add a `TypeNode#contains_type` method? Seems like a simple addition and would help this PR a lot.

bcardiff: Isn't `TypeNode#<=` is what you are looking for?

asterite: Yes, but `Int32?` doesn't work in macros. It's https://github.com/crystal-lang/crystal/issues/6576

I'll try to fix it later. But even if I fix it, I won't be able to use it until the next version (but this PR can wait).

asterite: I send #6617

With it, we can simply do `{% raise "..." unless U <= Int32? %}`, which is sooo much simpler :-)

RX14: I didn't know `<=` behaved correctly with union types :)

bcardiff: Should we create 0.28 milestone to keep track of this then?

asterite: I think it's the same code that's used in `is_a?`. There are many related methods that do more or less the same but eventually we should unify them. I now think improving stuff step by step is possible and much better than rewriting many parts.

straight-shoota:
@asterite Could you rebase this?

asterite:
@straight-shoota Done! :-)

src/comparable.cr
```diff
@@ -51,4 +70,13 @@ module Comparable(T)
   # [4, 7, 2].sort { |x, y| y <=> x } # => [7, 4, 2]
   # ```
   abstract def <=>(other : T)
+
+  private def _compare_with(other : T)
```
straight-shoota: Why does this method name begin with an underscore? This is not typically used for private methods.

asterite: I'm afraid that if someone includes `Comparable` and wants to define a `compare_with` method (for some other purpose) it will conflict with this method.

There are no real private methods in Crystal, everything leaks.

straight-shoota: Okay, sounds legit. But what about putting the code directly into each method? A little bit duplication is not an issue.  Then there would simply be no method that could be accidentally overridden.
And it would get rid of the somewhat ugly block short cuts `&.>(0)`.

asterite: Good idea. Done!

straight-shoota:
@asterite Needs a rebase!

src/array.cr
bcardiff: Note: This can be refactored now since 0.27 has the fix

asterite:
Rebased!

asterite:
There was a recent PR that improved `Comparable`'s docs. In this PR I had already done that so I kept my docs which also includes the fact that `nil` can be returned now from `<=>`. I should have mentioned that in the other PR to avoid wasting someone else's time.

src/array.cr
asterite: @bcardiff You are right! Updated :-)

src/comparable.cr
```diff
@@ -1,22 +1,40 @@
 # The `Comparable` mixin is used by classes whose objects may be ordered.
 #
 # Including types must provide an `<=>` method, which compares the receiver against
-# another object, returning a negative number, `0`, or a positive number depending
-# on whether the receiver is less than, equal to, or greater than the other object.
+# another object, returning:
+# - a value less than zero if `self` is less than the other object
+# - a value greater than zero if `self` is greater than the other object
+# - zero if `self` is equal to the other object
```
straight-shoota: I'd prefer to use the literal `0` instead of "zero", just like `nil`.

asterite: Done!

src/number.cr
```diff
@@ -166,6 +166,10 @@ struct Number
   # Returns `-1`, `0` or `1` depending on whether `self` is less than *other*, equals *other*
   # or is greater than *other*.
```
straight-shoota: The doc comment should mention that this method returns `nil` when comparing float with NaN value. Maybe also declare the return type as `Int32?`.

src/int.cr
```diff
@@ -237,6 +237,12 @@ struct Int
     end
   end
 
+  def <=>(other : Int)
```
straight-shoota: Maybe declare return type as `Int32` analog to my suggestion for `Number#<=>`. It also wouldn't hurt to copy the doc comment with a note that comparing ints won't return `nil`.

src/compiler/crystal/macros/methods.cr
```diff
@@ -422,7 +422,11 @@ module Crystal
       when "<="
         bool_bin_op(method, args) { |me, other| me <= other }
       when "<=>"
-        num_bin_op(method, args) { |me, other| me <=> other }
+        num_bin_op(method, args) do |me, other|
+          v = me <=> other
```
straight-shoota: Why not simply `(me <=> other) || NilLiteral.new`?

src/comparable.cr
```diff
@@ -1,22 +1,40 @@
...
-# `Comparable` uses `<=>` to implement the conventional comparison operators (`<`, `<=`, `==`, `>=`, and `>`).
+# `Comparable` uses `<=>` to implement the conventional comparison operators
+# (`<`, `<=`, `==`, `>=`, and `>`).
+#
+# Note that returning `nil` is only useful when defining a partial comparable
+# relationship. One such example is float values: they are generally comparable,
+# except for `NaN`. If none of the values of a type are comparable between each
+# other, `Comparable` shouldn't be included.
+#
+# NOTE: when `nil` is returned from `<=>`, `Array#sort` and related sorting
```
straight-shoota:
```suggestion
# NOTE: When `nil` is returned from `<=>`, `Array#sort` and related sorting
```

src/comparable.cr
```diff
@@ -1,22 +1,40 @@
...
-# on whether the receiver is less than, equal to, or greater than the other object.
+# another object, returning:
+# - a negative number if `self` is less than the other object
+# - a positive number if `self` is greater than the other object
+# - `0` if `self` is equal to the other object
+# - `nil` if `self` and the other object are not comparable
 #
-# `Comparable` uses `<=>` to implement the conventional comparison operators (`<`, `<=`, `==`, `>=`, and `>`).
+# `Comparable` uses `<=>` to implement the conventional comparison operators
+# (`<`, `<=`, `==`, `>=`, and `>`).
```
straight-shoota: Maybe it could be explicitly mentioned that these operators return `false` when the operands are not comparable (i.e. `<=>` returns `nil`).

src/compiler/crystal/macros/methods.cr
Sija: Because `me <=> other` can return `false`?

straight-shoota: The expected return type is `Int32?`. Implementations could obviously return whatever they like, but that will likely break anyway.

Besides, this is compiler code and we control all implementations of `#<=>` that could be called here and none of them returns `false`.

src/number.cr
asterite: Done

src/int.cr
asterite: Done

src/compiler/crystal/macros/methods.cr
asterite: Done

src/comparable.cr
asterite: Done

src/comparable.cr
asterite: Done

src/compiler/crystal/macros/methods.cr
asterite: Note that `(me <=> other) || NilLiteral.new` didn't work, the `return` here returns from the method , not from the block. I had to use `(me <=> other) || (return NilLiteral.new)`.

In general I'd prefer if we didn't discuss these minor style things, they lead to commits that fail, revertion, discussions, etc., for very little gain.

straight-shoota: It was really just meant as a question because I didn't know whether the `return` was necessary or not. You didn't need to change that. Sorry.

r00ster91:
Shouldn't the `src/partial_comparable.cr` file now be deleted?

asterite:
Yes, but it's a breaking change to do so, so we must be a bit careful.

r00ster91:
Unfortunately the `Deprecated` annotation doesn't support modules yet. I would say it should just be removed because it's not really used anywhere: https://github.com/search?l=Crystal&q=PartialComparable&type=Code.

straight-shoota:
Why don't we add a deprecation notice in the API docs and remove it in 0.29.0?

bcardiff:
We could flag the `<=>` method as deprecated during 0.28 if someone wants to tackle that.
The std lib path is not ignored by default, so that should warn about `PartialComparable` usages.

Maybe another check for deprecated methods is to detect overrides of abstract methods with `@[Deprecated]`.

PR #6616, opened 2018-08-28
Number: remove type suffixes in `inspect` output

### LINKED ISSUE #6615: How to get normal/clean numerical output display?
I raised this issue when this "problem" occurred with release 0.24.1 but I really need to know how to get around/fix it.

This code:
```
x = "73376604975172941348239581622402173809".to_big_i
print "factors  of #{x} are: "; print " => ", tm { print '\n', x.factors }
puts
```
currently produces this output for versions > 0.23.1
```
factors of 73376604975172941348239581622402173809 are: 
[[3_big_i, 1], [127_big_i, 1], [2388286630369_big_i, 1], [80639196284734379234981_big_i, 1]] => 00:00:00.118
```
I don't want numerical type references in my output display.  What do I have to do so the output displays normal/clean numerical values?
```
factors of 73376604975172941348239581622402173809 are: 
[[3, 1], [127, 1], [2388286630369, 1], [8063919628473437923498, 1]] => 00:00:00.118
```

asterite:
Fixes #6615
Reverts #4747

This change is about:

```crystal
# Current behavior
1_u64.inspect # => "1_u64"

# Proposed behavior
p 1_u64 # => "1"

# And same goes for any other numeric type: don't show the type suffix
```

I know I previously approved #4747, but I regret it.

As I previously mentioned, and #6615 is a clear example of that, the type information makes it harder to grasp and understand the output. If we want Crystal to be used for math/science stuff, I think we should show numbers as numbers, not as language constructs with type information.

I understand that type information can be valuable for debugging. But in that case it's simply a matter of using `typeof` or `.class` when debugging. We shouldn't always hammer the user with type information.

## Arguments against this change

And if the argument is "inspect should print something that is valid Crystal code and can construct back that value", well, that already doesn't apply for classes and structs, so it's generally not true.

## Other languages

No other language shows type information for numbers when printing them, either by themselves or inside other structures (arrays, hashes, even class fields).

Ruby doesn't show type information when printing numbers. I know, Ruby is dynamic and it looks like there's a single number type, but this is not true. There's Fixnum and Bignum, but you don't get this information when using `inspect` on them, even though it might be important from a debugging perspective (or is it important?).

[Julia has many number types](https://julia-ylwu.readthedocs.io/en/latest/manual/integers-and-floating-point-numbers.html) yet you don't see their types when outputting them.

## Let `Array#to_s` invoke `to_s`?

Now, one could say that the problem is that `Array#to_s` invokes `inspect` on each of its elements. Some suggested that `Array#to_s` should invoke `to_s` on its elements, while `Array#inspect` should invoke `inspect` on its elements. If we try it out, we get:

```crystal
ary = ["foo", "bar", "baz, qux"]
puts ary # =>[foo, bar, baz, qux]

# Or if we want to output the values to a file
File.open("file.txt", "w") do |file|
  file << ary
end
```

That is, `String#to_s` returns the string without quotes, so that doesn't seem to work very well. So using `inspect` makes sense, because we want to see each element clearly delimited. Or maybe we could let `Array#to_s` invoke `inspect` but only for String elements? I fear that if we go that direction it's more complex, and that we'll need these special rules for more types.

Alternatively, to print arrays we could say to always use `p` instead of `puts`, and to do `ary.inspect(file)`, but then `to_s` is also used in interpolation, for example `"foo #{ary}"`, so one would also have to do `"foo #{ary.inspect}"` (which is also a bit slower).

## Final words

More importantly, I think #4747 was merged without any associated bug or request, which makes me think someone maybe asked for this feature on IRC and a PR was rushed. There wasn't much discussion about this change either.

### DIFF
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -299,6 +299,10 @@ describe BigDecimal do
     BigDecimal.new(100, 4).to_s.should eq "0.01"
   end
 
+  it "#inspect" do
+    BigDecimal.new(1, 2).inspect.should eq "0.01"
+  end
+
   it "converts to other number types" do
     bd1 = BigDecimal.new(123, 5)
     bd2 = BigDecimal.new(-123, 5)
--- a/spec/std/big/big_float_spec.cr
+++ b/spec/std/big/big_float_spec.cr
@@ -169,7 +169,7 @@ describe "BigFloat" do
   end
 
   describe "#inspect" do
-    it { "2.3".to_big_f.inspect.should eq("2.3_big_f") }
+    it { "2.3".to_big_f.inspect.should eq("2.3") }
   end
 
   it "#hash" do
--- a/spec/std/big/big_int_spec.cr
+++ b/spec/std/big/big_int_spec.cr
@@ -259,7 +259,7 @@ describe "BigInt" do
   end
 
   describe "#inspect" do
-    it { "2".to_big_i.inspect.should eq("2_big_i") }
+    it { "2".to_big_i.inspect.should eq("2") }
   end
 
   it "does gcd and lcm" do
--- a/spec/std/float_spec.cr
+++ b/spec/std/float_spec.cr
@@ -190,7 +190,7 @@ describe "Float" do
     end
 
     it "does inspect for f32" do
-      3.2_f32.inspect.should eq("3.2_f32")
+      3.2_f32.inspect.should eq("3.2")
     end
 
     it "does inspect for f64 with IO" do
@@ -200,7 +200,7 @@ describe "Float" do
 
     it "does inspect for f32" do
       str = String.build { |io| 3.2_f32.inspect(io) }
-      str.should eq("3.2_f32")
+      str.should eq("3.2")
     end
   end
 
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -185,23 +185,23 @@ describe "Int" do
   end
 
   describe "#inspect" do
-    it "appends the type" do
+    it "doesn't append the type" do
       23.inspect.should eq("23")
-      23_i8.inspect.should eq("23_i8")
-      23_i16.inspect.should eq("23_i16")
-      -23_i64.inspect.should eq("-23_i64")
-      23_u8.inspect.should eq("23_u8")
-      23_u16.inspect.should eq("23_u16")
-      23_u32.inspect.should eq("23_u32")
-      23_u64.inspect.should eq("23_u64")
+      23_i8.inspect.should eq("23")
+      23_i16.inspect.should eq("23")
+      -23_i64.inspect.should eq("-23")
+      23_u8.inspect.should eq("23")
+      23_u16.inspect.should eq("23")
+      23_u32.inspect.should eq("23")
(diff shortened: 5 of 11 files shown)


asterite:
@Sija Good catch, thanks! I updated this PR.

bew:
What is supposed to be the difference between `#to_s` and `#inspect`?
Shouldn't `#inspect` give more information about the object, type etc..?

wontruefree:
I dont have a strong opion either way but it seems like a fair ammount of `inspects` more closely follow their literal form.  I would be fine with either one but I wanted to point this out.

```crystal
puts "foo".inspect #=> "foo"
puts "foo"         #=> foo
puts :foo.inspect  #=> :foo
puts :foo          #=> foo
puts /foo|bar/.inspect #=> /foo|bar/
puts /foo|bar/         #=>(?-imsx:foo|bar)
```

RX14:
The only time people complain about this output is when it's invoked by array. So really, I think we should fix that. Number type is very important in crystal, unlike in ruby.

```java
System.out.println(Arrays.asList("a", "b")); // => [a, b]
```
```cs
Console.WriteLine(new List<string>() { "a", "b" }); // System.Collections.Generic.List`1[System.String] (lol)
```
```js
console.log(["a", "b"].toString()) // => a,b
```
```py
print(str(["a", "b"])) # => ['a', 'b']
```

Maybe ruby and python are the odd ones out, and maybe `["a", "b"].to_s` being `[a, b]` isn't so confusing after all. If we consider `to_s` being for user display and `inspect` being for developer/debugging display it makes perfect sense.

asterite:
Here's one more argument against showing the type of literals in `inspect`.

For example, arrays are inspected like this:

```crystal
a = [1, 2, 3]
a.inspect # => [1, 2, 3]
```

However, we can also do:

```crystal
a = [1, 2, 3] of Int32 | Float64
a.inspect # => [1, 2, 3]
```

Note that we don't show the array's type. So if the reason for showing the type is "construct the object back", then the above won't work. If the reason if making it easier to understand the types in your program, then the above doesn't work either. We would have to show it like `[1, 2, 3] of Int32 | Float64`... basically we would always have to show the type. And I'm pretty sure we don't want to go with that.

Trying to define the difference between `to_s` and `inspect`, I can't come up with a good explanation. One that I think we can use is that `inspect` must print something clearly delimited, so that you can show it inside other types. For example, `"foo, bar".to_s` is just `foo, bar`, and showing that inside another type can be confusing: `[foo, bar]`. But `"foo, bar".inspect` is `"foo, bar"`, and inside other types there's no confusion. It will even escape quotes to make this clear. I _think_ String is the only example of this, but I think it's a good reason. And with this, the reason is not "show also the type, or be able to construct the object back to Crystal".

bew:
Sure you mark a point, but having a way to view the type alongside each values is very valuable, and as you pointed out, it would be best to show the array's type as well.
If not through the `inspect` method, I would like to see this through another method like (_warning: ugly name_) `debug_inspect`, that can be used when debugging some program.
Or maybe once we managed to define what `inspect` is for exactly, rename that method, because `inspect` to me is like `debug_inspect` !
Or if it only make sense for `String`s, maybe have a dedicated method to make a delimited string.

Sija:
I guess this can be closed too :)

PR #6617, opened 2018-08-28
Compiler: resolve generics in macros, and add `Genreic#resolve` and `Generic#resolve?` macro methods

### LINKED ISSUE #6576: Resolve Generics in macros

asterite:
Fixes #6576

This code right now works:

```crystal
{{ String }}
```

`String` (a `Path` AST node) is resolved in macros to a `TypeNode` denoting its type, or a compile-error ir raised. However, the same thing with generics (a `Generic` AST node) doesn't work:

```crystal
{{ Array(String) }} # can't execute Generic in a macro
```

This should work in the same way, resolving to a `TypeNode`. This is what this PR fixes.

It also adds `resolve` and `resolve?` macro methods to the `Generic`. For example:

```crystal
# Macros always receive AST nodes, and we need `resolve` to solve them to a TypeNode
macro abstract?(node)
  {{ node.resolve.abstract? }}
end

p abstract? IO            # => true
p abstract? String        # => false
p abstract? Array(String) # => false
```

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -1618,6 +1618,24 @@ module Crystal
       it "executes named_args" do
         assert_macro "x", %({{x.named_args}}), [Generic.new("Foo".path, [] of ASTNode, named_args: [NamedArgument.new("x", "U".path), NamedArgument.new("y", "V".path)])] of ASTNode, "{x: U, y: V}"
       end
+
+      it "executes resolve" do
+        assert_macro "x", %({{x.resolve}}), [Generic.new("Array".path, ["String".path] of ASTNode)] of ASTNode, %(Array(String))
+
+        expect_raises(Crystal::TypeException, "undefined constant Foo") do
+          assert_macro "x", %({{x.resolve}}), [Generic.new("Foo".path, ["String".path] of ASTNode)] of ASTNode, %(Foo(String))
+        end
+
+        expect_raises(Crystal::TypeException, "undefined constant Foo") do
+          assert_macro "x", %({{x.resolve}}), [Generic.new("Array".path, ["Foo".path] of ASTNode)] of ASTNode, %(Array(foo))
+        end
+      end
+
+      it "executes resolve?" do
+        assert_macro "x", %({{x.resolve?}}), [Generic.new("Array".path, ["String".path] of ASTNode)] of ASTNode, %(Array(String))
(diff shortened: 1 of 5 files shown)


vladfaust:
Thank you, @asterite ❤️

vladfaust:
What about Unions? They aren't resolvable as well ATM.

asterite:
Yes, you can write `Union(Sting, Int32) `

vladfaust:
@asterite,

```crystal
macro type(type)
  {% pp! type.resolve %}
end

type Int32         # type.resolve # => Int32
type Int32?        # type.resolve # => (Int32 | Nil)
type Array(Int32)? # type.resolve # => (Array(Int32) | Nil)
```

but

```crystal
macro type(declaration)
  {% pp! declaration.type.resolve %}
end

type id : Int32                     # declaration.type.resolve # => Int32
type id : Int32?                    # undefined macro method 'Union#resolve'
type id : Array(Int32)?             # undefined macro method 'Union#resolve'
type id : Union(Int32 | Nil)        # declaration.type.resolve # => (Int32 | Nil)
type id : Union(Array(Int32) | Nil) # declaration.type.resolve # => (Array(Int32) | Nil)
```

Why is it so?

vladfaust:
@asterite bump. Maybe you just forgot something little to make `?` work? If so, fixing it would make the PR complete

PR #6618, opened 2018-08-29
Compiler: don't "ambiguous match" if there's an exact match

### LINKED ISSUE #6601: ambiguous call matches both Int8 and Int16
The following code errors:

```cr
a = 1 || "foo"
a + 2 # instantiating '(Int32 | String)#+(Int32)':  ambiguous call matches both Int8 and Int16
```

It's odd that this mentions `Int8` and `Int16` which are not even used explicitly in this example. I guess it's because when the literal type is expanded it first notices that these could be ambiguous. It should definitely not compile because `String#+` doesn't accept any integer, but the error message is wrong.

Even if the integer literals are typed explicitly, the error is similar:
```cr
a = 1_i32 || "foo"
a + 2_i32 # instantiating '(Int32 | String)#+(Int32)': ambiguous call matches both Int8 and Int16
```

The expected error should be something like `no overload matches '(Int32 | String)#+' with type Int32`.

Extracted from #6593

Introduced by #6074

asterite:
Fixes #6601

The algorithm for looking up call matches is:
1. try lookup without automatic casts
2. if that fails, retry with automatic casts

In the second step, if an overload for a literal matched two different integer types (for example `1` matched `Int8` and `Int16`) then an "ambiguous call" error was raised. The problem is that this "ambiguous call" check didn't account for the possibility of 1 failing because of an error related to another type in the union, and the fact that there was no ambiguos call when using the new "literal types" for automatic casts, because the check was done eagerly instead of cancelling the check where we find an exact overload match.

This PR fixes that by tracking whether there's an ambiguous call, and noticing that there's no ambiguos call if there's an exact match. Then at the end of the lookup we check if there's an ambiguos call.


src/compiler/crystal/types.cr
```diff
@@ -1321,9 +1328,22 @@ module Crystal
...
+    # The literal associated with this type
     getter literal : NumberLiteral
+
+    # The most exact match type, or the first match otherwise
     @matched_type : Type?
 
+    # A type that signals it's an ambiguous call, because the literal
+    # can match two types (for example when 1 matches both Int8 and Int16
+    # restrictions, but there's no Int32 restriction)
+    @ambiguous_call : Type?
```
bcardiff: If the call is ambiguous the value of this ivar will be a union types of the possibilities?

src/compiler/crystal/types.cr
```diff
@@ -1321,9 +1328,22 @@ module Crystal
...
+
+    # True if this literal matched against the exact type (for example 1
+    # matched an Int32 restriction). If this is true, there's never an
+    # ambiguous call (so `@ambiguous_call` will be `nil`).
+    @exact_match = false
```
bcardiff: If `@exact_match == @ambiguous_call.nil?` always, then it should be a function directly. or just remove it since the usages are already guarded by `@ambiguous_call` been not `nil`.

src/compiler/crystal/types.cr
asterite: No, just the first type. I can change it to be an arrray of all types, but maybe it's not necessary.

src/compiler/crystal/types.cr
asterite: Not necessarily. Say you have the literal `1` and these overloads:
1. Int8: it matches, to `@matched_type` becomes `Int8`
2. Int16: it matches, so `@ambiguous_call` becomes `Int16`
3. Int32: it matches exactly... right now we set `@matched_type` to it, `@ambiguous_call` to `nil` and `@exact_match` to `true`...
4. Int64: it matches, but since `@exact_match` is true, we don't care anymore

I'm thinking how to do it without the `@exact_match` variable, if that's what you are suggesting. I think we can do it by comparing `@matched_type == literal.type`, but we'd have to do that check all the time. Maybe it's not that terrible because it's just a reference comparison.

Is this the change you mean?

asterite:
@bcardiff I just updated the PR, thanks for the comments!

I made a few changes:
- I refactored the code and, I think, improved it's readability
- The error message for "ambiguous call" is a bit more clear and will show all types that matched
- It now fixes #6620

asterite:
@bcardiff I think I can make it use less instance variables, just having `@matches` have all matches, and if there's an exact match we keep it with just one type. But I'm not sure that additional change will make the code easier to understand.

PR #6627, opened 2018-08-30
Allow comments after :nodoc: marker

Sija:
Right now comment contents must be exactly `# :nodoc:`, otherwise they'll still be included in the docs. This PR allows for following comments, checking only for `:nodoc:` marker at the beginning.

This is preferrable since at times there's a wish to include documentation in the code (usually related to the technical side of things), which won't be included in generated API docs.

### DIFF
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -163,7 +163,8 @@ class Crystal::Doc::Generator
   end
 
   def nodoc?(str : String?)
-    str == ":nodoc:" || str == "nodoc"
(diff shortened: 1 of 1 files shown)


asterite:
We could check for starts with and ends with.

Sija:
> We could check for starts with and ends with.

In my opinion it shouldn't be there since it reduces its visibility. You can't miss it in the beginning of the doc, but you might at the end, especially if it's a long one.

I mean it's not a big deal and I can add it, just sayin' :)

asterite:
You are right

bew:
Nice, ty all here :tada:

PR #6628, opened 2018-08-31
Fix to_h inspect in NamedTuple

icyleaf:
Test Code:

```crystal
def foo(**options)
  unless options.empty?
    options.to_h
  end
end

puts foo
```

Current wrong result: 

```
in line 3: Can't convert an empty NamedTuple to a Hash
```

Now correct result:

```
nil
```

### DIFF
--- a/spec/compiler/semantic/named_tuple_spec.cr
+++ b/spec/compiler/semantic/named_tuple_spec.cr
@@ -286,15 +286,6 @@ describe "Semantic: named tuples" do
       )) { named_tuple_of({"x": types["Foo"].virtual_type!, "y": types["Foo"].virtual_type!}) }
   end
 
-  it "does not compile to_h of empty tuples" do
-    # TODO change the location of this spec upon #2391
-    assert_error %(
-      require "prelude"
-      NamedTuple.new.to_h
-      ),
-      "Can't convert an empty NamedTuple to a Hash"
-  end
-
   it "types T as a tuple of metaclasses" do
     assert_type("
       struct NamedTuple
--- a/spec/std/named_tuple_spec.cr
+++ b/spec/std/named_tuple_spec.cr
@@ -272,10 +272,6 @@ describe "NamedTuple" do
     tup1 = {a: 1, b: "hello"}
     hash = tup1.to_h
     hash.should eq({:a => 1, :b => "hello"})
-
-    expect_raises Exception, "Can't convert an empty NamedTuple to a Hash" do
-      NamedTuple.new.to_h
-    end
   end
 
   it "does to_s" do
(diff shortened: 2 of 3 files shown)


asterite:
We should return `{} of NoReturn => NoReturn` for an empty named tuple, not raise an exception.

icyleaf:
Hi @asterite, i updated return `{} of Return => Return` and add new test

```crystal
NamedTuple.new.to_h.should eq({} of Return => Return)
```

It throws an exception:

```
Missing hash key: "key" (KeyError)
  from src/hash.cr:0:9 in 'fetch'
  from src/hash.cr:62:5 in '[]'
  from src/compiler/crystal/codegen/codegen.cr:1446:17 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:2068:7 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:602:9 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:2068:7 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:807:11 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:2068:7 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:602:9 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/codegen/codegen.cr:2068:7 in 'accept'
  from src/compiler/crystal/codegen/call.cr:286:13 in 'codegen_call_with_block'
```

If it could not to "visit` the value, it could not return this, right?

Sija:
@icyleaf You mean _**No**Return_ right?

icyleaf:
@Sija Oops, you got me, updated.

Sija:
Maybe `{} of Nil => Nil` or `{} of Void => Void` would do the trick?

MakeNowJust:
`Tuple.new.to_a` returns `[] of NoReturn`, so it looks good idea that `NamedTuple.new.to_h` returns `{} of NoReturn => NoReturn`

icyleaf:
@MakeNowJust `puts Tuple.new.to_a` is works, not `puts NamedTuple.new.to_h`, there is something wrong with `codegen.cr`, i did not see it deep.

MakeNowJust:
@icyleaf This problem is solved by #6659. But this fix is unavailable until next version even if #6659 is merged...

icyleaf:
@MakeNowJust Nice!!
@asterite Updated, review again please, thanks.

PR #6629, opened 2018-08-31
Check for FreeBSD 11.x instead of 12

myfreeweb:
64-bit inodes are used in 12 and newer. Since 12 will soon be released, and CURRENT will become 13, switch to checking for the older version.

### DIFF
--- a/src/compiler/crystal/semantic/flags.cr
+++ b/src/compiler/crystal/semantic/flags.cr
@@ -29,6 +29,7 @@ class Crystal::Program
     set = flags_name.map(&.downcase).to_set
     set.add "darwin" if set.any?(&.starts_with?("macosx")) || set.any?(&.starts_with?("darwin"))
     set.add "freebsd" if set.any?(&.starts_with?("freebsd"))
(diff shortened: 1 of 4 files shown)


RX14:
if the freebsd ABIs are completely different for each release then they should go into whole different libc folders.

myfreeweb:
That would be a ridiculous amount of duplication.

RX14:
@myfreeweb it's already duplicated between each target because these files are supposed to be machine-generated. They're currently not and so the current workflow is stupid.

ysbaddaden:
We already "support" FreeBSD 12 which will be released in November, so there is no rush to merge.

This patch breaks FreeBSD 10.x and I can't find when it won't be supported anymore officially. Is it when FreeBSD 12 is released?

myfreeweb:
@ysbaddaden when 12 is released, CURRENT will become "13" and break

myfreeweb:
CURRENT is 13 now (12 is branched into STABLE), please merge this

j8r:
It would be a nice to have for the 0.27.0.

PR #6632, opened 2018-09-01
Fix: don't modify ARGV array when parsing options

stakach:
You might want to reference the arguments somewhere else in the application


asterite:
This is not a fix. Ruby works like that too. The idea is that `OptionParser` will remove the options used, and then you are left with the program arguments. You can call `parse(ARGV.dup)` if you want to.

stakach:
Thanks for the pointer

PR #6635, opened 2018-09-01
Optimize `Indexable#join` when all elements are strings

### LINKED ISSUE #6634: Benchmark realtime on Linux
Ubuntu 16.04.1
1GB RAM (Free memory 400MB+)
On VPS machine

Each test is ran 3 times:

ruby -v
Ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]

crystal -v
Crystal 0.26.1 [391785249] (2018-08-27)
LLVM: 4.0.0

$ crystal run --release 1.rb
00:00:00.033265433

$ ruby 1.rb
0.015399671000011494

1.rb
```
require "benchmark"

cols = 10
rows = 1000
data = Array.new(rows) { Array.new(cols) { "x"*1000 } }

time = Benchmark.realtime do
  csv = data.map { |row| row.join(",") }.join("\n")
end

puts time
```

I'm surprise if an unopitimize code does affect Crystal to run slower than Ruby?
Is it due to LLVM 4.0 on Ubuntu being too old where LLVM 6.0 on macOS Homebrew is newer?
Tested latest Crystal 0.26.1 on macOS Mojave is definitely slower than Ruby.2.5 as other have reported.

asterite:
Fixes #6634

This PR overrides `Indexable(T)#join` and if:
- `T` is a `String` or 
- `T` is a union that contains `String`, and all elements are strings
then an optimized version of `join` is executed. In this case we can compute the total bytesize of the returned string: `separator.bytesize * (array.size - 1) + array.sum(&.bytesize)`. This prevents potential string reallocations that happen in the default algorithm when `String.build` is used.

I used a slightly modified version of the code in #6634 to benchmark this:

```crystal
require "benchmark"

cols = 10
rows = 1000
data = Array.new(rows) { Array.new(cols) { ("x"*1000) } }

Benchmark.ips do |x|
  x.report("doit") do
    csv = data.map { |row| row.join(",") }.join("\n")
  end
end
```

Before:

```
doit  91.19  ( 10.97ms) (± 1.70%)  75882435 B/op  fastest
```

After:

```
doit 183.46  (  5.45ms) (± 9.41%)  20050063 B/op  fastest
```

(note the allocated memory per op too)

In Ruby (change `require "benchmark"` to `require "benchmark/ips" using [this gem](https://github.com/evanphx/benchmark-ips)):

```
                doit    117.854  (±10.2%) i/s -    585.000  in   5.030671s
```

So now it's faster than Ruby. Just note that if you run the original code from #6634 Crystal seems to be just a little slower than Ruby depending on the run, but a proper benchmark (`Benchmark.ips`) shows the real stuff.

## Further optimization

There are two cases when this optimization is applied:
- `T` is a `String` or 
- `T` is a union that contains `String`, and all elements are strings

The first one will be trivially faster than the previous one.

The second one involves an `all?(&.is_a?(String))` check. I still think doing this check is faster than reallocating memory, so that's probably good. Ruby always does this check (it has no compile-time information) but while doing it it also computes the expected total bytesize, and switches to the default/slow implementation if an element is not a string. We could do the same, computing the length while checking `.is_a?(String)` but I think that will make the code a lot harder to read and understand, so maybe it's not worth it. But we could eventually do it if we find it's a significant performance difference (I doubt it).

Also, when there's no `String` in the array, the implementation is the default (slow) one.

### DIFF
--- a/src/indexable.cr
+++ b/src/indexable.cr
@@ -289,10 +289,10 @@ module Indexable(T)
     return "" if empty?
 
     {% if T == String %}
-      _join_strings(separator)
+      join_strings(separator)
     {% elsif String < T %}
       if all?(&.is_a?(String))
-        _join_strings(separator)
+        join_strings(separator)
       else
         super(separator)
       end
@@ -301,7 +301,7 @@ module Indexable(T)
     {% end %}
   end
 
-  private def _join_strings(separator)
+  private def join_strings(separator)
     separator = separator.to_s
 
     # The total bytesize of the string to return is:


src/indexable.cr
```diff
@@ -281,6 +281,66 @@ module Indexable(T)
...
+  def join(separator = "")
+    return "" if empty?
+
+    {% if T == String %}
+      _join_strings(separator)
+    {% elsif String < T %}
+      if all?(&.is_a?(String))
+        _join_strings(separator)
+      else
+        super(separator)
```
asterite: I wanted to use `super` here but #6636

src/indexable.cr
```diff
@@ -281,6 +281,66 @@ module Indexable(T)
     self
   end
 
+  # Optimized version of `Enumerable#join` that performs better when
+  # all of the elements in this indexable are strings: the total string
+  # btyesize to return can be computed before creating the final string,
```
Sija: typo: `btyesize`

asterite: Thanks, fixed!

src/indexable.cr
```diff
@@ -281,6 +281,66 @@ module Indexable(T)
...
+          buffer += separator.bytesize
+        end
+
+        # Copy element to buffer
+        buffer.copy_from(elem.to_unsafe, elem.bytesize)
+        buffer += elem.bytesize
+
+        # Check whether we'll know the final UTF-8 size
+        if elem.size_known?
+          size += elem.size
```
RX14: This is wrong because the bytesize of the seperator isn't added. The specs pass though, is there any way to test this in the specs?

src/string.cr
```diff
@@ -4168,7 +4168,8 @@ class String
     return 4
   end
 
-  protected def size_known?
+  # :nodoc:
+  def size_known?
```
asterite: I need this in `Indexable#join` so I made it `:nodoc:`. It might be confusing to expose it as a public method because calling `size` always gets you the correct `size`. We can discuss exposing this in a separate issue/PR.

src/indexable.cr
asterite: Oh, good catch! Yes, I'll check it now.

asterite: Fixed!

asterite:
OK, this PR got a tad bigger now. @RX14 found that I had a bug when computing a string's size. So:
- I first fixed that but first checking that the sizes (and just in case bytesizes) match
- I extracted a private method to do it
- then I remembered we also check for `size` and `bytesize` equality for stirngs in `string_spec.cr`, because we manually construct strings in `String`
- I thought about having an `eq_string` matcher to compare strings, adding it to the `support` directory and requiring it when needed, but that's bug prone: what if you compare two strings and forget to check for their `bytesize` and `size`, and you also forget to require that file from `support`?
- so I then changed the `eq` matcher to always perform a `bytesize` and `size` equality when comparing strings

This actually uncovered a bug in `String#gsub(Char, Char)`. So I think having `eq` do this comparison by default is good because we create such strings in the whole standard library, but anyone else is also free to use `String.new` in their shards and they might put incorrect values for `bytesize` and `size` there. The `eq` matcher got a bit more complex, and it might be just slightly slightly slower, but I'm pretty sure it's insignificant.

src/indexable.cr
```diff
@@ -281,6 +281,69 @@ module Indexable(T)
...
+        _join_strings(separator)
+      else
+        super(separator)
+      end
+    {% else %}
+      super(separator)
+    {% end %}
+  end
+
+  private def _join_strings(separator)
```
RX14: I don't like the convention of using `_name` for private names when crystal already has private methods, just `private def join_strings` is fine.

src/spec/expectations.cr
```diff
@@ -5,17 +5,49 @@ module Spec
...
     def failure_message(actual_value)
-      expected = @expected_value.inspect
-      got = actual_value.inspect
-      if expected == got
-        expected += " : #{@expected_value.class}"
-        got += " : #{actual_value.class}"
+      expected_value = @expected_value
+
+      # Check for the case of string equality when the content match
+      # but not the bytesize or size.
```
RX14: How about recalculating the size of both the actual and expected values and report an error that either the actual or expected values are invalid? This would be more helpful and precise information, and it would be easy to calculate.

src/indexable.cr
asterite: The problem with private methods is that they can still be accessed by subclasses or including types. But yeah, I'll change it.

src/spec/expectations.cr
asterite: Sorry, I don't understand what you mean.

RX14: instead of just comparing the bytesizes, work out their actual bytesize using `Char::Reader` and use that as the reference

asterite: I know, you are talking about the case when you pass a computed string to `eq`. I think doing that is a bit overkill.

asterite: Sorry for the long conversation, but for example one might do:

```crystal
some_computed_string = ...
some_computed_string.should eq("foo" * 3)
```

Now, there might be a bug in `"foo" * 3` that returns an invalid `bytesize` or `size`. But `"foo" * 3` is already tested in the standard library. If you use another custom method as an argument to `eq`, you should test it separately, and of course you won't test it by supplying values that use the same function that you are testing... so I think assuming what you pass in `eq` already has correct values of `bytesize` and `size` is fine.

PR #6641, opened 2018-09-02
Don't try to mess with closed FDs when creating a subprocess

Timbus:
So when on a mac, spawning a child process crashes if stdin is closed (and probably the other handles). This makes `entr -r` break my apps when trying to run them.
This is because when forking off a child, crystal tries to fiddle with the standard handles, which you cannot do if they're closed. This patch checks to see if the handle is closed before messing with it.

While I'm at it, the existing `Redirect::Close` behavior seems really strange to me, too. Opening `/dev/null`, duping it, closing it, *then* setting flags on it seems to bug out on the mac.. I've moved the `close_on_exec` call to inside the `File.open` block for now, which seems to work, but it feels like there is a bigger problem here with how BSD treats filehandles.

### DIFF
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -154,7 +154,8 @@ class IO::FileDescriptor < IO
 
   def reopen(other : IO::FileDescriptor)
     system_reopen(other)
-
+    # You shouldn't be able to reopen a closed FD
+    @closed = false
     other
   end
 
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -13,7 +13,7 @@ class IO::FileDescriptor < IO
   def initialize(@fd, blocking = false)
     @closed =
       {% unless flag?(:win32) %}
-        fcntl(LibC::F_GETFD) < 0
+        !fcntl(LibC::F_GETFD) rescue true
       {% else %}
         false
       {% end %}
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -11,12 +11,7 @@ class IO::FileDescriptor < IO
   getter fd
 
   def initialize(@fd, blocking = false)
-    @closed =
-      {% unless flag?(:win32) %}
-        !fcntl(LibC::F_GETFD) rescue true
-      {% else %}
-        false
-      {% end %}
+    @closed = system_closed?
 
     unless blocking || {{flag?(:win32)}}
       self.blocking = false
@@ -154,8 +149,7 @@ class IO::FileDescriptor < IO
 
   def reopen(other : IO::FileDescriptor)
     system_reopen(other)
-    # You shouldn't be able to reopen a closed FD
-    @closed = false
+
     other
   end
 
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -11,7 +11,12 @@ class IO::FileDescriptor < IO
   getter fd
 
   def initialize(@fd, blocking = false)
-    @closed = false
+    @closed =
+      {% unless flag?(:win32) %}
+        fcntl(LibC::F_GETFD) < 0
+      {% else %}
+        false
+      {% end %}
 
     unless blocking || {{flag?(:win32)}}
       self.blocking = false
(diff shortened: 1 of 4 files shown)


Timbus:
Test case in bash/fish:
```bash
crystal eval 'puts `echo "hello"`' <&-
```

The `<&-` closes stdin, which should in no way affect `echo`.

src/process.cr
```diff
@@ -415,17 +415,20 @@ class Process
     when IO::FileDescriptor
       src_io.blocking = true
       dst_io.reopen(src_io)
+      dst_io.close_on_exec = false
     when Redirect::Inherit
-      dst_io.blocking = true
+      if LibC.fcntl(dst_io.fd, LibC::F_GETFD) >= 0
```
RX14: Please use `dst_io.fcntl`.

RX14:
It concerns me that the exception fixed by this bug fails to raise
```
$ crystal eval 'puts `echo "hello"`' <&-
[0x560afce68296] ???
[0x560afc554c7b] __crystal_raise +43
[0x560afc5561fc] ???
[0x560afc558895] main +565
[0x7f3d81990223] __libc_start_main +243
[0x560afc5513be] _start +46
[0x0] ???
```

However, this bug affected me too and i'm glad the cause has been found. I wish the close_on_exec didn't have to be moved though.

Timbus:
I've had a few minutes this morning with my work machine, and yeah I could replicate it all in C.
It dawned on me that what's happening is that if stdin is closed.. the next file to be opened will be set to the next available FD.. Which is going to be 0! So we're opening devnull, duping stdin over itself, closing it.. and then trying to set flags on it -_-

Knowing that.. I feel that this fix is a (albeit robust) bandaid in some ways. Hm.

Timbus:
Side note; the reason the error is uncaught is likely because this stuff happens right after a fork.

And I figured out a good way to put `close_on_exec` back :)

Timbus:
I don't know if that travis test is failing because of these changes.. It's suspiciously close to the altered code, but the error seems totally unrelated

Sija:
@Timbus that's a known intermittent error unrelated to this PR.

Timbus:
This seems to be as polished as I can make it. It is annoying that libevent is reopening handles, which makes the whole 'closed' check a bit of a gamble in other cases.
Will need to look into that later.

src/io/file_descriptor.cr
```diff
@@ -149,7 +154,8 @@ class IO::FileDescriptor < IO
 
   def reopen(other : IO::FileDescriptor)
     system_reopen(other)
-
+    # You shouldn't be able to reopen a closed FD
+    @closed = false
```
RX14: This should be put in `system_reopen`, just after the `dup` call, so there's not a race between the reschedule and setting `@closed = false`.

src/io/file_descriptor.cr
```diff
@@ -11,7 +11,12 @@ class IO::FileDescriptor < IO
   getter fd
 
   def initialize(@fd, blocking = false)
-    @closed = false
+    @closed =
+      {% unless flag?(:win32) %}
+        !fcntl(LibC::F_GETFD) rescue true
+      {% else %}
+        false
+      {% end %}
```
RX14: perhaps we should define a `system_closed?` and check that here and after reopen.

src/io/file_descriptor.cr
```diff
@@ -11,7 +11,12 @@ class IO::FileDescriptor < IO
   getter fd
 
   def initialize(@fd, blocking = false)
-    @closed = false
+    @closed =
+      {% unless flag?(:win32) %}
+        !fcntl(LibC::F_GETFD) rescue true
```
straight-shoota: You could avoid raising and immediately catching an exception by using `LibC.fcntl` directly. Or alternatively, implement an non-raising `fctnl?` method.

ysbaddaden: Hence why @RX14 proposed to have a `system_closed?` method that would call `LibC.fcntl()` directly, I guess.

Timbus: Dang, I was just trying to follow the advice from the first review.

RX14 has a good idea, I'll move it down a level into `system_closed?`. I'm just trying to touch as little as possible though.

src/io/file_descriptor.cr
Timbus: No need to check after reopen since it has to be open if dup succeeded. Still made the method though, it looks nicer.

PR #6648, opened 2018-09-02
Semantic: correctly guess ivar type from splat argument

MakeNowJust:
Caught from https://twitter.com/n_siena/status/1026021219714269184 (Japanese).

Before this PR the type of ivar `@foo` of `def initialize(*@foo : Int32)` is guessed to `Int32`. It causes very strange error. For example:

```crystal
class Foo
  def initialize(*@foo : Int32)
  end
end

Foo.new(1)
# Error: instance variable '@foo' of Foo must be Int32, not Tuple(Int32)
```

I think it should not be guessed to any types because this restriction means variable-length tuple and we have no type for it. `Indexable(Int32)` is one of generic types of variable-length tuple, but it is not perfect. So, it should be specified explicitly if needed.

Nevertheless, `@foo` type of `def initialize(*@foo : *{Int32})` can be guessed to `{Int32}` easily. It should be guessed.

A double-splat and ivar has the same problems. This PR fixes them. Now we get this error and result.

```crystal
class Foo
  def initialize(*@foo : Int32)
  end
end

Foo.new(1)
# Error: Can't infer the type of instance variable '@foo' of Foo
```

```crystal
alias T1 = {Int32}

class Foo
  getter foo
  def initialize(*@foo : *T1)
  end
end

pp Foo.new(1).foo.class # => {Int32}
```

### DIFF
--- a/src/compiler/crystal/semantic/type_guess_visitor.cr
+++ b/src/compiler/crystal/semantic/type_guess_visitor.cr
@@ -25,6 +25,9 @@ module Crystal
 
     @args : Array(Arg)?
     @block_arg : Arg?
+    @splat_index : Int32?
+    @double_splat : Arg?
+    @splat : Arg?
 
     @args_hash_initialized = true
     @args_hash = {} of String => Arg
@@ -800,6 +803,20 @@ module Crystal
       if arg = args_hash[node.name]?
         # If the argument has a restriction, guess the type from it
         if restriction = arg.restriction
+          # It is for something like `def foo(*@foo : *T)`.
+          if @splat == arg
+            # If restriction is not splat (like `*foo : T`),
+            # we cannot guess the type.
+            # (We can also guess `Indexable(T)`, but it is not perfact.)
+            # And this early return is no problem because splat argument
+            # cannot have a default value.
+            return unless restriction.is_a?(Splat)
+            restriction = restriction.exp
+            # It is for something like `def foo(**@foo : **T)`.
+          elsif @double_splat == arg
+            return unless restriction.is_a?(DoubleSplat)
+            restriction = restriction.exp
+          end
           type = lookup_type?(restriction)
           return type if type
         end
@@ -1096,6 +1113,8 @@ module Crystal
       @found_self = false
       @args = node.args
       @block_arg = node.block_arg
+      @double_splat = node.double_splat
+      @splat_index = node.splat_index
       @args_hash_initialized = false
 
       if !node.receiver && node.name == "initialize" && !current_type.is_a?(Program)
@@ -1115,6 +1134,9 @@ module Crystal
       @initialize_info = nil
       @block_arg = nil
       @args = nil
+      @double_splat = nil
+      @splat_index = nil
+      @splat = nil
       @args_hash.clear
       @args_hash_initialized = true
       @outside_def = true
@@ -1170,7 +1192,12 @@ module Crystal
 
     def args_hash
       unless @args_hash_initialized
-        @args.try &.each do |arg|
+        @args.try &.each_with_index do |arg, i|
+          @splat = arg if @splat_index == i
+          @args_hash[arg.name] = arg
+        end
+
+        @double_splat.try do |arg|
           @args_hash[arg.name] = arg
         end
 
--- a/src/compiler/crystal/semantic/type_guess_visitor.cr
+++ b/src/compiler/crystal/semantic/type_guess_visitor.cr
@@ -804,7 +804,7 @@ module Crystal
         # If the argument has a restriction, guess the type from it
         if restriction = arg.restriction
           # It is for something like `def foo(*@foo : *T)`.
-          if @splat == arg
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/semantic/type_guess_visitor.cr
```diff
@@ -800,6 +803,20 @@ module Crystal
       if arg = args_hash[node.name]?
         # If the argument has a restriction, guess the type from it
         if restriction = arg.restriction
+          # It is for something like `def foo(*@foo : *T)`.
+          if @splat == arg
```
asterite: I'd do `@splat.same?(arg)`

src/compiler/crystal/semantic/type_guess_visitor.cr
```diff
@@ -800,6 +803,20 @@ module Crystal
...
+            # If restriction is not splat (like `*foo : T`),
+            # we cannot guess the type.
+            # (We can also guess `Indexable(T)`, but it is not perfact.)
+            # And this early return is no problem because splat argument
+            # cannot have a default value.
+            return unless restriction.is_a?(Splat)
+            restriction = restriction.exp
+            # It is for something like `def foo(**@foo : **T)`.
```
asterite: Same here: `@double_splat.same?(arg)`

src/compiler/crystal/semantic/type_guess_visitor.cr
MakeNowJust: Thank you. I think so too. I'll fix it later.

bcardiff:
I'm finding some issues with this change.

Before this PR methods with var args methods where able to be restricted without problems:

```crystal
def concat(*args : String)
  args.join(", ")
end
```

This means `args[i]` is of type `String`. Same happen with `**args : String`.

I understand that the guess visitor was not handling `def initialize(*@args : String)` since there is not straight solution to type `@args` that will satisfy all possible invocations of that method.

I don't understand how this PR improves that situation.

From the added specs the following works

```crystal
class Foo
  def initialize(*@foo : *{Int32})
  end
end

Foo.new(1)
```

But `Foo.new(1, 2)` no longer compiles.
Making `def initialize(*foo : {Int32, Int32})` will make `Foo.new(1, 2)` pass, but not `Foo.new(1)`.
So with the proposed change the arg is no longer of variable length.

In the case of named argument I don't understand why would we need to mention a name in the restriction if the only thing we should express, at most, is the type of _all_ the values. (Look at `def initialize(**@foo : **{foo: Int32})` in the specs)

Since the problem is how instance vars can (or can't) be initialized for splatted argument I would propose:

* revert this PR
* disallow using ivar initializers in splatted arguments
* keep the `m(*args : String)` and `m(**args : String)` syntax as the only constructs that could appear for splatted arguments to express the types of all the values.

Unless of course, someone helps me see what I am misunderstanding here.

MakeNowJust:
@bcardiff What is problem? I didn't disable `m(*args : String)` and `m(**args : String)` syntaxes in this PR. They are still available.

asterite:
@bcardiff the `*{Int32}` syntax means it's a splat of a tuple of one element. That's why the type can be guessed.

(I don't know how useful is that, but that's how it works)

RX14:
@bcardiff 

```cr
class Foo
  def initialize(*@foo : *{Int32, String})
  end
end
```
means

```cr
class Foo
  @foo : {Int32, String}
  def initialize(*foo)
    @foo = foo
  end
end
```

This seems useless at first but is useful when you have `class Foo(*T)` to let the compiler guess the generic type correctly from the constructor:

```cr
class Foo(*T)
  def initialize(*@args : *T)
    p typeof(@args)
  end
end

Foo.new("name", 1, 6.0, "foo")
```

This PR doesn't change the semantics of the above code snippet at all. It just takes care to *allow* that while disallowing non-splat type restrictions. At least thats what I think from reading the code.

RX14:
BTW: another bug https://carc.in/#/r/5e8b :)

bcardiff:
Ok, I was missing that the use case was splatting over a type argument.

I still don't like it that much. I can't see the value of using splats in this scenario, and if splats are removed then we can pass and store directly the tuple.

Most likely is because I only think of restricting splats by the type of values (not length, not names).

Thanks for clarifying everybody.

RX14:
@bcardiff how would you achieve this interface without using splats??

bcardiff:
I think it can't.

The itch comes from the fact that the number of arguments changes in a very strict way inside a generic family.

Instantiated generics clases of course have incompatible interfaces if the `T` appears in the type restriction. But that usually does not predicate over the number or names of argument.

But it's an itch, I am not sure if breaks something formally.

asterite:
In general generic types over a splat are a big itch of me too. It's nice that it works but it's curious that no other language provides this and you have `Tuple1`, `Tuple2`, etc. in other languages. And I think the reason is this itch that there's something wrong (magical) about it.

PR #6649, opened 2018-09-02
Simplify windows rescue block codegen

### LINKED ISSUE #6605: break inside a rescue block doesn't work on Windows
I have a small sample of code here:

```crystal
a = [1, 2, 3, 4, 5]

loop do
  begin
    puts a[-1]
  rescue IndexError
    break
  end

  a.pop
end

puts "End result: #{a.size}"
```

This is the output on Unix systems (including WSL):

```
5
4
3
2
1
End result: 0
```

However, this is the result on Windows when cross-compiling (as per the instructions [here](https://github.com/crystal-lang/crystal/wiki/Porting-to-Windows)):

```
5
4
3
2
1
```

Notice the absence of the last line. I'm not particularly sure what the error is, but it might have to do with exception handling.

RX14:
Fixes #6605

### DIFF
--- a/src/compiler/crystal/codegen/exception.cr
+++ b/src/compiler/crystal/codegen/exception.cr
@@ -159,7 +159,8 @@ class Crystal::CodeGenVisitor
         # Here we set up the rescue block for that purpose.
         if node_ensure
           rescue_ensure_block = new_block "rescue_ensure"
-          @rescue_block = rescue_ensure_block
+        else
+          rescue_ensure_block = @rescue_block
         end
 
         node_rescues.each do |a_rescue|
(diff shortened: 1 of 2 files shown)


RX14:
Please can someone review this.

RX14:
@ysbaddaden I was largely expecting bcardiff to review it, since he reviewed the original PR.

src/compiler/crystal/codegen/exception.cr
```diff
@@ -182,6 +183,21 @@ class Crystal::CodeGenVisitor
...
+            catch_ret_target_block = new_block "this_rescue_target"
+            builder.build_catch_ret catch_pad, catch_ret_target_block
+            position_at_end catch_ret_target_block
+          end
+
+          saved_catch_pad = @catch_pad
+          @catch_pad = old_catch_pad
+
+          # We are generating code for rescues, so set up the rescue block.
+          @rescue_block = rescue_ensure_block
```
bcardiff: why there is a need to set the `rescue_block` in every codegen of the node_rescues instead of setting the `@rescue_block` only once as before?

RX14: `match_any_type_id` can generate a call, and in that case on windows, the call would be before we catchret, but it would be inside the `rescue_ensure_block` so would generate an invoke. This would mean we'd have invokes to the same catchswitch from inside and outside the catchpad which is disallowed. All this moving around is simply to prevent the call to `~match<Type>` from being inside the `rescue_ensure_block`.

PR #6659, opened 2018-09-04
Clean up tuple elements after unreachable

MakeNowJust:
Resolve https://github.com/crystal-lang/crystal/pull/6628#issuecomment-417981480 problem.

Small reproducible example:

```crystal
def foo
  yield({1, while true; end})
end

foo { |a, b| }
```

Result:

```console
$ crystal run foo.cr
Missing hash key: "b" (KeyError)
  from Hash(String, Crystal::CodeGenVisitor::LLVMVar)+@Hash(K, V)#[]<String>:Crystal::CodeGenVisitor::LLVMVar
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::CodeGenVisitor>:Nil
  from Crystal::CodeGenVisitor#visit<Crystal::Call>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::CodeGenVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::CodeGenVisitor>:Nil
  from main
```

This reason is `yield` tuple destructing against the tuple having unreachable elements. An unreachable element should be treated by `CleanupTransformer`. This PR adds cleanup process for `TupleLiteral` and `NamedTupleLiteral` to `CleanupTransformer`.

Thank you.

### DIFF
--- a/spec/compiler/codegen/def_spec.cr
+++ b/spec/compiler/codegen/def_spec.cr
@@ -559,7 +559,7 @@ describe "Code gen: def" do
       ))
   end
 
-  it "codegens yield with destructing tuple having unrachable element" do
+  it "codegens yield with destructing tuple having unreachable element" do
     codegen(%(
       def foo
         yield({1, while true; end})
--- a/spec/compiler/codegen/def_spec.cr
+++ b/spec/compiler/codegen/def_spec.cr
@@ -558,4 +558,14 @@ describe "Code gen: def" do
       foo(1 || true)
       ))
   end
+
+  it "codegens yield with destructing tuple having unrachable element" do
+    codegen(%(
+      def foo
+        yield({1, while true; end})
+      end
+
+      foo { |a, b| }
+      ))
+  end
 end
--- a/spec/compiler/semantic/cleanup_spec.cr
+++ b/spec/compiler/semantic/cleanup_spec.cr
@@ -16,6 +16,14 @@ describe "cleanup" do
       ), "expression has no effect"
   end
 
+  it "strip tuple elements after unreachable element" do
+    assert_after_cleanup "{1, while true; end, 2}", "1\nwhile true\nend"
+  end
+
+  it "strip named-tuple elements after unreachable element" do
+    assert_after_cleanup "{foo: 1, bar: while true; end, baz: 2}", "1\nwhile true\nend"
+  end
+
   # it "errors comparison of unsigned integer with zero or negative literal" do
   #   error = "comparison of unsigned integer with zero or negative literal will always be false"
   #   assert_error "1_u32 < 0", error
--- a/src/compiler/crystal/semantic/cleanup_transformer.cr
+++ b/src/compiler/crystal/semantic/cleanup_transformer.cr
@@ -730,6 +730,28 @@ module Crystal
     def transform(node : TupleLiteral)
       super
       node.update
+
+      no_return_index = node.elements.index &.no_returns?
+      if no_return_index
+        exps = Expressions.new(node.elements[0, no_return_index + 1])
+        exps.bind_to(exps.expressions.last)
+        return exps
+      end
+
+      node
+    end
+
+    def transform(node : NamedTupleLiteral)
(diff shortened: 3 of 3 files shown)


bew:
Sorry but I don't understand the problem and the solution here..

From #6628 I think you're using `while true; end` to get a `NoReturn` type, right? Can you add a note about this?

Also it starts as a `Tuple` and you transform it to an `Expressions` right? How come this is not a problem?? You're changing the type arbitrarily, what if I'm using the result as a `Tuple`??

(I don't understand the problem and the change..)

asterite:
Yeah, the fix doesn't seem to be right...

MakeNowJust:
`while true; end` is simple way to get `NoReturn` expression. (@asterite please tell me if you have simpler way.)

We cannot get the tuple having `NoReturn` element, so we can eliminate elements after unreachable. In fact, this kind optimization is applied to all other cases by `CleanupTransformer`. For example, I have this hash literal:

```crystal
{:foo => 1, :bar => while true; end, :baz => 2}
```

Finally, through `CleanupTransformer`, it is transformed to:

```crystal
__temp_xxx = ::Hash(typeof(:foo, :bar, :baz), typeof(1, while true; end, 2)).new
__temp_xxx[:foo] = 1
__temp_xxx
:bar
while true; end
```

I think the reason of https://github.com/crystal-lang/crystal/pull/6628#issuecomment-417981480 is missing to apply this optimization to `TupleLiteral` and it affects the codegen phase as a result.

spec/compiler/codegen/def_spec.cr
```diff
@@ -558,4 +558,14 @@ describe "Code gen: def" do
       foo(1 || true)
       ))
   end
+
+  it "codegens yield with destructing tuple having unrachable element" do
```
r00ster91: Typo: unrachable

MakeNowJust:
@asterite Could you review it with seriousness? I think it is the only true way to fix https://github.com/crystal-lang/crystal/pull/6628#issuecomment-417981480.

Do you expect we can fix it by avoiding `KeyError` with using `[]?`, like:

```diff
diff --git a/src/compiler/crystal/codegen/codegen.cr b/src/compiler/crystal/codegen/codegen.cr
index 490abfe77..6ddca53ae 100644
--- a/src/compiler/crystal/codegen/codegen.cr
+++ b/src/compiler/crystal/codegen/codegen.cr
@@ -1437,15 +1437,17 @@ module Crystal
               if arg && arg.name != "_"
                 t_type = tuple_type
                 t_value = codegen_tuple_indexer(exp_type, exp_value, i)
-                block_var = block_context.vars[arg.name]
-                assign block_var.pointer, block_var.type, t_type, t_value
+                if block_var = block_context.vars[arg.name]?
+                  assign block_var.pointer, block_var.type, t_type, t_value
+                end
               end
             end
           else
             exp_values.each_with_index do |(exp_value, exp_type), i|
               if (arg = block.args[i]?) && arg.name != "_"
-                block_var = block_context.vars[arg.name]
-                assign block_var.pointer, block_var.type, exp_type, exp_value
+                if block_var = block_context.vars[arg.name]?
+                  assign block_var.pointer, block_var.type, exp_type, exp_value
+                end
               end
             end
           end
```

But the above patch does not resolve anything.

```console
$ cat no-return-hash.cr
p({} of NoReturn => NoReturn)
$ bin/crystal run no-return-hash.cr
Error in src/hash.cr:857: BUG: missing context var: key

        key.inspect(io)
        ^~~
```

The true problem is `CleanupTransformer`'s bug, that is missing to transform `TupleLiteral` with `NoReturn` element. The codegen phase expects the program eliminated expressions after `NoReturn` expression. When `CleanupTransform` is not perfect, it makes sense that problems happen on the codegen phase.

asterite:
Now that I see it again it's almost OK. What happens if it's a tuple with one element that's NoReturn? I think it'll be a compiler bug with this code.

Issue #6624, opened 2018-08-30
0.26.1 i386 deb package completely broke the compiler


stronny:
```
$ dpkg -S /usr/bin/crystal
crystal: /usr/bin/crystal
$ dpkg -l | grep crystal
ii  crystal     0.26.1-1       i386         no description given
$ /usr/bin/crystal -h
Failed to raise an exception: END_OF_STACK
[0xb5ca8813] ???
[0xb535e383] ???
[0xb535f411] ???
[0xb53645ec] ???
[0xb539c469] ???
[0xb5362991] ???
[0xb722b816] ???
```

as a side note, maybe add a description to the package?

bcardiff:
@stronny I am checking the installer using a vagrant [xenial 32 bits](https://app.vagrantup.com/ubuntu/boxes/xenial32) and I was able to run the installer as documented [here](https://crystal-lang.org/docs/installation/on_debian_and_ubuntu.html) without troubles.

```
vagrant@ubuntu-xenial:~$ dpkg -S /usr/bin/crystal
crystal: /usr/bin/crystal
vagrant@ubuntu-xenial:~$ dpkg -l | grep crystal
ii  crystal                          0.26.1-1                                   i386         no description given
vagrant@ubuntu-xenial:~$ /usr/bin/crystal -h
Usage: crystal [command] [switches] [program file] [--] [arguments]

Command:
   ....
```

stronny:
My setup is a debian stable virtualbox guest. I've just noticed that the bug is reproduced only as a particular user, but not as another one. I think it has to do with fd reopening in the update, here's the tail of an strace:
```
[pid 12842] clock_gettime(CLOCK_REALTIME, {tv_sec=1535657416, tv_nsec=132236618}) = 0
[pid 12842] epoll_create1(EPOLL_CLOEXEC) = 4
[pid 12842] pipe2([5, 6], O_NONBLOCK|O_CLOEXEC) = 0
[pid 12842] getcpu(0x9afd8, 0x10, NULL) = -1 EFAULT (Bad address)
[pid 12842] open("/dev/urandom", O_RDONLY|O_LARGEFILE|O_CLOEXEC) = 7
[pid 12842] fcntl64(7, F_SETFD, FD_CLOEXEC) = 0
[pid 12842] fstat64(7, {st_mode=S_IFCHR|0666, st_rdev=makedev(1, 9), ...}) = 0
[...]

PR #6660, opened 2018-09-04
Use blocking IO on a TTY if it can't be reopened

Timbus:
Should fix issue #6624

Kinda hard to test though. Doesn't seem to have any regressions at least.

### DIFF
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -20,19 +20,18 @@ class IO::FileDescriptor < IO
 
   # :nodoc:
   def self.from_stdio(fd)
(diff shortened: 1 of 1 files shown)


RX14:
Probably cleaner to do

```cr
ret = LibC.ttyname_r(fd, path, 256)
return new(fd: blocking: true) unless ret == 0

clone_fd = LibC.open(path, LibC::O_RDWR)
return new(fd: blocking: true) if clone_fd == -1

new(clone_fd).tap(&.close_on_exec = true)
```

Sija:
@RX14 Your example calls are missing `fd` / `clone_fd` argument.

Issue #5374, opened 2017-12-11
Wishlist: Bring back Time#ticks or equivalent


docelic:
It would be good to bring back Time#ticks that existed in 0.23.

It doesn't need to be exactly #ticks, it can be any equivalent which reduces Time or Span to a single number that can then be used for comparisons and in initializers.

to_time.ticks <=> other.to_time.ticks

Time.new( Time.now.ticks +- Int)

Time::Span.new( ticks)

etc.

ysbaddaden:
Well: `to_time <=> other.to_time`, `Time.now - 1.second`. You don't need ticks.

If you really insist, there is `.epoch` and `.epoch_ms` as well as `to_f`.

docelic:
In 0.24.0 the Time class has nanosecond precision, so it is not acceptable to use #epoch or #epoch_ms. (There is also epoch_f, but that loses precision.)

And one needs both seconds and nanoseconds value to re-create an identical (or at least equally precise) Time object.

ysbaddaden:
Sorry, `ticks` was a private implementation detail, and we won't bring it back, since we don't use it anymore.

I believe Time has solutions for the exposed use cases, such as comparing Time instances themselves (which compares seconds and nanoseconds precisely), or casting to desired precisions. These solutions should be more future prone, for example to the addition of locations and offsets.

straight-shoota:
> It doesn't need to be exactly #ticks, it can be any equivalent which reduces Time or Span to a single number that can then be used for comparisons and in initializers.

It is currently not even possible to express the full range of a time instance with nanosecond precision in a single primitive number because that would require more than 64 bits. Hence ´Time`uses `Int64` for `seconds` and `Int32` for `nanoseconds`.

However, it is possible to use `epoch` and `nanoseconds` together to read two numbers representing that time instance. However, it is not straightforward to recreate a `Time` instance from that, because `Time.epoch` does not have a nanosecond argument and for `Time.new` you'd have to manually add `UNIX_SECONDS`.
It might be worth thinking about exposing `total_seconds` publicly which allows to read the `seconds` ivar directly. It is not so much an implementation detail as `ticks` and will always be needed for `epoch` which is basically just an alias wich subtracts `UNIX_SECONDS`.

straight-shoota:
To clarify: I'm with @ysbaddaden in arguing that you shouldn't *need* to convert a `Time` instance to primitive and back.

PR #6661, opened 2018-09-04
Expose Time#total_seconds in public API

straight-shoota:
`Time#total_seconds` was a protected method, but it can be useful outside of `Time` namespace, for example to implement custom extensions to the Time API.
This method directly exposes an implementation detail, but it doesn't depend on the implementation and would be useful even if the implementation was changed. It supplements the already existing constructor `Time.new(seconds, nanoseconds)` which accepts `total_seconds` since epoch as argument.

I don't think `Time#offset_seconds` needs to be exposed since it can easily be duplicated as sum of `#total_seconds` and `#offset`. Instead I changed visibility to `private` because it doesn't need to be exposed in the namespace (and isn't even used).

See also #5374 (probably fixes that).

### DIFF
--- a/spec/std/time/time_spec.cr
+++ b/spec/std/time/time_spec.cr
@@ -21,7 +21,7 @@ describe Time do
     end
 
     it "initializes max value" do
-      time = Time.new(9999, 12, 31, 23, 59, 59, nanosecond: 999_999_999)
+      time = Time.new(9999, 12, 31, 23, 59, 59, nanosecond: 999_999_999, location: Time::Location::UTC)
       time.year.should eq(9999)
       time.month.should eq(12)
       time.day.should eq(31)
@@ -29,6 +29,7 @@ describe Time do
       time.minute.should eq(59)
       time.second.should eq(59)
       time.nanosecond.should eq(999_999_999)
+      time.total_seconds.should eq Time::MAX_SECONDS
     end
 
     it "fails with negative nanosecond" do
@@ -331,6 +332,17 @@ describe Time do
       t1 = Time.now
       t1.epoch.should eq(t1.to_utc.epoch)
       t1.epoch_f.should be_close(t1.to_utc.epoch_f, 1e-01)
+    end
+  end
+
+  it "#total_seconds" do
+    Time.utc(1, 1, 1).total_seconds.should eq 0
+
+    now = Time.utc_now
+    Time.utc(seconds: now.total_seconds, nanoseconds: now.nanosecond).should eq now
+
+    expect_raises(ArgumentError, "Invalid time") do
+      Time.utc(seconds: Int64::MAX, nanoseconds: 999_999_999)
     end
   end
 
--- a/src/time.cr
+++ b/src/time.cr
@@ -1208,11 +1208,14 @@ struct Time
     (day - 1) + temp + (365*(year - 1)) + ((year - 1)/4) - ((year - 1)/100) + ((year - 1)/400)
   end
 
-  protected def total_seconds
+  # Returns the total number of seconds elapsed since epoch (`0001-01-01 00:00:00`).
+  def total_seconds
     @seconds
   end
 
-  protected def offset_seconds
+  # Returns the total number of seconds elapsed since epoch (`0001-01-01 00:00:00`)
+  # plus the offset observed in `#location` at that instant.
+  private def offset_seconds
     @seconds + offset
   end
 


ysbaddaden:
Unless you have clear use cases, this method is private API that reflect internal implementation details of Time and Time::Span.

`Time.new(sec, nsec)` should be marked with nodoc for the same reason —I believe it used to be. Maybe some other methods should be, too.

ysbaddaden:
Actually, I believe the `#total_seconds` and `#offset_seconds` don't even have to exist. Just use `@seconds` and `@seconds + @offset` where needed. They're only used within `Time` itself and seem more confusing than really helpful.

straight-shoota:
Use case: custom extensions to the Time API need a direct way to extract essential information out of a `Time` instance.

If `#total_seconds` and the supplementing constructor didn't exist, this would only be possible through the unix timestamp or the calendrical interface. Both are not optimal and require additional calculations which might be completely useless.

Yes, it directly exposes the internal representation, but that isn't bad by itself. `seconds since epoch` is a common property of any time representation, and it would make sense to have this method even if the internal implementation was different (for example using fields).

I agree that `#offset_seconds` is not really helpful.

RX14:
Who says we don't change the epoch later? And then because this is public API, `#total_seconds` won't be useful for anything. I guess since the community decided `foo.@var` was a good feature you can just use that.

straight-shoota:
> Who says we don't change the epoch later?

It's the epoch of the ISO 8601 calendar. We can change the internal representation (including epoch) however we like, but the calendar epoch won't change unless we change the calendar.
And in that case, there are more breaking changes than `#total_seconds`.

`@seconds` is exactly the thing we'd like to avoid: having to rely on internal implementation.

With the current implementation, `#total_seconds` just returns `@seconds`. But imagine the case where we'd change the internal implementation, for example to use Unix epoch instead. Then `#total_seconds` would just change to `@seconds - SECONDS_SINCE_UNIX` and the public API doesn't break.

ysbaddaden:
> I don't see any benefit from hiding these methods.

You don't expose any use case for them to be exposed either. Time extensions that would run slower isn't a use case. It's an hypothetical case.

> `@seconds` is exactly the thing we'd like to avoid: having to rely on internal implementation.

No, this is exactly what we want: this is an internal value, use at your own risk, because it *will* likely break someday at the most unexpected time. We don't even need `Time#total_seconds` in stdlib, since only `Time` uses it (and thus should use `@seconds`). We already had the issue with `#ticks` which completely broke when we changed the internal representation, and didn't care to fix. Let's not reiterate.

straight-shoota:
> Time extensions that would run slower isn't a use case. It's an hypothetical case.

They don't exist yet. But there are enough features missing from the stdlib implementation (and shouldn't be needed there) that sooner or later there will be third party extensions. You can call it hypothetical, but that's a very realistic use case.

Yes, `Time` can use `@seconds` directly. But an external library shouldn't have to. `#total_seconds` doesn't expose implementation details, it just happens that the implementation uses that same property "seconds since epoch" as internal representation.

Let's take a look at `#epoch`: Its value didn't change when the internal representation changed from ticks to seconds/nanoseconds. And it won't change if it changes back to ticks or something different. The same is true for `#total_seconds` - and similarly, the constructor `new(seconds, nanoseconds)`.

Described in different words: `#total_seconds` is like `#epoch` (or `#unix_seconds` when #6662 is merged) just with the native epoch of the ISO calendar (`0001-01-01`) instead of Unix epoch (`1970-01-01`). And `.new(seconds, nanoseconds)` is like `.epoch` (or `.unix`) but with calendar epoch and nanosecond resolution.

The only argument against `#total_seconds` is that there is already a very similar `#epoch`/`#unix_seconds` and we might not need both.
That's certainly up for debate.

I still think it makes sense to have `#total_seconds` because together with `new(seconds, nanoseconds)` it provides a smooth and consistent interface based on the inherent design of the underlying calendar and allows to convert a `Time` instance to a primitive datatype and back without losing precision (unix methods don't support nanoseconds).

asterite:
I'm not sure about this. Without a use case it's hard to approve this PR. I think converting a Time from/to unix seconds is good enough. I can't see why the numbers of seconds since 01-01-01 is important (given that we have a "standard" measure against the unix epoch).

straight-shoota:
If we add nanosecond precision to the UNIX constructor it might be fine.

straight-shoota:
Closing stale PR.

PR #6664, opened 2018-09-04
Add docs for file open modes

r00ster91:
This documents all file open modes and also improves the invalid file open mode error output a bit.

### DIFF
--- a/src/crystal/system/file.cr
+++ b/src/crystal/system/file.cr
@@ -4,7 +4,7 @@ module Crystal::System::File
   # calls.
   private def self.open_flag(mode)
     if mode.size == 0
-      raise "Invalid access mode #{mode}"
+      raise "Invalid file open mode '#{mode}'"
     end
 
     m = 0
@@ -19,7 +19,7 @@ module Crystal::System::File
       m = LibC::O_WRONLY
       o = LibC::O_CREAT | LibC::O_APPEND
     else
-      raise "Invalid access mode #{mode}"
+      raise "Invalid file open mode '#{mode}'"
     end
 
     case mode.size
@@ -32,10 +32,10 @@ module Crystal::System::File
       when 'b'
         # Nothing
       else
-        raise "Invalid access mode #{mode}"
+        raise "Invalid file open mode '#{mode}'"
       end
     else
-      raise "Invalid access mode #{mode}"
+      raise "Invalid file open mode '#{mode}'"
     end
 
     oflag = m | o
--- a/src/crystal/system/file.cr
+++ b/src/crystal/system/file.cr
@@ -4,7 +4,7 @@ module Crystal::System::File
   # calls.
   private def self.open_flag(mode)
     if mode.size == 0
-      raise "Invalid file open mode '#{mode}'"
+      raise "No file open mode specified"
     end
 
     m = 0
@@ -19,7 +19,7 @@ module Crystal::System::File
       m = LibC::O_WRONLY
       o = LibC::O_CREAT | LibC::O_APPEND
     else
-      raise "Invalid file open mode '#{mode}'"
+      raise "Invalid file open mode: '#{mode}'"
     end
 
     case mode.size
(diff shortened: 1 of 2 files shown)


j8r:
Other exceptions have their subject and the value separated by a semicolon, we should stick with this: `"Invalid file open mode: #{mode}"`

src/crystal/system/file.cr
```diff
@@ -4,7 +4,7 @@ module Crystal::System::File
   # calls.
   private def self.open_flag(mode)
     if mode.size == 0
-      raise "Invalid access mode #{mode}"
+      raise "Invalid file open mode '#{mode}'"
```
straight-shoota: In this branch, `mode` is an empty string. So this should probably better complain about that. At least it doesn't need to expand the variable.

src/file.cr
```diff
@@ -36,6 +36,27 @@ class File < IO::FileDescriptor
...
+  #      | creates a new file if the file doesn't exists.
+  # a    | Write-only, starts at the end of the file,
+  #      | creates a new file if the file doesn't exists.
+  # a+   | Read-write, starts at the end of the file,
+  #      | creates a new file if the file doesn't exists.
+  # rb   | Same as the 'r' mode but in binary file mode.
+  # wb   | Same as the 'w' mode but in binary file mode.
+  # ab   | Same as the 'a' mode but in binary file mode.
+  # ```
+  # In binary file mode, line endings are not converted to CRLF on Windows.
```
straight-shoota: Binary mode treats files as binary data. Convertions of line endings is just an example of that (although it's probably the sole significant consequence).

straight-shoota:
Is there a specific reason to change all `access mode` to `file open mode`? I don't see a significant difference.

j8r:
"file open mode" makes easier to figure out that a `File.open` method may be the issue in our code.

Sija:
`access mode` is more commonly used term than `file open mode`.

r00ster91:
I'd rather like to stick with "file open mode". I think this is more descriptive and meaningful than just "access mode" and it's already used everywhere else.

straight-shoota:
> it's already used everywhere else.

What do you mean with everywhere? I can't find a single occurence in the API docs.

Sija:
@r00ster91 More descriptive - perhaps, widely used - certainly not - see https://trends.google.com/trends/explore?q=%22access%20mode%22,%22file%20open%20mode%22

straight-shoota:
@Sija Are you aware that `access mode` has multiple other meanings? As such is very likely to be more frequent in google trends.

I don't see any indication that one of both terms is decisively more popular than the other. Semantically there is no big difference either, I think both are very descriptive. And that's why both are commonly used for documentation purposes.

Unless there are any objective arguments not mentioned here, I'd suggest to stick with the status quo.

r00ster91:
> What do you mean with everywhere? I can't find a single occurence in the API docs.

Well ok without this PR, there's one occurence in the [comment for the private `self.open_flag` method](https://github.com/crystal-lang/crystal/blob/0.26.1/src/crystal/system/file.cr#L3).

But anyway, when I search for "file access modes" then it leads me to pages like https://www.tutorialspoint.com/unix/unix-file-permission.htm with file permission stuff like `-rwxrwxr--` which is probably related to something different. While "file open modes" leads me to pages like http://www.c4learn.com/c-programming/c-file-open-modes/ and http://www.cplusplus.com/reference/cstdio/fopen/.

src/file.cr
```diff
@@ -36,6 +36,27 @@ class File < IO::FileDescriptor
...
+  #      | creates a new file if the file doesn't exists.
+  # w+   | Read-write, truncates existing file to zero length or
+  #      | creates a new file if the file doesn't exists.
+  # a    | Write-only, starts at the end of the file,
+  #      | creates a new file if the file doesn't exists.
+  # a+   | Read-write, starts at the end of the file,
+  #      | creates a new file if the file doesn't exists.
+  # rb   | Same as the 'r' mode but in binary file mode.
+  # wb   | Same as the 'w' mode but in binary file mode.
+  # ab   | Same as the 'a' mode but in binary file mode.
```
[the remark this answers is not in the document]
r00ster91: These file open modes don't exist actually. Do you mean they should be added? It should probably be done in a different PR.

RX14:
:O `FATAL: uncaught exception while processing handler for CHLD, exiting`

Looks like a rare bug that needs investigating: https://circleci.com/gh/crystal-lang/crystal/11842

straight-shoota:
@RX14 I found a second incident and opened #7243 for this.

PR #6665, opened 2018-09-05
Fix `ASTNode#to_s` for `{% verbatim do %} ... {% end %}`

MakeNowJust:
For example, it shows wrong result.

Code:

```crystal
p!({% verbatim do %}1{{2}}3{% end %})
```

Result:

```crystal
{% verbatim do %}1
{{ 2 }}
3
{% end %} # => 3
```

This PR fixes it.


MakeNowJust:
ping. Could anyone review it?

Issue #6669, opened 2018-09-05
Downto with unsigned integers


gmarcais:
The `#downto` method behaves like an infinite loop for an unsigned integer and a target value of 0. For example:

```Crystal
UInt16(3).downto(1) { |i| p i } # Good, prints 3, 2, 1
UInt16(3).downto(1).each { |i| p i } # Same with iterator
UInt16(3).downto(0) { |i| p i } # Bad, infinite loop.
UInt16(3).downto(0).each { |i| p i } # Same with iterator.
```
The first two lines print 3,2,1 as expected. The last two lines do not print 3,2,1,0 as expected, but rather loop back to `UInt16::MAX` and are infinite loops.

The current code for `#downto` in the class `Int` is:
```Crystal
def downto(to, &block : self ->) : Nil
    x = self
    while x >= to
      yield x
      x -= 1
    end
  end
```
This does not work for an unsigned integer: `x -= 1` loops back to the max positive value and is always `>= 0`. It could be changed to:
```Crystal
def downto(to, &block : self ->) : Nil
  return unless self >= to
  x = self
  yield x
  while x > to
    x -= 1
    yield x
  end
end
```

Similarly, the `DowntotoIterator` private to the class `Int` could be implemented as:
```Crystal
private class DowntoIterator(T, N)
  include Iterator(T)

  @from : T
  @to : N
  @current : T
  @done : Bool

  def initialize(@from : T, @to : N)
    @current = @from
    @done = false
  end

  def next
    return stop if @done
    value = @current
    @done = @current == @to
    @current -= 1 unless @done
    value
  end

  def rewind
    @current = @from
    @done = false
    self
  end
end
```

straight-shoota:
This error is actually not specific to unsigned integers but to all types when counting down to the min value. And similarly when counting up to max value.

Implementation wise it is probably easier to just return when `x == to` inside the loop:

```cr
def downto(to, &block : self ->) : Nil
    x = self
    while true
      yield x
      return if x == to
      x -= 1
   end
end
```

gmarcais:
One more issue: the current implementation of `#downto` and the proposed new implementation differ from the Ruby version when `to > self`. That is: `3.downto(5) { |i| p i }` prints 3 in Ruby, and nothing in Crystal. Should they agree? Same difference with `#upto`.

Ruby compatible implementation:
```Crystal
def downto(to, &block : self ->) : NIL
  yield self
  return unless self > to
    x = self
    while true
      x -= 1
[...]

PR #6678, opened 2018-09-07
Fix Int#downto with unsigned int going down to 0.

gmarcais:
See issue #6669. When using unsigned integers with the downto method, and the to is equal to 0, the code is an endless loop:

```Crystal
3.downto(0) { |i| p i } # Prints 3, 2, 1, 0
3_u16.downto(0) { |i| p i } # Endless loop printing 3_u16, 2_u16, 1_u16, 0_u16, 65535_u16, ...
```

Both version of the downto method (with and without a block) are fixed by this patch.

### DIFF
--- a/src/int.cr
+++ b/src/int.cr
@@ -353,9 +353,11 @@ struct Int
   end
 
   def upto(to, &block : self ->) : Nil
+    return unless self <= to
     x = self
-    while x <= to
+    while true
       yield x
+      return if x == to
       x += 1
     end
   end
@@ -365,9 +367,11 @@ struct Int
   end
 
   def downto(to, &block : self ->) : Nil
+    return unless self >= to
     x = self
-    while x >= to
+    while true
       yield x
+      return if x == to
       x -= 1
     end
   end
@@ -539,23 +543,24 @@ struct Int
     @from : T
     @to : N
     @current : T
+    @done : Bool
 
     def initialize(@from : T, @to : N)
       @current = @from
+      @done = !(@from <= @to)
     end
 
     def next
-      if @current > @to
-        stop
-      else
-        value = @current
-        @current += 1
-        value
-      end
+      return stop if @done
+      value = @current
+      @done = @current == @to
+      @current += 1 unless @done
+      value
     end
 
     def rewind
       @current = @from
+      @done = !(@from <= @to)
       self
     end
   end
@@ -566,23 +571,24 @@ struct Int
     @from : T
     @to : N
     @current : T
+    @done : Bool
 
     def initialize(@from : T, @to : N)
       @current = @from
+      @done = !(@from >= @to)
     end
 
     def next
-      if @current < @to
-        stop
-      else
-        value = @current
-        @current -= 1
-        value
-      end
+      return stop if @done
+      value = @current
+      @done = @current == @to
+      @current -= 1 unless @done
+      value
     end
 
     def rewind
       @current = @from
+      @done = !(@from >= @to)
       self
     end
   end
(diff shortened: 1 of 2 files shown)


src/int.cr
```diff
@@ -365,11 +365,13 @@ struct Int
...
   def downto(to, &block : self ->) : Nil
+    return unless self >= to
     x = self
-    while x >= to
+    while true
       yield x
+      return if x == to
       x -= 1
-    end
+   end
```
Sija: Wrong indent.

asterite:
What happens with `3_u16.downto(4_u16).to_a`?

gmarcais:
Hum. `3_u16.downto(4_u16).to_a` does not work. Let me fix the case when it goes up to return an "empty" iterator.

asterite:
@gmarcais You can probably do something like `@done = false unless @from >= @to` in the iterator's constructor.

gmarcais:
I modified to `@done = !(@from >= @to)` in the constructor and in rewind.

Sija:
@gmarcais Would be good to add a spec for a case mentioned by @asterite.

gmarcais:
Added spec cases for .to_a.

straight-shoota:
Please also fix `#upto` at into max (see https://github.com/crystal-lang/crystal/issues/6669#issuecomment-418900135)

gmarcais:
I don't get that error. It does not seem to have anything to do with the current patch:

```
 1) Semantic: macro allows declaring class with macro expression
       fork: Cannot allocate memory (Errno)
```

I changed the comment on the patch. Does this have anything to do with it? (Seems unlikely, but I ask nevertheless). Any idea what went wrong?

Sija:
> It does not seem to have anything to do with the current patch

That's correct, nothing to worry about.

straight-shoota:
@gmarcais Why close?

gmarcais:
I did not do that. Not sure what happened. Can it be reopened (I personally cannot apparently)? I was still considering fixing the step method and related.

gmarcais:
I added a version of the step method that does not overflow as well.

asterite:
@gmarcais You probably need to rebase your branch against `master`

gmarcais:
@asterite Yes, I rebased and added the step method. That allowed me to reopen the issue. 

I am not yet convinced about my implementation because of the following issue.

The step function is possibly a mess of many different types: `self, to, by`. They may all be different types. Say `self` is an `Int32` equal to `Int32::MAX-2` and `to` is a `UInt32` equal to `UInt32::MAX/2 + 2` (which is higher than `Int32::MAX`), what should the method do? Yield `Int32` values up to `Int32::MAX`? Yield `UInt32` values up to `to`?

One can imagine crazier combination: `self` is an `Int32`, `by` is a `Float` and `to` is a `Double` larger than `Int32::MAX`.

Should there be some type restrictions?

straight-shoota:
I don't think it would hurt to just restrict the arguments to the same type as `self`.

asterite:
Yeah, it's a crazy combination. Right now you can even do:

```crystal
1.step(to: 3.5, by: 0.5) do |i|
  puts i # => 1.0, 1.5, ...
end
```

I'm not sure restricting the types is the way to go, but we could try it.

But I'd try to not involve `step` in this PR because it will make it harder to decide when and how to merge this. Let's focus on `downto`, and we can fix `upto` and `step` in separate PRs (or, well, `upto` is probably fine in this PR).

RX14:
I'd cast to the `self` type using `self.class.new(to)`, like like the lhs type rule for operators.

gmarcais:
OK, I reverted the step method changes. The current PR only contains the upto and downto methods. I'll create a new PR for step.

PR #6679, opened 2018-09-07
Crystal::Macros::ClassDef API

Sija:
Adds missing methods to `Crystal::Macros::ClassDef` API.


Sija:
@asterite that's weird, I got impression of such properties from https://github.com/crystal-lang/crystal/blob/b6032455bf77684bf03d20ffb69cf64b94f16732/src/compiler/crystal/syntax/ast.cr#L1273-L1283

asterite:
@Sija those are the properties, yes, but they are exposed in the file ai mentioned.

PR #6681, opened 2018-09-08
Add support for ISO calendar week to Time

### LINKED ISSUE #2533: Add `Time#iso_year_and_week`
We need a `Time#iso_year_and_week : {Int32, Int32}` method that returns a tuple with the [iso week and date](https://en.wikipedia.org/wiki/ISO_week_date). Then, we can also have `Time#iso_week` and `Time#iso_year` that would return each component of that tuple. Then use the same time format flags as [Ruby's strftime](http://ruby-doc.org/core-2.2.0/Time.html#method-i-strftime) for iso week: 

```
%G - The week-based year
%g - The last 2 digits of the week-based year (00..99)
%V - Week number of the week-based year (01..53)
```

The problem is I can't find an official definition of algorithm for this. I _could_ copy [Go's source code](https://golang.org/src/time/time.go?s=10978:11018#L319) for this, which is pretty clear, or some other language's soruce code, but I don't know if we'll be violating some copyright. The thing is, the algorithm is an ISO standard and so I'm not sure if copying such code would be infringing something.

straight-shoota:
This PR adds two methods to `Time` for handling ISO calendar week:

* `Time#calendar_week` returns the year and week number. The year is important because calendar weeks don't match up with calendar years and the calendar week might actually be in a different year than the regular calendar year (returned by `#year`).
* `Time.week_date` creates a `Time` instance from a week date specified by year, week number and day of week.

It also adds directives `%G, %g, %V` which utilize these methods to `Time::Format`.

The algorithms for calculating the week number and back are completely annotated in code. They're based on the description on [Wikipedia](https://en.wikipedia.org/wiki/ISO_week_date#Calculation) and the implementation in [Go's stdlib](https://golang.org/src/time/time.go?s=18399:18439#L520) but with a few improvements.

Closes #2533 
Depends on #6555


RX14:
needs a rebase

straight-shoota:
Rebased.

bcardiff:
@straight-shoota Sorry, could you rebase again this on master so we can merge in a more tidy fashion?

straight-shoota:
Sure, done. But I'm not sure what it adds, it could just be fast forwarded.

bcardiff:
Thanks! A rebase will not leave a text reference to the PR. Squash or merge commit will. But is better to avoid long standing branch merged and keep the repo as thin as possible.

PR #3493, opened 2016-11-01
allow rand with zero value

kostya:
this is useful in such code:
```crystal
def delay(interval : Float64)
  sleep(rand(interval))
end

# when you want just not to sleep
delay(0.0)
```

PR #6686, opened 2018-09-08
Revert "allow rand from zero value"

oprypin:
This reverts #3493.

That commit introduced a bad change: it silenced a valid error for an arbitrary case.
Saying "this is useful" while showing wrong code is not an argument; that code should have just used an inclusive range.

`rand(n)` is always equivalent to `rand(0...n)`.

`rand(0...0)` even now says "Invalid range for rand: 0...0".

`rand(0)` asks for a number ≥ 0 and < 0. This makes no sense and should obviously be an error for the same reason.

### DIFF
--- a/spec/std/random_spec.cr
+++ b/spec/std/random_spec.cr
@@ -36,8 +36,6 @@ describe "Random" do
     5.times do
       rand(Int64::MAX).should be >= 0
     end
-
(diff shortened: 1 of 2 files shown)


kostya:
but why rand(0..0) is ok?

PR #6688, opened 2018-09-08
Make asm parsing and formatting more robust

RX14:
Hopefully the last asm PR, because the parsing and formatting code is really simple and uniform this time.

Best to look at the end result for this, the diff is full of false-positives.


RX14:
@asterite you could say that about removing empty `()` from function calls though - which we already do when the call has a reciever. Given the lack of complaints about the former, and how infrequently people write `asm`, I'm not inclined to change this.

asterite:
@RX14 sounds good

PR #6689, opened 2018-09-09
Fix `==` and `!=` when comparing signed vs unsigned integers

### LINKED ISSUE #571: The result of comparing signed and unsigned variables of the same size is inconsistent
Let's take a look at the following test program:

```
a = 4294967295_u32
b = 0_i32

puts "a = #{a}"
puts "b = #{b}"

if b > a
  puts "(b > a) is true"
else
  puts "(b > a) is false"
end

if a < b
  puts "(a < b) is true"
else
  puts "(a < b) is false"
end
```

The resulting output is

```
a = 4294967295
b = 0
(b > a) is true
(a < b) is false
```

The statements "(b > a) is true" and "(a < b) is false" are obviously contradicting each other.

C/C++ compilers usually issue a warning about mixed signed/unsigned comparisons even though the language standard defines deterministic (albeit somewhat convoluted) conversion rules:
https://www.securecoding.cert.org/confluence/display/cplusplus/INT02-CPP.+Understand+integer+conversion+rules

asterite:
Fixes #571

### DIFF
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -537,6 +537,18 @@ describe "Int" do
     end
   end
 
+  {% if compare_versions(Crystal::VERSION, "0.26.1") > 0 %}
+    it "compares equality and inequality of signed vs. unsigned integers" do
+      x = -1
+      y = x.unsafe_as(UInt32)
+
+      (x == y).should be_false
+      (y == x).should be_false
+      (x != y).should be_true
+      (y != x).should be_true
+    end
+  {% end %}
+
   it "clones" do
     [1_u8, 2_u16, 3_u32, 4_u64, 5_i8, 6_i16, 7_i32, 8_i64].each do |value|
       value.clone.should eq(value)
--- a/src/compiler/crystal/codegen/primitives.cr
+++ b/src/compiler/crystal/codegen/primitives.cr
@@ -118,6 +118,8 @@ class Crystal::CodeGenVisitor
     when "<=" then return codegen_binary_op_lte(t1, t2, p1, p2)
     when ">"  then return codegen_binary_op_gt(t1, t2, p1, p2)
     when ">=" then return codegen_binary_op_gte(t1, t2, p1, p2)
+    when "==" then return codegen_binary_op_eq(t1, t2, p1, p2)
+    when "!=" then return codegen_binary_op_ne(t1, t2, p1, p2)
     end
 
     p1, p2 = codegen_binary_extend_int(t1, t2, p1, p2)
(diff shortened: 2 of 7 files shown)


bcardiff:
Why closing it? I was up to send the approval.

asterite:
@bcardiff Because it stucks at one spec, probably this broke something that depended on that equality. I'll reopen and investigate why it's failing, and try to fix that.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -161,6 +161,26 @@ class Crystal::CodeGenVisitor
     end
   end
 
+  # The below methods (lt, lt2, gt, gte, eq, ne) perform
```
RX14: s/lt2/lte/

RX14:
Wow, you corrected the typo before I even had time to red the whole diff!

src/compiler/crystal/codegen/primitives.cr
asterite: Good catch. Fixed!

asterite:
I keep adding fixes. This time, getting runtime integer values from LLVM returned `UInt64` and we sometimes compared that with `-1`, so now it fails. I fixed it by making that used method to return `Int32`, while also allowing to get the values as `Int64` and `UInt64`,  which is what LLVM provides (but in the tests we don't use such big values so it's always safe to use `Int32`).

RX14:
CI still failing :C

asterite:
CI passed 🎉

spec/std/int_spec.cr
```diff
@@ -537,6 +537,18 @@ describe "Int" do
     end
   end
 
+  {% if compare_versions(Crystal::VERSION, "2.6.1") > 0 %}
```
luislavena: @asterite is this version correct? Shouldn't be `0.26.1` instead?

asterite: Oops! Fixed :-)

spec/std/int_spec.cr
```diff
@@ -537,6 +537,18 @@ describe "Int" do
...
 
+  {% if compare_versions(Crystal::VERSION, "0.26.1") > 0 %}
+    it "compares equality and inequality of signed vs. unsigned integers" do
+      x = -1
+      y = x.unsafe_as(UInt32)
+
+      (x == y).should be_false
+      (y == x).should be_false
+      (x != y).should be_true
+      (y != x).should be_true
```
j8r: Wouldn't it be better to use `should eq` and `should_not eq` expectations to have a report of variables if it fails?

asterite: I don't know, here we are testing those operators and I wouldn't like to hide them behind a spec matcher.

RX14: this is better as-is

NoICE:
You guys are awesome. Thank you.

PR #6694, opened 2018-09-10
Int: remove `gcd` and `lcm`

### LINKED ISSUE #6683: integer overflow in gcd function
The Crystal ``gcd`` implementation has a bug. I haven't found its code yet to see what it is as I just discovered it because it was returning errors in my code.

Ex: All these results should be '1'
```
puts 614_889_782_588_491_410.gcd(53) # => 7
puts 614_889_782_588_491_410.gcd(59) # => 95
puts 614_889_782_588_491_410.gcd(61) # => 31
```
To fix function in my code I implemented ``Stein's gcd algorithm``

https://en.wikipedia.org/wiki/Binary_GCD_algorithm
https://lemire.me/blog/2013/12/26/fastest-way-to-compute-the-greatest-common-divisor/

This is inside a module, inside of ``Struct Int`` so I can do:  ``a.gcd b`` like in Ruby.

```
    def bbgcd(b)
      a, b = self.to_big_i, b.to_big_i
      return b if a == 0
      return a if b == 0
    
      k = typeof(a).zero
      while (a | b).even?
        a >>= 1
        b >>= 1
        k += 1
      end
    
      while a.even?; a >>= 1 end
    
      while b != 0
        while b.even?; b >>= 1 end
        a, b = b, a if a > b
        b -= a
      end
    
      a << k
    end
```
Here are the correct results using it:
```
puts 614_889_782_588_491_410.bbgcd(53) => 1
puts 614_889_782_588_491_410.bbgcd(59) => 1
puts 614_889_782_588_491_410.bbgcd(61) => 1
```
This is another instance where I was pulling my hair trying to figure out why the same Ruby code was producing incorrect results in Crystal, until I became resigned to the fact that the code was correct and this function was returning incorrect results.  And sure enough it was for certain numbers.  Again, I just verified this, and  haven't had a chance to look at the Crystal implementation, but it may be better to use Stein's implementation.  Whether to use it was also a discussion issue for Ruby.

asterite:
Trivially fixes #6683

### DIFF
--- a/spec/std/big/big_int_spec.cr
+++ b/spec/std/big/big_int_spec.cr
@@ -279,8 +279,8 @@ describe "BigInt" do
     (-a_17).gcd(17).should eq(17)
     (-17).gcd(a_17).should eq(17)
 
-    (a_17).gcd(17).should be_a(Int::Unsigned)
-    (17).gcd(a_17).should be_a(Int::Unsigned)
+    (a_17).gcd(17).should be_a(BigInt)
+    (17).gcd(a_17).should be_a(BigInt)
 
     (a_17).lcm(17).should eq(a_17)
     (17).lcm(a_17).should eq(a_17)
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -100,14 +100,6 @@ describe "Int" do
     end
   end
 
-  describe "lcm" do
-    it { 2.lcm(2).should eq(2) }
-    it { 3.lcm(-7).should eq(21) }
-    it { 4.lcm(6).should eq(12) }
-    it { 0.lcm(2).should eq(0) }
-    it { 2.lcm(0).should eq(0) }
-  end
-
   describe "to_s in base" do
     it { 12.to_s(2).should eq("1100") }
     it { -12.to_s(2).should eq("-1100") }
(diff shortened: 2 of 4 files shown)


asterite:
The other change is because it's not intuitive to return Int from a BigInt operation (in fact it returned a union)

yxhuvud:
I'm not a fan of this particular change. This would add both code and performance overhead to virtually all usages of these methods I've ever done. I'd prefer an exception on `lcm` overflow (a la the overflow issue bcardiff is working on), and to have the bigint operation always return a bigint.

sdogruyol:
@asterite do you still want to work on this?

asterite:
I don't know

jhass:
Somebody experimented with the original issue and active overflow checking yet? Does it improve the situation? If so, maybe we can close this and leave the original issue open for further discussion.

yxhuvud:
@jhass I am planning to do it on my summer vacation, now that the overflow stuff is in place.

yxhuvud:
@jhass Yes, the original implementation will give overflow error with overflow checking turned on. However, it will overflow even in cases it doesn't have to. I updated the issue with some benchmarks and comments.

vlazar:
Can be closed since #6683 was fixed by #7996?

asterite:
@vlazar Thanks!

PR #5003, opened 2017-09-18
Change Errno into a hierarchy of OSError subclasses

oprypin:
Initial exception hierarchy is:

- OSError (former Errno)
  - OSError::BlockingIO {EAGAIN, EALREADY, EINPROGRESS, EWOULDBLOCK}
  - OSError::FileExists {EEXIST}
  - OSError::FileNotFound {ENOENT}
  - OSError::IsADirectory {EISDIR}
  - OSError::NotADirectory {ENOTDIR}
  - OSError::PermissionError {EACCES, EPERM}
  - ConnectionError
    - BrokenPipe {EPIPE}
    - ConnectionAborted {ECONNABORTED}
    - ConnectionRefused {ECONNREFUSED}
    - ConnectionReset {ECONNRESET}

These are based on [what Python does](https://docs.python.org/3/library/exceptions.html#exception-hierarchy).

The best showcase of improvement is [src/ecr/process.cr](https://github.com/crystal-lang/crystal/pull/5003/files#diff-5a095c79df5efe1d16c00800706fdfa5).

Note that I did not get rid of the inherent dependency on `errno`/`GetLastError`, because it's just not viable to predict every possible error from those countless error codes, every time an OS API is called. `OSError.create` substitutes the exception class with some subclass of `OSError`, when such a class is defined, or otherwise the generic `OSError` (just like `Errno` before) with the error code available. This is also [what Python does](https://docs.python.org/3/library/exceptions.html#OSError).

Windows errors are integrated into this; see [windows_error.cr](https://github.com/crystal-lang/crystal/pull/5003/files#diff-0344db3a341eb107ea95f3a804c4bd6a).

PR #6698, opened 2018-09-11
Implement Windows Registry API

straight-shoota:
This PR adds an API for accessing the Windows Registry.

Access to the Registry is required in the standard library to get proper time zone names (see `Crystal::System::Time.translate_zone_name`, included in this PR) and MIME type mappings (extension to #5765, not yet included).

However, the Registry API is obviously only available on win32. Previous discussions (most recently in #6668) have manifested a common understanding that the Crystal standard library should be as much cross-platform as possible.
Having a highly platform-specific Registry API conflicts that proposition. Still, the stdlib needs that API to provide an implementation on win32 for cross-platform features.

I considered stripping down the implementation and moving it to the `Crystal::System` namespace. This would only contain the bare minimum required to implement the missing features for accessing the time and mime databases in the registry. But they already require a huge portion of the features and thus doesn't seem to make much sense as a stripped-down version.

Looking at Go, they have the registry API in an external package [`golang.org/x/sys/windows/registry`](https://github.com/golang/sys/tree/master/windows/registry) which needs to be imported to use it in user code. But, they also need the registry for the same stdlib implementations and thus provide a copy of that package in [`internal/syscall/windows/registry`](https://github.com/golang/go/tree/master/src/internal/syscall/windows/registry) which is only used by the internal implementation.
We could probably go a similar way, but I'm not too happy about maintaining the same code in two repositories. 
Maybe it could be somewhat manageable using git subtrees or submodules, but I don't have a clear strategy in mind.

So, before reviewing the implementation of this PR, please consider options for how to include this into the stdlib.


src/registry.cr
```diff
@@ -0,0 +1,588 @@
...
+#
+# The Windows API defines some [predefined root keys](https://docs.microsoft.com/en-us/windows/desktop/sysinfo/predefined-keys) that are always open.
+# They are available as constants and can be used as entry points to the registry.
+#
+# * `HKEY_CLASSES_ROOT`
+# * `HKEY_CURRENT_USER`
+# * `HKEY_LOCAL_MACHINE`
+# * `HKEY_USERS`
+# * `HKEY_CURRENT_CONFIG`
+module Registry
```
asterite: This should be in a `Windows` namespace.

src/crystal/system/win32/time.cr
```diff
@@ -163,7 +164,23 @@ module Crystal::System::Time
   # Searches the registry for an English name of a time zone named *stdname* or *dstname*
   # and returns the English name.
   private def self.translate_zone_name(stdname, dstname)
-    # TODO: Needs implementation once there is access to the registry.
-    nil
+    Registry::LOCAL_MACHINE.open("SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Time Zones",
+      Registry::SAM::ENUMERATE_SUB_KEYS | Registry::SAM::QUERY_VALUE) do |key|
+      key.each_key do |sub_name|
+        key.open(sub_name, Registry::SAM::READ) do |sub_key|
```
asterite: I _think_ you can write it as `key.open(sub_name, :read)`, right?

src/registry.cr
```diff
@@ -0,0 +1,588 @@
...
+      name.check_no_null_byte
+
+      get_raw?(name.to_utf16, buffer)
+    end
+
+    private def get_raw?(name : Slice(UInt16), buffer : Bytes) : {ValueType, UInt32}?
+      length = buffer.size.to_u32
+      status = LibC.RegQueryValueExW(self, name, nil, out valtype, buffer, pointerof(length))
+      case status
+      when WinError::ERROR_SUCCESS
```
asterite: I just now see this, but this should also probably be named `Windows::Error`. Basically anything that's windows specific must be in the `Windows` module.

straight-shoota: `ERROR_SUCCESS` should be a `LibC` constant. And `WinError` should eventually be entirely replaced by specific cross-platform error types, same as `Errno` (similar to what was proposed in #5003).

asterite: Right, that makes sense.

src/crystal/system/win32/time.cr
asterite: (no need to change it, though, I was just curious)

straight-shoota: I think I had actually implemented that part before #6074 was merged.

Looking at it, I figured the argument to the previous call didn't need to be a combination of *enumerate* and *query*. Changing this all to `:read` makes the code a little bit simpler.

src/lib_c/x86_64-windows-msvc/c/winreg.cr
```diff
@@ -0,0 +1,114 @@
...
+    # This flag must be combined using the OR operator with the other flags in this table that either query or access registry values.
+    # Windows 2000: This flag is not supported.
+    WOW64_32KEY = 0x00200
+
+    # Indicates that an application on 64-bit Windows should operate on the 64-bit registry view. This flag is ignored by 32-bit Windows.
+    # This flag must be combined using the OR operator with the other flags in this table that either query or access registry values.
+    # Windows 2000: This flag is not supported.
+    WOW64_64KEY = 0x00100
+
+    # Combines the STANDARD_RIGHTS_WRITE, `KEY_SET_VALUE`, and `KEY_CREATE_SUB_KEY` access rights.
```
r00ster91: ` missing around `STANDARD_RIGHTS_WRITE

src/registry.cr
```diff
@@ -0,0 +1,588 @@
...
+      end
+
+      getter sub_key_count : UInt32
+
+      # size of the key's subkey with the longest name, in Unicode characters, not including the terminating null byte.
+      getter max_sub_key_length : UInt32
+
+      getter value_count : UInt32
+      # size of the key's longest value name, in Unicode characters, not including the terminating null byte.
+
```
r00ster91: Newline needs to be removed

straight-shoota:
> it should be possible to do this using a few raw libc calls.

Yes, it should be possible. But including all error handling etc. around these libc calls, such a bare-bone implementation is far from trivial and would include most of the functionality proposed in this PR. Thus it seems much better to have a full-fledged and fully tested implementation than an ugly purpose-specific one.

RX14:
Whatever option gets chosen, I don't want a registry or anything 100% platform-specific exposed directly to the user.

ysbaddaden:
I'm fine with a `Windows::Registry` class only available for Windows. Ruby has `Win32::Registry` in stdlib, Python has `_winreg` in stdlib, Go has an official package `golang.org/x/sys/windows/registry` copied as `src/internal/windows/registry` in stdlib.

If we only need to read from the registry to extract data in core/stdlib, then maybe we could have a stripped down `Crystal::System::Registry` with only read methods.

Anyway, having a properly implemented and tested Registry class is a better idea than shelling out to `reg.exe` or wrapping C calls over and over (increasing bugs), whatever if it's private or public.

RX14:
@ysbaddaden i'm fine with whatever as long as it's in `Crystal::System` and undocumented, but I think it's certainly possible to have a "nice enough" interface that's very mininal: no structs, no OO, just calling methods that return handles/strings and raise. This also means minimal effect on the compile time of small programs on windows.

bew:
> i'm fine with whatever as long as it's in Crystal::System and undocumented

Or documented for us, but not public (please!)

straight-shoota:
> I don't want a registry or anything 100% platform-specific exposed directly to the user.

What about `UNIXSocket`? 🤷‍♂️

asterite:
I'm fine with having a full binding to the windows registry in the std. It's not much code either, just 600 lines of code which include comments.

RX14:
@straight-shoota https://docs.microsoft.com/en-us/windows/desktop/ipc/named-pipes can substitute for unix sockets most of the time

ysbaddaden:
1. [AF_UNIX is coming to windows](https://blogs.msdn.microsoft.com/commandline/2017/12/19/af_unix-comes-to-windows/);
2. don't "abstract" something as fundamental as UNIX sockets under some shady new term just because "it's not cross platform"; that would just bring confusion;
3. let's get back on topic: the windows registry.

RX14:
@ysbaddaden thanks for the info! Nice to hear that supporting real unix sockets on windows is possible. However, that further underlines my point that bringing registry into the stdlib, *let alone the prelude* would be unprecedented and unnecessary.

src/lib_c/x86_64-windows-msvc/c/winreg.cr
```diff
@@ -0,0 +1,114 @@
...
+  end
+
+  enum ValueType
+    NONE                       = 0
+    SZ                         = 1
+    EXPAND_SZ                  = 2
+    BINARY                     = 3
+    DWORD                      = 4
+    DWORD_LITTLE_ENDIAN        = DWORD
+    DWORD_BIG_ENDIAN           =  5
```
Fryguy: minor, but weird alignment here (though it's probably `crystal tool format`)... the `4` is aligned different from the `5`, likely because the DWORD breaks up the 2 sequences.

src/registry.cr
```diff
@@ -0,0 +1,588 @@
...
+# They are available as constants and can be used as entry points to the registry.
+#
+# * `HKEY_CLASSES_ROOT`
+# * `HKEY_CURRENT_USER`
+# * `HKEY_LOCAL_MACHINE`
+# * `HKEY_USERS`
+# * `HKEY_CURRENT_CONFIG`
+module Registry
+  # Information about file types and their properties.
+  CLASSES_ROOT = Key.new(LibC::HKEY_CLASSES_ROOT, "HKEY_CLASSES_ROOT")
```
Fryguy: I'm curious why you removed the HKEY_ prefix on the const name?  As a user of this module I'd expect to find it with the HKEY_ prefix

straight-shoota: Because they're redundant. There are no namespaces in C, hence constants are often prefixed by a type identifier (like `HKEY`). In Crystal it's obvious that they are registry constants because they're defined in the `Registry` module.

Fryguy: I can see your argument, but outside of the C-code context, the HKEY prefix has pervaded into everything.  It's even in the GUI regedit to the point that that is the actual name of the key now.  That was really the though process behind my statement of "As a user of this module I'd expect to find it with the HKEY_ prefix"

straight-shoota:
Closing.
The immediate functionality for time zone and mime db support on Windows was resolved by #11137.

If we decide to provide a public API for access to the Windows Registry in stdlib later, we can re-issue this implementation. Otherwise it may serve as a basis for a shard.

PR #6702, opened 2018-09-11
Add Errno#errno_message getter

r00ster91:
This adds `Errno#errno_message` getter which returns the last raised errno message.

Usecase:
I want to print the error message from errno in my application just like this: `No such file or directory` instead of this: `Unhandled exception: some_call: No such file or directory (Errno)`.

### DIFF
--- a/src/errno.cr
+++ b/src/errno.cr
@@ -204,6 +204,9 @@ class Errno < Exception
   # Returns the numeric value of errno.
   getter errno : Int32
 
+  # Returns the message of errno.
+  getter errno_message : String
+
   # Creates a new `Errno` with the given message. The message will
   # have concatenated the errno message denoted by *errno*.
   #
@@ -217,12 +220,8 @@ class Errno < Exception
   # ```
   def initialize(message, errno = Errno.value)
     @errno = errno
-    super "#{message}: #{Errno.message(errno)}"
-  end
-
-  # Returns the errno message denoted by *errno*.
-  def self.message(errno)
-    String.new(LibC.strerror(errno))
+    @errno_message = String.new(LibC.strerror(@errno))
+    super "#{message}: #{@errno_message}"
   end
 
   # Returns the value of libc's errno.
--- a/src/errno.cr
+++ b/src/errno.cr
@@ -205,7 +205,7 @@ class Errno < Exception
   getter errno : Int32
 
   # Creates a new `Errno` with the given message. The message will
-  # have concatenated the message denoted by *errno*.
+  # have concatenated the description denoted by *errno*.
   #
   # Typical usage:
   #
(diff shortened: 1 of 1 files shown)


src/errno.cr
```diff
@@ -217,7 +217,12 @@ class Errno < Exception
   # ```
   def initialize(message, errno = Errno.value)
     @errno = errno
-    super "#{message}: #{String.new(LibC.strerror(errno))}"
+    super "#{message}: #{description(errno)}"
+  end
+
+  # Returns the description denoted by *errno*.
+  def description(errno = Errno.value)
```
straight-shoota: This should either be a class method or use `@errno` instead of an argument.

RX14:
Perhaps this should be a method instead of a class method

asterite:
Once we make the error hierarchy independent of the OS, will this still be here?

I don't quite understand the use case.

src/errno.cr
```diff
@@ -204,8 +204,8 @@ class Errno < Exception
   # Returns the numeric value of errno.
   getter errno : Int32
 
-  # Creates a new Errno with the given message. The message will
-  # have concatenated the message denoted by `Errno#value`.
+  # Creates a new `Errno` with the given message. The message will
+  # have concatenated the errno message denoted by *errno*.
```
Sija: What's `errno message`? It should be `*message*`.

r00ster91:
@RX14 In the end there isn't really a difference and the other methods are also already class methods.
But if we make self.message an instance method, the others should be instance methods too I think.

@asterite The use case is just that when I for example read a file and an error happens, that I can print the reason why the file couldn't be read.
I could right now for example do:
```crystal
begin
  File.read("a")
rescue ex : Errno
  puts ex
end
```
and it would print `Error opening file 'a' with mode 'r': No such file or directory`. But I just want to print the errno message `No such file or directory` in my program.

asterite:
@r00ster91 then let's save `@message` in an instance variable and expose that as a `getter`.

asterite:
Hm, never mind, the name collides with `Exception`'s message. Maybe `@errno_message`, though...

r00ster91:
@asterite Hmm so we always save the last raised errno message in `@errno_message`. But what if I for example want to get another errno message instead of just the last one that has been raised? I think we should rather use a method.

asterite:
> But what if I for example want to get some other errno message with a number instead of just the last one that has been raised?

What would be an example for this? Errno gets re-written each time there's an error, so it only makes sense to reference the last one.

With that you could do:

```crystal
begin
  File.read("a")
rescue ex : Errno
  puts ex.errno_message
end
```

But more specifically, why don't you do:

```crystal
begin
  File.read("a")
rescue ex
  puts "Can't read file a"
end
```

I don't think `Errno` as an exception will survive in the future, we'll probably have a different exception hierarchy and exposing errno message will probably don't exist.

straight-shoota:
> But what if I for example want to get some other errno message with a number

You could always just call `LibC.error` directly.

r00ster91:
> But more specifically, why don't you do:
> ```
> begin
>  File.read("a")
> rescue ex
>  puts "Can't read file a"
> end
I prefer to give the user specific information *why* the program couldn't read the file. Because just this is IMO not very helpful to solve the problem.

But okay I will change it to a getter then. I agree. In the end when really need it, I could just call `LibC.strerror` myself.

PR #6704, opened 2018-09-11
Add Socket::Address#pretty_print and #inspect

straight-shoota:


asterite:
Why? How does it look if you print an array of addresses?

straight-shoota:
```cr
p [Socket::IPAddress.new("127.0.0.1", 80), Socket::IPAddress.new("192.168.0.1", 8000)]
#before:
# => [Socket::IPAddress(@family=INET, @size=16, @port=80, @address="127.0.0.1", @addr6=nil, @addr4=LibC::InAddr(@s_addr=16777343_u32)), Socket::IPAddress(@family=INET, @size=16, @port=8000, @address="192.168.0.1", @addr6=nil, @addr4=LibC::InAddr(@s_addr=16820416_u32))]

# after:
# => [127.0.0.1:80, 192.168.0.1:8000]
```

I don't think there is any reason for `#inspect` to print all the internal values when the string representation displays exactly the same information in a human readable format.

asterite:
Here's Ruby:

```
reqirb(main):001:0> require 'ipaddr'
=> true
irb(main):002:0> ipaddr1 = IPAddr.new "3ffe:505:2::1"
=> #<IPAddr: IPv6:3ffe:0505:0002:0000:0000:0000:0000:0001/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff>
irb(main):003:0> ipaddr3 = IPAddr.new "192.168.2.0/24"
=> #<IPAddr: IPv4:192.168.2.0/255.255.255.0>
irb(main):004:0> ipaddr3 = IPAddr.new "192.168.2.0"
=> #<IPAddr: IPv4:192.168.2.0/255.255.255.255>
```

As I mentioned in some other PR, `inspect` should produce an output that's clearly delimited. But "[1]::2]" looks a bit like a broken array literal.

straight-shoota:
There is [IPAddr](https://ruby-doc.org/stdlib-2.5.1/libdoc/ipaddr/rdoc/IPAddr.html) but it's quite a bit different from Crystal's `IPAddress` which is essentially a socket address consisting of IP and port. In Ruby, it's only an IP address but with subnet mask.

```rb
p IPAddr.new "3ffe:505:2::1" # => #<IPAddr: IPv6:3ffe:0505:0002:0000:0000:0000:0000:0001/ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff>
```

straight-shoota:
You're right. Let's make it `Socket::IPAddress(127.0.0.1:80)`.

RX14:
Aren't we throwing away a lot of the information here?

asterite:
@straight-shoota  Oh, and `to_s` should behave like what you originally proposed for `inspect`.

Again, in Ruby:

```
$ irb
irb(main):001:0> require 'ipaddr'
=> true
irb(main):002:0> puts IPAddr.new "192.168.2.0"
192.168.2.0
```

straight-shoota:
@RX14 Not really. All the information is expressed in the string representation. The data structures are just pretty complicated which makes it look like there is a whole lot more information in it.

@asterite `#to_s` has been like this before and it doesn't change with this PR. It's even used by `inspect`.
Additionally to the Ruby string representation, there is also a port.

straight-shoota:
Maybe we should really rename `IPAddress` to `IPSocketAddress` or `SocketAddress` to make it more clear, that it is not just an IP address.

PR #6705, opened 2018-09-11
Add examples to Socket::Server#accept API docs

straight-shoota:


src/socket/server.cr
```diff
@@ -2,12 +2,32 @@ class Socket
...
 
     # Accepts an incoming connection and returns the client socket.
     #
     # Returns `nil` if the server is closed after invoking this method.
+    #
+    # ```
+    # require "socket"
+    #
+    # server = TCPServer.new(2202)
+    # while socket = server.accept
```
asterite: `accept` -> `accept?`

src/socket/server.cr
```diff
@@ -2,12 +2,32 @@ class Socket
   module Server
     # Accepts an incoming connection and returns the client socket.
     #
+    # ```
+    # require "socket"
+    #
+    # server = TCPServer.new(2202)
+    # while socket = server.accept
```
asterite: But `server.accept` is always truthy...

src/socket/server.cr
```diff
@@ -2,12 +2,33 @@ class Socket
   module Server
     # Accepts an incoming connection and returns the client socket.
     #
+    # ```
+    # require "socket"
+    #
+    # server = TCPServer.new(2202)
+    # while true
```
jhass: `loop do`?

asterite: Both are fine. I'm fact, I'm starting to think we should remove `loop`, it adds absolutely nothing and it's just another way of doing something.

jhass: well I guess it was useful when it had its block argument :P

PR #6706, opened 2018-09-11
Fix Socket#reuse_port? if SO_REUSEPORT is not supported

straight-shoota:
Currently, if `SO_REUSEPORT` is not supported, `LibC.getsockopt` returns `-1`. When calling `Socket#reuse_port?` this causes an exception.

This PR fixes that to return `false` if errno is `ENOPROTOOPT` (meaning that `SO_REUSEPORT` is not supported).

### DIFF
--- a/src/socket.cr
+++ b/src/socket.cr
@@ -391,7 +391,14 @@ class Socket < IO
   end
 
   def reuse_port?
-    getsockopt_bool LibC::SO_REUSEPORT
+    0 != getsockopt(LibC::SO_REUSEPORT, 0) do |errno|
+      # If SO_REUSEPORT is not supported, the return value should be `false`
+      if errno.errno == Errno::ENOPROTOOPT
+        return false
+      else
+        raise errno
+      end
+    end
   end
 
   def reuse_port=(val : Bool)
@@ -445,9 +452,13 @@ class Socket < IO
 
   # Returns the modified *optval*.
   def getsockopt(optname, optval, level = LibC::SOL_SOCKET)
+    getsockopt(optname, optval, level) { |errno| raise errno }
+  end
+
+  protected def getsockopt(optname, optval, level = LibC::SOL_SOCKET)
     optsize = LibC::SocklenT.new(sizeof(typeof(optval)))
     ret = LibC.getsockopt(fd, level, optname, (pointerof(optval).as(Void*)), pointerof(optsize))
-    raise Errno.new("getsockopt") if ret == -1
+    yield Errno.new("getsockopt") if ret == -1
     optval
   end
 
--- a/src/socket.cr
+++ b/src/socket.cr
@@ -391,7 +391,7 @@ class Socket < IO
   end
 
   def reuse_port?
-    0 != getsockopt(LibC::SO_REUSEPORT, 0) do |errno|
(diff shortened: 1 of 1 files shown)


src/socket.cr
```diff
@@ -391,7 +391,14 @@ class Socket < IO
   end
 
   def reuse_port?
-    getsockopt_bool LibC::SO_REUSEPORT
+    0 != getsockopt(LibC::SO_REUSEPORT, 0) do |errno|
```
Sija: Ouch, that looks sooo weird... Please, assign the return value to a variable and compare with that:

```cr
ret = getsockopt(LibC::SO_REUSEPORT, 0) do |errno|
  # ...
end
ret != 0
```

ysbaddaden:
I'm not sure. `reuse_port?` is meant to say whether `SO_REUSEPORT` is set, not swallow the fact that it's unsupported.

straight-shoota:
My thinking is: If `SO_REUSEPORT` is not supported, it can't be set. So the method should return `false`.

`socket.reuse_port = true` also works if `SO_REUSEPORT` is not supported. If anything should raise, it's the setter. Or both. Or none. But it makes no sense if only the getter raises.

I prefer either both or only the setter should raise. `reuse_port?` should just honestly answer if the socket has the port flagged for reuse which is false if `SO_REUSEPORT` is not supported.

ysbaddaden:
My thinking is "can I check `reuse_port?` before setting it?" If so, I could be wrongly led to believe it worked since the setter won't set errno, which is probably an issue by itself.

So yeah, maybe the check should just return false, but trying to set the value must raise if unsupported.

straight-shoota:
I've done some research and from what I've read, it seams that `getsockopt` *should* actually fail with `ENOPROTOOPT`. Unfortunately, this case isn't handled in `#setsockopt_bool` so it won't raise or return `false` or anything.

But, at least on WSL, `setsockopt` doesn't even complain about setting `SO_REUSEPORT`, only `getsockopt`.

straight-shoota:
There are however reports about `setsockopt` failing on WSL: https://github.com/IBM-Swift/BlueSocket/issues/81 Microsoft/WSL#1419
But I can't reproduce it in Crystal on WSL.

ysbaddaden:
Let Microsoft seal with WSL bugs. `setsockopt` should fail which is fine.

src/socket.cr
```diff
@@ -391,7 +391,15 @@ class Socket < IO
   end
 
   def reuse_port?
-    getsockopt_bool LibC::SO_REUSEPORT
+    ret = getsockopt(LibC::SO_REUSEPORT, 0) do |errno|
+      # If SO_REUSEPORT is not supported, the return value should be `false`
```
ysbaddaden: no need for the comment: code is explicit enough.

straight-shoota:
@ysbaddaden 

> `setsockopt` should fail which is fine.

The point is, it doesn't fail on WSL :(

RX14:
@straight-shoota 

> Let Microsoft seal with WSL bugs

straight-shoota:
Yeah, sure. That's why I posted a comment in the WSL issue. It's just that Crystal's `#reuse_port=` method also behaves incorrect. It's not a big deal of course.

Issue #5778, opened 2018-03-06
Refactor Sockets API


straight-shoota:
[#4277 (issue-comment)](https://github.com/crystal-lang/crystal/issues/4277#issuecomment-293552825)

There are quite a few errors in the API design related to sockets:

`TCPServer` inherits from `TCPSocket` which is conceptually true, but `TCPSocket` is a `Socket` which inherits from `IO` - yet a server cannot be used as an IO. The same is true for `UNIXServer`.

`UDPSocket` also inherits from `Socket` and `IO`, yet cannot be used as an `IO` either since UDP communication does not work as a stream.

`IPSocket` as parent of `UDPSocket` and `TCPSocket` defines accessors `local_address` and `remote_address`, yet `TCPServer` does not have single remote address. Neither does a `UDPSocket` in listening mode.

While it's not such a big issue, I would also suggest to put all these types in the `Socket` namespace. These socket types are the only non-basic types in the global name space. And large parts of the API are already in the `Socket` namespace. `TCPSocket` -> `Socket::TCP`, `TCPServer` -> `Socket::Server::TCP` (or `Socket::TCPServer`) etc. I will use namespaced types from now on but that should not influence the logical considerations.

`TCPSocket` and `UDPSocket` should be an `IO` because they are essentially streaming interfaces. The basic `Socket` however should not be an `IO` because `UDPSocket` cannot be used as an IO. This would call for an intermediary class `Socket::IO` as parent for `TCPSocket` and `UDPSocket` which inherits from `IO` and includes `Socket`.

* Socket
** Socket::UDP
** Socket::IO
*** Socket::TCP
*** Socket::UNIX

`Socket` therefore needs to be a module.

bew:
May be related: https://github.com/crystal-lang/crystal/issues/4317#issuecomment-295789662

straight-shoota:
@bew yeah, this assumes UNIX sockets are only used as streams.

straight-shoota:
Additionally, I would suggest that all sockets and servers can be created using appropriate `Socket::IPAddress` or `Socket::UNIXAddress` as constructor parameters.

Maybe these could also be renamed to `Socket::Address::IP` and `Socket::Address::UNIX` to clean up the namespace and follow the same naming as for sockets and servers.

[...]

PR #6707, opened 2018-09-11
Change `Socket::Server#accept` to return `IO` and add default implementation

straight-shoota:
This PR changes the return type of  `Socket::Server#accept` and `#accept?` to `IO` and `IO?`.
It decouples the interface from a specific `Socket` design. Right now, every `Socket` inherits from `IO`, but for a UDP socket implementation, this doesn't make sense because it is not a stream (see #5778).
The purpose of the `Socket::Server#accept` methods is to accept an incoming connection to use it as a stream IO. That's what the `IO` class is meant for.

It also adds a default implementation for `#accept` which returns the value of `#accept?` or raises if it is `nil`.
It is the same implementation as already used in `Socket#accept`.
This makes it easier for classes inheriting the module to implement the interface, they might just need to override `#accept?`.

### DIFF
--- a/src/socket/server.cr
+++ b/src/socket/server.cr
@@ -3,12 +3,14 @@ class Socket
     # Accepts an incoming connection and returns the client socket.
     #
     # If the server is closed after invoking this method, an `IO::Error` (closed stream) exception must be raised.
-    abstract def accept : Socket
+    def accept : IO
+      accept? || raise IO::Error.new("Closed stream")
+    end
 
     # Accepts an incoming connection and returns the client socket.
     #
     # Returns `nil` if the server is closed after invoking this method.
-    abstract def accept? : Socket?
+    abstract def accept? : IO?
 
     # Accepts an incoming connection and yields the client socket to the block.
     # Eventually closes the connection when the block returns.


RX14:
There are datagram protocols which use accept(), such as DCCP. accept or not depends on whether the protocol has the concept of a "connection". That's entirely orhogonal to whether the protocol is stream-based or not.

straight-shoota:
@RX14 Yes, I see, there are two different definitions of what gets accepted by a server. Either a socket connection or an IO stream. Right now in the stdlib it is everywhere used as returning an IO stream. Changing the return type to `IO` makes it more versatile so this concept can be used without explicitly using sockets (for example to mock a network connection).

Since we have no type in stdlib that models a non-stream socket connection and probably won't need that ever, I'm not sure if it makes sense base an abstract interface on this concept. And if if does, it should be an additional one.

Maybe it would be a good idea to rename `Socket::Server` to `IO::Server` or something similar to more explicitly express its purpose to accept IO streams.

RX14:
I'm unsure of the overall direction of these socket PRs, what's the end goal, what's the design?

straight-shoota:
@ysbaddaden 
1. Why not? `Socket` doesn't say anything about its purpose. We actually want an `IO` in order to use it for example in `HTTP::Server` to read and write to it.
2. The example just illustrates that, if `accept` returned an UDP socket, it wouldn't be useful in a context where you'd expect an IO.

@RX14 That's exactly my idea to have a raw socket implementation and let protocol specific implementations build upon that. Because of the complexity involved, I figured to first propose individual changes which are independent of the large socket refactoring because they already make sense with the current structure.

> There's a million different combinations a socket can behave, and only a few which are commonly used.

An `accept` method returning a generic `Socket` instance isn't very useful as it doesn't specify *how*  this socket is supposed to be used. It essentially says the thing it returns is a socket, but you need to know from the context if it is an IO stream or some other connection.

The only meaningful use case is to return an IO stream. This PR reduces complexity. `accept` simply returns an IO and everything's fine.

RX14:
@straight-shoota I don't understand how `Socket#accept` doesn't return a `Socket`. We've already covered a situation where it doesn't return an `IO`. `Socket` shouldn't be an `IO`, and it should have no subclasses. Accept returning a socket is absolutely the most useful thing that can be done.

`TCPServer#accept` will return a `TCPSocket` which is an `IO`. That's what 99% of users will use.

straight-shoota:
We're talking about `Socket::Server#accept`, not `Socket#accept`! 

> because it's a low-level interface that extremely few people will use.

`Socket` is low-level. `Socket::Server` is actually a high-level interface and only implemented by `TCPServer` and `UNIXServer`. They're providing IO streams.

When the return type is `IO`, it could also be implemented by a `MockServer` which is used for testing to mock a network connection.

RX14:
@straight-shoota currently, `TCPServer < TCPSocket < IPSocket < Socket`. So `Socket::Server#accept` **is** `Socket#accept`! That's the insane part and needs to be tackled first.

ysbaddaden:
Please, let's close, this is going nowhere. An API musn't be design to "ease testing". Don't mock but test the real thing instead.

Let's spend energy to change TCPServer / UNIXServer to compose a TCPSocket / UNIXSocket instead, which has already been accepted numerous times.

asterite:
> An API musn't be design to "ease testing"

I don't quite agree with this.

RX14:
@asterite I think what @ysbaddaden means was that an API shouldn't be degraded only to ease testing.

asterite:
Ah, that makes more sense.

straight-shoota:
I agree. But testability is not the reason for this change. It's just a nice byproduct.

@ysbaddaden This was supposed to be one step in preparation for decoupling the socket types. But if you insist that this should not be merged as is, fine. It'll be in a later PR then.

asterite:
Note that specifying that some method returns IO doesn't conflict with subclasses returning a stricter type. And then, the compiler doesn't cast the return value tui the return type, so if it really returns a Socket you can still call Socket methods.

PR #6710, opened 2018-09-12
Add Socket::IPAddress predicates and constants

straight-shoota:
This PR adds methods to query information of `IPAddress` instances

* `Socket::IPAddress#loopback?` returns whether the address is a loopback address (`127.0.0.0/24` or `::1`)
* `Socket::IPAddress#unspecified?` returns whether the address is unspecified (`0.0.0.0` or `::`)

Default values for `LOOPBACK`, `UNSPECIFIED` and `BROADCAST` are also added as constants for both IPv4 and IPv6.

### DIFF
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -528,17 +528,12 @@ module Crystal::Playground
 
       server = HTTP::Server.new handlers
 
-      host = @host
-      if host
-        address = server.bind_tcp host, @port
-      else
-        address = server.bind_tcp @port
-      end
+      address = server.bind_tcp @host || Socket::IPAddress::LOOPBACK, @port
       @port = address.port
 
       puts "Listening on http://#{address}"
-      if host == "0.0.0.0"
-        puts "WARNING running playground with 0.0.0.0 is unsecure."
+      if address.unspecified?
+        puts "WARNING running playground on #{address.address} is insecure."
       end
 
       begin
--- a/src/socket/address.cr
+++ b/src/socket/address.cr
@@ -72,6 +72,13 @@ class Socket
   # resolve an IP, or don't know whether a `String` constains an IP or a domain
   # name, you should use `Addrinfo.resolve` instead.
   struct IPAddress < Address
+    UNSPECIFIED  = "0.0.0.0"
+    UNSPECIFIED6 = "::"
+    LOOPBACK     = "127.0.0.1"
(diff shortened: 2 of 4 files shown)


src/socket/address.cr
```diff
@@ -190,6 +197,23 @@ class Socket
...
 
+    # Returns `true` if this IP is a loopback address.
+    #
+    # In the IPv4 family, loopback addresses are all addresses in the subnet
+    # `127.0.0.0/24`. In IPv6 `::1` is the loopback address.
+    def loopback? : Bool
+      if family == Family::INET
+        @addr4.not_nil!.s_addr & 0x00000000ff_u32 == 0x0000007f_u32
+      else
+        @address == LOOPBACK6
```
bew: What if the address is `0000:0000:0000:0000:0000:0000:0000:0001` ? It should work too

straight-shoota: Yeah, true. But maybe it's better to normalize the address...

src/socket/address.cr
```diff
@@ -216,12 +216,24 @@ class Socket
...
+        ipv6_addr8(addr) == StaticArray[0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8, 0_u8]
       else
         raise "unreachable!"
       end
     end
 
+    private def ipv6_addr8(addr : LibC::In6Addr)
+      {% if flag?(:darwin) || flag?(:openbsd) || flag?(:freebsd) %}
+      addr.__u6_addr.__u6_addr8
+      {% elsif  flag?(:linux) && flag?(:musl) %}
```
bew: there's a double space here.
Also, can you indent the code in the macro conditions?

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -528,17 +528,12 @@ module Crystal::Playground
...
-      else
-        address = server.bind_tcp @port
-      end
+      address = server.bind_tcp @host || Socket::IPAddress::LOOPBACK, @port
       @port = address.port
 
       puts "Listening on http://#{address}"
-      if host == "0.0.0.0"
-        puts "WARNING running playground with 0.0.0.0 is unsecure."
+      if address.unspecified?
```
RX14: Perhaps a later PR to make this have better intentions: test for not-loopback not 0.0.0.0 specifically

PR #6711, opened 2018-09-12
Improve socket specs

straight-shoota:
Adds new and improves existing specs for `Socket`, `TCPSocket`, `TCPServer`, `UNIXServer` and `UDPSocket`.

The diff is best viewed with `-w`/`&w=1` to ignore whitespace.

<del>This is still work in progress, I'm publishing it as PR to validate if the specs perform correctly on all platforms.</del>

### DIFF
--- a/spec/std/socket/socket_spec.cr
+++ b/spec/std/socket/socket_spec.cr
@@ -27,6 +27,28 @@ describe Socket do
     client.protocol.should eq(Socket::Protocol::TCP)
   end
 
+  it "sends messages" do
+    port = unused_local_port
+    server = Socket.tcp(Socket::Family::INET6)
+    server.bind("::1", port)
+    server.listen
+    address = Socket::IPAddress.new("::1", port)
+    spawn do
+      client = server.not_nil!.accept
+      client.gets.should eq "foo"
+      client.puts "bar"
+    ensure
+      client.try &.close
+    end
+    socket = Socket.tcp(Socket::Family::INET6)
+    socket.connect(address)
+    socket.puts "foo"
+    socket.gets.should eq "bar"
+  ensure
+    socket.try &.close
+    server.try &.close
+  end
+
   describe "#bind" do
     each_ip_family do |family, _, any_address|
       it "binds to port" do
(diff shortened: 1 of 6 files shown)


ysbaddaden:
Some specs are lost. For example `SO_REUSEPORT` or ipv4 connections to ipv6 server running on `::`.

straight-shoota:
>  ipv4 connections to ipv6 server running on `::`.

Was there a explicit spec for that? I can't find it... But I can add it, of course.

straight-shoota:
The specs that are now failing might be bugs, but I'm not sure.

* LINUX: When querying a previously set `recv_buffer_size`, a different value is returned.
* ALL: Connecting with an an IPv6 address (`::1`) to an IPv4 server (`0.0.0.0`) fails. Not sure if that is expected. The other way around (`127.0.0.1` to `::`) works.
* MACOS: UDPSocket can't send broadcast to `255.255.255.255`.

RX14:
@straight-shoota connecting ipv6 to ipv4 is expected to fail. Binding to `::` binds to v4 and v6, binding to `0.0.0.0` binds to v4 only. Depending on kernel options though, which might make this test brittle

straight-shoota:
So, it's probably better not to spec this?

straight-shoota:
The spec connecting with `127.0.0.1` to `::` fails on FreeBSD.

RX14:
@straight-shoota whether `::` binds to ipv4 is entirely OS-dependant and depends on sysctl values, see http://man7.org/linux/man-pages/man7/ipv6.7.html ctrl-f `IPV6_V6ONLY`. It's default value is controlled via sysctl. So *if* we could specifically specify this flag then we could spec the behaviour, but as it is, we leave this at default setting, so we shouldn't spec this.

straight-shoota:
The two failing specs are now marked as pending and I'll open new issues for tracking them.
Now the specs compile on every platform.

The fails on travisci are unrelated to this PR.

straight-shoota:
Can this get a review, please?

straight-shoota:
I added two additional specs that came up during refactoring.

wontruefree:
Thank you @straight-shoota !

PR #6716, opened 2018-09-13
Move Socket.ip? to Socket::IPAddress.ip?

straight-shoota:
The method is specific to `IPAddress`, so it shouldn't be defined in the parent namespace.

It could also be `Socket::Address.ip?`, but it fits more closely in `IPAddress`.

### DIFF
--- a/spec/std/socket/address_spec.cr
+++ b/spec/std/socket/address_spec.cr
@@ -162,65 +162,65 @@ describe Socket::UNIXAddress do
   end
 end
 
-describe Socket do
+describe Socket::Address do
   # Tests from libc-test:
   # http://repo.or.cz/libc-test.git/blob/master:/src/functional/inet_pton.c
   it ".ip?" do
     # dotted-decimal notation
-    Socket.ip?("0.0.0.0").should be_true
-    Socket.ip?("127.0.0.1").should be_true
-    Socket.ip?("10.0.128.31").should be_true
-    Socket.ip?("255.255.255.255").should be_true
+    Socket::IPAddress.ip?("0.0.0.0").should be_true
+    Socket::IPAddress.ip?("127.0.0.1").should be_true
+    Socket::IPAddress.ip?("10.0.128.31").should be_true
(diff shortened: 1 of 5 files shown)


asterite:
```crystal
if Socket.ip?(address)
  # ...
end
```

I think that was the point, it's almost readable English.

This new way is longer and redundant.

bew:
Darn I agree with both of you here^^

In this case there could be a `Socket.ip?` that simply redirects to `Socket::Address.ip?` as a fancy-english-method?

straight-shoota:
@bew That would be an alias. I don't think anyone would want that.

bew:
Oh right ^^ didn't even see it coming

RX14:
Maybe `Socket::IPAddress.parse?` which returns the `Socket::IPAddress` if it's valid, nil otherwise? Then it'd fit.

straight-shoota:
That sounds good. `ip?` and `parse` (or rather `new(host, port)`) already do pretty much the same, so it looks like a nice solution. Will need some refactoring in the implementation, but I could give it a try.

straight-shoota:
Or, well. The only issue is `port`. `Socket.ip?` is about checking if a string is an IP address. `IPAddress.parse` creates an instance of `IPAddress` and expects the argument to be a URI including an IP and port.

RX14:
Oh, yeah. I keep forgetting `IPAddress` isn't an IP address. Really bad naming, if it has a port then it probably shouldn't be on `IPAddress` at all.

wontruefree:
would it make more sense to have `Socket::Address.ip?` be `Socket::Address.valid?`?

straight-shoota:
@wontruefree A path could also be a valid (UNIX) address.

wontruefree:
I just thought it might a little redundant to have `IPAddress.ip?`

ysbaddaden:
Asterite said it all: the proposed change doesn't improve anything. Further proposed alternatives bring more confusion. What's there to discuss further here? `Socket.ip?(address)` reads better and is simpler. Let's keep it.

PR #6717, opened 2018-09-13
Extract Socket constants in separate file.

straight-shoota:
`Socket::Address` and `Socket::Addrinfo` conceptually don't dependent on
`Socket`. The shared constants and lib includes are extracted to `common.cr` to separate concerns more clearly.

### DIFF
--- a/spec/std/socket/address_spec.cr
+++ b/spec/std/socket/address_spec.cr
@@ -1,5 +1,5 @@
 require "spec"
-require "socket"
+require "socket/address"
 
 describe Socket::Address do
   describe ".parse" do
--- a/spec/std/socket/addrinfo_spec.cr
+++ b/spec/std/socket/addrinfo_spec.cr
@@ -1,5 +1,5 @@
 require "spec"
-require "socket"
+require "socket/addrinfo"
 
 describe Socket::Addrinfo do
   describe ".resolve" do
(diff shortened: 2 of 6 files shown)


ysbaddaden:
I dont understand. If they're common constants, that they'll be included anyway, and they're still named Socket::X then I expect them to be defined in `socket.cr` not buried in some other file.

Can we avoid stylistic/opinion changes, and concentrate on fixing *actual* issues in Socket?

straight-shoota:
This is not about style, it is about separation of concerns. I'd like `socket.cr` to focus on the `Socket` type, not some other stuff that is in its namespace but not solely required by `Socket`.
The idea is, if you only need `Socket::Address`, you can just include `socket/address` instead of the entire package.
While this selective requiring doesn't have any relevant performance impact as it might have in other languages, the expression of explicit dependencies is still valuable.

This PR is in preparation for refactoring the entire socket API. In the process I found it was much easier to separate individual pieces of code into separate files instead of having the main type + some common stuff in the main file.

This PR is not an integral part of the refactoring, that's why I published it separately. But it makes refactoring easier.

jhass:
So, where do we want to go with this? 18 months later it didn't seem to have been a pain point for anybody given the activity here...

straight-shoota:
Closing stale PR.

PR #3536, opened 2016-11-12
Add dig method to Enumerable

samueleaton:
Add the `dig` method as seen in Ruby's [Hash#dig](https://ruby-doc.org/core-2.3.1/Hash.html#method-i-dig) and [Array#dig](https://ruby-doc.org/core-2.3.1/Array.html#method-i-dig)

~~Also adds a `dig?` method that will not raise an error but will instead return `nil` if the key path doesn't exist.~~

PR #6719, opened 2018-09-14
Add Hash/Indexable#dig/dig?

Sija:
Followup to #3536.

### DIFF
--- a/spec/std/hash_spec.cr
+++ b/spec/std/hash_spec.cr
@@ -105,6 +105,51 @@ describe "Hash" do
     end
   end
 
+  describe "dig?" do
+    it "gets the value at given path given splat" do
+      ary = [1, 2, 3]
+      h = {"a" => {"b" => {"c" => [10, 20]}}, ary => {"a" => "b"}}
+
+      h.dig?("a", "b", "c").should eq([10, 20])
+      h.dig?(ary, "a").should eq("b")
+    end
+
+    it "returns nil if not found" do
+      ary = [1, 2, 3]
+      h = {"a" => {"b" => {"c" => 300}}, ary => {"a" => "b"}}
+
+      h.dig?("a", "b", "c", "d", "e").should be_nil
+      h.dig?("z").should be_nil
+      h.dig?("").should be_nil
+    end
+  end
+
+  describe "dig" do
+    it "gets the value at given path given splat" do
+      ary = [1, 2, 3]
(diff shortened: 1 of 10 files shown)


j8r:
It would be also great to have `#dig(keys : Enumerable)`, where `keys` represents a variable sequence of keys.

RX14:
Thought we decided against this. It was `.dig`  or `JSON::Any` and `JSON::Any` won because my proposal forgot that `JSON::Type` included more than `Hash` and `Array`. I proposed it could be solved by making `JSON.parse` return `Hash | Array` because the 95% usecase for json is arrays or hashes at the top level, but that idea wasn't well recieved.

@j8r that method would just produce insane types.

Sija:
@RX14 I understood differently, also according to your own https://github.com/crystal-lang/crystal/pull/3536#issuecomment-403773699.
@j8r That would forgo all compile-time safety, ditto what @RX14 said.

I still see it as a very handy addition to stdlib and see no reason *not* to implement it.

asterite:
I was against `dig` because I thought the implementation would make the return type be the union of everything. But this implementation is **brilliant**. 

```crystal
a = {1 => [[1, 2], [3, 4]]}

p typeof(a.dig(1, 0)) # => Array(Int32)
p a.dig(1, 0) # => [1, 2]

p typeof(a.dig(1, 0, 1)) # => Int32
p a.dig(1, 0, 1) # => 2
```

So I see it as a very nice, convenient way, to traverse deep structures. Now we would just need to add it to `JSON::Any` and `YAML::Any`... or remove those ugly `Any` wrappers altogether (I'd start by adding `dig/dig?` to those types, and eventually decide how to remove them).

Sija:
> So I see it as a very nice, convenient way, to traverse deep structures.

Exactly.

>... or remove those ugly `Any` wrappers altogether (I'd start by adding `dig/dig?` to those types, and eventually decide how to remove them).

Yes, please! 🎉

asterite:
@Sija Do you want to add `dig/dig?` to `Any` then? At least that way this PR will be "complete".

Sija:
@asterite Done.
@codenoid Indeed, good to c u back! :)

Sija:
Is it GTG now? CI fails seems not related, btw.

bcardiff:
@Sija CircleCI had a bad week in their servers.

In darwin is still failing because it trying to compile the compiler with llvm 7, which is still not supported. The dependency of the formula is `llvm` plain and that leads to llvm 7 right now.

PR #6720, opened 2018-09-14
Treat untyped element in tuple and named tuple

### LINKED ISSUE #6718: Regression in cleanup trannsformer
The current [nightly build](https://circleci.com/gh/straight-shoota/crinja/28) of crinja shows a compiler regression in master. The build with a compiler at `master`  fails, while `0.26.1` is fine.

This was probably introduced by #6659

I couldn't exactly pinpoint the failing part yet (the code base is quite large).
But it is reproducible with:
```shell
$ git clone https://github.com/straight-shoota/crinja && cd crinja
$ crystal-master spec spec/lib/filter_spec.cr
```

Callstack:
```
Nil assertion failed at /crystal/src/compiler/crystal/semantic/bindings.cr:530 (Exception)
  from /crystal/src/nil.cr:107:5 in 'not_nil!'
  from /crystal/src/compiler/crystal/semantic/bindings.cr:0:5 in 'program'
  from /crystal/src/compiler/crystal/semantic/bindings.cr:539:26 in 'update'
  from /crystal/src/compiler/crystal/semantic/bindings.cr:532:5 in 'update'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:746:7 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:380:7 in 'transform_many'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:262:10 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:258:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:258:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:138:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:324:7 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:7 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:263:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:138:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:461:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:299:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:138:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:461:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:299:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:138:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/transformer.cr:10:7 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:167:16 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:138:9 in 'transform'
  from /crystal/src/compiler/crystal/syntax/transformer.cr:7:14 in 'transform'
  from /crystal/src/compiler/crystal/semantic/cleanup_transformer.cr:9:14 in 'cleanup'
  from /crystal/src/compiler/crystal/semantic/main_visitor.cr:13:7 in 'visit_main'
  from /crystal/src/compiler/crystal/semantic/main_visitor.cr:5:5 in 'visit_main:process_finished_hooks:cleanup'
  from /crystal/src/compiler/crystal/semantic.cr:22:7 in 'semantic'
  from /crystal/src/compiler/crystal/compiler.cr:152:7 in 'compile'
  from /crystal/src/compiler/crystal/command/spec.cr:74:14 in 'spec'
  from /crystal/src/compiler/crystal/command.cr:88:7 in 'run'
  from /crystal/src/compiler/crystal/command.cr:46:5 in 'run'
  from /crystal/src/compiler/crystal/command.cr:45:3 in 'run'
  from /crystal/src/compiler/crystal.cr:8:1 in '__crystal_main'
  from /crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from /crystal/src/crystal/main.cr:86:7 in 'main'
  from /crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```

/cc @MakeNowJust

MakeNowJust:
Fixed #6718

The reason of #6718 is calling `node.update` with untyped `TupleLiteral` or `NamedTupleLiteral`. An untyped node has no `node.program`, so it fails nil assertion.

I think this `node.update` is not needed because `CleanupTransformer` does not change the type of `node` except for untyped and unreachable node.  However such nodes should be transformed to other node. In other words `node.update` does nothing in general.

Thus, this replaces `node.update` with treating untyped element.

### DIFF
--- a/spec/compiler/semantic/named_tuple_spec.cr
+++ b/spec/compiler/semantic/named_tuple_spec.cr
@@ -302,4 +302,23 @@ describe "Semantic: named tuples" do
       meta
     end
   end
+
+  it "doesn't crash on named tuple in not executed block (#6718)" do
+    assert_type(%(
+      require "prelude"
+
+      def pending(&block)
+      end
+
+      def untyped(x = nil)
+      end
+
+      # To reproduce this bug, it is needed to the expression that is
+      # not typed on main phase but is typed on cleanup phase.
+      # `untyped(untyped)` is just one.
+      pending do
+        {s: untyped(untyped)}
+      end
+    )) { nil_type }
+  end
 end
--- a/spec/compiler/semantic/tuple_spec.cr
+++ b/spec/compiler/semantic/tuple_spec.cr
@@ -300,4 +300,23 @@ describe "Semantic: tuples" do
       )
     }
   end
+
+  it "doesn't crash on tuple in not executed block (#6718)" do
+    assert_type(%(
+      require "prelude"
+
+      def pending(&block)
+      end
+
+      def untyped(x = nil)
+      end
+
+      # To reproduce this bug, it is needed to the expression that is
+      # not typed on main phase but is typed on cleanup phase.
+      # `untyped(untyped)` is just one.
+      pending do
+        {untyped(untyped)}
+      end
+    )) { nil_type }
+  end
 end
--- a/src/compiler/crystal/semantic/cleanup_transformer.cr
+++ b/src/compiler/crystal/semantic/cleanup_transformer.cr
@@ -741,6 +741,12 @@ module Crystal
         return exps
       end
 
+      # `node.program` is assigned by `MainVisitor` usually, however
+      # it may not be assigned in some edge-case (e.g. this `node` is placed
+      # at not invoked block.). This assignment is for it.
+      node.program = @program
+      node.update
+
       node
     end
 
@@ -757,6 +763,9 @@ module Crystal
         exps.bind_to(exps.expressions.last)
         return exps
       end
+
+      node.program = @program
+      node.update
 
       node
     end
--- a/src/compiler/crystal/semantic/cleanup_transformer.cr
+++ b/src/compiler/crystal/semantic/cleanup_transformer.cr
@@ -729,7 +729,10 @@ module Crystal
 
     def transform(node : TupleLiteral)
       super
-      node.update
+
+      unless node.elements.all? &.type?
+        return untyped_expression node
+      end
 
       no_return_index = node.elements.index &.no_returns?
       if no_return_index
@@ -743,7 +746,10 @@ module Crystal
 
     def transform(node : NamedTupleLiteral)
       super
-      node.update
+
+      unless node.entries.all? &.value.type?
+        return untyped_expression node
+      end
 
       no_return_index = node.entries.index &.value.no_returns?
       if no_return_index


MakeNowJust:
Yes, there is no untyped expression in tests at least on cleapup phase after `super` called. It's important thing. (and I noted this as comment in spec.)

Maybe it is the simplest solution is moving `node.update` in behind of `NoReturn` processing and adding `node.program = @program` in front of `node.update`.
Perhaps it works fine. But, I'm wondering why this `node.update` is really needed.

MakeNowJust:
And, the above I commented is very conservative way.

asterite:
I don't understand, why is there a "Nil assertion failed"? Because `@program` is `nil`?

straight-shoota:
> Because @program is nil?

Yes, looks like it.

MakeNowJust:
Because it is not assigned to `node.program` (`node` is `TupleLiteral` or `NamedTupleLiteral`). This assignment is done in `MainVisitor` normally. However, in the test case, the block of `pending` is not processed by `MainVisitor` due to not invoked, so the tuple literal inside block is untyped and its `node.program` is `nil`.

asterite:
I see.

Well, we have two solutions:
1. Assign `@program` before calling `node.update`
2. Remove `@program` altogether. It's needed to create a `TupleType` (`program.tuple_type(...)`) but we all `Type`s have a reference to the `Program`, so we can simply do: `program = elements.first.type` inside `TupleLiteral#update` (after the first `return unless`...). Same goes for `NamedTupleLiteral`)

I much prefer the second option. It makes the code more robust (we don't need to remember to assign `@program`) and it makes `TupleLiteral` occupy less memory.

@MakeNowJust Would you like to do 2, or do you want me to do that?

MakeNowJust:
@asterite Removing `node.program` is good idea, but I don't know it works for an empty tuple and empty named tuple...

And it seems you are sure `node.update` is needed. But I'm unbelieving in it. Could you explain about this?

asterite:
@MakeNowJust Ah, right, I forgot about the empty case. Then yeah, let's set `node.program = ...` before calling `node.update`.

straight-shoota:
🏓

straight-shoota:
I'm a bit lost on how this should proceed... Who needs to take what action?

MakeNowJust:
Updated.

spec/compiler/semantic/named_tuple_spec.cr
```diff
@@ -302,4 +302,23 @@ describe "Semantic: named tuples" do
...
+    assert_type(%(
+      require "prelude"
+
+      def pending(&block)
+      end
+
+      def untyped(x = nil)
+      end
+
+      # To reproduce this bug, it is needed to the expression that is
```
RX14: FYI the correct english is "... it is needed to have an expression that is ..." and it reads even better as "... it is required to have an expression that is ..."

PR #6723, opened 2018-09-14
Adds ability to quote always in CSV.build

maiha:
Currently, `CSV.build` quotes only those fields which contain special characters.
PR adds ability to quote all fields by **quote_always** option.

```crystal
puts CSV.build {|csv| csv.row 101, "doesn't"}
puts CSV.build(quote_always: true) {|csv| csv.row 101, "doesn't"}
```

```text
101,doesn't
"101","doesn't"
```

This allows us to link to tools that are not RFC compliant. For example, **ClickHouse** can't parse the former.
https://github.com/yandex/ClickHouse/issues/2192

Best regards,


asterite:
Maybe `always_quote` reads better (I'm not a native English speaker so other's opinion is best)

src/csv/builder.cr
```diff
@@ -136,13 +150,20 @@ class CSV::Builder
     end
 
     private def needs_quotes?(value)
-      value.each_byte do |byte|
-        case byte.unsafe_chr
-        when @separator, @quote_char, '\n'
-          return true
+      case @quoting
+      when .rfc?
+        if value.is_a?(String)
```
bcardiff: If the value is a Char or Symbol the quoting logic is not applied.
That was a pre-existing corner case that was not handled.
Should we handled it now?

maiha: Well, I prefer this PR just provides a new **feature** about quoting behavior.
And then, I'd like to fix the corner case soon in another PR that provides **fix the existing issue**.

maiha:
FYI: There was almost no performance change in compatible mode(`Quoting::RFC`).

```crystal
Benchmark.ips(warmup: 0, calculation: 10) do |x|
  x.report("csv(orig)") { build_csv1(reports) }
  x.report("csv(RFC) ") { build_csv2(reports, "RFC") }
  x.report("csv(ALL) ") { build_csv2(reports, "ALL") }
end
```

```
csv(orig)   3.27  (305.83ms) (± 6.77%)  238787461 B/op        fastest
csv(RFC)    3.22  (310.67ms) (± 9.10%)  238787333 B/op   1.02× slower
csv(ALL)    2.52  (396.54ms) (± 9.99%)  238786808 B/op   1.30× slower
```
- sample data: rows=109,488, cols=16, bytes=26.5 MB

src/csv/builder.cr
```diff
@@ -35,15 +35,21 @@ require "csv"
 # 4,5,6,7,8
 # ```
 class CSV::Builder
+  enum Quoting
+    NONE # No quotes
```
RX14: This documentation doesn't work, it needs to be before the constants to show up in `crystal doc`.

src/csv.cr
```diff
@@ -111,16 +111,25 @@ class CSV
...
   #
+  # Takes optional *quoting* argument to define quote behavior.
+  #
   # ```
   # result = CSV.build do |csv|
   #   csv.row "one", "two"
   #   csv.row "three"
   # end
   # result # => "one,two\nthree\n"
+  # result = CSV.build(quoting: CSV::Builder::Quoting::ALL) do |csv|
```
RX14: `quoting: :all`

maiha: Wow, I didn't know such a cool syntax! Thank you!

RX14:
I don't see why is was made an enum, it should just be a boolean, `always_quote`.

RX14:
Actually i've changed my mind. If all these CSV libs have an option for no quoting, there must be a reason. An enum is fine.

asterite:
Right, but I'd like to know the reason.

maiha:
> What's the point of no quotes? Isn't that a broken format?

To be honest, I just imitated other libraries, so I don't know exact reasons.

In my opinion, it can be expected to bypass the judgment process and speed up a little, such as when it is composed only of numbers or alphabets.

asterite:
Makes sense

maiha:
Oh, I forgot about special keywords `All` and `None`. However, this should only be enabled in the case of `Flags`. Right?

bew:
@maiha the error is not about the special enum members All & None, it looks like an automatic cast error

maiha:
Since it seems that it takes time to solve the problem, I reverted the refactoring codes about "use Symbol for enum".

PR #6732, opened 2018-09-16
Allow `it` and `pending` to be used without blocks

tswicegood:
Both `it` and `pending` can be used in RSpec without a block. Calling `it` without a block is a way of specifying that it is pending. I was caught off guard when writing a spec and it failed in both instances because I lacked a block.

### DIFF
--- a/spec/std/spec_spec.cr
+++ b/spec/std/spec_spec.cr
@@ -117,6 +117,9 @@ describe "Spec matchers" do
 end
 
 describe "Spec" do
+  it "should treat it calls without blocks as pending"
+  pending "should allow pending calls without blocks"
+
   describe "use_colors?" do
     it "returns if output is colored or not" do
       saved = Spec.use_colors?
--- a/spec/std/spec_spec.cr
+++ b/spec/std/spec_spec.cr
@@ -117,9 +117,6 @@ describe "Spec matchers" do
 end
 
 describe "Spec" do
-  it "should treat it calls without blocks as pending"
-  pending "should allow pending calls without blocks"
-
   describe "use_colors?" do
     it "returns if output is colored or not" do
       saved = Spec.use_colors?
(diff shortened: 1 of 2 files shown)


spec/std/spec_spec.cr
```diff
@@ -117,6 +117,9 @@ describe "Spec matchers" do
 end
 
 describe "Spec" do
+  it "should treat it calls without blocks as pending"
+  pending "should allow pending calls without blocks"
```
tswicegood: Both of these add two "pending" test cases to the output. I'm open to other ideas on how to test this. The compiler does a really good job of providing a "failing" test case without the new methods in `spec/methods.cr` though.

asterite: Just don't test it. The code is trivially simple.

RX14:
`pending` without a block, yes. `it` without a block, no.

straight-shoota:
Is the `it` overload really required? Why not just use `pending` everywhere? It's essentially an alias...

RX14:
it is an alias

spec/std/spec_spec.cr
tswicegood: 👍

tswicegood:
@straight-shoota, @RX14 the [documented way][1] to mark tests as incomplete in RSpec is to write `it "should do something"` without the block. I imagine I'm not the only one who'd try to write out a bunch of pending tests via `it …` and no block and be surprised at a failure.

[1]: https://relishapp.com/rspec/rspec-core/v/2-0/docs/pending/pending-examples#pending-implementation

asterite:
Yes, but this is not RSpec and Crystal tries to avoid aliases. I guess anyone will be surprised to find out there's no let, before each, etc.

tswicegood:
In fairness, @asterite – when I came across no `it` as pending, the first thing I tried was rewriting it to `pending`, so I guess most folk's figure it out. 😄 

I'll remove the new `it` signature and push up the changes. Hopefully it'll run this time. Not sure what's up with those failed tests. Looks like they timed out.

PR #6735, opened 2018-09-16
 Give proper error when trying to instantiate Module

### LINKED ISSUE #3885: Improve error message when trying to instantiate module
Ruby has a slightly more helpful error message when you try to instantiate a module which has an `initialize` method (got this after an incomplete conversion of a module to a class).

```crystal
module Foo
  def initialize
  end
end

Foo.new

# crystal foo.cr
# Error in foo.cr:6: undefined local variable or method 'allocate'
#
# Foo.new
#     ^~~

# ruby foo.cr
# foo.cr:6:in `<main>': undefined method `new' for Foo:Module (NoMethodError)
```

r00ster91:
Fixes #3885 

Now this:
```cr
module Foo
  def initialize
  end
end
Foo.new
```
gives:
```c
Error in test.cr:5: instantiating 'Foo:Module#new()'

Foo.new
    ^~~

in test.cr:2: cannot instantiate Foo:Module

  def initialize
  ^
```
instead of:
```c
Error in line 5: instantiating 'Foo:Module#new()'

in line 2: undefined local variable or method 'allocate'
```

### DIFF
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -39,7 +39,7 @@ end
 class Crystal::Call
   def raise_matches_not_found(owner, def_name, arg_types, named_args_types, matches = nil, with_literals = false)
     if def_name == "allocate" && owner.is_a?(ModuleType) && owner.is_a?(MetaclassType)
-      raise "Cannot instantiate #{owner}"
+      raise "cannot instantiate #{owner}"
     end
 
     # Special case: Foo+.class#new
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -97,10 +97,11 @@ class Crystal::Call
             msg << "undefined method '#{def_name}'"
           else
             msg << "undefined local variable or method '#{def_name}'"
-            if def_name == "allocate" && owner.is_a?(ModuleType) && owner.is_a?(MetaclassType)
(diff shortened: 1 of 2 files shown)


straight-shoota:
I'm not sure this is a good change. `allocate` could just be a regular method on a module and could be called as suche. The error message is at least misleading in a case where the method was just not defined - because some include is missing, or the def is mispelled for example. This could very well be a valid call even on a module.
```cr
module Foo
  def self.allocate
  end
end

Foo.allocate`
```
Failing to call a method `allocate` on a module doesn't mean you're trying to instantiate it.

I'd leave the error message as is and maybe add a note about being unable to instantiate a module.

jhass:
Perhaps we rather should make sure the call to `.new` already fails unless it's defined explicitly

RX14:
Specs need updating

src/compiler/crystal/semantic/call_error.cr
```diff
@@ -100,6 +100,10 @@ class Crystal::Call
           end
         end
 
+        if def_name == "allocate" && owner.is_a?(ModuleType) && owner.is_a?(MetaclassType)
```
asterite: This condition should check if onwer is a MetaclassType and owner.instance_type.module?

The current check reads well but makes no sense because a MetaclassType inherits from ModuleType. It just happens to work because the only metaclasses that don't have an allocate are modules, though I think this check will also trigger for enums

Issue #6737, opened 2018-09-17
Segfault with OpenSSL 1.1.1


t-richards:
## System information

 - Crystal 0.26.1
 - OpenSSL 1.1.1

## Synopsis

Crystal programs will successfully compile against this OpenSSL version, but appear to have issues at runtime when performing read/write operations over TLS connections.

The new LTS version of OpenSSL, version 1.1.1, was released September 11, 2018.

 - https://www.openssl.org/blog/blog/2018/09/11/release111/

For example, carc.in just so happens to have OpenSSL 1.1.1
 - https://carc.in/#/r/50pp

## Example code

```crystal
require "http/client"

response = HTTP::Client.get "https://tomrichards.net/"
puts response.body
```

## Error output

```sh-session
Invalid memory access (signal 11) at address 0x8
[0x562297c94f96] *CallStack::print_backtrace:Int32 +118
[0x562297c8524d] __crystal_sigfault_handler +61
[0x7f99e48e93c0] ???
[0x562297ce95e0] ???
[0x562297c8582e] ~procProc(Pointer(LibCrypto::Bio), Pointer(UInt8), Int32, Int32) +46
[0x7f99e5005343] ???
[0x7f99e5005803] BIO_write +35
[0x7f99e50038eb] ???
[0x7f99e5005cd7] BIO_ctrl +119
[0x7f99e527a49e] ???
[0x7f99e527eee0] ???
[0x7f99e527999a] ???
[0x7f99e5265496] SSL_do_handshake +86
[0x562297d245a6] *OpenSSL::SSL::Socket::Client#initialize<TCPSocket, OpenSSL::SSL::Context::Client, Bool, String>:Nil +342
[0x562297d24436] *OpenSSL::SSL::Socket::Client::new:context:sync_close:hostname<TCPSocket, OpenSSL::SSL::Context::Client, Bool, String>:OpenSSL::SSL::Socket::Client +214
[0x562297d14c13] *HTTP::Client#socket:IO+ +499
[0x562297d1449c] *HTTP::Client#send_request<HTTP::Request>:Bool +60
[0x562297d1441d] *HTTP::Client#exec_internal_single<HTTP::Request>:(HTTP::Client::Response | Nil) +29
[0x562297d1423d] *HTTP::Client#exec_internal<HTTP::Request>:HTTP::Client::Response +29
[0x562297d14216] *HTTP::Client#exec<HTTP::Request>:HTTP::Client::Response +6
[...]

PR #6738, opened 2018-09-17
Support OpenSSL 1.1.1

ysbaddaden:
OpenSSL 1.1.0 introduced a public API to create and populate the BIO and BIO_METHOD structs. We now use it when available. We still keep the manual creation for OpenSSL 1.0.2 (and below) which is still supported.

This fixes an internal ABI breaking change introduced in OpenSSL 1.1.1 that causes a compiled program to segfault. See #6737.

This also implements the new BIO read/write methods that take `size_t` lengths instead of `int`. This skips the compatibility methods provided by OpenSSL.

NOTE: the patch also starts to use uppercased external symbols, such as `BIO_set_init`, to avoid `bio_set_init = BIO_set_init` aliases. This is limited to the `BIO_` and `BIO_meth_*` used in this patch.

### DIFF
--- a/src/openssl/bio.cr
+++ b/src/openssl/bio.cr
@@ -3,22 +3,19 @@ require "./lib_crypto"
 # :nodoc:
 struct OpenSSL::BIO
   CRYSTAL_BIO = begin
-    crystal_bio = LibCrypto::BioMethod.new
-    crystal_bio.name = "Crystal BIO".to_unsafe
-
-    crystal_bio.bwrite = LibCrypto::BioMethodWrite.new do |bio, data, len|
+    bwrite = LibCrypto::BioMethodWriteOld.new do |bio, data, len|
       io = Box(IO).unbox(bio.value.ptr)
       io.write Slice.new(data, len)
       len
     end
 
-    crystal_bio.bread = LibCrypto::BioMethodRead.new do |bio, buffer, len|
+    bread = LibCrypto::BioMethodReadOld.new do |bio, buffer, len|
       io = Box(IO).unbox(bio.value.ptr)
       io.flush
       io.read(Slice.new(buffer, len)).to_i
     end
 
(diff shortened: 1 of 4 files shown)


src/openssl/bio.cr
```diff
@@ -34,31 +47,56 @@ struct OpenSSL::BIO
...
 
-    crystal_bio.create = LibCrypto::BioMethodCreate.new do |bio|
-      bio.value.shutdown = 1
-      bio.value.init = 1
-      bio.value.num = -1
+    create = LibCrypto::BioMethodCreate.new do |bio|
+      {% if LibCrypto::OPENSSL_110 %}
+        LibCrypto.BIO_set_shutdown(bio, 1)
+        LibCrypto.BIO_set_init(bio, 1)
+        # bio.value.num = -1
```
ysbaddaden: I have no idea what `BIO#num` means and what we should do here. Anybody knows?

bcardiff: It seems to come originally from https://github.com/crystal-lang/crystal/commit/4995fb010316ac0cb3085ab18f8738f7b7415140#diff-65c7a3575f24669104f83007f62e90c6R109 if the documentation of 1.1.x says nothing about it it's fine as it is I guess.

bcardiff:
This kind of issues makes me want to enforce more infrastructure to test compatibility with different libraries. Otherwise we in lack of sight.

RX14:
@bcardiff we'd have had to keep track of the openssl release and add it to the test infastructure. It's honestly better to be reactionary about these things to avoid wasting effort. Maybe once there's far more manpower in the project.

RX14:
tested locally

ysbaddaden:
What's really needed is a manual review of the OpenSSL integration which dates back to the old OpenSSL 0.9.8. See what public API they provide now, but was kinda internal/private.

For example the changes here were already present in OpenSSL 1.1.0, and would have avoided the reported segfault.

But that requires having an active maintainer...

PR #6744, opened 2018-09-18
Refactor Process in preparation for Windows support

RX14:
Commits are best read in order, with their descriptions.

### DIFF
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -147,6 +147,7 @@ class IO::FileDescriptor < IO
   end
 
   def reopen(other : IO::FileDescriptor)
+    return other if self.fd == other.fd
     system_reopen(other)
 
     other
--- a/src/process.cr
+++ b/src/process.cr
@@ -194,9 +194,7 @@ class Process
   def self.exec(command : String, args = nil, env : Env = nil, clear_env : Bool = false, shell : Bool = false,
                 input : Stdio = Redirect::Inherit, output : Stdio = Redirect::Inherit, error : Stdio = Redirect::Inherit, chdir : String? = nil)
     command, args = prepare_args(command, args, shell)
-    unless exec_internal(command, args, env, clear_env, input, output, error, chdir)
-      raise Errno.new("execvp")
-    end
+    exec_internal(command, args, env, clear_env, input, output, error, chdir)
   end
 
   getter pid : Int32
@@ -261,19 +259,21 @@ class Process
       begin
         reader_pipe.close
         writer_pipe.close_on_exec = true
-        unless Process.exec_internal(
-                 command,
-                 args,
-                 env,
-                 clear_env,
-                 fork_input || input,
-                 fork_output || output,
-                 fork_error || error,
-                 chdir
-               )
-          writer_pipe.write_bytes(Errno.value)
-          writer_pipe.close
-        end
+        Process.exec_internal(
+          command,
+          args,
+          env,
+          clear_env,
+          fork_input || input,
+          fork_output || output,
+          fork_error || error,
+          chdir
+        )
+      rescue ex : Errno
+        writer_pipe.write_bytes(ex.errno)
+        writer_pipe.write_bytes(ex.message.try(&.bytesize) || 0)
+        writer_pipe << ex.message
+        writer_pipe.close
       rescue ex
         ex.inspect_with_backtrace STDERR
         STDERR.flush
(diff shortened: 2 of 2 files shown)


src/process.cr
```diff
@@ -428,31 +446,33 @@ class Process
 
     Dir.cd(chdir) if chdir
 
-    LibC.execvp(command, argv) != -1
+    argv = [command.check_no_null_byte.to_unsafe]
+    args.try &.each do |arg|
+      argv << arg.check_no_null_byte.to_unsafe
```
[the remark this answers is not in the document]
bew: Not exactly, the existing code was:
```cr
argv = [command.to_unsafe]
args.try &.each do |arg|
  argv << arg.to_unsafe
end
argv << Pointer(UInt8).null
```
Oh that's right, no idea where I got that idea then..

src/process.cr
```diff
@@ -285,9 +285,13 @@ class Process
     writer_pipe.close
     bytes = uninitialized UInt8[4]
     if reader_pipe.read(bytes.to_slice) == 4
-      reader_pipe.close
       errno = IO::ByteFormat::SystemEndian.decode(Int32, bytes.to_slice)
```
ysbaddaden: Can we clean this up? Passing errno and message size (both Int32) must have the same logic. IMO a simple `io.write_bytes(x)` and `x = io.read_bytes(Int32)` is enough. No need to deal with endianess in parent/child IPC —it should be on *both* sides anyway.

src/kernel.cr
```diff
@@ -2,12 +2,20 @@
...
 {% else %}
   require "c/unistd"
 
   STDIN  = IO::FileDescriptor.from_stdio(0)
   STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
+
+  ORIGINAL_STDIN  = IO::FileDescriptor.new(0, blocking: true)
+  ORIGINAL_STDOUT = IO::FileDescriptor.new(1, blocking: true).tap { |f| f.flush_on_newline = true }
+  ORIGINAL_STDERR = IO::FileDescriptor.new(2, blocking: true).tap { |f| f.flush_on_newline = true }
```
ysbaddaden: I'm not sure. This creates more constants for a single use: reopen stdio file descriptors on fork/exec for POSIX targets. I think it was better placed in `Process#exec_internal`.

Given the refactor, can we just init on-demand in `Process.to_real_fd`? With `Process#exec_internal` just passing `STDIN`, ... to `reopen_io`? That would then be abstracted to be system specific, with `to_real_fd` being a noop on windows (unless it needs something):

```crystal
protected def self.exec_internal(...) : NoReturn
  reopen_io(input, STDIN)
  reopen_io(output, STDOUT)
  reopen_io(error, STDERR)

  # ...
end

private def self.to_real_fd(fd)
  {% unless flag?(:windows) %}
    # reopen original stdio devices on POSIX systems:
    case fd
    when STDIN then IO::FileDescriptor(0, blocking: true)
    when STDOUT then IO::FileDescriptor(1, blocking: true)
    when STDERR then IO::FileDescriptor(2, blocking: true)
  {% end %}
  fd
end
```

src/process.cr
```diff
@@ -428,31 +446,33 @@ class Process
...
 
-    LibC.execvp(command, argv) != -1
+    argv = [command.check_no_null_byte.to_unsafe]
+    args.try &.each do |arg|
+      argv << arg.check_no_null_byte.to_unsafe
+    end
+    argv << Pointer(UInt8).null
+
+    LibC.execvp(command, argv)
+    raise Errno.new("execvp")
```
straight-shoota: Could we avoid the overhead of raising an exception that is immediately caught? That's how it was before.
Could still have an handler for other `Errno` raised in `reopen_io`.

RX14: I don't think it's worth the effort. An exception will still be raised either way, and it just duplicates code to have to check for errno in two ways.

src/process.cr
RX14: This is done because there's no `read_bytes?(Int32)` which does not raise on EOF. EOF is the hot path, the case where the exec succeeds. So the *first* read call has to be coded like this check for EOF without having the overhead of raising and rescuing an exception in the successful case. Adding `read_bytes?(Int32)` which allows you to check for EOF would be nice, but it's another PR.

src/kernel.cr
ysbaddaden: Are there other use cases than restoring the original file descriptors on fork/exec for POSIX targets? which must be achieved through Process#fork and Process#exec.

I'd let them in `#to_real_fd`, at least for the time being, and see with use-cases (if any) and after integrating windows (that doesn't need it) if it's worth changing.

RX14: Yeah I'll move these constants to private in `Process`.

RX14: Done.

ysbaddaden:
Let's merge.

RX14:
I was planning on rebasing the windows part of the process refactor on top of this PR to make 100% sure that this PR was complete, but I'm moving back to university so havent had time :(

I'm fine to merge this, but there's a very small but nonzero chance that there might be more amendments coming.

src/io/file_descriptor.cr
```diff
@@ -147,6 +147,7 @@ class IO::FileDescriptor < IO
   end
 
   def reopen(other : IO::FileDescriptor)
+    return if self.fd == other.fd
```
wontruefree: should this return self?
will this introduce a `Nil` into the types that can be returned?

RX14: Thanks for spotting! (this method should probably return nil but thats for another PR.

PR #6748, opened 2018-09-19
Configure Crystal compiler version in Makefile

### LINKED ISSUE #6715: Compiler version detection
When the compiler is built, it stores the version of the compiler source code in the executable which can later be printed with `crystal --version`.

There is some rather complicated logic in [`crystal/compiler/config.cr`](https://github.com/crystal-lang/crystal/blob/8154c2d1894e3e82a28284701cd5841bf749af31/src/compiler/crystal/config.cr#L32) (#2856) to determine the current version.

Essentially, it reads the current git tag and uses that if available. Otherwise, the version of the compiling compiler is used, either bare or with `+?` appended depending on whether it's inside a git repository. All this can be overriden by setting the env var `CRYSTAL_CONFIG_VERSION`.
The version is also accompanied by the SHA hash of the current commit. This is only available when building in a git repository and always uses HEAD.

There are a few issues with this approach, some have already been mentioned in #4642 and #4687.

The compiler provided as FreeBSD port tells it is `0.24.0+?+?+?+?`. This does not only hide the fact that it is actually a 0.25.1 compiler but also results in a `Crystal::VERSION` that can't be handled by `SemanticVersion`. This compiler currently fails to build `int_spec.cr` because it contains the `compare_versions` macro. It can be fixed by setting `CRYSTAL_CONFIG_VERSION` in the build process, but the compiler should not produce such invalid versions.
In general, when building from a tarball, the compiler has no information whatsoever on the current source version. Proper version reporting (including SHA) depends on using a git clone. This is not very practical for many use cases where a tarball would suffice.

Builds can be categorized in two different types:

* A *development* build compiles a shallow version somewhere in between releases. It is used for development and testing purposes (locally, in CI, etc.). The version should be based on the latest release version (which is NOT necessarily the compiling compiler's version) with some kind of suffix like `-dev` or an offset with the number of commits since the release. And it should of course include a commit hash if available. These are typically built in a git repository, but not necessarily.
* A *release* build compiles the compiler at a specific released version (this includes official binaries, pre-releases, OS packages etc.). The source code might live in a git clone, but can as well be from a tarball. These builds a typically automated in some kind of build script. Version and commit hash are fixed and should be configured in the build process.

I think it would be a good idea to make *release* builds always require explicit configuration to use the appropriate version information. Since it's mostly automated and the build systems already have the information, this shouldn't be a big deal.
Every other build without explicit configuration is automatically a development build and won't have a release version.  But the reported version of a development build should accurately describe the state of the source code.

Using the version of the current compiler (`Crystal::VERSION`) is not an option. A newer compiler might also compile an older source, or a recent source could be compiled by a compiler that is older than the last release.

When git is available, the currently used `git describe --tags --long --always 2>/dev/null` would return something like `0.26.1-60-g1337b09` for current master. This is really nice, especially because it shows the number of commits. But it is also pretty unreliable because git tags are not automatically synchronized. You can have the latest master branch, but unless you also have the `0.26.1` tag, it will report as being based on an older release (for example: `0.25.1-218-g1337b09`). To update the tags you need to run something like `git fetch --tags upstream`.

I think the only way to get this straight is to document the latest release in the source code. Then it will be available everywhere (not depend on git) and reliably correspond to the code.
This is already available in CHANGELOG and could be implemented similar to  #4687. But I would not rely on the git version at all. Maybe the git data could be to enhance the version info by the number of commits since the latest release tag (if it matches the latest release in the changelog) and the commit hash.
For builds without git it should be an issue to loose some information about the number of commits and the commit hash.

Regarding implementation, it might be preferable to decouple the logic and move it out of the compiler source. Then the compiler just needs to read `CRYSTAL_CONFIG_VERSION` and `CRYSTAL_CONFIG_COMMIT`.

Unless specified (for a release build) the Makefile would populate them with `$(grep -oEi '(^|\n)# ([0-9]+\.[0-9]+\.[0-9]\w*)' CHANGELOG.md | head -n1 | awk '{print $2}')-dev` and the commit hash of HEAD. Maybe some additional logic could validate the base version of `git describe --tags --long --always 2>/dev/null` and use the offset. But that's an advanced feature and I'd suggest to focus on getting the basics right first.

straight-shoota:
Some complicated logic was used to detect the current compiler version
from git tags. This method is unreliable and could produce invalid
results.

This is completely removed from Crystal source code and the version is
set as env var `CRYSTAL_CONFIG_VERSION` in the `Makefile`. The value is
determined from the most recent release listed in `CHANGELOG.md`. This
is a reliable method even if the code is not checked out in a git
repository.

By default, `-dev` suffix is appended to the version. For creating a
release build with a non-dev version, `CRYSTAL_CONFIG_VERSION` needs
to be explicitly specified.

Example usage (assuming inside a git repo):
```console
$ # Development build
$ make crystal
$ bin/crystal version
Crystal 0.26.1-dev [ga107c84] (2018-09-19)
$ # Release build
$ make crystal CRYSTAL_CONFIG_VERSION=0.27.0-rc1
$ bin/crystal version
Crystal 0.27.0-rc1 [ga107c84] (2018-09-19)
```

Previously, the development version would have been `Crystal 0.26.1+64 [ga107c84]`, or `Crystal 0.26.1+?` if not built inside a git repository. The release version would have been `Crystal 0.27.0 [ga107c84]`.

The number of commits since release has been lost. But I don't think it has much meaning and is unreliable. It wouldn't be difficult to add it back, though. A maybe even more important option could be including the branch name in a development build, if inside a git repository.

An alternative or supplementary source for the most recent release could also be a `VERSION` file. But  `CHANGELOG.md` should be pretty solid.

Fixes #6715

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,9 @@ SOURCES := $(shell find src -name '*.cr')
 SPEC_SOURCES := $(shell find spec -name '*.cr')
 override FLAGS += $(if $(release),--release )$(if $(stats),--stats )$(if $(progress),--progress )$(if $(threads),--threads $(threads) )$(if $(debug),-d )$(if $(static),--static )$(if $(LDFLAGS),--link-flags="$(LDFLAGS)" )
 SPEC_FLAGS := $(if $(verbose),-v )$(if $(junit_output),--junit_output $(junit_output) )
-EXPORTS := $(if $(release),,CRYSTAL_CONFIG_PATH="$(PWD)/src")
+CRYSTAL_CONFIG_VERSION ?= $(shell grep -oEi '(^|\n)\#\s+([0-9]+\.[0-9]+\.[0-9]\w*)' --max-count=1 CHANGELOG.md | awk '{print $$2}')-dev
+CRYSTAL_CONFIG_COMMIT ?= $(shell git rev-parse --short HEAD 2> /dev/null)
+EXPORTS := $(if $(release),,CRYSTAL_CONFIG_PATH="$(PWD)/src") CRYSTAL_CONFIG_VERSION=$(CRYSTAL_CONFIG_VERSION) CRYSTAL_CONFIG_COMMIT=$(CRYSTAL_CONFIG_COMMIT)
 SHELL = sh
 LLVM_CONFIG_FINDER := \
   [ -n "$(LLVM_CONFIG)" ] && command -v "$(LLVM_CONFIG)" || \
@@ -58,6 +60,13 @@ ifeq (${LLVM_CONFIG},)
   $(error Could not locate llvm-config, make sure it is installed and in your PATH, or set LLVM_CONFIG)
 else
   $(shell echo $(shell printf '\033[33m')Using $(LLVM_CONFIG) [version=$(shell $(LLVM_CONFIG) --version)]$(shell printf '\033[0m') >&2)
+endif
+
+ifeq (${CRYSTAL_CONFIG_VERSION},)
+  $(error Missing version information, could not read from CHANGELOG.md. Please provide a valid CRYSTAL_CONFIG_VERSION setting)
+endif
+ifeq (${CRYSTAL_CONFIG_VERSION},-dev)
+  $(error Missing version information, could not read from CHANGELOG.md. Please provide a valid CRYSTAL_CONFIG_VERSION setting)
 endif
 
 .PHONY: all
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -1,11 +1,21 @@
 module Crystal
   module Config
-    def self.path
-      {{env("CRYSTAL_CONFIG_PATH") || ""}}
+    def self.path : String
+      {{ env("CRYSTAL_CONFIG_PATH") || "" }}
     end
 
-    def self.version
-      version_and_sha.first
+    def self.version : String
+      {% begin %}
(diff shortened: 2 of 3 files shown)


bcardiff:
I am not in favor of using the changelog as a source of information for the compiler or the build process.

The current source code already allows a specific version to be used as branding (set a value in `CRYSTAL_CONFIG_VERSION` env var and that will get embedded in the generated binary).

In other projects I worked in, having the literal version that is been built is nice as simple. Then we enforce that tagged build match that description and non tag build have a `-dev` suffix. This is equivalent to what you are proposing in.

One issue I have with changing the tagging is that after a release, until some period of time we don't know if the next release will be a minor or patch. So there is some inconsistency if the version is not updated immediately after release.

I understand the recent glitch in travis, and that the process can be changed/simplified, but this also affect the whole build process.

I don't think we should invest effort in this area for the time being because:

* The official releases are always tagged correctly right now (fingers crossed :-P)
* The CI is happy
* There is working way to allow non official built to inject the version information. 

Maybe we should document better how to create releases of other distros to help mainteners a bit more. I believe the core nature of the PR is that.

j8r:
Ideally the solution should be in a separate project – `crystal-lang/crystal` isn't the only project facing this issue, and others can use the adopted solution. In a way this will standardize it. I'm thinking to `crystal-lang/shards`, or versioning bots.

Issue #6646, opened 2018-09-02
Crystal and GMP


proyb6:
The previous discussion was in 2015 but tested with pi digits using BigInteger and how do we implement GMP?
https://groups.google.com/forum/#!topic/crystal-lang/cY8bImdHWV4

Tested PHP 7.1 on macOS Mojave completed around 2 seconds for the same number of iterations:
https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/pidigits-php-5.html

C, Ruby and Crystal Source code:
https://github.com/kostya/crystal-benchmarks-game#pidigits
https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/pidigits-yarv-1.html

asterite:
What I mean is something like this:

```crystal
require "big"
N = (ARGV[0]? || 100).to_i

struct BigInt
  def add(other : BigInt) : self
    LibGMP.add(self, self, other)
    self
  end

  def add(other : Int) : self
    if other < 0
      sub(other.abs)
    elsif other <= LibGMP::ULong::MAX
      LibGMP.add_ui(self, self, other)
    else
      add(other.to_big_i)
    end
    self
  end

  def sub(other : Int) : self
    if other < 0
      add(other.abs)
    elsif other <= LibGMP::ULong::MAX
      LibGMP.sub_ui(self, self, other)
    else
      sub(other.to_big_i)
    end
    self
  end

  def sub(other : BigInt) : self
    LibGMP.sub(self, self, other)
    self
  end

  def mul(other : BigInt) : self
    LibGMP.mul(self, self, other)
    self
  end

  def mul(other : LibGMP::IntPrimitiveSigned) : self
    LibGMP.mul_si(self, self, other)
    self
  end

  def mul(other : LibGMP::IntPrimitiveUnsigned) : self
    LibGMP.mul_ui(self, self, other)
    self
  end
end

i = 0
k = 0
ns = 0.to_big_i
a = 0.to_big_i
t = 0
u = 0.to_big_i
k1 = 1
n = 1.to_big_i
d = 1.to_big_i

while true
  k += 1
  t = n << 1
  n.mul(k)
  k1 += 2
  a.add(t)
  a.mul(k1)
  d.mul(k1)
  if a >= n
    t, u = (n * 3 + a).divmod(d)
    u.add(n)
    if d > u
      ns.mul(10)
      ns.add(t)
      i += 1
      if i % 10 == 0
[...]

PR #6755, opened 2018-09-20
Issue #6646 extend GMP implementation

cfsamson:
I made this pull request after the discussion in thread #6646. This solved most of current BigInt implementations performance issues, and using the Benchmarks game pidigits benchmark as a measurement of performance sets this on par with the Rust implentation (within ~5%).

You can now mutate a BigInt variable and reuse a variable for the most common operations add, sub, mul, div, submul, addmul.

The api will look like this:
```
# mutating a variable comparable to a += b, avoiding new allocations:
a = # some big int
b = # some big int
a.add(b) # a += b

# Reusing a variable, avoiding new allocations:
a = # some big int
b = # some big int
c = # some big int
a.add(b, c) # a = b + c
```
See the suggested documentation for more information.

### DIFF
--- a/src/big/big_int.cr
+++ b/src/big/big_int.cr
@@ -172,7 +172,7 @@ struct BigInt < Int
     elsif op1 <= LibGMP::ULong::MAX
       LibGMP.add_ui(self, op2, op1)
     else
-      add(op1, op2.to_big_i)
+      add(op1.to_big_i, op2)
     end
     self
   end
--- a/src/big/big_int.cr
+++ b/src/big/big_int.cr
@@ -115,6 +115,78 @@ struct BigInt < Int
     end
   end
 
+  # a.add(b), a += b, avoiding allocations by mutating a
+  #
+  #    a = 100.to_big_i
+  #    a.add(10) => # a = 110
+  def add(other : BigInt) : self
+    LibGMP.add(self, self, other)
+    self
+  end
+
+  # a.add(b), a += b, avoiding allocations by mutating a
+  def add(other : Int) : self
+    if other < 0
+      sub(other.abs)
+    elsif other <= LibGMP::ULong::MAX
+      LibGMP.add_ui(self, self, other)
+    else
+      add(other.to_big_i)
+    end
+    self
+  end
+
+  # a.add(b,c), a = b+c avoiding allocations by mutating a. This is a nice way
+  # to "reuse" `BigInt` variables and avoid new allocations. Beware that
+  # the value of "a" will be overwritten with the result of b*c
+  #
+  #    a = 0.to_big_i
+  #    a.add(10_to_big_i, 5.to_big_i) => # a = 15
+  #
+  #    a = 100.to_big_i
+  #    a.add(10.to_big_i, 5.to_big_i) => # a = 15
+  def add(op1 : BigInt, op2 : BigInt) : self
+    LibGMP.add(self, op1, op2)
+    self
+  end
+
+  # a.add(b,c), a = b+c avoiding allocations by mutating a. This is a nice way
+  # to "reuse" `BigInt` variables and avoid new allocations. Beware that
+  # the value of "a" will be overwritten with the result of b*c
+  def add(op1 : BigInt, op2 : Int) : self
+    if op2 < 0
+      sub(op1, op2.abs)
+    elsif op2 <= LibGMP::ULong::MAX
+      LibGMP.add_ui(self, op1, op2)
+    else
+      add(op1, op2.to_big_i)
+    end
+    self
+  end
+
+  # ditto
+  def add(op1 : Int, op2 : BigInt) : self
+    if op2 < 0
+      LibGMP.abs(op2, op2)
+      sub(op1, op2)
+    elsif op1 <= LibGMP::ULong::MAX
+      LibGMP.add_ui(self, op2, op1)
+    else
+      add(op1, op2.to_big_i)
+    end
+    self
+  end
+
+  # ditto
+  def add(op1 : Int, op2 : Int) : self
+    if op2 < 0
+      sub(op1.to_big_i, op2.abs)
+    else
+      add(op1.to_big_i, op2)
+    end
+    self
+  end
+
   def -(other : BigInt) : BigInt
     BigInt.new { |mpz| LibGMP.sub(mpz, self, other) }
   end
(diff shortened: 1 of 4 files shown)


src/big/big_int.cr
```diff
@@ -115,6 +115,78 @@ struct BigInt < Int
...
+  # a.add(b,c), a = b+c avoiding allocations by mutating a. This is a nice way
+  # to "reuse" `BigInt` variables and avoid new allocations. Beware that
+  # the value of "a" will be overwritten with the result of b*c
+  #
+  #    a = 0.to_big_i
+  #    a.add(10_to_big_i, 5.to_big_i) => # a = 15
+  #
+  #    a = 100.to_big_i
+  #    a.add(10.to_big_i, 5.to_big_i) => # a = 15
+  def add(op1 : BigInt, op2 : BigInt) : self
```
bew: I suggest using another name like `save_add` to make it explicit that you're not adding anything to `self` but that you are saving the addition of 2 other numbers to `self`

cfsamson: I have no problem with that.

I just wonder if splitting them was necessary since a add taking two arguments normally add them together and return a result, but if you only have 1 argument the only logical thing that can happen is that it's added to the calling object.

If we divide them in two we could also rename `a.add(b)` to `a.add_with(b)` so it's clear that a is added with b, but `save_add` is good too. Do you know if there's any other places in std lib that has to tackle this problem and has a naming convention we could use? I can't think of anything myself.

src/big/big_int.cr
```diff
@@ -153,6 +299,133 @@ struct BigInt < Int
...
+    self
+  end
+
+  # a.mul(b) a*b avoiding allocations by mutating a
+  def mul(other : LibGMP::IntPrimitiveSigned) : self
+    LibGMP.mul_si(self, self, other)
+    self
+  end
+
+  #ditto
```
r00ster91: Please format the code using `crystal tool format`.

cfsamson: Yeah, I noticed. Will do.

src/big/big_int.cr
bew: Nothing like this in the stdlib iirc.

FWIW, when I saw `a.add(b, c)` I thought at first that it'd do `a = a + b + c`, not `a = b + c`.

cfsamson: OK, good point. I'll see if someone else chimes in with opinions, but I see we might need to change this.

Sija: or maybe even `add(*ops : BigInt) : self`

src/big/big_int.cr
```diff
@@ -115,6 +115,78 @@ struct BigInt < Int
     end
   end
 
+  # a.add(b), a += b, avoiding allocations by mutating a
+  #
+  #    a = 100.to_big_i
+  #    a.add(10) => # a = 110
```
Sija: Please, use triple backticks (```) for code blocks, ditto all below.

src/big/big_int.cr
straight-shoota: That being said, I also wouldn't use the method name `add` because it doesn't properly convey the exact meaning of this operation.
I've looked at the Ruby GMP bindings in the [gmp gem](https://www.rubydoc.info/gems/gmp/0.7.43/GMP/Z) where there is a class method `.add(result, op1, op2)` and an instance method `#add!(b)`.
That's not ideal either because it separates both variants (one and two arguments) into completely different methods, yet they implement the same behaviour (just with a default argument). Sill, from a naming perspective, this is much better because `#add!` and `.add` clearly communicate that they are different from say `#+`.

cfsamson: Yes I know, I thought about that, but also I thought when someone reads the api it will be clear there are two different uses, and defaulting to self doesn't work for sub so I thought I'd better make them consistent. But yes, no problem to change that.

straight-shoota: @Sija I don't think it makes any sense to have a splat of operator arguments. And I fail to see any benefit over using a aggregation method like `ops.each { |op| sum.add!(op) }`

Sija: > That's not ideal either because it separates both variants (one and two arguments) into completely different methods, yet they implement the same behaviour (just with a default argument).

Current implementation is better suited for `.add(result, op1, op2)`. `#add` without `self` is quite misleading.

IMO `#add(!)` implies using `self` as part of the addition - not as an argument (default or not - it's an implementation detail).

src/big/big_int.cr
cfsamson: OK, I'll do that.

src/big/big_int.cr
cfsamson: I looked at the Ruby code too, and didn't get much wiser. As I see it there are three good options:
1. Keep as is and document it
2. Change the `a.add(b, c)` to something else
3. Consider if we really need `a.add(b)`, `a.add(b, c)` is the more flexible of them I think since you can do `a.add(a, b)` if I'm not mistaken.

For 2. could something like `a.add_two(b, c)` work?

cfsamson: Go uses `c.Add(a, b)` where the sum a + b and stores the result in c

src/big/big_int.cr
```diff
@@ -116,15 +116,16 @@ struct BigInt < Int
   end
 
   # a.add(b), a += b, avoiding allocations by mutating a
-  #
-  #    a = 100.to_big_i
-  #    a.add(10) => # a = 110
+  # ```
```
Sija: Please, add an empty line between the comment and the code block, ditto all below.

src/big/big_int.cr
Sija: I'd model the API after Ruby, since IMO it's clear about how these operations work, i.e.:

- `#add!(other)` => `a = a + other`
- `.add(result, op1, op2)` => `result = op1 + op2`

src/big/big_int.cr
cfsamson: done

src/big/big_int.cr
bew: I agree for `#add!` but not for the second, it's unclear to me that it won't add all args and return a new instance with the result.

bew:
Continuing discussion after https://github.com/crystal-lang/crystal/pull/6755#discussion_r219320571 (@Sija)

> I'd model the API after Ruby, since IMO it's clear about how these operations work, i.e.:
> 
> * `#add!(other)` => `a = a + other`
> * `.add(result, op1, op2)` => `result = op1 + op2`

I agree for `#add!` but not for the second, it's unclear to me that it won't add all args and return a new instance with the result.

Sija:
> [...] it's unclear to me that it won't add all args and return a new instance with the result.

You pass a _result_ argument so judging from that, it's clear (at least to me) it's gonna be (re)used for the returned value.

Otherwise signature would be `.add(op1, op2)`.

straight-shoota:
> You pass a result argument

Unless it's a required named argument, this isn't clear at the call site. That's the place where this method needs to distinguish itself from `#+`.

I guess it could also be `#add(op1, op2)` but I'm not entirely fond of it, because it looks like adding two summands to `self`, when it is actually adding two summands and writing the result to `self`.

cfsamson:
@straight-shoota I agree. In addition to having the add methods spread around more by dividing them, there is no president (as far as I know) in other Crystal libraries to have operations that takes a result argument like this. But I agree that `#add` is not perfect either. 

Since `a.add(b) => #a += b` seems to be the "normal" way to read an add method, can it be a solution to make the `a.add(b, c)` to `a.add!(b, c)` to signal that this method deviates from what most will assume that the "normal" add operation does?

Just for reference, I send the link to @asterite 's thoughts from the other thread which also adress this very issue: https://github.com/crystal-lang/crystal/issues/6646#issuecomment-419756459

asterite:
Suggestion: move `BigInt` and other types to a shard, so it can be maintained outside of the core team (I think no one in the core team has experience with libgmp).

RX14:
Do we have any users of `Big*` in the stdlib?

asterite:
@RX14 No, not at all

cfsamson:
I'm fine with that. I'll close this then if there are no objections.

bcardiff:
For the record, I would vouch for `#add!` as a mutating add method. Even if there is no `#add` since `+` can be used for that. I think that it is more clear.

PR #6758, opened 2018-09-20
 Improve sprintf docs, minor String::Formatter refactors

r00ster91:
This removes redundant spaces in the [`sprintf`](https://crystal-lang.org/api/0.26.1/toplevel.html#sprintf%28format_string%2C%2Aargs%29%3AString-class-method) documentation. The spaces have just been copied from Ruby's [`sprintf.c`](https://github.com/ruby/ruby/blob/96db72ce38b27799dd8e80ca00696e41234db6ba/sprintf.c#L219). They are there because Ruby uses the 4-space-code-markdown but in Crystal that's not required because Crystal uses the ```-markdown.

And this also adds the missing `c` format flag. [It's documented](https://github.com/crystal-lang/crystal/blob/391785249f66aaf5929b787b29810aeb4af0e1e8/src/kernel.cr#L122) but not actually implemented.

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -73,7 +73,7 @@ end
 # The syntax for a format specifier is:
 #
 # ```text
-# %[flags][width][.precision][type]
+# %[flags][width][.precision]type
 # ```
 #
 # A format specifier consists of a percent sign, followed by optional flags,
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -73,7 +73,7 @@ end
 # The syntax for a format specifier is:
 #
 # ```text
-# %[flags][width][.precision]type
+# %[flags][width][.precision][type]
 # ```
 #
 # A format specifier consists of a percent sign, followed by optional flags,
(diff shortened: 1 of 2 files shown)


src/kernel.cr
```diff
@@ -73,7 +73,7 @@ end
 # The syntax for a format specifier is:
 #
 # ```text
-# %[flags][width][.precision]type
+# %[flags][width][.precision][type]
```
straight-shoota: What? The type is required. That's why it is not in brackets.

r00ster91: Oh I didn't realize the brackets indicate that the argument is optional.

src/string/formatter.cr
```diff
@@ -110,7 +110,7 @@ struct String::Formatter(A)
 
   private def consume_width(flags)
     case current_char
-    when '1'..'9'
+    when '0'..'9'
```
bew: I think this is wrong, unless the 0 is after a non-0 digit, 0 is a separate flag: https://github.com/crystal-lang/crystal/blob/391785249f66aaf5929b787b29810aeb4af0e1e8/src/kernel.cr#L150

(_I can't fully check the implementation, i'm on my phone_)

PR #6764, opened 2018-09-21
removed filter/blur properties

### LINKED ISSUE #6751: crystal api scrolling jitterness and distortion
windows 10, 64-bit. Firefox Quantum 62.0 (latest), 144hz display

### Issue
When visiting any page on https://crystal-lang.org/api/master/, and when you scroll, there is a small distortion and jitterness (most annoying thing ever!). 
It was not like this 5-6 months ago, something was changed i think

Might be hard to show for a gif example, but you can definitely notice it yourself.
Distorted scrolling:

Smooth:

girng:
fixes #6751

### DIFF
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -553,15 +553,6 @@ span.flag.purple {
 .js-modal-visible .modal-background {
   display: flex;
 }
-.types-list,
-.main-content {
-  filter: blur(0);
-  transition: filter 200ms;
-}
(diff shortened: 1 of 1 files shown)


RX14:
@straight-shoota this is your code, is this fine to remove?

straight-shoota:
It's not an essential feature. But blurring the background behind the modal window has some visual significance. I wouldn't simply remove it, but look for alternative ways to achieve this.
I haven't looked at the exact issue in detail (and won't be able to for another week). But I'm sure there should be a way to have a blurred background when the modal is active without any performance issues.

Issue #3546, opened 2016-11-15
restriction of abstract def return value, not works


kostya:
this is compiled, but should not:

```crystal
require "json"

abstract class A
  abstract def bla : String # <= expected to return string only

  def show
    p bla.to_json
  end
end

class B < A
  def bla
    1 # <= Int32
  end
end

B.new.show
```

in case when i change my api, but all code is compiled ok O_o, i want compiler to show this bugs

cjgajard:
see also #160

asterite:
Duplicate of #160

kostya:
i not sure that this is #160, because this is about abstract defs(which is declare interfaces and restrictions), which compiler should checks to implement by it declaration.

asterite:
OK, I'll reopen, but it's very similar

damaxwell:
I recently was surprised by this myself.  

After playing around a bit I'm starting to suspect that the return type on an abstract method has zero ramifications whatsoever.  E.g.,

```crystal
module Interface
  abstract def foo : Zork
end

struct A
  include Interface

  def foo
    3.14159
  end
end

struct B
  include Interface

  def foo
    "pi"
  end
end

struct C
  @x : Interface

  def initialize(@x : Interface)
  end

  def foo
    @x.foo
  end
end

a = A.new
c = C.new(a)
typeof(c.foo) # => (Float64 | String)
```

The nonexistence of the type `Zork` is irrelevant, the code compiles, and the type `Zork` appears to have no effect on any real return type.  I.e., `: Zork` at the end of the abstract method declaration seems to be a fancy form of whitespace.  

If that's the case, maybe it would make sense for now to just outlaw return types on abstract methods.  At least then there is then no surprise factor: the compiler tells you that it won't enforce restrictions on the return type, so don't even bother trying.  Allowing return types on abstract methods could be restored in the future if they ever become meaningful.

damaxwell:
Great.  I think this is a pretty easy change, and would be happy to give it a shot, unless someone else (@RX14 ?) already plans to.

Fryguy:
Just ran into this, and it was confusing.  I understand, and agree with, the arguments from #160, but I think in the case for abstract methods is a little different, because I would like to enforce a return type of my subclasses.  I'm not a fan of disallowing return types on abstract methods, as it feels like just a temporary band-aid.

jgaskins:
Looks like this bug may still be active. This compiles on v0.30.1:

```crystal
[...]

Issue #5275, opened 2017-11-11
Instance method loses type inheriting from abstract generic class


jclem:
Given an abstract generic class with an abstract instance method that returns the generic type, when a concrete class inherits from that class, the compiler does not enforce that the concrete class's implementation of the instance method returns the generic type.

See https://play.crystal-lang.org/#/r/32ld

```
abstract class Foo(T)
  abstract def foo : T
end

class Bar < Foo(String)
  def foo
    1
  end
end

puts Bar.new.foo
```

Crystal v0.23.1, MacOS 10.13.1

MakeNowJust:
This happens against not only generics:

```crystal
abstract class Foo
  abstract def foo : String
end

class Bar < Foo
  def foo
    1
  end
end

puts Bar.new.foo
```

bew:
Note: Adding the type restriction on `Bar#foo` will error as we would expect.

Looks like a bug to me

RX14:
Duplicate of https://github.com/crystal-lang/crystal/issues/3546

See also https://github.com/crystal-lang/crystal/issues/160

Issue #160, opened 2014-07-12
Child class's methods can change parent return type


booch:
I created a child class that overrides its parent class's method, but doesn't specify the return type. I expected the compiler to know that the child return type had to be the same as the parent's. I also created an overridden method that changed the return type of the parent class, and did not get an error about that.

jhass:
What's the outstanding issue here? Could someone update the title to clarify?

ozra:
+1 on being able to continue to specify return types. The "catch errors early" is what is of interest here. I find it much easier to "lock down" the inference in key places - not _all_ the time - to ensure getting an error in a reasonable place instead of a 10 mile long chain.

However, monkey patching is such a super thing, one of the good things Crystal's gotten from Ruby. So, sealing classes, or limiting the possibility of what you _can do_ with super classes is unnecessary. Instead, if you want a strict type restriction on a method. Declare that type on the methods in overrides and/or subclasses too. I don't think that's an unreasonable demand. This gives the best of both worlds imo.

bcardiff:
Reading some parts again, @asterite I would vote for either

1. explicit subclass check on return types, or
2. do the ducktype checking states in your mention to me at https://github.com/manastech/crystal/issues/160#issuecomment-48818326 only if that is consistent with other parts of the language. (ie: arguments type restrictions, casts, etc). Otherwise a compatible duck type return value will not work as an argument later.

Option 2 plays nicer with mocks. But implementing that now is too much effort, will probably slow the compiler. Is sound to start with option 1 and then migrate to 2 if, again, the effect of that is spread in the whole language.

Currently you can trust the type contract a base class states, in neither explicit subclass nor ducktype behavior.

PR #6771, opened 2018-09-22
disallow return types on abstract methods

damaxwell:
Addresses #3546

This is a breaking language change -- return types on abstract methods will need to be removed before code will compile.  These had no effect anyways, so after removal there should be no impacts. Nevertheless, breaking changes are an irritation.

The standard library had a handful of examples where abstract methods had return types.  For the most part the hoped-for return types were noted in the documentation instead.

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -844,8 +844,8 @@ module Crystal
     it_parses "def foo : Int32\n1\nend", Def.new("foo", body: 1.int32, return_type: "Int32".path)
     it_parses "def foo(x) : Int32\n1\nend", Def.new("foo", args: ["x".arg], body: 1.int32, return_type: "Int32".path)
 
-    it_parses "abstract def foo : Int32", Def.new("foo", return_type: "Int32".path, abstract: true)
-    it_parses "abstract def foo(x) : Int32", Def.new("foo", args: ["x".arg], return_type: "Int32".path, abstract: true)
+    assert_syntax_error "abstract def foo : Int32"
+    assert_syntax_error "abstract def foo(x) : Int32"
 
     it_parses "{% for x in y %}body{% end %}", MacroFor.new(["x".var], "y".var, "body".macro_literal)
     it_parses "{% if x %}body{% end %}", MacroIf.new("x".var, "body".macro_literal)
@@ -1665,7 +1665,6 @@ module Crystal
       assert_end_location "def foo; 1; end"
       assert_end_location "def foo; rescue ex; end"
       assert_end_location "abstract def foo"
-      assert_end_location "abstract def foo : Int32"
       assert_end_location "begin; 1; end"
       assert_end_location "class Foo; end"
       assert_end_location "struct Foo; end"
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -3366,6 +3366,9 @@ module Crystal
       end
 
       if @token.type == :":"
+        if is_abstract
+          raise "abstract methods cannot specify a return type"
+        end
         next_token_skip_space
         return_type = parse_single_type
         end_location = return_type.end_location
--- a/src/file/info.cr
+++ b/src/file/info.cr
@@ -78,32 +78,32 @@ class File
   # A `File::Info` contains metadata regarding a file. It is returned by
   # `File.info`, and `File#info`.
   abstract struct Info
-    # Size of the file, in bytes.
-    abstract def size : UInt64
+    # Size (`UInt64`) of the file, in bytes.
+    abstract def size
 
-    # The permissions of the file.
-    abstract def permissions : Permissions
+    # The `Permissions` of the file.
+    abstract def permissions
 
-    # The type of the file.
-    abstract def type : Type
+    # The `Type` of the file.
+    abstract def type
 
-    # The special flags this file has set.
-    abstract def flags : Flags
+    # The special `Flags` this file has set.
+    abstract def flags
 
-    # The last time this file was modified.
-    abstract def modification_time : Time
+    # The last `Time` this file was modified.
+    abstract def modification_time
 
-    # The user ID of the file's owner.
-    abstract def owner : UInt32
+    # The user ID (`UInt32`) of the file's owner.
+    abstract def owner
 
-    # The group ID that the file belongs to.
-    abstract def group : UInt32
+    # The group ID (`UInt32`) that the file belongs to.
+    abstract def group
 
     # Returns true if this `Info` and *other* are of the same file.
     #
     # On unix, this compares device and inode fields, and will compare equal for
     # hard linked files.
-    abstract def same_file?(other : File::Info) : Bool
+    abstract def same_file?(other : File::Info)
 
     # Returns true if this `Info` represents a standard file. Shortcut for
     # `type.file?`.
--- a/src/io.cr
+++ b/src/io.cr
@@ -109,7 +109,7 @@ abstract class IO
   # io.write(slice)
   # io.to_s # => "abcd"
   # ```
-  abstract def write(slice : Bytes) : Nil
+  abstract def write(slice : Bytes)
 
   # Closes this `IO`.
   #
--- a/src/random.cr
+++ b/src/random.cr
@@ -62,11 +62,11 @@ module Random
     PCG32.new
   end
 
-  # Generates a random unsigned integer.
+  # Generates a random unsigned integer (`UInt`).
   #
   # The integers must be uniformly distributed between `0` and
   # the maximal value for the chosen type.
-  abstract def next_u : UInt
+  abstract def next_u
 
   # Generates a random `Bool`.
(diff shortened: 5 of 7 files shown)


Sija:
How hard would be to do it other way around - that is to implement 'em, instead of removing?

asterite:
Sorry, let's not work on stuff that we'll later have to revert. Someone should implement this correctly, it's not that hard (just check the return type is `AbstractDefChecker`).

damaxwell:
Thanks for reopening @asterite.  I wasn't able to reply sooner because comments were shut down to outsiders for a bit.

I don't feel strongly about this PR, and had offered it only because I had the impression that @RX14 had green-lighted it in the conversation over at #3546.

@Sija, in fact I had thought when I was writing the PR to try my hand at implementing return type restrictions (how hard can it be?).  But I quickly decided there were some subtleties that required language semantic decisions.

Superclasses currently don't impose restrictions on their children, and the return type of a method defined in a superclass is determined by a global analysis of the whole program if the leaf type of the receiver isn't known by the compiler.  With that in mind, I'm not sure what

```crystal
abstract struct Foo
  abstract def foo( x : (String | Int32) ) : Bool
  abstract def bar( y : Int ) : String
end
```

is supposed to mean.   Does the following comply?

```crystal
struct FooChild < Foo
  def foo( x : String )
    return 3.14159
  end

  def foo( x : Int32 )
    return 3.14159
  end

  def foo( x : (String | Int32) ) : Bool
    return True
  end

 def bar( y : Int ) : String
    return "pi"
 end

 def bar( y : Int32 ) : Float64
    return 2.7
 end
```

`FooChild` implements all the abstract methods needed by `Foo`. But as far as I can tell, `foo` in `FooChild` with the same type signature as in `Foo` will never be called. 

And although `bar` when called with an `Int` will sometimes return a `String`, sometimes it won't.  I.e., because return types can widen even as the argument types become more specific, the subclass can meet the return type requirements of the abstract method and yet undermine them entirely.

Implementing this version of return type restriction would be fairly easy, I think.  The place to check is in `AbstractDefChecker.cr`, and I have a branch that does this, it just needs testing.  But it wasn't clear to me that this was a useful property for the language to have.  There is really no guarantee being provided by the compiler about the return type.  If you are defining an interface via abstract methods, you can't be certain you'll actually be getting the desired return types.

So maybe the enforcement rule would have to be "if a method of a subclass has the same name as an abstract superclass method, and if the argument types of the method are subtypes of the abstract method's arguments, then the declared return type has to be a subtype of the abstract method return type (if present)." But this is essentially the reverse policy to return types for methods of subclasses in general and #160 ought to be revisited at the same time.  And there are still complications to worry about.  E.g.

```crystal
abstract struct Foo
  abstract def foo( (Int32 | String) ) : Bool
  abstract def foo( Int32 ) : Float64
end
```

To analyse compliance for the first version of `foo` you need to be aware of the second version of `foo` which effectively overrides it for some arguments.  Currently, compliance is verified locally: for each abstract method, for each child class, verify that there is at least one method that has arguments at least as general as the abstract method's. That's a lot easier.

And then there are generics and splats to worry about.

So I'd come to the conclusion that maybe this wasn't so easy.  I'd be interested in hearing concretely what implementing return types for abstract methods is supposed to do.  Maybe I could give a hand with it, but without the details I'm not sure.

And I agree with the sentiment that  if return type enforcement were going to be implemented any time soon, then it's certainly not worth the exercise of removing the syntax, irritating end users, and then adding a shiny new feature.  But #3546 has been open for a while now, it's had a duplicate posting as an issue (#5275), and I ran into it as a surprise independently.  If it isn't going to be looked at soon, then maybe it's better to make the compiler more honest, even if it makes it a little less likable.

Sija:
@damaxwell Thank you for such a detailed explanation! Fixing that seems to be above "my paygrade", but maybe someone like our compiler ninja @MakeNowJust could tackle that? ;)

damaxwell:
Yup.

I think this is the key question. Given

```crystal
abstract class A 
  abstract def foo( x : B ) : C
end
```

should there be a guarantee that if `AChild` is a subclass of `A`, and its `foo` method is called with a variable that is a subtype of `B`, then the returned value is a subtype of `C`?

If the answer is "yes", I think it's a big change for the language (possibly confusing in light of #160), but the task to accomplish is clear.

If the answer is "no", then what purpose does the return type serve?

RX14:
The contract for `foo` is that any type that passes the `Int` type restriction when passed to the method must end up returning `Int`. I don't see any way of checking that other than typing the function for all the millions of possible combinations of parameter types, which is infeasible.

However, annotating the return types of abstract methods is really useful for documentation.

Perhaps this check could be done at instantiation time. When instantiating a method:
1. Check if the method has the same name as an abstract method in a superclass or module, i.e. if it has *any* chance of implementing the abstract interface.
2. For all abstract overloads found, check if the specific argument types *of this instantiation* match the argument prototype of the abstract method.
3. Type the method as we were going to do anyway.
4. Now we have the return type, check it against the return type annotation of the current method, and *all* abstract overloads which matched.

Does this sound like a workable algorithm?

It has the same downside of ordinary method type restrictions, in that it only works for all the instantiations actually used, instead of all the possible instantiations.

asterite:
Well, abstract methods started out as just being documentation. I would personally leave them as documentation until someone figures out what's the **entire** semantic of overrides, overloads, redefinitions, etc. (this isn't just a problem with return types in abstract methods, how the language works in this regards is not documented anywhere, and the only source of truth is the compiler's code, which has many edge cases and bugs). So I'd start by defining the spec of the language, then fixing the current implementation toward it.

asterite:
By the way, I don't see what's the problem with this:

```crystal
abstract struct Foo
  abstract def foo( x : Int ) : String
  abstract def foo( x : Int32 ) : Bool
end
```

They are two different overloads.

damaxwell:
No problem intrinsically.  But I think it was ruled out by @RX14's proposal.

RX14:
okay yeah, the most specific overload then

damaxwell:
Maybe this is the wrong place to have this conversation, but since it's already underway I'll leave a couple of thoughts in case they are helpful when the core team decides it's time to address these issues.

@rx14's proposal implies 
```crystal
abstract class Foo
  abstract foo( x : Int ) : String
end
```
creates a contract that subclasses must fulfill.  For this contract to be meaningful, I think you need to then outlaw a subsequent
```crystal
abstract class FooMiddle < Foo
  abstract foo( x : Int32 ) : Bool
end
```
Otherwise, whatever contract was supposed to be enforced can be undone.  

On the other hand,
```crystal
abstract class Foo
  abstract foo( x : Int ) : String
  abstract foo( x : Int32 ) : Bool
end
```
is fine; there is no contract subversion since the same class is drawing up the contracts simultaneously and knows about their interaction.

-------------------------------------

Alternatively, I think it would be consistent with the current language design to take the stance that abstract methods simply don't have return types.  A concrete method in Crystal can specify a return type because there is something to verify for a particular implementation.  But an abstract method is a proxy for a whole family of methods and Crystal's unusual permissiveness with regards to override return types is not compatible with return type specification for abstract methods.

damaxwell:
The current state of affairs leans towards the principle of maximal astonishment. In
```crystal
 abstract foo( x : Int) : Zork
```
one type restriction is so strong that it is enforced by the compiler even if the method is never called, and the other is so weak that it is ignored completely and the type doesn't even have to exist.

The `: Zork` is a comment masquerading as code, but I doubt that folks generally know this.  Cheekily, the comment is saying "When I designed this method, I was hoping that when you implement it, you might opt to return this type.  My code that calls the method might assume that you do, and if you happen not to, it's uncertain what will occur.  Duck typing might work in your favor, though, so things might work out just fine.  Or not; there might be a subtle bug. Best of luck to you.  P.S. The compiler is pretty laid back and won't warn you if you decide to return something else." 😉 

If this PR is closed without merging, and there isn't impetus to @asterite's call for someone to figure out "what's the **entire** semantic of overrides, overloads, redefinitions" and formalize the language spec, would there be warmth instead to a PR against `crystal-book` to make clear that the return type of an abstract method has zero enforcement, but will appear in docs?

dscottboggs:
Personally, I don't think having return types as documentation is semantically a good idea, even if it's well documented in the Book.

Why not just disallow specifying a return type in the case that you want to have more than one, or using a Union? With the first suggestion, specifying both of the following in the same abstract class would be a compile time error.

```crystal
abstract def foo( a : T ) U
abstract def foo( a : T ) V
```
The compiler could simply say "return type restriction overloads are not allowed in abstract definitions" and that solves the problem.

With the second, it would certainly be more of a complicated implementation, but the semantics would be that a type union like:
```crystal
abstract def foo( a : T ) U | V
```
Would allow a subclass to define foo as returning *either* `U` or `V`, but not both.

As far as the special case of the Int type, perhaps that is a separate semantic issue. Perhaps there need to be special rules regarding the handling of Int/Int32/Int64 overloads.

Blacksmoke16:
This can probably be closed since https://github.com/crystal-lang/crystal/pull/7956 is merged.

PR #6778, opened 2018-09-23
Use the third person in method docs

r00ster91:
There are some methods left that don't use the third person.

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -190,7 +190,7 @@ class Array(T)
     size <=> other.size
   end
 
-  # Set intersection: returns a new `Array` containing elements common to `self`
+  # Sets intersection: returns a new `Array` containing elements common to `self`
   # and *other*, excluding any duplicates. The order is preserved from `self`.
   #
   # ```
@@ -219,7 +219,7 @@ class Array(T)
     end
   end
 
-  # Set union: returns a new `Array` by joining `self` with *other*, excluding
+  # Sets union: returns a new `Array` by joining `self` with *other*, excluding
   # any duplicates, and preserving the order from `self`.
   #
   # ```
--- a/src/array.cr
+++ b/src/array.cr
@@ -190,7 +190,7 @@ class Array(T)
     size <=> other.size
   end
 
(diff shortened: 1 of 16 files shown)


vladfaust:
In my projects I always use no-person references in the method introduction and then the third person in description. E.g:

```crystal
# Set foo to bar. May raise. Uploads to server.
def foo
end
```

So in the docs summary it says `Set foo to bar` and in the full description there is a third person reference to the described method.

Which approach is right?

r00ster91:
According to [Documenting code](https://crystal-lang.org/docs/conventions/documenting_code.html) you should always document methods in the third person:
> - Use the third person: `Returns the number of horns this unicorn has` instead of `Return the number of horns this unicorn has`.

src/array.cr
```diff
@@ -190,7 +190,7 @@ class Array(T)
     size <=> other.size
   end
 
-  # Set intersection: returns a new `Array` containing elements common to `self`
+  # Sets intersection: returns a new `Array` containing elements common to `self`
```
z64: This is actually a noun, not a verb. ("Set intersection", as in the intersection of two sets)

Same below on the next diff line for this file

src/compiler/crystal/compiler.cr
```diff
@@ -83,7 +83,7 @@ module Crystal
     # one LLVM module is created for each type in a program.
     property? single_module = false
 
-    # Set to a `ProgressTracker` object which tracks compilation progress.
+    # Sets to a `ProgressTracker` object which tracks compilation progress.
```
RX14: This was correct before, although badly worded (my docs, sorry)

src/compiler/crystal/program.cr
```diff
@@ -110,7 +110,7 @@ module Crystal
     # The main filename of this program
     property filename : String?
 
-    # Set to a `ProgressTracker` object which tracks compilation progress.
+    # Sets to a `ProgressTracker` object which tracks compilation progress.
```
RX14: ditto

Issue #3973, opened 2017-02-02
Undeterministic builds


theodorton:
It looks like the compiler does not produce deterministic builds / (i.e. building the identical source files produce different binaries).

It seems to happen even for the `*.o` files that are used when cross-compiling:

```
⟩ crystal build --cross-compile src/cli.cr
⟩ mv cli.o cli.previousbuild.o
⟩ crystal build --cross-compile src/cli.cr
⟩ md5 cli*
MD5 (cli.o) = f6fd2cf2f5ef7b777b0e2b1f870848fa
MD5 (cli.previousbuild.o) = b520abab295c1e86528cdc9ff9c8b69
```

I'm guessing this is expected, so my question is if deterministic builds are on the roadmap or if it might be hard to do due to constraints from LLVM. 

## Crystal version

```
⟩ crystal -v
Crystal 0.20.5 (2017-01-28)
```

theodorton:
Figured it out. Looks like you can get deterministic builds by passing `--no-debug` to the compiler.

RX14:
Looks related to #3831 and #3960

theodorton:
Current state with 0.23.1 is:

1. Passing `--no-debug` causes the build to not include useful debugging information (like line numbers).
2. Passing `--single-module` produces deterministic build, but I'm not sure what this option represents. There's not a lot of documentation regarding this, and even searching for `llvm single module` doesn't seem to return anything relevant.

I would still like to ask if deterministic builds with debug information should be the expected scenario. I don't think build artifacts should change unless there are changes to the source code or the environment.

akzhan:
Please upgrade to 0.24.0, it works fine with debug info for me.

RX14:
@theodorton by default, non-release builds of crystal produce a seperate llvm module for each class, so that we can run the LLVM codegen step in parallel on 4 cores. `--single-module` disables that and makes your debug builds slower to compile. But I guess if you want reproducible builds, then that's an OK tradeoff.

PR #6788, opened 2018-09-24
Add support for specifying the build date

peterhoeg:
Currently the compiler gets the compilation date embedded into it - instead allow a date to be specified (would typically be the release date) so that subsequent compilations would not be different.

Further to https://github.com/crystal-lang/crystal/issues/3973

Do you still require specs for something this small?

Tangibly, this will be used by https://github.com/NixOS/nixpkgs/pull/47166 instead of the ugly replacement I'm doing over there.

### DIFF
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -54,7 +54,7 @@ module Crystal
     end
 
     def self.date
-      {{ `date "+%Y-%m-%d"`.stringify.chomp }}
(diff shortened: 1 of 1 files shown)


bew:
Why would you want to set a specific build date?

peterhoeg:
To have a more deterministic build. On NixOS we generally set the date in the build environment to the start of the unix epoch.

jhass:
Should this be `CRYSTAL_CONFIG_BUILD_DATE`?

peterhoeg:
I have changed the variable name to @jhass's suggestion.

I noticed that CircleCI fails on Darwin with this change. Is that expected? Also, do you require a test?

PR #6793, opened 2018-09-25
Fsync

carlhoerberg:
Implementing fsync (and fdatasync) (fflush in windows) in the `File` class.

The spec goes through even with a simple `File#flush`, suggestions on how to test it better are welcome. 

References:
https://linux.die.net/man/2/fsync
https://linux.die.net/man/2/fdatasync
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fflush?view=vs-2017
https://ruby-doc.org/core-2.5.1/IO.html#method-i-fsync

### DIFF
--- a/src/crystal/system/unix/file.cr
+++ b/src/crystal/system/unix/file.cr
@@ -158,4 +158,18 @@ module Crystal::System::File
 
     nil
   end
+
+  def fsync : Nil
+    flush
+    if LibC.fsync(fd) != 0
+      raise Errno.new("fsync")
+    end
+  end
+
+  def fdatasync : Nil
+    flush
+    if LibC.fdatasync(fd) != 0
+      raise Errno.new("fsync")
+    end
+  end
 end
--- a/src/crystal/system/unix/file.cr
+++ b/src/crystal/system/unix/file.cr
@@ -169,7 +169,7 @@ module Crystal::System::File
   def fdatasync : Nil
     flush
     if LibC.fdatasync(fd) != 0
-      raise Errno.new("fsync")
+      raise Errno.new("fdatasync")
     end
   end
 end
--- a/src/crystal/system/unix/file.cr
+++ b/src/crystal/system/unix/file.cr
@@ -159,17 +159,15 @@ module Crystal::System::File
     nil
   end
 
-  def fsync : Nil
-    flush
-    if LibC.fsync(fd) != 0
-      raise Errno.new("fsync")
-    end
-  end
-
-  def fdatasync : Nil
-    flush
-    if LibC.fdatasync(fd) != 0
-      raise Errno.new("fdatasync")
+  private def system_fsync(flush_metadata = true) : Nil
+    if flush_metadata
+      if LibC.fsync(fd) != 0
+        raise Errno.new("fsync")
+      end
+    else
+      if LibC.fdatasync(fd) != 0
+        raise Errno.new("fdatasync")
+      end
     end
   end
 end
--- a/src/file.cr
+++ b/src/file.cr
@@ -838,15 +838,14 @@ class File < IO::FileDescriptor
     system_truncate(size)
   end
 
-  # Flushes all File's page cache to the disk device so that
+  # Flushes all data written to this File to the disk device so that
   # all changed information can be retrieved even if the system
   # crashes or is rebooted. The call blocks until the device reports that
   # the transfer has completed.
-  # As well as flushing the file data, fsync also flushes the metadata
-  # information associated with the file. To reduce disk activity for
-  # applications that do not require all metadata to be synchronized
-  # with the disk the parameter *flush_metadata* can be set to false, then
-  # the syscall *fdatasync* is called instead.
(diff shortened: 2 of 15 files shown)


src/crystal/system/unix/file.cr
```diff
@@ -158,4 +158,18 @@ module Crystal::System::File
...
+    flush
+    if LibC.fsync(fd) != 0
+      raise Errno.new("fsync")
+    end
+  end
+
+  def fdatasync : Nil
+    flush
+    if LibC.fdatasync(fd) != 0
+      raise Errno.new("fsync")
```
Sija: `fsync` -> `fdatasync`

src/lib_c/x86_64-windows-msvc/c/io.cr
```diff
@@ -15,4 +15,5 @@ lib LibC
   fun _chsize(fd : Int, size : Long) : Int
   fun _get_osfhandle(fd : Int) : IntPtrT
   fun _pipe(pfds : Int*, psize : UInt, textmode : Int) : Int
+  fun fflush(fd : Int) : Int
```
RX14: `fflush` takes a `FILE*` not an fd. `FlushFileBuffers` is the correct function and takes a `HANDLE`.

RX14: (Although I messed up and made the `HANDLE` hard to get at here, so perhaps best to leave it unimplemented on windows for now)

carlhoerberg: ok! removing it and raising `NotImplementedError`

carlhoerberg:
Ruby have them distinct (if that's any indicator), and the "closer to metal" the better i feel like at least. For anyone that knows what fdatasync is and want to do it, and search the docs and nothing come up might wonder why it's not implemented, unless they search the source code of course.

RX14:
fdatasync isn't a thing in windows, so fdatasync would just call fsync. I'd be fine with having it as a parameter of fsync.

carlhoerberg:
That's true, but also "fsync" is not a thing in windows either so I don't know about that argument. Yes, we can call fsync if fdatasync is not supported by the OS (can be more than just windows). I for one very much appreciate closeness to the actual posix syscalls and less arbitrarily named flags. Is it a goal to minimise number of methods?

ysbaddaden:
I also like being close to POSIX, but Crystal aims to be a little more intuitive and simplify such details —while still having them accessible when possible.

A real thing is that we don't want aliases. Since `fdatasync` has no equivalent in Windows, `fdatasync` and `fsync` would end up being exactly the same. Hence my proposal to have a single `fsync` method with a named parameter that will be used on POSIX systems, and overlooked on Windows with a documentation to explain that.

RX14:
anyone searching for fdatasync will quickly fall back to searching for fsync, and then find the parameter documented.

ysbaddaden:
Note that POSIX AIO has a single [`aio_fsync`](http://man7.org/linux/man-pages/man3/aio_fsync.3.html) function that can be passed `O_SYNC` or `O_DSYNC` meaning respectively `fsync` or `fdatasync`.

carlhoerberg:
Updated with docs and a single method with a `flush_metadata` parameter

src/file.cr
```diff
@@ -838,6 +838,20 @@ class File < IO::FileDescriptor
     system_truncate(size)
   end
 
+  # Flushes all File's page cache to the disk device so that
```
RX14: Perhaps?

> Flushes all data written to this `File` to the disk device so that...

src/file.cr
```diff
@@ -838,6 +838,20 @@ class File < IO::FileDescriptor
...
+  # all changed information can be retrieved even if the system
+  # crashes or is rebooted. The call blocks until the device reports that
+  # the transfer has completed.
+  # As well as flushing the file data, fsync also flushes the metadata
+  # information associated with the file. To reduce disk activity for
```
RX14: This is mainly mtime, right? Perhaps mention that.

carlhoerberg: atime, mtime yes. from linux docs:

> fdatasync() is similar to fsync(), but does not flush modified metadata unless that metadata is needed in order to allow a subsequent data retrieval to be correctly handled. For example, changes to st_atime or st_mtime (respectively, time of last access and time of last modification; see stat(2)) do not require flushing because they are not necessary for a subsequent data read to be handled correctly. On the other hand, a change to the file size (st_size, as made by say ftruncate(2)), would require a metadata flush.

RX14:
Perhaps `flush_metadata` should even be false by default...

carlhoerberg:
Don't you think it would be a bit confusing for a method called `fsync` to actually do a `fdatasync` by default? 

On another note, it's only when doing small rewrites of an existing file that fdatasync has an advantage over fsync (about 30% faster when doing 1KB rewrites). Doing append only writes or large rewrites (~1MB) makes fsync and fdatasync equally fast.

PR #6873, opened 2018-09-25
Implement File#eof

Heaven31415:
It seems that right now you cannot ask directly whether you are positioned at the end of file, this PR changes that situation.

### DIFF
--- a/src/file.cr
+++ b/src/file.cr
@@ -830,7 +830,12 @@ class File < IO::FileDescriptor
   def size
     info.size
   end
-
+  
+  # Checks whether end of file is reached.
+  def eof?
+    pos >= info.size
+  end
+  
   # Truncates the file to the specified *size*. Requires that the current file is opened
   # for writing.
   def truncate(size = 0) : Nil
--- a/src/file.cr
+++ b/src/file.cr
@@ -830,12 +830,12 @@ class File < IO::FileDescriptor
   def size
     info.size
   end
-  
+
   # Checks whether end of file is reached.
   def eof?
     pos >= info.size
   end
-  
+
   # Truncates the file to the specified *size*. Requires that the current file is opened
   # for writing.
   def truncate(size = 0) : Nil


RX14:
This works only for files, not all file descriptors, and is also racey - `File#eof?` doesn't neccesarily mean a subsequent `File#read` returns 0.

What's the usecase instead of simply attempting a read?

Heaven31415:
> This works only for files, not all file descriptors, and is also racey - `File#eof?` doesn't neccesarily mean a subsequent `File#read` returns 0.
> 
> What's the usecase instead of simply attempting a read?

How should I rewrite it so File#read will always return 0 when File#eof? returns true?

I don't see any other usecases, but isn't this enough? It's seems a little redundant and overcomplicated to call File#read to know whether I have reached end of file.

RX14:
@Heaven31415 there's no way to rewrite it like that, since two processed can be modifying the same file at the same time.

Why do you need to know whether you have reached eof if you're not calling read anyway? And if you're calling read or using one of the many different read methods, they all signal EOF in their own way.

Do you have any code you can show were you need `eof?`

ysbaddaden:
We can't rely on the file size to detect EOF, because another process could be saving the file at the same time that we're reading from it. This could lead to falsely detect an EOF when there was a race condition.

I can understand wanting to know whether we're at EOF or not, but I'd also like this to be available for all file descriptors. This could be implemented on `IO::Buffered` by checking whether there is some unread bytes in the buffer (trying to refill it when it's empty), but I fail to see how this could be implemented for `IO::FileDescriptor` directly (no buffer).

The idiomatic way in Crystal is to read until EOF, which can be either expected with `?` methods such as `#read?` or with an `IO::EOFError` exception with other methods such as `#read`.

I'd like an use-case too, that can't be solved with the current solutions we have in place.

asterite:
Also, this can be monkey-patched into your project without the core team having to maintain this functionality which doesn't seem to be quite correct.

Heaven31415:
> Why do you need to know whether you have reached eof if you're not calling read anyway?

@RX14 I just simply wanted to know whether there is still something to read. Of course I could use IO#read, but then I would need to create a slice. I could also use File#read_byte but it just doesn't say what I really want to do, check whether I'm at eof.

> Do you have any code you can show were you need eof?

I  have a .wav parser and its using eof to check whether there is still some chunk to read (sorry for longish example). I could replace eof with something else, but it's just the best way to describe what I'm trying to do.

```
def load_wav_file(filename : String) : WavSound?
  unless File.file? filename
    return nil
  end

  ext = File.extname(filename)
  if ext != ".wav" && ext != ".wave"
    return nil
  end

  file = File.open(filename, "r")
  header = uninitialized WavRIFFHeader
  format_chunk = uninitialized FormatChunk
  data_chunk = uninitialized DataChunk
  valid_format = false

  header.group_id = file.read_string(4)
  header.size = file.read_bytes(Int32, LE)
  header.riff_type = file.read_string(4)

  if header.group_id != "RIFF" || header.riff_type != "WAVE"
    return nil
  end

  while !file.eof?
    tag = file.read_string(4)
    file.seek(-4, IO::Seek::Current)

    case tag
    when "fmt "
      format_chunk.id = file.read_string(4)
      format_chunk.size = file.read_bytes(Int32, LE)

      pos = file.pos

      format_chunk.format_tag = file.read_bytes(Int16, LE)
      format_chunk.channels = file.read_bytes(UInt16, LE)
      format_chunk.samples_per_sec = file.read_bytes(UInt32, LE)
      format_chunk.avg_bytes_per_sec = file.read_bytes(UInt32, LE)
      format_chunk.block_align = file.read_bytes(UInt16, LE)
      format_chunk.bits_per_sample = file.read_bytes(UInt16, LE)

      valid_format = true

      if format_chunk.size > (file.pos - pos)
        file.skip(format_chunk.size - (file.pos - pos))
      end
    when "data"
      data_chunk.id = file.read_string(4)
      data_chunk.size = file.read_bytes(Int32, LE)
      data_chunk.data = Bytes.new(data_chunk.size)

      data = data_chunk.data.to_slice
      bytes_read = 0

      while bytes_read != data_chunk.size
        count = file.read(data)
        bytes_read += count
        data += count
      end
    else
      chunk_id = file.read_string(4)
      chunk_size = file.read_bytes(Int32, LE)
      file.skip(chunk_size)
    end
  end

  file.close

  return nil unless valid_format

  data = data_chunk.data.pointer(0).unsafe_as(Pointer(Void))
  size = data_chunk.data.size
  frequency = format_chunk.samples_per_sec

  WavSound.new(data, size, frequency)
end
```

> We can't rely on the file size to detect EOF, because another process could be saving the file at the same time that we're reading from it. This could lead to falsely detect an EOF when there was a race condition.

@ysbaddaden Sorry, I just wasn't aware of that when I was submitting PR.

> The idiomatic way in Crystal is to read until EOF, which can be either expected with ? methods such as #read? or with an IO::EOFError exception with other methods such as #read.

There is only one small problem with that, which may be a little unrelated to this PR. For example File#read_bytes and File#read_string (which I was using in my code) doesn't have ? counterparts. Also there is no File#read?. So you are almost forced to work with exceptions, only File#read_byte can help you.

> Also, this can be monkey-patched into your project without the core team having to maintain this functionality which doesn't seem to be quite correct.

@asterite Sure, it's now clear that this PR in current form is not a good move. I just wanted to have an easier and cleaner way to express that I'm at the end of file. Right now it seems that IO#read_byte would be the best way to go.

I don't think that I'm competent enough to propose anything better, I will leave to you to decide whether there should be another way to express eof.

asterite:
I think we just need `read_string?` here. Or you can read a slice of four bytes and create a string from there. If the slice isn't filled, you stop the loop. But eof doesn't seem to be the right solution, because you really want to know if there's more to read.

ysbaddaden:
Thanks for the use case and explanation! I guess we could have `#read_string?` and maybe `#read_bytes?`. I never needed them myself, but they could be useful.

RX14:
(`read_bytes?` too but thats harder given the amount of indirection)

Heaven31415:
My PR in current form obviously won't be merged and I don't intend to change it, so I'm closing it.

PR #6888, opened 2018-09-28
Remove comment after :nodoc: in `URI.escape_one`

r00ster91:
The `:nodoc:` is visible in the docs because of the comment under it: [`URI.unescape_one`](https://crystal-lang.org/api/0.26.1/URI.html#unescape_one%28string%2Cbytesize%2Ci%2Cbyte%2Cchar%2Cio%2Cplus_to_space%3Dfalse%2C%26block%29-class-method).

### DIFF
--- a/src/uri.cr
+++ b/src/uri.cr
@@ -348,7 +348,6 @@ class URI
   end
 
   # :nodoc:
-  # Unescapes one character. Private API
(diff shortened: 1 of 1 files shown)


Sija:
It shouldn't anymore, check with `master`.

r00ster91:
Oh didn't know this works now.

PR #6891, opened 2018-09-28
Add floor divisions operator

bcardiff:
Adds semantics to the `//` operator as integer division / floor division.

`a // b` will represent the mathematical value ⌊a ÷ b⌋  and be of the same type as `a`. This is consistent with other arithmetic operators (lhs operand type is preserved).

Follow up of #6470 as part of #2968.

After next release `/` will be able to be changed to float division for all types and use `//` where floor or integer division is needed.

### DIFF
--- a/spec/std/float_spec.cr
+++ b/spec/std/float_spec.cr
@@ -130,11 +130,13 @@ describe "Float" do
     end
 
     it "applies floor" do
-      (7.0 // 2.0).should eq(3)
-      (-7.0 // 2.0).should eq(-4)
+      (7.0 // 2.0).should eq(3.0)
+      (-7.0 // 2.0).should eq(-4.0)
 
-      (6.0 // 2.0).should eq(3)
-      (-6.0 // 2.0).should eq(-3)
+      (6.0 // 2.0).should eq(3.0)
+      (-6.0 // 2.0).should eq(-3.0)
+
+      (30.3 // 3.9).should eq(7.0)
     end
   end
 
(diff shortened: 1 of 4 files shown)


r00ster91:
Trying to format
```cr
struct Int
  def //
  end
end
```
gives a bug:
```
Error: couldn't format './cr.cr', please report a bug including the contents of it: https://github.com/crystal-lang/crystal/issues

expecting keyword end, not `/, `, at :4:8 (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???


```
I think `//` needs to be added in this tuple: [`formatter.cr:2171`](https://github.com/crystal-lang/crystal/blob/25de751e005d509192f46228ff6e5c4c8b3e424e/src/compiler/crystal/tools/formatter.cr#L2171) or maybe here: [`formatter.cr:2188`](https://github.com/crystal-lang/crystal/blob/25de751e005d509192f46228ff6e5c4c8b3e424e/src/compiler/crystal/tools/formatter.cr#L2188).

bcardiff:
Yes this is still WIP or blocked do to the formatted. I will check if that is enough. I’m not that familiar with the formatter still. Since parser and formatter use the lexer with its own control fow the state of the lexer evolves differently for the same input and is hard to keep them aligned i think.

bcardiff:
Rebased on to master. Ready for another review (while waiting for CI)

RX14:
Can we add a `-D` flag in 0.27.0 which will make `/` return float, to allow people to test their code with the new behaviour before 0.28.0?

PR #6894, opened 2018-09-30
Allow creating sockets from raw file descriptors

myfreeweb:
This is useful for [socket activation](https://github.com/myfreeweb/soad), sandboxes/fd-passing, etc.

### DIFF
--- a/src/socket.cr
+++ b/src/socket.cr
@@ -79,7 +79,8 @@ class Socket < IO
     end
   end
 
-  protected def initialize(@fd : Int32, @family, @type, @protocol = Protocol::IP, blocking = false)
+  # Creates a Socket from an existing socket file descriptor.
+  def initialize(@fd : Int32, @family, @type, @protocol = Protocol::IP, blocking = false)
     @closed = false
     init_close_on_exec(@fd)
 
--- a/src/socket/tcp_server.cr
+++ b/src/socket/tcp_server.cr
@@ -48,6 +48,11 @@ class TCPServer < TCPSocket
     end
   end
 
+  # Creates a TCPServer from an already configured raw file descriptor
+  def initialize(*, fd : Int32, family : Family = Family::INET)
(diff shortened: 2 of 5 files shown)


ysbaddaden:
Use case is valid, but CI is red.

straight-shoota:
Needs fix and maybe a spec for the described behaviour.

straight-shoota:
@myfreeweb Are you still working on this?

myfreeweb:
I don't understand why that test is failing..

straight-shoota:
@myfreeweb The added `TCPServer#initialize(fd)` shadows `TCPServer.new(port)` because both can be called with a single Int32 argument.

I'd suggest to make `fd` an required named argument: `def initialize(*, fd : Int32 ...` This should be changed uniquely in all fd constructors.

straight-shoota:
It's a shame that such a change breaks only one mostly unrelated SSL spec 😢

myfreeweb:
Passes now

straight-shoota:
Awesome :+1: Could you add a doc to the now public `Socket#initialize` as well?

myfreeweb:
done

src/socket.cr
```diff
@@ -79,7 +79,8 @@ class Socket < IO
     end
   end
 
-  protected def initialize(@fd : Int32, @family, @type, @protocol = Protocol::IP, blocking = false)
+  # Creates a Socket from an existing socket file descriptor.
+  def initialize(@fd : Int32, @family, @type, @protocol = Protocol::IP, blocking = false)
```
asterite: Shouldn't this also use named arguments like the other `initialize` methods in subtypes?

myfreeweb: There's no conflict on `Socket`.. named arguments are only really necessary for TCP

asterite: I see. I meant for consistency, but it's not a big deal.

PR #6897, opened 2018-10-01
Fiber/Scheduler refactorings

ysbaddaden:
This pull request is a collection of refactors of Fiber, Event, Scheduler and EventLoop (previously merged into Scheduler). This is meant as a first step to isolate the Scheduler from the rest (fibers, libevent) and centralize the scheduling logic to the Scheduler.

The proposed changes shouldn't break anything, and don't seem to have much impact on the overall performance —from my limited testing. They should, however, simplify experiments on Scheduler itself (e.g. multithreads).

Each commit tries to be a single change, with the commit message explaining what and why.


src/thread.cr
```diff
@@ -48,51 +52,49 @@ class Thread
...
-  @@threads = Set(Thread).new
+  {% if flag?(:android) || flag?(:openbsd) %}
+    # no thread local storage (TLS) for OpenBSD or Android,
+    # we use pthread's specific storage (TSS) instead:
+    @@current_key = uninitialized LibC::PthreadKeyT
 
-  {% if flag?(:openbsd) %}
-    @@main = new
+    ret = LibC.pthread_key_create(pointerof(@@current_key), nil)
+    raise Errno.new("pthread_key_create") unless ret == 0
```
RX14: Perhaps make this an explicit initializer:

```cr
@@current_key = begin
  ret = LibC.pthread_key_create(out current_key, nil)
  raise Errno.new("pthread_key_create") unless ret == 0
  current_key
end
```

src/fiber.cr
```diff
@@ -116,32 +116,29 @@ class Fiber
...
-  def sleep(time : Number)
-    sleep(time.seconds)
+  # :nodoc:
+  def resume_event
+    @resume_event ||= Crystal::EventLoop.create_resume_event(self)
   end
 
-  def yield
-    sleep(0)
+  def self.sleep(time : Number)
```
RX14: Perhaps remove these methods and use `Crystal::Scheduler` directly from the top-level sleep?

ysbaddaden:
I removed `Fiber.sleep` and used an explicit initializer for `Thread.@@current_key`.

src/crystal/event_loop.cr
```diff
@@ -1,25 +1,19 @@
-require "event"
+require "./event"
 
-# :nodoc:
-class Scheduler
-  @@runnables = Deque(Fiber).new
-  @@eb = Event::Base.new
+class Crystal::EventLoop
```
straight-shoota: Why not a `module`?

ysbaddaden: Probably because Scheduler used to be a class with only class methods too, and I overlooked that.

bew: would it be possible to keep it as a class, and make all current class methods → instance methods, and ultimately allow to make multiple event loops? (no idea what usecase there could possibly be for that though)

src/crystal/scheduler.cr
```diff
@@ -0,0 +1,44 @@
+require "./event_loop"
+require "thread"
+
+class Thread
+  # :nodoc:
+  def scheduler
+    @scheduler ||= Crystal::Scheduler.new
```
bcardiff: Currently the scheduler is global. In order to keep the current behavior this should be a class var.
I think there will be multiple schedulers with MT, so the accessing _the scheduler_ via the current thread is ok.

src/fiber/aarch64.cr
```diff
@@ -0,0 +1,65 @@
+{% skip_file unless flag?(:aarch64) %}
+
+class Fiber
+  # :nodoc:
+  def makecontext(stack_ptr, fiber_main) : Void*
+    # align the stack pointer to 16 bytes
+    stack_ptr = Pointer(Void*).new(stack_ptr.address & ~0x0f_u64)
```
bcardiff: The alignment was done for every arch before. Wouldn't be better to keep that before the call to `makecontext`?

It would be great to leave somewhere a doc related to `makecontext` and `swapcontext` (and their required annotations).

src/thread.cr
```diff
@@ -48,51 +52,49 @@ class Thread
...
+
+    protected def self.current=(@@current : Thread) : Thread
+    end
   {% end %}
 
   protected def start
-    {% unless flag?(:openbsd) %}
-    @@current = self
-    {% end %}
+    Thread.current = self
```
bcardiff: For the main thread the `#start` is never called, so setting the current thread by this line wont happen.
In my WIP for mt-gc at https://github.com/bcardiff/crystal/commit/aedde16f9d012a8d8649b29974e9eec01cb29870 I changed how main fiber / thread will initialize.

On platforms with `ThreadLocal` there will be 2 instances of Thread objects with the current code since the `@@current = new` will be evaluated again.

For now, performing `Thread.current = self` in both constructors seem enough.

src/thread.cr
```diff
@@ -6,6 +6,10 @@ class Thread
   # Don't use this class, it is used internally by the event scheduler.
   # Use spawn and channels instead.
 
+  # all threads, so the GC can see them (GC doesn't scan thread locals)
+  @@threads = Set(Thread).new
```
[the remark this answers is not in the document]
ysbaddaden: I just moved the line, wondered avout changing it and finally kept it. An Array would be simpler and enough.

src/crystal/event_loop.cr
ysbaddaden: I don't have any plans to experiment with many event loops. Quite the opposite.

src/crystal/scheduler.cr
ysbaddaden: Is there a downside to attach the scheduler on the thread now?

src/fiber/aarch64.cr
ysbaddaden: It's still done for every arch but that's probably not required on all architectures, so evolutions can choose to align or not. Is this required for something (like BDWGC)?

Yes, there should be some documentation for the arch-specific methods (that take their names off the deprecated POSIX makecontext / setcontext / swapcontext). I just wasn't exactly sure where to put them. Maybe in `fiber.cr` itself?

bcardiff: I am not sure it is a requirement. But it itches me to repeat that "cryptic" call in every method.
Documentation in `fiber.cr` seems good.

src/thread.cr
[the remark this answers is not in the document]
bcardiff: We can remove the Set later for sure, but I found that the linked list is enough. If we guarantee there is only one Thread object per thread then we don't even need to lookup to remove it from the list.

Bonus point the thread safe linked list pointer implementation from fibers can be extracted and reused.

I think the set was added for openbsd, hence maybe I was missing something regarding why Set was used in the first place.

src/crystal/scheduler.cr
```diff
@@ -4,14 +4,20 @@ require "thread"
 class Thread
   # :nodoc:
   def scheduler
-    @scheduler ||= Crystal::Scheduler.new
+    @scheduler ||= Crystal::Scheduler.new(Fiber.new)
```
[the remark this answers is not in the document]
bcardiff: But it's already deciding something regarding the multiple schedulers and we should not introduce such change here.

src/thread.cr
ysbaddaden: I now always call `self.current = new` in `Thread`, but maybe it would be better placed in `src/kernel.cr` (eventually).

ysbaddaden:
I pushed some fixups, including more thread refactorings:
- abstracts the new `Fiber::LinkedList` struct as a thread-safe generic `Thread::LinkedList(T)` and is now used to keep a list of thread objects (for the GC to find/mark them);
- adds and always initializes `Thread#fiber_main`, passed to the `Scheduler` associated to the thread.

Oops. I broke GC.

straight-shoota:
The documentation added for implementing `makecontext` and `swapcontext` shouldn't be part of the public API docs as the methods themselves are already `nodoc`. That's just internal documentation.

Maybe the the arch-specific implementations could be moved to the `Crystal::Fiber` namespace and included into `Fiber`, similar to other platform-specific implementations such as `File`.

RX14:
This looks good, although i'd like to re-review once all fixups have been applied

ysbaddaden:
@straight-shoota read the patch again, the docs for the context methods is attached to the `require` not the Fiber class, hence not public AFAIK.

src/crystal/scheduler.cr
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  protected def enqueue(fibers : Enumerable(Fiber)) : Nil
+    @runnables.concat fibers
+  end
+
+  protected def resume(fiber : Fiber) : Nil
+    current, @current = @current, fiber
+    GC.stack_bottom = fiber.@stack_bottom
+    Fiber.swapcontext(pointerof(current.@stack_top), fiber.@stack_top)
```
bcardiff: Before the PR the swapcontext use `pointerof(fiber.@stack_top)`  why is that not needed? I see no changes in the assembler code in `x86_64` for example.

src/fiber.cr
```diff
@@ -10,103 +44,77 @@ end
 class Fiber
   STACK_SIZE = 8 * 1024 * 1024
 
-  @@first_fiber : Fiber? = nil
-  @@last_fiber : Fiber? = nil
+  @@fibers = Thread::LinkedList(Fiber).new
```
bcardiff: I imagined that a LinkedList module can be included and make the class a singleton linked list across ivars. And encapsulate the definition of next/previous there.

src/fiber.cr
```diff
@@ -10,103 +44,77 @@ end
...
+    @@fibers.delete(fiber)
+  end
+
   def initialize(@name : String? = nil, &@proc : ->)
     @stack = Fiber.allocate_stack
     @stack_bottom = @stack + STACK_SIZE
+
     fiber_main = ->(f : Fiber) { f.run }
 
+    # ???
```
bcardiff: I think this is because the `@stack_bottom` points to 1 position after the valid region of the whole `@stack`. So is allowing the `stack_ptr` to start from "END-1"

src/thread.cr
```diff
@@ -6,92 +6,125 @@ class Thread
   # Don't use this class, it is used internally by the event scheduler.
   # Use spawn and channels instead.
 
-  @th : LibC::PthreadT?
+  # all thread objects, so the GC can see them (it doesn't scan thread locals)
```
bcardiff: boehm does support thread local storage. I thought the original set was added because openbsd didn't have thread local support.

src/thread.cr
```diff
@@ -6,92 +6,125 @@ class Thread
...
+  end
+
+  # Returns the Fiber representing the thread's main stack.
+  def main_fiber
+    @main_fiber.not_nil!
+  end
+
+  # :nodoc:
+  def scheduler
+    @scheduler ||= Crystal::Scheduler.new(main_fiber)
```
bcardiff: So, besides the refactor this PR will allow each thread to have a the current ST scheduler, but the GC support is not yet guaranteed. Ok, we can go in that direction as intermediate stage.

src/thread/linked_list.cr
```diff
@@ -0,0 +1,64 @@
...
+  # `#previous : T?` and `#next : T?` methods.
+  struct LinkedList(T)
+    @mutex = Thread::Mutex.new
+    @head : T?
+    @tail : T?
+
+    # Iterates the list without acquiring the lock, to avoid a deadlock in
+    # stop-the-world situations, where a paused thread could have acquired the
+    # lock to push/delete a node, while still being "safe" to iterate (but just
+    # in this situation).
```
bcardiff: "this situation" means that the world needs to be stoped, right?

src/crystal/scheduler.cr
ysbaddaden: Because the assembly needs a pointer to the `stack_top` pointer of the current fiber to write the current stack pointer to (i.e. `void**`), but we don't need to dereference the pointer to load the new `stack_top` pointer, and can just pass the pointer itself (i.e. `void*`).

AArch64 used that form, and I harmonized the calls.

src/thread.cr
ysbaddaden: I reworded the doc, but yes, that's what it said. Maybe we're wrong, yet there's nothing bad to keep the list of threads anyway :)

src/thread/linked_list.cr
ysbaddaden: Yes.

src/fiber.cr
ysbaddaden: You mean something like :
```crystal
module Thread::LinkedList(T)
  struct Root(T)
    def unsafe_each; end
    def push(node : T); end
    def delete(node : T); end
  end

  macro included
    class_getter all = Root(T).new
    property next : T?
    property previous : T?
  end
end

File: src/crystal/scheduler.cr (hunk 9 above)
Comment (RX14) [reply src/crystal/scheduler.cr]: The difference is the `movq ($1), %rsp` vs `movq $1, %rsp` in the asm

File: src/fiber.cr (hunk 11 above)
Comment (RX14) [reply src/fiber.cr]: Yes, because the stack grows down, so issuing a store to stack_bottom would store off the end of the stack.

File: src/fiber.cr (hunk 10 above)
Comment (bcardiff) [reply src/fiber.cr]: Even more flattened:

```
module Thread::LinkedList(T)
  macro included
    @@first : T?
    @@last : T?
    @next : T?
    @previous : T?

    def self.unsafe_each; end
    def self.push(node : T); end
    def self.delete(node : T); end
  end
end
```

Comment (RX14) [reply src/fiber.cr]: Please don't jump to macros on this, this is an internal struct and the way it is in this PR is just fine.

File: src/crystal/scheduler.cr (hunk 9 above)
Comment (bcardiff) [reply src/crystal/scheduler.cr]: Ok, thanks for clarifying

File: src/fiber.cr (hunk 10 above)
Comment (bcardiff) [reply src/fiber.cr]: Ok

File: src/fiber.cr (hunk 11 above)
Comment (bcardiff) [reply src/fiber.cr]: So I think adding a comment here and apply the fixups in the commits is what is pending to re-review and merge this PR. Right?

Then we can move to add some changes to support MT GC and other required changes for parallelism in others PR.

**Comment by ysbaddaden** [comment]
I'm squashing the fixups to have clean comments. That will take some time.

**Comment by ysbaddaden** [comment]
Squashed.

File: src/concurrent.cr [hunk 15]
```diff
@@ -12,15 +12,15 @@ def sleep(seconds : Number)
     raise ArgumentError.new "Sleep seconds must be positive"
   end
 
-  Fiber.sleep(seconds)
+  Crystal::Scheduler.sleep(seconds.seconds)
```
straight-shoota: `Crystal::Scheduler.sleep` look a bit off as it seems that the scheduler would sleep when it actually puts the current fiber to sleep. I don't have better suggestion at hand, just wanted to mention it.

RX14: `Crystal::Scheduler.sleep` is used only used in these two methods so i think its acceptable

ysbaddaden:
Fixed typos & rebased again.

PR #6899, opened 2018-10-01
Blur is only activated when modal is opened

girng:
Blur is only activated when modal is opened

Coincides with straight-shoota's request https://github.com/crystal-lang/crystal/pull/6764#issuecomment-423694501

### DIFF
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -553,6 +553,10 @@ span.flag.purple {
 .js-modal-visible .modal-background {
   display: flex;
 }
+.js-modal-visible .types-list,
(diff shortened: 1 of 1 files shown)


straight-shoota:
`transition: filter 200ms` removed in #6764 should also be added again. Only `filter: blur(0)` should've been removed. I assumed it was a no-op, but apparently, it isn't (at least not in all browsers).

RX14:
we don't need a transition or blur at all

j8r:
I suggest to do like [devdocs](https://devdocs.io/help#shortcuts) - when pressing `?`, the helping page is loaded with all the help, including keyboard shortcuts, how to search, search aliases.
Note that the page is loaded with `#shortcuts` to point directly to the shortcuts title.

PR #4707, opened 2017-07-12
Seperate Socket and FileDescriptor IO

RX14:
On some platforms - notably windows - socket descriptors are different from file
descriptors so it makes little sense for them to be shared under a common hierarchy.

This is done in 3 stages:
- Firstly I refactored out the common logic for dealing with nonblocking IO using `read`/`write` like functions from `IO::FileDescriptor` into `IO::Syscall`. It takes care of checking for EAGAIN and managing the lists of waiting fibers.
- Then, I removed the `edge_triggerable` option from `IO::FileDescriptor`, it doesn't seem to be used anywhere and it seems to me like it would cause hangs and other weird behaviour if used.
- Finally, I changed `Socket` not to inherit `IO::FileDescriptor`. This involved quite a lot of copy-pasting code, which seems like a step backwards, but it means we can refactor `File` and `Socket` to use `Crystal::System` separately. Once these refactors are done both `File` and `Socket` class hierarchies will likely end up much cleaner.

PR #6906, opened 2018-10-03
Refactor Socket to reuse Crystal::System::FileDescriptor

straight-shoota:
`Socket` wraps a file descriptor, similar to `IO::FileDescriptor` and many of methods for dealing with the basic libc file descriptor API have duplicate implementations in `Socket` and `Crystal::System::FileDescriptor`.  Thus I figure it makes sense to include the latter into `Socket` in order to reuse the same methods and instance variables.

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -1197,6 +1197,32 @@ describe "File" do
     end
   end
 
+  it "#close_on_exec?" do
+    File.open(datapath("test_file.txt")) do |file|
+      file.close_on_exec?.should be_true
+      file.close_on_exec = false
+      file.close_on_exec?.should be_false
+      file.close_on_exec = true
+      file.close_on_exec?.should be_true
+    end
+  end
+
+  it "#blocking?" do
+    File.open(datapath("test_file.txt")) do |file|
+      file.blocking.should be_true
+      file.blocking = false
+      file.blocking.should be_false
+      file.blocking = true
+      file.blocking.should be_true
+    end
+  end
+
+  it "#tty?" do
+    File.open(datapath("test_file.txt")) do |file|
+      file.tty?.should be_false
+    end
+  end
+
   describe ".match?" do
     it "matches basics" do
       File.match?("abc", "abc").should be_true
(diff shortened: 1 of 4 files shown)


asterite:
I feel we are running in circles here. Socket was refactored to not inherit from `IO::FileDescriptor` some time ago, I think because in Windows it's not a file descriptor (not sure that was the reason), but now we are adding (almost) this back.

asterite:
Specifically: https://github.com/crystal-lang/crystal/pull/4707

RX14:
Specifically: https://docs.microsoft.com/en-us/windows/desktop/winsock/socket-handles-2

> it is recommended that socket handles be used only with Winsock functions.

straight-shoota:
A `Socket` is not an `IO::FileDescriptor` but it can use some of the features provided by `Crystal::System::FileDescriptor`. They are really just internal helper methods, not a shared hierarchy. Users shouldn't know about `Crystal::System::FileDescriptor` anyway.

For windows support, we have to reorganize system specific implementations for `Socket` anyway. This will probably lead to `Crystal::System::Socket` anyway. Once we're there, we can further decide if and what connection that should have with `Crystal::System::FileDescriptor` on POSIX platforms. If we end up duplicating the shared methods there, I'm fine with that.

My goal is just to simplify `Socket` and move complexity brought by system specific implementations out of it. If you'd like to start refactoring to `Crystal::System::Socket` right away, I could just duplicate the necessary methods there and then we can decide later if it should stay separate from `Crystal::System::FileDescriptor`.

RX14:
@straight-shoota this commit adds more lines than it deletes, and the diff is under +-100 anyway. It's not worth merging this PR if we're not 100% sure that this won't just have to be reverted anyway.

straight-shoota:
@RX14 It adds about 50 lines of specs for previously unspecced methods. And I don't think it needs to be reverted. Almost all code removed in this PR should not stay in `Socket` anyway. The only thing that might change when going towards Windows support is replacing `include Crystal::System::FileDescriptor` with `include Crystal::System::Socket`.

ysbaddaden:
I'm fine with the change, but as stated by Ary this is going in circles and leads to nowhere. Let's add Windows support, see how things must be (re)organized, and then start refactoring Socket, with facts and requirements in place.

straight-shoota:
It's a little bit of a chicken-and-egg problem because I had planned to simplify `Socket` in order to refactor it and then be able to extract platform-specifics and add Windows support. But okay, I can leave it as is for now.

PR #6908, opened 2018-10-04
Add UNIX_EPOCH

j8r:
This adds a documented `Time::UNIX_EPOCH` constant
Some instance methods have also been replaced to ivars on which they were pointing to. This further improve the performance a bit.

Simple benchmark:
```crystal
require "benchmark"

TIME = Time.utc_now

struct Time
  UNIX_EPOCH = Time.utc 1970, 1, 1

  def to_unix_epoch
    (@seconds - UNIX_EPOCH.total_seconds).to_i64
  end
end

Benchmark.ips do |bm|
  bm.report("int") do
    TIME.epoch
  end
  bm.report("span") do
    TIME.to_unix_epoch
  end
end
```
Results:
```
 int 257.98M (  3.88ns) (± 3.85%)  0 B/op   1.45× slower
span 375.05M (  2.67ns) (± 3.43%)  0 B/op        fastest
```


src/time.cr
```diff
@@ -252,6 +252,13 @@ struct Time
   # :nodoc:
   DAYS_PER_4_YEARS = 365*4 + 1
 
+  # This constant is defined to be "1970-01-01 00:00:00 UTC" on all systems with respect to the system clock.
+  # Can be used to create a `Time::Span` that represents an unix epoch time duration.
+  # ```
```
Sija: Missing blank line above.

src/time.cr
```diff
@@ -252,6 +252,13 @@ struct Time
...
+  # Time.utc_now - Time::UNIX_EPOCH
+  # ```
+  UNIX_EPOCH = Time.utc 1970, 1, 1
+
   # :nodoc:
   UNIX_SECONDS = SECONDS_PER_DAY.to_i64 * (1969*365 + 1969/4 - 1969/100 + 1969/400)
```
bcardiff: if `UNIX_SECONDS` is no longer used and has a `:nodoc:` then we should remove it (or schedule to remove it)

j8r: I was planning to use exclusively `Time::UNIX_EPOCH.total_seconds`, but I've encounter issues; possibility chicken and egg ones, because `Time::UNIX_EPOCH` is a `Time` inside a `Time` object.
To sum-up, `UNIX_SECONDS` is still in use, for now.

straight-shoota:
1. Please remove unrelated refactoring edits. This makes it really hard to follow the actual changes. They can be introduced in a separate PR (but then *all* uses of `total_seconds` etc. should be changed, maybe the method even removed, see #6661).
2. You need to create the `UNIX_EPOCH` instance using the `utc(seconds, nanoseconds)` constructor with the value of `UNIX_SECONDS`. Then you can remove the latter.
3. The benchmark is very likely flawed (see https://github.com/crystal-lang/crystal/pull/6662#issuecomment-426639421 for details). I don't think there could be *any* legitimate explanation for `(@seconds - UNIX_EPOCH.total_seconds).to_i64` being 50% faster than `(total_seconds - UNIX_SECONDS).to_i64`.

straight-shoota:
Overall, I think it makes sense to replace `UNIX_SECONDS` with a public constant `UNIX_EPOCH`.

RX14:
@straight-shoota the benchmark is real, i performed it many times on a completely unloaded desktop PC and reproduced this (+- 0.02%!):

```
 int 319.32M (  3.13ns) (± 0.01%)  0 B/op   1.20× slower
span 383.16M (  2.61ns) (± 0.03%)  0 B/op        fastest
 int 319.32M (  3.13ns) (± 0.01%)  0 B/op   1.20× slower
span 383.18M (  2.61ns) (± 0.03%)  0 B/op        fastest
 int  319.3M (  3.13ns) (± 0.02%)  0 B/op   1.20× slower
span 383.16M (  2.61ns) (± 0.02%)  0 B/op        fastest
 int  319.3M (  3.13ns) (± 0.02%)  0 B/op   1.20× slower
span 383.16M (  2.61ns) (± 0.02%)  0 B/op        fastest
 int  319.3M (  3.13ns) (± 0.01%)  0 B/op   1.20× slower
span 383.16M (  2.61ns) (± 0.02%)  0 B/op        fastest
```


However, that does not mean that the results are valid. LLVM can see that the result of `TIME.epoch` and `TIME.to_unix_epoch` are unused, and can optimize parts out. What's optimized will depend on specific details of the LLVM IR.

Using this code to make sure that the result was used, I got these results:
```cr
require "benchmark"

TIME = Time.utc_now

struct Time
  UNIX_EPOCH = Time.utc 1970, 1, 1

  def to_unix_epoch
    (@seconds - UNIX_EPOCH.total_seconds).to_i64
  end
end

Benchmark.ips do |bm|
  x = 0
  bm.report("int") do
    x ^= TIME.epoch
  end

  y = 0
  bm.report("span") do
    y ^= TIME.to_unix_epoch
  end
end
```

```
 int 294.49M (   3.4ns) (± 0.29%)  0 B/op        fastest
span  273.7M (  3.65ns) (± 0.01%)  0 B/op   1.08× slower
 int 294.69M (  3.39ns) (± 0.11%)  0 B/op        fastest
span  273.7M (  3.65ns) (± 0.03%)  0 B/op   1.08× slower
 int 294.37M (   3.4ns) (± 0.41%)  0 B/op        fastest
span  273.7M (  3.65ns) (± 0.02%)  0 B/op   1.08× slower
```

RX14:
My conclusion is that it doesn't matter much: either way is optimized *so fast* that it will never be your bottleneck.

RX14:
100% agreed with @straight-shoota  on not including unrelated refactors inside the diff though

straight-shoota:
@RX14 I was referring to LLVM's optimization invalidating the result, which means the benchmark doesn't show what is supposed to.

src/time.cr
```diff
@@ -253,8 +253,17 @@ struct Time
   DAYS_PER_4_YEARS = 365*4 + 1
 
   # :nodoc:
+  # TODO: remove after 0.27.0, and replace the value of UNIX_EPOCH by "Time.utc 1970, 1, 1"
```
straight-shoota: Why can't this constant be removed right away? It's only used internally and since it is no longer used, what's the purpose of keeping it until the following release?
Plus, the value of `UNIX_EPOCH` should be `utc(SECONDS_PER_DAY.to_i64 * (1969*365 + 1969/4 - 1969/100 + 1969/400))`.

src/time.cr
```diff
@@ -252,8 +252,13 @@ struct Time
   # :nodoc:
   DAYS_PER_4_YEARS = 365*4 + 1
 
-  # :nodoc:
-  UNIX_SECONDS = SECONDS_PER_DAY.to_i64 * (1969*365 + 1969/4 - 1969/100 + 1969/400)
+  # This constant is defined to be "1970-01-01 00:00:00 UTC" on all systems with respect to the system clock.
```
straight-shoota: What is `on all systems with respect to the system clock` supposed to mean? The constant always represents the date `1970-01-01`. That doesn't depend on the system clock.

j8r: I've taken this from the Rust's documentation https://doc.rust-lang.org/std/time/constant.UNIX_EPOCH.html.

straight-shoota: Yeah, but what does it mean for the Crystal implementation? I can't make any sense of that.
In Rust it is an instance of `SystemTime` which is inherently system-dependent. That's probably the reason why they wrote that.

RX14: Please remove the "on all systems with respect to the system clock" wording.

src/time.cr
```diff
@@ -252,8 +252,13 @@ struct Time
...
+  # Can be used to create a `Time::Span` that represents an unix epoch time duration.
+  #
+  # ```
+  # Time.utc_now - Time::UNIX_EPOCH
+  # ```
+  UNIX_EPOCH = utc(seconds: SECONDS_PER_DAY.to_i64 * (1969*365 + 1969/4 - 1969/100 + 1969/400), nanoseconds: 0)
```
RX14: Why is this not just `Time.utc(1970, 1, 1)` ??

j8r: I was using at first, but this doesn't work. @straight-shoota [explained it in 2.](https://github.com/crystal-lang/crystal/pull/6908#issuecomment-427022520)

straight-shoota: Alternatively, `zone_offset_at` could just be skipped if `location == UTC`. Then `Time.utc(1970, 1, 1)` wouldn't depend on `UNIX_EPOCH`.

src/time.cr
```diff
@@ -409,7 +414,7 @@ struct Time
               second
 
     # Normalize internal representation to UTC
-    seconds = seconds - zone_offset_at(seconds, location)
+    seconds = seconds - zone_offset_at(seconds, location) if !location.utc?
```
RX14: `unless`

src/time.cr
r00ster91: Maybe this should be done by the formatter. WDYT?

src/time.cr
j8r: I thought you was against this type of discussions... it's neither more idiomatic, nor shorter.

Sija: yes, it is more idiomatic.

j8r: There is no style guide, so the stdlib is the reference - and this syntax is used in it, formats and compile.
Besides that for me `unless` rhymes with useless, can we stop having this type of conversations?

Sija: `unless` keyword exists exactly to be used in such cases - which is way more natural to read _unless time is utc_ than _if not time is utc_ - that's a bad english. and yes, stdlib uses `unless`, so please use the language constructs in a way they're provided for.

RX14: I'm against forcing people to use `unless`. I'll point it out if I think you forgot - but if you say "I don't want to use unless" then i'll just mark it as resolved and merge.

There's no need for this conversation.

PR #3902, opened 2017-01-16
allow setting and deleting of attributes on supported XML::Node types

bmmcginty:
Allow node[name]=value and node.attributes.delete(name)

I'm not sure if Attributes#delete is the correct place for the delete method, but I'm open to suggestions. Maybe Node#delete_attribute?
Also, are we avoiding all methods in libxml2 that require the LIBXML_TREE_ENABLED define? (I've continued that pattern in this PR.)
That should probably be noted somewhere if so.

PR #6910, opened 2018-10-04
Added missing XML::Attributes#delete

joenas:
Hello!

This adds the missing `XML::Attributes#delete` referenced in #3902. I wasn't sure about what kind of return value that would be appropriate so I went with what felt natural.

### DIFF
--- a/spec/std/xml/xml_spec.cr
+++ b/spec/std/xml/xml_spec.cr
@@ -359,4 +359,17 @@ describe XML do
     root["bar"] = 1
     root["bar"].should eq("1")
   end
+
+  it "deletes an attribute" do
+    doc = XML.parse(%{<foo bar="baz"></foo>})
+    root = doc.root.not_nil!
+
+    res = root.delete("bar")
+    root["bar"]?.should be_nil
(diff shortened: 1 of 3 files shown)


straight-shoota:
"Delete" methods usually return the value that is being deleted. I think this would make sense here, too. `nil` would indicate that the attribute doesn't exist.

spec/std/xml/xml_spec.cr
```diff
@@ -359,4 +359,17 @@ describe XML do
...
   end
+
+  it "deletes an attribute" do
+    doc = XML.parse(%{<foo bar="baz"></foo>})
+    root = doc.root.not_nil!
+
+    res = root.delete("bar")
+    root["bar"]?.should be_nil
+    root.to_s.should eq(%{<foo/>})
+    res.should eq "bar"
```
straight-shoota: I meant it to return the *value*, in this case `baz`.

joenas: Right

joenas:
🎊 🎈

PR #6912, opened 2018-10-06
add support for 100-continue in HTTP::Server::Response

jreinert:


src/http/server/response.cr
```diff
@@ -79,6 +79,15 @@ class HTTP::Server
       raise "Can't read from HTTP::Server::Response"
     end
 
+    # Send a 100 Continue response to clients waiting for this before sending
+    # the request body
+    def continue
+      @status_code = 100
```
straight-shoota: I'd suggest to let the status code be configurable as argument (default to `100`). You might want to use this mechanic with other status codes (`102` for example).

straight-shoota: Alternatively, only set it to 100 if it was not explicitly set before, but that'd be kind of hart to detect.

jreinert: sure, why not. Should it still be called continue then?

straight-shoota: Good question. I'm not sure about the name anyway.
Maybe something like `send_intermediate_headers`?

jreinert: Maybe we should do some research how other web frameworks/servers call this...

straight-shoota:
Go's webserver implementation doesn't provide a means to send 100 Continue explicitly, but it automatically sends the continue header when `Expect: 100-continue` is present as soon as the handler reads from the IO.
See https://golang.org/pkg/net/http/#ResponseWriter

Copying this might even be more useful since it just works with any handler and doesn't require any custom logic.

jreinert:
@straight-shoota That is perfect!

src/http/common.cr
```diff
@@ -40,6 +40,10 @@ module HTTP
           body = UnknownLengthContent.new(io)
         end
 
+        if body.is_a?(Content) && expect_continue?(headers)
```
straight-shoota: `body` should be a `Content` here in any case. IMO restricting the type with `body : Content? = nil` above is cleaner.

src/http/common.cr
```diff
@@ -40,6 +40,10 @@ module HTTP
...
+          body.expect_continue
```
straight-shoota: Jus another nit bit: I'd prefer `expect_continue` to be an explicit setter for the ivar. `body.expect_continue = true` is more explicit about it's function.

straight-shoota:
I noticed there is an issue when the handler doesn't read from the IO and `100 Continue` status should not be sent. A spec for this case is missing. But `Content#close` calls `skip_to_end` which read the entire IO and thus sent `100 Continue` nonetheless. I suppose `#close` just needs to set `@expects_continue = false` before calling `skip_to_end`.

straight-shoota:
Actually, I'm not entirely sure *why* `skip_to_end` even is there... It seems to be the only reason for the `Content` type in the first place.

But I can't see how that is actually useful, and at least, it is confusing to the user. It doesn't just close the IO in it's current state but completely consumes it until the sender is done (maybe a happy `curl -T /dev/urandom example.com`? 😆 ).
I mean you don't *have* to call it, just exiting the handler will close the socket without consuming everything.
I'm just wondering what's the use case and if there is a better solution?

src/http/common.cr
jreinert: Unfortunately this doesn't work. The assignment `body = ChunkedContent.new(io)` then fails with `type must be (HTTP::Content | Nil), not (IO | Nil)` :(

src/http/common.cr
jreinert: Good point! Update incoming.

jreinert:
@straight-shoota i agree. this is weird indeed. But I think it should be a seperate issue.

spec/std/http/server/server_spec.cr
```diff
@@ -275,6 +275,66 @@ module HTTP
...
+
+        response = Client::Response.from_io(socket)
+        response.status_code.should eq(200)
+        response.body.should eq("hello")
+      end
+    end
+
+    it "handles Expect: 100-continue correctly when body isn't read" do
+      server = Server.new do |context|
+        context.response.respond_with_error("I don't want your body", 400)
```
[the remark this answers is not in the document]
jreinert: ;)

src/http/content.cr
```diff
@@ -3,16 +3,53 @@ require "http/common"
...
+    @continue_sent = false
+    @expects_continue = false
+
     def close
+      @expects_continue = false
       skip_to_end
       super
     end
+
+    def expects_continue=(value : Bool)
```
Sija: Perhaps cleaner: `property? expects_continue : Bool`

jreinert: :+1:

src/http/common.cr
straight-shoota: Oh, that looks like an error in the type inference. `ChunkedContent.new(io)` is actually a `HTTP::Content`. And it can't be `Nil`.

Sija: > And it can't be `Nil`.

@straight-shoota there's `body = nil` on line 32, that's the reason I guess.

straight-shoota: Ah, yes. It means the resulting type of `body`. I interpreted as trying to assign `IO | Nil` to the variable which is restricted to `Content | Nil`. But yeah, that makes sense.

src/http/content.cr
```diff
@@ -3,16 +3,49 @@ require "http/common"
 module HTTP
   # :nodoc:
   module Content
+    CONTINUE = "HTTP/1.1 100 Continue\r\n\r\n"
+
+    @continue_sent = false
```
Sija: Wouldn't be good to expose it to the outside world?
I'm thinking `getter? continue_sent : Bool = false`.

straight-shoota: For what purpose? The entire types are all `nodoc`.

straight-shoota: That means internal. And unless a getter would actually be used anywhere in stdlib, it shouldn't be necessary to add it.

Sija: ok, so `expects_continue` should be better defined with `setter?` instead of `property?` then.

spec/std/http/server/server_spec.cr
```diff
@@ -275,6 +275,66 @@ module HTTP
...
 
+    it "handles Expect: 100-continue correctly when body is read" do
+      server = Server.new do |context|
+        context.response << context.request.body.not_nil!.gets_to_end
+      end
+
+      address = server.bind_unused_port
+      spawn { server.listen }
+
+      Fiber.yield
```
straight-shoota: Please adjust this according to #6953

RX14:
@jreinert needs a format

jreinert:
> @jreinert needs a format

done

src/http/content.cr
```diff
@@ -3,16 +3,50 @@ require "http/common"
...
+      super
+    end
+
+    def peek
+      ensure_send_continue
+      super
+    end
+
+    def skip(bytes_count)
+      ensure_send_continue
```
RX14: Why do we `ensure_send_continue` in `#skip` here but not in `UnknownLengthContent`?

jreinert: Apparently because I'm just sloppy :P Sorry.

src/http/content.cr
```diff
@@ -26,18 +60,22 @@ module HTTP
...
       @io.read_byte
     end
 
     def peek
+      ensure_send_continue
       @io.peek
     end
 
     def skip(bytes_count)
+    enensure_send_continue
```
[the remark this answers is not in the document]
jreinert: uh jezzus.. i should sleep more. xD

src/http/content.cr
```diff
@@ -26,18 +60,22 @@ module HTTP
...
       @io.read_byte
     end
 
     def peek
+      ensure_send_continue
       @io.peek
     end
 
     def skip(bytes_count)
+      enensure_send_continue
```
Sija: One `en` too much.

jreinert: ... i don't... even..

Sija: Relax, that happens, especially when there's lack of sleep included :) I'd just advise "proofreading" your PRs before submitting ;)

PR #6914, opened 2018-10-07
Add step in Installation section in README template

r00ster91:
This documents the `shards install` step in the Installation section of the README template.
Because actually the Installation section is without this step incomplete. That's not everything you need to do to install the shard so you can use it.
So now the Installation section for a shard would look like this:

---

## Installation

Add this to your application's `shard.yml`:
 
```yaml
dependencies:
  shard:
    github: user/shard
```

Then run `shards install`.

### DIFF
--- a/src/compiler/crystal/tools/init/template/readme.md.ecr
+++ b/src/compiler/crystal/tools/init/template/readme.md.ecr
@@ -13,7 +13,7 @@ dependencies:
     github: <%= config.github_name %>/<%= config.name %>
 ```
 
-and run `shards install`.
+Then run `shards install`.
 <%- else -%>
 TODO: Write installation instructions here
 <%- end -%>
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/tools/init/template/readme.md.ecr
```diff
@@ -12,6 +12,8 @@ dependencies:
   <%= config.name %>:
     github: <%= config.github_name %>/<%= config.name %>
 ```
+
+and run `shards install`.
```
straight-shoota: This should be a whole sentence (for example `Then run `shards install`).

r00ster91: I think this is fine because `Add this to your application's shard.yml:`, the yaml code and this together is one sentence. So the whole installation is swifty and simply described in one sentence.
"Do this and run this" instead of "Do this. Then run this".

straight-shoota: It's hard to read that as a whole sentence given it's two paragraphs with a code block element in between.

PR #6915, opened 2018-10-07
Fix empty case statement normalizes to condition expression

straight-shoota:
After #6367, an empty case statement (without `when` or `else`) would be normalized to just the condition. This means that `case "foo"; else` is normalized to `"foo"`. But that's semantically incorrect. The return type of an empty `case` should just be `nil` because there is no branch that could return anything else.

This fix adds a `nil` after the condition when there are no `when` or `else` branches, so the example normalizes to `"foo"; nil`.

This PR also adds two codegen specs which are not necessary to document this fix, but they should have been added in #6367.

### DIFF
--- a/spec/compiler/codegen/case_spec.cr
+++ b/spec/compiler/codegen/case_spec.cr
@@ -13,6 +13,14 @@ describe "Code gen: case" do
     run("require \"prelude\"; case 1; when 0; 2; else; 3; end").to_i.should eq(3)
   end
 
+  it "codegens case without whens" do
+    run("require \"prelude\"; case 1; end").to_string.should eq ""
+  end
+
+  it "codegens case without whens but else" do
+    run("require \"prelude\"; case 1; else; 2; end").to_i.should eq(2)
+  end
+
   it "codegens case that always returns" do
     run("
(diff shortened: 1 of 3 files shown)


spec/compiler/codegen/case_spec.cr
```diff
@@ -13,6 +13,14 @@ describe "Code gen: case" do
     run("require \"prelude\"; case 1; when 0; 2; else; 3; end").to_i.should eq(3)
   end
 
+  it "codegens case without whens" do
+    run("require \"prelude\"; case 1; end").to_string.should eq ""
```
asterite: I'm not sure it's a good idea to turn a `nil` into a string like this... it probably works, but it's not clear why (I wouldn't add a test for this if the normalizer is already tested, and same for the test below).

straight-shoota: `to_string` actually just returns `@output` of `Crystal::SpecRunOutput` which is already a string.

straight-shoota: But I can remove these specs, if their not considered particularly useful.

PR #6916, opened 2018-10-07
Let Zlib::Reader, Gzip::Reader and Flate::Reader include IO::Buffered for optimized performance

asterite:
All of these readers wrap another IO from which they read. The problem is that when using methods like `gets` which read until some character is found, usually a newline, it's much faster to use input buffering to read a lot of bytes at once instead of going byte per byte (or worse, utf-8 char per char). The `gets` method actually relies on a `peek` method to peek next bytes if available to perform the scan faster, and the easiest way to accomplish this is by including `IO::Buffered`, which implements this.

(There _might_ be a way to implement this without include `IO::Buffered`, maybe for example in `Zlib::Reader` having an output buffer similar to the one in `IO::Buffered`, and reading there and copying from that to the requested slice by `read`, but that's basically duplicating the work that's done in `IO::Buffered`, so it's probably not worth it).

Note that this will require an additional 8KB of memory for each of these IOs, but that's fine because one usually doesn't deal with a lot of IOs at a time.

This optimization happened because the slowness was [discovered in IRC](https://gitter.im/crystal-lang/crystal?at=5bb999143844923661f7abb1).

Also checking what Python does, their gzip reader is buffered too (and I guess that's the case for Ruby too).

(An additional research, if someone wants to tackle this, is checking whether other IOs in the std, notably HTTP responses, might need this optimization too.)

Benchmark follows:

```crystal
require "gzip"
require "zlib"
require "flate"

FILENAME = "test.gzip"

def benchmark(type : T.class) forall T
  File.open(FILENAME, "w") do |file|
    T::Writer.open(file) do |writer|
      1_000_000.times do |i|
        writer.puts "This is line number #{i}"
      end
    end
  end

  i = 0
  puts "#{T}:"
  puts Time.measure {
    File.open(FILENAME) do |file|
      T::Reader.open(file) do |reader|
        reader.each_line do
          i += 1
        end
      end
    end
  }

  puts i
end

benchmark(Gzip)
benchmark(Zlib)
benchmark(Flate)
```

Old output:

```
Gzip:
00:00:03.052840054
1000000
Zlib:
00:00:02.439849829
1000000
Flate:
00:00:01.947586854
1000000
```

New output:

```
Gzip:
00:00:00.103666309
1000000
Zlib:
00:00:00.095306070
1000000
Flate:
00:00:00.090960507
1000000
```

Gzip: 29 times faster
Zlib: 25 times faster
Flate: 21 times faster

### DIFF
--- a/src/flate/reader.cr
+++ b/src/flate/reader.cr
@@ -4,11 +4,16 @@
 # instance, it reads data from the underlying IO, decompresses it, and returns
 # it to the caller.
 class Flate::Reader < IO
+  include IO::Buffered
+
   # If `#sync_close?` is `true`, closing this IO will close the underlying IO.
   property? sync_close : Bool
 
   # Returns `true` if this reader is closed.
   getter? closed = false
+
+  # Dictionary passed in the constructor
+  getter dict : Bytes?
 
   # Peeked bytes from the underlying IO
   @peek : Bytes?
@@ -22,6 +27,7 @@ class Flate::Reader < IO
     ret = LibZ.inflateInit2(pointerof(@stream), -LibZ::MAX_BITS, LibZ.zlibVersion, sizeof(LibZ::ZStream))
     raise Flate::Error.new(ret, @stream) unless ret.ok?
 
+    @peek = nil
     @end = false
   end
 
@@ -46,12 +52,12 @@ class Flate::Reader < IO
   end
 
   # Always raises `IO::Error` because this is a read-only `IO`.
-  def write(slice : Bytes)
+  def unbuffered_write(slice : Bytes)
     raise IO::Error.new "Can't write to Flate::Reader"
   end
 
   # See `IO#read`.
-  def read(slice : Bytes)
+  def unbuffered_read(slice : Bytes)
     check_open
 
     return 0 if slice.empty?
@@ -119,8 +125,12 @@ class Flate::Reader < IO
     end
   end
 
+  def unbuffered_flush
+    raise IO::Error.new "Can't flush Flate::Reader"
+  end
+
   # Closes this reader.
-  def close
+  def unbuffered_close
     return if @closed
     @closed = true
 
@@ -128,6 +138,14 @@ class Flate::Reader < IO
     raise Flate::Error.new(ret, @stream) unless ret.ok?
 
     @io.close if @sync_close
+  end
+
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
+
+    @io.rewind
+
+    initialize(@io, @sync_close, @dict)
   end
 
   # :nodoc:
--- a/src/gzip/reader.cr
+++ b/src/gzip/reader.cr
@@ -27,6 +27,8 @@
 # string # => "abc"
 # ```
 class Gzip::Reader < IO
+  include IO::Buffered
+
   # Whether to close the enclosed `IO` when closing this reader.
   property? sync_close = false
 
@@ -73,7 +75,7 @@ class Gzip::Reader < IO
   end
 
   # See `IO#read`.
-  def read(slice : Bytes)
+  def unbuffered_read(slice : Bytes)
     check_open
 
     return 0 if slice.empty?
@@ -121,16 +123,31 @@ class Gzip::Reader < IO
   end
 
   # Always raises `IO::Error` because this is a read-only `IO`.
-  def write(slice : Bytes) : Nil
+  def unbuffered_write(slice : Bytes) : Nil
     raise IO::Error.new("Can't write to Gzip::Reader")
   end
 
+  def unbuffered_flush
+    raise IO::Error.new "Can't flush Gzip::Reader"
+  end
+
   # Closes this reader.
-  def close
+  def unbuffered_close
     return if @closed
     @closed = true
 
     @flate_io.try &.close
     @io.close if @sync_close
+  end
+
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
+
+    @io.rewind
+
+    @header = nil
+    @flate_io = nil
+
+    initialize(@io, @sync_close)
   end
 end
--- a/src/zlib/reader.cr
+++ b/src/zlib/reader.cr
@@ -4,6 +4,8 @@
 # instance, it reads data from the underlying IO, decompresses it, and returns
 # it to the caller.
 class Zlib::Reader < IO
+  include IO::Buffered
+
   # Whether to close the enclosed `IO` when closing this reader.
   property? sync_close = false
 
@@ -56,7 +58,7 @@ class Zlib::Reader < IO
   end
 
   # See `IO#read`.
-  def read(slice : Bytes)
+  def unbuffered_read(slice : Bytes)
     check_open
 
     return 0 if slice.empty?
@@ -79,16 +81,28 @@ class Zlib::Reader < IO
   end
 
   # Always raises `IO::Error` because this is a read-only `IO`.
-  def write(slice : Bytes)
+  def unbuffered_write(slice : Bytes)
     raise IO::Error.new "Can't write to Zlib::Reader"
   end
 
-  def close
+  def unbuffered_flush
+    raise IO::Error.new "Can't flush Zlib::Reader"
+  end
+
+  def unbuffered_close
     return if @closed
     @closed = true
 
     @flate_io.close
     @io.close if @sync_close
+  end
+
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
+
+    @io.rewind
+
+    initialize(@io, @sync_close, @flate_io.dict)
   end
 
   protected def self.invalid_header
--- a/spec/std/flate/flate_spec.cr
+++ b/spec/std/flate/flate_spec.cr
@@ -1,14 +1,18 @@
 require "spec"
 require "flate"
 
+private def new_sample_io
+  io = IO::Memory.new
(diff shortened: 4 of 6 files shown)


bcardiff:
Can be composition of IOs used to reach the performance? 

It would be great to keep IO as orthogonal as possible IMO. 

If we need to chain two or more IOs that happen to be all buffered then we will end with intermediate buffers that will not be used at all and just filling memory I think.

asterite:
@bcardiff Yes, for that we need to remove the `IO::Buffered` module and make it a wrapper with which you wrap other IOs. It's also a silent breaking change if `File` for example stop doing buffering by default.

On the other hand, if you have a `File` and a `Flate::Reader`, I think first the raw bytes from the File will be buffered by `File`, and then the decompressed bytes will be buffered by `Flate::Reader`, so the data is different and it might make sense (as in, it might be faster) to have two buffers. I'm not sure though, I guess someone will have to benchmark that.

asterite:
Actually, setting both file and gzip::reader to `read_buffering = false` gives the same time, so probably `read_buffering` is not working well...

asterite:
I just added another commit that fixes `IO::Buffered#peek`: it wasn't honoring the `read_buffering` setting.

With that, if I set `read_buffering = false` in the File it works very, very slow. The reason is that:
- you ask from, say, Flate::Reader, a line to decompress
- Flate::Reader is now buffered, so it will ask 8KB to decompress instead of a line
- to do this, it needs to first read bytes from the file. But there's [a condition](https://github.com/crystal-lang/crystal/blob/master/src/flate/reader.cr#L65-L75) there to not read more than we need, and if we can't peek we have to read byte per byte (this is because we don't in turn implement flate in Crystal and rely on zlib). And this is slow. If it's buffered we can peek and it's faster.

Anyway, I think adding buffering at each level is good. And, again, 8KB pero IO is not a big deal, considering that the OS can't have many file descriptors open at once.

straight-shoota:
Yeah, it makes a difference to have multiple buffers in the IO chain. All potentially costly read operations should be buffered to speed things up.

`Gzip::Reader#read` for example invokes a number of smaller reads on the wrapped IO. When that's a file, it needs a buffer as well as the `Gzip::Reader` itself benefits from a buffer.

I've taken a quick looked at the other IO implementations
and most of them probably won't benefit from 

`OpenSSL::DigestIO`, `Zip::ChecksumReader` and `Zip::ChecksumWriter` could probably benefit from being buffered, because every read/write operation updates the digest.

HTTP content IO's look fine, though. They mostly pass through.

A buffer is only useful if there is a significant performance improvement between 1000 `read_byte` and `read(Bytes.new(1000))` (for example).

asterite:
@straight-shoota Thank you for the research!

I think buffering is not needed for wrapper as long as they provide a good `peek` implementation, which is what the `gets` and all the methods that need to find something while reading bytes need. `Zip::ChecksumReader` already has `peek`, but `OpenSSL::DigestIO` doesn't have it and could use one (but maybe for another PR).

straight-shoota:
Relying on the user to `peek` is an incomplete solution. It can eliminate the need for buffering in some situations. But when you have a known data format and just issue a number of smaller reads, you don't need to peek and it wouldn't help in any way.
Plus, not every IO necessarily supports `peek` at all.

RX14:
This is fine to implement now that we have `read_buffering=` available.

src/flate/reader.cr
```diff
@@ -119,8 +125,12 @@ class Flate::Reader < IO
     end
   end
 
+  def unbuffered_flush
+    raise IO::Error.new "Can't flush Flate::Reader"
```
RX14: Shouldn't this be: `io.flush`? It makes no sense to raise here.

src/flate/reader.cr
```diff
@@ -130,6 +140,14 @@ class Flate::Reader < IO
     @io.close if @sync_close
   end
 
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
```
RX14: `check_open`.

src/flate/reader.cr
asterite: Well, it's a reader, but yeah, we can delegate

RX14: Do other readers raise on flush? I thought the default semantic for flush was to silently ignore if it has nothing to do.

RX14:
Reading from any `IO::Buffered` with a buffer greater than 4096 bytes will bypass the buffering (safely) anyway, so `IO::Buffered` impls should compose well. Unfortunately `Flate::Reader` will only ever read by 1 byte or use peek, which avoids that...

I'm not sure `Gzip::Reader` should be buffered because it always delegates to `flate_io`. Same for `Zlib::Reader`. Either they should be unbuffered or turn off buffering on their `Flate::Reader`s.

straight-shoota:
@RX14 `Gzip` and `Zlib` both delegate reading the entire slice to `Flate`, so yes, as long as they themselves are buffered, a `Flate` reader wrapped in such a way shouldn't need a read buffer as well.

src/flate/reader.cr
asterite: Thanks! I updated the code.

asterite:
@RX14 @straight-shoota Yeah, probably the buffer isn't needed for gzip and zlib, but if I don't include `IO::Buffered` it works slow. Maybe we'll have to see how to implement `peek`, but I have no idea how to do it. But feel free to play with that, if you can implement `peek` we can remove `IO::Buffered`.

RX14:
with `Zlib::Reader` you can just forward peek to `@flate_io`.

with `Gzip::Reader` you just need to forward peek, but handle `peek.size == 0` specially, copying the contents of the `read_bytes == 0` branch, pretty much exactly

asterite:
But:
- `Zlib::Reader` has a section that reads a checksum, and that isn't returned in `read`.
- `Gzip::Reader` must handle a header, I'm not sure how to account for that in `peek`

I think it's better for me to close this PR and you can work on all of these things because I won't have time to update this PR with more changes.

RX14:
@asterite notice that in `Zlib::Reader` there's only  special cases when `@flate_io.read` returns zero, i.e. EOF, so they don't affect peek.

`Gzip::Reader` has a special case because `@flate_io`'s EOF doesn't mean the reader's EOF, but you just have to loop `peek = @flate_io.peek` and `read_next_header` until the peek size is > 0 to avoid accidentally returning EOF from peek.

straight-shoota:
No, `Zlib::Reader` and `Gzip::Reader` both need to include `IO::Buffered`. Their single-byte read is much slower than a bigger slice.
But they can disable the buffer in the wrapped `Flate::Reader` because when they are buffered themselves, they will only issue large slice reads to the Flate reader.

asterite:
Ah, that makes sense. I'll try it later and update the code.

RX14:
Might as well keep this PR as-is then because of the 4KiB rule.

asterite:
Yes, I think the current PR is fine. Feel free to merge it.

src/flate/reader.cr
asterite: Yes, other reads raise on flush.

asterite:
@RX14 Ping. Other readers also raise on flush, or at least `File::PReader` does. We can maybe leave this discussion for another issue/PR.

RX14:
I'm not sure what you mean regarding flush, I'm talking about this failing:

```cr
io.read_byte
# io now has a read buffer of size 8192, 8191 bytes are valid
io.read_buffering = false
io.gets
# io.gets will return only a maximum of 8191 bytes, even if the next `\n` or `\r\n` is beyond that size.
```

asterite:
For that I suggest removing `read_buffering`. As I said, it's only needed for secure random. For instance, you can't disable read buffering in Ruby.

asterite:
Or we can fix `IO::Buffered#gets` to check for `read_buffering` in addition to `peek`. But it's not related to this PR, really. Or yes, I can remove the "fix" for `read_buffering`, would that be OK?

asterite:
Okay, I removed the `read_buffering` changes.

RX14:
It's fine to remove that fix for now but the bug still persists and should be opened in another issue.

src/flate/reader.cr
Sija: Why not just use `check_open` call like @RX14 suggested?

src/gzip/reader.cr
```diff
@@ -121,16 +123,31 @@ class Gzip::Reader < IO
...
+  def unbuffered_close
     return if @closed
     @closed = true
 
     @flate_io.try &.close
     @io.close if @sync_close
   end
+
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
```
Sija: ditto

src/zlib/reader.cr
```diff
@@ -79,18 +81,30 @@ class Zlib::Reader < IO
...
+  def unbuffered_close
     return if @closed
     @closed = true
 
     @flate_io.close
     @io.close if @sync_close
   end
 
+  def unbuffered_rewind
+    raise IO::Error.new "Closed stream" if closed?
```
Sija: ditto

asterite: Yeah, it was like that but I had to rollback to a previous version and I forgot to add that.

asterite: Someone can send a PR, or commit directly to master with it (I'm trying but I can't).

Sija: #6958 :)

PR #5676, opened 2018-02-02
Fixed OpenSSL bindings to recognize LibreSSL

LVMBDV:
Used the C preprocessor from `cc` to get `LIBRESSL_VERSION_NUMBER` and `LIBRESSL_VERSION_NUMBER` which are used to determine `OPENSSL_110` and `OPENSSL_102`, as @RX14 suggested. This fixes #4676. Tested on my Void Linux machine with LibreSSL.

PR #6917, opened 2018-10-07
Fixed OpenSSL bindings to work with LibreSSL

LVMBDV:
- Reworked #5676 with sane fallbacks, like falling back to using just `pkg-config` if the header files are inaccessible.
- Replaced flags like `OPENSSL_110` with actual version number values e.g. `OPENSSL_VERSION` is `0x10102000` when its version is `1.1.2`.
- When we can't check the headers but can use `pkg-config`, we assume OpenSSL is installed and set the version values accordingly. Although this behavior is a bit iffy for systems that have LibreSSL installed without the headers but the current behavior also works exactly like this AFAIU.

### DIFF
--- a/src/openssl/lib_crypto.cr
+++ b/src/openssl/lib_crypto.cr
@@ -1,8 +1,30 @@
 {% begin %}
   lib LibCrypto
-    OPENSSL_111 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.1 libcrypto || printf %s false`.stringify != "false" }}
-    OPENSSL_110 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.0 libcrypto || printf %s false`.stringify != "false" }}
-    OPENSSL_102 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.0.2 libcrypto || printf %s false`.stringify != "false" }}
+    {% if `command -v pkg-config > /dev/null || printf %s false` != "false" %}
+      {% if `test -f $(pkg-config --variable=includedir libcrypto)/openssl/opensslv.h || printf %s false` != "false" %}
+        {% libressl_version_number = `printf %s "#include <openssl/opensslv.h>\nLIBRESSL_VERSION_NUMBER" | #{(env("CC") || "cc").id} #{`pkg-config --cflags --silence-errors libcrypto || true`.chomp}  -E -`.chomp.split.last %}
+        LIBRESSL_VERSION = {{ (libressl_version_number == "LIBRESSL_VERSION_NUMBER") ? 0 : libressl_version_number.split('L').first.split("0x").last.to_i(16) }}
+        {% if libressl_version_number == "LIBRESSL_VERSION_NUMBER" %}
+          OPENSSL_VERSION = {{ `printf %s "#include <openssl/opensslv.h>\nOPENSSL_VERSION_NUMBER" | #{(env("CC") || "cc").id} #{`pkg-config --cflags --silence-errors libcrypto || true`.chomp}  -E -`.chomp.split.last.split("L").first.id }}
+        {% else %}
+          OPENSSL_VERSION = 0
+        {% end %}
+      {% else %}
+        LIBRESSL_VERSION = 0
+        {% if `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.1 libcrypto || printf %s false`.stringify != "false" %}
+          OPENSSL_VERSION = 0x10101000
+        {% elsif `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.0 libcrypto || printf %s false`.stringify != "false" %}
+          OPENSSL_VERSION = 0x10100000
+        {% elsif `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.0.2 libcrypto || printf %s false`.stringify != "false" %}
+          OPENSSL_VERSION = 0x10002000
+        {% else %}
+          OPENSSL_VERSION = 0
+        {% end %}
+      {% end %}
+    {% else %}
+      OPENSSL_VERSION = 0
+      LIBRESSL_VERSION = 0
+    {% end %}
   end
 {% end %}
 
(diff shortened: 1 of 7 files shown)


LVMBDV:
The fail on [build 12969](https://circleci.com/gh/crystal-lang/crystal/12969) seems irrelevant. Ready for a review :)

bcardiff:
Build restarted.

I would prefer to use  `compare_versions` macro method instead of versions hex encoded. Is there any reason why not to do it that way?

straight-shoota:
> Is there any reason why not to do it that way?

Is there a string representation of the library version available that would fit with `compare_versions`? `OPENSSL_VERSION` is just a binary value.

LVMBDV:
There is a string representation of the library version inside the same header file.
For LibreSSL, it's `LIBRESSL_VERSION_TEXT` which is formatted like:
`"LibreSSL 2.7.4"`
And for OpenSSL, it's `OPENSSL_VERSION_TEXT` which is formatted like:
`"OpenSSL 1.0.2q-dev  xx XXX xxxx"`

I could use the MAJOR.MINOR.PATCH parts of both if you prefer, but I think
`OPENSSL_VERSION >= 0x10100000`
is more readable than
`compare_versions(OPENSSL_VERSION, "1.1.0") >= 0`

bcardiff:
Even if there is no string representation, I don't see the value of using custom integer value vs initializing manually a string semver value.

It's better to keep one single idiom for version. 

So either we needed to extract the semver from `..._VERSION_TEXT` or initialize those semver strings manually where this PR now uses integers. The later is good enough for me.

ysbaddaden:
The official version check in OpenSSL is the hexadecimal check. As ugly as it may look, we should follow it.

What really bothers me is the amount and complexity of macro run calls to extract 2 definitions from the C headers :-(

LVMBDV:
I reworked the checks to work with `compare_versions` and tried to get the semver strings from `pkg-config` but turns out it doesn't always output semver-compatible versions.

LVMBDV:
The second solution seems to pass too. It's a bit simpler as it uses the header file only to check for LibreSSL and uses `pkg-config` to get the semver strings.

Both commits are valid solutions, pick your favorite :)

src/openssl/lib_ssl.cr
```diff
@@ -2,9 +2,17 @@ require "./lib_crypto"
...
+    {% from_libressl = (`command -v pkg-config > /dev/null || printf %s false` != "false") &&
+                       (`test -f $(pkg-config --silence-errors --variable=includedir libssl)/openssl/opensslv.h || printf %s false` != "false") &&
+                       (`printf %s "#include <openssl/opensslv.h>\nLIBRESSL_VERSION_NUMBER" | #{(env("CC") || "cc").id} #{`pkg-config --cflags --silence-errors libssl || true`.chomp}  -E -`.chomp.split('\n').last == "LIBRESSL_VERSION_NUMBER") %}
+
+    {% if from_libressl %}
+      LIBRESSL_VERSION = {{ `command -v pkg-config > /dev/null && pkg-config --silence-errors --modversion libssl || printf %s 0.0.0`.split.last.gsub(/[^0-9.]/, "") }}
+      OPENSSL_VERSION = "0.0.0"
+    {% else %}
+      LIBRESSL_VERSION = "0.0.0"
+      OPENSSL_VERSION = {{ `command -v pkg-config > /dev/null && pkg-config --silence-errors --modversion libssl || printf %s 0.0.0`.split.last.gsub(/[^0-9.]/, "") }}
```
j8r: `hash pkg-config 2>/dev/null` is shorter and has the same effect than `command -v pkg-config >/dev/null`.

You can also do this command once, before `{% if from_libressl %}`:
```{% ssl_version = `hash pkgconfig 2>/dev/null && pkg-config --silence-errors --modversion libssl || printf %s 0.0.0`.split.last.gsub(/[^0-9.]/, "") %}```

And then:
`LIBRESSL_VERSION = {{ ssl_version }}`
`OPENSSL_VERSION = {{ ssl_version }}`

Comment (LVMBDV) [reply src/openssl/lib_ssl.cr]: Yeah, you're right. I remember thinking about storing the libssl version last night, wonder why it slipped my mind later. I will do that when I get on my computer.

File: src/openssl/lib_crypto.cr [hunk 2]
```diff
@@ -1,8 +1,17 @@
 {% begin %}
   lib LibCrypto
-    OPENSSL_111 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.1 libcrypto || printf %s false`.stringify != "false" }}
-    OPENSSL_110 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.1.0 libcrypto || printf %s false`.stringify != "false" }}
-    OPENSSL_102 = {{ `command -v pkg-config > /dev/null && pkg-config --atleast-version=1.0.2 libcrypto || printf %s false`.stringify != "false" }}
+    {% from_libressl = (`hash pkg-config 2> /dev/null || printf %s false` != "false") &&
+                       (`test -f $(pkg-config --silence-errors --variable=includedir libcrypto)/openssl/opensslv.h || printf %s false` != "false") &&
+                       (`printf %s "#include <openssl/opensslv.h>\nlibressl_version_number" | #{(env("cc") || "cc").id} #{`pkg-config --cflags --silence-errors libcrypto || true`.chomp}  -E -`.chomp.split('\n').last == "libressl_version_number") %}
```
RX14: perhaps use `$(pkg-config ...)` here instead of ``#{`pkg-config ...`}``

straight-shoota: `#{(env("cc") || "cc").id}` could be inserted directly into the command, as well.

LVMBDV:
The fail on [build 13549](https://travis-ci.org/crystal-lang/crystal/builds/440551925) seems irrelevant to me. Is it from master?

RX14:
@LVMBDV no it's just a transient OOM error, I've restarted the build.

RX14:
Tested on my machine, this works well. Has this been tested on alpine?

Sija:
> I could use the MAJOR.MINOR.PATCH parts of both if you prefer, but I think
> `OPENSSL_VERSION >= 0x10100000`
> is more readable than
> `compare_versions(OPENSSL_VERSION, "1.1.0") >= 0`

Somewhat agreed, perhaps it's worth adding 3rd argument which would specify the operator for comparison, turning your example into: `compare_versions(OPENSSL_VERSION, ">=", "1.1.0")`?

j8r:
To test this features, it would be really nice to have an Alpine Linux CI when the language is compiled with:
- Libressl
- musl-libc
- static linking?

PR #6927, opened 2018-10-10
Improve support for parsing and formatting //

### LINKED ISSUE #6761: 0.26 macro fails to accept keyword arguments with empty regex
The following code works with `0.24` and `0.25`, but not with `0.26.1` and `master`.
- master: Crystal 0.26.1+67 [421497e] (2018-09-21)

```crystal
macro foo(regex)
end

foo(regex: //)
```

```
Syntax error in /tmp/test.cr:4: unterminated call

foo(regex: //)
   ^
```

It works with `foo(regex: /.*/)` or `foo(//)` in `0.26.1`.

### versions

| version   | code             | result            |
|-----------|------------------|-------------------|
| 0.24.x    | foo(regex: //)   | OK                |
| 0.25.x    | foo(regex: //)   | OK                |
| 0.26.1    | foo(regex: //)   | unterminated call |
|           | foo(regex: /.*/) | OK                |
|           | foo(//)          | OK                |
| 0.26.1+67 | foo(regex: //)   | unterminated call |

### real codes
- https://github.com/maiha/comment-spec.cr/blob/master/src/comment-spec/regex_parser.cr#L158-L161

Best regards,

bcardiff:
Fixes #6761 
Required for #6891
and implements the same logic from 5cb28863f4950572287a47e60de0ab18920d4f8c in the formatter.

Essentially improves the support for dealing with // as a integer division operator or empty regex is some corner cases by the parser and by the formatter.

Some of the specs are not needed per-se but I added while learning a bit of the formatter. I left them to add more coverage.


bcardiff:
I was so lost while to debugging the formatter and the parser that ... [bcardiff/crystal-ast-helper](https://github.com/bcardiff/crystal-ast-helper) 🤓 

Ugly code that helps

cc: @MakeNowJust @asterite @ozra , in case you find it useful

PR #6928, opened 2018-10-11
Detect and warn on stack overflow

### LINKED ISSUE #271: Stack overflows

damaxwell:
This PR fleshes out PoC #6376.

Closes #271.

### DIFF
--- a/spec/std/kernel_spec.cr
+++ b/spec/std/kernel_spec.cr
@@ -220,3 +220,50 @@ describe "at_exit" do
                            OUTPUT
   end
 end
+
+describe "seg fault" do
+  it "reports SIGSEGV" do
+    status, _, error = build_and_run <<-'CODE'
+      puts Pointer(Int64).new(0x00FEDCBA).value
+    CODE
+
+    status.success?.should be_false
+    error.should contain("Invalid memory access")
+    error.should_not contain("Stack overflow")
+  end
+
+  it "detects stack overflow on the main stack" do
+    # This spec can take some time under FreeBSD where
+    # the default stack size is 0.5G.  Setting a
+    # smaller stack size with `ulimit -s 8192`
+    # will address this.
+    status, _, error = build_and_run <<-'CODE'
+      def foo
+        y = StaticArray(Int8,512).new(0)
+        foo
+      end
+      foo
+    CODE
+
+    status.success?.should be_false
+    error.should contain("Stack overflow")
+  end
+
+  it "detects stack overflow on a fiber stack" do
+    status, _, error = build_and_run <<-'CODE'
+      def foo
+        y = StaticArray(Int8,512).new(0)
+        foo
+      end
+
+      spawn do
+        foo
+      end
+
+      sleep 60.seconds
+    CODE
+
+    status.success?.should be_false
+    error.should contain("Stack overflow")
+  end
+end
--- a/src/fiber.cr
+++ b/src/fiber.cr
@@ -1,4 +1,7 @@
 require "c/sys/mman"
+{% unless flag?(:win32) %}
+require "c/sys/resource"
+{% end %}
 require "thread/linked_list"
 
 # Load the arch-specific methods to create a context and to swap from one
@@ -86,10 +89,25 @@ class Fiber
   # :nodoc:
   def initialize
     @proc = Proc(Void).new { }
-    @stack = Pointer(Void).null
     @stack_top = _fiber_get_stack_top
     @stack_bottom = GC.stack_bottom
     @name = "main"
+
+    # Determine location of the top of the stack.
+    # The technique here works only for the main stack on a POSIX platform.
+    # TODO: implement for Windows with GetCurrentThreadStackLimits
+    # TODO: implement for pthreads using
+    #    linux-glibc/musl: pthread_getattr_np
+    #              macosx: pthread_get_stackaddr_np, pthread_get_stacksize_np
+    #             freebsd: pthread_attr_get_np
+    #             openbsd: pthread_stackseg_np
+    @stack = Pointer(Void).null
+    {% unless flag?(:win32) %}
+      if LibC.getrlimit(LibC::RLIMIT_STACK, out rlim) == 0
+        stack_size = rlim.rlim_cur
+        @stack = Pointer(Void).new(@stack_bottom.address - stack_size)
+      end
+    {% end %}
 
     @@fibers.push(self)
   end
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -320,7 +320,18 @@ end
 # :nodoc:
 fun __crystal_sigfault_handler(sig : LibC::Int, addr : Void*)
   # Capture fault signals (SEGV, BUS) and finish the process printing a backtrace first
-  LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+
+  # Determine if the SEGV was inside or 'near' the top of the stack
+  # to check for potential stack overflow. 'Near' is a small
+  # amount larger than a typical stack frame, 4096 bytes here.
+  stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)
+
+  if stack_top <= addr < Fiber.current.@stack_bottom
+    LibC.dprintf 2, "Stack overflow (e.g., infinite or very deep recursion)\n"
+  else
+    LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+  end
+
   CallStack.print_backtrace
   LibC._exit(sig)
 end
--- a/src/lib_c/aarch64-linux-gnu/c/sys/resource.cr
+++ b/src/lib_c/aarch64-linux-gnu/c/sys/resource.cr
@@ -1,0 +1,12 @@
+lib LibC
+  alias RlimT = ULong
+
+  struct Rlimit
+    rlim_cur : RlimT
+    rlim_max : RlimT
+  end
+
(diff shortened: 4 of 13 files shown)


src/signal.cr
```diff
@@ -320,7 +320,17 @@ end
 # :nodoc:
 fun __crystal_sigfault_handler(sig : LibC::Int, addr : Void*)
   # Capture fault signals (SEGV, BUS) and finish the process printing a backtrace first
+
   LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+
+  # Determine if the SEGV was inside or 'near' the top of the stack
+  # to check for potential stack overflow. 'Near' is a small
+  # amount larger than a typical stack frame.
+  stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)
```
Sija: Maybe `4096` here should be a constant?

src/signal.cr
```diff
@@ -320,7 +320,17 @@ end
...
+  if stack_top <= addr < Fiber.current.@stack_bottom
+    LibC.dprintf 2, "Potential stack overflow (e.g., infinite or very deep recursion)\n"
```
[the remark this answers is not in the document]
asterite: Additionally, I would raise a `StackOverflowError` when this happen. Right now it's a bit slow to show the full trace for it, but we should improve that.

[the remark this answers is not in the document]
ysbaddaden: That's a good idea, but I don't think we can raise an exception from the SIGSEGV handler, because it runs on its own altstack, not the stack that segfaulted (it's full).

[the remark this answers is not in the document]
asterite: Ah, that's true... When I tried it I watched it print "Unhandled exception: ...", but I can't seem to `rescue` it. So printing an error and exiting is fine. After all, a stack overflow is probably a bug in the app and it usually happens when you start it.

src/signal.cr
damaxwell: I'm very sympathetic to your anti-magic number perspective. But since crystal won't allow constants to be declared inside a method, the declaration would have to be remote from its (one and only) use. That felt awkward. So I went with the precedent of `fiber.cr` and `io.cr` and used an explicit number. But I also tried to document what was going on ; I'll expand on that comment a little bit.

src/signal.cr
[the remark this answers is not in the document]
damaxwell: Forgive my overly cautious error messages.  I'll go out on a limb and claim (with certainty!) a stack overflow.

src/signal.cr
```diff
@@ -320,7 +320,18 @@ end
...
   # Capture fault signals (SEGV, BUS) and finish the process printing a backtrace first
-  LibC.dprintf 2, "Invalid memory access (signal %d) at address 0x%lx\n", sig, addr
+
+  # Determine if the SEGV was inside or 'near' the top of the stack
+  # to check for potential stack overflow. 'Near' is a small
+  # amount larger than a typical stack frame, 4096 bytes here.
+  stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)
+
+  if stack_top <= addr < Fiber.current.@stack_bottom
+    LibC.dprintf 2, "Stack overflow (invalid memory access at address 0x%lx)\n", addr
```
Sija: Perhaps expand _Stack overflow_ part into _Stack overflow: infinite or too deep recursion_ (copied from previous PR by @ysbaddaden)?

asterite: I mean, I wouldn't mix "stack ovefflow" with "invalid memory access" in the same message. Most developers don't know or don't care about this fact. Just "Stack overflow" is fine.

damaxwell: Hmm.  In @ysbaddaden 's  original work he also reported the address of the seg fault being handled, even when it was a stack overflow.  Easy enough to not do this and shorten the error message yet further.

asterite: Actually, don't listen to me.

damaxwell: Fair enough. ;-)

But your original point is very well taken: don't bury the main point (stack overflow) on the second line of the error message.  That's fixed.

ysbaddaden: Let's try to settle on something, then let's change the message once. My personal opinion:

- I'd keep the `infinite or too deep recursion` hint. It gives an explanation for the blunt `stack overflow`, which otherwise can be a little intimidating when not familiar with it.
- I'd drop the `invalid memory access at 0x...` since it's 99% likely to be a stack overflow, and debugging with gdb/lldb will give access to the actual address for the remaining 1%.

src/fiber.cr
```diff
@@ -86,11 +89,18 @@ class Fiber
...
     @stack_bottom = GC.stack_bottom
     @name = "main"
 
+    @stack = Pointer(Void).null
+    {% unless flag?(:win32) %}
+      if LibC.getrlimit(LibC::RLIMIT_STACK, out rlim) == 0
+        stack_size = rlim.rlim_cur
+        @stack = Pointer(Void).new(@stack_bottom.address - stack_size)
+      end
+    {% end %}
```
ysbaddaden: That should work for the main thread, but is it correct for created threads? I don't know what's the default stack size for them. Maybe we should use [pthread_getattr_np(3)](http://man7.org/linux/man-pages/man3/pthread_getattr_np.3.html) and [pthread_attr_getstacksize(3)](http://man7.org/linux/man-pages/man3/pthread_attr_getstacksize.3.html) for created thread?

RX14: Let's not worry about that for now until after threaded GC works. Otherwise it's stalling the PR.

ysbaddaden: Sure, but maybe we can add a comment about it :)

RX14: Yeah, a `TODO: ` would be nice

spec/std/kernel_spec.cr
```diff
@@ -220,3 +220,22 @@ describe "at_exit" do
                            OUTPUT
   end
 end
+
+describe "seg fault" do
+  it "detects stack overflow" do
```
RX14: How about a spec which stack overflows the main stack, not just a fiber stack?

How long does this spec run for (time)?

damaxwell: Each spec in `kernel_spec.cr` costs about 3 seconds apiece to run for a total of 36 seconds on my laptop.  The one I added is no different.  So I was leery about adding too many of these. 

In my actual testing on the various platforms I used the following:

* Random pointer deref does not trigger stack overflow
* Trivial infinite recursion on main stack triggers stack overflow
* Infinite recursion on main stack with modest stack allocations triggers stack overflow
* Trivial infinite recursion on fiber stack triggers stack overflow (this became the official spec)
* Infinite recursion on fiber with modest stack allocation triggers stack overflow

All of these take about the same amount of time, except on one of the BSDs (freebsd, I think) the default stack size is gargantuan and `ulimit` didn't seem to work to restrict it.  So testing for the main stack took an inordinate amount of time (a minute or two). This led to the choice of using the fibre stack for the official suite if there was going to be just one spec.

Please let me know which of these you'd like me to add.

RX14: @damaxwell I'd go for 1, 3 and 5. Speeding up the stack overflow by allocating a 512 byte staticarray should help the speed of the test on the BSDs, and testing main and fiber stacks is a good idea.

damaxwell: The new specs are added, @RX14.  It was FreeBSD that seems to have a default stack size of 0.5G, but `ulimit` works just fine to limit it; not sure what I had done wrong.   It takes 2 minutes (in a VM) to exhaust a stack that big. I left a comment in the spec to note the potential trouble and remedy.

ysbaddaden:
This spec fails systematically on Ubuntu 14.04 for me when running `make std_spec`:
```
  1) seg fault detects stack overflow on the main stack

     Failure/Error: error.should contain("Stack overflow")
       Expected:   "Invalid memory access (signal 11) at address 0x7ff92e7c2f08 <..snip...>"
       to include: "Stack overflow"

     # spec/std/kernel_spec.cr:249
```

damaxwell:
Thanks.  It was tested under Ubuntu 16.04 and passes there.  I'll get 14.04 up and running and see what's different.

damaxwell:
@ysbaddaden: the failure happens because gmake 3.81 sets the stack rlimit to RLIM_INFINITY=-1.  So our computation of the stack top is wrong.  Later versions of gmake reset the stack rlimit to something sane before forking out child processes; see  [this discussion](https://lists.gnu.org/archive/html/bug-make/2010-07/msg00123.html).  I'm looking into a fix.

asterite:
@damaxwell @ysbaddaden Maybe a bit late, but thank you so much for this! Segfaults that are stack overflows are probably the most common errors reported here, and now that traffic will probably be reduced a lot.

PR #4901, opened 2017-08-30
Make IO a class

asterite:
This changes `IO` to be a `class` instead of a `module`. The reasons are:

1. `IO` has an `@encoding` variable that can be set and changed. This makes `IO` as a struct very inconvenient because its mutable.
2. There are no struct `IO`s in the standard library.
3. Less code is generated (code for module dispatch is huge and repeats in every method call). This point can probably be improved by improving the compiler, but the other two points are still relevant. 

Because `Zip::File` now works with `IO` instead of a union of two specific IOs (and this can't be represented in the language because the types get merged to the closest ancestor) I added raising methods like `seek` and `pos`, which also make sense because all IOs should provide that interface, even if they can't effectively implement it (otherwise, let's break IO into many small interfaces like in Go, that's acceptable too but a much bigger and boring change).

PR #6929, opened 2018-10-11
(WIP) Refactor protocol sockets (TCP, UDP, UNIX)

### LINKED ISSUE #5778: Refactor Sockets API

straight-shoota:
Closes #5778

The protocol socket implementations previously all inherited `Socket` (and hence `IO`) which leads to problems in the type hierarchy. This is more detailed in the issue.

This PR splits the Sockets library into a lower level API `Socket::Raw` (previously `Socket`) which gives full access to the underlying OS socket implementation. Protocol sockets only wrap an instance of this raw socket and provide a protocol-specific API wrapper.Old type hierarchy:

Old type hierarchy:

* IO
  * Socket < IO::Buffered
    * TCPSocket
       * TCPServer < Server::Socket
    * UNIXSocket
      * UNIXServer < Server::Socket
    * UDPSocket

New type hierarchy:

* IO
  * Socket::Raw < IO::Buffered
  * TCPSocket
  * UNIXSocket
* TCPServer < Server::Socket
* UNIXServer < Server::Socket
* UDPSocket

**Implementation**

The protocol sockets are simply wrapping a `Socket::Raw` and delegating all socket interaction to that. They are immutable and could just be structs to be super-lightweight.
However, since `IO` is a class, and `TCPSocket` and `UNIXSocket` need to inherit `IO`, they have to be classes (for this use case we could think about revisting #4901 or something else).
All other implementations (server sockets and `UDPSocket`) are structs.

They sometimes share similar behaviour in delegating IO operations, TCP settings etc. but there is not really a point in creating a subordinate hierarchy for this. Instead, `src/socket/delegates.cr` provides a few macros that define common methods in the socket implementations. This is perhaps not really elegant, but the best solution I could come up with. They could later be expanded directly into the respective types to avoid macro calls in favour of a little bit code duplication, once we're settled on the general design.

Based on #6717


ysbaddaden:
I don't like `Socket::Raw`. It's confusing because it's not a RAW socket. I really prefer `Socket < IO` which makes much more sense in the general case. The refactor is far too complex for my taste: many macros that inject methods that delegate to composed objects...

Obviously, TCPServer/UNIXServer shouldn't inherit TCPSocket/UNIXServer but compose them, and just implement a few methods (`open` and `accept`), maybe UDPSocket should do same, but that's about it. No need to make it all more complex than necessary.

straight-shoota:
Yes, `Socket::Raw` could just be `Socket`. I'd prefer to not call it `Socket` and keep it more in the background because people shouldn't use it typically. But it could have a different name than `Raw`.

I explicitly mentioned that the macros can be removed later in the process, it just keeps it easier while refactoring this PR. But we just need a lot of delegators for this to work.

I don't get your last paragraph. Could you elaborate what you mean exactly?

straight-shoota:
Okay, let's please not bikeshed about the completely indifferent naming of a class. It should be clear to everyone what its purpose is. Naming can be discussed once we settle on the general API design.

src/socket/raw.cr
```diff
@@ -0,0 +1,693 @@
...
+  end
+
+  def getsockopt_bool(optname, level = LibC::SOL_SOCKET)
+    ret = getsockopt optname, 0, level
+    ret != 0
+  end
+
+  def setsockopt_bool(optname, optval : Bool, level = LibC::SOL_SOCKET)
+    v = optval ? 1 : 0
+    ret = setsockopt optname, v, level
```
j8r: `v` isn't needed:
`ret = setsockopt optname, (optval ? 1 : 0), level`

src/socket/raw.cr
```diff
@@ -0,0 +1,693 @@
...
+    @closed = false
+    fd = LibC.socket(family, type, protocol)
+    raise Errno.new("failed to create socket:") if fd == -1
+    init_close_on_exec(fd)
+    @fd = fd
+
+    self.sync = true
+    unless blocking
+      self.blocking = false
+    end
```
j8r: I would rather use `def self.new`, remove the last 5 duplicated lines along with `@closed = true`, and at the end call the protected initializer below `new(fd, family, type, protocol, blocking)`

src/socket/raw.cr
```diff
@@ -0,0 +1,693 @@
...
+    getsockopt_bool LibC::SO_KEEPALIVE
+  end
+
+  def keepalive=(val : Bool)
+    setsockopt_bool LibC::SO_KEEPALIVE, val
+  end
+
+  def linger
+    v = LibC::Linger.new
+    ret = getsockopt LibC::SO_LINGER, v
```
j8r: `v` can be avoided, use directly the method

src/socket/unix_server.cr
```diff
@@ -1,82 +1,151 @@
...
-    end
+  # Creates a named UNIX socket listening on *address*.
+  #
+  # Always deletes any existing filesystam pathname first, in order to cleanup
+  # any leftover socket file.
+  #
+  # ```
+  # UNIXServer.new(Socket::UNIXAddress.new("/tmp/dgram.sock"))
+  # ```
+  def self.new(address : Socket::UNIXAddress, *, mode : File::Permissions? = nil, backlog = 128) : UNIXServer
```
j8r: A common default, decently secure permission for sockets is `660`

wontruefree:
I like the changes to hierarchy

straight-shoota:
@bcardiff 

> change `TCPServer` and `UnixServer` to not inherit Socket and narrow the interface of them.

That's the main part of what this PR does.
It also removes `TCPSocket` and `UNIXSocket` inheriting from `Socket` to provide a clean API focused on what's relevant to the respective protocol.

> I see value in exposing as much functionality of the OS, but I would rather don't do it at the same time.

I'm not sure what you mean. This PR doesn't expose anything more or less than before. It just applies composition instead of inheritance to the socket implementations.

@j8r Thanks for your review, but I currently don't care about implementation details. We still need to figure out the API design. I marked your comments as resolved but will revisit them when it's time.

src/socket/ip_socket.cr
```diff
@@ -12,7 +20,15 @@ class IPSocket < Socket
     IPAddress.from(sockaddr, addrlen)
   end
 
+  # Returns the `IPAddress` for the remote end of the IP socket or `nil` if it
+  # is not connected.
+  def remote_address?
```
RX14: Is this really needed?

src/socket/raw.cr
```diff
@@ -483,6 +554,61 @@ class Socket::Raw < IO
     arg
   end
 
+  # Returns `true` if the Nable algorithm is disabled.
+  def tcp_nodelay?
```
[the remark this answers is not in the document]
straight-shoota: Oh great, I had intended to detail this option in the description but forgot to mention it. I'm fine with that :+1:

src/socket/ip_socket.cr
straight-shoota: I felt that I needed it at some point during refactoring the specs. It's not completely thought out, but it came in handy.
But it's certainly up for debate.

straight-shoota: Maybe it's even better to remove it from this PR and discuss it separately.

ysbaddaden:
`Socket.new` is the generic way to access sockets, without any arbitrary limitation or requiring to create a new type, composing a confusing new type, and require macros to inject method delegation. Like it or not, this is meant to stay —BSD sockets can be considered a confusing mess, I agree, yet limiting what we can achieve with them would be nonsensical.

The consensus among the core team is for the TCPServer abstraction to stop inheriting from TCPSocket, because a server isn't an IO steam. Nothing more.

Unless this is leading to something more complex than beneficiary, I'll gladly accept a pull request doing just that; same for UNIXServer of course.

I'm more wary of breaking the `UDPSocket < Socket` inheritance, that can quickly become a messy batch of delegated methods, and I'd like to see this tackled later.

RX14:
> The consensus among the core team

I disagree.

I don't think anyone is proposing arbitrary limitations to sockets here, in fact just to seperate concerns cleanly between a BSD-api `Socket` class and idiomatic Crystal `TCP::*` classes.

However, if separating the server classes is something we do all agree on, we can do that first. But I would like to see that as a first step not the last.

src/socket/tcp_server.cr
```diff
@@ -1,108 +1,212 @@
...
     begin
       yield server
     ensure
       server.close
     end
   end
 
-  # Creates a new TCP server, listening on all interfaces, and yields it to the
-  # block. Eventually closes the server socket when the block returns.
+  # Creates a new `TCPServer` listenting on *host* and *port*, and yields it to the block.
```
r00ster91: Typo: `listenting`

src/socket/unix_socket.cr
```diff
@@ -1,50 +1,60 @@
...
-      sock.close
+      socket.close
     end
   end
 
+  Socket.delegate_close
+  Socket.delegate_io_methods
+  Socket.delegate_buffer_sizes
+
   # Returns a pair of unamed UNIX sockets.
```
r00ster91: Typo: `unamed` should be `unnamed` and in the [other `self.pair`](https://github.com/crystal-lang/crystal/pull/6929/files#diff-e4de07adb127b6b3c4954da7e7e6b277R92) too.

straight-shoota:
I've added a commit that moves protocol-specific methods from `Socket::Raw` to the protocol socket types as proposed by @RX14 in https://github.com/crystal-lang/crystal/pull/6929#discussion_r224862609

This still uses macros for some of the duplicated code, but that's just for easier development and can be resolved later on in this PR so that the (duplicated) code lives directly inside the respective types.

I am not sure however about refactoring `IO` out of `Socket::Raw` because it is essentially designed as an `IO`. For now, I have left all IO-related methods in `Socket::Raw`.

wontruefree:
I still like `Socket` over `Socket::Raw` I know this was mentioned as bikeshedding earlier but I think it adds an empty namespace and I think it makes things less clear.

asterite:
Sorry if this was discussed before, but what's the problem with tcp socket and the other types being IOs?

asterite:
Also, class vs struct isn't very important here, one doesn't create thousands of sockets or servers.

straight-shoota:
@asterite `TCPSocket` should be an IO, but `TCPServer` or `UDPSocket` not.

asterite:
Then I don't understand the difference between Socket and Soxket::Raw

straight-shoota:
@ysbaddaden Crystal is a language with a focus on type safety. The stdlib currently makes this wobbly in places like this, where inheritance completely misguides the actual behaviour of a class. `UDPSocket` can't be used as an `IO` so it shouldn't inherit it. `TCPServer` is not a `TCPSocket` so it shouldn't inherit. etc. see #5778 for details. It's really confusing to have an API where protocol specific implementations inherit all the multi-purpose features of the parent `Socket` type making a lot of methods available where they don't make any sense. This obscures the purpose of these classes and makes it really hard for users to understand what they're *actually* supposed to do with them.

> The proposed solution introduces a complex refactor,

Well, the API is really wonky right now. This PR cleans it up. So the PR itself has some complexity (although it should be comprehensible when looking at the individual commits) but it reduces overall complexity by separating concerns and making the API simpler and thus easier to understand and use.

> that trades inherited methods for macros that inject lots of methods that merely delegate;

I've already mentioned it several times now that the delegator macros are not meant to stay. Please stop repeating this nonsense argument again and again.

> it introduces a confusing `Socket::Raw` type without any reason (just use Socket).

How is `Socket::Raw` in any way more confusing than current `Socket`? It's just been renamed because 90% of socket users won't use it directly. As already mentioned, the naming is completely irrelevant to the structural architecture and can be up for discussion. If you insist that `Socket::Raw` should stay being called `Socket`, I'm fine with that. I don't think it's the wisest idea, but it doesn't matter much.

asterite:
Is `Socket::Raw` exposed to the user? If not then the name doesn't mater. Otherwise, maybe it can be named `Socket::Base` or something like that.

straight-shoota:
@asterite Yes, it's a public API. It can be used to directly access the OS socket, for example for less common socket types (like Bluetooth connections, Unix datagram) where we don't provide protocol-specific socket implementations.

`Socket::Base` is also fine.

asterite:
So, I'm seeing now that `Socket` is a module and no longer a class. Can't `Socket::Raw` be renamed to `Socket` then? I think this will remove the confusion.

straight-shoota:
Yes, it could be renamed back to `Socket`. See https://github.com/crystal-lang/crystal/pull/6929#issuecomment-429013084

But discussing the naming that class is really not the main concern right now.

RX14:
If you look at the `TCPSocket` docs right now as a new user who doesn't know the BSD socket API already, you're lost. And that's a sign of a terrible API. The BSD sockets API is confusing. Added to that, the inheritance makes it such that all the important methods are lost from the page.

That's why I want to remove all the inheritance and replace it with composition. Not because of type safety, but because the current API is impossible to effectively document.

The socket API isn't meant to be object-oriented, so it shouldn't be. There should be one `Socket` class. Then, to make that `Socket` class easy to use, it makes sense to wrap that `Socket` class in `TCPClient` and `TCPServer` classes.

Making `Socket` not be an `IO` I don't *really* care about. If we don't want that, `Socket` can keep being an `IO`. 

So to set clear goals (for me):
- I want `Socket` to be a pure representation of the whole of the BSD API, and usable for weird edgecases, including platform-specific stuff like netlink sockets.
- I want `{TCP,UDP,UNIX}::{Client,Server}` to be simple to use and well documented classes which hide the complexities of the underlying sockets API.

The only way I can think to achieve that is to remove the inheritance.

src/socket/unix_server.cr
```diff
@@ -1,82 +1,163 @@
...
+
+  # Creates a named UNIX socket listening on a filesystem pathname.
   #
   # Always deletes any existing filesystam pathname first, in order to cleanup
   # any leftover socket file.
   #
-  # The server is of stream type by default, but this can be changed for
-  # another type. For example datagram messages:
   # ```
-  # UNIXServer.new("/tmp/dgram.sock", Socket::Type::DGRAM)
```
bcardiff: how is a `UNIXServer` created with `DRAM` type? Some docs left use `dgram.sock` but I am a bit lost how is that functionality preserved.

src/socket/delegates.cr
```diff
@@ -0,0 +1,180 @@
+module Socket
+  # :nodoc:
+  macro delegate_close
```
bcardiff: some of these macros could be modules that are included. If so, I'm ok with reusing code using that artifact.

src/socket/unix_server.cr
straight-shoota: Using `UNIXServer`/`UNIXSocket` with DGRAM type is no longer supported. It would be too weird and contradict the purpose of this PR if a stream socket (implementing `IO`) could be used as a datagram socket.

We could add an additional type for this but I don't think this use case is anywhere common enough for that. If you want to send datagrams over UNIX sockets, you should simply use `Socket::Raw`.

src/socket/delegates.cr
straight-shoota: Yeah, maybe. But I'm not sure it makes much sense. It doesn't hurt to just duplicate those methods (it's typically only two or three uses for each). I would only use a module if it communicates some sort of common behaviour that could be useful as an individual trait.

Maybe the IO delegator methods could be put in a `IO::Delegator` module. This might actually be reusable in other places (`HTTP::Server::Response`, `HTTP::UnknownLengthContent`, or `IO::Hexdump` come to mind).

straight-shoota:
100% agree to @RX14.

Compare https://crystal-lang.org/api/0.26.1/TCPSocket.html with https://straight-shoota.github.io/crystal/sockets-refactor/TCPSocket.html

 @bcardiff 

> I would still feel more comfortable if the initial work would be to only split the server vs socket issue.

This could probably be split in separate PR's. I'm happy to reorganize the commits if it helps to push this forward.

RX14:
This is still a big PR, is there any way we could split this up more once we're all agreed on the general approach?

For example, separating out some of the minor commits to a separate file, and I think at this point we're mostly agreed that `Socket` doesn't have to change names.

RX14:
Also, we can leave moving the getsockopt calls to the wrapper classes till later if ever. For now it's best to leave them on `Socket` and wrap them.

straight-shoota:
The first six commits are only preparatory and can be split off. In fact, I had already submitted #6707 and #6717 as solitary PRs but they didn't receive much appreciation. The other PRs created in preparation for this #6700 #6703 #6704 #6705 #6706 #6710 #6711 and #6909 have already been merged, though 👍 

> Also, we can leave moving the getsockopt calls to the wrapper classes till later if ever. For now it's best to leave them on Socket and wrap them.

I'm not sure if that's better. But we can leave it out and focus on porting the Sockets API to windows first. This needs fibers, though.

straight-shoota:
Closing stale PR.

PR #6941, opened 2018-10-14
Fix dead links in docs of Enumerable#chunks and Iterator#chunk

r00ster91:
Clicking on `Enumerable::Chunk::Drop` or `::Alone` leads to https://crystal-lang.org/api/0.26.1/Enumerable/Chunk/Drop.html which leads to https://crystal-lang.org/404 because `Enumerable::Chunk` is `:nodoc:`.


Sija:
Removing docs because of that doesn't seem like a solution. Perhaps it would be better to remove `:nodoc` from `Enumerable::Chunk` instead.

straight-shoota:
@r00ster91 Could you add a short description to each of the now expose types? A reference to `Enumerable#chunks` should suffice.

PR #6942, opened 2018-10-14
Change Installation section of README template to numbered list

r00ster91:
Follow-up of #6914 

Actually there's a step missing: `shards init` which creates `shard.yml`. But now after a third step, it would make sense to use a numbered list I think.
I didn't put a dot behind step **1** and **3** for consistency because there aren't dots either behind the steps in the **Contributing** section.

So now the **Installation** section would look like this:

---

## Installation

1. Run `shards init`
2. Add this to your application's `shard.yml`:
```yaml
dependencies:
  library:
    github: user/library
```
3. Run `shards install`

### DIFF
--- a/spec/compiler/crystal/tools/init_spec.cr
+++ b/spec/compiler/crystal/tools/init_spec.cr
@@ -93,13 +93,15 @@ module Crystal
       describe_file "example/README.md" do |readme|
         readme.should contain("# example")
 
+        readme.should contain(%{1. Run `shards init`.})
+        readme.should contain(%{2. Add this to your application's `shard.yml`:})
+
         readme.should contain(%{```yaml
 dependencies:
   example:
     github: jsmith/example
 ```})
-
-        readme.should contain(%{Then run `shards install`.})
+        readme.should contain(%{3. Run `shards install`.})
         readme.should contain(%{TODO: Write a description here})
         readme.should_not contain(%{TODO: Write installation instructions here})
         readme.should contain(%{require "example"})
@@ -110,14 +112,17 @@ dependencies:
       describe_file "example_app/README.md" do |readme|
         readme.should contain("# example")
 
+        readme.should contain(%{TODO: Write a description here})
+
+        readme.should_not contain(%{1. Run `shards init`.})
+        readme.should_not contain(%{2. Add this to your application's `shard.yml`:})
+
         readme.should_not contain(%{```yaml
 dependencies:
   example:
     github: jsmith/example
 ```})
-
-        readme.should_not contain(%{Then run `shards install`.})
-        readme.should contain(%{TODO: Write a description here})
+        readme.should_not contain(%{3. Run `shards install`.})
         readme.should contain(%{TODO: Write installation instructions here})
         readme.should_not contain(%{require "example"})
         readme.should contain(%{1. Fork it (<https://github.com/jsmith/example_app/fork>)})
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/tools/init/template/readme.md.ecr
```diff
@@ -5,15 +5,15 @@ TODO: Write a description here
 ## Installation
 
 <%- if config.skeleton_type == "lib" -%>
-Add this to your application's `shard.yml`:
+1. Run `shards init`
+2. Add this to your application's `shard.yml`:
```
j8r: Better to detail what "this" is:
> Add the dependency to your application's `shard.yml`

r00ster91: Let's exclude "application's". Because a `shard.yml` doesn't necessarily has to be used for just one application. I could just init the `shard.yml` on the desktop and use it for multiple applications.

So "Add the dependency to your `shard.yml`:"

RX14:
In fact, most of the time you don't have to run `shards init`, because you've used other shards before or you've used `crystal init`. So I don't think we should include this advice.

r00ster91:
Most of the time, but not always. I think the section shouldn't be incomplete and describe all steps to install the shard. But I agree, most of the time `shard.yml` is probably there.
Maybe "Run `shards init` unless `shard.yml` already exists" or something similar?

RX14:
I don't think we should assume that people don't know to use shards, there's obviously not enough space to explain the whole process of installing a shard.

asterite:
Looking at [one Ruby gem](https://github.com/ruby/csv) they say:

1. Add something to the Gemfile
2. Run `bundle`
3. (or install it with the `gem` command)

I think we just need to say:

1. Add to `shards.yml`
2. Execute `shards`

Nothing else. `shards init` is for initializing a project, and in fact if you use `crystal init` it will be there already.

spec/compiler/crystal/tools/init_spec.cr
```diff
@@ -113,14 +112,13 @@ dependencies:
...
-        readme.should_not contain(%{1. Run `shards init`})
-        readme.should_not contain(%{2. Add the dependency to your `shard.yml`:})
+        readme.should_not contain(%{1. Add the dependency to your `shard.yml`:})
         readme.should_not contain(%{```yaml
 dependencies:
   example:
     github: jsmith/example
 ```})
-        readme.should_not contain(%{3. Run `shards install`})
+        readme.should_not contain(%{2. Run `shards`})
```
ysbaddaden: Please keep the explicit `shards install`. The default action can change, and be removed to become a help message instead.

src/compiler/crystal/tools/init/template/readme.md.ecr
```diff
@@ -5,14 +5,13 @@ TODO: Write a description here
...
-1. Run `shards init`
-2. Add the dependency to your `shard.yml`:
+1. Add the dependency to your `shard.yml`:
 ```yaml
 dependencies:
   <%= config.name %>:
     github: <%= config.github_name %>/<%= config.name %>
 ```
-3. Run `shards install`
+2. Run `shards`
```
ysbaddaden: Same.

spec/compiler/crystal/tools/init_spec.cr
straight-shoota: Also it shows explicitly what this command is doing.

Sija:
@sdogruyol It seems since sometime you stopped squashing commits, which leads to pretty messy commit list, is it intentional?

sdogruyol:
@Sija nope it's not intentional. I just hate Github for changing my default merge strategy based on other repositories. I think eventually I need to write a Chrome plugin to avoid that.

PR #6946, opened 2018-10-16
Optimized Array#skip

asterite:
`Enumerable#skip` doesn't know the size of the resulting array, plus it's much efficient to do a mem-copy here.

A benchmark I did:

```crystal
require "benchmark"

a = Array.new(10_000) { |i| i }

Benchmark.ips do |x|
  x.report("old skip") do
    a.skip(1)
  end
  x.report("new skip 2") do
    a.skip2(1)
  end
  x.report("new skip 3") do
    a.skip3(1)
  end
end
```

where `skip` is the current one, `skip2` just called `[count..-1]` and `skip3` is the version in this PR.

Results:

```
  old skip  15.75k ( 63.48µs) (± 3.89%)  99072 B/op   2.62× slower
  new skip 2 41.28k ( 24.23µs) (± 6.37%)  40032 B/op   1.00× slower
new skip 3  41.31k ( 24.21µs) (± 6.05%)  40032 B/op        fastest
```

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -946,6 +946,24 @@ class Array(T)
     self
   end
 
+  # Returns an `Array` with the first *count* elements removed
+  # from the original array.
+  #
+  # If *count* is bigger than the number of elements in the array, returns an empty array.
+  #
+  # ```
+  # [1, 2, 3, 4, 5, 6].skip(3) # => [4, 5, 6]
+  # ```
+  def skip(count : Int) : Array(T)
+    raise ArgumentError.new("Attempt to skip negative size") if count < 0
+
+    new_size = Math.max(size - count, 0)
+    Array(T).build(new_size) do |buffer|
+      buffer.copy_from(to_unsafe + count, new_size)
+      new_size
+    end
+  end
+
   # Returns an `Array` with all possible permutations of *size*.
   #
   # ```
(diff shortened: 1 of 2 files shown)


asterite:
BTW, maybe we should rename `skip` to `drop`, it reads better. Also for `Iterator`... I'm thinking that if we want a bang version for this, `skip!` reads badly, even `skip` reads badly already.

yxhuvud:
"BTW, maybe we should rename skip to drop, it reads better."

Also, `skip` is not an antonym to the opposite, ie `take`. `drop` works better from that view too.

src/array.cr
```diff
@@ -946,6 +946,17 @@ class Array(T)
     self
   end
 
+  # Optimized version of `Enumerable#skip`.
```
straight-shoota: Shouldn't this better repeat the description of `Enumerable#skip`?

asterite: Yes. Actually, we should have some thingy to inherit doc from the parent. Let's work on that next (there are already a few methods in `Array` that say they are an optimized version of something else)

asterite:
@yxhuvud Yeah, `take` is something else we need. Right now we have `first`. The problem is that if we have `take` and `first` they will be aliases... maybe it's not a big deal. Maybe having aliases is not that bad after all.

src/array.cr
bew: For now I think it's better indeed, but it would be nice to have a `:previous_doc:` and/or `:super_doc:` to automatically copy the doc from previous def or super def or that same method

asterite: I just copied the doc.

yxhuvud:
Ah right. We have `first` instead of `take`, but still have `take_while`. That is a bit confusing.

Issue #6948, opened 2018-10-16
Channel::Unbuffered#send doesn't raise when resumed from close


ysbaddaden:
I noticed that closing a `Channel::Unbuffered` with a pending send merely resumes the sender fiber instead of raising an exception. This is surprising, because the value wasn't delivered, and further attempts to send a value will raise a Channel::Closed exception.

Is the behavior correct? or is it a bug, and we should raise when the value wasn't delivered? but maybe we can't be sure that the value was actually delivered?

To reproduce:

```crystal
ch = Channel::Unbuffered(Nil).new

spawn do
  ch.send(nil)   # will simply return
  # ch.send(nil) # will raise Channel::ClosedError
end

spawn { ch.close }
sleep(0)
```

bcardiff:
I think we can't be sure it is delivered unless there is a timeout. The unbuffered channel can hold one long living value waiting to be received, sometime in the future.

If the channel is closed at the moment of sending, then yes, an exception can and must be raised.

straight-shoota:
This has been fixed at some point:
```cr
ch = Channel(Nil).new

spawn do
  ch.send(nil) # raises Channel::ClosedError
end 

spawn { ch.close }
sleep(0)
```

PR #6953, opened 2018-10-17
Fix: some specs rely on Fiber.yield behavior

ysbaddaden:
Some specs are relying on the `Fiber.yield` behavior, that's actually the event loop behavior (libevent2). They start failing whenever the `Fiber.yield` algorithm changes —for example to push the current fiber to the runnables queue, instead of adding a resume event.

This patch proposes changes to make the fiber synchronization expectations explicit. Either by looping until somethig is ready, or using Channel::Unbuffered's sync ability, or with an explicit enqueue/resume for specs testing Channel itself.

### DIFF
--- a/spec/std/channel_spec.cr
+++ b/spec/std/channel_spec.cr
@@ -1,5 +1,10 @@
 require "spec"
 
+private def yield_to(fiber)
+  Crystal::Scheduler.enqueue(Fiber.current)
+  Crystal::Scheduler.resume(fiber)
+end
+
 describe Channel do
   it "creates unbuffered with no arguments" do
     Channel(Int32).new.should be_a(Channel::Unbuffered(Int32))
@@ -40,17 +45,22 @@ describe Channel::Unbuffered do
   it "blocks if there is no receiver" do
     ch = Channel::Unbuffered(Int32).new
     state = 0
-    spawn do
+    main = Fiber.current
+
+    sender = Fiber.new do
       state = 1
       ch.send 123
       state = 2
+    ensure
+      yield_to(main)
     end
 
-    Fiber.yield
+    yield_to(sender)
     state.should eq(1)
     ch.receive.should eq(123)
     state.should eq(1)
-    Fiber.yield
+
+    sleep
     state.should eq(2)
   end
 
@@ -67,8 +77,8 @@ describe Channel::Unbuffered do
     ch = Channel::Unbuffered(Int32).new
     ch.full?.should be_true
     ch.empty?.should be_true
-    spawn { ch.send 123 }
-    Fiber.yield
+    sender = Fiber.new { ch.send 123 }
+    yield_to(sender)
     ch.empty?.should be_false
     ch.full?.should be_true
     ch.receive.should eq(123)
@@ -88,8 +98,8 @@ describe Channel::Unbuffered do
 
   it "can send and receive nil" do
     ch = Channel::Unbuffered(Nil).new
-    spawn { ch.send nil }
-    Fiber.yield
+    sender = Fiber.new { ch.send nil }
+    yield_to(sender)
     ch.empty?.should be_false
     ch.receive.should be_nil
     ch.empty?.should be_true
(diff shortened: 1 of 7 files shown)


spec/std/channel_spec.cr
```diff
@@ -138,27 +157,44 @@ describe Channel::Unbuffered do
...
-    spawn do
-      ch.send nil
-      ch.send nil
-    rescue Channel::ClosedError
-      sender_closed = true
+    closed = false
+    main = Fiber.current
+
+    sender = Fiber.new do
+      ch.send(nil)
```
ysbaddaden: Related to #6948 : I'd expect `ch.send` to raise here since it couldn't deliver the value.

straight-shoota:
The behaviour provided by the spec helpers `yield_to` and `wait_for` is actually pretty generically usable. I'd suggest to include them into the stdlib API.

Maybe `Fiber#resume` should even enqueue the current fiber by default. If you don't want that, use `Crystal::Scheduler.resume` directly.

spec/std/mutex_spec.cr
```diff
@@ -26,9 +26,9 @@ describe Mutex do
     mutex = Mutex.new
 
     a = 1
-    two = true
-    three = true
-    four = true
+    two = false
+    three = false
+    four = false
```
RX14: Why?

ysbaddaden:
Fiber synchronization should be achieved through synchronous channels, as we do for the `future` and `lazy` specs; `yield_to` is only there to test Channel itself. We could abstract `wait_for`, it's a spin-lock with a timeout, but this isn't the goal of this PR.

`Fiber#resume` and `Crystal::Scheduler.resume(fiber)` having different semantics would only bring confusion. Maybe `Fiber#resume` will be removed, maybe not. Again, this isn't the topic of this PR.

spec/std/mutex_spec.cr
ysbaddaden: Because we eventually assert that they'll be `true`, so I wanted to also make sure the variables were changed —meaning that the fibers have run. We could even initialize them to `nil`.

bcardiff: I guess because the assertions at the end are `should be_true` so the test might have passed before if none of the spawns were executed.

spec/std/openssl/ssl/server_spec.cr
```diff
@@ -67,8 +67,6 @@ describe OpenSSL::SSL::Server do
           client.close
         end
 
-        Fiber.yield
```
bcardiff: For this spec, how can we be sure that we actually stressed the path we want?

That happened to me while testing crystal-db and I made a shameless helper `with_witness` that will fail if the `#check` is not called within the block. Because when the `.should` are inside the block, it might never get called.

```crystal
it "..." do
  with_witness do |w|
    spawn do
      some.should eq "..."
      w.check
    end
    # do something else, but we know the spawn 
    # is called before exiting the spec if we don't 
    # wrap with a with_witness?
  end
end
```

spec/std/concurrent/future_spec.cr
```diff
@@ -34,13 +34,12 @@ describe Concurrent::Future do
 
   describe "future" do
     it "computes a value" do
-      chan = Channel(Int32).new(1)
+      chan = Channel::Unbuffered(Int32).new
```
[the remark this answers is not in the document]
ysbaddaden: An unbuffered channel is synchronous, so it forces main/fiber to sync as we expect it. Unlike Buffered that are async.

Unrelated to this PR, but I think both buffered/unbuffered should have both (a)synchronous modes —in fact, I would have just buffered channels with a blocking argument to select sync/async.

spec/std/openssl/ssl/server_spec.cr
ysbaddaden: By sending a response and reading it?

spec/std/channel_spec.cr
```diff
@@ -40,17 +45,22 @@ describe Channel::Unbuffered do
...
     end
 
-    Fiber.yield
+    yield_to(sender)
     state.should eq(1)
     ch.receive.should eq(123)
     state.should eq(1)
-    Fiber.yield
+
+    sleep
```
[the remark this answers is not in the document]
bcardiff: :-( ok, let's keep the sleep if there is a segfault on double enqueue...

spec/std/concurrent/future_spec.cr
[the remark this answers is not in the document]
bcardiff: Ok, it's true the behavior is different. And that change does not affect the spec.

I think is more clear to have the async/sync nature in the type and not in the type argument.

spec/std/openssl/ssl/server_spec.cr
bcardiff: But all the code happen in blocks. This spec is `nop` if open don't yield to the block.
Before with the `Fiber.yield` it seems that there was an explicit path that was intended.

I know the flaw was present before actually. But wont such help be an idiom worth to use for this an maybe other?

ysbaddaden: Probably, but in this case expecting a response was simpler, and it enhances the spec a little —we can communicate both ways.

PR #3582, opened 2016-11-26
[WIP] Windows

lbguilherme:
> **Warning: This is very very initial.**

Fixes #26 

## Roadmap:

- [x] Make it compile code
- [ ] Add `{% if flag?(:windows) %}` all over the core library and make the prelude portable
  - [x] Basic IO
  - [x] Fibers
  - [x] Concurrent IO
  - [ ] Exceptions (backtraces)
  - [x] Threads
  - [ ] Mutex
  - [ ] Process
  - [ ] Regex (should work out of the box, but needs to compile libpcre)
  - [x] GC
  - [ ] Encodings
  - [ ] ...
- [ ] Port rest of std library
  - [ ] DNS
  - [ ] Sockets
  - [ ] FileUtils and Dir
  - [ ] ...
- [ ] Remove dependency on `wine clang.exe` hack
- [ ] Compile `crystal.exe` and self-host it
- [ ] Pass all specs
- [ ] Figure out how to do CI testing for Windows
- [ ] Add support for 32-bit Windows as well
- [ ] Release Crystal for Windows!

## How to use it: (from Linux)

1. Get the branch `win` from this fork:

       git clone https://github.com/lbguilherme/crystal.git
       git checkout win

2. Install Wine: https://www.winehq.org/
3. Download Clang 3.8.1 for Windows x64. Install it somewhere with Wine. Make sure this works:

       $ wine bin/clang.exe --version
       clang version 3.8.1 (branches/release_38)
       Target: x86_64-pc-windows-msvc

4. Install MXE from my fork (I'll get this patch upstream at some point https://github.com/mxe/mxe/pull/1539)

       git clone https://github.com/lbguilherme/mxe.git
       git checkout updated-gc
       # Check http://mxe.cc/#requirements for a list of dependencies
       # (or just run the comand below and get an error listing what is missing)
       make MXE_TARGETS=x86_64-w64-mingw32.static gcc gc libevent libiconv # This will take some considerable time

5. Build the crystal compiler having at least LLVM 3.8 in your system. Or maybe download some older clang.exe, compactible with the LLVM in your Crystal. If needed, build it: `make`

6. Write some simple Crystal:

       # foo.cr
       ch = Channel(String).new

       spawn do
         3.times { puts ch.receive }
       end

       ch.send "Hello Crystal"
       ch.send "  running on Windows!"
       sleep 1
       ch.send "\nYEAH!!"

7. And finally: Build code!

  ```sh
[...]

PR #6955, opened 2018-10-18
Extract platform-specifics of Fiber, Thread and EventLoop

straight-shoota:
This is a step to prepare `Fiber` for Windows support.


src/crystal/system/fiber.cr
```diff
@@ -0,0 +1,16 @@
+module Crystal::System::Fiber
+  # Allocates memory for a stack.
+  # def self.allocate_stack(stack_size : Int) : Void*
+
+  # Frees memory of a stack.
+  # def self.free_stack(stack : Void*, stack_size : Int) : Nil
+
+  # Determines location of the top of the main fiber's stack.
```
RX14: This is actually the main *process* stack. Happens to also be the main fiber stack.

RX14:
It doesn't abstract mmap, it abstracts a certain usage of mmap, particularly, allocating stacks.

I don't think that should be the same abstraction as other mmap usages.

straight-shoota:
@ysbaddaden Sure, it can be renamed to be more general. What about `main_fiber_stack`, though? It needs to be platform specific but it's doesn't really belong to memory mapping?

Adding the windows implementation right away is not an issue. I already have that code. It compiles but I can't validate it, because event handling hasn't been ported. I don't think it makes much sense to add dedicated specs for this?

src/crystal/system/unix/fiber.cr
```diff
@@ -0,0 +1,42 @@
...
+    {% end %}
+
+    LibC.mmap(nil, stack_size, LibC::PROT_READ | LibC::PROT_WRITE, flags, -1, 0).tap do |pointer|
+      raise Errno.new("Cannot allocate new fiber stack") if pointer == LibC::MAP_FAILED
+
+      {% if flag?(:linux) %}
+        LibC.madvise(pointer, stack_size, LibC::MADV_NOHUGEPAGE)
+      {% end %}
+
+      LibC.mprotect(pointer, 4096, LibC::PROT_NONE)
```
straight-shoota: Shouldn't this use `stack_size`?

straight-shoota:
I added the win32 implementation as far as I could get. I can't figure out the assembly for actually swapping the context on x86_64-win32. This is really not my area of expertise.

I'd be happy if somebody can take over this part. You guys now much better what to do.

src/crystal/system/win32/fiber.cr
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def self.free_stack(stack : Void*, stack_size) : Nil
+    if LibC.VirtualFree(stack, stack_size, LibC::MEM_RELEASE) == 0
+      raise WinError.new("VirtualFree")
+    end
+  end
+
+  def self.main_fiber_stack(stack_bottom : Void*) : Void*
+    LibC.GetCurrentThreadStackLimits(out lowLimit, out highLimit)
```
Sija: snake_case plz, `lowLimit` - > `low_limit`, `highLimit` -> `high_limit`

ysbaddaden:
Good job at adding Win32 support overall. I didn't review correctly, but that looks nice!

I stand to my original point: abstracting `mmap` as "anonymous memory mapping" and making it win32-compatible would be much more useful than moving a few methods to allocate stacks, then do the same for the GC, and again for other usages, for example:
- https://github.com/ysbaddaden/gc/blob/master/include/memory.h
- https://github.com/ysbaddaden/crystal/commit/b067a0f1da5140a9588894e680effe33f5de782b#diff-67b7365674b2cd768e530fdd8525b3dbR50

What differs are subsequent calls, but the `madvise` NOHUGEPAGE on Linux looks like a good option to be applied by default, and `mcollect` can be abstracted to protect a part of a memory mapping (win32: `VirtualProtect`). All that without requiring anything specific in `Crystal::System` for (de)allocating stacks.

Also, detecting a thread's stack should be the responsibility of Thread, not Fiber, that could ask the thread or be passed by the Thread when it initializes its main Fiber —I missed that in the stack overflow detection pull request.

Last but not least: I didn't know what I was doing while meddling with the arch-specific assembly, but I searched for reference, experimented with gdb and registers, and eventually learned a lot. Hint: look at ldc coroutines (D compiled based on LLVM) and search for win32 callee saved registers. For example https://en.wikipedia.org/wiki/X86_calling_conventions

RX14:
Assuming that a mmap even exists on the target platform is a mistake. mmap is a hugely complex syscall and the flags you want to use *even if all kernels supported the same flags* would differ depending on kernel behavior. "Allocate this amount of heap memory from the OS and return a pointer to the start" is a very different operation to "Allocate a stack" and there's no reason to put them through the same method. There's no need to expose madvise and mprotect.

`Crystal::System` should be the lowest common denominator for platform dependencies, and mmap is the opposite of that. Please don't think in terms of just POSIX and windows. They're hardly the only porting targets out there.

ysbaddaden:
@RX14 Do you have an example of a target that doesn't implement anonymous memory mappings, that allocates pages when they're accessed?

I don't intend to have a complete mmap interface, but the minimum we'll need in core/stdlib. See for example: https://github.com/ysbaddaden/crystal/commit/29547dc21ac6e1560848f2e8d335358ced2c81ad

straight-shoota:
@ysbaddaden Your example looks quite complicated. It adds a lot of specifics of the memory mapping API to the general `Fiber` implementation:

```cr
Crystal::System::MemoryMapping.new(STACK_SIZE, :READ_WRITE, :STACK).tap do |mapping|
  mapping.advise(:NOHUGEPAGE)
  mapping.protect(GUARD_SIZE, :NONE)
end
```

I'd prefer to keep the interface between `Fiber` and the platform-specific implementation as small as possible. It much simpler to have only `Crystal::System::Fiber.allocate_stack(STACK_SIZE)` (this could eventually be implemented as re-using `Crystal::System::MemoryMapping`).

RX14:
`Crystal::System` should be as *inflexible* as possible, because interfaces being inflexible keeps them simple and easy to implement on new platforms.

Allocating a fiber stack is an operation which you can easily stub, or make special platform-specific alterations for. mmap is a complex **unix** interface with a lot of complexity and a lot of assumptions.

ysbaddaden:
I'd still like a `Stack` struct (pointer + size) instead of a raw pointer, with a few methods:
1. new() to create the allocation;
2. protect() to declare guard page(s) explicitly, to avoid burying this as "implementation detail" —it's not, we expect them in the stack overflow detection;
3. free() to dispose of the allocation.

And ideally:

4. reserve() to grow pages manually, since Win32 can't do it automatically.

This could allow fibers to have differently sized stacks —defined manually or better automatically thanks to some compiler analysis, to reduce memory pressure on Win32 and virtual memory on POSIX.

We could have both a `Stack` and a `MemoryMapping` structs, but they're just the same to me: a low-level `Allocation`. Fiber doesn't care about `reserve()`... until we find a way to grow stacks (who knows), and a GC won't care about `protect()`... unless it wants to secure bounds.

RX14:
@ysbaddaden I'd be fine with a `Crystal::System::Memory.new_stack(size)` and `Crystal::System::Memory.allocate` which both return the same struct. We already have to deal with `MAP_STACK` on openbsd so there obvious is some differences between stack and general-purpose allocations without `MADV_NOHUGEPAGE`

rishavs:
@ysbaddaden any resolution on this topic? Its been a couple of months since the last comment and I was wondering if there were other discussions on this offline. Would be nice to move this PR along.

straight-shoota:
@rishavs Finding an agreement on the interface is not really an urging matter as it won't really make a noticeable change from a usage point of view. The important thing is to figure out the assembly code for a fiber switch on win32. That will have an impact as it brings us closer to running fibers on windows.

That being said, it would of course be great if we could settle on the interface and proceed from there.

src/kernel.cr
```diff
@@ -470,15 +470,15 @@ class Process
...
-  # Background loop to cleanup unused fiber stacks.
-  spawn do
-    loop do
-      sleep 5
-      Fiber.stack_pool_collect
-    end
+# Background loop to cleanup unused fiber stacks.
+spawn do
+  loop do
+    sleep 5
```
RX14: sleep isn't implemented on windows?

src/lib_c/x86_64-windows-msvc/c/basetsd.cr
```diff
@@ -4,4 +4,5 @@ lib LibC
   {% else %}
     alias ULONG_PTR = ULong
   {% end %}
+  alias PULONG_PTR = ULONG_PTR*
```
RX14: Don't make any of the `Pxxx` aliases, just use `xxx*`. It makes it much clearer whats going on.

src/thread/mutex.cr
```diff
@@ -1,55 +1,5 @@
...
-      raise Errno.new("pthread_mutex_destroy", ret) unless ret == 0
-    end
-
-    def to_unsafe
-      pointerof(@mutex)
-    end
-  end
-end
+{% if flag?(:win32) %}
+require "./mutex_win32"
```
RX14: No thanks, please abstract this to `Crystal::System` properly.

src/thread/mutex_win32.cr
```diff
@@ -0,0 +1,8 @@
+# TODO: Implement for multithreading.
+class Thread
+  class Mutex
```
RX14: If it's not implemented properly, it should be in the stubs file and the real code shouldn't be required in the prelude

src/concurrent.cr
```diff
@@ -3,6 +3,7 @@ require "channel"
 require "crystal/scheduler"
 require "./concurrent/*"
 
+{% unless flag?(:win32) %}
```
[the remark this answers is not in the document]
straight-shoota: This is already in `windows_stubs.cr`. You added it yourself 😁

src/kernel.cr
straight-shoota: It is in `windows_stubs.cr`.

src/concurrent.cr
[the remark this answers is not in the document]
RX14: Oops

RX14:
The `thread_win32` and `thread_pthread` still need to go. Are you sure this PR depends on `Thread` more than just a stub? I'd much prefer to discuss properly `Crystal::System`ing `Thread` in another PR.

straight-shoota:
This PR doesn't really add anything substantial, it merely just reorganizes code. It should be fine to revert to the original `threads.cr`. Fibers don't yet work on win32 anyway.

straight-shoota:
I have rebased this branch on master, building on top of the changes introducing `StackPool` and the updated `Threads` integration.

The actual fiber switch implementation for win32 x86_64 is not included. I can't get it to work. My current status is available in a separate branch: [fiber-win32](https://github.com/straight-shoota/crystal/tree/feature/fiber-win32) ([compare](https://github.com/straight-shoota/crystal/compare/jm/feature/fiber-prepare-port...straight-shoota:feature/fiber-win32)).
It uses the same approach as the original implementation in #3582 but the program seems to fail as soon as the asm code is included in the compiled binary.

Maybe it helps to somehow figure this out. Would be great if someone could take a look at it ❤️ 
/cc @lbguilherme @sorenbug @txe

wrejcodes:
@straight-shoota I just started looking through this thread this morning so I don’t know if this is going to be super helpful or not. But it looks like in your windows asm you have a typo. It should be movq $1, %rsp instead of movq ($1), %rsp based on what I saw from the other context swap implementations.

straight-shoota:
Yeah, I've noticed that as well. But I copied the code from the original implementation in #3582 
(see https://github.com/crystal-lang/crystal/pull/3582/files#diff-3633a2efaac123bcc373f98d811942c3R189) which supposedly worked. I tried without parenthesis as well, so there must be at least some other error anyway.

cfsamson:
@straight-shoota I think you also need to store NT_TIB stack info members to make this work properly or else Windows might kill the process. I accidentally wrote a book about context switches a few weeks ago where I do this in Rust on Macos, Linux and Windows. Rust also use the LLVM backend so the inline assembly and information should still be valid for Crystal. I also had to pass the `alignstack` option to LLVM for it to work correctly on Windows too. 

Here is a link to the chapter that explains it: https://cfsamson.gitbook.io/green-threads-explained-in-200-lines-of-rust/supporting-windows

You'll find reference to another article pointing out the same thing and a direct link to the Boost Context assembly that also stores NT_TIB information to the `%gs` register. 

I could try this out and test for you but I was just checking how you're doing on Windows support and I don't really know if this is something you still need. Let me know if there is anything I can do.

bcardiff:
@cfsamson this shows how to use asm options. https://github.com/crystal-lang/crystal/blob/03e6eba26879c778a4ac18c6e3df60c45d473e4b/spec/compiler/parser/parser_spec.cr#L1410

cfsamson:
@bcardiff Ok, so that's exactly the same as Rust. I think if we build upon the code you already got by using [Boost Context](https://github.com/boostorg/context/blob/develop/src/asm/ontop_x86_64_ms_pe_gas.asm) as a reference, and merging that with the inline asm I already got working with Rust we should have a good chance of getting the context switch working properly (on x86-64 at least) fairly quickly.

Right now it seems that https://github.com/straight-shoota/crystal/tree/feature/fiber-win32 is closest to making this work so that might be the best place to start? 

And lastly, testing this on Windows, If I cross-compile the compiler for Windows, is there a set of tests targeting fibers only?

bcardiff:
@cfsamson 

> Right now it seems that https://github.com/straight-shoota/crystal/tree/feature/fiber-win32 is closest to making this work so that might be the best place to start? 

It seems so, but since that is based on this PR branch and there is conflict, I would start with a rebase.

> And lastly, testing this on Windows, If I cross-compile the compiler for Windows, is there a set of tests targeting fibers only?

No, there is no suite specific to fibers AFAIK. channel_spec.cr will definitely stress it, but simple scheduling example will probably exhibit some issues. `aFiber.resume` should force a context switch so you can build small experiments.

I guess you already went through https://github.com/crystal-lang/crystal/wiki/Porting-to-Windows that explains the current state and procedure on how to cross compiler some small programs. Using `--prelude=empty` will allow you to avoid the whole stdlib.

@RX14 and @straight-shoota shold have more recent experience regarding windows.

RX14:
the best way to test is to cross-compile some manual tests and keep going until it doesn't segfault.

cfsamson:
@bcardiff and @straight-shoota Here's a short status update.

1. I couldn't get the [feature/fiber-win32 branch](https://github.com/straight-shoota/crystal/tree/feature/fiber-win32) to build, and rebasing it if anything made that more difficult. I started debugging it but only got deeper and deeper into things I really didn't want to change. I did however look through the code and it looked fine. 

2. I decided to pull out the interesting part of the problem, the context switch in Windows, and try to get the inline assembly to work. 

    [I made it all available here as a gist.](https://gist.github.com/cfsamson/8229407feefe91ef21a05aaa9b224729#file-crystal_fibers_asm_test-cr) It's a extremely simplified version of crystals context switch, focusing mostly on `makecontext` and `swapcontext` for `x86-64` on POSIX and Windows. 

    The advantage is that it's much easier to get the assembly working and cross compile these ~200 lines of code and then pull in `makecontext` and `swapcontext` instead of building the whole compiler every time. 

3. Crashing on windows
   The strange issue is that the code works on Ubutnu, and the program cross compiles to Windows, but it will not even start correctly and I couldn't get to the bottom of the problem profiling it either. You'll notice that it doesnt even print out the first line of text before crashing. 

    The assembly for Windows in this file _should_ work. Granted, I haven't stored the `XMM` registers yet but that's not really needed if we don't use them. I have no idea what's causing this to crash but it looks very similar to [https://github.com/crystal-lang/crystal/pull/6955#issuecomment-482795645](https://github.com/crystal-lang/crystal/pull/6955#issuecomment-482795645).

Hopefully it's easier to figure out what's happening using the code I created, and I wanted to share it if anyone else is interested to have a look at whats going on. I will investigate a bit further and see if I figure it out.

cfsamson:
@bcardiff @RX14 I have concluded that this most likely is not an issue with the inline assembly or the code beloning to the context switch but a problem in Crystal. I boiled down the issue as much as I could and resorted to implementing the same minimal example in both Rust and Crystal. In Rust it works as exepected on Windows and Linux, Crystal will not work with in Windows with the same code.

I tried to make a breif summary of the problem and presented how far I got with my research in https://github.com/crystal-lang/crystal/issues/7947. Hope it helps.

cfsamson:
@bcardiff @straight-shoota @RX14 
Good news. I found out what's causing the trouble mentioned in comment https://github.com/crystal-lang/crystal/pull/6955#issuecomment-482795645.

> The actual fiber switch implementation for win32 x86_64 is not included. I can't get it to work. My current status is available in a separate branch: fiber-win32 (compare).
It uses the same approach as the original implementation in #3582 but the program seems to fail as soon as the asm code is included in the compiled binary.

 See https://github.com/crystal-lang/crystal/issues/7947#issuecomment-508936915 for a more thorugh explanation (the issue title is now wildly misleading).

## TLDR:
Somehow calling `puts` from a new thread seems to cause an issue in Windows locking `stdout` so the program silently breaks. No **messages** OR **errors** are written so it "seems" like everything is brroken. I would assume there is some lock contention going on. The short term solution I got working is to instead of using `puts` like this:

```crystal
puts "spawn"
fiber = Fiber.new do
  puts "fiber says hello"
end

Crystal::Scheduler.yield(fiber)
```
Do this:
```crystal
puts "spawn"
fiber = Fiber.new do
  STDOUT << "fiber says hello\n"
  STDOUT.flush
end

Crystal::Scheduler.yield(fiber)
``` 

Now I haven't tested this myself on this implementation, since I cant get the branch to build, but I would guess that this is basically all that's needed to get the current implementation up an running. The next step is to fix the issue with `puts`.

## Proof of concept
That means the context switch is working. This gist example shows pretty much everything that needs to be done and is a provable example that it works as expected on Windows if you take the time to cross compile it (Note: we still need to save the XMM6-XMM15 registers for it to be finished).

[Gist to the working context switch on x86-64 Linux and Windows.](https://gist.github.com/cfsamson/3f3c694f7e235226f02188ee2f284c37)

RX14:
@straight-shoota are you interested in finishing this PR?

asterite:
I really hope @straight-shoota didn't get mad about Crystal for some reason, because he's such a great contributor. For example I know he didn't agree with my decision in #7809 and his last contribution was around that time...

straight-shoota:
No worries, I'm still around and happy to proceed with this. ;) I just couldn't get around here much in the last weeks.

Thank you very much @cfsamson, I'll be able to look into your findings in more detail in the next days. 
Sorry for the wait.

cfsamson:
@straight-shoota Perfect. I would start where you left off since as far as I can understand that code should work.

However I updated the gist[ I linked to above](https://gist.github.com/cfsamson/3f3c694f7e235226f02188ee2f284c37) just now adding everything you need (i hope) to also store/restore the `xmm` registers. The important takeaway is that the inline assembly looks like this:

```asm
asm("
        pushq %rcx
        pushq %gs: 0x10
        pushq %gs:0x08
        pushq %rdi
        pushq %rbx
        pushq %rbp
        pushq %rsi
        pushq %r12
        pushq %r13
        pushq %r14
        pushq %r15

        # XMM registers
        subq $$160, %rsp
        movups %xmm6, (%rsp)
        movups %xmm7, (%rsp)
        movups %xmm8, (%rsp)
        movups %xmm9, (%rsp)
        movups %xmm10, (%rsp)
        movups %xmm11, (%rsp)
        movups %xmm12, (%rsp)
        movups %xmm13, (%rsp)
        movups %xmm14, (%rsp)
        movups %xmm15, (%rsp)

        movq %rsp, ($0)

        movq $1, %rsp

        # XMM registers
        movups (%rsp), %xmm6
        movups (%rsp), %xmm7
        movups (%rsp), %xmm8
        movups (%rsp), %xmm9
        movups (%rsp), %xmm10
        movups (%rsp), %xmm11
        movups (%rsp), %xmm12
        movups (%rsp), %xmm13
        movups (%rsp), %xmm14
        movups (%rsp), %xmm15
        addq $$160, %rsp
        
        popq %r15
        popq %r14
        popq %r13
        popq %r12
        popq %rsi
        popq %rbp
        popq %rbx
        popq %rdi
        popq %gs:0x08
        popq %gs:010
        popq %rcx
        "
              :
              : "r"(current), "r"(to)
             
              )
```
And `makecontext` looks like this:

```crystal
 {% if flag?(:win32) %}
  def makecontext(proc : Proc)
    @f = proc

    s_top = @stack.to_unsafe
    s_bottom = s_top.as(UInt8*) + STACK_SIZE # stack grows down -> high address is bottom
    s_bottom = Pointer(UInt8).new(s_bottom.address & ~15)  # align stack on 16 bytes
    
    s_start = s_bottom - 32
    
    # 8 registers + 2 qwords for NT_TIB + 1 parameter + 10 128bit XMM registers
    @rsp = (s_start - (11*8 + 10*16)).as(Void*) 
    s_start_ptr = s_start.as(UInt64*)
    s_start_ptr.value = ->(f : FiberPoc) { f.run }.pointer.address

    # https://en.wikipedia.org/wiki/Win32_Thread_Information_Block
    # stack end = "stack limit" = low address
    stack_limit = (s_start - 24).as(UInt64*) # where to put the value on our stack
    stack_limit.value = s_top.address

    # stack top = "stack base" = high address
    stack_base = (s_start - 16).as(UInt64*) # where to put the value on our stack
    stack_base.value = s_bottom.address

    # First parameter
    first_param = (s_start - 8).as(UInt64*)
    first_param.value = self.as(Void*).address
  end
  {% else %}
```

The excessive comments is just for explanation. One thing to note is that I use the `movups` instruction instead of the `movaps` instruction you'll see in the Boost library. The reason is that we don't need to worry about aligning the instructions to a 16 byte boundary with `movups` and they have been equally fast for the last 5-6 generations of CPU's (`movaps` can be a bit faster on old CPU's).

If you take a look at the [Boost context switch](https://gist.github.com/cfsamson/3f3c694f7e235226f02188ee2f284c37) there is some more information they store/restore but I don't think it's strictly needed but worth making a note of.

straight-shoota:
I've rebased this PR on current master and also took another stab on the context swap implementation.

The major problem is actually just #7955. The context swap implementation in the `feature/fiber-win32` branch has really been working as it was supposed to all the time. That explains my previous frustration with this 😆 

Thank you so much @cfsamson for investigating and finding the culprit. Your book is really interesting. I'm really glad there is finally help from someone with expertise on windows 😍 

Now I've updated the [feature/fiber-win32](https://github.com/straight-shoota/crystal/tree/feature/fiber-win32) branch. Context switches are basically working now =) I also added the XMM and TIB registers as suggested by @cfsamson 

When running the concurrent_spec, there is a failure trying to resume a non-resumable main fiber. This might be related to channels. I'm not sure right now.

But this is already a big step ahead 🎉 

I'd suggest to continue with reviews on this PR and after it is merged, I'll open a new one adding the win32 implementation.

src/crystal/scheduler.cr
```diff
@@ -118,11 +136,25 @@ class Crystal::Scheduler
...
+      {% raise "unsupported platform" %}
+    {% end %}
   end
 
   protected def yield(fiber : Fiber) : Nil
-    @current.resume_event.add(0.seconds)
+    {% if flag?(:unix) %}
+      @current.resume_event.add(0.seconds)
+    {% elsif flag?(:win32) %}
+      @runnables << fiber
```
ysbaddaden: You can't push the current fiber into the runnables queue; it will lead to a situation where the current fiber will try to resume itself, for example.

You could, however, implement pseudos `Crystal::EventLoop` and `Crystal::Event` classes for win32 with a mere `Deque(Fiber)` and a Fiber that will pop & resume the next fiber from it. That would also avoid the unix/win32 checks everywhere (less noise), and have the API ready for implementing the event loop on win32.

src/kernel.cr
ysbaddaden: `sleep` can't be implemented on win32 without an event loop and timers; i.e. chicken-egg.

src/crystal/scheduler.cr
straight-shoota: I suppose it would probably be better to remove these specifics from this PR.

straight-shoota:
I've reduced the scope of this PR to refactoring and extracting the platform-specific implementations.
Implementations for win32 will be added in a followup. This should make it easier to review and merge.

Meanwhile, I've added an `EventLoop` stub implenentation in the [`feature/fiber-win32`](https://github.com/straight-shoota/crystal/tree/feature/fiber-win32) branch. `concurrent_spec` passes and channels work in general, but the channel spec seems to hang.

ysbaddaden:
Please close this PR or rebase it to match your last branch with the wIn32 context switch! It's much better than the preliminary work here, and I only have a few remarks. It won't delay the merge, and we'll have fibers working on win32! Huge feat!

In fact what I'm not really fond of are these PR commits. I'd prefer thread/pthread.cr, thread/win32.cr, thread/pthread/mutex.cr files, each fully implementing their "public" interface, instead of extracting generic Thread and Thread::Mutex classes. That would add duplication, but readability should improve.

asterite:
It's also the case that refactors like this will probably conflict much more against the current work on parallelism and so it's unlikely that they will get merged, or they will delay parallelism more.

straight-shoota:
@ysbaddaden 
I'm not sure if the entire thing will be easier to review and merge. But sure, I can make a PR for the entire feature branch.

> In fact what I'm not really fond of are these PR commits. I'd prefer thread/pthread.cr, thread/win32.cr, thread/pthread/mutex.cr files, each fully implementing their "public" interface, instead of extracting generic Thread and Thread::Mutex classes.

I'm not sure what you mean. There are no generic classes of either of them. [`src/thread.cr`](https://github.com/crystal-lang/crystal/blob/c184ab1281b5f40520cfb8d8b0dd055a8dd58466/src/thread.cr) has some platform-independent stuff and the specific implementations just reopen this class.

@asterite
If we want concurrency support for win32 we need these refactors.

ysbaddaden:
This PR mostly brings stubs without the win32 implementation of Fiber stacks and x86_64 context switch, those are huge progresses! That helps counterbalance the changes and conflicts with other branches (MT) that this PR introduces.

About the rework: I meant that POSIX and Win32 don't need to share a base Thread class. The amount of duplicated code isn't large enough to justify the split. That would allow them to have their own internal details, so we treat them as "opaque classes".

I'd also like this rework to settle on where we should put specific classes, namely `Thread`, `Thread::Mutex`, `Thread::ConditionVariable` (pthread, win32), `EventLoop` (libevent, win32, then probably epoll, kqueue, ...) to avoid further conflicts.

Being system specific, I suppose they should be moved under `src/crystal/system`?

For threads, I guess they should be `unix/pthread.cr`, `unix/pthread_mutex.cr`, `win32/thread.cr`, and so on, then have `src/thread.cr`, `src/thread/mutex.cr` require the correct one?

For the event loop, maybe `unix/libevent.cr` and `win32/iocp.cr` then have `src/crystal/event_loop.cr` require the correct one?

Just a question: would it be hard to bring real win32 thread support, instead of bringing a stub? I know that's not needed to implement fibers, but threads aren't needed on POSIX either, and it would be awesome to have a basic support for them.

cfsamson:
May I ask how you plan to implement IOCP support? And where the epoll/kqueue/iocp based event loop plays into this? 

Maybe I'm wrong but creating a new thread itself looks pretty easy using the [CreateThread ](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createthread) WinApi function, but I don't have the answer to what setup is required beyond that to fit it into Crystals thread support.

ysbaddaden:
@cfsamson we use libevent for now, which leverages epoll/kqueue for us. I'd like to use on Win32 too, but it may be more complex to integrate libevent than using IOCP directly on Win32.

straight-shoota:
@ysbaddaden I've opened #7995 which has everything in `feature/fiber-win32` included. I'll answer you there.

I think this PR can be closed.

cfsamson:
@ysbaddaden OK, thanks. Seems that libevent abstracts over IOCP also pretty effectively when using `bufferevents`. I guess it's better to discuss this later.

RX14:
I don't want to rely on libevent for windows. It will not work. We must implement our own event loop from scratch on windows for any kind of sane implementation.

PR #6962, opened 2018-10-19
Rewrite macro spec without executing a shell command

### LINKED ISSUE #6961: Compiler spec failing on travis-ci because of insufficient memory
I noticed a lot of CI builds failing on travis-ci because they can't allocate enough memory at some point.
 
* https://travis-ci.org/crystal-lang/crystal/jobs/443615214
* https://travis-ci.org/crystal-lang/crystal/jobs/442935761
* https://travis-ci.org/crystal-lang/crystal/jobs/441646668
* https://travis-ci.org/crystal-lang/crystal/jobs/443615214

The failure is always `fork: Cannot allocate memory (Errno)` in spec `Semantic: macro allows declaring class with macro expression` (https://github.com/crystal-lang/crystal/blob/master/spec/compiler/semantic/macro_spec.cr#L569).

This really doesn't seem to make any sense. It doesn't always happen, but it comes up from time to time. 

According to [travis docs](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) the machines running with Ubuntu Trusty on GCE have 7.5 GB of memory available. That should be plenty for compiling the std and compiler specs - even in one go.

asterite:
Maybe fixes #6961

Let's see if CI fails like this. I'll try to re-run it a few times.

This is probably a workaround and we should investigate why it fails for this specific spec.

### DIFF
--- a/spec/compiler/semantic/macro_spec.cr
+++ b/spec/compiler/semantic/macro_spec.cr
@@ -568,7 +568,7 @@ describe "Semantic: macro" do
 
   it "allows declaring class with macro expression" do
     assert_type(%(
-      {{ `echo "class Foo; end"` }}
+      {{ "class Foo; end".id }}
 
       Foo.new
       )) { types["Foo"] }


bcardiff:
Test passed. Let's do some rebuilds and see how it goes.

asterite:
@bcardiff We should also try [this](https://github.com/crystal-lang/crystal/issues/6961#issuecomment-431358737)

bew:
This can be merged as is, the intent "Rewrite macro spec without executing a shell command" is good, green if it didn't fix some other errors

PR #6963, opened 2018-10-19
Fix typos

r00ster91:
This fixes some typos.

### DIFF
--- a/src/io/stapled.cr
+++ b/src/io/stapled.cr
@@ -15,10 +15,10 @@ class IO::Stapled < IO
 
   # Returns `true` if this `IO` is closed.
   #
-  # Underlying ÌO`s might have a different status.
+  # Underlying `IO`s might have a different status.
   getter? closed : Bool = false
 
-  # Creates a new `IO::Stapled` which reads from *reader*  and writes to *writer*-
+  # Creates a new `IO::Stapled` which reads from *reader* and writes to *writer*.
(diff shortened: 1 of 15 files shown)


src/io/stapled.cr
```diff
@@ -73,7 +73,7 @@ class IO::Stapled < IO
     @writer.write(slice)
   end
 
-  # `Flushes `writer`.
+  # Flushes `writer`.
```
RX14: `*writer*`

Sija: That's incorrect. * are used to denote arguments, not properties.

r00ster91: But ` doesn't seem correct to me either. * seems more correct to me in this case because it's more related to a variable/parameter.

And currently here: [`stapled.cr:21`](https://github.com/crystal-lang/crystal/blob/master/src/io/stapled.cr#L21)
```cr
# Creates a new `IO::Stapled` which reads from *reader*  and writes to *writer*-
def initialize(@reader : IO, @writer : IO, @sync_close : Bool = false)
```
there's also * used to italicize the parameter. And now here using ` would be a bit incosistent I think.

Sija: Yes, because here they're arguments, not properties, that's the whole point.

RX14: My bad, this is correct as-is.

PR #6964, opened 2018-10-19
Auto increment patch version number on builds after a release

bcardiff:
In SemVer 
* `1.10.4+40` == `1.10.4`
* `1.10.4-rc1` < `1.10.4` since these prefix are used before the `1.10.4` is tagged.

Right now the latest tag in crystal is `0.26.1`. And nightly packages are annotated as `0.26.1+99` where 99 is the number of commits from the tag.

The current behavior (before the PR) make it hard to prepare libraries for upcoming breaking changes since there is no nice way to use `compare_versions(Crystal::VERSION, "0.26.1") <= 0`. Up to know there were some mostly `Crystal::VERSION.starts_with?("0.26.1+")` as a hack every now and then.

This PR chances the sufix from `+99` to `-c99` (c as in commits 🤷‍♂️ ) and also increment the patch number from the latest tag. So nightly packages currently could be branded as `0.26.2-c99` which matches the top description of how semver are used. Note: this does not implies that prefix (like `0.26.2`) will exists.

So after this PR libraries will be able to use `compare_versions(Crystal::VERSION, "0.26.1") <= 0` as differentiate nightly from latest.

### DIFF
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -38,19 +38,35 @@ module Crystal
 
       # Failed git and no explicit version set: ""
       # We inherit the version of the compiler building us for now.
-      return { {{Crystal::VERSION}}, nil } if git_version.empty?
+      return {next_patch({{Crystal::VERSION}}), nil} if git_version.empty?
 
       # Shallow clone with no tag in reach: abcd123
       # We assume being compiled with the latest released compiler
-      return {"#{{{Crystal::VERSION}}}+?", git_version} unless git_version.includes? '-'
+      return {"#{next_patch({{Crystal::VERSION}})}-c?", git_version} unless git_version.includes? '-'
 
       # On release: 0.0.0-0-gabcd123
       # Ahead of last release: 0.0.0-42-gabcd123
       tag, commits, sha = git_version.split('-')
-      sha = sha[1..-1]                                # Strip g
-      tag = "#{tag}+#{commits}" unless commits == "0" # Reappend commits since release unless we hit it exactly
+      sha = sha[1..-1] # Strip g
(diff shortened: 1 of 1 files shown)


asterite:
Another alternative is just forget about automating the version number. It was a cool idea but it's brining a lot of problems.

Why not hardcode the next compiler version in the source code? This is what Ruby does, for example. They even have [a bot that updates the date every day](https://github.com/ruby/ruby/commit/1aa8945d4dc402dc9be56f699040293381042e07). I'm not saying we should do the same, but maybe just change those values before a release and that's all.

bcardiff:
We can go slightly that way. a `VERSION` file could be in the root of the repo. We can stop using git at all. Update the file when the Changelog.md is merged. And in the following commit to the next patch version and a suffix like `-edge` or something.

That will remove `CRYSAL_CONFIG_VERSION` also.

I stumble here while trying to deal with the tempfile breaking changes that affect minitest that is used is shards.

straight-shoota:
The existing automatic version determination is already confusing and the behaviour from this PR makes it only more so.

Incrementing the patch version isn't even the correct approach according to semver. When the next release contains changes to the external API, it must be a new minor or major version and even pre-release / development versions should reflect that.

Storing the compiler version in the repository is exactly what I've proposed in #6715. A suffix such as `-edge` or `-dev` for development versions is great.

oprypin:
`{{ compare_versions(Crystal::VERSION, "0.26.99") > 0 }}` is a solution that currently works. May be worth allowing this too:
`{{ compare_versions(Crystal::VERSION, "0.26") > 0 }}` (currently an error "Not a semantic version")

oprypin:
Nevermind, what I said does not apply to nightlies in any way.

PR #4687, opened 2017-07-07
Get version number from CHANGELOG.md when cannot get it from git

MakeNowJust:
Fixed #4642

PR #6966, opened 2018-10-20
Add VERSION file

### LINKED ISSUE #4642: crystal 0.23 reports old version number
I installed crystal 0.23 from source. I've the following requirements:

- pre-compiled crystal-0.21.1-1-linux_x86-64
- gc 7.6
- llvm 3.9.1

In the source dir, I ran command 'make' and 'make spec'. All passed. But it seems crystal 0.23 is reporting 0.21.1 for me. See below.

$ ./crystal-0.21.1-1/bin/crystal -v
Crystal 0.21.1 [3c6c75e] (2017-03-06) LLVM 3.5.0

$ ./crystal-0.23.0-1/bin/crystal -v  
Crystal 0.21.1 (2017-06-29) LLVM 3.9.1

### LINKED ISSUE #6715: Compiler version detection

bcardiff:
So as an alternative to #6964. This PR moves to a plain `VERSION` file that is embedded at compile time. If git information is available the sha is present as before, but is not part of version number.

The version could have `0.26.2-dev` the only important thing is to be greater than latest. But since we know the next version is `0.27.0` we can use `0.27.0` (nb: `0.27.0-dev` < `0.27.0`)

Closes #6964. Fixes #6715. Fixes #4642. Closes #4687. 

---

Update:

Also a `CRYSTAL_CONFIG_BUILD_COMMIT` env var is used to specify the build commit. The Makefile grab this information from git if available.

Closes #6748.


src/compiler/crystal/config.cr
```diff
@@ -30,27 +30,15 @@ module Crystal
     end
 
     private def self.compute_version_and_sha
-      # Set explicitly: 0.0.0, ci, HEAD, whatever
-      config_version = {{env("CRYSTAL_CONFIG_VERSION")}}
-      return {config_version, nil} if config_version
+      config_version = {{ `cat #{__DIR__}/../../../VERSION`.stringify.chomp }}
```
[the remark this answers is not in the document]
bcardiff: Even if a macro is added, we can't use it until the next release. Until then an `ifdef`/`type` would be needed.

bcardiff:
It’s a nice asset to have the version in a plain file i think. Easier to change (eventually by a bot if you want ;-) )

straight-shoota:
I'd also like a way to specify the commit SHA explicitly, either in VERSION or an env var. Development builds shouldn't depend on using git on the build system. That's relevant for example when package managers provide edge builds.

src/compiler/crystal/config.cr
```diff
@@ -30,27 +30,19 @@ module Crystal
...
-      sha = sha[1..-1]                                # Strip g
-      tag = "#{tag}+#{commits}" unless commits == "0" # Reappend commits since release unless we hit it exactly
-
-      {tag, sha}
+      config_version = {% if flag?(:windows) %}
+                       {{ `type #{__DIR__}/../../../VERSION`.stringify.chomp }}
+                       {% else %}
+                       {{ `cat #{__DIR__}/../../../VERSION`.stringify.chomp }}
+                       {% end %}
+      git_describe = {{ `(git describe --tags --long --always 2>/dev/null) || true`.stringify.chomp }}
```
straight-shoota: `git rev-parse --short HEAD 2> /dev/null` returns the sha directly. No need to parse the describe string.

RX14:
perhaps it would also be a good idea to add a CI check to fail the build if we're building a tag and VERSION isn't the same as the tag name

oprypin:
Well, at that point it's too late anyway... Need some kind of preliminary check

RX14:
@oprypin it's not too late to release another version bump at that point. If you're following the release checklist, it will "never happen", but I'd rather have more layers of safety than are required.

straight-shoota:
If you run the specs locally before pushing the release, it's not even to late.

src/compiler/crystal/config.cr
```diff
@@ -30,27 +30,16 @@ module Crystal
...
-      return {"#{{{Crystal::VERSION}}}+?", git_version} unless git_version.includes? '-'
-
-      # On release: 0.0.0-0-gabcd123
-      # Ahead of last release: 0.0.0-42-gabcd123
-      tag, commits, sha = git_version.split('-')
-      sha = sha[1..-1]                                # Strip g
-      tag = "#{tag}+#{commits}" unless commits == "0" # Reappend commits since release unless we hit it exactly
-
-      {tag, sha}
+      {config_version, sha}
```
straight-shoota: I suppose this could be separated into two methods now. Unless we'd want to consider the option of specifying the sha in the `VERSION` file as well, there is no longer any shared logic between "computing" `version` and `sha`.

bcardiff:
@straight-shoota @RX14 I applied all the suggestions so far.

straight-shoota:
What do you think about specifying the SHA as an environment variable? It should at least be optional (when not building in a git environment). I'd even prefer it as the only means and provide the value from the Makefile/build script. It would be cleaner not to call `git` from the compiler.

bcardiff:
@straight-shoota ok. it will require some changes in the distribution scripts doing this (and I am a bit reluctant usually to do it)
 
A `CRYSTAL_CONFIG_SHA` env variable is used to grab the information after the last commit. If empty, then no sha is shown.

j8r:
`CRYSTAL_CONFIG_SHA` isn't explicit enough - perhaps `CRYSTAL_COMMIT_HASH`

bcardiff:
@j8r  It matches the method name `Crystal::Config.sha`

j8r:
That's fair. The point is the method/variable name don't tell what they are – what is a "Crystal config sha"? The method can be renamed to something from `Crystal::Config.sha` to something like `Crystal::Config.commit_hash`?

bcardiff:
I wanted to avoid changing api as much as possible. 

We can call the method `build_commit` and the param `CRYSTAL_CONFIG_BUILD_COMMIT` in order to align with the naming inside `define_crystal_constants`.

RX14:
I don't mind the naming.

src/compiler/crystal/config.cr
```diff
@@ -5,15 +5,18 @@ module Crystal
     end
 
     def self.version
-      version_and_sha.first
+      {% if flag?(:windows) %}
+      {{ `type #{__DIR__}/../../../VERSION`.stringify.chomp }}
```
straight-shoota: Indent this line?
Ditto below.

PR #2791, opened 2016-06-10
Add `read_file` macro method

asterite:
Check the diff for the docs.

The rationale for introducing this is that:
1. Embedding resources has already popped up several times, either here in github or talking with others personally.
2. One can currently use `cat filename.txt` but that's not portable, and one has to redirect errors to avoid getting a compilation error.
3. `read_file` returns `nil` if the file doesn't exist. With that, one can choose to deal with `nil` at compile time or runtime.
4. It should be a bit faster because an external process doesn't need to be started

~~This works for text files, but one can also use it for binary data: simply call `to_slice` afterwards to get the bytes (though we probably need a literal for binary data, but that's another issue)~~

Actually, it doesn't work for binary files because the string won't be able to be embedded at all. We could probably add a `read_binary_file` later, if we really want to do that.

Fixes #1649 and #2751

PR #6967, opened 2018-10-20
Add `read_file` macro method

Sija:
#2791, again! ;)

/cc @asterite

### DIFF
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -75,6 +75,8 @@ module Crystal::Macros
   # ```
   # read_file("#{__DIR__}/some_file.txt")
   # ```
+  #
+  # NOTE: Relative paths are resolved to the current working directory.
   def read_file(filename) : StringLiteral | NilLiteral
   end
 
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -67,7 +67,7 @@ module Crystal::Macros
   def raise(message) : NoReturn
   end
 
-  # Reads a file: if it exists, returns a StringLiteral with its contents;
+  # Reads a file: if it exists, returns a `StringLiteral` with its contents;
   # otherwise `nil` is returned.
   #
   # To read a file relative to where the macro is defined, use:
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -67,6 +67,17 @@ module Crystal::Macros
   def raise(message) : NoReturn
   end
 
+  # Reads a file: if it exists, returns a StringLiteral with its contents;
+  # otherwise `nil` is returned.
+  #
+  # To read a file relative to where the macro is defined, use:
+  #
+  # ```
+  # read_file("#{__DIR__}/some_file.txt")
+  # ```
+  def read_file(filename) : StringLiteral | NilLiteral
+  end
+
   # Compiles and execute a Crystal program and returns its output
   # as a `MacroId`.
   #
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -30,6 +30,8 @@ module Crystal
         interpret_system(node)
       when "raise"
         interpret_raise(node)
+      when "read_file"
+        interpret_read_file(node)
       when "run"
         interpret_run(node)
       else
@@ -140,6 +142,20 @@ module Crystal
 
     def interpret_skip_file(node)
       raise SkipMacroException.new(@str.to_s)
+    end
+
+    def interpret_read_file(node)
+      if node.args.size == 1
+        node.args[0].accept self
+        filename = @last.to_macro_id
+        if File.exists?(filename)
+          @last = StringLiteral.new(File.read(filename))
+        else
+          @last = NilLiteral.new
+        end
+      else
+        node.wrong_number_of_arguments "macro call 'read_file'", node.args.size, 1
+      end
     end
 
     def interpret_system(node)
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -182,8 +182,7 @@ module Crystal
 
       node.args[0].accept self
       filename = @last.to_macro_id
-      filename = find_file(filename) { }
-      if filename
+      if File.exists?(filename)
         @last = StringLiteral.new(File.read(filename))
       else
         @last = NilLiteral.new
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -144,20 +144,6 @@ module Crystal
       raise SkipMacroException.new(@str.to_s)
     end
 
-    def interpret_read_file(node)
-      unless node.args.size == 1
-        node.wrong_number_of_arguments "macro call 'read_file'", node.args.size, 1
-      end
-
-      node.args[0].accept self
-      filename = @last.to_macro_id
-      if File.file?(filename)
-        @last = StringLiteral.new(File.read(filename))
-      else
-        @last = NilLiteral.new
-      end
-    end
-
     def interpret_system(node)
       cmd = node.args.map do |arg|
         arg.accept self
@@ -177,6 +163,20 @@ module Crystal
 
     def interpret_raise(node)
       macro_raise(node, node.args, self)
+    end
+
+    def interpret_read_file(node)
+      unless node.args.size == 1
+        node.wrong_number_of_arguments "macro call 'read_file'", node.args.size, 1
+      end
+
+      node.args[0].accept self
+      filename = @last.to_macro_id
+      if File.file?(filename)
+        @last = StringLiteral.new(File.read(filename))
+      else
+        @last = NilLiteral.new
+      end
     end
 
     def interpret_run(node)
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -4,39 +4,6 @@ require "semantic_version"
 
 module Crystal
   class MacroInterpreter
-    private def find_file(filename)
-      filename = "#{filename}.cr" unless filename.ends_with?(".cr")
-
-      # Support absolute paths
-      if filename.starts_with?('/')
-        if File.exists?(filename)
-          unless File.file?(filename)
-            return yield "#{filename.inspect} is not a file"
-          end
-        else
-          return yield "can't find file #{filename.inspect}"
-        end
-      else
-        begin
-          relative_to = @location.try &.original_filename
-          found_filenames = @program.find_in_path(filename, relative_to)
-        rescue ex
-          return yield ex.message
-        end
-
-        unless found_filenames
-          return yield "can't find file #{filename.inspect}"
-        end
-
-        if found_filenames.size > 1
-          return yield "#{filename.inspect} is a directory"
-        end
-
-        filename = found_filenames.first
-      end
-      filename
-    end
-
     def interpret_top_level_call(node)
       interpret_top_level_call?(node) ||
         node.raise("undefined macro method: '#{node.name}'")
@@ -218,11 +185,37 @@ module Crystal
       end
 
       node.args.first.accept self
-      original_filename = @last.to_macro_id
+      filename = @last.to_macro_id
+      original_filename = filename
 
-      filename = original_filename
-      filename = find_file(filename) do |error_message|
-        node.raise "error executing macro 'run': #{error_message}"
+      # Support absolute paths
+      if filename.starts_with?("/")
+        filename = "#{filename}.cr" unless filename.ends_with?(".cr")
+
+        if File.exists?(filename)
+          unless File.file?(filename)
+            node.raise "error executing macro run: '#{filename}' is not a file"
+          end
+        else
+          node.raise "error executing macro run: can't find file '#{filename}'"
+        end
+      else
+        begin
+          relative_to = @location.try &.original_filename
+          found_filenames = @program.find_in_path(filename, relative_to)
+        rescue ex
+          node.raise "error executing macro run: #{ex.message}"
+        end
+
+        unless found_filenames
+          node.raise "error executing macro run: can't find file '#{filename}'"
+        end
+
+        if found_filenames.size > 1
+          node.raise "error executing macro run: '#{filename}' is a directory"
+        end
+
+        filename = found_filenames.first
       end
 
       run_args = [] of String
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -4,6 +4,37 @@ require "semantic_version"
 
 module Crystal
   class MacroInterpreter
+    private def find_file(filename)
+      # Support absolute paths
+      if filename.starts_with?('/')
+        if File.exists?(filename)
+          unless File.file?(filename)
+            return yield "#{filename.inspect} is not a file"
+          end
+        else
+          return yield "can't find file #{filename.inspect}"
+        end
+      else
+        begin
+          relative_to = @location.try &.original_filename
+          found_filenames = @program.find_in_path(filename, relative_to)
+        rescue ex
+          return yield ex.message
+        end
+
+        unless found_filenames
+          return yield "can't find file #{filename.inspect}"
+        end
+
+        if found_filenames.size > 1
+          return yield "#{filename.inspect} is a directory"
+        end
+
+        filename = found_filenames.first
+      end
+      filename
+    end
+
     def interpret_top_level_call(node)
       interpret_top_level_call?(node) ||
         node.raise("undefined macro method: '#{node.name}'")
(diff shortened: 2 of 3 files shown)


bcardiff:
I don't think the discussion will differ. Better to find a way to embed slices.

That been said. I care more about API. So, if the macro would return a slice, even if the string is embedded as a first implementation, that is something I could live with. Until a better implementation without breaking the API will come.

sdogruyol:
I've wanted to have this for a long time. Let's KISS and start with something easy to iterate on like this :+1:

sdogruyol:
I've wanted to have this for a long time. Let's KISS and start with something easy to iterate on like this :+1:

sdogruyol:
I've wanted to have this for a long time. Let's KISS and start with something easy to iterate on like this :+1:

src/compiler/crystal/macros/methods.cr
```diff
@@ -142,6 +144,20 @@ module Crystal
       raise SkipMacroException.new(@str.to_s)
     end
 
+    def interpret_read_file(node)
+      if node.args.size == 1
+        node.args[0].accept self
+        filename = @last.to_macro_id
+        if File.exists?(filename)
```
asterite: I think the full filename's logic should be enhanced. Check how the `run` macro method is implemented.

asterite:
Right now a string can contain any kind of bytes, so I think this is fine. If someone wants to embed a slice of bytes they can just do:

```crystal
DATA = {{ read_file("whatever") }}.to_slice
```

ysbaddaden:
I'd like a kwarg, to have a String (default) or Bytes (on demand), but maybe it would be confusing, and always having a String is better?
```crystal
{{ read_file("LICENSE.md") }}             # => String
{{ read_file("logo.png", binary: true) }} # => Bytes
```

jhass:
I don't see a strong reason to have it for string, worst case it should be a `String.new({{read_file...)` away, but if we want it I'd vote for having `read_file` to return `Bytes` and `read_file_to_string` or perhaps `read_text_file` to return `String`.

asterite:
The problem is that `read_file` is a macro method, and, like all macro methods, it returns an AST node. There's a `StringLiteral` node to represent a string. There's **no** node to represent a slice of bytes. So if you want that you'll have to:
- introduce such node
- define a syntax to write it (maybe `Bytes[...]` would be enough)
- provide operations on it

In my mind it's just simpler to make `read_file` return a string. I don't see any problem with it, and you can manipulate the data at compile-time just fine.

jhass:
The proper implementation should do something like https://github.com/crystal-lang/crystal/pull/2791#issuecomment-225299043, so generate an AST subtree equivalent to the inline assembly call. 

As Brian points out, introducing it now to return a string literal would just force a breaking  API change in the future.

On a compromise, maybe we can do a text only version for now and call it `read_text_file` or the like?

asterite:
If it returns an `asm` node there's no way to manipulate the data at compile-time. I still don't understand what's the problem with having it be a StringLiteral, could someone explain this to me? :-)

asterite:
But `read_file` returning `StringLiteral` is perfect and there won't be a need to change it's API. I mean, we are assuming it needs to be changed because it has something wrong with it. I want to know what that is.

jhass:
Maybe it's the old discussion about what `String` should be, generic data blob or representation of sequence of codepoints in some encoding and me strongly favoring the latter as I see doing the former one of the biggest mistakes of Ruby. Also if we would go for the former we should remove `Slice`.

jhass:
Also if we keep the concerns separate, that is read some text vs just give me the bytes, we can make the former pleasant to use at some point by applying any necessary encoding transformations.

asterite:
Then I guess a conclusion won't be reached here. I propose to close this then until someone figures out how to lay out everything related to strings and bytes.

straight-shoota:
@jhass `String.new({{ read_file }})` would copy the slice to the new String instance.

Having a `binary` argument as @ysbaddaden suggested seems like a good idea. If `false`, it returns a `StringLiteral`. If `binary: true`, it could just append a call to `#to_slice` to the `StringLiteral`. And later, when there is a proper bytes data type, it could just be upgraded to that without any changes in the API.

bcardiff:
* I'm ok with `read_text_file`.
* I'm ok with `read_file` to expand to a `Call(StringLiteral, :to_slice)` and whatever is needed to simulate a slice literal.
* I'm ok with not having `read_file` now, but adding `read_text_file` now
* I'm not ok with changing the returned type based on an argument. Even in macros. So no for `binary: true`.

So, if the macro name is changed and the review of @asterite regarding path is solved, this is GTG on my side.

RX14:
There's clearly two different things wanted here:

- a way to read files inside a macro and then use the contents of the file inside the macro to make decisions on the macro level. This *may* expose the contents of a file at runtime but doesn't have to. The most reasonable way to do this is returning a `StringLiteral`. Think `VERSION` file.
- a way to efficiently embed *large*, possibly binary, files inside the executable to make exevutables more self-contained. Think compiled image and JS assets for a webapp.

These are orthogonal and want different interfaces. The former can be `read_file` and would be a **macro method**. The latter can be achieved using module-level ASM to embed a file in the ELF directly using assembler directives. I'd call it `embed_file` and it'd be a **macro** (and macros are not macro methods)

jhass:
I still think we should be cautious with introducing the first case while the second doesn't exist, because during that period people will abuse it for the second case. So I stick to preferring `read_text_file` as a name, I don't think it harms the first case in any way and communicates its contract much more clearly.

straight-shoota:
@RX14 The embedded data could be binary but also text. For text, it should simple to get a `String` 
at runtime without copying the slice.

RX14:
@straight-shoota you could just use `read_text_file` for that.

RX14:
> Yeah, but it feels a bit weird to have `read_text_file` but `embed_file` for the same thing with a binary file.

But it's not the same thing, conceptually or in implementation. That was the whole point of my comment above.

straight-shoota:
Both fits in the second bullet point. If I want to embed a file into the binary, I can use `embed_file` for binary data. But if it's text data and it's supposed to be accessible as `String` at runtime, I need to use `read_text_file` instead.

asterite:
Any other language that lets you **embed** data using `asm`? It seems like a huge portability pain to do that.

RX14:
@asterite any language with `asm` which uses a real assembler lets you do it.

Doing file embedding with `global_asm` *just* means that we pass all the work over to LLVM, and don't have to add any code in the crystal compiler to do this. The asm isn't platform-specific because the ASM block consists only of assembler directives, no instructions.

The implementation doesn't matter anyway.

RX14:
@straight-shoota it'd be easy to add slice and string versions of `embed_file`.

`embed_file` isn't what this PR is about.

straight-shoota:
> `embed_file` isn't what this PR is about.

Agreed, but we're also taking future development into consideration.

And as far as I can tell, a string version of `embed_file` would be equivalent to `read_text_file`, wouldn't it?

asterite:
@straight-shoota No. With `embed_file` you won't have access to the contents of the file at compile-time.

straight-shoota:
Exactly. But what would then be the point of having that over `read_text_file`? It can be equally used to embed a string to be accessible at runtime *and* also gives access at compile time.

RX14:
@straight-shoota it's about compile-time performance. For large files, `embed_file` means that the string doesn't have to be read from the file into memory, escaped as a crystal string, passed back through the parser to unescape the string, and then stored in memory again and written into the LLVM module.

Using assembler directives means that the assembler can copy the input file directly into the output object file with no overhead. This matters with large files. However, large files probably won't be strings.

straight-shoota:
Yeah, that's true. I guess it all comes down to properly communicating the individual properties.

bcardiff:
So, poll time?

👍 adding `read_text_file` after applying the outstanding review + rename https://github.com/crystal-lang/crystal/pull/6967#issuecomment-431869656. 

👎 for closing this PR until the `embed_file` is implemented in a way that supports https://github.com/crystal-lang/crystal/pull/6967#issuecomment-433476252.

RX14:
I've always been +1 for this PR, just waiting for the fixes to be applied.

straight-shoota:
I'm not arguing against this PR :+1: I'm just emphasizing to think about the overall design of these macros.

Sija:
Remind me please why won't we like to have `read_file` (returning `String`) + `embed_file` (returning `Bytes`)?

straight-shoota:
@Sija We're talking about macros, so `read_text_file` returns a `StringLiteral`. But `embed_file` wouldn't return a `BytesLiteral` (which doesn't exist) but directly embed the data at codegen and doesn't need to keep it in memory.

RX14:
I'm no longer OK with `read_text_file`. It should just be `read_file`. It's a macro method, so of course it's going to return a string literal. And we'll probably never add a `read_binary_file`.

Sija:
@straight-shoota Yes I know, I meant literal version of those (acknowledging that there's no `SliceLiteral` atm).

bew:
What about we add a `BytesLiteral` that will currently be the same as `StringLiteral`, but will prepare the terrain for when we have a slice literal?

This way `read_text_file` could return a `StringLiteral`, a future `read_binary_file` could return a `BytesLiteral`, and `embed_file` could embed the file.

RX14:
We don't need a slice literal. You only need a string in a macro since you can only *do something* with a string in a macro. If you need to do advanced binary file reading you can use macro run file to process a binary file and emit any code (including possibly an `embed_file` call to a file generated on-the-fly) you need.

Just apply the suggestions and merge this PR.

RX14:
Or maybe it's obvious this feature is too confusing from the replies to this issue and we should just leave it out entirely.

Sija:
Applied suggestions from https://github.com/crystal-lang/crystal/pull/6967#discussion_r226964391, ready for another round of review.

jhass:
There's a couple of you who voted on both the comments arguing for naming it `read_file` and naming it `read_text_file`. Let's have another final round of voting. React here with:

*  🎉   for `read_file`
* ❤️  for `read_text_file`

src/compiler/crystal/macros.cr
```diff
@@ -67,6 +67,17 @@ module Crystal::Macros
   def raise(message) : NoReturn
   end
 
+  # Reads a file: if it exists, returns a StringLiteral with its contents;
```
straight-shoota: `` `StringLiteral` ``

src/compiler/crystal/macros.cr
```diff
@@ -67,6 +67,17 @@ module Crystal::Macros
...
+  # otherwise `nil` is returned.
+  #
+  # To read a file relative to where the macro is defined, use:
```
straight-shoota: Maybe add that paths are relative to the current working directory. That's already implied, but can't hurt to make it explicit.

src/compiler/crystal/macros/methods.cr
```diff
@@ -163,44 +180,50 @@ module Crystal
...
+      if node.args.size == 0
+        node.wrong_number_of_arguments "macro call 'run'", 0, "1+"
+      end
+
+      node.args.first.accept self
+      original_filename = @last.to_macro_id
+
+      filename = original_filename
+      filename = "#{filename}.cr" unless filename.ends_with?(".cr")
+      filename = find_file(filename) { |error_message|
```
straight-shoota: Could you please change this to `do ... end`? Multiline blocks with braces looks really odd.

src/compiler/crystal/macros/methods.cr
Sija: Done and ready for re-review :)

Sija:
@straight-shoota I've applied your comments, thanks!
@asterite wanna chime in, is it GTG?

src/compiler/crystal/macros.cr
```diff
@@ -67,6 +67,19 @@ module Crystal::Macros
   def raise(message) : NoReturn
   end
 
+  # Reads a file: if it exists, returns a `StringLiteral` with its contents;
+  # otherwise `nil` is returned.
+  #
+  # To read a file relative to where the macro is defined, use:
+  #
+  # ```
+  # read_file("#{__DIR__}/some_file.txt")
```
asterite: I think `./some_file.txt` should work, can you try it out? The idea is that this works like require.

straight-shoota: Hm, it probably works. But I don't think it should. Relative paths should be relative to the current working directory.
Either way, there should be a spec for reading a relative path.

straight-shoota:
I don't see any reason why a relative path should be relative to the file where the macro is called from. This doesn't make any sense. And it gets super confusing when the actual call to `read_file` is defined in a different file than the string literal it receives as an argument. This might be the case with a helper macro that relies on `read_file` and receives the file name as argument.

If you need this behaviour, there is `__DIR__` for that.

src/compiler/crystal/macros.cr
asterite: Well, we don't do require `"#{__DIR__}/..."`, we use relative paths. `require` can be thought as a macro call that requires the given file. `read_file` is another macro that works the same way but reads it at compile-time. Because it's compile-time, I think it makes sense to use the same rules as require. In fact, the whole change that @Sija made because of me requesting it is exactly that.

straight-shoota: The argument to `require` is always an (explicit) string literal. There is no reason to make it customizeable. And there is no reason to require a file relative to the working directory.

But to `read_file` you might want to pass a `StringLiteral` as argument that is originally defined in a completely different place. And it there are perfectly fine reasons to read a file relative to the current working directory instead of the source location.

asterite:
Please don't depend on my approval or advice to move forward with this PR.

Sija:
I'm inclined to agree with @straight-shoota on this, sharing path resolution method between `require` and macros - both, `run` and `read_file` seems counter-intuitive and too magical considering the logic inside: https://github.com/crystal-lang/crystal/blob/0ca54f8d8d5c063f625f0fc6009949653d6184e1/src/compiler/crystal/crystal_path.cr#L66-L124

I'd suggest removing this feature from both macros, allowing only absolute paths (+ relative using `#{__DIR__}`)

straight-shoota:
Could you please add a spec for relative paths? (i.e. `spec/std/data/build")
Relative paths are not forbidden (and shouldn't). `File.file?` will resolve a relative path in the current working directory.

bcardiff:
@Sija the las commit needs to go. The macro first needs to be ina released version before using it in the compiler.

Sija:
@bcardiff yep, my bad, I removed it.

src/compiler/crystal/macros/methods.cr
```diff
@@ -142,6 +177,20 @@ module Crystal
       raise SkipMacroException.new(@str.to_s)
     end
 
+    def interpret_read_file(node)
```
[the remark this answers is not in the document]
Sija: Indeed, after all I reckon that sharing logic between `require` / `run` and `read_file` is not desirable. Former ones use it to require/run code, latter is used to read generic files, not necessarily code, so having such behaviour would be confusing.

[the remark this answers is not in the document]
Sija: Seems like GTG if that's the last remaining thing :shipit:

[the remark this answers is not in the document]
bcardiff: But then the `find_file` refactor is not required to add the feature. Better to have a different PR for that refactor. Otherwise tracking the change on releases is harder ;-)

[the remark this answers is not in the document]
Sija: @bcardiff Fair 'nuff, I've reverted this.

bcardiff:
NB: We should wait for 1 more core team re review before merging. And it should be merged with a squash.

straight-shoota:
@Sija could squash it and add a nice commit message =)

bcardiff:
Merged :-)
The nice commit message was the title :-P

Thanks everybody for the discussion!

Sija:
It occured to me that perhaps it would be sensible to have nilable `read_file?` (current `read_file`) and raising `read_file` variants, whatcha think?

Sija:
@bcardiff yep, exactly. Although I would (add `read_file?` variant), since I can easily imagine use case where you'd like to read file with certain fallback which wouldn't be possible when using raising variant.

asterite:
Current implementation covers all use cases.

Sija:
@asterite true, yet it simplifies things in case like this:

```cr
{% version = read_file("../../VERSION") %}
{{ version ? version.chomp : nil }}
```

vs

```cr
{{ read_file("../../VERSION").chomp }}
```

straight-shoota:
```cr
{{ read_file("../../VERSION") || raise "Missing VERSION file" }}
```

Sija:
@straight-shoota nope, it ain't work since I'd need to use `.chomp`. Having `#try` would solve it but sadly, it doesn't exist in macros.

straight-shoota:
```cr
{{ (read_file("../../VERSION") || raise "Missing VERSION file").chomp }}
```

woodruffw:
:+1: For a version that errors at compile time if the file doesn't exist (or I/O fails in any way).

Sija:
I agree with @woodruffw, @straight-shoota's workaround is good enough, yet having it "baked" into the method itself feels more natural (as long as there's nilable variant too).

PR #6968, opened 2018-10-20
Remove quotes from overflow symbols in formatter spec

r00ster91:


straight-shoota:
For some reason this keeps repeatedly failing on circleci/test_linux because of either one of #6961 or #7243. I doubt it's related to this PR, but the statistical accumulation is at least suspicious.

PR #6970, opened 2018-10-21
determine thread stack size with pthread API

damaxwell:
Moves thread stack size computation out of Fiber and into Thread.

Switches to pthread APIs for computing stack size.


ysbaddaden:
Thanks a lot!

The only bit I'm wary of, is whether the stack addr/size/guard is reported correctly for the *main thread*, since there isn't a standard and portable `pthread_getattr`. I investigated and wrote my results below. Only **darwin** seems to behave incorrectly for the main thread.

- darwin:  `pthread_get_stackaddr_np` seems to be valid, but `pthread_get_stacksize_np` returns bogus values for the main thread (see https://opensource.apple.com/source/WTF/WTF-7602.1.50/wtf/StackBounds.cpp.auto.html). At least, it used to, and we should be conservative.

- freebsd: [pthread_attr_get_np(3)](https://www.unix.com/man-page/freebsd/3/pthread_attr_get_np/) seems correct (see https://github.com/ivmai/bdwgc/commit/49d7fe77033f0bdabd50a2912cd58effd85452c1).

- linux-gnu: [pthread_getattr_np(3)](http://man7.org/linux/man-pages/man3/pthread_getattr_np.3.html) seems correct (per the manpage).

- linux-musl: `pthread_getattr_np(3)` seems compliant with the GNU extension (see https://git.musl-libc.org/cgit/musl/tree/src/thread/pthread_getattr_np.c).

- openbsd: [pthread_stackseg_np(3)](http://man.openbsd.org/OpenBSD-6.3/pthread_stackseg_np.3) seems to take care of the main thread (see https://github.com/openbsd/src/blob/b66614995ab119f75167daaa7755b34001836821/lib/librthread/rthread_np.c#L78).

src/thread.cr
```diff
@@ -135,4 +135,58 @@ class Thread
...
+  # :nodoc:
+  def stack : Slice(Void)
+    bottom = Pointer(Void).null
+    top = Pointer(Void).null
+    {% if flag?(:darwin) %}
+      bottom = LibC.pthread_get_stackaddr_np(@th);
+      stack_size = LibC.pthread_get_stacksize_np(@th);
+      top = Pointer(Void).new(bottom.address - stack_size)
+    {% elsif flag?(:freebsd) %}
+      err = LibC.pthread_attr_init(out attr)
```
ysbaddaden: - Pedantic: can we harmonize the implementation within thread? I use `ret` (instead of `err`) and inline the check+raise.
- Leak: if any further call to pthread fails, the `pthread_attr_t` object is never destroyed (this applies to linux as well).

For example:
```crystal
ret = LibC.pthread_attr_init(out attr)
raise Errno.new("pthread_attr_init", ret) unless ret == 0

begin
  ret = LibC.pthread_attr_get_np(@th, pointerof(attr))
  raise Errno.new("pthread_attr_get_np", ret) unless ret == 0

  # ...
ensure
  LibC.pthread_attr_destroy(pointerof(attr))
end
```

src/fiber.cr
```diff
@@ -133,6 +118,17 @@ class Fiber
     end
   end
 
+  # :nodoc:
+  def stack : Slice(Void)
+    # If we allocated our own stack, we know its bounds.
+    if @stack != Pointer(Void).null
+      return Slice(Void).new(@stack, @stack_bottom.address - @stack.address)
+    end
+    # Otherwise, assume for now that there is only one thread,
```
ysbaddaden: Maybe use a more detailed explanation:
> the fiber represents a thread's main stack and we assume it's the currently running thread; this should be fine since a thread should never resume the main fiber of another thread (undefined behavior? chaos?)

src/signal.cr
```diff
@@ -324,9 +324,11 @@ fun __crystal_sigfault_handler(sig : LibC::Int, addr : Void*)
   # Determine if the SEGV was inside or 'near' the top of the stack
   # to check for potential stack overflow. 'Near' is a small
   # amount larger than a typical stack frame, 4096 bytes here.
-  stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)
 
-  if stack_top <= addr < Fiber.current.@stack_bottom
+  stack = Fiber.current.stack
+  stack_top = Pointer(Void).new(stack.pointer(0).address - 4096)
```
RX14: you can perform arithmetic on pointers directly.

src/fiber.cr
RX14: Why not pass the thread stack through the `Fiber.new` constructor. Then we simply have a fiber constructor for creating a fiber with a new stack and a fiber constructor for creating a fiber with an existing stack, and `Fiber` never has to worry about `Thread.current.stack` at all.

src/thread.cr
```diff
@@ -135,4 +135,58 @@ class Thread
       Fiber.inactive(fiber)
     end
   end
+
+  # :nodoc:
+  def stack : Slice(Void)
```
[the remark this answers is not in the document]
ysbaddaden: Yes, but we don't have Thread in Crystal::System, yet.

src/signal.cr
RX14: just make it a named local variable, like `page_size = 4096` then `stack.pointer(0) - page_size`. Constants don't need to be constants in the language if they're only used in one function. They just need to be constant.

src/fiber.cr
damaxwell: We can’t cache @stack in Fiber because it is dynamic in time. What is is at the time of Fibre construction need not be what it is when a segfault occurs later in the run. The stack bottom stays put but the stack size adjusts.

src/thread.cr
damaxwell: True enough about that leak.  I had seen the same leak also in #6944 in the changes to `ConditionVariable#initialize` and had assumed that you hadn't used full care to release the resources because if these calls are failing the house is on fire, and maybe the leaky faucet isn't so worrisome. I'll fix this here and submit a separate PR for that other one.

ysbaddaden: Oh, I totally missed it! I only overlooked some possible errnos that are very unlikely (per the manpages) and not worth raising.

src/fiber.cr
ysbaddaden: I did too much low level programming, I guess. It sounded safe enough, but it's better to be really safe.

Is the stack size really changing? We never tweak it, and I'm not sure we ever do.

If so, maybe linking a Fiber to its Thread when it represents a thread stack would be a better idea? We could then have something like `@thread.try(&.stack) || @stack`.

damaxwell: On `x86_64-linux-musl` at least, the stack sizes reported from `pthread_attr_get_np` & `pthread_attr_getstack` on the main thread reflect the current stack allocation, which can be much smaller than the ultimate limit, so it will grow in time. And although the Crystal code base never changes the limits on the main stack, there is nothing to stop a user from calling `setrlimit`, which will change the main stack size.

Saving the associated thread if the Fiber represents a thread's stack (main thread or otherwise), is the right thing to do.

src/thread.cr
damaxwell: For the requested changes:

1. On Darwin, on the main thread, we fall back to `getrlimit`.
2. Switched to `ret` not `err` and tried to match nearby code style (`throw ... unless ret == 0`)
3. This was the biggest change.  For platforms that do not have an equivalent of `pthread_getattr_np` we need to cache the thread attributes at the time of creation to access the guard size.  The stack is returned as a separate `Thread::Stack` structure that maintains a pointer, a size, and an optional guard size, which is `nil` if is unknown (i.e. on the main process thread).  If you think this is pedantic, we can simply report a 0 instead.  My previous use of `Slice(Void)` to represent a stack turned out to be a problem because it uses an `Int32` to represent its count, and (rarely, but in principle) a stack can have more bytes than that.

The leak was addressed as well.  For consistency in the code base, probably `ConditionVariable` also ought to get fixed; it doesn't look like the errno's that could occur there are much less likely than the errno's addressed here.

src/fiber.cr
damaxwell: Well, it turned out that it having the `Thread` know about its `Fiber` and the `Fiber` knowing about its `Thread` didn't mix well with the fact that `Thread` has a finalizer.  Numerous `GC Warning: Finalization cycle` warnings were generated running the spec suite.

So instead, a `Fiber` returns an optional `Thread::Stack` which is `Nil` if the `Fiber` is not responsible for maintaining its stack (i.e. the main `Fiber` for a `Thread`).  A `Thread` always knows about its own stack and can return it.  If you want the current stack, talk to the `Crystal::Scheduler`, which knows both the current `Thread` and the current `Fiber` and can figure out what is the right stack to report.

src/signal.cr
damaxwell: Addressed via a local variable.

@RX14 , my first experience with pointer arithmetic in Crystal was with subtracting a `UInt64`, which is natural to figure out a stack top when you know the bottom and the stack grows down.  But this seems to be broken, and I shied away from it afterwards.  But I think it's just subtracting a `UInt` that's problematic and could be fixed in a separate PR.

src/thread.cr
```diff
@@ -42,12 +51,19 @@ class Thread
   def initialize
     @func = ->{}
     @th = LibC.pthread_self
-    @main_fiber = Fiber.new
 
     @@threads.push(self)
   end
 
+  def is_process_thread?
```
Sija: Drop `is_` prefix please, there's already `?` at the end, which communicates it's a predicate method.

damaxwell: Done.

src/thread.cr
```diff
@@ -135,4 +151,72 @@ class Thread
...
     end
   end
+
+  # :nodoc:
+  def stack : Thread::Stack
+    top = Pointer(Void).null
+    stack_size : LibC::SizeT = 0
+
+    {% if flag?(:darwin) %}
+      bottom = LibC.pthread_get_stackaddr_np(@th);
```
Sija: Redundant `;`

src/thread.cr
```diff
@@ -135,4 +151,72 @@ class Thread
...
+    # glibc `pthread_getattr_np` makes no attempt to determine it:
+    # https://code.woboq.org/userspace/glibc/nptl/pthread_getattr_np.c.html
+    guard_size : UInt64? = nil
+    @attr.try do |attr|
+      ret = LibC.pthread_attr_getguardsize(pointerof(attr), out gs)
+      raise Errno.new("pthread_attr_getguardsize", ret) unless ret == 0
+      guard_size = gs.to_u64
+    end
+
+    return Thread::Stack.new(top, stack_size.to_u64, guard_size)
```
Sija: Redundant `return`

src/thread/stack.cr
```diff
@@ -0,0 +1,16 @@
+# :nodoc:
+class Thread
+  # :nodoc:
+  struct Stack
+    property top : Pointer(Void)  # Lowest address of a writable byte
```
Sija: As a general rule always put the comments one line above the code in question, otherwise they'll be misassigned by the doc generator (even though that's not the case here atm, since it's nodoc-ed).

src/thread.cr
```diff
@@ -25,14 +26,22 @@ class Thread
...
+
+    ret = GC.pthread_create(pointerof(@th), pointerof(attr), ->(data : Void*) {
       (data.as(Thread)).start
       Pointer(Void).null
     }, self.as(Void*))
 
     if ret == 0
       @@threads.push(self)
     else
+      @attr.try do |attr|
```
RX14: We already have non-nillable `attr` (not `@attr`) above.

RX14:
This is basically fine

src/thread/stack.cr
damaxwell: These comments aren't written in a style that's suitable for public docs either, so I'm going to leave the formatting as-is.  `crystal tool format` was happy to line up the comments to start in the same column for me when I had left them ragged, so I imagine this commenting style is not entirely foreign.

src/thread.cr
damaxwell: Ha! Silly me for copy/paste of my code from the finalizer.

damaxwell:
@bcardiff: Rebased.

PR #6971, opened 2018-10-21
Document Top Level Namespace constants

r00ster91:
This documents the Top Level Namespace constants [`STDIN`](https://crystal-lang.org/api/0.26.1/toplevel.html#STDIN), [`STDOUT`](https://crystal-lang.org/api/0.26.1/toplevel.html#STDOUT), [`STDERR`](https://crystal-lang.org/api/0.26.1/toplevel.html#STDERR), [`ARGV`](https://crystal-lang.org/api/0.26.1/toplevel.html#ARGV) and [`ARGF`](https://crystal-lang.org/api/0.26.1/toplevel.html#ARGF).

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -59,7 +59,7 @@ ARGV = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
 # ```
 # # Argument passed to the program: "file"
 # ARGF.gets_to_end # => "123"
-# ARGV # => []
+# ARGV             # => []
 # ```
 #
 # You can manipulate `ARGV` yourself to control what `ARGF` operates on.
@@ -68,7 +68,7 @@ ARGV = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
 # ```
 # ARGV.replace ["file1"]
 # ARGF.gets_to_end # => The content of file1
-# ARGV # => []
+# ARGV             # => []
 # ARGV << "file2"
 # ARGF.gets_to_end # => The content of file2
 # ```
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -16,7 +16,7 @@
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
 
-# Path to the current temporary program.
+# Path to the current program.
 PROGRAM_NAME = String.new(ARGV_UNSAFE.value)
 
 # The arguments passed to the program.
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -2,14 +2,18 @@
   # The standard input file descriptor. Contains data piped to the program.
   STDIN  = IO::FileDescriptor.new(0)
 
-  # The standard output file descriptor. Gets flushed when a newline is written.
+  # The standard output file descriptor.
   #
   # Typically used to output data and information.
+  #
+  # NOTE: Gets flushed when a newline is written.
   STDOUT = IO::FileDescriptor.new(1).tap { |f| f.flush_on_newline = true }
 
-  # The standard error file descriptor. Gets flushed when a newline is written.
+  # The standard error file descriptor.
   #
   # Typically used to output error messages and diagnostics.
+  #
+  # NOTE: Gets flushed when a newline is written.
   STDERR = IO::FileDescriptor.new(2).tap { |f| f.flush_on_newline = true }
 {% else %}
   require "c/unistd"
@@ -17,14 +21,18 @@
   # The standard input file descriptor. Contains data piped to the program.
   STDIN  = IO::FileDescriptor.from_stdio(0)
 
-  # The standard output file descriptor. Gets flushed when a newline is written.
+  # The standard output file descriptor.
   #
   # Typically used to output data and information.
+  #
+  # NOTE: Gets flushed when a newline is written.
   STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
 
-  # The standard error file descriptor. Gets flushed when a newline is written.
+  # The standard error file descriptor.
   #
   # Typically used to output error messages and diagnostics.
+  #
+  # NOTE: Gets flushed when a newline is written.
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
 
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -44,43 +44,43 @@ ARGV = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
 
 # An `IO` for reading files from `ARGV`.
 #
-# `ARGF` assumes that `ARGV` contains only filenames.
+# Usage example:
 #
-# After a file from `ARGV` has been read, it's removed from `ARGV`.
+# `program.cr`:
+# ```
+# puts ARGF.gets_to_end
+# ```
 #
-# Example:
-#
-# A file to read from: (called `file`)
-#
+# A file to read from: (`file`)
 # ```text
 # 123
 # ```
 #
+# ```text
+# $ crystal build program.cr
+# $ ./program file
+# 123
+# $ ./program file file
+# 123123
+# $ # If ARGV is empty, ARGF reads from STDIN instead:
+# $ echo "hello" | ./program
+# hello
+# $ ./program unknown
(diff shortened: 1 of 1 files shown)


RX14:
I don't want to add `ARGC`. If anything i'd rename ARGV to just ARGS but thats probably not worth changing.

straight-shoota:
`ARGC` is only necessary when accessing `ARGV` is unsafe. That's not the case, so we don't need it. The size is inherently included in `ARGV`.

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
...
 
+  # The standard input file descriptor.
   STDIN  = IO::FileDescriptor.from_stdio(0)
+  # The standard output file descriptor. Gets flushed when a newline is written.
   STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  # The standard error file descriptor. Gets flushed when a newline is written.
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
 
+# Path to the current temporary program.
```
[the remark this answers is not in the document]
r00ster91: It's deleted after the program exited. So maybe:
```suggestion
# Path to the current temporary program in the cache.
# The program is deleted after it finished executing.
```

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
 {% if flag?(:win32) %}
+  # The standard input file descriptor.
   STDIN  = IO::FileDescriptor.new(0)
+  # The standard output file descriptor. Gets flushed when a newline is written.
```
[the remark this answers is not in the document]
r00ster91: I don't think these one line comments here really need newlines but these below with this one long comment might look confusing so I'll add some newlines there.

src/kernel.cr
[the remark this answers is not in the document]
straight-shoota: Sorry, what are you talking about?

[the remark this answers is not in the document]
r00ster91: When I put out `PROGRAM_NAME`, I get e.g. `/home/r00ster/.cache/crystal/crystal-run-test.tmp`.
A `.tmp` file which stands for "temporary". And it's in the cache. When I run `File.exists?(PROGRAM_NAME)` in the program, it gives me `true` and when I then after the program finished executing, run `/home/r00ster/.cache/crystal/crystal-run-test.tmp`, it gives me `-bash: /home/r00ster/.cache/crystal/crystal-run-test.tmp: No such file or directory`.
So it's gone after the program finished.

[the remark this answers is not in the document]
straight-shoota: When you compile and execute the program with `crystal run`, only compiles to a temporary file and deletes it. That's simply the behaviour of this specific build mechanism and by no means relevant for the description of this constant.

The typical distribution is to compile once (using `crystal build`) and (be able to) run the compiled program multiple times. Nothing will be deleted.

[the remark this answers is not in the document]
r00ster91: Oooh now it makes sense. Yes you are right, sorry. It's not always a temporary file.

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
...
 
+  # The standard input file descriptor.
   STDIN  = IO::FileDescriptor.from_stdio(0)
+  # The standard output file descriptor. Gets flushed when a newline is written.
   STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  # The standard error file descriptor. Gets flushed when a newline is written.
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
 
+# Path to the current program.
```
straight-shoota: It's not necessarily a path. Better call it `Name of the current program`.

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
 {% if flag?(:win32) %}
+  # The standard input file descriptor.
```
RX14: "The standard input file descriptor" doesn't really add much more info than the name. I feel like more needs to be added here.

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
...
 
+  # The standard input file descriptor.
   STDIN  = IO::FileDescriptor.from_stdio(0)
+  # The standard output file descriptor. Gets flushed when a newline is written.
   STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  # The standard error file descriptor. Gets flushed when a newline is written.
   STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
 {% end %}
 
+# The name of the current program.
```
RX14: This is actually the name the program was called by under the shell. For example `/foo/bar arg` in the shell gets `/foo/var` as `PROGRAM_NAME`, whereas just `bar` in the path gets just `bar` as `PROGRAM_NAME`.

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
...
 PROGRAM_NAME = String.new(ARGV_UNSAFE.value)
-ARGV         = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
-ARGF         = IO::ARGF.new(ARGV, STDIN)
+
+# The arguments passed to the program.
```
RX14: An array containing the arguments passed the program?

src/kernel.cr
```diff
@@ -1,18 +1,39 @@
...
+ARGV = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
+
+# An `IO` which reads from `ARGV` if arguments are specified.
```
RX14: This description still seems confusing.

straight-shoota: Just copy what Ruby says. https://ruby-doc.org/core-2.5.0/ARGF.html

src/kernel.cr
r00ster91: Probably that it can contain output from piping and that keyboard input gets redirected to it?
```
Might contain piped output.
Keyboard input gets redirected here.
```

src/kernel.cr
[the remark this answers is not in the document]
r00ster91: I don't think `NOTE` is needed here.
It's already pretty much 2 times stated here with the code `{ |f| f.flush_on_newline = true }` that the file descriptor gets flushed on a newline.
And now since the last commit it would look like this:
```cr
# The standard output file descriptor.
#
# Typically used to output data and information.
#
# NOTE: Gets flushed when a newline is written.
```
I think that's just unnecessarily long and can be shorter.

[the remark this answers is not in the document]
Sija: I'd say it fits better as a `NOTE` since it's not in any way description of a constant but rather a _note_ about its behaviour. And it's not longer since `NOTE` is shown differently (and more prominently) in docs.

src/kernel.cr
```diff
@@ -1,18 +1,88 @@
...
 
+# The name, the program was called with.
 PROGRAM_NAME = String.new(ARGV_UNSAFE.value)
-ARGV         = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
-ARGF         = IO::ARGF.new(ARGV, STDIN)
+
+# An array of arguments passed to the program.
+ARGV = Array.new(ARGC_UNSAFE - 1) { |i| String.new(ARGV_UNSAFE[1 + i]) }
+
+# An `IO` for reading files from `ARGV`.
```
bcardiff: I would suggest to start this description with a usage example from an executable in a terminal.

Like
```
$ crystal build cli.cr
$ echo "foo" | ./cli
$ ./cli file.txt
```

What will `ARGF.gets_to_end` will have in both scenarios?
What will the behaviour be if multiple arguments are given or not?
What will happen if some arguments are not files and what can be done in that scenario.

I things that would be a more easy to understand flow for ARGF.

r00ster91: Like this?
```cr
# An `IO` for reading files from `ARGV`.
#
# Usage example:
#
# The program: (`program.cr`):
# ```
# puts ARGF.gets_to_end
# ```
#
# A file to read from: (`file`)
# ```text
# 123
# ```
#
# ```
# $ crystal build program.cr
# $ echo "hello" | ./program
# hello
# $ ./program file
# 123
# $ ./program file file
# 123123
# $ ./program unknown
# Unhandled exception: Error opening file 'unknown' with mode 'r': No such file or directory (Errno)
# ...
# ```
#
# After a file from `ARGV` has been read, it's removed from `ARGV`.
#
# You can manipulate `ARGV` yourself ...
```

PR #6972, opened 2018-10-21
Implement System for Win32

markrjr:
Pretty simple implementation of the hostname and cpu count functions.

### DIFF
--- a/src/crystal/system/win32/hostname.cr
+++ b/src/crystal/system/win32/hostname.cr
@@ -1,0 +1,12 @@
+require "c/ntsysteminfo"
+
+module Crystal::System
+  def self.hostname
+    name_size = LibC::NT_COMPUTER_NAME_SIZE.to_u32
+    unless LibC.NTGetComputerNameExA(LibC::ComputerNameFormat::ComputerNameDnsHostname, out machine_name, pointerof(name_size))
+        raise Errno.new("Failed to get machine hostname.")
+    end
+    name = String.new machine_name.to_unsafe
+    name
+  end
+end
--- a/src/lib_c/x86_64-windows-msvc/c/int_safe.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/int_safe.cr
@@ -1,3 +1,4 @@
 lib LibC
   alias DWORD = UInt32
+  alias DWORD_PTR = UInt32*
 end
--- a/src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
@@ -1,0 +1,62 @@
+require "c/winnt"
+require "c/win_def"
+require "c/int_safe"
+
+lib LibC
+
+  #GetSystemInfo - https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getsysteminfo
+  
+  fun NTGetSystemInfo = GetSystemInfo(system_info : SystemInfo*)
+  
+  PROCESSOR_ARCHITECTURE_AMD64 = 9
+  PROCESSOR_ARCHITECTURE_ARM = 5
+  PROCESSOR_ARCHITECTURE_ARM64 = 12
+  PROCESSOR_ARCHITECTURE_IA64 = 6
+  PROCESSOR_ARCHITECTURE_INTEL = 0
+  PROCESSOR_ARCHITECTURE_UNKNOWN = 0xffff
+
+  struct ProcessorInfo
+    wProcessorArchitecture : WORD
+    wReserved : WORD
+  end
+
+  union OEMProcessorInfo
+    dwOemId : DWORD
+    processorInfo : ProcessorInfo
+  end
+
+  struct SystemInfo
+    oemProcessorInfo : OEMProcessorInfo
+    dwPageSize : DWORD
+    lpMinimumApplicationAddress : LPVOID
+    lpMaximumApplicationAddress : LPVOID
+    dwActiveProcessorMask : DWORD_PTR
+    dwNumberOfProcessors : DWORD
+    dwProcessorType : DWORD
+    dwAllocationGranularity : DWORD
+    wProcessorLevel : WORD
+    wProcessorRevision : WORD
+
+  end
+
+
+  #GetComputerNameExA - https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getcomputernameexa
+
+  fun NTGetComputerNameExA = GetComputerNameExA(computer_name_format : ComputerNameFormat,
+                                              machine_name : CHAR[NT_COMPUTER_NAME_SIZE]*,
+                                              machine_name_size_ptr : DWORD_PTR) : BOOLEAN
+
+  NT_COMPUTER_NAME_SIZE = 256
+
+  enum ComputerNameFormat
+    ComputerNameNetBIOS
+    ComputerNameDnsHostname
+    ComputerNameDnsDomain
+    ComputerNameDnsFullyQualified
+    ComputerNamePhysicalNetBIOS
+    ComputerNamePhysicalDnsHostname
+    ComputerNamePhysicalDomain
+    ComputerNamePhysicalDnsFullyQualified
+  end
+
+end
--- a/src/lib_c/x86_64-windows-msvc/c/winnt.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/winnt.cr
@@ -8,6 +8,7 @@ lib LibC
   alias LPWSTR = WCHAR*
   alias LPWCH = WCHAR*
 
+  alias LPVOID = Void*
   alias HANDLE = Void*
 
   INVALID_FILE_ATTRIBUTES      = DWORD.new(-1)
--- a/src/crystal/system/win32/cpucount.cr
+++ b/src/crystal/system/win32/cpucount.cr
@@ -1,0 +1,8 @@
+require "c/ntsysteminfo"
+
+module Crystal::System
+  def self.cpu_count
+    LibC.NTGetSystemInfo(out system_info)
+    system_info.dwNumberOfProcessors
+  end
+end


Sija:
`crystal tool format` please

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
```diff
@@ -0,0 +1,54 @@
+require "c/winnt"
+require "c/win_def"
+require "c/int_safe"
+
+lib LibC
+  fun NTGetSystemInfo = GetSystemInfo(system_info : SystemInfo*)
```
straight-shoota: Is there a reason for `NT` prefix? I don't think that should be there.

straight-shoota: This function should be in `sysinfoapi.cr` (equivalent to `sysinfoapi.h`).

straight-shoota: It's probably better to use [`GetNativeSystemInfo`](https://docs.microsoft.com/de-de/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo), though, which is generally equivalent but also works as expected WOW64 (32-bit layer on 64-but Windows).

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
```diff
@@ -0,0 +1,54 @@
...
+    wProcessorArchitecture : WORD
+    wReserved : WORD
+  end
+
+  union OEMProcessorInfo
+    dwOemId : DWORD
+    processorInfo : ProcessorInfo
+  end
+
+  struct SystemInfo
```
straight-shoota: `SYSTEM_INFO`

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
```diff
@@ -0,0 +1,54 @@
...
+    lpMaximumApplicationAddress : LPVOID
+    dwActiveProcessorMask : DWORD_PTR
+    dwNumberOfProcessors : DWORD
+    dwProcessorType : DWORD
+    dwAllocationGranularity : DWORD
+    wProcessorLevel : WORD
+    wProcessorRevision : WORD
+  end
+
+  fun NTGetComputerNameExA = GetComputerNameExA(computer_name_format : ComputerNameFormat,
```
straight-shoota: Ditto, no `NT`.

straight-shoota: And use `GetComputerNameExW` for unicode compat.

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
```diff
@@ -0,0 +1,54 @@
...
+                                                machine_name : CHAR[NT_COMPUTER_NAME_SIZE]*,
+                                                machine_name_size_ptr : DWORD_PTR) : BOOLEAN
+
+  NT_COMPUTER_NAME_SIZE = 256
+
+  enum ComputerNameFormat
```
straight-shoota: `COMPUTER_NAME_FORMAT`

markrjr:
I think I hit all the PR comment fixes in the latest push. Let me know if there's anything else.

src/crystal/system/win32/hostname.cr
```diff
@@ -0,0 +1,12 @@
+require "c/sysinfoapi"
+
+module Crystal::System
+  def self.hostname
+    name_size = LibC::MAX_COMPUTER_NAME_SIZE.to_u32
+    unless LibC.GetComputerNameExW(LibC::COMPUTER_NAME_FORMAT::ComputerNameDnsHostname, out machine_name, pointerof(name_size))
+      raise WinError.new("Failed to get machine hostname")
```
straight-shoota: `WinError` and `Errno` usually only receive the name of the failing function as message (= `GetComputerNameExW`).

src/crystal/system/win32/hostname.cr
```diff
@@ -0,0 +1,12 @@
+require "c/sysinfoapi"
+
+module Crystal::System
+  def self.hostname
+    name_size = LibC::MAX_COMPUTER_NAME_SIZE.to_u32
+    unless LibC.GetComputerNameExW(LibC::COMPUTER_NAME_FORMAT::ComputerNameDnsHostname, out machine_name, pointerof(name_size))
```
straight-shoota: According to the description of [GetComputerNameExW](https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getcomputernameexw), I'm not sure this usage is correct. Might be reading it wrong... But `MAX_COMPUTER_NAME_SIZE` doesn't make sense as an size limit, since you're not providing a buffer. And the name could actually have more characters because it's a DNS name.

src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr
```diff
@@ -0,0 +1,54 @@
+require "c/winnt"
+require "c/win_def"
+require "c/int_safe"
+
+lib LibC
+  fun GetNativeSystemInfo = GetNativeSystemInfo(system_info : SYSTEM_INFO*)
+
+  PROCESSOR_ARCHITECTURE_AMD64   =      9
```
straight-shoota: These constants are not used anywhere.

src/crystal/system/win32/hostname.cr
ysbaddaden: According to the docs, yes:
- it needs the buffer to be allocated —`out` means that the function will allocate it which it doesn't;
- it could be called twice, once with `lpBuffer` as `nil` so that it will set `nSize`, so we can allocate a properly sized buffer and call the function again that will fill the value.

src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr
ysbaddaden: They're unrelated to this PR.

markrjr: Yeah, I can remove them.

I originally put them just in case we wanted to exposed a processor architecture function or something later which comes from the same `GetNativeSystemInfo` call, but I since such a function doesn't exist yet, the associated constants are just sort of hanging out as dead code.

src/crystal/system/win32/hostname.cr
markrjr: Will change this around.

src/crystal/system/win32/hostname.cr
markrjr: It's a bit weird of API design on MS's part, imo. I'd prefer if the function just returned name length (and perhaps -1 if the call failed.) 

@ysbaddaden is right. 

DNS names have a 255 character limit [IIRC](https://blogs.msdn.microsoft.com/oldnewthing/20120412-00/?p=7873), so I just initialized it to the max since the function wants it allocated beforehand and will set the value accordingly. Since I use a StaticArray on the other side, technically I'm wasting a tiny bit of memory (since the DNS name will probably be much less than 255 chars) but I do think that this is a bit cleaner than 2 calls.

RX14: It's a standard pattern for windows, you can use [`retry_wstr_buffer`](https://github.com/crystal-lang/crystal/blob/master/src/crystal/system/windows.cr#L3) for this.

src/lib_c/x86_64-windows-msvc/c/int_safe.cr
```diff
@@ -1,3 +1,4 @@
 lib LibC
   alias DWORD = UInt32
+  alias DWORD_PTR = UInt32*
```
RX14: I just used `DWORD*` instead of `DWORD_PTR` elsewhere, no need to copy *all* the aliases

src/lib_c/x86_64-windows-msvc/c/winnt.cr
```diff
@@ -8,6 +8,7 @@ lib LibC
   alias LPWSTR = WCHAR*
   alias LPWCH = WCHAR*
 
+  alias LPVOID = Void*
```
RX14: Ditto, `Void*` instead of `LPVOID*` is fine.

src/crystal/system/win32/hostname.cr
markrjr: Would this be the correct way to use `retry_wstr_buffer` since `GetComputerNameExW` wants a pointer to the buffer length? This compiles and runs.
```
def hostname
  retry_wstr_buffer do |buffer, small_buf|
    name_size = buffer.size.to_u
    if LibC.GetComputerNameExW(LibC::COMPUTER_NAME_FORMAT::ComputerNameDnsHostname, buffer, pointerof(name_size))
      return String.from_utf16(buffer[0, name_size])
    else
      raise WinError.new("Failed to get machine hostname")
    end
  end
end
```

RX14: No, you need to either `break` the final `String`, `next` with the next size of the buffer, or raise. You have no provision for actually retrying, making `retry_wstr_buffer` pointless.

straight-shoota:
@markrjr This PR needs to address the review comments.

markrjr:
I've got most of it done locally, but have not had time to finish the last bits. I'll commit by the end of the week.

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
markrjr: Switched to using [GetNativeSystemInfo](https://github.com/crystal-lang/crystal/blob/b9b590442aa9d0afcce2cb16af0e65e8a846ae76/src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr#L6).

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
markrjr: Renamed [here](https://github.com/crystal-lang/crystal/blob/b9b590442aa9d0afcce2cb16af0e65e8a846ae76/src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr#L18).

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
markrjr: Fixed [here](https://github.com/crystal-lang/crystal/blob/b9b590442aa9d0afcce2cb16af0e65e8a846ae76/src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr#L31).

src/lib_c/x86_64-windows-msvc/c/ntsysteminfo.cr
markrjr: Fixed [here](https://github.com/crystal-lang/crystal/blob/b9b590442aa9d0afcce2cb16af0e65e8a846ae76/src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr#L35).

src/crystal/system/win32/hostname.cr
markrjr: Fixed [here](https://github.com/crystal-lang/crystal/blob/b9b590442aa9d0afcce2cb16af0e65e8a846ae76/src/crystal/system/win32/hostname.cr).

src/crystal/system/win32/hostname.cr
markrjr: I was initially a little confused by what the two`yield` statements were doing. I know that they yield execution, but one was used for size and the other for the content of the buffer. I think I understand now, thanks for explaining. Newest PR has these changes.

src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr
markrjr: Removed constants.

src/lib_c/x86_64-windows-msvc/c/winnt.cr
markrjr: Removed this.

src/lib_c/x86_64-windows-msvc/c/int_safe.cr
markrjr: Removed this.

src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr
```diff
@@ -0,0 +1,45 @@
...
+    lpMaximumApplicationAddress : Void*
+    dwActiveProcessorMask : DWORD*
+    dwNumberOfProcessors : DWORD
+    dwProcessorType : DWORD
+    dwAllocationGranularity : DWORD
+    wProcessorLevel : WORD
+    wProcessorRevision : WORD
+  end
+
+  fun GetComputerNameExW = GetComputerNameExW(computer_name_format : COMPUTER_NAME_FORMAT,
```
Sija: `GetComputerNameExW = ` part is redundant.

markrjr: Fixed in 808e56c.

Sija: Nice! remember about running `crystal tool format` afterwards :)

markrjr: Yup, did it on the commit just before this one.

r00ster91: Needs to be done again, CI is failing:
```
Error: formatting './src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr' produced changes
```

markrjr: Yeah, weird, will fix.

src/lib_c/x86_64-windows-msvc/c/sysinfoapi.cr
```diff
@@ -0,0 +1,45 @@
...
+    wProcessorRevision : WORD
+  end
+
+  fun GetComputerNameExW(computer_name_format : COMPUTER_NAME_FORMAT,
+                         machine_name : LPWSTR,
+                         machine_name_size : DWORD*) : BOOLEAN
+
+  enum COMPUTER_NAME_FORMAT
+    ComputerNameNetBIOS
+    ComputerNameDnsHostname
```
Sija: Would it make sense to drop `ComputerName` prefix?

markrjr: It's there in the [docs](https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-getcomputernameexw), so I just left it in.

spec/std/system_spec.cr
```diff
@@ -4,16 +4,21 @@ require "system"
...
+      $?.success?.should be_true                        # The hostname command has to be available
       hostname = System.hostname
       hostname.should eq(shell_hostname)
     end
   end
 
   describe "cpu_count" do
     it "returns current CPU count" do
-      shell_cpus = `getconf _NPROCESSORS_ONLN 2>/dev/null || nproc --all 2>/dev/null || grep -sc '^processor' /proc/cpuinfo || sysctl -n hw.ncpu 2>/dev/null`.to_i
+      shell_cpus = 0
```
r00ster91: This line seems unnecessary to me.

spec/std/system_spec.cr
```diff
@@ -4,16 +4,21 @@ require "system"
 describe System do
   describe "hostname" do
     it "returns current hostname" do
-      shell_hostname = `hostname`.strip
-      $?.success?.should be_true # The hostname command has to be available
+      shell_hostname = {{ `hostname`.stringify }}.strip # Workaround for inability to execute shell commands on Windows.
```
r00ster91: Maybe add a TODO so this can be changed once `Process` is ported?

src/crystal/system.cr
```diff
@@ -14,4 +14,9 @@ require "./system/unix/hostname"
   require "./system/unix/sysctl_cpucount"
```
r00ster91: This whole macro condition should probably be changed to the one posted here: https://github.com/crystal-lang/crystal/pull/6972#issuecomment-445560027.
There's still one UNIX require here: [`system.cr:11`](https://github.com/crystal-lang/crystal/blob/ba46dd965f5b4871ff87f0fb2d936647e052c889/src/crystal/system.cr#L11) which is not in any macro branch.

spec/std/system_spec.cr
straight-shoota: This shouldn't be changed in the first place. The spec works fine as is on Unix and it will on Windows as soon as `Process` is ported.

src/crystal/system.cr
markrjr: I've changed that to the code from the comment.

spec/std/system_spec.cr
markrjr: I've changed this back, though I thought on gitter you suggested changing this to a macro.

spec/std/system_spec.cr
markrjr: If I don't declare it, the equality check `cpu_count.should eq(shell_cpus)` fails because it can't find a declaration for the variable.

straight-shoota: That's odd because it should be defined in all macro branches.

src/crystal/system.cr
r00ster91: No, this macro condition is still different from the one in https://github.com/crystal-lang/crystal/pull/6972#issuecomment-445560027.
The UNIX require in [`system.cr:11`](https://github.com/crystal-lang/crystal/blob/ba46dd965f5b4871ff87f0fb2d936647e052c889/src/crystal/system.cr#L11) is still on the top-level but it should be moved to the `flag?(:unix)` macro branch.
The `flag?(:freebsd) || flag?(:openbsd)` macro branch should be removed and this:
```cr
  {% if flag?(:freebsd) || flag?(:openbsd) %}
    require "./system/unix/sysctl_cpucount"
  {% else %}
    require "./system/unix/sysconf_cpucount"
  {% end %}
```
should be added to the `flag?(:unix)` macro branch.

src/crystal/system.cr
```diff
@@ -8,10 +8,17 @@ module Crystal::System
   # def self.cpu_count
 end
 
-require "./system/unix/hostname"
+{% if flag?(:unix) %}
+  require "./system/unix/hostname"
 
-{% if flag?(:freebsd) || flag?(:openbsd) %}
+  {% if flag?(:freebsd) || flag?(:openbsd) %}
   require "./system/unix/sysctl_cpucount"
```
r00ster91: A little bit indentation is missing here:
```suggestion
     require "./system/unix/sysctl_cpucount"
```

spec/std/system_spec.cr
```diff
@@ -4,16 +4,21 @@ require "system"
 describe System do
   describe "hostname" do
     it "returns current hostname" do
-      shell_hostname = `hostname`.strip
-      $?.success?.should be_true # The hostname command has to be available
+      shell_hostname = `hostname`.strip # Workaround for inability to execute shell commands on Windows.
```
straight-shoota: The comment is leftover

spec/std/system_spec.cr
straight-shoota: You should be able to remove this line.

src/crystal/system.cr
```diff
@@ -8,10 +8,17 @@ module Crystal::System
...
-require "./system/unix/hostname"
+{% if flag?(:unix) %}
+  require "./system/unix/hostname"
 
-{% if flag?(:freebsd) || flag?(:openbsd) %}
-  require "./system/unix/sysctl_cpucount"
-{% elsif flag?(:unix) %}
-  require "./system/unix/sysconf_cpucount"
+  {% if flag?(:freebsd) || flag?(:openbsd) %}
+     require "./system/unix/sysctl_cpucount"
```
r00ster91: Hmm there seems to be a little GitHub indentation bug?
I suggested 2 spaces (https://github.com/crystal-lang/crystal/pull/6972#discussion_r240786893) but now there are 3 here.
Can one space be removed? That'd be the last thing from me.

src/crystal/system/win32/hostname.cr
```diff
@@ -0,0 +1,16 @@
+require "c/sysinfoapi"
+
+module Crystal::System
+  def self.hostname
+    retry_wstr_buffer do |buffer, small_buf|
+      name_size = buffer.size.to_u
```
RX14: Probably `LibC::DWORD.new(buffer.size)` would be better here, in case `to_u` changes.

src/crystal/system.cr
markrjr: Fixed in 2c59c80.

src/crystal/system/win32/hostname.cr
markrjr: Fixed in 2c59c80.

spec/std/system_spec.cr
markrjr: Fixed in 2c59c80.

spec/std/system_spec.cr
markrjr: [CI](https://circleci.com/gh/crystal-lang/crystal/15436?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) fails with that removed.

RX14:
```cr
shell_cpus = {% if ...
```

markrjr: Ah, I didn't know that you could conditionally define stuff using macros.

straight-shoota:
@markrjr You're initial patch wasn't bad. Keep it on! =)

sdogruyol:
Thanks for the great first patch @markrjr :tada:

PR #5021, opened 2017-09-22
Added `pkcs5_pbkdf2_hmac` to `lib_crypto.cr` and `pkcs5.cr`

aurimasniekis:
Added `pkcs5_pbkdf2_hmac` function to support pbkdf2 with multiple algorithm

PR #6975, opened 2018-10-21
Expose LibCrypt's PKCS5_PBKDF2_HMAC

### LINKED ISSUE #6973: Expose LibCrypt PKCS5_PBKDF2_HMAC
### Issue Kind: Feature Request
Please add the binding to libcrypt's PKCS5_PBKDF2_HMAC method

Documented on https://www.openssl.org/docs/man1.1.0/crypto/PKCS5_PBKDF2_HMAC_SHA1.html

mniak:
Fixes #6973 
Replaces pull request #5021

### DIFF
--- a/spec/std/openssl/pkcs5_spec.cr
+++ b/spec/std/openssl/pkcs5_spec.cr
@@ -1,5 +1,6 @@
 require "spec"
 require "openssl/pkcs5"
+require "openssl/algorithm"
 
 describe OpenSSL::PKCS5 do
   it "computes pbkdf2_hmac_sha1" do
@@ -12,4 +13,45 @@ describe OpenSSL::PKCS5 do
       OpenSSL::PKCS5.pbkdf2_hmac_sha1("password", "salt", iterations, key_size).hexstring.should eq expected
     end
   end
+
+  {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.0.0") >= 0 %}
+  it "computes pbkdf2_hmac" do
+    [
+      {OpenSSL::Algorithm::MD4, 1, 16, "1857f69412150bca4542581d0f9e7fd1"},
+      {OpenSSL::Algorithm::MD4, 1, 32, "1857f69412150bca4542581d0f9e7fd19332ff5c0b820cb0172457a29c5519be"},
+      {OpenSSL::Algorithm::MD4, 2**16, 16, "3d87c982c8c4223f4af39406ac3882e6"},
+      {OpenSSL::Algorithm::MD4, 2**16, 32, "3d87c982c8c4223f4af39406ac3882e6e6b92685dcf89f74df8caf7500b41883"},
+      {OpenSSL::Algorithm::MD5, 1, 16, "f31afb6d931392daa5e3130f47f9a9b6"},
+      {OpenSSL::Algorithm::MD5, 1, 32, "f31afb6d931392daa5e3130f47f9a9b6e8e72029d8350b9fb27a9e0e00b9d991"},
+      {OpenSSL::Algorithm::MD5, 2**16, 16, "8b4ffd76e400c3b74b3d0fbfd9232048"},
+      {OpenSSL::Algorithm::MD5, 2**16, 32, "8b4ffd76e400c3b74b3d0fbfd9232048762c86fe7684992c6f581f073f6625ee"},
+      {OpenSSL::Algorithm::RIPEMD160, 1, 16, "b725258b125e0bacb0e2307e34feb16a"},
+      {OpenSSL::Algorithm::RIPEMD160, 1, 32, "b725258b125e0bacb0e2307e34feb16a4d0d6aed6cb4b0eee458fc1829020428"},
+      {OpenSSL::Algorithm::RIPEMD160, 2**16, 16, "93a8e007de2608e54911684cbebe2780"},
+      {OpenSSL::Algorithm::RIPEMD160, 2**16, 32, "93a8e007de2608e54911684cbebe27808cc39fa59de9acdf74492155b46c4d2d"},
+      {OpenSSL::Algorithm::SHA1, 1, 16, "0c60c80f961f0e71f3a9b524af601206"},
+      {OpenSSL::Algorithm::SHA1, 1, 32, "0c60c80f961f0e71f3a9b524af6012062fe037a6e0f0eb94fe8fc46bdc637164"},
+      {OpenSSL::Algorithm::SHA1, 2**16, 16, "1b345dd55f62a35aecdb9229bc7ae95b"},
+      {OpenSSL::Algorithm::SHA1, 2**16, 32, "1b345dd55f62a35aecdb9229bc7ae95b305a8d538940134627e46f82d3a41e5e"},
+      {OpenSSL::Algorithm::SHA224, 1, 16, "3c198cbdb9464b7857966bd05b7bc92b"},
+      {OpenSSL::Algorithm::SHA224, 1, 32, "3c198cbdb9464b7857966bd05b7bc92bc1cc4e6e63155d4e490557fd85989497"},
+      {OpenSSL::Algorithm::SHA224, 2**16, 16, "53a7f042a8154092058cfe87e7fbf1c1"},
+      {OpenSSL::Algorithm::SHA224, 2**16, 32, "53a7f042a8154092058cfe87e7fbf1c1f96826a9a2ffd8bcfda50bb9f60786f0"},
+      {OpenSSL::Algorithm::SHA256, 1, 16, "120fb6cffcf8b32c43e7225256c4f837"},
+      {OpenSSL::Algorithm::SHA256, 1, 32, "120fb6cffcf8b32c43e7225256c4f837a86548c92ccc35480805987cb70be17b"},
+      {OpenSSL::Algorithm::SHA256, 2**16, 16, "4156f668bb31db3a17f4d1b91424ef0d"},
+      {OpenSSL::Algorithm::SHA256, 2**16, 32, "4156f668bb31db3a17f4d1b91424ef0d417ad1f35d055aceaebd8da0f6a44b7e"},
+      {OpenSSL::Algorithm::SHA384, 1, 16, "c0e14f06e49e32d73f9f52ddf1d0c5c7"},
+      {OpenSSL::Algorithm::SHA384, 1, 32, "c0e14f06e49e32d73f9f52ddf1d0c5c7191609233631dadd76a567db42b78676"},
+      {OpenSSL::Algorithm::SHA384, 2**16, 16, "c7b5b0b726f6556587cced08d184253b"},
+      {OpenSSL::Algorithm::SHA384, 2**16, 32, "c7b5b0b726f6556587cced08d184253bc9d2eb802db134fb9029b86ab25e7cd0"},
+      {OpenSSL::Algorithm::SHA512, 1, 16, "867f70cf1ade02cff3752599a3a53dc4"},
+      {OpenSSL::Algorithm::SHA512, 1, 32, "867f70cf1ade02cff3752599a3a53dc4af34c7a669815ae5d513554e1c8cf252"},
+      {OpenSSL::Algorithm::SHA512, 2**16, 16, "6f64c3f8023813d8c2cab43cabfaa65e"},
+      {OpenSSL::Algorithm::SHA512, 2**16, 32, "6f64c3f8023813d8c2cab43cabfaa65ed061822afe974060d8079d122fb869f4"},
+    ].each do |(algorithm, iterations, key_size, expected)|
+      OpenSSL::PKCS5.pbkdf2_hmac("password", "salt", iterations, algorithm, key_size).hexstring.should eq expected
+    end
+  end
+  {% end %}
 end
--- a/spec/std/openssl/pkcs5_spec.cr
+++ b/spec/std/openssl/pkcs5_spec.cr
@@ -15,43 +15,43 @@ describe OpenSSL::PKCS5 do
   end
 
   {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.0.0") >= 0 %}
-  it "computes pbkdf2_hmac" do
-    [
-      {OpenSSL::Algorithm::MD4, 1, 16, "1857f69412150bca4542581d0f9e7fd1"},
-      {OpenSSL::Algorithm::MD4, 1, 32, "1857f69412150bca4542581d0f9e7fd19332ff5c0b820cb0172457a29c5519be"},
-      {OpenSSL::Algorithm::MD4, 2**16, 16, "3d87c982c8c4223f4af39406ac3882e6"},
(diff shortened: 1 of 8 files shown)


src/openssl/pkcs5.cr
```diff
@@ -8,4 +8,23 @@ module OpenSSL::PKCS5
     end
     buffer
   end
+
+  def self.pbkdf2_hmac(secret, salt, iterations = 2**16, algorithm : Symbol = :sha1, key_size = 64) : Bytes
```
j8r: Use `String` instead of `Symbol`. Here, I'd rather use an `enum`

mniak: Sure!
I declared a new enum named DigestAlgorithm.
I'm not sure about the following:
- Location (should I declare this inside the module?)
- Casing (for acronims in enum members)
- Enum naming (is the name DigestAlgorithm a good name?)

mniak:
I declared a new enum named DigestAlgorithm.
I declared a new enume `DigestAlgorithm`
I'm not sure about the following:
- Location (should I declare this inside the module?)
- Casing (for acronims in enum members)
- Enum naming (is the name DigestAlgorithm a good name?)

j8r:
IMO just `Digest` is fine, like https://crystal-lang.org/api/master/OpenSSL/Digest.html. You can also drop the `,` after each enum element. For the location, it's in the right place.

mniak:
Added a version check that enables the feature only when OpenSSL >= 1.0.0
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Added a version check that enables the feature only when OpenSSL >= 1.0.0
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Added a version check that enables the feature only when OpenSSL >= 1.0.0
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Added a version check that enables the feature only when OpenSSL >= 1.0.0
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Made the chanes you suggested.
Also added a version check that enables the feature only when OpenSSL >= 1.0.0 (when the underlying method became available).
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Made the chanes you suggested.
Also added a version check that enables the feature only when OpenSSL >= 1.0.0 (when the underlying method became available).
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Made the chanes you suggested.
Also added a version check that enables the feature only when OpenSSL >= 1.0.0 (when the underlying method became available).
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Made the chanes you suggested.
Also added a version check that enables the feature only when OpenSSL >= 1.0.0 (when the underlying method became available).
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

mniak:
Made the chanes you suggested.
Also added a version check that enables the feature only when OpenSSL >= 1.0.0 (when the underlying method became available).
Suggestion made in https://github.com/crystal-lang/crystal/pull/5021#issuecomment-334394154

src/openssl/pkcs5.cr
```diff
@@ -1,11 +1,41 @@
...
+          when DigestAlgorithm::MD4       then LibCrypto.evp_md4
+          when DigestAlgorithm::MD5       then LibCrypto.evp_md5
+          when DigestAlgorithm::RIPEMD160 then LibCrypto.evp_ripemd160
+          when DigestAlgorithm::SHA1      then LibCrypto.evp_sha1
+          when DigestAlgorithm::SHA224    then LibCrypto.evp_sha224
+          when DigestAlgorithm::SHA256    then LibCrypto.evp_sha256
+          when DigestAlgorithm::SHA384    then LibCrypto.evp_sha384
+          when DigestAlgorithm::SHA512    then LibCrypto.evp_sha512
+          else                                 raise "Unsupported digest algorithm: #{algorithm}"
+          end
```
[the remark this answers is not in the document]
mniak: Where should the enum be placed then?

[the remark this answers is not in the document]
mniak: If I use the enum in the class `OpenSSL::HMAC` then I should place it in a common location. The obvious choice would be put it in `OpenSSL::Digest` but there is already a class with this name https://github.com/crystal-lang/crystal/blob/master/src/openssl/digest/digest.cr

spec/compiler/crystal/tools/init_spec.cr
```diff
@@ -93,13 +93,13 @@ module Crystal
       describe_file "example/README.md" do |readme|
         readme.should contain("# example")
 
+        readme.should contain(%{1. Add the dependency to your `shard.yml`:})
```
mniak: # This change came when I pulled from the upstream master

spec/compiler/crystal/tools/init_spec.cr
```diff
@@ -93,13 +93,13 @@ module Crystal
...
         readme.should contain(%{```yaml
 dependencies:
   example:
     github: jsmith/example
 ```})
-
```
mniak: It dit not change this file also. It was changed in the upstream

src/compiler/crystal/compiler.cr
```diff
@@ -625,7 +625,7 @@ module Crystal
         # old one. Generating an `.o` file is what takes most time.
         #
         # However, instead of directly generating the final `.o` file
-        # from the `.bc` file, we generate it to a termporary name (`.o.tmp`)
+        # from the `.bc` file, we generate it to a temporary name (`.o.tmp`)
```
mniak: This change was not mine. Also upstream

src/compiler/crystal/semantic/method_lookup.cr
```diff
@@ -296,7 +296,7 @@ module Crystal
         end
       end
 
-      # If there's a restriction on a double splat, zero matching named arguments don't matc
+      # If there's a restriction on a double splat, zero matching named arguments don't match
       if double_splat && double_splat_restriction &&
```
mniak: this change is not mine

src/compiler/crystal/tools/init/template/readme.md.ecr
```diff
@@ -5,15 +5,13 @@ TODO: Write a description here
 ## Installation
```
mniak: not mine also

mniak:
Sorry for the mess, Im a total newbie

src/openssl/algorithm.cr
```diff
@@ -0,0 +1,28 @@
...
+    RIPEMD160
+    SHA1
+    SHA224
+    SHA256
+    SHA384
+    SHA512
+
+    def to_evp
+      case self
+      when OpenSSL::Algorithm::MD4       then LibCrypto.evp_md4
```
Sija: You can skip `OpenSSL::Algorithm::` part. ditto all below

mniak: I fixed this.

src/openssl/pkcs5.cr
```diff
@@ -1,11 +1,41 @@
 require "openssl"
 
 module OpenSSL::PKCS5
+  enum DigestAlgorithm
+    MD4,
```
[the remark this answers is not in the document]
mniak: ok

mniak:
I feel that I'm close...

spec/std/openssl/pkcs5_spec.cr
```diff
@@ -12,4 +13,45 @@ describe OpenSSL::PKCS5 do
       OpenSSL::PKCS5.pbkdf2_hmac_sha1("password", "salt", iterations, key_size).hexstring.should eq expected
     end
   end
+
+  {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.0.0") >= 0 %}
+  it "computes pbkdf2_hmac" do
```
Sija: Indent please, ditto below.

mniak:
done

src/openssl/hmac.cr
```diff
@@ -13,27 +14,9 @@ class OpenSSL::HMAC
...
-          when :md5       then LibCrypto.evp_md5
-          when :ripemd160 then LibCrypto.evp_ripemd160
-          when :sha1      then LibCrypto.evp_sha1
-          when :sha224    then LibCrypto.evp_sha224
-          when :sha256    then LibCrypto.evp_sha256
-          when :sha384    then LibCrypto.evp_sha384
-          when :sha512    then LibCrypto.evp_sha512
-          else                 raise "Unsupported digest algorithm: #{algorithm}"
-          end
+  # *algorithm* is of type `OpenSSL::Algorithm`.
```
bew: No need to precise the type here in the documentation, you could even not document it at all, it's pretty self explanatory!  If you still want to document this argument, say what it is for / what it'll do

spec/std/openssl/pkcs5_spec.cr
mniak: Not it is indented

src/openssl/hmac.cr
mniak: Ok, i've removed this comment

RX14:
just waiting for another review

src/oauth/signature.cr
```diff
@@ -19,7 +19,7 @@ struct OAuth::Signature
 
   def compute(request, tls, ts, nonce)
     base_string = base_string(request, tls, ts, nonce)
-    Base64.strict_encode(OpenSSL::HMAC.digest :sha1, key, base_string)
```
asterite: There's no need to change this line

src/openssl/pkcs5.cr
```diff
@@ -8,4 +8,15 @@ module OpenSSL::PKCS5
     end
     buffer
   end
+
+  {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.0.0") >= 0 %}
+    def self.pbkdf2_hmac(secret, salt, iterations = 2**16, algorithm : OpenSSL::Algorithm = OpenSSL::Algorithm::SHA1, key_size = 64) : Bytes
```
asterite: In my opinion this method should always exist, but raise in case it's not available because of the OpenSSL version. Please change it. Thank you!

RX14: Good catch!

j8r: To clarify, raising at compile time (e.g `method X not supported with OpenSSL version X.X`) instead of having an unknown method.

asterite: No, this should raise at runtime.

j8r: Why? The program is compiled with an OpenSSL version that doesn't support the aglorithm.

asterite: Imagine this function is used conditionally, with a condition at runtime. Now imagine this is in a shard. I link that shard and want to use it, configuring it to not use this function: I can't, I won't be able to compile the program, even if at runtime I know I won't use the function. It's better to raise at runtime (I can change a config and avoid the exception).

j8r:
You'll be able to compile with a macro condition at compile time. What if you think the function won't be used but it does in a particular situation? This is a case that could be dealed before runtime, and your application will be more robust because this error won't exist.

src/openssl/pkcs5.cr
j8r: You'll be able to compile with a macro condition at compile time. What if you think the function won't be used but it does in a particular situation? This is a case that could be dealed before runtime, and your application will be more robust because this error won't exist.

asterite: Right, then you need to use a macro condition everywhere, which is tedious.

This has already been discussed in the past, I can't remember the conclusion.

RX14: Methods shouldn't just appear and disappear based on environmental reasons. It makes documentation difficult and breaks basic expectations. This has already been decided, i can't find this issue but it was discussed in length.

j8r: That's fair. Probably this can be added somewhere (Wiki?) as contribution best practices.

asterite: Good idea. I'll later add it to the FAQ, or Contributing.

src/oauth/signature.cr
mniak: I've changed this to met the request in https://github.com/crystal-lang/crystal/pull/6975#discussion_r226915346.

I was using symbols in my contribution, but [I was suggested to create an enum](https://github.com/crystal-lang/crystal/pull/6975#discussion_r226878167). Done that.

Then, I was suggested to share this enum with other code pieces. And that is why I changed this line.

src/openssl/pkcs5.cr
mniak: I'm understanding that I should accept the suggestion of @asterite: remove the macro conditional and throw an error at runtime.

Am I right?

src/oauth/signature.cr
asterite: That's correct. However, there's an undocumented feature of the language where you can pass a symbol to a method that has an enum restriction, and the symbol will be automatically converted to the enum. That's why you can change **just this line** back to what it was before.

src/openssl/pkcs5.cr
j8r: Is it **really** needed to support OpenSSL versions older than 1.0.0? Even CentOS 6 isn't on this version (seems to be 1.0.1e).
1.0.0 was released in [2010](https://en.wikipedia.org/wiki/OpenSSL#Major_version_releases)

RX14: @mniak yes

@j8r thats for another issue

RX14:
Shit, this was a breaking change, so this needs a revert and resubmit and wait until 0.28.0

Or we could just say that it's been so long since 0.27.0 that we can have the next release be 0.28.0.

danielwestendorf:
This change does indeed break any code that deals with OpenSSL Ciphers. 

@RX14 Will it be reverted and wait until 0.28.0 will the 0.27.1 release just never happen? I think it's fine either way (from my perspective), there should just be some clarity on which it will be so I can update my [libraries](https://travis-ci.org/danielwestendorf/message_verifier.cr/builds/474902421) accordingly.

asterite:
It will be reverted.

Sija:
@asterite Given the amount of changes wouldn't be better to tag next release as `0.28`?

RX14:
@mniak please resubmit this PR! It's my mistake merging it too early, but it needs to wait until 0.28.0 so it's been reverted in `master`.

straight-shoota:
@Sija We discussed this option in the core team and agreed that it is be better to stick to the plan. Reverting this PR is cheap and doesn't cause any disturbance. We can proceed to prepare for 0.27.1 as the next release and 0.28.0 after that.

mniak:
@RX14 Do I do it simply by creating a new PR with these changes?

PR #6977, opened 2018-10-22
 Add nanoseconds to Time.unix constructor

straight-shoota:
This is an alternative to #6661 to provide a simple interface to convert a `Time` instance from and to primitive data types (`Int64`/`Int32`). Instead of publicly exposing `#total_seconds` which could be used together with `Time.new(seconds, nanoseconds)`, it adds a `nanoseconds` argument to `Time.unix` which can be used together with `#to_unix`.

In return, `Time.new(seconds, nanoseconds)` and `Time.utc(seconds, nanoseconds)` constructors are removed from the public API and only used as internal helper methods.

`#total_seconds` could be replaced by `@seconds`, but I fear it would be confusing in respect to `#seconds` which returns a different value. IMHO it makes sense to keep `#total_seconds` and `#offset_seconds` as internal helper methods (they can't be `private`, though).

See also #5374 (probably fixes that).

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -471,13 +471,14 @@ struct Time
     local(year, month, day, hour, minute, second, nanosecond: nanosecond, location: Location::UTC)
   end
 
+  # :nodoc:
   # Creates a new `Time` instance that corresponds to the number of *seconds*
   # and *nanoseconds* elapsed from epoch (`0001-01-01 00:00:00.0 UTC`)
   # observed in *location*.
   #
   # Valid range for *seconds* is `0..315_537_897_599`.
   # For *nanoseconds* it is `0..999_999_999`.
-  def initialize(*, @seconds : Int64, @nanoseconds : Int32, @location : Location)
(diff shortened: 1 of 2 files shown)


src/time.cr
```diff
@@ -466,13 +466,14 @@ struct Time
...
 
+  # :nodoc:
   # Creates a new `Time` instance that corresponds to the number of *seconds*
   # and *nanoseconds* elapsed from epoch (`0001-01-01 00:00:00.0 UTC`)
   # observed in *location*.
   #
   # Valid range for *seconds* is `0..315_537_897_599`.
   # For *nanoseconds* it is `0..999_999_999`.
-  def initialize(*, @seconds : Int64, @nanoseconds : Int32, @location : Location)
+  protected def initialize(*, @seconds : Int64, @nanoseconds : Int32, @location : Location)
```
wontruefree: why is this moved to protected?

straight-shoota:
Closing stale PR.

PR #6978, opened 2018-10-22
Add Time constructor receiving LibC::FILETIME (win32)

straight-shoota:
Time constructors receiving a LibC time struct should be publicly available because they're essential when converting time values from C bindings to Crystal `Time`.

### DIFF
--- a/src/crystal/system/win32/file_info.cr
+++ b/src/crystal/system/win32/file_info.cr
@@ -71,7 +71,7 @@ struct Crystal::System::FileInfo < ::File::Info
   end
 
   def modification_time : ::Time
-    Time.from_filetime(@file_attributes.ftLastWriteTime)
+    ::Time.new(@file_attributes.ftLastWriteTime)
   end
 
   def owner : UInt32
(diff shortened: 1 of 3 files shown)


RX14:
It's not worth exposing platform-specific methods to the user for something so trivial. `LibC::Timespec` also exists on Windows and C libraries on windows could easily use that. But binding timespec on windows just for one constructor we don't even use is unlikely to get merged.

Since it's a confusing situation, we should delete the platform-specific constructor or at least make it nodoc

asterite:
They were previously exposed, then hidden because of the reasons @RX14 mentions, so we shouldn't bring them back.

straight-shoota:
Converting a `FILETIME` to `Time` is far from trivia, see the implementation of [Crystal::System::Time.filetime_to_seconds_and_nanoseconds](https://github.com/crystal-lang/crystal/blob/698ebf688ccd4d5a5f32b96b070365ad7db0c269/src/crystal/system/win32/time.cr#L24).
The stdlib needs to convert those values, so the implementation is available. It's also likely that users need to convert those values for custom C bindings. Why should you have to implement this conversion from a standard time struct to `Time` again?

RX14:
@straight-shoota programs can call `Crystal::System` themselves if they really want - it just makes it extra clear it's unstable.

straight-shoota:
Fine.

PR #6982, opened 2018-10-23
Add --format option to docs command

### LINKED ISSUE #2772: Option for the doc command to generate JSON.
Refs rrrene/inch#38.
/cc @rrrene

mniak:
Fixes #2772 
https://github.com/crystal-lang/crystal/issues/2772#issuecomment-341139030

### DIFF
--- a/src/compiler/crystal/command/docs.cr
+++ b/src/compiler/crystal/command/docs.cr
@@ -4,7 +4,7 @@
 # is in `crystal/tools/doc/`
 
 class Crystal::Command
-  private VALID_OUTPUT_FORMATS = ["html", "json"]
+  private VALID_OUTPUT_FORMATS = %(html json)
 
   private def docs
     output_format = "html"
--- a/src/compiler/crystal/command/docs.cr
+++ b/src/compiler/crystal/command/docs.cr
@@ -23,7 +23,7 @@ class Crystal::Command
       opts.on("--output=DIR", "-o DIR", "Set the output directory (default: #{output_directory})") do |value|
         output_directory = value
       end
-      opts.on("-f FORMAT", "--format=FORMAT", "Set the output format [json, html] (default: #{output_format})") do |value|
+      opts.on("--format=FORMAT", "-f FORMAT", "Set the output format [#{VALID_OUTPUT_FORMATS.join(", ")}] (default: #{output_format})") do |value|
         if !VALID_OUTPUT_FORMATS.includes? value
           STDERR.puts "Invalid format '#{value}'"
           puts opts
--- a/src/compiler/crystal/command/docs.cr
+++ b/src/compiler/crystal/command/docs.cr
@@ -4,7 +4,7 @@
 # is in `crystal/tools/doc/`
 
 class Crystal::Command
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -28,7 +28,7 @@ class Crystal::Doc::Generator
     },
   }
 
-  def initialize(@program : Program, @included_dirs : Array(String), @output_dir : String, @canonical_base_url : String?)
+  def initialize(@program : Program, @included_dirs : Array(String), @output_dir : String, @output_format : String, sss @canonical_base_url : String?)
```
Sija: `sss`? ;)

mniak: Sorry, typo

src/compiler/crystal/command/docs.cr
```diff
@@ -4,7 +4,10 @@
 # is in `crystal/tools/doc/`
 
 class Crystal::Command
+  private VALID_OUTPUT_FORMATS = ["html", "json"]
```
Sija: Pedantic: `%w(html json)`

straight-shoota: It's not really an improvement, just a different way to express this.

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -74,6 +72,25 @@ class Crystal::Doc::Generator
...
+  end
+
+  def generate_docs_html(program_type, types)
+    copy_files
+    generate_types_docs types, @output_dir, types
+    generate_readme program_type, types
+  end
+
+  def generate_readme(program_type, types)
+    raw_body, body = read_readme
```
straight-shoota: The part for generating `body` is only needed in this method, so I suppose it should be moved here from `read_readme`.

mniak:
Should I skip the generation of `index.json` when `--format=html` (or missing)?

straight-shoota:
That would require to change the JavaScript for loading the docs. Currently, it loads `index.json` when served through HTTP and `search-index.js` when opened as a local file (using JSONP). But it's totally fine to use `search-index.js` in both cases.

https://github.com/crystal-lang/crystal/blob/7b8326facbf6ff07b88c8295d749c16776c46545/src/compiler/crystal/tools/doc/html/js/_search.js#L528

You can tackle it right now or leave that for a new PR.

mniak:
I'd prefer to do handle it in a different PR

Sija:
@mniak It would be good to add "to docs command" in the title since it's not clear now what's this option is about.

Also please change "Solves the issue" to "Fixes" or "Closes" since these two are the keywords used by GH to auto-close issues.

mniak:
@Sija thanks for these tips

mniak:
@bcardiff I tried to follow your tip avoiding duplicate the logic to grab the content of the readme file.

RX14:
@bcardiff this is a pretty simple PR to review :)

RX14:
Merging this broke master :(

bcardiff:
A conflict with the specs added during #7063

PR #6985, opened 2018-10-23
Make ARGF's ARGV modifiable

r00ster91:
This adds a `property` to `IO::ARGF` which allows to modify `ARGF`'s `argv`. That's useful if you e.g. want to remove values from the array that aren't filenames. Or if you just want to take the first argument.

### DIFF
--- a/src/io/argf.cr
+++ b/src/io/argf.cr
@@ -3,6 +3,8 @@ class IO::ARGF < IO
   @path : String?
   @current_io : IO?
 
+  property argv : Array(String)
+
   def initialize(@argv : Array(String), @stdin : IO)
     @path = nil
     @current_io = nil


asterite:
`ARGF`'s `argv` **is** `ARGV`, so modifying `ARGV` should do what you want. No need to add an extra property. This is the same as in Ruby.

r00ster91:
But `ARGV` is a constant. Trying to modify `ARGV` results in `already initialized constant ARGV`.

Sija:
@r00ster91 Modyfing (`#clear`, etc.), not reassigning.

r00ster91:
Wow that's suprising me... I always thought it's completely impossible to modify constants at all and that you can only read from them.

asterite:
You can't reassign constants. You can modify the data they hold.

PR #6988, opened 2018-10-24
Replace HTTP::Server::Response#respond_with_error with #respond_with_status

straight-shoota:
`Response#respond_with_error` would previously just write the error header and flush the IO.
That flush seems unnecessary. Instead the response should be closed immediately. A call to this method was typically followed by a call tall to `close` anyway (which is no longer needed with this PR)

The only benefit of not closing directly is that you could append to the body (for example to provide more error data). That's a pretty irrelevant use case however. `#respond_with_error` already writes a body and you can't replace it, only append. Hence custom error pages will need to a custom implementation anyway.

This also changes the error response to a fixed content length.

/cc @jhass

### DIFF
--- a/src/http/server/request_processor.cr
+++ b/src/http/server/request_processor.cr
@@ -35,7 +35,7 @@ class HTTP::Server::RequestProcessor
         break unless request
 
         if request.is_a?(HTTP::Status)
-          response.respond_with_error(request.description, request.value)
+          response.respond_with_status(request)
           return
         end
 
@@ -47,7 +47,7 @@ class HTTP::Server::RequestProcessor
         begin
           @handler.call(context)
         rescue ex
-          response.respond_with_error
+          response.respond_with_status(:internal_server_error)
           error.puts "Unhandled exception on HTTP::Handler"
           ex.inspect_with_backtrace(error)
           return
--- a/src/http/server/request_processor.cr
+++ b/src/http/server/request_processor.cr
@@ -36,7 +36,6 @@ class HTTP::Server::RequestProcessor
 
         if request.is_a?(HTTP::Status)
           response.respond_with_error(request.description, request.value)
-          response.close
           return
         end
 
@@ -49,7 +48,6 @@ class HTTP::Server::RequestProcessor
           @handler.call(context)
         rescue ex
           response.respond_with_error
-          response.close
           error.puts "Unhandled exception on HTTP::Handler"
           ex.inspect_with_backtrace(error)
           return
--- a/src/http/server/response.cr
+++ b/src/http/server/response.cr
@@ -125,6 +125,13 @@ class HTTP::Server
 
     # Sends an error response.
     #
+    # Calls `#reset`, writes the given status, and closes the response.
+    def respond_with_error(status : HTTP::Status)
+      respond_with_error(status.description, status.code)
+    end
+
+    # Sends an error response.
+    #
     # Calls `#reset`, writes the given message, and closes the response.
     def respond_with_error(message = "Internal Server Error", code = 500)
       reset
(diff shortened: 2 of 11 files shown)


ysbaddaden:
It buries something that used to be explicit in the request handler. Sending an error response and deciding to close the connection are different matters.

straight-shoota:
It doesn't close the connection, just the response. `#responde_with_error` sends a complete response, so it should be closed.

straight-shoota:
It could be considered to rename the method to `#close_with_error` to make the behaviour more explicit.

straight-shoota:
Rebased onto master.

src/http/server/response.cr
```diff
@@ -132,7 +132,7 @@ class HTTP::Server
       message ||= @status.description
       self.content_type = "text/plain"
       self << @status.code << ' ' << message << '\n'
-      flush
+      close
```
asterite: Maybe clarify in the docs that this also closes the response?

straight-shoota:
I added the closing behaviour to the method documentation.

While at it, I also added an overload accepting `HTTP::Status`, which is a nice convenience.

src/http/server/response.cr
```diff
@@ -123,16 +123,23 @@ class HTTP::Server
       @output.closed?
     end
 
-    # Generates an error response using *message* and *code*.
+    # Sends an error response.
     #
-    # Calls `reset` and then writes the given message.
+    # Calls `#reset`, writes the given status, and closes the response.
+    def respond_with_error(status : HTTP::Status)
+      respond_with_error(status.description, status.code)
```
Sija: Minor optimization: I'd use this method as the _main_ impl, since this way you could avoid allocating `HTTP::Status` twice.

straight-shoota: I thought about that, but `HTTP::Status` is an enum, so it's actually not an allocation, just a cast and some validation. In order to provide a custom error message, this would need an additional `message` argument as well. But this should be the second (optional) argument, which would be inverse to the existing method.
However, that might really be the better API.
Also, I figured this method should perhaps be renamed. After all, it is not only useful for errors, but can send any kind of status code.
So my suggestion would be to change the API to this:

```cr
def respond_with_status(status : HTTP::Status, message : String? = nil)
  message || = status.description
  # impl
end

def respond_with_status(status : Int, message : String? = nil)
  respond_with_status(HTTP::Status.new(status), message)
end
```

Sija: Sounds like a nice improvement!

src/http/server/response.cr
```diff
@@ -126,22 +126,32 @@ class HTTP::Server
...
       reset
-      @status = HTTP::Status.new(code)
       message ||= @status.description
       self.content_type = "text/plain"
       self << @status.code << ' ' << message << '\n'
       close
     end
 
+    # :ditto:
+    def respond_with_status(status : Int = 500, message : String? = nil)
```
Sija: *status* argument IMO should be required.

straight-shoota: Yeah, I guess that's better. Before, it was `respond_with_error` where one could argue that a default error status is 500. But even that's not necessarily the case. So, yeah, I'll make it required.

src/http/server/request_processor.cr
```diff
@@ -26,7 +26,6 @@ class HTTP::Server::RequestProcessor
 
         if request.is_a?(HTTP::Request::BadRequest)
           response.respond_with_error("Bad Request", 400)
```
Sija: This line should make use of the newly added method I guess, ditto below.

straight-shoota:
Rebased and squashed. Also added the last commit which uses `#respond_with_status` in more places.

straight-shoota:
Rebased again on master with #8032 to see test_linux32 succeed.

straight-shoota:
... aaand rebased again on master with #8056

PR #6990, opened 2018-10-25
Fiber context & Stack pool

ysbaddaden:
Here are my last refactors to Fiber (before MT experiments):

1. Extract a Fiber::StackPool to hold & collect stacks that can be recycled, and made it thread safe;
2. Create a Fiber::Context struct that is passed to the arch-specific ASM;
3. Add a `resumable` property to the fiber context that tells whether the fiber is resumable (it's context is saved), running, or dead;
4. Since a fiber can be enqueued before its context is saved —which should never happen with a single thread, unless a concurrency primitive is wrong— the scheduler waits for the fiber to be resumable, and raises if the fiber is dead.

Note: these changes shall impact the performance but hopefully not that much. With a single thread, mutexes should never be locked, and most checks will always succeed, unless something's wrong.


src/fiber/context.cr
```diff
@@ -0,0 +1,55 @@
+class Fiber
+  # :nodoc:
+  #
+  # The arch-specific make/swapcontext assembly relies on the Context struct and
+  # expects the following layout. Avoid moving the struct properties as it would
+  # require to update all the make/swapcontext implementations.
+  @[Extern]
+  struct Context
+    property stack_top : Void*
+    property resumable : LibC::Long
```
j8r: What about using an enum for `resumable`, like:
```crystal
property state : State
enum State
  Resumable 
  Running
  Dead
end
```

Heaven31415:
@ysbaddaden I got a question, it's a bit unrelated, but very important to consider for MT. Some libraries that are crucial for game and application development (like OpenGL) store thread local information which is used during their execution, thus forcing you to know on which thread you are currently running. 

Will there be a mechanism to lock execution of exactly one fiber to a specific thread (which will allow proper usage of those libraries)?

This problem already happened in golang. You can read about their solution [here](https://golang.org/pkg/runtime/#LockOSThread) and [here](https://github.com/golang/go/wiki/LockOSThread).

vladfaust:
I'm a simple man, I see `thread safe`, you got my thumb up

src/fiber/context.cr
ysbaddaden: Sadly, no, the value must be accessible from ASM, and the enumeration wouldn't do much.

waj:
Do you know how big is the impact within the same thread? Also, when is that spin wait actually meaningful? Why a running (and still runnable) fiber is being rescheduled into another thread instead of keep ot running on the same thread?

I’m probably missing details on the big picture that you have in mind for the MT scheduler and I’d like to understand more about it.

ysbaddaden:
I'll make more testing and can replace the spin wait with a simpler resumable / dead check, that prove an erroneous synchronization primitive such as a wrong assumption in channel, mutex or a manual resume somewhere. It's better to detect this than crash :)

Resumable and the spin lock in scheduler are because of MT with *stealing* schedulers where thread A enqueues the current fiber and thread B steals it and resumes it... before thread A saved it's context. Oops. Segfault.

ysbaddaden:
Even with a dispatch scheduler, that would push enqueued fiber to any thread, since there is a time window between the moment we enqueue the current fiber and its context is saved, there is a risk that another thread tries to resume the fiber before it's context is fully saved.

waj:
I see, but I wonder, if the current fiber is about to be resumed in another thread, then it's runnable (not waiting for IO, timer or channel data), so why reschedule it in a different thread in the first place instead of keep it running on the current one? Mm.. maybe there is a `Fiber.yield` call. Is there any other case?

bcardiff:
We could use a compile time flag to apply the proposed lock only when MT is enabled. This will still leave the fiber context refactor and the assembly code setting it, but that should really matter performance wise.

ysbaddaden:
I'll use this branch as a basis for MT schedulers.

TBH this resumable flag and the spin lock checks have been insignificant in muco. They always succeed on single thread, and provide incredible safety otherwise.

PR #6994, opened 2018-10-26
allow pointer/UInt subtraction

damaxwell:
Allow subtraction of a `Pointer` and an unsigned integer. 

 Previously
```crystal
a = Pointer(Int8).new(1234)
b = a - 1_u64
```
yielded

```
in src/pointer.cr:95: wrong number of arguments for 'UInt64#-' (given 0, expected 1)
Overloads are:
 - UInt64#-(other : Int8)
 - UInt64#-(other : Int16)
 - UInt64#-(other : Int32)
 - UInt64#-(other : Int64)
 - UInt64#-(other : Int128)
 - UInt64#-(other : UInt8)
 - UInt64#-(other : UInt16)
 - UInt64#-(other : UInt32)
 - UInt64#-(other : UInt64)
 - UInt64#-(other : UInt128)
 - UInt64#-(other : Float32)
 - UInt64#-(other : Float64)

    self + (-other)
```

### DIFF
--- a/spec/std/pointer_spec.cr
+++ b/spec/std/pointer_spec.cr
@@ -186,6 +186,24 @@ describe "Pointer" do
     Pointer(Int32).new(1234).address.should eq(1234)
   end
 
+  it "performs arithmetic with u64" do
+    p = Pointer(Int8).new(1234)
+    d = 4_u64
+    (p + d).address.should eq(1238)
+    (p - d).address.should eq(1230)
+
+    p = Pointer(Int8).new(UInt64::MAX)
+    d = UInt64::MAX - 1
+    (p - d).address.should eq(1)
+  end
+
+  it "performs arithmetic with u32" do
+    p = Pointer(Int8).new(1234)
+    d = 4_u32
+    (p + d).address.should eq(1238)
+    (p - d).address.should eq(1230)
+  end
+
   it "shuffles!" do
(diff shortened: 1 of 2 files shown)


RX14:
This conversaion can overflow...

The real issue is the `self + (-other)` hack instead of implementing `Pointer#-` as a primitive.

straight-shoota:
> This conversaion can overflow...

It can, but no computer system has anywhere near that amount of addressable memory space.

Still, a primitive would be better, obviously.

damaxwell:
Hmm.  Can you give a specific example where this computation gets the wrong answer?  Maybe I'm missing something....

RX14:
@damaxwell it won't. This code is currently correct. However, `to_i64` will probably throw on overflow in the next few releases.

There's no "pointer subtract" instruction in LLVM, the calculation has to be done this way. We'll use `unsafe_as` to convert the number types, bypassing the overflow check.

damaxwell:
@RX14, if you guys are likely to move to throwing on overflow, should I just add `UInt64` specific versions right now of `Pointer#+` and `Pointer#-` that use `unsafe_as`?  Then you don't need to remember to fix the casts in both of these methods the future.

RX14:
@damaxwell probably better to just leave a TODO.

PR #6995, opened 2018-10-26
fix failing SIGSEGV spec

damaxwell:
On linux the spec "seg fault reports SIGSEGV" is intermittently failing because the process is exiting with success.  I.e, it looks like the desired seg fault is not being generated.  I have not reproduced this, but presumably the hard-coded pointer address in the spec is intermittently in user space.

Fixed by dereferencing a null pointer instead, which should always generate a seg fault.


straight-shoota:
I've noticed that failing spec as well. Thanks for investigating!

RX14:
Probably lack of ASLR (boo)

PR #7003, opened 2018-10-28
Add SemanticVersion to API docs

Blacksmoke16:
* Adds `SemanticVersion` to `docs_main.cr` to include it in doc generation.
* Adds additional docs on the `SemanticVersion` class.


src/semantic_version.cr
```diff
@@ -47,21 +71,31 @@ class SemanticVersion
...
   end
 
+  # Contains additional pre-release metadata related to this semantic version
   struct Prerelease
+    # Parses a `Prerelease` from the given pre-release metadata string
+    #
+    # ```
+    # require "semantic_version"
+    #
+    # prerelease = SemanticVersion::Prerelease.parse("rc.1.3")
```
Sija: Usually you'd write `rc1` without a dot in between.

Blacksmoke16: Was going off of examples on https://semver.org/.  Seems they have it as like `2.0.0-rc.1` or `2.0.0-rc.2`

Sija:
Sorry, I was thinking about dash, like in the examples you've given.

src/semantic_version.cr
```diff
@@ -36,6 +54,12 @@ class SemanticVersion
                   end
   end
 
+  # Returns the string representation of this semantic version
+  #
+  # ```
```
r00ster91: There's `require "semantic_version"` missing here.

src/semantic_version.cr
```diff
@@ -47,21 +71,31 @@ class SemanticVersion
     end
   end
 
+  # :nodoc:
```
r00ster91: Why `:nodoc:`? The `<=>` comparison operator should be documented so users know it can be used here.
```suggestion
  # The comparison operator.
  #
  # ```
  # require "semantic_version"  
  # 
  # semver1 = SemanticVersion.new(1, 0, 0)
  # semver2 = SemanticVersion.new(2, 0, 0)
  # semver1 <=> semver2 # => -1
  # ```
```

src/semantic_version.cr
```diff
@@ -74,15 +108,26 @@ class SemanticVersion
...
+    # require "semantic_version"
+    #
+    # semver = SemanticVersion.parse("0.27.1-rc.1")
+    # semver.prerelease.to_s # => "rc.1"
+    # ```
     def to_s(io : IO)
       identifiers.join(".", io)
     end
 
+    # :nodoc:
```
r00ster91: Ditto

src/semantic_version.cr
```diff
@@ -23,6 +40,7 @@ class SemanticVersion
     end
   end
 
+  # Create a new `SemanticVersion` instance with the given major, minor, and patch versions
```
r00ster91:
```suggestion
  # Creates a new `SemanticVersion` instance with the given major, minor, and patch versions
  # and optionally build and pre-release metadata
  #
  # Raises `ArgumentError` if *prerelease* is invalid pre-release metadata.
```

src/semantic_version.cr
```diff
@@ -74,15 +108,26 @@ class SemanticVersion
       Prerelease.new identifiers
     end
 
+    # Array of identifiers that make up the pre-release metadata
     getter identifiers : Array(String | Int32)
 
+    # Create a new `Prerelease` instance with supplied array of identifiers
```
r00ster91: Use the third person: `Creates`.

src/semantic_version.cr
```diff
@@ -1,14 +1,31 @@
 # Conforms to Semantic Versioning 2.0.0
```
r00ster91: Maybe put a newline after this? I think currently it would be shown like this in the docs:

---

#### Conforms to Semantic Versioning 2.0.0 See [https://semver.org/](https://semver.org/) for more information.

src/semantic_version.cr
```diff
@@ -1,14 +1,31 @@
...
+  # The major version of this semantic version
   getter major : Int32
+
+  # The minor version of this semantic version
   getter minor : Int32
+
+  # The patch version of this semantic version
   getter patch : Int32
+
+  # The build number of this semantic version
```
r00ster91: It's not always just a number.
```suggestion
  # The build metadata of this semantic version
```

src/semantic_version.cr
```diff
@@ -1,14 +1,31 @@
...
   getter prerelease : Prerelease
 
+  # Parses a `SemanticVersion` from the given semantic version string
+  #
+  # ```
+  # require "semantic_version"
+  #
+  # semver = SemanticVersion.parse("2.61.4")
+  # semver # => #<SemanticVersion:0x55b3667c9e70 @major=2, @minor=61, @patch=4, ... >
+  # ```
```
r00ster91:
```suggestion
  # ```
  #
  # Raises `ArgumentError` if *str* is not a semantic version.
```

src/semantic_version.cr
straight-shoota: I don't think it matters either way.

src/semantic_version.cr
Blacksmoke16: Was thinking since its required by the `Comparable` that the user would look [there](file:///home/george/dev/git/crystal/docs/Comparable.html#%28other%3AT%29-instance-method).

But probably wouldn't hurt to add here as well.

src/semantic_version.cr
```diff
@@ -1,14 +1,34 @@
...
+  # The minor version of this semantic version
   getter minor : Int32
+
+  # The patch version of this semantic version
   getter patch : Int32
+
+  # The build metadata of this semantic version
   getter build : String?
+
+  # The pre-release metadata of this semantic version
```
RX14: semver.org calls this the pre-release version

src/semantic_version.cr
```diff
@@ -23,6 +43,10 @@ class SemanticVersion
     end
   end
 
+  # Creates a new `SemanticVersion` instance with the given major, minor, and patch versions
+  # and optionally build and pre-release metadata
```
r00ster91: Here and below are still some `pre-release metadata` instances.

Blacksmoke16:
Anything else need done on this?

Issue #6125, opened 2018-05-24
macros: Keep the location (file/line/column) of AST nodes


bew:
To get proper location in case of bad usage of macros we can use location pragma around a specific AST node, but it's very repetitive to do it each time, and so nobody does that (and also it's not really documented either).

Here is an example:
````cr
macro do_with_location(node)
  do_something(
    #<loc:{{ node.filename }},{{ node.line_number }},{{ node.column_number }}>{{ node }}
  )
end

do_with_location BadConst
````
This gives the proper error on `BadConst`:
````
Error in loc.cr:8: expanding macro

do_with_location BadConst
^~~~~~~~~~~~~~~~

in loc.cr:8: undefined constant BadConst

do_with_location BadConst
                 ^~~~~~~~
````

I think it's a good candidate for a macro method that does that when writing:
```cr
macro do_with_location(node)
  do_something {{ node.with_location }}
end
```

For stdlib usage I think this could make JSON.mapping error more clean when a type doesn't exist (instead of the huuuuge error https://carc.in/#/r/44p0 for example)

WDYT?

bew:
I did a simple implementation:
```diff
diff --git a/src/compiler/crystal/macros/methods.cr b/src/compiler/crystal/macros/methods.cr
index 724d27aaf..1b3ee42fd 100644
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -319,6 +319,12 @@ module Crystal
           column_number = end_location.try &.original_location.try &.column_number
           column_number ? NumberLiteral.new(column_number) : NilLiteral.new
         end
+      when "with_location"
+        interpret_argless_method("with_location", args) do
+          MacroId.new(String.build do |str|
+            to_s(str, emit_loc_pragma: true)
+          end)
+        end
       when "=="
         interpret_one_arg_method(method, args) do |arg|
           BoolLiteral.new(self == arg)
```
I can PR it if you want

asterite:
You explained what you want to do, but not why.

bew:
I agree the macro source is a bit too much, 'hiding' it under `--error-trace` seems a good idea!

But then what do you want to show the user when `--error-trace` is not activated?

If we want the nice error message as I showed, without the user telling us where the token is coming from, I fear this is impossible: The macro language as a 'templating'-like language makes token tracing impossible.
Please correct me if I'm wrong!

bew:
Any other opinions?

[...]

PR #3858, opened 2017-01-07
Fix Reference#dup of file private class

MakeNowJust:
For instance:

```crystal
private class Foo
end

Foo.new.dup
```

this code causes an error:

```
Error in foo.cr:4: instantiating 'Foo#dup()'

Foo.new.dup
        ^~~

in /.../src/reference.cr:40: expanding macro

    {% if @type.abstract? %}
    ^

in macro 'macro_XXX' /.../src/reference.cr:40, line 2:

   1.
>  2.       dup = Foo.allocate
   3.       dup.as(Void*).copy_from(self.as(Void*), instance_sizeof(Foo))
   4.       dup
   5.
```

This commit fixes it by wrapping `{{@type}}` expansion with `#<loc:>` pragma
of class defined location.

PR #5597, opened 2018-01-16
Compiler: add more locations

asterite:
Fixes #5593

This adds more missing locations.

In theory all nodes should have a location. I tried to fix most of them but many are still missing (but they can be done later). One easy way to check this is to puts a `pp node` in [here](https://github.com/crystal-lang/crystal/blob/80cbe6603f938fc2200b785bd3fb2edbd854a17d/src/compiler/crystal/codegen/debug.cr#L225) and slowly fix those cases (but I don't have time for that now).

PR #7008, opened 2018-10-30
Compiler: add locations to all expanded macro arguments

MakeNowJust:
This PR changes compiler to add locations to all expanded macro arguments if possible. It makes helpful error message on macro expansion.

For example the following code makes an error with message `undefined constant BadType`.

```crystal
class Foo
  getter foo : BadType
end
```

However, currently the compiler shows expanded macro code and reports expanded macro line as error location. It is confusing for Crystal beginners I think.

```console
$ crystal run foo.cr
Error in foo.cr:2: expanding macro

  getter foo : BadType
  ^~~~~~

in foo.cr:2: expanding macro

  getter foo : BadType
  ^

in macro 'getter' expanded macro: macro_4384842512:113, line 4:

   1.
   2.
   3.
>  4.             @foo : BadType
   5.
   6.             def foo : BadType
   7.               @foo
   8.             end
   9.
  10.
  11.
  12.

undefined constant BadType
```

After this PR, the compiler shows such a simple error:

```console
$ bin/crystal foo.cr
Error in foo.cr:2: expanding macro

  getter foo : BadType
  ^~~~~~

in foo.cr:2: undefined constant BadType

  getter foo : BadType
               ^~~~~~~

```

Great!

- - -

## Technical Note

I believe this PR is awesome and useful, however, some developers (especially compiler specialist @asterite) think it is unacceptable because emitting location pragmas more makes breaking change and they know it is unuseful in some cases. (ref. #3858)

Don't worry! This PR does not add emitting location pragmas actually. Instead of emitting location pragmas to expanded code, it saves a pair of expanded code position and location pragmas of its position on macro expansion (it is called as `invisible_loc_pragmas` in source code.), and the lexer processes this data on parsing expanded code. It is cleaner and better than embedding location pragmas.
(However we cannot remove location pragmas from language because it has use-case still, e.g. ECR)

- - -

@bew This PR solves #6125 indirectly. What do you think?

- - -

Thank you :smile:

### DIFF
--- a/src/compiler/crystal/macros/interpreter.cr
+++ b/src/compiler/crystal/macros/interpreter.cr
@@ -2,6 +2,7 @@ module Crystal
   class MacroInterpreter < Visitor
     getter last : ASTNode
     property free_vars : Hash(String, TypeVar)?
+    property invisible_loc_pragmas : Hash(Int32, Array(Lexer::LocPragma))
 
     def self.new(program, scope : Type, path_lookup : Type, a_macro : Macro, call, a_def : Def? = nil, in_macro = false)
       vars = {} of String => ASTNode
@@ -78,6 +79,7 @@ module Crystal
                    @vars = {} of String => ASTNode, @block : Block? = nil, @def : Def? = nil,
                    @in_macro = false)
       @str = IO::Memory.new(512) # Can't be String::Builder because of `{{debug}}`
+      @invisible_loc_pragmas = {} of Int32 => Array(Lexer::LocPragma)
       @last = Nop.new
     end
 
@@ -99,13 +101,16 @@ module Crystal
       node.exp.accept self
 
       if node.output?
-        # In the caseof {{yield}}, we want to paste the block's body
-        # retaining the original node's location, so error messages
-        # are shown in the block instead of in the generated macro source
         is_yield = node.exp.is_a?(Yield) && !@last.is_a?(Nop)
-        @str << " #<loc:push>begin " if is_yield
-        @last.to_s(@str, emit_loc_pragma: is_yield, emit_doc: is_yield)
-        @str << " end#<loc:pop> " if is_yield
+        if (loc = @last.location) && loc.filename.is_a?(String) || is_yield
+          (@invisible_loc_pragmas[@str.pos.to_i32] ||= [] of Lexer::LocPragma) << Lexer::LocStackPragma::Push
+          @str << "begin " if is_yield
+          @last.to_s(@str, emit_loc_pragma: @invisible_loc_pragmas, emit_doc: true)
+          @str << " end" if is_yield
+          (@invisible_loc_pragmas[@str.pos.to_i32] ||= [] of Lexer::LocPragma) << Lexer::LocStackPragma::Pop
+        else
+          @last.to_s(@str)
+        end
       end
 
       false
@@ -161,7 +166,7 @@ module Crystal
             @last.to_s(str)
           end
         end
-      end).at(node)
+      end)
       false
     end
 
@@ -512,12 +517,12 @@ module Crystal
     end
 
     def visit(node : TupleLiteral)
-      @last = TupleLiteral.map(node.elements) { |element| accept element }.at(node)
+      @last = TupleLiteral.map(node.elements) { |element| accept element }
       false
     end
 
     def visit(node : ArrayLiteral)
-      @last = ArrayLiteral.map(node.elements) { |element| accept element }.at(node)
+      @last = ArrayLiteral.map(node.elements) { |element| accept element }
       false
     end
 
@@ -525,7 +530,7 @@ module Crystal
       @last =
         HashLiteral.new(node.entries.map do |entry|
           HashLiteral::Entry.new(accept(entry.key), accept(entry.value))
-        end).at(node)
+        end)
       false
     end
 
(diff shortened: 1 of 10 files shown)


straight-shoota:
For your example, this is great because the failing `BadType` is both in the original source as well as the expanded macro.

But what if the error was introduced only in the macro, with no direct equivalent in the original code? Let's just use the same code and error for example, but a custom macro for defining it:
```cr
macro define_foo
  @foo : Bad type
  def foo : Bad type
    @foo 
  end
end

class Foo
  define_foo
end
```

What error would this produce?

MakeNowJust:
@straight-shoota no `Bad type`, just `BadType`?

```
Error in foo.cr:9: expanding macro

  define_foo
  ^~~~~~~~~~

in foo.cr:9: expanding macro

  define_foo
  ^

in macro 'define_foo' foo.cr:1, line 1:

>  1.   @foo : BadType
   2.   def foo : BadType
   3.     @foo
   4.   end

undefined constant BadType
```

It is same as current.

bew:
@MakeNowJust I read the title of the PR, I knew exactly what was going on, as I've been thinking about doing exactly that (but by naively using loc pragma^^)

I'm glad to see that you took the time and did it :D
I'll read that later

src/compiler/crystal/macros/interpreter.cr
```diff
@@ -512,33 +517,33 @@ module Crystal
     end
 
     def visit(node : TupleLiteral)
-      @last = TupleLiteral.map(node.elements) { |element| accept element }.at(node)
+      @last = TupleLiteral.map(node.elements) { |element| accept element }
```
bew: Why remove the location of these new nodes? (here and below)

MakeNowJust: For preventing to report macro location as error location. However, I forgot they are used to report macro interpreter error, so this is wrong.

asterite:
Hey @MakeNowJust ! I think this is a great solution. I found it hard to understand at first, but it makes total sense. I solves very well one of the problems of text macros, basically that you loose the reference to the original node locations. I guess the performance impact is low because pragmas will only be checked in the lexer when lexing a macro expansion.

I'm not sure about the `invisible` term. Maybe we could call them "macro expansion pragmas" to make it clear that these result from macro expansion.

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -26,6 +26,33 @@ module Crystal
     # Heredocs pushed when found. Should be processed when encountering a newline
     getter heredocs = [] of {Token::DelimiterState, HeredocItem}
 
+    property invisible_loc_pragmas : Hash(Int32, Array(LocPragma))? = nil
+
+    alias LocPragma = LocStackPragma | LocSetPragma
+
+    enum LocStackPragma
```
asterite: What do you think if we separate this into two types? It can be `LocPushPragma` and `LocPopPragma`. Maybe the code will be even smaller. And the `case` bellow will be "complete" (right now there's a hidden `else`)

src/compiler/crystal/macros/interpreter.cr
bew: > For preventing to report macro location as error location

sry but I still don't get it, could you give an example?

bew:
@MakeNowJust will the example at https://github.com/crystal-lang/crystal/issues/6125#issuecomment-423640385 work with this PR? Or does it need sth else?

src/compiler/crystal/macros/interpreter.cr
MakeNowJust: @bew You have this code:

```crystal
macro foo
  {% x = 1 %}
  {{ x }} + "foo"
end

foo
```

And you will get such a error with bad location when evaluated macro node has its location:

```console
$ bin/crystal run foo.cr
Error in foo.cr:6: expanding macro

foo
^~~

in foo.cr:2: no overload matches 'Int32#+' with type String
Overloads are:
 - Int32#+(other : Int8)
 - Int32#+(other : Int16)
 - Int32#+(other : Int32)
 - Int32#+(other : Int64)
 - Int32#+(other : Int128)
 - Int32#+(other : UInt8)
 - Int32#+(other : UInt16)
 - Int32#+(other : UInt32)
 - Int32#+(other : UInt64)
 - Int32#+(other : UInt128)
 - Int32#+(other : Float32)
 - Int32#+(other : Float64)
 - Number#+()

  {% x = 1 %}
    ^

```

I think now it is wrong that evaluated macro node has location.

- - -

@asterite You added locations to evaluated macro node in #5597, and you said:

> In theory all nodes should have a location.

However, what is **theory**? I'm wondering when these locations are used and useful.

I think it is creepy that evaluated macro node has location because it is just a value except passed arguments. And I think these locations are never used.

MakeNowJust:
@bew No, unfortunately... However I think it is another problem from this PR. The reason what https://github.com/crystal-lang/crystal/issues/6125#issuecomment-423640385 does not work is for re-raising an error with node calling macro location [here](https://github.com/crystal-lang/crystal/blob/43010c806cab99290b9361790159268fa3761c76/src/compiler/crystal/semantic/semantic_visitor.cr#L417-L418):

```crystal
  def eval_macro(node)
    yield
  rescue ex : MacroRaiseException
    node.raise ex.message, exception_type: MacroRaiseException
  rescue ex : Crystal::Exception
    node.raise "expanding macro", ex
  end
```

I don't know why this is needed. Sorry.

MakeNowJust:
ping. Who is interested in this?

PR #7011, opened 2018-10-31
Outdated version alert in docs

Willamin:
When I'm looking up Crystal docs, my search – either on a search engine or my browser history – often results in an old version of the docs (eg. `/api/0.24/Array.html`). I thought it would be a nice convenience to have a link on every docs page that points to the latest version (eg.`/api/latest/Array.html`). 

This PR adds such a link to the sidebar. When Javascript is disabled, the link indicates that the user "might" be looking at old docs. When Javascript is enabled, either the link indicates that the user "is" looking at old docs or the link is hidden. The way I've done this isn't the cleanest way, but I think it's the only way it will work without manual work or a change in the docs deployment process. 

On page load, the current version of the docs and the latest version of the docs are compared to determine if the link should be displayed. The current version is obtained by looking at the URL of the current page. The latest version is obtained by requesting `/api/latest` and checking the resulting URL.

Here's a screenshot of where I added the link:
image

To test that the links worked well locally I rigged up a preview web application that handles things like redirecting `/api/latest` to `/api/0.26.1`. If anyone wants to use that to test this PR out locally, you can use my (docs-preview app)[https://github.com/willamin/docs-preview]

I think it would also be interesting to evolve this feature into one where the version can be selected with a dropdown, but this seemed like a good first step at least.

### DIFF
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -87,6 +87,8 @@ class Crystal::Doc::Generator
 
     File.write File.join(@output_dir, "css", "style.css"), StyleTemplate.new
     File.write File.join(@output_dir, "js", "doc.js"), JsTypeTemplate.new
+
+    File.write File.join(@output_dir, "js", "version-checker.js"), JsVersionCheckerTemplate.new
   end
 
   def generate_types_docs(types, dir, all_types)
--- a/src/compiler/crystal/tools/doc/html/_head.html
+++ b/src/compiler/crystal/tools/doc/html/_head.html
@@ -13,6 +13,7 @@
 <link href="<%= base_path %>css/style.css" rel="stylesheet" type="text/css">
 
 <script type="text/javascript" src="<%= base_path %>js/doc.js"></script>
+<script type="text/javascript" src="<%= base_path %>js/version-checker.js"></script>
 <script type="text/javascript">
   CrystalDoc.base_path = "<%= base_path %>";
 </script>
--- a/src/compiler/crystal/tools/doc/html/_sidebar.html
+++ b/src/compiler/crystal/tools/doc/html/_sidebar.html
@@ -1,10 +1,12 @@
 <div class="sidebar">
   <div class="sidebar-header">
 
-    <a href="/api/latest/">
-      <div class="alert">
-        This documentation may be out of date. Click here to visit the latest version.
-      </div>
+    <a id="outdated-alert" href="/api/latest/">
+      <noscript>
+        <div class="alert">
+          This documentation may be out of date. Click here to visit the latest version.
+        </div>
+      </noscript>
     </a>
 
     <div class="search-box">
--- a/src/compiler/crystal/tools/doc/html/_sidebar.html
+++ b/src/compiler/crystal/tools/doc/html/_sidebar.html
@@ -1,5 +1,12 @@
 <div class="sidebar">
   <div class="sidebar-header">
+
+    <a href="/api/latest/">
+      <div class="alert">
+        This documentation may be out of date. Click here to visit the latest version.
+      </div>
+    </a>
+
     <div class="search-box">
       <input type="search" class="search-input" placeholder="Search..." spellcheck="false" aria-label="Search">
     </div>
--- a/src/compiler/crystal/tools/doc/html/_sidebar.html
+++ b/src/compiler/crystal/tools/doc/html/_sidebar.html
@@ -1,7 +1,7 @@
 <div class="sidebar">
   <div class="sidebar-header">
 
-    <a id="outdated-alert" href="/api/latest/">
+    <a id="outdated-alert" href="/api/latest/<%= current_type.try(&.path)%>">
       <noscript>
         <div class="alert">
           This documentation may be out of date. Click here to visit the latest version.
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -624,3 +624,17 @@ span.flag.purple {
   margin-left: -100px;
   margin-right: 12px;
 }
+.alert {
+  padding: 0.5rem;
+  margin: 0.5rem;
+  background-color: hsl(270, 25%, 80%);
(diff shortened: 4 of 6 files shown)


straight-shoota:
The initiative is great, this is definitely a point to be improved.

But I don't like this approach. It assumes properties about the environment (for example that API docs for different versions are served at `/api/<version>` and `/api/latest` points to the most recent release). This might reflect the current deployment of the stdlib API docs, but it certainly doesn't fit for any Crystal library's API docs.

As you've already mentioned, it would even be better to link to all available API versions directly on each page. This feature needs a different approach than the one in this PR anyway and would easily solve the issue to direct readers to the most recent version. That's where we should be heading, not this intermediary solution which would be insufficient in the long term.

RX14:
Ultimately, managing multiple versions of crystal docs is out of scope for the crystal doc tool. What is in scope, is an option to embed some "header" markdown, where the header markdown could be generated by some tool or script with domain-specific knowledge of what to write in the header. (`--canonical-base-url` should also go, being replaced by a generic option to include tags in the HTML `<header>`)

Unfortunately this would require regenerating the previous documentation versions every time the header was changed. Can we find a way around this that doesn't require using JS?

asterite:
I actually see no problem in having a standard way to check for the latest docs and show that alert. Having a convention around that is good. Maybe `/api/latest` is not good, or maybe yes for the crystal doc, but we just need a way to specify where (if any) the latest URL can be found, and that's it.

Willamin:
>  the docs generator is not only designed for the standard library

I totally agree, and think it would be good to determine a standard way to archive and access old documentation for all Crystal libraries. I did notice a `@is_crystal_repo` variable in `generator.cr` that could be used to limit this to only the stdlib at first. I don't like that approach but it's an option. 

> Unfortunately this would require regenerating the previous documentation versions every time the header was changed. Can we find a way around this that doesn't require using JS?

We could use an iframe that loads a small page containing a dropdown or other control that is generated every time new docs are added. It's not the cleanest solution.

> Maybe `/api/latest` is not good

I didn't want to go the route of checking to see what `/api/latest` points to, but it seemed like it would work well with the current setup. I think a really clean way to handle it would be to have a JSON api (could be a static file that gets updated with new docs) that lists the available versions and the latest version. 

> we just need a way to specify where (if any) the latest URL can be found

I think this is a good option. How about some logic like this?
- if it's the crystal repo, use the default `/api/:version` and `/api/latest`
- if it's another library, default to not including the version warning
- if it's another library, allow opting into including the version warning, defaulting to `/api/:version` and `/api/latest` via `$ crystal docs` CLI flags
- allow specifying alternatives to `/api/:version` and `/api/latest` via `$ crystal docs` CLI flags

---

If any of these ideas sound like a way to improve this PR, I'll go ahead and put them in code. I wanted to go ahead and get at least a simple solution merged in as quickly as possible so it can exist in the oldest possible release so it can start paying off. That was my main goal with the simple "go to the latest version" link.

RX14:
Ah yes, if we allow a custom HTML for the header, the custom HTML can just be an iframe. I like that solution.

Willamin:
> I don't want to accept any solution that depends on JS.

Agreed. With the current state of this PR, I tried to make it work as best as it could without JS. I'll look into the iframe solution to see what I can get working.

straight-shoota:
I'm sure we agree to rule out any solution that would require touching every single HTML file of every previous release on a new release.

Embedding an HTML page as an iframe could help, but its capabilities are very limited and it causes usability issues. It can only show static content and is not context-sensitive. If we want to provide *useful* information how the currently displayed version relates to other available versions, we'd still need to maintain one HTML file for every previous release.

This really looks like a very natural use case for Javascript. I don't think there is a practical solution without using a simple Javascript tool. And I don't see why that should be avoided at all costs. The API docs already use a fair amount of Javascript for the search and other features. So it's just adding one more feature which actually serves a quite important use case.

Looking for ideas, I found how elixir's docs generator does this and I like this implementation. The relevant documentation for ex_doc can be found here: https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html#module-additional-javascript-config

It works like this:

* When generating docs, you can pass a configuration option `javascript_config_path`. This is a URL pointing to a javascript file which provides information about available versions. The URL could either be absolute when the final deploy location is fixed, or could also be relative to the API docs root location (for example `../versions.js`).
* This `versions.js` file declares a list of available versions, each with version name and URL. Again, these URLs could be absolute or relative to the API docs' root location. Example file from ex_doc: https://hexdocs.pm/ex_doc/docs_config.js Only this file needs to be updated on every new release.
* A tiny Javascript helper inserts a version select dropdown element into the page content based on the values provided by `versions.js`. The implementation is very minimal, as the example of ex_doc shows: https://github.com/elixir-lang/ex_doc/blob/18fe1ad7c966708b947fc274626d5d7ef5513f97/assets/js/versions.js There's a [handlebars template](https://github.com/elixir-lang/ex_doc/blob/18fe1ad7c966708b947fc274626d5d7ef5513f97/assets/js/templates/versions-dropdown.handlebars) etc. but this would all be trivial to implement in plain JavaScript.

RX14:
The docs currently work fine without JS, and I want to support people who choose to use the web without JS. Many are close friends.

You can say that the old docs version warning isn't a neccesary feature for people not using JS but I think it would be nice if at all possible to support this. I guess it does make it a lot harder though.

straight-shoota:
There's always the possibility of a fallback solution via `<noscript>`. That could just be a note that the version data is not available and a link to a version overview page (which we should have in any case).

KimBurgess:
An alternative solution here, that keeps everything reaching the user as static HTML would be to use a [Cloudflare worker](https://developers.cloudflare.com/workers/reference/apis/html-rewriter/) (or similar) to insert this based on the path at the time of the request.

straight-shoota:
@KimBurgess I don't think that's an option because it adds unnecessary complexity to deploying docs with the version navigation feature. That should just be pushing some static html files somewhere.
A more feasible solution would be to pull the files from old versions, insert the version data and re-upload. Maybe that could be worth exploring, but for now a JavaScript based solution seems pretty good and should cover most use cases with minimal effort.

Issue #6905, opened 2018-10-03
Add IO#read_string? and #read_bytes?


straight-shoota:
As suggested in https://github.com/crystal-lang/crystal/pull/6873#issuecomment-424730829, the IO methods `read_string` and `read_bytes` don't have a non-raising counterpart which would be useful in situations where it's optional if the IO can provide an additional value.

See https://github.com/crystal-lang/crystal/pull/6501#discussion_r208516591 for another example where `read_bytes?` would be useful but a workaround needed to be employed.

Implementing `#read_string?` should be pretty easy. For `#read_bytes?`, we'd probably also need `IO::ByteFormat#decode?` or let `IO::ByteFormat#decode` return `nil` instead of raising. But then we'd need to ensure a proper exception is raised.

asterite:
I don't think these methods should exist. What if read_bytes? reads a few bytes but not enough to decode the data? How can you recover those lost bytes?

straight-shoota:
`read_bytes?` would be for use cases like https://forum.crystal-lang.org/t/read-nrrd-file-with-attached-header-and-gzipped-data/5355 where you know from the data format description that you can either read the full bytes or not.
Anything in between would be a failure state and should probably raise. Excess bytes could be attached to the exception.

HertzDevil:
> What if read_bytes? reads a few bytes but not enough to decode the data? How can you recover those lost bytes?

In C you would `ungetc` the bytes already read. However I guess it isn't very appropriate to implement a similar in-memory buffer for every `IO` in Crystal

asterite:
In the end I'm thinking that `read_bytes` can also fail in the same way, and right now there's no way to recover the lost bytes. So it's probably fine to add `read_bytes?` However, we must clearly document that if `nil` is returned it might be because there wasn't enough data to read, but some of it might have been read (IOs that have `peek` could sometimes avoid advancing the pointer)

PR #7014, opened 2018-11-01
Add IO#read_string?

straight-shoota:
Fixes part of #6905

### DIFF
--- a/spec/std/io/io_spec.cr
+++ b/spec/std/io/io_spec.cr
@@ -323,7 +323,7 @@ describe IO do
       io.read_char.should be_nil
     end
 
-    it "reads string" do
+    it "#read_string" do
       io = SimpleIOMemory.new("hello world")
       io.read_string(5).should eq("hello")
       io.read_string(1).should eq(" ")
@@ -331,6 +331,14 @@ describe IO do
       expect_raises(IO::EOFError) do
         io.read_string(6)
       end
+    end
+
+    it "#read_string?" do
+      io = SimpleIOMemory.new("hello world")
+      io.read_string?(5).should eq("hello")
+      io.read_string?(1).should eq(" ")
+      io.read_string?(0).should eq("")
+      io.read_string?(6).should be_nil
     end
 
     it "does each_line" do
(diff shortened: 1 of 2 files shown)


spec/std/io/io_spec.cr
```diff
@@ -333,6 +333,14 @@ describe IO do
       end
     end
 
+    it "#read_string?" do
+      io = SimpleIOMemory.new("hello world")
+      io.read_string?(5).should eq("hello")
+      io.read_string?(1).should eq(" ")
+      io.read_string?(0).should eq("")
```
luislavena: This scenario is not described in the documentation. I know originally isn't described for `#read_string` either, but might be worth adding it.

Thank you.

src/io.cr
```diff
@@ -443,14 +443,30 @@ abstract class IO
...
     String.new(bytesize) do |ptr|
       if decoder = decoder()
         read = decoder.read_utf8(self, Slice.new(ptr, bytesize))
         if read != bytesize
-          raise IO::EOFError.new
+          return
         end
       else
-        read_fully(Slice.new(ptr, bytesize))
+        unless read_fully?(Slice.new(ptr, bytesize))
```
j8r: Why you don't put this in an `elif`?

jhass:
Mh, sorry to derail a bit, but this made me realize it makes it easy to read a broken string if your bytelimit happens to be in the middle of a multi-byte character (whatever multi-byte encoding your IO is in really). I wonder if these should take the amount of codepoints/characters to read instead?

asterite:
I'm actually thinking this method shouldn't exist. The problem is that it can happen that you read some bytes but not enough to reach the desired bytesize, and at that point you get `nil` as a return value but you lost those read bytes.

With `read_fully?` at least you get the bytes on the passed slice so you don't lose them.

Or, well, at least I would document that this method might read bytes that will be lost, and most probably the caller should raise an error on `nil`.

RX14:
I also want `gets(limit)` without stopping at any delimiters. gets's behaviour is to only return nil if the current position is EOF, it returns strings whose bytesize is less than the limit. So ideally there should be both behaviours. The behaviour in this PR seems more like `read_string_fully` to me, and we need a `read_string`.

straight-shoota:
Closing stale PR.

PR #7022, opened 2018-11-03
Fix hovering methods not working correctly

### LINKED ISSUE #7020: HTML docs: hovering on an inherited method shows in the wrong place
https://crystal-lang.org/api/0.27.0/OpenSSL/SSL/Socket/Server.html

Scroll slightly down and hover on an inherited method.

image

The amount you scrolled by will be the amount that the label is off by.

ghost:
Fixes #7020 

#6764 probably removed a bit too much and caused that when hovering a method in the inherited section, the detailed representation of the method is shown at the wrong place.

This is fixed by using `position: relative` which was originally applied by `filter: blur(0);`.

grafik

### DIFF
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -554,7 +554,7 @@ span.flag.purple {
   display: flex;
 }
 .main-content {
-  filter: blur(0);
+  position: relative;
 }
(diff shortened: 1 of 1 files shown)


RX14:
This should use `position: relative` instead of `filter: blur(0)`. Using `filter` does apply `position: relative` (lol wtf??), but the former is better to solve this bug.

Issue #5351, opened 2017-12-05
Double splat args doesn't appear in generated doc for macros


bew:
```cr
# some doc
macro mapping(**properties)
end
```
Gives:
![macro_no_spec_in_doc](https://user-images.githubusercontent.com/9730330/33630662-76d3d32a-da08-11e7-9254-a1834fb9789a.png)
extracted from [YAML doc](https://crystal-lang.org/api/0.23.1/YAML.html#mapping-macro)

PR #7031, opened 2018-11-06
Fix `Crystal::Doc::Method#has_args?` to show double splat arg

MakeNowJust:
Fixed https://github.com/crystal-lang/crystal/issues/5351#issuecomment-435654252


RX14:
Love the specs!

PR #7040, opened 2018-11-07
Add type restriction to private Process constructor

### LINKED ISSUE #7038: Misleading error message on Process.new(nil)
This code: `Process.new(nil)` results in `private method 'new' called for Process.class`.

```

LLVM: 4.0.0
```

z64:
The unrestricted ctor argument causes the compiler to match a call like `Process.new(nil)` to this private ctor.

Restricting the argument to `@pid : Int32` gives a more helpful compiler error, explaining that no overload matches.

```cr
class Foo
  getter value : Int32
  
  # private def initialize(@value)
  private def initialize(@value : Int32)
  end
  
  def initialize
    @value = 0
  end
end

Foo.new(nil)

# Error in line 13: no overload matches 'Foo.new' with type Nil
# Overloads are:
#  - Foo.new(value : Int32)
#  - Foo.new()
```
https://carc.in/#/r/5gg1

Fixes #7038


bew:
BTW, the error suggesting that `Foo.new(value : Int32)` is a valid ctor is kinda wrong as it is private, and can't be used from the outside, no?

z64:
`Libc::PidT` sounds good to me, given there is `def self.pid : Libc::PidT`.

I just followed the [existing getter](https://github.com/crystal-lang/crystal/blob/3721ab8a6d3d2ea121cd62f4e395a1f6f4a7f785/src/process.cr#L220) :)
I will change them both if we like.

@Sija It is used legally in the public method `def self.fork`, there is a call to `new pid`

jhass:
Yeah updating both sounds good to me

RX14:
I'd rather not make the `Process#pid : LibC::SizeT` change, I don't want to expose platform-specific integer types.

jhass:
We already expose this one though, even expect it in some public API. Would you feel better about an alias at `Process::PidT` and using that for all public API?

RX14:
I'd feel better changing all the places where we expose this to not expose this, instead of going in the other direction.

RX14:
Note: I'm not suggesting we do that in this PR, we can just remove that part of the diff.

straight-shoota:
We can discuss not exposing platform-specific types in a separate issue. But the type of `Process#pid` is `LibC::PidT`, whether we declare it as an explicit return type or not. I think it's better to be explicit because it improves API documentation.

PR #7051, opened 2018-11-08
Refactor filename resolution logic out of `interpret_run`

Sija:
Spinoff from #6967 as per https://github.com/crystal-lang/crystal/pull/6967#discussion_r231622556

### DIFF
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -220,8 +220,7 @@ module Crystal
       node.args.first.accept self
       original_filename = @last.to_macro_id
 
-      filename = original_filename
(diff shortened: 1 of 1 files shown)


src/compiler/crystal/macros/methods.cr
```diff
@@ -185,37 +218,11 @@ module Crystal
...
-        end
-
-        if found_filenames.size > 1
-          node.raise "error executing macro run: '#{filename}' is a directory"
-        end
+      original_filename = @last.to_macro_id
 
-        filename = found_filenames.first
+      filename = original_filename
+      filename = find_source_file(filename) do |error_message|
```
RX14: `filename = find_source_file(original_filename) do`...?

PR #7053, opened 2018-11-09
Remove implicit null skip from Hash to JSON serialization

### LINKED ISSUE #7052: Hash(...)#from_json; NULL results in empty-hash.
Environment:

```
$ crystal -v
Crystal 0.27.0 (2018-11-04)

LLVM: 6.0.1
```

Bug:

```crystal
require "json"

hash = Hash(String, String | Nil).from_json %({"a":null})

puts hash # => {}
```

Bug-in-action:

https://play.crystal-lang.org/#/r/5gw4

MakeNowJust:
Fixed #7052 

I think `Hash(String, String).from_str(%{{"foo": null}})` should be failed because `String` does not contain `nil` and JSON `null` is just `nil`. In fact Rust (serde) and Swift does not allow this.

Rust:

```rust
#[macro_use]
extern crate serde_derive;

extern crate serde;
extern crate serde_json;

use std::collections::HashMap;

fn main() {
    let _: HashMap<String, String> = serde_json::from_str(r#"{"foo": null}"#).unwrap();
}
// Error("invalid type: null, expected a string", line: 1, column: 12)
```

Swift:

```swift
import Foundation

let decoder: JSONDecoder = JSONDecoder()
let _: [String: String] = try! decoder.decode(Dictionary<String, String>.self, from: """
  {"foo": null}
""".data(using: .utf8)!)
// Swift.DecodingError.valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [_DictionaryCodingKey(stringValue: "foo", intValue: nil)], debugDescription: "Expected String but found null value instead.", underlyingError: nil))
```

### DIFF
--- a/spec/std/json/serialization_spec.cr
+++ b/spec/std/json/serialization_spec.cr
@@ -45,10 +45,6 @@ describe "JSON serialization" do
       Hash(String, Int32).from_json(%({"foo": 1, "bar": 2})).should eq({"foo" => 1, "bar" => 2})
     end
 
-    it "does Hash(String, Int32)#from_json and skips null" do
-      Hash(String, Int32).from_json(%({"foo": 1, "bar": 2, "baz": null})).should eq({"foo" => 1, "bar" => 2})
-    end
-
     it "does for Array(Int32) from IO" do
       io = IO::Memory.new "[1, 2, 3]"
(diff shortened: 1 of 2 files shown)


MakeNowJust:
@straight-shoota Good! Added.

PR #7055, opened 2018-11-09
Fix RequestProcessor connection reuse with unconsumed requests

### LINKED ISSUE #6924: skip_to_end in HTTP::Content#close: Useful or problematic?
`HTTP::Content` provides a single method, `#close` that delegates to super (`IO#close`) but ensures the IO is completely consumed before closing via `skip_to_end`. 

I can't see how that is actually useful, and at least, it is confusing to the user. It doesn't just close the IO in it's current state but completely consumes it until the sender is done (a request from `curl -T /dev/urandom example.com`  would never actually close...).

This feature seems to be the only reason for the `HTTP::Content` type in the first place and was introduced in 5adc780f. Unfortunately, I can't find an explanation for the reasons behind that and there are no specs documenting a change in behaviour.

At the point of that commit, there was only a single `HTTP::Response` class and not yet separated between server and client. Not sure if that matters in any way.

I could only think of a possible explanation in the context of reusing a connection (HTTP keep-alive), where you would obviously need to completely consume the previously sent content before you could read the next request or response from the IO.
But that actually only matters in `RequestProcessor` and as far as I can see, wouldn't demand for an override of the close method. This would be it (and the special behaviour is even mentioned here):

```diff
--- i/src/http/server/request_processor.cr
+++ w/src/http/server/request_processor.cr
@@ -57,6 +57,7 @@ class HTTP::Server::RequestProcessor

         # Skip request body in case the handler
         # didn't read it all, for the next request
+        request.body.try &.skip_to_end
         request.body.try &.close
       end
     rescue ex : Errno
```
Without that, only `spec/std/http/server/server_spec.cr:579` would fail.

With this change, we could remove `HTTP::Content#closet`. However, it would still provide a point of attack when a client just keeps writing to the socket.

Some light improvement could be to provide a means for the handler to force closing the connectio (by checking the response has `Connection: keep-alive` and `Content-Length` headers or closing the request IO in the handler directly). There could also be some limit to the amount of bytes that will be read to avoid never-closing connections.

But all in all, I think it might perhaps be best to simply close the entire connection when there is still content on the input stream after the handler has completed. The handler should make sure to consume the body if there should be one.
The consequence is only that you can't reuse the connection if the client sent a body that the server didn't expect. I don't think that's too much to worry about.

/cc @waj

Initially mentioned in https://github.com/crystal-lang/crystal/pull/6912#issuecomment-428293598

straight-shoota:
Fixes #6924

This removes the `skip_to_end` call from `HTTP::Content#close`, which would involuntary consume a large (or infinite) HTTP body.

Instead, the logic is moved to `RequestProcessor` and it closes the connection if the body is too large (in this case `>16kb`).

I'm not entirely sure if it should skip any body at all automatically. I'm more leaning towards making this the responsibility of the handler. And the rule would just be if the handler doesn't entirely consume the request body, the connection can't be reused. It's probably fine to leave some leeway, though. That's why I included this into the PR. But we can easily remove it and break if the body contains any data at all.
I'll leave that up for discussion.

The handler can also tell the request processor to close the connection by setting `Connection: close` header.

### DIFF
--- a/src/http/server/request_processor.cr
+++ b/src/http/server/request_processor.cr
@@ -59,21 +59,17 @@ class HTTP::Server::RequestProcessor
         break unless HTTP.keep_alive?(response)
 
         # The request body is either FixedLengthContent or ChunkedContent.
-        # In case it has not entirely been consumed by the handler, try to
-        # skip to the end. If the request is larger than maxmum skippable size,
-        # we close the connection even if keep alive was requested.
+        # In case it has not entirely been consumed by the handler, the connection is
+        # closed the connection even if keep alive was requested.
         case body = request.body
         when FixedLengthContent
-          if body.read_remaining > 16_384
-            # Close the connection if remaining length exceeds the maximum skipable size.
+          if body.read_remaining > 0
+            # Close the connection if there are bytes remaining
             break
-          else
-            body.skip_to_end
           end
         when ChunkedContent
-          # Try to read maximum skipable number of bytes.
-          # Close the connection if the IO has still more to read.
-          break unless skip_to_end(body)
+          # Close the connection if the IO has still bytes to read.
+          break unless body.closed?
         end
       end
     rescue ex : Errno
@@ -85,17 +81,5 @@ class HTTP::Server::RequestProcessor
         # IO-related error, nothing to do
       end
     end
-  end
-
-  # Reads and discards bytes from `io` until there are no more bytes.
-  # If there are more than 16_384 bytes to be read from the IO, it returns `false`.
-  private def skip_to_end(io : IO) : Bool
-    buffer = uninitialized UInt8[4096]
-
-    4.times do
-      return true if io.read(buffer.to_slice) < 4096
-    end
-
-    false
   end
 end
--- a/src/http/server/request_processor.cr
+++ b/src/http/server/request_processor.cr
@@ -55,9 +55,26 @@ class HTTP::Server::RequestProcessor
 
         break unless request.keep_alive?
 
-        # Skip request body in case the handler
-        # didn't read it all, for the next request
-        request.body.try &.close
+        # Don't continue if the handler set `Connection` header to `close`
+        break unless HTTP.keep_alive?(response)
+
+        # The request body is either FixedLengthContent or ChunkedContent.
+        # In case it has not entirely been consumed by the handler, try to
+        # skip to the end. If the request is larger than maxmum skippable size,
+        # we close the connection even if keep alive was requested.
+        case body = request.body
+        when FixedLengthContent
+          if body.read_remaining > 16_384
+            # Close the connection if remaining length exceeds the maximum skipable size.
+            break
+          else
+            body.skip_to_end
+          end
+        when ChunkedContent
+          # Try to read maximum skipable number of bytes.
+          # Close the connection if the IO has still more to read.
+          break unless skip_to_end(body)
+        end
       end
     rescue ex : Errno
       # IO-related error, nothing to do
@@ -68,5 +85,17 @@ class HTTP::Server::RequestProcessor
         # IO-related error, nothing to do
       end
     end
+  end
+
+  # Reads and discards bytes from `io` until there are no more bytes.
+  # If there are more than 16_384 bytes to be read from the IO, it returns `false`.
+  private def skip_to_end(io : IO) : Bool
+    buffer = uninitialized UInt8[4096]
+
+    4.times do
+      return true if io.read(buffer.to_slice) < 4096
+    end
+
+    false
   end
 end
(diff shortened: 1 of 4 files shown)


asterite:
I didn't see this PR's diff yet, but I'm now thinking we should change `HTTP::Client` to be concurrent by default, kind of similar to what Go does.

I found [this article](https://awmanoj.github.io/tech/2016/12/16/keep-alive-http-requests-in-golang/) 
 [which mentions this stackoverflow question](https://stackoverflow.com/questions/17948827/reusing-http-connections-in-golang).

In short, it seems that when you create a client, do a request but you don't read the whole response and close it then keep-alive won't be used. But it still works: internally I guess it creates a new socket, connects and performs a separate request. If you then don't read and close the connection I guess it's closed after some time. So I'm pretty sure there's a connection pool going on.

We must do this change. Right now everything's fine if you use the http client in a single fiber, but if you use it in multiple fibers then there's a change that requests are intermixed and this should work by default, it's not good to have to wait for other responses to be read and closed.

src/http/server/request_processor.cr
```diff
@@ -69,4 +86,16 @@ class HTTP::Server::RequestProcessor
...
     end
   end
+
+  # Reads and discards bytes from `io` until there are no more bytes.
+  # If there are more than 16_384 bytes to be read from the IO, it returns `false`.
+  private def skip_to_end(io : IO) : Bool
+    buffer = uninitialized UInt8[4096]
+
+    4.times do
+      return true if io.read(buffer.to_slice) < 4096
```
asterite: `read` isn't guaranteed to read the full slice. You should use `read_fully?`

asterite:
Now I read the diff. The change seems fine, except that I'd like `HTTP::Client` to succeed performing a request even if another request is currently being processed (it's body was not yet consumed). I'm not sure we should merge this because it might break existing code... right? If you perform two consecutive requests and the response body is larger than the maximum skippable size.

straight-shoota:
@asterite It took my a while to understand why you're mentioning `HTTP::Client` so much. Might have been me being tired. But yes, this affects client side response consuming as well.

> I'm now thinking we should change HTTP::Client to be concurrent by default, kind of similar to what Go does.

That's #6011

We might consider using an auto-consuming IO wrapper for client response.
But I think I'd actually prefer to be strict in both server and client. If the request/response is not entirely consumed, the connection can't be reused. That transfers responsibility for deciding what should be consumed entirely to the user - instead of using some guesswork in the general implementation saying that maybe some amount of unconsumed data could be skipped.

asterite:
@straight-shoota Right. For me the logic should be:
- I make a request and get a response
- If I don't consume that response and make another request, that opens another connection (but still works)
- If I did consume the response and make another request, reuse the connection using keep-alive

The only problem here is having to eventually close that connection that never consumed the response to avoid leaks.

straight-shoota:
> having to eventually close that connection that never consumed the response to avoid leaks.

This could be implemented in `HTTP::Client::Response#close`. Users would need to call this method when finished.  If the request body was consumed, the connection is returned to the pool. Otherwise it is simply closed. Yielding methods would automatically take care of this.

src/http/server/request_processor.cr
straight-shoota: I don't think this part should be merged anyway. It's better to be strict and don't reuse the connection when there is even only one unconsumed byte.

straight-shoota:
I added a commit which removes skipping some minor amount of bytes and instead always closes the connection if the request was not entirely consumed.

src/http/server/request_processor.cr
```diff
@@ -55,9 +55,22 @@ class HTTP::Server::RequestProcessor
 
         break unless request.keep_alive?
 
-        # Skip request body in case the handler
-        # didn't read it all, for the next request
-        request.body.try &.close
+        # Don't continue if the handler set `Connection` header to `close`
```
RX14: The body still needs to be closed here.

straight-shoota:
Yes, the socket is always closed at the end of the loop in the `ensure` block. It's a bit convoluted, maybe there could be some more refactoring.

straight-shoota:
I'll also add docs about this to `HTTP::Server::Handler` when we're settled on the behaviour.

RX14:
I like the behaviour, personally.

straight-shoota:
I've added a section about reusing connections to the API docs of `HTTP::Server`. I figured this was the right place, because `RequestProcessor` is entirely undocumented (and should probably be removed #7248) and most of the documentation about behaviour of server handlers was already in `HTTP::Server`.

See also #7251 for more changes to the existing docs.

asterite:
Yes, `RequestProcessor` is just a helper class, it's just public to be able to test it.

straight-shoota:
Rebased and ready to ship.

PR #4803, opened 2017-08-06
[WIP] update freebsd vagrant box to lastest release (11.1)

tosmi:
This PR updates the Vagrantfile for building crystal with FreeBSD 11.1 and fixes #4802. The following changes are implement

- bump box version to FreeBSD 11.1
- change shell from /bin/csh to /bin/sh to avoid warnings
- set base_mac, otherwise you have to run vagrant up twice as the upstream box does not set base_mac
- update the package list
- use -q (quiet) for install packages
- disable folder sharing and just clone the upstream crystal repo. IMHO this makes the box easier to use as most people are not running NFSD on their machines. Git is installed inside the box so updating the crystal repos from inside the box is not a problem.

Issue #5843, opened 2018-03-20
Crystal::System::Random buffers /dev/urandom


kostya:
i find this while using https://github.com/kostya/run_with_fork. Crystal in `after_fork_child_callbacks` calls `Random::DEFAULT.new_seed`, but new seed for every fork get from Random::Secure.rand which is also the same for each fork. May be mix some Process.pid into seed?

```crystal
require "uuid"

10.times do
  fork do
    p({ rand, UUID.random.to_s })
    exit
  end
end

sleep 1.0
```

```
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
{0.9253318028114554, "d74c282a-0c1d-402c-9e59-56bccf6d1e87"}
```

kostya:
this fix rand, but not uuid

```crystal
require "uuid"

class Random::PCG32
  def new_seed
    new_seed(Random::Secure.rand(UInt64::MIN..UInt64::MAX) + Process.pid, Random::Secure.rand(UInt64::MIN..UInt64::MAX))
  end
end

10.times do
  fork do
    p({ rand, UUID.random.to_s })
    exit
  end
end

sleep 1.0
```

```
{0.4068051713856044, "653031dd-c2bb-4b18-9282-069d60b762bf"}
{0.021214472643506514, "653031dd-c2bb-4b18-9282-069d60b762bf"}
{0.5559970727791962, "653031dd-c2bb-4b18-9282-069d60b762bf"}
{0.7634387669881384, "653031dd-c2bb-4b18-9282-069d60b762bf"}
[...]

PR #7066, opened 2018-11-12
Update Vagrantfile

### LINKED ISSUE #4802: Update Vagrantfile to support latest FreeBSD release (11.1)
The current Vagrantfile uses FreeBSD 10.2 and has a number of problems

- the FreeBSD base box does not set base_mac so you have to run vagrant up twice
- the package list is outdated
- the shell for vagrant is set to csh, FreeBSD comes with /bin/sh which avoids some warnings
- folder sharing is based on NFS, IMHO this is not necessary for just building crystal. For example I'm not running NFSD on my workstation and starting one just for building crystal is overkill. I would propose just cloning the the upstream crystal repository to the homedir of the vagrant user.

I'm going to use the updated box for creating a FreeBSD port/package for crystal

I'm working on a PR for fixing these issues, just let me know if there is a chance of getting this merged.

thanks

### LINKED ISSUE #6722: Vagrantfile
The Vagrantfile is a mess. I tried to use it to test the sockets specs on different systems, but there are a lot of issues:

* The latest LTS Ubuntu Bionic is missing.
* FreeBSD 11 is missing (#4802) and the provisioning for FreeBSD is error-prone. There is a year-old stale PR #4803 with details.
* Ubuntu Xenial uses an outdated, custom build LLVM-3.5 even if already comes with 3.8.
* Ubuntu uses the setup script to install crystal, Debian doesn't. For no apparent reason.
* There are issues with integrating the crystal source code which should ideally be shared folder, not a copy. See https://github.com/crystal-lang/crystal/pull/4803#discussion_r131628383 for details.
* OpenBSD is entirely missing. Would be great to add it, considering it's a tier 2 platform.
* Alpine Linux with musl would also be great to add.
* The selection of Linux distributions is a bit strange. Debian and Ubuntu are both present but they are actually very similar. Relevant differences are mostly regarding available packages and versions. Other distributions such as CentOS, Arch, Gentoo are entirely missing. I'm not saying they should all be added, it just strikes me that there are 10 different Debian/Ubuntu releases (and not even including the latest LTS) and no other distributions.

I'm just wondering about whether the vagrant file is entirely supported, then these issues need to be addressed to be useful in any way. Or maybe it should just be removed and developers use their own VM specs (or separately maintained Vagrantfile). WDYT?

bcardiff:
I'm was trying to bring up to date the Vagrantfile.

Despite which distros/versions are included I wanted to be able to:

**a.** Install a crystal compiler using each distro package management (even if they are not maintained by crystal-lang) and reproduce better the end use experience.

**b.** Install a crystal compiler from a .tar.gz hosted in github. This will allow up to date test without waiting for each platform and compile master easily without non crystal-lang resource.

I am not sure how to achieve (b) for freebsd and alpine. Or if we need to port the compiler in other platform before using it in the target ones. Help is apreciated here :-)

Aiming for (b) helps to have a porting instructions written that depends only in crystal-lang resources. It's a great reference for the future IMO.

Supporting (b) means that dependencies needs to be installed explictly. Whether for (a) that is not needed. This is the reason why lots of dependencies appears.

I'm also refactored the configurate to have an explicit list of machines defined, easier to follow and probably to tweak.

I also added some docs and notes, but before going further I would apprecite input to this a resource we can all feel comfortable using it.

Closes #4802, Closes #4803, Closes #6722.

cc: @j8r, @jirutka, @myfreeweb

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,7 +36,7 @@
 # * alpine is unable to to use the github .tar.gz compiler
 #
 
-INSTALL_GITHUB_TARGZ = true
+INSTALL_GITHUB_TARGZ = false
 GITHUB_URL = "https://github.com/crystal-lang/crystal/releases/download/0.27.0/crystal-0.27.0-1"
 CRYSTAL_LINUX64_TARGZ = "#{GITHUB_URL}-linux-x86_64.tar.gz"
 CRYSTAL_LINUX32_TARGZ = "#{GITHUB_URL}-linux-i686.tar.gz"
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -34,7 +34,6 @@
 # * freebsd needs post install manual scripts in order to install crystal.
 # * alpine provisionning is failing due to openssl libressl issues.
 # * alpine is unable to to use the github .tar.gz compiler
-# * there is no crystal pre built package for alpine32
 #
 
 INSTALL_GITHUB_TARGZ = true
@@ -58,7 +57,6 @@ Vagrant.configure("2") do |config|
   define_freebsd config, name: 'freebsd11', box: '11.2-STABLE', github_targz: false
 
   define_alpine config, name: 'alpine64', bits: 64, github_targz: false
-  define_alpine config, name: 'alpine32', bits: 32, github_targz: false
 
   config.vm.provider "virtualbox" do |vb|
     vb.memory = 6*1024
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -186,7 +186,8 @@ def install_crystal(c, family, dist:, bits:, github_targz:)
       MSG
     when "alpine"
       c.vm.provision :shell, inline: %(
-        apk add --no-cache crystal shards
+        apk add --no-cache crystal shards \
+          || echo "WARNING: apk failed (ignored)"
       )
     end
   end
@@ -288,7 +289,7 @@ def define_freebsd(config, name:, box:)
   end
 end
 
-def define_alpine(config, name:, bits:)
+def define_alpine(config, name:, bits:, llvm: '4')
   config.vm.define name do |c|
     c.vm.box = "alpine/alpine#{bits}"
 
@@ -297,13 +298,18 @@ def define_alpine(config, name:, bits:)
     c.vm.provision :shell, inline: %(
       echo '' > /etc/profile.d/crystal.sh
 
-      echo 'export LLVM_CONFIG=llvm4-config' >> /etc/profile.d/crystal.sh
       echo 'export FLAGS="--target x86_64-linux-musl --link-flags=-no-pie --static"' >> /etc/profile.d/crystal.sh
+
+      apk update
 
       apk add --no-cache \
         tar curl git gcc g++ make automake libtool autoconf bash coreutils paxmark \
         zlib-dev yaml-dev pcre-dev libxml2-dev gmp-dev readline-dev libressl-dev libatomic_ops libevent-dev gc-dev \
-        llvm4-dev llvm4-static || echo "WARNING: apk failed (ignored)"
+        || echo "WARNING: apk failed (ignored)"
+
+      echo 'export LLVM_CONFIG=llvm#{llvm}-config' >> /etc/profile.d/crystal.sh
+      apk add --no-cache llvm#{llvm}-dev llvm#{llvm}-static \
+        || echo "WARNING: apk failed (ignored)"
     )
 
     install_crystal(c, "alpine", dist: nil, bits: bits, github_targz: false)
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -27,6 +27,8 @@
 # * trusty64 FLAGS=--no-debug is unable to build the compiler.
 # * precise32 is unable to build std_spec due to outdated libxml2. ~> 2.9.0 is required
 # * precise32 will fail running std_spec related to reuse_port & ipv6
+# * freebsd needs post install manual scripts in order to install crystal.
+#
 
 def clone_crystal_from_vagrant(config)
   # use ~/crystal directory instead of /vagrant
@@ -145,6 +147,46 @@ def define_debian(config, name:, dist:, bits:, llvm: nil)
   end
 end
 
+def define_freebsd(config, name:, box:)
+  config.vm.define name do |c|
+    c.ssh.shell = "sh"
+    c.vm.box = "freebsd/FreeBSD-#{box}"
+
+    c.vm.hostname = name
+    c.vm.base_mac = "6658695E16F0"
+
+    c.vm.network "private_network", type: "dhcp"
+    c.vm.synced_folder ".", "/vagrant", type: "nfs"
+
+    c.vm.provision :shell, inline: %(
+      pkg install -y git gmake pkgconf
+      pkg install -y libyaml gmp libunwind libevent pcre boehm-gc-threaded
+      pkg install -y llvm60
+    )
+
+    c.vm.post_up_message = <<-MSG
+
+    To install crystal from sources:
+
+    $ sudo portsnap fetch extract
+    $ sudo make -C/usr/ports/lang/crystal/ reinstall clean BATCH=yes
+    $ sudo make -C/usr/ports/devel/shards/ reinstall clean BATCH=yes
+
+    To install pre-built crystal:
+
+    $ sudo pkg install -y crystal shards
+
+    Compile crystal with:
+
+    $ cd ~/crystal
+    $ gmake crystal
+
+    MSG
+
+    clone_crystal_from_vagrant(c)
+  end
+end
+
 Vagrant.configure("2") do |config|
   define_ubuntu config, name: 'bionic64', dist: 'bionic', bits: 64
   define_ubuntu config, name: 'xenial64', dist: 'xenial', bits: 64
@@ -158,24 +200,8 @@ Vagrant.configure("2") do |config|
   define_debian config, name: 'stretch64', dist: 'stretch', bits: 64
   define_debian config, name: 'jessie64', dist: 'jessie', bits: 64
 
-  config.vm.define "freebsd" do |c|
-    c.ssh.shell = "csh"
-    c.vm.box = "freebsd/FreeBSD-10.2-RELEASE"
+  define_freebsd config, name: 'freebsd11', box: '11.2-STABLE'
 
-    c.vm.network "private_network", type: "dhcp"
-    c.vm.synced_folder ".", "/vagrant", type: "nfs"
-
-    # to build boehm-gc from git repository:
-    #c.vm.provision :shell, inline: %(
-    #  pkg install -y libtool automake autoconf libatomic_ops
-    #)
-
-    c.vm.provision :shell, inline: %(
-      pkg install -y git gmake pkgconf pcre libunwind clang36 libyaml gmp libevent2
-    )
-
-    clone_crystal_from_vagrant(c)
-  end
 
   config.vm.provider "virtualbox" do |vb|
     vb.memory = 6*1024
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -18,7 +18,17 @@
 #
 # The vagrant machines will be provisined with a native crystal
 # package or with the targz version downloaded from github.
-# Change INSTALL_GITHUB_TARGZ to choose the installation method.
+# Set `INSTALL_GITHUB_TARGZ` to download .tar.gz files from github.
+#
+# ```
+# $ INSTALL_GITHUB_TARGZ=true vagrant up xenial64
+# ```
+#
+# Configuration via environment variables
+#
+# * `INSTALL_GITHUB_TARGZ=true|false` native package vs github tar.gz installation
+# * `VM_CPUS=N` number of cpus (default: 2)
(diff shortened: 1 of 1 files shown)


j8r:
In my opinion for rolling release distributions like Arch Linux, and those supporting only the latest stable release like Alpine and Fedora, I'll rather use Docker to test them. They are all in a recent Linux kernel.

In the mean time we can have an official Crystal image for each of this distributions.
The Docker itself can be runned in a VM with a recent host like Alpine Latest.

For others using older kernels, like CentOS, Debian and Ubuntu, either:
- use a VM for each of them
- have also an official Crystal Docker images for each of them, and to catch issues related to old kernels like [this one](https://github.com/crystal-lang/crystal/issues/5843), with a Docker host with the minimum Linux kernel version supported: e.g. Ubuntu 14.04 (3.13) or CentOS 7 (3.10 with backports).

Furthermore, the Docker option can free time for the core maintainers by having this Dockerfiles maintained by the community. The best would be https://github.com/docker-library/crystal, or https://github.com/crystal-lang/dockerfiles . The release scheme could be like [node](https://github.com/nodejs/docker-node).

Finally, I also propose to use https://github.com/multiarch/qemu-user-static to test the compiler against `multiarch/alpine:aarch64-latest-stable`, and why not produce a tarball release for aarch64 :grinning: 

For other OS like BSDs, Windows and macOS, I don't see any other choice than VMs.

straight-shoota:
@j8r Using docker is certainly a good option for development but it serves a different purpose than using real VMs.  And on non-linux hosts, docker still needs some kind of virtualization or compatibility layer. 
IMHO it's a good idea to try to have all supported platforms available in the Vagrantfile. We can also have community-maintained Docker images. They're not mutually exclusive.

straight-shoota:
@bcardiff I'd suggest to leave (b) out of the Vagrantfile. This adds a lot of complexity. I think it would be better solved using separate installation scripts. These could be run repeatedly without needing to provision the machine again. And such scripts could also be useful in other contexts on the respective platforms.

They could be simple shell scripts in this repository and the appropriate one being copied to the Vagrant machine. In order to use a recent Crystal version, you boot up a Vagrant machine and execute the installer script. It will install the binary package first, but I don't think that's a big issue. The package is not that large and skipping the install doesn't have much benefit. And it doesn't hurt overriding the installed package.

bcardiff:
@straight-shoota  The things is that if (b) is not done it's halfway story. We won't be able to ensure the Vagrantfile is usable because it depends on external artifacts. On other side, downloading the .tar.gz is one more installation method we actually encourage. Also it would help having a path for platforms that do not have a package right now (like alpine32)

@j8r We won't be adding more native packages support soon.

I'm ok if we only keep (b) for debian/ubuntu. And not use it for freebsd/alpine for now.

I am more interested in bringing the Vagrantfile up to date in order to be able to test some platform sensible issues, like detecting thread stack size to mention one.

I thought that the support from the tar.gz to freebsd/alpine could be done in the provisioning in an automated way. Maybe we need some fixes to allow that to happen, I see some patches in alpine package.

j8r:
Can we also test aarch64 for Alpine? There is [vagrant-libvirt](https://github.com/vagrant-libvirt/vagrant-libvirt) that supports Qemu. There is a demand to run Crystal in their Raspberry Pi 3.

Lots of people are also using Docker in general, this could also be tested in a VM image (thus the `multiarch/alpine:aarch64-latest-stable` can be used to test aarch64)

straight-shoota:
> We won't be able to ensure the Vagrantfile is usable because it depends on external artifacts.

If the installers scripts are part of the repository and copied in the Vagrantfile, they're hardly external artifacts.

> Also it would help having a path for platforms that do not have a package right now (like alpine32)

Sure. In this case, there is no other way. But there is probably a reason why this package doesn't exist, it could be some effort to get it ported.

> I see some patches in alpine package.

They're actually no longer required for 0.27.0. It should work without patches. Only #7076 and #7077 are keeping Crystal 0.27.0 from working 100% on alpine. And they're just issues in the spec suite.

@j8r 

> There is vagrant-libvirt that supports Qemu.

I'm not sure how useful it would be to rely on emulation. It requires and additional plugin and it's going to run pretty slow.

Vagrantfile
```diff
@@ -1,71 +1,302 @@
...
+# * xenial32 FLAGS=--no-debug is unable to run compiler_spec.
+# * trusty64 FLAGS=--no-debug is unable to build the compiler.
+# * precise32 is unable to build std_spec due to outdated libxml2. ~> 2.9.0 is required.
+# * precise32 will fail running std_spec related to reuse_port & ipv6.
+# * freebsd needs post install manual scripts in order to install crystal.
+# * alpine provisionning is failing due to openssl libressl issues.
+# * alpine is unable to to use the github .tar.gz compiler
+#
+
+INSTALL_GITHUB_TARGZ = false
```
straight-shoota: Couldn't this use an environment variable? This makes it easy to toggle source install without changing the Vagrantfile.

Vagrantfile
```diff
@@ -1,71 +1,302 @@
...
+
+  define_debian config, name: 'stretch64', dist: 'stretch', bits: 64
+  define_debian config, name: 'jessie64', dist: 'jessie', bits: 64
+
+  define_freebsd config, name: 'freebsd11', box: '11.2-STABLE'
+
+  define_alpine config, name: 'alpine64', bits: 64
+
+  config.vm.provider "virtualbox" do |vb|
+    vb.memory = 6*1024
```
straight-shoota: Could we make this setting configurable? Allotting 6GB is quite a lot, especially if you're running multiple VMs in parallel (or don't have much memory available anyway).
So much space is only necessary for compiling the compiler, but you can safely go with much less when working only on the stdlib.

Vagrantfile
```diff
@@ -1,71 +1,302 @@
...
+      mkdir -p /opt/crystal
+      echo '#{targz_url}'
+      curl -sSL --output - '#{targz_url}' | tar xz -C /opt/crystal --strip-component=1 -f -
+
+      echo 'export LIBRARY_PATH=/opt/crystal/lib/crystal/lib/:${LIBRARY_PATH}' >> /etc/profile.d/crystal.sh
+      echo 'export PATH=/opt/crystal/bin:$PATH' >> /etc/profile.d/crystal.sh
+    )
+  else
+    case family
+    when "ubuntu"
```
straight-shoota: Why not merge the instructions for Ubuntu and Debian? They're essentially the same.

Vagrantfile
```diff
@@ -1,71 +1,302 @@
...
+      MSG
+    when "alpine"
       c.vm.provision :shell, inline: %(
-        apt-get -y install apt-transport-https dirmngr
-        apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54
-        echo 'deb https://dist.crystal-lang.org/apt crystal main' > /etc/apt/sources.list.d/crystal.list
-        apt-get update
-        apt-get -y install crystal curl git libgmp3-dev zlib1g-dev libedit-dev libxml2-dev libssl-dev libyaml-dev libreadline-dev g++ llvm-#{ver} llvm-#{ver}-dev
-        echo 'export LIBRARY_PATH="/opt/crystal/embedded/lib"' > /etc/profile.d/crystal.sh
+        apk add --no-cache crystal shards
```
straight-shoota:
```suggestion
        apk update
        apk add --no-cache crystal shards
```

Vagrantfile
```diff
@@ -1,71 +1,302 @@
...
+    c.vm.provision :shell, inline: %(
+      echo '' > /etc/profile.d/crystal.sh
+
+      echo 'export LLVM_CONFIG=llvm4-config' >> /etc/profile.d/crystal.sh
+      echo 'export FLAGS="--target x86_64-linux-musl --link-flags=-no-pie --static"' >> /etc/profile.d/crystal.sh
+
+      apk add --no-cache \
+        tar curl git gcc g++ make automake libtool autoconf bash coreutils paxmark \
+        zlib-dev yaml-dev pcre-dev libxml2-dev gmp-dev readline-dev libressl-dev libatomic_ops libevent-dev gc-dev \
+        llvm4-dev llvm4-static || echo "WARNING: apk failed (ignored)"
```
straight-shoota: Why only LLVM 4?

Vagrantfile
j8r: or `apk add --update`

Vagrantfile
bcardiff: done. I also added env vars for cpu and mem

Vagrantfile
bcardiff: I left 4gb as default as before, and add env var to change the available memory

Vagrantfile
bcardiff: Easier to follow I guess. I would leave them as duplicated actually.

Vagrantfile
bcardiff: I left a single `apk update` in the first provision

Vagrantfile
bcardiff: Is the one used in the build. But I added `llvm:` (default 4) argument and can be used with 5 to define another alpine machine.

bcardiff:
Ready for another review. I keep using this new vagrant file in other branches :-)

RX14:
Sorry, I won't be reviewing this PR because I don't use vagrant.

Vagrantfile
```diff
@@ -1,71 +1,320 @@
...
+      add-apt-repository "deb http://apt.llvm.org/#{dist}/ llvm-toolchain-#{dist}#{llvm_suffix} main"
+      curl -sSL https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+      apt-get update
+      apt-get install -y llvm#{llvm_suffix}-dev
+    )
+  end
+end
+
+# instructions to install crystal binary on all platforms
+def install_crystal(c, family, dist:, bits:, github_targz:)
```
straight-shoota: This method feels very convoluted. It's hard to follow the control flow from individual define methods here and then into different branches again.

I'd prefer to have an `install_from_targz` method and integrate the familiy-specific installation instructions into the respective `define_<family>` methods, maybe separate the install step into `install_crystal_<family>`.

Vagrantfile
```diff
@@ -1,71 +1,320 @@
...
+    clone_crystal_from_vagrant(c)
   end
+end
 
-  config.vm.provider "virtualbox" do |vb|
-    vb.memory = 4096
-    vb.cpus = 2
+def define_alpine(config, name:, bits:, llvm: '4')
+  config.vm.define name do |c|
+    c.vm.box = "alpine/alpine#{bits}"
```
straight-shoota: [`alpine/alpine64`](https://app.vagrantup.com/alpine/boxes/alpine64) doesn't seem to be a great vagrant box. It's current version is still 3.7.0 which was released in 2017.
The most-used alpine box is [`generic/alpine38`](https://app.vagrantup.com/generic/boxes/alpine38) (`alpine39` is also available).

Vagrantfile
```diff
@@ -1,71 +1,320 @@
...
+    setup_memory(config, bits)
+
+    c.vm.provision :shell, inline: %(
+      echo '' > /etc/profile.d/crystal.sh
+
+      echo 'export FLAGS="--target x86_64-linux-musl --link-flags=-no-pie --static"' >> /etc/profile.d/crystal.sh
+
+      apk update
+
+      apk add --no-cache \
```
j8r: you can merge the upper command inside this one: `apk add --update --no-cache`

Vagrantfile
```diff
@@ -1,71 +1,320 @@
...
+  define_ubuntu config, name: 'trusty64', dist: 'trusty', bits: 64, llvm: { url: "http://crystal-lang.s3.amazonaws.com/llvm/llvm-3.9.1-1-linux-x86_64.tar.gz", path: "llvm-3.9.1-1" }
+  define_ubuntu config, name: 'trusty32', dist: 'trusty', bits: 32, llvm: { url: "http://crystal-lang.s3.amazonaws.com/llvm/llvm-3.9.1-1-linux-i686.tar.gz", path: "llvm-3.9.1-1" }
+  define_ubuntu config, name: 'precise64', dist: 'precise', bits: 64, llvm: { url: "http://crystal-lang.s3.amazonaws.com/llvm/llvm-3.9.1-1-linux-x86_64.tar.gz", path: "llvm-3.9.1-1" }
+  define_ubuntu config, name: 'precise32', dist: 'precise', bits: 32, llvm: { url: "http://crystal-lang.s3.amazonaws.com/llvm/llvm-3.9.1-1-linux-i686.tar.gz", path: "llvm-3.9.1-1" }
+
+  define_debian config, name: 'stretch64', dist: 'stretch', bits: 64
+  define_debian config, name: 'jessie64', dist: 'jessie', bits: 64
+
+  define_freebsd config, name: 'freebsd11', box: '11.2-STABLE'
+
```
j8r: I guess we also want to test against FreeBSD 12

jhass:
This is probably pretty stale by now. Is anyone really using these or mabye it's okay to just drop them?

bcardiff:
It can come back later if needed. I haven’t been using it for a while, and even then I use the PR rather than master. So it’s the same for me to have a spare copy rather than force my workflow to others.

RX14:
Shall we remove the vagrantfile then?

PR #7067, opened 2018-11-12
Fix socket specs error messages for musl

### LINKED ISSUE #7057: Socket specs failing on musl
See https://github.com/alpinelinux/aports/pull/5560#issuecomment-437120481

The first errors are just that musl seems to have different error messages than glibc. We already account for differences between linux and bsd/darwin, so it's just adding another set of messages.

The last failure comes from OpenSSL error handling (with OpenSSL 1.1) where initializing an `OpenSSL::SSL::Error` actually raises `Errno`. This seems wrong, at this point it should always raise an SSL error, even if the cause is an IO error (which can be wrapped).

straight-shoota:
Fixes #7057 (the first part)

/cc alpinelinux/aports#5560

### DIFF
--- a/spec/std/socket/tcp_server_spec.cr
+++ b/spec/std/socket/tcp_server_spec.cr
@@ -33,7 +33,7 @@ describe TCPServer do
       end
 
       it "raises when port is negative" do
-        expect_raises(Socket::Error, linux? ? "getaddrinfo: Servname not supported for ai_socktype" : "No address found for #{address}:-12 over TCP") do
+        expect_raises(Socket::Error, linux? ? (musl? ? "getaddrinfo: Unrecognized service" : "getaddrinfo: Servname not supported for ai_socktype") : "No address found for #{address}:-12 over TCP") do
           TCPServer.new(address, -12)
         end
       end
--- a/spec/std/socket/spec_helper.cr
+++ b/spec/std/socket/spec_helper.cr
@@ -20,3 +20,7 @@ end
 def linux?
   {{ flag?(:linux) }}
 end
+
+def musl?
+  {{ flag?(:musl) }}
+end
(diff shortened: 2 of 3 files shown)


spec/std/socket/tcp_server_spec.cr
```diff
@@ -33,7 +33,7 @@ describe TCPServer do
       end
 
       it "raises when port is negative" do
-        expect_raises(Socket::Error, linux? ? "getaddrinfo: Servname not supported for ai_socktype" : "No address found for #{address}:-12 over TCP") do
+        expect_raises(Socket::Error, linux? ? (musl? ? "getaddrinfo: Unrecognized service" : "getaddrinfo: Servname not supported for ai_socktype") : "No address found for #{address}:-12 over TCP") do
```
RX14: This line is getting a bit long, how about checking for errno by number instead of by message?

straight-shoota: I'm not sure... Error numbers have portability issues.

RX14: We have constants under `Errno` for the values. We should do the same for `Socket::Error`.

ysbaddaden:
Alpine can apply this patch in the APKBUILD. They did previously.

I'd like the `linux?` and now `musl?` query methods to not exist. The flag checks should be inlined and happen at compile only, not delayed to runtime.

Checking Errno numbers would avoid all these issues, too, and thus be the correct patch.

straight-shoota:
> Alpine can apply this patch in the APKBUILD. They did previously.

Sure. But it's always easier if you don't need special patches.

Why don't we proceed like suggested in https://github.com/crystal-lang/crystal/pull/7067#discussion_r233145662. This PR is an improvement and fixes an existing error. Let's merge it and refactor afterwards. We should take the little steps, even if there are bigger ones to follow.

straight-shoota:
Oh, I noticed the error constants are already defined. So I guess it's relatively easy to implement expectations based on error codes. It just needs a new error class.

PR #7075, opened 2018-11-13
Add setter for server response headers

nhh:
While working with the basic HTTP library, i noticed that i cannot set the response headers for a HTTP::Server::Response

I also changed the usage of "headers" to "@headers" for clarity.

Consider the following situation:
(response is a HTTP::Client::Response)
``` crystal
client.exec(..) do |response|
  # context.response.headers = response.headers
  context.response.status_code = response.status_code
  context.response.content_type = "text/html"
  context.response.print(response.body_io.gets)
  context.response.close
end
```


asterite:
Why not use merge?

nhh:
Did i choose rebase? Or what do you mean by that?

I just wanted to have some feedback, because i dont know if that behavior was expected.

PR #7077, opened 2018-11-13
Add MIME::MediaType

### LINKED ISSUE #5686: Support parsing Content-Disposition: file headers
In a multi-file upload the files can be uploaded in a subpart with `Content-Disposition: file; filename="file1.txt"`

Currently one must detect `multipart/mixed` form data and parse the sub-multipart data outside of the standard library.

For example post data see https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2
References: https://tools.ietf.org/html/rfc2388
```
   either of the "content-disposition: form-data"
   header or, in the case of multiple files, in a "content-disposition:
   file" header of the subpart.
```

straight-shoota:
This PR adds a `MIME::MediaType` type with methods for parsing mime media types as per [RFC 2045](https://tools.ietf.org/html/rfc2045) and [RFC 2046](https://tools.ietf.org/html/rfc2046) including support for optional parameters.

This type is put to use for reading and writing HTTP `Content-Type` headers and `HTTP::Multipart` boundaries replacing existing incomplete implementations.

This is not directly integrated with the `MIME` registry (added in #5765) but can be optionally used together.

Closes #5686

### DIFF
--- a/spec/std/mime/media_type_spec.cr
+++ b/spec/std/mime/media_type_spec.cr
@@ -1,0 +1,291 @@
+require "spec"
+require "mime/media_type"
+
+private def parse(string)
+  type = MIME::MediaType.parse(string)
+
+  {type.media_type, type.@params}
+end
+
+private def assert_format(string, format = string, file = __FILE__, line = __LINE__)
+  MIME::MediaType.parse(string).to_s.should eq(format), file, line
+end
+
+describe MIME::MediaType do
+  describe ".new" do
+    it "create new instance" do
+      media_type = MIME::MediaType.new("foo/bar", {"param" => "value"})
+      media_type.media_type.should eq "foo/bar"
+      media_type["param"].should eq "value"
+    end
+
+    it "raises for invalid parameter name" do
+      expect_raises(MIME::Error, %q(Invalid parameter name "ß")) do
+        MIME::MediaType.new("foo/bar", {"param" => "value", "ß" => "invalid"})
+      end
+    end
+  end
+
+  describe ".parse" do
+    it "parses media type" do
+      MIME::MediaType.parse("text/html; charset=utf-8").media_type.should eq "text/html"
+    end
+
+    it "parses params" do
+      parse("text/html; charset=utf-8").should eq({"text/html", {"charset" => "utf-8"}})
+
+      parse("text/html; charset=us-ascii").should eq({"text/html", {"charset" => "us-ascii"}})
+
+      parse("text/html; foo = bar; bar= foo ;").should eq({"text/html", {"foo" => "bar", "bar" => "foo", "charset" => "utf-8"}})
+
+      parse(%(form-data; name="foo")).should eq({"form-data", {"name" => "foo"}})
+      parse(%(form-data; name ="foo")).should eq({"form-data", {"name" => "foo"}})
+      parse(%(form-data; name= "foo")).should eq({"form-data", {"name" => "foo"}})
+
+      parse(%( form-data ; name=foo)).should eq({"form-data", {"name" => "foo"}})
+
+      parse(%(FORM-DATA;name="foo")).should eq({"form-data", {"name" => "foo"}})
+
+      parse(%( FORM-DATA ; name="foo")).should eq({"form-data", {"name" => "foo"}})
+
+      expect_raises(MIME::Error, "Missing media type") { parse("") }
+      expect_raises(MIME::Error, "Missing media type") { parse(" ") }
+      expect_raises(MIME::Error, "Missing media type") { parse(";") }
+      expect_raises(MIME::Error, "Missing media type") { parse(" ;") }
+
+      expect_raises(MIME::Error, "Missing attribute name at 10") { parse("form-data;=foo") }
+      expect_raises(MIME::Error, "Missing attribute name at ") { parse("form-data; =foo") }
+      expect_raises(MIME::Error, "Missing attribute value") { parse("form-data;foo") }
+
+      parse(%(form-data; key=value;  blah="value";name="foo" )).should eq({"form-data", {"key" => "value", "blah" => "value", "name" => "foo"}})
+
+      expect_raises(MIME::Error, "Duplicate key 'key' at 15") { parse(%(foo; key=val1; key=the-key-appears-again-which-is-bogus)) }
+
+      parse(%(FORM-DATA;NAMe="foo")).should eq({"form-data", {"name" => "foo"}})
+
+      # From RFC 2231:
+
+      parse(%(application/x-stuff; title*=us-ascii'en-us'This%20is%20%2A%2A%2Afun%2A%2A%2A)).should eq({"application/x-stuff", {"title" => "This is ***fun***"}})
+
+      parse(%(message/external-body; access-type=URL; URL*0="ftp://";URL*1="cs.utk.edu/pub/moore/bulk-mailer/bulk-mailer.tar")).should eq({
+        "message/external-body",
+        {"access-type" => "URL", "url" => "ftp://cs.utk.edu/pub/moore/bulk-mailer/bulk-mailer.tar"},
+      })
+
+      # Tests from http://greenbytes.de/tech/tc2231/
+
+      # attonly
+      parse(%(attachment)).should eq({"attachment", {} of String => String})
+      # attonlyucase
+      parse(%(ATTACHMENT)).should eq({"attachment", {} of String => String})
+      # attwithasciifilename
+      parse(%(attachment; filename="foo.html")).should eq({"attachment", {"filename" => "foo.html"}})
+      # attwithasciifilename25
+      parse(%(attachment; filename="0000000000111111111122222")).should eq({"attachment", {"filename" => "0000000000111111111122222"}})
+      # attwithasciifilename35
+      parse(%(attachment; filename="00000000001111111111222222222233333")).should eq({"attachment", {"filename" => "00000000001111111111222222222233333"}})
+      # attwithasciifnescapedchar
+      parse(%(attachment; filename="f\\oo.html")).should eq({"attachment", {"filename" => "f\\oo.html"}})
+      # attwithasciifnescapedquote
+      parse(%(attachment; filename="\\"quoting\\" tested.html")).should eq({"attachment", {"filename" => %("quoting" tested.html)}})
+      # attwithquotedsemicolon
+      parse(%(attachment; filename="Here's a semicolon;.html")).should eq({"attachment", {"filename" => "Here's a semicolon;.html"}})
+      # attwithfilenameandextparam
+      parse(%(attachment; foo="bar"; filename="foo.html")).should eq({"attachment", {"foo" => "bar", "filename" => "foo.html"}})
+      # attwithfilenameandextparamescaped
+      parse(%(attachment; foo="\\"\\\\";filename="foo.html")).should eq({"attachment", {"foo" => %("\\), "filename" => "foo.html"}})
+      # attwithasciifilenameucase
+      parse(%(attachment; FILENAME="foo.html")).should eq({"attachment", {"filename" => "foo.html"}})
+      # attwithasciifilenamenq
+      parse(%(attachment; filename=foo.html)).should eq({"attachment", {"filename" => "foo.html"}})
+      # attwithasciifilenamenqs
+      parse(%(attachment; filename=foo.html ;)).should eq({"attachment", {"filename" => "foo.html"}})
+      # attwithfntokensq
+      parse(%(attachment; filename='foo.html')).should eq({"attachment", {"filename" => "'foo.html'"}})
+      # attwithisofnplain
+      parse(%(attachment; filename="foo-ä.html")).should eq({"attachment", {"filename" => "foo-ä.html"}})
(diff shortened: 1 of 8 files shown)


spec/std/mime/media_type_spec.cr
```diff
@@ -0,0 +1,226 @@
...
+      parse(%(attachment; foobar=x; filename="foo.html")).should eq({"attachment", {"foobar" => "x", "filename" => "foo.html"}})
+
+      # Browsers also just send UTF-8 directly without RFC 2231,
+      # at least when the source page is served with UTF-8.
+      parse(%(form-data; firstname="Брэд"; lastname="Фицпатрик")).should eq({"form-data", {"firstname" => "Брэд", "lastname" => "Фицпатрик"}})
+
+      # Empty string used to be mishandled.
+      parse(%(foo; bar="")).should eq({"foo", {"bar" => ""}})
+
+      # Microsoft browers in intranet mode do not think they need to escape \ in file name.
```
r00ster91: Typo: **browers**.

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+          io << value
+        else
+          io << '"'
+          MIME::MediaType.quote_string(value, io)
+          io << '"'
+        end
+      end
+    end
+
+    # Access a parameter value.
```
r00ster91:
```suggestion
    # Accesses a parameter value.
```

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+    end
+
+    # Parses a mime type string representation including any optional parameters,
+    # per RFC 1521.
+    # Media types are the values in `Content-Type` and `Content-Disposition` HTTP
+    # headers (RFC 2183).
+    #
+    # Media type is lowercased and trimmed of white space. Param keys are lowercased.
+    #
+    # Raises on error.
```
r00ster91:
```suggestion
    # Raises `MIME::Error` on error.
```

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+    # ```
+    # MIME::MediaType.new("text/plain charset=UTF-8")["charset"] # => "UTF-8"
+    # ```
+    #
+    # Returns `nil` if the parameter key is not set.
+    def [](key : String) : String
+      @params[key]
+    end
+
+    # Access a parameter value.
```
r00ster91: Ditto.

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+          MIME::MediaType.quote_string(value, io)
+          io << '"'
+        end
+      end
+    end
+
+    # Access a parameter value.
+    #
+    # ```
+    # MIME::MediaType.new("text/plain charset=UTF-8")["charset"] # => "UTF-8"
```
Sija:
```suggestion
    # MIME::MediaType.new("text/plain; charset=UTF-8")["charset"] # => "UTF-8"
```

ditto below

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+    property params : Hash(String, String)
+
+    # Creates a new `MediaType` instance.
+    def initialize(@media_type, @params = {} of String => String)
+    end
+
+    def to_s(io : IO)
+      io << media_type
+
+      params.keys.sort.each do |key|
```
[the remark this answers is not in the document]
straight-shoota: Because the order is insignificant, it makes sense to sort the parameters for display. Thus identical parameter sets will always print in the same way, no matter the order in which they were constructed.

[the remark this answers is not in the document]
RX14: I don't think it's worth the performance loss. You should do this in `inspect`.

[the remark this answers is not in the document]
straight-shoota: Having a different order in `inspect` and `to_s` doesn't seem like a good idea.
I'd prefer to leave it as is for now. Sorting on `to_s` is simply an implementation detail demanded by the way these params are stored. `Hash` is just not a very efficient data structure for this application. But we currently don't have any alternatives in the stdlib and I don't think it's worth introducing a custom implementation for this specific use case.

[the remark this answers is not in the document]
RX14: Completely disagree.

src/mime/media_type.cr
```diff
@@ -0,0 +1,320 @@
...
+        when ';'
+          return
+        when '='
+          key = string.byte_slice(key_start, reader.pos - key_start).strip.downcase
+
+          reader.next_char
+
+          quoted = nil
+          open_quote = false
+          if reader.current_char == '"'
```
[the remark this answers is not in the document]
straight-shoota: White space between `=` and `quoted-string` (delimiter `"`) is illegal.

src/mime/media_type.cr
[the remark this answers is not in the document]
straight-shoota: Could you be more specific? Yes, it's less performant than it could be. But performance can be improved by changing the internal data handling. That's not required to be implemented right now.

src/mime/media_type.cr
```diff
@@ -5,8 +5,7 @@ require "mime"
 module MIME
   # A MediaType describes a MIME content type with optional parameters.
   struct MediaType
-    property media_type : String
-    property params : Hash(String, String)
```
Sija: Why? Having access to raw params Hash could come handy, when removed there's no way to do that.

straight-shoota: `MediaType` directly exposes `#[]`, `#[]?`, `#fetch`, `#[]=` and `#each_parameter`. Why would yo need direct access to `params`?

Sija: For serialization as an example.

straight-shoota: What kind of serialization? I figure you'd either serialize with `#to_s` or directly access the ivar `@params` (like `JSON::Serializable`).

src/mime/media_type.cr
```diff
@@ -34,7 +34,7 @@ module MIME
     #
     # ```
     # MIME::MediaType.parse("text/plain; charset=UTF-8")["charset"] # => "UTF-8"
-    # MIME::MediaType.parse("text/plain; charset=UTF-8")["foo"] # KeyError
+    # MIME::MediaType.parse("text/plain; charset=UTF-8")["foo"]     # KeyError
```
Sija: `KeyError` -> `raises KeyError`

src/mime/media_type.cr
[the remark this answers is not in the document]
RX14: `to_s` is not for humans, it's for machines. We shouldn't make any performance concessions for humans. It's just how these methods are specified. Sort them for `inspect`.

src/mime/media_type.cr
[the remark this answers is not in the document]
RX14: I see nothing excepting `quoted-string` from the allowable free insertion of  linear-white-space between any atom.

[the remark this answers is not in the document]
straight-shoota: Hm, yeah okay. I didn't see that. RFC 2045 doesn't explicitly declare that these whitespace rules also apply, it only refers to some syntax rules of RFC 822. And RFC 822 allows whitespace in header fields, while mime types are not necessarily only used in headers.
But it looks safe to assume that white space should be allowed nonetheless, that's seems like the most sane way.

src/mime/media_type.cr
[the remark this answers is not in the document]
straight-shoota: Do you really think it's a good idea to have different sort oder for `to_s` and `inspect`?

src/mime/media_type.cr
```diff
@@ -0,0 +1,381 @@
...
+      if mediatype.starts_with?("text/")
+        params["charset"] ||= "utf-8"
+      end
+
+      MediaType.new mediatype, params
+    end
+
+    private def self.parse_parameter_value(reader)
+      quoted = nil
+      open_quoite = false
```
Sija:
```suggestion
      open_quote = false
```

src/mime/media_type.cr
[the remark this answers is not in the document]
RX14: Yes.

straight-shoota:
@RX14 I removed `sort` from `#to_s` without an different implementation for `#inspect`. This can be improved later, if necessary.

spec/std/mime/media_type_spec.cr
```diff
@@ -0,0 +1,259 @@
...
+    end
+
+    it "parses params" do
+      parse("text/html; charset=utf-8").should eq({"text/html", {"charset" => "utf-8"}})
+
+      parse("text/html; charset=us-ascii").should eq({"text/html", {"charset" => "us-ascii"}})
+
+      parse("text/html; foo = bar; bar= foo ;").should eq({"text/html", {"foo" => "bar", "bar" => "foo", "charset" => "utf-8"}})
+
+      parse(%(form-data; name="foo")).should eq({"form-data", {"name" => "foo"}})
```
RX14: Can we have an added spec for `key= "value"` and `key ="value`

src/mime/media_type.cr
```diff
@@ -0,0 +1,380 @@
...
+    # headers (RFC 2183).
+    #
+    # Media type is lowercased and trimmed of white space. Param keys are lowercased.
+    #
+    # Returns `nil` on error.
+    def self.parse?(string : String) : MediaType?
+      parse_impl(string) { return nil }
+    end
+
+    private def self.parse_impl(string : String, &block : String ->)
```
RX14: Does restricting this block to `String -> NoReturn` work?

src/mime/media_type.cr
```diff
@@ -0,0 +1,492 @@
...
+      # Stores extended parameters with continuating values
+      continuation = Hash(String, Hash(String, String)).new do |hash, key|
+        hash[key] = {} of String => String
+      end
+
+      while reader.has_next?
+        # Consume optional whitespace
+        while reader.current_char.ascii_whitespace?
+          reader.next_char
+        end
```
RX14: I'd suggest splitting this into `private def consume_whitespace(reader)`

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
...
+
+    # Creates a new `MediaType` instance.
+    def initialize(@media_type, @params = {} of String => String)
+    end
+
+    def to_s(io : IO)
+      io << media_type
+
+      @params.each do |key, value|
+        next unless MIME::MediaType.token? key
```
RX14: Should we really be silently ignoring this?

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
...
+          sub_type_start = reader.pos
+          reader.next_char
+        elsif token?(char)
+          reader.next_char
+        else
+          yield "Invalid character '#{char}' at #{reader.pos}"
+        end
+      end
+
+      # string is emtpy or first character was a ';'
```
r00ster91: Typo: **emtpy**

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
...
+    # headers (RFC 2183).
+    #
+    # Media type is lowercased and trimmed of white space. Param keys are lowercased.
+    #
+    # Raises `MIME::Error` on error.
+    def self.parse(string : String) : MediaType
+      parse_impl(string) { |err| raise MIME::Error.new("Parsing error: #{err} (#{string.dump})") }
+    end
+
+    # Parses a mime type string representation including any optional parameters,
```
r00ster91:
```suggestion
    # Parses a MIME type string representation including any optional parameters,
```

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
...
+    # ```
+    # MIME::MediaType.new("text/plain").sub_type # => "plain"
+    # MIME::MediaType.new("foo").sub_type        # => nil
+    # ```
+    def sub_type : String?
+      index = media_type.byte_index('/'.ord) || return
+      media_type.byte_slice(index + 1, media_type.bytesize - index - 1)
+    end
+
+    # Parses a mime type string representation including any optional parameters,
```
r00ster91:
```suggestion
    # Parses a MIME type string representation including any optional parameters,
```

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
+require "uri"
+require "http"
+require "mime"
+
+module MIME
+  # A MediaType describes a MIME content type with optional parameters.
```
r00ster91:
```suggestion
  # A `MediaType` describes a MIME content type with optional parameters.
```

src/mime/media_type.cr
straight-shoota: Maybe it should raise when setting such a key?

src/mime/media_type.cr
```diff
@@ -0,0 +1,491 @@
...
+require "http"
+require "mime"
+
+module MIME
+  # A `MediaType` describes a MIME content type with optional parameters.
+  struct MediaType
+    getter media_type : String
+
+    # Creates a new `MediaType` instance.
+    def initialize(@media_type, @params = {} of String => String)
```
RX14: We need to validate keys here too.

RX14:
Squash please!

PR #7080, opened 2018-11-13
Add .wasm (application/wasm) mime type

Sija:
SSIA, followup to #5765. Btw, Golang has it baked in, see: https://github.com/golang/go/blob/0098f8aeaceb5feec7462ae64f8ce91a473360c1/src/mime/type.go#L66


Sija:
Shall we have it 🚢 ?

Sija:
Ready to 🎲 ?

PR #7081, opened 2018-11-13
Use read_file macro in Crystal::Config.version

Sija:
Followup to #6967. If accepted, needs to be merged after `0.27.1` release.

### DIFF
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -5,11 +5,7 @@ module Crystal
     end
 
     def self.version
-      {% if flag?(:windows) %}
-        {{ `type #{__DIR__}/../../../VERSION`.stringify.chomp }}
-      {% else %}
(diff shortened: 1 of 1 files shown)


bcardiff:
@Sija do you want to update this with the ` || raise "..."` idiom?

Sija:
@bcardiff first I'd like to discuss adding a raising variant and I'll do it after that if there won't be a consensus.

Sija:
@bcardiff I don't think it's settled since as I wrote in https://github.com/crystal-lang/crystal/pull/6967#issuecomment-439232353 I believe having raising one is still more beneficial - raising outside of `read_file` will obscure the actual error - which might be other than missing file.

woodruffw:
I began work on a variant of `read_file` that raises here: #7094.

bcardiff:
@Sija would you mind pushing a rebase on master?

Sija:
@bcardiff Sure, done.

PR #7090, opened 2018-11-17
Markdown: don't parse markdown inside inline code

### LINKED ISSUE #7065: `#[](...)` is not rendered correctly in docs
When this:
```
`#[](...)`
```
(or without anything in the parentheses)
is in the documentation of for example a method then in the docs it's not rendered correctly and results in `#`.

Example: [`Indexable#unsafe_fetch`](https://crystal-lang.org/api/0.27.0/Indexable.html#unsafe_fetch%28index%3AInt%29-instance-method):
> Clients never invoke this method directly. Instead, they access elements with `#` and [`#[]?(index)`](https://crystal-lang.org/api/0.27.0/Indexable.html#%5B%5D%3F%28index%3AInt%29-instance-method).

Source:
```cr
  # Clients never invoke this method directly. Instead, they access
  # elements with `#[](index)` and `#[]?(index)`.
```

MakeNowJust:
Fixed #7065


MakeNowJust:
🛎 ping

PR #5564, opened 2018-01-10
fixed int128 implementation

omotorin:
I found some bugs in implementation of int128. It related with LLVMConstInt that works only with 64 bit integer and 128 bit will always cut to 64 bit. Also I've implemented to_int128/to_uint128 that needed to correct works codegen for 128 bit integer

PR #7093, opened 2018-11-18
Fix u128

wontruefree:
Pulling in @omotorin commits and trying to finish https://github.com/crystal-lang/crystal/pull/5564


wontruefree:
I am not seeing these errors locally is this a CI thing?

asterite:
You need to run `crystal tool format`

asterite:
By the way, we should probably run this check earlier in CI. It's a waste of time to run all specs and have them pass just to fail later because the formatting is not correct.

src/string.cr
```diff
@@ -459,7 +459,39 @@ class String
 
   # Same as `#to_i` but returns an `UInt64` or the block's value.
   def to_u64(base : Int = 10, whitespace = true, underscore = false, prefix = false, strict = true, &block)
-    gen_to_ u64
+    gen_to_ u64, 18446744073709551615_u64
+  end
+
+  # ## to_i128/to_u128
```
Sija: Redundant comment.

wontruefree: good catch

Sija:
@wontruefree & what about https://github.com/crystal-lang/crystal/pull/5564#discussion_r16066589?

wontruefree:
@Sija I did some googling a little bit ago and read that it is one cycle on 32bit machines anyways I dont know if I can just get that from the x86 spec but I am more then willing to explore that.  I am wondering if I can verify these problems in virtual box.

RX14:
A simple benchmark of `String#to_i` before and after this PR will show why this can't be merged.

wontruefree:
wow I didn't realize what the performance impact would be.
master:
```bash
$ ./.build/crystal benchmark.cr --release
String#to_i32  35.87M ( 27.88ns) (± 2.76%)  0 B/op   1.01× slower
String#to_i64  36.11M ( 27.69ns) (± 2.45%)  0 B/op        fastest
```

this branch:
```bash
$ ./.build/crystal benchmark.cr --release
String#to_i32   2.24M (446.08ns) (± 4.01%)  0 B/op   1.04× slower
String#to_i64   2.32M (430.29ns) (± 1.97%)  0 B/op        fastest
```

wontruefree:
@asterite Thanks for the explanation.  I am getting more understanding the more I get into this.  I am working on it now.

wontruefree:
Ok this is a little better but some extraction might be possible.

```bash
$ ./.build/crystal benchmark.cr --release
String#to_i32  35.96M ( 27.81ns) (± 2.35%)  0 B/op   1.00× slower
String#to_i64  36.12M ( 27.69ns) (± 1.86%)  0 B/op        fastest
```

RX14:
It's not so much a bug as just a lack of intrinsics.

src/string.cr
```diff
@@ -611,6 +641,133 @@ class String
...
+    invalid : Bool
+
+  private macro gen_to_128(method, max_positive = nil, max_negative = nil)
+    info = to_u128_info(base, whitespace, underscore, prefix, strict)
+    return yield if info.invalid
+
+    if info.negative
+      {% if max_negative %}
+        return yield if info.value > {{max_negative}}
+        -info.value.to_{{method}}
```
bcardiff: The same issues expressed in #7172 apply here.

wontruefree: good call I will update this.

wontruefree:
@Sija I have been kind of deep in the weeds in the compiler.
I think I have to add u128 and i128 to the compilers math interpreter. src/compiler/crystal/semantic/math_interpreter.cr

Do you know if that would be the next steps to working on UInt128 to crystal?

Sija:
@wontruefree I think so. Adding missing functions from `compiler-rt` (in the vein of https://github.com/crystal-lang/crystal/pull/7206/files#diff-b1ba6a49a8ea0b8a8425cebf9552302) seems needed too.

Sija:
@wontruefree Are you still planning on finishing this?

wontruefree:
@Sija I am actually working on this today.  But I am not as well versed in the compiler and dont have a 32 bit machine. If this is easy for you to finish you are more then welcome to take it.

wontruefree:
I have been trying to get things to work in vagrant.  I have not gotten repeatable results with cross compiling.

Sija:
If I'm not mistaken you'd need to port below functions to make it work on 32-bit machines:

- `__multi3` => https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/multi3.c
- `__umodti3` => https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/umodti3.c
- `__udivti3` => https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/udivti3.c
- `__modti3` => https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/modti3.c
- `__divti3` => https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/divti3.c

delef:
there any news?

wontruefree:
@delef I was working on it two weeks ago and having a really hard time getting consistent builds on 32bit machines.  Basically I cannot get master to build consistently on VirtualBox.  I have also tried cross-compiling, but I was not confident I got it working correctly.

If you want to take this over or have a way to build 32 bit consistently I would like to know.  If not I work on it when I have time but it is going really slowly until I get 32 bit builds working.

Sija:
@wontruefree any updates?

wontruefree:
@Sija I have identified the rt functions that would need to be implemented and found example implementations in the llvm source code.  I have been having trouble running code on 32-bit VMs, and I had some difficulties running rt functions in general.

If someone wants to take this over and has the experience to do this they should. I can take another pass at it but it probably won't be till later this month.

Also, I have been doing my work on another branch so I could experiment without pushing up code.  But if someone wants the code, it is here.
https://github.com/wontruefree/crystal/tree/fix_u128_bak

wontruefree:
If people are following this I have some updates.  I have figured out how to implement and test this.  It is just a fair amount of work and I have been working through it.  I am hoping to have this by 0.30.0

src/string.cr
```diff
@@ -489,13 +519,14 @@ class String
...
-  record ToU64Info,
-    value : UInt64,
-    negative : Bool,
-    invalid : Bool
+  record UIntInfo, value : (UInt64 | UInt128), negative : Bool, invalid : Bool
 
   private macro gen_to_(method, max_positive = nil, max_negative = nil)
-    info = to_u64_info(base, whitespace, underscore, prefix, strict, leading_zero_is_octal, unsigned: {{max_negative == nil}})
+    if base == UInt128
+      info = to_info(base, whitespace, underscore, prefix, strict, leading_zero_is_octal, unsigned: {{max_negative == nil}})
```
Sija: This branch seems redundant.

src/string.cr
```diff
@@ -489,13 +519,14 @@ class String
...
+    else
+      info = to_info(base, whitespace, underscore, prefix, strict, leading_zero_is_octal, unsigned: {{max_negative == nil}})
```
Sija: What's the reason for having two branches doing the same thing?

src/crystal/compiler_rt/i128_info.cr
```diff
@@ -0,0 +1,7 @@
+module CompilerRT
+  struct I128Info
+    property low : UInt64 = 0_u64, high : Int64 = 0_i64
+
+    def initialize; end
```
Sija: Empty initializers are redundant. ditto elsewhere

src/crystal/compiler_rt/modti3.cr
```diff
@@ -0,0 +1,11 @@
+# Function return the remainder of the signed division eg. `a % b`
+fun __modti3(a : Int128, b : Int128) : Int128
+  bits_in_tword_m1 = sizeof(Int128) * sizeof(Char) - 1
+  s = b >> bits_in_tword_m1 # s = b < 0 ? -1 : 0
+  b = (b ^ s) - s           # negate if s == -1
+  s = a >> bits_in_tword_m1 # s = a < 0 ? -1 : 0
+  a = (a ^ s) - s           # negate if s == -1
+  r = 0_u128
+  __udivmodti4(a.unsafe_as(UInt128), b.unsafe_as(UInt128), pointerof(r))
+  return (r ^ s).unsafe_as(Int128) - s # negate if s == -1
```
Sija: redundant `return`, ditto elsewhere

src/string.cr
wontruefree: I thought this was removed.  It is legacy from when I started this branch.

src/crystal/compiler_rt/modti3.cr
wontruefree: I liked the explicit returns in this case but I understand they are not necessary.

spec/std/crystal/compiler_rt/muloti4_spec.cr
```diff
@@ -0,0 +1,90 @@
+require "spec"
+
+# Ported from compiler-rt:test/builtins/Unit/muloti4_test.c
+
+private def test__muloti4(a : Int128, b : Int128, expected : Int128, expected_overflow : Int128, file = __FILE__, line = __LINE__)
+  it "passes compiler-rt builtins unit tests" do
+    actual_overflow : Int32 = 0
+    actual = __muloti4(a, b, pointerof(actual_overflow))
+    actual_overflow.should eq(expected_overflow), file, line
+    if !expected_overflow
```
Sija:
```suggestion
    unless expected_overflow
```

src/string.cr
```diff
@@ -572,8 +599,8 @@ class String
       end
     end
 
-    value = 0_u64
-    mul_overflow = ~0_u64 // base
+    value = ((base == 128) ? 0_u128 : 0_u64)
```
```suggestion
    value = (base == 128) ? 0_u128 : 0_u64
```

tmp.cr
```diff
@@ -0,0 +1,317 @@
+# # TODO: case off IO::ByteFormat::SystemEndian for struct order
```
Sija: Is this file needed? Looks like debug leftover.

benchmark.cr
```diff
@@ -0,0 +1,6 @@
+require "benchmark"
```
Sija: This file seems to be a debug leftover.

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ioft/i386-ubuntu:16.04
```
Sija: This file seems to be a debug leftover.

src/crystal/compiler_rt/i128_info.cr
```diff
@@ -0,0 +1,5 @@
+module CompilerRT
+  struct I128Info
```
Sija: Could be simplified using `record` macro. ditto elsewhere

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,122 @@
...
+fun __udivmodti4(a : UInt128, b : UInt128, rem : UInt128*)
+  n_udword_bits = sizeof(Int64) * sizeof(Char)
+  n_utword_bits = sizeof(Int128) * sizeof(Char)
+  n = a.unsafe_as(CompilerRT::U128Info)
+  d = b.unsafe_as(CompilerRT::U128Info)
+  q = CompilerRT::U128Info.new
+  r = CompilerRT::U128Info.new
+  sr = 0_u32
+
+  if (n.high == 0)
```
Sija:
```suggestion
  if n.high == 0
```

ditto all below

src/string.cr
```diff
@@ -489,13 +519,10 @@ class String
   end
 
   # :nodoc:
-  record ToU64Info,
-    value : UInt64,
-    negative : Bool,
-    invalid : Bool
+  record UIntInfo, value : (UInt64 | UInt128), negative : Bool, invalid : Bool
```
```suggestion
  record UIntInfo, value : UInt64 | UInt128, negative : Bool, invalid : Bool
```

spec/std/crystal/compiler_rt/umodti3_spec.cr
```diff
@@ -0,0 +1,20 @@
+require "spec"
+
+# Ported from compiler-rt:test/builtins/Unit/umodti3_test.c
+
+private def test__umodti3(a : UInt128, b : UInt128, expected : UInt128, file = __FILE__, line = __LINE__)
+  it "passes compiler-rt builtins unit tests" do
+    __umodti3(a, b).eq expected
```
Sija: `file` and `line` should be passed over

src/string.cr
bcardiff: It's better to keep a `ToU64Info` and `ToU128Info` so parsing does not always involve a 128 bits integer as a representation and 64 bits are used when possible.

Or a generic can be used if preferred, but I think two structs will be good enough.

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -1542,57 +1542,61 @@ module Crystal
...
         int64max = Int64::MAX.to_u64
         int64max += 1 if negative
 
-        int32max = Int32::MAX.to_u32
-        int32max += 1 if negative
-
-        if int_value > int64max
-          @token.number_kind = :u64
-        elsif int_value > int32max
-          @token.number_kind = :i64
```
Sija: Hmm, what's the reason for removal?

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -1542,57 +1542,61 @@ module Crystal
...
-        raise_value_doesnt_fit_in "Int64", string_value, start
+    def check_negative_value_fits_in_int128(string_value, num_size, start)
+      if num_size > 39
+        raise_value_doesnt_fit_in "Int128", string_value, start
       end
 
-      if num_size == 19
+      if num_size == 39
         i = 1 # skip '-'
         "9223372036854775808".each_byte do |byte|
```
Sija: Shouldn't this value be updated?

src/compiler/crystal/syntax/lexer.cr
```diff
@@ -1601,17 +1605,17 @@ module Crystal
...
-        raise_value_doesnt_fit_in "UInt64", string_value, start
+    def check_value_fits_in_uint128(string_value, num_size, start)
+      if num_size > 40
+        raise_value_doesnt_fit_in "UInt128", string_value, start
       end
 
-      if num_size == 20
+      if num_size == 40
         i = 0
         "18446744073709551615".each_byte do |byte|
```
Sija: ditto

spec/std/crystal/compiler_rt/muloti4_spec.cr
wontruefree: I agree with @bcardiff

src/compiler/crystal/syntax/lexer.cr
wontruefree: the max is now 128 so the line is here now.
https://github.com/crystal-lang/crystal/pull/7093/files#diff-627df40ff6c553c599dce26c4eeb3fb3R1576

Sija: IIUC int type should be determined dynamically

wontruefree: can you give an example.  This seems like an update of the int max?

bcardiff:
@wontruefree maybe it will make sense to split this PR in 3 parts.

1. for the port of compiler-rt, bringing uniform support for overflow and int128 in all platforms by default.
2. support for parsing int128 numbers (probably by introducing ToU64Info and ToU128Info instead of the union)
3. bring support to the compiler/parser itself to handle 128 bits literals

I think this order will make things easier to review and merge (also, big thanks for porting compiler-rt with the current design 🙇 )

WDYT?

wontruefree:
@bcardiff I like this idea.  I think this PR got more and more expansive then I thought.  It is kind of hard to verify all this works without getting to run in crystal.

src/crystal/compiler_rt/i128_info.cr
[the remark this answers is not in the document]
wontruefree: I didnt know this existed in crystal.  `@[Extern]` is way better.

PR #7094, opened 2018-11-19
compiler: Introduce non-nilable read_file macro

woodruffw:
Follow-up on #6967.

The original (nilable) macro has been renamed to `read_file?`.

### DIFF
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -65,6 +65,8 @@ module Crystal
         interpret_raise(node)
       when "read_file"
         interpret_read_file(node)
+      when "read_file?"
+        interpret_read_file(node, nilable: true)
       when "run"
         interpret_run(node)
       else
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/macros/methods.cr
```diff
@@ -198,16 +200,18 @@ module Crystal
...
       end
 
       node.args[0].accept self
       filename = @last.to_macro_id
-      if File.file?(filename)
+
+      begin
         @last = StringLiteral.new(File.read(filename))
-      else
+      rescue e
```
Sija: `e` -> `ex` plz

yxhuvud: Also, one question is if it is a good idea to rescue all possible errors instead of only the specific one that is raised if the file doesn't exist.

Sija: @yxhuvud I don't see a problem with it since they're re-raised and the only meaningful call (which might raise) is `File.read`.

woodruffw: > and the only meaningful call (which might raise) is `File.read`

Yep, this was my reasoning here :slightly_smiling_face:

RX14:
We should change the default behaviour of `read_file` to return nil. From my experience, all other macro methods return nil instead of raising in the "missing" case. Since the macro language is dynamically typed, returning nil instead of erroring is the better default.

woodruffw:
> all other macro methods return nil instead of raising in the "missing" case

Two counterexamples: Both the `system` and `run` macros raise on their error cases (failed execution and missing file, respectively). I'd argue that `read_file` bears more resemblance in kind to these two macros than others, and so should raise like them.

Similarly, from the end user's perspective, I'd argue that raising an error is a better default: returning a `nil` literal converts an I/O failure into a type mismatch, or (worse!) silently does the wrong thing if the user only ever calls `#to_s` or any other quack-y calls. That's avoidable with the `|| raise` workaround, but I think doing it by default is more programmer-friendly.

ysbaddaden:
@RX14 the `run` macro raises if the executed program exits with a non zero status.

If I embed a file, I expect the file to exist and be readable. I want a compile time error with the actual reason if it failed to. I don't want to deal with a nilable or have to raise myself a redundant `read_file("path/to/file.png") || raise "failed to embed `path/to/file.png`"` when the macro just could do it and give me the actual reason.

PR #7097, opened 2018-11-20
Format: prevent conflicting between nested tuple types and macro expressions

### LINKED ISSUE #7095: Formatter removes space between double brackets, causing code to not compile
The formatter changes this (valid program)

```crystal
class Foo
  getter matrix : { {Int32, Int32}, {Int32, Int32} }

  def initialize
    @matrix = { {1,0}, {0,1} }
  end
end

Foo.new
```

into this (invalid program)

```crystal
class Foo
  getter matrix : {{Int32, Int32}, {Int32, Int32}}

  def initialize
    @matrix = { {1,0}, {0,1} }
  end
end

Foo.new
```

This is quite a bit like #6299, except I can't use the workaround suggested there since a type declaration like this can't be split across multiple lines.

MakeNowJust:
Fixed #7095

Now `crystal tool format` keeps the following file format:

```crystal
foo : { {Int32, String} }
```

### DIFF
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -1107,12 +1107,12 @@ module Crystal
       if first_name == "Tuple" && @token.value != "Tuple"
         write_token :"{"
         found_comment = skip_space_or_newline
-        wrote_space_at_end = false
+        write_space_at_end = false
         node.type_vars.each_with_index do |type_var, i|
           # This is to prevent writing `{{` and `{%`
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/tools/formatter.cr
```diff
@@ -1106,14 +1106,22 @@ module Crystal
       # Check if it's {A, B} instead of Tuple(A, B)
       if first_name == "Tuple" && @token.value != "Tuple"
         write_token :"{"
-        skip_space_or_newline
+        found_comment = skip_space_or_newline
+        wrote_space_at_end = false
```
straight-shoota: Shouldn't this be called `write_space_at_end`?

vladfaust:
Thank you, @MakeNowJust!

PR #7109, opened 2018-11-24
Implement ArrayLiteral pop and shift

### LINKED ISSUE #6023: Feature request: ArrayLiteral#pop & #shift
`ArrayLiteral#push` works (although not documented), so I suppose it's possible to add `#pop` and `#shift` etc.

*Upd*
Nice place to start: https://github.com/crystal-lang/crystal/pull/5265

Blacksmoke16:
Resolve #6023

### DIFF
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -646,8 +646,16 @@ module Crystal::Macros
     def unshift : ArrayLiteral
     end
 
+    # Similar to `Array#shift`
+    def shift(n : NumberLiteral) : ArrayLiteral
+    end
+
     # Similar to `Array#push`.
     def push(value : ASTNode) : ArrayLiteral
+    end
+
+    # Similar to `Array#pop`
+    def pop(n : NumberLiteral) : ArrayLiteral
     end
 
     # Similar to `Array#<<`.
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -643,7 +643,7 @@ module Crystal::Macros
     end
 
     # Similar to `Array#unshift`.
-    def unshift : ArrayLiteral
+    def unshift(value : ASTNode) : ArrayLiteral
     end
 
     # Similar to `Array#shift`
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -661,8 +661,48 @@ module Crystal
         assert_macro "", %({% x = [1]; x.unshift(2); %}{{x}}), [] of ASTNode, %([2, 1])
       end
 
+      it "executes shift without a parameter" do
+        expect_raises(Crystal::TypeException, "Error in line 1: wrong number of arguments for shift (given 0, expected 1)") do
+          assert_macro "", %({% x = [1,2,3].shift %}{{x}}), [] of ASTNode, %([1])
+        end
+      end
+
+      it "executes shift invalid parameter type" do
+        expect_raises(Crystal::TypeException, "Error in line 1: argument to shift must be a number, not StringLiteral") do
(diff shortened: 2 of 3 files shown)


vladfaust:
Why accepting NumberLiteral arguments only? 😕

src/compiler/crystal/macros.cr
```diff
@@ -646,10 +646,18 @@ module Crystal::Macros
     def unshift : ArrayLiteral
     end
 
+    # Similar to `Array#shift`
+    def shift(n : NumberLiteral) : ArrayLiteral
```
Sija:
```suggestion
    def shift(n : ASTNode) : ArrayLiteral
```

src/compiler/crystal/macros.cr
```diff
@@ -646,10 +646,18 @@ module Crystal::Macros
...
+    end
+
     # Similar to `Array#push`.
     def push(value : ASTNode) : ArrayLiteral
     end
 
+    # Similar to `Array#pop`
+    def pop(n : NumberLiteral) : ArrayLiteral
```
Sija:
```suggestion
    def pop(n : ASTNode) : ArrayLiteral
```

Sija:
@vladfaust good catch, it should be `ASTNode` all the way.

asterite:
Why `ASTNode`? 🤔

It's fine as `NumberLiteral`.

Blacksmoke16:
@Sija Why tho?  `Array#shift` and pop are limited to integers.  `[1,2,3].pop("foo")` wouldn't make sense.

vladfaust:
Yeah, my bad. This PR is okay

Sija:
@Blacksmoke16 Sorry, I misread (was thinking about `#unshift`)

src/compiler/crystal/macros.cr
```diff
@@ -646,10 +646,18 @@ module Crystal::Macros
     def unshift : ArrayLiteral
```
Sija: Shouldn't `#unshift` accept an argument?

Blacksmoke16: It does but just isn't documented it looks like.

Sija: So it should - question is whether in this PR or the new one?

Blacksmoke16: fixed in 201c834

vladfaust: I'd put into another PR, because one is about adding new features, and another is a fix

bcardiff:
I think we should pursuit to remove the mutation methods over literal values. I stated that before.

The very nature of the mutation force a dependency of the execution order. If we aim for modular/incremental compilation that dependency would need to be tracked.

One of the first mutations `ArrayLiteral#push` introduced were to support registry-like feature. Pushing classes to an array for later iterate in a macro. This pattern can be substituted with other constructs (that don't exist yet) using annotations: "enumerate classes that has certain annotation applied".

I have see the usage of mutating literal like that. Append this for later iteration.

I would avoid adding more mutation if the reason is because other mutation operations exists.

Another benefit of the patter that use annotations is that it _would_ even be sound to semantically treat them as a declaration instead of an statement. That is, if a classes is later created with that annotation, the macro could be reevaluated; hence, there is no dependency of the order.

Blacksmoke16:
@bcardiff :+1: works for me.  I'll make another PR for the current missing arg on `unshift` then.

bcardiff:
I'm confused @Blacksmoke16 . Probably you have a use case that you need to unshift or pop the ArrayLiteral. Would you mind sharing it?

I get that adding an argument might be smaller change than adding an argument, but at the end of the day you seem to be needing something that I am not vouching in supporting. Again, can you share the use case?

vladfaust:
My use-case was an ORM where I'd need to store meta data about models in a compile-time Array. Since I'm moving to annotations as well, I don't need this functionality either.

PR #7118, opened 2018-11-27
change readme template

mamantoha:
Before:
screenshot_20181127_141628

After:
screenshot_20181127_141806

### DIFF
--- a/src/compiler/crystal/tools/init/template/readme.md.ecr
+++ b/src/compiler/crystal/tools/init/template/readme.md.ecr
@@ -6,11 +6,13 @@ TODO: Write a description here
 
 <%- if config.skeleton_type == "lib" -%>
 1. Add the dependency to your `shard.yml`:
-```yaml
-dependencies:
-  <%= config.name %>:
-    github: <%= config.github_name %>/<%= config.name %>
-```
+
+   ```yaml
+   dependencies:
+     <%= config.name %>:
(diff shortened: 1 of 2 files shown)


r00ster91:
You need to change the specs too: https://github.com/crystal-lang/crystal/blob/master/spec/compiler/crystal/tools/init_spec.cr.

src/compiler/crystal/tools/init/template/readme.md.ecr
```diff
@@ -6,11 +6,13 @@ TODO: Write a description here
 
 <%- if config.skeleton_type == "lib" -%>
 1. Add the dependency to your `shard.yml`:
-```yaml
-dependencies:
-  <%= config.name %>:
-    github: <%= config.github_name %>/<%= config.name %>
-```
+
+    ```yaml
```
r00ster91: Can there be 3 spaces instead of 4 before every line? Works on GitHub and would be better aligned to the line above.
Currently it's not very good aligned to the "Add" word above by one space as you can see here:
```
1. Add the dependency to your `shard.yml`:

    ```yaml
   ^
```

mamantoha: Yes. It's working. Should I change alignment?

r00ster91: Yes, that would be good.

PR #2004, opened 2016-01-10
ECR: support comments (<%# %>) and escapes (<%% %>)

ysbaddaden:
~~WIP: it's still missing the transform specs (comments are dropped, escaped are transformed to strings, minus a %).~~

I don't make a difference between `<%%` and `<%%=` since we'd treat them exactly the same (output them as is minus a %).

closes #1999

PR #7121, opened 2018-11-27
Add comments and escapes to ECR docs

KCErb:
PR #2004 added some new features to ECR, here are the docs to go with them.

Also, I felt like the language was a little awkward in the examples. Originally the class: `Greeting` would greet a user with the phrase `Greeting, KC` but I believe it is more natural to call the class a `Greeter` and I'm sure (as a native English speaker) that the idiom is "Greetings, KC".

Sorry if that kind of input isn't welcome, I don't mean to be pedantic. I'm happy to revert that aspect of this PR if requested :)

### DIFF
--- a/src/ecr.cr
+++ b/src/ecr.cr
@@ -60,7 +60,7 @@
 #   ECR.def_to_s "greeter.ecr"
 # end
 #
-# Greeter.new(nil).to_s # => "Greetings!\n"
+# Greeter.new(nil).to_s    # => "Greetings!\n"
 # Greeter.new("Jill").to_s # => "Greetings, Jill!\n"
 # ```
 #
--- a/src/ecr.cr
+++ b/src/ecr.cr
@@ -36,7 +36,7 @@
 #   ECR.def_to_s "greeter.ecr"
 # end
 #
-# Greeter.new("John").to_s # => Greetings, John!
+# Greeter.new("John").to_s # => "Greetings, John!\n"
 # ```
 #
 # Using logical statements:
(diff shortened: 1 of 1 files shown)


src/ecr.cr
```diff
@@ -12,61 +12,98 @@
...
 #
-#   ECR.def_to_s "greeting.ecr"
+#   ECR.def_to_s "greeter.ecr"
 # end
 #
-# # greeting.ecr
-# Greeting, <%= @name %>!
-#
-# Greeting.new("John").to_s #=> Greeting, John!
+# Greeter.new("John").to_s # => Greetings, John!
```
straight-shoota: Could you wrap the result in quotes?

src/ecr.cr
```diff
@@ -12,61 +12,98 @@
...
+#
+# class Greeter
+#   def initialize(@name : String)
+#   end
+#
+#   ECR.def_to_s "greeter.ecr"
+# end
 #
-# Greeting.new(nil).to_s #=> Greeting!
+# Greeter.new(nil).to_s # => Greetings!
```
straight-shoota: Ditto: quotes.
And maybe add a usage example with a name?

src/ecr.cr
```diff
@@ -12,61 +12,98 @@
...
 #
-#   ECR.def_to_s "greeting.ecr"
+#   ECR.def_to_s "greeter.ecr"
 # end
 #
-# # greeting.ecr
-# <%- @names.each do |name| -%>
-# Hi, <%= name %>!
-# <%- end -%>
+# Greeter.new("John", "Zoe", "Ben").to_s
```
straight-shoota: ` # => "Hi, John!\nHi, Zoe!\nHi, Ben!"`

src/ecr.cr
```diff
@@ -12,61 +12,98 @@
...
+# ```
 #
-# Greeting.new("John", "Zoe", "Ben").to_s
-# #=> Hi, John!
-# #=> Hi, Zoe!
-# #=> Hi, Ben!
+# ```
+# require "ecr"
+# foo = 2
+# string = ECR.render("demo.ecr")
```
straight-shoota: There is no need for storing the result in a variable, just directly show the result. `.inspect` is not essential, either.

KCErb:
@straight-shoota Thanks for the comments, I believe I implemented them all correctly.

PR #7123, opened 2018-11-28
HTTP::Client: move `@socket` instance variable at the end of the method to avoid wrong intermediate states

### LINKED ISSUE #7120: HTTP::Client, request after IO::Timeout raises "Unsupported HTTP version"
Reproducible code: 

```crystal
require "http/client"

http = HTTP::Client.new("www.google.com", tls: true)
http.read_timeout = 0.01
loop do
  sleep 1
  http.get "/"
rescue ex : IO::Timeout
  puts ex.inspect
end
```

result: 

```
$ crystal run --release src/bug.cr
#<IO::Timeout:Read timed out>
Unhandled exception: Unsupported HTTP version: ;7[Cs*vs#0UY1D[S+ (Exception)
  from /usr/local/Cellar/crystal/0.27.0/src/http/client/response.cr:30:12 in 'exec_internal_single'
  from /usr/local/Cellar/crystal/0.27.0/src/http/client.cr:499:5 in '__crystal_main'
```

asterite:
Fixes #7120

It seems that when you create an OpenSSL socket a few bytes are sent and then read, probably to establish the SSL connection. If an exception is raised then (in this case because of the timeout) then  `@socket` is left with a non-SSL socket, and then the next request will fail. The solution is to set `@socket` to `nil` so that the next call tries to establish the connection again.

Unfortunately I can't think of a spec other than the code that reproduces it, but that expects a read-timeout to happen under a certain condition and also a known server that supports TLS.

### DIFF
--- a/src/http/client.cr
+++ b/src/http/client.cr
@@ -676,16 +676,14 @@ class HTTP::Client
     socket = TCPSocket.new hostname, @port, @dns_timeout, @connect_timeout
     socket.read_timeout = @read_timeout if @read_timeout
     socket.sync = false
-    @socket = socket
 
     {% if !flag?(:without_openssl) %}
       if tls = @tls
-        tls_socket = OpenSSL::SSL::Socket::Client.new(socket, context: tls, sync_close: true, hostname: @host)
-        @socket = socket = tls_socket
+        socket = OpenSSL::SSL::Socket::Client.new(socket, context: tls, sync_close: true, hostname: @host)
       end
     {% end %}
 
-    socket
+    @socket = socket
   end
 
   private def host_header


RX14:
How about just moving the `@socket = socket` to the end of the function?

asterite:
Good idea about moving `@socket` to the end.

No idea how to test it, though...

asterite:
I mean, something like this works as a spec:

```crystal
    it "..." do
      http = HTTP::Client.new("www.google.com", tls: true)
      http.read_timeout = 0.01

      expect_raises(IO::Timeout) do
        http.get("/")
      end

      expect_raises(IO::Timeout) do
        http.get("/")
      end
    end
```

Not sure we want to include such spec.

j8r:
A localhost `HTTP::Server` could be created for the specs.

asterite:
Sure. If someone has the time please update my PR with a spec (I won't have time, sorry)

RX14:
I tried for 3 hours and you can't spec this.

straight-shoota:
How about putting a spec in `spec/manual`?

PR #7126, opened 2018-11-29
Add BigDecimal#round(digits, mode) overload

### LINKED ISSUE #5714: Implement rounding modes for BigDecimal
In the spirit of Ruby's [`BigDecimal#round(n, mode)`](http://devdocs.io/ruby~2.5/bigdecimal#method-i-round).

"[Bankers rounding](https://en.wikipedia.org/wiki/Rounding#Round_half_to_even)" example:

ruby:
```ruby
"1.5".to_d.round(0, :half_even) # => 2.0
"2.5".to_d.round(0, :half_even) # => 2.0
```

crystal:
```crystal
# no such thing atm :(

"1.5".to_big_d.round(0) # => 2.0
"2.5".to_big_d.round(0) # => 3.0
```

Sija:
Fixes #5714

TBD:
- [x] `BigDecimal.rounding_mode`
- [x] `BigDecimal.save_rounding_mode`
- [x] `BigDecimal.with_rounding_mode`
- [ ] Implementation review
- [x] Specs
- [x] Docs

### DIFF
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -271,6 +271,20 @@ describe BigDecimal do
     end
   end
 
+  describe ".with_rounding_mode" do
+    it "sets given mode as .rounding_mode within yielded block" do
+      begin
+        BigDecimal.rounding_mode = :to_zero
+        BigDecimal.with_rounding_mode(:ties_even) do
+          BigDecimal.rounding_mode.should eq(Number::RoundingMode::TIES_EVEN)
+        end
+        BigDecimal.rounding_mode.should eq(Number::RoundingMode::TO_ZERO)
+      ensure
+        BigDecimal.rounding_mode = nil
+      end
+    end
+  end
+
   it "performs arithmetic with other number types" do
     (1.to_big_d + 2).should eq(BigDecimal.new("3.0"))
     (2 + 1.to_big_d).should eq(BigDecimal.new("3.0"))
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -203,6 +203,8 @@ describe BigDecimal do
             .should eq(BigDecimal.new(BigInt.new({{sign.id}}5), 1))
           BigDecimal.new("{{sign.id}}2.00").round(10)
             .should eq(BigDecimal.new(BigInt.new({{sign.id}}20), 1))
+          BigDecimal.new("{{sign.id}}0.00").round(-10)
+            .should eq(BigDecimal.new(BigInt.zero))
         end
 
         it "rounds :to_zero" do
@@ -220,6 +222,8 @@ describe BigDecimal do
             .should eq(BigDecimal.new(BigInt.new({{sign.id}}979797799999666)))
           BigDecimal.new("{{sign.id}}979797799999666.5").round(0, mode: :ties_even)
             .should eq(BigDecimal.new(BigInt.new({{sign.id}}979797799999666)))
+          BigDecimal.new("{{sign.id}}123456078.999").round(-3, mode: :ties_even)
+            .should eq(BigDecimal.new(BigInt.new({{sign.id}}123456000)))
         end
 
         it "rounds :to_positive" do
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -235,6 +235,42 @@ describe BigDecimal do
     {% end %}
   end
 
+  describe ".rounding_mode" do
+    it "defaults to :ties_even" do
+      BigDecimal.rounding_mode.should eq(Number::RoundingMode::TIES_EVEN)
+    end
+
+    it "allows to be set" do
+      begin
+        BigDecimal.rounding_mode = :to_zero
+        BigDecimal.rounding_mode.should eq(Number::RoundingMode::TO_ZERO)
+      ensure
+        BigDecimal.rounding_mode = nil
+      end
+    end
+
+    it "acts as default value for mode argument in #round" do
+      BigDecimal.rounding_mode = :to_positive
+      BigDecimal.new("2.5").round.should eq(BigDecimal.new(BigInt.new(3)))
+      BigDecimal.rounding_mode = nil
+      BigDecimal.new("2.5").round.should eq(BigDecimal.new(BigInt.new(2)))
+    end
+  end
+
+  describe ".save_rounding_mode" do
+    it "properly retains and resets rounding mode state" do
+      begin
+        BigDecimal.rounding_mode = :to_zero
+        BigDecimal.save_rounding_mode do
+          BigDecimal.rounding_mode = :ties_even
+        end
+        BigDecimal.rounding_mode.should eq(Number::RoundingMode::TO_ZERO)
+      ensure
+        BigDecimal.rounding_mode = nil
+      end
+    end
+  end
+
   it "performs arithmetic with other number types" do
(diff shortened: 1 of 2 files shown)


src/big/big_decimal.cr
```diff
@@ -235,6 +255,67 @@ struct BigDecimal < Number
...
+
+    bd = n_value
+    bd += 1 if round_up
+    bd = self.class.new(bd, digits)
+    bd *= -1 if negative
+    bd
+  end
+
+  # :nodoc:
+  def round(digits, mode : Symbol) : BigDecimal
```
Sija: This definition is required due to enum autocasting bug. Otherwise user is left with `BigDecimal::RoundingMode::HALF_UP` - vs - `:half_up`

spec/std/big/big_decimal_spec.cr
```diff
@@ -169,6 +169,90 @@ describe BigDecimal do
...
 
+  describe "#round" do
+    {% for sign in %w[+ -] %}
+      context "(with {{sign.id}} sign)" do
+        it "returns self if digits are equal or more than significant digits" do
+          BigDecimal.new("{{sign.id}}2.5").round(1)
+            .should eq(BigDecimal.new(BigInt.new({{sign.id}}25), 1))
+        end
+
+        it "rounds :up" do
```
r00ster91:
```suggestion
        it "rounds up" do
```
The colons can probably be removed now since it's using enums.

Sija: I prefer it this way, since it denotes passed argument.

src/big/big_decimal.cr
```diff
@@ -235,6 +257,122 @@ struct BigDecimal < Number
...
+  # NOTE: Defaults to `RoundingMode::HALF_UP` if `nil`.
+  class_property rounding_mode : RoundingMode?
+
+  # :ditto:
+  def self.rounding_mode : RoundingMode
+    @@rounding_mode || RoundingMode::HALF_UP
+  end
+
+  # Executes the provided block, preserving the rounding mode:
+  #
```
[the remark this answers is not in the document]
r00ster91: Yes, description and code example is usually seperated by one line.

Sija:
@RX14 @bcardiff @asterite would you mind a rrrreview?

RX14:
This doesn't even *attempt* to be compatible with `Number#round`. It also seems completely unoptimized, have you looked at what other implementations do here?

Sija:
> This doesn't even _attempt_ to be compatible with `Number#round`.

That's true, some goes for `Float{32,64}#round` and `Int#round` (they handle neither *ndigits*, nor *base* arguments) - it would be nice to sort 'em out, but not before having uniformly working *mode* argument - which would be better done in a separate PR.

Ruby has more uniform API in this regard, with `#round([ndigits][, mode])` signature.

> It also seems completely unoptimized, have you looked at what other implementations do here?

Unfortunately, that's quite likely, but you're free to post your optimization suggestions, let's treat it as yet another Advent of Code puzzle ;)

RX14:
I don't want to merge this before we sort out a rounding API on `Number` and make `Int` and `Float` follow it. Then we can make this PR follow that API too.

Sija:
> That's true, some goes for `Float{32,64}#round` and `Int#round` (they handle neither _ndigits_, nor _base_ arguments) - it would be nice to sort 'em out, but not before having uniformly working _mode_ argument - which would be better done in a separate PR.

I'd like to correct my statement, since `Float{32,64}#round` and `Int#round` are more like optimizations of `Number#round`, with exception of missing handling *base* argument; btw, is it really needed? Aside of *digits* and sometimes *mode*, I haven't seen much of it in stdlibs of other languages.

Now, as for cleaning up the API, I'd suggest dropping *base* argument, would leave us with consistent API of `Float`, `Int` and `Number`.

Additional (named?) argument *mode* support would stay for `BigDecimal` only. Would be nice to have it working for `Number#round`, but to my (limited) understanding, that would mean loosing `Float#round` optimization when *mode* argument is used.

Sija:
@RX14 any thoughts?

RX14:
Same as before, we need to unify the API. We should keep base.

Sija:
@RX14 yeah, you said it before, any (counter) suggestions _how_? and why in your opinion _should_ we keep base?

Sija:
Hi, what's the status of this PR?

@RX14 proposed unification of `Number` rounding API stating that:

> [...] We should keep base.

I disagree, personally I've never had a need to specify _base_ while rounding, neither I've seen it anywhere in the wild. Ruby, JavaScript, Python, PHP - they all have pretty consistent APIs regarding rounding numbers, in which _base_ is nowhere to be seen.

In terms of `Number#round` API _unification_ I'd rather vote for removal of before-mentioned _base_ argument.

`BigDecimal#round` is especially useful for money-related shards, thus its absence hurts stdlib functionality and `BigDecimal` feature set.

I'd like to continue working on it once some course of action is settled.

Sija:
Mornin' ☀️ !

I've force-pushed refactored implementation which, I believe should be a tad more performant, and so I'd like to revisit this PR if possible since this feature comes rrrly handy for these few occasions (think financial, especially crypto), thanks for the attention and have a great day there :) 🙏

Side note: Some of the GHA CI jobs (CircleCI is ✅ ) seem to use a stale cache or sth - the spec failures doesn't seem to cover for the actual code - that or... might a be a red herring since it's already just a bit l8 for me - or early for that matter 🙃

Sija:
ping, could this get a review please? /cc @straight-shoota @asterite @waj

asterite:
@Sija I don't think this has any pressure to get into 1.0. There are a bunch of more important things right now to make sure 1.0 is properly released, so please be patient until we get some time to review this and other PRs. Thank you!

straight-shoota:
This can now be refactored on top of #10413.

straight-shoota:
I don't think we should add the global `rounding_mode` property. Or at least, it should be a different discussion. This should focus on implementing the missing rounding methods for use with `Number#round`. That is (probably?) just `round_away` and `round_even`.

It might actually be best to close this and start with a fresh PR.

Sija:
@straight-shoota rebased on top of the current `master`.

straight-shoota (approved):
Please exclude `rounding_mode` setting from this PR. This warrants a dedicated discussion. If at all, it should probably be global to `Number` and not specific to `BigDecimal` (see https://github.com/crystal-lang/crystal/issues/10414#issuecomment-798794701 ff.).

Also `BigDecimal` is missing the basic rounding mode methods `#round_away` and `#round_even`.

straight-shoota (changes requested):
Misclicked on Approve

Sija:
@straight-shoota Will do, sadly I'm not sure I can give you any sensible timeframe for that.

src/big/big_decimal.cr
```diff
@@ -237,6 +237,114 @@ struct BigDecimal < Number
...
+    round_up =
+      case mode
+      in .to_zero?     then false
+      in .ties_away?   then msd >= 5
+      in .ties_even?   then msd == 5 ? n_digits[-2]?.try(&.odd?) : msd > 5
+      in .to_positive? then !negative
+      in .to_negative? then negative
+      end
+
+    value += 1 if round_up
```
HertzDevil, line 342: This addition is incorrect, since `value` is already the `BigDecimal` itself, not the mantissa:

```crystal
BigDecimal.new("123.456789").round(2, mode: :ties_away) # => 124.45
```

HertzDevil:
With #10798 merged and the part about `BigDecimal.rounding_mode` deferred, this PR can probably be closed.

straight-shoota:
Thank you both, @Sija and @HertzDevil =)

PR #7135, opened 2018-12-01
Int128: fix how LLVM::Type.const_int emit 128 literals

bcardiff:
The llvm wrapper is not able to emit 128bits ints.

Instead of going the way #7093 did of using String. This will allow to use `LLVMConstIntOfArbitraryPrecision` method that accept the number as an array of `UInt64`.

### DIFF
--- a/src/llvm/type.cr
+++ b/src/llvm/type.cr
@@ -126,7 +126,12 @@ struct LLVM::Type
   end
 
   def const_int(value) : Value
-    Value.new LibLLVM.const_int(self, value, 0)
+    if !value.is_a?(Int128) && !value.is_a?(UInt128) && int_width != 128
+      Value.new LibLLVM.const_int(self, value, 0)
+    else
+      encoded_value = UInt64[value & UInt64::MAX, (value >> 64) & UInt64::MAX]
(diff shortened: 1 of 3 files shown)


src/llvm/type.cr
```diff
@@ -126,7 +126,12 @@ struct LLVM::Type
   end
 
   def const_int(value) : Value
-    Value.new LibLLVM.const_int(self, value, 0)
+    if !value.is_a?(Int128) && !value.is_a?(UInt128) && int_width != 128
+      Value.new LibLLVM.const_int(self, value, 0)
+    else
+      encoded_value = UInt64[value & UInt64::MAX, (value >> 64) & UInt64::MAX]
```
asterite: But this allocates memory. Why not use static_array?

bcardiff: AFAIK the reference is kept in llvm side. I couldn't find where the pointer was used to duplicate the data so I assume a reference was used. I could be wrong the documentation didn't said much.

asterite: If it's kept in LLVM side then we must keep a reference on our side, otherwise the GC will collect it.

RX14: It's copied: https://github.com/llvm-mirror/llvm/blob/master/lib/Support/APInt.cpp#L103

This should go on the stack.

PR #7140, opened 2018-12-03
CI: Format before running specs

j8r:
`crystal tool format` is a lot faster than running the whole specs.
This change provides a [Fail-fast](https://en.wikipedia.org/wiki/Fail-fast) system when the syntax isn't properly formatted.

I have also seen minor deprecated syntax (shell ticks), so I replaced them too (can be dropped from the PR)

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -73,10 +73,10 @@ prepare_system() {
 }
 
 build() {
-  with_build_env './bin/crystal tool format --check samples spec src'
-  with_build_env 'make std_spec clean'
   with_build_env 'make crystal std_spec compiler_spec docs'
+  with_build_env './bin/crystal tool format --check samples spec src'
   with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
+  with_build_env 'make std_spec clean'
 }
 
 prepare_build() {
--- a/bin/ci
+++ b/bin/ci
@@ -73,10 +73,10 @@ prepare_system() {
 }
 
 build() {
+  with_build_env './bin/crystal tool format --check samples spec src'
   with_build_env 'make std_spec clean'
   with_build_env 'make crystal std_spec compiler_spec docs'
   with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
-  with_build_env './bin/crystal tool format --check samples spec src'
 }
 
 prepare_build() {
@@ -86,7 +86,7 @@ prepare_build() {
 
   # Make sure binaries from llvm are available in PATH
   on_osx brew install jq
-  on_osx OSX_LLVM_PACKAGE=`brew info --json=v1 crystal | jq '.[].dependencies | .[] | select(startswith("llvm"))'`
+  on_osx OSX_LLVM_PACKAGE=$(brew info --json=v1 crystal | jq '.[].dependencies | .[] | select(startswith("llvm"))')
   on_osx brew link --force $OSX_LLVM_PACKAGE
 
   on_tag verify_version
@@ -94,7 +94,7 @@ prepare_build() {
 
 verify_version() {
   # If building a tag, check it matches with file
-  FILE_VERSION=`cat VERSION`
+  FILE_VERSION=$(cat VERSION)
 
   if [ "$FILE_VERSION" != "$CURRENT_TAG" ]
   then
--- a/bin/ci
+++ b/bin/ci
@@ -73,10 +73,11 @@ prepare_system() {
 }
 
 build() {
-  with_build_env 'make crystal std_spec compiler_spec docs'
-  with_build_env './bin/crystal tool format --check samples spec src'
-  with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
   with_build_env 'make std_spec clean'
+  with_build_env 'make crystal'
+  with_build_env './bin/crystal tool format --check samples spec src'
+  with_build_env 'make std_spec compiler_spec docs'
+  with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
 }
 
 prepare_build() {


j8r:
@bew see https://unix.stackexchange.com/questions/5778/whats-the-difference-between-stuff-and-stuff
Both are valid, but `$()` is newer and overall recommended

bew:
@j8r interesting didn't know `` `...` `` was older than `$(...)`

bew:
BTW with your last commit you removed `std_spec` to be run before the crystal compilation, I think that's wrong too

bew:
@asterite was there a reason to run the `std_spec` target before the `crystal` target ?

straight-shoota:
Whats the reason for this? I'd prefer to keep the formatter running last since it's the least important. When a PR is run on CI I'm mostly interested in a verification that it acts *correctly* and care less whether it's properly formatted. That's easy to fix later. But it doesn't help if you have a properly formatted patch that simply doesn't work.

asterite:
@straight-shoota The thing is that the formatter runs in less than a second, but the specs take minutes. So you send your PR, you wait like 30 minutes for the build, and then you realize you have a wrong formatting. It's much better to put the formatter first so you only have to wait a few minutes to realize you missed running `crystal tool format`.

straight-shoota:
@asterite I don't understand the problem. You can simply run `crystal tool format` locally to avoid it popping up late in the CI process. It can easily be installed as a pre-commit hook to fully automate this check locally.

Since getting correctness right is much more difficult than getting the format right (it's usually just one command), I'd want that to be tested first. Even if it takes much longer than checking the formatting.

j8r:
This change is free @straight-shoota , your proposition doesn't block this one – they can complement each others. Some contributors only use GUI, or only have Crystal in a container to test and then commit locally.

straight-shoota:
> This change is free

No, it's not. It postpones critical CI steps after a less important one.

j8r:
Essentially, what does it change? At the end, all checks are made.

straight-shoota:
If you don't think it changes anything, there is no point in this PR.

You didn't state a specific motivation for this PR, so I assume your reasoning is similar to what @asterite said. You want to avoid having to wait for the long running CI steps to process before being notified about a formatter error. 
I agree, that's fairly reasonable. But the solution to this is not changing the order CI checks run but to ensure the formatter runs locally. A simple pre-commit hook and you're done. That's even much faster than waiting for push to Github and CI worker initialization!

The problem with changing the CI order, putting less-significant checks first is that a CI run would fail for insignificant errors and not even bother executing any the significant tests until the formatting error is fixed.

Sija:
@straight-shoota I agree, thing is that there's no way to include `pre-commit` hook within the repo itself. Every committer would need to add it by hand (or script, but still). I still think that CI should fail first with code, and later with formatting - which is less important.

IMO best option (if possible) would be running specs and formatter concurrently and failing fast if either errors out - letting it finish though to make results available for inspection - in case you'd like to know whether test suite passes even if the formatting is borked and CI job fails.

straight-shoota:
Consider a contributor, especially a newcomer, not familiar with formatter posts a PR with a bug fix or a new feature. And the only thing our CI system tells is: "I don't like how you format your code".  And it doesn't even bother executing it, even if it's completely valid and correct. That's bonkers.

straight-shoota:
@Sija
> there's no way to include pre-commit hook within the repo itself. 

Yes, but those who care about getting fast formatter results can add it very easy.  We could even provide one in the repo(like https://github.com/golang/go/blob/master/misc/git/pre-commit), it just needs to be copied over.
Those who don't care shouldn't be bothered with delayed spec results just because some formatting is off.

j8r:
The formatting check can be a different Job, separated from Darwin and Linux.
This way we know we haven't properly formatted, but the code looks good.

ysbaddaden:
What's more important: broken compiler/lib specs? or broken formatting? I like that CI tells me that I broke something *important* (the compiler or core/stdlib) before telling me to fix some *cosmetic* details. At least we know that a pull request *works*, and just needs some additional :heart: to please the eye.

asterite:
I see it another way: what's better, spending resources of a CI CPU passing all specs to finally realize you missed formatting, or saving those resources/energy/money if we can detect there's something wrong with the formatting?

straight-shoota:
@asterite Spending machine resources is better than spending human resources.

Sure, failing fast on a formatter error is efficient, but it's simply not effective when the goal is to test if the code works.

asterite:
My thought of process is that if I want to contribute code I'll write code and tests and check that locally. When all is fine I'll submit a PR. Then I'll wait like 30 minutes for the build to pass, and specs will most likely pass because I check that locally. And then the build will fail because I missed running `crystal tool format`. By then I might be doing something else and not have time to continue contributing that day.

On the other hand, if the formatter check is run as soon as possible I might be able to know this after, say, 5 minutes, and then I'll go, run `crystal tool format` and be done with the PR.

asterite:
Otherwise, let's just create a  PR template that says in big bold letters "PLEASE EXECUTE `crystal tool format` before submitting a pR".

luislavena:
Not familiar with GitHub checks, but is not possible setup as individual check just for formatting, separated from the *actual* build? That way, the builds can be OK, but formatting is separate and feedback about it happens sooner.

straight-shoota:
@asterite Even in the short time it takes for the format test to finish, the creator might be occupied and only return to the PR after 30 minutes or more, only to find out that the intended specs were not executed because the formatting is not correct.

@luislavena It could just be a separate job on circle, besides `test_darwin`, `test_linux` and `test_linux32`. Maybe this would be a good compromise? Format errors will be show up fast, but won't stop actual test from running.

RX14:
The problem with executing `crystal tool format` first is that the format failing then prevents the rest of the spec suite from running. I prefer the current way around, because I share @straight-shoota's view. But I think ideally they'd run in parallel and report their status independently.

bew:
The only problem with running it independently is that if the PR changes the formatter, the formatting would be from the old compiler, not the new one.. But maybe that's not a big deal, and we can simply ignore the formatting check in that case

RX14:
yeah, you'd have a format job, which built a compiler and formatted, then simply remove the format step from the current jobs.

bcardiff:
Using a separate job for the formatter could be an improvement. Github reports all final jobs in circleci independently we could have a `test_format` job that will only build the compiler and run the formatter.

Building the compiler (in development) is not the heavy part so it should run fast enough.

RX14:
If someone could PR this change to circle, it'd be great

j8r:
Separating the formatting check to its own job is the best bet - closing

PR #7145, opened 2018-12-04
Add git pre-commit hook for `crystal tool format`

straight-shoota:
This adds a pre-commit git hook for applying `git tool format --check` on the current index as mentioned in https://github.com/crystal-lang/crystal/pull/7140#issuecomment-443912421. Git hooks need to be installed manually in the local repo, but this script makes it as simple as `ln -s scripts/git/pre-commit .git/hooks`.

Currently it doesn't work properly when only one Crystal file is in the index. The command will fail silently (no commit created) but don't print any error messages. #7144 fixes that. If that doesn't get merged soon, a workaround could be added.

A similar tool exists in Golangs repository: https://github.com/golang/go/blob/master/misc/git/pre-commit

### DIFF
--- a/scripts/git/pre-commit
+++ b/scripts/git/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/sh
+#! /bin/sh
 #
 # This script ensures Crystal code is correctly formatted before committing it.
 # It won't apply any format changes automatically.
@@ -17,4 +17,9 @@ changed_cr_files=$(git diff --cached --name-only --diff-filter=ACM | grep '\.cr$
 
 [ -z "$changed_cr_files" ] && exit 0
 
-exec crystal tool format --check $changed_cr_files >&2
+if [ -x bin/crystal ]; then
+  # use bin/crystal wrapper when available to run local compiler build
+  exec bin/crystal tool format --check $changed_cr_files >&2
+else
+  exec crystal tool format --check $changed_cr_files >&2
+fi
--- a/scripts/git/pre-commit
+++ b/scripts/git/pre-commit
@@ -1,0 +1,20 @@
+#!/bin/sh
+#
+# This script ensures Crystal code is correctly formatted before committing it.
+# It won't apply any format changes automatically.
+#
+# Only staged files (the ones to be committed) are being processed, but each file is checked
+# entirely as it is stored on disc, even parts that are not staged.
+#
+# To use this script, it needs to be installed in the local git repository. For example by running
+# `ln -s scripts/git/pre-commit .git/hooks` in the root folder.
+#
+# Called by "git commit" with no arguments. The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+
+changed_cr_files=$(git diff --cached --name-only --diff-filter=A | grep '\.cr$')
+
+[ -z "$changed_cr_files" ] && exit 0
+
+exec crystal tool format --check "$changed_cr_files" >&2
--- a/scripts/git/pre-commit
+++ b/scripts/git/pre-commit
@@ -13,8 +13,8 @@
 # exit with non-zero status after issuing an appropriate message if
 # it wants to stop the commit.
 
-changed_cr_files=$(git diff --cached --name-only --diff-filter=A | grep '\.cr$')
+changed_cr_files=$(git diff --cached --name-only --diff-filter=ACM | grep '\.cr$')
 
 [ -z "$changed_cr_files" ] && exit 0
 
-exec crystal tool format --check "$changed_cr_files" >&2
+exec crystal tool format --check $changed_cr_files >&2


straight-shoota:
Sure, running it with a compiler built from the current source would be better. But that's not reasonable. Running it with the installed compiler should be sufficient for almost all use cases. It only fails when editing the formatter rules, or when the staged code is affected from a formatter change in master. Such cases are pretty rare nowadays. There are some occasional changes to the formatter now and then, but usually with a relatively small footprint. For non-compiler work this is totally fine most of the time.

It could go in a wiki, but having it in the repo makes it much easier to install. If it's not welcome here, I'll add it to the wiki, but I'd prefer to add it to the repo directly.

bcardiff:
I agree with @RX14. 

I would add this only, if the pre-commit does something when the std / std spec things are changed.

Otherwise it would only add confusion when changing the compiler and formatter.

scripts/git/pre-commit
```diff
@@ -0,0 +1,20 @@
...
+#
+# Called by "git commit" with no arguments. The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+
+changed_cr_files=$(git diff --cached --name-only --diff-filter=A | grep '\.cr$')
+
+[ -z "$changed_cr_files" ] && exit 0
+
+exec crystal tool format --check "$changed_cr_files" >&2
```
MakeNowJust: `"$changed_cr_files"` is wrong, `$chnaged_cr_files` (without wrapping quote) is correct. `$changed_cr_files` is multiline text when there are two or more changed files, and bare variable expands each lines as one command arg.

scripts/git/pre-commit
```diff
@@ -0,0 +1,20 @@
...
+# entirely as it is stored on disc, even parts that are not staged.
+#
+# To use this script, it needs to be installed in the local git repository. For example by running
+# `ln -s scripts/git/pre-commit .git/hooks` in the root folder.
+#
+# Called by "git commit" with no arguments. The hook should
+# exit with non-zero status after issuing an appropriate message if
+# it wants to stop the commit.
+
+changed_cr_files=$(git diff --cached --name-only --diff-filter=A | grep '\.cr$')
```
MakeNowJust: Why `--difff-filter=A`? It means showing only added files.
I think `--diff-filter=ACM` is right like Go's one.

MakeNowJust:
I agree with @RX14 as formatter contributor.

Updating `CONTRIBUTING.md` to mention such a script is another way. What do you think?

scripts/git/pre-commit
bew: @MakeNowJust but that won't work if there is a space in the path to one of the file..
A workaround would be to make a shell array, and expand it with `"${the_array[@]}"` which is a shell trick to generate quotes arguments for each item of `the_array`

j8r: @bew in anyway we have to avoid Bashisms (no arrays). Some of us are using Alpine/BSD

bew: How do you handle path with space then?

j8r: `git diff --cached --name-only --diff-filter=ACM | awk '/^.*\.cr$/ { gsub(" ", "\ "); print $0 }'` do the trick - spaces need to be escaped on UNIX

MakeNowJust: It is hard problem. Escaping space by hand is not so hard. `sed` is useful, but, we should escape any other **shell special characters** even if space is escaped. It is hard.  Additionally, filename may have newline, it breaks one-line-one-filename rule. We can use `-z` option of `git diff`, but handling null character on POSIX shell is so hard (almost impossible). Otherwise, `git diff --name-only ...` output is changed by `core.quotepath` and some git configurations, so we should set correct values if consistent results are needed.

In the real world, when you add a file whose name contains spaces to this repository? I never do, and I believe you never too.

It is not valuable that we mind this problem. In fact Go's `pre-commit` hook script does not pay attention for it.

It is terrible when CI script is broken. But, when `pre-commit` script is broken, we can skip it via `-n` option.

This `"$changed_cr_files"` is **absolutely broken**. It must not work against multiple staged files. It should be fixed.

MakeNowJust: Of course we should add escaping space trick if we add path with space file in near future.

j8r: Agree @MakeNowJust . In crystal/crystal-lang, we'll never have files containing spaces (or I hope, because there is no point to).
I the project absolutely need a space in a file name, a trick could be reconsidered.

straight-shoota: I don't think we need to care about pathnames including spaces or special characters. Yes, there are possible flaws but it's a non-trivial problem (sadly).

I'm sure such paths wouldn't be approved in the Crystal repo anyway. So there is no need to bother validating them in a pre-commit hook.

MakeNowJust: @straight-shoota It has trivial problem for now. You should fix.

straight-shoota:
> Updating `CONTRIBUTING.md` to mention such a script is another way. What do you think?

Sure, that would be a good place to mention it, alongside the book and maybe in the wiki somewhere.

The question is only, what should people have to do in order to install it? `ln -s scripts/git/pre-commit .git/hooks` or `curl https://raw.githubusercontent.com/crystal-lang/crystal/92ca0c03fed623d6645e3059fc12a8b2e45a9e8b/scripts/git/pre-commit > .git/hooks/pre-commit && chmod +x .git/hooks/pre-commit` (the URL would obviously change, maybe just a gist?).

straight-shoota:
I've created a gist at https://gist.github.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6 and #7617 adds it to `CONTRIBUTING.md`.

Issue #7142, opened 2018-12-04
More Iterator methods: slice_before, slice_after, slice_when, chunk_while


asterite:
While coding this year's [Advent of Code](https://adventofcode.com/2018) it would have been nice to have these `Iterator` methods. They exist in Ruby but are missing in Crystal. They are sometimes useful in this type of problems, where using Crystal is more fun :-P

Is there interest in these? I can submit a PR with a commit for each of these.

I guess the names can remain the same as in Ruby.

mamantoha:
While coding the fifth day of Advent of Code it would have been nice to have `String#swapcase` and `Char.swapcase` methods. Yeah 🙂

asterite:
@mamantoha Hehe, indeed. At this point, given that `swpacase`'s implementation is trivial, I'd say we could add it just fine.

yxhuvud:
`to_h(&:block)` from Ruby 2.6 could be nice to have too, if we are window shopping ruby methods :)

straight-shoota:
Please keep in mind though that a time travelling Santa solving elves puzzles is not entirely representing *real world* use cases.

Just to put some restraints on the shopping spree ;)

asterite:
This is all done now :-)

(open a separate issue if you'd like `swapcase`, I'm still not sure it's that useful)

PR #7146, opened 2018-12-04
Add `Iterator#slice_after`

asterite:
Part of #7142

### DIFF
--- a/src/iterator.cr
+++ b/src/iterator.cr
@@ -1290,4 +1290,91 @@ module Iterator(T)
       self
     end
   end
+
+  # Returns an iterator for each chunked elements where the ends
+  # of chunks are defined by the block, when the block's value
+  # is _truthy_. Thus, the enumerable is sliced just **after** each
+  # time the block returns a _truthy_ value.
+  #
+  # For example, to get chunks that end at each uppercase letter:
+  #
+  # ```
+  # ary = ['a', 'b', 'C', 'd', 'E', 'F', 'g', 'h']
+  # iter = ary.slice_after(&.uppercase?)
+  # iter.next # => ['a', 'b', 'C']
+  # iter.next # => ['d', 'E']
+  # iter.next # => ['F']
+  # iter.next # => ['g', 'h']
+  # iter.next # => Iterator::Stop
+  # ```
+  #
+  # By default, a new array is created and yielded for each slice when invoking `next`.
+  # * If *reuse* is `false`, the method will create a new array for each chunk
+  # * If *reuse* is `true`, the method will create a new array and reuse it.
+  # * If *reuse* is an `Array`, that array will be reused
+  #
+  # This can be used to prevent many memory allocations when each slice of
+  # interest is to be used in a read-only fashion.
+  def slice_after(reuse : Bool | Array(T) = false, &block : T -> B) forall B
+    SliceAfter(typeof(self), T, B).new(self, block, reuse)
+  end
+
+  # :nodoc:
+  class SliceAfter(I, T, B)
+    include Iterator(Array(T))
+
+    def initialize(@iterator : I, @block : T -> B, reuse)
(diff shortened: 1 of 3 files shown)


asterite:
Note that this PR is complete by itself. I will add the other methods `slice_before`, `slice_when` after this one is merged. I might reuse some code between them, but I think adding one PR for each method will be easier to review.

bew:
I don't understand why it's called `slice_after` based on what it does.. I seems better to me to call it `slice_until`, as it'll add things to the slice **until** the block is true.

asterite:
@bew After it finds the value it "slices" the enumerable. Or put another way, it slices the enumerable after each time returns a truthy value.

bew:
Ooh ok I get it, is it possible to add an explanation like this in the doc? I think it'd help memorize the naming logic (and probably help with the other iterator you're planning to add)

bcardiff:
An spec for why the reuse is not `Bool | Array(B)` might be worth it ;-)

asterite:
Yes, I thought about putting a restriction to `reuse`, but none of the other methods have that restriction. I'll add it just for this one and in a separate PR I'll add the rest of the restrictions.

asterite:
Updated:
- put a restriction on `reuse`
- document `reuse`

asterite:
Also added a small explanation to why it's called `slice_after`.

src/iterator.cr
```diff
@@ -1290,4 +1290,92 @@ module Iterator(T)
...
+  def slice_after(reuse : Bool | Array(T) = false, &block : T -> B) forall B
+    SliceAfter(typeof(self), T, B).new(self, block, reuse)
+  end
+
+  # :nodoc:
+  class SliceAfter(I, T, B)
+    include Iterator(Array(T))
+
+    def initialize(@iterator : I, @block : T -> B, reuse)
+      @values = [] of T
```
bew: if `reuse` is an `Array`, it'll still create an empty array here for nothing :(

asterite: Good catch! Fixed! That's the code I originally had before adding all the `reuse` logic. Thank you!

bcardiff:
For the record: removing the restriction is not possible (or not worth the effort).

src/iterator.cr
```diff
@@ -1290,4 +1290,91 @@ module Iterator(T)
...
+  class SliceAfter(I, T, B)
+    include Iterator(Array(T))
+
+    def initialize(@iterator : I, @block : T -> B, reuse)
+      @end = false
+      @clear_on_next = false
+
+      if reuse
+        if reuse.is_a?(Array)
+          @values = reuse
```
RX14: Shouldn't `@clear_on_next = true` here?

asterite: Hmm... you are supposed to pass an empty array as a reuse value. It's not documented what happens if you pass an array that has some values already, but in this case they will be part of the first chunk. Maybe that's good. I wouldn't worry about this now.

RX14: The obvious usecase for passing an array as `reuse` is when you've just used the array from another iterator, meaning the array might not be empty. I don't think it's good to pass the burden of calling clear onto the caller, since it's easy to forget here. Also, `clear` on an already-empty array is very cheap, so why not?

asterite: What if you want to get chunks but want the first chunk to have some elements beforehand? If the iterator calls `clear` there's no way you can do that.

I think we are just inventing use cases. Whenever I used `reuse` I passed a boolean. We can maybe remove the ability to pass an array altogether.

RX14: If you want the first chunk to have extra elements:
1. that's pretty inflexible, since you can only do it with the first chunk, not the last
2. using a named arg called `reuse: ` is not the most clear way to achieve this.

For me, `reuse` is for avoiding memory allocations only.

PR #7150, opened 2018-12-05
Add Enumerable#to_h(&block)

Sija:
Re: 🎄 shopping spree at Ruby 2.6 land ;)

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1292,16 +1292,15 @@ module Enumerable(T)
     end
   end
 
-  # Creates a `Hash` out of an returned from a *block* Enumerable, where each
-  # element is a 2 element structure (for instance a `Tuple` or an `Array`).
+  # Creates a `Hash` out of `Tuple` pairs (key, value) returned from the *block*.
   #
   # ```
   # (1..3).to_h { |i| {i, i ** 2} } # => {1 => 1, 2 => 4, 3 => 9}
   # ```
   def to_h(&block : T -> Tuple(K, V)) forall K, V
     each_with_object({} of K => V) do |item, hash|
-      item_pair = yield item
-      hash[item_pair[0]] = item_pair[1]
+      key, value = yield item
+      hash[key] = value
     end
   end
 end
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1291,4 +1291,18 @@ module Enumerable(T)
       hash[item[0]] = item[1]
     end
   end
+
+  # Creates a `Hash` out of an returned from a *block* Enumerable, where each
+  # element is a 2 element structure (for instance a `Tuple` or an `Array`).
+  #
+  # ```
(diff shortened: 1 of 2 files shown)


Sija:
@asterite spot on, fixed.

Sija:
I was thinking of that but then it would have to be `Array(X | Y) | Tuple(X, Y)`.

asterite:
Why Array? Tuple is enough. If you have an Array as the block value the caller can always convert that to a duple. Enforcing tuple guarantees a minimum level of performance.

Sija:
Fair 'enuff, pushed an update.

src/enumerable.cr
```diff
@@ -1291,4 +1291,17 @@ module Enumerable(T)
...
+
+  # Creates a `Hash` out of an returned from a *block* Enumerable, where each
+  # element is a 2 element structure (for instance a `Tuple` or an `Array`).
+  #
+  # ```
+  # (1..3).to_h { |i| {i, i ** 2} } # => {1 => 1, 2 => 4, 3 => 9}
+  # ```
+  def to_h(&block : T -> Tuple(K, V)) forall K, V
+    each_with_object({} of K => V) do |item, hash|
+      item_pair = yield item
```
asterite: What do you think about unpacking this as `key, value = yield`?

src/enumerable.cr
```diff
@@ -1291,4 +1291,17 @@ module Enumerable(T)
       hash[item[0]] = item[1]
     end
   end
+
+  # Creates a `Hash` out of an returned from a *block* Enumerable, where each
+  # element is a 2 element structure (for instance a `Tuple` or an `Array`).
```
asterite: The docs are now outdated. Also it reads a bit strange "out of an returned"

Sija: > Creates a `Hash` out of `Tuple` pairs (key, value) returned from *block*.

sounds good?

asterite: Yes! Maybe saying that the first element of the tuple will be used as a key, and the second as a value, though I think this will be clear from the example. Alternatively check the Ruby docs (but they will probably mention Array).

Sija: Fixed.

src/enumerable.cr
```diff
@@ -1291,4 +1291,16 @@ module Enumerable(T)
       hash[item[0]] = item[1]
     end
   end
+
+  # Creates a `Hash` out of `Tuple` pairs (key, value) returned from the *block*.
+  #
+  # ```
+  # (1..3).to_h { |i| {i, i ** 2} } # => {1 => 1, 2 => 4, 3 => 9}
+  # ```
+  def to_h(&block : T -> Tuple(K, V)) forall K, V
```
straight-shoota: The block type you're using is actually `T, self -> Tuple(K, V)`.

PR #7152, opened 2018-12-06
Add `Iterator#slice_before`

asterite:
Part of #7142

This also improves the documentation of `slice_after`, now the two docs are similar and make it more clear why they are called that way.

I thought about reusing the implementation of `slice_before` but it might not be worth it. The algorithms need different variables. Plus at least in Ruby I think they don't seem to reuse the algorithm either. (in Ruby it's also much, much longer, in Crystal the core algorithm is just 30 lines)

### DIFF
--- a/src/iterator.cr
+++ b/src/iterator.cr
@@ -1291,15 +1291,14 @@ module Iterator(T)
     end
   end
 
-  # Returns an iterator for each chunked elements where the ends
-  # of chunks are defined by the block, when the block's value
-  # is _truthy_. Thus, the enumerable is sliced just **after** each
-  # time the block returns a _truthy_ value.
+  # Returns an iterator over chunks of elements, where each
+  # chunk ends right **after** the given block's value is _truthy_.
   #
   # For example, to get chunks that end at each uppercase letter:
   #
   # ```
   # ary = ['a', 'b', 'C', 'd', 'E', 'F', 'g', 'h']
+  # #                   ^         ^    ^
   # iter = ary.slice_after(&.uppercase?)
   # iter.next # => ['a', 'b', 'C']
   # iter.next # => ['d', 'E']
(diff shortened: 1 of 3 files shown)


asterite:
`ary` is short for `Array`. Ruby uses it in its examples. For example: https://ruby-doc.org/core-2.2.0/Range.html#method-i-bsearch

It also happens to be my name :-P

src/iterator.cr
```diff
@@ -1291,15 +1291,14 @@ module Iterator(T)
     end
   end
 
-  # Returns an iterator for each chunked elements where the ends
-  # of chunks are defined by the block, when the block's value
-  # is _truthy_. Thus, the enumerable is sliced just **after** each
-  # time the block returns a _truthy_ value.
+  # Returns an iterator for each chunked elements where each chunk
```
RX14: Perhaps

>Returns an iterator over chunks of elements, where each chunk ends right after the given block's value is _truthy_

src/iterator.cr
```diff
@@ -1377,4 +1376,96 @@ module Iterator(T)
       self
     end
   end
+
+  # Returns an iterator for each chunked elements where each chunk
+  # ends right **before** the given block's value is _truthy_.
```
RX14: either bold before above or remove it here

src/iterator.cr
```diff
@@ -1377,4 +1376,96 @@ module Iterator(T)
...
+        if value.is_a?(Stop)
+          @end = true
+          if @values.empty?
+            return stop
+          else
+            return @reuse ? @values : @values.dup
+          end
+        end
+
+        if @block.call(value) && !@values.empty?
```
RX14: `!@values.empty?` is cheaper and should go before the `@block.call`. The optimizer won't be able to reorder as it won't be able to prove that `@block` doesn't modify `@values`.

src/iterator.cr
asterite: Done!

src/iterator.cr
asterite: Done!

src/iterator.cr
asterite: Good idea!

RX14:
Might be good to hold off merging this until we reach consensus on what happens with non-empty `reuse: ` though.

asterite:
But we have `reuse` in many other methods and they behave the same.

We can probably tackle the `reuse` "issue" in a separate discussion.

asterite:
Plus we already merged `slice_after`

PR #7155, opened 2018-12-06
Replace inappropriate uses of `Time.now`

straight-shoota:
For measuring elapsed time, `Time.measure` should be used. In applications, where a time zone won't be used anyway, it's better to use `Time.utc_now` instead of `Time.now` to avoid loading a time zone which can be expensive (loading from disk) when it is completely unnecessary.

### DIFF
--- a/samples/red_black_tree.cr
+++ b/samples/red_black_tree.cr
@@ -384,24 +384,26 @@ class RedBlackTreeRunner
 end
 
 def bench(name, n = 1)
-  t = Time.now
-  print "#{name}: "
   res = nil
-  n.times do
-    res = yield
+  elapsed = Time.measure do
+    print "#{name}: "
+    res = nil
+    n.times do
+      res = yield
+    end
   end
-  puts "#{Time.now - t}, res: #{res}"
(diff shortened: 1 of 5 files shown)


straight-shoota:
@asterite Sure, the method body of `#measure` could just be inlined directly:
```cr
start = Time.monotonic
do_stuff
elapsed = Time.monotonic - start
```
This would be closer to the previous usage of `Time.now` but with a proper clock.

I'd prefer to use `.measure`, but `.monotonic` is also possible.

ysbaddaden:
@asterite really?

ysbaddaden:
OK, I understand.

samples/red_black_tree.cr
```diff
@@ -384,24 +384,26 @@ class RedBlackTreeRunner
 end
 
 def bench(name, n = 1)
-  t = Time.now
```
r00ster91: The changes to this file are actually already done in #6454, [samples/red_black_tree.cr](https://github.com/crystal-lang/crystal/pull/6454/files#diff-f08c2a87edafd9e94150222141b6f45a).

PR #7158, opened 2018-12-07
Fix 'NamedTupleLiteral#[]` to raise an error for invalid key type

MakeNowJust:
Ref #7153

And also update doc.

 `NamedTupleLiteral#[]=` denies key typed neither `StringLiteral`, `SymbolLiteral` nor `MacroId` already. I think `NamedTuple#[]` should do same.

### DIFF
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -937,7 +937,7 @@ module Crystal
           when StringLiteral
             key = key.value
           else
-            return NilLiteral.new
+            raise "argument to [] must be a symbol or string, not #{key.class_desc}:\n\n#{key}"
           end
 
           entry = entries.find &.key.==(key)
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/macros/methods.cr
```diff
@@ -937,7 +937,7 @@ module Crystal
           when StringLiteral
             key = key.value
           else
-            return NilLiteral.new
+            raise "argument to [] must be a symbol or string, not #{key.class_desc}:\n\n#{key}"
```
Sija: I'd suggest avoiding newlines in exception messages if possible.

MakeNowJust: https://github.com/crystal-lang/crystal/pull/7153#discussion_r239688064

asterite:
I don't think this is right. With this, there's no way to know whether a key is present or not in the named tuple literal.

RX14:
Guess we need `[]?` lol

asterite:
Maybe. Or just returning `nil` in macros is fine. It's done in many other places (like in `Annotation#[]`)

Sija:
AFAIK this exception is raised only when wrong type of node passed as *key*, value can still be `nil`.

asterite:
@Sija oh, right, I looked too quickly. Thanks!

PR #7159, opened 2018-12-07
Add `Iteratory#slice_when` and `Iterator#chunk_while`

asterite:
Part of #7142

Having both `slice_when` and `chunk_while`, given that the difference is only whether the condition is inverted or not, is debatable. However, one might be more readable than other depending on each case. Same as why `while`/`until`, `if`/`unless` exist.


RX14:
Merge conflicts!

asterite:
@RX14 rebased

PR #7165, opened 2018-12-09
Add missing Float#weeks method similar to Int#weeks

vlazar:
Both `Int` and `Float` have `days`, `hours`,`minutes`, `seconds`, `milliseconds`, `microseconds`, `nanoseconds`.

On top of that `Int` also has `weeks` method, but `Float` does not.

### DIFF
--- a/spec/std/time/span_spec.cr
+++ b/spec/std/time/span_spec.cr
@@ -277,5 +277,6 @@ describe Time::Span do
     1.hours.should eq(60.minutes)
     1.week.should eq(7.days)
     2.weeks.should eq(14.days)
+    1.1.weeks.should eq(7.7.days)
   end
 end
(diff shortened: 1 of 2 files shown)


r00ster91:
Isn't `Float#months` and `Float#years` missing as well?

vlazar:
@r00ster91 You are right, but these return `Time::MonthSpan`, not `Time::Span` like all other methods (`weeks` .. `nanoseconds`) and `Time::MonthSpan` stores values as `Int64`, so unless that's changed the `months` and `years` for `Float` wouldn't make much sense.

vlazar:
Maybe there is a reason behind `Time::MonthSpan` not using floats. And if not, changing `Time::MonthSpan` implementation feels like a bigger change than just adding missing method(s). I guess it's better to discuss it first and address in separate PR. But I'm happy to address it here, if there is a need for it of course.

straight-shoota:
`MontSpan` can't express a fraction of a month, that's why float months and years don't make much sense. 
This requires an appropriate datatype, see #6522.

PR #7173, opened 2018-12-10
Add Int128, UInt128 to arithmetics_spec

bcardiff:
Some big numbers were (unintentionally) left out of the specs


asterite:
@bcardiff Int128 support doesn't exist for 32 bits. That's why I never continued work on it, and I think the easiest way right now is to completely drop its support. Otherwise we have to do some juggling to include a small runtime for the missing functions (`__udivti3`, `__umodti3`, etc.). They seems to be defined [here](https://github.com/llvm-mirror/compiler-rt), for example:

https://github.com/llvm-mirror/compiler-rt/blob/39adb65b17d62b2bd0742e454ebb6fd923bdbf79/lib/builtins/multi3.c#L45-L56

bcardiff:
I've been digging this for the last couple of hours. compiler-rt seems to have support for 32-bits 

> https://bugs.swift.org/browse/SR-3472
>
> compiler-rt supports 32-bit Linux.

but I am not finding how to link to a prebuilt library, not to build with a 32 bits environment with the missing symbols.

So we just spec 128 bits on 64 bits, right?
128 bits will come, sooner or later. As long as we have specs for 32 bits they will avoid us breaking the compiler for 32 bits even if 128 bits are supported for 64 bits systems.

asterite:
Yes, we should probably use `compiler-rt` but I'm not sure how to cleanly do that.

bcardiff:
The CI is "green". GTG?

PR #7179, opened 2018-12-11
Add begin-less and end-less ranges

### LINKED ISSUE #7170: [RFC] Endless ranges
Ruby 2.6 introduces endless ranges:

```ruby
[1, 2, 3, 4, 5][2..] # => [3, 4, 5]
```

I think it's a small nice addition. It removes the need to write `-1` and the confusion of `2..-1` vs `2...-1` (in the above snippet `2..` and `2...` would give the same result).

Ruby implements this by having the end of the range be `nil`. We could do the same.

The only "ugly" thing about it is that you always have to write that range inside parentheses, brakcets, or another expression (you can't just write `a = 1..` because the parser will wait for the end value, and a newline makes it still expect the end value). But one can write `a = 1..nil` so it's a minor issue (plus these ranges are mostly useful in nested positions).

I don't think it's too hard to implement this.

Thoughts?

(side note, I also thought about having `..3` meaning "from the start", but it's not that useful because one could always put a zero there, it's always zero, and then there's the confusion of which zero is it (int or float?))

asterite:
Fixes #7170

This PR introduces begin-less and end-less ranges. If most of us think this is a nice and useful feature we could merge it.

An end-less range is just a range with `nil` as its end. It can be written as `(3..)` or `(3..nil)`, but of course without an explicit `nil` is nicer.

A begin-less range is just a range with `nil` as its beginning. It can be written as `..3` or `nil..3`.

These ranges have two semantics:

## As a range

A range has methods like `each`, `step`, etc. An endless range doesn't stop iteration (well, it will, eventually, when it overflows):

```crystal
(3..).each do |x|
  puts x
  break if some_condition
end
```

You can't invoke `each` on a begin-less range (it errors at compile-time, unless the begin is a nilable int, and in that case it raises at runtime). But you can invoke `reverse_each`:

```crystal
(..3).reverse_each do |x|
  # ... eventually yields 0, -1, -2
end
```

Note that a begin-less range isn't equivalent to a range that starts with `0` in these case.

A range also has a `===` method which can be use in `case` ([and soon in some `Enumerable` methods ](#7174)). In this case an endless range means "greater than the beginning of the range":

```crystal
number = rand(1..10)
case number
when ..3 then "foo"
when (3..5) then "bar"
when (6..) then "bar" # or just else
end
```

(well, maybe the above isn't a good example because there's still an else `nil` branch there)

Another example:

```crystal
numbers = [1, 10, 3, 4, 5, 8]
numbers.select(6..) # => [10, 8]
numbers.select(..6) # => [1, 3, 4, 5]
```

The above is the same as:

```crystal
numbers.select(&.>=(6))
numbers.select(&.<=(6))
```

I think the former is a bit more readable.

There's also `Number#clamp`, so now there's another way of doing this:

```crystal
num = some_number
num = 10 if num <= 10
```

We can do:

```crystal
num = num.clamp(10..)
```

Same goes with capping to a max.

In Ruby 2.6 there was also an example like this:

```ruby
[1, 2, 3].zip(6..) # => [[1, 6], [2, 7], [3, 8]]
```

which is nice (no need to explicitly specify an upper bound) but currently doesn't work in Crystal (but I might make it work soon).

## As an indexer

As an indexer, an endless range means "until the end of the collection". It's equivalent to passing `-1`, except that you don't have to write that and you don't have to care about `..` vs. `...`:

```crystal
ary = [1, 2, 3, 4, 5]
ary[2..] # => [3, 4, 5]
ary[2...] # => [3, 4, 5]
```

This works for any indexable (`Array`, `Deque`) and `String`.

For a begin-less range it's equivalent to passing `0`, but maybe it's a bit more concise:

```crystal
ary = [1, 2, 3, 4, 5]
ary[..2] # => [1, 2, 3]
```

## The catch-all range

Because the begin and end are now optional, we can also write `(..)`. It doesn't have much use, really. We could forbid it. But right now you can use it:

```crystal
# replace all elements of an array with a single element
ary = [1, 2, 3, 4, 5]
ary[..] = 3
ary # => [3]

# replace all elements of an array
ary = [1, 2, 3, 4, 5]
ary[..] = [6, 7, 8]
ary # => [6, 7, 8]

# fill an array with a value
ary = [1, 2, 3]
ary.fill(8, ..)
ary # => [8, 8, 8]

# another way to dup an array
ary = [1, 2, 3]
b = ary[..]
```

And also:

```crystal
case number
when .. then "duh"
end
```

Yeah... not a lot of uses, but the semantics are consistent.

## Implementation details

Ideally I would change the `Range(Int, Int)` restrictions to `Range(Int?, Int?)`. However, that gives a compile-error in some cases saying that you can't have `Int` inside a union type. Instead of trying to fix that I decided to just drop those inner restrictions for two reasons:
- I don't have much time to think about fixing that error
- It doesn't matter much because you will still get an (probably uglier) error if you pass a non-number range, but:
- I don't think there will be many there passing ranges of non-integers to these methods
- It's still not clear what we want to do with these restrictions: is a `Range(Int, Int)` actually a thing, or should it be more like `Range(< Int, < Int)` saying "anything that inherits `Int`"?

Changing the parser was relatively simple.
Making the changes to `Range` methods was also simple, but it took a bit of time.
Changing the logic of indexers (`ary[3..]`) was very easy because all of that logic is in `Indexable.range_to_index_and_count` (we might want to move that method to `Range` because it's also used in `String` which is not an `Indexable`)

Thoughts?

### DIFF
--- a/spec/std/indexable_spec.cr
+++ b/spec/std/indexable_spec.cr
@@ -161,6 +161,18 @@ describe Indexable do
     last_element.should eq(3)
   end
 
+  it "iterates within a range of indices, no end" do
+    indexable = SafeIndexable.new(5)
+    last_element = nil
+
+    return_value = indexable.each(within: 2..nil) do |elem|
+      last_element = elem
+    end
+
+    return_value.should eq(indexable)
+    last_element.should eq(4)
+  end
+
   it "joins strings (empty case)" do
     indexable = SafeStringIndexable.new(0)
     indexable.join.should eq("")
--- a/spec/std/indexable_spec.cr
+++ b/spec/std/indexable_spec.cr
@@ -173,6 +173,17 @@ describe Indexable do
     last_element.should eq(4)
   end
 
+  it "iterates within a range of indices, no beginning" do
+    indexable = SafeIndexable.new(5)
+
+    elements = [] of Int32
+    return_value = indexable.each(within: nil..2) do |elem|
+      elements << elem
+    end
+
+    elements.should eq([0, 1, 2])
+  end
+
   it "joins strings (empty case)" do
     indexable = SafeStringIndexable.new(0)
     indexable.join.should eq("")
--- a/spec/std/indexable_spec.cr
+++ b/spec/std/indexable_spec.cr
@@ -101,87 +101,88 @@ describe Indexable do
 
   it "iterates through a subset of its elements (#3386)" do
     indexable = SafeIndexable.new(5)
-    last_element = nil
+    elems = [] of Int32
 
     return_value = indexable.each(start: 2, count: 3) do |elem|
-      last_element = elem
+      elems << elem
     end
 
+    elems.should eq([2, 3, 4])
     return_value.should eq(indexable)
-    last_element.should eq(4)
   end
 
   it "iterates until its size (#3386)" do
     indexable = SafeIndexable.new(5)
-    last_element = nil
+    elems = [] of Int32
 
     indexable.each(start: 3, count: 999) do |elem|
-      last_element = elem
+      elems << elem
     end
 
-    last_element.should eq(4)
+    elems.should eq([3, 4])
   end
 
   it "iterates until its size, having mutated (#3386)" do
     indexable = SafeIndexable.new(10)
-    last_element = nil
+    elems = [] of Int32
 
     indexable.each(start: 3, count: 999) do |elem|
-      indexable.size += 1 if elem <= 5
       # size is incremented 3 times
-      last_element = elem
+      indexable.size += 1 if elem <= 5
+      elems << elem
     end
 
     # last was 9, but now is 12.
-    last_element.should eq(12)
+    elems.should eq([3, 4, 5, 6, 7, 8, 9, 10, 11, 12])
   end
 
   it "iterates until its size, having mutated (#3386)" do
     indexable = SafeIndexable.new(10)
-    last_element = nil
+    elems = [] of Int32
 
     indexable.each(start: 3, count: 5) do |elem|
       indexable.size += 1
-      last_element = elem
+      elems << elem
     end
 
     # last element iterated is still 7.
-    last_element.should eq(7)
+    elems.should eq([3, 4, 5, 6, 7])
   end
 
   it "iterates within a range of indices (#3386)" do
     indexable = SafeIndexable.new(5)
-    last_element = nil
+    elems = [] of Int32
 
     return_value = indexable.each(within: 2..3) do |elem|
-      last_element = elem
+      elems << elem
     end
 
+    elems.should eq([2, 3])
     return_value.should eq(indexable)
-    last_element.should eq(3)
   end
 
   it "iterates within a range of indices, no end" do
     indexable = SafeIndexable.new(5)
-    last_element = nil
+    elems = [] of Int32
 
     return_value = indexable.each(within: 2..nil) do |elem|
-      last_element = elem
+      elems << elem
     end
 
+    elems.should eq([2, 3, 4])
     return_value.should eq(indexable)
-    last_element.should eq(4)
   end
 
   it "iterates within a range of indices, no beginning" do
     indexable = SafeIndexable.new(5)
 
-    elements = [] of Int32
+    elems = [] of Int32
     return_value = indexable.each(within: nil..2) do |elem|
-      elements << elem
+      elems << elem
     end
 
-    elements.should eq([0, 1, 2])
+    elems.should eq([0, 1, 2])
+    return_value.should eq(indexable)
   end
 
   it "joins strings (empty case)" do
--- a/spec/std/range_spec.cr
+++ b/spec/std/range_spec.cr
@@ -56,6 +56,7 @@ describe "Range" do
   it "does to_s" do
     (1...5).to_s.should eq("1...5")
     (1..5).to_s.should eq("1..5")
+    (1..nil).to_s.should eq("1..")
   end
 
   it "does inspect" do
@@ -188,6 +189,16 @@ describe "Range" do
       range.each { any = true }
       any.should eq(false)
     end
+
+    it "endless" do
+      range = (3..nil)
+      ary = [] of Int32
+      range.each do |x|
+        ary << x
+        break if ary.size == 5
+      end
+      ary.should eq([3, 4, 5, 6, 7])
+    end
   end
 
   describe "reverse_each" do
@@ -210,6 +221,13 @@ describe "Range" do
       any = false
       range.reverse_each { any = true }
       any.should eq(false)
+    end
+
+    it "raises on endless range" do
+      range = (3..(true ? nil : 1))
+      expect_raises(ArgumentError, "Can't reverse_each endless range") do
+        range.reverse_each { }
+      end
     end
   end
 
@@ -235,6 +253,17 @@ describe "Range" do
 
       iter.rewind
       iter.next.should eq(1)
+    end
+
+    it "does with endless range" do
+      r = (3..nil)
+      iter = r.each
+      iter.next.should eq(3)
+      iter.next.should eq(4)
+
+      iter.rewind
+      iter.next.should eq(3)
+      iter.next.should eq(4)
     end
 
     it "cycles" do
@@ -301,6 +330,42 @@ describe "Range" do
     it "is not empty with ... and begin.succ == end" do
       (1...2).reverse_each.to_a.should eq([1])
     end
+
+    it "raises on endless range" do
+      expect_raises(ArgumentError, "Can't reverse_each endless range") do
+        (1..(true ? nil : 1)).reverse_each
+      end
+    end
+  end
+
+  describe "step" do
+    it "does with inclusive range" do
+      a = 1..5
+      elems = [] of Int32
+      iter = a.step(2) do |x|
+        elems << x
+      end
+      elems.should eq([1, 3, 5])
+    end
+
+    it "does with exclusive range" do
+      a = 1...5
+      elems = [] of Int32
+      iter = a.step(2) do |x|
+        elems << x
+      end
+      elems.should eq([1, 3])
+    end
+
+    it "does with endless range" do
+      a = (1...nil)
+      elems = [] of Int32
+      iter = a.step(2) do |x|
+        elems << x
+        break if elems.size == 5
+      end
+      elems.should eq([1, 3, 5, 7, 9])
+    end
   end
 
   describe "step iterator" do
@@ -354,6 +419,17 @@ describe "Range" do
     it "is not empty with ... and begin.succ == end" do
       (1...2).step(1).to_a.should eq([1])
     end
+
+    it "does with endless range" do
+      a = (1...nil)
+      iter = a.step(2)
+      iter.next.should eq(1)
+      iter.next.should eq(3)
+
+      iter.rewind
+      iter.next.should eq(1)
+      iter.next.should eq(3)
+    end
   end
 
   describe "map" do
@@ -386,5 +462,27 @@ describe "Range" do
     clone.should eq(range)
     clone.begin.should_not be(range.begin)
     clone.end.should_not be(range.end)
+  end
+
+  describe "===" do
+    it "inclusive" do
+      ((1..2) === 0).should be_false
+      ((1..2) === 1).should be_true
+      ((1..2) === 2).should be_true
+      ((1..2) === 3).should be_false
+    end
+
+    it "exclusive" do
+      ((1...2) === 0).should be_false
+      ((1...2) === 1).should be_true
+      ((1...2) === 2).should be_false
+    end
+
+    it "endless" do
+      ((1...nil) === 0).should be_false
+      ((1...nil) === 1).should be_true
+      ((1...nil) === 2).should be_true
+      ((1..nil) === 2).should be_true
+    end
   end
 end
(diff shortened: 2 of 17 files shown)


straight-shoota:
`Range` doesn't have to be restricted to `Int` anyway. Theoretically it can be used with any `Comparable` that implements `#succ`.

j8r:
I mean `col[..]` is supposed to return the whole collection, `col[...]` all but the last one? I've seen a spec for `nil..nil` but not for `nil...nil`

asterite:
`..` and `...` are the same. Inclusive and exclusive against infinite is infinite.

j8r:
@asterite We have to do `..-2` instead of `...` to have all but the last element?
Because we have:
```crystal
a = [0, 1, 2]
puts a[0...-1] #=> [0, 1] 
puts a[0..-2]  #=> [0, 1]
```
`...` won't be equal to `[0...-1]`, but equal to `[0..-1]` and `..` then?

asterite:
Your reasoning is correct. However an endless range means "up to infinite" or without an end. In the case of indexing a collection, because the size is finite, it just means to take everything available. There the exclusive/inclusive distinction doesn't make sense.

Maybe to prevent this small confusion we could make an endless exclusive range a syntax error. Thoughts?

j8r:
Yes the nillable end of exclusive ranges could be forbidden (like now).

sudo-nice:
@asterite There is an example above you wrote:

```crystal
(3..).reverse_each do |x|
  # ... eventually yields 0, -1, -2
end
```

Did you really mean it should yield `3, 2, 1, 0, -1, -2`?

sudo-nice:
So that `.reverse_each` on end-less range would mean *start when the range starts, but go the opposite direction*?

asterite:
Oops, I had a typo in the example. Now it's fixed.

yxhuvud:
Regarding the catch-all range, while it isn't amazing for one dimensional arrays, it will probably be a godsend to people implementing matrices with multiple dimensions.

spec/std/range_spec.cr
```diff
@@ -301,6 +356,49 @@ describe "Range" do
     it "is not empty with ... and begin.succ == end" do
       (1...2).reverse_each.to_a.should eq([1])
     end
+
+    it "raises on endless range" do
```
RX14: Can we have a spec for the reverse iterator working too?

spec/std/range_spec.cr
```diff
@@ -301,6 +356,49 @@ describe "Range" do
...
+      expect_raises(ArgumentError, "Can't reverse_each endless range") do
+        (1..(true ? nil : 1)).reverse_each
+      end
+    end
+  end
+
+  describe "step" do
```
RX14: Can we test negative steps?

spec/std/range_spec.cr
```diff
@@ -301,6 +356,49 @@ describe "Range" do
...
+
+    it "raises on beginless range" do
+      a = nil..3
+      expect_raises(ArgumentError, "Can't step beginless range") do
+        a.step(2) { }
+      end
+    end
   end
 
   describe "step iterator" do
```
RX14: ditto

src/range.cr
```diff
@@ -101,11 +104,16 @@ struct Range(B, E)
...
+      raise ArgumentError.new("Can't each beginless range")
+    end
+
+    end_value = @end
+    while end_value.nil? || current < end_value
       yield current
       current = current.succ
     end
-    yield current if !@exclusive && current == @end
+    yield current if !@exclusive && !end_value.nil? && current == end_value
```
RX14: `current == nil` will never be true, so i don't think this extra check is required

src/range.cr
```diff
@@ -125,9 +137,21 @@ struct Range(B, E)
   # # prints: 14 13 12 11 10
   # ```
   def reverse_each : Nil
-    yield @end if !@exclusive && !(@end < @begin)
-    current = @end
-    while @begin < current
+    {% if E == Nil %}
+      {% raise "Can't reverse_each endless range" %}
+    {% end %}
```
RX14: Why not the same macro check in `#each`?

spec/std/indexable_spec.cr
```diff
@@ -161,6 +161,29 @@ describe Indexable do
...
+  it "iterates within a range of indices, no end" do
+    indexable = SafeIndexable.new(5)
+    last_element = nil
+
+    return_value = indexable.each(within: 2..nil) do |elem|
+      last_element = elem
+    end
+
+    return_value.should eq(indexable)
+    last_element.should eq(4)
```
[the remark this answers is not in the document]
asterite: All existing tests were like this, so I just followed the style. I'll add an extra commit that improves all of this.

spec/std/range_spec.cr
asterite: Unless I'm misunderstanding something, this is the spec for the reverse iterator. Or did you mean something else?

spec/std/range_spec.cr
asterite: Good idea, I totally forgot about this. I'll add it as an extra commit.

src/range.cr
asterite: Good thinking :-)

src/range.cr
asterite: Strange, I though I wrote it... I'll add it

asterite:
I originally had two commits, one for end-less range and another for begin-less range. I'll add more commits that fix the existing issues and then we can squash it into a single commit, after all the changes aren't very big.

spec/std/range_spec.cr
asterite: Actually, at least in Ruby `step` can't be negative because a Range is always increasing, if it's decreasing (for example `5..1`) then it's "empty".

Let's discuss this in a separate issue, the current behavior is not yielding at all, but maybe giving an error would be better, or even making it work (using `pred`).

asterite: Small correction: a negative step makes it hang. Definitely not good but I still think we should discuss it in a separate issue (I'll open one soon).

spec/std/indexable_spec.cr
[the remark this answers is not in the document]
asterite: Done!

spec/std/indexable_spec.cr
```diff
@@ -161,6 +161,29 @@ describe Indexable do
...
+  end
+
+  it "iterates within a range of indices, no beginning" do
+    indexable = SafeIndexable.new(5)
+
+    elements = [] of Int32
+    return_value = indexable.each(within: nil..2) do |elem|
```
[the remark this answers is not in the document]
asterite: Done!

src/range.cr
asterite: Done!

src/range.cr
asterite: Done!

asterite:
@RX14 All comments addressed (except one I didn't understand).

Given that this PR also adds specs for `Range#step`, I'd like to merge this first and then I can send a PR fixing it for negative steps (it's very similar to with positive steps except we have to use `pred`).

spec/std/range_spec.cr
RX14: This is the spec for the `reverse_each` iterator raising, but not actually working (calling `.next`)

spec/std/range_spec.cr
RX14: :+1:

spec/std/range_spec.cr
asterite: Ooooh... you are right! In fact this wasn't working. I added a spec and also fixed the code.

bew:
Is `foo ..2` the same as `foo(..2)` or `foo..2` ?

asterite:
@bew Excellent question!

I decided it's parsed as `foo..2`, so as a range, not as `foo(..2)`. I think parsing it the other way would mean a breaking change, plus you can always disambiguate the two cases with parentheses.

But we can discuss it, of course.

bew:
Yes, `foo..2` is the most logical and easiest to parse from the compiler point of view, but for us humans it's not that easy and logical I think..

The following thought is for ranges in general:
I would prefer to make `1 .. 3` an error, and force the `1..3` or `1..(2 + 1)` (no spaces around `..`) for ranges.
_btw, iirc I never saw ranges with spaces_

This way you would easily disambiguate the 2: when there is no space it would _ALWAYS_ be a range, when there is spaces it can only be a partial range (begin and/or end -less range)

j8r:
Agree with @bew, removing spaces disambiguate with a good side effect to ensure a consistent style for ranges.

bew:
Also, can you add a TODO somewhere to change `x..nil` & `nil..x` to `x..` & `..x` after next release (in specs)?

asterite:
@bew But you said:

> (this is important for method calls when generated from macros for example)

For this reason we shouldn't change semantics based on a single space.

Otherwise, let's remove begin-less ranges, but I think the current behavior (`foo ..2` is `foo..2`) is fine.

bew:
Hmm that's true too.. No idea how useful begin-less range can be, and how much it would impact people to not have it..

RX14:
I'm fine with the parsing as-is

RX14:
@bcardiff any reason why?

asterite:
Same question. This is backwards compatible. The only reason to wait for 0.28.0 is that this is a new language feature (and I wouldn't mind that).

src/range.cr
```diff
@@ -207,11 +256,16 @@ struct Range(B, E)
...
   def includes?(value)
-    if @exclusive
-      @begin <= value < @end
-    else
-      @begin <= value <= @end
-    end
+    begin_value = @begin
+    end_value = @end
+
+    # TOOD: change to `nil?` after removing the `nil?` error related to `Pointer`
```
bcardiff:
```suggestion
    # TODO: change to `nil?` after removing the `nil?` error related to `Pointer`
```

bcardiff: Also, which error?

Sija: @bcardif #7180

asterite: Thanks, fixed.

This is the error: https://play.crystal-lang.org/#/r/5xke

It was an old error because at some pointer you would use `nil?` to check for a null pointer, but this was later changed to `null?` and the error was introduced to avoid silent breaking changes.

I already removed it in master (#7180) but of course the current compiler would still complain.

asterite:
> It's a bit weird that beginless range are Iterable(Nil) Enumerable(Nil), yet they can be reversed iterable and serve fine as range of Ints for example.

Maybe it's not that bad, if it's `Iterable(Nil)` it looks like there's nothing useful to iterate, and in fact before reaching that point you would get a compile-time or runtime error.

> I would add some specs to clarify foo + 1 .. bar + 1

How does that change with this PR? I can't see endless or beginless ranges there.

> slightly unrelated but since we are making decisions regaring foo .. bar being foo..bar vs foo(..bar) (which make this PR backward compatible 👍) we can also settle the precedence in the specs.

There's already a parser spec for `foo..2` that shows that it's parsed as a range and not as a call, and there's also one for `foo(..2)` so I think we are good there.

The thing we must remember to do later is to add this to the crystal-book docs.

bcardiff:
> > I would add some specs to clarify `foo + 1 .. bar + 1`

> How does that change with this PR? I can't see endless or beginless ranges there.

because there is a decision regarding `foo .. bar` to be `foo..bar` and not `foo(..bar)`. The same could apply `foo + 1 .. bar + 1` wrt operators precedence `+` `..` and call.

it's not mandatory but if would be good to have some spec to check the precedence will hold in the future.

bcardiff:
Please rebase on master so we can merge commit the PR. :shipit:

asterite:
@bcardiff Done! :-)

bcardiff:
There is definitely something odd in this PR, not only travis, but circle also fails sometimes also. Travis is likely to fail more often.

I've spent time in my fork checking/tweaking travis because I thought it was scoped to it, but CircleCI sometimes also fails. Build of master does not exhibit this error.

I tried changing ulimit, update docker, update travis dist, same behavior.

```
...
./bin/crystal build  -o .build/std_spec spec/std_spec.cr
cc: error: _main.o: No such file or directory
cc: error: P-ointer40V-oid41.o: No such file or directory
cc: error: S-lice40U-I-nt841.o: No such file or directory
cc: error: P-ointer40U-I-nt841.o: No such file or directory
```

* https://circleci.com/gh/bcardiff/crystal/8
* https://travis-ci.org/bcardiff/crystal/jobs/483559179

@asterite any clue? Maybe some interaction between path/ranges?

asterite:
I will check it later when I have time. We can probably delay this until 0.28.0 or even later, it's not super important.

j8r:
No problem on my side with the `feature/xless-ranges` branch. I've ran `make` before doing this.
```
/app # ./bin/crystal build  -o .build/std_spec spec/std_spec.cr?
/app # uname -a
Linux f5b436f201b2 4.18.0-11-generic #12-Ubuntu SMP Tue Oct 23 19:22:37 UTC 2018 x86_64 Linux
/app # cat /etc/alpine-release 
3.9.0
/app # git describe --tags --long --always
0.25.0-579-gcbf479d31
```

asterite:
@j8r Thank you! But that's not exactly the command that's run. The full command is:

```
./bin/crystal build -D preview_overflow -D compiler_rt  -o .build/std_spec spec/std_spec.cr
```

Check the `-D` flags there.

j8r:
No problem too @asterite  :
```
./bin/crystal build -D preview_overflow -D compiler_rt  -o .build/std_spec spec/
std_spec.cr
```

asterite:
(@j8r thank you!)

asterite:
CI passed this time after merging #7397 🎉 😁

asterite:
Merge? 😺

straight-shoota:
Thanks @asterite!

PR #7182, opened 2018-12-13
Compiler: allow trailing commas inside tuple types

asterite:
This is a small enhancement. Trailing commas are super useful when generating code with macros, to not care about putting that one extra comma at the end.


bcardiff:
I would say it's a language change rather and a compiler bug :-)

asterite:
@bcardiff Good call, I changed the label to `kind:feature`.

pynixwang:
join

PR #7184, opened 2018-12-13
Fix formatting for heredoc having interpolation of string literal

### LINKED ISSUE #7183: Formatter cannot format heredoc having interpolation of string literal
Extracted from https://github.com/crystal-lang/crystal/pull/7099#issuecomment-446491007

This code causes an error.

```crystal
<<-FOO
foo#{"foo"}
FOO
```

```console
$ crystal tool format -< foo.cr
Error: couldn't format STDIN, please report a bug including the contents of it: https://github.com/crystal-lang/crystal/issues

expecting DELIMITER_END, not `INTERPOLATION_START, foo`, at :2:1 (Exception)
  from Crystal::Formatter#check<Symbol>:Nil
  from Crystal::Formatter#visit_string_interpolation<Crystal::StringInterpolation, Crystal::Token, Int32, Int32, Int32, Int32, Bool>:Bool
  from Crystal::Formatter#next_token:Crystal::Token
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::Formatter>:(Bool | Nil)
  from Crystal::format<String>:String
  from Crystal::Command#tool:(Bool | Nil)
  from __crystal_main
  from main
```

MakeNowJust:
Fixed #7183

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -1944,7 +1944,9 @@ module Crystal
           @inside_interpolation = true
           exp = preserve_stop_on_do { parse_expression }
 
-          if exp.is_a?(StringLiteral)
+          # We cannot reduce `StringLiteral` of interpolation inside heredoc into `String`
+          # because heredoc try to remove its indentation.
+          if exp.is_a?(StringLiteral) && delimiter_state.kind != :heredoc
             pieces << Piece.new(exp.value, line_number)
           else
             pieces << Piece.new(exp, line_number)
@@ -2025,7 +2027,7 @@ module Crystal
     def remove_heredoc_indent(pieces : Array, indent)
       current_line = IO::Memory.new
       remove_indent = true
-      new_pieces = [] of ASTNode
+      new_pieces = [] of ASTNode | String
       previous_line_number = 0
       pieces.each_with_index do |piece, i|
         value = piece.value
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/syntax/parser.cr
```diff
@@ -2083,15 +2085,22 @@ module Crystal
         line = remove_heredoc_from_line(line, indent, pieces.last.line_number) if remove_indent
         add_heredoc_piece new_pieces, line
       end
-      new_pieces
+      new_pieces.map do |piece|
```
asterite: Can this `map` be avoided and instead this logic put in `add_heredoc_piece` somehow?

asterite:
Feel free to merge, the `map` probably doesn't affect overall performance and can always be improved later.

src/compiler/crystal/syntax/parser.cr
MakeNowJust: To avoid this `map`, `add_heredoc_piece` must take extra argument `last_is_string` and it must return new `last_is_string` state. Like:

```crystal
    def remove_heredoc_indent(pieces : Array, indent)
      ...
      last_is_string = false
      new_pieces = [] of ASTNode
      ...

      pieces.each_with_index do |piece, index|
        ...
        last_is_string = add_heredoc_piece(pieces, piece, last_is_string)
        ...
      end

      ...
    end

    private def add_heredoc_piece(pieces, piece : String, last_is_string : Bool)
      if last_is_string
        pieces.last!.as(StringLiteral).value += piece
      else
        pieces << StringLiteral.new(piece)
      end
      true
    end

    private def add_heredoc_piece(pieces, piece : ASTNode, last_is_string : Bool)
      pieces << piece
      false
    end
``

Issue #5110, opened 2017-10-12
[rfc] Import Object#yield_self from Ruby 2.5


MakeNowJust:
Ruby 2.5 decide to introduce `Object#yield_self` (see [here](https://bugs.ruby-lang.org/issues/6721)). This method yields `self` to the block and then returns its result. The implementation is very simple:

```crystal
class Object
  def yield_self
    yield self
  end
end
```

Should we import this method into Crystal? What do you think?

straight-shoota:
It's basically `Object#try` which also works for `Nil` with a descriptive yet still uninviting long name.

RX14:
I don't really see the usecase..?

MakeNowJust:
@RX14 Yes, I'd like to ask Ruby committer about usecase.

lbguilherme:
Not that I defend it, but could be used to create a variable, without creating a variable. To help keep calculations inline:

    result = long.chained.computation(12).foo(1, 2, 3).yield_self {|list| list.sum * list.product }.to_s.size

RethinkDB (a database whose query language is mostly functional) has a similar operation, called `do`: https://rethinkdb.com/api/ruby/do/.

It is like a map, but for the entire thing.

ysbaddaden:
It's similar to `.tap` in behavior and intent, expect it returns the block's value (not the tapped value).

I don't see much usage. Naming is verbose, and assigning to a value feels better:
```crystal
list = long.chained.computation(12).foo(1, 2, 3)
result = (list.sum * list.product).to_s.size
```

MakeNowJust:
@RX14 Why do you close this issue? (I can guess but...) If you could explain this, you should do it, or if you couldn't explain this, you shouldn't close this.

RX14:
The consensus was that it wasn't really very useful. If anyone has a counterexample to discuss we can always reopen.

asterite:
`yield_self` is for those that want a "pipe operator" in Ruby: http://mlomnicki.com/yield-self-in-ruby-25/

Like others here, I don't think it looks nice, nor I think it's readable. And since we have `try` (and you will probably never want to have a `nil` in a chain), it can work as good as `yield_self`.

MakeNowJust:
I found very useful usecase of `Object#yield_self`! See this simple fizz buzz example:

```crystal
def fizz?(n)
  n % 3 == 0
end

def buzz?(n)
  n % 5 == 0
end

(1..100).each do |i|
[...]

PR #7188, opened 2018-12-14
Add Object#then

Sija:
- Brought from Ruby 2.6, which aliased `Object#yield_self` to `Object#then`
- Followup to #5110

### DIFF
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -133,6 +133,17 @@ private class HashedTestObject
 end
 
 describe Object do
+  describe "then" do
+    it "returns output of given block" do
+      "Hello".then { "World" }.should eq("World")
+      "Hello".then(&.upcase).should eq("HELLO")
+    end
+
+    it "works on nils" do
+      nil.then { |x| x.should be_nil; true }.should be_true
+    end
+  end
+
   describe "delegate" do
     it "delegates" do
       wrapper = StringWrapper.new("HellO")
--- a/src/object.cr
+++ b/src/object.cr
@@ -143,6 +143,29 @@ class Object
     end
   end
 
+  # Yields `self` to the block, and returns its output.
+  #
+  # Overall, `then` improves readability of the code by promoting chaining
+  # over nested function calls.
+  #
+  # ```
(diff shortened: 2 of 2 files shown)


asterite:
What are good use cases of this in Crystal? All the given examples don't look very useful.

Sija:
@asterite In short, method chaining.

```cr
"https://api.github.com/repos/crystal-lang/crystal"
  .then(&->URI.parse(String))
  .then(&->HTTP::Client.get(URI))
  .then(&.body)
  .then(&->JSON.parse(String))
  .then(&.["stargazers_count"])
  .then { |x| "Crystal has #{x} stargazers" }
  .then(&->pp(String))
```

straight-shoota:
This has already been discussed in #5110 and was essentially rejected.
I'm not strictly against such a method, but continuing from the previous RFC, I would expect a compelling argument demonstrating the benefits and usefulness of this feature.

Besides, the name is not very smart. `#yield_self` is to bulky and `#then` looks more like a promise pattern.

The example for method chaining is really ugly because it requires to explicitly specify types for turning methods into process. That's hardly any improvement!

Sija:
> I'm not strictly against such a method, but continuing from the previous RFC, I would expect a compelling argument demonstrating the benefits and usefulness of this feature.

My arguments for its usefulness would be:

- ability to transform the value within the method chain
- ability to use it in `when` branches in cases where you'd need to use compound statement (`... && ...`)

> Besides, the name is not very smart. `#yield_self` is to bulky and `#then` looks more like a promise pattern.

True that, but ya know, _naming is hard_, just next to cache invalidation ;)

> The example for method chaining is really ugly because it requires to explicitly specify types for turning methods into process. That's hardly any improvement!

Would look better without a need to pass 'em types, but then OTOH it makes it pretty clear how the transformation progresses (type-wise). It always could be written in long form (`{ |...| ... }`).

straight-shoota:
> It always could be written in long form

It could. But then it would loose it's conciseness, reducing the benefit over using simple variables in between steps.

RX14:
We already rejected this, I doubt there will be a change in consensus.

PR #7202, opened 2018-12-18
OpenSSL: Refactor const specs to reduce changes of failing

bcardiff:
Some of us use to have a `foo.cr` in the project root 🙈 and building it would make this spec fail.

ref: https://twitter.com/bcardiff/status/1075123585210085377


bcardiff:
How did `const` got typed instead of `context` !? 🤷‍♂️

PR #7206, opened 2018-12-20
Add overflow detection with preview opt-in

### LINKED ISSUE #3103: Add suport for Int overflow checking
LLVM provides some builtin functions for checking overflow in arithmetic operations (http://llvm.org/docs/LangRef.html#arithmetic-with-overflow-intrinsics) with a miniumum overhead.
A PR would be accepted with support for overflow checking?

bcardiff:
This PR adds overflow checks for `+` `-` `*`, to number conversions `to_iX`, `to_uX`, `to_fX` and so to constructs like `Int32.new(value)`.

The raised exception will contain, when possible, the location of the method/operator that originate it.

~~It adds some new methods to be able to keep previous wrapping behaviour. `to_iX!`/ `Int32.new!(value)`. This methods should be used to be ready for upcoming breaking changes.~~ (Split at: #7226)

~~It prepares some new primitives to be able to have cleaner compiler code in the future. That is drop the `:cast` internal in favor of `:convert` and `:unchecked_convert` to reflect better the nature of `to_X`/`to_X!` methods.~~ (Split at: #7226)

~~There are many refactors to the stdlib~~ (Split at #7256) and ~~compiler code~~ (Split at #7262) ~~where the wrapping/unsafe behaviour was needed. Although some changes might look odd they should be reviewed as keeping the current code since operators and methods will be changing in a future version. Some changes show a need for _future_ refactor (JSON/YAML parsing for dealing with UInt64, Enum flags with UInt64 base types, conversions for big types). In future version the overflow check on Time and other can be simplified.~~

In order to avoid breaking changes and to allow a preview of this feature, the overflow behavior is only enabled if compiled with `-D preview_overflow`.

I checked locally that even a 2nd compiler generation pass the specs. I did catch some left overs with that.

```
$ make clean_crystal crystal && sleep 1 && touch src/compiler/crystal.cr && make crystal
$ make std_spec compiler_spec
```

Closes #6223
Closes #3103

#### Sample

```
# file: foo.cr
Int32::MAX + 1
puts "ok"
```

```
$ ./bin/crystal foo.cr 
ok
```

```
$ ./bin/crystal foo.cr -D preview_overflow
Unhandled exception: Overflow (OverflowError)
...
```

### DIFF
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -231,6 +231,42 @@ describe "Code gen: arithmetics primitives" do
           ), flags: PreviewOverflowFlags).to_i.should eq(1)
         end
       {% end %}
+
+      {% for float_type in [Float32, Float64] %}
+        {% if type != UInt128 || float_type != Float32 %}
+          # skip for type == UInt128 && float_type == Float32
+          # since Float32::MAX < UInt128::MAX
+          it "raises overflow if greater than {{type}}::MAX (from {{float_type}})" do
+            run(%(
+              require "prelude"
+
+              v = {{float_type}}.new({{type}}::MAX) * {{float_type}}.new(1.5)
+
+              begin
+                v.{{method}}
+                0
+              rescue OverflowError
+                1
+              end
+            ), flags: PreviewOverflowFlags).to_i.should eq(1)
+          end
+        {% end %}
+
+        it "raises overflow if lower than {{type}}::MIN (from {{float_type}})" do
+          run(%(
+            require "prelude"
+
+            v = {{float_type}}.new({{type}}::MIN) * {{float_type}}.new(1.5) - {{float_type}}.new(1.0)
+
+            begin
+              v.{{method}}
+              0
+            rescue OverflowError
+              1
+            end
+          ), flags: PreviewOverflowFlags).to_i.should eq(1)
+        end
+      {% end %}
     {% end %}
   end
 end
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -1,11 +1,13 @@
 require "../../spec_helper"
 
-{% if flag?(:bits64) %}
+{% if flag?(:darwin) %}
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, UInt128, Int8, Int16, Int32, Int64, Int128]
 
 PreviewOverflowFlags = ["preview_overflow"]
 {% else %}
-# skip Int128 and UInt128 on 32 bits platforms
+# Skip Int128 and UInt128 on linux platforms due to compiler-rt dependency.
+# PreviewOverflowFlags includes compiler_rt flag to support Int64 overflow
+# detection in 32 bits platforms.
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64]
 
 PreviewOverflowFlags = ["preview_overflow", "compiler_rt"]
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -2,9 +2,13 @@ require "../../spec_helper"
 
 {% if flag?(:bits64) %}
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, UInt128, Int8, Int16, Int32, Int64, Int128]
+
+PreviewOverflowFlags = ["preview_overflow"]
 {% else %}
 # skip Int128 and UInt128 on 32 bits platforms
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64]
+
+PreviewOverflowFlags = ["preview_overflow", "compiler_rt"]
 {% end %}
 
 describe "Code gen: arithmetics primitives" do
@@ -73,7 +77,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
 
       it "raises overflow for {{type}} + Int64" do
@@ -85,7 +89,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
     {% end %}
   end
@@ -101,7 +105,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
 
       it "raises overflow for {{type}} - Int64" do
@@ -113,7 +117,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
     {% end %}
   end
@@ -129,7 +133,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
 
       it "raises overflow for {{type}} * Int64" do
@@ -141,7 +145,7 @@ describe "Code gen: arithmetics primitives" do
           rescue OverflowError
             1
           end
-        ), flags: ["preview_overflow"]).to_i.should eq(1)
+        ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
     {% end %}
   end
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -61,4 +61,32 @@ describe "Code gen: arithmetics primitives" do
       end
     {% end %}
   end
+
+  describe "+ addition" do
+    {% for type in SupportedInts %}
+      it "raises overflow for {{type}}" do
+        run(%(
+          require "prelude"
+          begin
+            {{type}}::MAX + {{type}}.new(1)
+            0
+          rescue OverflowError
+            1
+          end
+        ), flags: ["preview_overflow"]).to_i.should eq(1)
+      end
+
+      it "raises overflow for {{type}} + Int64" do
+        run(%(
+          require "prelude"
+          begin
+            {{type}}::MAX + 1_i64
+            0
+          rescue OverflowError
+            1
+          end
+        ), flags: ["preview_overflow"]).to_i.should eq(1)
+      end
+    {% end %}
+  end
 end
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -269,4 +269,36 @@ describe "Code gen: arithmetics primitives" do
       {% end %}
     {% end %}
   end
+
+  describe ".to_f conversions" do
+    it "raises overflow if greater than Float32::MAX" do
+      run(%(
+        require "prelude"
+
+        v = Float64.new(Float32::MAX) * 1.5_f64
+
+        begin
+          v.to_f32
+          0
+        rescue OverflowError
+          1
+        end
+      ), flags: PreviewOverflowFlags).to_i.should eq(1)
+    end
+
+    it "raises overflow if lower than Float32::MIN" do
+      run(%(
+        require "prelude"
+
+        v = Float64.new(Float32::MIN) * 1.5_f64
+
+        begin
+          v.to_f32
+          0
+        rescue OverflowError
+          1
+        end
+      ), flags: PreviewOverflowFlags).to_i.should eq(1)
+    end
+  end
 end
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -2,6 +2,10 @@ require "../../spec_helper"
 
 {% if flag?(:darwin) %}
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, UInt128, Int8, Int16, Int32, Int64, Int128]
+SupportedIntsConversions = {
+                             to_i8: Int8, to_i16: Int16, to_i32: Int32, to_i64: Int64, to_i128: Int128,
+                             to_u8: UInt8, to_u16: UInt16, to_u32: UInt32, to_u64: UInt64, to_u128: UInt128,
+                           }
 
 PreviewOverflowFlags = ["preview_overflow"]
 {% else %}
@@ -9,6 +13,10 @@ PreviewOverflowFlags = ["preview_overflow"]
 # PreviewOverflowFlags includes compiler_rt flag to support Int64 overflow
 # detection in 32 bits platforms.
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, Int8, Int16, Int32, Int64]
+SupportedIntsConversions = {
+                             to_i8: Int8, to_i16: Int16, to_i32: Int32, to_i64: Int64,
+                             to_u8: UInt8, to_u16: UInt16, to_u32: UInt32, to_u64: UInt64,
+                           }
 
 PreviewOverflowFlags = ["preview_overflow", "compiler_rt"]
 {% end %}
@@ -149,6 +157,80 @@ describe "Code gen: arithmetics primitives" do
           end
         ), flags: PreviewOverflowFlags).to_i.should eq(1)
       end
+    {% end %}
+  end
+
+  describe ".to_i conversions" do
+    {% for method, path_type in SupportedIntsConversions %}
+      {% type = path_type.resolve %}
+
+      {% if ![UInt64, Int128, UInt128].includes?(type) %}
+        it "raises overflow if greater than {{type}}::MAX" do
+          run(%(
+            require "prelude"
+
+            v = UInt64.new({{type}}::MAX) + 1_u64
+
+            begin
+              v.{{method}}
+              0
+            rescue OverflowError
+              1
+            end
+          ), flags: PreviewOverflowFlags).to_i.should eq(1)
+        end
+      {% end %}
+
+      {% if ![UInt128].includes?(type) && SupportedInts.includes?(UInt128) %}
+        it "raises overflow if greater than {{type}}::MAX (using UInt128)" do
+          run(%(
+            require "prelude"
+
+            v = UInt128.new({{type}}::MAX) + 1_u128
+
+            begin
+              v.{{method}}
+              0
+            rescue OverflowError
+              1
+            end
+          ), flags: PreviewOverflowFlags).to_i.should eq(1)
+        end
+      {% end %}
+
+      {% if ![Int64, Int128, UInt128].includes?(type) %}
+        it "raises overflow if lower than {{type}}::MIN" do
+          run(%(
+            require "prelude"
+
+            v = Int64.new({{type}}::MIN) - 1_i64
+
+            begin
+              v.{{method}}
+              0
+            rescue OverflowError
+              1
+            end
+          ), flags: PreviewOverflowFlags).to_i.should eq(1)
+        end
+      {% end %}
+
+      {% if ![Int128].includes?(type) && SupportedInts.includes?(Int128) %}
+        it "raises overflow if lower than {{type}}::MIN (using Int128)" do
+          run(%(
+            require "prelude"
+
+            v = Int128.new({{type}}::MIN) - 1_i128
+
+            begin
+              v.{{method}}
+              0
+            rescue OverflowError
+              1
+            end
+          ), flags: PreviewOverflowFlags).to_i.should eq(1)
+        end
+      {% end %}
     {% end %}
   end
 end
--- a/spec/compiler/codegen/arithmetics_spec.cr
+++ b/spec/compiler/codegen/arithmetics_spec.cr
@@ -117,4 +117,32 @@ describe "Code gen: arithmetics primitives" do
       end
     {% end %}
   end
+
+  describe "* multiplication" do
+    {% for type in SupportedInts %}
+      it "raises overflow for {{type}}" do
+        run(%(
+          require "prelude"
+          begin
+            ({{type}}::MAX / {{type}}.new(2) &+ {{type}}.new(1)) * {{type}}.new(2)
+            0
+          rescue OverflowError
+            1
+          end
+        ), flags: ["preview_overflow"]).to_i.should eq(1)
+      end
+
+      it "raises overflow for {{type}} * Int64" do
+        run(%(
+          require "prelude"
+          begin
+            ({{type}}::MAX / {{type}}.new(2) &+ {{type}}.new(1)) * 2_i64
+            0
+          rescue OverflowError
+            1
+          end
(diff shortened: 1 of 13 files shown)


src/float.cr
```diff
@@ -223,6 +228,11 @@ struct Float64
     value.to_f64
   end
 
+  # Returns an `Float64` by invoking `unsafe_to_f64` on *value*.
+  def Float64.unsafe_new(value)
```
Sija:
```suggestion
  def self.unsafe_new(value)
```

src/big/big_decimal.cr
```diff
@@ -306,57 +306,93 @@ struct BigDecimal < Number
...
+  end
+
+  # Converts to `Int32`. Truncates anything on the right side of the decimal point.
+  def unsafe_to_i
+    unsafe_to_i32
+  end
+
+  # Converts to `BigInt`. Truncates anything on the right side of the decimal point,
+  # converting negative to positive.
+  def to_big_u
```
Sija: Shouldn't this raise for negative values?

src/primitives.cr
```diff
@@ -280,10 +280,22 @@ end
                              to_u8: UInt8, to_u16: UInt16, to_u32: UInt32, to_u64: UInt64, to_u128: UInt128,
                              to_f32: Float32, to_f64: Float64,
                            } %}
+        # TODO 0.28.0 replace with @[Primitive(:convert)]
+
         # Returns `self` converted to `{{type}}`.
+        # Raises`OverflowError`.
```
Sija:
```suggestion
        # In case of overflow raises `OverflowError`.
```

RX14:
Is there any way to make this multiple PRs? For example, a PR to add unsafe casts, a PR to update the stdlib, a PR to enable the option for standard operators to raise on overflow? Nearly 50 commits is a bit too much

src/primitives.cr
r00ster91:
```cr
        # Raises `OverflowError` in case of overflow.
```

src/big/big_decimal.cr
bcardiff: Unsigned BitInt are not really a thing I will make that method private. This is not changing the current behavior which should also be reviewed because:

```crystal
-2.to_big_i.unsafe_to_u8 # => 2_u8
-2.unsafe_to_u8          # => 254_u8
```

src/float.cr
bcardiff: The change was done that way to minimize the diff. In some cases the type is mention explicitly and in others `self` is used.

src/primitives.cr
bcardiff: done

bcardiff:
Once the build is happy I will split in probably 3 PRs.
- prepare unsafe ops
- refactor std libs and specs
- implement overflow semantics

The Int128 support for 32 bits or even in linux is causing some issues.
Up to now Int128 multiplication was never stressed in the CI and now we are getting `undefined reference to __muloti4`.
I though it was due to llvm-4.0, but even building against llvm-6.0 is not changing that issue.
I will try to find a workaround for that mainly.

RX14:
Don't bother with `Int128` specs which fail for now. Solving that will require a lot more work.

asterite:
I didn't have time to review this yet, but I will during this week (but feel free to merge it if you all think it's good)

RX14:
@bcardiff Turns out that `__muloti4` is actually an overflow-checking 64-bit multiply, not related to 128bit specs at all.

RX14:
Actually, it's a 64bit multiply with overflow checking used for the i128 support, weird.

src/exception.cr
```diff
@@ -129,6 +129,18 @@ class DivisionByZeroError < Exception
...
 end
 
+# Raised when the result of an arithmetic operation is outside of the range
+# that can be represented within the given operands types.
+#
+# ```
+# Int32::Max + 1 # raises OverflowError (Overflow)
+# ```
+class OverflowError < Exception
+  def initialize(message = "Overflow")
```
asterite: Maybe this should be "Overflow or underflow"? (unless we have a way to detect whether it was overflow or underflow, but I don't think so)

r00ster91: And maybe also be a bit more specific about what kind of overflow it is?
For example `Integer overflow or underflow`.

spec/compiler/codegen/arithmetics_spec.cr
```diff
@@ -2,9 +2,17 @@ require "../../spec_helper"
 
 {% if flag?(:bits64) %}
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, UInt128, Int8, Int16, Int32, Int64, Int128]
```
r00ster91: Shouldn't these constants be capitalized and in snake_case?

src/exception.cr
```diff
@@ -129,6 +129,18 @@ class DivisionByZeroError < Exception
   end
 end
 
+# Raised when the result of an arithmetic operation is outside of the range
+# that can be represented within the given operands types.
+#
+# ```
+# Int32::Max + 1 # raises OverflowError (Overflow)
```
r00ster91: What about adding an example about underflow?
```suggestion
# Int32::MAX + 1 # raises OverflowError (because of overflow)
# Int32::MIN - 0 # raises OverflowError (because of underflow)
```

src/exception.cr
bcardiff: Underflow occur on floating point numbers. I'm ok to use the same exception. But we are not checking for underflow here.

If we aim to use the same exception, the proposal is to change the message or the name also?

spec/compiler/codegen/arithmetics_spec.cr
bcardiff: no, it's not mandatory. It is valid and used in other places, like `c_struct_spec.cr` to mention one.

src/exception.cr
asterite: I mean, when you subtract one from the most negative integer, is that an underflow or an overflow?

asterite: Nevermind, it seems that's an overflow too. Ignore this comment :-)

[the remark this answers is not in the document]
bcardiff: OverflowError is raised with in `#to_f` methods. It's not just integer overflow. :-)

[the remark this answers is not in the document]
r00ster91: Ah alright then it's probably better to keep it simple.

bcardiff:
First stage of the PR split at #7226

bcardiff:
Second stage of the PR split at #7256.

bcardiff:
Third stage of the PR split at #7262 . The fourth and will be this PR after a merge & rebase.

bcardiff:
Forth stage rebased in this PR. The only thing I imagine is splitting further is the operators changes `+`,`-`,`/` from the conversion changes `to_X`,`T.new`.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -173,6 +174,90 @@ class Crystal::CodeGenVisitor
...
+                 binary_overflow_fun "llvm.uadd.with.overflow.i16", llvm_context.int16
+               when :u32
+                 binary_overflow_fun "llvm.uadd.with.overflow.i32", llvm_context.int32
+               when :u64
+                 binary_overflow_fun "llvm.uadd.with.overflow.i64", llvm_context.int64
+               when :u128
+                 binary_overflow_fun "llvm.uadd.with.overflow.i128", llvm_context.int128
+               else
+                 raise "unreachable"
+               end
```
RX14: I suggest modifying `binary_overflow_fun` to take `"add"` and the `IntegerType` and work out the intrinsic name and type. Then, you can make a `codegen_binary_op_result("add", tmax, t1, t2, p1, p2)`, which additionally uses the `"add"` to work out which `codegen_trunc_binary_op_result` to use when overflow checking is disabled.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -173,6 +174,90 @@ class Crystal::CodeGenVisitor
...
+
+    res = extract_value res_with_overflow, 0
+    o_bit = extract_value res_with_overflow, 1
+
+    op_overflow = new_block "overflow"
+    op_normal = new_block "normal"
+
+    # TODO raise overflow if outside of t1 type range interpreted as a t2
+
+    cond o_bit, op_overflow, op_normal
```
RX14: I think we should use `llvm.expect.i1(o_bit, false)` here.

spec/compiler/codegen/arithmetics_spec.cr
```diff
@@ -151,4 +159,78 @@ describe "Code gen: arithmetics primitives" do
       end
     {% end %}
   end
+
+  describe ".to_i conversions" do
+    {% for method, path_type in SupportedIntsConversions %}
+      {% type = path_type.resolve %}
+
+      {% if ![UInt64, Int128, UInt128].includes?(type) %}
```
RX14: `unless` is supported in macros.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -326,6 +326,21 @@ class Crystal::CodeGenVisitor
     )
   end
 
+  private def codegen_out_of_range(target_type : IntegerType, arg_type : FloatType, arg)
+    if arg_type.kind == :f32 && target_type.kind == :u128
+      # since Float32::MAX < UInt128::MAX
+      # the range checking is replaced by a positive check only
+      builder.fcmp(LLVM::RealPredicate::OLT, arg, llvm_type(arg_type).const_float(0))
```
RX14: Can't we generalize this to `arg_type.max < target_type.max`? I'd prefer express it in terms of that than explicitly code the types.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -341,6 +341,21 @@ class Crystal::CodeGenVisitor
     end
   end
 
+  private def codegen_out_of_range(target_type : FloatType, arg_type : IntegerType, arg)
+    if arg_type.kind == :u128 && target_type.kind == :f32
```
RX14: ditto, please make this check in terms of ranges instead of explicit types.

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -668,6 +683,16 @@ class Crystal::CodeGenVisitor
   end
 
   def codegen_convert(from_type : IntegerType, to_type : FloatType, arg, *, checked : Bool)
+    # TODO remove conditional after 0.28
+    if @program.has_flag?("preview_overflow")
+      if checked
+        if from_type.kind == :u128 && to_type.kind == :f32
```
RX14: this check is done twice

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -173,6 +174,90 @@ class Crystal::CodeGenVisitor
...
+
+    op_overflow = new_block "overflow"
+    op_normal = new_block "normal"
+
+    # TODO raise overflow if outside of t1 type range interpreted as a t2
+
+    cond o_bit, op_overflow, op_normal
+    position_at_end op_overflow
+
+    location = @call_location.not_nil!
```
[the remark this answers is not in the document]
bcardiff: It is assigned in `CodeGenVisitor#codegen_primitive`. Note that since currently `src/compiler/crystal/codegen/cast.cr` calls to codegen_cast, there is no always a call_location to use. The call to codegen_cast might happen implicitly.

spec/compiler/codegen/arithmetics_spec.cr
```diff
@@ -1,11 +1,13 @@
 require "../../spec_helper"
 
-{% if flag?(:bits64) %}
+{% if flag?(:darwin) %}
 SupportedInts = [UInt8, UInt16, UInt32, UInt64, UInt128, Int8, Int16, Int32, Int64, Int128]
```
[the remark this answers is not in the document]
bcardiff: Yes, I am not sure how did you check that compiler-rt is the default runtime in osx but not in linux. Definitely something we will need to include in distribution-scripts. I'm not sure if alpine packages come with everything bundled as we would need, if not https://embeddedartistry.com/blog/2017/2/20/compilerrt seems to be a good resource.

Maybe later we could avoid the `compiler_rt.cr` file entirely once the compiler-rt runtime is bundled and linked by the compiler. If so, it's better to include that lib in 0.27.1 packages. Thoughts?

spec/compiler/codegen/arithmetics_spec.cr
bcardiff: I know, but it's not worse. Before there was `||` / `&&` and I parse easier the `if` than the `unless` when they are present.

I think I prefer to keep `unless` when the have no body and stop de execution a block. Like `return unless C` or `break unless C`, because then you know that `C` holds through the rest of the statements. I found it more intuitive, but is a personal choice.

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
RX14: I meant in the sequence of commits, not in the code.

spec/compiler/codegen/arithmetics_spec.cr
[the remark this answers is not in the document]
RX14: >  I am not sure how did you check that compiler-rt is the default runtime in osx

It's just a good guess based on knowing that osx uses clang by default, not gcc.

This should probably be extracted into a different issue. The `ldc` D compiler also faces a similar problem to this (they use `compiler-rt` for profiling) and they have some code for guessing the location of `compiler-rt`. Alternatively, Crystal could require linking with `clang`, but that doesn't seem like a great solution.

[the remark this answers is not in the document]
bcardiff: That or port compiler-rt to crystal, but I would rather link against a shipped library.

src/compiler/crystal/codegen/primitives.cr
bcardiff: I like to have the intrinsics explicitly written so the can be searched. I looked for explicitly avoiding the concatenation. Other alternative would be to leave some of the names of the intrinsics commented in the function that will return them.

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
bcardiff: It is defined in 28c167896476ce15a06515e37e360c83c15f9d20 "Keep AST call operator location for primitive calls" which is the second commit in the PR. GitHub does not shows the commit in the best order after rebase it seems.

src/compiler/crystal/codegen/primitives.cr
bcardiff: That would require the compiler to support comparing 128 bits with float or doubles. And the comparison of u128 and f32 will convert the u128 to f32 so we first will need to force everything to be f64. (Actually I think that `Int128#<(Float32)` might not be sound)

On other side these check are in order to emit as few instructions as possible in the runtime check.

And there is a need for specific instructions on each types combinations.

I feel the pain :-) but I think it needs to remain as is.

src/compiler/crystal/codegen/primitives.cr
bcardiff: yes, in order to make `codegen_out_of_range(target_type : FloatType, arg_type : IntegerType, arg)` a total function. But there is no check needed if converting from Int to Float64 (or any Int but UInt128 to Float32), so is to emit less code.

Would you prefer to make `codegen_raise_overflow_cond` check if the `overflow` argument is a literal false value and avoid emitting code?

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
RX14: Look at the commits tab in this pr.

src/compiler/crystal/codegen/primitives.cr
RX14: I'm not sure I understand.

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
bcardiff: I know, that what I meant with "GitHub does not shows the commit in the best order after rebase it seems". The commits list of the branch in the fork https://github.com/bcardiff/crystal/commits/feature/preview-overflow shows a different order, and it also shows the parent of each commit.

src/compiler/crystal/codegen/primitives.cr
bcardiff: I don't see that checking `arg_type.max < target_type.max` is a generalization. Because, without checking the inner code of the `then` side, that assume that all the cases are of non negative integer types.

Making all (or this) `codegen_out_of_range` more general would require to check the number types ranges probably in double. I prefer to keep the code using only the type involving in the check (even if it's a bit less declarative). I think it's easier to follow.

RX14: OK, I can imagine how this method would get more complex if you had to check `target_type.min < arg_type.min` too.

bcardiff: For the record: I feel the pain :-)

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
RX14: That doesn't make sense at all... I guess github's UI is buggy?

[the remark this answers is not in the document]
bcardiff: I know it's been there for a while. After a rebase in a PR I stop trusting that list. But yeah, it's definitely a GitHub issue.

bcardiff:
@RX14 regarding the amount of commit I find only two to squash and meet the "made the right design decision first time".

* Refactor: Extract method out_of_range check and raise_overflow
* Refactor: Extract method for codegen check/raise of exception

I can squash them, no prob. But, did you notice anything else you would change in the commit history?

src/int.cr
```diff
@@ -285,6 +289,8 @@ struct Int
   # Raises `ArgumentError` if *exponent* is negative: if this is needed,
   # either use a float base or a float exponent.
   #
+  # Intermediate multiplication will wrap arround silently in case of overflow.
```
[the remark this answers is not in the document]
bcardiff: Thanks!

src/compiler/crystal/codegen/primitives.cr
bcardiff: Done at "Emit llvm_expect_i1_fun to favor non overflow results"

spec/spec_helper.cr
```diff
@@ -175,6 +175,7 @@ def run(code, filename = nil, inject_primitives = true, debug = Crystal::Debug::
 
     compiler = Compiler.new
     compiler.debug = debug
+    compiler.flags.concat flags if flags
```
RX14: Nice! This is compatible with my refactor in https://github.com/crystal-lang/crystal/pull/7282 :)

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -274,6 +274,7 @@ class Crystal::CodeGenVisitor
     op_overflow = new_block "overflow"
     op_normal = new_block "normal"
 
+    builder.call(llvm_expect_i1_fun, [overflow_condition, llvm_false])
```
RX14: You need to use the return value of the call in the branch instruction, no it's not documented, you need to read the code here: https://github.com/llvm-mirror/llvm/blob/master/lib/Transforms/Scalar/LowerExpectIntrinsic.cpp

RX14: Alternatively we could emit [branch weight metadata](https://llvm.org/docs/BranchWeightMetadata.html) directly, but that's a pain and best left if/when LLVM actually removes the expect intrinsic/

RX14:
There was an overflow in this CI job: https://circleci.com/gh/crystal-lang/crystal/16790

src/compiler/crystal/codegen/primitives.cr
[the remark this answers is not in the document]
asterite: Hehe, I came here to comment the same thing. So strange that GitHub put the commits in a different order...

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -274,6 +274,7 @@ class Crystal::CodeGenVisitor
     op_overflow = new_block "overflow"
     op_normal = new_block "normal"
 
+    overflow_condition = builder.call(llvm_expect_i1_fun, [overflow_condition, llvm_false])
```
asterite: Cool! I wonder what difference it makes in terms of performance, but it's nice that LLVM has this feature.

src/crystal/compiler_rt.cr
```diff
@@ -0,0 +1,38 @@
...
+    end
+    return result
+  end
+  if b == min
+    if a != 0 && a != 1
+      overflow.value = 1
+    end
+    return result
+  end
+  sa = a >> (n - 1)
```
asterite: Is it OK to use `-` here instead of `-&`? (same everywhere in this function)

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -173,6 +174,120 @@ class Crystal::CodeGenVisitor
...
+  private def codegen_out_of_range(target_type : IntegerType, arg_type : IntegerType, arg)
+    min_value, max_value = target_type.range
+    # arg < min_value || arg > max_value
+    or(
+      codegen_binary_op_lt(arg_type, target_type, arg, int(min_value, target_type)),
+      codegen_binary_op_gt(arg_type, target_type, arg, int(max_value, target_type))
+    )
+  end
+
+  private def codegen_raise_overflow
```
RX14: Since the generated code here is always the same and very common, I think it would be best to make it at least a single basic block for each function. Ideally it'd be a single `__crystal_overflow` function, with a single basic block for each function which is a call to that function. From my testing, passing `-D preview_overflow` seems to add about 50% to the build time, which isn't great.

bcardiff: You mean to have a `__crystal_overflow` well-known function that will raise the exception? What I don't like about that is that the location of the exception will not be the location of the operator. It will be along the trace.

If the slow down comes from using the `MainVisitor` maybe there is a another way to emit that fixed set of code.

Or maybe we can reuse the main visitor. There were only a pair of calls to `@program.visit_main` in the code (one with a manually created main visitor and other with a custom visitor.

asterite: I think that if a function/method doesn't have a location it's skipped from the trace. So you could try defining that function in the compiler (so it doesn't get a location) and then use it.

RX14: @bcardiff finding the exact location of the overflowing operation already depends on the backtrace line numbers surely? So nothing changed.

src/crystal/compiler_rt.cr
bcardiff: Good catch. Fixed in the last commit. I will squash that after approval.

bcardiff:
> There was an overflow in this CI job: https://circleci.com/gh/crystal-lang/crystal/16790

Unable to repro and get valuable info :-(

src/compiler/crystal/codegen/primitives.cr
bcardiff: done at "Add __crystal_raise_overflow as well known function". I think we can leave that commit not squashed.

The times of the codegen went from:
* 11.83s (master) 
* 37.14s (without __crystal_raise_overflow)
* 12.66s (with __crystal_raise_overflow)

asterite: Nice! And thank you @RX14 for the great idea!

RX14:
@bcardiff the CI failure happened twice now on different CIs, I think you should run the spec in a loop and see if you can get a repro.

bcardiff:
I squashed the compiler_rt fix.
I left the refactor `__crystal_raise_overflow` alone in a commit to keep that part of the history.
I made some changes in `codegen/debug.cr`, the offset variable is expressed in bits, although I couldn't reproduce the Overflow, the argument is UInt64, so change the operation to take place in that type, despite the other operand been Int32, Int64, etc. That should fix the overflow, as long as the values are correct.
I also fix the binding of `LibLLVM.offset_of_element` to return `UInt64`.

For the record the largest number I got was when creating debug info for LibC::PthreadMutexT an offset of 4299161654. Which is still far from the overflow point.

Let's see what the CI thinks... 🍿

Issue #7147, opened 2018-12-05
Macro raise doesn't keep location


bew:
Extracted from #6125 (https://github.com/crystal-lang/crystal/issues/6125#issuecomment-423640385)

```cr
macro raise_on(node)
  {% node.raise "node '#{node}' is here!" %}
end

macro will_raise_soon(node)
  # this is a helper macro
  raise_on {{node}}
end

macro will_raise(node)
  # this is a helper macro
  will_raise_soon {{node}}
end

will_raise :this
```
I think this should report the error `node ':this' is here!` at the last line (`will_raise :this`), without any trace (but allow to show it with `--error-trace`)

(might need a change in how macro's `raise` is reported)

---
_thought on this by @MakeNowJust:_ https://github.com/crystal-lang/crystal/pull/7008#issuecomment-434555247
> The reason what [#6125 (comment)](https://github.com/crystal-lang/crystal/issues/6125#issuecomment-423640385) does not work is for re-raising an error with node calling macro location [here](https://github.com/crystal-lang/crystal/blob/43010c806cab99290b9361790159268fa3761c76/src/compiler/crystal/semantic/semantic_visitor.cr#L417-L418):
> 
> ```crystal
>   def eval_macro(node)
>     yield
>   rescue ex : MacroRaiseException
>     node.raise ex.message, exception_type: MacroRaiseException
>   rescue ex : Crystal::Exception
>     node.raise "expanding macro", ex
>   end
> ```

bew:
Actually it doesn't even work correctly (IMO) currently with a single macro call:

For exemple, with this (normal macro raise):
```cr
macro macro_raise
  {% raise "Raised in here" %}
end

macro_raise
```
We get:
```
Error in macro_raise.cr:5: Raised in here

macro_raise
^~~~~~~~~~~

```
Which is nice.

---

Now if we raise on a specific node in a macro (doesn't work as I would expect):
```cr
macro macro_raise_on(arg)
  {% arg.raise "Raised here on #{arg}" %}
end

macro_raise_on :this
```
We get:
```
Error in macro_raise.cr:5: Raised here on :this

macro_raise_on :this
^~~~~~~~~~~~~~

```
But I would expect:
```
Error in macro_raise.cr:5: Raised here on :this

macro_raise_on :this
               ^~~~~

```

Fixing this might fix the original code (https://github.com/crystal-lang/crystal/issues/7147#issue-387545421), but I'm not sure.

asterite:
[...]

PR #7207, opened 2018-12-20
Spec: detect nesting `it`/`pending` and raise compilation error

### LINKED ISSUE #7192: pending test causes incorrect test count and malformed output
Version of crystal:
```

LLVM: 4.0.0
```

When I mark a test as pending the count of tests and test output are malformed.

I wrote the following test.
```crystal
describe "my tests" do
  it "1 equals 1.0" do
    1.should eq 1.0
  end
end
```

Running the test correctly outputs one green test displays the expected output.
screen shot 2018-12-16 at 8 24 31 am

When I mark the test as pending, things will get interesting.
```crystal
describe "my tests" do
  it "1 equals 1.0" do
    pending "something"
    1.should eq 1.0
  end
end
```

The output shows two examples and the verbose output is malformed with the word `something` showing up twice on the first example listed and only `1.0` from the test name is shown, not the full name `1 equals 1.0`.
screen shot 2018-12-16 at 8 23 49 am

The non-verbose spec output also shows two examples, although is not malformed.
screen shot 2018-12-16 at 8 25 30 am

MakeNowJust:
Fixed #7192

Fixed specs following changes also.

```crystal
require "spec"

it "foo" do
  pending
end
```

```console
$ crystal run foo.cr
Error in foo.cr:3: instantiating 'it(String)'

it "foo" do
^~

in foo.cr:3: cannot nest 'pending' of spec

it "foo" do
^~

```

NOTE: Reported location looks wrong due to #7147.

### DIFF
--- a/spec/std/colorize_spec.cr
+++ b/spec/std/colorize_spec.cr
@@ -1,7 +1,7 @@
 require "spec"
 require "colorize"
 
-private def colorize(obj, *args)
+private def colorize_wrap(obj, *args)
   obj.colorize(*args).toggle(true)
 end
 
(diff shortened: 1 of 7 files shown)


spec/std/colorize_spec.cr
```diff
@@ -17,116 +17,116 @@ private class ColorizeToS
   end
 end
 
-describe "colorize" do
+describe "colorize_wrap" do
```
Sija: it's still about `colorize`, `colorize_wrap` is just an internal helper.

spec/std/colorize_spec.cr
```diff
@@ -1,7 +1,7 @@
 require "spec"
 require "colorize"
 
-private def colorize(obj, *args)
+private def colorize_wrap(obj, *args)
```
Sija: why?

spec/std/colorize_spec.cr
MakeNowJust: It is failure of `%s/colorize/colorize_wrap`. Thank you.

spec/std/colorize_spec.cr
MakeNowJust: `Scope.colorize` is resolved in `it` blocks before file-scope `colorize` method, so it needs to change name.

MakeNowJust:
Indeed. And this check pass over nesting `it` in some cases.

```crystal
def nest_test
  it "nest" do
    (1 + 1).should eq(2)
  end
end

it "test" do
  nest_test
end
```

I'll change this to a runtime check. However, finding problem before running is useful in fact... Could anyone have another idea?

MakeNowJust:
Ok, I decide to rewrite runtime-check this. Thank you @asterite, @straight-shoota, and @bcardiff.

bcardiff:
@straight-shoota on rspec a pending with block is executed, allowing the developer to code de spec and even check that is not able to run and will fail when the code success (a bit confusing for first timers).

I think that a pending that replaces the it is enough and node/comments in the spec should be enough. I wont expect that asserting things in the middle of an in progress spec will be so much used.

PR #7211, opened 2018-12-23
Expose HTTP response headers in HTTP::WebSocket::Protocol

skateman:
Sometimes you might be curious about the HTTP response headers after a successful websocket upgrade. I haven't found any tests for the Protocol class, so I did not create any tests. However, if you really need tests for this, please let me know if I have to create a file or add something into the spec file for `HTTP::Websocket`.

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -1,5 +1,6 @@
 class HTTP::WebSocket
   getter? closed = false
+  delegate :response_headers, to: @ws
 
   # :nodoc:
   def initialize(io : IO)
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -33,7 +33,9 @@ class HTTP::WebSocket::Protocol
     size : Int32,
     final : Bool
 
-  def initialize(@io : IO, masked = false)
+  getter :response_headers
+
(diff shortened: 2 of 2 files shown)


straight-shoota:
The `Protocol` class is an implementation detail and hence not tested. So I'm not sure how this patch is supposed to be usable. People don't directly interact with `WebSocket::Protocol`. Since there are no public methods, the response headers are entirely inaccessible.

If this is to be added, it should probably be added in the `WebSocket` class.

skateman:
@straight-shoota you're right, I forgot to push the getter/delegation.

straight-shoota:
Still, the ivar should go to `WebSocket`, not `Protocol` it has nothing to do with the protocol implementation.

skateman:
@straight-shoota I agree, but as far as I understand, the response headers are only available [this](https://github.com/crystal-lang/crystal/pull/7211/files#diff-c30599f135415930b128db8e9064bbe2R242) method. So I have to expose it somehow, you're saying the instance variable is not a good solution can you please tell me in what direction should I change it? I'm not that familiar with Crystal, not sure how else can I pass these headers to the `WebSocket` class.

straight-shoota:
Why don't you open an issue first, detailling what exactly you want to achieve? 
That's probably better to discuss how this can be implemented.

straight-shoota:
Closing because this PR doesn't do what it's supposed to. The issue is tracked at #7213.

PR #7218, opened 2018-12-27
Factor powers of ten before dividing BigDecimals

### LINKED ISSUE #7122: Division of BigDecimal with unnecessary scale results in ridiculous scale
```cr
require "big/big_decimal"
var = BigDecimal.new(100, 1) / BigDecimal.new(100000000, 8)
var.value # => 1_big_i
var.scale # => 18446744073709551615_u64
```
Dividing two `BigDecimal`s with an unnecessary large scale results in a `BigDecimal` with a ridiculous scale.

​
Trying to convert such a `BigDecimal` to an `Int` will result in:
```
gmp: overflow in mpz type
Program received and didn't handle signal IOT (6)
```
```cr
require "big/big_decimal"

var = BigDecimal.new(100, 1) / BigDecimal.new(100000000, 8)
var.to_i64
```

Sija:
Fixes #7122

### DIFF
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -148,6 +148,7 @@ describe BigDecimal do
       BigDecimal.new(-1) / BigDecimal.new(0)
     end
     BigDecimal.new(1).should eq(BigDecimal.new(1) / BigDecimal.new(1))
+    BigDecimal.new(10).should eq(BigDecimal.new(100, 1) / BigDecimal.new(100000000, 8))
     BigDecimal.new(5.to_big_i, 1_u64).should eq(BigDecimal.new(1) / BigDecimal.new(2))
     BigDecimal.new(-5.to_big_i, 1_u64).should eq(BigDecimal.new(1) / BigDecimal.new(-2))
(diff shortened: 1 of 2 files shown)


spec/std/big/big_decimal_spec.cr
```diff
@@ -148,6 +148,7 @@ describe BigDecimal do
       BigDecimal.new(-1) / BigDecimal.new(0)
     end
     BigDecimal.new(1).should eq(BigDecimal.new(1) / BigDecimal.new(1))
+    BigDecimal.new(10).should eq(BigDecimal.new(100, 1) / BigDecimal.new(100000000, 8))
```
asterite: Not for this PR, but all of these assertions are written backwards: what one expects to get should be at the right-hand side.

src/big/big_decimal.cr
```diff
@@ -204,6 +204,7 @@ struct BigDecimal < Number
   # ```
   def div(other : BigDecimal, max_div_iterations = DEFAULT_MAX_DIV_ITERATIONS) : BigDecimal
     check_division_by_zero other
+    other.factor_powers_of_ten
```
Sija: Since calling `factor_powers_of_ten` is changing internal object state, should we `dup` the *other* before calling it?

Issue #4593, opened 2017-06-19
[  topic:stdlib] [community:to-design] Add support for authenticated encryption/decryption using OpenSSL


bmmcginty:
As it stands, Crystal doesn't support authenticated encryption/decryption using the OpenSSL::Cipher module.
According to
`https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption`
we need to be able to:
.set custom IV length? (not sure when this would be used, research)
.add AAD (authenticated data) using EVP_EncryptUpdate
.get/set the authentication tag with EVP_CIPHER_CTX_ctrl

I don't mind writing the code for this, but suggestions are welcome as to organization/naming.
An update_aad function could be used to set authenticated data.
I could use the iv_len call still, and just make an iv_len= call to set a custom IV lenggth.
Similarly, authentication_tag and authentication_tag= calls could get/set the authentication tag.

It looks like this will work for CCM and GCM AES modes.

Any thoughts?

danielwestendorf:
I'm working on porting Ruby on Rails' [`MessageEncryptor`](https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/activesupport/lib/active_support/message_encryptor.rb) to Crystal for safe communication between rails apps and crystal apps. It's dependent on implementation this ticket suggests.

How can I help?

My current roadblock is a lack of the `authenticated?` method.

```ruby
def new_cipher
  OpenSSL::Cipher.new(@cipher)
end

def aead_mode?
  @aead_mode ||= new_cipher.authenticated?
end
```

washu:
if your working on the Cihper you should also add teh auth tag method
@[Link("crypto")]
lib LibCrypto
	EVP_CTRL_GCM_GET_TAG = 0x10
	EVP_CTRL_GCM_SET_TAG = 0x11
	EVP_GCM_TLS_TAG_LEN  = 16
	fun evp_cipher_ctx_ctrl = EVP_CIPHER_CTX_ctrl(ctx : EVP_CIPHER_CTX, type : Int32, arg : Int32, ptr : Void*) : Int32
end
class OpenSSL::Cipher
	def auth_tag() : Bytes
	    tag = Bytes.new(LibCrypto::EVP_GCM_TLS_TAG_LEN)
[...]

PR #7223, opened 2018-12-27
Check the cipher flags to see if the cipher supports aead

danielwestendorf:
Supports checking if a cipher supports aead, with an `authenticated?` helper method.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
 
   test_darwin:
     macos:
-      xcode: "9.0"
+      xcode: "9"
     environment:
       <<: *env
       TRAVIS_OS_NAME: osx
@@ -447,4 +447,3 @@ workflows:
             - dist_darwin
             - dist_docker
             - dist_docs
-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ jobs:
 
   test_darwin:
     macos:
-      xcode: "9"
+      xcode: "9.0"
     environment:
       <<: *env
       TRAVIS_OS_NAME: osx
@@ -447,3 +447,4 @@ workflows:
             - dist_darwin
             - dist_docker
             - dist_docs
+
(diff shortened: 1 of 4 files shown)


danielwestendorf:
Need is explained [here](https://github.com/crystal-lang/crystal/issues/4593#issuecomment-414342989).

danielwestendorf:
The `darwin` CI suite seems to be failing because `aes-128-gcm` isn't an available cipher, however, when run locally on my macOS system (with Crystal installed via homebrew) it is available and the spec passes.

ysbaddaden:
CI seems to run on macOS sierra, which probably has an old OpenSSL version (maybe 0.9.8zh).

Do you know which OpenSSL versions support `aes-128-gcm`?

danielwestendorf:
@ysbaddaden I think you're on to something there. I'm trying different macOS/OSX versions to see what the oldest version this change would require is, as I can't seem to find any definitive information about when `aes-128-gcm` was added to OpenSSL.

.circleci/config.yml
```diff
@@ -447,4 +447,3 @@ workflows:
             - dist_darwin
             - dist_docker
             - dist_docs
-
```
asterite: This line should be kept.

danielwestendorf: Whoops!

spec/std/openssl/cipher_spec.cr
```diff
@@ -44,4 +44,18 @@ describe OpenSSL::Cipher do
     s3.to_s.should eq(data)
     s3.to_slice.should eq(s4.to_slice)
   end
+
+  it "authenticated?" do
+    begin
+      cipher = OpenSSL::Cipher.new("aes-128-gcm")
+      cipher.authenticated?.should eq(true)
+    rescue ArgumentError
```
straight-shoota: Maybe add a check that its actually an expected ArgumentError by verifying the error message. It should be re-raised if it doesn't match.

danielwestendorf: Could definitely do that, however if the `ArgumentError` is originating from anything other than the cipher not being available, it'd raise on the next invocation (line 58/59).

bcardiff:
I'm late to the party. The changes are fine. At most I would've try to only accept the exception as a success if the error message is certain, or wrap the whole block of the `it` in a feature check `if`.

PR #7225, opened 2018-12-28
Disallow non-invariant types in `Pointer#value=`

### LINKED ISSUE #6997: Module validation failed: Function return type does not match operand type of return inst
Found this compiler bug when duplicating and modifying test data hashes in my specs.

## Code to reproduce

```
x = {
  "y" => [] of Hash(String, String | Hash(String, String))
}
x["y"].dup.concat [{ "a" => "b" }]
```

## Compiler output

```
Module validation failed: Function return type does not match operand type of return inst!
  ret i32* %1, !dbg !36
 %"Hash(String, Hash(String, String) | String)"* (Exception)
  from main
```

## Environment

```
Crystal 0.26.1 (2018-09-26)

LLVM: 6.0.1
```

## References

Might be related to #6253

asterite:
Fixes #6997

The code in #6997 shouldn't compile at all, as I explained in the issue.

This PR changes `Pointer#value=` to disallow non-invariant types. If you have a `Pointer(Array(String | Int32))` then you must insert an `Array(String | Int32)` into it. You can't insert `Array(String)` or `Array(Int32)`.

As a second step to this I would like to change all restrictions to always check invariance for generic types. Right now we can do `x : Range(Int, Int)` as a type restriction to say "any type that inherits `Int`" but that gives the problem of #6997. We can change the semantic to mean "must be an `Int`" (in this case it doesn't make sense) and eventually introduce some syntax to mean "must be an `Int` or inherit it".

### DIFF
--- a/spec/compiler/semantic/pointer_spec.cr
+++ b/spec/compiler/semantic/pointer_spec.cr
@@ -163,4 +163,15 @@ describe "Semantic: pointer" do
       pointerof(LibFoo.extern)
       )) { pointer_of(int32) }
   end
+
+  it "doesn't allow non-invariant type in value= (#6997)" do
+    assert_error %(
+      class Gen(T)
+      end
+
+      ptr = Pointer(Gen(Char | Int32)).malloc(1_u64)
+      ptr.value = Gen(Char).new
+      ),
+      "can't insert Gen(Char) into Pointer(Gen(Char | Int32))"
+  end
 end
--- a/src/compiler/crystal/semantic/main_visitor.cr
+++ b/src/compiler/crystal/semantic/main_visitor.cr
@@ -2426,7 +2426,11 @@ module Crystal
 
       value = @vars["value"]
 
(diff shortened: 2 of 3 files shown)


asterite:
We should probably not merge this because it breaks for tuples. That is, one should be able to do:

```crystal
ptr = Pointer(Tuple(Int32 | String)).malloc(1)
ptr.value = {1}
```

That's because the compiler automatically expands "smaller" tuples to "bigger" tuples. But it doesn't do that for other types. So... we should change the matching semantic: for tuples it's fine to be covariant, for other types it's not.

asterite:
(or we can remove that extra tuple logic, it's probably better to simplify the language)

j8r:
@asterite I guess this compiles:
```cr
ptr = Pointer(Tuple(Int32 | String)).malloc(1)
ptr.value = {1.as(Int32 | String)}
```
I'm not sure `Tuple` with unions are as used as for `Array`

PR #7226, opened 2018-12-28
Add unsafe number ops

bcardiff:
This PR adds some new methods to be able to keep current wrapping/unsafe behaviour once the overflow is activated. 

New methods/constructs:
*  `value.to_iX!`/`value.to_uX!`/`value.to_fX!` ~~`value.unsafe_to_iX`/`value.unsafe_to_uX`/`value.unsafe_to_fX`~~  
* `Int32.new!(value)` ~~`Int32.unsafe_new(value)`~~. 
* `Int#&**` is also added, and `Int#**` was refactored to avoid possible overflow in last iteration.

It prepares some new primitives to be able to have cleaner compiler code in the future. That is, drop the `:cast` internal in favor of `:convert` and `:unchecked_convert` ~~`:unsafe_convert`~~ to reflect better the nature of to_X methods.

This new methods should be used to be ready for upcoming breaking changes.

### DIFF
--- a/spec/std/big/big_int_spec.cr
+++ b/spec/std/big/big_int_spec.cr
@@ -296,13 +296,13 @@ describe "BigInt" do
   it "can be casted into other Number types" do
     big = BigInt.new(1234567890)
     big.to_i.should eq(1234567890)
-    big.to_i8.should eq(-46)
-    big.to_i16.should eq(722)
+    big.to_i8!.should eq(-46)
+    big.to_i16!.should eq(722)
     big.to_i32.should eq(1234567890)
     big.to_i64.should eq(1234567890)
     big.to_u.should eq(1234567890)
-    big.to_u8.should eq(210)
-    big.to_u16.should eq(722)
+    big.to_u8!.should eq(210)
+    big.to_u16!.should eq(722)
     big.to_u32.should eq(1234567890)
 
     u64 = big.to_u64
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -46,6 +46,42 @@ describe "Int" do
     end
   end
 
+  describe "&**" do
+    it "with positive Int32" do
+      x = 2 &** 2
+      x.should eq(4)
+      x.should be_a(Int32)
+
+      x = 2 &** 0
+      x.should eq(1)
+      x.should be_a(Int32)
+    end
+
+    it "with UInt8" do
+      x = 2_u8 &** 2
+      x.should eq(4)
+      x.should be_a(UInt8)
+    end
+
+    it "raises with negative exponent" do
+      expect_raises(ArgumentError, "Cannot raise an integer to a negative integer power, use floats for that") do
+        2 &** -1
+      end
+    end
+
+    it "works with large integers" do
+      x = 51_i64 &** 11
+      x.should eq(6071163615208263051_i64)
+      x.should be_a(Int64)
+    end
+
+    it "wraps with larger integers" do
+      x = 51_i64 &** 12
+      x.should eq(-3965304877440961871_i64)
+      x.should be_a(Int64)
+    end
+  end
+
   describe "#===(:Char)" do
     it { (99 === 'c').should be_true }
     it { (99_u8 === 'c').should be_true }
--- a/spec/std/number_spec.cr
+++ b/spec/std/number_spec.cr
@@ -1,6 +1,48 @@
 require "spec"
 
+private macro it_initializes_from_value_to(number_type)
+  it "initialize from value to {{number_type}}" do
+    {{number_type}}.new(1).should be_a({{number_type}})
+    {{number_type}}.new(1).should eq(1)
+
+    {{number_type}}.new(1u32).should be_a({{number_type}})
+    {{number_type}}.new(1u32).should eq(1)
+
+    {{number_type}}.new(1.0).should be_a({{number_type}})
+    {{number_type}}.new(1.0).should eq(1)
+  end
+
+  it "unchecked initialize from value to {{number_type}}" do
+    {{number_type}}.new!(1).should be_a({{number_type}})
+    {{number_type}}.new!(1).should eq(1)
+
+    {{number_type}}.new!(1u32).should be_a({{number_type}})
+    {{number_type}}.new!(1u32).should eq(1)
+
+    {{number_type}}.new!(1.0).should be_a({{number_type}})
+    {{number_type}}.new!(1.0).should eq(1)
+  end
+end
+
 describe "Number" do
+  it_initializes_from_value_to Int8
+  it_initializes_from_value_to Int16
+  it_initializes_from_value_to Int32
+  it_initializes_from_value_to Int64
+
+  it_initializes_from_value_to UInt8
+  it_initializes_from_value_to UInt16
+  it_initializes_from_value_to UInt32
+  it_initializes_from_value_to UInt64
+
+  {% if flag?(:bits64) %}
+    it_initializes_from_value_to Int128
+    it_initializes_from_value_to UInt128
+  {% end %}
+
+  it_initializes_from_value_to Float32
+  it_initializes_from_value_to Float64
+
   describe "significant" do
     it "10 base" do
       1234.567.significant(1).should eq(1000)
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -307,28 +307,33 @@ struct BigDecimal < Number
     self
   end
 
+  # Converts to `BigInt`. Truncates anything on the right side of the decimal point.
+  def to_big_i
+    if @value >= 0
+      (@value / TEN ** @scale)
+    else
+      -(@value.abs / TEN ** @scale)
+    end
+  end
+
   # Converts to `Int64`. Truncates anything on the right side of the decimal point.
   def to_i64
-    if @value >= 0
-      (@value / TEN ** @scale).to_i64
-    else
-      -(@value.abs / TEN ** @scale).to_i64
-    end
+    to_big_i.to_i64
   end
 
   # Converts to `Int32`. Truncates anything on the right side of the decimal point.
   def to_i32
-    to_i64.to_i32
+    to_big_i.to_i32
   end
 
   # Converts to `Int16`. Truncates anything on the right side of the decimal point.
   def to_i16
-    to_i64.to_i16
+    to_big_i.to_i16
   end
 
   # Converts to `Int8`. Truncates anything on the right side of the decimal point.
   def to_i8
-    to_i64.to_i8
+    to_big_i.to_i8
   end
 
   # Converts to `Int32`. Truncates anything on the right side of the decimal point.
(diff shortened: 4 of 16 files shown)


src/compiler/crystal/codegen/primitives.cr
```diff
@@ -18,6 +18,9 @@ class Crystal::CodeGenVisitor
             when "binary"
               codegen_primitive_binary node, target_def, call_args
             when "cast"
+              # TODO 0.28.0 delete :cast
+              codegen_primitive_cast node, target_def, call_args
+            when "convert"
```
RX14: Just use `when "cast", "convert"`?

bcardiff: The convert case will have some changes in the overflow branch and is to flag that the cast should be removed

ysbaddaden:
I agree with @asterite there is nothing unsafe in those conversions. It's unchecked for over/underflow, or it's maybe wrapping, as in "be careful, it may not be the value you meant", but it's never unsafe as in "be careful, this can segfault".

Maybe just `as_x` as in `-1.as_u8` for example?

asterite:
@ysbaddaden I think `as_...` and `to_...` can be easily confused. I like that you said "unchecked", maybe `unchecked_to_i`? But it's quite long, so `unsafe` might still be a good alternative.

Sija:
@asterite `unchecked_to_...` is long, but IMO most descriptive of all the given options so far. `as_...` OTOH is short and nice but as you said could be easily confused with `to_...`.

Sija:
@RX14 but they're not, _truncating_ is way different then _overflowing_:

> **truncate**
> verb | trʌŋˈkeɪt, ˈtrʌŋkeɪt | [with object]
> - (often as adjective truncated) shorten (something) by cutting off the top or the end: _a truncated cone shape | discussion was truncated by the arrival of tea._

RX14:
@Sija you're cutting off the top bits when you cast to a smaller type. It's exactly truncation.

src/big/big_int.cr
```diff
@@ -414,6 +414,30 @@ struct BigInt < Int
...
+  def unsafe_to_i16
+    LibGMP.get_si(self).unsafe_to_i16
+  end
+
+  def unsafe_to_i32
+    LibGMP.get_si(self).unsafe_to_i32
+  end
+
+  def unsafe_to_i64
+    if LibGMP::Long == Int64 || (self <= Int32::MAX && self >= Int32::MIN)
```
j8r:
```suggestion
    if LibGMP::Long == Int64 || (Int32::MIN <= self <= Int32::MAX)
```

src/big/big_int.cr
```diff
@@ -438,6 +462,30 @@ struct BigInt < Int
...
+  def unsafe_to_u16
+    LibGMP.get_ui(self).unsafe_to_u16
+  end
+
+  def unsafe_to_u32
+    LibGMP.get_ui(self).unsafe_to_u32
+  end
+
+  def unsafe_to_u64
+    if LibGMP::Long == Int64 || (self <= Int32::MAX && self >= Int32::MIN)
```
j8r:
```suggestion
    if LibGMP::Long == Int64 || (Int32::MIN <= self <= Int32::MAX)
```

straight-shoota:
@RX14 What about when the size is kept? Converting Int32 to UInt32 doesn't truncate. It simply changes the interpretation of the byte values.

This re-interpretation is similar to `Object#unsafe_as`, except that it's not unsafe. So `#as_x` as suggested by @ysbaddaden would fit pretty well.
Yes, it could be difficult to distinguish from `#to_x`. But I'd prefer that over any of the mentioned alternatives.

Maybe `as_unchecked_x` would be a compromise but it's really long :/

Sija:
@RX14 so it seems that's a semantic issue as well, for me the result of `-256.to_u8 # => 0_u8` expression exhibits truncation + overflowing, but maybe it's because I'm lacking proper low-level IT knowledge in this regard.

j8r:
Got an idea, what about appending a exclamation mark: `#to_i32!`.
This syntax is already used for alternate methods like `p!`, `pp!`, `not_nil!`, `sort!`, `reject!`.
This can say: "Hey, beware of overflows!"

bcardiff:
* `to_unsafe_u` vs `unsafe_to_u`, what is unsafe is the conversion, not the "u". So I think it make more sense the current proposal: `unsafe_to_u`

* `unsafe_` vs `unchecked_`/`trunc_`, the unsafe means that the result might not be semantically correct. Not necessary that will segfault. There are other "unsafe" methods with that express things that the risk is not to segfault. Also the unsafe comes from keeping the semantic with C, where to_unsafe is used.

Maybe the bang alternative is not bad, `to_i32!` is like convert it to i32 no matter what, but then `unsafe_as` should be also renamed?

Since it's not a method I would expect to be used that much by the user, using a more verbose name should be fine.

src/big/big_decimal.cr
```diff
@@ -307,57 +307,91 @@ struct BigDecimal < Number
...
     if @value >= 0
-      (@value / TEN ** @scale).to_i64
+      (@value / TEN ** @scale)
     else
-      -(@value.abs / TEN ** @scale).to_i64
+      -(@value.abs / TEN ** @scale)
     end
   end
 
+  # Converts to `Int64`. Truncates anything on the right side of the decimal point.
```
r00ster91: The documentation of the unsafe and the safe versions are the same. I think that's pretty confusing.
```suggestion
  # Converts to `Int64`. Truncates anything on the right side of the decimal point.
  # Raises `OverflowError` on overflow.
```
Ditto all below.

straight-shoota:
To throw another option into the ring: `#as_x!`. It's similarly distinctive as `#to_x!` but it signals clearer that it's not converting the value but the interpretation.

I don't follow why `#unsafe_as` would need to be renamed. It is *unsafe* because it can result in invalid values, whereas the converted unchecked integers are always valid.

ysbaddaden:
@asterite it won't be used just when interacting with C but anytime you must implement a binary protocol where overflow is expected: random number generators, HTTP/2 HPACK headers (huffman compression), probably in decoding DWARF sections, ...

I'm fine with a longer name, but would prefer a rust-like `#wrapping_to_x` over `#unsafe_to_x`. It's more explicit as to what happens, and keeps `unsafe` for raw pointers, interacting with external functions (C, C++), `#unsafe_as` that casts a type to whatever other type, ... things that Crystal can't assert to be safe operations.  I don't get why a wrapping integer conversion would be unsafe.

bcardiff:
@straight-shoota regarding

> I don't follow why #unsafe_as would need to be renamed. It is unsafe because it can result in invalid values, whereas the unchecked conversions are always valid integers.

I meant for consistency. `#unsafe_as(T)` will always read the portion of memory as a `T` no matter what. `to_iX!` could be thought in the same way.

bcardiff:
I would like some more final comments or proposals in the light of my last comments regarding why `#unsafe_to_X` is still what I would choose as the conversions to go out of the crystal land.

We could have a method for truncation as public API and still have a `#unsafe_to_X` mainly for libs, effectively an alias. But that could be a separate PR.

asterite:
I think `trunc` might be a good prefix. In LLVM there's a [trunc](https://llvm.org/docs/LangRef.html#trunc-to-instruction) operation that we use when converting an integer to a smaller size. The only "problem" is that the Crystal operation can also (sign) extend the value, and also allows converting from signed to unsigned and floating point to integer so it's not really truncation in those cases.

Then, there's nothing `unsafe` about these conversion, they just might return a number that's not very intuitive.

I also thought about `reintepret` but maybe it's too long and not that good.

I'm thinking `to_i32!` and alternatives might be good. Like @bcardiff said, it reads as "convert this to an `Int32` no matter what, even if it overflows. But that doesn't mean we have to rename `unsafe_as`, which is unsafe (could very easily lead to segmentation fault if used incorrectly, while these number conversions can never lead to one). But we'd need to have the equivalent `Int32.new!(...)`.

bcardiff:
Ok so the api can be `#to_X!`, `.new!` and the primitives could be `convert` and `unchecked_convert`. 👍 before rewriting?

bcardiff:
ready `to_review!`

j8r:
@bcardiff Have you seen https://github.com/crystal-lang/crystal/pull/7226#discussion_r244425180 and https://github.com/crystal-lang/crystal/pull/7226#discussion_r244425499 ? That's minor style improvement, still it reads better :)

bcardiff:
Sorry, @j8r done!

I would like one more core team member review before merging and go to the next partial PR for overflow.

spec/std/big/big_int_spec.cr
```diff
@@ -296,13 +296,13 @@ describe "BigInt" do
   it "can be casted into other Number types" do
     big = BigInt.new(1234567890)
     big.to_i.should eq(1234567890)
-    big.to_i8.should eq(-46)
-    big.to_i16.should eq(722)
+    big.to_i8!.should eq(-46)
```
straight-shoota: Is this intended?

spec/std/int_spec.cr
```diff
@@ -46,6 +46,42 @@ describe "Int" do
...
+      x = 2 &** 2
+      x.should eq(4)
+      x.should be_a(Int32)
+
+      x = 2 &** 0
+      x.should eq(1)
+      x.should be_a(Int32)
+    end
+
+    it "with positive UInt8" do
```
straight-shoota:
```suggestion
    it "with UInt8" do
```

spec/std/int_spec.cr
```diff
@@ -46,6 +46,42 @@ describe "Int" do
...
+      x.should be_a(UInt8)
+    end
+
+    it "raises with negative exponent" do
+      expect_raises(ArgumentError, "Cannot raise an integer to a negative integer power, use floats for that") do
+        2 &** -1
+      end
+    end
+
+    it "should work with large integers" do
```
straight-shoota:
```suggestion
    it "works with large integers" do
```

spec/std/int_spec.cr
```diff
@@ -46,6 +46,42 @@ describe "Int" do
...
+      x = 51_i64 &** 11
+      x.should eq(6071163615208263051_i64)
+      x.should be_a(Int64)
+    end
+
+    it "should wrap with larger integers" do
```
straight-shoota:
```suggestion
     it "wraps with larger integers" do
```

spec/std/number_spec.cr
```diff
@@ -1,6 +1,48 @@
 require "spec"
 
+private macro it_initialize_from_value_to(number_type)
```
straight-shoota: Why is this a macro? Seems like a method would do.

straight-shoota: Minor grammar issue: `it_initializes_from_value_to` (ditto below)

src/float.cr
```diff
@@ -139,6 +139,11 @@ struct Float32
     value.to_f32
   end
 
+  # Returns an `Float32` by invoking `to_f32!` on *value*.
```
straight-shoota:
```suggestion
  # Returns a `Float32` by invoking `to_f32!` on *value*.
```

src/float.cr
```diff
@@ -223,6 +228,11 @@ struct Float64
     value.to_f64
   end
 
+  # Returns an `Float64` by invoking `to_f64!` on *value*.
```
straight-shoota:
```suggestion
  # Returns a `Float64` by invoking `to_f64!` on *value*.
```

spec/std/big/big_int_spec.cr
asterite: I think so, the same happens when you do `1234567890.to_i8`, it just truncates the contents to the right-most 8 bits (`11010010`) and reinterprets that as `Int8`, and it ends up negative because of the left-most `1`.

spec/std/int_spec.cr
Sija: one leading space too much :)

spec/std/big/big_int_spec.cr
bcardiff: It is intended as in "to keep the current behavior"

spec/std/int_spec.cr
straight-shoota: I don't know why this happens. Github editor inserts that automatically. I removed that in all the other comments, just missed this one.

spec/std/big/big_int_spec.cr
straight-shoota: Sry, that was too unspecific. I meant the change. It removes test coverage of `big.to_i8` etc.

bcardiff: It's because it will overflow on future changes. The coverage for the semantics of the numbers are not that great, and am not focusing that much in Big

spec/std/number_spec.cr
straight-shoota: Oh. That's unexpected. Why is that...

straight-shoota: Hm. I think `be_a` doesn't need to be a macro.

straight-shoota: https://carc.in/#/r/5wnn

src/big/big_decimal.cr
bcardiff: Yes it should be after OverflowError is introduced

spec/std/number_spec.cr
asterite: @straight-shoota That's true and I tried to change it in the pass and then reverted it because it currently doesn't work for actual class types: https://carc.in/#/r/5woq

The reason is that the type of `String.class` is `Class` and matching `T.class` of `Class` gives `Object` because a `Class` is an instance of an Object... it's probably not correct, but there's no way to go from `Class` to the actual instance type.

The actual problem is that the type of `"foo"` is String, and matching that with `T.class` gives the `String` **class** (called metaclass in the compiler) type, whereas when matching against `T` it gives you the String **instance** type. Now the value of `String.class` is the String **class*, but it's type is `Class`. And `Class` is a generic placeholder for all classes. Instead, we should probably make it be an endless type creation, so it would be the "String class class", and matching it against `T.class` would give the "String class". It's a bit confusing, but I think in Objective-C it works like that. I'll try it out soon.

PR #7227, opened 2018-12-28
CI: Allow travis build with no output for 20min

bcardiff:
This PR extends the timeout in travis to match the one used already in CircleCI

Ref: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - travis_retry bin/ci prepare_build
 
 script:
-  - travis_retry bin/ci build
+  - travis_wait 40 travis_retry bin/ci build
 
 branches:
   only:


RX14:
Why? Are there any builds which failed because of this?

bcardiff:
@rx14 yes, the overflow pr is reaching the timeout in travis.

RX14:
No, it failed because it ran out of memory.

straight-shoota:
Building the compiler really shouldn't take more than 20 minutes on a travis VM. If it does, I think that should be considered failing CI.

straight-shoota:
Most of that time is spent running the compiler specs, which constantly adds output.

Not having any output over a longer time only happens during compilation, i.e. while building the compiler and building the specs. But neither of them should ever take anywhere near than 20 minutes alone on a somewhat reasonably performant machine.

bcardiff:
The last build in the overflow PR shows that is a timeout while building, not running the compiler spec.

https://travis-ci.org/crystal-lang/crystal/jobs/473028069

straight-shoota:
That's failing because the timeout for maximum job runtime is reached. I thought we're only talking about the job getting cancelled because no output was received for 20 minutes.

The title is misleading. 

But `travis_wait 40` extends the overall runtime to 40 minutes, which is obviously fine. More specs take longer. And more specs are good :)

RX14:
From reading the docs, I don't think travis_wait changes the glloball timeout. It just writes lines to the log every so often. Can you apply this to your branch to make sure the fix works?

RX14:
There's no way to make a Travis job have a total runtime greater than 50 minutes. This renders `travis_retry` useless when a single try takes longer than 25 minutes. Its time to abandon Travis.

straight-shoota:
Similar issue on circleci: https://circleci.com/gh/crystal-lang/crystal/16256?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Times out after receiving no output for 10 minutes. But there were some memory issues as well right before that.

RX14:
@straight-shoota the memory issues cause a hang, which causes no build output. That build could have run forever, and not succeeded.

During a build which is making any forward progress, we don't even get close to 10 minutes with no output. And if we did, it'd be during compiling the spec suite, not during running the spec suite.

bcardiff:
Yup it was an OOM and this didn't help in the branch.

I guess we might need to disable to travis 32 bits at least.

I must confess I am starting to get tired of keep the 32 bits torch lighted on...

asterite:
I'm fine with dropping 32 bits support.

PR #7228, opened 2018-12-28
CI: Split formatting check

bcardiff:
This will allow early failure on formatting issues and keep useful spec failure in the long run.

---

I can't find the issue/PR where this was discussed. 🙈

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,20 @@ jobs:
       TRAVIS_OS_NAME: osx
     steps: *ci_steps
 
+  check_format:
+    machine: true
+    environment:
+      <<: *env
+      TRAVIS_OS_NAME: linux
+      ARCH: x86_64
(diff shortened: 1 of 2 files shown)


j8r:
Related PR: https://github.com/crystal-lang/crystal/pull/7140

ysbaddaden:
> The only "bad" thing is that the compiler is compiled twice

We could [use a workspace](https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs) to share data across jobs of the same workflow, and even share a same "executor" machine (with CircleCI 2.1). This would allow to have an initial "build" followed by "format", "std_spec" and "compiler_spec" running in parallel and reusing `.build/crystal` for example.

RX14:
The whole formatter build takes under 2 minutes. We absolutely don't need to make this more complex.

PR #7231, opened 2018-12-29
Break out standalone StatusCode class

dwightwatson:
This attempts to break out a new `HTTP::StatusCode` class that takes care of all responsibilities to do with status codes:
* It provides constants to access standard HTTP codes, so code can use human-readable constants in place of magic integers.
* It provides quick helpers to infer the type of status code.
* It takes `.default_status_message_for` off of `HTTP` and brings it inline.

### DIFF
--- a/src/http/status_code.cr
+++ b/src/http/status_code.cr
@@ -1,162 +1,111 @@
-class HTTP::StatusCode
-  # A support class that provides additional around HTTP status codes.
-  #
-  # Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
-  #
-  # It provides constants for the defined HTTP status codes as well as helper
-  # methods to easily identify the type of response.
-
-  CONTINUE                           = 100
-  SWITCHING_PROTOCOLS                = 101
-  PROCESSING                         = 102
-  EARLY_HINTS                        = 103
-  OK                                 = 200
-  CREATED                            = 201
-  ACCEPTED                           = 202
-  NON_AUTHORITATIVE_INFORMATION      = 203
-  NO_CONTENT                         = 204
-  RESET_CONTENT                      = 205
-  PARTIAL_CONTENT                    = 206
-  MULTI_STATUS                       = 207
-  ALREADY_REPORTED                   = 208
-  IM_USED                            = 226
-  MULTIPLE_CHOICES                   = 300
-  MOVED_PERMANENTLY                  = 301
-  FOUND                              = 302
-  SEE_OTHER                          = 303
-  NOT_MODIFIED                       = 304
-  USE_PROXY                          = 305
-  TEMPORARY_REDIRECT                 = 307
-  PERMANENT_REDIRECT                 = 308
-  BAD_REQUEST                        = 400
-  UNAUTHORIZED                       = 401
-  PAYMENT_REQUIRED                   = 402
-  FORBIDDEN                          = 403
-  NOT_FOUND                          = 404
-  METHOD_NOT_ALLOWED                 = 405
-  NOT_ACCEPTABLE                     = 406
-  PROXY_AUTHENTICATION_REQUIRED      = 407
-  REQUEST_TIMEOUT                    = 408
-  CONFLICT                           = 409
-  GONE                               = 410
-  LENGTH_REQUIRED                    = 411
-  PRECONDITION_FAILED                = 412
-  PAYLOAD_TOO_LARGE                  = 413
-  URI_TOO_LONG                       = 414
-  UNSUPPORTED_MEDIA_TYPE             = 415
-  RANGE_NOT_SATISFIABLE              = 416
-  EXPECTATION_FAILED                 = 417
-  MISDIRECTED_REQUEST                = 421
-  UNPROCESSABLE_ENTITY               = 422
-  LOCKED                             = 423
-  FAILED_DEPENDENCY                  = 424
-  TOO_EARLY                          = 425
-  UPGRADE_REQUIRED                   = 426
-  PRECONDITION_REQUIRED              = 428
-  TOO_MANY_REQUESTS                  = 429
-  REQUEST_HEADER_FIELDS_TOO_LARGE    = 431
-  CONNECTION_CLOSED_WITHOUT_RESPONSE = 444
-  UNAVAILABLE_FOR_LEGAL_REASONS      = 451
-  INTERNAL_SERVER_ERROR              = 500
-  NOT_IMPLEMENTED                    = 501
-  BAD_GATEWAY                        = 502
-  SERVICE_UNAVAILABLE                = 503
-  GATEWAY_TIMEOUT                    = 504
-  HTTP_VERSION_NOT_SUPPORTED         = 505
-  VARIANT_ALSO_NEGOTIATES            = 506
-  INSUFFICIENT_STORAGE               = 507
-  LOOP_DETECTED                      = 508
-  NOT_EXTENDED                       = 510
-  NETWORK_AUTHENTICATION_REQUIRED    = 511
+# A support class that provides additional around HTTP status codes.
+#
+# Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
+#
+# It provides constants for the defined HTTP status codes as well as helper
+# methods to easily identify the type of response.
+enum HTTP::StatusCode
+  CONTINUE                        = 100
+  SWITCHING_PROTOCOLS             = 101
+  PROCESSING                      = 102
+  EARLY_HINTS                     = 103
+  OK                              = 200
+  CREATED                         = 201
+  ACCEPTED                        = 202
+  NON_AUTHORITATIVE_INFORMATION   = 203
+  NO_CONTENT                      = 204
+  RESET_CONTENT                   = 205
+  PARTIAL_CONTENT                 = 206
+  MULTI_STATUS                    = 207
+  ALREADY_REPORTED                = 208
+  IM_USED                         = 226
+  MULTIPLE_CHOICES                = 300
+  MOVED_PERMANENTLY               = 301
+  FOUND                           = 302
+  SEE_OTHER                       = 303
+  NOT_MODIFIED                    = 304
+  USE_PROXY                       = 305
+  SWITCH_PROXY                    = 306
+  TEMPORARY_REDIRECT              = 307
+  PERMANENT_REDIRECT              = 308
+  BAD_REQUEST                     = 400
+  UNAUTHORIZED                    = 401
+  PAYMENT_REQUIRED                = 402
+  FORBIDDEN                       = 403
+  NOT_FOUND                       = 404
+  METHOD_NOT_ALLOWED              = 405
+  NOT_ACCEPTABLE                  = 406
+  PROXY_AUTHENTICATION_REQUIRED   = 407
+  REQUEST_TIMEOUT                 = 408
+  CONFLICT                        = 409
+  GONE                            = 410
+  LENGTH_REQUIRED                 = 411
+  PRECONDITION_FAILED             = 412
+  PAYLOAD_TOO_LARGE               = 413
+  URI_TOO_LONG                    = 414
+  UNSUPPORTED_MEDIA_TYPE          = 415
+  RANGE_NOT_SATISFIABLE           = 416
+  EXPECTATION_FAILED              = 417
+  IM_A_TEAPOT                     = 418
+  MISDIRECTED_REQUEST             = 421
+  UNPROCESSABLE_ENTITY            = 422
+  LOCKED                          = 423
+  FAILED_DEPENDENCY               = 424
+  UPGRADE_REQUIRED                = 426
+  PRECONDITION_REQUIRED           = 428
+  TOO_MANY_REQUESTS               = 429
+  REQUEST_HEADER_FIELDS_TOO_LARGE = 431
+  UNAVAILABLE_FOR_LEGAL_REASONS   = 451
+  INTERNAL_SERVER_ERROR           = 500
+  NOT_IMPLEMENTED                 = 501
+  BAD_GATEWAY                     = 502
+  SERVICE_UNAVAILABLE             = 503
+  GATEWAY_TIMEOUT                 = 504
+  HTTP_VERSION_NOT_SUPPORTED      = 505
+  VARIANT_ALSO_NEGOTIATES         = 506
+  INSUFFICIENT_STORAGE            = 507
+  LOOP_DETECTED                   = 508
+  NOT_EXTENDED                    = 510
+  NETWORK_AUTHENTICATION_REQUIRED = 511
 
   # Returns `true` if the response status code is between 100 and 199.
   def self.informational?(status_code : Int)
-    (100..199).includes?(status_code)
+    100 <= status_code <= 199
   end
 
   # Returns `true` if the response status code is between 200 and 299.
   def self.success?(status_code : Int)
-    (200..299).includes?(status_code)
+    200 <= status_code <= 299
   end
 
   # Returns `true` if the response status code is between 300 and 399.
   def self.redirection?(status_code : Int)
-    (300..399).includes?(status_code)
+    300 <= status_code <= 399
   end
 
   # Returns `true` if the response status code is between 400 and 499.
   def self.client_error?(status_code : Int)
-    (400..499).includes?(status_code)
+    400 <= status_code <= 499
   end
 
   # Returns `true` if the response status code is between 500 and 599.
   def self.server_error?(status_code : Int)
-    (500..599).includes?(status_code)
+    500 <= status_code <= 599
   end
 
   # Returns the default status message of the given HTTP status code.
-  def self.default_message_for(status_code : Int) : String
-    case status_code
-    when 100 then "Continue"
-    when 101 then "Switching Protocols"
-    when 102 then "Processing"
+  def default_message : String
+    case value
     when 200 then "OK"
-    when 201 then "Created"
-    when 202 then "Accepted"
     when 203 then "Non-Authoritative Information"
-    when 204 then "No Content"
-    when 205 then "Reset Content"
-    when 206 then "Partial Content"
     when 207 then "Multi-Status"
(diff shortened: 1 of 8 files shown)


src/http/status_code.cr
```diff
@@ -0,0 +1,162 @@
+class HTTP::StatusCode
+  # A support class that provides additional around HTTP status codes.
+  #
+  # Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
+  #
+  # It provides constants for the defined HTTP status codes as well as helper
+  # methods to easily identify the type of response.
+
+  CONTINUE                           = 100
```
r00ster91: I don't like how there need to be two HTTP status code lists now. The constants here and the `case` in `default_message_for`. It's not nice when you have to maintain two almost identical things. I wish there was a way to have only one list.

src/http/status_code.cr
```diff
@@ -0,0 +1,162 @@
...
+    when 502 then "Bad Gateway"
+    when 503 then "Service Unavailable"
+    when 504 then "Gateway Timeout"
+    when 505 then "HTTP Version Not Supported"
+    when 506 then "Variant Also Negotiates"
+    when 507 then "Insufficient Storage"
+    when 508 then "Loop Detected"
+    when 510 then "Not Extended"
+    when 511 then "Network Authentication Required"
+    else          ""
```
r00ster91: Is this really the best solution? Maybe `"Unknown Status Code"` or something?

dwightwatson:
Hey @r00ster91, thanks for the feedback. I've just been rethinking this PR and will probably resubmit.

I agree, not a huge fan of the duplication in going back and forth between a status code and it's description. I'll take a look at a better way of doing it - starting by looking at enums.

I took the list of status codes from the link that was already provided in the existing code (https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). I too noticed that I'm a teapot wasn't referenced there, so opted to stick with it.

src/http/status_code.cr
```diff
@@ -0,0 +1,162 @@
+class HTTP::StatusCode
```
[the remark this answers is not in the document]
j8r: Totally agree with @r00ster91 , the class doesn't bring anything.

src/http/status_code.cr
j8r: When you'll use an enum, this can be changed to a macro: `{% for member in @type.constants %}`
You'll have to replace `_` by a space.
I don't know for the dash, maybe a double underscore `__`

dwightwatson: Thanks for that tip - hadn't even considered macros.

Looking at the Crystal docs for enums it looked like the keys would be camelcased, like `HTTP::StatusCode::NetworkAuthenticationRequired`. I was hoping there might be some sort of `humanize` method that would give me `Network authentication required` back but doesn't seem like it. 

From the looks of it it's not particularly important that the string returned here matches the spec, but that it gives a human representation of it - so I think losing some capitalisation or a hyphen might not be an issue.

src/http/status_code.cr
straight-shoota: No, this is fine. The list doesn't change often anyway. It could obviously be dry-ed but that would cost either at runtime or compile time.

src/http/status_code.cr
```diff
@@ -0,0 +1,162 @@
...
+  def self.server_error?(status_code : Int)
+    (500..599).includes?(status_code)
+  end
+
+  # Returns the default status message of the given HTTP status code.
+  def self.default_message_for(status_code : Int) : String
+    case status_code
+    when 100 then "Continue"
+    when 101 then "Switching Protocols"
+    when 102 then "Processing"
```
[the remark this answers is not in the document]
straight-shoota: I don't see any reason two leave out any assigned codes.

418 is of course debatable, and not even strictly an HTTP status code. But many other implementations include it (node.js, Go), so why not have it in Crystal as well? This status code certainly won't be assigned for any other use case.

src/http/status_code.cr
straight-shoota: I don't think this should be implemented with macros. The list is completely fine. The returned string message should match the official description.

Enum values need to start with an upper case letter, as does any constant. But the name can be camel cased or snake cased and there are instances of both styles in the stdlib.

dwightwatson:
Hi @asterite, thanks. Was going to have another crack in a another branch but can continue working here.

One concern I've got with using an enum is the usability when trying to set a status code - because you get back an instance of the enum and have to call value on it instead

```crystal
# as a class
response.status_code = HTTP::StatusCode::NOT_FOUND

# as an enum
response.status_code = HTTP::StatusCode::NOT_FOUND.value
```

Do you have any thoughts surrounding this?

I'll match up the status codes with those listed on Wikipedia.

straight-shoota:
```cr
def status_code=(status : HTTP::StatusCode)
  status_code = status.value
end
```

j8r:
I think now it can better fit a struct:
```cr
struct HTTP::Status
  enum Code
    CONTINUE = 100
    OK = 200
    ...
    def sefl.new(status_code : Int32)
      case status_code
      when 100 then CONTINUE
      wheen 200 then
      ...
    end
  end

  getter code : Code
  def initialize(status_code : Int32)
    @code = Code.new status_code
  end
  def self.success?
    200 <= status_code <= 299
  end
  ...
end
```

straight-shoota:
@j8r Why two types instead of one?

And the status code itself should be stored as an integer, not an enum so it can be used with status codes not defined in the enum (custom ones or simply missing).

dwightwatson:
Here's what I'm tinkering with in respect to moving to an enum. I've re-worked the default message method to support the non-standard strings, and fall back to converting snakecase to titlecase.

Let me know if this is the sort of direction that you're after. I really appreciate all the feedback, this is my first real PR to Crystal.

```crystal
# A support class that provides additional around HTTP status codes.
#
# Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
#
# It provides constants for the defined HTTP status codes as well as helper
# methods to easily identify the type of response.
enum HTTP::StatusCode
  CONTINUE                        = 100
  SWITCHING_PROTOCOLS             = 101
  PROCESSING                      = 102
  EARLY_HINTS                     = 103
  OK                              = 200
  CREATED                         = 201
  ACCEPTED                        = 202
  NON_AUTHORITATIVE_INFORMATION   = 203
  NO_CONTENT                      = 204
  RESET_CONTENT                   = 205
  PARTIAL_CONTENT                 = 206
  MULTI_STATUS                    = 207
  ALREADY_REPORTED                = 208
  IM_USED                         = 226
  MULTIPLE_CHOICES                = 300
  MOVED_PERMANENTLY               = 301
  FOUND                           = 302
  SEE_OTHER                       = 303
  NOT_MODIFIED                    = 304
  USE_PROXY                       = 305
  SWITCH_PROXY                    = 306
  TEMPORARY_REDIRECT              = 307
  PERMANENT_REDIRECT              = 308
  BAD_REQUEST                     = 400
  UNAUTHORIZED                    = 401
  PAYMENT_REQUIRED                = 402
  FORBIDDEN                       = 403
  NOT_FOUND                       = 404
  METHOD_NOT_ALLOWED              = 405
  NOT_ACCEPTABLE                  = 406
  PROXY_AUTHENTICATION_REQUIRED   = 407
  REQUEST_TIMEOUT                 = 408
  CONFLICT                        = 409
  GONE                            = 410
  LENGTH_REQUIRED                 = 411
  PRECONDITION_FAILED             = 412
  PAYLOAD_TOO_LARGE               = 413
  URI_TOO_LONG                    = 414
  UNSUPPORTED_MEDIA_TYPE          = 415
  RANGE_NOT_SATISFIABLE           = 416
  EXPECTATION_FAILED              = 417
  IM_A_TEAPOT                     = 418
  MISDIRECTED_REQUEST             = 421
  UNPROCESSABLE_ENTITY            = 422
  LOCKED                          = 423
  FAILED_DEPENDENCY               = 424
  UPGRADE_REQUIRED                = 426
  PRECONDITION_REQUIRED           = 428
  TOO_MANY_REQUESTS               = 429
  REQUEST_HEADER_FIELDS_TOO_LARGE = 431
  UNAVAILABLE_FOR_LEGAL_REASONS   = 451
  INTERNAL_SERVER_ERROR           = 500
  NOT_IMPLEMENTED                 = 501
  BAD_GATEWAY                     = 502
  SERVICE_UNAVAILABLE             = 503
  GATEWAY_TIMEOUT                 = 504
  HTTP_VERSION_NOT_SUPPORTED      = 505
  VARIANT_ALSO_NEGOTIATES         = 506
  INSUFFICIENT_STORAGE            = 507
  LOOP_DETECTED                   = 508
  NOT_EXTENDED                    = 510
  NETWORK_AUTHENTICATION_REQUIRED = 511

  # Returns `true` if the response status code is between 100 and 199.
  def self.informational?(status_code : Int)
    (100..199).includes?(status_code)
  end

  # Returns `true` if the response status code is between 200 and 299.
  def self.success?(status_code : Int)
    (200..299).includes?(status_code)
  end

  # Returns `true` if the response status code is between 300 and 399.
  def self.redirection?(status_code : Int)
    (300..399).includes?(status_code)
  end

  # Returns `true` if the response status code is between 400 and 499.
  def self.client_error?(status_code : Int)
    (400..499).includes?(status_code)
  end

  # Returns `true` if the response status code is between 500 and 599.
  def self.server_error?(status_code : Int)
    (500..599).includes?(status_code)
  end

  # Returns the default status message of the given HTTP status code.
  def default_message : String
    case value
    when 200 then "OK"
    when 203 then "Non-Authoritative Information"
    when 207 then "Multi-Status"
    when 226 then "IM Used"
    when 414 then "URI Too Long"
    when 418 then "I'm a teapot"
    else
      to_s.split("_").map do |keyword|
        keyword.capitalize
      end.join(" ")
    end
  end
end
```

j8r:
@dwightwatson change `Range#includes` to `<= status_code <= ` for better performance

src/http/status_code.cr
```diff
@@ -0,0 +1,111 @@
...
+  def default_message : String
+    case value
+    when 200 then "OK"
+    when 203 then "Non-Authoritative Information"
+    when 207 then "Multi-Status"
+    when 226 then "IM Used"
+    when 414 then "URI Too Long"
+    when 418 then "I'm a teapot"
+    else
+      to_s.split("_").map do |keyword|
```
straight-shoota: Please revert to the previous implementation with all mappings hard coded. These string conversions are unnecessarily affecting runtime performance.

j8r:
Despite that, good overall (better to commit instead of writing as a comment nevertheless).
@straight-shoota and @dwightwatson  else we can have:
```cr
enum HTTP::StatusCode
  CONTINUE = 100
  OK = 200
  ...
  def self.new(status_code : Int32)
    case status_code
    when 100 then CONTINUE
    when 200 then OK
    ...
  end
  def self.success?
    200 <= self.value <= 299
  end
 ...
end
```

src/http/status_code.cr
```diff
@@ -0,0 +1,111 @@
...
+    400 <= status_code <= 499
+  end
+
+  # Returns `true` if the response status code is between 500 and 599.
+  def self.server_error?(status_code : Int)
+    500 <= status_code <= 599
+  end
+
+  # Returns the default status message of the given HTTP status code.
+  def default_message : String
```
straight-shoota: This should also be a class methods accepting an integer to use this with status codes unknown to the enumeration (returning nil in that case).
The instance method could just delegate to the class method using `value`.

straight-shoota: Or wait... `Enum.new` accepts unknown values, doesn't it?

dwightwatson: I believe it does, but this method already just returns an empty string if it doesn't recognise the status code provided. So I guess this would work if you think it looked better...

```crystal
enum StatusCode
  CONTINUE                        = 100

  def self.default_message_for(status_code : Int) : String
    case new(status_code)
    when CONTINUE then "Continue"
    else ""
    end
  end
end
```

dwightwatson:
What would you think about naming it just `HTTP::Status` - wonder if that reads better.

src/http/status_code.cr
j8r: Why we won't use OOP in this object?

dwightwatson: I'm not sure I understand, what did you have in mind?

j8r: Having an instance variable (or use the enum valuee with `self.value` ) to avoid passing `(status_code : Int)` in each class method.

src/http/status_code.cr
```diff
@@ -0,0 +1,163 @@
...
+#
+# Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
+#
+# It provides constants for the defined HTTP status codes as well as helper
+# methods to easily identify the type of response.
+enum HTTP::StatusCode
+  CONTINUE                        = 100
+  SWITCHING_PROTOCOLS             = 101
+  PROCESSING                      = 102
+  EARLY_HINTS                     = 103
```
r00ster91: This one needs to be added to the `case` as well.

RX14:
If we're going this far, why not refactor to use `StatusCode` instead of numbers throughout the code and move the class methods to instance methods etc.

dwightwatson:
Thanks @straight-shoota.

* I've renamed the enum `HTTP::Status` and made all the methods instance methods.
* I've renamed `default_message_for` to just `#description`

How might I go about validating the status code when initialising a new `HTTP::Status` instance?

Also I wonder if I should attempt to refactor `Response#status_code` in a seperate PR rather than all in one.

src/http/status.cr
```diff
@@ -0,0 +1,163 @@
...
+    when 501 then "Not Implemented"
+    when 502 then "Bad Gateway"
+    when 503 then "Service Unavailable"
+    when 504 then "Gateway Timeout"
+    when 505 then "HTTP Version Not Supported"
+    when 506 then "Variant Also Negotiates"
+    when 507 then "Insufficient Storage"
+    when 508 then "Loop Detected"
+    when 510 then "Not Extended"
+    when 511 then "Network Authentication Required"
```
j8r: You can use this macro:
```cr
enum HTTP::Status
  Continue           = 100
  SwitchingProtocols = 101
  Processing         = 102
  EarlyHints         = 103

  def description
    {% begin %}
     case value
     {% for member in @type.constants %}
     when {{member.id}}.value then "{{member.gsub(/([A-Z])/, " \\1")[1..-1]}}"
     {% end %}
     else ""
     end
    {% debug %}
    {% end %}
  end
end
puts HTTP::Status::EarlyHints.description #=> Early Hints
```
If others think that the dash matter, a gsub could be added to replace `_` to `-`, same for `Im ` to `I'm `,

straight-shoota: This is really not a good use case for macros. There is no effective benefit. It unnecessarily makes the code more complex and increases compiler work.
Also it's nice when you can grep for the message string.

straight-shoota:
> How might I go about validating the status code when initialising a new `HTTP::Status` instance?

Hm, it seems an enum's `.new` method can't be redefined 😕 I can only think of using a different method name for creating an instance from an int:
``cr
def self.code(code : Int32)
  raise "Invalid HTTP status code #{code}" unless 100 <= code <= 999
  new(code)
end
```
It obviously can't enforce this limit when `.new` is called in user code. I don't think there is a way to make `.new` protected, either.
But maybe someone has a better idea.

> Also I wonder if I should attempt to refactor `Response#status_code` in a seperate PR rather than all in one.

I'd say it fits here. It's only a small change and conceptually closely related.

File: src/http/status.cr [hunk 9]
```diff
@@ -0,0 +1,163 @@
...
+    else          ""
```
straight-shoota: Maybe this should return `nil`?

dwightwatson:
Looking into it more - sort of reluctant to make the `Response#status_code` to `Response#status` code change without a little more buy-in. It would change the public API for both client and server responses, making it a breaking change. The same thing applies to an empty string being returned from `default_message_for`. Though, I suppose if you relied on `default_message_for` then this is already a breaking change.

straight-shoota:
Yes, it's a breaking change. `default_message_for` is public API. So this won't be merged for `0.27.1` anyway.

src/http/status.cr
j8r: macro aren't designed to reduce boilerplate?

src/http/status.cr
j8r: It could be `description?` that returns `String?` and `description` that raises on `nil`.

src/http/status.cr
j8r: This also prevent https://github.com/crystal-lang/crystal/pull/7231#discussion_r244495514 . I don't see a use case for grep, you can search in the enum values directly.

src/http/status.cr
straight-shoota: Why should that method ever raise?? It should either return `nil` or empty string. That's only a minor semantic difference.

src/http/status.cr
straight-shoota: You shouldn't try to rewrite something into a macro, when plain code is perfectly fine.

src/http/status.cr
j8r: empty string is fine then, dealing with nil would add an annoyance for nothing.

straight-shoota: Depends. Having to distinguish empty string and nil is also annoying. It's often better to return `nil` which clearly states that there is no value available. Whereas empty string seems like there is something, but it has no content.

src/http/status.cr
```diff
@@ -0,0 +1,163 @@
...
+  NOT_IMPLEMENTED                 = 501
+  BAD_GATEWAY                     = 502
+  SERVICE_UNAVAILABLE             = 503
+  GATEWAY_TIMEOUT                 = 504
+  HTTP_VERSION_NOT_SUPPORTED      = 505
+  VARIANT_ALSO_NEGOTIATES         = 506
+  INSUFFICIENT_STORAGE            = 507
+  LOOP_DETECTED                   = 508
+  NOT_EXTENDED                    = 510
+  NETWORK_AUTHENTICATION_REQUIRED = 511
```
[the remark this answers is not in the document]
RX14: These are enum values, not constant names, I always assumed the standard was camel case.

Issue #7189, opened 2018-12-15
each_cons should be able to use deques  when reusing the buffer. 


yxhuvud:
So, for some reasons I investigated the performance of `each_cons`, with the reuse flag set, and what would happen if the reused `Array` was replaced by a `Deque`. The results, as far as I can see, points to using Deque being consistently faster except when working on arrays with very few elements. Low counts, like 2, does not seem to impact the conclusion (though counts in the hundreds make Deque perform even better in comparison, but I didn't include those results). 

**Suggestions based on below benchmarks**: 
1: At the very least, each_cons should be able to accept a buffer to reuse that isn't bound to be an `Array` like today.
2: Change default choice with the reuse flag set to be Deque 

Regarding 2, it would be interesting to have good benchmarks for the case when reuse is false. I tried to run the benchmarks with those flags set, but got results that was a lot closer, and where the winner each time seemed more or less random. I guess GC pauses affected that one pretty hard.

Thoughts? 

I'll be happy to provide code if we can agree on scope.

Disclaimer: the machine I test on is something like 5 years old. Other machines may have other results.

-**Full benchmark below:**

```
require "benchmark"

class Array
  def each_cons2(count : Int, reuse = false)
    each.cons2(count, reuse)
  end

  def each_cons2(count : Int, reuse = false)
    if reuse
      unless reuse.is_a?(Deque)
        reuse = Deque(T).new(count)
      end
      cons = reuse
    else
      cons = Deque(T).new(count)
      reuse = nil
    end

    each do |elem|
      cons << elem
      cons.shift if cons.size > count
      if cons.size == count
        if reuse
          yield cons
        else
          yield cons.dup
        end
      end
    end
    nil
  end
end

module Iterator(T)
  def cons2(n : Int, reuse = false)
    raise ArgumentError.new "Invalid cons size: #{n}" if n <= 0
    Cons2(typeof(self), T, typeof(n)).new(self, n, reuse)
  end

  struct Cons2(I, T, N)
    include Iterator(Array(T))
    include IteratorWrapper

    def initialize(@iterator : I, @n : N, reuse)
      if reuse
        if reuse.is_a?(Deque)
          @values = reuse
        else
          @values = Deque(T).new(@n)
        end
        @reuse = true
      else
        @values = Deque(T).new(@n)
        @reuse = false
      end
    end

    def next
      loop do
        elem = wrapped_next
        @values << elem
        @values.shift if @values.size > @n
        break if @values.size == @n
      end

[...]

PR #7233, opened 2018-12-29
Enumerable#each_cons support for accepting deque as a reuse buffer

yxhuvud:
This removes the restriction on `Enumerable#each_cons` and `Iterator#cons` to only accept arrays as a container to reuse. With this, it will be possible to use a Deque instead, for better performance.

Also adds some missing tests in iterator_spec.cr

Solves #7189

### DIFF
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -286,6 +286,15 @@ describe "Enumerable" do
       a.should be(reuse)
     end
 
+    it "returns each_cons iterator with reuse = deque" do
+      reuse = Deque(Int32).new
+      iter = [1, 2, 3, 4, 5].each_cons(3, reuse: reuse)
+
+      a = iter.next
+      a.should eq(Deque{1, 2, 3})
+      a.should be(reuse)
+    end
+
     it "yields running pairs with reuse = true" do
       array = [] of Array(Int32)
       object_ids = Set(UInt64).new
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -306,6 +306,16 @@ describe "Enumerable" do
       end
       array.should eq([[1, 2], [2, 3], [3, 4]])
     end
+
+    it "yields running pairs with reuse = deque" do
+      array = [] of Deque(Int32)
+      reuse = Deque(Int32).new
+      [1, 2, 3, 4].each_cons(2, reuse: reuse) do |pair|
+        pair.should be(reuse)
+        array << pair.dup
+      end
+      array.should eq([Deque{1, 2}, Deque{2, 3}, Deque{3, 4}])
+    end
   end
 
   describe "each_slice" do
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -286,7 +286,7 @@ describe "Enumerable" do
       a.should be(reuse)
     end
 
-    it "returns running pairs with reuse = true" do
+    it "yields running pairs with reuse = true" do
       array = [] of Array(Int32)
       object_ids = Set(UInt64).new
       [1, 2, 3, 4].each_cons(2, reuse: true) do |pair|
@@ -297,7 +297,7 @@ describe "Enumerable" do
       object_ids.size.should eq(1)
     end
 
-    it "returns running pairs with reuse = array" do
+    it "yields running pairs with reuse = array" do
       array = [] of Array(Int32)
       reuse = [] of Int32
       [1, 2, 3, 4].each_cons(2, reuse: reuse) do |pair|
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -283,22 +283,21 @@ module Enumerable(T)
   #
   # By default, a new array is created and yielded for each consecutive slice of elements.
   # * If *reuse* is given, the array can be reused
-  # * If *reuse* is an `Array`, this array will be reused
-  # * If *reuse* is truthy, the method will create a new array and reuse it.
+  # * If *reuse* is `true`, the method will create a new array and reuse it.
+  # * If *reuse*  is an instance of `Array`, `Deque` or a similar collection type (implementing `#<<`, `#shift` and `#size`) it will be used.
+  # * If *reuse* is falsey, the array will not be reused.
   #
   # This can be used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def each_cons(count : Int, reuse = false)
-    if reuse
-      unless reuse.is_a?(Array)
-        reuse = Array(T).new(count)
-      end
-      cons = reuse
+    if reuse.nil? || reuse.is_a?(Bool)
+      each_cons_internal(count, reuse, Array(T).new(count)) { |slice| yield slice }
     else
-      cons = Array(T).new(count)
-      reuse = nil
+      each_cons_internal(count, true, reuse) { |slice| yield slice }
     end
+  end
 
+  private def each_cons_internal(count : Int, reuse, cons)
     each do |elem|
       cons << elem
       cons.shift if cons.size > count
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -290,6 +290,7 @@ module Enumerable(T)
   # This can be used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def each_cons(count : Int, reuse = false)
+    raise ArgumentError.new "Invalid cons size: #{count}" if count <= 0
     if reuse.nil? || reuse.is_a?(Bool)
       each_cons_internal(count, reuse, Array(T).new(count)) { |slice| yield slice }
     else
--- a/src/iterator.cr
+++ b/src/iterator.cr
@@ -293,34 +293,30 @@ module Iterator(T)
   # iter.next # => Iterator::Stop::INSTANCE
   # ```
   #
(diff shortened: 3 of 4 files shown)


src/iterator.cr
```diff
@@ -297,30 +297,26 @@ module Iterator(T)
...
   # * If *reuse* is an `Array`, this array will be reused
   # * If *reuse* is truthy, the method will create a new array and reuse it.
+  # * If *reuse* is falsey, the array will not be reused.
   #
   # This can be used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def cons(n : Int, reuse = false)
     raise ArgumentError.new "Invalid cons size: #{n}" if n <= 0
-    Cons(typeof(self), T, typeof(n)).new(self, n, reuse)
+    if reuse.is_a?(Bool)
```
bcardiff: If reuse is nil this might be falling in the undesired branch. The docs state about falsey values.

yxhuvud: Good point. Added a check for nil and a test for that case.

src/enumerable.cr
```diff
@@ -281,14 +281,15 @@ module Enumerable(T)
   #
   # By default, a new array is created and yielded for each consecutive slice of elements.
   # * If *reuse* is given, the array can be reused
-  # * If *reuse* is an `Array`, this array will be reused
-  # * If *reuse* is truthy, the method will create a new array and reuse it.
+  # * If *reuse* is `true`, the method will create a new array and reuse it.
+  # * If *reuse* is a truthy, the passed `Array` or `Deque` will be used.
```
RX14: typo: a truthy

src/iterator.cr
```diff
@@ -297,30 +297,26 @@ module Iterator(T)
...
-          @values = Array(T).new(@n)
-        end
-        @reuse = true
-      else
-        @values = Array(T).new(@n)
-        @reuse = false
-      end
+    def initialize(@iterator : I, @n : N, values : V, reuse)
+      @values = values
+      @reuse = reuse ? true : false
```
RX14: Consider `@reuse = !!reuse`, it seems to be a fairly common idiom, are there objections?

straight-shoota:
Should `reuse == true` use a `Deque` as default for better performance than array?
Or does it need the standard collection type?

RX14:
@straight-shoota it'd break a lot of code, since a lot of code only takes arrays.

src/iterator.cr
bcardiff: No objections on my side, but we don't need to enforce it IMO.

yxhuvud: I have no opinion on the matter, except that I didn't like spending 5 lines on it which is what the original code looked like after my changes in the logic.

yxhuvud:
Should I rebase it down to three commits again?

RX14:
yeah

src/iterator.cr
```diff
@@ -297,30 +297,26 @@ module Iterator(T)
   # * If *reuse* is given, the array can be reused
   # * If *reuse* is an `Array`, this array will be reused
```
straight-shoota: This should be updated to allow `Array`, `Deque` or any collection type with a matching interface.

spec/std/enumerable_spec.cr
```diff
@@ -286,7 +286,16 @@ describe "Enumerable" do
       a.should be(reuse)
     end
 
-    it "returns running pairs with reuse = true" do
+    it "returns each_cons iterator with reuse = deque" do
+      reuse = Deque(Int32).new
+      iter = [1, 2, 3, 4, 5].each_cons(3, reuse: reuse)
+
+      a = iter.next
+      a.should eq(Deque(Int32).new([1, 2, 3]))
```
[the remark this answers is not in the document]
straight-shoota: Any type with argless constructor and `#<<`. See https://crystal-lang.org/reference/syntax_and_semantics/literals/array.html#array-like-type-literal

spec/std/enumerable_spec.cr
```diff
@@ -306,6 +315,16 @@ describe "Enumerable" do
       end
       array.should eq([[1, 2], [2, 3], [3, 4]])
     end
+
+    it "yields running pairs with reuse = deque" do
+      array = [] of Array(Int32)
+      reuse = Deque(Int32).new
+      [1, 2, 3, 4].each_cons(2, reuse: reuse) do |pair|
+        pair.should be(reuse)
+        array << pair.to_a
```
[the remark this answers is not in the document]
yxhuvud: Because Deque doesn't do structural equality. I tried the way you suggested at first, but `Deque(Foo).new != Deque(Foo).new`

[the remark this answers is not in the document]
straight-shoota: It should. See https://carc.in/#/r/5xoj

src/iterator.cr
yxhuvud: In practice, the only ones I found in stdlib that have `<<`, `shift` and `size` seems to be `Array` and `Deque`. I suppose this is where java or whatever would have reached for a Qeueable interface.

I'll synch the texts of the two methods, though.

spec/std/enumerable_spec.cr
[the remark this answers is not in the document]
yxhuvud: Hmm. Changing it exposed a type issue. I'll try to figure out a way to fix that.

src/iterator.cr
straight-shoota: Yeah, the stdlib has only those two matching collections. But it should also be usable with any custom collection type.

spec/std/enumerable_spec.cr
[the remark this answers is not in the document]
yxhuvud: Reused the solution from `Iterator`. Good enough?

src/enumerable.cr
```diff
@@ -281,22 +281,22 @@ module Enumerable(T)
   #
   # By default, a new array is created and yielded for each consecutive slice of elements.
   # * If *reuse* is given, the array can be reused
-  # * If *reuse* is an `Array`, this array will be reused
-  # * If *reuse* is truthy, the method will create a new array and reuse it.
+  # * If *reuse* is `true`, the method will create a new array and reuse it.
+  # * If *reuse* is truthy, the passed `Array` or `Deque` will be used.
```
straight-shoota: An integer would also be truthy, but can't be passed.
I'd suggest something along:
```
If *reuse*  is an instance of `Array`, `Deque` or a similar collection type (implementing `#<<`, `#shift` and `#size`) it will be used.
```

We should probably have an abstract module for a generic list interface.

yxhuvud: Well, updated, and that list interface is way out of scope.

src/enumerable.cr
```diff
@@ -283,22 +283,22 @@ module Enumerable(T)
...
   #
   # This can be used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def each_cons(count : Int, reuse = false)
-    if reuse
-      unless reuse.is_a?(Array)
-        reuse = Array(T).new(count)
-      end
-      cons = reuse
+    raise ArgumentError.new "Invalid cons size: #{count}" if count <= 0
```
straight-shoota: Please call it `Invalid count`, that's the name of the argument.

src/iterator.cr
```diff
@@ -293,34 +293,30 @@ module Iterator(T)
...
   # * If *reuse* is given, the array can be reused
-  # * If *reuse* is an `Array`, this array will be reused
-  # * If *reuse* is truthy, the method will create a new array and reuse it.
+  # * If *reuse* is `true`, the method will create a new array and reuse it.
+  # * If *reuse*  is an instance of `Array`, `Deque` or a similar collection type (implementing `#<<`, `#shift` and `#size`) it will be used.
+  # * If *reuse* is falsey, the array will not be reused.
   #
   # This can be used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def cons(n : Int, reuse = false)
```
straight-shoota: This argument should also be called `count` to line up with `Enumerable#each_cons`. But that would be a breaking change (and I don't think the other changes are breaking), so it can be left for another PR.

src/iterator.cr
```diff
@@ -293,34 +293,30 @@ module Iterator(T)
...
+    if reuse.nil? || reuse.is_a?(Bool)
+      Cons(typeof(self), T, typeof(n), Array(T)).new(self, n, Array(T).new(n), reuse)
+    else
+      Cons(typeof(self), T, typeof(n), typeof(reuse)).new(self, n, reuse, reuse)
+    end
   end
 
-  private struct Cons(I, T, N)
+  private struct Cons(I, T, N, V)
     include Iterator(Array(T))
```
straight-shoota: I think this should be changed to `Iterator(V)`?

src/iterator.cr
Sija: Pedantic: there's a double space in the comment above.

straight-shoota: I noticed it, too, but didn't want to be too pedantic :P

src/iterator.cr
yxhuvud: Yes. But I wonder though, how to trigger the difference in a test.

src/iterator.cr
straight-shoota: @Sija, that would be breaking. Let's change it in another PR.

yxhuvud:
(rebased)

PR #7236, opened 2018-12-30
[WIP] Zip64 writer+reader

### LINKED ISSUE #7103: Adding Zip64 support to Zip::Writer
We have ported our zip_tricks library to Crystal at least in as far as writing ZIPs is concerned. What we needed was Zip64 support and the ability to manually "simulate" writes for keeping offsets. The low-level implementation differs slightly from Crystal's ZIP module, but the high-level API is very, very similar (a streaming Writer object that handles any IO within a block). The low-level difference is that there is a single Writer object that handles low-level byte packing and writing, and that is the object that handles the tricky logic with switching Zip64 on and off, setting the correct masked permission bits on things and so forth. The shard is in https://github.com/WeTransfer/cr_zip_tricks

So, in light of #5215 etc would it be something that belongs in the Crystal stdlib? I could potentially port the writing to the standard library, but I would prefer to then remove the "self writing" feature in `Entry` so that the serializer we already have can be reused and stay in one module. The Entry object then gets a much smaller API surface as well. @asterite would love to know your thoughts on this

julik:
Opened to track progress, not much to see yet. Closes #7103


spec/std/zip/zip_serializer_spec.cr
```diff
@@ -0,0 +1,533 @@
...
+    @io.rewind
+  end
+
+  def read_1b
+    slice = read_n(1)
+    slice[0]
+  end
+
+  def read_2b
+    @io.read_bytes(UInt16, format = IO::ByteFormat::LittleEndian)
```
Sija: `format =` -> `format:` (former way is declares local variable named `format` and passes it as an positional argument, while the latter passes `format` as named argument, which I reckon is what you wanted)

ditto all below

spec/std/zip/zip_serializer_spec.cr
```diff
@@ -0,0 +1,533 @@
...
+
+      br.read_2b.should eq(0x5455) # Extended timestamp extra tag
+      br.read_2b.should eq(5)      # Size of the timestamp extra
+      br.read_1b.should eq(128)    # The timestamp flag
+
+      ext_mtime = br.read_4b_signed
+      ext_mtime.should eq(1_468_763_280) # The mtime encoded as a 4byte uint
+
+      parsed_time = Time.unix(ext_mtime)
+      parsed_time.year.should eq(2_016)
```
Sija: `2_016` -> `2016`

julik:
@Sija Maybe you could advise. I changed `Writer#add` to be more predictable in its behavior ("always use data descriptor with a block, do not use data descriptor without and assume the Entry is correct"). The problem is in the _reading_ of the entries. The current reader in the stdlib assumes it is possible to read the ZIP file "straight ahead", which in a way is a good idea if you want to, say, open a ZIP right when it is being uploaded to you by the client (neat feature! smart! no buffering!) It is however almost impossible to do it correctly in the presence of data descriptors - your only way to catch where the file entry body ends and the file descriptor begins is to scan for the 0x08074b50 byte sequence. In practice it can (and will) also occur in file bodies. It will also occur inside ZIPs inside ZIPs and so forth.

The consensus so far seems to be that it is not possible to open such ZIP files without having a seekable input. Our solution was to actually read the ZIP from the "tail", starting with the central directory, using that to recover offsets and sizes within the entire ZIP file being read, and then address those offsets directly. This works very well. We ended up only using the "straight ahead" reading for recovering truncated ZIP files where no central directory could be found.

With the changes I made, there is a bit of an embarrassing situation that we are trying to read our own written file with the reader and failing, because the reader is trying to do the impossible (do straight ahead read with data descriptors). Possible courses of action:

* Scan for the data descriptor signature in the input stream and hope for the best (this _is_ not a good approach, really)
* Change the ZIP reader to require a seekable IO. This is a breaking API change, but a necessary one. Then rewrite the entire reading code for "from tail" seeks. Doable, but change will be much larger and this Reader class alone in zip_tricks is pushing a few hundred lines
* Make the existing reader fail explicitly when it encounters an entry with bit 3 set in the local headers, since once you have hit one you don't really know how many bytes you need to skip to end up at the next entry

If I have to do both reader _and_ writer changes at the same time I will pretty much have to do 2 + 3, and the changes are _big_. I would prefer doing first the writing correctly, then the reading in a separate PR. But the problem, if not addressed, will remain - that it is impossible to reliably forward-decode a ZIP file in presence of data descriptors. In zip_tricks we split the two code paths completely and it is up to the user of the code to decide whether to use the central directory reading approach (much faster, much more reliable _and_ allows random access to entries) or use straight-ahead parsing (potentially unreliable, may lead to extracting corrupt data unless every entry is buffered over the entire remaining body of the ZIP and then verified for length).

Sija:
@julik that's a good question! I'd like to say more, but researching the subject without a computer at hand makes is a real PITA, let me dig into that and come back to you around end of the week, once rejoined with a bigger screen. In the meantime I wish you a great New Year! 🍾🎊🌲

RX14:
I think that probably the best approach is to support both option 2 and option 3 as modes, with option 2 as the default.

julik:
@RX14 🤷‍♂️ Let me know when there is a decision on this. I would be happy to maintain a Zip shard, but I got no idea what the current requirements regarding API compatibility requirements are etc

j8r:
I still think a modular approach is better. It should be a library that *can* be included, or not, in to the stdlib.

RX14:
@julik If there's a good reason, then compatibility should be broken, but obviously kept at a minimum. So changing to require a seekable IO is fine, since it's the most sane thing to do. I'd still like an option to read straight-ahead if the user can assert that there are no data descriptors, but that can come in a later PR if you wish. 

If it's at all easier for you to split into reader+writer PRs then by all means do so.

Splitting into a shard can come later, we should move forward with this PR for now.

asterite:
Note that `Zip::File` already requires an IO to be seekable. I think requiring that for `Zip::Reader` is fine too. In fact it also happens that in Go it's exactly like that.

julik:
Closing in favor of https://github.com/crystal-lang/crystal/pull/11396

PR #7239, opened 2018-12-31
Implement StringLiteral#count

Blacksmoke16:
Implements StringLiteral#count.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -292,6 +292,10 @@ module Crystal
         assert_macro "", %({{"hello".size}}), [] of ASTNode, "5"
       end
 
+      it "executes count" do
+        assert_macro "", %({{"aabbcc".count('a')}}), [] of ASTNode, "2"
+      end
+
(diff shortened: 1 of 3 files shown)


bcardiff:
I’m courious, was this feature needed in some use case?

Blacksmoke16:
I needed it to count the number of placeholder values in a URL path, i.e. `"/users/:user_id/posts/:post_id".count (':')`.

Other way to do it would be like `"path".chars.select { |chr| chr == ':' }.size`

j8r:
@Blacksmoke16 you can do `"/users/:user_id/posts/:post_id".split(':').size}}`

straight-shoota:
At some point we need to establish some consistent policy how to decide which methods should be available in the macro language. We shouldn't port all of `String` to `StringLiteral`.

But I don't see a reason why not to merge this and it has a use case. There are workarounds, but they obscure meaning.

asterite:
Yes, right now there's no clear policy about which macro methods should exit. However, we should probably have all methods from string and all non-mutating methods from array/hash to allow simpler macro code to be written.

PR #7241, opened 2018-12-31
Fix error message on crystal tool format

MakeNowJust:
Formatting error message is wrong. For example it shows:

> `$ crystal tool format foo.cr`

But the above command makes the same result because single argument
special case is removed by #7144.

STDIN case `-` is still alived, so I think the right message is:

> `$ crystal tool format - < foo.cr`

/cc @straight-shoota

### DIFF
--- a/src/compiler/crystal/command/format.cr
+++ b/src/compiler/crystal/command/format.cr
@@ -193,7 +193,7 @@ class Crystal::Command
 
       ex.inspect_with_backtrace STDERR
     else
-      STDERR << "there's a bug formatting #{file}, to show more information, please run:\n\n  $ crystal tool format #{file}"
+      STDERR << "there's a bug formatting #{file}, to show more information, please run:\n\n  $ crystal tool format - < #{file}"
     end
 
     STDERR.puts


straight-shoota:
A bit of context:
When the formatter encounters an exception which is not a syntax error it only prints this:
```
there's a bug formatting #{file}, to show more information, please run:

  $ crystal tool format #{file}
```

When formatting a single file before #7144, the exception would be printed directly, hence the advise to run the formatter on a single file.


I don't think this is the right solution. Piping the file through stdin is a hack an shouldn't be advertised as a means to get more verbose error information.

Either errors should be printed in more detail right away, or this could be activated by a flag like `--error-trace`.

I'm not sure about the reasoning why error messages are hidden in the first place.

MakeNowJust:
> I'm not sure about the reasoning why error messages are hidden in the first place.

Because it makes confusing when many files causes formatter bugs. It is for preventing to report independent formatter bugs as a single issue.

And, `--error-trace` is bad name at least. It is confusing about compiler's `--error-trace` flag.

However I'm not sure this behavior is right. IMHO to show detailed error and exit immediately (like ICE) is better...?

bcardiff:
I think that in case of an exception we should abort and print details information if possible. And suggest the user to isolate the file / submit bug.

straight-shoota:
@MakeNowJust I meant `--error-trace` as a reference to the compiler option. It should obviously be called differently for the formatter. :+1:

What's the downside of outputting all errors instead of aborting on the first one? In the rare case that there are lots of errors, the output might be long, but it's also complete. You don't need to annoyingly iterate on avoiding formatter bugs you can't directly fix anyway.

I'd prefer to print all errors. Whether it's full error output or not can be configurable using a CLI option. We only need to decide which should be the default.

straight-shoota:
I think we can trust the formatter developer to use an appropriate option such as `--hide-errors` or `--fail-fast`. That's not a huge burden.

Every-day users shouldn't hopefully encounter formatter bugs often, but when it happens, the output should be complete.

MakeNowJust:
@straight-shoota When and who feels useful to this complete log? I can't understand it.

straight-shoota:
As long as it's not failing on the first error, I'm fine with either verbosity on default.

MakeNowJust:
[rufo](https://github.com/ruby-formatter/rufo) behaves fail-fast for unexpected internal error in fact.

@asterite Could you have an opinion to this?

bew:
If we go that way, there should probably be a flag like `--skip-error-files` to not format files that errors and still format other files

straight-shoota:
I wouldn't want a formatter bug which typically only occurs in maybe one or two files at most to prevent formatting all other files. Or make it rather complicated by having to manually exclude failing files.

MakeNowJust:
`--skip-error-files` flag is really distrusting. It says the formatter have many bugs (it is true, but, it is not a sales point of the formatter...) So, to extend `Dir.glob` to support excluding pattern is right way.

straight-shoota:
> So, to extend `Dir.glob` to support excluding pattern is right way.

I don't follow... why should a Crystal user have to bother manually excluding individual files just because the formatter shows a bug with them? The formatter can and should handle this gracefully on it's own.

`--skip-error-files` should just be the default behavior instead of a flag.

MakeNowJust:
> why should a Crystal user have to bother manually excluding individual files just because the formatter shows a bug with them? 

Because this is just a bug. A bug bother us every time. If a bug does not bother us, we may lose a passion to get rid of bugs 😄 

Nevertheless, a way to skip file involving formatter bug is needed. When someone invoke formatter bug under a project, but the project must go on developing. It is needed for continuing this until next release including a bug fix, if we cannot escape from a bug anyway.
I considered an hour... and now, I feel `--skip-error-files` is acceptable. At least it is good for this purpose. However I'm unbelieving the case that "we cannot escape from a bug anyway". În my experience, we can escape from formatter bug by modifying target file little...

straight-shoota:
A bug should bother compiler developers. But not ordinary developers just trying to format their non-compiler Crystal files. The difference in numbers as well as the understanding of the compiler and its tools probably covers several orders of magnitude.

>  În my experience, we can escape from formatter bug by modifying target file little...

Sure, but that can be hard for someone without your experience. At the very least it's annoying and doesn't have a particularly compelling reason from the user's perspective. It's unlikely a typical developer would start investigating a compiler bug when he just wants to format some files.

MakeNowJust:
There is two reasons that I stick with fail-fast and deny changing `--skip-error-files` as default.

1. Generally a program abort on fatal error, and a bug is fatal always. The formatter should follow this mannar. And, if `--skip-error-files` is default, the formatter will succeed when a bug is found. It seems terrible.
2. When a user encounter a formatter bug, I expect the user will submit a new issue. I think the formatter can report the best suitable information for this on fail-fast. Meanwhile, `--skip-error-files` as default inhibit this.

@straight-shoota's opinion is surely right. It reduces frastrations by users who encounter a formatter bug. However, a typical user does't encounter a formatter bug, and we lose chances to find a formatter bug by `--skip-error-files` as default. Is it valuable?

straight-shoota:
1. A bug in the formatter is not a fatal error. It only affects a single file and the program can continue without any issues.
2. I don't know how many users will actually open an issue when they encounter an error. But if they do, the error trace can be easily printed by providing a specific flag.

I think the behaviour should stay mostly as it is, only change from single file mode to show the full trace to require a flag.  Essentially, that would only add a flag and print that with `crystal tool format #{file}` for showing more information.

MakeNowJust:
> I think the behaviour should stay mostly as it is, only change from single file mode to show the full trace to require a flag.

It is highly better than `--skip-error-files` as default.

straight-shoota:
@MakeNowJust Oh, that's actually what I have all along assumed "`--skip-error-files` by default" would do. Seems like a misunderstanding, then.

straight-shoota:
It should obviously return non-zero status code when there is a bug.

MakeNowJust:
Now, `--show-backtrace` option is added. It show backtrace on a bug when it is specified, and default bug error message uses this.

In addition, status code is also fixed. `crystal tool format` returns non-zero status code When there are some errors.

src/compiler/crystal/command/format.cr
```diff
@@ -69,27 +75,24 @@ class Crystal::Command
...
+        when .invalid_byte_sequence?
+          error "'#{result.filename}' is not a valid Crystal source file", exit_code: nil
+        when .bug?
+          error "there's a bug formatting '#{result.filename}'", exit_code: nil
         end
-        exit 1
       end
     end
+
+    exit check_files.empty? ? 0 : 1
```
bcardiff: So now if check is false, and there are files with syntax error or invalid byte sequence the formatter will return a non zero status.

I am not sure that an element should be appended if `check == false` when handling `ex : InvalidByteSequenceError` and `ex : Crystal::SyntaxException`.

When the user is half way editing a file and the editor use the formatter I think we want that to succeed (even if it is unable to do some format)

MakeNowJust: `SyntaxException` and `InvalidByteSequenceError` are also error. So, it should report as error generally. But they has difference than a bug. They are expected on formatter, and a bug is unexpected.
It may be needed to distinguish both a bug (unexpected error) and other errors (expected error), so `crystal tool format` returns `1` on expected errors and `2` on unexpected errors. And, editor plugin applying formatter checks status code of `crystal tool format`, then it shows pop-up only if the status code is `2`, but ignores errors when status code is `1` because it is not useful on editor plugin.

asterite:
I didn't follow the entire conversation, but:
- Did #7144 make something break? We could just revert it, I didn't see the need to merge it.
- A formatter bug should be reported (like it is right now) but shouldn't affect exit code or halt formatting other files. Imagine I put a formatter check on travis and there's a bug, I can't get CI to pass and I have to do some ugly workaround for that.
- There's no need to show the backtrace or anything like that. In fact we don't need a backtrace when a compiler bug happens because the ones that care about the backtraces are the people who are going to fix the bug. And for that you write some code that reproduces the bug, run it, and when running it you'll see the error trace in the spec itself.

Please let's not make things more complex than they should.

straight-shoota:
@asterite Before #7144, when a formatter error happened on an invocation with a single-file argument, it would immediately print the entire error trace. With multiple arguments, it would just print that there was an error and to run the formatter with the breaking file as single argument in order to show the trace. This advice is still there, but it doesn't work anymore because #7144 removed any semantic difference between single and multiple file arguments. I don't think this demands for reverting that change, the fix is simply to use a explicit flag to enable error traces, instead of relying on implicit meaning.

> Imagine I put a formatter check on travis and there's a bug, I can't get CI to pass and I have to do some ugly workaround for that.

The formatter check (run with `--check`) *should* definitely fail CI if it encounters a bug. When the formatter breaks, it can't verify if the file is correctly formatted or not. For all we know it could be complete jibberish. There is no way this should turn green.

src/compiler/crystal/command/format.cr
straight-shoota: @bcardiff Currently, the formatter returns `1` in these cases. It doesn't seem to be an issue.

On the contrary, changing that would probably break things.

straight-shoota: I'm fine with signaling the type of error by different return codes. But I wouldn't bother unless there is a real use case for this.

asterite:
I understand. But as I said there's never a need to show the error trace (not even in compiler bugs not related to the formatter).

I disagree about a bug halting CI. How do you make CI pass then without waiting for a new compiler release?

MakeNowJust:
@asterite As a workaround, you can use `--exclude` option to exclude the casual file.

asterite:
But that's annoying, you have to modify for example travis.yml and you also can't manually format all files before committing without also remembering to exclude that file and so on. But if others think this shouldn't be like that it's fine too.

straight-shoota:
> How do you make CI pass then without waiting for a new compiler release?

Yes, that's a pretty strong drawback. But I'd always prefer that over silently ignoring bugs which would compromise the entire purpose of the formatter check.

As @MakeNowJust mentioned above, it's usually only a little change to avoid hitting the formatter bug and it should be fairly easy to detect (if nothing else, you can binary search through the file by commenting out code blocks).

The other option would be to either ignore failures or entirely disable the formatter check on CI. That's a huge cutback but it's at least explicit.

MakeNowJust:
@asterite 

> you also can't manually format all files before committing without also remembering to exclude that file and so on.

It is not true because the formatter try to format all files, but it skips that file when error occur, so we can use `crystal tool format` simply.

straight-shoota:
I someone really wants to pass the formatter check if there are bugs, this might be a use case for returning different return codes.

But still, I wouldn't implement this without any explicit demand.

After all, formatter bugs are pretty rare and hopefully become even less frequent in the future.

bcardiff:
To bypass a failure in the formatter the use can `crystal tool format || echo "Ignore formatter error"`.

Having the backtrace reported will help to diagnose early the bug probably. I don't think it would be used to actually fix the bug because it's usually an iterative, but it helps to diagnose.

src/compiler/crystal/command/format.cr
bcardiff: Ok, I guess I misunderstood the proposed changes regarding the exit status.

RX14:
I want to just revert #7144. I made a mistake merging it.

This issue is bikeshedding. Nobody cares what the semantics are, as long as their formatter works.

If the formatter doesn't fail by reporting an exit code, developers just don't notice and report these bugs. So the formatter needs to fail. What was wrong with the old behavior?

straight-shoota:
#7144 should not be reverted. It's really weird when the formatter tool behaves differently depending on how many files a glob matches.

Assuming that #7144 stays, the short bug message needs to change, because the described method doesn't work anymore.

The current state of this PR fixes that. And unless I'm missing something, this is a change we can hopefully all agree on. Essentially it changes the message from `run "crystal tool format file.cr"` to `run "crystal tool --show-backtrace file.cr"` and actives the flag. It also simplifies some code in the process.

src/compiler/crystal/command/format.cr
```diff
@@ -17,14 +17,15 @@ class Crystal::Command
     @format = "text"
     excludes = ["lib"] of String
     includes = [] of String
-    check = nil
+    check = false
```
straight-shoota: Maybe make this an instance var to avoid passing it around?

MakeNowJust: `format` is `Crystal::Command`'s method, and `Crystal::Command` is extended by any files, so instance var conflicts easily. Other command methods under `src/compiler/crystal/command` doesn't use instance var also.

straight-shoota: Then maybe make it `@formatter_check`? It would simplify the code a lot.

MakeNowJust: This pull request is for bug fix, and I have more ideas for refactoring `Crystal::Command#format` around. I'd like to work after merged this PR.

PR #7245, opened 2019-01-01
Fix: detect OpenSSL/LibreSSL from official C headers

### LINKED ISSUE #7244: OpenSSL version detection fails with OpenSSL 1.1.1 in FreeBSD base (no pkg-config file)
The `ssl_version` in `lib_ssl.cr` / `lib_crypto.cr` relies on `pkg-config` only. The OpenSSL included in the FreeBSD base system (i.e. without installing a version from packages, which is currently not newer — 1.1.1 is in base) does not include a pkg-config file. So Crystal assumes it's not 1.1:

```
ld: error: undefined symbol: SSL_library_init
>>> referenced by lib_ssl.cr:227 (/usr/home/greg/src/github.com/crystal-lang/crystal/src/openssl/lib_ssl.cr:227)
>>>               _main.o:(__crystal_main)

ld: error: undefined symbol: SSL_load_error_strings
>>> referenced by lib_ssl.cr:227 (/usr/home/greg/src/github.com/crystal-lang/crystal/src/openssl/lib_ssl.cr:227)
>>>               _main.o:(__crystal_main)

ld: error: undefined symbol: OPENSSL_add_all_algorithms_noconf
>>> referenced by lib_ssl.cr:227 (/usr/home/greg/src/github.com/crystal-lang/crystal/src/openssl/lib_ssl.cr:227)
>>>               _main.o:(__crystal_main)
…
```

Would it be a good idea to check `openssl version`? I can't think of any other way to detect the version…

ysbaddaden:
Detecting the OpenSSL version using pkg-config is prone to bugs when a libssl.pc or libcrypto.pc definition aren't available, or pkg-config itself isn't available. Using a C preprocessor to extract the official `OPENSSL_VERSION_NUMBER` and `LIBRESSL_VERSION_NUMBER` constants from the `openssl/opensslv.h` header should always be correct.

The version numbers are a little cryptic when compared to their text counterpart, but this is what's expected by OpenSSL, so we should follow.

Of course, if `pkg-config` if installed we use it to detect the correct version (linking will use it).

fixes #7244

### DIFF
--- a/src/openssl/version.sh
+++ b/src/openssl/version.sh
@@ -1,0 +1,38 @@
+#! /bin/sh
+
+LIBNAME=$1
+
+if [ -z $OPENSSL_DIR ]; then
+    # use pkg-config if available:
+    OPENSSL_CFLAGS=$(command -v pkg-config > /dev/null && pkg-config --cflags --silence-errors $LIBNAME)
+else
+    # use specified prefix:
+    OPENSSL_CFLAGS="-I$OPENSSL_DIR/include"
+fi
+
+# extract version numbers from OpenSSL/LibreSSL C headers:
+LIBRESSL_VERSION_NUMBER=$(printf "#include <openssl/opensslv.h>\nLIBRESSL_VERSION_NUMBER" | ${CC:-cc} $OPENSSL_CFLAGS -E - 2> /dev/null | tail -n 1)
+OPENSSL_VERSION_NUMBER=$(printf "#include <openssl/opensslv.h>\nOPENSSL_VERSION_NUMBER" | ${CC:-cc} $OPENSSL_CFLAGS -E - 2> /dev/null | tail -n 1)
+
+if [ $LIBRESSL_VERSION_NUMBER = LIBRESSL_VERSION_NUMBER ]; then
+    # not libressl:
+    echo LIBRESSL_VERSION_NUMBER = 0x0
+
+    if [ $OPENSSL_VERSION_NUMBER = OPENSSL_VERSION_NUMBER ]; then
+        echo '{% raise "Error: failed to find OpenSSL or LibreSSL development headers" %}'
+    else
+        # detected OpenSSL:
+        echo OPENSSL_VERSION_NUMBER = $(echo $OPENSSL_VERSION_NUMBER | sed 's/L//')
+    fi
+else
+    # detected LibreSSL:
+    echo LIBRESSL_VERSION_NUMBER = $(echo $LIBRESSL_VERSION_NUMBER | sed 's/L//')
+    echo OPENSSL_VERSION_NUMBER = 0x0
+fi
+
+if [ -z $OPENSSL_DIR ]; then
+    echo LDFLAGS = nil
+else
+    # pass specified lib folder:
+    echo LDFLAGS = \"-L$OPENSSL_DIR/lib\"
+fi
--- a/src/openssl/version.sh
+++ b/src/openssl/version.sh
@@ -10,6 +10,22 @@ else
     OPENSSL_CFLAGS="-I$OPENSSL_DIR/include"
 fi
 
+if [ -z $OPENSSL_CFLAGS ] && [ "`uname -s`" = "Darwin" ]; then
+    OPENSSL_DIR=/usr/local/opt/openssl
+
+    # check default homebrew install directory (fast):
+    if [ -d "$OPENSSL_DIR/include" ]; then
+        OPENSSL_CFLAGS="-I$OPENSSL_DIR/include"
+    else
+        # ask homebrew (slower):
+        OPENSSL_DIR=$(command -v brew > /dev/null && brew --prefix openssl)
+
+        if [ -n "$OPENSSL_DIR" ] && [ -d "$OPENSSL_DIR/include" ]; then
+            OPENSSL_CFLAGS="-I$OPENSSL_DIR/include"
+        fi
+    fi
+fi
+
 # extract version numbers from OpenSSL/LibreSSL C headers:
 LIBRESSL_VERSION_NUMBER=$(printf "#include <openssl/opensslv.h>\nLIBRESSL_VERSION_NUMBER" | ${CC:-cc} $OPENSSL_CFLAGS -E - 2> /dev/null | tail -n 1)
 OPENSSL_VERSION_NUMBER=$(printf "#include <openssl/opensslv.h>\nOPENSSL_VERSION_NUMBER" | ${CC:-cc} $OPENSSL_CFLAGS -E - 2> /dev/null | tail -n 1)
--- a/bin/ci
+++ b/bin/ci
@@ -86,7 +86,7 @@ format() {
 prepare_build() {
   on_linux verify_linux_environment
 
-  on_osx brew install crystal
+  on_osx brew install crystal openssl
 
   # Make sure binaries from llvm are available in PATH
   on_osx brew install jq
--- a/spec/std/openssl/ssl/context_spec.cr
+++ b/spec/std/openssl/ssl/context_spec.cr
@@ -53,7 +53,7 @@ describe OpenSSL::SSL::Context do
     context.should be_a(OpenSSL::SSL::Context::Client)
     context.verify_mode.should eq(OpenSSL::SSL::VerifyMode::NONE)
     context.options.no_ssl_v3?.should_not be_true
-    {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.1.1") >= 0 %}
+    {% if LibSSL::OPENSSL_VERSION_NUMBER >= 0x1_01_01_00_0 %}
     context.modes.should eq(OpenSSL::SSL::Modes::AUTO_RETRY)
     {% else %}
     context.modes.should eq(OpenSSL::SSL::Modes::None)
@@ -67,7 +67,7 @@ describe OpenSSL::SSL::Context do
     context.should be_a(OpenSSL::SSL::Context::Server)
     context.verify_mode.should eq(OpenSSL::SSL::VerifyMode::NONE)
     context.options.no_ssl_v3?.should_not be_true
-    {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.1.1") >= 0 %}
+    {% if LibSSL::OPENSSL_VERSION_NUMBER >= 0x1_01_01_00_0 %}
     context.modes.should eq(OpenSSL::SSL::Modes::AUTO_RETRY)
     {% else %}
     context.modes.should eq(OpenSSL::SSL::Modes::None)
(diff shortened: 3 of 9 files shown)


ysbaddaden:
I'm all in favor to abstract a TLS module in Crystal's stdlib, allowing to choose or switch to any library (OpenSSL, BearSSL, GnuTLS, ...) by using a selected shard.

I have some concerns with mbedtls. Looking at TLS linux distributions, there are differently named libraries: libpolarssl.so, libmbedtls.so, and some different ABI versions, which will probably be confusing. Packaged versions of PolarSSL are far outdated (1.3.9 in Debian Jessie for example). It probably requires to switch to mbedtls which may be available or not. One must enable jessie-backports to install it on Debian Jessie, and it's an old 2.4.2 release, that didn't get any update since sept 2017, despite seeing some CVE in newer releases... I'm not sure it has the kind of extended support that old OpenSSL releases can have (in RHEL, Ubuntu or Debian).

OpenSSL is ubiquitous. The API doesn't break much. I spend a couple hours once a year to fix some removed deprecations when a new version is released, and that's it. A real issue has been detecting and supporting LibreSSL, which is supposed to be an OpenSSL 1.0.0 replacement, but isn't exactly, and choosing to use text version numbers, instead of extracting the real version numbers was probably a bad choice. The last issue is that some OS are stuck with a far outdated 0.9.8 for which we should just drop support, like Ruby did 2 years ago: https://github.com/ruby/openssl/commit/4eb4b3297a92149d0a95c674107d091b18fe76af.

ysbaddaden:
Note that the `crystal` brew depends on `openssl`, which does install a supported version (1.0.2) but:
```
openssl is keg-only, which means it was not symlinked into /usr/local,
```

This means we need to set some things up to use the library we asked to install (sic).

Anyway: CI/darwin is fixed by defaulting to 0.9.8.

straight-shoota:
@RX14 Let's make that a separate thread, first of all focusing on extracting an implementation-independent TLS API.

ysbaddaden:
Alright, we have 2 choices here:

1. Default the version to 0.9.8 when no header can be found;

2. Configure CI/darwin to use the OpenSSL library we asked to install (using `pkg-config` and `PKG_CONFIG_PATH`) at least with macOS 10.12; 10.13 may work better since it ships libressl.

We can also fail compilation when headers can't be found. It's an error after all.

I believe the TLS interface and mbedtls are different topics :)

RX14:
I much prefer version 2.

We shouldn't support unmaintained openssl versions.

bcardiff:
* I would prefer to have semantic version formatted strings used all around :-(
* The brew formula only depends on pkgconfig on build, when bottle the dependency is not installed. Maybe that should change?
* I don't see where https://github.com/Homebrew/homebrew-core/blob/master/Formula/crystal.rb depends on openssl (ref: https://github.com/crystal-lang/crystal/pull/7245#issuecomment-450756001)

ysbaddaden:
> I would prefer to have semantic version formatted strings used all around

Yes, it's ugly, but that's the way to distinguish versions for OpenSSL, let's not fight it (we lose). We can't expect the pkgconfig definition to be present. OPENSSL_VERSION_TEXT is an arbitrary text and parsing it to extract the version will probably break. We could build the version text by parsing OPENSSL_VERSION_NUMBER, but is that really worth the hassle?

> The brew formula only depends on pkgconfig on build, when bottle the dependency is not installed. Maybe that should change?

I don't know. Can someone run the openssl specs from this pull request on a recent macOS release (high sierra or later)?

> I don't see where `Formula/crystal.rb` depends on openssl

The formula may not, but the log of `brew install crystal` does installs openssl on CircleCI. I don't know why. Maybe it's being updated because it was installed in the CircleCI VM?

straight-shoota:
Openssl doesn't even follow semver ;)

src/openssl/version.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/sh
+
+# use pkg-config when available:
+CFLAGS=`command -v pkg-config > /dev/null && pkg-config --cflags --silence-errors $1 || true`
```
j8r: Use `$()`, like you have done below

bcardiff:
> > I would prefer to have semantic version formatted strings used all around

> Yes, it's ugly, but that's the way to distinguish versions for OpenSSL, let's not fight it (we lose). We can't expect the pkgconfig definition to be present. OPENSSL_VERSION_TEXT is an arbitrary text and parsing it to extract the version will probably break. We could build the version text by parsing OPENSSL_VERSION_NUMBER, but is that really worth the hassle?

Oh well...

> > I don't see where Formula/crystal.rb depends on openssl

> The formula may not, but the log of brew install crystal does installs openssl on CircleCI. I don't know why. Maybe it's being updated because it was installed in the CircleCI VM?

ruby formula which is updated to 2.3.7 seems to be the one that installs openssl.

RX14:
>     * The brew formula only depends on pkgconfig on build, when bottle the dependency is not installed. Maybe that should change?

Yes, it should. Crystal's debian packages already depend on pkg-config. A lot of Crystal assumes pkg-config is available in practice.

The crystal homebrew formula should ensure that the correct `PKG_CONFIG_PATH` is set for the crystal compiler to see openssl, however it decides to do that.

RX14:
>     2\. We can check the default homebrew directory on macOS (`/usr/local/opt/openssl`) and even ask brew about it with `brew --prefix openssl`. That's simple and should just work, without requiring any configuration by the user —it's a shame that openssl is poorly managed on macOS, but that's Apple's fault, not Crystal users' fault.

No, I refuse to merge this. This hack belongs in homebrew, not Crystal. If crystal even needs to know homebrew exists, I will strongly argue against merging it.

Homebrew should wrap crystal (or `pkg-config`), and export the correct `PKG_CONFIG_DIR` so that the build environment is correct. It is not crystal's job to fix the build environment!

ysbaddaden:
I don't care about homebrew and macOS, but lots of developers do use them. Adding a few lines to a script file to help the most common scenario ain't much.

Especially since this patch will break all macOS+brew installations, because it only worked by chance until now —defaulting to a 0.9.8/1.0.0 version that happened to be the system installed one.

RX14:
> I don't care about homebrew and macOS, but lots of developers do use them. Adding a few lines to a script file to help the most common scenario ain't much.

I'm not advocating breaking anything, I'm just advocating moving the hack where it belongs.

ysbaddaden:
Then I let someone that's actually using macOS and brew to handle the fix properly in the crystal brew formula. For example:
- patching `src/openssl/version.sh` to search for `/usr/local/opt/openssl` then fallback to `brew --prefix openssl`;
- make `pkg-config` a mandatory requirement and fix `PKG_CONFIG_PATH` in the crystal wrapper shell script;
- whatever other solution that works.

Until some as an acceptable solution for both crystal and homebrew, we should wait before merging this PR, or just merge it with the brew fix. I'm fine with whatever, as long as we do fix openssl for homebrew users in the end.

RX14:
>   * make `pkg-config` a mandatory requirement and fix `PKG_CONFIG_PATH` in the crystal wrapper shell script (hopefully we won't need a fix for each and every binding in stdlib);

This was what I was essentially proposing.

Looking at homebrew: [`pkg-config` is compiled](https://github.com/Homebrew/homebrew-core/blob/193b6b8c9654d310591767a9c87ae581633b35b3/Formula/pkg-config.rb) with a default pkg-config path which contains the default homebrew locations. So I think just installing `pkg-config` should be enough?

src/openssl/version.sh
```diff
@@ -0,0 +1,54 @@
+#! /bin/sh
+
+LIBNAME=$1
+
+if [ -z $OPENSSL_DIR ]; then
```
RX14: Where is `OPENSSL_DIR` coming from?

ysbaddaden:
> pkg-config is compiled with a default pkg-config path (...) I think just editing the homebrew formula to install pkg-config should be enough.

No it won't. `openssl.pc` isn't in a standard location (like `/usr/local/lib/pkgconfig`) but in a specific installation folder for openssl (`/usr/local/opt/openssl/lib/pkgconfig`), this is on purpose.

> Where is OPENSSL_DIR coming from?

Just a simple mean to quickly select an installed version. Could be removed, but it's nice to just type `OPENSSL_DIR=/opt/openssl-1.1.1`. It's simpler and less error prone than typing `PKG_CONFIG_PATH=/opt/openssl-1.1.1/lib/pkgconfig:$PKG_CONFIG_PATH` (or wait, was it `lib/pkg-config` or `PKGCONFIG_DIR`? I always forget), especially when trying to test one specific version, among many installed versions.

pkgconfig is a nice tool when things are installed globally on the system, yet awful when they're not.

RX14:
> No it won't. `openssl.pc` isn't in a standard location (like `/usr/local/lib/pkgconfig`) but in a specific installation folder for openssl (`/usr/local/opt/openssl/lib/pkgconfig`), this is on purpose.

I don't understand why they'd do this :/ If Homebrew wants to be awkward, we need to discuss with Homebrew what the proper fix is. Hacking around it without understanding Homebrew's point of view seems ill-advised.



> Just a simple mean to quickly select an installed version. Could be removed, but it's nice to just type `OPENSSL_DIR=/opt/openssl-1.1.1`. It's simpler and less error prone than typing `PKG_CONFIG_PATH=/opt/openssl-1.1.1/lib/pkgconfig:$PKG_CONFIG_PATH` (or wait, was it `lib/pkg-config` or `PKGCONFIG_DIR`? I always forget and dig stackoverflow), especially when trying to test one specific version, among many installed versions.

Then can it become `CRYSTAL_OPENSSL_DIR`? I can imagine conflicts with that name...

RX14:
This just need to be merged without the macos-specific stuff and then use a wrapper script to fix the build environment in Homebrew. I don't see why homebrew wouldn't accept that.

InfRandomness:
Hey, sorry to dig-up such an old thread,
I've read the discussion and it looks like this got very close to being merged, I was wondering what was the status of this as of today (3 years later)
my goal being to try to push crystal on freebsd to a better state

dmgk:
Echoing @InfRandomness question above.

On FreeBSD, openssl is available both from the base system and from ports, and Crystal currently [requires](https://github.com/freebsd/freebsd-ports/blob/main/lang/crystal/files/extra-patch-src_openssl_lib__crypto.cr) [patching](https://github.com/freebsd/freebsd-ports/blob/main/lang/crystal/files/extra-patch-src_openssl_lib__ssl.cr) to use openssl from the base (https://github.com/crystal-lang/crystal/issues/7244).

It looks that HomeBrew issue was resolved by https://github.com/Homebrew/homebrew-core/pull/48028, which seemed to be the only objection to merging this PR.

straight-shoota:
I wouldn't mind polishing up this PR and merging it.

But overall, I'd rather like a more flexible, general approach for configuring lib bindings. There should be easy options to specify which libraries and versions to link to at compile time. This should technically work without having OpenSSL sources locally installed (or rather, despite), for example for cross-compiling workflows.
This problem is by far not a specific for libssl/libcrypto but applies to many other library bindings, hence I'd look for a more generic solution.

ysbaddaden:
Honestly, I don't see any alternative, especially a generic one. There is #8336 but it only works for types and "constants", so we still need checking versions to know whether a function is available or not. It would help avoid the shell script, though, which does just that (preprocess the C headers to extract the version constants).

asterite:
We can always improve #8336 or do something similar that checks whether a function is define. This can be done by creating a small C program and check if it compiles or not.

ysbaddaden:
OpenSSL vs LibreSSL strikes again: the OpenBSD 7.6 base ships LibreSSL v4.0.0 but the `pkg-config` for libssl and libcrypto reports v2.0.0 which is breaking the openssl integration.

Trying to install the openssl pkg doesn't help because there are 3 versions that can be installed in parallel in non standard directories, and each with their own exotic pkg-config name :sob:

PR #7247, opened 2019-01-02
Replace #status_code with #status

dwightwatson:
This is a follow-up to #7231 but broken out into another PR because the direction changed a bit from the discussion over there - thought it might be wise in case that direction is reversed. This basically introduces a number of breaking changes:

* `#status_code` on the response object is replaced with `#status` which is an instance of `HTTP::Status`. It encapsulates everything needed to do with HTTP status codes.
* `HTTP.default_status_message_for` was replaced by the `HTTP::Status` object's `description` and it will now return `nil` instead of an empty string when it doesn't have an answer.
* You can still set `status_code=` if you like, and it will be mapped under the hood to an instance of `HTTP::Status`. 
*`HTTP::Status` has `#code` so can call `status.code` to get the integer value instead of needing to know to call `status.value`.

Open to more guidance and feedback here. Hope it's going in the direction you'd like to see - there were a lot of comments on the last PR and I've tried to take it all into consideration.


straight-shoota:
When starting with a fresh PR, could you please squash the commits into logical pieces (probably similar to your bullet points)? That makes it much easier to review and follow the individual changes. Thanks :+1:

src/http/client/response.cr
```diff
@@ -75,8 +93,8 @@ class HTTP::Client::Response
     end
   end
 
-  def self.mandatory_body?(status_code) : Bool
-    !(status_code / 100 == 1 || status_code == 204 || status_code == 304)
+  def self.mandatory_body?(status : HTTP::Status) : Bool
+    !(status.informational? || status == HTTP::Status::NO_CONTENT || status == HTTP::Status::NOT_MODIFIED)
```
Sija:
```suggestion
    !(status.informational? || status.no_content? || status.not_modified?)
```

src/oauth2/client.cr
```diff
@@ -145,8 +145,8 @@ class OAuth2::Client
     }
 
     response = HTTP::Client.post(token_uri, form: body, headers: headers)
-    case response.status_code
-    when 200, 201
+    case response.status
+    when HTTP::Status::OK, HTTP::Status::CREATED
```
Sija:
```suggestion
    when .ok?, .created?
```

src/http/web_socket/protocol.cr
```diff
@@ -272,8 +272,8 @@ class HTTP::WebSocket::Protocol
       handshake.to_io(socket)
       socket.flush
       handshake_response = HTTP::Client::Response.from_io(socket)
-      unless handshake_response.status_code == 101
-        raise Socket::Error.new("Handshake got denied. Status code was #{handshake_response.status_code}.")
+      unless handshake_response.status == HTTP::Status::SWITCHING_PROTOCOLS
```
Sija:
```suggestion
      unless handshake_response.status.switching_protocols?
```

src/http/client/response.cr
dwightwatson: Oh cool, didn't realise you got these methods for free. Will update.

dwightwatson:
I suspect the failure on test_linux32 might be a false positive - is it possible to restart that one?

src/http/server/handlers/log_handler.cr
```diff
@@ -11,7 +11,7 @@ class HTTP::LogHandler
     elapsed = Time.measure { call_next(context) }
     elapsed_text = elapsed_text(elapsed)
 
-    @io.puts "#{context.request.method} #{context.request.resource} - #{context.response.status_code} (#{elapsed_text})"
+    @io.puts "#{context.request.method} #{context.request.resource} - #{context.response.status.code} (#{elapsed_text})"
```
straight-shoota: This line doesn't need to change.

src/http/client/response.cr
```diff
@@ -3,20 +3,33 @@ require "../common"
...
       @body = "" unless @body || @body_io
     else
       if (@body || @body_io) && (headers["Content-Length"]? != "0")
-        raise ArgumentError.new("Status #{status_code} should not have a body")
+        raise ArgumentError.new("Status #{status.code} should not have a body")
+      end
+    end
+  end
+
+  def initialize(status_code : Int32, @body : String? = nil, @headers : Headers = Headers.new, status_message = nil, @version = "HTTP/1.1", @body_io = nil)
```
straight-shoota: This should just be a `.new` override accepting Int32 and delegating to the automatically defined `.new` method of the previously declared constructor:
```cr
def self.new(status_code : Int32 body : String? = nil, headers : Headers = Headers.new, status_message = nil, version = "HTTP/1.1", body_io = nil)
  new(HTTP::Status.new(status_code), body, headers, status_message, version, body_io)
end
```
There is no need to duplicate the entire logic.

straight-shoota:
This looks good!

There is however still an issue with creating a new `HTTP::Status` instance from an arbitrary status code number. `.new` on enums is implicitly defined and technically not part of the public API. As far as I can tell, it can neither be overriden nor documented.
Thus, we'd probably need to use a method with a different name that can be documented as part of the public API and validate the assigned value.
Probably something like this:
```cr
# Create a new `HTTP::Status` instance from *status_code* number.
# The values are not limited to defined enum members, but also allow to use custom status codes. Any number in the range 100..999 is allowed.
def HTTP::Status.from_code(status_code : Int32)
  raise "Invalid status code" unless 100 <= status_code <= 999
  new(status_code)
end

File: spec/std/http/status_spec.cr [hunk 6]
```diff
@@ -0,0 +1,82 @@
+require "spec"
+require "http"
+
+describe HTTP::Status do
+  describe "#from_code" do
```
straight-shoota: This should be called `.from_code` (class methods are prefixed by `.`, instance methods by `#`).

Could you also add a passing spec for a valid but undefined status code?

src/http/client/response.cr
```diff
@@ -131,14 +140,15 @@ class HTTP::Client::Response
       raise "Invalid HTTP status code: #{pieces[1]}"
     end
 
+    status = HTTP::Status.from_code(status_code)
     status_message = pieces[2]? ? pieces[2].chomp : ""
 
     body_type = HTTP::BodyType::OnDemand
-    body_type = HTTP::BodyType::Mandatory if mandatory_body?(status_code)
+    body_type = HTTP::BodyType::Mandatory if mandatory_body?(status)
```
asterite: Maybe we could move this method to a `body_mandatory?` in the `HTTP::Status` enum (it could be done in a separate PR).

src/http/server/handler.cr
```diff
@@ -23,7 +23,7 @@ module HTTP::Handler
     if next_handler = @next
       next_handler.call(context)
     else
-      context.response.status_code = 404
+      context.response.status = HTTP::Status::NOT_FOUND
```
asterite: Optional: just use `:not_found` instead of the full enum name. Same in other places where the status is assigned.

src/http/server/response.cr
```diff
@@ -113,17 +123,18 @@ class HTTP::Server
     # Generates an error response using *message* and *code*.
     #
     # Calls `reset` and then writes the given message.
-    def respond_with_error(message = "Internal Server Error", code = 500)
+    def respond_with_error(status : HTTP::Status = HTTP::Status::INTERNAL_SERVER_ERROR, status_message = nil)
```
asterite: Note that this is a breaking change. We should either postpone merging this PR until 0.28.0 or else preserve the old overload.

src/http/status.cr
```diff
@@ -0,0 +1,172 @@
+# A support enum that provides additional around HTTP status codes.
```
asterite: additional what?

src/http/status.cr
```diff
@@ -0,0 +1,172 @@
...
+  SERVICE_UNAVAILABLE             = 503
+  GATEWAY_TIMEOUT                 = 504
+  HTTP_VERSION_NOT_SUPPORTED      = 505
+  VARIANT_ALSO_NEGOTIATES         = 506
+  INSUFFICIENT_STORAGE            = 507
+  LOOP_DETECTED                   = 508
+  NOT_EXTENDED                    = 510
+  NETWORK_AUTHENTICATION_REQUIRED = 511
+
+  def self.from_code(status_code : Int32)
```
asterite: Please document the difference between this and `new`, mainly that this will raise an `ArgumentError` if the status code is not inside (100..999) (also change `raise` to effectively raise `ArgumentError`).

src/http/status.cr
```diff
@@ -0,0 +1,172 @@
...
+    when 502 then "Bad Gateway"
+    when 503 then "Service Unavailable"
+    when 504 then "Gateway Timeout"
+    when 505 then "HTTP Version Not Supported"
+    when 506 then "Variant Also Negotiates"
+    when 507 then "Insufficient Storage"
+    when 508 then "Loop Detected"
+    when 510 then "Not Extended"
+    when 511 then "Network Authentication Required"
+    else          nil
```
asterite: I know this has been discussed, but I would change this to an empty string. Imagine having to pass the description to some place that expects a string, then you won't be able to use this method and you'll have to use `not_nil!` (which is probably incorrect, because you  don't want it to fail like that) or use `|| ""` which is what this method should return.

asterite: Basically, if we don't know the description because the code isn't officially registered we shouldn't signal that with something noteworthy as `nil`, which requires additional handling: an empty string is fine and easier to use.

bew: or `"Unknown http code"` ?

src/http/client/response.cr
dwightwatson: Happy to tackle that one separately.

src/http/status.cr
```diff
@@ -0,0 +1,172 @@
...
+  GATEWAY_TIMEOUT                 = 504
+  HTTP_VERSION_NOT_SUPPORTED      = 505
+  VARIANT_ALSO_NEGOTIATES         = 506
+  INSUFFICIENT_STORAGE            = 507
+  LOOP_DETECTED                   = 508
+  NOT_EXTENDED                    = 510
+  NETWORK_AUTHENTICATION_REQUIRED = 511
+
+  def self.from_code(status_code : Int32)
+    raise "Invalid HTTP status code: #{status_code}" unless 100 <= status_code <= 999
```
straight-shoota: The upper limit should probably be only 599. According to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6) there are five categories of status codes, indicated by the first digit.

src/http/status.cr
straight-shoota: @bew The status message is supposed to explain the meaning of the code to human readers. Stating that it's unknown, doesn't explain anything about the status code, only about the processing handler. Unknown codes should either return `nil` or empty string which expresses that there is no explanation.

src/http/server/handler.cr
ysbaddaden: Same in specs, for example `Response.new(:ok)`.

src/http/status.cr
j8r: I agree with @asterite about `nil` - if we can avoid an Union and a nil, that's all good :)

bew: ok, thank you @straight-shoota

src/http/status.cr
jwoertink: Yeah, and linkedin actually returns a [server 999](https://stackoverflow.com/questions/27571419/how-to-avoid-http-1-1-999-request-denied-response-from-linkedin) when scraping their site.

src/http/client/response.cr
```diff
@@ -131,14 +140,15 @@ class HTTP::Client::Response
       raise "Invalid HTTP status code: #{pieces[1]}"
     end
 
+    status = HTTP::Status.from_code(status_code)
```
asterite: This is wrong. We shouldn't raise if we receive an invalid http status code in the response. Please use new here.

straight-shoota: That depends on what is considered invalid, see thread above. It should definitely raise if the status code is not a three-digit number.

We might consider allowing to receive `600..999` and maybe `000`..`099` yet let these ranges fail for `from_code`. But I think that's to complicated having different restrictions.

It's probably best if `000..999` is allowed everywhere.

src/http/status.cr
Sija: One could always do `description.to_s` which would return an empty string on `nil`.

asterite: Good point. Lets leave nil then.

src/http/client/response.cr
```diff
@@ -61,7 +70,7 @@ class HTTP::Client::Response
   end
 
   def to_io(io)
-    io << @version << ' ' << @status_code << ' ' << @status_message << "\r\n"
+    io << @version << ' ' << @status.value << ' ' << @status_message << "\r\n"
```
Sija:
```suggestion
    io << @version << ' ' << @status.code << ' ' << @status_message << "\r\n"
```

src/http/status.cr
```diff
@@ -0,0 +1,174 @@
...
+    new(status_code)
+  end
+
+  def code
+    value
+  end
+
+  # Returns `true` if the response status code is between 100 and 199.
+  def informational?
+    100 <= value <= 199
```
Sija:
```suggestion
    100 <= code <= 199
```

imo reads better. ditto all below.

src/http/client/response.cr
```diff
@@ -48,6 +52,11 @@ class HTTP::Client::Response
     process_content_type_header.content_type
   end
 
+  # Convenience method to retrieve the HTTP status code.
```
Sija: Should this be marked as `DEPRECATED`?

src/http/status.cr
straight-shoota: Golang implementation makes it `"status code #{code}"` which is not very helpful IMO.

src/http/client/response.cr
straight-shoota: Oh, but values < 100 need extra care when printing. Would be printf `%03d` for inserting leading zeros to make it three digits always.

src/http/status.cr
```diff
@@ -0,0 +1,174 @@
...
+  GATEWAY_TIMEOUT                 = 504
+  HTTP_VERSION_NOT_SUPPORTED      = 505
+  VARIANT_ALSO_NEGOTIATES         = 506
+  INSUFFICIENT_STORAGE            = 507
+  LOOP_DETECTED                   = 508
+  NOT_EXTENDED                    = 510
+  NETWORK_AUTHENTICATION_REQUIRED = 511
+
+  # Create a new status instance with the given status code, or raise an
+  # error if the status code given is not inside 000..999.
```
Sija: Outdated range in the description.

src/http/status.cr
```diff
@@ -0,0 +1,174 @@
...
+
+  # Create a new status instance with the given status code, or raise an
+  # error if the status code given is not inside 100..999.
+  def self.from_code(status_code : Int32)
+    raise ArgumentError.new("Invalid HTTP status code: #{status_code}") unless 100 <= status_code <= 999
+    new(status_code)
+  end
+
+  def code
+    value
```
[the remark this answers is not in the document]
dwightwatson: I added it just for convenience because it read a little better, and it doesn't require the person using it to know it's an enum. Originally I didn't use it anywhere else - I left it just as a helpful addition, but another comment on the PR asked that I use `#code` everywhere.

spec/std/http/client/response_spec.cr
```diff
@@ -169,14 +169,14 @@ class HTTP::Client
...
-      [100, 101, 204, 304].each do |status|
+      [HTTP::Status::CONTINUE, HTTP::Status::SWITCHING_PROTOCOLS, HTTP::Status::NO_CONTENT, HTTP::Status::NOT_MODIFIED].each do |status|
         response = Response.new(status)
         response.headers.size.should eq(0)
       end
     end
 
     it "raises when creating 1xx, 204 or 304 with body" do
-      [100, 101, 204, 304].each do |status|
+      [HTTP::Status::CONTINUE, HTTP::Status::SWITCHING_PROTOCOLS, HTTP::Status::NO_CONTENT, HTTP::Status::NOT_MODIFIED].each do |status|
```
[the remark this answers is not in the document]
dwightwatson: Could you please explain this further? I'm still new to Crystal and I'm not sure how I would go about using tuples here.

src/http/status.cr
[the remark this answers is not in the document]
straight-shoota: It's an alias but only because we can't remove `value`.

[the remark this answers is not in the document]
j8r: Therefore:
- internally, `value` can be used because we know that's the enum,'s value and this avoid performance penalties (despite being minor) 
- `#code` has to be documented with comments

spec/std/http/client/response_spec.cr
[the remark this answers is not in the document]
j8r: @dwightwatson https://crystal-lang.org/api/master/Tuple.html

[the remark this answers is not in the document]
j8r: TL;DR: they are fixed size and allocated on stack - this may help a bit the specs to be lighter on execution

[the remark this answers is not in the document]
dwightwatson: Oh, so you mean to literally just replace the array with a tuple syntax?

```crystal
{HTTP::Status::CONTINUE, HTTP::Status::SWITCHING_PROTOCOLS, HTTP::Status::NO_CONTENT, HTTP::Status::NOT_MODIFIED}.each do |status|
```

I'm open to it if others agree but I feel that sacrifices readability for little gain.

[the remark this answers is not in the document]
asterite: It matters for performance reasons, but given that this is a test it doesn't matter to (it's not like this is in the middle of a loop or something). You can leave it as is.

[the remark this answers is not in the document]
j8r: Yeah it doesn't matter much, but still we sometimes refer to the test for code samples - it doesn't hurt. Furthermore, you have learnt a new thing today :smile:

src/http/status.cr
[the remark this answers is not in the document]
j8r: Sure really minor, doesn't it calls `#code` that calls `#value`?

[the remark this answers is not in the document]
j8r: Just asking what the gain of using a new method internally, we clearly see we are inside an Enum doesn't it?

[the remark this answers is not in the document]
r00ster91: No, there's no perfomance penalty at all. There would eventually be one when you use Ruby but Crystal is *compiled* and efficient. As far as I know for example this:
```cr
def method1
  123
end
def method2
  method1
end
method2()
```
simply becomes
```cr
123
```
at compile-time.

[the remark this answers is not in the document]
j8r: oh ok, same for `@ivar` and `ivar` I guess?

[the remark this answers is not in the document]
asterite: Yes, the compiler (LLVM, but in some cases Crystal) inlines stuff. **A lot**.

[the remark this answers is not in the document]
j8r: Thanks for the explanations.
Another good point of having this method is for documentation.

dwightwatson:
There's been quite a bit of discussion on this PR so I'm not sure exactly which changes are still being requested. Keen to finish off whatever remains to get this one through.

straight-shoota:
This looks good. I don't think there is any immediate action required, all requests have been met. Just waiting for reviews.

src/http/server/handler.cr
Sija: This is still unresolved, just FYI.

j8r:
Why using symbols? The purpose of enums are compile-time safety – despite being more verbose. Symbols can be any name without an error at compilation.

dwightwatson:
As per this comment: https://github.com/crystal-lang/crystal/pull/7247#discussion_r244744646

From my understanding the symbols map to the values of the enum. If you try to set a symbol that doesn't reflect a possible value the compiler errors.

```crystal
enum HTTP::Status
	OK = 200
end

class Response
  def initialize(@status : HTTP::Status)
  end
end


Response.new(:ok) # no issue
Response.new(:not_ok) # compiler error
```

j8r:
That's strange in this cas you're right. You pass a symbol but the error in your example is:
```cr
Error in line 9: no overload matches 'Response.new' with type Symbol
Overloads are: - Response.new(status : HTTP::Status)
```
Why it works in line 8?? That what confused me in the first place.
I thought we would like to avoid Symbols for incremental compilation etc...

j8r:
On my mind, a Symbol overload was needed. This Symbol => Enum feature is weird, it creates a "ghost overload "didn't see it in docs (See `Usage` in https://crystal-lang.org/docs/syntax_and_semantics/enum.html#Usage)

```cr
enum HTTP::Status
  OK = 200
end
class Response
  def initialize(status : Symbol)
    case status
    when :ok then HTTPS::Status::OK
    else raise "unknown"
    end
  end
end
Response.new(:ok)
```

asterite:
@j8r the automatic casting from symbol to enum works as long as there's an enum restriction. It was implemented some time ago but never documented.

spec/std/http/client/response_spec.cr
```diff
@@ -169,14 +169,14 @@ class HTTP::Client
     end
 
     it "doesn't sets content length for 1xx, 204 or 304" do
-      [100, 101, 204, 304].each do |status|
+      [HTTP::Status::CONTINUE, HTTP::Status::SWITCHING_PROTOCOLS, HTTP::Status::NO_CONTENT, HTTP::Status::NOT_MODIFIED].each do |status|
```
RX14:
```suggestion
      response_codes = [:Continue, :SwitchingProtocols, :NoContent, :NotModified] of HTTP::Status
      response_codes.each do |status|
```

spec/std/http/client/response_spec.cr
RX14: ditto

src/http/server/response.cr
```diff
@@ -115,15 +125,16 @@ class HTTP::Server
     # Calls `reset` and then writes the given message.
     def respond_with_error(message = "Internal Server Error", code = 500)
       reset
-      @status_code = code
+      @status = HTTP::Status.from_code(code)
+      message = message || @status.description
```
RX14:
```suggestion
      message ||= @status.description
```

src/http/server/response.cr
```diff
@@ -115,15 +125,16 @@ class HTTP::Server
     # Calls `reset` and then writes the given message.
     def respond_with_error(message = "Internal Server Error", code = 500)
```
RX14: There needs to be another overload of this which takes in a `HTTP::Status` as `code`, but that can be in another PR.

src/http/server/response.cr
```diff
@@ -115,15 +125,16 @@ class HTTP::Server
...
-      self << code << ' ' << message << '\n'
+      self << @status.code << ' ' << message << '\n'
       flush
     end
 
     protected def write_headers
-      status_message = HTTP.default_status_message_for(@status_code)
-      @io << @version << ' ' << @status_code << ' ' << status_message << "\r\n"
+      status_message = @status.description
+      @io << @version << ' ' << @status.code << ' ' << status_message << "\r\n"
```
RX14: I'd inline the `status_message` variable here.

src/http/status.cr
```diff
@@ -0,0 +1,174 @@
+# An enum that provides additional support around HTTP status codes.
+#
+# Based on [Hypertext Transfer Protocol (HTTP) Status Code Registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)
+#
+# It provides constants for the defined HTTP status codes as well as helper
+# methods to easily identify the type of response.
+enum HTTP::Status
+  CONTINUE                        = 100
```
RX14: I'd prefer if there were in `PascalCase`.

straight-shoota: I don't oppose standardization here, but I don't think there is an established rule for either case.

j8r: You mean CamelCase? I find also this case closer to the actual status code message, they are never in all caps.

spec/std/http/client/response_spec.cr
Sija: CamelCased symbols look odd, they're almost explicitly used with snake_case.

src/http/status.cr
Sija: @RX14 that's a bikeshedding, it can be changed in another PR if needed along with other enums like `Flate::Strategy`

spec/std/http/client/response_spec.cr
RX14: This isn't about the camel cased symbols, it's about using symbols in the first place.

Sija: so let's use snake_case then.

straight-shoota:
In light of #7266 suggest to rename `Status.code` to `Status.new` because it was only a workaround in the first place and is no longer needed.
The compiler change is non-breaking and goes into 0.27.1, so this feature can safely be merged to 0.28.0 (where this PR needs to go anyway, because it's breaking the API).

NOTE: Implementing this needs a very recent compiler build from master.

src/http/server/response.cr
dwightwatson: Yeah this was discussed a little earlier - will handle this separately.

dwightwatson:
@straight-shoota Do you mean replace `.from_code` with `.new`?

dwightwatson:
Seems like the tests fail now (but work fine locally). Do the tests not run with a fresh build of Crystal?

spec/std/http/client/response_spec.cr
Sija: I don't see how that's related - here we're talking about symbols and the issue you mentioned is about enum members - 2 different things.

bcardiff:
@dwightwatson I notice you are already merging from master. Once you make the CI happy we need a 👍 from @asterite / @RX14 before **saquash merging** this PR. Thanks!

Sija:
Needs a rebase to make CI (with Crystal 0.27.1) green.

dwightwatson:
Back to green now, please let me know if there is anything more I can do to get this merged in.

Sija:
Btw, is this GTG already?

asterite:
@bcardiff `HTTP::Response#status_code` is renamed to `HTTP::Response#status`. This is a breaking change.

asterite:
@bcardiff It also changes it from a number to an enum (once you rename `status_code` to `status` you'll need to do some more changes)

bcardiff:
@asterite but `#status_code` is preserved as a getter and the initializer with and Int is available also.

z64:
@bcardiff It currently removes the public `HTTP.default_status_message_for(int)` method, replaced by `HTTP::Status.new(int).description`

asterite:
@bcardiff You are right, sorry for the confusion

straight-shoota:
Thank you @dwightwatson ❤️

PR #7251, opened 2019-01-02
Improve HTTP::Server docs

straight-shoota:
This is a general overhaul of the `HTTP::Server` API docs, describing the updated behaviour after #5776.
It adds some more details and structures the text.

I've left out the section about reusing connections which will be added with the changed behaviour in #7055.

### DIFF
--- a/src/http/server.cr
+++ b/src/http/server.cr
@@ -8,60 +8,88 @@ require "./common"
   require "openssl"
 {% end %}
 
-# An HTTP server.
+# A concurrent HTTP server implementation.
 #
-# A server is given a handler that receives an `HTTP::Server::Context` that holds
-# the `HTTP::Request` to process and must in turn configure and write to an `HTTP::Server::Response`.
+# A server is initialized with a handler chain responsible for processing each
+# incoming request.
 #
-# The `HTTP::Server::Response` object has `status` and `headers` properties that can be
-# configured before writing the response body. Once response output is written,
-# changing the `status` and `headers` properties has no effect.
+# ```
+# require "http/server"
 #
-# The `HTTP::Server::Response` is also a write-only `IO`, so all `IO` methods are available
-# in it.
+# server = HTTP::Server.new do |context|
+#   context.response.content_type = "text/plain"
+#   context.response.print "Hello world!"
+# end
+#
+# address = server.bind_tcp 8080
+# puts "Listening on http://#{address}"
+# server.listen
+# ```
+#
+# ## Binding to sockets
+#
+# The server can be bound to one ore more server sockets (see )
+#
+# Supported types:
+#
+# * TCP socket: `#bind_tcp`, `#bind_unused_port`
+# * TCP socket with TLS/SSL: `#bind_tls`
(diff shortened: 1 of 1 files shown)


src/http/server.cr
```diff
@@ -54,14 +51,45 @@ require "./common"
 #   context.response.print "Hello world!"
 # end
 #
-# server.bind_tcp "0.0.0.0", 8080
-# puts "Listening on http://0.0.0.0:8080"
+# address = server.bind_tcp "0.0.0.0", 8080
+# puts "Listening on http:/#{address}"
```
bew: typo: missing a `/`

src/http/server.cr
```diff
@@ -39,12 +21,27 @@ require "./common"
...
-# puts "Listening on http://127.0.0.1:8080"
+# address = server.bind_tcp 8080
+# puts "Listening on http://#{address}"
 # server.listen
 # ```
 #
-# ### With non-localhost bind address
+# ## Binding to sockets
+#
+# The server can be bound to one ore more server sockets (see )
```
r00ster91:
```suggestion
# The server can be bound to one ore more server sockets (see `#bind`)
```

src/http/server.cr
```diff
@@ -39,12 +21,27 @@ require "./common"
...
+#
+# Supported types:
+#
+# * TCP socket: `#bind_tcp`, `#bind_unused_port`
+# * TCP socket with TLS/SSL: `#bind_tls`
+# * Unix socket `#bind_unix`
+#
+# `#bind(uri : URI)` and `bind(uri : String)` parse socket configuration for
```
r00ster91:
```suggestion
# `#bind(uri : URI)` and `#bind(uri : String)` parse socket configuration for
```

straight-shoota:
Thanks @r00ster91 for prove reading! ❤️

src/http/server.cr
```diff
@@ -54,14 +51,45 @@ require "./common"
...
+# ### Handler chain
+#
+# The handler given to a server can simply be a block that receives an `HTTP::Server::Context`,
+# or it can be an instance of `HTTP::Handler`. An `HTTP::Handler` has an optional
+# `#next` method to forward processing to the next handler in the chain.
+#
+# For example, an initial handler might handle exceptions raised from subsequent
+# handlers and return a `500 Server Error` status (see `HTTP::ErrorHandler`).
+# The next handler might log all incoming requests (see `HTTP::LogHandler`).
+# And the final handler deals with routing and application logic.
```
bcardiff: We are dropping how the optional block interact with the rest of the handlers.

straight-shoota: I don't follow. Are you referencing an intended change to the behaviour?

RX14:
I think I'd prefer the ordering to talk about how the block form works first, and the role of `HTTP::Context` and then digress into `HTTP::Handler`, followed by talking about binding. I think this order reflects the order which a newbie would want to explore the server: first how to integrate your logic, since that's the interesting part. Then only after how to actually use the interface.

PR #4499, opened 2017-06-03
Compiler: use Colorize.on_tty_only!

MakeNowJust:
`Colorize.on_tty_only!` is introduced in #4271, but it is not used in compiler. I think it is useful that colorize is turned on/off depending on tty detection result.

In addition, I fixed `bin/crystal` wrapper script to detect TTY. `bin/crystal` script's `SUPPORTS_COLORS` is really buggy, at least it isn't working correct now.

PR #7252, opened 2019-01-02
Refactor coloring in compiler

r00ster91:
This:
* Resolves unnecessary colorizing complexity in the compiler.
  Originally I just wanted to fix that when you do `''`, that the did-you-mean in the error is colored in bold yellow (currently it's not) but it turned out to be actually quite difficult to retrieve the `@color` value from the `Compiler` (whether `--no-color` has been passed or not) from [this spot in the lexer](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/syntax/lexer.cr#L734).
  So now at the beginning of a compilation, `Colorize.enabled` is simply used to turn colors off when `--no-color` has been passed. Which is much simpler than having to define a `colorize` method first and then having to figure out how to even retrieve the `color` value from the compiler. Now strings can be colored wherever you want just like you would in your own code without `colorize` helper methods.
* Removes 2 unnecessary `"colorize"` requires.
* Fixes some did-you-means or other helpful hints not being colored in bold yellow in multiple places.
* For example `Time.new(seconds: 0)` results in the error
  error
  with the last item not being colored in bold/white (or more specifically with the items after the `(trying this one)` not being colored in bold/white). That's fixed as well.


src/compiler/crystal/util.cr
```diff
@@ -4,7 +4,7 @@ module Crystal
     if filename.starts_with?(dir)
       filename = filename[dir.size..-1]
       if filename.starts_with? '/'
-        filename[1..-1]
+        filename.lchop
```
Sija:
```suggestion
        filename.lchop('/')
```

ditto below

r00ster91: Actually the more unspecific `lchop` without a char is faster:
```
lchop(char)  30.81M ( 32.46ns) (±11.64%)  32 B/op   1.21× slower
      lchop  37.38M ( 26.75ns) (±15.55%)  32 B/op        fastest
```
Also, I don't think there's really a readability difference. If the filename starts with `/`, remove that from the beginning.

straight-shoota: With `lchop('/')` you can spare the entire if clause. That should result in pretty much the same instructions and more concise code.

r00ster91: Ooh yes overlooked that the `if` can just be removed then. Will change that. But for the `lchop` below I don't even think I'm operating on a filename so `lchop('/')` there would probably be wrong. Will leave that one as-is.

r00ster91:
Phew CI finally passes! That took way too long.

src/compiler/crystal/semantic/call_error.cr
```diff
@@ -394,10 +394,12 @@ class Crystal::Call
       str << "\n - "
       append_def_full_name a_def.owner, a_def, arg_types, str
       if defs.size > 1 && a_def.same?(matched_def)
-        str << colorize(" (trying this one)").blue
+        str << " (trying this one)".colorize.blue
+        str << "\e[1m" # Make the rest of the string bold
```
Sija: This seems hackish... isn't there any other way to avoid using raw ansi escape codes?

ditto below

r00ster91: Yes, I know. I would really like to avoid this but I couldn't find a way so far.
Let's cut the problem down to this:
```cr
require "colorize"

str = IO::Memory.new

str << "\n - "
str << "item 1"
str << "\n - "
str << "item 2"

str << " (did you mean this one?)".colorize.yellow

str << "\n - "
str << "item 3"

puts str.to_s.colorize.bold
```

did you mean

Internally the string is:
```cr
"\e[1m\n - item 1\n - item 2\e[33m (did you mean this one?)\e[0m\n - item 3\e[0m"
 ^                          ^                              ^               ^
 Bold                       Yellow                         Terminates      Terminates
                                                           Yellow          Bold
```
The problem here is that the `Terminates Yellow` escape sequence doesn't just terminate `Yellow` but also `Bold` while actually the *last* sequence (`Terminates Bold`) is supposed to terminate `Bold`.
And that's fixed by appending a new `Bold` (`"\e[1m"`) after `Terminates Yellow`.

did you mean

straight-shoota: You just need to call `colorize.bold` on each individual component instead of the entire list.

straight-shoota: Might seem a bit irrational as it could normally be summarized. But that's the way these colorize  commands work. You can't nest them.

src/compiler/crystal/semantic/call_error.cr
```diff
@@ -391,13 +391,11 @@ class Crystal::Call
 
   def append_matches(defs, arg_types, str, *, matched_def = nil, argument_name = nil)
     defs.each do |a_def|
-      str << "\n - "
-      append_def_full_name a_def.owner, a_def, arg_types, str
+      str << "\n - #{def_full_name(a_def.owner, a_def, arg_types)}".colorize.bold
```
straight-shoota: With this change, I suppose the `append_def_full_name` method(s) can be removed in favour of `def_full_name`? They're no longer used anywhere. At the very least the instance method.

src/compiler/crystal/util.cr
```diff
@@ -17,26 +13,22 @@ module Crystal
...
     source.lines.map_with_index do |line, i|
       str = "#{"%4d" % (i + 1)}. #{line.to_s.chomp}"
       target = i + 1 == highlight_line_number
       if target
-        if color
-          str = ">".colorize.green.bold.to_s + str[1..-1].colorize.bold.to_s
-        else
-          str = ">" + str[1..-1]
-        end
+        str = ">".colorize.green.bold.to_s + str.lchop.colorize.bold.to_s
```
straight-shoota: Instead of concatenating the colorized strings only to collect them in an array and finally join them all, they could just be  appended in a `String.build` directly.
Not necessarily in scope of this PR though.

src/compiler/crystal/semantic/call_error.cr
r00ster91: I didn't remove them because `def_full_name` actually depends on `append_def_full_name` because it's used to build the string returned by `def_full_name`.
But I could of course put `append_def_full_name`'s code directly into the `String.build { |io| }` of `def_full_name` but I'm not sure if that'd be a good readability improvement.

straight-shoota: I don't know why these are defined as both instance and class methods anyway. Clearly, they're not bound to the state of the object, but they're not used from outside the class either. Maybe it's left over. I guess this could all need a refactoring, but not in this PR.

src/compiler/crystal/util.cr
```diff
@@ -17,28 +13,28 @@ module Crystal
...
-    source.lines.map_with_index do |line, i|
-      str = "#{"%4d" % (i + 1)}. #{line.to_s.chomp}"
-      target = i + 1 == highlight_line_number
-      if target
-        if color
-          str = ">".colorize.green.bold.to_s + str[1..-1].colorize.bold.to_s
+  def self.with_line_numbers(source : String, highlight_line_number = nil)
+    String.build do |io|
+      i = 0
+      source.each_line do |line|
```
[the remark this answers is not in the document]
r00ster91: So `source.lines.each_with_index(1)`? That'd be slower since it would first iterate through `source`'s lines and put it in array and then iterate through it again with an index while `source.each_line do |line|` only iterates through the lines.

src/compiler/crystal/util.cr
```diff
@@ -17,28 +13,28 @@ module Crystal
...
+        i += 1
+        if i == highlight_line_number
+          line = ">".colorize.green.bold.to_s + "#{"%3d" % i}. #{line}".colorize.bold.to_s
```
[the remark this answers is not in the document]
r00ster91: Oh, yes I get it now.

src/compiler/crystal/util.cr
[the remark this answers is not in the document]
straight-shoota: Oh, never mind. I'll add a proposal in a separate PR.

src/compiler/crystal/semantic/call_error.cr
r00ster91: I figured out now why the specs are failing.
Currently I'm colorizing each item invidually bold but the whole error message is also getting colorized bold at the end nearly before it's getting raised. Thus the error ends in `\e[0m\e[0m`. So I could fix the specs by manually appending `\e[0m` at the end. Or by colorizing everything here in this string build invidually too: [`call_error.cr:553`](https://github.com/crystal-lang/crystal/blob/36de8d89131aad4bda97a83221898fa07f770d1a/src/compiler/crystal/semantic/call_error.cr#L553) and then disabling that the whole string gets colorized bold by adding an ugly parameter here: [`exception.cr:140`](https://github.com/crystal-lang/crystal/blob/9c42d31daf1331e31ba2c07679571acea699fa7a/src/compiler/crystal/semantic/exception.cr#L140) just for this one case. But I don't think that's the right way.
But I can't really think of anything else which does not involve using ANSI escape sequences. Maybe modifying `Colorize` but I also don't think that's the right way because the user just expects `\e[0m\e[0m` to be at the end in this case.

j8r: @r00ster91 if you want to reset the colors, you can use `Colorize.reset` instead of directly the raw ANSI escape code.

r00ster91: But `Colorize.reset` only works for `IO`'s, not strings.

src/colorize.cr
```diff
@@ -111,9 +111,15 @@ module Colorize
     self.enabled = STDOUT.tty? && STDERR.tty?
   end
 
-  def self.reset(io = STDOUT)
+  # Resets the color and mode of *io* if `enabled` is `true`.
+  def self.reset(io : IO = STDOUT)
     io << "\e[0m" if enabled?
```
j8r: You can use `Colorize.reset(io) if enabled?`

src/colorize.cr
```diff
@@ -111,9 +111,15 @@ module Colorize
...
   end
+
+  # Resets the color and mode of *string* if `enabled` is `true`.
+  def self.reset(string : String)
+    enabled? ? string + "\e[0m" : string
```
j8r: Same here, but you have to use a `String.build`

src/colorize.cr
r00ster91: But this is the definition of `Colorize.reset(IO)`. Why should I use it inside of it? That would cause a stack overflow.

r00ster91:
#4499 looks much better than this, I'll close this and perhaps do an extra PR someday with some tiny things of this that #4499 doesn't have, after #4499 is merged.

PR #7254, opened 2019-01-02
Refactor be_a spec to a method

straight-shoota:
This expectation helper doesn't need to be a macro.

/cc https://github.com/crystal-lang/crystal/pull/7226#discussion_r244845729


straight-shoota:
We'll definitely need a spec for this, either way.

PR #7255, opened 2019-01-03
Compiler: let a metaclass' metaclass be a new type instead of Class

asterite:
This PR changes the language a bit regarding metaclasses. An explanation follows.

Every expression has a value and a type. For example:

- `1` has a value of `1` and a type of `Int32`. Methods for `1` are looked up in the `Int32` type. These are denoted as `Int32#method`.
- `Int32` has a value of `Int32` and a type of `Int32.class`. In the compiler this is called a _metaclass_. Methods for `Int32` are looked up in the `Int32.class` type. These are denoted as `Int32.method`.

When we have a restriction like:

```crystal
def capture(x : T) forall T
  T
end

capture(1)
```

what happens is that `1` is typed (as usual) and we get the `Int32` type. Then when passed to the method this is matched against `T` and `T` becomes `Int32`.

When we have this (note it's `T.class`):

```crystal
def capture(x : T.class) forall T
  T
end

capture(Int32)
```

what happens is that `Int32` is typed and it gets the `Int32.class` type. Then matching against `T.class` we get `Int32`. Internally to implement this we go from the metaclass to the instance type (every metaclass has an instance type, in the case of `Int32.class` it's `Int32` and so on). 

So far, so good.

The problem comes when we want to pass `Int32.class` and capture that type. The current use case is using it in the `be_a` matcher in spec, but I stumbled upon this use case a few other types (but I can't remember now what they were):

```crystal
def capture(x : T.class) forall T
  T
end

capture(Int32.class)
```

Here `Int32.class` is typed, and currently this gets a type of `Class`, because the type of a metaclass is always `Class`. And from there the compiler would go to the instance type, but the instance type of `Class` is `Object`, because a class is an object. (Note: in this point I'm a bit confused, but the important thing to know is that we go from `Int32.class` to `Class` and we lost the information of from which class we came from, to go back to it when matching against `T.class`).

This PR changes the above so that the type of `Int32.class` is a new metaclass called `Int32.class.class`, whose instance type is `Int32.class` (and this is repeated indefinitely, so you have `Int32.class.class.class`, they are created on the fly). The good thing about this is that it can go back to `Int32.class` via its instance type.

And so, the above matching works as expected, and we can write `be_a` as a method (an in general many macros that are based on types can be now always written as methods, which is always good).

I'm not sure which metamodel representation is better, but it might not matter much: one usually doesn't dabble at that level, and as long as it works as expected it's probably fine.

I also thought of an alternative: having the type of `Int32.class` be `Class`, but one whose instance type point to `Int32.class`. In that way we would have many `Class` types on the implementation side, but from the outside they would look like a single one. I think this is a bit harder to implement and at this point it might not be worth it, so I went with this simpler approach.

### DIFF
--- a/spec/compiler/codegen/is_a_spec.cr
+++ b/spec/compiler/codegen/is_a_spec.cr
@@ -715,16 +715,4 @@ describe "Codegen: is_a?" do
       1.to_s.is_a?(B)
       )).to_b.should be_true
   end
-
-  it "says true for Class.is_a?(Class.class) (#4374)" do
-    run("
-      Class.is_a?(Class.class)
-    ").to_b.should be_true
-  end
-
-  it "says true for Class.is_a?(Class.class.class) (#4374)" do
-    run("
-      Class.is_a?(Class.class.class)
-    ").to_b.should be_true
-  end
 end
--- a/src/class.cr
+++ b/src/class.cr
@@ -129,7 +129,7 @@ class Class
   # ```
   # Int32 | Char # => (Int32 | Char)
   # ```
-  def self.|(other : U.class) forall U
+  def |(other : U.class) forall U
     t = uninitialized self
     u = uninitialized U
     typeof(t, u)
--- a/src/compiler/crystal/tools/typed_def_processor.cr
+++ b/src/compiler/crystal/tools/typed_def_processor.cr
@@ -3,6 +3,8 @@
 #
 # It is used for `crystal tool context/expand/implementation`.
 module Crystal::TypedDefProcessor
+  @processed_types = Set(Type).new
+
   private def process_typed_def(typed_def : Def)
     typed_def.accept self
   end
@@ -17,6 +19,9 @@ module Crystal::TypedDefProcessor
   end
 
   private def process_type(type : Type) : Nil
+    return if @processed_types.includes?(type)
+    @processed_types << type
+
     if type.is_a?(NamedType) || type.is_a?(Program) || type.is_a?(FileModule)
       type.types?.try &.each_value do |inner_type|
         process_type inner_type
@@ -29,7 +34,7 @@ module Crystal::TypedDefProcessor
       end
     end
 
-    process_type type.metaclass if type.metaclass != type
+    process_type type.metaclass unless type.metaclass?
 
     if type.is_a?(DefInstanceContainer)
       type.def_instances.each_value do |typed_def|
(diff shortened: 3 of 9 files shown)


src/compiler/crystal/tools/typed_def_processor.cr
```diff
@@ -3,6 +3,8 @@
 #
 # It is used for `crystal tool context/expand/implementation`.
 module Crystal::TypedDefProcessor
+  @processed_types = Set(Type).new
```
asterite: This is now needed because otherwise it would recurse endlessly, I'm not sure if it's because of aliases or what... I'm surprised it worked before, whenever I traverse types in code (`AbstractDefChecker`, `RecursiveStructChecker`, etc.) I keep a registry of all processed types.

src/class.cr
```diff
@@ -129,7 +129,7 @@ class Class
   # ```
   # Int32 | Char # => (Int32 | Char)
   # ```
-  def self.|(other : U.class) forall U
+  def |(other : U.class) forall U
```
asterite: I don't know why this was a class method before, or even why it worked, but with this change it now needs to be an instance method.

spec/compiler/codegen/is_a_spec.cr
```diff
@@ -715,16 +715,4 @@ describe "Codegen: is_a?" do
       1.to_s.is_a?(B)
       )).to_b.should be_true
   end
-
-  it "says true for Class.is_a?(Class.class) (#4374)" do
```
asterite: These specs now make no sense.

straight-shoota:
I love it.

But should it really be metaclass-ception and recurse indefinitely? There are use cases for `X.class.class` but `X.class.class.class` etc.?

asterite:
We can probably stop at `X.class.class`. Let me try that...

asterite:
Actually, what would you get from doing `X.class.class.class`? I'm not sure... So maybe for now we can do it like this and eventually, thinking it well, change it but still making it work with the `T.class` restriction.

asterite:
@bew No

vladfaust:
Hey, @asterite, quick question: is it possible to get instance type from a class in runtime? I'd like to init a `Proc` with a variable type. Tried this code (https://carc.in/#/r/5wyf):

```crystal
type = Int32
pp Proc(type)
```

Prints `Syntax error in eval:2: expecting token 'CONST', not 'type'`.

Then I tried `typeof` (https://play.crystal-lang.org/#/r/5wye):

```crystal
type = Int32
pp Proc(typeof(type))
```

Almost got there: `Proc(Int32.class)`. So how to get `Int32` from `Int32.class` without macros? Given that I cannot use `42` as `type`, only explicit class.

asterite:
@vladfaust No, there's no way.

PR #7256, opened 2019-01-03
Overflow prepare stdlib

bcardiff:
The PR updates the stdlib to use unchecked operations & conversions where needed.

The changes should be read as: keep the current semantic once the operations will default to overflowing.

Some changes direct make sense because they manipulate data as bits mainly (eg: hasher).
Some might look a bit odd (eg: dwarf,  floatprinter). The changes were required to make specs pass once the overflow is turned on.
Other changes can be revised in future version (eg: time) since there is manual detection of overflow that could be simpler in the future.

### DIFF
--- a/src/debug/dwarf/info.cr
+++ b/src/debug/dwarf/info.cr
@@ -45,7 +45,7 @@ module Debug
           code = DWARF.read_unsigned_leb128(@io)
           attributes.clear
 
-          if abbrev = abbreviations[code - 1]? # abbreviations.find { |a| a.code == abbrev }
+          if abbrev = abbreviations[code &- 1]? # abbreviations.find { |a| a.code == abbrev }
             abbrev.attributes.each do |attr|
               value = read_attribute_value(attr.form)
               attributes << {attr.at, attr.form, value}
--- a/src/json/from_json.cr
+++ b/src/json/from_json.cr
@@ -68,7 +68,7 @@ end
 
 {% for type in %w(Int8 Int16 Int32 Int64 UInt8 UInt16 UInt32 UInt64) %}
   def {{type.id}}.new(pull : JSON::PullParser)
-    {{type.id}}.new(pull.read_int)
+    {{type.id}}.new!(pull.read_int)
   end
 {% end %}
 
--- a/src/json/pull_parser.cr
+++ b/src/json/pull_parser.cr
@@ -250,35 +250,35 @@ class JSON::PullParser
   end
 
   def read?(klass : Int8.class)
-    read_int.to_i8 if kind == :int
+    read_int.to_i8! if kind == :int
   end
 
   def read?(klass : Int16.class)
-    read_int.to_i16 if kind == :int
+    read_int.to_i16! if kind == :int
   end
 
   def read?(klass : Int32.class)
-    read_int.to_i32 if kind == :int
+    read_int.to_i32! if kind == :int
   end
 
   def read?(klass : Int64.class)
-    read_int.to_i64 if kind == :int
+    read_int.to_i64! if kind == :int
   end
 
   def read?(klass : UInt8.class)
-    read_int.to_u8 if kind == :int
+    read_int.to_u8! if kind == :int
   end
 
   def read?(klass : UInt16.class)
-    read_int.to_u16 if kind == :int
+    read_int.to_u16! if kind == :int
   end
 
   def read?(klass : UInt32.class)
-    read_int.to_u32 if kind == :int
+    read_int.to_u32! if kind == :int
   end
 
   def read?(klass : UInt64.class)
-    read_int.to_u64 if kind == :int
+    read_int.to_u64! if kind == :int
   end
 
   def read?(klass : Float32.class)
--- a/src/number.cr
+++ b/src/number.cr
@@ -47,7 +47,7 @@ struct Number
   macro slice(*nums, read_only = false)
     %slice = Slice({{@type}}).new({{nums.size}}, read_only: {{read_only}})
     {% for num, i in nums %}
-      %slice.to_unsafe[{{i}}] = {{@type}}.new({{num}})
+      %slice.to_unsafe[{{i}}] = {{@type}}.new!({{num}})
     {% end %}
     %slice
   end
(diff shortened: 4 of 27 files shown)


src/json/from_json.cr
```diff
@@ -68,7 +68,7 @@ end
 
 {% for type in %w(Int8 Int16 Int32 Int64 UInt8 UInt16 UInt32 UInt64) %}
   def {{type.id}}.new(pull : JSON::PullParser)
-    {{type.id}}.new(pull.read_int)
+    {{type.id}}.new!(pull.read_int)
```
asterite: Shouldn't this raise on overflow?

src/json/pull_parser.cr
```diff
@@ -250,35 +250,35 @@ class JSON::PullParser
   end
 
   def read?(klass : Int8.class)
-    read_int.to_i8 if kind == :int
+    read_int.to_i8! if kind == :int
```
asterite: Shouldn't this, and all of the others, raise on overflow? Or maybe not raise, but return `nil` because this is named `read?`

src/yaml/from_yaml.cr
```diff
@@ -48,7 +48,7 @@ end
 
 {% for type in %w(Int8 Int16 Int32 Int64 UInt8 UInt16 UInt32 UInt64) %}
   def {{type.id}}.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
-    {{type.id}}.new parse_scalar(ctx, node, Int64)
+    {{type.id}}.new! parse_scalar(ctx, node, Int64)
```
asterite: Same comment as in JSON: shouldn't this raise on overflow?

src/yaml/from_yaml.cr
```diff
@@ -57,7 +57,7 @@ def String.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
 end
 
 def Float32.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
-  parse_scalar(ctx, node, Float64).to_f32
+  parse_scalar(ctx, node, Float64).to_f32!
```
asterite: Same comment as in JSON: shouldn't this raise on overflow?

src/number.cr
```diff
@@ -47,7 +47,7 @@ struct Number
   macro slice(*nums, read_only = false)
     %slice = Slice({{@type}}).new({{nums.size}}, read_only: {{read_only}})
     {% for num, i in nums %}
-      %slice.to_unsafe[{{i}}] = {{@type}}.new({{num}})
+      %slice.to_unsafe[{{i}}] = {{@type}}.new!({{num}})
```
asterite: I'm thinking this should raise on overflow. Same with `static_array`. That it works now and there's a test for it is just because it was working like that so far.

src/json/from_json.cr
bcardiff: Not quite. The json lexer tocken has a `int_value : Int64` and the support for big UInt64 values comes from an intermediate negative number stored in `int_value`. So... unless there is a broader refactor we need the `new!`.

asterite: What about the other types? (`Int32, `Int8`, etc.).

But yes, it worked like that before (no raise on overflow) so once we have overflow working we can probably tackle it separately, so don't worry.

src/json/pull_parser.cr
bcardiff: I would answer all json/yaml comments with: the change is to keep the current behavior. And can't be refactored until overflow is turned on by default.

asterite: Makes total sense, thanks!

src/number.cr
bcardiff: There is a spec that explicitly use out of range values. We either break introduce a breaking change (that maybe nobody cares) or we keep the current behavior.

I agree that it's odd to allow negative numbers in a `UInt8.slice(-1)`. I just didn't want to introduce breaking changes here.

asterite: If it's to avoid a breaking change then yes, let this silently overflow for now 👍

src/debug/dwarf/info.cr
```diff
@@ -45,7 +45,7 @@ module Debug
           code = DWARF.read_unsigned_leb128(@io)
           attributes.clear
 
-          if abbrev = abbreviations[code - 1]? # abbreviations.find { |a| a.code == abbrev }
+          if abbrev = abbreviations[code &- 1]? # abbreviations.find { |a| a.code == abbrev }
```
RX14: There should be heavily TODO'd, or compiled into an issue. If we just do a blanket "everything which fails due to overflow gets the old semantics" we miss a fantastic opportunity to get rid of bugs!

src/json/from_json.cr
RX14: This must be commented in the code, or compiled into an issue, etc.

This is rushing.

src/pointer.cr
```diff
@@ -95,7 +95,7 @@ struct Pointer(T)
     # TODO: If throwing on overflow for integer conversion is implemented,
     # then (here and in `Pointer#-`) for a `UInt64` argument the call to
     # `to_i64` should become `as_unsafe`.
```
RX14: This comment should have been replaced with an explanation.

RX14:
This is a huge missed opportunity. The non-obvious cases where unchecked arithmetic is used *must* be accompanied by a comment, and all the used *must* be audited.

I have no confidence that if we say we'll do those bugfixes and refactors "after we've implemented overflow", that they'll ever get done. There at the very least needs to be a high-priority issue enumerating these cases, and actively working on fixing them. And this issue needs to be completed before 0.27.1. Silently copying the existing behaviour is dangerous.

bcardiff:
I do want to refactor but we can't for example refactor Time to use overflow until overflow is the default behavior (or we need to have a huge macro if to use overflow if enabled). I think that is better to than after.

Other cases like dwarf requires some more digging and I wasn't able to dig deeper without further time. At worst, we are emitting the same code as before. Not worse.

I expect that there are some more missing parts of the std lib that will need s/+/&+/ but were not detected with the current spec coverage.

This PR is a whole TODO after overflow is turned on IMO. But doing it before will requiere code duplication or efforts that can happen later and improve stdlib once we have more a solid ground.

RX14:
@bcardiff I'm fine with leaving time, since it has a comment. I'm fine with leaving DWARF too, but it needs a comment or an issue or something to track this technical debt which has just been created.

JSON and YAML should be refactored before 0.27.1. `Number.slice` should be raise-on-overflow for 0.28.0. There are also some comments that need to be updated, but thats less of an issue.

RX14:
Have we done any of these refactors?

Issue #4165, opened 2017-03-19
crystal init assumes Git


rkeene:
crystal init creates a Git repo, even if some other version control system would be more preferable to the user

kirbyfan64:
...like what? Shards only supports Git right now, and IMO Crystal shouldn't pretend to support others unless they're fully supported in the ecosystem.

If you want to use something else, just do `rm -rf .git`.

bcardiff:
The git command is used to also grab author and email when expanding templates.
If there is no git in the system it should expand even so.

I pretty much agree with @kirbyfan64.

I wouldn't be against of a `--skip-git` option in crystal init that would skip invoking git related commands (.gitignore/git init/git config). But that is as much as we should consider today.

ozra:
`init` etc. commands would be very well off in their own binaries, called from the `crystal` binary. Like git calls external helper utils.

Then it's piece of cake switching out functionality like sniffing user name, and initializing repo in `ag silver-nugget-vcs` format (_obviously made up_) or whatever.

- configure your preferred settings in a user conf that crystal looks at (`vcs: ag`), to avoid flag every time
- crystal calls `crystal-vcs-hook-ag sniff-user-info`, `crystal-vcs-hook-ag init`, etc.

Code is not dependant on compiler libs, nor is the compiler dependant on util code like this - so it should bring down unnecessary compilation time also, separate concerns.

Not that I find this urgent at all. I'm completely fine with git, it does everything better than anything I used _before_ - and there is nothing that I miss really.

drosehn:
FWIW, I do prefer `subversion` for my own personal projects.  And just to show how really odd I am, I do keep meaning to try `mercurial` too.  (I keep installing it on all my systems, but haven't gotten around to using it yet).  However, I don't feel it's urgent for crystal to do anything about that.  The ideas that ozra described seem pretty reasonable to me.  And for now, supporting `git` is better than **not** supporting `git`!

ysbaddaden:
Shards is meant, and capable, to support any VCS. Missing support for SVN and Mercurial is just because I don't use them. Pull requests are welcome :-)

Then, we could allow to configure the VCS with `crystal init --svn` for example, but still defaulting to `git`.

j8r:
[...]

PR #7259, opened 2019-01-03
WIP: Remove init command

j8r:
This PR removes the `crystal init` command.

## Initial problem
Beginners probably wants an easy way to initialize a new *standard* Crystal project:

To do it, (s)he has to choose between two commands:
- `shards init`
- `crystal init`

The first issue with this is: which one to use?

## Design considerations

As said in https://github.com/crystal-lang/crystal/pull/7138#issuecomment-445372645 , there is an inter-dependency where it shouldn't have one:

- the Crystal compiler itself shall not know shards
- shards, of course, know and depends of Crystal

Other languages like Rust with Cargo, Nodejs with npm, Go with dep, ruby with gem/bundler have this separation design.
Usually when there are project initialization logic, it lives on the package manager, not in the compiler: `cargo new`, `npm init`, `dep init`, `bundle init`.

## Solution

I'm working on merging it to `shards init` in https://github.com/crystal-lang/shards/blob/master/src/commands/init.cr , unless @ysbaddaden has a better idea?


asterite:
Note: Crystal does now about `shards`. The whole `require` logic is built around that. I personally think `crystal init` should stay.

ysbaddaden:
Don't spend time on this. Shards isn't a Crystal project management tool, but a dependency manager *for* Crystal projects. The `shards init` command is only meant to quickly create a `shard.yml` file in an *existing* project, for people like me that just creates a `foo/src` folder and start hacking *then* need to add a dependency, and don't want to create the `shard.yml` by hand.

j8r:
@ysbaddaden So shards is similar to gem, and potentially a new project would be similar to `bundler`?

j8r:
Whats the `crystal` command supposed to be?

Is there a compiler that provides project management, in addition to source code management?
For example, for go:
```
$ go
Go is a tool for managing Go source code.
```
Project management isn't going to be light, people will want new features to better initialize/manage their projects, which would likely be refused:
- adding other [CVS than git](https://github.com/crystal-lang/crystal/issues/4165)
- adding [shards version integration](https://github.com/crystal-lang/crystal/pull/5428)

asterite:
Yes, there is: Crystal

straight-shoota:
@asterite 
> Note: Crystal does now about shards. The whole require logic is built around that.

For the record: That's is not entirely correct. The compiler expects dependencies in the `lib` folder and looks for them in specific locations. This has certainly be designed with shards in mind (or literally *with* shards), but Crystal does not care *how* dependencies get there.

On the matter: While I'm not strictly opposed to the idea, I don't think there is much benefit worth the hassle. Besides, this PR puts the cart before the horse.

RX14:
Maybe in the future we will remove `crystal init` (unlikely, bikeshedding over clean separation of compiling code and dependency management and project management tools isn't productive), but this PR is not it. It will need far more discussion and a concrete plan, not just a PR to remove the code with no replacement.

j8r:
I guess we need a RFC repo like in Rust for stuff like this - opening an issue.

PR #7262, opened 2019-01-03
Overflow prepare compiler

bcardiff:
The PR updates the stdlib to use unchecked operations & conversions where needed.
It's easier to review per commit than per changes.

compiler_specs need to use `&+` rather than `+` or including prelude would be required.
The `to_X!` operations are used when calling libs instead of `to_X`.

### DIFF
--- a/src/compiler/crystal/semantic/top_level_visitor.cr
+++ b/src/compiler/crystal/semantic/top_level_visitor.cr
@@ -676,10 +676,10 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
           if counter == 0 # In case the member is set to 0
             1
           else
-            counter * 2
+            counter &* 2
           end
         else
-          counter + 1
+          counter &+ 1
         end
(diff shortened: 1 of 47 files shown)


src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -676,10 +676,10 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
           if counter == 0 # In case the member is set to 0
             1
           else
-            counter * 2
+            counter &* 2
           end
         else
-          counter + 1
+          counter &+ 1
```
RX14: An issue to fix or discuss this behaviour should be created.

bcardiff: Done at #7268. Is there any other feedback of the changes of the PR?

RX14: No, the PR is good! Thanks!

PR #7266, opened 2019-01-04
Compiler: error if defining initialize inside enum

### LINKED ISSUE #7238: Compiler bug redefining an enum's initialize
Code:
```cr
enum Foo
  FOO = 1

  def initialize
  end
end
```
Compiler error:
```
Missing hash key: Foo (KeyError)
  from src/hash.cr:126:9 in 'fetch'
  from src/hash.cr:62:5 in '[]'
  from src/compiler/crystal/semantic/type_guess_visitor.cr:1131:9 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/semantic/semantic_visitor.cr:105:27 in 'visit'
  from src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from src/compiler/crystal/syntax/ast.cr:169:27 in 'accept_children'
  from src/compiler/crystal/syntax/visitor.cr:28:11 in 'accept'
  from src/compiler/crystal/semantic/hooks.cr:26:7 in 'visit_with_finished_hooks'
  from src/compiler/crystal/semantic/type_declaration_processor.cr:140:5 in 'process'
  from src/compiler/crystal/semantic.cr:70:7 in 'top_level_semantic'
  from src/compiler/crystal/semantic.cr:22:23 in 'semantic'
  from src/compiler/crystal/compiler.cr:152:7 in 'compile'
  from src/compiler/crystal/command/eval.cr:29:14 in 'eval'
  from src/compiler/crystal/command.cr:82:7 in 'run'
  from src/compiler/crystal/command.cr:46:5 in 'run'
  from src/compiler/crystal/command.cr:45:3 in 'run'
  from src/compiler/crystal.cr:8:1 in '__crystal_main'
  from __libc_start_main
  from _start
  from ???
```

Redefining `initialize` in an enum type should probably be a syntax error.

asterite:
Fixes #7238

### DIFF
--- a/spec/compiler/semantic/enum_spec.cr
+++ b/spec/compiler/semantic/enum_spec.cr
@@ -436,4 +436,16 @@ describe "Semantic: enum" do
     method = result.program.types["Foo"].lookup_first_def("bar", block: false).not_nil!
     method.always_inline?.should be_true
   end
+
+  it "errors if defining initialize in Enum (#7238)" do
+    assert_error %(
+      enum Foo
+        FOO = 1
+
+        def initialize
+        end
+      end
+      ),
+      "enums can't define an `initialize` method, try using `def self.new`"
+  end
 end
--- a/spec/compiler/semantic/enum_spec.cr
+++ b/spec/compiler/semantic/enum_spec.cr
@@ -448,4 +448,18 @@ describe "Semantic: enum" do
       ),
       "enums can't define an `initialize` method, try using `def self.new`"
   end
+
+  it "can redefine Enum.new" do
+    assert_type(%(
+      enum Foo
+        FOO = 1
+
+        def self.new(x : Int32)
+          "hello"
+        end
+      end
+
+      Foo.new(1)
+      )) { string }
+  end
 end
--- a/spec/compiler/codegen/enum_spec.cr
+++ b/spec/compiler/codegen/enum_spec.cr
@@ -322,4 +322,19 @@ describe "Code gen: enum" do
       x
       )).to_i.should eq(1)
   end
+
+  it "can redefine Enum.new and use previous_def" do
+    run(%(
+      enum Foo
+        FOO = 1
+        BAR = 2
+
+        def self.new(x : Int32)
+          previous_def(2)
+        end
+      end
+
+      Foo.new(1)
+      )).to_i.should eq(2)
+  end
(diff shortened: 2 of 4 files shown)


spec/compiler/semantic/enum_spec.cr
```diff
@@ -436,4 +436,16 @@ describe "Semantic: enum" do
...
+  it "errors if defining initialize in Enum (#7238)" do
+    assert_error %(
+      enum Foo
+        FOO = 1
+
+        def initialize
+        end
+      end
+      ),
+      "enums can't define an `initialize` method, try using `def self.new`"
```
straight-shoota: This advise is not good, because `def self.new` doesn't work either:
```cr
enum Foo
  FOO = 1
  
  def self.new(value)
    puts "self.new caleld"
    new(value)
  end
end

Foo.new(1)
```

straight-shoota:
That would fall under "don't suggest using it" 😬 

For `HTTP::Status` it would be nice if `.new` could be overriden instead of having to resort to `.code` for adding a value verification to the initializer.

asterite:
I just pushed a commit to allow redefining `Enum.new`. But to do it you have to put the correct type restriction, otherwise it will take the other `new` (for example if you don't put any type restriction it will choose the other one because it's more strict). And then you have to use `previous_def` to invoke the original `new`.

spec/compiler/semantic/enum_spec.cr
```diff
@@ -436,4 +436,30 @@ describe "Semantic: enum" do
...
+  end
+
+  it "can redefine Enum.new" do
+    assert_type(%(
+      enum Foo
+        FOO = 1
+
+        def self.new(x : Int32)
+          "hello"
```
straight-shoota: Maybe make this test ensure that overriding with `previous_def` works as expected?

asterite: I just added a separate test in `codegen` that runs the code and checks this (with just the type you can't know whether it passed the correct value). Let me know if you think this is fine.

RX14:
This isn't a breaking change since defining `def initialize` didn't work before, right?

asterite:
@RX14 Exactly

PR #7269, opened 2019-01-04
Add Char#+(Char) and fix <=>'s behaviour

r00ster91:
This adds the concatenation of two chars and fixes the behaviour of the `<=>` operator of `Char`.

I feel like `Char#+(Char)` was missing and right now in my code I have `'"' + char + '"'` which I would like to use instead of `%("#{char}")` or even worse `"\"#{char}\""`. I personally find the first one the easiest to read.
Also, the first one is much faster:
```cr
              +  11.55M ( 86.58ns) (±25.21%)   48 B/op        fastest
percent literal   6.18M ( 161.7ns) (±38.42%)  208 B/op   1.87× slower
       escaping   7.45M (134.16ns) (±33.66%)  208 B/op   1.55× slower
```
---
The behaviour of the [spaceship operator](https://crystal-lang.org/api/0.27.0/Char.html#%28other%3AChar%29-instance-method) is currently wrong. It's just an alias of [`Char#-`](https://crystal-lang.org/api/0.27.0/Char.html#-%28other%3AChar%29-instance-method).
Right now `'a' <=> 'c'` returns `-2` instead of `-1`.

This also adds some missing specs.

### DIFF
--- a/src/char.cr
+++ b/src/char.cr
@@ -63,6 +63,28 @@ struct Char
     ord - other.ord
   end
 
+  # Concatenates this char and *other*.
+  #
+  # ```
+  # 'a' + 'b' # => "ab"
+  # ```
+  def +(other : Char)
+    bytesize = other.bytesize + self.bytesize
+    String.new(bytesize) do |buffer|
+      index = 0
+      other.each_byte do |byte|
+        buffer[index] = byte
+        index += 1
+      end
+      self.each_byte do |byte|
+        buffer[index] = byte
+        index += 1
+      end
+
+      {bytesize, 2}
+    end
+  end
+
   # Concatenates this char and *string*.
   #
   # ```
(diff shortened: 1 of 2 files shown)


src/char.cr
```diff
@@ -63,6 +63,28 @@ struct Char
...
+  # ```
+  def +(other : Char)
+    bytesize = other.bytesize + self.bytesize
+    String.new(bytesize) do |buffer|
+      index = 0
+      other.each_byte do |byte|
+        buffer[index] = byte
+        index += 1
+      end
+      self.each_byte do |byte|
```
Sija: This loop should come first.

bcardiff:
Seconded. I think we can relax the documentation of `<=>`. 

Regarding Char + String. I would vote for no. But then we actually have String + Char. I would actually remove that as well to favor more the usage of the StringBuilder in that case. If there is a single char the user can do string + char.to_s, but allowing string + char can lead to bad performant code IMO.

Closing to move on that directions. But, thanks @r00ster91 to bring the discussion.

PR #7275, opened 2019-01-05
 Add yielding line number to {String,IO}#each_line

straight-shoota:
When iterating through lines you often need a line number. This PR adds them as a second (optional) argument to the block provided to `String#each_line`, `IO#each_line` and `File.each_line`.
In the updated API docs I removed unrelated features `upcase` and `reverse` from the examples.

Alternatively, it could be a separate method called `#each_line_with_number` but I think the proposed option is better.

A second commit in this PR refactors related specs in `io_spec.cr` which I noticed could be simplified while adding the specs for yielding a line number.

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -908,17 +908,26 @@ abstract class IO
   # ```
   # io = IO::Memory.new("hello\nworld")
   # io.each_line do |line|
-  #   puts line.chomp.reverse
+  #   puts line
   # end
+  # # output:
+  # # hello
+  # # world
   # ```
   #
-  # Output:
+  # The second argument yielded to the block is the line number starting at `1`.
+  # The optional *offset* argument is added to the first line number.
   #
-  # ```text
-  # olleh
-  # dlrow
   # ```
-  def each_line(*args, **options) : Nil
+  # io = IO::Memory.new("hello\nworld")
+  # io.each_line do |line, lino|
+  #   puts "#{lino}: #{line}"
+  # end
+  # # output:
+  # # 1: hello
+  # # 2: world
+  # ```
+  def each_line(*args, **options, &block : String -> _) : Nil
     while line = gets(*args, **options)
       yield line
     end
--- a/src/io.cr
+++ b/src/io.cr
@@ -933,6 +933,30 @@ abstract class IO
     end
   end
 
+  # Reads this `IO` line by line and yields each line and the line number to a block.
+  #
+  # See `#each_line` for basic behaviour.
+  #
+  # The second argument yielded to the block is the line number starting at `1`.
+  # The optional *offset* argument is added to the first line number.
+  #
+  # ```
+  # io = IO::Memory.new("hello\nworld")
+  # io.each_line do |line, lino|
+  #   puts "#{lino}: #{line}"
+  # end
+  # # output:
+  # # 1: hello
+  # # 2: world
+  # ```
+  def each_line_with_number(*args, offset = 0, **options, &block : String, Int32 -> _) : Nil
+    line_number = offset
+    each_line do |line|
+      line_number += 1
+      yield line, line_number
+    end
+  end
+
   # Returns an `Iterator` for the *lines* in this `IO`, where a line
   # is defined by the arguments passed to this method, which can be the same
   # ones as in the `gets` methods.
--- a/src/string.cr
+++ b/src/string.cr
@@ -3365,23 +3365,49 @@ class String
   def each_line(chomp = true, &block : String -> _) : Nil
     return if empty?
 
-    offset = 0
-
-    while byte_index = byte_index('\n'.ord.to_u8, offset)
-      count = byte_index - offset + 1
+    byte_offset = 0
+    while byte_index = byte_index('\n'.ord.to_u8, byte_offset)
+      count = byte_index - byte_offset + 1
       if chomp
         count -= 1
-        if offset + count > 0 && to_unsafe[offset + count - 1] === '\r'
+        if byte_offset + count > 0 && to_unsafe[byte_offset + count - 1] === '\r'
           count -= 1
         end
       end
 
-      yield unsafe_byte_slice_string(offset, count)
-      offset = byte_index + 1
+      yield unsafe_byte_slice_string(byte_offset, count)
+      byte_offset = byte_index + 1
     end
 
-    unless offset == bytesize
-      yield unsafe_byte_slice_string(offset)
+    unless byte_offset == bytesize
+      yield unsafe_byte_slice_string(byte_offset)
+    end
+  end
+
+  # Splits the string after each newline and yields each line and the line number to a block.
+  #
+  # See `#each_line` for basic behaviour.
+  #
+  # The second argument yielded to the block is the line number starting at `1`.
+  # The optional *offset* argument is added to the first line number.
+  #
+  # ```
+  # haiku = "the first cold shower
+  # even the monkey seems to want
+  # a little coat of straw"
+  # haiku.each_line do |stanza, lino|
+  #   puts "#{lino}: #{stanza}"
+  # end
+  # # output:
+  # # 1: the first cold shower
+  # # 2: even the monkey seems to want
+  # # 3: a little coat of straw
+  # ```
+  def each_line_with_number(chomp = true, offset : Int32 = 0, &block : String, Int32 -> _) : Nil
+    line_number = offset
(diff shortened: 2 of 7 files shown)


src/io.cr
```diff
@@ -908,19 +908,30 @@ abstract class IO
...
+  # io = IO::Memory.new("hello\nworld")
+  # io.each_line do |line|
+  #   puts "#{lino}: #{line}"
+  # end
+  # # output:
+  # # 1: hello
+  # # 2: world
+  # ```
+  def each_line(*args, offset = 0, **options, &block : String, Int32 -> _) : Nil
+    lino = offset
```
RX14: `line_number`, please.

src/io.cr
```diff
@@ -908,19 +908,30 @@ abstract class IO
   # ```
   # io = IO::Memory.new("hello\nworld")
   # io.each_line do |line|
-  #   puts line.chomp.reverse
+  #   puts line
```
bcardiff: The chomp maybe was there from times when the each_line didn't have a chomp option.

I think it's unclear what are the available options. We should add see `IO#gets` for available options.

src/string.cr
```diff
@@ -3340,17 +3340,33 @@ class String
...
+  # end
+  # # output:
+  # # 1: the first cold shower
+  # # 2: even the monkey seems to want
+  # # 3: a little coat of straw
+  # ```
+  def each_line(chomp = true, offset lino : Int32 = 0, &block : String, Int32 -> _) : Nil
     return if empty?
 
     offset = 0
```
bcardiff: Can we avoid the clashing of internal variable and external param name?

src/io.cr
straight-shoota: I'd like to change that in a different PR.

src/io.cr
```diff
@@ -908,19 +908,30 @@ abstract class IO
...
-  # olleh
-  # dlrow
   # ```
-  def each_line(*args, **options) : Nil
+  # io = IO::Memory.new("hello\nworld")
+  # io.each_line do |line|
+  #   puts "#{lino}: #{line}"
+  # end
+  # # output:
+  # # 1: hello
```
r00ster91: Maybe put the output in an external ` ```text ` code block.

src/io.cr
```diff
@@ -908,19 +908,30 @@ abstract class IO
...
+  # The second argument yielded to the block is the line number starting at `1`.
+  # The optional *offset* argument is added to the first line number.
   #
-  # ```text
-  # olleh
-  # dlrow
   # ```
-  def each_line(*args, **options) : Nil
+  # io = IO::Memory.new("hello\nworld")
+  # io.each_line do |line|
```
Sija: Missing 2nd `lino` argument.

src/io.cr
```diff
@@ -908,19 +908,30 @@ abstract class IO
...
+  # # output:
+  # # 1: hello
+  # # 2: world
+  # ```
+  def each_line(*args, offset = 0, **options, &block : String, Int32 -> _) : Nil
+    line_number = offset
     while line = gets(*args, **options)
-      yield line
+      line_number += 1
+      yield line, lino
```
Sija: Typo: `lino`

src/io.cr
straight-shoota: No, that separates it from the example.

src/string.cr
```diff
@@ -3340,32 +3340,50 @@ class String
...
-  def each_line(chomp = true) : Nil
+  #
+  # The second argument yielded to the block is the line number starting at `1`.
+  # The optional *offset* argument is added to the first line number.
+  #
+  # ```
+  # haiku = "the first cold shower
+  # even the monkey seems to want
+  # a little coat of straw"
+  # haiku.each_line do |stanza, lino|
```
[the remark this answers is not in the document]
Sija: Btw, conventially this shortcut is written as `lineno` which I'd suggest using instead of `lino`.

ysbaddaden:
What happens if you have a file with more than Int32::MAX lines? Or connect a TCP socket to a server that will eventually stream more than this number of lines?

I guess overflow, whatever if we actually use the line_number variable

src/string.cr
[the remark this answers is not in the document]
r00ster91: `index` is also an option. Let's please use full written names.

Sija:
Having unified `Integer` type would be a bliss...

asterite:
I played a bit with implementing a unified Integer type, it's super slow, or at least like 1000 times slower than primitives. Maybe it could be further optimized, I don't know, but I think it's very unlikely we'll have something like that unless someone starts to really play with the idea and finds a way to make it fast (the gist is a type that holds either an int or a big int under the hood)

RX14:
you can always use `offset: 0_u64` to stop any overflows.

ysbaddaden:
@RX14 It makes it less likely, but can still happen, if I create a connection and write messages for some weeks. I can specify a u128, but then `io.read_line(offset: 0_u128)` becomes more complex/confusing than `while line = io.gets; end` which is simpler.

I'm not saying it's a bad idea to pass the iteration line number, but... aren't we repeating the `loop` yields the iteration count idea (not bad) but that we eventually dropped because of this very reason (overflow)?

straight-shoota:
We could also make this a separate method `#each_line_with_number`.

Another option is to simply count on the offending counter being optimized out by LLVM when the line number block arg is not used 😏

asterite:
I don't think overflowing is a problem. I'd probably use the line number when iterating a file's lines, and file lines are super unlikely to overflow.

That said, iterating and using an extra local variable for this not-that-common use case is fine too. Otherwise, I'd name it `each_line_with_index` like we have in other places.

straight-shoota:
Yes, when using files you're probably save, but @ysbaddaden's example of a long running server connection doesn't sound too far-fetched. And it's definitely a use case to count the number of received lines.

straight-shoota:
Separated into `each_line_with_number`

RX14:
Then I don't see the point. Is this really used commonly enough to justify a new method?

straight-shoota:
I know I have used this quite a few times, that's why I'd welcome it as a helpful addition.
There is one use case in stdlib, #7252 shows another one.

If that's enough to justify adding another method (or actually three), I don't know.

straight-shoota:
I have extracted the uncontested parts of this PR to #7419.

ysbaddaden:
No need to add another method.

I wanted to emphasize that maybe it's not a good idea to introduce a counter in a loop function, which we did once and eventually reverted.

If we expect `#each_line` to be mostly for files, then it's fine to have a counter, even an Int32 should be fine. I just wish it wouldn't raise on index overflow for other cases, especially if they don't use the index.

sudo-nice:
I dream someday this works:
```crystal
File.each_line("some_file").with_index(1) do |line, idx|
  puts "#{idx}: #{line}"
end
```

j8r:
Is it that complicate to create our own counter? This extends the API to do a simple thing, I don't see the need.

Having the control of the counter, one can decide to be Int32 or even UInt64 to reduce the chance of overflow, depending of the use case.

sudo-nice:
This works already:
```crystal
File.open("some_file").each_line.with_index(1) do |line, idx|
  puts "#{idx}: #{line}"
end
```

straight-shoota:
@sudo-nice Using an iterator is less efficient than `#each_line` with a block.

@ysbaddaden I agree that adding a separate method is not a great idea. Either we integrate line numbers into `#each_line` or don't.

How about using a wrapping counter? This way it won't raise but simply overflow. I figure this would be fine because when using line numbers makes any sense, they're gonna be far smaller than `Int32::MAX`. But this leaves the option to use `#each_line` without line numbers for infinite lines and it won't raise.

ysbaddaden:
A wrapping on overflow counter would kill the whole point of having integer overflow checks in the first place. IMO the iterator is the best solution: no need for yet-another-method, no need for always having a counter.

asterite:
We have many variants of `each`, `each_with_index` in the std, why not have `each_line` and `each_line_with_index` then? The "problem" is that the index starts at zero, but it's just about doing `each_line_with_index(1)` and that's it. Then it's consistent with the rest of the std.

jhass:
I have to agree here that the usecases are to few to really warrant the extra API. Using the iterator or a closured var is simple enough. Also there hasn't been any interest in this anymore for over a year, so the pain can't be big. I vote to close this :)

PR #7277, opened 2019-01-05
Allow different values than only -1, 0 and 1 to be returned by <=>

### LINKED ISSUE #7276: Allow <=> to not only return -1, 0 or 1
Currently the documentation of [`Comparable`](https://crystal-lang.org/api/0.27.0/Comparable.html) and [`PartialComparable`](https://crystal-lang.org/api/0.27.0/PartialComparable.html) says that classes which include one of them must implement `<=>` and it must return either `1`, `0` or `-1` (or `nil` in case of `PartialComparable`). But why? Is there a specific reason for that? There's not really a benefit behind being so strict. It should probably be changed so that the spaceship is allowed to return either a positive number, zero or a negative number. A good example is `Char`'s [`<=>`](https://crystal-lang.org/api/0.27.0/Char.html#%28other%3AChar%29-instance-method) which already does that.

See also the discussion in: #7269

r00ster91:
Closes #7276 

This changes the spaceship's documentation to allow more values than only `-1`, `0` and `1` to be returned.

### DIFF
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -1180,7 +1180,7 @@ describe "Array" do
     b = [4, 5, 6]
     c = [1, 2]
 
-    (a <=> b).should be < 1
+    (a <=> b).should be < 0
     (a <=> c).should be > 0
     (b <=> c).should be > 0
     (b <=> a).should be > 0
--- a/src/array.cr
+++ b/src/array.cr
@@ -167,11 +167,11 @@ class Array(T)
     false
   end
 
-  # Combined comparison operator. Returns *0* if `self` equals *other*, *1* if
-  # `self` is greater than *other* and *-1* if `self` is smaller than *other*.
+  # Combined comparison operator. Returns `0` if `self` equals *other*, `1` if
+  # `self` is greater than *other* and `-1` if `self` is smaller than *other*.
   #
   # It compares the elements of both arrays in the same position using the
-  # `<=>` operator.  As soon as one of such comparisons returns a non-zero
+  # `<=>` operator. As soon as one of such comparisons return a non-zero
   # value, that result is the return value of the comparison.
   #
   # If all elements are equal, the comparison is based on the size of the arrays.
--- a/src/array.cr
+++ b/src/array.cr
@@ -171,7 +171,7 @@ class Array(T)
   # `self` is greater than *other* and `-1` if `self` is smaller than *other*.
   #
   # It compares the elements of both arrays in the same position using the
-  # `<=>` operator. As soon as one of such comparisons return a non-zero
+  # `<=>` operator. As soon as one of such comparisons returns a non-zero
   # value, that result is the return value of the comparison.
   #
   # If all elements are equal, the comparison is based on the size of the arrays.
--- a/src/array.cr
+++ b/src/array.cr
@@ -167,8 +167,10 @@ class Array(T)
     false
   end
 
-  # Combined comparison operator. Returns `0` if `self` equals *other*, `1` if
-  # `self` is greater than *other* and `-1` if `self` is smaller than *other*.
+  # Combined comparison operator.
+  #
+  # Returns `-1`, `0` or `1` depending on whether `self` is less than *other*, equals *other*
+  # or is greater than *other*.
   #
   # It compares the elements of both arrays in the same position using the
   # `<=>` operator. As soon as one of such comparisons returns a non-zero
--- a/src/char.cr
+++ b/src/char.cr
@@ -106,7 +106,7 @@ struct Char
   # The comparison operator.
   #
   # Returns the difference of the codepoint values of `self` and *other*.
-  # The result is either negative, zero or positive based on whether `other`'s codepoint is
+  # The result is either negative, `0` or positive based on whether `other`'s codepoint is
   # less, equal, or greater than `self`'s codepoint.
   #
   # ```
--- a/src/char.cr
+++ b/src/char.cr
@@ -105,6 +105,11 @@ struct Char
 
   # The comparison operator.
   #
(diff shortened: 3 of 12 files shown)


src/array.cr
```diff
@@ -167,11 +167,11 @@ class Array(T)
...
   end
 
-  # Combined comparison operator. Returns *0* if `self` equals *other*, *1* if
-  # `self` is greater than *other* and *-1* if `self` is smaller than *other*.
+  # Combined comparison operator. Returns `0` if `self` equals *other*, `1` if
+  # `self` is greater than *other* and `-1` if `self` is smaller than *other*.
   #
   # It compares the elements of both arrays in the same position using the
-  # `<=>` operator.  As soon as one of such comparisons returns a non-zero
+  # `<=>` operator. As soon as one of such comparisons return a non-zero
```
RX14: returns was correct before

src/char.cr
```diff
@@ -103,10 +103,12 @@ struct Char
     (ord - other).chr
   end
 
-  # Implements the comparison operator.
+  # The comparison operator.
```
RX14: This should be more verbose like the array docs

src/partial_comparable.cr
```diff
@@ -62,5 +62,16 @@ module PartialComparable(T)
     end
```
r00ster91: There's `def compare_with(other : T)` above this and it's only used by this module. Seems to be a helper method so shouldn't it be private?

src/number.cr
```diff
@@ -161,9 +161,10 @@ struct Number
     {(self / number).floor, self % number}
   end
 
-  # Implements the comparison operator.
+  # The comparison operator.
   #
-  # See also: `Object#<=>`.
```
r00ster91: Removed because `Object#<=>` no longer exists.

src/number.cr
```diff
@@ -161,9 +161,10 @@ struct Number
...
+  # Returns `-1` if `self` is less than *other*, `0` if `self` equals *other*
```
straight-shoota: IMO readability would be improved if this followed the schema used in the other descriptions to first name the different types of outcome and then the reasons for each.

src/partial_comparable.cr
straight-shoota: This module is scheduled to be removed in `0.28.0` anyway, so don't bother (see #6611).

straight-shoota: This module is scheduled to be removed in 0.28.0 anyway, so don't bother (see #6611).

spec/std/array_spec.cr
```diff
@@ -1180,7 +1180,7 @@ describe "Array" do
     b = [4, 5, 6]
     c = [1, 2]
 
-    (a <=> b).should be < 1
+    (a <=> b).should be < 0
```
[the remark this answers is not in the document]
r00ster91: No, I don't think there was any reason. Looks like a mistake to me. `a` and `b` are never equal.

src/comparable.cr
```diff
@@ -1,22 +1,22 @@
...
   end
 
-  # Compares this object to *other* based on the receiver’s `<=>` method, returning `true` if it returns `-1` or `0`.
+  # Compares this object to *other* based on the receiver’s `<=>` method, returning `true` if it returns a negative number or zero.
   def <=(other : T)
     (self <=> other) <= 0
   end
 
-  # Compares this object to *other* based on the receiver’s `<=>` method, returning `true` if it returns `0`.
+  # Compares this object to *other* based on the receiver’s `<=>` method, returning `true` if it returns zero.
```
straight-shoota: I'd prefer to use `0` instead of "zero" everywhere. It's an explicit value and we should use the appropriate literal for that.

src/comparable.cr
```diff
@@ -30,18 +30,20 @@ module Comparable(T)
...
+  # Compares this object to *other* based on the receiver’s `<=>` method, returning `true` if it returns a positive number or zero.
   def >=(other : T)
     (self <=> other) >= 0
   end
 
-  # Comparison operator. Returns `0` if the two objects are equal,
-  # a negative number if this object is considered less than *other*,
+  # The comparison operator.
+  #
+  # Returns zero if the two objects are equal,
```
straight-shoota: This should be ordered like all the other cases: negative, zero, positive.

src/partial_comparable.cr
```diff
@@ -62,5 +62,18 @@ module PartialComparable(T)
...
   end
 
+  # The comparison operator.
+  #
+  # Returns a negative number if the object
```
straight-shoota: Minor detail, but that's an awkward line break.

src/partial_comparable.cr
```diff
@@ -62,5 +62,18 @@ module PartialComparable(T)
...
+  # is considered to be less than *other*, zero if the two objects are equal,
+  # a positive number if this object is considered to be greater than *other*,
+  # or `nil` if no order can be established.
+  #
+  # Subclasses define this method to provide class-specific ordering.
+  #
+  # ```
+  # # Sort in a descending way
```
straight-shoota: This example is unneccesarily hard to follow. Maybe just prepend the same example with ascending order?

src/partial_comparable.cr
r00ster91: `PartialComparable` is going to be removed anyway.

src/comparable.cr
r00ster91: I used "zero" because "negative number" and "positive number" is also written in words instead of `-1` and `+1`, so for consistency. But "negative number" and "positive number" is very general while `0` is a unique value so they are still pretty different and `0` reads faster, so I will change that.

straight-shoota:
Thank you @r00ster91

PR #7280, opened 2019-01-06
Add #has_key? method to YAML::Any and JSON::Any?

delef:
```
data = YAML.parse <<-END
          ---
          foo:
            bar:
              baz:
                - qux
                - fox
          END

data.has_key? "foo" # => true
data.has_key? "fo" # => false
data["foo"]["bar"]["baz"].has_key? 0 # => true
data["foo"]["bar"]["baz"].has_key? 2 # => false
```

works the same way with `JSON::Any`

Thank you for your attention.

### DIFF
--- a/src/json/any.cr
+++ b/src/json/any.cr
@@ -115,6 +115,28 @@ struct JSON::Any
     end
   end
 
+  # Assumes the underlying value is a `Array`
+  # Returns `true` when index given by *index* exists, otherwise `false`
+  def has_key?(index : Int) : Bool
+    case object = @raw
+    when Array
+      object[index]? != nil
+    else
+      raise "Expected Array for #has_key?(index : Int), not #{object.class}"
+    end
+  end
+
+  # Assumes the underlying value is a `Hash`
+  # Returns `true` when key given by *key* exists, otherwise `false`
+  def has_key?(key : String) : Bool
+    case object = @raw
+    when Hash
+      object.has_key?(key)
+    else
+      raise "Expected Hash for #has_key?(key : String), not #{object.class}"
+    end
+  end
+
   # Traverses the depth of a structure and returns the value.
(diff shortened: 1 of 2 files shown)


src/json/any.cr
```diff
@@ -115,6 +115,28 @@ struct JSON::Any
     end
   end
 
+  # Assumes the underlying value is a `Array`
+  # Returns `true` when index given by *index* exists, otherwise `false`
+  def has_key?(index : Int) : Bool
+    case object = @raw
+    when Array
+      object[index]? != nil
```
j8r: `Array#has_key` doesn't exist,` Any#has_key?` must not exist too – what's the point?

src/yaml/any.cr
```diff
@@ -126,6 +126,23 @@ struct YAML::Any
     end
   end
 
+  # Assumes the underlying value is an `Array` or `Hash`
+  # Returns `true` when index_or_key given by *index_or_key* exists, otherwise `false`
+  def has_key?(index_or_key) : Bool
+    case object = @raw
+    when Array
+      if index_or_key.is_a?(Int)
+        object[index_or_key]? != nil
```
j8r: Ditto: `Array#has_key?` doesn't exists, `Any#[]?` is perfectly fine

asterite:
Note that we pretty much decided not to add more methods to these types. If you need them, just cast to the appropriate type.

delef:
> Note that we pretty much decided not to add more methods to these types. If you need them, just cast to the appropriate type.

I didn't know that you decided to freeze the current API

asterite:
It's not that. It's that you want `has_key?`, someone else will wants `keys`, and before you notice it the whole Hash and Array APIs will be replicated, when you could just have written `as_h.has_key?`.

Issue #7196, opened 2018-12-17
Default target triple baked into compiler fails on musl


straight-shoota:
Linux builds of the Crystal compiler produced by [crystal-lang/distribution-scripts](https://github.com/crystal-lang/distribution-scripts) (such as the ones provided for releases and nightlies) have a default target triple `x86_64-unknown-linux-gnu`. This seems to be hard coded into the binary.

Binaries are statically linked and run as well on linux-gnu as on linux-musl. When running on Alpine however, this default target triple is incorrect and results in linker errors (e.g. ``undefined reference to `errno'``). This makes it pretty much unusable. The correct target triple would be `x86_64-alpine-linux-musl`. This can be mitigated by explicitly passing `--cross-compile --target x86_64-alpine-linux-musl`, but that's not very user-friendly and doesn't work when running a crystal file as macro because the default target triple is used for compiling it.

**TL;DR** The target triple should default to the system the compiler is running on. Because the same binary runs on linux-gnu and linux-musl it needs to detect which one to use.

asterite:
> TL;DR The target triple should default to the system the compiler is running on

Yes, it is. We just ask this from LLVM via the C API, which should be equivalent of running `llvm-config --host-target` on the command line.

If you run that on that machine, what's the output?

RX14:
It doesn't.

straight-shoota:
@asterite I literally replaced the one occurence of `x86_64-unknown-linux-gnu` by `x86_64-alpine-linux-musl` in the compiler binary and it works as expected. So it's *definitely* baked in.

RX14:
I suggest defining a host triple in `Crystal::Config`, which is always the `--target` the compiler is built with. This is different from the `default_target_triple`, which should be `{{env("CRYSTAL_CONFIG_TARGET")}} || host_triple`. I don't think we will ever need to define the `--host` at runtime, we can leave it out until we get a usecase.

straight-shoota:
The issue is that "`--target` the compiler is built with" can differ from the host it is running on (e.g. gnu vs. musl).

I think it shouldn't matter what target triple the compiler was built with, only what it's running on. This needs to read from the system environment like `llvm-config --host-target`.

straight-shoota:
Having the host baked into the binary is usually not an issue, because most targets won't run on any other system anyway.
[...]

Issue #6903, opened 2018-10-03
Issues with Platform Support wiki page


straight-shoota:
The Wiki page [Platform Support](https://github.com/crystal-lang/crystal/wiki/Platform-Support) was created a year ago and hasn't seen much change since then. It's still considered a draft. That doesn't need to be changed, but I noticed a few issues, mostly inconsistencies in target names between wiki page and `src/lib_c`.

* ✅ `aarch64-linux-musl` was added in #5861. I suppose it should be listed in Tier 2.
* ✅ The wiki page lists `x86_64-unknown-openbsd` but that target is not available. There is a folder called `amd64-unknown-openbsd` which seems like the same target. LLVM doesn't know a `amd64` architecture <ins>on my system</ins> (it's `x68-64`). I suppose the target should be renamed in the source code unless there is some explicit reason for this being `amd64`? 
* ✅ The wiki page lists `x86_64-apple-darwin`, but the source folder is called `x86_64-macosx-darwin`
* ✅ The wiki page lists `armv6-linux-gnueabihf` and `armv7-linux-gnueabihf` but in the source there is only `arm-linux-gnueabihf`.
* ✅ `x86_64-portbld-freebsd` is missing from the wiki page (it's the same as `x86_64-unknown-freebsd`, though).
* ✅ The wiki page lists `x86_64-pc-windows-msvc` but the source folder is called `x86_64-windows-msvc`.

<del>Additionally, since musl is the only way to reliably compile with statically linked libraries, I'd suggest to improve its support level. Mainly, automatic tests would be great.
This shouldn't be very difficult to integrate into CI. Both travis and circle support docker images and the specs can just be executed in an alpine image for `i686-linux-musl` and `x86_64-linux-musl`.</del> #6943

For all other targets (at least on Tier 2), there could be a simple smoke test to ensure that the stdlib specs at least compile for that target. It would probably be enough to run this prior to a release (or on nightly), not on every PR.

```bash
for target in $(ls -1 src/lib_c/); do
  echo "Cross compiling for $target"
  make .build/std_spec clean FLAGS="--cross-compile --target $target"
done
```

BTW: I can't edit wiki any pages anymore. Has there been a change to access rights?

/cc @mverzilli @RX14 @bcardiff

mverzilli:
[...]

PR #7282, opened 2019-01-06
Add Crystal::Codegen::Target and consolidate triple handling

RX14:
This has been bugging me for a while, this fixes the way crystal handles target triples to be more correct. 

This means that the names of some `lib_c` folders has changed, symlinks are temporarily in place to let the old compiler version work. There will need to be a later PR to clean up and finish the second (far smaller) part of these changes.


RX14:
This isn't strictly required for https://github.com/crystal-lang/crystal/issues/7196, but it's something I wanted to do while I was poking around the target handling code.

asterite:
@RX14 Looks good! But could you describe what this change is about? What's the problem with the current code and how does this improve it?

straight-shoota:
See #6903 for some of the issues this should be fixing.

src/compiler/crystal/codegen/target.cr
```diff
@@ -0,0 +1,135 @@
...
+      raise Target::Error.new("Unsupported architecture for target triple: #{self}")
+    end
+
+    opt_level = release ? LLVM::CodeGenOptLevel::Aggressive : LLVM::CodeGenOptLevel::None
+
+    target = LLVM::Target.from_triple(self.to_s)
+    target.create_target_machine(self.to_s, cpu: cpu, features: features, opt_level: opt_level).not_nil!
+  end
+
+  def to_s
```
asterite: Shouldn't `to_s(io)` be defined instead?

src/compiler/crystal/compiler.cr
```diff
@@ -86,9 +86,9 @@ module Crystal
     # A `ProgressTracker` object which tracks compilation progress.
     property progress_tracker = ProgressTracker.new
 
-    # Target triple to use in the compilation.
+    # Codegen target to use in the compilation.
     # If not set, asks LLVM the default one for the current machine.
-    property target_triple : String?
+    property codegen_target = Codegen::Target.new
```
asterite: Could this be made a lazy property?

RX14:
@asterite the main thing it introduces is the `Codegen::Target` class, which uses LLVM's normalization code to turn mangled human-inputted triples into something which always has the `architecture-vendor-environment` format. So the compiler now accepts more triples. Then `Codegen::Target` provides a bunch of query methods and ways to reason about the triple so that the correct flags can be set more robustly. And finally the format in `lib_c` is updated to always be `architecture-environment`, which was the case for architectures where the environment happened to have multiple parts (like `linux-gnu` or `linux-musl`), but the compiler didn't normalize `x86_64-unknown-freebsd11.0` to `x86_64-freebsd`, it left it as `x86_64-unknown-freebsd`, which includes the vendor still (unknown).

So it's mostly cleanups for minor things which worked well already, but wasn't perfect.

src/compiler/crystal/compiler.cr
asterite: Because the comments say it will be the default one unless you set one. From that I expect it to be lazily created (unless you set one). But it's fine if that was not the intention.

RX14:
@asterite fixed!

straight-shoota:
@RX14 Could you also take a look at updating the Platform Support wiki page (cf #6903).

PR #7291, opened 2019-01-09
Get SNI for OpenSSL

bararchy:
This PR adds the ability to query the [SNI](https://www.openssl.org/docs/man1.1.1/man3/SSL_get_servername.html) string, having this ability is important when needing to decide which certificate to show the connecting client, or, when creating a proxy that should show the relevant certificate for each host.
There are other reasons for having this ability.

Overall this change is minor and have been tested locally.

### DIFF
--- a/spec/std/openssl/ssl/server_spec.cr
+++ b/spec/std/openssl/ssl/server_spec.cr
@@ -76,4 +76,21 @@ describe OpenSSL::SSL::Server do
       end
     end
   end
+
+  it "detects SNI hostname" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open tcp_server, server_context do |server|
+      spawn do
+        client = server.accept
+        client.hostname.should eq("example.com")
+        client.close
+      end
+
+      OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context) do |socket|
+        socket.hostname = "example.com"
+      end
+    end
+  end
 end
--- a/spec/std/openssl/ssl/server_spec.cr
+++ b/spec/std/openssl/ssl/server_spec.cr
@@ -83,13 +83,14 @@ describe OpenSSL::SSL::Server do
 
     OpenSSL::SSL::Server.open tcp_server, server_context do |server|
       spawn do
-        client = server.accept
-        client.hostname.should eq("example.com")
-        client.close
+        sleep 1
+        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
+        end
       end
 
-      OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
-      end
+      client = server.accept
+      client.hostname.should eq("example.com")
+      client.close
     end
   end
 end
--- a/spec/std/openssl/ssl/server_spec.cr
+++ b/spec/std/openssl/ssl/server_spec.cr
@@ -88,8 +88,7 @@ describe OpenSSL::SSL::Server do
         client.close
       end
 
-      OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context) do |socket|
-        socket.hostname = "example.com"
+      OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
       end
     end
   end
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -73,7 +73,7 @@ class OpenSSL::SSL::Server
   end
 
   # This method allows to fetch the SSL SNI string if present
-  def get_sni : String
+  def sni : String
     String.new(LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name))
   end
 
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -73,9 +73,9 @@ class OpenSSL::SSL::Server
   end
 
   # This method allows to fetch the SSL SNI string if present
-  def sni : String?
-    if hostname = LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name)
-      String.new(hostname)
+  def hostname : String?
+    if host_name = LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name)
+      String.new(host_name)
     end
   end
 
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -73,8 +73,9 @@ class OpenSSL::SSL::Server
   end
 
   # This method allows to fetch the SSL SNI string if present
-  def sni : String
-    String.new(LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name))
+  def sni : String?
+    host_name = String.new(LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name))
+    host_name.empty? ? nil : host_name
   end
 
   # Closes this SSL server.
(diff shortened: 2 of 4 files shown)


src/openssl/ssl/server.cr
```diff
@@ -72,6 +72,11 @@ class OpenSSL::SSL::Server
     end
   end
 
+  # This method allows to fetch the SSL SNI string if present
```
r00ster91: `if present`? Does that mean this value is sometimes not available? What gets returned in that case? An empty string? Maybe that should be `nil`.
Also I think this method should be simply called `sni` because as far as I know in Crystal's method world you don't put `set` or `get` in front of method names.

bararchy: @r00ster91 done!

asterite:
Looks good, though in Ruby I think the method is called `hostname` and there's also `hostname=`. I know nothing about openssl, but I just found that.

asterite:
(note: I might be wrong)

bararchy:
@asterite the `hostname=` should be in the `Client` part of `SSL::Socket` as it sets the SNI for the connection.
I wanted to divide those into small manageable PRs because C bindings + OpenSSL is a scary concept and it's better to take in small doses

src/openssl/ssl/server.cr
```diff
@@ -72,6 +72,12 @@ class OpenSSL::SSL::Server
...
+  def sni : String?
+    host_name = String.new(LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name))
+    host_name.empty? ? nil : host_name
+  end
```
ysbaddaden: It should check for a NULL pointer instead of instantiating a String then checking it's size:

```crystal
if hostname = LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name)
  String.new(hostname)
end
```

bararchy: @ysbaddaden done!
keep in mind that `hostname=` will be part of the Client socket (as I answered to @asterite)

src/openssl/ssl/server.cr
```diff
@@ -72,6 +72,13 @@ class OpenSSL::SSL::Server
     end
   end
 
+  # This method allows to fetch the SSL SNI string if present
+  def hostname : String?
+    if host_name = LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name)
```
straight-shoota: Where is `@ssl` declared? And shouldn't hostname be a property of the SSL server socket? The point of this is that you have *one*  server wich can accepts connections with different hostnames.

bararchy: @straight-shoota -> https://github.com/crystal-lang/crystal/blob/530b9134582f9c97feea381c23d8e66037e72e8d/src/openssl/ssl/socket.cr#L82

straight-shoota: This is `OpenSSL::SSL::Server`, not `OpenSSL::SSL::Socket::Server`. Did you mix up the classes?

straight-shoota: These names *are* exceptionally confusing. That needs to improve with the new generic TLS API.

bararchy: Oh! damn yes I did, fixing now! damn...

bararchy: @straight-shoota fixed, sorry

bararchy:
@straight-shoota regarding specs, sadly can't make those until the `client` side will be finished, unless you wish me to do an ugly hack like 
```
system("openssl s_client")
```
etc..

jhass:
For which openssl versions are these methods available? Do we need or want to conditionalize defining them for nicer compilation errors? 

Btw, maybe I'm wrong, but I vaguely remember that you can bundle up your certs and keys into a single file, load that into a context and openssl will pick the first matching one.

bararchy:
@asterite @jhass @straight-shoota @RX14 

I totally missed we already have the `hostname` ability in the client socket already, so, Added only server side, added specs which should pass.

Let's squash and merge, then I'll start working on adding the callback which is the needed part for the whole SNI server flow

bararchy:
@asterite @ysbaddaden @straight-shoota can I get another approval ?

src/openssl/ssl/socket.cr
```diff
@@ -181,4 +181,11 @@ abstract class OpenSSL::SSL::Socket < IO
   def unbuffered_rewind
     raise IO::Error.new("Can't rewind OpenSSL::SSL::Socket::Client")
   end
+
+  # This method allows to fetch the SSL SNI string if present
```
asterite: Is this description correct? It's a bit weird that the method name is `hostname` but here it says "SSL SNI string".

bararchy: SNI or Server Name Indication is the used name for a hostname being passed inside the OpenSSL protocol.

I think saying `SNI` here makes more sense then hostname because we are already in the context of a Socket and I don't want mixups, if someone thinks this should change let me know

jhass: Let's strip the "SSL", though. It's redundant in this context, let alone SSL as a protocol is pretty dead by now (everybody is using TLS). 

We could also say something like "Returns the hostname provided through Server Name Indication (SNI)"

ysbaddaden: I like @jhass proposal, SSL is redundant and SNI may not be a known acronym.

bararchy:
@jhass @ysbaddaden done. Updated docs comment 
Feel free to merge or let the CI ran for 50min for a comment hahahah

bararchy:
@ysbaddaden all green LGTM :tada:

spec/std/openssl/ssl/server_spec.cr
```diff
@@ -76,4 +76,20 @@ describe OpenSSL::SSL::Server do
...
   end
+
+  it "detects SNI hostname" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open tcp_server, server_context do |server|
+      spawn do
+        client = server.accept
+        client.hostname.should eq("example.com")
```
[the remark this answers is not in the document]
bararchy: :cry: this process of PRing is really frustrating , I can see why people push something and disappear after 3 review cycles where there is always this one last thing to fix.
Anyway, I'll make the change, could we just try and make sure there is nothing else to change on the same time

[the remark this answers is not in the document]
bararchy: @straight-shoota done.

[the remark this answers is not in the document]
straight-shoota: Well, the point of reviewing is to make sure the patched code is fine. Sometimes, issues are not directly spotted on the first view.

spec/std/openssl/ssl/server_spec.cr
```diff
@@ -76,4 +76,21 @@ describe OpenSSL::SSL::Server do
...
+
+  it "detects SNI hostname" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open tcp_server, server_context do |server|
+      spawn do
+        sleep 1
+        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
+        end
```
Sija: one `end` too much.

spec/std/openssl/ssl/server_spec.cr
```diff
@@ -76,4 +76,21 @@ describe OpenSSL::SSL::Server do
...
   end
+
+  it "detects SNI hostname" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open tcp_server, server_context do |server|
+      spawn do
+        sleep 1
+        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
```
Sija: Could you please split this line? Its complexity makes it hard to read.

spec/std/openssl/ssl/server_spec.cr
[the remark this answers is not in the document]
bararchy: Oh I'm well aware, nothing against you or the reviewers , just pointed out the frustration :shrug: 
anyway, we are on the path to merge so all is good

spec/std/openssl/ssl/server_spec.cr
bararchy: I can, keep in mind its a copy paste of the above tests, I just wanted to keep the same structure, maybe it should be better to merge this PR and make a new one fixing all the examples in the file which use this exact same line?

bararchy:
@straight-shoota can I get your approval?

PR #7295, opened 2019-01-10
Class: let `to_s` be independent of the `name` method

### LINKED ISSUE #7292: Class: redefining the `name` property affects the behavior of `to_s`
```Crystal
abstract class Command
  class_getter name : String = ""
end

class CreateUserCommand < Command
end


pp {{Command.subclasses}} # =. []
```

However removing the `class_getter name : String = ""` causes it to work fine.

https://play.crystal-lang.org/#/r/5ypk/edit

asterite:
Fixes #7292

### DIFF
--- a/spec/std/class_spec.cr
+++ b/spec/std/class_spec.cr
@@ -12,6 +12,12 @@ end
 private class B2 < A
 end
 
+private class ClassWithRedefinedName
+  def self.name
+    "OtherName"
+  end
+end
+
 describe Class do
   it "does ===" do
     (Int32 === 1).should be_true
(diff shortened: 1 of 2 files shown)


straight-shoota:
Can we also remove setting the `name` class property by default? It's weird to have this automatically defined but overidable.

straight-shoota:
Yeah, that's true. But the issue is that you can't rely on `name` because it could be easily overridden, even by accident. That's less likely with `.to_s`. But I'd even consider disallowing to override `.to_s`. I fail to see any valid use case were you would want to conciously let a class report an incorrect name. This will only lead to unexpected behaviour.

Sija:
That's theoretically true, yet I've encountered maybe 2 cases in which it ended up as an issue really (and with this PR merged they're resolved). I'm against creating rules for no actual reason. There's much more that could go wrong within the stdlib itself because of the open nature of the language (like overriding `Object#id` or other important methods) and it's up to the developer to be aware of those (like in Ruby).

PR #7296, opened 2019-01-10
Set certificate_chain from X509 object

bararchy:


bararchy:
This fails for me too with `SSL_CTX_use_certificate: error:0609E09C:digital envelope routines:PKEY_SET_TYPE:unsupported algorithm (OpenSSL::Error)` But looking at the docs I don't see why.

@RX14 @jhass @asterite @straight-shoota any idea? 

docs at: https://www.openssl.org/docs/man1.0.2/ssl/SSL_CTX_use_certificate.html

bararchy:
@asterite Ruby has a way to add a `Certificate` object directly to context, yes.
https://docs.ruby-lang.org/en/2.6.0/OpenSSL/SSL/SSLContext.html#method-i-add_certificate

(as in, not only from file)

bararchy:
@asterite seems like it works in Ruby

```ruby
require "openssl"
=> true
cert = OpenSSL::X509::Certificate.new
=> #<OpenSSL::X509::Certificate
 subject=#<OpenSSL::X509::Name >,
 issuer=#<OpenSSL::X509::Name >,
 serial=#<OpenSSL::BN 0>,
 not_before=nil,
 not_after=nil>
context = OpenSSL::SSL::SSLContext.new
=> #<OpenSSL::SSL::SSLContext:0x00005587e589ef00>
context.cert = cert
=> #<OpenSSL::X509::Certificate
 subject=#<OpenSSL::X509::Name >,
 issuer=#<OpenSSL::X509::Name >,
 serial=#<OpenSSL::BN 0>,
 not_before=nil,
 not_after=nil>

```

ysbaddaden:
It's possible that we don't initialize the X509 certificate correctly. What if you extract one from the cert file, then try to set it?

bararchy:
@ysbaddaden we currently have no way to initiate a new X509 object from cert file, only setup SSL::Context from it.

HertzDevil:
#13187 added an `OpenSSL::X509::Certificate.from_der?` method that might be usable as a constructor

PR #7297, opened 2019-01-11
Detect nesting `it` and `pending` at run-time

### LINKED ISSUE #7192: pending test causes incorrect test count and malformed output

MakeNowJust:
Fix #7192
Close #7207

### DIFF
--- a/src/spec/context.cr
+++ b/src/spec/context.cr
@@ -169,10 +169,6 @@ module Spec
 
     @@spec_nesting = false
 
-    def self.check_nesting_spec(file, line)
-      check_nesting_spec(file, line) { }
-    end
-
     def self.check_nesting_spec(file, line, &block)
       raise NestingSpecError.new("can't nest `it` or `pending`", file, line) if @@spec_nesting
 
--- a/src/spec/context.cr
+++ b/src/spec/context.cr
@@ -174,7 +174,7 @@ module Spec
     end
 
     def self.check_nesting_spec(file, line, &block)
-      raise NestingSpecError.new("cannot nest `it` and `pending`", file, line) if @@spec_nesting
+      raise NestingSpecError.new("can't nest `it` or `pending`", file, line) if @@spec_nesting
 
       @@spec_nesting = true
       begin
--- a/src/spec/context.cr
+++ b/src/spec/context.cr
@@ -81,7 +81,7 @@ module Spec
             puts
             puts "#{(i + 1).to_s.rjust(3, ' ')}) #{fail.description}"
 
-            if ex.is_a?(AssertionFailed)
+            if ex.is_a?(SpecError)
               source_line = Spec.read_line(ex.file, ex.line)
               if source_line
                 puts Spec.color("     Failure/Error: #{source_line.strip}", :error)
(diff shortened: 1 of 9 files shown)


src/spec/context.cr
```diff
@@ -166,6 +166,23 @@ module Spec
...
     end
+
+    @@spec_nesting = false
+
+    def self.check_nesting_spec(file, line)
+      check_nesting_spec(file, line) { }
+    end
+
+    def self.check_nesting_spec(file, line, &block)
+      raise NestingSpecError.new("cannot nest `it` and `pending`", file, line) if @@spec_nesting
```
Sija: `and` -> `or`

MakeNowJust: Thank you!

src/spec/expectations.cr
```diff
@@ -385,6 +385,11 @@ module Spec
         raise ex
       end
 
+      # `NestingSpecError` is also treated.
+      if ex.is_a?(Spec::NestingSpecError) && klass != Spec::NestingSpecError
```
asterite: What is this for?

MakeNowJust:
Please don't merge this for now because this doesn't fix #7192 entirely. Sorry 🙇 I'll fix it asap.

@straight-shoota At least we have an issue #7192 and this is the simplest way to fix it.

MakeNowJust:
Now, nesting `pending`/`it` doesn't break `crystal spec -v` output.

Example:

```crystal
require "spec"

it "foo" do
  it "bar" do
  end
end

it "baz" do
  pending
end
```

Old (broken) output:



New (fixed) output:

src/spec/expectations.cr
MakeNowJust: See full code:

```crystal
    def expect_raises(klass : T.class, message = nil, file = __FILE__, line = __LINE__) forall T
      yield
    rescue ex : T
      # We usually bubble Spec::AssertaionFailed, unless this is the expected exception
      if ex.is_a?(Spec::AssertionFailed) && klass != Spec::AssertionFailed
        raise ex
      end

      # `NestingSpecError` is treated as the same above.
      if ex.is_a?(Spec::NestingSpecError) && klass != Spec::NestingSpecError
        raise ex
      end

      # ...
    end
```

Please tell me more correct sentence which means this if you know.

src/spec/context.cr
```diff
@@ -166,6 +166,19 @@ module Spec
     def matches?(pattern, line, locations)
       false
     end
+
+    @@spec_nesting = false
```
Fryguy: If we were to ever get to parallel test execution, would a class variable cause a problem?

asterite: Yes

asterite: Oh, actually, I was replying generically that class variables are to be used carefully in a parallel context. Not sure about this case, though

Issue #5784, opened 2018-03-07
Expose connection metadata in HTTP::Server::Context


straight-shoota:
With #5776 upcoming, `HTTP::Server` can listen on multiple networks sockets - and even without that, it would be useful for a handler to be able to access information about the network connection used to process.

My initial idea was to expose the socket as `Context#socket` but that would also allow to take over the IO which should be avoided (there's `Response#upgrade` for that).

There really just needs to be access to metadata about the socket, for example `local_address` and `remote_address` (note that `remote_address` is not particularly useful to identify the actual client's IP because of proxies and whatnot).
I don't know if there would be any other useful data.

We could think about encapsulating the values in a struct like `Socket::Metadata` but I'm not sure it's worth adding another type for just two data points.

A question is whether to expose it: on `Context` or `Response`. I'd say the most logical place would be `Context` as it relates to both request and response. On the other hand, other network related features (such as the mentioned `#upgrade` as well as `#close`) are on `Response`. 

Another issue is how to implement it. It could either be just delegates to the `Socket` instance. this would be easy in `Reponse` because the network socket is already available internally as `@io`. For `Context` it would either need access to `Response`'s internal io or store a reference to the socket itself. A different solution would be to just copy the values to the `Context` (or `Response`) instance.

And the third question I'm not sure about, how this should behave when no network connection is involved (for example in specs that don't go through the network stack). Should it just return `nil` or a default value like the loopback address (but what would that be for Unix sockets?).

straight-shoota:
I've looked at some other API's and I found a nice addition in Java's `ServletRequest`: It has methods named `getServerName()` and `getServerPort()` which expose the values from the `Host` header if available and default to the local address.
Go's `http.Request` similarly has a `Host` property which is either the value of the `Host` header or the host included in the request URI. It may or may not include a port.

RX14:
It should conceptually be on `Request` instead of `Response`, `Context` should continue holding just a request and response.

straight-shoota:
@RX14 Having it on `Request` would mean, these methods were also exposed in a client context, because this class is used as representing a request both in the client and server. A client request can't have any network connection data unless it is modified when it is being sent over a network connection. And it would still probably need different implementations for client and server context.

There are separate `Response` classes for client and server, which makes it far more easily to incorporate access to connection data into the server response. Conceptually, this is not great, I agree.
But I don't think we can implement this with a sane API unless we had a separate server `Request` type.

RX14:
[...]

PR #1722, opened 2015-10-10
HTTP::Request, added remote ip and peer_addr

benoist:

PR #5870, opened 2018-03-26
Add connection metadata to HTTP::Server::Response

straight-shoota:
This PR adds methods `#local_address` and `#remote_address` address to `HTTP::Server::Response`.

I'm not entirely sure whether the method's return type should be nilable. It's not very typical to have a Response not wrapping a socket. This is mostly used as mockup for specs. Maybe we can avoid forcing usually unnecessary nil-checks in actual code. This could be done either by raising (though there should probably also be a non-raising option) or returning some default address if the IO does not provide one.

Closes #5784
Based on #5869

PR #7302, opened 2019-01-11
[WIP] Add remote_address to HTTP::Server::Context

### LINKED ISSUE #453: Add remote ip field to Http::Request
It is useful to get access from `HTTP::Handler` to underlying socket properties. I am interested in the client ip address (aka 'remote_ip'). Currently `Http::Request` has fields that are parsed out of the request body but no client ip.

omarroth:
Would appear to close #453 and #5784.

This PR adds support for `remote_address` by pulling it from the underlying IO.

I'm currently unsure if adding this as a separate field to `Context` is appropriate. It may make more sense to provide a `Context::Info` or `Context::Metadata`, which could contain `remote_address` and `local_address`, or other useful information that would be relevant in some cases, with functionality similar to #1722.

Example use:
```crystal
# A very basic HTTP server
require "http/server"

  context.response.print "Hello to #{context.remote_address}!"
end

server.listen
```

```
$ curl http://127.0.0.1:8080
Hello to 127.0.0.1:37024!
```

This PR also patches `OpenSSL::SSL::Socket` so the `IPAddress` can be determined for servers with SSL enabled.

It might be good to change [this line](https://github.com/crystal-lang/crystal/compare/master...omarroth:add-remote-ip?expand=1#diff-157d96b494d2033fd2ed0f4d8d12a856R46) to a warning or safer default.

Still needs tests.

### DIFF
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -19,8 +19,9 @@ class HTTP::Request
   @cookies : Cookies?
   @query_params : Params?
   @uri : URI?
+  @io : IO?
 
-  def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1")
+  def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1", @io : IO? = nil)
     @headers = headers.try(&.dup) || Headers.new
     self.body = body
   end
@@ -99,11 +100,25 @@ class HTTP::Request
     return BadRequest.new unless HTTP::SUPPORTED_VERSIONS.includes?(http_version)
 
     HTTP.parse_headers_and_body(io) do |headers, body|
-      return new method, resource, headers, body, http_version
+      return new method, resource, headers, body, http_version, io
     end
 
     # Malformed or unexpectedly ended http request
     BadRequest.new
+  end
+
+  def remote_address
+    if (io = @io).responds_to?(:remote_address)
+      io.remote_address
+    else
+      raise "#{io} doesn't have a remote address"
+    end
+  end
+
+  def remote_address?
+    if (io = @io).responds_to?(:remote_address)
+      io.remote_address
+    end
   end
 
   # Lazily parses and return the request's path component.
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -173,6 +173,20 @@ abstract class OpenSSL::SSL::Socket < IO
     raise IO::Error.new("Can't rewind OpenSSL::SSL::Socket::Client")
   end
 
+  def remote_address
+    if (io = @bio.io).responds_to?(:remote_address)
+      io.remote_address
+    else
+      raise "#{io} doesn't have a remote address"
+    end
+  end
+
+  def remote_address?
+    if (io = @bio.io).responds_to?(:remote_address)
+      io.remote_address
+    end
+  end
+
   # Returns the hostname provided through Server Name Indication (SNI)
   def hostname : String?
     if host_name = LibSSL.ssl_get_servername(@ssl, LibSSL::TLSExt::NAMETYPE_host_name)


sam0x17:
this is so needed

straight-shoota:
IMHO this PR is premature. We should continue the discussion about how to expose and implement this in #5784.

omarroth:
I agree it's premature, that's why it's marked as [WIP].

[This](https://github.com/crystal-lang/crystal/issues/453#issuecomment-333295282) comment #453 is mostly the reason for this PR. What is implemented here is minimal functionality with the expectation that it will be expanded to include support for e.g. `X-Forwarded-For` similar to #1722.

My hope for this PR is to further discussion. There doesn't appear to be much consensus reached in #5784, although I would be happy to move this functionality into `HTTP::Server::Response` if that is more appropriate.

omarroth:
A couple thoughts:

In #1722 there's an option to use headers to determine the IP, namely `X-Forwarded-For`, `Client-Ip`, and several others were mentioned. Would it make sense to leave this to the user to implement? There appear to be a pretty varied list of headers in use that would likely need to be supported, while the user would have a better understanding of what they need.

[This example](https://github.com/jhass/carc.in/blob/master/src/web.cr#L96-L102) exposes `client_ip`, which *is* determined from headers, so it might make sense to implement that here. So for example, if a client is connecting to a server via proxy, `client_ip` would return the result of `X-Forwarded-For`, and `remote_ip` would return the proxy.

It makes sense to me that this should belong in `HTTP::{Server,Client}::Response`, since there are already several other methods for handling the socket itself, namely `close`, `flush`, and `write`. I would consider this PR a way of providing information about the socket. It is also keeps `HTTP::Server::Context` clean. I'd be interested in other thoughts.

Currently this PR supports HTTP servers, with and without SSL. Are there any other protocols/sockets that need to be supported?

src/http/server/context.cr
```diff
@@ -7,8 +7,11 @@ class HTTP::Server
     # The `HTTP::Server::Response` to configure and write to.
     getter response : Response
 
+    # The `Socket::IPAddress` of the remote
+    getter remote_address : Socket::IPAddress
+
     # :nodoc:
-    def initialize(@request : Request, @response : Response)
+    def initialize(@request : Request, @response : Response, @remote_address : Socket::IPAddress = Socket::IPAddress.new("127.0.0.1", 3000))
```
ysbaddaden: There should be no default. It's misleading.

Furthermore, it's not necessarily an IP address, it can be an UNIX address.

src/http/server/request_processor.cr
```diff
@@ -30,10 +31,25 @@ class HTTP::Server::RequestProcessor
...
+          io = input.io
+
+          case io
+          when TCPSocket
+            remote_address = io.remote_address
+          end
+        end
+
+        # TODO: Check default for remote_address
+        remote_address ||= Socket::IPAddress.new("127.0.0.1", 3000)
```
ysbaddaden: Again: there must be no default.

src/openssl/ssl/socket.cr
```diff
@@ -95,6 +95,10 @@ abstract class OpenSSL::SSL::Socket < IO
     LibSSL.ssl_set_bio(@ssl, @bio, @bio)
   end
 
+  def io
```
ysbaddaden: Maybe there could be a delegate to `io.remote_address` instead, so we don't care whether `io` is a Socket or SSLsocket.

straight-shoota:
@omarroth 

> In #1722 there's an option to use headers to determine the IP, namely `X-Forwarded-For`, `Client-Ip`, and several others were mentioned. Would it make sense to leave this to the user to implement?

As you state in the next paragraph, these are two different concepts: a) The remote address of the immediate network connection. b) The client address of a HTTP connection which may be equal to the remote address, but that could also be some kind of proxy or other intermediary.

I think both should be made available in stdlib's HTTP implementation, but this PR focuses on a) which needs to come first. b) is the next step.

> Currently this PR supports HTTP servers, with and without SSL. Are there any other protocols/sockets that need to be supported?

The server is specific to HTTP, so no other application servers need to be supported. But obviously all kinds of HTTP transport streams, i.e. TCP and Unix sockets with and without SSL.

src/openssl/ssl/socket.cr
omarroth: I think it makes sense that it is exposed here, since we're "unwrapping" the SSLSocket.

If you were to delegate `remote_address` to SSLSocket there would likely be duplication in handling various sockets.

src/http/server/response.cr
```diff
@@ -40,6 +40,30 @@ class HTTP::Server
...
     end
 
+    def remote_address
+      case @io
+      when IPSocket
+        io = @io
+      when UNIXSocket
+        io = @io
+      when OpenSSL::SSL::Socket
+        io = @io.as(OpenSSL::SSL::Socket).io
```
Sija: No need for calling `.as`, ditto below.

src/http/server/response.cr
```diff
@@ -40,6 +40,30 @@ class HTTP::Server
       output.response = self
     end
 
+    def remote_address
+      case @io
+      when IPSocket
+        io = @io
+      when UNIXSocket
```
Sija: You can merge it with the above branch using `when IPSocket, UNIXSocket`. ditto below.

src/http/server/response.cr
```diff
@@ -40,6 +40,30 @@ class HTTP::Server
...
+      case io
+      when IPSocket
+        remote_address = io.as(IPSocket).remote_address
+      when UNIXSocket
+        remote_address = io.as(UNIXSocket).remote_address
+      else
+        raise "Unsupported socket"
+      end
+
+      return remote_address
```
Sija: Redundant `return`. ditto below.

src/http/server/response.cr
omarroth: Unfortunately there appears to be an issue with type inference, removing the cast results in:
```
in src/http/server/response.cr:50: undefined method 'io' for IO::ARGF (compile-time type is IO+)

        io = @io.io
```

omarroth:
Thanks for taking a look at this @ysbaddaden and @Sija.

I moved `remote_address` into `HTTP::Server:Response`, and it now will raise an exception on unsupported socket types.

Currently it's handling the socket twice. Once to unwrap it if necessary and then again to determine the underlying type. It may be necessary to delegate it to `SSLSocket` if it's possible for a socket to be wrapped multiple times, but otherwise it seems cleaner to handle that in one place.

src/http/server/response.cr
```diff
@@ -40,6 +40,26 @@ class HTTP::Server
...
+      end
+
+      case io
+      when IPSocket, UNIXSocket
+        remote_address = io.remote_address
+      else
+        raise "Unsupported socket"
+      end
+
+      remote_address
```
Sija: Actually this variable can be removed too only leaving `case`.

src/openssl/ssl/socket.cr
```diff
@@ -95,6 +95,10 @@ abstract class OpenSSL::SSL::Socket < IO
...
+    return @bio.io
```
Sija: ditto

src/http/server/response.cr
omarroth: Would it still be worth leaving for readability?

Sija: I'm not sure, it (readability) is pretty subjective but perhaps... anyone else would like to chime in?

src/http/server/response.cr
```diff
@@ -40,6 +40,26 @@ class HTTP::Server
       output.response = self
     end
 
+    def remote_address
+      case @io
```
straight-shoota: Make this `case io = @io`. Using the local variable `io` in each branch will have it restricted to the appropriate type.

straight-shoota: This entire method could be much simpler:

```cr
io = @io
if io.is_a?(OpenSSL::SSLSocket)
  io = io.@bio.io
end

if io.responds_to?(:remote_address)
  io.remote_address
end
```

It's probably fine to just return `nil` in case `remote_address` is not available. Or we could have both a raising and nilable variant.

straight-shoota: Maybe it would even be better to delegate `remote_address` on `SSLSocket`.

omarroth: So a `remote_address?` and `remote_address`?

src/http/server/request_processor.cr
```diff
@@ -1,4 +1,5 @@
 require "../server"
+require "openssl"
```
straight-shoota: This require seems off-placed.

src/http/server/response.cr
omarroth: The reason for `case` is for extending it to support any other wrappers for socket, although I think SSLSocket is the only one so I don't mind changing it.

straight-shoota: Yes. But I'm not sure if we need the raising variant.

ysbaddaden: We want the raising variant. The underlying IO for a HTTP server should always be a Socket, so we don't have to deal with a nilable. Just like `array[index]` is expected to return a value, `remote_address` is expected to return an address.

src/openssl/ssl/socket.cr
ysbaddaden: I don't understand your claim. We musn't care if it's a SSLSocket just like we don't care whether is a UnixSocket or TCPSocket or whatever other IO. If `io` responds to `#remote_address` then it returns an address. Nothing more.

omarroth:
Where would you suggest exposing this?

ysbaddaden:
The request object. Later we can implement `#client_ip` on the request, that will use the remote address and the x-forwarded request headers to return a client ip.

straight-shoota:
That was essentially the consensus in #5784 and why previous attempts like #5870 failed. It just didn't work out yet because `HTTP::Request` is used in both the client and server and we need to agree whether we want to split it into two separate classes (`HTTP::{Client,Server}}::Request`) like it is already the case for `Response`.

omarroth:
Apologies for not giving this PR as much attention. I moved `remote_address` into `HTTP::Request` and made a minor change to `HTTP::Request` to allow for pulling it from the underlying IO.

I'll add specs when I have a free moment.

src/http/request.cr
```diff
@@ -19,8 +19,9 @@ class HTTP::Request
   @cookies : Cookies?
   @query_params : Params?
   @uri : URI?
+  @io : IO?
 
-  def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1")
+  def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1", @io : IO | Nil = nil)
```
samueleaton: `IO | Nil` can be changed to `IO?` to match `Headers?`

omarroth:
Closing in favor of #7610.

PR #7311, opened 2019-01-13
remove debian 32 bit from Vagrantfile

wontruefree:
Debian seems to only support 64 bit operating systems on vagrant now as listed here https://app.vagrantup.com/debian.
This was breaking vagrant up so I removed the 32 bit instances.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,8 +26,8 @@ Vagrant.configure("2") do |config|
     end
   end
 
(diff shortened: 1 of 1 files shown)


bcardiff:
This is actually handled in #7066 . Feel free to provide additional feedback on the Vagrantfile in that PR if it's a resource you are using. Thanks!

PR #7313, opened 2019-01-14
Refactor getter/property macros to work properly with Bool types

### LINKED ISSUE #7312: Lazy getter macros doesn't work with Bool
Since `||=` operator invokes rhs on `nil` **and** `false`, using it inside `getter` macro with `Bool` type will execute lazy block whenever the value is `false` - which I'd say is incorrect behavior since lazy initialization should happen only for `nil` values.

```cr
class Foo
  getter(bar : Bool) { puts "Foo#bar executed"; false }
end

foo = Foo.new
foo.bar # prints "Foo#bar executed"
foo.bar # prints "Foo#bar executed" again
```

Should the implementation be changed to check for `.nil?` instead of using `||=`?

Sija:
Fixes #7312

### DIFF
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -35,6 +35,9 @@ private class TestObject
   def self.getter12_value=(@@getter12_value)
   end
 
+  @getter13_counter = 0
+  getter(getter13) { @getter13_counter += 1; false }
+
   setter setter1
   setter setter2 : Int32
   setter setter3 : Int32 = 3
@@ -57,6 +60,9 @@ private class TestObject
 
   property(property11) { 11 }
   property property12 : Int32 { 10 + 2 }
+
+  @property13_counter = 0
+  property(property13) { @property13_counter += 1; false }
 
   def initialize
     @getter1 = 1
@@ -205,6 +211,13 @@ describe Object do
       obj2 = TestObject.new
       obj2.getter12.should eq(24)
     end
+
+    it "defines lazy getter with block returning false" do
+      obj = TestObject.new
+      obj.@getter13_counter.should eq(0)
+      3.times { obj.getter13.should be_false }
+      obj.@getter13_counter.should eq(1)
+    end
   end
 
   describe "getter!" do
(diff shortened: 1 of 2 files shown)


src/object.cr
```diff
@@ -355,11 +359,19 @@ class Object
           {{var_prefix}}\{{name.var.id}} : \{{name.type}}?
 
           def {{method_prefix}}\{{name.var.id}}
-            {{var_prefix}}\{{name.var.id}} ||= \{{yield}}
+            if {{var_prefix}}\{{name.var.id}}.nil?
+              {{var_prefix}}\{{name.var.id}} = \{{yield}}
+            else
+              {{var_prefix}}\{{name.var.id}}.not_nil!
```
asterite: This (and all) `not_nil!` should be avoided because it will check again for nil, and also generate code to raise in case it's nil. It's better to assign the ivar to a local variable first.

spec/std/object_spec.cr
```diff
@@ -205,6 +211,14 @@ describe Object do
       obj2 = TestObject.new
       obj2.getter12.should eq(24)
     end
+
+    # TODO: enable after 0.27.1
```
asterite: This doesn't need to be pending

src/object.cr
Sija: Thanks for the tip! I forgot that I'm dealing here with with c/ivars and out the frustration tucked `not_nil!` all over the place ;)

In the updated code I named the local variable after the c/ivar name + `_` prefix to avoid clashing with language keywords, yet I'm thinking whether would be cleaner to use some generic name like `value` instead, what do you think?

bew: @Sija what about a fresh variable `%value` ?

Sija:
@bew I thought about it but since there's no danger of name collision or anything like that would be just wasteful.

bew:
then why not use `value` instead of `_\{{name.var.id}}` ?

Sija:
`value` it is then.

PR #7318, opened 2019-01-17
Benchmark improvements & fixes

r00ster91:
This:
* (https://github.com/crystal-lang/crystal/commit/16eb8bb58a051a53039db29cabe5a57fc6d0994d) Fixes that when doing this: `Benchmark.ips {}`, that this happens:
  ```
  Unhandled exception: Empty enumerable (Enumerable::EmptyError)
    from /usr/lib/crystal/benchmark/ips.cr:727:5 in 'run_comparison'
    from /usr/lib/crystal/benchmark/ips.cr:40:9 in 'execute'
    from /usr/lib/crystal/benchmark.cr:113:5 in '__crystal_main'
    from /usr/lib/crystal/crystal/main.cr:97:5 in 'main_user_code'
    from /usr/lib/crystal/crystal/main.cr:86:7 in 'main'
    from /usr/lib/crystal/crystal/main.cr:106:3 in 'main'
    from __libc_start_main
    from _start
    from ???
  ```
  Or for `Benchmark.bm {}`, this happens:
  ```
  ```
  That error can be quite confusing and the empty benchmark doesn't seem quite right to me either. Now instead, nothing happens and the benchmark will get skipped. And if you for whatever reason want to do macro stuff in a `Benchmark.ips`, you don't have to handle the case where the benchmark is empty.
* (https://github.com/crystal-lang/crystal/commit/16eb8bb58a051a53039db29cabe5a57fc6d0994d) Currently this:
  ```cr
  Benchmark.bm do |bm|
    bm.report
  end
  ```
  results in:
  ```
  in line 2: 'Benchmark::BM::Job#report' is expected to be invoked with a block, but no block was given
  ```
  which is of course the expected behaviour.
  But this?
  ```cr
  Benchmark.ips do |bm|
    bm.report
  end
  ```
  you will get this!
  ```
  Unhandled exception: Empty enumerable (Enumerable::EmptyError)
    from /usr/lib/crystal/benchmark/ips.cr:761:5 in 'report'
    from /eval:3:3 in '__crystal_main'
    from /usr/lib/crystal/crystal/main.cr:97:5 in 'main_user_code'
    from /usr/lib/crystal/crystal/main.cr:86:7 in 'main'
    from /usr/lib/crystal/crystal/main.cr:106:3 in 'main'
    from __libc_start_main
    from _start
    from ???
  ```
  That's not a very helpful error. But why does this happen? It's because the method to print out the results is called `report` without a block. But now this method is renamed to `print_results` in order to fix this.
* (https://github.com/crystal-lang/crystal/commit/0f1ff3573cbacc0d219ff94cef3d293e61c9f515) Minorly improves documentation.


asterite:
You'll now notice that compiling the spec for benchmark will print this warning, which is very annoying. It was previously done at compile time and later change to do that at runtime for that reason.

r00ster91:
Oh actually the two specs I've added print this as well now in the middle of the spec-dots:
```
.................................................Warning: benchmarking without the `--release` flag won't yield useful results

.Warning: benchmarking without the `--release` flag won't yield useful results

.................................................................................................
```
Should I remove the specs?

r00ster91:
Ever planned on merging this?

asterite:
In Ruby the output is:

```
```

Is that the output in this PR? I think because it ends early nothing gets printed, which might be confusing.

PR #7320, opened 2019-01-17
Discourage the use of Bool#to_unsafe other than for C bindings

oprypin:
Dependent codebases shouldn't be infested with the word "unsafe" even if this happens to do the job. Directly using ternary is clearer, too.

### DIFF
--- a/src/bool.cr
+++ b/src/bool.cr
@@ -46,8 +46,10 @@ struct Bool
     hasher.bool(self)
   end
 
-  # Returns `1` for `true` and `0` for `false`.
-  def to_unsafe
+  # Returns an integer derived from the boolean value, for interoperability with C-style booleans.
+  #
+  # The value is `1` for `true` and `0` for `false`.
+  def to_unsafe : LibC::Int
     self ? 1 : 0
   end
 


elorest:
I for one think this makes sense.

src/bool.cr
```diff
@@ -46,8 +46,10 @@ struct Bool
     hasher.bool(self)
   end
 
-  # Returns `1` for `true` and `0` for `false`.
-  def to_unsafe
+  # Returns an integer derived from the boolean value, for interoperability with C-style booleans.
+  #
+  # The value is `1` for `true` and `0` for `false`.
+  def to_unsafe : LibC::Int
```
asterite: I think this return type should be `Int32`, this should fail when `LibC::Int` is `Int64`, for instance (not sure that happens, but if it does then it will fail)

oprypin: @asterite, *at the moment* `LibC::Int` is always `Int32` and there are a lot of assumptions about it. And I think it's good that such things will start failing when that's no longer the case, because it won't happen by itself but rather when trying to port to a new (currently hypothetical) platform.

By definition, this should be interoperable with C `int` primarily. And I thought that setting the return type to `LibC::Int` exactly matches that.

asterite: That's fine, but then the return value should be `LibC::Int.new(self ? 1 : 0)`

PR #7323, opened 2019-01-18
Add bool#to_i

Blacksmoke16:
Based on #7320, this PR adds a similar method to `to_s` but instead returns the integer representation of a bool, 1 for true and 0 for false.

This method, currently, is the same as `to_unsafe` but will be isolated from C binding changes.

### DIFF
--- a/spec/std/bool_spec.cr
+++ b/spec/std/bool_spec.cr
@@ -41,6 +41,11 @@ describe "Bool" do
     it { false.to_s.should eq("false") }
   end
 
+  describe "to_i" do
+    it { true.to_i.should eq 1 }
+    it { false.to_i.should eq 0 }
+  end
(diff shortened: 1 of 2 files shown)


RX14:
What's the usecase?

vladfaust:
@RX14 I could think of a smaller serialization result, for example to store bools more efficiently in Redis ("0" instead of "false"). Or more compact JSONs. Additionally, this method would not require any attention from the language contributors in the foreseeable future

asterite:
Plus there might be a reason why it's not there in Ruby

j8r:
For those who knows nothing about C, or any language/technology accepting `0` and `1` as booleans,  converting a `Bool` to an `Int32` looks strange.

PR #7328, opened 2019-01-21
Add String#rchop? and String#lchop?

j8r:
`String#rchop?` and `String#lchop?` are the nillable variants of `String#rchop` and `String#lchop`.

A use case for them is to remove a prefix/suffix when a string starts/ends with a certain string, else continue.

```cr
# This is inefficient because lchop also do a starts_with? check internally 
if mystring.starts_with? "prefix"
  substring = mystring.lchop "prefix"
# This is also not optimal and verbose
elsif (substring = mystring.lchop "prefix") != mystring
  substring
# The new way to do - try to lchop
elsif substring = mystring.lchop? "prefix"
  substring
end
```

### DIFF
--- a/src/spec/source.cr
+++ b/src/spec/source.cr
@@ -15,10 +15,10 @@ module Spec
   # :nodoc:
   def self.relative_file(file)
     cwd = Dir.current
-    if file.starts_with?(cwd)
-      file = file[cwd.size..-1]
-      file = file[1..-1] if file.starts_with?('/')
+    if basename = file.lchop? cwd
+      basename.lchop '/'
+    else
+      file
     end
-    file
   end
 end
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -553,8 +553,28 @@ describe "String" do
     it { "foo".rchop('o').should eq("fo") }
     it { "foo".rchop('x').should eq("foo") }
 
+    it { "".rchop("").should eq("") }
     it { "foobar".rchop("bar").should eq("foo") }
     it { "foobar".rchop("baz").should eq("foobar") }
+  end
+
+  describe "rchop?" do
+    it { "".rchop?.should be_nil }
+    it { "foo".rchop?.should eq("fo") }
+    it { "foo\n".rchop?.should eq("foo") }
+    it { "foo\r".rchop?.should eq("foo") }
+    it { "foo\r\n".rchop?.should eq("foo\r") }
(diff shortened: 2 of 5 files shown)


Sija:
Why not just calling `lchop/rchop`? You'd want to remove prefix/suffix anyway.

```cr
# This is inefficient because lchop also do a starts_with? check internally 
if mystring.starts_with? "prefix"
  substring = mystring.lchop "prefix"
# This is also not optimal and verbose
elsif (substring = mystring.lchop "prefix") != mystring
  substring
# The new way to do - try to lchop
elsif substring = mystring.lchop? "prefix"
  substring
end

# why all this dance if you can do:
mystring = mystring.lchop "prefix"
```

j8r:
@Sija this are just examples, take this one:

```cr
FILE = <<-E
name=john
num=123
E

name = nil
num = nil
FILE.each_line do |line|
  if line.starts_with? "name="
    name = line.lchop "name="
  elsif line.starts_with? "num="
    num = line.lchop "num="
  end
end
p name, num
```

Sija:
With such implementation I find the API confusing:

```cr
"foo".rchop("bar")  # => "foo"
"foo".rchop?("bar") # => nil
```

It's behavior is nothing like we know from from other `?` suffixed methods from stdlib - either query or nilable - and this one is neither of those.

j8r:
The last commit brings improvements in the language to take advantage of this changes.

For me this methods are similar to `String#[index : Int]?` - try to perform the operation and return a value, else return nil.
Maybe that's the `rchop` and `lchop` behavior that is confusing, not `rchop?` and `lchop?`.

j8r:
You're right @bcardiff , I naively modified the original methods and add new ones.

src/spec/source.cr
```diff
@@ -15,10 +15,10 @@ module Spec
   # :nodoc:
   def self.relative_file(file)
     cwd = Dir.current
-    if file.starts_with?(cwd)
-      file = file[cwd.size..-1]
-      file = file[1..-1] if file.starts_with?('/')
+    if file_prefix = file.lchop? cwd
+      file = file_prefix.lchop '/'
```
asterite:
```sugestion
      file_prefix.lchop '/'
```

asterite:
```suggestion
      file_prefix.lchop '/'
```

j8r: I also renamed prefix to basename

PR #7330, opened 2019-01-21
Make Nil#not_nil! raise NilAssertionError

### LINKED ISSUE #7278: Error class for Nil assertion failure
```crystal
def initialize(value : Array)
  job_class_index = value.index("cls").not_nil! + 1
  job_args_index = value.index("arg").not_nil! + 1
  # ... more ditto
rescue ex : Exception
  if ex.message =~ /Nil assertion failed/
    raise InvalidPayload.new
  else
    raise ex
  end
end
```

Sometimes it's easier to explicitly define `.not_nil!` rather than handling `nil` case for every value. In these situations I just rescue the whole block. The problem is that "Nil assertion failed" error doesn't have its own class. Wouldn't it be better to write this instead:

```crystal
def initialize(value : Array)
  job_class_index = value.index("cls").not_nil! + 1
  job_args_index = value.index("arg").not_nil! + 1
  # ... more ditto
rescue ex : NilAssertionFailed
  raise InvalidPayload.new
end
```

?

r00ster91:
Fixes #7278 

This makes [`Nil#not_nil!`](https://crystal-lang.org/api/0.27.0/Nil.html#not_nil%21%28caller_file%3D__FILE__%2Ccaller_line%3D__LINE__%29-instance-method) raise `NilAssertionError` instead of just `Exception`.
This also makes the `caller_file` and `caller_line` arguments redundant.

### DIFF
--- a/src/exception.cr
+++ b/src/exception.cr
@@ -138,3 +138,14 @@ class NotImplementedError < Exception
     super("Not Implemented: #{item}")
   end
 end
+
+# Raised when a `not_nil!` assertion fails.
+#
+# ```
+# "hello".index('x').not_nil! # raises NilAssertionError ("hello" does not contain 'x')
+# ```
+class NilAssertionError < Exception
+  def initialize(message = "Nil assertion failed")
+    super(message)
+  end
+end
--- a/src/nil.cr
+++ b/src/nil.cr
@@ -99,12 +99,11 @@ struct Nil
     self
   end
 
-  # Raises an exception.
-  # Reports the file and line at which the `not_nil!` assertion is called.
+  # Raises `NilAssertionError`.
   #
   # See also: `Object#not_nil!`.
-  def not_nil!(caller_file = __FILE__, caller_line = __LINE__)
-    raise "Nil assertion failed at #{caller_file}:#{caller_line}"
+  def not_nil!
+    raise NilAssertionError.new
   end
 
   def clone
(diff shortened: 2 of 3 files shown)


bew:
> The location of where the error happened exactly is also somewhere in the stack trace here and you have to find it first.

But the stacktrace is not always visible, sometimes you just have some `???`, and you can't debug a thing... At least knowing where to look, as given in the error message is better imo.

> There should be rather a general solution to this problem for all errors, not just for the not_nil! error.

True, but having it for `not_nil!` is a start!

r00ster91:
> But the stacktrace is not always visible, sometimes you just have some `???`, and you can't debug a thing... At least knowing where to look, as given in the error message is better imo.

Did this ever happen to you at an error where not just `Exception` was raised? Then I believe that's the problem and caused by that just an `Exception` has been raised instead of a proper error exception class (like `NilAssertionError`). But I'm not too sure about this. That should probably be a bit more investigated.

> True, but having it for `not_nil!` is a start!

But not a good start IMO. Do you think in the future all the errors should have `(caller_file = __FILE__, caller_line = __LINE__)` as the arguments? I think that'd be a pretty bad solution and there are much better ways to go.

asterite:
Yes, I don't think the `__FILE__` and `__LINE__` PR should have been merged, the info is already provided in the stack trace. Otherwise we'd need to pass that in every `raise`, but it's a bit redundant.

PR #7334, opened 2019-01-23
Fix parsing and behaviour of `->foo.[]` and other operators 

### LINKED ISSUE #7319: Bug: Can't create Proc from index operator
I want to get the index operator method (`[]`) of an `Array` as `Proc`, but I keep getting the same error: 

```
expecting token 'IDENT', not [].
```

Here is how I try to reference the index operator (just like a normal method):

```crystal
->array.[](Int32)
```

→ _[StackOverflow Question](https://stackoverflow.com/questions/54228749/how-to-get-index-operator-method-as-proc)_

|    |   |
| ---------  | ---------- |
| **Crystal version** | 0.27.0 |
| **OS** | elementary OS (Ubuntu 18.04) |

asterite:
Fixes #7319

Split into several commits, but basically:
- parsing of `->foo.op` where `op` was `[]`, `+`, etc. didn't work, so that's fixed
- after fixing that I tried taking a proc pointer where the variable was an integer and it crashed. The reasons are many: most of these methods (like `+`) are primitives and inlined by the compiler, and this didn't work when used as proc pointer. Another reason is that primitives are not passed by reference (pointer) and that's needed when the proc pointer closures over something (it's a pointer). Both of these things are fixed.
- there's another commit that improves something by making it include `Indexable`

The code in this PR feels slightly hacky, but primitives are a special case in the language and so they usually need special code to handle them, so it's probably fine.


asterite:
@bcardiff It seems something overflows in the codegen debug code...

bcardiff:
@asterite yup debugging in CI since nightly also failed, it seems `LibC::PthreadMutexT#@__align` offset is `5998840007512307828` _sometimes_. Which triggers the overflow in even in u64. I'm trying to narrow it down...

asterite:
@bcardiff The plot thickens :-)

But I can't find anywhere where `__align` is used in the code... it's a bit strange.

bcardiff:
It's https://github.com/crystal-lang/crystal/blob/master/src/lib_c/x86_64-linux-gnu/c/sys/types.cr#L45-L49 that is required in the prelude , now I need to CI to fail again in my fork 👷 🛠 ...

src/compiler/crystal/codegen/primitives.cr
```diff
@@ -366,8 +366,8 @@ class Crystal::CodeGenVisitor
   end
 
   private def codegen_raise_overflow
-    location = @call_location.not_nil!
-    set_current_debug_location(location) if @debug.line_numbers?
+    location = @call_location
```
bcardiff: Using `#not_nil!` is intentional here. Having debug location information is mandatory for when debug information is wanted to be emitted. Otherwise the program will fail with `--debug` but run ok, so it was to catch that missing information even when not using `--debug`.

asterite: That makes sense. However, with this PR this code can now be triggered when there's no call at all. For example:

```crystal
a = 1
f = ->a.+(Int32) # this ends up generation a function that implements + and can raise overflow
f.call(Int32::MAX)
```

The `->a.+(Int32)` can happen multiple times in code so there's no single "call" to use, unlike when you write `a + 1`. I don't know how to handle this...

bcardiff: Can't we reach the location of the proc literal when inlining the primitive?

asterite: We could. But that will not always work correctly because two proc literals that point to the same method will result in the same function being generated, so the location will point to the first literal, which is not always correct.

Maybe we need a reproducible test case to check for when this breaks.

bcardiff: Oh, I see. Well the story needs to continue with more support for debugging information then.

asterite: I'll see what I can do :-)

bcardiff: I mean in a future PR, but if you want to tackle that now I wont interfere :-P

sdogruyol:
This seems go to :ship: :+1:

straight-shoota:
@asterite Can you rebase this branch on master to make sure it's all still good?

asterite:
@straight-shoota Rebased and green.

straight-shoota:
Hm, should the last commit be squashed and then merge three indivdual commits?

asterite:
@straight-shoota Done

straight-shoota:
Awesome! Thanks @asterite

PR #7338, opened 2019-01-27
Add nillable range fetching

j8r:
This adds `String#[Range]?`, `String#[Int, Int]?`, `Array[Range]?` and `Array#[Int, Int]?`.
Like the existing `String#[Int]?` and `Array#[Int]?`, the bound check is done internally.

Before:
```cr
ary = [1, 2, 3]
if var.size > 1
  sub = var[1..-1]
  ...
end
```
The  typical `.size > start_index` can now be removed.
```cr
ary = [1, 2, 3]
if sub = ary[1..-1]?
  ...
end
```

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -453,7 +453,7 @@ class Array(T)
   # element). Additionally, an empty array is returned when the starting index
   # for an element range is at the end of the array.
   #
-  # Raises `IndexError` if the starting index is out of range.
+  # Raises `IndexError` if the range's start is out of range.
   #
   # ```
   # a = ["a", "b", "c", "d", "e"]
@@ -468,6 +468,16 @@ class Array(T)
     self[*Indexable.range_to_index_and_count(range, size)]
   end
 
+  # Like `#[Range(Int, Int)]`, but returns `nil` if the range's start is out of range.
+  #
+  # ```
+  # a = ["a", "b", "c", "d", "e"]
+  # a[6..10]? # => nil
+  # ```
+  def []?(range : Range(Int, Int))
+    self[*Indexable.range_to_index_and_count(range, size)]?
+  end
+
   # Returns count or less (if there aren't enough) elements starting at the
   # given start index.
   #
@@ -475,34 +485,37 @@ class Array(T)
   # element). Additionally, an empty array is returned when the starting index
   # for an element range is at the end of the array.
   #
-  # Raises `IndexError` if the starting index is out of range.
+  # Raises `IndexError` if the *start* index is out of range.
+  #
+  # Raises `ArgumentError` if *count* is negative.
   #
   # ```
   # a = ["a", "b", "c", "d", "e"]
   # a[-3, 3] # => ["c", "d", "e"]
-  # a[6, 1]  # raise IndexError
   # a[1, 2]  # => ["b", "c"]
   # a[5, 1]  # => []
+  # a[6, 1]  # raises IndexError
   # ```
   def [](start : Int, count : Int)
-    raise ArgumentError.new "Negative count: #{count}" if count < 0
+    self[start, count]? || raise IndexError.new
+  end
 
-    if start == size
-      return Array(T).new
-    end
+  # Like `#[Int, Int]` but returns `nil` if the *start* index is out of range.
+  def []?(start : Int, count : Int)
+    raise ArgumentError.new "Negative count: #{count}" if count < 0
+    return Array(T).new if start == size
 
     start += size if start < 0
-    raise IndexError.new unless 0 <= start <= size
 
-    if count == 0
-      return Array(T).new
-    end
+    if 0 <= start <= size
+      return Array(T).new if count == 0
 
-    count = Math.min(count, size - start)
+      count = Math.min(count, size - start)
 
-    Array(T).build(count) do |buffer|
-      buffer.copy_from(@buffer + start, count)
-      count
+      Array(T).build(count) do |buffer|
+        buffer.copy_from(@buffer + start, count)
+        count
+      end
     end
   end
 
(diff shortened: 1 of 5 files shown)


src/array.cr
```diff
@@ -457,14 +457,26 @@ class Array(T)
     self[*Indexable.range_to_index_and_count(range, size)]
   end
 
+  # like `#[Range(Int, Int)]`, but returns nil if the range's start is out of range.
```
r00ster91: Let's capitalize on the beginning of a method doc please. Also can we put markdown on `nil`?
```suggestion
  # Like `#[Range(Int, Int)]`, but returns `nil` if the range's start is out of range.
```

src/string.cr
```diff
@@ -721,31 +721,44 @@ class String
...
   # ```
-  # "hello"[0..2]   # "hel"
-  # "hello"[0...2]  # "he"
-  # "hello"[1..-1]  # "ello"
-  # "hello"[1...-1] # "ell"
+  # "hello"[0..2]   # => "hel"
+  # "hello"[0...2]  # => "he"
+  # "hello"[1..-1]  # => "ello"
+  # "hello"[1...-1] # => "ell"
+  # "hello"[6..7]?  # raise IndexErrors
```
```suggestion
  # "hello"[6..7]?  # raises IndexError
```

src/array.cr
```diff
@@ -474,24 +486,27 @@ class Array(T)
...
+      count = Math.min(count, size - start)
 
-    Array(T).build(count) do |buffer|
-      buffer.copy_from(@buffer + start, count)
-      count
+      Array(T).build(count) do |buffer|
+        buffer.copy_from(@buffer + start, count)
+        count
+      end
+    else
```
Sija: This whole `else` branch is redundant.

src/string.cr
```diff
@@ -780,13 +792,9 @@ class String
         {count, 0}
       end
     elsif start == i
-      if count >= 0
-        return ""
-      else
-        raise ArgumentError.new "Negative count"
-      end
+      ""
     else
```
Sija: ditto

src/string.cr
```diff
@@ -861,13 +873,9 @@ class String
         {count, slice_size}
       end
     elsif start == bytesize
-      if count >= 0
-        return ""
-      else
-        raise ArgumentError.new "Negative count"
-      end
+      ""
     else
```
Sija: ditto

src/string.cr
r00ster91:
```suggestion
  # "hello"[6..7]   # raises `IndexError`
```

Sija: @r00ster91 surprisingly not, grep the code to see precedents.

src/string.cr
```diff
@@ -707,11 +707,11 @@ class String
...
-  # "hello"[0]  # 'h'
-  # "hello"[1]  # 'e'
-  # "hello"[-1] # 'o'
-  # "hello"[-2] # 'l'
-  # "hello"[5]  # raises IndexError
+  # "hello"[0]  # => 'h'
+  # "hello"[1]  # => 'e'
+  # "hello"[-1] # => 'o'
+  # "hello"[-2] # => 'l'
+  # "hello"[5]  # => raises IndexError
```
r00ster91: Let's stay consistent with the "raises" in the example below and not use a `=>` here. Also I think `=>` should only be used when that actually returns something.
```suggestion
  # "hello"[5]  # raises IndexError
```

src/string.cr
r00ster91: Deleted that suggestion, noticed it's in code so surrounding `IndexError` by grave accents wouldn't be useful.

j8r:
Docs polished, thanks @Sija and @r00ster91 :+1:

src/string.cr
```diff
@@ -721,31 +721,44 @@ class String
...
   # ```
-  # "hello"[0..2]   # "hel"
-  # "hello"[0...2]  # "he"
-  # "hello"[1..-1]  # "ello"
-  # "hello"[1...-1] # "ell"
+  # "hello"[0..2]   # => "hel"
+  # "hello"[0...2]  # => "he"
+  # "hello"[1..-1]  # => "ello"
+  # "hello"[1...-1] # => "ell"
+  # "hello"[6..7]?  # raises IndexError
```
```suggestion
  # "hello"[6..7]   # raises IndexError
```

RX14:
Let's get consensus if we want these methods before looking at the implementation please. I think they're a good addition.

straight-shoota:
The semantics need to be solidified:

Why does `[1, 2, 3][1, 3]?`return `[2, 3]`? Maybe it should be `nil` because there are not 3 elements to read starting from index 1.
On the other hand, `[1, 2, 3][4, 0]?` returns `nil`. Shouldn't that better be `[]` then?

straight-shoota:
As a more general question: Shouldn't these methods (including the already existing `Array#[](Range)` etc.) be defined on `Indexable`?

Sija:
> Why does `[1, 2, 3][1, 3]?`return `[2, 3]`? Maybe it should be `nil` because there are not 3 elements to read starting from index 1.

FYI, that's the same as in Ruby.

>  On the other hand, `[1, 2, 3][4, 0]?` returns `nil`. Shouldn't that better be `[]` then?

I think not, since `nil` in this case signifies missing start index - also, it's the same as in Ruby.

Personally I found this distinction informative, returning:
- `nil`: for invalid start index
- `[]`: for valid start index, but lacking elements to return

straight-shoota:
Thank you @j8r 
I've opened #7619 to discuss moving these methods to `Indexable`

bcardiff:
I've just noticed that given this PR and #7439, `Slice` should also have a `#[]?(Range)` version probably for consistency.

RX14:
@bcardiff is there an issue or PR for this, and if not I think that would be a better way to make sure this wasn't forgotten

PR #7348, opened 2019-01-30
Scheduler: resume only alive and non-current fibers

### LINKED ISSUE #3900: Crash in fiber switching with select
```crystal
ch1 = Channel::Buffered(Int32).new(1)
ch2 = Channel::Buffered(Int32).new(1)
res = [] of Int32

spawn do
  loop do
    select
    when x = ch1.receive
      res << x
    when y = ch2.receive
      res << y
    end
  end
end

spawn do
  3.times do |i|
    select
    when ch1.send(i)
    when ch2.send(i)
    end
  end
end

Fiber.yield
p res
```

```
[4466037] *CallStack::print_backtrace:Int32 +117
[4444920] __crystal_sigfault_handler +56
[4758788] sigfault_handler +40
[140503573855408] ???
[0] ???
```

forksaber:
Fixes #3900

`select` when used with buffered channels can cause a single fiber to be enqueued multiple times.
This can cause `Crystal::Scheduler.reschedule` to segfault when:
- resuming an already dead fiber
- resuming the currently running fiber from itself

This PR fixes that by having an `@alive` flag associated with every fiber and resuming only alive and non-current fibers.

### DIFF
--- a/spec/std/concurrent/select_spec.cr
+++ b/spec/std/concurrent/select_spec.cr
@@ -100,4 +100,63 @@ describe "select" do
     sleep
     x.should eq 1
   end
+
+  it "select with buffered channels shouldn't segfault on resuming already completed fiber, fixed #3900" do
+    ch1 = Channel::Buffered(Int32).new(1)
+    ch2 = Channel::Buffered(Int32).new(1)
+    res = [] of Int32
+
+    spawn do
+      3.times do
+        select
+        when x = ch1.receive
+          res << x
+        when y = ch2.receive
+          res << y
+        end
+      end
+    end
+
+    spawn do
+      3.times do |i|
+        select
+        when ch1.send(1)
+        when ch2.send(2)
+        end
+      end
+    end
+
+    sleep 0
+    res.should eq([1, 2, 1])
+  end
+
+  it "select with buffered channels shouldn't segfault when current fiber resumes itself" do
+    ch1 = Channel::Buffered(Int32).new(1)
+    ch2 = Channel::Buffered(Int32).new(1)
+    res = [] of Int32
+
+    spawn do
+      3.times do
+        select
+        when x = ch1.receive
+          res << x
+        when y = ch2.receive
+          res << y
+        end
+      end
+    end
+
+    spawn do
+      3.times do |i|
+        select
+        when ch1.send(1)
+        when ch2.send(2)
+        end
+      end
+      sleep 0
+    end
+
+    sleep 0
+    res.should eq([1, 2, 1])
+  end
 end
--- a/src/crystal/scheduler.cr
+++ b/src/crystal/scheduler.cr
@@ -63,7 +63,11 @@ class Crystal::Scheduler
   end
 
   protected def reschedule : Nil
-    if runnable = @runnables.shift?
+    while runnable = @runnables.shift?
+      next if runnable == @current
+      break if runnable.alive
+    end
+    if runnable
       runnable.resume
     else
       Crystal::EventLoop.resume
--- a/src/fiber.cr
+++ b/src/fiber.cr
@@ -62,6 +62,7 @@ class Fiber
 
   # :nodoc:
   property previous : Fiber?
+  property alive : Bool = true
 
   # :nodoc:
   def self.inactive(fiber : Fiber)
@@ -155,6 +156,7 @@ class Fiber
     ex.inspect_with_backtrace(STDERR)
     STDERR.flush
   ensure
+    @alive = false
     @@stack_pool << @stack
 
     # Remove the current fiber from the linked list


kostya:
fix for https://github.com/crystal-lang/crystal/issues/3900?

kostya:
Not really understand how it works, but looks like hack, is not the fix should be in Channel::Buffered or in select?

asterite:
At this point in time I don't think it's worth trying to fix things without considering everything together. Adding an `alive` instance variable might work (not sure) but it does feel like a hack. Why is a fiber enqueued after it's finished?

ysbaddaden:
The problem is how `select` is implemented: it creates *many* events to resume the current fiber. One event will resume the fiber and then it deletes other events, which is usually fine, but it appears there is a time-window during which another resume event could slip by (even with a single thread), and a context switch can end up trying to resume the current fiber, which is impossible (it's running, so it's context isn't saved on the stack) and leads to a segfault.

I'm not sure `select` can lead to resume a dead fiber, maybe on *very* bad luck.

I think #6990 with a little rewrite would make the bug explicit by raising an exception instead of segfaulting. I'll reopen it in a few days.

forksaber:
I'll try to explain where the problem is with current `select` implementation. Just to make it simpler i'll use a single channel example.

Consider the following example which segfaults:
```crystal
ch = Channel::Buffered(Int32).new(1)

spawn do # spawned fiber
  ch.send(1)
  select
  when ch.send(1)
  when ch.send(2)
  end
end

ch.receive
ch.receive
```
2 fibers here:`main_fiber` and `spawned_fiber`
3 variables to hold state: `ch.@senders`, `ch.@receivers`, `@runnables`

Execution flow:

<pre>
 ------- main_fiber ----------
 - create channel `ch`
 - create and enqueue spawned_fiber (<b>State</b> @runnables: [spawned_fiber])
 - ch.receive
     channel is empty
     ch.@receivers << Fiber.current (<b>State</b> ch.@receivers: [main_fiber])
     <b>reschedule</b>
 
-------spawned_fiber-----------    (<b>State</b> @runnables: [])
 - ch.send(1)
     @buffer << 1
     enqueue all @receivers       (<b>State</b> @runnables: [main_fiber], @ch.receivers: [])
 - select....end
     ch is full for both <b>when cases</b> so append to @senders for both branches
     2.times { ch.@senders << Fiber.current } (ops.each &.wait)
     (<b>State</b> @senders: [spawned_fiber, spawned_fiber])
     <b>reschedule</b>

---------main_fiber--------------  (<b>State</b> @runnables: [])
- ch.receive(1st) resumes
     receives value = 1
     enqueues all ch.@senders  (<b>State</b> @runnables: [spawned_fiber, spawned_fiber],
                                      ch.@senders: [])
- ch.receive(2nd)
     no value in channel
     ch.@receivers << main_fiber (<b>State</b> ch.@receivers: [main_fiber])
     <b>reschedule</b>

---------spawned_fiber----------- (<b>State</b>@runnables: [spawned_fiber])
- select....end resumes
    ch1.send(1) succeeds
    enqueues all ch1.@receivers  (<b>State</b>@runnables : [spawned_fiber, main_fiber])
      
- spawned_fiber ends
     <b>reschedule</b>

Schedule tries to resume spawned_fiber which is already dead....segfaults  
</pre>

The problem is that the `selecting` fiber is enqueued multiple times before the `select` implementation gets a chance to call `unwait` because of the non blocking nature of buffered channels ( as long as the buffer is not empty/full).

This issue doesn't happen with unbuffered channels because as soon as a send or receive succeeds, control is ceded back to the `selecting` fiber which then calls `unwait` removing the `selecting` fiber from any `@senders/@receivers` before its enqueued.

@asterite, @kostya I'd argue that this is a pragmatic solution given that 
   - the core team is unlikely to accept a patch which changes the implementation significantly
   - this doesn't change any external api
   -  much of this is likely to be rewritten once multithreading lands

ysbaddaden:
Thanks for the clarification!

I believe the problem is two fold:
1. `select` implementation is broken —I knew it was on MT (Multi Thread);
2. the current scheduler is unsafe against unexpected situations (switch to current fiber, resume dead fiber).

Yet, hiding the issue by overlooking "trying to switch to current fiber" or "trying to resume a dead fiber" isn't a good idea. I'd like the scheduler to raise and make the bug obvious & acknowledged. Hence why I'm proposing to resume #6990 instead, tweaked to raise in these 2 situations (not just dead fiber).

ysbaddaden:
I think `select` should be rewritten to:

1. spawn a fiber for each select clause (A, B);
2. reschedule the current fiber (C);
3. eventually, on wakeup (from A or B):
   1. raise an exception if one was set (e.g. timeout);
   2. set the returned value.

Each spawned fiber (A, B) shall eventually:
1. be run, and wait on its clause to become true;
2. then check & set a flag:
3. on failure: the fiber does nothing more and returns;
4. on success:
    1. cancel other fibers/events (A, B);
    2. get a value (e.g. channel);
    3. set `select` value or an exception to raise (e.g. timeout);
    4. resume the initial fiber (C).

I probably overlooked some details. Slower, but safe, and almost MT-safe too (using an atomic flag).

forksaber:
closing in favour of #7414

PR #7350, opened 2019-01-31
Adding libxml2 xmlAddChild binding to XML standard library

Vici37:
### What
Adding the libxml2 xmlAddChild binding to the XML node class so that XML nodes can ... add children.

### Why
I'm working on a personal project involving some XML that requires some nodes being shuffled around, and couldn't find an approach to move child nodes from one parent node to another. Please let me know if I'm just blind and there's a completely valid approach to doing this.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -716,6 +716,10 @@ module Crystal
         assert_macro "", %({{{:a => 1, :b => 3}.size}}), [] of ASTNode, "2"
       end
 
+      it "executes sort_by" do
+        assert_macro "", %({{["abc", "a", "ab"].sort_by { |x| x.size }}}), [] of ASTNode, %(["a", "ab", "abc"])
+      end
+
       it "executes empty?" do
         assert_macro "", %({{{:a => 1}.empty?}}), [] of ASTNode, "false"
       end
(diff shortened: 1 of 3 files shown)


src/xml/node.cr
```diff
@@ -477,6 +477,11 @@ struct XML::Node
     LibXML.xmlUnlinkNode(self)
   end
 
+  # Adds the node as the child of this node in the XML document
+  def addChild(child : Node)
```
r00ster91: In Crystal (or in the standard library at least) methods have the `snake_case`, not the `camelCase`: (spec needs to be changed too)
```suggestion
  def add_child(child : Node)
```

Vici37: Good catch! Fixed it up. I've been spending too much time in Java land.

asterite:
Also note that in Nokogiri (libxml2 bindings for Ruby) [this method](https://github.com/sparklemotion/nokogiri/blob/c021f65d94f979bb8656d989949fc1bfc57b91a6/ext/nokogiri/xml_node.c#L163) is not just a single line: it checks that the node is OK to be reparented, and removes the previous parent from the given node, etc. It's **very complex**. So let's be careful with this.

Vici37:
That's fair @asterite - for my project, I'm not too concerned if I shoot myself in the foot by doing something improper. To me, it's more valuable to have the xmlAddChild method that may break me in certain edge cases than to not have xmlAddChild at all. Obviously a less than stellar experience for the standard library if something does go sideways, so I understand the desire to get it right first.

I'll work on porting over the Nokogiri implementation (thanks for the link!) and will update this PR when I've gotten through that rabbit hole.

Thanks for the feedback!

PR #7352, opened 2019-01-31
Optimize String#compare in case of ASCII only

r00ster91:
This optimizes `String#compare` when *`other`* and `self` is ASCII:
```cr
Benchmark.ips do |bm|
  bm.report "String#new_compare with ASCII" do
    "HeLlo".new_compare("hELlO", true) == 0
  end

  bm.report "String#compare with ASCII" do
    "HeLlo".compare("hELlO", true) == 0
  end
end
```
```
String#new_compare with ASCII 504.84M (  1.98ns) (±31.44%)  0 B/op        fastest
    String#compare with ASCII  12.06M ( 82.89ns) (±33.57%)  0 B/op  41.85× slower
```

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -2410,10 +2410,6 @@ class String
   # this string is less, equal or greater than *other*, optionally in a *case_insensitive*
   # manner.
   #
-  # If *case_insitive* is `false`, this method delegates to `<=>`. Otherwise,
-  # the strings are compared char-by-char, and ASCII characters are compared in a
-  # case-insensitive way.
-  #
   # ```
   # "abcdef".compare("abcde")   # => 1
   # "abcdef".compare("abcdef")  # => 0
@@ -2422,29 +2418,55 @@ class String
   #
   # "abcdef".compare("ABCDEF", case_insensitive: true) # => 0
   # "abcdef".compare("ABCDEG", case_insensitive: true) # => -1
+  #
+  # "heIIo".compare("heııo", case_insensitive: true, Unicode::CaseOptions::Turkic) # => 0
   # ```
-  def compare(other : String, case_insensitive = false)
+  def compare(other : String, case_insensitive = false, options = Unicode::CaseOptions::None)
     return self <=> other unless case_insensitive
 
-    reader1 = Char::Reader.new(self)
-    reader2 = Char::Reader.new(other)
-    ch1 = reader1.current_char
-    ch2 = reader2.current_char
+    if ascii_only? && other.ascii_only?
+      position = 0
 
-    while reader1.has_next? && reader2.has_next?
-      cmp = ch1.downcase <=> ch2.downcase
-      return cmp.sign if cmp != 0
+      while position < bytesize && position < other.bytesize
+        byte1 = to_unsafe[position]
+        byte2 = other.to_unsafe[position]
 
-      ch1 = reader1.next_char
-      ch2 = reader2.next_char
-    end
+        # Lowercase both bytes
+        if 65 <= byte1 <= 90
+          byte1 += 32
+        end
+        if 65 <= byte2 <= 90
+          byte2 += 32
+        end
 
-    if reader1.has_next?
-      1
-    elsif reader2.has_next?
-      -1
+        comparison = byte1 <=> byte2
+        return comparison unless comparison == 0
+
+        position += 1
+      end
+
+      bytesize <=> other.bytesize
     else
-      0
+      reader1 = Char::Reader.new(self)
+      reader2 = Char::Reader.new(other)
+      char1 = reader1.current_char
+      char2 = reader2.current_char
+
+      while reader1.has_next? && reader2.has_next?
+        comparison = char1.downcase(options) <=> char2.downcase(options)
+        return comparison.sign unless comparison == 0
+
+        char1 = reader1.next_char
+        char2 = reader2.next_char
+      end
+
+      if reader1.has_next?
+        1
+      elsif reader2.has_next?
+        -1
+      else
+        0
+      end
     end
   end
 
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -2245,22 +2245,38 @@ describe "String" do
     end
   end
 
-  it "compares non-case insensitive" do
-    "fo".compare("foo").should eq(-1)
-    "foo".compare("fo").should eq(1)
-    "foo".compare("foo").should eq(0)
-    "foo".compare("fox").should eq(-1)
-    "fox".compare("foo").should eq(1)
-    "foo".compare("Foo").should eq(1)
-  end
+  describe "compare" do
+    it "compares case-sensitive" do
+      "fo".compare("foo").should eq(-1)
+      "foo".compare("fo").should eq(1)
+      "foo".compare("foo").should eq(0)
+      "foo".compare("fox").should eq(-1)
+      "fox".compare("foo").should eq(1)
+      "foo".compare("Foo").should eq(1)
+      "hällo".compare("Hällo").should eq(1)
+      "".compare("").should eq(0)
+    end
 
-  it "compares case insensitive" do
-    "fo".compare("FOO", case_insensitive: true).should eq(-1)
-    "foo".compare("FO", case_insensitive: true).should eq(1)
-    "foo".compare("FOO", case_insensitive: true).should eq(0)
-    "foo".compare("FOX", case_insensitive: true).should eq(-1)
-    "fox".compare("FOO", case_insensitive: true).should eq(1)
-    "fo\u{0000}".compare("FO", case_insensitive: true).should eq(1)
+    it "compares case-insensitive" do
+      "foo".compare("FO", case_insensitive: true).should eq(1)
+      "FOO".compare("fo", case_insensitive: true).should eq(1)
+      "fo".compare("FOO", case_insensitive: true).should eq(-1)
+      "FOX".compare("foo", case_insensitive: true).should eq(1)
+      "foo".compare("FOX", case_insensitive: true).should eq(-1)
+      "foo".compare("FOO", case_insensitive: true).should eq(0)
+      "hELLo".compare("HellO", case_insensitive: true).should eq(0)
+      "fo\u{0}".compare("FO", case_insensitive: true).should eq(1)
+      "fo".compare("FO\u{0}", case_insensitive: true).should eq(-1)
+      "\u{0}".compare("\u{0}", case_insensitive: true).should eq(0)
+      "z".compare("hello", case_insensitive: true).should eq(1)
+      "h".compare("zzz", case_insensitive: true).should eq(-1)
+      "ä".compare("äA", case_insensitive: true).should eq(-1)
+      "äÄ".compare("äÄ", case_insensitive: true).should eq(0)
+      "heIIo".compare("heııo", case_insensitive: true, options: Unicode::CaseOptions::Turkic).should eq(0)
+      "".compare("abc", case_insensitive: true).should eq(-1)
+      "abc".compare("", case_insensitive: true).should eq(1)
+      "abcA".compare("abca", case_insensitive: true).should eq(0)
+    end
   end
 
   it "builds with write_byte" do


ysbaddaden:
What about longer and much longer strings?

src/string.cr
```diff
@@ -2405,25 +2404,52 @@ class String
...
-
-    while reader1.has_next? && reader2.has_next?
-      cmp = ch1.downcase <=> ch2.downcase
-      return cmp.sign if cmp != 0
+    if ascii_only? && other.ascii_only?
+      if self.size > other.size
+        return 1
+      elsif self.size < other.size
+        return -1
+      end
```
r00ster91: These greater-or-less checks are pretty ugly and generally make this method slightly slower I think.
What is the usecase for a method which in addition to being case-insensitive also returns which of the two is greater or smaller? I think this feature is/will be very rarely used and mostly a ` == 0` will be put after this method.
I really wish this method would just return `true` or `false` and would be case-insensitive by default...

r00ster91:
@ysbaddaden then it's even slower:
```cr
string = "aA" * 500
Benchmark.ips do |bm|
  bm.report "String#new_compare for ASCII" do
    string.new_compare(string, true) == 0
  end

  bm.report "String#compare for ASCII" do
    string.compare(string, true) == 0
  end
end

```
```
String#new_compare for ASCII 367.12M (  2.72ns) (±22.18%)  0 B/op          fastest
    String#compare for ASCII  92.58k (  10.8µs) (±19.51%)  0 B/op  3965.56× slower
```

ysbaddaden:
Great!

src/string.cr
ysbaddaden: The method is named `#compare` not `#equals?`.

This boils down to 2 integer comparisons, performance impact is likely negligible compared to the whole method.

r00ster91: Yes that was basically a request to rename it to something like `equals?` and make it case-insensitive by default.

Actually 4 integer comparisons. One before the loop (these two^) and 2 integer comparisons in every single iteration of `@bytesize.times do |i|`. I would really like to avoid these and make the `== 0` redundant.

r00ster91: Thanks with that it's as fast as [this](https://gist.github.com/r00ster91/ef1cdfd0c20b3ab2f69918c43a8e9e83#file-string-cr) now.
Any opinions on making it this instead of the current `compare` though?

asterite: If you write the method that compares case-insensitive and returns `true` or `false`, how it performs compared to this?

asterite: (but in any case sometimes knowing less-than or greater-than is good, so we won't change the fact that it returns a number)

asterite: Hey, I think this condition is wrong: if a string is longer/shorter than other it doesn't mean it's lower/greater. For example "z" is greater than "hello", while "h" is lower than "zzz".

src/string.cr
```diff
@@ -2405,25 +2404,47 @@ class String
   def compare(other : String, case_insensitive = false)
     return self <=> other unless case_insensitive
 
-    reader1 = Char::Reader.new(self)
-    reader2 = Char::Reader.new(other)
-    ch1 = reader1.current_char
-    ch2 = reader2.current_char
+    if ascii_only? && other.ascii_only?
+      size_cmp = self.bytesize <=> other.bytesize
+      return size_cmp if size_cmp != 0
```
bew: Quoting @asterite (from earlier comments) about this condition:
> Hey, I think this condition is wrong: if a string is longer/shorter than other it doesn't mean it's lower/greater. For example "z" is greater than "hello", while "h" is lower than "zzz".

r00ster91: Oh I think this can be fixed by simply making it:
```cr
cmp = self <=> other
return cmp if cmp != 0
```
And that means all the specs are wrong too, right? Here's the first one:
```cr
 "fo".compare("FOO", case_insensitive: true).should eq(-1)
```
Here it says `-1` but actually `"fo" <=> "FOO"` returns `1` which this method should actually return.

r00ster91:
I changed `String#compare` to `String#equals?` now and it returns a `Bool` because I couldn't really find a way to make `compare` as fast as `equals?` while making it have the same behavior as `<=>` in the ASCII-only branch.

But also, I've scrolled for quite some time through this: https://github.com/search?l=C&o=asc&q=strcasecmp&s=indexed&type=Code. `strcasecmp` is similar to `String#compare`. It returns a positive number when the first string is greater than the other, zero if they are equal and a negative number if the string is less than other (e.g. `strcasecmp("hi", "h");` returns `105`, that's the codepoint of `h`). But I couldn't find a single instance in all this code where the user didn't only want to know whether the strings are equal regardless of the case. The codepoint that is returned by that method is never actually used. They only check if it's negative or positive (you often see `!` before `strcasecmp` or `== 0`, `!= 0` at the end).
So I think it would be better to just let this method return a bool already.

asterite:
A use case: sort strings case insensitive.

src/string.cr
```diff
@@ -2385,46 +2385,45 @@ class String
...
-  # "abcdef".compare("ABCDEF", case_insensitive: true) # => 0
-  # "abcdef".compare("ABCDEG", case_insensitive: true) # => -1
+  # "HellO".equals?("hELLo")             # => true
+  # "hello world".equals?("HELLO WORLD") # => true
+  # "Ü".equals?("ü")                     # => true
+  # "hello".equals?("world")             # => false
   # ```
-  def compare(other : String, case_insensitive = false)
-    return self <=> other unless case_insensitive
+  def equals?(other : String)
```
bew: Why is `equals?` case insensitive now? It's not intuitive, and also not configurable..

r00ster91:
ok I think I've found a relatively efficient way to do it now.

r00ster91:
A bug in the Crystal compiler in the [test_darwin CircleCI](https://circleci.com/gh/crystal-lang/crystal/17741?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

src/string.cr
```diff
@@ -2397,12 +2397,26 @@ class String
...
+  #
+  # "heIIo".compare("heııo", case_insensitive: true, Unicode::CaseOptions::Turkic) # => 0
   # ```
-  def compare(other : String, case_insensitive = false)
+  def compare(other : String, case_insensitive = false, options = Unicode::CaseOptions::None)
     comparison = self <=> other
-    return comparison if (!case_insensitive || comparison != 0) && self.bytesize != other.bytesize
+    return comparison unless case_insensitive
+
+    if ascii_only? && other.ascii_only? && (options.none? || options.ascii?)
```
Sija: I don't think `&& (options.none? || options.ascii?)` is needed here. Both strings are ASCII so we can safely take a shortcut ignoring the *options* value.

src/string.cr
```diff
@@ -2397,12 +2397,26 @@ class String
...
+      if self.size != other.size
+        if self.starts_with?(other[0]? || return 1)
```
Sija: That's unreadable, please move the `return` into its own line.

src/string.cr
r00ster91: Just wanted to prove you that it's needed but surprisingly you are right, thanks!

Sija:
CI failed on Darwin with:

```
./bin/crystal build  -o .build/std_spec spec/std_spec.cr

while requiring "prelude" (Exception)
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::Program#top_level_semantic<Crystal::ASTNode+>:Tuple(Crystal::ASTNode+, Crystal::TypeDeclarationProcessor)
  from Crystal::Program#semantic<Crystal::ASTNode+, Bool>:Crystal::ASTNode+
  from __crystal_main
  from main
Caused by: while requiring "exception" (Exception)
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::Program#top_level_semantic<Crystal::ASTNode+>:Tuple(Crystal::ASTNode+, Crystal::TypeDeclarationProcessor)
  from Crystal::Program#semantic<Crystal::ASTNode+, Bool>:Crystal::ASTNode+
  from __crystal_main
  from main
Caused by: while requiring "callstack" (Exception)
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::Program#top_level_semantic<Crystal::ASTNode+>:Tuple(Crystal::ASTNode+, Crystal::TypeDeclarationProcessor)
  from Crystal::Program#semantic<Crystal::ASTNode+, Bool>:Crystal::ASTNode+
  from __crystal_main
  from main
Caused by: while requiring "c/dlfcn" (Exception)
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::Program#top_level_semantic<Crystal::ASTNode+>:Tuple(Crystal::ASTNode+, Crystal::TypeDeclarationProcessor)
  from Crystal::Program#semantic<Crystal::ASTNode+, Bool>:Crystal::ASTNode+
  from __crystal_main
  from main
Caused by: can't find file 'c/dlfcn'

If you're trying to require a shard:
- Did you remember to run `shards install`?
- Did you make sure you're running the compiler in the same directory as your shard.yml? (Crystal::CrystalPath::Error)
  from Crystal::Program#find_in_path<String, (String | Nil)>:Array(String)
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::TopLevelVisitor#visit<Crystal::Expressions>:Bool
  from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
  from Crystal::Program#top_level_semantic<Crystal::ASTNode+>:Tuple(Crystal::ASTNode+, Crystal::TypeDeclarationProcessor)
  from Crystal::Program#semantic<Crystal::ASTNode+, Bool>:Crystal::ASTNode+
  from __crystal_main
  from main

```

src/string.cr
```diff
@@ -2401,29 +2397,71 @@ class String
...
-    reader2 = Char::Reader.new(other)
-    ch1 = reader1.current_char
-    ch2 = reader2.current_char
+    if ascii_only? && other.ascii_only?
+      if self.size != other.size
+        others_first_char = other.at(0) do
+          return 1
+        end
+        if self.starts_with?(others_first_char)
+          if self.size > other.size
```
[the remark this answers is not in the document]
r00ster91: Yes, I'm just copying `String#<=>`'s behavior here.
When both strings start with the same character, then the length of both is the decisive factor.
`"aAA" <=> "aA"` returns `1` because the first string is longer.
`"Hzz" <=> "Hzzz"` returns `-1` because the latter string is shorter.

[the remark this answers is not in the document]
asterite: That's not correct. The strings are compared, char by char. If you end up with no chars in one of the strings then the other is "greater". Check the logic in the existing branch.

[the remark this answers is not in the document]
r00ster91: Oh, just copying the logic of the non-ASCII branch worked perfectly.

src/string.cr
```diff
@@ -2401,29 +2397,62 @@ class String
...
+        position += 1
+        byte1 = to_unsafe[position]
+        byte2 = other.to_unsafe[position]
+      end
 
-    if reader1.has_next?
-      1
-    elsif reader2.has_next?
-      -1
+      if position < bytesize
```
straight-shoota: This could just be `(bytesize - other.bytesize).sign` which is shorter and probably even more efficient (not that it matters for this).

src/string.cr
```diff
@@ -2401,29 +2397,62 @@ class String
...
+  def compare(other : String, case_insensitive = false, options = Unicode::CaseOptions::None)
     return self <=> other unless case_insensitive
 
-    reader1 = Char::Reader.new(self)
-    reader2 = Char::Reader.new(other)
-    ch1 = reader1.current_char
-    ch2 = reader2.current_char
+    if ascii_only? && other.ascii_only?
+      position = 0
+      byte1 = to_unsafe[0]
```
straight-shoota: Why not move this to the beginning of the loop to avoid duplicate code?

src/string.cr
r00ster91: Yes, it's a little bit faster, thanks!

src/string.cr
```diff
@@ -2401,29 +2397,55 @@ class String
...
+        return comparison unless comparison == 0
+
+        position += 1
+      end
 
-    if reader1.has_next?
-      1
-    elsif reader2.has_next?
-      -1
+      (bytesize - other.bytesize).sign
```
bcardiff: Isn't this the same as:

```suggestion
      bytesize <=> other.bytesize
````

?

File: src/string.cr [hunk 10]
```diff
@@ -2389,10 +2389,6 @@ class String
   # this string is less, equal or greater than *other*, optionally in a *case_insensitive*
   # manner.
   #
-  # If *case_insitive* is `false`, this method delegates to `<=>`. Otherwise,
```
Sija: This should be brought back.

r00ster91: Why is it useful to know that this method delegates to `<=>`? Also not just ASCII characters are compared in a case-insensitive way, all characters.

Sija: Fair 'nuff, good point.

bcardiff:
@r00ster91 would you mind doing a rebase on master so the CI can be happy?

bcardiff:
Thanks @r00ster91 🙇

PR #7353, opened 2019-01-31
Add exception overload to `raise`

j8r:
The goal of this PR is to add the possibility to add an optional cause to `raise`,  as allowed by `Exception.new message, cause` – object used in the implementation.

### DIFF
--- a/spec/std/raise_spec.cr
+++ b/spec/std/raise_spec.cr
@@ -13,7 +13,7 @@ describe "raise" do
   it "should reuse the previous callstack" do
     exception = expect_raises Exception, "with callstack" do
       begin
-        raise
+        raise ""
       rescue ex
         raise "with callstack", ex
       end
--- a/spec/std/raise_spec.cr
+++ b/spec/std/raise_spec.cr
@@ -10,6 +10,17 @@ describe "raise" do
     exception.callstack.should_not be_nil
   end
 
+  it "should reuse the previous callstack" do
+    exception = expect_raises Exception, "with callstack" do
+      begin
+        raise
+      rescue ex
+        raise "with callstack", ex
+      end
+    end
+    exception.callstack.should_not be_nil
+  end
+
   it "shouldn't overwrite the callstack on re-raise" do
     exception_after_reraise = expect_raises Exception, "exception to be rescued" do
       begin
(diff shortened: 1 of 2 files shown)


Sija:
`raise` without *message* argument was discussed earlier and rejected. Adding the *cause* argument seems pointless. What's the point of this PR?

j8r:
The nillable message can be removed, yes. What the point of `raise(message)` then, it's just an alias?
We can already do `raise Exception.new "message"` like in Python. Ruby compatibility isn't an aim of the language, doesn't it?

asterite:
Sorry, but this hasn't been discussed first. And I personally can't see a use case for it.

j8r:
Adding the cause is useful when raising on a rescue block. If it's useless, don't add it – it's an optional nillable

straight-shoota:
I don't see any benefit of `raise "message", cause` over `raise Exception.new("message", cause)`.

It's a little bit shorter, but expressiveness is better for understanding purpose. `raise` with string argument is only a short cut for raising *simple* exceptions.
When specifying a `cause`, you'd most likely use a specific exception type anyway.

j8r:
I don't quite get it: why having a half-baked shortcut? It doesn't hurt to improve it with an optional argument.

Otherwise, the `raise(message)` string overload isn't needed? In fact this is an exception we raise, not a string.

Sija:
> I don't quite get it: why having a half-baked shortcut? It doesn't hurt to improve it with an optional argument.

Because that's what _shortcuts_ are for? If they would be "fully-baked" they wouldn't be shortcuts anymore...

ysbaddaden:
I only consider `raise string` for reporting a bug with `raise "BUG: unexpected x y z"` or something that should be unreachable with `raise "unreachable"`. I fail to see the need for passing a cause exception, since I'd let the exception bubble, or just raise it again; can you enlighten me?

j8r:
I would like to raise the exception again yes, but I need to modify the message.

In my case, I do a task with subtasks which may have ones too. If one fails, it's reported on the top, with an user-friendly message like `main task failed: subtask failed: cannot download X`. Then, there is a cleanup on the main task.

Eventually, I would replace all this Exception raising by returning an Union of a custom Error type.

ysbaddaden:
> I would replace all this Exception raising by (...) a custom Error type.

Sounds like a perfect solution.

PR #7354, opened 2019-01-31
add DTLS support

stakach:
available in OpenSSL 1.0.2

### DIFF
--- a/src/openssl/lib_ssl.cr
+++ b/src/openssl/lib_ssl.cr
@@ -219,6 +219,8 @@ lib LibSSL
   {% if compare_versions(OPENSSL_VERSION, "1.0.2") >= 0 %}
     alias X509VerifyParam = LibCrypto::X509VerifyParam
 
+    fun dtls_method = DTLS_method : SSLMethod
+
     fun ssl_get0_param = SSL_get0_param(handle : SSL) : X509VerifyParam
     fun ssl_ctx_get0_param = SSL_CTX_get0_param(ctx : SSLContext) : X509VerifyParam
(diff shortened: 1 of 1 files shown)


asterite:
This needs to be wrapped and exposed (users never use lib directly), plus some test added.

stakach:
it works with the existing [OpenSSL Context](https://crystal-lang.org/api/0.27.0/OpenSSL/SSL/Context/Client.html#new%28method%3ALibSSL%3A%3ASSLMethod%3DContext.default_method%29-class-method) where you can pass in an alternative method.

So this is all that should be required. I'll add some tests.

asterite:
Hm, that should be changed...

stakach:
I quite like how everything is structured generally, it's very elegant in my opinion.
However I did hit a snag with testing.. DTLS works well as a client however I can't make it play nicely as a server with the current setup. Would probably need a`OpenSSL::SSL::Socket::DTLSServer` class to handle "connection management", which seems fairly complex.

You can see how I'm using DTLS as a client here: (which is super simple with the current class structure)
https://github.com/aca-labs/crystal-engine-driver/blob/master/src/engine-driver/transport/udp.cr#L52

The use case for me is communicating with IoT devices like [Zigbee dot dot](https://www.zigbee.org/zigbee-for-developers/dotdot/) and [Thread](https://www.threadgroup.org/)

So agree that just adding the method is a half-baked but I'll probably have time in the next 6 months to work on implementing the server side. Might see some IoT devices running crystal :)

asterite:
What I mean is that for now this change is probably fine, but we should eventually change the whole thing to expose those values outside of `lib` (an `alias` is usually enough).

PR #7356, opened 2019-02-01
Add Array#starts_with? and Array#ends_with?

delef:
Hello everyone 🙂

```
a = [1, 2, 3, 4]

a.starts_with? [1, 2] # => true
a.starts_with? [1] # => true
a.starts_with? [2] # => false
a.starts_with? 1 # => true
a.starts_with? 2 # => false

a.ends_with? [3, 4] # => true
a.ends_with? [4] # => true
a.ends_with? [3] # => false
a.ends_with? 4 # => true
a.ends_with? 3 # => false
```

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -1450,6 +1450,7 @@ class Array(T)
   # a.starts_with?(2) # => false
   # ```
   def starts_with?(el : T)
+    return false if empty?
     self[0] == el
   end
   
@@ -1471,6 +1472,7 @@ class Array(T)
   # a.ends_with?(2) # => false
   # ```
   def ends_with?(el : T)
+    return false if empty?
     self[-1] == el
   end
 
--- a/src/array.cr
+++ b/src/array.cr
@@ -1430,6 +1430,50 @@ class Array(T)
     self
   end
 
+  # Returns `true` when array starts with *other*, otherwise `false`.
+  #
+  # ```
+  # a = [1, 2, 3]
+  # a.starts_with?([1, 2]) # => true
+  # a.starts_with?([2, 3]) # => false
+  # ```
+  def starts_with?(other : Array)
+    return false if other.size > size
+    self[0, other.size] == other
+  end
+  
+  # Returns `true` when array starts with *element*, otherwise `false`.
+  #
+  # ```
+  # a = [1, 2, 3]
+  # a.starts_with?(1) # => true
+  # a.starts_with?(2) # => false
+  # ```
+  def starts_with?(el : T)
+    self[0] == el
+  end
+  
+  # Returns `true` when array ends with *other*, otherwise `false`.
+  # ```
+  # a = [1, 2, 3]
+  # a.ends_with?([2, 3]) # => true
+  # a.ends_with?([1, 2]) # => false
+  # ```
+  def ends_with?(other : Array)
+    return false if other.size > size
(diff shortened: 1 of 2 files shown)


src/array.cr
```diff
@@ -1430,6 +1430,52 @@ class Array(T)
...
   end
 
+  # Returns `true` when array starts with *other*, otherwise `false`.
+  #
+  # ```
+  # a = [1, 2, 3]
+  # a.starts_with?([1, 2]) # => true
+  # a.starts_with?([2, 3]) # => false
+  # ```
+  def starts_with?(other : Array)
```
j8r: It can even be `Enumerable`

src/array.cr
```diff
@@ -1430,6 +1430,52 @@ class Array(T)
...
+    self[0] == el
+  end
+  
+  # Returns `true` when array ends with *other*, otherwise `false`.
+  # ```
+  # a = [1, 2, 3]
+  # a.ends_with?([2, 3]) # => true
+  # a.ends_with?([1, 2]) # => false
+  # ```
+  def ends_with?(other : Array)
```
j8r: ditto, can be `Enumerable`

delef:
@j8r thanks!

konovod:
@Sija `Indexable`, maybe? `Enumerable` doesn't have `[]` method.

src/array.cr
```diff
@@ -1430,6 +1430,62 @@ class Array(T)
...
+  end
+  
+  # Returns `true` when array starts with *element*, otherwise `false`.
+  #
+  # ```
+  # a = [1, 2, 3]
+  # a.starts_with?(1) # => true
+  # a.starts_with?(2) # => false
+  # ```
+  def starts_with?(el : T)
```
r00ster91: Let's not use such confusing abbreviations, please. Also there's `element` in the description.
```suggestion
  def starts_with?(element : T)
```

Sija:
@konovod true but it has `each`, which is good enough. `Indexable` could be a candidate too though.

asterite:
Hey, sorry to interrupt everyone, but was this ever discussed? Please, create an issue discussing something.

I personally don't think this has enough use cases to merit going into the standard library. Additionally, the implementation is sub-optimal, creating an intermediate array.

I vote to close this.

delef:
@asterite I can move this into Enumerable. Thus, I will try to optimize avoiding an intermediate array.

asterite:
@delef Thank you, but I don't think we are going to merge this. It's not in Ruby and I can't see the need for it.

src/array.cr
```diff
@@ -1430,6 +1430,62 @@ class Array(T)
...
+  
+  # Returns `true` when array ends with *element*, otherwise `false`.
+  # ```
+  # a = [1, 2, 3]
+  # a.ends_with?(3) # => true
+  # a.ends_with?(2) # => false
+  # ```
+  def ends_with?(element : T)
+    return false if empty?
+    self[-1] == element
```
r00ster91: You can use the unsafe `unsafe_fetch` here since it won't be empty here. Ditto above.

Sija:
@delef would you mind posting some real-world use-cases for these? IMO the best ones would be taken directly from stdlib, where we could see some improvement in terms of code complexity, verbosity, etc. Otherwise value of this addition stays pretty theoretical (i.e. code bloat).

PR #7366, opened 2019-02-03
make wait_readable and wait_writable public

stakach:
These methods are useful for integrating 3rd party libraries that require you to wait for a readable or writable socket.
Example [libssh2](https://github.com/libssh2/libssh2/blob/master/example/ssh2_exec.c)
How this is implemented in crystal: https://github.com/spider-gazelle/ssh2.cr/blob/a93eb957420b9e9f31bb8cd46e50ff862c5bc54b/src/session.cr#L39-L49

### DIFF
--- a/src/io/syscall.cr
+++ b/src/io/syscall.cr
@@ -103,10 +103,12 @@ module IO::Syscall
     end
   end
 
+  # :nodoc:
   def wait_readable(timeout = @read_timeout)
     wait_readable(timeout: timeout) { |err| raise err }
   end
 
+  # :nodoc:
   def wait_readable(timeout = @read_timeout)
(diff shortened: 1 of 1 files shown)


stakach:
@asterite how should one integrate libssh2 to prevent it blocking the event loop?
Using those methods definitely works to achieve what is required for libssh2

asterite:
How is it done in Go?

asterite:
Actually, nevermind, I have no idea what I am talking about :blush:

stakach:
haha yeah, looks like they have a native golang implementation..

ysbaddaden:
Can we make them `:nodoc:`? It would allow to use them, but make no guarantee these methods will continue to exist.

PR #7371, opened 2019-02-04
http client: Fallback to default mime type 'application/octet-stream'

### LINKED ISSUE #7370: MIME parsing raises exception on empty Content-Type response
We have a file on AWS S3 that does not have a Content-Type set when downloaded:

```
$ curl -v <redacted>
< HTTP/1.1 200 OK
< x-amz-id-2: <redacted>
< x-amz-request-id: <redacted>
< Date: Mon, 04 Feb 2019 09:54:10 GMT
< Last-Modified: Tue, 29 Jan 2019 12:43:53 GMT
< ETag: <redacted>
< x-amz-server-side-encryption: AES256
< Accept-Ranges: bytes
< Content-Type:
< Content-Length: 1073741824
< Server: AmazonS3
```
When trying to fetch it, the MIME parser chokes on it:
```
  from /usr/share/crystal/src/mime/media_type.cr:0:25 in 'parse'
  from /usr/share/crystal/src/hash.cr:244:13 in 'check_content_type_charset'
  from src/copy_upstream.cr:63:9 in 'copy_upstream_single_without_resume'
  from src/copy_upstream.cr:53:19 in '???'
  from src/cr_sarkar_proxy.cr:150:5 in '->'
  from lib/kemal/src/kemal/route.cr:12:9 in '->'
  from lib/kemal/src/kemal/route_handler.cr:255:3 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:24:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:24:7 in 'call_next'
  from lib/kemal/src/kemal/static_file_handler.cr:67:9 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:24:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:24:7 in 'call'
  from /usr/share/crystal/src/http/server/handler.cr:24:7 in 'call'
  from /usr/share/crystal/src/http/server/request_processor.cr:39:11 in 'process'
  from /usr/share/crystal/src/fiber.cr:255:3 in '???'
  from ???
```
It's not really obvious from the stack trace, but this happens in a `HTTP::Client.get` call to GET the file from S3.

This happens on version 0.27.1:
```
$ crystal -v
Crystal 0.27.1 [64137d045] (2019-01-30)

LLVM: 4.0.0
```

If a MIME type can not be determined, we should use `application/octet-stream`, [since it is the default mime type](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types) to use if the type cannot be determined. While this is arguably a bug in S3 rather than crystal, (we should be conservative in what we send and liberal in what we accept.)[https://en.wikipedia.org/wiki/Robustness_principle] Also, `curl` downloads the file correctly even with the empty `Content-Type` header.

bew:
Fixes #7370 

Uses a default mime type as a fallback when the `Content-Type` header cannot be parsed.

I'm not sure about the `DEFAULT_MIME_TYPE` as a constant, should the user be able to customize the default mime type?

### DIFF
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -74,12 +74,16 @@ module HTTP
   private def self.check_content_type_charset(body, headers)
     return unless body
 
-    if content_type = headers["Content-Type"]?
-      mime_type = MIME::MediaType.parse(content_type)
-      if charset = mime_type["charset"]?
-        body.set_encoding(charset, invalid: :skip)
-      end
-    end
+    content_type = headers["Content-Type"]?
+    return unless content_type
+
+    mime_type = MIME::MediaType.parse?(content_type)
+    return unless mime_type
+
+    charset = mime_type["charset"]?
+    return unless charset
+
+    body.set_encoding(charset, invalid: :skip)
   end
 
   # :nodoc:
--- a/spec/std/http/client/client_spec.cr
+++ b/spec/std/http/client/client_spec.cr
@@ -3,13 +3,13 @@ require "openssl"
 require "http/client"
 require "http/server"
 
-private def test_server(host, port, read_time = 0)
+private def test_server(host, port, read_time = 0, content_type = "text/plain")
   server = TCPServer.new(host, port)
   begin
     spawn do
       io = server.accept
       sleep read_time
-      response = HTTP::Client::Response.new(200, headers: HTTP::Headers{"Content-Type" => "text/plain"}, body: "OK")
+      response = HTTP::Client::Response.new(200, headers: HTTP::Headers{"Content-Type" => content_type}, body: "OK")
       response.to_io(io)
       io.flush
     end
@@ -209,6 +209,13 @@ module HTTP
       test_server("localhost", 0, 0) do |server|
         client = Client.new("localhost", server.local_address.port)
         client.connect_timeout = 0.5
+        client.get("/")
+      end
+    end
+
+    it "tests empty Content-Type" do
+      test_server("localhost", 0, content_type: "") do |server|
+        client = Client.new("localhost", server.local_address.port)
         client.get("/")
       end
     end


WJWH:
Whoa, I had just cloned the repo and opened up my editor but you already made a fix! That is less than 45 minutes from issue creation to pr submission ❤️

src/http/common.cr
```diff
@@ -71,11 +71,13 @@ module HTTP
...
   end
 
+  DEFAULT_MIME_TYPE = MIME::MediaType.parse("application/octet-stream")
+
   private def self.check_content_type_charset(body, headers)
     return unless body
 
     if content_type = headers["Content-Type"]?
-      mime_type = MIME::MediaType.parse(content_type)
+      mime_type = MIME::MediaType.parse?(content_type) || DEFAULT_MIME_TYPE
```
straight-shoota: `mime_type` is not used anywhere except for querying the `charset` parameter (which isn't set in `application/octet-stream` anyway). So the following condition should simply be skipped when `parse`?  returns `nil`.

bew: ok, so something like that?
```cr
    if (content_type = headers["Content-Type"]?) &&
       (mime_type = MIME::MediaType.parse?(content_type)) &&
       (charset = mime_type["charset"]?)
      body.set_encoding(charset, invalid: :skip)
    end
```

Sija: Why not just `return unless mime_type`?

asterite: Yes. In fact we should `return unless` a lot more here, to avoid nested code.

bew:
Squashed!

bcardiff:
Thanks @bew

straight-shoota:
Honestly, I really don't like the change to unnest all conditions from `if` to `return unless`. IMHO that makes it harder to follow control flow. I'm definitely in favour of using `return unless` more often, but having it in every second line is simply confusing.

Obviously, this is a matter of stylistic preference. But the agreement is not to change from one style to another for no semantic reason.

PR #7373, opened 2019-02-04
Fix: integer overflow in main thread stack base detection

### LINKED ISSUE #7368: Crystal 0.27.1 built with --release crashes immediately with OverflowError
If I compile 0.27.1 (or master) with `--release -Dpreview_overflow`, then I can't compile anything anymore, not even Crystal itself. The compiler crashes *immediately*:

```
$ make -B
CRYSTAL_CONFIG_BUILD_COMMIT="0ef090a" ./bin/crystal build -D preview_overflow -D compiler_rt --release  -o .build/crystal src/compiler/crystal.cr -D without_openssl -D without_zlib
Unhandled exception: Overflow (OverflowError)
[0xd30ca6] ???
[0x417cfb] __crystal_raise +43
[0x429963] ???
[0x422979] ???
[0x41f388] ???
[0x4522bd] ???
[0x41bdbd] main +45
[0x2acaf91c0f45] __libc_start_main +245
[0x41416f] ???
[0x0] ???
make: *** [.build/crystal] Erreur 5
```

Note that I can build a working compiler with `-Dpreview_overflow` only (without `--release`).

All my Travis jobs are currently broken, since the Linux packages of 0.27.1 have been built with these options.

OS: Ubuntu 14.04
Target: x86_64-linux-gnu

### LINKED ISSUE #7369: OverflowError always lead to can't raise END_OF_STACK
Happens systematically on Ubuntu 14.04 / x86_64-linux-gnu for me.

It seems Debug::ELF and/or Debug::DWARF are trying to raise an exception but fail to —because we're already inside an exception handler?

Yet, compiling with `--no-debug` has no impact.

ysbaddaden:
Using `getrlimit` led to invalid STACK SIZE limits in some cases which in turn led to an integer overflow while detecting the stack base of the main stack. This patch replaces `getrlimit` with proper, but non portable, calls to pthread functions that return the stack of a given thread.

Note: only tested on x86_64-linux-gnu. I didn't test on macOS, FreeBSD or OpenBSD.

fixes #7368
fixes #7369


src/lib_c/x86_64-openbsd/c/signal.cr
```diff
@@ -42,6 +42,12 @@ lib LibC
   SIG_DFL = SighandlerT.new(Pointer(Void).new(0_u64), Pointer(Void).null)
   SIG_IGN = SighandlerT.new(Pointer(Void).new(1_u64), Pointer(Void).null)
 
+  struct StackT
+    *ss_sp : Void*;
```
Sija: Redundant `;`, ditto below.

ysbaddaden: oops, C header copy-paste :)

Sija: C C (si si) ;)

src/thread.cr
```diff
@@ -135,4 +135,42 @@ class Thread
       Fiber.inactive(fiber)
     end
   end
+
+  private def stack_address : Void*
+    address = Pointer(Void).null
+
+    {% if flag?(:darwin) %}
+      address = LibC.pthread_get_stackaddr_np(@th) - pthread_get_stacksize_np(@th)
```
bcardiff: In https://github.com/crystal-lang/crystal/pull/6970#issuecomment-431652079 you said that these functions return bogus values. Any new insight regarding this?

bcardiff:
```suggestion
      address = LibC.pthread_get_stackaddr_np(@th) - LibC.pthread_get_stacksize_np(@th)
```

ysbaddaden: Urgh, I totally forgot about #6970 :/

To be honest, I don't know. Rust does just that, and doesn't care to check the main thread (when it does for OpenBSD). BDWGC doesn't need `pthread_get_stacksize_np`. I'm not sure what should be done, here.

ysbaddaden: It seems it was fixed in macOS 10.10

ysbaddaden: Here is how openjdk does it:
http://hg.openjdk.java.net/jdk8/jdk8/hotspot/file/87ee5ee27509/src/os_cpu/bsd_x86/vm/os_bsd_x86.cpp#l836

bcardiff: Ok, from https://github.com/rust-lang/rust/issues/43347#issuecomment-316783599 and https://github.com/dmlloyd/openjdk/blob/2f680fec58a8ffb818a48f542e493306d1113c69/src/hotspot/os_cpu/bsd_x86/os_bsd_x86.cpp#L902-L961 lets move forward.

Also, the osx binaries are targeted for `>= 10.11`.

ysbaddaden: Sadly I can't just take their solution, it's licensed under the GPLv2.

ysbaddaden:
CI should pass now :crossed_fingers:

bcardiff:
Manually checked in macOS & FreeBSD. :shipit:

bcardiff:
Thanks @ysbaddaden  💯 👑

oprypin:
Presumably this will require a new release?

bcardiff:
Yes @oprypin we will release 0.27.2 tomorrow probably with the fixed indicated in that milestone.

PR #7374, opened 2019-02-04
 Add `options` argument to String#camelcase and String#underscore

r00ster91:
That are the only methods that upcase or downcase a string internally and don't have an `options` argument. Because what if you speak Turkish and want to camelcase `"interesting_integer"` to `"İnterestingİnteger"` instead of `"InterestingInteger"`? And `String#capitalize` has it too so I don't see why these shouldn't have it.

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -913,7 +913,7 @@ class String
   def downcase(options = Unicode::CaseOptions::None)
     return self if empty?
 
-    if ascii_only?
+    if ascii_only? && (options.none? || options.ascii?)
       String.new(bytesize) do |buffer|
         bytesize.times do |i|
           buffer[i] = to_unsafe[i].unsafe_chr.downcase.ord.to_u8
--- a/src/string.cr
+++ b/src/string.cr
@@ -3382,7 +3382,7 @@ class String
   # "HTTP_CLIENT".underscore            # => "http_client"
   # "3.14IsPi".underscore               # => "3.14_is_pi"
   # ```
-  def underscore
+  def underscore(options = Unicode::CaseOptions::None)
     first = true
     last_is_downcase = false
     last_is_upcase = false
@@ -3391,12 +3391,12 @@ class String
 
     String.build(bytesize + 10) do |str|
       each_char do |char|
-        digit = '0' <= char <= '9'
-        downcase = 'a' <= char <= 'z' || digit
-        upcase = 'A' <= char <= 'Z'
+        digit = char.ascii_number?
+        downcase = char.ascii_lowercase? || digit
+        upcase = char.ascii_uppercase?
 
         if first
-          str << char.downcase
+          str << char.downcase(options)
         elsif last_is_downcase && upcase
           if mem
             # This is the case of A1Bcd, we need to put 'mem' (not to need to convert as downcase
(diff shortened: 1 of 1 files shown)


src/string.cr
```diff
@@ -913,7 +913,7 @@ class String
   def downcase(options = Unicode::CaseOptions::None)
     return self if empty?
 
-    if ascii_only? && (options.none? || options.ascii?)
+    if ascii_only?
```
ysbaddaden: Are you sure about this one?

r00ster91: I removed this from `String#upcase`, `String#capitalize` and from `String#downcase` and all specs failed except for this method. And from all three this one actually has the most specs: [`spec/std/string_spec.cr:476`](https://github.com/crystal-lang/crystal/blob/master/spec/std/string_spec.cr#L476). It especially has many that use the `options` argument. But I could maybe add some more specs to `String#downcase` to be 100 percently sure.

ysbaddaden: I prefer to be sure because shaving 2 integer checks after `ascii_only?` that iterated the whole string should have a negligible impact on performance.

r00ster91: Alright there's no difference anyway.
```
    downcase 300.45  (  3.33ms) (±14.33%)  350256 B/op        fastest
new_downcase 286.03  (   3.5ms) (±15.35%)  350257 B/op   1.05× slower
```

wontruefree:
Can you update the documentation as well?

r00ster91:
I added two turkic examples and fixed `String#underscore` to properly work with case options.

src/string.cr
```diff
@@ -3377,12 +3377,13 @@ class String
...
-  # "HTTP_CLIENT".underscore            # => "http_client"
-  # "3.14IsPi".underscore               # => "3.14_is_pi"
+  # "DoesWhatItSaysOnTheTin".underscore                         # => "does_what_it_says_on_the_tin"
+  # "PartyInTheUSA".underscore                                  # => "party_in_the_usa"
+  # "HTTP_CLIENT".underscore                                    # => "http_client"
+  # "3.14IsPi".underscore                                       # => "3.14_is_pi"
+  # "InterestingImage".underscore(Unicode::CaseOptions::Turkic) # => "ınteresting_ımage"
   # ```
-  def underscore
+  def underscore(options = Unicode::CaseOptions::None)
```
j8r: You can add a type annotation

src/string.cr
```diff
@@ -3442,16 +3449,17 @@ class String
...
 
   # Converts underscores to camelcase boundaries.
   #
   # ```
-  # "eiffel_tower".camelcase # => "EiffelTower"
+  # "eiffel_tower".camelcase                                   # => "EiffelTower"
+  # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase
+  def camelcase(options = Unicode::CaseOptions::None)
```
j8r: Type annotation here too

src/string.cr
straight-shoota: I'd suggest to add it because it improves documentation. Readers can directly follow the type link to see available values.
And more importantly, it allows to use autocasted symbols for the enum members. This only works when the argument is typed as an enum type. So you can write `"InterestingImage".underscore(:turkic)`.

r00ster91: Ooooh... I completely forgot about that awesome feature somehow! That's a good reason to add it.

PR #7376, opened 2019-02-04
CI: Avoid duplicate runs in maintenance branches

bcardiff:
The maintenance plan runs also the test. This will reduce a bit the load in the CI


bcardiff:
Closing. It needs to be sent from origin instead of fork to run fully.

PR #7378, opened 2019-02-04
CI: Test linux 64 bits package using docker image recent build

bcardiff:
This PR will run the specs using the distributable linux 64 package inside the docker image built.

For nightlies and maintenance releases it uses de same docker image that is pushed to docker hub.
For tagged release it uses a docker image that was already built but that is not pushed since the binaries are not signed.

This would have caught some of the regressions in 0.27.1


bcardiff:
Will open again from origin instead of fork

PR #7379, opened 2019-02-05
spawn: give compile error if passing an exp and a block

### LINKED ISSUE #7365: Named fibers seem subtly broken
#1284

```
spawn               do p :hi end
spawn "hey"         do p :hey end
spawn name: "hello" do p :hello end
sleep 1
```

```
$ ./test.cr
:hi
:hello
```

This was a real WTF moment for me. Do I understand correctly that any positional arguments to `spawn` will silently prevent it from running? Why?

:thinking: is it because of `spawn` macro?

asterite:
Fixes #7365

### DIFF
--- a/src/concurrent.cr
+++ b/src/concurrent.cr
@@ -94,7 +94,11 @@ end
 # This is because in the first case all spawned fibers refer to
 # the same local variable, while in the second example copies of
 # *i* are passed to a `Proc` that eventually invokes the call.
-macro spawn(call, *, name = nil)
+macro spawn(call, *, name = nil, &block)
+  {% if block %}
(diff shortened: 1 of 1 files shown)


src/concurrent.cr
```diff
@@ -94,7 +94,11 @@ end
 # This is because in the first case all spawned fibers refer to
 # the same local variable, while in the second example copies of
 # *i* are passed to a `Proc` that eventually invokes the call.
-macro spawn(call, *, name = nil)
+macro spawn(call, *, name = nil, &block)
+  {% if block %}
+    {% raise "`spawn(call)` can't be invoked with a block, did you mean `spawn(name: ...) { ... }" %}
```
Sija:
```suggestion
    {% raise "`spawn(call)` can't be invoked with a block, did you mean `spawn(name: ...) { ... }`?" %}
```

Issue #7285, opened 2019-01-07
Remove None and All default values for flag enums.


straight-shoota:
Continuation from https://github.com/crystal-lang/crystal/issues/7270#issuecomment-451947228

With most flag enums the semantics of `None` and `All` probably make sense. That's why they were added as defaults.

But they're not necessarily always semantically valid. When some of the flags are mutually exclusive, `All` is not possible. When setting at least one flag is required, `None` doesn't make sense either.

Because of this they better shouldn't be added to every flag enum.

Autogeneration can simply be removed. Where appropriate, these members can be added manually as `None = 0` and `All = ?` (would it be `Int32::MAX`?).

Another option could be to configure the generation of these special enum members via the `Flags` annotation. For example, `@[Flags(all: false)]` would not autogenerate a `None` member.
We could keep the current semantics by default (plain `@[Flags]` annotation generates both special members) but allow to opt-out when the autogenerated values are not applicable. This is reasonable when most flag enums use `All` and `None` members (which I assume, but it's not based on evidence).
This solution adds more complexity to the compiler and another feature to the language. The alternative (remove + manual creation) is easier to follow because there is only one syntax and it's explicit.

Considering the highest desired state where each enum member would be documented with a individual description, autogeneration (in either form) is not useful because it can't generate meaningful documentation anyway.

Thus, I think I would prefer to remove autogeneration completely and require explicit declaration.

N.B. If we keep autogeneration in any form, there should be a fix to ensure autogenerated members are visible in the API docs. The documentation needs to be manually overridable.

bcardiff:
The point of the autogeneration is

1. to avoid doing all the bitwise-or with values or manually calculate them (All)
2. ensure there is always a 
3. to have a clear semantics to exclude All and None constants as members in `Enum.each` and `Enum#each` and in the compiler (`num_members` in `TopLevelVisitor`)

In order to keep that we would need to go with the option of `@[Flags(all: false)]` or settle that All/None will be constants like the others and might be less fun to use because the All and None will be members.

If we allow `@[Flags(all: false)]` we can also allow `@[Flags(all: ALL)]` to allow custom names btw.

I'm not voting for anyone yet, just dumping thoughts.

bew:
@asterite are thinking about something like this?
```cr
enum Foo
  A
  B
  MyAll = all
  
  def self.all
    # generated with macros.. 
    # but how to detect that
    # MyAll should be skipped? 
    A | B
  end
end

p Foo::MyAll
[...]

PR #7382, opened 2019-02-06
enum parse macro to no longer parse virtual `None` or `All` members

stakach:
Using crystal 0.27.2 on OSX
When I execute the following code:

```crystal
OpenSSL::SSL::VerifyMode.parse("peer")
```

I get `duplicate when "none" in case`:
image

However looking the [definition](https://github.com/crystal-lang/crystal/blob/master/src/openssl/lib_ssl.cr#L52) of that enum:
```crystal
  @[Flags]
  enum VerifyMode : Int
    NONE                 = 0
    PEER                 = 1
    FAIL_IF_NO_PEER_CERT = 2
    CLIENT_ONCE          = 4
  end
```

The macro that builds `.parse?` needs to ignore the built in `None` and `All` members for this code to compile.

### DIFF
--- a/src/enum.cr
+++ b/src/enum.cr
@@ -405,6 +405,7 @@ struct Enum
   # the enum members names, so a member named "FourtyTwo" or "FOURTY_TWO"
   # is found with any of these strings: "fourty_two", "FourtyTwo", "FOURTY_TWO",
   # "FOURTYTWO", "fourtytwo".
+  # It won't parse the `None` and `All` members of flags enums.
   #
   # ```
   # Color.parse?("Red")    # => Color::Red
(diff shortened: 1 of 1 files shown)


asterite:
I think we should actually disallow `NONE` as an enum member (we already disallow `None`). Let me fix that, and then this will be trivially fixed (though it will be a breaking change).

I think `None` and `All` should still be parsed...

asterite:
Actually, the problem will still happen if we have constants like `Foo` and `FOO`. The solution here is to not generate duplicate `when` branches. Unfortunately we don't have a `uniq` macro that accepts a block... but maybe it's possible to do, somehow

asterite:
Also, let's remove that `NONE` constant from there, it's redundant.

ysbaddaden:
I think we shouldn't inject our own constants into Enums. It's acceptable for Crystal enums (documented behavior), but awful for enums to external bindings :disappointed:

asterite:
> I think we shouldn't inject our own constants into Enums

Period :-)

Please, please, let's remove these generated enum constants, they make the code complex (many special cases to consider) and lead to bugs. One can simply write `None = 0` or `NONE = 0` and that's it. `All` is a bit more troublesome to define but it's not a big deal and it's not always needed.

bew:
@asterite @ysbaddaden related to #7285

PR #7384, opened 2019-02-06
Compiler: always mention the scope (and with ... yield scope, if any) on undefined method error

### LINKED ISSUE #7357: `with` . `yield` compilation error should mention current receiver
While writing a DSL using `with` ... `yield`, I ran into a compilation error which took me a long time to fix because the actual receiver was not mentioned in the error (which turned out to be nil).

A reduced example:

```crystal
def make_object
  "test"
  puts "This was my error"
end

def doit
	with make_object() yield
end

doit {
  size
}
```

Which gave the following error:

```
Error in line 10: instantiating 'doit()'

in line 10: instantiating 'doit()'

in line 11: undefined method 'size'
```

I was surprised by this error because the string object should implement the `size` method. Of course my error would have been much easier to discover if the error message would have included the fact that the receiver was of class `Nil`.

Contrast this with:

```crystal
def make_object
  "test"
  puts "this was my error"
end

make_object().size
```
giving the more helpful:

```
Error in line 7: undefined method 'size' for Nil

Rerun with --error-trace to show a complete error trace.
```

asterite:
Fixes #7357

For the code in the issue:

```crystal
def make_object
  "test"
  puts "This was my error"
end

def doit
	with make_object() yield
end

doit {
  size
}
```

You now get this error:

```
Error in baz.cr:11: instantiating 'doit()'

doit {
^~~~

in baz.cr:11: instantiating 'doit()'

doit {
^~~~

in baz.cr:12: undefined local variable or method 'size' for Nil (for `with ... yield`) and top-level (current scope)

  size
  ^~~~
```

The error now mentions both scopes (in order) where the method was searched.

Additionally, the scope is now always mentioned to make it easy to understand what went wrong. For example, for this code:

```crystal
hello
```

Before: `undefined local variable or method 'hello'`
After: `undefined local variable or method 'hello' for top-level`

Similarly, for this code:

```crystal
class Foo
  def foo
    hello
  end
end

Foo.new.foo
```

Before: `undefined local variable or method 'hello'`
After: `undefined local variable or method 'hello' for Foo`

Or...

```crystal
class Foo
  def self.foo
    hello
  end
end

Foo.foo
```

Before: `undefined local variable or method 'hello'`
After: `undefined local variable or method 'hello' for Foo.class` ("Ooooh... I totally forgot I was in a class method!")

### DIFF
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -92,21 +92,36 @@ class Crystal::Call
       similar_name = owner.lookup_similar_def_name(def_name, self.args.size, block)
 
       error_msg = String.build do |msg|
-        if obj && owner != program
-          msg << "undefined method '#{def_name}' for #{owner}"
-        elsif convert_to_logical_operator(def_name)
-          msg << "undefined method '#{def_name}'"
-          similar_name = convert_to_logical_operator(def_name)
+        if obj
+          could_be_local_variable = false
+        elsif logical_op = convert_to_logical_operator(def_name)
+          similar_name = logical_op
+          could_be_local_variable = false
         elsif args.size > 0 || has_parentheses?
-          msg << "undefined method '#{def_name}'"
+          could_be_local_variable = false
         else
+          # This check is for the case `a if a = 1`
           similar_name = parent_visitor.lookup_similar_var_name(def_name) unless similar_name
           if similar_name == def_name
-            # This check is for the case `a if a = 1`
-            msg << "undefined method '#{def_name}'"
+            could_be_local_variable = false
           else
-            msg << "undefined local variable or method '#{def_name}'"
+            could_be_local_variable = true
           end
+        end
+
+        if could_be_local_variable
+          msg << "undefined local variable or method '#{def_name}'"
+        else
+          msg << "undefined method '#{def_name}'"
+        end
+
+        owner_name = owner.is_a?(Program) ? "top-level" : owner.to_s
+        with_scope = @with_scope
+
+        if with_scope && !obj && with_scope != owner
+          msg << " for #{with_scope} (with ... yield) and #{owner_name} (current scope)"
+        else
+          msg << " for #{owner_name}"
         end
 
         if def_name == "allocate" && owner.is_a?(MetaclassType) && owner.instance_type.module?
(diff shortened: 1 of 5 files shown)


bew:
For: `undefined local variable or method 'hello' for Foo`

We should mention `top-level` too?

asterite:
@bew I think at that point it will be too confusing... the `top-level` is always there. Plus if `Foo` inherited some other class or included some other modules, should we mention all the hierarchy?

Or maybe yes... but if we go that way, I'd actually like to change all error messages to be more friendly and comprehensive, kind of like what Elm does.

bew:
> @bew I think at that point it will be too confusing... the `top-level` is always there. Plus if `Foo` inherited some other class or included some other modules, should we mention all the hierarchy?

Indeed, let's keep it that way for now.

mpcjanssen:
@asterite is the `Int32` in the initial comment an example from a different piece of code? I would expect the error here to reference nil (the result of `make_object`:
```
in baz.cr:12: undefined local variable or method 'size' for Nil (for `with ... yield`) and top-level (current scope)
```

asterite:
@mpcjanssen Oops, you are right. It actually prints `Nil` instead of `Int32`. That was from some other example I had.

asterite:
I added a second commit to not use `colorize` in compiler specs because it was making testing the messages a bit tedious and error-prone.

src/compiler/crystal/semantic/call_error.cr
```diff
@@ -92,23 +92,38 @@ class Crystal::Call
...
+          msg << "undefined local variable or method '#{def_name}'"
+        else
+          msg << "undefined method '#{def_name}'"
+        end
+
+        owner_name = owner.is_a?(Program) ? "top-level" : owner.to_s
+        with_scope = @with_scope
+
+        if with_scope && !obj && with_scope != owner
+          msg << " for #{with_scope} (with ... yield) and #{owner_name} (current scope)"
```
straight-shoota: Is `current scope` actually a proper term? Could maybe be lexical or static scope...

asterite: I don't know, I'm not a native English speaker. Write a suggestion and I'll accept it.

straight-shoota: No, the language is fine. I'm just wondering if it's the best technical description to be used here.

asterite: I don't think we mention "scope" anywhere else. I don't know what to call this. Maybe it would be better to just restrict `with ... yield` to a single scope, then we wouldn't even have to think about this.

straight-shoota: I guess it's fine as it is.

PR #7395, opened 2019-02-08
[DO NOT MERGE] Let's debug the compiler in the CI

asterite:
I'll be sending a couple of updates to the PR in order, which is on top of the `begin/end-less ranges` which is currently failing in the CI, in order to find out why it's failing.

As a first step I'll just compile the compiler and then try to run the std specs with it.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -73,9 +73,10 @@ prepare_system() {
 }
 
 build() {
+  with_build_env 'make crystal'
   with_build_env 'make std_spec clean'
-  with_build_env 'make crystal std_spec compiler_spec docs'
-  with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
+  # with_build_env 'make crystal std_spec compiler_spec docs'
+  # with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
 }
 
 format() {
--- a/bin/ci
+++ b/bin/ci
@@ -73,10 +73,9 @@ prepare_system() {
 }
 
 build() {
-  with_build_env 'make crystal'
-  with_build_env 'make std_spec clean'
-  # with_build_env 'make crystal std_spec compiler_spec docs'
-  # with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
+  with_build_env 'make std_spec clean threads=1'
+  with_build_env 'make crystal std_spec compiler_spec docs threads=1'
+  with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
 }
 
 format() {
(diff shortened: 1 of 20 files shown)


asterite:
Now I'm almost sure this problem has to do with:
- `fork` having some bug
- CI reaching out of memory inside `fork`

For some reason the output isn't shown sometimes... I added debug output and now it's shown and you can clearly see the "channel closed" errors and the "out of memory" errors.

asterite:
One explanation can be that we have more and more std specs and that needs more and more memory to compile everything. One solution, for CI, is to compile the std specs by groups, for example primitives in one pass, array and hash in other, json in other and son on. This is really just a CI problem that we can solve with these changes, it only affects the compiler. Probably anyone with a decent computer can compile Crystal and std specs on their computer just fine.

asterite:
Another solution: we can pass `--threads 1` (or some other numbers smaller than the current one). I tried using just one `fork` and it worked well. It might be that when we fork we duplicate a lot of memory somehow and it eventually brings down the whole process. That will make CI just a bit slower because compiling is not the bottleneck, running the specs is.

asterite:
I just tried passing `--threads 1` and it worked.

We should probably try that in the existing PRs that are failing, and of course we can improve the compiler a bit by special-casing that case (there's no need to `fork` nor `spawn` at all).

I'm not saying that this should be a definitive solution. We should definitely improve the situation in some other way. But this way will let us continue developing other parts of Crystal without the CI blocking us or giving us false positives. (we should track this problem in a separate issue).

Thoughts?

asterite:
Okay, I lied. It didn't work. It did work for the first compilation of the std specs but not for the second one... for some reason. But I'm pretty sure the reason is "out of memory" in `fork`.

asterite:
Okay, that worked. No fork, spawn or Channel now. We could probably go with this solution for now (not with this code, I have to special case it for one thread).

straight-shoota:
Maybe this is related to #6961?

I still doubt that compiling the std specs could fill up 7,5 GB of memory unless there is some kind of memory leak. I'm pretty sure it used to compile with far less memory requirements.

ysbaddaden:
That doesn't surprises me, `fork` shouldn't copy memory at first, but will whenever the parent or child process writes to a memory page. If CI has no swap, disables overcommit, or something like that, then `fork` will crash.

Maybe we can use `--threads=0` to disable spawn/fork, and make it the default on CI? That looks acceptable to me.

asterite:
> That doesn't surprises me, `fork` shouldn't copy memory at first, but will whenever the parent or child process writes to a memory page. If CI has no swap, disables overcommit, or something like that, then `fork` will crash.

Exactly.

> Maybe we can use `--threads=0` to disable spawn/fork, and make it the default on CI? That looks acceptable to me.

Just `--threads=1` should be enough. If there's one "thread" there's no need to fork/spawn at all. I'll write a PR for that.

I'm still not satisfied with the memory usage the compiler has, but for now there's not much we can do other than fixing things for the CI to work a bit better.

asterite:
Nothing else to do here.

Issue #7243, opened 2018-12-31
Rarely failing nil assertion in signal_spec


straight-shoota:
```
Nil assertion failed (Exception)
  from src/class.cr:148:0 in 'not_nil!'
  from spec/std/signal_spec.cr:0:23 in '->'
  from src/signal.cr:255:3 in '->'
  from src/signal.cr:255:3 in 'process'
  from src/signal.cr:198:9 in '->'
  from src/fiber.cr:255:3 in 'run'
  from src/fiber.cr:29:34 in '->'
  from ???
FATAL: uncaught exception while processing handler for CHLD, exiting
```

It seems to be a race condition in the `Singal::CHLD` handler and happens very rarely.

I' might have seen this error more often, but here are at least two instances where it failed in unrelated CI runs:

* [circleci/test_linux](https://circleci.com/gh/crystal-lang/crystal/11987)
* [travis-ci/linux64](https://travis-ci.org/crystal-lang/crystal/jobs/424542591) (unfortunately, travis doesn't seem to keep the logs of a job after it has been restarted)

straight-shoota:
Another occurrance: https://circleci.com/gh/crystal-lang/crystal/18298?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

asterite:
I GOT IT!!! 💡 

This is the spec:

```crystal
it "CHLD.trap is called after default Crystal child handler" do
  called = false
  child = nil

  Signal::CHLD.trap do
    called = true
    Process.exists?(child.not_nil!.pid).should be_false
  end

  child = Process.new("true", shell: true)
  child.not_nil!.wait # doesn't block forever
  called.should be_true
end
```

TIL I learned about SIGCHLD: it's trapped when a child process **ends**.

The spec fails because of a `not_nil!` (the trace isn't entirely correct with the filenames and line numbers). But my guess was that `child.not_nil!` is raising. How could that happen?

Well, if `Process.new("true", shell: true)` runs and ends **before** it gets assigned to `child`, which is super rare, then the handler will be invoked and `child` will still be `nil` ⚠️ 

This race condition can be reproduced if you change the `child` assignment to:

```crystal
child = begin
  sleep 1
  Process.new("true", shell: true)
end
```

Now to fix it. I guess if `child` is `nil` I'll just `Fiber.yield` until `child` gets a non-`nil` value. It's a bit hard fixing this because something happens before an expression gets assigned to a variable 🤔

PR #7402, opened 2019-02-09
HTTP::Client spec: don't write server response if expecting read timeout

### LINKED ISSUE #7401: "Unhandled exception in spawn: Error writing to socket: Broken pipe (Errno)" in socket client specs
If you put a `1000.times do` around all socket client specs this is easy to reproduce. I won't have time these days to find out why it happens, but if somebody has the time, please investigate it.

With that and my other researches we should be reaching a point where no specs will fail on CI.

asterite:
Fixes #7401

I found out that the problem was in the ["tests read_timeout"](https://github.com/crystal-lang/crystal/blob/4e0f731a7314cf03a37f76453799e9de54f7a550/spec/std/http/client/client_spec.cr#L192) spec. This fixes it but I'm not _entirely_ sure why the exception happens.

What I think happens is:
- the client is set a read timeout, and it will timeout
- the server is said to wait for more time than that of the read timeout and then write a response
- by the time the server gets to write the response, the client's socket might be already closed because the GC already finalized it

At least I couldn't reproduce it with `GC_DONT_GC=1` (but then it failed because of too many open file requests), so I'm not sure. But I think my logic above is correct.

In this PR, when we expect a read timeout to happen we don't even bother instructing the server to write a response because that doesn't matter anymore.

However, this leads me to think... if I write an HTTP server and I accept a request and the client closes the socket before I get to write the response... an exception will be raised and it will be printed in the server's `spawn` handling that request. Is that OK? How do other languages handle this?

### DIFF
--- a/spec/std/http/client/client_spec.cr
+++ b/spec/std/http/client/client_spec.cr
@@ -3,15 +3,17 @@ require "openssl"
 require "http/client"
 require "http/server"
 
-private def test_server(host, port, read_time = 0, content_type = "text/plain")
+private def test_server(host, port, read_time = 0, content_type = "text/plain", write_response = true)
   server = TCPServer.new(host, port)
   begin
     spawn do
       io = server.accept
       sleep read_time
-      response = HTTP::Client::Response.new(200, headers: HTTP::Headers{"Content-Type" => content_type}, body: "OK")
-      response.to_io(io)
-      io.flush
+      if write_response
+        response = HTTP::Client::Response.new(200, headers: HTTP::Headers{"Content-Type" => content_type}, body: "OK")
+        response.to_io(io)
+        io.flush
+      end
     end
 
     yield server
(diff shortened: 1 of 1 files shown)


bew:
> if I write an HTTP server and I accept a request and the client closes the socket before I get to write the response... an exception will be raised and it will be printed in the server's spawn handling that request. Is that OK?

That's what I would expect at least..

asterite:
Ugh, that failure in darwin... come on, CI, I'm trying my best here, don't give me **new** errors...

asterite:
For history purposes, this is the darwin build that failed: https://circleci.com/gh/crystal-lang/crystal/18231?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I'll re-run it now.

straight-shoota:
I'm pretty sure the immediate issue should be fixed by #7197.

asterite:
Another one to investigate, a `Nil assertion failed` in `signal_spec.cr`: https://circleci.com/gh/crystal-lang/crystal/18298?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

straight-shoota:
That's already tracked in #7243

asterite:
We can probably merge this PR first, which is really small, and then focus on reviewing #7197, which is much larger and also is currently failing on my machine.

PR #7403, opened 2019-02-09
Specs: avoid GC finalization warning in Thread specs

### LINKED ISSUE #7400: "GC Warning: Finalization cycle" in std specs
It seems since some time ago the specs are printing "GC Warning: Finalization cycle" all over the place. The warning always points to a same object address. I debugged this a bit, printing each object passed to `GC.add_finalizer_impl` and I found out that `Thread` (!) is the culprit.

It seems `Thread` has a `finalize` on it, and some specs are doing `Thread.new` to test `Thread`, `Thread::Mutex` and `Thread::ConditionVariable`, all of which are currently not used at all because we don't support parallelism.

I don't know why a finalization cycle happens, but I think we should do these things:
- Remove `finalize` from `Thread`, because once we'll support parallelism we'll control its lifecycle
- Add a `destroy` method to `Thread` (or some other name) to replace `finalize`, and invoke it after using a `Thread` (in all `Thread`-related specs, for now)

What do you think?

asterite:
Fixes #7400

### DIFF
--- a/spec/std/thread_spec.cr
+++ b/spec/std/thread_spec.cr
@@ -21,6 +21,10 @@ describe Thread do
     thread.join
     current.should be(thread)
     current.should_not be(Thread.current)
+  ensure
+    # avoids a "GC Warning: Finalization cycle" caused by *current*
+    # referencing the thread itself, preventing the finalizer to run:
+    current = nil
   end
 
   it "yields the processor" do


ysbaddaden:
The problem may be the double reference between Thread and Fiber that prevents the finalization. Using `GC_register_finalizer_ignore_self` could solve the issue.

Now, calling an explicit destroy is fine, Thread isn't a public class.

asterite:
@ysbaddaden do you know where the reference from Fiber to thread is? I can't find it.

ysbaddaden:
Err, I mixed up the different MT branches, it's not in master :bowing_man:

asterite:
Because, yes, we can fix the cycle but right now I don't know which one is it.

ysbaddaden:
We already use `GC_register_finalizer_ignore_self` anyway.

If I make [this spec](https://github.com/crystal-lang/crystal/blob/master/spec/std/thread_spec.cr#L18) pending, then I don't have the warning anymore. The issue seems to be the `current` local variable of the captured `@func` block that keeps a reference to the thread!

asterite:
Wow, good catch! Then we shouldn't merge this PR, we should fix that particular spec, or altogether remove it or comment it for now.

ysbaddaden:
Thanks for pointing the issue, I'd never have thought of looking at closured data.

Just setting `current` to `nil` fixes the issue:
```diff
diff --git a/spec/std/thread_spec.cr b/spec/std/thread_spec.cr
index c84e500..9f75142 100644
--- a/spec/std/thread_spec.cr
+++ b/spec/std/thread_spec.cr
@@ -21,6 +21,10 @@ describe Thread do
     thread.join
     current.should be(thread)
     current.should_not be(Thread.current)
+  ensure
+    # avoids a "GC Warning: Finalization cycle" caused by *current*
+    # referencing the thread itself, preventing the finalizer to run:
+    current = nil
   end
   it "yields the processor" do
```

asterite:
@ysbaddaden Excellent! I updated this PR with that and reverted the removal of `finalize`. Let's see what CI says.

PR #7405, opened 2019-02-09
Parse IIS Set-Cookie date format

### LINKED ISSUE #7404: Error parsing cookies
```crystal
require "http"

require "http/server"

  context.response.headers["Set-Cookie"] = "my_cookie=; expires=Tue, 12-Oct-1999 07:00:00 GMT; path=/; secure; HttpOnly"
end

address = server.bind_tcp 5050
spawn server.listen

response = HTTP::Client.get "http://localhost:5050/"

response.cookies

server.close
```

Cookie will show no `secure` or `http_only` (both false)
Also, example by @Blacksmoke16 : https://play.crystal-lang.org/#/r/67na

This is a standard set-cookie header coming from IIS.

Sija:
Fixes #7404

### DIFF
--- a/src/time/format/custom/http_date.cr
+++ b/src/time/format/custom/http_date.cr
@@ -1,16 +1,18 @@
 require "./rfc_2822"
 
 struct Time::Format
-  # Parse a time string using the formats specified by [RFC 2616](https://tools.ietf.org/html/rfc2616#section-3.3.1).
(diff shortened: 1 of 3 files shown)


bararchy:
Thanks @Sija  :)

src/time/format/custom/http_date.cr
```diff
@@ -61,7 +61,7 @@ struct Time::Format
         char '-'
         short_month_name
         char '-'
-        year_modulo_100
+        full_or_short_year
```
straight-shoota: This derivates from the described format. The change is fine, but could you please mention the non-RFC-compliant IIS date format in the docs and add a code comment here to reference that?

Sija: Done.

PR #7406, opened 2019-02-10
 Suggest `next` when trying to break from captured block 

r00ster91:
When doing this:
```cr
def method(&block)
  block.call
end

method do
  break
end
```
then you get `Error in line 6: can't break from captured block`.
Now this makes the error message suggest `next` because when using `break` in a captured block you most likely tried to escape from the captured block and `next` makes that possible.
Also when you instead try to `return` from a captured block, it gives you a helpful error message: `Error in line 6: can't return from captured block, use next` so why not suggest `next` in case of `break` too?

This also downcases `next`'s and `break`'s error message just like all the other error messages.

### DIFF
--- a/spec/compiler/semantic/block_spec.cr
+++ b/spec/compiler/semantic/block_spec.cr
@@ -1075,7 +1075,7 @@ describe "Block inference" do
 
       bar
       ),
-      "can't break from captured block"
+      "can't break from captured block, use next"
   end
 
   it "errors if doing next in proc literal" do
--- a/spec/compiler/semantic/block_spec.cr
+++ b/spec/compiler/semantic/block_spec.cr
@@ -1075,7 +1075,7 @@ describe "Block inference" do
 
       bar
       ),
-      "can't break from captured block, use `next`, however `next` will have behave differently when used in a block or a loop inside of the captured block."
+      "can't break from captured block, try using `next` to exit the block."
   end
 
   it "errors if doing next in proc literal" do
--- a/spec/compiler/semantic/block_spec.cr
+++ b/spec/compiler/semantic/block_spec.cr
@@ -1085,7 +1085,7 @@ describe "Block inference" do
       }
       foo.call
       ),
-      "Invalid next"
+      "invalid next"
   end
 
   it "does next from captured block" do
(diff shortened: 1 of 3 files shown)


straight-shoota:
In general, this is probably a good feature. But I'm not sure if "use next" is the best advice. It's not always as simple as that, because `next` and `break`/`return` have different semantics. If the block is called multiple times, `next` might not give you what you want. So I think it should be a more relaxed suggestion.

asterite:
@straight-shoota Right on! I was thinking about `did you mean 'next'`? But that doesn't give much info either.

In general, I think we should improve all of our error messages to look a bit more like Elm, where they take a lot of time to explain what went wrong and suggestions with their explanations.

r00ster91:
What about sometimes linking to the docs?
```
can't break from captured block.
If you want to exit the captured block, use `next`, however `next` will have behave differently when used in a block or a loop inside of the captured block.
For more information see https://crystal-lang.org/reference/syntax_and_semantics/next.html.
```

Sija:
Ruby allows `break` within blocks which makes sense, since `next` provide different semantics and breaking from the block is a side-effect.

Sija:
Actually, why it's not possible to use `break` in captured blocks?

asterite:
@Sija `break` in non-captured blocks means "exit the enclosing method". For a capture block it can't work.

```crystal
def capture(&block)
  block
end

def foo
  # This break looks like it will exit foo, but that can't happen
  block = capture { break }
  puts 1
  block
end

block = foo
block.call
```

r00ster91:
I hope this is clear enough now. I don't think it needs to be as long as I proposed. In the end it might rather confuse the user.

straight-shoota:
Indeed, that is way to long and too much detail. A simple "Maybe you can use `next` instead?`" would suffice.

spec/compiler/semantic/block_spec.cr
```diff
@@ -1075,7 +1075,7 @@ describe "Block inference" do
 
       bar
       ),
-      "can't break from captured block, use `next`, however `next` will have behave differently when used in a block or a loop inside of the captured block."
+      "can't break from captured block, try using `next` to exit the block."
```
straight-shoota: I think stating that `next` exits the block is misleading exactly in the way we're trying to clarify things.

Sija:
@asterite Now it makes sense, thanks for explanation!

PR #7409, opened 2019-02-10
Fix `Signal::CHLD.reset` not clearing previously set handler, and fix related race condition in spec

### LINKED ISSUE #7243: Rarely failing nil assertion in signal_spec

asterite:
Fixes #7243

I thought about many ways to fix the race condition in this spec... I think the only safe alternative is to use a channel and send that over to the signal handler. If the handler runs before we send the channel, it will wait. If it runs later, the process will be already finished.

The only problem with this spec is that if the CHLD handler doesn't run for some reason (bug) then the spec will wait forever in `chan.send` because nobody will consume it. However, given that everything is working well now I don't this as a big problem, plus signal handling code doesn't change that frequently.

### DIFF
--- a/spec/std/signal_spec.cr
+++ b/spec/std/signal_spec.cr
@@ -50,5 +50,25 @@ describe "Signal" do
     called.should be_true
   end
 
+  it "CHLD.reset removes previously set trap" do
+    call_count = 0
+
+    Signal::CHLD.trap do
+      call_count += 1
+    end
+
+    Process.new("true", shell: true).wait
+    Fiber.yield
+
+    call_count.should eq(1)
+
+    Signal::CHLD.reset
+
+    Process.new("true", shell: true).wait
+    Fiber.yield
+
+    call_count.should eq(1)
+  end
+
   # TODO: test Signal::X.reset
 end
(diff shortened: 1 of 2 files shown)


asterite:
Okay... my first commit was missing `Signal::CHLD.reset` to clear that handler, because later if another spec did `Process.new(...)` and that process ended then that same handler would be called, hanging everything (on `chan.receive`), which effectively happened.

I added `Signal::CHLD.reset` and ran the specs locally... only to find out that that **didn't work**. The intention of keeping a default handler when `Signal::CHLD.reset` was good, but any previously set handler wasn't cleared. So this PR also fixes that. Phew!

PR #7413, opened 2019-02-11
Optimize dwarf line numbers decoding

asterite:
The first time an Exception's full backtrace (filenames, line numbers, etc.) must be built, the runtime will first decode all DWARF information from the dwarf file (or somewhere else? not sure).

I noticed this process takes about 6 seconds for the compiler's std specs (because it's a huge program, the bigger the program the bigger the dwarf file and the longer it will take to decode it), specifically you can see this doing:

```
$ bin/crystal build spec/std_spec.cr
$ ./std_spec -e "inspects with cause"
.

Finished in 5.88 seconds
```

I thought 6 seconds to decode the backtrace was too much so I started to dig in the code to see if I can improve it. I managed to improve **part** of the decoding process by about 6 times.

The improvement is explained in the PR code comments. There's a `decode_sequences` method in `DWARF::LineNumbers`. Previously it took 4.84 seconds to do it for the std_spec. Now it takes 0.82 seconds.

Now running that spec above looks like this:

```
$ bin/crystal build spec/std_spec.cr
14:05 $ ./std_spec -e "inspects with cause"
.

Finished in 1.34 seconds
```

Note that this improvement is only see the first time `ex.inspect_with_backtrace` is called for any exception. Later on the dwarf info should have been decoded. Still, it's good to win a lot of time in this first decoding.

I believe there might be more things to improve, but so far this is the easiest change that gives the most boost.

### DIFF
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -279,7 +279,7 @@ struct CallStack
           end
 
           strings = mach_o.read_section?("__debug_str") do |sh, io|
-            Debug::DWARF::Strings.new(io, sh.offset)
+            Debug::DWARF::Strings.new(io, sh.offset, sh.size)
           end
 
           mach_o.read_section?("__debug_info") do |sh, io|
@@ -381,7 +381,7 @@ struct CallStack
           end
 
           strings = elf.read_section?(".debug_str") do |sh, io|
-            Debug::DWARF::Strings.new(io, sh.offset)
+            Debug::DWARF::Strings.new(io, sh.offset, sh.size)
           end
 
           elf.read_section?(".debug_info") do |sh, io|
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -223,10 +223,7 @@ struct CallStack
       read_dwarf_sections unless @@dwarf_line_numbers
       if ln = @@dwarf_line_numbers
         if row = ln.find(pc)
-          path = ln.files[row.file]?
-          if dirname = ln.directories[row.directory]?
-            path = "#{dirname}/#{path}"
-          end
+          path = "#{row.directory}/#{row.file}"
           return {path, row.line, row.column}
         end
       end
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -282,11 +282,7 @@ struct CallStack
           mach_o.read_section?("__debug_info") do |sh, io|
             names = [] of {LibC::SizeT, LibC::SizeT, String}
 
-            while true
-              pos = io.pos
-              offset = pos - sh.offset
-              break unless offset < sh.size
-
+            while (offset = io.pos - sh.offset) < sh.size
               info = Debug::DWARF::Info.new(io, offset)
 
               mach_o.read_section?("__debug_abbrev") do |sh, io|
@@ -387,11 +383,7 @@ struct CallStack
           elf.read_section?(".debug_info") do |sh, io|
             names = [] of {LibC::SizeT, LibC::SizeT, String}
 
-            while true
-              pos = io.pos
-              offset = pos - sh.offset
-              break unless offset < sh.size
-
+            while (offset = io.pos - sh.offset) < sh.size
               info = Debug::DWARF::Info.new(io, offset)
 
               elf.read_section?(".debug_abbrev") do |sh, io|
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -285,8 +285,11 @@ struct CallStack
           mach_o.read_section?("__debug_info") do |sh, io|
             names = [] of {LibC::SizeT, LibC::SizeT, String}
 
-            while io.tell - sh.offset < sh.size
-              offset = io.tell - sh.offset
+            while true
+              pos = io.pos
+              offset = pos - sh.offset
+              break unless offset < sh.size
+
               info = Debug::DWARF::Info.new(io, offset)
 
               mach_o.read_section?("__debug_abbrev") do |sh, io|
@@ -387,8 +390,11 @@ struct CallStack
           elf.read_section?(".debug_info") do |sh, io|
             names = [] of {LibC::SizeT, LibC::SizeT, String}
 
-            while io.tell - sh.offset < sh.size
-              offset = io.tell - sh.offset
+            while true
+              pos = io.pos
+              offset = pos - sh.offset
+              break unless offset < sh.size
+
               info = Debug::DWARF::Info.new(io, offset)
 
               elf.read_section?(".debug_abbrev") do |sh, io|
--- a/src/debug/dwarf/line_numbers.cr
+++ b/src/debug/dwarf/line_numbers.cr
@@ -115,8 +115,8 @@ module Debug
       record Row,
         address : UInt64,
         op_index : UInt32,
-        directory : Int32,
-        file : Int32,
+        directory : String,
+        file : String,
         line : Int32,
         column : Int32,
         end_sequence : Bool
@@ -168,19 +168,11 @@ module Debug
       # reduce the memory usage of repeating a String many times.
       getter matrix : Array(Array(Row))
 
-      # The array of indexed directory paths.
-      getter directories : Array(String)
-
-      # The array of indexed file names.
-      getter files : Array(String)
-
       @offset : LibC::OffT
 
       def initialize(@io : IO::FileDescriptor, size)
         @offset = @io.tell
         @matrix = Array(Array(Row)).new
-        @directories = [] of String
-        @files = [] of String
         decode_sequences(size)
       end
 
@@ -213,12 +205,6 @@ module Debug
 
       # Decodes the compressed matrix of addresses to line numbers.
       private def decode_sequences(size)
-        # Map names to indexes to avoid doing linear scans over @files and @directories
-        indexes = {
-          directory: {} of String => Int32,
-          filename:  {} of String => Int32,
-        }
-
         while true
           pos = @io.tell
           offset = pos - @offset
@@ -245,7 +231,7 @@ module Debug
           read_filename_table(sequence)
 
           if @io.tell - @offset < sequence.offset + sequence.total_length
-            read_statement_program(sequence, indexes)
+            read_statement_program(sequence)
           end
         end
       end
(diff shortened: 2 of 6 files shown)


yxhuvud:
Replacing the implementation of #find at line 187 with a binary search looks like another low hanging fruit, assuming it is a nontrivial amount of rows.

ysbaddaden:
Impressive boost, especially since the files/directories arrays are usually small if I remember correctly.

asterite:
@yxhuvud Yes, `find` could use a binary search, I thought of it. That said, I don't understand why `@matrix` is defined like that, maybe pushing all `Row`s in a single array would be enough? Not sure.

But that's not the bottleneck, at least when I profiled it. The bottleneck are:
- `Debug::DWARF::Strings#decode`: most of the time spent it `gets` and `tell`/`pos` (this can be optimized just a bit by using `gets('\0', chomp: true)` but I didn't want to make this PR too big
- `Debug::DWARF.read_unsigned_leb128`: most of the time is spent in `read_byte` 

I'm not sure there's something we can do about those two points. In both cases (I think?) the IO is repositioned and read, and because it's buffered an entire block is consumed. Maybe changing the buffer size could improve this because `decode` usually needs just a few bytes (it's just a name) and I guess the same applies to `read_unsigned_leb128`. But for that we should first make the buffer size of `IO::Buffered` configurable.

asterite:
@ysbaddaden Yes, the arrays are small. The problem is that the `register_filename` and `register_directory` end up being called, at least in `std_spec`, 1 million times, and most of those calls are always with the same filename/directory that was passed the last time. In fact, another way I tried to optimize this is by only recomputing the file/path indexes when `register.file` was set, or when `Register.new` was done, or at the beginning of `read_statement_program` but the code ended up being a bit more complex and taking more time if I also didn't do the "is it the same as the last time" check.

src/debug/dwarf/line_numbers.cr
```diff
@@ -389,20 +389,54 @@ module Debug
...
-          return index
+        if name.same?(@last_filename)
+          return @last_filename_index
         end
-        @files << name
-        @files.size - 1
+
+        @last_filename = name
+
+        index = @files.index(name)
```
yxhuvud: How often does it go back and look up a file that it has already seen? If it happens often, perhaps changing `@files`to be a `Hash(String,Int32)` would be a good idea? Your cache would probably still be worth it, but the rest of the method could be

```
if index = @files[name]?
  index
else
  @files[name] = @files.size 
end
```

.. but if it doesn't happen often then it would not be worth it.

src/debug/dwarf/line_numbers.cr
```diff
@@ -389,20 +389,54 @@ module Debug
...
+
+        unless index
+          @files << name
+          index = @files.size - 1
```
yxhuvud: Swap the order here and there would be no need to decrement.

ysbaddaden:
The matrix is defined as an Array(Array(Row)) to match the dwarf definition, but more importantly to skip a bunch of addresses to reach the actual block more quickly.

src/debug/dwarf/line_numbers.cr
asterite: Thanks, fixed!

src/debug/dwarf/line_numbers.cr
asterite: Yes, that's a good idea. I already tried it before sending this PR (we actually need a separate Hash because `@files` is used from the outside as an Array). It reduced the time from 0.82s to 0.79s, so maybe all the trouble of having more variables and more logic might not be worth it, but I can include it in this PR. What do you think?

ysbaddaden: It's nice. Using a Hash *and* an Array will require more memory, but that should be contained since there aren't that much entries.

asterite: Yes. Also, I had code that used the Hash just for computing this faster but then it's dropped because only the Array matters to clients (from what I've seen).

I'd really like to write test cases for all of these. I might do that these weeks.

asterite: I added the hash.

asterite:
I added another commit that introduces two temporary hashes that map dir/file-names to indexes, to avoid doing linear scans over `@files` and `@directories`. That reduces the time of `Debug::DWARF::LineNumbers#decode_sequences` from 0.82s to 0.78s.

src/debug/dwarf/line_numbers.cr
```diff
@@ -389,20 +395,56 @@ module Debug
...
-      private def register_filename(name)
-        if index = @files.index(name)
-          return index
+      # When decoding statement programs when asking for the current
+      # filename it's usually the case that it's the same as the old
+      # one (because all info from a single file comes first, then
+      # another file comes next, etc.). So we remember the last
+      # mapped index to avoid unnecessary lookups.
+      @last_filename : String?
+      @last_filename_index = 0
```
yxhuvud: BTW, if the files are read in some order (and not jumping from file to file, do `@last_filename` ever differ from `@files[-1]`, (and similar for the index)?

asterite: Yes, they differ. They are not read in order. I don't know why, though, I'm not familiar with the format. Maybe understanding the format first, then optimizing it might be better... 😊

asterite:
I just pushed another commit with more improvements. I tried to do small commits for this but I was experimenting and in the end it was a bit hard for me to separate all the changes.

Essentially, I changed some `UInt32 | UInt64` unions to just `UInt64` (the cost of dealing with unions is bigger than just tracking the exact type). Another change is avoiding calling `io.tell` multiple times (this is expensive; I also used `io.pos`, I think we should remove the `tell` alias).

But the biggest change is in `DWARF::Strings`: we now use a small (16Kb) buffer to read strings from instead of seeking and reading from the IO (which is done **a lot** of times).

So now:

```crystal
$ bin/crystal build spec/std_spec.cr
$ ./std_spec -e "inspects with cause"
.

Finished in 1.31 seconds
```

(down from 1.83s)

I'm almost sure more things can be optimized, but with these changes I'm pretty satisfied now.

asterite:
Meh, I'll redo the last commits in smaller commits because CI failed and I can't figure out why...

wontruefree:
@asterite why use 2 hashes?  Directories and filenames would not collide.  It seems like you could reduce this to one register method and one hash.

asterite:
@wontruefree `files` and `directories` are public and are used by other parts of the code that need the "index -> name" mapping.

wontruefree:
@asterite I am not sure I understand.

But if you have considered it and know why it won't work that is cool.  I cannot make a PR into your fork but here is what I was thinking.

https://github.com/crystal-lang/crystal/compare/master...wontruefree:opt/dwarf-decoding

asterite:
@wontruefree the `@files` and `@directories` variables are exposed through the getters of `LineNumbers` and used [here](https://github.com/crystal-lang/crystal/blob/323ed75fb9edfd99134627b635e1325e605ad8df/src/callstack.cr#L226-L227). In your code you just stop putting data there. Try it with this code:

```crystal
begin
  raise "oh no"
rescue ex
  puts ex.inspect_with_backtrace
end

begin
  [1][2]
rescue ex
  puts ex.inspect_with_backtrace
end
```

 Do you get something correct in the output?

src/debug/dwarf/line_numbers.cr
ysbaddaden: If I recall correctly, the format follows functions (symbols). An instruction group of the function refers to a file:line, then the following group likely refers to the same file at a further line... unless the compiler inlined code, which will cause jumps to another file, etc.

If I recall correctly, that is.

Sija:
@asterite CI just failed again with `GC Warning: Failed to expand heap by 16777216 bytes`, seems like crusade is not over yet...

wontruefree:
@asterite I see what you are saying.
I updated the branch and reduced the external interface to use a method called paths.

After thinking about this why not just make the `index` the instance variable and have the callstack use a hash lookup.  Seems like that would be faster.

asterite:
I think some tests are allocating too much memory and it dies on Linux 32 bits. But I didn't start investigating that yet.

asterite:
I just pushed yet another commit. Thanks to what @wontruefree said I noticed that it doesn't make sense to build those `files`/`directories` arrays. Instead, we can store them directly inside each `Row`.

@ysbaddaden Do you see a problem with this?

src/callstack.cr
```diff
@@ -381,14 +381,17 @@ struct CallStack
...
-            Debug::DWARF::Strings.new(io, sh.offset)
+            Debug::DWARF::Strings.new(io, sh.offset, sh.size)
           end
 
           elf.read_section?(".debug_info") do |sh, io|
             names = [] of {LibC::SizeT, LibC::SizeT, String}
 
-            while io.tell - sh.offset < sh.size
-              offset = io.tell - sh.offset
+            while true
```
ysbaddaden: Maybe:
```crystal
while (offset = io.pos - sh.offset) < sh.size
```

asterite: I like it! Done :-)

ysbaddaden:
@asterite It's probably fine. I don't recall exactly why I used arrays and indices. Maybe it helped limit the overall memory usage by avoiding to keep duplicated string allocations?

asterite:
@ysbaddaden I don't think so because all of the strings are stored inside arrays inside `Sequence`s so a same string shares the same memory for others. But maybe it wasn't like that at some point, I don't know. This way it seems to work :-)

Maybe if I have time I'll learn that format and write specs for it, so we can continue simplifying it (with more confidence). However, right now I think the times are more or less fine so I might focus on something else after this.

ysbaddaden:
Once the sequences are decoded we don't retain any references to them, so their `@directories` and `@files` arrays will be collected by the GC. I think most folder names and enough file names are shared across sequences that de-duplicating can be significant on memory usage, especially with large programs, but that should be verified.

Sija:
@asterite Would you mind updating PR description with the newest run-time numbers?

asterite:
@Sija done!

asterite:
Good to go?

asterite:
I'll merge it. Should I squash or not? I never know. I think each commit is meaningful by itself.

ysbaddaden:
Let's keep each commit. We could squash "simplify while loop" but that's nitpicking.

PR #7414, opened 2019-02-11
channel: call ops.unwait as soon as fiber is resumed/enqueued; fixes select segfault

### LINKED ISSUE #3900: Crash in fiber switching with select

forksaber:
Fixes #3900

A `select` call registers multiple wait events (one for each of its branches) and yields. When any of the branches is ready, `select` fiber resumes and unwaits the remaining events by calling 
`
ops.each &.unwait
`. This works correctly with unbuffered channels as upon readiness, they immediately switch to the waiting/selecting fiber.

However, since buffered channels don't block unless buffer is full(send blocks) or empty(receive blocks), the `select`ing fiber is not resumed upon readiness. Hence, `unwait` is not called on time. By the time `unwait` is called, the `select`ing fiber may be enqueued multiple times.

Resuming this fiber after it has finished executing causes a segfault.

So, the fix here to call `ops.each &.unwait` as soon as the first resume/enqueue happens.

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -5,7 +5,7 @@ abstract class Channel(T)
     abstract def ready?
     abstract def execute
     abstract def wait
-    abstract def unwait
+    abstract def unwait(fiber : Fiber)
   end
 
   class ClosedError < Exception
@@ -30,10 +30,8 @@ abstract class Channel(T)
 
   def close
     @closed = true
-    Crystal::Scheduler.enqueue @senders
-    @senders.clear
-    Crystal::Scheduler.enqueue @receivers
-    @receivers.clear
+    enqueue! @senders
+    enqueue! @receivers
     nil
   end
 
@@ -57,20 +55,28 @@ abstract class Channel(T)
     pp.text inspect
   end
 
+  private def enqueue!(fibers : Deque(Fiber))
+    # not using .each here as fiber.enqueue method can mutate the deque through callback
+    # which will cause segfault with .each
+    while fiber = fibers.shift?
+      fiber.enqueue
+    end
+  end
+
   protected def wait_for_receive
     @receivers << Fiber.current
   end
 
-  protected def unwait_for_receive
-    @receivers.delete Fiber.current
+  protected def unwait_for_receive(fiber : Fiber)
+    @receivers.delete fiber
   end
 
   protected def wait_for_send
     @senders << Fiber.current
   end
 
-  protected def unwait_for_send
-    @senders.delete Fiber.current
+  protected def unwait_for_send(fiber : Fiber)
+    @senders.delete fiber
   end
 
   protected def raise_if_closed
@@ -111,9 +117,11 @@ abstract class Channel(T)
         return ops.size, nil
       end
 
+      fiber = Fiber.current
+      fiber.callback = ->{ ops.each &.unwait(fiber) }
+
       ops.each &.wait
       Crystal::Scheduler.reschedule
-      ops.each &.unwait
     end
   end
 
@@ -146,8 +154,8 @@ abstract class Channel(T)
       @channel.wait_for_receive
     end
 
-    def unwait
-      @channel.unwait_for_receive
+    def unwait(fiber : Fiber)
+      @channel.unwait_for_receive(fiber)
     end
   end
 
@@ -170,8 +178,8 @@ abstract class Channel(T)
       @channel.wait_for_send
     end
 
-    def unwait
-      @channel.unwait_for_send
+    def unwait(fiber : Fiber)
+      @channel.unwait_for_send fiber
     end
   end
 end
@@ -192,9 +200,7 @@ class Channel::Buffered(T) < Channel(T)
     raise_if_closed
 
     @queue << value
-    Crystal::Scheduler.enqueue @receivers
-    @receivers.clear
-
+    enqueue! @receivers
     self
   end
 
@@ -206,8 +212,7 @@ class Channel::Buffered(T) < Channel(T)
     end
 
     @queue.shift.tap do
-      Crystal::Scheduler.enqueue @senders
-      @senders.clear
+      enqueue! @senders
     end
   end
 
@@ -264,7 +269,7 @@ class Channel::Unbuffered(T) < Channel(T)
 
     @value.tap do
       @has_value = false
-      Crystal::Scheduler.enqueue @sender.not_nil!
+      @sender.not_nil!.enqueue
       @sender = nil
     end
   end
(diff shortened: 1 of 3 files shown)


src/channel.cr
```diff
@@ -14,10 +14,34 @@ abstract class Channel(T)
     end
   end
 
+  # A Fiber wrapper which executes a callback
+  # before being resumed or enqueued
+  # :nodoc:
```
Sija: This should be moved into the first line to make it work (with blank line in between plz).

src/channel.cr
```diff
@@ -111,9 +135,12 @@ abstract class Channel(T)
         return ops.size, nil
       end
 
-      ops.each &.wait
+      fiber = SelectFiber.current
+      unwait_callback = ->{ ops.each &.unwait(fiber); nil }
```
Sija: Why not inline it?

straight-shoota:
Hey @forksaber, thanks for your initiative!

This PR most likely conflicts with #7407 which also applies substantial changes to `Channel.select`. That PR has a far broader scope on multithreading concurrency, and it's quite heavy but definitely worth a look!

/cc @ysbaddaden

src/channel.cr
forksaber: resolved

src/channel.cr
forksaber: fixed

ysbaddaden:
I think the approach is correct: the ready fiber must unwait/cancel other clauses *before* yielding to another fiber. The wrapper class should be a `struct`, thought, to avoid an allocation on each and every channel operation, which would noticeably slow down channels, even outside of select.

ysbaddaden:
For example:
```crystal
struct SelectFiber
  def initialize(@fiber = Fiber.current)
    @callback = nil
  end

  def initialize(@fiber = Fiber.current, &@callback : ->)
  end

  def resume
    @callback.try(&.call)
    @fiber.resume
  end

  def enqueue
    @callback.try(&.call)
    Crystal::Scheduler.enqueue(@fiber)
  end
end

f1 = SelectFiber.new
f2 = SelectFiber.new { p 123 }
```

Or maybe add it to `Fiber` itself, as a self erasing callback:

```crystal
class Fiber
  # :nodoc:
  setter callback : Proc(Nil, Nil)?

  def resume
    exec_callback
    Crystal::Scheduler.resume(self)
  end

  def enqueue
    exec_callback
    Crystal::Scheduler.enqueue(self)
  end

  private def exec_callback
    if callback = @callback
      @callback = nil
      callback.call
    end
  end
end

f = Fiber.current
f.callback = -> { ops.each(&.unwait(f)) }
```

forksaber:
Thanks @ysbaddaden for your input. I prefer adding the self erasing callback to `Fiber` itself. I also had the same approach in mind....but was reluctant to make a change to `Fiber` . Now with your go-ahead, I'll update this PR with changes to `Fiber`.

ysbaddaden:
Yes, the patch will be smaller, and it's pushing more of the scheduling API to Fiber itself, which I think if good (less calls to `Crystal::Scheduler`).

forksaber:
updated PR to store callback in `Fiber` object

src/channel.cr
```diff
@@ -111,9 +117,11 @@ abstract class Channel(T)
         return ops.size, nil
       end
 
+      fiber = Fiber.current
+      fiber.callback = ->{ ops.each &.unwait(fiber); nil }
```
Sija: You can drop `; nil`, since `#each` has `Nil` return type already.

forksaber: fixed

Sija: It's still there :)

forksaber: oops...hopefully fixed this time :)

Sija: It is indeed :)

Sija:
Is this GTG? Would be nice to have it 🚢 :)

forksaber:
It's GTG from my side.

src/fiber.cr
```diff
@@ -60,6 +60,9 @@ class Fiber
   # :nodoc:
   property previous : Fiber?
 
+  # :nodoc:
+  setter callback : Proc(Nil)?
```
straight-shoota: Could you please add a comment to explain the purpose of this?
Comments are welcome in other places as well ;)

src/channel.cr
```diff
@@ -111,9 +117,11 @@ abstract class Channel(T)
         return ops.size, nil
       end
 
+      fiber = Fiber.current
+      fiber.callback = ->{ ops.each &.unwait(fiber) }
```
asterite: This forms a closure and so it allocates memory, which is not ideal. Please change it to store a reference to `ops` directly in `Fiber`. For example you can have `record UnwaitCallback, ops : ...` and then `callback : UnwaitCallback` or something like that (and if we need more callbacks we create a union and son on).

forksaber: @asterite I tried implementing your suggestion but I am not able to figure out the `type` for `ops`. If I use `Tuple | Array`, I get the following error:
```
can't use Array(T) in unions yet, use a more specific type
```

asterite: Oh, right, it can't be implemented. Nevermind.

straight-shoota: Would converting the tuple to an array be more efficient than a closure?

asterite: The problem is that w don't know the types inside the array

straight-shoota: Yeah, right...
Yet we know it's all `Channel`, we just don't know the generic arguments. But that's actually irrelevant for calling `#unwait`. Just thinking here, but we could hack around it by introducing a module which is included in `Channel`:
```cr
abstract class Channel(T)
  module Unwaitable
    def unwait(fiber : Fiber)
    end
  end
    
  include Unwaitable
end

record UnwaitCallback, ops : Array(Channel::Unwaitable)
```

It works. Question is, is it worth the effort of having an additional type and needing to explicitly cast tuple and array arguments to `Array(Channel::Unwaitable)`? Probably not.

asterite:
Closing because #3900 was closed because of #8112

PR #7417, opened 2019-02-12
Extract Fiber::StackPool

ysbaddaden:
Keeps stack operations (allocate, recycle, free) out of Fiber itself.

Just a mere extraction; not MT-safe but will be easy to add —e.g. use a mutex to protect the `deque.pop?`, `deque.push` and `deque.shift?` calls.

### DIFF
--- a/src/fiber.cr
+++ b/src/fiber.cr
@@ -1,5 +1,6 @@
 require "c/sys/mman"
 require "thread/linked_list"
+require "./fiber/stack_pool"
 
 # Load the arch-specific methods to create a context and to swap from one
 # context to another one. There are two methods: `Fiber#makecontext` and
@@ -42,10 +43,10 @@ fun _fiber_get_stack_top : Void*
 end
 
 class Fiber
-  STACK_SIZE = 8 * 1024 * 1024
-
   @@fibers = Thread::LinkedList(Fiber).new
-  @@stack_pool = [] of Void*
+
+  # :nodoc:
+  class_getter stack_pool = StackPool.new
 
   @stack : Void*
   @resume_event : Crystal::Event?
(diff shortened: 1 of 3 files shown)


ysbaddaden:
The circle/linux job failed, but the error seems unrelated:
```
1) TCPSocket fails when connection is refused
   Failure/Error: expect_raises_errno(Errno::ECONNREFUSED, "Error connecting to 'localhost:#{port}'") do
     Expected Errno but nothing was raised
   # spec/std/socket/tcp_socket_spec.cr:132
```

asterite:
We should note those CI errors and try to fix them. Lately I've been trying to get a CI that we can trust but it seems we are still missing a few cases.

straight-shoota:
@asterite The hard thing is recognizing repeated instances of the same CI failure. A single random connection error simply isn't really worth noting because it could happen for many reasons and most of them are out of our control.

sdogruyol:
Are we holding this just because of the CI failure?

asterite:
I just re-ran it.

src/fiber/stack_pool.cr
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    # Removes a stack from the bottom of the pool, or allocates a new one.
+    def checkout
+      stack = @deque.pop? || allocate
+      {stack, stack + STACK_SIZE}
+    end
+
+    # Appends a stack to the bottom of the pool.
+    def release(stack)
```
Sija: Isn't it bit of a weird name when you're actually _push_ing the stack into the pool?

ysbaddaden: I used the namings from `DB::Pool`.

sdogruyol:
@asterite should be good to :ship:

src/fiber.cr
```diff
@@ -65,9 +63,13 @@ class Fiber
     @@fibers.delete(fiber)
   end
 
+  # :nodoc:
+  def self.stack_pool
+    @@stack_pool ||= StackPool.new
```
kostya: why this need to be lazy initialized? it add little overhead

asterite: Does this need to be lazily created? It will never be `nil` in any program.

ysbaddaden: Thanks, I overlooked that!

PR #7418, opened 2019-02-12
Atomic::Flag

ysbaddaden:
An atomic flag, that can be set or not.

Obviously a requirement for MT; not really useful on a single thread, except for the slightly nicer API (I think); definitely an improvement over doing this manually with an `Atomic(Int32)` (see second commit).

```crystal
flag = Atomic::Flag.new

if flag.test_and_set
  do_something
end
```

```crystal
flag = false

unless flag
  flag = true
  do_something
end
```

### DIFF
--- a/src/atomic.cr
+++ b/src/atomic.cr
@@ -222,3 +222,33 @@ struct Atomic(T)
     end
   end
 end
+
+# An atomic flag, that can be set or not.
+#
+# Concurrency safe. If many fibers try to set the atomic in parallel, only one
+# will succeed.
+#
+# Example:
+# ```
+# flag = Atomic::Flag.new
+# flag.test_and_set # => true
+# flag.test_and_set # => false
+# flag.clear
+# flag.test_and_set # => true
+# ```
+struct Atomic::Flag
+  def initialize
+    @value = 0_u8
+  end
+
+  # Atomically tries to set the flag. Only succeeds and returns `true` if the
+  # flag wasn't previously set; returns `false` otherwise.
+  def test_and_set : Bool
+    Atomic::Ops.atomicrmw(:xchg, pointerof(@value), 1_u8, :sequentially_consistent, false) == 0_u8
+  end
+
+  # Atomically clears the flag.
+  def clear : Nil
+    Atomic::Ops.store(pointerof(@value), 0_u8, :sequentially_consistent, true)
+  end
+end
--- a/spec/std/atomic_spec.cr
+++ b/spec/std/atomic_spec.cr
@@ -186,3 +186,18 @@ describe Atomic do
     atomic.get.should eq(nil)
   end
 end
+
+describe Atomic::Flag do
+  it "#test_and_set" do
+    flag = Atomic::Flag.new
+    flag.test_and_set.should be_true
+    flag.test_and_set.should be_false
+  end
+
+  it "#clear" do
+    flag = Atomic::Flag.new
+    flag.test_and_set.should be_true
+    flag.clear
+    flag.test_and_set.should be_true
+  end
+end
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -122,15 +122,12 @@ enum Signal : Int32
     LibC.sigismember(pointerof(@@sigset), self) == 1
   end
 
-  @@setup_default_handlers = Atomic(Int32).new(0)
+  @@setup_default_handlers = Atomic::Flag.new
 
   # :nodoc:
   def self.setup_default_handlers
-    _, success = @@setup_default_handlers.compare_and_set(0, 1)
-    return unless success
-
+    return unless @@setup_default_handlers.test_and_set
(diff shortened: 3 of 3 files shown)


src/atomic.cr
```diff
@@ -222,3 +222,34 @@ struct Atomic(T)
...
+#
+# Example:
+# ```
+# flag = Atomic::Flag.new
+# flag.test_and_set # => true
+# flag.test_and_set # => false
+# flag.clear
+# flag.test_and_set # => true
+# ```
+struct Atomic::Flag
```
Sija: Shouldn't this be a Class instead, in case such a flag would be passed around?

ysbaddaden: `Atomic(T)` is a struct.

asterite:
Would it be hard to make the existing `Atomic` allow `Bool`? There are a number of compile-time checks to support different types (integers, references), I think we can add another one for `Bool` and then it's just `Atomic.new(false)`, etc.

src/atomic.cr
```diff
@@ -222,3 +222,34 @@ struct Atomic(T)
...
+  def initialize
+    @value = 0
```
asterite: Maybe `0_u8`? That way it occupies less memory.

ysbaddaden:
That would be identical to `Atomic(Int32).new(0)`, which is noisy:
```crystal
flag = Atomic(Bool).new(false)

_, success = flag.compare_and_set(false, true)
if success
  do_something
end
```

I like the simplified interface of Atomic::Flag: `#set : Bool`, `#clear`. No more details. We could probably set some memory order to speed things up a bit, too.

In C `atomic_flag` implementations are guaranteed to be lock-free, when `_Atomic` could use locks, though I'm not sure how to guarantee it with LLVM.

asterite:
@ysbaddaden Makes sense. What about a method to also `test_and_set` to `false`? The API seems designed with some code you have in mind but it doesn't look "generic" enough to me.

src/atomic.cr
```diff
@@ -222,3 +222,34 @@ struct Atomic(T)
...
+  end
+
+  # Atomically tries to set the flag. Only succeeds and returns `true` if the
+  # flag wasn't previously set; returns `false` otherwise.
+  def test_and_set : Bool
+    Atomic::Ops.atomicrmw(:xchg, pointerof(@value), 1, :sequentially_consistent, false) == 0
+  end
+
+  # Atomically clears the flag. Should only be called from the fiber that
+  # successfully set the flag.
```
asterite: > Should only be called from the fiber that successfully set the flag
Why?

straight-shoota:
@ysbaddaden Just thinking... What about adding `#test_and_set` directly to `Atomic`, so it can be used with `Atomic(Bool)`? This method could just raise a compile time error it `T != Bool`.

But it would essentially just trade one type (`Atomic::Flag`) for another (`Atomic(Bool)`) with a not-as-concise API...

ysbaddaden:
@asterite this comes from C11 / C++11 :
- https://en.cppreference.com/w/c/atomic/atomic_flag
- https://en.cppreference.com/w/cpp/atomic/atomic_flag

It looked generic enough to me, but maybe we don't need this now and I can just keep it in the MT branch. My use cases are mostly run an initializer once and MT-safe spin locks, cases where we don't care about values (true/false, 0/1, whatever). I just need a flag that can be set once then cleared.

> it would essentially just trade one type (`Atomic::Flag`) for another (`Atomic(Bool)`) with a not-as-concise API

Exactly. I prefer an explicit type over adding a method to Atomic that can only be used with *some* type, and always having to specify this type *and* the default value.

src/atomic.cr
ysbaddaden: Hum, maybe this is too restricted to using a flag as a lock, I'll remove it.

wontruefree:
Is there a way to test the atomic nature of this feature?
something like

```crystal
flag = Atomic::Flag.new
spawn {
  sleep 0.1
  flag.test_and_set.should be flase
}
spawn {
  flag.test_and_set.should be true
}
Fiber.yield
```

straight-shoota:
@wontruefree There is no way two fibers could be accessing the flag simultaneously because they can only switch execution at specific points (sleep, IO, yield etc.). This is only relevant for multithreading. And race conditions are always hard to demonstrated.

wontruefree:
@straight-shoota Yeah I dont know how to test this.  I just thought it would be nice to have a test for the functionality and the interface.

src/atomic.cr
bcardiff: This is still pending

src/atomic.cr
bcardiff: Is there any inherent issue with calling the method from other fiber/thread?

ysbaddaden:
I applied the changes.

Sija:
Seems like it's all good here! :shipit:

PR #7419, opened 2019-02-12
Improve specs and docs for #each_line

straight-shoota:
This extracts the uncontested improvements of #7275 which can be merged regardless of the future of `#each_line_with_number`.

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -908,17 +908,13 @@ abstract class IO
   # ```
   # io = IO::Memory.new("hello\nworld")
   # io.each_line do |line|
-  #   puts line.chomp.reverse
+  #   puts line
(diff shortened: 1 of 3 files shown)


src/io.cr
```diff
@@ -908,17 +908,26 @@ abstract class IO
...
-  #   puts line.chomp.reverse
+  #   puts line
   # end
+  # # output:
+  # # hello
+  # # world
   # ```
   #
-  # Output:
+  # The second argument yielded to the block is the line number starting at `1`.
```
Sija: Leftovers from previous PR here, ditto below.

straight-shoota: Oops

Issue #6934, opened 2018-10-12
Segfault when running a static binary built on Alpine musl 


j8r:
I use a library I made,  [clicr](https://github.com/j8r/clicr) library, which consists of recursive macros.
When compiling this program statically with  `crystal build --static cli.cr` in an `alpine:latest`, running `./cli ver` give this result:
```
Invalid memory access (signal 11) at address 0x0
[1]    21902 segmentation fault (core dumped)  ./cli
```
Code:
```crystal
require "clicr"
def run
  Clicr.create(
    commands: {
      ver: {
        info:   "Version with general system information",
        action: "version",
      },
    }
  )
end

def version
  puts 0.1
end

run
```

In `gdb`, when ` run` and then `backtrace`, I got:
```sh
Program received signal SIGBUS, Bus error.
0x00007ffff7f4d428 in GC_find_limit_with_bound ()
(gdb) backtrace
#0  0x00007ffff7f4d428 in GC_find_limit_with_bound ()
#1  0x00007ffff7f4d4eb in GC_init_linux_data_start ()
#2  0x00007ffff7f4c521 in GC_init ()
#3  0x00007ffff7f09770 in init () at /usr/lib/crystal/core/gc/boehm.cr:88
#4  0x00007ffff7f456eb in main () at /usr/lib/crystal/core/crystal/main.cr:35
#5  0x00007ffff7ef1996 in main () at /usr/lib/crystal/core/crystal/main.cr:106
```

Building the code with dynamic linking works -  `crystal build cli.cr && ./cli ver` returns `0.1`

Crystal version:
```
/ # crystal -v
Crystal 0.26.1 (2018-08-30)

LLVM: 5.0.1
Default target: x86_64-alpine-linux-musl
```

RX14:
looks like a bdwgc bug

RX14:
actually, that gdb `SIGBUS` is expected, you should keep using `continue` when you get a `SIGBUS` inside `GC_find_limit*` until you get to the real segfault

RX14:
useful...

ysbaddaden:
Sounds a lot like https://github.com/crystal-lang/crystal/issues/4276

j8r:
Compiling with `crystal build --link-flags "-static -lunwind"` seems to do the trick...

j8r:
Reduced example:
```crystal
raise ""
```
```
Invalid memory access (signal 11) at address 0x0
[...]

PR #7420, opened 2019-02-12
Add test_alpine to circleci config

### LINKED ISSUE #6943: Run CI on Alpine/musl
Since musl is the only way to reliably compile with statically linked libraries, I'd suggest to improve its support level. Mainly, automatic tests would be great.
This shouldn't be very difficult to integrate into CI. Both travis and circle support docker images and the specs can just be executed in an alpine image for `i686-linux-musl` and `x86_64-linux-musl`.

/cc #6903 #6917

straight-shoota:
Fixes #6943

This is just experimental for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,6 +62,22 @@ jobs:
       - store_artifacts:
           path: ~/test-results/spec
 
+  test_alpine:
+    docker:
+    - image: jrei/crystal-alpine:latest
+    environment:
+      <<: *env
+      TRAVIS_OS_NAME: linux
+      ARCH: x86_64
+      ARCH_CMD: linux64
+    steps:
+    - checkout
+    - run: echo 'export CURRENT_TAG="$CIRCLE_TAG"' >> $BASH_ENV
+    - run: bin/ci prepare_build
+    - run: make std_spec clean threads=1
+    - run: make crystal std_spec compiler_spec docs threads=1
+    - run: find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen
+
   test_darwin:
     macos:
       xcode: "11.1.0"
@@ -504,6 +520,8 @@ workflows:
                 - /.*\bci\b.*/
       - test_linux32_std:
           filters: *unless_maintenance
+      - test_alpine:
+          filters: *unless_maintenance
       - test_darwin:
           filters: *unless_maintenance
       - test_preview_mt:
@@ -527,6 +545,8 @@ workflows:
             tags:
               only: /.*/
       - test_linux32_std:
+          filters: *per_tag
+      - test_alpine:
           filters: *per_tag
       - test_darwin:
           filters: *per_tag
@@ -602,6 +622,7 @@ workflows:
     jobs:
       - test_linux
       - test_linux32_std
+      - test_alpine
       - test_darwin
       - test_preview_mt
       - check_format
@@ -659,6 +680,8 @@ workflows:
                 - /release\/.+/
                 - /.*\bci\b.*/
       - test_linux32_std:
+          filters: *maintenance
+      - test_alpine:
           filters: *maintenance
       - test_darwin:
           filters: *maintenance
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,6 @@ jobs:
       TRAVIS_OS_NAME: linux
       ARCH: x86_64-musl
       ARCH_CMD: linux64
-      DOCKER_TEST_PREFIX: crystallang/crystal:ci-test-dev # TODO: Remove after 0.33.0
     steps: *ci_steps
 
   test_darwin:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,15 @@ defaults:
     - run: echo 'export CURRENT_TAG="$CIRCLE_TAG"' >> $BASH_ENV
     - run: bin/ci prepare_build
     - run: bin/ci build
+    - run:
+        when: always
+        command: |
+          mkdir -p ~/test-results/spec
+          cp .junit/*.xml ~/test-results/spec/
+    - store_test_results:
+        path: ~/test-results
+    - store_artifacts:
+        path: ~/test-results/spec
 
 version: 2
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,20 +63,14 @@ jobs:
           path: ~/test-results/spec
 
   test_alpine:
-    docker:
-    - image: jrei/crystal-alpine:latest
+    machine: true
     environment:
       <<: *env
       TRAVIS_OS_NAME: linux
       ARCH: x86_64
       ARCH_CMD: linux64
-    steps:
-    - checkout
-    - run: echo 'export CURRENT_TAG="$CIRCLE_TAG"' >> $BASH_ENV
-    - run: bin/ci prepare_build
-    - run: make std_spec clean threads=1
-    - run: make crystal std_spec compiler_spec docs threads=1
-    - run: find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen
+      DOCKER_TEST_IMAGE: jrei/crystal-alpine:latest
+    steps: *ci_steps
 
   test_darwin:
     macos:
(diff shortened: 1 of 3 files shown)


j8r:
Great @straight-shoota ! You use my image 😮 
Considering Alpine is often used for static compilation, why not enabling it too here?
Another point, I hope my image will be put in the crystal-lang organization and will become official :)

straight-shoota:
Yes, we'll likely need to use a custom image from crystal org which gets updated on every release build and doesn't have to depend on apk being up to date.

But your image is fine for a preliminary exploration :+1:

ysbaddaden:
You can extract the static crystal binary from the official tarball. Here is a Dockerfile that will configure an Alpine Linux 3.9 image with everything needed to execute a complete workflow on CI:

```Dockerfile
FROM alpine:3.9

ARG VERSION=0.27.2
ARG REVISION=1
ARG ARCH=x86_64
ARG TARBALL=https://github.com/crystal-lang/crystal/releases/download/${VERSION}/crystal-${VERSION}-${REVISION}-linux-${ARCH}.tar.gz

RUN apk add --virtual .build-deps \
        build-base \
        curl \
        git

RUN apk add --virtual .crystal-deps \
        gc-dev \
        libevent-dev \
        libunwind-dev \
        llvm5-dev \
        pcre-dev

RUN apk add --virtual .crystal-specs-deps \
        gmp-dev \
        libxml2-dev \
        openssl-dev \
        readline-dev \
        yaml-dev \
        zlib-dev

RUN set -e; \
    curl -sL ${TARBALL} | tar xz -C /tmp; \
    cp /tmp/crystal-${VERSION}-${REVISION}/lib/crystal/bin/* /usr/local/bin/; \
    rm -rf /tmp/crystal-${VERSION}-${REVISION}
```

straight-shoota:
Unfortunately, the generic linux binary doesn't work on alpine/musl because the baked in gnu target triple (see #7196).

But if it did, that would exactly be the way I've described in https://github.com/crystal-lang/crystal/issues/6943#issuecomment-463179102

ysbaddaden:
@straight-shoota just `export FLAGS=--target $(llvm-config --host-target)`

We can't pass `--target` to `crystal run` *but* the Makefile only calls `crystal build` where we can pass `--target` and it will just work (verified).

straight-shoota:
Yes, macro runs is the main issue that can't be worked around (unless with `sed -i -e 's/x86_64-unknown-linux-gnu/x86_64-alpine-linux-musl/' /usr/local/bin/crystal`). The default target should not be baked in but depend on `ldd --version` (or `llvm-config --host-target`), at least to distinguish between linux-musl and linux-gnu.

ysbaddaden:
It would be great to have a minimal bootstrap compiler with just `build` command. No playground tools, or docs, not even run or eval. ping @asterite

asterite:
@ysbaddaden Yes, I always wanted to move those tools to separate binaries, even if it means having 4 or 5 "big" binaries (right now crystal is about 42MB, so maybe all of them would occupy at most 200MB, maybe less because each binary will be smaller). I'd really like that to happen. Also removing some parts of the std to separate shards (that could live in the `crystal-lang` organization, like XML, YAML, OAuth, etc.). But last thing I remember is that most didn't want to go into that direction (one thing was avoiding a more complex build/release that would involve multiple binaries, I don't think that's a big issue given how automated things are now).

j8r:
That's already nice to have moved out package management. I've worked to move out Crystal playground, apparently only few wanted to. I would work again on it if there is a consensus on it.

straight-shoota:
What does this all have to do with ci builds on alpine?

asterite:
I think nothing 😊

j8r:
I think because i complicates portability and porting on other environments.

ysbaddaden:
@straight-shoota a bootstrap compiler would allow to use the released static crystal binary on alpine, and overall ease cross platform development and experimentations :)

straight-shoota:
I don't see why that would make a difference. We just need to make the target host configurable for *all* compilation steps (thus including macro evaluation) and use a default depending on the environment. That's the same for both a small as well as a fat compiler.

asterite:
Macros run in the host, why would that need to be configurable?

straight-shoota:
Configurable as in "configures itself", instead of being baked in. So the same compiler binary can work with both musl and gnu by default (and without requiring any manual configuration).

straight-shoota:
Now the specs are running into `Failed to raise an exception: END_OF_STACK` https://circleci.com/gh/crystal-lang/crystal/19693?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Maybe this is related to #6934? Although it's not a static build.

j8r:
@straight-shoota do you still work on this? The CI can be re-triggered, mps has pushed Crystal 0.29 to Alpine, my image use this version.
This CI will be very useful to prevent issues like [this one](https://github.com/crystal-lang/crystal/pull/7871).

straight-shoota:
@j8r I've triggered a rebuild but it still fails for some reason I'm not going to investigate further. This is also an old branch which would need a rebase.

But anyway, this should build on crystal-lang/distribution-scripts#29 for a custom alpine-based Dockerfile so we don't need to rely on alpine package updates (which might take a few days after a Crystal release).

j8r:
Indeed, the image will have to change. I've removed g++ on the image, because only the Crystal compiler needs it. That's offtopic: why g++ over Clang?

j8r:
Any plan to have an Alpine CI any time soon?
I experimented [with Drone CI](https://github.com/crystal-lang/crystal/issues/8080#issuecomment-539579581), but it looks like the tests are too long to finish (cancelled because >1hour).

straight-shoota:
I'd like to move forward, but it is held up by crystal-lang/distribution-scripts#29 Waiting for feedback there.

straight-shoota:
This depends on crystal-lang/distribution-scripts#55 which is already used for building `crystallang/crystal:ci-test-dev` docker image.
There's one failure in stdlib specs: #7482

bcardiff:
It would be great to emit the junit output as other jobs. The `*ci_steps` does not have those instructions, it can be added there. 

NB: the *ci_steps is currently not used due to the workaround of splitting the specs. But let's improveit and keep it. The DRY will come back at some point, will come back at some point.

j8r:
Just a thought: if Alpine aarch64 CI is planned, wouldn't be better to have all Alpine CIs in GH Actions?
Since it is free, perhaps it would save some hours credits from Circle CI (I don't know the billing plan).

straight-shoota:
There are currently no plans for Alpine aarch64. At least I'm not aware of any. Duplicating the workflow on circleci is far easier integration than setting up a new GitHub action. And AFAIK circleci credits are not a concern atm.

So we should stick with this PR which is ready to merge. The important goal is to gain spec coverage linking against musl-libc. We can always reconsider whenever we get around implementing CI for alpine on aarch64. But that's currently not a high-priority concern. Maybe at that point we might even have advanced further towards a single, integrated CI platform.

bin/ci
```diff
@@ -148,6 +148,9 @@ with_build_env() {
     x86_64)
       export DOCKER_TEST_IMAGE="$DOCKER_TEST_PREFIX-build"
       ;;
+    x86_64-musl)
```
bcardiff: Shouldn't it be the other way around to mimic the target triple?

```suggestion
    musl-x86_64)
```

.circleci/config.yml
```diff
@@ -62,6 +71,16 @@ jobs:
...
           path: ~/test-results/spec
 
+  test_alpine:
+    machine: true
+    environment:
+      <<: *env
+      TRAVIS_OS_NAME: linux
+      ARCH: x86_64-musl
+      ARCH_CMD: linux64
+      DOCKER_TEST_PREFIX: crystallang/crystal:ci-test-dev # TODO: Remove after 0.33.0
```
bcardiff: I'm pushing a `0.32.1-alpine[-build]` with the changes of https://github.com/crystal-lang/distribution-scripts/pull/55 included. So this line can be removed and we can verify how it works.

bin/ci
straight-shoota: How does that mimic? The target triples are `x86_64-linux-gnu` and `x86_64-linux-musl`.

bcardiff: I got confused with the name of the .tar.gz which are *-linux-x86_64.

straight-shoota:
I've removed the explicit config, now the job uses the regular alpine image. This is ready to merge!

Issue #3825, opened 2017-01-03
Support for UDP Multicasting


jaitaiwan:
I spent a bit of time trying to get UDP multicasting happening with Crystal. What I learned is that it is theoretically possible to do with Crystal, but you need to understand the underlying system calls to make that happen.

So my request is for wrappers around the UDP socket to allow for easily joining multicast groups and sending broadcasts similar to Go's `net.ListenMulticastUDP` function and associated functionality so that one does not have to play with `setsockopts`.

Thanks!

ysbaddaden:
This is already supported. Just set `sock.broadcast = true` then connect the UDP socket. See [this spec](https://github.com/crystal-lang/crystal/blob/b1416e2b4e53e08bcc6769b6908f2eacff8b5e0a/spec/std/socket_spec.cr#L503) for example.

```crystal
sock = UDPSocket.new(Socket::Family::INET)
sock.broadcast = true
sock.connect("255.255.255.255", 1234)
sock.send("my broadcasted message")
```

Some documentation seems to be lacking, though.

jaitaiwan:
Are you on IRC @ysbaddaden?

I saw that broadcasts were supported which is fantastic but unless my knowledge is mistaken, to listen to a multicast address like say `239.255.60.60` you need to bind to `0.0.0.0` and then do something funky with `setsockopts` to join that multicast group?

ysbaddaden:
Then no, we don't provide a helper for the `IP_ADD_MEMBERSHIP` socket option.

PR #7423, opened 2019-02-13
add support for UDP multicast

stakach:
currently I've only added the LibC flags for Darwin but thought I'd get some feedback before proceeding further.

fixes: #3825

### DIFF
--- a/spec/std/socket/udp_socket_spec.cr
+++ b/spec/std/socket/udp_socket_spec.cr
@@ -1,8 +1,9 @@
 require "./spec_helper"
+require "../../support/errno"
 require "socket"
 
 describe UDPSocket do
-  each_ip_family do |family, address|
+  each_ip_family do |family, address, unspecified_address|
     it "#bind" do
       port = unused_local_port
       socket = UDPSocket.new(family)
@@ -51,6 +52,63 @@ describe UDPSocket do
 
       client.close
       server.close
+    end
+
+    if {{ flag?(:darwin) }} && family == Socket::Family::INET6
+      # Darwin is failing to join IPv6 multicast groups on older versions.
+      # However this is known to work on macOS Mojave with Darwin 18.2.0.
+      # Darwin also has a bug that prevents selecting the "default" interface.
+      # https://lists.apple.com/archives/darwin-kernel/2014/Mar/msg00012.html
+      pending "joins and transmits to multicast groups"
+    else
+      it "joins and transmits to multicast groups" do
+        udp = UDPSocket.new(family)
+        udp.bind(unspecified_address, 2000)
+
+        udp.multicast_loopback = false
+        udp.multicast_loopback?.should eq(false)
+
+        udp.multicast_hops = 4
+        udp.multicast_hops.should eq(4)
+        udp.multicast_hops = 0
+        udp.multicast_hops.should eq(0)
+
+        addr = case family
+               when Socket::Family::INET
+                 expect_raises(Socket::Error, "Unsupported IP address family: INET. For use with IPv6 only") do
+                   udp.multicast_interface 0
+                 end
+                 udp.multicast_interface Socket::IPAddress.new(unspecified_address, 0)
+
+                 Socket::IPAddress.new("224.0.0.254", 2000)
+               when Socket::Family::INET6
+                 expect_raises(Socket::Error, "Unsupported IP address family: INET6. For use with IPv4 only") do
+                   udp.multicast_interface(Socket::IPAddress.new(unspecified_address, 0))
+                 end
+                 udp.multicast_interface(0)
+
+                 Socket::IPAddress.new("ff02::102", 2000)
+               else
+                 raise "Unsupported IP address family: #{family}"
+               end
+
+        udp.join_group(addr)
+        udp.multicast_loopback = true
+        udp.multicast_loopback?.should eq(true)
+
+        udp.send("testing", addr)
+        udp.receive[0].should eq("testing")
+
+        udp.leave_group(addr)
+        udp.send("testing", addr)
+
+        # Test that nothing was received after leaving the multicast group
+        spawn do
+          sleep 100.milliseconds
+          udp.close
+        end
+        expect_raises_errno(Errno::EBADF, "Error receiving datagram: Bad file descriptor") { udp.receive }
+      end
     end
   end
 
--- a/src/socket/udp_socket.cr
+++ b/src/socket/udp_socket.cr
@@ -92,4 +92,148 @@ class UDPSocket < IPSocket
     bytes_read, sockaddr, addrlen = recvfrom(message)
     {bytes_read, IPAddress.from(sockaddr, addrlen)}
   end
+
+  # Reports whether transmitted multicast packets should be copied and sent
+  # back to the originator.
+  def multicast_loopback?
+    case @family
+    when Family::INET
+      getsockopt_bool LibC::IP_MULTICAST_LOOP, LibC::IPPROTO_IP
+    when Family::INET6
+      getsockopt_bool LibC::IPV6_MULTICAST_LOOP, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
+
+  # Sets whether transmitted multicast packets should be copied and sent back
+  # to the originator, if the host has joined the multicast group.
+  def multicast_loopback=(val : Bool)
+    case @family
+    when Family::INET
+      setsockopt_bool LibC::IP_MULTICAST_LOOP, val, LibC::IPPROTO_IP
+    when Family::INET6
+      setsockopt_bool LibC::IPV6_MULTICAST_LOOP, val, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
+
+  # Returns the current value of the `hoplimit` field on uni-cast packets.
+  # Datagrams with a `hoplimit` of `1` are not forwarded beyond the local network.
+  # Multicast datagrams with a `hoplimit` of `0` will not be transmitted on any
+  # network, but may be delivered locally if the sending host belongs to the
+  # destination group and multicast loopback is enabled.
+  def multicast_hops
+    case @family
+    when Family::INET
+      getsockopt LibC::IP_MULTICAST_TTL, 0, LibC::IPPROTO_IP
+    when Family::INET6
+      getsockopt LibC::IPV6_MULTICAST_HOPS, 0, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
+
+  # The multicast hops option controls the `hoplimit` field on uni-cast packets.
+  # If `-1` is specified, the kernel will use a default value.
+  # If a value of `0` to `255` is specified, the packet will have the specified
+  # value as `hoplimit`. Other values are considered invalid and `Errno` will be raised.
+  # Datagrams with a `hoplimit` of `1` are not forwarded beyond the local network.
+  # Multicast datagrams with a `hoplimit` of `0` will not be transmitted on any
+  # network, but may be delivered locally if the sending host belongs to the
+  # destination group and multicast loopback is enabled.
+  def multicast_hops=(val : Int)
+    case @family
+    when Family::INET
+      setsockopt LibC::IP_MULTICAST_TTL, val, LibC::IPPROTO_IP
+    when Family::INET6
+      setsockopt LibC::IPV6_MULTICAST_HOPS, val, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+    val
+  end
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This function overrides the default
+  # IPv4 interface address for subsequent transmissions. Setting the interface
+  # to `0.0.0.0` will select the default interface.
+  # Raises `Socket::Error` unless the socket is IPv4 and an IPv4 address is provided.
+  def multicast_interface(address : IPAddress)
+    if @family == Family::INET
+      if addr = address.@addr4
+        setsockopt LibC::IP_MULTICAST_IF, addr, LibC::IPPROTO_IP
+      else
+        raise Socket::Error.new "Expecting an IPv4 interface address. Address provided: #{address.address}"
+      end
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}. For use with IPv4 only"
+    end
+  end
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This function overrides the default
+  # IPv6 interface for subsequent transmissions. Setting the interface to
+  # index `0` will select the default interface.
+  def multicast_interface(index : UInt32)
+    if @family == Family::INET6
+      setsockopt LibC::IPV6_MULTICAST_IF, index, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}. For use with IPv6 only"
+    end
+  end
+
+  # A host must become a member of a multicast group before it can receive
+  # datagrams sent to the group.
+  # Raises `Socket::Error` if an incompatible address is provided.
+  def join_group(address : IPAddress)
+    case @family
+    when Family::INET
+      group_modify(address, LibC::IP_ADD_MEMBERSHIP)
+    when Family::INET6
+      group_modify(address, LibC::IPV6_JOIN_GROUP)
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
+
+  # Drops membership to the specified group. Memberships are automatically
+  # dropped when the socket is closed or the process exits.
+  # Raises `Socket::Error` if an incompatible address is provided.
+  def leave_group(address : IPAddress)
+    case @family
+    when Family::INET
+      group_modify(address, LibC::IP_DROP_MEMBERSHIP)
+    when Family::INET6
+      group_modify(address, LibC::IPV6_LEAVE_GROUP)
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
+
+  private def group_modify(ip, operation)
+    case @family
+    when Family::INET
+      if ip_addr = ip.@addr4
+        req = LibC::IpMreq.new
+        req.imr_multiaddr = ip_addr
+
+        setsockopt operation, req, LibC::IPPROTO_IP
+      else
+        raise Socket::Error.new "Expecting an IPv4 multicast address. Address provided: #{ip.address}"
+      end
+    when Family::INET6
+      if ip_addr = ip.@addr6
+        req = LibC::Ipv6Mreq.new
+        req.ipv6mr_multiaddr = ip_addr
+
+        setsockopt operation, req, LibC::IPPROTO_IPV6
+      else
+        raise Socket::Error.new "Expecting an IPv6 multicast address. Address provided: #{ip.address}"
+      end
+    else
+      raise Socket::Error.new "Unsupported IP address family: #{@family}"
+    end
+  end
 end
--- a/src/lib_c/aarch64-linux-gnu/c/netinet/in.cr
+++ b/src/lib_c/aarch64-linux-gnu/c/netinet/in.cr
@@ -3,6 +3,7 @@ require "../stdint"
 
 lib LibC
   IPPROTO_IP   =   0
+  IPPROTO_IPV6 =  41
   IPPROTO_ICMP =   1
   IPPROTO_RAW  = 255
   IPPROTO_TCP  =   6
@@ -38,5 +39,30 @@ lib LibC
     sin6_flowinfo : UInt32T
     sin6_addr : In6Addr
     sin6_scope_id : UInt32T
+  end
+
+  IP_MULTICAST_IF   = 32
+  IPV6_MULTICAST_IF = 17
+
+  IP_MULTICAST_TTL    = 33
+  IPV6_MULTICAST_HOPS = 18
+
+  IP_MULTICAST_LOOP   = 34
+  IPV6_MULTICAST_LOOP = 19
+
+  IP_ADD_MEMBERSHIP = 35
+  IPV6_JOIN_GROUP   = 20
+
+  IP_DROP_MEMBERSHIP = 36
+  IPV6_LEAVE_GROUP   = 21
+
+  struct IpMreq
+    imr_multiaddr : InAddr
+    imr_interface : InAddr
+  end
+
+  struct Ipv6Mreq
+    ipv6mr_multiaddr : In6Addr
+    ipv6mr_interface : UInt
   end
 end
--- a/src/lib_c/aarch64-linux-musl/c/netinet/in.cr
+++ b/src/lib_c/aarch64-linux-musl/c/netinet/in.cr
@@ -3,6 +3,7 @@ require "../stdint"
 
 lib LibC
   IPPROTO_IP   =   0
+  IPPROTO_IPV6 =  41
   IPPROTO_ICMP =   1
   IPPROTO_RAW  = 255
   IPPROTO_TCP  =   6
@@ -38,5 +39,30 @@ lib LibC
     sin6_flowinfo : UInt32T
     sin6_addr : In6Addr
     sin6_scope_id : UInt32T
+  end
+
+  IP_MULTICAST_IF   = 32
+  IPV6_MULTICAST_IF = 17
+
+  IP_MULTICAST_TTL    = 33
+  IPV6_MULTICAST_HOPS = 18
+
+  IP_MULTICAST_LOOP   = 34
+  IPV6_MULTICAST_LOOP = 19
+
+  IP_ADD_MEMBERSHIP = 35
+  IPV6_JOIN_GROUP   = 20
+
+  IP_DROP_MEMBERSHIP = 36
+  IPV6_LEAVE_GROUP   = 21
+
+  struct IpMreq
+    imr_multiaddr : InAddr
+    imr_interface : InAddr
+  end
+
+  struct Ipv6Mreq
+    ipv6mr_multiaddr : In6Addr
+    ipv6mr_interface : UInt
   end
 end
--- a/src/lib_c/arm-linux-gnueabihf/c/netinet/in.cr
+++ b/src/lib_c/arm-linux-gnueabihf/c/netinet/in.cr
@@ -3,6 +3,7 @@ require "../stdint"
 
 lib LibC
   IPPROTO_IP   =   0
+  IPPROTO_IPV6 =  41
   IPPROTO_ICMP =   1
   IPPROTO_RAW  = 255
   IPPROTO_TCP  =   6
@@ -38,5 +39,30 @@ lib LibC
     sin6_flowinfo : UInt32T
     sin6_addr : In6Addr
     sin6_scope_id : UInt32T
+  end
+
+  IP_MULTICAST_IF   = 32
+  IPV6_MULTICAST_IF = 17
+
+  IP_MULTICAST_TTL    = 33
+  IPV6_MULTICAST_HOPS = 18
+
+  IP_MULTICAST_LOOP   = 34
+  IPV6_MULTICAST_LOOP = 19
+
+  IP_ADD_MEMBERSHIP = 35
+  IPV6_JOIN_GROUP   = 20
+
+  IP_DROP_MEMBERSHIP = 36
+  IPV6_LEAVE_GROUP   = 21
+
+  struct IpMreq
+    imr_multiaddr : InAddr
+    imr_interface : InAddr
+  end
+
+  struct Ipv6Mreq
+    ipv6mr_multiaddr : In6Addr
+    ipv6mr_interface : UInt
   end
 end
--- a/src/lib_c/i386-linux-gnu/c/netinet/in.cr
+++ b/src/lib_c/i386-linux-gnu/c/netinet/in.cr
@@ -3,6 +3,7 @@ require "../stdint"
 
 lib LibC
   IPPROTO_IP   =   0
+  IPPROTO_IPV6 =  41
   IPPROTO_ICMP =   1
   IPPROTO_RAW  = 255
   IPPROTO_TCP  =   6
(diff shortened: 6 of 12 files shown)


jaitaiwan:
Awesome! I'll see about what needs to be done for Linux...

stakach:
ping @ysbaddaden what do you think?

@RX14 the Windows implementation sockets, the stuff in `netinet/in.cr` for BSD and linux, is almost identical to the BSD implementation, two minor differences. Was thinking of adding a `winsock2.cr` file with that in it, but not sure if it will compile / don't want to break anything you are working on.
Definitions for things like `SaFamilyT` are probably missing on Windows.
Did you want me to include it?

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
     bytes_read, sockaddr, addrlen = recvfrom(message)
     {bytes_read, IPAddress.from(sockaddr, addrlen)}
   end
+
+  # multicast loopback reports whether transmitted multicast packets
+  # should be copied and send back to the originator.
```
Sija:
```suggestion
  # should be copied and sent back to the originator.
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
     bytes_read, sockaddr, addrlen = recvfrom(message)
     {bytes_read, IPAddress.from(sockaddr, addrlen)}
   end
+
+  # multicast loopback reports whether transmitted multicast packets
```
Sija:
```suggestion
  # Multicast loopback reports whether transmitted multicast packets
```

src/socket/udp_socket.cr
Sija:
```suggestion
  # should be copied and sent back to the originator.
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+      optname = LibC::IP_MULTICAST_LOOP
+      level = LibC::IPPROTO_IP
+    else
+      optname = LibC::IPV6_MULTICAST_LOOP
+      level = LibC::IPPROTO_IPV6
+    end
+    getsockopt_bool optname, level
+  end
+
+  # multicast loopback sets whether transmitted multicast packets should be
```
Sija:
```suggestion
  # Multicast loopback sets whether transmitted multicast packets should be
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+  # copied and send back to the originator, if the host has joined the
```
Sija:
```suggestion
  # copied and sent back to the originator, if the host has joined the
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+  # multicast group
```
Sija:
```suggestion
  # multicast group.
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+      optname = LibC::IP_MULTICAST_TTL
+      level = LibC::IPPROTO_IP
+    else
+      optname = LibC::IPV6_MULTICAST_HOPS
+      level = LibC::IPPROTO_IPV6
+    end
+    getsockopt optname, 0, level
+  end
+
+  # The multicast hops option controls hoplimit field on uni-cast packets.
```
Sija:
```suggestion
  # The multicast hops option controls `hoplimit` field on uni-cast packets.
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+    else
+      optname = LibC::IPV6_MULTICAST_HOPS
+      level = LibC::IPPROTO_IPV6
+    end
+    setsockopt optname, val, level
+    val
+  end
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This fucntion over-rides the default
```
Sija:
```suggestion
  # from the primary network interface. This function overrides the default
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+  # for subsequent transmissions. Setting the interface to "0.0.0.0" for IPv4
```
Sija:
```suggestion
  # for subsequent transmissions. Setting the interface to `0.0.0.0` for IPv4
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+      optname = LibC::IPV6_MULTICAST_HOPS
+      level = LibC::IPPROTO_IPV6
+    end
+    getsockopt optname, 0, level
+  end
+
+  # The multicast hops option controls hoplimit field on uni-cast packets.
+  # If -1 is specified, the kernel will use a default value.
+  # If a value of 0 to 255 is specified, the packet will have the specified
+  # value as hoplimit.  Other values are considered invalid.
```
Sija:
```suggestion
  # value as `hoplimit`. Other values are considered invalid.
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+  # Datagrams with a hoplimit of 1 are not forwarded beyond the local net-
+  # work.  Multicast datagrams with a hoplimit of 0 will not be transmitted
```
Sija:
```suggestion
  # work. Multicast datagrams with a `hoplimit` of 0 will not be transmitted
```

stakach:
Thanks @Sija I've fixed those issues with the comments

Sija:
@stakach Great! I'd also put all integer values inside backticks, since they ought to be treated literally.

stakach:
@ysbaddaden I've made those changes.
I believe the bindings are correct, what's wrong with the OpenBSD bindings?

I was looking in 
* https://github.com/openbsd/src/blob/master/sys/netinet/in.h
* https://man.openbsd.org/ip6#IPV6_JOIN_GROUP

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+    group_modify(ip, true)
+  end
+
+  # Drops membership to the specified group. Memberships are automatically
+  # dropped when the socket is closed or the process exits.
+  def leave_group(ip : IPAddress)
+    group_modify(ip, false)
+  end
+
+  protected def group_modify(ip, operation)
```
straight-shoota: Why is this protected instead of private?

ysbaddaden:
I can't really review. I don't know anything about multicast —for example I have no idea what "hops" is.

I looked at how Ruby handles that, but it doesn't have any high-level interface for multicast in stdlib, so it doesn't help either :(

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,117 @@ class UDPSocket < IPSocket
...
+    end
+    setsockopt optname, val, level
+    val
+  end
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This function overrides the default
+  # for subsequent transmissions. Setting the interface to `0.0.0.0` for IPv4
+  # or index `0` for IPv6 will select the default interface.
+  def multicast_interface(ifout : IPAddress | UInt32)
```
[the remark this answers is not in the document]
stakach: I disagree with this to some degree. Firstly almost no one is going to use this function, however for those who need it, it should work for both IPv4 and v6.

As ethernet ports would be assigned to VLans which might run DHCP, the IPv6 index method to me makes the most sense. If you were to use this function, you'd be enumerating the network cards on your machine and then setting an ENV var, or equivalent configuration, that would be picked up by the application (it's unlikely to need the application to auto detect anything)

What might be a cleaner solution is to split it up into two functions - as it's the only method who's interfaces are not equivalent between IPv4 and 6

i.e. `ipv4_multicast_interface(ifout : IPAddress)` and `ipv6_multicast_interface(ifout : UInt32)`
But I think this is actually more ugly and anyone using this method should already be aware of the differences between IP implementations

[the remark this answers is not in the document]
stakach: I did update this function to accept IPv6 ANY as a valid param for IPv6 though.

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,30 @@ describe UDPSocket do
...
+
+      udp.multicast_hops = 4
+      udp.multicast_hops.should eq(4)
+      udp.multicast_hops = 1
+      udp.multicast_hops.should eq(1)
+
+      udp.multicast_interface Socket::IPAddress.new("0.0.0.0", 0)
+
+      addr = Socket::IPAddress.new("224.0.0.254", 2000)
+      udp.join_group(addr)
```
wontruefree: should more than 1 addr be in the group to check that multicast can reach both of them?  also to distinguish from broadcast it might make sense to have an addr and check that it is not receiving anything.

stakach: Broadcast isn't enabled and I don't think we need to complicate the test. We're not testing the OS level socket behaviour, we're testing crystals ability to configure a socket - which the test achieves.

wontruefree:
This is really cool to see in Crystal.  Thanks for working on this.

src/socket/udp_socket.cr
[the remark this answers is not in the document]
stakach: thoughts @straight-shoota ?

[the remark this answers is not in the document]
straight-shoota: Sry, I don't have any decisive opinion on this.
I guess we can keep the `UInt32` argument, but I think it should better be two separate methods, because they're working differently. This would also make it easier to document because the different behaviour is clearly separated.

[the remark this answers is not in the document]
stakach: OK. I'll separate them :)

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,139 @@ class UDPSocket < IPSocket
...
+    val
+  end
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This function overrides the default
+  # IPv4 interface address for subsequent transmissions. Setting the interface
+  # to `0.0.0.0` will select the default interface.
+  def multicast_interface(address : IPAddress)
+    if @family == Family::INET
+      setsockopt LibC::IP_MULTICAST_IF, address.addr4.not_nil!, LibC::IPPROTO_IP
```
straight-shoota: When `address` is a not a v4 address, this would raise `NilAssertionFailed`. Could you make sure that `address.addr4` is valid and otherwise raise a custom error?

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,139 @@ class UDPSocket < IPSocket
...
+  def multicast_loopback?
+    case @family
+    when Family::INET
+      optname = LibC::IP_MULTICAST_LOOP
+      level = LibC::IPPROTO_IP
+    when Family::INET6
+      optname = LibC::IPV6_MULTICAST_LOOP
+      level = LibC::IPPROTO_IPV6
+    else
+      raise "Unsupported IP address family: #{family}"
```
straight-shoota: This should raise a more specific `Socket::Error`. DItto for all other generic `raise` calls.

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,139 @@ class UDPSocket < IPSocket
     bytes_read, sockaddr, addrlen = recvfrom(message)
     {bytes_read, IPAddress.from(sockaddr, addrlen)}
   end
+
+  # Reports whether transmitted multicast packets should be copied and sent
+  # back to the originator.
+  def multicast_loopback?
+    case @family
+    when Family::INET
+      optname = LibC::IP_MULTICAST_LOOP
```
straight-shoota: It would be better if you simply duplicate the call to `getsockopt_bool` in both branches using the constants directly as arguments. This makes the code shorter overall and easier to follow. Ditto all below.

src/socket/udp_socket.cr
straight-shoota: Also it would be nice if you'd consistently use either `family` or `@family`.

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,139 @@ class UDPSocket < IPSocket
...
+  def leave_group(ip : IPAddress)
+    group_modify(ip, false)
+  end
+
+  private def group_modify(ip, operation)
+    resp = case ip.family
+           when Family::INET6
+             operation = operation ? LibC::IPV6_JOIN_GROUP : LibC::IPV6_LEAVE_GROUP
+             req = LibC::Ipv6Mreq.new
+             req.ipv6mr_multiaddr = ip.addr6.not_nil!
```
straight-shoota: This nil check should be guarded to provide a specific error when a non-matching address family is used. Ditto below.

stakach:
Great feedback @straight-shoota - I've implemented all those changes.

src/socket/udp_socket.cr
```diff
@@ -155,6 +155,7 @@ class UDPSocket < IPSocket
   # from the primary network interface. This function overrides the default
   # IPv4 interface address for subsequent transmissions. Setting the interface
   # to `0.0.0.0` will select the default interface.
+  # Raises `Socket::Error` unless the socket is IPv4 and an IPv4 address is provided
```
straight-shoota: A set of periods would be nice. :+1:

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,30 @@ describe UDPSocket do
       client.close
       server.close
     end
+
+    it "can join and transmit to multicast groups" do
+      udp = UDPSocket.new Socket::Family::INET
```
straight-shoota: There should also be a test for IPv6. A simple solution would be to use `family` here (it comes from `each_ip_family` in line 5). The any address (for the next line) is provided as fourth argument to the block by `each_ip_family`.

Only the lines for `multicast_interface` and goup handling need to be duplicated and distinguished between v4 and v6.

straight-shoota: Perhaps it would be better then to just duplicate the entire spec or split the familiy-specific parts to separate specs. Maybe that's even better to have smaller specs focused on single features.

stakach: Ahh nice, didn't notice that. Will update.

stakach:
@straight-shoota looking good with the IPv6 tests in place.

stakach:
@straight-shoota so not sure how to resolve the OSX tests for IPv6.

I dev on OSX so can confirm that setting the default multicast interface on `Darwin 18.2.0` does not work as described in this bug: https://lists.apple.com/archives/darwin-kernel/2014/Mar/msg00012.html

However I can join the group successfully on my laptop (with IPv6) using crystal so I assume it's either an older version running on circle ci or possibly a permissions thing? not sure..

I see there is a similar issue with the `it "sends broadcast message" do` test for darwin sockets, so I think it's probably related.

straight-shoota:
Yeah, the CI uses xcode version 9.0 which runs on [macOS 10.12.6 / Darwin 16.7.0](https://circle-macos-docs.s3.amazonaws.com/image-manifest/build-282/index.html). The bug has probably been fixed somewhere in between. We could consider simply updating the CI, but I'm not sure whether that's possible or desired. We aim to support macOS 10.7+ and the specs should work on all supported versions.

So for now, it's probably okay to simply disable the spec for `#multicast_interface`  with IPv6 on darwin. But the rest of the spec should run on this platform!

I'd suggest to structure the tests like this, possibly even separating the individual methods (loopback, hops, join/leave) into separate specs.

```cr
each_ip_family do 
  # ...
  it "can join and transmit to multicast groups" do
    # setup
    # multicast_loopback
    # multicast_hops
    # {join,leave}_group + send/receive
  end
end
describe "#multicast_interface" do
  it "on IPv4" do
    # setup ipv4
    udp.multicast_interface Socket::IPAddress.new("0.0.0.0", 0)
    # send/receive test
  end
  {% unless flag?(:darwin) %}
    it "on IPv6" do
      # setup ipv6
      udp.multicast_interface 0
      # send/receive test
    end
  {% end %}
end      
```

We could consider activating the darwin spec based on the actual OS version. But for this it would be great to know when this was fixed.

`leave_group` is currently also missing from the specs, please add some expectation for that. Should also make sure that send/receive doesn't work after leaving.

stakach:
@straight-shoota all specs are passing for all platforms except Darwin IPv6 which displays as pending now.
image

I've just added the test for group leave.

`multicast_interface`, `group_join` and `group_leave` IPv6 don't work on the CI version of Darwin unfortunately.
I think the fix for `group_join` and `group_leave` was introduced in late 2017: https://github.com/apple/darwin-xnu/commit/0a798f6738bc1db01281fc08ae024145e84df927#diff-e2c4bcf26a9843a55acac3eb88326083

stakach:
@straight-shoota what do you think? Ready for merge?
Would love to see this in the next release

straight-shoota:
Please separate the specs and make sure that the working specs run on darwin as suggested in my previous comment.

I think I'd also prefer to run all specs on recent darwin versions. This would just need to be a dynamic check based on `sw_vers` combined with `flag?(:darwin)`. This would be the best solution, even without exactly pinpointing when it was fixed.

stakach:
@straight-shoota was just looking at this again.
Having no luck using `sw_vers` to determine if this version of OSX supports supports what. Have a feeling this issue lies not in the OS level code, but in the user level libraries supplied by XCode.

So I was thinking:

1. Would it be OK if the code in the current state is accepted?
   * One test is pending: IPv6 multicast on Darwin
2. I open a new pull request to implement [getifaddrs](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getifaddrs.3.html)
   * This will allow me to implement the missing behaviour of Darwin [manually](https://github.com/pothosware/SoapyRemote/commit/43a5befce9ab1d76e9abb3a70b268881da225767)
   * I can fix IPv6 on OSX as part of that pull request - otherwise I feel like it's going to make this pull request very messy

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,58 @@ describe UDPSocket do
       client.close
       server.close
     end
+
+    is_darwin = {% if flag?(:darwin) %} true {% else %} false {% end %}
```
Sija:
```suggestion
    is_darwin = {{ flag?(:darwin) }}
```

stakach: nice one

straight-shoota:
Yes of course, it's fine to simply skip the spec on darwin for now if a dynamic condition is too complicated.` 

But as far as I understand it, only `#multicast_interface` fails, so the other specs should be executed on darwin with IPv6. A structure as proposed in https://github.com/crystal-lang/crystal/pull/7423#issuecomment-470063033 would be ideal for this.

stakach:
@straight-shoota it works on the latest version of OSX with what are probably the latest XCode libraries?

This is a minimal reproduction that can you can use to test on your laptop if you have OSX
(this works for me - but not on CircleCI)

```crystal
require "socket"

struct Socket::IPAddress
  def addr6
    @addr6
  end
end

lib LibC
struct Ipv6Mreq
    ipv6mr_multiaddr : LibC::In6Addr
    ipv6mr_interface : LibC::UInt
  end
end

s = UDPSocket.new Socket::Family::INET6
s.bind "::", 0
# multicast_interface selection
#s.setsockopt 9, 1, 41
#s.getsockopt 9, 41
sock = Socket::IPAddress.new("ff02::102", 2000)
addr = sock.addr6.not_nil!
  
req = LibC::Ipv6Mreq.new
req.ipv6mr_multiaddr = addr
req.ipv6mr_interface = 0
# Join IPv6 group
s.setsockopt 12, req, 41
Fiber.yield
s.send("testing", sock)
s.receive[0]
```

Thanks, I'll rebase.

stakach:
@straight-shoota I didn't split the tests as none of the IPv6 multicast functions work on CircleCI version of Darwin or the build tools running on it.
I added [this check](https://github.com/crystal-lang/crystal/pull/7423/files#diff-aa6c3f436b2862e1cc16359a2da17ae9R86), which should have worked based on my setup (you can confirm if you have OSX with the code I posted above)

I'll implement `getifaddrs` which should allow us to use either IP address or index for both IPv4 and v6 `multicast_interface` selection as well as working around the limitations of IPv6 on darwin.

Given not many people will be using `multicast_interface` selection and IPv6 multicast works on the latest versions of OSX (even if it's not being tested currently) I think this pull request is good to go

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,58 @@ describe UDPSocket do
...
+               when Socket::Family::INET
+                 expect_raises(Socket::Error) { udp.multicast_interface 0 }
+                 udp.multicast_interface Socket::IPAddress.new(unspecified_address, 0)
+                 Socket::IPAddress.new("224.0.0.254", 2000)
+               when Socket::Family::INET6
+                 expect_raises(Socket::Error) { udp.multicast_interface(Socket::IPAddress.new(unspecified_address, 0)) }
+
+                 # Darwin has a bug that prevents selecting the "default" interface.
+                 # https://lists.apple.com/archives/darwin-kernel/2014/Mar/msg00012.html
+                 udp.multicast_interface(0) unless is_darwin
```
straight-shoota: You can omit this check. `is_darwin` can never be true here, because IPv6 on darwin is already skipped in line 57.

stakach:
@straight-shoota what do we think?

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,58 @@ describe UDPSocket do
...
+        udp.multicast_loopback?.should eq(true)
+
+        udp.multicast_hops = 4
+        udp.multicast_hops.should eq(4)
+        udp.multicast_hops = 1
+        udp.multicast_hops.should eq(1)
+
+        addr = case family
+               when Socket::Family::INET
+                 expect_raises(Socket::Error) { udp.multicast_interface 0 }
```
straight-shoota:
```suggestion
                 expect_raises(Socket::Error, "Unsupported IP address family") { udp.multicast_interface 0 }
```

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,58 @@ describe UDPSocket do
...
+                 udp.multicast_interface Socket::IPAddress.new(unspecified_address, 0)
+
+                 Socket::IPAddress.new("224.0.0.254", 2000)
+               when Socket::Family::INET6
+                 expect_raises(Socket::Error) { udp.multicast_interface(Socket::IPAddress.new(unspecified_address, 0)) }
```
straight-shoota:
```suggestion
                 expect_raises(Socket::Error, "Expecting an IPv4 interface address") { udp.multicast_interface(Socket::IPAddress.new(unspecified_address, 0)) }
```

(probably better to split these lines)

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +52,58 @@ describe UDPSocket do
...
+
+        udp.leave_group(addr)
+        udp.send("testing", addr)
+
+        # Test that nothing was received after leaving the multicast group
+        spawn do
+          sleep 1
+          udp.close
+        end
+        expect_raises(Errno) { udp.receive }
```
straight-shoota: Please use `expect_raises_errno` from `support/errno` to validate the error code.

src/socket/address.cr
```diff
@@ -80,10 +80,10 @@ class Socket
     BROADCAST6   = "ff0X::1"
 
     getter port : Int32
+    getter addr6 : LibC::In6Addr?
```
straight-shoota: These ivars are implementation details. They're not part of the public interface on purpose.

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,142 @@ class UDPSocket < IPSocket
...
+  # Raises `Socket::Error` unless the socket is IPv4 and an IPv4 address is provided.
+  def multicast_interface(address : IPAddress)
+    if @family == Family::INET
+      if addr = address.addr4
+        setsockopt LibC::IP_MULTICAST_IF, addr, LibC::IPPROTO_IP
+      else
+        raise Socket::Error.new "Expecting an IPv4 interface address. Address provided: #{address.address}"
+      end
+    else
+      raise Socket::Error.new "For use with IPv4 only. Unsupported IP address family: #{@family}"
```
straight-shoota: The reason for failure should come first:

```suggestion
      raise Socket::Error.new "Unsupported IP address family: #{@family}. For use with IPv4 only"
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,142 @@ class UDPSocket < IPSocket
...
+
+  # For hosts with multiple interfaces, each multicast transmission is sent
+  # from the primary network interface. This function overrides the default
+  # IPv6 interface for subsequent transmissions. Setting the interface to
+  # index `0` will select the default interface.
+  def multicast_interface(index : UInt32)
+    if @family == Family::INET6
+      setsockopt LibC::IPV6_MULTICAST_IF, index, LibC::IPPROTO_IPV6
+    else
+      raise Socket::Error.new "For use with IPv6 only. Unsupported IP address family: #{@family}"
```
straight-shoota:
```suggestion
      raise Socket::Error.new "Unsupported IP address family: #{@family}. For use with IPv6 only"
```

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,142 @@ class UDPSocket < IPSocket
...
+    end
+  end
+
+  # A host must become a member of a multicast group before it can receive
+  # datagrams sent to the group.
+  # Raises `Socket::Error` if an incompatible address is provided.
+  def join_group(address : IPAddress)
+    group_modify(address, true)
```
straight-shoota: It would be ore concise if `operation` directly takes the appropriate value (e.g. `LibC::IPV6_JOIN_GROUP`) instead of an indirection over a boolean flag.

src/socket/udp_socket.cr
```diff
@@ -88,4 +88,142 @@ class UDPSocket < IPSocket
...
+               req.imr_multiaddr = ip_addr
+
+               setsockopt operation, req, LibC::IPPROTO_IP
+             else
+               raise Socket::Error.new "Expecting an IPv4 multicast address. Address provided: #{ip.address}"
+             end
+           else
+             raise Socket::Error.new "Unsupported IP address family: #{@family}"
+           end
+    # We yield so that when we perform a send we'll receive the loopback if that
```
straight-shoota:
```suggestion
    # FIXME: We yield so that when we perform a send we'll receive the loopback if that
```

src/socket/address.cr
straight-shoota: I meant, these getters shouldn't be added at all, neither as documented nor undocumented methods. You can use the ivars directly in your code.

stakach: How do I do that?

r00ster91: Simply revert the changes of this file: remove the getters and readd:
```cr
@addr6 : LibC::In6Addr?
@addr4 : LibC::InAddr?
```

stakach: wow had no idea you could do that.
Sweet

stakach:
@straight-shoota I've implemented all those requested changes. Thank you for your patience btw.

asterite:
@stakach it's actually your patience that we must thank, others would have probably abandoned this PR already :-)

stakach:
need me to do anything more for this pull to be merged?

straight-shoota:
Would be great if you could rebase against current master, to make sure it still works fine with 0.28.0-dev.

stakach:
@straight-shoota looking good :)

stakach:
@ysbaddaden @asterite would one of you be available to review this pull?

spec/std/socket/udp_socket_spec.cr
```diff
@@ -2,7 +2,7 @@ require "./spec_helper"
 require "socket"
```
bcardiff: The rest of the specs in this directory uses

```crystal
require "./spec_helper"
require "../../support/errno"
```

And that will allow running this file independently

```shell
$ ./bin/crystal spec spec/std/socket/udp_socket_spec.cr 
```

(I have Darwin 18.0.0 still and as the comment stated it requires 18.2.0)

stakach: Good catch, I've added that include

stakach:
The spec error for linux32 seems unrelated to this pull

src/socket/udp_socket.cr
```diff
@@ -92,4 +92,154 @@ class UDPSocket < IPSocket
...
+               req.ipv6mr_multiaddr = ip_addr
+
+               setsockopt operation, req, LibC::IPPROTO_IPV6
+             else
+               raise Socket::Error.new "Expecting an IPv6 multicast address. Address provided: #{ip.address}"
+             end
+           else
+             raise Socket::Error.new "Unsupported IP address family: #{@family}"
+           end
+    # FIXME: We yield so that when we perform a send we'll receive the loopback if
```
asterite: Was this discussed at some point? What code fails without this? Maybe there is a way to avoid `Fiber.yield` here.

(but it can be discussed in a separate issue if everything is working fine)

stakach: @asterite in my testing I found that without the yield the following wouldn't work as expected:

```crystal
# Loop back indicates we want to receive anything we send back on the same socket
udp.multicast_loopback = true
udp.join_group(addr) # (yielding here)
udp.send("testing", addr)

# without the yield when joining the group above then the multicast send above succeeds but is not received locally even though loopback is enabled
udp.receive[0].should eq("testing")
```

The processes handling multicast in the background seem to be quite complicated so I assume it's related to some kernel data-structures but I am not certain
https://medium.com/tcp-ip/igmp-query-group-join-leave-681ee7a00db1
https://www.eventhelix.com/RealtimeMantra/Networking/igmp/igmp-query-join-leave.pdf

spec/std/socket/udp_socket_spec.cr
```diff
@@ -52,6 +53,62 @@ describe UDPSocket do
...
+        udp.send("testing", addr)
+        udp.receive[0].should eq("testing")
+
+        udp.leave_group(addr)
+        udp.send("testing", addr)
+
+        # Test that nothing was received after leaving the multicast group
+        spawn do
+          sleep 1
+          udp.close
```
asterite: Why is this done in `spawn`? And having `sleep 1` in specs is not ideal...

src/socket/udp_socket.cr
asterite: Okay, so what `Fiber.yield` does is to put the current fiber to sleep and see if another fiber can execute. If that's the case the other fiber executes until it gets blocked, then other fibers (like the first one) resume execution.

I guess what I'm not seeing here is is that in your code there's just a single fiber, so `Fiber.yield` would have no effect, except maybe waiting a few (milli/nano)seconds to resume execution. That's why I think that `Fiber.yield` there is a big smell. There's probably some misunderstanding around how this behaves and it's probably not good to introduce something that we don't understand.

spec/std/socket/udp_socket_spec.cr
stakach: We're checking that we've left the multicast group and I couldn't think of any other way to check for the lack of something.

Otherwise we'd just hang on receive, waiting for nothing as we've left the group.
So the sleep is to ensure that receive is not going to receive anything and then we close the socket which raises the error.

stakach: We could probably lower the wait time to 100 milliseconds?

src/socket/udp_socket.cr
stakach: So remove the fiber yield?
My feeling is that the loopback isn't enabled until the IGMP join group packet is transmitted.

So the tests might pass if I remove this yield and set `multicast_hops = 0` so that the packet send isn't required.

stakach:
@asterite I think this should resolve your concerns.

Sija:
Merge 🕦?

bcardiff:
Thanks @stakach and all reviewers involved!

PR #7424, opened 2019-02-13
Implement resource owner credentials

### LINKED ISSUE #7086: Support OAuth 2.0 Resource Owner Password Credentials Grant Type
Please include support for the OAuth 2.0 Resource Owner Password Credentials grant type in the `OAuth2::Client` class. The class currently includes support for Authorization Code and Client Credentials grant types, but not Resource Owner Password Credentials.

The Resource Owner Password Credentials grant type is defined in the OAuth 2.0 standard IETF RFC 6749 Section 1.3.3:

https://tools.ietf.org/html/rfc6749#section-1.3.3

This is supported in other OAuth 2.0 libs including:

Ruby `oauth2` gem:

[https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2/client.rb#L165-L170](https://github.com/oauth-xx/oauth2/blob/b83b65c5b45d63cb5ca99e568db6b7aec0dbe23e/lib/oauth2/client.rb#L165-L170)

Go `oauth2` package:

https://godoc.org/golang.org/x/oauth2#Config.PasswordCredentialsToken

Blacksmoke16:
Resolves #7086

* Orders methods in order they appear in the RFC
* Implements `get_access_token_using_resource_owner_credentials`
* Adds `"Content-Type" => "application/x-www-form-urlencoded"` to better adhere to the RFC.

I think it would be nice to test the `get_access_token_using_*` methods create the proper request.  However not sure how to best handle that without setting up a mock HTTP server...

### DIFF
--- a/spec/std/http/spec_helper.cr
+++ b/spec/std/http/spec_helper.cr
@@ -7,7 +7,7 @@ private def wait_for(timeout = 5.seconds)
     Fiber.yield
 
     if (Time.monotonic - now) > timeout
-      raise "server failed to start within 5 seconds"
+      raise "server failed to start within #{timeout}"
     end
   end
 end
--- a/spec/std/http/spec_helper.cr
+++ b/spec/std/http/spec_helper.cr
@@ -1,0 +1,46 @@
+require "spec"
+
+private def wait_for(timeout = 5.seconds)
+  now = Time.monotonic
+
+  until yield
+    Fiber.yield
+
+    if (Time.monotonic - now) > timeout
+      raise "server failed to start within 5 seconds"
+    end
+  end
+end
+
+# Helper method which runs *server*
+# 1. Spawns `server.listen` in a new fiber.
+# 2. Waits until `server.listening?`.
+# 3. Yields to the given block.
+# 4. Ensures the server is closed.
+# 5. After returning from the block, it waits for the server to gracefully
+#    shut down before continuing execution in the current fiber.
+# 6. If the listening fiber raises an exception, it is rescued and re-raised
+#    in the current fiber.
+def run_server(server)
+  server_done = Channel(Exception?).new
+
+  spawn do
+    server.listen
+  rescue exc
+    server_done.send exc
+  else
+    server_done.send nil
+  end
+
+  begin
+    wait_for { server.listening? }
+
+    yield server_done
+  ensure
+    server.close unless server.closed?
+
+    if exc = server_done.receive
+      raise exc
+    end
+  end
+end
--- a/spec/std/oauth2/client_spec.cr
+++ b/spec/std/oauth2/client_spec.cr
@@ -44,5 +44,7 @@ describe OAuth2::Client do
     client.get_access_token_using_refresh_token("some_refresh_token")
     client.get_access_token_using_refresh_token("some_refresh_token", scope: "some scope")
     client.get_access_token_using_client_credentials(scope: "some scope")
+    client.get_access_token_using_resource_owner_credentials(username: "user123", password: "monkey")
+    client.get_access_token_using_resource_owner_credentials(username: "user123", password: "monkey", scope: "foo")
   end)
 end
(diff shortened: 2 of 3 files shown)


src/oauth2/client.cr
```diff
@@ -106,42 +106,54 @@ class OAuth2::Client
 
   # Gets an access token using an authorization code, as specified by
   # [RFC 6749, Section 4.1.1](https://tools.ietf.org/html/rfc6749#section-4.1.3).
```
r00ster91: What about also fixing this link? In the text it says "Section 4.1.1" but in the URL it says "section-4.1.3".

straight-shoota:
> However not sure how to best handle that without setting up a mock HTTP server...

There is nothing wrong with that. OAuth is an HTTP-based protocol, so in order to test it, you'll need an HTTP server.
You could get around going through the network stack by modifying `get_access_token` in some way to make it not strictly rely on `HTTP::Client.get`, but I don't think that's worth it.
Just spin up an HTTP server for testing. Examples for this can be found in the HTTP specs.

asterite:
It would be nice not having to spin up an HTTP::Server for testing what request should be generated. We need some kind of thing that we can mock that receives a request and returns a response. Maybe that concept is called a Transport in Go, I'm not sure.

In any case, for now I guess the only way is spinning up an HTTP::Server...

Blacksmoke16:
:+1: I can maybe look into doing that in another PR.

straight-shoota:
@asterite Agreed. But we don't have a mockable HTTP::Client yet. It's one of the features suggested in #6011

Blacksmoke16:
@straight-shoota @asterite I added some specs for the getting an access token.  Not sure what else could be done to make it better, but i'm open to ideas.

spec/std/oauth2/client_spec.cr
```diff
@@ -1,6 +1,20 @@
 require "spec"
 require "oauth2"
 
+private def assert_request
```
straight-shoota: This is not an assertion method. Let's call it `oauth2_server`.

straight-shoota:
I'm sorry, I thought #7197 was already merged into master, but it isn't. Please take a look at the changes in this PR to avoid leaking fibers and sockets from the spec.

spec/std/oauth2/client_spec.cr
```diff
@@ -1,6 +1,20 @@
...
+  server = HTTP::Server.new do |context|
+    body = context.request.body.not_nil!.gets_to_end
```
asterite: Hm... this `not_nil!` here looks ugly. This has nothing to do with this PR, but I wonder how we ended up with `body` begin nilable. I mean, it makes sense, a request might not have a body. But maybe we should have `body?` (returns `nil` if none) and `body` (raises if none) as we do in other places. That way accessing the body is an assertion itself. Thoughts? (for another discussion)

straight-shoota: Alternatively, `body` could always be an IO, but could be an empty one. I don't think the HTTP specification differentiates between "no body" and "empty body".

asterite: Yes, that could work too. We'd need some `IO::Empty` or something...

Blacksmoke16:
@straight-shoota Done.  I created the `http/spec_helper` in my branch as well and used the setup there to reuse some code.  

Shouldn't conflict assuming no further changes are made to it.

spec/std/http/spec_helper.cr
```diff
@@ -0,0 +1,46 @@
+require "spec"
+
+private def wait_for(timeout = 5.seconds)
+  now = Time.monotonic
+
+  until yield
+    Fiber.yield
+
+    if (Time.monotonic - now) > timeout
+      raise "server failed to start within 5 seconds"
```
straight-shoota: I'm just seeing this now, for better re-usability this should report the actual `timeout` value.

Blacksmoke16: @straight-shoota 👍 I'll fix that up when i get home.

straight-shoota: I suppose this will be ready to be merged first and #7197 needs to be rebased anyway.

PR #7426, opened 2019-02-13
Spec: do Fiber.yield after each test

asterite:
When you are running a lot of specs and you press CTRL+C you have to wait until there's some IO that would result in fiber switching for the signal event to be trapped. Not all specs do IO so if there are a bunch of them that only do CPU stuff you won't be able to halt the specs from running.

This PR adds a `Fiber.yield` before each spec to give a chance for the signal event to be handled.

I didn't notice any slowdown when doing this.

I think this won't be needed once we have multiple threads... but for now we don't.

### DIFF
--- a/src/spec/methods.cr
+++ b/src/spec/methods.cr
@@ -57,7 +57,7 @@ module Spec::Methods
 
         # We do this to give a chance for signals (like CTRL+C) to be handled,
         # which currently are only handled when there's a fiber switch
-        # (IO stuff, sleep, etc.). Without it the use might way more than needed
+        # (IO stuff, sleep, etc.). Without it the user might wait more than needed
         # after pressing CTRL+C to quit the tests.
         Fiber.yield
       end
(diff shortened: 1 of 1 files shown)


straight-shoota:
A comment in the code would be great.

asterite:
@straight-shoota You are totally right. Done!

src/spec/methods.cr
```diff
@@ -54,6 +54,12 @@ module Spec::Methods
         Spec.abort! if Spec.fail_fast?
       ensure
         Spec.run_after_each_hooks
+
+        # We do this to give a chance for signals (like CTRL+C) to be handled,
+        # which currently are only handled when there's a fiber switch
+        # (IO stuff, sleep, etc.). Without it the use might way more than needed
```
bew:
```suggestion
        # (IO stuff, sleep, etc.). Without it the user might wait more than needed
```

asterite:
Sorry to interrupt, but can we continue discussing this in the forums? This is an issue tracker and this is not an issue (it's brainstorming).

asterite:
Good to go? 😄

PR #7436, opened 2019-02-15
Add missing `String#to_i` type annotations

j8r:


r00ster91:
Why are they missing? Also what's with the `underscore` argument?

asterite:
What do you mean by missing? Type annotations are optional. This is a breaking change.

j8r:
Yeah I miss the underscore, thanks @r00ster91

j8r:
By missing, I mean they are inferred to booleans - but their types are missing in the docs.

src/string.cr
```diff
@@ -568,7 +568,7 @@ class String
         next
       end
 
-      last_is_underscore = false
+      last_is_underscore : Bool = false
```
straight-shoota: This change is nonsense.

src/string.cr
```diff
@@ -3470,7 +3470,7 @@ class String
     return self if empty?
 
     first = true
-    last_is_underscore = false
+    last_is_underscore : Bool = false
```
straight-shoota: ditto.

src/string.cr
```diff
@@ -3480,7 +3480,7 @@ class String
           last_is_underscore = true
         elsif last_is_underscore
           str << char.upcase
-          last_is_underscore = false
+          last_is_underscore : Bool = false
```
straight-shoota: ditto.

asterite:
They are not inferred to booleans. With this change one cannot pass `nil` or any `truthy` value other than `true` anymore. This is why this is a breaking change.

It's not that I'm against this change, though... but we have to be careful about merging this.

straight-shoota:
I think it's better to only allow `Bool` instead of any truthy or falsey values.

src/string.cr
j8r: I have searched & replaced then push before going to week-end without reviewing... oops :)

j8r:
Anything blocking for merging?

PR #7442, opened 2019-02-16
Add String::Builder#clear

r00ster91:
One of my instance variables in my class is a `String::Builder` and I would really like to clear it efficiently instead of having to build a new `String::Builder` every time I want to build a new string with it.
So I added this which simply sets the bytesize of the builder to zero so the old bytes from the other string build are simply overwritten by the new string build.

```crystal
io = String::Builder.new

Benchmark.ips do |bm|
  bm.report("new") do
    io = String::Builder.new
  end

  bm.report("clear") do
    io.clear
  end
end
```
```
  new   6.86M (145.79ns) (±12.55%)  176 B/op  88.42× slower
clear  606.5M (  1.65ns) (± 6.06%)    0 B/op        fastest
```
This also adds a NOTE to `String.build` which suggests that when you use `String.build` in a loop, that you should use the `String::Builder` class and simply clear the old contents of the builder and then build a new string with the same builder. That's of course much faster than creating a new `String::Builder` everytime which is done by `String.build` in a loop. I did these improvements in `src/mime/media_type.cr`, for example.
```crystal
io = String::Builder.new

Benchmark.ips do |bm|
  bm.report("build") do
    10.times do
      String.build do |io|
        io << "abc"
      end
    end
  end

  bm.report("clear") do
    10.times do
      io << "abc"
      io.to_s
      io.clear
    end
  end
end
```
```
build 516.91k (  1.93µs) (±21.86%)  2083 B/op  19.69× slower
clear  10.18M ( 98.27ns) (±11.60%)     0 B/op        fastest
```

### DIFF
--- a/spec/std/string_builder_spec.cr
+++ b/spec/std/string_builder_spec.cr
@@ -10,7 +10,7 @@ describe String::Builder do
   end
 
   it "clears" do
-    builder = String::Builder.new
+    io = String::Builder.new
     io << 123
     io.to_s.should eq("123")
 
--- a/spec/std/string_builder_spec.cr
+++ b/spec/std/string_builder_spec.cr
@@ -9,6 +9,17 @@ describe String::Builder do
     str.should eq("123456")
   end
 
+  it "clears" do
+    builder = String::Builder.new
+    io << 123
+    io.to_s.should eq("123")
+
+    io.clear
+
+    io << "hello"
+    io.to_s.should eq("hello")
+  end
+
   it "raises if invokes to_s twice" do
     builder = String::Builder.new
     builder << 123
--- a/src/mime/media_type.cr
+++ b/src/mime/media_type.cr
@@ -277,6 +277,8 @@ module MIME
         end
       end
 
+      io = String::Builder.new
+
       # 3. Resolve continuation parameters
       #
       #     extended-parameter := (extended-initial-name "=" extended-value) /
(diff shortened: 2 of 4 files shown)


spec/std/string_builder_spec.cr
```diff
@@ -9,6 +9,17 @@ describe String::Builder do
     str.should eq("123456")
   end
 
+  it "clears" do
+    builder = String::Builder.new
```
Sija:
```suggestion
    io = String::Builder.new
```

r00ster91: Oh whoops, thanks.

ysbaddaden:
AFAIK this is **unsafe** and there is good reason for the builder to not be reusable: the allocated buffer is merely casted to a String. This means that this clear method is modifying the previously generated String (unexpected) and possibly reallocating the buffer (segfault).

Sija:
@ysbaddaden Good call! Shouldn't be possible to realloc buffer with `@buffer = GC.malloc_atomic(capacity.to_u32).as(UInt8*)` call within `#clear` to avoid case you've mentioned?

r00ster91:
Oh I see what you mean:
```crystal
class String::Builder < IO
  def clear
    @bytesize = 0
    @finished = false
  end
end

io = String::Builder.new

io << "hello"

string = io.to_s

io.clear

io << "123"

string # => "123lo"
```
But yes it seems to be possible to fix by reallocating the buffer:
```crystal
class String::Builder < IO
  def clear
    @bytesize = 0
    @finished = false
    @buffer = GC.malloc_atomic(@capacity.to_u32).as(UInt8*)
  end
end

io = String::Builder.new

io << "hello"

string = io.to_s

io.clear

io << "123"

string # => "hello"
```

asterite:
String::Builder shouldn't be public API nor used directly. Use IO::Memory. We should probably add mutating methods to it to make it easy to build and modify a string.

ysbaddaden:
Closing. String::Builder is internal detail to implement `String.build`. Messing with it is dangerous.

As stated by Asterite, please use an IO::Memory object.

Sija:
Anyhow, it seems there's a performance improvement to be done in `MIME::MediaType.parse_impl`.

/cc @straight-shoota

straight-shoota:
Yeah, although it's probably not a huge improvement, in practice continuation parameters are actually quite rare I think. However, there are probably other places in the stdlib as well where we could reuse a string buffer.

PR #7444, opened 2019-02-17
Unprotect HTTP::Params#raw_params getter

Sija:
My main motivation is efficiently modifying values inside without needless (re)parsing/(re)creating it manually.  atm if one needs to access the raw value, the only solutions I've found is to access it via `.@raw_params` syntax which is a serious no-no, or to recreate it manually:

```cr
query_string = "foo=bar&foo=baz"
query_hash = {} of String => Array(String)
HTTP::Params.parse(query_string).each do |key, value|
  query_hash[key] ||= [] of String
  query_hash[key] << value
end
# modify the *query_hash*, in order to cleanup values or for whatever other reason...
modified_query_string = HTTP::Params.new(query_hash).to_s
```

`HTTP::Params#initialize` takes `@raw_params` as an argument so it's already (sort of) part of a public API, so why not give access to `#raw_params` which would help in cases like above?


asterite:
Can the value returned by `HTTP::Params.parse(query_string)` be modified instead? I can't see any parsing done by the `[]=` setter in `HTTP::Params`.

Sija:
@asterite If I understand you correctly that would be difficult at times since sometimes you need to pass `Hash(String, Array(String))` to an external method for validation/sanitization purposes so in order to achieve it you'll be back again to manual creation of such `Hash`.

Sija:
🏓 any decisions to be made here?

asterite:
Maybe we can call this `to_h`? I wouldn't mind that.

Sija:
@asterite `HTTP::Params#to_h` already exists and returns `Hash` with flattened values.

asterite:
I guess this is fine then. Could you add some docs?

Sija:
@asterite done.

straight-shoota:
Could you please add a reference to `#to_h` (and vice versa) explaining the respective differences.

Sija:
@straight-shoota sure, I'll do it once I get back home.

Sija:
@straight-shoota force-pushed with added doc reference to `#to_h` method.

src/http/params.cr
```diff
@@ -132,7 +132,18 @@ module HTTP
       form_builder.to_s
     end
 
-    protected getter raw_params
+    # In contrast to `#to_h`, which returns flattened key-value pairs,
```
straight-shoota: Documentation of a method should not start with comparing it to another method. The first sentence should only be a very basic description.
E.g.
```cr
# Returns a `Hash` with raw parameters mapping from `String` to `Array(String)`.
#
# The return value can be supplied to `.new` to create a new instance.
#
# ```
# params = HTTP::Params.parse("foo=bar&foo=baz&qux=zoo")
# params.raw_params # => {"foo" => ["bar", "baz"], "qux" => ["zoo"]}
#
# params2 = HTTP::Params.new(params.raw_params)
# params2.to_s # => "foo=bar&foo=baz&qux=zoo"
# ```
#
# In contrast, `#to_h` returns flattened key-value pairs.
getter raw_params : Hash(String, Array(String))
```

straight-shoota:
Actually, `#to_h` doesn't work as expected. Values of duplicate keys are overridden instead of aggregated. That's not how HTTP params are supposed to work.

```cr
HTTP::Params.parse("foo=bar&foo=baz&qux=zoo").to_h # => {"foo" => "baz", "qux" => "zoo"}
```

I think it makes sense to rename `#raw_params` to `#to_h` then, because the generic `#to_h` implementation doesn't make any sense.

Sija:
> Actually, `#to_h` doesn't work as expected.

And how's it expected to work? I haven't seen any SPEC to which it should adhere, so IMO it's hard to talk about expectations here...

AFAIR parsing query params differs in every language/framework, so it's far from standardized.

> I think we should rename `#raw_params` to `#to_h` then, because the generic `#to_h` implementation doesn't make any sense.

I'm not sure about that, in most of the cases you might want to ignore duplicated values, so the current logic is what you'd like to have.

straight-shoota:
Okay, you're right, there is not specification about this and different approaches are applied by different frameworks. But the most common solution is to aggregate multiple values into an array. That's also the internal representation used by `HTTP::Params`, so it feels like a natural fit for a `#to_h` method and IMO makes more sense than the current implementation. It allows the user to decide the strategy for multiple values:
```cr
# all values
params.to_h
# only first value:
params.to_h.transform_values &.first
# only last value:
params.to_h.transform_values &.last
# concatenation of all values:
params.to_h.transform_values &.join(',')
```

And this is definitely unexpected behaviour:
```cr
params = HTTP::Params.parse("foo=bar&foo=baz")
params["foo"]      # => "bar"
params.to_h["foo"] # => "baz"
```

The hash should contain the same values as returned by either `HTTP::Params#fetch` ("first" strategy) or `HTTP::Params#fetch_all` ("all" strategy), but not an entirely different one ("last" strategy).

PR #7445, opened 2019-02-17
OpenSSL::SSL::Socket#cipher and #tls_version

carlhoerberg:


spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -0,0 +1,39 @@
+require "spec"
+require "socket"
+require "../../spec_helper"
+require "../../../support/ssl"
+
+describe OpenSSL::SSL::Socket do
+  it "knows which cipher that is in use" do
```
r00ster91:
```suggestion
  it "returns the cipher that is currently in use" do
```

spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -0,0 +1,39 @@
...
+        end
+      end
+
+      client = server.accept
+      client.cipher.should contain "RSA"
+      client.close
+    end
+  end
+
+  it "knows which TLS version that is in use" do
```
r00ster91:
```suggestion
  it "returns the TLS version" do
```

src/openssl/ssl/socket.cr
```diff
@@ -179,4 +179,16 @@ abstract class OpenSSL::SSL::Socket < IO
...
   end
+
+  # Returns the cipher currently in use
+  def cipher : String
+    String.new(LibSSL.ssl_cipher_get_name(LibSSL.ssl_get_current_cipher(@ssl)))
+  end
+
+  # Returns the TLS version currently in use
+  def tls_version : String?
+    if version = LibSSL.ssl_get_version(@ssl)
```
r00ster91: This function returns `"unknown"` when no version has been set, not `null` as stated [here](https://linux.die.net/man/3/ssl_get_version). Should this maybe be
```cr
if (version = LibSSL.ssl_get_version(@ssl)) == "unknown"`
```
?

carlhoerberg: Ok, maybe a test case for that too, how did you test it?

carlhoerberg: Ok, or just return "unknown" instead of being smart?

r00ster91: I think when it can return `"unknown"` then that should become a `nil` but I wonder if it's even possible that this method returns `"unknown"`? I don't think it's possible to call this without a connection being established. But I'm not sure.

straight-shoota: No, this should simply return the value returned by `ssl_get_version`. If that's `"unknown"`, that's fine. But the semantics shouldn't change to `nil`.

src/openssl/ssl/socket.cr
```diff
@@ -179,4 +179,14 @@ abstract class OpenSSL::SSL::Socket < IO
       String.new(host_name)
     end
   end
+
+  # Returns the cipher currently in use
```
straight-shoota:
```suggestion
  # Returns the current cipher used by this socket.
```

src/openssl/ssl/socket.cr
```diff
@@ -179,4 +179,14 @@ abstract class OpenSSL::SSL::Socket < IO
...
+  def cipher : String
+    String.new(LibSSL.ssl_cipher_get_name(LibSSL.ssl_get_current_cipher(@ssl)))
+  end
+
+  # Returns the TLS version currently in use
```
straight-shoota:
```suggestion
  # Returns the name of the TLS protocol version used by this socket.
```

PR #7446, opened 2019-02-17
Optimize String#to_u methods for the case of a negative number

r00ster91:
This makes `String`'s `to_u64_info` method return early when the number that should be converted to an unsigned integer is negative.
```crystal
Benchmark.ips do |bm|
  bm.report("old") do
    "-112978314788778".to_u64 { false }
  end

  bm.report("new") do
    "-112978314788778".new_to_u64 { false }
  end
end
```
```
old  58.13M (  17.2ns) (± 5.40%)  0 B/op  10.21× slower
new  593.7M (  1.68ns) (± 4.43%)  0 B/op        fastest
```

---

```crystal
Benchmark.ips do |bm|
  bm.report("old") do
    "-112978314788778913431478989314731437891343734789943413789474187178137914713173484137134780".to_u16 { false }
  end

  bm.report("new") do
    "-112978314788778913431478989314731437891343734789943413789474187178137914713173484137134780".new_to_u16 { false }
  end
end
```
```
old  43.73M ( 22.87ns) (± 8.58%)  0 B/op  13.42× slower
new 586.69M (   1.7ns) (± 3.40%)  0 B/op        fastest
```


src/string.cr
```diff
@@ -522,18 +522,19 @@ class String
     end
 
     negative = false
-    found_digit = false
-    mul_overflow = ~0_u64 / base
 
     # Check + and -
     case ptr.value.unsafe_chr
-    when '+'
-      ptr += 1
```
r00ster91: I'm pretty sure numbers like `+123` are less common than a negative number so I've moved this below the `'-'` check.

src/string.cr
```diff
@@ -522,18 +522,19 @@ class String
     end
 
     negative = false
-    found_digit = false
-    mul_overflow = ~0_u64 / base
```
r00ster91: I've moved these above the if-checks where the variables are used to avoid a

src/string.cr
```diff
@@ -522,18 +522,19 @@ class String
...
 
     # Check + and -
     case ptr.value.unsafe_chr
-    when '+'
-      ptr += 1
     when '-'
+      return ToU64Info.new 0, true, true if unsigned
```
straight-shoota: The other little optimizations are probably worth the change, but I'm questioning this one. Parsing negative numbers as unsigned integer should be a very, very rare case. UInts are usually only used in binary protocols anyway.

This line gives a nice performance boost for this rare case 10 times. But it also affects the other cases. It just costs a little performance, but parsing a negative number as a signed integer is a much more common scenario.

src/string.cr
```diff
@@ -491,7 +491,7 @@ class String
     invalid : Bool
 
   private macro gen_to_(method, max_positive = nil, max_negative = nil)
-    info = to_u64_info(base, whitespace, underscore, prefix, strict)
+    info = to_u64_info(base, whitespace, underscore, prefix, strict, unsigned: {{max_negative == nil}})
```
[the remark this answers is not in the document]
r00ster91: I did that because at all the other arguments you can also see what the name of this argument is (because the variables are named after the argument names) and so I wanted to adapt this one to the others by also clearly saying what the name of this argument is. It's also faster to understand so you don't have to scroll down to the `to_u64_info` def first.

src/string.cr
r00ster91: > Parsing negative numbers as unsigned integer should be a very, very rare case. UInts are usually only used in binary protocols anyway.

I actually have a case in my code. I'm converting a string to an integer and the integer must not be negative so I simply do: `string.to_u32?` which will obviously also return `nil` when the integer in the string is negative. I don't think cases where you don't want to allow negative numbers are that rare.

Most of the time in the benchmarks the new `to_i` methods were `1.04× slower` or less slower for negative numbers but it's actually quite easy to get the runtime `if unsigned` check over to compile-time by simply making `to_u64_info` a macro and putting the `if unsigned` check in a macro branch.
So now there are no longer trade-offs.

straight-shoota: > I don't think cases where you don't want to allow negative numbers are that rare.

Yeah, but such cases would usually parse a signed integer and check for `>= 0`. Using an unsigned integer type for this is a creative solution but should not really be recommended.

> So now there are no longer trade-offs.

Maybe, but I don't see a relevant benefit either.

r00ster91: > Maybe, but I don't see a relevant benefit either.

But you said this previously:

> And it also affects the other cases. It just costs a little performance, but parsing a negative number as a signed integer is a much more common scenario. I wouldn't want to sacrifice this for a doubtful improvement.

Didn't you say by this that the `if unsigned` runtime check could cost perfomance for parsing negative signed integers? That's what I resolved with the macro. The `if unsigned` check no longer costs any perfomance for parsing negative numbers as signed integers. It no longer affects any other cases.

straight-shoota: That was *a* point. Others are general complexity, increased compile time, resources needed for implementing, reviewing and applying this patch, successive maintenance etc. All for an IMHO very marginal benefit.

straight-shoota: I don't oppose this PR if others think it's worth merging. But I won't approve it either.

src/string.cr
```diff
@@ -509,7 +509,7 @@ class String
     end
   end
 
-  private def to_u64_info(base, whitespace, underscore, prefix, strict)
+  private macro to_u64_info(base, whitespace, underscore, prefix, strict, unsigned)
```
wontruefree: why does this need to be a macro instead of a method?

Sija: See https://github.com/crystal-lang/crystal/pull/7446/files#diff-87ebb896824e7dc3bd8d5b29293bcf00R529

wontruefree: Does it have to be a maco. it seems like it could just check for the parameter.

Sija: FWIW it's about trading runtime for compile-time check.

asterite: This is a no-no, it duplicates the binary size and compile time for all methods.

ysbaddaden: And probably optimized just the same by LLVM anyway.

r00ster91: > And probably optimized just the same by LLVM anyway

The new `to_i` methods were always something like `1.04× slower` on negative numbers when I ran the benchmark about 6 times in a row. But it could be a coincidence.

> This is a no-no, it duplicates the binary size and compile time for all methods.

But how does it duplicate the binary size? Isn't it even less binary size with a macro? A macro is always evaluated/inlined by Crystal but for a method the chance that it gets inlined isn't always 100% as far as I know.
But I understand the compile-time part and will revert this. The compile times take too long already.

straight-shoota: Expanded macros and inlined methods multiply the amount of code because it's repeated for each time they're used. On the contrary, the code of non-inlined methods is shared between all it's callers.

RX14:
It improves performance in an edge case only, but that doesn't mean it's not worth merging, as it's only a small bit of code.

Issue #5988, opened 2018-04-22
formatter bug


kostya:
```crystal
h = {
  xxx: Int32,
  y:   {a: 1,
      b: 2},
}
```

b line is wrong indentation

Sija:
IMO it should be formatted like:

```cr
h = {
  xxx: Int32,
  y:   {a: 1, b: 2},
}
```

or:

```cr
h = {
  xxx: Int32,
  y: {
    a: 1,
    b: 2,
  },
}
```

But that changes the format quite a bit...

@kostya How's your ideal final formatting of this?

PR #7450, opened 2019-02-17
Format: fix indent of nested array elements

### LINKED ISSUE #7443: Formatter regression with hashes inside array
Code below:

```cr
very_long_variable_name = [{
  :foo => :bar,
}]
```

gets formatted to:

```cr
very_long_variable_name = [{
                             :foo => :bar,
                           }]
```

which seems like a regression after merging #7048.

IMHO it should keep the format from the original snippet, i.e. 2 space indentation.

/cc @MakeNowJust

MakeNowJust:
Fixed #7443

With more precision, this change treats multiline elements.
This change detects multiline elenent and keeps current indent in this case, or elements will be aligned at start column.

Now, the formatter keeps these formats:

```crystal
# #7048
[
  {
    "some" => "hash",
    "with" => "values",
  }, {
    "another"  => "hash",
    "will get" => "deindented",
  }, {
    "now" => "totally",
  }, {
    "flat" => "against the left",
  },
]

# #7443
very_long_variable_name = [{
  :foo => :bar,
}]

very_long_variable_name = [
  {
    :foo => :bar,
  }, {
    :bar => :baz,
  },
]

very_long_variable_name = [
  {
    :foo => :bar,
  },
  {
    :bar => :baz,
  },
]

# single-line elements case

foo = [1, 2, 3,
       4, 5, 6]

foo = [1, 2, 3, # foo
       4, 5, 6]

foo = [
  1, 2, 3,
  4, 5, 6,
]
```

@asterite Sorry. I know you are working for #7443 just now. However I wrote this patch yesterday (but I wrote a spec today.)

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -1310,4 +1310,11 @@ describe Crystal::Formatter do
     "  2  \n" +
     "end"
   )
+
+  # #7443
+  assert_format "long_variable_name = [{\n  :foo => 1,\n}, {\n  :bar => 2,\n}]"
+  assert_format "long_variable_name = [\n  {\n    :foo => 1,\n  }, {\n    :bar => 2,\n  },\n]"
+  assert_format "long_variable_name = [\n  {\n    :foo => 1,\n  },\n  {\n    :bar => 2,\n  },\n]"
+  assert_format "long_variable_name = [1, 2, 3,\n                      4, 5, 6]"
+  assert_format "long_variable_name = [1, 2, 3, # foo\n                      4, 5, 6]"
 end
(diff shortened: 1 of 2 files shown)


asterite:
@MakeNowJust Thank you! I didn't start working on this yet, it was just my intention if it was too hard to implement. But you did it so it's great!

Sija:
You guys are awesome, thanks! 🙇

kostya:
will this fix https://github.com/crystal-lang/crystal/issues/5988?

asterite:
@kostya I don't think so. However, an idea to fix that is to use the start line and end line of the hash/namedtuple to know what to align. After this is merged I can fix it (or @MakeNowJust might ^_^)

Issue #7451, opened 2019-02-18
Unify Time::Span and Time::MonthSpan


anykeyh:
This Issue is related to:
https://forum.crystal-lang.org/t/time-span-and-time-monthspan/430

# Current implementation

Crystal offers method à-la active-support for dealing with advance an rewind in time, for example:

```crystal
tomorrow = Time.now + 1.day
last_month = 1.month.ago
```

Theses methods rely on `Time::Span` and `Time::MonthSpan`. The current implementation is not unified, therefore:

```crystal
time_span = 1.month + 1.day # Throw compile time error because we can't add Time::MonthSpan with Time::Span 
```

# The problem

It has been done like this because months and days are treated differently, therefore, add a month THEN a day is not the same than add a day THEN a month:

```crystal
puts Time.new(2018,1,28) + 1.month + 1.day
# > 2018-03-01 00:00:00 +07:00
puts Time.new(2018,1,28) + 1.day + 1.month
# > 2018-02-28 00:00:00 +07:00
```

# Unification

In Active Support, the type is unified by stacking the parts of the transformations in an Array like this:

```ruby
(1.day + 1.month + 1.day).parts
# => [[:days, 1], [:months, 1], [:days, 1]]
```

The goal here would be to rewrite the Time::Span and Time::MonthSpan to handle it like Active Support, and allow to store and combine directly months and days.

After the rewriting, it would be possible to do:

```crystal
struct MyStruct
  @duration : Time::Span

  def initialize(@duration)
  end
end

MyStruct.new(1.month + 1.day) #< Currently impossible.
```

## Caveats

Current Time::Span offers methods which doesn't exists in Time::MonthSpan.

### Like PG's Intervals

I let you discuss about the best way to do it. In PostgreSQL, the type interval apply the month followed by the days as example below:

```sql
SELECT DATE('2018-01-28') + INTERVAL '1 day 1 month';
-- 2018-03-01 00:00:00

SELECT DATE('2018-01-28') + INTERVAL '1 month 1 day';
-- 2018-03-01 00:00:00
```

Pro:
Performance-wise, it stores only the month, day and microseconds in a fixed-length structure (16 bytes)

Cons:
- Some methods of Time::Span are non-sensical in the new implementation and should be removed (?) (total_days etc...)
- This would cause a problem when applying transformation from left to right:

```crystal
Time.new(2018,1,28) + (1.day + 1.month) != Time.new(2018,1,28) + 1.day + 1.month
```

[...]

PR #7454, opened 2019-02-19
Combine Time::Span and Time::MonthSpan

jgaskins:
I was trying to map [Neo4j's `Duration` type](https://neo4j.com/docs/cypher-manual/current/syntax/temporal/#cypher-temporal-durations) to Crystal's `Time::Span` but kept running into issues due to `Time::Span` and `Time::MonthSpan` being separate types. This PR combines them into `Time::Span`.

If I'm being completely honest, I can't imagine a realistic scenario where someone would need to specify a duration of `1.year + 1.day` for anything but right now that's impossible with Crystal in a scenario where anyone _could_ need it. In my case, if someone does actually store that duration in their Neo4j database, we have no way to represent it.

I was going to write a wrapper around both types for the Neo4j driver but I figured it might be worth combining them in the core library. I could imagine people running into the same issues elsewhere, such as with [Postgres's `INTERVAL` type](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT) or any time you need to deal with user-supplied time spans.

_Note:_ There were previously no specs for `Time::MonthSpan`. This PR adds a spec regarding months/years for `Time::Span` and a few for `Time#{+,-}(Time::Span)`.

### DIFF
--- a/spec/std/time/span_spec.cr
+++ b/spec/std/time/span_spec.cr
@@ -118,17 +118,17 @@ describe Time::Span do
   end
 
   it "test add" do
-    t1 = Time::Span.new 2, 3, 4, 5, 6_000_000
-    t2 = Time::Span.new 1, 2, 3, 4, 5_000_000
+    t1 = 1.year  + 2.months + 3.days + 4.hours + 5.minutes + 6.seconds + 7_000_000.nanoseconds
+    t2 = 2.years + 3.months + 4.days + 5.hours + 6.minutes + 7.seconds + 8_000_000.nanoseconds
     t3 = t1 + t2
 
-    t3.days.should eq(3)
-    t3.hours.should eq(5)
-    t3.minutes.should eq(7)
-    t3.seconds.should eq(9)
-    t3.milliseconds.should eq(11)
-    t3.nanoseconds.should eq(11_000_000)
-    t3.to_s.should eq("3.05:07:09.011000000")
+    t3.years.should eq 3
+    t3.months.should eq 5
+    t3.days.should eq 7
+    t3.hours.should eq 9
+    t3.minutes.should eq 11
+    t3.seconds.should eq 13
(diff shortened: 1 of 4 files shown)


jgaskins:
Oh, apparently someone just opened #7451 regarding this exact problem. :joy:

anykeyh:
It's fortunate you made it, what a coincidence indeed ! :+1: 

Note: In your implementation, this will produce weird behavior in certain cases:

```crystal
now =Time.now 

now + 1.day + 1.month == now + (1.day + 1.month) # Return false
```

I don't know what to think about it honestly. In ActiveSupport, operations are triggered in the order of the definition (the Time::Span object store an array of operation), while in Postgres the implementation is sensibly the same than yours.

straight-shoota:
Thanks for pushing this proposal forward! As mentioned in #7451 there is already a lengthy discussion about this topic in #6522. The todo list which seems to be mostly agreed on is in https://github.com/crystal-lang/crystal/issues/6522#issuecomment-413344502

I don't think we should merge `Time::Span` and `Time::MonthSpan` into a single type because they represent different concepts and the data structure would become much larger for simple seconds-based durations (which is the most common use case). Just combining months and seconds is not enough anyway, we'll at least need a field for days as well.

asterite:
I don't think making `Time::Span` bigger is really an issue. Let's first find some code where doing time operations is the bottleneck of a program.

anapsix:
imho, it would be nice to have a unified `Time::Span`, but what would be the value of `1.month.total_days`?

jgaskins:
The GitHub issues related to this have been closed. I extracted a shard last year called [`duration`](https://github.com/jgaskins/duration) that handles seconds-based and calendar-based units in the same type, so I'm gonna go ahead and close this, as well.

Issue #7456, opened 2019-02-19
OpenSSL exception when reading/writing at line-rate


carlhoerberg:
When reading from a `OpenSSL::SSL::Socket::Server` that has negotiated a GCM cipher (eg. TLS_AES_256_GCM_SHA384 or ECDHE-RSA-AES128-GCM-SHA256) at full speed I very quickly get

```
OpenSSL::SSL::Error:SSL_read: error:1408F119:SSL routines:ssl3_get_record:decryption failed or bad record mac
```

However, if i phase the write side I don't get it. 

Will add an example but opening the issue as an placeholder.

carlhoerberg:
It does not seem to be related to GCM ciphers specifically, can reproduce with a number of different ciphers, instead it seems to be some alignment problem. 

```crystal
require "socket"
require "openssl"

def server
  socket = TCPServer.new(5555)
  context = OpenSSL::SSL::Context::Server.new
  context.private_key = "localhost.key"
  context.certificate_chain = "localhost.crt"

  loop do
    client = socket.accept? || break
    ssl_socket = OpenSSL::SSL::Socket::Server.new(client, context, sync_close: true)
    spawn echo(ssl_socket)
  end
end

def echo(ssl_socket)
  puts "Client connected"
  i = 0
  buf = uninitialized UInt8[4096]
  loop do
    i += 1
    ssl_socket.read(buf.to_slice)
    ssl_socket.write(buf.to_slice)
  end
rescue ex
  puts "Failed at the #{i} read"
  raise ex
end

def client
  socket = TCPSocket.new("127.0.0.1", 5555)
  context = OpenSSL::SSL::Context::Client.new
  context.verify_mode = OpenSSL::SSL::VerifyMode::NONE
  ssl_socket = OpenSSL::SSL::Socket::Client.new(socket, context)
  spawn do
    buf = StaticArray(UInt8, 1000).new(0)
    loop do
      ssl_socket.write buf.to_slice
    end
  end
  buf = uninitialized UInt8[4096]
  loop do
    ssl_socket.read buf.to_slice
  end
end

unless File.exists? "localhost.key"
  system "openssl req -x509 -out localhost.crt -keyout localhost.key -newkey rsa:2048 -nodes -sha256 -subj '/CN=localhost'"
end

spawn client
server
```

Output: 

```
Client connected
Failed at the 137 read
[...]

PR #7460, opened 2019-02-20
Disable double write buffering in OpenSSL sockets

### LINKED ISSUE #7458: Double buffering in OpenSSL::SSL::Socket
This seems like a contradiction: 

https://github.com/crystal-lang/crystal/blob/c9d1eef8fde5c7a03a029d64c8483ed7b4f2fe86/src/openssl/ssl/socket.cr#L87-L92

I suspect it should be `io.sync = true`?

carlhoerberg:
This fixes #7458 and #7456

### DIFF
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -87,7 +87,7 @@ abstract class OpenSSL::SSL::Socket < IO
     # Since OpenSSL::SSL::Socket is buffered it makes no
     # sense to wrap a IO::Buffered with buffering activated.
     if io.is_a?(IO::Buffered)
-      io.sync = false
+      io.sync = true
       io.read_buffering = false
     end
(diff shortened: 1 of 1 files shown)


asterite:
I approve this because:
- it apparently fixes two issues
- I'm almost sure the code was supposed to match the comment, but it wasn't the case

@carlhoerberg Thank you!

asterite:
I wonder if tests can be included for this. That way this stays working forever.

asterite:
At least by including some manual specs (which are run before releasing, I think). This is in `spec/manual`

carlhoerberg:
I don't know if it's reliable for all kind of things related to OpenSSL buffering, but for this case this code triggered it every time: https://github.com/crystal-lang/crystal/issues/7456#issuecomment-465319901

PR #7466, opened 2019-02-21
Experiment: add MAIN_FILE constant

asterite:
This PR lets the compiler include a predefined constant, `MAIN_FILE` with the full filename of the first file that's being compiled.

With this you can do:

```crystal
if __FILE__ == MAIN_FILE
  puts "Main program..."
end
```

This is similar to Python's `if __name__ == __main__` and Ruby's not that common `if __FILE__ == $0`.

I just want to gather opinions on this because it's asked from time to time, like today: https://forum.crystal-lang.org/t/is-there-any-equivalent-to-cs-main-or-pythons-if-main-statement/450

(note: Ideally this would be named `__MAIN__` but that parses as a variable, not a constant, so we would need a small hack to make that parse as a constant, that's why I went with `MAIN_FILE`)

### DIFF
--- a/src/compiler/crystal/program.cr
+++ b/src/compiler/crystal/program.cr
@@ -108,7 +108,7 @@ module Crystal
     property? show_error_trace = false
 
     # The main filename of this program
-    property filename : String?
+    getter filename : String?
 
     # A `ProgressTracker` object which tracks compilation progress.
     property progress_tracker = ProgressTracker.new
@@ -228,6 +228,14 @@ module Crystal
       types["ThreadLocal"] = @thread_local_annotation = AnnotationType.new self, self, "ThreadLocal"
 
       define_crystal_constants
+    end
+
+    def filename=(filename : String?)
+      @filename = filename
+
+      if filename
+        types["MAIN_FILE"] = Const.new self, self, "MAIN_FILE", StringLiteral.new(filename)
+      end
     end
 
     # Returns a `LiteralExpander` useful to expand literal like arrays and hashes


asterite:
Also note that if we go with this, it's much better to do:

```crystal
{% if __FILE__ == MAIN %}
  puts "Main program"
{% end %}
```

because otherwise the main program will be always compiled-in, even if you don't actually use it.

jhass:
I still reject this usecase, having two files, one for the executable and one for the library is cleaner in my opinion. Introducing this will fragment the community on which approach to use (when).

straight-shoota:
I second @jhass here. I don't think there are any other use cases for the path of the main file, it's only for comparing with the current file. If we want to have a feature like this, it would probably be better to just make it a macro method like `main_file?` which returns a `BoolLiteral`.

I'm not decided whether this would be a good idea. I generally like it, because it avoids having to put tiny bootstrap code like `require "myapp"; MyApp.run` in a separate file.

stronny:
I don't think python's approach was any good in the first place. Conceptually the _program_ deals with global system state, `ENV` and `ARGV`, while _libraries_ provide an abstract API with internal types. Making the file produce different code depending on some implicit parameter is confusing, unless this is already an established idiom in the community. Also, what's wrong with small bootstrap files anyway?

dscottboggs:
I like the idea of having it be a macro method, what if it was a compiler hook?

I'm not sure if I see the problem with having both methods, it doesn't seem confusing to me that either would work. What I mean is, if I'm reading the source for something and I see `macro when_run_as_main` or `if MAIN_FILE`, it's pretty clear what is happening. By the same token, if I see a separate executable, which requires a library and then does something with it based on ENV and ARGV, it's also pretty clear what that does.

My reasoning for this is that the "main" file, `src/shard_name.cr` is also the location of the root module, so if you have your CLI in a separate file, and you just `crystal build src/shard_name.cr` the resulting executable `shard_name` wouldn't be capable of handling ARGV and ENV.

asterite:
Just a note that I'm also against this, I just wanted to see what others think.

Then there's [executables in shards](https://github.com/crystal-lang/shards/blob/master/SPEC.md#executables)

> The executables can be of any type or language (e.g., shell, binary, ruby), must exist in the bin folder of the Shard, and have the executable bit set (on POSIX platforms)

I can't see how can that rule interact with the `MAIN_FILE` check.

I mean, if we go this route then you as a shard developer will be able to execute a binary with that check, but nobody else will because it'll be buried inside the `lib` directory of that project.

I also personally think it's more organized to have lib code in one place, executable code that uses that lib code in another place. It's also more flexible because you can provide multiple binaries for your shard/project. Well, with the `MAIN_FILE` check you could do it too but I'm not sure executables map one to one with library modules/code.

dscottboggs:
So, relative to the project root, you should compile binaries and place them in `/bin`? Is there a recommended place to put the files that will be compiled into `/bin` under `/src`?

straight-shoota:
@asterite I think the part about shards executables doesn't really fit here.

@dscottboggs Only if you want them to be available as shards executables.

There are no recommendation currently for naming app and library entry files. But I think it would be great to discuss some kind of convention for this.

asterite:
Yeah, sorry, I misunderstood the shards spec. I thought it compiled the binaries found in the shards `bin` directory and put them into the project's `bin` directory. I think one is supposed to do that using a postscript...

In any case, I'll close this issue. This is just an experiment to see people's opinions. I wouldn't like to have such thing in the language. Sorry for the confusion.

PR #7467, opened 2019-02-21
OAuth2: don't parse error as JSON

### LINKED ISSUE #7464: oauth2 errors are making some incorrect json assumptions for Facebook
Hello thanks for all your hard work on crystal!

I use the multi_auth shard which itself use oauth2 from the crystal stdlib.

the OAuth2::Error class is trying to decode the error from the payload expecting something like this:

````json
{ "error": "some oauth error" }
````

This is what I understand according to the code in src/oauth2/error.cr:

```crystal
 pull.read_object do |key|
      case key
      when "error"             then error = pull.read_string
      when "error_description" then error_description = pull.read_string
      else
        raise "Unknown key in oauth2 error json: #{key}"
      end
    end
```

I suppose that must be the case with some oauth2 providers but this assumption is not valid for Facebook, which returns in case of error something like this:

```json
{"error":{"message":"This authorization code has expired.","type":"OAuthException","code":100,"fbtrace_id":"GyzZxJWzHMP"}}
```

Because of error being an object instead of the string it makes this json reader crash with this error:

```
Unhandled exception: Expected string but was begin_object at 1:11 (JSON::ParseException)
  from /usr/share/crystal/src/json/pull_parser.cr:518:5 in 'parse_exception'
  from /usr/share/crystal/src/json/pull_parser.cr:510:5 in 'expect_kind'
  from /usr/share/crystal/src/json/pull_parser.cr:180:5 in 'read_string'
  from /usr/share/crystal/src/json/pull_parser.cr:83:5 in 'read_object_key'
  from /usr/share/crystal/src/oauth2/error.cr:89:13 in 'new'
  from /usr/share/crystal/src/json/from_json.cr:13:3 in 'from_json'
  from /usr/share/crystal/src/oauth2/client.cr:154:7 in 'get_access_token_using_authorization_code'
[... more stack ]
```

I would have gladly fixed it myself but I'm a beginner crystal developer and I'm clearly not good enough yet to fix this issue.

asterite:
Fixes #7464

Let's not parse the error as the JSON documented in the [RFC](https://tools.ietf.org/html/rfc6749#section-4.1.2.1). Instead, just treat the whole response body as the exception message. One can parse it from there if needed, but I think it's usually not needed.

### DIFF
--- a/src/oauth2/client.cr
+++ b/src/oauth2/client.cr
@@ -149,7 +149,7 @@ class OAuth2::Client
     when 200, 201
       OAuth2::AccessToken.from_json(response.body)
     else
-      raise OAuth2::Error.from_json(response.body)
+      raise OAuth2::Error.new(response.body)
     end
   end
 
--- a/src/oauth2/error.cr
+++ b/src/oauth2/error.cr
@@ -1,28 +1,2 @@
 class OAuth2::Error < Exception
-  getter error : String
-  getter error_description : String?
-
-  def initialize(@error, @error_description)
-    if error_description = @error_description
-      super("#{@error}: #{error_description}")
-    else
-      super(@error)
-    end
-  end
-
-  def self.new(pull : JSON::PullParser)
-    error = nil
-    error_description = nil
-
-    pull.read_object do |key|
-      case key
-      when "error"             then error = pull.read_string
-      when "error_description" then error_description = pull.read_string
-      else
-        raise "Unknown key in oauth2 error json: #{key}"
-      end
-    end
-
-    new error.not_nil!, error_description
-  end
 end


Sija:
I'm not sure it's need to go that far, why not cover the use-case from the linked issue and throw the catcheable error containing the response body the user can parse as a fallback solution?

asterite:
If there's an OAuth error there's not much you can programmatically do: log the error, check it later, fix the bug, if any, or wait until the server fixes the problem, etc. You won't programmatically fix the error by checking the response body. That's why I think that it's better we don't do anything with the body. Just include it as an exception message.

If you really want to parse it, just do `JSON.parse(...)` or map it to a model of your own.

straight-shoota:
We could consider keeping the `#error` and `#error_description` methods but let them lazily parse the JSON on demand.

asterite:
@straight-shoota We could, but this way users can parse it if they want it so it's a bit faster too (no need to parse it if you don't want to use it)

asterite:
I'm open to suggestions. Let's think about the API and implementation for this (though I really think that parsing this is unnecessary in most of the cases and if you really want to parse it you can use `JSON.parse(ex.message)`.

straight-shoota:
I'm not very familiar with the OAuth2 API.

Actually, it's probably okay to merge this and think about alternatives when such a feature is requested (if ever).

PR #7469, opened 2019-02-21
Handle signals in a separate fiber

### LINKED ISSUE #7336: signal traps and unhandled SIGCHLD
```
#!/usr/bin/env crystal
child = Process.new "sleep", ["60"]
spawn { p child.wait }
Signal::INT.trap { exit 0 }
at_exit {
  child.kill
p :before
  sleep 1
p :after
}

sleep 10
raise "boom"
```

If left to run (as expected):
```
$ ./test.rb
:before
#<Process::Status:0x7fd86638cfd0 @exit_status=15>
:after
Unhandled exception: boom (Exception)
```

With `^C` pressed:
```
$ ./test.rb
^C:before
:after
```

As far as I was able to gather this happens because trap handlers are run signal-free, however it's not obvious that a call to `exit` will pull at_exit to its context.

If this is not a bug, please tell me a way to make sure the subprocess is reliably killed and reaped before the master dies. I've spent half a day to figure this all out and I'm all out of ideas short of taking back SIGCHLD from the event loop if it's even possible and safe.

asterite:
Fixes #7336

I had to assign `handler` to a separate variable because when it's closured all types assigned to it are in the closure (this includes `Nil` because it's the type that it gets after the `if`). Another alternative is to use `not_nil!` but that's maybe uglier and more cryptic.

The main problem this fixes is that when a signal is handled the runtime currently executes it in the same fiber in which the signal trapping/checking happens, and this prevents multiple signals to run concurrently. This in particular affects SIGCHLD, which is called when a child process ends.

### DIFF
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -208,13 +208,16 @@ module Crystal::Signal
 
   private def self.process(signal) : Nil
     if handler = @@handlers[signal]?
-      handler.call(signal)
+      non_nil_handler = handler # if handler is closured it will also have the Nil type
+      spawn do
+        non_nil_handler.call(signal)
+      rescue ex
+        ex.inspect_with_backtrace(STDERR)
+        fatal("uncaught exception while processing handler for #{signal}")
+      end
     else
       fatal("missing handler for #{signal}")
     end
-  rescue ex
-    ex.inspect_with_backtrace(STDERR)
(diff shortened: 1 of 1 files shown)


asterite:
@ysbaddaden Good catch! Fixed.

PR #7470, opened 2019-02-21
IO: remove `flush_on_newline` and only `sync` on TTY devices

### LINKED ISSUE #7431: Should STDOUT always set flush_on_newline by default?
I am just starting with Crystal, being a Ruby programmer for many years. I am impressed with its performance, but playing around, I found one thing that badly affects performance when writing to standard output: that `flush_on_newline` is always set by default on `STDOUT`. My trivial test:

    s = "a" * 60

    10000000.times do
      puts s
    end

takes `real: 10.4s user: 2.4s sys: 8.0s` when its output is redirected to a file. (FYI the same thing run by the Ruby 2.5.1p57 interpreter takes `real: 3.8s user: 3.5s sys: 0.3s`)

Yet if I add the line:

`STDOUT.flush_on_newline = false`

at the top, the time drops dramatically to: `real: 0.5s user: 0.3s sys: 0.2s`, which is more what I was expecting.

I suggest that the definition of STDOUT in kernel.cr be changed to:

`STDOUT = IO::FileDescriptor.new(1).tap { |f| f.flush_on_newline = true if f.tty? }`

so that flush_on_newline is only set when STDOUT is a TTY. I am fairly sure that this is how Unix & Linux C programs normally behave. Typically I believe that standard error is completely unbuffered, but appears to be line buffered in Crystal. This, however, probably isn't a big deal.

asterite:
Fixes #7431

Explanation for this [here](https://github.com/crystal-lang/crystal/issues/7431#issuecomment-463836358)

Merging this or not is still debatable, but the pros I see are:
- remove the strange `flush_on_newline` property that's actually a hack to be able to see TTY output as it's written, when `sync` does that in a more unified way (and works too if you call `io.print`). This also simplifies code because there's less code :-)
- improve performance when you redirect the output to a file

I checked and `Logger` already flushes the IO when a message is written to it.

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -5,16 +5,12 @@
   # The standard output file descriptor.
   #
   # Typically used to output data and information.
-  #
-  # NOTE: Gets flushed when a newline is written.
-  STDOUT = IO::FileDescriptor.new(1).tap { |f| f.flush_on_newline = true }
+  STDOUT = IO::FileDescriptor.new(1)
 
   # The standard error file descriptor.
   #
   # Typically used to output error messages and diagnostics.
-  #
-  # NOTE: Gets flushed when a newline is written.
-  STDERR = IO::FileDescriptor.new(2).tap { |f| f.flush_on_newline = true }
+  STDERR = IO::FileDescriptor.new(2)
 {% else %}
   require "c/unistd"
 
@@ -25,15 +21,17 @@
   #
   # Typically used to output data and information.
   #
-  # NOTE: Gets flushed when a newline is written.
-  STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  # When this is a TTY device, `sync` will be true for it
+  # at the start of the program.
+  STDOUT = IO::FileDescriptor.from_stdio(1)
 
   # The standard error file descriptor.
   #
   # Typically used to output error messages and diagnostics.
   #
-  # NOTE: Gets flushed when a newline is written.
-  STDERR = IO::FileDescriptor.from_stdio(2).tap { |f| f.flush_on_newline = true }
+  # When this is a TTY device, `sync` will be true for it
+  # at the start of the program.
+  STDERR = IO::FileDescriptor.from_stdio(2)
 {% end %}
 
 # The name, the program was called with.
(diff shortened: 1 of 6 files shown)


ysbaddaden:
Maybe I'd use a String.build in logger's default formatter, to avoid too many sync writes, but that can wait / be benchmarked to see if there is any real benefit.

src/kernel.cr
```diff
@@ -25,15 +21,17 @@
   #
   # Typically used to output data and information.
   #
-  # NOTE: Gets flushed when a newline is written.
-  STDOUT = IO::FileDescriptor.from_stdio(1).tap { |f| f.flush_on_newline = true }
+  # If it is a TTY device, `sync` will be true for this it
```
r00ster91:
```suggestion
  # When this is a TTY device, no write buffering will be done.
```
The line below this one can be removed. Ditto below.

asterite: I changed the text a bit but:
- changing `sync` to `no write buffering` is wrong because what's really happening is that `sync` becomes `true` and you can always change that (if we say "no write buffering" it's not clear how to revert it)
- this check happens at the start of the program so you can revert it, and I think it's important to say this

r00ster91: I agree but shouldn't `sync` be maybe `sync?`? `sync` does not exist.

asterite: there's `sync=` and `sync?`. In any case this doesn't matter much, eventually we should rename `sync` to `write_buffering`.

RX14:
This seems to me like a bad idea, since pipes are not TTYs, this will break crystal programs which are run as subprocesses and cause hangs. I suggest making STDIO sync always. If people need more performance they can always turn it off.

RX14:
I've run across this again, I've written a program which implements the [i3bar protocol](https://i3wm.org/docs/i3bar-protocol.html) which streams JSON over `STDOUT`. It's certainly confusing that the behavior of the program is different when you run it in the terminal and when it's run by another program (and this confused me for a couple of days *despite* having prior knowledge of this issue).

Does anyone have any comments on my above suggestion or I'll create a PR to make STDIO always sync.

asterite:
I don't understand why this is a problem in Crystal but not in Ruby.

RX14:
@asterite I've just looked at ruby and python, they take the approach from this PR: behaving differently based on whether the output is a TTY or not. So it's clearly debatable, I just think that choosing consistency over performance is better. People can always turn off `STDOUT.sync = false` if they hit a performance bottleneck in their code

PR #7474, opened 2019-02-22
Attempted String::Formatter has no type fix

### LINKED ISSUE #7461: BUG: `def consume_formatted_substitution ... at /usr/share/crystal/src/string/formatter.cr:53:11 has no type (Exception)
error:

```
BUG: `def consume_formatted_substitution
  key = consume_substitution_key('>')
  next_char
  arg = current_arg
  if if false
    arg.is_a?(Hash)
  else
    false
  end
    target_arg = arg[key]
  else
    raise(ArgumentError.new("One hash or named tuple required"))
  end
end` at /usr/share/crystal/src/string/formatter.cr:53:11 has no type (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
```

I do not have code that can reproduce it atm.  Came across it while i was writing a tutorial for Athena + Granite.  I can share the repo once its done if anyone wants to poke around.

The issue seems to be with something Athena is doing.  I'll poke around later and see if i can figure anything out.

Blacksmoke16:
Fix #7461

I'm not super familiar with what overall is going on but this seems to fix it.  If I had to guess it would be because the type from the arg doesn't exist (for some reason, maybe the cause of that is at a lower level/someplace else?) and this causes `target_arg` to be typeless.  This change makes `target_arg` nil if a type does not exist, which is an allowable type for it.

### DIFF
--- a/src/string/formatter.cr
+++ b/src/string/formatter.cr
@@ -55,7 +55,7 @@ struct String::Formatter(A)
     next_char
     arg = current_arg
     if arg.is_a?(Hash) || arg.is_a?(NamedTuple)
-      target_arg = arg[key]
+      target_arg = arg[key] unless arg.empty?
     else
       raise ArgumentError.new "One hash or named tuple required"
     end


asterite:
Sorry, but this is a compiler bug. Something should be fixed in the compiler. We shouldn't workaround the issue because then we'll forget about it and it will be triggered in a different situation.

PR #4874, opened 2017-08-22
Add configurable default target

RX14:
This supports, for example, a compiler compiled for the `x86_64-unknown-linux-musl` architecture which doesn't require the `--target` command-line option to be set when ran on `x86_64-unknown-linux-gnu`. This is equivalent to the `--target` configure option in autoconf.

PR #7479, opened 2019-02-23
Add detection of linux environment in compiler config

### LINKED ISSUE #7196: Default target triple baked into compiler fails on musl

straight-shoota:
Fixes #7196

This change allows to use the same compiler binary with both gnu and musl.

In the process, I renamed `Config.default_target_triple` to `Config.default_target` and it now returns a `Codegen::Target` instance. We need the instance anyway to check whether it's linux.

I also replaced all usages of `Target.new` with `Config.default_target` which better communicates the purpose and avoids weird interdependency.

This does not add a `--host-target` option, which was also discussed in #7196 but that can be left for another PR.

### DIFF
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -37,7 +37,7 @@ module Crystal
     @@default_target : Crystal::Codegen::Target?
 
     def self.default_target : Crystal::Codegen::Target
-      @@default_target || begin
+      @@default_target ||= begin
         target = Crystal::Codegen::Target.new({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple)
 
         if target.linux?
@@ -51,7 +51,7 @@ module Crystal
           target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
         end
 
-        @@default_target = target
+        target
       end
     end
 
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -41,9 +41,9 @@ module Crystal
         target = Crystal::Codegen::Target.new({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple)
 
         if target.linux?
-          # The linux binary runs as well on linux-gnu as linux-musl, but the
-          # default target needs to match the C library used on the current
-          # system.
+          # The statically linked linux binary runs as well on linux-gnu as
+          # on linux-musl, but the default target needs to match the C
+          # library used on the current system.
           # This can be automatically detected from the output of `ldd --version`
           # in order to use the appropriate environment target.
           default_libc = target.gnu? ? "-gnu" : "-musl"
@@ -57,7 +57,7 @@ module Crystal
 
     private def self.runtime_libc
       ldd_version = String.build do |io|
-        Process.new("ldd", {"--version"}, output: io, error: io).wait
+        Process.run("ldd", {"--version"}, output: io, error: io)
       rescue Errno
         # In case of an error (eg. `ldd` not available), we assume it's gnu.
         return "gnu"
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -18,7 +18,7 @@ module Crystal
         Crystal #{version} #{formatted_sha}(#{date})
 
         LLVM: #{llvm_version}
-        Default target: #{self.default_target_triple}
+        Default target: #{self.default_target}
         DOC
     end
 
@@ -34,8 +34,40 @@ module Crystal
       Time.unix(time).to_s("%Y-%m-%d")
     end
 
-    def self.default_target_triple
-      {{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple
+    @@default_target : Crystal::Codegen::Target?
+
+    def self.default_target : Crystal::Codegen::Target
+      @@default_target || begin
+        target = Crystal::Codegen::Target.new({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple)
+
+        if target.linux?
+          # The linux binary runs as well on linux-gnu as linux-musl, but the
+          # default target needs to match the C library used on the current
+          # system.
+          # This can be automatically detected from the output of `ldd --version`
+          # in order to use the appropriate environment target.
+          musl_environment = musl_environment?
+          if musl_environment && target.gnu?
+            target = Crystal::Codegen::Target.new(target.to_s.sub("-gnu", "-musl"))
+          elsif !musl_environment && target.musl?
+            target = Crystal::Codegen::Target.new(target.to_s.sub("-musl", "-gnu"))
+          end
+        end
+
+        @@default_target = target
+      end
+    end
+
+    private def self.musl_environment?
+      ldd_version = String.build do |io|
+        Process.new("ldd", ["--version"], output: io, error: io).wait
+      rescue Errno
+        # In case of an error (for example `ldd` not available), we simply
+        # assume it's not musl
+        return false
+      end
+
+      ldd_version.starts_with?("musl")
     end
   end
 end
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -46,28 +46,29 @@ module Crystal
           # system.
           # This can be automatically detected from the output of `ldd --version`
           # in order to use the appropriate environment target.
-          musl_environment = musl_environment?
-          if musl_environment && target.gnu?
-            target = Crystal::Codegen::Target.new(target.to_s.sub("-gnu", "-musl"))
-          elsif !musl_environment && target.musl?
-            target = Crystal::Codegen::Target.new(target.to_s.sub("-musl", "-gnu"))
-          end
+          default_libc = target.gnu? ? "-gnu" : "-musl"
+
+          target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
         end
 
         @@default_target = target
       end
     end
 
-    private def self.musl_environment?
+    private def self.runtime_libc
       ldd_version = String.build do |io|
         Process.new("ldd", ["--version"], output: io, error: io).wait
       rescue Errno
         # In case of an error (for example `ldd` not available), we simply
-        # assume it's not musl
-        return false
+        # assume it's gnu.
+        return "gnu"
       end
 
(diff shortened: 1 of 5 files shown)


src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,40 @@ module Crystal
       Time.unix(time).to_s("%Y-%m-%d")
     end
 
-    def self.default_target_triple
-      {{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple
+    @@default_target : Crystal::Codegen::Target?
+
+    def self.default_target : Crystal::Codegen::Target
```
asterite: (optional) This can be written as

```crystal
getter(default_target : Crystal::Codegen::Target) do 
  # code here
end
```

asterite:
I guess my question is: why are we changing the default target triple that LLVM gives us? As far as I understand it, it's because we must check whether we use libc or musl...? If that's the case, shouldn't LLVM do that check and return us the correct target triple?

(I'm trying to understand whether this is something we should change on our side or it should be changes on LLVM's side)

src/compiler/crystal/config.cr
straight-shoota: I'm not particularly drawn to a specific style here, but I like being able to grep for `def self.default_target`.

straight-shoota:
LibLLVM.get_default_target_triple` seems to return the target triple the compiler was built for.

I've just confirmed this by compiling this simple program on alpine/musl: `require "llvm"; puts LLVM.default_target_triple`. It  prints `x86_64-alpine-linux-musl` on both alpine and ubuntu.

LLVM's default target triple doesn't take into account the current environment at runtime. On release builds it's overriden by `CRYSTAL_CONFIG_TARGET` anyway. But that's not the issue here as both values are fixed in at compile time.

In order to use a single binary for both gnu and musl we need to look at the runtime system and adjust the environment accordingly. I don't think LLVM provides a tool for that.

straight-shoota:
Actually, I think we can get rid of `CRYSTAL_CONFIG_TARGET` which was introduced by #4874 in order to be able to build a compiler with gnu as default target from musl.
With this change we shouldn't need that anymore because it won't matter whether a musl or gnu target is baked in, it should work on both.

asterite:
Oooh... I understand now. I though LLVM would dynamically compute it from the machine.

Then, shouldn't we completely not use LLVM and compute thd triple from lld, uname, etc.?

straight-shoota:
@asterite Yeah, that's essentially what @RX14 said in https://github.com/crystal-lang/crystal/issues/7196#issuecomment-450639640 to somehow use or re-implement `config.guess`.
But that's a story for another day. Right now, this fixes the only issue we're aware of, the interoperability of gnu and musl on linux.

straight-shoota:
I added a second commit that makes the implementation extensible and reduces code.

j8r:
Does it potentially allows to remove the Debian part in the multistage build in distribution-scripts, and only use Alpine?
If so, this can solve #7480

straight-shoota:
@j8r Yes, this should help to simplify the build process. But that's an entirely different story.

This may or may not affect #7480 and I don't see anything more than a hunch suggesting that.
It doesn't matter anyway, because we need to solve #7480 *before* the next release (in order to make the release) and this environment detection can only be used *after* the next release.

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,41 @@ module Crystal
...
+      @@default_target || begin
```
RX14: perhaps

```cr
@@default_target ||= begin
  ...
  target
end
```

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,41 @@ module Crystal
...
+          target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
+        end
+
+        target
+      end
+    end
+
+    private def self.runtime_libc
+      ldd_version = String.build do |io|
+        Process.new("ldd", ["--version"], output: io, error: io).wait
```
j8r: `["--version"]` can be a Tuple `{"--version"}`

straight-shoota:
@bcardiff I'd suggest to leave `CRYSTAL_CONFIG_TARGET` in for now until we're certain we don't need it any more in the distribution scripts. In case there arise some issues.

bcardiff:
@straight-shoota Definitely! I meant in the future.

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,41 @@ module Crystal
...
+
+        target
+      end
+    end
+
+    private def self.runtime_libc
+      ldd_version = String.build do |io|
+        Process.new("ldd", {"--version"}, output: io, error: io).wait
+      rescue Errno
+        # In case of an error (for example `ldd` not available), we simply
```
wontruefree: make comment more succinct.
```suggestion
        # In case of an error (eg. `ldd` not available), we assume gnu.
```

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,41 @@ module Crystal
...
+
+        if target.linux?
+          # The linux binary runs as well on linux-gnu as linux-musl, but the
+          # default target needs to match the C library used on the current
+          # system.
+          # This can be automatically detected from the output of `ldd --version`
+          # in order to use the appropriate environment target.
+          default_libc = target.gnu? ? "-gnu" : "-musl"
+
+          target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
```
wontruefree: why not make setters for `architecture`, `vendor` and `environment`.  You would not have to make new targets and you would not have to rely on a `.sub` for the replacement, it would be an exact replacement.

straight-shoota: Yeah, that would actually be more complicated because `environment` is an array. A proper implementation would have to loop through it to find the item to replace. It should usually be the first one, but AFAIK there are no guarantees.

This implementation is nice and simple and should be fine for now. If we ever change `Target` to have mutable fields, it can be refactored. But I don't think this needs to come with this change. It's fine as it is.

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,40 @@ module Crystal
...
+    @@default_target : Crystal::Codegen::Target?
+
+    def self.default_target : Crystal::Codegen::Target
+      @@default_target ||= begin
+        target = Crystal::Codegen::Target.new({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple)
+
+        if target.linux?
+          # The linux binary runs as well on linux-gnu as linux-musl, but the
+          # default target needs to match the C library used on the current
+          # system.
```
ysbaddaden: That's not exactly true: this only happens *if* crystal was linked statically.

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,40 @@ module Crystal
...
+          target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
+        end
+
+        target
+      end
+    end
+
+    private def self.runtime_libc
+      ldd_version = String.build do |io|
+        Process.new("ldd", {"--version"}, output: io, error: io).wait
```
ysbaddaden: Why not `Process.run` directly?

src/compiler/crystal/config.cr
straight-shoota: Right. I'll state that more precisely.

This code could also be guarded by `flag?(:static)` but it shouldn't make a difference because a non-static binary can only run on one target platform (which should be equivalent to its default target).

src/compiler/crystal/config.cr
straight-shoota: Probably just overlooked that. Maybe I did something else in the beginning...

src/compiler/crystal/config.cr
```diff
@@ -34,8 +34,40 @@ module Crystal
...
+
+        if target.linux?
+          # The statically linked linux binary runs as well on linux-gnu as
+          # on linux-musl, but the default target needs to match the C
+          # library used on the current system.
+          # This can be automatically detected from the output of `ldd --version`
+          # in order to use the appropriate environment target.
+          default_libc = target.gnu? ? "-gnu" : "-musl"
+
+          target = Crystal::Codegen::Target.new(target.to_s.sub(default_libc, "-#{runtime_libc}"))
```
wontruefree: Is there a way to do the detection of the environment before creating the target?  It is probably not that much of an overhead but it seems like since you are using string detection you only need to make the target once.

```crystal
if ({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple).includes?("linux")
  Crystal::Codegen::Target.new(target.to_s.sub({"-gnu" => "-#{runtime_libc}", "-musl" => "-#{runtime_libc}"}))
else
  Crystal::Codegen::Target.new({{env("CRYSTAL_CONFIG_TARGET")}} || LLVM.default_target_triple)
end
```

bcardiff: It's true, but by using `#linux?` we are also checking that the name occurs only in a specific part, and the check is case insensitive already due to `LLVM.normalize_triple(target_triple.downcase)`.

It's good enough as is IMO.

straight-shoota: I agree. This code is executed exactly once per compiler run. There is no point in optimizing out a little bit of string operations while repeating already implemented logic.

wontruefree: That makes sense :)

straight-shoota:
Today's nightly build runs on alpine and ubuntu :+1:

Here is the x86_64 package if you want to give it a test drive: https://19521-6887813-gh.circle-artifacts.com/0/dist_packages/crystal-nightly-20190307-1-linux-x86_64.tar.gz

PR #7483, opened 2019-02-26
WebSocket#close accepts code and reason arguments

### LINKED ISSUE #6469: WebSocket close frame messages aren't being parsed
The WebSocket specification says, regarding the body of close frames, that `the first two bytes of
   the body MUST be a 2-byte unsigned integer (in network byte order)
   representing a status code [...]`. However, the `HTTP::WebSocket` client simply treats the whole thing as a string. It would be more useful if the body were parsed into a `code` integer and a `reason` string.

vladfaust:
Fixes #6469

~~I saw WebSocket#close specs and they are pretty simple, but I'm absolutely not sure how to convert `String.new(bytes[2, bytes[1]]).should eq(message)` line to test both reason and code arguments.~~

Also see https://developer.mozilla.org/ru/docs/Web/API/WebSocket/close.

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -147,9 +147,9 @@ class HTTP::WebSocket
             code = @current_message.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
             reason = @current_message.gets_to_end
             @on_close.try &.call(code, reason)
-            close(code, reason) unless closed?
           rescue IO::EOFError
             @on_close.try &.call(nil, nil)
+          ensure
             close unless closed?
           end
 
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -53,7 +53,7 @@ class HTTP::WebSocket
   def on_binary(&@on_binary : Bytes ->)
   end
 
-  def on_close(&@on_close : String ->)
+  def on_close(&@on_close : Int16?, String? ->)
   end
 
   protected def check_open
@@ -90,11 +90,18 @@ class HTTP::WebSocket
     end
   end
 
-  # Closes the websocket with *reason* and *code*.
-  def close(reason : String? = nil, code : Int16? = nil)
+  # Closes the websocket.
+  def close
     return if closed?
     @closed = true
-    @ws.close(reason, code)
+    @ws.close
+  end
+
+  # Closes the websocket with *code* and *reason*.
+  def close(code : Int16, reason : String? = nil)
+    return if closed?
+    @closed = true
+    @ws.close(code, reason)
   end
 
   def run
@@ -102,7 +109,7 @@ class HTTP::WebSocket
       begin
         info = @ws.receive(@buffer)
       rescue IO::EOFError
-        @on_close.try &.call("")
+        @on_close.try &.call(nil, nil)
         break
       end
 
@@ -136,9 +143,16 @@ class HTTP::WebSocket
       when Protocol::Opcode::CLOSE
         @current_message.write @buffer[0, info.size]
         if info.final
-          message = @current_message.to_s
-          @on_close.try &.call(message)
-          close(message) unless closed?
+          begin
+            code = @current_message.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
+            reason = @current_message.gets_to_end
+            @on_close.try &.call(code, reason)
+            close(code, reason) unless closed?
+          rescue IO::EOFError
+            @on_close.try &.call(nil, nil)
+            close unless closed?
+          end
+
           @current_message.clear
           break
         end
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -90,10 +90,11 @@ class HTTP::WebSocket
     end
   end
 
-  def close(message = nil)
+  # Closes the websocket with *reason* and *code*.
+  def close(reason : String? = nil, code : Int16? = nil)
     return if closed?
     @closed = true
-    @ws.close(message)
+    @ws.close(reason, code)
   end
 
   def run
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -229,7 +229,20 @@ class HTTP::WebSocket::Protocol
     end
   end
 
-  def close(message = nil)
+  def close(reason : String? = nil, code : Int16? = nil)
+    if code
+      raw = uninitialized UInt8[2]
+      IO::ByteFormat::BigEndian.encode(code, raw.to_slice)
+
+      message = String.new(raw)
+
+      if reason
+        message += reason
+      end
+    else
+      message = reason
+    end
+
     if message
       send(message.to_slice, Opcode::CLOSE)
     else
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -229,25 +229,15 @@ class HTTP::WebSocket::Protocol
     end
   end
 
-  def close(reason : String? = nil, code : Int16? = nil)
-    if code
-      raw = uninitialized UInt8[2]
-      IO::ByteFormat::BigEndian.encode(code, raw.to_slice)
+  def close
+    send(Bytes.empty, Opcode::CLOSE)
+  end
 
-      message = String.new(raw)
-
-      if reason
-        message += reason
-      end
-    else
-      message = reason
-    end
-
-    if message
-      send(message.to_slice, Opcode::CLOSE)
-    else
-      send(Bytes.empty, Opcode::CLOSE)
-    end
+  def close(code : Int16, reason : String? = nil)
+    slice = Bytes.new(sizeof(Int16) + (reason ? reason.bytesize : 0))
+    IO::ByteFormat::NetworkEndian.encode(code, slice)
+    reason.to_slice.copy_to(slice + sizeof(Int16)) if reason
+    send(slice, Opcode::CLOSE)
   end
 
   def self.new(host : String, path : String, port = nil, tls = false, headers = HTTP::Headers.new)
--- a/spec/std/http/web_socket_spec.cr
+++ b/spec/std/http/web_socket_spec.cr
@@ -275,14 +275,16 @@ describe HTTP::WebSocket do
   end
 
   describe "close" do
-    it "closes with message" do
-      message = "bye"
+    it "closes with code and response" do
+      response = "bye"
+      code = 4000_i16
       io = IO::Memory.new
       ws = HTTP::WebSocket::Protocol.new(io)
-      ws.close(message)
+      ws.close(code, response)
       bytes = io.to_slice
       (bytes[0] & 0x0f).should eq(0x8) # CLOSE frame
-      String.new(bytes[2, bytes[1]]).should eq(message)
+      IO::ByteFormat::NetworkEndian.decode(Int16, bytes[2, 3]).should eq(code)
+      String.new(bytes[4, bytes[1] - 2]).should eq(response)
     end
 
     it "closes without message" do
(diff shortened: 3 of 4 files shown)


vladfaust:
I'm also thinking of parsing on socket close frame as well: 

https://github.com/crystal-lang/crystal/blob/60760a5460aa13a1f80c5a1a7410782dc4076b77/src/http/web_socket.cr#L138

So a developer would write

```crystal
socket.on_close do |reason, code|
  puts "Closed websocket with #{code} #{reason}"
end
```

src/http/web_socket.cr
```diff
@@ -90,10 +90,11 @@ class HTTP::WebSocket
     end
   end
 
-  def close(message = nil)
+  # Closes the websocket with *reason* and *code*.
+  def close(reason : String? = nil, code : Int16? = nil)
```
z64: I think `code` should be the first parameter. This matches the way its encoded in the payload, and lots of services I've dealt with more often than not will send *only* `code`.

You also cannot send *just* reason, so I suggest:

```cr
def close
  # ..
end

def close(code : Int16, reason : String? = nil)
  # ..
end
```
See RFC 6455: https://tools.ietf.org/html/rfc6455#section-5.5.1

>    The Close frame MAY contain a body (the "Application data" portion of
   the frame) that indicates a reason for closing, such as an endpoint
   shutting down, an endpoint having received a frame too large, or an
   endpoint having received a frame that does not conform to the format
   expected by the endpoint.  **If there is a body, the first two bytes of
   the body MUST be a 2-byte unsigned integer** (in network byte order)
   representing a status code with value /code/ defined in Section 7.4.
   Following the 2-byte integer, **the body MAY contain UTF-8-encoded data
   with value /reason/, the interpretation of which is not defined by
   this specification.**  This data is not necessarily human readable but
   may be useful for debugging or passing information relevant to the
   script that opened the connection.

z64:
Yes, we need to also have close frame decoding, but similar to my last comment, please:
```cr
ws.on_close do |code, reason|
  # ..
end
```

src/http/web_socket/protocol.cr
```diff
@@ -229,7 +229,20 @@ class HTTP::WebSocket::Protocol
     end
   end
 
-  def close(message = nil)
+  def close(reason : String? = nil, code : Int16? = nil)
+    if code
+      raw = uninitialized UInt8[2]
+      IO::ByteFormat::BigEndian.encode(code, raw.to_slice)
+
+      message = String.new(raw)
```
z64: I can't benchmark this right now, but I don't think creating this string is necessary over just creating the slice we need to send and writing everything to that:

```cr
  slice = Bytes.new(2 + reason.bytesize)
  IO::ByteFormat::NetworkEndian.encode(code, slice)
  reason.to_slice.copy_to(slice + 2)

  # ..

  send(slice, Opcode::CLOSE)
```
https://carc.in/#/r/6dg3

vladfaust:
Gonna work on it tomorrow

vladfaust:
Please help me writing `socket.on_close { |code, reason| }` specs (or write them for me, which is better)

z64:
Hi @vladfaust, following the [http negotiation spec](https://github.com/crystal-lang/crystal/blob/e09bfeb2695e5031f506905a4b4e56e293d6e5a5/spec/std/http/web_socket_spec.cr#L301-L336) I believe this should be acceptable. This passes for the existing 0.27.2 implementation:

```cr
it "calls close handler with code and reason" do
  address_chan = Channel(Socket::IPAddress).new

  spawn do
    http_ref = nil
    ws_handler = HTTP::WebSocketHandler.new do |ws, ctx|
      ws.on_message do
        # ws.close(1000, "goodbye")
        ws.close("goodbye")
        http_ref.not_nil!.close
      end
      ws.send("hello")
    end

    http_server = http_ref = HTTP::Server.new([ws_handler])
    address = http_server.bind_unused_port
    address_chan.send(address)
    http_server.listen
  end

  listen_address = address_chan.receive
  ws = HTTP::WebSocket.new("ws://#{listen_address}")

  ws.on_message do
    ws.send("hello")
  end

  # ws.on_close do |code, reason|
  ws.on_close do |message|
    # code.should eq 1000
    # reason.should eq "goodbye"
    message.should eq "goodbye"
  end

  ws.run
end
```
Edits for your new spec as commented.

Seeing as there aren't individual specs like this for the other handlers, maybe it should just be added to the spec I linked. I'm not sure though. 

I would add whatever you think is best so that we can be sure it works, and then spec style can be addressed in review 👍

src/http/web_socket.cr
```diff
@@ -90,18 +90,26 @@ class HTTP::WebSocket
...
+  # Closes the websocket.
+  def close
     return if closed?
     @closed = true
-    @ws.close(message)
+    @ws.close
+  end
+
+  # Closes the websocket with *code* and *reason*.
+  def close(code : Int16, reason : String? = nil)
```
z64: Actually, shouldn't `code` be `UInt16` everywhere?

>a 2-byte unsigned integer (in network byte order) representing a status code with value /code/

src/http/web_socket/protocol.cr
```diff
@@ -229,12 +229,15 @@ class HTTP::WebSocket::Protocol
...
   end
 
-  def close(message = nil)
-    if message
-      send(message.to_slice, Opcode::CLOSE)
-    else
-      send(Bytes.empty, Opcode::CLOSE)
-    end
+  def close
+    send(Bytes.empty, Opcode::CLOSE)
```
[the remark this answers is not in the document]
z64: Actually, maybe just remove this method and do:
```cr
def close(code : UInt16 = 1000, reason : String? = nil)
```

src/http/web_socket.cr
```diff
@@ -90,19 +90,26 @@ class HTTP::WebSocket
...
+    @ws.close(code, reason)
   end
 
   def run
     loop do
       begin
         info = @ws.receive(@buffer)
       rescue IO::EOFError
-        @on_close.try &.call("")
+        @on_close.try &.call(nil, nil)
```
Sija: You can skip `nil`s.

src/http/web_socket.cr
Sija: Why split it into 2 methods?

src/http/web_socket.cr
```diff
@@ -135,9 +143,16 @@ class HTTP::WebSocket
...
-          message = @current_message.to_s
-          @on_close.try &.call(message)
-          close(message) unless closed?
+          begin
+            code = @current_message.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
+            reason = @current_message.gets_to_end
+            @on_close.try &.call(code, reason)
+            close(code, reason) unless closed?
+          rescue IO::EOFError
+            @on_close.try &.call(nil, nil)
```
Sija: ditto

src/http/web_socket.cr
```diff
@@ -135,9 +143,16 @@ class HTTP::WebSocket
...
         if info.final
-          message = @current_message.to_s
-          @on_close.try &.call(message)
-          close(message) unless closed?
+          begin
+            code = @current_message.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
+            reason = @current_message.gets_to_end
+            @on_close.try &.call(code, reason)
+            close(code, reason) unless closed?
+          rescue IO::EOFError
```
Sija: Shouldn't this be `ensure`?

src/http/web_socket/protocol.cr
```diff
@@ -229,12 +229,15 @@ class HTTP::WebSocket::Protocol
...
+  end
+
+  def close(code : Int16, reason : String? = nil)
```
Sija: ditto

src/http/web_socket.cr
vladfaust: Nope, it raises if skipping `nil`s.

src/http/web_socket.cr
vladfaust: ditto

src/http/web_socket.cr
vladfaust: Makes sense

src/http/web_socket.cr
vladfaust: @Sija see in the discussion above

Sija: Can't find it, which discussion?

z64: @Sija See:

https://github.com/crystal-lang/crystal/pull/7483#discussion_r260421899

https://github.com/crystal-lang/crystal/pull/7483#discussion_r260759767

src/http/web_socket.cr
Sija: Well, the context of the comment doesn't explain it in any way, so I'll repeat my question: why split the `#close` method into 2 implementations?

z64: The RFC excerpt above lays it out.

> The Close frame MAY contain a body

So we have two methods:

1. `def close` - this sends no body (`Bytes.empty`), which will be interpreted implicitly by the server/client as close 1005 [RFC](https://tools.ietf.org/html/rfc6455#section-7.4.1)

2. `def close(code, reason = nil)` - Sends a body. When you send a body, `code` is required, and `reason` is optional.

This captures the two behaviors of section 5.5.1.

Sija: @z64 Implementation on the Crystal side have nothing to do with RFC(s).

IMO it should be simply defined as:

```cr
def close(code : UInt16? = nil, reason : String? = nil)
  # ...
end
```

z64: The problem with that is this allows a user to do

`ws.close(reason: "foo")`

This is invalid per ws protocol. You *must* supply a code. Having `code : UInt16` gives us compile time safety instead of a runtime check.

Sija: @z64 Oops, I missed that part, thanks for bearing with me! :) If that's the case, then implementation enforcing it during compile-time is definitely a good thing 👍

asterite:
Can't code have a default value other than nil?

z64:
@asterite In my suggestions I considered close code 1005, which is a special case; it is never meant to be sent by clients, but receiving clients interpret a body-less close frame as close 1005.

I suppose we could do

```cr
def close(code : UInt16 = 1005, reason : String? = nil)
  if code == 1005
    send(Bytes.empty, Opcode::CLOSE)
  else
    # send body
  end
end
```
but then this is confusing, because reason will never be sent:
```cr
ws.close(1005, "reason")
# or
ws.close(reason: "reason")
```
I think two methods is a cleaner separation of these semantics.

src/http/web_socket.cr
Sija: @vladfaust Please, don't _resolve_ the conversation until is actually implemented.

bew:
Reading https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#Properties
I think that documented error codes should be available in an enum (with their reason), and a custom close with code+reason should be available for other lib/framework/application errors

Sija:
@bew That would be too limiting. Docs especially state that some codes are:

>Reserved for use by WebSocket extensions.

or

>Reserved for future use by the WebSocket standard.

z64:
I don't think an enum would be bad.

Enums are not strict in accepting integer values by default, so I don't see how it would be limiting. `reason` though is just a payload, and is application specific. The *name* of the code is standard, though, much like HTTP status codes.

Basically, this proposal is #7247 but for websockets. I think this could be done in another PR though.

vladfaust:
FYI: I'm not very pushy about that because ~~developing Crystal itself is hard on my machine~~ I have this working [in my own code](https://github.com/onyxframework/http/blob/master/src/onyx-http/channel.cr#L175-L177). Should I add server specs here? :thinking:

src/http/web_socket.cr
Sija: 🏓

Issue #7422, opened 2019-02-12
TcpServer specs failure on Ubuntu 18.10


yxhuvud:
On clean master branch: 

Running 
`make clean && make && make spec`
consistently fails (for both ipv4 and ipv6), at 
`spec/std/socket/tcp_server_spec.cr:51` with the following error:

```
  1) TCPServer .new using IPv4 reuse_port raises when not binding with reuse_port
     Failure/Error: expect_raises_errno(Errno::EADDRINUSE, {% if flag?(:linux) %}"listen: "{% else %}"bind: "{% end %}) do

       Expected Errno with "listen: ", got #<Errno: bind: Address already in use> with backtrace:
         # /usr/share/crystal/src/socket/tcp_server.cr:73:15 in 'initialize:reuse_port'
         # /usr/share/crystal/src/socket/tcp_server.cr:32:3 in 'new:reuse_port'

```

Versions: 
System Crystal: Crystal 0.27.2 [60760a546] (2019-02-05)
Ubuntu 18.10
LLVM: 4.0.0
libc: 2.28-0ubuntu1

Looking at the spec, it would pass if not checking for `listen`, but for `bind` instead, but the code special cases linux to look for `listen`, and the spec changed somewhat recently, so the behaviour is probably intentional (though failing, which it obviously shouldn't).

Sija:
I'm curious why it needs to check for exception message at all, shouldn't `Errno::EADDRINUSE` be enough?

straight-shoota:
`EADDRINUSE` is not very precise. It still makes a difference if bind or listen fails. I'm not sure if that's worth keeping at all costs, but until now it didn't show any issues. Eventually, we want to move away from generic `Errno` exceptions anyway towards more specific exception classes, at which point this distinction vanishes anyway.

That aside, it's weird that Ubuntu 18.10 would already fail at `bind`. I'm pretty sure that this spec passed up to 18.10.

yxhuvud:
Could there be anything else that is triggering the issue?

straight-shoota:
This spec already fails in bionic 18.04. I did not yet check atomic 17.10.

I guess we should expect either `bind` or `listen` - at least for `gnu`.

asterite:
Can't we just check the exception type and not bother about the message, which is out of our control? (we don't generate it)

straight-shoota:
[...]

PR #7484, opened 2019-02-26
Fix expected error in tcp_server_spec for musl

straight-shoota:
It seems only glibc errors at listen. This spec failed on musl because it errors at bind.

We could also consider removing these strict checks for very specific error messages. It would probably suffice to check that it either contains `bind` or `listen` with any libc implementation.
But I guess being a bit more strict avoids accidentally triggering unexpected errors which would then still meet the expectation.

And in the end, it doesn't really matter if such specs happen to occasionally fail on a new platform. Having a few false positives isn't an issue.
In Crystal's aport this spec is simply muted (https://git.alpinelinux.org/aports/tree/community/crystal/fix-spec-std-socket-tcp_server_spec.cr.patch?id=29c6464be19088283c52dd74e90bbbd4f9829a6c).

### DIFF
--- a/spec/std/socket/tcp_server_spec.cr
+++ b/spec/std/socket/tcp_server_spec.cr
@@ -50,7 +50,7 @@ describe TCPServer do
 
         it "raises when not binding with reuse_port" do
           TCPServer.open(address, 0, reuse_port: true) do |server|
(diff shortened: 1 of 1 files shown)


straight-shoota:
Ceterum censeo continua integratio pro musl esse habendum #6943

yxhuvud:
quite related to #7422. Is that one also affected?

PR #7486, opened 2019-02-28
added documentation to BigInt, BigFloat, and BigDecimal

dscottboggs:
I feel that it's important to note that in order to use the big numbers, you need to `require "big"`. This adds some basic usage to the top of each of the Big... structs that don't already have it.

### DIFF
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -8,6 +8,17 @@ require "big"
 #
 # The general idea and some of the arithmetic algorithms were adapted from
 # the MIT/APACHE -licensed https://github.com/akubera/bigdecimal-rs
+#
+# ### Usage:
+# ```
+# require "big"
+# # From a string representation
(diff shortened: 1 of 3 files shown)


src/big/big_decimal.cr
```diff
@@ -8,6 +8,17 @@ require "big"
 #
 # The general idea and some of the arithmetic algorithms were adapted from
 # the MIT/APACHE -licensed https://github.com/akubera/bigdecimal-rs
+#
+# ### Usage:
+# ```
+# require "big"
+# # From a string representation
+# BigDecimal.new("123456789123456789.123456789123456789") # => 123456789123456789.123456789123456789
```
straight-shoota: I would actually suggest to use smaller, comprehensible numbers in these examples. The `Big*` types are not meant only for handling big numbers but for providing the *ability* to have arbitrary large and precise numbers. Smaller numbers work perfectly fine in these examples, too.

straight-shoota:
Closing stale PR.

PR #7490, opened 2019-02-28
Replace some as by ans

r00ster91:
This replaces some **a**s by **an**s.
There needs to be an **an** in front of "int" for example because "int" starts with a vowel, **i**.

This also adds `indefinitive_article` arguments to the `type_desc` methods so they return the correct indefinitive article for the word which is used in a lot of error messages.


spec/compiler/semantic/primitives_spec.cr
```diff
@@ -9,15 +9,15 @@ describe "Semantic: primitives" do
...
     assert_type("1_i64") { int64 }
   end
 
-  it "types a int128" do
+  it "types an int128" do
     assert_type("1_i128") { int128 }
   end
 
-  it "types a uint128" do
+  it "types an uint128" do
```
r00ster91: This is arguable because it depends on your preference to pronounce "uint": "unsigned integer" (an) or "you int" (a).
I think "an" is fine here because otherwise other strings have to be changed too then and now it would be consistent.

src/compiler/crystal/types.cr
```diff
@@ -593,7 +593,7 @@ module Crystal
       0
     end
 
-    def type_desc
+    def type_desc(indefinitive_article = false)
```
r00ster91: I don't think this method is ever invoked in the error messages with `type_desc(true)` but I'm not sure. Does anyone know for sure?
If it is, then an algorithm has to be made which determines whether "a" or "an" needs to be in front of the word and that could be quite complicated because this type could theoretically be anything, any word. But then the errors should probably just be reordered to avoid the "a".

j8r:
I think it's a bad idea to add the logic of a/an everywhere, I would rather prefer to have a new method to avoid this.
```cr
def type_desc_indefinitive
  "a union"
end
```
I haven't looked at the inheritance, but it may possible to have a default `type_desc_indefinitive` which would be `"a " + type_desc`, or checking if `type_desc` starts with a vowel logic, and override it for exceptions.

j8r:
What to we do?
Not sure about the added logic/complexity with `indefinitive_article`, it is not really a big deal IMHO.

straight-shoota:
If there is to be an automatic solution to use grammatically correct "a"/"an" it should probably be completely generic based on the first character of the interpolated name.

While that may be nice to have it is very, very low priority. We can certainly live with having "a" everywhere, even if "an" would be correct.

PR #7492, opened 2019-03-01
Fix time format RFC 3339 to not include offset seconds

### LINKED ISSUE #7489: Default serialization of Time in JSON  not RFC2822 nor ISO8601 compliant
I'm playing with crystal and moment.js; when I serialize the date into a json, it returns this string:

```crystal
{
  x: Time.now
}.to_json # => {"x": "2019-02-28T14:48:12+07:00:00"}
```

However, moment.js tell me the date format is invalid and should be RFC2822 compliant.

Basically, the seconds in the time zone stamp are not accepted. `2019-02-28T14:48:12+07:00` is accepted however.

I don't know if the time format outputted by crystal is RFC2822 compliant or not - if so, it could be a problem on moment.js library -; it seems any way overkill to have seconds in timezone, no?

### LINKED ISSUE #7491: Misleading error message when dispatching named arguments through double splat
The following code produces an error message that's not really helpful in identifying the issue.
```cr
def foo(foo = false)
end

def bar(**options)
  foo(**options)
end

bar(foo: true, baz: true) # Error: argument 'foo' already specified
```
The argument `foo` is specified, but actually only once. The real error is that there is no argument named `baz` for `foo`.
When calling `foo` directly instead of going through `bar` and the double splat argument, the error is:
```
Error in line 8: no argument named 'baz'
Matches are:
 - foo(foo = false)
```

straight-shoota:
Fixes #7489

This still allows parsing offset seconds even if seconds are not specified in RFC 3339. I think it's better to be a bit lenient here to be able to parse values even if they're not 100% correct according to the spec.

### DIFF
--- a/src/time/format/formatter.cr
+++ b/src/time/format/formatter.cr
@@ -175,7 +175,7 @@ struct Time::Format
     end
 
     def time_zone(with_seconds = false)
-      time_zone_offset(allow_seconds: with_seconds)
+      time_zone_offset(format_seconds: with_seconds)
     end
 
     def time_zone_z_or_offset(**options)
@@ -186,12 +186,12 @@ struct Time::Format
       end
     end
 
-    def time_zone_offset(force_colon = false, allow_colon = true, allow_seconds = true)
-      time.zone.format(io, with_colon: force_colon, with_seconds: allow_seconds)
+    def time_zone_offset(force_colon = false, allow_colon = true, format_seconds = true, parse_seconds = true)
+      time.zone.format(io, with_colon: force_colon, with_seconds: format_seconds)
     end
 
     def time_zone_colon(with_seconds = false)
-      time_zone_offset(force_colon: true, allow_seconds: with_seconds)
+      time_zone_offset(force_colon: true, format_seconds: with_seconds)
     end
 
     def time_zone_colon_with_seconds
@@ -203,7 +203,7 @@ struct Time::Format
     end
 
     def time_zone_rfc2822
-      time_zone_offset(allow_colon: false, allow_seconds: false)
+      time_zone_offset(allow_colon: false, format_seconds: false)
     end
 
     def time_zone_gmt_or_rfc2822(**options)
--- a/src/time/format/formatter.cr
+++ b/src/time/format/formatter.cr
@@ -186,7 +186,7 @@ struct Time::Format
       end
     end
 
-    def time_zone_offset(force_colon = false, allow_colon = true, format_seconds = true, parse_seconds = true)
+    def time_zone_offset(force_colon = false, allow_colon = true, format_seconds = false, parse_seconds = true)
       time.zone.format(io, with_colon: force_colon, with_seconds: format_seconds)
     end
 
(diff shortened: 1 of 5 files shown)


src/time/format/formatter.cr
```diff
@@ -186,12 +186,12 @@ struct Time::Format
       end
     end
 
-    def time_zone_offset(force_colon = false, allow_colon = true, allow_seconds = true)
-      time.zone.format(io, with_colon: force_colon, with_seconds: allow_seconds)
+    def time_zone_offset(force_colon = false, allow_colon = true, format_seconds = true, parse_seconds = true)
+      time.zone.format(io, with_colon: force_colon, with_seconds: format_seconds)
```
ysbaddaden: `parse_seconds` isn't used in this method;

asterite:
Yeah, I never saw seconds in time offset...

src/time/format/formatter.cr
straight-shoota: Yes, but `Format::Parser.time_zone_offset` does and both types should have the same interface for `Time::Pattern` (even though it's currently not used anywhere).

straight-shoota:
You're right. `format_seconds = true` was a mistake. There is not really any practical use of a time zone offset in seconds.

src/time/format/formatter.cr
ysbaddaden: Still, the parser and formatter shouldn't have irrelevant parameters. Please keep `allow_seconds`.

Note that I'd drop support for seconds in timezone offsets altogether, since neither RFC 3339, RFC 2822 or ISO 8601 support it.

straight-shoota: The time zone database is based on seconds and so is `Time:: Location`. It should be possible to completely serialize it. Offset seconds are used by the format pattern `%::z`.

straight-shoota: I'm not sure what you mean by keeping `allow_seconds`. That's semantically the same as `parse_seconds`. But I think the latter name fits better with the new argument `write_seconds`.

ysbaddaden: I mean: don't introduce new parameters that make no sense in some context (`parse_seconds` in formater?! `format_seconds` in parser?!). Keeping the existing and generic `allow_seconds` would fit in both context, and fit with other namings (such as `allow_colon`) too.

straight-shoota: We can't use the same name for different arguments. Both the formatter and parser implementations of the method can be called from the same pattern method (for example in `Pattern#rfc_3339`).

ysbaddaden: So? If a pattern wants seconds in the parser, then it also wants seconds in the formatter.

straight-shoota: No, not necessarily. A format could leniently support parsing seconds but don't print them. That's essentially what I'd suggested for the RFC 3339 format.

straight-shoota:
This fix would be a low-hanging fruit. @ysbaddaden Do you think the implementation is okay to merge?

asterite:
If we want to release 0.28.0 soon I'd say let's merge this. We can worry about low-level API details later on. It's better if we don't include offset seconds in the next release.

Sija:
GTG?

Issue #7209, opened 2018-12-20
Crystal Play linker issue: -lssl error


codingdracula:
```
OS: MacOS Mojave 10.14.2
```

```
Crystal 0.27.0 (2018-11-04)

LLVM: 6.0.1
Default target: x86_64-apple-macosx
```
```
OpenSSL version: LibreSSL 2.6.5
```

I was trying out crystal for the first time, so I followed: https://crystal-lang.org/docs/installation/on_mac_osx_using_homebrew.html

I did the following

```
brew update
brew install crystal
crystal play
```

I encountered the following error
```
Listening on http://127.0.0.1:8080
ld: library not found for -lssl (this usually means you need to install the development package for libssl)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
Error: execution of command failed with code: 1: `cc "${@}" -o '/Users/dlahr/.cache/crystal/crystal-run-play-1-1.tmp'  -rdynamic  -lz `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libssl || printf %s '-lssl -lcrypto'` `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libcrypto || printf %s '-lcrypto'` -lpcre -lgc -lpthread /usr/local/Cellar/crystal/0.27.0/src/ext/libcrystal.a -levent -liconv -ldl -L/usr/lib -L/usr/local/lib`
```

I then did the following
```
xcode-select --install
xcode-select --switch /Library/Developer/CommandLineTools
```

I restarted my machine and once again ran,
```
crystal play
```

The error does not occur right away, it occurs once I visit the playground i.e. `http://127.0.0.1:8080`

straight-shoota:
Hey @codingdracula, the error message says you need to install the libssl development package. Do you have that installed? I'm not a Mac user but AFAIK those `xcode-select` doesn't magically install some dependencies.

You can probably resolve this by running `brew install openssl` if you're using homebrew.

codingdracula:
Hey @straight-shoota I am aware some dependencies are missing, but there is no documentation that states what I need. There is no libssl package to download for example. (At least I can't easily find one.) As far as I can gather on the internet it is part of OpenSSL, which why I showed in my original post that I had OpenSSL...likely there is another package I need, but I don't know what that package is. 

This specifically happens when running the crystal playground. I can compile files and use the crystal compiler fine. 

[...]

Issue #6875, opened 2018-09-26
Crystal not working in Mac OS X Mojave


thebradhimself:
After the upgrade I reinstalled xcode command line tools.  
I've removed and reinstalled OpenSSL, MariaDB, Crystal via homebrew and added it back via crenv.  

I get the following error. 

> ld: library not found for -lssl (this usually means you need to install the development package for libssl)
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> Error: execution of command failed with code: 1: `cc "${@}" -o '/Users/bradleyhintze/workspace/nikola/truck/bin/truck'  -rdynamic  -lyaml -lz `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libssl || printf %s '-lssl -lcrypto'` `command -v pkg-config > /dev/null && pkg-config --libs --silence-errors libcrypto || printf %s '-lcrypto'` -lpcre -lgc -lpthread /usr/local/Cellar/crystal/0.26.1_1/src/ext/libcrystal.a -levent -liconv -ldl -L/usr/lib -L/usr/local/lib`

jasdeepsingh:
Solution:

`export PKG_CONFIG_PATH=/usr/local/opt/openssl/lib/pkgconfig`
Fish Shell: `set -U -x PKG_CONFIG_PATH /usr/local/opt/openssl/lib/pkgconfig`

KevinSjoberg:
To add to what @jasdeepsingh is saying. Ensure you have installed `pig-config`. This can be done running `brew install pig-config`. OpenSSL also need to be installed `brew install openssl`.

Let `pkg-config` know about your OpenSSL installation.

`export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/opt/openssl/lib/pkgconfig`

Verify `PKG_CONFIG_PATH` is set by running `echo $PKG_CONFIG_PATH`. Now, reinstall crystal `brew uninstall crystal` and `brew install crystal`.

joshuapinter:
In my case all I had to do was add `
export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/opt/openssl/lib/pkgconfig` to my `.bash_profile` file.

Should this be in the docs or installation feedback so people don't have to find it via the Googles?

Happy to write a PR for that.

bcardiff:
@joshuapinter thanks, It's already submitted in a PR https://github.com/crystal-lang/crystal-book/pull/312

[...]

PR #7494, opened 2019-03-01
Playground: Compile programs using without_openssl flag

bcardiff:
So, after upgrading to mojave...

I was able to experience #7209 / #6875 

Every program input in the playground will include `playground/agent.cr` which triggers the following dependencies

* https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/playground/agent.cr#L1
* https://github.com/crystal-lang/crystal/blob/master/src/http.cr#L2
* https://github.com/crystal-lang/crystal/blob/master/src/http/server.cr#L8

The websocket communication with the playground server does not require openssl, that is the reasone `without_openssl` flag exists actually.

Adding that flag to the compilation process will prevent this issues. 

⚠️  It will also mean that openssl is not available to be used in the scripts in the playground. ⚠️ 

Another alternative would be to detect if it is available, and if not, ignore it.
It seems that the previous default openssl library was enough to build the samples in osx.

cc: @dustinlahr, @thebradhimself

### DIFF
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -55,6 +55,7 @@ module Crystal::Playground
       output_filename = tempfile "play-#{@session_key}-#{tag}"
       compiler = Compiler.new
       compiler.color = false
+      compiler.flags << "without_openssl"
       begin
         @logger.info "Instrumented code compilation started (session=#{@session_key}, tag=#{tag})."
         result = compiler.compile sources, output_filename


Sija:
IMHO unconditionally compiling Playground without OpenSSL seems bit problematic choice, since there might be many situations where you'd like to use `HTTP` calls to ssl-enabled endpoints.

bcardiff:
I agree with @Sija but it was good to get to the bottom line :-)

Another alternative would be to use another mechanism to communicate the running script with the playground server.

straight-shoota:
Closing stale PR. There have been no further reportings that linking openssl was a problem with the playground. So it doesn't appear to be in need of fixing. Instead, it could break things.

PR #7495, opened 2019-03-01
Implement Set#add? method

Sija:
Same as [`Set#add?`](https://ruby-doc.org/stdlib-2.5.3/libdoc/set/rdoc/Set.html#method-i-add-3F) from Ruby's stdlib.

_chickendan_ mentioned on a Gitter chat about a need for such API to easily (in one go) distinguish whether upserted value was set before.

### DIFF
--- a/spec/std/set_spec.cr
+++ b/spec/std/set_spec.cr
@@ -40,6 +40,19 @@ describe "Set" do
     end
   end
 
+  describe "add?" do
+    it "returns self when object is not in the set" do
+      set = Set(Int32).new
+      set.add?(1).should eq(set)
+    end
+
+    it "returns nil when object is in the set" do
+      set = Set(Int32).new
+      set.add?(1)
+      set.add?(1).should be_nil
+    end
+  end
+
   describe "delete" do
     it "deletes an object" do
       set = Set{1, 2, 3}
--- a/spec/std/set_spec.cr
+++ b/spec/std/set_spec.cr
@@ -41,15 +41,16 @@ describe "Set" do
   end
 
   describe "add?" do
-    it "returns self when object is not in the set" do
+    it "returns true when object is not in the set" do
       set = Set(Int32).new
-      set.add?(1).should eq(set)
+      set.add?(1).should be_true
     end
 
-    it "returns nil when object is in the set" do
+    it "returns false when object is in the set" do
       set = Set(Int32).new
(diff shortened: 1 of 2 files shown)


j8r:
Why having this type of methods only for `Set`, whereas it can be in other `Enumerable` that implements `includes?`, like `push?` and `unshift?`?

No, I don't see a real need to have it in the stdlib. It's already very easy to implement by ourself.

Sija:
> When a value is added to a Hash you can know whether it existed or not.

Can you? I haven't any way to do that...

Sija:
@j8r Because `Set` is designed with unique values in mind, whereas other `Enumerable`-compatible classes are not.

straight-shoota:
@asterite The hash lookup could be improved, but I suppose we can merge this without it (and maybe add a `TODO`). 

I'm somehow wondering about the return value `Set | Nil`. It's the same as in Ruby. But why not return `Bool`? I can't think of a use case where you would have any benefit from returning `self`.

asterite:
Yeah, bool sounds better ai think. And I agree about delaying the optimization.

Sija:
@straight-shoota @asterite I've pushed a commit with proposed changes.

spec/std/set_spec.cr
```diff
@@ -40,6 +40,19 @@ describe "Set" do
     end
   end
 
+  describe "add?" do
+    it "returns true when object is not in the set" do
```
straight-shoota: Please add an expectation to make sure the value was actually added to the set.

src/set.cr
```diff
@@ -71,6 +71,19 @@ struct Set(T)
     self
   end
 
+  # If the *object* is not in the set, adds it and returns `true`,
```
straight-shoota: The description should not start with a conditional clause, makes it hard to understand.
```suggestion
  # Adds *object* to the set and returns `true` on success and `false` if the value was already in the set.
```

spec/std/set_spec.cr
Sija: Done.

straight-shoota: I meant without relying on the method that's currently being tested. Just append a `set.includes?(1).should be_true`.

Sija: Good point, I added it in force-pushed commit.

PR #7498, opened 2019-03-03
Allow to check Hash subset in ContainExpectation

### LINKED ISSUE #7497: Add named overloads to the contain spec macro for use with key-value pairs.
There was a discussion in the Gitter chat recently surround some confusion around using the `contain` spec macro. Basically, someone was just trying to do this...

```crystal
describe "a hash" do
  it "has a particular value" do
    a_hash.should contain "a value"
  end
end
```

which was naturally always returning false, as `Hash#includes?` checks for the presence of a key-value pair, rather than either a key or a value. I propose adding required-named-argument overloads to `Hash#includes?` and refactor `contain` to be able to forward those on, so that we could do:

```crystal
describe "a hash" do
  it "has a particular value" do
    a_hash.should contain value: "a value"
    a_hash.should contain key: "a key"
  end
end
```

Sija:
Fixes #7497

### DIFF
--- a/src/spec/expectations.cr
+++ b/src/spec/expectations.cr
@@ -220,7 +220,14 @@ module Spec
     end
 
     def match(actual_value)
-      actual_value.includes?(@expected_value)
+      expected_value = @expected_value
+      if expected_value.is_a?(Hash) && actual_value.is_a?(Hash)
+        expected_value.all? do |key, value|
+          actual_value[key]? == value
+        end
+      else
+        actual_value.includes?(expected_value)
+      end
     end
 
     def failure_message(actual_value)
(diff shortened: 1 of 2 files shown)


src/spec/expectations.cr
```diff
@@ -220,7 +220,14 @@ module Spec
     end
 
     def match(actual_value)
-      actual_value.includes?(@expected_value)
+      expected_value = @expected_value
+      if expected_value.is_a?(Hash) && actual_value.is_a?(Hash)
+        expected_value.all? do |key, value|
+          actual_value[key]? == value
```
j8r: This behavior isn't consistent with how `includes?` works. I don't know what problem it's supposed to solve.

Sija: It was discussed at length at Gitter as well as in several comments in before-mentioned issue. IMHO giving downvotes before you understand the problem at hand doesn't make much sense and if anything it shows an ill-will you hold for some reason...

PR #7499, opened 2019-03-03
Add Hash#includes?(other : Hash) overload

### LINKED ISSUE #7497: Add named overloads to the contain spec macro for use with key-value pairs.

Sija:
This PR allows checking if a `Hash` includes given subset denoted by given *other* `Hash`.
Functionality asked for instance at [SO](https://stackoverflow.com/questions/7584801/how-to-check-if-an-hash-is-completely-included-in-another-hash) - Ruby version.

Supersedes #7498 since it's a more versatile implementation, enhancing `Hash` API without a need for introducing special cases in `ContainExpection` matcher.

### DIFF
--- a/spec/std/hash_spec.cr
+++ b/spec/std/hash_spec.cr
@@ -97,6 +97,22 @@ describe "Hash" do
     end
   end
 
+  describe "#includes?" do
+    it "returns true if all key-values are matching" do
+      h = (0..10).to_h { |i| {i.to_s, i} }
+      h.includes?({"0" => 0}).should be_true
+      h.includes?({"0" => 0, "10" => 10}).should be_true
+    end
+
+    it "returns false if any key-value is different" do
+      h = (0..10).to_h { |i| {i.to_s, i} }
+      h.includes?({0 => 0}).should be_false
+      h.includes?({"0" => 10}).should be_false
+      h.includes?({"0" => 0, "10" => 11}).should be_false
+      h.includes?({"0" => 0, "11" => 10}).should be_false
+    end
+  end
+
   describe "[]=" do
     it "overrides value" do
       a = {1 => 2}
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -141,6 +141,23 @@ class Hash(K, V)
     end
   end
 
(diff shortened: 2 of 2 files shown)


j8r:
This doesn't closes #7497 ...
I'm not for or against this feature, at least I'm glad you understood https://github.com/crystal-lang/crystal/pull/7498/files#r261847933 – no need to have custom code related to `Hash` on the expectation anymore.

asterite:
Ruby provides `<`, `<=`, `>` and `>=` for this (subset, subset or equals, superset, superset or equals). Let's do the same.

dscottboggs:
I don't think it's accurate to say that this closes #7497. That issue was about a totally different question.

Sija:
@asterite Cool, these are way better! How about also delegating `Hash#includes?(other : Hash)` to `other <= self` to allow `contain` expectation to work smoothly with it?

straight-shoota:
`#includes?(a)` is defined as returning true when `a` is part of the collection. You're proposing an entirely different semantic, which is actually a subset predicate. This should be a different method and would be implemented by #7500.

I'd suggest to close this PR.

straight-shoota:
I've given this some thought and maybe we could consider adding an overload `Enumerable(T).includes?(collection : Enumerable)` which returns true if all elements in `collection` are included in `self`. So that's essentially `other.all? { |elem| includes?(elem) }`.

/cc https://github.com/crystal-lang/crystal/pull/7500#issuecomment-472214553

asterite:
If we do that it should be named `includes_all? `

PR #7505, opened 2019-03-04
Refactor IO::Syscall as IO::Evented

ysbaddaden:
I started this refactor in my MT experiments, but I think it stands on its own:

`Evented` is closer to what the module is trying to achieve (abstract event loop details), and method naming follows that from `IO::Buffered` (e.g. `evented_read` instead of `read_syscall_helper`).

Refactors the `FileDescriptor` and `Socket` IO classes to move implementation details into `IO::Evented` to 1. reduce duplication, and 2. avoid direct accesses to internal details (such as `@writers` or `@read_event`).

### DIFF
--- a/src/crystal/event_loop.cr
+++ b/src/crystal/event_loop.cr
@@ -22,60 +22,32 @@ module Crystal::EventLoop
     end
   end
 
-  def self.create_fd_write_event(io : IO::FileDescriptor, edge_triggered : Bool = false)
+  def self.create_fd_write_event(io : IO::Evented, edge_triggered : Bool = false)
     flags = LibEvent2::EventFlags::Write
     flags |= LibEvent2::EventFlags::Persist | LibEvent2::EventFlags::ET if edge_triggered
-    event = @@eb.new_event(io.fd, flags, io) do |s, flags, data|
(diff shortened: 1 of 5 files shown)


ysbaddaden:
Huum, I think it should be `:nodoc:`. Maybe moved to `Crystal` namespace, along `Event` and `EventLoop`. Maybe even moved to `Crystal::System` since it's only used on POSIX platforms.

yxhuvud:
(quite OT, but related to the greater topic of evented IO on Linux) I wonder how much of this will change when https://lwn.net/Articles/776703/ is fully merged and accessible from userspace.

ysbaddaden:
Having them moved back into IO::Evented helped in my MT experiments, thought I moved most libevent responsibility into EventLoop, and use a single Event on Fiber directly, which holds the status (e.g. added, canceled). See https://github.com/ysbaddaden/crystal/commit/2ea421a2858db3398d1c6314b56031ed4137f79f

ysbaddaden:
@bcardiff it was already :)

bew:
It looks like everybody agree, let's get going! :rocket: 
ok

straight-shoota:
Thank you @ysbaddaden!

Issue #7493, opened 2019-03-01
Automatically launch browser window when running `crystal play`.


joshuapinter:
### Preamble
I'm new to Crystal but I'm absolutely loving it as I've been a Ruby and Rails dev/contributor for over a decade.

This is a feature request that I'm willing write a PR for but as this is my first intro to the Crystal Github community, I wanted to gauge your reaction first before writing a PR.

### Feature Request
`crystal play` is incredible, especially when learning. However, it's a bit of a pain to have to manually open up a browser window and put in the correct address and port where the Crystal Playground is being served.

I would much prefer to just type `crystal play` and have it launch the server as well as the default browser window to the correct IP address and port. Something as simple as `open 127.0.0.1:8080`.

If for whatever reason you think that this should not be the default behaviour, we could optionally keep it tucked behind a command line option on `play`, like `-o` or `--open`.

Anyway, thanks for all your work on this impressive and exciting language, and let me know your feelings on this. If you like it, either as the default or as an option, I'll write up a PR against it this weekend.

Thanks! 

### Versions
```bash
$ crystal --version
Crystal 0.27.2 (2019-02-05)

LLVM: 6.0.1
Default target: x86_64-apple-macosx
```

r00ster91:
That's a good idea! I think the browser should be launched by default whenever it's possible. Sometimes it won't be possible like on linux for me when I try `sensible-browser`:
```shell
$ sensible-browser
Couldn't find a suitable web browser!
Set the BROWSER environment variable to your desired browser.
```
Or `xdg-open`:
```shell
$ xdg-open
-bash: xdg-open: command not found
```
In that case the tab should be manually opened.

Sija:
I think that would be pretty useful, although:

> If for whatever reason you think that this should not be the default behaviour, we could optionally keep it tucked behind a command line option on `play`, like `-o` or `--open`.

I'd use flag for that.

bcardiff:
I'm not sure what is the most portable way to do that. 
Also, note that port and binding might change, but the HTTP server API should help to return the actual values used for that.

The changes to make this happen should be at https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/command/playground.cr

r00ster91:
> I'd use flag for that.

Or the user is asked if he wants to open it. I would find that more convenient than having to manually add a flag which is a bit similar to manually opening the tab:
```
$ crystal play
Do you want to open http://127.0.0.1:8080 in your browser?
```

[...]

PR #7507, opened 2019-03-04
WIP: Automatically launch browser window when running `crystal play`.

joshuapinter:
Only tested on macOS 10.14 so far. Works fantastic.

Want to get the eyes of the core team on this just to get their initial feedback on where this should be, etc.

Related to #7493.

### TODO

- [ ] Ubuntu 18.04
- [ ] Ubuntu 18.10
- [ ] Windows 10?
- [ ] Others?


src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,10 @@ class Crystal::Command
       end
     end
 
+    playground_url = "http://#{server.host || "127.0.0.1"}:#{server.port || "8080"}"
+
+    system "open #{playground_url}" # Open up the default browser automatically to the Playground.
```
jwoertink: Might as well move this out in to a method since you'll probably need to use some `flag` checks on the system anyway. `open_browser(playground_url)` or something like that. Then you can use:

```crystal
def open_browser(playground_url)
  {% if flag?(:darwin) %}
    system "open #{playground_url}"
  {% end %}
end
```

j8r: That's `xdg-open` on Linux.

joshuapinter: Great suggestion. Will definitely do that. For these `flag` values, is there a listing of these values I can use? I'm currently looking for a way to tell which OS we're on in order to use the right command.

jwoertink: I don't know about a specific place, but there's `:linux`, `:freebsd`, `:openbsd`, `:win32` being used around in the language. I would imagine the bsd ones probably won't need it, and the linux one might need it's own method for checking if `xdg-open` or `sensible-browser`.

straight-shoota: The supported values for `flags` are essentially components of the target triples. In this case you're interested in the operating system. Supported platforms are listed as subdirectories of `src/lib_c` and on https://github.com/crystal-lang/crystal/wiki/Platform-Support
Currently, that's `linux`, `darwin`, `freebsd`, `openbsd` (and `windows`).

r00ster91:
You can use `start (URL)` on Windows. (put it in a `{% elsif flag?(:win32) %}` macro branch)

src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,10 @@ class Crystal::Command
       end
     end
 
+    playground_url = "http://#{server.host || "127.0.0.1"}:#{server.port || "8080"}"
```
r00ster91: What about simply using `localhost`?
```suggestion
    playground_url = "http://#{server.host || "localhost"}:#{server.port || "8080"}"
```
That would be better especially for people who are on Linux and don't have `xdg-open` installed. (The people who have to manually open the tab). They can easily remember `localhost:8080` instead of remembering that number.

joshuapinter: This was done in d0c8a6bc34a8b1aad3473dd48928fc871bcfc59d. 

Didn't realize I could just commit your suggestion. New feature of Github I guess. :)

src/compiler/crystal/command/playground.cr
joshuapinter: This is now a private method called `open_browser` and I added commands for `darwin`, `win32` and `linux` with flag checks. I stubbed out `freebsd` and `openbsd` but I don't know the proper commands to use for those.

I had [trouble building crystal for Ubuntu 18.04](https://github.com/crystal-lang/crystal/issues/5556#issuecomment-469404469) so I wasn't able to test beyond macOS.

Also, I wasn't sure if the windows flag was `win32` or `windows` as @jwoertink said `win32` and @straight-shoota said `windows`.

src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,26 @@ class Crystal::Command
...
+  # Open up the default browser automatically to the Playground.
+  private def open_browser(playground_url)
+    {% if flag?(:darwin) %}
+      system "open #{playground_url}"
+    {% elsif flag?(:win32) %}
+      system "start #{playground_url}"
+    {% elsif flag?(:linux) %}
+      system "xdg-open #{playground_url}"
+    # TODO
+    # {% elsif flag?(:freebsd) %}
```
r00ster91: If you plan to uncomment these macro branches, I think `:freebsd` and `:openbsd` are included in `:linux` so you probably don't need them.

joshuapinter: Perfect, thanks @r00ster91!

src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,21 @@ class Crystal::Command
       end
     end
 
+    playground_url = "http://#{server.host || "localhost"}:#{server.port || "8080"}"
+
+    open_browser(playground_url)
```
Sija: Please, put it behind a flag as mentioned in the https://github.com/crystal-lang/crystal/issues/7493#issuecomment-468737132.

src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,21 @@ class Crystal::Command
...
   end
+
+  # Open up the default browser automatically to the Playground.
+  private def open_browser(playground_url)
+    {% if flag?(:darwin) %}
+      system "open #{playground_url}"
+    {% elsif flag?(:win32) %}
+      system "start #{playground_url}"
+    {% elsif flag?(:linux) %}
+      system "xdg-open #{playground_url}"
```
r00ster91: When this command fails (it does for me on Linux), it will print an ugly `--: 1: --: xdg-open: not found`.
This can be avoided by running the command without configuring it with an input, output or error. `system` configures it with an input, output or error as you can see [here](https://github.com/crystal-lang/crystal/blob/60760a5460aa13a1f80c5a1a7410782dc4076b77/src/process.cr#L536) (it uses [`Process::Redirect::Inherit`](https://crystal-lang.org/api/0.27.2/Process/Redirect.html#Inherit)) but `Process.run` doesn't do that by default and discards the IO (it uses [`Process::Redirect::Close`](https://crystal-lang.org/api/0.27.2/Process/Redirect.html#Close)).
```suggestion
      Process.run("xdg-open #{playground_url}", shell: true)
```

src/compiler/crystal/command/playground.cr
straight-shoota: What? Why should BSDs match the Linux flag?

r00ster91: Oh, then  probably `:unix` which will definitely include them.

src/compiler/crystal/command/playground.cr
joshuapinter: Can we get an official stance on whether this should be contained behind a flag or not?

With the latest changes suggested by @r00ster91 to use `Process.run` instead of `system`, that should hide any negative effects (e.g. error output) if the command fails.

I imagine the only thing left to discuss would be whether people would find it annoying that the browser window automatically opens to the Playground when using `crystal play`.

As a newcomer to the language, I certainly would prefer it to open automatically but I'm not sure if that is inconvenient to others.

src/compiler/crystal/command/playground.cr
joshuapinter: So, should we change `flag?(:linux)` to `flag?(:unix)`?

src/compiler/crystal/command/playground.cr
straight-shoota: I think this definitely needs to be configurable. As already mentioned, this is a very opinionated features.
You should be able to run the playground for example as a background service and not try to open any http URL.

And I'd still question whether this flag should be active by default (is it opt-in or opt-out?). I know that opt-in would largely reduce it's usefulness, but again, I'm not even sure this needs to be implemented as a playground feature.

src/compiler/crystal/command/playground.cr
```diff
@@ -34,6 +34,21 @@ class Crystal::Command
       end
     end
 
+    playground_url = "http://#{server.host || "localhost"}:#{server.port || "8080"}"
```
j8r: `8080` is already the default (https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/playground/server.cr#L460)

src/compiler/crystal/command/playground.cr
Sija: I can't imagine situation of having such feature enabled without the ability to opt-out, c'mon... In any case I'd also question whether should it be turned on by default.

bcardiff: I think the official should be to open the browser. Is better for newcomers and offers discoverability of the feature.

So I would go with an opt-out flag `-n --no-launch`.

src/compiler/crystal/command/playground.cr
j8r: I even think  `"localhost"` can to be dropped to when looking at https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/playground/server.cr#L549 .

You can change to `property host : String = Socket::IPAddress::LOOPBACK` in https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/playground/server.cr#L453

src/compiler/crystal/command/playground.cr
r00ster91: Yes.

src/compiler/crystal/command/playground.cr
joshuapinter: Thanks for the review @j8r!

To be clear: 

1. Can I just remove the ` || "8080"` from `server.port || "8080"`?

2. Can I just remove ` || "localhost"` from `server.host || "localhost"`? I'll have to check later tonight but as I recall the value of `server.host` was not set.

Thanks and sorry for my newbness.

j8r: You delete both yes. `server.host` is nillable (`String?`), with `nil` as default value. You have to change it to `String` in `Playground::Server`, with `Socket::IPAddress::LOOPBACK` as default value. See my links above to have hints where changing the code.

joshuapinter: Ah, gotcha. I'll make those changes and get you to look at it. Thanks again.

straight-shoota: The exact address to connect to is only known after the socket is bound. For example, with `--port 0`, the port is chosen by the socket implementation.

So this needs to be moved after `server.start`.

joshuapinter: @straight-shoota If I move the `open_browser` call after the `server.start` it never gets called, or at least on macOS the browser window never launches. Any way on getting around that?

joshuapinter: @j8r When I change `property host : String?` to `property host : String = Socket::IPAddress::LOOPBACK` I get the following errors when running `make`: 

```bash
Error in src/compiler/crystal.cr:8: instantiating 'Crystal::Command.class#run()'

Crystal::Command.run
                 ^~~

in src/compiler/crystal/command.cr:45: instantiating 'run(Array(String))'

  def self.run(options = ARGV)
  ^

in src/compiler/crystal/command.cr:46: instantiating 'Crystal::Command#run()'

    new(options).run
                 ^~~

in src/compiler/crystal/command.cr:70: instantiating 'playground()'

      playground
      ^~~~~~~~~~

in src/compiler/crystal/command/playground.cr:8: instantiating 'Crystal::Playground::Server.class#new()'

    server = Playground::Server.new
                                ^~~

in src/compiler/crystal/tools/playground/server.cr:459: instance variable '@host' of Crystal::Playground::Server must be String, not Nil

      @host = nil
      ^~~~~

make: *** [.build/crystal] Error 1
```

Do I need to set `@host = nil` to `@host = Socket::IPAddress::LOOPBACK`?

If so, something very strange happens and I'm unsure if it's related or just a coincidence.

First off, the `make` command completes, so that's great, but then when I run `bin/crystal play` the crystal text that is in the `playground` is different than before the change. See the following comparison (left is before change and right is after change):

Screen Shot 2019-03-13 at 6 00 46 PM

Anyway, let me know what your thoughts are on this before I proceed with removing the fallback to `localhost`. 

Thanks.

src/compiler/crystal/command/playground.cr
joshuapinter: Added `-n` option in 1c90e117c1cb64b2cf7cec9ca4dc0d6d0f2e32e3. Works great.

src/compiler/crystal/command/playground.cr
joshuapinter: Replaced `linux` with `unix` in 987e539a2ed645f68a874a96812275fbdd9bc589.

src/compiler/crystal/command/playground.cr
```diff
@@ -27,13 +28,32 @@ class Crystal::Command
...
+      end
+
       opts.unknown_args do |before, after|
         if before.size > 0
           server.source = gather_sources([before.first]).first
         end
       end
     end
 
+    playground_url = "http://#{server.host || "localhost"}:#{server.port}"
```
Sija: I'd put these 2 lines within `if` condition.

src/compiler/crystal/command/playground.cr
j8r: Remove the `@host = nil`

straight-shoota: You must not guess the adress. Use the one the server is bound to. `server.start` not only binds the socket but also listens wich will block the program. Execute launch code between bind and listen, for example by yielding to a block.

j8r: To set the host and the port:
```cr
if host
  server.host = host
end
if port
  server.port = port
end
```

You can yield the playground address, then inside the block open the browser. In `Playground::Server` the yield can be just before the `server.listen`.

PR #7510, opened 2019-03-04
Compiler: let `protected initialize` define a protected `new`

### LINKED ISSUE #7501: Able to access protected initialize method outside of module
```
module Db
  class Row
    getter cells : Array(String)
    
    protected def initialize(array : Array(String))
      @cells = array.clone
    end
  end
end

array = ["hi", "hello", "howdy"]
row = Db::Row.new array
puts row.cells
```

This code works and prints the array, but I expect an error because the initialize method is protected.

asterite:
Fixes #7501


asterite:
There seems to be a problem with included modules. This needs a bit more tweaking... I'll fix it later.

Issue #7508, opened 2019-03-04
Compiler error execvp: No such file or directory (Errno)


earlonrails:
```
sh-4.2# crystal build --release server.cr
execvp: No such file or directory (Errno)
  from ???
  from ???
  from ???
...
Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
sh-4.2# crystal -v
Crystal 0.26.1 [391785249] (2018-08-27)

LLVM: 4.0.0
Default target: x86_64-unknown-linux-gnu
sh-4.2# uname -a
Linux 64809d7bf557 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

Dockerfile to reproduce
```
FROM amazonlinux:latest

RUN curl https://dist.crystal-lang.org/rpm/setup.sh | bash
RUN yum install -y crystal-0.26.1 git gcc gcc-c++

ENV PORTS=""
WORKDIR /app

COPY shard.* ./
COPY server.cr ./
RUN shards install
RUN crystal build --release server.cr

EXPOSE ${PORTS}

CMD server
```

Simple websocket kemal server
```
require "kemal"
require "random/secure"

Kemal.config.port = 8080

ws "/" do |socket|
  socket_id = Random::Secure.hex(32)
  puts "[#{socket_id}}] connected!"
  socket.send "connected!"
  socket.on_message do |message|
    puts "[#{socket_id}}] message: #{message}"
    socket.send "echo: #{message}"
  end

  # Remove clients from the list when it’s closed
  socket.on_close do
    puts "[#{socket_id}}] closed"
  end
end

Kemal.run
```

straight-shoota:
This seems to be an issue with `PATH`, either it's malconfigured or a required executable is missing. Probably similar to #7082.

If you prefix the `crystal build` command with `strace -f` it should print the arguments to `execvp`.

But I think we should really improve the error message for `execvp` to include the executable name. Users shouldn't be required to run strace just to fine out which file can't be executed.

earlonrails:
Thanks I will try the `strace -f`

earlonrails:
```
[pid    17] lseek(12, 51912232, SEEK_SET) = 51912232
[...]

PR #7511, opened 2019-03-04
Add command and args to execvp error message

straight-shoota:
When `LibC.execvp` fails, it would previously raise an `Errno` with message `execvp` followed by the reason of failure (for example `No such file or directory`). But that's not a great help for debugging, unless you know *which* command was to be executed.

This PR adds the command and it's arguments to the error message. I'm not sure if the arguments are really necessary.

Related to #7508

### DIFF
--- a/spec/std/process_spec.cr
+++ b/spec/std/process_spec.cr
@@ -17,8 +17,8 @@ describe Process do
 
   it "raises if command could not be executed" do
     # FIXME: Oddly doubled error message
-    expect_raises_errno(Errno::ENOENT, "execvp: No such file or directory: No such file or directory") do
-      Process.new("foobarbaz")
+    expect_raises_errno(Errno::ENOENT, %(execvp (foobarbaz "foo"): No such file or directory: No such file or directory)) do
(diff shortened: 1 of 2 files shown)


spec/std/process_spec.cr
```diff
@@ -17,8 +17,8 @@ describe Process do
 
   it "raises if command could not be executed" do
     # FIXME: Oddly doubled error message
-    expect_raises_errno(Errno::ENOENT, "execvp: No such file or directory: No such file or directory") do
-      Process.new("foobarbaz")
+    expect_raises_errno(Errno::ENOENT, %(execvp (foobarbaz "foo"): No such file or directory: No such file or directory)) do
```
yxhuvud: Does the printed command need a length restriction? If it has many arguments, then there could be some issues with the usability.

straight-shoota: Restricting the length could also have usability implications. The best solution would be to change the order and print the failed command after the libc error message. `Errno` doesn't support that though.

PR #7513, opened 2019-03-06
Document Unicode::CaseOptions enum

r00ster91:
I think the `Unicode::CaseOptions` enum should be publicly documented when it can be used in methods like `String#upcase` so you can easily see all the available case options.


asterite:
Sorry, I don't understand this change. The enum is already documented. And the module too.

r00ster91:
But not publicy documented. It's not in the API documentation. I think stuff that can be used in public methods (like `upcase`) should be documented too. When one wants to know the case options he needs to open GitHub first, find `unicode.cr` by opening `src` then scroll through till you eventually find the `unicode` folder, open it, then you perhaps don't know whether it's `data.cr` or `unicode.cr` (I don't think that's obvious for everyone) and then you finally see those 3 case options. That's pretty inconvient.

asterite:
Oh, I see, this is the relevant line:

```crystal
require "./unicode/unicode"
```

src/unicode/unicode.cr
```diff
@@ -1,13 +1,7 @@
-# Provides methods that answer questions about unicode characters,
-# and the `Unicode::CaseOptions` enum for special case conversions
-# like Turkic.
-#
-# There's no need to use the methods defined in this module
-# because they are exposed in `Char` in a more convenient way
-# (`Char#upcase`, `Char#downcase`, `Char#whitespace?`, etc.)
+# Provides the `Unicode::CaseOptions` enum for special case conversions like Turkic and
+# private and undocumented methods that are exposed in `Char`.
```
RX14: I'd rather not reference the undocumented methods at all. Leave this class with no doc comment and seperate the old doc comment from the module with a line.

r00ster91: You mean like this?
```cr
# ...
# because they are exposed in `Char` in a more convenient way
# (`Char#upcase`, `Char#downcase`, `Char#whitespace?`, etc.)

# :nodoc:
module Unicode
...
```
But then the module wouldn't show up in the API documentation and so neither the `CaseOptions` enum.

straight-shoota: "no doc comment" != "nodoc comment"

In your example, simply remove the nodoc line. `Unicode` will be included in the docs, but the documentation for this module will be empty.

src/unicode/unicode.cr
```diff
@@ -1,13 +1,12 @@
...
 #
-# There's no need to use the methods defined in this module
+# There's no need to use the methods defined here
 # because they are exposed in `Char` in a more convenient way
 # (`Char#upcase`, `Char#downcase`, `Char#whitespace?`, etc.)
+
 module Unicode
-  # Options to pass to `upcase`, `downcase`, `uppercase?`
-  # and `lowercase?` to control their behaviour.
+  # Case options to pass to various char and string methods such as `upcase` or `downcase`.
```
Sija:
```suggestion
  # Case options to pass to various `Char` and `String` methods such as `upcase` or `downcase`.
```

src/unicode/unicode.cr
```diff
@@ -1,13 +1,12 @@
-# Provides methods that answer questions about unicode characters,
-# and the `Unicode::CaseOptions` enum for special case conversions
-# like Turkic.
+# The `Unicode` module provides methods that answer questions about unicode characters,
+# and the `Unicode::CaseOptions` enum for special case conversions like Turkic.
 #
-# There's no need to use the methods defined in this module
+# There's no need to use the methods defined here
```
asterite: I think this line here is incorrect. It says "there's no need to use the methods in this module"... but all the methods are `:nodoc:`! So which methods is this comment talking about?

I'd simply hide this comment and say that the `Unicode` module provides just the `Unicode::CaseOptions` enum.

bcardiff: @r00ster91 the comment above still applies. It's not a priority but in order to merge this PR, those need to be addressed.

r00ster91: Oh, I must've missed this one.

I left this in because this comment will not show up in the API docs. This whole comment will only be visible to people visiting this file on GitHub. And for those these methods are visible. However it doesn't make sense to say that these nodoc methods shouldn't be used only here in this file. So I will remove this section of this comment and add a public comment saying that it provides the `CaseOptions` enum.

src/unicode/unicode.cr
```diff
@@ -1,13 +1,6 @@
-# Provides methods that answer questions about unicode characters,
-# and the `Unicode::CaseOptions` enum for special case conversions
-# like Turkic.
```
Sija: I'd just leave this part and remove the rest.

r00ster91: This is included in the public comment now.

Sija: Got misled by the diff, right.

Blacksmoke16:
@r00ster91 Am I missing something or is this still not actually visible in the online API docs?  

Specifically on https://crystal-lang.org/api/0.28.0/String.html#upcase%28options%3DUnicode%3A%3ACaseOptions%3A%3ANone%29-instance-method, the arg type isnt linked nor is `Unicode` in the sidebar.

asterite:
https://github.com/crystal-lang/crystal/blob/db9cb72763af14cfa5d950d4e5903f6ac17fc04f/src/unicode/data.cr#L7

asterite:
I guess nobody tried running `makes docs` to see if the docs appeared?

Blacksmoke16:
Apparently not, that would do it.

Sija:
Seems like `:nodoc:` from `unicode/data.cr` has a priority over https://github.com/crystal-lang/crystal/blob/db9cb72763af14cfa5d950d4e5903f6ac17fc04f/src/unicode/unicode.cr#L1-L2

PR #7518, opened 2019-03-07
Number: remove type suffix from inspect. Add `#dump` method to show it.

### LINKED ISSUE #6196: [RFC] Remove suffixes of integers in inspect
It's so nice now that with Crystal 0.25.0 one can do:

```crystal
a = [1, 2, 3] of Int64
a << 4
p a
```

However, the output is:

```
[1_i64, 2_i64, 3_i64, 4_i64]
```

That's **unreadable**. The numbers gets mixed with the `64` from the suffix. Imagine debugging some code related to math/numbers, it's very hard to read. The code already states that the integer is `Int64`. Why does it need to be explicitly stated in the output? It's just noise, and very annoying noise.

I propose we revert this change. Otherwise to print such code.

Note that even using `puts a` will print it this way, because `Array#to_s` uses `inspect` for each element (like in Ruby, so that's actually fine/good, specially if the Array has objects).

This is further a problem if such array is part of an object:

```crystal
class Foo
  property a = [1, 2, 3] of Int64
end

p Foo.new
```

Output:

```
#<Foo:0x10e4a2f20 @a=[1_i64, 2_i64, 3_i64]>
```

So much noise...

Can we please revert this change? :-)

Thank you ❤️

asterite:
Fixes #6196

I added `dump` because many seem to want it. I'll personally never use it (we can use `typeof`) so if there isn't a strong reason to want it we can remove it.

### DIFF
--- a/src/number.cr
+++ b/src/number.cr
@@ -268,6 +268,35 @@ struct Number
     self == 0
   end
 
+  # Returns a string representation of this number including a suffix
+  # that shows it's type. The suffix is not shown for `Int32` and `Float64`
+  # because they are the default integer and float types respectively, in Crystal.
+  #
+  # ```
+  # 1.dump     # => "1"
+  # 1_i64.dump # => "1_i64"
+  # 1_u8.dump  # => "1_u8"
+  # ```
+  def dump : String
+    String.build do |io|
+      dump(io)
+    end
+  end
+
+  # Appends a string representation of this number including a suffix
+  # that shows it's type. The suffix is not shown for `Int32` and `Float64`
+  # because they are the default integer and float types respectively, in Crystal.
+  #
+  # See `#dump`.
+  def dump(io : IO)
+    to_s(io)
+    io << dump_suffix
+  end
+
+  # Returns the suffix used for `#dump`.
+  # Can return an empty string if there's no suffix.
+  abstract def dump_suffix : String
+
   private class StepIterator(T, L, B)
     include Iterator(T)
 
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -394,4 +394,12 @@ describe BigDecimal do
     negative_one.normalize_quotient(negative_one, positive_ten).should eq(positive_ten)
     negative_one.normalize_quotient(negative_one, negative_ten).should eq(negative_ten)
   end
+
+  describe "#dump" do
+    it { "123".to_big_d.dump.should eq("123_big_d") }
+  end
+
+  describe "#inspect" do
+    it { "123".to_big_d.inspect.should eq("123") }
+  end
 end
--- a/spec/std/big/big_float_spec.cr
+++ b/spec/std/big/big_float_spec.cr
@@ -187,8 +187,12 @@ describe "BigFloat" do
     it { "1234567890123456789".to_big_f.to_s.should eq("1234567890123456789") }
   end
 
+  describe "#dump" do
+    it { "2.3".to_big_f.dump.should eq("2.3_big_f") }
+  end
+
   describe "#inspect" do
-    it { "2.3".to_big_f.inspect.should eq("2.3_big_f") }
+    it { "2.3".to_big_f.inspect.should eq("2.3") }
   end
 
   it "#hash" do
--- a/spec/std/big/big_int_spec.cr
+++ b/spec/std/big/big_int_spec.cr
@@ -259,8 +259,12 @@ describe "BigInt" do
     a.to_big_f.should eq(BigFloat.new("1234567890123456789.0"))
   end
 
+  describe "#dump" do
+    it { "2".to_big_i.dump.should eq("2_big_i") }
+  end
+
   describe "#inspect" do
-    it { "2".to_big_i.inspect.should eq("2_big_i") }
+    it { "2".to_big_i.inspect.should eq("2") }
   end
 
   it "does gcd and lcm" do
--- a/spec/std/big/big_rational_spec.cr
+++ b/spec/std/big/big_rational_spec.cr
@@ -192,6 +192,14 @@ describe BigRational do
     x = br(10, 3)
     x.clone.should eq(x)
   end
+
+  describe "#dump" do
+    it { 123.to_big_r.dump.should eq("123_big_r") }
+  end
+
+  describe "#inspect" do
+    it { 123.to_big_r.inspect.should eq("123") }
+  end
 end
 
 describe "BigRational Math" do
--- a/spec/std/float_spec.cr
+++ b/spec/std/float_spec.cr
@@ -204,13 +204,33 @@ describe "Float" do
     end
   end
 
+  describe "#dump" do
+    it "does dump for f64" do
+      3.2.dump.should eq("3.2")
+    end
+
+    it "does dump for f32" do
+      3.2_f32.dump.should eq("3.2_f32")
+    end
+
+    it "does dump for f64 with IO" do
+      str = String.build { |io| 3.2.dump(io) }
+      str.should eq("3.2")
+    end
+
+    it "does dump for f32" do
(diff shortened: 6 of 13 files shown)


src/number.cr
```diff
@@ -268,6 +268,35 @@ struct Number
     self == 0
   end
 
+  # Returns a string representation of this number including a suffix
+  # that shows it's type. The suffix is not shown for `Int32` and `Float64`
```
r00ster91:
```suggestion
  # that shows its type. The suffix is not shown for `Int32` and `Float64`
```

src/number.cr
```diff
@@ -268,6 +268,35 @@ struct Number
...
+  # 1_u8.dump  # => "1_u8"
+  # ```
+  def dump : String
+    String.build do |io|
+      dump(io)
+    end
+  end
+
+  # Appends a string representation of this number including a suffix
+  # that shows it's type. The suffix is not shown for `Int32` and `Float64`
```
r00ster91:
```suggestion
  # that shows its type. The suffix is not shown for `Int32` and `Float64`
```

Sija:
> The suffix is not shown for `Int32` and `Float64`

Maybe it should? If we're talking about `#dump` I'd expect to see **all** types there.

src/number.cr
```diff
@@ -268,6 +268,35 @@ struct Number
...
+  # because they are the default integer and float types respectively, in Crystal.
+  #
+  # See `#dump`.
+  def dump(io : IO)
```
bcardiff: I would expect to `dump` emit something that I can copy/paste in crystal code and runs.
That is not true for the Big* family that does not have a built-in literal.
What about adding a `dump_prefix` (or change the `#dump` implementation in Big*) so they can be emitted as `BigInt.new("` + `.to_s` + `")` ?

Sija: My expectations differ in this regard, I'd expect to clearly see the types (see PHP's [`var_dump`](http://php.net/manual/en/function.var-dump.php)), not necessarily be able to copy-paste it - I'd rather use `#inspect` for that.

bew: > I'd rather use `#inspect` for that

But you don't always have the correct types then, since there is no type suffix on `#inspect`

Sija: @bew Oh, right. I guess I don't copy-paste too much `#inspect`-ed output... 😅

straight-shoota: Well, currently you can copy&paste `inspect` output of numbers. Only when this is merged, you can't.

Sija: The whole argument about copy pasting is for me pretty weak since how often you actually need to do so? And such behavior would IMO rather fit into `#serialize/deserialize` terms and not `#inspect` or `#dump` - these, as I understand them, are provided for debugging purposes and mainly to be used for visual grepping, not copy-paste.

ysbaddaden: > how often you actually need to do so?
Never for me.

asterite: > I would expect to dump emit something that I can copy/paste in crystal code and runs.
That is not true for the Big* family that does not have a built-in literal.

And this is why I think we shouldn't provide a `dump` method at all, because it will always be incomplete.

asterite:
My proposal: let's remove `dump` for now. We can think of better ways to generally dump objects and try to get a representation that's as close as Crystal code as possible (this isn't always possible). Right now `_big_i` and such don't work well so even though `1_i8` works, `1_big_i` doesn't and it's inconsistent.

Thoughts?

RX14:
I still prefer challenging the assumption that `["foo"].to_s == "[\"foo\"]"` for closing #6196, but I'd rather remove the suffixes than add a `#dump`.

straight-shoota:
I'd prefer to keep type suffixes and solve the issue of noisy output differently. It mostly affects collections like an array, where you have multiple numbers and don't want to be overwhelmed by the type suffixes (often all are of the same type anyway). When inspecting a single number, I don't think it hurts that much to print the type suffix.

There have been two alternative proposals and I'd prefer either of them over removing type suffixes entirely and `#dump`.

* When printing `Array(T)` (or similar collection) and `T` is a number type, call `#to_s` instead of `#inspect` on each member (https://github.com/crystal-lang/crystal/issues/6196#issuecomment-397645983). That's essentially what we're doing in `Slice#to_s` right now, when `T` is `UInt`. This would remove the most annoying clutter, but you still get the type when inspecting a number. That's what I would expect for `Number#inspect` and if you don't want it, there is `#to_s`.
* Move type suffixes from `#inspect` to `#pretty_print` (https://github.com/crystal-lang/crystal/issues/6615#issuecomment-416740878).

Since this is a new edition of #6616 I guess we can close that then.
If we want merge this one, the change to `Slice#to_s` should be brought over from there.

asterite:
> It mostly affects collections like an array, where you have multiple numbers and don't want to be overwhelmed by the type suffixes 

I disagree. It affects **everything**. When you do `p some_type` and `some_type` is a class and you get to see the value of instance variables, and these are numbers, or arrays of numbers, you get the suffixes all over the place.

asterite:
> Move type suffixes from #inspect to #pretty_print

That wouldn't be good because suffixes are ugly, not pretty.

straight-shoota:
> When you do p some_type and some_type is a class and you get to see the value of instance variables, and these are numbers, or arrays of numbers, you get the suffixes all over the place.

With both alternatives, there would be no type suffixes for arrays of numbers, even if they're values of instance variables. And how numerical instance variables are printed depends on the individual type's implementation of `#inspect`. If the number type is not important, you should simply use ´#to_s`. The auto-generated `#inspect` methods in `Struct` and `Reference` could use the same logic to print number ivars with `#to_s`.

RX14:
> you get the suffixes all over the place

and this is what I want when i'm debugging

asterite:
> and this is what I want when i'm debugging

If you are debugging you can do `p typeof(something)`. There's no need to infect every output of every program with debug info...

straight-shoota:
> There's no need to infect every output of every program with debug info...

I don't follow this argument... I always thought the purpose of `#inspect` is to provide more information on an object for purposes such as debugging. Not something for regular user-facing output. Who would be interested in object ids and stuff anyway? So every program simply shouldn't use `#inspect` for every output.

asterite:
Array#to_s invokes inspect. Ruby got it right. We had it right. Why can't we revert that change?

straight-shoota:
> Array#to_s invokes inspect.

And we could change that for numbers, as proposed in alternative solution number 1.
But who would want to use `Array#to_s` (including square brackets) for user-facing output anyway?  `array.join(", ")` is the popular choice. Or something even more fancy.

asterite:
Actually, I was wrong. I do use `inspect` and related for debugging and it bugs me to see all those number prefixes.

I'll close this, it seems you all want to keep somehow this behavior. Until this goes back to how it was originally (like in Ruby and every other language in the universe) I'm not interested in developing or using Crystal anymore. Sorry ¯\_(ツ)_/¯

j8r:
If we want a Pretty Print, we do `pp` – sounds logical. I think this change fits for improving the Pretty Print readability.

sdogruyol:
IMHO, this should be the default behavior like it used to be and every other practical language on the planet. I don't get all the excessive bikeshedding going over the types and I'm really sad to see the impulsiveness involved in the comments. 

The only concern should be making Crystal move forward for the greater good.

Sija:
AFAIR introducing this change in #4747 was done just for theoretical reasons based on assumptions (_it's useful to know the type_) with no real-world backing. There was no bug report opened and no one wished to have them in the language before that PR. We all know the aftermath, which polarized developers over the topic, which was purely theoretical from the beginning.

IMHO the only sensible thing to do is to revert it entirely and open an issue, in which interested parties can discuss if, and how they'd like to implement such behavior as an opt-in, rather than the other way around - which stands in obvious conflict with many devs and other languages (even ones with strong types).

PR #7520, opened 2019-03-08
Implement Int#{leading,trailing}_zeros_count

### LINKED ISSUE #7517: Missing 'count the number of leading/trailing zeroes' on Int and BigInt
Modern compilers have a set of **builtin compiler functions** that can be availed of to optimize code (speed and code size).

Nim does a good job of this for some of them, as shown at the following links.

https://github.com/nim-lang/Nim/blob/devel/lib/pure/bitops.nim
https://github.com/nim-lang/Nim/blob/master/lib/pure/math.nim#L103

It would nice for Crystal to provide these same type of optimizations for standard methods that use them.

I have an immediate use for **countOfTrailingZeroes** and **popcount**.

Sija:
Closes #7517

Missing `BigInt#leading_zeros_count` implementation, which I have no idea how to tackle since `LibGMP` uses variable width integers.

### DIFF
--- a/src/big/big_int.cr
+++ b/src/big/big_int.cr
@@ -385,6 +385,10 @@ struct BigInt < Int
     LibGMP.popcount(self)
   end
 
+  def trailing_zeros_count
+    LibGMP.scan1(self, 0)
+  end
+
   def to_i
     to_i32
   end
--- a/src/int.cr
+++ b/src/int.cr
@@ -542,6 +542,9 @@ struct Int
   # ```
   abstract def popcount
 
+  # Returns the number of trailing `0`-bits.
+  abstract def trailing_zeros_count
+
   private class TimesIterator(T)
     include Iterator(T)
 
@@ -629,6 +632,15 @@ struct Int8
     Intrinsics.popcount8(self)
   end
 
+  # Returns the number of leading `0`-bits.
+  def leading_zeros_count
+    Intrinsics.countleading8(self, false)
+  end
+
+  def trailing_zeros_count
+    Intrinsics.counttrailing8(self, false)
+  end
+
   def clone
     self
   end
@@ -654,6 +666,15 @@ struct Int16
 
   def popcount
     Intrinsics.popcount16(self)
+  end
+
+  # Returns the number of leading `0`-bits.
+  def leading_zeros_count
+    Intrinsics.countleading16(self, false)
+  end
+
+  def trailing_zeros_count
+    Intrinsics.counttrailing16(self, false)
   end
 
   def clone
@@ -683,6 +704,15 @@ struct Int32
     Intrinsics.popcount32(self)
   end
 
+  # Returns the number of leading `0`-bits.
+  def leading_zeros_count
+    Intrinsics.countleading32(self, false)
+  end
+
+  def trailing_zeros_count
+    Intrinsics.counttrailing32(self, false)
+  end
+
   def clone
     self
   end
@@ -708,6 +738,15 @@ struct Int64
 
   def popcount
     Intrinsics.popcount64(self)
+  end
+
+  # Returns the number of leading `0`-bits.
+  def leading_zeros_count
+    Intrinsics.countleading64(self, false)
+  end
+
+  def trailing_zeros_count
+    Intrinsics.counttrailing64(self, false)
   end
 
   def clone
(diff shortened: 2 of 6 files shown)


konovod:
`BigInt#leading_zeros_count` just doesn't make sense, so i think raising an exception is a correct implementation and raising should be reflected in specs.

ysbaddaden:
Or not implement it, so it will fail to compile.

yxhuvud:
This will be useful for me immediately, but I wonder about the naming - wouldn't it make more sense to have it called `count_leading_zeros`? I guess it is a question of if it should be a order or a lookup of a property.

Are there any conventions around this? LLVM intrinsics clearly use the CLZ order.

Sija:
My favorite naming option would be `leading_zeros` / `trailing_zeros`.

/cc @asterite @crystal-lang ?

asterite:
My thought process:
- in LLVM it's called `count_leading_zeros` because it's a function: `count_leading_zeros(value)` which can be read as "count leading zeros (of) value"
- `number.leading_zeros_count` expressed that as an accessor/property
- If I see "001000" and think `leading_zeros` I think about "00" so maybe I'd expect `leading_zeros` to return "00"

But I wouldn't mind ending up with `count_leading_zeros` or just `leading_zeros` if it won't be confusing.

Sija:
Verb at the beginning in `count_..._zeros` reads badly for my taste, I'd prefer using a noun there, as in `..._zeros_count` - if at all.

src/big/big_int.cr
```diff
@@ -390,6 +390,14 @@ struct BigInt < Int
     LibGMP.popcount(self)
   end
 
+  def leading_zeros_count
```
straight-shoota: I'd prefer to simply not declare this method on `BigInt`. It does only apply to fixed-width integers.

asterite: Agreed, this is just repeating compiler's work.

Sija: That's impossible since it's declared as `abstract`.

bew:
I'd suggest `binary_leading_zeros` and `binary_trailing_zeros` to explicit the fact that it checks the binary representation

src/big/big_int.cr
straight-shoota: Yes, it shouldn't be an abstract method on `Int`, because the concept of leading zeros doesn't apply to all integer types.

Sija: @straight-shoota so what do you suggest?

ysbaddaden: Drop the `Int#leading_zeros_count` abstract method.

Sija: > Yes, it shouldn't be an abstract method on `Int`, because the concept of leading zeros doesn't apply to all integer types.

For the time being it applies to all of the `Int` types, besides `BigInt` so I'm more inclined to treat `BigInt` as an exception to the rule then the other way around.

Besides, it's symmetric to `Int#trailing_zeros_count`. What's exactly the problem with raising in compile-time in `BigInt#leading_zeros_count` impl.?

j8r:
That's indeed pedantic, after reading the docs, there are `Indexable#first`, `Indexable#last`, `Hash#first_key`, `Hash#first_value`,  `Hash#last_key`, `Hash#last_value`, `XML::Node#first_element_child`.
Wouldn't it be even more on par with other methods in the stdlib to name it `first_binary_zeros` and `last_binary_zeros`?

konovod:
@j8r it would ne something like `first_binary_one_index`, because we seek for a first 1, not first 0. But that's not needed because "count leading\trailing zeros" is used in other languages and libraries.

src/big/big_int.cr
ysbaddaden: It declares a method that makes no sense, and duplicates what the compiler already does.

ysbaddaden:
The feature is very specific, already known by some name. Giving it another name will only make it more difficult to find when it's needed.

Let's just use `leading_zeros_count`.

src/big/big_int.cr
Sija: Yeah, but having it declared directly on `Int` as `abstract` makes it possible to use it for all `Int` types without duck typing or any other checks.

asterite: It's the other way around: if we ever support holding a variable of type Int, then invoking a method in it will have the compiler type all methods of all subclasses. It will find the one in BigInt and the compilation will fail.

Sija: Oh, I've forgot about that case, I'll remove it then. Is it ok to just comment it out, to keep it more visible?

Sija: Done.

src/int.cr
```diff
@@ -542,6 +542,12 @@ struct Int
   # ```
   abstract def popcount
 
+  # Returns the number of leading `0`-bits.
+  # abstract def leading_zeros_count
```
straight-shoota: I don't think this should be left as a comment either. But please copy the doc comment to all implementations.

Sija: Done as requested.

PR #7525, opened 2019-03-08
Number: let `inspect` not show the type suffix

### LINKED ISSUE #6196: [RFC] Remove suffixes of integers in inspect

asterite:
Fixes #6196 

Also clarifies the difference between `to_s` and `inspect`.

I'll repeat it here: the idea of `inspect` is to give a string representation that's **clearly delimited** and so it's suitable to be embedded inside other expressions. That's why `Array#to_s` invokes `inspect` on each element, because in the case of for example `String`, `to_s` doesn't show the quotes but `inspect` does. `inspect` has **nothing to do** with showing something that resembles a Crystal expression, or something that shows types.

If many of you want a way to make it easier to know the types of an object tree then maybe you can open an issue to discuss it. For example the method could check the constructors of an object to maybe create expressions that recreate an object, thought this isn't always possible... but it's an idea. And it probably belongs in a shard (in my opinion).

Finally, my opinion expressed [here](https://github.com/crystal-lang/crystal/pull/7518#issuecomment-470673749) about me not wanting to continue developing or using Crystal is not a threat, it's just how I feel: why would I use a language that I started which I find ugly and verbose in its output and which doesn't show the clean and beautiful output of Ruby? This is just something I feel super strong about, that's all.

### DIFF
--- a/src/object.cr
+++ b/src/object.cr
@@ -104,10 +104,38 @@ class Object
   # as this will in turn call `to_s(io)` on it.
   abstract def to_s(io : IO)
 
-  # Returns a `String` representation of this object.
+  # Returns a `String` representation of this object suitable
+  # to be embedded inside other expressions, sometimes providing
+  # more information about this object.
   #
-  # Similar to `to_s`, but usually returns more information about
-  # this object.
+  # `#inspect` (and `#inspect(io)`) are the methods used when
+  # you invoke `#to_s` or `#inspect` on an object that holds
+  # other objects and wants to show them. For example when you
+  # invoke `Array#to_s`, `#inspect` will be invoked on each element:
+  #
+  # ```
+  # ary = ["one", "two", "three, etc."]
+  # ary.inspect # => ["one", "two", "three, etc."]
+  # ```
+  #
+  # Note that if Array invoked `#to_s` on each of the elements
+  # above, the output would have been this:
+  #
+  # ```
+  # ary = ["one", "two", "three, etc."]
+  # # If inspect invoked to_s on each element...
+  # ary.inspect # => [one, two, three, etc.]
+  # ```
+  #
+  # Note that it's not clear how many elements the array has,
+  # or which are they, because `#to_s` doesn't guarantee that
+  # the string representation is clearly delimited (in the case
+  # of `String` the quotes are not shown).
+  #
+  # Also note that sometimes the output of `#inspect` will look
+  # like a Crystal expression that will compile, but this isn't
+  # always the case, nor is it necessary. Notably, `Reference#inspect`
+  # and `Struct#inspect` return values that don't compile.
   #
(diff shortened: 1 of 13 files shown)


src/object.cr
```diff
@@ -104,10 +104,38 @@ class Object
...
+  # ```
+  #
+  # Note that it's not clear how many elements the array has,
+  # or which are they, because `#to_s` doesn't guarantee that
+  # the string representation is clearly delimited (in the case
+  # of `String` the quotes are not shown).
+  #
+  # Also note that sometimes the output of `#inspect` will look
+  # like a Crystal expression that will compile, but this isn't
+  # always the case, nor is necessary. Notably, `Reference#inspect`
```
r00ster91:
```suggestion
  # always the case, nor is it necessary. Notably, `Reference#inspect`
```

src/object.cr
```diff
@@ -104,10 +104,38 @@ class Object
...
+  # ary = ["one", "two", "three, etc."]
+  # ary.inspect # => ["one", "two", "three, etc."]
+  # ```
+  #
+  # Note that if Array invoked `#to_s` on each of the elements
+  # above, the output would have been this:
+  #
+  # ```
+  # ary = ["one", "two", "three, etc."]
+  # # If inspect invoked to_s on each element...
```
r00ster91: Already stated above and also not stated in the previous example.
```suggestion
```

asterite: This is if you read the example out of context, to avoid confusion.

src/object.cr
asterite: thanks

asterite:
Thank you all! ❤️❤️❤️

ysbaddaden:
Thanks! :clap:

straight-shoota:
I am a bit surprised how fast this was rushed, but I am ok with the change.

PR #7527, opened 2019-03-08
Compiler: merge procs with the same arguments type and Nil | T return type to Nil return type

### LINKED ISSUE #3655: Proc with Void return type
```crystal
class A
  @@a : String -> Void
  @@a = ->(y : String){
    	1
  }
end
```

compiler complains Proc type doesn't match.

asterite:
Fixes #3655

The idea here is that a proc that returns any type `T` is compatible with a proc with the same argument types that returns `Nil`. The union type of those things is just the type of the proc that returns `Nil`, as in "in the general case we don't know anything about the value being returned".

This is a rule that's currently being applied in a few places, this fixes the language to apply it in more (all?) places.

### DIFF
--- a/src/compiler/crystal/semantic/type_merge.cr
+++ b/src/compiler/crystal/semantic/type_merge.cr
@@ -271,12 +271,24 @@ module Crystal
 
   class ProcInstanceType
     def common_ancestor(other : ProcInstanceType)
+      # For Proc(..., NoReturn), Proc(..., T) we keep Proc(..., T)
       if return_type.no_return? && arg_types == other.arg_types
         return other
       end
 
+      # Same but the other way around
       if other.return_type.no_return? && arg_types == other.arg_types
         return self
+      end
+
+      # For Proc(..., Nil), Proc(..., T) we keep Proc(..., Nil)
+      if return_type.nil_type? && arg_types == other.arg_types
+        return self
+      end
+
+      # Same but the other way around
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/semantic/type_merge.cr
```diff
@@ -271,14 +271,26 @@ module Crystal
...
         return other
       end
 
+      # Same but the other way around
       if other.return_type.no_return? && arg_types == other.arg_types
         return self
       end
 
+      # For Proc(..., Nil), Proc(..., T) we keep Proc(..., Nil)
+      if return_type.nil_type? && arg_types == other.arg_types
```
straight-shoota: This entire method could return early if `arg_types != other.arg_types`.

asterite: Yes, but doing that is expensive if the return types are other.

paulcsmith:
Incredible! That was so fast. Thank @asterite

asterite:
@paulcsmith Glad to help! 😊 

It would be great if you can try it out in the code you have, if you are familiar with compiling the compiler (or maybe now there are nightlies that can make this easier).

Blacksmoke16:
@asterite This PR ended up breaking Athena.  (The commit before this was merged works fine).  I'll see if i can come up with a reduced example of what is breaking.

PR #7529, opened 2019-03-08
Compiler: fix named args type matching

### LINKED ISSUE #7523: Using keyword arg with Proc restriction changes the type of the Proc
Crystal 0.27.2

```crystal
class ProcTest
  def self.run_it(proc : ->)
    proc.call
  end
end

# Works!
ProcTest.run_it(->{ 123 })

# Does not work
ProcTest.run_it(proc: ->{ 123 })

# Raises
Error in line 7: no overload matches 'ProcTest.run_it', proc: Proc(Int32)
Overloads are:
 - ProcTest.run_it(proc : (-> ))
Couldn't find overloads for these types:
 - ProcTest.run_it(proc : Proc(Int32))
```

So using a keyword arg seems to change the type of the proc, which is quite strange and makes it very hard to use proc as instance variables in initializers when you want to use a keyword arg to make it clear which Proc you are trying to use

asterite:
Fixes #7523

The order of matching named arguments type was wrong (match vs. signature but should have been the other way around).

### DIFF
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -906,4 +906,14 @@ describe "Semantic: proc" do
       Foo.new.block
       )) { proc_of(nil_type) }
   end
+
+  it "can pass proc that returns T as Void with named args (#7523)" do
+    assert_type(%(
+      def foo(proc : ->)
+        proc
+      end
+
+      foo(proc: ->{ 1 })
+      )) { proc_of(nil_type) }
+  end
 end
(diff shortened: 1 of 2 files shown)


asterite:
Yeah, it should probably be named "can be cast to", but in general the whole matching logic should be refactored. This is just a small patch.

paulcsmith:
Awesome! Can't wait to try this out on some stuff I'm working on. I just noticed there is a highly docker build so I'll give that a try later. Thanks for fixing this!

PR #7531, opened 2019-03-09
Change bsearch's `/ 2` to `>> 1` for faster performance

Fryguy:
Using the benchmark:

```crystal
require "benchmark"

RANGE = (0..100_000_000)
ARRAY = RANGE.to_a

Benchmark.ips do |x|
  x.report("Array#bsearch") { ARRAY.bsearch { |number| number > 77_777_777 } }
  x.report("Range#bsearch") { RANGE.bsearch { |number| number > 77_777_777 } }
end
```

Before:
```
Array#bsearch   6.83M (146.38ns) (± 2.92%)  0 B/op        fastest
Range#bsearch   6.11M ( 163.7ns) (± 2.31%)  0 B/op   1.12× slower
```

After:
```
Array#bsearch   13.1M ( 76.35ns) (± 3.12%)  0 B/op        fastest
Range#bsearch   9.88M (101.23ns) (± 3.00%)  0 B/op   1.33× slower
```

---

I was surprised to see how much slower bsearch was than find (originally noticed in @konung's [fast-crystal](https://github.com/konung/fast-crystal#arraybsearch-vs-arrayfind-code)), so I'm looking into better algorithms.  Until I find one that works well, I figured I'd submit this simple win.  It uses the less expensive shift-right operation to do integer division by 2 as opposed to the division operation.  This change provides a 48% speedup in Array#bsearch and a 38% speedup in Range#bsearch

### DIFF
--- a/src/range/bsearch.cr
+++ b/src/range/bsearch.cr
@@ -48,8 +48,8 @@
     saved_to = to
     satisfied = nil
     while from < to
-      mid = (from < 0) == (to < 0) ? from + (to - from) / 2
-          : (from < -to) ? -((- from - to - 1) / 2 + 1) : (from + to) / 2
+      mid = (from < 0) == (to < 0) ? from + ((to - from) >> 1)
+          : (from < -to) ? -(((- from - to - 1) >> 1) + 1) : ((from + to) >> 1)
 
       if yield mid
         satisfied = mid
@@ -102,7 +102,7 @@ struct Range(B, E)
     saved_to = to
     satisfied = nil
     while from < to
-      mid = from + (to - from) / 2
+      mid = from + ((to - from) >> 1)
 
       if yield mid
         satisfied = mid


asterite:
The reason is that division checks for zero and might raise. That makes it slower. I think there's also unsafe_div. But this is fine.

Fryguy:
I had originally tried unsafe_div and it was slower than `>> 1`.  I had also tried unsafe_shr, but there was no change over `>> 1`

asterite:
Actually /2 shouldn't be slower because it's a hardcoded value. This is strange...

Fryguy:
I was actually surprised that LLVM didn't have a built-in optimization for `/ 2`

Fryguy:
FWIW:

/ 2
```
Array#bsearch   6.83M (146.38ns) (± 2.92%)  0 B/op        fastest
Range#bsearch   6.11M ( 163.7ns) (± 2.31%)  0 B/op   1.12× slower
```

\>> 1
```
Array#bsearch   13.1M ( 76.35ns) (± 3.12%)  0 B/op        fastest
Range#bsearch   9.88M (101.23ns) (± 3.00%)  0 B/op   1.33× slower
```

unsafe_div(2)
```
Array#bsearch   9.09M (109.97ns) (± 2.65%)  0 B/op        fastest
Range#bsearch   7.35M (136.13ns) (± 3.28%)  0 B/op   1.24× slower
```

unsafe_shr(1)
```
Array#bsearch  13.06M ( 76.54ns) (± 3.66%)  0 B/op        fastest
Range#bsearch   9.92M (100.84ns) (± 3.46%)  0 B/op   1.32× slower
```

konovod:
Signed division could be slower because it rounds up to zero (at least as [LLVM docs](https://llvm.org/docs/LangRef.html#sdiv-instruction) say), and shift rounds to negative infinity. Benchmarking confirms that unsigned division has same speed and signed is 1.23 times slower. 

On the other hand looks like signed division in Crystal always rounds to negative infinity - i have no idea how to explain it.

Fryguy:
Just tried being more explicit by doing an unsigned div with `/ 2_u32` and `/2_u8`.  No change from regular `/ 2`.  Updated my post above.

konovod:
The left side is still signed. Can't currently check myself, but I'm sure that if you will convert left side to UInt32 (unsafely) and then do unsafe_div, performance of bsearch will match the one with `>> 1`.

Fryguy:
@konovod Doesn't seem to have worked, but it's possible I did it wrong:

```
Array#bsearch   6.91M (144.62ns) (± 2.82%)  0 B/op        fastest
Range#bsearch   6.16M (162.37ns) (± 3.25%)  0 B/op   1.12× slower
```

Is this the "right" way to do it?

```diff
-      mid = from + ((to - from) >> 1)
+      mid = from + ((to - from).unsafe_as(Int32) / 2_u32)
```

konovod:
unsafe_as(UInt32)

Fryguy:
LOL oops..of course :smile:

So, yes, using `.unsafe_as(Int32).unsafe_div(2_u32)` has similar performance to `>> 1`

```
Array#bsearch  13.19M ( 75.81ns) (± 3.50%)  0 B/op        fastest
Range#bsearch   9.92M (100.76ns) (± 3.81%)  0 B/op   1.33× slower
```

konovod:
I think that in this (and similar) places, `>> 1` is easier to understand and looks better than converting to unsigned.
For a general case (unsigned division by, say, 3 is faster then signed too), maybe there should be something like `unsafe_positive_div` but this looks like a low priority and debatable thing.

straight-shoota:
Thank you @Fryguy

PR #7536, opened 2019-03-10
Fix restriction of valid type vs free vars

### LINKED ISSUE #6446: cannot overload forall method
```crystal

module A
  def self.func(c : Array(String).class)
    puts "xxxx"
  end
  def self.func(c : T.class) forall T
    puts "yyyy"
  end
  def self.func(c : Int32.class)
    puts "zzzz"
  end
end

module A
  def self.func(c : Array(Int32).class)
    puts "override"
  end
end
A.func Array(String) # xxxx
A.func Array(Int32) # yyyy, should be override
A.func Int32 # yyyy should be zzzz
```

bew:
Fixes #6446

---
```cr
def bar(a : T.class) forall T
  puts "Hello from T"
end
  
def bar(a : Int32.class)
  puts "Hello from Int32"
end

# Before:
bar(Int32) # "Hello from T"

# After
bar(Int32) # "Hello from Int32"
```
Compiler hacking is fun :tada:


asterite:
I'll review it tomorrow.

bew:
Thanks for the review :smiley:

Just a note: https://carc.in/#/r/6hcr
```cr
# No free vars here, and Foo doesn't exist
def bar(a : Foo.class)
  puts "Hello from Foo"
end
  
def bar(a : Int32.class)
  puts "Hello from Int32"
end

# Before:
bar(Int32) # Error: undefined constant Foo

# After
bar(Int32) # "Hello from Int32"
```

asterite:
@bew Yes, I'm not sure. I think it's bad that if you don't invoke the method you don't get the "undefined constant Foo" either way. We should probably have a pass to compute those types before analyzing calls, and then in the restriction you would know whether a type is a free var or not. For now maybe this is fine.

bew:
I think the new behavior is fine.

I'm not sure how you could fix that, since you could have:
```cr
def bar(a : Foo.class)
  puts "Hello from Foo"
end
  
def bar(a : Int32.class) # restriction happens here when this method is processed
  puts "Hello from Int32"
end

record Foo # define Foo here!

bar(Int32) # => "Hello from Foo"
```
But maybe if there's first a pass to find all types, then a pass to process al methods that could do it, but there will be other problems I think, like macros that creates types / create methods / use `@type.methods` etc..

asterite:
@bew Macros run on the very first pass. This second pass would check the types of all method arguments to see if they exist (they must either be in scope or declared as free variables). I'm almost sure it's something doable. But it's tricky because methods are also added on the first pass (though they are not type-checked there, and this is why it's tricky to compute which one is stricter than the other). We should probably also have some context of `is_restriction?` to know whether something is a free var.

But let's merge this for now.

asterite:
@bew Macros run on the very first pass. This second pass would check the types of all method arguments to see if they exist (they must either be in scope or declared as free variables). I'm almost sure it's something doable. But it's tricky because methods are also added on the first pass (though they are not type-checked there, and this is why it's tricky to compute which one is stricter than the other). We should probably also have some context of `is_restriction?` to know whether something is a free var.

But let's merge this for now.

bew:
> We should probably also have some context of `is_restriction?` to know whether something is a free var.

Yes that's what I was thinking, for exemple the `owner` param could be a `ctx` param instead, with an `owner` field, but also a `def` field (usually nil) representing the current def.

bew:
Actually I've found a much simpler solution while playing with #7543:
```cr
class Metaclass
  def restriction_of?(other : Metaclass, ctx)
    self.name.restriction_of?(other.name, ctx)
  end
end
```
I'll send a PR soon

NOTE: #7543 is sooo going to solve many restriction issues with free vars :rocket:

PR #7537, opened 2019-03-11
Fix GenericClassType vs GenericClassInstanceType restrictions

bew:
I can't find an issue this may have solved (hard to search!), but basically:

```cr
class Foo(T)
end

def bar(a : Foo)
  1
end

def bar(a : Foo(Int32))
  true
end

# Before:
bar(Foo(Int32).new) # => 1

# After
bar(Foo(Int32).new) # => true
```
:tada:

### DIFF
--- a/src/compiler/crystal/semantic/restrictions.cr
+++ b/src/compiler/crystal/semantic/restrictions.cr
@@ -281,10 +281,10 @@ module Crystal
 
   class Generic
     def restriction_of?(other : Path, owner)
-      other_type = owner.lookup_type?(self)
-      if other_type
-        self_type = owner.lookup_path(other)
-        if self_type
+      self_type = owner.lookup_type?(self)
+      if self_type
+        other_type = owner.lookup_path(other)
+        if other_type
           return self_type.restriction_of?(other_type, owner)
         end
       end
@@ -300,6 +300,44 @@ module Crystal
         return false unless type_var.restriction_of?(other_type_var, owner)
       end
 
+      true
+    end
+  end
+
+  class GenericClassType
+    def restriction_of?(other : GenericClassInstanceType, owner)
+      # ```
+      # def foo(param : Array)
+      # end
+      #
+      # def foo(param : Array(Int32))
+      # end
+      # ```
+      #
+      # Here, self is `Array`, other is `Array(Int32)`
+
+      return unless self == other.generic_type
+
+      # `SomeGeneric` is never a restriction of `SomeGeneric(X)`
+      false
+    end
+  end
+
+  class GenericClassInstanceType
+    def restriction_of?(other : GenericClassType, owner)
+      # ```
+      # def foo(param : Array(Int32))
+      # end
+      #
+      # def foo(param : Array)
+      # end
+      # ```
+      #
+      # Here, self is `Array(Int32)`, other is `Array`
+
+      return unless self.generic_type == other
+
+      # `SomeGeneric(X)` is always a restriction of `SomeGeneric`
       true
     end
   end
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/semantic/restrictions.cr
```diff
@@ -304,6 +304,44 @@ module Crystal
...
+      # def foo(param : Array)
+      # end
+      #
+      # def foo(param : Array(Int32))
+      # end
+      # ```
+      #
+      # Here, self is `Array`, other is `Array(Int32)`
+
+      return unless self == other.generic_type
```
Sija:
```suggestion
      return false unless self == other.generic_type
```

src/compiler/crystal/semantic/restrictions.cr
```diff
@@ -304,6 +304,44 @@ module Crystal
...
+      # def foo(param : Array(Int32))
+      # end
+      #
+      # def foo(param : Array)
+      # end
+      # ```
+      #
+      # Here, self is `Array(Int32)`, other is `Array`
+
+      return unless self.generic_type == other
```
Sija:
```suggestion
      return false unless self.generic_type == other
```

src/compiler/crystal/semantic/restrictions.cr
Sija: It seems that there's no way that this method will return `true` anyways, right?

src/compiler/crystal/semantic/restrictions.cr
```diff
@@ -281,10 +281,10 @@ module Crystal
...
   class Generic
     def restriction_of?(other : Path, owner)
-      other_type = owner.lookup_type?(self)
-      if other_type
-        self_type = owner.lookup_path(other)
-        if self_type
+      self_type = owner.lookup_type?(self)
+      if self_type
+        other_type = owner.lookup_path(other)
+        if other_type
```
bew: @asterite you'd have to explain to my boss why I didn't slept last night :joy: :innocent:

src/compiler/crystal/semantic/restrictions.cr
bew: Indeed, but I'd like to keep the check explicit, otherwise the explanation below (``# `SomeGeneric` is never a restriction of `SomeGeneric(X)` ``) is not :100: % accurate, as it could become ``# `Foo` is never a restriction of `Bar(X)` `` which makes no sense.

src/compiler/crystal/semantic/restrictions.cr
bew: Hum no, that's wrong.. it'd mean that `Foo` is a restriction of `Foo(Int32)` (when the generic type `Foo` is equal)

src/compiler/crystal/semantic/restrictions.cr
Sija: So better change the docs instead of hardcoding redundant condition.

src/compiler/crystal/semantic/restrictions.cr
Sija: IIUC it's a simplification of your conditions within this method.

bew: Oh right, I was thinking about the other one ><

I'll change that!

asterite:
I'll review it tomorrow.

src/compiler/crystal/semantic/restrictions.cr
asterite: Wow! How did that even work? 😮

straight-shoota:
@asterite Your announcement to review tomorrow would have been almost true in CET, just off by 2 minutes... 😆

bew:
Cool, thanks for the review and merge :smiley:

asterite:
@straight-shoota Haha, yes. I commented that but then realized I had time to do it then... but I figured maybe what I said is true in _some_ timezone out there :-P

Issue #7032, opened 2018-11-06
Stack overflow for BigInt


proyb6:
Testing an old code from someone gist, an input of 300000 (smaller number like 30000 is working fine) get an invalid memory error?

```
require "big"
  
def fact(n)
  return 1 if n == 0
  n * fact(n - 1)
end

n = BigInt.new(ARGV[0])
puts fact(n)
```
> crystal --release --no-debug fact.cr
> ./fact 300000

0.27.0
```
Stack overflow (e.g., infinite or very deep recursion)
[0x41c416] ???
[0x41c121] __crystal_sigfault_handler +433
[0x41c530] sigfault_handler +40
[0x7f1688534390] ???
[0x7f168875f59d] GC_malloc +13
[0x7f1688757708] GC_realloc +184
[0x7f1688c4335a] __gmpz_realloc +58
[0x7f1688c454a1] __gmpz_sub_ui +1041
[0x41c1b1] ???
[0x41c1c2] ??? (104762 times)
[0x404152] __crystal_main +5874
[0x40680c] main +44
[0x7f1687b19830] __libc_start_main +240
[0x402989] _start +41
[0x0] ???

```

0.26.1
```
Invalid memory access (signal 11) at address 0x7ffc7d9e7d40
[0x424176] ???
[0x423fb0] __crystal_sigfault_handler +32
[0x7fef86777390] ???
[0x42b7de] GC_clear_stack_inner +30
[0x42b7f6] GC_clear_stack_inner +54 (5 times)
[0x42b85e] GC_clear_stack +78
[0x427e93] GC_generic_malloc_many +563
[0x42fcc8] GC_malloc +216
[0x7fef86c0d858] __gmpz_init +24
[0x42401c] ???
[0x424058] ??? (104616 times)
[0x40c2c6] __crystal_main +5110
[0x40fe9c] main +44
[0x7fef85d5c830] __libc_start_main +240
[0x40adf9] _start +41
[0x0] ???

```

bcardiff:
It's the case of a very deep recursion.
You might to code it in an iterative way to workaround that limitation.

```crystal
require "big"
  
def fact(n)
  res = 1
  while n > 1
  	res *= n
    n -= 1
	end
  return res
end

n = BigInt.new(40000)
puts fact(n)
```

PR #7542, opened 2019-03-13
Improve two errors

r00ster91:
This improves the error when you do for example `pointerof(123)` which can be fixed by putting `123` in a variable. That's needed because you can't just modify and operate on the raw value `123`, that doesn't make sense. The value of a variable however can be modified.

And this also makes the stack overflow error slighlty more detailed to make sure there's no confusion. Even after stack overflow detection there was still some confusion about the error: #7032.


src/compiler/crystal/semantic/main_visitor.cr
```diff
@@ -2482,7 +2482,7 @@ module Crystal
 
     def visit(node : PointerOf)
       var = pointerof_var(node)
-      node.exp.raise "can't take address of #{node.exp}" unless var
+      node.exp.raise "can't take address of #{node.exp}, put #{node.exp} in a variable" unless var
```
bew: This might be incorrect, sometime you simply can't take the address of it I think (I don't have an example OTOH though..)

asterite: Yes, I think this suggestion is incorrect. Moving things to a variable will give a result that you wouldn't expect. For example if you do `pointerof(self)` and get the error and you move it to a var, and `self` as a struct, then you get a totally different result.

straight-shoota: If at all, I'd phrase this something like "can't take address of #{node.exp}, pointerof() can only be used with variables"

src/compiler/crystal/syntax/parser.cr
```diff
@@ -5337,7 +5337,7 @@ module Crystal
       next_token_skip_space_or_newline
 
       if @token.keyword?(:self)
-        raise "can't take pointerof(self)", @token.line_number, @token.column_number
+        raise "can't take address of pointerof(self), put self in a variable", @token.line_number, @token.column_number
```
straight-shoota: It's either `address of self` or `pointerof(self)`. `address of pointerof(self)` doesn't make sense. That would be `pointerof(pointerof(self))`.

src/signal.cr
```diff
@@ -327,7 +327,21 @@ fun __crystal_sigfault_handler(sig : LibC::Int, addr : Void*)
   stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)
 
   if stack_top <= addr < Fiber.current.@stack_bottom
-    LibC.dprintf 2, "Stack overflow (e.g., infinite or very deep recursion)\n"
+    LibC.dprintf 2,
```
straight-shoota: An error message should be concise, not explain the general concept of stack overflow errors. This is fine as is.

src/compiler/crystal/semantic/main_visitor.cr
r00ster91: How about `can't take address of #{node.exp}, try putting #{node.exp} in a variable`? In the most cases this works.

src/compiler/crystal/syntax/parser.cr
asterite: I agree with this for runtime errors (though stack overflow is a border case because it usually means you coded something wrong). For compile-time errors I'm starting to really like how Elm shows them.

straight-shoota: @asterite You're referring to https://github.com/crystal-lang/crystal/pull/7542#discussion_r265163050 ?

bew:
Also, it surprises me that `pointerof(self)` (and other pointerof(X) checks) is a parse error instead of a check done later in the processing of the parse tree.

src/compiler/crystal/syntax/parser.cr
asterite: Oops, yes

straight-shoota:
@bew Only `pointerof(self)` is a parser error. `self` can't be a variable and it can't be used in `pointerof`, so its definitely an error at this stage.

src/signal.cr
straight-shoota: Please let's not detail the concept of stack overflow in a runtime error message. While it is likely a code issue, this error message is presented to the *user*, not the *developer* and should not try to explain a developer's tasks to a user.

straight-shoota: Maybe we could consider a more detailed error message for development builds, but there is no clear indicator for that. Maybe `flag?(:release)` would do, but that's not its purpose.

r00ster91: Ok you are right with that point. The user will see this error with code for the language the developer is using and that comes a bit unprofessional too. Like the user could come "yeah, developer I think you have to do that and this to fix this error in my application".
Only compile time errors should be as detailed as this error.

bcardiff:
Isn't it misleading that the message says 

> pointerof only works with variables

but `pointerof(obj.@ivar)` and `pointerof(LibFoo.external_var)` are allowed?

r00ster91:
The error message doesn't specify what kind of variable. It could be an instance variable, a class variable or an external variable. It would be misleading if it would say "local variables".

asterite:
I agree with @bcardiff . It's misleading because at the syntax level `obj.@ivar` and `LibFoo.external_var` don't look like variables, even though they point to variables.

What's wrong with the current error message?

r00ster91:
> What's wrong with the current error message?

I don't think there's something wrong with the error. I think at the syntax level `obj.@ivar` isn't confusing because you can clearly see that it's an instance **variable**. Same for `LibC.variable`. If you know that it's a variable, the error makes sense. The error doesn't specify what kind of variable it needs to be, it can be any variable.

asterite:
People usually read "variable" as "local variable". Ask anyone "what's a variable in Ruby" and they will say "the 'x' in 'x = 1'". They will never say "oh, and variable could also mean instance or class variable".

I think, if it's not broken, we shouldn't fix it.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -5364,7 +5364,7 @@ module Crystal
       next_token_skip_space_or_newline
 
       if @token.keyword?(:self)
-        raise "can't take pointerof(self)", @token.line_number, @token.column_number
+        raise "can't take address of self, pointerof only works with variables", @token.line_number, @token.column_number
```
r00ster91:
```suggestion
        raise "can't take address of self", @token.line_number, @token.column_number
```

src/compiler/crystal/semantic/main_visitor.cr
```diff
@@ -2481,7 +2481,7 @@ module Crystal
         # "undefined local variable or method"
         node.exp.accept self
 
-        node.exp.raise "can't take address of #{node.exp}"
+        node.exp.raise "can't take address of #{node.exp}, pointerof only works with variables"
```
r00ster91:
```suggestion
        node.exp.raise "can't take address of #{node.exp}"
```

spec/compiler/semantic/pointer_spec.cr
```diff
@@ -79,11 +79,11 @@ describe "Semantic: pointer" do
   end
 
   it "pointer of class raises error" do
-    assert_error "pointerof(Int32)", "can't take address of Int32"
+    assert_error "pointerof(Int32)", "can't take address of Int32, pointerof only works with variables"
   end
 
   it "pointer of value error" do
-    assert_error "pointerof(1)", "can't take address of 1"
+    assert_error "pointerof(1)", "can't take address of 1, pointerof only works with variables"
```
r00ster91:
```suggestion
    assert_error "pointerof(1)", "can't take address of 1"
```

spec/compiler/semantic/pointer_spec.cr
```diff
@@ -79,11 +79,11 @@ describe "Semantic: pointer" do
   end
 
   it "pointer of class raises error" do
-    assert_error "pointerof(Int32)", "can't take address of Int32"
+    assert_error "pointerof(Int32)", "can't take address of Int32, pointerof only works with variables"
```
r00ster91:
```suggestion
    assert_error "pointerof(Int32)", "can't take address of Int32"
```

spec/compiler/parser/parser_spec.cr
```diff
@@ -1500,7 +1500,7 @@ module Crystal
     assert_syntax_error "case when .foo? then 1; end"
     assert_syntax_error "macro foo;{%end};end"
     assert_syntax_error "foo {1, 2}", "unexpected token: ,"
-    assert_syntax_error "pointerof(self)", "can't take pointerof(self)"
+    assert_syntax_error "pointerof(self)", "can't take address of self, pointerof only works with variables"
```
r00ster91:
```suggestion
    assert_syntax_error "pointerof(self)", "can't take address of self"
```

r00ster91:
Okay but maybe at least turn `can't take pointerof(self)` into `can't take address of self` for consistency with the other errors then?

Sija:
I vote to close this PR.

r00ster91:
@Sija why? I'm trying to do something out of this PR.

Sija:
@r00ster91 because the actual error message is imo good enough.

PR #7543, opened 2019-03-14
Feature/add context to `restriction_of`

bew:
So I added a context for `restriction_of` But I don't know how useful this can be... And what to do with it..

See commit: https://github.com/crystal-lang/crystal/pull/7543/commits/1328d0b522db7972a8da59ed89e51f14c68fd855 for usage trial
@asterite Any ideas?

/ref https://github.com/crystal-lang/crystal/pull/7536#issuecomment-472430019

### DIFF
--- a/src/compiler/crystal/semantic/restrictions.cr
+++ b/src/compiler/crystal/semantic/restrictions.cr
@@ -364,11 +364,11 @@ module Crystal
   end
 
   class Type
-    def restrict(other : Nil, context)
+    def restrict(other : Nil, match_ctx)
       self
     end
 
-    def restrict(other : Type, context)
+    def restrict(other : Type, match_ctx)
       if self == other
         return self
       end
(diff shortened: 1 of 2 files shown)


asterite:
@bew Thank you!

The existing logic is already a bit convoluted, and this is makes it a bit more so. I think all of this should be rethought and maybe written (almost) from scratch. I mean, the whole "restrictions" logic.

For now I wouldn't worry much about trying to improve the existing logic.

bew:
For me as a novice is compiler hacking/reading I think the current way is pretty neat! How would you do this differently?

asterite:
I think having all the restriction logic split across different types is a bit confusing and hard to follow. 

I thought `restriction_of?` was used for other parts of the code beside sorting defs, but apparently it's only there for that. Maybe your PR is good. I'll try to review it a bit.

src/compiler/crystal/semantic/restrictions.cr
```diff
@@ -32,40 +32,44 @@ require "../types"
...
 
-    def restriction_of?(type : ASTNode, owner)
+    def restriction_of?(type : ASTNode, ctx)
       false
     end
   end
 
   struct DefWithMetadata
-    def restriction_of?(other : DefWithMetadata, owner)
+    def restriction_of?(other : DefWithMetadata, ctx)
```
asterite: I think this was fine before. We can create the context inside this method because we have everything: `self`, `other` and `ctx`. That means not creating the context in `ModuleType#add_def`.

asterite:
I actually think the change looks really good. I got confused because I saw a huge diff, also in the `restrict` methods, but it's just renaming `context` to `match_context`. I guess that's fine.

src/compiler/crystal/semantic/restrictions.cr
bew: I thought about it but thought it would be confusing at call site or while reading the code, but I've found a way using a named arg for `DefWithMetadata#restriction_of?` to make it clear everywhere :smiley:

bew:
> I actually think the change looks really good.

Thank you, but I still don't why I did it... Would it be useful for later?

> I got confused because I saw a huge diff, also in the `restrict` methods, but it's just renaming `context` to `match_context`.

Yeah sorry about that, I wanted to show that the context given to `#restriction_of?` and `#restrict` are different by naming them differently. renaming a `MatchContext` to `match_ctx` is short and easy to understand, but I've yet to come up with a 'short' name for `RestrictionContext`..

asterite:
> Thank you, but I still don't know why I did it... Would it be useful for later?

In a previous PR you changes something to say "if I can't resolve a type that this is less strict than it". We could now change it to say "if the other type is a free variable (you can get that from the context, a def knows the free variable) then this is less strict than it, otherwise use the logic before your PR".

bew:
Ok I think I got it right @asterite

bew:
I have a big question @asterite:
```cr
it "inserts typed Path before untyped Path" do
  assert_type(%(
    def foo(a : T) forall T
      1
    end

    def foo(a : Int32)
      true
    end

    foo(1)
    )) { bool }
end
```
It currently works because impl is:
```cr
class Path
  def restriction_of?(other : Path, ctx)
    return true if self == other

    self_type = ctx.owner.lookup_path(self)
    if self_type
      other_type = ctx.owner.lookup_path(other)
      if other_type
        return self_type.restriction_of?(other_type, ctx)
      else
        return true # >>>> When `other` cannot be typed, it's ALWAYS a restriction
      end
    end

    false
  end
end
```
This means that:
- `Int32` vs `T` (free var): `Int32` wins
- `Int32` vs `UndefinedType`: `Int32` wins

---
But now I can check if `other` is a free var:
```cr
class Path
  def restriction_of?(other : Path, ctx)
    return true if self == other

    self_type = ctx.owner.lookup_path(self)
    if self_type
      if other_type = ctx.owner.lookup_path(other)
        return self_type.restriction_of?(other_type, ctx)
      end

      if ctx.free_var_in_other_def?(other)
        # `other` is a free var!
        return true
      end

      # >>>>>> WHAT TO DO WHEN `other` IS UNKNOWN ?
      true
    end

    false
  end
end
```

When `other` is unknown (untyped and not a free var), if I return `true` (which is correct I think), the behavior will be the same as current. But then what's the point of checking if T is a free var if in the end it's treated the same as if it wasn't ?

bew:
I don't think this change is worth it.. I think we can close it, WDYT?

jhass:
No answer for a year is an answer I guess... :D

PR #7546, opened 2019-03-14
MT GC

bcardiff:
This PR allows the GC to work with threads running different fibers. It does not change the scheduling at all. There is a sample program added that manually schedule fibers to threads.

To test this branch you need a custom build of the bdwgc.

The path to find the `libgc.a` is fixed for now.

To be able to run this you need checkout and build in a specific location:
In your crystal working directory with `mt/gc` branch checked out:

```
$ pushd ..
$ git clone http://github.com/crystal-lang/bdwgc
$ cd bdwgc
$ git checkout release-7_6-crystal
$ git clone https://github.com/ivmai/libatomic_ops
$ cd libatomic_ops
$ git checkout v7.6.10
$ cd ..
$ autoreconf -vif
$ ./configure --disable-debug --disable-dependency-tracking --disable-shared --enable-large-config
$ make 
$ ls .libs/
libcord.a   libcord.la  libcord.lai libgc.a     libgc.la    libgc.lai

$ popd # Go back to the crystal working directory

$ make clean deps

# To run the samples/mt_gc_test.cr (no need to build the compiler)
$ ./bin/crystal samples/mt_gc_test.cr -D preview_mt_gc

# To build the compiler itself
$ make crystal FLAGS="-D preview_mt_gc"
```

* The PR adds a simple flag to check whether a fiber is running or not. This is only used for GC purposes and the GC disable/enable in the context switch ensures the information is only read in a consistent state. Yet that check might be affected by the work in #7407
* One of the extensions of bdwgc `GC_get_stackbottom` is to get the stackbottom of the current thread.  That might not be needed if #6970 evolves successfully.
* The other extension of the bdwgc `GC_switch_to_corutine` and how it is used in the `Scheduler`/`Fiber` are the real game changer here.
* The whole patch to bdwgc can be found at https://github.com/crystal-lang/bdwgc/commit/417756b3b6f00f08b9d92f8dcd51c539402a3403

---

This is based on some of the past efforts of (and present talks with 😄) @ggiraldez and @waj


samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
+# Experiment to stress the GC with multi-threading.
+# The experiment manually allocates multiple times a queue of fibers
+# on a list of worker threads. The main thread does the orchestration.
+# The fibers will grow and shrink it's stack size by doing a recursion.
```
r00ster91:
```suggestion
# The fibers will grow and shrink their stack size by doing recursion.
```

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
...
+
+  def initialize(@fibers : Int32, @threads : Int32)
+    @fibers_reached = Atomic(Int32).new(0)
+    @fiber_dephs = Array(Int32).new(@fibers, 0)
+    @mutex = Thread::Mutex.new
+    @threads_state = :wait
+
+    # Create worker fibers but do not start them yet.
+    # Each fiber will try to reach the `expected_depth` value
+    # by increasing or decreasing it's callstack.
```
r00ster91:
```suggestion
    # by increasing or decreasing its callstack.
```

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
...
+    @fibers.times do |index|
+      @worker_fibers << Fiber.new("f:#{index}") do
+        Context.fiber_run(self, index, 1)
+      end
+    end
+
+    # Create worker threads.
+    # they will perform operations when `@threads_state == :run`
+    # otherwise they will remain in a tighy busy loop.
```
r00ster91:
```suggestion
    # otherwise they will remain in a tight busy loop.
```

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
...
+end
+
+class Context
+  property expected_depth : Int32 = 0
+  @worker_fibers = Array(Fiber).new(0)
+  @pending_fibers = Array(Fiber).new(0)
+
+  def initialize(@fibers : Int32, @threads : Int32)
+    @fibers_reached = Atomic(Int32).new(0)
+    @fiber_dephs = Array(Int32).new(@fibers, 0)
```
r00ster91:
```suggestion
    @fiber_depths = Array(Int32).new(@fibers, 0)
```

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
...
+        fiber_run(context, fiber_index, depth + 1)
+      else
+        context.notifify_depth_reached
+        context.yield
+      end
+    end
+  end
+
+  def set_fiber_depth(index, depth)
+    @fiber_dephs[index] = depth
```
r00ster91:
```suggestion
    @fiber_depths[index] = depth
```

src/crystal/scheduler.cr
```diff
@@ -57,9 +57,15 @@ class Crystal::Scheduler
   end
 
   protected def resume(fiber : Fiber) : Nil
+    GC.disable
```
yxhuvud: I wonder, would it be a good idea to have a block version of `GC.disable` that reenables GC when the block is exited?

src/fiber.cr
```diff
@@ -85,16 +89,18 @@ class Fiber
...
-    @stack_bottom = GC.stack_bottom
+    @stack_bottom = GC.current_thread_stack_bottom
     @name = "main"
 
     @@fibers.push(self)
   end
 
   # :nodoc:
   def run
+    GC.enable
```
yxhuvud: Hmm, why is this necessary? Where is the corresponding `disable`?

vladfaust:
So, the `Thread.new` method has already been available before, but new threads were not GC'ed, right? And if this got merged, then we can create true multi-threaded applications?

src/fiber.cr
asterite: It's on the other comment you left ;-)

yxhuvud: No, that one reenables itself, unless there are some really strange shenanigans going on.

ysbaddaden:
Do you get acceptable performance?

I initially used `GC.disable` and `GC.enable` while running a context switch, but the performance was *awful*. I also tried masking the signal used to stop-the-world, which improved performance to not that horrible with multiple threads (but very bad with a single one), but it was linux only, so not portable. See https://github.com/crystal-lang/crystal/pull/7214#issuecomment-450115098 for benchmarks.

I eventually did what the `single-msqueue` branch did: use a rwlock and performance got surprisingly acceptable (not using ConcurrencyKit but a custom multi-reader single-writer lock). See https://github.com/ysbaddaden/crystal/commit/ab7bfb4431f6eecaa26ef982b813936cd7fca587 and a benchmark comparable to the previous disable/sigmask link (1, 2, or 4 threads switching between 128 fibers):

```
switch[1/128]: crystal: 10000000 yields in 411 ms, 24285940 yields per second
switch[2/128]: crystal: 10000000 yields in 991 ms, 10082371 yields per second
switch[4/128]: crystal: 10000000 yields in 949 ms, 10526425 yields per second
```

I also push thread stacks on GC collect, not on each and every context switch. I never benchmarked, but since there are only a few threads, but I assume it's faster.

src/fiber.cr
asterite: Ah, you are right. I think this one might be off.

ysbaddaden: The world is suspended during the context switch: the scheduler switched context, which means that it called `GC.disable` in the current fiber then jumped to `Fiber#run` of the new fiber, and it will only call `GC.enable` when the current fiber is eventually resumed.

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,165 @@
...
+  def gc_stats
+    __p "GC.stats: #{GC.stats}"
+    __p "Foo.collections: #{Foo.collections}"
+  end
+
+  def create_thread
+    Thread.new do
+      while true
+        case @threads_state
+        when :run
```
j8r: `@threads_state` having only two values, `:run` or `:wait` value, I suggest to change it to a Boolean like `@threads_running` or `@threads_waiting`. Else, if it's planned to have more than two values, it can be an `enum`.

src/crystal/scheduler.cr
bcardiff: Due to how things work, I would prefer to not use a block version. This `GC.disable` is paired with the enable of the `Fiber#run` or the trailing `GC.enable` in `Scheduler#resume` depending on whether the `fiber` argument is new or not.

samples/mt_gc_test.cr
bcardiff: It's just a sample. I like symbols. And it used to have 4 states 🙈

bcardiff:
@ysbaddaden currently just using the std spec, compiler specs and the sample. I didn't experience slowdown but probably because the only MT code checked is the sample (and is unable to run in ST).

The `GC.disable`/`GC.enables` could be wrapped by `flag?(:preview_mt_gc)` if needed to allow checking things in MT before enabling the feature.

I try to avoid calling `GC_switch_to_corutine` (aka `set_thread_stack_bottom`) in the context switch and calling it just in `GC.before_collect` but that didn't work. Some collections can be triggered by the gc after a context switch but did not trigger the before_collect callback and then the stack information in the GC was wrong causing a lovely segfault. Having a `set_thread_stack_bottom(thread, stack_bottom)` would be better in terms of performance but I didn't succeed. So I ended up using a more declarative API like `GC_switch_to_corutine`. The sample script with 8 threads and 100 fibers did consistently trigger a collection with wrong stack information for one of the threads.

* Would be better to wrap then `GC.disable`/`GC.enables` to at least avoid hitting down the ST performance for now, right?
* Should I use `:mt` or `:preview_mt` as a flag directly so other MT features can be enabled all together?
* From https://github.com/ysbaddaden/crystal/commit/ab7bfb4431f6eecaa26ef982b813936cd7fca587 , we could include the RWLock and add the integration with boehm. I am not sure if it's sound to do the change on `atomic.cr` from `:sequentially_consistent` to `:acquire`/`:release`. It seems the former puts stronger constraints

  > [Sequentially-consistent ordering](https://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering) 
  > Atomic operations tagged memory_order_seq_cst not only order memory the same way as release/acquire ordering (everything that happened-before a store in one thread becomes a visible side effect in the thread that did a load), but also establish a single total modification order of all atomic operations that are so tagged.

* I will submit another set of instructions that will require creating a `.pc` file ¬¬ and remove the hack path so we can have better chances on merging and testing this.
* I would like to have some specific benchmark ready to run and try to decide better if the rwlock is worth it. Do you have the http stress test that you used?

bcardiff:
@vladfaust `Thread.new` was already there, not as a public api. As long as you don't use fibers/io on secondary threads everything is fine. So, don't use it 😸.

Fryguy:
Minor, but is `corutine` a typo (expected `coroutine`), or was that intentional for name collision reasons?

bcardiff:
I need to learn english. Thanks @Fryguy 😸

ysbaddaden:
It took me some time to understand the atomic memory ordering constraints, but AFAIK the memory ordering is declaring memory barriers for LLVM optimizer's to *not reorder across atomic instructions*. Since I only use `Atomic::Flag` for creating locks, I only care for instructions within `:acquire` and `:release` to never be moved *before* or *after* the atomic instructions, but instructions outside of the atomics can be reordered within. `:seq_cst` is more strict, and nothing can cross the memory barriers.

I believe that acquire/release is fine for locks. I'm not sure there are other use cases for Atomic::Flag that would need a strict sequential memory ordering?

bcardiff:
I've rebased this on (almost) master and integrate changes from #7551, https://github.com/crystal-lang/crystal/pull/7551#discussion_r266594361, and compare with the GC aspects of #7547.
#7547 helped me identify the problem mention at https://github.com/crystal-lang/crystal/pull/7546#issuecomment-473021383, I missed swapping the procs call in `GC.before_collect`.

I've created a couple of branches to compare strategies (more on that later).

To check this you first need to build GC with custom extensions. It match this [patch](https://gist.github.com/bcardiff/b3bf67bd04b8112ea3804fa6289ff495).

```
$ pushd ..
$ git clone http://github.com/crystal-lang/bdwgc
$ cd bdwgc
$ git checkout mt-gc-benchmark
$ git clone https://github.com/ivmai/libatomic_ops
$ cd libatomic_ops
$ git checkout v7.6.10
$ cd ..
$ autoreconf -vif
$ ./configure --disable-debug --disable-dependency-tracking --disable-shared --enable-large-config
$ make 
$ ls .libs/
libcord.a   libcord.la  libcord.lai libgc.a     libgc.la    libgc.lai
```

Also, create the following `gc.pc` file, changing the `/path/to/bdwgc` with the full path of the bdwgc working directory.

```
$ cat gc.pc
prefix=/path/to/bdwgc
libdir=${prefix}/.libs

Name: bdwgc
Description:
Version: 7.6.12+mt
Libs: ${libdir}/libgc.a
```

Then we build the `samples/mt_gc_test.cr`. It now has some command line options to measure its performance.
I've have checked some alternatives. Each of it in their own branch.

* branch: `mt/gc-switch_to_coroutine`. The `GC_switch_to_coroutine` is used in `Fiber.resume`. `GC_enable`/`GC_disable` for locking.
* branch: `mt/gc-set_stackbottom-in-resume-lock`. The `GC_set_stackbottom` is used in `Fiber.resume` with `Thread.current`. `GC_enable`/`GC_disable` for locking.
* branch: `mt/gc-set_stackbottom-lock`. The `GC_set_stackbottom` is used in `GC.before_collect`. `GC_enable`/`GC_disable` for locking.
* branch: `mt/gc-set_stackbottom-rwlock`. The `GC_set_stackbottom` is used in `GC.before_collect`. Custom `GC::RWLock` for locking.

> Note: these branches are in [my fork](https://github.com/bcardiff/crystal)

The last three uses the same GC extension. So I wanted to check if there was any benefit of using the implicit current thread or not.
If not is better to have a more flexible API since it can be used, as shown, in different ways (`Fiber.resume` vs `GC.before_collect`).

After checking out each `STRATEGY` branch, to build the `mt_gc_test` the following can be used:

```
$ rm -rf ~/.cache/crystal/ # clear temp files because reasons
$ (export PKG_CONFIG_PATH=/path/to/bdwgc:$PKG_CONFIG_PATH && ./bin/crystal build samples/mt_gc_test.cr --release -D preview_mt -o gc-STRATEGY)
```

> Note: I changed the flag to `preview_mt`.

Additionally as a baseline a `-D gc_none` can also be compiled:

```
$ rm -rf ~/.cache/crystal/ # clear temp files because reasons
$ ./bin/crystal build samples/mt_gc_test.cr --release -D gc_none -o gc-none
```

Finally, things like the following can be used to measure the performance of different strategies.

```
$ ./gc-none -m -f 1000 -t 2 -l 20
$ ./gc-switch_to_coroutine -m -f 1000 -t 2 -l 20
$ ./gc-set_stackbottom-in-resume-lock -m -f 1000 -t 2 -l 20
$ ./gc-set_stackbottom-lock -m -f 1000 -t 2 -l 20
$ ./gc-set_stackbottom-rwlock -m -f 1000 -t 2 -l 20
```

I've tested this on OSX for a couple of variables. Results can be viewed [here](https://docs.google.com/spreadsheets/d/17tZUHtQL9zbeS8_jBXQ6JYKStxFbVsEfnEr3FXZ4dIo/edit?usp=sharing).

So far it seems that the best alternative is `gc-set_stackbottom-in-resume-lock`, but they are all very similar.
The `gc-set_stackbottom-rwlock` seems to hang on some examples, but if that strategy ends up working the same GC extension could be used.

I have moved the `mt/gc` branch to `mt/gc-set_stackbottom-in-resume-lock` so that can be reviewed/approved in this PR.

Of course, if more experiments shows that another alternative is better we can switch that. But for now, that strategy seems enough.

bcardiff:
I did the edits in the samples code that @ysbaddaden the most notable changes is that the gc_none version is even faster. The others didn't change much for now. So I will stick with the previous decision of using `mt/gc-set_stackbottom-in-resume-lock` as a starting point.

ysbaddaden:
I tweaked my implementation a bit to try different lock strategies. See https://github.com/ysbaddaden/crystal/commit/093060042a149c777fee81678daae4e3a62dfeaa and https://github.com/ysbaddaden/crystal/commit/58af519543a56040eb59e950d8e325064a828a32.

Using my HTTP::Server with different tests (IO bound, GC bound, CPU bound) I got slightly better values with `pthread_rwlock` and my custom `rwlock` than with `GC.disable` but nothing really significant. I think the overhead mostly lies in libevent and IO syscalls and somewhat erases the context switch overhead.

Yet, using my context switch benchmarks, the overhead of a context switch (including the resumable CAS 1 to 2) is significantly different using the different strategies. See this plot, using `gc_none` as a reference (zero overhead):
switch

I'm worried about the hangup you got with my rwlock, but I couldn't reproduce it, despite running HTTP::Server with a GC collect running every some requests :/

ysbaddaden:
I'm fine with using `GC.disable` for now, but I'd love the GC to have the 4 lock_read, lock_write, unlock_read and unlock_write methods, so we could change what's used.

vladfaust:
I'm so happy to see the `0.28` milestone on it! :rocket: 

*Drum beat intensifies :drum:*

bcardiff:
@ysbaddaden with the last push

* The GC has the 4 lock_read, lock_write, unlock_read and unlock_write methods, yet only one strategy is used.
* Changing the strategy should imply changes only in boehm.cr
* I fixed the hang problems. There was sync issue with the worker threads when finishing/starting each round of the stress test that is fixed in https://github.com/crystal-lang/crystal/pull/7546/commits/81783cfd7b1ceb5b34570a3686ab232ded4bf53f.
* `Thread.current` is used several times already in a context switch: to access the scheduler, to set the stack_bottom. Accessing it one more time to check the single resume isn't problematic. I still prefer to have the `Fiber::Context` as an `@[Extern]`

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,229 @@
...
+    # allocate fibers on each thread queue.
+    @pending_fibers_queues.each &.clear
+    @worker_fibers.dup.shuffle!.each_with_index do |f, index|
+      @pending_fibers_queues[index % @threads] << f
+    end
+
+    @threads_state.set(State::Run)
+
+    # spin wait for all fibers to finish
+    while (c = @fibers_reached.get) < @fibers
```
Sija: What `c = ...` assignment is for? ditto below.

bcardiff: old puts debugging 🙈

samples/mt_gc_test.cr
```diff
@@ -0,0 +1,229 @@
...
+    STDERR.puts parser
+    exit(1)
+  end
+end
+
+case mode
+when :run
+  run(threads_num, fibers_num, loops_num, true)
+when :ips
+  Benchmark.ips do |x|
```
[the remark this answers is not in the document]
bcardiff: It was copy pasted from Benchmark docs.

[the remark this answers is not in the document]
r00ster91: Then that could be changed too.

src/crystal/scheduler.cr
```diff
@@ -57,9 +57,42 @@ class Crystal::Scheduler
...
+  private def ensure_single_resume(fiber)
+    # Set the current thread as the running thread of *fiber*,
+    # but only if *fiber.@current_thread* is effectively *nil*
+    if fiber.@current_thread.compare_and_set(nil, Thread.current).last
+      # the current fiber will leave the current thread shortly
+      # although it's not resumable yet we need to clear
+      # *@current.@current_thread* for a future `Scheduler.resume(@current)`
+      @current.@current_thread.set(nil)
+    else
+      fatal_resume_error(fiber, "tried to resume multiple times a fiber")
```
straight-shoota:
```suggestion
      fatal_resume_error(fiber, "tried to resume the same fiber multiple times")
```

ysbaddaden:
I don't have time to review properly. From your last comment, the patch sounds acceptable.

About `Thread.current` there should ideally be one single call to it during a context switch: to reach the current thread's scheduler. Setting the stack bottom can be moved to `GC.before_collect` without delaying GC marker threads (see https://github.com/crystal-lang/crystal/pull/7547/commits/093060042a149c777fee81678daae4e3a62dfeaa#diff-859e2b09d96c61167d13e65208236dd8R332).

bcardiff:
* Rebased on master
* Include https://github.com/crystal-lang/crystal/pull/7546#discussion_r273862225 suggestion in 825b9e4d9f15151287f96e70f2d617c8558d82ec
* Include https://github.com/crystal-lang/crystal/pull/7546#issuecomment-481627266 suggestion to move the call to set the stack bottom to `GC.before_collect` in the last commit a3632e1be4331e885bb2c7b1b8fca17d8442f903

As I mentioned in before using setting the stack bottom in the context switch or before_collect didn't show impact in the current tests. Adding the change seems reasonable.

bcardiff:
Today we've been reviewing this PR with waj and we found a missing piece introduced in the last refactor at a14b4aaf2a56131fc31fe7bf3e3e62f2db4eea25 where for single thread we were not setting the stackbottom.

Now that is fixed at for single thread app there should be no impact.

ysbaddaden:
Can we just use `mt` instead of `preview_mt`?

waj:
I prefer to use `preview_mt` for now. It requires less explanation about why `mt` is not usable yet.

Issue #6468, opened 2018-07-30
[RFC] Structured Concurrency


straight-shoota:
Crystal has an great concurrency model based on Fibers and Channels, which can be used to pass messages around.

Fibers are conceptually pretty simple. If you spawn one, it takes off from the main context and runs concurrently for an indefinite amount of time, depending on what it does and how long it takes to do this. From the perspective of the main control flow, it's essentially *fire and forget*.

Real life problems typically ask for a more sophisticated way of handling concurrent tasks. Sometimes you need to wait for either one, some or all tasks to be finished before continuing the main scope. Error handling in concurrent tasks is also important and the ability to cancel the remaining tasks if others have finished or errored.

Fibers and Channels can be used to implement a model for structured concurrency. Given that this is a pretty common idiom, I'd like to see a generalized implementation in Crystal's stdlib. 

## What we have

`HTTP::Server#listen` uses a custom implementation of a wait group executing a number of tasks simultaneously and waiting for all to finish. Other examples are in the `parallel` macro or `Crystal::Compiler#codegen_many_units`. `parallel` is the only feature of structured concurrency currently available in the stdlib, but it is only suitable for a fixed number of concurrent tasks that are known at compile time.

A more generalized approach would help to make this concept easily re-useable.
It can be implemented based on the existing features that Fiber and Channel provide. The only thing that's missing is a way to deliberatly kill fibers and unwrap their stack (see #3561, and a proposed implementation in #6450).

## Background

I recommend reading the articles referenced below. They both describe a model of structured concurrency which essentially restricts the execution of concurrent tasks to a specific scope and having tools to manage them. This contrasts with the model of `go` (Go) and `spawn` (Crystal) which just fires off a new fiber without caring about it's life cycle. This makes it hard to follow control flow: what happens where and when in which scope.

The main idea of this proposal is to understand that each fiber is limited to the scope it is executed in:

> Every time our control splits into multiple concurrent paths, we want to make sure that they join up again.

This ensures that fibers don't get lost doing whatever stuff they might not even be supposed to do anymore.
I believe this concept can be applied to almost any real-life use case of fibers.
Having a structured flow of control also allows for a proper exception flow. Right now, unhandled exceptions within a fiber are just printed and ignored. When a fiber is scoped to some parent context, an exception can just be propagated there.

## Prototype

[...]

Issue #6461, opened 2018-07-28
ability to output stack trace of all fibers


rdp:
Initial thread here:
https://groups.google.com/forum/#!topic/crystal-lang/y51iKWVIXFA

Sometimes it would be "nice to know" where all the fibers in, say, a web app are currently at, for debug and profiling purposes.

Something like Ruby's

`Thread.list.each {|t| puts t.name, t.status, t.backtrace}`

rdp:
Looks like I forgot I already requested this in #4683 closing in favor of that...

PR #7547, opened 2019-03-14
Experimental MT

ysbaddaden:
Rework of #7214 with much better integration, and acceptable overall performance: -10% slower with `NPROCS=1` and `-Dmt` than with the current scheduler for a simple HTTP::Server stress test using `wrk`.

EDIT: despite achieving good single thread performance despite being thread safe, adding more threads won't improve performance. One of four threads will roughly be equal —allright, 4 threads is slightly faster than 1, but multiple processes are still much faster.

NOTE: I want clean commits for each feature and improvement. I'll force-push revised commits instead of pushing lots of fixups —I already squashed dozens of them.


asterite:
It would be really cool if you and @bcardiff combined efforts on this topic.

From my point of view, I think it would be cool to have any safe infrastructure for running `spawn` across multiple threads, even if it's slower than now (for example using `GC.enable` and `GC.disable`). Once that's done std code can be upgraded to be multithread-safe, and also work can be done to improve the performance. In the meantime also libraries can be upgraded to consider multithreading and so on. So in a way this would be parallelizing this feature's implementation :-)

bararchy:
Its quite funny , I was 100% that you are both synced on the releases hahah

bcardiff:
Can we get https://github.com/crystal-lang/crystal/pull/7547/commits/2e19060fb007336a9002048297b6d3af9829d8be and https://github.com/crystal-lang/crystal/pull/7547/commits/491bdaea37d91d44049d66b3bc055fd38860fc87 in their own PRs?

I think we can merge those right away.

Note that https://github.com/crystal-lang/crystal/pull/7547/commits/35d30af3e05dc56e6d1d4611b2ebece0ec7368ad is already in master.

spec/std/channel_spec.cr
```diff
@@ -5,6 +5,14 @@ private def yield_to(fiber)
   Crystal::Scheduler.resume(fiber)
 end
 
+private macro it_unless_mt(name, &block)
+  {% if flag?(:mt) %}
+    pending({{name}}) { {{yield}} }
+  {% else %}
+    describe({{name}}) { {{yield}} }
```
Sija: Why not `it`?

ysbaddaden:
@asterite this is why I pushed the PR instead of keeping it hidden in some branch :)

I'm not aiming to merge for now.

I'm disappointed by the performance of MT for the time being. I'm starting to have acceptable single thread performance, despite being thread safe (overhead is contained) which is a nice achievement, but I don't get any improvement with more threads... channels are awfully slower, and select is disabled —I also broke IO::Evented but that's fixable.

The best performance is currently achieved with current master in "sharding" (distinct processes) and no GC, which has a higher impact on performance than I thought (actual collections, not the safe context switches).

I'll try a Chase and Lev queue for the scheduler. It would avoid a mmap and use a circular queue (that could grow and shrink), and allow to steal many fibers are once, instead of stealing only one at a time.

spec/std/channel_spec.cr
ysbaddaden: Typo :)

ysbaddaden:
@bcardiff extracted and rebased.

stakach:
Not entirely sure if this is the place to discuss this, however I feel like this might be the wrong direction to take the language. Fibers are awesome, they are quite mutation safe, very easy to reason with.

Might it not be better for the language to embrace single thread IO reactor with a thread pool accessible via something like a `ThreadChannel` for processor intensive operations that don't have access to IO operations.

I personally really like the look of #6468 and it feels like the complexity of managing fibers will go up once threads are introduced - given fibers should always run on the same thread.

In short, processes feel like the right way to achieve parallelism however it would be nice to perform some processor intensive operations without blocking the reactor.

ysbaddaden:
#6461 is unrelated, there can be structured concurrency with MT and stealing schedulers.

The complexity you refer to is fearless concurrency which is already a problem, even without threads, it's just not as obvious. This should have it's own discussion. I'll try to open it during the weekend with some ideas (initialize constants at start time, thread safe globals only, *moving* values accross fibers) and links to interesting articles.

As stated above, I'm disappointed with the performance, but I didn't try to add an actor to verify a bcrypt hash, or to calculate a random number in fibers, those cases should be dramatically faster the more threads we add (limited by hardware).

ysbaddaden:
Sigh. Benchmarking is tricky.

With default `wrk` settings (2 threads, 10 connections), running 2 or more threads isn't better than 1 (unless requests are heavily CPU bound) and slower than current master, which is *very* disappointing. But `wrk` with 1 thread and 128 shows something completely different. Using relative results:

```
single-thread 1 process:   × 1.00
single-thread 2 processes: × 1.35
single-thread 3 processes: × 1.65
```
```
multi-thread 1 thread:     × 1.10
multi-thread 2 threads:    × 1.50
multi-thread 3 threads:    × 1.40
```

With enough incoming connections, MT is starting to get faster then running different processes :fireworks:, yet it's still slower with 3+ threads (too much contention? scheduler? gc?).

src/concurrent/mt_scheduler.cr
```diff
@@ -0,0 +1,304 @@
...
+      else
+        park
+      end
+    end
+  end
+
+  protected def steal_loop : Fiber?
+    Thread.log "scheduler#steal_loop", self
+
+    100.times do
```
wontruefree: I am just curious why 100?
This seems like a high number but I do not do much MT programming.

ysbaddaden: Magic number. It should probably be much higher, to avoid parking the thread to quickly, while enqueue should wait for the queue to have some entries before unparking a thread, to make it worth it.

ysbaddaden: Maybe it should be adaptive, too, depending on how much time was spent waiting in the previous steal loop, and whether the thread was parked or not.

cpunion:
Great! How to build and test this?

asterite:
I think this can be closed now that #8112 was merged.

@ysbaddaden Sorry this didn't make it through!

PR #7551, opened 2019-03-14
MT: Fiber::Context struct

ysbaddaden:
The fiber context holds the current stack top pointer (for a saved stack), as well as a resumable flag that tells whether the fiber can be resumed (its context was fully saved) or is currently running, or is dead (its proc returned).

The resumable flag is required to prevent a fiber to be resumed while it's still running. This should usually not happen with the monothreaded scheduler, but still useful to find a buggy synchronisation primitive that would enqueue the same fiber multiple times (e.g. the current `select` implementation). This will be very frequent with a multithreaded scheduler where a thread B could try to resume a fiber that has just been enqueued by thread A but didn't fully store its context yet.

The fiber status, running or dead, is monitored using an `@alive` boolean. We can't set `context.resumable` to another value, because the scheduler eventually swaps the context to another fiber, which would overwrite `context.resumable`.

The scheduler now aborts with a fatal error whenever it would resume a running or dead fiber, which would otherwise lead to a segfault. It doesn't just ignore the fiber since to enqueue a fiber twice or a dead fiber is an error that must be fixed.


src/fiber.cr
```diff
@@ -67,6 +35,7 @@ class Fiber
   end
 
   def initialize(@name : String? = nil, &@proc : ->)
+    @context = Context.new
```
bcardiff: I would expect that each `Fiber#initialize` would set `@context.resumable` explicitly at least on one of the overloads.

Currently, all `Fiber#makecontext` implementations are doing `@context.resumable = 1` which seems a bit odd.

src/crystal/scheduler.cr
```diff
@@ -57,9 +57,24 @@ class Crystal::Scheduler
   end
 
   protected def resume(fiber : Fiber) : Nil
+    validate_resumable(fiber)
```
bcardiff: Should we have another state and validate that the `fiber` is `resuming`? Otherwise we are not enforcing that the `fiber` is not trying to be resumed from two threads at the same time.

Maybe the flag is not just a boolean but a pointer to the thread so we can set & validate the flag is still owned by the current thread. Maybe I am missing something but although there are not a lot of code between the validation and the swapcontext, there is still a chance for other thread to try to do the same.

src/fiber.cr
ysbaddaden: I'm not sure. I meant for the context to be manipulated by the arch-specific functions, and `makecontext` is responsible for telling when the context is actually resumable. I'm not sure Fiber should assume anything about it's resumable state (only about being alive/dead).

src/crystal/scheduler.cr
ysbaddaden: Yes, I assume the following, otherwise this is still MT unsafe:
1. a Fiber must only be queued once;
2. a Scheduler's queue must have atomic operations so a Fiber may only dequeued (including stolen) once.

Since Fiber isn't a public API, and developers should use provided synchronization primitives (e.g. Channel, Mutex, select), I believe these assumptions to be acceptable. Assumption 1 is the whole reason I disabled `select` in my MT branch, by the way —I'll eventually reenable it by spawning distinct fibers for each wait operation.

bcardiff: Wouldn't be safer and sound to add this check I propose then? For the right usages, there should be almost no overhead and will prevent invalid state. For me, it sounds like the same you wanted to avoid by rescheduling a dead fiber.

src/fiber.cr
bcardiff: Ok, let's keep it as is.

src/crystal/scheduler.cr
ysbaddaden: You're right, it's safer to add another check. It'll add an atomic op, to make sure one scheduler only gets the right to resume the fiber.

ysbaddaden: I'm trying to add this extra check, or changing `resumable` to be an atomic, but that's really tricky, and I'm not sure how to implement it :/

bcardiff: I was thinking of something like this: https://github.com/bcardiff/crystal/commit/ddf3fc18ff61a14e7b499ba837cd4852381d7438 

WDYT?

The following snippet stresses the check:

```crystal
f = Fiber.new do
  while true
  end
end

Thread.new do
  Crystal::Scheduler.resume(f) # Boom!
end

Crystal::Scheduler.resume(f) # or Boom!
```

ysbaddaden: Thanks!

I found a hack in the meantime in my MT branch (add a cas on resumable to set it from 1 to 2), but didn't check that it did work yet.

bcardiff: If you agree with that commit, do you want to add it to this PR then so we can get some ✅from others?

I tried to come up with a version using just ints to flag the state, but it was more complex to mark the lifecycle of the "incarnations" of each fiber and, at the end of the day, knowing which thread is actually executing/willing to execute each fiber is probably valuable information.

bcardiff: (also, it needs a rebase on master so the CI can be happy ...)

ysbaddaden:
I added a third state to `resumable` an made it an atomic long. Weirdly enough, it didn't make the context switch slower, it may even be slightly faster :open_mouth: 

The drawback is that we lost the `@[Extern]` annotation, because Atomic isn't allowed for extern structs.

The ASM still non atomically sets `resumable`, which I expect to be fine, since schedulers will atomically check `resumable` to be `1` then atomically transition it to `2`, which are enough guarantees.

ysbaddaden:
In the MT branch, the following segfaulted quickly. With the current patch, it didn't segfault anymore:

```crystal
fiber = Fiber.new { loop { Fiber.yield } }
100.times { Crystal::Scheduler.enqueue(fiber) }
sleep 0
```

ysbaddaden:
Note that adding `@current_thread` to Fiber, and atomically changing it wasn't as bad as I thought; roughly 20% slower than changing `resumable` to an atomic.

bcardiff:
Also, can we use `:preview_mt` as a flag until the feature is stable enough?

bcardiff:
With #7546 merged we can close this PR. That does not discard future improvements to the current state.

PR #7561, opened 2019-03-19
Add IO overload to Benchmark

j8r:
This change allows benchmarkers to directly send the report to an IO, in order to build a Markdown benchmark report for example.  Before creating a subprocess was the only way to do this.
The `interactive` ivar is no longer needed.

### DIFF
--- a/spec/std/benchmark_spec.cr
+++ b/spec/std/benchmark_spec.cr
@@ -64,7 +64,7 @@ end
 describe Benchmark::IPS::Entry, "#calculate_stats" do
   it "correctly calculates basic stats" do
     e = create_entry
-    e.calculate_stats([2, 4, 4, 4, 5, 5, 7, 9])
+    e.calculate_stats({2, 4, 4, 4, 5, 5, 7, 9})
 
     e.size.should eq(8)
     e.mean.should eq(5.0)
--- a/spec/std/benchmark_spec.cr
+++ b/spec/std/benchmark_spec.cr
@@ -9,13 +9,12 @@ typeof(begin
 end)
 
 describe Benchmark::IPS::Job do
-  it "works in general / integration test" do
+  it "executes in non-interactive" do
     # test several things to avoid running a benchmark over and over again in
     # the specs
-    j = Benchmark::IPS::Job.new(0.001, 0.001, interactive: false)
+    j = Benchmark::IPS::Job.new(0.001, 0.001)
     a = j.report("a") { sleep 0.001 }
     b = j.report("b") { sleep 0.002 }
-
     j.execute
 
     # the mean should be calculated
@@ -27,6 +26,17 @@ describe Benchmark::IPS::Job do
     first, second = [a.slower, b.slower].sort
     first.should eq(1)
     second.should be > 1
+  end
+
+  it "executes interactively with an IO" do
+    io = IO::Memory.new
+    Benchmark.ips(0.001, 0.001, io) do |bm|
+      bm.report("first benchmark") { sleep 0.001 }
+      bm.report("second benchmark") { sleep 0.002 }
+    end
+    result = io.to_s
+    result.should contain "first benchmark"
+    result.should contain "second benchmark"
   end
 end
 
--- a/src/benchmark.cr
+++ b/src/benchmark.cr
@@ -85,15 +85,15 @@ module Benchmark
 
   # Main interface of the `Benchmark` module. Yields a `Job` to which
   # one can report the benchmarks. See the module's description.
-  def bm
+  def bm(io : IO = STDOUT)
     {% if !flag?(:release) %}
       puts "Warning: benchmarking without the `--release` flag won't yield useful results"
     {% end %}
 
-    report = BM::Job.new
-    yield report
-    report.execute
-    report
+    job = BM::Job.new
+    yield job
+    job.execute io
+    job
   end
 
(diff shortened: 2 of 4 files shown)


RX14:
No, the interactive ivar is especially needed now. Outputting VT100 escape sequences unconditionally is bad behaviour, especially when outputting to something like markdown.

j8r:
Or we can simply drop the use of escape sequences

j8r:
We can also drop the use of escape sequences

j8r:
That's right, interactive behavior allows to show the results progressively.

j8r:
I've changed `interactive` to a method overload instead of an ivar, which isn't used anywhere else than in `Benchmal::IPS::Job#report`

spec/std/benchmark_spec.cr
```diff
@@ -12,21 +12,22 @@ describe Benchmark::IPS::Job do
...
+      # the mean should be calculated
+      a.mean.should be > 10
+
+      # one of the reports should be normalized to the fastest but do to the
+      # timer precisison sleep 0.001 may not always be faster than 0.002 so we
+      # don't care which
+      first, second = [a.slower, b.slower].sort
+      first.should eq(1)
+      second.should be > 1
+    end
```
yxhuvud: Shouldn't the old usage still be supported? If so, why change the test instead of adding a new one?

j8r: Because we don't want the output to be sent in the STDOUT, which is the default (as before)

j8r: I can create a new tests

j8r: I can create new tests

j8r:
I've modified the API: if we pass an IO, it must be a valid TTY to send the report interactively in it.
I don't know if it's better to raise of not send the report to the IO.
If we want to generate a custom report, we can use the `report` method.

Sija:
> I've modified the API: if we pass an IO, it must be a valid TTY to send the report interactively in it.

Why? What if you'd like to pipe it to the file?

j8r:
If you pipe to a fine @Sija , it can end up having escape sequences.
Generally if you define a custom IO other the default `STDOUT`, you know what you do and can directly use `Benchmark::IPS` instead of the `Benchmark.ips` wrapper.

j8r:
If you pipe to a file @Sija , it can end up having escape sequences.
Generally if you define a custom IO other the default STDOUT, you know what you do and can directly use Benchmark::IPS instead of the Benchmark.ips wrapper.

Sija:
@j8r `Colorize` module should handle that (escape sequences).

j8r:
All escape sequences aren't about color: this one is used to "delete" lines - (try `printf "\e[2A"`)

asterite:
@j8r Originally this PR was about being able to send the report through an IO, say, a file. But a file is not a tty. And it seems you are raising on non-tty devices. So how does this change support your original idea? I think I might be missing something...

Sija:
> All escape sequences aren't about color: this one is used to "delete" lines - (try `printf "\e[2A"`)

And where is it used?

This PR becomes really messy and your changes are way out of initially defined scope (`class` -> `struct`, tty related changes, etc.).

j8r:
Yes @asterite. I can add an example somewhere in the API docs or in the gitbook. If you want to send the benchmark to a non TTY IO:
```crystal
require "benchmark"

job = Benchmark::IPS::Job.new
job.report "a" do
  sleep 0.2  
end

job.report "b" do
  sleep 0.1
end

job.execute nil

File.open "./report", "w" do |fd|
  job.report  fd
end
```
Content in `./report`:
```
a   4.99  (200.29ms) (± 0.04%)  0.0B/op   2.00× slower
b   9.98  (100.18ms) (± 0.07%)  0.0B/op        fastest
```

j8r:
I've rebased in two commits:
- one to specifically add IO support to Benchmark. This directly relates to TTY checks, etc
- another one add minior optimizations/code improvements here and there (can be another PR of course)

asterite:
This looks weird:

```crystal
job.execute nil
```

Can't we have `execute()` and `execute_interactive(io)`?

j8r:
> And where is it used?

@Sija please read more carefully, see https://github.com/crystal-lang/crystal/blob/master/src/benchmark/ips.cr#L106

`print "\e[#{ran_items.size}A"`

j8r:
@asterite or we can have `execute()` and `execute(io)` with a mandatory IO.

asterite:
Yes, that can work too. But having to pass `nil` is a no-no.

Sija:
@j8r oh, I see. Thanks for pointing me into a right line. Perhaps it should be replaced with `Colorize.reset` (I don't remember if such method actually exists).

j8r:
Not really, I rather think we can have an API for [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) in general, not only colors - it can control the cursor, the scroll, clear the display etc.
I don't know if it has its place in the stdlib, maybe since Windows officially support them now.
That's beyond this PR, I'll add a topic in the forum.

straight-shoota:
I don't follow why this removes the `interactive: false` behaviour on `Benchmark.ips`. As far as I can see, it should work just fine with a configurable IO (default value `io.tty?`). There seems to be no reason to remove this feature and require to use a lower-level API for this.

j8r:
I have found an other way to generate a benchmark report, PR no longer needed

Issue #251, opened 2014-11-19
Add CRYSTAL_CONFIG_LIBRARY_PATH


asterite:
Similar to `CRYSTAL_CONFIG_PATH` and `CRYSTAL_CONFIG_VERSION`, add a `CRYSTAL_CONFIG_LIBRARY_PATH` environment variable that would hardcode a path in the compiler for libraries lookup.

This will make it possible to create a compiler that knows where to find essential libs (libgc, pcre) without having to pass it as en environment variable on each compilation.

Additionally, add a compiler flag to get these configurations (for example: `crystal --config PATH`) so that you can compile a compiler carrying these configurations from a previous compiler.

@jhass What are you thoughts on this?

PR #7562, opened 2019-03-19
Add CRYSTAL_LIBRARY_PATH for lookup static libraries

bcardiff:
The compiler will use the `CRYSTAL_LIBRARY_PATH` environment variable as a first lookup destination for the static libraries that are wanted to be linked.

It follows the other configuration options conventions:
* Upon building the compiler the `CRYSTAL_CONFIG_LIBRARY_PATH` variable can be used to define the default embedded.
* The `Crystal::LIBRARY_PATH` constant is defined on compiled programs.

Ref: https://forum.crystal-lang.org/t/rfc-link-configuration/546

### DIFF
--- a/src/compiler/crystal/codegen/link.cr
+++ b/src/compiler/crystal/codegen/link.cr
@@ -75,6 +75,16 @@ module Crystal
     end
   end
 
+  class CrystalLibraryPath
+    def self.default_path : String
+      ENV.fetch("CRYSTAL_LIBRARY_PATH", Crystal::Config.library_path)
+    end
+
+    class_getter paths : Array(String) do
+      default_path.split(':', remove_empty: true)
+    end
+  end
+
   class Program
     def lib_flags
       has_flag?("windows") ? lib_flags_windows : lib_flags_posix
@@ -111,7 +121,9 @@ module Crystal
 
             static = has_flag?("static") || ann.static?
 
-            if has_pkg_config && (libflags = pkg_config_flags(libname, static, library_path))
+            if static && (static_lib = find_static_lib(libname, CrystalLibraryPath.paths))
+              flags << ' ' << static_lib
+            elsif has_pkg_config && (libflags = pkg_config_flags(libname, static, library_path))
               flags << ' ' << libflags
             elsif static && (static_lib = find_static_lib(libname, library_path))
               flags << ' ' << static_lib
--- a/src/compiler/crystal/codegen/link.cr
+++ b/src/compiler/crystal/codegen/link.cr
@@ -137,6 +137,10 @@ module Crystal
           end
         end
 
+        # Append the CRYSTAL_LIBRARY_PATH values as -L flags.
+        CrystalLibraryPath.paths.each do |path|
+          flags << " -L#{path}"
+        end
         # Append the default paths as -L flags in case the linker doesn't know
         # about them (eg: FreeBSD won't search /usr/local/lib by default):
         library_path.each do |path|
--- a/src/compiler/crystal/command/env.cr
+++ b/src/compiler/crystal/command/env.cr
@@ -7,9 +7,10 @@ class Crystal::Command
     end
 
     vars = {
-      "CRYSTAL_CACHE_DIR" => CacheDir.instance.dir,
-      "CRYSTAL_PATH"      => CrystalPath.default_path,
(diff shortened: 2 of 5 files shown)


asterite:
Fixes #251 :-)

src/compiler/crystal/codegen/link.cr
```diff
@@ -75,6 +75,16 @@ module Crystal
     end
   end
 
+  class CrystalLibraryPath
+    def self.default_path : String
+      ENV.fetch("CRYSTAL_LIBRARY_PATH", Crystal::Config.library_path)
+    end
+
+    class_getter paths : Array(String) do
+      default_path.split(':').reject &.empty?
```
straight-shoota:
```suggestion
      default_path.split(':', remove_empty: true)
```

ysbaddaden:
If the goal is to fix #251 then this is fine, otherwise I fail to see to point. We can already set `LIBRARY_PATH` (for static libraries) and `LD_LIBRARY_PATH` (for shared libraries) when building, to help the linker find libraries.

bcardiff:
The goal is to have a priority lookup of static libraries. Independent of having pkg-config or not. 
That will also allow

* linking against embedded shipped libraries easier.
* end user to link against other builds of the library without learning to write a `.pc` and setup pkg-config.

I'm following what it was more or less discussed in the forum.
After this I would like to implement some version of link annotation override from compiler options.

straight-shoota:
If this only applies to static libraries, should that maybe be reflected in the env var's name? Like `CRYSTAL_STATIC_LIBRARY_PATH`?

bcardiff:
@straight-shoota maybe, but it was too long for my eyes. 

We could also append it to `-L` so it can be used for shared library also if it is prefered. (👍/👎 in comment if this is/isn't wanted)

bcardiff:
When the upcoming changes of a custom GC and embedded compiler-rt, I think we need to push this.

If the user has a bdwgc package installed, or another llvm version that is not the one the compiler is used, the pkg-config, if configured, will default to those. Ignoring the ones shipped with the compiler. 

We need some setting for libraries that should be prioritized more than pkg-config.

src/compiler/crystal/command/env.cr
```diff
@@ -7,9 +7,10 @@ class Crystal::Command
     end
 
     vars = {
-      "CRYSTAL_CACHE_DIR" => CacheDir.instance.dir,
-      "CRYSTAL_PATH"      => CrystalPath.default_path,
-      "CRYSTAL_VERSION"   => Config.version || "",
+      "CRYSTAL_CACHE_DIR"    => CacheDir.instance.dir,
+      "CRYSTAL_PATH"         => CrystalPath.default_path,
+      "CRYSTAL_VERSION"      => Config.version || "",
```
wontruefree: Could this ever be nil?

bcardiff: Crystal.version probably not since now there is a file with it. This line is from time where git describe was used. It can be handled in another PR if wanted

r00ster91: I agree. I think that's better than just `CRYSTAL_VERSION=` and Crystal developers will probably understand `CRYSTAL_VERSION=nil`.
```suggestion
      "CRYSTAL_VERSION"      => Config.version || "nil",
```

j8r: `nil` isn't a proper semantic version...

r00ster91: It should just indicate that there is no semantic version.

bew: Or leave it as `""` with a BIG `TODO: Remove nil condition` ? :stuck_out_tongue:

bew: Actually the version can be nil, see #7602

wontruefree: I didn't expect that. seems like config should not allow `nil` to be returned.
https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/config.cr#L7

Also. This might want to be its own PR.

RX14:
I think I would have preferred a `CRYSTAL_LIBRARY_OVERRIDES="foo=/path/too/foo.a bar=..."` variable, the name seems clearer and easier to use. It'd just override the entire linking lookup logic.

Issue #7487, opened 2019-02-28
Document all stdlib features that need a require statement


dscottboggs:
I think it's really important to note at the top of each feature in the docs that needs to be `require`d. For example, I just submitted [a PR](https://github.com/crystal-lang/crystal/pull/7486) which adds this documentation to the Big... structs, but I want to make an issue to note that all of these cases should be documented.

straight-shoota:
That would certainly be very helpful. Examples should ideally work by copy and paste (except maybe a little custom setup might be required sometimes).
Including the necessary `require`s for the example to work is a no-brainer and already present in some API docs.

PRs are welcome.

Maroo-b:
@straight-shoota should we close this issue as the PR was already merged?

PR #7564, opened 2019-03-20
Add missing `require` statements for the API

Maroo-b:
The purpose of this PR is to add missing `require` statements in examples, it addresses this [issue](https://github.com/crystal-lang/crystal/issues/7487)

I went through the current docs (0.27.2) and check all the provided examples.

### Issues
Iterator example: 
[link](https://github.com/crystal-lang/crystal/blob/master/src/iterator.cr#L180)
```
array_of_iters = [[1], [2, 3], [4, 5, 6]]
iter = Iterator(Int32).chain array_of_iters
ter.next # => 1
iter.next # => 2
iter.next # => 3
iter.next # => 4
```

returns this error:
```
instantiating 'Iterator::ChainsAll(Array(Int32), Int32)#next()'
src/iterator.cr:207: undefined method 'next' for Array(Int32)
```

### DIFF
--- a/src/big/big_int.cr
+++ b/src/big/big_int.cr
@@ -15,6 +15,7 @@ struct BigInt < Int
   #
   # ```
   # require "big"
+  #
   # BigInt.new # => 0
   # ```
   def initialize
@@ -26,6 +27,8 @@ struct BigInt < Int
   # Raises `ArgumentError` if the string doesn't denote a valid integer.
   #
   # ```
+  # require "big"
+  #
   # BigInt.new("123456789123456789123456789123456789") # => 123456789123456789123456789123456789
   # BigInt.new("123_456_789_123_456_789_123_456_789")  # => 123456789012345678901234567890
   # BigInt.new("1234567890ABCDEF", base: 16)           # => 1311768467294899695
@@ -350,6 +353,8 @@ struct BigInt < Int
   # Returns a string representation of self.
   #
   # ```
+  # require "big"
+  #
   # BigInt.new("123456789101101987654321").to_s # => 123456789101101987654321
   # ```
   def to_s : String
@@ -365,6 +370,8 @@ struct BigInt < Int
   # Returns a string containing the representation of big radix base (2 through 36).
   #
   # ```
+  # require "big"
+  #
   # BigInt.new("123456789101101987654321").to_s(8)  # => "32111154373025463465765261"
   # BigInt.new("123456789101101987654321").to_s(16) # => "1a249b1f61599cd7eab1"
   # BigInt.new("123456789101101987654321").to_s(36) # => "k3qmt029k48nmpd"
@@ -583,6 +590,7 @@ struct Int
   # Returns a `BigInt` representing this integer.
   # ```
   # require "big"
+  #
   # 123.to_big_i
   # ```
   def to_big_i : BigInt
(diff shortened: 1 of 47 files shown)


Sija:
IMHO `require` statements in docs should be added only at the module level. Repeating those for **every** example seems repetitive and unnecessary.

src/big/big_int.cr
```diff
@@ -350,6 +352,7 @@ struct BigInt < Int
   # Returns a string representation of self.
   #
   # ```
+  # require "big"
```
bcardiff:
```suggestion
  # require "big"
  #  
```

src/big/big_rational.cr
```diff
@@ -130,6 +130,7 @@ struct BigRational < Number
   # Divides the rational by (2 ** *other*)
   #
   # ```
+  # require "big"
```
bcardiff:
```suggestion
  # require "big"
  #  
```

src/big/big_rational.cr
```diff
@@ -139,6 +140,7 @@ struct BigRational < Number
   # Multiplies the rational by (2 ** *other*)
   #
   # ```
+  # require "big"
```
bcardiff:
```suggestion
  # require "big"
  #  
```

Maroo-b:
@bcardiff thanks for the quick review, I updated the PR to cover all the API and fix the style inconsistency.

Sija:
That's a serious bloat within the docs, whereas it pretty easy to just check main module doc comment for a needed `require`, c'mon...

bcardiff:
@Sija having self-contained examples is good. Even at the cost of 2 duplicate lines IMO.
Maybe for libs where more context is needed that does not apply. But at least for the stdlib is something I think we should aim.

straight-shoota:
I've been questioning the noise this generates, but having them is better than not having them.

And it could perhaps in the future be an option to optimize the display of API docs. It's easy to recognize repeating `require` statements and make them fade or disappear or whatever. It doesn't work the other way around.

bcardiff:
@Maroo-b I think we are ready to merge this as is.
The issue you notice is fixed in #7569 
I think there is no conflict between both PR so it's better to merge this, then the other and then if wanted the [M-Z] range can be tackled.

ok?

Maroo-b:
@bcardiff I guess it can be merged, also I already covered [A-Z] (which means every doc page for **0.27.2** version.

Maroo-b:
@bcardiff @straight-shoota I would like to contribute more to Crystal in my free time, is there a tag or a board for highest priority issues or tasks that I can work on?

straight-shoota:
Thank you very much @Maroo-b! I'm looking forward for additional contributions from you.

There is no real priority list. Important issues are usually ones [tagged with the next milestone](https://github.com/crystal-lang/crystal/issues?utf8=%E2%9C%93&q=is%3Aopen+milestone%3A0.28.0+is%3Aissue) but they're often more complex bugs (or features that are already worked). Another starting point are the [`community:` labels](https://github.com/crystal-lang/crystal/labels), for example [`community:to-implement`](https://github.com/crystal-lang/crystal/labels/community%3Ato-implement). Apart from that, just pick whatever you're interested in. You can browse the list of open issues and take what you like ;)

Maroo-b:
@straight-shoota thank you for your quick feedback!

Sure, I'll check the labeled issues :)

Issue #5338, opened 2017-12-01
OptionParser should allow flag with optional value


straight-shoota:
In #4499 it has come up, that `OptionParser` does not allow for a flag wich could either be assigned a value (`--flag=value`) or not be used without value (`--flag`). In the references PR it happend that `--color a.cr` was misinterpreted as flag `--color` with value `a.cr`, whereas it is meant to be a value-less flag `--color` and an unnamed argument `a.cr`.

Currently, the format `--flag [value]` allows a flag value to be missing, but this only works if it is the last argument or the next argument is also a flag. It would also match an unnamed argument. There should be a way to prevent that.

Now, obviously, this can't work with an empty space as separater between flag an value as this would always be ambigous, but it would work if it was enforced to be an equals sign: `--foo arg` would be an value-less flag, `--flag=value` would be a flag with value `value`.

MakeNowJust:
> this only works if it is the last argument or **the next argument is also a flag**.

It depends options handler order. For example:

```crystal
require "option_parser"

def parse(args)
  p [:parse, args]
  OptionParser.parse(args) do |opts|
    opts.on("-a [val]", "") do |val|
      p [:a, val]
    end

    opts.on("-b [val]", "") do |val|
      p [:b, val]
    end

    opts.unknown_args do |args|
      p [:unknown_args, args]
    end
  end
  puts
end

parse(%w(-a -b b))
# [:parse, ["-a", "-b", "b"]
# [:a, "-b"]
# [:unknown_args, ["b"]]

parse(%w(-b -a a))
# [:parse, ["-b", "-a", "a"]
# [:a, "a"]
# [:unknown_args, []]
# NOTE: why "-b" is not called?
```

straight-shoota:
Wow, this looks like there are some serious issues with optional parameters.

urde-graven:
This is patched in commit 0c306cad23f906d87a82ca667c862d3afde9011d, which fixes 4 issues - see new specs in this commit against crystal main branch: 

```
Failures:

  1) OptionParser optional option doesn't get separated value if specified with merged flag
     Failure/Error: flag.should eq(expect_value)

       Expected: ""
            got: "123"

     # spec/std/option_parser_spec.cr:15

  2) OptionParser optional option doesn't get separated value that looks like flag if specified with separated flag
     Failure/Error: flag.should eq(expect_value)

       Expected: ""
            got: "-g -h"

     # spec/std/option_parser_spec.cr:15

[...]

PR #7565, opened 2019-03-20
OptionParser: optional options and arguments shifting

urde-graven:
This is patch from https://github.com/crystal-lang/crystal/issues/5338#issuecomment-452672895. Its argument's parsing changes are compatible with Ruby, but not interface to maintain backward compatibility. I also converted option_parser_spec.cr from crystal to Ruby ([option_parser_spec.rb](https://gist.github.com/urde-graven/1f4a501b2cc0ee2175f2910a8fd55a13)) with following results:
```
rspec ./option_parser_spec.rb:134 # OptionParser has required option with = (3) raises
rspec ./option_parser_spec.rb:228 # OptionParser does to_s with very long flag (#3305)
rspec ./option_parser_spec.rb:255 # OptionParser calls the handler for invalid options
rspec ./option_parser_spec.rb:268 # OptionParser calls the handler for missing options
rspec ./option_parser_spec.rb:511 # OptionParser raises if flag doesn't start with dash (#4001)
rspec ./option_parser_spec.rb:368 # OptionParser -- returns a pair with things coming before and after --
rspec ./option_parser_spec.rb:385 # OptionParser -- returns a pair with things coming before and after --, without --
```

1. Argument `--flag=` generates  empty string for option `--flag=FLAG` in Ruby, but exception OptionParser::MissingArgument in Crystal. This behavior was not changed in this patch.
2. Cosmetic difference:
```
Diff:
       @@ -1,5 +1,5 @@
        Usage: foo
       -    --very_long_option_kills=formatter
       +        --very_long_option_kills=formatter
                                             long
            -f, --flag                       some flag
            -g[FLAG]                         some other flag

```
3. `opts.invalid_option` callback does't exist in Ruby
4. ` opts.missing_option` callback does't exist in Ruby
5. `opts.on("foo", "")` is supported in Ruby and its result is `Usage: prog_name [options] [foo]`
6. (and 7.) `opts.unknown_args` callback does't exist in Ruby

Ruby also  respects for example `POSIXLY_CORRECT` env. var. ("If the environment variable POSIXLY_CORRECT is set, all options must precede non-options on the command line.") and have hidden options ` --*-completion-bash` and `--*-completion-zsh`.

### DIFF
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -1,18 +1,43 @@
 require "spec"
 require "option_parser"
 
-private def expect_capture_option(args, option, value)
+private def expect_capture_option(args, option, expect_value, expect_args = [] of String,
+                                  add_args = ["arg"], block_first = false, &block)
+  args += add_args
   flag = nil
   OptionParser.parse(args) do |opts|
+    yield opts if block_first
     opts.on(option, "some flag") do |flag_value|
       flag = flag_value
     end
+    yield opts unless block_first
   end
-  flag.should eq(value)
-  args.size.should eq(0)
+  flag.should eq(expect_value)
+  args.should eq(expect_args + add_args)
 end
 
-private def expect_doesnt_capture_option(args, option)
+private def expect_capture_option(args, option, expect_value, expect_args = [] of String, add_args = ["arg"])
+  expect_capture_option(args, option, expect_value, expect_args) { }
+end
+
+private def expect_capture_optional_option(args_type, flag_type, value = "123",
+                                           expect_value = value, expect_args = [] of String, &block)
+  expect_capture_option args_type == :args_separated ? ["--flag", value] : ["--flag=#{value}"],
+    flag_type == :flag_separated ? "--flag [FLAG]" : "--flag=[FLAG]", expect_value, expect_args do |opts|
+    yield opts
+  end
+  expect_capture_option args_type == :args_separated ? ["-f", value] : ["-f#{value}"],
+    flag_type == :flag_separated ? "-f [FLAG]" : "-f[FLAG]", expect_value, expect_args do |opts|
+    yield opts
+  end
+end
+
+private def expect_capture_optional_option(args_type, flag_type, value = "123",
+                                           expect_value = value, expect_args = [] of String)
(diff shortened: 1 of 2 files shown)


straight-shoota:
Compatibility with Ruby is not a huge issue. But could you please detail the breaking changes this PR introduces?

urde-graven:
I am not aware of any for common programs that does not rely on some fixed behavior, which is unlikely. Perhaps for programs that have nonstandard  definition of flags, such as text at the end or beginning before "-" character - see changes in process_handler.

straight-shoota:
Okay, from your description I understood there would be (relevant) breaking changes.

urde-graven:
I don't understand, which one do you mean? Of programs, which depends on wrong behavior? So what next?

urde-graven:
Even If there was any such marginal breaking change, I felt that I had seen that you did not hesitate to do big breaking changes somewhere else fairly commonly. So do you want to change process_handler to be sure, or what?

asterite:
@urde-graven I think it's just to know whether this change is a breaking change so that it appears in the Changelog and people are aware of it 🙂

straight-shoota:
@urde-graven You wrote "but not interface to maintain backward compatibility." `backward compatibility` being the keyword.

But yeah, as @asterite said, breaking changes are acceptable, but if they happen, it needs to be understood.

urde-graven:
:smile: I'll look into the changelog, to know what you consider as breaking change. OptionParser is not a documented and has not defined interface, so you can not tell what is bug fix and what is change of behavior. In any case, I tried to make it backward-compatible, and changed only what I thought, no one would probably use and should not use. Why don't we take a look into code maybe we'll block right on it?

asterite:
OptionParser is the first thing people use to parse options. I'm almost sure it'd used in most Crystal code out there. Why shouldn't it be used?

urde-graven:
I didn't meant OptionParser as a whole, but its changes, for example text at the end or beginning of the flags definition.

asterite:
Oooh... I see. Yes, makes sense.

straight-shoota:
Closing stale PR. The changes might still be good (although it's hard to grasp, so 🤷), but the implementation isn't usable anymore.
Feel free to open a new issue about discussing these or other changes to `OptionParser` features.

PR #7568, opened 2019-03-20
Compiler: fix passing recursive alias to proc

### LINKED ISSUE #7567: Iterator doesn't seem to be matching type aliases
Trying to use an `Iterator` to reduce memory consumption of a large result pulled back from a database (the nodes go through a few transformations before being serialized as JSON) and getting this error:

```crystal
instantiating 'Iterator::Map(Indexable::ItemIterator(Array(Neo4j::Type), Neo4j::Type), Neo4j::Type, DatabaseMetric)#next()'
in /usr/local/Cellar/crystal/0.27.2/src/iterator.cr:722: type must be Neo4j::Type, not (Array(Neo4j::Type) | Bool | Float64 | Hash(String, Neo4j::Type) | Int16 | Int32 | Int64 | Int8 | Neo4j::Failure | Neo4j::Ignored | Neo4j::LatLng | Neo4j::Node | Neo4j::Path | Neo4j::Point2D | Neo4j::Point3D | Neo4j::Relationship | Neo4j::Success | Neo4j::UnboundRelationship | String | Time | Nil)
```

In this case, `Neo4j::Type` is an [alias for all of those types](https://github.com/jgaskins/neo4j.cr/blob/0038216f0984b84417c09edb12e61032caf2edae/src/neo4j/type.cr#L111-L131), including `Nil`.

I unfortunately don't have a reduced example for this yet. I'll see if I can find one soon.

asterite:
Fixes #7567

### DIFF
--- a/spec/compiler/semantic/alias_spec.cr
+++ b/spec/compiler/semantic/alias_spec.cr
@@ -317,4 +317,21 @@ describe "Semantic: alias" do
       Foo(Bar).new(Foo.new(1).as(Bar))
     ), "can't cast Foo(Int32) to Bar"
   end
+
+  it "can pass recursive alias to proc" do
+    assert_type(%(
+      class Object
+        def itself
+          self
+        end
+      end
+
+      alias Rec = Int32 | Array(Rec)
+
+      a = uninitialized Rec
+
+      f = ->(x : Rec) {}
+      f.call(a.itself)
+      )) { nil_type }
+  end
 end
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -268,6 +268,8 @@ module Crystal
     def covariant?(other_type : Type)
       return true if self == other_type
 
+      other_type = other_type.remove_alias
+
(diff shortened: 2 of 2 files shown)


wontruefree:
Small side note but I think that was blocking an implementation of cons, car, cdr.  Could you do this in Crystal now?

https://gist.github.com/tenderlove/908d016b611726389d03b887d63c49ca

This is from a conversation on Twitter.
https://twitter.com/tenderlove/status/1011346534158368768

asterite:
@wontruefree I don't think so. How's the Crystal implementation of that? What are the types of those lambda arguments?

wontruefree:
There was a solution with boxing and unboxing but I don't think there was ever a solution with recursive types.

wontruefree:
@asterite I was hoping something like this would work.
```crystal
alias M = Proc(Int32, M?)
cons = ->(a: Int32, b : M) { ->(m: M?) { m ? m.call(a, b.not_nil!) : nil }}
```

PR #7569, opened 2019-03-20
fix example codes (2019-03)

maiha:
Hi all, it is time to catch up the examples code once a year.
- target src: master branch ( `a17a65617 Implement Int#{leading,trailing}_zeros_count (#7520)`)
- compiler: Crystal 0.28.0-dev [a17a656] (2019-03-20)

#### NOTE
Aside from PR, there are three points that I was worried about.

##### 1. humanize

```crystal
1073741824.humanize_bytes(format: :IEC)
```

```
ambiguous call, implicit cast of :IEC matches all of
Int::BinaryPrefixFormat
```

##### 2. `Hash#dig?`

```crystal
h = {"a" => {"b" => [10, 20, 30]}}
h.dig? "a", "b"                # => [10, 20, 30]
h.dig? "a", "b", "c", "d", "e" # => nil
```
```
Error in tmp/hash/all.cr:71: instantiating 'Hash(String, Hash(String, Array(Int32)))#dig?(String, String, String, String, String)'

h.dig? "a", "b", "c", "d", "e" # => nil
  ^~~~

in crystal/src/hash.cr:169: instantiating 'Hash(String, Array(Int32))#dig?(Tuple(String, String, String, String))'

      value.dig?(*subkeys)
            ^~~~

in crystal/src/hash.cr:169: no overload matches 'Array(Int32)#dig?' with types String, String, String
Overloads are:
 - Indexable(T)#dig?(index : Int)
 - Indexable(T)#dig?(index : Int, *subindexes)

      value.dig?(*subkeys)
            ^~~~
```

##### 3. `Int32::MAX + 1` doesn't raise `Overflow`

Although the compiled compiler was build with default setting `-D preview_overflow -D compiler_rt ...`, I can't find the overflow action.
```console
$ crystal --version
Crystal 0.28.0-dev [a17a656] (2019-03-20)

LLVM: 4.0.0

$ cat test.cr
pp! Int32::MAX + 1

$ crystal test.cr
Int32::MAX + 1 # => -2147483648
```

Best regards,


wontruefree:
That overflow behavior seems worrying.

straight-shoota:
The error with `Int#humanize_bytes` looks odd, too. Something wrong with autocasting symbols? Using `:JEDEC` (the other enum value) instead of `:IEC` results in the same message: `ambiguous call, implicit cast of :JEDEC matches all of Int::BinaryPrefixFormat`

asterite:
There's a bug with named arguments and autocasting but I can't remember what was it or how to reproduce it.

src/io.cr
```diff
@@ -81,6 +81,7 @@ abstract class IO
   #
   # ```
   # STDIN.read_timeout = 1
+  # sleep 1
```
r00ster91: This is correct as-is, an `IO::Timeout` is raised when the user didn't enter something at the `gets` after 1 second. Adding this sleep would make the delay 2 seconds.
```suggestion
```

bcardiff:
> 3. Int32::MAX + 1 doesn't raise Overflow
> Although the compiled compiler was build with default setting -D preview_overflow -D compiler_rt ..., I can't find the overflow action.

You need to compiler the snippet with `-D preview_overflow` in https://github.com/maiha/crystal-examples/blob/bf2ac63fcbd51dce97065d21de3163e18111609d/src/job/compile.cr#L109 to let the compiled program have overflow behavior.

The flag that you mention is for the compiler to overflow on runtime, but does not mean that the generated program have that. It's an opt-in preview feature.

bcardiff:
I extracted a bug from `Hash#dig?` 🕵️ 🐛 

Thanks a lot @maiha this is great.

I guess that making all the examples a full crystal program (Ref: #7564) is something that benefits the tool you made, right?

src/io.cr
maiha: @r00ster91 Oh, you are completely right. I understood my misunderstanding. Thank you for the nice point! 😺

maiha:
> I guess that making all the examples a full crystal program (Ref: #7564) is something that benefits the tool you made, right?

Yep, it helps to reduce the heuristics built in here. 😺 
https://github.com/maiha/crystal-examples/blob/v0.5.1/bundled/heuristic.jnl#L110-L133

straight-shoota:
I've created an issue for the autocasting bug (#7578). `Hash.dig?` is #7570.

@maiha Have you checked the overflow issue with correct flags? Then this should be good to go! 🚀

maiha:
> @maiha Have you checked the overflow issue with correct flags? Then this should be good to go! 🚀

Yep, it worked! 😺

straight-shoota:
Thank you very much @maiha :+1:

PR #7575, opened 2019-03-22
 Don't include private constants in the docs 

r00ster91:
This no longer includes private constants in the documentation, just like private methods.


vladfaust:
Thank you for many QoL PRs you make, @r00ster91! 🙏 :heart:

PR #7580, opened 2019-03-23
Restriction update: generic vs path - specs

bew:
* Makes this work:
```cr
def foo(a : T) forall T
  1
end

def foo(a : Array(T)) forall T
  true
end

foo(Array(Int32).new) # => true
```
* Adds some restriction specs for Path vs Path & Generic vs path
* Simplify Metaclass vs Metaclass restriction implementation


bew:
Cool! thx for the review / merge!

straight-shoota:
Thanks for the PR =)

Issue #7004, opened 2018-10-29
Weird Slice#pointer method semantics


Heaven31415:
Hi, I'm not really sure what is going on with this method: [Slice#pointer](https://github.com/crystal-lang/crystal/blob/391785249f66aaf5929b787b29810aeb4af0e1e8/src/slice.cr#L229)

1. It takes size argument, which is supposed to do some kind of bounds checking, but it's off by one, as far as I can see.
2. No matter what size value you pass, it returns the same pointer (I think it was supposed to return different addresses depending on size).
3. A bit unrelated, but right now there is no explicit way to say that you want to get a pointer from slice without passing any 'dummy argument', I'm not counting **to_unsafe** as explicit.

RX14:
@Heaven31415 you're meant to pass as an argument the number of offsets from the pointer you're going to use. So `slice.pointer(5)` means you can use `slice.pointer(5)[0..4]`. It's supposed to just return the pointer, it only does bounds checking, no pointer arithmetic. For getting the pointer without bounds checking `to_unsafe` is fine.

asterite:
Yeah, that method wasn't proposed by me and I always found it a bit odd. I think we can safely remove it.

Maroo-b:
@asterite I was checking the reason why `Slice#pointer` and I found this commit https://github.com/crystal-lang/crystal/commit/22c7881586c6214896485745d109815f545f0ec5 
It seems that this method provide a validation layer and it's mentioned in the docs that Slice performs bound checks: ([link](https://crystal-lang.org/api/0.27.2/Slice.html))
> A Slice is a Pointer with an associated size.
> While a pointer is unsafe because no bound checks are performed when reading from and writing to it, reading from and writing to a slice involve bound checks. In this way, a slice is a safe alternative to Pointer.

Also, I tried locally to use `Slice#to_unsafe` instead of `Slice#pointer` and the following example doesn't raise an exception:
https://github.com/crystal-lang/crystal/blob/master/src/slice.cr#L320 
is it safer in this case to keep it?

straight-shoota:
The copy and move methods rely on the behaviour of `#pointer` for bounds checking. That would still need to be implemented in these methods.

I agree that this method should be removed, because it doesn't make much sense: It applies a bounds check but returns a pointer which has no bounds. There are only a hand full of uses in stdlib. Most are related to copy and move methods and look like `a.copy_to(b.pointer(size), b.size)`. That could just be `a.copy_to(b)` anyway.

[...]

PR #7581, opened 2019-03-23
Remove Slice#pointer

Maroo-b:
Issue: https://github.com/crystal-lang/crystal/issues/7004
As discussed with @straight-shoota and @asterite `Slice#pointer` is a confusing method and it should be removed.
 The code was refactored to keep the verification logic in `Slice` mainly for copy methods.

I was thinking if we should add a deprecation warning or just mention this as a breaking change in case someone is using `Slice#pointer`.

### DIFF
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -105,7 +105,7 @@ abstract class OpenSSL::SSL::Socket < IO
     count = slice.size
     return 0 if count == 0
 
-    LibSSL.ssl_read(@ssl, slice.pointer(count), count).tap do |bytes|
+    LibSSL.ssl_read(@ssl, slice.to_unsafe, count).tap do |bytes|
       if bytes <= 0 && !LibSSL.ssl_get_error(@ssl, bytes).zero_return?
         raise OpenSSL::SSL::Error.new(@ssl, bytes, "SSL_read")
       end
@@ -118,7 +118,7 @@ abstract class OpenSSL::SSL::Socket < IO
     return if slice.empty?
 
     count = slice.size
-    bytes = LibSSL.ssl_write(@ssl, slice.pointer(count), count)
+    bytes = LibSSL.ssl_write(@ssl, slice.to_unsafe, count)
     unless bytes > 0
       raise OpenSSL::SSL::Error.new(@ssl, bytes, "SSL_write")
     end
--- a/src/regex.cr
+++ b/src/regex.cr
@@ -513,7 +513,8 @@ class Regex
       capture_number = (name_table[capture_offset].to_u16 << 8) | name_table[capture_offset + 1].to_u16
 
       name_offset = capture_offset + 2
-      name = String.new((name_table + name_offset).pointer(name_entry_size - 3))
+      checked = name_table[name_offset, name_entry_size - 3]
+      name = String.new(checked.to_unsafe)
 
       lookup[capture_number] = name
     end
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -147,9 +147,7 @@ struct Slice(T)
   # slice2 # => Slice[12, 13, 14]
   # ```
   def +(offset : Int)
-    unless 0 <= offset <= size
-      raise IndexError.new
-    end
+    check_size(offset)
 
     Slice.new(@pointer + offset, @size - offset, read_only: @read_only)
   end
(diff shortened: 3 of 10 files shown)


src/openssl/ssl/socket.cr
```diff
@@ -105,7 +105,7 @@ abstract class OpenSSL::SSL::Socket < IO
     count = slice.size
     return 0 if count == 0
 
-    LibSSL.ssl_read(@ssl, slice.pointer(count), count).tap do |bytes|
+    LibSSL.ssl_read(@ssl, slice.to_unsafe, count).tap do |bytes|
```
asterite: I think you can just pass `slice` here, no need to invoke `to_unsafe` (this is called automatically for C bindings). Check this in other places too, please. Thank you!

Maroo-b: @asterite thank you for the quick review!
This is interesting to learn, I'll update the PR quickly. :)

Maroo-b: @asterite I tried to pass `slice` directly but it seems that calling `to_unsafe` is required in order to have a `Pointer` type.

asterite: Ah... I guess the pointer type doesn't match then. It's fine like that.

src/slice.cr
```diff
@@ -566,6 +562,12 @@ struct Slice(T)
   protected def check_writable
     raise "Can't write to read-only Slice" if @read_only
   end
+
+  protected def check_size(count : Int)
+    unless 0 <= count <= size
+      raise IndexError.new
```
Sija: I'm not sure that `IndexError` is the most appropriate here...

straight-shoota: Why not? It's raised when trying to access an index in the slice that is not in its range.

Maroo-b: @Sija  I saw that `IndexError.new` is expected to be raised for Slice size issues (destination of copy, index,...) it's also mentioned in the docs for example:
- [Slice#copy_to](https://crystal-lang.org/api/0.27.2/Slice.html#copy_to%28target%3Aself%29-instance-method)
- [Slice#copy_from](https://crystal-lang.org/api/0.27.2/Slice.html#copy_from%28source%3Aself%29-instance-method)

Sija: Fair 'nuff, I had my doubts because `Slice#copy_{to,from}` methods, which do the bounds-checking, seem to me like they should throw a dedicated Exception class instead, but maybe it's just me ;)

src/slice.cr
```diff
@@ -566,6 +562,12 @@ struct Slice(T)
   protected def check_writable
     raise "Can't write to read-only Slice" if @read_only
   end
+
+  protected def check_size(count : Int)
```
straight-shoota: Why is this method `protected`?

Maroo-b: The only reason this is a`protected` method is to follow the same visibility of `check_writable` method. 
Do you suggest to make it private or public?

straight-shoota: I'm not sure why `check_writable` is `protected` either... I guess they could both be private.

Maroo-b: I'll update the visibility of those 2 methods to `private` then.

straight-shoota: Wait a minute, maybe someone else has input on this.

asterite: There's `target.check_writable` in the code in `slice.cr`. That means it needs to be `protected`.

(I don't like the existing visibility rules, but that's what we have)

asterite: If you don't need to do `obj.check_size` (with an explicit receiver) then there's no need to make it `protected`.

Maroo-b: Thank you @asterite for the explanation, I updated the visibility of `check_size` to be private.

src/regex.cr
```diff
@@ -513,7 +513,8 @@ class Regex
       capture_number = (name_table[capture_offset].to_u16 << 8) | name_table[capture_offset + 1].to_u16
 
       name_offset = capture_offset + 2
-      name = String.new((name_table + name_offset).pointer(name_entry_size - 3))
+      checked = (name_table + name_offset)[0, (name_entry_size - 3)]
```
straight-shoota:
```suggestion
      checked = name_table[name_offset, name_entry_size - 3]
```

Maroo-b: @straight-shoota  can you please explain to me the difference?
It seems that `name_table + name_offset` returns a new Slice, is this the reason for the changer?
I'm sorry, I'm asking a lot of questions because I'm curious to understand why.

straight-shoota: Sure. `slice + x` is essentially the same as `slice[x, slice.size - x]`. Both return a new slice. Instead of chaining two calls to advance the pointer and then limit the size, this can be done in a single call. The result is the same, but it avoids some unnecessary duplicate instructions.

Issue #6685, opened 2018-09-08
Disallow question mark (?) and exclamation mark (!) at the end of the LHS in an assignment.


dawkot:
```crystal
module M
  @@x? = nil # => Syntax error: unexpected token: =
end

class C
  def initialize
    @x? = nil # => Syntax error: unexpected token: =
  end
end
```

jwoertink:
@dawkot I'm pretty sure this is expected behavior. The question mark is used at the end of a method to notate that it returns a boolean value, or used in the case of returning nil as opposed to throwing an exception like in the case of an array `x[0]?` or `active?`. It doesn't really make sense with a variable since you're assigning the value and not questioning if the value will exist, or if it's returning a boolean.

j8r:
This works at the top level, with e.g. `x? = 1`.
It can be a good to have for ivars/cvars too, specially for class macros like `property enabled?  = true`.
The `property?`, `getter?`, `setter?` macros could be removed then.

j8r:
Ok `setter?` doesn't exist.
Does the class macros ending with `?` only adds a `?` to the method, compared to the non question mark variant?!

asterite:
This doesn't work and it's expected behaviour. The actual bug is that `x? = 1` is allowed (it isn't allowed in Ruby). The `?` and `!` suffixes are exclusive for method calls, not for variables.

jhass:
Did we make that valid for `setter x? = 1`or something perhaps?

asterite:
I think it's a side-effect of how the parser works. We never designed it specifically, it just worked from the beginning.

j8r:
There are two options:
1) Forbid names ending with `?` for every type of variables
2) Allow it everywhere, thus reducing the need of having `setter? var` and `property? var` since we will be able to use `setter var?` and `property var?`

jwoertink:
If voting is a thing, then I vote option 1 lol. `x? = 1` feels like dipping pizza in mayo.

vladfaust:
I personally liked the `?` syntax:

```crystal
x? = rand > 0.5 ? 1 : nil

if x = x?
  puts x
end
```

It was also useful for nilable arguments:

```crystal
def foo(x x? : Int32? = nil)
  if x = x?
  end
end
```

It worker well in those cases when thinking of non-nil variable name variant is hard. For example, `if user = user?` is easier than `if the_user = user` or `if non_nil_user = user`.

straight-shoota:
[...]

PR #7582, opened 2019-03-24
Disallow '!' or '?' at the end of the LHS in an assignment

Maroo-b:
Issue: https://github.com/crystal-lang/crystal/issues/6685
This affects mainly variable defined in the top level, instance and class variable raise already an exception.

I'm curious how do you guys debug the compiler, I'm still new here and I was just adding `puts` statement for debugging, is there any trick to know that could make me or someone else more productive in the future?

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -315,7 +315,6 @@ module Crystal
           break
         when :"="
           slash_is_regex!
-
           if atomic.is_a?(Call) && atomic.name == "[]"
             next_token_skip_space_or_newline
 
@@ -331,6 +330,11 @@ module Crystal
 
             if atomic.is_a?(Var) && atomic.name == "self"
               raise "can't change the value of self", location
+            end
+
+            if atomic.is_a?(Call) && (atomic.name.ends_with?('?') ||
+               atomic.name.ends_with?('!'))
+              raise "unexpected token: =", location
             end
 
             atomic = Var.new(atomic.name).at(atomic) if atomic.is_a?(Call)
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -333,7 +333,7 @@ module Crystal
             end
 
             if atomic.is_a?(Call) && (atomic.name.ends_with?('?') ||
-               atomic.name.ends_with?('!'))
+               (atomic.name.ends_with?('?') || atomic.name.ends_with?('!'))
               raise "unexpected token: =", location
             end
 
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -332,8 +332,7 @@ module Crystal
               raise "can't change the value of self", location
             end
 
-            if atomic.is_a?(Call) && (atomic.name.ends_with?('?') ||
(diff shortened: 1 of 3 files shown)


asterite:
We debug with puts like you

src/compiler/crystal/syntax/parser.cr
```diff
@@ -333,6 +332,10 @@ module Crystal
               raise "can't change the value of self", location
             end
 
+            if atomic.is_a?(Call) && atomic.name.ends_with?(/[?!]/)
```
j8r: Using `ends_with?('?')` and `ends_with?('!')` is faster than a regexp
```cr
require "benchmark"

REG = /[?!]/
STR = "123!"

Benchmark.ips do |x|
  x.report("regexp") do
    STR.ends_with? REG
  end

  x.report("chars") do
    STR.ends_with?('?') || STR.ends_with?('!')
  end
end
```
```
regexp 710.41k (  1.41µs) (± 4.25%)  496 B/op  639.10× slower
 chars 454.02M (   2.2ns) (± 2.09%)    0 B/op         fastest
```

Maroo-b: @j8r thank you for the benchmarking! 
I'll edit the condition to be  `atomic.name.ends_with?('?') || atomic.name.ends_with?('!')`

j8r:
@Maroo-b there is also the @bcardiff's [puts](https://github.com/crystal-lang/crystal/issues/6615#issuecomment-416740878):  `pp!({x, x.class, typeof(x)})`
One can have:
```cr
def pp_bcardiff(x)
  pp!({x, x.class, typeof(x)}
end
```
Profit :smile:

Maroo-b:
@j8r I was actually searching for this method in the code source 😄 . Maybe we should add for internal use only ⚒

src/compiler/crystal/syntax/parser.cr
```diff
@@ -333,6 +332,11 @@ module Crystal
               raise "can't change the value of self", location
             end
 
+            if atomic.is_a?(Call) && (atomic.name.ends_with?('?') ||
+               atomic.name.ends_with?('!'))
```
straight-shoota: Minor: Logic is easier to follow when both name checks are on the same line:
```suggestion
               (atomic.name.ends_with?('?') || atomic.name.ends_with?('!'))
```

src/compiler/crystal/syntax/parser.cr
```diff
@@ -333,6 +332,11 @@ module Crystal
               raise "can't change the value of self", location
             end
 
+            if atomic.is_a?(Call) && (atomic.name.ends_with?('?') ||
```
straight-shoota: You need to remove the part from this line. Couldn't post a multi-line diff suggestion.

Maroo-b:
Thank you @straight-shoota for the review, I updated the PR as suggested.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -315,7 +315,6 @@ module Crystal
           break
         when :"="
           slash_is_regex!
-
```
Sija: Pedantic: That line could be left as is :)

Maroo-b: @Sija Done :)

Issue #7566, opened 2019-03-20
New offsetof() expression


malte-v:
I thought it would be nice to have a C-style `offsetof()` macro besides `sizeof()`, which could be practical when interfacing with C. My implementation currently looks like this:

```cr
macro offsetof(path, property_name)
  {%
    type = path.resolve
    properties = type.instance_vars
    offset_expr = "0"
  %}

  {% for current_property in properties %}
    {% if current_property.name == property_name %}
      ({{offset_expr.id}})
    {% end %}
    {% offset_expr = offset_expr + " + sizeof(" + "#{current_property.type.name}" + ")" %}
  {% end %}
end
```

This seems to work for me, although I don't know whether Crystal performs some sort of packing/alignment which would break this. If so, is there a different way to approach this?

straight-shoota:
It's not safe to calculate member offsets by hand. There is no guarantee that the binary representation corresponds to the order and size in the type declaration. LLVM can and will move members around.

So only LLVM can tell the actual offset used of a specific ivar, but we can expose this as an intrinsic method like `sizeof`.

malte-v:
Thanks for the quick clarification! Probably, [`LLVMOffsetOfElement`](http://llvm.org/doxygen/group__LLVMCTarget.html#ga9971347f4072d348862519bbacbd71a7) is what should be used here (it's already in `LibLLVM`). I would be happy to help implementing this if you point me in the right direction. This requires introducing a new AST node, right? And if so, what's the process of creating one?

asterite:
Before adding any new feature to the standard library or the language we should answer the question of why this is needed. Why do you need the `offsetof` function?

malte-v:
@asterite Concrete use case:
I'm currently getting into Vulkan (the graphics API) and I'm using Crystal since it's my programming language of choice. At some point you will need tell Vulkan how to pass data from the main program (written in Crystal) to some shader (written in GLSL/HLSL/...). In order to do that, you must define a data structure which contains all per-vertex data. A minimalistic example would be:

```
struct Vertex
  property pos : Vec2
  property color : Vec3

  def initialize(@pos = Vec2.zero, @color = Vec3.zero)
  end
end
```

Then, you fill in a structure containing the information on how this data is laid out:

```
struct Vertex
  ...

  # this is just for completeness, see below for the relevant part
  def self.get_binding_description
    description = Vk::VertexInputBindingDescription.new
    description.binding = 0
    description.stride = sizeof(Vertex)
[...]

PR #7589, opened 2019-03-25
New offsetof expression

malte-v:
This adds a new `offsetof` expression to the language, as discussed in issue #7566.


spec/compiler/semantic/offsetof_spec.cr
```diff
@@ -0,0 +1,27 @@
...
+  it "errors on typeof inside offsetof expression" do
+    assert_error "struct Foo; @a = 0; end; foo = Foo.new; offsetof(typeof(foo), @a)", "can't use typeof inside offsetof expression"
+  end
+
+  it "gives error if using offsetof on something that can't have instance variables" do
+    assert_error "offsetof(Int32, @foo)", "type Int32 can't have instance variables"
+  end
+
+  it "gives error if using offsetof on something that's neither a class nor a struct" do
+    assert_error "module Foo; @a = 0; end; offsetof(Foo, @a)", "Foo is neither a class nor a struct, it's a module"
```
bcardiff: I think we need a spec for `class Foo` since the error message states that is should work with classes.

src/compiler/crystal/macros.cr
```diff
@@ -1187,6 +1187,17 @@ module Crystal::Macros
   class Out < UnaryExpression
   end
 
+  # An `offsetof` expression.
+  class OffsetOf < ASTNode
+    # Returns the type that has been used in this `offsetof` expression.
+    def structure : TypeNode
```
bcardiff: I'm not sure about the naming if classes are allowed.

src/compiler/crystal/tools/playground/public/vendor/codemirror-5.38.0/mode/crystal/crystal.js
```diff
@@ -30,7 +30,7 @@
     var keywords = wordRegExp([
       "abstract", "alias", "as", "asm", "begin", "break", "case", "class", "def", "do",
       "else", "elsif", "end", "ensure", "enum", "extend", "for", "fun", "if",
-      "include", "instance_sizeof", "lib", "macro", "module", "next", "of", "out", "pointerof",
+      "include", "instance_sizeof", "lib", "macro", "module", "next", "of", "offsetof", "out", "pointerof",
```
bcardiff: This change should go against codemirror. It doesn't hurt to keep it, but I would change the version to `codemirror-5.38.0-offsetof` to hightlight that we are using a custom version.

spec/compiler/semantic/offsetof_spec.cr
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  it "errors on offsetof uninstantiated generic type" do
```
straight-shoota: Maybe there should also be a successful example for `offsetof(Foo(T), @a)`

src/compiler/crystal/macros.cr
straight-shoota: Why not simply call it `type`?

src/compiler/crystal/macros.cr
```diff
@@ -1187,6 +1187,17 @@ module Crystal::Macros
...
+    end
+
+    # Returns the name of the instance variable used in this `offsetof` expression as a MacroId.
+    def member : MacroId
```
straight-shoota: We're using the name `instance_variable` in other places, so this should follow.

src/compiler/crystal/tools/playground/public/vendor/codemirror-5.38.0/mode/crystal/crystal.js
straight-shoota: I don't think it's worth it making a custom change to this vendored file. We should just adapt the upstream defintion and updated accordingly.

src/compiler/crystal/macros.cr
malte-v: We can call it `type` in the macros module, but its internal name must be something else since `ASTNode` already defines `@type`. Maybe `base`, like "we take the offset of an element from its base"?

straight-shoota: Right. I'd simply go with `offset_type` then. No need to introduce another name.

malte-v:
@straight-shoota Sorry, didn't look into the issue once more. I'd definitely go with the `offsetof(Type, @instance_var)` syntax. It resembles the C style while also showing that this only works with instance variables. `Type.@instance_var` looks a bit noisy and like invoking a class method to me.  If you're using `offsetof`, you're likely to know it from C and you would know that `@instance_var` references some instance variable of the specified type. But that's my opinion.

src/compiler/crystal/macros.cr
```diff
@@ -1187,6 +1187,17 @@ module Crystal::Macros
   class Out < UnaryExpression
   end
 
+  # An `offsetof` expression.
+  class OffsetOf < ASTNode
+    # Returns the type that has been used in this `offsetof` expression.
+    def offsetof_type : ASTNode
```
r00ster91: Just `type` perhaps? Seems a bit redundant when you have to say again that you want the type of the `offsetof` when you already call it on an `offsetof` expression.

Sija: https://github.com/crystal-lang/crystal/pull/7589#discussion_r269084433

malte-v: @r00ster91 Good point. Though its internal name must stay `offsetof_type`: https://github.com/crystal-lang/crystal/pull/7589#discussion_r269084433

src/compiler/crystal/semantic/main_visitor.cr
```diff
@@ -2581,6 +2581,40 @@ module Crystal
...
+      @in_type_args -= 1
+
+      type = node.offsetof_type.type?
+
+      node.offsetof_type.raise "type #{type} can't have instance variables" unless type.is_a?(InstanceVarContainer)
+      node.offsetof_type.raise "can't use typeof inside offsetof expression" if node.offsetof_type.is_a?(TypeOf)
+
+      instance_vars = type.all_instance_vars
+      ivar_name = node.instance_var.as(InstanceVar).name
+      ivar_index = instance_vars.keys.index(ivar_name)
```
asterite: All of this can be replaced with `type.index_of_instance_var(node.instance_var.as(InstanceVar).name)` (extracting `node.instance_var.as(InstanceVar).name` to `ivar_name` is fine)

src/compiler/crystal/semantic/main_visitor.cr
```diff
@@ -2581,6 +2581,40 @@ module Crystal
...
+      node.offsetof_type.raise "can't take offsetof element #{ivar_name} of uninstantiated generic type #{type}" if type.is_a?(GenericType)
+
+      if type && type.struct?
+        expanded = NumberLiteral.new(@program.offset_of(type.sizeof_type, ivar_index).to_s, :i32)
+        expanded.type = @program.int32
+        node.expanded = expanded
+      elsif type && type.instance_type.devirtualize.class?
+        expanded = NumberLiteral.new(@program.instance_offset_of(type.sizeof_type, ivar_index).to_s, :i32)
+        expanded.type = @program.int32
+        node.expanded = expanded
```
asterite: The only difference between these two branches is:

```crystal
@program.offset_of(type.sizeof_type, ivar_index)
# and
program.instance_offset_of(type.sizeof_type, ivar_index)
```

we can assign that to a variable and do all of the `expanded = ...; expanded.type = ...; node.expanded = ...` outside of the `if`s.

Issue #3624, opened 2016-12-02
When slices are converted to array, code is not always executed


benoist:
The following shows code not being executed at all when using slices of a union type.

```crystal
slice = Slice(Int32).new(1)
puts slice.to_a[0] == 1

slice = Slice(Float32).new(1)
puts slice.to_a[0] == 1

slice = Slice(Int32 | Float32).new(1)
if slice.to_a[0] == 1
  puts "true"
else
  puts "false"
end
puts "finished"

## Output
# false
# false
# finished
```
https://play.crystal-lang.org/#/r/1fcr

It also stops printing the contents.

```crystal
slice = Slice(Int32).new(1)
puts slice.to_a

slice = Slice(Float32).new(1)
puts slice.to_a

slice = Slice(Int32 | Float32).new(1)
puts slice.to_a

## Output
# [0]
# [0]
# [
```

asterite:
`Slice.new(size)` allocates memory of that size initialized to zeros. Not zero (not 0_i32 or 0_f32), but memory zero. However, memory zero doesn't denote any value of a union of `Int32 | Float32`. So it's unsafe, and you get undefined behaviour.

We should probably note this in the method docs, saying that unless you are using a primitive type like an integer or float, chances are that you are using the wrong method.

`Slice(T).new(size)` is usually used when T is UInt8, so for `Bytes.new(3)` (ask 3 bytes of memory heap and get a slice to it)

RX14:
@asterite Maybe things like this are a valid use case for an `unsafe` attribute which makes sure people at least know what they're getting into. I probably wouldn't have clocked this error at first because in my mind `Slice` is a "safe" type.

ysbaddaden:
Is it really meaningful to have slices of unions?

asterite:
@ysbaddaden I don't know, but a slice is a generic type. Maybe we can limit `Slice.new(size)` to primitive types such as integers and floats and error in another case. That will make the method (and the whole type) safe, because that's the only unsafe method.

PR #7591, opened 2019-03-26
Fix Slice#clone for non-primitive types and deep copy

### LINKED ISSUE #7590: possible issue with Slice#clone
Code source:
https://carc.in/#/r/6l2p
```
slice = Slice["abc", "a"]

slice.clone
```
```
Error in line 3: instantiating 'Slice(String)#clone()'

in /usr/lib/crystal/slice.cr:125: Can only use primitive integers and floats with Slice.new(size), not String

    copy = self.class.new(size)
```

It's strange how Slice was created with String elements but it fails to clone the instance.
Is this an intended behavior?

straight-shoota:
Fixes `Slice#clone` for non-primitive types (#7590) and fixes `Sice#clone` to perform a deep copy, `Slice#dup` a memory copy.

### DIFF
--- a/spec/std/slice_spec.cr
+++ b/spec/std/slice_spec.cr
@@ -413,6 +413,11 @@ describe "Slice" do
     slice = Slice.new(6, read_only: true) { |i| i + 1 }
     slice[2..4].read_only?.should be_true
   end
+
+  it "#clone non-primitive" do
+    slice = Slice["abc", "a"]
+    slice.clone.should eq slice
+  end
 end
 
 private def itself(*args)
--- a/spec/std/slice_spec.cr
+++ b/spec/std/slice_spec.cr
@@ -414,9 +414,46 @@ describe "Slice" do
     slice[2..4].read_only?.should be_true
   end
 
-  it "#clone non-primitive" do
-    slice = Slice["abc", "a"]
-    slice.clone.should eq slice
+  describe "#clone" do
+    it "clones primitive" do
+      slice = Slice[1, 2]
+      slice.clone.should eq slice
+    end
+
+    it "clones non-primitive" do
+      slice = Slice["abc", "a"]
+      slice.clone.should eq slice
+    end
+
+    it "buffer copy" do
+      slice = Slice["foo"]
+      copy = slice.clone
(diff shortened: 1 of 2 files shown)


asterite:
I think it's still wrong. `clone` should perform a deep copy, so the elements inside the slice must be `clone`d too.

straight-shoota:
You're right. That was already broken before #3624 was fixed.

straight-shoota:
So, we have three levels of copy:

1. struct copy: duplicates the ivar values pointer and size (current `#dup`).
2. buffer copy: the data is duplicated to a new memory location (current `#clone`).
3. deep copy: calls `#clone` on each item and stores the results in newly allocated memory (new `#clone`).

When fixing `#clone`, we're missing a method for buffer copy. Should we maybe rename the current behaviour to `#dup`? It would be different from other struct's default `#dup` method, but that should be fine. Struct copy is always implicit when passed as value.

bcardiff:
I don't see it as three levels. 
When cloning a container structure you want to avoid aliasing of values (deep copy)
When dupping a container structure you want to have aliasing of values. (shallow copy)

asterite:
I think for 2 he means allocating a new buffer for the data, but use the same data. If you move the struct around (struct copy) it will be the same pointer/buffer. So I think yes, 2 is missing and we could use `#dup` for that.

straight-shoota:
Exactly. I've implemented my suggestion and provided specs to demonstrate the behaviour.

kematy:
How can this ruby language like code be simplified when deep cloning?
class Object
      def dclone
        case self
          when Fixnum,Bignum,Float,NilClass,
               FalseClass,TrueClass
            klone = self
          when Hash
            klone = self.clone
            self.each{|k,v| klone[k] = v.dclone}
          when Array
            klone = self.clone
            klone.clear
            self.each{|v| klone << v.dclone}
          else
            klone = self.clone
        end
        klone.instance_variables.each {|v|
          klone.instance_variable_set(v,
            klone.instance_variable_get(v).dclone)
        }
        klone
      end
    end

Sija:
@kematy Please use the forum for questions like that.

PR #7594, opened 2019-03-26
Improve documentation of Hash

### LINKED ISSUE #7593: Is Hash ordered based on when the entry was inserted?
Hi, everybody.

  I just want to make sure whether `Hash#keys` or `Hash#each` returns entries in order based on when they were inserted. I found it's mentioned in [`Hash#key_index`](https://crystal-lang.org/api/0.27.2/Hash.html#key_index(key)) in document,  and it looks like it's ordered viewing the code of `key_index`. But nothing mentioned in any other place about this, and the document of [`Set` ](https://crystal-lang.org/api/0.27.2/Set.html) say `Set` based on `Hash` is unordered.

straight-shoota:
Closes #7593

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1,8 +1,36 @@
 require "crystal/hasher"
 
-# A `Hash` represents a mapping of keys to values.
+# A `Hash` represents a collection of key-value mappings, similar to a dictionary.
 #
-# See the [official docs](http://crystal-lang.org/docs/syntax_and_semantics/literals/hash.html) for the basics.
+# Main operations are storing a key-value mapping (`#[]=`) and
+# querying the value associated to a key (`#[]`). Key-value mappings can also be
+# deleted (`#delete`).
+# Keys are unique within a hash. When adding a key-value mapping with a key that
+# is already in use, the old value will be forgotten.
+#
+# ```
+# # Create a new Hash for mapping String to Int32
+# hash = Hash(String, Int32).new
+# hash["one"] = 1
+# hash["two"] = 2
+# hash["one"] # => 1
+# ```
+#
+# [Hash literals](http://crystal-lang.org/reference/syntax_and_semantics/literals/hash.html)
+# can also be used to create a `Hash`:
+#
+# ```
+# {"one" => 1, "two" => 2}
+# ```
+#
+# Implementation is based on an open hash table.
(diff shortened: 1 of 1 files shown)


src/hash.cr
```diff
@@ -397,6 +425,8 @@ class Hash(K, V)
   #   key_and_value # => {"foo", "bar"}
   # end
   # ```
+  #
+  # The enumeration follows the order the keys were inserted.
```
Sija: I'd add it as `NOTE`, ditto all below.

src/hash.cr
```diff
@@ -1,8 +1,36 @@
...
+# ```
+#
+# Implementation is based on an open hash table.
+# Two objects refer to the same hash key when their hash value (`Object#hash`)
+# is identical and both objects are equal to each other (`Object#==`).
+#
+# Enumeration follows the order that the corresponding keys were inserted.
+#
+# NOTE: When using mutable data types as keys, changing the value of a key after
+# it was inserted into the `Hash` may lead to undefined behaviour.
```
asterite: Optional: maybe mention that in those cases you can call `rehash`?

src/hash.cr
straight-shoota: I don't think these labels should be overused.

src/hash.cr
```diff
@@ -1,8 +1,37 @@
...
 
-# A `Hash` represents a mapping of keys to values.
+# A `Hash` represents a collection of key-value mappings, similar to a dictionary.
 #
-# See the [official docs](http://crystal-lang.org/docs/syntax_and_semantics/literals/hash.html) for the basics.
+# Main operations are storing a key-value mapping (`#[]=`) and
+# querying the value associated to a key (`#[]`). Key-value mappings can also be
+# deleted (`#delete`).
+# Keys are unique within a hash. When adding a key-value mapping with a key that
+# is already in use, the old value will be forgotten.
```
r00ster91:
```suggestion
# is already in use, the old value will be overwritten.
```

asterite: I'm not sure, "overwritten" sounds like the actual value will be modified, but that's not the case. I think "forgotten" is fine.

Sija: Maybe _replaced_ would do?

PR #7595, opened 2019-03-26
feature(time): add Time.now_seconds class method

ndudnicz:
Hi, added Time.now_seconds method. JS's Time.now like method, that give the current ts in second.
Didn't add spec since it uses Time.utc and Time::Span which already have specs.

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -741,6 +741,11 @@ struct Time
     )
   end
 
+  # Returns an epoch current timestamp in second
+  def self.now_seconds : Int64
+    (Time.utc - Time::UNIX_EPOCH).to_i
+  end
+
   # Returns a copy of `self` with time-of-day components (hour, minute, second,
   # nanoseconds) set to zero.
   #


straight-shoota:
Hi @ndudnicz, welcome and thank you for contributing to Crystal!

Did you notice there is already a `#to_unix` method? So using `Time.utc.to_unix` is essentially the same as `Time.now_seconds`. It's a trivial method chain and the same amount to type, but it's more descriptive of the meaning. `now_seconds` doesn't specify what kind of seconds it relates to (it might also be similar to `Time.local.seconds` for example).

So, I don't think adding this method provides a reasonable benefit.

ndudnicz:
Hmm yes I didn't see. It is in the `0.27.2` doc, but `Time.utc.to_unix` does not work with `crystal 0.27.2` installed from brew:
```bash
$> crystal -v
Crystal 0.27.2 (2019-02-05)

LLVM: 6.0.1
$> crystal run main.cr
Error in main.cr:7: missing arguments: seconds, nanoseconds


  puts Time.utc.to_unix
            ^~~
```
So I think I got confused. thx

straight-shoota:
Yeah, sry I didn't point out that it is `.utc_now` in 0.27.2. The method has been renamed to `.utc` in master (see https://crystal-lang.org/api/master/Time.html#utc%3ATime-class-method).

Issue #7027, opened 2018-11-05
Compile-time deprecation warnings


j8r:
There are no proper way to deprecate methods/objects.
The common way is to add a comment like `# TODO: Remove after 0.28.0` but the end user won't be informed of this future breaking-change when compiling its program. (S)he will only notice the missing object/methods when using the new release that remove them, and has to read the `CHANGELOG` to figure out what breaking-change is the source of this, and what to do.

## Solution

C# has the `Obsolete` annotation for deprecation, java has `@Deprecated`

An annotation can be used to deprecate method and objects, along with a deprecation message:
```crystal
@[Deprecated("will be removed in 0.28.0, use `new_method` instead")]
def old_method
end
```
This will print a colored deprecation notice like: ``Deprecated: old_method: will be removed in 0.28.0, use `new_method` instead"``

This could also add a deprecation message below the method in the generated docs.

RX14:
This can be - and has been - done by putting 

```cr
def old_method
  {{ puts "old_method is depecated and will be removed in xxx, use yyy" }}
end
```

see https://github.com/crystal-lang/crystal/pull/3563/files#diff-4760ade465a047460c75394908828da7R377


I'm not so sure we need a better solution for this *right now*, but later (close to 1.0) adding one will help external tooling.

j8r:
I guess we can consider this issue solved with https://github.com/crystal-lang/crystal/pull/7661 and all the other PRs @bcardiff :smile: 
This functionality was implemented faster than I thought. A sign that Crystal is gaining stability and approaching to 1.0?

straight-shoota:
Quick adoption of new features is not necessarily a sign of stability 😝 

And it's not complete yet. Discussion continues in #7655 👍

PR #7596, opened 2019-03-26
Quality checks / Deprecated methods

bcardiff:
This PR adds to `build` and `spec` commands a `--check` option.

The `--check` option will notify if the emitted program is using any method marked as `@[Deprecated("migration message")]`.

For the `build` command if `--check` detect issues the exit status is `1`.

For the `spec` command I think it would be better to, even if quality checks errors are detected, to run the specs. But to do so `Crystal::Command#execute` needs to be refactored to overwrite the normal exit path. Before doing so, feedback is welcome.

I detect two issues, these are fixed in the PR but could be extracted if preferred.
* The bash autocompletion was still referring to compile instead of the build command
* The annotations were not cloned in `Def`

I didn't cover and I think it can be delayed:
* Deprecation of a whole class/module.
* Structured json output format of this messages
* A proper refactor to product warning messages without creating a `Crystal::TypeException`

The tool as is, already help me detect lots of `/` that we need to change to `//`.

### DIFF
--- a/spec/compiler/codegen/warnings_spec.cr
+++ b/spec/compiler/codegen/warnings_spec.cr
@@ -1,0 +1,157 @@
+require "../spec_helper"
+
+describe "Code gen: warnings" do
+  it "detects top-level deprecated methods" do
+    assert_warning %(
+      @[Deprecated("Do not use me")]
+      def foo
+      end
+
+      foo
+    ), "Warning in line 6: Deprecated top-level foo. Do not use me",
+      inject_primitives: false
+  end
+
+  it "deprecation reason is optional" do
+    assert_warning %(
+      @[Deprecated]
+      def foo
+      end
+
+      foo
+    ), "Warning in line 6: Deprecated top-level foo.",
+      inject_primitives: false
+  end
+
+  it "detects deprecated instance methods" do
+    assert_warning %(
+      class Foo
+        @[Deprecated("Do not use me")]
+        def m
+        end
+      end
+
+      Foo.new.m
+    ), "Warning in line 8: Deprecated Foo#m. Do not use me",
+      inject_primitives: false
+  end
+
+  it "detects deprecated class methods" do
+    assert_warning %(
+      class Foo
+        @[Deprecated("Do not use me")]
+        def self.m
+        end
+      end
+
+      Foo.m
+    ), "Warning in line 8: Deprecated Foo.m. Do not use me",
+      inject_primitives: false
+  end
+
+  it "detects deprecated generic instance methods" do
+    assert_warning %(
+      class Foo(T)
+        @[Deprecated("Do not use me")]
+        def m
+        end
+      end
+
+      Foo(Int32).new.m
+    ), "Warning in line 8: Deprecated Foo(Int32)#m. Do not use me",
+      inject_primitives: false
+  end
+
+  it "detects deprecated generic class methods" do
+    assert_warning %(
+      class Foo(T)
+        @[Deprecated("Do not use me")]
+        def self.m
+        end
+      end
+
+      Foo(Int32).m
+    ), "Warning in line 8: Deprecated Foo(Int32).m. Do not use me",
+      inject_primitives: false
+  end
+
+  it "detects deprecated module methods" do
+    assert_warning %(
+      module Foo
+        @[Deprecated("Do not use me")]
+        def self.m
+        end
+      end
+
+      Foo.m
+    ), "Warning in line 8: Deprecated Foo.m. Do not use me",
+      inject_primitives: false
+  end
+
+  it "ignore deprecation excluded locations" do
+    with_tempfile("check_warnings_excludes") do |path|
+      FileUtils.mkdir_p File.join(path, "lib")
+
+      # NOTE tempfile might be created in symlinked folder
+      # which affects how to match current dir /var/folders/...
+      # with the real path /private/var/folders/...
+      path = File.real_path(path)
+
+      main_filename = File.join(path, "main.cr")
+      output_filename = File.join(path, "main")
+
+      Dir.cd(path) do
+        File.write main_filename, %(
+          require "./lib/foo"
+
+          bar
+          foo
+        )
+        File.write File.join(path, "lib", "foo.cr"), %(
+          @[Deprecated("Do not use me")]
+          def foo
+          end
+
+          def bar
+            foo
+          end
+        )
+
+        compiler = Compiler.new
+        compiler.warnings = Warnings::All
+        compiler.warnings_exclude << Crystal.normalize_path "lib"
+        compiler.prelude = "empty"
+        result = compiler.compile Compiler::Source.new(main_filename, File.read(main_filename)), output_filename
+
+        result.program.warning_failures.size.should eq(1)
+      end
+    end
+  end
+
+  it "errors if invalid argument type" do
+    assert_error %(
+      @[Deprecated(42)]
+      def foo
+      end
+      ),
+      "Error in line 3: first argument must be a String"
+  end
+
+  it "errors if too many arguments" do
+    assert_error %(
+      @[Deprecated("Do not use me", "extra arg")]
+      def foo
+      end
+      ),
+      "Error in line 3: wrong number of deprecated annotation arguments (given 2, expected 1)"
+  end
+
+  it "errors if missing link arguments" do
+    assert_error %(
+      @[Deprecated(invalid: "Do not use me")]
+      def foo
+      end
+      ),
+      "Error in line 3: too many named arguments (given 1, expected maximum 0)"
+  end
+end
--- a/src/compiler/crystal/codegen/warnings.cr
+++ b/src/compiler/crystal/codegen/warnings.cr
@@ -1,0 +1,91 @@
+module Crystal
+  struct DeprecatedAnnotation
+    getter message : String?
+
+    def initialize(@message = nil)
+    end
+
+    def self.from(ann : Annotation)
+      args = ann.args
+      named_args = ann.named_args
+
+      if named_args
+        ann.raise "too many named arguments (given #{named_args.size}, expected maximum 0)"
+      end
+
+      message = nil
+      count = 0
+
+      args.each do |arg|
+        case count
+        when 0
+          arg.raise "first argument must be a String" unless arg.is_a?(StringLiteral)
+          message = arg.value
+        else
+          ann.wrong_number_of "deprecated annotation arguments", args.size, "1"
+        end
+
+        count += 1
+      end
+
+      new(message)
+    end
+  end
+
+  class Def
+    def short_reference
+      case owner
+      when Program
+        "top-level #{name}"
+      when .metaclass?
+        "#{owner.instance_type}.#{name}"
+      else
+        "#{owner}##{name}"
+      end
+    end
+  end
+
+  class CodeGenVisitor
+    def check_call_to_deprecated_method(node : Call)
+      return unless @program.warnings.all?
+
+      if (ann = node.target_def.annotation(@program.deprecated_annotation)) &&
+         (deprecated_annotation = DeprecatedAnnotation.from(ann))
+        return if ignore_warning_due_to_location(node.location)
+
+        message = deprecated_annotation.message
+        message = message ? " #{message}" : ""
+
+        full_message = node.warning "Deprecated #{node.target_def.short_reference}.#{message}"
+
+        @program.warning_failures << full_message
+      end
+    end
+
+    private def ignore_warning_due_to_location(location : Location?)
+      return false unless location
+
+      filename = location.original_filename
+      return false unless filename
+
+      return @program.warnings_exclude.any? do |path|
+        filename.starts_with?(path)
+      end
+    end
+  end
+
+  class Command
+    def report_warnings(result : Compiler::Result)
+      return if result.program.warning_failures.empty?
+
+      result.program.warning_failures.each do |message|
+        STDERR.puts message
+      end
+      STDERR.puts "A total of #{result.program.warning_failures.size} warnings were found."
+    end
+
+    def warnings_fail_on_exit?(result : Compiler::Result)
+      result.program.error_on_warnings && result.program.warning_failures.size > 0
+    end
+  end
+end
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -64,7 +64,10 @@ class Crystal::Command
       init
     when "build".starts_with?(command)
       options.shift
-      build
+      result = build
+      report_warnings result
+      exit 1 if warnings_fail_on_exit?(result)
+      result
     when "play".starts_with?(command)
       options.shift
       playground
@@ -168,7 +171,9 @@ class Crystal::Command
   private def run_command(single_file = false)
     config = create_compiler "run", run: true, single_file: single_file
     if config.specified_output
-      config.compile
+      result = config.compile
+      report_warnings result
+      exit 1 if warnings_fail_on_exit?(result)
       return
     end
 
@@ -177,6 +182,9 @@ class Crystal::Command
     result = config.compile output_filename
 
     unless config.compiler.no_codegen?
+      report_warnings result
+      exit 1 if warnings_fail_on_exit?(result)
+
       execute output_filename, config.arguments, config.compiler
     end
   end
@@ -197,7 +205,7 @@ class Crystal::Command
     {config, result}
   end
 
-  private def execute(output_filename, run_args, compiler)
+  private def execute(output_filename, run_args, compiler, *, error_on_exit = false)
     time? = @time && !@progress_tracker.stats?
     status, elapsed_time = @progress_tracker.stage("Execute") do
       begin
@@ -226,7 +234,7 @@ class Crystal::Command
     end
 
     if status.normal_exit?
-      exit status.exit_code
+      exit error_on_exit ? 1 : status.exit_code
     else
       case status.exit_signal
       when ::Signal::KILL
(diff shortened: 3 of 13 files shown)


j8r:
Related to https://github.com/crystal-lang/crystal/issues/7027
In my opnion it would be better to have the deprecation warnings by default.
Another thing, why not using the stdlib logger? Then, having an argument to set the loglevel of the compiler.

spec/compiler/codegen/quality_check_spec.cr
```diff
@@ -0,0 +1,117 @@
...
+        def self.m
+        end
+      end
+
+      Foo.m
+    ), "Warning in line 8: Deprecated Foo.m. Do not use me",
+    inject_primitives: false
+  end
+
+  it "errors if too string argument" do
```
j8r:
```suggestion
  it "errors if invalid argument type" do
```

bcardiff:
@j8r I think that either

* you care about future compatibility (ie: you want build to fail on the usage of deprecated message) 
* or not (ie: you don't even care to know about it).

Having a configurable error level for this check opens the door to compiler messages that will bother the output.

j8r:
For example in Rust there are [Lint levels](https://doc.rust-lang.org/rustc/lints/levels.html), I don't think this warnings are useless to not print them by default.
Else, I'm afraid only few aware people will benefit from it.

ysbaddaden:
I'm supportive of adding support for proper deprecations, I think the annotation is good, but why are they opt-in? If I don't actively run with `--check` then I can totally miss a deprecation warning, and will later be faced with a broken compile.

Isn't `check` quite a broad term? It's not clear what it will actually check by its name alone. Unless there is a plan to check for different things later on?

I wonder about the failing status code, too. It will, eventually, fail to compile, but for now it did compile.

Maybe a `--skip-deprecate` would be more helpful and self explanatory?

bcardiff:
> Isn't check quite a broad term? It's not clear what it will actually check by its name alone. Unless there is a plan to check for different things later on?

No strict plan, but other checks I have in mind are deprecation of language constructs (probably `with ... yield`). And maybe some precision issues could appear. But what I don't like about detecting potential issues in a code base is that then we would need pragmas or something to allow them in specific situations. So I would stick initially regarding future compatibility of the code.

> I wonder about the failing status code, too. It will, eventually, fail to compile, but for now it did compile.

There is an `exit 1` in case warnigns were found, it should not compile.

Sija:
> No strict plan, but other checks I have in mind are deprecation of language constructs (probably `with ... yield`). And maybe some precision issues could appear. But what I don't like about detecting potential issues in a code base is that then we would need pragmas or something to allow them in specific situations. So I would stick initially regarding future compatibility of the code.

Wouldn't extending (and perhaps integrate into the CI pipeline) [Ameba](https://github.com/veelenga/ameba) be more beneficial?

bcardiff:
@Sija ameba works at a syntax level so these checks can't be detected with it's current scope.

asterite:
My thoughts on this (similar to those of @ysbaddaden):
- Why it's a `--check` flag and not a `check` (or a better name) command? I can't see why I'd like to build a program or run specs **and** also check that I don't use deprecated things. I'd only want to do one of those things.
- `check` sounds too broad. 
- This is an optional thing to do in your project, so "if you want to use it, you use it, otherwise you don't" initially sounds fine... until we remember that Crystal will compile all of the code at once, including your code and that of libraries. If a library doesn't care about deprecations they will pop up on your checks and there's nothing you can do about it. Well, you can bug the library's author to fix those deprecations. But I think a much better way to accomplish this is to **always** show these deprecation warnings. Because you can't remove these warnings, you'll be almost forced to fix them right away. That way libraries will stop using deprecated features as soon as possible. And note that it's just a warning: the compile output will be polluted, but your code will still compile and run fine. But because it's annoying you will most likely get rid of that (and then you could have `--check` or `--treat-warnings-as-errors` flag).

bcardiff:
What about having `--warnings=emit|error|ignore` and `--ignore-warnings-from=paths` (in `build` and `spec`) to allow the following use cases:

* (default) `--warnings=emit --ignore-warnings-from=lib`. Bother the author with respect it's app/shard code but let it compile them.
* `--warnings=error` good for CI since it will change the exit status
* `--warnings=ignore` turn warnings off

konovod:
There could be times when you just can't fix some warning (the feature doesn't work correctly in current compiler release, or you have to support previous release, or you are using old unsupported shard and don't have a knowledge\time to fix it). These situations are, of course, temporary, so it is ok to have warnings on by default, but it would be nice to have an ability to disable some on them - either using a compiler switch or by annotating in code.

ysbaddaden:
@bcardiff your proposal sounds good.

Note: I like how Rust as an optional `since` parameter ([Rust RFC 1270](https://rust-lang.github.io/rfcs/1270-deprecation.html)), that improves the message and gives a reference version to look at for the deprecation:
```rust
#[deprecated(since="1.2.3" note="use bar() instead")]
fn foo() {}

// => warning: foo() is marked deprecated as of version 1.2.3
//    note: use bar() instead
```

bcardiff:
@ysbaddaden But the `since` should be the version each package I guess that that information is not available in a uniform way. For now the deprecated annotation can accept only the message and we can extend that later. The `since:` can be simulated by

```crystal
  {% if compare_versions(Crystal::VERSION, "0.29.0-0") >= 0 %}
    @[Deprecated("Lorem ipsum")]
  {% end %}
```

I guess the `since:` only make sense if we want to deprecate this in a future version that is not the very next one. Otherwise, the deprecation could just be commited. Right?

asterite:
Unless that `since` is used programmatically somewhere, I think just doing `@[Deprecated("Deprecated since .... Please use ...")]` is the same.

straight-shoota:
Giving a time frame of the deprecation should always help. But I suppose it would even be better to announce the future version in which the feature is scheduled to be removed.

straight-shoota:
@asterite Even if not used programmatically, allowing separate keys helps to provide standardized information in a uniform manner and can be used for example in API docs.

bcardiff:
The following pattern can be used to schedule a deprecation and removal.

```crystal
{% if compare_versions(Crystal::VERSION, "0.29.0-0") <= 0 %}
  {% if compare_versions(Crystal::VERSION, "0.28.0-0") >= 0 %}
    @[Deprecated("Lorem ipsum")]
  {% end %}
  def foo
    # ...
  end
{% end %}
```

We can add additional information but I would prefer to find a way to enforce that semantic if the information is available. But I think we can start without them.

Without a way to know which is the package version we will need to inform it to the annotation.

```crystal
@[Deprecated("Lorem ipsum")]
def foo
end

@[Deprecated(note: "Lorem ipsum")]
def foo
end

@[Deprecated(note: "Lorem ipsum", version: Crystal::VERSION, since: "0.28.0")]
def foo
end

@[Deprecated(note: "Lorem ipsum", version: Crystal::VERSION, since: "0.28.0", drop: "0.29.0")]
def foo
end
```

Again, let's start with the note and the [cli options](https://github.com/crystal-lang/crystal/pull/7596#issuecomment-476905254) and leave this ideas for the future.

straight-shoota:
> But I think we can start without them.

Definitely. We can ship a MVP without that and collect experience on how it can be used and what features we're missing.

bcardiff:
The last push implements https://github.com/crystal-lang/crystal/pull/7596#issuecomment-476905254 for the run, build and spec commands.

It's ready for another round of reviews.

src/compiler/crystal/command/format.cr
```diff
@@ -72,29 +72,19 @@ class Crystal::Command
...
-      includes = normalize_paths includes
-      excludes = normalize_paths excludes
+      includes = includes.map { |p| Crystal.normalize_path p }
+      excludes = excludes.map { |p| Crystal.normalize_path p }
       excludes = excludes - includes
       if files.empty?
         files = Dir["./**/*.cr"]
       else
-        files = normalize_paths files
+        files = files.map { |p| Crystal.normalize_path p }
```
Sija: Why not move `#map { |path| Crystal.normalize_path path }` into the `normalize_paths` helper instead? :)

bcardiff: 'cause the atomic version was needed in the other command and doing another overload for a map was not worth it IMO.

spec/compiler/codegen/warnings_spec.cr
```diff
@@ -0,0 +1,157 @@
+require "../spec_helper"
+
+describe "Code gen: warnings" do
+  it "detects top level depreacted methods" do
```
ysbaddaden: typo: deprecated (same below).

src/compiler/crystal/codegen/warnings.cr
```diff
@@ -0,0 +1,93 @@
...
+        full_message = node.warning "Deprecated #{node.target_def.short_reference}.#{message}"
+
+        @program.warnings_failures << full_message
+      end
+    end
+
+    private def ignore_warning_due_to_location(location : Location?)
+      return false unless location
+
+      filename = location.filename
```
asterite: Maybe use `location.original_filename` to take macro expansions into account.

src/compiler/crystal/command.cr
```diff
@@ -511,9 +520,30 @@ class Crystal::Command
...
+      compiler.warnings = case w
+                          when "emit"
+                            Crystal::Warnings::Emit
+                          when "error"
+                            Crystal::Warnings::Error
+                          when "ignore"
+                            Crystal::Warnings::Ignore
+                          else
+                            error "--warnings should be emit, error, or ignore"
+                            raise "unreachable"
```
asterite: It's strange that this line is needed.

src/compiler/crystal/command/format.cr
```diff
@@ -72,29 +72,19 @@ class Crystal::Command
     )
       @format_stdin = files.size == 1 && files[0] == "-"
 
-      includes = normalize_paths includes
-      excludes = normalize_paths excludes
+      includes = includes.map { |p| Crystal.normalize_path p }
```
asterite: Why not `map!`? Same in other places below.

src/compiler/crystal/command.cr
bcardiff: Do you mean the `raise "unreachable"`? 

The `error` function is not a `NoReturn` it depends on an argument value whether or not it does an `exit`.

asterite: Aah... right 👍

src/compiler/crystal/semantic/ast.cr
```diff
@@ -13,6 +13,21 @@ module Crystal
       ::raise exception_type.for_node(self, message, inner)
     end
 
+    def warning(message, inner = nil, exception_type = Crystal::TypeException)
+      # TODO extract message formatting from exceptions
+      String.build do |io|
+        exception = exception_type.for_node(self, message, inner)
+        io << "Warning "
+        exception.append_to_s(nil, io)
+        # Errors of macro will frist include the trace of the macro
```
r00ster91:
```suggestion
        # Macro errors will first include the trace of the macro
```

src/compiler/crystal/semantic/ast.cr
```diff
@@ -13,6 +13,21 @@ module Crystal
...
+        # expansions, and then the warning message.
+        # In other warnign messages the code snippet includes a newline
```
r00ster91:
```suggestion
        # In other warning messages the code snippet includes a newline
```

src/compiler/crystal/semantic/ast.cr
```diff
@@ -13,6 +13,21 @@ module Crystal
...
+        if exception.@filename.is_a?(VirtualFile)
+          io << "\n"
```
r00ster91:
```suggestion
          io.puts
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1049,6 +1049,12 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
         node.returns_twice = true
       when @program.raises_annotation
         node.raises = true
+      when @program.deprecated_annotation
+        # Check a DeprecatedAnnotation can be built
+        # there is no need to store it, but enforce
```
r00ster91:
```suggestion
        # There is no need to store it, but enforcing
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1049,6 +1049,12 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
...
+        # arguments make sense here.
```
r00ster91:
```suggestion
        # arguments makes sense here.
```

src/compiler/crystal/codegen/warnings.cr
```diff
@@ -0,0 +1,93 @@
...
+
+      new(message)
+    end
+  end
+
+  class Def
+    def short_reference
+      case owner
+      when Program
+        "top level #{name}"
```
r00ster91:
```suggestion
        "top-level #{name}"
```

src/compiler/crystal/program.cr
```diff
@@ -115,6 +115,15 @@ module Crystal
...
     property codegen_target = Config.default_target
 
+    # How to treat the generated warnings.
+    property warnings : Warnings = Warnings::Emit
+
+    # Paths to ignore for warnings detection.
+    property warnings_exclude : Array(String) = [] of String
+
+    # Detected warnings failures
+    property warnings_failures = [] of String
```
r00ster91: This should be `warning_failures` everywhere.

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -1049,6 +1049,12 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
         node.returns_twice = true
       when @program.raises_annotation
         node.raises = true
+      when @program.deprecated_annotation
+        # Check a DeprecatedAnnotation can be built
```
r00ster91: I'm not exactly sure what this comment means but perhaps
```suggestion
        # Check whether the DeprecatedAnnotation can be built.
```

bcardiff: It means to check the arguments are the expected ones. Annotations are a bit loosely defined. There is currently no way to declare which arguments are expected.

bcardiff:
Last comment and naming feedback incorporated. Thanks @j8r.
Rebased & squashed on master. 
Ready to merge after CI is happy :shipit: .

bcardiff:
Oh, I will give some time and see if @straight-shoota or @ysbaddaden have more feedback before merging.

src/compiler/crystal/command.cr
```diff
@@ -511,9 +520,30 @@ class Crystal::Command
       @color = false
       compiler.color = false
     end
+    setup_compiler_warning_options(opts, compiler)
     opts.invalid_option { }
   end
 
+  private def setup_compiler_warning_options(opts, compiler)
+    compiler.warnings_exclude << Crystal.normalize_path "lib"
+    opts.on("--warnings emit|error|ignore", "How to treat warning. (default: emit)") do |w|
```
Sija:
```suggestion
    opts.on("--warnings emit|error|ignore", "How to treat warnings. (default: emit)") do |w|
```

src/compiler/crystal/command.cr
```diff
@@ -511,9 +520,30 @@ class Crystal::Command
...
     end
+    setup_compiler_warning_options(opts, compiler)
     opts.invalid_option { }
   end
 
+  private def setup_compiler_warning_options(opts, compiler)
+    compiler.warnings_exclude << Crystal.normalize_path "lib"
+    opts.on("--warnings emit|error|ignore", "How to treat warnings. (default: emit)") do |w|
+      compiler.warnings = case w
+                          when "emit"
```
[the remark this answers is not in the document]
asterite: We should probably not try to invent the wheel. I think everywhere else in the world there are two options: "warnings on or off", and "treat warnings as errors".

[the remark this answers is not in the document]
ysbaddaden: You're maybe right. I usually associate `emit` as the output format (object file, LLVM IR, JSON...).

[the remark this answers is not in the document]
bcardiff: Ok, do we prefer to start with `--warnings all|none` and `--error-on-wargnings` in case the 
warnings categories increase in the future so we can have `--warnings deprecated,X,Y,Z` to specify which warnings we want?

Or seems enough to go with `--warnings warn|error|ignore`?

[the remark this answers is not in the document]
j8r: Having warning categories is great. For instance on `pylint`, there is a `-d, --disable=` to disable messages/reports. I think this will be appropriate to copy this opt-in by default approach, for example having `disable-warnings` , and only disable warnings we don't want.

There is also a `disable=all` to disable all messages/reports, and `enable` to enable some.
For example `disable=all enable=classes`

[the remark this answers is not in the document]
asterite: I think `--warnings all|none` with `--error-on-warnings` is fine. Then we could eventually list specific warnings, as you suggest.

bcardiff:
The last push implements the `--warnings all|none --error-on-warnings` CLI options.
Although `Crystal::Warnings` could be a `@[Flags]` there is currently no need.

Once reviewed I will rebase + squash for merging.

cc: @straight-shoota @asterite

j8r:
It would be also good to have a way to send warnings without relying only on annotations. It's not always possible to use them, for example when changing a syntax in the compiler.
One can say we have already `{{ puts "deprecated notice"  }}`, but that's not really a clean API for deprecation warnings.

bcardiff:
@j8r feel free to open an issue to discuss the design around that. Probably a deprecated macro method would work to check other constructs like deprecating specific C library version.

The compiler itself has everything needed to add ad-hoc code to detect usage of language constructs to be deprecated in the future.

straight-shoota:
Do we want to include `--check` in the compiler build and spec runs on CI?

bcardiff:
In the compiler specs we can. In the std lib it could only be when running against the compiled compiler.

Once 0.28 is released, definitely!

Issue #6498, opened 2018-08-07
suggestion: add sort method to Slice and StaticArray


chenkovsky:
currently, we cannot sort Slice or StaticArray directly.

jhass:
Could you expand a bit on your usecases?

chenkovsky:
@jhass  when interacting with C API, a pointer is passed from C API which represents an array. I want to do something on this pointer without create a new array.

NIFR91:
Creating a new array from a Slice or StaticArray just to sort is unnecessary, adding sort to the slice object sound good to me.

asterite:
I think we could implement this, the algorithm in Array works on `Pointer` so it would be easy to move this code to `Slice` or maybe `Pointer` and use it in `Array`, `Slice` and `StaticArray`.

RX14:
@asterite Definitely slice, constructing a slice from a pointer is "free" and it provides a far better interface for working on what is conceptually a slice

Maroo-b:
@straight-shoota @asterite I would like to work on this, should I add the following methods to `Slice` only?

- `def sort(&block : T, T -> Int32) : Slice(T)`
- `def sort! : Slice(T)`
- `def sort!(&block : T, T -> Int32) : Slice(T)`
- `def sort_by(&block : T -> UNDERSCORE) : Slice(T)`
- `def sort_by!(&block : T -> UNDERSCORE) : Slice(T)`

straight-shoota:
Yeah, that seems to be the course of action :+1:
While doing that, I'd also suggest to move the implementation methods (`intro_sort` etc.) to a separate file, for example `src/sort.cr` or `src/slice/sort.cr` to better separate this feature for better code readability.

Issue #5173, opened 2017-10-24
Array#slice


coderhs:
Array#slice is a popular feature in ruby https://ruby-doc.org/core-2.4.2/Array.html#method-i-slice would this be a welcomed feature. I know that we can achieve the same directly but this is an API that ruby developers are used it.

PS: Would sending PR with methods that are in Ruby STD but not in crystal be welcomed? if so I would like to help out by sending more of them.

akzhan:
Crystal propose it's own https://crystal-lang.org/api/0.23.1/Slice.html, and you allowed to get it from arrays etc. It looks more powerful feature (for me).

jhass:
Ruby's `Array#slice` really is just an alias for `Array#[]`. If you look at our `Array#[]` you'll find it already covers most, if not all, of the same usecases.

RX14:
In the ruby docs it says that `[]` and `slice` are just aliases. We try pretty hard not to introduce aliases into crystal, so I don't see any way this would be accepted.

coderhs:
@akzhan How do you use the Slice you mentioned with Array(Int32) or such arrays? I am planning to add this method to `crystal_on_steroids`, if slice is better then I would love to use that to implement. `Array#slice`.

RX14:
@coderhs `Array#slice` is *literally* an alias for `Array#[]`. There's no need to add this method because it already exists under another name.

coderhs:
@RX14 I agree. 

Right now I am just now curious on how to use `Slice` to slice an array. From the example I find that we need to enter elements one by one to a pointer and then slice it, or is there a way in which it can accepts an array and slice it.

RX14:
@coderhs you can't. We removed `Array#to_slice` because it would cause a segfault if you kept a reference to the slice returned from that method and the array was resized. @akzhan is incorrect, `Slice` has nothing to do with this.

crysbot:
This issue has been mentioned on **Crystal Forum**. There might be relevant details there:

https://forum.crystal-lang.org/t/indexable-mutable-methods-acting-on-part-of-the-sequence/7911/2

PR #7597, opened 2019-03-26
Add sort methods to Slice

Maroo-b:
- Issue: https://github.com/crystal-lang/crystal/issues/6498

This PR depends on this fix: https://github.com/crystal-lang/crystal/pull/7591

I extracted the sort logic from `Array` to a module. If this approach is approved, I'll try to add some docs for the module (or if someone has more skills with writing docs :) )


asterite:
You can now rebase against master :-)

straight-shoota:
What's the purpose of `Sortable` module?  It doesn't define an actual interface, just some class methods  can easily be placed in an existing namespace like `Slice` or `Array`. Only `intro_sort` is publicly visible at all (that might even be questionable).

yxhuvud:
Is it possible to move `sort`, `sort!`, `sort_by` etc into Sortable?

straight-shoota:
@yxhuvud Yes that should work. The exclamation methods need a tiny adjustment, but `#to_unsafe` returns the pointer in both `Array` and `Slice`.

But I'm not sure if this module would be much of a benefit. Duplicating the code for sort methods is one thing, but that's not really an issue. The question is, is having a `Sortable` interface useful to users?

Maroo-b:
@straight-shoota
I extracted the sort logic from Array. Should I just rename the module to `Sort` for example and just require it in `Array` and `Slice`? or do you have another suggestion?

asterite:
@Maroo-b Your changes (the `Sort` module) don't show up for some reason...

My thoughts:
- there's no need for a `Sort` or `Sortable` module that can be included. In fact there's no need for it to be generic, the `T` isn't used anywhere
- I agree that the `sort` helper methods (currently in `Sortable`) should exist somewhere where this logic can be shared, but let's not use global names like `Sort` or `Sortable` which can collide with user-defined types. I think just putting the helper methods in `Slice` or `Pointer::SortUtils` or `Pointer::SortHelpers` is good enough.

Maroo-b:
@asterite thank you for the feedback! 

Sure I agree, I'll move those methods to `Pointer::SortUtils` then :)

straight-shoota:
IMHO the name `Pointer::SortUtils` doesn't make much sense. The sort algorithm is implemented on pointers, but conceptually, it operates on collections.

I'm still not convinced that we need a separate namespace for these methods anyway, but if we do, I'd call it `Slice::Sort` or `Array::Sort`. Or maybe `Crystal::IntroSort`?

asterite:
The way I imagine it, the `sort` methods belong to `Slice`. Then Array simply does `Slice.new(to_unsafe, size).sort`.

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
...
+  # b = a.sort_by { |word| word.size }
+  # b # => Slice["fig", "pear", "apple"]
+  # a # => Slice["apple", "pear", "fig"]
+  # ```
+  def sort_by(&block : T -> _) : Slice(T)
+    dup.sort_by! { |e| yield(e) }
+  end
+
+  # Modifies `self` by sorting all elements. The given block is called for
+  # each element, then the comparison method #<=> is called on the object
```
r00ster91:
```suggestion
  # each element, then the comparison method `<=>` is called on the object
```

src/pointer/sort_utils.cr
```diff
@@ -0,0 +1,211 @@
+# :nodoc:
+#
```
r00ster91:
```suggestion
```

src/pointer/sort_utils.cr
```diff
@@ -0,0 +1,211 @@
...
+class Pointer::SortUtils
```
r00ster91: This can be a module since there are only class methods here.
```suggestion
module Pointer::SortUtils
```

src/pointer/sort_utils.cr
```diff
@@ -0,0 +1,211 @@
...
+  def self.intro_sort!(a, n)
```
r00ster91: Shouldn't this and the other method be protected too just like all the others?
```suggestion
  protected def self.intro_sort!(a, n)
```

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
     @pointer
   end
 
+  # Returns a new slice with all elements sorted based on the return value of
+  # their comparison method `#<=>`
```
r00ster91: There's no instance method `<=>` in this struct.
```suggestion
  # their comparison method `<=>`.
```

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
...
+  # a      # => Slice[3, 1, 2]
+  # ```
+  def sort : Slice(T)
+    dup.sort!
+  end
+
+  # Returns a new slice with all elements sorted based on the comparator in the
+  # given block.
+  #
+  # The block must implement a comparison between two elements *a* and *b*,
```
r00ster91: *This* style is usually used for arguments. But there's no argument named `a` or `b` here.
```suggestion
  # The block must implement a comparison between two elements a and b,
```

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
...
+  def sort(&block : T, T -> U) : Slice(T) forall U
+    {% unless U <= Int32? %}
+      {% raise "expected block to return Int32 or Nil, not #{U}" %}
+    {% end %}
+
+    dup.sort! &block
+  end
+
+  # Modifies `self` by sorting all elements based on the return value of their
+  # comparison method `#<=>`
```
r00ster91:
```suggestion
  # comparison method `<=>`.
```

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
...
+  def sort! : Slice(T)
+    Pointer::SortUtils.intro_sort!(to_unsafe, size)
+    self
+  end
+
+  # Modifies `self` by sorting all elements based on the comparator in the given
+  # block.
+  #
+  # The given block must implement a comparison between two elements
+  # *a* and *b*, where `a < b` returns `-1`, `a == b` returns `0`,
```
r00ster91:
```suggestion
  # a and b, where `a < b` returns `-1`, `a == b` returns `0`,
```

src/slice.cr
```diff
@@ -545,6 +546,106 @@ struct Slice(T)
...
+    {% unless U <= Int32? %}
+      {% raise "expected block to return Int32 or Nil, not #{U}" %}
+    {% end %}
+
+    Pointer::SortUtils.intro_sort!(to_unsafe, size, block)
+    self
+  end
+
+  # Returns a new array with all elements sorted. The given block is called for
+  # each element, then the comparison method #<=> is called on the object
```
r00ster91:
```suggestion
  # each element, then the comparison method `<=>` is called on the object
```

src/pointer/sort_utils.cr
```diff
@@ -0,0 +1,211 @@
...
+    return if n < 2
+    quick_sort_for_intro_sort!(a, n, Math.log2(n).to_i * 2)
+    insertion_sort!(a, n)
+  end
+
+  def self.intro_sort!(a, n, comp)
```
r00ster91:
```suggestion
  protected def self.intro_sort!(a, n, comp)
```

Maroo-b:
@straight-shoota @asterite just to confirm, should I move the sort logic to `Slice` and update `Array` to use Slice for sorting?
I'm concerned about creating a Slice instance just for sorting. is it fine?

asterite:
It's fine, Slice is a struct. And yes, please move the methods to Slice. Thank you!

straight-shoota:
Creating a struct is basically free of cost.

I'd suggest to add `Array#to_slice` as a shortcut for `Slice.new(to_unsafe, size)`. Might be useful as a public method, too.

bew:
@straight-shoota but we'd get back to https://github.com/crystal-lang/crystal/issues/5173#issuecomment-340286755 ?
> @RX14 wrote: We removed `Array#to_slice` because it would cause a segfault if you kept a reference to the slice returned from that method and the array was resized.

straight-shoota:
Looks great!

I'd prefer to put the sort methods in a separate file (e.g. `src/slice/sort.cr`) though for some kind of compartmentalization.

src/slice/sort.cr
```diff
@@ -0,0 +1,211 @@
+# :nodoc:
```
straight-shoota: This shouldn't have any effect.

bew:
What about moving the specs to `spec/std/slice/sort_spec.cr` ?

Maroo-b:
@bew it's an option, but I prefer to keep `Slice` specs in one file for the moment.

Maroo-b:
PR updated to resolve conflicts and apply changes from this PR: https://github.com/crystal-lang/crystal/pull/7639

Sija:
Seems like GTG 🚀

PR #7598, opened 2019-03-27
add some Time static methods: now_utc_to_unix*

ndudnicz:
Hi, I did some Time helpers that give the current utc timestamp in second `Time.now_utc_to_unix`, msecond `Time.now_utc_to_unix_ms` and second as `Float64` `Time.now_utc_to_unix_f`. Because sometimes I don't want to make an instance of Time when I only want the current timestamp.

Here's a simple benchmark:
```
$> cat main.cr
ITER = 10_000_000

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.utc.to_unix
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.utc.to_unix: " + (_end - _start).to_s + " ms"

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.utc.to_unix_ms
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.utc.to_unix_ms: " + (_end - _start).to_s + " ms"

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.utc.to_unix_f
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.utc.to_unix_f: " + (_end - _start).to_s + " ms"

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.now_utc_to_unix
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.now_utc_to_unix: " + (_end - _start).to_s + " ms"

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.now_utc_to_unix_ms
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.now_utc_to_unix_ms: " + (_end - _start).to_s + " ms"

_start = Time.utc.to_unix_ms
i = 0
while i < ITER
  Time.now_utc_to_unix_f
  i += 1
end
_end = Time.utc.to_unix_ms
puts ITER.to_s + " Time.now_utc_to_unix_f: " + (_end - _start).to_s + " ms"
```
```
$> ./repo/bin/crystal build main.cr
$> ./main
10000000 Time.utc.to_unix: 664 ms
10000000 Time.utc.to_unix_ms: 928 ms
10000000 Time.utc.to_unix_f: 732 ms
10000000 Time.now_utc_to_unix: 355 ms
10000000 Time.now_utc_to_unix_ms: 593 ms
10000000 Time.now_utc_to_unix_f: 433 ms
```

Regards

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -273,6 +273,13 @@ struct Time
   # :nodoc:
   MAX_SECONDS = 315537897599_i64
 
+  # epoch's timestamp in second since `0001-01-01 00:00:00`
+  EPOCH_SECONDS_TIMESTAMP = 62135596800_i64
+  EPOCH_SECONDS_TIMESTAMP_F64 = 62135596800_f64
+
+  # epoch's timestamp in msecond since `0001-01-01 00:00:00`
+  EPOCH_MSECONDS_TIMESTAMP = 62135596800000_i64
+
   # `DayOfWeek` represents a day of the week in the Gregorian calendar.
   #
   # ```
@@ -739,6 +746,36 @@ struct Time
       seconds: total_seconds - other.total_seconds,
       nanoseconds: nanosecond - other.nanosecond,
     )
+  end
+
+  # Returns an utc epoch current timestamp in second
+  #
+  #```
+  # Time.now_utc_to_unix # => 1452567845
+  #```
+  def self.now_utc_to_unix : Int64
+    seconds, nanoseconds = Crystal::System::Time.compute_utc_seconds_and_nanoseconds
+    seconds - EPOCH_SECONDS_TIMESTAMP
+  end
+
+  # Returns an utc epoch current timestamp in msecond
+  #
+  #```
+  # Time.now_utc_to_unix_ms # => 1452567845876
+  #```
+  def self.now_utc_to_unix_ms : Int64
+    seconds, nanoseconds = Crystal::System::Time.compute_utc_seconds_and_nanoseconds
+    (seconds * 1_000 + (nanoseconds / NANOSECONDS_PER_MILLISECOND)) - EPOCH_MSECONDS_TIMESTAMP
+  end
+
+  # Returns an utc epoch current timestamp in second as float64
+  #
+  #```
+  # Time.now_utc_to_unix_f # => 1452567845.876736
+  #```
+  def self.now_utc_to_unix_f : Float64
+    seconds, nanoseconds = Crystal::System::Time.compute_utc_seconds_and_nanoseconds
+    (seconds.to_f + (nanoseconds.to_f / 1e9)) - EPOCH_SECONDS_TIMESTAMP_F64
   end
 
   # Returns a copy of `self` with time-of-day components (hour, minute, second,


asterite:
Making an instance of `Time` shouldn't affect the performance because it's a struct and LLVM _should_ optimize things. Maybe using `Benchmark.ips` can give us a better idea of what's going on...

jhass:
For me your benchmark results in:
```
$ bin/crystal build bench.cr
$ ./bench
10000000 Time.utc.to_unix: 773 ms
10000000 Time.utc.to_unix_ms: 981 ms
10000000 Time.utc.to_unix_f: 745 ms
10000000 Time.now_utc_to_unix: 332 ms
10000000 Time.now_utc_to_unix_ms: 627 ms
10000000 Time.now_utc_to_unix_f: 428 ms
$ bin/crystal build bench.cr --release
$ ./bench
10000000 Time.utc.to_unix: 306 ms
10000000 Time.utc.to_unix_ms: 303 ms
10000000 Time.utc.to_unix_f: 299 ms
10000000 Time.now_utc_to_unix: 237 ms
10000000 Time.now_utc_to_unix_ms: 240 ms
10000000 Time.now_utc_to_unix_f: 247 ms
```
So the difference is much less significant in a release build, for your benchmark. Using `Benchmark.ips` in a release build:

```
       Time.utc.to_unix  33.96M ( 29.45ns) (± 1.73%)  0.0B/op   1.17× slower
    Time.utc.to_unix_ms  33.94M ( 29.46ns) (± 1.89%)  0.0B/op   1.17× slower
     Time.utc.to_unix_f  33.98M ( 29.43ns) (± 1.66%)  0.0B/op   1.16× slower
   Time.now_utc_to_unix  39.51M ( 25.31ns) (± 1.70%)  0.0B/op   1.00× slower
Time.now_utc_to_unix_ms  39.57M ( 25.27ns) (± 1.46%)  0.0B/op        fastest
 Time.now_utc_to_unix_f  39.50M ( 25.31ns) (± 1.43%)  0.0B/op   1.00× slower
``` 

So there is a fairly consistent overhead of about 4ns, might be about the overhead of an additional function call I guess, but seems negligible for standard library functionality. Maybe somebody find a way so this is inlined better but until then I would say let's have projects that depends on that kind of performance difference implement this on their own and vote to close this :)

straight-shoota:
Yeah, I agree. Having this method is not a huge improvement in performance as the benchmark show.

But nonetheless, thank you very much, @ndudnicz and welcome to the team =)

Remember to run benchmarks using `Benchmark.ips` and in release mode.

ndudnicz:
`might be about the overhead of an additional function call I guess` that's what I wanted to avoid, useless function calls. Even with `--release` compilation and `benchmark.ips`, it run ~10% faster.
Sure it's not game changing but I thought it could match with the `fast as C` paradigm.
I close it.
thx

asterite:
I just checked it and I think the main difference is that `Time.utc` uses the constants `Time::Location::UTC`, which in turns uses `Time::Zone::UTC`and there's also the constant `Time::UNIX_EPOCH`.

When you use a constant that can't be computed at compile-time (like a struct), each time you access a constant the runtime will verify whether it's initialized or not, and initialize it if not. This is because a constant might be used before the line where it's actually declared. I'm almost sure these checks are the ones causing the little slowdown.

I believe that we can optimize the compiler... I'm just not sure how yet. Maybe there is a way to detect whether a constant is never used before the line where it's declared, and in that case the checks will be gone.

asterite:
Also, "fast as C" means you can write a Crystal program that runs as fast as C: just use pointers, static array, avoid allocations, etc. But that doesn't mean that the standard library or programs written in Crystal will generally run "as fast as C". It just means you can get there. For example you could provide a shard with those optimized functions if you want. But from the std's point of view we'd like to keep it simple, and hope that compiler optimizations will improve this over time.

Issue #7599, opened 2019-03-27
Bad comment formatting on next line when there's a comment after def


r00ster91:
```cr
def method # this is a method
  # this is a comment
end
```
gets formatted to:
```cr
def method # this is a method
# this is a comment
end
```
But I expect nothing to happen.

PR #7604, opened 2019-03-30
Formatter - fix comment indentation

Maroo-b:
issue: https://github.com/crystal-lang/crystal/issues/7599
Note the variable `@inside_def` was a leftover from a previous cleanup that's why I remove it.
with this PR it's possible to have the correct comment indentation for this example:

```
# module comment
module Foo
  # class comment
  class Bar
    # method comment
    def method # this is a method
      # this is a comment
      if 1 == 1
        # block comment
      end
    end
  end
end

```

### DIFF
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -76,7 +76,6 @@ module Crystal
     @implicit_exception_handler_indent : Int32
     @last_write : String
     @exp_needs_indent : Bool
-    @inside_def : Int32
     @when_infos : Array(AlignInfo)
     @hash_infos : Array(AlignInfo)
     @assign_infos : Array(AlignInfo)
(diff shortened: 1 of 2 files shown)


asterite:
I'm not sure the fix is correct. I think when formatting the newline after a def the new indentation should be passed. I'll try to fix this.

Maroo-b:
@asterite you are right, this fix messes up with comments inside method blocks. (if, each, ...)
I'm still trying to understand how the formatter is working :/

src/compiler/crystal/tools/formatter.cr
```diff
@@ -4342,7 +4336,9 @@ module Crystal
       while @token.type == :COMMENT
         empty_line = @line_output.to_s.strip.empty?
         if empty_line
-          write_indent if needs_indent
+          if needs_indent
+            write_indent
+          end
```
Sija: This change is redundant.

asterite:
@Maroo-b Thank you for this! But I have more of the formatter context in my mind to know the least amount of changes to make it work. I created a PR to fix this and another issue too: #7605

Maroo-b:
I'm closing this PR in favor of: https://github.com/crystal-lang/crystal/pull/7605

PR #7605, opened 2019-03-30
Fix a couple of formatter issues

### LINKED ISSUE #7599: Bad comment formatting on next line when there's a comment after def

### LINKED ISSUE #7600: Bad formatting for comma enum with comments
```cr
enum E
  A, # hello
  B, # hello
  C, # hello
end
```
gets formatted to:
```cr
enum E
  A,  # hello
B,    # hello
C,    # hello
end
```
But I expect nothing to happen.

asterite:
Fixes #7599
Fixes #7600

### DIFF
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -2061,9 +2061,16 @@ module Crystal
 
       # This is the case of an enum member
       if node.name[0].ascii_uppercase? && @token.type == :","
-        write ", "
-        next_token_skip_space
-        @exp_needs_indent = @token.type == :NEWLINE
+        write ","
+        next_token
+        @lexer.skip_space
+        if @token.type == :COMMENT
+          write_comment
+          @exp_needs_indent = true
+        else
+          write " "
+          @exp_needs_indent = @token.type == :NEWLINE
+        end
       end
 
       false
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/tools/formatter.cr
```diff
@@ -2058,9 +2061,16 @@ module Crystal
 
       # This is the case of an enum member
       if node.name[0].ascii_uppercase? && @token.type == :","
```
j8r: Is there any advantage of supporting both the comma and space as member delimiters in enums?

asterite: That space is supported is a bug. Only comma or newline should be supported. Please open a bug for this.

j8r: Excluding the comma, why supporting both the comma and new line (and therefore the semicolon)?
```cr
enum E
  A; B; C
  D, E, F
end
```

asterite: I copied the comma from C#. I wouldn't mind dropping it, but it's a breaking change (I don't mind either)

bcardiff:
Dropping the comma from the formatter and from the parser can be done in another PR

PR #7610, opened 2019-04-01
Add HTTP::Request#remote_address

### LINKED ISSUE #453: Add remote ip field to Http::Request

asterite:
This PR adds a `remote_address` **property** to `HTTP::Request`.

Some details:
- the property is of type `String?`. It can be `nil` if the address can't be determined, but this usually won't happen. It's also nilable because for `HTTP::Client` it doesn't make sense. It's `String` and not a structured object like `IPAddress` because a middleware might want to overwrite this property, for example using the `X-Forwarded-By` or `X-Real-IP` headers. Also, this field is apparently commonly used for logging so it being a String is probably fine and the most flexible option.
- I'm not super convinced about using `responds_to?(:remote_address)` to determine this, I'd maybe like a module that defines this as an abstract method, but we can change the implementation later.
- I know I didn't document the `local_address` and `remote_address` getters of `OpenSSL::SSL::Socket`, but only because the underlying `IO` is not necessarily a `IPSocket`, so the type of those getters isn't `IPAddress`. Maybe we can assume the underlying `IO` is `IPSocket`, otherwise raise in this case. But this is something that we can improve later.
- If you have a better idea for the doc comments of `HTTP::Request#remote_address`, let me know

The idea here is to define an API that's usable and we won't modify later on. I think adding `HTTP::Request#remote_address : String?` is enough for this. We can redefine the implementation details I mention above in subsequent PRs. 

Fixes #453
Related to #5784
Supersedes #7302

### DIFF
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -20,6 +20,15 @@ class HTTP::Request
   @query_params : Params?
   @uri : URI?
 
+  # The network address that sent the request to an HTTP server.
+  #
+  # `HTTP::Server` will try to fill this property, and its value
+  # will have a format like "IP:port", but this format is not guaranteed.
+  # Middlewares can overwrite this value.
+  #
+  # This property is not used by `HTTP::Client`.
+  property remote_address : String?
+
   def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1")
     @headers = headers.try(&.dup) || Headers.new
     self.body = body
@@ -99,7 +108,13 @@ class HTTP::Request
     return BadRequest.new unless HTTP::SUPPORTED_VERSIONS.includes?(http_version)
 
     HTTP.parse_headers_and_body(io) do |headers, body|
-      return new method, resource, headers, body, http_version
+      request = new method, resource, headers, body, http_version
+
+      if io.responds_to?(:remote_address)
+        request.remote_address = io.remote_address.try &.to_s
+      end
+
+      return request
     end
 
     # Malformed or unexpectedly ended http request
--- a/spec/std/http/server/server_spec.cr
+++ b/spec/std/http/server/server_spec.cr
@@ -835,6 +835,51 @@ module HTTP
       end
     end
 
+    describe "#remote_address" do
+      it "for http server" do
+        remote_address = nil
+
+        server = Server.new do |context|
+          remote_address = context.request.remote_address
+        end
+
+        tcp_server = TCPServer.new("127.0.0.1", 0)
+        server.bind tcp_server
+        address1 = tcp_server.local_address
+
+        run_server(server) do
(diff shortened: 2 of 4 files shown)


asterite:
Also, we should probably add some way to parse a network address that looks like "ip:port" into `{String, Int32}?`. Right now we have `Socket::IPAddress.parse(URI)` but you have to pass something like "tcp://ip:port". But here we only care about the `"ip:port"` part (would work for IPv6 too) and getting those values as IP (string) and port (int).

But again, this can be done later on.

src/http/request.cr
```diff
@@ -20,6 +20,15 @@ class HTTP::Request
   @query_params : Params?
   @uri : URI?
 
+  # The network address that sent the request to an HTTP server.
+  #
+  # `HTTP::Server` will try to fill this property, and it's value
```
r00ster91:
```suggestion
  # `HTTP::Server` will try to fill this property, and its value
```

asterite: Thanks!

asterite:
Rebased against master. I noticed `spec/std/http/spec_helper` didn't include the general `spec/std/spec_helper` which contains `datapath` so I added that too.

asterite:
Also thank you @omarroth for the original code, which I used almost as is. It's just that this needed a bit more forward pushing from us core team members and having it on our side makes it easier to rebase, edit, etc.

omarroth:
No worries, thanks @asterite!

rdp:
I noticed today that determining remote_address "up front" for  request makes this benchmark:

https://gist.github.com/jgaskins/ba74aaca0ce45f714caaa7571703beba [1]

2.5% slower.(12500 -> 12225 req/s) single threaded, and about 5% slower multi-threaded.  Maybe there's a way to avoid the slowdown?

[1] https://forum.crystal-lang.org/t/multithreaded-crystal-initial-thoughts/1089

straight-shoota:
What exactly are you comparing? Crystal Vs Go? Different Crystal versions?

rdp:
Crystal versus crystal with or without this patch applied:
```
diff --git a/src/http/request.cr b/src/http/request.cr
index f942c33f3..ade289eb1 100644
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -110,7 +110,7 @@ class HTTP::Request
       request = new line.method, line.resource, headers, body, line.http_version, internal: nil
 
       if io.responds_to?(:remote_address)
-        request.remote_address = io.remote_address.try &.to_s
+        #request.remote_address = io.remote_address.try &.to_s
       end
 
       return request
```

First up any "small" HTTP server, here's one: https://gist.github.com/rdp/b136f6a998919fe0eae39dc7990d0ba0
then load test it with wrk: `$ wrk  http://localhost:9000`
Thanks!

PR #6768, opened 2018-09-22
Optimized Images by the GitHub [ImgBot] (2.89kb -> 1.86kb (35.49%))

AYIDouble:
[ImgBot] optimizes images
/src/compiler/crystal/tools/playground/public/icon.png -- 2.89kb -> 1.86kb (35.49%)

PR #7616, opened 2019-04-03
Replace PNG icon with optimized SVG for playground

straight-shoota:
Closes #6768


bcardiff (dismissed):
🐛 The `` ends with size 0x0 (tested in Chrome and Firefox)

wontruefree:
Does this change the Browsers that can run this?
I would imagine we are targeting Firefox, Chrome, Safari and Opera right?

straight-shoota:
SVG support has been in all relevant browsers for 10 years or so.

wontruefree:
I thought so but I was not sure and thought I would ask.

oprypin (approved):

PR #7617, opened 2019-04-03
Mention git pre-commit hook in CONTRIBUTING.md

straight-shoota:
This adds the git pre-commit hook proposed in #7145 to `CONTRIBUTING.md`. The hook is provided as a gist instead of directly inside the repository.

Gist: https://gist.github.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6

Closes #7145

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -169,6 +169,17 @@ your own projects; a not so clean history is to prefer once a commit landed ther
 git push -f origin my_feature_branch
 ```
 
+## Git pre-commit hook
+
+Code submitted to this repository should be formatted according to `crystal tool format`.
+A pre-commit hook can be installed into the local git repo to ensure the formatter validates every commit: https://gist.github.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6
+
+Install the pre-commit hook:
+
+```sh
+curl https://gist.githubusercontent.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6/raw/pre-commit > .git/hooks/pre-commit && chmod +x .git/hooks/pre-commit
+```
+
 ## Code of Conduct
 
 Please note that this project is released with a [Contributor Code of Conduct][ccoc].
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -172,7 +172,7 @@ git push -f origin my_feature_branch
 ## Git pre-commit hook
 
 Code submitted to this repository should be formatted according to `crystal tool format`.
-A pre-commit hook can be installed into the local git repo to ensure the formatter validates every commit: https://gist.github.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6
+A pre-commit hook can be installed into the local git repo to ensure the formatter validates every commit: https://github.com/crystal-lang/crystal/blob/master/scripts/git/pre-commit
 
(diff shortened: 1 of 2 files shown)


bcardiff:
The gist should use the ./bin/crystal. 
And it should use a built compiler (or at least warn the user that .build/crystal does not exist.

straight-shoota:
Using `bin/crystal` makes it only usable for the Crystal repo. But since it's a standalone gist, I think it should work with any Crystal repo.

I could add a check if `bin/crystal` is available and call that.

Checking for the existence of `.build/crystal` doesn't make much sense because it's not guaranteed that a local build is actually a recent build. It could even be an older build than the global `crystal` binary.

As mentioned in #7145, the use cases where the formatter has different results between master and last release are pretty rare for most users. Only for those who are working directly with the compiler/formatter would meet this more often. But for this work you probably wouldn't use such a pre-commit hook anyway because it's not flexible enough.

bcardiff:
Ok, I missed that the idea was to have a gist that can be used in multiple repos.

Using `./bin/crystal`, if exists, would be helpful.

wontruefree:
I think the URL for the pre commit hook should be under the crystal-lang group.  You can always check the source but it would add more confidence and adoption if it was under control of the crystal lang github.

bcardiff:
@wontruefree I agree, but AFAIK that is not possible. Gists can't be created under an organization.
A workaround would be to check in the file in this or other git repo.

straight-shoota:
Or a redirect from crystal-lang.org...

wontruefree:
it would be nice to have a src.crystal-lang.org domain to have scripts like this but it seems easier to add them to a repo.

bcardiff:
We have files.crystal-lang.org setup but without https. There is dist.crystal-lang.org that has a setup.sh is not the best location.

I would store in this repo directly and raw link to head if possible.

straight-shoota:
So essentially merge this and #7145?

bcardiff:
@straight-shoota actually... yes :-). With an update in #7145 so it works on other repos it should be fine.

straight-shoota:
Rebased on top of #7145

CONTRIBUTING.md
```diff
@@ -169,6 +169,17 @@ your own projects; a not so clean history is to prefer once a commit landed ther
 git push -f origin my_feature_branch
 ```
 
+## Git pre-commit hook
+
+Code submitted to this repository should be formatted according to `crystal tool format`.
+A pre-commit hook can be installed into the local git repo to ensure the formatter validates every commit: https://gist.github.com/straight-shoota/fdaf4cf1954e084cd5abccf8f77975f6
```
Sija:
```suggestion
A pre-commit hook can be installed into the local git repo to ensure the formatter validates every commit: https://github.com/crystal-lang/crystal/blob/master/scripts/git/pre-commit
```

sdogruyol: Good catch :+1:

RX14:
This hook isn't a good idea because any formatter changes since the last release which affect crystal's source itself will mean that you need to `make crystal` before you commit to get the latest formatter. I realize that this hook is optional but I don't like encouraging it for that reason.

straight-shoota:
@RX14 I don't think that matters much. Formatter changes are relatively rare and usually affect only some edge cases.

Issue #7608, opened 2019-03-30
Remove the comma as an enum member delimiter


j8r:
Following [this discussion](https://github.com/crystal-lang/crystal/pull/7605#discussion_r270628680), delimiting enum members can be done with either a comma, a newline or a semicolon (and also a space at this time, which is a bug). Allowing an additional delimiter character like the comma isn't needed, as it acts similarly as the semicolon:
```cr
enum E
  A; B; C
  D, E, F
end
```

Sija:
I'd keep it as a valid syntax, yet remove it by formatter as a preferred style.

jhass:
We don't use semicolons anywhere else I think, why keep them over the comma?

jhass:
Ah, nvm, because it's the same as newline. I'd reformat it to a newline here though I think.

asterite:
I'd remove the comma syntax. It's probably a breaking change but it's a weird syntax. There's nothing like that anywhere else in the language. Newline and semicolon are probably fine and enough.

j8r:
The closer syntax I see in the language is arguments:
```cr
puts(
  STDIN, STDOUT, STDERR
)
```
But enum members aren't arguments at all.

sevk:
comma is more human readable .

r00ster91:
>comma is more human readable .

The enums with the members delimited by a newline or semicolon are human readable too IMO. And it's also about consistency. As @j8r said, enum members aren't arguments. You define something.

ysbaddaden:
I don't think there is anything to brainstorm here: enum members must be separated by a linefeed, period. 

A semicolon is only accepted because it acts as a linefeed, similar to how it's syntactically valid to inline constants, but we never do it, because it's ugly, and hinders readability:
```crystal
FOO = 1
BAR = 2; BAZ = 3
```

PR #7618, opened 2019-04-03
Remove comma as enum member delimiter

j8r:
Resolves https://github.com/crystal-lang/crystal/issues/7608

### DIFF
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -2060,7 +2060,7 @@ module Crystal
       end
 
       # This is the case of an enum member
-      if node.name[0].ascii_uppercase? && @token.type == :";"
+      if @token.type == :";"
         next_token
         @lexer.skip_space
         if @token.type == :COMMENT
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -2060,7 +2060,8 @@ module Crystal
       end
 
       # This is the case of an enum member
-      if @token.type == :";"
+      # TODO: remove comma support after 0.28.0
+      if @token.type == :";" || (node.name[0].ascii_uppercase? && @token.type == :",")
         next_token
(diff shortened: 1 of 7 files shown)


src/compiler/crystal/tools/formatter.cr
```diff
@@ -2060,14 +2060,14 @@ module Crystal
       end
 
       # This is the case of an enum member
-      if node.name[0].ascii_uppercase? && @token.type == :","
-        write ","
```
oprypin: I'm not sure what the usual policy for the formatter is, but might it be possible to tell it to chew up old code with commas and re-format it, rather than outright failing?

asterite: Good idea! We've done it in the past. But for that the parser will still have to support commas, or at least the parser used in the formatter (could be toggled with a flag). I like that idea.

straight-shoota: That would be similar to an autocorrection feature for `check`, but implemented in the formatter.
I don't think it's urgent to completely abolish commas right away, so formatting them to semicolons seems like a good idea :+1:

src/compiler/crystal/tools/formatter.cr
```diff
@@ -2060,14 +2060,15 @@ module Crystal
...
+      # TODO: remove comma support after 0.28.0
```
Sija: Support for commas should stay regardless.

j8r: This is not the consensus from https://github.com/crystal-lang/crystal/issues/7608, though no dead line is decided. It could be pushed further.

straight-shoota: Having this TODO is certainly better than not, even if we later decided to keep this conversion indefinitely. But I don't think that's reasonable. We don't need to throw it out in the next version, but eventually it would just be useless dead code.

bcardiff:
Thanks @j8r and reviewers for this addition

PR #7626, opened 2019-04-04
Compiler: Avoid showing duplicate entries as warnings

bcardiff:
A leftover from #7596. Warnings should appear once per deprecated methods and not once per invocation.

### DIFF
--- a/src/compiler/crystal/codegen/warnings.cr
+++ b/src/compiler/crystal/codegen/warnings.cr
@@ -46,17 +46,26 @@ module Crystal
   end
 
   class CodeGenVisitor
+    @deprecated_methods_detected = Set(String).new
+
     def check_call_to_deprecated_method(node : Call)
       return unless @program.warnings.all?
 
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/codegen/warnings.cr
```diff
@@ -46,17 +46,21 @@ module Crystal
   end
 
   class CodeGenVisitor
+    @deprecated_methods_detected = Set(String).new
```
bew: Could you use a Set of Annotation nodes instead of method names? So that we could use the annotations on many other places

bcardiff: The annotation is stateless. And they are cloned between typed defs. But also this PR is wrong. I will send a push soon that handle better some cases.

I am stress testing it to avoid more pushes and reviews.

bcardiff:
@asterite @sdogruyol the first push was not right. I added a couple of more tests to ensure that a warning is reported once per receiver's type of the call plus the invocation location.

PR #7627, opened 2019-04-04
Fix infinite loop in asm with 3 colons and a variable

r00ster91:
This fixes an infinite compile loop when doing `asm("" ::: ""(var))`. The `until @token.type == :")"` loop never gets exited.


bcardiff:
Thanks @r00ster91 . Beware of using asm too much 😸

PR #7628, opened 2019-04-04
Pending leftovers to support // and &-ops in multiple places

bcardiff:
Some missing bits to fully support `//` in all places were `/` is supported today.

Ref: #2968


bcardiff:
Thanks for all the reviews, but the CI wasn't happy.
There are other left overs I've just found regarding the &-ops and //. I will send them in another PR.

spec/compiler/semantic/static_array_spec.cr
```diff
@@ -83,11 +83,20 @@ describe "Semantic: static array" do
 
   it "types static array new with size being a computed constant" do
     assert_type(%(
+      struct Int
+        def //(other : Int)
```
bcardiff: I've found it a bit odd that the value of `SIZE` can be computed by the math interpreter but the definition is still needed. It's similar to what is done in https://github.com/crystal-lang/crystal/blob/master/spec/compiler/semantic/const_spec.cr#L237-L239 .

bcardiff: It's just needed for the typing phase, so the implementation can be just a `0`. I've just pushed that change

Issue #7619, opened 2019-04-03
Move Array index methods to Indexable


straight-shoota:
`Array#[](range : Range)` and similar methods (including the ones added in #7338) should be valid for any `Indexable`. Most methods could be implemented directly in `Indexable` and only `#[]?(start : Int, count : Int)` needs to be abstract and implementation-specific.

Currently, only `Array` implement a full set of range methods, but most other including classes implement at least some method like `#[](range : Range)` or `#[](start : Int, count : Int)`.

The only trouble are `Tuple` and `StaticArray` which can't create dynamically sized instances of subsets of themselves. So the question is, what type should `Tuple#[](range : Range)` return. Or should it be possible at all? If not, we can't add this to `Indexable`.

Instead we could consider adding a module (for example `List`) which inherits from `Indexable` and implements range methods and allows modifications using `[]=`, `<<` etc.

asterite:
The problem with `[]` that returns several values is what type to return. For `Array` it's clear: `Array`. Should `Deque` return a `Deque`? What about `Tuple` and `StaticArray` as you mention?

That's why I don't think `[](range)` belongs in `Indexable`. Maybe another module... but I'm not sure if it will be a big win.

j8r:
It's possible to return `Tuple` for `Tuple#[range]`, but this means having potentially huge macro expands, with every possible subsets size of the instance pre-defined in it.

asterite:
For `Tuple` I always thought about supporting `[](range)` in a built-in way, similar to how `[](index)` is magical. But I didn't find good use cases for that so it might not be worth it.

j8r:
I've a use case:
```cr
def first_n_args(num : Int, *args)                                                                 
  args[0..n]                                                                             
end                                                                                    
puts first_n_args 2, 'a', 'b', 'c'
```
I'm using a similar logic for an interpreter, for now I use `case/when`s with `case num; when 0 {...}; when 1 {...}` etc, then another `case/when` to match this `Tuple` against others.

I don't think it's that necessary, but I like the idea of having `#[]` for all Indexables.

j8r:
And what about `String`? We could `include Indexable` in it. This means having `String#each`, `String#first` and loads of new methods form `Indexable`

asterite:
Ruby had each for String, I think. The problem is that a String is a collection of chars and bytes so it's not clear what should be iterated by default. Plus String already implement those tons of methods from Indexable.

j8r:
[...]

PR #7630, opened 2019-04-05
Include Indexable into String

j8r:
Follow up of https://github.com/crystal-lang/crystal/issues/7619

The end goal is to remove duplication with the `Indexable` module, benefit from methods from the module and no longer use the `Indexable.self.range_to_index_and_count` [class method](https://github.com/crystal-lang/crystal/blob/master/src/indexable.cr#L622).

I had a hard time to figure out why I got `Error opening file '/app/s/r/c/c/o/m/p/i/l/e/r/c/r/y/s/t/a/l/c/r' with mode 'r': No such file or directory (Errno)`. That's because `String` was assumed not to be an `Enumerable`

I'm still worry about slower performance, for example with `String#size` and `String#each`.

Feedbacks welcome.

### DIFF
--- a/spec/compiler/lexer/lexer_objects/strings.cr
+++ b/spec/compiler/lexer/lexer_objects/strings.cr
@@ -36,7 +36,7 @@ module LexerObjects
     def next_unicode_tokens_should_be(expected_unicode_codes)
       @token = lexer.next_string_token(token.delimiter_state)
       token.type.should eq(:STRING)
-      token.value.as(String).char_at(0).ord.should eq(expected_unicode_codes)
+      token.value.as(String).at(0).ord.should eq(expected_unicode_codes)
     end
 
     def next_string_token_should_be(expected_string)
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1707,10 +1707,6 @@ describe "String" do
     "\u{AB}1".codepoint_at(1).should eq('1'.ord)
   end
 
-  it "does char_at" do
-    "いただきます".char_at(2).should eq('だ')
-  end
-
   it "does byte_at" do
     "hello".byte_at(1).should eq('e'.ord)
     expect_raises(IndexError) { "hello".byte_at(5) }
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -194,7 +194,7 @@ class Hash(K, V)
   # h.dig? "a", "b", "c", "d", "e" # => nil
   # ```
   def dig?(key : K, *subkeys)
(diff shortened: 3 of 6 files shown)


oprypin:
You're going against a deliberate design decision that a string is not inherently iterable. At least it's relatively clear what the `chars` of a string are, and they are indeed what's iterable. But what does it really mean to iterate over, say, a poem? Over bytes, codepoints, visible characters, maybe even lines?

asterite:
I guess one thing we can do is to have views. I think Swift does that. Do you could have `string.byte_view`, `string.char_view`, `string.codepoint_view`, etc. These would just be wrapper structs over string that could be `Indexable` and `Enumerable`. So you could do:

```crystal
"hello".char_view.map { |c| c + 1 } # => ['i', 'f', 'm', 'm', 'p']
```

That's basically the same than `"hello".chars.map ...` except that an intermediate array for `chars` is not needed. Same with `bytes`, `codepoints`, etc.

bew:
@asterite About view, I have this in my wishlist for a loong time: StringView, for example you could do `some_string.view` (or `to_view`) and get a `StringView` object with all the (at least) lookup methods of `String`, but without the overhead of creating a `String` object each times..

For example for `String#split` instead of creating and returning new strings, you could do `my_string.view.split(' ', 2)` and get very lightweight StringView referencing the original `String`..

(_I should do a separate issue for this one day_)

j8r:
Why not having then a `String::Chars` struct, that is returned for `String#chars`? Same for codepoints, bytes etc. To have an array, one can do `"mystring".chars.to_a` then.

oprypin:
But that's exactly what @asterite said?

j8r:
On the idea yes, on my side I was proposing to replace the existing `String#chars : Array(Char)` by `String#chars : String::Chars`  (not creating a new `String#chars_view`).
I said this because:
- `String#each_char` will be replaced by `String::Chars#each`
- `String#chars` by `String::Chars#to_a`
- `String#each_char_with_index` by `String::Chars#each_with_index`

j8r:
Thanks for the feedbacks everyone, I wasn't sure it was a good idea to include Indexable into String.
We'll move this discussion to a proper issue about having indexable structs for chars, bytes and maybe even lines in the `String` namespace – I like the idea too.

j8r:
Thanks for the feedbacks everyone, I wasn't sure it was a good idea to include Indexable into String.
We'll move this discussion to a proper issue about having indexable structs for chars, bytes and maybe even lines in the `String` namespace – I like the idea too.

PR #7633, opened 2019-04-05
Remove String#at

j8r:
It is a duplicate of `String#char_at`, which is less explicit.
Nor `String#at` or `String#char_at` are documented, though public methods.
Not sure about the degree of deprecation we want (only a comment, using deprecated annotation or none?).

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -825,6 +825,16 @@ class String
     self[regex, group]?.not_nil!
   end
 
+  @[Deprecated("Use `String#char_at` instead.")]
+  def at(index : Int)
+    char_at(index)
+  end
+
+  @[Deprecated("Use `String#char_at` instead.")]
+  def at(index : Int)
+    char_at(index) { yield }
+  end
+
   def char_at(index : Int)
     char_at(index) { raise IndexError.new }
   end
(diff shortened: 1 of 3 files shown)


wontruefree:
Should there be a deprecation warning before removing something from the public API?

src/string.cr
```diff
@@ -825,11 +825,21 @@ class String
     self[regex, group]?.not_nil!
   end
 
+  @[Deprecated("Use `String#char_at` instead.")]
```
wontruefree: 👍

j8r:
The CI is still using 0.27.2 :confused:

bcardiff:
@j8r please rebase on master, the CI is updated now

asterite:
If we go this route we should probably deprecate/remove `char_at`. There's already `#[]`.

straight-shoota:
I'm not sure about removing `#char_at` as well, but it might be a good idea.

j8r:
... why we'll remove `String#char_at` and keep others like `Hash#fetch`?

asterite:
Because fetch accepts a block that is used as the value, when the key is missing.

j8r:
Good to merge? If deprecating `String#char_at` is wanted, it can be another PR.

RX14:
@bcardiff do we apply breaking change labels to deprecations?

bcardiff:
@RX14 yes 🙇

PR #7634, opened 2019-04-05
Remove the adler32/ and crc32/ directories

r00ster91:
This removes the directories `adler32` and `crc32` and puts the files directly in `src/`. It's unnecessary to have them in directories, it's just one file.


j8r:
For those wondering what are `adler32` and `crc32`:
- [Adler32](https://en.wikipedia.org/wiki/Adler-32)
- [CRC-32](https://en.wikipedia.org/wiki/Cyclic_redundancy_check)

asterite:
I actually want all libs to be inside directories.

asterite:
And probably move these under a namespace.

j8r:
@asterite you moving mean under `Crypto`? There are already `Crypto::Blowfish`, and `Crypto::Subtle`. We can have `Crypto::Adler32` and `Crypto::CRC32`

j8r:
@asterite you mean moving under `Crypto`? There are already `Crypto::Blowfish`, and `Crypto::Subtle`. We can have `Crypto::Adler32` and `Crypto::CRC32`

asterite:
I think so. But please let's not design using PRs. We can open issues to discuss these things.

straight-shoota:
Closing. We don't want to go in this direction.

Let's please discuss these refactorings first, thanks :+1:

PR #7639, opened 2019-04-05
Int div deprecation

bcardiff:
Depends on #7638.

This PR replaces (hopefully) all usages of `Int#/` with `Int#//`.

`Int#/` is marked as deprecated. This will cause some noise while building the compiler and specs since the method is still tested and used in places like macro interpreter. Without turning off warnings on some lexical scope we will need to tolerate that noise for a bit.

Since the stdlib is not ignored bye defualt (only `./lib`) if there are some functions without coverage in the stdlib users will notice them and will be able to report :-)

Ref: #2968

---

Submitting as draft until #7638 is merged

### DIFF
--- a/src/crystal/hasher.cr
+++ b/src/crystal/hasher.cr
@@ -244,7 +244,7 @@ struct Crystal::Hasher
   end
 
   private def read_u24(ptr, rest)
-    ptr[0].to_u64 | (ptr[rest/2].to_u64 << 8) | (ptr[rest - 1].to_u64 << 16)
+    ptr[0].to_u64 | (ptr[rest // 2].to_u64 << 8) | (ptr[rest - 1].to_u64 << 16)
   end
 
   private def read_u32(ptr)
--- a/src/int.cr
+++ b/src/int.cr
@@ -117,6 +117,7 @@ struct Int
     self.class.new(to_f // other)
   end
 
+  @[Deprecated("Int#/ will return a Float in 0.29.0. Use Int#// for integer division.")]
   def /(other : Int)
     self // other
   end
--- a/src/int.cr
+++ b/src/int.cr
@@ -362,7 +362,7 @@ struct Int
   end
 
   def lcm(other : Int)
-    (self * other).abs / gcd(other)
+    (self * other).abs // gcd(other)
   end
 
   def divisible_by?(num)
--- a/spec/std/deque_spec.cr
+++ b/spec/std/deque_spec.cr
@@ -75,10 +75,10 @@ describe "Deque" do
     it "works the same as array when inserting at 1/8 size and deleting at 3/4 size" do
       DequeTester.new.test do
         1000.times do
-          step { c.insert(c.size / 8, i) }
+          step { c.insert(c.size // 8, i) }
         end
         1000.times do
-          step { c.delete_at(c.size * 3 / 4) }
+          step { c.delete_at(c.size * 3 // 4) }
         end
       end
     end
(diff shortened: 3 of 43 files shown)


bcardiff:
Ready for the review!

src/int.cr
```diff
@@ -117,6 +117,7 @@ struct Int
     self.class.new(to_f // other)
   end
 
+  @[Deprecated("Int#/ will return a Float. Might need to use Int#// for integer division.")]
```
straight-shoota: I'd prefer to mention when this is expected to change.
Im not sure if we settled on that, though. 0.29.0 or give it more time?

ysbaddaden: As soon as the compiler is ready. I believe 0.28 is ready, and announces the deprecation, so 0.29 will introduce the breaking change.

bcardiff: We can say it's schedule in 0.29, but I haven't prepared the change to fully check if the compiler is ready.

And if some place of the std a replacement is missing then we need a patch or minor release in the middle probably.

So either 0.29.0 with the risk of forcing 0.28.1 or delaying until 0.30.0.
Or 0.30.0 and might need to delay more than necessary.

Both (or none) works for me.

straight-shoota: Let's just say 0.29.0. If it should be further delayed for some reason, it's not a big issue.

bcardiff: Done

bew:
What about a more imperative style:
```
Use Int#// for integer division.
```
Instead of:
```
Might need to use Int#// for integer division.
```

src/crystal/hasher.cr
```diff
@@ -244,7 +244,7 @@ struct Crystal::Hasher
   end
 
   private def read_u24(ptr, rest)
-    ptr[0].to_u64 | (ptr[rest/2].to_u64 << 8) | (ptr[rest - 1].to_u64 << 16)
+    ptr[0].to_u64 | (ptr[rest//2].to_u64 << 8) | (ptr[rest - 1].to_u64 << 16)
```
Sija:
```suggestion
    ptr[0].to_u64 | (ptr[rest // 2].to_u64 << 8) | (ptr[rest - 1].to_u64 << 16)
```

bcardiff:
Both suggestions addressed.

j8r:
I have a problem: `Int#/` **isn't** deprecated, it behaves differently.
It's possible to be needlessly spammed for a very valid code, like @RX14 point out.
It should be another type of warning, with a lower priority, disabled by default. information/note maybe.
The seprecated annotation doesn't fit for every API change.

j8r:
I have a problem: `Int#/` **isn't** deprecated, it behaves differently.
It's possible to be needlessly spammed for a very valid code, like @RX14 point out.
It should be another type of warning, with a lower priority, disabled by default. information/note maybe.
The deprecated annotation doesn't fit for every API change.

bcardiff:
@j8r it's disabled by default. When there is a change of the semantic although it's not a formal deprecation you need to call for a review at least. What it is lacking is a way to mark the warning as solved for example in the code. Although I am fine with dealing some false-positive reports.

A fully smooth migration would be to force every `/` to be either `//` or `temp_float_div` and then change `temp_float_div` to `/`, but I hardly believe somebody would embrace that journey.

j8r:
Yes I know warnings are disabled by default @bcardiff. Still, this is misleading to mark `Int#/` as deprecated – it's not. This syntax will stay valid, and won't go away (just bwhave differently)

straight-shoota:
> just behave differently

But that's exactly why its current use needs to be deprecated.

You could just see it like this: `Int#/(Int) : Int` is going to be removed in 0.29.0. If your code uses this method, it shouldn't. That's what the deprecation warning says.
In the same release (probably, except we make the extra detour @bcardiff described 😆 ) we're going to add `Int#/(Int) : Float` which essentially takes its place. But it's a different method with a different signature. Code using this method might continue to build or might fail, depending how the return value is used.

j8r:
The signature change, right. I doubt this really matter anyway. This isn't like this warnings will impact non insiders, unless they are curious about tryings flags or read the changelog. That's fine then.

PR #7641, opened 2019-04-06
Update `for` expression error

r00ster91:
This updates the `for` expression error a bit. It was slightly outdated. It also changes `array, hash` into `ArrayLiteral, HashLiteral` which is more consistent with `exp.class_desc` being a type.


r00ster91:
Now when I looked at it again I noticed that it doesn't really match the style of all the other errors, so I changed it a bit again and made it shorter.

PR #7642, opened 2019-04-07
Replace SemanticVersion with SoftwareVersion

### LINKED ISSUE #7637: Replace SemanticVersion
[`SemanticVersion`](https://crystal-lang.org/api/0.27.2/SemanticVersion.html) is a data type for representing version numbers.
It's used in the compiler to implement the [`compare_versions`](https://crystal-lang.org/api/0.27.2/Crystal/Macros.html#compare_versions(v1:StringLiteral,v2:StringLiteral):NumberLiteral-instance-method) macro, which in turn is used to adapt C bindings depending on the library version.

The implementation of `SemanticVersion` and `compare_versions` is solid. But the design is problematic. It's only targeting version numbers conforming to [`semver`](https://semver.org/) (`major.minor.patch-pre`). `semver` is a popular versioning system, but far from the only one. The reality shows a great variety of software version schemes employed in the wild.

For the implementation of `SemanticVersion`, a great number of current software versions can be considered compatible to semver, even if it doesn't actually follow semver, but has a similar format. But as soon as the format doesn't match up, `SemanticVersion` and `compare_versions` can't handle a version (such as `1.0` for example). For a general-purpose implementation, this is unacceptable. It should support a wide variety of version number formats.

`SemanticVersion` is therefore pretty much unusable for anything, unless all versions it needs to handle are guaranteed to conform to semver. In any real-world application, with versions coming from different vendors, this is very unlikely. `shards` for example uses a custom implementation for comparing version numbers that is more flexible.

Ruby has [`Gem::Version`](http://ruby-doc.org/stdlib-2.0.0/libdoc/rubygems/rdoc/Gem/Version.html) which is very versatile. I have a Crystal version of that type with a similar interface but a different internal implementation which should be more efficient and could even be further improved for comparison performance.
I think it could be used as a drop-in replacement of `SemanticVersion` and renamed to `SoftwareVersion`. It's available here: https://gist.github.com/straight-shoota/d8eec4a346018b0a51c1730b2a5c2bf8
I could also consider publishing it as a shard, but I think it would also fit in the stdlib and would improve `compare_versions` as well.

straight-shoota:
Fixes #7637

This allows to handle a wider variety of version numbers than only semver compatible ones.

The implementation can certainly be improved for performance. This PR focuses on establishing a more versatile behaviour and API. I'd like to postpone performance issues in a subsequent discussion. This can always be improved once the API is settled.


src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+  end
+
+  # Returns the string representation of this version.
+  def to_s : String
+    @version
+  end
+
+  # Returns `true` if this version is a pre-release version.
+  #
+  # A version is considered pre-release if it contains an letter or a dash (`-`).
```
r00ster91:
```suggestion
  # A version is considered pre-release if it contains a letter or a dash (`-`).
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+    @version.each_char_with_index do |char, index|
+      if char.ascii_letter? || char == '-' || char == '+'
+        return self.class.new(@version.byte_slice(0, index))
+      end
+    end
+
+    self
+  end
+
+  # Compares this version with *other* returning `-1`, `0`, or `1` if the
```
r00ster91:
```suggestion
  # Compares this version with *other* returning `-1`, `0`, or `1` depending on whether
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+  # other version is lower, equal or greater than `self`.
```
r00ster91:
```suggestion
  # *other*'s version is lower, equal or greater than `self`.
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+
+      # If one string is consumed, the other is either ranked higher (char is a digit)
+      # or lower (char is letter, making it a pre-release tag).
+      if lchar.nil?
+        return rchar.not_nil!.ascii_letter? ? 1 : -1
+      elsif rchar.nil?
+        return lchar.ascii_letter? ? -1 : 1
+      end
+
+      # Try to consume consequitive digits into a number
```
r00ster91:
```suggestion
      # Try to consume consecutive digits into a number
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+  end
+
+  # Parses an instance from a string.
+  #
+  # A version string is a series of digits or ASCII letters separated by dots.
+  #
+  # Returns `nil` if *string* describes an invalid version.
+  def self.parse?(string : String) : self?
+    # If string is an empty string convert it to 0
+    string = "0" if string =~ /\A\s*\Z/
```
r00ster91:
```suggestion
    string = "0" if string.blank?
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+  protected def initialize(@version : String)
+  end
+
+  # Parses an instance from a string.
+  #
+  # A version string is a series of digits or ASCII letters separated by dots.
+  #
+  # Returns `nil` if *string* describes an invalid version.
+  def self.parse?(string : String) : self?
+    # If string is an empty string convert it to 0
```
r00ster91:
```suggestion
    # If the string is a blank string, convert it to 0
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+# ASCII characters. The first segment is expected to contain only digits.
+#
+# If the string contains a dash (`-`) or a letter, it is considered a
+# pre-release.
+#
+# Optional version metadata may be attached and is separated by a plus character (`+`).
+# All content following a `+` is considered metadata.
+#
+# This format is described by the regular expression:
+# `/[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-.]+)?/`
```
r00ster91:
```suggestion
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+# Periods (and dash) delimit numbers but don't affect sort order by themselves.
+# Thus `1.0a` is considered equal to `1.0.a`.
+#
+# Pre-releases are sorted lower than the corresponding release version which
+# includes the characters up to the first dash or letter. For instance `1.0-b`
+# compares lower than `1.0` but greater than `1.0-a`.
+struct SoftwareVersion
+  include Comparable(self)
+
+  # :nodoc:
```
r00ster91:
```suggestion
  # Describes the version format.
```

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+# If the string contains a dash (`-`) or a letter, it is considered a
+# pre-release.
+#
+# Optional version metadata may be attached and is separated by a plus character (`+`).
+# All content following a `+` is considered metadata.
+#
+# This format is described by the regular expression:
+# `/[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-.]+)?/`
+#
+# This implementation is compatible to popular versioning schemes such as
```
r00ster91:
```suggestion
# This implementation is compatible with popular versioning schemes such as
```

spec/std/software_version_spec.cr
```diff
@@ -0,0 +1,168 @@
...
+    v("1.0.0+build1").should eq v("1.0.0+build1")
+
+    v("1.8.2.a").should be < v("1.8.2")
+    v("1.2").should be < v("1.3")
+    v("0.2.0").should be < v("0.2.0.1")
+    v("1.2.rc1").should be < v("1.2.rc2")
+  end
+
+  it "sort" do
+    list = ["0.1.0", "0.2.0", "5.333.1", "5.2.1", "0.2", "0.2.0.1", "5.8", "0.0.0.11"].map { |v| v(v) }
```
r00ster91: Avoid creating a new array by modifying the existing array:
```suggestion
    list = ["0.1.0", "0.2.0", "5.333.1", "5.2.1", "0.2", "0.2.0.1", "5.8", "0.0.0.11"].map! { |v| v(v) }
```

spec/std/software_version_spec.cr
```diff
@@ -0,0 +1,168 @@
...
+
+    list.sort.map(&.to_s).should eq ["0.0.0.11", "0.1.0", "0.2", "0.2.0", "0.2.0.1", "5.2.1", "5.8", "5.333.1"]
```
r00ster91:
```suggestion
    list.sort.map!(&.to_s).should eq ["0.0.0.11", "0.1.0", "0.2", "0.2.0", "0.2.0.1", "5.2.1", "5.8", "5.333.1"]
```

spec/std/software_version_spec.cr
bew: Well no, `v` call doesn't return a string..

straight-shoota:
@j8r The regex is not optimal and probably a major point to improve (btw. that's why the constant shouldn't be public @r00ster91, I'd like it to vanish eventually). But I really want to focus this PR on the API, not the implementation.

It's designed that the regex can be removed and instead use a custom parser + array storage (for example) without breaking the API. But before we can do that, I'd like to be able to *use* the new behaviour to help decide on how performance can be improved more effectively. There are a number of different possibilities and it's hard to tell which way to got without further research.

@r00ster91 `Version` is to unspecific IMO. `SoftwareVersion` is explicit and unambiguous.

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+# segments separated by periods. Each segment contains one ore more alphanumerical
+# ASCII characters. The first segment is expected to contain only digits.
+#
+# If the string contains a dash (`-`) or a letter, it is considered a
+# pre-release.
+#
+# Optional version metadata may be attached and is separated by a plus character (`+`).
+# All content following a `+` is considered metadata.
+#
+# This format is described by the regular expression:
```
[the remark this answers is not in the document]
straight-shoota: I don't think there is any benefit from having the regex in public API. It's not needed and should eventually vanish.

straight-shoota:
How did the first build on darwin succeed? https://circleci.com/gh/crystal-lang/crystal/21290?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
It contained an unguarded `@[Deprecated]` annotation which should fail on 0.27.2...

src/software_version.cr
```diff
@@ -0,0 +1,330 @@
...
+    # the constraint so that `0.1` doesn't match `~> 0.2`.
+    # self >= constraint
+    constraint_number, _ = consume_number(constraint, last_period_index + 1)
+    own_number, _ = consume_number(@version, last_period_index + 1)
+
+    own_number >= constraint_number
+  end
+
+  # Custom hash implementation which produces the same hash for `a` and `b` when `a <=> b == 0`
+  def hash(hasher)
```
j8r: An example would be welcome, along with type annotation if possible. Is there any advantage over `<=>` – performance?

straight-shoota: This is just the implementation of `Object#hash` and ensures that when `a <=> b == 0` both instances have the same hash.
here is no advantage over `<=>`. Both methods serve different methods.

spec/std/software_version_spec.cr
```diff
@@ -0,0 +1,168 @@
...
+    it "empty version" do
+      ["", "   ", " "].each do |empty|
+        SoftwareVersion.parse(empty).to_s.should eq "0"
+      end
+    end
+  end
+
+  it "#<=>" do
+    # This spec has changed from Gems::Version where both where considered equal
+    v("1.0").should be < v("1.0.0")
```
yxhuvud: I find it quite unintuitive that these are not equal. What is the motivation?

straight-shoota: They are in fact *different* with a different number of segments. Both interpretations are legitimate and I don't think there is a definitive argument for one or the other. It mostly depends on the context which behaviour would be expected. The practical relevance however is pretty small anyway. You're usually comparing software versions in a specific context (such as a versioned software package) where changes in version format are pretty rare and unlikely to assign a version that could be considered equal twice. So I don't think it matters much what the comparison returns.

When in doubt, I'd usually prefer to stick to the example of `Gem::Version`, but there are implications on hash performance. Equality in `<=>` should also result in equal hashes and ignoring trailing zero segments would add more complexity to the hasher, thus reducing performance.

ysbaddaden: It has been a cause of bugs in shards, and now shards considers 1 == 1.0 == 1.0.0 = ...

straight-shoota: @ysbaddaden What kind of bugs?

straight-shoota:
It seems that for `Gem::Version` the equality of `1.0 == 1.00` has not matched the hashes of these version numbers. This was only fixed relatively recently and there is a current discussion about the inconsistencies this caused in the Ruby implementation.

https://github.com/rubygems/rubygems/issues/2595
https://github.com/rubygems/rubygems/pull/2597

RX14:
I think things like `Gem::Version` prove that comparing version numbers isn't nearly as easy as it looks, there's a bunch of edgcases and tradeoffs. Semver fixed that, by being a rigid spec with rigid semantics. That means it's simple and unopinionated to make a `SemanticVersion` class.

I'd prefer this to be in a shard, where there are no backwards-compatibility guarantees past 1.0 so there won't be problems with locking in a bad behaviour.

As for what happens to `SemanticVersion` i'd be fine with moving it into the compiler sources and making it private.

j8r:
Can someone remind me why the compiler absolutely needs SemanticVersion?
For example with OpenSSL versions there was one approach by comparing integers and an other with the `compare_versions` macro.

It may be worth it to see to what extent this module is used, and how, before taking any decision.

straight-shoota:
Can we please confine the general feature discussion to #7637? This PR is about specific implementation.

straight-shoota:
Closing stale PR.

PR #7646, opened 2019-04-07
Allow Spec's .pending and .it to accept constant

straight-shoota:
This enables to use `pending` and `it` with a constant as description argument:

```cr
pending Foo do
end
it Foo do
end
```

This already works for `describe` and `context`, but for both `pending` and `it` raised an error because of mismatching types.

### DIFF
--- a/src/spec/methods.cr
+++ b/src/spec/methods.cr
@@ -36,6 +36,7 @@ module Spec::Methods
   #
   # It is usually used inside a `#describe` or `#context` section.
   def it(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, &block)
+    description = description.to_s
     Spec::RootContext.check_nesting_spec(file, line) do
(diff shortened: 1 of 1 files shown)


konovod:
Isn't spec name supposed to be a clear description of what test do?
Maybe I'm missing some cases, but
`Char::Reader raises IndexError when smth` - nice name for a test
`Char::Reader IndexError` - not nice.

straight-shoota:
@konovod Yeah, I'm not sure if it makes much sense for `it` to use a constant as description. But `pending` should definitely support it because you can simply convert a `describe` or `context` to `pending`. Given that all the other methods support a constant, it's better to have it on `it` as well.

straight-shoota:
Maybe a useful example for `it`:
```cr
describe "#read" do
  it Int { }
  it String { }
end
```

Sija:
I wouldn't go for `it` but pending makes sense.

Issue #7274, opened 2019-01-05
Add Annotation#types, #vars, #methods


vladfaust:
These methods would allow to iterate over all types/vars/methods which have the annotation, e.g:

```crystal
annotation Foo
end

@[Foo]
class Bar
end

{% pp Foo.types %} # => [Bar]
```

/cc @Blacksmoke16, this could be useful in [Athena](https://github.com/Blacksmoke16/athena), right?

I also can find it useful in ORM implementations.

Blacksmoke16:
Yes, this would allow me to just get a list of all methods that are tagged as a route action.  Currently I am requiring the controller classes to inherit from a predefined class, then iterating over the subclasses of that parent class, then filtering the methods down to just ones with the annotation.

This would be a bit cleaner, and allow the user to define their own inheritance structure.

vladfaust:
@Blacksmoke16 there is an alternative possible solution: https://forum.crystal-lang.org/t/type-all-includers/254

Blacksmoke16:
@vladfaust What would your use case be for wanting an array of _ALL_ methods/instance vars?

Would something like

```crystal
annotation Foo
end

class Klass
  @[Foo]
  def one
    1
  end
  
  @[Foo]
  def two
    2
  end

  def three
    3
  end
end

pp {{Klass.methods(Foo)}} # => [one, two]
```
Where `methods` on `TypeNode` optionally takes an `AnnotationType`, and only returns methods with that annotation?  This is already currently possible, would be fore sure be cleaner.

HertzDevil:
If https://github.com/crystal-lang/crystal/pull/7648#issuecomment-598167731 works for types, do we still need this feature for instance / class variables and defs, or are there similar workarounds for them?

Blacksmoke16:
@HertzDevil You can basically use the same workaround but with `@type.methods`.  Only thing this feature would allow for is like getting an `Array(Def)` of _all_ methods that have a given annotation irrespective of a specific type.  However I'm not sure of how that be useful/why you would need that.

kalinon:
I also realized today that the workaround im used to using (`Object.all_subclasses`) misses any modules with annotations as well.

PR #7648, opened 2019-04-07
TypeNode#annotated_types

Blacksmoke16:
Returns an array of all types (struct/class/module) with the given annotation.  Addresses some of #7274.  

```crystal
annotation Foo
end

@[Foo(val: 1)]
class Bar
end

@[Foo(val: 2)]
class Baz
end

class Blah
end

{% for a in Foo.annotated_types %}
  pp {{a.annotation(Foo)[:val]}}
{% end %}

# => 1
# => 2
```

A future PR could add `#annotated_methods`/`#annotated_variables`, but `#annotated_types` is the more useful one IMO.

/cc @vladfaust

### DIFF
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -2685,7 +2685,7 @@ module Crystal
   end
 
   class AnnotationType < NamedType
-    property class_types : Array(Type) = [] of Type
+    property class_types : Set(Type) = Set(Type).new
 
     def type_desc
       "annotation"
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -2685,6 +2685,8 @@ module Crystal
   end
 
   class AnnotationType < NamedType
+    property class_types : Array(Type) = [] of Type
+
     def type_desc
       "annotation"
     end
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -2685,7 +2685,7 @@ module Crystal
   end
 
   class AnnotationType < NamedType
-    property class_types : Set(Type) = Set(Type).new
+    property annotated_types : Set(Type) = Set(Type).new
 
     def type_desc
       "annotation"
--- a/spec/compiler/semantic/annotation_spec.cr
+++ b/spec/compiler/semantic/annotation_spec.cr
@@ -794,4 +794,116 @@ describe "Semantic: annotation" do
       ))
     end
   end
+
+  describe "#annotated_types" do
+    it "returns an empty array if there are none defined" do
+      assert_type(%(
+        annotation Foo
+        end
+
+        module Moo
+        end
+
+        {% if Foo.annotated_types.empty? %}
+          1
+        {% else %}
+          'a'
+        {% end %}
+      )) { int32 }
+    end
+
+    it "includes annotated modules" do
+      assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo]
+        module Moo
+        end
+
+        {% if Foo.annotated_types.size == 1 %}
+          1
+        {% else %}
+          'a'
+        {% end %}
+      )) { int32 }
+    end
+
+    it "includes annotated classes" do
+      assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo]
+        class Moo
+        end
+
+        {% if Foo.annotated_types.size == 1 %}
+          1
+        {% else %}
+          'a'
+        {% end %}
+      )) { int32 }
+    end
+
+    it "includes annotated structs" do
+      assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo]
+        struct Moo
+        end
+
+        {% if Foo.annotated_types.size == 1 %}
+          1
+        {% else %}
+          'a'
+        {% end %}
+      )) { int32 }
+    end
+
+    it "does not duplicate types if multiple annotations of same type are applied" do
+      assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo]
+        @[Foo]
+        struct Moo
+        end
+
+        {% if Foo.annotated_types.size == 1 %}
+          1
+        {% else %}
+          'a'
+        {% end %}
+      )) { int32 }
+    end
+
+    it "works with multiple types/annotations" do
+      assert_type(%(
+        annotation Foo; end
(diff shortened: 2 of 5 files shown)


vladfaust:
Small note – you don't actually check for *types* in specs, only for their size. I'm also concerned of how does it work with #7326...

Blacksmoke16:
In the case of #7326 you actually want to keep duplicates since the values on each could be different.  However, `annotated_types` you only want the list of types that are annotated, you wouldn't want `Klass` to show up twice just because you added two annotations to it.

Even if those two annotations had separate values, you would still be able to use `#annotations` in the for loop to get access to them.

j8r:
This can be confused with classic type annotations. The choice over Attributes was decided in https://github.com/crystal-lang/crystal/pull/6063

asterite:
Just a note that things like this push us further and further away from incremental/modular compilation (at this point we should probably drop that idea altogether and accept the existing nature of Crystal).

j8r:
@asterite if each library/shard is independent from each other, won't it be possible to compile each one and link them to the main program?

Blacksmoke16:
@asterite @straight-shoota anything else need to be done?

RX14:
@asterite incremental macros just means careful tracking of exactly which AST nodes a macro "sees" and being able to verify they're unchanged. I think that can be automated fairly well. We should remove the main blocker for incremental/modular compilation (type restrictions only being restrictions) *before* blocking other features over that though.

I think this is good to go.

src/compiler/crystal/types.cr
```diff
@@ -2685,6 +2685,8 @@ module Crystal
   end
 
   class AnnotationType < NamedType
+    property class_types : Set(Type) = Set(Type).new
```
straight-shoota: Is `class_types` a good name for this? Why not call it `annotated_type` as well?

Blacksmoke16: Its a set of the classes that have a given annotation, so makes sense yea.

bcardiff:
I would like to have the ability to query types/ivars/methods given an annotation type. 

I think that the incremental is doable (at some cost).

But what I would like to have a discussion/design before merging this is regarding subtyping. If Bar < Foo, are annotations inherited or not? Would that behavior be better specified in the annotation declaration itself (like in C# IIRC) or to define that upon querying (adding params to the proposed API)

src/compiler/crystal/types.cr
straight-shoota: => `annotated_types`

Blacksmoke16:
I think it would be better if they were not, and just added to each type explicitly; however I could be convinced otherwise.  IMO, This would make it less likely for unintended things to happen.  Being more verbose but allowing for more control seems to be a fair trade to me.

Blacksmoke16:
Renamed that property @straight-shoota, CI failure seems unrelated?

Blacksmoke16:
I don't suppose this will make it into `0.29.0`?

bcardiff:
@Blacksmoke16 I am interested in the feature but I haven't invested time yet to have proper design discussions. From my past experience, I am still inclined to mimic C# interface. But we should also consider how other language features impact: an annotation that is applied in a module that is included in a class does contribute to the annotations?

Blacksmoke16:
I'd like to get the discussion around this started again.  I'm getting close to a point where this is kind of a blocker to some Athena stuff I'm working on.  #8133 will be a good temporary solution, but having both this and that would be ideal.

watzon:
Any movement on this? This really is probably the ideal way for annotations to work, since they won't be hamstrung by needing to loop over annotations inside of a class. I really need this working for a project as well.

kalinon:
> Any movement on this? This really is probably the ideal way for annotations to work, since they won't be hamstrung by needing to loop over annotations inside of a class. I really need this working for a project as well.

Yes +1 for this feature

asterite:
I think this will be needed for the new logging stuff, otherwise how do you find all registered logs?

(I wonder why there's no registry and instead the proposal is using annotations, but it's a separate issue).

watzon:
It does seem necessary for a number of things, and it's so close to approval

bcardiff:
@asterite it's not needed. The following works.

```crystal
  def self.build_backend(type : String, config : Backend::Config)
    {% begin %}
      case type
      {% for t in Log::Backend.all_subclasses %}
        {% if reg = t.annotation(Log::RegisterBackend) %}
          when {{reg[0]}}
            {{t}}.new(config)
        {% end %}
      {% end %}
      else
        raise "There is no Log::Backend with @[Log::RegisterBackend(\"#{type}\")] annotation"
      end
    {% end %}
  end
```

It is not a registry as DB drivers because that might lead to a situation where the backend is not registered on time to initialize the `Log` constants.

Back to coding ... 👋

asterite:
So the annotation is only for the name, but you still need to inherit `Log::Backend`. Interesting!

Why not rely on a class method like `backend_name`? Why use annotations which are not type-safe?

asterite:
Also a warning: if a generic type inherits `Log::RegisterBackend` then it will stop working because you won't be able to use that uninstantiated generic type. Maybe there's no use case for generic types, though...

bcardiff:
> Why not rely on a class method like backend_name? Why use annotations which are not type-safe?

I liked the registry. The case is generated with literals. It can be changed of course, it is not the more important aspect :-). But the annotation based registry is doable.

Blacksmoke16:
> Also a warning: if a generic type inherits `Log::RegisterBackend` then it will stop working because you won't be able to use that uninstantiated generic type

This is what #8458 would help with, as it could print `Foo(T)` as `Foo`.

> @asterite it's not needed. The following works.

It works yes, but IMO it seems less than ideal to have to have a parent type/module just to do this; hence this PR.  Annotating a type should be enough, not for the `Log` use case specifically but are cases where it would be nice.

straight-shoota:
@Blacksmoke16 Totally agree. But I don't think anyone made the argument that this PR wasn't useful because there are simple alternatives. It's just not strictly necessary for the `Log::Backend` example. 👍

Blacksmoke16:
As a workaround for now, I realized you could just do `Object.all_subclasses.select &.annotation(MyAnn)`

Blacksmoke16:
Closing you can just use the workaround as defined in https://github.com/crystal-lang/crystal/pull/7648#issuecomment-598167731.

PR #7650, opened 2019-04-08
Add annotation Deprecated

straight-shoota:
This allows us to use `@[Deprecated]` without needing a version check to guard whether the compiler is 0.28.0 and supports it. As an additional benefit, this also adds it to the API docs.

Other compiler annotations should be documented similarly, but I'd like to get `Deprecated` merged soon in order to be able to finalize deprecation messages for 0.28.0 (for example for #7586).

### DIFF
--- a/src/annotations/deprecated.cr
+++ b/src/annotations/deprecated.cr
@@ -3,7 +3,7 @@
 # It receives a `StringLiteral` as single argument containing a deprecation notice.
 #
 # ```cr
-# @[Deprecated("#foo has been deprecated, use #bar instead")]
+# @[Deprecated("use #bar instead")]
 # def foo
 # end
 # ```
--- a/src/annotations/deprecated.cr
+++ b/src/annotations/deprecated.cr
@@ -3,7 +3,7 @@
 # It receives a `StringLiteral` as single argument containing a deprecation notice.
 #
 # ```cr
-# @[Deprecated("use #bar instead")]
+# @[Deprecated("Use #bar instead")]
 # def foo
 # end
 # ```
--- a/src/annotations/deprecated.cr
+++ b/src/annotations/deprecated.cr
@@ -1,0 +1,11 @@
+# This annotations marks methods, types or constants as deprecated.
+#
+# It receives a `StringLiteral` as single argument containing a deprecation notice.
+#
+# ```cr
+# @[Deprecated("#foo has been deprecated, use #bar instead")]
+# def foo
+# end
+# ```
+annotation Deprecated
+end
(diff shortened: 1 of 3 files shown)


src/docs_main.cr
```diff
@@ -2,6 +2,7 @@
 # It, for example, doesn't include API for the compiler, but does include
 # the fictitious API for the Crystal::Macros module.
 
+require "./annotations/*"
```
bcardiff: We need to add it to the prelude also.

src/annotations/deprecated.cr
```diff
@@ -0,0 +1,11 @@
+# This annotations marks methods, types or constants as deprecated.
+#
+# It receives a `StringLiteral` as single argument containing a deprecation notice.
+#
+# ```cr
+# @[Deprecated("#foo has been deprecated, use #bar instead")]
```
bcardiff: `top-level foo` is added as a prefix of the warning. There is no need to mention it.

src/annotations/deprecated.cr
```diff
@@ -0,0 +1,11 @@
+# This annotations marks methods, types or constants as deprecated.
```
bcardiff: Only methods

straight-shoota: The compiler `--check` feature only supports methods, but the annotation can be used on anything, right?

bcardiff: That should be right. There is no check whether the annotation makes sense unless it is used in the semantic or codegen phase.

bcardiff: Also, the `--check` is not the final option. I need to update the PR description...

straight-shoota: I'd like to view this annotation independent of the specific compiler feature. `--check` is just one application where it is used. The doc generator would be another one. And I could easily see both supporting types and constants as well in the future.

straight-shoota: Yeah, better rename it. I'm aware that it has changed, but `--check` feels like a good way to address it... :P

asterite: I can imagine people using `Deprecated` on types and opening an issue here because it doesn't work.

bcardiff: And PRs will be welcomed

r00ster91:
Does this really need to be in a new folder? IMO this could easily be in a `annotations.cr`. There aren't that many annotations that there needs to be a new file for every single annotation. As an example the macros aren't in 4 different files either: https://github.com/crystal-lang/crystal/blob/master/src/macros.cr. So why the annotations?

PR #7651, opened 2019-04-08
add bool types to Bool

wontruefree:
This adds some types to the Bool struct.


bcardiff:
@wontruefree those restriction are left out intentionally

`false || "hi" => "hi"`

PR #7653, opened 2019-04-08
Display deprecated label when @[Deprecated] is used

bcardiff:
~~Depends on #7652 (marked as draft until merged)~~

The last commit of the PR adds support of the doc generator tool to understand the `@[Deprecated]` and emit the same label that is currently emitted when a `# DEPRECATED: ...` comment is found.

Even before this PR the labels are only emitted for methods. But the implementation will be ready to handle them once other annotations are preserved by the parser/compiler.

<details><summary>Show me a sample (with screenshot)</summary><p>

Given the following code:

```crystal
module FOO
  # DEPRECATED: Use `Bar` instead1
  @[Deprecated("Use `Bar` instead2", "sdfs")]
  class Foo
    # DEPRECATED: Use `VAL_B` instead1
    @[Deprecated("Use `VAL_B` instead2")]
    VAL_A = 10
    VAL_B = 20

    # DEPRECATED: Use `#bar` instead1
    @[Deprecated("Use `#bar` instead2")]
    def foo
    end

    # DEPRECATED
    @[Deprecated]
    def foo2
    end

    def bar
    end

    # DEPRECATED: Use `#qux` instead1
    @[Deprecated("Use `#qux` instead2")]
    macro baz
    end

    macro qux
    end
  end

  class Bar
  end
end
```

This is the generated output:

screencapture-file-Users-bcardiff-Projects-crystal-master-docs-FOO-Foo-html-2019-04-08-17_50_00

</p></details>

### DIFF
--- a/src/compiler/crystal/tools/doc/constant.cr
+++ b/src/compiler/crystal/tools/doc/constant.cr
@@ -38,4 +38,8 @@ class Crystal::Doc::Constant
       builder.field "summary", formatted_summary
     end
   end
+
+  def annotations(annotation_type)
+    @const.annotations(annotation_type)
+  end
 end
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -300,6 +300,7 @@ class Crystal::Doc::Generator
 
   def doc(context, string)
     string = isolate_flag_lines string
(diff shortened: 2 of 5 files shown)


bcardiff:
Yes @straight-shoota I agree. It's cheap. 

Merging this will mean that the time deprecation does not need to be in the comments 🚀

PR #7654, opened 2019-04-08
add types to slice

wontruefree:
I added some types to slice and renamed the `hexstring` function that takes a pointer to `to_unsafe_hexstring`.

### DIFF
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -64,7 +64,7 @@ struct Slice(T)
   #
   # String.new(slice) # => "abc"
   # ```
-  def initialize(@pointer : Pointer(T), size : Int, *, @read_only = false)
+  def initialize(@pointer : Pointer(T), size : Int, *, @read_only : Bool = false)
     @size = size.to_i32
   end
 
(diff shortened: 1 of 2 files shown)


konovod:
@wontruefree `hexstring` doesn't take only pointer, it can take a Slice.

src/slice.cr
```diff
@@ -493,11 +495,11 @@ struct Slice(T)
...
   end
 
-  private def to_hex(c)
+  private def to_hex(c : Int) : UInt8
     ((c < 10 ? 48_u8 : 87_u8) + c)
   end
 
-  def bytesize
+  def bytesize : Int32
     sizeof(T) * size
```
yxhuvud: Whoa, this may be a candidate for being upscaled to Int64.

wontruefree: or maybe this could just be `Int`

wontruefree:
@konovod It also mutates the values in place though.  This seems like an unsafe action to me.
Also, I don't mind leaving it untyped until the API gets more specific.

src/slice.cr
yxhuvud: The problem is with the implementation, not the type. It will be wrong if sizeof(T) * size > Int32::MAX.

wontruefree: That makes sense but I would want to pull that into another PR

PR #7656, opened 2019-04-09
Textual improvements

r00ster91:
This fixes more typos and does some other minor textual improvements. It for example makes the `BigDecimal` documentation visible in the API docs.


src/file/info.cr
```diff
@@ -75,8 +75,7 @@ class File
     end
   end
 
-  # A `File::Info` contains metadata regarding a file. It is returned by
-  # `File.info`, and `File#info`.
+  # A `File::Info` contains metadata regarding a file. It is returned by `File.info`.
```
RX14: Why was this changed?

r00ster91: Hmm I probably thought for some reason that the method doesn't exist.

src/big/big_decimal.cr
```diff
@@ -1,20 +1,19 @@
...
   def initialize(big_decimal_str : String, reason : String)
     super("Invalid BigDecimal: #{big_decimal_str} (#{reason})")
   end
 end
 
+# A `BigDecimal` can represent arbitrarily large precision decimals.
+#
+# It is internally represented by a pair of `BigInt` and `UInt64`: value and scale.
+# Value contains the actual value, and scale tells the decimal point place.
+# E.g. when value is 1234 and scale 2, the result is 12.34.
```
straight-shoota:
```suggestion
# E.g. when value is `1234` and scale `2`, the result is `12.34`.
```

PR #7657, opened 2019-04-09
Specify an initial capacity if possible

r00ster91:
This uses `Array`'s `initial_capacity` argument more often which avoids repeately reallocating the internal buffer because the size changed. It also optimizes `Enumerable#size` slightly to use the including type's `size` to avoid iterating through each element to find the size (which `count { true }` does).

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1172,14 +1172,7 @@ module Enumerable(T)
   # [1, 2, 3, 4].size # => 4
   # ```
   def size
-    if self.is_a?(Array) || self.is_a?(Slice) ||
-       self.is_a?(Deque) || self.is_a?(Tuple) ||
-       self.is_a?(Range) || self.is_a?(Set) ||
-       self.is_a?(StaticArray)
-      self.size
-    else
-      count { true }
-    end
+    count { true }
   end
 
   # Returns an `Array` with the first *count* elements removed
@@ -1193,7 +1186,7 @@ module Enumerable(T)
   def skip(count : Int)
     raise ArgumentError.new("Attempt to skip negative size") if count < 0
 
-    array = Array(T).new(self.size - count)
+    array = Array(T).new(size - count)
     each_with_index do |e, i|
       array << e if i >= count
     end
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -517,10 +517,17 @@ module Enumerable(T)
   def in_groups_of(size : Int, filled_up_with : U = nil) forall U
     raise ArgumentError.new("Size must be positive") if size <= 0
 
-    ary = Array(Array(T | U)).new(self.size*size)
+    initial_capacity = self.size / size
+
+    if self.size.odd?
+      initial_capacity += 1
+    end
+
+    ary = Array(Array(T | U)).new(initial_capacity)
     in_groups_of(size, filled_up_with) do |group|
       ary << group
     end
+
     ary
   end
 
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -517,7 +517,7 @@ module Enumerable(T)
   def in_groups_of(size : Int, filled_up_with : U = nil) forall U
     raise ArgumentError.new("Size must be positive") if size <= 0
 
-    ary = Array(Array(T | U)).new
+    ary = Array(Array(T | U)).new(self.size*size)
     in_groups_of(size, filled_up_with) do |group|
       ary << group
     end
@@ -705,7 +705,7 @@ module Enumerable(T)
   # [1, 2, 3].map { |i| i * 10 } # => [10, 20, 30]
   # ```
   def map(&block : T -> U) forall U
-    ary = [] of U
+    ary = Array(U).new(size)
     each { |e| ary << yield e }
     ary
   end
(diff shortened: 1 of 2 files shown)


src/enumerable.cr
```diff
@@ -1172,7 +1172,14 @@ module Enumerable(T)
   # [1, 2, 3, 4].size # => 4
   # ```
   def size
-    count { true }
+    if self.is_a?(Array) || self.is_a?(Slice) ||
+       self.is_a?(Deque) || self.is_a?(BitArray) ||
+       self.is_a?(Tuple) || self.is_a?(StaticArray) ||
+       self.is_a?(Range) || self.is_a?(Set)
+      self.size
```
bew: If it worked, you'd called yourself recursively...
This is a no-op, since `Enumerable` is usually included at the beginning or the class, and `size` is overridden later in that class.

r00ster91: I'm doing all this for methods like `Enumerable#to_a`. So the types's `size` is used there instead of `count { true }`.
Or the array creation in `to_a` becomes `ary = Array(T).new(self.size)` (which won't call this method then) but that wouldn't work for all types. Not all have `#size`.

bew: But... using `size` in `to_a` will call the overridden `size` method if there is one, or the 'default' one in `Enumerable` if not

src/enumerable.cr
```diff
@@ -517,7 +517,7 @@ module Enumerable(T)
   def in_groups_of(size : Int, filled_up_with : U = nil) forall U
     raise ArgumentError.new("Size must be positive") if size <= 0
 
-    ary = Array(Array(T | U)).new
+    ary = Array(Array(T | U)).new(self.size*size)
```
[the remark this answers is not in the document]
r00ster91: Perhaps it's better to avoid floats here and `to_f` and leave it simple: `self.size / size` which will still avoid quite some reallocations.

src/enumerable.cr
r00ster91: Really? Hmm I understood it the way that `self.size` will call the overridden `size` method if possible and `size` always calls `Enumerable`'s `size`. But you are probably right, I will remove these if checks now.

src/enumerable.cr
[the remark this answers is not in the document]
r00ster91: No, using floats doesn't seem to be very problematic. I benchmarked and your way is the fastest one.

Sija:
```suggestion
    ary = Array(Array(T | U)).new(self.size * size)
```

src/enumerable.cr
```diff
@@ -705,7 +705,7 @@ module Enumerable(T)
   # [1, 2, 3].map { |i| i * 10 } # => [10, 20, 30]
   # ```
   def map(&block : T -> U) forall U
-    ary = [] of U
+    ary = Array(U).new(size)
```
kostya: is not `size` is a function which have computation cost?

r00ster91:
Hmmm I'm about unsure about why this happens:
```cr
module Enumerable
  def other_to_a
    # ary = Array(T).new(size)
    
    self.each { |e| puts e } # => Values get printed, `each` works
    
    # ary
  end
end

puts [1, 2, 3].cycle.first(5).other_to_a
puts [1, 2, 3].cycle.first(5).to_a # => [1, 2, 3, 1, 2]
```
And when now uncommenting the comments and replacing `puts e` by `ary << e`:
```cr
module Enumerable
  def other_to_a
    ary = Array(T).new(size)
    
    self.each { |e| ary << e } # Nothing gets yielded
    
    ary
  end
end

puts [1, 2, 3].cycle.first(5).other_to_a # => []
puts [1, 2, 3].cycle.first(5).to_a # => [1, 2, 3, 1, 2]
```
Replacing `ary = Array(T).new(size)` by `ary = [] of T` makes it work (that's also what `to_a` uses).

src/enumerable.cr
bew: Basically if you have a local variable called `size`, doing `size` will access the local var not the method, instead one can use `self.size` to call the method. There is no way to call `Enumerable`'s `size` when the includer type overrides the method `size`.

src/enumerable.cr
r00ster91: No, `size` has only in very very rare cases a computation cost: when `Enumerables`'s `size` is used because it iterates through every element to find out the size. However on arrays for example, this will call `Array`'s `size` which is a `getter` and has zero computation cost.
https://github.com/crystal-lang/crystal/blob/60760a5460aa13a1f80c5a1a7410782dc4076b77/src/array.cr#L55

bew:
It's because `size` will consume the iterator to know its size, so once you do `size`, there is nothing left in the iterator to read..

src/enumerable.cr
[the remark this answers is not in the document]
r00ster91: But this is off by 1 when the number of elements is odd so I will just do `self.size / size` and add 1 if `self.size` is odd.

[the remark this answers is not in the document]
bew: That's why I use floats, and `ceil` to get the upper value (`3.5` → 4)
Using floats, with group size is 2:
- when self.size is 7, `(7.to_f / 2).ceil.to_i` → 4
- when self.size is 8, `(8.to_f / 2).ceil.to_i` → 4

asterite:
Thank you for the PR but it's not correct.

Please, please, please, people, open an issue ti discuss things first.

I really want now a contribution guideline that reads: PRs will be closed if they are not related to an existing issue that's been discussed and approved before.

Otherwise I feel everyone jumps to review every PR, which can be a waste of time if the PR is not needed or not wanted.

PR #7660, opened 2019-04-09
Let IO#copy return UInt64

### LINKED ISSUE #7659: Upload large file `Content-Length` Int32 overflow
# Summary
I am trying to upload a 5 GibiBytes(`5515055111 Bytes`) file but this problem occurs.

If the full upload is completed, you will get: 
`Content-Length header is 5515055111 but body had 1220087815 bytes (ArgumentError)`

When uploaded to `2521.792672 MegaBytes ≈ 2521792672 Bytes`, you will get: 
`Content-Length header is 5515055111 but body had -1773174624 bytes (ArgumentError)`

Maybe it's Int32 overflow? (I guess).
Update: Yes, I confirm that it is caused by Int32 overflow.

### Crystal - ICR
```text
icr(0.27.2) > 2521792672.to_i32
 => -1773174624
```

# Position
Source of this error: [common.cr#L141](https://github.com/crystal-lang/crystal/blob/60760a5460aa13a1f80c5a1a7410782dc4076b77/src/http/common.cr#L141)

# Problem
```text
Unhandled exception on HTTP::Handler
Content-Length header is 5515055111 but body had -1773174624 bytes (ArgumentError)
  from HTTP::Client#send_request<HTTP::Request>:Bool
  from ~procProc(HTTP::Server::Context, Nil)@src/toucan/relay/server.cr:40
  from HTTP::Server::RequestProcessor#process<IO+, IO+, IO::FileDescriptor>:Nil
  from ~procProc(Nil)@/usr/local/Cellar/crystal/0.27.2/src/http/server.cr:380
  from Fiber#run:(IO::FileDescriptor | Nil)
```

# Screenshot
螢幕截圖 2019-04-10 01 01 42

asterite:
Fixes #7659

`IO.copy` shouldn't overflow. Plus the return value isn't normally used. But when it is, it's better if it has the correct value, accounting for large uploads/copies.

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -1123,14 +1123,14 @@ abstract class IO
   #
   # io2.to_s # => "hello"
   # ```
-  def self.copy(src, dst)
+  def self.copy(src, dst) : UInt64
     buffer = uninitialized UInt8[4096]
-    count = 0
+    count = 0_u64
     while (len = src.read(buffer.to_slice).to_i32) > 0
       dst.write buffer.to_slice[0, len]
       count += len
     end
-    len < 0 ? len : count
+    count
   end
 
   # Copy at most *limit* bytes from *src* to *dst*.
@@ -1143,8 +1143,10 @@ abstract class IO
   #
   # io2.to_s # => "hel"
   # ```
-  def self.copy(src, dst, limit : Int)
+  def self.copy(src, dst, limit : Int) : UInt64
     raise ArgumentError.new("Negative limit") if limit < 0
+
+    limit = limit.to_u64
 
     buffer = uninitialized UInt8[4096]
     remaining = limit


j8r:
Unrelated thing: it can be a instance methods, and add the annotation `dst : IO`?

straight-shoota:
@j8r An instance method would be weird. Does it copy the receiver to the argument or is it the other way around? We'd need to go with `#copy_to` or `#copy_from`. But `IO.copy(from, to)` reads very nice and is self-exlpanatory.

src/io.cr
```diff
@@ -1123,14 +1123,14 @@ abstract class IO
   #
   # io2.to_s # => "hello"
   # ```
-  def self.copy(src, dst)
+  def self.copy(src, dst) : UInt64
```
r00ster91: A little external argument maybe? I like to use them because they improve readabilty.
```suggestion
  def self.copy(src, to dst) : UInt64
```

src/io.cr
```diff
@@ -1143,9 +1143,11 @@ abstract class IO
   #
   # io2.to_s # => "hel"
   # ```
-  def self.copy(src, dst, limit : Int)
+  def self.copy(src, dst, limit : Int) : UInt64
```
r00ster91:
```suggestion
  def self.copy(src, to dst, limit : Int) : UInt64
```

asterite:
We could actually have copy_to and copy_from but it's a separate discussion

olbat:
Does this PR fixes https://github.com/crystal-lang/crystal/issues/3209 as well?

PR #7664, opened 2019-04-11
Add deprecation message to PartialComparable

straight-shoota:
Follow up on #6611 

Adding `@[Deprecated]` annotation to `#<=>` method is not detected by the warning feature. I don't think it makes sense to add annotations to the other methods. This module seems to be pretty much not used anywhere at all.

### DIFF
--- a/src/partial_comparable.cr
+++ b/src/partial_comparable.cr
@@ -7,8 +7,7 @@
 #
 # `PartialComparable` uses `<=>` to implement the conventional
 # comparison operators (`<`, `<=`, `==`, `>=`, and `>`).
-#
-# DEPRECATED: This module is deprecated as of Crystal 0.28.0. Its behaviour has been fully integrated into `Comparable`.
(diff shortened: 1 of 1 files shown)


bcardiff:
The issue with detecting the deprecation of `<=>` is that it's an abstract method. And annotations of abstrac methods are not passed to their implementation.

```crystal
module Foo
  @[Deprecated]
  abstract def foo
end

class Bar
  include Foo

  def foo
    1
  end
end

puts Bar.new.foo
```

That is the same story as identifing if a method is overriding or declaring a new method.

I this case the best thing to do is deprecate the constant/module itself.

It would be great to detect mentions of deprecated constants after 0.28.

bcardiff:
@straight-shoota, when I checked locally the deprecation, is only emitted if it goes **before** the docs comment.

> The @[Deprecated] can be used before the docs comment and will be output in the html docs.
> There might be something with the parser regarding the order of docs and annotations.

straight-shoota:
Sry, missed that.

It's super weird though :)

PR #7665, opened 2019-04-11
Add documentation for Flags and Link annotations

bcardiff:


src/annotations.cr
```diff
@@ -9,3 +9,42 @@
...
+# 2. will pass `-lpcre` to the linker.
+#
+# `@[Link("pcre", static: true)]` will favor static libraries over shared libraries.
+# 1. will lookup `libpcre.a` in `CRYSTAL_LIBRARY_PATH`, if not found
+# 2. will lookup `pcre` using `pkg-config --static`, if not found,
+# 3. will lookup `libpcre.a` in `/usr/lib`, `/usr/local/lib`
+#
+# `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker.
+#
+# The `CRYSTAL_LIBRARY_PATH`, `/usr/lib`, and `/usr/local/lib` are added for shared libraries lookup using `-L`.
```
straight-shoota: This line is not related to `@[Link]`?

bcardiff: I think it doesn't hurt. I see that is not about the usage of `@[Link]`, but it complements its semantics. 

Saying that `@[Link("foo")]` will append `-l foo` to the linker might not be completely helpful unless the default context/configuration of the linker is known.

straight-shoota: Yes, it's probably useful context information. But the sentence could be interpreted that this is an effect of `@[Link]`. It should simply be rephrased.

Suggestion:
```
When an `-l` option is passed to the linker, it will lookup the libraries in paths passed
with the `-L` option. `CRYSTAL_LIBRARY_PATH`, `/usr/lib`, and `/usr/local/lib` are added
by default. Custom paths can be passed using `ldflags`: `@[Link(ldflags: "-Lvendor/bin")]`.
```

PR #7668, opened 2019-04-12
Restore Time.new as deprecated method

bcardiff:
As part of #7586 to avoid major breaking changes in #5346 the following constructor should be added.


bcardiff:
`def self.new(location : Location = Location.local) : Time` should have been also added... . 

I noticed thanks to Lucky. Lucky me.

Issue #7645, opened 2019-04-07
[RFC] Indexable String wrappers for char, bytes and lines


j8r:
## Background

After some thought about [this issue to move Array index methods to Indexable](https://github.com/crystal-lang/crystal/issues/7619). What about `String`?

There are lots of methods that are mirrored from `Indexable` like `each_char`,  `each_char_with_index`, `#[]`, `#[]?`... Same for method related to bytes `each_bytes` and lines `each_line`.

Another problem is having `"str".each_char` and `"str".chars.each`. Several `Indexable` methods like `map` can only be used by creating an intermediate array.
 
Including `Indexable(Char)` into `String` isn't [a proper solution ](https://github.com/crystal-lang/crystal/pull/7630), because a string can be, by nature, represented as bytes, chars, codepoints.

## Proposal

On the high level API, the changes will be:
- `"str".each_char` => `"str".chars.each`
- `"str".chars` => `"str".chars.to_a`
- `"str".each_char_with_index` => `"str".chars.each_with_index`
- `"str".count` => `"str".chars.count`
- `"str".blank?` => `"str".chars.all? &.whitespace?`
- `"str".codepoints` => `"str".chars.map &.ord`
- `"str".each_codepoint`=> `"str".chars.each { |c| c.ord }`
- `"str".lines` =>  `"str".lines.to_a`
- `"str".each_line` => `"str".lines.each`
- `"str".bytes` =>  `"str".bytes.to_a`
- `"str".each_byte` => `"str".bytes.each`

That's a preview, with possibly even more method changes.
The codepoints changes are less user-friendly, but being quite low-level and rarely used, on my opinion it's fine.

## Implementation

Better to concentrate first on the proposal's API, but if you wonder what type `String#chars` can be and how this can be implemented by using structs: `String::Chars`, `String::Bytes` and `String::Lines`. `CharIterator` and `LineIterator` can be moved to `String::Chars::Iterator` and `String::Lines::Iterator`.

This proposal could superset https://github.com/crystal-lang/crystal/pull/7275

straight-shoota:
The idea doesn't sound too bad, but what actual problem does it solve?

j8r:
@straight-shoota remove duplicates with `Indexable`, improve the API (no more `each_char` vs `chars.each`), avoid creating intermediate arrays to use Enumerable methods like `map`, `reject` or `select`.

asterite:
[...]

PR #7672, opened 2019-04-15
[RFC] Implement String wrappers

j8r:
This wrappers represents the string as a sequence of Chars, Lines or Bytes

Implements #7645 

This PR mainly exists to show how String wrappers can be implemented in practice, with as little breaking changes as possible.

Main advantages are:
- performance gain by removing the need of an intermediate Array
- less boilerplating by using `Indexable` (no need to reimplement)
- benefit from all methods brought by `Indexable`

Disadvantages - breaking changes:
- `String#chars` -> `String#chars.to_a`
- `String#bytes` -> `String#bytes.to_a`
- `String#lines` -> `String#lines(chomp = true).to_a`

Another thing is the chomp argument to pass for every lines related methods, `@chomp` is now an ivar to pass only once in the constructor.

We can see latter what we do for other existing methods `each_char`, `each_char_with_index`, `each_line` etc

### DIFF
--- a/samples/llvm/brainfuck.cr
+++ b/samples/llvm/brainfuck.cr
@@ -137,7 +137,7 @@ class Program
     new source.chars
   end
 
-  def self.new(source : Array(Char))
+  def self.new(source : String::Chars)
     new parse(source, 0, source.size)
   end
 
--- a/spec/std/io/io_spec.cr
+++ b/spec/std/io/io_spec.cr
@@ -606,7 +606,7 @@ describe IO do
         buffer = uninitialized UInt8[1024]
         bytes_read = io.read_utf8(buffer.to_slice) # => 3
         bytes_read.should eq(3)
-        buffer.to_slice[0, bytes_read].to_a.should eq("你".bytes)
+        buffer.to_slice[0, bytes_read].to_a.should eq("你".bytes.to_a)
       end
 
       it "raises on incomplete byte sequence" do
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1722,7 +1722,7 @@ describe "String" do
   end
 
   it "does chars" do
-    "ぜんぶ".chars.should eq(['ぜ', 'ん', 'ぶ'])
+    "ぜんぶ".chars.to_a.should eq(['ぜ', 'ん', 'ぶ'])
   end
 
   it "allows creating a string with zeros" do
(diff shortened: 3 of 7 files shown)


asterite:
> This PR mainly exists to show how String wrappers can be implemented in practice, with as little breaking changes as possible.

If you want to avoid breaking changes, call then char_view, etc. I will never accept changing the return type of chars and bytes.

j8r:
> If you want to avoid breaking changes, call them char_view, etc. I will never accept changing the return type of chars and bytes.

Nobody will ever use it then. Am I missing an important point of why it's mandatory that `String#chars` returns an array (excluding ruby)? Having both `String#chars` and `String#char_view.to_a ` is a duplication of the API for no apparent reason.

There are lots of languages doing `for char in mystring` like Python, Nim, or Rust `for char in mystring.chars() { }`. The natural Crystal equivalent is `mystring.chars.each`.

asterite:
> Having both String#chars and String#char_view.to_a is a duplication of the API for no apparent reason.

It's the same reason `array.map` returns an Array, not an Iterator. In most cases the performance you get from iterators is not needed. If iterator is returned by default it's unintuitive because you have to remember to stick `to_a` at the end. Make it the other way around (`Array` is returned by default) and it's intuitive. If you need more performance, turn it into an iterator and call `to_a`.

I'd also like to stop doing breaking changes. The current API is fine. I believe we are just inventing a new API for no reason (no real use case asked for this).

j8r:
What about `String#each_char_with_index`? This RFC avoid having this type of methods (IMHO ugly)

straight-shoota:
It's still not a real use case and doesn't justify this change.

straight-shoota:
I'd rather invest time and effort in feature that are actually needed.

j8r:
@straight-shoota by seing https://github.com/search?l=Crystal&p=0&q=chars&type=Code, you'll how can be the impact of this change for those using `map`, `reverse`, `zip` etc. 

Do we count chars, bytes, codepoints? Not that clear just by reading the code. That's why there are all this underscore prefixes, which are fine if we want a functional style à la Erlang.

The main point was for method organization. If I am the only one to prefer to have only one (no more `chars.each` vs `each_char`), efficient (no Array), OOP way (`chars.each_with_index` vs `each_char_with_index`) to do a thing, without copying what is existing in Indexable (like `with_index`, `to_a`, `count` etc), then I close this.

Again, this isn't a major pain, but frequent – each time I use this methods I feel something hackish out place.

asterite:
@j8r The thing is that `String::Chars` is not efficient. You can do:

```crystal
string = # some string...
chars = string.chars
chars[0] # => ...
chars[10] # => ...
chars[20] # => ...
```

by making `string.chars` return `String::Chars` we effectively introduced a performance regression in the above code: it compiles, but accessing a string by char index is slow (it needs to UTF-8 decode it to know that, every time the method is called).

Right now `chars` returns an `Array(Char)`. Accessing that by index is as efficient as possible. Traversing those chars too. Reversing them too. It's expensive to build the array of chars, but once you get that then you can have an optimized structure to manipulate them (and if you take a look at the GitHub search you provided you'll see that's the case).

The use cases of chars as `Iterator` can be accessed with `each_char`, and this name can be immediately related to an iterator. A view of chars as an `Indexable` is not performant because of what I said above... so maybe me suggesting that was wrong in the first place because I didn't think of this performance issue.

j8r:
Thanks for the explanation @asterite. I guess it's the never ending memory vs perfomance debate.


`char_view` could

Issue #5801, opened 2018-03-10
No docs for Channel


Jens0512:
[The 0.24.1 (and current) api-docs](https://crystal-lang.org/api/0.24.1/Channel.html) has no documentation whatsoever for the `Channel` class .

This is currently quite a central piece in Crystal's concurrency, so I feel that this is important and should be documented for next release.

straight-shoota:
There is a [guide on concurrency](https://crystal-lang.org/docs/guides/concurrency.html) in the language reference which also includes some information about `Channel`. This could be used as a starting ground for improving API docs.

PR #7673, opened 2019-04-15
Add docs for Channel

j8r:
Fixes https://github.com/crystal-lang/crystal/issues/5801

I'm not very familiar with channels and its concepts, this docs can be improved.

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -57,6 +57,7 @@ abstract class Channel(T)
   end
 
   # Receive a value from the channel.
+  #
   # ```
   # channel = Channel(Int32).new
   # channel.send(1)
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -10,7 +10,7 @@ require "fiber"
 #   channel.send(nil)
 #   puts "After send"
 # end
-
+#
 # puts "Before receive"
 # channel.receive
 # puts "After receive"
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -71,7 +71,10 @@ abstract class Channel(T)
     receive_impl { raise ClosedError.new }
   end
 
-  # Receive a value from the channel, if any.
+  # Receives a value from the channel.
+  # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
+  # 
+  # Returns `nil` if the channel is closed or closes while waiting for receive.
   def receive?
     receive_impl { return nil }
   end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -62,7 +62,6 @@ abstract class Channel(T)
   # channel.send(1)
   # channel.receive # => 1
   # ```
-  #
   def receive
     receive_impl { raise ClosedError.new }
   end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -1,6 +1,6 @@
 require "fiber"
 
-# Channels enable concurrent communication between fibers.
+# A `Channel` enables concurrent communication between fibers.
 #
 #
 # ```
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -1,6 +1,6 @@
 require "fiber"
 
-# Channels enables to execute multiple tasks concurrently.
+# Channels enable concurrent communication between fibers.
 # ```
 # channel = Channel(Nil).new
 
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -10,8 +10,8 @@ require "fiber"
 #   channel.send(1)
 # end
 #
-# channel.receive #=> 0
-# channel.receive #=> 1
+# channel.receive # => 0
+# channel.receive # => 1
 # ```
 abstract class Channel(T)
   module SelectAction
@@ -70,7 +70,7 @@ abstract class Channel(T)
 
   # Receives a value from the channel.
   # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
-  # 
+  #
   # Returns `nil` if the channel is closed or closes while waiting for receive.
   def receive?
     receive_impl { return nil }
@@ -172,8 +172,7 @@ abstract class Channel(T)
     def wait
       @channel.wait_for_receive
     end
-    
-    
+
     def unwait
       @channel.unwait_for_receive
     end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -1,6 +1,7 @@
 require "fiber"
 
 # Channels enable concurrent communication between fibers.
+
 # ```
 # channel = Channel(Nil).new
 
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -2,6 +2,8 @@ require "fiber"
 
 # A `Channel` enables concurrent communication between fibers.
 #
+# They allow communicating data between fibers without sharing memory and without having to worry about locks, semaphores or other special structures.
+#
 # ```
 # channel = Channel(Int32).new
 #
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -157,57 +157,50 @@ abstract class Channel(T)
     ReceiveAction.new(self)
   end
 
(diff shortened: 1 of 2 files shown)


src/channel.cr
```diff
@@ -1,5 +1,19 @@
 require "fiber"
 
+# Channels enables to execute multiple tasks concurrently.
```
r00ster91: No, channels are for communicating between fibers. They don't execute multiple tasks concurrently, that's what fibers do.

src/channel.cr
```diff
@@ -1,5 +1,19 @@
 require "fiber"
 
+# Channels enable concurrent communication between fibers.
+# ```
```
Sija:
```suggestion

# ```
```

src/channel.cr
```diff
@@ -1,5 +1,19 @@
...
+# channel = Channel(Nil).new
+
```
Sija:
```suggestion
#
```

src/channel.cr
```diff
@@ -41,10 +55,18 @@ abstract class Channel(T)
     @closed
   end
 
+  # Receive a value from the channel.
+  # ```
+  # channel = Channel(Int32).new
+  # channel.send(1)
+  # channel.receive # => 1
+  # ```
+  #
```
Sija:
```suggestion
```

src/channel.cr
```diff
@@ -1,5 +1,19 @@
...
+# Channels enable concurrent communication between fibers.
+# ```
+# channel = Channel(Nil).new
+
+# spawn do
+#   puts "Before send"
+#   channel.send(nil)
+#   puts "After send"
+# end
+
```
Sija:
```suggestion
#
```

src/channel.cr
```diff
@@ -41,10 +56,17 @@ abstract class Channel(T)
     @closed
   end
 
+  # Receive a value from the channel.
+  # ```
```
Sija:
```suggestion
  #
  # ```
```

src/channel.cr
```diff
@@ -1,5 +1,20 @@
 require "fiber"
 
+# Channels enable concurrent communication between fibers.
+
+# ```
```
Sija:
```suggestion
#
# ```
```

src/channel.cr
```diff
@@ -127,61 +151,71 @@ abstract class Channel(T)
     ReceiveAction.new(self)
   end
 
-  # :nodoc:
```
z64: Docs comments on these are fine, but these look like implementation details that should retain the `:nodoc:` flag?

src/channel.cr
```diff
@@ -41,10 +57,18 @@ abstract class Channel(T)
...
+  # ```
+  # channel = Channel(Int32).new
+  # channel.send(1)
+  # channel.receive # => 1
+  # ```
   def receive
     receive_impl { raise ClosedError.new }
   end
 
+  # Receive a value from the channel, if any.
```
z64: The difference between `receive` and `receive?` is that `receive?` will not raise an exception if the channel is closed. Other `Channel` methods will raise exceptions if it is closed; I recommend reading [`channel_spec.cr`](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr) and grep for `expect_raises` to better understand when this is expected to occur, and maybe mention this in some docs ("Raises `Channel::ClosedError` if the channel is closed", etc..)

src/channel.cr
```diff
@@ -1,5 +1,21 @@
 require "fiber"
 
+# Channels enable concurrent communication between fibers.
+
```
Sija:
```suggestion
#
```

src/channel.cr
```diff
@@ -1,5 +1,21 @@
 require "fiber"
 
+# Channels enable concurrent communication between fibers.
```
straight-shoota:
```suggestion
# A `Channel` enables concurrent communication between fibers.
```

Singular should be used to describe a singular entity. The plural could be misinterpreted that you'd need multiple channels for concurrent communication.

src/channel.cr
```diff
@@ -1,5 +1,21 @@
...
+# ```
+# channel = Channel(Nil).new
+#
+# spawn do
+#   puts "Before send"
+#   channel.send(nil)
+#   puts "After send"
+# end
+#
+# puts "Before receive"
```
straight-shoota: The messages could be numbered to illustrate control flow.
```suggestion
# puts "1. Before receive"
```

src/channel.cr
```diff
@@ -41,10 +57,18 @@ abstract class Channel(T)
     @closed
   end
 
+  # Receive a value from the channel.
```
straight-shoota:
```suggestion
  # Receives a value from the channel.
  # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
  #
  # Raises `ClosedError` if the channel is closed or closes while waiting for receive.
```

src/channel.cr
straight-shoota: Yes, raised exceptions and other error behaviour (like returning `nil`) should be explicitly mentioned.

```suggestion
  # Receives a value from the channel.
  # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
  # 
  # Returns `nil` if the channel is closed or closes while waiting for receive.
```

Ditto for `#send`.

src/channel.cr
straight-shoota: I don't think these structs are part of the public API. Nor should they. Adding comments is great, but please keep `:nodoc:`.

j8r: I remove it to add a comment, and I forget to put it again.

src/channel.cr
```diff
@@ -1,5 +1,21 @@
 require "fiber"
 
+# A `Channel` enables concurrent communication between fibers.
+#
+#
```
Sija: These suggestions are really tiresome, desired outcome should be:

```cr
# A `Channel` ...
#
# ```
# ...
# ```
```

i.e. with 1 empty line between the doc comment and example below.

src/channel.cr
ysbaddaden: Don't document the Action structs unless you're writing actual implementation docs that will help developers to understand the internal API design.

src/channel.cr
ysbaddaden: I fail to see the point of the whole example. The `puts` don't mean anything and are unerelated to Channel. In fact a buffered and unbuffered channel will behave differently. I would prefer the example to spawn one fiber that will send a few messages then let the main fiber receive those messages in the order they were sent, which is the most important thing.

src/fiber.cr
```diff
@@ -129,6 +129,7 @@ class Fiber
     @resume_event ||= Crystal::EventLoop.create_resume_event(self)
   end
 
+  # Set the current fiber to wait, in order to eventually execute others fibers.
```
ysbaddaden: This is completely unrelated to documenting Channel. AFAIK developers are encouraged to use `sleep(0)` for the time being anyway.

j8r:
@ysbaddaden I whish I could but I have not the experience to go into the details, and tell pros and cons of buffered vs unbuffered channels. Mine barely reach `Channel(T).new`, `spawn`, `send` and `receive` :sweat_smile: 
That's a reason why I opened this PR, I think it's important to do our best to document this major language feature.

j8r:
@ysbaddaden I wish I could but I have not the experience to go into the details, and tell pros and cons of buffered vs unbuffered channels. Mine barely reach `Channel(T).new`, `spawn`, `send` and `receive` :sweat_smile: 
That's a reason why I opened this PR, I think it's important to do our best to document this major language feature.

j8r:
@sdogruyol Maybe, I don't know when I would have enough experience/knowledge with Channels to further improve the documentation. I won't probably the best one for this task.

Issue #291, opened 2014-12-03
Feature request: typed splats


sferik:
I would like to be able to define the following methods:

```crystal
# Fetch multiple users by their IDs
def find_users(*ids : Int32)
end

# Fetch multiple users by their usernames
def find_users(*usernames : String)
end
```

such that:

```crystal
find_users(1, 2, 3) # dispatches to the first version of the method
find_users("erik") # dispatches to the second version of the method
find_users(1, "erik") # error: no overload matches 'find_users' with types Int32, String
```

Essentially, I want the ability to define a method with an *arbitrary number* of parameters of a *specific type*. Of course, this is possible with an array but it should be possible to invoke the method without square braces (e.g. `find_users([1, 2, 3])`).

Basically, I’d like a splat of type T to be syntactic sugar for Array(T) when it appears as a parameter in a method definition? Do you think that would be possible?

asterite:
I'm reopening becuase we'll probably finally implement this

sferik:
\o/

vendethiel:
:+1: sounds really good :-).

asterite:
Oh, with this I (initially) mean that you'll be able to put a type restriction, not to have overloaded splats. We have to think a bit more about the overload case, specially because of the empty-tuple case.

asterite:
On second thought, I think we can also implement typed splats. In fact I have this in a stash :-)

The idea would be that typed splats don't match empty tuples. For the empty-tuple case use another overload. So:

```crystal
# Fetch multiple users by their IDs. Requires at least one ID.
def find_users(*ids : Int32)
end

# Fetch multiple users by their usernames. Requires as least one username.
def find_users(*usernames : String)
end

# Maybe return all of them? (just for demo)
def find_users
end

find_users(1, 2, 3)       # matches first
find_users("a", "b", "c") # matches second
first_users               # matches third
```

Of course, without the third overload the last call would give a compile-time error.

[Multi-dispatch over tuple types](https://github.com/manastech/crystal/issues/223) is still a different issue, so implementing this won't magically make #223 work (but in the future it might).

Thoughts?

sferik:
\o/

vendethiel:
Great :).

PR #7674, opened 2019-04-15
Restore File.join for single arguments

bcardiff:
After [a change in file.cr in #5635](https://github.com/crystal-lang/crystal/pull/5635/files#diff-f8eae7ee99e3c68635bfdd55a95eb2c0R693) `File.join(str)` no longer compiles.

Although it can be argued that if someone uses that call it can be removed, it is useful in case of tuple and splats.

NB: found it while smoke testing the release with amber

### DIFF
--- a/src/file.cr
+++ b/src/file.cr
@@ -690,8 +690,8 @@ class File < IO::FileDescriptor
   # File.join("foo/", "/bar/", "/baz")   # => "foo/bar/baz"
   # File.join("/foo/", "/bar/", "/baz/") # => "/foo/bar/baz/"
   # ```
-  def self.join(first : String | Path, *parts : String | Path) : String
-    Path.new(first, *parts).to_s
+  def self.join(*parts : String | Path) : String
+    Path.new(*parts).to_s
   end
 
   # Returns a new string formed by joining the strings using `File::SEPARATOR`.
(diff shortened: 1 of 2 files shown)


src/file.cr
```diff
@@ -694,6 +694,11 @@ class File < IO::FileDescriptor
     Path.new(first, *parts).to_s
   end
 
+  # :ditto:
+  def self.join(first : String | Path) : String
```
asterite: Does it work if you change the method above to just be `def self.join(*parts : String | Path) : String` ?

bcardiff: Yes! 😸

oprypin:
What about adding a test for 0 arguments?

bcardiff:
@oprypin would you expect `File.join` with no arg to compile?

oprypin:
@bcardiff, with the current state of the source code, yes, actually, seems like it would compile.
But there should be an explicit decision regarding whether it should compile (in my opinion it shouldn't).
And it's too bad that we can't really have a test for "doesn't compile".

bcardiff:
@oprypin ok, it does not compile :-)

```
$ ./bin/crystal eval 'puts File.join'
Error in line 1: no overload matches 'File.join'
Overloads are:
 - File.join(parts : Array | Tuple)
 - File.join(*parts : String | Path)
```

oprypin:
Oh :open_mouth: then I'm missing some knowledge about \*args in Crystal. Thought it would have been 0-or-more.

oprypin:
Some clarification on this behavior

```crystal
def f(*x : String)
  p x
end

f()
```
```
Error in line 5: no overload matches 'f'
Overloads are:
 - f(*x : String)
```
---
```crystal
def f(*x)
  p x
end

f()
```
```
{}
```

asterite:
@oprypin  https://github.com/crystal-lang/crystal/issues/291#issuecomment-95576090 and there's a 105% chance that it's not documented.

In Ruby `File.join` returns an empty string, I don't think that as a bad thing.

PR #7678, opened 2019-04-16
Move bsearch-related private top-level methods to Range

r00ster91:
This moves methods that are used for implementing binary search into `Range` instead of leaving it outside on the top-level, they don't belong on the top-level. Then the `private` attribute of the methods also makes sense because private methods on the top-level can still be called and `private` does nothing. I found these when doing `{% puts @type.methods %}`.

### DIFF
--- a/src/range/bsearch.cr
+++ b/src/range/bsearch.cr
@@ -1,73 +1,73 @@
-{% for p in [64, 32] %}
-  # Cast integer as floating number value with keeping binary structure.
-  private def int_as_float(i : Int{{ p }})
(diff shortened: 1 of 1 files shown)


asterite:
They make a lot of sense as private top-level functions: they can't be called from anywhere else other than that file. `private` methods on a type can be called when reopening the type.

bew:
https://carc.in/#/r/6qen nice one, looks like a bug, can you open an issue instead?

r00ster91:
Yes: #7681.

PR #7682, opened 2019-04-16
HTTP: Make return value of Response#status_code an Int32

bcardiff:
In #7247 the idea was to keep the `Response#status_code` API. The setter should return the same `Int32` value that it is assigned and not the new `HTTP::Status`. cc: @dwightwatson

---

NB: Discovered while testing amber controller fixtures where

```crystal
respond_with do
  html ->{ redirect_to "/some_path" }
end
```

stopped compiling because the block was no longer returning `Int32` since the value was determined by the setter called in [Amber's HTTP::Server::Context#halt!](https://github.com/amberframework/amber/blob/master/src/amber/router/context.cr#L66).

### DIFF
--- a/spec/std/http/server/server_spec.cr
+++ b/spec/std/http/server/server_spec.cr
@@ -146,7 +146,8 @@ module HTTP
       it "sets status code" do
         io = IO::Memory.new
         response = Response.new(io)
-        response.status_code = 201
+        return_value = response.status_code = 201
+        return_value.should eq 201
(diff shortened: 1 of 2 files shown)


asterite:
Ruby makes the return value of `foo.x = y` to be `y` independently of what the `x=` method returns. We should probably do the same in Crystal (I'm not sure how yet).

bcardiff:
The `@last` and type of a call to `method=` might need to be a special case then.

PR #7686, opened 2019-04-17
Fix Path docs

r00ster91:
The `Path` docs are currently not very helpful: https://crystal-lang.org/api/0.28.0/Path.html.


r00ster91:
This is done in #7718 now.

PR #7690, opened 2019-04-18
Improve Colorize

### LINKED ISSUE #6736: Remove manual symbols in Crystal with automatic casting enums?
I'm personally not sure if there are other examples, but #6427 could be easily replaced with an enum listing all the colours instead.

This has several advantages:
- if a user enters in a wrong colour, the error is a compile-time error, not a runtime error
- the code for `colorize` can be much cleaner

However, this might require an overhaul of the entire module since a lot of code in that module is dependent on the symbols themselves - for example, this section of code:

```crystal
{% for name in COLORS %}
  def {{name.id}}
    @fore = ColorANSI::{{name.camelcase.id}}
    self
  end

  def on_{{name.id}}
    @back = ColorANSI::{{name.camelcase.id}}
    self
  end
{% end %}
```

As said in the beginning, I'm not sure if there are any other examples of enums being manually duplicated as symbols within the compiler, but if they are I think they should be converted fully to enums as well, to conform with #6074.

r00ster91:
This:
* Adds more overall documentation to `Colorize` and improves some.
* Removes `Colorize.on_tty_only!` and makes it the default so you don't have to put it at the beginning of every program that uses `Colorize` because you don't want that your program behaves badly in non-TTY terminals (i.e. that the ANSI escape codes become visible). Most people don't do this/forget it.
* Uses the explosive automatic casting feature instead of macros in order to support symbols which results in cleaner code and compile time errors instead of run time errors.
  Fixes #6736 because as to my knowledge there are no other examples where enums are being manually duplicated as symbols.
* Adds `Object.colorize(r, g, b)` which is a syntactic sugar for `Object.colorize(Colorize::ColorRGB.new(r, g, b))`. It's a bit of an alias but I'd say it's rather syntactic sugar and less worse than all the other aliases in `Colorize`.
* Fixes `"hello".colorize(:red).inspect` returning `"hello"` in red instead of `"\e[31mhello\e[0m"` which is probably what you want when you inspect a colored string.
* Deprecates the alias `bright`. See https://github.com/crystal-lang/crystal/pull/7690#discussion_r279160997.


src/colorize.cr
```diff
@@ -138,18 +140,25 @@ def with_color
...
   end
 
-  def colorize(fore)
+  # Turns `self` into a `Colorize::Object` and colors it.
+  def colorize(fore : Color)
     Colorize::Object.new(self).fore(fore)
   end
+
+  # Turns `self` into a `Colorize::Object` and colors it with an RGB color.
+  def colorize(r : UInt8, g : UInt8, b : UInt8)
```
j8r: `red, green, blue` will be more explicit

src/colorize.cr
```diff
@@ -214,10 +226,33 @@ module Colorize
...
+    Reverse
+    # Makes the text invisible.
+    Hidden
+  end
 end
 
+# A colorize object colors and decorations can be applied to.
 struct Colorize::Object(T)
-  private MODE_DEFAULT   = '0'
+  private MODE_NONE      = '0'
```
j8r: `MODE_*` private constants can be changed to a private enum annotated with `@[Flags]`, which will bring `Mode::None`.

r00ster91: I would like to not make these constants an enum because enums can't contain `Char`s so they would need to be integers and that would mean a bunch of `to_s`s which I would like to avoid.

j8r: The performance difference isn't significant. I think this enum should be exposed and documented, in order to know available modes.

```cr
require "benchmark"

io = File.open File::DEVNULL, "w"

enum E
  A
  B
  C
end

E_A = '1'
E_B = '2'
E_C = '3'

Benchmark.ips do |x|
  x.report("enum") do
    io << E::A
    io << E::B
    io << E::C
  end

  x.report("char") do
    io << E_A
    io << E_B
    io << E_C
  end
end
```
```
enum  42.39M ( 23.59ns) (± 6.34%)  0.0B/op   1.72× slower
char  73.07M ( 13.69ns) (± 4.67%)  0.0B/op        fastest
```

r00ster91: This enum is actually exposed and documented here: [line 235](https://github.com/crystal-lang/crystal/pull/7690/files#diff-9fc1ac5e1582956c996fd04dde1f3049R235). These constants are basically part of this enum and just there because the `Mode` enum is a flags enum and not in normal order which is required here: [line 430](https://github.com/crystal-lang/crystal/pull/7690/files#diff-9fc1ac5e1582956c996fd04dde1f3049R430).

j8r: The enum already exists, why not directly use it?

r00ster91: Because the `Mode` enum can't be used directly. The `Mode` enum is a flags enum and is thus not in normal order from 1 to 7 but that is required in [line 430](https://github.com/crystal-lang/crystal/pull/7690/files#diff-9fc1ac5e1582956c996fd04dde1f3049R430). The constants however are in the correct normal number order from 1 to 7.

j8r: Sorry I don't understand, what is line 430?

r00ster91: The links broke because of a commit, the line is at [404](https://github.com/crystal-lang/crystal/pull/7690/files#diff-9fc1ac5e1582956c996fd04dde1f3049R404) now.

src/colorize.cr
```diff
@@ -226,108 +231,98 @@ struct Colorize::Object(T)
...
     end
   end
 
+  # Inspects this object and makes the ANSI escape codes visible.
   def inspect(io : IO) : Nil
-    surround(io) do
-      @object.inspect(io)
-    end
+    internal_io = IO::Memory.new
+    (surround(internal_io) do
```
Sija: Why extra `()`?

src/colorize.cr
j8r: Maybe `Bright` can be removed, it has the same effect has bold ([reference](https://en.wikipedia.org/wiki/ANSI_escape_code))

Anyway you can do
```cr
{% for mode in Mode.constants %}
  {% if !["All", "None", "Bright"].includes? mode.stringify %}\
            Mode::{{mode.id}}
  {% end %}      
{% end %}  
```
Then drop private constants.

j8r: If you really want chars, a counter can be also used

src/colorize.cr
r00ster91: Oh, you are right. I probably forgot this when testing something.

src/colorize.cr
r00ster91: I think bright and bold both exist because on some terminals they actually make the text **bold** and on some terminals (like mine) it just brightens up the color. So you can choose one based on how they act in your terminal.

So we are talking about this, right?
https://github.com/crystal-lang/crystal/blob/e3b84b632e7012d079fe57a1b5df4b7f50593b71/src/colorize.cr#L399-L408
Here it iterates through the mode array (`%w(bold dim underline blink reverse hidden)`) and generates an if check for each mode and if the local variable `mode` (not the one from the macro `for`) includes for example `Mode::Blink`, a semicolon is appended and then `MODE_BLINK`'s value is appended, `'5'`. So everything fine. Now when I however use the value from the `Mode` enum, I would append `8` which is wrong and would result in an incorrect ANSI escape code.
Or am I misunderstanding something?

j8r: The `Mode` enum is incorrect. After looking at [the VT100 specification](https://vt100.net/docs/vt100-ug/chapter3.html),  there is table:
```
0 | Attributes off
1 | Bold or increased intensity
4 | Underscore
5 | Blink
7 | Negative (reverse) image
```

The enum should be
```cr
enum Mode
  Normal     = 0
  Bold       = 1
  Underscore = 4
  Blink      = 5
  Reverse    = 7
  Conceal    = 8
end
```
This enum maps to incorrect values, and can't be used outside `Colorize`.

j8r: According to the [VT510 specification](https://vt100.net/docs/vt510-rm/chapter12.html), more values can be added to the enum.

r00ster91: The values of the `Mode` enum don't matter. It's just a flags enum so you can have multiple modes. The values of the private constants matter and they are correct according to https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters.

j8r: Why don't make them correct, and only use this very enum?
The private constants won't be needed then.

r00ster91: As I said the `Mode` enum needs to be a flags enum so that you can have multiple modes at once for one text. Otherwise you couldn't make a text both underlined and bold.

src/colorize.cr
```diff
@@ -226,108 +245,100 @@ struct Colorize::Object(T)
...
     @back = ColorANSI::Default
-    @mode = 0
+    @mode = Mode::None
     @enabled = Colorize.enabled?
   end
 
-  {% for name in COLORS %}
-    def {{name.id}}
-      @fore = ColorANSI::{{name.camelcase.id}}
+  {% for color in COLORS %}
```
straight-shoota: I don't think we need `COLORS` anymore.


```suggestion
  {% for color in ColorANSI.constants %}
```

Ditto for `MODES`.

src/colorize.cr
```diff
@@ -401,12 +412,12 @@ struct Colorize::Object(T)
         printed = true
       end
 
-      unless mode_is_default
+      unless no_mode
         # Can't reuse MODES constant because it has bold/bright duplicated
-        {% for name in %w(bold dim underline blink reverse hidden) %}
-          if mode.bits_set? MODE_{{name.upcase.id}}_FLAG
+        {% for mode in %w(Bold Dim Underline Blink Reverse Hidden) %}
```
straight-shoota:
```suggestion
        Mode.values.each do |mode|
```

r00ster91: I assume you mean
`{% for mode in Mode.constants.reject { |constant| constant == "Bright" } %}`? However that's even longer than `{% for mode in %w(Bold Dim Underline Blink Reverse Hidden) %}` so maybe this little boilerplate code is fine.

straight-shoota: No, I meant a runtime loop. I don't think we need a macro for this. The only reason for the macro is to use `MODE_*` constants. But this could just be a simple mapping from `Mode` to mode id.

r00ster91: So a hash? I'm not sure if this little change is really worth the slow down. I would like to keep this as performant as possible and avoid hash look-ups especially in this method which is pretty frequently called.
Unfortunately I can't use a named tuple either because only symbols and strings are accepted as keys, not constants like `Mode::Bold`.

straight-shoota: Nobody's suggesting a hash. Simpy use `case ... when`

src/colorize.cr
```diff
@@ -218,9 +218,9 @@ module Colorize
   @[Flags]
   enum Mode
     # Makes the text bold.
-    Bold = 1
+    Bright
```
straight-shoota: This doesn't seem right.

r00ster91: This is the behavior when `Bright` and `Bold` are `1`:
```cr
"hello".colorize.bright.bright.inspect # => "\e[1;1mhello\e[0m"
"hello".colorize.bright.bold.inspect # => "\e[1;1mhello\e[0m"
"hello".colorize.bold.inspect # => "\e[1;1mhello\e[0m"
"hello".colorize.bright.inspect # => "\e[1;1mhello\e[0m"
```

And this is the behavior after this commit, without them:
```cr
"hello".colorize.bright.bright.inspect # => "\e[1mhello\e[0m"
"hello".colorize.bright.bold.inspect # => "\e[1;1mhello\e[0m"
"hello".colorize.bold.inspect # => "\e[1mhello\e[0m"
"hello".colorize.bright.inspect # => "\e[1mhello\e[0m"
```
The behavior afterwards is better because many useless `;1`s are omitted and this decreases the terminal ANSI escape code parsing time. There is no visual difference between `"\e[1mhello\e[0m"` and `"\e[1;1mhello\e[0m"`.

r00ster91: Also, I've found a disadvantage of no longer using macros and the private `MODE_*` constants. 
This is the behavior without this PR:
```cr
"hello".colorize.bright.bright.to_s.inspect # => "\e[1mhello\e[0m"
"hello".colorize.bright.bold.to_s.inspect # => "\e[1mhello\e[0m"
"hello".colorize.bold.to_s.inspect # => "\e[1mhello\e[0m"
"hello".colorize.bright.to_s.inspect # => "\e[1mhello\e[0m"
```
Here, all unnecessary `;1`s are omitted, the perfect behavior.
But with this PR and using this long run time `case` rather than macros, `"hello".colorize.bright.bold.inspect` returns `"\e[1;1mhello\e[0m"` (as you can also see in the other comment in the second code block).

So maybe it would be better afterall to revert to the private `MODE_*` constants since it has more disadvantages than advantages (slower ANSI escape code generation because of run time `case` and what I mentioned above).

j8r: The main problem is Bright and Bold are the same, the API suggests they aren't.
For example one can choose bright, and in its terminal it prints a bold text or vice-versa.

Either we choose to call it `Bright` or `Bold`, but not both, and we mention that the output may vary across terminals.
I'd vote to keep `Bold`, because both the official [VT100 docs](https://vt100.net/docs/tp83/chapter12.html) and Wikipedia use this therm.

r00ster91: I don't really agree. When we keep `Bold` then it wouldn't make sense when you get bright text, when we keep `Bright`, it wouldn't make sense when you get bold text.

I will revert to not using a `case` now. It doesn't really have advantages and this PR is rather stalled now after this.

j8r: Yes but that's how it works, Terminals interprets as they wish. Even colors can be represented as another one, I've seen yellow close to a brown.
`Bright` or not, the text is nearly always at least `Bold` - wider characters.

r00ster91: Well not for me for example:
image

j8r: After looking at other libraries, the consensus seems to be `bold`, also on par with the VT100 documentation:
- https://github.com/fazibear/colorize/blob/master/lib/colorize/class_methods.rb#L78
- https://github.com/logrusorgru/aurora/blob/master/color.go#L60
- https://github.com/ogham/rust-ansi-term/blob/master/src/ansi.rs#L35

I can see issues opened by users saying they set bright and it does bold, or vice-versa, and this doesn't work. Look at this questions:
- [gnome-terminal use bright colors for bold](https://askubuntu.com/questions/875102/gnome-terminal-use-bright-colors-for-bold)
- [How do I make XTerm not use bold?](https://superuser.com/questions/15226/how-do-i-make-xterm-not-use-bold)
- [Disable Using Bright Colors for Bold Font in gnome-terminal](https://superuser.com/questions/516123/disable-using-bright-colors-for-bold-font-in-gnome-terminal)

I suggest having at least a description:
> Makes the text bold **or** bright, depending the terminal

j8r: After looking at other libraries, the consensus seems to be `bold`, also on par with the VT100 documentation:
- https://github.com/fazibear/colorize/blob/master/lib/colorize/class_methods.rb#L78
- https://github.com/logrusorgru/aurora/blob/master/color.go#L60
- https://github.com/ogham/rust-ansi-term/blob/master/src/ansi.rs#L35

I can see issues opened by users saying they set bright and it does bold, or vice-versa, and this doesn't work. Look at this questions:
- [gnome-terminal use bright colors for bold](https://askubuntu.com/questions/875102/gnome-terminal-use-bright-colors-for-bold)
- [How do I make XTerm not use bold?](https://superuser.com/questions/15226/how-do-i-make-xterm-not-use-bold)
- [Disable Using Bright Colors for Bold Font in gnome-terminal](https://superuser.com/questions/516123/disable-using-bright-colors-for-bold-font-in-gnome-terminal)

I suggest having at least a description:
> Makes the text bold and/or bright, depending the terminal

r00ster91: Actually, yeah I think it wouldn't be so bad to remove `Mode::Bright` because "bold" can really mean both "bright" and "bold" because sometimes "bold" is also used in other cases like "you made a bold statement there" which means you made like a "powerful"/"brave"/"big" statement and these words are in some way similar to "bright".

r00ster91: Actually, it would be better to deprecate this method at least because it's used sometimes: https://github.com/search?l=Crystal&q=bright&type=Code.

r00ster91:
I resolved the conflicts, fixed some bugs and deprecated a method. Can this please be reviewed?

samples/2048.cr
```diff
@@ -3,24 +3,26 @@
...
+      128 => {Color::Red, Color::Yellow},
+      256 => {Color::Magenta, Color::Yellow},
+      512 => {Color::White, Color::Yellow},
+     1024 => {Color::White, Color::Yellow},
+     2048 => {Color::White, Color::Yellow},
+     4096 => {Color::White, Color::Black},
+     8192 => {Color::White, Color::Black},
+    16384 => {Color::White, Color::Black},
+    32768 => {Color::White, Color::Black},
+    65536 => {Color::White, Color::Black},
```
Sija: You can leave the symbols by adding ` of Int32 => {Color, Color}` to this hash declaration.

src/colorize.cr
```diff
@@ -61,95 +72,83 @@
...
 
-  def colorize(fore)
+  # Turns `self` into a `Colorize::Object` and colors it.
+  def colorize(fore : Color)
     Colorize::Object.new(self).fore(fore)
   end
+
+  # Turns `self` into a `Colorize::Object` and colors it with an RGB color.
+  def colorize(red : UInt8, green : UInt8, blue : UInt8)
+    Colorize::Object.new(self).fore(Colorize::ColorRGB.new(red, green, blue))
```
Sija:
```suggestion
    colorize(Colorize::ColorRGB.new(red, green, blue))
```

samples/2048.cr
r00ster91: Oh, right!

r00ster91: Hmm, no it doesn't really seem to work:
```cr
  TILES = {
        0 => {:white, nil},
        2 => {:black, :white},
        4 => {:blue, :white},
        8 => {:black, :yellow},
       16 => {:white, :red},
       32 => {:black, :red},
       64 => {:white, :magenta},
      128 => {:red, :yellow},
      256 => {:magenta, :yellow},
      512 => {:white, :yellow},
     1024 => {:white, :yellow},
     2048 => {:white, :yellow},
     4096 => {:white, :black},
     8192 => {:white, :black},
    16384 => {:white, :black},
    32768 => {:white, :black},
    65536 => {:white, :black},
  } of Int32 => {Colorize::ColorANSI, Colorize::ColorANSI?}
```
```

In samples/2048.cr:6:11

 6 | TILES = {
             ^-
Error: no overload matches 'Hash(Int32, Tuple(Colorize::ColorANSI, Colorize::ColorANSI | Nil))#[]=' with types Int32, Tuple(Symbol, Nil)

Overloads are:
 - Hash(K, V)#[]=(key : K, value : V)
```

Sija: That's weird... it works fine without tuples, see https://carc.in/#/r/7nts

Looks like it might be a bug /cc @asterite

r00ster91: I created #8333.

Blacksmoke16:
Bump, this still on anyone's radar?  Seems it needs another rebase...

r00ster91:
Hello, I don't think there is interest in this PR but I resolved the conflicts anyway. It's unfortunate that this wasn't merged before 1.0 because it does some breaking changes like removing `on_tty_only`. This makes it hard to merge now.

straight-shoota:
I think most of this is really good, but it's also a bunch of different changes merged together. It's hard to review. IMO it would be easier to merge in spearate chunks.

Ideally, I think each of the bullet points in the OP should be a separate PR. Maybe some could be merged, but to me it seems they're for the most part independent changes.

As you already mentioned, we can't simply remove `on_tty_only!` for backward compatibility. But there's no need to. We can just keep this method and make it deprecated.

src/colorize.cr
```diff
@@ -226,120 +220,147 @@ module Colorize
...
 
-    raise ArgumentError.new "Unknown mode: #{mode}"
-  end
+  # Adds *mode* to the text's decorations.
+  def mode(mode : Mode)
+    # TODO: Remove this in the future
+    if mode == Mode::Bright
+      puts "Warning: The text decoration `bright` is deprecated. Please use `bold` instead.".colorize(:light_yellow) unless @@warning_printed
+      @@warning_printed = true
+    end
```
Blacksmoke16, line 315: Do deprecation warnings work on enum members?  Don't think the annotation existed when this PR was made but that would prob be more ideal.

r00ster91: Yes, it would be nice if that worked but they don't work.

r00ster91:
I have re-added `on_tty_only!` so this PR should no longer break anything.

HertzDevil:
`Object.colorize(r, g, b)` is resolved by #11832.

PR #7691, opened 2019-04-18
Change String#to_i to parse octals with prefix 0o

### LINKED ISSUE #7689: String#to_i with prefix raises on 0
The following code raises an error:

```crystal
"0".to_i(prefix: true)
```

```text
Unhandled exception: Invalid Int32: 0 (ArgumentError)
```

https://play.crystal-lang.org/#/r/6qsn

As far as I can tell from the source, it sees the leading zero and thinks it's an octal.
Then it aborts because it thinks there is no more to the number.
I have a scenario where I'm parsing numbers that may or may not have a prefix,
and using a built-in method is certainly preferred over writing my own.
This is a bit of an odd scenario, since it is expecting a prefix, but the following works:

```crystal
"5".to_i(prefix: true)
```

Perhaps this just needs to be documented, but it should be a simple fix.

Crystal version 0.27.2

icy-arctic-fox:
This is a breaking change.

Previously, numbers starting with 0 would be parsed as octals. Now that is not the case, and they will be parsed as base-10. The `0o` prefix must be present to treat it as an octal. This addresses parsing 0 with `prefix: true` raising an error.

This PR should resolve #7689 

Update:
The current behavior in Crystal can be preserved by adding `leading_zero_is_octal: true` to the `#to_i` calls. By default however, `leading_zero_is_octal` is false and expects `0o` as a prefix. The YAML parser uses this so that both formats are recognized.


icy-arctic-fox:
Turns out this would also break YAML parsing of integers. The tests show that the YAML parser still behaves in the old way, where a `0` prefix means octal, instead of `0o`. Working on a fix for that now.

asterite:
Maybe a `leading_zero_is_octal = false` argument is good enough (because some software, like YAML, might still need this behavior).

icy-arctic-fox:
> Maybe a `leading_zero_is_octal = false` argument is good enough (because some software, like YAML, might still need this behavior).

I agree. There's a lot of places that still use a leading zero for octal notation. Linux permissions being one of them - 0755, 0644. Don't want to break that.

src/string.cr
```diff
@@ -539,16 +543,31 @@ class String
...
-        when 'b'
-          base = 2
-          ptr += 1
-        when 'x'
-          base = 16
-          ptr += 1
-        else
-          base = 8
-        end
+        found_digit = case ptr.value.unsafe_chr
```
Sija: I'd refactor this to create a new variable with default (`found_digit = false`), then assign it to `true` in just one branch, where it's needed.

icy-arctic-fox: Yup, that code is silly. I was too fixated on using a return value from the case statement.

src/string.cr
```diff
@@ -546,10 +550,20 @@ class String
...
           base = 16
           ptr += 1
-        else
+        when 'o'
           base = 8
+          ptr += 1
+        else
+          if leading_zero_is_octal
+            base = 8
+            ptr -= 1
```
straight-shoota: Why decrement the pointer?

straight-shoota:
The YAML changes look fine.
According to [YAML 1.2 Core Spec](https://yaml.org/spec/1.2/2009-07-21/spec.html#id2604185), octal numbers are prefixed by `0o`. `0` prefix is not mentioned, but it's implicitly included in the regex for base 10 integers. But it's probably okay to interpret them as octal numbers. That would be in accordance with the `tag:yaml.org,2002:int` [YAML 1.1 Integer type](https://yaml.org/type/int.html) (which the Core Spec references) where an octal number is represented by a `0` prefix and `0o` is not mentioned in the valid value spec.
So this is kind of contrdictory, but I guess it's fine to simply resolve both prefixes to octal number.

src/string.cr
icy-arctic-fox: After looking closer, I don't think it is needed. It was to cover the case where there's a leading zero, but no `o` right after it. It would handle strings like `0123` and `0abc` (invalid). I think I put it in because there's no prefix character (like b or x), so it needed to back up.

icy-arctic-fox: Travis [build 15307.1](https://travis-ci.org/crystal-lang/crystal/builds/527505579) failed when this was removed. Looks like it broke parsing of `!!int 0` in YAML. I'll see if there's a way I can address this without decrementing the pointer, unless you're ok with putting it back in.

asterite:
Sorry, I tried to fix the conflicts after merging another PR and I forgot some stuff (I think default values).

@icy-arctic-fox Do you have time to fix this? I think CI passed fine, it just had a hiccup in the previous run.

icy-arctic-fox:
Looks like it should be fixed. Somehow the merge dropped one default value.

PR #7694, opened 2019-04-19
Implement named/pos_args annotation methods

### LINKED ISSUE #7692: Extra Annotation methods
While writing https://github.com/crystal-lang/crystal-book/pull/335, I came across some places that annotations fall a bit short, mainly around positional values.  The `Annotation` macro class could do with some extra methods.

Currently positional arguments can only be accessed if the user knows how many there will be at compile time.  I.e. accessing each one with a `NumberLiteral`.  However this makes it not very useful IMO.

I was thinking of adding:
* size - Would return the total number of arguments on the annotation (positional + named).
  * Would allow for dynamic access of positional args.
    * `{% for idx in 0...Type.annotation(MyAnnotation).size %}`.
    * `idx` could then be used to access the positional value on the annotation.
    * If named arguments were used as well, accessing them would just return `nil`.
* keys - Would return the keys of all named arguments on the annotation.
  * Would allow for reading values off of an annotation in a loop.
* values - Would return all positional values (and maybe named arg values?)
  * Would allow looping over the values themselves, without directly having to read them off the annotation.  Most useful for positional values.

My main concerns are if an annotation has both positional AND named arguments, what should the `values` method return?  Or, how should iterating using `size` work if you have a named argument with the same key as the index?

Possibly could have separate methods for `values`/`size`?  One for positional and another for `named` arguments?

Thoughts?

Blacksmoke16:
Resolves #7692.

Adds `#named_args` and `#args` methods to `Annotation`.

### DIFF
--- a/spec/compiler/semantic/annotation_spec.cr
+++ b/spec/compiler/semantic/annotation_spec.cr
@@ -12,7 +12,43 @@ describe "Semantic: annotation" do
     type.name.should eq("Foo")
   end
 
-  describe "#*_args" do
+  describe "#args/named_args" do
+    describe "#args" do
+      it "returns an empty TupleLiteral if there are none defined" do
+        assert_type(%(
+          annotation Foo
+          end
+
+          @[Foo]
+          module Moo
+          end
+
+          {% if (args = Moo.annotation(Foo).args) && args.is_a? TupleLiteral && args.empty? %}
+            1
+          {% else %}
+            'a'
+          {% end %}
+        )) { int32 }
+      end
+
+      it "returns a TupleLiteral if there are positional arguments defined" do
+        assert_type(%(
+          annotation Foo
+          end
+
+          @[Foo(1, "foo", true)]
+            module Moo
+          end
+
+          {% if Moo.annotation(Foo).args == {1, "foo", true} %}
+            1
+          {% else %}
+            'a'
+          {% end %}
+        )) { int32 }
+      end
+    end
+
     describe "#named_args" do
       it "returns an empty NamedTupleLiteral if there are none defined" do
         assert_type(%(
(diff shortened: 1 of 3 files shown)


spec/compiler/semantic/annotation_spec.cr
```diff
@@ -12,6 +12,97 @@ describe "Semantic: annotation" do
...
+      )) { int32 }
+      end
+
+      it "returns a NamedTupleLiteral if there are named arguments defined" do
+        assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo(extra: "three", "foo": 99)]
+          module Moo
```
bcardiff: wrong formatting

spec/compiler/semantic/annotation_spec.cr
```diff
@@ -12,6 +12,97 @@ describe "Semantic: annotation" do
...
+      )) { int32 }
+      end
+
+      it "returns a TupleLiteral if there are positional arguments defined" do
+        assert_type(%(
+        annotation Foo
+        end
+
+        @[Foo(1, "foo", true)]
+          module Moo
```
bcardiff: :ditto:

src/compiler/crystal/macros.cr
```diff
@@ -874,9 +874,17 @@ module Crystal::Macros
...
+    # used on `self`.
     def [](name : SymbolLiteral | StringLiteral | MacroId) : ASTNode
     end
+
+    # Returns a `NamedTupleLiteral` representing the named arguments on `self`.
+    def named_args : NamedTupleLiteral
+    end
+
+    # Returns a `TupleLiteral` representing the positional arguments on `self`.
+    def pos_args : TupleLiteral
```
bcardiff: I would prefer to name them `args`, like in `Call` `ASTNode`.

Blacksmoke16: How would you differentiate between the positional/named args then?

Blacksmoke16:
Good for review again @bcardiff @straight-shoota

bew:
Imo `args` should return something with pos + named args, that why i suggested `pos_args` in the first place to target only positional args. (the same comment goes for Call)

Blacksmoke16:
@bcardiff @bew Would it be worth to add `post_args` back in addition to the other two?  Then you could either target all args, only named, or only positional?

Blacksmoke16:
Thinking of this more, the challenge with returning them both would be how to do that, since they are of different "types".  Could have `#args` return like `Tuple(Tuple(Positional), NamedTuple(Named))` Or even a named tuple like `NamedTuple(positional: Tuple(Positional), named: NamedTuple(Named))`.

Blacksmoke16:
@bcardiff @bew @straight-shoota Could you review this again?  I added `#args` method that will return a named tuple representing both positional and named args.

`{named: {foo: "bar"}, positional: {1, true, 17}}`

So now it's possible to get positional arguments, named arguments, or both off of an annotation.

RX14:
I'd prefer not to return a named tuple, no other macro methods behave like this.

Blacksmoke16:
@RX14 I'd also be fine with removing `#argss` and just keeping `#pos_args` and `#named_args`.

RX14:
@Blacksmoke16 that would be great

bcardiff:
Using only `args` (for positional args) and `named_args` to match the AST seems the best alternative to me.

Blacksmoke16:
@RX14 @bcardiff Done.

Blacksmoke16:
Works for me.  This is a big improvement over my current hack I'm using :P

straight-shoota:
Thank you @Blacksmoke16

PR #7695, opened 2019-04-19
Format: don't produce unnecessary newline before named args following heredoc

### LINKED ISSUE #7684: Formatter is not satisfied with formatting arguments with heredoc only once
```cr
def method(*arguments)
end

method(
  <<-HEREDOC,
  hello
  HEREDOC
  argument: 123
)
```
gets formatted to:
```cr
def method(*arguments)
end

method(
  <<-HEREDOC,
  hello
  HEREDOC

  argument: 123
)
```
and this gets formatted to:
```cr
def method(*arguments)
end

method(
  <<-HEREDOC,
  hello
  HEREDOC


  argument: 123
)
```
The formatter is satisfied after formatting the file two times. But I expect it to stop after the first format.

MakeNowJust:
Fixed #7684

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -1333,4 +1333,9 @@ describe Crystal::Formatter do
   # #7631
   assert_format "x.try &.[] 123"
   assert_format "x.try &.[]= 123, 456"
+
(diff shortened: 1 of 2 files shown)


straight-shoota:
@MakeNowJust needs a rebase

Issue #7680, opened 2019-04-16
BigDecimal missing ** method


jzakiya:
Converting this snippet from Ruby to Crystal.

```
require "bigdecimal"
 
LN2 = BigMath::log(2,16) 

def xx(n)
     ...
     2 * LN2  ** (n + 1)
end
```

Because Crystal has no ``**`` and ``log`` methods, I had to do this, which has less precision.

```
require "big"
include Math

LN2 = Math.log(2)

def xx(n)
     ...
     2 * LN2.to_big_d  ** (n + 1)     # can't do, no ** method for BigDecimal
     2 * LN2.to_big_f  ** (n + 1)     # not as much precision (digits) as Ruby
end
```
BigInt and BigFloat have ``**`` while BigDecimal and BigRational don't.

Maroo-b:
Is it possible to possible to add this change using func `pow_ui` from `LibGMP` https://github.com/crystal-lang/crystal/blob/master/src/big/lib_gmp.cr#L81? I saw it's used for `BigInt`

jzakiya:
Commented on correctness of https://github.com/crystal-lang/crystal/pull/7696

jwbuiter:
I'll try my hand at this for my first contribution

PR #7696, opened 2019-04-19
Creating ** method in BigDecimal and BigRational

ananagame:
Resolves [#7680](https://github.com/crystal-lang/crystal/issues/7680)

Adds `**` method to BigRational and BigDecimal

### DIFF
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -210,7 +210,7 @@ struct BigDecimal < Number
     k = self
     result = k.class.new(1)
     while exponent > 0
-      result *= k
+      result *= k if exponent & 0b1 != 0
       exponent = exponent.unsafe_shr(1)
       k *= k if exponent > 0
     end
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -199,11 +199,20 @@ struct BigDecimal < Number
     (self / other).floor
   end
 
-  # Returns the value of raising `self` to the power of *other* exponent.
-  def **(other : Int)
-    result = 1
-    other.times do
-      result *= self
+  # Returns the value of raising `self` to the power of *exponent* exponent.
+  # 
+  # Raise `ArgumentError` if *exponent* is negative.
+  def **(exponent : Int)
+    if exponent < 0
+      raise ArgumentError.new "Cannot raise a BigDecimal to a negative integer power."
+    end
+
+    k = self
+    result = k.class.new(1)
+    while exponent > 0
+      result *= k
+      exponent = exponent.unsafe_shr(1)
+      k *= k if exponent > 0
     end
     result
   end
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -199,6 +199,10 @@ struct BigDecimal < Number
     (self / other).floor
   end
 
+  def **(other : Int)
+    ([self] * other).reduce { |i, j| i * j }
+  end
+
   # Divides `self` with another `BigDecimal`, with a optionally configurable *max_div_iterations*, which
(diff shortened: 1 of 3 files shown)


Sija:
Tests please :)

Blacksmoke16:
Doc comments would be nice as well.

bcardiff:
Thanks @ananagame for implementing this. The current implementation creates an Array, potentially a big one. An iterative implementation would be faster and stress less the GC. If you are able to implement it that way it would be great. Thanks!

ananagame:
Sure @bcardiff

ananagame:
```crystal
require "big"
include Math

LN2 = Math.log(2)

def xx(n)
     puts 2 * LN2.to_big_d  ** (n + 1)     # new implementation of ** method
     puts 2 * LN2.to_big_f  ** (n + 1)     # not as much precision with BigFloat
end

xx(2)
```

And here are the results: 

```crystal
0.666049303977858932290481832092188441370295065354 # BigDecimal
0.666049303977858892586                            # BigFloat
```

src/big/big_decimal.cr
```diff
@@ -199,6 +199,15 @@ struct BigDecimal < Number
     (self / other).floor
   end
 
+  # Returns `self` at the power `other`
```
Sija:
```suggestion
  # Returns the value of raising `self` to the power of *other* exponent.
```

src/big/big_decimal.cr
```diff
@@ -199,6 +199,15 @@ struct BigDecimal < Number
...
+  def **(other : Int)
+    result = 1
```
Sija:
```suggestion
    result = self.class.new(1)
```

Sija:
Implementation in this PR is too naive, please look for instance how's `Int#**` implemented.

ananagame:
Sure, I'll work on it. Sorry for that!

jzakiya:
Maybe I'm just not understanding something, but I was looking at the code for implementing ``**`` and I don't see how it's correct. This is the code from https://en.wikipedia.org/wiki/Exponentiation_by_squaring

Modified to mimic committed code variables names. (Tested all versions in Ruby)
```
def power(x, exponent)
  result = 1
  while exponent > 0
    (result *= x; exponent -= 1) if exponent.odd?
    x *= x
    exponent >>= 1
  end
  result
end
```
I ran the committed code and it doesn't produce expected results.
How is that version correct?

ananagame:
You're right I forgot a condition in the code

Sija:
Tests please! :)

ananagame:
Here are the tests

```crystal 
require "big"
include Math

LN2 = Math.log(2)

def xx(n)
  puts 2 * LN2.to_big_d ** (n + 1) # new implementation of ** method
  puts 2 * LN2.to_big_f ** (n + 1) # not as much precision with BigFloat

  puts 2.to_big_d ** 10 # 2 to the power 10 using big decimal method
  puts 2 ** 10 # 2 to the power 10 using int method
end

xx(2)
```

Result
```
0.666049303977858932290481832092188441370295065354
0.666049303977858892586
1024
1024
```

Sija:
I meant unit tests.

jzakiya:
This is for my enlightenment.
Is there a performance reason for writing the code like this

```
result *= x if exponent & 0b1 != 0
exponent = exponent >> 1
```
versus
```
result *= x if exponent.odd?
exponent >>= 1
```
since the later is more concise and, IMHO, more idiomatic and human understandable?

Sija:
@ananagame What's the reason for closing?

ananagame:
I don't think that I have the skills to complete a good PR

PR #7697, opened 2019-04-19
add Uri#+ method, to get relative uri

kostya:


oprypin:
It can't be named  `+`. Nobody expects the addition of two items to completely get rid of the first item. Could be `URI.join`.

kostya:
i think you never expect that `+` modify something, that why `+=` exists

ysbaddaden:
Adding URIs sounds very weird, and the behavior is confusing(are domains and paths added?!).

Based on the proposed specs, this looks more like a `merge` instead, where `a.merge(b)` means that domains and paths from b (if present) will replace those from a, params will be merged from b to a, and so on.

kostya:
renamed to merge, this is method quite useful, when need to redirect from relative location for example, or parse links from html and expand relative ones it to original location.

oprypin:
Oh the behavior is very useful.
It's what a web browser does when you're at a page with the URL `a` and click a link with the URL `b`.

See also: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

Merge would be a really unusual name though. This is fully equivalent to joining filesystem paths
...at least in Python.
I was surprised about Crystal's behavior in [`File.join`](https://crystal-lang.org/api/0.28.0/File.html#join%28parts%3AArray%7CTuple%29%3AString-class-method), when joining paths starting with a slash.
[Python's behavior](https://docs.python.org/3/library/os.path.html#os.path.join) seems more useful (and that is the equivalency I was alluding to): 

> If a component is an absolute path, all previous components are thrown away and joining continues from the absolute path component.

straight-shoota:
I already have a branch for this with probably a more complete implementation (though I didn't have an in-depth look at this PR).  Transformations should be applied as specified in [RFC 3986 section 5.2](https://tools.ietf.org/html/rfc3986#section-5.2.2).
I've just been waiting for #7635 because it depend on the improved `#normalize`. I'll need to rebase and publish my branch in a few days, then we can compare the implementations.

straight-shoota:
Closed by #7716

PR #7703, opened 2019-04-21
Fix offsetof not being usable with macros

### LINKED ISSUE #7702: offsetof can't be used in macros
EDIT: Nevermind, I solved the issue. I'll open a pull request soon.

I just discovered a pretty important bug with `offsetof`, so sorry it's already released :(

`offsetof` produces weird syntax when used in macros:

```cr
macro do_nothing(exp)
  {{exp}}
end

do_nothing(offsetof(SomeType, @some_ivar))
```

I get the following error:
```
Error in src/main.cr:14: macro didn't expand to a valid program, it expanded to:

   1.   offsetof(SomeType, @some_ivar)SomeType@some_ivar
Syntax error in src/main.cr:14: unexpected token: SomeType

do_nothing(offsetof(SomeType, @some_ivar))
                    ^

do_nothing(offsetof(SomeType, @some_ivar))
^~~~~~~~~~
```
As you can see, the `offsetof(SomeType, @some_ivar)` becomes `offsetof(SomeType, @some_ivar)SomeType@some_ivar`. I have no idea where this comes from because `OffsetOf` basically does the same as `SizeOf` when compiling.

What would really help me fix this bug is knowing:
- Is it normal that AST nodes get transformed when used in macros?
- If so, why and where does this happen?

malte-v:
This fixes #7702.

### DIFF
--- a/spec/compiler/parser/to_s_spec.cr
+++ b/spec/compiler/parser/to_s_spec.cr
@@ -160,4 +160,5 @@ describe "ASTNode#to_s" do
   expect_to_s %(asm("nop" ::: "e" : "volatile"))
   expect_to_s %[(1..)]
   expect_to_s %[..3]
+  expect_to_s "offsetof(Foo, @bar)"
 end
--- a/src/compiler/crystal/syntax/to_s.cr
+++ b/src/compiler/crystal/syntax/to_s.cr
@@ -1268,6 +1268,7 @@ module Crystal
       @str << ", "
       node.instance_var.accept(self)
       @str << ')'
+      false
     end
 
     def visit(node : IsA)


Sija:
Some specs would nice :)

asterite:
I say it because visitor should be respected. This is probably some issue in ToSVisitor, check that you implement that and that you return false in the method so children aren't revisited/printed.

asterite:
I can confirm that's the issue. Just add a new line at the end of the method overload for OffsetOf in ToSVisitor that says false (but please keep the test).

asterite:
Or even better, remove this test and add a test to the to s visitor.

malte-v:
I just realized my "solution" breaks much more that it fixes, e.g. type aliases stopped working with `offsetof`.
@asterite Thank you so much! I wouldn't have noticed that small detail. That must be the proper solution.

PR #7711, opened 2019-04-24
A setter type is the argument's type

### LINKED ISSUE #7707: Inconsistent setter method behavior with Ruby
When using a setter method in Ruby, it always returns its assigned value, regardless of the value returned by the method. Maybe this is intentional, but is inconsistent with Ruby and can cause unexpected problems (which lead me to report this).

For instance:

```crystal
class Foo
  @value = 0
  
  def value=(new_value)
    @value = new_value
    "Not a number"
  end
end

foo1 = Foo.new
puts foo1.value = 42 # Works

foo2 = Foo.new
foo2.value = foo1.value = 42 # Breaks, trying to assign string to Int32
```

https://play.crystal-lang.org/#/r/6ryq

The last line is something logical users may want to do. But because the return method of `value=` is used instead of `new_value`, it causes an unexpected compiler error. It tries to assign the string "Not a number" to `foo2.value` instead of 42.

With the exact code in Ruby:

```ruby
class Foo
  @value = 0
  
  def value=(new_value)
    @value = new_value
    "Not a number"
  end
end

foo1 = Foo.new
puts foo1.value = 42 # Works

foo2 = Foo.new
foo2.value = foo1.value = 42 # Works (42) assigned to both
```

https://repl.it/repls/FixedJuniorFinance

I ran into this when I wanted to return an assigned value, but the setter method does some extra computation after setting the value, and that computation result was returned unexpectedly.

Crystal v0.28.0

asterite:
Fixes #7707

The way it's done is probably not very efficient, but it probably doesn't matter.


j8r:
What about `#[]=`?

asterite:
Oh, some obscure codegen bug. Maybe later...

src/compiler/crystal/semantic/call.cr
```diff
@@ -77,8 +77,13 @@ class Crystal::Call
 
     @target_defs = matches
 
-    bind_to matches if matches
-    bind_to block.break if block
+    # For a setter like `foo.x = y` we'd like to use `y`'s type
+    if self.setter?
+      bind_to args.first
```
bew: Note for later: it should probably be `bind_to args.last` instead to handle `[]=`

straight-shoota:
I've updated this PR and fixed the original codegen failure (it seems to be caused by the `setter?` property conflicting with the method). Also added support for `[]=`.

One spec causes a codegen bug. I've deactivated it for now, to check other things. That spec was introduced by #9972. There's no obvious connection between both changes. The compiler change from #9972 itself doesn't seem to have an influence on any other code as far as I can see, it's just that single spec that breaks.
So it's unclear to me whether the observed behaviour is an actual issue with this change or if it just surfaces an issue with the code from #9972 (or anything else).

But there's also an ICE not covered by specs.

asterite:
@straight-shoota Thanks! I'll take a look at this in about two weeks

asterite:
@straight-shoota I'm not sure what I'm supposed to do here. Which spec is failing with an ICE? I see CI just gives an ICE but I don't know where.

straight-shoota:
I don't know anything more, really.

The deactivated spec I talked about is https://github.com/asterite/crystal-1/blob/d864a13b1c6df229b520a9e98b1982aea646b1b9/spec/compiler/codegen/proc_spec.cr#L885

As for the ICE, I didn't dig any further since you said you would take a look at it.

asterite:
Oh, I thought there was a single spec that ICEd. Hmm... I might not have time to look at this these days, not sure.

asterite:
I think we won't go with this because it's considered a breaking change, so closing.

straight-shoota:
For the record: Breaking change doesn't mean this feature will never come. We can have a feature flag for that and make it the default in 2.0.

Issue #7391, opened 2019-02-07
Merge the /bin/crystal Shell script in the compiler?


j8r:
The [/bin/crystal](https://github.com/crystal-lang/crystal/blob/master/bin/crystal) script is setting environment variables, then calling the compiler.

The questions are:
- is it a limitation of the language to not be able to write this script directly in Crystal?
- can't this environment variables like our other ones in our `.bashrc` or `.zshrc`?
- How about Windows, would it be in Powershell?

Issues:
- As noted by jemc on Gitter/IRC, this complicates compiler debugging with tools like lldb.
- This also render counter-intuitive to have multiple crystal - one may think to copy the shell script, which isn't going to work.

I think exposing `CRYSTAL_PATH=$CRYSTAL_ROOT/src:lib`, and potentially other variables, through an API that can be used by `shards` and other projects (ameba?)  would be a nice plus, instead of assuming the library path in each project and hard-coding it.

asterite:
This isn't a problem.

j8r:
This is a problem for the Windows port, or we create `bin/crystal.bat`. Whatever the solution chosen, something has to be done.

j8r:
Prevents also to set `i_know_what_im_doing` when we think we know what we are doing like https://github.com/crystal-lang/crystal/issues/5835, but in fact not so much :sweat_smile:

straight-shoota:
I think we should keep `bin/crystal` as a developer tool. But I don't think the distribution packages should need it.

asterite:
@straight-shoota Then who sets `CRYSTAL_PATH` to point it to the standard library?

j8r:
For example look on Alpine, `file $(which crystal)` returns directly the binary.
`docker run -it --rm -v $PWD:/app -w /app jrei/crystal-alpine sh -c 'ldd $(which crystal)' `
```
        /lib/ld-musl-x86_64.so.1 (0x7f1d127c1000)
        libLLVM-5.0.so => /usr/lib/libLLVM-5.0.so (0x7f1d0f7d2000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x7f1d0f67d000)
        libpcre.so.1 => /usr/lib/libpcre.so.1 (0x7f1d0f620000)
        libgc.so.1 => /usr/lib/libgc.so.1 (0x7f1d0f5b1000)
        libevent-2.1.so.6 => /usr/lib/libevent-2.1.so.6 (0x7f1d0f566000)
[...]

PR #7715, opened 2019-04-24
Remove shell script wrapper

j8r:
Resolves https://github.com/crystal-lang/crystal/issues/7391
To sumup the points:
- simplifies debugging
- removes any confusion about which crystal binary is used
- helps for Windows support (or one prefer writing a bat file? :smile: )

For now this is a quick dirty draft to ping core members before continuing.

Please, submit any opinion in the related issue: https://github.com/crystal-lang/crystal/issues/7391


asterite:
Please explain how I can run a specific spec file without this wrapper.

Right now I do:

```
bin/crystal spec spec/std/array_spec.cr
```

the `bin/crystal` wrapper sets the value of `CRYSTAL_PATH` to use the standard library in my local copy rather than the one that comes with the crystal installation.

How do I do that without the wrapper?

j8r:
Environment variable, a bit like `GOPATH`?
The Makefile is already capable of doing this.

j8r:
By the way this is only needed when doing compiler development, not for common users.

`export CRYSTAL_PATH=$PWD/src` once doesn't seem too bad, versus having to learn why using `./bin/crystal` and not `.build/crystal`.

But either the script is kept or not, `which go`, `which rustc`, or whatever language directly returns a binary – it's also possible to do it for Crystal for the end user.

ysbaddaden:
This patch makes it harder to use a custom built crystal (binary + its stdlib), since CRYSTAL_PATH must be exported manually (regression), with different compiler builds it becomes a nightmare (huge regression) or... requires to write wrapper scripts manually :confused:

We can't run some compiler/std specs (not all of them) quickly without analyzing Makefile beforehand (regression) or manually exporting CRYSTAL_PATH again.

We can talk about not requiring a wrapper script, having a compile time configurable default CRYSTAL_PATH when the envvar is unset —thought paths will be relative, which is tricky to resolve flawlessly across platforms (hence why we leverage bash).

> export CRYSTAL_PATH=$PWD/src once doesn't seem too bad, versus having to learn why using ./bin/crystal and not .build/crystal.

Please. Calling bin/crystal is easy; .build/crystal is hidden to plain sight, and it even comes with a warning when run directly; manually exporting CRYSTAL_PATH all the time is tedious, error prone, and unfriendly and unwelcomed behavior.

j8r:
Let me try something, I've to look deeper how other languages do.
It's strange that only Crystal require this.

asterite:
It's not that it's required. It's a compiler developer tool. Remove it and we have to do some things manually (like always passing the ENV var, or manually choosing a compiler compiler).

I'll close this and the related issue.

j8r:
@asterite if it's a developer compiler tool, don't ship it to the user.
I think it's possible to directly use `.build/crystal` without having a wrapper or setting env variables.

PR #7716, opened 2019-04-24
Add URI#resolve and URI#relativize

straight-shoota:
This PR adds two complementary methods to `URI`:

* `URI#resolve(uri)` resolves `uri` against `self` according to [RFC 3986 section 5.2](https://tools.ietf.org/html/rfc3986#section-5.2.2).
* `URI#relativize(uri)` relativizes `uri` against `self` using `query` and `fragment` components from `uri` and a path component that describes a minimum-difference relative path between the paths of `self` and `uri`.

```cr
URI.parse("http://foo.com/bar/baz").resolve("../quux")                # => http://foo.com/quux
URI.parse("http://foo.com/bar/baz").relativize("http://foo.com/quux") # => ../quux`
```

#7697 contains a similar implementation for `URI#resolve` called `URI#merge`

### DIFF
--- a/src/uri.cr
+++ b/src/uri.cr
@@ -21,6 +21,28 @@ require "./uri/encoding"
 # uri.to_s   # => "http://foo.com/posts?id=30&limit=5#time=1305298413"
 # ```
 #
+# ## Resolution and Relativization
+#
+# *Resolution* is the process of resolving one URI against another, *base* URI.
+# The resulting URI is constructed from components of both URIs in the manner specified by
+# [RFC 3986 section 5.2](https://tools.ietf.org/html/rfc3986#section-5.2.2), taking components
+# from the base URI for those not specified in the original.
+# For hierarchical URIs, the path of the original is resolved against the path of the base
+# and then normalized. See `#resolve` for examples.
+#
+# *Relativization* is the inverse of resolution as that it procudes an URI that
+# resolves to the original when resolved against the base.
+#
+# For normalized URIs, the following is true:
+#
+# ```
+# a.relativize(a.resolve(b)) # => b
+# a.resolve(a.relativize(b)) # => b
+# ```
+#
+# This operation is often useful when constructing a document containing URIs that must
+# be made relative to the base URI of the document wherever possible.
+#
 # # URL Encoding
 #
 # This class provides a number of methods for encoding and decoding strings using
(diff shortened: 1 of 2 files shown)


ysbaddaden:
I prefer merge. It's more explicit than resolve which is confusing —what is is supposed to resolve?

It's also explicitly named "merge" and "remove dot segments" in RFC 3986 sections 5.2.3 and 5.2.4. The term "resolve" is never used, even in section 5.2.2.

I had to dig the specs to start to understand what relativise could mean. I don't see any use case. I don't think it's frequent enough to be in stdlib.

straight-shoota:
The entire section 5 of  RFC 3986 is titled *URI resolution* and that's the common term for this, also used in many APIs. The RFC only refers to merging of paths.
Calling the entire resolution process `merge` seems to be a Rubyism. I don't think this is a good name, because it's pretty unspecific. Using the more expressive name `resolve` is better.

> I had to dig the specs to start to understand what relativise could mean.

The docs explicitly state it's the inverse of `#resolve`.
`#relativize` is definitely not as commonly used as `#resolve`, so yeah, it might not fit into stdlib. But it's still useful and complements `#resolve`. The implementation is already not very complex and  I suppose the implementation of `#relativize_path` can eventually move to `Path` because having a similar method on `Path` would be nice.

oprypin:
@ysbaddaden is the only one suggesting the name `merge` (and really insisting on it, for whatever reason).

But I think it just doesn't fit. Programmers tend to think of merging as taking the "best" parts of two sequences, possibly by indiscriminately iterating over them.

And that's very misleading in this context because it's explicitly *not* this type of operation.

`resolve` makes a lot of sense to me, and @straight-shoota provides good justification as well.

oprypin:
But I think there's a mistake in the original post. The last example seems to be demonstrating `relativize` but actually calls `resolve`.

straight-shoota:
Fixed

bcardiff:
I think there is still an issue in the opening comment:

It should probably be:

```crystal
URI.parse("http://foo.com/bar/baz").relativize("http://foo.com/quux") # => ../quux`
```

src/uri.cr
```diff
@@ -19,6 +19,28 @@ require "./uri/uri_parser"
 # uri.query  # => "id=30&limit=5"
 # uri.to_s   # => "http://foo.com/posts?id=30&limit=5#time=1305298413"
 # ```
+#
+# ## Resolution and Relativization
+#
+# *Resulution* is the process of resolving one URI against another, *base* URI.
```
r00ster91:
```suggestion
# *Resolution* is the process of resolving one URI against another, *base* URI.
```

straight-shoota:
Rebased on master

straight-shoota:
Rebased onto master.

src/uri.cr
RX14: This change was lost

straight-shoota: Thanks

Issue #7709, opened 2019-04-24
Titlecase vs Camelcase


wontruefree:
This api [String#camelcase](https://crystal-lang.org/api/0.28.0/String.html#camelcase%28options%3AUnicode%3A%3ACaseOptions%3DUnicode%3A%3ACaseOptions%3A%3ANone%29-instance-method) seems to be what I have heard called titlecase.  Should there be two apis `camelcase` and `titlecase`?

```crystal
"hello_world".camelcase #=> "helloWorld"
"hello_world".titlecase #=> "HelloWorld"
```

Blacksmoke16:
Yes they are separate things.  `camelcase` does not upcase the first character while `titlecase` does.

oprypin:
camelcase does indeed have two conflicting definitions...
or perhaps the only required property is that it has uppercase letters in the middle (regardless of the first letter).

This is why I say that your suggested change provides no improvement over the current status.
And, since it's backwards-incompatible (in potentially dangerous ways), I'd just like to shoot this down entirely...

I'd support mixedcase + titlecase though (get rid of the ambiguous camelcase)

vladfaust:
Wiki [says](https://en.wikipedia.org/wiki/Camel_case) that **camelCase** has single definition.

asterite:
There's camelCase and PascalCase

konovod:
@vladfaust wiki says:
> Some programming styles prefer camel case with the first letter capitalised, others not.[1][2][3] For clarity, this article calls the two alternatives upper camel case (initial uppercase letter, also known as Pascal case) and lower camel case (initial lowercase letter, also known as Dromedary case[4]). Some people and organizations, notably Microsoft,[2] use the term camel case only for lower camel case.

I like the PascalCase idea though (because i like pascal).

ysbaddaden:
It's much more confusing than that. Citing the [wikipedia article](https://en.wikipedia.org/wiki/Camel_case):

- Title Case (...) capitalises all words but retains the spaces between them;
- upper camel case (initial uppercase letter, also known as Pascal case);
- lower camel case (initial lowercase letter, also known as Dromedary case[4]) ;
- other synonyms include: <too many to list, check the wikipedia article>

Title case is use in English literature and newspapers to capitalize titles, for example "very blunt title" will become "Very Blunt Title".

[...]

PR #7717, opened 2019-04-24
add lowercase option to String#camelcase

wontruefree:
resolves https://github.com/crystal-lang/crystal/issues/7709

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -3500,13 +3500,14 @@ class String
   # Converts underscores to camelcase boundaries.
   #
   # ```
-  # "eiffel_tower".camelcase                                   # => "EiffelTower"
-  # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
+  # "eiffel_tower".camelcase                                            # => "EiffelTower"
+  # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
+  # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, lowercase_first_letter = false)
+  def camelcase(downcase = false, options : Unicode::CaseOptions = Unicode::CaseOptions::None)
     return self if empty?
 
-    first = !lowercase_first_letter
+    first = !downcase
     last_is_underscore = false
 
     String.build(bytesize) do |str|
--- a/src/string.cr
+++ b/src/string.cr
@@ -3504,7 +3504,7 @@ class String
   # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
   # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase = false)
+  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase : Bool = false)
     return self if empty?
 
     first = !downcase
--- a/src/string.cr
+++ b/src/string.cr
@@ -3504,7 +3504,7 @@ class String
   # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
   # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(downcase = false, options : Unicode::CaseOptions = Unicode::CaseOptions::None)
+  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase = false)
     return self if empty?
 
     first = !downcase
--- a/src/string.cr
+++ b/src/string.cr
@@ -3503,10 +3503,10 @@ class String
   # "eiffel_tower".camelcase                                   # => "EiffelTower"
   # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None)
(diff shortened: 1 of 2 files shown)


ysbaddaden:
Just `lower` would be enough. Both variants (lower camel case, upper camel case) should be documented, too.

src/string.cr
```diff
@@ -3503,10 +3503,10 @@ class String
   # "eiffel_tower".camelcase                                   # => "EiffelTower"
   # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None)
+  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, lowercase_first_letter = false)
```
konovod:
```suggestion
  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, lowercase_first_letter = false)
```
to prevent just `false` as argument, suggested in https://github.com/crystal-lang/crystal/issues/7709#issuecomment-486559644

vladfaust: I'd stand for just `lower` as the argument name. It's clearly understandable that `lower` in context of `camelcase` means exactly `lower camelcase`, which is a common term. No need for a long and explicit name here.

straight-shoota:
`lowercase_first_letter` is too long.

I'd suggest one of the following (in order of preference):
* `downcase_first`
* `lower`
* `lower_first`

wontruefree:
@straight-shoota @ysbaddaden what do you think about `downcase` as the argument?

src/string.cr
```diff
@@ -3500,13 +3500,14 @@ class String
...
   #
   # ```
-  # "eiffel_tower".camelcase                                   # => "EiffelTower"
-  # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
+  # "eiffel_tower".camelcase                                            # => "EiffelTower"
+  # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
+  # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None)
+  def camelcase(downcase = false, options : Unicode::CaseOptions = Unicode::CaseOptions::None)
```
r00ster91: Maybe this argument should be at the end to avoid a breaking change?

wontruefree: I moved it to the beginning because I feel like `options` parameters are usually at the end of a signature.

konovod: `options` is self-descriptive because uses Enum. `downcase` should either use Enum, or be a named argument (appear after `*,`), otherwise people will write `str.camelcase(true)`.

wontruefree: @konovod can you take another look at this?

konovod: Looks 👍 now.

src/string.cr
```diff
@@ -3500,13 +3500,14 @@ class String
...
   #
   # ```
-  # "eiffel_tower".camelcase                                   # => "EiffelTower"
-  # "isolated_integer".camelcase(Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
+  # "eiffel_tower".camelcase                                            # => "EiffelTower"
+  # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
+  # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None)
+  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase = false)
```
j8r:
```suggestion
  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase : Bool = false)
```
I would also add that `downcase` is the first character in downcase variant of camelcase

vladfaust:
That’s bad. ‘camelcase(downcase: true)’ is confusing. ‘camelcase(lower: true)’ is better. And I am for breaking change if needed to move the argument to the first place.

j8r:
Or `lower_first`, that's about a bit longer than the current `downcase` and remove any ambiguity

straight-shoota:
> And I am for breaking change if needed to move the argument to the first place.

No. It should only be usable as a named argument to clearly express the semantics.

wontruefree:
@crystal-lang/team CI seems to be failing due to running too long.  Is this common for CI and how can I rerun it?

wontruefree:
@straight-shoota thank you for re-running this.
Can we merge this?

vladfaust:
That's disturbing.

wontruefree:
@vladfaust Which part is disturbing?

ysbaddaden:
I'm late to the party, but where the hell did downcase came from? Downcase camelcase doesn't mean anything, it's just confusing.

This is lower or upper camelcase (which affects the letter) not downcase —a completely different method and meaning :confused:

straight-shoota:
@ysbaddaden It downcases the first character, the equivalent method is `#downcase` not `#lower`, so it makes some sense. As mentioned, I'd have prefered `downcase_first` or `lower_first`.

straight-shoota:
Hm, actually, it doesn't even downcase the first character when it's uppercase: `"Foo".camelcase(downcase: true) => "Foo"`.  I think this behaviour should be changed.

asterite:
I think the API is fine, the above is just a bug in this PR's implementation.

r00ster91:
Isn't this method generally behaving incorrectly? This method returns `"EiffelTower"` for `"eiffel_tower"` but that's wrong. `"EiffelTower"` is PascalCase, not camelCase. The first letter should be lowercase by default. I think there should be another method, `pascalcase` that has the current behavior of `camelcase`.

j8r:
@r00ster91 Pascal case was discussed in [the related issue](https://github.com/crystal-lang/crystal/issues/7709), please do not duplicate the discussion in this PR.

PR #7719, opened 2019-04-25
Optimize Enumerable#join

j8r:
This is an insignificant PR; I noticed this when working on a lexer/builder.

```cr
require "benchmark"

module Enumerable
  def new_join(separator, io)
    start = false
    each do |elem|
      start ? io << separator : (start = true)
      yield elem, io
    end
  end
end

SMALL = StaticArray(Int32, 4).new 123
BIG = StaticArray(Int32, 999).new 123

Benchmark.ips do |x|
  x.report("small old join") { SMALL.join("", STDOUT) {|e|} }
  x.report("small new join") { SMALL.new_join("", STDOUT) {|e|} }

  x.report("big old join") { BIG.join("", STDOUT) {|e|} }
  x.report("big new join") { BIG.new_join("", STDOUT) {|e|} }
end
```
Result:
```
small old join  73.28M ( 13.65ns) (± 0.81%)  0 B/op         fastest
small new join  72.57M ( 13.78ns) (± 2.34%)  0 B/op    1.01× slower
  big old join 226.14k (  4.42µs) (± 1.16%)  0 B/op  324.02× slower
  big new join 231.49k (  4.32µs) (± 1.36%)  0 B/op  316.54× slower
```
 
Edit My bad I mess up the condition, it wasn't doing anything. Fixed and result updated.
Using booleans are easier to understand.

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -693,9 +693,9 @@ module Enumerable(T)
   # (1), (2), (3), (4), (5)
   # ```
   def join(separator, io : IO)
-    start = 0
+    start = false
     each do |elem|
-      start == 1 ? io << separator : (start = 1)
+      start ? io << separator : (start = true)
       yield elem, io
     end
   end
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -695,8 +695,9 @@ module Enumerable(T)
   def join(separator, io : IO)
     start = false
     each do |elem|
-      start ? io << separator : (start = true)
+      io << separator unless first
       yield elem, io
+      first = false
     end
   end
 
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -693,7 +693,7 @@ module Enumerable(T)
   # (1), (2), (3), (4), (5)
   # ```
   def join(separator, io : IO)
-    start = false
+    first = true
     each do |elem|
       io << separator unless first
       yield elem, io
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -695,7 +695,7 @@ module Enumerable(T)
   def join(separator, io : IO)
     start = 0
     each do |elem|
-      start == 1 ? io << separator : (start = 0)
+      start == 1 ? io << separator : (start = 1)
       yield elem, io
     end
   end
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -692,9 +692,10 @@ module Enumerable(T)
   # ```text
   # (1), (2), (3), (4), (5)
   # ```
-  def join(separator, io)
-    each_with_index do |elem, i|
-      io << separator if i > 0
+  def join(separator, io : IO)
+    start = 0
+    each do |elem|
+      start == 1 ? io << separator : (start = 0)
       yield elem, io
     end
   end


asterite:
Could you write it like this?

```crystal
  def join(separator, io : IO)
    first = true
    each_with_index do |elem, i|
      io << separator unless first
      yield elem, io
      first = false
    end
  end
```

Doing an assignment inside a ternary operator is very hard to read and understand.

asterite:
There's no magic here: the existing code increments a counter on each loop. The new code just sets a variable to false. I guess that's slightly faster.

j8r:
Ok, but the `each_wth_index` is no longer needed @asterite

src/enumerable.cr
```diff
@@ -692,10 +692,12 @@ module Enumerable(T)
   # ```text
   # (1), (2), (3), (4), (5)
   # ```
-  def join(separator, io)
-    each_with_index do |elem, i|
-      io << separator if i > 0
+  def join(separator, io : IO)
+    start = false
```
Sija:
```suggestion
    first = false
```

konovod:
So there is no performance bonus in benchmark with latest changes? What's a point of change then?

j8r:
@konovod there is a slight difference when using bigger enumerable:
```
big old join 202.45k (  4.94µs) (± 3.37%)  0.0B/op   1.01× slower
big new join 204.20k (  4.90µs) (± 2.88%)  0.0B/op        fastest
```
I reproduced it several time to be sure.
I admit this is a bit useless, I was initially mainly wondering if there is an advantage of incrementing a counter compared to a simple boolean.

straight-shoota:
Did you check if moving `first = false` into the `unless first` branch provides any notable benefit?

```cr
first = true
each do |elem|
  unless first
    io << separator
    first = false
  end
  yield elem, io
end
```

j8r:
@straight-shoota the code inside the unless will never run.
I've checked
```cr
first = true
each do |elem|
  if first
    first = false
  else
    io << separator
  end
  yield elem, io
end
```
No difference compared to doing `first = false` on each iteration (@asterite code). Maybe it doesn't necessarily means an operation if `first` is already `false`, or it is as expensive as a `if/else` conditional.

wontruefree:
@j8r I did a few benchmarks and found that code you mentioned had some benefits with big arrays (~ 3.5%).

But I found that checking the variable before writing back to it has about a ~10% benefit with big arrays (9999999 elements) and about a ~10% loss with small arrays (99 elements).

```crystal
  def new_join(separator, io)
    first = true
    each do |elem|
      io << separator unless first
      yield elem, io
      first = false if first
    end
  end
```

Don't know if it is worth it because I imagine most joins will be with small arrays.  Also, I am not entirely sure why this happens but I was thinking setting the variable would break the CPU cache and that would cause a performance hit.

wontruefree:
@konovod I would agree with you which is why I am stating I don't know why this is happening.  From your benches, I am not sure which strategy is best.  It seems like they are all the fastest at some point.

jhass:
The benchmark results are essentially random and within error margin towards each other. The optimizer is smart enough here and this is bikeshedding. I'm in favor of closing this as there's no clear evidence of improvement.

PR #5627, opened 2018-01-22
Adds System Users and Groups

chris-huxtable:
Attempts to support Darwin, all Linux's, OpenBSD, and FreeBSD

Also adds features like:
- `Process.owner = System::User.get("nobody")`
- `Process.group = System::Group.get("nobody")`
- `Process.user`
- `Process.group`
- `System::User.get("root").shell`
- `System::Group.get("wheel").members`
- `File.chown("some/path", "root", "wheel")`
- `File.chown("some/path", System::User.get("root"), System::Group.get("wheel"))`
- `a_pointer.to_slice_null_terminated(a_limit)` (useful for converting the Char** to an Array(String))

Previous PR #5615

PR #5615, opened 2018-01-19
Adds System Users and Groups

chris-huxtable:
Attempts to support Darwin, all Linux's, OpenBSD, and FreeBSD

Also adds features like:
- `Process.become(User.get("nobody"))`
- `Process.become(Group.get("nobody"))`
- `Process.user`
- `Process.group`
- `User.get("root").shell`
- `Group.get("wheel").members`
- `a_pointer.to_slice_null_terminated(a_limit)` (useful for converting the Char** to an Array(String))

PR #7725, opened 2019-04-29
Introduce System::User and System::Group

woodruffw:
Introduces the `System::User` and `System::Group` classes, which in turn delegate to the `Crystal::System::User` and `Crystal::System::Group` classes.

This needs documentation and unit tests, but I'll hold off on the latter until a few maintainers :+1:  this general structure/approach.

Example usage:

```crystal
require "system/group"
require "system/user"

group = System::Group.from_id(1000_u32)
puts group.name, group.members

user = System::User.from_username(group.members.first)
puts user.directory, user.shell
```

See #5627, #5615


src/system/user.cr
```diff
@@ -0,0 +1,62 @@
...
+    sys_user = Crystal::System::User.from_name?(name)
+    new(sys_user) if sys_user
+  end
+
+  # Returns the user associated with the given name.
+  #
+  # Raises `NotFound` if no such user exists.
+  # See `from_name?`.
+  def self.from_name(name)
+    self.from_name?(name) || raise NotFound.new("no such user: #{name}")
```
Sija:
```suggestion
    from_name?(name) || raise NotFound.new("No such user: #{name}")
```

- Exception messages should start with uppercase
- using `self` is redundant

ditto elsewhere

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,53 @@
...
+    buf = Bytes.new(1024)
+
+    ret = LibC.getpwnam_r(username, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    while ret == LibC::ERANGE
+      buf = Bytes.new(buf.size * 2)
+      ret = LibC.getpwnam_r(username, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    end
+
+    raise Errno.new("getpwnam_r") if ret != 0
+    return nil if pwd_pointer.null?
```
Sija:
```suggestion
    return if pwd_pointer.null?
```

ditto elsewhere

straight-shoota:
Just to be sure: This essentially extracts a part of #5627 and only adds the `System::User` and `System::Group` types?

Looks good. Specs and doc would be awesome :+1:

src/docs_main.cr
```diff
@@ -62,3 +62,4 @@ require "./uuid"
 require "./uuid/json"
 require "./zip"
 require "./zlib"
+require "./system/*"
```
Blacksmoke16: Is there something we're wanting to add to the public API docs? /cc @straight-shoota

woodruffw:
> Just to be sure: This essentially extracts a part of #5627 and only adds the `System::User` and `System::Group` types?

Correct! My thought here was to keep the initial changeset small, and work to expand the API/interaction with other classes (`File::Info` and `Process` come to mind) in subsequent PRs.



> Specs and doc would be awesome 👍

Will do! I'll put some initial specs in this evening (EST).

woodruffw:
Alright, added initial specs.

Some notes:

* Unless we augment the testing environment to add special users and groups, there's no good testing story for the `Group#password`, `Group#members`, `User#password`, `User#group_id`, `User#directory`, or `User#shell`. The `User` attributes could be tested by adding something like `System::User.each_user` that wraps `getpwent_r`, but that's a nonstandard call. Not a huge deal, just something to consider.

woodruffw:
Failure on `linux32` looks unrelated:

```
 1) TCPServer .new using IPv4 reuse_port raises when not binding with reuse_port

     Failure/Error: expect_raises_errno(Errno::EADDRINUSE, {% if flag?(:gnu) %}"listen: "{% else %}"bind: "{% end %}) do



       Expected Errno with "listen: ", got #<Errno: bind: Address already in use> with backtrace:
```

straight-shoota:
Splitting #5627 into separate pieces is a good idea :+1:

Did you take a look at how other implementations test this? I don't think setting up a specific environment for these specs is feasible. So we need to accept that some features can't be tested effectively. But they're all in all pretty simple wrappers to OS features, so this shouldn't be a big deal. Make sure to add `typeof` calls though to make sure the code is at least valid even if it's not run.

But you could take some inspiration from Golang's test suite for example (e.g. https://github.com/golang/go/blob/master/src/os/user/user_test.go) on how others implement tests for this.

woodruffw:
> But you could take some inspiration from Golang's test suite for example (e.g. https://github.com/golang/go/blob/master/src/os/user/user_test.go) on how others implement tests for this.

Thanks! Looks like Golang uses the running process's `uid` and `gid` (presumably via `getuid`/`getgid`) to create objects with testable state. We could duplicate this approach by adding `Process.user` and `Process.group`, and then yanking the relevant names/IDs out of those objects. But perhaps that's better for a future PR, to avoid ballooning this one any more.

Either way, I'll add type checks now 👍

straight-shoota:
> But perhaps that's better for a future PR, to avoid ballooning this one any more.

Sounds like a good idea.

woodruffw:
> But in the specs you could simply use `` `id -un` `` and `` `id -gn` `` for now.

Ah, totally forgot I can shell out in tests. Thanks! I'll use those.

j8r:
You can use `struct` instead of `class`.
Why having private getters for `sys_group` and `sys_user`?

woodruffw:
> You can use `struct` instead of `class`

:+1: 

> Why having private getters for `sys_group` and `sys_user`?

I might be mistaken about this, but my understanding is that the `Crystal::System` hierarchy should never be exposed directly to users -- it's purely for filling out the actual user-facing classes. That's why I made the two private, but I'm 👍 on exposing them publicly if that's preferred.

j8r:
@woodruffw the questiom wasn't really why tey are private, this was more about why `private getter` instead of a regular ivar `@sys_user`. Only `private def initialize(@sys_user  : Crystal::System::User` should be enough (same for `sys_group`)

j8r:
@woodruffw the questiom wasn't really why they are private, this was more about why `private getter` instead of a regular ivar `@sys_user`. Only `private def initialize(@sys_user  : Crystal::System::User` should be enough (same for `sys_group`)

woodruffw:
@j8r Ah, gotcha. Yes, a regular ivar would be just fine. Will change.

woodruffw:
Hmm, looks like `id -un` chokes on (some) of the linux CIs:

```
id: cannot find name for user ID 1001

Error in spec/std_spec.cr:2: while requiring "./std/**"
```

Will investigate further tomorrow.

woodruffw:
Guessing it has something to do with the `-u 1001` in this `docker` invocation:

```
docker run --rm -t -u 1001 -v /home/circleci/project:/mnt -w /mnt -e CRYSTAL_CACHE_DIR=/tmp/crystal crystallang/crystal:0.28.0-build linux64 /bin/sh -c 'make std_spec clean threads=1'
```

straight-shoota:
This problem has been hit in the previous PR, see https://github.com/crystal-lang/crystal/pull/5627#issuecomment-380619156 and  https://github.com/crystal-lang/crystal/pull/5627/files#diff-a8da778d56fdb695e50daf67e50c0fbc

woodruffw:
Gotcha, thanks for that! I'll apply the same fix.

src/crystal/system/unix/group.cr
```diff
@@ -0,0 +1,63 @@
+require "c/grp"
+
+class Crystal::System::Group
```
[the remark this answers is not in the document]
woodruffw: I might be misunderstanding, but I don't see how `File` reopens the original type -- the base `File` class uses `include` on `Crystal::System::File`, which in turn references some constants from `::File` but doesn't reopen it. 

(I agree with the overarching point about unnecessary duplication, I just want to make sure I understand your recommendation :smile:)

spec/std/system/group_spec.cr
```diff
@@ -0,0 +1,93 @@
...
+    it "raises on nonexistent group" do
+      expect_raises System::Group::NotFound, "No such group" do
+        System::Group.from_id(1234567_u32)
+      end
+    end
+  end
+
+  describe "name" do
+    it "is a String" do
+      System::Group.from_name(GROUP_NAME).name.should be_a(String)
```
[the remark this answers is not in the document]
woodruffw: Yeah, agree. I'll remove the pointless type checks and add some (unvalidated) calls to the other members of `User` and `Group`.

woodruffw:
> However, there are some derivations from the original implementation that I don't think are an improvement (like `password` property), and some neat features missing (such as `#to_s` and `def_equals_and_hash` on the new types).

Went ahead and removed the `password` property from both `User` and `Group`. I'll add `to_s` in a bit.

Re: `def_equals_and_hash`: I think just the user/group ID should be sufficient, at least for POSIX. Thoughts?

I'll also follow up on the general design process in #7738.

src/crystal/system/unix/group.cr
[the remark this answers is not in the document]
straight-shoota: Oh sry, `File` was an example for including a `Crystal::System` type. That might be a good approach here, too. But not two separate types with delegation.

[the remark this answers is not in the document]
woodruffw: Alright, refactored so that `User` and `Group` each `extend` their `Crystal::System` counterpart (`include` doesn't work because we're mixing class methods in).

One slight problem: `from_name?` and `from_id?` don't appear on their extending classes in the documentation. The docs correctly show that e.g. `Crystal::System::User` is being extended, but don't provide documentation for it (correctly, IMO).

Is there a workaround to this? I suppose I could just move the `?` methods out of `Crystal::System` and into `System` and have them call helper methods, but that feels a little clunky.

woodruffw:
Anything else this PR needs? We should definitely continue the discussion in #7738, but I think (unless I'm mistaken) that there's nothing objectionable about the current implementation/API.

straight-shoota:
Looks good so far, only the data types of group and user ids should be `String` instead of platform-specific types. Also, from #7738 I'd like to have `User#username` return the username and `User#name` the real/display name.

woodruffw:
Bleh, this got lost in the stack. I'll catch up on this and make the suggested changes latter this week/weekend.

woodruffw:
> Looks good so far, only the data types of group and user ids should be `String` instead of platform-specific types. Also, from #7738 I'd like to have `User#username` return the username and `User#name` the real/display name.

Instead of making `uid` and `gid` into strings, might it make sense to only expose them as part of the API for `:unix` targets? That way we get the best of both worlds: Linux/macOS/BSD users can treat UIDs and GIDs as the integers they are, and any Crystal code compiled for Windows that uses them will fail at compile time. We could then add the appropriate (feature-gated) methods for Windows as well, overlapping names where appropriate.

I got this idea from looking at the `File::Info` source for Win32: instead of just failing at compile time when someone tries to use `File::Info#owner` or `#group` on a system where those don't make sense, we currently return `0_u32`. As a result, a user attempting to re-build their Crystal project on Windows will (1) miss the semantic bug of using a method on a platform that it doesn't make sense for, and (2) get a very confusing result (`0` is usually root's UID).

In any case, I'll change `#name` to `#username` and extract the full name from the GECOS field.

straight-shoota:
The standard library is supposed to be portable. When a program runs on Linux, it should just run on MacOS, BSD, Windows by default. A developer should not have to worry about platform-specifics.

Thus the standard library needs to provide an appropriate abstraction layer and platform-specific APIs can be provided by shards.

So the gid/uid should be encoded as strings. That's the typical platform-independent implementation. And I don't think there is any downside, except that POSIX developers are used to use integers. When writing platform-specific code interfacing with POSIX APIs, you'll need to convert the datatype. But that's hardly a big issue.

src/system/user.cr
```diff
@@ -7,23 +7,25 @@ class System::User
...
   getter group_id : LibC::GidT
+  getter name : String
   getter directory : String
   getter shell : String
 
   def_equals_and_hash @user_id
 
-  private def initialize(@name, @user_id, @group_id, @directory, @shell)
+  private def initialize(@username, @user_id, @group_id, gecos, @directory, @shell)
+    @name = gecos.split(",").first
```
straight-shoota: This extraction from gecos should be in the platform-specific code. `System::User.new` should directly accept a `@name` argument.

woodruffw:
> The standard library is supposed to be portable. When a program runs on Linux, it should just run on MacOS, BSD, Windows by default. A developer should not have to worry about platform-specifics.
> 
> Thus the standard library needs to provide an appropriate abstraction layer. Platform-specific APIs can be provided by shards.
> 
> So the gid/uid should be encoded as strings. That's the typical platform-independent implementation. And I don't think there is any downside, except that POSIX developers are used to use integers. When writing platform-specific code interfacing with POSIX APIs, you'll need to convert the datatype. But that's hardly a big issue.

All fair points, and I ultimately want to use these features so I won't argue too much here :wink: 

But as a general observation: I think users are sometimes best served by a standard library that *isn't* cross-platform, as least not when "cross-platform" is defined as "the same calls on all platforms." There are some things (user and group management here, but also fine-tuned socket and I/O control, extended permissions, etc.) that are really nice to have in a standard library, but don't fit cleanly into abstract, platform-agnostic hierarchies.

I think Crystal is uniquely positioned to *ameliorate* a lot of the conventional hassle of writing platform-specific code in a platform-agnostic codebase: DCE, the type system, hygenic macros, feature flags, etc. make it possible to do platform-specific things cleanly, and I think users would appreciate a standard library that stops the compilation on a platform where (e.g.) `File::Info#owner` is meaningless rather than implicitly doing the wrong thing.

In any case, I'll change `uid` and `gid` to Strings in a moment.

didactic-drunk:
Why not do what go does?
```
type User struct {
        // Uid is the user ID.
        // On POSIX systems, this is a decimal number representing the uid.
        // On Windows, this is a security identifier (SID) in a string format.
        // On Plan 9, this is the contents of /dev/user.
}
```

`Crystal::System::UidT` can specify the type.

RX14:
As far as I know we decided a long time ago, democratically, by consistently merging code which followed this convention, that compiling on one platform meant compiling on all platforms, and that platform specific behaviour would be runtime exceptions. So as far as I'm concerned that's policy and any bikeshedding about it should be redirected somewhere else.

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,45 @@
...
+    user = String.new(pwd.pw_gecos).split(",").first
+    new(String.new(pwd.pw_name), pwd.pw_uid.to_s, pwd.pw_gid.to_s, user, String.new(pwd.pw_dir), String.new(pwd.pw_shell))
+  end
+
+  def from_username?(username : String)
+    username.check_no_null_byte
+
+    pwd = uninitialized LibC::Passwd
+    pwd_pointer = pointerof(pwd)
+    buf = Bytes.new(1024)
```
ysbaddaden: Tiny optimization: we could use a StaticArray on the stack, instead of allocating HEAP memory, only allocating a dynamic Bytes if 1024 bytes aren't enough:

```crystal
buf = uninitialized UInt8[1024]
```

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,45 @@
...
+
+    ret = LibC.getpwuid_r(id, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    while ret == LibC::ERANGE
+      buf = Bytes.new(buf.size * 2)
+      ret = LibC.getpwuid_r(id, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    end
+
+    raise Errno.new("getpwuid_r") if ret != 0
+    return nil if pwd_pointer.null?
```
ysbaddaden: Pedantic: I notice you sometimes use `return nil if ptr.null?` or `return if ptr.null?`, please stick to either for consistency; thought *I* would use something more Ruby-esque:
```crystal
from_struct(pwd) if pwd_pointer
```

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,45 @@
...
+
+  def from_username?(username : String)
+    username.check_no_null_byte
+
+    pwd = uninitialized LibC::Passwd
+    pwd_pointer = pointerof(pwd)
+    buf = Bytes.new(1024)
+
+    ret = LibC.getpwnam_r(username, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    while ret == LibC::ERANGE
```
ysbaddaden: It's allocating 2× more memory in a loop. I assume the buffer should eventually be large enough, but I'm not fond of such an assumption, and would add an upper limit (e.g. 16KB) and raise Errno if it still fails.

src/crystal/system/unix/user.cr
woodruffw: Nice! Should have read the docs; didn't realize a null `Pointer` is falsey :smile:

src/crystal/system/unix/user.cr
woodruffw: Good call, I'll add a test to the loop. The `raise` that follows the loop should suffice, since `ret` should be `ERANGE`.

src/crystal/system/unix/user.cr
woodruffw: One slight annoyance: I had to go with 

```crystal
initial_buf = uninitialized UInt8[1024]
buf = initial_buf.to_slice
```

since the compiler likes neither the union type (maybe `StaticArray` could be special-cased like `Slice` for C FFI?) nor either of these syntaxes:

```crystal
buf = (uninitialized UInt8[1024]).to_slice
```

```crystal
buf = uninitialized UInt8[1024].to_slice
```

woodruffw:
> I really like the Pull Request, it's simple and just abstract enough, but I'm not fond of the `from_` prefix in constructors. What about using kwargs? and maybe an activerecord like `find`? So we'd have `Process::User.find(username: "julien")` and `Process::User.find(id: "101")`?

I don't have any strong opinions here (I use the `from_` pattern in some of my Ruby code, but it's less useful in a strongly-typed language with multiple dispatch). Haven't made the changes yet, but if other maintainers :+1: the `find`-style API then I'll update the PR :smile:

woodruffw:
Let me know if there's anything else I can do here. AFAIK the only thing this is waiting on is more opinions about the current API design.

ysbaddaden:
Nothing really, apart from my small remark on naming the API.

@asterite @bcardiff @RX14 @straight-shoota are `User.from_name` and `User.from_id` fine, or shall we have something nicer, like `User.find(*, name: nil)` and `User.find(*, id: nil), or something more in harmony with the rest of the core/stdlib —thought I don't remember having an API to seek data?

Also is searching an user or group by id a POSIX only thing? How are user/group represented on Windows?

No strong opinion anyway, I'd just like to merge a nice API that we'll be happy to use.

src/crystal/system/unix/group.cr
```diff
@@ -0,0 +1,58 @@
...
+      ret = LibC.getgrnam_r(groupname, grp_pointer, buf, buf.size, pointerof(grp_pointer))
+    end
+
+    raise Errno.new("getgrnam_r") if ret != 0
+
+    from_struct(grp) if grp_pointer
+  end
+
+  def from_id?(groupid : String)
+    groupid = groupid.to_i.to_u32!
```
RX14: Why not just `groupid.to_u32`?

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,47 @@
...
+      ret = LibC.getpwnam_r(username, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    end
+
+    raise Errno.new("getpwnam_r") if ret != 0
+
+    from_struct(pwd) if pwd_pointer
+  end
+
+  def from_id?(id : String)
+    id = id.to_i.to_u32!
```
RX14: ditto

RX14:
Looks very good, just two minor nitpicks

RX14:
> Also is searching an user or group by id a POSIX only thing? How are user/group represented on Windows?

Go uses SID on windows for user and group IDs, and supports looking up by them.

I'm fine with `User.from_name`, although I think `User.find(name: "rx14")` may be neater. It's up to @woodruffw.

src/crystal/system/unix/group.cr
woodruffw: Thanks!

woodruffw:
> I'm fine with `User.from_name`, although I think `User.find(name: "rx14")` may be neater. It's up to @woodruffw.

Yep, I agree, will make these changes in a bit.

bcardiff:
@wontruefree I think it should be `User.find_by(name: ...)` as in ActiveRecord API

woodruffw:
Okay, updated to expose an ActiveRecord-style API:

```crystal
User.find_by name: "william"
Group.find_by? id: 0
```

Let me know if anything else needs be done :slightly_smiling_face:

src/crystal/system/unix/group.cr
```diff
@@ -0,0 +1,58 @@
...
+
+  private def from_name?(groupname : String)
+    groupname.check_no_null_byte
+
+    grp = uninitialized LibC::Group
+    grp_pointer = pointerof(grp)
+    initial_buf = uninitialized UInt8[1024]
+    buf = initial_buf.to_slice
+
+    ret = LibC.getgrnam_r(groupname, grp_pointer, buf, buf.size, pointerof(grp_pointer))
```
bcardiff: `out grp` could be used here and in other places https://crystal-lang.org/reference/syntax_and_semantics/c_bindings/out.html

spec/std/system/group_spec.cr
```diff
@@ -0,0 +1,89 @@
+require "spec"
+require "system/group"
+
+GROUP_NAME = {{ `id -gn`.stringify.chomp }}
+GROUP_ID   = {{ `id -g`.stringify.chomp }}
+
+describe System::Group do
+  describe "find_by" do
```
straight-shoota: Just a minor detail: `find_by(*, name)` and `find_by(*, id)` are entirely different methods. I wouldn't group them in the same describe block. Instead, the nilable and not-nilable variants can be grouped because they do the exact same thing and only differ in error handling.

spec/std/system/group_spec.cr
```diff
@@ -0,0 +1,89 @@
...
+      group.name.should eq(GROUP_NAME)
+    end
+
+    it "returns nil on nonexistent group id" do
+      group = System::Group.find_by?(id: "1234567")
+      group.should eq(nil)
+    end
+  end
+
+  describe "name" do
```
straight-shoota: Also minor detail: It's good practice to prefix spec groups for instance methods by `#` and class methods by `.`.
```suggestion
  describe "#name" do
```
This is not a requirement, though.

src/system/user.cr
```diff
@@ -0,0 +1,53 @@
...
+  class NotFoundError < Exception
+  end
+
+  extend Crystal::System::User
+
+  getter username : String
+  getter user_id : String
+  getter group_id : String
+  getter name : String
+  getter directory : String
```
straight-shoota: I'd prefer to call this property `home` or `home_directory`. Any thoughts on this?

src/system/group.cr
```diff
@@ -0,0 +1,50 @@
+require "crystal/system/group"
+
+class System::Group
```
asterite: This is missing docs describing what is a Group and how you generally use this class.

src/system/group.cr
```diff
@@ -0,0 +1,50 @@
...
+  # Raised on group lookup failure.
+  class NotFoundError < Exception
+  end
+
+  extend Crystal::System::Group
+
+  getter name : String
+  getter id : String
+  getter members : Array(String)
```
asterite: All these properties are missing docs.

src/system/group.cr
```diff
@@ -0,0 +1,50 @@
...
+
+  def_equals_and_hash @id
+
+  private def initialize(@name, @id, @members)
+  end
+
+  # Returns the group associated with the given name.
+  #
+  # Raises `NotFoundError` if no such group exists.
+  def self.find_by(*, name)
```
asterite: Can we add a type restriction to `name`, like `name : String`? Same with all other methods that lack a type restriction. And put a return type too.

src/system/group.cr
```diff
@@ -0,0 +1,50 @@
...
+
+  # Returns the group associated with the given ID.
+  #
+  # Returns `nil` if no such group exists.
+  def self.find_by?(*, id)
+    from_id?(id)
+  end
+
+  def to_s(io)
+    io << "#{name} (#{id})"
```
asterite: This should be:

```crystal
def to_s(io)
  io << name << " (" << id << ")"
end
```

otherwise it's allocating memory for the string interpolation.

src/system/user.cr
```diff
@@ -0,0 +1,53 @@
+require "crystal/system/user"
+
+class System::User
```
asterite: This is missing docs and general example usage.

src/system/user.cr
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  extend Crystal::System::User
+
+  getter username : String
+  getter user_id : String
+  getter group_id : String
+  getter name : String
+  getter directory : String
+  getter shell : String
```
asterite: All of these properties are missing docs.

src/crystal/system/unix/user.cr
```diff
@@ -0,0 +1,47 @@
...
+      ret = LibC.getpwnam_r(username, pwd_pointer, buf, buf.size, pointerof(pwd_pointer))
+    end
+
+    raise Errno.new("getpwnam_r") if ret != 0
+
+    from_struct(pwd) if pwd_pointer
+  end
+
+  private def from_id?(id : String)
+    id = id.to_u32
```
asterite: This line can raise, and because this is a `?` method maybe we should do `id.to_u32?` and if it's `nil` return `nil` (not sure).

src/crystal/system/unix/group.cr
```diff
@@ -0,0 +1,58 @@
+require "c/grp"
+
+module Crystal::System::Group
+  private GETGR_R_SIZE_MAX = 1024 * 16
+
+  private def extract_members(gr_mem)
+    members = Array(String).new
```
asterite: (optional!) One way to optimize this is to first count `n`. Then we can do `members = Array(String).new(n)` which should avoid potential memory reallocations.

src/crystal/system/unix/user.cr
straight-shoota: Sounds good. When using a name that's not a valid UInt32, it can't find a user by that id, so it should return `nil`.

src/crystal/system/unix/group.cr
woodruffw: Yep, this is what I ran into.

woodruffw:
Thanks for the reviews @bcardiff / @straight-shoota / @asterite! I'll address these tonight (EST).

RX14:
I see no reason to drop `Group#members`. It's useful, and optimizations can easily be applied later.

straight-shoota:
Is it actually useful? I fail to see any common use case for this.
We don't even have an API to list all users or all groups. These are all things that should go into a shard.

Regardless of usefulness, I'm not sure this method should be in the stdlib implementation, regardless of performance implications. I'd rather not add it here. It's certainly not as important as the entire rest of this PR.

src/system/user.cr
woodruffw: I like `home_directory`!

asterite:
Yes, I'm also not sure about `Group#members`, mainly because it's not there in Go so I'm sure there must be a good reason (maybe there's no such thing in Windows or it's expensive to compute, and with this way we'd be making this call always expensive there?).

We can always add it later if we find out there's no problem.

Also, we can always remove it later if we find a problem, after all we are 0.x.

straight-shoota:
> I'm sure there must be a good reason

I'm now pretty convinced that this is for memory implications. Imagine iterating a list of N files which all belong to a group of M users. When retrieving the group for each file, this will allocate N arrays of M `System::User` instances. And you'll most likely don't ever need any of them.

woodruffw:
Okay, nixed `#members`.

asterite:
Cool! Should this make it into 0.30.0?

bcardiff:
Yes, it's a new feature, it's safe to be added. 

Thanks @woodruffw and everybody involved

woodruffw:
Yay! Many thanks to all who reviewed.

PR #7730, opened 2019-05-02
Fix String#camelcase(lower: true)

asterite:
Correctly implements #7717

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -3499,21 +3499,23 @@ class String
 
   # Converts underscores to camelcase boundaries.
   #
+  # If *lower* is true, lower camelcase will be returned (the first letter is downcased).
+  #
   # ```
   # "eiffel_tower".camelcase                                            # => "EiffelTower"
-  # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
+  # "empire_state_building".camelcase(lower: true)                      # => "empireStateBuilding"
   # "isolated_integer".camelcase(options: Unicode::CaseOptions::Turkic) # => "İsolatedİnteger"
   # ```
-  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, downcase : Bool = false)
+  def camelcase(options : Unicode::CaseOptions = Unicode::CaseOptions::None, *, lower : Bool = false)
     return self if empty?
 
-    first = !downcase
+    first = true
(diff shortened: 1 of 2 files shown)


vladfaust:
I'm sorry, but `camelcase(downcase: true)` is tautology. It's even worse given the fact that we have `downcase` method.

asterite:
The thing is that programming will never read as a natural language so from time to time you have to make compromises. It's also not a big deal, this is not a method that's going to be used a lot.

asterite:
The thing is that programming will never read as a natural language so from time to time you have to make compromises. It's also not a big deal, this is not a method that's going to be used a lot.

bew:
What about `camelcase(first_low: true)` ?

asterite:
Maybe we can just have `lower_camelcase` and that's it.

j8r:
@asterite @vladfaust ideally, we have to express:
- camelcase - already in the method name (nothing to do)
- lower/down case - `down` or `lower`
- the first character is impacted - `first`

So we either have `down_first` or `lower_first`, as @straight-shoota suggested before.
`first_down` and `first_lower` are also solutions, but less readable.

src/string.cr
```diff
@@ -3499,21 +3499,23 @@ class String
 
   # Converts underscores to camelcase boundaries.
   #
+  # If *lower* is true, lower camelcase will be returned (the first letter is downcased).
+  #
   # ```
   # "eiffel_tower".camelcase                                            # => "EiffelTower"
   # "empire_state_building".camelcase(downcase: true)                   # => "empireStateBuilding"
```
bew:
```suggestion
  # "empire_state_building".camelcase(lower: true)                      # => "empireStateBuilding"
```

straight-shoota:
From https://github.com/crystal-lang/crystal/pull/7731#issuecomment-488674607 by @asterite 

> Actually, I think `lower` is fine. In wikipedia there's "upper camel case" and "lower camel case", so `camelcase(lower: true)` would use the "lower" variant.

+1 from me

asterite:
@vladfaust Thank you for the suggestion, I applied the changes.

src/string.cr
straight-shoota: @asterite This needs to be changed

asterite: Thanks

PR #7731, opened 2019-05-02
Fix String#camelcase(lower: true)

vladfaust:
:eyes:


j8r:
I don't see how `lower` is better than `downcase`, really.
That's a bit offtopic, but should the API use `lowercase` or `downcase`, `uppercase` or `upcase` generally? Their use are mixed as for now.

asterite:
Actually, I think `lower` is fine. In wikipedia there's "upper camel case" and "lower camel case", so `camelcase(lower: true)` would use the "lower" variant.

vladfaust:
👍

straight-shoota:
For reference: duplicate of #7730

PR #7732, opened 2019-05-02
Restrict bsearch block type

### LINKED ISSUE #7727: why ruby and crystal are giving completly different result when using bsearch ?
**This is my program:**
```
statuses = ["Cancelled", "Completed", "Open", "Optional", "Shipped"]
p %{bsearch  s <=> "Completed"}
p statuses.bsearch{|s| s <=> "Completed"}
p "     ------   "
p %{ bsearch "Completed" <=> s}
p statuses.bsearch{|s| "Completed" <=> s}
p "     ------   "
p %{ bsearch s == "Open"}
p statuses.bsearch{|s| s == "Open"}
p "     ------   "
p %{ bsearch s == "Completed"}
p statuses.bsearch{|s| s == "Completed"}
p "     ------   "
p %{ bsearch "Completed" == s}
p statuses.bsearch{|s| "Completed" == s}
p "     ******   "
p "     ------   "
p %{ bsearch_index  s <=> "Completed"}
p statuses.bsearch_index { |s, i|  s <=> "Completed"}
p "     ------   "
p %{ bsearch_index  "Completed" <=> s}
p statuses.bsearch_index { |s, i|  "Completed" <=> s}
p "     ------   "
p %{ bsearch_index s == "Open"}
p statuses.bsearch_index{|s,i| s == "Open"}
p "     ------   "
p %{ bsearch_index s == "Completed"}
p statuses.bsearch_index{|s,i| s == "Completed"}
```

Result with **ruby**:
```
"bsearch  s <=> \"Completed\""
nil
"     ------   "
" bsearch \"Completed\" <=> s"
"Completed"
"     ------   "
" bsearch s == \"Open\""
"Open"
"     ------   "
" bsearch s == \"Completed\""
nil
"     ------   "
" bsearch \"Completed\" == s"
nil
"     ******   "
"     ------   "
" bsearch_index  s <=> \"Completed\""
nil
"     ------   "
" bsearch_index  \"Completed\" <=> s"
1
"     ------   "
" bsearch_index s == \"Open\""
2
"     ------   "
" bsearch_index s == \"Completed\""
nil
```

Result with **crystal**:
```
"bsearch  s <=> \"Completed\""
"Cancelled"
"     ------   "
" bsearch \"Completed\" <=> s"
"Cancelled"
"     ------   "
" bsearch s == \"Open\""
"Open"
"     ------   "
" bsearch s == \"Completed\""
nil
"     ------   "
" bsearch \"Completed\" == s"
nil
"     ******   "
"     ------   "
" bsearch_index  s <=> \"Completed\""
0
"     ------   "
" bsearch_index  \"Completed\" <=> s"
0
"     ------   "
" bsearch_index s == \"Open\""
2
"     ------   "
" bsearch_index s == \"Completed\""
nil
```

Can anyone tell me which language has it right, and why, or are implementations of **bsearch** doing it right but with completly different intents ;)

asterite:
Fixes #7727

Currently `Indexable#bsearch`, `Indexable#bsearch_index` and `Range#bsearch` all take a block that accepts anything as a return value. This is a bit unintuitive because the block should only return
`true` or `false`. It also happens that Ruby provides the same functionality but an integer can be returned from the block and this might confuse people coming from Ruby.

I thought about making the block return `Bool | Nil` but I think it doesn't make sense for `Nil` to appear there (`nil` is usually combined with non-bool types, but those types can't make it into `bsearch`'s block).

### DIFF
--- a/src/indexable.cr
+++ b/src/indexable.cr
@@ -147,7 +147,7 @@ module Indexable(T)
   # [2, 5, 7, 10].bsearch { |x| x >= 4 } # => 5
   # [2, 5, 7, 10].bsearch { |x| x > 10 } # => nil
   # ```
-  def bsearch
+  def bsearch(&block : T -> Bool)
     bsearch_index { |value| yield value }.try { |index| unsafe_fetch(index) }
   end
 
@@ -166,7 +166,7 @@ module Indexable(T)
   # [2, 5, 7, 10].bsearch_index { |x, i| x >= 4 } # => 1
   # [2, 5, 7, 10].bsearch_index { |x, i| x > 10 } # => nil
   # ```
-  def bsearch_index
+  def bsearch_index(&block : T, Int32 -> Bool)
     (0...size).bsearch { |index| yield unsafe_fetch(index), index }
   end
 
--- a/src/range/bsearch.cr
+++ b/src/range/bsearch.cr
@@ -81,7 +81,7 @@ struct Range(B, E)
   # ```
   #
   # Returns `nil` if the block didn't return `true` for any value.
-  def bsearch
+  def bsearch(&block : B | E -> Bool)
     from = self.begin
     to = self.end
 


src/indexable.cr
```diff
@@ -166,7 +166,7 @@ module Indexable(T)
   # [2, 5, 7, 10].bsearch_index { |x, i| x >= 4 } # => 1
   # [2, 5, 7, 10].bsearch_index { |x, i| x > 10 } # => nil
   # ```
-  def bsearch_index
+  def bsearch_index(&block : T, Int32 -> Bool)
```
wontruefree: Since this is returning an index should the return type also be restricted?

```suggestion
  def bsearch_index(&block : T, Int32 -> Bool) : Int32
```

asterite: I guess so, but it has nothing to do with the intention of this PR.

wontruefree: Just thought I would mention it :)

PR #7734, opened 2019-05-02
Compiler: let puts string in macro omit quotes

### LINKED ISSUE #7544: Macro puts without quotes
Currently `{% puts "Foo" %}` produces the following output:

```
"Foo"
```

I expect the output to be without quotes:

```
Foo
```

And if I need quotes, I should use `{% pp "Foo" %}` to follow the runtime methods.

asterite:
Fixes #7544


asterite:
@straight-shoota I agree. That's what I had originally and later I refactored it, but I'll bring the separate methods back.

PR #7742, opened 2019-05-06
 Implementation: allow to lookup class and module implementations

### LINKED ISSUE #4941: Allow to find modules and classes using crystal tool
Hi people!

Currently `crystal tool hierarchy` isn't showing class location.

```crystal
class MiClase
end

a = MiClase.new
a
```

```
$ crystal tool hierarchy mi_clase.cr -e MiClase --no-color
- class Object (4 bytes)
  |
  +- class Reference (4 bytes)
     |
     +- class MiClase (4 bytes)
```

I think it is very useful to find clases using extensions like [vscode-crystal-lang](https://github.com/crystal-lang-tools/vscode-crystal-lang)

This idea was suggested  by @Qwerp-Derp

> Can you add support for finding the definition of a function/class/module/etc., like with other language plugins in VSCode?
> So for example, in this code:

```crystal
def foo
  Bar::Baz(10)
end
```

> If I go to Bar and right-click, and select "Peek definition", it should bring me to where Bar is defined

Some similar to https://github.com/crystal-lang/crystal/issues/4766

/cc @MakeNowJust :sweat_smile:

MakeNowJust:
Fixed #4941

Image:

I think it is totally useful, however this implementation cannot work on some corner case. One corner case, method argument restriction:

```crystal
def foo(x : ‸Foo)
end

foo
```

It is hard to implement for now. Another corner case, macro argument:

```crystal
record Foo,
  x : ‸Bar
```

It is impossible.

- - -

@faustinoaq Sorry for the late. Is this okay?

### DIFF
--- a/spec/compiler/crystal/tools/implementations_spec.cr
+++ b/spec/compiler/crystal/tools/implementations_spec.cr
@@ -468,4 +468,12 @@ describe "implementations" do
     foo
     )
   end
+
+  it "find const implementation" do
+    assert_implementations %(
+    ༓Foo = 42
+
+    F‸oo
+    )
+  end
 end
--- a/spec/compiler/crystal/tools/implementations_spec.cr
+++ b/spec/compiler/crystal/tools/implementations_spec.cr
@@ -368,4 +368,104 @@ describe "implementations" do
     Bar.bar
     )
   end
+
+  it "find class implementation" do
+    assert_implementations %(
+    ༓class Foo
+    end
+
+    F‸oo
+    )
+  end
+
+  it "find open class implementation" do
+    assert_implementations %(
+    ༓class Foo
+      def foo
+      end
+    end
+
+    ༓class Foo
+      def bar
+      end
+    end
+
+    F‸oo
+    )
+  end
+
+  it "find struct implementation" do
+    assert_implementations %(
+    ༓struct Foo
+    end
+
+    F‸oo
+    )
+  end
+
+  it "find module implementation" do
+    assert_implementations %(
+    ༓module Foo
+    end
+
+    F‸oo
+    )
+  end
+
+  it "find enum implementation" do
+    assert_implementations %(
+    ༓enum Foo
+      Foo
+    end
+
+    F‸oo
+    )
+  end
+
+  it "find enum value implementation" do
+    assert_implementations %(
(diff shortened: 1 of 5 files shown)


spec/compiler/crystal/tools/implementations_spec.cr
```diff
@@ -368,4 +368,104 @@ describe "implementations" do
     Bar.bar
     )
   end
+
+  it "find class implementation" do
+    assert_implementations %(
+    ༓class Foo
+    end
+
+    F‸oo
```
bew: There are some weird chars before `class` and just after the `F` of `Foo` on L377 (here and on all other specs)

Sija: That's intended (see whole spec for details).

bew: Oooh right, that's neat actually :smiley:

MakeNowJust:
ping. Could anyone review this?

spec/compiler/crystal/tools/implementations_spec.cr
```diff
@@ -368,4 +368,104 @@ describe "implementations" do
...
+    Foo::F‸oo
+    )
+  end
+
+  it "find class implementation via alias" do
+    assert_implementations %(
+    ༓class Foo
+    end
+
+    alias Bar = Foo
```
bcardiff: Was there an explicit decision to not resolve to this line instead the Foo? What happens if the alias has a union?

MakeNowJust:
@bcardiff Thank you for reviewing!

I fixed `crystal tool implementation` to lookup alias name without resolve. For example:

`foo.cr`:

```crystal
class Foo
end

class Bar
end

alias Baz = Foo | Bar

Baz
```

```console
$ bin/crystal tool implementation -c foo.cr:9:1 foo.cr
1 implementation found
foo.cr:7:1
```

Unfortunately this doesn't work:

```console
$ bin/crystal tool implementation -c foo.cr:7:15 foo.cr
no implementations or method call found
```

I think implementing this is hard for now...

faustinoaq:
Hi @MakeNowJust Yeah, this is a progress, Thank you! :+1:

PR #7743, opened 2019-05-06
Compiler: fix bug related to types that can't be inferred

### LINKED ISSUE #6390: [BUG] "cast from Nil to Crystal::GenericInstanceType+ "
Hello, I found a bug while solving [this](urhttps://projecteuler.net/problem=82l)

Here are some details:
```
cast from Nil to Crystal::GenericInstanceType+ failed, at /crystal/src/compiler/crystal/types.cr:1543:20:1543 (TypeCastError)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
$ crystal -v

LLVM: 4.0.0

$ uname -a
Linux ubuntu 4.15.0-23-generic #25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

```

The source is:

```crystal
# @matrix = File.read_lines("../081.in")
#              .map do |row|
#                row.split(',').map { |x| x.to_i }
#              end
matrix = [
  [131, 673, 234, 103, 18],
  [201, 96, 342, 965, 150],
  [630, 803, 746, 422, 111],
  [537, 699, 497, 121, 956],
  [805, 732, 524, 37, 331]
]

class P082
  Steps = [{ 1, 0 }, { -1, 0 }, { 0, 1 }]

  getter :best

  def initialize(matrix)
    @best = 10**9
    @n = matrix.size
    @matrix = matrix
    @visited = Array.new(N) { Array.new(N, false) }
  end

  def find(i, j, current)
    return if current > @best
    if j == @n - 1
      @best = [current, @best].min
      return
    end
    @visited[i][j] = true
    current += @matrix[i][j]
    Steps.each do |di, dj|
      new_i = di + i
      new_j = dj + j
      next if new_i < 0 || new_i >= @n
      next if new_j < 0 || new_j >= @n
      next if @visited[new_i][new_j]
      find(new_i, new_j, @visited, current)
    end

    current -= @matrix[i][j]
    @visited[i][j] = false
  end

  def solve
    (0...@n).each do |i|
      find(i, 0, 0)
    end
  end
end

solution = P082.new(matrix)
solution.solve
puts solution.best
```

I'm happy of helping at least with a bug reported. 
If I can help don't hesitate on contact me. 

Thanks guys and keep up the good job!

asterite:
This PR fixes #6390 and also improves error messages a bit.

For example before this PR, if you had:

```crystal
class Foo
  def initialize
    @x = Array.new(3, 10)
  end
end

Foo.new
```

the compiler would say:

```
Error in foo.cr:3: can't use Array(T) as the type of instance variable @x of Foo, use a more specific type

    @x = Array.new(3, 10)
         ^~~~~
```

That's probably a bit misleading: it's leaking the fact that the compiler infers the above to be an uninstantiated array type and says "you can't do that".

With this PR the error message is:

```
Error in foo.cr:7: instantiating 'Foo.class#new()'

Foo.new
    ^~~

in foo.cr:3: can't infer the type of instance variable '@x' of Foo
...
```

I think that's better: it's less misleading and it might hint that your best choice is to put a type annotation on `@x`.

This is also better because for this code:

```crystal
class Foo
  def initialize
    @x = Array.new(3, 10 + "x")
  end
end

Foo.new
```

you used to get:

```
Error in foo.cr:3: can't use Array(T) as the type of instance variable @x of Foo, use a more specific type

    @x = Array.new(3, 10 + "x")
         ^~~~~
```

that is, even if we have a semantic error in that line (`10 + "x"`) the compiler would first complain that it can't infer the type. With this PR the error is:

```
Error in foo.cr:7: instantiating 'Foo.class#new()'

Foo.new
    ^~~

in foo.cr:3: no overload matches 'Int32#+' with type String
Overloads are:
 - Int32#+(other : Int8)
 - Int32#+(other : Int16)
 - Int32#+(other : Int32)
 - Int32#+(other : Int64)
 - Int32#+(other : Int128)
 - Int32#+(other : UInt8)
 - Int32#+(other : UInt16)
 - Int32#+(other : UInt32)
 - Int32#+(other : UInt64)
 - Int32#+(other : UInt128)
 - Int32#+(other : Float32)
 - Int32#+(other : Float64)
 - Number#+()

    @x = Array.new(3, 10 + "x")
                         ^
```

### DIFF
--- a/src/compiler/crystal/semantic/type_guess_visitor.cr
+++ b/src/compiler/crystal/semantic/type_guess_visitor.cr
@@ -1073,21 +1073,16 @@ module Crystal
     end
 
     def check_allowed_in_generics(node, type)
-      # Types such as Object, Int, etc., are not allowed in generics
-      # and as variables types, so we disallow them.
-      if type && !type.allowed_in_generics?
-        @error = Error.new(node, type)
-        return nil
-      end
-
-      case type
-      when GenericClassType
+      if type.is_a?(GenericClassType)
+        nil
+      elsif type.is_a?(GenericModuleType)
+        nil
+      elsif type && !type.allowed_in_generics?
+        # Types such as Object, Int, etc., are not allowed in generics
+        # and as variables types, so we disallow them.
         @error = Error.new(node, type)
         nil
-      when GenericModuleType
-        @error = Error.new(node, type)
-        nil
-      when NonGenericClassType
+      elsif type.is_a?(NonGenericClassType)
         type.virtual_type
       else
         type
--- a/src/compiler/crystal/semantic/type_guess_visitor.cr
+++ b/src/compiler/crystal/semantic/type_guess_visitor.cr
@@ -631,6 +631,14 @@ module Crystal
       return type if type
 
       type = guess_type_call_with_type_annotation(node)
+
+      # If the type is unbound (uninstantiated generic) but the call
+      # wasn't something like `Gen(Int32).something` then we can never
+      # guess a type, the type is probably inferred from type restrictions
+      if !obj.is_a?(Generic) && type.try &.unbound?
+        return nil
+      end
+
       return type if type
 
       nil
--- a/spec/compiler/semantic/class_spec.cr
+++ b/spec/compiler/semantic/class_spec.cr
@@ -516,7 +516,7 @@ describe "Semantic: class" do
       foo = Foo.new
       foo.@y
       ),
-      "Can't infer the type of instance variable '@y' of Foo"
+      "can't infer the type of instance variable '@y' of Foo"
   end
 
   it "errors if reading ivar from non-ivar container" do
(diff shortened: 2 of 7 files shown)


asterite:
@straight-shoota Because it's lower case everywhere else. It was accidentally changed by someone else (I think @Sija) when we standarized uppercase name for stdlib runtime errors, but this is a compiler error and there we are using lowecase (I think we should use lowercase everywhere, but that's a separate discussion).

src/compiler/crystal/semantic/type_guess_visitor.cr
```diff
@@ -631,6 +631,14 @@ module Crystal
       return type if type
 
       type = guess_type_call_with_type_annotation(node)
+
+      # If the type is unbound (uninstantiated generic) but the call
+      # wasn't something like `Gen(Int32).something` then we can never
+      # guess a type, the type is probably inferred from type restrictions
+      if !obj.is_a?(Generic) && type.try &.unbound?
```
asterite: To be honest, I'm not super happy with this solution but:
- it kind of makes sense
- it gets rid of a bug that seems to be very popular (it was reported countless times already)
- fixing it another way might involve a bigger refactor but I don't have time for it right now

PR #7746, opened 2019-05-07
 Compiler: make automatic cast work with `with ... yield`

### LINKED ISSUE #7736: Automatic casting of symbols to enums broken within blocks
There is a bug with automatic casting of symbols to enum values when using with self yield in a block. 

Reproducible example:
```crystal
enum BorderSide
  Top
  Bottom
end

class Table
 
  def config
    with self yield
  end

  def row(border : BorderSide)
    "bar"
  end
end

Table.new.config do
  row(border: :top)
end
```

with this error
`undefined method 'row' for Table (with ... yield) and top-level (current scope)`

its works fine when `BorderSide::Top` is used
```crystal
Table.new.config do
  row(border: BorderSide::Top)
end
```

asterite:
Fixes #7736

There's also a separate commit that makes exceptions show the full (crystal) backtrace when they fail on specs. It was driving me a bit crazy, not being able to see why spec failed (it got broke some time ago by someone else).

### DIFF
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -38,6 +38,9 @@ end
 
 class Crystal::Call
   def raise_matches_not_found(owner, def_name, arg_types, named_args_types, matches = nil, with_literals = false)
+    obj = @obj
+    with_scope = @with_scope
+
     # Special case: Foo+.class#new
     if owner.is_a?(VirtualMetaclassType) && def_name == "new"
       raise_matches_not_found_for_virtual_metaclass_new owner
@@ -71,10 +74,13 @@ class Crystal::Call
       end
     end
 
+    # Also check with scope
+    if with_scope
+      defs.concat with_scope.lookup_defs(def_name)
+    end
+
     # Check if it's the case of an abstract def
     check_abstract_def_error(owner, matches, defs, def_name)
-
-    obj = @obj
 
     # Check if this is a `foo` call and we actually find it in the Program
     if !obj && defs.empty?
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/semantic/call_error.cr
```diff
@@ -71,6 +71,12 @@ class Crystal::Call
       end
     end
 
+    # Also check with scope
+    with_scope = @with_scope
```
asterite: I don't inline it because it's also used later on.

bew:
```cr
if foo = @foo 
  # ...
end
# foo is still usable here
```
But I understand. To make it more clear, I'd add a blank line after the assign:
```
foo = @foo 

if foo
```
To show that it's not only for the `if`

asterite: Good idea! Done, and also moved another var like that.

src/compiler/crystal/semantic/exception.cr
```diff
@@ -72,6 +72,17 @@ module Crystal
       end
     end
 
+    def inspect_with_backtrace(io : IO) : Nil
```
RX14: I'm confused - this method is defined in `::Exception` and it doesn't seem to be overridden.

asterite: What do you mean? It's overwritten here.

bew: I think RX14 wants to know you're overwritting it, doesn't the method in Exception already does that?

asterite: The bodies are different: `::Exception` appends the `message`, but this one calls `to_s(io)`

Issue #7553, opened 2019-03-15
Revamp compile error formatting & output


paulcsmith:
Based on proposal in https://forum.crystal-lang.org/t/rfc-revamp-how-compile-time-errors-are-shown/548/3

## The Proposed Changes

* By default, just show the code with the error, not the entire trace. This is what Elm does and it works quite well. Helps focus on what is necessary.
* Add a command line option and/or ENV to show the full trace. Sometimes it is useful to see more, but let's make that optional instead of making the current verbose output the default. 
* Some changes to how the error is displayed to make it easier to scan at a glance. Code + pics attached

## Example Code and Screenshots

https://github.com/paulcsmith/crystal-errors

### For errors without more info (just one line)

* Prefix with "Error"
* Print error message in yellow and bold
* (not sure about this) show how to print the full trace (like it does now)
* Show what file it is
* Show line number right by the source code line
* Show code line (maybe bolded?) 

https://github.com/paulcsmith/crystal-errors/blob/master/errors/type_restriction.cr

<img alt="Screen Shot 2019-03-14 at 9 49 07 PM" src="https://user-images.githubusercontent.com/22394/54437417-f0ff0280-470a-11e9-9e2d-8503b61d9282.png">

### For errors with more than one line

* Print first line in yellow just like other errors
* **Everything after first newline goes after the source line.**

https://github.com/paulcsmith/crystal-errors/blob/master/errors/no_overload.cr

<img alt="Screen Shot 2019-03-14 at 9 48 37 PM" src="https://user-images.githubusercontent.com/22394/54437329-c44aeb00-470a-11e9-9df5-7746284b90b7.png">

### Macro errors

* Print surrounding lines
* Only bold line with the problem
* **Show the expanded code, and underneath it show the unexpanded code**

It is important to show both. The unexpanded part show what the user wrote. The expanded part helps figure out why the issue might be there. Use --verbose/--debug for more in-depth look if it still is not enough

https://github.com/paulcsmith/crystal-errors/blob/master/errors/macro_error.cr

<img alt="Screen Shot 2019-03-14 at 10 18 12 PM" src="https://user-images.githubusercontent.com/22394/54437336-c90f9f00-470a-11e9-8347-ff711b9aeef0.png">

### Flag name `--trace-error`

`crystal build src/my_app.cr --trace-error`

This will print all the frames. Surround error message with `--` and keep it yellow and bold to highlight the final frame and the error message

[...]

PR #7748, opened 2019-05-07
Revamp compile error formatting & output

martimatix:
Suggested in #7553

### DIFF
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -134,17 +134,25 @@ module Crystal
       end
     end
 
+    def filename_row_col_message(io, filename, line_number, column_number)
+      io << "In "
+      io << colorize("#{relative_filename(filename)}:#{line_number}:#{column_number}").underline
+    end
+
     def format_error(filename, lines, line_number, column_number, size = 0)
       String.build do |io|
-        return "In #{filename}" unless line_number
+        return "In #{relative_filename(filename)}" unless line_number
         line = lines[line_number - 1]?
-        return "In #{filename}:#{line_number}:#{column_number}" unless line
+
+        unless line
+          return filename_row_col_message(io, filename, line_number, column_number)
+        end
 
         case filename
         when String
-          io << "In #{relative_filename(filename)}"
+          filename_row_col_message(io, filename, line_number, column_number)
         when VirtualFile
-          io << "In macro '#{filename.macro.name}'"
+          io << "In macro '" << colorize("#{filename.macro.name}").underline << '\''
         else
           io << "In unknown location"
         end
@@ -204,18 +212,21 @@ module Crystal
     def append_macro_definition_location(io, filename : VirtualFile)
       macro_source = filename.macro.location
       source_filename = macro_source.try &.filename
+      line_number = macro_source.try &.line_number
+      column_number = macro_source.try &.column_number
 
-      io << "Macro defined in " << case source_filename
+      io << "Macro defined in "
+
+      case source_filename
       when String
-        "#{relative_filename(source_filename)}"
+        io << colorize("#{relative_filename(source_filename)}:#{line_number}:#{column_number}").underline
       when VirtualFile
-        "macro '#{source_filename.macro.name}'"
+        io << "macro '" << colorize("#{source_filename.macro.name}").underline << '\''
       else
         "unknown location"
       end
 
       lines = source_lines(source_filename)
-      line_number = macro_source.try &.line_number
 
       if lines && line_number
         io << "\n\n"
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -186,9 +186,7 @@ module Crystal
       String.build do |io|
         return if lines.empty?
         io << "\n\n"
-        lines
-          .skip_while(&.blank?)
-          .each { |line| io << line << '\n' }
+        io << lines.skip_while(&.blank?).join('\n')
       end
     end
 
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -188,11 +188,11 @@ module Crystal
         if show_where_macro_expanded
           append_where_macro_expanded(io, virtual_file)
           io << '\n'
-          io << "called macro "
         end
+        io << "Called macro defined in "
         append_macro_definition_location(io, virtual_file)
         io << "\n\n"
-        io << "which expanded to:"
+        io << "Which expanded to:"
         io << "\n\n"
         append_expanded_macro(io, virtual_file.source)
       end
@@ -222,8 +222,6 @@ module Crystal
       source_filename = macro_source.try &.filename
       line_number = macro_source.try &.line_number
       column_number = macro_source.try &.column_number
-
-      io << "defined in "
 
       case source_filename
       when String
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -181,12 +181,14 @@ module Crystal
     end
 
     def format_macro_error(virtual_file : VirtualFile)
+      show_where_macro_expanded = !(@error_trace && self.responds_to?(:error_trace=))
       String.build do |io|
         io << "There was a problem expanding macro '#{virtual_file.macro.name}'"
         io << "\n\n"
-        unless @error_trace && self.responds_to?(:error_trace=)
+        if show_where_macro_expanded
           append_where_macro_expanded(io, virtual_file)
-          io << "\n\n"
+          io << '\n'
+          io << "called macro "
         end
         append_macro_definition_location(io, virtual_file)
         io << "\n\n"
@@ -221,7 +223,7 @@ module Crystal
       line_number = macro_source.try &.line_number
       column_number = macro_source.try &.column_number
 
-      io << "called macro defined in "
+      io << "defined in "
 
       case source_filename
       when String
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -141,15 +141,14 @@ module Crystal
     end
 
     def format_error(filename, lines, line_number, column_number, size = 0)
+      return "#{relative_filename(filename)}" unless line_number
+      
+      unless line = lines[line_number - 1]?
+        return "in #{filename_row_col_message(filename, line_number, column_number)}"
+      end
+
       String.build do |io|
         io << "in "
-        return "#{relative_filename(filename)}" unless line_number
-        line = lines[line_number - 1]?
-
-        unless line
-          return filename_row_col_message(filename, line_number, column_number)
-        end
-
         case filename
         when String
           io << filename_row_col_message(filename, line_number, column_number)
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -142,7 +142,7 @@ module Crystal
 
     def format_error(filename, lines, line_number, column_number, size = 0)
       return "#{relative_filename(filename)}" unless line_number
-      
+
       unless line = lines[line_number - 1]?
         return "in #{filename_row_col_message(filename, line_number, column_number)}"
       end
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -174,7 +174,7 @@ module Crystal
         io << "Was expanded to:"
         io << "\n\n"
         append_expanded_macro(io, virtual_file.source)
-        next if @all_frames && self.is_a? TypeException
+        next if @all_frames && self.responds_to?(:all_frames=)
         io << "\n\n"
         append_where_macro_expanded(io, virtual_file)
       end
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -107,10 +107,10 @@ module Crystal
     def error_body(source, default_message) : String | Nil
       case filename = @filename
       when VirtualFile
-        return format_error(filename)
+        return format_macro_error(filename)
       when String
         if File.file?(filename)
-          return format_error(File.read_lines(filename)).capitalize
+          return format_error_from_file(filename)
         end
       end
 
@@ -169,18 +169,21 @@ module Crystal
       end
     end
 
-    def format_error(lines : Array(String))
+    def format_error_from_file(filename : String)
+      lines = File.read_lines(filename)
       format_error(
         filename: @filename,
         lines: lines,
         line_number: @line_number,
         column_number: @column_number,
         size: @size
-      )
+      ).capitalize
     end
 
-    def format_error(virtual_file : VirtualFile)
+    def format_macro_error(virtual_file : VirtualFile)
       String.build do |io|
+        io << "There was a problem expanding macro '#{virtual_file.macro.name}'"
+        io << "\n\n"
         unless @error_trace && self.responds_to?(:error_trace=)
           append_where_macro_expanded(io, virtual_file)
           io << "\n\n"
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -227,23 +227,29 @@ module Crystal
     end
 
     def expanded_source_subsection(source, line_number)
-      expanded_source = Crystal.with_line_numbers(source, line_number, @color, hide_after_highlight: true, join_lines: false)
+      from_index = [0, line_number - MACRO_LINES_TO_SHOW].max
+      source_slice = source.lines[from_index...line_number]
 
-      case expanded_source
-      when String # just a single line
-        expanded_source
-      when Array(String)
-        if expanded_source.size <= MACRO_LINES_TO_SHOW
-          expanded_source.join '\n'
-        else
-          expanded_source[-(MACRO_LINES_TO_SHOW)..-1].join '\n'
+      min_leading_white_space =
+        source_slice.map do |line|
+          if match = line.match(/^(\s+)\S/)
+            spaces = match[1]?
+            spaces.size if spaces
+          end
         end
-      end
+        .compact
+        .min
+
+      source_slice = source_slice.map do |line|
+        replace_leading_tabs_with_spaces(line).lchop(" " * min_leading_white_space)
+      end.join('\n')
+
+      Crystal.with_line_numbers(source_slice, line_number, @color, line_number_start = from_index + 1)
     end
 
     def append_expanded_macro(io, source)
       line_number = @line_number
-      if @error_trace
+      if @error_trace || !line_number
         io << Crystal.with_line_numbers(source, line_number, @color)
       else
         io << expanded_source_subsection(source, line_number)
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -144,11 +144,10 @@ module Crystal
       return "#{relative_filename(filename)}" unless line_number
 
       unless line = lines[line_number - 1]?
-        return "in #{filename_row_col_message(filename, line_number, column_number)}"
+        return "#{filename_row_col_message(filename, line_number, column_number)}"
       end
 
       String.build do |io|
-        io << "in "
         case filename
         when String
           io << filename_row_col_message(filename, line_number, column_number)
@@ -170,13 +169,14 @@ module Crystal
 
     def format_error_from_file(filename : String)
       lines = File.read_lines(filename)
-      format_error(
+      formatted_error = format_error(
         filename: @filename,
         lines: lines,
         line_number: @line_number,
         column_number: @column_number,
         size: @size
-      ).capitalize
+      )
+      "In #{formatted_error}"
     end
 
     def format_macro_error(virtual_file : VirtualFile)
@@ -281,7 +281,7 @@ module Crystal
       lines = source_lines(source_filename)
       return unless lines
 
-      io << "Code " << format_error(
+      io << "Code in " << format_error(
         filename: source_filename,
         lines: lines,
         line_number: expanded_source.line_number,
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -4,11 +4,16 @@ require "colorize"
 module Crystal
   abstract class Exception < ::Exception
     property? color = false
+    property? all_frames = false
 
     @filename : String | VirtualFile | Nil
 
     def to_s(io) : Nil
       to_s_with_source(nil, io)
+    end
+
+    def warning=(warning)
+      @warning = !!warning
     end
 
     abstract def to_s_with_source(source, io)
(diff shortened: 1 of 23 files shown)


martimatix:
If you get the opportunity, please take it for a test drive. I would love to get some feedback.

src/compiler/crystal/syntax/virtual_file.cr
```diff
@@ -24,6 +24,6 @@ class Crystal::VirtualFile
   end
 
   def pretty_print(pp)
-    pp.text inspect
+    pp.text inspectatoma toma toma toma toma toma toma toma toma toma toma toma toma toma toma toma toma toma tom
```
Sija: Oops ;)

martimatix: Oops 😅

asterite: https://youtu.be/-0AX3_JhxyM?t=22

jwoertink:
I'm not totally sure on this, but according to [this commit](https://github.com/crystal-lang/crystal/commit/38029c416d35b4d7d1b58d7aad451d70aa47b0f4), and there was a comment somewhere from @asterite about compiler errors being lowercase (I think), you may need to update those.

paulcsmith:
So cool! Gonna try to give this a whirl soon! If it isn't too much to ask, could you post some screenshots?

paulcsmith:
Ok I got this working and have a few screenshots to share

Screen Shot 2019-05-07 at 1 04 42 PM
Screen Shot 2019-05-07 at 1 05 27 PM

I *really* love the macro one. Much easier to parse than what is there now. But both are fantastic.

A few things I think could make it even better:

* Unindent the code so that there isn't so much empty space and so it all lines up
* I noticed both macro tests I did showed `'macro_xxxxxxx` is that  a bug? Or maybe just something weird about how the Crystal self compiles? I don't know a lot about Crystal internals so this might work fine elsewhere 

I couldn't figure out how to get `--all-frames` working. I tried `bin/crystal spec --all-frames` but it seems to be ignored. How did you get this working? I don't work on Crystal core much so I'm probably just doing something dumb :)

paulcsmith:
One more to show how errors show details below the code line

Screen Shot 2019-05-07 at 1 19 15 PM

Looks like there may be an extra newline under the code, but otherwise I think this is really slick

asterite:
@paulcsmith With this code:

```crystal
# foo.cr
def foo
  "foo".abs
end

def bar
  foo
end

bar
```

I get:

```
$ bin/crystal foo.cr
Error: undefined method 'abs' for String
Showing last frame. Use --all-frames for full trace.

In foo.cr

 2 |   "foo".abs
             ^~~
```

With `--all-frames I get`:

```
$ bin/crystal foo.cr --all-frames
Error: instantiating 'bar()'

In foo.cr

 9 | bar
     ^~~
--
Error: instantiating 'foo()'

In foo.cr

 6 |   foo
       ^~~
--
Error: undefined method 'abs' for String

In foo.cr

 2 |   "foo".abs
             ^~~
```

paulcsmith:
Thanks @asterite. I must have been working with an error that only had 1 frame 🤦‍♂️

paulcsmith:
One other thing. What is the difference between `--error-trace` and `--all-frames`? I imagine this will be confusing to users because they seem very similar. It seems like maybe they should be combined or maybe we need to remove `--error-trace` and have `--all-frames` cover what it does?

asterite:
If I remember correctly, `--error-trace` will show you where a type comes from. Try it with this program:

```crystal
"foo".index('a').abs
```

But maybe it's not very useful... 

(though I agree that we should combine `--all-frames` with `--error-trace`, if we go this route)

martimatix:
This is great feedback @paulcsmith !

I can work on merging `--all-frames` and `--error-trace` and then we can agree on a flag name.

A note about `--all-frames` - it does not get displayed for syntax errors because they always show the last frame. I though about leaving the dimmed "showing last frame" message in there as a sanity check for whether the flag is being used but felt that in most cases, it is better to show less lines.

Re: `macro_xxxxxxx`, it's possible that I'm not looking up the macro source properly. Can anyone give me a small example to reproduce this?

I'll investigate `--all-frames` and spec.

And I'll also investigate that extra new line. Sometimes it's hard to tell because the caret symbol only uses the upper half of the line.

Also, I'll look into unindenting the code. Could get interesting if the code uses tabs.

oprypin:
> Could get interesting if the code uses tabs.

I wouldn't mess too much with it. Just remove the longest common prefix that matches `\s+`.

martimatix:
@paulcsmith regarding the macro names `macro_xxxxxxx`, are the macros defined in macros?

In the event that macros are defined within macros, it will use the macro name. If macros are defined in a file, then it will reference the file path.

martimatix:
Regarding the new line shown below in purple here - what do people think?

image

Below is a side by side of with and without.

image

From an implementation point of view, it's easier to have the new line because not all errors will have the error indicator (`^~~~~~`).

martimatix:
Funny that you mention replacing tildes with dashes because that's one of the first things I did as per the _first_ diagram in the [RFC](https://github.com/crystal-lang/crystal/issues/7553). But then the second diagram used tildes so I reverted the change. Made sense for me to do that at the time because Crystal presently uses tildes and thought the first diagram had a typo.

Personally, I like the dashes so I'll make that change - it's literally a one character change! If others feel strongly about the tilde, please let me know.

paulcsmith:
@martimatix I believe I took some screenshots before I changed the tilde to a dash. Oops! Thanks for making the change though. I hope others agree, I think it looks a lot cleaner :)

martimatix:
### Status update

I've addressed the suggestions made by @paulcsmith 

* `--all-frames` has now been merged with `--error-trace`. So in future, use `--error-trace`.
* Spec accepts `--error-trace`
* Indentation has been minimised
* Replace tilde with dash

Presently, I'm looking at splitting out the "Did you mean?" messages into multiple lines. Here's a screenshot:

image

There are several kinds of "Did you mean?" errors  - 12 occurrences across 5 non-spec files. They're all implemented differently and contexts vary. Therefore, I'm going to aim for only modifying the ones listed in [this spec file](https://github.com/crystal-lang/crystal/blob/master/spec/compiler/semantic/did_you_mean_spec.cr).

martimatix:
On review, I'm thinking that showing the `Rerun with --error-trace to show a complete error trace.` message that appears at the bottom is somewhat tautological because of the "Showing last frame..." message.

I suppose it indicates that there's some more information than revealing all error frames but I think anyone who's struggling to figure out a complex error would use `--error-trace` anyway. Therefore, I propose hiding the message for a cleaner look.

martimatix:
@paulcsmith I've implemented the suggestions you made. Let me know what you think.

paulcsmith:
Been super busy but will try to take a look soon. I did take a brief look earlier and what I saw looked freakin awesome!

paulcsmith:
@martimatix AWESOME work. I'll post some screenshots for everyone to see the updated stuff

*Misspelled method name*
Screen Shot 2019-05-31 at 2 52 32 PM
*Misspelled method name with --error-trace*
Screen Shot 2019-05-31 at 2 48 14 PM
*Misspelled method name inside a macro*
Screen Shot 2019-05-31 at 2 51 45 PM
*Method type mismatch*
Screen Shot 2019-05-31 at 2 55 30 PM

I think this looks great and is a big improvement over what's there. I might suggest a few small changes after playing wit this a bit

* Move the "Showing last frame..." To the very bottom. It kind of gets in the way of reading the error message right now. My bad :)
* If not too much work, I think it'd be good to bold/highlight the suggested method/class/whatever name in `Did you mean?` so it pops a bit more

None of these are show stoppers though and could be fixed/changed later! I think this is ready for a core team review at this point.

martimatix:
Thanks for the feedback @paulcsmith 

I'm also thinking that the filename, in addition to being bolded and underlined, could have the line number and column number. In the example above, it would be:

```
In spec/std/http/cookie_spec.cr:11:31
```

I noticed that other compilers -- such as the Rust compiler -- does this and it's super handy because you can copy and paste that line into your editor's file search and it will take you straight to the error. For iterm2 users, the story is even better because command clicking the filename with row and column numbers will take you to the error in a single click. I'm sure other terminal emulators also support this feature.

I'll implement the suggested changes first before asking the core team to review this PR. 🙂

r00ster91:
These 2 dashes between the errors don't look very pretty (rather quick and dirty) in my opinion. Can they maybe just be replaced by a newline? So like this:
image
IMO still pretty readable.

martimatix:
> These 2 dashes between the errors don't look very pretty

I like the dashes because it signifies where one error frame ends and the next one starts. My eyes feel a bit lost when replacing the dashes with a newline. Then again, I've been staring at these error messages for far too long and may have just got used to the dashes. 😄

What are everyone else's thoughts?

oprypin:
Keep the separation but make it anything other than just 2 dashes 😅

paulcsmith:
Maybe two dashes or two equal signs before the “Error” text rather than above it?

I agree that just the beeline makes it hard to see where one line ends and one begins

And yeah like and column number sounds good!

martimatix:
Perhaps we can do something with colour. For example, the error lines can be yellow but not bold except for the final frame which is both yellow and bold.

paulcsmith:
@martimatix that’s a great idea. The color probably differentiates it enough

martimatix:
image

As you can see in the screenshot above, I've removed the `--` between error frames.

Locations have been underlined and row and column numbers have been added. @paulcsmith suggested bold _and_ underline but I found that in my terminal it started to look overwhelming hence the bolding wasn't applied.

Moving the `Showing last frame. Use --error-trace for full trace.` message to the bottom of the error output sounds trivial but it has the additional burden of adding logic to another class (`MethodTraceException` for any interested readers). Therefore, I will not implement this feature in this first cut.

Bolding `Did you mean?` has also been left out because the logic is spread across multiple methods in several files. It's possible but just tedious work so I'll leave this as is for now.

paulcsmith:
>  @paulcsmith suggested bold and underline but I found that in my terminal it started to look overwhelming hence the bolding wasn't applied.

Just the underline is plenty. Good call!

> Moving the Showing last frame. Use --error-trace for full trace. message to the bottom of the error output sounds trivial but it has the additional burden of adding logic to another class (MethodTraceException for any interested readers). Therefore, I will not implement this feature in this first cut.

That sounds like a solid plan 👍 

> Bolding Did you mean? has also been left out because the logic is spread across multiple methods in several files. It's possible but just tedious work so I'll leave this as is for now.

Yeah let's not hold up this PR for that. It could be done later as a separate PR

Nice work! I'm stoked for this!!

martimatix:
Would core team members be able to review this PR?

asterite:
@martimatix I can review it but next week. Thank you!

asterite:
So, I reviewed this (I checked the error messages produced by the compiler, I didn't take a look at the code yet) and I'm not sure how things are shown is substantially better.

In particular, the main error message appears at the top, and then a large callstack might appear, and then you always have to scroll back to see what the error was.

With this code:

```crystal
def bar
  foo 1, 2
end

{% if true %}
  def foo(x, y)
    1 + 'a'
  end
{% end %}

bar
```

Current error message:

```
Error: no overload matches 'Int32#+' with type Char

Macro defined in foo.cr:5:1

 5 | {% if true %}

Was expanded to:

 > 1 |
 > 2 | def foo(x, y)
 > 3 |   1 + 'a'
           ^

In foo.cr:5:1

 5 | {% if true %}
     ^

Overloads are:
 - Int32#+(other : Int8)
 - Int32#+(other : Int16)
 - Int32#+(other : Int32)
 - Int32#+(other : Int64)
 - Int32#+(other : Int128)
 - Int32#+(other : UInt8)
 - Int32#+(other : UInt16)
 - Int32#+(other : UInt32)
 - Int32#+(other : UInt64)
 - Int32#+(other : UInt128)
 - Int32#+(other : Float32)
 - Int32#+(other : Float64)
 - Number#+()
```

Previous error message:

```
Error in foo.cr:11: instantiating 'bar()'

bar
^~~

in foo.cr:2: instantiating 'foo(Int32, Int32)'

  foo 1, 2
  ^~~

in foo.cr:5: expanding macro

{% if true %}
^

in macro 'macro_4396812144' /Users/asterite/Projects/crystal/foo.cr:5, line 3:

   1.
   2.   def foo(x, y)
>  3.     1 + 'a'
   4.   end

no overload matches 'Int32#+' with type Char
Overloads are:
 - Int32#+(other : Int8)
 - Int32#+(other : Int16)
 - Int32#+(other : Int32)
 - Int32#+(other : Int64)
 - Int32#+(other : Int128)
 - Int32#+(other : UInt8)
 - Int32#+(other : UInt16)
 - Int32#+(other : UInt32)
 - Int32#+(other : UInt64)
 - Int32#+(other : UInt128)
 - Int32#+(other : Float32)
 - Int32#+(other : Float64)
 - Number#+()
```

As you can see, the last thing I saw before this change was that "no overload matches", a list of overloads, and just above that the code related to that. With this PR all of these things are scattered.

I also find it a bit strange that code is highlighted (kind of bold), that the error message is yellow, and the rest is normal. The most important thing is the error message, but I don't feel yellow screams more than white (for some reason). Maybe I'm used to the old error messages.

But even for a simpler error... with this code:

```crystal
def fox(a, b)
end

foo 1, 2
``` 

The new error:

```
Error: undefined method 'foo' for top-level

In foo.cr:4:1

 4 | foo 1, 2
     ^--

Did you mean 'fox'?
```

The old error:

```
Error in foo.cr:4: undefined method 'foo' for top-level (did you mean 'fox'?)

foo 1, 2
^~~
```

The new error shows too much info and it's distracting. The old error was simple: just 3 lines, one showing the error message and location, the other one the faulty code.

I don't think verbosity is bad, but if we really want that then I'd like the error messages to be more similar to Elm. Right now it's just a bit of makeup, but with Elm-like error messages it's a complete change. For example for the last code I'd like to see something like this:

```
-- METHOD ERROR -------------------- foo.cr:4:1

I couldn't find a method named 'foo' at the top-level:

4|   foo 1, 2

I did find these methods, though, which have a similar name:

  - fox
```

But I understand that this is a bigger change that's harder to accomplish (plus it involves changing existing messages rather than just changing the formatting).

...

However, if others think this is a good change, we can try it. I just find the error messages more scattered now and harder to process.

asterite:
I edited a bit my previous response, including screenshots.

paulcsmith:
@asterite I think you make some good points. I've got a few thoughts

## Future changes

 I think this the goal of this PR is to improve the formatting for now. Then introduce more elm-like messages. I think it is best to do these in steps though

I think for now if we get this in that would be a small improvement. Then we can really take it up a notch later with more Elm-like error messages

## Colors

I feel the white all blends together. Especially when there are lots of code frames. It's hard to tell where to look. I like yellow, but I think there should be some kind of color or visual indicator that sets the line apart

Could be that we use `-----` like Elm. But I think that may require adjusting the error messages and might be better to tackle at another time. Also worth noting Elm colorizes that part. I believe they use cyan

## The macro message

I think the message you posted with the current implementation is the best case scenario. It gets significantly worse when the macro is defined in a shard and the error in the macro is in the middle of 100+ lines of generated code. It is nearly impossible to see where in the macro the issue is, it is hard to see where the macro is defined, and hard to see what called the macro and where. These are problems that if solved, would greatly improve the experience of using macros in Crystal

I think that part of the problem is how we structure the message, and part of the problem is that there may be a bug. Here is how I believe it should look (with a few tweaks to ordering):

The problem code:

```crystal
def bar
  foo 1, 2
end

{% if true %}
  def foo(x, y)
    1 + 'a'
  end
{% end %}

bar
```

And the error

```
There was a problem expanding macro 'foo'

Code in foo.cr:3:1

 3 | foo 1, 2
      ^
Called macro defined in foo.cr:5:1

 5 | {% if true %}

Which expanded to:

 > 1 |
 > 2 | def foo(x, y)
 > 3 |   1 + 'a'
           ^

Error: no overload matches 'Int32#+' with type Char
Overloads are:
 - Int32#+(other : Int8)
 - Int32#+(other : Int16)
 - Int32#+(other : Int32)
 - Int32#+(other : Int64)
 - Int32#+(other : Int128)
 - Int32#+(other : UInt8)
 - Int32#+(other : UInt16)
 - Int32#+(other : UInt32)
 - Int32#+(other : UInt64)
 - Int32#+(other : UInt128)
 - Int32#+(other : Float32)
 - Int32#+(other : Float64)
 - Number#+()
```

### Suggested changes for macro errors

* Replace message from first line with a more generic: "There was a problem expanding macro 'foo'"
* Move the error message next to the extended description. (in this case "Overloads are:")
* Re-order the display of the macros. Start with the non-macro code (the code that called the macro). Then show where the macro was defined. Then show what it expanded to (with just a few lines above/below). That way the expanded code is by the error. It also is how people think about how their code.
* Also note that the bug is fixed. It says `Code in foo.cr:3:1` and `foo 1, 2`. This is what should be shown. The code that called the macro

> Also worth noting that I believe the PR author @martimatix said passing `--error-trace` will fully expand the macro so you can still get at that when needed.

### How this solves the current issues with macro errors

* Clearly shows how we got to this macro (what code did I write that called this?)
* Shows which macro method was called
* Shows the expanded version but limits the surrounding code so you can actually see where the issue is.

## Non-macro messages

This one I think needs fewer changes. What I love about it is that it is extremely scannable. The filename is highlighted and is near the code line which makes it super easy to see where the problem is. 

I think it also sets us up for Elm-like messages where we can display more info down below the code.

In your example screenshot it is harder to see the did you mean. I believe it should be yellow and then it will be as scannable as before. 

One important note is that in this simple example the error message is easier to see, but in a more realistic example it gets much trickier because the file path is much longer. Which pushes the error messages around. Splitting things up gives a lot more organization (IMO)

Here's an example of how it looks now:

Screen Shot 2019-06-17 at 10 54 29 AM

It's hard to even tell where the error message begins because the path is long and because there is no separation or visual formatting. It also moves the code line far from the path. And if there were more overloads it would be even further. This is why we put the file path right by the printed code in the new version. It is also part of they the extended error description goes below it. It is kind of "extra information" or "help" and so I believe makes sense to be below it. It also makes it ready for Elm-like messages.

### Suggested changes

In this case I think the main thing is highlighting the "Did you mean" in yellow and making sure the extended error messages is well-written. But I think this can be done as a separate PR.

I'd love to hear what people think of this

asterite:
@paulcsmith Good suggestions, thank you!

But what about the issue about the error message being away from the code related to it? Here's another example:

Screen Shot 2019-06-17 at 12 02 10 PM

My eyes first focus on the white highlighted code (I believe it's more prominent than the yellow above, at least for me). Then following that and closer to the cursor (below everything) I see "Overloads are..." which reads kind of strange. What's the error message? Then I have to look back up and see that, oh, the error message is that thing in yellow. But then that message and the list of overloads are pretty far away, and in between I have the code with the actual issue.

What I'd like to see is:

```

In foo.cr:2:2

 2 | a["foo"]
      ^

Error: no overload matches 'Array(Int32)#[]' with type String
Overloads are:
 - Array(T)#[](start : Int, count : Int)
 - Array(T)#[](range : Range)
 - Indexable(T)#[](index : Int)
 ```

That is, I think the most relevant error message should be at the bottom, next to where the cursor is left after I try to compile something.

**Comment by paulcsmith** [comment]
I like the general structure of "here is the error, here's where it is, here's extra details", but I get what you are saying now and I think that is a real issues. It's like your eye is going to "here's where it is, here's the extra details, and then here's the error" which is not the right order. 

Maybe we can do something with the visual to improve this. I've got some ideas I'll try put here so we can try stuff out. Just need to whip up some screenshots

**Comment by paulcsmith** [comment]
How's this for the non-macro error?

Screen Shot 2019-06-17 at 11 22 19 AM

**Comment by martimatix** [comment]
Thanks for taking the time to give feedback on this @asterite and @paulcsmith

I'll start looking at the macro messages first. The only change that I would like to suggest is that `There was a problem expanding macro 'foo'` is moved below the `--error-trace` message. That way the `--error-trace` message consistently appears first whether it's a macro or non-macro error.

As for the non-macro error, I like option 2 because my eyes gravitate to the all yellow error message and then a slight upwards glance takes me to the error source.

**Comment by paulcsmith** [comment]
Totally agree on moving the —error-trace message above for the macro too! 

I’d be totally ok with the yellow error message too. I liked all three options and only had minor preferences between them. Overall I like the new structure a lot. So glad @asterite had those ideas

I think this is gonna look awesome!

**Comment by RX14** [comment]
A key thing to keep in mind is that error messages are read (in top-to-bottom blocks) from the bottom to the top since the bottom is what's shown first in the terminal. That should apply to each error frame as well as the overall trace.

So I like the bottom one most.

**Comment by martimatix** [comment]
### Seeking feedback.

This is what a macro error looks like now:

image

and with `--error-trace`

image

- There's a double line break between error frames.
- The error message is butted up against error source.

Without this, it wasn't clear if the error message referenced the text above or below the message.

**Comment by paulcsmith** [comment]
@martimatix nice! Could you also post the source code so it is easier to see how the error relates to the code?

A few suggestions:

* I think the lines should show the file location. “Code in src/myfile.cr:12” and doesn’t need to say what method or macro it is since it shows the code beneath it. The other messages should also show the file name and line number I think. “Called macro in src/mymacro.cr:12” etc. 
* I think each line should start with a capital letter. 

I’m not totally sure I understand the full error trace but I think that’s partly because I can’t see the source. Once you post that I’ll take another look 🎉

**Comment by asterite** [comment]
@martimatix it's looking great!

**Comment by martimatix** [comment]
@paulcsmith Below is the code for the above error trace. It's a contrived example where a macro is defined inside a macro. When a macro is defined inside a macro, the error can only be referenced by its parent macro. Hence, we see the error referenced by a macro name and not the containing file.

```crystal
macro my_outer_macro(name, content)
  macro define_method
    def {{name}}
      {% if content == 1 %}
        "one"
      {% elsif content == 2 %}
        this_will_cause_an_error
      {% else %}
        {{content}}
      {% end %}
    end
  end

  class InnerMacroClass
    define_method
    def trigger
      bar #=> two
    end
  end
end

my_outer_macro bar, 2
InnerMacroClass.new.trigger
```

Below is an example with the file being referenced by its filename. Also, the first letter of each line is capitalised as per your suggestion.

image

paulcsmith:
That looks fantastic!  🎉

martimatix:
@asterite Fixed specs and build is green. Ready for you to review at your convenience.

asterite:
@martimatix This looks beautiful! 😍 

I found this:

```crystal
Foo.bar
```

Error:

```

In foo.cr:1:1

 1 | foo.bar
     ^--
Error: undefined constant Foo
```

It seems to be showing the line in lowercase for some reason...

Otherwise, this is great!

asterite:
Later today I'll review the code, now I just reviewed what it looks like.

src/compiler/crystal/exception.cr
```diff
@@ -94,4 +99,194 @@ module Crystal
...
+
+    def format_error_from_file(filename : String)
+      lines = File.read_lines(filename)
+      format_error(
+        filename: @filename,
+        lines: lines,
+        line_number: @line_number,
+        column_number: @column_number,
+        size: @size
+      ).capitalize
```
asterite: This seems to be the problem of the downcase...

martimatix: Nice catch! Fixed.

src/compiler/crystal/exception.cr
```diff
@@ -94,4 +99,194 @@ module Crystal
...
+      String.build do |io|
+        io << colorize("#{relative_filename(filename)}:#{line_number}:#{column_number}").underline
+      end
+    end
+
+    def format_error(filename, lines, line_number, column_number, size = 0)
+      return "#{relative_filename(filename)}" unless line_number
+
+      unless line = lines[line_number - 1]?
+        return "#{filename_row_col_message(filename, line_number, column_number)}"
```
Sija:
```suggestion
        return filename_row_col_message(filename, line_number, column_number)
```

paulcsmith:
This is stunning! Nice work @martimatix. I can hardly wait for this to be released :D

Raising at compile time has never been so nice
Screen Shot 2019-06-28 at 11 36 28 AM

src/compiler/crystal/exception.cr
```diff
@@ -94,4 +99,194 @@ module Crystal
       end
     end
   end
+
+  module ErrorFormat
```
asterite: Is this module needed? Can't we just put the methods in `Crystal::Exception`?

martimatix:
I've been using this forked version to build a CLI for work. So far, it's been a great experience.

Just pushed another commit because I discovered that "can't find file" error messages were being printed twice and the font was bold. Wasn't a show stopper but if I didn't correct it now, someone else would have raised an issue.

image

src/compiler/crystal/exception.cr
martimatix: Only two Exception classes utilise these methods: `TypeException` and `SyntaxException`. Therefore, I felt that adding them to `Crystal::Exception` would bloat the instances that don't use these methods.

Happy to move it if that would result in a better outcome.

asterite:
@martimatix Thank you!!

Once CI passes I'm going to merge this. It's so much better than what we have right now.

src/compiler/crystal/exception.cr
asterite: The end result is actually more bloat: methods from modules are typed once for each included type, and also it generates duplicate code. This isn't something that you should or need know... but that's the current state of Crystal and so the less bloat in the compiler's generated code, the better, I think.

In any case, it's probably unnoticeable.

asterite:
@martimatix in #7956 I wanted to have an error that spawns two lines (like "foo\nbar") but what happened is that the first line was shown in yellow, and the second line ended up separated by two lines and in white.

Is that intentional?

How do I make an error spawn multiple lines, where both lines are important for the error message?

martimatix:
Yes, it was intentional as it was described in #7553 

As for how to overcome it, my suggestion is to use `colorize` in the error message itself.

PR #7750, opened 2019-05-07
Compiler: consider macro calls in ivar initializer

### LINKED ISSUE #7666: Macros in type declaration values
When using a macro in a field assignment, I always get the error: ```macro '....' must be defined before this point but is defined later```

```crystal
class Example
  macro val
    "test"
  end

  # Sets up fields for pasing and dumping the struct
  @one : String = val
end
```

Thanks!

asterite:
Fixes #7666


vladfaust:
Ary, thank you!

PR #7756, opened 2019-05-08
Compiler: fix error on named args forwarding

### LINKED ISSUE #7491: Misleading error message when dispatching named arguments through double splat

asterite:
Fixes #7491

### DIFF
--- a/spec/compiler/semantic/named_args_spec.cr
+++ b/spec/compiler/semantic/named_args_spec.cr
@@ -308,4 +308,14 @@ describe "Semantic: named args" do
       ),
       "no overload matches"
   end
+
+  it "says correct error when forwarding named args (#7491)" do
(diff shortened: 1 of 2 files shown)


asterite:
Later I might check what goes on with the linux32 build that it reaches out of memory.

PR #7757, opened 2019-05-08
Compiler: error on macro call in proc pointer

### LINKED ISSUE #7465: The error given when trying to send a macro as a proc could be more helpful
Sample code: 

```crystal
def accepts_block(arg, &block)
  if arg
    return yield
  else
    return "not yielded"
  end
end

macro some_action
  puts "doing a thing"
	"thing done"
end

accepts_block "argument", &->some_action

```

( [playground link](https://carc.in/#/r/6bnw) )

I recognize that you can't take a proc from a macro like that but the resulting error message `wrong number of arguments for macro 'some_action' (given 0, expected 0)` could be more helpful.

asterite:
Fixes #7465

### DIFF
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -916,4 +916,40 @@ describe "Semantic: proc" do
       foo(proc: ->{ 1 })
       )) { proc_of(nil_type) }
   end
+
+  it "errors when using macro as proc value (top-level) (#7465)" do
+    assert_error %(
+      macro bar
+      end
+
+      ->bar
+      ),
+      "undefined method 'bar'\n\n'bar' exists as a macro, but macros can't be used in proc pointers"
+  end
+
+  it "errors when using macro as proc value (top-level with obj) (#7465)" do
+    assert_error %(
+      class Foo
+        macro bar
+        end
+      end
+
+      ->Foo.bar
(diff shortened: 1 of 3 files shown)


spec/compiler/semantic/proc_spec.cr
```diff
@@ -916,4 +916,28 @@ describe "Semantic: proc" do
...
   end
+
+  it "errors when using macro as proc value (top-level) (#7465)" do
+    assert_error %(
+      macro bar
+      end
+
+      ->bar
+      ),
+      "'bar' is a macro and macros can't be used in proc pointers"
```
straight-shoota: Can we make the error message `undefined method 'bar' for top-level` as it would be if `bar` was not defined or a local var? That's the error here: there is no method of that name. Whether there a synonymous macro might just be a minor coincidence.

It doesn't hurt to keep the current message as an additional hint though.

asterite: Yeah, that probably makes sense. I'll change it later.

j8r:
You can now use Draft PRs instead of the `WIP ` text.

asterite:
Can I turn this into a draft?

straight-shoota:
No, you can't go back to draft.

asterite:
Updated!

PR #7758, opened 2019-05-08
Compiler: fix require relative path resolution

### LINKED ISSUE #7408: "require" resolves nonsensical relative paths
In a filesystem structure such as

```
src
|-- foo.cr
spec
|-- spec_helper.cr
|-- foo_spec.cr
```

where `spec_helper.cr` contains:

```crystal
require "spec"
require "../../src/foo" # << This path is wrong and doesn't exist.
# ...
```

The above code does _not_ result in a compile time error but instead requires the file `src/foo.cr`.
This is highly confusing in my opinion and should be a compile time error.

I can't quite catch this behaviour from https://crystal-lang.org/reference/syntax_and_semantics/requiring_files.html. I do believe the code is running into this https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/crystal_path.cr#L71 line and is effectively working because `../src/../src/foo.cr` is found, but still I think this shouldn't resolve, since `require "../../src/foo` from inside `spec_helper` is simply plain wrong.

I am running on macOS:
```
Crystal 0.27.2 (2019-02-05)

LLVM: 6.0.1
```

but obviously the code is still the same in HEAD.

asterite:
Fixes #7408

### DIFF
--- a/src/compiler/crystal/crystal_path.cr
+++ b/src/compiler/crystal/crystal_path.cr
@@ -45,60 +45,69 @@ module Crystal
     end
 
     private def find_in_path_relative_to_dir(filename, relative_to)
-      if relative_to.is_a?(String)
-        # Check if it's a wildcard.
-        if filename.ends_with?("/*") || (recursive = filename.ends_with?("/**"))
-          filename_dir_index = filename.rindex('/').not_nil!
-          filename_dir = filename[0..filename_dir_index]
-          relative_dir = "#{relative_to}/#{filename_dir}"
-          if File.exists?(relative_dir)
-            files = [] of String
-            gather_dir_files(relative_dir, files, recursive)
-            return files
-          end
-        else
-          relative_filename = "#{relative_to}/#{filename}"
+      return unless relative_to.is_a?(String)
 
-          # Check if .cr file exists.
-          relative_filename_cr = relative_filename.ends_with?(".cr") ? relative_filename : "#{relative_filename}.cr"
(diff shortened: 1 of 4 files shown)


asterite:
Cool, I thought I implemented something wrong but it turned out we had a couple of invalid `require` that this PR now catches 💪

src/compiler/crystal/crystal_path.cr
```diff
@@ -45,60 +45,69 @@ module Crystal
...
-          filename_dir = filename[0..filename_dir_index]
-          relative_dir = "#{relative_to}/#{filename_dir}"
-          if File.exists?(relative_dir)
-            files = [] of String
-            gather_dir_files(relative_dir, files, recursive)
-            return files
-          end
-        else
-          relative_filename = "#{relative_to}/#{filename}"
+      return unless relative_to.is_a?(String)
```
Fryguy: I know this was in the original code, but why not have relative_to be typed to a String right in the signature?

asterite: Good point. I will do this, I think it will simplify code (I thought it was being passed other things than `String` or `Nil` but it doesn't seem to be the case). Thanks!

asterite: Done!

asterite: Actually, never mind. This can be done in a separate PR. If somebody wants to do it, please go ahead.

PR #7759, opened 2019-05-09
Compiler: give precedence to T.class over Class

### LINKED ISSUE #7392: Precedence of overloaders with metaclass arguments
### Code example:

_https://carc.in/#/r/672i_

```crystal
def foo(x : Class)
  puts "1"
end

def foo(x : Int32.class)
  puts "2"
end

foo(Int32)
```

### Expected:

```
2
```

### Got:

```
1
```

I'm not sure if this a bug, I may be doing it wrong. It also could be affected by https://github.com/crystal-lang/crystal/pull/7255, please point me if it is.

asterite:
Fixes #7392 

This is just a small thing, but it makes sense, and I'm trying to get rid of every bug we have reported, so...

### DIFF
--- a/spec/compiler/semantic/restrictions_spec.cr
+++ b/spec/compiler/semantic/restrictions_spec.cr
@@ -712,4 +712,18 @@ describe "Restrictions" do
       ),
       "no overload matches"
   end
+
+  it "gives precedence to T.class over Class (#7392)" do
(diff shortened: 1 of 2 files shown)


straight-shoota:
Thank you @asterite ! Your bug killing spree is really amazing 🚀

Issue #7688, opened 2019-04-18
Undefined method 'split' for Path


sudo-nice:
This code fails with `Crystal 0.28.0 [639e4765f] (2019-04-17)`

```crystal
path = Path.new "/tmp/foo/bar"
Dir.mkdir_p path
```

```
Error in r.cr:2: instantiating 'Dir.class#mkdir_p(Path)'

Dir.mkdir_p path
^~~~~~~

in /usr/share/crystal/src/dir.cr:241: instantiating 'mkdir_p(Path, Int32)'

def self.mkdir_p(path, mode = 0o777)
  ^

  in /usr/share/crystal/src/dir.cr:244: undefined method 'split' for Path

components = path.split(File::SEPARATOR)
```

asterite:
`Dir` isn't aware of `Path` yet. It seems only `File` is.

I wonder how can this be solved... imagine I want to define an API (a shard) that works with paths. Should I handle `String | Path`? Anyone will have to handle both of these things all the time? Maybe it's fine...

j8r:
... Why using `Dir` in the first place? Keep using the `Path` object will be real nice, with methods like `path.create_dir(all: true)`, with `all` to create all subdirectories.

asterite:
@j8r I mean, it's probably a good idea, but the comment is confusing. It makes it look like `path.create_dir` is already a thing but it's not. So the only option for OP is to do `Dir.mkdir_p(path.to_s)`

j8r:
What I've understood is the issue is mainly of creating a directory from a `Path`. I think @sudo-nice knows that `Dir.mkdir_p` works with `String` (thus with `to_s`), else he won't bother to do a `Path.new` from a `String` before.
I should be more explicit on the comment. about the proposed draft.

bew:
@asterite I think creating a `Path` should be almost transparent (understand: very lightweight), so that all methods that deals with file paths can only take a `Path` as input. And when you have a `String` you simply create a `Path.new some_str` and give it.

So instead of:
```cr
input = "foo/bar"
Dir.mkdir_p input
```
You'd do:
```cr
input = "foo/bar"
Dir.mkdir_p Path.new(input)
```
This way `Dir` can be properly typed (and can't throw unhelpful error messages like OP' one), and only has to deal with `Path`.

ysbaddaden:
@bew this is exactly what I don't want, and why I don't like `Path`.

[...]

PR #7762, opened 2019-05-10
Deprecate Path#dirname

j8r:
Here is the use case I faced, how to get the directory name?

With either
`File.basename(File.dirname(Dir.current))`
or
`Path.new(Path.new(Dir.current).dirname).basename.to_s`
or a mix of both.

One can also use this, but this isn't as clear nor as efficient:
`Path.new(Dir.current).parts.last.lchop`

Now, one can do:
`Path.new(Dir.current).dirname.basename.to_s`
and the root directory of the current one:
`Path.new(Dir.current).dirname.dirname.basename.to_s`

If a `String` is wanted, `#to_s` can still be called.

---
**EDIT**: There is already `Path#parent`, deprecating `Path#dirname`

### DIFF
--- a/src/path.cr
+++ b/src/path.cr
@@ -174,9 +174,9 @@ struct Path
   # Returns all components of this path except the last one.
   #
   # ```
-  # Path["/foo/bar/file.cr"].dirname # => "/foo/bar"
+  # Path["/foo/bar/file.cr"].dirname # => Path["/foo/bar]"
   # ```
-  def dirname : String
+  def dirname : Path
     reader = Char::Reader.new(at_end: @name)
     separators = self.separators
 
@@ -199,22 +199,22 @@ struct Path
       current = reader.current_char
 
       if separators.includes?(current)
-        return current.to_s
+        return Path.new current.to_s
       else
         # skip windows here for next condition regarding anchor
         if windows? && reader.has_next? && reader.peek_next_char == ':'
           reader.next_char
         else
-          return "."
+          return Path.new "."
         end
       end
     end
 
     if windows? && reader.current_char == ':' && reader.pos == 1 && (anchor = self.anchor)
-      return anchor.to_s
+      return anchor
     end
 
-    @name.byte_slice(0, reader.pos + 1)
+    Path.new @name.byte_slice(0, reader.pos + 1)
   end
 
   # Returns the parent path of this path.
--- a/src/path.cr
+++ b/src/path.cr
@@ -174,9 +174,27 @@ struct Path
   # Returns all components of this path except the last one.
   #
   # ```
-  # Path["/foo/bar/file.cr"].dirname # => Path["/foo/bar]"
+  # Path["/foo/bar/file.cr"].dirname # => "/foo/bar]"
   # ```
+  @[Deprecated("Use `Path#parent.to_s` instead")]
   def dirname : Path
+    parent.to_s
+  end
+
+  # Returns the parent path of this path, all components of this path except the last one.
+  #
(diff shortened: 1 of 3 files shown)


j8r:
I'am confused, what's the difference between `Path#parent` and `Path#dirname`?
`Path.new(Dir.current).parent.basename` seems to do what I'm aiming for.

asterite:
@j8r I guess `dirname` is just `parent.to_s`, in case you need the parent as a string.

I do agree that it might be confusing.

So in your case you can do `Path.new(Dir.current)`.parent.basename`, right?

I'd also like to see `Path.current`, which is simply `Path.new(Dir.current)`.

bew:
> I'd also like to see Path.current, which is simply Path.new(Dir.current)

With `Path.current`it's not clear that it's the current dir (even if it's the only possibility), maybe `Path.current_dir` ?

j8r:
But having `Dir.current` inside `Path` goes against your ideas https://github.com/crystal-lang/crystal/issues/7688 that `Path` is only a wrapper that doens't perform any syscall. This is another topic.

I've added a deprecation for `Path#dirname`, waiting for more thoughts.

bew:
True @j8r, another solution would be to make `Dir.current` return a `Path` directly, so `Path` doesn't have to do the call itself

j8r:
No @bew, because:
- no `File` and `Dir` return `Path`. Some methods are accepting `Path`, and [there are some issues](https://github.com/crystal-lang/crystal/issues/7688)
- this would be a breaking change
- if a new method is added, it can be in `Path`
- this isn't related to the issue, the main point was `Path#dirname`, `Dir.current` was only an example. I could have pick `"/var/log"` instead.

src/path.cr
```diff
@@ -174,9 +174,27 @@ struct Path
   # Returns all components of this path except the last one.
   #
   # ```
-  # Path["/foo/bar/file.cr"].dirname # => "/foo/bar"
+  # Path["/foo/bar/file.cr"].dirname # => "/foo/bar]"
```
```suggestion
  # Path["/foo/bar/file.cr"].dirname # => "/foo/bar"
```

asterite:
I vote to close this. There's nothing wrong with `parent` and `dirname`.

j8r:
I think it's confusing, others can be too like me. Any good reason to have both?

asterite:
One returns a string, which is familiar and known from File.dirname, the other one returns a Path. Why force users to write to_s when they can simply get something with one method call?

j8r:
The user that is familiar with `File.dirname`, can use `File.dirname`?
`parent.to_s` vs `dirname`... for me the result is the same, I was using `Path[mypath.dirname]` because I was familiar with it, and doesn't know an other twin method returning `Path` existed.
I may be alone on this, I admit.

straight-shoota:
Yes, the result of `path.dirname` is equivalent to `path.parent.to_s`. So it's essentially a short cut. One could argue whether it's worth having this method. And this was in fact brought up during the review of #5635. But in the end, it was agreed on the current API. And I'd like to keep it that way - at least for now. The `Path` API is brand new and still needs to be integrated with other APIs. I'd like to put more effort into this than disucussions about whether to have single method which doesn't make a big difference either way.
We can surely revisit the Path API, but please let's gather some some experience in using it first.

`File.dirname` is supposed to go away, so this is not an alternative.

j8r:
Maybe one day all paths would be represented as `Path` instead of `String`.
For the time being, `String` is the most common denominator in the stdlib, I use `Path#to_s` very (too) often.

PR #7764, opened 2019-05-11
Update docs link in README

nsuchy:
This pull request updates the docs link in the README file. The url http://crystal-lang.org/docs redirects to https://crystal-lang.org/reference/ so it makes sense to link to it directly. Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Try it online
 Documentation
 -------------
 
-* [Official docs](http://crystal-lang.org/docs)
+* [Official docs](https://crystal-lang.org/reference/)
 * [Standard library API](https://crystal-lang.org/api)
 * [Roadmap](https://github.com/crystal-lang/crystal/wiki/Roadmap)
 


README.md
```diff
@@ -53,7 +53,7 @@ Try it online
 Documentation
 -------------
 
-* [Official docs](http://crystal-lang.org/docs)
+* [Official docs](https://crystal-lang.org/reference/)
```
r00ster91:
```suggestion
* [Official language reference](https://crystal-lang.org/reference/)
```
There's also still a `https://crystal-lang.org/docs` link above.

Sija:
Personally I find `/docs` URL way more clean, shorter and rememberable.

bcardiff:
Leaving /docs is fine. Eventually a landing page for docs should land there and point the user to either the book, api or other resources. That was the motivation to use /reference.

nsuchy:
Closing this as based on feedback the change isn’t a good idea. Curious about what the landing page would look like.

bcardiff:
Nothing too fancy. A page to link to doc resources and shorten the nav bar a bit. Probably something like https://elm-lang.org/docs

PR #7765, opened 2019-05-11
implement IO::MultiWriter#flush

mamantoha:
Hi.

I'd like to start with an issue I faced while trying to log output to stdout as well as file.

Logging to file doesn't work in the next example:

```crystal
require "logger"

log_file = File.new("test.log", "a")
writer = IO::MultiWriter.new(log_file, STDOUT)

log = Logger.new(writer)
log.level = Logger::DEBUG
log.debug("Created logger")
```
It is fixed by patching `IO::MultiWriter` class.

```crystal
class IO::MultiWriter < IO
  def flush
    @writers.each(&.flush)
  end
end
```

This PR includes fixes and some specs.
Also, I added an example of how to do logging to file and stdout.

### DIFF
--- a/spec/std/io/multi_writer_spec.cr
+++ b/spec/std/io/multi_writer_spec.cr
@@ -1,4 +1,5 @@
 require "spec"
+require "../spec_helper"
 
 describe "IO::MultiWriter" do
   describe "#write" do
@@ -47,6 +48,23 @@ describe "IO::MultiWriter" do
       writer.close
 
       io.closed?.should eq(true)
+    end
+  end
+
+  describe "#flush" do
+    it "writes to IO and File" do
+      with_tempfile("multiple_writer_spec") do |path|
+        file = File.new(path, "w")
+        io = IO::Memory.new
+
+        writer = IO::MultiWriter.new(io, file)
+
+        writer.puts "foo bar"
+        writer.flush
+
+        io.to_s.should eq("foo bar\n")
+        File.read(path).should eq("foo bar\n")
+      end
     end
   end
 end
--- a/src/logger.cr
+++ b/src/logger.cr
@@ -7,9 +7,6 @@
 # When you are developing the system, however, you probably want to know about the program’s internal state,
 # and would set the `Logger` to `DEBUG`.
 #
-# If logging to multiple locations is required, an `IO::MultiWriter` can be
-# used.
-#
 # ### Example
 #
 # ```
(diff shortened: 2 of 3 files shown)


src/logger.cr
```diff
@@ -36,6 +33,19 @@
 #   log.fatal(err)
 # end
 # ```
+#
+# If logging to multiple locations is required, an `IO::MultiWriter` can be
+# used.
+#
+# ```crystal
```
Sija:
```suggestion
# ```
```

spec/std/io/multi_writer_spec.cr
```diff
@@ -49,4 +50,22 @@ describe "IO::MultiWriter" do
       io.closed?.should eq(true)
     end
   end
+
+  describe "#flush" do
+    it "writes to IO and File" do
+      path = datapath("test_file.log")
```
straight-shoota: Better use a `tempfile` here.

mamantoha: Done

spec/std/io/multi_writer_spec.cr
```diff
@@ -49,4 +49,19 @@ describe "IO::MultiWriter" do
       io.closed?.should eq(true)
     end
   end
+
+  describe "#flush" do
+    it "writes to IO and File" do
+      file = File.tempfile("foo")
```
straight-shoota: Sry, I meant the `with_tempfile` helper from `spec/support/tempfile.cr`. It automatically cleans up the tempfile after the spec is run.

mamantoha: done

src/logger.cr
Sija:
```suggestion
# ```
```

spec/std/io/multi_writer_spec.cr
```diff
@@ -49,4 +50,21 @@ describe "IO::MultiWriter" do
       io.closed?.should eq(true)
     end
   end
+
+  describe "#flush" do
+    it "writes to IO and File" do
+      with_tempfile("multiple_writer_spec") do |path|
+        file = File.new(path, "a")
```
straight-shoota: Are you using append mode on purpose? It probably doesn't matter be cause the temp path should not exist, but as far as I understand it, write mode would be more fitting.

mamantoha: done

src/logger.cr
```diff
@@ -36,6 +33,19 @@
...
 # end
 # ```
+#
+# If logging to multiple locations is required, an `IO::MultiWriter` can be
+# used.
+#
+# ```
+# file = File.new("production.log", "a")
+# stdout = STDOUT
+# writer = IO::MultiWriter.new(file, stdout)
```
Sija:
```suggestion
# writer = IO::MultiWriter.new(file, STDOUT)
```

mamantoha: done

asterite:
@mamantoha Thank you!

Issue #7738, opened 2019-05-05
System::User and System::Group API


straight-shoota:
There have been some PR's adding `System::User` and `System::Group` features, originally #5615 then it's reboot #5627 and now there's #7725 with a subset of the features, which is much easier to review.

We've had discussions about the design of these APIs directly in the PRs, but that's not a good procedure, especially when we're talking about quite substantial, completely new APIs.

*Instantiation*

There have been several proposal for getting instances of `System::{Group,User}`. The last state from #5627 was `.get(name_or_id)`, `.from_gid(id)`, `.from_groupname(name)` and equivalent for `User` (plus nilable `?` methods). 

(There were also some methods to retrieve the name from id, or id from name directly (`.name`, `.gid`). I'm not sure we really need them and they can easily be added later, so we don't need to discuss them right away.)

#7725 now comes with `.from_name` and `.from_id` methods for both types (plus nilable `?` methods). This is IMO an improvement, because we don't need `gid`/`uid` and `groupname´/`username` explicitly in the method name, when the fact that it's about a group/user is already determined by the type this method is called on.

A generic method for looking up either by name or id might be useful, but this is a feature that can be added as an enhancement later.

*Properties*

The current proposals are mainly based on exposing properties provided from `getgrnam`, `getpwnam` etc. But we need to take a wider look at exposing a portable API that is not POSIX-specific.

#5627 exposed the following properties:

User:
* name : String
* uid : String
* raw_uid : UInt32 | String
* gid : String
* raw_gid : UInt32 | String
* home : String
* shell : String
* info : String
* group : Group
* root? : Bool

Group:
* name : String
* gid : String
* raw_gid : UInt32 | String
* user_names : Array(String)
* users : Array(User)

In #7725 we have

User:
* name : String
* password : String
* user_id : LibC::UidT
* group_id : LibC::GidT
* directory : String
* shell : String

Group:
* name : String
* password : String
* id : LibC::GidT
* members : Array(String)

There are already a few differences between both of them, and I think the correct solution is somewhere in between.

The interface in Golang is also a nice reference to a very neat API:

[User:](https://golang.org/pkg/os/user/#User)
* Uid string
* Gid string
* Username string
* Name string
* HomeDir string
* GroupIds() string[]

[Group:](https://golang.org/pkg/os/user/#Group)
* Gid  string
* Name string

I propose to use the exact same properties as used in Golang in our API. They are a good collection of essential features working cross-platform. 

[...]

PR #7768, opened 2019-05-11
Better handle file names starting with ~ in Path#expand

byroot:
Currently `p Path.new("~foo.txt").expand` gives: `Path["/home/crystal/oo.txt"]`.

The expected return value could either be `Path["/home/crystal/~foo.txt"]` or `Path["/home/foo.txt"]` depending on wether or not you'd like to support the `~username` notation like Ruby's `File.expand_path` does.

I implemented the former as it's much simpler.

### DIFF
--- a/spec/std/path_spec.cr
+++ b/spec/std/path_spec.cr
@@ -569,6 +569,10 @@ describe Path do
       it_expands_path("~", "/", "\\", base: "/tmp/gumby/ddd", env_home: "/")
       it_expands_path("~/a", "/a", "\\a", base: "/tmp/gumby/ddd", env_home: "/")
     end
+
+    describe "ignores name starting with ~" do
+      it_expands_path("~foo.txt", "/current/~foo.txt", "\\current\\~foo.txt", base: "/current", env_home: "/")
+    end
   end
 
   describe "#<=>" do
--- a/src/path.cr
+++ b/src/path.cr
@@ -564,14 +564,10 @@ struct Path
 
     name = @name
 
-    if name.starts_with?('~')
-      home = home.to_kind(@kind).normalize
-
-      if name.size == 1
-        name = home.to_s
-      else
-        name = home.join(name.byte_slice(2, name.bytesize - 2)).to_s
-      end
+    if name == "~"
+      name = home.to_kind(@kind).normalize.to_s
+    elsif name.starts_with?("~/")
+      name = home.to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
     end
 
     unless new_instance(name).absolute?


straight-shoota:
This is clearly a bug, yes. IIRC I had considered implementing `~username` to expand to `/home/username` (or whatever the home of `username` is), but this is not yet supported because of the additional effort of a cross-platform implementation. But it should be possible.

I guess we could fix this for now. However, that would lead to a breaking change when home directory resolution is implemented.
So maybe it would be better to already cover that case and either return a value that is only a good guess or raise an error that home path expansion cannot be resolved. Or provide a basic implementation right away.

straight-shoota:
We could also decide we don't ever want username-based home resolution. Then it's fine to merge this anyway.

But I think it's a neat feature and typically provided by shells, so it would be great to have it available. Even if it's probably not very commonly used.

byroot:
> I guess we could fix this for now. However, that would lead to a breaking change when home directory resolution is implemented.

Regardless of wether or not this fix is merged, implementing home directory resolution will be a breaking change.

If people are sure they want to implement it, I can try to take a stab at it, but it's far from trivial, there's a lot of subtilities to it. For instance, what if the user doesn't exist? Ruby will raise an error:

```ruby
>> File.expand_path('~foo/bar')
ArgumentError (user foo doesn't exist)
```

This is to say that there's quite a lot of API design consideration in such feature, and beside mimicking Ruby I don't really feel I can make these decisions myself.

straight-shoota:
> Regardless of wether or not this fix is merged, implementing home directory resolution will be a breaking change.

Currently, the `~username` format is broken anyway, so this is not an issue.

> For instance, what if the user doesn't exist?

This should result in a `No such file or directory` error.

Implementation of this feature requires #7738 and then it shouldn't be hard to do:
```cr
elsif name.starts_with?("~")
  username = parts.first.lchop('~')
  user = System::User.from_name?(username)
  raise "No such file or directory" unless user
  name = new_instance(user.home_dir).join(*parts[1..])
end
```

byroot:
> Currently, the `~username` format is broken anyway, so this is not an issue.

Well, based on the method source, I think it's fair to say that it's not broken but simply not implemented. I get that this might seem like splitting hairs, but the distinction is important.

To give more context to this PR, here's what led me to it: https://old.reddit.com/r/crystal_programming/comments/bn9veq/fileexpand_path_with_filename_that_starts_with/

The user on this post didn't expect `File.expand_path("~foo")` to try to resolve a `foo` user home.

straight-shoota:
No, it's broken. The first character after the tilde is chopped off. `Path.new("~foo").expand_path` returns the equivalent to `Path.new("~/oo").expand_path`.

"not implemented" would not change the path at all.

byroot:
@straight-shoota yes it's broken, but the source code doesn't show any kind of attempt at supporting `~username`.

It returns `~/oo` because it check for `start_with?("~")` and then assume the second character is `/`.

So it's I'm just trying to fix an obvious bug. Sure we could think wether we'd like to also implement a feature at the same time, but IMO they are two different things, and this simple fix should make it now, and eventually the feature be added later on it someone feels like it.

straight-shoota:
The current implementation already handles path starting with `~` specially. It does so in a weird way, that needs fixing. But when we fix that, I'd like to fix it with further feature expansion in mind to avoid a breaking change later.

ysbaddaden:
The current implementation _is broken_: it chops the tilde and the following character in cases where it _should not_, since only `~` and `~/path` are supported.

Choosing to handle `~username` is _another matter_. It involves cross platform discrepancies and additional syscalls to get the home of a given user, potentially failing if the user doesn't exist.

If someone is willing to implement the latter with support for all platforms, please do so, it would be a nice to have feature.

In the meantime and otherwise, let's just merge this fix.

PR #7774, opened 2019-05-13
Hide hidden matches in non-recursive directories if `match_hidden` is false.

### LINKED ISSUE #7769: Dir.glob doesn't hide hidden files in non-recursive directories
As per [this reddit thread](https://www.reddit.com/r/crystal_programming/comments/bn8ij8/dirglob_with_match_hidden_somehow_not_working/), calling `Dir.glob` on a non-recursive directory matches hidden files even if `match_hidden` is set to false.

```crystal
# both have the same output
Dir.glob("*", match_hidden: true)
Dir.glob("*", match_hidden: false)
```

[Playground](https://play.crystal-lang.org/#/r/6w66)

Crystal version:

```bash
Crystal 0.28.0 (2019-04-24)

LLVM: 6.0.1
```

---

It looks like this is caused by not checking if the `match_hidden` option is false in the EntryMatch case of Glob.run:

https://github.com/crystal-lang/crystal/blob/639e4765f3f4137f90c5b7da24d8ccb5b0bfec35/src/dir/glob.cr#L166-L170

Adding the line that skips hidden matches in the RecursiveDirectory case before `yield` seems to resolve this issue:

https://github.com/crystal-lang/crystal/blob/639e4765f3f4137f90c5b7da24d8ccb5b0bfec35/src/dir/glob.cr#L222

ayazhafiz:
Fixes #7769

### DIFF
--- a/src/dir/glob.cr
+++ b/src/dir/glob.cr
@@ -166,6 +166,7 @@ class Dir
         when EntryMatch
           return if sequence[pos + 1]?.is_a?(RecursiveDirectories)
           each_child(path) do |entry|
+            next if entry[0] == '.' && !options[:match_hidden]
             yield join(path, entry) if cmd.matches?(entry)
           end
         when DirectoryMatch
--- a/src/dir/glob.cr
+++ b/src/dir/glob.cr
@@ -166,7 +166,7 @@ class Dir
         when EntryMatch
           return if sequence[pos + 1]?.is_a?(RecursiveDirectories)
           each_child(path) do |entry|
-            next if entry[0] == '.' && !options[:match_hidden]
+            next if !options[:match_hidden] && entry.starts_with?('.')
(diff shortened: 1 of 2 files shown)


src/dir/glob.cr
```diff
@@ -166,6 +166,7 @@ class Dir
         when EntryMatch
           return if sequence[pos + 1]?.is_a?(RecursiveDirectories)
           each_child(path) do |entry|
+            next if entry[0] == '.' && !options[:match_hidden]
```
asterite:
```suggestion
            next if !options[:match_hidden] && entry.starts_with?('.')
```

ayazhafiz: Updated both cases this code was present.

bcardiff:
Thanks @ayazhafiz

Issue #7271, opened 2019-01-05
Compiler raises exception for the specific expression of UInt32 enum value


lugia-kun:
For the following Crystal source code, the compiler will raise an exception:
```crystal
module Tests
  enum Test : UInt32
    VAR = (1 << 31)
  end
end

puts Tests::Test::VAR.value
```
and backtrace is here:
```
Using compiled compiler at `.build/crystal'
Invalid UInt32: -2147483648 (ArgumentError)
  from crystal/src/string.cr:0:60 in 'to_u32'
  from crystal/src/compiler/crystal/codegen/codegen.cr:432:17 in 'visit'
  from crystal/src/compiler/crystal/syntax/visitor.cr:27:12 in 'accept'
  from crystal/src/compiler/crystal/codegen/codegen.cr:2062:7 in 'accept'
  from crystal/src/compiler/crystal/codegen/const.cr:73:7 in 'initialize_simple_const'
  from crystal/src/compiler/crystal/codegen/codegen.cr:255:11 in 'initialize_simple_class_vars_and_constants'
  from crystal/src/compiler/crystal/codegen/codegen.cr:229:10 in 'initialize'
  from crystal/src/compiler/crystal/codegen/codegen.cr:145:5 in 'new:single_module:debug'
  from crystal/src/compiler/crystal/codegen/codegen.cr:63:7 in 'codegen'
  from crystal/src/compiler/crystal/codegen/codegen.cr:62:5 in 'codegen:debug:single_module'
  from crystal/src/compiler/crystal/compiler.cr:22:7 in 'codegen'
  from crystal/src/compiler/crystal/compiler.cr:153:16 in 'compile'
  from crystal/src/compiler/crystal/command.cr:258:7 in 'compile'
  from crystal/src/compiler/crystal/command.cr:177:14 in 'run_command'
  from crystal/src/compiler/crystal/command.cr:99:7 in 'run'
  from crystal/src/compiler/crystal/command.cr:46:5 in 'run'
  from crystal/src/compiler/crystal/command.cr:45:3 in 'run'
  from crystal/src/compiler/crystal.cr:8:1 in '__crystal_main'
  from crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from crystal/src/crystal/main.cr:86:7 in 'main'
  from crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from ../sysdeps/x86_64/start.S:122:0 in '_start'
  from ???
[...]

PR #7776, opened 2019-05-13
Compiler: fix enum overflow on non-default values

### LINKED ISSUE #7268: Enum that inherit UInt64 do not get right values
The compiler uses 32 bit integer to keep the values assigned to enum members.
If the Enum is declared with UInt64 as in https://github.com/crystal-lang/crystal/blob/master/src/llvm/enums.cr#L4 the user needs to assign values explicitly.

Affected lines https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/semantic/top_level_visitor.cr#L683-L686

Extracted from #7262

---

$ crystal --version
Crystal 0.27.0

asterite:
Fixes #7268

There's also #7271, which is slightly different, but I'll fix it in a separate PR (it's harder).

### DIFF
--- a/spec/compiler/semantic/enum_spec.cr
+++ b/spec/compiler/semantic/enum_spec.cr
@@ -462,4 +462,14 @@ describe "Semantic: enum" do
       Foo.new(1)
       )) { string }
   end
+
+  it "gives error on enum overflow" do
+    assert_error %(
+      @[Flags]
+      enum Foo : UInt8
+        #{Array.new(9) { |i| "V#{i + 1}" }.join "\n"}
+      end
+      ),
+      "value of enum member V9 would overflow the base type UInt8"
+  end
 end
--- a/spec/compiler/codegen/enum_spec.cr
+++ b/spec/compiler/codegen/enum_spec.cr
@@ -337,4 +337,15 @@ describe "Code gen: enum" do
       Foo.new(1)
       )).to_i.should eq(2)
   end
+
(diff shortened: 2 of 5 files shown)


spec/compiler/semantic/enum_spec.cr
```diff
@@ -462,4 +462,14 @@ describe "Semantic: enum" do
...
   end
+
+  it "gives error on enum overflow" do
+    assert_error %(
+      @[Flags]
+      enum Foo : UInt8
+        #{Array.new(9) { |i| "V#{i + 1}" }.join "\n"}
+      end
+      ),
+      "can't automatically compute value of enum member V9 because it would overflow"
```
straight-shoota: IMO this makes more sense:

```suggestion
      "value of enum member V9 would overflow the base type UInt8"
```

asterite: I'll try to do it tomorrow

asterite: Done, thanks for the suggestion!

asterite:
> I guess you are thinking in interpret the constants based on the enum target type

Yes, that's what we are doing right now. So if you have `1 << 40`, it will work fine if the enum is defined as `UInt64` but it will give an error if it's `Int32`. I think that makes sense, there's no point to have the user type `1_u64 << 40` when they expect to get a `UInt64`.

I thought to fix #7271 I just had to catch `OverflowException` but that doesn't seem to be the case.

Issue #7602, opened 2019-03-28
Can't require installed compiler


bew:
So I've been trying to play with the compiler on `Crystal 0.27.2`:
```cr
# file foo.cr
require "compiler/crystal"

Crystal::Compiler.new
```
Compile with:
```cr
crystal foo.cr -Di_know_what_im_doing
```
And I'm _gently_ welcomed with the following error:
```
cat: /usr/lib/crystal/compiler/crystal/../../../VERSION: No such file or directory
```
(for fun, here is [the full trace](https://gist.github.com/bew/fa20b049d2e67d3a268394739a29287a))

straight-shoota:
Yeah we should probably include the `VERSION` file in the source folder provided by the Crystal package or make reading it optional.

j8r:
How can we simply get the Crystal version, build commit, LLVM version etc. inside Crystal code?

I've tried this, but that's really messy to have to require the whole compiler:
```cr
require "compiler/crystal/**"

Crystal::Config.version
Crystal::Config.build_commit
Crystal::Config.llvm_version
Crystal::Config.default_target
Crystal::Config.date
```

asterite:
https://crystal-lang.org/api/0.28.0/Crystal.html

No need to require anything.

j8r:
That does the job, only missing the `default_target`

asterite:
Yes, because the target is something specific to the compiler, not something you want to use outside of it.

PR #7784, opened 2019-05-15
Make reading VERSION file optional

j8r:
Fix https://github.com/crystal-lang/crystal/issues/7602
Fix https://github.com/crystal-lang/distribution-scripts/issues/35

### DIFF
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -5,7 +5,7 @@ module Crystal
     end
 
     def self.version
-      {{ read_file("#{__DIR__}/../../../VERSION").chomp }}
+      {{ (read_file?("#{__DIR__}/../../../VERSION") || "").chomp }}
     end
 
     def self.llvm_version


src/compiler/crystal/config.cr
```diff
@@ -5,7 +5,7 @@ module Crystal
     end
 
     def self.version
-      {{ read_file("#{__DIR__}/../../../VERSION").chomp }}
+      {{ (read_file?("#{__DIR__}/../../../VERSION") || "").chomp }}
```
Sija: That's incorrect. If there's no `VERSION` file it should return `nil`.

j8r: That's how CRYSTAL_PATH has done

bcardiff:
I would prefer to include the VERSION file in the distribution packages.

j8r:
No @bcardiff , that's the purpose of the file. If we download the crystal source as a tar.gz, we have the version inside.

asterite:
What's wrong with including the `VERSION` file in our distribution?

j8r:
Because not everybody use your distribution scripts, packages like Alpine, Arch Linux and others. Everybody must use the source code from crystal-lang/crystal, some use directly https://github.com/crystal-lang/crystal/archive/0.28.0.tar.gz.

ysbaddaden:
This isn't a good example because the release tarballs do have the `VERSION` file, since they're snapshots of the repository. Do you have examples where the `VERSION` file is _actually_ missing?

j8r:
I was responding to @asterite and @bcardiff @ysbaddaden, of course the `VERSION` file is present. If the file is moved to distribution-scripts, it will no longer be present when packaging for Alpine and Arch Linux, then no version will be available.

For your question about when `VERSION` is missing, look at the two issues in the description (you can ask to @bew about its usecase)

bcardiff:
The `VERSION` file is part of the source. There is no need to make it an optional dependency.
If the file would have been in the ~/src directory it would have worked out of the box.
distribution packages already might already be packaging docs, samples, license, stdlib, etc.

They need to be updated to include the VERSION file to support the scenario of requiring the compiler.

j8r:
I misunderstood. I was thinking that including `VERSION` meant moving it from the repo to distribution scripts.

RX14:
Arch installs the `src` directory to `/usr/lib/crystal`. We shouldn't require any files *outside* the `src` directory to compile what's inside `src` because we cannot assume that they're there.

A simple fix is moving `VERSION` to `src/VERSION`.

PR #7785, opened 2019-05-15
Compiler: fix type flow of var in if with &&

### LINKED ISSUE #7434: Setting more than one local var in "return unless ..." leaves it nilable
Below code should work for both methods (`Foo#bar` and `Foo#baz`) equally I'd say, but for `Foo#baz` it fails with `no overload matches 'Int32#+' with type (Int32 | Nil)` compiler error.

```cr
class Foo
  @x : Int32?
  @y : Int32?

  def initialize(@x, @y)
  end

  def bar
    return unless (x = @x)
    return unless (y = @y) && y > 0
    x + y
  end

  def baz
    return unless (x = @x) && (y = @y) && y > 0
    x + y
  end

  def baz_no_return
    unless (x = @x) && (y = @y) && y > 0
      nil
    else
      x + y
    end
  end

  def baz_if
    if (x = @x) && (y = @y) && y > 0
      x + y
    end
  end
end

# This works
pp Foo.new(2, 2).bar

# This doesn't :( - no overload matches 'Int32#+' with type (Int32 | Nil)
# pp Foo.new(2, 2).baz

# That's ok, `return` breaks stuff...
pp Foo.new(2, 2).baz_no_return

# That's ok too, using `if` seems also fine for the compiler
pp Foo.new(2, 2).baz_if
```

https://carc.in/#/r/6bko

asterite:
Fixes #7434

This one is harder to explain so I'm going to skip that.

### DIFF
--- a/spec/compiler/semantic/if_spec.cr
+++ b/spec/compiler/semantic/if_spec.cr
@@ -337,4 +337,18 @@ describe "Semantic: if" do
       foo
       )) { nilable int32 }
   end
+
+  it "doesn't consider nil type in else branch with if with && (#7434)" do
+    assert_type(%(
+      def foo
+        if true && (y = 1 || nil)
+        else
+          return 1
+        end
+        y
+      end
+
+      foo
+      ), inject_primitives: false) { int32 }
+  end
 end
--- a/src/compiler/crystal/semantic/main_visitor.cr
+++ b/src/compiler/crystal/semantic/main_visitor.cr
@@ -1999,7 +1999,12 @@ module Crystal
         next if then_var.same?(else_var)
 
         if_var = MetaVar.new(name)
-        if_var.nil_if_read = !!(then_var.try(&.nil_if_read?) || else_var.try(&.nil_if_read?))
+
+        # Only copy `nil_if_read` from each branch if it's not unreachable
+        then_var_nil_if_read = !then_unreachable && then_var.try(&.nil_if_read?)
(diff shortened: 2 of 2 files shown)


spec/compiler/semantic/if_spec.cr
```diff
@@ -337,4 +337,18 @@ describe "Semantic: if" do
...
+      def foo
+        if true && (y = 1 || nil)
+        else
+          return 1
+        end
+        y
+      end
+
+      foo
+      ), inject_primitives: false) { int32 }
```
bew: I don't understand your test, in all cases the return type can only be int32, I think you want to `return false` or some other type?

straight-shoota: That's the point: This test ensures that the compiler recognizes that the return type is always `Int32`, because the `if` branch can never be reached.

bew: oooh ok I think I get it, it'd be far easier to understand if we could do:
```cr
if true && (y = 1 || nil)
end
assert_type(y, Int32)
```
Or maybe just do:
```cr
assert_type(%(
  if true && (y = 1 || nil)
  end
  y
  ), inject_primitives: false) { int32 }
```
?

asterite: But in your examples y ends up nilable. It's the return in the else branch that makes it non nilable. But that what was not working well (when in a right side of &&).

PR #7788, opened 2019-05-16
Compiler: raise on new on abstract class

### LINKED ISSUE #3835: Abstract classes can be created via class type
```cr
abstract class Foo
end

a = [] of Foo.class
a << Foo
a[0].new # => #<Foo:0x1102c5fc0 >
```

The above should probably be allowed at compile-time but raise at runtime, because otherwise there would be no way to generically create an instance from a class.

asterite:
Fixes #3835

This one is a weird one. Given:

```crystal
abstract class Foo
end

class Bar < Foo
end

class Baz < Foo
end
```

If you write `Bar.new || Baz.new`, the compiler will unify the type to "`Foo` or any of its subclasses" (`Foo+`). The same happens when you do `Bar || Baz`, it will be `Foo.class+`.

You can do:

```crystal
(Bar || Baz).new
```

and everything's fine. But you could also do:

```crystal
(Foo || Baz).new
```

and that would somehow create a `Foo` instance, even though `Foo` is abstract! The compiler has no way to know that when you end up with `Foo.class+`, the real types might be abstract types.

To solve this, we allow to invoke `new` but when you try to do it with an abstract class it will raise at runtime. This is not type-safe, but if we disallow this then we couldn't implement, for example, a lookup table:

```crystal
lookup = {"Bar" => Bar, "Baz" => Baz} # gets the type Hash(String, Foo.class+) 
lookup["Bar"].new
```

### DIFF
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -1182,6 +1182,20 @@ module Crystal
       false
     end
 
+    protected def define_allocate(metaclass)
+      if metaclass.instance_type.abstract?
+        metaclass.add_def(
+          Def.new("allocate",
+            body: Call.new(nil, "raise",
+              args: [StringLiteral.new("can't instantiate abstract class #{metaclass.instance_type}")] of ASTNode,
+              global: true
+            ))
+        )
+      else
+        metaclass.add_def Def.new("allocate", body: Primitive.new("allocate"))
+      end
+    end
+
     def type_desc
       case
       when extern? && extern_union?
@@ -1200,7 +1214,7 @@ module Crystal
     include ClassVarContainer
 
     protected def initialize_metaclass(metaclass)
-      metaclass.add_def Def.new("allocate", body: Primitive.new("allocate"))
+      define_allocate(metaclass)
     end
 
     def virtual_type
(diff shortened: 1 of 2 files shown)


straight-shoota:
I don't like having a compiler-generated runtime error. But I suppose there is no other way to solve this unless we had a way to differentiate `Foo+`.

Do we have any other incident where the compiler injects a `raise` call? This is one of these examples where the compiler is tightly coupled with the stdlib and expects certain methods to be available at runtime.

src/compiler/crystal/types.cr
```diff
@@ -1182,6 +1182,20 @@ module Crystal
       false
     end
 
+    protected def define_allocate(metaclass)
+      if metaclass.instance_type.abstract?
+        metaclass.add_def(
+          Def.new("allocate",
+            body: Call.new(nil, "raise",
+              args: [StringLiteral.new("can't instantiate abstract class #{metaclass.instance_type}")] of ASTNode,
```
Sija:
```suggestion
              args: [StringLiteral.new("Can't instantiate abstract class #{metaclass.instance_type}")] of ASTNode,
```

IIRC runtime errors are written in _Title case_.

asterite: No, compiler errors are lowercase.

asterite:
Yes, as does raise at the codegen level.

asterite:
I agree that this is a bit rushed, plus it doesn't work.

PR #7790, opened 2019-05-17
Bcrypt: rename `Password#==` to `Password#verify`

### LINKED ISSUE #7753: Crypto::Bcrypt::Password#== is not commutative
As discussed [in the forums](https://forum.crystal-lang.org/t/using-crypto-password-for-authentication/743?u=asterite), it's a bit unintuitive that given `password : Crypto::Bcrypt::Password` and `string : String`, `password == string` will work but `string == password` won't.

Two solutions:
1. Add an overload `String==#(other : Crypto::Bcrypt::Password)` so it's commutative
2. Use a different method name. Ruby's BCrypt uses `is_password?` as an alias, but `verify`, `check` or `matches?` might be equally good.

I'm not sure which approach is better. I think the second one is better because it's also more type-safe. You can do `password == 1.23` and it will still compile, but it's obviously incorrect. You can even do `password == password` and it will return true, but it makes little sense.

Thoughts?

asterite:
Fixes #7753

### DIFF
--- a/src/crypto/bcrypt/password.cr
+++ b/src/crypto/bcrypt/password.cr
@@ -9,8 +9,8 @@ require "../subtle"
 # password = Crypto::Bcrypt::Password.create("super secret", cost: 10)
 # # => $2a$10$rI4xRiuAN2fyiKwynO6PPuorfuoM4L2PVv6hlnVJEmNLjqcibAfHq
 #
-# password == "wrong secret" # => false
-# password == "super secret" # => true
+# password.verify("wrong secret") # => false
+# password.verify("super secret") # => true
 # ```
 #
 # See `Crypto::Bcrypt` for hints to select the cost when generating hashes.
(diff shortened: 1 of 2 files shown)


src/crypto/bcrypt/password.cr
```diff
@@ -61,14 +61,19 @@ class Crypto::Bcrypt::Password
...
+  # password.verify("wrong secret") # => false
+  # password.verify("super secret") # => true
   # ```
-  def ==(password : String) : Bool
+  def verify(password : String) : Bool
     hashed_password = Bcrypt.new(password, salt, cost)
     Crypto::Subtle.constant_time_compare(@raw_hash, hashed_password)
   end
 
+  @[Deprecated("Use Crypto::Bcrypt::Password#verify")]
```
bcardiff:
```suggestion
  @[Deprecated("Use `Crypto::Bcrypt::Password#verify`")]
```

oprypin:
Maybe not breaking-change if it's just deprecated

bcardiff:
@oprypin I know, but we agree that at least for the time being we are tagged them as breaking for better visibility in the release notes.

Sija:
I'd still suggest to rename the method to `verify?` to make it obvious it's returning a `Bool`.

asterite:
@Sija But `verify?` isn't proper English, same reason we have `includes?` and not `include?`. I think `verify` is fine. I would have liked `matches?` but others agreed with using `verify` or `verify?` (but `verify?` is inconsistent to me).

PR #7797, opened 2019-05-19
remove realpath from bin/crystal

wontruefree:
I had to a fair amount of debugging and found that readlink with the --canonicalize flag is the same on OSX BSD and Linux.  This seems to simplify this script.

### DIFF
--- a/bin/crystal
+++ b/bin/crystal
@@ -24,10 +24,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-realpath() {
-    canonicalize_path "$(resolve_symlinks "$1")"
-}
-
 resolve_symlinks() {
     _resolve_symlinks "$1"
 }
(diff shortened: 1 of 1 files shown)


bin/crystal
```diff
@@ -133,7 +110,7 @@ __warning_msg() {
 }
 
 
-SCRIPT_PATH="$(realpath "$0")"
+SCRIPT_PATH="$(readlink --canonicalize "$0")"
```
j8r: No present on Alpine Linux, use `f` instead.

```
Usage: readlink [-fnv] FILE

Display the value of a symlink

        -f      Canonicalize by following all symlinks
        -n      Don't add newline
        -v      Verbose
```

j8r:
There is also [crystal-wrapper](https://github.com/crystal-lang/distribution-scripts/blob/master/linux/files/crystal-wrapper). If we really want to simplify this mess, [we can do it in Crystal](https://github.com/crystal-lang/crystal/issues/7391). For some unclear reasons, some are attached to this scripts.

bin/crystal
ysbaddaden: It's not Alpine: only `-f` is supported on OpenBSD and FreeBSD, for example, and seems to be somewhat portable (no guarantees, thought).

Thought `-f` doesn't seem to be supported on macOS. In fact I'm not sure `--canonicalize` is available until you install GNU coreutils...

I.e. the current script is portable, that's why it's here in the first place.

jhass:
Is there any real pain here? Is it broken somewhere? Why touch this?

wontruefree:
I am going to close this because of the of the cross platform problems.

@jhass I had to do some debugging and this caused me some frustration.  But now that I have learned a few things I will not run into the same problems.

PR #7801, opened 2019-05-20
 Compiler: don't precompute sizeof on abstract structs and modules

### LINKED ISSUE #7741: Invalid memory access (signal 11) at address 0x0
Extracted from https://github.com/crystal-lang/crystal/issues/7728

https://play.crystal-lang.org/#/r/6u14/edit

```cr
abstract struct ColumnBase; end

record ColumnInfo(T, M) < ColumnBase, column_name : String, default : T? = nil, type : T.class = T do
  # Returns the SQL representation for this column.
  def to_column(io : IO) : Nil
  end
end

cols = [
  ColumnInfo(String, Int32).new("name", "Jim"),
  ColumnInfo(Int32, Int32).new("age", 123),
  ColumnInfo(Array(Bool), Int32).new("values", [true, false, false, true, true]),
  ColumnInfo(Int64, Int32).new("id", nil),
]

io = IO::Memory.new

cols.sort_by { |c| true ? -1 : 1 }.each &.to_column io
ColumnInfo(Time, Int32).new("created_at", nil).to_column io
```

```
[0x55af537d7006] *CallStack::print_backtrace:Int32 +118
[0x55af537c9e90] __crystal_sigfault_handler +192
[0x7f82cf1b93c0] ???
[0x0] ???
```

Same error against crystal master.

I was able to reduce it a bit, however the stack trace changed a bit.

https://play.crystal-lang.org/#/r/6ujl

```cr
abstract struct ColumnBase; end

record ColumnInfo(T) < ColumnBase, column_name : String, default : T? = nil

cols = [
  ColumnInfo(String).new("name"),
  ColumnInfo(Int32).new("age"),
  ColumnInfo(Array(Bool)).new("values"),
  ColumnInfo(Int32).new("id"),
]

io = IO::Memory.new

cols.sort_by { |c| true ? -1 : 1 }.each &.to_s io
ColumnInfo(Time?).new("created_at").to_s io
```

```
[0x563058616686] *CallStack::print_backtrace:Int32 +118
[0x563058608bc0] __crystal_sigfault_handler +192
[0x7f76ae0be3c0] ???
[0x56305861f926] *Slice(T)::new:read_only<Pointer(UInt8), Int32, Bool>:Slice(UInt8) +22
[0x563058619215] *String#to_slice:Slice(UInt8) +37
[0x56305861ca38] *String#to_s<IO::Memory>:Nil +24
[0x563058662961] *IO::Memory +17
[0x563058668bc0] *ColumnBase+ +32
[0x0] ???
```

asterite:
Fixes #7741

And a few other fixes/improvements too.

I'm marking it as a breaking change too because `instance_sizeof` could be incorrectly invoked with non-class stuff (like structs) and so this will make compiling (but incorrect) code to stop compiling... but I think it's fine.

### DIFF
--- a/src/compiler/crystal/semantic/main_visitor.cr
+++ b/src/compiler/crystal/semantic/main_visitor.cr
@@ -2568,7 +2568,8 @@ module Crystal
       # Try to resolve the sizeof right now to a number literal
       # (useful for sizeof inside as a generic type argument, but also
       # to make it easier for LLVM to optimize things)
-      if type && !node.exp.is_a?(TypeOf)
+      if type && !node.exp.is_a?(TypeOf) &&
+         !(type.module? || (type.abstract? && type.struct?))
         expanded = NumberLiteral.new(@program.size_of(type.sizeof_type).to_s, :i32)
         expanded.type = @program.int32
         node.expanded = expanded
--- a/spec/compiler/codegen/sizeof_spec.cr
+++ b/spec/compiler/codegen/sizeof_spec.cr
@@ -199,4 +199,40 @@ describe "Code gen: sizeof" do
       size.should eq(16)
     end
   {% end %}
+
+  it "doesn't precompute sizeof of abstract struct (#7741)" do
+    run(%(
+      abstract struct Base
+      end
+
+      struct Foo(T) < Base
+        def initialize(@x : T)
+        end
+      end
+
+      z = sizeof(Base)
+
+      Foo({Int32, Int32, Int32, Int32})
+
+      z)).to_i.should eq(16)
+  end
+
+  it "doesn't precompute sizeof of module (#7741)" do
+    run(%(
+      module Base
+      end
+
+      struct Foo(T)
+        include Base
+
+        def initialize(@x : T)
+        end
+      end
+
+      z = sizeof(Base)
+
+      Foo({Int32, Int32, Int32, Int32})
+
+      z)).to_i.should eq(16)
+  end
 end
--- a/spec/compiler/semantic/sizeof_spec.cr
+++ b/spec/compiler/semantic/sizeof_spec.cr
@@ -22,7 +22,10 @@ describe "Semantic: sizeof" do
   end
 
   it "gives error if using instance_sizeof on something that's not a class" do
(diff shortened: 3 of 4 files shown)


asterite:
Some explanation: at some point we made `sizeof` be computed, if possible, at the semantic phase so that it can be used for generic type arguments, like `StaticArray(UInt8, sizeof(Int32))`. That works fine, but when the type given to `sizeof` is an abstract thing like a struct or a module, and there are generic subtypes, the final size (which is the size of the union of all actual instantiations in the program) can only be known at the end of compilation. This is what this PR fixes.

src/compiler/crystal/semantic/main_visitor.cr
```diff
@@ -2568,7 +2568,8 @@ module Crystal
       # Try to resolve the sizeof right now to a number literal
       # (useful for sizeof inside as a generic type argument, but also
       # to make it easier for LLVM to optimize things)
-      if type && !node.exp.is_a?(TypeOf)
+      if type && !node.exp.is_a?(TypeOf) &&
+         !(type.module? || (type.abstract? && type.struct?))
```
bcardiff: Can't this be moved upwards together with the check of `type.is_a?(GenericType)`? I thought the notion of an instantiable type was already defined somewhere.

Blacksmoke16:
Can confirm this fixed my issue.  Thanks again.

src/compiler/crystal/semantic/main_visitor.cr
asterite: Here in `MainVisitor` we compute it if we can. Later in `CleanupTransformer` we check that it's only a class. So, doing `sizeof` an abstract struct or module can be done and works, just not eagerly.

To be honest, now I'm not convinced that `sizeof` should work eagerly at all. I did that so one can do, for example, `StaticArray(Int32, sizeof(Int32))`, but that's a pretty contrived example. But for now I'm just fixing broken things, not changing the semantics of the language or introducing breaking changes. But eventually we can define the semantics we want for this and fix it.

And no, unfortunately there's no `type.can_be_instantiated?` method.

PR #7804, opened 2019-05-21
Move VERSION inside ./src

bcardiff:
Move VERSION inside ./src due to https://github.com/crystal-lang/crystal/pull/7784#issuecomment-494387548 

Fixes https://github.com/crystal-lang/distribution-scripts/issues/35
Fixes https://github.com/crystal-lang/crystal/issues/7602
Closes https://github.com/crystal-lang/crystal/pull/7784


bin/ci
```diff
@@ -98,7 +98,7 @@ prepare_build() {
 
 verify_version() {
   # If building a tag, check it matches with file
-  FILE_VERSION=`cat VERSION`
+  FILE_VERSION=`cat ./src/VERSION`
```
j8r:
```suggestion
  FILE_VERSION="$(cat ./src/VERSION)"
```

asterite: Why?

asterite: I mean, it was using backticks before and it was working fine. Why risk changing it to something that apparently works the same but might break in some other subtle ways?

oprypin: Even if it is a positive change, I would keep it separate from this PR as unrelated.

bcardiff: Ok, commit removed.

RX14:
:heart:

j8r:
Any reason why the version isn't directly a constant inside the code?

asterite:
I was thinking the same thing. In fact, everything was good before introducing this file.

However, I think having it in a separate file makes the commit history for other files a bit cleaner.

oprypin:
But why not have it in a *separate* *source file*? Isn't it best of both worlds?

asterite:
Yes, I think that makes sense. I think (but I don't know) the idea of the VERSION file was also to have it in the root directory so it's visible, but we even lost that. At this point I think it should be in code (except that it's a bit harder to read externally).

j8r:
It can have a file with only `Crystal::VERSION = "0.28.0"`, then require it.
Easy to parse for other tools outside Crystal, too.

PR #7811, opened 2019-05-22
Fix record macro with inheritance

asterite:
This didn't work:

```crystal
abstract struct Base; end

record Point < Base, x = 0, y = 0

Point.new(1, 2).clone
```

But with this PR it does.


oprypin:
Could add a code sample as a test that this works

PR #7812, opened 2019-05-22
Compiler: fix extend from generic types

### LINKED ISSUE #7167: Extending a module doesn't work for generics
This works fine (https://carc.in/#/r/5qsl):

```crystal
module Foo
  def foo
  end
end

class String
  extend Foo
end

pp! String.foo # => nil
```

But this fails (https://carc.in/#/r/5qsm):

```crystal
module Foo
  def foo
  end
end

class Array(T)
  extend Foo
end

pp! Array(Int32).foo
# undefined method 'foo' for Array(Int32).class
```

And also this (https://carc.in/#/r/5qsn):

```crystal
module Foo(T)
  def foo
  end
end

class Array(T)
  extend Foo(T)
end

pp! Array(Int32).foo
# undefined method 'foo' for Array(Int32).class
```

asterite:
Fixes #7167

### DIFF
--- a/spec/compiler/semantic/module_spec.cr
+++ b/spec/compiler/semantic/module_spec.cr
@@ -1304,4 +1304,52 @@ describe "Semantic: module" do
       ),
       "no overload matches"
   end
+
+  it "extends module from generic class and calls class method (#7167)" do
+    assert_type(%(
+      module Foo
+        def foo
+          1
+        end
+      end
+
+      class Gen(T)
+        extend Foo
+      end
+
+      Gen(Int32).foo
+      )) { int32 }
+  end
+
+  it "extends generic module from generic class and calls class method (#7167)" do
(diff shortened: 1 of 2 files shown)


wontruefree:
I don't know if this belongs in this PR but why would parents return a `nil`.  Seems like if all classes return an array of `Types` and you call `parents` on a class will no parents you should get an empty array.

asterite:
@wontruefree It's a good observation. I think I did it like that to avoid creating an empty array. But it probably makes sense to return an empty array in the default case because that will never be called, except for Object. I'll try it later (in a separate PR).

asterite:
Actually a module has no parents by default. I don't think I'll make this change for now.

wontruefree:
sounds good.  I was just making an observation.

PR #7815, opened 2019-05-23
Compiler: don't virtualize types in Union(...)

### LINKED ISSUE #7814: JSON serialization issue when inheriting from Array
If you create a class that inherits from an Array, and you override the `initialize` method, regardless of that class being used, it will throw errors with serialization.

```crystal
class Collection < Array(String)
  def initialize
    super
  end
end

require "json"

class Foo
  include JSON::Serializable

  property bar : Array(String)
end

Foo.from_json(%({"bar": ["a"]}))
```
https://play.crystal-lang.org/#/r/6yl2

In this case, I only care about `Foo` because `Collection` may come from some shard my app is requiring ([example](https://github.com/luckyframework/wordsmith/blob/master/src/wordsmith/inflector/inflections.cr#L6)). 

This code will throw 

> \> 35.                     ::Union(Array(String)).new(pull)
> wrong number of arguments for 'Array(String)#initialize' (given 1, expected 2, 0)

```
Crystal 0.28.0 (2019-04-24)

LLVM: 6.0.1
```

asterite:
Fixes #7814

### DIFF
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -2868,7 +2868,7 @@ module Crystal
           type_var.raise "argument to Proc must be a type, not #{type_var}"
         end
         # There's no need for types to be virtual because at the end
-        # `type_merge` will take of that.
+        # `type_merge` will take care of that.
         # The benefit is that if one writes `Union(T)`, that becomes exactly T
         # and not T+ (which might lead to some inconsistencies).
         type_var.devirtualize.as(Type)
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/types.cr
```diff
@@ -2867,7 +2867,11 @@ module Crystal
         unless type_var.is_a?(Type)
           type_var.raise "argument to Proc must be a type, not #{type_var}"
         end
-        type_var
+        # There's no need for types to be virtual because at the end
+        # `type_merge` will take of that.
```
bew:
```suggestion
        # `type_merge` will take care of that.
```

PR #7818, opened 2019-05-23
Compiler: fix new/initialize lookup regarding modules

### LINKED ISSUE #7007: Including JSON::Serializable or YAML::Serializable lets you invoke a default constructor
https://play.crystal-lang.org/#/r/5dii

```Crystal
require "json"

class Foo
  include JSON::Serializable

  property name : String
end

pp Foo.new
```

```

Invalid memory access (signal 11) at address 0x4
[0x55679ffc5476] *CallStack::print_backtrace:Int32 +118
[0x55679ffb9b0d] __crystal_sigfault_handler +61
[0x7f56342143c0] ???
[0x55679ffc9e60] *String#chomp:String +352
[0x55679ffcb6f3] *String#pretty_print<PrettyPrint>:Nil +275
[0x5567a000496b] *Foo +1419
[0x5567a0004c95] *PrettyPrint::format<Foo, IO::FileDescriptor, Int32, String, Int32>:IO::FileDescriptor +117
[0x5567a0004c13] *PrettyPrint::format<Foo, IO::FileDescriptor, Int32>:IO::FileDescriptor +67
[0x55679ffb9b60] *pp<Foo>:Foo +32
[0x55679ffacaab] __crystal_main +1515
[0x5567a000b016] *Crystal::main_user_code<Int32, Pointer(Pointer(UInt8))>:Nil +6
[0x5567a000af79] *Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32 +41
[0x55679ffb7a66] main +6
[0x7f5633de1223] __libc_start_main +243
[0x55679ffac3ee] _start +46
[0x0] ???
```

asterite:
Fixes #7007

It turned out the issue was that even this compiled:

```crystal
module Base
  def initialize(x)
  end
end

class Foo
  include Base
end

Foo.new # Works, finds then one in Reference, but shouldn't!
```

The problem was that `new`/`initialize` lookup looked past modules that defined an `initialize` method and this was wrong.

### DIFF
--- a/spec/compiler/semantic/module_spec.cr
+++ b/spec/compiler/semantic/module_spec.cr
@@ -1352,4 +1352,20 @@ describe "Semantic: module" do
       Gen(Int32).foo
       )) { int32.metaclass }
   end
+
+  it "doesn't look up initialize past module that defines initialize (#7007)" do
+    assert_error %(
+      module Moo
+        def initialize(x)
+        end
+      end
+
+      class Foo
+        include Moo
+      end
+
+      Foo.new
+      ),
+      "wrong number of arguments"
+  end
 end
--- a/src/compiler/crystal/semantic/call_error.cr
+++ b/src/compiler/crystal/semantic/call_error.cr
@@ -61,10 +61,10 @@ class Crystal::Call
       end
     end
 
-    # Another special case: initialize is only looked up one level,
+    # Another special case: `new` and `initialize` are only looked up one level,
     # so we must find the first one defined.
     new_owner = owner
-    while defs.empty? && def_name == "initialize"
+    while defs.empty? && (def_name == "initialize" || def_name == "new")
       new_owner = new_owner.superclass
       if new_owner
         defs = new_owner.lookup_defs(def_name)
--- a/src/compiler/crystal/semantic/method_lookup.cr
+++ b/src/compiler/crystal/semantic/method_lookup.cr
@@ -36,7 +36,7 @@ module Crystal
       # `new` must only be searched in ancestors if this type itself doesn't define
       # an `initialize` or `self.new` method. This was already computed in `new.cr`
       # and can be known by invoking `lookup_new_in_ancestors?`
-      if my_parents && !(!lookup_new_in_ancestors? && is_new)
+      if my_parents && !(is_new && !lookup_new_in_ancestors?)
         my_parents.each do |parent|
           matches = parent.lookup_matches(signature, owner, parent, matches_array)
           if matches.cover_all?
@@ -44,6 +44,11 @@ module Crystal
           else
             matches_array = matches.matches
           end
+
+          # If this is a `new` method, once a parent defines an `initialize`
+          # method and we couldn't find any matches we must not go up in the
+          # hierarchy.
+          break if is_new && parent.has_def_without_parents?(signature.name)
         end
       end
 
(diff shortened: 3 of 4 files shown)


bew:
Does that mean that this won't work intentionally?

```cr
module Base1
  def initialize(x)
  end
end

module Base2
  def initialize(x, y)
  end
end

class Foo
  include Base1
  include Base2
end

Foo.new 1
Foo.new 1, 2
```
(not sure which one would fail based on the order of parents, the one in Base1 or Base2?)

asterite:
@bew Yes, that's correct. You can't have `initialize` defined at multiple levels. This is probably a breaking change, but it simply doesn't make sense to allow `initialize` bypass a level, because then you are potentially bypassing initializing an instance variable.

asterite:
Maybe later we can refine this rule, but for now I think this is what makes most sense. I really wouldn't advice using `initialize` in a module anyway.

asterite:
Put another way, I coded the compiler to always check the first level of `initialize` to determine whether all instance variables are initialized. That an `initialize` on a later level can be called is a bug and it wasn't intentional at all.

asterite:
You can always do:

```
module Base1
  def initialize(x)
  end
end

module Base2
  def initialize(x, y)
  end
end

class Foo
  include Base1
  include Base2

  def initialize(x)
    super
  end
end

Foo.new 1
Foo.new 1, 2
```

(in fact `JSON::Serializable` does something similar, it injects a `self.new` on `inherited`)

asterite:
> User is able to declare multiple initialize across multiple modules and include them.

This is not like that. If a type defines an initialize method (any arity) then initialize in super types (parent classes or included modules) are not looked up. Think of Java or C#: constructors are never inherited. The exception in Crystal is that if a type doesn't define a constructor, parent constructors are looked up.

bcardiff:
I didn't mention inheritance, but I should have. Ok with that semantics.

I was trying to state what will happen after this PR with:

```crystal
module Base1
  def initialize(x)
  end
end

module Base2
  def initialize(x, y)
  end
end

class Foo
  include Base1
  include Base2
end
```

Foo will have both initialize or only one?

asterite:
@bcardiff Only the one from `Base2`.

Maybe we can change that... but if we do that we should formally define how we want `initialize`/`new` to work in all cases, not just in this particular case, and how to check that all non-nilable instance vars are initialized for all including classes.

bcardiff:
Ok, the last context for now then. I see how this can be understood as consistent with not inheriting initializer from the parent from some point of view.

Let's merge this as is for now.

It's not 100% neat that `super` is used to point to some initializers of a module. But is ok for now I think.

Blacksmoke16:
@asterite @bcardiff I'm assuming it's known that this code would no longer work come next release?  Seems undesired IMO.

```crystal
require "json"

abstract class Parent
  @name : String

  def initialize(@name : String); end
end

class Child < Parent
  include JSON::Serializable
end

Child.new "Fred"
```
```
Error in test.cr:39: no overload matches 'Child.new' with type String
Overloads are:
 - Child.new(pull : ::JSON::PullParser)

Child.new "Fred"
```

asterite:
Yes, constructors are not (and shouldn't) be inherited.

That said, I always wanted JSON serialization to be decoupled from constructors, but nobody else liked that idea, so...

Blacksmoke16:
@asterite Why shouldn't constructors be inherited from abstract parent classes/structs?  Like currently, anything that has constructors shared in a parent abstract class/struct that includes a serializable modules, just doesn't work.

asterite:
Because a parent constructor has no guarantee of initializing instance variables of a subclass.

PR #7820, opened 2019-05-24
forward read_/write_timeout in ssl socket to underlaying socket

carlhoerberg:


src/openssl/ssl/socket.cr
```diff
@@ -189,4 +189,15 @@ abstract class OpenSSL::SSL::Socket < IO
     io = @bio.io
     io.responds_to?(:remote_address) ? io.remote_address : nil
   end
+
+  {% for name in ["read_timeout", "read_timeout=", "write_timeout", "write_timeout="] %}
```
asterite: Could you please define each method separately, without using macros? Not only because it's simpler, but only because this is wrong: `read_timeout=` expects an argument, same as `write_timeout=`. Thank you!

src/openssl/ssl/socket.cr
```diff
@@ -189,4 +189,40 @@ abstract class OpenSSL::SSL::Socket < IO
...
+      io.read_timeout = value
+    else
+      raise NotImplementedError.new("#{io.class}#read_timeout=")
+    end
+  end
+
+  def write_timeout
+    io = @bio.io
+    if io.responds_to? :write_timeout
+      io.read_timeout
```
```suggestion
      io.write_timeout
```

straight-shoota:
I'm a bit skeptical of this. Adding a bunch of `NotImplementedErrors` errors for this doesn't seem right.

First of all, what's the intended use case?
Maybe there could be other (more type-safe) solutions?

If we stay with the suggested methods, we could consider moving them to a more generic location than `SSL::Socket` (maybe even `IO::Buffered`).

carlhoerberg:
So in our server software we listen both for plain text and TLS connections. We then hand over the socket (`TCPSocket`, or `OpenSSL::SSL::Socket`) down the line and treat them all the same. In parts of the protocol a read_timeout has be set, and sometimes we're just waiting endlessly (timeout = nil) for data on the socket. And instead of both having to pass the both raw TCPSocket AND the OpenSSL::SSL::Socket in the case of TLS connections down the line, we currently monkey patch the SSL::Socket with the above code.

carlhoerberg:
I don't see how this could be implemented in `IO::Buffered`, but if `BIO#io` was a `Socket` instead of a `IO` here: https://github.com/crystal-lang/crystal/blob/3817a99db9b1a9bb2fa007a91ef0c857d80d097f/src/openssl/bio.cr#L115
we could avoid `responds_to?`.

carlhoerberg:
So i guess my question is, can `OpenSSL::BIO` ever be passed an `IO` object that isn't a `Socket`?

asterite:
It seems so: https://github.com/crystal-lang/crystal/blob/73518893dff6592287257c78574dcbe9d0b32903/src/openssl/x509/extension.cr#L60-L66

carlhoerberg:
ah yeah, so what do you think? better for ppl to pass around both the underlaying socket and the openssl::socket, or to raise NotImplemented if you try to set read_timeout on an openssl::socket where you've used a IO::Memory as underlying socket?

asterite:
I have no problem with this PR and I think it's fine. There's really no other way to do this in Crystal without proper interfaces.

asterite:
Or maybe we can define these methods to `IO` with a default `NotImplemented` and then make `IO::Buffered` forward those methods to the underlying IO. That way there's no `responds_to?`.

ysbaddaden:
@asterite an in-memory buffer with read/write timeouts sounds plausible to me, so :+1: for this transparent and more generic solution.

asterite:
I was actually thinking that in-memory buffer would never timeout. So for IO::Memory it would be a nop.

carlhoerberg:
But how would `IO::Buffered` forward to the underlying socket? As the underlying socket is `@bio.io` in `OpenSSL::SSL::Socket`, but in most other cases in `@io`? Or an abstract method in `IO::Buffered`, so that all classes that extends `IO::Buffered` has to implement `read_timeout` et al?

carlhoerberg:
`IO::Buffered` feels misplaced too, timeouts doesn't have much to do with buffering either I feel? Is it just that SSL::Socket happens to extend that module?

straight-shoota:
Yes, `SSL::Socket` can wrap a generic `IO`. And that makes sense. None of it's behaviour is strictly bound to a socket, it can work on anything (wrapping a socket is certainly the most common application, though).

This change essentially merges unrelated behaviour into this class, which only belongs to `Socket` but not to generic IOs. I guess `local_address` and `remote_address` are already precedents for this. They're not in the inherent scope of `SSL::Socket` either.

Having these specific delegators on `SSL::Socket` only helps as far as it directly wraps a `Socket`. Should there be any other wrapping, for example compression, the entire thing breaks. 

Promoting this to the generic IO interface makes more sense, I guess.

An alternative would be to introduce a method to access the wrapped IO, which would allow to navigate to the source socket and call methods on it directly. However, this would already be complicated for `IO::MultiWriter` and `IO::Stapled` which wrap multiple IOs.

asterite:
Oh, I thought `IO::Buffered` wrapped another `IO` but that's not the case. Sorry, disregard what I said before.

Again, I see no problem with this PR. One usually works with a socket, and the socket can be a raw socket `Socket` or a secure socket `OpenSSL::SSL::Socket`. You want to work with those things uniformly. Then both types should have more or less the same methods.

What I don't understand is that when you have something that can either be a `Socket` or `OpenSSL::SSL::Socket`, the compiler will infer its type to be `IO`. Then you can't do `io.read_timeout = ...` because it's not defined on `IO`. So you have to check if it's Socket or ssl Socket. @carlhoerberg is that what you are doing?

If so, I think `read_timeout` and others should be in `IO`. Then you can work uniformly with any `IO`. Same reason we have `flush` on every `IO` even though not every `IO` can be flushed.

At least that's how Crystal works. Other languages have interfaces so you would have `IO::Waitable` or something and you could check that.

In any case, it makes sense to put a timeout on an ssl socket, and have that work properly without having to deal with the underlying IO.

carlhoerberg:
Absolutely, putting `read_timeout` et al in `IO` and in `SSL::Socket` forward it to `@bio.io` could work. Do you think the default implementation should be noop like `flush` or `raise NotImplemented`?

straight-shoota:
@asterite I suppose `SSL::Socket` is actually a bad name, because this type is not a socket. It's just an IO wrapper (similar to `Gzip::Reader` for example) and has nothing to do with a socket.

In Crystal's IO design there is no *raw socket* and *secure socket*. There is only `Socket`, and such a socket can be wrapped in `SSL::Socket` for providing encryption.

We should consider this when re-designing a generic `TLS` API, this type should not be be called "socket" but maybe `TLS::IO` or similar.

straight-shoota:
I think the core issue here is that `Socket` represents two separate concerns: connection configuration (for example timeouts, addresses) and the stream interface (`IO` methods). `SSL::Socket` only wraps the latter (on purpose because because it's supposed to works with any `IO`).

So `SSL::Socket` can't fully replace `Socket` in a handler that needs direct access to the underlying socket configuration.

Moving the respective methods to `IO` mitigates that. And for `{read,write}_timeout` it might still make some kind of sense. But `{remote,local}_address` really shouldn't belong to `IO`, that's entirely specific to `Socket`.

IMO it would be better to pass both `Socket` and `SSL::Socket` to handlers, where the former is responsible for connection configuration and the latter for handling the stream IO.
This would allow users access to even more specific socket configuration, for example the `tcp_keepalive_*` options.

A practical implementation could simply add `socket` as a parameter to `HTTP::RequestProcessor#process` and property to `HTTP::Context`. In both cases it can be optional to allow for non-socket application.

carlhoerberg:
or give access to the underlying `IO` through the `SSLSocket`?  eg. `@ssl_socket.io.read_timeout = 15`

carlhoerberg:
it's annoying though in the handler code to distinguish if it's a SSL or a "normal" socket, i just want to read and write to it. and in some steps set the read_timeout as it's part of the protocol. passing around `(socket : Socket, ssl_socket : SSLSocket?)` is super ugly.

asterite:
I think we can merge this, right? It's harmless and it works fine.

PR #7822, opened 2019-05-27
Add POSIX (get|set)(uid|gid) functionality.

didactic-drunk:
* Adds `Process.uid`

* Adds `Process.euid`

* Adds `Process.gid`

* Adds `Process.egid`

* Adds `Process.setresuid`

* Adds `Process.setresgid`


straight-shoota:
Thank you for this PR @didactic-drunk!

There has already been another proposal for this in #5627 (among other features). But this huge PR should be split in smaller ones, and some parts of it are already extracted.
#7725 introduces the `System::User` and `System::Group` API. Having a separate PR for the process API is great.

And the implementation looks pretty neat so far. However, we probably need to take a step back to first discuss the intended API in more detail. Your implementation pretty closely resembles the underlying POSIX system calls. We probably want to have a higher level of abstraction and a platform-independent API that can work for non-POSIX systems like Windows.

I've opened a new issue for this: #7829

src/process.cr
```diff
@@ -491,6 +491,116 @@ class Process
     io.close if io
   end
 
+  # Returns the **real** user id of the current process.
+  def self.uid : LibC::UidT
```
asterite: LibC types shouldn't be exposed in public Crystal APIs. We should either always convert stuff to Int64, Int32 or whatever is appropriate.

In fact, in Go this is exposed as a String. Maybe we should even return a struct representing this where the internal representation is hidden from the user (then we can use LibC::Uidt internally), because you won't be performing math or anything else with a uid.

didactic-drunk:
@ysbaddaden This PR doesn't handle users and groups.  It handles the uid getting/switching as a companion to `Process.user` which returns `System::User`.  One of the many process PR's #7829 #5627 #5615 (I think there's one more) said a single PR was too large and to break it up.

This  PR provides the underlying implementation for `Process.user` to obtain the user_id and also provide the actual user_id's for those interested.

Some applications may only need the euid/uid, not the user records. Also it should work in a chroot environment where /etc/passwd is unavailable.  An API that returns only the id is still necessary.  I'm attempting to use these features right now which is why this PR is implemented the way it is taking in to consideration the common cases with a high level API returning `System::User` and providing access to the actual id's for applications where it's a necessity.

There are also security implications in always returning a user when attempting to fetch the id.
```
def Process.root?
  # Opens "/etc/master.passwd" on BSD systems
  # potentially leaving root's or other hashed passwords in memory.
  Process.user.user_id == 0
end
```

vs 

```
def Process.root?
  # Call getuid and skip getpwnam
  Process.user_id == 0
end
```


Are `Process.user_id`, `Process.effective_user_id` acceptable names?

src/process.cr
didactic-drunk: Sometimes users id's need to be displayed, saved, restored, copied or compared.    For example applications that provide persistent data owned by a particular user_id not stored in a filesystem.  Or sending user_id credentials via a socket where you can't send a name.  You can't reify it to a name on the sender or receiver as the process sending or receiving may be chroot'd and not have access to the names.

There is simple math on uid's via ==.  If euid == 0 or uid == 0 privileges are elevated.  Also comparisons are sometimes done against gid == egid to detect setgid programs. 

Anyone attempting to create a user add function in pure crystal needs math operations to find the highest uid that's not nobody.

Before you criticize the user creation example which I admit I have no use for, the socket example is on my list of things to do, checking setuid/gid is needed for the app I'm porting, sockets and uid maintained data are also part of this app.  I haven't submitted a PR for socket as that seems quite distant in getting approved.  

My point being various applications need the actual id's, not a abstract return type with hidden values. There's a reason that [go, nim, node, python, ruby] along with various other programming languages include getuid functions by default.

asterite: The Go API used strings.

didactic-drunk:
@ysbaddaden Names changed to `Process.user_id`, `Process.effective_user_id`, etc.

src/process.cr
didactic-drunk: How do I safely convert from `"0"` to `LibC::UidT`?  Is there an `IntX.from_s`?

didactic-drunk: I was almost finished with converting to String's with the exception of the above problem when I decided to read the documentation.

The documentation claims it's an `int`. [https://golang.org/pkg/os/#Getegid](https://golang.org/pkg/os/#Getegid)
```
func Geteuid() int
Geteuid returns the numeric effective user id of the caller.
On Windows, it returns -1. 
```

Actual code: [https://golang.org/src/os/proc.go](https://golang.org/src/os/proc.go)
```
// Geteuid returns the numeric effective user id of the caller.

// On Windows, it returns -1.

func Geteuid() int { return syscall.Geteuid() }
```

Also windows doesn't have a concept of effective vs real vs saved user or group id's.

Personally I'd prefer to keep these as Int's as that's what everyone using them expects them to be and if someone does what they've been taught in every unix programming book to check for root by using `if euid == 0` it will actually work instead of creating security holes.

didactic-drunk:
> If we're introducing `Process.user = User.new(id)` then maybe we don't want `Process.user_id = `?

1. Does that work with `chroot`?
2. Does it allow for avoiding getpwent calls for performance or security reasons?
```
# If this uses getpwent:
Process.user = User.new(Process.real_or_effective_user_id)
# Is not the same as:
Process.user_id = Process.real_or_effective_user_id
```

didactic-drunk:
On second thought Process#become is probably a better than `Process#user_id=` for one way transitions.
```
def become(user, group = user.group)
  setresgid group.gid if group
  setresuid user.id
end
```

Is the above API acceptable?

didactic-drunk:
@chris-huxtable proposed the following:

`Process.become(user, ...)` handles complete user/group/groups transitions in the correct order.

Rename `Process.setresuid` to `Process.become_user(euser, effective: user)`
Rename `Process.setreguid` to `Process.become_group(euser, effective: user)`
`become_x` is still necessary for temporary user/group changes or changing ruid/euid/suid to different values.

Would these changes be accepted?

didactic-drunk:
Bump?

didactic-drunk:
Is there anything else for me to do?  @straight-shoota thumbs up'd the API.

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
     $?.exit_code.should eq(0)
   end
 
+  it "getuid" do
```
Sija:
```suggestion
  it ".effective_user_id" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+    Process.effective_user_id.should eq `id -u`.to_i
+    Process.effective_user_id.should eq Process.effective_user_id
+  end
+
+  it "getgid" do
```
Sija:
```suggestion
  it ".group_id" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+    Process.group_id.should eq `id -g`.to_i
+    Process.group_id.should eq Process.effective_group_id
+  end
+
+  it "become_user" do
```
Sija:
```suggestion
  it ".become_user" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+    end
+
+    if euid == 0
+      success.should be_true
+    else
+      success.should be_false
+    end
+  end
+
+  it "become_group" do
```
Sija:
```suggestion
  it ".become_group" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+    end
+
+    if Process.effective_user_id == 0
+      success.should be_true
+    else
+      success.should be_false
+    end
+  end
+
+  it "become_user raises when only setting suid on unsupported platforms" do
```
Sija:
```suggestion
  it ".become_user raises when only setting suid on unsupported platforms" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+    if !LibC.responds_to?(:setresuid) && LibC.responds_to?(:setreuid)
+      expect_raises_errno(Errno::ENOSYS) do
+        Process.become_user(suid: 55)
+      end
+    end
+  end
+
+  it "become_group raises when only setting sgid on unsupported platforms" do
```
Sija:
```suggestion
  it ".become_group raises when only setting sgid on unsupported platforms" do
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+      Process.user_id.should eq 0
+      Process.effective_user_id.should eq 8888
+      success = true
+    rescue ex : Errno
+      # Ok for non-root
+    ensure
+      Process.become_user(ruid: ruid, euid: euid) if success
+    end
+
+    if euid == 0
```
straight-shoota: This spec is hard to follow. Why not streamline flow like this?

```cr
if euid == 0
    Process.become_user(ruid: 0, euid: 8888, suid: 9999)
    Process.user_id.should eq 0
    Process.effective_user_id.should eq 8888
    Process.become_user(ruid: ruid, euid: euid)
else
  expect_raises_errno(Errno::EPERM) do
    Process.become_user(ruid: 0, euid: 8888, suid: 9999)
  end
end
```

Or maybe even better, separate both cases into individual examples and only execute them according to the value of `euid`. This clearly communicates what is actually being tested.
```cr
if euid == 0
  it ".become_user (root)" do
    # ...
  end
else
  it ".become_user (non-root)" do
    # ....
  end
end
```

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
...
+
+    if Process.effective_user_id == 0
+      success.should be_true
+    else
+      success.should be_false
+    end
+  end
+
+  it "become_user raises when only setting suid on unsupported platforms" do
+    if !LibC.responds_to?(:setresuid) && LibC.responds_to?(:setreuid)
```
straight-shoota: I'd move this conditional outside the example, so that it doesn't show up as successful when it actually doesn't do anything.

src/process.cr
```diff
@@ -507,6 +507,150 @@ class Process
...
+  # user = System::User.find_by name: "crystal"
+  # Process.become user
+  #
+  # user = System::User.find_by name: "crystal"
+  # group = System::Group.find_by name: "wheel"
+  # Process.become user, group # Use a different group other than the user's.
+  # ```
+  def self.become(user : System::User, group : System::Group? = nil)
+    group ||= user.group
+    # TODO setgroups
```
straight-shoota: Could you elaborate on this TODO? I don't understand what it means.

spec/std/process_spec.cr
```diff
@@ -85,6 +85,79 @@ describe Process do
     $?.exit_code.should eq(0)
   end
 
+  it "getuid" do
+    Process.effective_user_id.should eq `id -u`.to_i
+    Process.effective_user_id.should eq Process.effective_user_id
```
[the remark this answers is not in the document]
didactic-drunk: It should test `user_id` against `effective_user_id`.

src/process.cr
didactic-drunk: It means [setgroups](https://linux.die.net/man/2/setgroups) should be called before `become_user`, but the `User` class doesn't provide a group list.

Also, the API and naming was never discussed.  I would use `Process.groups=` before `become_` was settled on.   Now I don't know what to call it.

spec/std/process_spec.cr
```diff
@@ -85,37 +85,32 @@ describe Process do
...
     Process.group_id.should eq `id -g`.to_i
     Process.group_id.should eq Process.effective_group_id
   end
 
-  it ".become_user" do
-    success = false
-    ruid = Process.user_id
-    euid = Process.effective_user_id
+  if euid == 0
+    it ".become_user" do
```
straight-shoota: The essential part was adding the scopes `as root` and `as non-root` to the example descriptions. Otherwise there is really no difference to having a single example.

src/process.cr
straight-shoota: All right. This extensive explanation should be added to the comment.

src/process.cr
```diff
@@ -513,6 +513,150 @@ class Process
...
+  # * *suid*: saved user id.
+  #
+  # Example:
+  #
+  # ```
+  # Process.become_user ruid: 1004, euid: 0 # Changes real and saved user id to 1004 and effective to 0 (root)
+  # Process.become_user euid: 0             # Only changes the euid.
+  # Process.become_user suid: 0             # Platform specific and may not function.
+  # ```
+  def self.become_user(*, ruid : LibC::UidT = UID_DEFAULT, euid : LibC::UidT = UID_DEFAULT, suid : LibC::UidT = ruid)
```
straight-shoota: I propose to name these arguments simply `real`, `effective` and `saved` instead of cryptic abbreviations that require an explanation.

src/process.cr
```diff
@@ -513,6 +513,150 @@ class Process
     io.close if io
   end
 
+  # Returns the **real** user id of the current process.
```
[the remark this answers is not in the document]
didactic-drunk: # Returns the **real** user id of the current process.
# Returns the **effective** user id of the current process.

When skimming the documentation to quickly see the method differences.  If the use of bold is inappropriate I'll remove it.

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+  # Process.become_group saved: 0             # Platform specific and may not function.
+  # ```
+  def self.become_group(*, real : Int? = nil, effective : Int? = nil, saved : Int? = nil)
+    real ||= GID_DEFAULT
+    effective ||= GID_DEFAULT
+    saved ||= real
+
+    {% if LibC.has_method?(:setresaved) %}
+      if LibC.setresgid(real, effective, saved) != 0
+        raise Errno.new("setresaved failed")
```
```suggestion
        raise Errno.new("setresgid failed")
```

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+      end
+    {% else %}
+      if saved != GID_DEFAULT && real == GID_DEFAULT && effective == GID_DEFAULT
+        Errno.value = Errno::ENOSYS
+        raise Errno.new("only setting the saved gid not supported")
+      end
+
+      if LibC.setregid(real, effective) != 0
+        raise Errno.new("setreffective failed")
```
```suggestion
        raise Errno.new("setregid failed")
```

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+        raise Errno.new("setresaved failed")
+      end
+    {% else %}
+      if saved != UID_DEFAULT && real == UID_DEFAULT && effective == UID_DEFAULT
+        Errno.value = Errno::ENOSYS
+        raise Errno.new("only setting the saved uid not supported")
+      end
+
+      if LibC.setreuid(real, effective) != 0
+        raise Errno.new("setreffective failed")
```
```suggestion
        raise Errno.new("setreuid failed")
```

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+  # Process.become_user saved: 0                 # Platform specific and may not function.
+  # ```
+  def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int? = nil)
+    real ||= UID_DEFAULT
+    effective ||= UID_DEFAULT
+    saved ||= real
+
+    {% if LibC.has_method?(:setresuid) %}
+      if LibC.setresuid(real, effective, saved) != 0
+        raise Errno.new("setresaved failed")
```
```suggestion
        raise Errno.new("setresuid failed")
```

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+  # Process.become_group real 1, effective: 2 # Changes real and saved group id to 1, effective to 2.
+  # Process.become_group effective: 0         # Only changes the effective.
+  # Process.become_group saved: 0             # Platform specific and may not function.
+  # ```
+  def self.become_group(*, real : Int? = nil, effective : Int? = nil, saved : Int? = nil)
+    real ||= GID_DEFAULT
+    effective ||= GID_DEFAULT
+    saved ||= real
+
+    {% if LibC.has_method?(:setresaved) %}
```
```suggestion
    {% if LibC.has_method?(:setresgid) %}
```

src/process.cr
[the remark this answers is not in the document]
straight-shoota: Given that the other method is already called `effective_user_id`, I don't think that's necessary.

src/process.cr
```diff
@@ -513,6 +513,156 @@ class Process
...
+  # * *saved*: saved user id (suid).
+  #
+  # Example:
+  #
+  # ```
+  # Process.become_user real: 1004, effective: 0 # Changes real and saved user id to 1004 and effective to 0 (root)
+  # Process.become_user effective: 0             # Only changes the effective.
+  # Process.become_user saved: 0                 # Platform specific and may not function.
+  # ```
+  def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int? = nil)
```
straight-shoota:
```suggestion
  def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int = real)
```
`saved` cannot stay unchanged, so it should not accept `nil` which would have different semantics than for the other arguments. Assigning `real` as default value seems logical because that's the default.

src/process.cr
```diff
@@ -513,6 +513,150 @@ class Process
...
+    LibC.getgid
+  end
+
+  # Returns the **effective** group id of the current process.
+  def self.effective_group_id : LibC::GidT
+    LibC.getegid
+  end
+
+  # :nodoc:
+  UID_DEFAULT = if LibC::UidT.new(0).is_a?(Int::Signed)
```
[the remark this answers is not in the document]
didactic-drunk: `UID_DONT_CHANGE`?  Do you have a suggestion?

src/process.cr
chris-huxtable: Sorry to dig this up from the grave but I think it deserves to be said.

>> Personally I'd prefer to keep these as Int's as that's what everyone using them expects them to be

I think @didactic-drunk is correct. Anyone using this is expecting integers. Using strings just adds a more trouble to its use for the end user. More trouble means more screw ups more screw-ups mean more bugs. Seeing as these would be during runtime thats even more troublesome. Ideally this is what you try to avoid troublesome runtime things with frameworks/libraries. If it isn't supported on a OS the API shouldn't be available on that OS, throwing a compile time error. Requiring the user to support OS specific features on any given OS is to be expected.

src/process.cr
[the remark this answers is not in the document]
straight-shoota: Maybe `UID_NO_CHANGE`?

src/process.cr
didactic-drunk: What happens when setting `effective` without changing `real`?  Requiring `saved` would break that. 

It's a common case. There's a dedicated syscall for it (`seteuid`).  Some services operate in the following manner:
```
while request = ...
  seteuid request.user.id
  open file as user
ensure
  seteuid 0
end
```


Most programs will do one of several things:
* Set them all once when switching away from root.  This is probably the most common case and why Process.become(user) exists.
* Swap euid/ruid back and forth before lowering privileges (setuid or root starting applications).  Example `sudo`.
* Set euid to many users possibly swapping euid/ruid (root running service or setuid).
* I probably missed a few.

Setting `saved` by itself is rare but I wanted to allow it on platforms that support it for the cases when setting `ruid => other, euid => other, suid => current` and switching back later.

Leaving `saved` as the only way back can add additional security since I think it's dropped on exec under some circumstances on some platforms.   Or maybe all platforms.  My memory is foggy on this point as I haven't tried it in > 20 years and don't care to look it up.

src/process.cr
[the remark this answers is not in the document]
didactic-drunk: It's private now if that makes a difference.

src/process.cr
straight-shoota: Oh yeah, I overlooked that case. But `saved = real || effective` should work, right?

straight-shoota: I just don't want its default value to be `nil` and `nil` having a different meaning than for effective and real.

src/process.cr
didactic-drunk: Maybe the difference in opinions is one of trying to create portable API (@asterite) and trying to use features we know aren't portable like `setuid`/`chroot` and would like the program to break at compile time if it can't handle the operation.

When I want a Windows `SID`, I'll ask for one and not expect it to work on POSIX.  Likewise I won't bother with `getgid` on Windows. 

100% portable applications are a nice goal, but I'm not making 100% portable applications.  I'm making POSIX applications that expect to use POSIX calls for security.  Otherwise I can't use crystal.

I could use shards but...

Maintaining shards is hard for these types of calls as the definitions vary between platforms and crystal doesn't include a c header parser.  Luckily the bindings were generated under `src/lib_c`, but using them means half the calls are in base crystal and the shard monkey patches to get access.  That seems like they could get out of sync quickly and be a maintenance nightmare.

My preference is break at compile and document as "Nonportable.  Use X if you can".

src/process.cr
didactic-drunk: I think I based this on the behavior of using `setreuid`.  Providing `ruid` or `euid` leaves `saved` as whatever `ruid` is, which sometimes means not changing it.  This is probably the most portable option as some platforms don't support `setresuid` and the most secure.

didactic-drunk: Maybe i should document that it attempts to mimic the behavior of setreuid if `saved` is not provided.

didactic-drunk:
Review?

src/process.cr
```diff
@@ -513,6 +513,158 @@ class Process
...
+  # Process.become_user 0 # Changes real, effective, saved user id to 0 (root).
+  # ```
+  def self.become_user(uid : Int)
+    become_user real: uid, effective: uid, saved: uid
+  end
+
+  # Attempts to change real, effective, and/or saved user id's of the current process.
+  # Uid's not supplied in arguments are kept at their current values (if supported).
+  # Explicit setting of saved id's is not supported on all platforms.
+  # Attempts to mimic the behavior of setreuid() if saved is not provided.
```
straight-shoota:
```suggestion
  # Attempts to mimic the behavior of `setreuid()` if *saved* is not provided.
```
Maybe add a link to documentation of `setreuid`?

src/process.cr
straight-shoota: I'm still not fond of the fact that `nil` argument has a different meaning for real/effective vs. saved

didactic-drunk: If I changed this
`def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int? = nil)`
to this:
`def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int? = real)`

Would that be acceptable?

straight-shoota: This still means `nil` has the same meaning when passed as argument value like `become_user(real: 1, effective: nil, saved: nil)`.
/cc @ysbaddaden @asterite @RX14 Any input on this?

didactic-drunk: What about two methods?  One that has `setreuid` semantics and another with `setresuid`?
```
def self.become_user(*, real : Int? = nil, effective : Int? = nil)
def self.become_user(*, real : Int? = nil, effective : Int? = nil, saved : Int = nil)
```

straight-shoota: Yeah, that could work. But it would be `saved : Int = real`, right?

didactic-drunk: Is `become_user` acceptable now?

src/process.cr
```diff
@@ -513,6 +513,183 @@ class Process
...
+  def self.group_id
+    LibC.getgid.to_i
+  end
+
+  # Returns the effective group id of the current process.
+  def self.effective_group_id
+    LibC.getegid.to_i
+  end
+
+  private UID_NO_CHANGE = if LibC::UidT.new(0).is_a?(Int::Signed)
```
RX14: Could this be just `LibC::UidT.new!(-1)`? Which then wraps to `MAX`.

RX14:
How's this going to work on windows when user and group IDs become strings and most of these are unsupported...

Perhaps this is platform-specific enough to just be a shard... Which I, reluctantly suggest.

didactic-drunk:
> How's this going to work on windows when user and group IDs become strings and most of these are unsupported...
> 
> Perhaps this is platform-specific enough to just be a shard... Which I, reluctantly suggest.

Some systems have `setresuid` and other only `setreuid`.  Crystal doesn't have a way for shards to access lib feature detection but the base crystal does have it's own platform detection which is also unavailable to shards.

How would I differentiate between `setreuid` and `setresuid` in a shard?

If there was c header support this would be a lot easier.

didactic-drunk:
> How's this going to work on windows when user and group IDs become strings and most of these are unsupported...

Changing user's and groups is not supported but cygwin manages to do it somehow.


> Perhaps this is platform-specific enough to just be a shard... Which I, reluctantly suggest.

Does shards have a way to indicate platform specific shards?

```
dependencies:
  posix:
    github: ...
    platform: posix
    platform: !windows
```

Also, `chroot` can't be used safely without `setuid` functionality.  Why isn't it in a shard?

ysbaddaden:
This can easily a shard (e.g. named `posix`).

- There are zero differences in the `uid_t` and `gid_t` types in all supported platforms, they're all `unsigned int`. The functions are POSIX compliant and also identical across supported platforms. A single LibC definition will cover them all —if it needed to be cross-platform, it would be a great use case to push #8307 forwards;
- Windows support is likely to be different and we don't know about it (or care actually);
- We don't need user/group management to build the Crystal runtime (unlike Signal);

Instead of debating forever on whether we include this or not, it would be available *today* to developers, instead of arguing about function names and how to interact with an unknown target (i.e. Windows), it would merely wrap the well known and documented POSIX methods and it would be perfect. We'll resume debating *much later*, when we have a better knowledge of the Windows API, and whether we provide we'll provide this feature, or not.

Of course, we'd move `chroot` to that shard, too.

ysbaddaden:
> How would I differentiate between `setreuid and `setresuid` in a shard?

Just like we do in crystal/system: with target compile flags; e.g. `:unix`, `:darwin`, `:linux`, ...

straight-shoota:
Golang provides `Getegid()`, `Geteuid()`, `Getgid()`, `Getgroups()`, `Getuid()` in stdlib package `os`.
They return `-1` on windows.

The respective setters are available with other system specific features in https://golang.org/x/sys

I think we could go a similar way and provide getters in stdlib. They only provide information and might have broader use cases than going for a decidedly platform-specific implementation.

asterite:
For types that might be int in some platforms or strings in others, the best thing to do would be to retruenf an opaque wrapper struct type that you pass around. That type could provide to_i and to_s methods if needed.

didactic-drunk:
> Instead of debating forever on whether we include this or not, it would be available _today_ to developers, instead of arguing about function names and how to interact with an unknown target (i.e. Windows),

I'm starting to think that's the only way forward considering:
* It's been 2 years of discussions and 2-3 implementations with no clear way to get this in to core.
* Those with domain knowledge want compile time errors if `geteuid` and the like are used on windows while core members want runtime errors.  I still don't understand the reasoning as it doesn't provide portability, only the illusion of it.  
* We also want native types for interaction with other systems and API's.
* The windows `SID` and `(get|set)euid` are used for different purposes.  I don't see why you'd mix the two in any code.  Requiring platform agnostic return values doesn't make sense to me.

RX14:
The question for me is is this worth our time to make sure this is *100%* correct, and will *never* have to be changed again after 1.0, when so few people need this functionality.

I understand the usecase, it's important, but ultimately not that popular. We can always move this into the core later, after it's a shard.

For now I would much prefer to start with what go provides, since Go's API design is the most sane and cross-platform I have *ever* seen. (For reference, they also make sure stdlib code compiles on all platforms, and returns errors at runtime when unsupported)

chris-huxtable:
For any one else interested! 

- https://github.com/chris-huxtable/user_group.cr

PR's welcome.

RX14:
I think it'd be nice to have a `crystal-lang/posix` shard which was somewhat officially supported, but not with the same stability guarantees as the stdlib proper.

chris-huxtable:
@RX14 - I (and I am sure many others) would happily contribute the `crystal-lang/posix`.

watzon:
Yeah I'd be interested in helping with `crystal-lang/posix` as well seeing as I use Linux pretty exclusively

PR #7827, opened 2019-05-28
Compiler: fully track AST nodes name location

### LINKED ISSUE #6937: Error for undefined method after do, end points to the wrong line
```cr
def method
  yield
end

method do
end.undefined_method
```
gives this error with the green pointer at the wrong line:
```
Error in spec/spec_helper.cr:5: undefined method 'undefined_method' for Nil

method do
    ^~~~~~~~~~~~~~~~

Rerun with --error-trace to show a complete error trace.
```
but a one-line bracket block
```cr
def method
  yield
end

method {}.undefined_method
```
gives a correct error:
```
Error in spec/spec_helper.cr:5: undefined method 'undefined_method' for Nil

method {}.undefined_method
          ^~~~~~~~~~~~~~~~

Rerun with --error-trace to show a complete error trace.
```

asterite:
Fixes #6937

Contains a minor refactor too.

### DIFF
--- a/src/compiler/crystal/syntax/ast.cr
+++ b/src/compiler/crystal/syntax/ast.cr
@@ -55,8 +55,8 @@ module Crystal
     def doc=(doc)
     end
 
-    def name_column_number
-      @location.try(&.column_number) || 0
+    def name_location
+      nil
     end
 
     def name_size
@@ -538,7 +538,7 @@ module Crystal
     property block : Block?
     property block_arg : ASTNode?
     property named_args : Array(NamedArgument)?
-    property name_column_number : Int32
+    property name_location : Location?
     property name_size = -1
     property doc : String?
     property visibility = Visibility::Public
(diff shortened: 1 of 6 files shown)


src/compiler/crystal/syntax/ast.cr
```diff
@@ -1026,7 +1023,9 @@ module Crystal
     end
 
     def clone_without_location
-      Macro.new(@name, @args.clone, @body.clone, @block_arg.clone, @splat_index, @double_splat.clone)
+      m = Macro.new(@name, @args.clone, @body.clone, @block_arg.clone, @splat_index, @double_splat.clone)
```
Sija: Possible pedantic refactor:

```cr
Macro.new(@name, @args.clone, @body.clone, @block_arg.clone, @splat_index, @double_splat.clone)
  .tap(&.name_location = name_location)
```

asterite: I really don't care much about shorter code anymore, but thanks!

PR #7831, opened 2019-05-29
Add type restrictions to INI

j8r:
Having type restrictions helps to know which type is expected when looking at the API docs, but still supporting exotic types with Unions like `NamedTuple` with `Int32` and `Char` values.

### DIFF
--- a/spec/std/ini_spec.cr
+++ b/spec/std/ini_spec.cr
@@ -54,22 +54,8 @@ describe "INI" do
       it "parses empty section" do
         INI.parse("[section]").should eq({"section" => Hash(String, String).new})
       end
-
-      it "parses a file" do
-        INI.parse(File.read datapath("test_file.ini")).should eq({
-          "general" => {
-            "log_level" => "D",
-          },
-          "section1" => {
-            "foo" => "1.1",
-            "bar" => "2",
-          },
-          "section2" => {
-            "x.y.z" => "coco lala",
-          },
-        })
-      end
     end
+
     describe "from IO" do
       it "parses a file" do
         File.open datapath("test_file.ini") do |file|
--- a/spec/std/ini_spec.cr
+++ b/spec/std/ini_spec.cr
@@ -2,71 +2,91 @@ require "./spec_helper"
 require "ini"
 
 describe "INI" do
-  describe "parse from string" do
-    it "fails on malformed section" do
-      expect_raises(INI::ParseException, "unterminated section") do
-        INI.parse("[section")
-      end
-    end
-
-    it "fails on data after section" do
-      expect_raises(INI::ParseException, "data after section") do
-        INI.parse("[section] foo  ")
-      end
-    end
-
-    it "fails on malformed declaration" do
-      expect_raises(INI::ParseException, "expected declaration") do
-        INI.parse("foobar")
+  describe "parse" do
+    describe "from String" do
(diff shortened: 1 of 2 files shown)


src/ini.cr
```diff
@@ -21,7 +21,7 @@ class INI
   #
   # INI.parse("[foo]\na = 1") # => {"foo" => {"a" => "1"}}
   # ```
-  def self.parse(str) : Hash(String, Hash(String, String))
+  def self.parse(str : String) : Hash(String, Hash(String, String))
```
bew: I think it could be an IO too

j8r: I was wondering too, if the same method can be used.
I've added a test.

spec/std/ini_spec.cr
```diff
@@ -2,71 +2,77 @@ require "./spec_helper"
 require "ini"
 
 describe "INI" do
-  describe "parse from string" do
-    it "fails on malformed section" do
-      expect_raises(INI::ParseException, "unterminated section") do
-        INI.parse("[section")
+  describe "parse" do
+    describe "from String" do
```
Sija:
```suggestion
    context "from String" do
```

spec/std/ini_spec.cr
```diff
@@ -2,71 +2,77 @@ require "./spec_helper"
...
-        },
-        "section1" => {
-          "foo" => "1.1",
-          "bar" => "2",
-        },
-        "section2" => {
-          "x.y.z" => "coco lala",
-        },
-      })
+    describe "from IO" do
```
Sija:
```suggestion
    context "from IO" do
```

src/ini.cr
```diff
@@ -65,12 +65,12 @@ class INI
   #
   # INI.build({"foo" => {"a" => "1"}}, true) # => "[foo]\na = 1\n\n"
   # ```
-  def self.build(ini, space : Bool = false) : String
+  def self.build(ini : Hash(String, Hash(String, String)) | T, space : Bool = false) : String forall T
```
asterite: This makes no sense, saying "well, it can be a Hash or basically anything else".

j8r: Yes, else we don't have any idea of the "shape" is supposed to have, nor other types supported.
It could also be `Hash | NamedTuple`, or we enforce that an INI must only be `Hash(String, Hash(String, String))`

asterite: How about removing the type restriction? It can be anything that's iterable. It can even be an array of tuples. But in Crystal we don't have a way to talk about those things (no interfaces, no shapes, etc.) so it's better to remove this.

j8r: I must bet at least an `Enumerable`, doesn't it?

asterite: No, you can have `each` on a type, not have it be `Enumerable`, and the code will work.

j8r: Ok, however I still don't see the use case to do this.

asterite: The benefit is not introducing a breaking change.

spec/std/ini_spec.cr
j8r: What does it change? I don't see the improvement.

Sija: Semantics, `describe` is typically used at top level, whereas `context` groups tests within.

j8r:
Anything remains to do?

PR #7835, opened 2019-05-31
Add Benchmark.memory

r00ster91:
Sometimes not only execution speed is of interest but also the consumed memory, when your program runs on a device with low RAM for example, like a Raspberry Pi. So I think it would be good to abstract IPS' internal GC stats total bytes computation into a method which can be conveniently used instead of having to wait for a benchmark entry to finish first just to see the consumed memory. There's also that `` Warning: benchmarking without the `--release` flag won't yield useful results `` message then which is unwanted output in that case.

### DIFF
--- a/src/benchmark/ips.cr
+++ b/src/benchmark/ips.cr
@@ -87,9 +87,8 @@ module Benchmark
           target = Time.monotonic + @calculation_time
 
           loop do
-            bytes_before_measure = GC.stats.total_bytes
-            elapsed = Time.measure { item.call_for_100ms }
-            bytes += (GC.stats.total_bytes - bytes_before_measure).to_i64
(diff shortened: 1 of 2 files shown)


src/benchmark/ips.cr
```diff
@@ -87,11 +87,11 @@ module Benchmark
           target = Time.monotonic + @calculation_time
 
           loop do
-            bytes_before_measure = GC.stats.total_bytes
-            elapsed = Time.measure { item.call_for_100ms }
-            bytes += (GC.stats.total_bytes - bytes_before_measure).to_i64
+            elapsed = nil
+            bytes_taken = Benchmark.memory { elapsed = Time.measure { item.call_for_100ms } }
```
RX14: this should at the least be on multiple lines

wontruefree:
This is great I have been wanting to do this with some projects but have not looked into it.
I would totally use this.

PR #7836, opened 2019-05-31
Fixed ‎CVE-2019-11358

GermanAizek:


Sija:
This CVE refers to jQuery - which isn't used - could you please elaborate on the problem and your fix?

GermanAizek:
Excuse me - I wrote the wrong number.
Correct: CVE-2019-11358

Sija:
New CVE id you've posted still refers to jQuery and it's even less clear about the reason for this fix - could you please explain it?

j8r:
Crystal Docs doesn't use JQuery, but Crystal Playground does. Is it impacted? Another, more brutal solution, is to get rid of this library :smile:

GermanAizek:
Example:
localhost#q=

after executing query, will display 0 in alert window

straight-shoota:
@GermanAizek Thanks for providing a fix, but honestly, I can't tell what it is for.

I can't reproduce this vulnerability with Crystal docs (for example in https://crystal-lang.org/api/0.28.0/index.html#q=). As already mentioned, the docs generator does not even use jQuery.

asterite:
Nobody seems to be able to reproduce this, so I'm going to close this.

PR #7837, opened 2019-05-31
Allow running compiler_specs with specific flags

### LINKED ISSUE #7805: Default compiler flags (or cli options)
I would like to be able to run the compiler specs against some specific flags.
This would help to test some opt-in features behind flags (overflow, mt, etc).

1. I would like to agree on the scope:
  1.a) This change could be either be scoped only the compiler_specs (changing https://github.com/crystal-lang/crystal/blob/0.28.0/spec/spec_helper.cr#L160) or,
  1.b) to the whole compiler (changing https://github.com/crystal-lang/crystal/blob/0.28.0/src/compiler/crystal/compiler.cr#L198).

2. I would like to agree on the interface
  2.a) Use an env variable like `CRYSTAL_COMPILER_FLAGS` that will contain a space separated list of flags to be used.
  2.b) Use an env variable to mimic any CLI option of the compiler via an env variable like `CRYSTAL_COMPILER_OPTS`. The implementation will need to parse them (instead of just split spaces).

I would be happy with 1.b + 2.a.

bcardiff:
Fixes #7805 

This PR solves the mentioned issue in the narrowest way: 1.a (only affecting the specs) + 2.a (only allowing compiler flags).

This is use full for testing compiler features behind flags locally and could be used for testing compiler with different configurations in CI.

### DIFF
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -117,6 +117,7 @@ def warnings_result(code, inject_primitives = true)
   compiler.warnings = Warnings::All
   compiler.error_on_warnings = false
   compiler.prelude = "empty" # avoid issues in the current std lib
+  apply_program_flags(compiler.flags)
   result = compiler.compile Compiler::Source.new("code.cr", code), output_filename
 
   result.program.warning_failures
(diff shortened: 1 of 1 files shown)


spec/spec_helper.cr
```diff
@@ -160,9 +161,24 @@ end
...
+# Separate flags with a space.
+# Using CRYSTAL_SPEC_COMPILER_FLAGS="foo bar" will mimic -Dfoo -Dbar options.
+private def apply_program_flags(target)
+  ENV["CRYSTAL_SPEC_COMPILER_FLAGS"]?.try { |f| target.concat(f.split) }
+end
+
+private def encode_program_flags : String
+  f = [] of String
+  apply_program_flags(f)
+  f.map { |x| "-D#{x}" }.join(" ") || ""
```
j8r:
```suggestion
  f.map { |x| "-D#{x}" }.join(' ')
```

PR #7845, opened 2019-06-02
Add option stop_at_subcommand to OptionParser.parse

didactic-drunk:
<ins>This PR adds an option `stop_at_subcommand: true` to `OptionParser.parse` which stops argument parsing at the first subcommand. This allows to parse subcommands separately</ins> (updated description inserted by @straight-shoota).

Original text:

Most of the changes are from a variable rename to be more descriptive or method additions for consistency with parse().  The only method actually changed is ParseTask#initialize.

### DIFF
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -395,6 +395,47 @@ describe "OptionParser" do
     end
   end
 
+  describe "stop_at_subcommand" do
+    it "stops at foo" do
+      args = ["-f", "foo", "-b", "bar", "--", "baz"]
+      f = false
+      b = false
+      OptionParser.parse(args, stop_at_subcommand: true) do |opts|
+        opts.on("-f", "some flag") do
+          f = true
+        end
+        opts.on("-b", "some flag") do
+          b = true
+        end
+      end
+      f.should be_true
+      b.should be_false
+      args.should eq(["foo", "-b", "bar", "--", "baz"])
+    end
+
+    it "still stops at --" do
+      args = ["-f", "--", "-g", "foo", "-b", "bar", "--", "baz"]
+      f = false
+      g = false
+      b = false
+      OptionParser.parse(args, stop_at_subcommand: true) do |opts|
+        opts.on("-f", "some flag") do
+          f = true
+        end
+        opts.on("-g", "some flag") do
+          g = true
+        end
+        opts.on("-b", "some flag") do
+          b = true
+        end
+      end
+      f.should be_true
+      g.should be_false
+      b.should be_false
+      args.should eq(["-g", "foo", "-b", "bar", "--", "baz"])
+    end
+  end
+
   describe "forward-match" do
     it "distinguishes between '--lamb VALUE' and '--lambda VALUE'" do
       args = %w(--lamb value1 --lambda value2)
--- a/src/option_parser.cr
+++ b/src/option_parser.cr
@@ -52,10 +52,11 @@ class OptionParser
 
   # Creates a new parser, with its configuration specified in the block,
   # and uses it to parse the passed *args* (defaults to `ARGV`).
-  def self.parse(args = ARGV) : self
+  # Stops at first non-dash argument when stop_at_subcommand is `true`.
+  def self.parse(args = ARGV, *, stop_at_subcommand : Bool = false) : self
     parser = OptionParser.new
     yield parser
-    parser.parse(args)
+    parser.parse(args, stop_at_subcommand: stop_at_subcommand)
     parser
   end
 
@@ -181,8 +182,9 @@ class OptionParser
   end
 
   # Parses the passed *args* (defaults to `ARGV`), running the handlers associated to each option.
-  def parse(args = ARGV)
-    ParseTask.new(self, args).parse
+  # Stops at first non-dash argument when stop_at_subcommand is `true`.
+  def parse(args = ARGV, *, stop_at_subcommand = false)
+    ParseTask.new(self, args, stop_at_subcommand: stop_at_subcommand).parse
   end
 
   @[Deprecated("Use `parse` instead.")]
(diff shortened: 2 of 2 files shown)


chocolateboy:
I don't think `order` is a good name for this method in Ruby's OptionParser, so it would be nice to take this opportunity to use a more descriptive name. Given that the parse methods are written infrequently (i.e. typically no more than once per application/codebase), there's no particular need for the name to be Huffman-coded.

Some alternatives:

- match
- **parse?**<sup>1</sup>
- parse_initial
- parse_known
- parse_partial
- parse_valid

<sup>1</sup> Although the `?` suffix is mainly used for predicates, it's occasionally used to mark a safe version of a method which doesn't raise on error e.g. `Hash#dig?` and `Indexable#[]?` .

didactic-drunk:
parse_subcommand ?

Happy to rename when a decision is made.

didactic-drunk:
The circleci error looks like a spurious OOM.  I've started testing this against a 4 layer deep subcommand ruby conversion and it seems to work.

chocolateboy:
> parse_subcommand ?

I think it's clearer to describe the behavior rather than a particular usage. `parse_subcommand` suggests that the method parses options for a subcommand, when in practice (for me at least) it's the other way round i.e. the parsed options belong to a wrapper command (e.g. rlwrap or xargs) and the remaining *unparsed* arguments are passed through to the wrapped (sub)command.

jhass:
How about making this an optional keyword argument to `parse`? As for the name... `parse(args, stop_at_subcommand: true)`, idk I'm not finding a good name either.

chocolateboy:
It might be too late for this (since it would be a breaking change<sup>1</sup>), but if [this proposal](https://github.com/crystal-lang/crystal/issues/7846) is accepted, another option could be to use `parse!` for the current method, to reflect the fact that it's "unsafe" (i.e. may raise an exception), and then this safe version could just be called `parse`:

```ruby
# parse args (default: ARGV) and raise an exception on error
parse!
parse!(args)

# parse args (default: ARGV) up to the first non-option
parse
parse(args)
```

<sup>1</sup> Though in practice, code using `parse!` (i.e. most code) would be unaffected by the change.

didactic-drunk:
I suspect #7846 would need @deprecated on parse! for a while.

#7846 could be combined with this PR
```
def parse(args = ARGV, stop_at_subcommand = false)
end

@[Deprecated("Use `#parse` instead")]
def parse!
end
```

or 

```
def parse_until_the_first_nondash_or_doubledash_probably_for_a_subcommand_but_maybe_you_have_other_reasons_you_can_list_them_in_pr_7845_have_a_nice_day(args)
```

Either way is fine.

chocolateboy:
Combining the PRs sounds like a good idea to me.

> ```ruby
> def parse_until_the_first_nondash_or_doubledash_probably_for_a_subcommand_but_maybe_you_have_other_reasons_you_can_list_them_in_pr_7845_have_a_nice_day(args)
> ```

As happens all too often on Crystal issues, this discussion has quickly descended into downvotes (as a substitute for discussion) and snide asides, so I'll take my bikeshedding elsewhere. I use this method a fair bit in Ruby CLI tools and look forward to using it in Crystal, whatever it ends up being called.

didactic-drunk:
@chocolateboy My comment was an attempt at humor with a hint of IDGAS what it's called (the 2nd bit of code) with a preference for an argument added to a method (the first bit of code).

Is there a humor markdown somewhere?

#### **Serious mode**
My silence on your comments regarding my suggestion of parse_subcommand in my mind was an admission that my idea sucked, you were right, I moved on.

`order()` was chosen for ruby compatibility.   Personally I see parse() and parse!() having different behaviors confusing and more confusing than `order`.   The additional param tells me what's going on when I go to read code years down the line.  If there was a clear concise name I would probably go for it but right now I don't think one has been proposed.

Either way it doesn't matter.  No one has mentioned what will be accepted. As soon as someone tells me what will be accepted for merging I'll update code and push it. IDK what the name is.

didactic-drunk:
Should I downvote my own parse_subcommand suggestion or delete it? I left it for others to see the full conversation with a downvote for now.

didactic-drunk:
Recently rebased and removed `OptionParser#order!`.  

Is `order` for a name acceptable?  `parse_subcommand`?  `stop_at_subcommand` as an option?  Anything?

src/option_parser.cr
```diff
@@ -199,10 +216,19 @@ class OptionParser
...
 
   private struct ParseTask
-    @double_dash_index : Int32?
+    @stop_index : Int32?
+
+    def initialize(@parser : OptionParser, @args : Array(String), order = false)
+      double_dash_index = @stop_index = @args.index("--")
+
+      # stop at first nondash param unless "--" comes earlier
+      if order && (order_index = @args.index { |arg| arg =~ /^[^\-]/ })
```
RX14: `|arg| !arg.starts_with?('-')`?

src/option_parser.cr
```diff
@@ -199,10 +216,19 @@ class OptionParser
...
+        unless (ddi = double_dash_index) && ddi < order_index
```
RX14: shouldn't need the `ddi = double_dash_index` here, since `double_dash_index` is a local variable.

didactic-drunk:
> I really don't like the `order` name either, I'd suggest `.parse(stop_at_subcommand: true)`.

`order` removed and changed to `.parse(stop_at_subcommand: true)`.

spec/std/option_parser_spec.cr
```diff
@@ -395,6 +395,47 @@ describe "OptionParser" do
     end
   end
 
+  describe "order" do
```
Sija: Old name left.

RX14:
@straight-shoota no, it does stop at the first subcommand (`foo --foo --bar subcommand --baz` will get you `--foo` and `--bar`, then stop at `subcommand`, so you can `case subcommand; ...; end` then continue parsing the rest with *another* `OptionParser` instance.

straight-shoota:
But what if `--bar` expects a value?

didactic-drunk:
@straight-shoota is correct.  I'm working on it but it requires more changes to the parser that I don't have time for right now.

RX14:
@straight-shoota ah! I understand the issue now, sorry.

PR #7854, opened 2019-06-04
Fix Def#to_s with **options and &block

### LINKED ISSUE #7853: Crystal playground bug
I seem to have encountered a bug in the Crystal playground. Here is the input code:

```crystal
class Agent
  
  property sessions : Array(String)
  property cookies : Array(String)
  
  def initialize(*, sessions = nil, cookies = nil)
    @sessions = sessions || [] of String
    @cookies = cookies || [] of String
  end
  
  def self.new(**options, &block)
    yield new(**options)
  end
  
end

Agent.new do |agent|
  pp agent.cookies
end
```

Which I know might be terrible, but I was trying to figure out why some other code of mine wasn't working so I made this little example. When I ran it at carc.in it timed out so I opened up the playground and tried it. It returned the error `Syntax error in play:12: expecting token ')', not '&'` after saying I'd found a bug.

Without the `&block` it compiles.

MakeNowJust:
Fixed #7853

It's quite simple bug: missing a line `printed_arg = true` on outputting `**options`. Thank you.

### DIFF
--- a/spec/compiler/parser/to_s_spec.cr
+++ b/spec/compiler/parser/to_s_spec.cr
@@ -161,4 +161,5 @@ describe "ASTNode#to_s" do
   expect_to_s %[(1..)]
   expect_to_s %[..3]
   expect_to_s "offsetof(Foo, @bar)"
+  expect_to_s "def foo(**options, &block)\nend"
 end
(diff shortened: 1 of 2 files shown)


bcardiff:
Thanks @MakeNowJust !

PR #7858, opened 2019-06-05
Adds File.readlink to match File.symlink

didactic-drunk:


src/crystal/system/unix/file.cr
```diff
@@ -105,6 +105,25 @@ module Crystal::System::File
...
   end
 
+  def self.readlink(path) : String
+    buf = Bytes.new 256
+    3.times do
+      ret = LibC.readlink(path, buf, buf.bytesize)
+      if ret == -1
+        raise Errno.new("readlink")
+      elsif ret == buf.bytesize
+        buf = Bytes.new(buf.bytesize * 4)
```
ysbaddaden: On the 3rd pass, if the buffer is still too small, it will allocate another memory chunk that will never be used.

src/crystal/system/unix/file.cr
```diff
@@ -105,6 +105,25 @@ module Crystal::System::File
     ret
   end
 
+  def self.readlink(path) : String
+    buf = Bytes.new 256
```
ysbaddaden: Reading the [readlink(2)](http://man7.org/linux/man-pages/man2/readlink.2.html) manpage, it asks `lstat` for the actual buffer size and falls back to `PATH_MAX if it's 0. That leads to a race condition between `lstat` and `readlink` and the result can still be truncated, and must be dealt with (retry or raise if buf.size == PATH_MAX).

src/crystal/system/unix/file.cr
```diff
@@ -105,6 +105,25 @@ module Crystal::System::File
...
+    3.times do
+      ret = LibC.readlink(path, buf, buf.bytesize)
+      if ret == -1
+        raise Errno.new("readlink")
+      elsif ret == buf.bytesize
+        buf = Bytes.new(buf.bytesize * 4)
+      else
+        buf[ret] = 0_u8
+        return String.new(buf[0, ret])
+        break
```
ysbaddaden: 1. you can pass the actual bytesize as `String.new(buf, ret)`;
2. `break` is unreachable after `return`;

ysbaddaden: BTW: `ret` could have a more explicit name, like `bytesize` for example.

didactic-drunk: 1. Should I use `String.new(buf.pointer, ret)`?  Add a new overload to String?  Keep String.new(buf[0, ret])? Do something else?

```
in src/crystal/system/unix/file.cr:120: no overload matches 'String.new' with types Slice(UInt8), Int64
Overloads are:
 - String.new(bytes : Bytes, encoding : String, invalid : Symbol | ::Nil = nil)
 - String.new(chars : ::Pointer(UInt8), bytesize, size = 0)
 - String.new(chars : ::Pointer(UInt8))
 - String.new(slice : Bytes)
 - String.new(capacity : Int, &block)
```

2. Changed to bytesize.  Not pushed yet.

src/crystal/system/unix/file.cr
didactic-drunk: The choice of initial buffer size and multiplication factor are [space, time, race condition] trade offs.

The longest link on my Mac is 138.  On Linux 135.

All requests on a normal system are handled by a single system call with buffer size 256. (Race condition free)
The 2nd request at 1024 will handle what I assume are outliers.
The 3rd requests hits the max on either Linux or Mac at 4096.
The 3 iter limit is designed for systems like GNU Hurd which may have extremely high or no practical limits.

Alternatively I could:
* lstat
* alloc + readlink
* loop on failure?

But this would add a race condition on every call.
Along with the overhead of 2 system calls.
And more heap allocations for the File::Info and underlying stat structure.

Race condition avoidance alone was enough for me to avoid the lstat method.

RX14:
@didactic-drunk CI is failing due to a typo `SSize` vs `SSizeT` on all platforms except for macos

spec/std/file_spec.cr
```diff
@@ -270,6 +270,12 @@ describe "File" do
     end
   end
 
+  describe "symlink?" do
```
straight-shoota:
```suggestion
  describe ".readlink" do
```

src/crystal/system/unix/file.cr
```diff
@@ -105,6 +105,28 @@ module Crystal::System::File
...
+    3.times do |iter|
+      bytesize = LibC.readlink(path, buf, buf.bytesize)
+      if bytesize == -1
+        raise Errno.new("readlink")
+      elsif bytesize == buf.bytesize
+        break if iter >= 2
+        buf = Bytes.new(buf.bytesize * 4)
+      else
+        buf[bytesize] = 0_u8
+        return String.new(buf[0, bytesize])
```
straight-shoota: This would be an alternative, but there is no real difference between both:

```suggestion
        return String.new(buf.to_unsafe, bytesize)
```

`buf[bytesize] = 0_u8` is unneccessary, though. The `String` constructor ensures the string ends with a zero byte. And the C string should be zero delimited anyway.

didactic-drunk: `= 0_u8` was removed.

From the linux man page:

> Conforming applications should not assume that the returned contents of the symbolic link are null-terminated.

RX14:
@bcardiff the github "PRs need two reviews to be merged" setting is blocking me from merging this PR because github counts @ysbaddaden's review as still pending. Please revert this change, this PR is ready to go.

straight-shoota:
You can dismiss outdated reviews. But yeah, if this setting was changed lately, it would probably be better to go back.

straight-shoota:
Thank you @didactic-drunk

RX14:
Huh, I tried to find the button to dismiss others reviews but I couldn't find it...

straight-shoota:
=> https://help.github.com/en/articles/dismissing-a-pull-request-review

RX14:
That's the most hidden UX i've seen in a while...

Issue #5467, opened 2017-12-27
Support `arm-linux-musleabihf` and `aarch64-linux-musl` targets


j8r:
`arm-linux-musleabihf` and `aarch64-linux-musl` targets will be needed to compile the crystal compiler and [package it](https://github.com/ysbaddaden/crystal-alpine/issues/6) for Alpine Linux on ARM architectures.

Looking at [Rust's targets](https://github.com/rust-lang/rust/tree/master/src/librustc_back/target) might be useful.

We should also change the triple `amd64-unknown-openbsd` to `x86_64-unknown-openbsd`

straight-shoota:
I don't think it makes much sense to keep this issue open. It originally started with different goals and the main one has been implemented for a long time.
We've had support for `aarch64-linux-musl` since 0.25 ([#5861](https://github.com/crystal-lang/crystal/pull/5861)).

`arm-linux-musleabihf` is a different story. The focus has clearly shifted towards more modern architectures, so this is probably not that relevant.
We're open to contributions for any new target platform. But `arm-linux-musleabihf` is not a priority.

ysbaddaden:
The `arm-linux-musleabihf` target should only be missing the `LibC` bindings since we already support `arm-linux-gnueabihf`. I can generate them if anybody needs the target.

PR #5862, opened 2018-03-25
WIP: Add support for arm-linux-musleabihf

jirutka:
⚠️ There’s some problem, see comments…

Related to #5467

PR #7859, opened 2019-06-05
Add support for arm-linux-musleabihf

j8r:
Follow up of #5862.
All credits go to @jirutka .


RX14:
is it tested?

ysbaddaden:
The original PR was as good as the `arm-linux-gnueabi` target, which means that it had the exact same hard/soft float issues on the same hardware. I couldn't test it better because I miss an armv7 Alpine image that I can just run in qemu (I use Raspbian for glibc).

j8r:
I got this errors when cross compiling with a newly compiled compiler:
```
export  LLVM_CONFIG=/usr/bin/llvm5-config
make FLAGS="--target arm-linux-musleabihf --link-flags=-no-pie --verbose"
```
```
...
/usr/lib/gcc/x86_64-alpine-linux-musl/8.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: _main.o: relocations in generic ELF (EM: 40)
/usr/lib/gcc/x86_64-alpine-linux-musl/8.3.0/../../../../x86_64-alpine-linux-musl/bin/ld: _main.o: error adding symbols: file in wrong format
```

j8r:
I would like to go forward on this, but struggling with cross-compilation :confused:

straight-shoota:
Resolves #5467

flying-sausages:
Bump?

PR #7860, opened 2019-06-05
Implement pow operator for BigDecimal and BigRational

### LINKED ISSUE #7680: BigDecimal missing ** method

jwbuiter:
Resolves #7680

Since @ananagame closed his PR, I decided to try to implement this for my first contribution.

Instead of an iterative approach, this implementation uses the operators of the underlying BigInt representation to achieve the same results.

I also added tests and tried to format the styles of the tests to match those in the surrounding file.

### DIFF
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -205,7 +205,7 @@ describe BigDecimal do
   end
 
   it "exponentiates" do
-    result = ("12.34".to_big_d ** 5)
+    result = "12.34".to_big_d ** 5
     result.should be_a(BigDecimal)
     result.to_s.should eq("286138.1721051424")
   end
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -204,6 +204,12 @@ describe BigDecimal do
     (3 / 2.to_big_d).should eq(BigDecimal.new("1.5"))
   end
 
+  it "exponentiates" do
+    result = ("12.34".to_big_d ** 5)
+    result.should be_a(BigDecimal)
+    result.to_s.should eq("286138.1721051424")
+  end
+
   it "can be converted from other types" do
     1.to_big_d.should eq (BigDecimal.new(1))
     "1.5".to_big_d.should eq (BigDecimal.new(15, 1))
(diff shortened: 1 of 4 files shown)


spec/std/big/big_decimal_spec.cr
```diff
@@ -204,6 +204,12 @@ describe BigDecimal do
     (3 / 2.to_big_d).should eq(BigDecimal.new("1.5"))
   end
 
+  it "exponentiates" do
+    result = ("12.34".to_big_d ** 5)
```
Sija:
```suggestion
    result = "12.34".to_big_d ** 5
```

jwbuiter: I added this to the batch commit.

src/big/big_rational.cr
```diff
@@ -170,6 +170,23 @@ struct BigRational < Number
     BigRational.new { |mpq| LibGMP.mpq_neg(mpq, self) }
   end
 
+  # Raises the rational to the *other*th power
+  #
+  # This will raise an exception if rational is 0 and other is negative.
```
r00ster91:
```suggestion
  # This will raise `ArgumentError` if rational is 0 and *other* is negative.
```

jwbuiter: Actually, it raises a `DivisionByZeroError`

r00ster91: edited

straight-shoota:
@jwbuiter could you apply the suggestions? You might've fixed them, but they're not in the branch.

straight-shoota:
Thank you @jwbuiter

Issue #7826, opened 2019-05-28
Request: HTTP::Params.encode Array value support


Daniel-Worrall:
Coming across #5172 and interacting with APIs that take multiple of the same param, it would be very handy to be able to pass in an array of values to the parameter to be parsed out. I don't think there's a different expectation that could be made to passing in an array of values

```cr
data = {"key" => ["a", "b"], "foo" => "bar"}

puts HTTP::Params.encode(data) # => "key=a&key=b&foo=bar"
```

https://carc.in/#/r/702o

Sija:
You can do it by passing `Hash(String, Array(String))` to the `HTTP::Params` constructor and invoking `#to_s`.

```crystal
data = {"key" => ["a", "b"], "foo" => ["bar"]}

puts HTTP::Params.new(data)
```

https://carc.in/#/r/702v

vladfaust:
If we speak about keeping stdlib slim, there is https://github.com/vladfaust/http-params-serializable which acts like `JSON::Serializable`, but for HTTP params :slightly_smiling_face:

straight-shoota:
Sounds like a good idea to add an overload to `HTTP::Params.encode` delegating to `.new` and `#to_s`, as in the example by @Sija.

bcardiff:
Or to extend the `HTTP::Params::Builder` to handle array like values in https://github.com/crystal-lang/crystal/blob/master/src/http/params.cr#L347

rodrigopinto:
I think overloading the `HTTP::Params.encode` is a way to go but also extending the `HTTP::Params::Builder` would allow to do `HTTP::Params::Builder#add("key", ["a", "b"])`. 

As no one started it yet, I can take care of this implementation. Is it ok?

rodrigopinto:
Working on the PR I realized that [HTTP::Params.encode(hash : Hash(String, String))](https://github.com/crystal-lang/crystal/blob/master/src/http/params.cr#L90) signature is restrictive in case someone wishes to passes other values types like `Int32`, 'Float64`, etc.

As we are implementing a method signature with `hash : Hash(String, String | Array(String)` for this issue, I was wondering if would make sense to have a Union type to allow other types of params for the value part of the hash.

About @Sija suggestion, we could extend the `new` + `to_s`,  but would match in the same situation as it expects a `Hash(String, Array(String))`. WDYT?

oprypin:
[...]

PR #7862, opened 2019-06-06
Adds support to HTTP::Params.encode encode an arrays of values for a key.

rodrigopinto:
# What is this PR for?

This PR is for adding support to `HTTP::Params.encode` encode a key with an array of values. Like:

```Crystal
data = {"key" => ["a", "b"], "foo" => "bar"}

puts HTTP::Params.encode(data) # => "key=a&key=b&foo=bar"
```

This closes https://github.com/crystal-lang/crystal/issues/7826

### DIFF
--- a/src/http/params.cr
+++ b/src/http/params.cr
@@ -85,9 +85,9 @@ module HTTP
     # ```
     # require "http/params"
     #
-    # HTTP::Params.encode({"foo" => "bar", "baz" => "qux"}) # => "foo=bar&baz=qux"
+    # HTTP::Params.encode({"foo" => "bar", "baz" => ["quux", "quuz"]}) # => "foo=bar&baz=quux&baz=quuz"
     # ```
-    def self.encode(hash : Hash(String, String))
+    def self.encode(hash : Hash(String, String | Array(String)))
       build do |builder|
         hash.each do |key, value|
           builder.add key, value
@@ -100,7 +100,7 @@ module HTTP
     # ```
     # require "http/params"
     #
-    # HTTP::Params.encode({foo: "bar", baz: "qux"}) # => "foo=bar&baz=qux"
+    # HTTP::Params.encode({foo: "bar", baz: ["quux", "quuz"]}) # => "foo=bar&baz=quux&baz=quuz"
     # ```
     def self.encode(named_tuple : NamedTuple)
       build do |builder|
@@ -344,12 +344,18 @@ module HTTP
       end
 
       # Adds a key-value pair to the params being built.
-      def add(key, value)
+      def add(key, value : String?)
         @io << '&' unless @first
         @first = false
         URI.escape key, @io
         @io << '='
         Params.encode_www_form_component value, @io if value
+        self
+      end
+
+      # Adds all of the given *values* as key-value pairs to the params being built.
+      def add(key, values : Array)
+        values.each { |value| add(key, value) }
         self
       end
 
--- a/spec/std/http/params_spec.cr
+++ b/spec/std/http/params_spec.cr
@@ -57,13 +57,13 @@ module HTTP
 
     describe ".encode" do
       it "builds from hash" do
-        encoded = Params.encode({"foo" => "bar", "baz" => "quux"})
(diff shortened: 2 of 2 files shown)


jwoertink:
How would slightly more complex structures look?

```crystal
data = {"key" => {"subkey" => ["a", "b"], "foo" => "bar", "baz" => [["c"], ["d"]]}
puts HTTP::Params.encode(data) # => ??
```

Or would just say that the default lib only supports very simple structures, and all complex matters need to be handled by individual frameworks?

oprypin:
application/x-www-form-urlencoded  doesn't define any such complex structures

oprypin:
So the thing is,
this already exists
```crystal

data = {"key" => ["a", "b"], "foo" => ["bar"]}

puts HTTP::Params.new(data) # => "key=a&key=b&foo=bar"
```
https://carc.in/#/r/71jb

so why introduce another slightly different and inconsistent approach?

oprypin:
The implementation here looks good though, and actually doesn't add much complexity.

rodrigopinto:
@oprypin The problem on extending `Params.new` is that it expects `HTTP::Params.new(raw_params : Hash(String, Array(String)))` so the parameters should always be ` String => Array(String)` and would never accept ` String => String`. 

So instead of extending the `Params.new` I  added the behavior mentioned on the issue.

From my point of view, be too strict on only accepting values as Array, is like passing the problem to the clients that will use `Params.new`. I don't mind to have another implementation of the `Params.new` accepting `HTTP::Params.new(raw_params : Hash(String, String))`, but then I would change the current implementation. WDYT?

src/http/params.cr
```diff
@@ -85,12 +85,12 @@ module HTTP
...
     #
-    # HTTP::Params.encode({"foo" => "bar", "baz" => "qux"}) # => "foo=bar&baz=qux"
+    # HTTP::Params.encode({"foo" => "bar", "baz" => ["quux", "quuz"]}) # => "foo=bar&baz=quux&baz=quuz"
     # ```
-    def self.encode(hash : Hash(String, String))
+    def self.encode(hash : Hash(String, String | Array(String)))
       build do |builder|
         hash.each do |key, value|
-          builder.add key, value
+          builder.add(key, value)
```
asterite:
```suggestion
          builder.add key, value
```

This change is not necessary (less diff, better git history).

src/http/params.cr
```diff
@@ -100,12 +100,12 @@ module HTTP
...
     # require "http/params"
     #
-    # HTTP::Params.encode({foo: "bar", baz: "qux"}) # => "foo=bar&baz=qux"
+    # HTTP::Params.encode({foo: "bar", baz: ["quux", "quuz"]}) # => "foo=bar&baz=quux&baz=quuz"
     # ```
     def self.encode(named_tuple : NamedTuple)
       build do |builder|
         named_tuple.each do |key, value|
-          builder.add key.to_s, value
+          builder.add(key.to_s, value)
```
asterite:
```suggestion
          builder.add key.to_s, value
```

This change is not necessary (less diff, better git history)

src/http/params.cr
```diff
@@ -353,6 +353,11 @@ module HTTP
         self
       end
 
+      def add(key, values : Array)
```
asterite: This is missing some docs. I suggest:

```
# Adds all of the given *values* as key-value pairs to the params being built.

File: src/http/params.cr [hunk 4]
```diff
@@ -343,8 +343,8 @@ module HTTP
         @first = true
       end
 
-      # Adds a key-value pair to the params being built.
-      def add(key, value)
+      # Adds a key-value pair to the/ params being built.
```
asterite:
```suggestion
      # Adds a key-value pair to the params being built.
```

yxhuvud:
In practice, one problem here is that there is no standardized way to do it, so people have invented a whole bunch of ways. See https://swagger.io/docs/specification/describing-parameters/#query-parameters for a list of a bunch of different options.

rodrigopinto:
@asterite Applied the CR suggestions and pushed. The red on CI is something not related.

PR #7864, opened 2019-06-06
Natively support SameSite cookies

### LINKED ISSUE #7839: Natively support SameSite cookie option
Currently `HTTP::Cookie` only supports the `SameSite` option via the extension property.  I propose to add native support for it, similar to secure or http_only options.  

The new property would be nilable with a default of nil (to represent disabled samesite option).  The two valid options could be stored in an enum to provide type safety, as opposed to manually giving it "lax" or "strict".

Blacksmoke16:
Resolves #7839

### DIFF
--- a/spec/std/http/cookie_spec.cr
+++ b/spec/std/http/cookie_spec.cr
@@ -81,7 +81,7 @@ module HTTP
         cookie = parse_set_cookie("key=value; Secure")
         cookie.name.should eq("key")
         cookie.value.should eq("value")
-        cookie.secure.should eq(true)
+        cookie.secure.should be_true
         cookie.to_set_cookie_header.should eq("key=value; path=/; Secure")
       end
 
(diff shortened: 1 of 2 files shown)


src/http/cookie.cr
```diff
@@ -1,24 +1,33 @@
 require "./common"
 
 module HTTP
-  # Represents a cookie with all its attributes. Provides convenient
-  # access and modification of them.
+  # Represents a cookie with all its attributes. Provides convenient access and modification of them.
   class Cookie
+    # Possible values for the SameSite cookie as described in the [Same-site Cookies Draft](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1.1).
```
straight-shoota:
```suggestion
    # Possible values for the `SameSite` cookie as described in the [Same-site Cookies Draft](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1.1).
```

spec/std/http/cookie_spec.cr
```diff
@@ -143,16 +175,17 @@ module HTTP
...
-        cookie.secure.should eq(true)
-        cookie.expires.should eq(time)
+        cookie.name.should eq "key"
+        cookie.value.should eq "value"
+        cookie.path.should eq "/test"
+        cookie.domain.should eq "www.example.com"
+        cookie.http_only.should be_true
+        cookie.secure.should be_true
+        cookie.expires.should eq time
+        cookie.samesite.should eq HTTP::Cookie::SameSite::Strict
```
Sija:
```suggestion
        cookie.samesite.should eq :strict
```

spec/std/http/cookie_spec.cr
```diff
@@ -81,18 +81,50 @@ module HTTP
...
+            cookie.to_set_cookie_header.should eq "key=value; path=/; SameSite=Lax"
+          end
+        end
+
+        context "Strict" do
+          it "parses samesite" do
+            cookie = parse_set_cookie("key=value; SameSite=Strict")
+            cookie.name.should eq "key"
+            cookie.value.should eq "value"
+            cookie.samesite.should eq HTTP::Cookie::SameSite::Strict
```
Sija:
```suggestion
            cookie.samesite.should eq :strict
```

spec/std/http/cookie_spec.cr
```diff
@@ -81,18 +81,50 @@ module HTTP
...
         cookie.to_set_cookie_header.should eq("key=value; path=/; HttpOnly")
       end
 
+      describe "SameSite" do
+        context "Lax" do
+          it "parses samesite" do
+            cookie = parse_set_cookie("key=value; SameSite=Lax")
+            cookie.name.should eq "key"
+            cookie.value.should eq "value"
+            cookie.samesite.should eq HTTP::Cookie::SameSite::Lax
```
Sija:
```suggestion
            cookie.samesite.should eq :lax
```

src/http/cookie.cr
```diff
@@ -113,6 +125,7 @@ module HTTP
           domain: match["domain"]?,
           secure: match["secure"]? != nil,
           http_only: match["http_only"]? != nil,
+          samesite: (ss = match["samesite"]?) ? SameSite.parse? ss : nil,
```
Sija:
```suggestion
          samesite: match["samesite"]?.try &->SameSite.parse?(String),
```

spec/std/http/cookie_spec.cr
straight-shoota: This is not an improvement. I'd accept both variants, but changing it has no benefit.

PR #7871, opened 2019-06-08
Fix cipher expectation in OpenSSL Socket spec

### LINKED ISSUE #7869: OpenSSL Socket spec fails on Alpine
When running all crystal specs on Alpine at edge, we got this error, as reported by mps on IRC:
```
  1) OpenSSL::SSL::Socket returns the cipher that is currently in use
     Failure/Error: client.cipher.should contain "RSA"

       Expected:   "TLS_AES_256_GCM_SHA384"
       to include: "RSA"

     # spec/std/openssl/ssl/socket_spec.cr:18
```
This spec was integrated in https://github.com/crystal-lang/crystal/pull/7445

I think this is due to a default cipher change in recent OpenSSL. I don't know precisely since when (seems 1year?), this [doc](https://github.com/openssl/openssl/blob/master/doc/man3/SSL_CTX_set_cipher_list.pod) says that `TLS_AES_256_GCM_SHA384` is the default cipher. The spec is also not ideal, why the cypher should contain RSA?
Not empty will be an improvement, and maybe check if the string is in uppercase.

j8r:
Recent OpenSSL defaults to TLS_AES_256_GCM_SHA384.
Fixes #7869.

### DIFF
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -15,7 +15,9 @@ describe OpenSSL::SSL::Socket do
       end
 
       client = server.accept
-      client.cipher.should contain "RSA"
+      cipher = client.cipher
+      cipher.upcase.should eq cipher
+      cipher.should_not be_empty
       client.close
     end
   end
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -15,9 +15,7 @@ describe OpenSSL::SSL::Socket do
       end
 
       client = server.accept
-      cipher = client.cipher
-      cipher.upcase.should eq cipher
-      cipher.should_not be_empty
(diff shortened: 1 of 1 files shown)


spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -15,7 +15,9 @@ describe OpenSSL::SSL::Socket do
       end
 
       client = server.accept
-      client.cipher.should contain "RSA"
+      cipher = client.cipher
+      cipher.upcase.should eq cipher
```
straight-shoota: I'm not entirely sure about this, maybe checking for not empty would be sufficient.

j8r: The best would be to check if the cipher is one of https://github.com/openssl/openssl/blob/master/include/openssl/ssl.h#L176, but this was complicating the test.

straight-shoota:
Thank you @j8r!

PR #7886, opened 2019-06-13
Fix edge cases in code parser and stringifier (found by fuzzing)

oprypin:
https://github.com/oprypin/crystal-fuzzing

### DIFF
--- a/spec/compiler/lexer/lexer_spec.cr
+++ b/spec/compiler/lexer/lexer_spec.cr
@@ -222,6 +222,8 @@ describe "Lexer" do
   it_lexes_i64 ["2147483648", "-2147483649"]
   it_lexes_i64 [["2147483648.foo", "2147483648"]]
   it_lexes_u64 ["18446744073709551615", "14146167139683460000", "9223372036854775808"]
+  it_lexes_number :u64, ["10000000000000000000_u64", "10000000000000000000"]
+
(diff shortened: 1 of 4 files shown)


oprypin:
It's failing though... Wasn't able to figure out ranges+newlines yet

oprypin:
Excluded the failing commit, will get back to that one later.

oprypin:
(in another PR)

bcardiff:
Thanks! Can the PR be named in a way it reflects the impact and no the procedure so it’s more meaningful as a changelog entry in the future?

oprypin:
The commits are meaningful changelog entries, the PR is indeed arbitrarily grouped by procedure.

asterite:
I think Range and newline doesn't need to be supported because `1.. \n` is parsed as an endless range now.

oprypin:
Alright, then I will not attempt to cover it. Seems consistent with the rest of the language, things like 
```
(1
+2) == 2
```
But still doing it in a separate PR.
https://github.com/crystal-lang/crystal/pull/7888

asterite:
Yes, now I'm actually not sure, probably endless range only makes sense when `)` comes afterwards.

oprypin:
Well then, good that I split out the controversial thing.
This PR is still good though! :D

straight-shoota:
Thank you @oprypin !

Issue #1388, opened 2015-09-06
[RFC] Pipe Operator


felixbuenemann:
It would be great if Crystal had a construct similar to [Elixir's pipe operator](http://elixir-lang.org/docs/v1.0/elixir/Kernel.html#%7C%3E/2) `|>`:

```crystal
class Importer
  def import(io)
    xml = parse_xml(io)
    root = find_root_node(xml)
    tree = build_tree(root)
  end

  def import2(io)
    io |> parse_xml |> find_root_node |> build_tree
  end
end
```
I think that the latter example is much easier to read and requires far less eye tracking to comprehend.

The implementation in Elixir is simply syntactic sugar using a macro, not sure if Crystal's macros could do this transform as well.

This has been briefly discussed in #1099, but I don't think the arguments against it were valid. Crystal is just like Ruby a hybrid between an OO and functional language and I really like the funtional core, imperative shell pattern, which leads to well testable units by avoiding shared/hidden state.

jhass:
We're fairly thin on the available syntax space already, so we should be very careful in using that up.

I'm not sure I find it more clear personally, it's something that makes sense because you're used to it, but you have to learn it. Syntax is always something to learn, I hang on on #ruby a lot, and "What does `&:` do" is one of the most common questions, people don't even recognize that it's `&(:foo)` not `&:(foo)`. The point is not about drawing an analogous ambiguity in your proposal, but why I think that too much syntax can be harmful in learning a language.

felixbuenemann:
People already now how to use pipes from the command line, so I think it's pretty easy to grasp.

jhass:
Or are they? On the command line the pipe passes the output to an invisible parameter (stdin) that doesn't even exist in Crystal, here it magically turns a argument-less call into one with an argument (now think about the ambiguity that adds if there's an overload that takes one and one that doesn't). The analogy would rather be "`a |> b` looks just `a | xargs b`", which it does not.

asterite:
This was briefly discussed in point 4 of #1099.

Right in your first example you have:

```crystal
xml = parse_xml(io)
root = find_root_node(xml)
```

In Crystal, right now, you can do this:

```crystal
require "xml"

xml = XML.parse("<foo>1</foo>")
xml.root
```

[...]

PR #7887, opened 2019-06-14
Implement Pipe |> operator

S-YOU:
```
% .build/crystal eval '"Hello World!" |> puts'
Hello World!
```

What do you think?

### DIFF
--- a/src/compiler/crystal/syntax/ast.cr
+++ b/src/compiler/crystal/syntax/ast.cr
@@ -2236,6 +2236,23 @@ module Crystal
     def_equals_and_hash constraint, exp
   end
 
+  class Pipe < ASTNode
+    property body : ASTNode
+
+    def initialize(@body : ASTNode)
+    end
+
+    def accept_children(visitor)
+      @body.accept visitor
+    end
+
+    def clone_without_location
+      Pipe.new(@body.clone)
+    end
+
+    def_equals_and_hash @body
+  end
+
   enum Visibility : Int8
     Public
     Protected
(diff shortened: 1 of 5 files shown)


ysbaddaden:
That it was already refused a long time ago #1388 as "doesn't fit the language".

Crystal uses chaining instead. See https://github.com/crystal-lang/crystal/issues/1388#issuecomment-138083911

asterite:
Note that pipe in Ruby 2.7 is equal to a dot. It's different from your implementation. I don't think it's worth adding it like Ruby. Maybe like Elixir and Elm yes, but it's already been discussed in the past and rejected.

asterite:
See https://github.com/ruby/ruby/commit/f169043d81524b5b529f2c1e9c35437ba5bc3a7a and all the comments against it.

S-YOU:
> See ruby/ruby@f169043 and all the comments against it.

That's because of Ruby implementation is very similar to dot and not much meaning. My Implementation is similar to Elixir I believe.

asterite:
@S-YOU That's fine, but we never agreed that a pipeline operator for a non-functional language was a good idea.

shuya-takano:
I doubt that too. Let’s see how it is going on with Ruby.

absolutejam:
Is it possible to achieve this in a shard?

ShalokShalom:
Ruins the whole language for me. I wonder why people, who have never coded proficiency in F#, the language it comes from, make remarks on how useful it is.

PR #7888, opened 2019-06-14
Don't allow ranges to span across lines

oprypin:
`1..\n2..` was turning into `(1..2)..`

This is a backwards-incompatible change because `1..\n2` is also no longer parsed as one range.


oprypin:
Also note:
```crystal
(1..
2) == (1..2)
```
After:
```crystal
(1..
2) == 2
```
----
Also also note (regardless of this change):
```crystal
(1+
2) == 3)
```
```crystal
(1
+2) == 2)
```

PR #7889, opened 2019-06-14
Formatter: format top-level inline macros

asterite:
This PR lets the formatter also format code inside top-level inline macros that have no interpolation. For example:

```crystal
a = 1

{% if flag?(:foo) %}
  a += 2     # this will get formatted
{% else %}
  a *=  3    # this will also get formatted
{% end %}
```

Note that code inside `{% if ... %}` found in inline macros (not the ones defined with `macro`) and that doesn't have interpolation (`{{ ... }}`) nor iteration (`{% ... %}`) must produce a valid Crystal expression, and so the formatter can take this into account and try to format that as a sub-program.

It might seem that this isn't a lot, but I still think it's useful. Many times one has to do conditional programming based on flags (`{% if flag?(:darwin) %}`) and usually the code there doesn't have interpolation nor iteration. And in those cases it's useful to be able to edit while having the formatter help along.

The formatter will indent the code inside such macros, so in the end some might think that it doesn't align well with the outer code. Maybe. But I think having consistency and not having to even think about "should I indent this or not" is a great benefit that is already provided by the formatter in other places.

I thought the implementation was going to be a bit more trivial but it turns out that macro code can access variables outside of them, so for example `a |= 3` will give a syntax error if `a` isn't defined before. Or for example macros inside `lib` and `enum` use different parsers. Like those, I had to consider other edge cases and for each I added a formatter spec.

Original idea by @waj :-)


asterite:
Thinking a bit more about this, maybe instead of changing the formatter I can make the `Parser` already produce valid AST nodes if there's no interpolation/iteration inside such macros. Then the formatter will format those things as usual. And also if there's a syntax error, or even a semantic error, the error won't show "in expanded macro" because there's no need for that.

I'll give it a try if I have time... so we might want to hold on a bit before merging this.

asterite:
Actually, I think what I said above might make things more complex than just keeping this thing in the formatter. I was thinking maybe the compiler could avoid showing "in expanded macro" for such cases on syntax/semantic error, but we can do the same trick if needed, and there's no need to parse both branches of an `if` "just in case" in the parser.

PR #7890, opened 2019-06-15
Add String#occurrences

r00ster91:
This adds `String#occurrences` which returns the amount of occurrences of a string in a string.
This is already possible with `"hello hello".scan("ll").size # => 2` but that's very inefficient (creates an array + a bunch of strings, just to get the array's size).
The best name for this method, `count` is unfortunately already taken by `String#count`.

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1303,6 +1303,12 @@ describe "String" do
     end
   end
 
+  describe "occurrences" do
+    it "returns the amount of occurrences of a string in a string" do
+      "hello hello".occurrences("ll").should eq(2)
+    end
+  end
+
   describe "gsub" do
     it "gsubs char with char" do
       "foobar".gsub('o', 'e').should eq("feebar")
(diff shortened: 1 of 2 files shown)


oprypin:
`scan` with a block also works fine.
And why is this not implemented using that?

Where is the rationale/discussion that this is needed in addition to `scan`?

You must define and test the behavior for overlapping matches.
`ababa.occurences(aba)`.

akadusei:
@r00ster91 There, currently, is no overload for `String#count` that accepts a `String`, so the name could still be `#count`?

oprypin:
As a side note, how many times will we need to implement aliases "execute this iterator but actually gather it into array" and "execute this iterator but actually just count the number of results"?

Oh wait, we actually have a common implementation for all iterators.
And then just deny block-yielding methods the status of iterators.

And instead for each method we duplicate the implementation and arbitrarily choose whether the dupe should return a custom-made iterator or an array.

And we still don't have a good way of just counting the results.

r00ster91:
I will just use `String#scan`'s block.

PR #7899, opened 2019-06-19
Allow setting default capacity for StringPool

carlhoerberg:
I think the previous default capacity of 8 entries is way too low for any read world use case, but maybe I'm wrong, can revert it if you want to.

### DIFF
--- a/spec/std/string_pool_spec.cr
+++ b/spec/std/string_pool_spec.cr
@@ -65,4 +65,12 @@ describe StringPool do
     end
     pool.size.should eq(10_000)
   end
+
+  it "can be created with larger initial capacity" do
+    pool = StringPool.new(capacity: 32)
+    s1 = pool.get "foo"
+    s2 = pool.get "foo"
+    s1.should be(s2)
+    pool.size.should eq(1)
+  end
 end
--- a/spec/std/string_pool_spec.cr
+++ b/spec/std/string_pool_spec.cr
@@ -67,7 +67,7 @@ describe StringPool do
   end
 
   it "can be created with larger initial capacity" do
-    pool = StringPool.new(capacity: 32)
+    pool = StringPool.new(initial_capacity: 32)
     s1 = pool.get "foo"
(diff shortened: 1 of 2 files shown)


RX14:
allocating two half-page allocations doesn't seem that bad

spec/std/string_pool_spec.cr
```diff
@@ -65,4 +65,12 @@ describe StringPool do
     end
     pool.size.should eq(10_000)
   end
+
+  it "can be created with larger initial capacity" do
```
yxhuvud: It is smaller with the other change, though.

asterite:
So, String Pool is also used when parsing JSON to reuse object keys. I doubt there will be 256 different keys. Can we keep it small but make it configurable?

(I don't know how this change affects the performance of parsing JSON, though)

src/string_pool.cr
```diff
@@ -33,8 +33,18 @@ class StringPool
...
+  # of the internal buffers in case of growth. If you have an estimate
+  # of the maximum number of elements the pool will hold it should
+  # be initialized with that capacity for improved performance.
+  #
+  # ```
+  # pool = StringPool.new(256)
+  # pool.size # => 0
+  # ```
+  def initialize(initial_capacity : Int = 8)
+    @capacity = initial_capacity
```
asterite: The type restriction is too wide and it won't work if you pass anything other than Int32. Let's make that Int32 instead.

spec/std/string_pool_spec.cr
carlhoerberg: hehe, yes, but reverted back to 8 entries now, so now it's true again ;)

spec/std/string_pool_spec.cr
```diff
@@ -65,4 +65,12 @@ describe StringPool do
...
+    pool = StringPool.new(capacity: 32)
```
asterite:
```suggestion
    pool = StringPool.new(initial_capacity: 32)
```

PR #7900, opened 2019-06-19
StringPool#clear

carlhoerberg:
Allow clearing a `StringPool`, resetting the buffers to their initial capacity

### DIFF
--- a/spec/std/string_pool_spec.cr
+++ b/spec/std/string_pool_spec.cr
@@ -65,4 +65,12 @@ describe StringPool do
     end
     pool.size.should eq(10_000)
   end
+
+  it "can be cleared" do
+    pool = StringPool.new
+    pool.get "foo"
+    pool.size.should eq(1)
+    pool.clear
+    pool.size.should eq(0)
+  end
 end
--- a/src/string_pool.cr
+++ b/src/string_pool.cr
@@ -34,7 +34,7 @@ class StringPool
 
   # Creates a new empty string pool.
   def initialize
-    @capacity = 8
+    @capacity = @initial_capacity = 8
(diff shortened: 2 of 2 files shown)


RX14:
At this point you might as well make a new stringpool, right?

Sija:
> At this point you might as well make a new stringpool, right?

Probably, but that's probably less performant than clearing the existing instance.

carlhoerberg:
Yes, but not if the string pool is a constant.

ysbaddaden:
This is exactly the same as creating a new StringPool: it creates new HEAP allocations and loses references that must be garbage collected. Maybe it could be interesting to keep the HEAP allocations and only the size was reset to `0` instead.

But, if the only use case is to modify a constant, then wouldn't a class/module property be a better fit?

carlhoerberg:
In our code it's difficult/impossible to pass around the pool. But in some
cases we need to reclaim the memory if the pool has grown too large,
clearing it is a very crude way of doing it. A size limit, and a way throw
away the least used strings would've been ideal..

asterite:
Also, I created `StringPool` exclusively for the compiler but poorly made it public, then I though about using it in JSON... so the API is not the best one. Just some thoughts on this.

(this is kind of similar to what happened with `Markdown`)

ysbaddaden:
@carlhoerberg I meant to replace an `App::POOL = StringPool.new` definition to:
```crystal
module App
  class_property pool = StringPool.new
end
```

Then change calls to `POOL` for `App.pool`, which behaves the same, except that the later but can be changed (unlike constants, which IMO should be thought as immutable).

carlhoerberg:
Fair enough

PR #7901, opened 2019-06-19
Fix: UUID#to_slice returns a dangling pointer

### LINKED ISSUE #7894: UUID slice has different value when passed to a method
From the conversation https://gitter.im/crystal-lang/crystal?at=5d07a49b2313502d385b7a3c

https://play.crystal-lang.org/#/r/730d

```cr
require "uuid"

def convert(value : ::UUID)
  value.to_slice
end

uuid = UUID.new "24b408d0-5f08-43d3-9027-c0d4ec097e3a"

uuid.to_slice # => Bytes[36, 180, 8, 208, 95, 8, 67, 211, 144, 39, 192, 212, 236, 9, 126, 58]
convert uuid # => Bytes[36, 180, 8, 208, 95, 8, 67, 211, 0, 0, 0, 0, 16, 0, 0, 0]
```

I can't imagine this being the desired behavior; but if it is, it should probably be documented somewhere.

ysbaddaden:
I thought we had `#to_unsafe_slice` somewhere in the stdlib but we don't, and we probably don't want to introduce it (unsafe is _bad_)~~, so I changed `UUID#to_slice` to allocate HEAP memory and thus return a safe slice instead, and changed the internal implementation to still take advantage of the unsafe slice~~. I removed `UUID#to_slice` and added `UUID#bytes` that returns a StaticArray that can be passed around, and sometimes be accessible as a slice (in safe contexts).

fixes #7894

### DIFF
--- a/src/uuid.cr
+++ b/src/uuid.cr
@@ -161,9 +161,9 @@ struct UUID
     end
   end
 
-  # Returns 16-byte slice.
-  def to_slice
-    @bytes.to_slice
+  # Returns the binary representation of the UUID.
+  def bytes : StaticArray(UInt8, 16)
+    @bytes.dup
   end
 
   # Returns unsafe pointer to 16-bytes.
@@ -173,7 +173,7 @@ struct UUID
 
   # Returns `true` if `other` UUID represents the same UUID, `false` otherwise.
   def ==(other : UUID)
-    to_slice == other.to_slice
+    @bytes == other.@bytes
   end
 
   # Convert to `String` in literal format.
@@ -184,7 +184,7 @@ struct UUID
   end
 
   def to_s(io : IO) : Nil
-    slice = to_slice
+    slice = @bytes.to_slice
 
     buffer = uninitialized UInt8[36]
     buffer_ptr = buffer.to_unsafe
@@ -200,7 +200,7 @@ struct UUID
   end
 
   def hexstring
-    to_slice.hexstring
+    @bytes.to_slice.hexstring
   end
 
   def urn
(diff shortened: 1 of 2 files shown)


asterite:
As I said in the original issue, I think we should remove `to_slice` and instead introduce a method to return a `StaticArray`. Then you return a copy which doesn't allocate, and you can invoke `to_slice` on it and pass it to an IO without allocating memory at all.

ysbaddaden:
Sure, how to name it? Just `UUID#bytes`?

ysbaddaden:
I removed `UUID#to_slice` and added `UUID#bytes` instead which is a much better solution!

src/uuid.cr
```diff
@@ -161,9 +161,9 @@ struct UUID
     end
   end
 
-  # Returns 16-byte slice.
-  def to_slice
-    @bytes.to_slice
+  # Returns the binary representation of the UUID.
+  def bytes : StaticArray(UInt8, 16)
+    @bytes.dup
```
ysbaddaden: I know `dup` is probably not required, but it makes the copy explicit.

src/uuid.cr
```diff
@@ -161,9 +161,9 @@ struct UUID
     end
   end
 
-  # Returns 16-byte slice.
-  def to_slice
```
bew: What about adding a deprecation annotation? 
But still fix #7894 by doing `@bytes.to_slice.dup` until we remove the method completely

src/uuid.cr
```diff
@@ -173,7 +173,7 @@ struct UUID
 
   # Returns `true` if `other` UUID represents the same UUID, `false` otherwise.
   def ==(other : UUID)
-    to_slice == other.to_slice
+    @bytes == other.@bytes
```
oprypin: Why introduce another case of `.@` for no reason?

src/uuid.cr
wontruefree: This should probably have a deprecation warning.

RX14:
Should we create another issue for `StatisArray#to_slice` being unsafe?

PR #7903, opened 2019-06-19
Remove expand home (`~`) by default in File.expand_path and Path#expand, now opt-in argument

didactic-drunk:
Path#absolute doesn't expand "~/" and may be used for secure path
resolution.

---

(final behavior)

`home` specifies the home directory which `~` will expand to.

* "~" is expanded to the value passed to `home`.
* If it is `false` (default), home is not expanded.
* If `true`, it is expanded to the user's home directory (`Path.home`).

### DIFF
--- a/spec/std/path_spec.cr
+++ b/spec/std/path_spec.cr
@@ -18,14 +18,6 @@ private def it_expands_path(path, posix, windows = posix, *, base = nil, env_hom
       ENV["HOME"] = env_home || (path.windows? ? HOME_WINDOWS : HOME_POSIX)
 
       base_arg = base || (path.windows? ? BASE_WINDOWS : BASE_POSIX)
-      home_arg = case home
-                 when Path   then home
-                 when String then home
-                 when Bool   then Path.home
-                   #       home ? Path.home : raise "invalid value 'false' for home"
-                 else raise "invalid type for home #{home}"
-                 end
-      #      puts "#{typeof(home_arg)} #{home_arg.class} #{home_arg.inspect} #{home.inspect}"
       path.expand(base_arg.not_nil!, expand_base: !!expand_base, home: home)
     ensure
       ENV["HOME"] = prev_home
@@ -552,9 +544,11 @@ describe Path do
       it_expands_path("C:\\foo", "D:/C:\\foo", "C:\\foo", base: "D:/")
     end
 
-    describe "doesn't expand ~" do
-      path = Path["~"]
-      path.expand.should_not eq path.expand(home: "/foo")
+    it "doesn't expand ~" do
+      ["~", "~/foo"].each do |path|
+        path = Path[path]
+        path.expand(base: "", expand_base: false).should eq path
+      end
     end
 
     describe "checks all possible types for expand(home:)" do
--- a/spec/std/path_spec.cr
+++ b/spec/std/path_spec.cr
@@ -10,7 +10,7 @@ private def it_normalizes_path(path, posix = path, windows = path, file = __FILE
   assert_paths(path, posix, windows, "normalizes", file, line, &.normalize)
 end
 
-private def it_expands_path(path, posix, windows = posix, *, base = nil, env_home = nil, expand_base = false, file = __FILE__, line = __LINE__)
+private def it_expands_path(path, posix, windows = posix, *, base = nil, env_home = nil, expand_base = false, home = false, file = __FILE__, line = __LINE__)
   assert_paths(path, posix, windows, %((base: "#{base}")), file, line) do |path|
     prev_home = ENV["HOME"]
 
@@ -18,7 +18,15 @@ private def it_expands_path(path, posix, windows = posix, *, base = nil, env_hom
       ENV["HOME"] = env_home || (path.windows? ? HOME_WINDOWS : HOME_POSIX)
 
       base_arg = base || (path.windows? ? BASE_WINDOWS : BASE_POSIX)
-      path.expand(base_arg.not_nil!, expand_base: !!expand_base)
+      home_arg = case home
+                 when Path   then home
+                 when String then home
+                 when Bool   then Path.home
+                   #       home ? Path.home : raise "invalid value 'false' for home"
+                 else raise "invalid type for home #{home}"
+                 end
+      #      puts "#{typeof(home_arg)} #{home_arg.class} #{home_arg.inspect} #{home.inspect}"
+      path.expand(base_arg.not_nil!, expand_base: !!expand_base, home: home)
     ensure
       ENV["HOME"] = prev_home
     end
@@ -544,34 +552,49 @@ describe Path do
       it_expands_path("C:\\foo", "D:/C:\\foo", "C:\\foo", base: "D:/")
     end
 
+    describe "doesn't expand ~" do
+      path = Path["~"]
+      path.expand.should_not eq path.expand(home: "/foo")
+    end
+
+    describe "checks all possible types for expand(home:)" do
+      home_posix2 = Path.posix(BASE_POSIX).join("foo").to_s
+      home_windows2 = Path.windows(BASE_WINDOWS).join("foo").to_s
+
+      home = Path[""].windows? ? home_windows2 : home_posix2
+      it_expands_path("~/a", {BASE_POSIX, "~/a"}, {BASE_WINDOWS, "~\\a"}, home: false)
+      it_expands_path("~/a", {home_posix2, "a"}, {home_windows2, "a"}, home: home)
+      it_expands_path("~/a", {home_posix2, "a"}, {home_windows2, "a"}, home: Path[home])
+    end
+
     describe "converts a pathname to an absolute pathname, using ~ (home) as base" do
-      it_expands_path("~/", {HOME_POSIX, ""}, {HOME_WINDOWS, ""})
-      it_expands_path("~/..badfilename", {HOME_POSIX, "..badfilename"}, {HOME_WINDOWS, "..badfilename"})
-      it_expands_path("..", "/default", "\\default")
-      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "~/b")
-      it_expands_path("~", HOME_POSIX, HOME_WINDOWS)
-      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, base: "/tmp/gumby/ddd")
-      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "/tmp/gumby/ddd")
+      it_expands_path("~/", {HOME_POSIX, ""}, {HOME_WINDOWS, ""}, home: true)
+      it_expands_path("~/..badfilename", {HOME_POSIX, "..badfilename"}, {HOME_WINDOWS, "..badfilename"}, home: true)
+      it_expands_path("..", "/default", "\\default", home: true)
+      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "~/b", home: true)
+      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, home: true)
+      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, base: "/tmp/gumby/ddd", home: true)
+      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "/tmp/gumby/ddd", home: true)
     end
 
     describe "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do
-      it_expands_path("~/", {HOME_POSIX, ""}, {HOME_WINDOWS, ""})
-      it_expands_path("~/..badfilename", {"#{HOME_POSIX}/", "..badfilename"}, {"#{HOME_WINDOWS}\\", "..badfilename"}, base: "")
-      it_expands_path("~/..", "/home", "C:\\Users")
-      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "~/b")
-      it_expands_path("~", HOME_POSIX, HOME_WINDOWS)
-      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, base: "/tmp/gumby/ddd")
-      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "/tmp/gumby/ddd")
+      it_expands_path("~/", {HOME_POSIX, ""}, {HOME_WINDOWS, ""}, home: true)
+      it_expands_path("~/..badfilename", {"#{HOME_POSIX}/", "..badfilename"}, {"#{HOME_WINDOWS}\\", "..badfilename"}, base: "", home: true)
+      it_expands_path("~/..", "/home", "C:\\Users", home: true)
+      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "~/b", home: true)
+      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, home: true)
+      it_expands_path("~", HOME_POSIX, HOME_WINDOWS, base: "/tmp/gumby/ddd", home: true)
+      it_expands_path("~/a", {HOME_POSIX, "a"}, {HOME_WINDOWS, "a"}, base: "/tmp/gumby/ddd", home: true)
     end
 
     describe "converts a pathname to an absolute pathname, using ~ (home) as base (HOME=/)" do
-      it_expands_path("~/", "/", "\\", env_home: "/")
-      it_expands_path("~/..badfilename", "/..badfilename", "\\..badfilename", env_home: "/")
-      it_expands_path("..", "/default", "\\default", env_home: "/")
-      it_expands_path("~/a", "/a", "\\a", base: "~/b", env_home: "/")
-      it_expands_path("~", "/", "\\", env_home: "/")
-      it_expands_path("~", "/", "\\", base: "/tmp/gumby/ddd", env_home: "/")
-      it_expands_path("~/a", "/a", "\\a", base: "/tmp/gumby/ddd", env_home: "/")
+      it_expands_path("~/", "/", "\\", env_home: "/", home: true)
+      it_expands_path("~/..badfilename", "/..badfilename", "\\..badfilename", env_home: "/", home: true)
+      it_expands_path("..", "/default", "\\default", env_home: "/", home: true)
+      it_expands_path("~/a", "/a", "\\a", base: "~/b", env_home: "/", home: true)
+      it_expands_path("~", "/", "\\", env_home: "/", home: true)
+      it_expands_path("~", "/", "\\", base: "/tmp/gumby/ddd", env_home: "/", home: true)
+      it_expands_path("~/a", "/a", "\\a", base: "/tmp/gumby/ddd", env_home: "/", home: true)
     end
 
     describe "ignores name starting with ~" do
--- a/src/file.cr
+++ b/src/file.cr
@@ -338,8 +338,9 @@ class File < IO::FileDescriptor
   # Converts *path* to an absolute path. Relative paths are
   # referenced from the current working directory of the process unless
   # *dir* is given, in which case it will be used as the starting point.
-  # "~/" is expanded to the value passed to *home* or the user's home directory
-  # when *home* is `true`.
+  # "~" is expanded to the value passed to *home*.
+  # If it is `false` (default), home is not expanded.
+  # If `true`, it is expanded to the user's home directory (`Path.home`).
   #
   # ```
   # File.expand_path("foo")                 # => "/home/.../foo"
--- a/src/file.cr
+++ b/src/file.cr
@@ -338,14 +338,16 @@ class File < IO::FileDescriptor
   # Converts *path* to an absolute path. Relative paths are
   # referenced from the current working directory of the process unless
   # *dir* is given, in which case it will be used as the starting point.
+  # "~/" is expanded to the value passed to *home* or the user's home directory
+  # when *home* is `true`.
   #
   # ```
-  # File.expand_path("foo")             # => "/home/.../foo"
-  # File.expand_path("~/crystal/foo")   # => "/home/crystal/foo"
-  # File.expand_path("baz", "/foo/bar") # => "/foo/bar/baz"
+  # File.expand_path("foo")                 # => "/home/.../foo"
+  # File.expand_path("~/foo", home: "/bar") # => "/bar/foo"
+  # File.expand_path("baz", "/foo/bar")     # => "/foo/bar/baz"
   # ```
-  def self.expand_path(path, dir = nil) : String
-    Path.new(path).expand(dir || Dir.current).to_s
+  def self.expand_path(path, dir = nil, *, home = false) : String
+    Path.new(path).expand(dir || Dir.current, home: home).to_s
   end
 
   class BadPatternError < Exception
--- a/src/path.cr
+++ b/src/path.cr
@@ -15,10 +15,10 @@
 # # Examples
 #
 # ```
-# Path["foo/bar/baz.cr"].parent   # => Path["foo/bar"]
-# Path["foo/bar/baz.cr"].basename # => "baz.cr"
-# Path["./foo/../bar"].normalize  # => Path["bar"]
-# Path["~/bin"].expand            # => Path["/home/crystal/bin"]
+# Path["foo/bar/baz.cr"].parent         # => Path["foo/bar"]
+# Path["foo/bar/baz.cr"].basename       # => "baz.cr"
+# Path["./foo/../bar"].normalize        # => Path["bar"]
+# Path["~/bin"].expand(home: Path.home) # => Path["/home/crystal/bin"]
 # ```
 #
 # For now, its methods are purely lexical, there is no direct filesystem access.
@@ -548,17 +548,19 @@ struct Path
   # unless *base* is given, in which case it will be used as the reference path.
   #
   # ```
-  # Path["foo"].expand             # => Path["/current/path/foo"]
-  # Path["~/crystal/foo"].expand   # => Path["/home/crystal/foo"]
-  # Path["baz"].expand("/foo/bar") # => Path["/foo/bar/baz"]
+  # Path["foo"].expand                 # => Path["/current/path/foo"]
+  # Path["~/foo"].expand(home: "/bar") # => Path["/bar/foo"]
+  # Path["baz"].expand("/foo/bar")     # => Path["/foo/bar/baz"]
   # ```
   #
-  # *home* specifies the home directory which `~` will expand to. If not given
-  # (or `nil` is given) then `Path.home` will be used.
+  # *home* specifies the home directory which `~` will expand to.
+  # If `true` is given then `Path.home` will be used.
+  # "~/" is expanded to the value passed to *home* or the user's home directory
+  # when *home* is `true`
   # If *expand_base* is `true`, *base* itself will be exanded in `Dir.current`
   # if it is not an absolute path. This guarantees the method returns an absolute
   # path (assuming that `Dir.current` is absolute).
-  def expand(base : Path | String = Dir.current, *, home : String | Path | Nil = nil, expand_base = true) : Path
+  def expand(base : Path | String = Dir.current, *, home : Path | String | Bool = false, expand_base = true) : Path
     base = Path.new(base) unless base.is_a?(Path)
     base = base.to_kind(@kind)
     if base == self
@@ -568,10 +570,12 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = (home || Path.home).to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = (home || Path.home).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    if home
+      if name == "~"
+        name = (resolve_home(home)).to_kind(@kind).normalize.to_s
+      elsif name.starts_with?("~/")
+        name = (resolve_home(home)).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+      end
     end
 
     unless new_instance(name).absolute?
@@ -623,6 +627,15 @@ struct Path
 
     expanded = new_instance(expanded) unless expanded.is_a?(Path)
     expanded.normalize(remove_final_separator: false)
+  end
+
+  private def resolve_home(home)
+    case home
+    when Path   then home
+    when String then Path[home]
+    when true   then Path.home
+    else             raise ArgumentError.new("Expected Path | String | Bool, not #{home.class}")
+    end
   end
 
   # Appends the given *part* to this path and returns the joined path.
--- a/src/path.cr
+++ b/src/path.cr
@@ -15,10 +15,10 @@
 # # Examples
 #
 # ```
-# Path["foo/bar/baz.cr"].parent         # => Path["foo/bar"]
-# Path["foo/bar/baz.cr"].basename       # => "baz.cr"
-# Path["./foo/../bar"].normalize        # => Path["bar"]
-# Path["~/bin"].expand(home: Path.home) # => Path["/home/crystal/bin"]
+# Path["foo/bar/baz.cr"].parent    # => Path["foo/bar"]
+# Path["foo/bar/baz.cr"].basename  # => "baz.cr"
+# Path["./foo/../bar"].normalize   # => Path["bar"]
+# Path["~/bin"].expand(home: true) # => Path["/home/crystal/bin"]
 # ```
 #
 # For now, its methods are purely lexical, there is no direct filesystem access.
@@ -554,10 +554,11 @@ struct Path
   # ```
   #
   # *home* specifies the home directory which `~` will expand to.
-  # If `true` is given then `Path.home` will be used.
-  # "~/" is expanded to the value passed to *home* or the user's home directory
-  # when *home* is `true`
-  # If *expand_base* is `true`, *base* itself will be exanded in `Dir.current`
+  # "~" is expanded to the value passed to *home*.
+  # If it is `false` (default), home is not expanded.
+  # If `true`, it is expanded to the user's home directory (`Path.home`).
+  #
+  # If *expand_base* is `true`, *base* itself will be expanded in `Dir.current`
   # if it is not an absolute path. This guarantees the method returns an absolute
   # path (assuming that `Dir.current` is absolute).
   def expand(base : Path | String = Dir.current, *, home : Path | String | Bool = false, expand_base = true) : Path
@@ -572,9 +573,9 @@ struct Path
 
     if home
       if name == "~"
-        name = (resolve_home(home)).to_kind(@kind).normalize.to_s
+        name = resolve_home(home).to_s
       elsif name.starts_with?("~/")
-        name = (resolve_home(home)).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+        name = resolve_home(home).join(name.byte_slice(2, name.bytesize - 2)).to_s
       end
     end
 
@@ -631,11 +632,11 @@ struct Path
 
   private def resolve_home(home)
     case home
-    when Path   then home
-    when String then Path[home]
-    when true   then Path.home
-    else             raise ArgumentError.new("Expected Path | String | Bool, not #{home.class}")
+    when String then home = Path[home]
+    when Bool   then home = Path.home
     end
+
+    home.to_kind(@kind).normalize
   end
 
   # Appends the given *part* to this path and returns the joined path.
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -582,26 +582,26 @@ describe "File" do
     end
 
     it "converts a pathname to an absolute pathname, using ~ (home) as base" do
-      File.expand_path("~/").should eq(File.join(home, ""))
-      File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-      File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(%r{\A//}, "/"))
-      File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
-      File.expand_path("~").should eq(home)
-      File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
-      File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
+      File.expand_path("~/", home: true).should eq(File.join(home, ""))
+      File.expand_path("~/..badfilename", home: true).should eq(File.join(home, "..badfilename"))
+      File.expand_path("..", home: true).should eq("/#{base.split('/')[0...-1].join('/')}".gsub(%r{\A//}, "/"))
+      File.expand_path("~/a", "~/b", home: true).should eq(File.join(home, "a"))
+      File.expand_path("~", home: true).should eq(home)
+      File.expand_path("~", "/tmp/gumby/ddd", home: true).should eq(home)
+      File.expand_path("~/a", "/tmp/gumby/ddd", home: true).should eq(File.join([home, "a"]))
     end
 
     it "converts a pathname to an absolute pathname, using ~ (home) as base (trailing /)" do
       prev_home = home
       begin
         ENV["HOME"] = File.expand_path(datapath)
-        File.expand_path("~/").should eq(File.join(home, ""))
-        File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(%r{\A//}, "/"))
-        File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
-        File.expand_path("~").should eq(home)
-        File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
-        File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
+        File.expand_path("~/", home: true).should eq(File.join(home, ""))
+        File.expand_path("~/..badfilename", home: true).should eq(File.join(home, "..badfilename"))
+        File.expand_path("..", home: true).should eq("/#{base.split('/')[0...-1].join('/')}".gsub(%r{\A//}, "/"))
+        File.expand_path("~/a", "~/b", home: true).should eq(File.join(home, "a"))
+        File.expand_path("~", home: true).should eq(home)
+        File.expand_path("~", "/tmp/gumby/ddd", home: true).should eq(home)
+        File.expand_path("~/a", "/tmp/gumby/ddd", home: true).should eq(File.join([home, "a"]))
       ensure
         ENV["HOME"] = prev_home
       end
@@ -611,13 +611,13 @@ describe "File" do
       prev_home = home
       begin
         ENV["HOME"] = "/"
-        File.expand_path("~/").should eq(home)
-        File.expand_path("~/..badfilename").should eq(File.join(home, "..badfilename"))
-        File.expand_path("..").should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
-        File.expand_path("~/a", "~/b").should eq(File.join(home, "a"))
-        File.expand_path("~").should eq(home)
-        File.expand_path("~", "/tmp/gumby/ddd").should eq(home)
-        File.expand_path("~/a", "/tmp/gumby/ddd").should eq(File.join([home, "a"]))
+        File.expand_path("~/", home: true).should eq(home)
+        File.expand_path("~/..badfilename", home: true).should eq(File.join(home, "..badfilename"))
+        File.expand_path("..", home: true).should eq("/#{base.split('/')[0...-1].join('/')}".gsub(/\A\/\//, "/"))
+        File.expand_path("~/a", "~/b", home: true).should eq(File.join(home, "a"))
+        File.expand_path("~", home: true).should eq(home)
+        File.expand_path("~", "/tmp/gumby/ddd", home: true).should eq(home)
+        File.expand_path("~/a", "/tmp/gumby/ddd", home: true).should eq(File.join([home, "a"]))
       ensure
         ENV["HOME"] = prev_home
       end


spec/std/path_spec.cr
```diff
@@ -10,6 +10,15 @@ private def it_normalizes_path(path, posix = path, windows = path, file = __FILE
   assert_paths(path, posix, windows, "normalizes", file, line, &.normalize)
 end
 
+private def it_absolutes_path(path, posix, windows = posix, *, base = nil, env_home = nil, expand_base = false, file = __FILE__, line = __LINE__)
+  assert_paths(path, posix, windows, %((base: "#{base}")), file, line) do |path|
+    begin
```
Sija: redundant `begin`

didactic-drunk: Fixed.

asterite:
I don't think I can relate `absolute` to what it does. In my mind, turning "foo" to absolute would mean it's "/foo".

Maybe we can add an optional argument to `expand` that is `true` by default and which specified whether to expand `~`.

But it's always better to discuss this is an issue, not a PR.

didactic-drunk:
@asterite Absolute is pretty standard and the normal term. 

* Ruby `File.absolute_path`
  Converts a pathname to an absolute pathname. Relative paths are referenced from the current working directory of the process unless dir_string is given, in which case it will be used as the starting point. If the given pathname starts with a “~” it is NOT expanded, it is treated as a normal directory name.
* Go `func Abs(path string) (string, error)`
  https://golang.org/pkg/path/filepath/#Abs
  Abs returns an absolute representation of path. If the path is not absolute it will be joined with the current working directory to turn it into an absolute path. The absolute path name for a given file is not guaranteed to be unique. Abs calls Clean on the result.
* Java `getAbsolutePath()`
  This file path method returns the absolute path of the file. If File is created with absolute pathname, it simply returns the pathname.
  If the file object is created using a relative path, the absolute pathname is resolved in a system-dependent way. On UNIX systems, a relative pathname is made absolute by resolving it against the current user directory.
* C++ `std::filesystem::absolute`
  Returns a path referencing the same file system location as p, for which is_absolute() is true
* D `absolutePath("some/file", "/foo/bar")`
* Nim had a PR to add `absolutePath` and already has `isAbsolute`.
* Elixir `absname(path)`
  Converts the given path to an absolute one. Unlike expand/1, no attempt is made to resolve .., . or ~.
* Python `os.path.abspath(path)`
  Return a normalized absolutized version of the pathname path. On most platforms, this is equivalent to calling the function normpath() as follows: normpath(join(os.getcwd(), path)).

C# is an outlier calling the function `GetFullPath`.

Almost all implementations have the 2nd arg as the root directory to allow for caching of `Dir.current` equivalents.

[Absolute path](https://trends.google.com/trends/explore?date=today%205-y&geo=US&q=absolute%20path) is what people search for. Not [fullpath](https://trends.google.com/trends/explore?date=today%205-y&geo=US&q=full%20path).

And finally `absolute(path)` matches `absolute?` in Crystal.

asterite:
I think the outlier is actually correct. Absolute seems like unix baggage. Just my opinion.

Sija:
@asterite Seeing `full_path` makes me think that `~` should be expanded. `absolute` doesn't convey such notion, thus is closer to the given implementation. just my 2c.

didactic-drunk:
@asterite Perhaps I didn't write enough about the problems with `expand`. 

Most application probably **want** `absolute` and **NOT** `expand` for most uses except finding user configuration files.  See #7768 where misuse of `expand` accidentally broke an application.  He didn't want `expand`.  He wanted `absolute` but no API provided the necessary function.

There are also serious security implications for any application using `expand` if an attacker can choose file names or the application calls expand on any filename outside of it's own control.

Depending on the application  `expand` can be used to escape out of a specified file structure to gain read or write access to other files.
```shell
mkdir -p expand/~/.ssh
touch expand/~/.ssh/authorized_keys
```
```crystal
Dir.cd "expand"
Dir.glob("*/*/*") do |file| # Not sure why ** isn't working.
#Dir.new(".").children.each do |file|
        path = Path[file]
        puts "path #{path} expanded to #{path.expand}"
end
```
Output:
```
path ~/.ssh/authorized_keys expanded to /Users/test/.ssh/authorized_keys
```

If anything expand should work the other way around and **not** expand ~ by default.  Or provide `absolute` along with `expand` letting the developer choose the appropriate method.

asterite:
Okay, disregard my comments.

src/file.cr
```diff
@@ -338,6 +338,19 @@ class File < IO::FileDescriptor
   # *dir* is given, in which case it will be used as the starting point.
   #
   # ```
+  # File.absolute_path("foo")             # => "/home/.../foo"
+  # File.absolute_path("baz", "/foo/bar") # => "/foo/bar/baz"
+  # ```
+  def self.absolute_path(path, dir = nil) : String
```
RX14: *If* we have reached a consensus that the `File...` methods will be going away in favour of `Path` then perhaps we shouldn't add this.

didactic-drunk:
All of the code reviews were addressed.  Is there anything else needed to get this PR approved?

asterite:
Someone needs to approve it. I won't because I'm not too familiar with Path's API so you'll have to wait for someone else. But don't worry, I'm sure this will be merged.

src/path.cr
```diff
@@ -540,6 +540,18 @@ struct Path
     end
   end
 
+  # Converts this path to an absolute path. Relative paths are
+  # referenced from the current working directory of the process (`Dir.current`)
+  # unless *base* is given, in which case it will be used as the reference path.
```
RX14: There doesn't seem to be enough difference in the first paragraph of these doc comments. You should probably add the 

>"~/" is expanded to the current user's home directory.

in `#expand`'s docs

straight-shoota:
The described behaviour is not different from `expand` as evidenced by the implementation. The only difference is an option to disable home path resolution in `expand`. This seems like a good addition for `expand` anyway.

But I'm not convinced we need a separate method for this when it's just a simple delegate.
We could however debate whether `home` argument should perhaps be `nil` by default. 

The return value of `Path#expand` is an absolute path with the only exception when `expand_base` is intentionally set to `false`. 

> There are also serious security implications for any application using expand if an attacker can choose file names or the application calls expand on any filename outside of it's own control.
> Depending on the application expand can be used to escape out of a specified file structure to gain read or write access to other files.

This is completely unrelated. `#expand` is the same as `#absolute`. Whether you expand `~` to a user's home directory doesn't add any more security implications than expanding a series of `../` to parent directories.

If you want to restrict the resulting path to a specific path prefix, you need a different solution anyway.

didactic-drunk:
> But I'm not convinced we need a separate method for this when it's just a simple delegate.
We could however debate whether home argument should perhaps be nil by default.

Great.  Tell me which will be accepted and I'll rework this PR.  Home argument on or off by default? 

Finding the absolute path is a more common operation than home directory lookups.  Perhaps off should be the default?

> This is completely unrelated. `#expand` is the same as `#absolute`. Whether you expand `~` to a user's home directory doesn't add any more security implications than expanding a series of `../` to parent directories.

Incorrect.  You are describing sanitizing user input which is a different problem.  I am describing taking data returned from the file system and turning a relative path in to an absolute path. A file can't be named `..` for this trick to work with `Dir`.

See my post above on using `glob` with `expand` for an example.  `Dir.new` gives similar results.

straight-shoota:
I don't follow what you describe as a security issue regarding `expand`. Could you elaborate?

didactic-drunk:
Let's say you make a file watching application which watches an input dir and performs actions on any file saved there.  Maybe it's a javascript minifier, compressor, decompressor, encryption etc service.  It works as follows:
* Save a file in an input directory.
* A process chdir's to "input" and periodically polls "./" using Dir.glob or Dir.entries to get a recursive list of files.
* The process uses File.expand on data returned from Dir.* expecting it will only return an absolute path.  Instead it is redirected to another file completely.
* The watching process may need the expanded path if another application takes part in job processing (when using a queue) or for notification purposes.

```
mkdir -p input/~/.ssh
touch input/~/.ssh/authorized_keys
```

```
Dir.cd "input"
Dir.glob("*/*/*") do |file| # Not sure why ** isn't working.
#Dir.new(".").children.each do |file|
        path = Path[file]
        puts "path #{path} expanded to #{path.expand}"
end
```

Output:
```
path ~/.ssh/authorized_keys expanded to /Users/test/.ssh/authorized_keys
```

Another example is #7768 which was almost hit by this misfeature.  Depending on how he wrote his application the files he downloaded may have read or overwritten any files under his home directory including .bashrc or .ssh/authorized_keys.  The files he downloaded were in the form of `~foo` which was addressed in the PR, but this does not take in to account intentional malicious acts of creating file structures to exploit the current behavior.  If crystal gains popularity I would not expect the situation to improve if applications are already creating exploitable conditions.  Wait until torrents start showing up with "~/.bashrc" in the file name.

The default is not safe and there's no current standardized way to get a **safe** absolute path to a file.

straight-shoota:
Okay, I see. This is about the ambiguity of a `~` as first character of a path name.

I agree there needs to be a way to expand a path without `~` expansion.
A solution for that would be to make `home` parameter nilable. As said before, we could consider making it `nil` by default, which would be more strict and avoid surprises. You'd have to opt in for that.

There is unfortunately no reliable way to escape a `~` character. In a shell you could use a backslash to escape it, but such a path would not work outside a shell, so this should not be included in any return value.
Since the expansion only considers the very first character, adding a `./` prefix would sufficiently escape the `~` and make it unambiguously refer to a path in the working directory. But this would not be a single path component, but actually a path of two components. And `#normalize` would blow it away; there could be a special case for this, but this only adds layers of complexity and still carries some rough edges.

It seems fine to me when you can toggle `~` expansion:
* When dealing with paths from user input, you can enable `~` to offer this convenience feature. In order to specify a path starting with `~`, it should be fine to insert the path with a `./` prefix.
* When dealing with paths read from the file system, you would disable `~` because it's not necessary anyway.

Alternatively, we could also remove `~` expansion entirely and leave that for a shard to implement (which could also cover more complex variants like `~foo`).
It seems Go, Rust, Node.js don't have this in their stdlib. So it can work without it. However, given this is a pretty commonly used feature, I think it makes sense to have it available in stdlib.

straight-shoota:
I think I wouldn't accept a `Bool` parameter for `home`, just `Path | Nil` should suffice. Calling it with `home: Path.home` is not much more effort than `home: true` and is more explicit about what it does.

But don't implement that right away. I'd like to hear some thoughts from others as well.

didactic-drunk:
> But don't implement that right away. I'd like to hear some thoughts from others as well.

Too late.

src/path.cr
```diff
@@ -564,15 +564,27 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = home.to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = home.to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    home = case home
+           when Path, String
```
RX14: `when Path` should be `home`, and `when String` should be `Path.new(string)`, surely?

RX14:
None of the specs pass in `home` as a `String` or a `Path`.

spec/std/path_spec.cr
```diff
@@ -10,21 +10,26 @@ private def it_normalizes_path(path, posix = path, windows = path, file = __FILE
...
-      path.expand(base_arg.not_nil!, expand_base: !!expand_base)
+      #      home = Path.home if home == true
+      path.expand(base_arg.not_nil!, expand_base: !!expand_base, home: home)
     ensure
       ENV["HOME"] = prev_home
     end
   end
 end
 
+private def it_expands_path_with_home(path, posix, windows = posix, *, base = nil, env_home = nil, expand_base = false, file = __FILE__, line = __LINE__)
```
straight-shoota: Does this need to be a separate helper method? The difference at the callsite is `_with_home` vs. adding a `, home: true` argument.

src/path.cr
```diff
@@ -545,16 +545,16 @@ struct Path
...
   # ```
-  # Path["foo"].expand             # => Path["/current/path/foo"]
-  # Path["~/crystal/foo"].expand   # => Path["/home/crystal/foo"]
-  # Path["baz"].expand("/foo/bar") # => Path["/foo/bar/baz"]
+  # Path["foo"].expand                       # => Path["/current/path/foo"]
+  # Path["~/crystal/foo"].expand(home: true) # => Path["/home/crystal/foo"]
+  # Path["baz"].expand("/foo/bar")           # => Path["/foo/bar/baz"]
   # ```
   #
   # *home* specifies the home directory which `~` will expand to.
```
straight-shoota: The documentation needs to be extended for the new behaviour.

src/file.cr
```diff
@@ -336,14 +336,15 @@ class File < IO::FileDescriptor
   # Converts *path* to an absolute path. Relative paths are
   # referenced from the current working directory of the process unless
   # *dir* is given, in which case it will be used as the starting point.
+  # "~/" is expanded to the current user's home directory.
```
straight-shoota: This is not correct (by default).

src/path.cr
```diff
@@ -564,15 +564,30 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = home.to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = home.to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    home = case home
+           when Path
```
Sija: `when x then y` one-line style would make this whole `case` bit more succinct.

spec/std/path_spec.cr
didactic-drunk: Can the `Bool` argument for home stay?  If not I need to understand why the commented line in the review I posted below causes spec failures.

didactic-drunk: Also, the original documentation for home expansion was incorrect.

src/path.cr
```diff
@@ -564,15 +566,25 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = home.to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = home.to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    home = case home
+           when Path      ; home
```
```suggestion
           when Path then home
```

ditto below

didactic-drunk: I didn't know that was a keyword.

src/path.cr
```diff
@@ -545,16 +545,18 @@ struct Path
...
   # ```
   #
-  # *home* specifies the home directory which `~` will expand to.
+  # "~/" is expanded to the value passed to *home* or the user's home directory
+  # when *home* is `true`
   # If *expand_base* is `true`, *base* itself will be exanded in `Dir.current`
   # if it is not an absolute path. This guarantees the method returns an absolute
   # path (assuming that `Dir.current` is absolute).
-  def expand(base : Path | String = Dir.current, *, home = Path.home, expand_base = true) : Path
+  def expand(base : Path | String = Dir.current, *, home = nil, expand_base = true) : Path
```
straight-shoota: I'd suggest to only accept `Path | String | Nil` for `home`.
Opinions?

Sija: Why not allowing `Bool` as a shorthand as well?

didactic-drunk: `true` is the most common and likely use and allows for future `~username/foo` expansion.

straight-shoota: I'd question whether we even want/need `~username` expansion at all. But that is a different discussion. As is how this would be integrated into the API. We don't need to consider that *right now*. Expanding the type to accept `Bool` later is easy and doesn't break anything. We should only consider if `Bool` makes sense at the present.

@Sija `Path.home` is more explict, see https://github.com/crystal-lang/crystal/pull/7903#issuecomment-515983270 
Also, it avoids doubts about the difference between `home: nil` and `home: false`.

didactic-drunk: To remove of the `Bool` argument would anyone be willing to look at the spec failures after adding ```home = Path.home if home == true``` to `src/spec/std_path.cr`?

asterite: @didactic-drunk Which failure? There's at least one `home: true` in the code.

Also, `it_expands_path` should call `it "..." do ... end`, otherwise the "tests" don't run inside an "it" block and they are not properly reported. Same goes for `it_normalizes_path`.

asterite: Though I guess there's some old code there that's incorrect and wasn't written by you? I'm not sure...

asterite: It's also easier if you add a type restriction to the `home` arguments with the types you want to allow, then you'll get a compile error where you are passing a `Bool`.

didactic-drunk: > Which failure? There's at least one `home: true` in the code.

Line 22 `src/std/path_spec.cr` .

didactic-drunk: I plan on adding type restrictions after resolving this issue.

didactic-drunk: > @didactic-drunk Which failure? There's at least one `home: true` in the code.
> 
> Also, `it_expands_path` should call `it "..." do ... end`, otherwise the "tests" don't run inside an "it" block and they are not properly reported. Same goes for `it_normalizes_path`.

Most of the path specs weren't written by me.

Are `it` blocks in `assert_paths_raw` sufficient?

straight-shoota: @asterite The `it` spec blocks are in `assert_paths_raw`. That method is probably misnamed, but everything works as expected.

didactic-drunk: When line 22 is uncommented all home directory specs fail. I don't understand why.  The output shows bizarre half windows paths:
```
       Expected: Path.windows("C:\\Users\\Crystal\\a")
            got: Path.windows("\\home\\crystal\\a")
```

asterite:
I can send a PR to fix the existing specs, then we can work on top of that.

straight-shoota:
@asterite The specs are not broken.

asterite:
Ah, you are right. It's just that some calls happen before the `it` block in this PR, that's what confused me.

didactic-drunk:
@asterite Comment line 22 in the specs and the specs succeed.  Uncomment and they fail even though the same type conversion happens in a different method.

asterite:
The problem is that `assert_paths` ends up calling `assert_paths_raw` which ends up calling the block twice. The first time `home` is changed and then the second time it has the value from after the first block was called, not from the original argument.

It works fine if you do:

```crystal
      home_arg = home
      home_arg = Path.home if home_arg == true
      path.expand(base_arg.not_nil!, expand_base: !!expand_base, home: home_arg)
```

didactic-drunk:
@asterite Type restrictions were added as you suggested, but I can't get rid of the Bool type with a case statement.

asterite:
Sorry, I don't know what that means, but with type restrictions you just have to fix the errors the compiler tells you.

didactic-drunk:
> Sorry, I don't know what that means, but with type restrictions you just have to fix the errors the compiler tells you.

Shouldn't the following always produce a `Path | String | Nil`?
```
      home_arg = case home
                  when Path   then home
                  when String then home
                  when Nil    then home
                  when Bool   then Path.home
                    #       home ? Path.home : raise "invalid value 'false' for home"
                  else raise "invalid type for home #{home}"
                  end
```

Instead it says home is `(Bool | Path)`:
```
in spec/std/path_spec.cr:30: no overload matches 'Path#expand' with types String, expand_base: Bool, home: (Bool | Path)
Overloads are:
 - Path#expand(base : Path | String = Dir.current, *, home : Path | String | Nil = nil, expand_base = true)
Couldn't find overloads for these types:
 - Path#expand(base : String, expand_base : Bool, home : Bool)

      path.expand(base_arg.not_nil!, expand_base: !!expand_base, home: home_arg)
```

asterite:
Well, everything inside `assert_paths` end up happening inside an `it` block which basically closures over `home`. What that happens you can't restrict the type of `home` because it could be accessed concurrently by another fiber (hypothetically).

In this case you can do something like:

```crystal
      home2 = home
      home_arg = case home2
                 when Path   then home2
                 when String then home2
                 when Nil    then nil
                 when Bool   then Path.home
                   #       home ? Path.home : raise "invalid value 'false' for home"
                 else raise "invalid type for home #{home2}"
                 end
```

One might argue that this is a problem in the language. I personally think that spec code has become way too complex. I mean, why do we transform `home` for the `Bool` case if it's not supported by the API? If we remove it then we can always pass `home` directly.

straight-shoota:
> If we remove it then we can always pass home directly.

Yes, please!

didactic-drunk:
`Path.home` is different on Windows and POSIX.  The HOME environment variable is set depending on which is being tested.  Passing in `Path.home` would lead to spec failures.

didactic-drunk:
> I mean, why do we transform `home` for the `Bool` case if it's not supported by the API? If we remove it then we can always pass `home` directly.

So we don't have to pass it twice, once for POSIX and a 2nd time for Windows.  Should I keep it the way it is with or double the number of test cases by copy and pasting code for each one manually?

straight-shoota:
Since #8128 was merged, `nil` value for `home` argument now means to lazily call `Path.home` only when needed. So I guess we'll need to use a Bool after all in order to signal deactivation of home path expansion. But then we could probably remove the `nil` value because it would be equivalent to `false`.

So `home` could either be:
* `String | Path`: specific value
* `true`: default value `Path.home`
* `false`: don't expand home path

didactic-drunk:
> So `home` could be one of the following:
> 
>     * `String | Path`: specific value
> 
>     * `true`: default value `Path.home`
> 
>     * `false`: don't expand home path

Pushed.

src/path.cr
```diff
@@ -568,10 +570,20 @@ struct Path
...
-    if name == "~"
-      name = (home || Path.home).to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = (home || Path.home).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    if home
+      home = case home
+             when Path   then home
+             when String then Path[home]
+             when true   then Path.home
+             when false  then nil
```
straight-shoota: This condition can't match because `home` is already ensured to be truthy.

src/path.cr
```diff
@@ -568,10 +570,20 @@ struct Path
...
+             else             raise ArgumentError.new("Expected Path | String | Bool, not #{home.class}")
```
straight-shoota: This branch should be unreachable because of the argument's type restriction.

src/path.cr
```diff
@@ -568,10 +570,20 @@ struct Path
...
+             end
+
+      if name == "~"
+        name = (home || Path.home).to_kind(@kind).normalize.to_s
```
straight-shoota:
```suggestion
        name = home.to_kind(@kind).normalize.to_s
```

src/path.cr
didactic-drunk: I botched the merge slightly.  See current code where (home || Path.home) is still in use.  I have a use case where a program may use `path.expand(home: true)` and not want `Path.home` resolved.

```
download_dir = settings_from_user_editable_config_file["downloads"] || "~/Downloads"
Path[download_dir].expand(home: true)
```

"downloads" could be "/Volumes/some_drive" or /Users/Shared/Downloads".  There are many similar use cases and I think some of them quite common for intended use such as user editable configuration settings, command line flags or possibly environment variables.  I recognize the shell may expand `~/` by default but that's not where the difference is likely to be.

Example where user supplied Path expansion used by default but unlikely from a command line argument.
```
foo_path = "~/.foo"
# Path.home shouldn't run based on #8128
option_parser.on("--foo-path=[PATH]) { |arg| foo_path = arg }
Path[foo_path].expand(home: true)
```


Dropping `(home || Path.home)` within the if block would bring back #8128

straight-shoota: True, I forgot about the lazy evaluation. I suppose it would probably be worth to extract `home` resolution in a helper method and use that in the name branches.
```cr
private def resolve_home(home)
  case home
    when String then home = Path[home]
    when true   then home = Path.home
  end
  home.to_kind(@kind).normalize
end
```
This reads clearer than the current implementation, it centralizes the handling and removes code duplication.

spec/std/path_spec.cr
```diff
@@ -544,34 +552,49 @@ describe Path do
       it_expands_path("C:\\foo", "D:/C:\\foo", "C:\\foo", base: "D:/")
     end
 
+    describe "doesn't expand ~" do
+      path = Path["~"]
+      path.expand.should eq path
```
didactic-drunk: @straight-shoota I don't think the review was correct.  The test makes sure `path.expand` without a home is **not** the same as with a home.  Also the spec fails.  

It's the last commit.  Should I undo it?
```
-      path.expand.should_not eq path.expand(home: "/foo")
+      path.expand.should eq path
```

src/path.cr
```diff
@@ -568,10 +570,12 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = (home || Path.home).to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = (home || Path.home).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    if home
+      if name == "~"
+        name = (resolve_home(home)).to_kind(@kind).normalize.to_s
```
```suggestion
        name = resolve_home(home).to_kind(@kind).normalize.to_s
```

src/path.cr
```diff
@@ -568,10 +570,12 @@ struct Path
...
+      elsif name.starts_with?("~/")
+        name = (resolve_home(home)).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
```
```suggestion
        name = resolve_home(home).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
```

spec/std/path_spec.cr
straight-shoota: This spec is supposed to make sure that a `~` component will be left untouched by `#expand`.
In order to spec this correctly, we need to remove base expansion:
```cr
path.expand(base: "", expand_base: false).should eq path
```
And please also add a test for `Path["~/foo"]`

src/path.cr
```diff
@@ -568,10 +570,12 @@ struct Path
 
     name = @name
 
-    if name == "~"
-      name = (home || Path.home).to_kind(@kind).normalize.to_s
-    elsif name.starts_with?("~/")
-      name = (home || Path.home).to_kind(@kind).normalize.join(name.byte_slice(2, name.bytesize - 2)).to_s
+    if home
+      if name == "~"
+        name = resolve_home(home).to_kind(@kind).normalize.to_s
```
straight-shoota: `.to_kind(@kind).normalize` should be moved into the helper.

spec/std/path_spec.cr
```diff
@@ -544,34 +544,51 @@ describe Path do
       it_expands_path("C:\\foo", "D:/C:\\foo", "C:\\foo", base: "D:/")
     end
 
+    describe "doesn't expand ~" do
```
straight-shoota:
```suggestion
    it "doesn't expand ~" do
```

src/path.cr
straight-shoota: Could you please update the docs to the same phrasing as `File.expand_path`?

src/path.cr
```diff
@@ -625,6 +629,17 @@ struct Path
     expanded.normalize(remove_final_separator: false)
   end
 
+  private def resolve_home(home)
+    home = case home
+           when Path   then home
+           when String then Path[home]
+           when true   then Path.home
+           else             raise ArgumentError.new("Expected Path | String | Bool, not #{home.class}")
```
straight-shoota: This else branch should not raise `ArgumentError` because the method is not callable directly and from the calls in `#expand`, the only possible argument values are either `Path | String` or `true`.

So I'd suggest to either `raise "unreachable"` or remove the entire else branch and replace `true` with `Bool`, because for any boolean value (which in this case can only be `true`), it should use `Path.home`.

straight-shoota: Also IMO it's a bit more concise to put the assign inside the branches. You can even leave out the `Path` branch.  This requires to omit the `else` branch by matching against `Bool`.

```cr
case home
  when String then home = Path[home]
  when Bool   then home = Path.home
end
```
I'd strongly prefer this notation, but it's not mandatory.

didactic-drunk: `else raise` is to catch future code change changes causing unexpected behavior.  Are you sure you want them gone?

asterite:
Is it intentional that with this PR `FIle.expand_path("~/foo")` doesn't work like before? That is, "~" isn't expanded anymore unless you pass `home: true`.

straight-shoota:
Hm, I thought we had discussed to change the default behaviour in the course of this discussion. But it seems, this was never really decided. So it's up for debate.
I can live with both. Personally, I'd probably prefer `home: true` as default.

asterite:
I was confused because if you read this PR from top to bottom the main description is this:

> Path#absolute doesn't expand "~/" and may be used for secure path resolution.

So I thought this was just adding a new method that didn't expand "~" by default, leaving `expand` untouched.

I also think we should be passing `home: true` to `File.expand_path`. First because otherwise it's a huge breaking change, and a silent one. Second, it works like that in Ruby too (less important, but still...).

asterite:
Oooh... there's no `Path#absolute` anymore. I guess I'm a bit lost. I would have preferred a separate issue to discussing this, and a PR that just has the code for what we agree in the separate issue. Otherwise it's really hard to follow what's going on.

Otherwise, let's just make `home: true` by default and that's it.

didactic-drunk:
@asterite I thought I was very clear in my [third comment](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-504211199) that the default is not safe and should be changed **or** add a separate method `#absolute` ([my 2nd comment](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-504181405)) so the programmer can choose the correct method call.

@asterite I mentioned you at the top in both comments.

@asterite If you think default `~/` expansion is safe then what happens when less skilled programmer uses`#expand` when they really want `#absolute`?  Arbitrary files get read or overwritten.  It's not theoretical, it already happened in #7768.  He got lucky because the payload wasn't malicious.  See my [third comment](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-504211199) for an example of a malicious payload you can run yourself.

@asterite @straight-shoota I think the clarity of a separate API call is warranted given the security implications especially since they search for absolute and will end up with a secure method by default. 

Not wanting to break backwards compatibility silently also favors a separate method.  

Also you gain compatibility with ruby which has two separate methods.  As do several other programming languages.

asterite:
The thing is, the only times I used expand. in Ruby scripts and programs was because I wanted to expand the home directory. I don't know of that is true for others. So at least I find it really un intuitive that the only use case I had suddenly becomes. broken and I have to type more code. But I'm probably wrong.

straight-shoota:
I don't think there is a compelling reason to have both `#absolute` and `#expand` when they do almost exactly the same thing. It would be more difficult to recall the difference between them than understanding the difference between `expand(path, home: true)` and `expand(path, home: false)`.
It doesn't make sense in my opinion to add a `#absolute` method when we already have `#expand`.

But we can discuss whether home expansion should be enabled by default or not. As I understand this entire PR, that's what you're actually concerned about.

I can't see any significant security implications. `expand` just expands a path in a specific path context.

The only implication for expanding `~` by default is - as you described in https://github.com/crystal-lang/crystal/pull/7903#issuecomment-515638594 - that a relative path starting with a component named `~` can't be accessed, because it points to the user's home directory. Example:
```shell
> mkdir \~
> echo bar > \~/foo
> tree
.
└── ~
    └── foo
> crystal eval 'puts File.read(Path["~/foo"].expand)'
Error opening file '/home/johannes/foo' with mode 'r': No such file or directory (Errno)
```

This is a *super rare* edge case where the convenience feature of expanding `~` by default would lead to unexpected behaviour. And it can be fixed simply by prefixing the path with `./` so the `~` is no longer the first character.
```cr
> crystal eval 'puts File.read(Path["./~/foo"].expand)'
bar
```
So, I don't think this is a valid reason to make `home: false` be default.

#7768 was clearly a bug. And I'm certain we don't want to expand `~username` to `username`'s home directory by default (if that would be implemented at all), given that it is much more common that a file or folder name starts with `~`.

didactic-drunk:
> I don't think there is a compelling reason to have both `#absolute` and `#expand` when they do almost exactly the same thing. It would be more difficult to recall the difference between them than understanding the difference
> between `expand(path, home: true)` and `expand(path, home: false)`.

Python, Ruby, and Go all have `absolute path` and a second `expand path` function.  Go's is expand implementation is third party.

Python
```
os.path.abspath(path)
  Return a normalized absolutized version of the pathname path. On most platforms, this is equivalent to calling the function normpath() as follows: normpath(join(os.getcwd(), path)).

os.path.expanduser(path)
  On Unix and Windows, return the argument with an initial component of ~ or ~user replaced by that user’s home directory.

```

Ruby
```
# If the given pathname starts with a “~'' it is NOT expanded, it is treated as a normal directory name.
File.absolute_path("~oracle/bin")       #=> "<relative_path>/~oracle/bin"
# The given pathname may start with a “~'', which expands to the process owner's home directory (the environment variable HOME must be set correctly). “~user'' expands to the named user's home directory.
File.expand_path("~oracle/bin")           #=> "/home/oracle/bin"

```

Go
```
func Abs(path string) (string, error)
homedir.Expand() to expand the ~ in a path to the home directory.
```

didactic-drunk:
On a related note I saw an opportunity to improve the `Time` API based on @straight-shoota's various reviews stating to prefer arguments with defaults over two methods that are almost identical.

So I changed the following:
```
Time.parse(time : String, pattern : String, location : Location) : Time
Time.parse!(time : String, pattern : String) : Time
```
To
```
Time.parse(time : String, pattern : String, location : Location = Time::Location::UTC) : Time
```
Which matches existing formatting methods.
```
src/time/format/custom/iso_8601.c
def self.parse(string, location : Time::Location? = Time::Location::UTC)
```

After giving myself a pat on the back for this tiny improvement I checked the history wondering why someone would do this.   It must be old code, but no.
```
Author: Johannes Müller <x>
Date:   Wed Aug 1 14:28:14 2018 +0200

    Add location as required argument for Time.parse (#6369)
    
    * Add location as required argument for Time.parse
    
    This makes it more clear in the API that a default value is required instead of defaulting to `nil` which would raise in case of time zone information missing from the formatted string. `Time.parse_utc` and `Time.parse_local` are added as short cuts to the most common use cases.

```

I think you were attempting to reduce bugs for programmer errors when timezones were or weren't specified and provided an API that made the intent clear rather than having a single default argument that could potentially be misused.  


I think we're in the same boat.  Gathering statistics from my locally installed ruby gems the API is heavily misused.
* 16 calls to `File.absolute_path`
* 51 calls to `File.expand_path` with `~`
1014 calls to `File.expand_path` with a fixed string no `~` (Incorrect usage, they wanted `absolute_path`)
211 calls to `File.expand_path` with `File.dirname` (Incorrect uage, they wanted `absolute_path`)
3 calls to `File.expand_path` referencing home or user.
495 other calls, most of which appear to want `absolute`.

Weirdest use: `filename = File.absolute_path(File.expand_path(filename))`

didactic-drunk:
I'm willing to admit I was wrong.  Giving people a choice between `absolute_path` and `expand_path` doesn't lead to correct usage.
@asterite was incorrect in thinking expanding home is the normal use case.  In ruby it's measurably not.  From google searches stats it's not.  My personal experience shows `absolute_path` is way more common than expanding home.  The types of software he and I develop may be different but a quick grep of python or ruby will show home expansion is an outlier.

Considering the following:
* `expand_path` in ruby is mostly misused even with an `absolute_path` option.
* A programmer has already misused `expand_path` to create a potential security hole in a crystal  program when crystal isn't 1.0 yet.  Expect more with more users.
* Silently breaking backwards compatibility is unwanted by changing the `expand_path` default.

Can we add clearer API that shows programmer intent and is easy for programmers to see when it's misused?
* Add `absolute_path` since it is the most common use case.
* Add `expand_user` (python) or `expand_home` for the **rare** home expansion cases and to improve clarity as @straight-shoota did for the `Time` API.
* deprecate `expand_path` as the use is ambigious and not safe.

didactic-drunk:
> This is a _super rare_ edge case 

[10201](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=buffer+overflow) buffer overflows were super rare edge cases.  Crystal has buffer size checking to stop those sorts of problems.  Why bother if it's rare?

[22](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=%2Ftmp+race) tmp race flaws were even more rare edge cases.  Yet crystal attempts to create `Tempfile`'s with random unpredictable numbers.

TLS server name validation?  That's not even an edge case.  It requires an active attacker redirecting traffic.  There's an entire file `src/openssl/ssl/hostname_validation.cr` dedicated to fixing it.

All security holes are super rare edge cases.  Until someone writes an exploit and makes it self replicate.


> And it can be fixed simply by prefixing the path with `./` so the `~` is no longer the first character.

How will programmers without security training to notice subtle flaws **and** come up with a correct solution that can't be bypassed?  Btw that is not how I would fix it and not how most other people **with** training would fix it either.  Most people are taught to use input validation.  Anyone  following advice to reject "begins with /", contains some variant of "/." or contains "\0" would think they were safe when they're not because the expand API has an unsafe default.

> where the convenience feature of expanding `~` by default would lead to unexpected behaviour. 

Convienence vs security?  In this case it's not one or the other.  You can have both.  See solution [one comment up](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-528936113).

straight-shoota:
I don't like your argumentation citing a completely different problem and suggesting that it would need to lead to exactly the same solution. It is not. When parsing time, there is no sound default value, it's either no default location (which raises when the formatted string doesn't include one) or you need to specify one explicitly.
When expanding the home path, the question is only whether it should be enabled by default or not (and maybe if there should be different methods for this).

And I don't like you're insisting that expanding `~` by default is a security flaw. I'm all in for improving security and it should obviously win over convenience. But here I can't see any significant security implications.

As I understand it, this is not about a trade-off between convenience and security. We're talking about convenience for a relatively common use-case (path referencing home directory) vs. a super rare use case (a path component consisting of only a `~` at the beginning of a path).

Let's face it that way: With a single method `Path#expand, there is little chance to get it wrong compared to having two separate methods like in Ruby etc. This one method with default `home: true` should be good for almost any use case. IMO the drawback when you don't need or want home expansion is negligible.

didactic-drunk:
@asterite You said you always used it for home expanion.  Where exactly?  Every instance I looked at uses `expand_path`to get an absolute path with 2 exceptions.

Home expansion is absolutely the outlier.

Absolute only (not thorougly checked)
```
src/compiler/crystal/codegen/cache_dir.cr
src/compiler/crystal/command/cursor.cr 
src/compiler/crystal/command/spec.cr File.expand_path("./src")
src/compiler/crystal/command/spec.cr File.expand_path("spec")
src/compiler/crystal/command.cr        
src/compiler/crystal/compiler.cr
src/http/server/handlers/static_file_handler.cr
src/process/executable_path.cr
src/spec/dsl.c
```

Isn't this absolute_path twice?
```
src/compiler/crystal/crystal_path.cr
    private def make_relative_unless_absolute(filename)
      filename = "#{Dir.current}/#{filename}" unless filename.starts_with?('/')
      File.expand_path(filename)
    end
```

Uncertain
```
src/spec/dsl.c

```

didactic-drunk:
> I think I wouldn't accept a `Bool` parameter for `home`, just `Path | Nil` should suffice. Calling it with `home: Path.home` is not much more effort than `home: true` and is more explicit about what it does.

There is an obvious option that has eluded us.  Path expansion is either finding an absolute path or finding a file/dir relative to the user's home that's often fixed.  In the few hundred uses I've looked at in the last few hours it's never been both.

@straight-shoota You were ok with an explicit argument for
```path.expand("foo", home: Path.home)```

Why not
```Path.home.join "foo"```

It's both shorter and more explicit.

Separate home expansion from absolute resolution.
```
path.absolute
Path.home.join ".crystal"
```

For some reason they're combined in to a single method for an extremely rare use case that I can't find in ruby or crystal.  But maybe you have a code example?

**Comment by Sija** [comment]
> ``
> 
> Why not
> `Path.home.join "foo"`

Because you couldn't easily parse it from a string intended to be expanded, like `~/Downloads` for instance.

**Comment by straight-shoota** [comment]
> Why not `Path.home.join "foo"`

`join` is completely wrong behaviour, when the path (`"foo"`) is not under your control. It may be a relative path, an absolute path or a path relative to the home directory. Joining it to `Path.home` is only correct in the second case. The others need a different behaviour. And to figure this all out there is `Process#expand`.

> For some reason they're combined in to a single method for an extremely rare use case that I can't find in ruby or crystal. But maybe you have a code example?

The use case exists for dealing with user-provided data (for example: configuration values). If you only look at source code references, yes, you'll naturally find pretty few cases, where this behaviour is necessary.

**Comment by asterite** [comment]
I think I wouldn't mind `File.expand_path` to have `home: false` by default now, if it implies more security by default.

**Comment by didactic-drunk** [comment]
> I can't see any significant security implications. `expand` just expands a path in a specific path context.

Code the following:
1. Unarchive a file from the internet.
2. Dir.cd "unpacked_dir"
3a. Start a http server serving files that uses `expand_path`.  You've exposed every file in your home directory including your ssh keys.
3b. Process files rewriting them in some way using `expand_path`.  You may have rewritten your .bash_profile, .ssh/authorized_keys or any other file in your home dir.

Spam the following on torrent sites.:
4. File structures with a payload of ~/.ssh/bash_profile etc.

#7768 did 1. and 2.  I couldn't find his published code.  Maybe he didn't make it to to step 3.  Maybe he did.

**Comment by didactic-drunk** [comment]
> > Why not `Path.home.join "foo"`
> 
> @didactic-drunk Because you couldn't easily parse it from a string intended to be expanded, like `~/Downloads` for instance.

@Sija I get that, but the use is so rare.  It's not in crystal, not in any of the ruby code I've read, and not obvious in any I've grepped.

**Comment by straight-shoota** [comment]
Expanding `~/.ssh/private_key` is typically identical to expanding something like `../../../../../../../../../../../../../../../../home/username/.ssh/private_key`. You'd only need to guess the username, which shouldn't be hard. Just because `#expand` wouldn't expand your home directory doesn't make your secrets safe or prevents overwriting files in your home dir.

Whether you're dealing with paths using `#expand` or `#absolute`, whether `home` is true or false by default: Ensuring access is restricted to allowed file path in a specific context is a separate task.

**Comment by Sija** [comment]
> @Sija I get that, but the use is so rare. It's not in crystal, not in any of the ruby code I've read, and not obvious in any I've grepped.

It's pretty standard way to expand configurable paths, and of course you won't find it in any language source code because its usecases are almost found entirely in _userland_, not stdlib.

**Comment by didactic-drunk** [comment]
> Expanding `~/.ssh/private_key` is typically identical to expanding something like `../../../../../../../../../../../../../../../../home/username/.ssh/private_key`. 

1. Normal filename sanitization would catch starts with `/`, `.` or contains `./`.  It's the most common documented solution.  It won't catch starts with `~`.  `~` is a valid file name character. I have 59 files in my homedir that contain `~`.
2. **You can't store '../' In a filename.**  Walking the file system in most languages is generally a safe operation and may catch people by surprise when `expand` parses `~`.  
3. The example provided in #7768 may be two different programs.  He may shell to `7z` for extraction which contains path traversal sanitization.  Then he's caught by 2. thinking everything was safe.

My point is what's normally safe under almost all circumstances can catch people by surprise leading to security holes.  Yes it's rare, but so are race conditions.

**Comment by didactic-drunk** [comment]
> > @Sija I get that, but the use is so rare. It's not in crystal, not in any of the ruby code I've read, and not obvious in any I've grepped.
> 
> It's pretty standard way to expand configurable paths, and of course you won't find it in any language source code because its usecases are almost found entirely in _userland_, not stdlib.

You're right.  If this topic continues I'll look at the applications more closely.

Maybe renaming to `expand_user` or `expand_home` with a comment detailing when to us `absolute` would cause less confusion and security holes?

**Comment by didactic-drunk** [comment]
> The use case exists for dealing with user-provided data (for example: configuration values). If you only look at source code references, yes, you'll naturally find pretty few cases, where this behaviour is necessary.

Why would you keep the unsafe default for a rare use case? The common case is non-expansion of home.

Let the tiny fraction of applications that need it specify an additional argument, rather than the majority of applications specifying `home: false`.

Or split the API.  `absolute` with `expand_user` or `expand_home`.  If when to use which is well documented and they get it wrong it's their fault.

**Comment by straight-shoota** [comment]
I'm sorry, I don't follow what your mean.

> 1. Normal filename sanitization would catch starts with `/`, `.` or contains `./`.  It's the most common documented solution.  It won't catch starts with `~`.  `~` is a valid file name character. I have 59 files in my homedir that contain `~`.

What is "normal filename sanitization"? If a filename just contains `~` it doesn't matter. `home: true` only makes a difference when the entire filename is just `~`.

> 2. **You can't store '../' In a filename.**  Walking the file system in most languages is generally a safe operation and may catch people by surprise when `expand` parses `~`.  

Just to smart off: `..` is a valid filename as well. On regular file systems it just points to the parent path.
Walking the file system can't be generally safe unless you have complete control over it. A completely static expansion method can't provide that. `Path#expand` doesn't have any connection to the filesystem the path are used for. 

> 3. The example provided in #7768 may be two different programs.  He may shell to `7z` for extraction which contains path traversal sanitization.  Then he's caught by 2. thinking everything was safe..

No idea what you mean with that. #7768 was a bug and it's fixed.

> Why would you keep the unsafe default for a rare use case? The common case is non-expansion of home.

Because it's not unsafe and the use case is not really rare. In fact, it's probably a few magnitudes more common than having a filename consisting of only `~`. Every program that uses user-configurable local paths benefits.

**Comment by straight-shoota** [comment]
> Or split the API.  `absolute` with `expand_user` or `expand_home`.  If when to use which is well documented and they get it wrong it's their fault.

That might be an option: having separate `#expand_home` and `#absolute` methods with different behaviour. In order to get the current behaviour of `path.expand`, you'd need to chain them `path.expand_home.absolute`. That's certainly more explicit. But I'm not convinced it's better when you can just have it in a single method without significant downsides.

**Comment by asterite** [comment]
I didn't follow the entire conversation, but maybe we could have home be false by default and try it for a. version or two. If it's more secure by default then it might be fine. It's just a matter of adding home: true.

I'm just thinking, maybe someone makes an app that asks the user for a filename. I need to remember to pass home: true because the user might reference the home directory, right? It's just that, that one has to remember to enable it when asking for user input.

**Comment by didactic-drunk** [comment]
>  I need to remember to pass home: true because the user might reference the home directory, right? It's just that, that one has to remember to enable it when asking for user input.

@asterite It's not a choice of one case having to remember or a secure default. 
It's a choice between which of the 2 use cases has to remember and if there's a secure default or not.  Also the secure default is more common.

Here are the use cases.  One of them has to pass an argument depending which default you choose.  

### User uses File.expand("~/.config/...")

* Based on statistics posted above this use is extremely rare.  Less than 1%.
* If they forget to pass `home: true` they will probably figure out quickly during development because their config file loading is broken. 
* No security issues.

### User uses File.expand on arbitrary file system input in an arbitrary directory or on arbitrary input. 

* 99% of the time File.expand is used to get absolute path.  Every use in the crystal compiler confirms this.
* To be safe 99/100 times **they must remember to pass `home: true`**.
* The security hole is silent and won't be caught during development unless they accidentally stumble upon it like in #7768.
* Months or years later many people are exploited.

### Summary
`expand with home: false (default)`
* Safe default
* 1% of users need to remember to pass something or it breaks quickly and visibly to the developer.

`expand with home: true (default)`
* Unsafe default
* 99% of users need to remember to pass something or it can cause silent security issues that developers are very unlikely to catch before it gets to users.  
* #7768 was only caught by a user.  Not any crystal tests.  #7768 is the tip of the iceberg.  Fixing it did not close *any* security problems.

**Comment by straight-shoota** [comment]
@jgaskins You keep insisting there is a security issue and the default behaviour would be unsafe. I can't see any proof of that.

It's just a matter of which expectations users have about how the code behaves. Do they expect that path starting with `~/` expand to the home directory or not? Every time a path comes directly or indirectly from user input, it's a nice convenience to have. Most other cases might not explicitly need it, but it doesn't hurt either. And there might be few situations where you explicitly want to disable home expansion.

**Comment by jgaskins** [comment]
@straight-shoota I've insisted a lot of things but I assure you this one's not me. :-)

**Comment by didactic-drunk** [comment]
> > 1. Normal filename sanitization would catch starts with `/`, `.` or contains `./`.  It's the most common documented solution.  It won't catch starts with `~`.  `~` is a valid file name character. I have 59 files in my homedir that contain `~`.
> 
> What is "normal filename sanitization"? 

Rejecting common methods to escape a restricted file system path.  A partial list is at the top of this comment.

>If a filename just contains `~` it doesn't matter. `home: true` only makes a difference when the entire filename is just `~`.

It's a problem when a **directory** contains a bare `~`  Then you can escape

  It makes a difference when an application is running but the filenames are chosen by another input source.  The input could be an extracted zip file.  Or network drive.  Or upload path for a web service.  Or many others.

My main problems are
1. The API is mostly misused which leads me to think it's a bad design, mostly a bad default.
2. The default is also insecure.  If it was secure I wouldn't care.
3. The results of `expand` are surprising and not on anyone's radar for file name sanitization. `~` is not something that normally needs sanitization.  Until I saw #7768 I didn't know to look for it.  Other security auditors don't know to look for it either.

**Comment by didactic-drunk** [comment]
> @jgaskins You keep insisting there is a security issue and the default behaviour would be unsafe. I can't see any proof of that.
> 
> It's just a matter of which expectations users have about how the code behaves. Do they expect that path starting with `~/` expand to the home directory or not? Every time a path comes directly or indirectly from user input, it's a nice convenience to have. Most other cases might not explicitly need it, but it doesn't hurt either. 

Pass an existing path from `Dir.entries` to `File.absolute_path` on ruby.  You dont' need to sanitize anything.
Do the same for `os.path.abspath(..)` in python.  No issues.
Do the same for `filepath.Abs()` in go.  No issues.
Do the same for `path.resolve()` in nodejs.  No issues.
Do the same for `path.expand()` in crystal. **Now you need to audit for security problems with every single use**.  That far outweights any benefits of having `home: true` as a default.

**Comment by straight-shoota** [comment]
> Rejecting common methods to escape a restricted file system path. A partial list is at the top of this comment.

I'm still not following.

> It's a problem when a directory contains a bare ~ Then you can escape

No, there is only a problem when a directory (or file, for that matter) is named exactly `~`. When `~` is somewhere in the name (even at the beginning), it is not recognized for home expansion.

> Do the same for path.expand() in crystal. Now you need to audit for security problems with every single use.

Why would you need to do that?

**Comment by didactic-drunk** [comment]
> You keep insisting there is a security issue and the default behaviour would be unsafe. I can't see any proof of that.

Examples were provided in this PR [here](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-515638594) and [here](https://github.com/crystal-lang/crystal/pull/7903#issuecomment-504211199).  You can run the code examples.  Some fill in the blanks are required.  

For data processing try this:
```
unzip "attacker_controlled_file"
Dir.chdir "unzipped_dir" do
  Dir.glob("**/**") do |file| # Not sure why ** isn't working.
    # Pretend to process some files in the package
    # Cleanup
    File.delete path.expand # Boom.
  end
end
```

On ruby/python/go/nodejs using `abspath` this would be safe.  On crystal it deletes any files in your home directory.

@asterite See above line.

straight-shoota:
Anyway, this discussion is far too long already, especially considering this is a pull request. This should really have started as an issue, not directly a PR.
I'm sure nobody will be able to follow the course of this discussion and thus there won't be any decision.

So, I'd suggest to close this PR at this point. Continuing here leads nowhere.

Let's start with an issue detailing the issue and possible solutions.

As a personal favour, I'd ask you to read your texts after writing them and consider whether the argumentation would be comprehensible for someone else.

Sija:
@straight-shoota I don't think it's fair to state it leads nowhere. IMO OP has a point regarding security implications of leaving home expansion enabled from default - which can be avoided by simply turning it off from default - as suggested by @asterite. That would be a breaking change but for vast majority of users un-noticible since they wouldn't use it anyway. All the rest can get back to the previous behaviour just by "flipping the switch", so no harm is done.

straight-shoota:
@Sija I'm not saying that the argument isn't justified. It's just that this discussion is far too long for anyone to follow. That's why it's a dead end.

@didactic-drunk Thanks for your last example! I now understand that you're talking about a pathname beginning with `~` as a result from an method like `Dir.glob`. The good thing about these values is, that we can control it. So, it would be feasible to escape a `~`-only name.

didactic-drunk:
> So, it would be feasible to escape a `~`-only name.

Do you mean escape results from Dir.glob?  But Dir.glob results can go other places like printing, filenames for Zip, http/ftp/sftp/etc uploads that don't understand the escaped format.

straight-shoota:
I was thinking about some kind of internal escaping that's only relevant when used for `expand` or similar. But yeah, that probably adds too much complexity for little gain.

src/path.cr
didactic-drunk: `#expand` contains documentation on when and what `home` is expanded to.

RX14:
Would be nice to have `openat2` wrappers, when `openat2` lands in the linux kernel, which allows sandboxing path lookups beneath a given dirfd *correctly*, ignoring magic links like `/proc/`, symlinks, paths which use `..`, etc. Obviously needs a more expensive fallback for older kernel versions to be useful.

Something like `dir.open(path, :beneath) #=> File`. This would solve all filesystem escapes automatically.

RX14:
I guess this can be merged for `0.32.0`?

asterite:
Maybe. It's a breaking change and I'd like feedback from @waj before merging it.

RX14:
0.32.0 is the release after next

straight-shoota:
@asterite The breaking change is probably not a huge issue because `Path` is relatively recent addition and not well integrated yet, anyway.

straight-shoota:
Never mind. `File.expand_path` changes behaviour as well.

didactic-drunk:
If `File.expand_path` is scheduled for removal and `Path` is new why not keep home expansion on by default for `expand_path` and remove it later.  Combatibility issues are minimized that way. The security issues will sort themselves when `expand_path` is removed.

straight-shoota:
I'd avoid having different behaviour.

didactic-drunk:
Is there anything for me to do?

bcardiff:
@didactic-drunk May I ask for a rebase on master? After that, if we have a couple of confirmation of the reviews (since it's been a while) we are good to go.

I think the other needed thing is to update the PR title and opening comment to reflect the final state of the PR.

didactic-drunk:
@bcardiff Rebased.

asterite:
Actually, maybe not that many apps will fail. Only those with a hardcoded "~/" in their code will, because I guess if you want to pass that as a command line argument the shell will expand it for you first. The other possibility is that these paths are in a config file but I'm not sure how likely is that.

straight-shoota:
I think this title makes it more clear that this is a breaking change. The main change here is not adding an argument, it's changing a feature that was previously enabled by default.

PR #7908, opened 2019-06-20
HTTP::Server: don't double-close response

### LINKED ISSUE #7832: HTTP::Server Leftovers after chunking
## Summary
Server response will have leftovers. (Transfer-Encoding)
I can't find the source of the error. :(

## Example
* Server
```crystal
require "http/client"

  puts [context.request]

  context.response.headers.add "Transfer-Encoding", "chunked"
  context.response << String.new Bytes[10, 13]
  context.response.flush
  context.response.close
end

tcp = server.bind_tcp "0.0.0.0", 8888
puts "Listening on #{tcp}"
server.listen
```

* Curl Leftovers
```text
$ curl -v 127.0.0.1:8888 --raw -k
* Rebuilt URL to: 127.0.0.1:8888/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 1087 (#0)
> GET http://127.0.0.1:8888/ HTTP/1.1
> Host: 127.0.0.1:8888
> User-Agent: curl/7.54.0
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 200 OK
< Connection: keep-alive
< Transfer-Encoding: chunked
< Proxy-Connection: keep-alive
< 
2


0

0

* Leftovers after chunking: 5 bytes
* Connection #0 to host 127.0.0.1 left intact
```

asterite:
Fixes #7832

Sorry this has no tests but it's really hard to test.

### DIFF
--- a/src/http/server/response.cr
+++ b/src/http/server/response.cr
@@ -113,6 +113,8 @@ class HTTP::Server
     # Closes this response, writing headers and body if not done yet.
     # This method must be implemented if wrapping the response output.
     def close
+      return if closed?
+
       @output.close
     end
 
@@ -204,6 +206,8 @@ class HTTP::Server
       end
 
       def close
+        return if closed?
+
         unless response.wrote_headers?
           response.content_length = @out_count
         end


wontruefree:
This seems like a good candidate for a test.

wontruefree:
I cannot make a pr into your branch.  But I added one spec. 
https://github.com/crystal-lang/crystal/compare/master...wontruefree:bug/http-server-leftover

asterite:
Does that spec passes if you remove my diff?

asterite:
@wontruefree The spec passes even without my check. It's really hard to test it.

wontruefree:
Maybe it is not worth writing a spec for.
I just thought it would be nice to have covered.

Issue #7910, opened 2019-06-20
Use same naming scheme for OAuth::AccessToken and OAuth2::AccessToken


watzon:
Currently `OAuth::AccessToken` and `OAuth2::AccessToken` have a different method to access the actual token value, with the former being `token` and the latter being `access_token`. It would make more sense for `OAuth2::AccessToken` to also have the parameter be named `token` as it's less verbose and makes the naming more consistent.

asterite:
I think I used the names mentioned in the RFC, but I wouldn't mind making the APIs more consistent.

asterite:
I take it back: I don't think we need a different kind of consistency here. The current names match those in the RFCs.

PR #7911, opened 2019-06-21
rename access_token to token

wontruefree:
This renames `OAuth2::AccessToken#access_token` to `OAuth2::AccessToken#token`.  I also added some depreciation warnings for `OAuth2::AccessToken#access_token`

fixes https://github.com/crystal-lang/crystal/issues/7910

### DIFF
--- a/src/oauth2/access_token/access_token.cr
+++ b/src/oauth2/access_token/access_token.cr
@@ -59,12 +59,6 @@ abstract class OAuth2::AccessToken
   # will be the string "[1,2,3]".
   property extra : Hash(String, String)?
 
-  @[Deprecated("Use `token` instead of `access_token`")]
-  def initialize(access_token : String, expires_in : Int?, @refresh_token : String? = nil, @scope : String? = nil, @extra = nil)
-    @token = access_token
-    @expires_in = expires_in.try &.to_i64
-  end
-
   def initialize(@token : String, expires_in : Int?, @refresh_token : String? = nil, @scope : String? = nil, @extra = nil)
     @expires_in = expires_in.try &.to_i64
   end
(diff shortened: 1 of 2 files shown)


src/oauth2/access_token/access_token.cr
```diff
@@ -54,7 +59,13 @@ abstract class OAuth2::AccessToken
   # will be the string "[1,2,3]".
   property extra : Hash(String, String)?
 
-  def initialize(@access_token : String, expires_in : Int?, @refresh_token : String? = nil, @scope : String? = nil, @extra = nil)
+  @[Deprecated("Use `token` instead of `access_token`")]
+  def initialize(access_token : String, expires_in : Int?, @refresh_token : String? = nil, @scope : String? = nil, @extra = nil)
```
asterite: This doesn't work. The arguments are exactly the same as the next method. and so this will get overwritten and it wi. never show up in docs. Just delete this overload, there's no other way to do it.

wontruefree: @asterite I removed this can you review this again?

asterite:
Ooooh... but the JSON property in the spec is called `access_token`.

Now I don't think this is needed, sorry. It's just a cosmetic change and it deviates from the OAuth RFC/Spec. Plus it introduces a breaking change. I don't see why we need to do this.

But if others agree this is fine, let's go for it.

wontruefree:
@asterite I thought `OAuth2::AccessToken#access_token` was redundant.  I Think `#value` might make more sense.

asterite:
@wontruefree Sorry about getting into this, but now I don't think it's worth introducing a breaking change, less if the names won't match the ones in the RFC.

wontruefree:
@asterite no problem I was just trying to help with some low hanging fruit from issues.

PR #7917, opened 2019-06-23
number: Add `#positive?` and `#negative?`

woodruffw:
Adds `#positive?` and `#negative?` to Crystal's `Number`.

~Will add unit tests in a moment.~ Done.

### DIFF
--- a/spec/std/number_spec.cr
+++ b/spec/std/number_spec.cr
@@ -231,6 +231,30 @@ describe "Number" do
     1f32.zero?.should eq false
   end
 
+  it "test positive?" do
+    0.positive?.should eq false
+    0.0.positive?.should eq false
+    0f32.positive?.should eq false
+    1.positive?.should eq true
+    1.0.positive?.should eq true
+    1f32.positive?.should eq true
+    -1.positive?.should eq false
+    -1.0.positive?.should eq false
+    -1f32.positive?.should eq false
+  end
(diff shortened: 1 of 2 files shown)


j8r:
Duplicate of https://github.com/crystal-lang/crystal/pull/6552

woodruffw:
To summarize my argument for this (on Gitter):

1. Crystal's `Number` represents real numbers, and the definitions of "positive" and "negative" are settled for real numbers.
2. `positive?` and `negative?` make for much nicer reading in conditions and blocks and are consistent with "say what you mean" semantics (one of the things that makes writing Crystal from a background in Ruby so pleasant)
3. `Number` already defines both `sign` and `zero?`, accepting the settled definition of `sgn(R)`. The addition of `positive?` and `negative?` is consistent with the current semantics.

j8r:
This isn't really the same as `#zero?`, because it isn't only defined in `Number`:
- [Complex#zero?](https://crystal-lang.org/api/master/Complex.html#zero%3F%3ABool-instance-method)
- [Time::Span#zero?](https://crystal-lang.org/api/master/Time/Span.html#zero?:Bool-instance-method)

Even if `Number#zero?` isn't that necessary, it's an already known method name.

asterite:
Sorry, duplicate of something we rejected in the past.

Issue #7920, opened 2019-06-24
Enumerable#tally


kachick:
Feature request
===

How about adding `Enumerable#tally`? Ruby has it.

https://bugs.ruby-lang.org/issues/11076
https://github.com/ruby/spec/pull/668

asterite:
I think this could be a good addition, it's pretty common to want to count things in a list.

asterite:
Closed by #7921

PR #7921, opened 2019-06-24
Implement Enumerable#tally

kachick:
For https://github.com/crystal-lang/crystal/issues/7920

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1391,8 +1391,12 @@ module Enumerable(T)
   # ```
   def tally
     each_with_object(Hash(T, Int32).new) do |item, hash|
-      hash[item] ||= 0
-      hash[item] += 1
+      count = hash[item]?
+      if count
+        hash[item] = count + 1
+      else
+        hash[item] = 1
+      end
     end
   end
 
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1389,7 +1389,7 @@ module Enumerable(T)
   # ```
   # %w[1 2 3 3 3 2].tally # => { "1" => 1, "2" => 2, "3" => 3 }
   # ```
-  def tally
+  def tally : Hash(T, Int32)
     each_with_object(Hash(T, Int32).new) do |item, hash|
       count = hash[item]?
       if count
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1384,10 +1384,12 @@ module Enumerable(T)
     result
   end
 
-  # Returns a `Hash` with counts the elements in the collection.
+  # Tallys the collection.  Returns a hash where the keys are the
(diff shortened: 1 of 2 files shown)


src/enumerable.cr
```diff
@@ -1384,6 +1384,18 @@ module Enumerable(T)
...
   end
 
+  # Returns a `Hash` with counts the elements in the collection.
+  #
+  # ```
+  # %w[1 2 3 3 3 2].tally # => { "1" => 1, "2" => 2, "3" => 3 }
+  # ```
+  def tally
+    each_with_object(Hash(T, Int32).new) do |item, hash|
+      hash[item] ||= 0
```
asterite: I would do:

```
count = hash[item]?
if count
  hash[item] = count + 1
else
  hash[item] = 1
end
```

That should be a bit more efficient to avoid multiple lookups in `hash`.

wontruefree:
`tally` seems a little odd compared to the other method names.  What about `count_by`?

Sija:
@wontruefree `*_by` suffix is typically used for methods accepting block - see `Enumerable#{group,index}_by` as an example. Given its purpose `tally` IMO is a pretty sensible name - short, descriptive and apt for the occasion, I don't see anything wrong with it.

src/enumerable.cr
kachick: Thank you! Updated with your one https://github.com/crystal-lang/crystal/pull/7921/commits/c3e69663b2e579868f01030794051e130c71d945

wontruefree:
@sija makes sense to me

src/enumerable.cr
```diff
@@ -1384,6 +1384,22 @@ module Enumerable(T)
     result
   end
 
+  # Returns a `Hash` with counts the elements in the collection.
```
asterite: What do you think about using the same docs as Ruby:

https://github.com/ruby/ruby/blob/trunk/enum.c#L1004-L1013

src/enumerable.cr
```diff
@@ -1384,6 +1384,22 @@ module Enumerable(T)
...
+  #
+  # ```
+  # %w[1 2 3 3 3 2].tally # => { "1" => 1, "2" => 2, "3" => 3 }
+  # ```
+  def tally
```
asterite:
```suggestion
  def tally : Hash(T, Int32)
```

This makes it a bit more explicit what are the types in the hash.

src/enumerable.cr
kachick: Sounds good 👍 Applied in https://github.com/crystal-lang/crystal/pull/7921/commits/49d9c6880ce0578eea3b17249130b239f5acf660

kachick:
Thank you!

PR #5042, opened 2017-09-26
Cookie expiration

bmmcginty:
As mentioned in #5033, this PR adds consistent handling for the *expires* parameter for cookies. It changes #expires to #expiration_time and allows #expires to act as a property for the *expires* parameter. It also adds #max_age to allow access to the *max-age* parameter. Finally, documentation has been added for #expiration_time.

PR #7925, opened 2019-06-26
Proper handling of max-age and expires for cookies

watzon:
This PR is based off of #5042. Seeing as it has been sitting there for some time with no movement, and the current Cookie parsing is breaking with large `max-age` values, I decided to get to ball rolling again.


spec/std/http/cookie_spec.cr
```diff
@@ -194,36 +194,80 @@ module HTTP
 
       it "parse max-age as seconds from current time" do
```
Sija: I guess this description should follow-up with your changes.

watzon: Good catch. I've updated it.

Sija: Actually I thought about the `from current time` part ;)

watzon: Oops, yeah you're right :joy:

watzon:
It looks like the failures are due to a memory leak of some kind, any idea what could be causing it?

asterite:
I think compiler specs are leaking. I just noticed that codegen specs that don't use the prelude are allocating memory using `Pointer.malloc` or `Foo.new` which end up calling `LibC.malloc`and that memory is never freed.

However, specs usually don't allocate that much memory in tests so maybe there's something else going on... but it's hard to debug. Plus it only happens on linux 32, mostly.

watzon:
It's weird. I can't even run the full library specs on my machine or it eats up all my RAM. I've had to force shutdown my laptop twice. The cookie specs themselves pass though.

src/http/cookie.cr
```diff
@@ -53,9 +58,20 @@ module HTTP
...
+      else
+        @expires
+      end
+    end
+
+    # Returns the expiration status of this cookie as a `Bool` given a creation time
+    def expired?(time_reference = @creation_time)
+      if @max_age == 0.seconds
+        true
+      elsif (time = expiration_time(time_reference)) && time < Time.now
```
asterite: Maybe use `Time.utc_now` to avoid fetching location information? Not sure.

src/http/cookie.cr
```diff
@@ -25,23 +26,27 @@ module HTTP
...
       expires = @expires
+      max_age = @max_age
       domain = @domain
       samesite = @samesite
       String.build do |header|
         header << "#{URI.escape @name}=#{URI.escape value}"
         header << "; domain=#{domain}" if domain
         header << "; path=#{path}" if path
         header << "; expires=#{HTTP.format_time(expires)}" if expires
+        header << "; max-age=#{max_age.total_seconds}" if max_age
```
asterite: I think `total_seconds` it a `Float64`. Does `max-age` allow float values in the RFC?

watzon: According to the RFC, Max-Age has to be a non-zero digit. It doesn't say anything about Float vs Int though.

Blacksmoke16: https://tools.ietf.org/html/rfc6265#section-5.2.2

Mentions:

```
   Let delta-seconds be the attribute-value converted to an integer.

   If delta-seconds is less than or equal to zero (0), let expiry-time
   be the earliest representable date and time.  Otherwise, let the
   expiry-time be the current date and time plus delta-seconds seconds.

   Append an attribute to the cookie-attribute-list with an attribute-
   name of Max-Age and an attribute-value of expiry-time.
```
so looks like it should be converted to an int.

Sija: Seems like `max_age.to_i` (number of full seconds) should be enough.

watzon: At the very least I feel like it should be an `Int64`. I have run into cases where Max-Age is too large and can't be parsed into an `Int32`.

Sija: @watzon [`Time::Span#to_i`](https://crystal-lang.org/api/0.29.0/Time/Span.html#to_i%3AInt64-instance-method) returns an `Int64`, so no problem here.

watzon: Oh it does? Well awesome then

watzon: Done

src/http/cookie.cr
```diff
@@ -25,23 +26,27 @@ module HTTP
...
       expires = @expires
+      max_age = @max_age
       domain = @domain
       samesite = @samesite
       String.build do |header|
         header << "#{URI.escape @name}=#{URI.escape value}"
         header << "; domain=#{domain}" if domain
         header << "; path=#{path}" if path
         header << "; expires=#{HTTP.format_time(expires)}" if expires
+        header << "; max-age=#{max_age.total_seconds.to_i}" if max_age
```
Blacksmoke16:
```suggestion
        header << "; max-age=#{max_age.to_i}" if max_age
```

src/http/cookie.cr
watzon: Good question..

watzon:
Holy moley, the specs actually passed

watzon:
Looks like we just need to get @straight-shoota in here for review

Sija:
@watzon Two approvals is what's needed (would gr8 to have @straight-shoota review nevertheless).

RX14:
I'd like another approval of the changes made

spec/std/http/cookie_spec.cr
```diff
@@ -192,38 +192,82 @@ module HTTP
...
-        delta.should be > 0.seconds
-        delta.should be < 1.seconds
+        cookie.max_age.should eq 0.seconds
+      end
+    end
+
+    describe "expiration_time" do
+      it "sets expiration_time to be current when max-age=0" do
+        cookie = parse_set_cookie("bla=1; max-age=0")
+        (Time.now - cookie.expiration_time.not_nil!).should be <= 1.seconds
```
straight-shoota: `Time.now` is deprecated. You should use `Time.utc` **everywhere** because time zone doesn't matter here.

Also I think this spec is a bit more concise:
```suggestion
        cookie.expiration_time.not_nil!.should be_close(Time.utc, 1.seconds)
```

spec/std/http/cookie_spec.cr
```diff
@@ -192,38 +192,82 @@ module HTTP
...
+
+      it "sets future expiration_time with max-age" do
+        cookie = parse_set_cookie("bla=1; max-age=1")
+        (cookie.expiration_time.not_nil!).should be > Time.now
+      end
+
+      it "sets future expiration_time with max-age and future cookie creation time" do
+        cookie = parse_set_cookie("bla=1; max-age=1")
+        cookie_expiration = cookie.expiration_time(time_reference: Time.now + 20.seconds).not_nil!
+        (Time.now + 20.seconds).should be < cookie_expiration
```
straight-shoota: When using a specific reference time, please store it in a variable like `time_reference = Time.utc + 20.seconds`. This makes this spec much better readable.

Also, like above, `be_close` would be a good fit:
```suggestion
        cookie_expiration.should be_close(time_reference, 1.second)
```

spec/std/http/cookie_spec.cr
```diff
@@ -192,38 +192,82 @@ module HTTP
...
+        (Time.now + 21.seconds).should be >= cookie_expiration
+      end
+
+      it "sets future expiration_time with expires" do
+        cookie = parse_set_cookie("bla=1; expires=Thu, 01 Jan 2020 00:00:00 -0000")
+        (cookie.expiration_time.not_nil! >= Time.utc(2020, 1, 1, 0, 0, 0)).should be_true
```
straight-shoota: This should test equality because the exact time is explicitly specified:

```suggestion
        cookie.expiration_time.not_nil!.should eq Time.utc(2020, 1, 1, 0, 0, 0)
```

spec/std/http/cookie_spec.cr
```diff
@@ -192,38 +192,82 @@ module HTTP
...
-      it "not expired" do
-        parse_set_cookie("bla=1; max-age=1").expired?.should eq false
+      it "not expired with future expires" do
+        cookie = parse_set_cookie("bla=1; expires=Thu, 01 Jan 2020 00:00:00 -0000")
+        cookie.expired?.should be_false
       end
 
-      it "not expired" do
-        parse_set_cookie("bla=1; expires=Thu, 01 Jan 2020 00:00:00 -0000").expired?.should eq false
+      it "sets past expiration_time with max-age and future time reference" do
```
straight-shoota: This spec seems to be mostly a duplication of `sets future expiration_time with max-age and future cookie creation time`, except for the last line.

src/http/cookie.cr
```diff
@@ -25,23 +26,27 @@ module HTTP
...
 
     def initialize(@name : String, value : String, @path : String = "/",
-                   @expires : Time? = nil, @domain : String? = nil,
-                   @secure : Bool = false, @http_only : Bool = false,
-                   @samesite : SameSite? = nil, @extension : String? = nil)
+                   @expires : Time? = nil, @max_age : Time::Span? = nil,
+                   @domain : String? = nil, @secure : Bool = false,
+                   @http_only : Bool = false, @samesite : SameSite? = nil,
+                   @extension : String? = nil)
+      @creation_time = Time.now
```
straight-shoota: This doesn't need a time zone, better use UTC directly:
```suggestion
      @creation_time = Time.utc
```

src/http/cookie.cr
straight-shoota: Yes, please use `Time.utc` everywhere.

src/http/cookie.cr
```diff
@@ -53,9 +58,20 @@ module HTTP
...
-        e < Time.utc
+    def expiration_time(time_reference = @creation_time)
+      if max_age = @max_age
+        time_reference + max_age
+      else
+        @expires
+      end
+    end
+
+    # Returns the expiration status of this cookie as a `Bool` given a creation time
```
straight-shoota:
```suggestion
    # Returns the expiration status of this cookie as a `Bool`.
```

Please explain the argument including its default value in another paragraph.

src/http/cookie.cr
```diff
@@ -53,9 +58,20 @@ module HTTP
...
+    def expired?(time_reference = @creation_time)
+      if @max_age == 0.seconds
```
straight-shoota:
```suggestion
      if @max_age.zero?
```

src/http/cookie.cr
```diff
@@ -25,23 +26,27 @@ module HTTP
     def_equals_and_hash name, value, path, expires, domain, secure, http_only
 
     def initialize(@name : String, value : String, @path : String = "/",
-                   @expires : Time? = nil, @domain : String? = nil,
-                   @secure : Bool = false, @http_only : Bool = false,
-                   @samesite : SameSite? = nil, @extension : String? = nil)
+                   @expires : Time? = nil, @max_age : Time::Span? = nil,
```
straight-shoota: There should be a validation that `@max_age` must not have a negative value.

straight-shoota (dismissed):

yujiri8:
Does this PR need someone to adopt it? I could really use the change.

watzon:
@yui-knk I believe so. I just don't have time to work on the requested changes right now.

src/http/cookie.cr
```diff
@@ -24,27 +24,49 @@ module HTTP
...
+      new(
+        _name, _value,
+        path: _path, expires: _expires, domain: _domain, secure: _secure,
+        http_only: _http_only, samesite: _samesite, extension: _extension
+      )
+    end
+
+    def self.new(_name : String, _value : String) : self
+      new(name: _name, value: _value)
+    end
```
Sija, line 46: Why's this initializer needed?

straight-shoota: Otherwise `Cookie.new(name, value)` would be a deprecation warning because the deprecated `.new` method shadows the implicit `.new` from `initialize`. But this basic constructor should continue to work, only the optional config arguments should be named-only.

Sija: So why not use `def self.new(name : String, value : String) : self` signature?

straight-shoota: The named arguments must not match the ones it delegates to or it would create an infinite recursion.

Sija: Ah, of course, thanks for bearing with me :)

beta-ziliani:
Closing this one now that we have #10564 to focus on. Thanks for the effort @watzon !

Issue #7857, opened 2019-06-05
File.new API deficiencies and proposals


didactic-drunk:
### There are several problems with the existing File.new API.

1. No way to create a file for read/write without truncating the existing file (O_CREAT). #7849
2. No way to open a file for create without overwriting or opening an existing one (O_EXCL).
3. No way to atomically lock a file when opening (O_EXLOCK).
4. No way to use to avoid following symlinks (O_NOFOLLOW).
5. Following the POSIX fopen API may seem easy but it doesn't scale to current or future uses.
6. File modes as strings are runtime errors, not compile time.

### Solutions
7. Solutions used by other languages.
8. Proposed solutions.

### 1. Creating without truncating 

It's not possible without race conditions.
See #7849

## 2. Creating a file without overwriting (O_EXCL)

Race condition avoidance is one of the numerous reasons it's **necessary**.  This flag may be combined with O_EXLOCK or O_NOFOLLOW for various uses.  Other examples are plentiful using google.

### 3-4. Atomic locking and O_NOFOLLOW

Opening a file with a lock **atomically** lets you easily modify files without race conditions or risk of data loss.  The [atomic_write shard](https://github.com/chris-huxtable/atomic_write.cr) would benefit from this.

Example of race condition free atomically rewriting a file:
```crystal
File.open "file", O_EXLOCK | ... do |old_file|
  buf = modify old_file.read
  File.open "file.tmp", O_EXCL | O_EXLOCK | ... do |new_file|
    new_file.write buf
  end
  File.rename old_file, new_file
end
```

Existing application already do this.  This feature is required for compatibility with other applications as File#flock_* has unavoidable race conditions.  Existing flock functions are still useful for upgradable locks or multi reader/writers on an existing file.

It seems like all major targeted platforms already support compatible open modes.

Linux/MacOS/FreeBSD/OpenBSD all support the following modes:
- O_SHLOCK Atomically obtain a shared lock.
- O_EXLOCK Atomically obtain an exclusive lock.
- O_NOFOLLOW If last path element is a symlink, don't follow it.

Windows
- CreateFile dwShareMode has similar options for locking.
- O_NOFOLLOW can be simulated using FILE_FLAG_OPEN_REPARSE_POINT, checking if the opened file is a symlink and raising an error if it is.  
- I don't have or use windows so this would need verification.

Creating an OS abstraction layer isn't needed as the necessary functions are widely supported.

## 5. Runtime vs compile time errors:

[...]

PR #7931, opened 2019-06-27
Allow passing custom modes to File.new/open

carlhoerberg:


carlhoerberg:
Allow you to open files with custom flags eg. `O_SYNC | O_DIRECT`. 

If it matters, Ruby supports this way of specifying flags: https://ruby-doc.org/core-2.6.3/IO.html#method-c-new

wontruefree:
@carlhoerberg have you checked out this [issue](https://github.com/crystal-lang/crystal/issues/7857)

carlhoerberg:
@wontruefree had not, thanks! i guess this is an implementation of the posix flag option

wontruefree:
I dont know what direction people are leaning now.  But, I wanted to let you know in case people did not respond to the PR.

asterite:
@carlhoerberg I can't see how this fixes anything, the `open_flag(mode)` call is behind a `mode : String` restriction, so there's no way you can pass an int to the `File` constructor.

carlhoerberg:
Where's that restriction? And if there is one, one could just remove that
it?

asterite:
For example here: https://github.com/crystal-lang/crystal/blob/3df0fd5c84e3019bd8709edcf9fb215b7fcb93d0/src/file.cr#L103

I think this needs a bit more design, though (which is happening in #7857)

carlhoerberg:
Works locally for me without modifying File.new. Does setting a default
value restrict  type?

asterite:
Oh, I got confused because I saw the String type restriction. So I guess that's fine. But I'm still unsure about merging this.

ysbaddaden:
This is a nice monkey-patch workaround to avoid the issue, but I don't think we should merge it in std. We should design and implement a proper solution in #7857.

didactic-drunk:
It looks like everyone was leaning for POSIX flags or renamed POSIX flags.   Would a PR for one of those be approved?

PR #7935, opened 2019-06-28
Add debug log helper function for codegen

bcardiff:
While working with some changes in the codegen is convenient to printf some messages. This PR improves the state of the art to include such messages.

The `debug_codegen_log` function has some documentation for the allowed variations.

Including some `debug_codegen_log` calls in `codegen_dispatch` for example would print

```
<block: next_def @ src/compiler/crystal/codegen/call.cr:385> 
<block: current_def1 @ src/compiler/crystal/codegen/call.cr:368> 
```

Allowing easy tracing of the execution of the LLVM IR code and mapping which part of the codegen code generates the llvm code.

Those logs messages will not generate code unless `CRYSTAL_DEBUG_CODEGEN` env var is set.

To achieve this some additional things were covered:

* The codegen `printf` needed a fix (missing argument)
* Add some code name extraction with through LLVMExt
* Refactor the codegen `unreachable` to use `debug_codegen_log` instead of a custom `printf` call.

### DIFF
--- a/src/llvm/basic_block.cr
+++ b/src/llvm/basic_block.cr
@@ -17,4 +17,9 @@ struct LLVM::BasicBlock
   def to_unsafe
     @unwrap
   end
+
+  def name
+    block_name = LibLLVMExt.basic_block_name(self)
+    block_name ? LLVM.string_and_dispose(block_name) : nil
+  end
 end
--- a/src/compiler/crystal/codegen/codegen.cr
+++ b/src/compiler/crystal/codegen/codegen.cr
@@ -1813,6 +1813,42 @@ module Crystal
       call @program.printf(@llvm_mod, llvm_context), [builder.global_string_pointer(format)] + args
     end
 
+    # Emits a debug message that shows the current llvm basic block name,
+    # the location within the codegen that was used to emit this log.
+    #
+    # The message is only generated if `CRYSTAL_DEBUG_CODEGEN` is set
+    #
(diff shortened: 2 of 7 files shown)


src/llvm/basic_block.cr
```diff
@@ -17,4 +17,9 @@ struct LLVM::BasicBlock
   def to_unsafe
     @unwrap
   end
+
+  def name
+    block_name = LibLLVMExt.basic_block_name(self)
+    block_name ? LLVM.string_and_dispose(block_name) : "unknown"
```
asterite: Since a name is optional, I would expect this to return `String?`.

asterite: (the "unknown" can be inserted on the caller site, like `block.name || "unknown"`)

bcardiff: I'm not sure names are optional. In llvm < 4.0 we might not be able to know the name, but basic blocks always have a name I think. So I was catching that and ensuring to return a `String`.

If you think is still better to return a `String?` I can apply the suggested changes.

asterite: Yes, basic blocks can have no name, for example the entry block is usually left without a name. In the API for creating blocks the default name is `""` if not given. So I think we should either return `nil` or `""`, but not "undefined" (because there could be a block named like that).

bcardiff: Ok. done!

PR #7938, opened 2019-06-29
YAML: deserialize String last in a union type

### LINKED ISSUE #7936: Incorrect Parsing of YAML Hash with Union Value
When trying to do a YAML.mapping in a class with the unique combination of String, Nil, and another YAML.mapping class `value: null` gets captured as a string "null" and not as nil as it should. null without quotes is a reserved YAML word and should always be treated as such.

Crystal Version 0.29.0

```crystal
require "yaml"

class Foo
  YAML.mapping(
    bar: Hash(String, String | Nil | Raz)
  )
end

class Raz
  YAML.mapping(
    taz: String?
  )
end

yaml = <<-YAML
bar: 
  baz: null
YAML

foo = Foo.from_yaml(yaml)

puts foo
# => {"baz" => "null"}
```

Also, when the value is empty in yaml, it defaults to the first empty class.

```crystal
require "yaml"

class Foo
  YAML.mapping(
    bar: Hash(String, String | Nil | Raz)
  )
end

class Raz
  YAML.mapping(
    taz: String?
  )
end

yaml = <<-YAML
bar: 
  baz:
YAML

foo = Foo.from_yaml(yaml)

puts foo
# => {"baz" => #<Raz:0x103f7e880 @taz=nil>}
```

asterite:
Fixes #7936

I can hear @straight-shoota's voice saying "Told ya!" but this fix is good because:
- When mapping to a String it will always work
- When mapping a union which has a String, String is tried last, so numbers, nulls and anything else will have precedence

### DIFF
--- a/spec/std/yaml/serialization_spec.cr
+++ b/spec/std/yaml/serialization_spec.cr
@@ -175,8 +175,8 @@ describe "YAML serialization" do
       value.should eq(Time.utc(2014, 1, 2))
     end
 
-    it "deserializes union" do
-      Array(Int32 | String).from_yaml(%([1, "hello"])).should eq([1, "hello"])
+    it "deserializes union with nil, string and int (#7936)" do
+      Array(Int32 | String | Nil).from_yaml(%([1, "hello", null])).should eq([1, "hello", nil])
     end
 
     it "deserializes time" do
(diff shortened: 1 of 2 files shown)


bcardiff:
It seems there are no specs for empty values. From the null specs i guess they are not interpreted as nil, which is good. But can a spec be added to ensure that?

Otherwise LGTM

asterite:
@bcardiff I just added specs for that:

```crystal
String.from_yaml("") # => ""
Nil.from_yaml("") # => nil
```

So in the case of a union of `String | Nil`, empty means null. I'm not sure how correct is that, but it's not clear from the YAML spec.

I don't think this is too problematic. If someone needs a special logic they can always use a custom converter.

PR #7940, opened 2019-06-29
Refactor codegen of unions

bcardiff:
This PR performs refactors and some code cleanups to move all the code related to the binary representation of the unions to a new file `codegen/unions.cr`.

The code cleanup is mainly related to how caching of the `llvm_value`/`union_value_cache` was done.

This will serve as a basis for changing the representation of the unions later.

I thought it would be easier to understand if the refactor itself was put in an independent PR.

### DIFF
--- a/src/compiler/crystal/codegen/unions.cr
+++ b/src/compiler/crystal/codegen/unions.cr
@@ -1,0 +1,85 @@
+module Crystal
+  class LLVMTyper
+    private def create_llvm_type(type : MixedUnionType, wants_size)
+      llvm_name = llvm_name(type, wants_size)
+      if s = @structs[llvm_name]?
+        return s
+      end
+
+      @llvm_context.struct(llvm_name) do |a_struct|
+        if wants_size
+          @wants_size_cache[type] = a_struct
+        else
+          @cache[type] = a_struct
+          @structs[llvm_name] = a_struct
+        end
+
+        max_size = 0
+        type.expand_union_types.each do |subtype|
+          unless subtype.void?
+            size = size_of(llvm_type(subtype, wants_size: true))
+            max_size = size if size > max_size
+          end
+        end
+
+        max_size /= pointer_size.to_f
+        max_size = max_size.ceil.to_i
+
+        max_size = 1 if max_size == 0
+
+        llvm_value_type = size_t.array(max_size)
+
+        if wants_size
+          @wants_size_union_value_cache[type] = llvm_value_type
+        else
+          @union_value_cache[type] = llvm_value_type
+        end
+
+        [@llvm_context.int32, llvm_value_type]
+      end
+    end
+  end
+
+  class CodeGenVisitor
+    def union_type_id(union_pointer)
+      aggregate_index union_pointer, 0
+    end
+
+    def union_value(union_pointer)
+      aggregate_index union_pointer, 1
+    end
+
+    def store_in_union(union_pointer, value_type, value)
+      store type_id(value, value_type), union_type_id(union_pointer)
+      casted_value_ptr = cast_to_pointer(union_value(union_pointer), value_type)
+      store value, casted_value_ptr
+    end
+
+    def store_bool_in_union(union_type, union_pointer, value)
+      store type_id(value, @program.bool), union_type_id(union_pointer)
+
+      # To store a boolean in a union
+      # we sign-extend it to the size in bits of the union
+      union_value_type = llvm_union_value_type(union_type)
+      union_size = @llvm_typer.size_of(union_value_type)
+      int_type = llvm_context.int((union_size * 8).to_i32)
+
+      bool_as_extended_int = builder.zext(value, int_type)
+      casted_value_ptr = bit_cast(union_value(union_pointer), int_type.pointer)
+      store bool_as_extended_int, casted_value_ptr
+    end
+
+    def store_nil_in_union(union_pointer, target_type)
+      union_value_type = llvm_union_value_type(target_type)
+      value = union_value_type.null
+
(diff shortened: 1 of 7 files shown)


src/compiler/crystal/codegen/unions.cr
```diff
@@ -0,0 +1,122 @@
...
+# Together these operations should encapsulate the binary representation of the MixedUnionType.
+#
+# Other unions like ReferenceUnionType that have a more trivial
+# representation are not handled here.
+#
+module Crystal
+  class LLVMTyper
+    private def create_llvm_type(type : MixedUnionType, wants_size)
+      llvm_name = llvm_name(type, wants_size)
+      if s = @structs[llvm_name]?
```
Sija: Please, use full names.

bcardiff: The scope of `s` is only the `if`, which consist of a 1 line. Otherwise, I would agree with using the longer variable name.

Sija: Well, yeah I know, but still... it won't hurt to avoid cryptic one-letter variable names...

asterite: I think the issue here is that a good name is probably `struct`, but that's a keyword... so one could use `a_struct` or similar. But given that it's just one line, I think `s` is fine.

src/compiler/crystal/codegen/unions.cr
```diff
@@ -78,6 +78,10 @@ module Crystal
       store value, casted_value_ptr
     end
 
+    def store_void_in_union(union_pointer, target_type)
```
asterite: 😻

src/compiler/crystal/codegen/unions.cr
```diff
@@ -41,6 +41,10 @@ module Crystal
   end
 
   class CodeGenVisitor
+    def union_type_and_value_pointer(union_pointer)
+      {load(union_type_id(union_pointer)), union_value(union_pointer)}
```
asterite: What do you think if we define `union_type_id` and `union_value` as local variables here:

```crystal
def union_type_and_value_pointer(union_pointer)
  union_type_id = aggregate_index union_pointer, 0
  union_value = aggregate_index union_pointer, 1
  {load(union_type_id, union_value)}
end
```

Then we can redefine `def union_value` and `def union_type_id`:

```crystal
def union_type_id(ptr)
  union_type_and_value_pointer(ptr)[0]
end

def union_value(ptr)
  union_type_and_value_pointer(ptr)[1]
end
```

Then you can still use `union_value(...)` and `union_type_id(...)` instead of using `_, value = union_type_and_value_pointer(...)` in many places.

And the knowledge of how a union is stored is still kept in a single place: `union_type_and_value_pointer`.

Another nice thing about this is that we could eventually "optimize" the two methods to avoid generating LLVM code that's not used (I don't think the extra code will appear in the final code, but it's still invoking some LLVM apis).

bcardiff: I guess there is `)` missplaced:

```cr
def union_type_and_value_pointer(union_pointer)
  union_type_id = aggregate_index union_pointer, 0
  union_value = aggregate_index union_pointer, 1
  {load(union_type_id), union_value}
end
```

The llvm values are not emitted unless it's used I think. Or they should be removed since they are not used.

I understand that having those 2 methods seems easier to read, but will allow misguided use of unions in a MT environment. The value could change between the two operations. Alternative when reading a value the supposed type must be passed and that API is weirder.

I would even remove (or make private) the `union_type_id` and `union_value` methods.

asterite: > but will allow misguided use of unions in a MT environment

How is this possible if the methods only read?

I mean, how is this:

```crystal
_, value = union_type_and_value_pointer(ptr, type)
```

safer than this:

```crystal
value = union_value(ptr, type)
```

given that `union_value(ptr, type)` is just `union_type_and_value_pointer(ptr, type)[1]`?

asterite: I'm only proposing `union_value(ptr, type)` to be a shorter way to write `_, value = union_type_and_value_pointer(ptr, type)`.

bcardiff: It won't be wrong by itself, but later one will use `union_type_id(ptr)` and `union_value(ptr)` independently in the same codegen method and miss the point that type and value should be read in the same operation.

asterite: Ooooh... I see what you mean.

Then yes, that makes sense.

Should we remove the methods `union_value` and `union_type_id` then, and inline them in the other one?

bcardiff: On second thoughts, the union_type_id and union_value have some occurrences here. We could either leave them as private or even more accurate, file's private. Since the codegen is a heavily used partial class, making it instance private method will not prevent its usage.

Is there any objection to converting them into private file methods?

bcardiff: I see `union_type_and_value_pointer` as the union read operation.
`store*_in_union` as the writer operations.

And `union_type_id` and `union_value` as helpers for the particular representation of the unions. That should be used in the above methods. I don't want to replace `union_type_id` for a call to `union_type_and_value_pointer`.

asterite: Sounds good. File private or not, it's fine to leave the helpers 👍

bcardiff: Ok, let's merge it as is. Making it file private requires passing the codegen as argument, the refactor can be done later. There are other implementations helper methods in the codegen.

PR #7944, opened 2019-07-01
Json keys

### LINKED ISSUE #7942: JSON::Serializable for deserialization of hash with integer keys doesn't work
JSON::Serializable for deserialization of hash with integer keys doesn't work:

```crystal
require "json"

class Building
	include JSON::Serializable
	property locations = {} of Int32 => String
#	property locations = {} of String => String

	def initialize(); end
end

b = Building.new
b.locations[2] = "Alfa"
#b.locations["2"] = "Alfa"

ser = b.to_json
p ser

y = Building.from_json(ser)
p y
```

Is it a bug or a feature?

It works for YAML serialization. And it works for string keys in JSON serialization.
But it produces awful huge error messages for integer keys - error messages that are absolutely unhelpful. 

If it is a feature then it is a rather trap for newcomer to Crystal programming. It seems to be very unnatural for me. I was rewriting a bigger piece of code to Crystal and it was difficult to spot the cause of the errors. I expected that serialization is always bidirectional if I use `include JSON::Serializable`.

Maybe it is OK. I just wanted to let you know.

I am on Xubuntu 18.04

Crystal 0.29.0 [fbfe8b6] (2019-06-05)
LLVM: 4.0.0

cat /etc/issue
Ubuntu 18.04.2 LTS

asterite:
Fixes #7942

Also document a bit the type-safe way to parse JSON (I don't think it was documented before, and mapping/serializable doesn't count because it doesn't talk about primitive types, Array, Hash, etc.)

I decided to only provide out-of-the-box serialization of json object keys for ints and floats, because their text representation is pretty obvious. Other types are more controversial but could be provided by shards.

### DIFF
--- a/src/json.cr
+++ b/src/json.cr
@@ -1,5 +1,51 @@
 # The JSON module allows parsing and generating [JSON](http://json.org/) documents.
 #
+# ### General type-safe interface
+#
+# The general type-safe interface for parsing JSON is to invoke `T.from_json` on a
+# target type `T` and pass either a `String` or `IO` as an argument.
+#
+# ```
+# require "json"
+#
+# json_text = %([1, 2, 3])
+# Array(Int32).from_json(json_text) # => [1, 2, 3]
+#
+# json_text = %({"x": 1, "y": 2})
+# Hash(String, Int32).from_json(json_text) # => {"x" => 1, "y" => 2}
+# ```
+#
+# Serializing is achieved by invoking `to_json`, which returns a `String`, or
+# `to_json(io : IO)`, which will stream the JSON to an `IO`.
+#
+# ```
+# require "json"
+#
+# [1, 2, 3].to_json            # => "[1, 2, 3]"
+# {"x" => 1, "y" => 2}.to_json # => "{\"x\": 1, \"y\": 2}
+# ```
+#
+# Most types in the standard library implement these methods. For user-defined types
+# you can define a `self.new(pull : JSON::PullParser)` for parsing and
+# `to_json(builder : JSON::Builder)` for serializing. The following sections
+# show convenient ways to do this using either `JSON.mapping` or `JSON::Serializable`.
+#
+# NOTE: JSON object keys are always strings but they can still be parsed
+# and deserialized to other types. To deserialize, define a
+# `T.from_json_object_key?(key : String) : T?` method, which can return `nil`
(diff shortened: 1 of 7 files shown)


src/json.cr
```diff
@@ -1,5 +1,51 @@
...
+#
+# json_text = %([1, 2, 3])
+# Array(Int32).from_json(json_text) # => [1, 2, 3]
+#
+# json_text = %({"x": 1, "y": 2})
+# Hash(String, Int32).from_json(json_text) # => {"x" => 1, "y" => 2}
+# ```
+#
+# Serializing is achieved by invoking `to_json`, which returns a `String`, or
+# `to_json(io : IO)`, which will streat the JSON to an `IO`.
```
jessedoyle: I think there's a typo here (`streat` -> `stream`?).

asterite: Thanks, fixed

src/json.cr
```diff
@@ -1,5 +1,51 @@
...
+# ```
+#
+# Most types in the standard library implement these methods. For user-defined types
+# you can define a `self.new(pull : JSON::PullParser)` for parsing and
+# `to_json(builder : JSON::Builder)` for serializing. The following sections
+# show convenient ways to do this using either `JSON.mapping` or `JSON::Serializable`.
+#
+# NOTE: JSON object keys are always strings but they can still be parsed
+# and deserialized to other types. To deserialize, define a
+# `T.from_json_object_key(key : String) : T?` method, which can return `nil`
```
RX14: `T.from_json_object_key?(`

asterite: Thanks, done!

src/json.cr
```diff
@@ -1,5 +1,51 @@
...
+# ```
+#
+# Serializing is achieved by invoking `to_json`, which returns a `String`, or
+# `to_json(io : IO)`, which will stream the JSON to an `IO`.
+#
+# ```
+# require "json"
+#
+# [1, 2, 3].to_json    # => "[1, 2, 3]"
+# {"x" => 1, "y" => 2} # => "{\"x\": 1, \"y\": 2}
```
luislavena: Missing the `.to_json` call in the example 😁 

```suggestion
# {"x" => 1, "y" => 2}.to_json # => "{\"x\": 1, \"y\": 2}
```

asterite: Thank you! ❤️

PR #7952, opened 2019-07-04
Allow rendering tags on methods without any docs

### LINKED ISSUE #7895: Deprecated annotation not adding for methods that don't have any docs
I remember the work around #7661 should add the `DEPRECATED` tag to the API docs when the annotation is added to a method.  However that does not seem to be working currently.

https://crystal-lang.org/api/master/String.html#at%28index%3AInt%29-instance-method

https://github.com/crystal-lang/crystal/blob/099b9ecc28f6ee015718c140ef2d62566876d675/src/string.cr#L846-L847

Blacksmoke16:
Fixes #7895.

* Only skip generating docs if the method does not have any docs and doesn't have `Deprecated` annotation
* Adds specs for generating summary/docs for a method
* Fixes some type links on Int `Deprecated` messages, so that they are linkable in the API docs

### DIFF
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -274,14 +274,14 @@ class Crystal::Doc::Generator
 
   def summary(obj : Type | Method | Macro | Constant)
     doc = obj.doc
-    return nil unless doc
 
-    summary obj, doc
+    return if !doc && !obj.annotations(@program.deprecated_annotation)
+
+    summary obj, doc || ""
   end
 
   def summary(context, string)
-    line = fetch_doc_lines(string).lines.first?
-    return nil unless line
+    line = fetch_doc_lines(string).lines.first? || ""
 
     dot_index = line =~ /\.($|\s)/
     if dot_index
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -274,14 +274,14 @@ class Crystal::Doc::Generator
...
-    summary obj, doc
+    return if !doc && !obj.annotations(@program.deprecated_annotation)
+
+    summary obj, doc || ""
   end
 
   def summary(context, string)
-    line = fetch_doc_lines(string).lines.first?
-    return nil unless line
+    line = fetch_doc_lines(string).lines.first? || ""
```
bcardiff: from :280 the `string` is never `nil`.

Blacksmoke16:
@bcardiff Not sure I follow.  Isn't that what it's doing already?  If there are no docs nor annotations on the method, it'll return early.  Otherwise if there is one or the other, there are docs to generate.

bcardiff:
I meant that the new `return if !doc && !obj.annotations(@program.deprecated_annotation)` seems odd. If the `summary` and `doc` methods returned `nil` in those cases, then the check is not needed.

It's just an idea for (maybe) cleaner code.

But as I said. Is good for approving. If the refactor make sense to you feel free to iterate.

Thanks!

PR #7958, opened 2019-07-11
HTTP::Client: check host is just a host

### LINKED ISSUE #7902: Socket::Addrinfo::Error when using HTTP::Client constructor
```
Crystal 0.29.0 (2019-06-06)

LLVM: 6.0.1
```

```crystal
require "http/client"

client = HTTP::Client.new("api.airbnb.com/v2", tls: true)
client.compress = false
r = client.post("/logins", body: "{}")
pp r
```

```log
Unhandled exception: No address found for api.airbnb.com/v2:443 over TCP (Socket::Addrinfo::Error)
  from /usr/lib/crystal/socket/tcp_socket.cr:124:9 in 'initialize'
  from /usr/lib/crystal/socket/tcp_socket.cr:27:3 in 'new'
  from /usr/lib/crystal/http/client.cr:736:5 in 'socket'
  from /usr/lib/crystal/http/client.cr:624:5 in 'send_request'
  ...
```

This happens so far only with `api.airbnb.com`. Executing `HTTP::Client.post(...)` works without `Socket::Addrinfo::Error` but it doesn't decompress the gzip response, which is why I tried to do it manually

asterite:
Fixes #7902

To be honest, I'd really like to change `HTTP::Client` to work like in other languages (Go, C#), like some of what is proposed in #6011

But until that happens I think it's good if we make it easier to use the existing HTTP::Client.

### DIFF
--- a/spec/std/http/client/client_spec.cr
+++ b/spec/std/http/client/client_spec.cr
@@ -60,6 +60,21 @@ module HTTP
     typeof(Client.post("http://www.example.com", body: Bytes[65]))
     typeof(Client.new("host").post("/", body: Bytes[65]))
 
+    describe "from String" do
+      it "raises when not a host" do
+        ["http://www.example.com",
+         "www.example.com:8080",
+         "example.com/path",
+         "example.com?query",
+         "http://example.com:bad_port",
+         "user:pass@domain"].each do |string|
+          expect_raises(ArgumentError, "The string passed to create an HTTP::Client must be just a host, not #{string.inspect}") do
+            Client.new(string)
+          end
+        end
+      end
+    end
+
     describe "from URI" do
       it "has sane defaults" do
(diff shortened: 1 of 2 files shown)


bcardiff:
I guess the `def self.new(uri : URI, tls = nil)` should also have a runtime check that the URI does not contains a path. 

The protected method `validate_host` seems to be used in other places to ensure there is a host, but making it validate that the URI has no path will break some stuff probably.

asterite:
@bcardiff Yeah, I saw that too but I think that's a different issue to fix (I just wanted to fix one thing, the other one isn't reported so it might not be a problem, but we'd still want to do it). Plus maybe when you create an `HTTP::Client` with a path it makes sense to use that as a base path, even merging the query string. I don't know...

PR #7965, opened 2019-07-15
validate xml element names

### LINKED ISSUE #7963: XML::Builder allows creating elements with invalid names
Noticed in https://github.com/Blacksmoke16/oq/issues/9.

```cr
require "xml"

xml = XML.build(indent: 2) do |xml|
  xml.element("1") do
    xml.text "foo"
  end
end

puts xml
<?xml version="1.0"?>
<1>foo</1>
```

```cr
require "xml"

xml = XML.build(indent: 2) do |xml|
  xml.element("a b=\"c\"") do
    xml.text "foo"
  end
end

puts xml
<?xml version="1.0"?>
<a b="c">foo</a b="c">
```

I would expect this to error since an element name cannot start with a number nor contain a space, equal sign etc as defined [here](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-Name).

Blacksmoke16:
Uses `LibXML.xmlValidateNameValue` to validate the provided name when using `XML::Builder` to build elements.

Fixes #7963 
Fixes https://github.com/Blacksmoke16/oq/issues/9

### DIFF
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -44,11 +44,14 @@ struct XML::Builder
 
   # Emits the start of an element.
   def start_element(name : String) : Nil
-    call StartElement, string_to_unsafe(name)
+    element_name = string_to_unsafe name
+    raise XML::Error.new("Invalid element name: '#{name}'", __LINE__) if LibXML.xmlValidateNameValue(element_name).zero?
+    call StartElement, element_name
   end
 
   # Emits the start of an element with namespace info.
   def start_element(prefix : String?, name : String, namespace_uri : String?) : Nil
+    raise XML::Error.new("Invalid element name: '#{prefix}#{name}'", __LINE__) if LibXML.xmlValidateNameValue(string_to_unsafe("#{prefix}#{name}")).zero?
     call StartElementNS, string_to_unsafe(prefix), string_to_unsafe(name), string_to_unsafe(namespace_uri)
   end
 
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -44,8 +44,9 @@ struct XML::Builder
 
   # Emits the start of an element.
   def start_element(name : String) : Nil
-    check_valid_element_name name, string_to_unsafe(name), "element name"
-    call StartElement, name
+    unsafe_name = string_to_unsafe(name)
+    check_valid_element_name name, unsafe_name, "element name"
+    call StartElement, unsafe_name
   end
 
   # Emits the start of an element with namespace info.
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -44,15 +44,19 @@ struct XML::Builder
 
   # Emits the start of an element.
   def start_element(name : String) : Nil
-    check_valid_element_name name, "element name"
+    check_valid_element_name name, string_to_unsafe(name), "element name"
     call StartElement, name
   end
 
   # Emits the start of an element with namespace info.
   def start_element(prefix : String?, name : String, namespace_uri : String?) : Nil
-    check_valid_element_name name, "element name"
-    check_valid_element_name prefix, "prefix" if prefix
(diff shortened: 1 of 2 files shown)


src/xml/builder.cr
```diff
@@ -44,11 +44,14 @@ struct XML::Builder
...
   def start_element(name : String) : Nil
-    call StartElement, string_to_unsafe(name)
+    element_name = string_to_unsafe name
+    raise XML::Error.new("Invalid element name: '#{name}'", __LINE__) if LibXML.xmlValidateNameValue(element_name).zero?
+    call StartElement, element_name
   end
 
   # Emits the start of an element with namespace info.
   def start_element(prefix : String?, name : String, namespace_uri : String?) : Nil
+    raise XML::Error.new("Invalid element name: '#{prefix}#{name}'", __LINE__) if LibXML.xmlValidateNameValue(string_to_unsafe("#{prefix}#{name}")).zero?
```
asterite: Shouldn't `prefix` and `name` be validated separately? Or at least, would it be the same to do that? Then we don't need a string interpolation, plus we could say "Invalid prefix" or "Invalid element name".

Blacksmoke16: Ah, good call.  I'll get that fixed.

src/xml/builder.cr
```diff
@@ -44,14 +44,14 @@ struct XML::Builder
...
-    raise XML::Error.new("Invalid element name: '#{name}'", __LINE__) if LibXML.xmlValidateNameValue(element_name).zero?
-    call StartElement, element_name
+    raise XML::Error.new("Invalid element name: '#{name}'", __LINE__) unless validate_element_name name
+    call StartElement, name
   end
 
   # Emits the start of an element with namespace info.
   def start_element(prefix : String?, name : String, namespace_uri : String?) : Nil
-    raise XML::Error.new("Invalid element name: '#{prefix}#{name}'", __LINE__) if LibXML.xmlValidateNameValue(string_to_unsafe("#{prefix}#{name}")).zero?
+    raise XML::Error.new("Invalid element name: '#{name}'", __LINE__) unless validate_element_name name
```
asterite: Sorry I forgot to mention this but the `line` argument passed to `XML::Error` is the line number in the XML document. I think that's not available when building so I would pass zero here (just like we pass zero in other places). Probably something to improve later on.

Another idea: have a `check_valid_element_name(name, element_kind)` which does the checking and the raising, so for example like `check_valid_element_name(name, "element name")` and `check_valid_element_name(prefix, "prefix")`.

src/xml/builder.cr
asterite: Thank you! But now `string_to_unsafe` is called twice for each `name` and `prefix`, and that does a null byte checking so now that is done twice. So an optional thing to do is to call that first and then pass those things to `check_valid...`, though I can see it starts to become a bit annoying. Sorry! So feel free to ignore this comment, eventually we can optimize that.

src/xml/builder.cr
```diff
@@ -44,12 +44,19 @@ struct XML::Builder
 
   # Emits the start of an element.
   def start_element(name : String) : Nil
-    call StartElement, string_to_unsafe(name)
+    check_valid_element_name name, string_to_unsafe(name), "element name"
+    call StartElement, name
```
bcardiff: Isn't this changing the argument?

Shouldn't it be:

```cr
    unsafe_name = string_to_unsafe(name)
    check_valid_element_name name, unsafe_name, "element name"
    call StartElement, unsafe_name
```

Blacksmoke16: Oh, good catch.  Wonder why that didnt cause things to brek..

asterite: It's the same

asterite: Actually it's better the new way to avoid double call

PR #7966, opened 2019-07-16
JSON: use enums instead of symbols

asterite:
This changes `JSON::Lexer` and `JSON::PullParser` to expose enums instead of symbols for token kinds and pull parser state kinds. Symbols work but they are not type-safe. Plus enums gives enumerate (duh) all possible states.

The PullParser has already proven to be very useful but moving forward I wouldn't like to keep the Symbols interface so it's better now or never.

This is a breaking change:
- `JSON::Lexer::Token#type : Symbol` is replaced by `JSON::Lexer::Token#kind : JSON::Lexer::Token::Kind`
- `JSON::PullParser#kind : Symbol` is replaced by `JSON::PullParser#kind : JSON::PullParser::Kind`

To make it easy to upgrade I kept `JSON::Lexer::Token#type : Symbol` as deprecated and also made `JSON::PullParser::Kind` be comparable to symbols, with a deprecation notice. That way we don't actually break anything but if you run the compiler with `--warnings=all` then you can fix things at your time.

Once merged, in a subsequent PR I'll try to document all of these types, including the ones in the YAML module.

I also added `JSON::Builder#next_is_object_key?` because I think it's useful and I at least needed it for optimizing one thing in https://github.com/Blacksmoke16/oq (though I didn't send a PR there yet).

### DIFF
--- a/src/json/builder.cr
+++ b/src/json/builder.cr
@@ -280,6 +280,13 @@ class JSON::Builder
     end
   end
 
+  # Returns `true` if the next thing that must pushed into this
+  # builder is an object key (so a string) or the end of an object.
+  def next_is_object_key? : Bool
+    state = @state.last
+    state.is_a?(ObjectState) && state.name
+  end
+
   private def scalar(string = false)
     start_scalar(string)
     yield.tap { end_scalar(string) }
--- a/src/json/token.cr
+++ b/src/json/token.cr
@@ -1,5 +1,21 @@
 class JSON::Token
-  property type : Symbol
+  enum Kind
+    Null
+    False
+    True
+    Int
+    Float
+    String
+    BeginArray
+    EndArray
+    BeginObject
+    EndObject
+    Colon
+    Comma
+    EOF
+  end
+
+  property kind : Kind
   property string_value : String
   property int_value : Int64
   property float_value : Float64
@@ -8,7 +24,7 @@ class JSON::Token
   property raw_value : String
 
   def initialize
-    @type = :EOF
+    @kind = :EOF
     @line_number = 0
     @column_number = 0
     @string_value = ""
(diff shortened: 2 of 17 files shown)


src/json/builder.cr
```diff
@@ -280,6 +280,13 @@ class JSON::Builder
     end
   end
 
+  # Returns `true` if the next thing that must pushed into this
```
Sija: Actually it returns a string (`state.name`) and `false` otherwise.

asterite: Thanks! But no, it returns a `Bool` (I added a return type just in case).

src/json/token.cr
```diff
@@ -17,16 +33,70 @@ class JSON::Token
...
-      @float_value.to_s(io)
-    when :STRING
-      @string_value.to_s(io)
+    case @kind
+    when .null?
+      io << "null"
+    when .false?
+      io << "false"
+    when .true?
+      io << true
```
Sija:
```suggestion
      io << "true"
```

src/json/token.cr
```diff
@@ -17,16 +33,70 @@ class JSON::Token
...
+    when .int?
+      raw_value.to_s(io)
+    when .float?
+      raw_value.to_s(io)
+    when .string?
+      string_value.to_s(io)
+    when .begin_array?
+      io << "["
```
Sija: All these could be `Char`s.

asterite: Yes, but appending a Char is slower! The char needs to be decoded from its codepoint into UTF-8. With a string it's just a buffer copy.

asterite: Actually, it seems appending a char is faster for some reason... maybe for a hardcoded char LLVM optimizes it better.

src/json/token.cr
asterite: Thanks, done!

j8r:
Even the `Token` class could be avoided, see https://github.com/j8r/con/blob/master/src/lexer.cr .
I was planning to do it, but never get enough time to.

RX14:
I'm not so sure about the symbol compatibility using `==`, it seems very limited in the code which it will not cause a breaking change in. But I might be wrong.

asterite:
You mean it might cause a breaking change after all?

RX14:
Maybe? The only way to tell would be to find a corpus of projects which use these classes

asterite:
Even if it's not completely backwards compatibility I'd move forward with this. Upgrading it should be relatively straight forward.

PR #7969, opened 2019-07-17
Allow defining default value with getter!

### LINKED ISSUE #7968: Can't define a default value with getter!
https://play.crystal-lang.org/#/r/78lv

```cr
class Foo
  getter! name : String = "Jim"
end
```

```
There was a problem expanding macro 'getter!'

Called macro defined in macro 'macro_139694884568576'

 223 | macro getter!(*names)

Which expanded to:

 >  1 |       
 >  2 | 
 >  3 |   @name : String = "Jim"?
 >  4 |   
 >  5 | 
 >  6 | 
 >  7 | def name?
 >  8 |   @name
 >  9 | end
 > 10 | 
 > 11 | def name
 > 12 |   @name.not_nil!
 > 13 | end
 > 14 |       
 > 15 |     
Error: expecting token ':', not 'def'
```

Looks like it's adding the `?` incorrectly.

Blacksmoke16:
Fixes #7968

### DIFF
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -21,6 +21,7 @@ private class TestObject
   @getter5 : Int32?
 
   getter! getter6 : Int32
+  getter! getter_default : String = "Jim"
 
   getter? getter7
   getter? getter8 : Bool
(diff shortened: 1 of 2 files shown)


asterite:
As I mentioned in the original issue, I can't see what's the point of having a default value here.

PR #7970, opened 2019-07-17
Make some macro methods more uniform

asterite:
Related issue: https://forum.crystal-lang.org/t/add-nilable-to-more-macro-types/273

This PR does these things (you can check each commit too):
- Add `Union#resolve` and `Union#resolve?`: this was missing. For example [this code](https://play.crystal-lang.org/#/r/78lk) doesn't compile but it should (if you replace `decl.type.nilable?` with `decl.type.resolve.nilable?`).
- Add `TypeNode#resolve` and `TypeNode#resolve?`: so in case you either get a Path, Union, Generic or an already resolved TypeNode you can always call `resolve` on it and get a TypeNode
- Add `Path#types` and `Generic#types`: there's already `Union#types` which returns a union's types. Instead of casing on it we can define `types` in these two other types two so you can always call `types` on something that looks like a type and work with that in a uniform way
- Let `TypeNode#union_types` also work for non-union types: same reason as before to make it easier to work uniformly with unions and non-unions by providing a way to convert a type to a list of types it holds (for a non-union it's just that type).

These are small improvements that, I believe, will make the life of macro users a bit more simple.

### DIFF
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -1297,6 +1297,12 @@ module Crystal::Macros
     # returns a `NilLiteral`.
     def resolve? : ASTNode | NilLiteral
     end
+
+    # Returns this path inside an array literal.
+    # This method exists so you can call `types` on the type of a type
+    # declaration and get all types, whether it's a Generic, Path or Union.
+    def types : ArrayLiteral(ASTNode)
+    end
   end
 
   # A class definition.
@@ -1343,6 +1349,12 @@ module Crystal::Macros
     # Resolves this path to a `TypeNode` if it denotes a type,
     # or otherwise returns a `NilLiteral`.
     def resolve? : ASTNode | NilLiteral
+    end
+
+    # Returns this generic inside an array literal.
+    # This method exists so you can call `types` on the type of a type
+    # declaration and get all types, whether it's a Generic, Path or Union.
+    def types : ArrayLiteral(ASTNode)
     end
   end
 
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -1697,7 +1697,7 @@ module Crystal::Macros
     end
 
     # Returns the types forming a union type, if this is a union type.
-    # Gives a compile error otherwise.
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/macros.cr
```diff
@@ -1297,6 +1297,12 @@ module Crystal::Macros
     # returns a `NilLiteral`.
     def resolve? : ASTNode | NilLiteral
     end
+
+    # Returns this path inside an array litera or Union.
```
bcardiff:
```suggestion
    # Returns this path inside an array literal or Union.
```

asterite: Thanks!

bcardiff: If you squash it to the actual commit we can merge the PR without squashing it if you want.

asterite: I don't know how to do that (my git knowledge is a bit limited) but I think we shouldn't worry too much about the commit history. A typo fix is a typo fix and I think it's fine if that appears in a separate commit.

oprypin: Yes, please just merge this without squashing.
But if there are more followups, manual squashing may become more relevant.

Then you can do `git rebase -i`, move the line of "fix typo" right under its relevant commit and also change it to be "fixup". And `git push --force`...

asterite: Do I need to do git rebase -I master to fix this? I'm happy to learn.

bcardiff: `$ git rebase -i master`

once the editor open move the last line 1 up so 2a4e6b6b3 is squashed with 746cc33f8.

```
pick f85222a82 Macros: add Union#resolve and Union#resolve?
pick d0562bbd2 Macros: add TypeNode#resolve and TypeNode#resolve?
pick 746cc33f8 Macros: add Path#types and Generic#types
s 2a4e6b6b3 Fix typo
pick 2cb3c1104 Macros: let TypeNode#union_types work for non-union types
```

save file
reword the commit message when the editor appears.

Or use GitKraken :-P (?)

asterite: Wow, so easy! Thank you @oprypin and @bcardiff . This will save me a ton of time!

PR #7972, opened 2019-07-17
Compiler and LLVM refactors and minor additions

bcardiff:
Add methods to emit LLVM's `switch` and helpers for `load volatile` and `store volatile`.

This also removes a method_missing used for delegation in favor of explicit macro expansion. This is easier to track and to customize later if needed.

### DIFF
--- a/src/compiler/crystal/codegen/crystal_llvm_builder.cr
+++ b/src/compiler/crystal/codegen/crystal_llvm_builder.cr
@@ -65,10 +65,18 @@ module Crystal
       @builder.to_unsafe
     end
 
-    macro method_missing(call)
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/codegen/crystal_llvm_builder.cr
```diff
@@ -65,10 +65,18 @@ module Crystal
       @builder.to_unsafe
     end
 
-    macro method_missing(call)
-      return llvm_nil if @end
+    {% for name in %w(add add_handler alloca and ashr atomicrmw bit_cast build_catch_ret call
```
asterite: Thank you! 😍

PR #7977, opened 2019-07-19
Random::ISAAC: fix initial seed

### LINKED ISSUE #7976: Random::ISAAC does not randomize the outcome
The output of the code below should yield different values. However it returns mostly the same value (once in a while a different value is returned, but mostly it is the same.). 

I should point out that I am using Crystal 0.28.0 but based on posting the question in the forum, the bug likely exists even in Crystal 0.29.0.

```crystal
require "random/isaac"

def rnd_alpha_numeric_id()
    srcstr = "abcdefghjkmnpqrstuvwxyz" ## all except i, l and o
    rndinst = Random::ISAAC.new
    rndstr = srcstr.chars
                   .sample(6, rndinst)
                   .join
                
    rndnum = rndinst.rand(100000..999999)
    return "#{rndstr}-#{rndnum}"
end
            
10.times do |i|
    puts "#{i}: #{rnd_alpha_numeric_id()}"
end
```

asterite:
Fixes #7976

In the old code `unsafe_as` returned a copy of the static array and so it was initializing something else other than what was returned.

### DIFF
--- a/spec/std/random/isaac_spec.cr
+++ b/spec/std/random/isaac_spec.cr
@@ -349,4 +349,10 @@ describe "Random::ISAAC" do
   it "can be initialized without explicit seed" do
     Random::ISAAC.new.should be_a Random::ISAAC
   end
+
+  it "different instances generate different numbers (#7976)" do
+    isaacs = Array.new(1000) { Random::ISAAC.new }
+    values = isaacs.map(&.rand(10_000_000))
+    values.uniq.size.should be > 2
+  end
 end
--- a/src/random.cr
+++ b/src/random.cr
@@ -289,6 +289,28 @@ module Random
     end
   end
 
+  # Returns a StaticArray filled with random integers.
+  # Note that this only works with StaticArray of integer types.
+  #
+  # ```
+  # Random.new.rand(StaticArray(Int32, 4))  # => StaticArray[4795590, -1987398873, 1859211601, 546845569]
+  # Random.new.rand(StaticArray(UInt16, 2)) # => StaticArray[49247, 53674]
+  # ```
+  def rand(type : StaticArray(T, N).class) forall T, N
+    {%
+      if T != Int8 && T != UInt8 &&
+         T != Int16 && T != UInt16 &&
+         T != Int32 && T != UInt32 &&
+         T != Int64 && T != UInt64
+        raise "Random::Secure#random_static argument must be StaticArray of integer type"
+      end
+    %}
+
+    buffer = uninitialized UInt8[sizeof(StaticArray(T, N))]
+    random_bytes(buffer.to_slice)
+    buffer.unsafe_as(StaticArray(T, N))
+  end
+
   # Fills a given slice with random bytes.
   #
   # ```
--- a/src/random.cr
+++ b/src/random.cr
@@ -289,26 +289,39 @@ module Random
     end
   end
 
-  # Returns a StaticArray filled with random integers.
-  # Note that this only works with StaticArray of integer types.
-  #
-  # ```
-  # Random.new.rand(StaticArray(Int32, 4))  # => StaticArray[4795590, -1987398873, 1859211601, 546845569]
-  # Random.new.rand(StaticArray(UInt16, 2)) # => StaticArray[49247, 53674]
-  # ```
-  def rand(type : StaticArray(T, N).class) forall T, N
-    {%
-      if T != Int8 && T != UInt8 &&
-         T != Int16 && T != UInt16 &&
-         T != Int32 && T != UInt32 &&
-         T != Int64 && T != UInt64
-        raise "Random::Secure#random_static argument must be StaticArray of integer type"
-      end
-    %}
+  {% for type, values in {
+                           "Int8".id   => %w(20 -66 89 19),
+                           "UInt8".id  => %w(186 221 127 245),
+                           "Int16".id  => %w(-32554 32169 -20152 -7686),
+                           "UInt16".id => %w(39546 44091 2874 17348),
+                           "Int32".id  => %w(1870830079 -1043532158 -867180637 -1216773590),
+                           "UInt32".id => %w(3147957137 4245108745 2207809043 3184391838),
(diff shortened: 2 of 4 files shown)


spec/std/random/isaac_spec.cr
```diff
@@ -349,4 +349,10 @@ describe "Random::ISAAC" do
   it "can be initialized without explicit seed" do
     Random::ISAAC.new.should be_a Random::ISAAC
   end
+
+  it "different instances generate different numbers (#7976)" do
+    isaacs = Array.new(1000) { Random::ISAAC.new }
+    values = isaacs.map(&.rand(10_000_000))
+    values.uniq.size.should be > 2
```
asterite: This spec might fail, but it's veeeeery unlikely.

src/random/isaac.cr
```diff
@@ -12,13 +12,10 @@ class Random::ISAAC
...
   private getter cc
 
-  private alias Seeds = StaticArray(UInt32, 8)
-
   private def random_seeds
-    result = uninitialized Seeds
-    result_slice = result.unsafe_as(StaticArray(UInt8, sizeof(Seeds))).to_slice
-    Random::Secure.random_bytes(result_slice)
-    result
+    StaticArray(UInt32, 8).new do
```
asterite: This is probably a bit slower than before (assuming the old code did it right) but this is on initialization and I prefer safe code even if a bit slower (not in hot path).

src/random/isaac.cr
```diff
@@ -12,13 +12,10 @@ class Random::ISAAC
...
+      Random::Secure.rand(0_u32..UInt32::MAX)
```
Blacksmoke16:
```suggestion
      Random::Secure.rand(UInt32::MIN..UInt32::MAX)
```

ysbaddaden:
What about an uninitialized `UInt8[32]`, fill it in one call to `random_bytes` instead of 8 calls to rand, then cast it to a `StaticArray(UInt32, 8)`?

That would spare 7 unbuffered reads from urandom (or 7 calls to getrandom, arc4random, ...) yet fix the issue, wouldn't it?

asterite:
I think so, yes, but to do that we need to use "unsafe_as". I wanted to avoid unsafe code. But I'll try it anyway.

Another idea is to have such code in Secure::Random, somehow, so that it's unsafe but in a single place.

asterite:
Updated!

I added a general `Random#rand(StaticArray(T, N))` that, for now, only works with integer types. Its implementation uses `unsafe_as` but it's not really unsafe (or it shouldn't be).

Then I used that method in the initialization of `Random::ISSAC` seed.

In the future I'd also like to be able to do `rand(Int32)`, `rand(UInt64)` and so on.

spec/std/random/isaac_spec.cr
asterite: Any other way to test this? Otherwise I can remove the test altogether.

asterite:
I'll do it differently.

asterite:
Added more things and implemented it in a way that's easier to read in docs (now there's one overload per static array type).

Also added the ability to do `rand(Int32)`, `rand(Int8)` and so on, to get a value in that type's range. For that I use `random_bytes` to fill all bytes and then cast it to that type (same thing as with static arrays). I don't know if that's sound but we could eventually change the implementation if needed.

src/random.cr
```diff
@@ -289,26 +289,39 @@ module Random
...
+  {% for type, values in {
+                           "Int8".id   => %w(20 -66 89 19),
+                           "UInt8".id  => %w(186 221 127 245),
+                           "Int16".id  => %w(-32554 32169 -20152 -7686),
+                           "UInt16".id => %w(39546 44091 2874 17348),
+                           "Int32".id  => %w(1870830079 -1043532158 -867180637 -1216773590),
+                           "UInt32".id => %w(3147957137 4245108745 2207809043 3184391838),
+                           "Int64".id  => %w(4438449217673515190 8514493061600538358 -4874671083204037318 -7825896160729246667),
+                           "UInt64".id => %w(15004487597684511003 12027825265648206103 11303949506191212698 6228566501671148658),
+                         } %}
```
Sija: Since these numbers will change anyway isn't it pointless to include them in docs?

asterite: I wanted to include example values. How would you do it otherwise?

src/random.cr
```diff
@@ -289,6 +289,41 @@ module Random
...
+    # Returns a random {{type}}
+    #
+    # ```
+    # rand({{type}}) # => {{values[0].id}}
+    # ```
+    def rand(t : {{type}}.class) : {{type}}
```
Sija:
```suggestion
    def rand(type : {{type}}.class) : {{type}}
```

asterite: Thanks, fixed

src/random.cr
Sija: I'd write the values like `<random {{type}} integer>`.

asterite: I think a real example is better

src/random.cr
```diff
@@ -289,6 +289,41 @@ module Random
...
+                           "Int64".id  => %w(4438449217673515190 8514493061600538358 -4874671083204037318 -7825896160729246667),
+                           "UInt64".id => %w(15004487597684511003 12027825265648206103 11303949506191212698 6228566501671148658),
+                         } %}
+    # Returns a random {{type}}
+    #
+    # ```
+    # rand({{type}}) # => {{values[0].id}}
+    # ```
+    def rand(type : {{type}}.class) : {{type}}
+      rand_type_from_bytes(type)
```
oprypin: It's strange to have a separate implementation for this when `rand(type::MIN..type::MAX)` works fine.

Also kind of funny that `random_bytes` unrolls integers into bytes (and inverts for endinanness) and then this rolls them back into an integer (without inverting for endianness - though double-inverting would also be weird).

So yes, this implementation is endianness-dependent, unlike everything else.

And then, just by chance,
`Random.new(x).rand(Int32::MIN..Int32::MAX) == Random.new(x).rand(Int32)`
but
`Random.new(x).rand(Int64::MIN..Int64::MAX) != Random.new(x).rand(Int64)`

PR #7980, opened 2019-07-21
Fix incorrect Int#% overflow

### LINKED ISSUE #7979: Somehow mod operation causes overflow
Related: #6683 

```crystal
53 % 614_889_782_588_491_410 # -617331001(53 is expected)
```

With `-D preview_overflow` option it raises `OverflowError`.

asterite:
Fixes #7979

Doing `(x ^ y) >= 0` doesn't work well for determining whether `x` and `y` have the same sign when their types are not the same (for example `Int32` and `Int64`).

### DIFF
--- a/src/int.cr
+++ b/src/int.cr
@@ -172,7 +172,7 @@ struct Int
   def %(other : Int)
     if other == 0
       raise DivisionByZeroError.new
-    elsif (self ^ other) >= 0
+    elsif (self < 0) == (other < 0)
       self.unsafe_mod(other)
     else
       me = self.unsafe_mod(other)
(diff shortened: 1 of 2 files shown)


src/int.cr
```diff
@@ -172,7 +172,7 @@ struct Int
   def %(other : Int)
     if other == 0
       raise DivisionByZeroError.new
-    elsif (self ^ other) >= 0
+    elsif (self.sign ^ other.sign) >= 0
```
oprypin: Yes, but what about the equivalent expression `(a<0) == (b<0)`?
https://carc.in/#/r/79lf

asterite: You are right, I was overcomplicating things. Changed, thanks!

Issue #7974, opened 2019-07-18
BigFloat#to_s doesn't show decimal for whole numbers


watzon:
I could be wrong here, but for sake of consistency I think this makes sense. Here's the scenario:

```crystal
# This works as expected
(1.3).to_big_f.to_s
# => "1.3"

# This however does not
(1.0).to_big_f.to_s
# => "1"
```

As you can see, when the second example is transformed to a String it doesn't show the decimal place. As it is a float I would expect it to always show, at the very least, a zero following the decimal.

vlazar:
Shouldn't these give the same results?

```crystal
"1.1".to_big_f.to_s
# => 1.1

1.1.to_big_f.to_s
# => 1.10000000000000008882
```

watzon:
@vlazar I'd think so

Lasvad:
fix for @watzon initially issue.
I started looking at the problem you identified @vlazar. It looks a bit more complex than the first one. Not sure if it should be a separate issue?

Hopefully I submitted the PR in proper form. Scanned over the contributing guide. Please let me know if something should be added.

Cheers.

konovod:
About second problem: maybe it is not a problem but a correct behavior.
1.1 can't be represented exactly in floating form, only approximately. In a Float32 type it is represented as `1.10000000000000008882`, and in a BigFloat it is represented as `1.10000.(many zeros)...0000012345` (sorry, i don't know exact precision and representation of BigFloat). 

When we convert `1.1` from string to BigFloat we get `1.10000.(many zeros)...0000012345` and it is close enough to 1.1 to be converted to 1.1 on output.

Now when we use 1.1 numeral it has an Float32 type and value of `1.10000000000000008882`, that is close to 1.1 for a Float32, but isvery far from 1.1 by BigFloat standards. So it is no correct to convert it from `1.10000000000000008882` to `1.10000.(many zeros)...0000012345` - we just don't have the information that `1.10000000000000008882` is going to represent `1.1` and not say `1.1000000000000001`.

Possible solution is a BigFloat literals, but that is definitely another issue.

asterite:
Another solution is to convert the float to string and then use that as the initial value for the big float. Because `1.1.to_s` gives "1.1".

asterite:
Closed by #7982

PR #7982, opened 2019-07-22
Bugfix: BigFloat#to_s not showing decimals for whole numbers

Lasvad:
add ".0" if value contains no decimal points for to_s.
update specs to reflect having ".0" for whole numbers.

ex.
```
Previous
3.to_big_f.to_s -> "3"
(3.0).to_big_f.to_s -> "3"

Changed to
3.to_big_f.to_s -> "3.0"
(3.0).to_big_f.to_s -> "3.0"

```

Issue #7974

### DIFF
--- a/spec/std/big/big_float_spec.cr
+++ b/spec/std/big/big_float_spec.cr
@@ -4,15 +4,16 @@ require "big"
 describe "BigFloat" do
   describe "new" do
     string_of_integer_value = "123456789012345678901"
+    string_of_integer_value_as_float = "123456789012345678901.0"
     bigfloat_of_integer_value = BigFloat.new(string_of_integer_value)
     string_of_float_value = "1234567890.12345678901"
     bigfloat_of_float_value = BigFloat.new(string_of_float_value)
 
     it "new(String)" do
-      bigfloat_of_integer_value.to_s.should eq(string_of_integer_value)
+      bigfloat_of_integer_value.to_s.should eq(string_of_integer_value_as_float)
       bigfloat_of_float_value.to_s.should eq(string_of_float_value)
-      BigFloat.new("+#{string_of_integer_value}").to_s.should eq(string_of_integer_value)
-      BigFloat.new("-#{string_of_integer_value}").to_s.should eq("-#{string_of_integer_value}")
+      BigFloat.new("+#{string_of_integer_value}").to_s.should eq(string_of_integer_value_as_float)
+      BigFloat.new("-#{string_of_integer_value}").to_s.should eq("-#{string_of_integer_value_as_float}")
       BigFloat.new("123_456_789.123_456_789").to_s.should eq("123456789.123456789")
     end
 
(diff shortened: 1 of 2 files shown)


spec/std/big/big_float_spec.cr
```diff
@@ -190,13 +191,16 @@ describe "BigFloat" do
...
     it { "48600000".to_big_f.to_s.should eq("48600000") }
     it { "12345678.87654321".to_big_f.to_s.should eq("12345678.87654321") }
     it { "9.000000000000987".to_big_f.to_s.should eq("9.000000000000987") }
-    it { "12345678901234567".to_big_f.to_s.should eq("12345678901234567") }
-    it { "1234567890123456789".to_big_f.to_s.should eq("1234567890123456789") }
+    it { "12345678901234567".to_big_f.to_s.should eq("12345678901234567.0") }
+    it { "1234567890123456789".to_big_f.to_s.should eq("1234567890123456789.0") }
+
+    it { 3.to_big_f.to_s.should eq("3.0") }
+    it { (3.0).to_big_f.to_s.should eq("3.0") }
```
rodrigopinto: I would write more test cases covering more cases. I was looking into this issue yesterday, and all the specs we have until now were covering `String` to `BigFloat`. The issue aims to cover `Float32 | Float64` to `BigFloat` and I wind up with trick responses like:

`it { 1.3.to_big_f.to_s.should eq("1.3") }`

```shell
Expected: "1.3"
got: "1.30000000000000004441"
```

`it { 0.000001.to_big_f.to_s.should eq("0.000001") }`

```shell
Expected: "0.000001"
got: "0.000000999999999999999954748"
```

For those cases, the numbers are Float64, and there is does not have an initializer on the BigFloat that covers `Floats` without passing the precision. For now, I extend more the test cases for floating numbers to see what need to be covered

asterite: @rodrigopinto That's a relevant suggestion but it's unrelated to this PR's fix.

The problem you are describing is the result value of when you initialize a BigFloat with a Float. Not a part of this PR. This PR is good as is.

rodrigopinto: I also saw there were some discussions about it on the issue. So, let's move ahead then.

vlazar:
With changes on PR these still fail to add `.0`
```
"10".to_big_f.to_s # => 10
"100".to_big_f.to_s # => 10
"150".to_big_f.to_s # => 150
```

Lasvad:
hmm...

``` 
it { "1234567890123456789".to_big_f.to_s.should eq("1234567890123456789.0") }
```
this test passed so im surprised that 

```
"10".to_big_f.to_s # => 10
```
doesn't work.

I'll take a look as soon as I can and put a fix in.

vlazar:
@Lasvad Any numbers ending with 0 won't add `.0` at the end.

asterite:
My suggestion for implementing this: keep a bool that tracks whether `.` was printed. If not, print it at the end.

Lasvad:
@vlazar thanks for confirming the source of the error. I'll add your test cases when I implment the fix.
@asterite Cool. I'll look into implementing something like that in a few hours when I get home.

Lasvad:
I've made the changes and used a flag to check if a decimal was set.

src/big/big_float.cr
```diff
@@ -279,20 +279,28 @@ struct BigFloat < Float
...
       next if cstr[i] == 45 # '-'
-      io << '.' if i == expptr
+      if i == expptr
+        io << '.'
+        decimal_set = true
+      end
       io << cstr[i].unsafe_chr
     end
     (expptr - length).times { io << 0 } if expptr > 0
+    if !decimal_set
```
Sija:
```suggestion
    unless decimal_set
```

asterite: Not needed, they are the same.

Issue #7293, opened 2019-01-10
build failure aarch64


silmanduin66:
I am trying to build and run a simple HelloWorld.cr
arch : aarch64
OS : ArchLinux
Device : Raspberry pi 3 ( ArmV8)
LLVM : 6.0.1
Crystal: 0.27.0

this is the error:

```[root@archlinuxPi1 crystal]# crystal build hello.cr
Invalid memory access (signal 11) at address 0xaaaac1cf0000
[0xaaaabc56cb88] ???
[0xaaaabbded850] __crystal_sigfault_handler +208
[0xaaaabc5700e8] sigfault_handler +40
[0xffffbc27c66c] ???
[0xffffb81b8300] memset +256
[0xaaaabc10edf8] ???
[0xaaaabc10dfe8] ???
[0xaaaabc14507c] ???
[0xaaaabc18b604] ???
[0xaaaabc1445d0] ???
[0xaaaabc141998] ???
[0xaaaabc13f444] ???
[0xaaaabc4f8f5c] ???
[0xaaaabbe01488] ???
[0xaaaabbdd1e9c] __crystal_main +10728
[0xaaaabbdd5534] main +44
[0xffffb815a2a4] _libc_start_main +228
[0xaaaabbdce3e4] ???
```

I crossed-compiled crystal from my ArchLinux PC with LLVM 6.0.1 and crystal 0.27.0, transfered the crystal.o to my aarch64 based device and compiled there. Everything worked fine.
I installed all the necessary libs.

the command ```crystal -v``` gives me :

```
Crystal 0.27.0 (2019-01-10)
LLVM: 6.0.1
Default target: aarch64-unknown-linux-gnu
```

silmanduin66:
I am trying to build and run a simple HelloWorld.cr 
arch : aarch64
OS : ArchLinux
Device : Raspberry pi 3  ( ArmV8)
LLVM : 6.0.1
Crystal: 0.27.0

this is the error:

`[root@archlinuxPi1 crystal]# crystal build  hello.cr
Invalid memory access (signal 11) at address 0xaaaac1cf0000
[0xaaaabc56cb88] ???
[0xaaaabbded850] __crystal_sigfault_handler +208
[0xaaaabc5700e8] sigfault_handler +40
[0xffffbc27c66c] ???
[0xffffb81b8300] memset +256
[0xaaaabc10edf8] ???
[0xaaaabc10dfe8] ???
[0xaaaabc14507c] ???
[...]

Issue #7299, opened 2019-01-11
cross-compilation to aarch64 mysterious error


silmanduin66:
This is not a real issue as it sometimes works

Crystal: 0.27.0
LLVM: 6.0.1
OS: archLinux
arch: x86-64 to aarch64 (PC to raspberry Pi3)

I cross-compile my crystal files to aarch64 , and i get this error every time i cross-compile a file for the 1st time. But if i try to cross-compile it again , it mysteriously works. ( sometimes i have to do it 3 times in a row)

```
[root@archLinux blogsy]# crystal build src/blogsy.cr --cross-compile --target "aarch64-unknown-linux-gnu"
Invalid memory access (signal 11) at address 0x7ffcbbf770b0
[0x5599e0bcfe86] ???
[0x5599e032123e] __crystal_sigfault_handler +238
[0x7fdd154313c0] ???
[0x7fdd15314f10] ???
[0x7fdd16220380] _ZN4llvm15SmallVectorBase8grow_podEPvmm +128
[0x7fdd175b3d2a] _ZNK4llvm19AArch64CallLowering17splitToValueTypesERKNS_12CallLowering7ArgInfoERNS_15SmallVectorImplIS2_EERKNS_10DataLayoutERNS_19MachineRegisterInfoEjRKSt8functionIFvjmEE +1146
[0x7fdd175b4116] _ZNK4llvm19AArch64CallLowering9lowerCallERNS_16MachineIRBuilderEjRKNS_14MachineOperandERKNS_12CallLowering7ArgInfoENS_8ArrayRefIS7_EE +326
[0x7fdd169acbbb] _ZNK4llvm12CallLowering9lowerCallERNS_16MachineIRBuilderENS_17ImmutableCallSiteEjNS_8ArrayRefIjEESt8functionIFjvEE +1387
[0x7fdd169b4c8c] _ZN4llvm12IRTranslator13translateCallERKNS_4UserERNS_16MachineIRBuilderE +1132
[0x7fdd169b83fe] _ZN4llvm12IRTranslator20runOnMachineFunctionERNS_15MachineFunctionE +2894
[0x7fdd16554870] _ZN4llvm19MachineFunctionPass13runOnFunctionERNS_8FunctionE +144
[0x7fdd16369b98] _ZN4llvm13FPPassManager13runOnFunctionERNS_8FunctionE +792
[0x7fdd16369c12] _ZN4llvm13FPPassManager11runOnModuleERNS_6ModuleE +50
[...]

PR #7983, opened 2019-07-22
Fixes issue with aarch64 invalid memory access

drujensen:
This change properly maps the 64bit Intrinsics memory methods for `aarch64` architecture.

This fixes issue #7293 and possibly #7299

### DIFF
--- a/src/intrinsics.cr
+++ b/src/intrinsics.cr
@@ -1,6 +1,6 @@
 lib Intrinsics
   fun debugtrap = "llvm.debugtrap"
-  {% if flag?(:x86_64) %}
+  {% if flag?(:x86_64) || flag?(:aarch64) %}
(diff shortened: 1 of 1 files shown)


asterite:
Thank you!

I think we have a bits64 flag. It's checked in other parts of the code.

asterite:
Maybe also fixes #7983

src/intrinsics.cr
```diff
@@ -1,6 +1,6 @@
 lib Intrinsics
   fun debugtrap = "llvm.debugtrap"
-  {% if flag?(:x86_64) %}
+  {% if flag?(:x86_64) || flag?(:aarch64) %}
```
asterite:
```suggestion
  {% if flag?(:bits64) %}
```

ysbaddaden:
Thank you 🙇

PR #7987, opened 2019-07-23
Add support for LLVM 8

bcardiff:
~~The `IsLocalToUnit`, `IsDefinition`, `IsOptimized` were always used with the same values and the API regarding them changed in llvm 8.0.~~ 

Update: Use `DISubprogram::toSPFlags` to keep functionality & LibLLVMExt API

t - 2 months before llvm 9.0 is released.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -89,9 +89,16 @@ prepare_build() {
   on_osx curl -L https://github.com/crystal-lang/crystal/releases/download/0.29.0/crystal-0.29.0-1-darwin-x86_64.tar.gz -o ~/crystal.tar.gz
   on_osx 'pushd ~;gunzip -c ~/crystal.tar.gz | tar xopf -;mv crystal-0.29.0-1 crystal;popd'
 
-  on_osx brew install llvm@7 gmp libevent pcre pkg-config
+  on_osx brew install llvm@8 gmp libevent pcre pkg-config
   on_osx brew upgrade libyaml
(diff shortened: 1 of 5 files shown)


bcardiff:
I've found a better way to implement it via `DISubprogram::toSPFlags` 💯

refi64:
This is actually really sweet, I was working on a Crystal Flatpak SDK a few days ago, but the freedesktop sdk 19.08 beta only ships LLVM 8 so I couldn't use it yet.

bin/ci
```diff
@@ -70,6 +70,7 @@ prepare_system() {
   on_linux sudo service docker restart
 
   on_osx brew update
+  on_osx brew update
```
bew: Why 2 times?

bcardiff: I'm debugging why llvm@8 is not appearing in CircleCI.

anatol:
Awesome, thank you for porting the codebase to LLVM8.

While we are here, is there any chance to prepare crystal for upcoming LLVM9? Like building the project with LLVM master and see if there are any API changes that can be fixed in advance, before LLVM9 becomes public.

bcardiff:
@anatol There could be preliminary support. Since LLVM 9 is not yet released and there might be pending changes I personally prefer to wait for a release. But definitely, PRs are welcome to have initial support for it if someone wants to tackle it earlier.

PR #7994, opened 2019-07-25
Update distribution-scripts

bcardiff:
This update will make darwin .tag.gz and .pkg official packages to use llvm 6.0.1 (instead of 3.9.1 👻)

This is not upgrading to llvm 8 since its support is landing in this very release.


bcardiff:
For homebrew users, the formula will be updated to llvm 8 there upon release.

bcardiff:
A 0.30.0-2-darwin package was uploaded that uses llvm 3.9.1 due to https://bugs.llvm.org/show_bug.cgi?id=42932

Issue #5430, opened 2017-12-21
Coordinate porting to Windows


straight-shoota:
*UPDATE 2023-04-24:*

*All the main platform features for Windows are finished! 🚀*

*There are still some smaller stories pending. Project board: https://github.com/orgs/crystal-lang/projects/11/views/5*

*You can support the ongoing development financially: [Windows suppport](https://opencollective.com/crystal-lang/projects/windows-support) project on Open Collective.*

*UPDATE 2021-11-18:*

*By now, the compiler works pretty well on windows and most of the standard library has been ported as well. Besides a couple other smaller missing pieces, the major gap is in concurrency-related features held back by the missing event loop implementation.*

-----

With #5339 being merged, the first step of bringing Crystal to Windows has been accomplished!
With current master branch it is possible to cross-compile and run simple programs (alá *Hello, World!*) on native Windows.

Obviously we're still far away from porting the compiler or entire standard library to windows. This issue is supposed to coordinate and keep track of ongoing efforts to add support for more and more of the crystal standard library.

# TODO

The primary focus should be on the core library (somewhat odered by priority):

* [x]  File descriptors #5339
* [x] GC #5339
* [x] Errno #5339 
* [x] Basic Exceptions #5339 
* [x] Regex #5339
* [x] Random #5539 
* [x]  Time #5422 
* [x]  Raising #6419 
* [x]  File #5623
* [x]  Dir #5623 
* [x] ENV #6333
* [x] Fibers #7995
* [x] Event loop #12149
* [x]  Signal #7339
* [x]  Mutex  #12213
* [x]  Sockets #10696
* [x]  Process
* [x]  Backtraces #11461
* [x]  Iconv (#11480)
* [x]  System #6972
* [x]  Threads (#11647)
* [x] Compiler #9032

And of course a CI infrastructure once specs are running on windows.

# Proceedings

The general course of action is to extract platform-specific implementations and then add the implementation for win32 platform.

A lot of work has already been done in #3582 but it needs to be updated and chopped into smaller pieces of changes.

The first goal should probably be to get specs running on windows to be able to verify the windows implementations. Essential for this are as far as I can see file, dir, exceptions (backtrace!), process and signal.

If you want to help with that effort, or want to "take" a piece of that work to work on, please comment below.

# Useful References 

* Wiki page [*Porting to Windows*](https://github.com/crystal-lang/crystal/wiki/Porting-to-Windows)
[...]

PR #7995, opened 2019-07-25
Implement fibers for win32

straight-shoota:
This PR continues the efforts to bring concurrency features to windows (see #5430).

There are some refactorings to the the existing Fiber, Thread and EventLoop classes to extract the platform-specific implementation focused on POSIX and libevent.

Then it adds implementations of these classes for windows. `Thread` and `EventLoop` are only stubbed so far.
The core functionality is the context swap on win32.

`concurrent_spec` passes and channels work in general, but the channel spec seems to hang.

Supersedes #6955


src/fiber.cr
```diff
@@ -46,9 +45,21 @@ class Fiber
...
+    # FIXME: This line shouldn't be necessary
+    stack_ptr = nil
+    {% if flag?(:win32) %}
+      # align stack bottom to 16 bytes
+      @stack_bottom = Pointer(Void).new(@stack_bottom.address & ~0x0f_u64)
+
+      # It's the caller's responsibility to allocate 32 bytes of "shadow space" on the stack right
+      # before calling the function (regardless of the actual number of parameters used)
+
+      stack_ptr = @stack_bottom - sizeof(Void*) * 4
```
cfsamson: I have been thinking a bit about the shadow space, since at first I didn't think that we needed it at all. I have concluded that we do need it but I think the way we do it here gives the wrong end result. 

Right now we first reserve 4 bytes of shadow space, and then write `rsp` over one of them when writing `stack_ptr[0] = fiber_main.pointer`. I think we have to do like I show in this diagram since the next 16 byte aligned offset is 48 bytes from the bottom of the stack.
Skjermbilde 2019-07-26 kl  23 25 14
(it's just exemplified with "addresses" on a 64 byte stack).

The fact that we don't have a return address should not matter, the important part is that `rsp` ends up in the shadow space which might be used when debugging for example. `stack_ptr = @stack_bottom - sizeof(Void*) * 6` should avoid that.

Take a look at how [Boost Context does it](https://github.com/boostorg/context/blob/d4608a4e8ba2e8a0bdd45613243e6c53a23d23dc/src/asm/ontop_x86_64_ms_pe_gas.asm#L190). I linked to the place int the asm where they set up`rsp` and compare that offset with the position in the chart in the top of the file.
 
@ysbaddaden had some questions about the shadow space earlier and i have some references here besides the Boost Context one:
https://docs.microsoft.com/en-us/cpp/build/stack-usage?view=vs-2019#stack-allocation
https://github.com/simon-whitehead/assembly-fun/tree/master/windows-x64#shadow-space
https://wiki.lazarus.freepascal.org/Win64/AMD64_API#Shadow_space

straight-shoota:
@ysbaddaden in https://github.com/crystal-lang/crystal/pull/6955#issuecomment-515912947 (answered here in the new PR):

> About the rework: I meant that POSIX and Win32 don't need to share a base Thread class. The amount of duplicated code isn't large enough to justify the split. That would allow them to have their own internal details, so we treat them as "opaque classes".

Sure, we can do that. I don't know yet what the proper `Thread` implementation for win32 will look like, so I just did the minimal required thing to get fibers working.

We need to consider: Once we get proper multithreading, I suppose `Thread` might become a public API. I'm not sure if that's planned yet, but at least I think it would make sense to allow access to threads directly. In that case, we'll need to define a single public interface, like we do with other system-specific types.

> I'd also like this rework to settle on where we should put specific classes, namely Thread, Thread::Mutex, Thread::ConditionVariable (pthread, win32), EventLoop (libevent, win32, then probably epoll, kqueue, ...) to avoid further conflicts.
> 
> Being system specific, I suppose they should be moved under src/crystal/system?
> 
> For threads, I guess they should be unix/pthread.cr, unix/pthread_mutex.cr, win32/thread.cr, and so on, then have src/thread.cr, src/thread/mutex.cr require the correct one?

Sounds good.

> For the event loop, maybe unix/libevent.cr and win32/iocp.cr then have src/crystal/event_loop.cr require the correct one?

I'd like to include `event_loop` in the system-specific file names. So `unix/event_loop_libevent.cr` and `win32/event_loop_iocp.cr`.
For now, I've moved them all to `src/crystal/event_loop` because it's not exactly system-specific. Basically, the `libevent` bindings for unix should in theory work with win32 as well. But yeah, it probably still makes more sense to put that in `src/crystal/system/{unix,win32}`.

> Just a question: would it be hard to bring real win32 thread support, instead of bringing a stub? I know that's not needed to implement fibers, but threads aren't needed on POSIX either, and it would be awesome to have a basic support for them.

I really don't know. Haven't looked into that yet. And I'd rather like to get fibers working right away because that's the limiting factor for porting other features as well.

src/thread.cr
```diff
@@ -1,16 +1,19 @@
-require "c/pthread"
-require "c/sched"
-require "./thread/*"
+require "./thread/linked_list"
+{% if flag?(:win32) %}
+  {% raise "thread not supported" %}
```
RX14: This feels like it should be `Crystal::System`...

src/crystal/event_loop.cr
```diff
@@ -1,60 +1,30 @@
...
 
-  def self.create_dns_request(nodename, servname, hints, data, &callback : LibEvent2::DnsGetAddrinfoCallback)
-    dns_base.getaddrinfo(nodename, servname, hints, data, &callback)
-  end
+  # Adds a new timeout to this event.
+  # def add(time_span : Time::Span?) : Nil
 end
+
+{% if flag?(:unix) %}
+  require "./event/event_loop_libevent"
```
RX14: This feels like it should be in `Crystal::System` too, but less so.

src/lib_c/x86_64-windows-msvc/c/winbase.cr
```diff
@@ -90,4 +90,16 @@ lib LibC
   fun GetEnvironmentStringsW : LPWCH
   fun FreeEnvironmentStringsW(lpszEnvironmentBlock : LPWCH) : BOOL
   fun SetEnvironmentVariableW(lpName : LPWSTR, lpValue : LPWSTR) : BOOL
+
+  MEM_COMMIT     = 0x00001000
+  MEM_RESERVE    = 0x00002000
+  MEM_RESET      = 0x00080000
+  MEM_RESET_UNDO = 0x01000000
+
+  fun VirtualAlloc(lpAddress : Void*, dwSize : SizeT, flAllocationType : DWORD, flProtect : DWORD) : Void*
```
RX14: These two functions are actually in `memoryapi.h`, according to MSDN, not winbase

src/crystal/scheduler.cr
```diff
@@ -97,11 +97,23 @@ class Crystal::Scheduler
...
+    print_error "\nFATAL: #{message}: #{fiber}\n"
+    {% unless flag?(:win32) %}
+      # FIXME: Enable when caller is supported on win32
+      caller.each { |line| print_error "  from #{line}\n" }
+    {% end %}
+
     exit 1
   end
 
+  private def print_error(message)
```
RX14: Maybe move this to `Crystal::System.print_error`?

src/fiber/context/x86_64.cr
```diff
@@ -3,39 +3,78 @@
...
-      pushq %rdi        // push 1st argument (because of initial resume)
-      pushq %rbx        // push callee-saved registers on the stack
-      pushq %rbp
-      pushq %r12
-      pushq %r13
-      pushq %r14
-      pushq %r15
-      movq %rsp, 0($0)  // current_context.stack_top = %rsp
-      movl $$1, 8($0)   // current_context.resumable = 1
+    {% if flag?(:win32) %}
```
RX14: Should we have a `x86_64-windows-msvc.cr` and a `x86_64-linux-gnu.cr`?

straight-shoota: Sounds reasonable.

straight-shoota:
I pushed some updates but didn't move any files yet. I'm waiting for feedback on this before pushing things around.

ysbaddaden:
@straight-shoota `Thread` is meant to stay has a private internal API, there is no plan to eventually make it public.

src/fiber/context/x86_64.cr
ysbaddaden: Why `x86_64-linux-gnu.cr`? The context switch is valid for all System-V targets (linux gnu/musl, openbsd, freebsd, ...). 

Are there alternatives to win32-msvc ABI, for example is MinGW different? In that case probably extract `x86_64-windows-msvc.cr` and have a`x86_64-posix.cr`.

straight-shoota: `x86_64-posix.cr` => `x86_64-unix.cr`? We're typically using `unix` everywhere else...

ysbaddaden: Oh yeah, sure.

RX14: I didn't know the standard abi had a name! If it's called systemv, then it should be named that!

`x86_64-sysv.cr` and `x86_64-microsoft.cr` would be my preference

straight-shoota: That naming looks good, see https://en.wikipedia.org/wiki/X86_calling_conventions#x86-64_calling_conventions

straight-shoota:
I've moved `Thread` and `EventLoop` to `crystal/system`, split the `x86_64` fiber context  implementation and also squashed all those fixup commits.

src/crystal/system/print_error.cr
```diff
@@ -0,0 +1,12 @@
+module Crystal::System
+  # Prints directly to stderr without going through an IO.
+  # This is useful for error messages from components that are required for
+  # IO to work (fibers, scheduler, event_loop).
+  def self.print_error(message)
```
Sija: I'd add a simple logic for appending `\n` to the *message*, since it's the most common usage case.
Perhaps sth along the lines of `message = message + "\n" unless message.ends_with?('\n')`?

src/prelude.cr
```diff
@@ -79,7 +79,7 @@ require "static_array"
 require "struct"
 require "symbol"
 no_win require "system"
-no_win require "thread"
+require "thread"
```
ysbaddaden: Shouldn't it become `crystal/thread.cr`?

straight-shoota:
Yes, exactly. It kind of makes sense to have `Thread` in the `Crystal` namespace, but that breaks some things. That's the reason why the tests keep failing because there are ever more places with broken references. It's probably worth it, though. When it's not intended to be public API, then it should go into the `Crystal` namespace like everything else.

The PR changes everything to `Crystal::Thread` but that's easy to swing back should we decide to keep it in the top level namespace.

src/windows_stubs.cr
```diff
@@ -73,3 +73,11 @@ end
 def sleep(time : Time::Span)
   LibC.Sleep(time.total_milliseconds)
 end
+
+class Crystal::Thread
```
RX14: Perhaps move this to `src/crystal/system/win32/thread.cr`

fiber.cr
```diff
@@ -0,0 +1,22 @@
+puts "running"
```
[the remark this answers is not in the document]
straight-shoota: Definitely :+1:

src/windows_stubs.cr
straight-shoota: IIRC you previously asked to have this stub in `windows_stubs` :D

src/crystal/system/event_loop.cr
```diff
@@ -0,0 +1,32 @@
+module Crystal::EventLoop
```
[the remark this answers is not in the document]
straight-shoota: I figured the `Crystal::System` namespace is only really useful when there are both a system-specific type and a general type in `Crystal`. Otherwise it only adds clutter.
There is however the possibility of name clashes with the compiler which also uses the `Crystal` namespace. But that's a general issue unless we put *everything* in `Crystal::System`.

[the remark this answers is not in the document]
RX14: I put everything which has a platform-specific implementation - and is using the new `Crystal::System` structure for it in `Crystal::System`.

I'm fine with not renaming, but I personally would have done it, since the files are in `src/crystal/system`.

src/windows_stubs.cr
RX14: haha, well, oops

I think that was when everything was ifdefs?

Either way, I think i've changed my mind since the rest of `Thread` is a stub too. What do you think? It's not particularly important either way.

ysbaddaden: I agree, let's prepare for the win32 implementations by having stubs defined at their expected place, with a TODO: implement :-)

src/crystal/system/event_loop.cr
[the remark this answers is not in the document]
straight-shoota: I see it that way that the things in `Crystal::System` are not used anywhere directly except as an implementation detail of their specific counterpart in the public API. `EventLoop`, `Thread` etc. however are general APIs for Crystal's runtime. They only have system-specific implementations, but they're directly used in the runtime. `Crystal::EventLoop` for example is referenced 10 times in the stdlib (without specs), `Crystal::Thread` 17 times.

src/crystal/system/print_error.cr
straight-shoota: This is a very low-level method. I think it is fine without convenient, but ultimately unnecessary logic enhancements.

straight-shoota:
Rebased and squashed.

RX14:
and CI failed

ysbaddaden:
> Unhandled exception: BUG: Thread.current returned NULL (Exception)

It fails to build crystal itself (ECR.embed for the init command).

straight-shoota:
@ysbaddaden `check_format` fails while building the compiler, but that's the only thing being built in that spec. All `test` checks fail while executing `ecr/process` in a `run` macro for `ecr_spec`. `ecr/process` is probably just the first crystal program executed with the changed (local) standard library-

To reproduce the error directly:
```console
$ bin/crystal run src/ecr/process.cr -- spec/std/data/test_template2.ecr
Unhandled exception: BUG: Crystal::Thread.current returned NULL (Exception)
  from src/raise.cr:191:1 in 'raise'
  from src/raise.cr:229:3 in 'raise'
  from src/crystal/system/unix/pthread.cr:64:3 in 'current'
  from src/crystal/scheduler.cr:17:5 in 'enqueue'
  from src/concurrent.cr:63:3 in 'spawn'
  from src/kernel.cr:540:1 in '__crystal_main'
  from __libc_start_main
  from _start
  from ???
```

Out of a hunch, I re-inserted `require "thread"` into [scheduler.cr](https://github.com/crystal-lang/crystal/pull/7995/commits/f47ec7f6bfdc2d646f0ba109230baf277dd547d0#diff-3920767d4904eb70109bd332570d8cc4L3) (which seemed unnecessary at first), but it makes the error vanish. I guess there is some effect from require order. CI should pass now.

src/crystal/system/win32/thread.cr
```diff
@@ -0,0 +1,74 @@
...
+
+  @@current : Thread? = nil
+
+  # Associates the Thread object to the running system thread.
+  protected def self.current=(@@current : Thread) : Thread
+  end
+
+  # Returns the Thread object associated to the running system thread.
+  def self.current : Thread
+    @@current || raise "BUG: Thread.current returned NULL"
```
Sija:
```suggestion
    @@current || raise "BUG: Crystal::Thread.current returned NULL"
```

straight-shoota:
Rebased on master with #8032 to see test_linux32 succeed.

src/fiber.cr
```diff
@@ -46,9 +45,21 @@ class Fiber
 
     fiber_main = ->(f : Fiber) { f.run }
 
-    # point to first addressable pointer on the stack (@stack_bottom points past
-    # the stack because the stack grows down):
-    stack_ptr = @stack_bottom - sizeof(Void*)
+    # FIXME: This line shouldn't be necessary (#7975)
```
asterite: One way to remove this FIXME is to do the if flag check twice, but the second one inside the assignment to stack_ptr. But it's not necessary, I think this way is fine too.

asterite:
Once we merge this, how far would we be regarding running Crystal in Windows?

wontruefree:
@asterite looks like there are eight more bullet points to work on.
https://github.com/crystal-lang/crystal/issues/5430

asterite:
This now needs to be rebased on top of `master`.

src/crystal/system/thread_linked_list.cr
```diff
@@ -1,13 +1,13 @@
-require "thread/mutex"
+require "./thread_mutex"
 
 # :nodoc:
-class Thread
+class Crystal::Thread
```
asterite: Could we revert this change and leave `Thread` instead of `Crystal::Thread`? I don't think we can merge this PR with this. But @waj and @bcardiff might be able to provide more context about why we don't want to make this change.

RX14: It should at least be another PR, honestly.

I like the change though.

asterite: Maybe. I just don't think we should delay Windows support just because some want to rename a class and some don't.

straight-shoota: Okay, I'll revert to `Thread`.

src/crystal/system/win32/event_loop_iocp.cr
```diff
@@ -0,0 +1,70 @@
+require "crystal/system/print_error"
+
+module Crystal::EventLoop
+  @@queue = Deque(Fiber).new
+
+  private def self.loop_fiber
```
ysbaddaden: The EventLoop API changed a lot with MT. There is no more loop fiber, and less methods.

straight-shoota:
I've updated the win32 implementation to reflect the recent changes to `EventLoop` from #8173.
Also rebased on top of #8195 to get it compile on win32.

RX14:
Another coflict?

RX14:
would love another review

src/fiber/context/x86_64-microsoft.cr
```diff
@@ -0,0 +1,78 @@
...
+          popq %r15         // pop callee-saved registers from the stack
+          popq %r14
+          popq %r13
+          popq %r12
+          popq %rsi
+          popq %rbp
+          popq %rbx
+          popq %rdi         // pop 1st argument (for initial resume)
+          popq %gs:0x08
+          popq %gs:0x10
```
firejox: The order it popped is different from the order it stored in `makecontext`. Is it correct?

ysbaddaden: Sorry, I didn't mean to resolve the discussion....

I believe it's a documentation error of setting rbx in makecontext (or we push rbx when we don't need to?).

We only pop up till rcx and the return address is left on the stack, I think.

firejox: @ysbaddaden  Well, I meant these `popq %gs:0x08` and `popq %gs:0x10` two instructions , not about `rbx` and `rcx`. Sorry for including too large region. 

In `makecontext`, they store in the following order.
```
    # The following two values are stored in the Thread Information Block (NT_TIB)
    # and are used by Windows to track the current stack limits
    stack_ptr[-2] = @stack_bottom # %gs:0x08: Stack Bottom
    stack_ptr[-3] = @stack        # %gs:0x10: Stack Limit
```

Shouldn't the pop order be this?
```
  popq %gs:0x10
  popq %gs:0x08
```

cfsamson: I think you're right. `GS:[0x08]` is the "high address", and `GS:[0x10]` is the "low address" and if I'm not mistaken `@stack_bottom` is the high address, and `@stack` is the low address so they're put on the initial stack in the wrong order (or we need to change both push and pop orders in `asm`).

I would just change the initial setup to:
```
stack_ptr[-2] = @stack              # %gs:0x10: Stack Limit
stack_ptr[-3] = @stack_bottom # %gs:0x08: Stack Bottom
```
@ysbaddaden might be right that we don't need to push/pop `%rcx` on every switch. I don't have the setup here to confirm that, but it might be a few cycles wasted there.

ysbaddaden: Ah yes. They're inverted in makecontext. I totally missed that.

PR #7996, opened 2019-07-26
Integer: Improve #gcd types and improve performance a bit.

### LINKED ISSUE #6683: integer overflow in gcd function

yxhuvud:
This provide an implementation of binary gcd.
If called with different types, gcd will no longer respond with a
union type. EDIT: Now restricted to only accept the same type, to avoid method clutter. 

Also adds tests for gcd - previously only BigInt#gcd and #lcm had tests.

Alternative to #6694
Would resolve #6683

### DIFF
--- a/src/int.cr
+++ b/src/int.cr
@@ -395,10 +395,9 @@ struct Int
     (self & mask) == mask
   end
 
-  # Returns the greatest common divisor of `self` and `other` (which
-  # is an `Int` of some kind). The return type is whichever side can
-  # fit the biggest values. If either is signed and has value equal to
-  # `MIN` of its type, then overflow will be raised.
+  # Returns the greatest common divisor of `self` and `other`. Signed
+  # integers may raise overflow if either has value equal to `MIN` of
+  # its type.
   #
   # ```
   # 5.gcd(10) # => 2
--- a/src/int.cr
+++ b/src/int.cr
@@ -395,8 +395,47 @@ struct Int
     (self & mask) == mask
   end
 
-  def gcd(other : Int)
-    self == 0 ? other.abs : (other % self).gcd(self)
+  # Returns the greatest common divisor of `self` and `other` (which
+  # is an `Int` of some kind). The return type is whichever side can
+  # fit the biggest values. If either is signed and has value equal to
+  # `MIN` of its type, then overflow will be raised.
+  #
+  # ```
+  # 5.gcd(10) # => 2
+  # 5.gcd(7)  # => 1
+  # ```
+  def gcd(other : self) : self
+    u = self.abs
+    v = other.abs
+    shift = self.class.zero
+    return v if u == 0
+    return u if v == 0
+
+    # Let shift := lg K, where K is the greatest power of 2
+    # dividing both u and v.
+    while (u | v) & 1 == 0
+      shift &+= 1
+      u = u.unsafe_shr 1
+      v = v.unsafe_shr 1
+    end
+    while u & 1 == 0
+      u = u.unsafe_shr 1
+    end
+    # From here on, u is always odd.
+    loop do
+      # remove all factors of 2 in v -- they are not common
+      # note: v is not zero, so while will terminate
+      while v & 1 == 0
+        v = v.unsafe_shr 1
+      end
+      # Now u and v are both odd. Swap if necessary so u <= v,
+      # then set v = v - u (which is even).
+      u, v = v, u if u > v
+      v &-= u
+      break if v.zero?
+    end
+    # restore common factors of 2
+    u.unsafe_shl shift
   end
 
   def lcm(other : Int)
--- a/src/int.cr
+++ b/src/int.cr
@@ -399,6 +399,9 @@ struct Int
   # integers may raise overflow if either has value equal to `MIN` of
   # its type.
   #
+  # Implementation heavily inspired by
+  # https://en.wikipedia.org/wiki/Binary_GCD_algorithm#Iterative_version_in_C
+  #
   # ```
   # 5.gcd(10) # => 2
   # 5.gcd(7)  # => 1
--- a/src/int.cr
+++ b/src/int.cr
@@ -406,10 +406,10 @@ struct Int
   def gcd(other : self) : self
     u = self.abs
     v = other.abs
-    shift = self.class.zero
     return v if u == 0
     return u if v == 0
 
+    shift = self.class.zero
     # Let shift := lg K, where K is the greatest power of 2
     # dividing both u and v.
     while (u | v) & 1 == 0
--- a/spec/std/big/big_int_spec.cr
+++ b/spec/std/big/big_int_spec.cr
@@ -334,19 +334,15 @@ describe "BigInt" do
     a_17 = a * 17
 
     (abc * b).gcd(abc * c).should eq(abc)
+    abc.gcd(a_17).should eq(a)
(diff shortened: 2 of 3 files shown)


straight-shoota:
This adds way too much complexity for a method with questionable usefulness.
Previously, it was one-line. Now it's dozens LOC and multiple method overload for each number type. That doesn't seem reasonable to me.

I'd strictly prefer #6694 over this.

yxhuvud:
Would you prefer a solution that always cast the argument to the self type? That would take a lot less code, cause a lot more overflows, but at least not have union type issues.

straight-shoota:
Without a real use case, I'd rather remove it entirely.

If you want to do anything serious, you'll probably use `BigInt` anyway.

yxhuvud:
If you want something serious, it is probably to solve questions like "I have this thing with a period of n, and this other thing with a period of m, when will they coincide?", or it is working in modulo fields. 

Neither of which usually need BigInt.

asterite:
I agree with @straight-shoota . I believe this is useful but I think the complexity it brings (mainly we'd have to maintain that) vs. the actual number of times you need it might not be worth it. It's totally acceptable as a shard, though.

yxhuvud:
How about restricting arguments to be of the same type? That would remove the type complexity of accidental union types and would be more obvious to the user?

asterite:
@yxhuvud Oh, that's a very good idea. Then users will have to take care of matching the two types, which should be pretty straight-forward (just extend the smaller type to the larger type).

I like it :-)

@straight-shoota What do you think?

src/int.cr
```diff
@@ -357,8 +357,47 @@ struct Int
...
+  def gcd(other : self) : self
+    u = self.abs
+    v = other.abs
+    shift = self.class.zero
+    return v if u == 0
+    return u if v == 0
+
+    # Let shift := lg K, where K is the greatest power of 2
+    # dividing both u and v.
+    while (u | v) & 1 == 0
```
yxhuvud: Oddly enough, using `odd?`  actually gave noticeably worse performance. I tried to replace the implementation of `odd?` to be this under the hood, but benchmarking that did not give any noticeable difference that I could see. :man_shrugging:

RX14: Just for interest - try adding `AlwaysInline` to the `odd?` implementation. This might just be a LLVM inlining heuristics filure.

yxhuvud: Yeah, no difference. Replacing the method helps `gcd` for everything but `UInt8 ` (and by help, I mean it makes the whole difference if it is categorically faster than the old one or several times slower), but I'm not managing to replicate that performance difference benchmarking `odd?` itself. There is some improvement for signed numbers, but not for unsigned, while the performance difference shows in both. 

All regardless of `@[AlwaysInline]` or not.

src/int.cr
```diff
@@ -357,8 +357,47 @@ struct Int
     (self & mask) == mask
   end
 
-  def gcd(other : Int)
-    self == 0 ? other.abs : (other % self).gcd(self)
+  # Returns the greatest common divisor of `self` and `other` (which
+  # is an `Int` of some kind). The return type is whichever side can
```
asterite: > The return type is whichever side can fit the biggest values.

This can probably be removed.

yxhuvud: How about the new version?

src/int.cr
```diff
@@ -357,8 +357,46 @@ struct Int
...
+  # its type.
+  #
+  # ```
+  # 5.gcd(10) # => 2
+  # 5.gcd(7)  # => 1
+  # ```
+  def gcd(other : self) : self
+    u = self.abs
+    v = other.abs
+    shift = self.class.zero
```
Sija: This line can be moved beneath two `return` lines below.

yxhuvud: :+1:

yxhuvud:
Rebased.

src/int.cr
RX14: This loop is doing `shift = {u.trailing_zeroes, v.trailing_zeroes}.min` right? Perhaps we could bind the `llvm.cttz` intrinsics in the future for more speed :)

yxhuvud: We already have `cttz` bound at `Int#trailing_zeros_count`. However, I tried it and didn't notice any noticable improvement, so I left that out. Which isn't all that surprising, as on average, most numbers don't really have lots of trailing zeros.

But perhaps I did something wrong in my implementation, or it could be affected by me having a quite old desktop computer which I used to test it.. :man_shrugging:

RX14:
Ping.

asterite:
> It's not a hard review.

I think it is. The original `gcd` algorithm was the euclidean, which is very simple to understand. This method is longer than many other methods. I need some time to understand it, or at least... could the code provide a link to the algorithm explanation and proof? I can try it with a bunch of numbers, but where is the proof that this terminates and works correctly?

yxhuvud:
@asterite Ok. I rebased and added a comment with the base behind my implementation.

src/int.cr
```diff
@@ -395,8 +395,49 @@ struct Int
...
   end
 
-  def gcd(other : Int)
-    self == 0 ? other.abs : (other % self).gcd(self)
+  # Returns the greatest common divisor of `self` and `other`. Signed
+  # integers may raise overflow if either has value equal to `MIN` of
+  # its type.
+  #
+  # Implementation heavily inspired by
+  # https://en.wikipedia.org/wiki/Binary_GCD_algorithm#Iterative_version_in_C
```
j8r: I don't think such implementation details should be in the API docs, but it can still be on the source code below.
It doesn't impact how the user will use the method.

asterite: Oh, I missed this. Yeah, for some reason I thought that comment was inside the method definition, not part of its documentation. PRs are welcome, thank you!

PR #3515, opened 2016-11-07
A method to escape a string into uri safe structure

tbrand:
Hi there,
There is a critical difference between Ruby and Crystal for an implementation of URI.escape.

```
require "uri"
p URI.escape("https://crystal-lang.org/")
```
Then,
```
$ ruby sample.cr
-> "https://crystal-lang.org/"
$ crystal sample.cr
-> "https%3A%2F%2Fcrystal-lang.org%2F"
```
So it is difficult to escape a String in uri-safe structure like as Ruby's behavior.
So I added a method to keep the String url structure.
```
require "uri"
p URI.escape_uri_safe(":/?#[]@!$&\'()*+,;= ")
p URI.escape_uri_safe("https://crystal-lang.org?test=あああ")
```
Then,
```
$ crystal sample.cr
-> ":/?#[]@!$&'()*+,;=%20"
   "https://crystal-lang.org?test=%E3%81%82%E3%81%82%E3%81%82"
```

May be [this](https://github.com/crystal-lang/crystal/issues/2098) issue is involved.

Thanks for your review.

PR #7997, opened 2019-07-26
Improve URL encoding

straight-shoota:
* Adds methods URI.encode and URI.decode for URL encoding without
  escaping reserved characters. These are to be used when encoding URLs.
* URI.escape and URI.unescape are renamed to URI.encode_www_form and
URI.decode_www_form. These are to be used for `x-www-form-urlencoded` serialization.

The specific implementations are explained in the documentation.
Method names have been chosen according to the names of other APIs, especially Elixir's, per https://github.com/crystal-lang/crystal/pull/3515#issuecomment-335878803 and https://github.com/crystal-lang/crystal/pull/3515#issuecomment-509262082

All these methods have been moved to a separate file in `src/uri/encoding.cr` to clean up the main file. They're an isolated domain just live in URI's class namespace.

There were two variants of escape/unescape receiving a block, which allow for custom modifications of the encoding rules. I'm not sure whether we need to expose them at all in the public API. Use cases should be pretty rare.
So maybe, we should `nodoc` them. I only kept the variant writing to IO. Such low-level methods to not need to provide a string variant. This at least reduces the impact on the public API a bit.

Closes #3515

### DIFF
--- a/spec/std/uri_spec.cr
+++ b/spec/std/uri_spec.cr
@@ -8,6 +8,46 @@ private def assert_uri(string, file = __FILE__, line = __LINE__, **args)
   end
 end
 
+private def it_encodes(string, expected_result, file = __FILE__, line = __LINE__, **options)
+  it "encodes #{string.inspect}", file, line do
+    URI.encode(string, **options).should eq(expected_result), file, line
+
+    String.build do |io|
+      URI.encode(string, io, **options)
+    end.should eq(expected_result), file, line
+  end
+end
+
+private def it_decodes(string, expected_result, file = __FILE__, line = __LINE__, **options)
+  it "decodes #{string.inspect}", file, line do
+    URI.decode(string, **options).should eq(expected_result), file, line
+
+    String.build do |io|
+      URI.decode(string, io, **options)
+    end.should eq(expected_result), file, line
+  end
+end
+
+private def it_encodes_www_form(string, expected_result, file = __FILE__, line = __LINE__, **options)
+  it "encodes #{string.inspect}", file, line do
+    URI.encode_www_form(string, **options).should eq(expected_result), file, line
+
+    String.build do |io|
+      URI.encode_www_form(string, io, **options)
+    end.should eq(expected_result), file, line
+  end
+end
+
+private def it_decodes_www_form(string, expected_result, file = __FILE__, line = __LINE__, **options)
+  it "decodes #{string.inspect}", file, line do
+    URI.decode_www_form(string, **options).should eq(expected_result), file, line
+
+    String.build do |io|
+      URI.decode_www_form(string, io, **options)
+    end.should eq(expected_result), file, line
+  end
+end
+
 describe "URI" do
   describe ".parse" do
     # scheme
@@ -224,6 +264,7 @@ describe "URI" do
     it { URI.parse("http://www.example.com").userinfo.should be_nil }
     it { URI.parse("http://foo@www.example.com").userinfo.should eq("foo") }
     it { URI.parse("http://foo:bar@www.example.com").userinfo.should eq("foo:bar") }
+    it { URI.new(user: "ä /", password: "ö :").userinfo.should eq("%C3%A4+%2F:%C3%B6+%3A") }
   end
 
   describe "to_s" do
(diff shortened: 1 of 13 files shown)


src/uri.cr
```diff
@@ -19,6 +20,31 @@ require "./uri/uri_parser"
...
+# * `.encode(string : String, *, space_to_plus : Bool = false) : String`: URL-encodes a string.
+# * `.encode(string : String, io : IO, *, space_to_plus : Bool = false) : Nil`: URL-encodes a string to an IO.
+# * `.decode_www_form(string : String, *, plus_to_space : Bool = true) : String`: Decodes an `x-www-form-urlencoded` string component.
+# * `.decode_www_form(string : String, io : IO, *, plus_to_space : Bool = true) : Nil`: Decodes an `x-www-form-urlencoded` string component to an IO.
+# * `.encode_www_form(string : String, *, space_to_plus : Bool = true) : String`: Encodes a string as a `x-www-form-urlencoded` component.
+# * `.encode_www_form(string : String, io : IO, *, space_to_plus : Bool = true) : Nil`: Encodes a string as a `x-www-form-urlencoded` component to an IO.
+#
+# The main difference is that `.encode_www_form` encodes reserved characters
+# (see `.reserved?`), while `.encode` does not. The decode methods are exactly
+# similar except for the handling of `+` characters.
```
Sija: _exactly similar_? 🤔 ;)

straight-shoota: Yes, that's the only difference. `decode` decodes `+` literally while `decode_www_form` decodes it as space.

Sija: I'd rephrase it into sth like `The decode methods are almost identical, except for...`.

straight-shoota: Okay, I get what you mean. I'm not a native speaker but AFAIK "exactly similar" is a common phrase to describe something that is almost identical. But, sure, I can rephrase that.

spec/std/uri_spec.cr
```diff
@@ -305,112 +346,96 @@ describe "URI" do
...
+    it_decodes_www_form("%", "%")
+    it_decodes_www_form("%1", "%1")
+    it_decodes_www_form("123%45%6", "123E%6")
+    it_decodes_www_form("%zzzzz", "%zzzzz")
   end
 
-  describe "reserved?" do
-    reserved_chars = Set.new([':', '/', '?', '#', '[', ']', '@', '!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '='])
+  it ".reserved?" do
+    reserved_chars = {':', '/', '?', '#', '[', ']', '@', '!', '$', '&', '\'', '(', ')', '*', '+', ',', ';', '='}
```
asterite: I don't think there's a need to use a tuple here.

asterite: That is, each different tuple type means a new type the compiler has to track, with new method instantiations and so on. So if the idea of using a tuple is for performance, this actually had the opposite result. And in specs the performance of using a tuple vs a set doesn't matter (and I think a Set is more efficient for lookup).

src/http/cookie.cr
```diff
@@ -50,7 +50,7 @@ module HTTP
     end
 
     def to_cookie_header
-      "#{@name}=#{URI.escape value}"
+      "#{URI.encode_www_form(@name)}=#{URI.encode_www_form(value)}"
```
asterite: optional optimization: stream to the IO instead of creating many intermediate strings.

src/uri.cr
asterite: I don't think this matters much.

Sija: @asterite yes it does, this wording is confusing and docs are meant to explain things and not obscure, don't they?

straight-shoota: I don't think it's confusing, but the alternative is also good, so I'll change it.

straight-shoota:
@RX14 This branch could've used a rebase before merging. Now we have this fixup in master: https://github.com/crystal-lang/crystal/commit/2c416baf35d5cfc64b212634be84231acd4e2e0e

RX14:
Whoops, I always forget that github remembers the "last" merge mode used.

asterite:
I personally don't think the commit history is that important.

straight-shoota:
It's by far not critical, but keeping the history clean is somewhat important, so that the changes are comprehensible. If we had lot's of isolated fixup commits, it could be a nightmare to figure out how some change came to be.

bcardiff:
I think I've found an issue with this PR. Reported as #8020

oprypin:
I don't think the motivation behind this change is right.

Consider
Python:
```python
urllib.parse.quote("item #9") == "item%20%239"
```
Crystal:
```crystal
URI.escape("item #9") == "item%20%239"
```
which is exactly what people usually want, but **it's gone now**.

Now we have:
```crystal
URI.encode("item #9") == "item%20#9"
URI.encode_www_form("item #9") == "item+%239"

```

And, while `encode_www_form` is fine -- for WWW forms, yes,
what exactly is the usecase for `URI.encode`?
I see the examples showing that you are supposed to pass the whole URI to it. I think that's almost always dubious to do. Could you provide a real example when one would want to convert a URI using the current `URI.encode`? I can't.

Anyway,
If I'm making a variable part in my URL,
such as `/items/#{item}`,
I want to form the URL by writing
`"/items/#{URI.escape(item)}"` (but it's gone now).

Leaving the `#` in there is a mistake (i.e. using `URI.encode` is a mistake - yes, see https://github.com/oprypin/nightly.link/issues/15).
And don't get me started on `encode_www_form`. I'm **not encoding a form**. I don't want to explicitly have to disable the `+` replacement (which is also a mistake here).

So to recap, these are the available options for the **typical usecase** now:

```crystal
URI.encode_www_form("item #9", space_to_plus: false) == "item%20#9"
String.build { |io| URI.encode("item #9", io, &->URI.unreserved?(UInt8)) } == "item%20#9"
```

And honestly, I prefer the latter, because, remember, **I'm not encoding a WWW form**.
I guess in my projects I'll be adding a helper function doing this action.`

oprypin:
I have also looked at most usages of this method and concluded that most of them are wrong.

https://pryp.in/p/906ce163818aa867a9d876739aa8f031217cab5e/results.html

Note that this is a page **with my own remarks added**

matthewmcgarvey:
Just to chime in a bit, you can see problems with this change that people (including myself) had through this issue and the connected PRs https://github.com/taylorfinnell/awscr-s3/issues/86

straight-shoota:
The original discussion of the scope of these methods is in #3515. This is just the PR that implemented the change. I don't like reviving discussion zombies. Can you please open a new issue about this, ideally incorporating insights from the original discussion?

oprypin:
Done, #10603

PR #7998, opened 2019-07-26
Fiber#wakeup to wake up sleeping fibers

carlhoerberg:
Allow waking up sleeping fibers. I know this is a controversial topic, "fibers shouldn't have a public API" etc., but `Channel` doesn't have timeout mechanism and in our app we can over time accumulate hundreds of thousands of sleeping fibers that eventually will wake up and do nothing, so we've monkey patched our app with this functionality. 

Very simplified we got something like this now:
```crystal
ch = Channel(Int32).new
loop do
  waiting = true
  f = spawn do
    sleep 60
    ch.close if waiting
  end
  v = ch.receive? || break
  waiting = false
  f.wakeup
  work(v)
end
``` 

Of course a "native" timeout argument to `Channel#receive` and `Channel#send` would be even more elegant (and one that don't necessarily closes the channel, but raises a `Channel::Timeout` exception maybe).

### DIFF
--- a/spec/std/fiber_spec.cr
+++ b/spec/std/fiber_spec.cr
@@ -1,0 +1,16 @@
+require "spec"
+
+describe Fiber do
+  it "can be woken up from sleep" do
+    took = Time.measure do
+      ch = Channel(Nil).new
+      f = spawn do
+        sleep 2
+        ch.send nil
+      end
+      f.wakeup
+      ch.receive
+    end
+    took.should be < 1.seconds
+  end
+end
--- a/src/crystal/event.cr
+++ b/src/crystal/event.cr
@@ -33,6 +33,12 @@ struct Crystal::Event
     @freed = true
   end
 
+  def delete
+    unless LibEvent2.event_del(@event) == 0
+      raise "Error deleting event"
+    end
+  end
+
   # :nodoc:
   struct Base
     def initialize
--- a/src/fiber.cr
+++ b/src/fiber.cr
@@ -119,6 +119,15 @@ class Fiber
     Crystal::Scheduler.resume(self)
   end
 
+  # Wakeup a sleeping fiber
+  def wakeup
+    raise "Can't wakeup dead fibers" if dead?
+    raise "Can't wakeup one self" if self == Fiber.current
+    @resume_event.try &.delete
+    Crystal::Scheduler.enqueue(Fiber.current)
+    Crystal::Scheduler.resume(self)
+  end
+
   # :nodoc:
   def resume_event
     @resume_event ||= Crystal::EventLoop.create_resume_event(self)


bararchy:
I get why you want this, and TBH this is the pains you see in Ruby with green-threads, because we lack parallelism (which is "coming" (c) - trademark by manas)  I think this and other ways to "make it work" won't be really needed once multi-threading lands (soon, really, in 1-100 years)

asterite:
I would abstain from modifying anything related to concurrency.

The core team is working on this. Eventually we'll have parallelism, and once we have that we'll start having more primitives and methods for timeouts, scheduling, etc.

Anything touching this will most likely conflict with the existing work.

straight-shoota:
It doesn't really add anything new. Calling `Crystal::Scheduler.enqueue(Fiber.current)` and `    Crystal::Scheduler.resume(self)` is already available anyway. Essentially, it's just a delayed version of `Fiber#resume`.

Deleting the resume event should probably happen in `Crystal::Scheduler.enqueue` in any case.

asterite:
> Channel doesn't have timeout mechanism

The idea would be to use `select` and have something like this:

```crystal
chan = Channel(Int32).new

select
when x = chan.receive
  puts "Got #{x}"
when sleep(3)
  puts "Timeout!"
end
```

but we never got to implement `sleep_select_action` (which is what `sleep` turns into inside a `select`).

But then you can do it like this:

```crystal
chan = Channel(Int32).new
sleep_chan = Channel(Nil).new

spawn do
  sleep(3)
  sleep_chan.send(nil)
end

select
when x = chan.receive
  puts "Got #{x}"
when sleep_chan.receive
  puts "Timeout!"
end
```

And then you can have something like this:

```crystal
def receive_with_timeout(chan, timeout)
  sleep_chan = Channel(Nil).new
  spawn do
    sleep(timeout)
    sleep_chan.send(nil)
  end

  select
  when x = chan.receive
    x
  when sleep_chan.receive
    raise "Timeout!"
  end
end

chan = Channel(Int32).new

x = receive_with_timeout(chan, 3.seconds)
puts x
```

(or with a block, different exception, etc.).

asterite:
This seems to work too:

```crystal
def sleep_select_action(timeout)
  chan = Channel(Nil).new
  spawn do
    sleep(timeout)
    chan.send(nil)
  end
  Channel::ReceiveAction.new(chan)
end

chan = Channel(Int32).new

select
when x = chan.receive
  puts "Got #{x}"
when sleep(3)
  puts "Timeout!"
end
```

However, I think @waj's idea was to avoid having to create a Channel just for a timeout.

asterite:
(but with could go with that implementation for now and eventually optimize it)

carlhoerberg:
@straight-shoota  I'm not sure deleting the event in `Crystal::Scheduler.resume` is smart, the "sleep event" it self is calling `Crystal::Scheduler.resume` when fired: https://github.com/crystal-lang/crystal/blob/0.29.0/src/crystal/event_loop.cr#L21
So it would in most cases just try to delete something that isn't there.

@asterite Sure, that's a nice way to solve the same problem with the existing API, but if the sleep is say an hour or even days (real use case in our code), won't there still be an massive amount of fibers hanging around eventually?

asterite:
@carlhoerberg But if you want to eventually kill it, instead of calling `wakeup` you can just send a message over a channel and have that fiber consume it... I think?

carlhoerberg:
no, it's sleeping, not receiving on a channel, so how would that work? 

Note that this is not like killing or canceling a fiber with all gory details that entails, this just cancels the event and resumes the fiber gracefully (https://github.com/crystal-lang/crystal/pull/6450), it's in no way "dangerous" or changes API/behavior, just adds a capability, canceling sleeps.

asterite:
I mean, you have this:

```crystal
ch = Channel(Int32).new
loop do
  waiting = true
  f = spawn do
    sleep 60
    ch.close if waiting
  end
  v = ch.receive? || break
  waiting = false
  f.wakeup
  work(v)
end
```

You can do it without wakeup by using a separate channel:

```crystal
def sleep_select_action(timeout)
  chan = Channel(Nil).new
  spawn do
    sleep(timeout)
    chan.send(nil)
  end
  Channel::ReceiveAction.new(chan)
end

ch = Channel(Int32).new
loop do
  waiting = true
  wakeup = Channel(Nil).new
  f = spawn do
    select
    when sleep 60
      # Timeout...
    when wakeup.receive
      # Wakeup
    end
    ch.close if waiting
  end
  v = ch.receive? || break
  waiting = false
  wakeup.send(nil)
  work(v)
end
```

I'm just saying that:
1. This seems to be already doable without changing anything
2. There's probably a good reason there's no "wakeup" in Go

carlhoerberg:
No, now you got two fibers for no reason. The "sleeping" fiber in `sleep_select_action` will still be hanging around even after `wakeup.send(nil)`. And if the timeout is hours or days and you do this very often you will create a memory leak. I'm pretty sure go cancel its timeout timer when the channel has received something, your example doesn't.

asterite:
Yes, the idea was to find a way to do the timeout without `spawn` and `Channel`.

asterite:
Actually, maybe this can work. But we'll need input from someone else from the core team that knows more about concurrency, and whether this is safe/intuitive.

asterite:
Another thing: when we have multiple threads it would be incorrect to wake up a fiber from another thread... meaning it will crash badly.

I think we'll probably also make `spawn` return `Nil` so you can't hold a reference to a fiber.

Also, why is the fiber doing `sleep`? I'm not sure I understand.

carlhoerberg:
If you can implement a channel timeout without sleep (and memory leaks) I'm all ears, I guess with some custom libevent event just for channel timeouts

carlhoerberg:
Have anyone tried implementing `Channel` with libevent user events (`event_action`)? Or should I give it a try? Much like `IO::Evented` but manually activating the read/write event from the sender/receiver.

asterite:
As I said, I don't think it's worth spending time on these things. When parallelism lands it won't share a single bit of similarity with the code that exists now.

PR #7999, opened 2019-07-26
Turn abstract return type erros into warnings

asterite:
Follow up to #7956

We agreed that making them errors will cause a massive breaking change.
Making them warnings first will make fixing these things easier while things still keep working.

### DIFF
--- a/spec/compiler/semantic/abstract_def_spec.cr
+++ b/spec/compiler/semantic/abstract_def_spec.cr
@@ -395,8 +395,8 @@ describe "Semantic: abstract def" do
       )
   end
 
-  it "errors if missing return type" do
-    assert_error %(
+  it "warning if missing return type" do
+    assert_warning %(
       abstract class Foo
         abstract def foo : Int32
       end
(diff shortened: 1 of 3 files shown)


bcardiff:
After this PR these warnings are always emitted, despite `--warnings=none`.
In other places, before appending something to `@program.warning_failures` a check is done after: 

* checking `@program.warnings.all?` (or eventually some other flag configuration per type of warning)
* checking location of the code via `ignore_warning_due_to_location` or similar.

Without these two checks, the compiler is a bit more pedantic.

I'm checking if we are satisfied with that current state before releasing

asterite:
Oh, I didn't see that... I though you would emit warnings and in that method they would be ignored if they are not turned on. This is probably slightly slower because the warning message is created, but I'm not sure how much of a slowdown that is...

bcardiff:
If we would go that route the warning message would need to have a category field and location so that logic is handled in a single place.

Or we can change the api to be `.warning(category, location) { "message" }` to lazy build the message if category / location matches. But at least in the deprecation message warning, it was better to have logic upper in a separate place for performance (and that was the only case up to now)

asterite:
I see... I guess for now we can just fix the existing warnings that are always shown then. Could you take care of that? All of them are in `abstract_def_checker.cr`. Otherwise let me know and I'll do it, but much later today.

RX14:
The warning system is only used here for 1 release, and warnings (iirc) are still off by default because I was removing the warning system in favour of a simpler system. I was working on this but I'm still struggling to even keep up with github notifications with the time I have to work on crystal...

bcardiff:
I wasn’t able to spent time on this. I might tomorrow if I finish checking the ecosystem. 

I see this as a first time where we can evolve smoothly the language semantics. I find it valuable to build the abstractions to implement these kind of things in the less error prone way as possible.

asterite:
@bcardiff No worries, I already sent a PR for this :-) (I hope it's fine 🤞)

PR #8002, opened 2019-07-27
Http optimizations

asterite:
These are a series of refactors to improve parsing of HTTP requests. Check each commit for more details (each has a description in it).

This is more code, more complex code, and more low-level code, but that's the whole idea of Crystal: if you want to go deeper and optimize hot paths you can do it in Crystal itself.

I wrote this benchmark:

```crystal
require "benchmark"
require "http"

# # request.txt was generated using this code
# request = <<-REQUEST.lines.join("\r\n") + "\r\n\r\n"
# GET /hello.htm HTTP/1.1
# User-Agent: Mozilla/4.0 (compatible; MSIE5.01; Windows NT)
# Host: www.tutorialspoint.com
# Accept-Language: en-us
# Accept-Encoding: gzip, deflate
# Connection: Keep-Alive
# REQUEST
# File.write("request.txt", request)

io = File.open("request.txt")

Benchmark.ips do |x|
  x.report("from_io") do
    io.rewind
    HTTP::Request.from_io(io)
  end
end
```

I read from a file to simulate reading from an external resource, though the entire file probably fits `IO::Buffered`'s buffer (but the same should be true for `Socket`). But later I'll show the results with `IO::Memory` too.

If I run the above benchmark against `master`:

```
$ crystal foo.cr --release
from_io 335.90k (  2.98µs) (± 1.07%)  2.1kB/op  fastest
```

When against this PR:

```
$ bin/crystal foo.cr --release
from_io 495.93k (  2.02µs) (± 0.85%)  816B/op  fastest
```

So a 30% improvement! 😄 

Also note the memory allocated per op: 2.1kB before, now 816B. This is the main reason it's faster.

If I use `IO::Memory` instead of `File` I get:

```
Before: from_io 509.34k (  1.96µs) (± 0.75%)  2.1kB/op  fastest
After:  from_io   1.01M (991.82ns) (± 0.75%)  816B/op  fastest
```

Since HTTP servers are pretty common in Crystal I thought this is a good way to optimize all apps out there. I know usually a lot more goes on in a typical web server (for example rendering) but the less time and memory the framework takes for itself, the better.

I didn't benchmark an HTTP::Server with `ab` or similar after this change, but you are welcome to do that and post the results here!

### DIFF
--- a/spec/std/http/headers_spec.cr
+++ b/spec/std/http/headers_spec.cr
@@ -160,6 +160,21 @@ describe HTTP::Headers do
     headers.includes_word?("foo", "baz").should be_true
   end
 
+  it "doesn't match word with comma separated value, partial match" do
+    headers = HTTP::Headers{"foo" => "bar, bazo"}
+    headers.includes_word?("foo", "baz").should be_false
+  end
+
+  it "matches word with comma separated value, partial match (array)" do
+    headers = HTTP::Headers{"foo" => ["foo", "baz, bazo"]}
+    headers.includes_word?("foo", "baz").should be_true
+  end
+
+  it "doesn't match word with comma separated value, partial match (array)" do
+    headers = HTTP::Headers{"foo" => ["foo", "bar, bazo"]}
+    headers.includes_word?("foo", "baz").should be_false
+  end
+
   it "matches word among headers" do
     headers = HTTP::Headers.new
     headers.add("foo", "bar")
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -18,15 +18,18 @@ module HTTP
   SUPPORTED_VERSIONS = {"HTTP/1.0", "HTTP/1.1"}
 
   # :nodoc:
+  record EndOfRequest
+  # :nodoc:
+  record HeaderLine, name : String, value : String, bytesize : Int32
+
+  # :nodoc:
   def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true)
     headers = Headers.new
 
     headers_size = 0
-    while line = io.gets(MAX_HEADER_SIZE, chomp: true)
-      headers_size += line.bytesize
-      break if headers_size > MAX_HEADER_SIZE
-
-      if line.empty?
+    while header_line = read_header_line(io)
+      case header_line
+      when EndOfRequest
         body = nil
 
         if body_type.prohibited?
@@ -64,11 +67,45 @@ module HTTP
 
         yield headers, body
         break
-      end
+      else # HeaderLine
+        headers_size += header_line.bytesize
+        break if headers_size > MAX_HEADER_SIZE
 
-      name, value = parse_header(line)
-      break unless headers.add?(name, value)
+        break unless headers.add?(header_line.name, header_line.value)
+      end
     end
+  end
+
+  private def self.read_header_line(io) : HeaderLine | EndOfRequest | Nil
+    # Optimization: check if we have a peek buffer
+    if peek = io.peek
+      # peek.empty? means EOF (so bad request)
+      return nil if peek.empty?
+
+      # Find \r\n (first \n, then \r before it)
+      index = peek.index('\n'.ord.to_u8)
+      if index && index > 0 && peek[index - 1] == '\r'.ord.to_u8
+        # `index == 1` means we just have "\r\n", so end of request
+        if index == 1
+          io.skip(2)
+          return EndOfRequest.new
+        end
+
+        name, value = parse_header(peek[0, index - 1])
+        io.skip(index + 1) # Must skip until after \r\n
+        return HeaderLine.new name: name, value: value, bytesize: index + 1
+      end
+    end
+
+    line = io.gets(MAX_HEADER_SIZE, chomp: true)
+    return nil unless line
+
+    if line.empty?
+      return EndOfRequest.new
+    end
+
+    name, value = parse_header(line)
+    return HeaderLine.new name: name, value: value, bytesize: line.bytesize
   end
 
   private def self.check_content_type_charset(body, headers)
@@ -87,7 +124,12 @@ module HTTP
   end
 
   # :nodoc:
-  def self.parse_header(line)
+  def self.parse_header(line : String) : {String, String}
+    parse_header(line.to_slice)
+  end
+
+  # :nodoc:
+  def self.parse_header(slice : Bytes) : {String, String}
     # This is basically
     #
     # ```
@@ -100,12 +142,12 @@ module HTTP
     # instead of 3 (two from the split and one for the lstrip),
     # and there's no need for the array returned by split.
 
-    cstr = line.to_unsafe
-    bytesize = line.bytesize
+    cstr = slice.to_unsafe
+    bytesize = slice.size
 
     # Get the colon index and name
-    colon_index = cstr.to_slice(bytesize).index(':'.ord) || 0
-    name = line.byte_slice(0, colon_index)
+    colon_index = slice.index(':'.ord.to_u8) || 0
+    name = header_name(slice[0, colon_index])
 
     # Get where the header value starts (skip space)
     middle_index = colon_index + 1
@@ -117,15 +159,54 @@ module HTTP
     right_index = bytesize
     if middle_index >= right_index
       return {name, ""}
-    elsif right_index > 1 && cstr[right_index - 2] === '\r' && cstr[right_index - 1] === '\n'
+    elsif right_index > 1 && cstr[right_index - 2] == '\r'.ord.to_u8 && cstr[right_index - 1] == '\n'.ord.to_u8
       right_index -= 2
-    elsif right_index > 0 && cstr[right_index - 1] === '\n'
+    elsif right_index > 0 && cstr[right_index - 1] == '\n'.ord.to_u8
       right_index -= 1
     end
 
-    value = line.byte_slice(middle_index, right_index - middle_index)
+    value = String.new(slice[middle_index, right_index - middle_index])
 
     {name, value}
+  end
+
+  private COMMON_HEADERS = %w(
+    accept-language
+    accept-encoding
+    allow
+    cache-control
+    connection
+    content-disposition
+    content-encoding
+    content-language
+    content-length
+    content-type
+    etag
+    expires
+    host
+    last-modified
+    user-agent
+  )
+
+  # :nodoc:
+  def self.header_name(slice : Bytes)
+    # Check if the header name is a common one.
+    # If so we avoid having to allocate a string for it.
+    if slice.size < 20
+      buffer = uninitialized UInt8[20]
+
+      # Copy the slice to buffer, downcased
+      slice.each_with_index do |byte, i|
+        buffer[i] = 65 <= byte <= 90 ? byte + 32 : byte
+      end
+
+      buffer_slice = buffer.to_slice[0, slice.size]
+
+      name = COMMON_HEADERS.find { |value| value.to_slice == buffer_slice }
+      return name if name
+    end
+
+    String.new(slice)
   end
 
   # :nodoc:
--- a/src/http/headers.cr
+++ b/src/http/headers.cr
@@ -49,11 +49,16 @@ struct HTTP::Headers
   end
 
   def initialize
-    @hash = Hash(Key, Array(String)).new
+    # We keep a Hash with String | Array(String) values because
+    # the most common case is a single value and so we avoid allocating
+    # memory for arrays.
+    @hash = Hash(Key, String | Array(String)).new
   end
 
   def []=(key, value : String)
-    self[wrap(key)] = [value]
+    check_invalid_header_content(value)
+
+    @hash[wrap(key)] = value
   end
 
   def []=(key, value : Array(String))
@@ -83,21 +88,31 @@ struct HTTP::Headers
   def includes_word?(key, word)
     return false if word.empty?
 
-    word = word.downcase
-    # iterates over all header values avoiding the concatenation
-    get?(key).try &.each do |value|
-      value = value.downcase
-      offset = 0
-      while true
-        start = value.index(word, offset)
-        break unless start
-        offset = start + word.size
-
-        # check if the match is not surrounded by alphanumeric chars
-        next if start > 0 && value[start - 1].ascii_alphanumeric?
-        next if start + word.size < value.size && value[start + word.size].ascii_alphanumeric?
-        return true
+    values = @hash[wrap(key)]?
+    case values
+    when Nil
+      false
+    when String
+      includes_word_in_header_value?(word.downcase, values.downcase)
+    else
+      word = word.downcase
+      values.any? do |value|
+        includes_word_in_header_value?(word, value.downcase)
       end
+    end
+  end
+
+  private def includes_word_in_header_value?(word, value)
+    offset = 0
+    while true
+      start = value.index(word, offset)
+      return false unless start
+      offset = start + word.size
+
+      # check if the match is not surrounded by alphanumeric chars
+      next if start > 0 && value[start - 1].ascii_alphanumeric?
+      next if start + word.size < value.size && value[start + word.size].ascii_alphanumeric?
+      return true
     end
 
     false
@@ -165,15 +180,15 @@ struct HTTP::Headers
 
     other.each do |key, value|
       this_value = @hash[wrap(key)]?
-      if this_value
-        case value
-        when String
-          return false unless this_value.size == 1 && this_value[0] == value
-        when Array(String)
-          return false unless this_value == value
-        else
-          false
-        end
+      case {value, this_value}
+      when {String, String}
+        return false unless value == this_value
+      when {Array, Array}
+        return false unless value == this_value
+      when {String, Array}
+        return false unless this_value.size == 1 && this_value[0] == value
+      when {Array, String}
+        return false unless value.size == 1 && value[0] == this_value
       else
         return false unless value.nil?
       end
@@ -184,16 +199,16 @@ struct HTTP::Headers
 
   def each
     @hash.each do |key, value|
-      yield({key.name, value})
+      yield({key.name, cast(value)})
     end
   end
 
   def get(key)
-    @hash[wrap(key)]
+    cast @hash[wrap(key)]
   end
 
   def get?(key)
-    @hash[wrap(key)]?
+    @hash[wrap(key)]?.try { |value| cast(value) }
   end
 
   def dup
@@ -218,8 +233,13 @@ struct HTTP::Headers
       io << ", " if index > 0
       key.name.inspect(io)
       io << " => "
-      if values.size == 1
-        values.first.inspect(io)
+      case values
+      when Array
+        if values.size == 1
+          values.first.inspect(io)
+        else
+          values.inspect(io)
+        end
       else
         values.inspect(io)
       end
@@ -259,9 +279,13 @@ struct HTTP::Headers
     key = wrap(key)
     existing = @hash[key]?
     if existing
-      existing << value
+      if existing.is_a?(Array)
+        existing << value
+      else
+        @hash[key] = [existing, value]
+      end
     else
-      @hash[key] = [value]
+      @hash[key] = value
     end
   end
 
@@ -269,7 +293,13 @@ struct HTTP::Headers
     key = wrap(key)
     existing = @hash[key]?
     if existing
-      existing.concat value
+      if existing.is_a?(Array)
+        existing.concat value
+      else
+        new_value = [existing]
+        new_value.concat(value)
+        @hash[key] = new_value
+      end
     else
       @hash[key] = value
     end
@@ -287,7 +317,11 @@ struct HTTP::Headers
     value
   end
 
-  private def concat(values)
+  private def concat(values : String)
+    values
+  end
+
+  private def concat(values : Array(String))
     case values.size
     when 0
       ""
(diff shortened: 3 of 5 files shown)


src/http/headers.cr
```diff
@@ -83,24 +86,31 @@ struct HTTP::Headers
...
+        value = value.downcase
+        offset = 0
+        while true
+          start = value.index(word, offset)
+          break unless start
+          offset = start + word.size
+
+          # check if the match is not surrounded by alphanumeric chars
+          next if start > 0 && value[start - 1].ascii_alphanumeric?
+          next if start + word.size < value.size && value[start + word.size].ascii_alphanumeric?
```
straight-shoota: Shouldn't the surroundings be checked in the single-value branch as well?

src/http/request.cr
```diff
@@ -93,32 +93,119 @@ class HTTP::Request
...
+    # Optimization: avoid allocating a string for common HTTP version
+    http_version = HTTP::SUPPORTED_VERSIONS.find { |version| version.to_slice == subslice }
+    return BadRequest.new unless http_version
 
-      return request
+    # Skip trailing spaces
+    space_index = next_space_index
+    while space_index < slice.size
+      # Oops, we find something else (more than three parts)
+      return BadRequest.new unless slice[space_index] == ' '.ord.to_u8
```
straight-shoota: Half the character comparisons use `===` the other half `==` + `ord.to_u8`. I'd stick with either one of them, but keep it homogenous.

src/http/request.cr
```diff
@@ -93,32 +93,119 @@ class HTTP::Request
...
+    subslice = slice[space_index...next_space_index]
+
+    # Optimization: avoid allocating a string for common HTTP version
+    http_version = HTTP::SUPPORTED_VERSIONS.find { |version| version.to_slice == subslice }
+    return BadRequest.new unless http_version
 
-      return request
+    # Skip trailing spaces
+    space_index = next_space_index
+    while space_index < slice.size
```
straight-shoota: This loop is missing an index increment.

src/http/request.cr
```diff
@@ -93,32 +93,119 @@ class HTTP::Request
...
   # `nil` on EOF or `BadRequest` otherwise.
-  def self.from_io(io)
+  def self.from_io(io) : HTTP::Request | BadRequest | Nil
+    line = parse_request_line(io)
+
+    case line
+    when EOF
+      nil
+    when BadRequest
+      BadRequest.new
```
straight-shoota: Why not simply return `line`? This entire `case` could even be replaced by `return line unless line.is_a?(RequestLine)` when `EOF` would just be `nil` instead of an extra type. Having lesser types is better and I don't see no reason why this couldn't be `nil`.

src/http/headers.cr
asterite: I think you are right. I thought multiple values in http headers were separates by commas, but it can be a single value with commas in it. I'll fix it.

asterite: Fixed, thanks! The specs were passing because they create headers with the special hash syntax, which uses `[]=`, and apparently that was always turning the value into an array... now that's no longer the case which would make working with `HTTP::Headers` consume less memory in general (parsing of `HTTP::Request` uses `HTTP::Headers#add?` which doesn't create an array if needed, so this doesn't affect the benchmarks I showed before).

straight-shoota: Yes. This method doesn't look for values but words.

src/http/request.cr
asterite: Thank you! ❤️ 

Fixed and added a spec to prove that it works fine (just a request line with many spaces all around, except before the first word).

src/http/request.cr
asterite: Yes, this was initially `nil` but I thought it was clearer with `EOF` because then I don't have to document that `nil` means `EOF`. I'll try it and see how it looks. After all it's just a small internal method.

asterite: Fixed!

src/http/request.cr
asterite: Good idea. I left them all like `==` + `ord.to_u8` for consistency because when calling `index` and others you have to pass an `UInt8`.

spec/std/http/headers_spec.cr
```diff
@@ -160,6 +160,21 @@ describe HTTP::Headers do
     headers.includes_word?("foo", "baz").should be_true
   end
 
+  it "doesn't match word with comma separated value, partial match" do
+    headers = HTTP::Headers{"foo" => "bar, bazo"}
+    headers.includes_word?("foo", "baz").should be_false
+  end
+
+  it "matches word with comma separated value, partial match (array)" do
+    headers = HTTP::Headers{"foo" => ["bar", "bazo, baz"]}
```
straight-shoota: These examples values are pretty confusing, using the same order is easier to read:
```suggestion
    headers = HTTP::Headers{"foo" => ["foo", "baz, bazo"]}
```

src/http/request.cr
```diff
@@ -121,6 +121,91 @@ class HTTP::Request
     BadRequest.new
   end
 
+  # :nodoc:
+  METHODS = %w(GET HEAD POST PUT DELETE CONNECT OPTIONS PATCH)
```
j8r: I think they can be public, HTTP request methods are used in macros in web frameworks like [Kemal](https://github.com/kemalcr/kemal/blob/v0.25.2/src/kemal/dsl.cr#L9) and [Amber](https://github.com/amberframework/amber/blob/v0.29.0/src/amber/dsl/router.cr), and in [HTTP::Client](https://github.com/crystal-lang/crystal/blob/0.29.0/src/http/client.cr#L370).

The `TRACE` method is missing.
The RFCs can also be put in the comment description:
`PATCH`:  https://tools.ietf.org/html/rfc5789#section-2
other methods: https://tools.ietf.org/html/rfc7231#section-4.

Can be changed to a Tuple, or maybe an enum.

src/http/common.cr
```diff
@@ -117,17 +160,57 @@ module HTTP
...
     end
 
-    value = line.byte_slice(middle_index, right_index - middle_index)
+    value = String.new(slice[middle_index, right_index - middle_index])
 
     {name, value}
   end
 
+  # :nodoc:
+  COMMON_HEADERS = %w(
```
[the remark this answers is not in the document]
asterite: No. A tuple creates a new type for the compiler for no reason. Tuples main use case is returning multiple values from a method. All other uses are a smell.

src/http/common.cr
```diff
@@ -64,11 +67,46 @@ module HTTP
...
+
+        break unless headers.add?(header_line.name, header_line.value)
+      end
+    end
+  end
+
+  private def self.read_header_line(io) : HeaderLine | EndOfRequest | Nil
+    # Optimization: check if we have a peek buffer
+    peek = io.peek
+    if peek
```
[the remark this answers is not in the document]
asterite: Done, thanks! (not really needed, but...)

src/http/request.cr
asterite: Thanks, I added `TRACE`.

There's no need to cover all cases here, if a method is missing it will still work though a string will be allocated for that method.

src/http/request.cr
```diff
@@ -121,6 +121,91 @@ class HTTP::Request
...
+
+  private def self.parse_request_line(io : IO) : RequestLine | BadRequest | Nil
+    # Optimization: see if we have a peek buffer
+    # (avoids a string allocation for the entire request line)
+    peek = io.peek
+    if peek
```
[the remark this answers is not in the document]
asterite: Done, thanks! (not really needed, but...)

asterite:
Sorry about the constant rebase, next time I'll keep adding commits to make it easier to review and just at the end I'll rebase everything before merging (I learned `rebase -i` and now I can't stop, hehe :-P)

src/http/common.cr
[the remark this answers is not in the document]
j8r: Thanks for this tip.
But here, having an Array allocates to the heap, right? Or there is an optimization I am not aware of.
Or perhaps a `StaticArray`?

src/http/request.cr
j8r: How about making them public, as in [Golang](https://golang.org/pkg/net/http/#pkg-constants)?

src/http/common.cr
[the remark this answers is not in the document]
asterite: It allocates once for the entire duration of the program. It's negligible. It's a problem if you allocate an array in a tight loop, or even more than once when you could use a tuple. But if it's just once then Array is the best option.

src/http/request.cr
asterite: This PR isn't about adding new API. I'd rather not add API that I'm not sure how would one use.

j8r: Ok so, better make it private then.

src/http/common.cr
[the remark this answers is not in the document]
j8r: Thanks for the explanation.
However, having an Array constant is mutable (we can append elements to it), but not a Tuple constant.
This one can be made private.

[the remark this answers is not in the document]
asterite: Done private, thanks.

The mutability argument is good. But I still think in this case it doesn't matter because it's used in just one line of code.

straight-shoota:
`git commit --fixup`  is great for this. It annotates a commit as fixup of a previous one and with `git rebase -i --autosquash` they get automatically inserted at the right place.

asterite:
Ooooh... I didn't know that. I'll try it next time. Thanks!

src/http/request.cr
```diff
@@ -29,8 +29,12 @@ class HTTP::Request
   # This property is not used by `HTTP::Client`.
   property remote_address : String?
 
-  def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1")
-    @headers = headers.try(&.dup) || Headers.new
+  def initialize(method : String, resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, version = "HTTP/1.1")
```
straight-shoota: I think it would be better to provide a custom `self.new` instead of `initialize`. The `dup_headers` logic is also unnecessary, it can just be duped right here:

```cr
def self.new(method : String, resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, version = "HTTP/1.1")
  # duplicate headers to prevent the request from modifying data that the user might hold.
  new(method, resource, headers.try(&.dup), body, version, internal: nil)
end
  
private def initialize(@method : String, @resource : String, headers : Headers? = nil, body : String | Bytes | IO | Nil = nil, @version = "HTTP/1.1", *, internal)
  @headers ||= Headers.new
  self.body = body
end
```

asterite: I though about internal but it looks strange to use a named argument like thar. In fact it was my first way of doing it. Maybe we should actually expose the dup_headers argument publicly in case someone needs this performance and they know what they are doing. Thoughts?

straight-shoota: Yes, it's a bit strange. Alternatively, `headers` could just be differently named in the internal method (for example `safe_headers`) and we call it with a named argument.
In any case, I think the duplication should not happen in the internal method.

I don't think exposing it is necessary. When you create a `Request` without `headers` argument, it won't be duped and you can just use `request.headers` to build the headers.

asterite: I can't find any of these ways objectively better so for now I'd like to leave it like this.

asterite: Plus this PR doesn't change any API so we can take care of this little detail later.

straight-shoota: IMO having the duplication logic in the internal constructor is unnecessary complexity. But it's okay to leave it as is.

asterite: I changed it (so I could try `fixup`, though I already squashed... but it's useful!)

src/http/request.cr
```diff
@@ -121,6 +117,89 @@ class HTTP::Request
...
+      # TODO: eventually increase the 4096 bytes limit
+      if index && index > 0 && index < 4096 && peek[index - 1] == '\r'.ord.to_u8
+        parts = parse_request_line(peek[0, index - 1])
+        io.skip(index + 1) # Must skip until after \r\n
+        return parts
+      end
+    end
+
+    request_line = io.gets(4096, chomp: true)
+    return BadRequest.new unless request_line
```
RX14: Should this be `return nil` for EOF?

src/http/request.cr
```diff
@@ -121,6 +117,89 @@ class HTTP::Request
...
+
+  private def self.parse_request_line(slice : Bytes) : RequestLine | BadRequest
+    space_index = slice.index(' '.ord.to_u8)
+
+    # Oops, only a single part (should be three)
+    return BadRequest.new unless space_index
+
+    subslice = slice[0...space_index]
+
+    # Optimization: see if it's one of the common methods
```
RX14: Is using the `StringPool` hashing method faster or slower than searching the array here? I suspect this is faster, but I could be wrong.

src/http/request.cr
```diff
@@ -121,6 +117,89 @@ class HTTP::Request
     BadRequest.new
   end
 
+  private METHODS = %w(GET HEAD POST PUT DELETE CONNECT OPTIONS PATCH TRACE)
```
RX14: Is this faster as a tuple? (could allow LLVM to optimize the loop since size is known?)

src/http/request.cr
RX14: Of course, the fastest way is probably a manually constructed parse tree like the crystal lexer uses. That can be added later.

straight-shoota:
Thank you @asterite

RX14:
I was still reviewing this and had changes to request :<

src/http/common.cr
```diff
@@ -64,11 +67,45 @@ module HTTP
...
+    end
+  end
+
+  private def self.read_header_line(io) : HeaderLine | EndOfRequest | Nil
+    # Optimization: check if we have a peek buffer
+    if peek = io.peek
+      # peek.empty? means EOF (so bad request)
+      return nil if peek.empty?
+
+      # Find \r\n (first \n, then \r before it)
```
RX14: Previously we allowed just `\n` to end a header line. Some bad clients may well send just `\n`. They should probably be supported. NGINX supports just `\n`: https://github.com/nginx/nginx/blob/4bf4650f2f10f7bbacfe7a33da744f18951d416d/src/http/ngx_http_parse.c#L770-L777.

straight-shoota:
Oh, I'm sorry 😢 I should've just merged it yesterday right away^^

Just a suggestion: When I'm reviewing a PR that might get merged while doing that, I tend do request a review from myself to signal that I'm currently looking at it (or intend to do so).

src/http/common.cr
```diff
@@ -117,17 +159,56 @@ module HTTP
...
+    content-type
+    etag
+    expires
+    host
+    last-modified
+    user-agent
+  )
+
+  # :nodoc:
+  def self.header_name(slice : Bytes)
```
RX14: This may well be faster, easier to maintain and more generally applicable by using `StringPool`.

src/http/common.cr
RX14: This actually applies to the request line parsing too...

asterite:
By the way, I benchmarked the [simple http server](https://github.com/crystal-lang/crystal/blob/master/samples/http_server.cr) on the samples directory before and after this change with `ab`. It's just a silly example but it can show whether this change had an effect on the overall http roundtrip.

Doing:

```
ab -k -c 100 -n 200000 127.0.0.1:8080/
```

Before:

```
Time taken for tests:   1.918 seconds
Requests per second:    104299.05 [#/sec] (mean)
Time per request:       0.959 [ms] (mean)
Time per request:       0.010 [ms] (mean, across all concurrent requests)
Transfer rate:          10287.31 [Kbytes/sec] received
```

After:

```
Time taken for tests:   1.716 seconds
Requests per second:    116551.95 [#/sec] (mean)
Time per request:       0.858 [ms] (mean)
Time per request:       0.009 [ms] (mean, across all concurrent requests)
Transfer rate:          11495.85 [Kbytes/sec] received
```

And I'm almost sure if we change Hash to have an open addressing implementation it could go up to `121750.26` requests per second (just a number out of the top of my head 😉).

By comparison, doing the same benchmark against a [simple server in Go](https://gowebexamples.com/http-server/) gives these results:

```
Time taken for tests:   2.767 seconds
Requests per second:    72285.75 [#/sec] (mean)
Time per request:       1.383 [ms] (mean)
Time per request:       0.014 [ms] (mean, across all concurrent requests)
Transfer rate:          10800.51 [Kbytes/sec] received
```

However, Go handles parallelism and when we'll have parallelism the performance will get a bit worse, but on the other hand a single request doing expensive CPU won't be able to stop the server from receiving other requests.

RX14:
@asterite could you test with `wrk` instead in the future? It gives much better and more realistic results, and is the industry standard http benchmarking tool these days.

asterite:
@RX14 Sure! Here it it:

```
wrk -t12 -c400 -d30s http://127.0.0.1:8080/
```

(I don't know if those values are good, I just copied them from their github repo)

Before this PR:

```
Running 30s test @ http://127.0.0.1:8080/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.51ms  750.62us  14.12ms   80.25%
    Req/Sec     9.31k   687.19    26.23k    96.78%
  3339715 requests in 30.10s, 321.69MB read
  Socket errors: connect 0, read 244, write 0, timeout 0
Requests/sec: 110942.98
Transfer/sec:     10.69MB
```

After this PR:

```
Running 30s test @ http://127.0.0.1:8080/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.26ms  603.93us   9.90ms   82.14%
    Req/Sec     9.92k   594.01    14.69k    90.38%
  3562875 requests in 30.10s, 343.18MB read
  Socket errors: connect 0, read 239, write 0, timeout 0
Requests/sec: 118355.73
Transfer/sec:     11.40MB
```

With a "hypothetical" Hash with open addressing:

```
Running 30s test @ http://127.0.0.1:8080/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.13ms  446.90us  12.34ms   81.88%
    Req/Sec    10.33k   434.48    12.97k    83.14%
  3699575 requests in 30.00s, 356.35MB read
  Socket errors: connect 0, read 237, write 0, timeout 0
Requests/sec: 123302.09
Transfer/sec:     11.88MB
```

Go:

```
Running 30s test @ http://127.0.0.1:8080/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.78ms    7.57ms 336.03ms   94.45%
    Req/Sec     8.50k     1.07k   45.00k    96.03%
  3046719 requests in 30.10s, 374.82MB read
  Socket errors: connect 0, read 250, write 0, timeout 0
Requests/sec: 101219.02
Transfer/sec:     12.45MB
```

So the results are very different from `ab` but we can see each optimization is a bit noticeable. Also Go does pretty good too.

RX14:
@asterite is that Go with `GOMAXPROCS` or is go actually using all cores and still losing?

RX14:
Also, are you going to address the review?

asterite:
> @asterite is that Go with GOMAXPROCS=1 or is go actually using all cores and still losing?

No, that's without specifying `GOMAXPROCS` (so I guess using all cores).

If I pass `GOMAXPROCS=1` I get:

```
Running 30s test @ http://127.0.0.1:8080/
  12 threads and 400 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    18.14ms   88.69ms   2.00s    98.73%
    Req/Sec     3.14k   339.80     8.62k    85.96%
  1127120 requests in 30.10s, 138.66MB read
  Socket errors: connect 0, read 397, write 0, timeout 118
Requests/sec:  37442.65
Transfer/sec:      4.61MB
```

So I guess if with parallelism we can be close to Go numbers it'll be more than good ( /cc @waj @bcardiff )

> Also, are you going to address the review?

Yes, but in a couple of hours.

src/http/common.cr
straight-shoota: When the request line is parsed from an IO that can't peek, it already accepts `\n` because `IO#gets` reads the `\n` optionally. It should also be optional for the peek branch and header parser.

RX14: Thanks for pointing out that this does fall back to the old `gets` code if `\r\n` isn't used so it's not a regression!

straight-shoota: Yes, it falls back, but we could treat that case directly in the peek branch.

src/http/request.cr
asterite: Good catch! I think this is indeed a breaking change (though it's very unlikely that the underlying `IO`s would be peekable, but still...). Fixed in #8009

src/http/request.cr
asterite: I don't think we need a `StringPool` for this because the number of methods is very small. A parse tree is also an option, if someone else wants to try it they can do it but I think what takes most of the time is the string allocation and the IO, not the actual parsing of the method.

asterite: Plus a `StringPool` would means people could keep sending different methods, eventually consuming all memory because of it.

src/http/common.cr
asterite: Thanks! Fixed in #8009

src/http/common.cr
asterite: Same reasons for not doing that as my other comment.

src/http/request.cr
asterite: Someone else commented it. I tried it and there's no difference. I also said before that tuples shouldn't be used for everything, just for returning multiple things from methods.

PR #8003, opened 2019-07-27
Fix examples 2019 07

maiha:
Congrats 0.29.0! Let's catch up the examples code too!

- target src: master branch ( 3b2bde644 `Merge pull request #7987 from crystal-lang/ci/llvm8` )
- compiler: `Crystal 0.30.0-dev [3b2bde6] (2019-07-27)`

Best regards,

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -63,13 +63,8 @@ abstract class Channel(T)
   #
   # ```
   # channel = Channel(Int32).new
-  # spawn do
-  #   channel.send(1)
-  #   channel.send(2)
-  # end
+  # channel.send(1)
   # channel.receive # => 1
-  # channel.close
-  # channel.receive # raises Channel::ClosedError
   # ```
   def receive
     receive_impl { raise ClosedError.new }
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -63,8 +63,13 @@ abstract class Channel(T)
   #
   # ```
   # channel = Channel(Int32).new
-  # channel.send(1)
+  # spawn do
+  #   channel.send(1)
+  #   channel.send(2)
+  # end
   # channel.receive # => 1
+  # channel.close
+  # channel.receive # raises Channel::ClosedError
   # ```
   def receive
     receive_impl { raise ClosedError.new }
--- a/src/file.cr
+++ b/src/file.cr
@@ -60,7 +60,7 @@ class File < IO::FileDescriptor
   # return EOF, and any data written will be immediately discarded.
   #
   # ```
-  # File.open(File::NULL) do |file|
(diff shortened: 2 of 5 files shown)


src/channel.cr
```diff
@@ -63,8 +63,13 @@ abstract class Channel(T)
...
   # ```
   # channel = Channel(Int32).new
-  # channel.send(1)
+  # spawn do
+  #   channel.send(1)
+  #   channel.send(2)
+  # end
   # channel.receive # => 1
+  # channel.close
+  # channel.receive # raises Channel::ClosedError
```
maiha: I do not completely understand the original code's intent,
but `channel.send(1)` without `spawn` locks the operation.

Therefore, I modified the example code based on the following two points as the docs says.
1. works well when the channel is active
2. raises when the channel is not active

src/file.cr
```diff
@@ -60,7 +60,7 @@ class File < IO::FileDescriptor
   # return EOF, and any data written will be immediately discarded.
   #
   # ```
-  # File.open(File::NULL) do |file|
+  # File.open(File::NULL, "w") do |file|
```
maiha: The original code causes an error as follows:

```
Unhandled exception: File not open for writing (IO::Error)
  from crystal/src/crystal/system/unix/file_descriptor.cr:20:11 in 'unbuffered_write'
  from crystal/src/io/buffered.cr:192:5 in 'flush'
  from crystal/src/io/buffered.cr:200:5 in 'close'
  from crystal/src/file.cr:595:7 in '__crystal_main'
```

src/io.cr
```diff
@@ -29,7 +29,7 @@ require "c/errno"
 #     slice.size
 #   end
 #
-#   def write(slice : Bytes)
+#   def write(slice : Bytes) : Nil
```
maiha: #7956 makes codes more reliable 👍

src/channel.cr
bcardiff: I would prefer to leave this change out of the PR. I agree that the sample is not the best, but let’s wait for some upcoming changes in the channel. The rest of the PR looks good as usual 🎩

maiha: Yep. Just revert? or put some comments like this?

```
channel = Channel(Int32).new
channel.send(1) # TODO: This blocks
channel.receive # => 1
```

straight-shoota: When there is a sample code, it should work. Either we fix this in some way or remove it.
Maybe just use a buffered channel?

bcardiff: Just revert the changes in this file for this PR please.

maiha: Reverted. I hope Brian will write cool example or implementation soon :)

PR #8004, opened 2019-07-27
Channel::Buffered, only enqueue one waiting fiber at a time

carlhoerberg:
Speeds up `Channel::Buffered` with many senders and/or receivers. 

```
original 184.39k (  5.42µs) (± 2.92%)  0.0B/op  42.21× slower
improved   7.78M (128.49ns) (± 2.07%)  0.0B/op        fastest
```

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -223,8 +223,9 @@ class Channel::Buffered(T) < Channel(T)
     raise_if_closed
 
     @queue << value
-    Crystal::Scheduler.enqueue @receivers
-    @receivers.clear
+    if receiver = @receivers.shift?
+      Crystal::Scheduler.enqueue receiver
+    end
 
     self
   end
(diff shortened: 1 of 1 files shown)


carlhoerberg:
Gist for the benchmark: https://gist.github.com/carlhoerberg/103a130171e9ce1256317152eb9ee6e4

bcardiff:
We will check if this improvement holds/are needed with the upcoming changes. Thanks for tracking this down.

straight-shoota:
Why not merge it in right now? It's just a minimal change but provides a considerable performance improvement.
And it seems absolutely reasonable to do this. Any excess fibers are just going to be send back to wait in the `Channel` implementation.

bcardiff:
Because there is ongoing work in Channels. I would rather merge/review altogether in this case and check this in different scenarios.

RX14:
@bcardiff what ongoing work on channels? I feel I never know what's being worked on...

asterite:
This PR is now obsolete after #8112

But please open a new PR if there are optimization opportunities. Thank you!

PR #8005, opened 2019-07-28
get peer certificates and signatures

will:
There are two patches here, that go along with doing SCRAM channel binding, with this draft pr in https://github.com/will/crystal-pg/pull/177 using them as monkey patches.

The `OpenSSL::SSL::Socket#peer_certificate` I think for sure makes sense in the standard library, but the `OpenSSL::X509::Certificate#signature`, I'm not as sure. The upgrade of MD5 and SHA1 while, required in https://tools.ietf.org/html/rfc5929#section-4.1 is maybe not general purpose. The upgrade could be made optional. Then again maybe nothing else needs these signatures, so I don't know. Also I couldn't figure out how to get a test to go all the way through the method, since I don't see a way to make a `OpenSSL::X509::Certificate` with a signature.


RX14:
Could we add `Certificate#signature_algorithm` and `Certificate#digest(algorithm)` instead and move the SCRAM specific logic outside the stdlib?

will:
Yeah I think that makes sense, I’ll give it a shot soon and see what it
looks like.

will:
Updated with it split into two methods

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
       raise Error.new("X509_add_ext") if ret.null?
       extension
     end
+
+    # Returns the name of the signature algorithm.
+    def signature_name : String
+      sigid = LibCrypto.x509_get_signature_nid(@cert)
+      result = LibCrypto.obj_find_sigid_algs(sigid, out algo_nid, nil)
+      raise "could not determine certificate signature algorithm" if result == 0
```
Sija:
```suggestion
      raise "Could not determine certificate signature algorithm" if result == 0
```

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
...
+
+      sn = LibCrypto.obj_nid2sn(algo_nid)
+      raise "unknown algo NID #{algo_nid}" if sn == Pointer(UInt8).null
```
Sija:
```suggestion
      raise "Unknown algo NID #{algo_nid.inspect}" if sn.null?
```

src/openssl/ssl/socket.cr
```diff
@@ -235,4 +240,17 @@ abstract class OpenSSL::SSL::Socket < IO
...
+
+  # Returns the `OpenSSL::X509::Certificate` the peer presented, if a
+  # connection was esablished.
+  #
+  # NOTE: Due to the protocol definition, a TLS/SSL server will always send a
+  # certificate, if present. A client will only send a certificate when
+  # explicitly requested to do so by the server (see SSL_CTX_set_verify(3)). If
+  # an anonymous cipher is used, no certificates are sent. That a certificate
+  # is returned does not indicate information about the verification state.
+  def peer_certificate
```
Sija: Since the return type is nilable wouldn't be moar apt to name it `peer_certificate?`?.

src/openssl/ssl/socket.cr
```diff
@@ -235,4 +240,17 @@ abstract class OpenSSL::SSL::Socket < IO
       raise NotImplementedError.new("#{io.class}#write_timeout=")
     end
   end
+
+  # Returns the `OpenSSL::X509::Certificate` the peer presented, if a
+  # connection was esablished.
+  #
+  # NOTE: Due to the protocol definition, a TLS/SSL server will always send a
+  # certificate, if present. A client will only send a certificate when
+  # explicitly requested to do so by the server (see SSL_CTX_set_verify(3)). If
```
```suggestion
  # explicitly requested to do so by the server (see `SSL_CTX_set_verify(3)`). If
```

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
...
+
+    # Returns the digest of the certificate using *algorithm_name*
+    #
+    # ```
+    # cert.digest("SHA1").hexstring   # => "6f608752059150c9b3450a9fe0a0716b4f3fa0ca"
+    # cert.digest("SHA256").hexstring # => "51d80c865cc717f181cd949f0b23b5e1e82c93e01db53f0836443ec908b83748"
+    # ```
+    def digest(algorithm_name : String) : Slice(UInt8)
+      algo_type = LibCrypto.evp_get_digestbyname algorithm_name
+      raise ArgumentError.new "could not find digest for '#{algorithm_name}'" if Pointer(Void).null == algo_type
```
Sija:
```suggestion
      raise ArgumentError.new "Could not find digest for #{algorithm_name.inspect}" if algo_type.null?
```

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
...
+      hash = Slice(UInt8).new(64) # EVP_MAX_MD_SIZE for SHA512
+      result = LibCrypto.x509_digest(@cert, algo_type, hash, out size)
+      raise "could not generate certificate hash" unless result == 1
```
Sija:
```suggestion
      raise "Could not generate certificate hash" unless result == 1
```

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
...
+    # Returns the digest of the certificate using *algorithm_name*
+    #
+    # ```
+    # cert.digest("SHA1").hexstring   # => "6f608752059150c9b3450a9fe0a0716b4f3fa0ca"
+    # cert.digest("SHA256").hexstring # => "51d80c865cc717f181cd949f0b23b5e1e82c93e01db53f0836443ec908b83748"
+    # ```
+    def digest(algorithm_name : String) : Slice(UInt8)
+      algo_type = LibCrypto.evp_get_digestbyname algorithm_name
+      raise ArgumentError.new "could not find digest for '#{algorithm_name}'" if Pointer(Void).null == algo_type
+      hash = Slice(UInt8).new(64) # EVP_MAX_MD_SIZE for SHA512
```
Sija:
```suggestion
      hash = Bytes.new(64) # EVP_MAX_MD_SIZE for SHA512
```

src/openssl/ssl/socket.cr
straight-shoota: Not necessarily, but might be a good idea.

In any case, please specify the return type in the methods.

ysbaddaden: Adding a `?` is meant to denote a nilable _variant_ of a raising method. There is no raising `peer_certificate` method, so the method is simply returns a nilable, and shouldn't have a `?` suffix.

straight-shoota: Oh I see, the other `#peer_certificate` is in a subclass. Then no `?`.

ysbaddaden: We should start to write a style guide with such idioms, they're only said here and there. It would be nice to collect them.

src/openssl/x509/certificate.cr
```diff
@@ -59,5 +59,32 @@ module OpenSSL::X509
       raise Error.new("X509_add_ext") if ret.null?
       extension
     end
+
+    # Returns the name of the signature algorithm.
+    def signature_name : String
```
RX14: I prefer `#signature_algorithm` here.

src/openssl/ssl/socket.cr
will: I'm not sure if I followed this thread correctly. I think if I just add the return type explicitly then it'll be good?

spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -3,36 +3,53 @@ require "socket"
...
-    server_context, client_context = ssl_context_pair
+private alias Server = OpenSSL::SSL::Socket::Server
+private alias Client = OpenSSL::SSL::Socket::Client
 
-    OpenSSL::SSL::Server.open(tcp_server, server_context) do |server|
-      spawn do
-        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com") do |socket|
-        end
-      end
+private def socket_test(server_tests, client_tests)
```
straight-shoota: Maybe add empty procs as default values in order to avoid clutter in the samples?

will: > I'd like to have a positive spec on `signature_algorithm'.
> Maybe this could help: https://github.com/sfackler/rust-openssl/blob/487963d17a71dd7f4b239d7284c623f1c6a0b64b/openssl/src/x509/tests.rs#L303

Yeah, but we'd also have to add a way to load a cert from a pem which it looks like they have (`X509::from_pem`) . That's probably doable, but I'll have to figure out which functions do that.

straight-shoota: Oh right, that's missing. So go ahead if you like, either in this PR or a followup. But it doesn't need to be in here.

straight-shoota: It might be pretty straightforward, actually. See https://zakird.com/2013/10/13/certificate-parsing-with-openssl for an example. It's probably better to use `PEM_read_bio_X509` though, and we already have `OpenSSL::BIO` for bridging to `IO`.

will: :/ I'm not understanding how to use `BIO`. I tried this

```crystal
# LibCrypto
fun pem_read_bio_x509 = PEM_read_bio_X509(bp : Bio*, x590 : X509*, cb : Void*, u : Void*) : X509
```

```crystal
# Certificate
    def initialize(pem : String)
      bio = OpenSSL::BIO.new(IO::Memory.new(pem))
      # also tried this way in case it mattered if the data was already in the io
      # io = IO::Memory.new(pem.size)
      # bio = OpenSSL::BIO.new io
      # bio.io << pem
      @cert = LibCrypto.pem_read_bio_x509(bio, nil, nil, nil)
      raise "Unable to parse certificate" if @cert.null?
    end
```

```
Certificate.new("-----BEGIN CERTIFICATE-----
MIICsDCC...blah
")
```

but `PEM_read_bio_X509` is always returning null pointers. I looked into using exactly what the link you sent did, but since you have to later `free` the bio, that seems not great

straight-shoota: If you raise `OpenSSL::Error.new` it shows you the OpenSSL error message: `error:20068079:BIO routines:BIO_gets:unsupported method`.  Seems like the issue is that `BIO_meth_set_gets` is not implemented in `Bio`.
That should make it work:
```diff
--- i/src/openssl/bio.cr
+++ w/src/openssl/bio.cr
@@ -48,6 +48,19 @@ struct OpenSSL::BIO
       1
     end

+    gets = LibCrypto::BioMethodGets.new do |bio, buffer, size|
+      io = Box(IO).unbox(BIO.get_data(bio))
+      io.flush
+      string = io.gets(size)
+      if string
+        p string
+        string.to_slice.copy_to(buffer, string.bytesize)
+        string.bytesize
+      else
+        0
+      end
+    end
+
     ctrl = LibCrypto::BioMethodCtrl.new do |bio, cmd, num, ptr|
       io = Box(IO).unbox(BIO.get_data(bio))

@@ -92,6 +105,7 @@ struct OpenSSL::BIO
         LibCrypto.BIO_meth_set_write(biom, bwrite)
         LibCrypto.BIO_meth_set_read(biom, bread)
       {% end %}
+      LibCrypto.BIO_meth_set_gets(biom, gets)

       LibCrypto.BIO_meth_set_ctrl(biom, ctrl)
       LibCrypto.BIO_meth_set_create(biom, create)
@@ -103,6 +117,7 @@ struct OpenSSL::BIO
       biom.value.name = "Crystal BIO"
       biom.value.bwrite = bwrite
       biom.value.bread = bread
+      biom.value.bgets = gets
       biom.value.ctrl = ctrl
       biom.value.create = create
       biom.value.destroy = destroy
```

But it still runs into an error: `error:0906D064:PEM routines:PEM_read_bio:bad base64 decode` 
Dunno what this is about. `openssl x509 -noout -in cert.pem -text` works perfectly.

will: Yeah it's strange. It seems to be getting the data, since if you screw up the end cert line, it complains about that instead of the base64. Hmmmm

RX14: This much sounds like a new PR though.

RX14:
@will if the usage of this PR has been tested in crystal-pg, i'm fine to merge for 0.30.0.

bcardiff:
Darwin specs are failing due to missing symbols:

* _OBJ_find_sigid_algs
* _X509_get_signature_nid

Since this PR adding and not changing is easy to patch the library until work is done to make the CI in happy.

will:
@RX14 I have it on a branch for crystal-pg, but it's not merged yet. I wanted to wait for the changes to make it in before committing to it. The branch is over at https://github.com/will/crystal-pg/commits/scram_cbind

@bcardiff the darwin failures are strange to me. I develop on mac, and both openssl 1.0.2 and 1.1.0 work fine for me:

```
~/c/crystal-pg (scram_cbind)> set -gx PKG_CONFIG_PATH /usr/local/opt/openssl/lib/pkgconfig
~/c/crystal-pg (scram_cbind)> pkg-config --atleast-version=1.0.2 libcrypto || printf %s false
~/c/crystal-pg (scram_cbind)> pkg-config --atleast-version=1.1.0 libcrypto || printf %s false
false⏎                                                                                                                                                                                   ~/c/crystal-pg (scram_cbind)> crystal spec
..............................................................................................................................................

Finished in 546.24 milliseconds
142 examples, 0 failures, 0 errors, 0 pending
~/c/crystal-pg (scram_cbind)> set -gx PKG_CONFIG_PATH "/usr/local/opt/openssl@1.1/lib/pkgconfig"
~/c/crystal-pg (scram_cbind)> pkg-config --atleast-version=1.0.2 libcrypto || printf %s false
~/c/crystal-pg (scram_cbind)> pkg-config --atleast-version=1.1.0 libcrypto || printf %s false
~/c/crystal-pg (scram_cbind)> crystal spec
..............................................................................................................................................

Finished in 556.07 milliseconds
142 examples, 0 failures, 0 errors, 0 pending
~/c/crystal-pg (scram_cbind)> uname -a
Darwin Wills-MBP.guest.corp.microsoft.com 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
```

will:
Is it possible that the mac ci build isn't actually setting up openssl 1.0.2 by exporting pkg_config_path? When I don’t do that on mac everything fails to link at all (even before these changes), so I would have expected more things to fail sooner than this patch, but I don’t see "pkg_config" in the circle yaml.

RX14:
`X509_get_signature_nid` was introduced in openssl 1.1: https://github.com/openssl/openssl/commit/dfcf48f499f19fd17a3aee03151ea301814ea6ec

The linking is likely succeeding on your laptop because both openssl libraries are making their way into the library search path.

will:
Hm, the implementation requires getting info out of the X509 struct. There isn't an easy way to do that in crystal except to redefine the entire struct right? Like you cant just get one field out. 

It looks like in 1.0.2, the field is only the second one in https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/x509/x509.h#L270, and it's in the same spot today https://github.com/openssl/openssl/blob/faa9dcd4d468441422254ab2d887bb267e0245b6/crypto/include/internal/x509_int.h#L161 but not all the other fields are the same. 

1.0.2 is from January 22, 2015, what are the chances we can drop support for it?

will:
Sorry I haven't followed up on this in a while. I'm unsure of what changes to make to move this along, or what I should be doing here.

RX14:
There's no chance we can drop old openssl support, you'll probably have to get it from the struct directly, depending on openssl version.

will:
Rebased this. Curious what the current CI situation will be since the last time.

straight-shoota:
CI looks mostly good, but we'll still need support for OpenSSL 1.0.x. Unfortunately we don't have automatic tests set up against older OpenSSL versions.

AArch64 times out which doesn't happen in master, so it's likely caused by this change.

will:
> CI looks mostly good, but we'll still need support for OpenSSL 1.0.x. 

Is this still true? The OpenSSL project itself dropped support for 1.0.2 on 2019-12-31. Also 1.1.0 (though that version works with this patch) was dropped earlier on 2019-09-11. The only supported current OpenSSL version is 1.1.1 which will last until 2023-09-11. 

If it is still true, is there a timeline for how far beyond the library being unsupported by its maintainers that Crystal will support it? I'm okay doing the work here, but it'd be unfortunate if Crystal has to support dead openssl versions forever. 

Also would it be sufficient to only have these bindings and functions available for >= 1.1.0 versions? Or does it need to work for 1.0.2 also? I stopped work on this because I couldn’t figure out how to get a single field out of a struct with crystal's c biding support. The only thing I could think of was to manually read some fixed number of bytes offset, which felt gross, but maybe that's the only way.

> 
> AArch64 times out which doesn't happen in master, so it's likely caused by this change.

Do you have any tips on debugging this?

straight-shoota:
Even if the openssl project has dropped support, 1.0.x is still the supported version in older distributions like Ubuntu xenial which has LTS til April.
However, using such outdated versions really is far from recomended. So maybe it's fine to drop full support for older versions, i.e. merge this as is. Everything else still works on openssl < 1.1 but the new additions only work with more recent releases.
I would be fine with that, but let's hear other opinions on this.

Regarding Aarch64, I guess the first step would be to find out where exactly the spec execution breaks.  I don't think you can SSH into a Github actions runner. Best to use native hardware for this. I have an Raspberry Pi laying around somewhere, maybe I can take a stab at this.

will:
> I have an Raspberry Pi laying around somewhere, maybe I can take a stab at this. 
I'm taking a stab at it on an ec2 arm instance. I'm not really sure if I'm using docker right or anything, but it seems like I got some of the tests to run and build at least. I can add the ssh key you have in your github if you want to use this instance instead of a raspberry pi.

I'm not sure, but the issue may be `spec/std/openssl/ssl/socket_spec.cr` "accept raises timeout error if read_timeout is specified"

```
ubuntu@ip-172-31-24-182:~/crystal$ sudo docker run -it  -v ${PWD}:/tmp jhass/crystal:0.35.0-alpine-build /bin/sh
/ $ cd /tmp
/tmp $ bin/crystal spec spec/std/openssl/
..........................*.......................................................

Pending:
  OpenSSL::SSL::Context ciphers uses intermediate default ciphers

Finished in 2.66 seconds
82 examples, 0 failures, 0 errors, 1 pending
/tmp $ git checkout peer_cert
Switched to branch 'peer_cert'
/tmp $ bin/crystal spec spec/std/openssl/
..........................*.....................Unhandled exception in spawn: Error connecting to ':::38381': Address not available (Socket::ConnectError)
  from src/socket/addrinfo.cr:73:15 in 'initialize'
  from src/socket/tcp_socket.cr:27:3 in 'initialize'
  from src/socket/tcp_socket.cr:27:3 in 'new'
  from spec/std/openssl/ssl/socket_spec.cr:55:19 in '->'
  from src/primitives.cr:255:3 in 'run'
  from src/fiber.cr:92:34 in '->'
  from ???
^C

Pending:
  OpenSSL::SSL::Context ciphers uses intermediate default ciphers

Aborted!
Finished in 1:22 minutes
48 examples, 0 failures, 0 errors, 1 pending
```

will:
Maybe found it, trying this, but squashed it into the commit that introduced the test

``` diff
diff --git c/spec/std/openssl/ssl/socket_spec.cr i/spec/std/openssl/ssl/socket_spec.cr
index fdf16d686..f4265cd0a 100644
--- c/spec/std/openssl/ssl/socket_spec.cr
+++ i/spec/std/openssl/ssl/socket_spec.cr
@@ -47,7 +47,7 @@ private alias Server = OpenSSL::SSL::Socket::Server
 private alias Client = OpenSSL::SSL::Socket::Client

 private def socket_test(server_tests, client_tests)
-  tcp_server = TCPServer.new(0)
+  tcp_server = TCPServer.new("127.0.0.1", 0)
   server_context, client_context = ssl_context_pair

   OpenSSL::SSL::Server.open(tcp_server, server_context) do |server|
```

It looks like the localhost address was added, in these tests, and my small test refactor didn't pick it up in the rebase.

will:
I tried building on a clean xenial instance, to avoid the problem I was having earlier with multiple openssl libs on the same machine. However I did not get the same 'cant find symbol' errors that were happening on the old Darwin builds from before.

Instead it was these errors, which happened both with and without my patch 
```
ubuntu@ip-172-31-29-169:~/crystal$ lsb_release -c
Codename:	xenial
ubuntu@ip-172-31-29-169:~/crystal$ openssl version
OpenSSL 1.0.2g  1 Mar 2016
ubuntu@ip-172-31-29-169:~/crystal$ bin/crystal --version
Crystal 1.0.0-dev [cafe978] (2021-01-03)

LLVM: 8.0.0

ubuntu@ip-172-31-29-169:~/crystal$ bin/crystal spec spec/std/openssl
................................*..E..F...........................................

Pending:
  OpenSSL::SSL::Context ciphers uses intermediate default ciphers

Failures:

  1) OpenSSL::SSL::Context changes security level
     Failure/Error: context.security_level.should eq(level + 1)

       Expected: 1
            got: 0

     # spec/std/openssl/ssl/context_spec.cr:140

  2) OpenSSL::SSL::Context ciphers sets modern ciphers

       SSL_CTX_set_cipher_list: error:1410D0B9:SSL routines:SSL_CTX_set_cipher_list:no cipher match (OpenSSL::Error)
         from src/openssl/ssl/context.cr:263:5 in 'ciphers='
         from src/openssl/ssl/context.cr:284:5 in 'set_modern_ciphers'
         from spec/std/openssl/ssl/context_spec.cr:124:7 in '->'
         from src/primitives.cr:255:3 in 'internal_run'
         from src/spec/example.cr:33:16 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:330:7 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:330:7 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:147:7 in 'run'
         from src/spec/dsl.cr:274:7 in '->'
         from src/primitives.cr:255:3 in 'run'
         from src/crystal/main.cr:45:14 in 'main'
         from src/crystal/main.cr:114:3 in 'main'
         from __libc_start_main
         from _start
         from ???


Finished in 2.28 seconds
82 examples, 1 failures, 1 errors, 1 pending

Failed examples:

crystal spec spec/std/openssl/ssl/context_spec.cr:136 # OpenSSL::SSL::Context changes security level
crystal spec spec/std/openssl/ssl/context_spec.cr:123 # OpenSSL::SSL::Context ciphers sets modern ciphers
```

maybe support for 1.0.2 is already broken in crystal? I am surprised that the linking error didn't happen though

straight-shoota:
Awesome! 

The error message `Error connecting to ':::38381': Address not available` actually already showed up in the original CI failure message. It's *above* the spec progress dots. Maybe stderr is just written before stdout for some reason...

Side note: Please do not squash and force push fixes for better traceability of the intermediary changes.

straight-shoota:
It looks like #9831 already introduced optional features that don't work with openssl < 1.1, so I'd say this PR can be accepted as is.

We should probably consider documenting the support levels somewhere, though.

bcardiff:
In #9831 the policy was to allow functionality that will work only on certain version as long as

- the modules still compiles in older version
- a warn is logged

I tried to use 1.0.2 via nix and things build fine there. (`X509_get_signature_nid` was indeed present in 1.0.2 by https://github.com/openssl/openssl/commit/bd9fc1d6673f110411d03d2c2c80a25dbb5339dc)

In debian:9 de openssl is 1.0.1 and there I got

```
# bin/crystal spec spec/std/openssl/**/*_spec.cr 
O-penS-S-L-5858X-5095858C-ertificate.o: In function `signature_algorithm':
/src/src/openssl/x509/certificate.cr:65: undefined reference to `X509_get_signature_nid'
collect2: error: ld returned 1 exit status
```

I guess something similar will happen with libressl, I am not sure. It's fine if we support the propsed features in this PR only with openssl (again, as long as things compile if it's not used)

In order to move forward I think the following things are needed.

1. wrap fun declaration with `{% if compare_versions(OPENSSL_VERSION, "1.0.2") >= 0 %}`
2. use `{% if compare_versions(OPENSSL_VERSION, "1.0.2") >= 0 %}` to emit `Log.warn { "X509_get_signature_nid not supported" }` when that method is not supported and avoid trying to link an undefined symbol. The crystal method could either raise or return a stub value depending on what make sense.

bcardiff:
Using libressl (replacing openssl with libressl and `$ nix-shell --pure`)

```
[nix-shell:~/Projects/crystal/master]$ bin/crystal eval 'require "openssl"; pp! LibSSL::LIBRESSL_VERSION'
LibSSL::LIBRESSL_VERSION # => "3.0.2"

[nix-shell:~/Projects/crystal/master]$ bin/crystal eval 'require "openssl"; pp! LibSSL::OPENSSL_VERSION'
LibSSL::OPENSSL_VERSION # => "0.0.0"

[nix-shell:~/Projects/crystal/master]$ bin/crystal spec spec/std/openssl/**/*_spec.cr
........*..E..F...........F................................

Pending:
  OpenSSL::SSL::Context ciphers uses intermediate default ciphers

Failures:

  1) OpenSSL::SSL::Context changes security level
     Failure/Error: context.security_level.should eq(level + 1)

       Expected: 1
            got: 0

     # spec/std/openssl/ssl/context_spec.cr:140

  2) OpenSSL::SSL::Context .from_hash errors
     Failure/Error: expect_raises(OpenSSL::Error, "SSL_CTX_use_PrivateKey_file: error:02001002:system library:fopen:No such file or directory") do

       Expected OpenSSL::Error with "SSL_CTX_use_PrivateKey_file: error:02001002:system library:fopen:No such file or directory", got #<OpenSSL::Error: SSL_CTX_use_PrivateKey_file: error:02FFF002:system library:func(4095):No such file or directory> with backtrace:
         # src/openssl/ssl/context.cr:254:5 in 'private_key='
         # src/openssl/ssl/context.cr:474:7 in 'from_hash'
         # src/openssl/ssl/context.cr:78:7 in 'from_hash'
         # spec/std/openssl/ssl/context_spec.cr:243:9 in '->'
         # src/primitives.cr:255:3 in 'internal_run'
         # src/spec/example.cr:33:16 in 'run'
         # src/spec/context.cr:18:23 in 'internal_run'
         # src/spec/context.cr:330:7 in 'run'
         # src/spec/context.cr:18:23 in 'internal_run'
         # src/spec/context.cr:330:7 in 'run'
         # src/spec/context.cr:18:23 in 'internal_run'
         # src/spec/context.cr:147:7 in 'run'
         # src/spec/dsl.cr:274:7 in '->'
         # src/primitives.cr:255:3 in 'run'
         # src/crystal/main.cr:45:14 in 'main'
         # src/crystal/main.cr:119:3 in 'main'

     # spec/std/openssl/ssl/context_spec.cr:242

  3) OpenSSL::SSL::Context ciphers sets modern ciphers

       SSL_CTX_set_cipher_list: error:14FFF0B9:SSL routines:(UNKNOWN)SSL_internal:no cipher match (OpenSSL::Error)
         from src/openssl/ssl/context.cr:263:5 in 'ciphers='
         from src/openssl/ssl/context.cr:284:5 in 'set_modern_ciphers'
         from spec/std/openssl/ssl/context_spec.cr:124:7 in '->'
         from src/primitives.cr:255:3 in 'internal_run'
         from src/spec/example.cr:33:16 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:330:7 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:330:7 in 'run'
         from src/spec/context.cr:18:23 in 'internal_run'
         from src/spec/context.cr:147:7 in 'run'
         from src/spec/dsl.cr:274:7 in '->'
         from src/primitives.cr:255:3 in 'run'
         from src/crystal/main.cr:45:14 in 'main'
         from src/crystal/main.cr:119:3 in 'main'
       

Finished in 1.48 seconds
59 examples, 2 failures, 1 errors, 1 pending

Failed examples:

crystal spec spec/std/openssl/ssl/context_spec.cr:136 # OpenSSL::SSL::Context changes security level
crystal spec spec/std/openssl/ssl/context_spec.cr:234 # OpenSSL::SSL::Context .from_hash errors
crystal spec spec/std/openssl/ssl/context_spec.cr:123 # OpenSSL::SSL::Context ciphers sets modern ciphers
```

bcardiff:
To clarify 2. the partially supported methods should either a) warn + return a stub value or b) raise (no need to warn in this case).

will:
Hi @bcardiff, I just did your suggestions. Sorry this took so long. I was trying to get nix set up having never used it before, then ran into a lot of issues with that, then other irl stuff came up and it fell off my radar. Please let me know if this looks acceptable.

src/openssl/x509/certificate.cr
```diff
@@ -62,13 +62,17 @@ module OpenSSL::X509
...
+        raise "Could not determine certificate signature algorithm" if result == 0
 
-      sn = LibCrypto.obj_nid2sn(algo_nid)
-      raise "Unknown algo NID #{algo_nid.inspect}" if sn.null?
-      String.new sn
+        sn = LibCrypto.obj_nid2sn(algo_nid)
+        raise "Unknown algo NID #{algo_nid.inspect}" if sn.null?
+        String.new sn
+      {% else %}
+        raise "Could not determine certificate signature algorithm due to missing required OpenSSL functions"
```
straight-shoota, line 74: Maybe?
```suggestion
        raise "Missing OpenSSL function for certificate signature algorithm (requires OpenSSL 1.0.2)"
```

will: Good call

straight-shoota (approved):

will:
@bcardiff can you please take another look at this? Happy to do any other changes you think if there is anything

bcardiff (approved):
I don't follow why the `Client#peer_certificate` method is specialized to raise on nil while the base implementation is nilable.

Besides that, on my end the PR can be merged as is.

(Sorry I've been a bit disconnected)

will:
> I don't follow why the `Client#peer_certificate` method is specialized to raise on nil while the base implementation is nilable.

If you're a client, there will always be a peer certificate (the one the server is using), but if you are a server there might not bee a peer cert (the client may or may not have one). So I added `not_nil!` for the client class.

straight-shoota:
We'll squash on merge.

Merging master would be good before merging. I can do that myself.

straight-shoota:
Thank you @will

PR #8006, opened 2019-07-28
Allow timeout in select statements

carlhoerberg:
timeout keyword for use in `select`

```crystal
select
when x = ch.receive
  puts "got #{x}"
when timeout(1.seconds)
  puts "timeout"
end
```

Implemented by a `timeout_event` on the fiber, much like the `resume_event` but that sets a new Fiber property called `timed_out`, before resuming the fiber.

In contrast to:
```crystal
def sleep_select_action(timeout)
  chan = Channel(Nil).new
  spawn do
    sleep(timeout)
    chan.send(nil)
  end
  Channel::ReceiveAction.new(chan)
end
```
this is much more resource efficient, as no allocations are made (new fiber and new channel in the case above) and the timer is deleted as soon as any other event has occurred.

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -208,6 +208,8 @@ abstract class Channel(T)
   struct TimeoutAction
     include SelectAction
 
+    @timeout_at : Time::Span
+
     def initialize(timeout : Time::Span)
       @timeout_at = Time.monotonic + timeout
     end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -208,20 +208,20 @@ abstract class Channel(T)
   struct TimeoutAction
     include SelectAction
 
-    def initialize(@timeout : Time::Span)
+    def initialize(timeout : Time::Span)
+      @timeout_at = Time.monotonic + timeout
     end
 
     def ready?
-      Fiber.current.timed_out
+      Fiber.current.timed_out?
     end
 
-    def execute
+    def execute : Nil
       Fiber.current.timed_out = false
-      nil
     end
 
     def wait
-      Fiber.timeout @timeout
+      Fiber.timeout @timeout_at - Time.monotonic
     end
 
     def unwait
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -203,6 +203,31 @@ abstract class Channel(T)
       @channel.unwait_for_send
     end
   end
+
+  # :nodoc:
+  struct TimeoutAction
+    include SelectAction
+
+    def initialize(@timeout : Time::Span)
+    end
+
+    def ready?
+      Fiber.current.timed_out
+    end
+
+    def execute
+      Fiber.current.timed_out = false
+      nil
+    end
+
+    def wait
+      Fiber.timeout @timeout
+    end
+
+    def unwait
+      Fiber.cancel_timeout
+    end
+  end
 end
 
 # Buffered channel, using a queue.
--- a/src/concurrent.cr
+++ b/src/concurrent.cr
@@ -30,14 +30,16 @@ def sleep
   Crystal::Scheduler.reschedule
 end
 
-# timeout keyword for use in `select`
+# Timeout keyword for use in `select`.
 #
+# ```
 # select
 # when x = ch.recieve
 #   puts "got #{x}"
 # when timeout(1.seconds)
 #   puts "timeout"
 # end
+# ```
 def timeout_select_action(timeout)
   Channel::TimeoutAction.new(timeout)
 end
--- a/src/concurrent.cr
+++ b/src/concurrent.cr
@@ -30,6 +30,18 @@ def sleep
   Crystal::Scheduler.reschedule
 end
 
+# timeout keyword for use in `select`
+#
+# select
+# when x = ch.recieve
+#   puts "got #{x}"
+# when timeout(1.seconds)
+#   puts "timeout"
+# end
+def timeout_select_action(timeout)
+  Channel::TimeoutAction.new(timeout)
+end
+
 # Spawns a new fiber.
 #
 # The newly created fiber doesn't run as soon as spawned.
--- a/src/crystal/scheduler.cr
+++ b/src/crystal/scheduler.cr
@@ -34,6 +34,14 @@ class Crystal::Scheduler
     Thread.current.scheduler.sleep(time)
   end
 
+  def self.timeout(time : Time::Span) : Nil
+    Thread.current.scheduler.timeout(time)
+  end
+
+  def self.cancel_timeout : Nil
+    Thread.current.scheduler.cancel_timeout
+  end
+
   def self.yield : Nil
     Thread.current.scheduler.yield
   end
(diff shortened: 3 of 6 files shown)


carlhoerberg:
That `Fiber#timed_out` property must be reset, right now it's done in the `TimeoutSelectAction`, but maybe it should be a "read-once" property in the Fiber instead? 

```crystal
# Indicated that the fiber has been woken up by a timeout timer.
# Can only be read once, is automatically restored to false.
def timed_out
  @timed_out
ensure
  @timed_out = false
end

# Should only be set from the event callback
def timed_out=(value)
  @timed_out = value
end
```

bcardiff:
Thanks! This looks good overall. Let's wait until the next changes regarding parallelism land before merging.

bararchy:
@bcardiff any idea when those land? +\-

bcardiff:
@bararchy as soon as we can. We are trying to finish the release first and then keep iterating with some open stories still.

src/channel.cr
```diff
@@ -203,6 +203,31 @@ abstract class Channel(T)
...
+      Fiber.current.timed_out
+    end
+
+    def execute
+      Fiber.current.timed_out = false
+      nil
+    end
+
+    def wait
+      Fiber.timeout @timeout
```
RX14: This is incorrect, as `#wait` can be called multiple times.

RX14:
This can be merged for 0.30.0 @bcardiff. This is orthogonal to any parallelism changes.

asterite:
@RX14 I believe if we merge this it means more merge conflicts on the paralleism branch, with the only result of delaying its landing. 

If someone really wants this now they can always monkey-patch things.

(what I'm saying is, I don't think this is super urgent)

RX14:
This isn't difficult to fix any merge conflicts... it's largely a copy and paste of the existing `sleep` code.

src/concurrent.cr
```diff
@@ -30,6 +30,18 @@ def sleep
   Crystal::Scheduler.reschedule
 end
 
+# timeout keyword for use in `select`
+#
+# select
```
Sija: Example code should be fenced with triple backticks.

src/concurrent.cr
```diff
@@ -30,6 +30,18 @@ def sleep
   Crystal::Scheduler.reschedule
 end
 
+# timeout keyword for use in `select`
```
Sija:
```suggestion
# Timeout keyword for use in `select`.
```

src/fiber.cr
```diff
@@ -19,8 +19,10 @@ class Fiber
   @context : Context
   @stack : Void*
   @resume_event : Crystal::Event?
+  @timeout_event : Crystal::Event?
   protected property stack_bottom : Void*
   property name : String?
+  property timed_out = false
```
Sija:
```suggestion
  property? timed_out = false
```

src/channel.cr
```diff
@@ -203,6 +203,31 @@ abstract class Channel(T)
...
+    include SelectAction
+
+    def initialize(@timeout : Time::Span)
+    end
+
+    def ready?
+      Fiber.current.timed_out
+    end
+
+    def execute
```
Sija:
```suggestion
    def execute : Nil
```

src/channel.cr
```diff
@@ -203,6 +203,31 @@ abstract class Channel(T)
...
+
+  # :nodoc:
+  struct TimeoutAction
+    include SelectAction
+
+    def initialize(@timeout : Time::Span)
+    end
+
+    def ready?
+      Fiber.current.timed_out
```
Sija:
```suggestion
      Fiber.current.timed_out?
```

src/crystal/scheduler.cr
```diff
@@ -123,4 +131,12 @@ class Crystal::Scheduler
     @current.resume_event.add(0.seconds)
     resume(fiber)
   end
+
+  protected def timeout(time : Time::Span) : Nil
+    @current.timeout_event.add(time)
+  end
+
+  protected def cancel_timeout : Nil
```
carlhoerberg: Or should it be called `delete_timeout`, to make it more coherent?

carlhoerberg:
@ysbaddaden is the MT branch publicly available some where?  

Is MT expected to be included in the next crystal release? If not, maybe this can still be merged for the next release so that we have the API for it? The implementation can later change of course.

ysbaddaden:
I don't know which branch @bcardiff is working on. There is my experimental branch #7547 where... `select` is disabled because of the issue detailed above + only *one* branch must ever be executed (otherwise we could dequeue N channels in parallel and lose N-1 messages), which is complex. The argument still stands anyway: using `select` for timeouts to avoid spawning fibers will probably be reverted when MT lands, or select is refactored for MT.

That being said, I like this pull request. It's simple an implements timeouts for channels as it was designed to be: as a `select` condition. I'm just not fond of `select` myself, and wanted to emphasize that maybe it wasn't the best solution in the mid/long term, or the expected one (use less memory by avoiding to spawn fibers), thought there may be solutions to that (spawn fibers to use a stack from the HEAP instead of necessarily be a mmap).

So, all in all, maybe this pull request is perfectly fine. We must just wait for @bcardiff to tell whether it really is (if select & channel are being refactored, as it would seem).

carlhoerberg:
I see. The problem we want to avoid is to allocate memory that isn't released as soon as `select` returns, as that will inevitably lead to a memory leak (and OOM if the timeout is long and the `select` called often enough).

RX14:
@ysbaddaden In the future select will probably be implemented entirely differently, definitely without fibers being spawned, and probably using a custom event loop implementation which will handle the select logic (discard the fiber resume if the select has already been fired)

RX14:
Actually, this could be implemented with a per-fiber "select resume counter", the count at the start of the select is passed to channels or libevent timers which may want to resume the timer. On resume time, you cas the counter in the `Fiber` with count + 1, if you succeed you resume, else you discard your resume (and in the case of channels, resume the next possible fiber on the waitlist). 

Then you have to deal with overflows, but 64 bits is probably enough?

src/channel.cr
```diff
@@ -203,6 +203,33 @@ abstract class Channel(T)
...
+    def ready?
+      Fiber.current.timed_out?
+    end
+
+    def execute : Nil
+      Fiber.current.timed_out = false
+    end
+
+    def wait
+      Fiber.timeout @timeout_at - Time.monotonic
```
RX14: What happens if `@timeout_at - Time.monotonic` goes negative?

src/channel.cr
```diff
@@ -203,6 +203,33 @@ abstract class Channel(T)
...
     end
   end
+
+  # :nodoc:
+  struct TimeoutAction
+    include SelectAction
+
+    @timeout_at : Time::Span
+
+    def initialize(timeout : Time::Span)
```
RX14: I'd prefer to do the `timeout_at` calculation in `timeout_select_action`. That means a `timeout_at_select_action` can be added in the future, which accepts a timestamp instead of a time from now.

bcardiff:
@carlhoerberg Would you be willing to rebase this on top of #8322? The `SelectAction` API changes are stable in that PR but there are still some implementation details before merging it.

If you are not available to do so, let us know and we can pick it from here.

Thanks!

Issue #7829, opened 2019-05-29
Process user API


straight-shoota:
There have been some PR's adding features for querying and changing the user and group of a process, originally as part of #5615 then it's reboot #5627 and now there's #7822, which is a completely new approach.

We've had discussions about the design of these APIs directly in the PRs, but I'd like to discuss the general design in separate and focus the PR on the implementation.

From #5615
```cr
Process.become(user : User)
Process.become(group : Group)
Process.user : User
Process.group : Group
```

From #5627
```cr
Process.owner=(user : System::User)
Process.group=(group : System::Group)
Process.user : System::User
Process.group : System::Group
```

From #7822
```cr
Process.uid=(uid : LibC::UidT)
Process.uid : LibC::UidT
Process.euid : LibC::UidT
Process.gid=(gid : LibC::GidT)
Process.gid : LibC::GidT
Process.egid : LibC::GidT
Process.setresuid(uid : LibC::UidT)
Process.setresgid(gid : LibC::GidT)
```

The most important difference between these proposals (besides varying feature support) is clearly whether they work on plain uid/gid or on `System::User`/`System::Group` instances (see #7738 for that API). Using `LibC::UidT`/`LibC::GidT` is obviously not portable to non-POSIX platforms, but returning `Int32` (or `String`) could work. Int values are used for example in the [Ruby](https://ruby-doc.org/core-2.6.1/Process.html) and [Golang](https://golang.org/pkg/os/#Getuid) implementations.

Both approaches are good. Maybe we could even want both, I don't know.

It seems pretty difficult to create a portable interface, that can work for non-POSIX platforms. Most API's I've seen only support POSIX and more ore less closely resemble the POSIX calls.
So, ideas for cross-platform implementation would be great =)

j8r:
Having `Process.owner=(user : System::User)` is strange, like [File::Info#owner : Int32](https://crystal-lang.org/api/master/File/Info.html#owner%3AUInt32-instance-method) (that returns and user id, or uid).
Whatever is the API, we should use the `user` word instead of `owner`. It's used on POSIX, and on Windows (not sure).

yxhuvud:
If we want to make it very obvious that it only work on POSIX, then perhaps doing something like `Process::POSIX.owner=` could be reasonable?

didactic-drunk:
Most of the proposed high level API's seems to assume the program only wants to switch user's once for privilege dropping.

[...]

PR #8007, opened 2019-07-29
Rework File::Info user and group methods

j8r:
Since https://github.com/crystal-lang/crystal/pull/7725 is merged, the `File::Info` can be improved accordingly.
Several points are addressed in this PR, devised into separate commits:
- `File::Info#owner` and `File::Info#group`  are renamed to `File::Info#user_id` and `File::Info#group_id`
  - to match `System::User#id` and `System::Group#id`
  - more importantly because they return IDs (as String)
- [Controversial Commit] Use `UInt32` instead of `String` for user/group ids
  - because that what they are - Integer Inentifiers
  - this drops the need to cast several time with `to_s` then back to `to_u32`, the user need to pre-validate its type if coming from a `String`
  - for those worried about cross-platform compatibility,  see the last commit solving this issue adding system `#user`/`#group` to `File::Info`
-  Add `File::Info#user` and `File::Info#group`
   - this is supposed to be a generic user/group interface
   - this prevents to be forces to use UInt32 IDs as String, and then casting them back.
___
In fact, `File::Info#user_id` and `File::Info#group_id` are UN*X specific (on Windows as for now, `0` is hard-coded), there are several options, not sure what's would be the best:
- having only `File::Info#user` and `File::Info#group`, then call `#id`, but this means having extra calls and operations when wanting to only retrieve an ID
- having something like `File::Info#unix` with all UN*X specificities inside it, like `gid`, `uid`, `mtime`, `atime`, etc. This method will raise at compile time if called on a Windows system.

### DIFF
--- a/src/crystal/system/unix/user.cr
+++ b/src/crystal/system/unix/user.cr
@@ -4,7 +4,7 @@ module Crystal::System::User
   private GETPW_R_SIZE_MAX = 1024 * 16
 
   private def from_struct(pwd)
-    user = String.new(pwd.pw_gecos).split(",").first
+    user = String.new(pwd.pw_gecos).partition(',')[0]
     new(String.new(pwd.pw_name), pwd.pw_uid.to_s, pwd.pw_gid.to_s, user, String.new(pwd.pw_dir), String.new(pwd.pw_shell))
   end
 
--- a/src/file/info.cr
+++ b/src/file/info.cr
@@ -93,11 +93,23 @@ class File
     # The last time this file was modified.
     abstract def modification_time : Time
 
-    # The user ID of the file's owner.
-    abstract def owner : UInt32
+    # The user ID that the file belongs to.
+    abstract def owner_id : String
+
+    # :ditto:
+    @[Deprecated("Use File#owner_id instead")]
+    def owner : UInt32
+      owner_id.to_u32
+    end
 
     # The group ID that the file belongs to.
-    abstract def group : UInt32
+    abstract def group_id : String
+
+    # :ditto:
+    @[Deprecated("Use File#group_id instead")]
+    def group : UInt32
+      group_id.to_u32
+    end
 
     # Returns true if this `Info` and *other* are of the same file.
     #
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -393,58 +393,58 @@ describe "File" do
     end
   end
 
-  it "gets info for this file" do
-    info = File.info(datapath("test_file.txt"))
-    info.type.should eq(File::Type::File)
-  end
-
-  it "gets info for this directory" do
-    info = File.info(datapath)
-    info.type.should eq(File::Type::Directory)
-  end
-
-  # TODO: support stating nul on windows
-  pending_win32 "gets info for a character device" do
-    info = File.info(File::NULL)
-    info.type.should eq(File::Type::CharacterDevice)
-  end
-
-  it "gets info for a symlink" do
-    info = File.info(datapath("symlink.txt"), follow_symlinks: false)
-    info.type.should eq(File::Type::Symlink)
-  end
-
-  it "gets info for open file" do
-    File.open(datapath("test_file.txt"), "r") do |file|
-      info = file.info
+  describe "File::Info" do
+    it "gets for this file" do
+      info = File.info(datapath("test_file.txt"))
       info.type.should eq(File::Type::File)
     end
-  end
 
-  it "gets info for pipe" do
-    IO.pipe do |r, w|
-      r.info.type.should eq(File::Type::Pipe)
-      w.info.type.should eq(File::Type::Pipe)
+    it "gets for this directory" do
+      info = File.info(datapath)
+      info.type.should eq(File::Type::Directory)
     end
-  end
 
-  it "gets info for non-existent file and raises" do
-    expect_raises_errno(Errno::ENOENT, "Unable to get info for 'non-existent'") do
-      File.info("non-existent")
+    # TODO: support stating nul on windows
+    pending_win32 "gets for a character device" do
+      info = File.info(File::NULL)
+      info.type.should eq(File::Type::CharacterDevice)
     end
-  end
 
-  it "gets info mtime for new file" do
-    with_tempfile("mtime") do |path|
-      File.touch(path)
-      File.open(path) do |file|
-        file.info.modification_time.should be_close(Time.utc, 1.seconds)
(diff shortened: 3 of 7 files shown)


straight-shoota:
We decided for the type of ids to be `String` for more flexibility. That may not be set in stone, but I don't think we'll just change that in the next PR, unless there are some convincing new arguments (which I'm totally missing here). They are not strictly *integer identifiers* but *user identifiers*. Most operating systems just happen to represent them using integers (btw. the POSIX standard allows for signed integers).

A good solution to avoid converting between uint32 and string for `File::Info#user` would be to utilize the private, internal API to create a `System::User` instance directly from UInt32. But that's just a performance improvement and doesn't need to be implemented right away.

RX14:
We literally just discussed and decided that IDs will be `String` and then merged it.

Of course it's not changing.

j8r:
A public, `:nodoc:` overload `from_id(id : UInt32)` can be added to overcome this.

Therefore, do we agree on only having `File::Info#user : System::User` and `File::Info#group : System::Group`?

RX14:
we can have `#user_id : String`

j8r:
Does it mean `user_id` will return an SID for Windows, in the future?
It doesn't look like portable.

In UN*X, we won't what `#user : String` will return by only looking at the method name (UID, name?), and in Windows, same for `#user_id` (SID, RID?)

I think the only more or less portable solution is `#user : System::User`.

j8r:
Does it mean `user_id` will return an SID for Windows, in the future?
It doesn't look like portable.

In UN*X, we won't know what `#user : String` will return by only looking at the method name (UID, name?), and in Windows, same for `#user_id` (SID, RID?)

I think the only more or less portable solution is `#user : System::User`.

didactic-drunk:
Can the platform agnostic API's be `File::Info#owner` and `File::Info#group` and the platform specific methods be `[user_id, group_id, security_id]`?  

Someone attempting to get the `SID` is likely doing it for a platform specific purpose, just like someone using user/group id's is likely attempting something platform specific.


An application in a chroot environment may not have access to /etc/passwd.  WIth the current implementation that would raise an exception on `file.user`.

Here's an [visudo like example](https://github.com/chris-huxtable/user_group.cr/issues/1#issuecomment-508589202) program that switches uid's/gid's without opening /etc/passwd.  If the program is modified to either validate a file user/group id against the process user/group id or chown a file, actual id's and not System::User/Group's are necessary under at least the following conditions.

* `chroot` where /etc/passwd is unavailable
* Running setuid and attempting to not open /etc/passwd so root's and other password's aren't left in memory.
* Adding resiliency to a long running program (web server) against account deletion or username/group/uid/gid reuse.

j8r:
We can also be able to create `System::User` with an ID, then, eventually, lazily load others getters when needed.

`#owner` is confusing, neither UN*X, Windows or any other OS I know (please prove me wrong, still it would be a minotrity).

j8r:
We can also be able to create `System::User` with an ID, then, eventually, lazily load others getters when needed.

`#owner` is confusing, neither in UN*X, Windows or any other OS I know (please prove me wrong, still it would be a minotrity).

src/crystal/system/unix/user.cr
```diff
@@ -4,7 +4,7 @@ module Crystal::System::User
   private GETPW_R_SIZE_MAX = 1024 * 16
 
   private def from_struct(pwd)
-    user = String.new(pwd.pw_gecos).split(",").first
+    user = String.new(pwd.pw_gecos).partition(',')[0]
```
Sija: Why?

asterite: `partition` might be a bit faster (it returns a tuple instead of an array)

j8r:
@RX14 why having the `owner` name, everywhere else `user` is used? POSIX and Windows even use the "user" word. Except that, I'm ok with the API.

src/crystal/system/unix/user.cr
Sija: If so, maybe this would read a bit better?

```suggestion
    user = String.new(pwd.pw_gecos).partition(',').first
```

RX14:
It's called an owner in both APIs...

See the `chown` manpage: `int chown(const char *path, uid_t owner, gid_t group);`

in `sys/stat.h`:
```
    __uid_t st_uid;     /* User ID of the file's owner. */
```

It's called an owner in windows too. It's just got a *type* of "user"

j8r:
We don't implement `#owner : System::User` and `#group : System::Group` at the end?

RX14:
No, because that would be a breaking change. They must be deprecated first.

j8r:
Can we decide either having `user` or `owner` everywhere, please. Same for the [Process API](https://github.com/crystal-lang/crystal/issues/7829).
Otherwise, we will always mix up the two.

straight-shoota:
I think `File::Info#owner` and `Process.user` are fine. They are separate concepts and don't need to share a name.

RX14:
@j8r the owner of a file is a user, but a process is run *by* a user. The naming is fine and will not change.

src/crystal/system/unix/user.cr
straight-shoota: `partition` still allocates two unnecessary strings. A more efficient implementation would just scan for the first index of `,` and take a slice. This is even possible before creating a `String` instance, reducing the total amount of memory allocations to one.
```cr
user = String.new(pwd.pw_gecos, pwd.pw_gecos.to_slice(LibC.strlen(pwd.pw_gecos)).index(','.ord))
```

This could even avoid `LibC.strlen` because the `,` index will always be smaller. But this would mean to search through the char pointer manually, making the code more convoluted.

j8r: Hum, not sure it's worth to complicate that much for the gain, at this point.

j8r:
Ready for reviews

RX14:
This is a small deprecation, should it make it into the release?

src/file/info.cr
```diff
@@ -93,11 +93,23 @@ class File
...
+    @[Deprecated("Use File#owner_id instead")]
+    def owner : UInt32
+      owner_id.to_u32
+    end
 
     # The group ID that the file belongs to.
-    abstract def group : UInt32
+    abstract def group_id : String
+
+    # :ditto
```
straight-shoota:
```suggestion
    # :ditto:
```

j8r:
Is it good to merge for the core team?

straight-shoota:
Thanks @j8r

PR #8008, opened 2019-07-29
Add doc property to CStructOrUnionDef AST node

olbat:
I'm not sure why the CStructOrUnionDef AST node doesn't have a _doc_ property but adding it will be useful in https://github.com/crystal-lang/crystal_lib/pull/61 .

### DIFF
--- a/src/compiler/crystal/syntax/ast.cr
+++ b/src/compiler/crystal/syntax/ast.cr
@@ -1819,6 +1819,7 @@ module Crystal
     property name : String
     property body : ASTNode
     property? union : Bool
+    property doc : String?
 
     def initialize(@name, body = nil, @union = false)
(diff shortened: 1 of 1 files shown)


asterite:
It's because these don't show up in docs and will probably never will (users of your API shouldn't have to interact with C structs).

olbat:
I was thinking it could be useful for people that are generating and maintaining bindings to C libraries but maybe it's not a good idea in the end...

RX14:
But does just adding the property do anything if it's not used anywhere? I vote to close this.

asterite:
It does absolutely nothing. Well, it does increase memory footprint.

So yeah, let's close it. We can reopen it if we decide we want these to show in API docs.

PR #8009, opened 2019-07-29
HTTP::Request: follow up fixes

asterite:
Follow up to #8002

### DIFF
--- a/spec/std/http/request_spec.cr
+++ b/spec/std/http/request_spec.cr
@@ -1,6 +1,15 @@
 require "spec"
 require "http/request"
 
+private class EmptyIO < IO
+  def read(slice : Bytes)
+    0
+  end
+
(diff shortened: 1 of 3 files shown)


spec/std/http/request_spec.cr
```diff
@@ -138,6 +154,14 @@ module HTTP
       request.headers.should eq({"Host" => "host.example.org"})
     end
 
+    it "parses empty string (EOF), returns nil" do
+      Request.from_io(IO::Memory.new("")).should be_nil
+    end
+
+    it "parses empty string (EOF), returns nil (no peek)" do
+      Request.from_io(EmptyIO.new).should be_nil
```
straight-shoota: This is fine, but it's actually the only spec explicitly for a non-peek IO. We should probably run every request spec against both a peekable and non-peekable IO (not in this PR though).

PR #8010, opened 2019-07-29
Only report abstract-related warnings on warnings=all

asterite:
Follow up to #7999

### DIFF
--- a/src/compiler/crystal/semantic/abstract_def_checker.cr
+++ b/src/compiler/crystal/semantic/abstract_def_checker.cr
@@ -160,7 +160,7 @@ class Crystal::AbstractDefChecker
 
     original_base_return_type = base_type.lookup_type?(base_return_type_node)
     unless original_base_return_type
-      report_warning(base_return_type_node.warning "can't resolve return type #{base_return_type_node}\n#{this_warning_will_become_an_error}")
+      report_warning(base_return_type_node, "can't resolve return type #{base_return_type_node}\n#{this_warning_will_become_an_error}")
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/semantic/abstract_def_checker.cr
```diff
@@ -221,6 +221,10 @@ class Crystal::AbstractDefChecker
     @program.colorize("The above warning will become an error in a future Crystal version.").yellow.bold
   end
 
+  private def report_warning(warning)
```
bcardiff: We should be filtering by location also. The `ignore_warning_due_to_location` method has the logic for this. 
I think that calling it with `base_method.location` should do it. Otherwise, the warnings triggered by the dependencies will appear.

Of course, it's a balance and that hides a bit the need for an update in dependencies. Similar to the formatter tool that by default ignore libs, warnings do that as well (yet due to how does flags work the user is unable to _include_ libs in the check...).

So, if you explicitly prefer to not filter by location I am ok for this PR.

asterite: Oh, I didn't know warnings were ignored for libs. That makes sense. I'll fix it now.

asterite: Done! Let me know what you think, I moved the `ignore_warning_due_to_location` to `Program` so I could reuse it (I also think it makes sense for it to be there). I also added a `?` at the end of it and returned an extra `return`.

Issue #5651, opened 2018-01-28
Errors of macro def variables


delef:
Hello everyone! I faced such behavior, is this normal?

Syntax error in eval:5: expecting token 'CONST', not '{{'
```crystal
module Asd::But::Me::Ar
  class Test
    def remote_api
      Other::{{@type.name.split("::").last.capitalize.id}}.new
    end
  end
end

module Other
  class Test
    def hi
      pp "hi"
    end
  end
end

Asd::But::Me::Ar::Test.new.remote_api.hi
```

#### Working example
```crystal
module Asd::But::Me::Ar
  class Test
    def remote_api
      {{("Other::" + (@type.name.split("::").last.capitalize)).id}}.new
    end
  end
end

module Other
  class Test
    def hi
      pp "hi"
    end
  end
end

Asd::But::Me::Ar::Test.new.remote_api.hi
```
=> "hi" # => "hi"

#### Other problems:
Error in line 18: instantiating 'Asd::But::Me::Ar::Test#remote_api()'
in line 5: expanding macro
in line 5: undefined macro variable 'name'

```crystal
module Asd::But::Me::Ar
  class Test
    def remote_api
      {% name = @type.name.split("::").last %}
      "{{name.capitalize.id}}"
    end
  end
end

module Other
  class Test
    def hi
      pp "hi"
    end
  end
end

Asd::But::Me::Ar::Test.new.remote_api
```

Thanks for your help in advance!

```
Crystal 0.24.1 (2017-12-26)

LLVM: 5.0.1
Default target: x86_64-apple-macosx
```

asterite:
`Other:{{...}}` is invalid syntax. You can't just stick macros there. They must appear where a single expression appears (not in the middle of an identifier).

If you need that, you can wrap everything in a macro:

```crystal
{% begin %}
  # your code here
{% end %}
```

asterite:
Note that `{{...}}.new` is valid, because `exp.new` is valid. However, `Foo::{{bar}}` is not, because the parser consumes `Foo::Bar::...` as a single expression (a `Path` in ASTNode terms), not as a composition of several expressions.

delef:
Thanks

```crystal
{% name = @type.name.split("::").last %}
"{{name.capitalize.id}}"
```

why name is undefined macro variable?

asterite:
Because those are two, unrelated macro expansions.

To make them related, wrap both in a single `{% begin %} ...{% end %}`.

delef:
[...]

PR #8011, opened 2019-07-30
[WIP] File.new POSIX like modes.

didactic-drunk:
This PR only contains basic functionality.  Locking will be handled in another PR.


src/file/mode.cr
```diff
@@ -0,0 +1,16 @@
+class File
+  @[Flags]
+  enum Mode
+    Read      = LibC::O_RDONLY
```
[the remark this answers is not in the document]
didactic-drunk:
```
enum Crystal::System::File::Flags
  Read = LibC...
end

enum File::Flags
  Read = Crystal::System::File::Flags::Read
end
```

Is this what it should look like?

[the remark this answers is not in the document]
straight-shoota: No, `File::Mode` should be an enum with completely independent values that only need to have a meaning inside the Crystal API. Then, in `Crystal::System` these are translated to system-specific flags.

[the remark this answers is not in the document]
didactic-drunk: Do you mean:
```
enum File::Mode
 Read, Write, ...
end
```

```
Crystal::System::File::ModeMap = {
  File::Mode::Read => LibC::O_RDONLY,
  ...
}

```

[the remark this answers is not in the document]
straight-shoota: Probably not implemented as a map, but essentially, yes.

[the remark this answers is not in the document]
didactic-drunk: Right now it's a map.  How else could it be done concisely?

There is one odd error commented in `Crystal::System::File` that I don't understand.  The commented statements are more robust against Mode additions/subtractions.

RX14:
Do you mind if I take this PR for myself? I'd like to implement it and play around with it in a certain way on my own.

didactic-drunk:
If you are wondering why this implementation is so bare, it's based on other PR's (#5651, #5627, #7857, #7931 etc) with tons of design commentary requiring redesigns or closing/reopening issues with completely new code.  

I thought a minimalist design would open avenues for comments or a shared code base rather than me writing a complete detailed implementation that has to be scrapped, which already seems like that may be happening.

@RX14 I added you as a collaborator on my fork.

didactic-drunk:
Robustness achieved.  Missing flags are caught.  The prior error disappeared after rebasing.  Maybe my crystal was out of sync.

src/crystal/system/file.cr
```diff
@@ -1,8 +1,34 @@
+require "file/mode"
+
 # :nodoc:
 module Crystal::System::File
+  private ModeMap = {
```
RX14: I'd prefer this not to be implemented with a hash, it's slow. Manual logic is better here:

```cr
flags |= O_CREAT if mode.create?
```

didactic-drunk: Aaaaaasssssss yoooouuuuu wwissssshhhhhh.

didactic-drunk: I just noticed you had two comments.

> I'd rather have `:read, :write` than `:readwrite`. That's what I thought we agreed on in #7857 

Ok.

> Using `:append` with `:write` should be an error.

Not harmless?  Ok.

> And using `:sync` should probably set `IO#sync`??? Not sure on the last one.


Applications using sync would probably flush at specific points and possibly want `IO` buffering to reduce write load no?

```
io.write_bytes frame_length
io.write data
io.write_bytes checksum
io.flush
```

> And it needs a `*mode : Mode` overload which allows calling `File.open("foo", :read, :write)`.

How does that work with additional args?  Which of these if any should I use?
```
def self.new(filename : Path | String, mode : Mode = File::Mode.flags(Read), perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil)
def self.new(filename : Path | String, mode : Mode = File::Mode.flags(Read), *, perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil)
def self.new(filename : Path | String, mode : Mode = File::Mode.flags(Read))
```

RX14: You want to add a new

```cr
def self.new(filename : Path | String, *modes : Mode, perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil)
  new(filename, modes.reduce(&.|), encoding, invalid)
end
```

RX14: > Applications using sync would probably flush at specific points and possibly want `IO` buffering to reduce write load no?

yeah that's what I thought, let's not do this

RX14: > Not harmless? Ok.

It's just a bit confusing behaviour, is all. Since writes will all be at the end but reads won't?

I'm not 100% sure about disallowing it, it just seems confusing behaviour for the end user if they accidentally do this. And if it's got a usecase, then we can easily allow it in the future.

didactic-drunk: You said "Using :append with :write should be an error.".  Now your talking about read. Did you mean :read with :append?

didactic-drunk: Splats are new to me.  I have two errors I'm not sure how to resolve.

`File.new "foo", :read, :write`

```
In src/file.cr:111:62

 111 | fd = Crystal::System::File.open(filename, modes.reduce(&.|), perm)
                                                                ^
Error: wrong number of arguments for 'File::Mode#|' (given 0, expected 1)
```


`File.new "foo", :read, :write, :create_new` 

```
In src/file.cr:78:10

 78 | self.set_encoding(encoding, invalid: invalid) if encoding
           ^-----------
Error: no overload matches 'File#set_encoding' with types Symbol, invalid: Nil
```

I tried `[:create_new, :createnew, :CreateNew, :Create_New]`.  Do I need to type restrict `encoding` and `invalid`?  Are enum splats broken with camelcase?

RX14: > Did you mean :read with :append?

Shit, my bad, yes I did.

didactic-drunk: A replicated transaction log may want :read with :append.  One writer appends.  Clients dynamically connect and start streaming from an arbitrary point.  Kafka uses this design.  Many database replication implementations probably do something similar.

RX14: Okay, we can allow it.

BTW, `O_RDONLY | O_WRONLY != O_RDWR`, which this code currently assumes.

I'll work on a workaround for the bug above.

didactic-drunk: A replicated transaction log may use :read with :append.  One writer appends.  Clients dynamically connect and streaming from an arbitrary point.  Kafka uses this design.  Many database replication implementations probably do something similar.

didactic-drunk: I assumed you were busy in the past week so I did my own.

src/file.cr
```diff
@@ -583,6 +619,7 @@ class File < IO::FileDescriptor
   # permissions may be set using the *perm* parameter.
   #
   # See `self.new` for what *mode* can be.
+  @[Deprecated("Use `File.new(filename, :read, :write, :create)`")]
```
RX14: `File.open`

src/file.cr
```diff
@@ -628,7 +629,31 @@ class File < IO::FileDescriptor
...
+  def self.open(filename : Path | String, mode : Mode = File::Mode.flags(Read), perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil)
+    file = new filename, mode, perm, encoding, invalid
+    begin
+      yield file
+    ensure
+      file.close
+    end
+  end
+
+  @[Deprecated("Use `File.new(filename, :read, :write, :create)`")]
```
RX14: ditto

src/crystal/system/file.cr
```diff
@@ -6,6 +6,10 @@ module Crystal::System::File
   # calls.
   private def self.open_flag(mode : ::File::Mode)
     flags = 0
+    if mode.read? && mode.write?
+      flags |= LibC::O_RDWR
+      mode ^= ::File::Mode::Read | ::File::Mode::Write
+    end
```
RX14: I'd rather handle these flags in two stages (pseudocode):

```cr
mode |= File::Mode::Write if mode.append?

flags = 0
if mode.read? && mode.write?
  flags |= RDWR
elsif mode.read?
  flags |= RDONLY
elsif mode.write?
  flags |= WRONLY
else
  raise "Must specify one of ..."
end

flags |= O_CREAT if mode.create?
...etc
```

and there's no need for the mode.each loop, in fact the code is shorter (and more easily optimizable) without.

spec/std/file_spec.cr
```diff
@@ -224,6 +224,21 @@ describe "File" do
...
   end
 
+  describe "open with symnofollow" do
+    it "doesn't follow symlinks" do
+      file = datapath("test_file.txt")
+      other = datapath("test_file.ini")
+
+      with_tempfile("test_file_symlink.txt") do |symlink|
+        File.symlink(File.real_path(file), symlink)
+        File.new(symlink, File::Mode.flags(Read))
```
RX14: Could we have this test using the `:read` syntax just to make sure it works?

spec/std/file_spec.cr
```diff
@@ -224,6 +224,21 @@ describe "File" do
...
+        expect_raises(Errno) do
+          File.new(symlink, File::Mode.flags(Read, SymlinkNoFollow))
```
RX14: ditto but for the `*modes` overload.

src/crystal/system/file.cr
didactic-drunk: The loop is to prevent unhandled modes from slipping through by later additions.  Are you sure you want it removed?

spec/std/file_spec.cr
didactic-drunk: Small issue with splat args and extra params.  Is this a crystal bug or intended design?  Do I need to specify types for every parameter?

```
File.open(path, :write, :truncate, perm) do |file|
```

```
Error: no overload matches 'File.open' with types String, Symbol, Symbol, Int32

Overloads are:
 - File.open(filename : Path | String, mode = "r", perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil)
 - File.open(filename : Path | String, mode : Mode = File::Mode.flags(Read), perm : File::Permissions | Int = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil, &block)
 - File.open(filename : Path | String, mode : String, perm = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil, &block)
 - File.open(filename : Path | String, *modes : Mode, perm : File::Permissions | Int = DEFAULT_CREATE_PERMISSIONS, encoding = nil, invalid = nil, &block)
```

didactic-drunk: IF I don't specify a type for `mode` It passes the second `Symbol` as `encoding`.

didactic-drunk: Should I keep tests that generate deprecated warnings?  Like `File.new(..., "w")`?

RX14: Can you fix this by adding more type restrictions to the deprecated modes?

didactic-drunk: Should `File.new(..., "w")` be tested?  If so it generates a warning because it's deprecated.

src/crystal/system/file.cr
RX14: It's up to you, but I think the loop will be slower and specs will cover unhandled modes (who will add a mode without testing it and writing a spec??)

The handling for the read/write flags is much better though.

spec/std/file_spec.cr
RX14: @bcardiff what is the testing strategy for deprecated methods? Should the spec suite be run with warnings off?

didactic-drunk: @bcardiff ?

didactic-drunk: Is there anything I need to do?

bcardiff: Let the CI emit the warnings for now. This will push the story to allow warnings in specific call-sites or remind us to finally remove the deprecated methods.

RX14: Sorry, I have some commits and changes locally for this PR which I haven't published.

didactic-drunk: Bump.  You missed our 6 month anniversary (on this issue).  I'm breaking up with you.

RX14: I'm so sorry! I forget about things too often >_<

didactic-drunk:
How many years should I wait for a review?  The opportunity to get this in 1.0 has passed.

didactic-drunk:
I am not in favor of #9073.  It's confusing.  `Object.new` creates a new object in memory in most programming languages, not creates the resource it represents.

If moving to `File.open` please tell me which of the following should also happen:
```
Aws::S3.new # Creates a new s3 object - use .open to read an existing one.
Aws::EC2.new # Starts a new ec2 instance that you're billed for.
Mysql.new # Creates a new database server instance
Memcached.new # Starts a new memcached server.
...
```

@straight-shoota pointed out that this goes against all other uses of `.new`/`.open` where `.open` takes a block and closes the resource before returning.  `.new` means you manage it yourself.  When did that convention change?

didactic-drunk:
Specs pass, the modes are simplified based on RX14's suggestions, SYNC was removed.

What  else is required to get this merged?

asterite:
@didactic-drunk Is this change backwards-compatible? That is, can I still do `File.open("file.txt", "w")`?

didactic-drunk:
> @didactic-drunk Is this change backwards-compatible? That is, can I still do `File.open("file.txt", "w")`?

Yes.  Proof:
* All specs with string modes are unmodified.
* `open_flags(mode : String)` is utilized and unmodified.  `open_flags(mode: File::Mode)` handles the new symbol(s)/enum.
* `File.new/File.open` contain explicit `String` mode methods with deprecation warnings.

straight-shoota:
This change consists of serveral pieces:

* Add `File::Mode` flag enum
* Add handling of `File::Mode` in `Crystal::System::File`
* Add overloads for `File.new`/`.open` with `File::Mode`
* Refactor to use new `File.new`/`.open` overloads in stdlib and the compiler
* Refactor some uses of `File.open` to `File.new` ✖️ 
* Deprecate old overloads of `File.new`/`.open` with `String` mode ✖️ 

If possible, I'd like to separate some of these changes.

I'm not sure if we're already commited to deprecating the string-based file modes. It would be possible to defer that.
Several refactorings from `File.open` to `File.new` throughout stdlib could be extracted as separate changes. We can apply that independent of the other changes.
So I'd like to move the items marked with ✖️ out of the main PR. The others are tightly related and should work as a single change set.
WDYT?

It's probably best to start a completely new PR to get rid of the old history. Let's finalize the discussion on the API first, though.
I've created #11901 for that to keep the PR focused on implementation. The API looks good overall, but I'm proposing a few minor adjustments. Let's continue there!

didactic-drunk:
> * Refactor some uses of `File.open` to `File.new` heavy_multiplication_x
> 
> * Deprecate old overloads of `File.new`/`.open` with `String` mode heavy_multiplication_x
> 
> 
> If possible, I'd like to separate some of these changes.

Ok.

> I'm not sure if we're already commited to deprecating the string-based file modes. It would be possible to defer that. Several refactorings from `File.open` to `File.new` throughout stdlib could be extracted as separate changes. 

Ok.

> We can apply that independent of the other changes. So I'd like to move the items marked with heavy_multiplication_x out of the main PR. The others are tightly related and should work as a single change set. WDYT?

Ok.

> It's probably best to start a completely new PR to get rid of the old history.

Is that preferred to remove changes, rebase, force push?

straight-shoota:
Yes, I think starting a new PR is better. It means we keep the old review history attached here.
And start with a new review history and a new scope in a new PR.

Let's finalize the API in https://github.com/crystal-lang/crystal/issues/11901 first, though.

PR #8013, opened 2019-07-30
Handle too long URIs and too large header fields in HTTP::Request.from_io

### LINKED ISSUE #7838: HTTP::Server long url 400 Bad Request
## Summary
I found that HTTP::Server will appear 400 Bad Request for very long URLs.
Some requests require the use of "extra long characters".
Maybe it's because the request URL is longer than 4096 bytes (I guess).
[src/http/request.cr#L100](https://github.com/crystal-lang/crystal/blob/639e4765f3f4137f90c5b7da24d8ccb5b0bfec35/src/http/request.cr#L100) 

## Example
* curl
```text
curl -v "http://127.0.0.1:8080/list_detail_rate.htm/_____tmd_____/verify/?nc_token=33bffaa55be619b46359e6353d1bd6d9&nc_session_id=01VsUAH0NTPMVtfd7lyQMahoibWdyK1P8oUMd5M98WlcQ84bFi0PnmHETc3rGPtk3VN8HPWMx6QlMWV-Nw9LKhFln6-0AXmw2z9nEi93BnEET98VsjMF0mhWdmdvkUgS9eiw8IygcyecEQzuZucDqgaEY3jNRF8BNoe2JrjpADDmLoEzvs7GjkrYL_EbxDjGJoHvJG0DpdSHZOdyTDPcKc6A&nc_sig=059HQ0XsykTiYwN9Dn38G9XAcFHfXDvRtfFNlgtdIua129sbzmSlbWHFUzKuj3U0IGA3bzvprIaQ2IoO5j4l7TyEi806jDSf_kjePvkLjnFM0XA-ChnD7rcqrnGe1b_Mp-1YxplkweRUguUHMefTFifKoi1GowPa_H7SLr8CzbOCPNs6gWLiqxnwjLVLEgRlHUhzIiGJzKxLpS1LnjuS6F8jnEPs_3D0-lJiwaplLuVDbhA7fg4QFn39t4VWlR50gyAaalQwqubUnJlldueM0Dq6XPfM_6Ndy6_nQAW-3Asvt74RX7LO3w2WYnYgtfI6m2UL8jXC9Yiq4UBV1U7xZ-DVecZNUTkrHO1432KkP1TBgo2o3GOxbowBqfBsLfhdzrze9Jac4FCUrw_wieJX98p1sAxQHojeOF39MSL4bxBD7NQBcMbHWkrAmSui0NXBs_ie1EDml3wzWml9OAPM062sXW_TewgrvFacZBClGCSEA&x5secdata=5e0c8e1365474455070961b803bd560607b52cabf5960afff39b64ce58073f7844a8dd94c89edef01763e7bda2153e72fcab45d62d62d1d0fb9af411e98157f55151e037d30b894fa1b9f0617640df860a901a36bd6d324642091753f1253ec47fad7705c7bf26a7ef1aa109e380a620464f3a20661c6f7fc8f51770469f5a56dfb299a2f7461e469e75ebb7a26a42c122e1e84e4d46a143d8a56b702df58595e0496fd369fe90d5a5cb0968745f36293173959473a52df991db9a057aa860e6ec7493fb62f9278e24f1101ec2a88496be64fe7c2b8f69454c3030f98912b20e57c906a34cd3389933259d1cdc60837267dab14593dfc4341433f7b16aa8ec901a66fc14f736ed45bc53517cde99db61ef652477ccb4bf7caafe10e57362d226cd6a27906d66221653de01d0c8621b7f55c9edda4450e5b8708806aa671a4bb3bcfb0c3aefac537538fb3c4518b2f3cc9bebe3999de89b715520fa7bf25efa24321bd982834025777f5d720b283432993f456a22bf2ab1f7cadbabe78d5bcac146261ddfd78f42f556a153d04afeba97bfabe7e1b1bab5dc205ea1f1772a9e65ebfada88c577dfe8684438af0009bcfe91f53e8c9985011a1825bee7f3ef3adc7f6d15cb7c55489af26f039b67e10ebc3a2a53505010ede26797ee181e46d402671754d17dd5564f0e6611fe3655c144b68760cd919376ac3255c9fdbe366c7f7491c9324eb505b51100f3fd67251bb6e2a290d7c56b4a58098f06437f2e333ec3f6cf76e3ee75950ff412008f8c60927d405808a1ec5d84986c8b4bfe6c1e8a02d3f35015456c911ce421c3db22e7dd7e8d2edbf804b54b3801df2b2e32f800688d952d5e81b28316b4a8faf6e0540b8f4f9af6b1939a7d4c1814504c58f7372dfab298da2b638e7353318951b2757802182c5a4d949f7b7cfb0c1514ffde316a0778c2614fbf02ef5a29f033ba085d9500a8d98a373378e3ee3378f7ded7a3d74c77b43f11261b7b5f9e067a45acd7049b011eb1eb6e406071ebfa38086534305cddfd9db238b6528227ddc64a8214a4fa79e1b099381f6184d8b9c22d44361260404ad583b0233d028008edadc8580c9131e1410ec11e8297afd457bfd2817f80767c4d0c0a99304b99dfaf59d60322c79b3c37149d26b625db6229276943592642f84cd7a095cb5ad2399a28978f59d5e9d86b51cc6aeff759b1dae767c0c84d5a6408e232f59efbf383b923ad6738552238aae00e4ef5273f3f5b494ef2dd23f6c603fbb0c26db20e320529bc443d9b58aef1bc5836971b4f2acdf250b8eaf5df0c4800d019ccf2d4ded894b11504dc360b969530e713e02e5dbb6c5fb99b7729e541499ce3b3fbadf0bb88f623a6eee018a68a18b6f6f57292612caaebcfc792fa88dbf234633645d5865da5fe3bcf39c2600b79f3139dec529c808a88f9b065b25cf09c510dc5ddf56de47dd462e881c5061646c4b408768ad67cd2d8169abaf9c277889456442b3945a56eb91dd6836a1974366dd76c88c3bc1f35791e3cb48ef26868ec56dfa812950018110c97f8a6c1c39298c1f4ffcca95d95763f08473cd9a7dd538b15c3b99f3b8355dead71c32615ed937157fbf55d5b324abaee42404d6aff7d2caa016798b16d603567006c733c01a8f1728ec15bccc9d999a0c5d5271e68f3b53046e1b78c87ff1e0e343d02c851be03a191caddb580b1ea01eadb271272c50e4903455e4f80978e507b4d09db62dd59fd0134bb84f4cf72a713a47b125a9b090df20c3deed1c4679609bb800706191dec560a966dfd1bb36c01ab8433e9cedd47ca33aaca9305a8f190f356f78f9b9db2eefc41e0a0b402d48bb4e7125f25c0c632c5f2f2cccd5f5a4ffa813fe009cdc216dc960b657a675d487b0c0c172e97e5dc19a64974c66abd2203b4080fd9399cf1643d1c951dbc90e2eb1f6656a20d2c9eec9ae5af685031c4de9c02987273551a7383c54d7e16d6281bb4d02c1f6cff44d4c5f30d925a321a444351c29418abe135cf4f7954f786e32852fe5258f493bc9dcbfcdff230c31f036b565b1cdef31ecb62172ceaba403026211aec825c4bd8a8b1c71fd35be13938c426cb2fdd3a19f5f44c77d8675d6f5b846473a0e4247623faad34e271cced3d4ecb531650820913eafb414d95b956ff1a5e2d48105115312ee8ba9cd0045d9f1efb69a18b35502ac6bf035c52f679ead961a4c9b50c2a0bc7055232b5d3fe3bb5246562980d3e13c4f821158f623135bc32c61b49e558e226a11aef5d9a678964b11679bf2531504047e854&x5step=100&nc_app_key=X82Y__6eaa13982eaa135add1e2d6e36ba98f9"
```

* Server
```crystal
require "http/server"

end

server.listen
```

## Verbose
```text
$ curl -v "http://127.0.0.1:8080/list_detail_rate.htm/_____tmd_____/verify/?nc_token=33bffaa55be619b46359e6353d1bd6d9&nc_session_id=01VsUAH0NTPMVtfd7lyQMahoibWdyK1P8oUMd5M98WlcQ84bFi0PnmHETc3rGPtk3VN8HPWMx6QlMWV-Nw9LKhFln6-0AXmw2z9nEi93BnEET98VsjMF0mhWdmdvkUgS9eiw8IygcyecEQzuZucDqgaEY3jNRF8BNoe2JrjpADDmLoEzvs7GjkrYL_EbxDjGJoHvJG0DpdSHZOdyTDPcKc6A&nc_sig=059HQ0XsykTiYwN9Dn38G9XAcFHfXDvRtfFNlgtdIua129sbzmSlbWHFUzKuj3U0IGA3bzvprIaQ2IoO5j4l7TyEi806jDSf_kjePvkLjnFM0XA-ChnD7rcqrnGe1b_Mp-1YxplkweRUguUHMefTFifKoi1GowPa_H7SLr8CzbOCPNs6gWLiqxnwjLVLEgRlHUhzIiGJzKxLpS1LnjuS6F8jnEPs_3D0-lJiwaplLuVDbhA7fg4QFn39t4VWlR50gyAaalQwqubUnJlldueM0Dq6XPfM_6Ndy6_nQAW-3Asvt74RX7LO3w2WYnYgtfI6m2UL8jXC9Yiq4UBV1U7xZ-DVecZNUTkrHO1432KkP1TBgo2o3GOxbowBqfBsLfhdzrze9Jac4FCUrw_wieJX98p1sAxQHojeOF39MSL4bxBD7NQBcMbHWkrAmSui0NXBs_ie1EDml3wzWml9OAPM062sXW_TewgrvFacZBClGCSEA&x5secdata=5e0c8e1365474455070961b803bd560607b52cabf5960afff39b64ce58073f7844a8dd94c89edef01763e7bda2153e72fcab45d62d62d1d0fb9af411e98157f55151e037d30b894fa1b9f0617640df860a901a36bd6d324642091753f1253ec47fad7705c7bf26a7ef1aa109e380a620464f3a20661c6f7fc8f51770469f5a56dfb299a2f7461e469e75ebb7a26a42c122e1e84e4d46a143d8a56b702df58595e0496fd369fe90d5a5cb0968745f36293173959473a52df991db9a057aa860e6ec7493fb62f9278e24f1101ec2a88496be64fe7c2b8f69454c3030f98912b20e57c906a34cd3389933259d1cdc60837267dab14593dfc4341433f7b16aa8ec901a66fc14f736ed45bc53517cde99db61ef652477ccb4bf7caafe10e57362d226cd6a27906d66221653de01d0c8621b7f55c9edda4450e5b8708806aa671a4bb3bcfb0c3aefac537538fb3c4518b2f3cc9bebe3999de89b715520fa7bf25efa24321bd982834025777f5d720b283432993f456a22bf2ab1f7cadbabe78d5bcac146261ddfd78f42f556a153d04afeba97bfabe7e1b1bab5dc205ea1f1772a9e65ebfada88c577dfe8684438af0009bcfe91f53e8c9985011a1825bee7f3ef3adc7f6d15cb7c55489af26f039b67e10ebc3a2a53505010ede26797ee181e46d402671754d17dd5564f0e6611fe3655c144b68760cd919376ac3255c9fdbe366c7f7491c9324eb505b51100f3fd67251bb6e2a290d7c56b4a58098f06437f2e333ec3f6cf76e3ee75950ff412008f8c60927d405808a1ec5d84986c8b4bfe6c1e8a02d3f35015456c911ce421c3db22e7dd7e8d2edbf804b54b3801df2b2e32f800688d952d5e81b28316b4a8faf6e0540b8f4f9af6b1939a7d4c1814504c58f7372dfab298da2b638e7353318951b2757802182c5a4d949f7b7cfb0c1514ffde316a0778c2614fbf02ef5a29f033ba085d9500a8d98a373378e3ee3378f7ded7a3d74c77b43f11261b7b5f9e067a45acd7049b011eb1eb6e406071ebfa38086534305cddfd9db238b6528227ddc64a8214a4fa79e1b099381f6184d8b9c22d44361260404ad583b0233d028008edadc8580c9131e1410ec11e8297afd457bfd2817f80767c4d0c0a99304b99dfaf59d60322c79b3c37149d26b625db6229276943592642f84cd7a095cb5ad2399a28978f59d5e9d86b51cc6aeff759b1dae767c0c84d5a6408e232f59efbf383b923ad6738552238aae00e4ef5273f3f5b494ef2dd23f6c603fbb0c26db20e320529bc443d9b58aef1bc5836971b4f2acdf250b8eaf5df0c4800d019ccf2d4ded894b11504dc360b969530e713e02e5dbb6c5fb99b7729e541499ce3b3fbadf0bb88f623a6eee018a68a18b6f6f57292612caaebcfc792fa88dbf234633645d5865da5fe3bcf39c2600b79f3139dec529c808a88f9b065b25cf09c510dc5ddf56de47dd462e881c5061646c4b408768ad67cd2d8169abaf9c277889456442b3945a56eb91dd6836a1974366dd76c88c3bc1f35791e3cb48ef26868ec56dfa812950018110c97f8a6c1c39298c1f4ffcca95d95763f08473cd9a7dd538b15c3b99f3b8355dead71c32615ed937157fbf55d5b324abaee42404d6aff7d2caa016798b16d603567006c733c01a8f1728ec15bccc9d999a0c5d5271e68f3b53046e1b78c87ff1e0e343d02c851be03a191caddb580b1ea01eadb271272c50e4903455e4f80978e507b4d09db62dd59fd0134bb84f4cf72a713a47b125a9b090df20c3deed1c4679609bb800706191dec560a966dfd1bb36c01ab8433e9cedd47ca33aaca9305a8f190f356f78f9b9db2eefc41e0a0b402d48bb4e7125f25c0c632c5f2f2cccd5f5a4ffa813fe009cdc216dc960b657a675d487b0c0c172e97e5dc19a64974c66abd2203b4080fd9399cf1643d1c951dbc90e2eb1f6656a20d2c9eec9ae5af685031c4de9c02987273551a7383c54d7e16d6281bb4d02c1f6cff44d4c5f30d925a321a444351c29418abe135cf4f7954f786e32852fe5258f493bc9dcbfcdff230c31f036b565b1cdef31ecb62172ceaba403026211aec825c4bd8a8b1c71fd35be13938c426cb2fdd3a19f5f44c77d8675d6f5b846473a0e4247623faad34e271cced3d4ecb531650820913eafb414d95b956ff1a5e2d48105115312ee8ba9cd0045d9f1efb69a18b35502ac6bf035c52f679ead961a4c9b50c2a0bc7055232b5d3fe3bb5246562980d3e13c4f821158f623135bc32c61b49e558e226a11aef5d9a678964b11679bf2531504047e854&x5step=100&nc_app_key=X82Y__6eaa13982eaa135add1e2d6e36ba98f9"
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 1087 (#0)
> GET http://127.0.0.1:8080/list_detail_rate.htm/_____tmd_____/verify/?nc_token=33bffaa55be619b46359e6353d1bd6d9&nc_session_id=01VsUAH0NTPMVtfd7lyQMahoibWdyK1P8oUMd5M98WlcQ84bFi0PnmHETc3rGPtk3VN8HPWMx6QlMWV-Nw9LKhFln6-0AXmw2z9nEi93BnEET98VsjMF0mhWdmdvkUgS9eiw8IygcyecEQzuZucDqgaEY3jNRF8BNoe2JrjpADDmLoEzvs7GjkrYL_EbxDjGJoHvJG0DpdSHZOdyTDPcKc6A&nc_sig=059HQ0XsykTiYwN9Dn38G9XAcFHfXDvRtfFNlgtdIua129sbzmSlbWHFUzKuj3U0IGA3bzvprIaQ2IoO5j4l7TyEi806jDSf_kjePvkLjnFM0XA-ChnD7rcqrnGe1b_Mp-1YxplkweRUguUHMefTFifKoi1GowPa_H7SLr8CzbOCPNs6gWLiqxnwjLVLEgRlHUhzIiGJzKxLpS1LnjuS6F8jnEPs_3D0-lJiwaplLuVDbhA7fg4QFn39t4VWlR50gyAaalQwqubUnJlldueM0Dq6XPfM_6Ndy6_nQAW-3Asvt74RX7LO3w2WYnYgtfI6m2UL8jXC9Yiq4UBV1U7xZ-DVecZNUTkrHO1432KkP1TBgo2o3GOxbowBqfBsLfhdzrze9Jac4FCUrw_wieJX98p1sAxQHojeOF39MSL4bxBD7NQBcMbHWkrAmSui0NXBs_ie1EDml3wzWml9OAPM062sXW_TewgrvFacZBClGCSEA&x5secdata=5e0c8e1365474455070961b803bd560607b52cabf5960afff39b64ce58073f7844a8dd94c89edef01763e7bda2153e72fcab45d62d62d1d0fb9af411e98157f55151e037d30b894fa1b9f0617640df860a901a36bd6d324642091753f1253ec47fad7705c7bf26a7ef1aa109e380a620464f3a20661c6f7fc8f51770469f5a56dfb299a2f7461e469e75ebb7a26a42c122e1e84e4d46a143d8a56b702df58595e0496fd369fe90d5a5cb0968745f36293173959473a52df991db9a057aa860e6ec7493fb62f9278e24f1101ec2a88496be64fe7c2b8f69454c3030f98912b20e57c906a34cd3389933259d1cdc60837267dab14593dfc4341433f7b16aa8ec901a66fc14f736ed45bc53517cde99db61ef652477ccb4bf7caafe10e57362d226cd6a27906d66221653de01d0c8621b7f55c9edda4450e5b8708806aa671a4bb3bcfb0c3aefac537538fb3c4518b2f3cc9bebe3999de89b715520fa7bf25efa24321bd982834025777f5d720b283432993f456a22bf2ab1f7cadbabe78d5bcac146261ddfd78f42f556a153d04afeba97bfabe7e1b1bab5dc205ea1f1772a9e65ebfada88c577dfe8684438af0009bcfe91f53e8c9985011a1825bee7f3ef3adc7f6d15cb7c55489af26f039b67e10ebc3a2a53505010ede26797ee181e46d402671754d17dd5564f0e6611fe3655c144b68760cd919376ac3255c9fdbe366c7f7491c9324eb505b51100f3fd67251bb6e2a290d7c56b4a58098f06437f2e333ec3f6cf76e3ee75950ff412008f8c60927d405808a1ec5d84986c8b4bfe6c1e8a02d3f35015456c911ce421c3db22e7dd7e8d2edbf804b54b3801df2b2e32f800688d952d5e81b28316b4a8faf6e0540b8f4f9af6b1939a7d4c1814504c58f7372dfab298da2b638e7353318951b2757802182c5a4d949f7b7cfb0c1514ffde316a0778c2614fbf02ef5a29f033ba085d9500a8d98a373378e3ee3378f7ded7a3d74c77b43f11261b7b5f9e067a45acd7049b011eb1eb6e406071ebfa38086534305cddfd9db238b6528227ddc64a8214a4fa79e1b099381f6184d8b9c22d44361260404ad583b0233d028008edadc8580c9131e1410ec11e8297afd457bfd2817f80767c4d0c0a99304b99dfaf59d60322c79b3c37149d26b625db6229276943592642f84cd7a095cb5ad2399a28978f59d5e9d86b51cc6aeff759b1dae767c0c84d5a6408e232f59efbf383b923ad6738552238aae00e4ef5273f3f5b494ef2dd23f6c603fbb0c26db20e320529bc443d9b58aef1bc5836971b4f2acdf250b8eaf5df0c4800d019ccf2d4ded894b11504dc360b969530e713e02e5dbb6c5fb99b7729e541499ce3b3fbadf0bb88f623a6eee018a68a18b6f6f57292612caaebcfc792fa88dbf234633645d5865da5fe3bcf39c2600b79f3139dec529c808a88f9b065b25cf09c510dc5ddf56de47dd462e881c5061646c4b408768ad67cd2d8169abaf9c277889456442b3945a56eb91dd6836a1974366dd76c88c3bc1f35791e3cb48ef26868ec56dfa812950018110c97f8a6c1c39298c1f4ffcca95d95763f08473cd9a7dd538b15c3b99f3b8355dead71c32615ed937157fbf55d5b324abaee42404d6aff7d2caa016798b16d603567006c733c01a8f1728ec15bccc9d999a0c5d5271e68f3b53046e1b78c87ff1e0e343d02c851be03a191caddb580b1ea01eadb271272c50e4903455e4f80978e507b4d09db62dd59fd0134bb84f4cf72a713a47b125a9b090df20c3deed1c4679609bb800706191dec560a966dfd1bb36c01ab8433e9cedd47ca33aaca9305a8f190f356f78f9b9db2eefc41e0a0b402d48bb4e7125f25c0c632c5f2f2cccd5f5a4ffa813fe009cdc216dc960b657a675d487b0c0c172e97e5dc19a64974c66abd2203b4080fd9399cf1643d1c951dbc90e2eb1f6656a20d2c9eec9ae5af685031c4de9c02987273551a7383c54d7e16d6281bb4d02c1f6cff44d4c5f30d925a321a444351c29418abe135cf4f7954f786e32852fe5258f493bc9dcbfcdff230c31f036b565b1cdef31ecb62172ceaba403026211aec825c4bd8a8b1c71fd35be13938c426cb2fdd3a19f5f44c77d8675d6f5b846473a0e4247623faad34e271cced3d4ecb531650820913eafb414d95b956ff1a5e2d48105115312ee8ba9cd0045d9f1efb69a18b35502ac6bf035c52f679ead961a4c9b50c2a0bc7055232b5d3fe3bb5246562980d3e13c4f821158f623135bc32c61b49e558e226a11aef5d9a678964b11679bf2531504047e854&x5step=100&nc_app_key=X82Y__6eaa13982eaa135add1e2d6e36ba98f9 HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Proxy-Connection: Keep-Alive
> 
< HTTP/1.1 400 Bad Request
< Content-Type: text/plain
< Transfer-Encoding: chunked
< Proxy-Connection: keep-alive
< 
400 Bad Request
* Connection #0 to host 127.0.0.1 left intact
```

straight-shoota:
This change removes `HTTP::Request::BadRequest` which was just a status type to signal an error in the request parser. Instead, `Request.from_io` directly returns an `HTTP::Status`  which can designate a different status code than `400 Bad Request`.

When a URI is too long (max size is configurable, defaults to 1MB), it returns `414 Request-URI too long`.
When the header fields are to large (`HTTP::MAX_HEADER_SIZE` = 1MB), it returns `431 Request Header Fields Too Large`.

Fixes #7838

### DIFF
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -29,8 +29,8 @@ module HTTP
   def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true, *, max_headers_size : Int32 = MAX_HEADERS_SIZE) : HTTP::Status?
     headers = Headers.new
 
-    headers_size = 0
-    while header_line = read_header_line(io, max_headers_size)
+    max_size = max_headers_size
+    while header_line = read_header_line(io, max_size)
       case header_line
       when EndOfRequest
         body = nil
@@ -71,15 +71,15 @@ module HTTP
         yield headers, body
         return
       else # HeaderLine
-        headers_size += header_line.bytesize
-        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if headers_size > max_headers_size
+        max_size -= header_line.bytesize
+        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if max_size < 0
 
         return HTTP::Status::BAD_REQUEST unless headers.add?(header_line.name, header_line.value)
       end
     end
   end
 
-  private def self.read_header_line(io, max_headers_size) : HeaderLine | EndOfRequest | Nil
+  private def self.read_header_line(io, max_size) : HeaderLine | EndOfRequest | Nil
     # Optimization: check if we have a peek buffer
     if peek = io.peek
       # peek.empty? means EOF (so bad request)
@@ -101,14 +101,19 @@ module HTTP
           return EndOfRequest.new
         end
 
+        return HeaderLine.new name: "", value: "", bytesize: index + 1 if index > max_size
+
         name, value = parse_header(peek[0, end_index])
         io.skip(index + 1) # Must skip until after \n
         return HeaderLine.new name: name, value: value, bytesize: index + 1
       end
     end
 
-    line = io.gets(max_headers_size, chomp: true)
+    line = io.gets(max_size + 1, chomp: true)
     return nil unless line
+    if line.bytesize > max_size
+      return HeaderLine.new name: "", value: "", bytesize: max_size
+    end
 
     if line.empty?
       return EndOfRequest.new
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -5,8 +5,11 @@ require "mime/media_type"
 {% end %}
 
 module HTTP
-  # :nodoc:
-  MAX_HEADER_SIZE = 1_048_576 # 1 MB
+  # Default maximum permitted size (in bytes) of the request line in an HTTP request.
+  MAX_REQUEST_LINE_SIZE = 8192 # 8 KB
+
+  # Default maximum permitted combined size (in bytes) of the headers in an HTTP request.
+  MAX_HEADERS_SIZE = 16_384 # 16 KB
 
   # :nodoc:
   enum BodyType
@@ -23,11 +26,11 @@ module HTTP
   record HeaderLine, name : String, value : String, bytesize : Int32
 
   # :nodoc:
-  def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true) : HTTP::Status?
+  def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true, *, max_headers_size : Int32 = MAX_HEADERS_SIZE) : HTTP::Status?
     headers = Headers.new
 
     headers_size = 0
-    while header_line = read_header_line(io)
+    while header_line = read_header_line(io, max_headers_size)
       case header_line
       when EndOfRequest
         body = nil
@@ -69,14 +72,14 @@ module HTTP
         return
       else # HeaderLine
         headers_size += header_line.bytesize
-        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if headers_size > MAX_HEADER_SIZE
+        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if headers_size > max_headers_size
 
         return HTTP::Status::BAD_REQUEST unless headers.add?(header_line.name, header_line.value)
       end
     end
   end
 
-  private def self.read_header_line(io) : HeaderLine | EndOfRequest | Nil
+  private def self.read_header_line(io, max_headers_size) : HeaderLine | EndOfRequest | Nil
     # Optimization: check if we have a peek buffer
     if peek = io.peek
       # peek.empty? means EOF (so bad request)
@@ -104,7 +107,7 @@ module HTTP
       end
     end
 
-    line = io.gets(MAX_HEADER_SIZE, chomp: true)
+    line = io.gets(max_headers_size, chomp: true)
     return nil unless line
 
     if line.empty?
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -6,7 +6,7 @@ require "mime/media_type"
 
 module HTTP
   # :nodoc:
-  MAX_HEADER_SIZE = 16_384
+  MAX_HEADER_SIZE = 1_048_576 # 1 MB
 
   # :nodoc:
   enum BodyType
@@ -23,7 +23,7 @@ module HTTP
   record HeaderLine, name : String, value : String, bytesize : Int32
 
   # :nodoc:
-  def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true)
+  def self.parse_headers_and_body(io, body_type : BodyType = BodyType::OnDemand, decompress = true) : HTTP::Status?
     headers = Headers.new
 
     headers_size = 0
@@ -66,12 +66,12 @@ module HTTP
         check_content_type_charset(body, headers)
 
         yield headers, body
-        break
+        return
       else # HeaderLine
         headers_size += header_line.bytesize
-        break if headers_size > MAX_HEADER_SIZE
+        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if headers_size > MAX_HEADER_SIZE
 
-        break unless headers.add?(header_line.name, header_line.value)
+        return HTTP::Status::BAD_REQUEST unless headers.add?(header_line.name, header_line.value)
       end
     end
   end
--- a/src/http/content.cr
+++ b/src/http/content.cr
@@ -192,7 +192,7 @@ module HTTP
     end
 
     private def read_chunk_size
-      line = @io.read_line(HTTP::MAX_HEADER_SIZE, chomp: true)
+      line = @io.read_line(HTTP::MAX_HEADERS_SIZE, chomp: true)
 
       if index = line.byte_index(';'.ord)
         chunk_size = line.byte_slice(0, index)
@@ -205,7 +205,7 @@ module HTTP
 
     private def read_trailer
       while true
-        line = @io.read_line(HTTP::MAX_HEADER_SIZE, chomp: true)
+        line = @io.read_line(HTTP::MAX_HEADERS_SIZE, chomp: true)
         break if line.empty?
 
         key, value = HTTP.parse_header(line)
--- a/src/http/content.cr
+++ b/src/http/content.cr
@@ -97,7 +97,17 @@ module HTTP
     # in the chunked trailer part (see [RFC 7230 section 4.1.2](https://tools.ietf.org/html/rfc7230#section-4.1.2)).
     getter headers : HTTP::Headers { HTTP::Headers.new }
 
-    def initialize(@io : IO)
+    # Returns the maximum permitted combined size for the trailing headers.
+    #
+    # When parsing the trailing headers `ChunkedContent` keeps track of the
+    # amount of total bytes consumed for all headers (including line breaks).
+    # If the combined byte size of all headers is larger than the permitted size,
+    # `IO::Error` is raised.
+    #
+    # Default: `HTTP::MAX_HEADERS_SIZE`
+    getter max_headers_size : Int32
+
+    def initialize(@io : IO, *, @max_headers_size : Int32 = HTTP::MAX_HEADERS_SIZE)
       @chunk_remaining = -1
       @received_final_chunk = false
     end
@@ -192,7 +202,7 @@ module HTTP
     end
 
     private def read_chunk_size
-      line = @io.read_line(HTTP::MAX_HEADERS_SIZE, chomp: true)
+      line = @io.read_line(@max_headers_size, chomp: true)
 
       if index = line.byte_index(';'.ord)
         chunk_size = line.byte_slice(0, index)
@@ -204,9 +214,15 @@ module HTTP
     end
 
     private def read_trailer
+      max_size = @max_headers_size
+
       while true
-        line = @io.read_line(HTTP::MAX_HEADERS_SIZE, chomp: true)
+        line = @io.read_line(max_size + 1, chomp: true)
         break if line.empty?
+        max_size -= line.bytesize
+        if max_size < 0
+          raise IO::Error.new("Trailing headers too long")
+        end
 
         key, value = HTTP.parse_header(line)
         break unless headers.add?(key, value)
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -101,11 +101,11 @@ class HTTP::Request
 
   # Returns a `HTTP::Request` instance if successfully parsed,
   # `nil` on EOF or `HTTP::Status` otherwise.
-  def self.from_io(io, *, max_request_line_size : Int32 = 1_048_576) : HTTP::Request | HTTP::Status | Nil
+  def self.from_io(io, *, max_request_line_size : Int32 = HTTP::MAX_REQUEST_LINE_SIZE, max_headers_size : Int32 = HTTP::MAX_HEADERS_SIZE) : HTTP::Request | HTTP::Status | Nil
     line = parse_request_line(io, max_request_line_size)
     return line unless line.is_a?(RequestLine)
 
-    status = HTTP.parse_headers_and_body(io) do |headers, body|
+    status = HTTP.parse_headers_and_body(io, max_headers_size: max_headers_size) do |headers, body|
       # No need to dup headers since nobody else holds them
       request = new line.method, line.resource, headers, body, line.http_version, internal: nil
 
@@ -122,7 +122,7 @@ class HTTP::Request
 
   private METHODS = %w(GET HEAD POST PUT DELETE CONNECT OPTIONS PATCH TRACE)
 
-  private def self.parse_request_line(io : IO, max_size : Int32) : RequestLine | HTTP::Status | Nil
+  private def self.parse_request_line(io : IO, max_request_line_size) : RequestLine | HTTP::Status | Nil
     # Optimization: see if we have a peek buffer
     # (avoids a string allocation for the entire request line)
     if peek = io.peek
@@ -132,7 +132,7 @@ class HTTP::Request
       # See if we can find \n
       index = peek.index('\n'.ord.to_u8)
       if index
-        return HTTP::Status::URI_TOO_LONG if index > max_size
+        return HTTP::Status::URI_TOO_LONG if index > max_request_line_size
 
         end_index = index
 
@@ -147,11 +147,11 @@ class HTTP::Request
       end
     end
 
-    request_line = io.gets(max_size + 1, chomp: true)
+    request_line = io.gets(max_request_line_size + 1, chomp: true)
     return nil unless request_line
 
     # Identify Request-URI too long
-    if request_line.bytesize > max_size
+    if request_line.bytesize > max_request_line_size
       return HTTP::Status::URI_TOO_LONG
     end
 
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -101,8 +101,8 @@ class HTTP::Request
 
   # Returns a `HTTP::Request` instance if successfully parsed,
   # `nil` on EOF or `HTTP::Status` otherwise.
-  def self.from_io(io) : HTTP::Request | HTTP::Status | Nil
-    line = parse_request_line(io)
+  def self.from_io(io, *, max_request_line_size : Int32 = 1_048_576) : HTTP::Request | HTTP::Status | Nil
+    line = parse_request_line(io, max_request_line_size)
     return line unless line.is_a?(RequestLine)
 
     HTTP.parse_headers_and_body(io) do |headers, body|
@@ -122,7 +122,7 @@ class HTTP::Request
 
   private METHODS = %w(GET HEAD POST PUT DELETE CONNECT OPTIONS PATCH TRACE)
 
-  private def self.parse_request_line(io : IO) : RequestLine | HTTP::Status | Nil
+  private def self.parse_request_line(io : IO, max_size : Int32) : RequestLine | HTTP::Status | Nil
     # Optimization: see if we have a peek buffer
     # (avoids a string allocation for the entire request line)
     if peek = io.peek
@@ -131,8 +131,9 @@ class HTTP::Request
 
       # See if we can find \n
       index = peek.index('\n'.ord.to_u8)
-      # TODO: eventually increase the 4096 bytes limit
-      if index && index < 4096
+      if index
+        return HTTP::Status::URI_TOO_LONG if index > max_size
+
         end_index = index
 
         # Also check (and discard) \r before that
@@ -146,8 +147,13 @@ class HTTP::Request
       end
     end
 
-    request_line = io.gets(4096, chomp: true)
+    request_line = io.gets(max_size + 1, chomp: true)
     return nil unless request_line
+
+    # Identify Request-URI too long
+    if request_line.bytesize > max_size
+      return HTTP::Status::URI_TOO_LONG
+    end
 
     parse_request_line(request_line)
   end
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -105,7 +105,7 @@ class HTTP::Request
     line = parse_request_line(io, max_request_line_size)
     return line unless line.is_a?(RequestLine)
 
-    HTTP.parse_headers_and_body(io) do |headers, body|
+    status = HTTP.parse_headers_and_body(io) do |headers, body|
       # No need to dup headers since nobody else holds them
       request = new line.method, line.resource, headers, body, line.http_version, internal: nil
 
@@ -117,7 +117,7 @@ class HTTP::Request
     end
 
     # Malformed or unexpectedly ended http request
-    HTTP::Status::BAD_REQUEST
+    status || HTTP::Status::BAD_REQUEST
   end
 
   private METHODS = %w(GET HEAD POST PUT DELETE CONNECT OPTIONS PATCH TRACE)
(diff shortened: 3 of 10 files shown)


RX14:
1MiB is way too large.

asterite:
Go does the same, so it might be fine.

RX14:
No, we should make the limits configurable and leave the defaults as-is. Maybe doubled, but no larger.

It's so easy to DoS a server, and the current values work for any *sane* usecase.

If someone's already designed an insane API which you can't break, you should be able to raise the limits. But the defaults should be left as-is.

asterite:
So, if you hit `www.google.com` with such large request it works fine.

However, it might make sense to have a compromise and choose a smallish value but let the user configure it if they really need it.

straight-shoota:
I think it makes sense to look at what default values others are using.

Request line:

* [apache](http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestline): 8K
* [nginx](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers): 8K
* [golang](https://golang.org/pkg/net/http/#pkg-constants): 1M (total)

Request Header:
* [apache](http://httpd.apache.org/docs/2.2/mod/core.html#limitrequestfieldsize): 8K
* [nginx](http://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers): 4 * 8K (8K per header field, 32K in total with request line)
* [golang](https://golang.org/pkg/net/http/#pkg-constants): 1M (total)

This list can be expanded.

asterite:
I forgot to mention this in my previous review: `HTTP::Server` should have a configurable property that is passed to `HTTP::Request.from_io`, so that one can choose the maximum size.

straight-shoota:
@asterite Do you mean max header size or max request line size? Or both?

asterite:
Probably both.

RX14:
Going with the same as NGINX is OK with me. I'd be fine with copying just the 32k without the 8k per-header too.

straight-shoota:
> Realistically, most everything running crystal is gonna pass through NGINX on the way.

In this case Crystal wouldn't need to enforce any limits at all. In fact, it might even be counterproductive because increasing the limits in NGINX wouldn't have any effect unless they're increased in the app server as well.

I think that's why golang has such high limits: they expect their HTTP server to be exposed behind a reverse proxy. For this use case, you want high limits in order to accept everything coming from the frontend.
Only when the app server is exposed directly, it needs to enforce reasonable limits by itself to reduce DoS attack surface.

From that line of thought, you could say that an app server supposed to be exposed without a reverse proxy needs to take special precautions anyway. One of them would be to decrease the default limits.
I'm not sure which limits Caddy uses. The documentation says it's configurable (even depending on path) but not what they use as default value.

Given that Crystal and Go are used for writing the same kinds of software, I'd be inclined to follow Go's example here and use very high limits.
It's not safe by default, but honestly, you probably shouldn't hook up your `HTTP::Server` directly to the internet anyway. That's what reverse proxies are for.

RX14:
I don't follow that we should support DoSing the server by default because we assume that we're behind NGINX. While we're behind NGINX most of the time right now, I want to make sure that Crystal is robust with good defaults for graduating *beyond* that!

If we cannot agree to *new limits*, then perhaps we can all agree to make this configurable, then change limits later.

straight-shoota:
> I want to make sure that Crystal is robust with good defaults for graduating beyond that!

The question is whether the stdlib can and should provide that.

But disussing limits and long-term goals is not the purpose of this PR. So I agree to leave this part of the discussion out.

straight-shoota:
I've added these limits as class properties of `HTTP` with the defaults:
* `max_request_line_size`: 8 KB
* `max_headers_size`: 16 KB (as before)

These properties are currently the only way to configure this, making them global values. I'm not sure whether we would need more fine-grained control, for example per `HTTP::Server` instance. I'd say this could be easily added later, should there be a demand.

I've also changed `read_header_line` to read only as much bytes that are available, otherwise a purposly desgined request could make it initially read up to `2 * max_header_size - 1` bytes.

src/http/common.cr
```diff
@@ -5,8 +5,11 @@ require "mime/media_type"
 {% end %}
 
 module HTTP
-  # :nodoc:
-  MAX_HEADER_SIZE = 16_384
+  # Maximum permitted size of the request line in an HTTP request.
```
Sija:
```suggestion
  # Maximum permitted size (in bytes) of the request line in an HTTP request.
```

src/http/common.cr
```diff
@@ -5,8 +5,11 @@ require "mime/media_type"
...
+  class_property max_request_line_size = 8192 # 8 KB
+
+  # Maximum permitted combined size of the headers in an HTTP request.
```
Sija:
```suggestion
  # Maximum permitted combined size (in bytes) of the headers in an HTTP request.
```

src/http/common.cr
```diff
@@ -66,17 +69,17 @@ module HTTP
         check_content_type_charset(body, headers)
 
         yield headers, body
-        break
+        return
       else # HeaderLine
-        headers_size += header_line.bytesize
-        break if headers_size > MAX_HEADER_SIZE
+        max_size -= header_line.bytesize
+        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if max_size <= 0
```
Sija:
```suggestion
        return HTTP::Status::REQUEST_HEADER_FIELDS_TOO_LARGE if max_size < 0
```

src/http/request.cr
```diff
@@ -148,21 +147,26 @@ class HTTP::Request
       end
     end
 
-    request_line = io.gets(4096, chomp: true)
+    request_line = io.gets(HTTP.max_request_line_size + 1, chomp: true)
     return nil unless request_line
 
+    # Identify Request-URI too long
+    if request_line.bytesize > HTTP.max_request_line_size
```
Sija: I'd assign `HTTP.max_request_line_size` to a variable, in order to save couple of calls here.

```suggestion
    if request_line.bytesize > HTTP.max_request_line_size
```

asterite:
> I've added these limits as class properties of HTTP with the defaults

Thank you! But I was thinking about this as properties of an `HTTP::Server`. What do you think?

I know an app will usually have just one http server, but I usually try to avoid global variables. One reason is that it's easier to test (no need to remember unsetting the property after each test).

RX14:
I'd definitely find it far cleaner to have them on `HTTP::Server` - even if the code for passing it around is a bit messy.

straight-shoota:
I've moved configuration to `HTTP::Server`.

src/http/content.cr
```diff
@@ -192,7 +192,7 @@ module HTTP
     end
 
     private def read_chunk_size
-      line = @io.read_line(HTTP::MAX_HEADER_SIZE, chomp: true)
+      line = @io.read_line(HTTP::MAX_HEADERS_SIZE, chomp: true)
```
asterite: Do you think we can make `HTTP::Content` hold a `@max_headers_size` and set it in the constructor? Then it will take the same value configured in `HTTP::Server`.

straight-shoota: Sure, that should work :+1:

straight-shoota: Only `HTTP::ChunkedContent` needs this.

asterite: Ah, yes, sorry, I meant to say `HTTP::ChunkedContent`

src/http/common.cr
```diff
@@ -5,8 +5,11 @@ require "mime/media_type"
 {% end %}
 
 module HTTP
-  # :nodoc:
-  MAX_HEADER_SIZE = 16_384
+  # Maximum permitted size (in bytes) of the request line in an HTTP request.
```
RX14: These are only defaults now - perhaps mention that?

src/http/server/request_processor.cr
```diff
@@ -19,13 +25,17 @@ class HTTP::Server::RequestProcessor
...
+          max_request_line_size: max_request_line_size,
+          max_headers_size: max_headers_size,
+        )
 
         # EOF
         break unless request
 
-        if request.is_a?(HTTP::Request::BadRequest)
-          response.respond_with_error("Bad Request", 400)
+        if request.is_a?(HTTP::Status)
```
Sija:
```suggestion
        if status = request.as?(HTTP::Status)
```

straight-shoota: I've tried that, unfortunately it doesn't work because it doesn't restrict the type of `request` to `HTTP::Request`.
I figure this should probably work, though because `HTTP::Status` runs into the branch and returns.

PR #8017, opened 2019-07-30
Hash now uses an open addressing algorithm

### LINKED ISSUE #4557: Reimplementation a Hash using open addressing
Crystal Hash(K, V) should be reimplemented using open addressing and powers of two.

To be read: 

* http://www.idryman.org/blog/2017/05/03/writing-a-damn-fast-hash-table-with-tiny-memory-footprints/.
* https://github.com/konovod/testhash
* https://github.com/yxhuvud/crystal-hash
* https://developers.redhat.com/blog/2017/02/27/towards-faster-ruby-hash-tables/
* (sometimes 508) http://blog.redpanthers.co/behind-scenes-hash-table-performance-ruby-2-4

Thanks to @RX14, @konovod, @yxhuvud for links.

Refs #2817, #3932, #2331 and https://github.com/crystal-lang/crystal/pull/4033#issuecomment-307779149 to continue discussion and implementation.

asterite:
This improves its performance, both in time (always) and memory (generally).

Fixes #4557

## The algorithm

The algorithm is basically the one that [Ruby](https://github.com/ruby/ruby/blob/master/st.c#L1-L101
) uses.

## How to review this

There's just a single commit. My recommendation is to first look at the final code, which is thoroughly documented (the important bits are at the top), then at the diff.

I tried to document this really well because otherwise in a month I won't remember any of this. It also enables anyone to understand how it works and possibly keep optimizing things.

## Why?

The old Hash implementation uses closed addressing: buckets with linked lists. This apparently isn't great because of pointer chasing and not having good cache locality. Using open addressing removes pointer chasing and improves cache locality. It could just be a theory but empirically it performs much better.

## Give me some benchmarks!

I used this code as a benchmark:

<details>
 <summary>Code for benchmarking old Hash vs. new Hash</summary>

```crystal
require "benchmark"
require "./old_hash"
require "random/secure"

def benchmark_create_empty
  Benchmark.ips do |x|
    x.report("old, create") do
      OldHash(Int32, Int32).new
    end
    x.report("new, create") do
      Hash(String, String).new
    end
  end
end

def benchmark_insert_strings(sizes)
  sizes.each do |size|
    values = Array.new(size) { Random::Secure.hex }

    Benchmark.ips do |x|
      x.report("old, insert (type: String, size: #{size})") do
        hash = OldHash(String, String).new
        values.each do |value|
          hash[value] = value
        end
      end
      x.report("new, insert (type: String, size: #{size})") do
        hash = Hash(String, String).new
        values.each do |value|
          hash[value] = value
        end
      end
    end
  end
end

def benchmark_insert_ints(sizes)
  sizes.each do |size|
    values = Array.new(size) { rand(Int32) }

    Benchmark.ips do |x|
      x.report("old, insert (type: Int32, size: #{size})") do
        hash = OldHash(Int32, Int32).new
        values.each do |value|
          hash[value] = value
        end
      end
      x.report("new, insert (type: Int32, size: #{size})") do
        hash = Hash(Int32, Int32).new
        values.each do |value|
          hash[value] = value
        end
      end
    end
  end
end

def benchmark_read_strings(sizes)
  sizes.each do |size|
    values = Array.new(size) { Random::Secure.hex }

    old_hash = OldHash(String, String).new
    new_hash = Hash(String, String).new

    values.each do |value|
      old_hash[value] = value
      new_hash[value] = value
    end

    Benchmark.ips do |x|
      x.report("old, read (type: String, size: #{size})") do
        values.each do |value|
          old_hash[value]
        end
      end
      x.report("new, read (type: String, size: #{size})") do
        values.each do |value|
          new_hash[value]
        end
      end
    end
  end
end

def benchmark_read_ints(sizes)
  sizes.each do |size|
    values = Array.new(size) { rand(1_00_000) }

    old_hash = OldHash(Int32, Int32).new
    new_hash = Hash(Int32, Int32).new

    values.each do |value|
      old_hash[value] = value
      new_hash[value] = value
    end

    Benchmark.ips do |x|
      x.report("old, read (type: Int32, size: #{size})") do
        values.each do |value|
          old_hash[value]
        end
      end
      x.report("new, read (type: Int32, size: #{size})") do
        values.each do |value|
          new_hash[value]
        end
      end
    end
  end
end

sizes = [5, 10, 15, 20, 30, 50, 100, 200, 500, 1_000, 10_000, 100_000]

benchmark_create_empty()
puts
benchmark_insert_strings(sizes)
puts
benchmark_insert_ints(sizes)
puts
benchmark_read_strings(sizes)
puts
benchmark_read_ints(sizes)
```
</details>

Results:

```
old, create  18.50M ( 54.06ns) (± 1.82%)   160B/op   2.34× slower
new, create  43.32M ( 23.08ns) (± 0.88%)  64.0B/op        fastest

old, insert (type: String, size: 5)   3.60M (277.58ns) (± 7.01%)  480B/op   1.38× slower
new, insert (type: String, size: 5)   4.97M (201.28ns) (± 4.77%)  384B/op        fastest
old, insert (type: String, size: 10)   1.95M (513.74ns) (± 1.64%)  801B/op   1.33× slower
new, insert (type: String, size: 10)   2.59M (386.02ns) (± 3.01%)  832B/op        fastest
old, insert (type: String, size: 15)   1.27M (786.17ns) (± 3.56%)  1.09kB/op   1.55× slower
new, insert (type: String, size: 15)   1.98M (505.71ns) (± 4.48%)    832B/op        fastest
old, insert (type: String, size: 20) 922.03k (  1.08µs) (± 0.99%)  1.41kB/op   1.37× slower
new, insert (type: String, size: 20)   1.27M (788.95ns) (± 2.86%)  1.67kB/op        fastest
old, insert (type: String, size: 30) 639.38k (  1.56µs) (± 5.63%)  2.03kB/op   1.58× slower
new, insert (type: String, size: 30)   1.01M (992.38ns) (± 6.01%)  1.67kB/op        fastest
old, insert (type: String, size: 50) 355.49k (  2.81µs) (± 5.77%)  3.28kB/op   1.49× slower
new, insert (type: String, size: 50) 530.85k (  1.88µs) (± 3.95%)  3.81kB/op        fastest
old, insert (type: String, size: 100) 154.93k (  6.45µs) (± 4.04%)  7.06kB/op   1.81× slower
new, insert (type: String, size: 100) 280.47k (  3.57µs) (± 3.92%)  7.09kB/op        fastest
old, insert (type: String, size: 200)  71.71k ( 13.94µs) (± 3.27%)  13.3kB/op   1.91× slower
new, insert (type: String, size: 200) 136.71k (  7.31µs) (± 2.21%)  14.5kB/op        fastest
old, insert (type: String, size: 500)  23.46k ( 42.62µs) (± 3.53%)  36.1kB/op   2.46× slower
new, insert (type: String, size: 500)  57.82k ( 17.30µs) (± 4.78%)  28.4kB/op        fastest
old, insert (type: String, size: 1000)  12.78k ( 78.27µs) (± 1.72%)  67.4kB/op   2.20× slower
new, insert (type: String, size: 1000)  28.12k ( 35.56µs) (± 4.20%)  56.4kB/op        fastest
old, insert (type: String, size: 10000)   1.00k (997.78µs) (± 6.41%)  662kB/op   2.12× slower
new, insert (type: String, size: 10000)   2.13k (469.91µs) (± 5.39%)  896kB/op        fastest

old, insert (type: Int32, size: 5)   4.83M (207.19ns) (± 0.80%)  400B/op   1.78× slower
new, insert (type: Int32, size: 5)   8.60M (116.23ns) (± 1.76%)  240B/op        fastest
old, insert (type: Int32, size: 10)   2.78M (359.30ns) (± 0.91%)  640B/op   1.72× slower
new, insert (type: Int32, size: 10)   4.79M (208.73ns) (± 1.11%)  448B/op        fastest
old, insert (type: Int32, size: 15)   1.90M (525.28ns) (± 2.96%)  880B/op   1.75× slower
new, insert (type: Int32, size: 15)   3.33M (300.60ns) (± 5.98%)  448B/op        fastest
old, insert (type: Int32, size: 20)   1.30M (769.02ns) (± 6.52%)  1.09kB/op   1.54× slower
new, insert (type: Int32, size: 20)   2.00M (500.54ns) (± 4.76%)    976B/op        fastest
old, insert (type: Int32, size: 30) 972.78k (  1.03µs) (± 4.94%)  1.56kB/op   1.72× slower
new, insert (type: Int32, size: 30)   1.67M (597.26ns) (± 4.29%)    976B/op        fastest
old, insert (type: Int32, size: 50) 582.06k (  1.72µs) (± 5.23%)   2.5kB/op   1.50× slower
new, insert (type: Int32, size: 50) 872.34k (  1.15µs) (± 7.97%)  1.88kB/op        fastest
old, insert (type: Int32, size: 100) 234.48k (  4.26µs) (± 7.09%)   5.5kB/op   2.08× slower
new, insert (type: Int32, size: 100) 488.20k (  2.05µs) (± 6.79%)  4.14kB/op        fastest
old, insert (type: Int32, size: 200) 126.98k (  7.88µs) (± 4.74%)  10.2kB/op   1.84× slower
new, insert (type: Int32, size: 200) 233.78k (  4.28µs) (± 7.45%)  8.47kB/op        fastest
old, insert (type: Int32, size: 500)  40.86k ( 24.47µs) (± 4.14%)  28.3kB/op   2.55× slower
new, insert (type: Int32, size: 500) 104.34k (  9.58µs) (± 7.55%)  16.5kB/op        fastest
old, insert (type: Int32, size: 1000)  17.87k ( 55.97µs) (± 3.24%)  51.8kB/op   2.57× slower
new, insert (type: Int32, size: 1000)  45.98k ( 21.75µs) (± 7.72%)  32.5kB/op        fastest
old, insert (type: Int32, size: 10000)   1.52k (658.91µs) (± 3.88%)  506kB/op   2.16× slower
new, insert (type: Int32, size: 10000)   3.28k (305.04µs) (± 4.09%)  513kB/op        fastest

old, read (type: String, size: 5)  11.64M ( 85.90ns) (± 6.51%)  0.0B/op   2.25× slower
new, read (type: String, size: 5)  26.20M ( 38.16ns) (± 2.98%)  0.0B/op        fastest
old, read (type: String, size: 10)   5.69M (175.80ns) (± 4.64%)  0.0B/op   1.39× slower
new, read (type: String, size: 10)   7.93M (126.07ns) (± 6.97%)  0.0B/op        fastest
old, read (type: String, size: 15)   4.15M (240.69ns) (± 2.01%)  0.0B/op   1.16× slower
new, read (type: String, size: 15)   4.82M (207.47ns) (± 3.30%)  0.0B/op        fastest
old, read (type: String, size: 20)   2.93M (341.71ns) (± 1.47%)  0.0B/op   1.52× slower
new, read (type: String, size: 20)   4.46M (224.21ns) (± 1.25%)  0.0B/op        fastest
old, read (type: String, size: 30)   1.87M (534.24ns) (± 2.90%)  0.0B/op   1.45× slower
new, read (type: String, size: 30)   2.71M (369.30ns) (± 1.90%)  0.0B/op        fastest
old, read (type: String, size: 50) 992.08k (  1.01µs) (± 4.51%)  0.0B/op   1.71× slower
new, read (type: String, size: 50)   1.70M (589.77ns) (± 1.65%)  0.0B/op        fastest
old, read (type: String, size: 100) 596.10k (  1.68µs) (± 1.60%)  0.0B/op   1.40× slower
new, read (type: String, size: 100) 832.91k (  1.20µs) (± 1.82%)  0.0B/op        fastest
old, read (type: String, size: 200) 260.30k (  3.84µs) (± 1.63%)  0.0B/op   1.53× slower
new, read (type: String, size: 200) 398.84k (  2.51µs) (± 1.84%)  0.0B/op        fastest
old, read (type: String, size: 500) 111.03k (  9.01µs) (± 2.89%)  0.0B/op   1.40× slower
new, read (type: String, size: 500) 155.71k (  6.42µs) (± 3.33%)  0.0B/op        fastest
old, read (type: String, size: 1000)  41.59k ( 24.04µs) (± 3.07%)  0.0B/op   1.71× slower
new, read (type: String, size: 1000)  71.27k ( 14.03µs) (± 3.81%)  0.0B/op        fastest
old, read (type: String, size: 10000)   2.66k (376.32µs) (± 4.55%)  0.0B/op   2.46× slower
new, read (type: String, size: 10000)   6.54k (152.95µs) (± 4.03%)  0.0B/op        fastest

old, read (type: Int32, size: 5)  23.67M ( 42.24ns) (± 1.64%)  0.0B/op   2.30× slower
new, read (type: Int32, size: 5)  54.35M ( 18.40ns) (± 3.14%)  0.0B/op        fastest
old, read (type: Int32, size: 10)  11.66M ( 85.77ns) (± 4.63%)  0.0B/op   1.85× slower
new, read (type: Int32, size: 10)  21.58M ( 46.33ns) (± 3.20%)  0.0B/op        fastest
old, read (type: Int32, size: 15)   7.81M (128.00ns) (± 4.15%)  0.0B/op   1.49× slower
new, read (type: Int32, size: 15)  11.63M ( 86.00ns) (± 3.99%)  0.0B/op        fastest
old, read (type: Int32, size: 20)   5.80M (172.48ns) (± 5.34%)  0.0B/op   1.73× slower
new, read (type: Int32, size: 20)  10.02M ( 99.77ns) (± 4.54%)  0.0B/op        fastest
old, read (type: Int32, size: 30)   3.67M (272.51ns) (± 2.99%)  0.0B/op   1.84× slower
new, read (type: Int32, size: 30)   6.74M (148.30ns) (± 1.97%)  0.0B/op        fastest
old, read (type: Int32, size: 50)   2.05M (488.04ns) (± 5.57%)  0.0B/op   1.89× slower
new, read (type: Int32, size: 50)   3.87M (258.41ns) (± 4.24%)  0.0B/op        fastest
old, read (type: Int32, size: 100)   1.09M (921.59ns) (± 8.72%)  0.0B/op   1.70× slower
new, read (type: Int32, size: 100)   1.84M (542.80ns) (± 5.52%)  0.0B/op        fastest
old, read (type: Int32, size: 200) 535.83k (  1.87µs) (± 5.84%)  0.0B/op   1.66× slower
new, read (type: Int32, size: 200) 891.31k (  1.12µs) (± 5.49%)  0.0B/op        fastest
old, read (type: Int32, size: 500) 236.68k (  4.23µs) (± 3.61%)  0.0B/op   1.52× slower
new, read (type: Int32, size: 500) 360.85k (  2.77µs) (± 4.31%)  0.0B/op        fastest
old, read (type: Int32, size: 1000) 106.13k (  9.42µs) (± 4.88%)  0.0B/op   1.66× slower
new, read (type: Int32, size: 1000) 175.92k (  5.68µs) (± 4.08%)  0.0B/op        fastest
old, read (type: Int32, size: 10000)   4.60k (217.62µs) (± 2.94%)  0.0B/op   3.00× slower
new, read (type: Int32, size: 10000)  13.80k ( 72.47µs) (± 1.67%)  0.0B/op        fastest
```

As you can see the new implementation is always faster than the old one. Sometimes more memory is used, sometimes less.

I also ran some @kostya [benchmarks](https://github.com/kostya/benchmarks) that used Hash in their implementation. Here are the results:

#### [Havlak](https://github.com/kostya/benchmarks/blob/master/havlak/havlak.cr):
```
old: 12.49s, 375.1Mb
new: 7.58s, 215.7Mb
```

Havlak seems to be a benchmark measuring how well a language performs in general algorithmic tasks... the new results look good! 😊

#### [Brainfuck](https://github.com/kostya/benchmarks/blob/master/brainfuck/brainfuck.cr):
```
old: 5.20s, 1.8Mb
new: 4.22s, 1.8Mb
```

#### [JSON](https://github.com/kostya/benchmarks/blob/master/json/test.cr) (when using `JSON.parse`):
```
old: 2.20s, 1137.0Mb
new: 2.07s, 961.3Mb
```

#### [Knucleotide](https://github.com/kostya/crystal-benchmarks-game/blob/master/knucleotide/knucleotide.cr):
```
old: 1.63s, 26.5Mb
new: 1.01s, 32.4Mb
```

Then some more benchmarks...

There's `HTTP::Request#from_io` which I [recently optimized](https://github.com/crystal-lang/crystal/pull/8002):
```
old: from_io 498.47k (  2.01µs) (± 0.89%)  816B/op  fastest
new: from_io 549.22k (  1.82µs) (± 1.89%)  720B/op  fastest
```

(using `wrk` against the sample http server increases the requests/sec from 118355.73 to about 122000)

Also for curiosity I compared creating a Hash with 1_000_000 elements and seeing how it compares to Ruby and the old Hash.

<details>
  <summary>Code for creating a Hash with 1_000_000 elements in Ruby and Crystal</summary>

```crystal
size = 1_000_000

h = {0 => 0}

time = Time.now
size.times do |i|
  h[i] = i
end
puts "Insert: #{Time.now - time}"

time = Time.now
size.times do |i|
  h[i]
end
puts "Read: #{Time.now - time}"
```
</details>

Results:

```
Ruby 2.7-dev:
  Insert: 0.151813
  Read:   0.13749
Crystal old:
  Insert: 0.238662
  Read:   0.129462
Crystal new:
  Insert: 0.070804
  Read:   0.041008
```

Ruby was faster than Crystal! Ruby is simply amazing ❤️. But now Crystal is even faster!

## The compiler uses Hash all over the place!

So compile times now should go down! Right? ... Right?

Well, unfortunately no. I think the main reason is that the times are bound by the number of method instances, not by the performance of the many hashes used.

Memory consumption did seem to go down a bit.

## When will I have this?

~We could push this to 0.31.0. Or... we could have it in 0.30.0 if we delay the release a bit more (note: I don't manage releases, but if the community doesn't mind waiting a bit to get a huge performance boost in their apps then I think we could relax the release date).~

In 0.31.0.

## Final thoughts

1. Thank you Ruby! ❤️ ❤️ ❤️ 
1. Thank you Vladimir Makarov! ❤️ ❤️ ❤️ 
1. Algorithms are cool!
1. There's an optimization in the original Ruby algorithm which uses bitmasks instead of `remainder` or `%` to fit a number inside a range... I did that as almost the last thing because I didn't believe it would improve performance a lot... and it doubled the performance! 😮 
1. Feel free to target this PR and benchmark your code and post any interesting speedups here!
1. I hope CI 32 bits passes! 😊

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -218,16 +218,16 @@ class Hash(K, V)
   # less than 11 are ignored.
   def initialize(block : (Hash(K, V), K -> V)? = nil, *, initial_capacity = nil)
     initial_capacity = (initial_capacity || 0).to_i32
-    initial_indices_size = Math.pw2ceil(initial_capacity)
 
     # Same as the empty hash case
     # (but this constructor is a bit more expensive in terms of code execution).
-    if initial_indices_size == 0
+    if initial_capacity == 0
       @entries = Pointer(Entry(K, V)).null
       @indices = Pointer(UInt8).null
       @indices_size_pow2 = 0
       @indices_bytesize = 1
     else
+      initial_indices_size = Math.pw2ceil(initial_capacity)
       @entries = Pointer(Entry(K, V)).malloc(initial_indices_size / 2)
 
       # Check if we can avoid allocating the `@indices` buffer for
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -598,9 +598,10 @@ class Hash(K, V)
             else
               @indices.as(UInt32*)[index].to_i32!
             end
-    # Because 0 means empty we return -1,
-    # otherwise we decrement the value (so we can represent the index "0")
-    value == 0 ? -1 : value - 1
+
+    # Because we increment the value by one when we store the value
+    # here we have to substract one
+    value - 1
   end
 
   # Sets `@indices` at `index` with the given value.
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -607,6 +607,7 @@ class Hash(K, V)
     @size = 0
     @deleted_count = 0
     @indices_bytesize = 1
+    @first = 0
   end
 
   # Gets from `@indices` at the given `index`.
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -374,8 +374,8 @@ class Hash(K, V)
   # Upserts the key-value-hash triplet by doing a linear scan
   # first to see if the key already exists.
   # Returns true if the key was updated or inserted without needing
-  # a resize. Returns false if a resize was needed (and one) and
-  # the key wasn't inserted.
+  # a resize. Returns false if a resize was needed and the key
+  # wasn't inserted.
   private def upsert_linear_scan(key, value, hash) : Bool
     # Just do a linear scan...
     each_entry_with_index do |entry, index|
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1376,7 +1376,8 @@ class Hash(K, V)
 
   # Returns the first key in the hash.
   def first_key
-    first_key?.not_nil!
+    entry = first_entry?
+    entry ? entry.key : raise "Can't get first key of empty Hash"
   end
 
   # Returns the first key if it exists, or returns `nil`.
@@ -1393,7 +1394,8 @@ class Hash(K, V)
 
   # Returns the first value in the hash.
   def first_value
-    first_value?.not_nil!
+    entry = first_entry?
+    entry ? entry.value : raise "Can't get first value of empty Hash"
   end
 
   # Returns the first value if it exists, or returns `nil`.
@@ -1410,7 +1412,8 @@ class Hash(K, V)
 
   # Returns the last key in the hash.
   def last_key
-    last_key?.not_nil!
+    entry = last_entry?
+    entry ? entry.key : raise "Can't get last key of empty Hash"
   end
 
   # Returns the last key if it exists, or returns `nil`.
@@ -1427,7 +1430,8 @@ class Hash(K, V)
 
   # Returns the last value in the hash.
   def last_value
-    last_value?.not_nil!
+    entry = last_entry?
+    entry ? entry.value : raise "Can't get last value of empty Hash"
   end
 
   # Returns the last value if it exists, or returns `nil`.
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -699,7 +699,9 @@ class Hash(K, V)
   end
 
   # Yields each non-deleted Entry with its index inside `@entries`.
-  private def each_entry_with_index
+  private def each_entry_with_index : Nil
+    return if @size == 0
+
     entries_size.times do |i|
       entry = get_entry(i)
       yield entry, i unless entry.deleted?
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -379,7 +379,7 @@ class Hash(K, V)
   private def upsert_linear_scan(key, value, hash) : Bool
     # Just do a linear scan...
     each_entry_with_index do |entry, index|
-      if hash == entry.hash && entry.key == key
+      if entry.matches?(hash, key)
         set_entry(index, Entry(K, V).new(entry.hash, entry.key, value))
         return true
       end
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -525,7 +525,7 @@ class Hash(K, V)
     # have many deleted elements.
     if @deleted_count < @size
       # First grow `@entries`
-      @entries = @entries.realloc(indices_size)
+      realloc_entries(indices_size)
       double_indices_size
 
       # If we didn't have `@indices` and we still don't have 16 entries
@@ -828,6 +828,10 @@ class Hash(K, V)
   # Allocates `size` number of entries for `@entries`.
   private def malloc_entries(size)
     Pointer(Entry(K, V)).malloc(size)
+  end
+
+  private def realloc_entries(size)
+    @entries = @entries.realloc(size)
   end
 
   # Marks all existing entries as deleted
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -521,29 +521,32 @@ class Hash(K, V)
   # deleted. In that case the entries table is simply compacted.
   # However, in case of a resize deleted entries are also compcated.
   private def resize : Nil
-    # If we don't have `@indices` it means we have less than 16 elements
-    # in `@entries` so just grow that. Here we don't try to compact
-    # deleted elements (might not be worth it).
-    if @indices.null?
+    # Only do an actual resize (grow `@entries` buffer) if we don't
+    # have many deleted elements.
+    if @deleted_count < @size
+      # First grow `@entries`
       @entries = @entries.realloc(indices_size)
       double_indices_size
-      # Stop doing linear scan when we have more than 16 entries
-      if indices_size > MAX_INDICES_SIZE_LINEAR_SCAN
-        @indices_bytesize = compute_indices_bytesize(indices_size)
-        @indices = malloc_indices(indices_size)
-      else
+
+      # If we didn't have `@indices` and we still don't have 16 entries
+      # we keep doing linear scans (not using `@indices`)
+      if @indices.null? && indices_size <= MAX_INDICES_SIZE_LINEAR_SCAN
         return
       end
-    else
-      # Only resize if we don't have many deleted elements.
-      if @deleted_count < @size
-        @entries = @entries.realloc(indices_size)
-        double_indices_size
-        @indices_bytesize = compute_indices_bytesize(indices_size)
-        @indices = realloc_indices(indices_size)
-      end
 
-      # Must must clear the indices because we'll rebuild them from scratch
+      # Otherwise, we must start using `@indices`
+      @indices_bytesize = compute_indices_bytesize(indices_size)
+      @indices = malloc_indices(indices_size)
+    end
+
+    # `@indices` might still be null if we are compacting in the case where
+    # we are still doing a linear scan (and we had many deleted elements)
+    if @indices.null?
+      has_indices = false
+    else
+      # If we do have indices we must clear them because we'll rebuild
+      # them from scratch
+      has_indices = true
       clear_indices
     end
 
@@ -554,14 +557,16 @@ class Hash(K, V)
       # First we move the index to its new index (if we need to do that)
       set_entry(new_entry_index, entry) if entry_index != new_entry_index
 
-      # Then we try to find an empty index slot (we should find one now
-      # that we have more space)
-      index = fit_in_indices(entry.hash)
-      until get_index(index) == -1
-        index = next_index(index)
+      if has_indices
+        # Then we try to find an empty index slot
+        # (we should find one now that we have more space)
+        index = fit_in_indices(entry.hash)
+        until get_index(index) == -1
+          index = next_index(index)
+        end
+        set_index(index, new_entry_index)
       end
 
-      set_index(index, new_entry_index)
       new_entry_index += 1
     end
 
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -534,9 +534,14 @@ class Hash(K, V)
         return
       end
 
-      # Otherwise, we must start using `@indices`
+      # Otherwise, we must either start using `@indices`
+      # or grow the ones we had.
       @indices_bytesize = compute_indices_bytesize(indices_size)
-      @indices = malloc_indices(indices_size)
+      if @indices.null?
+        @indices = malloc_indices(indices_size)
+      else
+        @indices = realloc_indices(indices_size)
+      end
     end
 
     do_compaction
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -539,6 +539,18 @@ class Hash(K, V)
       @indices = malloc_indices(indices_size)
     end
 
+    do_compaction
+
+    # After compaction we no longer have deleted entries
+    @deleted_count = 0
+
+    # And the first valid entry is the first one
+    @first = 0
+  end
+
+  # Compacts `@entries` (only keeps non-deleted ones) and rebuilds `@indices.`
+  # If `rehash` is `true` then hash values inside each `Entry` will be recomputed.
+  private def do_compaction(rehash : Bool = false) : Nil
     # `@indices` might still be null if we are compacting in the case where
     # we are still doing a linear scan (and we had many deleted elements)
     if @indices.null?
@@ -554,8 +566,13 @@ class Hash(K, V)
     # while moving non-deleted entries to the beginning (compaction).
     new_entry_index = 0
     each_entry_with_index do |entry, entry_index|
-      # First we move the index to its new index (if we need to do that)
-      set_entry(new_entry_index, entry) if entry_index != new_entry_index
+      if rehash
+        # When rehashing we always have to copy the entry
+        set_entry(new_entry_index, Entry(K, V).new(key_hash(entry.key), entry.key, entry.value))
+      else
+        # First we move the entry to its new index (if we need to do that)
+        set_entry(new_entry_index, entry) if entry_index != new_entry_index
+      end
 
       if has_indices
         # Then we try to find an empty index slot
@@ -576,12 +593,6 @@ class Hash(K, V)
     if entries_to_clear > 0
       (entries + new_entry_index).clear(entries_to_clear)
     end
-
-    # After compaction we no longer have deleted entries
-    @deleted_count = 0
-
-    # And the first valid entry is the first one
-    @first = 0
   end
 
   # After this it's 1 << 28, and with entries being Int32
@@ -1667,17 +1678,7 @@ class Hash(K, V)
   # it was inserted into the `Hash` may lead to undefined behaviour.
   # This method re-indexes the hash using the current key values.
   def rehash : Nil
-    new_size = calculate_new_size(@size)
-    @buckets = @buckets.realloc(new_size)
-    new_size.times { |i| @buckets[i] = nil }
-    @buckets_size = new_size
-    entry = @last
-    while entry
-      index = bucket_index entry.key
-      entry.next = @buckets[index]
-      @buckets[index] = entry
-      entry = entry.back
-    end
+    do_compaction(rehash: true)
   end
 
   # Inverts keys and values. If there are duplicated values, the last key becomes the new value.
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -237,7 +237,7 @@ class Hash(K, V)
         initial_indices_size = Math.pw2ceil(initial_capacity)
       end
 
-      @entries = Pointer(Entry(K, V)).malloc(initial_indices_size / 2)
+      @entries = malloc_entries(initial_indices_size / 2)
 
       # Check if we can avoid allocating the `@indices` buffer for
       # small hashes.
@@ -326,7 +326,7 @@ class Hash(K, V)
     # Empty hash table so only initialize entries for now
     if @entries.null?
       @indices_size_pow2 = 3
-      @entries = Pointer(Entry(K, V)).malloc(4)
+      @entries = malloc_entries(4)
     end
 
     hash = key_hash(key)
@@ -626,6 +626,61 @@ class Hash(K, V)
     @first = 0
   end
 
+  # Initializes a `dup` copy from the contents of `other`.
+  protected def initialize_dup(other)
+    return if other.empty?
+
+    initialize_dup_entries(other)
+    initialize_copy_non_entries_vars(other)
+  end
+
+  # Initializes a `clone` copy from the contents of `other`.
+  protected def initialize_clone(other)
+    return if other.empty?
+
+    initialize_clone_entries(other)
+    initialize_copy_non_entries_vars(other)
+  end
+
+  # Initializes `@entries` for a dup copy.
+  # Here we only need tu duplicate the buffer.
+  private def initialize_dup_entries(other)
+    return if other.@entries.null?
+
+    @entries = malloc_entries(other.entries_capacity)
+
+    # Note that we only need to copy `entries_size` which
+    # are the effectives entries in use.
+    @entries.copy_from(other.@entries, other.entries_size)
+  end
+
+  # Initializes `@entries` for a clone copy.
+  # Here we need to copy entries while cloning their values.
+  private def initialize_clone_entries(other)
+    return if other.@entries.null?
+
+    @entries = malloc_entries(other.entries_capacity)
+
+    other.each_entry_with_index do |entry, index|
+      set_entry(index, entry.clone)
+    end
+  end
+
+  # Initializes all variables other than `@entries` for a copy.
+  private def initialize_copy_non_entries_vars(other)
+    @indices_bytesize = other.@indices_bytesize
+    @first = other.@first
+    @size = other.@size
+    @deleted_count = other.@deleted_count
+    @indices_size_pow2 = other.@indices_size_pow2
+    @block = other.@block
+
+    unless other.@indices.null?
+      @indices = malloc_indices(other.indices_size)
+      @indices.copy_from(other.@indices, indices_malloc_size(other.indices_size))
+    end
+  end
+
   # Gets from `@indices` at the given `index`.
   # Returns the index in `@entries` or `-1` if the slot is empty.
   private def get_index(index : Int32) : Int32
@@ -661,7 +716,7 @@ class Hash(K, V)
   end
 
   # Returns the capacity of `@indices`.
-  private def indices_size
+  protected def indices_size
     1 << @indices_size_pow2
   end
 
@@ -679,17 +734,22 @@ class Hash(K, V)
 
   # Allocates `size` number of indices for `@indices`.
   private def malloc_indices(size)
-    Pointer(UInt8).malloc(size * @indices_bytesize)
+    Pointer(UInt8).malloc(indices_malloc_size(size))
+  end
+
+  # The actual number of bytes needed to allocate `@indices`.
+  private def indices_malloc_size(size)
+    size * @indices_bytesize
   end
 
   # Reallocates `size` number of indices for `@indices`.
   private def realloc_indices(size)
-    @indices.realloc(size * @indices_bytesize)
+    @indices.realloc(indices_malloc_size(size))
   end
 
   # Marks all existing indices as empty.
   private def clear_indices : Nil
-    @indices.clear(indices_size * @indices_bytesize)
+    @indices.clear(indices_malloc_size(indices_size))
   end
 
   # Returns the entry in `@entries` at `index`.
@@ -747,11 +807,11 @@ class Hash(K, V)
 
   # Returns true if there's no place for new entries without doing a resize.
   private def entries_full? : Bool
-    entries_size == indices_size / 2
+    entries_size == entries_capacity
   end
 
   # Yields each non-deleted Entry with its index inside `@entries`.
-  private def each_entry_with_index : Nil
+  protected def each_entry_with_index : Nil
     return if @size == 0
 
     @first.upto(entries_size - 1) do |i|
@@ -760,9 +820,14 @@ class Hash(K, V)
     end
   end
 
+  # Allocates `size` number of entries for `@entries`.
+  private def malloc_entries(size)
+    Pointer(Entry(K, V)).malloc(size)
+  end
+
   # Marks all existing entries as deleted
   private def clear_entries
-    @entries.clear(indices_size / 2)
+    @entries.clear(entries_capacity)
   end
 
   # Computes the next index in `@indices`, needed when an index is not empty.
@@ -806,6 +871,11 @@ class Hash(K, V)
   # deleted and non-deleted ones.
   protected def entries_size
     @size + @deleted_count
+  end
+
+  # Returns the capacity of `@entries`.
+  protected def entries_capacity
+    indices_size / 2
   end
 
   # Computes the hash of a key.
@@ -1601,10 +1671,8 @@ class Hash(K, V)
   # hash_a # => {"foo" => "bar"}
   # ```
   def dup
-    hash = Hash(K, V).new(initial_capacity: @size)
-    each do |key, value|
-      hash[key] = value
-    end
+    hash = Hash(K, V).new
+    hash.initialize_dup(self)
     hash
   end
 
@@ -1617,10 +1685,8 @@ class Hash(K, V)
   # hash_a # => {"foobar" => {"foo" => "bar"}}
   # ```
   def clone
-    hash = Hash(K, V).new(initial_capacity: @size)
-    each do |key, value|
-      hash[key] = value.clone
-    end
+    hash = Hash(K, V).new
+    hash.initialize_clone(self)
     hash
   end
 
@@ -1725,6 +1791,10 @@ class Hash(K, V)
       {% else %}
         @hash == hash && @key == key
       {% end %}
+    end
+
+    def clone
+      Entry(K, V).new(hash, key, value.clone)
     end
   end
 
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -754,6 +754,9 @@ class Hash(K, V)
   # Internal implementation ends
   # ===========================================================================
 
+  # Returns the number of elements in this Hash.
+  getter size : Int32
+
   # Sets the value of *key* to the given *value*.
   #
   # ```
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -589,7 +589,7 @@ class Hash(K, V)
   # whether the maximum hash size is reached.
   private def double_indices_size : Nil
     if indices_size == MAXIMUM_INDICES_SIZE
-      raise "Maximim Hash size reached"
+      raise "Maximum Hash size reached"
     end
 
     @indices_size_pow2 += 1
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -230,7 +230,13 @@ class Hash(K, V)
       @indices_size_pow2 = 0
       @indices_bytesize = 1
     else
-      initial_indices_size = Math.pw2ceil(initial_capacity)
+      # Translate initial capacity to the nearest power of 2, but keep it a minimum of 8.
+      if initial_capacity < 8
+        initial_indices_size = 8
+      else
+        initial_indices_size = Math.pw2ceil(initial_capacity)
+      end
+
       @entries = Pointer(Entry(K, V)).malloc(initial_indices_size / 2)
 
       # Check if we can avoid allocating the `@indices` buffer for
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -147,10 +147,13 @@ class Hash(K, V)
   # Then all other methods use the internal methods, usually using other high-level
   # methods.
 
-  # The first valid entry in `@entries`.
+  # The index of the first non-deleted entry in `@entries`.
   # This is useful to support `shift`: instead of marking an entry
   # as deleted and then always having to ignore it we just increment this
   # variable and always start iterating from it.
+  # The invariant of `@first` always pointing to a non-deleted entry holds
+  # (unless `@size` is 0) and is guaranteed because of how
+  # `delete_and_update_counts` is implemented.
   @first : Int32 = 0
 
   # The buffer of entries.
@@ -506,16 +509,6 @@ class Hash(K, V)
     nil
   end
 
-  # Implementation on shift when we are sure there's something
-  # to shift (called from `shift`).
-  private def shift_impl
-    # We have to mark the entry as deleted for the GC to clear it
-    delete_entry(@first)
-    @first += 1
-    @deleted_count += 1
-    @size -= 1
-  end
-
   # Tries to resize the hash table in the condition that there are
   # no more available entries to add.
   # Might not result in a resize if there are many entries marked as
@@ -699,6 +692,29 @@ class Hash(K, V)
     delete_entry(index)
     @size -= 1
     @deleted_count += 1
+
+    # If we are deleting the first entry there are some
+    # more optimizations we can do
+    return if index != @first
+
+    # If the Hash is now empty then the first effective
+    # entry starts right after all the deleted ones.
+    if @size == 0
+      @first = @deleted_count
+    else
+      # Otherwise, we bump `@first` and keep bumping it
+      # until we find a non-deleted entry. It's guaranteed
+      # that this loop will end because `@size != 0` so
+      # there will be a non-deleted entry.
+      # It's better to skip the deleted entries once here
+      # and not every next time someone accesses the Hash.
+      # With this we also keep the invariant that `@first`
+      # always points to the first non-deleted entry.
+      @first += 1
+      while @entries[@first].deleted?
+        @first += 1
+      end
+    end
   end
 
   # Returns true if there's no place for new entries without doing a resize.
@@ -732,14 +748,10 @@ class Hash(K, V)
 
   # Returns the first `Entry` or `nil` if non exists.
   private def first_entry?
-    return nil if @size == 0
-
-    each_entry_with_index do |entry|
-      return entry
-    end
-
-    # Might happen if the Hash is modified concurrently
-    nil
+    # We always make sure that `@first` points to the first
+    # non-deleted entry, so `@entries[@first]` is guaranteed
+    # to be non-deleted.
+    @size == 0 ? nil : @entries[@first]
   end
 
   # Returns the first `Entry` or `nil` if non exists.
@@ -1503,7 +1515,7 @@ class Hash(K, V)
   def shift
     first_entry = first_entry?
     if first_entry
-      shift_impl
+      delete_entry_and_update_counts(@first)
       {first_entry.key, first_entry.value}
     else
       yield
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -147,6 +147,12 @@ class Hash(K, V)
   # Then all other methods use the internal methods, usually using other high-level
   # methods.
 
+  # The first valid entry in `@entries`.
+  # This is useful to support `shift`: instead of marking an entry
+  # as deleted and then always having to ignore it we just increment this
+  # variable and always start iterating from it.
+  @first : Int32 = 0
+
   # The buffer of entries.
   # Might be null if the hash is empty at the very beginning.
   # Has always the size of `indices_size` / 2.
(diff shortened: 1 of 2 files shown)


src/hash.cr
```diff
@@ -36,12 +36,162 @@ class Hash(K, V)
...
+  #
+  # There are two main data structures:
+  #
+  # - @entries:
+  #     A continguous buffer (Pointer) of hash entries (Entry) in the order
+  #     they were inserted. This makes it possible for Hash to preserve
+  #     order of insertion.
+  #     An entry holds a key-value pair together with the key's hash code.
+  #     An entry can also be marked as deleted. This is accomplished by using
+  #     0 as the hash code value. Becuase 0 is a valid hash code value, when
```
rumenzu:
```suggestion
  #     0 as the hash code value. Because 0 is a valid hash code value, when
```

src/hash.cr
```diff
@@ -782,7 +1376,7 @@ class Hash(K, V)
 
   # Returns the first key in the hash.
   def first_key
-    @first.not_nil!.key
+    first_key?.not_nil!
```
```suggestion
    first_key?.as K
```

src/hash.cr
```diff
@@ -794,12 +1388,12 @@ class Hash(K, V)
...
   # ```
   def first_key?
-    @first.try &.key
+    first_entry?.try &.key
   end
 
   # Returns the first value in the hash.
   def first_value
-    @first.not_nil!.value
+    first_value?.not_nil!
```
```suggestion
    first_value?.as V
```

src/hash.cr
```diff
@@ -828,12 +1422,12 @@ class Hash(K, V)
...
   # ```
   def last_key?
-    @last.try &.key
+    last_entry?.try &.key
   end
 
   # Returns the last value in the hash.
   def last_value
-    @last.not_nil!.value
+    last_value?.not_nil!
```
```suggestion
    last_value?.as V
```

src/hash.cr
```diff
@@ -811,12 +1405,12 @@ class Hash(K, V)
...
   # ```
   def first_value?
-    @first.try &.value
+    first_entry?.try &.value
   end
 
   # Returns the last key in the hash.
   def last_key
-    @last.not_nil!.key
+    last_key?.not_nil!
```
```suggestion
    last_key?.as K
```

src/hash.cr
asterite: Thanks, fixed!

src/hash.cr
asterite: That's not good because if `K` is nilable then this will succeed.

In any case thanks to your comments I realized the implementation of `first_key`, `last_key`, `first_value` and `last_value` were wrong. I fixed them.

src/hash.cr
asterite: See my other comment.

src/hash.cr
asterite: See my other comment.

src/hash.cr
asterite: See my other comment

src/hash.cr
straight-shoota: Why?

src/hash.cr
straight-shoota: Why? The type of `last_key?` is `K | Nil`, so `.not_nil!` removes the `Nil` type. That's much clearer than `.as(K)`.

j8r: @asterite you're right, we don't want to raise if the hash keys exists and are nil :+1:

lribeiro:
This gives a nice speedup on our internal app for log processing!

@bcardiff will this be considered for 0.30?

Time in seconds

|test|0.29|0.30dev|speedup|
|---|---:| ---:| ---:|
|forti|7.353|6.722|9.4%|
|forti_kv|5.462|4.867|12,2%|
|dns_grt|4.714|4.466|5,6%|


Details 

`~/Work/crystal-1/bin/crystal version`
```bash
Crystal 0.30.0-dev [ec2a26f9a] (2019-07-31)

LLVM: 8.0.0
```
```
~/Work/crystal-1/bin/crystal build --release -o bin/log_clean src/log_clean.cr 
bin/log_clean -c test/forti/forti.yml >    /dev/null  6.83s user 1.17s system 119% cpu 6.722 total
bin/log_clean -c test/forti/forti_kv.yml > /dev/null  5.11s user 2.02s system 146% cpu 4.867 total
bin/log_clean -c config/dns_grt.yml >      /dev/null  4.50s user 0.74s system 117% cpu 4.466 total

```

`crystal version`
```bash
Crystal 0.29.0 (2019-06-06)

LLVM: 6.0.1
```
```
bin/log_clean -c test/forti/forti.yml >    /dev/null  7.49s user 1.26s system 119% cpu 7.353 total
bin/log_clean -c test/forti/forti_kv.yml > /dev/null  5.80s user 2.08s system 144% cpu 5.462 total
bin/log_clean -c config/dns_grt.yml >      /dev/null  4.78s user 0.79s system 118% cpu 4.714 total
```

yxhuvud:
I wish github had properly threaded review comments. This is a comment to my own review: I note that you get lets of repeated allocation warnings and that one CI fail due to out of memory.

src/hash.cr
yxhuvud: Can not nil be used as a key, anyhow?

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+  private def each_entry_with_index
+    entries_size.times do |i|
+      entry = get_entry(i)
+      yield entry, i unless entry.deleted?
+    end
+  end
+
+  # Computes the next index in `@indices`, needed when an index is not empty.
+  private def next_index(index : Int32) : Int32
+    fit_in_indices(index + 1)
```
funny-falcon: "linear probing" is a bad idea.
Use at least "quadratic probing". It has same good cache locality on first three probes as for linear probing.

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+      # Try to do an upsert by doing a linear scan
+      upserted = upsert_linear_scan(key, value, hash)
+      return if upserted
+
+      # If we couldn't upsert it means the table was full
+      # so a resize might have been done.
+      # Now, it could happen that we are still with less than 16 elements
+      # and so `@indices` will be null, in which case we only need to
+      # add the key-value pair at the end of the `@entries` buffer.
+      if @indices.null?
```
funny-falcon: why not move this conditional block into `upsert_linear_scan`?

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+    @deleted_count += 1
+  end
+
+  # Returns true if there's no place for new entries without doing a resize.
+  private def entries_full? : Bool
+    entries_size == indices_size / 2
+  end
+
+  # Yields each non-deleted Entry with its index inside `@entries`.
+  private def each_entry_with_index
```
funny-falcon: `return if @size == 0`

straight-shoota:
The OOM failure on test_linux32 is pretty regular and unrelated to this PR.

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+      end
+    end
+  end
+
+  # Finds an Entry with the given key by doing a linear scan.
+  private def find_entry_linear_scan(key) : Entry(K, V)?
+    # If we have less than 8 elements we avoid computing the hash
+    # code and directly compare the keys (might be cheaper than
+    # computing a hash code of a complex structure).
+    if entries_size <= 8
```
funny-falcon: I doubt it is useful. I could be mistaken.
Was it benchmarked?

yxhuvud:
@straight-shoota Are all the warnings about large allocations also old?

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+            when 1
+              @indices[index].to_i32!
+            when 2
+              @indices.as(UInt16*)[index].to_i32!
+            else
+              @indices.as(UInt32*)[index].to_i32!
+            end
+    # Because 0 means empty we return -1,
+    # otherwise we decrement the value (so we can represent the index "0")
+    value == 0 ? -1 : value - 1
```
funny-falcon: value is Int32. Then why not just `value - 1` ?

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+      @indices[index] = value.to_u8!
+    when 2
+      @indices.as(UInt16*)[index] = value.to_u16!
+    else
+      @indices.as(UInt32*)[index] = value.to_u32!
+    end
+  end
+
+  # Marks `@indices` at `index` as empty.
+  private def empty_index(index) : Nil
```
funny-falcon: unused unnecessary method

src/hash.cr
```diff
@@ -966,7 +1559,7 @@ class Hash(K, V)
   # hash_a # => {"foobar" => {"foo" => "bar"}}
   # ```
   def clone
-    hash = Hash(K, V).new(initial_capacity: @buckets_size)
+    hash = Hash(K, V).new(initial_capacity: @size)
```
funny-falcon: clone could be extremely optimized (in case when `@deleted_count < @size`): we need just copy entries and indices, no need in iteration.

src/hash.cr
```diff
@@ -36,12 +36,162 @@ class Hash(K, V)
...
+  # Might be null if the hash is empty at the very beginning or when the hash
+  # size is less than or equal to 16.
+  # Could be a Slice but this way we might save a few bounds checking.
+  @indices : Pointer(UInt8)
+
+  # The size of `@indices` given as a power of 2.
+  # For example if it's 4 it means 2**4 so size 16.
+  # Can be zero when hash is totally empty.
+  # Otherwise guaranteed to be at least 3.
+  @indices_size_pow2 : UInt32
```
funny-falcon: no need for `@indices_size_pow2` to be `UInt32`. It could be just `UInt8`. Then placed near `@indices_bytesize` it will be better fitting.

src/hash.cr
```diff
@@ -1047,91 +1640,65 @@ class Hash(K, V)
...
+      {% else %}
+        @hash == hash && @key == key
+      {% end %}
     end
   end
 
   private module BaseIterator
-    def initialize(@hash, @current)
+    def initialize(@hash)
+      @index = 0
```
funny-falcon: I'd rather track index of a first element (and update it in `delete_entry_and_update_counts`).
If one calls `hash.shift` in a loop, they will suffer from `O(n^2)` without that. (Queue or LRU usecases).

If `@indices_size_pow2` will be changed to `UInt8` and placed near `@indices_bytesize`, then addition if `@first : UInt32` will not change size of hash structure.

funny-falcon:
Rather clean code! I liked to read it.

src/hash.cr
asterite: In Ruby they tried quadratic probing and it turned out to be a bit slower. They use linear probing. That's why I also chose that, and it's simpler.

Ref: https://github.com/ruby/ruby/blob/c94cc6d968a7241a487591a9753b171d8081d335/st.c#L869-L872

But if you want we can try it. I know nothing about quadratic probing, is it just `index, `index + 1, index + 4, index + 9, index + 16, etc.`?

konovod: According to the ref they aren't using linear probing now, they are using Double Hashing.

src/hash.cr
asterite: I think it might be a bit clearer this way: the `@indices.null?` check is done once again here and if it doesn't hold you know the code will continue running and `@indices` won't be null. If that's hidden inside `upsert_linear_scan` it's harder to tell.

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+
+  # Upserts the key-value-hash triplet by doing a linear scan
+  # first to see if the key already exists.
+  # Returns true if the key was updated or inserted without needing
+  # a resize. Returns false if a resize was needed (and one) and
+  # the key wasn't inserted.
+  private def upsert_linear_scan(key, value, hash) : Bool
+    # Just do a linear scan...
+    each_entry_with_index do |entry, index|
+      if hash == entry.hash && entry.key == key
```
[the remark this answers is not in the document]
asterite: Thank you, I missed this one.

src/hash.cr
```diff
@@ -118,6 +290,479 @@ class Hash(K, V)
...
+        # Otherwise we have to keep looking...
+        index = next_index(index)
+      end
+    end
+  end
+
+  # Upserts the key-value-hash triplet by doing a linear scan
+  # first to see if the key already exists.
+  # Returns true if the key was updated or inserted without needing
+  # a resize. Returns false if a resize was needed (and one) and
```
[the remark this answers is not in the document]
asterite: I don't know 😅 Fixed! (removed)

src/hash.cr
asterite: Good catch, thanks!

src/hash.cr
asterite: You are right! I don't know what I was thinking, hehe... thanks!

src/hash.cr
asterite: Thanks! I used that before using `Pointer#clear` and later forgot to remove this.

src/hash.cr
asterite: You are right! I think a similar optimization could have been done in the previous code. I'd like to leave this optimization for a separate PR so this PR doesn't become a bit harder to review (and get right 😅).

src/hash.cr
asterite: I'm sorry but I don't understand what you mean.

src/hash.cr
asterite: You are right! I changed it to `UInt8`. I think previously I tracked `@indices_size` and that was `UInt32` and then I replaced it with pow2 but didn't realized the size could be slower.

src/hash.cr
asterite: Yes. I just benchmarked it again because I introduced that optimization some time ago, before introducing more optimizations and changing other things.

I ran the same bencharmks as in my main post but only for String. The benchmark for `insert` don't have a difference. For `read` it gives this (old is without the optimization, new is with it):

```
old, read (type: String, size: 1)  70.65M ( 14.15ns) (± 2.49%)  0.0B/op   2.68× slower
new, read (type: String, size: 1) 189.05M (  5.29ns) (± 3.89%)  0.0B/op        fastest
old, read (type: String, size: 2)  39.07M ( 25.59ns) (± 2.05%)  0.0B/op   2.26× slower
new, read (type: String, size: 2)  88.16M ( 11.34ns) (± 4.01%)  0.0B/op        fastest
old, read (type: String, size: 3)  26.56M ( 37.65ns) (± 1.84%)  0.0B/op   2.00× slower
new, read (type: String, size: 3)  53.24M ( 18.78ns) (± 1.89%)  0.0B/op        fastest
old, read (type: String, size: 4)  19.98M ( 50.06ns) (± 2.17%)  0.0B/op   1.72× slower
new, read (type: String, size: 4)  34.35M ( 29.11ns) (± 3.70%)  0.0B/op        fastest
old, read (type: String, size: 5)  16.00M ( 62.49ns) (± 1.61%)  0.0B/op   1.49× slower
new, read (type: String, size: 5)  23.78M ( 42.05ns) (± 2.05%)  0.0B/op        fastest
old, read (type: String, size: 6)  13.11M ( 76.25ns) (± 1.49%)  0.0B/op   1.39× slower
new, read (type: String, size: 6)  18.17M ( 55.02ns) (± 2.68%)  0.0B/op        fastest
old, read (type: String, size: 7)  11.21M ( 89.24ns) (± 1.37%)  0.0B/op   1.22× slower
new, read (type: String, size: 7)  13.71M ( 72.93ns) (± 3.94%)  0.0B/op        fastest
old, read (type: String, size: 8)   9.57M (104.47ns) (± 3.21%)  0.0B/op   1.16× slower
new, read (type: String, size: 8)  11.14M ( 89.77ns) (± 3.77%)  0.0B/op        fastest
```

So you can see it's a big optimization. You can also see it becomes faster and faster to start compare the hash first when there are more and more elements, starting from 9 it's almost the same.

I wonder if this optimization could also be applied in the Ruby code... 🤔 

I might give it a try later (in Ruby), now I'm curious!

asterite:
@funny-falcon 

> Rather clean code! I liked to read it.

Thank you! 😊 And thank you for the thorough review, you found many good things to fix.

src/hash.cr
asterite: Actually, I think if the key type is more complex and `==` takes more time then it will be slower. So maybe it should also be applied for a few cases... maybe just do it if the key is a String because we know it's true.

src/hash.cr
asterite: It's true! However I don't understand how that works:

     Still other bits of the hash value are used when the mapping
     results in a collision.  In this case we use a secondary hash
     value which is a result of a function of the collision bin
     index and the original hash value.  The function choice
     guarantees that we can traverse all bins and finally find the
     corresponding bin as after several iterations the function
     becomes a full cycle linear congruential generator because it
     satisfies requirements of the Hull-Dobell theorem.

I don't know about that theorem so it's harder for me to implement something I don't understand. But I'll definitely have that in mind for further optimizing this (but it'll me some more time, but PRs are welcome!).

src/hash.cr
asterite: I just tried it with keys being `Array(String)` and the differences are actually much bigger, meaning that it's faster to avoid checking the `hash` code altogether for some reason... probably because to compute the hash code you need to consider the entire structure, but to compare elements you bail out as soon as you find a difference.

src/hash.cr
asterite: Ah, yes, I see what you mean. I think we should either remove `Hash#shift` or implement your suggestion. We copied `Hash#shift` from Ruby but I don't know what's a real use for that it.

asterite:
@yxhuvud 

> This is a comment to my own review comment: I note that you get lets of repeated allocation warnings and that one CI fail due to out of memory.

Yes, but I think the errors were there before. I think `Hash` might require a bit more memory because now it always doubles its size when a resize is needed. Also we keep adding tests to the spec suite and right now the compiler specs leak memory. I'll open another issue to track this problem. The solution for now is to just run specs on CI in smaller chunks instead of in two big chunks. This is a problem that only affects CI, regular machines usually have a lot of memory and don't need to compile the entire Crystal's spec suite.

funny-falcon:
`Hash.shift` is rather useful for:
- simple LRU implementation
- keyed FIFO queue
Given it is quite easy to add @first , it is better to add it than to not.

rumenzu:
A quick [benchmark](https://github.com/rumenzu/quartz/blob/master/benchmarks/devstone.cr) from my simulation framework reveals a really nice speedup of **~1.37x**!
Hash was already identified as a bottleneck in my case after profiling.

`./devstone 10000 3` results:

Crystal version | Wall clock time (s)
--------------- | --------------------
Crystal 0.30.0-dev [008647480] (2019-07-31) LLVM: 8.0.0 | 7.797805118
Crystal 0.29.0 (2019-06-06) LLVM: 6.0.1 | 5.703113211

asterite:
> Given it is quite easy to add @FIRST , it is better to add it than to not.

Yes, but it adds 4 bytes for all hashes even if you never use `shift`. Same reason why we didn't do that for `Array`. If you need `shift` like that you can use `Deque`, which is in the standard library and it's more efficient (at the cost of a bit more memory).

Said another way, we can consider this as a separate issue.

yxhuvud:
@asterite: I think the warnings isn't due to requiring more memory but because libgc doesn't like big allocations. That is, it is perfectly happy to allocate a GB if you do it in 4KB chunks, but if you do it all at once you get a warning. So doing fewer tests wouldn't really help with the warnings as you would still get one once the amount of entries in a hash get big enough.

asterite:
@yxhuvud Makes sense. I think I added a couple of tests that exercise some big-ish hashes.

asterite:
@funny-falcon On second thought it seems we can put the 4 bytes of `@first` in a place where it doesn't increase the size of `Hash`, at least in 64 bits where everything is aligned to 8 bytes. And the same is true for Array! So I think I'll add those (I'll add `@first` for Hash in this PR, and for Array in a separate PR). It'll be cool to have O(1) `shift`. I'm thinking for Array I'll just increment `@first` but also clear the value behind that pointer. Though I'll probably wait until parallelism or coordinate with @bcardiff because I know he's been doing some changes to Array to make it a bit more safe in parallel environments.

src/hash.cr
asterite: Done in https://github.com/crystal-lang/crystal/pull/8017/commits/01e072cdfa948aeb236b642543a7ed89990a346f !

@funny-falcon  I'll be happy if you can review the commit. The change turned out to be pretty simple thanks to the built abstractions.

asterite:
Now with optimized `shift`, thanks to @funny-falcon 's suggestion.

Benchmark. Code:

```crystal
old_hash = OldHash(Int32, Int32).new
new_hash = Hash(Int32, Int32).new

p instance_sizeof(typeof(old_hash))
p instance_sizeof(typeof(new_hash))

100_000.times do |i|
  old_hash[i] = i
  new_hash[i] = i
end

time = Time.now
while old_hash.shift?
end
puts "Old: #{Time.now - time}"

time = Time.now
while new_hash.shift?
end
puts "New: #{Time.now - time}"
```

Results:

```
56
56
Old: 00:00:02.272458000
New: 00:00:00.000244000
```

(meaning the bytesize of a Hash didn't change with this optimization)

RX14:
This is absolutely amazing! I think robin hood hashing would be a neat future optimization

src/hash.cr
```diff
@@ -60,13 +210,35 @@ class Hash(K, V)
...
-    initial_capacity = initial_capacity.to_i
-    @buckets = Pointer(Entry(K, V)?).malloc(initial_capacity)
-    @buckets_size = initial_capacity
+  def initialize(block : (Hash(K, V), K -> V)? = nil, *, initial_capacity = nil)
+    initial_capacity = (initial_capacity || 0).to_i32
+    initial_indices_size = Math.pw2ceil(initial_capacity)
+
+    # Same as the empty hash case
+    # (but this constructor is a bit more expensive in terms of code execution).
+    if initial_indices_size == 0
```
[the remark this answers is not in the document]
asterite: Good idea! Done, thanks

src/hash.cr
```diff
@@ -118,6 +290,473 @@ class Hash(K, V)
...
+  # After this it's 1 << 28, and with entries being Int32
+  # (4 bytes) it's 1 << 30 of actual bytesize and the
+  # next value would be 1 << 31 which overflows `Int32`.
+  private MAXIMUM_INDICES_SIZE = 1 << 28
+
+  # Doubles the value of `@indices_size` but first checks
+  # whether the maximum hash size is reached.
+  private def double_indices_size : Nil
+    if indices_size == MAXIMUM_INDICES_SIZE
+      raise "Maximim Hash size reached"
```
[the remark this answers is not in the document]
asterite: Thanks, fixed!

funny-falcon:
It is better to maintain `@first` in `delete_entry_and_update_counts`. It is single point of all deletions.
And it is better not simply `@first+=1` but:

```ruby
delete_entry(index)
@size-=1
@deleted_count+=1
return if index != @first
if @size == 0
  @first = @deleted_count
else
  @first += 1
  @first += 1 while @entries[@first].deleted?
end
```

And then first_entry could be optimized as well:
```ruby
@size == 0 ? nil : @entries[@first]
```

And shift

```ruby
if first_entry
  delete_entry_and_update_counts(@first)
  {first_entry.key, first_entry.value}
else
```

ysbaddaden:
You may also consider the [Robin Hound Hashing](https://cs.uwaterloo.ca/research/tr/1986/CS-86-14.pdf) (site may not be responding) optimization. Basically: 
> when you probe for a position to insert a new element, if the probe length for the existing element is less than the current probe length for the element being inserted, swap the two elements and keep going.

- [Robin Hood Hashing should be your default Hash Table implementation](https://www.sebastiansylvan.com/post/robin-hood-hashing-should-be-your-default-hash-table-implementation/)

Plus the *Learn Hash Table the Hard Way* series that talks about probes (linear, quadratic, double hash) and Robin Hound Hashing:
- [Part 1: Probe Distributions](http://www.idryman.org/blog/2017/07/04/learn-hash-table-the-hard-way/)
- [Part 2: Probe Distributions With Deletions](http://www.idryman.org/blog/2017/07/18/learn-hash-table-the-hard-way-2/)
- [Part 3: Probe Distributions and Run Time Performance](http://www.idryman.org/blog/2017/08/06/learn-hash-table-the-hard-way-3/)

funny-falcon:
@ysbaddaden Robin Hood hashing is useful for high load factor (close to 0.8-0.9). Proposed hash structure never exceeded load factor 0.5 by design (because indexes array always twice as large than entries array). Even already encoded linear search performs reasonably well with such low load factor. I believe, quadratic probing could be a bit better, but I'm not really sure. Ruby uses 

Robin Hood hashing will add some bunch of complexity without real benefits, imho. I could be mistaken.

RX14:
@funny-falcon the benefit is that we can increase the load factor, reducing the memory footprint.

funny-falcon:
> @funny-falcon the benefit is that we can increase the load factor, reducing the memory footprint.

@RX14 increasing load factor along will make this code much more complex. Current it's beauty is from simplicity of sizes: always power of two, indices are twice larger than entries.

In my attempts for this issue I always tried to make load factor greater, and to untangle indices and entries size. But my code were much more complex.

RX14:
> increasing load factor along will make this code much more complex

I realise that, but `Hash` performance is extremely important for the speed of the language. The performance/complexity tradeoff here is complicated. That's why I suggest we discuss this later. The performance benefits of robin hood hashing probably need to be tried *in practice* to get an idea of whether it's wanted. The gains in this PR are very substantial, and should not be delayed for such an investigation.

funny-falcon:
> I realise that, but Hash performance is extremely important for the speed of the language.

If hash performance is extremely important, then you don't want to increase load factor.
Rubin Hood has no performance benefits. If you read [paper](http://www.idryman.org/blog/2017/08/06/learn-hash-table-the-hard-way-3/) mentioned by @ysbaddaden carefully, then you noticed:

> Actually, the mean of quadratic probing is slightly better than quadratic robin hood hashing (1.3 vs 1.45).

Robin Hood hashing doesn't give better "average" performance. It only gives guarantee of bounded "worst" performance. In other words, application will not be magically faster in average with Robin Hood hashing. But there will be more guarantee about time of each separate operation.

Rust used to have load factor 0.9 with Robin Hood hashing, but it gives measurable performance overhead. https://github.com/rust-lang/rust/issues/38003

Now they [moved to other hash implementation](https://github.com/rust-lang/rust/pull/58623) that combines bulk SSE check of a bucket of entries with [quadratic probing of buckets](https://github.com/rust-lang/hashbrown/blob/ed0b240192c3a5c5c20995c7f8188453ce4e3a2f/src/raw/mod.rs#L156-L168).
Yes, it has impressive [0.875 load factor](https://github.com/rust-lang/hashbrown/blob/ed0b240192c3a5c5c20995c7f8188453ce4e3a2f/src/raw/mod.rs#L199-L208).
But it is not easily mapped to "ordered" hash. Bits of hash values have to be embedded into index table to use same trick.

Note that main benefit from Robin Hood hashing is a CPU cache locality. But given we have to journey from index table to entries table to check value, and order of elements in index table does not correlate to order in entries table, we already have pure cache locality. Embedding bits of hash value into index table will help, but will increase memory consumption.

Other trick to shortage average collision chain in quadratic probing is to use "collision" bit: on insertion set "collision" on all collisioned index entries, and during lookup iteration should stop if current element has no "collision" bit set. It trades some CPU for shorter collision chain. I believe, it makes "quadratic probing" much closer to "robin hood" hashing in term of "collision chain length distribution".

Any way, all my words should be taken with grain of salt. I often make false assumtions.

RX14:
@funny-falcon thank you! The above articles on RH hashing weren't entirely clear on this.

Memory overhead of Hash is important too, and I'd be interested in a graph of the allocation size of `@indexes` and `@entries` in bytes for varying `Hash(Reference, Reference)` sizes. I suspect that the load factor in `@indexes` is fairly insignificant compared to the size of `@entries`.

asterite:
@funny-falcon Thank you! I implemented all your suggestions. Let me know if I got something wrong 😊 

Although I think that many things can be still optimized (representation of things, the probing, the load factor, etc.) I think this is already a good improvement and we can continue refining things after this PR.

asterite:
Could someone explain to me with some ascii art of graphics how Robin Hood hashing works?

> When a collision occur, compare the two items’ probing count, the one with larger probing number stays and the other continue to probe. Repeat until the probing item finds an empty spot.

I can't understand this. Or... I can understand what it says but not what it means.

Let's suppose we start with empty buckets:

```
[_, _, _, _, _, _] 
```

Now we want to insert 10, which hashes to bucket 1:

```
[_, 10, _, _, _, _] 
```

Say we want to insert 20, which hashes to bucket 1 too... What do we do? The probe count for 10 is 0 (I guess?) and for 20 we don't know yet but it's definitely higher than 0 because it's occupied. So we swap them...?

```
[_, 20, 10, _, _, _] 
```

Next comes 30, again with bucket 1. So 20 has probe count 0, 10 has probe count 1, and 30 will have probe count 2... so we span everything again?

```
[_, 30, 20, 10, _, _] 
```

I can't see the benefit of this, but I'm sure there's something I'm not understanding.

konovod:
In Robin Hood hashing, we compare each element 
`[_, _, _, _, _, _] `
Now we want to insert 10, which hashes to bucket 1:
```
[_, 10, _, _, _, _]
  ,  0,  ,  ,  ,  ```

konovod:
In Robin hood hashing, we compare old element probe count with current probe count of our element.
So in your example nothing will be shifted - hash table will be 
`[_, 10, 20, 30, _, _]`
Now if we insert element 21 that hashes to 2, we will compare it to 20 (0 vs 1) and 30(1 vs 2) and still insert at 4:
`[_, 10, 20, 30, 21, _]`
But now if we insert 40 that hashes to 1:
it will skip 10, 20, 30 as they have same probe count, but will steal a place from 21 as it will have probe count 3 and 21 will have just 1:
`[_, 10, 20, 30, 40, _]`
and then we continue to search place for 21, finding it at next position:
`[_, 10, 20, 30, 40, 21]`

asterite:
@konovod Thanks, I understand now!

asterite:
Thank you everyone for the suggestions!

If you find optimizations please send PRs and benchmarks.

asterite:
Some optimizations ideas:
- Robin Hood hashing
- Double hashing
- Quadratic probing
- Reduce the allocated size of `@entries` (it could be possible to allocate less than `@indices_size / 2`)
- Other ideas...?

j8r:
That's good points @asterite, better to create issue to track them. It can even be a RFC.

asterite:
No. If someone finds an optimization please send a PR with benchmark. I don't plan to change anything else otherwise, and there's no need to track anything.

funny-falcon:
Gold words.

вс, 4 авг. 2019 г., 23:03 Ary Borenszweig <notifications@github.com>:

j8r:
I think it would be better to have proper RFCs about the design decisions, instead of PRs, just a thought.
Of course that's up to the core members to organize their project as they wish, and what they consider the best.

PR #8024, opened 2019-08-01
WebSocket: add blockable recieve and recieve_raw methods

### LINKED ISSUE #5600: WebSockets should have blocking methods
WebSockets in Crystal are very difficult to work with, not just because they are basically undocumented, but also because it's methods don't block which is not very Crystal-like.

I believe, and @oprypin will back me up, that WebSockets should be handled in the following way:

```crystal
ws = HTTP::WebSocket.new("ws://echo.websocket.org", "/")

ws.send("hello", "This is a message")
message = ws.receive
```
Where ws.receive blocks until a response comes.
If anyone wants this to be asynchronously they can always spawn a fiber.

```crystal
ws = HTTP::WebSocket.new("ws://echo.websocket.org", "/")
my_channel = Channel(String).new

spawn do
  ws.send("hello", "This is a message")
  message = ws.receive
  my_channel.send(message)
end

puts my_channel.receive
```

zenithlight:
To resolve #5600, I added a couple new methods to WebSocket. 

Like TCP sockets, these methods block until a message is received, and returns it. Unlike TCP sockets, a WebSocket message can consist of several wire packets. In addition, it returns one of 5 message structs depending on what opcode was received. The body of the message can be either String or Bytes depending on the opcode (Bytes for Binary, String for everything else).

I also refactored `run` to use the new `receive` method.

`receive_raw` simply returns a message struct. `receive` wraps `receive_raw` to provide behaviors like responding to PING with PONG, and closing the socket on a CLOSE message.

I have three concerns with this approach which I would like input on:

1. This is technically a breaking change, as before, during `run`, those behaviors (pong, close) were executed after the user-supplied callback was called, and now they are executed before. Probably the code can be changed to maintain compatibility.

2. I would like to add unit tests for `receive` and `receive_raw`, but not sure how to mock the `HTTP::WebSocket::Protocol` inside. Are there any examples in the stdlib using mocks?

3. I'm not sure what nomenclature to use for the "body" of the message struct. The spec references "payload data" so maybe `payload` would make more sense than `message`.

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -1,5 +1,15 @@
 class HTTP::WebSocket
   getter? closed = false
+  getter! close_reason : String
+
+  enum Opcode : UInt8
+    CONTINUATION = 0x0
+    TEXT         = 0x1
+    BINARY       = 0x2
+    CLOSE        = 0x8
+    PING         = 0x9
+    PONG         = 0xA
+  end
 
   # :nodoc:
   def initialize(io : IO)
@@ -100,52 +110,101 @@ class HTTP::WebSocket
     @ws.close(message)
   end
 
-  def run
+  struct Message
+    getter opcode
+
+    def initialize(@opcode : Opcode, @data : String | Bytes)
+    end
+
+    def text
+      @data.as(String)
+    end
+
+    def binary
+      @data.as(Bytes)
+    end
+  end
+
+  private def handle_opcode(message)
+    case message.opcode
+    when .close?
+      close(message.text) unless closed?
+      @close_reason = message.text
+    when .ping?
+      pong(message.text) unless closed?
+    end
+
+    if message.opcode.text? || message.opcode.binary?
+      return true
+    else
+      return false
+    end
+  end
+
+  # Waits until a full String or Bytes message is received on the socket and returns it.
+  # The data can be accessed in the `message` property of the returned struct.
+  #
+  # This method will respond to PING frames with PONG and close when a CLOSE frame is received.
+  # Use `#receive_raw` if you don't want that.
+  def receive
+    loop do
+      data = receive_raw
+
+      if handle_opcode(receive_raw)
+        return data
+      end
+    end
+  end
+
+  def receive_raw!
     loop do
       begin
         info = @ws.receive(@buffer)
       rescue IO::EOFError
-        @on_close.try &.call("")
+        return Message.new(:close, "")
         break
       end
 
+      @current_message.write @buffer[0, info.size]
+      next unless info.final
+
       case info.opcode
-      when Protocol::Opcode::PING
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          message = @current_message.to_s
-          @on_ping.try &.call(message)
-          pong(message) unless closed?
-          @current_message.clear
-        end
-      when Protocol::Opcode::PONG
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          @on_pong.try &.call(@current_message.to_s)
-          @current_message.clear
-        end
-      when Protocol::Opcode::TEXT
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          @on_message.try &.call(@current_message.to_s)
-          @current_message.clear
-        end
-      when Protocol::Opcode::BINARY
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          @on_binary.try &.call(@current_message.to_slice)
-          @current_message.clear
-        end
-      when Protocol::Opcode::CLOSE
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          message = @current_message.to_s
-          @on_close.try &.call(message)
-          close(message) unless closed?
-          @current_message.clear
-          break
-        end
+      when .text?
+        message = Message.new(:text, @current_message.to_s)
+      when .binary?
+        message = Message.new(:binary, @current_message.to_slice)
+      when .close?
+        message = Message.new(:close, @current_message.to_s)
+      when .ping?
+        message = Message.new(:ping, @current_message.to_s)
+      when .pong?
+        message = Message.new(:pong, @current_message.to_s)
       end
+
+      @current_message.clear
+      return message.not_nil!
+    end
+  end
+
+  def run
+    loop do
+      message = receive_raw!
+
+      case message.opcode
+      when .ping?
+        @on_ping.try &.call(message.text)
+      when .pong?
+        @on_pong.try &.call(message.text)
+      when .text?
+        @on_message.try &.call(message.text)
+      when .binary?
+        @on_binary.try &.call(message.binary)
+      when .close?
+        @on_close.try &.call(message.text)
+        break
+      end
+
+      handle_opcode(message)
     end
   end
 end
--- a/spec/std/http/web_socket_spec.cr
+++ b/spec/std/http/web_socket_spec.cr
@@ -5,19 +5,19 @@ require "random/secure"
 require "../../support/ssl"
 
 private def assert_text_packet(packet, size, final = false)
-  assert_packet packet, HTTP::WebSocket::Protocol::Opcode::TEXT, size, final: final
+  assert_packet packet, HTTP::WebSocket::Opcode::TEXT, size, final: final
 end
 
 private def assert_binary_packet(packet, size, final = false)
-  assert_packet packet, HTTP::WebSocket::Protocol::Opcode::BINARY, size, final: final
+  assert_packet packet, HTTP::WebSocket::Opcode::BINARY, size, final: final
 end
 
 private def assert_ping_packet(packet, size, final = false)
-  assert_packet packet, HTTP::WebSocket::Protocol::Opcode::PING, size, final: final
+  assert_packet packet, HTTP::WebSocket::Opcode::PING, size, final: final
 end
 
 private def assert_close_packet(packet, size, final = false)
-  assert_packet packet, HTTP::WebSocket::Protocol::Opcode::CLOSE, size, final: final
+  assert_packet packet, HTTP::WebSocket::Opcode::CLOSE, size, final: final
 end
 
 private def assert_packet(packet, opcode, size, final = false)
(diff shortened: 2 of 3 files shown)


RX14:
`receive` must handle ping and close messages, to avoid breaking the protocol. It should return just  `String | Bytes | Nil`. The close message would be stored in `ws.close_reason`, accessible after `receive` returns `nil`.

`receive_raw` can stay as-is. `run` can use `receive_raw` to re-implement the exact current semantics.

src/http/web_socket.cr
```diff
@@ -100,52 +101,86 @@ class HTTP::WebSocket
     @ws.close(message)
   end
 
-  def run
+  record Ping, message : String
+  record Pong, message : String
+  record Text, message : String
+  record Binary, message : Bytes
+  record Close, message : String
```
straight-shoota: Does this need to be five different types with almost the same layout? Maybe it would be better to have only a single struct with an enum property to designate the type of message?

src/http/web_socket.cr
```diff
@@ -1,5 +1,6 @@
 class HTTP::WebSocket
   getter? closed = false
+  getter? close_reason : String? = nil
```
RX14: I think this would be nicer as `getter!` actually

src/http/web_socket.cr
```diff
@@ -100,52 +101,86 @@ class HTTP::WebSocket
...
+    data
+  end
+
+  # Waits until a full message is received on the socket and returns it.
+  # The data can be accessed in the `message` property of the returned struct.
+  #
+  # This method will respond to PING frames with PONG and close when a CLOSE frame is received.
+  # Use `#receive_raw` if you don't want that.
+  def receive
+    handle_opcodes(receive_raw)
```
RX14: This needs to loop until either a `Close` or a `Binary`/`Text` message appears,

src/http/web_socket.cr
```diff
@@ -100,52 +101,86 @@ class HTTP::WebSocket
...
   end
 
-  def run
+  record Ping, message : String
+  record Pong, message : String
+  record Text, message : String
+  record Binary, message : Bytes
+  record Close, message : String
+
+  def handle_opcodes(data)
```
RX14: This should probably return just `true`/`false` depending if it's a `Ping`/`Pong` or other packet.

It should definitely be a private method.

src/http/web_socket.cr
zenithlight: Interesting. Do you think it makes sense to reuse the Protocol::Opcodes? And for the message property, it would have a type of String | Bytes?

straight-shoota: Yeah, message type is a bit tricky. As long as `String` support ordinary non-utf8 data it would suffice alone, but we can't count on that...

asterite: I would store it internally as a union and expose it through two properties that cast the value with `as`.

zenithlight:
I created an `Opcode` enum and a `Message` struct.

`Message` has an `opcode` property and stores the data internally as a `String | Bytes`. This is exposed as `String` and `Bytes` respectively by the getters `text` and `binary`. Is this roughly what you had in mind?

src/http/web_socket.cr
```diff
@@ -101,86 +101,109 @@ class HTTP::WebSocket
...
+  private def handle_opcodes(message)
+    case message.opcode
+    when Opcode::Close
+      close(message.text) unless closed?
+      @close_reason = message.text
+    when Opcode::Ping
+      pong(message.text) unless closed?
+    end
+
+    if message.is_a?(String | Bytes)
```
Sija: This will always fail.

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
...
+      @data.as(Bytes)
+    end
+
+    def initialize(@opcode : Opcode, @data : String | Bytes)
+    end
+  end
+
+  private def handle_opcodes(message)
+    case message.opcode
+    when Opcode::Close
```
Sija:
```suggestion
    when.close?
```

ditto all below

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
...
+    end
+  end
+
+  def receive_raw!
     loop do
       begin
         info = @ws.receive(@buffer)
       rescue IO::EOFError
-        @on_close.try &.call("")
+        return Message.new(Opcode::Close, "")
```
Sija:
```suggestion
        return Message.new(:close, "")
```

ditto all below

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
...
+
+    def text
+      @data.as(String)
+    end
+
+    def binary
+      @data.as(Bytes)
+    end
+
+    def initialize(@opcode : Opcode, @data : String | Bytes)
```
Sija: `@data` could be left nilable.

```suggestion
    def initialize(@opcode : Opcode, @data : String | Bytes | Nil = nil)
```

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
...
+    end
+  end
+
+  private def handle_opcodes(message)
```
Sija:
```suggestion
  private def handle_opcode(message)
```

src/http/web_socket.cr
zenithlight: Oops, forgot to change this when I changed the message to a struct.

src/http/web_socket.cr
zenithlight: Didn't know about this, thanks.

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
     @ws.close(message)
   end
 
-  def run
+  enum Opcode
```
straight-shoota: Why not reuse `HTTP::WebSocket::Protocol::Opcode` for this? `Protocol` is internal API, but you could simply move that enum here and use it in both classes. The `CONTINUATION` value is not needed for this high-level API, but I don't think having that is a huge issue.
But I don't know what others think about this.

src/http/web_socket.cr
```diff
@@ -100,52 +101,109 @@ class HTTP::WebSocket
...
+    Binary
+    Close
+    Ping
+    Pong
+  end
+
+  struct Message
+    getter opcode
+
+    def initialize(@opcode : Opcode, @data : String | Bytes | Nil = nil)
```
straight-shoota: In what case can `@data` be nil?

Sija: `:close` without a message?

straight-shoota: That should be empty String I think.

src/http/web_socket.cr
zenithlight: Actually, I was thinking about this too.

Mostly I was concerned about the `CONTINUATION` and the fact that the enum names are all-caps instead of title-cased. It makes it seem like it's for a low-level API (IMO). I'm not sure what the convention is within Crystal.

But I also like to reuse things if possible. Personally I think I lean more towards reusing the enum but wanted feedback from people more familiar with the language. Thanks for your input.

src/http/web_socket.cr
zenithlight: True, I think the data should always be an empty String or empty Bytes...

src/http/web_socket.cr
straight-shoota: > enum names are all-caps instead of title-cased.

There ist actually no style guide preferring either of them. And this could easily be adapted because the current enum is just internal API.

src/http/web_socket.cr
zenithlight: The `@data` should always have a value even if it's `""` so leaving it not-nilable.

src/http/web_socket.cr
zenithlight: I moved the enum into WebSocket class and imported it from the Protocol. I left it as all-caps in lieu of a style guide.

src/http/web_socket.cr
```diff
@@ -100,52 +110,101 @@ class HTTP::WebSocket
...
+  private def handle_opcode(message)
+    case message.opcode
+    when .close?
+      close(message.text) unless closed?
+      @close_reason = message.text
+    when .ping?
+      pong(message.text) unless closed?
+    end
+
+    if message.opcode.text? || message.opcode.binary?
```
Sija: This could be simplified to just `message.opcode.text? || message.opcode.binary?`. `return` is implicit anyway.

src/http/web_socket.cr
```diff
@@ -100,52 +110,101 @@ class HTTP::WebSocket
...
-      when Protocol::Opcode::CLOSE
-        @current_message.write @buffer[0, info.size]
-        if info.final
-          message = @current_message.to_s
-          @on_close.try &.call(message)
-          close(message) unless closed?
-          @current_message.clear
-          break
-        end
+      when .text?
```
Sija: you can group branches with common behavior:

```cr
when .text?, .close?, .ping?, .pong?
  message = Message.new(info.opcode, @current_message.to_s)
when .binary?
  message = Message.new(:binary, @current_message.to_slice)
end
```

ditto below

src/http/web_socket.cr
```diff
@@ -100,52 +110,101 @@ class HTTP::WebSocket
...
+      when .close?
+        message = Message.new(:close, @current_message.to_s)
+      when .ping?
+        message = Message.new(:ping, @current_message.to_s)
+      when .pong?
+        message = Message.new(:pong, @current_message.to_s)
+      end
+
+      @current_message.clear
+      return message.not_nil!
```
straight-shoota: I'd prefer a raising `else` branch (which would remove `Nil` type from `message`) instead of calling `not_nil!`.

asterite:
Could you provide a snippet of code of how you plan to use this?

z64:
Hi @zenithlight - I'd love to see this added, so thank you for getting it started.
Can you address the reviews above, or can someone else pick up this PR?

vlazar:
Typo in PR title. Needs to be `WebSocket: add blockable receive and receive_raw methods` (recieve -> receive)

straight-shoota:
Closing stale PR. There has been no activity in a long time. If someone wants to pick this up, they should start a new PR.
#5600 still tracks this feature.

PR #8034, opened 2019-08-02
Formatter: consider consecutive macro literals when subformatting

asterite:
The lexer might produce consecutive macro literals instead of just a single one, so when subformatting consider this case too.

Spotted by @waj


src/llvm/enums.cr
```diff
@@ -2,70 +2,70 @@ module LLVM
...
+      SwiftSelf                   = 1 << 48
+      UWTable                     = 1 << 49
+      WriteOnly                   = 1 << 50
+      ZExt                        = 1 << 51
 
       @@kind_ids = load_llvm_kinds_from_names.as(Hash(Attribute, UInt32))
 
       def each_kind(&block)
         return if value == 0
-        \{% for member in @type.constants %}
```
Sija: That removal seems like a bug.

asterite: Thanks, hopefully fixed! 🤞

src/compiler/crystal/tools/formatter.cr
```diff
@@ -2020,6 +2031,17 @@ module Crystal
       false
     end
 
+    # If we are formatting macro contents, if there are nested macro
+    # control structures they are definitely escaped with `\`,
+    # because otherwise we wouln't be able to format the contents.
+    # So here we appent those slashes. In theory the nesting can be
```
Sija: typo: append

src/compiler/crystal/tools/formatter.cr
```diff
@@ -2020,6 +2031,17 @@ module Crystal
...
+    # very deep but it's usually just one nesting.
```
Sija: nesting -> level

src/compiler/crystal/tools/formatter.cr
asterite: Thanks, fixed

src/compiler/crystal/tools/formatter.cr
asterite: Thanks, fixed

PR #5148, opened 2017-10-19
Make Array to be workable as a queue, ie respect push+shift pattern.

funny-falcon:
To achieve it, pointer to allocation is preserved, and @buffer is moved
instead of moving elements on `#shift`.

This is draft for #5126

PR #8036, opened 2019-08-02
Array: optimize shift

### LINKED ISSUE #5126: Array#shift is slow
I was using array shift a lot, but I found out it's pretty slow compared to using an index lookup

This is the current implementation

```crystal
  def shift
    if @size == 0
      yield
    else
      value = @buffer[0]
      @size -= 1
      @buffer.move_from(@buffer + 1, @size)
      (@buffer + @size).clear
      value
    end
  end
```

If I change this function by

```crystal
  def shift
    if @size == 0
      yield
    else
      value   = @buffer[0]
      @size   -= 1
      @buffer += 1
      value
    end
  end
```

It is a lot faster
Shifting Array(String).new(100_000, "a") shows the following speed improvement

```
shift        1.600000   0.000000   1.600000 (  1.602299)
each         0.000000   0.000000   0.000000 (  0.000487)
fast shift   0.000000   0.000000   0.000000 (  0.000530)
```

Question is, can it really be implemented like this or is there in important reason the buffer move and clear is required for this operation? I've made the change in the current master branch and tests all pass.

asterite:
Fixes #5126
Closes #5148

Some time ago I thought that making `shift` be efficient (O(1)) wasn't hard. The problem was that we needed to add an extra instance variable to `Array` for that and I didn't want to increase its size just to optimize a single method.

I turns out, Array's previous layout was like this:

```
  type_id    : Int32   # 4 bytes -|
  size       : Int32   # 4 bytes  |- packed as 8 bytes

  capacity   : Int32   # 4 bytes, aligned to 8 bytes

  buffer     : Pointer # 8 bytes  |- another 8 bytes
```

Because, at least in 64 bits, memory is aligned to 8 bytes, the `capacity` field was still taking 8 bytes. So... we still have 4 bytes left.

(and, well, for 32 bits it'll be a bit bigger, but I don't think by too much because the main bulk is the buffer)

My first idea was to include an `@first : Int32` instance variable which gets incremented on `shift`. Then replace `to_unsafe` with `@buffer + @first` and use `to_unsafe` everywhere instead of `@buffer.` That works but it turns out to be really slow to perform that math all the time.

Then I realized we can do something better: when we shift we can increment that variable but also increment `@buffer`! I called this new variable `@offset_to_buffer`, and doing `@buffer - @offset_to_buffer` points to the originally allocated buffer so we can reallocate it.

The diff isn't big but I added a lot of comments so it's clear. With this `shift` is optimized and also `insert 0, object`, `delete_at(0)`, `[]=(index, count, value)` when index is 0, and `unshift` (if previously shifted).

So we got it! `Array#shift` is efficient like in Ruby, and `Array` can be used like a queue or dequeue. `Deque` is still useful because of its circular nature, though.

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -47,6 +47,38 @@ class Array(T)
   include Indexable(T)
   include Comparable(Array)
 
+  # The size of this array.
+  @size : Int32
+
+  # The capacity of `@buffer`.
+  # Note that, because `@buffer` moves on shift, the actual
+  # capacity (the allocated memory) starts at `@buffer - @offset_to_buffer`.
+  # The actualy capacity is also given by the `remaining_capacity` internal method.
+  @capacity : Int32
+
+  # Offset to the buffer that was originally allocated, and which needs to
+  # be reallocated on resize. On shift this value gets increased, together with
+  # `@buffer`. To reach the root buffer you have to do `@buffer - @offset_to_buffer`,
+  # and this is also provided by the `root_buffer` internal method.
+  @offset_to_buffer : Int32 = 0
+
+  # The buffer where elements start.
+  @buffer : Pointer(T)
+
+  # In 64 bits the Array is composed then by:
+  # - type_id            : Int32   # 4 bytes -|
+  # - size               : Int32   # 4 bytes  |- packed as 8 bytes
+  #
+  # - capacity           : Int32   # 4 bytes -|
+  # - offset_to_buffer   : Int32   # 4 bytes  |- packed as 8 bytes
+  #
+  # - buffer             : Pointer # 8 bytes  |- another 8 bytes
+  #
+  # So in total 24 bytes. Without offset_to_buffer it's the same,
+  # because of aligning to 8 bytes (at least in 64 bits), and that's
+  # why we chose to include this value, because with it we can optimize
+  # `shift` to let Array be used as a queue/deque.
+
   # Size of an Array that we consider small to do linear scans
   # or other optimizations instead of using a lookup Hash.
   private SMALL_ARRAY_SIZE = 16
@@ -57,7 +89,6 @@ class Array(T)
   # [:foo, :bar].size # => 2
   # ```
   getter size : Int32
-  @capacity : Int32
 
   # Creates a new empty Array.
   def initialize
@@ -153,6 +184,13 @@ class Array(T)
     ary.size = (yield ary.to_unsafe).to_i
     ary
   end
+
+  # Returns the number of elements in the array.
+  #
+  # ```
+  # [:foo, :bar].size # => 2
+  # ```
+  getter size : Int32
 
   # Equality. Returns `true` if each element in `self` is equal to each
   # corresponding element in *other*.
@@ -410,9 +448,32 @@ class Array(T)
       @buffer[index] = value
     else
       diff = count - 1
-      (@buffer + index + 1).move_from(@buffer + index + count, size - index - count)
-      (@buffer + @size - diff).clear(diff)
-      @buffer[index] = value
+
+      # If index is 0 we can avoid a memcopy by doing a shift.
+      # For example if we have:
+      #
+      #    a = ['a', 'b', 'c', 'd']
+      #
+      # and someone does:
+      #
+      #    a[0..2] = 'x'
+      #
+      # we can change the value at 2 to 'x' and repoint `@offset_to_buffer`:
+      #
+      #    [-, -, 'x', 'd']
+      #           ^
+      #
+      # (we also have to clear the elements before that)
+      if index == 0
+        @buffer.clear(diff)
+        shift_buffer_by(diff)
+        @buffer.value = value
+      else
+        (@buffer + index + 1).move_from(@buffer + index + count, size - index - count)
+        (@buffer + @size - diff).clear(diff)
+        @buffer[index] = value
+      end
+
       @size -= diff
     end
 
@@ -642,26 +703,24 @@ class Array(T)
   # ```
   def concat(other : Array)
     other_size = other.size
-    new_size = size + other_size
-    if new_size > @capacity
-      resize_to_capacity(Math.pw2ceil(new_size))
-    end
+
+    resize_if_cant_insert(other_size)
 
     (@buffer + @size).copy_from(other.to_unsafe, other_size)
-    @size = new_size
+    @size += other_size
 
     self
   end
 
   # ditto
   def concat(other : Enumerable)
-    left_before_resize = @capacity - @size
+    left_before_resize = remaining_capacity - @size
     len = @size
     buf = @buffer + len
     other.each do |elem|
       if left_before_resize == 0
         double_capacity
-        left_before_resize = @capacity - len
+        left_before_resize = remaining_capacity - len
         buf = @buffer + len
       end
       buf.value = elem
@@ -704,10 +763,16 @@ class Array(T)
   def delete_at(index : Int)
     index = check_index_out_of_bounds index
 
+    # Deleting the first element is the same as a shift
+    if index == 0
+      return shift_when_not_empty
+    end
+
     elem = @buffer[index]
     (@buffer + index).move_from(@buffer + index + 1, size - index - 1)
     @size -= 1
     (@buffer + @size).clear
+
     elem
   end
 
@@ -763,7 +828,7 @@ class Array(T)
   # ary2 # => [[5, 2], [3, 4], [7, 8]]
   # ```
   def dup
-    Array(T).build(@capacity) do |buffer|
+    Array(T).build(@size) do |buffer|
       buffer.copy_from(@buffer, size)
       size
     end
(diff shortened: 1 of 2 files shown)


src/array.cr
```diff
@@ -1547,18 +1622,40 @@ class Array(T)
     shift { raise IndexError.new }
   end
 
+  # Removes the first value of `self`, at index 0, or otherwise invokes the given block.
+  # This method returns the removed value.
+  # If the array is empty, it invokes the given block.
```
straight-shoota: "... and returns it's value"

src/array.cr
```diff
@@ -1909,7 +2021,67 @@ class Array(T)
...
   private def check_needs_resize
-    double_capacity if @size == @capacity
+    # We have to compare against the actual capacity in case `@buffer` was moved
+    if @size == remaining_capacity
+      # If the array is not empty and more than half of the elements were shifted
+      # then we avoid a resize and just move the elements to the left.
+      # This is an heuristic. We could always try to move the elements if
+      # `@offset_to_buffer` is positive but it might happen that a user does
+      # `shift` + `push` in succession and it will produce a lot of memcopies.
+      if @capacity != 0 && @offset_to_buffer != 0 && @offset_to_buffer >= @capacity / 2
```
straight-shoota: `@offset_to_buffer != 0` is unnecessary.

asterite: It is if capacity is 1. Maybe I should mention it.

src/array.cr
asterite: Thanks, added!

src/array.cr
asterite: Clarified!

wdhwg001:
That's awesome! It would be better if you can attach some benchmark results, just for the convenience of those who come from a future changelog 😁

asterite:
I tried this:

```crystal
require "./old_array"

[5, 10, 20, 50, 100, 1_000, 10_000, 100_000, 200_000].each do |size|
  old_ary = OldArray.new(size, &.itself)
  new_ary = Array.new(size, &.itself)

  puts size
  puts "~" * 23

  time = Time.measure do
    while old_ary.shift?
    end
  end
  puts "Old: #{time}"

  time = Time.measure do
    while new_ary.shift?
    end
  end
  puts "New: #{time}"
  puts
end
```

Results:

```
5
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000000159
New: 00:00:00.000000080

10
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000000250
New: 00:00:00.000000067

20
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000000338
New: 00:00:00.000000096

50
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000000499
New: 00:00:00.000000141

100
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000000785
New: 00:00:00.000000247

1000
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.000022499
New: 00:00:00.000001696

10000
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.002324574
New: 00:00:00.000016691

100000
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:00.316955066
New: 00:00:00.000199766

200000
~~~~~~~~~~~~~~~~~~~~~~~
Old: 00:00:01.601885569
New: 00:00:00.000388967
```

At least for arrays of reasonable size the difference isn't that big, actually. But I guess the speed up is good anyway.

bcardiff:
I have doubts regarding this optimization. Is not that every Array will get shifted in its lifetime. Is having a more complex representation of array really worth it? Wouldn't be better to have a specialized structure for when shifting is expected to be used a lot?

asterite:
@bcardiff This optimization costs nothing: we just need to keep track of another integer, which doesn't change the total size of the Array in 64 bits architectures, and it's used in just a bunch of methods (most methods just deal with `@buffer` and that is always pointing to the correct place). With this optimization Array can be used as a queue or as a dequeue, making it ideal for any situation (we could even remove `Deque`, though that is a bit more performant in some situations).

I know there might be a chance that Array gets completely rewritten, but I don't see this optimization as something that makes Array's code a lot harder.

asterite:
@bcardiff Another way to see it: if you ever need `Array#shift` in the middle of other operations it's a shame if that's slow when having a fast implementation without any penalties (this one) exists. Or, well, the only penalty is a bit more code, but I don't think it's that complex.

RX14:
@asterite want to merge?

asterite:
@bcardiff was hesitant about this so I'll merge if he thinks this is okay

bcardiff:
I extracted some of the refactors I had in mind in Array in #8166. The conflict is that encapsulating the pointers in a propper buffer structure (with capacity) make it a little harder to keep the shift buffer as internal representation.

If we could get both things it would be great. Otherwise, depending on how Array might evolve for MT the optimization in shift might go away (temporarily (?))

RX14:
Please have a chat and work out how to merge both of these changes.

wontruefree:
Why is the `type_id` apart of the memory footprint?
I thought that would be removed at compile time.

asterite:
@wontruefree it's need for `is_a?` checks and resolving overloads at runtime. At least when it's a class type with children.

wontruefree:
Thanks for explaining it to me.  I didnt know you could overload at runtime I thought it was only available to overload at compile time.

asterite:
You can't overload at runtime. But at runtime, if you have:

```crystal
class Foo
end

class Bar
end

def method(x : Foo)
end

def method(x : Bar)
end

obj = rand < 0.5 ? Foo.new : Bar.new
method(obj)
```

at runtime the type_id of `obj` is checked and if it's the one for `Foo` then the first overload is called, otherwise the one for `Bar` is called.

wontruefree:
I think I remember someone opening an issue for this but I didn't know that is how it was solved.

src/array.cr
```diff
@@ -154,6 +185,13 @@ class Array(T)
     ary
   end
 
+  # Returns the number of elements in the array.
+  #
+  # ```
+  # [:foo, :bar].size # => 2
+  # ```
+  getter size : Int32
```
BrianHawley: Isn't this code also declared earlier in the same file and context? Same comment too.

asterite: You are right! I always forget to implement an error check in the compiler for this case...

jwoertink:
@asterite @bcardiff Here's another PR that looks like it's been approved by a few core team. Maybe this can be merged in or tagged for next release too? 😄

bcardiff:
There is still some conflict with how array could evolve regarding MT. And that depends on unsolved integration with the GC to avoid needing to clear the memory of unneeded elements.

#8166 introduce some of the initial work to do that, although without accomplishing a [successful integration with the GC](https://github.com/ivmai/bdwgc/issues/290) there is no strong reason to merge it.

asterite:
Could we maybe merge this so that people benefit from the performance improvement for a while? We could always revert this and go with the MT approach later on.

bcardiff:
I would let the rest chime in. My opinions were stated :-)

Note: there is a [pending comment](https://github.com/crystal-lang/crystal/pull/8036#discussion_r329296336) and of course a rebase in case a (potential temporal?) merge is preferred.

RX14:
I don't think it's fair to punt a PR infinitely down the line based on some exploratory work, but #8166 is already a PR. It'd be fair to discuss and vote on how to integrate the approaches, or choose one if that fails.

asterite:
@bcardiff Can we move forward with this for 1.0?

This is what I think: it seems there are no plans to make Array fiber-safe before 1.0. So, we can optimize `shift` now and then Array becomes the perfect collection, efficiently usable as a queue or as a stack (like in Ruby). If after 1.0 we decide to make changes to Array and this optimization can't be done, we drop it. BUT, if that never happens then we can benefit from this right now and forever.

Like @RX14 said, I don't think it's fair to remove this optimization just because of hypothetical cases. And personally, I don't think we'll ever make these collections fiber-safe by default because it will make them extremely slow for the general case.

bcardiff:
> make these collections fiber-safe by default because it will make them extremely slow for the general case

It depends on the guarantee we want. If `a = [];  a << 1 /in paralell with/ a << 2` is such that either insertion will happend (but not necessarily both) there is a way to make it efficient and wait-free. That was wait I was aiming but got stuck with the GC integration.

I think this optimization can come after 1.0 either way, I will not stop if you want to send it before. The `CrystalArraySyntheticProvider` in `crystal_formatters.py` should also be updated at that point.

But if we don't contemplate that the optimization can go away in the future, I am not sure we will be free to pick the best guarantee we could for MT. Maybe we still can; I haven't done the work to extend the idea of the wait-free algorithm to this new representation.

I also think that in the presence of Deque there less benefit of this optimization. Adding new operations to Array becomes more error prone as the structure becomes more complex.

PR #8040, opened 2019-08-04
Compiler: fix sizeof Nil and other empty types

### LINKED ISSUE #7644: Shifting a `Deque(Nil)` more times than its initial capacity crashes the program
I'm having a `Deque` with initial capacity of `N` and try to push `N + 1` values to read and then read them one-by-one.

This simple program works as expected (https://carc.in/#/r/6obc):

```crystal
deque = Deque(Bool).new(1)

2.times do
  deque << true
end

2.times do
  deque.shift
end

puts "ok"
```

Howerver, if using `Nil` as the deque's type, it crashes (https://carc.in/#/r/6obd):

```crystal
deque = Deque(Nil).new(1)

2.times do
  deque << nil
end

2.times do
  deque.shift
end

puts "ok"
```

```
```

The last offending line I've found is [this](https://github.com/crystal-lang/crystal/blob/60760a5460aa13a1f80c5a1a7410782dc4076b77/src/deque.cr#L531) — `Pointer(Nil).realloc` may lead to the bug. The issue happens with `Nil` only. Other types, as well as custom structs and classes, work as expected.

asterite:
Fixes #7644

### DIFF
--- a/spec/compiler/codegen/sizeof_spec.cr
+++ b/spec/compiler/codegen/sizeof_spec.cr
@@ -80,13 +80,19 @@ describe "Code gen: sizeof" do
   end
 
   it "gets sizeof Void" do
-    # Same as the size of a byte
+    # Same as the size of a byte, because doing
+    # `Pointer(Void).malloc` must work like `Pointer(UInt8).malloc`
     run("sizeof(Void)").to_i.should eq(1)
   end
 
   it "gets sizeof NoReturn" do
-    # Same as the size of a byte
-    run("sizeof(NoReturn)").to_i.should eq(1)
+    # NoReturn can't hold anything
+    run("sizeof(NoReturn)").to_i.should eq(0)
+  end
+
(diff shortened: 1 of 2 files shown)


asterite:
I forgot to explain this about more: `Pointer(T).malloc` is implemented as a primitive, and there we delegate the allocated size to LLVM using its `sizeof` function, which for `Nil` (and other "empty" types returns 0). So mallocing a Nil pointer always asks for 0 memory. But when doing `Pointer#clear` we use our `sizeof`, which calls LLVM's once and then would change it to 1 if it's zero. That caused a segfault trying to access something past that pointer. I guess I made it like that for the `Pointer(Void)` and `sizeof(Void)` case, but I generalized too much.

I'm still not sure about us being able to do `Pointer(Void)` because it's just so it looks like what you can do with C, but at least this fixes the issue without introducing any breaking change.

PR #8045, opened 2019-08-05
Int#divisible_by? doesn't care about the value being floored.

yxhuvud:
The if statement in % might be optimized away by LLVM, but why
force an optimization if the code isn't harder to understand when doing the optimized thing?


straight-shoota:
Thank you @yxhuvud :)

PR #8046, opened 2019-08-05
Remove standard library paths compiler

dburnsii:
Defining the Linux library paths without having an option to ignore them with compiling creates potential cross-compilation issues when using tools like Yocto, that attempt to insulate the build environment from the working environment. Linux should add these paths by default, so the vast majority of builds should still run just fine.

### DIFF
--- a/src/compiler/crystal/codegen/link.cr
+++ b/src/compiler/crystal/codegen/link.cr
@@ -105,7 +105,7 @@ module Crystal
     end
 
     private def lib_flags_posix
-      library_path = ["/usr/lib", "/usr/local/lib"]
+      library_path = [] of String
       has_pkg_config = nil
 
       String.build do |flags|


ysbaddaden:
I understand the concern, but there isn't just Linux out there, but many UNIX systems. These folders aren't necessarily added by the system (they're not on FreeBSD for example) despite having some required libs being installed at those locations. Removing these defaults will break building on those systems.

straight-shoota:
Thanks for bringing this up @dburnsii, though!
As @ysbaddaden explained, this fix won't work on all platforms and thus can't be implemented like this. We might look for a way to add these defaults on platforms that needs them, but this might be a more complex topic.
I'd suggest to open an issue detailing the original problem, possibly including a reproducible example with `yocto` or something similar. This enables us to look for solutions in broad.

ysbaddaden:
It's not complex to fix, but it must be tested on a number of platforms to make sure that it works smoothly everywhere, which is complex.

ysbaddaden:
@RX14 it's a bold assumption; the FreeBSD linker didn't include `/usr/local` *by default* despite ports installing libraries there. It can even be possible for the `/usr/local` pkgconfig directory to not be used by default. It's not misconfiguration.

RX14:
If it's not misconfiguration what is it? How does the C tool chain know to look in /usr/local? The C compilers do not have this ugly hack which prevents cross-linking.

If crystal should look in /usr/local (I presume this is where the ports system installs stuff to, which is why it's not looked at by default) the system or user must have a way of telling crystal to do that.

RX14:
I'm sure this change will break some systems, but it is up to distro packagers to work out how to solve this in the idiomatic way for their operating system. These hacks to "make things work" prevent other things working. 

If it turns out the answer is for ports to hard code these paths just for freebsd with a `CRYSTAL_CONFIG_DEFAULT_LINKER_PATHS` that's a good patch. But we need to emulate how the C tool chain operates since we are linking to C libraries.

asterite:
Why don't we make it configurable (an env var?) instead of possibly breaking things?

dburnsii:
I thought making it configurable would be a good idea, but at the same time there are other switches I'd also have to use (for GCC for instance, I had to add -nostdlib), that I think make adding this redundant. It sounds like there are setups that won't add these as defaults though, so could the switch be to enable them rather than having them enabled by default?

RX14:
This PR should have been merged. The freebsd `/usr/local` mess is [well documented](https://wiki.freebsd.org/WarnerLosh/UsrLocal) and *not our problem*. The fix is either to edit the crystal port to use `CC=gcc` or edit the crystal port to always pass link flags.

I'm not aware of any other system where this is an issue.

ysbaddaden:
I agree. It's much better than introducing a CLI argument to disable this. It looked harmless at first, but itnintroduces issues in valid use cases.

The compiler shell wrapper script can specify `LIBRARY_PATH` and/or `LD_LIBRARY_PATH` when needed on some platform.

RX14:
I'll re-submit this PR or similar then. We can add more knobs for packagers as requires.

bcardiff:
@RX14 The knobs you mention are adding `CRYSTAL_CONFIG_LINK_FLAGS` at build time and `CRYSTAL_LINK_FLAGS` so default libs can be overridden when bundling the compiler and via de wrapper script, right? Ref: https://github.com/crystal-lang/crystal/pull/8326#issuecomment-547117112

ysbaddaden:
FreeBSD can use a wrapper shell script that adds `/usr/lib:/usr/local/lib` to LIBRARY_PATH` so CC then LD will find system installed libraries. It doesn't need anything more.

RX14:
@bcardiff maybe, but they're not the correct solution here. The problem with compiling those in is that it hinders the compiler's ability to cross-compile for other platforms. @ysbaddaden has the correct solution for this case.

Issue #8043, opened 2019-08-05
Iterator idempotence


KimBurgess:
The `Iterator` module has a number of methods inherited from `Enumerable` that when invoked result in the original iterator being exhausted. This can lead to some confusing behaviour when first encountered.

[For example](https://play.crystal-lang.org/#/r/7csc):
```crystal
iter = (0...100).each

puts iter.size # => 100

puts iter.to_a # => []

puts iter.size # => 0
```

Similar behaviour also results from calls to `#all?`, `#any?`, `#count`, `#none?` and `#one?` due to internal calls to `#each`, which forces full evaluation of the original sequence.

Two possible solutions may exist here:
1. Make note of this behaviour in the docs to help reduce [WTFs/minute](https://www.osnews.com/story/19266/wtfsm/).
2. Override these methods within `Iterator` so that calls become idempotent. This would likely require internal caching of generated elements so that subsequent calls involving `#each` are yielded the originally generated objects. There's an obvious resource impact here as well as change to existing API behaviour so this comes with it's own challenges.

Happy to put together a PR for either of these, but would be interested in some feedback on the preferred approach.

straight-shoota:
Implicitly caching an iterator is not an option as it entirely defeats the purpose of an iterator (which can have a potentially unlimited size). When such behaviour is needed, it should be implemented by deliberate choice, for example by solidifying into an array.

So we can only improving the documentation on the behaviour related to re-iterating an iterator.

KimBurgess:
Possible counterpoint to that is full evaluation and caching does already take place with calls it `Iterator#cycle`. But I do completely agree that this is dangerous territory.

straight-shoota:
`#cycle` is a single method and caches only for its internal behaviour.

PR #8053, opened 2019-08-06
Docs on idempotence and methods involving eager evaluation in Iterators

KimBurgess:
Adds notes on underlying single pass / stream behaviour of Iterators and associated implications when using them.

Results of discussion in #8043.

### DIFF
--- a/src/iterator.cr
+++ b/src/iterator.cr
@@ -29,6 +29,16 @@ require "./enumerable"
 # (1..10_000_000).each.select(&.even?).map { |x| x * 3 }.first(3).to_a # => [6, 12, 18]
 # ```
 #
+# Because iterators only go forward, when using methods that consume it entirely or partially –
+# `to_a`, `any?`, `count`, `none?`, `one?` and `size` – subsequent calls will give a different
+# result as there will be less elements to consume.
+#
+# ```
+# iter = (0...100).each
+# iter.size # => 100
+# iter.size # => 0
+# ```
+#
 # To implement an `Iterator` you need to define a `next` method that must return the next
 # element in the sequence or `Iterator::Stop::INSTANCE`, which signals the end of the sequence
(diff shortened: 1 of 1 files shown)


src/iterator.cr
```diff
@@ -29,6 +29,18 @@ require "./enumerable"
 # (1..10_000_000).each.select(&.even?).map { |x| x * 3 }.first(3).to_a # => [6, 12, 18]
 # ```
 #
+# Due to the lazy computation provided by `Iterator`, emmitted objects may be considered as a
+# potentially infinite stream. Calls that pass to the underlying `Enumerable` module which force
+# eager evaluation of the complete sequence should be used with care. These include `#all?`,
```
asterite: > should be used with care

I don't think we should let people think that using iterators is dangerous or might lead to bugs. I would simply add a comment like this:

> Because iterators only go forward, when using methods that consume it entirely or partially, like `to_a`, `all?`, `any?` and `size`, subsequent calls to these methods will give a different result because there will be less elements to consume.

The example below is good though 👍

KimBurgess: Good call. I really like that wording too!

KimBurgess:
Turns out git works better if I actually add the changes before committing...

src/iterator.cr
```diff
@@ -29,6 +29,16 @@ require "./enumerable"
 # (1..10_000_000).each.select(&.even?).map { |x| x * 3 }.first(3).to_a # => [6, 12, 18]
 # ```
 #
+# Because iterators only go forward, when using methods that consume it entirely or partially -
```
straight-shoota: OCD: Could you use en dash please? Same in the following line.
```suggestion
# Because iterators only go forward, when using methods that consume it entirely or partially –
```

KimBurgess: I appreciate the OCD :). Was going to pop that in, but wasn't what the feeling towards unicode chars were.

PR #8055, opened 2019-08-06
Compiler: fix crash when formatting syntax error inside macro

### LINKED ISSUE #8038: Unhandled compiler exception: Empty enumerable (Enumerable::EmptyError)
```cr
{% begin %}
HASH = {
  {% for i in 0..10 %}
    "hello" => "hello",
  {% end %},
}
{% end %}
```
```
Unhandled exception: Empty enumerable (Enumerable::EmptyError)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
```
Before 0.30.0 (0.29.0) this resulted in a proper error:
```
Error in line 1: macro didn't expand to a valid program, it expanded to:

   1. 
   2. HASH = {
   3.   
   4.     "hello" => "hello",
   5.   
   6.     "hello" => "hello",
   7.   
   8.     "hello" => "hello",
   9.   
  10.     "hello" => "hello",
  11.   
  12.     "hello" => "hello",
  13.   
  14.     "hello" => "hello",
  15.   
  16.     "hello" => "hello",
  17.   
  18.     "hello" => "hello",
  19.   
  20.     "hello" => "hello",
  21.   
  22.     "hello" => "hello",
  23.   
  24.     "hello" => "hello",
  25.   ,
  26. }
Syntax error in expanded macro: macro_94113831048960:2: unterminated hash literal

HASH = {
       ^

```
But now there's only a compiler crash.

asterite:
Fixes #8038

Also improves the code a bit.

💭  I wonder whether we should make `min`, `max`, etc., always be nilable... that way you would always remember to handle the empty case, and in case you are sure you aren't dealing with an empty collection you can always stick a `not_nil!` there.

### DIFF
--- a/src/compiler/crystal/exception.cr
+++ b/src/compiler/crystal/exception.cr
@@ -242,20 +242,25 @@ module Crystal
 
     def minimize_indentation(source)
       min_leading_white_space =
-        source.map do |line|
-          if match = line.match(/^(\s+)\S/)
-            spaces = match[1]?
-            spaces.size if spaces
-          end
-        end
-          .compact
-          .min
+        source.min_of? { |line| leading_white_space(line) } || 0
 
-      source = source.map do |line|
-        replace_leading_tabs_with_spaces(line).lchop(" " * min_leading_white_space)
+      if min_leading_white_space > 0
+        source = source.map do |line|
+          replace_leading_tabs_with_spaces(line).lchop(" " * min_leading_white_space)
+        end
       end
 
       {source, min_leading_white_space}
+    end
+
+    private def leading_white_space(line)
+      match = line.match(/^(\s+)\S/)
+      return 0 unless match
+
+      spaces = match[1]?
+      return 0 unless spaces
+
+      spaces.size
     end
 
     def append_expanded_macro(io, source)
--- a/spec/compiler/semantic/macro_spec.cr
+++ b/spec/compiler/semantic/macro_spec.cr
@@ -1427,4 +1427,12 @@ describe "Semantic: macro" do
       ),
       "can't instantiate abstract class Foo"
   end
+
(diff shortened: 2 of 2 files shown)


bcardiff:
Given that an enumerable can be empty as a value, it could make sense to min/max be nilable, and have min!/max! as an unsafe “trust-me-is-not-empty” version. But on other hand the operation itself is not defined for empty sets, so is reasonable to have a non trivial precondition.

oprypin:
>  I wonder whether we should make min, max, etc., always be nilable

That's the kind of thing that causes a knee-jerk negative reaction for me.
But then if I actually think about it, it totally is an improvement.

> always remember to handle the empty case

bcardiff:
The linux32 CI is failing constantly this time. Maybe #8065 could help

src/compiler/crystal/exception.cr
```diff
@@ -242,22 +242,27 @@ module Crystal
...
     end
 
+    private def leading_white_space(line)
+      match = line.match(/^(\s+)\S/)
+      return 0 unless match
+
+      spaces = match[1]?
+      return 0 unless spaces
+
+      spaces.size
```
j8r: Since this part is refactored, we can simplify it by avoiding using a regex:
```cr
line.each_char_with_index do |char, index|
  return index if char != ' '
end
0
```

asterite: No, `\s` matches tab, spaces, maybe also unicode space, I don't know.

This code runs once when the compiler finds an error, optimizing here has no impact at all. Maybe it's 5ms faster, the user can't possibly tell.

I also prefer avoiding changing code if it's not broken.

j8r: Ho, there is `char.ascii_whitespace?`

But that's basically what's done here, the code is already changed.

PR #8032, opened 2019-08-02
Update CI to use 0.30.1

bcardiff:

PR #8056, opened 2019-08-07
Split http/server_spec

straight-shoota:
This PR extracts the specs for `HTTP::Server::RequestProcessor` and `HTTP::Server::Response` from `http/server_spec.cr` and puts them into separate files to improve readability. They're both individual components and the specs have 200+ lines each.

It turns out the spec file has been so convoluted that the `#remote_address` specs incorrectly landed in the example group for `RequestProcessor` instead of `Server`. So splitting is definitely a good idea.

### DIFF
--- a/spec/std/http/server/request_processor_spec.cr
+++ b/spec/std/http/server/request_processor_spec.cr
@@ -1,0 +1,267 @@
+require "spec"
+require "http/server/request_processor"
+
+private class RaiseErrno < IO
+  def initialize(@value : Int32)
+  end
+
+  def read(slice : Bytes)
+    Errno.value = @value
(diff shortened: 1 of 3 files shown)


straight-shoota:
Formatter fails because CI still runs with 0.29.0 (#8032)

asterite:
@straight-shoota I think you might need to run the formatter with a compiled compiler (it should now correctly format code inside top-level macros without expanded expressions)

straight-shoota:
Oh, it's a post-0.30.0 change. I misread the output.

straight-shoota:
Rebased on master with #8032 to see test_linux32 succeed.

PR #8061, opened 2019-08-08
HTTP::Request: store common header names in a normalized way

### LINKED ISSUE #8060: HTTP::Headers gets downcase
This shows it perfectly: 
```crystal
require "http"

headers = <<-EOF
Content-Length: 0
User-Agent: Me!12
\n
EOF

io_mem = IO::Memory.new
HTTP.parse_headers_and_body(IO::Memory.new(headers)) do |headers, body|
  HTTP.serialize_headers(io_mem, headers)
  pp body
end

io_mem.to_s # => "content-length: 0\r\nuser-agent: Me!12\r\n\r\n"
```

The expected result is to get back the same "case" for the headers, but instead it gets downcased

asterite:
Fixes #8060 

Doesn't actually fix it because we aren't storing exactly what was sent, but since headers are case insensitive this shouldn't be a problem. I think the "normalized" way is pretty common and used everywhere so in most cases this way will match what was sent.

If this is a real problem, though, we can revert this optimization.

### DIFF
--- a/spec/std/http/request_spec.cr
+++ b/spec/std/http/request_spec.cr
@@ -215,6 +215,21 @@ module HTTP
         request.should eq HTTP::Status::BAD_REQUEST
       end
 
+      it "stores normalized case for common header name (lowercase) (#8060)" do
+        request = Request.from_io(IO::Memory.new("GET / HTTP/1.1\r\ncontent-type: foo\r\n\r\n")).as(Request)
+        request.headers.to_s.should eq(%(HTTP::Headers{"content-type" => "foo"}))
+      end
+
+      it "stores normalized case for common header name (capitalized) (#8060)" do
+        request = Request.from_io(IO::Memory.new("GET / HTTP/1.1\r\nContent-Type: foo\r\n\r\n")).as(Request)
+        request.headers.to_s.should eq(%(HTTP::Headers{"Content-Type" => "foo"}))
+      end
+
+      it "stores normalized case for common header name (mixed) (#8060)" do
+        request = Request.from_io(IO::Memory.new("GET / HTTP/1.1\r\nContent-type: foo\r\n\r\n")).as(Request)
+        request.headers.to_s.should eq(%(HTTP::Headers{"Content-type" => "foo"}))
+      end
+
       describe "long request lines" do
         it "handles long URI" do
           path = "a" * 8177
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -185,9 +185,37 @@ module HTTP
     {name, value}
   end
 
+  # Important! These have to be in lexicographic order.
   private COMMON_HEADERS = %w(
-    accept-language
+    Accept-Encoding
+    Accept-Language
+    Accept-encoding
+    Accept-language
+    Allow
+    Cache-Control
+    Cache-control
+    Connection
+    Content-Disposition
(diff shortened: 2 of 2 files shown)


spec/std/http/request_spec.cr
```diff
@@ -215,6 +215,11 @@ module HTTP
         request.should eq HTTP::Status::BAD_REQUEST
       end
 
+      it "is stores normalized case for common header name (#8060)" do
+        request = Request.from_io(IO::Memory.new("GET / HTTP/1.1\r\nconetnt-type: foo\r\n\r\n")).as(Request)
```
Sija:
```suggestion
        request = Request.from_io(IO::Memory.new("GET / HTTP/1.1\r\ncontent-type: foo\r\n\r\n")).as(Request)
```

asterite: Thanks! I think I was running a different, spec, this one actually fails 😊

spec/std/http/request_spec.cr
```diff
@@ -215,6 +215,11 @@ module HTTP
         request.should eq HTTP::Status::BAD_REQUEST
       end
 
+      it "is stores normalized case for common header name (#8060)" do
```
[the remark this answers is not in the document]
asterite: Thanks!

ysbaddaden:
From HTTP/2 and onwards the normalized form of HTTP headers is... downcased. And HTTP/2 is much more common that it would seem. We should embrace the future (that started yesterday) and just store headers lowercased, we're only going to receive them more and more downcased, even in HTTP/1.

asterite:
Updated!

I think we could later use a binary search to optimize this but doing it with slices proved to be a bit hard for me when I tried to timebox it.

asterite:
Done with binary search :-)

(I think I got it right)

j8r:
Ho, @ysbaddaden is right. For example in GitHub response headers are all downcased:
```
HTTP/2.0 200 OK
access-control-allow-methods: GET
access-control-max-age: 3600
last-modified: Fri, 09 Aug 2019 18:56:47 GMT
etag: "d6756a52bf468deae45841b35756e6d8"
content-type: text/css
server: AmazonS3
content-encoding: gzip
via: 1.1 varnish
accept-ranges: bytes
date: Fri, 09 Aug 2019 20:46:48 GMT
via: 1.1 varnish
age: 2890
x-served-by: cache-iad2126-IAD, cache-cdg20764-CDG
x-cache: HIT, HIT
x-cache-hits: 1, 340
x-timer: S1565383608.050355,VS0,VE0
vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method, Accept-Encoding
access-control-allow-origin: *
x-fastly-request-id: b7cdd43b9564b5ec584b54bfde1b8233512b2697
content-length: 75635
X-Firefox-Spdy: h2
```
But my request headers are in camelcase with dashes.

straight-shoota:
@j8r Yes, this is likely the way we're going to move forward. But it's not the point here. We don't want to introduce a breaking change at this point. On the contrary, this PR is supposed to be a bugfix in 0.30.1 for the behaviour that was unintentionally broken in 0.30.0.

The idea to normalize all headers downcased should be discussed in a separate issue and has a broader scope than just `parse_header`.

PR #8065, opened 2019-08-09
Split std_specs in 32 bits ci

bcardiff:
Only for std_specs run against the compiled compiler which seems to be the one failing

This introduces a way to run a subset of the specs via an env var `SPEC_SPLIT`. Is not expected to be used as a public API but is a workaround to help a bit the CI.

Let's see how it works...

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -77,8 +77,19 @@ build() {
 
   case $ARCH in
     i386)
-      with_build_env 'make crystal std_spec threads=1'
-      with_build_env 'CRYSTAL_SPEC_COMPILER_THREADS=1 make compiler_spec docs threads=1'
+      with_build_env 'make crystal threads=1'
+      with_build_env 'SPEC_SPLIT="0%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="1%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="2%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="3%4" make std_spec threads=1'
+
+      parts=16
+      i=0
+      while [ $i -lt $parts ]; do
+        with_build_env "CRYSTAL_SPEC_COMPILER_THREADS=1 SPEC_SPLIT=\"$i%$parts\" make compiler_spec threads=1"
+        i=$((i + 1))
+      done
+
       with_build_env 'make docs threads=1'
       ;;
     *)
--- a/bin/ci
+++ b/bin/ci
@@ -74,7 +74,18 @@ prepare_system() {
 
 build() {
   with_build_env 'make std_spec clean threads=1'
-  with_build_env 'make crystal std_spec compiler_spec docs threads=1'
+
+  case $ARCH in
+    i386)
+      with_build_env 'make crystal std_spec threads=1'
+      with_build_env 'CRYSTAL_SPEC_COMPILER_THREADS=1 make compiler_spec docs threads=1'
+      with_build_env 'make docs threads=1'
+      ;;
+    *)
+      with_build_env 'make crystal std_spec compiler_spec docs threads=1'
+      ;;
+  esac
+
   with_build_env 'find samples -name "*.cr" | xargs -L 1 ./bin/crystal build --no-codegen'
 }
 
--- a/src/spec/dsl.cr
+++ b/src/spec/dsl.cr
@@ -109,6 +109,30 @@ module Spec
     lines << line
   end
 
+  @@split_filter : NamedTuple(remainder: Int32, quotient: Int32)? = nil
+
+  def self.add_split_filter(filter)
+    if filter
+      r, m = filter.split('%').map &.to_i
+      @@split_filter = {remainder: r, quotient: m}
+    else
+      @@split_filter = nil
+    end
+  end
+
+  @@spec_counter = -1
+
+  def self.split_filter_matches
+    split_filter = @@split_filter
+
+    if split_filter
+      @@spec_counter += 1
+      @@spec_counter % split_filter[:quotient] == split_filter[:remainder]
+    else
+      true
+    end
+  end
+
   # :nodoc:
   def self.matches?(description, file, line, end_line = line)
     spec_pattern = @@pattern
--- a/spec/compiler/codegen/private_spec.cr
+++ b/spec/compiler/codegen/private_spec.cr
@@ -2,7 +2,7 @@ require "../../spec_helper"
 
 describe "Codegen: private" do
   it "codegens private def in same file" do
-    compiler = Compiler.new
+    compiler = create_spec_compiler
     sources = [
       Compiler::Source.new("foo.cr", %(
                                         private def foo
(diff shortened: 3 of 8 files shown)


bin/ci
```diff
@@ -74,7 +74,21 @@ prepare_system() {
...
+  case $ARCH in
+    x86_64)
+      with_build_env 'make crystal std_spec compiler_spec docs threads=1'
+      ;;
+    i386)
+      with_build_env 'make crystal threads=1'
+      with_build_env 'SPEC_SPLIT="0%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="1%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="2%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="0%4" make std_spec threads=1'
```
asterite: Shouldn't this be `3%4`?

src/spec/dsl.cr
```diff
@@ -109,11 +109,27 @@ module Spec
     lines << line
   end
 
+  @@split_filter : NamedTuple(reminder: Int32, quotient: Int32)? = nil
```
Sija:
```suggestion
  @@split_filter : NamedTuple(remainder: Int32, quotient: Int32)? = nil
```

bcardiff:
Still. failing, but I noticed, the ci uses `threads=1` but for the compilers run in the compiler_spec `n_threads` will still be `8`. Could that be something impacting this @asterite?

bin/ci
```diff
@@ -74,7 +74,21 @@ prepare_system() {
...
+    x86_64)
+      with_build_env 'make crystal std_spec compiler_spec docs threads=1'
+      ;;
+    i386)
+      with_build_env 'make crystal threads=1'
+      with_build_env 'SPEC_SPLIT="0%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="1%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="2%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="3%4" make std_spec threads=1'
+      with_build_env 'make compiler_spec docs threads=1'
```
asterite: You should use `SPEC_SPLIT` here too. This is where it's failing.

straight-shoota:
How's this supposed to work? If I understand the issue correctly, the compiler runs out of memory while compiling the specs. Unless I'm missing some key concept here, changing the logic in `Spec.matches?` only changes which specs are executed (at runtime). But they're still compiled into the spec binary. In fact, this is now running the compiler four times.

We don't need to split spec execution but already spec compilation.

asterite:
@straight-shoota As far as I can tell from CI's output, it's failing when running the specs, not when compiling them.

src/spec/dsl.cr
```diff
@@ -109,11 +109,27 @@ module Spec
...
+  end
+
   # :nodoc:
   def self.matches?(description, file, line, end_line = line)
     spec_pattern = @@pattern
     spec_line = @@line
     locations = @@locations
+    split_filter = @@split_filter
+
+    if split_filter && line % split_filter[:quotient] != split_filter[:remainder]
```
asterite: The way I thought about `SPLIT_SPEC`, each time you do `it` a counter would get incremented, and it's this counter you would check against.

By using just the line number it might happen that more specs are in lines 1 modulo 4, for example.

But if it works then it's a good start :-)

asterite:
Maybe we can try running all specs (like this, splitted) with `-v` to see in which one it fails. Maybe there's a bunch of specs that allocate too much memory and they are the only problem... 🤔

src/spec/dsl.cr
bcardiff: Yup, that's another way to filter them. The distribution of the std_spec at least was good enough.

```
$ make std_spec
7744 examples, 0 failures, 0 errors, 8 pending

$ SPEC_SPLIT="0%4" make std_spec
1958 examples, 0 failures, 0 errors, 2 pending

$ SPEC_SPLIT="1%4" make std_spec
1856 examples, 0 failures, 0 errors, 2 pending

$ SPEC_SPLIT="2%4" make std_spec
1958 examples, 0 failures, 0 errors, 2 pending

$ SPEC_SPLIT="3%4" make std_spec
1972 examples, 0 failures, 0 errors, 2 pending
```

straight-shoota:
Hm, yeah it seems you're right. Compiling the specs is not the issue. But when it's at runtime, this might actually mean there are some memory leaks. So `-v` and perhaps a GC status report between examples would help identify the issues.

asterite:
@straight-shoota Oh, I know for sure there's a memory leak. When we compiler specs without the prelude the `malloc` that's used when you do `SomeClass.new` or `Pointer.malloc` is C malloc, and there's no matching `free` for that (there's no GC), so all specs that run without a prelude that create classes or allocate memory through `Pointer` will currently leak that memory. And we have a **a lot** of those tests.

Unfortunately I can't think of an easy way to fix that.

straight-shoota:
@asterite I see. There is the option to disable JIT compilation and run specs without prelude in a separate process as well. So it's a trade-off between speed and memory. I'm not sure how much a difference it makes, is it worth looking into that? But anyway, this is just relevant for specs and not a critical problem. So as long as we get the 32-bit specs to work by splitting them up, I guess this should be fine.

straight-shoota:
Another idea, just from the top of the head: Maybe we could somehow rig the `malloc` calls inside the JIT-executed spec code. So that we can keep track of the memory and free it after the spec has completed. Codegen can probably be changed for that, so it would be a little bit different code from normal execution, but that might be acceptable.

j8r:
The issue isn't just for 32bits, in general too much memory is used for tests. It would be better to fix the root cause instead. The memory in the CI isn't infinite, and the limit might be eventually reached in the future even for 64bit. It also impacts devs who are using machines with not much RAM.

asterite:
j8r PRs and ideas are welcome. This just affects us Crystal developers. I think this fix is fine and it let's us move forward.

bin/ci
```diff
@@ -74,7 +74,28 @@ prepare_system() {
...
+      with_build_env 'make crystal threads=1'
+      with_build_env 'SPEC_SPLIT="0%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="1%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="2%4" make std_spec threads=1'
+      with_build_env 'SPEC_SPLIT="3%4" make std_spec threads=1'
+
+      parts=16
+      for i in $(seq $parts)
+      do
+        with_build_env "CRYSTAL_SPEC_COMPILER_THREADS=1 SPEC_SPLIT=\"$((i-1))%$parts\" make compiler_spec threads=1"
```
j8r: That's strange, a while loop would be clearer (no need to use seq nor `i -1`)

bcardiff: my bash-fu is limited.

bcardiff:
I'm pushing temporarily @asterite's commit to check if that fix the ci. Other than that the PR is ready for review and squash

bin/ci
j8r: A classic counter:
```
i=0
while [ $i -lt $parts ]; do
  ...
  i=$((i + 1))
done
```

Anyway, wouldn't it be better to have this logic in the makefile, in order to make it more generic and use it outside the CI context?

bcardiff: seq replaced by while.

The whole goal of this is for the CI. I don't spec the index filter to become a public API.

PR #8066, opened 2019-08-09
Fix ipv6 hostname connection

MrSorcus:


straight-shoota:
Thanks :heart: Can you please add a spec for this?

PR #8067, opened 2019-08-09
Avoid lazy inizialization of class variables assigned with simple values

waj:
Currently all class variables are lazily initialized even when they are declared with a simple initializer with a literal of a simple type. This prevents that and do a direct initialization of the global variable, quite similar to simple constants.

I'm not so happy with this patch as I do with the result, so I'm open to suggestions about how to write this better.

### DIFF
--- a/spec/compiler/codegen/class_var_spec.cr
+++ b/spec/compiler/codegen/class_var_spec.cr
@@ -551,4 +551,14 @@ describe "Codegen: class var" do
       a &+ b &+ c &+ d
       )).to_i.should eq(1 + 1 + 10 + 20)
   end
+
+  it "inline initialization of simple class var" do
+    mod = codegen(%(
+      class Foo
+        @@x = 1
+      end
+    ))
+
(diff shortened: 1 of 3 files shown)


asterite:
@waj Maybe we can add a couple of specs for this:
1. A spec that proves that inherited simple class variables are initialized
2. A spec that check that the "init" stuff doesn't appear for simple class variables. Something [like this](https://github.com/crystal-lang/crystal/blob/61bc4d79f9c2857c474c9ec3ce22e26813193ec6/spec/compiler/codegen/const_spec.cr#L384-L391) (but looking for the "init:..." thing we use in LLVM).

What do you think?

waj:
@asterite I just added a spec for 2. Isn't 1 already covered by this spec? https://github.com/crystal-lang/crystal/blob/4a0ca220e8ea60c7a085fbc98815220b65a05eb9/spec/compiler/codegen/class_var_spec.cr#L359-L379

asterite:
@waj You are right, just that spec is good, thanks!

PR #8068, opened 2019-08-09
Add ability to tag specs

Fryguy:
This PR adds support for tagging specs similar (but not identical) to RSpec's tagging.

Example:
```crystal
describe Foo do
  it "#my_fast_method", tags: "fast" do
  end

  it "#my_slow_method", tags: "slow" do
  end
end
```

To run it, you can do `crystal spec --tag fast` to run the specs tagged "fast" only, or `crystal spec --tag ~slow` to run the specs *not* tagged "slow"

Additionally,
- Tagging can be added at any level (describe/context/it).
- Multiple tags can be specified at any level.
- Tags can be specified as a single string or as anything that can become a `Set(String)` via `to_set`, typically an `Array(String)` or a `Tuple` of Strings.
  - `it "stuff", tags: "slow"`
  - `it "stuff", tags: ["slow", "cli"]`
  - `it "stuff", tags: {"slow", "cli"}`
- Multiple tags can be passed to `crystal spec` by specifying `--tag` multiple times.

#### Differences to RSpec tagging

- RSpec tagging is done via general metadata options on each describe/context/it.  This PR is instead a purpose driven `tags` keyword parameter.
- In RSpec, you tag as follows:
  - `it "stuff", :slow => true do` and run it with `rspec --tag slow`
  - `it "stuff", :custom_key => custom_value do` and run it with `rspec --tag custom_key:custom_value`

  Implementation wise (and performance-wise) I felt it much cleaner to just make them all strings.  So the "equivalent" in this PR is

  - `it "stuff", tags: "slow" do` and run it with `crystal spec --tag slow`
  - `it "stuff", tags: "custom_key:custom_value" do` and run it with `crystal spec --tag custom_key:custom_value`

  Note that in RSpec you *can* do `rspec --tag custom_key` to run anything tagged with that key regardless of value.  However, I felt the complexity to support Hash style args (and likely NamedTuple as well) wasn't worth it for this one feature, when the workaround is to just tag it explicitly with a second tag (e.g. `it "stuff", tags: ["custom_key", "custom_key:custom_value"] do`)

#### TODO

- ~~I still have to verify all of the various filtering combinations on the RSpec side (between tags, anti-tags, description, pattern, file, locations) and determine precendence and logic (i.e. AND/OR), then see if that's what we also want it to do here.~~
  See https://github.com/crystal-lang/crystal/pull/8068#issuecomment-526372069 and subsequent discussion in https://github.com/crystal-lang/crystal/pull/8125#issuecomment-526372807
- I'm not sure how to write specs for this, and there aren't specs for the existing filtering code (like locations).  I did a lot of manual testing with this, so it works.

### DIFF
--- a/src/spec/context.cr
+++ b/src/spec/context.cr
@@ -162,10 +162,10 @@ module Spec
       end
     end
 
-    def describe(description, file, line, end_line, focus, &block)
+    def describe(description, file, line, end_line, focus, tags, &block)
       Spec.focus = true if focus
 
-      context = Spec::ExampleGroup.new(@@current_context, description, file, line, end_line, focus)
+      context = Spec::ExampleGroup.new(@@current_context, description, file, line, end_line, focus, tags)
       @@current_context.children << context
 
       old_context = @@current_context
@@ -177,19 +177,19 @@ module Spec
       end
     end
 
-    def it(description, file, line, end_line, focus, &block)
-      add_example(description, file, line, end_line, focus, block)
+    def it(description, file, line, end_line, focus, tags, &block)
+      add_example(description, file, line, end_line, focus, tags, block)
     end
 
-    def pending(description, file, line, end_line, focus)
-      add_example(description, file, line, end_line, focus, nil)
+    def pending(description, file, line, end_line, focus, tags)
+      add_example(description, file, line, end_line, focus, tags, nil)
     end
 
-    private def add_example(description, file, line, end_line, focus, block)
+    private def add_example(description, file, line, end_line, focus, tags, block)
       check_nesting_spec(file, line) do
         Spec.focus = true if focus
         @@current_context.children <<
-          Example.new(@@current_context, description, file, line, end_line, focus, block)
+          Example.new(@@current_context, description, file, line, end_line, focus, tags, block)
       end
     end
 
@@ -277,7 +277,8 @@ module Spec
 
     def initialize(@parent : Context, @description : String,
                    @file : String, @line : Int32, @end_line : Int32,
-                   @focus : Bool)
+                   @focus : Bool, tags)
+      initialize_tags(tags)
     end
 
     # :nodoc:
--- a/src/spec/dsl.cr
+++ b/src/spec/dsl.cr
@@ -133,6 +133,15 @@ module Spec
     lines << line
   end
 
+  # :nodoc:
+  def self.add_tag(tag)
+    if anti_tag = tag.lchop?('~')
+      (@@anti_tags ||= Set(String).new) << anti_tag
+    else
+      (@@tags ||= Set(String).new) << tag
+    end
+  end
+
   record SplitFilter, remainder : Int32, quotient : Int32
 
   @@split_filter : SplitFilter? = nil
@@ -280,7 +289,7 @@ module Spec
 
   # :nodoc:
   def self.run_filters
-    root_context.run_filters(@@pattern, @@line, @@locations, @@split_filter, @@focus)
+    root_context.run_filters(@@pattern, @@line, @@locations, @@split_filter, @@focus, @@tags, @@anti_tags)
   end
 end
 
--- a/src/spec/methods.cr
+++ b/src/spec/methods.cr
@@ -14,8 +14,8 @@ module Spec::Methods
   # ```
   #
   # If `focus` is `true`, only this `describe`, and others marked with `focus: true`, will run.
-  def describe(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, &block)
-    Spec.root_context.describe(description.to_s, file, line, end_line, focus, &block)
+  def describe(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags : String | Enumerable(String) | Nil = nil, &block)
+    Spec.root_context.describe(description.to_s, file, line, end_line, focus, tags, &block)
   end
 
   # Defines an example group that establishes a specific context,
@@ -25,8 +25,8 @@ module Spec::Methods
   # It is functionally equivalent to `#describe`.
   #
   # If `focus` is `true`, only this `context`, and others marked with `focus: true`, will run.
-  def context(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, &block)
-    describe(description.to_s, file, line, end_line, focus, &block)
+  def context(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags : String | Enumerable(String) | Nil = nil, &block)
+    describe(description.to_s, file, line, end_line, focus, tags, &block)
   end
 
   # Defines a concrete test case.
@@ -41,8 +41,8 @@ module Spec::Methods
   # It is usually used inside a `#describe` or `#context` section.
   #
   # If `focus` is `true`, only this test, and others marked with `focus: true`, will run.
-  def it(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, &block)
-    Spec.root_context.it(description.to_s, file, line, end_line, focus, &block)
+  def it(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags : String | Enumerable(String) | Nil = nil, &block)
+    Spec.root_context.it(description.to_s, file, line, end_line, focus, tags, &block)
   end
 
   # Defines a pending test case.
@@ -58,15 +58,15 @@ module Spec::Methods
   # It is usually used inside a `#describe` or `#context` section.
   #
   # If `focus` is `true`, only this test, and others marked with `focus: true`, will run.
-  def pending(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, &block)
-    pending(description, file, line, end_line, focus)
+  def pending(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags : String | Enumerable(String) | Nil = nil, &block)
+    pending(description, file, line, end_line, focus, tags)
   end
 
   # Defines a yet-to-be-implemented pending test case
   #
   # If `focus` is `true`, only this test, and others marked with `focus: true`, will run.
-  def pending(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false)
-    Spec.root_context.pending(description.to_s, file, line, end_line, focus)
+  def pending(description = "assert", file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags : String | Enumerable(String) | Nil = nil)
+    Spec.root_context.pending(description.to_s, file, line, end_line, focus, tags)
   end
 
   # DEPRECATED: Use `#it`
--- a/spec/std/spec/context_spec.cr
+++ b/spec/std/spec/context_spec.cr
@@ -31,8 +31,8 @@ describe Spec::ExampleGroup do
   describe "#report" do
     it "should include parent's description" do
       root = FakeRootContext.new
-      child = Spec::ExampleGroup.new(root, "child", "f.cr", 1, 10, false)
-      grand_child = Spec::ExampleGroup.new(child, "grand_child", "f.cr", 2, 9, false)
+      child = Spec::ExampleGroup.new(root, "child", "f.cr", 1, 10, false, nil)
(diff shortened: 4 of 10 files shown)


Fryguy:
@bcardiff I thought you might be interested in this with respect to #8065.  Not sure if it helps you there or not.

Fryguy:
Gah...of course I broke anti-tags right before I pushed the commit 😄 .  Marking as WIP for now...will unWIP when ready.

Fryguy:
Ok, fixed anti-tags, but since I have those couple of TODO items at the top I'll leave this WIP.

As I said, some combinations aren't working as might be expected, but the code is still in a state for discussion.

src/spec/dsl.cr
```diff
@@ -130,8 +144,16 @@ module Spec
       return true if lines && lines.any? { |l| line == l || line <= l <= end_line }
     end
 
-    if spec_pattern || spec_line || locations
-      Spec::RootContext.matches?(description, spec_pattern, spec_line, locations)
+    if tags
+      tags = tags.to_set
+      return false if spec_anti_tags.try &.intersects?(tags)
+      return true if spec_tags.try &.intersects?(tags)
```
asterite: I think this is wrong. It seems this is matching if the tags match, but what if the line doesn't match? I know it's maybe uncommon to specify a line and a tag, but... why is this check otuside of `Spec::RootContext.matches?`?

src/spec/dsl.cr
```diff
@@ -110,10 +110,24 @@ module Spec
   end
 
   # :nodoc:
-  def self.matches?(description, file, line, end_line = line)
+  def self.add_tag(tag)
+    set =
+      if tag.starts_with?("~")
```
asterite: I think this can be replaced with:

```crystal
if antitag = tag.lchop?('~')
  tag = antitag
  @@anti_tags ||= Set(String).new
else
  @@tags ||= Set(String).new
end
```

src/spec/dsl.cr
Fryguy: Yeah, I agree...I have to do much more testing with respect to combinations of various filtering options

vlazar:
I love the idea for simplicity.

I feel the visual appearance is a bit noisy though, especially for applying a single tag:
```
describe Foo do
  it "#my_fast_method", tags: ["fast"] do
  end

  it "#my_slow_method", tags: ["slow"] do
  end
end
```

Now that's probably a bad idea, but visually I would love to see something like this instead:
```
describe Foo do
  it "#my_fast_method", tags(fast) do
  end

  it "#my_slow_method", tags(slow) do
  end
end
```

sam0x17:
Good place to use symbols imo

Fryguy:
I was considering allowing single item (i.e. `tags: "fast"`) as well as symbol (e.g. `tags: :fast`), but I was also trying to avoid having multiple ways to do something as per Crystal's mantra.  In the end I need to compare a Set of Strings to another Set of Strings, so symbols require conversion, and non-Arrays require wrapping, and thus I avoided them for now.  That being said it's not hard to add that in if it's really wanted.

Fryguy:
I've finally figured out the RSpec filtering rules, and they are kind of complicated (and unintuitive, IMO), but here's what I've gathered.  (Note that this is from manual testing various combinations, and not from the [source](https://github.com/rspec/rspec-core/blob/4ab682260afe252b90c4330bb7f91430f68f68c4/lib/rspec/core/filter_manager.rb))

- if example filter (i.e. description pattern) is supplied
  - ignore all tags and ignore all line numbers, and then filter within all specified files (including the file part of locations
    - Example: `rspec spec/a_spec.rb spec/b_spec.rb:5 --tag c -e foo` will ignore the `c` tag entirely and ignore the `:5` part, and then search for `/foo/` in the entirely of a_spec.rb and b_spec.rb
- else if locations are supplied
  - tags are ignored for the locations supplied
    - Example: `rspec spec/b_spec.rb:5 --tag c` will run all of b_spec:5 regardless of the presence of `c` tags in there.
  - However, if additional file/globs are supplied, then tags are honored within those.
    - Example: `rspec spec/a_spec.rb spec/b_spec.rb:5 --tag c` will run all of b_spec:5, and it will run only those examples in a_spec.rb with the `c` tag.
- else
  - tags are honored within file/glob if specified (otherwise the entire default pattern)
    - Example: `rspec spec/*_spec.rb --tag c` will run only those examples with the `c` tag

Fryguy:
@asterite Updated based on #8125.  You were right, this was *WAY* easier after the refactor.

Fryguy:
@asterite Opinion on opening up the syntax at the context/example level? I'm not sure if the Crystal "only have one way to do things" applied in this case or not.

This PR:

```crystal
it "foo", tags: ["slow"] do
it "foo", tags: ["slow", "cli"] do
```

Requested

- Allow single String for the common case of a single tag, in addition to multi-tag

  ```crystal
  it "foo", tags: "slow" do
  it "foo", tags: ["slow", "cli"] do
  ```

- Allow Symbol in addition to String

  ```crystal
  it "foo", tags: :slow do
  it "foo", tags: [:slow, :cli] do
  ```

straight-shoota:
I wouldn't accept symbols here. It doesn't make sense to use a symbol, this is clearly a string. Rubyists tend to be too much focused on symbols ;)

Also I think it would be fine to only accept `Array(String)`.
It could be suggested to write it as `tags: %w(foo)`, which is less verbose than using a regular array literal.

asterite:
Symbols definitely no. But accepting a single string might be fine.

Fryguy:
Updated with a commit to allow for single String (e.g. `tags: "slow"`)

straight-shoota:
What happens when `tags: "slow superslow"` specified? Should it be a single tag (with whitespace)  or should it be understood as two separate tags?

Sija:
@straight-shoota I'd assume it's a single tag.

Fryguy:
Yeah it's a single tag with a space in it...from the cli you could do `--tag "slow superslow"` to filter on it

Fryguy:
The `test_preview_mt` run is failing on something unrelated (but probably useful for @waj to see for the mt stuff).  I can't rerun the workflow as I don't have permissions anyway, so someone else will have to kick it.

```
Failures:

  1) closes the server

     Failure/Error: ch.receive.should eq(:end)

       Expected: :end
            got: :timeout

     # spec/std/http/server/server_spec.cr:73


Finished in 55.67 seconds
8552 examples, 1 failures, 0 errors, 6 pending

Failed examples:

crystal spec spec/std/http/server/server_spec.cr:58 # closes the server

Makefile:99: recipe for target 'std_spec' failed

make: *** [std_spec] Error 1

on_os linux docker run --rm -t -u 1001 -v /home/circleci/project:/mnt -v /etc/passwd:/etc/passwd -v /etc/group:/etc/group -w /mnt -e CRYSTAL_CACHE_DIR=/tmp/crystal crystallang/crystal:0.30.1-build linux64 /bin/sh -c 'CRYSTAL_WORKERS=4 make std_spec threads=1 verbose=1 FLAGS="-D preview_mt"' exited with 2
Exited with code 1
```

src/spec/dsl.cr
```diff
@@ -110,6 +110,18 @@ module Spec
...
+  # :nodoc:
+  def self.add_tag(tag)
+    set =
+      if anti_tag = tag.lchop? '~'
+        tag = anti_tag
+        @@anti_tags ||= Set(String).new
+      else
+        @@tags ||= Set(String).new
+      end
+    set << tag
```
RX14:
```cr
if anti_tag = tag.lchop('~')
  (@@anti_tags ||= Set(String).new) << anti_tag
else
  (@@tags ||= Set(String).new) << tag
end
```

Bonus points for just using `class_getter tags = Set(String).new`

RX14:
This is really good though.

src/spec/dsl.cr
Sija: @RX14 Shouldn't that be `class_getter tags : Set(String) { Set(String).new }` (the lazy variant)?

RX14: Why does it matter? Whats the benefit of shaving a new ns of the class initialization time here?

Fryguy: The code as is follows the pattern of the other filters with respect to how it sets an @@ var, so while it may make it cleaner to change it, I think that's best for a follow up PR.

I don't think we want the class_getter with a default because a nil `@@tags` (or `@@locations` or any other filter) actually means that it wasn't passed on the cli, and that knowledge is used later.  I could see changing that, and relying on, say, checking for an empty set, but as I said that's definitely for a follow up PR

Fryguy:
@RX14 Updated with your suggestion of cleanup within the add_tag method.

src/spec/methods.cr
```diff
@@ -14,8 +14,8 @@ module Spec::Methods
   # ```
   #
   # If `focus` is `true`, only this `describe`, and others marked with `focus: true`, will run.
-  def describe(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, &block)
-    Spec.root_context.describe(description.to_s, file, line, end_line, focus, &block)
+  def describe(description, file = __FILE__, line = __LINE__, end_line = __END_LINE__, focus : Bool = false, tags = nil, &block)
```
straight-shoota: Could you please add type restrictions for `tags` everywhere? It should be `String | Enumerable(String) | Nil`. This helps give better error messages if there is a mistake.

straight-shoota: Also I think all arguments after `description` should be named arguments. But that's a breaking change and should probably be a different PR.

Fryguy: I don't understand the named args request.  I thought all params were by default callable with names in Crystal.

asterite: You can force some to only be called by name. I think it's documented in the reference.

Fryguy: Oh right...had to look it up as I've never used it, but it's this syntax:

```crystal
def foo(a, *, c)
```

There, after the splat, c is forced to be a named arg.

Fryguy: Update with the type expansion.  I didn't do the named args though, as I think that should be a follow up PR since it should encompass more than just the tags parameter.

Fryguy:
> What happens when multiple --tag arguments are given?
> 
>     `--tag foo --tag bar` (all examples which have at least one of these tags?)

Yes, examples tagged with "foo" OR "bar"

>     `--tag ~foo --tag ~bar` (all examples which have none of these tags?)

Yes, examples that are not tagged OR tagged with neither with foo NOR bar

>     `--tag foo --tag ~bar` (all examples which have foo bur not bar tag?)

Yes, examples tagged with foo but NOT bar.

> And how does this work when tags are applied in a hierarchy?

Tagging a context effectively tags all examples within that context with the tag.  So if you do `--tag foo` and the context is tagged as foo you get everything in that context.  Then, if you could filter that further with `--tag ~bar` to eliminate anything within that list should you choose to.

> It would be really great to have a formal specification of expected behaviour for these combinations of tag definitions and filters.

I agree.  I actually have a PR to https://github.com/crystal-lang/crystal-book ready to go once this is merged.  One issue is that documentation wise, it's kind of complicated to present complicated combinations, particularly if it's mixed with things like focus, description, pattern, etc.  In general, though, this is just another level of filtering as was presented in #8125 .  Describing it as layers of filtering examples seems to be the simplest way to describe it rather than presenting combinations.

> Do you think you could add some specs for this?

I would love to but I don't know how, because it needs to test the spec cli itself.  There are presently no specs for `Spec`.  Even #8125 didn't have specs unfortunately.  If you have thoughts on how to add I can work on that though I think it would be better in a follow up PR.

Fryguy:
I think I found a nice way to test the filters.  I'll create a separate PR that can be merged before this one that tests all of the existing filters, then I'll update this PR to extend those tests with ones for tags.

Fryguy:
Added specs for all other filters in #8242.  Once that is merged, I'll rebase this one and add tag specs.

Fryguy:
@straight-shoota @RX14 I've rebased this PR and added tests, so this should be good to go.

src/spec/context.cr
```diff
@@ -277,7 +277,8 @@ module Spec
 
     def initialize(@parent : Context, @description : String,
                    @file : String, @line : Int32, @end_line : Int32,
-                   @focus : Bool)
+                   @focus : Bool, tags)
```
straight-shoota: Maybe this should provide a default value `tags = nil`? Ditto for `Example`.

Fryguy: I could do that.  I hadn't done it initially because `@focus` doesn't have a default, which is a similar type of parameter.

straight-shoota: Yeah, they could perhaps both get default values. It's not important though.

Fryguy: Let's hold off for a separate PR then.  I'm thinking there are a lot of quality of life changes around these parameters that you and I have have spoken of elsewhere, like enforcing named parameters, for example.

PR #8070, opened 2019-08-11
Extend Big* classes support for JSON serialization

### LINKED ISSUE #7856: Implement to_json(json : JSON::Builder) on BigDecimal
Implement method `to_json(json : JSON::Builder)` on BigDecimal class. BigInt and BigFloat are OK. `to_yaml(...)` is properly defined in the file "yaml/to_yaml.cr" on the Number superclass, `to_json` can be probably defined on the Number class too (in the file "json/to_json.cr").

Sija:
- Allows to serialize `BigDecimal` objects (to strings)
- Changes `Big*` classes to serialize as strings (in order to avoid overflows)
- Allows to deserialize `Big*` objects from strings
- Allows to deserialize `BigFloat` from ints
- Allows `BigDecimal` to be used as JSON keys

Closes #7856

### DIFF
--- a/spec/std/json/serialization_spec.cr
+++ b/spec/std/json/serialization_spec.cr
@@ -338,12 +338,12 @@ describe "JSON serialization" do
 
     it "does for BigInt" do
       big = BigInt.new("123456789123456789123456789123456789123456789")
-      big.to_json.should eq("123456789123456789123456789123456789123456789")
+      big.to_json.should eq(%("123456789123456789123456789123456789123456789"))
     end
 
     it "does for BigFloat" do
       big = BigFloat.new("1234.567891011121314")
-      big.to_json.should eq("1234.567891011121314")
+      big.to_json.should eq(%("1234.567891011121314"))
     end
 
     it "does for UUID" do
--- a/src/big/json.cr
+++ b/src/big/json.cr
@@ -1,44 +1,81 @@
 require "json"
 require "big"
 
-def BigInt.new(pull : JSON::PullParser)
-  pull.read_int
-  BigInt.new(pull.raw_value)
-end
-
-def BigInt.from_json_object_key?(key : String)
-  BigInt.new(key)
-rescue ArgumentError
-  nil
-end
-
-def BigFloat.new(pull : JSON::PullParser)
-  pull.read_float
-  BigFloat.new(pull.raw_value)
-end
-
-def BigFloat.from_json_object_key?(key : String)
-  BigFloat.new(key)
-rescue ArgumentError
-  nil
-end
-
-def BigDecimal.new(pull : JSON::PullParser)
-  case pull.kind
-  when .int?
-    pull.read_int
-    value = pull.raw_value
-  when .float?
-    pull.read_float
-    value = pull.raw_value
-  else
-    value = pull.read_string
+struct BigInt
+  def self.new(pull : JSON::PullParser)
+    case pull.kind
+    when .int?
+      pull.read_int
+      value = pull.raw_value
+    else
+      value = pull.read_string
+    end
+    new(value)
   end
-  BigDecimal.new(value)
+
+  def self.from_json_object_key?(key : String)
+    new(key)
+  rescue ArgumentError
+    nil
+  end
+
+  def to_json(json : JSON::Builder)
+    json.string(to_s)
+  end
 end
 
-def BigDecimal.from_json_object_key?(key : String)
-  BigDecimal.new(key)
-rescue InvalidBigDecimalException
-  nil
+struct BigFloat
+  def self.new(pull : JSON::PullParser)
+    case pull.kind
+    when .int?
+      pull.read_int
+      value = pull.raw_value
+    when .float?
+      pull.read_float
+      value = pull.raw_value
+    else
+      value = pull.read_string
+    end
+    new(value)
+  end
+
+  def self.from_json_object_key?(key : String)
+    new(key)
+  rescue ArgumentError
+    nil
+  end
+
+  def to_json(json : JSON::Builder)
+    json.string(to_s)
+  end
+end
+
+struct BigDecimal
+  def self.new(pull : JSON::PullParser)
+    case pull.kind
+    when .int?
+      pull.read_int
+      value = pull.raw_value
+    when .float?
+      pull.read_float
+      value = pull.raw_value
+    else
+      value = pull.read_string
+    end
+    new(value)
+  end
+
+  def self.from_json_object_key?(key : String)
+    new(key)
+  rescue InvalidBigDecimalException
+    nil
+  end
+
+  def to_json_object_key
+    to_s
+  end
+
+  def to_json(json : JSON::Builder)
+    json.string(to_s)
+  end
 end
--- a/src/big/yaml.cr
+++ b/src/big/yaml.cr
@@ -1,26 +1,29 @@
 require "yaml"
 require "big"
 
-def BigInt.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
-  unless node.is_a?(YAML::Nodes::Scalar)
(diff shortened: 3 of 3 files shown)


Sija:
@bcardiff @asterite Is there a chance for this PR to land in `v0.30.1`?
Right now one have to monkey-patch `BigDecimal` to fully support JSON (de)serialization.

asterite:
@Sija No, sorry, it would be too hasty

spec/std/json/serialization_spec.cr
```diff
@@ -338,12 +338,12 @@ describe "JSON serialization" do
 
     it "does for BigInt" do
       big = BigInt.new("123456789123456789123456789123456789123456789")
-      big.to_json.should eq("123456789123456789123456789123456789123456789")
+      big.to_json.should eq(%("123456789123456789123456789123456789123456789"))
```
asterite: Is this change needed?

spec/std/json/serialization_spec.cr
```diff
@@ -338,12 +338,12 @@ describe "JSON serialization" do
...
     end
 
     it "does for BigFloat" do
       big = BigFloat.new("1234.567891011121314")
-      big.to_json.should eq("1234.567891011121314")
+      big.to_json.should eq(%("1234.567891011121314"))
```
asterite: Is this change needed?

src/big/json.cr
```diff
@@ -1,44 +1,81 @@
...
-  BigInt.new(pull.raw_value)
-end
+struct BigInt
+  def self.new(pull : JSON::PullParser)
+    case pull.kind
+    when .int?
+      pull.read_int
+      value = pull.raw_value
+    else
+      value = pull.read_string
```
asterite: I don't think a string should be implicitly converted to a big int... we don't do that for regular ints, why should we do it for BigInt?

src/big/json.cr
```diff
@@ -1,44 +1,81 @@
...
+    new(key)
+  rescue ArgumentError
+    nil
+  end
 
-def BigFloat.new(pull : JSON::PullParser)
-  pull.read_float
-  BigFloat.new(pull.raw_value)
+  def to_json(json : JSON::Builder)
+    json.string(to_s)
```
asterite: BigInt should be serialized into an int, not a string...

src/big/json.cr
```diff
@@ -1,44 +1,81 @@
...
+  def self.new(pull : JSON::PullParser)
+    case pull.kind
+    when .int?
+      pull.read_int
+      value = pull.raw_value
+    when .float?
+      pull.read_float
+      value = pull.raw_value
+    else
+      value = pull.read_string
```
asterite: We shouldn't be able to deserialize a string into float just like that.

asterite:
Sorry, I think this PR is not good. Deserializing from string should cause an error. Serializing to string is not good either, int should serialize to int, float to float, etc.

Sija:
@asterite Serialization to `String` is a must-have for `Big*` classes since JSON doesn't natively supports any of them.

src/big/json.cr
Sija: Why not?

src/big/json.cr
Sija: Wrong, this will cause an overflow.

asterite:
> @asterite Since JSON doesn't natively supports any of them, isn't serialization to String a must-have for Big* classes?

JSON supports them. Where does it say they are not supported?

https://stackoverflow.com/a/13502497/4990850

straight-shoota:
Technically, the JSON number type has no limit on the length of a number. But practically, number literals bigger than Float64 or Int64 can introduce interoperability issues.
When JSON data is only handled by Crystal programs or other JSON libraries that support mapping JSON numbers to big numbers, everything's fine. But as soon as some other software is involved, it might fail reading the data, even if it doesn't even need to understand it. So there is a reasonable cause for serializing big numbers to and from JSON strings.

The question is: which format should we serialize to by default? And which should be supported for deserialzation by default: both or only a single one?

Sija:
@straight-shoota that was exactly my point, thank you :)

> The question is: which format should we serialize to by default? And which should be supported for deserialzation by default: both or only a single one?

I'd support both when deserializing, while serializing to `String` by default - or when value is bigger than `Int64/Float64` but that approach can lead to overflowing numbers, in for instance JavaScript, which handles numbers as big as `9007199254740991`, whereas `Int64::MAX` is `9223372036854775807`, so several magnitudes larger.

asterite:
@Sija I don't think it's out responsibility to cover how other libraries deserialize numbers. We can perfectly serialize big int from an integer JSON literal. Same goes for big float. That's because we serialize from the raw string that comes in the JSON without necessarily converting it to a crystal primitive integer (like Int64).

I can't see why we would ever need to parse a big integer from a string. Say you have a big int in some other language, maybe Ruby, maybe Java, maybe C#. Can they serialize it to a JSON integer? Yes! Because they are unlimited. So `123456789123456789` in any language can be represented as `123456789123456789` in JSON.

Can you provide the use case you want to cover? Is there any API out there that is returning big numbers as strings instead of JSON numbers?

Blacksmoke16:
A good compromise could be to deserialize from number by default, but provide a built in converter similar to https://crystal-lang.org/api/master/String/RawConverter.html for the String => Big* case.

straight-shoota:
@asterite You might not be in control of the source/target of JSON data and how they expect big numbers to be represented. Having them as a String type is perfectly reasonable compromise, even if number would technically do fine. So having a converter for this would be good. It's probably best to stick to number literals by default.

Sija:
@asterite I'll just chip in [one story](http://derpturkey.com/untitled/) how problematic is trying to parse bigint (or any bigger number) in js-land (regex!), for hopefully better understanding the problem at hand :)

asterite:
@Sija Thank you for the example! So it seems in javascript parsing big numbers by default will "fail", or, well, give incorrect results because of javascript's inability to represent these numbers. And so you have to use another library. 

The way I see it, Crystal has no problem parsing or emitting such numbers, and so we shouldn't change the way Crystal does it. In any case Javascript should improve that and support big numbers, somehow.

But if you really want to emit big integers as strings then you can use a converter. But I don't think emitting strings by default is not a good idea.

Parsing big integers or big floats as strings is probably not a big problem, but I'm hesitant. It would be nice to see how other language libraries behave: if you parse an int from a string, does it work or does it fail?

Sija:
> @Sija Thank you for the example! So it seems in javascript parsing big numbers by default will "fail", or, well, give incorrect results because of javascript's inability to represent these numbers. And so you have to use another library.

Exactly. Parsing strings on the other hand is very easy and in most dynamic languages you can easily discern between regular int values and number-like strings during normalization.

>But if you really want to emit big integers as strings then you can use a converter `[1]`. But I don't think emitting strings by default is not a good idea. `[2]`

`[1]`: That's some option (btw, do we have any _to_string_ converter in stdlib?), though IMHO a bit "bulky"
`[2]`: So, in other words: you think it is a good idea? :)

RX14:
In my opinion, `Big*` should serialize/deserialize to numbers and a converter should be provided in the standard library for serializing/deserializing to strings.

NoICE:
IMO: Crystal should not care about other implementations - it's their responsibility to implement the spec correctly. They might fix it some day and then Crystal would not be able to behave up to spec.

So Crystal should serialize to number literals, like spec says. If Crystal user wants to use String for serialization - guess what, he might do so! Just use String type instead of BigDecimal and do the conversion in the user code.

Anyway - this need to be implemented :-)

0.32.1: 
```crystal
Error: no overload matches 'BigDecimal#to_json' with type JSON::Builder`
```

caspiano:
Closed by https://github.com/crystal-lang/crystal/pull/9898

PR #8077, opened 2019-08-14
Regex: let a regex be multiline by default, `m` means DOTALL

### LINKED ISSUE #8062: Scan should be multiline by default?
I wanted to search all lines that started with something ("foo"), I wrote this in Ruby: 

```ruby
string = "foo bar\nfoo baz"
p string.scan(/^foo/) # => ["foo", "foo"]
```

But in Crystal the above result is:

```crystal
[Regex::MatchData("foo")]
```

Ignoring the fact that Crystal returns a different type (which makes sense for Crystal), should `scan` always do a multiline lookup? It works the same as in Ruby if I do:

```ruby
string = "foo bar\nfoo baz"
p string.scan(/^foo/m) # note the m modifier
```

asterite:
Fixes #8062

It's more intuitive this way, and it's also closer to how Ruby works.

I renamed `Options` members to separate values from PCRE from the ones we expose to the user because they are slightly different. In particular `MULTILINE` and `PCRE_MULTILINE` have different values.

### DIFF
--- a/spec/std/regex_spec.cr
+++ b/spec/std/regex_spec.cr
@@ -13,7 +13,9 @@ describe "Regex" do
 
   it "does inspect" do
     /foo/.inspect.should eq("/foo/")
-    /foo/.inspect.should eq("/foo/")
+    /foo/i.inspect.should eq("/foo/i")
+    /foo/m.inspect.should eq("/foo/m")
+    /foo/x.inspect.should eq("/foo/x")
     /foo/imx.inspect.should eq("/foo/imx")
   end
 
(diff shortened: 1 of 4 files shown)


oprypin:
Why make `m` mean DOTALL if Ruby is the only one that does that?

ysbaddaden:
How to disable multiline?

oprypin:
`/(?-m)pattern/` 😂

asterite:
There's no way to disable multiline. But I never needed that in Ruby (nor heard of anyone needing it).

> Why make m mean DOTALL if Ruby is the only one that does that?

Maybe everyone else is wrong? :-)

I don't know, there's no need to merge this. I just find Ruby intuitive in this regard and I think we should behave in the same way, but if it's fine to have to stick `/m` from time to time then it's okay.

ysbaddaden:
Then we must be careful, just like in Ruby, when validating user input, like an email for example, where we often want `\A` and `\Z` instead of `^` and `$` which is a not-known-enough caveat of activerecord validations (since it's matching multiline, you can have a botched input like "foo@example.com\ninjected: header\n" and it would still match, because the first line is fine).

asterite:
I guess there's no need to take this route.

PR #8082, opened 2019-08-14
Path: optimize `join` for the case of joining one single part

asterite:
Follow up to #8078

This optimizes `Path#join` (and `File.join`) for the most common case:

```crystal
File.join("prefix", "suffix")
```

When we just have two pieces to combine it becomes easier to compute the total bytesize needed for the string and we can use `String.new` instead of `String.build`, the latter involving an extra allocation for the `String::Builder` class (and there's also no need to write things though a pointer, check capacity, etc.)

This results in more code, but I think it's good if we optimize for this case because it's so common.

Benchmark code:

```crystal
require "benchmark"

base = "/Users/someone/foo/bar"
parts = ["baz.cr"]

Benchmark.ips do |x|
  x.report("join") do
    Path.new(base).join(parts)
  end
end
```

Results:

```
before: 10.90M ( 91.78ns) (± 2.80%)  144B/op
after:  19.14M ( 52.23ns) (±11.85%)  48.0B/op
```

That's twice as fast and with a third of the memory used.

Another benchmark (`Dir.glob` uses a lot of `File.join(x, y)`):

```crystal
require "benchmark"

Benchmark.ips do |x|
  x.report("Dir.glob") do
    Dir.glob("*/*.cr") do
    end
  end
end
```

Results:

```
before: 1.85k (541.76µs) (± 4.57%)  56.0kB/op
after:  1.89k (528.95µs) (±10.02%)  42.3kB/op
```

Not a lot of speed improvement because the `File.info?` call is the bottleneck (but that's solved by #8081), but look at the memory: 42.3kB/op vs. 56kB/op. Huge difference!

### DIFF
--- a/src/path.cr
+++ b/src/path.cr
@@ -621,6 +621,73 @@ struct Path
     expanded.normalize(remove_final_separator: false)
   end
 
+  # Appends the given *part* to this path and returns the joined path.
+  #
+  # ```
+  # Path["foo"].join("bar")     # => Path["foo/bar"]
+  # Path["foo/"].join("/bar")   # => Path["foo/bar"]
+  # Path["/foo/"].join("/bar/") # => Path["/foo/bar/"]
+  # ```
+  def join(part) : Path
+    # If we are joining a single part we can use `String.new` instead of
+    # `String.build` which avoids an extra allocation.
+    # Given that `File.join(arg1, arg2)` is the most common usage
+    # it's good if we can optimize this case.
+
+    if part.is_a?(Path) && posix? && part.windows?
+      part = part.to_posix.to_s
+    else
+      part = part.to_s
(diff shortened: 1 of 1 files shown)


src/path.cr
```diff
@@ -648,6 +648,9 @@ struct Path
   # ```
   def join(parts : Enumerable) : Path
     if parts.is_a?(Indexable)
+      # If it's just a single part we can avoid one allocation of String.build
+      return join_one(parts.first) if parts.size == 1
```
RX14: Would it be cheaper on the optimizer to add a `def join(part : String | Path)` and delegate immediately to `join_one`?

asterite:
@RX14 Thank you! 🙏

Issue #3983, opened 2017-02-02
OpenSSL server seems to leak memory


rdp:
Somehow, it seems that the OpenSSL sockets are never being GC'ed, and thus leaking memory.

How to repro:

```
$ cat use_it_ssl.cr
require "http/server"

context = OpenSSL::SSL::Context::Server.new
context.private_key = "key.pem"
context.certificate_chain = "cert.pem"

server = HTTP::Server.new("0.0.0.0", 8081) do |context|
  context.response.content_type = "text/plain"
  context.response.print "Hello world!"
end
server.tls = context
puts "Listening on https://:8081"
server.listen
```

and run this ruby script:

`100.times { Thread.new { loop { system("curl -k https://localhost:8081") } } }`

and watch memory go up slowly...

what is odd is also that the OpenSSL socket finalize method never seems to get called (even if I do a GC.start once/request).

This may also be related to some "accruing sockets" issue in https://github.com/crystal-lang/crystal/issues/3953

ysbaddaden:
Doesn't it hit a max memory usage, after a while? We do call `SSL_free` in the `finalize` method, so it should eventually collect memory.

ysbaddaden:
I just tried (master, ubuntu 14.04), and it can't reproduce a memory leak. As expected with a GC, memory climbs up continuously, but it quickly hits a maximum memory (both VIRT and RES) after which it stops climbing up.

The `finalize` method is called after the 5th or 6th request in my tests.

rdp:
I suppose it would after the RAM in your system is exhausted.  I put a puts
in the finalize method and I never saw it outputted, but didn't look
further than that.

rdp:
That is interesting...I tried it in a 16.10 64 bit and it seems stable, tried it in a 16.10 32 bit and it seemed stable.  However on my 15.04 32-bit it goes up to and uses like 300MB of VIRT.  And in OS X task manager says it grows up to like 1.5GB but it's not immediately clear if that means a leak or not...I'll post back here if I can figure out how to repro it.

kirbyfan64:
Did you mean to close?

RX14:
You don't want to be looking at virtual memory usage when comparing memory, you should be looking at resident memory. Each crystal fiber will use 8MiB of virtual memory regardless of how much is actually in use.

rdp:
[...]

PR #8088, opened 2019-08-16
avoid leaking from SSL_new if ssl_accept fails #3983

rdp:
Not the cleanest way but pattern seems OK, and it fixes up the leak.

### DIFF
--- a/src/openssl/ssl/socket.cr
+++ b/src/openssl/ssl/socket.cr
@@ -49,11 +49,15 @@ abstract class OpenSSL::SSL::Socket < IO
   class Server < Socket
     def initialize(io, context : Context::Server = Context::Server.new, sync_close : Bool = false)
       super(io, context, sync_close)
-
-      ret = LibSSL.ssl_accept(@ssl)
-      unless ret == 1
-        io.close if sync_close
-        raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
+      begin
+        ret = LibSSL.ssl_accept(@ssl)
+        unless ret == 1
+          io.close if sync_close
+          raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
+        end
+      rescue ex
+        finalize # otherwise GC never calls finalize, mem leak
+        raise ex
       end
     end
 


src/openssl/ssl/socket.cr
```diff
@@ -49,11 +49,15 @@ abstract class OpenSSL::SSL::Socket < IO
...
-      unless ret == 1
-        io.close if sync_close
-        raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
+      begin
+        ret = LibSSL.ssl_accept(@ssl)
+        unless ret == 1
+          io.close if sync_close
+          raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
+        end
+      rescue ex
```
bcardiff: The exception caught here is from `io.close` or the `raise OpenSSL::SSL::Error`?
If it's for the later then the `finalize` call could be put between creating the exception and the raise

```cr
ex = OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
finalize # or even LibSSL.ssl_free(@ssl) directly.
raise ex
```

rdp: Can io.close raise an exception?  I also forgot to add the client  equivalent but I have added it now. The client kind of feels like it needs an ensure clause.  Should I keep the same pattern for both do you think?  I just thought it'd be easier to maintain since it's more obvious that intermediate code goes "in the ensure/catch block" for the server side, it follows a pattern, etc.  Thanks.

straight-shoota: Ensure would run even if everything goes right. I'm pretty sure you don't want to free the socket then.

straight-shoota: `io.close`can theoretically raise. But I don't think that matters here because when `ret == 1` you need to free the SSL socket anyway.
So I think this change could just be a call to `LibSSL.ssl_free` right before the `io.close`.

straight-shoota:
I guess the culprit here is that `LibSSL.ssl_new` is initialized in `Socket#initialize` and needs to be freed if subsequent calls in the subclass initialized go wrong. It's hard to see that and probably the reason why this went unnoticed.
It's fine to fix this just in place, but maybe we could change the implementation to use a design that somehow avoid this by putting the `ssl_new` call and handling errors in subsequent calls in the same place, for example by using a block to do the custom initialization.
It doesn't matter much for this particular case but could be an example for similar situations.

src/openssl/ssl/socket.cr
rdp: OK how do I remove a commit and replace it in these github review PR's, do the fixes and create a new PR?  Add another commit?  Thanks.

bcardiff: Add another commit, we can squash on merge. No worries :-)

rdp: OK I tried putting `LibSSL.ssl_free` right before `io.close`.  It segfaults on the line after that `raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")` My hunch is because the ssl_free call right before that was actually successful, so it lost error context...hmm..so I guess we could reorder everything there or use the rescue block...

bcardiff: If you do

```cr
ex = OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
LibSSL.ssl_free(@ssl)
raise ex
```

It should let you create the exception, free and _then_ raise. I’m assuming `@ssl` is not used after the exception is initialized.

src/openssl/ssl/socket.cr
```diff
@@ -2,36 +2,40 @@ abstract class OpenSSL::SSL::Socket < IO
...
+        end
 
-      ret = LibSSL.ssl_connect(@ssl)
-      unless ret == 1
-        raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_connect")
+        ret = LibSSL.ssl_connect(@ssl)
+        unless ret == 1
+          raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_connect")
+        end
+      rescue ex
```
bcardiff: The `Server#initialize` looks good. But on `Client#initialize` still has the raise, rescue, raise.

rdp: Client#initialize has far more return points..do you want me to insert it carefully before all of them?  It seems to me like it would clutter things quite a bit...

rdp: It has come to my knowledge that even SSL_Accept can raise an Exception (ex "connection reset by peer") if the underlying read/writes within it fail.  Mind blown.  I propose we just go with begin rescue blocks?

straight-shoota: `ssl_accept` is a c function. How would it raise?

rdp: `ssl_accept` makes calls to "read" from its BIO which calls back into Crystal land.  Crystal then calls "socket read" which can raise Crystal exceptions.  Surprised me...

jhass:
@asterite unfortunately we cannot get rid of the rescue/reraise part due to https://github.com/crystal-lang/crystal/pull/8088#discussion_r316918433

jhass:
It's working <3

Screenshot 2019-08-28 at 16 59 21

(friend of mine's camo.cr deployment before and after this PR)

RX14:
@jhass wow, someone's using that? And it still compiles???

jhass:
It does and works really well, better (and way more efficient) than the original. After this leak was fixed anyway :D I use it two in a couple of places.

Issue #8085, opened 2019-08-15
`URI#query_params`


Blacksmoke16:
I propose to add a `#query_params` (or something along those lines) method to `Uri`.  This method would return a `HTTP::Params` object based on the query params defined on the `Uri` instance.

This would mainly just be an easier shortcut to:

```cr
require "http/params"
require "uri"

uri = URI.parse "example.com/error?e=invalidurl"

if p = uri.query
 HTTP::Params.parse(p) # => HTTP::Params(@raw_params={"e" => ["invalidurl"]})
end
```

With the new method
```cr
require "uri"

uri = URI.parse "example.com/error?e=invalidurl"
uri.query_params # => HTTP::Params(@raw_params={"e" => ["invalidurl"]})
```

The `HTTP::Params` instance would be empty if there were no query params defined on the `Uri`.

PR #8090, opened 2019-08-16
Adds URI#query_params method

rodrigopinto:
This PR is for adding a `URI#query_params` which will generate a `HTTP::Params` instance parsed from the `URI#query`.

Example:

```crystal
require "uri"

uri = URI.parse "http://foo.com?id=30&limit=5#time=1305298413"
uri.query_params # => HTTP::Params(@raw_params={"id" => ["30"], "limit" => ["5"]})
```
👉 it closes https://github.com/crystal-lang/crystal/issues/8085

### DIFF
--- a/spec/std/uri_spec.cr
+++ b/spec/std/uri_spec.cr
@@ -180,13 +180,13 @@ describe "URI" do
     end
   end
 
-  describe "hostname" do
+  describe "#hostname" do
     it { URI.new("http", "www.example.com", path: "/foo").hostname.should eq("www.example.com") }
     it { URI.new("http", "[::1]", path: "foo").hostname.should eq("::1") }
     it { URI.new(path: "/foo").hostname.should be_nil }
   end
 
-  describe "full_path" do
+  describe "#full_path" do
     it { URI.new(path: "/foo").full_path.should eq("/foo") }
     it { URI.new.full_path.should eq("/") }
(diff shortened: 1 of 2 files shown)


spec/std/uri_spec.cr
```diff
@@ -322,6 +322,21 @@ describe "URI" do
     end
   end
 
+  describe "#query_params" do
+    it do
+      uri = URI.parse("http://foo.com?id=30&limit=5#time=1305298413")
+      uri.query_params.should eq(HTTP::Params.parse("id=30&limit=5"))
```
straight-shoota: Technically, the spec would be more reliant if the `HTTP::Params` instance was created from a spec. We're not testing `Params.parse` here, but still...

rodrigopinto: What do you mean? Something like:

```crystal
expected_params = HTTP::Params.new({ "id" => ["30"], "limit" => ["5"]} )
uri.query_params.should eq(expected_params)
```

straight-shoota:
@rodrigopinto Could you please squash the commits except the first one? I'd like to merge this as two commits, a spec refactor and a feature addition.

rodrigopinto:
@straight-shoota it is done.

straight-shoota:
Thank you @rodrigopinto

PR #8097, opened 2019-08-18
Increase precision of `Process.times`

### LINKED ISSUE #7964: Port `Process.times` to RUsage?
While running `Benchmark.measure` on some code to see how much CPU time it consumes, I noticed it only reports with a granularity of 0.01 seconds. Because this code does I/O, this requires running it for a _long_ time to get results with enough significant digits in it to be able to understand how it scales.

Ruby's `Benchmark.measure` offers microsecond precision, so I looked into [Ruby's implementation of `Process.times`](https://github.com/ruby/ruby/blob/0c6c937904aafc1809386bd892a2d114d22d01fe/process.c#L7364-L7393) and I noticed they use LibC's `getrusage` API. I have an implementation of `Process.times` in Crystal I've been using locally to get those 4 orders of magnitude more precision, but I assume there was a reason Crystal doesn't use `getrusage`. I've found mentions of it as far back as Linux 2.6 (released 16 years ago) and it works fine on my Mac, so it probably has decent support.

I haven't posted a PR for it yet because I want to make sure there isn't a major blocker for it before I have to add `LibC` bindings into all of these:

```
aarch64-linux-gnu      i686-linux-gnu         x86_64-macosx-darwin
aarch64-linux-musl     i686-linux-musl        x86_64-openbsd
amd64-unknown-openbsd  x86_64-darwin          x86_64-portbld-freebsd
arm-linux-gnueabihf    x86_64-freebsd         x86_64-unknown-freebsd
i386-linux-gnu         x86_64-linux-gnu       x86_64-windows-msvc
i386-linux-musl        x86_64-linux-musl
```

jgaskins:
Moving to LibC's `getrusage` API gives us 4 orders of magnitude more precision.

Resolves #7964

### DIFF
--- a/src/lib_c/x86_64-darwin/c/sys/times.cr
+++ b/src/lib_c/x86_64-darwin/c/sys/times.cr
@@ -1,4 +1,7 @@
 require "./types"
+require "c/stdlib"
+require "c/sys/times"
+require "c/unistd"
 
 lib LibC
   struct Tms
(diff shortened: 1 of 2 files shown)


jgaskins:
Just realized I marked the PR as ready for review but I didn't update the `lib_c` files for architectures that aren't run on CI. If anyone has a good way of doing that before I get to it, feel free because I'm too tired right now. :-)

asterite:
Unless we get all C bindings for every OS we support there's no chance this PR will get merge. 

@jgaskins I recommend at least copying the existing C bindings you added to the other `sys/times.cr` files. If CI passes then we can merge this. If eventually the release build fails for some OS we can fix the bindings. But if they are missing we can't merge this.

asterite:
Alternatively, introduce a `HAS_RUSAGE = true` in existing C bindings and ``HAS_RUSAGE = false` in others and conditionally call `get_rusage` based on this (doing `{% if HAS_RUSAGE %}`). That way we can support rusage on most common systems and otherwise fall back to the slow alternative.

jgaskins:
omg I don't know how I completely missed where you mentioned `LibC::Long`.

PR #8098, opened 2019-08-18
Compiler: fix abstract def check regarding generic ancestor lookup

### LINKED ISSUE #8096: Abstract method implemented via extend self not considered to be implemented in some circumstances
This worked in 0.29.0:
```crystal
module ICallable(T)
  abstract def call(foo : Proc(T))
end

module Caller
  extend self
  extend ICallable(Nil)

  def call(foo : Proc(Nil))
  end
end
```
In 0.30.1 it says that the abstract method ```call``` must be implemented by ```Caller:Module```.
- Compiles by changing ```def call``` to ```def self.call```.
- Also compiles if generics are not involved.

asterite:
Fixes #8096

Thank you @Exilor ! Thanks to your bug report I was able to understand why other (rare) cases were failing too.

The code is this:

```crystal
module ICallable(T)
  abstract def call(foo : Proc(T))
end

module Caller
  extend self
  extend ICallable(Nil)

  def call(foo : Proc(Nil))
  end
end
```

And the compiler will check that `ICallable(T)#call` is implemented by all concrete subtype. One such subtype is `Caller:Module` (where class methods live). Does it provide a `call` method? No. Then the compiler goes and check whether any ancestor of `Caller:Module` provides it. Because it extends `self` it checks whether `Caller` (where instance methods live) provides it. And it does! But we must check whether that matches the method in the generic type. For that the compiler will try to find the generic instantiation of that module. That is, because we are checking a subtype of `ICallable(T)` then at some point there must have been some `include ICallable(X)` in the chain, to know what type `X` is. The problem was that the compiler would start the search from the current ancestor, in this case `Caller`, and not from the main type we were checking in the first place (`Caller:Module`).

It's a bit confusing. That's why I also added a similar spec which is a bit simpler:

```crystal
module ICallable(T)
  abstract def call(foo : T)
end

module Moo
  def call(foo : Int32)
  end
end

module Caller
  extend ICallable(Int32)
  extend Moo
end
```

Here the compiler again will check whether `Caller:Module` implements `call`. It will find that `Moo` has it and then it will try to find the `include/extend ICallable`, but it won't find starting from `Moo`, it will find it starting from `Caller:Module`. Well, at least in this case, in other cases it might happen that `Moo` is the one that is including the other module.

### DIFF
--- a/spec/compiler/semantic/abstract_def_spec.cr
+++ b/spec/compiler/semantic/abstract_def_spec.cr
@@ -602,4 +602,38 @@ describe "Semantic: abstract def" do
       end
     ))
   end
+
+  it "implements through extend (considers original type for generic lookup) (#8096)" do
+    semantic(%(
+      module ICallable(T)
+        abstract def call(foo : T)
+      end
+
+      module Moo
+        def call(foo : Int32)
+        end
+      end
+
+      module Caller
+        extend ICallable(Int32)
+        extend Moo
+      end
+    ))
+  end
+
+  it "implements through extend (considers original type for generic lookup) (2) (#8096)" do
+    semantic(%(
+      module ICallable(T)
+        abstract def call(foo : T)
+      end
+
+      module Caller
+        extend ICallable(Int32)
+        extend self
+
+        def call(foo : Int32)
+        end
+      end
+    ))
+  end
 end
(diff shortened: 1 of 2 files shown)


asterite:
Also undoes #8035 because this fix is the right solution to the problem, the other one was not.

asterite:
@RX14 Thanks! 🙏

PR #8104, opened 2019-08-20
Hash: use integer division instead of float division

asterite:
Fixes a couple of warnings in `src/hash.cr`:

```
$ bin/crystal spec/std/hash_spec.cr --warnings=all
In src/hash.cr:240:54

 240 | @entries = malloc_entries(initial_indices_size / 2)
                                                      ^
Warning: Deprecated Int32#/. `Int#/` will return a `Float` in 0.29.0. Use `Int#//` for integer division.

In src/hash.cr:887:18

 887 | indices_size / 2
                    ^
Warning: Deprecated Int32#/. `Int#/` will return a `Float` in 0.29.0. Use `Int#//` for integer division.

A total of 2 warnings were found.
```

Thanks @Blacksmoke16 for finding this (though next time please report it as an issue, I might forget what I read in gitter 😊)

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -237,7 +237,7 @@ class Hash(K, V)
         initial_indices_size = Math.pw2ceil(initial_capacity)
       end
 
-      @entries = malloc_entries(initial_indices_size / 2)
+      @entries = malloc_entries(initial_indices_size // 2)
 
       # Check if we can avoid allocating the `@indices` buffer for
       # small hashes.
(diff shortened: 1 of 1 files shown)


RX14:
Did we forget to make the `Int#/` change for `0.29.0` and `0.30.0`?

bcardiff:
No, but the llvm migration took more time than expected. So it didn’t fit in my plate at least.

straight-shoota:
IMO we shouldn't put out such message with a precise schedule like "will return a `Float` in 0.29.0". There can always be delays and cause that we can't fulfill the announced plan. For example "will return a `Float` after 0.28.0" is almost equally expressive but doesn't claim to happen in a specific release.

PR #8106, opened 2019-08-21
Do not build empty INI section

j8r:
Having an INI file with no section is valid. However, the current builder will add an empty section `[]`, which will result of a changed file.

### DIFF
--- a/src/ini.cr
+++ b/src/ini.cr
@@ -71,7 +71,12 @@ class INI
 
   # Appends INI data to the given IO.
   def self.build(io : IO, ini, space : Bool = false) : Nil
+    # An empty section has to be at first, to prevent being included in another one.
+    ini[""]?.try &.each do |key, value|
+      io << key << (space ? " = " : '=') << value << '\n'
+    end
     ini.each do |section, contents|
+      next if section.to_s.empty?
       io << '[' << section << "]\n"
       contents.each do |key, value|
         io << key << (space ? " = " : '=') << value << '\n'
--- a/spec/std/ini_spec.cr
+++ b/spec/std/ini_spec.cr
@@ -77,7 +77,7 @@ describe "INI" do
   end
 
   describe "build to an INI-formatted output" do
-    it "build from a Hash" do
+    it "builds from a Hash" do
       INI.build({
         "general" => {
           "log_level" => 'D',
@@ -91,7 +91,7 @@ describe "INI" do
         },
       }, true).should eq(File.read datapath("test_file.ini"))
     end
-    it "build from a NamedTuple" do
+    it "builds from a NamedTuple" do
       INI.build({
         "general": {
           "log_level": 'D',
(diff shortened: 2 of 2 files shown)


src/ini.cr
```diff
@@ -71,7 +71,12 @@ class INI
 
   # Appends INI data to the given IO.
   def self.build(io : IO, ini, space : Bool = false) : Nil
+    # An empty section have to be at first, to prevent being included in another one.
+    ini[""]?.try &.each do |key, value|
+      io << key << (space ? " = " : '=') << value << '\n'
+    end
     ini.each do |section, contents|
+      next if !section.is_a?(Symbol) && section.empty?
```
asterite: Better ask if it's a string. Or even better: if to_s is empty.

j8r: I thought of this option, but:
- `section` may be any object, because the `ini` argument isn't type restricted - `empty?` can be implemented by other objects than `String`
- using `to_s` for each `section` can impact performance if `section` isn't a `String`

asterite: Okay, then `next if section.is_a?(String) && section.empty?`.

But calling `to_s` here is probably good too because most of the time sections are going to be strings or symbols and there's no memory allocation in those cases.

straight-shoota: I think we can safely assume that section names are strings or can easily convert to string. Calling `#empty?` on an completely unknown type (the only thing we know is that it's not `Symbol`) might result in unexpected behaviour.

j8r: The method accept every objects implementing `#each(&block)`, that yields 2 values.
Calling `#to_s` may be dangerous too.

asterite: No, that's for `contents`. `section` should be a simple object.

src/ini.cr
```diff
@@ -71,7 +71,12 @@ class INI
 
   # Appends INI data to the given IO.
   def self.build(io : IO, ini, space : Bool = false) : Nil
+    # An empty section have to be at first, to prevent being included in another one.
+    ini[""]?.try &.each do |key, value|
+      io << key << (space ? " = " : '=') << value << '\n'
+    end
     ini.each do |section, contents|
+      next if !section.to_s.empty?
```
asterite:
```suggestion
      next if section.to_s.empty?
```

src/ini.cr
```diff
@@ -71,6 +71,12 @@ class INI
 
   # Appends INI data to the given IO.
   def self.build(io : IO, ini, space : Bool = false) : Nil
+    # An empty section have to be at first, to prevent being included in another one.
+    if ini.is_a? Hash
+      ini.delete("").try &.each do |key, value|
```
asterite: This is wrong! You are modifying data the user is passing.

What was wrong with the previous approach?

asterite: I think the safest way to do this is to iterate `ini` twice, once looking for an empty section, once looking for non-empty sections.

j8r: Right, yes

j8r:
Remains your approval @asterite

PR #8107, opened 2019-08-22
expose .clear method on HTTP::Cookies collections

sam0x17:
Originated from a discussion on the crystal gitter room. As it turns out the `clear` method on the underlying cookies collection in `HTTP::Cookies` is not exposed, so there is no direct way of clearing cookie collections without using some sort of iterator or overwriting the variable storing the collection (which usually isn't possible).

Have been using the following monkey-patch in projects for a while, with success:
```crystal
class HTTP::Cookies
  def clear
    @cookies.clear
  end
end
```

This PR is equivalent to the above.

Thanks to @watzon for encouraging me to submit a PR.

Right now there are not specs for similar methods in `HTTP::Cookies`, but I would be happy to add a spec if we think that is necessary.


sam0x17:
real world example using https://github.com/kemalcr/kemal:
```crystal
  get "/logout" do |env|
    env.response.cookies.clear
    "successfully signed out"
  end
```

luislavena:
Thanks for the PR @sam0x17, will be possible to add some specs to complete the behavior and make sure future updates do not break it?

Thank you! ❤️ ❤️ ❤️

sam0x17:
@luislavena sure I'll do that right now

spec/std/http/cookie_spec.cr
```diff
@@ -252,6 +252,36 @@ module HTTP
...
+    end
+
+    it "allows clearing the cookies collection" do
+      cookies = Cookies.new
+      cookies << Cookie.new("test_key", "test_value")
+      cookies << Cookie.new("a", "b")
+      cookies << Cookie.new("c", "d")
+      cookies.clear
+      cookies.size.should eq 0
+      cookies.empty?.should be_true
```
Blacksmoke16: size/empty expectations are a bit redundant.

You could prob just do `cookies.should be_empty`

sam0x17: ah, didn't know about that helper

sam0x17: Noted

Yeah my thought was this will fail if either `size` or `empty?` breaks but it's probably totally unnecessary.

sam0x17:
should be good now @Blacksmoke16  @luislavena

PR #8109, opened 2019-08-22
Fix JSON and YAML parsing of NamedTuple with nilable fields

### LINKED ISSUE #8089: Support JSON/YAML parsing of NamedTuple with nilable field
Actually class or struct with JSON.mapping could have nilable field.
It's really needed with NoSQL databases driver who need crystal types for adding behavior.

This feature could be used with question mark (_?_) like this : 

```crystal
alias Car = NamedTuple (
  name: String,
  type: String,
  clients: Array(String)?
)

```

asterite:
Fixes #8089


asterite:
@jhass Thank you for all the reviews! ❤️

PR #8112, opened 2019-08-23
Multithreading

waj:
This is a big PR that adds initial and experimental support for multithreading. It works side by side with the current single thread implementation and must be enabled with the `preview_mt` flag.

There is an upcoming blog post explaining details about the implementation and decisions made on this work.

Things covered by this PR:

- One scheduler per thread (workers): each thread has a queue of runnable fibers. Each fiber is assigned to a worker when it starts and currently it lives in the same worker until it finishes (no job stealing)
- One event loop per thread: whenever a thread becomes idle (no runnable fibers left) it waits on the event loop for events (IO, timers). The same event loop also receives notifications about new fibers created or awaken from other threads.
- New implementation of a thread safe `Channel`: re-created from scratch because with the previous implementation it was harder to make a multi thread `select` operation.
- Atomic initialization of constants and class variables: to avoid race conditions
- Thread safe `Mutex`

Things not covered by this PR:

- Job stealing: naive approaches to implement this are detrimental to performance
- Soundness of types: union types might become unsound in MT mode
- Thread safety of std classes: for example, `Array` is not thread safe and must be protected by a mutex in case it's shared between fibers


src/crystal/scheduler.cr
```diff
@@ -123,4 +138,84 @@ class Crystal::Scheduler
...
+
+    protected def find_target_thread
+      if @@workers.empty?
+        Thread.current
+      else
+        @rr_target += 1
+        @@workers[@rr_target % @@workers.size]
+      end
+
+      # target = Thread.workers[@rr_target]
```
wontruefree: Will this be removed before merge?

src/crystal/scheduler.cr
```diff
@@ -123,4 +138,84 @@ class Crystal::Scheduler
...
+    end
+
+    @@workers = [] of Thread
+
+    def self.workers
+      @@workers
+    end
+
+    def self.init_workers
+      count = ENV["CRYSTAL_WORKERS"]?.try &.to_i || 4
```
wontruefree: I dont see this documented anywhere.  should this be added to the docs?
Also should this be an environment variable?  I would expect a command line flag.
What do other languages do to set the number of threads in a program.

wontruefree:
Thank you so much for all this work!!!! I am super excited to build out something that uses this.

I am curious but do you have any benchmarks with this?

src/crystal/scheduler.cr
j8r: And why `4`?

j8r: Perhaps it's the Crystal's equivalent of `GOMAXPROCS`, then it would be better to defaulting it to the number of CPU  cores.

j8r: Perhaps it's the Crystal equivalent of GOMAXPROCS, then it would be better to defaulting it to the number of CPU cores.

src/crystal/scheduler.cr
waj: Actually that's a leftover from an attempt to select the least busy thread for each new fiber. But that was buggy and unreliable. I'll remove all that and revisit in future.

src/crystal/scheduler.cr
waj: For example, you can use GOMAXPROCS in Go to set a similar configuration. Command line flags would require every Crystal app to parse the arguments. Remember this is a dynamic setting: once the program is compiled with MT support (`-D preview_mt`) the number of threads can be changed using that environment variable.

waj:
@wontruefree thank you!! we're also really excited to release this work. We have some benchmarks coming soon in a blog post.

src/crystal/scheduler.cr
waj: Ideally we could use `System.cpu_count` as the default value but we still found high contention that needs to be addressed

src/crystal/scheduler.cr
```diff
@@ -123,4 +138,68 @@ class Crystal::Scheduler
...
```
asterite:
```suggestion
      count = ENV["CRYSTAL_WORKERS"]?.try &.to_i? || 4
```

If a non-integer like "foo" is passed to `CRYSTAL_WORKERS` this will crash the program. Maybe it's fine.

How do we want to handle it? I just checked in Go, they seem to completely ignore the value if it's not an integer. In our case it might be good to raise or abort the program, it might be a programmer mistake and we should detect that.

spec/std/concurrent/select_spec.cr
```diff
@@ -100,4 +110,87 @@ describe "select" do
...
+    x.should eq 1
+
+    select
+    when x = ch2.receive
+    when x = ch1.receive
+    end
+    x.should eq 2
+  end
+
+  it "stress select with send/receive in multiple fibers" do
```
asterite: Does this spec belong here? I can't see `select` used, but maybe the methods called here are an implementation detail for that?

src/crystal/scheduler.cr
```diff
@@ -42,24 +58,30 @@ class Crystal::Scheduler
     Thread.current.scheduler.yield(fiber)
   end
 
+  @worker_in : IO
+  @worker_out : IO
```
asterite: I saw a commit where these were extracted to a `class` and then that was reverted? Why? Was it because of the overhead of using a `class`? Maybe using `struct` would have worked equally well without the memory overhead?

src/channel.cr
```diff
@@ -15,12 +16,55 @@ require "fiber"
...
-  module SelectAction
-    abstract def ready?
-    abstract def execute
-    abstract def wait
+class Channel(T)
+  @lock = Crystal::SpinLock.new
+  @queue : Deque(T)?
+
+  module NotReady
+    extend self
```
asterite: Is this `extend self` needed here?

bcardiff: It's a way to create unit types. Same as empty structs, but you don't need to put a `T.new` to refer the unit value. https://twitter.com/bcardiff/status/1157007260985495552

src/crystal/scheduler.cr
wontruefree: I imagine there are a few cases you could cause this to fail.  I would be curious what negative numbers would do.  What about white listing the numbers?

```suggestion
      count = ("1".."100").include?(ENV["CRYSTAL_WORKERS"]?.try &.to_i) || 4
```

I would rather a raise an error for bad input rather to a logic error being risen from the results due to bad inputs.

src/thread/linked_list.cr
```diff
@@ -6,7 +6,7 @@ class Thread
   #
   # Thread-safe doubly linked list of `T` objects that must implement
   # `#previous : T?` and `#next : T?` methods.
-  struct LinkedList(T)
+  class LinkedList(T)
```
wontruefree: What is the cause for this change?

src/channel.cr
asterite: Oooooh... I see. But you can also use `is_a?(NotReady.class)`, but I guess it's easier to use the `extend self` trick. Nice!

src/thread/linked_list.cr
asterite: LinkedList has a `Mutex` which was previously immutable but now it's mutable (has `@lock_count`) so using a struct means that locking effectively doesn't work. This has to be a class. But @waj or @bcardiff can confirm this.

bcardiff: The lazy initialization `Fiber.@@fibers` plus the `Fiber.fiber` getter didn't play nice if the linked list was a value type. `Fiber.fiber.push` would've been modifying copy instead of the expected list.

spec/compiler/codegen/const_spec.cr
```diff
@@ -446,4 +468,42 @@ describe "Codegen: const" do
 
     mod.to_s.should_not contain("CONST")
   end
+
+  it "synchronize initialization of contants" do
```
asterite:
```suggestion
  it "synchronize initialization of constants" do
```

src/compiler/crystal/codegen/once.cr
```diff
@@ -0,0 +1,38 @@
+require "./codegen"
+
+class Crystal::CodeGenVisitor
+  ONCE_STATE = "~ONCE_STATE"
+  @have_once_state = false
```
asterite: It seems this is never read, only assigned to.

src/crystal/event_loop.cr
```diff
@@ -1,28 +1,56 @@
 require "./event"
 
-module Crystal::EventLoop
-  @@eb = Crystal::Event::Base.new
-  @@dns_base : Crystal::Event::DnsBase?
+class Thread
```
asterite: Could we make this `Crystal::Thread`? Just to hide it more from the public API and also it will play nicely with #7995 which does that too.

src/thread/linked_list.cr
j8r: This doen't need to, `Mutex` is a class, thus passed by reference, even if inside a struc.
```cr
class Thread::Mutex
  getter lock_count : Int32 = 0
  def inc_lock
    @lock_count += 1
  end
end

def some_method(obj)
  obj.@mutex.inc_lock
end

ll = Thread::LinkedList(Nil).new
puts "before call: #{ll.@mutex.lock_count}"  #=> 0
some_method ll
puts "after call: #{ll.@mutex.lock_count}"   #=> 1
```

j8r: This isn't needed, `Mutex` is a class, thus passed by reference, even if inside a struct.
```cr
class Thread::Mutex
  getter lock_count : Int32 = 0
  def inc_lock
    @lock_count += 1
  end
end

def some_method(obj)
  obj.@mutex.inc_lock
end

ll = Thread::LinkedList(Nil).new
puts "before call: #{ll.@mutex.lock_count}"  #=> 0
some_method ll
puts "after call: #{ll.@mutex.lock_count}"   #=> 1
```

bcardiff: What I said before applies to the whole list. It has nothing to do with the mutex IIRC

j8r: @bcardiff sorry I was responding to @asterite , I didn't see your message.

asterite: Yeah, I don't know what I was thinking when I replied that.

asterite:
Should we add an entry in CI to also run the std specs with preview_mt?

bcardiff:
We could, or at least for branches named `mt/`. It will definitely consume more time. 🔥🔥🔥

src/channel.cr
```diff
@@ -127,23 +275,50 @@ abstract class Channel(T)
...
-          return index, result
-        end
-      end
+  def self.select(ops : Indexable(SelectAction), has_else = false)
+    # Sort the operations by the channel they contain
+    # This is to avoid deadlocks between concurrent `select` calls
+    ops_locks = ops
+      .to_a
+      .uniq(&.lock_object_id)
+      .sort_by(&.lock_object_id)
```
yxhuvud: Considering that `Channel.select` will often be called in a loop, allocations inside might end up costly if the actual work to handle each op is cheap. This might perhaps be microoptimized for `ops` with few entries along the same lines asterite optimized small Hashes recently?

(To hyperoptimize it, it might even make sense to create some sort of presorted registry that keep them in order that allow registrying or de-registrying selectables. But that might be overkill outside  #6468 :) ).

src/crystal/scheduler.cr
ysbaddaden: I actually believe it should be 1 by default and tweakable at compile time (maybe even at runtime).

Defaulting to all available cores only makes sense in some scenarios, for example long lived programs (which can easily define an ENV variable to enable more cores) but may not make sense in short lived programs, where initializing a dozen threads/schedulers will be more costly than just running

src/crystal/scheduler.cr
ysbaddaden: Why limit the number of cores? Cavium ThunderX (armv8, fully supported by Crystal) can have upto 128 cores for example. We don't know how many cores may be available tomorrow (or even today).

Thought coordinating that many cores without too much  will be huge challenge.

src/crystal/spin_lock.cr
```diff
@@ -0,0 +1,39 @@
+# :nodoc:
+class Crystal::SpinLock
+  {% if flag?(:preview_mt) %}
+    @m = Atomic(Int32).new(0)
```
j8r: What is `m`?

src/fiber/stack_pool.cr
```diff
@@ -5,13 +5,14 @@ class Fiber
 
     def initialize
       @deque = Deque(Void*).new
+      @m = Thread::Mutex.new
```
j8r: Rename `@m` to `@mutex`

src/crystal/event_loop.cr
```diff
@@ -1,28 +1,56 @@
...
+  @eb : Crystal::Event::Base?
+  @dns_base : Crystal::Event::DnsBase?
+  @loop_fiber : Fiber?
 
-  def self.after_fork
-    @@eb.reinit
+  # :nodoc:
+  def eb : Crystal::Event::Base
+    @eb ||= Crystal::Event::Base.new
```
j8r: Rename `@eb` to `@event_base`.
A getter can be used (and for `dns_base`)
`getter event_base : Crystal::Event::Base { Crystal::Event::Base.new }`

spec/compiler/codegen/const_spec.cr
yxhuvud: BTW, I think `synchronize` should have an `s` too.

src/crystal/spin_lock.cr
vlazar: I might be mistaken, but the question probably was more about giving a meaningful name to `@m`. Same as something like `@round_robin_target` instead of `@rr_target`  above.

src/crystal/scheduler.cr
```diff
@@ -123,4 +138,68 @@ class Crystal::Scheduler
     @current.resume_event.add(0.seconds)
     resume(fiber)
   end
+
+  {% if flag?(:preview_mt) %}
+    @rr_target = 0
```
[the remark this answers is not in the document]
j8r: Thanks for the explanation, it can be `@round_robin_target` then (I am not intimate with this algorithm vocabulary :sweat_smile: )

src/crystal/scheduler.cr
```diff
@@ -123,4 +138,86 @@ class Crystal::Scheduler
...
+        @runnables << fiber
+      end
+      @lock.unlock
+    end
+
+    @@workers = [] of Thread
+
+    def self.workers
+      @@workers
+    end
```
j8r: If you like, a `class_getter` can be used

src/crystal/scheduler.cr
```diff
@@ -123,4 +138,86 @@ class Crystal::Scheduler
...
+      end
+    end
+
+    private def self.worker_count
+      env_workers = ENV["CRYSTAL_WORKERS"]?
+
+      if env_workers && !env_workers.empty?
+        workers = env_workers.to_i?
+        if !workers || workers < 1
+          LibC.dprintf 2, "FATAL: Invalid value for CRYSTAL_WORKERS: #{env_workers}\n"
```
j8r: Why using `dprintf` instead of `STDERR.puts`, to not depend on `IO`?

src/thread.cr
```diff
@@ -8,7 +8,7 @@ class Thread
   # Use spawn and channels instead.
 
   # all thread objects, so the GC can see them (it doesn't scan thread locals)
-  @@threads = Thread::LinkedList(Thread).new
+  @@threads : Thread::LinkedList(Thread)?
```
j8r:
```suggestion
  protected class_getter threads : Thread::LinkedList(Thread) { Thread::LinkedList(Thread).new }
```
then remove the line 24 below (`def self.threads`)

src/fiber.cr
```diff
@@ -11,7 +11,7 @@ fun _fiber_get_stack_top : Void*
 end
 
 class Fiber
-  @@fibers = Thread::LinkedList(Fiber).new
+  @@fibers : Thread::LinkedList(Fiber)?
```
j8r:
```suggestion
  protected class_getter fibers : Thread::LinkedList(Fiber) { Thread::LinkedList(Fiber).new }
```
And remove line 33 below

src/crystal/thread_local_value.cr
```diff
@@ -0,0 +1,27 @@
+# :nodoc:
+class Crystal::ThreadLocalValue(T)
```
j8r: Not sure why a `class` would be necessary, since the underlying store is a `Hash`, thus passed always passed by reference.

src/compiler/crystal/codegen/once.cr
```diff
@@ -0,0 +1,36 @@
+require "./codegen"
+
+class Crystal::CodeGenVisitor
+  ONCE_STATE = "~ONCE_STATE"
+
+  def once_init
+    once_init_fun = @main_mod.functions[ONCE_INIT]?
+    if once_init_fun
```
j8r: The two above lines can be merged into a single one.

src/crystal/spin_lock.cr
```diff
@@ -0,0 +1,39 @@
...
+    {% end %}
+  end
+
+  def sync
+    begin
+      lock
+      yield
+    ensure
+      unlock
+    end
```
j8r: `begin` and `end` can be removed

src/crystal/spin_lock.cr
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def unsync
+    begin
+      unlock
+      yield
+    ensure
+      lock
+    end
```
j8r: Here too, `begin` and `end` can be removed

src/process.cr
```diff
@@ -399,6 +407,14 @@ class Process
   end
 
   private def channel
+    if channel = @channel
+      channel
+    else
+      raise "BUG: Notification channel was not initialized for this process"
+    end
+  end
```
j8r: This method is only used for `Process#wait`.
And even in its method, it could be `ex = channel.try &.receive`.
I propose to remove this method or move the logic to `Process#wait`.

src/crystal/spin_lock.cr
j8r: makes sense

j8r: Don't now. Can `lock` raise and a lock be partially set? Can `unlock` raise?

src/crystal/thread_local_value.cr
```diff
@@ -0,0 +1,27 @@
...
+  @values = Hash(Thread, T).new
+
+  def get(&block : -> T)
+    th = Thread.current
+    @values.fetch(th) do
+      @values[th] = yield
```
j8r:
```suggestion
      yield
```

src/crystal/scheduler.cr
ysbaddaden: Because IO depends on Fiber, EventLoop and Scheduler, which we aren't defined. dprintf will always work, using IO will likely fail to print anything.

src/crystal/event_loop.cr
```diff
@@ -1,28 +1,41 @@
...
 
   def self.resume
     loop_fiber.resume
   end
 
+  private def self.event_base
+    Thread.current.event_base
+  end
+
   private def self.loop_fiber
```
j8r: Not used anywhere.

j8r: I didn't found it in this PR nor with `grep -rn` in master.

j8r: I'm talking about this very private class method, not the instance method and variable.

j8r: Sure, but sorry I didn't undestand the code URM you mentioned, where `EventLoop.loop_fiber` is used then?

j8r: Sure, but sorry I didn't undestand the code URL you mentioned, where `EventLoop.loop_fiber` is used then?

j8r: Oh ok sorry, thanks. I didn't know this language feature, to call a static method like that.

src/crystal/event_loop.cr
```diff
@@ -1,28 +1,41 @@
...
+        Crystal::Scheduler.reschedule
+      end
+    end
   end
+end
+
+module Crystal::EventLoop
+  {% unless flag?(:preview_mt) %}
+    def self.after_fork
+      Thread.current.event_base.reinit
```
j8r:
```suggestion
      event_base.reinit
```

src/concurrent.cr
```diff
@@ -58,8 +58,11 @@ end
 #
 # 2.times { ch.receive }
 # ```
-def spawn(*, name : String? = nil, &block)
+def spawn(*, name : String? = nil, same_thread = false, &block)
```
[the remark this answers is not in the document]
jhass: Just make sure nobody changes the contract from "stays on the same thread" to "is initially queued for the same thread" in case job stealing is implemented :D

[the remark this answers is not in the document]
j8r: I just notice that `same_thread` should be typed as `Bool` here

Issue #8113, opened 2019-08-23
Tiny bugs in the Markdown library


Hentioe:
# Submitting questions

When I used the Markdown library to convert Markdown text to HTML, I found a small problem.
I tried to solve this problem, but I didn't find the reason in a short time.

# Submitting bugs

Code:

``` crystal
require "markdown"

mdtext = "[[google]](https://www.google.com)"
html = Markdown.to_html mdtext

puts html # => <p><a href="https://www.google.com">[google</a></p>
```

You can see that it has one `]` missing.

Play address: https://play.crystal-lang.org/#/r/7g29

Version:
```
~ crystal version
Crystal 0.30.1 [5e6a1b672] (2019-08-12)

LLVM: 4.0.0
Default target: x86_64-unknown-linux-gnu
```

straight-shoota:
The stdlib Markdown implementation has dozens of bugs and missing features. It should be considered broken, and I don't recommend using it.
I'd suggest to try https://github.com/icyleaf/markd instead, which is a solid and feature-rich Markdown implementation.

Hentioe:
Ok, thank you for your suggestion. It seems that this standard library is still immature and the problem should be a lot. I closed this issue.

Issue #7463, opened 2019-02-20
Markdown escapes HTML


Sija:
Which is obviously incorrect, since Markdown is [allowed](https://daringfireball.net/projects/markdown/syntax#html) to contain HTML tags.

For an example of such erroneous behavior see
https://hanneskaeufler.github.io/crytic/api/ generated by `crystal doc` out of https://raw.githubusercontent.com/hanneskaeufler/crytic/eb13b67362fc6646f9278120ebe5ba0c9526e772/README.md

Refs #4613

straight-shoota:
For this specific case, a simple workaround would be to avoid using HTML, it's not really necessary anyway.

But in general, I think we should find a way to use a proper Markdown library (like Makrd) in the docs generator.

Sija:
@straight-shoota Can't do, some things can't be done with Markdown alone.

asterite:
I still think we should split everything (std, compiler tools, etc.). Things would be so much simpler and will advance much smoother and in a more decentralized way.

asterite:
And we should probably close this as a duplicate.

Sija:
@asterite Duplicate of what? I haven't seen any issue regarding this explicit problem.

asterite:
Of #4613. Std Markdown is not feature complete. Not supporting HTML is just one more missing feature.

Sija:
IMO it's not a duplicate, rather a related issue.
For the time being there's still no decision anyway, after over a year from the last comment... :/

straight-shoota:
The main issue is that our Markdown implementation is very limited. But everyone should expect it to support Markdown reasonably well. I don't think this can be achieved by improving the current implementation. It needs to be replaced either by a complete rewrite or an existing library.

I don't see us fixing this issue without #4613. So this should be a duplicate.

Issue #7009, opened 2018-10-30
Numbered list not rendered correctly in docs when using code block


r00ster91:
This is how the current **Installation** section looks in the generated docs:
![grafik](https://user-images.githubusercontent.com/35064754/47743238-e291dc00-dc7e-11e8-91d8-aaa09c063e1a.png)
**1** and **1** instead of **1** and **2**.
Even if this is actually in the README:

`````md
## Installation

1. Add the dependency to your `shard.yml`:
```yaml
dependencies:
  library:
    github: user/library
```
2. Run `shards install`
``````

asterite:
That's probably a bug with the markdown renderer

r00ster91:
The bug is very likely here: [`parser.cr:253`](https://github.com/crystal-lang/crystal/blob/master/src/markdown/parser.cr#L253).

j8r:
Another solution is to [stop investing time for this in the stdlib](https://github.com/crystal-lang/crystal/issues/5215) and use e.g. [markd](https://github.com/icyleaf/markd)

straight-shoota:
Let's just ignore the error for now and change the list to be unordered. A fix can be delayed until we have some final decision on how to provide proper markdown support to the compiler. Or maybe we don't even need that if the compiler just outputs plain markdown and an external doc generator renderes it to HTML or whatever format.

Sija:
There are battle tested markdown libraries, that's not a problem if we'd like to go in that direction.

RX14:
We've been through the logistics of the compiler depending on external shards before and it's not going to happen any time soon.

r00ster91:
@bcardiff shouldn't this be rather fixed before releasing 0.27.0?

asterite:
I'd say not necessarily, the readme is usually read from GitHub, etc. The docs are used for the API (yes, the readme appears there too but someone will have read GitHub before, or having 1., 1. isn't a big deal).

r00ster91:
Unfortunately now after #7118, the **Installation** section is rendered even buggier:

![Installation](https://user-images.githubusercontent.com/35064754/49340042-a44f5a00-f63a-11e8-8166-b607eb730fa8.png)

``````md
## Installation

1. Add the dependency to your `shard.yml`:

   ```yaml
   dependencies:
     library:
       github: user/library
   ```

2. Run `shards install`
``````

straight-shoota:
[...]

PR #8115, opened 2019-08-24
Move Markdown to inside the compiler.

asterite:
The Markdown module is buggy and not feature complete. We shouldn't expose it in the standard library. For the compiler's purposes (`crystal doc`) it might be good but users should be encouraged to use a better library. Eventually the `crystal doc` tool should somehow use a different, better markdown library.

If with time we decide to improve it and make it feature complete and without bugs we could consider re-exposing it again. But right now this can't go like this in 1.0.

Related to #7463, #7009, #4613, #8113


drujensen:
This light markdown is very useful for simple cases i.e. doc generation.  It's being used by Amber's html helpers.  The reason given for hiding this was to encourage use of `better` libraries but the only one I see available is https://github.com/icyleaf/markd.  It is a full featured common markdown but it is slow.  What is the recommended replacement?

asterite:
I would recommend trying to optimize markd. You can also use the internal. Markdown type, it will still work but it's not recommended.

asterite:
Also if you tell me how it's slow I can try to optimize it because I like those problems.

drujensen:
I’m not sure why but the creator has benchmarks.  We don’t need full markdown. I guess we can extract what we need and write our own. 😔

asterite:
@drujensen I'm currently optimizing markd, it's fun :-)

So far I went from 11.21 times slower than std's Markdown to 5.40 times slower. It's still slower, but the gap is getting smaller. And I plan to continue optimizing. Of course it won't reach std's performance because std is incomplete and has bugs and implementing the missing stuff will make it slower.

In fact maybe we should eventually just move markd into the standard library somehow and use it for the docs.

straight-shoota:
> In fact maybe we should eventually just move markd into the standard library somehow and use it for the docs.

-> #4613

PR #8116, opened 2019-08-24
Add `Hash#put` and optimize `Set#add?`

asterite:
Right now `Set#add?` is implemented doing:

```crystal
if includes?(object)
  false
else
  add(object)
  true
end
```

That's a bit inefficient because there's a double hash lookup.

It would be cool to be be able to insert a key-value pair in a Hash and know if that was a new value or not (well, what `Set#add?` does) but that method doesn't exist. So here we introduce it as `Hash#store`.

Then we rewrite `Set#add?` using that new method.

Benchmark code:

```crystal
require "benchmark"

set = Set{0}
1000.times do |i|
  set.add?(i)
end

Benchmark.ips do |x|
  x.report("add?") do
    set.clear
    1000.times do |i|
      set.add?(i)
    end
  end
end
```

Results:

```
old: 66.60k ( 15.01µs) (± 2.27%)
new: 99.17k ( 10.08µs) (± 4.75%)
```

Not a big difference but I still think it's bad if this can't be implemented efficiently, and I also think `Hash#put` might be a generally useful method because you can't do that efficiently right now.

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -322,7 +322,8 @@ class Hash(K, V)
   private MAX_INDICES_BYTESIZE_2 = 65536
 
   # Inserts or updates a key-value pair.
-  private def upsert(key, value) : Nil
+  # Returns an `Entry` if it was updated, otherwise `nil`.
+  private def upsert(key, value) : Entry(K, V)?
     # Empty hash table so only initialize entries for now
     if @entries.null?
       @indices_size_pow2 = 3
@@ -333,18 +334,25 @@ class Hash(K, V)
 
     # No indices allocated yet so try to do a linear scan
     if @indices.null?
-      # Try to do an upsert by doing a linear scan
-      upserted = upsert_linear_scan(key, value, hash)
-      return if upserted
+      # Try to do an update by doing a linear scan
+      updated_entry = update_linear_scan(key, value, hash)
+      return updated_entry if updated_entry
 
-      # If we couldn't upsert it means the table was full
-      # so a resize might have been done.
+      # If we still have space, add an entry.
+      if !entries_full?
+        add_entry_and_increment_size(hash, key, value)
+        return nil
+      end
+
+      # No more space so we need to do a resize
+      resize
+
       # Now, it could happen that we are still with less than 16 elements
       # and so `@indices` will be null, in which case we only need to
       # add the key-value pair at the end of the `@entries` buffer.
       if @indices.null?
         add_entry_and_increment_size(hash, key, value)
-        return
+        return nil
       end
 
       # Otherwise `@indices` became non-null which means we can't do
@@ -370,7 +378,7 @@ class Hash(K, V)
         # We have free space: store the index and then insert the entry
         set_index(index, entries_size)
         add_entry_and_increment_size(hash, key, value)
-        break
+        return nil
       end
 
       # We found a non-empty slot, let's see if the key we have matches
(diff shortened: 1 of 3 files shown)


src/hash.cr
```diff
@@ -912,6 +910,21 @@ class Hash(K, V)
...
+  #
+  # If a value already exists for `key`, that (old) value is returned.
+  # Otherwise the given block is invoked and its value is returned.
+  #
+  # ```
+  # h = {} of Int32 => String
+  # h.store(1, "one") { "didn't exist" } # => "didn't exist"
+  # h.store(1, "uno") { "didn't exist" } # => "one"
+  # ```
+  def store(key : K, value : V)
```
```suggestion
  def store(key : K, value : V, &block)
```

asterite: This is not strictly needed.

j8r: IMO it would not bad to become mandatory.
Here this is for the API docs. I forgot to add the type signature, I edit the suggestion.

asterite: API docs include this even if you don't do it in code. If it becomes mandatory I'll add it.

src/hash.cr
```diff
@@ -912,6 +910,21 @@ class Hash(K, V)
...
+    updated_entry = upsert(key, value)
+    updated_entry ? updated_entry.value : yield
```
yxhuvud: `store` might perhaps be considered a symmetric operation to `fetch`. `fetch` yields the key to its block. Should `store` also yield something?

src/hash.cr
j8r: Not the return type:
- without: `def store(key : K, value : V, &block)`
- with: `def store(key : K, value : V, &block : -> U) V | U forall U`

One will know what the method returns without reading the comments.
I only see advantages to add it.

asterite: Yes, that's what I originally wrote. Then I looked at fetch and it has no return type or block annotation.

I understand that the type annotations make it more formal, but at the same time they make the signature more complex. It reminds me of those long Rust function signatures.

I prefer brevity and clarity. It's easy to understand how the method works by reading the example.

src/hash.cr
asterite: Great idea! I just did that, thanks. I forgot `fetch` yielded the key.

jhass:
I think the most common name for the "insert into map" operation is `put`. We could even do `put(key, value) #=> old_value` besides `put(key) { value } #=> old_value` and let '[]=` be `put(key, value); value`. This also brings me to the somewhat weird behavior of `store` returning the old value in case of an existing value and the new value in case of an updated value, I think it might be more consistent to return the default value in the latter case, or just always `nil`  (again,  in the no existing value case) even.

jhass:
I think I wouldn't mind `put`'s return value being `V?` and ignoring the default value, that's already the case for `[]?`.

RX14:
Looks like everyone's agreed on `put`, this can probably be moved forward.

asterite:
Renamed! If we merge we can squash the commits or I can rewrite the history, but maybe it's not bad to keep the commit about the decision to rename `store` to `put`.

RX14:
I'm having second thoughts about this, since I was confused by the name and thought this was a "put this value if it doesn't exist, else do nothing" call, which `Hash` is missing (but seems useful).

This method should probably be `#replace`, and have a version without the block which always returns the previous value, in addition to the current version.

straight-shoota:
Java has `Map.put(key, value)` (equal to `Hash#[]=`) and `HashMap.putIfAbsent(key, value)` which only stores if the key does not exist. The counterpart `Map.replace(key, value)` only stores if the key exists. There is also an overload `Map.replace(key, old_value, new_value)` which only stores if the key is set to `old_value`.

PR #8120, opened 2019-08-27
Arithmetic division for all types

### LINKED ISSUE #2968: Make `/` be float division

bcardiff:
This PR makes `/` work as arithmetic division for all types.

```crystal
1 / 2 # => 0.5
```

As a reminder, for integer or floor division `//` should be used.

```crystal
1 // 2 # => 0
```

Given a binary operator `⨁`, the result type is usually the type of the left hand side: `typeof(a ⨁ b) == typeof(a)`. This helps to reduce the unions in expressions like `a ⨁= b`.

For `/` this rule does not apply since the type might be a different to the operands. We want `anInt32 / anInt32` to be `Float32`.

There are 17 "numeric" types in the stdlib: `[U]Int{8,16,32,64,128}, Float{32,64}, Big{Int,Float,Decimal,Rational}, Complex`. (Actually, `Complex` does not inherit from `Number`.)

I propose `typeof(a / b) == max(fa(typeof(a)), fa(typeof(b)))`, where `fa(t)` defines a float affinity type.

* `Float32` if `t = [U]Int{8,16,32,64}`
* `Float64` if `t = [U]Int128`
  * 1 / `UInt128::MAX` ≈ 2.938e-39 < 1.175e-38 ≈ `Float32::MIN_POSITIVE`
* `BigFloat` if `t = BigInt`
* `t` if `t = Float{32,64},Big{Float,Decimal,Rational},Complex`

And `max` is computed under the partial order inferred by `{Float32 < Float64, Float64 < BigFloat < BigDecimal < BigRational, Float64 < Complex}`

All the combinations can be viewed in the following table:

Screen Shot 2019-08-27 at 13 52 20

The `max` relation is not defined for `Complex` and `Big{Int,Float,Decimal,Rational}`, they are extensions to the basic numbers but they do not interact well within each other. And I don't see us adding `BigRationalComplex` in the std-lib.

Note: I checked if `typeof(a / b) == fa(typeof(a))` would be a good rule, to be more similar to the other operators . But checking how things like `loop { a = a / 2 }` and `loop { a = 2 / a }` would work under one or another make me decide in favor of `typeof(a / b) == max(fa(typeof(a)), fa(typeof(b)))`.

I have a couple of leftovers or further thoughts after these changes:

* The former float division is still needed, but it was only exposed as a primitive through "/". "fdiv" can now be used as primitive method name, but will be available after the release.
* Although Big and Complex can be moved out of the std-lib, having the right behavior was worth it.
* I was not sure if BigDecimal < BigRational or BigRational < BigDecimal. Rational has more precision, but if working with decimal they are usually preferred to stay. But this should also impact other operators. So I stay went with BigDecimal < BigRational.
* It might make sense to move the minimal operators (those that operate within the same types) to a core module and leave the extensions to make things more comfortable in the stdlib. I tried to move the implementation a bit in that direction so see how it feels.
* While reviewing all numbers as a whole there are different criteria applied here and there: are the conversions handled in `to_X` methods or in `initialize`/`self.new`? Is there a generic implementation overridden or implementation should be only in the concrete classes?
* I would like to have more specs related to the contract in numbers in `spec/support/number.cr`

I tried to avoid changing as less as possible, but there is still some work to do in order to get some neat implementation. But for now is an improvement and I needed to fill some missing gaps in `Big*` related to conversions.

Fixes #2968


ysbaddaden:
I second Asterite. I believe the cases where / should return a Float32 is when either lhs or rhs is a Float32 and the other isn't more significant (not Float64, BigX or Complex).

ysbaddaden:
I still don't quite understand why `Int / Float32` would return a `Float64`? Isn't the float type in a binary operation significant?

asterite:
Ah, yes, I think Int / Float32 should be Float32. Float64 should only be the default when two integers are. involved. And when doing Int / Float64 of course.

bcardiff:
@yabaddaden, but [U]Int128 / Float32 => Float64, right?

It’s good that I kept the colored table 😅

asterite:
I don't think [U]Int128 / Float32 should be an exception. Any primitive integer divided by Float32 should be Float32.

bcardiff:
But then to keep the typing rule symmetrical there are some operations that will not fit:

1f32 / UInt128::MAX ≈ 2.938e-39 < 1.175e-38 ≈ Float32::MIN_POSITIVE

It also bugs me a little that if Int / Float32 is Float32, then the typing rule formulation is a bit more complex, but is paperwork on the formality side. I will leave on the side.

asterite:
I think `Float64` by default might not be bad then. You can always convert to Float32 without losing precision. And `Float32` is not common at all.

ysbaddaden:
@bcardiff the same reasoning applies to all other binary operations with a Float32. Lower precision means that the value will be rounded, so `1_f32 + UInt128::MAX` is `Infinity` (actually 3.402823669209385e+38) and `1_f32 / UInt128::MAX` is 0_f32 (actually `2.938735877055719e-39`). I believe this is expected behavior.

I don't understand why `1 + Float32` returns a Float32 but `1 / Float32` would suddenly be changed to return a Float64 with higher precision. That's unexpected. If the rounding/low precision is a problem, one should use a Float64 (the default) not a Float32 in its operations (uncommon).

There are use cases where we want lower precision. While documenting on alternative float sizes (16, 80 an 128-bit) I read that AI / ML algorithms sometimes cast to Float16 to specifically *lose precision*, which reduces memory usage, but also improves results: AI behaves better with blurry results, being too precise is counter productive.

bcardiff:
@ysbaddaden yes I recall the same use cases even losing some precision.

For native types:

* Int / Float should keep the float (whatever it is) and
* FloatA / FloatB should max the floats, 
* IntA / IntB should max using the float affinitiy

For extensions:
* NumA / NumB should max using the float affinitiy

with that, the table will look like follows

Screen Shot 2019-08-29 at 11 57 28

Separating native types and extensions seem needed because otherwise Float64 * BigInt would've been Float64 according to the rules for native types, instead of BigFloat which I think is preferred.

👍 ?

spec/std/int_spec.cr
```diff
@@ -437,7 +437,7 @@ describe "Int" do
   end
 
   it "raises when divides by zero" do
-    expect_raises(DivisionByZeroError) { 1 / 0 }
+    (1 / 0).should eq(Float32::INFINITY)
```
asterite: Should this be `Float64::INFINITY`?

bcardiff: Updated, and also change the description of the specs that used to raise.

spec/std/int_spec.cr
```diff
@@ -385,16 +385,32 @@ describe "Int" do
...
+      (-6 / 2).should eq(-3.0)
+      (6 / -2).should eq(-3.0)
+      (-6 / -2).should eq(3.0)
+    end
 
-    (6 / 2).should eq(3)
-    (-6 / 2).should eq(-3)
-    (6 / -2).should eq(-3)
-    (-6 / -2).should eq(3)
+    it "does not raises when divides by zero" do
```
straight-shoota:
```suggestion
    it "divides by zero" do
```

spec/std/int_spec.cr
```diff
@@ -385,16 +385,32 @@ describe "Int" do
...
+      (1 / 0).should eq(Float64::INFINITY)
+      (4 / 2).should eq(2)
```
straight-shoota: What's the purpose of this example?

spec/std/int_spec.cr
```diff
@@ -385,16 +385,32 @@ describe "Int" do
...
+    end
+
+    it "does not raises when divides Int::MIN by -1" do
```
straight-shoota:
```suggestion
    it "divides Int::MIN by -1" do
```

spec/std/int_spec.cr
bcardiff: leftovers from refactors, removing it since it's covered by others.

spec/std/int_spec.cr
bcardiff: handled and squashed in the last commit

spec/std/int_spec.cr
bcardiff: handled and squashed in the last commit

bcardiff:
Rebased on master and apply some suggested changes in the spec. I triggered a test build so I can check early how good we to ship this in 0.31.0

elorest:
Do other languages do this? I'm so used to 3/2==1 being true.

asterite:
@elorest Nim, Python, Elm, Haskell, Perl, to name a few.

elorest:
@asterite thanks.

j8r:
For the floor division of a Float, it will be logical to return an Integer instead of a Float.
According to [the definition of the floor division](https://en.wikipedia.org/wiki/Floor_and_ceiling_functions):
> [..] the floor function is the function that takes as input a real number x, and gives as output the greatest **integer** less than or equal to x

For now we have `3.0 // 2 #=> 1.0`

What do you think? It will makes sense on the mathematical sense, but maybe not on the Crystal language one, where usually the calling a method on a type returns the same type.

j8r:
For example, in Julia:
```julia
x = floor(1 / 2)
print(typeof(x)) #=> Float64
```
```julia
x = floor(Int, 1 / 2)
print(typeof(x)) #=> Int
```
I guess the actual behavior in Crystal makes also sense, I'm just going to add `to_i32`.

asterite:
The reason is that the range of a Float64 is bigger than that of an Int32 (or Int64). So if you do `big_float // 1` you really want a float as a result because it can't be represented as an integer.

j8r:
Another solution I came across for those wanting more mathematical "correctness" is to use Big.
For my side I won't because of performance reasons.

PR #8121, opened 2019-08-28
Add alias for Set#+ => Set#|

sam0x17:
Adds the ability to use ` Set{1, 2, 3} + Set{3, 4, 5}` syntax as an alias for the little-known `Set{1, 2, 3} | Set{3, 4, 5}` syntax for adding (unioning) two `Set(T)` objects. I have long used a similar monkey-patch to enable idiomatic set addition in my projects.

Includes a spec that merely checks that the alias is in place, since existing specs cover `|`.

This PR is equivalent to this monkey patch:
```crystal
struct Set(T)
  def +(other)
    self | other
  end
end
```

Thanks to @asterite for encouraging me to submit this PR and @Blacksmoke16 and @watzon for vetting the idea in gitter.

### DIFF
--- a/src/set.cr
+++ b/src/set.cr
@@ -208,7 +208,7 @@ struct Set(T)
   # ```
   # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
   # ```
-  def +(other : Set(U))
+  def +(other : Set(U)) forall U
     self | other
   end
 
--- a/src/set.cr
+++ b/src/set.cr
@@ -203,9 +203,7 @@ struct Set(T)
     set
   end
 
-  # Addition: returns a new, unique set containing the elements from both sets.
-  # Because sets cannot have duplicates, this is equivalent to the union of
-  # both sets.
+  # Addition: returns a new set containing the unique elements from both sets.
   #
   # ```
   # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
--- a/src/set.cr
+++ b/src/set.cr
@@ -203,13 +203,13 @@ struct Set(T)
     set
   end
 
-  # Addition: returns a new set containing the elements from both sets.
+  # Addition: returns a new, unique set containing the elements from both sets.
+  # Because sets cannot have duplicates, this is equivalent to the union of
+  # both sets.
   #
   # ```
   # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
   # ```
-  #
-  # Alias for `#|` (Union).
   def +(other)
     self | other
   end
--- a/src/set.cr
+++ b/src/set.cr
@@ -203,6 +203,17 @@ struct Set(T)
     set
   end
 
+  # Addition: returns a new set containing the elements from both sets.
+  #
+  # ```
+  # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
+  # ```
+  #
+  # Alias for `#|` (Union).
+  def +(other)
+    self | other
+  end
+
   # Difference: returns a new set containing elements in this set that are not
   # present in the other.
   #
--- a/src/set.cr
+++ b/src/set.cr
@@ -210,7 +210,7 @@ struct Set(T)
   # ```
   # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
   # ```
-  def +(other)
+  def +(other : Set(U))
     self | other
   end
 
--- a/spec/std/set_spec.cr
+++ b/spec/std/set_spec.cr
@@ -142,6 +142,14 @@ describe "Set" do
     set3.should eq(Set{1, 2, 3, 4, 5, "3"})
   end
 
+  it "aliases + to |" do
+    set1 = Set{1, 1, 2, 3}
+    set2 = Set{3, 4, 5}
+    set3 = set1 + set2
+    set4 = set1 | set2
+    set3.should eq(set4)
+  end
+
   it "does -" do
     set1 = Set{1, 2, 3, 4, 5}
     set2 = Set{2, 4, 6}


oprypin:
But we don't do aliases..?

oprypin:
OK but it is an alias though. And whatever reasons applied to not liking aliases in the first place, apply just as well here.

`&` represents "and", which is the operation applied to each membership in a set when performing an intersection, `Set#&`.
`|` always represents "or", which is the operation applied to each membership in a set when performing a union, `Set#|`.

jhass:
I'm confused, how is this operation not https://en.wikipedia.org/wiki/Union_(set_theory)?

j8r:
Concatenation result to an union for `Set`.
Having to use `|` is good to remind us this.
Having a similar method of `Array#+` will add more confusion, because one may expect a similar result.

Examples to illustrate:
```
[0, 1] | [1, 2] # => [0, 1, 2]
[0, 1] + [1, 2] # => [0, 1, 1, 2]
Set{0, 1} + Set{1, 2} # => Set{0, 1, 2}
```

oprypin:
I can follow the logic: we could always imagine `+` to mean writing out the elements of a collection side by side. like this: `Foo{1,2,3} + Foo{3,4,5} -> Foo{1,2,3 , 3,4,5}`. And then yes, `Set{1,2,3 , 3,4,5}` just happens to be `Set{1,2,3,4,5}`

j8r:
I would like to point also that there are no `Hash#+` nor `Hash#|`. This would bean having two aliases to `Hash#merge`.

What other languages have:
[Ruby Set](https://ruby-doc.org/stdlib-2.6.3/libdoc/set/rdoc/Set.html) has `|` aliasing to `+`
[Python set](https://docs.python.org/3/library/stdtypes.html#set) has only `|`
[Nim set](https://nim-lang.org/docs/sets.html#union%2CHashSet%5BA%5D%2CHashSet%5BA%5D) has `union(a, b)` and `proc +`

Maybe it's fine to have an alias in this case, but they should be made more clear in the API docs. This feature then could be used in IDE through `scry` (I open an issue)

oprypin:
`|` is in the exact same situation as `&`, and I will not accept different treatment for them.

Anyway, the bitwise operation is not so alien to sets as you're trying to make it seem. A set of integers, for example, can be imagined as a huge bit array, where each bit represents presence of one member in the set. In fact, in Pascal that's exactly how sets are implemented (they have horrible limitations on them for that reason). And then union and intersection do exactly match bitwise `|` and `&` even in the way of implementation.

oprypin:
Hey I'm not saying you shouldn't get to use `+` (not anymore, anyway).

Just that `|` must not be removed without also removing `&`.

Sija:
> add `+` as a separate operation instead of alias (or did we want to keep as alias)

IMO that would just needlessly hide the fact it **is** an alias in this case, and an implementation detail, so even if for other container types it's a different operation indeed, it shouldn't rly matter here.

> And I'll also add it for the other collection types that don't have it?

Sounds like a swell idea :)

src/set.cr
```diff
@@ -203,6 +203,17 @@ struct Set(T)
     set
   end
 
+  # Addition: returns a new set containing the elements from both sets.
```
RX14: I'd like to add something which makes it clear elements are not duplicated, and that this operation is the same as set union, *without* mentioning "alias".

src/set.cr
```diff
@@ -203,6 +203,17 @@ struct Set(T)
...
+  #
+  # ```
+  # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
+  # ```
+  #
+  # Alias for `#|` (Union).
```
RX14: I'd remove this, in favour of the above.

src/set.cr
sam0x17: See newest commit.

sam0x17:
@RX14 let me know if you need any other changes -- see newest commit

src/set.cr
```diff
@@ -203,6 +203,17 @@ struct Set(T)
...
   end
 
+  # Addition: returns a new, unique set containing the elements from both sets.
+  # Because sets cannot have duplicates, this is equivalent to the union of
+  # both sets.
+  #
+  # ```
+  # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
+  # ```
+  def +(other)
```
rodrigopinto:
```suggestion
  def +(other : Set(U))
```
The typing keep consistency across the implementations.

src/set.cr
```diff
@@ -203,6 +203,17 @@ struct Set(T)
     set
   end
 
+  # Addition: returns a new, unique set containing the elements from both sets.
+  # Because sets cannot have duplicates, this is equivalent to the union of
+  # both sets.
+  #
+  # ```
+  # Set{1, 1, 2, 3} + Set{3, 4, 5} # => Set{1, 2, 3, 4, 5}
+  # ```
```
rodrigopinto: I do think the documentation could follow the methods `add` and `inspect` and just mention as ```# Alias for `|` ```. 

@RX14 about the docs, the methods `add` and `inspect` are aliases for `<<` and `to_s` respectively. 

Not making clear that `def +(other)` is an alias for the pipe(`Set#|`) might be confusing with `Set#add`, even that implementation and conceptually they are different.

sam0x17: I agree that we should leave the alias note, but am willing to do either -- thoughts @RX14 ?

src/set.cr
```diff
@@ -203,13 +203,13 @@ struct Set(T)
     set
   end
 
-  # Addition: returns a new set containing the elements from both sets.
+  # Addition: returns a new, unique set containing the elements from both sets.
```
RX14: > returns a new set containing the unique elements from both sets

src/set.cr
RX14: This is not an alias, just two operators which happen to have the same semantics on this one datatype.

sam0x17:
@RX14 ready for final review

PR #8128, opened 2019-08-29
Path: don't ask for `Path.home` on expand if not needed

### LINKED ISSUE #8126: Path.expand evaluates Path.home even when no ~ is in the path
### Problem

`Path.expand` evaluates `Path.home` even when it's not needed, which results in this slightly odd behavior:
```
$ crystal eval <<< 'require "uri"; puts File.expand_path("/")'
/
$ unset HOME
$ crystal eval <<< 'require "uri"; puts File.expand_path("/")'
Missing ENV key: "HOME" (KeyError)
```

I discovered this because systemd's `DynamicUser` makes `HOME` be unset.

### Reason

`Path.expand` takes `Path.home` as a default argument: https://github.com/crystal-lang/crystal/blob/36a7557630d696acfe3496adcfdd17286d0e0393/src/path.cr#L560

which will evaluate `Path.home`, which in turn looks up `HOME`:
https://github.com/crystal-lang/crystal/blob/36a7557630d696acfe3496adcfdd17286d0e0393/src/path.cr#L949-L950

Version: master (36a7557630d696acfe3496adcfdd17286d0e0393)

asterite:
Fixes #8126

### DIFF
--- a/src/path.cr
+++ b/src/path.cr
@@ -553,11 +553,12 @@ struct Path
   # Path["baz"].expand("/foo/bar") # => Path["/foo/bar/baz"]
   # ```
   #
-  # *home* specifies the home directory which `~` will expand to.
+  # *home* specifies the home directory which `~` will expand to. If not given
+  # (or `nil` is given) then `Path.home` will be used.
   # If *expand_base* is `true`, *base* itself will be exanded in `Dir.current`
   # if it is not an absolute path. This guarantees the method returns an absolute
   # path (assuming that `Dir.current` is absolute).
-  def expand(base : Path | String = Dir.current, *, home = Path.home, expand_base = true) : Path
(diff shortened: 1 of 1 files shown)


src/path.cr
```diff
@@ -568,9 +569,9 @@ struct Path
     name = @name
```
```suggestion
     name = @name
     home ||= Path.home
```

ysbaddaden: Wouldn't that force `Path.home` to always be evaluated? which is precisely what we want to avoid here?

Sija: Ouch, you're right 👍

straight-shoota:
I guess it's fine to use `nil` as a value for lazy `Path.home` evaluation. But it's maybe not very intuitive. In #7903 we discussed making `nil` mean to disable home path expansion.

PR #8133, opened 2019-08-31
Add method to get an array of types a module is directly included in

Blacksmoke16:
```cr
module Foo
end

module Baz
  module Tar
    include Baz
  end
end

abstract class Parent
end

class Bar < Parent
  include Foo
  include Baz
end

struct Str
  include Baz
end

struct Gen(T)
  include Baz
end

abstract struct AStr
  include Baz
end

abstract class ACla
  include Baz
end

class SubT(T)
  include Baz
end

class ChildT(T) < SubT(T)
end

{{Baz.all_includers.map &.name.stringify}} # => ["Baz::Tar", "Bar", "Str", "Gen(T)", "AStr", "ACla", "SubT(T)"]
```

Questions:
* ~How do I test this?~
* ~Are there any other types that can include a module i should try?~
* ~Any other ideas for a better name?~

I also noticed that types with generics include the generic type vars in their name.  I would have thought `Gen(T)` would have just been `Gen`...not sure if this is intended or not.

EDIT:  I added it to `macros.cr`, will include that in feedback fixes.

### DIFF
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -1534,6 +1534,8 @@ module Crystal
         interpret_argless_method(method, args) { TypeNode.subclasses(type) }
       when "all_subclasses"
         interpret_argless_method(method, args) { TypeNode.all_subclasses(type) }
+      when "all_includers"
+        interpret_argless_method(method, args) { TypeNode.including_types type }
       when "constants"
         interpret_argless_method(method, args) { TypeNode.constants(type) }
       when "constant"
@@ -1562,8 +1564,6 @@ module Crystal
         fetch_annotation(self, method, args) do |type|
           self.type.annotation(type)
         end
-      when "included"
-        interpret_argless_method(method, args) { TypeNode.including_types type }
       when "annotations"
         fetch_annotation(self, method, args) do |type|
           annotations = self.type.annotations(type)
@@ -1669,14 +1669,15 @@ module Crystal
     end
 
     def self.including_types(type)
-      if type.is_a? NonGenericModuleType
+      case type
+      when NonGenericModuleType, GenericModuleType
         return empty_no_return_array unless types = type.raw_including_types
         return ArrayLiteral.map(types) do |including_type|
           TypeNode.new including_type
         end
+      else
+        empty_no_return_array
       end
-
-      empty_no_return_array
     end
 
     def self.type_vars(type)
--- a/src/compiler/crystal/macros/methods.cr
+++ b/src/compiler/crystal/macros/methods.cr
@@ -1562,6 +1562,8 @@ module Crystal
         fetch_annotation(self, method, args) do |type|
           self.type.annotation(type)
         end
+      when "included"
+        interpret_argless_method(method, args) { TypeNode.including_types type }
       when "annotations"
         fetch_annotation(self, method, args) do |type|
           annotations = self.type.annotations(type)
@@ -1664,6 +1666,17 @@ module Crystal
       else
         super
       end
+    end
+
+    def self.including_types(type)
+      if type.is_a? NonGenericModuleType
+        return empty_no_return_array unless types = type.raw_including_types
+        return ArrayLiteral.map(types) do |including_type|
+          TypeNode.new including_type
+        end
+      end
+
+      empty_no_return_array
     end
 
     def self.type_vars(type)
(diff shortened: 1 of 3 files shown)


kalinon:
Awesome. this will be quite useful if merged. I cant think of anything off the top of my head that also includes modules.

sam0x17:
idea: you could make a macro that generates a bunch of arbitrary modules to test

asterite:
Any use case for this?

Blacksmoke16:
Mainly as a compliment to `#subclassess`.  ATM only way to get an array of types is using that.  This allows for using modules to achieve a similar effect if inheritance isn't the best solution.

Ref https://forum.crystal-lang.org/t/type-all-includers/254

src/compiler/crystal/macros/methods.cr
```diff
@@ -1654,6 +1656,17 @@ module Crystal
       end
     end
 
+    def self.including_types(type)
+      if type.is_a? NonGenericModuleType
```
asterite: I think this logic also applies to `GenericModuleType` (but you'll have to do it in a separate branch with duplicate logic).

src/compiler/crystal/macros/methods.cr
```diff
@@ -1522,6 +1522,8 @@ module Crystal
         interpret_argless_method(method, args) { TypeNode.subclasses(type) }
       when "all_subclasses"
         interpret_argless_method(method, args) { TypeNode.all_subclasses(type) }
+      when "all_includers"
+        interpret_argless_method(method, args) { TypeNode.all_includers type }
```
Sija: Pedantic - to match the coding style used around.

```suggestion
        interpret_argless_method(method, args) { TypeNode.all_includers(type) }
```

RX14:
How well does this work with querying generic modules?

Blacksmoke16:
@RX14 Seems to work fine.

```cr
module Bar(T)
end

module Foo
  include Bar(String)
end

module Baz
  include Bar(Bool)
end

{{Bar.all_includers.map &.name.stringify}} # => ["Foo", "Baz"]
```

asterite:
I think the main issue is what happens if a module is included by a generic type. It's probably the same issue we have with `subclasses`. That's why I'm not sure how and whether this should be added to the macro methods.

Blacksmoke16:
```cr
module Bar(T)
end

class Test(T)
  include Bar(T)
end

{{Bar.all_includers.map &.name.stringify}} # => ["Test(T)"]
```

Is what seems to happen.  Same thing as `#subclasses`.

kalinon:
> ```crystal
> module Bar(T)
> end
> 
> class Test(T)
>   include Bar(T)
> end
> 
> {{Bar.all_includers.map &.name.stringify}} # => ["Test(T)"]
> ```
> 
> Is what seems to happen. Same thing as `#subclasses`.

I think this is acceptable behavior.

Blacksmoke16:
Anything left to do here?

Blacksmoke16:
Bump, this should be good to review.

src/compiler/crystal/macros/methods.cr
```diff
@@ -1654,6 +1656,18 @@ module Crystal
       end
     end
 
+    def self.all_includers(type)
+      case type
+      when NonGenericModuleType, GenericModuleType
+        return empty_no_return_array unless types = type.raw_including_types
```
straight-shoota: The main logic should not be hidden in a suffix condition. Please change this to either two separate expressions:
```suggestion
        types = type.raw_including_types
        return empty_no_return_array unless types
```
Alternatively, change the order of condition and effect:
```suggestion
        types = type.raw_including_types || return empty_no_return_array
```

src/compiler/crystal/macros/methods.cr
```diff
@@ -1654,6 +1656,18 @@ module Crystal
...
+        return ArrayLiteral.map(types) do |including_type|
```
straight-shoota: `return` is implicit here:

```suggestion
        ArrayLiteral.map(types) do |including_type|
```

Blacksmoke16:
Also FWIW, https://github.com/crystal-lang/crystal/pull/8483 will allow you to skip the generic type vars of the module (for both this and `#subclasses/#all_subclasses` (ref: https://github.com/crystal-lang/crystal/pull/8133#issuecomment-529950943)

Blacksmoke16:
@asterite @RX14 @bcardiff Anyone able to do another review of this?  Would be :100: if this could make it into `0.32.0`.

RX14:
I want ary or brian to sign off on this.

Blacksmoke16:
After working with this locally a bit I'm starting to think I should rename it to `#includers`, since this only returns types that directly include the module, not children of those types.

I'll push that tonight.

Blacksmoke16:
@asterite @bcardiff Bump

PR #8134, opened 2019-08-31
XML:: rename XML::Type to XML::Node::Type, introduce XML::Reader::Type

### LINKED ISSUE #8129: XML::Type has wrong names
I noticed last night that `XML::Type` member names seem to not be correct.  For example:

https://play.crystal-lang.org/#/r/7gwt

Based on https://www.gnu.org/software/dotgnu/pnetlib-doc/System/Xml/XmlNodeType.html it looks like the `DTD_NODE` should be like `SIG_WHITESPACE` and `ELEMENT_DECL` should be `END_ELEMENT`, etc.

asterite:
Fixes #8129

We incorrectly had a single enum for xml node types when in reality there are two enums: one for node types for `XML::Node`, another for the node types returned by `XML::Reader`.

Marking as a breaking change because `XML::Type` is renamed to `XML::Node::Type`, and the type of `XML::Reader#node_type` changed too.


src/xml/node.cr
```diff
@@ -172,27 +172,27 @@ struct XML::Node
...
-    when XML::Type::HTML_DOCUMENT_NODE then io << "HTMLDocument"
-    when XML::Type::DTD_NODE           then io << "DTD"
-    when XML::Type::ELEMENT_DECL       then io << "Element"
-    when XML::Type::ATTRIBUTE_DECL     then io << "AttributeDecl"
-    when XML::Type::ENTITY_DECL        then io << "EntityDecl"
-    when XML::Type::NAMESPACE_DECL     then io << "NamespaceDecl"
-    when XML::Type::XINCLUDE_START     then io << "XIncludeStart"
-    when XML::Type::XINCLUDE_END       then io << "XIncludeEnd"
-    when XML::Type::DOCB_DOCUMENT_NODE then io << "DOCBDocument"
+    when XML::Node::Type::ELEMENT_NODE       then io << "Element"
```
Sija: Why not use shorter `.element_node?` notation?

ditto elsewhere

asterite: I stated doing that but then kept it simple. It was easier to do the rename like this and from a user perspective there's no difference. I didn't want to introduce extra changes here. But we could do this in a separate PR.

Issue #7967, opened 2019-07-16
Wrong offsetof result in reference types


bcardiff:
The expression `offsetof(AClass, @ivar)` does not consider that the first element of the llvm struct used to represent an instance stores the type id. For value types the result is fine.

```cr
class Foo
  @x = 1
  @y = 2
end

x_offset = offsetof(Foo, @x) # => 0
y_offset = offsetof(Foo, @y) # => 4

f = Foo.new
ptr = f.as(Void*)

Foo.crystal_instance_type_id       # => 262
(ptr + x_offset).as(Int32*).value  # => 262 (the type_id value)
(ptr + y_offset).as(Int32*).value  # => 1 (the value of @x)
```

https://play.crystal-lang.org/#/r/78hv

Crystal 0.29.0

bcardiff:
Fixing https://github.com/crystal-lang/crystal/pull/7589/files#diff-7fe1f16e8917382c1e83ed318f73e8e8R2603 and adding a `compiler/codegen/offsetof_spec.cr` to check the actual value would be great if someone wants to fix a bug in the compiler.

PR #8137, opened 2019-09-02
take type id into account when calling offsetof on reference types

mcr431:
Previously all offsets of reference types were off by 4 bytes.
This changes the offsetof implementation for reference types to
add an additional 4 bytes and adds tests that ensure using the
offset on classes and structs result in access to the intended
variable.

see https://github.com/crystal-lang/crystal/issues/7967 for context

*Note this is my first time contributing and I'm not super familiar with the entirety of the codebase. If something should done in a different way, please let me know. This is also my first time working with an LLVM project. I wasn't sure if the type_id is the crystal one which seems to be of type Int32, or something internal to LLVM like this (https://llvm.org/doxygen/classllvm_1_1Metadata.html#a91c7b9c7cf6694f41b9030429b582d26) which is an unsigned int.

### DIFF
--- a/spec/compiler/codegen/offsetof_spec.cr
+++ b/spec/compiler/codegen/offsetof_spec.cr
@@ -1,0 +1,35 @@
+require "../../spec_helper"
+
+describe "Code gen: offsetof" do
+  it "returns offset allowing manual access of first struct field" do
+    code = "struct Foo; @x = 42; def x; @x; end; end;
+            f = Foo.new
+            (pointerof(f).as(Void*) + offsetof(Foo, @x).to_i64()).as(Int32*).value == f.x"
+
+    run(code).to_b.should be_true
+  end
+
+  it "returns offset allowing manual access of struct field that isn't first" do
+    code = "struct Foo; @x = 1; @y = 42; def x; @x; end; def y; @y; end; end;
+            f = Foo.new
+            (pointerof(f).as(Void*) + offsetof(Foo, @y).to_i64()).as(Int32*).value == f.y"
+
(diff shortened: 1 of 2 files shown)


asterite:
I think this is wrong. It's not always 4 bytes. It depends on padding. For example if you have type_id + string, in 64 bits that's padded to 8 bytes and so the offset will be 8 bytes.

I think you just need to add 1 to the offset in case it's a class, that's all.

bcardiff:
Thanks @mcr431 for working on this 💚 .

As pointed by @asterite adding a `+ 1` to de element_index when the type is a class (ie: reference type) is the right approach.

Regarding the specs, I would suggest writing asserts based on the values of the ivars and not teir offsets. That way they are less sensitive to changes in the structure layout.

May I ask, how did you find/pick the #7967? I'm curious.

mcr431:
Made the + 1 to element_index change but haven't addressed spec changes yet. Will update later today.

@bcardiff I picked this issue up because it seemed fairly straight forward, had the community:newcomer tag, and you had left a comment with a starting place in the comments so seemed like a good issue to gently get familiar with the compiler.

mcr431:
Made changes to the spec file. not sure if it was exactly what you were intending. I removed the first four tests since they were essentially testing the same thing as the last four but were relying on implementation details rather than use case. I updated the last four to use ivars instead of hard coded values.

bcardiff:
Thanks @mcr431! Happy first-time contribution!

bew:
Nice fix! Would have been nice to have a little comment in the code explaining the + 1 though

PR #8143, opened 2019-09-03
Compiler: check whether rescue var is closured

### LINKED ISSUE #8141: KeyError when begin-rescue blocks are adjacent
I found an odd little compiler bug in some code I was writing. Here's the smallest amount of code I was able to reproduce it with:

## test.cr
```crystal
def invoke(&block)                                                              
   block.call                                                                    
end                                                                                 
                                       
begin                     
rescue ex                       
end                             
                                                                                          
invoke do                                 
   begin                        
   rescue ex                             
   end                          
end
```

This also produces the error (the begin-rescue blocks are now nested):

## test2.cr
```crystal
def invoke(&block)                                                              
   block.call                                                                    
end                                                                                 
                                       
begin                     
rescue ex
   invoke do                                 
      begin                        
      rescue ex                             
      end                          
   end                       
end
```

## Error message
```
$ crystal build test.cr --threads=1
Missing hash key: "ex" (KeyError)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from __crystal_main
  from main
  from __libc_start_main
  from _start
  from ???
```

## Note
The error only seems to occur when both rescue keywords use the same variable name. Changing the second one to `exc`, for example, fixes it.

```sh
$ crystal -v
Crystal 0.30.0 (2019-08-06)
LLVM: 8.0.1
```

Operating system: Manjaro i3

I'm afraid it's far above my familiarity with the language to even speculate what's going on here, so I'll leave it to you crystal hot shots! Happy to provide more info if you have trouble reproducing this bug.

asterite:
Fixes #8141

### DIFF
--- a/spec/compiler/codegen/exception_spec.cr
+++ b/spec/compiler/codegen/exception_spec.cr
@@ -1301,4 +1301,22 @@ describe "Code gen: exception" do
       foo
       )).to_i.should eq(5)
   end
+
+  it "closures rescue variable (#8141)" do
+    codegen(%(
+      require "prelude"
+
+      def invoke(&block)
+        block.call
+      end
+
+      ex = nil
+
+      invoke do
+        begin
+        rescue ex
+        end
+      end
+    ))
+  end
 end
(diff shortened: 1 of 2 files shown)


bew:
What does this `check_closure` method do in understandable english? :pray:

asterite:
Checks whether the variable is in an enclosing scope. For example:

```crystal
def capture(&block)
  block
end

foo = 1

capture do
  # here we check whether `foo` appears outside the block,
  # if the block is captured. If so then it will need to be allocated
  # on the heap and pass via a hidden pointer to the block.
  puts foo
end
```

PR #8144, opened 2019-09-03
Remove unsafe Time contructor

straight-shoota:
#8112 added a new unsafe `Time` constructor to avoid a circular dependency on `UNIX_EPOCH` (see https://github.com/crystal-lang/crystal/pull/8112/commits/975c08f11041142164e2f77ef3309cb1844a0fe5). But this is unncessary because there is already a constructor that does not rely on `UNIX_EPOCH`, we can just use that one for the constant value.

/cc @waj

### DIFF
--- a/src/time.cr
+++ b/src/time.cr
@@ -268,7 +268,7 @@ struct Time
   # ```
   # Time.utc - Time::UNIX_EPOCH
   # ```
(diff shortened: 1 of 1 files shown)


straight-shoota:
Oh, it doesn't work because the safe constructor uses `offset_seconds` which in turn calls `to_unix`.

Issue #8140, opened 2019-09-03
puts in a multithreaded environment


asterite:
Now that we have multiple threads, running this program:

```crystal
100.times do |i|
  spawn do
    puts i
  end
end

sleep 0.1
```

Will output stuff like this:

```
2
6
0
8
4
1
13
105

149
...
```

Note that there are multiple newlines together, and some numbers are mixed.

The reason is that `puts(obj)` is implemented by doing `print(obj)` followed by `print '\n'` and so a thread might switch between these two calls.

There are multiple solutions to this. For example:
- let `puts(obj)` just call `print "#{obj}\n"`, but that involves a memory allocation for the string, at least in this simple form. To avoid this allocation Go for example has a pool of buffers that it uses to write to first, and then dumps this buffer into the final IO.
- use a mutex to synchronize the two calls

There are probably other solutions to this. Maybe we can choose the simplest one and eventually improve its performance.

The question is: is this expected behavior? Do we want to change it? I think that from the outside I would expect a call to `puts(obj)` to look kind of atomic.

/cc @waj

bcardiff:
The array to_s algorithm needs to be reviewed also since the detection of recursive values is not thread safe

sam0x17:
My thought is that this is the expected behavior.

asterite:
@bcardiff I think for that one we just need to change the current approach to use a thread local variable. And now that I think about it, maybe we can use one too for `puts`: a buffer per thread that we write to first, then we copy to the IO.

waj:
Let’s discuss one issue at a time ;)
I agree that `puts` behavior should be atomic, at least for tty. Having a pool of buffers sounds great but I’d only add it if we figure out the performance degrades (do not underestimate the GC that already manages pools of memory blocks)
It’s sad that we need to allocate an extra buffer for this but the option is use a Mutex for every operation. Or have atomic/non atomic IOs. Just thinking...

straight-shoota:
For a TTY, performance is not critical, so allocating and additional string buffer shouldn't be a huge issue.

For a more performant implementation, it might be worth to revisit vectored IO (#6078).

RX14:
Big +1 to vectored IO from me. `read()` and `write()` should support vectored IO overloads.

[...]

PR #8146, opened 2019-09-04
Let exec_recursive use a thread-local data structure

asterite:
ExecRecursive.hash might be used by different threads at the same time so we have to make sure its accessed is bound to the current thread.

Addresses https://github.com/crystal-lang/crystal/issues/8140#issuecomment-527260754

### DIFF
--- a/src/reference.cr
+++ b/src/reference.cr
@@ -135,8 +135,20 @@ class Reference
 
   # :nodoc:
   module ExecRecursive
+    alias Registry = Hash({UInt64, Symbol}, Bool)
+
+    {% if flag?(:preview_mt) %}
+      @@exec_recursive = Crystal::ThreadLocalValue(Registry).new
+    {% else %}
+      @@exec_recursive = Registry.new
+    {% end %}
+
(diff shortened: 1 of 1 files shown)


src/reference.cr
```diff
@@ -135,8 +135,21 @@ class Reference
...
+    {% if flag?(:preview_mt) %}
+      @@exec_recursive : Crystal::ThreadLocalValue(Registry)?
+    {% else %}
+      @@exec_recursive : Registry?
+    {% end %}
+
     def self.hash
-      @@exec_recursive ||= {} of {UInt64, Symbol} => Bool
+      {% if flag?(:preview_mt) %}
+        exec_recursive = @@exec_recursive ||= Crystal::ThreadLocalValue(Registry).new
```
waj: This initialization should be protected by a Mutex or moved as a class variable initializer. Otherwise a race condition will happen and multiple instances of `Crystal::ThreadLocalValue` are created.

asterite: Fixed!

PR #8147, opened 2019-09-04
[WIP] Synchronize logger severities with RFC 5424.

didactic-drunk:
Also attempt to provide compatibility with other logger implementations by @Blacksmoke16 and @watzon.

### Problem scope

Logger severities aren't portable between different crystal loggers.  This tends to lock applications in to a particular logger or reduce the number of message types they can log.

### Severities differences

EMERG in [crylog](https://github.com/Blacksmoke16/crylog/blob/ee97471bb247a1b415d38ec24c0e4eb771ff40cd/src/logger.cr#L35) and [strange](https://github.com/hydecr/strange/blob/92a1397575e33b60ab899c12d83f2866af1a3352/src/strange/formatter/color_formatter.cr#L34).
EMERG missing in crystal.

WARNING in crylog and strange.
WARN in crystal.

CRITICAL missing in crystal.
ALERT missing in crystal.
NOTICE missing in crystal.

FATAL in crystal isn't described in the RFC.  If kept where does it fit?  Higher or lower than EMERGENCY?

UNKNOWN in crystal.  What purpose does it serve?

### Numerical order

Sometimes in filters (like color) it's useful to check severity > INFO.  The RFC numbers seem backwards when used for this purpose.  Numerical syslog compatility may not be necessary since it's not popular in any serialized format and they aren't used except when making `syslog()` calls.  An output class can easily map `severity name -> syslog  name` when there is a 1 <-> mapping.

### Customized logging levels with sane defaults?

Should there be a way to customize logging levels for compatility with other applications/libraries?

### Personal preferences

When I started writing this I only cared about the portability issues from a programmers perspective.  Writing `.emerg` or `.emergency` is not worth fighting over.  Thinking about it more reminded me of hours and hours of combing through log files and always significantly slowed by shifts in name length.

Example output:
```
[2019-09-03T20:28:00.032959000Z] DEBUG: foo bar baz
[2019-09-03T20:28:00.034013000Z] CRITICAL: baz foo
[2019-09-03T20:28:00.034019000Z] ALERT: baz
[2019-09-03T20:28:00.034024000Z] ERROR: baz foo
[2019-09-03T20:28:00.034029000Z] CRITICAL: baz foo bar
[2019-09-03T20:28:00.034034000Z] INFO: baz
[2019-09-03T20:28:00.034039000Z] EMERGENCY: foo bar baz
[2019-09-03T20:28:00.034044000Z] NOTICE: bar baz
```

EMERGENCY is particularly jarring when attempting to read multiple log messages that are related to each other.  Wanting EMERGENCY to stick out more seems natural at first, except people don't read logs start to end.  They search them.  Many tools or loggers color the output highlighting important messages like EMERG and CRIT.

Shorter names such as EMERG and CRIT throw off formatting less and allow more data per line when reading logs.

### Single character spacing.

One extreme option is using a single character to represent levels with uniform spacing.

EMERGENCY and DEBUG in perfect harmony with uniform spacing.
```
[2019-09-03T20:28:00.034959000Z] N: foo bar baz
[2019-09-03T20:28:00.034013000Z] E: baz foo
[2019-09-03T20:28:00.034019000Z] E: baz
[2019-09-03T20:28:00.034024000Z] D: baz foo
[2019-09-03T20:28:00.034029000Z] W: baz foo bar
[2019-09-03T20:28:00.034034000Z] E: baz
[2019-09-03T20:28:00.034039000Z] E: foo bar baz
[2019-09-03T20:28:00.034044000Z] D: bar baz
```

### Next steps

1. Civilized discussion
2. Bike shedding
3. Battle Royal

### DIFF
--- a/src/logger.cr
+++ b/src/logger.cr
@@ -90,11 +90,23 @@ class Logger
     # Generic (useful) information about system operation
     INFO
 
+    # Normal but significant conditions
+    NOTICE
+
     # A warning
     WARN
 
     # A handleable error condition
     ERROR
+
+    # Critical conditions
+    CRITICAL
+
+    # Action must be taken immediately
+    ALERT
+
+    # System is unusable
+    EMERGENCY
 
     # An unhandleable error that results in a program crash
     FATAL


watzon:
I'd be fine with just skipping right to the battle royal. No need for niceties :wink:

straight-shoota:
For reference: https://tools.ietf.org/html/rfc5424#page-11

This sounds like a good idea.
Let's just remove `FATAL` in favour of `EMERGENCY`/`ALERT`. I'd also assign the specified numerical codes.

I don't think we need any standardization regarding abbreviated identifiers. Representation should lie in the responsibility of individual formatter implementations. Using single characters is just `severity.to_s[0]`.

asterite:
The RFC seems to be about syslog. Does every logger follow syslog convention? Logging in rust doesn't. Logging in Go doesn't. Logging in Ruby doesn't (unless you use the syslog module).

Im not saying we shouldn't do it. I just wonder why rfc 5424 isn't honored in other languages.

straight-shoota:
Syslog and application logging surely have different requirements. But so do different kinds of applications. Some might do fine with maybe two or three severity states, others need more.
It's probably unlikely that an application would need three different fatality states. Maybe it's a good simplification to combine them into `FATAL`. But it's not a huge gain. And I think it can be left up to the individual developers, which ones to use.

Building on a standardized specification makes sense IMO if if it is for a slightly different use case. But I don't see a reason why it wouldn't fit for application logging as well. You don't need to use the full set, if you don't need it.

straight-shoota:
In addition to the differences mentioned in the OP, `Logger::Severity` doesn't have a `NOTICE` value either.

And I'd question the usefulness of the `UNKNOWN` value. Is there a specific reason for this or is it just there because Ruby has it? If there is a place where a severity might not be known at all times, it could just be nilable.

didactic-drunk:
> Some might do fine with maybe two or three severity states, others need more.
> It's probably unlikely that an application would need three different fatality states.

[Some need more](https://www.youtube.com/watch?v=Ku_Z7U8Zhak)

didactic-drunk:
Heavier loggers like log4j, log4r often use RFC 5424 or remain very close to it.  Maintaining compatibility with RFC 5424 makes it easy for applications to use syslog without modification, although i admit the use cases for that are getting rarer.

Most rust loggers have a `.trace` level.  That doesn't fit with the RFC.

In some ways the RFC seems like a throw back to the 80's in the wrong directly.  > 3 failure states noting how to notify the "operator" says the machine is more expansive than a human and every message is constantly monitored.

On the opposite end many debugging workflows involve adding more v's or d's to `-v` or -`-d` to indicate more logging so you can fix a f'n bug.  

One approach I've used in the past:
```
def debug(level) # 
  logger.debug block if level > @debug_level
end
```

Should that be built in to the logger interface?  Is that what crylog's spaced number severities are for?

I didn't mention it before because it's closer to bike shedding and the unavoidable battle royale.

Blacksmoke16:
> Is that what crylog's spaced number severities are for?

Essentially yes, see [this](https://github.com/Blacksmoke16/crylog/blob/master/src/handlers/abstract_handler.cr#L18-L20) and [this](https://github.com/Blacksmoke16/crylog/blob/master/src/logger.cr#L81) for implementation, and the [docs](https://github.com/Blacksmoke16/crylog/tree/master/docs#handles) for how its used.

I pretty much used what [Monolog](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php) had defined.  See https://github.com/Seldaek/monolog/issues/138#issuecomment-11570214.

Sija:
CI in MT mode failed :(

```
Invalid memory access (signal 11) at address 0x7f56be7fe000

[0x10e5c86] *CallStack::print_backtrace:Int32 +118
[0xb34499] __crystal_sigfault_handler +217
[0x25cdbb7] sigfault_handler +40
[0x7f56e6b90390] ???
[0x25c1b50] GC_push_all_eager +64
[0x25ca1fd] GC_push_all_stacks +141
[0xb1f823] ~procProc(Nil) +275
[0x25c20ad] GC_mark_some +141
[0x25b9add] GC_stopped_mark +189
[0x25ba5c9] GC_try_to_collect_inner +217
[0x25bafbc] GC_collect_or_expand +476
[0x25bea4b] GC_alloc_large +187
[0x25beeac] GC_generic_malloc +348
[0x25bf0bc] GC_malloc_kind_global +252
[0x113e196] *GC::malloc_atomic<UInt64>:Pointer(Void) +6
[0x113e18e] *GC::malloc_atomic<UInt32>:Pointer(Void) +14
[0x1562208] *Socket+ +120
[0x15620d3] *Socket+ +19
[0x1562fc1] *Socket+ +49
[0x14cc53b] *IO+ +2027
[0x14cbd3e] *IO+ +238
[0x23d4e7b] *HTTP::Client::Response::from_io?<IO+, Bool, Bool>:(HTTP::Client::Response | Nil) +107
[0x2465fd0] *HTTP::Client#exec_internal_single<HTTP::Request>:(HTTP::Client::Response | Nil) +80
[0x2465cbc] *HTTP::Client#exec_internal<HTTP::Request>:HTTP::Client::Response +44
[0x2465c86] *HTTP::Client#exec<HTTP::Request>:HTTP::Client::Response +6
[0x24677c9] *HTTP::Client#post:form:headers<String, String, HTTP::Headers>:HTTP::Client::Response +137
[0x2467702] *HTTP::Client::post:form:headers<String, String, HTTP::Headers>:HTTP::Client::Response +770
[0x25004be] *OAuth2::Client#get_access_token_using_client_credentials<String>:OAuth2::AccessToken+ +334
[0xfd4119] ~procProc(Nil) +873
[0xb49b4f] *it<String, String, Int32, Int32, &Proc(Nil)>:(Array(Spec::Result) | Nil) +879
[0xfd2cba] ~procProc(Nil) +250
[0x20dd21d] *Spec::RootContext::describe<String, String, Int32, &Proc(Nil)>:Spec::Context+ +365
[0xb4d486] *describe<String, String, Int32, &Proc(Nil)>:Spec::Context+ +6
[0xfd22cc] ~procProc(Nil) +172
[0x20dd21d] *Spec::RootContext::describe<String, String, Int32, &Proc(Nil)>:Spec::Context+ +365
[0xfd4da7] *describe<OAuth2::Client.class, String, Int32, &Proc(Nil)>:Spec::Context+ +55
[0xb0fccd] __crystal_main +22717
[0x1c85486] *Crystal::main_user_code<Int32, Pointer(Pointer(UInt8))>:Nil +6
[0x1c853e9] *Crystal::main<Int32, Pointer(Pointer(UInt8))>:Int32 +41
[0xb1c5d6] main +6
[0x7f56e5d6a830] __libc_start_main +240
[0xb0a339] _start +41
[0x0] ???

Makefile:99: recipe for target 'std_spec' failed
make: *** [std_spec] Error 11

on_os linux docker run --rm -t -u 1001 -v /home/circleci/project:/mnt -v /etc/passwd:/etc/passwd -v /etc/group:/etc/group -w /mnt -e CRYSTAL_CACHE_DIR=/tmp/crystal crystallang/crystal:0.30.1-build linux64 /bin/sh -c 'CRYSTAL_WORKERS=4 make std_spec threads=1 verbose=1 FLAGS="-D preview_mt"' exited with 2

Exited with code 1
```

asterite:
> CI in MT mode failed :(

That's why it's behind a flag: it's experimental :-)

I wouldn't worry too much about it right now.

RX14:
I'm fine with removing fatal and unknown.

didactic-drunk:
Does anyone want to add TRACE?  It's extremely common with other loggers in many languages.

Blacksmoke16:
I wouldn't be opposed to it if people think it would be helpful to have.

ysbaddaden:
I'll welcome NOTICE. I sometimes wish there was a level between INFO and WARN, in order to have debug level (DEBUG), verbose level (INFO) and normal level (NOTICE) in applications.

That could also be achieved with TRACE, thought. We should only introduce one, so either NOTICE or TRACE, but not both.

UNKNOWN comes from Ruby: a *message that should always be logged*. I never had a need for it. I'd actually prefer the opposite: a SILENT mode where *nothing* is ever logged.

I never had any need for FATAL. Having more levels by adding CRITICAL, ALERT and EMERGENCY feels *really* outstretched. If an application reaches those levels, it feels like it *already* crashed, and such logs would never be logged. I prefer a single FATAL level, simple and descriptive... maybe it could act as UNKNOWN and *always* be logged (even if level if SILENT)?

didactic-drunk:
> That could also be achieved with TRACE, thought. We should only introduce one, so either NOTICE or TRACE, but not both.

NOTICE is not a replacement for TRACE.  NOTICE is between INFO and WARN.  TRACE is less than DEBUG.

Rust log:
`ERROR > WARN > INFO > DEBUG > TRACE`

Log4J:
`FATAL > ERROR > WARN > INFO > DEBUG > TRACE`

TRACE is often recomended for nonproduction or short burst levels of logging that can't be left on due to disk space concerns.

NOTICE is incredibly useful.  I use it all the time.

didactic-drunk:
My current recommendation based on most of the feedback:
`FATAL > ERROR > WARN > NOTICE > INFO > DEBUG`

@straight-shoota and @ysbaddaden said a single `FATAL` level for application crash is or may be acceptable .

`NOTICE` could be the default logger application level for most applications with `INFO` and `DEBUG` providing additional granularity.

I don't know about `TRACE`.

ysbaddaden:
I didn't mean for TRACE to replace or be the same than NOTICE. I was referring to the trio normal / verbose / debug (or very verbose) levels, that can be mapped to either INFO / DEBUG / TRACE *or* NOTICE / INFO / DEBUG, hence why I believe we should only have either TRACE or NOTICE but not both.

I'm fine with FATAL > ERROR > WARN > NOTICE > INFO > DEBUG. I'd love a SILENT level, too.

straight-shoota:
`SILENT` would be a filter mode, not a severity assigned to a message. I'm not sure if that should be a `Severity` value. I'd rather leave it as a configuration for filter/formatter implementations. The current `Logger` implementation supports this essentially by setting `@io = nil`.

ysbaddaden:
SILENT as a severity makes it *very* easy to implement, by setting its value to -1 (leading to no extraneous silent check in addition to severity check) then, obviously, don't implement `.silent` log methods :)

didactic-drunk:
> SILENT as a severity makes it _very_ easy to implement, by setting its value to -1 (leading to no extraneous silent check in addition to severity check) then, obviously, don't implement `.silent` log methods :)
> 
> It's also easy to set and avoids extraneous configuration: `logger.level = Logger::SILENT`. Done.

Is FATAL sufficient?  Are there cases when you wouldn't want to log a fatal error before crashing?

Issue #7981, opened 2019-07-21
JSON#mapping with custom converters and array types


HCLarsen:
When trying to parse JSON with an array of time objects, and using a custom converter, I'm getting an error that #to_json doesn't output an array. I would have expected the code generated from JSON#mapping macro to know to apply the converter to each individual element in the array, but it seems otherwise. I'm not sure, is this the intended functionality, or is it a bug?

Sample code:

```crystal
require "json"

class MyClass
  JSON.mapping(
    times: {type: Array(Time), converter: Time::EpochConverter}
  )
end

times = [] of Int64

times << Time.new.to_unix
times << Time.utc.to_unix

times.to_json

json = MyClass.from_json(times.to_json)
json.to_json
```

The specific error is:

> no overload matches 'Time::EpochConverter.to_json' with types Array(Time), JSON::Builder
> Overloads are:
>  - Time::EpochConverter.to_json(value : Time, json : JSON::Builder)
>  - Object#to_json(io : IO)
>  - Object#to_json()

asterite:
`Time::EpochConverter` is for converting a `String` into `Time` and the other way around, not from `Array` into `Time`.

This is an interesting problem.

I think one way we can achieve this is by having `JSON::ArrayConverter(Converter)` which basically encodes/decodes an array using another converter as a generic type argument. Then you would use it like this:

```crystal
require "json"

class MyClass
  JSON.mapping(
    times: {type: Array(Time), converter: JSON::ArrayConverter(Time::EpochConverter)}
  )
end

times = [] of Int64

times << Time.new.to_unix
times << Time.utc.to_unix

times.to_json

json = MyClass.from_json(times.to_json)
json.to_json
```

We should probably have `JSON::HashConverter(Converter)` too. And do the same with YAML.

What do you think?

I did a small test and it seems to work fine. The implementation I have so far is:

```crystal
module JSON::ArrayConverter(Converter)
  def self.from_json(pull : JSON::PullParser)
    ary = Array(typeof(Converter.from_json(pull))).new
    pull.read_array do
      ary << Converter.from_json(pull)
    end
    ary
  end

  def self.to_json(values : Array, builder : JSON::Builder)
    builder.array do
      values.each do |value|
        Converter.to_json(value, builder)
      end
    end
  end
end
```

HCLarsen:
I think it's a great solution, along with the HashConverter and adding both to YAML#mapping as well.

[...]

PR #8156, opened 2019-09-05
Allowing JSON.mapping & YAML.mapping converter attribute to be applied to Array and Hashes.

rodrigopinto:
# What is this PR for?

This PR is for making `JSON.mapping` & `YAML.mapping` support a converter to be applied to an Array or a Hash. 

👉 Closes https://github.com/crystal-lang/crystal/issues/7981

### DIFF
--- a/src/json/to_json.cr
+++ b/src/json/to_json.cr
@@ -161,6 +161,33 @@ struct Time
   end
 end
 
+# Converter to be used with `JSON.mapping`
+# to serialize an `Array(T)` instance with a custom converter
+# since the unix epoch. See `Time#to_unix`.
+#
+# ```
+# require "json"
+#
+# class Timestamp
+#   JSON.mapping({
+#     values: {type: Array(Time), converter: JSON::ArrayConverter(Time::EpochConverter)},
+#   })
+# end
+#
+# timestamp = Timestamp.from_json(%({"dates":[1459859781,1567628762]}))
+# timestamp.values  # => [2016-04-05 12:36:21 UTC, 2019-09-04 20:26:02 UTC]
+# timestamp.to_json # => %({"dates":[1459859781,1567628762]})
+# ```
+module JSON::ArrayConverter(Converter)
+  def self.to_json(values : Array, builder : JSON::Builder)
+    builder.array do
+      values.each do |value|
+        Converter.to_json(value, builder)
+      end
+    end
+  end
+end
+
 # Converter to be used with `JSON.mapping` and `YAML.mapping`
 # to serialize a `Time` instance as the number of seconds
 # since the unix epoch. See `Time#to_unix`.
@@ -232,3 +259,4 @@ module String::RawConverter
     json.raw(value)
   end
 end
+
--- a/src/json/to_json.cr
+++ b/src/json/to_json.cr
@@ -162,8 +162,7 @@ struct Time
 end
 
 # Converter to be used with `JSON.mapping`
-# to serialize an `Array(T)` instance with a custom converter
-# since the unix epoch. See `Time#to_unix`.
+# to serialize the `Array(T)` elements with the custom converter.
 #
 # ```
 # require "json"
@@ -189,8 +188,7 @@ module JSON::ArrayConverter(Converter)
 end
 
 # Converter to be used with `JSON.mapping`
-# to serialize a `Time` instance as the number of seconds
-# from a `Hash(K, V)` values.
+# to serialize the `Hash(K, V)` values elements with the custom converter.
 #
 # ```
 # require "json"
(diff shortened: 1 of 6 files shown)


src/json/to_json.cr
```diff
@@ -188,7 +188,24 @@ module JSON::ArrayConverter(Converter)
   end
 end
 
-module JSON::HashConverter(Converter)
+# Converter to be used with `JSON.mapping`
```
Sija: These descriptions mentioning `Time` are IMO very confusing. One could think that's the only use case, whereas IIUC that's supposed to be the example.

rodrigopinto: True. When writing it I had in mind only the `Time` converters.

RX14:
This looks really good! If you don't have the time to implement `YAML::HashConverter`, this can be merged now since it's a strict improvement.

rodrigopinto:
@RX14 Thanks for the review. I actually started it already, I hadn't enough time last week to push it forward. Hope to get it on ready for final review this week.

rodrigopinto:
@RX14 @straight-shoota I finished the implementation of the YAML converters, it is ready for another review.  I did some squash but a few "independent commits", let me know if more squashes are needed.

src/yaml/from_yaml.cr
```diff
@@ -290,6 +290,38 @@ module Time::EpochMillisConverter
...
+  end
+end
+
+module YAML::HashValueConverter(Converter)
+  def self.from_yaml(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
+    unless node.is_a?(YAML::Nodes::Mapping)
+      node.raise "Expected mapping, not #{node.class}"
+    end
+
+    hash = Hash(String, typeof(Converter.from_yaml(ctx, node))).new
```
Sija: (unlike JSON) YAML allows keys of different types, not just `String`.

RX14: JSON allows this too, on the crystal types side. This can be improved later with a different converter if you want to convert the keys too.

rodrigopinto: @Sija, that is true, we can improve it. I do believe that it could take another PR for that. Considering Hash also allows union types for the Key, the converter would need to handle the cast possibilities of different types, or would we make it strict?

straight-shoota: Both `Hash` converters simply need two generic arguments, one for converting the key, the other for converting the value. Just like `Hash` has two generic arguments as well.

straight-shoota: And I'd rather have this included here to avoid breaking the API later.

RX14: I mean, it's called `HashValueConverter`, not `HashConverter`.

rodrigopinto: So, I played a bit with a possible `YAML::HashConverter`, but wind up with some points.

Different from the basic types `Int32, String, Time, etc...`, the `Time::EpochConverter, Time::EpochMillisConverter` and others custom converters are modules which expose the method `self.from_yaml` that is used from the macro [mapping here](https://github.com/crystal-lang/crystal/blob/master/src/yaml/mapping.cr#L147).

So, to make a Hash works with union types, for example, `(Time::EpochConverter | String)` would be better to refactor these custom converters to have similar API of the basic types.

For example:
```crystal
struct Time::EpochConverter
  def from_yaml(ctx : YAML::ParseContext, node : YAML::Nodes::Node) : Time
    unless node.is_a?(YAML::Nodes::Scalar)
      node.raise "Expected scalar, not #{node.class}"
    end

    Time.unix(node.value.to_i)
  end
end
```

So, with that, we would avoid the situation commented on the snippet below.

```crystal
module YAML::HashConverter(K, V)
  def self.from_yaml(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
    unless node.is_a?(YAML::Nodes::Mapping)
      node.raise "Expected mapping, not #{node.class}"
    end

    hash = Hash(K, V).new

    YAML::Schema::Core.each(node) do |key, value|
      # here on the V.from_yaml(ctx, value)
      # It would work for if the type is Time::EpochConverter.from_yaml(ctx, value)
      # but in case it is a String, would be required something like String.new(ctx, value) instead
      hash[K.new(ctx, key)] = V.from_yaml(ctx, value)
    end

    hash
  end
end
```

Does it make sense @straight-shoota? Not sure if I missed something, so feel free to point out if I took the wrong path.

Concluding, I am happy to continue evolving this PR until we agree it is in a good shape, or either get it merged and open another PR with the changes + a `YAML::HashConverter` for keys and values.

RX14: `String.from_yaml` should have the same shape as a converter - this is how it works for `JSON`.

Please make an issue about this. Would you be okay leaving the `YAML` stuff off before that is refactored? What would others prefer?

RX14: Yes, please remove the hash converter and we can merge.

rodrigopinto: @RX14 done. I will start another one soon with tha yaml hash converter.

bcardiff:
This is going to be used so much 🚀 
Thanks @rodrigopinto !

PR #8159, opened 2019-09-06
Compiler: virtualize Proc types before instantiation

### LINKED ISSUE #6789: Compiler say: Proc(Int32, Socket) is not Proc(Int32, Socket) ?!
Following code is failed to compile with the message `'@func' of ... must be Proc(Int32, Socket), not Proc(Int32, Socket)`.

``` crystal
require "socket"
servers = 10.times.map do |i|
  Socket.new Socket::Family::UNIX, Socket::Type::STREAM, Socket::Protocol::IP
end
```

Error message:
```
Error in line 2: instantiating 'Int::TimesIterator(Int32)#map()'

in /usr/lib/crystal/iterator.cr:658: instantiating 'Iterator::Map(Int::TimesIterator(Int32), Int32, Socket).class#new(Int::TimesIterator(Int32), Proc(Int32, Socket))'

    Map(typeof(self), T, U).new(self, func)
                            ^~~

in /usr/lib/crystal/iterator.cr:665: instance variable '@func' of Iterator::Map(Int::TimesIterator(Int32), Int32, Socket) must be Proc(Int32, Socket), not Proc(Int32, Socket)

    def initialize(@iterator : I, @func : T -> U)
```

It happens on both `0.26.1` and `0.25.1`, see [play.crystal](https://play.crystal-lang.org/#/r/52uv).

asterite:
Fixes #6789

We must always use virtual types as generic arguments in types. It's a bit annoying that we have to remember to do so before calling `instantiate` or `proc_of`. The alternative would be to do that inside `instantiate` but for that we would need to create a separate array to hold the types and it would be a bit slower. This fixes the issue but eventually it would be nice to refactor it in that way if performance isn't impacted.

### DIFF
--- a/src/compiler/crystal/semantic/bindings.cr
+++ b/src/compiler/crystal/semantic/bindings.cr
@@ -415,8 +415,8 @@ module Crystal
       return unless self.def.args.all? &.type?
       return unless self.def.type?
 
-      types = self.def.args.map &.type
-      return_type = @force_nil ? self.def.type.program.nil : self.def.type
+      types = self.def.args.map &.type.virtual_type
+      return_type = @force_nil ? self.def.type.program.nil : self.def.type.virtual_type
 
       expected_return_type = @expected_return_type
       if expected_return_type && !expected_return_type.nil_type? && !return_type.implements?(expected_return_type)
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -958,4 +958,31 @@ describe "Semantic: proc" do
 
     ex.to_s.should contain "'bar' exists as a macro, but macros can't be used in proc pointers"
   end
+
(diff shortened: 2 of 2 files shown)


src/compiler/crystal/semantic/bindings.cr
```diff
@@ -415,8 +415,8 @@ module Crystal
       return unless self.def.args.all? &.type?
       return unless self.def.type?
 
-      types = self.def.args.map &.type
-      return_type = @force_nil ? self.def.type.program.nil : self.def.type
+      types = self.def.args.map &.type.virtual_type
```
bcardiff: the spec is not asserting virtual_type is used in arguments

asterite: Good catch. I'll update the spec tomorrow.

asterite: Done!

PR #8161, opened 2019-09-08
Specifies return types for some channel methods

### LINKED ISSUE #8160: MT compiler warnings
Some warnings are happening on nightly, most likely do the the MT changes.

```
 345 | def execute
           ^------
Warning: this method overrides Channel::SelectAction(S)#execute() which has an explicit return type of (Channel::NotReady | S).
Please add an explicit return type ((Channel::NotReady | T) or a subtype of it) to this method as well.
The above warning will become an error in a future Crystal version.
In /snap/crystal/96/share/crystal/src/channel.cr:381:9
 381 | def execute
           ^------
Warning: this method overrides Channel::SelectAction(S)#execute() which has an explicit return type of (Channel::NotReady | S).
Please add an explicit return type ((Channel::NotReady | Nil) or a subtype of it) to this method as well.
The above warning will become an error in a future Crystal version.
In /snap/crystal/96/share/crystal/src/channel.cr:349:9
 349 | def result
           ^-----
Warning: this method overrides Channel::SelectAction(S)#result() which has an explicit return type of S.
Please add an explicit return type (T or a subtype of it) to this method as well.
The above warning will become an error in a future Crystal version.
In /snap/crystal/96/share/crystal/src/channel.cr:386:9
 386 | def result
           ^-----
Warning: this method overrides Channel::SelectAction(S)#result() which has an explicit return type of S.
Please add an explicit return type (Nil or a subtype of it) to this method as well.
The above warning will become an error in a future Crystal version.
```

Blacksmoke16:
Fixes #8160

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -342,11 +342,11 @@ class Channel(T)
       @state = DeliveryState::None
     end
 
-    def execute
+    def execute : Channel::NotReady | T
       @channel.receive_internal { return NotReady }
     end
 
-    def result
+    def result : T
       @value
     end
 
@@ -378,12 +378,12 @@ class Channel(T)
     def initialize(@channel : Channel(T), @value : T)
     end
 
-    def execute
+    def execute : Nil
       @channel.send_internal(@value) { return NotReady }
       nil
     end
 
-    def result
+    def result : Nil
       nil
     end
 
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
I fixed the specs.  Should be good to review now.

PR #8162, opened 2019-09-08
enrich API for scanning strings 

c-cube:
- adds an optional "overlapping" argument
- add "each_match" for a lazy iterator, not enumerable.

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -1,6 +1,8 @@
 require "c/stdlib"
 require "c/string"
 
+require "regex/*"
+
 # A `String` represents an immutable sequence of UTF-8 characters.
 #
 # A `String` is typically created with a string literal, enclosing UTF-8 characters
@@ -3729,6 +3731,36 @@ class String
     end
 
     self
+  end
+
+  private class MatchIterator
+    include Iterator(Regex::MatchData)
+
+    def initialize(@s : String, @pattern : Regex, @overlapping : Bool)
+      @stop = false
+      @byte_offset = 0
+    end
+
+    def next
+      return stop if @stop
+
+      if match = @pattern.match_at_byte_index(@s, @byte_offset)
+        index = match.byte_begin(0)
+        shift_by = match[0].bytesize
+        shift_by = 1 if shift_by == 0 || @overlapping
+        @byte_offset = index + shift_by
+        return match
+      else
+        @stop = true
+        return stop
+      end
+    end
+  end
+
+  # Searches the string for instances of *pattern*,
+  # yielding an Iterator of `Regex::MatchData` for each match.
+  def each_match(pattern : Regex, overlapping = false)
+    MatchIterator.new(self, pattern, overlapping: overlapping)
   end
 
   # Searches the string for instances of *pattern*,
--- a/src/string.cr
+++ b/src/string.cr
@@ -3716,7 +3716,7 @@ class String
 
   # Searches the string for instances of *pattern*,
   # yielding a `Regex::MatchData` for each match.
-  def scan(pattern : Regex)
+  def scan(pattern : Regex, overlapping = false)
     byte_offset = 0
 
     while match = pattern.match_at_byte_index(self, byte_offset)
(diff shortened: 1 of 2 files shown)


src/string.cr
```diff
@@ -1,6 +1,8 @@
 require "c/stdlib"
 require "c/string"
 
+require "regex/*"
```
wontruefree: Is this added because of the order in the prelude file?

c-cube: yes, without that referring to `Regex::MatchData` would fail :/

src/string.cr
```diff
@@ -3716,48 +3718,78 @@ class String
...
+      else
+        @stop = true
+        return stop
+      end
+    end
+  end
+
+  # Searches the string for instances of *pattern*,
+  # yielding an Iterator of `Regex::MatchData` for each match.
+  def each_match(pattern : Regex, overlapping = false)
```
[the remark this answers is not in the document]
c-cube: StringScanner looks a lot more annoying to use. Isn't symmetry between `each(&block)` and `each : Iterator` common throughout the stdlib? To scan a large string lazily seems pretty useful (if you're only going to consume a small set of the `MatchData`, for example), and avoids building a large array upfront.

src/string.cr
```diff
@@ -3716,48 +3718,78 @@ class String
 
   # Searches the string for instances of *pattern*,
   # yielding a `Regex::MatchData` for each match.
-  def scan(pattern : Regex)
+  def scan(pattern : Regex, overlapping = false)
```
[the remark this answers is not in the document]
c-cube: It's the kind of feature that you rarely need (which is why it's false by default) but can come in handy. For example if you analyze logs with a multi-line regex to detect multiple line patterns, these can typically overlap.

src/string.cr
```diff
@@ -3716,48 +3718,78 @@ class String
...
     byte_offset = 0
 
     while match = pattern.match_at_byte_index(self, byte_offset)
       index = match.byte_begin(0)
       $~ = match
       yield match
       match_bytesize = match[0].bytesize
-      match_bytesize += 1 if match_bytesize == 0
+      match_bytesize = 1 if match_bytesize == 0 || overlapping
```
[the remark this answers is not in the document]
c-cube: Interesting trick. Is it as efficient? Seems less intuitive to me.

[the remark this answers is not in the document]
Sija: @c-cube it ain't trick but documented functionality.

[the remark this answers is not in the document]
c-cube: Ok, I can remove that part then.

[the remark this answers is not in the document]
Sija: It's still a regex though, which in general is slow(er), did you benchmarked both methods?

src/string.cr
wontruefree: I think This is should be added at the prelude level.
Since string is required in prelude we might as well make it explicit at that level.

straight-shoota: I don't think this can be accepted. `string` is a core type. It must not depend on `regex`.

asterite: Regex is also a core type. It has specialized syntax for it. There's `$1`, `$~`, etc.

wontruefree: It looks like `String` at master already refers to `Regex::MatchData` [here](https://github.com/crystal-lang/crystal/pull/8162/files#diff-87ebb896824e7dc3bd8d5b29293bcf00L3737).  Is there a way why this is different and needs to require `regex/*`

c-cube: I don't know why, but without this require the code doesn't compile.

wontruefree: What is the error?

c-cube:
```
./bin/crystal spec spec/std/string_spec.cr --error-trace
error in line 1
Error: while requiring "prelude"


In src/prelude.cr:29:1

 29 | require "string"
      ^
Error: while requiring "string"


In src/string.cr:3735:22

 3735 | include Iterator(Regex::MatchData)
                         ^---------------
Error: undefined constant Regex::MatchData
```

src/string.cr
```diff
@@ -3731,6 +3733,36 @@ class String
     self
   end
 
+  private class MatchIterator
```
c-cube: should this be a `struct`, btw?

asterite: Probably not because it's mutable.

j8r: A struct can be mutable, but only if not passed around. Otherwise the changes won't be propagated.

PR #8165, opened 2019-09-09
Add `Int#bits` for accessing bit ranges

stakach:
inspired by https://twitter.com/iximeow/status/1170530444112097282?s=21

### DIFF
--- a/src/int.cr
+++ b/src/int.cr
@@ -333,30 +333,17 @@ struct Int
     self >> bit & 1
   end
 
-  # Returns true if this number's *bit*th bit is 1, starting with the least-significant.
-  #
-  # ```
-  # 11[0] # => true
-  # 11[1] # => true
-  # 11[2] # => false
-  # 11[3] # => true
-  # 11[4] # => false
-  # ```
-  def [](index : Int)
-    bit(index) > 0
-  end
-
   # Returns the requested range of bits
   #
   # ```
-  # 0b10011[0..1] # => 0b11
-  # 0b10011[0..2] # => 0b11
-  # 0b10011[0..3] # => 0b11
-  # 0b10011[0..4] # => 0b10011
-  # 0b10011[0..5] # => 0b10011
-  # 0b10011[1..4] # => 0b1001
+  # 0b10011.bits(0..1) # => 0b11
+  # 0b10011.bits(0..2) # => 0b11
+  # 0b10011.bits(0..3) # => 0b11
+  # 0b10011.bits(0..4) # => 0b10011
+  # 0b10011.bits(0..5) # => 0b10011
+  # 0b10011.bits(1..4) # => 0b1001
   # ```
-  def [](range : Range)
+  def bits(range : Range)
     start_index = range.begin
     if start_index
       raise IndexError.new("start index (#{start_index}) must be positive") if start_index < 0
--- a/src/int.cr
+++ b/src/int.cr
@@ -363,7 +363,7 @@ struct Int
 
     # Generates a mask `count` bits long maintaining the correct type
     count = end_index - start_index
-    mask = (self.class.new(1) << count) - 1
+    mask = (self.class.new(1) << count) &- 1
 
     if self < 0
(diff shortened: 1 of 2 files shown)


stakach:
Weird... I don't see the overflow running on my Mac.
Any hints on how I should ignore that?

asterite:
In the past we decided to explicitly not add this. Imagine passing an integer, or an array size, or whatever number to a method that you expect is something that indexes an array. Then your code compiles and you have no idea why. It can be really confusing and I experienced that in Ruby. We have the bit method for this.

stakach:
Good point!
What about a `#bits` function?

asterite:
> What about a #bits function?

Maybe :-)

RX14:
Specs fail cause of overflow.

stakach:
Yeah, any pointers on how to ignore that?

src/int.cr
```diff
@@ -333,6 +333,51 @@ struct Int
...
+  #
+  # ```
+  # 0b10011.bits(0..1) # => 0b11
+  # 0b10011.bits(0..2) # => 0b11
+  # 0b10011.bits(0..3) # => 0b11
+  # 0b10011.bits(0..4) # => 0b10011
+  # 0b10011.bits(0..5) # => 0b10011
+  # 0b10011.bits(1..4) # => 0b1001
+  # ```
+  def bits(range : Range)
```
j8r: A nillable `bits?` can be considered too.
If done, the return type can be defined to clarify the difference.

bcardiff:
@stakach the overflow is in the new code: https://github.com/crystal-lang/crystal/pull/8165/files#diff-cf769fe2e361e14888004d0eae8368fdR366

Try building with `-Dpreview_overflow` and it should raise on your side aswell.

If you need wrapping around operations for `+` `-` `*`, use `&+`, `&-`, `&*`.

src/int.cr
stakach: I'm not sure I understand, when would / should one expect `bits?` return nil?

RX14: `bits?` would return `nil` instead of an `IndexError`. Not sure it's required though... I'd leave it for now.

asterite: I think it's wrong to raise `IndexError`. Bits past the bytesize of an int are turned off, imagining like an infinite series of zeros to represent the number. The `bit` method never raises.

Then there's the question about negative indices... in Ruby for some reason it gives you 0 instead of raising. In Crystal it works like that, maybe by chance. Maybe it's not bad to raise on negative indices. But I would not raise on positive indices.

RX14: That's an interesting point.

I'd definitely raise on negative ones.

stakach: @asterite it currently doesn't raise on positive indices - it's also tested in the spec
`it { 0b10011_u8.bits(0..16).should eq(0b10011_u8) }` note that we are grabbing `0..16` of `8`

It does raise on negative as it doesn't make sense to me for that to work and the logic to implement would be computationally expensive

asterite: Ah, you are right. Thanks!

asterite:
The code looks code, the method is fancy and we could merge it. I just like to know what use cases this has.

stakach:
It's useful for dealing with binary data structures.
https://en.wikipedia.org/wiki/IPv4#/media/File:Ipv4_address.svg

IPv4 is probably not a great example, but I've had to do things like extracts the components of a custom floating point number format. So many weird and wonderful protocols out there where this is more descriptive and less error prone than bit shifting and masking.

refi64:
Would it be worth changing the PR title prior to merging to avoid confusion
as to what was actually added?

icy-arctic-fox:
Ruby 2.7 now has the original syntax used in this PR (without the `#bits` method). https://blog.saeloun.com/2019/09/24/ruby-2-7-integer-with-range.html

stakach:
I am so glad to be programming crystal primarily now:
https://github.com/ruby/ruby/commit/6bedbf462544a7917fdc8d8c44276079a6e156cf

PR #8166, opened 2019-09-09
Refactor Array with Crystal::Buffer

bcardiff:
I extracted some of the refactors to make Array thread-safe. I think they are worth in its own and can be used in other structures: `Deque(T)`, `String::Builder`, `IO::Memory` since they use pointer-based buffers like Array.

This is clash a bit with #8036 since it affects the internal representation of the Array.

I wanted to remove as much as possible the usage of `to_unsafe` method in Array implementation. Those were overused to get a pointer to the data of the buffer but should be framed for C interop when possible. This aspect is one of the conflicts with #8036.

The `Crystal::Buffer` is allocated with the capacity and then with the buffer in an inlined way. The buffer can't be resized inplace have convenient methods to move data, ensure capacity, use power of two sizes, etc. as used in other containers.

Some future benefits of using `Crystal::Buffer`:

* Remove duplication logic of containers `Deque(T)`, `String::Builder`, `IO::Memory` 
* Probably allow allocation of atomic buffers
* It encapsulates the direct usage of Pointer

### DIFF
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -460,11 +460,11 @@ describe "Array" do
     it "concats enumerable to empty array (#2047)" do
       a = [] of Int32
       a.concat(1..1)
-      a.@capacity.should eq(3)
+      a.@buffer.capacity.should eq(3)
 
       a = [] of Int32
       a.concat(1..4)
-      a.@capacity.should eq(6)
+      a.@buffer.capacity.should eq(6)
     end
 
     it "concats a union of arrays" do
--- a/src/array.cr
+++ b/src/array.cr
@@ -1,3 +1,5 @@
+require "crystal/buffer"
+
 # An `Array` is an ordered, integer-indexed collection of objects of type T.
 #
 # Array indexing starts at 0. A negative index is assumed to be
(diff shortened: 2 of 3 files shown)


RX14:
I don't quite get the point of this refactor... It seems to be like a slice without bounds checking but a worse API. The `Buffer` class has almost zero logic inside itself. 90% of the diff is just adding another `.data` and very little logic is moved out of `Array`.

wontruefree:
There are a few times where I want to rezise an array before doing a bunch of inserts.  With this patch can I resize the array before a some insert logic.

```crystal
x = [1, 2 3]
x.@buffer = @buffer.realloc(100)
```

I think this is an edge case and possibly a premature optimization.  It is intended to limit memory allocations and copying.

asterite:
@wontruefree I think for that we should introduce a public API to pre-allocate size. Something like `ensure_capacity(...)` or `reserve(...)`, etc.

wontruefree:
@asterite that would be awesome.  I have been using the capacity hints to try and increase performance.

asterite:
But please nobody send a PR for that without first opening an issue to discuss the API. Thank you!

straight-shoota:
Closing stale PR. The array implementation has evolved since this generalization was proposed, so such a generalization would need to be re-designed. I'm not even sure that even makes much sense. In any case we need to discuss where we even want to go with this. I understand the main goal was simplification for implementing thread-safe data structures. This discussion is still pretty open though (#9548).

PR #8168, opened 2019-09-09
Crystal::ThreadLocalValue is now thread safe

waj:
Apparently we forgot to make this class thread safe 🤦‍♂ 

This affected for example when the same IO is used in different threads.

### DIFF
--- a/src/crystal/thread_local_value.cr
+++ b/src/crystal/thread_local_value.cr
@@ -1,27 +1,38 @@
 # :nodoc:
 struct Crystal::ThreadLocalValue(T)
   @values = Hash(Thread, T).new
+  @mutex = Crystal::SpinLock.new
 
   def get(&block : -> T)
     th = Thread.current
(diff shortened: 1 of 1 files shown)


src/crystal/thread_local_value.cr
```diff
@@ -1,27 +1,38 @@
...
   @values = Hash(Thread, T).new
+  @mutex = Crystal::SpinLock.new
 
   def get(&block : -> T)
     th = Thread.current
-    @values.fetch(th) do
-      @values[th] = yield
+    @mutex.sync do
+      @values.fetch(th) do
+        @values[th] = yield
```
RX14: Using a spinlock here seems really dangerous, since it'd be really easy to accidentally put an expensive operation in the block. *especially* so for the `#each` below.

I'm not sure there's a good solution, but even allocating memory inside a spinlock seems too expensive for that kind of lock, to me.

Issue #8153, opened 2019-09-05
I am not getting overflow errors


wontruefree:
I am not getting overflow errors where I am expecting an error.  The crystal compiler I am using is being compiled with the preview on.

```bash
CRYSTAL_CONFIG_PATH="/Users/jack/src/crystal/src" CRYSTAL_CONFIG_LIBRARY_PATH="/Users/jack/.anyenv/envs/crenv/versions/0.30.0/embedded/lib" CRYSTAL_CONFIG_BUILD_COMMIT="b6169b5af" ./bin/crystal build -D preview_overflow -D compiler_rt --link-flags="-L/usr/local/opt/llvm/lib"  -o .build/crystal src/compiler/crystal.cr -D without_openssl -D without_zlib
```

I would expect this to throw and error and it does not.
```crystal
puts "#{Int64::MAX} + #{1_u64} = #{Int64::MAX + 1_u64}"
puts "#{Int64::MAX} &+ #{1_u64} = #{Int64::MAX &+ 1_u64}"
```

```bash
9223372036854775807 + 1 = -9223372036854775808
9223372036854775807 &+ 1 = -9223372036854775808
```

[here](https://github.com/crystal-lang/crystal/blob/master/spec/compiler/codegen/arithmetics_spec.cr#L87) is a test that I think should be checking this.

Also [here](https://play.crystal-lang.org/#/r/7i50) is a playground that should throw and error.

asterite:
> The crystal compiler I am using is being compiled with the preview on.

If you do that, if an overflow happens during compilation you'll get an error.

You need to compile your program with `-D preview_overflow`.

Just run `crystal foo.cr -Dpreview_overflow` assuming `foo.cr` has the code you put above and you'll see the error.

bcardiff:
You need to pass `-D preview_overflow` when compiling the program you want to test. The flag in the compiler itself does not affect the program you want to test.

wontruefree:
Thanks that works for everything now.

Issue #7798, opened 2019-05-19
compiler_rt testing


wontruefree:
I have been trying to test the compiler_rt flag and am a bit confused.  It seems like the file is being loaded and the functions are not being used.

I have put a `raise "testing"` inside the `__mulodi4` rt function.
https://github.com/crystal-lang/crystal/blob/master/src/crystal/compiler_rt.cr#L4

After building crystal and running `.build/crystal eval "puts 1_i64 * 1_i64"` I would expect crystal to raise "testing" and it does not.

I have checked the build output and the "compiler_rt" flag is being passed.

```bash
$ make
Using /usr/local/opt/llvm@6/bin/llvm-config [version=6.0.1]
c++ -c  -o src/llvm/ext/llvm_ext.o src/llvm/ext/llvm_ext.cc -I/usr/local/Cellar/llvm@6/6.0.1_1/include  -stdlib=libc++ -fPIC -fvisibility-inlines-hidden -Werror=date-time -Werror=unguarded-availability-new -std=c++11 -Wall -W -Wno-unused-parameter -Wwrite-strings -Wcast-qual -Wmissing-field-initializers -pedantic -Wno-long-long -Wcovered-switch-default -Wnon-virtual-dtor -Wdelete-non-virtual-dtor -Wstring-conversion -DNDEBUG   -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
cc -fPIC    -c -o src/ext/sigfault.o src/ext/sigfault.c
ar -rcs src/ext/libcrystal.a src/ext/sigfault.o
CRYSTAL_CONFIG_PATH="/Users/jack/source_code/open_source/crystal/src" CRYSTAL_CONFIG_LIBRARY_PATH="/usr/local/Cellar/crystal/0.28.0/embedded/lib" CRYSTAL_CONFIG_BUILD_COMMIT="8e3075027" ./bin/crystal build -D preview_overflow -D compiler_rt  -o .build/crystal src/compiler/crystal.cr -D without_openssl -D without_zlib
```

also if I put a `raise` in the file itself after the [compile check](https://github.com/crystal-lang/crystal/blob/master/src/crystal/compiler_rt.cr#L1) and it seems like the file is being loaded. 

Am I missing something?

asterite:
You need to use bin/crystal

asterite:
instead of .build/crystal

wontruefree:
[...]

Issue #7514, opened 2019-03-06
Embed compiler-rt when building programs


bcardiff:
To support overflow and 128 ints property we need to either port the whole library or ship it with the compiler and link against them during compile time (as it's done with the boehm-gc).

I think the plan would be:

- [ ] update  omnibus distribution-scripts llvm formula to 6.0.1. 
  * This will allow dropping llvm < 4 support in the codebase. 
    * Currently the OSX pkg link against llvm 3.9.1. 
  * These binaries could also be used in Linux packages if wanted.
    * Currently we use the APK packages for llvm-4.0
- [ ] update omnibus distribution-scripts to build llvm with compiler-rt support. 
- [ ] upload llvm + compiler-rt binaries for OSX & Linux in S3 
  * as it is done for llvm 3.9.1 for OSX
- [ ] do a release to have the compiler-rt libraries ready to be embedded
- [ ] make compiler link the embedded compiler-rt.

When building compiler-rt there are a bunch of options and they are well described in https://embeddedartistry.com/blog/2017/2/20/compilerrt
>  * hard vs soft: this is floating point. Is your platform configured to support hard or soft floating point operations?
>  * static vs pic: is the code compiled as a static library, or are you compiling with position-independent-code? (PIC)
>  * i386 vs armv7x: this will be dependent upon your platform's processor. You need to pick the instruction set to match.
>  * The last portion of the name is the library format.

Some doubts I have are:
* Should we have a compiler option to choose specifically with which compiler-rt link against?
* Should we ship all versions of compiler-rt? of all tier-1 platforms?
* Would be safe to embed compiler-rt 6.0.1 version in Linux that use llvm 4.0? Or should we build compiler-rt 4.0 also?, or update Linux to 6.0.1?

bcardiff:
The current direction is to port compiler-rt to crystal code as done in #8170 

Meanwhile, if something is compiler-rt would need to be compiled and linked. More info at https://embeddedartistry.com/blog/2017/2/20/compilerrt in case the environment does not ship compiler-rt directly.

RX14:
I'm concerned that a crystal compiler-rt will end up conflicting with other libraries which depend on compiler-rt.

bcardiff:
@RX14 
>  If the user wants to link an official compiler-rt, the `skip_crystal_compiler_rt` flag can be used. But I expect the port will be easier to use.
https://github.com/crystal-lang/crystal/pull/8170#issue-315959399

[...]

PR #8170, opened 2019-09-10
Enable overflow by default

bcardiff:
This PR makes the overflow behavior active by default.

A temporal flag `disable_overflow` is added in case this change cause stopper issues. Expect to be removed in the near future.

Instead of linking to compiler-rt, the existing minimal port will be used. The specs of `mulodi4` were ported also.

The `mulodi4` implementation was moved to a `compiler_rt` directory. More compiler-rt functions might be ported if we aim for Int128 bit support across all targets. But that is a different story.

The ported `mulodi4` is not always used (it depends on the architecture) but is always emitted. If the user wants to link an official compiler-rt, the `skip_crystal_compiler_rt` flag can be used. But I expect the port will be easier to use.

cc: @wontruefree
Ref: #7514, #8153, #7798


spec/compiler/codegen/and_spec.cr
```diff
@@ -18,11 +18,11 @@ describe "Code gen: and" do
   end
 
   it "codegens and with bool and int 1" do
-    run("struct Bool; def to_i; 0; end; end; (false && 2).to_i").to_i.should eq(0)
+    run("struct Bool; def to_i!; 0; end; end; (false && 2).to_i!").to_i.should eq(0)
```
asterite: Is this rename really needed? It's just a spec, I don't think it breaks anything if we use `to_i`, right?

asterite: Oh, nevermind, it does because we later invoke it on `2`.

bcardiff: Yeah, ideally we should be able to use exceptions with the JIT runner that does not include prelude. But that is a pending thing to be done. But the specs test more the dispatch and typing than the actual behavior.

wontruefree:
Thanks for the CC.  This broke my build and it was nice to find the reason why.

PR #8173, opened 2019-09-10
Remove the event loop fiber

waj:
Since the commit https://github.com/crystal-lang/crystal/commit/284fb1e35abce2ecbebe41b3d6a78bf757145b9b there is actually no need to have a dedicated fiber to run the event loop. This is because now the event loop callbacks do not resume fibers directly and also it's not executing in endless non-returning mode.

With these changes the event loop is executed by any fiber that calls for a `reschedule` and there are no more fibers in the queue. This removes an extra context switch in those cases and provide a small but noticeable performance improvement for IO intensive apps.

### DIFF
--- a/src/crystal/event_loop.cr
+++ b/src/crystal/event_loop.cr
@@ -3,16 +3,6 @@ require "./event"
 class Thread
   # :nodoc:
   getter(event_base) { Crystal::Event::Base.new }
-
-  # :nodoc:
-  getter(loop_fiber) do
-    Fiber.new(name: "Event Loop") do
-      loop do
-        self.event_base.run_once
-        Crystal::Scheduler.reschedule
-      end
-    end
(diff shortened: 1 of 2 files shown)


wontruefree:
Does this have a performance bump?
Have you seen this thread https://forum.crystal-lang.org/t/multithreaded-crystal-initial-thoughts/1089

waj:
@wontruefree yes, I've seen that thread and yes, it should have a performance improvement although don't expect it to be really big. It just removes an extra context switch made every time there are no more runnable fibers and the runtime holds on the event loop for something to happen.

sdogruyol:
For the curious and @wontruefree this might be a pointer of the performance boost :+1:  https://twitter.com/sdogruyol/status/1171814376212508672

Issue #8127, opened 2019-08-29
Add ids to headings in generated docs


Blacksmoke16:
This feature is intended to be specific to the generated docs, but would require updates to the markdown module.  However, https://github.com/crystal-lang/crystal/pull/8115 makes that a non-issue.

I propose to:
* Add ids to the headings in the docs. 
  * Would allow linking directly to a section using the ids as targets for `#some-heading`.
* Have some easier way to link to a specific section.
  * Something like how Github does it, showing an icon you can click next to a heading that scrolls you down to that section.
  * Having the link copy the link to your clipboard would also be a nice to have.

Having this would make documenting shards directly within the API docs much much more user friendly and allow for more specific links to other sections within the docs themselves.

absolutejam:
It'd also be awesome to provide functionality like Elixir's Hexdocs, where you can provide the full name of a symbol (eg. `Foo.Bar.baz/2` in Elixir) in documentation comments and it will automatically link to it.

asterite:
> It'd also be awesome to provide functionality like Elixir's Hexdocs, where you can provide the full name of a symbol (eg. Foo.Bar.baz/2 in Elixir) in documentation comments and it will automatically link to it.

You can already do that. For example `Foo.bar`, `Foo#bar`, `#baz`, `#qux(x, y)`, etc.

straight-shoota:
With overloads based on argument types it can quickly get pretty verbose. But I guess there's no way around that.

absolutejam:
> > It'd also be awesome to provide functionality like Elixir's Hexdocs, where you can provide the full name of a symbol (eg. Foo.Bar.baz/2 in Elixir) in documentation comments and it will automatically link to it.
> 
> You can already do that. For example `Foo.bar`, `Foo#bar`, `#baz`, `#qux(x, y)`, etc.

I could have sworn I tried this before 🤦‍♂  This is awesome, thanks.

PR #8174, opened 2019-09-11
Use markd for internal markdown implementation of `crystal docs`

### LINKED ISSUE #4613: Improve Markdown implementation - Or: Should we separate compiler tools?

### LINKED ISSUE #7009: Numbered list not rendered correctly in docs when using code block

### LINKED ISSUE #7463: Markdown escapes HTML

asterite:
Fixes #4613
Fixes #7463
Fixes #7009

It's finally been decided that Markdown shouldn't be in the standard library. Eventually `crystal docs` (and others) will be moved to separate repos and their binaries will be provided in Crystal distributions. That means `crystal docs` could use [markd](https://github.com/icyleaf/markd) just fine.

In the meantime, I think it's no problem at all, if @icyleaf agrees, to make the internal Markdown implementation use markd instead of our buggy and incomplete one.

Why?

- It doesn't tie the std API to anything, we don't expose Markdown. People can still use it if they require the right path but it's their problem if they do that (we don't encourage that, and one can always use `markd` as a shard which is the same)
- It makes `crystal docs` feel like a mature tool where Markdown behaves correctly, **right now**. We don't have to wait to extract `crystal docs` to another repository. And when we'll do that we can simply delete the markdown-related code and use `markd` as a shard.

I mean, imagine 0.31.0 having experimental MT, Markdown that works well in crystal docs and many std optimizations. Doesn't that sound great?

/cc @waj @bcardiff


Blacksmoke16:
Do you know if this inherently fixes #8127 as well?

asterite:
@Blacksmoke16 No, it doesn't. I guess it should be easy to do, but I'd do it after merging this PR (if it gets merged).

j8r:
@asterite markd is licensed under MIT, copyrighted by @icyleaf.
The license **needs** to be kept, along with the copyright.
For instance, Ruby has [LEGAL](https://github.com/ruby/ruby/blob/master/LEGAL).

> It's finally been decided that Markdown shouldn't be in the standard library

?? We got no information about this in https://github.com/crystal-lang/crystal/issues/4613, nothing has been officially decided then.

Finally, this change will fix bugs for sure, and also new ones will be found.

What will you do if bugs are found? Fix them only in the std without contributing back? A bit silly.
Fix them both in std + markd? Twice the work to do.

More work to do, compared to the current "markdown broken, won't fix". Broken, but kind of stable.

asterite:
> @asterite markd is licensed under MIT, copyrighted by @icyleaf.
> The license needs to be kept, along with the copyright.

Sounds good, I'll do it if we decide that this PR should be merged.

> It's finally been decided that Markdown shouldn't be in the standard library
> ?? We got no information about this in #4613, nothing has been officially decided then.

Brian and Juan told me this. Not everything happens in GitHub or the forum. They are the main core team members.

> Finally, this change will fix bugs for sure, and also new ones will be found.
> What will you do if bugs are found? Fix them only in the std without contributing back? A bit silly.
> Fix them both in std + markd? Twice the work to do.

Right now we are not fixing any bugs related to Markdown. We don't have to fix bugs in the new Markdown using markd. The difference is that markd has fewer (maybe none?) bugs, and it passes the entire Markdown spec suite.

oprypin:
Why exactly doesn't this keep the Git history?

In any case, please let me take a look at this first, don't merge yet.

ibraheemdev:
If this pull request was updated to preserving fit history, would it be accepted?

asterite:
I don't think so. The other core team members didn't like pulling an external markdown library into the project, even if it's just for the compiler.

j8r:
The only possible option remaining would be to move the docs generation to its own project, in order to use external dependencies. Not going to happen anytime soon.

straight-shoota:
I've sent a PR to upstream some of the patches to markd: https://github.com/icyleaf/markd/pull/36

Issue #8111, opened 2019-08-23
[RFC] Simplify integers and floats


asterite:
Right now Crystal has a variety of integer and float types:
- Integers: `Int8`, `Int16`, `Int32`, `Int64`, `UInt8`, `UInt16`, `UInt32`, `UInt64`
- Floats: `Float32`, `Float64`

The default integer type when you don't use a suffix is `Int32` and the default float type is `Float64`.

This kind of works but I imagine something better.

## Int32 and Float64

Given that Int32 and Float64 are the default types it feels a bit redundant to type those 32 and 64 numbers all the time.

So here's an initial idea: what if we name those types `Int` and `Float`? We would of course need to rename the existing base types `Int` and `Float` but that's not a problem, we can maybe call them `IntBase` and `FloatBase` or `Integral` and `Floating`, it doesn't matter much because those names won't be used a lot.

Then talking about ints and floats is so much simpler: just use `Int` and `Float` everywhere. In the case where you do need a specific limit, which is rare and usually only useful in low-level code such as interfacing with C or writing binary protocols, you can still use the names `Int32`, `Int64`, `Float32` or whatever you need.

## What to alias to

Now, we could make `Int` be an alias of `Int32` and `Float` an alias of `Float64`, but maybe it's better if we make `Int` depend on the architecture. That means `Int` would be equivalent to `Int64` in 64 bits architectures.

This is also how Go works. [They recommend using `int` everywhere unless you have good reasons to use a specific size](https://tour.golang.org/basics/11). It's probably the case that using `Int64` by default instead of `Int32` works equally fine (maybe even better because the range is bigger so overflow is less possible) without a real performance degradation.

Another nice thing is that if eventually 128 bit architectures appear all programs will automatically start using this bigger range (if we want to) without needing to change any code.

## To alias or not

Now, we _could_ make `Int` be an alias of the respective underlying type, but I don't think that's a good idea. The reason is that if you have a program that does:

```crystal
x : Int = 1
y : Int32 = 2
x = y
```

that would compile in 32 bits but would stop compiling in 64 bits. Ideally we'd like our programs to always compile regardless of the architecture.

So, we could make `Int` and `Float` be **different** types. To assign `Int32` or `Int64` to them you would need to call `to_i` first. Then programs on 32 and 64 bits will go through that explicit conversion process.

[...]

PR #8175, opened 2019-09-12
refactor: add some type annotations to the stdlib

c-cube:
improves the documentation, imho. I'm not sure the ones for `index` or `rindex` are good, but `Int?` was rejected by the compiler.


src/string.cr
```diff
@@ -903,23 +905,23 @@ class String
...
     char_at(index).ord
   end
 
-  def byte_at(index)
+  def byte_at(index) : Int
     byte_at(index) { raise IndexError.new }
   end
 
-  def byte_at?(index)
+  def byte_at?(index) : (UInt8 | Int8 | UInt16 | Int16 | Int32 | Int64 | UInt32 | UInt64)?
```
yxhuvud: I'm quite skeptical about these annotations, but why does this return such a strange type?

src/string.cr
```diff
@@ -3716,48 +3719,78 @@ class String
...
       match_bytesize = match[0].bytesize
-      match_bytesize += 1 if match_bytesize == 0
+      match_bytesize = 1 if match_bytesize == 0 || overlapping
       byte_offset = index + match_bytesize
     end
 
     self
   end
 
+  private class MatchIterator
```
yxhuvud: Why is this class introduced?

src/indexable.cr
```diff
@@ -479,7 +479,7 @@ module Indexable(T)
   # ```
   # [1, 2, 3, 1, 2, 3].index(2, offset: 2) # => 4
   # ```
-  def index(object, offset : Int = 0)
+  def index(object, offset : Int = 0) : Int32 | UInt32 | UInt64 | Nil
```
wontruefree: no `Int64`?

src/indexable.cr
```diff
@@ -554,7 +554,7 @@ module Indexable(T)
   # [1, 2, 3, 2, 3].rindex(2)            # => 3
   # [1, 2, 3, 2, 3].rindex(2, offset: 2) # => 1
   # ```
-  def rindex(value, offset = size - 1)
+  def rindex(value, offset = size - 1) : Int32 | UInt32 | UInt64 | Nil
```
wontruefree: same as above

src/string.cr
```diff
@@ -1,6 +1,8 @@
 require "c/stdlib"
 require "c/string"
 
+require "regex/*"
```
wontruefree: Why is this included?
Should this be added explicitly in the prelude?

src/indexable.cr
asterite: This should be `Int32 | Nil` or `Int32?`

c-cube: does crystal upcast implicitly? I think it's used at least once in a context expecting each of these types. Will check again.

straight-shoota: The return type is the same as `offset` (or `Nil`), because it's derived from that value. So it should probably just be specified as `Int?`.

asterite: The return type is just checked but not used. Well, it's used in one specific case where a type can't be determined because of recursion. So it's better to put the more strict type, which is `Int32?`.

c-cube: @straight-shoota that was my first attempt, but the compiler rejects `Int?` (because it's unsized I imagine)

yxhuvud: I'd take a step back and first ask: Should it be mandatory for `Indexable(T)` to have a Int32 size? Or should it be up to the implementing T to decide whatever type its size has? If the latter, I'd vote for removing the return type signature here altogether.

c-cube: I'm surprised there isn't a `Size` integer type, which Indexable could use. Or the index type could be a parameter of `Indexable` itself.

asterite: There is a discussion about it: https://github.com/crystal-lang/crystal/issues/8111

But for now Int32 is the default integer type that should be used everywhere

src/string.cr
c-cube: if I remove this the file doesn't compile, as somehow `Regex::MatchData` isn't resolved in the `private class MatchIterator`.

src/indexable.cr
wontruefree: I am not sure this is resolved I think this should be `Int32?`. I would not care as much if this was not locking down public APIs and could cause breaking changes in the future

c-cube: putting `Int32?` makes some specs fail because they use `index` and expect a `UInt64?` or `UInt32?` (which are otherwise properly inferred).

As it seems there's no clean way to say something like `def index(…) : U? forall U < Int`, I removed the annotation.

asterite:
Sorry, I vote no against this PR. One of the nicest things about Crystal, what makes it unique, is that you can omit types if you want to. This PR makes things more strict, possibly leading to backwards incompatibility, and in other cases it makes method signatures long and verbose, similar to those of Java and Rust in some cases (like Hash fetch)

c-cube:
Ah, too bad :slightly_frowning_face: 
To me types also act as additional documentation.

straight-shoota:
@asterite  I agree to your argument, but only where it applies. Most of the changes neither are overly verbose or break anything. They just document behaviour. And that's a good thing. Even as an experienced Crystal developer, you might not always recall the exact return types of each method. But sometimes it's important to know them (for example for declaring ivars, or simply debugging).

asterite:
Let's keep them then, but only when we are sure of the types. For example I would remove the return of Indexable#index

straight-shoota (dismissed):

src/string.cr
```diff
@@ -903,23 +905,23 @@ class String
...
     byte_at(index) { raise IndexError.new }
   end
 
-  def byte_at?(index)
+  def byte_at?(index) : UInt8?
     byte_at(index) { nil }
   end
 
-  def byte_at(index)
+  def byte_at(index, &block : -> U) : UInt8 | U forall U
```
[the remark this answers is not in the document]
c-cube: this doesn't work, because of an overload of `byte_at`:

```crystal

  def byte_at?(index) : UInt8?
    byte_at(index) { nil }
  end
```
which returns `UInt8?` (so the `U` here is `Nil`).

src/string.cr
```diff
@@ -853,11 +855,11 @@ class String
     char_at(index) { yield }
   end
 
-  def char_at(index : Int)
+  def char_at(index : Int) : Char
     char_at(index) { raise IndexError.new }
   end
 
-  def char_at(index : Int)
+  def char_at(index : Int, &block : -> U) : Char | U forall U
```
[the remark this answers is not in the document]
c-cube: same as below, `&block : -> Char` is too restrictive.

src/string.cr
```diff
@@ -1980,11 +1982,11 @@ class String
   # This returns `true` if this string has `'\\'` in it. It might not be a back reference,
   # but `'\\'` is probably used for back references, so this check is faster than parsing
   # the whole thing.
-  def has_back_references?
+  def has_back_references? : Int32?
```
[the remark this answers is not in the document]
c-cube: it's actually `Int32?` but I just updated it with `!!` to make it a proper bool. And added a spec.

src/string.cr
```diff
@@ -814,32 +816,32 @@ class String
...
   end
 
-  def []?(regex : Regex, group)
+  def []?(regex : Regex, group : Int | String) : String?
     match = match(regex)
     match[group]? if match
   end
 
-  def [](str : String | Char)
+  def [](str : String | Char) : String | Char
```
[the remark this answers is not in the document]
c-cube: removed.

RX14:
Aiming to annotate all return types isn't good. If you're not sure, or it's complex, leave it off.

This should be documentation of existing implicit types, not restricting them more.

wontruefree:
Also it restricts types that might not need to be restricted.

src/string.cr
```diff
@@ -2040,7 +2042,7 @@ class String
   # "hello".gsub { |char| char + 1 } # => "ifmmp"
   # "hello".gsub { "hi" }            # => "hihihihihi"
   # ```
-  def gsub(&block : Char -> _)
+  def gsub(&block : Char -> Char | String) : String
```
RX14: The return type of this block can actually be anything with a `.to_s` method. Please leave the block type restriction as-is.

src/string.cr
```diff
@@ -2165,7 +2167,7 @@ class String
   # ```
   # "hello yellow".gsub("ll", "dd") # => "heddo yeddow"
   # ```
-  def gsub(string : String, replacement)
+  def gsub(string : String, replacement : Char | String)
```
RX14: ditto, revert this.

src/string.cr
```diff
@@ -2179,7 +2181,7 @@ class String
   # ```
   # "hello yellow".gsub("ll") { "dd" } # => "heddo yeddow"
   # ```
-  def gsub(string : String, &block)
+  def gsub(string : String, &block : String -> Char | String) : String
```
RX14: ditto

src/string.cr
```diff
@@ -2271,7 +2273,7 @@ class String
   # ```
   # "aabbcc".count { |c| ['a', 'b'].includes?(c) } # => 4
   # ```
-  def count
+  def count(&block : Char -> _) : Int
```
RX14: `Int32`.

src/string.cr
```diff
@@ -2284,14 +2286,14 @@ class String
   # ```
   # "aabbcc".count('a') # => 2
   # ```
-  def count(other : Char)
+  def count(other : Char) : Int
```
RX14: ditto

src/string.cr
```diff
@@ -2284,14 +2286,14 @@ class String
...
     count { |char| char == other }
   end
 
   # Sets should be a list of strings following the rules
   # described at `Char#in_set?`. Returns the number of characters
   # in this string that match the given set.
-  def count(*sets)
+  def count(*sets) : Int
```
RX14: ditto

src/string.cr
```diff
@@ -2302,7 +2304,7 @@ class String
   # ```
   # "aabbcc".delete { |c| ['a', 'b'].includes?(c) } # => "cc"
   # ```
-  def delete
+  def delete(&block : Char -> Bool) : String
```
RX14: Sometimes you have typed these blocks as `-> _`, so that truthiness works, sometimes you typed as `-> Bool` so you need to explicitly return a bool.

The former is descriptive, the latter is prescriptive, a breaking change, and belongs in another PR if at all.

src/string.cr
```diff
@@ -2339,7 +2341,7 @@ class String
   # "aaabbbccc".squeeze { |c| ['a', 'b'].includes?(c) } # => "abccc"
   # "aaabbbccc".squeeze { |c| ['a', 'c'].includes?(c) } # => "abbbc"
   # ```
-  def squeeze
+  def squeeze(&block : Char -> Bool) : String
```
RX14: ditto

src/string.cr
```diff
@@ -3716,7 +3718,7 @@ class String
 
   # Searches the string for instances of *pattern*,
   # yielding a `Regex::MatchData` for each match.
-  def scan(pattern : Regex)
+  def scan(pattern : Regex, &block) : Nil
```
RX14: `&block : Regex::MatchData ->`?

src/string.cr
```diff
@@ -3805,7 +3807,7 @@ class String
   # end
   # array # => [{'a', 0}, {'b', 1}, {'☃', 2}]
   # ```
-  def each_char_with_index
+  def each_char_with_index(&block)
```
RX14: `Char, Int32 ->`...

src/string.cr
```diff
@@ -3837,7 +3839,7 @@ class String
   # ```
   #
   # See also: `Char#ord`.
-  def each_codepoint
+  def each_codepoint(&block)
```
RX14: `Int32 ->`

src/string.cr
```diff
@@ -4157,7 +4159,7 @@ class String
   # "hello".size # => 5
   # "你好".size    # => 2
   # ```
-  def size
+  def size : Int
```
RX14: This returns `typeof(@length)`, so `Int32`.

src/string.cr
c-cube: adding this annotation makes the spec fail with:

```
Failures:

  1) scans using $~

       Nil assertion failed (NilAssertionError)
         from src/nil.cr:106:5 in 'not_nil!'
         from spec/std/string_spec.cr:1487:39 in '->'
         from src/spec/example.cr:255:3 in 'run'
         from src/spec/context.cr:207:23 in 'run'
         from src/spec/context.cr:42:23 in 'run'
         from src/spec/dsl.cr:162:7 in '->'
         from src/kernel.cr:255:3 in 'run'
         from src/crystal/main.cr:47:5 in 'main'
         from __libc_start_main
         from _start
         from ???
```
‽

RX14: @asterite any reason why adding this type restriction would break `$1`??

asterite: A compiler bug

straight-shoota:
@c-cube Are you still up to continue with this PR? It needs some fixing and conflict resolution.
Otherwise let us know and someone else can take over.

c-cube:
I wasn't aware this was still on the table, but by all means, please, take over :)

straight-shoota (approved):

src/string.cr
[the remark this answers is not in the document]
funny-falcon: @c-cube looks like you forgot to push: it is still here.

PR #8178, opened 2019-09-13
Revert semantical evaluation of pending blocks

straight-shoota:
Follow up to #8125 which reverts the change that the contents of `pending` blocks are evaluated in the compiler.

It's not necessary to compile pending blocks since they're not executed anyway. Avoiding this allows to define pending specs which uses code that is not currently working but supposed to be implemented/fixed at some point. It still needs to be valid Crystal, though obviously.

The alternative would be to comment out the code that does not compile. But I prefer leaving the spec explicitly written in code, and when the implementation is done, you simply need to replace `pending` with `it` and it will compile.

To only check if some piece of code compiles it can simply be wrapped in `typeof()`. This is applied in numerous places around the stdlib specs. You shouldn't use `pending` for this.

/cc https://github.com/crystal-lang/crystal/pull/8125#issuecomment-530708236

### DIFF
--- a/spec/std/spec_spec.cr
+++ b/spec/std/spec_spec.cr
@@ -126,6 +126,12 @@ describe "Spec matchers" do
     ex.message.should eq "can't nest `it` or `pending`"
     ex.file.should eq __FILE__
   end
+
+  describe "pending block is not compiled" do
+    pending "pending has block with valid syntax, but invalid semantics" do
+      UndefinedConstant.undefined_method
+    end
+  end
 end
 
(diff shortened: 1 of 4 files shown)


straight-shoota:
I'm putting this on the 0.31.0 milestone to avoid a breaking change introduced by #8125. 
This is already breaking the nightly specs for [crinja](https://circleci.com/gh/straight-shoota/crinja/506) and [sass.cr](https://circleci.com/gh/straight-shoota/sass.cr/411).

PR #8183, opened 2019-09-14
[WIP] Add Reader, Writer for OpenSSL::Cipher

omarroth:
This is an example implementation for adding `Cipher::Reader` and `Cipher::Writer`, similarly to `Zlib`/`Gzip` already in the standard library.

It makes code for encryption/decryption much cleaner IMO. The following, for example:

```crystal
# Assuming key, iv is already defined

cipher = OpenSSL::Cipher.new("aes-128-cbc")
cipher.encrypt
cipher.key = key
cipher.iv = iv

io = IO::Memory.new
io.write(cipher.update("Some plaintext"))
io.write(cipher.final)
```

becomes

```crystal
# Assuming key, iv is already defined

io = IO::Memory.new
OpenSSL::Cipher::Writer.open(io, "aes-128-cbc", key, iv) do |writer|
    writer << "Some plaintext"
end
```

Additionally it becomes possible to wrap IOs (a socket or file, for example) like so:

```crystal
file = File.open("my_secrets", mode: "w")
file = OpenSSL::Cipher::Writer.new(file, "aes-128-cbc", key, iv)
file << "Some plaintext"
file.close
```

I consider this an interesting proposal for discussion but in no way a finished PR.

Couple thoughts/issues currently:

It may make sense to implement this instead as a `CipherBase`, `CipherIO`, similar to `DigestBase`, `DigestIO`, etc.

Currently `Cipher::Reader` will write padding to the IO, which causes the added spec to fail. I'm unsure of the best way to go about fixing this.

### DIFF
--- a/spec/std/openssl/cipher_spec.cr
+++ b/spec/std/openssl/cipher_spec.cr
@@ -1,5 +1,6 @@
 require "../spec_helper"
 require "openssl/cipher"
+require "openssl/cipher/*"
 
 describe OpenSSL::Cipher do
   it "encrypts/decrypts" do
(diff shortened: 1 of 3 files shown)


didactic-drunk:
You'll also have issues when padding is off and writes aren't aligned to the cipher's block size.

(Not in a position to approve your PR)

PR #8185, opened 2019-09-15
JSON and YAML: allow deserializing abstract types

asterite:
This is something that wasn't possible right now, and I think just trying to deserialize every concrete subtype makes sense. In a way this is similar to how unions are deserialized.

### DIFF
--- a/src/json/serialization.cr
+++ b/src/json/serialization.cr
@@ -115,6 +115,12 @@ module JSON
   #   @a : Int32?
   # end
   # ```
+  #
+  # ### Abstract types
+  #
+  # When deserializing an abstract type all concrete subtypes will be
+  # tried to deserialized one by one: the first one that succeeds will
+  # be returned. Otherwise a `JSON::ParseException` is raised.
   module Serializable
     annotation Options
     end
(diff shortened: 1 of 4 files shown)


bcardiff:
A way to prioritize subtypes should be considered. Like an `attribute DeserializePriority` or similar.

asterite:
@bcardiff Maybe, I don't know. If you have many possible types that an API allows I think there shouldn't be ambiguous parsing.

But... should it be part of this PR?

And also, if you need an order you can always override `self.new` and do custom logic there.

straight-shoota:
@bcardiff That sounds like adding a lot of complexity and is IMO not strictly necessary.

src/json/serialization.cr
```diff
@@ -115,6 +115,12 @@ module JSON
   #   @a : Int32?
   # end
   # ```
+  #
+  # ### Abstract types
+  #
+  # When deserializing an abstract type all concrete subtypes will be
```
RX14: Perhaps say explicitly "when including this module in an abstract type"?

PR #8186, opened 2019-09-15
XML::Reader#expand changes

Blacksmoke16:
* Makes `XML::Reader#expand` raise a `XML::Error` if the node could not be expanded.
* Adds a new method `XML::Reader#expand?` that retains the previous behavior of returning `nil` on error.


Blacksmoke16:
Anything left to do here?

straight-shoota:
Thank you

PR #8192, opened 2019-09-16
Compiler: fix missing `name_location` of some calls

### LINKED ISSUE #8191: Failed to compile a very basic HTTP server with --release flag
```crystal
# server.cr
require "http/server"

  context.response.print "Hello world, got #{context.request.path}!"
end

address = server.bind_tcp("127.0.0.1", 8090)

puts "Listening on http://127.0.0.1:8090"

server.listen
```

```console
crystal server.cr --release 
Module validation failed: inlinable function call in a function with debug info must have a !dbg location
  call void @__crystal_raise_overflow()
inlinable function call in a function with debug info must have a !dbg location
  call void @__crystal_raise_overflow()
 (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
```

Ubuntu 19.04.
Crystal installed from Snap store.

```console
crystal -v                         
Crystal 0.31.0-dev [d1a8e7723] (2019-09-16)

LLVM: 8.0.0
```

asterite:
Fixes #8191

I _think_ these were all the missing cases, but I'm not sure.

We have two choices:
1. use `location` if `name_location` is missing: that way we'll never know if we missed setting a `name_location` somewhere
2. just use `name_location` and crash if it's missing: bad for users, though they can still compile their code with `--no-debug`, but we'll know we are missing something and we can fix it

Probably 2 is better, so I didn't change anything else.

In any case, this line is smelly:

https://github.com/crystal-lang/crystal/blob/d1a8e77237aa8af11a63ac4f98368d6d432cd113/src/compiler/crystal/codegen/primitives.cr#L17

it's strange that we check `call.location` to determine copying `call.name_location`. Shouldn't we check for `name_location`?

(though this seems to work, probably because `location` is set in more places than `name_location`)

### DIFF
--- a/spec/compiler/normalize/op_assign_spec.cr
+++ b/spec/compiler/normalize/op_assign_spec.cr
@@ -3,7 +3,9 @@ require "../../spec_helper"
 describe "Normalize: op assign" do
   ["+", "-", "*", "&+", "&-", "&*"].each do |op|
     it "normalizes var #{op}=" do
-      assert_normalize "a = 1; a #{op}= 2", "a = 1\na = a #{op} 2"
+      node = assert_normalize "a = 1; a #{op}= 2", "a = 1\na = a #{op} 2"
+      assert_name_location node.as(Expressions).expressions[1].as(Assign).value,
+        1, 10
     end
   end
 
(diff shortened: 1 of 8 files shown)


bew:
What's the difference between `location` and `name_location` ?

asterite:
@bew

```
foo.bar(1 + 2)
^   ^
|   |------ name_location
|
|--------- location
```

When an error says `undefined method "bar"` it will point and underline "bar" instead of somewhere at "foo". This is specially needed when "bar" appears in a different line:

```
foo.
  bar(1 + 2)
```

In the case of `1 + 2` I think the `name_location` will point to `+` instead of `1`.

PR #8194, opened 2019-09-16
Fix examples 2019 09

maiha:
Congrats 0.31.0! Let's catch up the examples code before release!

- target src: master branch ( 390aa89b9 `Merge pull request #8192 from asterite/bug/missing-name-location` )
- compiler: `Crystal 0.31.0-dev [390aa89] (2019-09-16)`

Best regards,

### DIFF
--- a/src/system/user.cr
+++ b/src/system/user.cr
@@ -5,8 +5,10 @@ require "crystal/system/user"
 # Users can be retrieved by either username or their user ID:
 #
 # ```
+# require "system/user"
+#
 # System::User.find_by name: "root"
-# System::User.find_by id: "1000"
+# System::User.find_by id: "0"
 # ```
 class System::User
   # Raised on user lookup failure.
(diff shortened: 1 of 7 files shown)


src/system/user.cr
```diff
@@ -5,8 +5,10 @@ require "crystal/system/user"
 # Users can be retrieved by either username or their user ID:
 #
 # ```
+# require "system/user"
+#
 # System::User.find_by name: "root"
-# System::User.find_by id: "1000"
+# System::User.find_by id: "0"
```
maiha: I changed the value from 1000 to 0 for two reasons.

1. This explains "two ways" of getting user data by **id** and **name**. Therefore, the intention of the code becomes clearer when the target is the same data.

2. I am using docker to verify the sample code, and id = 1000 does not exist in the container, so spec fails in macro level.

bcardiff:
Thanks @maiha! 

The `--error-on-warnings` could be used to prevent docs using deprecated methods or language constructs. Warnings are now turned on by default but I am not sure if that raise a flag in maiha/crystal-examples

PR #8195, opened 2019-09-17
Fix compile on win32

straight-shoota:
This applies the arithmetic operator change to win32-specific implementation and moves references to `Fiber` behind a flag. The latter can be removed after #7995, but right now it is required to build specs on win32.

### DIFF
--- a/src/gc/boehm.cr
+++ b/src/gc/boehm.cr
@@ -254,20 +254,22 @@ module GC
   end
 
   # pushes the stack of pending fibers when the GC wants to collect memory:
-  GC.before_collect do
-    Fiber.unsafe_each do |fiber|
-      fiber.push_gc_roots unless fiber.running?
-    end
-
-    {% if flag?(:preview_mt) %}
-      Thread.unsafe_each do |thread|
-        if scheduler = thread.@scheduler
-          fiber = scheduler.@current
-          GC.set_stackbottom(thread, fiber.@stack_bottom)
-        end
+  {% unless flag?(:win32) %}
+    GC.before_collect do
+      Fiber.unsafe_each do |fiber|
+        fiber.push_gc_roots unless fiber.running?
       end
-    {% end %}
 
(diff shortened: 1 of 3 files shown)


src/gc/boehm.cr
```diff
@@ -254,20 +254,22 @@ module GC
   end
 
   # pushes the stack of pending fibers when the GC wants to collect memory:
-  GC.before_collect do
-    Fiber.unsafe_each do |fiber|
-      fiber.push_gc_roots unless fiber.running?
-    end
+  {% unless flag?(:win32) %}
```
bcardiff: I don't understand the reason for this change.
The GC in windows still needs to push the gc root, and the extensions are already behind preview_mt.

src/spec/example.cr
```diff
@@ -35,11 +35,13 @@ module Spec
         ensure
           Spec.run_after_each_hooks
 
-          # We do this to give a chance for signals (like CTRL+C) to be handled,
-          # which currently are only handled when there's a fiber switch
-          # (IO stuff, sleep, etc.). Without it the user might wait more than needed
-          # after pressing CTRL+C to quit the tests.
-          Fiber.yield
+          {% unless flag?(:win32) %}
```
bcardiff: I guess this is because Fiber.yield is not allowed in windows. If so, then maybe is better to add the `flag?(:win32)` check inside `Fiber.yield`

src/gc/boehm.cr
straight-shoota: Master fails to compiler because `Fiber` is not yet implemented on win32. It is only added in #7995

straight-shoota: This is only temporary and will be removed in #7995. But it's necessary to run the specs prior to that.

src/spec/example.cr
straight-shoota: `Fiber` itself is currently undefined in win32.

PR #8196, opened 2019-09-17
Fix corruption of thread linked list

### LINKED ISSUE #8184: GC.collect might not be thread-safe
Many CI failure sin the `preview_mt` part always fail in a spec that checks that `finalize` is called, and for that `GC.collect` is invoked.

I believe there is some missing thread-safety there. I know the usual GC collection should be thread safe by now: the one that is automatically triggered when there's no more heap memory and it needs to be expanded. `GC.collect` is different: it's an explicit request to run the GC.

I already reported this through other mediums, I'm just tracking it here so we don't forget it.

Proofs (all of these, and more, crash after a call to `GC.collect`):
- https://circleci.com/gh/crystal-lang/crystal/28929
- https://circleci.com/gh/crystal-lang/crystal/28894
- https://circleci.com/gh/crystal-lang/crystal/28886
- https://circleci.com/gh/crystal-lang/crystal/28872

waj:
When a new thread finishes very quickly it could happen that the removal from the linked list was executed before the push. The linked list didn't check for that situation and it ends corrupted.

This is what was causing errors while the GC runs on the specs. There are some tests that create new threads with an almost empty body. 

Fixes #8184

### DIFF
--- a/src/thread.cr
+++ b/src/thread.cr
@@ -34,9 +34,7 @@ class Thread
       Pointer(Void).null
     }, self.as(Void*))
 
-    if ret == 0
-      Thread.threads.push(self)
-    else
+    if ret != 0
       raise Errno.new("pthread_create", ret)
     end
   end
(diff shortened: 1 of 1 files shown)


asterite:
A comment about this: it looks like a simple PR but it took days to find out the cause. The work these fixes take isn't related to the code change size at all!

Issue #6431, opened 2018-07-22
More helpful `Nil assertion failed` error message


Blacksmoke16:
https://play.crystal-lang.org/#/r/4kwn

```Crystal
class Foo
  property! name : String
end


foo = Foo.new

p foo.name
```

The exception for `Nil assertion failed`, right now, is quite lacking on details.  Including the variable name and/or line number would aid greatly in trying to debug larger programs where it may not be super obvious where the error is coming from.

j8r:
What do you think of having a reserved method `#not Nil`, that will be the opposite of `#as T`?

bew:
> no more cryptic `Nil assertion failed`.

Well you'd still have the same kind of message.. How would that `#not` be better than `#not_nil!` ?

j8r:
Hum that was my first thought - it may be only possible with a reserved method.

bew:
@j8r I don't understand, what are you expecting exactly? Can you give an example?

j8r:
I can be done with a macro at least, but that's not really OOP. I think something at least like `String | Nil shouldn't be Nil`

ysbaddaden:
The `property!` macro can be fixed to print something like:
> nil assertion error, expected T but got Nil

Where `T` is the non nilable property type, when its present, otherwise we just print the basic "nil assertion error".

The `#not(T)` proposal is unrelated to this issue, and wouldn't help to print a nicer error message over `#not_nil!`.

bew:
Could we allow something like this:
```cr
class Object
  def foo
    p! typeof(self)
  end
end

v = 1.as(Int32 | String)
v.foo
```
To print `Int32 | String` instead of `Int32` ?

_Maybe not this exact syntax_

bcardiff:
@bew That is unrelated. The `1` is a value and the type is `Int32`. That value can be used as a value of `Int32 | String` but the value, semantically speaking, is the same. Even if the memory representation differs, the **value** is `1`.

bew:
Hmm maybe `v : Int32 | String = 1; v.foo` is more accurate. Basically a way to get the compile type of `v` from a method on `Object`..

But I guess this is stupid/useless/not wanted.

straight-shoota:
@bew That's not possible. `typeof` shows the type of an expression. So you `typeof(v)` is `Int32 | String`. But the instance itself doesn't know anything about being assigned to a variable.
This should become obvious when you consider assigning the same value to a second variable having different type restrictions.

[...]

PR #8200, opened 2019-09-19
Add name to error message in non-nil asserting getters

icy-arctic-fox:
Adds the getter's name to the error message when the non-nil assertion fails. This resolves issue #6431 . Affects `getter!`, `property!`, `class_getter!`, and `class_property!`.

```crystal
class Foo
  getter! bar

  def initialize(@bar : Int32? = nil)
  end
end

Foo.new.bar # NilAssertionError: bar can't be nil
```

### DIFF
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -229,7 +229,7 @@ describe Object do
   describe "getter!" do
     it "uses getter!" do
       obj = TestObject.new
-      expect_raises(Exception, "Nil assertion failed") do
+      expect_raises(NilAssertionError, "getter5 cannot be nil") do
         obj.getter5
       end
       obj.getter5 = 5
@@ -240,7 +240,7 @@ describe Object do
 
     it "uses getter! with type declaration" do
       obj = TestObject.new
-      expect_raises(Exception, "Nil assertion failed") do
+      expect_raises(NilAssertionError, "getter6 cannot be nil") do
         obj.getter6
       end
       obj.getter6 = 6
@@ -379,7 +379,7 @@ describe Object do
   describe "property!" do
     it "uses property!" do
       obj = TestObject.new
-      expect_raises(Exception, "Nil assertion failed") do
+      expect_raises(NilAssertionError, "property5 cannot be nil") do
         obj.property5
       end
       obj.property5 = 5
@@ -388,7 +388,7 @@ describe Object do
 
     it "uses property! with type declaration" do
       obj = TestObject.new
-      expect_raises(Exception, "Nil assertion failed") do
+      expect_raises(NilAssertionError, "property6 cannot be nil") do
         obj.property6
       end
       obj.property6 = 6
--- a/src/object.cr
+++ b/src/object.cr
@@ -496,7 +496,7 @@ class Object
 
         def {{method_prefix}}\{{name.id}}
           if (\%value = {{var_prefix}}\{{name.id}}).nil?
-            raise NilAssertionError.new("\{{name.id}} cannot be nil")
+            ::raise NilAssertionError.new("\{{name.id}} cannot be nil")
           else
             \%value
           end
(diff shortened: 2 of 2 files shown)


straight-shoota:
The improved error message for getters is good. 

But I'm not in support of adding a custom error message to `#not_nil!`. Using this method should be avoided as much as possible and it's generally considered a code smell. There are rare occasions where it's necessary, but these should really be the exception. So adding too much convenience seems counter-productive, as it might encourage using it.

Besides, you can usually just raise a custom exception:
```cr
foo.not_nil! "because bar"
foo || raise "because bar"
```
This has the same size, but offers more flexibility because you can fully customize the exception being raised.
It doesn't work with falsey `Bool` values, though. But calling `.not_nil!` on a `Bool?` type is pretty rare.

icy-arctic-fox:
Not trying to argue for the `#not_nil!` variant, but rather have a discussion and provide use cases. I'm totally fine with removing it and just improving the `getter!` error message.

I'm not much of a fan of `#not_nil!` either, since it is [discouraged](https://github.com/rubocop-hq/ruby-style-guide#no-non-nil-checks) to explicitly check for nil. In its current state, `#not_nil!` is not very useful, especially for debugging. The only message provided is "Nil assertion failed," which doesn't tell what or why. It would be much more helpful to have a message explaining it. But to your point, you can just raise your own exception.

`#not_nil!` is more explicit that nil should not be given.

```crystal
nil.not_nil!("Nil not expected here")
```

It's also shorter than writing:

```crystal
nil || raise NilAssertionError.new("Nil not expected here")
```

Which is equivalent to what it does now. If you care about the exception type, then it gets lengthy. I prefer the "or raise" case myself. Or for a workaround, I used:

```crystal
if (parent = @parent)
  parent.do_something
else
  raise NilAssertionError.new("Parent node can't be nil")
end
```

Where `#not_nil!` is beneficial over the "or raise" approach is in method chaining. When searching through the Crystal repository for usages, using `#not_nil!` in a method chain is quite common. Most of the time though, the value is not nil, so this is a quick way to remove nil from the type list. But in the case it is nil, a "Nil assertion failed" message and a stack trace is produced, then it's up to the developer to figure out why. This is especially bad, for example, in a recursive data structure, like JSON, expecting a specific layout. So add a custom message, right?

```crystal
(foo.bar || raise "bar can't be nil").baz
# or...
if (b = foo.bar)
  b.baz
else
  raise "bar can't be nil"
end
# or even...
def bar_not_nil
  foo.bar || raise "bar can't be nil"
end
bar_not_nil.baz
```

It gets quite ugly. An alternative:

```crystal
foo.bar.not_nil!("bar can't be nil").baz
```

This is shorter and cleaner, but still kind of weird having a message thrown in there.

straight-shoota:
`#not_nil!` shouldn't be used for data validation. Its single use case IMO is for when a nilable variable can't be nil in a specific context, but the compiler can't prove it.

icy-arctic-fox:
Update this PR to just have improved error messages for `getter!` and related macros? or create a new one?

j8r:
Not needed, the block variant of getter can already be used for that:
```cr
class Klass
  getter str : String { raise "oops" }
end
```
Even if `getter!` is in certain (rather rare) cases handy, we could easily live without it – and even better because one will be encouraged to specify an error message.

straight-shoota:
@j8r Improving the error message for `getter!` is a good thing. The macro knows about the context (the name of the variable), so this should be included in the error message.

@icy-arctic-fox New PR or this one, your choice.

bcardiff:
@straight-shoota why you consider this a breaking-change? The exception type raised is the same. The message changes but not its type.

asterite:
I think it would be better to somehow include the type name in the message, like "`Foo#bar` was nil". Because seeing "name was nil" in a trace isn't much better than "something was nil" because "name" can appear in multiple types.

Sija:
sth along `{{@type}}{{method_prefix}}\{{name.id}}` would be IMO ideal.

straight-shoota:
@bcardiff Maybe because `Exception` was changed to `NilAssertionError` in the specs. But that's not breaking, the actual type didn't change.

Issue #8169, opened 2019-09-10
Wrong type suggestion by the compiler


BerkhanBerkdemir:
Today, I was playing with Crystal and saw something interesting, and that could be UX bug. I posted the sample code under the error message. The expected result should be `Hello, 10` something; however, compiler complains:

```
$ crystal run src/lorem.cr
Showing last frame. Use --error-trace for full trace.

In src/lorem.cr:9:17

 9 | private def foo() : UInt16
                 ^--
Error: type must be UInt16, not Int32
```

So, I believe that should be `Int32`, but the compiler says it should be `UInt16` which makes me confused.

## Sample code

```crystal
require "random"

module Lorem
  class Ipsum
    def initialize
     puts "Hello, #{foo()}"
    end

    # It's just a way to define a random function
    private def foo() : UInt16
      Random.new.rand(0..99)
    end
  end
end

Lorem::Ipsum.new
```

## 

## Extra

```
$ crystal -v
Crystal 0.30.1 [5e6a1b672] (2019-08-12)

LLVM: 4.0.0
Default target: x86_64-unknown-linux-gnu
```

Blacksmoke16:
```cr
private def foo() : UInt16
  Random.new.rand(0..99)
end
```

Numbers default to `Int32` unless explicitly specified, so your `0..99` would cause `#rand` to return an `Int32` number between 0 and 99.  However, you have a n `UInt16` type restriction on the `#foo` method.  `Int32 != UInt16` hence the error.

The way to fix this would be to specify `0..99` as `UInt16`. I.e. `0_u16..99_u16`.

RX14:
This error message should probably be "method is typed UInt16, but returns Int32" which is a far far less confusing message. Also a very simple PR for someone to make :)

BerkhanBerkdemir:
> This error message should probably be "method return type is restricted to UInt16, but the method returns Int32 instead" (or something similar) which is a far far less confusing message. Also a very simple PR for someone to make :)

After a few days later playing with Crystal, I believe this will be an attempt to improve UX. I can write the PR if you like, @RX14.

PR #8201, opened 2019-09-20
Compiler: change the UX of errors

BerkhanBerkdemir:
This commit closes issue #8169


oprypin:
Why breaking-change

oprypin:
Actually all the labels seem off

asterite:
Hi! Thank you for this!

But this doesn't fix the issue (otherwise some. specs would have failed). I was already working on this and I'll send a PR soon.

I'll close this because this is actually not that trivial to change, or it's not super obvious where to change it. Sorry!

PR #8203, opened 2019-09-20
Compiler: better error for return type mismatch

### LINKED ISSUE #8169: Wrong type suggestion by the compiler

asterite:
Fixes #8169

### DIFF
--- a/spec/compiler/semantic/def_spec.cr
+++ b/spec/compiler/semantic/def_spec.cr
@@ -272,7 +272,7 @@ describe "Semantic: def" do
 
       foo
       ),
-      "type must be Int32, not Char"
+      "method must return Int32 but it is returning Char"
   end
 
   it "errors if return type doesn't match on instance method" do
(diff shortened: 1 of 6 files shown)


RX14:
hell yes

bcardiff:
Seems a safe change to be merged for 0.31.0 IMO.

asterite:
> Seems a safe change to be merged for 0.31.0 IMO.

Let's do it!

PR #8204, opened 2019-09-20
Fix regressions in Number#round to handle medium amount of digits

bcardiff:
I've found in the ecosystem that round was used with some big numbers that cause overflow on integers.


RX14:
RIP CI??

bcardiff:
The last spec was added a bit blindly 🙈 .

Any objections to ship this in 0.31.0?

asterite:
> Any objections to ship this in 0.31.0?

Nope!

PR #8208, opened 2019-09-22
Fix spec reporting with nested context

### LINKED ISSUE #8207: Spec failures no longer include parent describe block descriptions
Using the following test case:

```cr
require "spec"

describe Bool do
  describe "that is true" do
    it "should be true" do
      false.should be_true
    end
  end
end
```

On `0.30.1`:
```bash
crystal test.cr 
F

Failures:

  1) Bool that is true should be true
     Failure/Error: false.should be_true

       Expected: true
            got: false

     # test.cr:6

Finished in 73 microseconds
1 examples, 1 failures, 0 errors, 0 pending

Failed examples:

crystal spec test.cr:5 # Bool that is true should be true
```

On latest nightly
```bash
crystal test.cr 
F

Failures:

  1) should be true
     Failure/Error: false.should be_true

       Expected: true
            got: false

     # test.cr:6

Finished in 88 microseconds
1 examples, 1 failures, 0 errors, 0 pending

Failed examples:

crystal spec test.cr:5 # should be true
```

Notice how the error message lost the parent describe block descriptions.

bcardiff:
Fixes #8207 

The report of failure was using directly the root context instead of going through the whole hierarchy of contexts.

### DIFF
--- a/spec/std/spec/context_spec.cr
+++ b/spec/std/spec/context_spec.cr
@@ -1,0 +1,20 @@
+require "spec"
+
+private class FakeRootContext < Spec::RootContext
+  def report_formatters(result)
+  end
+end
+
+describe Spec::NestedContext do
+  describe "#report" do
+    it "should include parent's description" do
+      root = FakeRootContext.new
+      child = Spec::NestedContext.new(root, "child", "f.cr", 1, 10, false)
+      grand_child = Spec::NestedContext.new(child, "grand_child", "f.cr", 2, 9, false)
+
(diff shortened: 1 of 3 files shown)


RX14:
can we have specs... for `Spec`?

bcardiff:
Done, but the spec is a plain unit test. The changes in example.cr are not tested and requires some other refactors to do them properly I think.

RX14:
The spec is fine, this PR is great, thank you!

PR #8214, opened 2019-09-24
Allow abort to run without arguments

dbackeus:
There are cases when terminal output is handled separately via logging classes or similar and but we still want to `abort` nicely.

Also aligns functionality with ruby equivalent.

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -515,7 +515,7 @@ end
 
 # Terminates execution immediately, printing *message* to `STDERR` and
 # then calling `exit(status)`.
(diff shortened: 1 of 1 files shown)


RX14:
Why not just use `exit 1` then?

ysbaddaden:
We already expect message to be a nilable. I guess it's fine to make it nil by default.

dbackeus:
@RX14 For the same reason we generally prefer to type `list.each` rather than `for item in list` I suppose. Reads like human language rather than machine language.

j8r:
`abort` is also more user-friendly for those who don't know well return codes.

PR #8215, opened 2019-09-24
 Remove newlines from base64digest 

j8r:
We very often want a SHA512 hash to be single-lined.

Also adds docs to `OpenSSL::DigestBase`

### DIFF
--- a/spec/std/openssl/digest_spec.cr
+++ b/spec/std/openssl/digest_spec.cr
@@ -6,7 +6,7 @@ describe OpenSSL::Digest do
     {"SHA1", "dcf4a1e3542b1a40a4ac2a3f7c92ffdb2d19812f"},
     {"SHA256", "df81eea14671ce970fb1052e9f5dd6dbda652ed37423ed3624120ec1534784a7"},
     {"SHA512", "082907b85fe25c33bba4765185b52993a493cfd24454edf4b977ccd9301a890659c52592456cbd8aeb5215055d9dd4a7d50a4db9961715fb764fb6c393a83192"},
-  ].each do |(algorithm, expected)|
+  ].each do |algorithm, expected|
     it "should be able to calculate #{algorithm}" do
       digest = OpenSSL::Digest.new(algorithm)
       digest << "fooø"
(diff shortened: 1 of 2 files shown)


yxhuvud:
> We very often want a SHA512 hash to be single-lined.

Why do you want that?

j8r:
> Why do you want that?

Because we usually want a SHA512 hash, not having lines breaks inside this very hash.
For example, `yarn.lock`, `APKBUILD`, or even the `sha512sum` command.

What's the use case of having a multi-line hash?!

j8r:
I never see the need of multi-line hash, anywhere, at least for SHA* and MD5

jhass:
@yxhuvud This only changes the `Digest` classes, not the `Base64` implementation itself.

j8r:
@yxhuvud Ok, but real world put the base64 string on a singe-line. It should be at least the default, because the most widespread.
If people really want multi-line, we may add a boolean argument then.

j8r:
@yxhuvud Ok, but real world applications put the base64 string on a singe-line. It should be at least the default, because the most widespread.

If people really want multi-line, we may add a boolean argument then.

Another option is to remove this method, it may be fine to use `Base64.strict_encode digest` - more flexible and no confusion possible (no questions about strict or not).

RX14:
For `Digest` purposes, `strict_encode` seems like the correct representation.

j8r:
Any update, should we add an argument `strict : Bool = true`?

spec/std/openssl/digest_spec.cr
```diff
@@ -2,11 +2,11 @@ require "spec"
 require "../../../src/openssl"
 
 describe OpenSSL::Digest do
-  [
+  {
```
straight-shoota: This should stay an array. Changing to tuple has no benefit.

src/openssl/digest/digest_base.cr
```diff
@@ -2,36 +2,43 @@ require "base64"
...
-    def digest
-      self.clone.finish
+    # :ditto:
+    def <<(data) : Digest
+      update(data)
     end
 
-    def <<(data)
-      update(data)
+    # Clones and finish the digest.
```
straight-shoota:
```suggestion
    # Clones and finishes the digest.
```

PR #8218, opened 2019-09-24
Channel`s non-blocking receive method added

Odebe:
I think having receive method that does not blocks current fiber is great idea. 
Synthetical example:
```
spawn do
  i = 0
  loop do
    sleep 3
    channel.send(i)
    i += 1
  end
end

spawn do
  loop do
     value = channel.nb_receive
     if value
      puts "#{Time.local}, empty: #{false}, value: #{value}"
     else
      puts "#{Time.local}, empty: #{true}"
     end

     sleep 1
  end
end

sleep 1000000
```

Real use case:
```
class NewClientsHandler
  def initialize(server : TCPServer, channel : Channel(TCPSocket))
    @channel = channel
    @server = server
  end

  def call
    loop do
      if client_socket = @server.accept?
        @channel.send(client_socket)
      end
    end
  end
end


class Server
  def initialize
    @server = TCPServer.new("localhost", '1234')
    @new_clients_channel = Channel(TCPSocket).new
    @clients = [] of ClientHandler
  end

  def start
    start_new_users_handling

    loop do
      handle_new_users

      # process messages from clients
    end
  end

  def start_new_users_handling
    spawn NewClientsHandler.new(@server, @new_clients_channel).call
  end

  def handle_new_users
    new_client_socket = @new_clients_channel.nb_receive
    return if new_client_socket.nil?

    client_handler = ClientHandler.new(new_client_socket)
    @clients << client_handler

    spawn client_handler.call
  end
end

Server.new.start

```

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -157,6 +157,14 @@ class Channel(T)
     receive_impl { return nil }
   end
 
+  # Receives a value from the channel.
+  # If there is a value waiting, it is returned immediately. Otherwise, this method returns nil.
+  #
+  # Returns `nil` if the channel is closed.
+  def nb_receive
+    @lock.sync { closed? ? nil : receive_internal { nil } }
+  end
+
   def receive_impl
     @lock.sync do
       receive_internal do
--- a/spec/std/channel_spec.cr
+++ b/spec/std/channel_spec.cr
@@ -187,13 +187,13 @@ describe "unbuffered" do
     closed.should be_true
   end
 
-  it "return nil on empty channel" do
+  it "returns nil on empty channel" do
     ch = Channel(Int32).new
     value = ch.nb_receive
     value.should eq(nil)
   end
 
-  it "return sended value" do
+  it "returns sent value" do
     ch = Channel(Int32).new
     ch.send(1)
     ch.nb_receive.should eq(1)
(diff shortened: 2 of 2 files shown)


src/channel.cr
```diff
@@ -157,6 +157,14 @@ class Channel(T)
     receive_impl { return nil }
   end
 
+  # Receives a value from the channel.
+  # If there is a value waiting, it is returned immediately. Otherwise, this method returns nil.
+  #
+  # Returns `nil` if the channel is closed.
+  def nb_receive
+    @lock.sync { closed? ? nil : receive_internal { nil } }
```
RX14: The closed check needs to move inside `receive_internal` like in `receive_impl`. And the behaviour of returning values after the channel is closed needs to have a spec (see https://github.com/crystal-lang/crystal/commit/3a9fa4ebd9922036d37f47c61ccee73a59d6d95e).

asterite:
Before sending a PR this should be discussed in a separate issue.

RX14:
Yeah, there's a reason we ask people to open issues before sending PRs when adding or changing API:

The PR should be used to discuss the implementation, the issue should be used to discuss and agree on the API beforehand. When these two discussions happen on the same PR simultaneously, it becomes messy.

src/channel.cr
Odebe: I agree that it should be named differently (maybe `nreceive`? like `index` and `rindex`) and I like idea about add new behaviour to existed method but it requires more complex changes and really should be discussed.

Odebe:
Ok. I am sorry. I will take this into account. 
I have opened [issue](https://github.com/crystal-lang/crystal/issues/8222).

asterite:
Closed until we finish discussing the issue.

Issue #8193, opened 2019-09-16
[RFC] Improve the design of context switch


firejox:
Currently, Fiber switching in crystal is done by invoking Fiber.swapcontext. However, it is hard to keep thread-safe during switching and return dead fiber back to stackpool. I have better idea to handle these problems.

## Split the Fiber.swapcontext into suspend_func and resume_func

If use Fiber.swapcontext, the fiber stack and code segment will look like this :

```
Fiber Stack

hi |             :             |
   |             :             |
   |             :             |
   |             :             |
   +---------------------------+
   |       return address      |
   +---------------------------+ 
   |                           |
   |      stored registers     |
   |                           |
   +---------------------------+ <---- context.stack_top
   |             :             |
   |             :             |
lo |             :             |


Fiber.swapcontext:

    [save register]
    
    [swap stack register pointer]
                                  <---- context switch point
    [restore registers]
    
    [return]

```

If use suspend_func and resume_func, the stack and code segment will be as the following : 

```
Fiber Stack

hi |             :             |
   |             :             |
   |             :             |
   |             :             |
   +---------------------------+
   |       return address      |
   +---------------------------+ 
   |                           |
   |      stored registers     |
   |                           |
   +---------------------------+
   |     resume_func address   |
   +---------------------------+ <---- context.stack_top
   |             :             |
   |             :             |
lo |             :             |

suspend_func:
    
    [save registers]
    
    [save resume_func address]
    
    [swap stack register pointer]
                                  <---- context switch point
    [return]

resume_func:

    [restore registers]
    
    [return]

```

## Benefits

Although there are two more instructions in each context switch, we can make sure some operations can be done atomatically in the future with some mechanism. For example, immediately unlock a mutex after switch.

```
code segment:
    mutex.lock
    target = get_targeted_fiber
    target.add_unlocker(mutex) ---------+
    target.switch                       |
                                        |
                                        |
                                        |
target fiber                            |
                                        |
hi |             :             |        |    |             :             |
   |             :             |        |    |             :             |
   |             :             |        |    +---------------------------+
   |             :             |        |    |       return address      |
   +---------------------------+        |    +---------------------------+
   |       return address      |        |    |      stored registers     |
   +---------------------------+        |    +---------------------------+
   |                           |        V    |     resume_func address   |
   |      stored registers     |------------>+---------------------------+
   |                           |             |       unlock function     |
   +---------------------------+             +---------------------------+
   |     resume_func address   |             |        mutex address      |
[...]

PR #8220, opened 2019-09-24
Improve the design of context switch

firejox:
refer to #8193


wontruefree:
This looks awesome I saw your issue but did not know enough to really respond to it.  Do you have any benchmarks of this change?

firejox:
I benchmark the **threadring** of [kostya/crystal-benchmarks-game](https://github.com/kostya/crystal-benchmarks-game) today. The new design performs slightly better than the origin one. Hence, the change will not cause the run-time overhead.

[result gist](https://gist.github.com/firejox/104ef75e2f8a0d4c5717b3c22182ba09)

wontruefree:
Awesome thank you for the benchmarks.

straight-shoota:
I haven't been able to look at this in depth, yet. But it looks interesting. However, I'd ask to wait with this until after #7995 is merged. This is a performance improvement and new features have a higher priority.

firejox:
@straight-shoota I can wait for that pr because it is just a direction about what we can do. I also agree that making crystal work on windows is :rocket: .

straight-shoota:
@firejox Could you please add back comments that describe the code's behaviour and reasons for that? Instead of removing the existing comments, maybe you can update them for the changes in this PR. This would help reviewing this PR and understanding the code later. Thanks.

firejox:
@straight-shoota Ok, I will add them back.

straight-shoota:
This looks really great, especially with the comments that even improve what we had before.

However, I'm a bit overcharged with understanding everything that's going on. But I'm trusting in your expertise which seems to reach far further than mine in this regard 😄 It would be great if someone else could review this who is more familiar with the topic.

Did you validate this is working as expected on native hardware? I've tested it on x86-64 both linux and windows. But especially the ARM variants might be bit tricky?

ysbaddaden:
Yes, this MUST be tested on real ARM and AArch64 hardware if it wasn't. QEMU isn't enough, and real hardware will immediately exhibit any issues —run the std spec suite for example.

yxhuvud:
@straight-shoota At least some ARM variants should be pretty simple to test nowadays that Apple M1 is everywhere. (I don't have one though, so I can't do it). And we also have aarch64 variants of CI, but perhaps those are QEMU and thus not sufficient?

straight-shoota:
Yeah, indeed the infrastructure situation has improved. Our CI workers run on native ARM machines, so that should be good.

Issue #8213, opened 2019-09-24
Port MT support to upstream bdw-gc getstackbottom patch


zhangkaizhao:
Hello. It is very exciting to hear that 0.31.0 was released with multi-threading preview support.
But I failed to build with preview_mt flag on Arch Linux.

file: main.cr
```crystal
def main
  channel = Channel(Int32).new(9)

  (1..10).each do |i|
    spawn do
      channel.send(i)
    end
  end

  result = [] of Int32
  9.times do
    result << channel.receive
  end

  puts result
end

main
```

It is OK to build without the `preview_mt` flag:
```shell
$ crystal build -o main main.cr
$ ./main
[1, 2, 3, 4, 5, 6, 7, 8, 9]
```
But failed to build with the `preview_mt` flag:
```shell
$ crystal build -Dpreview_mt -o main main.cr
/usr/bin/ld: G-C-.o: in function `set_stackbottom':
/usr/lib/crystal/gc/boehm.cr:202: undefined reference to `GC_set_stackbottom'
/usr/bin/ld: G-C-.o: in function `current_thread_stack_bottom':
/usr/lib/crystal/gc/boehm.cr:194: undefined reference to `GC_get_stackbottom'
collect2: error: ld returned 1 exit status
Error: execution of command failed with code: 1: `cc "${@}" -o '/home/kaizhao/lab/crystal/mt/main'  -rdynamic  -lpcre -lm -lgc -lpthread /usr/lib/crystal/ext/libcrystal.a -levent_pthreads -levent -lrt -ldl -L/usr/lib -L/usr/local/lib`
```

I have read the great article [Parallelism in Crystal](https://crystal-lang.org/2019/09/06/parallelism-in-crystal.html) and I think the code should be valid. Please correct me if I made mistakes.
Just don't know or this is an issue caused by packaging in Arch Linux or crystal itself. So I create this issue.
The version of crystal which is installed via Arch Linux `pacman` package management tool:
```shell
$ crystal -v
Crystal 0.31.0 (2019-09-23)

LLVM: 8.0.1
Default target: x86_64-pc-linux-gnu
```

yxhuvud:
As this is the GC that is failing, it would have made sense to also include the version that is used there. In particular, the missing functions you are seeing comes from very recent versions of the boehm GC (that I think @bcardiff was part of developing, for this purpose).

oprypin:
Maybe the feature hasn't been in a release of libgc yet

[...]

PR #8225, opened 2019-09-24
GC: Use upstream patch for green threads support

bcardiff:
This requires also an update of the distribution-scripts (~~reason why this PR is draft~~).
A compiler with the updated GC 8.0.4 + upstream patch will be available in circleci for a smoke testing. So far it seems to work locally.

The preview_mt jobs will fail with compilation issues since the 0.31.0 has the old patch bundled.

As of the changes itself, bdw-gc patch ended returning an opaque thread handler that is stored now in `Thread.gc_thread_handler` and a `StackBase` structure that is the stackbottom in most architectures. In IA-64 includes the reg_base, but we don't support IA-64. Since the fibers only knows about the stackbottom, I preferred to leave the reg_base commented.

Depends on: https://github.com/crystal-lang/distribution-scripts/pull/47

Fixes: #8213

### DIFF
--- a/src/gc/boehm.cr
+++ b/src/gc/boehm.cr
@@ -17,6 +17,13 @@ lib LibGC
   alias SizeT = LibC::SizeT
   alias Word = LibC::ULong
 
+  struct StackBase
+    mem_base : Void*
+    # reg_base : Void* should be used also for IA-64 when/if supported
+  end
+
+  alias ThreadHandle = Void*
+
   fun init = GC_init
   fun malloc = GC_malloc(size : SizeT) : Void*
   fun malloc_atomic = GC_malloc_atomic(size : SizeT) : Void*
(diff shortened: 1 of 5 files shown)


bcardiff:
A build with the upstream patch is available in `snap refresh crystal --channel=edge/ci-update-gc-dev`.

Unsigned rpm, deb packages and tar.gz can be downloaded from https://circleci.com/gh/crystal-lang/crystal/29489#artifacts/containers/0 in case someone is willing to smoke on their environment.

bcardiff:
On my side the smoke test went well. I've run the benchmarks and nothing explodes, memory consumption was similar than before.
 
If someone else wants to confirm/try and review the changes we can move forward here.

bararchy:
@bcardiff works for us :+1:

src/gc/boehm.cr
```diff
@@ -17,6 +17,13 @@ lib LibGC
   alias SizeT = LibC::SizeT
   alias Word = LibC::ULong
 
+  struct StackBase
+    mem_base : Void*
+    # reg_base : Void* should be used also for IA-64 when/if supported
```
RX14: We will never support IA-64, the architecture is thoroughly dead.

anatol:
I tested this patch with upstream GC changes at Arch Linux and it looks good. Thank you for the work.

One the patch lands crystal repo I'll push a new Arch `crystal` package.

bararchy:
@bcardiff seems both @anatol and our team verified it's working, this will really help up streamline development.
Can we cut a new version? :pray:

Issue #8231, opened 2019-09-25
Channel.select bug when closing channel


carlhoerberg:
The following code: 

```crystal
ch = Channel(Nil).new

spawn do
  loop do
    _, _ = Channel.select(ch.receive_select_action)
    puts "got it"
  end
end

ch.send nil
ch.close
sleep
```

results in:
```
 ✗ crystal run close.cr
got it
Unhandled exception in spawn: BUG: Fiber was awaken from select but no action was activated (Exception)
  from /usr/local/Cellar/crystal/0.31.0/src/channel.cr:321:5 in 'select'
  from /usr/local/Cellar/crystal/0.31.0/src/channel.cr:278:3 in 'select'
  from /usr/local/Cellar/crystal/0.31.0/src/channel.cr:275:5 in 'select'
  from close.cr:5:12 in '->'
  from /usr/local/Cellar/crystal/0.31.0/src/fiber.cr:255:3 in 'run'
  from /usr/local/Cellar/crystal/0.31.0/src/fiber.cr:48:34 in '->'
```

asterite:
Also:

```crystal
ch = Channel(Nil).new

spawn do
  loop do
    select
    when ch.receive
      puts "got it"
    end
  end
end

ch.send nil
ch.close
sleep
```

(it's the same, but please try to use "public" API when reporting these bugs)

carlhoerberg:
sure, but i don't think `select` is documented anywhere, can't find it here: 
https://crystal-lang.org/api/0.31.0/
or here: 
https://crystal-lang.org/reference/

asterite:
Yeah, unfortunately it's not documented because the entire concurrency thing is still being revised and implemented. But `select` is the public API, similar to the one in Go. The others are not (they are implementation details).

I just wanted to show that using public API it breaks so it's a legit bug, and not something that you can only cause by using internal API.

bcardiff:
@carlhoerberg In #8243 put some explanation for the alternatives syntax and behavior. 

If you want to have and "infinite" loop that will exit gracefully (for example upon a Ctrl+C) something like the following might work for you.

```cr
ch = Channel(Nil).new
exit1 = Channel(Nil).new
exit2 = Channel(Nil).new

Signal::INT.trap do
  exit1.send nil
end

spawn do
  i = 0
  run = true
  while run
    i = (i + 1) % 1_000
    select
    when ch.receive
      puts "got it"
    when exit1.receive
      run = false
    else
      if i == 0
[...]

Issue #8230, opened 2019-09-25
Index out of bounds (IndexError) with HTTP::Server


mamantoha:
Hi there!

After updating my Kemal web-app on Heroku with Crystal 0.31.0 I started to get a lot of error messages `Index out of bounds (IndexError)`.

Minimal reproducible code:

```crystal
# main.cr
require "http/server"

server = HTTP::Server.new([
  HTTP::LogHandler.new
]) do |context|
  context.response.content_type = "text/plain"
  context.response.print "Hello world!\n" * 1000 # << HERE
end

puts "Listening on http://0.0.0.0:3000"

server.listen(3000)
```

It is reproducible only with `--release` flag and large body in response.

```
crystal build --release main.cr
```

Each request causes an error

```
GET / - Unhandled exception:
Index out of bounds (IndexError)
  from /snap/crystal/118/share/crystal/src/crystal/thread_local_value.cr:26:7 in 'unbuffered_write'
  from /snap/crystal/118/share/crystal/src/http/server/response.cr:217:11 in 'unbuffered_write'
  from /snap/crystal/118/share/crystal/src/http/server/handler.cr:0:3 in 'call'
  from /snap/crystal/118/share/crystal/src/http/server/request_processor.cr:48:11 in '->'
  from /snap/crystal/118/share/crystal/src/fiber.cr:255:3 in 'run'
  from ???
```

P.S.

```
crystal -v                         
Crystal 0.31.0 [8d4715a59] (2019-09-24)

LLVM: 8.0.0
Default target: x86_64-unknown-linux-gnu
```

mamantoha:
BTW, can't reproduce on macOS.

```
crystal -v
Crystal 0.31.0 (2019-09-23)

LLVM: 8.0.1
Default target: x86_64-apple-macosx
```

vlazar:
Notice it's `LLVM: 8.0.1` on MacOS but `LLVM: 8.0.0` on Linux.

bcardiff:
llvm 8.0.0 or 8.0.1 makes no difference. But the release mode does as @mamantoha said.
The error is also present in llvm 4. I checked since that was updated recently on the linux builds.

bcardiff:
I don't get it still. My guess is that there is something odd with the state of the Slice somewhere. 

[...]

PR #8234, opened 2019-09-25
Compiler: store and restore debug location when going to main

### LINKED ISSUE #6920: Bug when calling module method before module definition
This:
```cr
Module.method

module Module
  @@zero : Int32 = 0

  def self.zero
    @@zero
  end

  @@variable : Int32 = zero

  def self.variable
    @@variable
  end

  def self.method
    @@variable = 0
  end
end
```
results in:
```cr
Module validation failed: !dbg attachment points at wrong subprogram for function
!4 = distinct !DISubprogram(name: "method", linkageName: "method", scope: !5, file: !5, line: 16, type: !6, isLocal: true, isDefinition: true, scopeLine: 16, isOptimized: false, unit: !0, variables: !2)
i32 ()* @"*Module::method:Int32"
  %0 = call i32* @"~Module::variable:read"(), !dbg !9
!9 = !DILocation(line: 6, column: 3, scope: !10)
!10 = distinct !DISubprogram(name: "zero", linkageName: "zero", scope: !5, file: !5, line: 6, type: !6, isLocal: true, isDefinition: true, scopeLine: 6, isOptimized: false, unit: !0, variables: !2)
!10 = distinct !DISubprogram(name: "zero", linkageName: "zero", scope: !5, file: !5, line: 6, type: !6, isLocal: true, isDefinition: true, scopeLine: 6, isOptimized: false, unit: !0, variables: !2)
 (Exception)
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
  from ???
```
But the interesting thing is that this: (when calling `Module.method` **after** the definition of `Module`)
```cr
module Module
  @@zero : Int32 = 0

  def self.zero
    @@zero
  end

  @@variable : Int32 = zero

  def self.variable
    @@variable
  end

  def self.method
    @@variable = 0
  end
end

Module.method
```
compiles.

The `Int32` type doesn't matter, it could also be a `String` or something.
And `Module` could also be a class/struct and the bug still happens.

This might be related to #5656 and #4858 since the `!dbg attachment points at wrong subprogram for function` reason is there too.

### LINKED ISSUE #7060: Crash depending on require order
There's a crash in Crystal 0.27.0 that did not happen in 0.26.1 when using [minitest.cr](https://github.com/ysbaddaden/minitest.cr) (0.4.0 for 0.26.1 / 0.4.2 for 0.27.0) and `HTTP::Client` in the same file that depends on the order of the `minitest/autorun` and `http/client` requires.

The following example works:

```crystal
require "http/client"
require "minitest/autorun"

class FooTest < Minitest::Test
  def test_http_client
    response = HTTP::Client.get("http://www.example.org")
    assert_equal true, response.success?
  end
end
```

However the following code crashes (require order switched):

```crystal
require "minitest/autorun"
require "http/client"

class FooTest < Minitest::Test
  def test_http_client
    response = HTTP::Client.get("http://www.example.org")
    assert_equal true, response.success?
  end
end
```

```
Module validation failed: !dbg attachment points at wrong subprogram for function
!10 = distinct !DISubprogram(name: "initialize", linkageName: "initialize", scope: !5, file: !5, line: 115, type: !6, isLocal: true, isDefinition: true, scopeLine: 115, isOptimized: false, unit: !0, variables: !2)
i8* (%"OpenSSL::BIO"*, %TCPSocket*)* @"*OpenSSL::BIO#initialize<TCPSocket>:Pointer(Void)"
  %2 = call %"struct.LibCrypto::BioMethod"** @"~OpenSSL::BIO::CRYSTAL_BIO:read"(), !dbg !12
!12 = !DILocation(line: 13, column: 3, scope: !13)
!13 = distinct !DISubprogram(name: "set_data", linkageName: "set_data", scope: !5, file: !5, line: 13, type: !6, isLocal: true, isDefinition: true, scopeLine: 13, isOptimized: false, unit: !0, variables: !2)
!13 = distinct !DISubprogram(name: "set_data", linkageName: "set_data", scope: !5, file: !5, line: 13, type: !6, isLocal: true, isDefinition: true, scopeLine: 13, isOptimized: false, unit: !0, variables: !2)
 (Exception)
  from __crystal_main
  from main
```

Example entry for `shard.yml`:

```yaml
development_dependencies:
  minitest:
    github: ysbaddaden/minitest.cr
```

System info:

```
$ crystal -v
Crystal 0.27.0 (2018-11-04)

LLVM: 6.0.1
```

System is macOS Mojave 10.14.1 and Crystal was installed from Homebrew using the precompiled bottle for Mojave.

Maybe @ysbaddaden (author of minitest.cr) has an idea what could be causing this.

### LINKED ISSUE #8233: Trouble finding failure in stack trace
I am upgrading a lib from an earlier version of crystal.  I am have been trying to debug this error and cannot find what the compiler is complaining about.

```bash
crystal run  test/*_test.cr -- 
Module validation failed: !dbg attachment points at wrong subprogram for function
!10 = distinct !DISubprogram(name: "initialize", linkageName: "initialize", scope: !5, file: !5, line: 115, type: !6, isLocal: true, isDefinition: true, scopeLine: 115, isOptimized: false, unit: !0, variables: !2)
i8* (%"OpenSSL::BIO"*, %TCPSocket*)* @"*OpenSSL::BIO#initialize<TCPSocket>:Pointer(Void)"
  %2 = call %"struct.LibCrypto::BioMethod"** @"~OpenSSL::BIO::CRYSTAL_BIO:read"(), !dbg !12
!12 = !DILocation(line: 13, column: 3, scope: !13)
!13 = distinct !DISubprogram(name: "set_data", linkageName: "set_data", scope: !5, file: !5, line: 13, type: !6, isLocal: true, isDefinition: true, scopeLine: 13, isOptimized: false, unit: !0, variables: !2)
!13 = distinct !DISubprogram(name: "set_data", linkageName: "set_data", scope: !5, file: !5, line: 13, type: !6, isLocal: true, isDefinition: true, scopeLine: 13, isOptimized: false, unit: !0, variables: !2)
 (Exception)
  from main
```

It seems like `!5` should be a value that I should be able to read.

asterite:
Fixes #8233
Fixes #7060
Fixes #6920

(which are all just the same bug, but this way it looks like this is a killer PR! 😛)

@bcardiff I'm putting this for 0.31.1 because even though it's not a bug introduced there, it's something that will likely fix a lot of the existing issues with debug info.

### DIFF
--- a/spec/compiler/codegen/debug_spec.cr
+++ b/spec/compiler/codegen/debug_spec.cr
@@ -170,4 +170,24 @@ describe "Code gen: debug" do
       Foo.new.bar
       ), debug: Crystal::Debug::All)
   end
+
+  it "stores and restores debug location after jumping to main (#6920)" do
+    codegen(%(
+      require "prelude"
+
+      Module.method
+
+      module Module
+        def self.value
+          1 &+ 2
+        end
+
+        @@x : Int32 = value
+
+        def self.method
+          @@x
+        end
+      end
+    ), debug: Crystal::Debug::All)
+  end
 end
--- a/src/compiler/crystal/codegen/codegen.cr
+++ b/src/compiler/crystal/codegen/codegen.cr
@@ -1565,6 +1565,8 @@ module Crystal
       old_entry_block = @entry_block
       old_alloca_block = @alloca_block
       old_needs_value = @needs_value
+      old_debug_location = @current_debug_location
+
       @llvm_mod = @main_mod
       @llvm_context = @main_llvm_context
       @llvm_typer = @main_llvm_typer
@@ -1573,6 +1575,8 @@ module Crystal
       @ensure_exception_handlers = nil
       @rescue_block = nil
       @catch_pad = nil
+
+      clear_current_debug_location if @debug.line_numbers?
 
       block_value = yield
 
@@ -1589,6 +1593,7 @@ module Crystal
       @alloca_block = old_alloca_block
       @needs_value = old_needs_value
       context.fun = old_fun
+      set_current_debug_location old_debug_location if @debug.line_numbers?
 
       block_value
     end


paulcsmith:
Woo hoo! This is awesome. I’ve run into this a couple times

paulcsmith:
@bcardiff When do you think 0.31.1 could be released. Right now we're blocked on releasing a new version of Lucky because we can't shutdown HTTP servers/chromedriver in specs (due to the specs being run in `at_exit` now). We tried monkey patching `Spec` but it raises this compile time error so we have to run specs with `--no-debug` to get them to run :S

Long story short, if this was released we'd be able to get a new version of Lucky out with support for the newest Crystal. Don't want to pressure anything, just explain the situation and see if that may affect the release cycle. Or if you have some ideas for helping us that'd be great too

Here is what we're doing now that requires `--no-debug` andI believe would  be fixed by this PR https://github.com/luckyframework/lucky_flow/pull/60/files#diff-906deee07fd614ba421ecd7a4a66ffd1R6

jwoertink:
I can try to run this against master and see how that goes.

asterite:
@paulcsmith What do we need from `Spec`'s public API to be able to fix the problem? I'm thinking maybe `Spec.after(:suite)` or something like that could solve it. I'm not sure whether the problem is that `at_exit` in your code runs before or after Spec.

asterite:
Oh, I just see #8235

jwoertink:
Ok, so I ran 2 different repos that had this issue against master, and both of them give me the same response:

```
[15:57PM] selenium-webdriver-crystal (master)$ /Users/jeremywoertink/Development/crystal/lang/.build/crystal run test/session_test.cr

In test/test_helper.cr:2:1

 2 | require "minitest/autorun"
     ^
Error: can't find file 'minitest/autorun'

If you're trying to require a shard:
- Did you remember to run `shards install`?
- Did you make sure you're running the compiler in the same directory as your shard.yml?
```

```
[15:58PM] lucky_flow (jaw/v0.6.0)$ /Users/jeremywoertink/Development/crystal/lang/.build/crystal spec/

In src/lucky_flow.cr:1:1

 1 | require "selenium"
     ^
Error: can't find file 'selenium'

If you're trying to require a shard:
- Did you remember to run `shards install`?
- Did you make sure you're running the compiler in the same directory as your shard.yml?
```

But both of these throw the module validation error with 0.31.0, so progress! 😄

paulcsmith:
@asterite Yes the `after_suite` method would fix this as well and would be the preferred solution since we wouldn't need to monkey patch `Spec`.

@jwoertink sometimes using Crystal master doesn't get the paths quite right when run like that so I'm guessing there is something up there but I  don't know the fix. I think you can mv the built binary to `/usr/local/bin/crystal-master` or something and then run `crystal-master spec`

jwoertink:
@paulcsmith I tried moving the binary to `/usr/local/bin`, but I get the same error.

```
[16:03PM] lucky_flow (jaw/v0.6.0)$ crystal_dev spec/

In src/lucky_flow.cr:1:1

 1 | require "selenium"
     ^
Error: can't find file 'selenium'
```

I ran `shards install`. Maybe it's cache or something? 🤷‍♂

asterite:
The way to do it is to invoke `bin/crystal` located in the compiler's source code that you have in your machine. For example:
- I have Crystal in `/Users/asterite/Projects/crystal`
- I run: `/Users/asterite/Projects/crystal/bin/crystal program.cr`

That `bin/crystal` wrapper will set `CRYSTAL_PATH` to use the `src` from the compiler's source code in that directory, and then `lib`.

To simplify things I have an alias `ccrystal` that points to `/Users/asterite/Projects/crystal/bin/crystal` so to try things with the new compiler I:
- run `make crystal` in the compiler's directory
- use `ccrystal` to use that new compiler

`ccrystal` is also useful to test the standard library in `master`.

Let me know if using any of those methods works for you guys with `master`. Fingers crossed 🤞

bcardiff:
@paulcsmith there are two outstanding regressions / bugs  I would like to try fix before 0.31.1 #8230 and #8231 . If things go smooth we should be ready for a release tomorrow or on Monday.

paulcsmith:
That’s great news! Thanks @bcardiff

PR #8238, opened 2019-09-26
Spec: add `before_suite` and `after_suite`

### LINKED ISSUE #8235: Feature: Add a hook for before/after all specs run.
After [this PR](https://github.com/crystal-lang/crystal/pull/8125) landed, the way specs ran changed quite a bit. Now if you have any `at_exit` handler in your spec_helper file, that block is ran before your specs instead of at the end. 

It's pretty common to have some code you run first thing before all of your specs, and then another bit when all specs have completed running. In the case of [LuckyFlow](https://github.com/luckyframework/lucky_flow/blob/master/spec/spec_helper.cr#L23-L26) we want to boot up a chromedriver instance, run our specs against that, then when all specs are done, we make sure to shut that process down as well as the HTTP server that's running. 

It would be great to have something like:

```crystal
# Runs once before any actual spec runs
Spec.before_suite do
  setup_test_stuff
end

# Runs once when all specs have completed running (filtered or not)
Spec.after_suite do
  shutdown_test_stuff
end
```

ref: https://gitter.im/crystal-lang/crystal?at=5d8ba89d63b06968fbea378d
ref: https://github.com/luckyframework/lucky/pull/918/files#diff-e6b95b08ee4663a22e90491d70a0cbdaR131-R154

asterite:
Fixes #8235

Also changes the way `Spec.after_each` run: they now run in reverse order.

I tried it in RSpec and it works the same, and it makes sense because one usually does this:

```crystal
# In one part of the code
Spec.before_suite do
  # set up thing 1
end

Spec.after_suite do
  # tear down thing 1
end

# In some other part of the code
Spec.before_suite do
  # set up thing 2
end

Spec.after_suite do
  # tear down thing 2
end
```

and in the end, because of the order things are called, you get:

```
# set up thing 1
# set up thing 2
# tear down thing 2
# tear down thing 1
```

Same goes for `before_each` and `after_each`: they are normally registered in pairs one after the other.

No specs for this, sorry, but the code is relatively trivial.

💭 Now that I think about it, what happens if one hook fails with an exception? I just tried it in RSpec and you get something like "An error occurred in an `after(:suite)` hook" and then a nicely printed exception. We should probably do the same, but I'd like to do that in a separate PR (and it doesn't need to be there in 0.31.1, after all if you get a failure it's something wrong in your setup).

### DIFF
--- a/src/spec/dsl.cr
+++ b/src/spec/dsl.cr
@@ -143,6 +143,44 @@ module Spec
     after_each << block
   end
 
+  # Instructs the spec runner to execute the given block
+  # before the entire spec suite.
+  #
+  # If multiple blocks are registered they run in the order
+  # that they are given.
+  #
+  # For example:
+  #
+  # ```
+  # Spec.before_suite { puts 1 }
+  # Spec.before_suite { puts 2 }
+  # ```
+  #
+  # will print, just before the spec suite starts, 1 and then 2.
+  def self.before_suite(&block)
+    before_suite = @@before_suite ||= [] of ->
+    before_suite << block
+  end
+
+  # Instructs the spec runner to execute the given block
+  # after the entire spec suite.
+  #
+  # If multiple blocks are registered they run in the reversed
+  # order that they are given.
+  #
+  # For example:
+  #
+  # ```
+  # Spec.after_suite { puts 1 }
+  # Spec.after_suite { puts 2 }
+  # ```
+  #
+  # will print, just after the spec suite ends, 2 and then 1.
+  def self.after_suite(&block)
+    after_suite = @@after_suite ||= [] of ->
+    after_suite << block
+  end
+
   # :nodoc:
   def self.run_before_each_hooks
     @@before_each.try &.each &.call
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
It would be nice if you could scope the `before/after_each` to only `it` blocks that are within the `describe` block it was defined.  Having something that executes before _every_ `it` block makes it far less useful.

asterite:
@Blacksmoke16 Yes, I want to do that and I see no problem in that. However I don't want to add too many things for 0.31.1. I think this PR is the necessary minimum to make the ecosystem work again.

asterite:
💭  I would also really like to have `around` filters, though I have no idea how to implement that yet.

Blacksmoke16:
>  I would also really like to have `around` filters, though I have no idea how to implement that yet.

What would that look like?

asterite:
> What would that look like?

```
describe "..." do
  around_each do |example|
    server = ...
    example.run
    server.stop
  end
end
```

though of course I have no idea how would `server` be reachable to each `it`... maybe if `server is a variable outside of that scope it can be done, though that would break the possibility of running specs in parallel in the future because of shared state.

Blacksmoke16:
Oh i see, so essentially something similar to 

```cr
  describe Int32 do
    Spec.before_all do
      server = Server.new
    end

    Spec.after_all do
      server.stop
    end
  
    it "should work" do
      server.some_method 
    end
  end
```

Where it defines variables that all `it` block could use?

asterite:
> Where it defines variables that all it block could use?

You can do that if you define a variable outside of the describe block, but it's not something that `around` would enable just because it's there. So in essence it's the same "hack" but instead of using two hooks (one before, one each) you just need one. Probably not useful...

jwoertink:
@asterite you're amazing! This boosts the core Spec API 10x 🎉 🌮

paulcsmith:
So fast :D Thank you

Issue #6449, opened 2018-07-26
Doc generation and custom annotations issue


anykeyh:
When a comment appears on top of attribute annotation, it won't be catch by crystal docs generation.

Example:
```crystal
    class BillingProperties
      include JSON::Serializable

      # When set to true, Requester Pays is enabled for this bucket.
      @[JSON::Field(key: "requesterPays")]
      property requester_pays : Bool
    end
```

```bash
crystal docs example.cr
```

Obviously, changing the order of the lines appears to work, however I do think the readability of the code would be impacted, further more if the comment is multi-lines.

asterite:
The actual issue is a comment on top of an annotation on top of a macro call which in turns defines a method. Comment on top of annotation on top of regular method works fine.

asterite:
Actually, the issue is that `property` expands to several things and the docs aren't copied to all of them.

bew:
I think that macros should not copy docs automatically, but add a way to insert the doc where we want, for ex:
```cr
macro foo
  {{ @doc }}
  def bar
  end
end

# some doc
foo
```
So we have precise control of where to doc is applied (and how)

KimBurgess:
I've just been taking a look at this - the issue doesn't appear to be on the macro expansion side (that I can see). To narrow the scope a little though, the same behavior occurs with the `getter` macro.

In existing setup, the expansion is:
```crystal
@var : Type

def var : Type
  @var
end
```

If this is flipped around to
```crystal
def var : Type
  @var
end

@var : Type
```
the doc generation works as intended with an without annotations.

Is there a downside to flipping these around until the root issue can be resolved?

Blacksmoke16:
Wouldn't that make the annotation apply to the method and not the ivar?

straight-shoota:
There is already a mechanism implemented in the `PropagateDocVisitor` which duplicates the doc comment of a macro to the features defined in it. A result of this is that the doc comment of a `property` macro is assigned to both the setter and getter method it defines.
So that should be all good.

It seems the issue here is just that for some reason this somehow gets mixed up when an annotation is involved, as per https://github.com/crystal-lang/crystal/issues/6449#issuecomment-495419391

KimBurgess:
@Blacksmoke16 the annotation applies to both the `InstanceVar` and `Def`(s) that these macros pump out.

Blacksmoke16:
[...]

PR #8239, opened 2019-09-26
Workaround IO::Evented#evented_write invalid IndexError error

### LINKED ISSUE #8230: Index out of bounds (IndexError) with HTTP::Server

bcardiff:
Fixes #8230

The issue started appearing after overflow is turned on by default.
But it is not exhibited if `Slice#+` is not inlined in `Evented#evented_write`.
If overflows are avoided in `@size &- offset` but it seems more a hack and fragile since it would be changing the semantics.

### DIFF
--- a/src/io/evented.cr
+++ b/src/io/evented.cr
@@ -70,7 +70,7 @@ module IO::Evented
 
     begin
       loop do
-        bytes_written = yield slice
+        bytes_written = (yield slice).to_i64
         if bytes_written != -1
           slice += bytes_written
           return if slice.size == 0
--- a/src/io/evented.cr
+++ b/src/io/evented.cr
@@ -70,6 +70,7 @@ module IO::Evented
 
     begin
       loop do
+        # TODO: Investigate why the .to_i64 is needed as a workaround for #8230
         bytes_written = (yield slice).to_i64
         if bytes_written != -1
           slice += bytes_written
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -163,7 +163,11 @@ struct Slice(T)
   # slice2 = slice + 2
   # slice2 # => Slice[12, 13, 14]
   # ```
+  @[NoInline] 
   def +(offset : Int)
+    # This method has NotInline as a workaround for #8230. 
+    # To avoid been inlined in Evented#evented_write.
+    # Making `@size &- offset` also seems to fix the issue.
     check_size(offset)
 
     Slice.new(@pointer + offset, @size - offset, read_only: @read_only)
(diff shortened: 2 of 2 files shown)


bcardiff:
I've found nothing wrong at the `.ll` level. And if I `puts` debug I can't reproduce. And from gdb I am unable to view the values of the expression. 

My guess is that in the loop of `Evented#evented_write` something is no in sync.

This is the only workaround I've found that does not change the semantics of the code, but changes the input to llvm.

src/slice.cr
```diff
@@ -163,7 +163,11 @@ struct Slice(T)
   # slice2 = slice + 2
   # slice2 # => Slice[12, 13, 14]
   # ```
+  @[NoInline]
```
Blacksmoke16: I think this would have to be applied above the doc comments, otherwise the method would lose the documentation.

Ref https://github.com/crystal-lang/crystal/issues/6449

bcardiff: Not on regular methods https://github.com/crystal-lang/crystal/issues/6449#issuecomment-495419391 . Just double checked

bcardiff:
There there is another workaround that seems to fix the issue while keeping the inlineable `Slice#+`. It's still a workaround though.

mamantoha:
Verified :tada:

src/io/evented.cr
```diff
@@ -70,7 +70,8 @@ module IO::Evented
 
     begin
       loop do
-        bytes_written = yield slice
+        # TODO: Investigate why the .to_i64 is needed as a workaround for #8230
```
Sija: 👍

PR #8240, opened 2019-09-27
Spec: let `should be_nil` and `should be_a(...)` cast the value

asterite:
A way to solve [this forum request](https://forum.crystal-lang.org/t/var-should-not-be-nil-does-not-eliminate-var-of-nil-from-subsequent-type-inference/1165)

Before this PR we had to do this:

```crystal
index = "foo".index('o')
index.should_not be_nil
foo[index.not_nil!].should eq('o')
```

The `not_nil!` is needed because the compiler has no way to know that the `should_not be_nil` means that `index` will not be `nil`.

With this PR we can now do:

```crystal
index = "foo".index('o')
index = index.should_not be_nil
foo[index].should eq('o')
```

Basically, the result of `obj.should_not be_nil` will return `obj` as a non-nil object.

The same applies for `obj.should be_a(T)` and `obj.should_not be_a(T)`.

This still doesn't automatically restrict the var but is significantly better than the old option.

This PR is a breaking change because I introduced the `Expectation` module that all expectations must include. An alternative to implement this PR is to overload `should` and `should_not` with `BeNil` and `BeA` but it felt like a worse option and it's less extensible.

This is just a draft because together with this PR we can actually change many `should_not be_nil` combined with `not_nil!` in the stdlib specs.

Finally, this PR also finally documents the way to extend `spec` with custom expectations.

💭 In RSpec they are called "matchers", maybe we could rename them to be consistent with the Ruby world? It's also shorter.

### DIFF
--- a/src/spec/expectations.cr
+++ b/src/spec/expectations.cr
@@ -40,6 +40,26 @@ module Spec
     # Returns a failure message for the `should_not` case against this
     # expectation.
     abstract def negative_failure_message(actual_value) : String
+
+    # Casts a value to possibly some other, narrower type, for the result
+    # of a `should` invocation.
+    #
+    # By default this method returns `value` but some matchers, particularly
+    # `be_nil` and `be_a(T)`, will return a non-nil and `T` instance,
+    # respectively.
+    def cast_should(value)
+      value
+    end
+
+    # Casts a value to possibly some other, narrower type, for the result
+    # of a `should_not` invocation.
+    #
+    # By default this method returns `value` but some matchers, particularly
+    # `be_nil` and `be_a(T)`, will return a nil and non-`T` instance,
+    # respectively.
+    def cast_should_not(value)
+      value
+    end
   end
 
   # :nodoc:
@@ -169,6 +189,14 @@ module Spec
     def negative_failure_message(actual_value) : String
       "Expected: #{actual_value.inspect} not to be nil"
     end
+
+    def cast_should(value)
+      value.nil? ? value : (raise "expected #{value} to be nil")
+    end
+
+    def cast_should_not(value)
+      value.nil? ? (raise "expected #{value} to not be nil") : value
+    end
   end
 
   # :nodoc:
@@ -205,6 +233,14 @@ module Spec
 
     def negative_failure_message(actual_value) : String
       "Expected #{actual_value.inspect} (#{actual_value.class}) not to be a #{T}"
+    end
+
+    def cast_should(value)
+      value.is_a?(T) ? value : (raise "expected #{value} to be a #{T}")
+    end
+
+    def cast_should_not(value)
+      value.is_a?(T) ? (raise "expected #{value} to not be a #{T}") : value
     end
   end
 
@@ -388,6 +424,19 @@ module Spec
     end
 
     # Creates an `Expectation` that passes if actual is nil (`== nil`).
+    #
+    # When using it for a `should_not` expectation, the result of the
+    # `should_not` call will be the receiver without the Nil type.
+    #
+    # For example:
+    #
+    # ```crystal
+    # value = "foo".index('o')
+    # typeof(value) # => Int32 | Nil
+    #
+    # result = value.should_not be_nil
+    # typeof(result) # => Int32
+    # ```
     def be_nil : Expectation
       Spec::BeNilExpectation.new
     end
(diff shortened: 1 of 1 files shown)


Fryguy:
Isn't the `.not_nil!` sufficient by itself? That is, you don't need the `should_not be_nil`, because `not_nil!` will raise an exception and fail that example. Similary `.as` can be used to verify a type and raise an runtime assertion (though admittedly it does casting, which *may* not be desirable)

```crystal
require "spec"

context "foo" do
  it "be_nil" do
    x = nil
    x.not_nil!.should eq "x"
  end

  it "be_a" do
    x = rand > 1 ? 1 : nil
    x.as(Int32).should eq 1
  end
end
```

```
  1) foo be_nil

       Nil assertion failed (NilAssertionError)
         from src/nil.cr:106:5 in 'not_nil!'
         from foo.cr:6:5 in '->'
         ...


  2) foo be_a

       cast from Nil to Int32 failed, at /Users/jfrey/projects/external/crystal/foo.cr:11:5:11 (TypeCastError)
         ...
```

Fryguy:
> Finally, this PR also finally documents the way to extend spec with custom expectations.
>
> 💭 In RSpec they are called "matchers", maybe we could rename them to be consistent with the Ruby world? It's also shorter.

+1 for naming them Matchers.  Calling them expectations is confusing because those already have a meaning in RSpec.

In RSpec, "expectations" are the result of `expect(thing)` or `expect { thing }` in the newer expectation syntax, and those can be compared to matchers with the `.to` and `.to_not` methods.  (e.g. `expect(thing).to eq 5`)

I actually prefer the `expect` syntax, personally, and it avoids having to patch methods. @asterite would you be ok with that being added as an alternate spec syntax?

asterite:
@Fryguy 

> Isn't the .not_nil! sufficient by itself?

Compare the output:

## with not_nil!

```
1) foo bar

       Nil assertion failed (NilAssertionError)
         from /usr/local/Cellar/crystal/0.31.0_1/src/nil.cr:106:5 in 'not_nil!'
         from foo.cr:6:9 in '->'
         ...
```

## with should_not be_nil

```
  1) foo bar
     Failure/Error: x = a.should_not be_nil

       Expected: nil not to be nil

     # foo.cr:6
```

I think `should_not be_nil` has a much better output. It also points you to the exact location, and it shows you the line that failed the assertion.

asterite:
@Fryguy 

> I actually prefer the expect syntax, personally, and it avoids having to patch methods. @asterite would you be ok with that being added as an alternate spec syntax?

No, we already had this discussion in the past and we kind of agreed that `should` involves less typing and it's easier to read and write.

src/spec/expectations.cr
```diff
@@ -50,109 +115,139 @@ module Spec
...
       "Expected: #{actual_value.inspect} to be nil"
     end
 
-    def negative_failure_message(actual_value)
+    def negative_failure_message(actual_value) : String
       "Expected: #{actual_value.inspect} not to be nil"
     end
+
+    def cast_should(value)
+      value.nil? ? value : (raise "expected #{value} to be nil")
```
RX14: These should be `BUG: `right?

asterite: Well, you can always instantiate this and call the method and you'll get an exception. I use BUG for compiler bugs.

asterite:
I pushed another commit to make it backwards-compatible with a deprecation message, so now this should be fine to merge.

We can change `not_nil!` in existing specs to use `should_not be_nil` in subsequent PRs: there are a lot of them and they are not necessary to merge this PR.

asterite:
I also want to rename `Spec::Expectation` to `Spec::Matcher`, but again, in a separate PR.

straight-shoota:
I might actually prefer implementing the special cases for `BeNil` and `BeA` directly in `should`/`should_not`. This would avoid a lot of overhead introduced by `cast_should` and `cast_should_not` and be easy and simple to understand.

I don't think we need to care for extensibility here, because I honestly can't think of any other valid use case for a generic solution. And if there was, it could always be changed later without breaking anything.

But we can reduce complexity by going the simple way.

Now, adding `Spec::Expectation` or `Spec::Matcher` might be a useful addition on its own. But this could be discussed separately.

asterite:
I agree, I'll send a simpler PR later.

straight-shoota:
@asteite Is this still on your radar? Otherwise we should make an issue to track this.

asterite:
Yes, please make an issue. I'm in the middle of a move and I don't have time for Crystal other than just replying on commenting on issues. Thank you!

RX14:
I'm fine with merging this PR without simplifying. It's not a big deal one way or the other, and can be changed in the future.

asterite:
Thanks for reminding me about this, I'll send a PR with the simplified version tomorrow.

PR #5671, opened 2018-02-01
[WIP] Add specs for spec output

straight-shoota:
This a work in progress trying to add specs for spec output.

Until know, there are no specs for this at all. Unfortunately the spec implementation is not very modular, so I had to mock it quite extensively to be able to run it in a spec context. But it work and required only a minimal change in the actual implementation.

It's far from ready but I'd like to get some feedback on this approach.

This could be useful for implementing #5633

PR #8242, opened 2019-09-27
Add specs for Spec filters

Fryguy:
@asterite Please review.

This adds basic specs for most of the filters as extracted in #8125.

### DIFF
--- a/spec/std/spec/filters_spec.cr
+++ b/spec/std/spec/filters_spec.cr
@@ -1,0 +1,115 @@
+require "./spec_helper"
+
+module Spec::Item
+  setter focus : Bool
+end
+
+describe Spec::RootContext do
+  describe "#run_filters" do
+    describe "by pattern" do
+      it "on an example" do
+        root = build_spec("f.cr")
+        root.run_filters(pattern: /example_f_2_2/)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_2]
+      end
+
+      it "on a context" do
+        root = build_spec("f.cr")
+        root.run_filters(pattern: /context_f_2/)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_1 example_f_2_2]
+      end
+    end
+
+    describe "by line" do
+      it "on a context's start line'" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 11)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_1 example_f_2_2]
+      end
+
+      it "between examples" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 15)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_1 example_f_2_2]
+      end
+
+      it "on an example's start line" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 16)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_2]
+      end
+
+      it "in an example" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 17)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_2]
+      end
+
+      it "on an example's end line" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 18)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_2]
+      end
+
+      it "on a context's end line'" do
+        root = build_spec("f.cr")
+        root.run_filters(line: 19)
+        all_spec_descriptions(root).should eq %w[root context_f_2 example_f_2_1 example_f_2_2]
+      end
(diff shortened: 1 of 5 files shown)


spec/std/spec/filters_spec.cr
```diff
@@ -0,0 +1,141 @@
...
+    "root"
+  end
+end
+
+private def all_descriptions(item) : Array(String)
+  child_descriptions = item.responds_to?(:children) ? item.children.flat_map { |c| all_descriptions(c) } : [] of String
+  child_descriptions.unshift(item.description)
+end
+
+private def build_spec(filename, root = nil, count = 2)
```
Fryguy: I was trying to find a way to parse a heredoc of an example file, but even if I did, I wouldn't know how to tell it to use the alternate `FakeRootContext` instead if `Spec.root_context`.  This approach here is similar to what's done in the other specs, so I thought it was good enough.

---

This method produces the equivalent of a spec file named `f.cr` with:

```crystal
context "context_f_1" do
  it "example_f_1_1" do
    # noop
  end

  it "example_f_1_2" do
    # noop
  end
end

context "context_f_2" do
  it "example_f_2_1" do
    # noop
  end

  it "example_f_2_2" do
    # noop
  end
end
```

asterite:
Given the massive breakage that this caused, I wonder if we should simply revert this and keep the old, simpler spec. It's also more flexible. But no focus, no rand test and no parallelism in the future. Maybe it's fine. What do you think?

ysbaddaden:
@asterite don't revert, the benefits (filters, random and parallel runs) clearly trump the breaking change, which is merely to not trust mutable global/shared data.

The breaking change could have been more explicit in the changelog, thought.

bararchy:
@asterite I took me 3 days to fix everything, please don't revert XD

Fryguy:
I think the change that caused the real issue was using `at_exit` to run the specs.  Perhaps we can tweak that part, but the rest of these changes have been really useful.  I think these features are very beneficial overall.

Total aside, I have a local branch with random specs, as well as seeded-random specs.  It's ready to go, just wanted to get these specs + tags in first, so I'm not building PRs on top of PRs.

Fryguy:
I'm not sure if there's something more I need to do for this PR...the 32 bit tests are failing, but I'm not sure if it's a general issue or something to do with my code here.

ysbaddaden:
Nothing but a review from someone with spec internals knowledge.

asterite:
I'll be busy these weeks so at least a review from me well have to wait, sorry.

Fryguy:
@straight-shoota Thanks for the pointer to #5671 ... I'll see what I can use from there.

src/spec/dsl.cr
```diff
@@ -230,7 +230,7 @@ module Spec
     start_time = Time.monotonic
 
     at_exit do
-      run_filters
+      root_context.run_filters(@@pattern, @@line, @@locations, @@split_filter, @@focus)
```
Sija: I'd leave this line within the local method `run_filters`.

spec/std/spec/filters_spec.cr
```diff
@@ -0,0 +1,141 @@
...
+  root ||= FakeRootContext.new
+  name = filename.chomp(".cr")
+
+  1.upto(count) do |i|
```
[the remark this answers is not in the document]
Fryguy: @straight-shoota I left this as original because `1.upto(count)` gives `1, 2, 3` whereas `count.times` gives `0, 1, 2`.  Aside from the lines, the value is also used as part of the context and example naming, and that ends up switching them all to 0 based, which I found harder to read.

src/spec/dsl.cr
Fryguy: You mean keep the local method that just does this one-liner?  I can do that.  I had figured the extra method wasn't worth it considering there are other calls to root_context (i.e. `root_context.run`) just below it and having them together in visual context made the code clearer to me.

Fryguy:
@straight-shoota I've reviewed #5671 and I think for now I'd like to keep this PR as is and refactor to something like that later.

---

The SpecEnvironment there is a copy of a lot of stuff, and I'd like to avoid the duplication, but the concept it really good.  Looking at Spec itself, the big issue with not being able to test is that it uses a lot of class variables which are in turn called directly from the various methods.  Detangling that I think is the key.  I've started a branch locally where instead of using the Spec module directly with class variables, we instead have a Spec::Environment class (or perhaps part of RootContext itself).  We can still keep the top-level Spec DSL methods for convenience, but they defer to an instance of RootContext with its own environment.  Then, in specs, we can create a new RootContext, with a separate environment and it won't collide with the singleton environment.  cc @asterite

Fryguy:
Moved the helper methods into a spec_helper.cr similar to #8310 .  This will also make it easier to rebase those PRs later.

RX14:
Needs a rebase.

Fryguy:
Done.

PR #8251, opened 2019-09-30
Update CI to use 0.31.1

bcardiff:


wontruefree:
This is changed every release.  should this be configured in the CI itself?

bcardiff:
@wontruefree it's a bit of bookkeeping but it also serves as a log of the lineage. If they end up as a configuration in the CI then that information is lost in time and forks will need to be configured also.

bcardiff:
The 32 bits packages are failing more frequently. It will need some further investigation in another PR. For now, the specs will be allowed to fail in order to avoid delaying the whole build. But the dist_linux32 job might also fail from time to time. So there is a bit more pressure to fix the memory issues.

Issue #7795, opened 2019-05-19
Time::Span constructor with Int32 for days, minutes, seconds


jaredmoody:
I'm new to crystal so forgive me if I'm misunderstanding, but it seems to me the `Time::Span` constructor is not behaving correctly:

```
Time::Span.new(seconds: 1)
# => no overload matches 'Time::Span.new', seconds: Int32
#  Overloads are:
#  - Time::Span.new(days : Int, hours : Int, minutes : Int, seconds : Int, nanoseconds : Int = 0)
#  ...
```

Passing `1` for `days`, `minutes`, `seconds`, all give the same error, but for `nanoseconds` it works:

```
Time::Span.new(nanoseconds: 1)
# => 00:00:00.000000001 : Time::Span
```

It seems to me all of these _should_ work, but if not then it seems like `nanoseconds` is at least inconsistently accepting `1` as an `Int` where the other arguments are not.

Here's my crystal version info:

```
> crystal -v
LLVM: 6.0.1
Default target: x86_64-apple-macosx

**Comment by cockroachking** [comment]
It's because there is no an overload in which you can only pass a `seconds` argument. In your case you have to either pass mandatory `hours`, `minutes` and `seconds` arguments like this:
```crystal
Time::Span.new hours: 0, minutes: 0, seconds: 1
```
or `seconds` and `nanoseconds`
```crystal
Time::Span.new seconds: 1, nanoseconds: 0  # Note that for this overload `seconds` and `nanoseconds` have to be named arguments
```
Or even you can do
```crystal
1.seconds
```
which reads much better.

Check the API reference for more details

straight-shoota:
I think the API should have allow to omit any component, defaulting all to zero.

So you can do `Time.new(minutes: 1, seconds: 30)` without having to specify `days: 0, hours: 0` or compose the value as `1.minute + 30.seconds`.

vladfaust:
Also https://github.com/crystal-lang/crystal/issues/7143

jaredmoody:
Thanks for the explanation, I'll close this as it's not a bug.

However, the compiler error could certainly be more helpful. "No overload match" made me think a type was incorrect; a missing argument error would have led me to the answer.

straight-shoota:
Not a bug, but I see it as a feature request.

jmbarnardgh:
[...]

PR #8257, opened 2019-10-02
Improve Time::Span initialization API

dnamsons:
Resolves https://github.com/crystal-lang/crystal/issues/7795

I also refactored the `self.compute_seconds` as it now can safely rely on `OverflowError`, which was introduced in 0.31

### DIFF
--- a/src/time/span.cr
+++ b/src/time/span.cr
@@ -5,9 +5,9 @@
 # Check all `#new` methods for details.
 #
 # ```
-# Time::Span.new(nanoseconds: 10_000) # => 00:00:00.000010000
-# Time::Span.new(10, 10, 10)          # => 10:10:10
-# Time::Span.new(10, 10, 10, 10)      # => 10.10:10:10
+# Time::Span.new(nanoseconds: 10_000)                           # => 00:00:00.000010000
+# Time::Span.new(hours: 10, minutes: 10, seconds: 10)           # => 10:10:10
+# Time::Span.new(days: 10, hours: 10, minutes: 10, seconds: 10) # => 10.10:10:10
 # ```
 #
 # Calculation between `Time` also returns a `Time::Span`.
@@ -21,7 +21,7 @@
 # Inspection:
 #
 # ```
-# span = Time::Span.new(20, 10, 10)
+# span = Time::Span.new(hours: 20, minutes: 10, seconds: 10)
 # span.hours   # => 20
 # span.minutes # => 10
 # span.seconds # => 10
@@ -30,8 +30,8 @@
 # Calculation:
 #
 # ```
-# a = Time::Span.new(20, 10, 10)
-# b = Time::Span.new(10, 10, 10)
+# a = Time::Span.new(hours: 20, minutes: 10, seconds: 10)
+# b = Time::Span.new(hours: 10, minutes: 10, seconds: 10)
 # c = a - b # => 10:00:00
 # c.hours   # => 10
 # ```
@@ -53,14 +53,16 @@ struct Time::Span
   # @nanoseconds can either be negative or positive).
   @nanoseconds : Int32
 
-  def self.new(hours : Int, minutes : Int, seconds : Int)
-    new(0, hours, minutes, seconds)
+  @[Deprecated("Use `new` with named arguments instead.")]
+  def self.new(_hours : Int, _minutes : Int, _seconds : Int)
+    new(0, _hours, _minutes, _seconds)
   end
 
-  def self.new(days : Int, hours : Int, minutes : Int, seconds : Int, nanoseconds : Int = 0)
+  @[Deprecated("Use `new` with named arguments instead.")]
+  def self.new(_days : Int, _hours : Int, _minutes : Int, _seconds : Int, _nanoseconds : Int = 0)
     new(
-      seconds: compute_seconds!(days, hours, minutes, seconds),
-      nanoseconds: nanoseconds.to_i64,
+      seconds: compute_seconds!(_days, _hours, _minutes, _seconds),
+      nanoseconds: _nanoseconds.to_i64,
     )
   end
 
@@ -88,6 +90,13 @@ struct Time::Span
     new(
       seconds: nanoseconds.to_i64.tdiv(NANOSECONDS_PER_SECOND),
       nanoseconds: nanoseconds.to_i64.remainder(NANOSECONDS_PER_SECOND),
+    )
+  end
+
+  def self.new(*, days : Int = 0, hours : Int = 0, minutes : Int = 0, seconds : Int = 0, nanoseconds : Int = 0)
+    new(
+      seconds: compute_seconds!(days, hours, minutes, seconds),
+      nanoseconds: nanoseconds.to_i64,
     )
   end
 
@@ -398,7 +407,7 @@ end
 struct Int
   # Returns a `Time::Span` of `self` weeks.
   def weeks : Time::Span
-    Time::Span.new 7 * self, 0, 0, 0
+    Time::Span.new(days: (7 * self))
   end
 
   # :ditto:
@@ -408,7 +417,7 @@ struct Int
 
   # Returns a `Time::Span` of `self` days.
   def days : Time::Span
-    Time::Span.new self, 0, 0, 0
+    Time::Span.new(days: self)
   end
 
   # :ditto:
@@ -418,7 +427,7 @@ struct Int
 
   # Returns a `Time::Span` of `self` hours.
   def hours : Time::Span
-    Time::Span.new self, 0, 0
+    Time::Span.new(hours: self)
   end
 
   # :ditto:
@@ -428,7 +437,7 @@ struct Int
 
   # Returns a `Time::Span` of `self` minutes.
   def minutes : Time::Span
-    Time::Span.new 0, self, 0
+    Time::Span.new(minutes: self)
   end
 
   # :ditto:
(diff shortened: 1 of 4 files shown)


src/time/span.cr
```diff
@@ -84,70 +87,14 @@ struct Time::Span
...
-      nanoseconds: nanoseconds.to_i64.remainder(NANOSECONDS_PER_SECOND),
-    )
-  end
-
-  private def self.compute_seconds!(days, hours, minutes, seconds)
-    compute_seconds(days, hours, minutes, seconds, true).not_nil!
-  end
-
-  private def self.compute_seconds(days, hours, minutes, seconds, raise_exception)
-    # TODO once overflow is the default this can be refactored
```
dnamsons: I decided to refactor this method upon seeing this TODO - I remembered that overflowing as a default had been added just recently

src/time/span.cr
```diff
@@ -53,18 +53,21 @@ struct Time::Span
   # @nanoseconds can either be negative or positive).
   @nanoseconds : Int32
 
-  def self.new(hours : Int, minutes : Int, seconds : Int)
-    new(0, hours, minutes, seconds)
+  def self.new(days : Int, hours : Int, minutes : Int, seconds : Int, nanoseconds : Int)
```
straight-shoota: Do we even want to keep this overload? Why not require named arguments always?

src/time/span.cr
```diff
@@ -53,18 +53,21 @@ struct Time::Span
...
     new(
-      seconds: compute_seconds!(days, hours, minutes, seconds),
-      nanoseconds: nanoseconds.to_i64,
+      compute_seconds(days, hours, minutes, seconds),
+      nanoseconds.to_i64,
     )
   end
 
-  def initialize(*, seconds : Int, nanoseconds : Int)
+  def initialize(seconds : Int, nanoseconds : Int)
```
straight-shoota: Please revert this. There's no reason to make these arguments positional.

src/time/span.cr
```diff
@@ -84,70 +87,14 @@ struct Time::Span
     @nanoseconds = nanoseconds.to_i32
   end
 
-  def self.new(*, nanoseconds : Int)
```
straight-shoota: This overload must not be removed. There should be a constructor that allows passing nanoseconds with overflow.

src/time/span.cr
dnamsons: I was slightly worried about backwards compatability, but if this seems unnecessary, I will remove it

straight-shoota: We're breaking backwards compatibility anyway. But it might make sense to leave it for now as deprecated.

src/time/span.cr
dnamsons: I am quite unsure how to deal with overloads, as having named arguments here leads to stack overflow

dnamsons: Never mind, I just realized i misunderstood the docs 😓

dnamsons:
I currently have kept the old initialization methods.
Is desirable to keep them, and if so, do I have to do something about adding the `deprecated` status?

src/time/span.cr
```diff
@@ -59,7 +59,7 @@ struct Time::Span
 
   def self.new(days : Int, hours : Int, minutes : Int, seconds : Int, nanoseconds : Int = 0)
```
dnamsons: I currently have kept the old initialization methods.
Is desirable to keep them, and if so, do I have to do something about adding the deprecated status?

straight-shoota: This overload should probably be removed because it shadows the new one. Or maybe just prefix every argument with `_` and add `@[Deprecated]` annotation.

straight-shoota: This overload should probably be removed because it shadows the new one. Or maybe just prefix every argument with `_` and add `@[Deprecated]` annotation.

dnamsons: I added `@[Deprecated]` annotations with a message to use `#new` with named args instead

src/time/span.cr
```diff
@@ -53,13 +53,15 @@ struct Time::Span
   # @nanoseconds can either be negative or positive).
   @nanoseconds : Int32
 
-  def self.new(hours : Int, minutes : Int, seconds : Int)
+  @[Deprecated("Use `new` with named arguments instead.")]
+  def self.new(_hours : Int, _minutes : Int, _seconds : Int)
     new(0, hours, minutes, seconds)
```
Sija: This ain't compile anymoar.

src/time/span.cr
```diff
@@ -53,13 +53,15 @@ struct Time::Span
...
   end
 
-  def self.new(days : Int, hours : Int, minutes : Int, seconds : Int, nanoseconds : Int = 0)
+  @[Deprecated("Use `new` with named arguments instead.")]
+  def self.new(_days : Int, _hours : Int, _minutes : Int, _seconds : Int, _nanoseconds : Int = 0)
     new(
-      seconds: compute_seconds!(days, hours, minutes, seconds),
+      seconds: compute_seconds(days, hours, minutes, seconds),
       nanoseconds: nanoseconds.to_i64,
```
Sija: ditto

src/time/span.cr
straight-shoota: What do you mean?

jhass: `_hours` vs `hours` etc

src/time/span.cr
```diff
@@ -91,63 +93,21 @@ struct Time::Span
...
 
-    if overflow
-      if raise_exception
-        raise ArgumentError.new "Time::Span too big or too small"
-      end
-      return nil
-    end
+  private def self.compute_seconds(days, hours, minutes, seconds)
+    days_to_seconds = SECONDS_PER_DAY.to_i64 * days
+    hours_to_seconds = 3600_i64 * hours
```
straight-shoota: These values should either all use the respective constants or all be literals.

src/time/span.cr
bcardiff: We are still missing to fix this `_arg` vs `arg`.
It's fine to leave a at least a typeof spec to ensure it compiles. Or even that it still behaves as wanted. Warnings are allowed in the CI.

src/time/span.cr
bcardiff: ditto ditto

src/time/span.cr
bcardiff:
```suggestion
    hours_to_seconds = SECONDS_PER_HOUR.to_i64 * hours
```

src/time/span.cr
```diff
@@ -91,63 +93,21 @@ struct Time::Span
...
+    minutes_to_seconds = 60_i64 * minutes
```
bcardiff:
```suggestion
    minutes_to_seconds = SECONDS_PER_MINUTE.to_i64 * minutes
```

straight-shoota:
@bcardiff We could extract the `compute_seconds` commit from this PR and discuss the overflow issues separately.
The rest can be merged quickly once the outstanding issues are fixed.

RX14:
All passing `-Ddisable_overflow` would do after this PR is disable overflow checking. I don't think that's unacceptable, since that's what you sign up for with that flag.

dnamsons:
With `-Ddisable_overflow` enabled, `compute_seconds` will not raise on overflow, so I guess that should be removed from this PR

src/time/span.cr
```diff
@@ -66,6 +66,15 @@ struct Time::Span
     )
   end
 
+  # Creates a new `Time::Span` from the *seconds* and *nanoseconds* given
```
straight-shoota:
```suggestion
  # Creates a new `Time::Span` from *seconds* and *nanoseconds*.
```

src/time/span.cr
```diff
@@ -66,6 +66,15 @@ struct Time::Span
...
+  #
+  # Nanoseconds get normalized in the range of 0...1_000_000_000,
```
straight-shoota:
```suggestion
  # Nanoseconds get normalized in the range of `0...1_000_000_000`,
```

src/time/span.cr
straight-shoota: Full stops everywhere, please.

src/time/span.cr
```diff
@@ -84,70 +95,46 @@ struct Time::Span
     @nanoseconds = nanoseconds.to_i32
   end
 
+  # Creates a new `Time::Span` from the *nanosenconds* given
+  #
+  # Nanoseconds get normalized in the range of 0...1_000_000_000,
```
straight-shoota:
```suggestion
  # Nanoseconds get normalized in the range of `0...1_000_000_000`,
```

src/time/span.cr
```diff
@@ -84,70 +95,46 @@ struct Time::Span
...
-      else
-        nanos = s
-        s += sd
-        # negative days -> total nanos should be lower
-        overflow = s > nanos
-      end
-    end
+  # Creates a new `Time::Span` from the *days*, *hours*, *minutes*, *seconds* and *nanoseconds* given
+  #
+  # Any time unit can be ommited
```
straight-shoota:
```suggestion
  # Any time unit can be ommited.
```

src/time/span.cr
dnamsons: Thanks, sorry for wasting your time 😓

straight-shoota: Don't worry. Thanks to you for your contribution!

straight-shoota:
Please rebase on master to fix the CI failures.

RX14:
It won't create invalid `Time::Span` values, just incorrect ones. This is no different to any other integer in the stdlib, and i cannot understand the deal with why this specific piece of maths is anything special.

dnamsons:
Rebased to the latest `master`

straight-shoota:
Rebased on current master, otherwise this seems ready to merge.

straight-shoota:
Thanks @dnamsons

PR #8260, opened 2019-10-02
Remove dupilcated method definitions  

asterite:


ysbaddaden:
Those macros for example?

```crystal
class User
  DB.mapping({ name: {type: Int32} })   # => def name
  JSON.mapping({ name: {type: Int32} }) # => def name
end
```

asterite:
Yes, those macros. But I'd like to remove them in favor of attributes and JSON::Serializable. There's no point in having two ways to do the same thing, and with attributes it's more extensible and doesn't inject methods (just a constructor).

asterite:
That is, I'd like to deprecate and eventually remove those mapping macros and use the attributes style.

asterite:
As an alternative I can try to not take methods defined via macros into account.

straight-shoota:
I like this!

But I'd like to question if methods using `previous_def` should be exempted:
```cr
class Foo
  def foo
    1
  end

  def foo
    previous_def
  end
end
```
I'm not sure about this, though. Using `previous_def` signals that the developer is aware of a previous definition and there might be reasons for this (although I can't think of one ATM). OTOH, `previous_def` might reference the definition from a different scope, in that case we'd like the above code to be a warning/error nonetheless.

asterite:
Very good question!

I think @bcardiff has a pattern of enhancing a single def by redefining it and using `previous_def`. In that case my PR will break his code.

Now I think this is not a good thing as a general rule, but maybe as a lint it's fine, at least to spot the (minor) mistakes that exist in the standard library.

bcardiff:
Definitely a good catch. Another one might be that due to compile flags or monkey patching raising a compiler error might be too much.

I guess this is better for ameba then or for a future `crystal doctor`.

And as with all the warnings it would be good to have a way for toggling off in a specific lexical scope (a thing that we don't do currently).

Let's move the cleanup commits to a separate refactor PR.
Open an issue in ameba linking to this PR in case this can be handled there.

asterite:
Sounds good. I'll send a separate PR later. Actually, I'll just force push excluding the last commit and change the title here.

bcardiff:
Better separate so the description matches and this intent history is kept for reference.

asterite:
Done!

asterite:
#8270

PR #8264, opened 2019-10-03
Add `offset` argument to all `map_with_index` methods

asterite:
Since `each_with_index` accepts an optional `offset` argument I think it makes sense for `map_with_index` to accept one too.

Ref: https://forum.crystal-lang.org/t/difference-between-array-elements/1189

### DIFF
--- a/spec/std/pointer_spec.cr
+++ b/spec/std/pointer_spec.cr
@@ -231,6 +231,14 @@ describe "Pointer" do
     a[2].should eq(5)
   end
 
+  it "maps_with_index!, with offset" do
+    a = Pointer(Int32).malloc(3) { |i| i + 1 }
+    a.map_with_index!(3, offset: 10) { |e, i| e + i }
+    a[0].should eq(11)
+    a[1].should eq(13)
+    a[2].should eq(15)
+  end
+
   it "raises if mallocs negative size" do
     expect_raises(ArgumentError) { Pointer.malloc(-1, 0) }
   end
--- a/spec/std/static_array_spec.cr
+++ b/spec/std/static_array_spec.cr
@@ -129,12 +129,27 @@ describe "StaticArray" do
     b.should eq(StaticArray[1, 2, 4, 5])
   end
 
+  it "does map_with_index, with offset" do
+    a = StaticArray[1, 1, 2, 2]
+    b = a.map_with_index(10) { |e, i| e + i }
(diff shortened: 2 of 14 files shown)


spec/std/pointer_spec.cr
```diff
@@ -231,6 +231,14 @@ describe "Pointer" do
     a[2].should eq(5)
   end
 
+  it "maps_with_index!, with offset" do
+    a = Pointer(Int32).malloc(3) { |i| i + 1 }
+    a.map_with_index!(3, offset: 10) { |e, i| e + i }
+    a[0].should eq(11)
```
wontruefree: you could probably do what you did above in https://github.com/crystal-lang/crystal/pull/8264/files#diff-fbe3f630f4ca2da753ea76cf3a0b95baR1594

```suggestion
    a.should eq([11, 13, 15])
```

spec/std/static_array_spec.cr
```diff
@@ -138,6 +144,15 @@ describe "StaticArray" do
     a.should be_a(StaticArray(Int32, 3))
   end
 
+  it "does map_with_index!, with offset" do
+    a = StaticArray(Int32, 3).new { |i| i + 1 }
+    a.map_with_index!(10) { |e, i| i * 2 }
+    a[0].should eq(20)
```
wontruefree: same as above

src/array.cr
```diff
@@ -1058,13 +1058,19 @@ class Array(T)
   end
 
   # Optimized version of `Enumerable#map_with_index`.
-  def map_with_index(&block : T, Int32 -> U) forall U
-    Array(U).new(size) { |i| yield @buffer[i], i }
+  #
+  # Accepts an optional *offset* parameter, which tells it to start counting
+  # from there.
+  def map_with_index(offset = 0, &block : T, Int32 -> U) forall U
+    Array(U).new(size) { |i| yield @buffer[i], offset + i }
```
wontruefree: I dont really know the difference here but why not use `build`

```suggestion
    Array(U).build(size) { |i| yield @buffer[i], offset + i }
```

straight-shoota: Please refer to the documentation of [Array.build](https://crystal-lang.org/api/0.31.1/Array.html#build(capacity:Int,&block):self-class-method) vs. [Array.new](https://crystal-lang.org/api/0.31.1/Array.html#new(size:Int,&block:Int32-%3ET)-class-method).

PR #8270, opened 2019-10-04
Remove duplicated defs

asterite:
Extracted from #8260

### DIFF
--- a/src/io/byte_format.cr
+++ b/src/io/byte_format.cr
@@ -43,8 +43,6 @@ module IO::ByteFormat
   abstract def encode(int : UInt64, io : IO)
   abstract def encode(int : Int128, io : IO)
   abstract def encode(int : UInt128, io : IO)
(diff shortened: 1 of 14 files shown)


src/io/byte_format.cr
```diff
@@ -43,8 +43,6 @@ module IO::ByteFormat
   abstract def encode(int : UInt64, io : IO)
   abstract def encode(int : Int128, io : IO)
   abstract def encode(int : UInt128, io : IO)
-  abstract def encode(int : Float32, io : IO)
-  abstract def encode(int : Float64, io : IO)
```
RX14: Why are these useless?

asterite: Because they are defined as non-abstract a few lines below. It doesn't make sense to define a method as abstract and as non-abstract in the same type.

RX14: Whoops, I'm blind :)

Issue #8261, opened 2019-10-02
export TERM=dumb


ilanpillemer:
When setting `TERM=dumb` I should not get funny stuff.

```
hello-world -> export TERM=dumb
hello-world -> crystal spec
[32m.[0m

Finished in 119 microseconds
[32m1 examples, 0 failures, 0 errors, 0 pending[0m
hello-world -> 
```

But I do...

straight-shoota:
Sounds like a good idea to disable color output by default when `ENV["TERM"] == "dumb"`.
This should be implemented in both `spec.cr` and `compiler/command.cr`, and `colorize.cr` could honour this setting as well.

asterite:
Out of curiosity: if you do `TERM=dumb rspec` with a Ruby project, what happens?

(I tried it on iTerm2 and Terminal on Mac and nothing changes)

Is `TERM=dumb` something that all programs must take into account? Also, why are you using `TERM=dumb`? Finally, you can pass `--no-color` to remove color.

ilanpillemer:
1. I am happy to try work out how to contribute and send a pull request for such a change
2. I have never used Ruby so cannot comment on whether it honours TERM in the terminal.
3. I am using the `acme` editor which does not work with colors, and I set TERM in general to solve this problem. See http://acme.cat-v.org/ , https://ilanpillemer.github.io/subjects/acme.html
4. I have just started using it together with `acme-lsp` (I need to update my page above with how I use it) but I have not yet tried to integrate with a Crystal lsp server.
5. I can use `--no-color` buts its clunky and means I have to treat Crystal as special snowflake. This is what I am doing now, after I figured out you could.
5. Here are some sample links where `TERM=dumb` is mentioned in which program are meant to honour the environment variable. 

- https://news.ycombinator.com/item?id=4533156 (scroll down to Russ Cox's comment on this`
- https://github.com/dexen/plan9port-custom/commit/51c437db8733264ef978216efeb813987071131c
- https://hn.nuxtjs.org/item/17026490
- https://github.com/rust-lang/rust/issues/50645

- This stack overflow page mentioned Ruby colorizer and sayts it honours `TERM=dumb`

asterite:
I guess we could read Term ENV var when booting "colorize" and always disable it if it's dumb. Spec needs a special treatment too because it uses "\r", which I guess dumb doesn't support?

ilanpillemer:
`dumb|80-column dumb tty,
        am,
        cols#80,
[...]

PR #8271, opened 2019-10-04
export TERM=dumb #8261

ilanpillemer:
This is a first attempt at #8261

### DIFF
--- a/src/colorize.cr
+++ b/src/colorize.cr
@@ -124,9 +124,12 @@ module Colorize
   # ```
   class_property? enabled : Bool = true
 
-  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?` and `STDERR.tty?` are `true`.
+  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?`
+  # and `STDERR.tty?` are `true` and the tty is not considered a dumb terminal.
+  # This is determined by the environment variable called `TERM`.
+  # If `TERM=dumb`, color won't be enabled.
   def self.on_tty_only!
-    self.enabled = STDOUT.tty? && STDERR.tty?
+    self.enabled = STDOUT.tty? && STDERR.tty? && ENV["TERM"]? != "dumb"
   end
 
   def self.reset(io = STDOUT)
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -49,7 +49,7 @@ class Crystal::Command
   private getter options
 
   def initialize(@options : Array(String))
-    @color = true
+    @color = ENV["TERM"]? != "dumb"
     @error_trace = false
     @progress_tracker = ProgressTracker.new
   end
@@ -563,7 +563,7 @@ class Crystal::Command
 
   private def error(msg, exit_code = 1)
     # This is for the case where the main command is wrong
-    @color = false if ARGV.includes?("--no-color")
+    @color = false if ARGV.includes?("--no-color") || ENV["TERM"]? == "dumb"
     Crystal.error msg, @color, exit_code: exit_code
   end
 end
--- a/src/spec.cr
+++ b/src/spec.cr
@@ -78,6 +78,8 @@ require "./spec/dsl"
 module Spec
 end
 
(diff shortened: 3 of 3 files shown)


bcardiff:
As long as we don't need to force the output to an 80 cols format this looks good enough.

src/colorize.cr
```diff
@@ -125,8 +125,9 @@ module Colorize
   class_property? enabled : Bool = true
 
   # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?` and `STDERR.tty?` are `true`.
+  # And the terminal is not a dumb terminal that will not understand escape sequences.
```
RX14: This doesn't make sense as a standalone sentence.

ilanpillemer: I have changed `Spec` to use `on_tty_only!`

src/compiler/crystal/command.cr
```diff
@@ -563,7 +564,7 @@ class Crystal::Command
 
   private def error(msg, exit_code = 1)
     # This is for the case where the main command is wrong
-    @color = false if ARGV.includes?("--no-color")
+    @color = false if ARGV.includes?("--no-color") || ENV["TERM"]? == "dumb"
```
Sija: This line IMO should be in general refactored due to a hacky way of (re)setting `@color` to a proper value - which should be done in line 524 already.

ilanpillemer: Will line 524 run "in the case where the main command is wrong?".. If so the line could be deleted. But if not it can't. If it would run, why was this line added with this comment?

ilanpillemer: Will line 524 run "in the case where the main command is wrong?".. If so the line could be deleted. But if not it can't. If it would run, why was this line added with this comment?

src/spec.cr
```diff
@@ -78,6 +78,8 @@ require "./spec/dsl"
 module Spec
 end
 
+Spec.use_colors = Colorize.on_tty_only!
```
RX14: In fact, I think `Spec.use_colors` can be removed, since `Colorize.enabled` exists. Just call `Colorize.on_tty_only!` once in `spec.cr` and the rest should sort itself out.

ilanpillemer: done

src/colorize.cr
```diff
@@ -124,9 +124,12 @@ module Colorize
   # ```
   class_property? enabled : Bool = true
 
-  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?` and `STDERR.tty?` are `true`.
+  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?`
+  # and `STDERR.tty?` are `true`.  It also checks if the TERM environment
```
straight-shoota:
```suggestion
  # and `STDERR.tty?` are `true`. It also checks the `TERM` environment variable.
  # If `TERM=dumb`, color won't be enabled either. 
```

src/compiler/crystal/command.cr
```diff
@@ -50,6 +50,7 @@ class Crystal::Command
 
   def initialize(@options : Array(String))
     @color = true
+    @color = false if ENV["TERM"]? == "dumb"
```
straight-shoota:
```suggestion
    @color = ENV["TERM"]? != "dumb"
```
And scrap the previous line.

src/colorize.cr
```diff
@@ -124,9 +124,11 @@ module Colorize
   # ```
   class_property? enabled : Bool = true
 
-  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?` and `STDERR.tty?` are `true`.
+  # Makes `Colorize.enabled` `true` if and only if both of `STDOUT.tty?`
+  # and `STDERR.tty?` are `true`.
+  # If `TERM=dumb`, color won't be enabled either.
```
straight-shoota: This needs to explain that `TERM` is an environment variable. You removed that part from my suggestion.

ilanpillemer: when I accepted the suggestion the comments got all mangled, so I guessed what was meant to be there.

Issue #3997, opened 2017-02-03
GC doesn't seem aggressive enough at times


rdp:
See if this one seems to use up an inordinate amount of RAM for me (it seems to for me on all OS's I tried anyway):

```
class Foo
  def finalize
    # Invoked when Foo is garbage-collected
    a = 3
    "bye #{3+4}"
  end

end

# Prints "Bye bye ...!" for ever
loop do
   Foo.new
end
```

I wonder if it's related to boehm gc finalization cycle or something.  Let me know if reproducible.  Thank you.

waj:
Objects with finalizers are collected after two GC cycles. The first cycle will run the finalizer and the second one will collect the object memory. If you add a call to `GC.collect` inside the loop the memory usage will stay really low.

For this reason I think it's a bad idea to allocate so many unreachable objects with finalizers in a tight loop. In fact, I think finalizers should be avoided whenever possible. They are useful to release OS resources but otherwise there might be always another (more deterministic) option to release resources or finish some work.

Can you give more details about what are you trying to accomplish with finalizers?

rdp:
It does stay down with GC.collect nice.   That exception trace still occurs occasionally (at least on OS X). I'm trying to debug here why OpenSSL server on "some machines" seems to leak memory.  Its sockets do use finalizers FWIW.  https://github.com/crystal-lang/crystal/issues/3983

benoist:
What I'm seeing in regards to GC is that the amount of heap memory seems to keep growing, but the actual memory usage is mostly lower see #3992
If I print GC stats I get this result:

```code
GC::Stats(@heap_size=3233890304, @free_bytes=3233832960, @unmapped_bytes=0, @bytes_since_gc=0, @total_bytes=8000019728)
```
As you can see in my example the free bytes is almost the same as the heap size, maybe that's the case here as well?

rdp:
The good news here is that if I add in an explicit GC.malloc every so often:
```

spawn do
  loop do
    sleep 5
    GC.collect
  end
end
```

It seems RES memory is "constrained" to around 1GB.  Or if it's `sleep ` then about 100 MB.  So the GC feels like it's working...not allowing leaks...just not being called somehow?

RX14:
@rdp GC.collect is expensive, ram not being used by other programs is almost free. For that reason, it's likely that the GC will instead simply use more ram instead of reclaiming more often under heavy allocation pressure (i.e. allocating in a loop).

[...]

Issue #2104, opened 2016-02-05
GC Warning: Repeated allocation of very large block


stugol:
I'm getting this warning sometimes when running my code:
```
GC Warning: Repeated allocation of very large block (appr. size 8392704):
	May lead to memory leak and poor performance.
```
I understand the error, but I don't know where in my code the allocation is happening. How can I track this down?

The size of the block changes each time, incidentally. Which is strange.

asterite:
```
Issue tracker warning: Repeated bug reports without code to reproduce them.
    May result in no answer.
```

asterite:
See also: https://groups.google.com/d/msg/crystal-lang/aEOCoytMpAk/ZXa8eptoCgAJ

stugol:
@asterite: I know Crystal is free and you don't owe me anything, but is it strictly necessary to go out of your way to be sarcastic and unhelpful? I can't produce code to reproduce the problem without first tracking down the source. I was asking how - given that Crystal has ZERO debugging support, without even line numbers in exceptions - I should go about it.

Seriously, debugging Crystal code is like pulling teeth. And when I ask for help you just take the piss. Rather than criticising MY attitude, I suggest you spend some time considering how yours could be improved.

Incidentally, I found the problem. If a large chunk of data is returned from a recursive function, it is copied each time up the call stack. Maybe you should fix that.

asterite:
Yes, sorry, instead of asking you again for some code I decided to reply in a more funny way, but my intention was not to offend you, although later I saw that that might have been the case. So then I sent that link showing that I did try to track those things down but was unlucky.

I'll try again later, maybe I was doing something wrong.

Could you still provide the code that returns a large chunk of data in the recursive function?

stugol:
All good, not to worry. I'll dig out the code when I get a chance.

Could you make Crystal easier to debug, perhaps? Like a "break into debugger and single-step" command? With variable inspection?

asterite:
I was able to set the warning callback from GC :-)

```crystal
lib LibGC
  alias WarnProc = LibC::Char*, Word ->
  fun set_warn_proc = GC_set_warn_proc(WarnProc)
  $warn_proc = GC_current_warn_proc : WarnProc
end

LibGC.set_warn_proc ->(msg, word) {
  raise String.new(msg)
}

x = [] of Pointer(Int32)
loop do
  x << Pointer.malloc(10_000_000, 0)
end
```

This will show the stack trace of where the problem originates.

[...]

PR #8281, opened 2019-10-06
GC: profiling stats and warning bindings

bcardiff:
These bindings allow further profiling stats and binding to warnings of boehm.

Extracted from:
* https://github.com/crystal-lang/crystal/issues/3997#issuecomment-320500210
* https://github.com/crystal-lang/crystal/issues/2104#issuecomment-180471871

### DIFF
--- a/src/gc.cr
+++ b/src/gc.cr
@@ -56,6 +56,18 @@ module GC
     bytes_since_gc : LibC::ULong,
     total_bytes : LibC::ULong
 
+  record ProfStats,
+    heap_size : LibC::ULong,
+    free_bytes : LibC::ULong,
+    unmapped_bytes : LibC::ULong,
+    bytes_since_gc : LibC::ULong,
+    bytes_before_gc : LibC::ULong,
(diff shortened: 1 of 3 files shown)


bcardiff:
Bdw seems to design the api in a backward compatible way. If fields are no longer user they will remain in that struct probably and the size is used as a version check.

Better to decouple for that and remove unused fields in the GC module.

src/gc.cr
```diff
@@ -56,6 +56,18 @@ module GC
...
 
+  record ProfStats,
+    heap_size : LibC::ULong,
+    free_bytes : LibC::ULong,
+    unmapped_bytes : LibC::ULong,
+    bytes_since_gc : LibC::ULong,
+    bytes_before_gc : LibC::ULong,
+    non_gc_bytes : LibC::ULong,
+    gc_no : LibC::ULong,
+    markers_m1 : LibC::ULong,
```
RX14: Perhaps rename this to `marker_threads`?

PR #8282, opened 2019-10-06
Fix parser crash for 'alias Foo?'

oprypin:
> cast from Crystal::Generic to Crystal::Path failed

(found by fuzzing)

Also make all `parse_ident()` calls more consistent

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -5414,7 +5414,9 @@ module Crystal
 
       next_token_skip_space_or_newline
 
-      name = parse_ident(allow_type_vars: false).as(Path)
+      name = parse_ident(allow_type_vars: false, parse_nilable: false)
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/syntax/parser.cr
```diff
@@ -1675,7 +1675,8 @@ module Crystal
       next_token_skip_space_or_newline
 
       name_location = @token.location
-      name = parse_ident allow_type_vars: false
+      name = parse_ident(allow_type_vars: false, parse_nilable: false)
+      raise "BUG: ModuleDef name can only be a Path" unless name.is_a?(Path)
```
asterite: What do you think if we extract these two lines into a new `parse_path` method? The `raise` message can simply be "BUG: expected Path" because it never triggers in practice.

asterite: (then of course we use `parse_path` everywhere else)

oprypin: Done, thanks

Issue #8269, opened 2019-10-04
Channel#send raise ClosedError even though successful send


firejox:
The code
```cr
ch = Channel(Int32).new

spawn do
  ch.send 1
end

spawn do
  puts ch.receive?
end

spawn do
  ch.close
end

Fiber.yield
```

The output
```
1
Unhandled exception in spawn: Channel is closed (Channel::ClosedError)
  from /home/developer/workspace/src/channel.cr:252:5 in 'raise_if_closed'
  from /home/developer/workspace/src/channel.cr:122:9 in 'send'
  from test2.cr:5:3 in '->'
  from /home/developer/workspace/src/fiber.cr:255:3 in 'run'
  from /home/developer/workspace/src/fiber.cr:48:34 in '->'
  from ???
```

Crystal version:
```
[developer@crystal-dev tmp]$ ../bin/crystal --version
Using compiled compiler at `.build/crystal'
Crystal 0.32.0-dev [cdafa0e85] (2019-10-04)

LLVM: 8.0.1
Default target: x86_64-pc-linux-gnu
```

PR #8284, opened 2019-10-07
Fix : Channel can send successfully but raise

firejox:
refer to #8269 
It uses additional state to track whether the send is successful. If not, it will raise error.

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -79,7 +79,7 @@ class Channel(T)
     Closed
   end
 
-  private record Sender(T), fiber : Fiber, value : T, select_context : SelectContext(Nil)?
+  private record Sender(T), fiber : Fiber, value : T, state_ptr : DeliveryState*, select_context : SelectContext(Nil)?
   private record Receiver(T), fiber : Fiber, value_ptr : T*, state_ptr : DeliveryState*, select_context : SelectContext(T)?
 
   def initialize(@capacity = 0)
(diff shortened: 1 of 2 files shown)


src/channel.cr
```diff
@@ -115,11 +118,17 @@ class Channel(T)
...
         @lock.unsync do
           Crystal::Scheduler.reschedule
         end
-        raise_if_closed
+
+        if state == DeliveryState::Closed
+          raise ClosedError.new
+        elsif state != DeliveryState::Delivered
+          raise "BUG: Fiber was awaken without channel delivery state set"
+        end
```
Sija: I'd rewrite it using `case`:

```cr
case state
when .closed?
  raise ClosedError.new
when .delivered?
  raise "BUG: Fiber was awaken without channel delivery state set"
end
```

firejox: Nice suggestion. However, Channel#receive use `case` with the form
```crystal
case state
when DeliveryState::Delivered
  value
when DeliveryState::Closed
  yield
else
  raise "BUG: Fiber was awaken without channel delivery state set"
end
```

If use `case`, I think this will be more symmetric.
```crystal
case state
when DeliveryState::Delivered
  # ignore
when DeliveryState::Closed
  raise ClosedError.new
else
  raise "BUG: Fiber was awaken without channel delivery state set"
end
```

bcardiff: Can you update the PR @firejox to the last snippet? `when DeliveryState::Closed`?

firejox: @bcardiff sure

bcardiff:
@firejox I will rebase this over master so the test_preview_mt can build

PR #8287, opened 2019-10-07
allow to initialize Time::Span only with seconds

mamantoha:
This is a really tiny change to allow initialize `Time::Span` only with seconds.


mamantoha:
After creating PR I found we already have this in https://github.com/crystal-lang/crystal/pull/8257.
Closed.

PR #8295, opened 2019-10-08
Faster mutex implementation (optionally reentrant)

waj:
I made some improvements in the `Mutex` implementation to optimize the time it takes to lock/unlock and handoff the ownership to waiting fibers.

Current implementation is quite slow (specially in MT mode) because it always tries to be extremely fair by passing the ownership to the first fiber in the waiting queue. But that fiber might be running in a thread that is already sleeping, and waking up threads is costly and slow. So now the fiber is still awaken but it doesn't automatically have the ownership of the lock. Instead it will compete again with other fibers that might not be sleeping yet. Also, the lock will spin some time before sleeping allowing small critical sections to be acquired without passing through the waiting queue at all.

So, the changes included in this PR are:
  * The `lock` will spin some time before sleeping
  * The `Mutex` is not reentrant by default. Reentrancy requires keeping track of the locking mutex. Non-reentrant mutex just stores an atomic bit and it's much faster to "swap" the content than doing an atomic CAS operation.
  * Ownership is not passed to the waiting fibers. They are just awaken and will try again to obtain the lock.
  * An atomic count of the number of waiting fibers is kept, so the lock over the queue can be optimistically skipped. This makes the `unlock` faster if no fibers are waiting.

I was running a benchmark https://gist.github.com/waj/0c4b5835af088e8921fee4cc2c6006ed#file-bm_lock-cr (provided by @carlhoerberg, thanks Carl!) and these are the results:

Before changes:
```
$ crystal run --release bm_lock.cr
Mutex.synchronize  78.03M ( 12.82ns) (± 1.39%)  0.0B/op  fastest

$ crystal run --release -D preview_mt bm_lock.cr
Mutex.synchronize 110.29k (  9.07µs) (± 2.15%)  0.0B/op  fastest
```

After changes:
```
$ bin/crystal run --release bm_lock.cr
Mutex.synchronize 107.36M (  9.31ns) (± 2.08%)  0.0B/op  fastest

$ bin/crystal run --release -D preview_mt bm_lock.cr
Mutex.synchronize  40.15M ( 24.91ns) (±41.89%)  0.0B/op  fastest
```

I was also running another test (https://gist.github.com/waj/0c4b5835af088e8921fee4cc2c6006ed#file-test-mutex-cr) to evaluate the correctness of the mutex:

Before changes:
```
$ crystal run --release test-cr.cr -- 1000000
00:00:00.042806000
4000000

$ crystal run --release -D preview_mt test-cr.cr -- 1000000
00:00:18.215747000
4000000
```

After changes:
```
$ bin/crystal run --release test-cr.cr -- 1000000
00:00:00.029512000
4000000

$ bin/crystal run --release -D preview_mt test-cr.cr -- 1000000
00:00:00.124867000
4000000
```

### DIFF
--- a/src/mutex.cr
+++ b/src/mutex.cr
@@ -2,35 +2,71 @@ require "crystal/spin_lock"
 
 # A fiber-safe mutex.
 class Mutex
+  @state = Atomic(Int32).new(0)
   @mutex_fiber : Fiber?
   @lock_count = 0
   @queue = Deque(Fiber).new
+  @queue_count = Atomic(Int32).new(0)
   @lock = Crystal::SpinLock.new
 
-  def lock
-    @lock.lock
-    mutex_fiber = @mutex_fiber
-    current_fiber = Fiber.current
+  def initialize(*, @reentrant = false)
+  end
 
-    if !mutex_fiber
-      @mutex_fiber = current_fiber
-      @lock.unlock
-    elsif mutex_fiber == current_fiber
-      @lock_count += 1 # recursive lock
-      @lock.unlock
-    else
-      @queue << current_fiber
-      @lock.unlock
-      Crystal::Scheduler.reschedule
+  @[AlwaysInline]
+  def lock
+    if @state.swap(1) == 0
+      @mutex_fiber = Fiber.current if @reentrant
+      return
     end
 
+    if @reentrant && @mutex_fiber == Fiber.current
+      @lock_count += 1
+      return
+    end
+
+    lock_slow
     nil
   end
 
-  def unlock
-    @lock.lock
+  @[NoInline]
+  private def lock_slow
+    loop do
+      break if try_lock
 
-    begin
+      @lock.sync do
+        @queue_count.add(1)
+        if @state.get == 0
+          if @state.swap(1) == 0
+            @queue_count.add(-1)
+            @mutex_fiber = Fiber.current
+            return
+          end
+        end
+
+        @queue.push Fiber.current
+      end
+      Crystal::Scheduler.reschedule
+    end
+
+    @mutex_fiber = Fiber.current
+    nil
+  end
+
+  private def try_lock
+    i = 1000
+    while @state.swap(1) != 0
+      while @state.get != 0
+        Intrinsics.pause
+        i &-= 1
+        return false if i == 0
+      end
+    end
+
+    true
+  end
+
+  def unlock
+    if @reentrant
       unless @mutex_fiber == Fiber.current
         raise "Attempt to unlock a mutex which is not locked"
       end
@@ -40,15 +76,26 @@ class Mutex
         return
       end
 
-      if fiber = @queue.try &.shift?
-        @mutex_fiber = fiber
-        fiber.enqueue
-      else
-        @mutex_fiber = nil
-      end
-    ensure
-      @lock.unlock
+      @mutex_fiber = nil
     end
+
+    @state.lazy_set(0)
+
+    if @queue_count.get == 0
+      return
+    end
+
+    fiber = nil
+    @lock.sync do
+      if @queue_count.get == 0
+        return
+      end
+
+      if fiber = @queue.shift?
+        @queue_count.add(-1)
+      end
+    end
+    fiber.enqueue if fiber
 
     nil
   end
--- a/spec/std/mutex_spec.cr
+++ b/spec/std/mutex_spec.cr
@@ -7,15 +7,15 @@ describe Mutex do
     mutex.unlock
   end
 
-  it "raises if unlocks without lock" do
-    mutex = Mutex.new
+  it "raises if unlocks without lock (reentrant)" do
+    mutex = Mutex.new(reentrant: true)
     expect_raises(Exception, "Attempt to unlock a mutex which is not locked") do
       mutex.unlock
     end
   end
 
-  it "can be locked many times from the same fiber" do
-    mutex = Mutex.new
+  it "can be locked many times from the same fiber (reentrant)" do
+    mutex = Mutex.new(reentrant: true)
     mutex.lock
     mutex.lock
     mutex.unlock
--- a/src/crystal/once.cr
+++ b/src/crystal/once.cr
@@ -12,7 +12,7 @@
 class Crystal::OnceState
   @rec = [] of Bool*
   {% if flag?(:preview_mt) %}
-    @mutex = Mutex.new
+    @mutex = Mutex.new(reentrant: true)
   {% end %}
 
   def once(flag : Bool*, initializer : Void*)


ysbaddaden:
> now the fiber is still awaken but it doesn't automatically have the ownership of the lock. Instead it will compete again with other fibers that might not be sleeping yet.

Weird, we usually want the exact opposite to avoid contention —waking all resources will lead them to fight again to get the lock... when n - 1 of them will return back to sleep.

waj:
@ysbaddaden this is not waking all resources. It just wake one of the fibers and it will have to compete with those still not sleeping, if they exist. If they don't it will acquire the lock right away. Otherwise, if it looses, it will spin a little bit one more time, and hopefully get the lock before going back to sleep. It might sound weird for that fiber, but the global throughput is improved because it reduces the time the mutex is not owned by any non-sleeping fiber.

ysbaddaden:
Ah, thanks for the clarification. It makes sense now. Yes, spinning is always a good idea :)

src/mutex.cr
```diff
@@ -2,35 +2,71 @@ require "crystal/spin_lock"
 
 # A fiber-safe mutex.
 class Mutex
+  @m = Atomic(Int32).new(0)
```
luislavena: For posterity sake, could `@m` be renamed to something more descriptive? Thank you.

waj: Just replaced with `@state`

RX14:
I really don't think having non-reentrant (recursive) be the default is a good idea. What happens if you accidentally recurse? A deadlock? Corruption? The only acceptable state is an exception, or having a recursive lock. There may be an unsafe option to make the lock much faster for the cost of having to be extra careful not to get it wrong, but safety *must* be the default.

asterite:
I agree. Crystal should always put safety before performance.

RX14:
@bcardiff @waj ?

waj:
If you are using mutexes is because you're dealing with "unsafe" code. I mean, there is some shared state that must be protected with a mutex in order to deal with potential race conditions. In most cases a non-reentrant mutex is all you need. A non-reentrant mutex is not owned by any fiber and that means it can be unlocked by a different fiber. That is actually a feature of the mutex. You can lock, pass a message to another fiber and assume the ownership there and unlock later.

If you accidentally try to recursively lock a mutex it will deadlock and the program will fail, way much faster and deterministic than failures that can occur for not locking at all.

In other words, we're not adding safety just making the mutex reentrant, non-reentrant mutex is way much faster and all you need in most cases. That was the rationale behind the decision of making non-reentrant option the default.

bcardiff:
The story of having better diagnostics/debugging tools is independent of the API mutex should have. Using a non-reentrant mutex as a reentrant mutex will block and not segfault, so the behavior is in the safe/sound side. The performance difference and the more common requirement for non-reentrant seems enough to stay with the _optionally reentrant_ API.

asterite:
I think this will mostly hit users in a silent way when doing stuff like:

```crystal
# called from multiple functions
def helper_function
  @mutex.synchronize do
    # some data
  end
end

def main_function
  @mutex.synchronize do
    do_something
    helper_function # Oops, deadlock, not easy to debug
  end
end
```

jhass:
Would it cost too much performance to require an explicit lock owner akin to Java's `synchronized`? It shouldn't be more than a pointer comparison and an , right? Something like

```cr
def lock(owner)
  raise "Already locked" if @current_owner = owner.object_id
  real_lock
  @current_owner = owner.object_id
end
```

Or actually a CAS operation for that.

Then Ary's example would raise.

bcardiff:
But if you are using a lock that is hold you want to wait for it to be free, not fail immediately. Maybe a reasonable timeout on debug mode can help, but I am not sure I even like that alternative.

Or you mean that only for non-reentrant mutex to signal a hey: you might need a reentrant mutex?

jhass:
Yes, the reentrant one wouldn't raise in that case, or even do anything with the owner I guess (other than keeping it for debug purposes maybe).

RX14:
It's not about debugging tools, it's about being able to debug in the first place. Users will not appreciate having to dig out gdb on a production server to debug their deadlocks. This is especially annoying because hung process detection and restarting is far less common and tricky to set up than auto-restart on crash. They'd much prefer if the process just crashed and have them a backtrace to investigate.

Is a benchmark of the non-reentrant vs reentrant available?

asterite:
I still don't understand why we have a deadlocking but less performance behavior as a default one, instead of a safer but slower one. One can always optimize a bottleneck once it's found. Having to deal with deadlocks or crashes upfront is not fun. But we'll see with feedback.

RX14:
Still:

> Is a benchmark of the non-reentrant vs reentrant available?

ysbaddaden:
With reentrant you must acces Fiber.current, thus access a thread local, then make more writes to memory when locking and unlocking. For the tight usages we make of Mutex, its necessarily faster to not have all these writes. I had 2 mutexes in my MT experiment, one private (no validation) and 1 public (deadlock detection, optionally reentrant) wrapping the private one. No need to say the private one was much more performant (for tight usages, such as in Channel, ...)

I wish there was a third option, on by default: detect and raise on deadlock/wrong fiber unlocking the mutex. Then, you know there is an issue and have a backtrace to understand it, or use a reentrant mutex if it's acceptable (thought, they're a smell for me, they bypass a bug, and can still lead to deadlock when using lock/unlock) instead of a hanging process.

I agree with Ary: validation shouldn't be opt-in. You can disable all checks *if* there is a performance issue (or you're a boss), thought unless in tight benchmarks/loops it may not be that significant (except for MT synchronization primitives).

RX14:
> I wish there was a third option, on by default: detect and raise on deadlock/wrong fiber unlocking the mutex. Then, you know there is an issue and have a backtrace to understand it, or use a reentrant mutex if it's acceptable (thought, they're a smell for me, they bypass a bug, and can still lead to deadlock when using lock/unlock) instead of a hanging process.

This is what I have been asking for all along... Have I been unclear?

lribeiro:
That would be my preference as well, safe by default, fast by need.
`.new(reentrant: true)` seems clear

RX14:
Well, there's non-reentrancy, and skipping deadlock checks in unsafe code for speed. They're different things to me (feature vs safety).

ysbaddaden:
Yes, and that's why I insisted above on having all 3 states:
1. check by default: raise on deadlock + wrong fiber unlocking (safe);
2. reentrant option (may still deadlock when skipping an unlock, but otherwise safe);
3. disable checks option: full speed (unsafe: can deadlock + any fiber can unlock).

I.e. similar to pthreads options but checked by default. It prevents deadlocks by default and reports where a deadlock/invalid unlock occurred —without having to reproduce in GDB. It keeps some flexibility with reentrant, and it won't impact performance when checks are disabled (we already check for reentrant).

jhass:
Somebody wants to open an issue about this so we don't forget it?

RX14:
@jhass this is blocking the release so it certainly won't be forgotten

asterite:
Is it blocking the release? I think we can release it like that for 0.32.0 and adjust in 0.32.1 or 0.33.0

RX14:
It's too easy to forget about a regression if it makes it into a release.

Issue #5287, opened 2017-11-13
Macro: Parsing string fail when `#{{...}}` is inside the string


anykeyh:
Parser got wrong while parsing macro with string

```crystal
macro do_something(a, b)
  "{{a.id}}#{{b.id}}"
end

puts do_something("1", "2") #Expected: 1#2
```

This will fail to compile, as macro will ignore `#{{b.id}}`

However I'm not sure I'm doing it right. I've tried with different position of backslashes  `\` to escaping it, but nothing is working.

maiha:
@anykeyh 
It would be fixed in next release. As a workaround you can use this.
```crystal
  "%s#%s" % [{{a.id}}, {{b.id}}]
```

MakeNowJust:
Better workaround for performance:

```crystal
macro do_something(a, b)
  "#{{{a.id}}}##{{{b.id}}}"
end

puts do_something("1", "2") # puts "1#2"
```

bew:
Better workaround for performance again! (nothing at runtime):
```cr
macro do_something(a, b)
  "{{a.id}}{{'#'.id}}{{b.id}}"
end

puts do_something "1", "2" # => "1#2"
```

straight-shoota:
Having stumbled across this issue, I find it only intuitive that `"#{{foo}}"` is interpreted as a macro expansion prefixed by `#` and not string interpolation of a tuple. Macro expressions are on a higher level than other code so they should have precedence.

So #5288 is IMO wrong as it provides a workaround for what should be valid by default.

straight-shoota:
There is another solution: We could completely disable macro expansion inside literals. I don't think that is a necessary feature. There doesn't even seem to be a spec for that, so maybe it's something that just happens to be but nobody really thought about it?

The thing is: There are currently two types of expansions in string literals: string interpolation (`#{}`) and macro expansion (`{{}}`). The latter is rarely talked about and probably many developers are not aware of it. It is still applied in some occasions in stdlib and probably also in shards.

If macro expansions were removed from literals, it would make the language a bit simpler. They can easily be replaced by interpolation, either as a macro or at runtime. 
```cr
# macro expansion
"foo {{ bar.id }}"
# macro string interpolation
{{ "foo #{bar.id}" }}
# runtime interpolation
"foo #{ {{ bar }} }"
```
The latter has a small performance impact because the string is composed at runtime. But that can be mitigated by using a  macro string interpolation. Apart from that, all three variants should be equivalent.

PR #8296, opened 2019-10-08
Add class info to not-nil assertion of getter!

icy-arctic-fox:
Continuation of #8200 to address some comments about adding class information. Thought it was a good idea.

The error messages now look like:

```
Class#some_property cannot be nil
Class.some_class_property cannot be nil
```

Added type annotations to generated getter methods when they are available.

### DIFF
--- a/src/object.cr
+++ b/src/object.cr
@@ -489,11 +489,11 @@ class Object
         \{% if name.is_a?(TypeDeclaration) %}
           {{var_prefix}}\{{name}}?
 
-          def {{method_prefix}}\{{name.var.id}}?
+          def {{method_prefix}}\{{name.var.id}}? : \{{name.type}}?
             {{var_prefix}}\{{name.var.id}}
           end
 
-          def {{method_prefix}}\{{name.var.id}}
+          def {{method_prefix}}\{{name.var.id}} : \{{name.type}}
             if (value = {{var_prefix}}\{{name.var.id}}).nil?
               ::raise NilAssertionError.new("\{{@type}}\{{"{{doc_prefix}}".id}}\{{name.var.id}} cannot be nil")
             else
@@ -620,7 +620,7 @@ class Object
         \{% if name.is_a?(TypeDeclaration) %}
           {{var_prefix}}\{{name.var.id}} : \{{name.type}}?
 
-          def {{method_prefix}}\{{name.var.id}}?
+          def {{method_prefix}}\{{name.var.id}}? : \{{name.type}}?
             if (value = {{var_prefix}}\{{name.var.id}}).nil?
               {{var_prefix}}\{{name.var.id}} = \{{yield}}
             else
@@ -904,7 +904,7 @@ class Object
         \{% if name.is_a?(TypeDeclaration) %}
           {{var_prefix}}\{{name.var.id}} : \{{name.type}}?
 
-          def {{method_prefix}}\{{name.var.id}}
+          def {{method_prefix}}\{{name.var.id}} : \{{name.type}}?
             if (value = {{var_prefix}}\{{name.var.id}}).nil?
               {{var_prefix}}\{{name.var.id}} = \{{yield}}
             else
--- a/src/object.cr
+++ b/src/object.cr
@@ -257,11 +257,12 @@ class Object
     pointerof(x).as(T*).value
   end
 
-  {% for prefixes in { {"", "", "@"}, {"class_", "self.", "@@"} } %}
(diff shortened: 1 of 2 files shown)


src/object.cr
```diff
@@ -495,7 +495,11 @@ class Object
         end
 
         def {{method_prefix}}\{{name.id}}
```
Blacksmoke16: How about adding the not nil return type?  Is helpful for documentation.

Sija: It's not always available and IMO seems like a bit redundant piece of information.

Blacksmoke16: If its provided it should be included TBH.

src/object.cr
```diff
@@ -496,7 +497,7 @@ class Object
...
           if (\%value = {{var_prefix}}\{{name.id}}).nil?
-            ::raise NilAssertionError.new("\{{name.id}} cannot be nil")
+            ::raise NilAssertionError.new("\{{@type}}" + "{{doc_prefix}}" + "\{{name.id}} cannot be nil")
```
icy-arctic-fox: If anyone can find a better way of doing this, that'd be great.

Blacksmoke16: You should be able to just call the macro within the string, something like  

`"\{{@type}}{{doc_prefix}}\{{name.id}} cannot be nil"`

icy-arctic-fox: I thought I could, but when I tried it, it tried to interpolate `#{{name.id}}`.

src/object.cr
icy-arctic-fox: I can add it, but would like a definite answer on whether it should be added.

icy-arctic-fox: It also looks like there might be a bug with the existing type annotations for these macros. Using a union of types puts the nillable `?` on the last type instead of the whole thing.

```crystal
getter? name : String | Symbol

# produces:

@name : String | Symbol?

def name? # No type annotation
  @name
end

# should be:

@name : (String | Symbol)?

def name? : (String | Symbol)?
  @name
end
```

bew: Note that `X | Y?` is exactly the same as `X | Y | Nil`

Blacksmoke16: Yea the `?` getter should include `Nil` in the return type.  The non `?` getter should exclude the `Nil` type in the restriction.

If it's provided, i.e. the `TypeDeclaration` context, I see no reason not to include them.  The method is already inherently that type, i.e. you can't add custom logic into it.  It will just help API docs since it would now include the expected types.

src/object.cr
Blacksmoke16: What about `"\{{@type}}#{{{doc_prefix}}}\{{name.id}} cannot be nil"`

bew: Looks like you're hitting https://github.com/crystal-lang/crystal/issues/5287 you can try to use the workaround [here](https://github.com/crystal-lang/crystal/issues/5287#issuecomment-343929797) with something like:

```
"\{{@type}}\{{ "{{doc_prefix}}".id }}\{{name.id}} cannot be nil"
```

src/object.cr
icy-arctic-fox: @bew Oh, duh, you're right. It doesn't matter if it's in parens or not.

@Blacksmoke16 I agree. If the type annotation is there, it should be applied to the getter methods. I'll add that in for all related macros.

src/object.cr
icy-arctic-fox: Thank you! Was really irritating why I could escape it. I've tested it and it works perfect.

src/object.cr
```diff
@@ -488,20 +488,31 @@ class Object
       \{% for name in names %}
         \{% if name.is_a?(TypeDeclaration) %}
           {{var_prefix}}\{{name}}?
-          \{% name = name.var %}
-        \{% end %}
 
-        def {{method_prefix}}\{{name.id}}?
-          {{var_prefix}}\{{name.id}}
-        end
+          def {{method_prefix}}\{{name.var.id}}? : \{{name.type}}?
```
Blacksmoke16: To prevent some duplication you could prob do like `{% if name.is_a?(TypeDeclaration) %} : \{{name.type}}? {% end %}`  Or something along those lines.

icy-arctic-fox: I had that originally, but then I noticed that all the other macros use this pattern, so I made it consistent. I'd be fine to change it.

straight-shoota:
@icy-arctic-fox Could you please squash this branch into two commits, one adding type name in error message, the other adding type annotations?

icy-arctic-fox:
Done and done.

Issue #8300, opened 2019-10-10
Channel#close could renqueue select fiber


firejox:
The code
```crystal
ch1 = Channel(Int32).new
ch2 = Channel(Int32).new
ch3 = Channel(Int32).new

spawn do
  select
  when x = ch1.receive
  when x = ch2.receive
  end

  puts x
end

spawn do
  ch1.send 1
  ch3.send 3
  ch2.close
end

puts ch3.receive
Fiber.yield
```

The output
```
1
3
Invalid memory access (signal 11) at address 0x0
[0x55bb997b26b6] *CallStack::print_backtrace:Int32 +118
[0x55bb997a4b09] __crystal_sigfault_handler +217
[0x55bb998105f0] sigfault_handler +40
[0x7f6d961414d0] ???
[0x0] ???
```

The crystal version
```
Crystal 0.32.0-dev [37ca414a4] (2019-10-10)

LLVM: 8.0.1
Default target: x86_64-pc-linux-gnu
```

PR #8301, opened 2019-10-10
Fix Channel#close could renqueue select fiber

firejox:
refer to #8300 . If the `select_context` has been triggered in other channel, channel will not enqueue the fiber into scheduler.

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -96,13 +96,21 @@ class Channel(T)
       @closed = true
 
       @senders.each do |sender|
-        sender.state_ptr.value = DeliveryState::Closed
-        sender.fiber.enqueue
+        if (select_context = sender.select_context) && !select_context.try_trigger
+          next
+        else
+          sender.state_ptr.value = DeliveryState::Closed
+          sender.fiber.enqueue
+        end
       end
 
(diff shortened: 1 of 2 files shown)


src/channel.cr
```diff
@@ -96,13 +96,21 @@ class Channel(T)
       @closed = true
 
       @senders.each do |sender|
-        sender.state_ptr.value = DeliveryState::Closed
-        sender.fiber.enqueue
+        if (select_context = sender.select_context) && !select_context.try_trigger
+          next
```
yxhuvud: As `next` would exit the current block, this could be rewritten to a guard like row, ie 
```
next if (select_context = sender.select_context) && !select_context.try_trigger

sender.state_ptr.value = DeliveryState::Closed
sender.fiber.enqueue
```

Sija:
Hmm, what happened to the CI?

src/channel.cr
firejox: I prefer to use explicit `if-else` rather than one line. The explicit way can indicate which fiber is allow to be enqueued into the scheduler. And this line `next if (select_context = sender.select_context) && !select_context.try_trigger` is too long for me.

PR #8302, opened 2019-10-10
Spec hooks

asterite:
This PR introduces the following hooks to spec:
- `before_each`
- `after_each`
- `before_all`
- `after_all`
- `around_each`
- `around_all`

<details>
  <summary>Example</summary>

  ```crystal
  require "spec"

  describe "outer context" do
    around_all do |context|
      puts "around before all 1"
      context.run
      puts "around after all 1"
    end

    around_all do |context|
      puts "around before all 2"
      context.run
      puts "around after all 2"
    end

    around_each do |example|
      puts "around before each 1"
      example.run
      puts "around after each 1"
    end

    around_each do |example|
      puts "around before each 2"
      example.run
      puts "around after each 2"
    end

    before_each do
      puts "before each 1"
    end

    before_each do
      puts "before each 2"
    end

    after_each do
      puts "after each 1"
    end

    after_each do
      puts "after each 2"
    end

    it "test A" do
      puts "test A"
    end

    it "test B" do
      puts "test B"
    end

    describe "nested context" do
      before_each do
        puts "before each 3"
      end

      after_each do
        puts "after each 3"
      end

      around_each do |example|
        puts "around before each 3"
        example.run
        puts "around after each 3"
      end

      it "test C" do
        puts "test C"
      end
    end
  end
  ```
</details>

<details>
  <summary>Output</summary>

```
around before all 1
around before all 2
around before each 1
around before each 2
before each 1
before each 2
test A
.after each 2
after each 1
around after each 2
around after each 1
around before each 1
around before each 2
before each 1
before each 2
test B
.after each 2
after each 1
around after each 2
around after each 1
around before each 1
around before each 2
around before each 3
before each 1
before each 2
before each 3
test C
.after each 3
after each 2
after each 1
around after each 3
around after each 2
around after each 1
around after all 2
around after all 1
```
</details>

Remaining things:
- Right now the `example` that you call `run` on in a `around_each` is a simple proxy. In RSpec it has access to the real Example that will be run, we could probably do the same.
- Same thing for `around_all` (which seems to be kind of broken in RSpec, but here it works fine).
- Add docs to everything
- Maybe add specs

And of course we need to determine if we really want all these features and they wouldn't make spec needlessly more complex.


j8r:
Don't `after_`/ `before_` hooks and `around_` hooks duplicate each other?

src/spec/context/procsy.cr
```diff
@@ -0,0 +1,15 @@
+module Spec
+  abstract class Context
+    struct Procsy
```
Sija: That's one weird name... did you mean `Proxy`?

asterite: I stole it from RSpec 😊 

https://www.rubydoc.info/github/rspec/rspec-core/RSpec/Core/Example/Procsy

It's called Procsy because it's a proxy to a Proc 😉

j8r: It's likely a word pun for a proc proxy, that's pretty funny.

asterite:
@j8r No if you only want to set up things without tearing them down, or the other way around. Plus `around` hooks are a bit more cumbersome to use (you have to call `example.run`)

j8r:
@asterite I think having to call `example.run` is fine, if it removes 4 hook and limit the increased complexity,you were previously pointing out.
That's an opinion, others may disagree of course.

asterite:
Okay, this is ready for review. I did a bunch of refactors too.

Now the two `Procsy` types provide access to the underlying `Example` or `ExampleGroup`. You can't do much with that right now, but I can imagine in the future you'd maybe want to have around hooks that do different things based on some attributes (for example tags). At least I've seen that done in RSpec and in many cases it's useful.

Issue #4835, opened 2017-08-16
Exceptions on windows


txe:
@RX14 , it's a question related to #4832. I just wanted to continue here because the conversation can turn out to be long. You said:
> I deeply disagree with creating a WinError class. Errno is platform-specific, and it needs to go. Exception hierarchy should never depend on the platform you're running on.

>For now mapping to errno will work but we must remove errno before the first release. We should abstract the exception hierarchy to be platform specific. There's a good document on how python did this that we can take inspiration from. Doing that is beyond the scope of this pr though.

I'm a bit confused by these words so decided to take "how python handles exceptions now" as a core statement :)

There is [PEP 3151](https://www.python.org/dev/peps/pep-3151/#exceptions-ignored-by-this-pep) which describes the OS and IO exception hierarchy in python. So basically, in spite of the fact that python has a number of different exceptions, it only uses one [OSError](https://docs.python.org/3/library/exceptions.html#OSError) when it comes to work with a system function which returns a system-related error.

OSError has errno attribute under Unix but errno and winerror attributes under Windows. Under Windows, the errno attribute is **loosely** determined from the Windows error code, [mapper source](https://github.com/python/cpython/blob/master/PC/errmap.h).

For me, Errno class in crystal looks the same as OSError in python. So did you suggest to add winerror field into Errno class?

RX14:
That python PEP keeps `OSError` for compatibility reasons, we don't need to be compatible and so we should remove `Errno`. A single exception type which represents multiple different logical exceptions is surely bad design. We should design an exception hierarchy which represents the error which *actually* happened, not represents the platform-specific way that error was signaled to you (like `Errno` does). Yes, it's difficult but surely it's worth it. I'm tired of checking `errno.value` every time I catch `Errno` and this is a perfect time to fix this cruft from ruby.

txe:
OK, now I see your point. But I don't think it's a perfect time for that thing, it's just beyond my competence.
Is there any way to simplify the situation with exceptions under Windows? I really don't want to do big changes because it would slow down my work on PRs.

Can I suggest the following things:
1. leave Errno class as it is 
2. use Errno for exceptions under Windows like this, for example:
```
def win_error(message)
    Errno.new (message, winerror_to_errno(LibC.get_last_error()))
end
....
raise win_error("some message")
```
3. or use class WinError < Errno where WinError simply initializes Errno like win_error mentioned before.

ysbaddaden:
[...]

Issue #1124, opened 2015-08-07
add exception filters


technorama:
This is a feature request to add exception filters similar to C# 6.

``` crystal
begin
  ...
rescue Errno => ex if ex.errno == Errno::EPERM
rescue => ex if log(ex)
rescue
end
```

jhass:
For the example I think I'd prefer #445 though.

bcardiff:
What would be the semantic if an exception occurs inside the condition? Throw it to the caller? Or keep trying to the following rescues first?

technorama:
@jhass the log() portion of the example isn't covered by #445.

asterite:
You can do:

```crystal
begin
   ...
rescue ex : Errno
  if ex.errno == Errno::EPERM
    # do something
  else
    raise ex
  end
rescue ex
  if log(ex)
    # do something
  else
    raise ex
  end
end
```

I know it's more typing, but you can do this right now without introducing features and thinking about corner cases like the one @bcardiff mentions.

Plus, my guess is that this kind of logic might be needed in 1% of the code, so it's not such a big deal.

On the other hand, the original snippet could be easily rewritten by the compiler to the expansion I've. But you won't be able to do:

```crystal
begin
   ...
rescue ex : Errno if ex.errno == Errno::EPERM
  # ...
rescue ex : Errno if ex.errno == Errno::EBADF
 # ...
end
```

Or yes, but the compiler needs more logic to handle these cases. So I think it's better if we keep things simple here.

And neither @waj nor me think that we should expand all the Errno hierarchy into exceptions, it will just bloat things up for, again, a use case that will happen 1% of the time.

So if @waj agrees, I'll close this and #445.

technorama:
@asterite Applications that deal with lots of files will need to rescue specific exceptions and to handle them differently, especially when dealing with files created by multiple users.

Your example for what can be done is NOT the same.  You can use if statements if they are the same exception class but not if they are different.  In order to rescue the example below you would need a single rescue statement with multiple if or case statements.
```
rescue Errno if condition 1
rescue Errno if condition 2
rescue SomeOtherException
rescue Exception if condition 3
end
```

Consider this request syntactic sugar for 
```
rescue ex : Errno if ex.errno == Errno::EPERM
```
is equivalent to
```
rescue ex : Errno
  if ex.errno == Errno::EPERM
    ...
  else
    CONTINUE_SEARCHING_EXCEPTIONS
  end
```

As far as corner cases there isn't one.
[...]

PR #445, opened 2015-02-23
Added specific Errno exceptions (Errno::ENOENT, ...)

ysbaddaden:
I love that Ruby raises specific Errno exceptions, so we can rescue only selected and let others go through:

```ruby
begin
  File.open("whatever") { ... }
rescue Errno::ENOENT
  ...
end
```

This is nicer to read than the current Crystal way, and avoids useless catches:

```crystal
begin
  File.open("whatever") { ... }
rescue ex : Errno
  if ex.errno == Errno::ENOENT
    ...
  else
    raise ex
  end
end
```

The drawback is that `Errno.find_exception_class(errno)` references all the specific exceptions, which in turn requires them all in the compiled binary, even if we only need a few of them. Maybe there is a better way than having a big case. Maybe this isn't even a problem.

PR #8303, opened 2019-10-10
Add missing errno constants

j8r:
That's mostly a POC to improve the error handling around `Errno`, as pointed by @bararchy on Gitter.

## Content of this PR

- Reordering the `LibC` Errno constants to facilitate their read.
- Add missing Errno constants (which also make the language compiling for the next commit).
- Add the `Error::Errno` exception class.

## Goal

Having different error class for each `Errno` allows to rescue them, instead of having to compare their integer values with `==`.
The end user would also be able to figure out what's the error name instead of having the `Errno` class with `#errno` as an integer.

## What's next

There are still places where the `Errno::Error` is exposed directly instead of a subclass.

This is an **important breaking change**, we have to weigh carefully the pros/cons before any action.


ysbaddaden:
Having different classes for Errno was rejected years ago. It creates lots of classes and increased compilation by a non negligible factor.

All I see here is noise (don't mess with libc please) and confusion around Errno::Error vs the Errno Enum for the time being.

straight-shoota:
This topic should be discussed in an issue, not a PR.

And we're certainly not going to map each errno value to a separate error type.
But we could replace `Errno` with a number of more specific error classes, which would essentially group related errnos into one type. This has been proposed before, but never really discussed. Related discussions are in #1124 and #4835. Please open an issue for this.

ysbaddaden:
See #445 (quite a small number).

straight-shoota:
@ysbaddaden #445 was duplicating each errno constant as a distinct type (same as this PR). That's not going to work, yes. But an alternative approach as described in https://github.com/crystal-lang/crystal/issues/1124#issuecomment-300585907 would be worth looking at.

RX14:
`Errno` needs to go. No other language apart from ruby couples their exception hierarchy to a decades old hack around C not being able to return tuples like this.

My vision for exceptions at `1.0` is that they not only are well-named and not platform-specific, but they contain accessible fields relating to the error. For example, a DNS lookup failure would have a getter for the name which was being resolved. A network connection failure would have the `IPAddress` attached. `errno` would still exist, for the few people who need it, but it'd largely be ignored because it's just not necessary.

PR #8304, opened 2019-10-10
Fix select with receive? and closed channels

### LINKED ISSUE #8300: Channel#close could renqueue select fiber

bcardiff:
This PR implements https://gist.github.com/bcardiff/289953a80eb3a0512a2a2f8c8dfeb1db while fixing the behavior of `select` over closed or closing channels.

The main guidance where:

* `Channel#receive` vs `Channel#receive?` differs on their behavior for closed channels. Using them directly is always blocking.
* Performing an operation over a closed and closing the channel during the operation must behave in the same way.

There was a need to extend the design of `SelectAction` and some overload/exceptions were used used to make typing work.
Regarding typing, blocking and non-blocking select have different methods so their typing can differ. Blocking select will never return `NotReady`.

Closes #8301
Fixes #8300
Follow up of https://github.com/crystal-lang/crystal/pull/8243#issuecomment-539686184

### DIFF
--- a/spec/std/concurrent/select_spec.cr
+++ b/spec/std/concurrent/select_spec.cr
@@ -144,6 +144,56 @@ describe "select" do
     x.should eq 1
   end
 
+  it "select fiber has one chance to be enqueued into scheduler (1)" do
+    ch1 = Channel(Int32).new
+    ch2 = Channel(Int32).new
+    ch3 = Channel(Int32).new
+    x = nil
+
+    spawn do
+      select
+      when x = ch1.receive
+      when x = ch2.receive
+      end
+    end
+
+    spawn do
+      ch1.send 1
+      ch3.send 3
+      ch2.close
+    end
+
+    ch3.receive.should eq(3)
+    Fiber.yield
+    x.should eq(1)
+  end
+
+  it "select fiber has one chance to be enqueued into scheduler (2)" do
+    ch1 = Channel(Int32).new
+    ch2 = Channel(Int32).new
+    ch3 = Channel(Int32).new
+    x = nil
+
+    spawn do
+      select
+      when ch1.send 1
+        x = 1
+      when ch2.send 2
+        x = 2
+      end
+    end
+
+    spawn do
+      ch1.receive
+      ch3.send 3
+      ch2.close
+    end
+
+    ch3.receive.should eq(3)
+    Fiber.yield
+    x.should eq(1)
+  end
+
   it "select same channel multiple times" do
     ch = Channel(Int32).new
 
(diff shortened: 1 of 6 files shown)


src/channel.cr
```diff
@@ -97,11 +120,13 @@ class Channel(T)
 
       @senders.each do |sender|
         sender.state_ptr.value = DeliveryState::Closed
+        sender.select_context.try &.try_trigger
```
firejox: I think this should be `next if (sender.select_context.try &.try_trigger) == false`

bcardiff: why? If there are multiple selects waiting for the same channel you want all of them to unwait knowing the reason was that the underlying channel got closed.

firejox: The reason is the issue #8300. Without `next`, the fiber whose select operation has been completed and not entered `unwait` stage can be enqueued into the scheduler.  This
is about unexpected fiber enqueue.

bcardiff: Ok, I see.

I pushed some changes that I think makes more explicit when the fiber is enqueued (or not).
And I also added the spec for both receive and send.

Thanks!

spec/std/concurrent/select_spec.cr
```diff
@@ -107,6 +144,56 @@ describe "select" do
...
+    ch2 = Channel(Int32).new
+    ch3 = Channel(Int32).new
+
+    spawn do
+      select
+      when x = ch1.receive
+      when x = ch2.receive
+      end
+
+      x.should eq(1)
```
kostya: assertions, seems not work correct when inside spawn, it just crash fiber when failed.

bcardiff: Yup, the others use helper functions that catch spec exceptions.

bcardiff:
squashed fixups and rebased on master after today's review to improve the helper methods.

PR #8308, opened 2019-10-11
Refactor callstack_spec using compile and run helper

straight-shoota:
This PR simply extracts the behaviour to compile a crystal file and run the resulting executable into a reusable helper method.

### DIFF
--- a/spec/std/callstack_spec.cr
+++ b/spec/std/callstack_spec.cr
@@ -1,68 +1,58 @@
 require "./spec_helper"
 
+private def compile_and_run_file(source_file)
+  with_tempfile("executable_file") do |executable_file|
+    puts executable_file
+    Process.run("bin/crystal", ["build", "--debug", "-o", executable_file, source_file])
+    File.exists?(executable_file).should be_true
+
+    output, error = IO::Memory.new, IO::Memory.new
+    Process.run executable_file, output: output, error: error
+
+    {output.to_s, error.to_s}
+  end
+end
+
 describe "Backtrace" do
   it "prints file line:colunm" do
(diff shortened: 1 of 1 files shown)


spec/std/callstack_spec.cr
```diff
@@ -1,68 +1,62 @@
...
-    with_tempfile("compiler_spec_output") do |path|
-      sample = datapath("backtrace_sample")
+private def compile_and_run_file(source_file)
+  with_tempfile("executable_file") do |executable_file|
+    puts executable_file
+    Process.run("bin/crystal", ["build", "--debug", "-o", executable_file, source_file])
+    File.exists?(executable_file).should be_true
+
+    {% if flag?(:darwin) %}
+      Process.run("dsymutil", ["--flat", executable_file])
```
asterite: Now that I think about it, this might not be needed anymore because IIRC we always run dsymutil in mac now

straight-shoota: I was not sure about this either, but don't have a mac for testing. I'll remove it.

spec/std/callstack_spec.cr
```diff
@@ -1,68 +1,58 @@
 require "./spec_helper"
 
-describe "Backtrace" do
-  it "prints file line:colunm" do
-    with_tempfile("compiler_spec_output") do |path|
-      sample = datapath("backtrace_sample")
+private def compile_and_run_file(source_file)
+  with_tempfile("executable_file") do |executable_file|
+    puts executable_file
```
RX14: This debug code shouldn't have been merged.

straight-shoota: Ooops.

PR #8310, opened 2019-10-11
Spec: Add ability to randomize specs

Fryguy:
This PR adds the ability to run specs in a random order.  To do so, just pass `--order random` to `crystal spec`.  Randomized specs will output a seed which can be used if the tests need to be run in the same previous random order.  To do so, just pass `--order <seed-value>`.

Default
crystal 2019-10-11 13-41-07

With random
crystal 2019-10-11 16-02-41

With seed
crystal 2019-10-11 16-03-33

---

Note that this PR takes some of the test infrastructure stuff I built in #8242 in order to test the randomization.

### DIFF
--- a/spec/std/spec/context_spec.cr
+++ b/spec/std/spec/context_spec.cr
@@ -1,15 +1,31 @@
 require "./spec_helper"
 
 describe Spec::ExampleGroup do
-  it "#randomize" do
-    root = build_spec("f.cr", count: 20)
+  describe "#randomize" do
+    it "by default" do
+      root = build_spec("f.cr", count: 20)
 
-    before_randomize = all_spec_descriptions(root)
-    root.randomize
-    after_randomize = all_spec_descriptions(root)
+      before_randomize = all_spec_descriptions(root)
+      root.randomize(Random::DEFAULT)
+      after_randomize = all_spec_descriptions(root)
 
-    after_randomize.should_not eq before_randomize
-    after_randomize.sort.should eq before_randomize.sort
+      after_randomize.should_not eq before_randomize
+      after_randomize.sort.should eq before_randomize.sort
+    end
+
+    it "with a seed" do
+      seed = Random::Secure.rand(1..99999).to_u64
+
+      root = build_spec("f.cr", count: 20)
+      root.randomize(Random::PCG32.new(seed))
+      after_randomize1 = all_spec_descriptions(root)
+
+      root = build_spec("f.cr", count: 20)
+      root.randomize(Random::PCG32.new(seed))
+      after_randomize2 = all_spec_descriptions(root)
+
+      after_randomize1.should eq after_randomize2
+    end
   end
 
   describe "#report" do
--- a/spec/std/spec/context_spec.cr
+++ b/spec/std/spec/context_spec.cr
@@ -1,6 +1,17 @@
 require "./spec_helper"
 
 describe Spec::ExampleGroup do
+  it "#randomize" do
+    root = build_spec("f.cr", count: 20)
+
+    before_randomize = all_spec_descriptions(root)
+    root.randomize
+    after_randomize = all_spec_descriptions(root)
+
+    after_randomize.should_not eq before_randomize
+    after_randomize.sort.should eq before_randomize.sort
+  end
+
   describe "#report" do
     it "should include parent's description" do
       root = FakeRootContext.new
--- a/spec/std/spec/context_spec.cr
+++ b/spec/std/spec/context_spec.cr
@@ -14,7 +14,7 @@ describe Spec::ExampleGroup do
     end
 
     it "with a seed" do
-      seed = Random::Secure.rand(1..99999).to_u64
+      seed = 12345_u64
 
       root = build_spec("f.cr", count: 20)
       root.randomize(Random::PCG32.new(seed))
--- a/src/spec.cr
+++ b/src/spec.cr
@@ -79,7 +79,11 @@ require "./spec/dsl"
 # ## Randomizing order of specs
 #
 # Specs, by default, run in the order defined, but can be run in a random order
-# by passing --order random to `crystal spec`.
+# by passing `--order random` to `crystal spec`.
+#
+# Specs run in random order will display a seed value upon completion. This seed
+# value can be used to rerun the specs in that same order by passing the seed
+# value to `--order`.
 module Spec
 end
 
@@ -107,12 +111,13 @@ OptionParser.parse do |opts|
       exit 1
     end
   end
-  opts.on("--order MODE", "run examples in random order by passing MODE as 'random'") do |mode|
-    unless %[default random].includes?(mode)
-      STDERR.puts "order must be either 'default' or 'random'"
+  opts.on("--order MODE", "run examples in random order by passing MODE as 'random' or to a specific seed by passing MODE as the seed value") do |mode|
+    if mode =~ /\A(?:default|random|\d{1,5})\Z/
+      Spec.order = mode
+    else
+      STDERR.puts "order must be either 'default', 'random', or a seed value"
       exit 1
     end
-    Spec.order = mode
   end
   opts.on("--junit_output OUTPUT_DIR", "generate JUnit XML output") do |output_dir|
     junit_formatter = Spec::JUnitFormatter.file(output_dir)
--- a/src/spec.cr
+++ b/src/spec.cr
@@ -114,11 +114,10 @@ OptionParser.parse do |opts|
   opts.on("--order MODE", "run examples in random order by passing MODE as 'random' or to a specific seed by passing MODE as the seed value") do |mode|
     if mode == "default" || mode == "random"
       Spec.order = mode
-    elsif mode.to_u64?
-      Spec.order = mode.to_u64
+    elsif seed = mode.to_u64?
+      Spec.order = seed
     else
-      STDERR.puts "order must be either 'default', 'random', or a seed value"
-      exit 1
+      abort("order must be either 'default', 'random', or a numeric seed value")
     end
   end
   opts.on("--junit_output OUTPUT_DIR", "generate JUnit XML output") do |output_dir|
--- a/src/spec.cr
+++ b/src/spec.cr
@@ -75,6 +75,11 @@ require "./spec/dsl"
 # ```
 #
 # If any such thing is marked with `focus: true` then only those examples will run.
+#
+# ## Randomizing order of specs
+#
+# Specs, by default, run in the order defined, but can be run in a random order
+# by passing --order random to `crystal spec`.
 module Spec
 end
 
@@ -101,6 +106,13 @@ OptionParser.parse do |opts|
       STDERR.puts "location #{location} must be file:line"
       exit 1
     end
+  end
+  opts.on("--order MODE", "run examples in random order by passing MODE as 'random'") do |mode|
+    unless %[default random].includes?(mode)
+      STDERR.puts "order must be either 'default' or 'random'"
+      exit 1
+    end
+    Spec.order = mode
   end
   opts.on("--junit_output OUTPUT_DIR", "generate JUnit XML output") do |output_dir|
     junit_formatter = Spec::JUnitFormatter.file(output_dir)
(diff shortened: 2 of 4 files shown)


src/random/pcg32.cr
```diff
@@ -38,20 +38,22 @@ class Random::PCG32
...
     # initialize to zeros to prevent compiler complains
     @state = 0_u64
     @inc = 0_u64
-    new_seed(initstate, initseq)
+    new_seed(@initstate, @initseq)
   end
 
-  def new_seed
-    new_seed(Random::Secure.rand(UInt64::MIN..UInt64::MAX), Random::Secure.rand(UInt64::MIN..UInt64::MAX))
+  def init_seed
```
asterite: I don't think we need this. We should generate the seed using a random, then use that seed to initialize another random. Then we don't need to ask the seed from the random (we generate it ourselves for the case of `random`).

src/spec/context.cr
```diff
@@ -7,6 +7,11 @@ module Spec
   abstract class Context
     # All the children, which can be `describe`/`context` or `it`
     getter children = [] of NestedContext | Example
+
+    def randomize(randomizer = Random::DEFAULT)
```
asterite: There's no need for a default value here because it's never called without a value.

src/random/pcg32.cr
Fryguy: That's a good idea...I'll update

src/spec/dsl.cr
```diff
@@ -230,6 +243,9 @@ module Spec
     start_time = Time.monotonic
 
     at_exit do
+      if randomizer = @@randomizer
+        root_context.randomize(randomizer)
+      end
```
asterite: maybe extract these three lines into a `maybe_randomize` helper function (all the code in a function should have the same level of abstraction).

src/spec/context.cr
Fryguy: I had original had that for testing purposes, but backed it out, so yeah I'll remove

Fryguy:
@asterite Updated.

bew:
Does it randomize things level per level? Or is it global?

Given 2 describes A/B/C, with 3 `it` each, a normal run would do:
```
A.1
A.2
A.3
B.1
B.2
B.3
C.1
C.2
C.3
```

With your PR, will it randomize like:
```
B.1
B.3
B.2
C.2
C.1
C.3
A.1
A.3
A.2
```
Or like:
```
B.3
C.2
B.2
A.1
A.3
C.1
B.1
C.3
A.2
```

?

src/spec/dsl.cr
```diff
@@ -64,6 +65,23 @@ module Spec
     exit
   end
 
+  # :nodoc:
+  class_getter randomizer_seed : Tuple(UInt64, UInt64)?
+  class_getter randomizer : Random::PCG32?
+
+  # :nodoc:
+  def self.order=(mode)
+    if mode == "random"
```
Sija: I'd rewrite it with `case mode`.

asterite:
@bew I would expect randomization to happen inside individual `describe`/`context` blocks, because doing it globally is basically impossible considering we'll have `before_all` hooks and such

straight-shoota:
Comparing with rspec, it has two different options: 
```
--order TYPE[:SEED]            Run examples by the specified order type.                            
                                 [defined] examples and groups are run in the order they are defined
                                 [rand]    randomize the order of groups and examples               
                                 [random]  alias for rand                                           
                                 [random:SEED] e.g. --order random:123                              
--seed SEED                    Equivalent of --order rand:SEED.                                     
```

I suppose it's fine to merge them into one since there don't seem to be any additional use cases.

rspec's default seed value is in `0 .. 2 ** 16`, which means a decimal number with at most 5 digits. This is much easier to handle than a 16 digit hexadecimal seed. And these smaller seed values should be totally sufficient.

Fryguy:
> Does it randomize things level per level? Or is it global?

Per-level.  This is particularly important for verbose output and before/after hooks, otherwise it would interleave improperly.

> Comparing with rspec

I find rspec's CLI confusing, and this seemed a lot simpler.  I can change though if we all agree on a different CLI interface.

> rspec's default seed value is in 0 .. 2 ** 16, which means a decimal number with at most 5 digits. This is much easier to handle than a 16 digit hexadecimal seed. And these smaller seed values should be totally sufficient.

When I started this exercise I was pulling the values out of the Random::PCG32, which was a pair of UInt64s, and I converted those to a hex string for ease of use.  But now, after @asterite's suggestion to generate a seed right in the Spec module, I can definitely do this, since I control the value.

Fryguy:
Updated with all suggestions.

For the random seed range I chose `1..99999` mainly because it's simpler to ensure 5 or less digits with a basic regex that have to try to convert it and ensure it's `1..2^16`.  As the number is arbitrary anyway, I figured this was good.

spec/std/spec/context_spec.cr
```diff
@@ -1,7 +1,31 @@
...
+      before_randomize = all_spec_descriptions(root)
+      root.randomize(Random::DEFAULT)
+      after_randomize = all_spec_descriptions(root)
+
+      after_randomize.should_not eq before_randomize
+      after_randomize.sort.should eq before_randomize.sort
+    end
+
+    it "with a seed" do
+      seed = Random::Secure.rand(1..99999).to_u64
```
straight-shoota: For testing purposes, the seed should be specified, not random.

src/spec.cr
```diff
@@ -100,6 +109,14 @@ OptionParser.parse do |opts|
       exit 1
     end
   end
+  opts.on("--order MODE", "run examples in random order by passing MODE as 'random' or to a specific seed by passing MODE as the seed value") do |mode|
+    if mode =~ /\A(?:default|random|\d{1,5})\Z/
```
straight-shoota: A Reged is not appropriate here. You should rather compare the string values directly, and else try to parse as an int.

src/spec.cr
```diff
@@ -100,6 +109,14 @@ OptionParser.parse do |opts|
...
+      Spec.order = mode
+    else
+      STDERR.puts "order must be either 'default', 'random', or a seed value"
+      exit 1
+    end
```
straight-shoota: Please use `abort`

src/spec/context.cr
```diff
@@ -7,6 +7,11 @@ module Spec
   abstract class Context
     # All the children, which can be `describe`/`context` or `it`
     getter children = [] of NestedContext | Example
+
+    def randomize(randomizer)
+      children.each { |c| c.randomize(randomizer) if c.responds_to?(:randomize) }
```
straight-shoota:
```suggestion
      children.each do |child|
        child.randomize(randomizer) if child.is_a?(NestedContext)
      end
```

src/spec/dsl.cr
```diff
@@ -64,6 +65,26 @@ module Spec
...
+  class_getter randomizer_seed : Int32?
+  class_getter randomizer : Random::PCG32?
+
+  # :nodoc:
+  def self.order=(mode)
+    seed =
+      case mode
+      when "random"
+        Random::Secure.rand(1..99999) # 5 digits or less for simplicity
+      when /\A(\d{1,5})\Z/
```
straight-shoota: Please no Regex for parsing a number! Also I don't think there needs to be a limit enforced to the allowed seed number.

src/spec/dsl.cr
```diff
@@ -64,6 +65,26 @@ module Spec
...
+        $1.to_i
+      end
```
Sija: I'd suggest adding an `else` branch possibly raising an `ArgumentError`.

src/spec.cr
Fryguy: I can do that, but it's not like that elsewhere in this file (not sure what the general approach is for things like that - stay consistent, or move forward)

Fryguy: Done...I can revert back if consistency is preferable

src/spec/dsl.cr
```diff
@@ -64,6 +65,28 @@ module Spec
...
+
+  # :nodoc:
+  def self.order=(mode)
+    seed =
+      case mode
+      when "default"
+        nil
+      when "random"
+        Random::Secure.rand(1..99999).to_u64 # 5 digits or less for simplicity
+      when UInt64
```
Sija: I'd use `Int`

Fryguy: I opened up the number based on the feedback from @straight-shoota :

> Also I don't think there needs to be a limit enforced to the allowed seed number.

I chose UInt64 for 2 reasons. 1) The unsigned part prevents negative numbers and 2) the input to Randon::PCG32 takes a UInt64, so it's cleaner to convert once (as opposed to Int, then to UInt64 later)

src/spec/dsl.cr
```diff
@@ -64,6 +65,28 @@ module Spec
...
+  class_getter randomizer : Random::PCG32?
+
+  # :nodoc:
+  def self.order=(mode)
+    seed =
+      case mode
+      when "default"
+        nil
+      when "random"
+        Random::Secure.rand(1..99999).to_u64 # 5 digits or less for simplicity
```
oprypin: Why this arbitrary range? Can't you just use `new_seed`?

src/spec/dsl.cr
```diff
@@ -64,6 +65,28 @@ module Spec
...
 
+  # :nodoc:
+  class_getter randomizer_seed : UInt64?
+  class_getter randomizer : Random::PCG32?
+
+  # :nodoc:
+  def self.order=(mode)
+    seed =
+      case mode
+      when "default"
```
oprypin: Hmm I don't like the strings being used here. At least it could be symbols. Or change it to `seed=`, `random_seed`.

src/spec/dsl.cr
oprypin: Or is the seed being printed somewhere? That would be nice.

yxhuvud: It needs to be printed, as you need to be able to set the seed to repeat a failing spec.

ysbaddaden: Yes the seed must be printed to be able to reproduce a randomly failing run —thought with parallelism involved, it will never be exactly the same run. Hence the simple seed.

src/spec/dsl.cr
Fryguy: The methods in this DSL module are basically the implementation of the CLI, so I'd like to keep it as close to the CLI as possible.  This includes incoming values as strings, and the method being named `order=` to match the `--order` CLI param.

src/spec.cr
```diff
@@ -100,6 +109,15 @@ OptionParser.parse do |opts|
       exit 1
     end
   end
+  opts.on("--order MODE", "run examples in random order by passing MODE as 'random' or to a specific seed by passing MODE as the seed value") do |mode|
+    if mode == "default" || mode == "random"
+      Spec.order = mode
+    elsif mode.to_u64?
```
straight-shoota: You can avoid parsing the number twice:
```suggestion
    elsif seed = mode.to_u64?
      Spec.order = seed
```

src/spec.cr
```diff
@@ -100,6 +109,15 @@ OptionParser.parse do |opts|
...
+      Spec.order = mode.to_u64
+    else
+      abort("order must be either 'default', 'random', or a seed value")
```
straight-shoota: Maybe this is more clear?
```suggestion
      abort("order must be either 'default', 'random', or a numeric seed value")
```

src/spec/dsl.cr
```diff
@@ -64,6 +65,28 @@ module Spec
...
+        nil
+      when "random"
+        Random::Secure.rand(1..99999).to_u64 # 5 digits or less for simplicity
+      when UInt64
+        mode
+      else
+        raise ArgumentError.new("order must be either 'default', 'random', or a seed value")
```
straight-shoota:
```suggestion
        raise ArgumentError.new("order must be either 'default', 'random', or a numeric seed value")
```

src/spec/dsl.cr
```diff
@@ -64,6 +65,28 @@ module Spec
...
+      end
+
+    @@randomizer_seed = seed
+    @@randomizer = seed && Random::PCG32.new(seed)
```
straight-shoota: I'd prefer this initialization in an `if` conditional instead of guarded by a fancy logic expression.

Fryguy: This is equivalent to 

```crystal
@@randomizer = seed ? Random::PCG32.new(seed) : nil
```

Is that preferable?

Fryguy:
@straight-shoota @RX14 I've rebased this and fixed the small comments, so this should be good to go.

PR #8312, opened 2019-10-11
Remove original backers from repo

bcardiff:
Can be found at https://crystal-lang.org/sponsors/original-sponsors.html

It caused some confusion for some sponsors. The file was no touched for 4yrs.


wontruefree:
didnt realize matz was an original supporter

PR #8313, opened 2019-10-11
Add Compiler-RT for 128bit integer support

wontruefree:
This is extracted from github.com/crystal-lang/crystal/pull/7093.

This implements all the expected functions 128bit integer support.


spec/std/crystal/compiler_rt/divti3_spec.cr
```diff
@@ -0,0 +1,28 @@
+{% skip_file unless flag?(:bits32) %}
```
Sija: Shouldn't this `unless` be `if`?
ditto elsewhere

wontruefree: good catch

wontruefree: I am on a mac that is 64bit and this flag is working as is.  When I change it to if 32bits it fails.

wontruefree:
how can I make this a draft?

bcardiff:
You can’t. Draft can only be created and be marked as ready. No worries

wontruefree:
I am not sure the 32bits flag is working as expected

spec/std/crystal/compiler_rt/divti3_spec.cr
bcardiff: I would expect that the specs can run in all the platforms. They call explicitly the functions.

src/crystal/compiler_rt/i128_info.cr
```diff
@@ -1,3 +1,4 @@
+@[Extern]
```
Sija: This should be attached to the `struct` itself and not the `module`.
ditto below

wontruefree:
@bcardiff can you take a look at this when you get a chance?

src/crystal/compiler_rt/udivti3.cr
```diff
@@ -0,0 +1,6 @@
+# Functions for returning the product of signed multiplication with overflow eg. `a * b`
+
+fun __udivti3(a : UInt128, b : UInt128) : UInt128
+  r = 0_u128
+  __udivmodti4(a, b, pointerof(r))
+end
```
Sija:
```suggestion
  r
end
```

src/crystal/compiler_rt/multi3.cr
```diff
@@ -0,0 +1,12 @@
...
+
+fun __multi3(a : Int128, b : Int128) : Int128
+  x = a.unsafe_as(CompilerRT::I128)
+  y = b.unsafe_as(CompilerRT::I128)
+  r = CompilerRT::I128.new
+
+  r.info.low = x.info.low &* y.info.low
+  r.info.high = r.info.high &+ (x.info.high &* y.info.low &+ x.info.low &* y.info.high)
+
+  return r.all
```
Sija:
```suggestion
  r.all
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+  n_udword_bits = sizeof(Int64) &* sizeof(Char)
+  n_utword_bits = sizeof(Int128) &* sizeof(Char)
+  n = a.unsafe_as(CompilerRT::UI128)
+  d = b.unsafe_as(CompilerRT::UI128)
+  q = CompilerRT::UI128.new
+  r = CompilerRT::UI128.new
+  sr = 0_u32
+
+  if n.info.high == 0
+    if (d.info.high == 0)
```
Sija:
```suggestion
    if d.info.high == 0
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+      if rem
+        rem.value = (n.info.low % d.info.low).to_u128
+      end
+      n.info.low = n.info.low // d.info.low
+      return n.all
+    end
+    rem.value = n.info.low.to_u128 if rem
+    return 0_u128
+    if (d.info.low == 0)
```
Sija:
```suggestion
    if d.info.low == 0
```

src/crystal/compiler_rt/udivti3.cr
wontruefree: r is the remainder and this is a `div` function not a `mod` function so I believe r should not be returned.

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+      if (d.info.high == 0)
```
Sija:
```suggestion
      if d.info.high == 0
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        if rem
+          rem.value = n.info.high % d.info.low
+        end
+        n.info.high = n.info.high // d.info.low
+        return n.all
+      end
+      if (n.info.low == 0)
```
Sija:
```suggestion
      if n.info.low == 0
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        if rem
+          r.info.high = n.info.high % d.info.high
+          r.info.low = 0
+          rem.value = r.all
+        end
+        n.info.high = n.info.high // d.info.high
+        return n.all
+      end
+      if ((d.info.high & (d.info.high &- 1)) == 0) # if d is a power of 2
```
Sija:
```suggestion
      if (d.info.high & (d.info.high &- 1)) == 0 # if d is a power of 2
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        if rem
+          r.info.low = n.info.low
+          r.info.high = n.info.high & (d.info.high &- 1)
+          rem.value = r.all
+        end
+        n.info.high = n.info.high >> d.s.info.high.trailing_zeros_count
+        return n.all
+      end
+      sr = d.info.high.trailing_zeros_count &- n.info.high.trailing_zeros_count
+      if (sr > n_udword_bits &- 2)
```
Sija:
```suggestion
      if sr > n_udword_bits &- 2
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        return 0_u128
+      end
+      sr = sr &+ 1
+      q.info.low = 0
+      q.info.high = n.info.low << (n_udword_bits &- sr)
+      r.info.high = n.info.high >> sr
+      r.info.low = (n.info.high << (n_udword_bits &- sr)) | (n.info.low >> sr)
+    end
+  else
+    if (d.info.high == 0)
```
Sija:
```suggestion
    if d.info.high == 0
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+      if ((d.info.low & (d.info.low &- 1)) == 0)
```
Sija:
```suggestion
      if (d.info.low & (d.info.low &- 1)) == 0
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        end
+        return n.all if d.info.low == 1
+
+        sr = d.info.low.trailing_zeros_count
+        q.info.high = n.info.high >> sr
+        q.info.low = (n.info.high << (n_udword_bits &- sr)) | (n.info.low >> sr)
+        return q.all
+      end
+      sr = 1 &+ n_udword_bits &+ d.info.low.trailing_zeros_count &- n.info.high.trailing_zeros_count
+      if (sr == n_udword_bits)
```
Sija:
```suggestion
      if sr == n_udword_bits
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        q.info.low = 0
+        q.info.high = n.info.low
+        r.info.high = 0
+        r.info.low = n.info.high
+      elsif (sr < n_udword_bits)
```
Sija:
```suggestion
      elsif sr < n_udword_bits
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        r.info.low = (n.info.high << (n_udword_bits &- sr)) | (n.info.low >> sr)
+      else
+        q.info.low = n.info.low << (n_utword_bits &- sr)
+        q.info.high = (n.info.high << (n_utword_bits &- sr)) | (n.info.low >> (sr &- n_udword_bits))
+        r.info.high = 0
+        r.info.low = n.info.high >> (sr &- n_udword_bits)
+      end
+    else
+      sr = d.info.high.trailing_zeros_count &- n.info.high.trailing_zeros_count
+      if (sr > n_udword_bits &- 1)
```
Sija:
```suggestion
      if sr > n_udword_bits &- 1
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+        rem.value = n.all if rem
+        return 0_u128
+      end
+      sr = sr &+ 1
+      q.info.low = 0
+      if (sr == n_udword_bits)
```
Sija:
```suggestion
      if sr == n_udword_bits
```

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,127 @@
...
+  if n.info.high == 0
+    if (d.info.high == 0)
+      if rem
+        rem.value = (n.info.low % d.info.low).to_u128
+      end
+      n.info.low = n.info.low // d.info.low
+      return n.all
+    end
+    rem.value = n.info.low.to_u128 if rem
+    return 0_u128
```
Sija: Is this intentional?

wontruefree: this is intentional it is ported from here https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/udivmodti4.c#L47

Sija: Making the code below essentially pointless...? I don't get it :/

bcardiff:
Oh my! those huge files of specs! @wontruefree I'm sorry!

I would not expect `_u128` constants to be used. Although for small values they are parsed.

The `udivmodti4_test.c` uses shifts and mask to build the 128 numbers

```
{(tu_int)0x0000000000000000uLL << 64 | 0x0000000000000000uLL, (tu_int)0x0000000000000000uLL << 64 | 0x0000000000000001uLL,  (tu_int)0x0000000000000000uLL << 64 | 0x0000000000000000uLL, (tu_int)0x0000000000000000uLL << 64 | 0x0000000000000000uLL},
```

instead of 

```
#   StaticArray[0x00000000000000000000000000000000.to_u128!, 0x00000000000000000000000000000001.to_u128!, 0x00000000000000000000000000000000.to_u128!, 0x00000000000000000000000000000000.to_u128!],
```

UInt64 static arrays can be used to input the hex literals and then casted.

That trick should allow you to include the commented specs also. There is no need for them to use 128 literals.

bcardiff:
I think `mulotiX_spec.cr` should require explicitly `mulotiX`. (I know it is not the case currently).

Then, the `compiler_rt.cr` will be able to include some operations based on distro and arch.

`compiler_rt.cr` can be left unchanged for this PR if you prefer, and once this is merged we can work on finding the inclusions to make `spec/compiler/codegen/arithmetics_spec.cr` and `spec/support/number.cr` work accross all platforms with the same definitions.

wontruefree:
@bcardiff would that be something like
```crystal
StaticArray[0x0000000000000000 << 64 | 0x0000000000000000]
```

Why not use the 128 literals?

src/crystal/compiler_rt/udivmodti4.cr
```diff
@@ -0,0 +1,132 @@
...
+      if rem
+        r.info.low = n.info.low
+        r.info.high = n.info.high & (d.info.high &- 1)
+        rem.value = r.all
+      end
+      n.info.high = n.info.high >> d.info.high.trailing_zeros_count
+      return n.all
+    end
+
+    sr = d.info.high.trailing_zeros_count &- n.info.high.trailing_zeros_count
```
bcardiff: Should this be leading instead of trailing zeros?
https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/udivmodti4.c#L85

```
Built-in Function: int __builtin_clz (unsigned int x)
Returns the number of leading 0-bits in x, starting at the most significant bit position. If x is 0, the result is undefined.

Built-in Function: int __builtin_ctz (unsigned int x)
Returns the number of trailing 0-bits in x, starting at the least significant bit position. If x is 0, the result is undefined.
```

bcardiff: There are more mentions of `clz` in https://github.com/llvm-mirror/compiler-rt/blob/master/lib/builtins/udivmodti4.c#L85 but no leading_zeros_count here.

wontruefree:
@bcardiff so I get errors unless I use 128 bit numbers in the tests.

```bash
$ .build/crystal spec/std/crystal/compiler_rt/divti3_spec.cr 

In spec/std/crystal/compiler_rt/divti3_spec.cr:23:3

 23 | test__divti3(HEX_0_8000000000000000, 1, HEX_0_8000000000000000)
      ^-----------
Error: no overload matches 'test__divti3' with types UInt64, Int32, UInt64
```

bcardiff:
I mean to construct the Int128 values of the spec using something like `StaticArray[0xffffffffffffffff, 0xffffffffffffffff].unsafe_as(UInt128)` where the high part does not need to be shifted.

Using a macro might help to reduce the diff between the original C code and the port.

src/crystal/compiler_rt/udivmodti4.cr
wontruefree: Thanks I took a journey down llvm MathExtras but it looks like the Crystals Int `trailing_zeros_count` and `leading_zeros_count` works.  I have updated udivmodti4 if you want to take another look.

wontruefree:
@bcardiff I have updated the tests that I can with the new Static Arrays.  However due to the current implementation of Int128 not a lot of the tests can be run.

RX14:
I don't think we can accept *this many* specs...

asterite:
Is all of this to support Int128? Is anyone using Int128 anywhere?

Note that this isn't to disregard all the work put here, but I think we are making things more complex than needed. If nobody is using Int128 and we don't have concrete use cases for it, I think it's so much simpler to drop it now and maybe add it post 1.0

bcardiff:
My take is that if the specs are too heavy in the compiler-rt, but the actual code is small enough once this PR is complete maybe we can leave the specs and implementation in a separate repo. The implementation is unlikely to change since it is low level and a port of original compiler-rt. Or putting the specs in a manual directory might work. But we can leave code layout discussion after the for when the port of the code and it's specs is done.

And yes AFAIK all the new additions in this PR are to support Int128 uniformly.

ysbaddaden:
I agree with @asterite. This is a huge work and lot of code to port from compiler-rt and to compile all the time, and to *maintain* eventually. Wouldn't it be simpler to depend on and distribute compiler-rt with crystal like we depend on and distribute libgc? It would still be possible to use another version than the distributed one, if needed.

bcardiff:
The specs are huge but the compiler-rt is small.

Requiring a compiler-rt prebuild library will impact on the different distributions.

compiler-rt has also some other functionality besides the ones been ported here. I recall some concerns that every Crystal program would have that embedded / required (either static or dynamically linked)

The most comprehensive guide I've found to build compiler-rt was https://embeddedartistry.com/blog/2017/2/20/compilerrt . Since there is no compiler-rt only package we can depend on, it will need to be compiled or extracted for each target platform.

Having a crystal port seems painless in the downstream, yet it can still allow compiler-rt prebuild library to be used if wanted / worked out.

I see the port as more doable, but I confess part of that is not wanting to deal with all the options and flavors of embedding compiler-rt upfront.

My main concern was not blocking releases downstream and it seems doable by compiling the port (which is small enough). Again, if the specs are the problem that can be worked out in a different path or repo.

asterite:
Sorry to ask again, but is compiler-rt needed for Int128 and something else (like math with overflow) or just for Int128?

bcardiff:
@asterite this PR is for Int128 but there is already a port of `mulodi4` that is used to have `Int64` overflow in 32 bits.

wontruefree:
@asterite compiler-rt is needed for 32 bit machines.  I did some more digging last night and think I might have something that will help me move forward.

These tests seem like a lot but they have been really helpful in debugging.  I can make a smaller set of tests these where the tests ported over from llvm codebase.

I can close these PRs until I am ready I have gotten some good feedback and it will take some time to get this all to work.

asterite:
>  this PR is for Int128 but there is already a port of mulodi4 that is used to have Int64 overflow in 32 bits
> compiler-rt is needed for 32 bit machines

But in #8283

> Meanwhile, we are near to drop 32bits packages (and CI). LLVM 8 is not even available in their apt repo for 32 bits

So all the work we (as a community) put into compiler-rt go into:
- Int128, for which there are no use cases and nobody asked it (I added it in the past because it seemed easy, but I was wrong)
- Some operations in 32 bits machines, which will soon not be supported anymore

That's why I question why we need to continue putting effort in these things. It's not that I don't find this PR useful in its own, but given the context where it could be applied it seems to provide very little value.

asterite:
Regarding the many specs, I don't think we need one spec per operation. We can put all specs inside a single `it` block... if having multiple specs is a concern (I don't think it poses any problems).

wontruefree:
I was doing a project with Int128 integers a while ago and found that this was uncompleted.
I would like 128bit integer support in crystal and it seems like the majority if not all the work so far has gone into supporting 32 bit machines.

asterite:
@wontruefree What did you need Int128 in your project?

wontruefree:
@asterite I was integrating with a cpp library that was using Int128.  I was looking into slices and other technical solutions but in the end I discontinued the project.  I picked this up because I thought it was interesting and I thought it would help us get to 1.0.

asterite:
Sorry for my interruption and discussion. If we can get this merged and support 32 bits and int128, I think this is the best result. Please go on!

spec/std/crystal/compiler_rt/muloti4_spec.cr
```diff
@@ -0,0 +1,95 @@
...
+  test__muloti4(Int128RT[4398046511103_i128].negate!, 2097152_i128, Int128RT[9223372036852678656_i128].negate!, 0)
+  test__muloti4(4398046511103_i128, Int128RT[2097152_i128].negate!, Int128RT[9223372036852678656_i128].negate!, 0)
+  test__muloti4(Int128RT[4398046511103_i128].negate!, Int128RT[2097152_i128].negate!, 9223372036852678656_i128, 0)
+
+  test__muloti4(2097152_i128, 4398046511103_i128, 9223372036852678656_i128, 0)
+  test__muloti4(Int128RT[2097152_i128].negate!, 4398046511103_i128, Int128RT[9223372036852678656_i128].negate!, 0)
+  test__muloti4(2097152_i128, Int128RT[4398046511103_i128].negate!, Int128RT[9223372036852678656_i128].negate!, 0)
+  test__muloti4(Int128RT[2097152_i128].negate!, Int128RT[4398046511103_i128].negate!, 9223372036852678656_i128, 0)
+
+  # test__muloti4(Int128RT[0x00000000000000B5_i64, 0x04F333F9DE5BE000_u64], Int128RT[0x0000000000000000_i64, 0x00B504F333F9DE5B_u64], Int128RT[0x7FFFFFFFFFFFF328_i64, 0xDF915DA296E8A000_u64], 0)
```
RX14: Why are there so many commented out tests?

wontruefree: they are broken.  The main problem is `__udivmodti4` but it is hard to debug and I work on one failure at a time.  I have gotten enough help from these PRs that I can close them and reopen them when I am closer if that helps.

RX14: If this is WIP, then can you indicate that in the title?

PR #5234, opened 2017-11-02
Format: indent multiline call chaining with 4 spaces

MakeNowJust:
Closes #5195 

See https://github.com/crystal-lang/crystal/pull/5195#issuecomment-340790563

PR #8315, opened 2019-10-12
Format: Fix indent in arguments

### LINKED ISSUE #8197: `.method` calls on new lines has wrong indent when nested
```cr
top_call("foo")
  .first_indent(SomeType.new("bar")
    .second_indent
    .second_indent)
  .first_indent(SomeType.new("baz")
    .second_indent
    .second_indent)
```
is wrongly formatted as:
```cr
top_call("foo")
  .first_indent(SomeType.new("bar")
  .second_indent
  .second_indent)
  .first_indent(SomeType.new("baz")
  .second_indent
  .second_indent)
```
Which is pretty bad to read..

For now I'm using:
```cr
top_call("foo")
  .first_indent(
    SomeType.new("bar")
      .second_indent
      .second_indent)
  .first_indent(
    SomeType.new("baz")
      .second_indent
      .second_indent)
```
Which is ok, but adds an additional indentation.
Nevertheless, I think the above formatting should be fixed

MakeNowJust:
Fixed #8197

Also remove `@multiline_call_indent` trick introduced by #5234.
It was originally introduced for dot alignment as `@dot_column`.
However it is removed and this trick can be implemented without any trick.

Thank you.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -1570,4 +1570,19 @@ describe Crystal::Formatter do
     end
     ex.line_number.should eq(5)
   end
+
+  # #8197
+  assert_format <<-CODE
+    foo
+      .foo1(bar
+        .bar1
+        .bar2)
+    CODE
+
+  assert_format <<-CODE
+    foo.foo1(
+      bar
+        .bar1
+        .bar2)
+    CODE
 end
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -73,7 +73,6 @@ module Crystal
     @inside_lib : Int32
     @inside_struct_or_union : Int32
     @inside_enum : Int32
-    @multiline_call_indent : Int32?
     @implicit_exception_handler_indent : Int32
     @last_write : String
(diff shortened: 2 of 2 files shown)


spec/compiler/formatter/formatter_spec.cr
```diff
@@ -1570,4 +1570,19 @@ describe Crystal::Formatter do
     end
     ex.line_number.should eq(5)
   end
+
+  # #8197
+  assert_format <<-CODE
+    foo
+      .foo1(bar
+        .bar1
+        .bar2)
```
asterite: I think this format should be preserved.

For example if I write this:

```crystal
foo
  .foo1(bar
    .bar1
    .bar2)
  .foo2(baz
    .baz2
    .baz3)
```

I'd like it to be formatted like that (without change).

MakeNowJust: This spec intend to what you say. What is problem?

asterite: Oooooh... sorry! For some reason I thought the lines below these ones are the formatted version, but now I see they are separate test cases. 😊

bew:
Thank you @MakeNowJust :heart:

PR #8320, opened 2019-10-13
Parse newline before clsoing paren of pseudo method call

### LINKED ISSUE #8318: Multi-line as, as?, is_a?, and other special methods cause syntax error
The following code generates a syntax error:

```crystal
x = 42.as(Int32?)
x.as(Int32
)
```

```
Error: expecting token ')', not 'NEWLINE'
```

This seems to happen with all "special" or "pseudo" methods. These also produce the issue:
`as?`, `is_a?`, `sizeof`, `pointerof`, `typeof`

Crystal v0.31.1

MakeNowJust:
Fixed #8318

Note that this fix does not need to modify formatter because all pseudo methods are formatted as if these methods are usual `Call`.

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -831,7 +831,7 @@ module Crystal
       if @token.type == :"("
         next_token_skip_space_or_newline
         type = parse_single_type
-        skip_space
+        skip_space_or_newline
         check :")"
         next_token_skip_space
       else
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/syntax/parser.cr
```diff
@@ -4914,6 +4914,9 @@ module Crystal
         exps << parse_op_assign
         if @token.type == :","
           next_token_skip_space_or_newline
+        else
+          skip_space_or_newline
+          check :")"
```
MakeNowJust: Maybe this fix seems funny. But these lines are derived from usual arguments parsing ([here](https://github.com/crystal-lang/crystal/blob/b01ce093ece9cbad39e081015511f8e5e4255b4d/src/compiler/crystal/syntax/parser.cr#L3397-L3400))

PR #8321, opened 2019-10-13
Fix `%` and remainder of min int value against -1

### LINKED ISSUE #8306: (Int64::MIN % -1_i64) -> Program received and didn't handle signal FPE (8)
```
$ crystal -v
Crystal 0.31.0 [8d4715a59] (2019-09-23)

LLVM: 8.0.0
```
```
$ echo 'puts (Int64::MIN % -1_i64)' | crystal eval
Program received and didn't handle signal FPE (8)
```

asterite:
Fixes #8306

### DIFF
--- a/src/int.cr
+++ b/src/int.cr
@@ -155,14 +155,18 @@ struct Int
   #
   # See `Int#/` for more details.
   def %(other : Int)
-    if other == 0
-      raise DivisionByZeroError.new
-    elsif (self < 0) == (other < 0)
-      self.unsafe_mod(other)
-    else
-      me = self.unsafe_mod(other)
-      me == 0 ? me : me + other
-    end
+    {% begin %}
+      if other == 0
+        raise DivisionByZeroError.new
+      elsif self < 0 && self == {{@type}}::MIN && other == -1
+        self.class.new(0)
+      elsif (self < 0) == (other < 0)
+        self.unsafe_mod(other)
+      else
+        me = self.unsafe_mod(other)
+        me == 0 ? me : me + other
+      end
+    {% end %}
   end
 
   # Returns `self` remainder *other*.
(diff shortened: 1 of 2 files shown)


src/int.cr
```diff
@@ -155,14 +155,18 @@ struct Int
...
-    end
+    {% begin %}
+      if other == 0
+        raise DivisionByZeroError.new
+      elsif self < 0 && self == {{@type}}::MIN && other == -1
+        self.class.new(0)
+      elsif (self < 0) == (other < 0)
+        self.unsafe_mod(other)
+      else
+        me = self.unsafe_mod(other)
```
r00ster91: What does `me` stand for? Could you maybe while you are on it write this variable out to make the meaning more clear?

asterite: It's a good idea but it can be done in a separate PR to avoid mixing intentions.

r00ster91: Okay, sure. Should I open an issue or will you make the PR? I would do it but I don't know what `me` stands for and can't find it anywhere else.

asterite: It's just two lines. It's the result of a function.

asterite: I'll maybe send a PR later, I don't think it's a big deal. It's just a temporary result. The code is easy to understand looking at just those two lines.

src/int.cr
```diff
@@ -155,14 +155,18 @@ struct Int
...
-    elsif (self < 0) == (other < 0)
-      self.unsafe_mod(other)
-    else
-      me = self.unsafe_mod(other)
-      me == 0 ? me : me + other
-    end
+    {% begin %}
+      if other == 0
+        raise DivisionByZeroError.new
+      elsif self < 0 && self == {{@type}}::MIN && other == -1
```
yxhuvud: What happens for `other == -2` ?

asterite: It seems to work fine. It gives 0.

oprypin: That should be a spec

asterite: But the bug mentioned `-1` specifically. I'm not sure we need to add specs for -2, -3 or possibly any number that might have a bug.

Sija: IIUC it stresses different path in the code, so it might be worth having a test coverage for it?

PR #8322, opened 2019-10-13
Channel refactor and optimize

firejox:
Several changes are applied:
1. use internal storage list to reduce heap allocation
2. move cleaning operations out of the lock section during close
3. remove redundant lock by making internal send/receive return state
4. because of 3, non blocking select can be determined by state
5. the wait/unwait of select action can be done in O(1)

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -1,5 +1,6 @@
 require "fiber"
 require "crystal/spin_lock"
+require "crystal/pointer_linked_list"
 
 # A `Channel` enables concurrent communication between fibers.
 #
@@ -29,7 +30,7 @@ class Channel(T)
   record UseDefault
 
   module SelectAction(S)
-    abstract def execute : S | NotReady | UseDefault
+    abstract def execute : DeliveryState
     abstract def wait(context : SelectContext(S))
     abstract def wait_result_impl(context : SelectContext(S))
     abstract def unwait
@@ -37,7 +38,6 @@ class Channel(T)
     abstract def lock_object_id
     abstract def lock
     abstract def unlock
-    abstract def available? : Bool
 
     def create_context_and_wait(state_ptr)
       context = SelectContext.new(state_ptr, self)
@@ -102,42 +102,72 @@ class Channel(T)
     Closed
   end
 
-  private record Sender(T), fiber : Fiber, value : T, state_ptr : DeliveryState*, select_context : SelectContext(Nil)?
-  private record Receiver(T), fiber : Fiber, value_ptr : T*, state_ptr : DeliveryState*, select_context : SelectContext(T)?
+  private module SenderReceiverCloseAction
+    def close
+      self.state = DeliveryState::Closed
+      _select_context = self.select_context
+      if _select_context.nil? || _select_context.try_trigger
+        self.fiber.enqueue
+      end
+    end
+  end
+
+  private struct Sender(T)
+    include Crystal::PointerLinkedList::Node
+    include SenderReceiverCloseAction
+
+    property fiber : Fiber
+    property data : T
+    property state : DeliveryState
+    property select_context : SelectContext(Nil)?
+
+    def initialize
+      @fiber = uninitialized Fiber
+      @data = uninitialized T
+      @state = DeliveryState::None
+    end
+  end
+
+  private struct Receiver(T)
+    include Crystal::PointerLinkedList::Node
+    include SenderReceiverCloseAction
+
+    property fiber : Fiber
+    property data : T
+    property state : DeliveryState
+    property select_context : SelectContext(T)?
+
+    def initialize
+      @fiber = uninitialized Fiber
+      @data = uninitialized T
+      @state = DeliveryState::None
+    end
+  end
 
   def initialize(@capacity = 0)
     @closed = false
-    @senders = Deque(Sender(T)).new
-    @receivers = Deque(Receiver(T)).new
+
+    @senders = Crystal::PointerLinkedList(Sender(T)).new
+    @receivers = Crystal::PointerLinkedList(Receiver(T)).new
+
     if capacity > 0
       @queue = Deque(T).new(capacity)
     end
   end
 
-  def close
+  def close : Nil
+    sender_list = Crystal::PointerLinkedList(Sender(T)).new
+    receiver_list = Crystal::PointerLinkedList(Receiver(T)).new
+
     @lock.sync do
       @closed = true
 
-      @senders.each do |sender|
-        sender.state_ptr.value = DeliveryState::Closed
-        select_context = sender.select_context
-        if select_context.nil? || select_context.try_trigger
-          sender.fiber.enqueue
-        end
-      end
-
-      @receivers.each do |receiver|
-        receiver.state_ptr.value = DeliveryState::Closed
-        select_context = receiver.select_context
-        if select_context.nil? || select_context.try_trigger
-          receiver.fiber.enqueue
-        end
-      end
-
-      @senders.clear
-      @receivers.clear
+      @senders, sender_list = sender_list, @senders
+      @receivers, receiver_list = receiver_list, @receivers
     end
-    nil
+
+    sender_list.each(&.value.close)
+    receiver_list.each(&.value.close)
   end
 
   def closed?
@@ -145,39 +175,52 @@ class Channel(T)
   end
 
   def send(value : T)
-    @lock.sync do
-      raise_if_closed
+    sender = Sender(T).new
 
-      send_internal(value) do
-        state = DeliveryState::None
-        @senders << Sender(T).new(Fiber.current, value, pointerof(state), select_context: nil)
-        @lock.unsync do
-          Crystal::Scheduler.reschedule
-        end
+    @lock.lock
 
-        case state
-        when DeliveryState::Delivered
-          # ignore
-        when DeliveryState::Closed
-          raise ClosedError.new
-        else
-          raise "BUG: Fiber was awaken without channel delivery state set"
-        end
+    case send_internal(value)
+    when DeliveryState::Delivered
+      @lock.unlock
+    when DeliveryState::Closed
+      @lock.unlock
+      raise ClosedError.new
+    else
+      sender.fiber = Fiber.current
+      sender.data = value
+      @senders.push pointerof(sender)
+      @lock.unlock
+
+      Crystal::Scheduler.reschedule
+
+      case sender.state
+      when DeliveryState::Delivered
+        # ignore
+      when DeliveryState::Closed
+        raise ClosedError.new
+      else
+        raise "BUG: Fiber was awaken without channel delivery state set"
       end
-
-      self
     end
+
+    self
   end
 
   protected def send_internal(value : T)
-    if receiver = dequeue_receiver
-      receiver.value_ptr.value = value
-      receiver.state_ptr.value = DeliveryState::Delivered
-      receiver.fiber.enqueue
+    if @closed
+      DeliveryState::Closed
+    elsif receiver_ptr = dequeue_receiver
+      receiver_ptr.value.data = value
+      receiver_ptr.value.state = DeliveryState::Delivered
+      receiver_ptr.value.fiber.enqueue
+
+      DeliveryState::Delivered
     elsif (queue = @queue) && queue.size < @capacity
       queue << value
+
+      DeliveryState::Delivered
     else
-      yield
+      DeliveryState::None
     end
   end
 
@@ -204,25 +247,34 @@ class Channel(T)
   end
 
   def receive_impl
-    @lock.sync do
-      receive_internal do
-        yield if @closed
+    receiver = Receiver(T).new
 
-        value = uninitialized T
-        state = DeliveryState::None
-        @receivers << Receiver(T).new(Fiber.current, pointerof(value), pointerof(state), select_context: nil)
-        @lock.unsync do
-          Crystal::Scheduler.reschedule
-        end
+    @lock.lock
 
-        case state
-        when DeliveryState::Delivered
-          value
-        when DeliveryState::Closed
-          yield
-        else
-          raise "BUG: Fiber was awaken without channel delivery state set"
-        end
+    state, value = receive_internal
+
+    case state
+    when DeliveryState::Delivered
+      @lock.unlock
+      raise "BUG: Unexpected UseDefault value for delivered receive" if value.is_a?(UseDefault)
+      value
+    when DeliveryState::Closed
+      @lock.unlock
+      yield
+    else
+      receiver.fiber = Fiber.current
+      @receivers.push pointerof(receiver)
+      @lock.unlock
+
+      Crystal::Scheduler.reschedule
+
+      case receiver.state
+      when DeliveryState::Delivered
+        receiver.data
+      when DeliveryState::Closed
+        yield
+      else
+        raise "BUG: Fiber was awaken without channel delivery state set"
       end
     end
   end
@@ -230,43 +282,52 @@ class Channel(T)
   def receive_internal
     if (queue = @queue) && !queue.empty?
       deque_value = queue.shift
-      if sender = dequeue_sender
-        sender.state_ptr.value = DeliveryState::Delivered
-        sender.fiber.enqueue
-        queue << sender.value
+      if sender_ptr = dequeue_sender
+        queue << sender_ptr.value.data
+        sender_ptr.value.state = DeliveryState::Delivered
+        sender_ptr.value.fiber.enqueue
       end
-      deque_value
-    elsif sender = dequeue_sender
-      sender.state_ptr.value = DeliveryState::Delivered
-      sender.fiber.enqueue
-      sender.value
+
+      {DeliveryState::Delivered, deque_value}
+    elsif sender_ptr = dequeue_sender
+      value = sender_ptr.value.data
+      sender_ptr.value.state = DeliveryState::Delivered
+      sender_ptr.value.fiber.enqueue
+
+      {DeliveryState::Delivered, value}
+    elsif @closed
+      {DeliveryState::Closed, UseDefault.new}
     else
-      yield
+      {DeliveryState::None, UseDefault.new}
     end
   end
 
   private def dequeue_receiver
-    while receiver = @receivers.shift?
-      if (select_context = receiver.select_context) && !select_context.try_trigger
+    while receiver_ptr = @receivers.shift?
+      select_context = receiver_ptr.value.select_context
+      if select_context && !select_context.try_trigger
+        receiver_ptr.value.state = DeliveryState::Delivered
         next
       end
 
       break
     end
 
-    receiver
+    receiver_ptr
   end
 
   private def dequeue_sender
-    while sender = @senders.shift?
-      if (select_context = sender.select_context) && !select_context.try_trigger
+    while sender_ptr = @senders.shift?
+      select_context = sender_ptr.value.select_context
+      if select_context && !select_context.try_trigger
+        sender_ptr.value.state = DeliveryState::Delivered
         next
       end
 
       break
     end
 
-    sender
+    sender_ptr
   end
 
   def inspect(io : IO) : Nil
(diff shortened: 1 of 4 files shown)


lribeiro:
Hi firejox, how does it affect performance, do you have any benchmarks you can show?

asterite:
I don't know if anyone told you, but the PRs you are sending related to concurrency are great!

src/channel.cr
```diff
@@ -102,82 +102,136 @@ class Channel(T)
...
+    end
+  end
 
   def initialize(@capacity = 0)
     @closed = false
-    @senders = Deque(Sender(T)).new
-    @receivers = Deque(Receiver(T)).new
+
+    @senders = uninitialized Crystal::StaticList
+    @senders.init
```
ysbaddaden: StaticList could just have an initialize method. It would be proper to have StaticList.new (allocated onnstackz because its a struct) rather than creating an uninitialized value then initializing it, with negligible (if any) performance impact.

firejox: Unfortunately we can't do that with simple `StaticList.new`. The `init` method will make the list become self-reference, so it requires the address of `@senders`. And  `@senders = StaticList.new` is a copy behavior. If move into initialize method, it will be initialized with incorrect address.

firejox:
@lribeiro  Here is the benchmarks

### [channel close benchmark](https://gist.github.com/firejox/d1c9c72a68d8092ad5bc31a0e0c74889)

before
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-close-bench.cr -Dpreview_mt --release -- 10000
00:54:15.125022741
```

after
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-close-bench.cr -Dpreview_mt --release -- 10000
00:40:58.541846658
```

### [channel select unwait benchmark](https://gist.github.com/firejox/6400e2d40aff03fe4ed66f329e5466c8)

before
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-select-unwait-bench.cr --release -- 100000
00:00:00.450562818
```

after
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-select-unwait-bench.cr --release -- 100000
00:00:00.210562496
```

### [channel ping pong benchmark](https://gist.github.com/firejox/11579afacf47701a4c9a5071ecdc7bce)

before
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-ping-pong-bench.cr -Dpreview_mt --release -- 500000
00:00:20.045008184
```

after
```
[developer@crystal-dev tmp]$ ../bin/crystal channel-ping-pong-bench.cr -Dpreview_mt --release -- 500000
00:00:16.457026952
```

src/channel.cr
ysbaddaden: Ah, I missed that. Sorry for the noise.

wontruefree:
Could static list be in the stdlib and not included in the prelude?

wontruefree:
Thank you @firejox for all the work on concurrency!!

src/crystal/static_list.cr
```diff
@@ -0,0 +1,98 @@
+# :nodoc:
+macro container_of(ptr, type, member)
```
asterite: Even though this is `:nodoc:` it leaks to the top-level namespace. So in my code I can call `container_of` and it will call this macro. This should probably be moved inside `Crystal::StaticList` and called like `Crystal::StaticList.container_of`. Same for the other macro.

firejox: OK, I will move these two macros inside the struct. Thank you.

firejox: @asterite Sorry for disturbing.  I have a problem when I put them into `Crystal::StaticList`. The macros seem to be unable to know the `Sender` and `Receiver` because of visibility. And now I find three ways to resolve it.

1. Remove the `private` of `Sender` and `Receiver` so that it can call `Crystal::StaticList.container_of`. But the problem is it will leak `Sender` and `Receiver` to public.

2. Pack macros into a module like `Crystal::StaticList::Reflect`, and include the module into channel. It will call the` container_of` instead of `Crystal::StaticList.container_of`. And this way will be ugly and hard to understand.

3. Use annotation and macro to generate corresponding method. Then it will invoke
something like `Receiver(T).from_{{ member_name }}(ptr)`. This could increase the compile time because of extra computation.

Here is the code pieces about these changes
https://gist.github.com/firejox/a6f2193a5e0820eef7e7e36c937c2330

Maybe there are other clever ways that I didn't come out. What do you think?

asterite: Instead of private visibility you can just put :nodoc: in a doc comment.

Sija: @asterite IMO they shouldn't leak into the top-level scope anyway.

firejox: OK, I will use the :nodoc: . Thank you.

src/channel.cr
```diff
@@ -204,69 +260,86 @@ class Channel(T)
...
+    receiver = Receiver(T).new
+
+    @lock.lock
+
+    state, value = receive_internal
+
+    case state
+    when DeliveryState::Delivered
+      @lock.unlock
+      value.as(T)
```
bcardiff: this cast is needed because the distinguished value `UseDefault` is removed and not used in `receive_internal` when returning `{DeliveryState::Closed, nil}`. I like to be stricter here I think.

src/channel.cr
```diff
@@ -352,19 +405,18 @@ class Channel(T)
 
     ops_locks.each &.lock
 
-    # Check that no channel is closed
-    unless ops.all?(&.available?)
```
bcardiff: If this is removed, then the `available?` is no longer used.

But I don't follow where the check that non of the channel is originally closed is done. Because if there is a channel ready and other closed I think the current code in this PR will yield and not raise.

src/channel.cr
```diff
@@ -373,8 +425,8 @@ class Channel(T)
       return ops.size, NotReady.new
     end
 
-    state = Atomic(SelectState).new(SelectState::Active)
-    contexts = ops.map &.create_context_and_wait(pointerof(state))
+    state_ptr = Pointer(Atomic(SelectState)).malloc(1, Atomic(SelectState).new(SelectState::Active))
```
bcardiff: Why we need for the state to be in the heap instead of the stack? The fiber will continue from this point and that was the reason we can use the stack.

src/crystal/static_list.cr
```diff
@@ -0,0 +1,97 @@
+# :nodoc:
+struct Crystal::StaticList
```
bcardiff: I would like to find a more crystal way to build this static list.
I think it should be possible to use a generic struct that will enforce typing and maybe allow some macros to hide the boilerplate as done here. But all the pointer, casting and final interface make me doubt.

Also, although it `:nodoc:` I think there could be some specs on the implementation.

The main problem is to build a linked list with data that should act as nodes and let them be stored in the stack, right?

src/channel.cr
firejox: Ok, I will make `UseDefault` back. Returning {DeliveryState::Closed, UseDefault} sounds better. Thanks.

src/channel.cr
firejox: Because `op.execute` will return proper state, I think there is no need to check channel twice. The case you mentioned is a new case, and current spec doesn't cover it. The following code:

```crystal
# ch1 is ready and ch2 is closed

# a.cr
select
when m1 = ch1.receive?
when m2 = ch2.receive
end

# b.cr
select
when m1 = ch1.receive
when m2 = ch2.receive?
end
```

What are the expect behaviors of the two code pieces?

src/channel.cr
firejox: Because I move the cleaning operations out of lock section during close. Consider the case. A `channel` is closing and needs to clean up the **long list**. And a `select` operation has been done at the same time. The `select` had a `Sender` or `Receiver` in the **list**. The action will not be removed because we don't know the action has been iterated during close. If it iterate the action after the `select` return, the `select_context.try_trigger` may modify the invaild address. Hence, the simple solution is put the state in the heap.

src/crystal/static_list.cr
firejox: I think this is a difference between [internal storage and external storage](https://en.wikipedia.org/wiki/Linked_list#Internal_and_external_storage). The external storage will enforce type explicit, and the internal storage requires some way to know which type stores the list node. That's why need the `Crystal::StaticList.container_of`. Because we store data in stack, it cannot prevent to handle without pointer.

And I will add some spec about the `StaticList`.

src/channel.cr
bcardiff: Ok. We can avoid the double check and do not have a guarantee regarding which channel, the closed or the ready one, will fulfill the select.

* `a.cr` will either execute  ch1's `when` with `m1 = nil`, or ch2's `when` with data.
* `b.cr` will either raise due to ch1 been closed, or ch2's `when` with data.

There are no specs mixing `receive` and `receive?` though. 

If `ch2` would not be ready I would expect that `b.cr` will **always** raise. Both on a blocking and non-blocking `select`. That is a spec that could be added. But I don't expect that `receive` and `receive?` are mixed.

So, this change is good.

src/channel.cr
bcardiff: Great. Can we add that as a comment at this point in the code?

firejox: Sure, I will add some description about it.

src/channel.cr
firejox: I see. Although your precondition is different from me, I will add the spec about mixing `receive` and `receive?`.

src/crystal/static_list.cr
bcardiff: What do you think about something like this https://gist.github.com/bcardiff/4327823bb43cbb93734fbb6240315dac ?

* there are less casts
* the pointer tricker is hidden in the module
* the API is a bit more clean: sometimes as methods, sometimes as macro calls.

there might be some subtle differences regarding the representation of the list. and this solution can only be used in `structs`. I didn't care to apply them to classes since I think it is only used in structs.

In case we need to extend it, in the worst case macros can be injected in the type that is including the module to specialize if the next/prev need to be pointer(T) or T.

firejox: Sorry for late on reply because I'm busy today. In my opinion, this change is a bit worse than mine. `Listify.prepend(list, node)` will make code like c, not crystal. Although I currently use `node.append_to pointerof(list)` in code, the implementation can be easily replaced by `list.append node.self_pointer`. `list.append` and `list.shift` will be better than `Listify.append` and `Listify.shift`.

Besides, the `each` operation should not hide the pointer. If you hide the pointer, the object you get is the copy object of the pointer point to. Any changes will not apply on the original object. This will be a problem.

Including a module is good, but I don't want to limit the name of member variable in  struct/class which use the list. (If possible, they may need the list with different name for future refactor/optimize)

firejox:
Updates:

1. add specs of `Crystal::StaticList`
2. add specs of `mixing receive and receive? select`

spec/std/crystal/static_list_spec.cr
```diff
@@ -0,0 +1,299 @@
...
+private class TestedObject2
+  property value : Int32
+  property link : Crystal::StaticList
+
+  def initialize(@value : Int32)
+    @link = uninitialized Crystal::StaticList
+  end
+end
+
+describe "Crysta::StaticList" do
```
RX14: `Crystal`.

src/channel.cr
```diff
@@ -373,8 +426,11 @@ class Channel(T)
       return ops.size, NotReady.new
     end
 
-    state = Atomic(SelectState).new(SelectState::Active)
-    contexts = ops.map &.create_context_and_wait(pointerof(state))
+    # Because channel#close may clean up a long list, `select_context.try_trigger` may
+    # be called after the select return. In order to prevent invalid address access,
+    # the state is allocated in the heap.
+    state_ptr = Pointer(Atomic(SelectState)).malloc(1, Atomic(SelectState).new(SelectState::Active))
```
RX14: Can probably just do `Pointer.malloc` here - the compiler will guess the type from the `Atomic(SelectState).new(SelectState::Active)`.

src/crystal/static_list.cr
RX14: Look at the way that `Thread::LinkedList` does it, and i'm pretty sure you can generalise that to work on the raw pointers. Then you don't need the `init_empty_list` either, just store the head and tail and use null pointers to signify the start/end of the list.

spec/std/crystal/static_list_spec.cr
straight-shoota: This typo can be avoided by removing the quotes.

RX14:
I still don't understand the reason why you need `container_of`. Could you explain it for me?

RX14:
Like, why not just make your list generic and assume that `T` has the `next` and `prev` pointers already, just like `Thread::LinkedList` does. You don't have to do all this awful fancy pointer magic casting, the types are just... correct.

firejox:
@RX14 I want to keep some flexibility, so I use `container_of` and do not assume that `T` has `next` and `prev` pointer. All things is just for flexibility.

You cannot do that as `Atomic(T)` does. I have tried before I made this PR. The main problem will occur on the design of `push` operation. It should be expected to accept `Pointer` type, not `Value` type.

Now I have made the code in generic way. It seems that my implementation is not widely accepted by core team.

asterite:
>  It seems that my implementation is not widely accepted by core team

I personally like the performance improvements but I think the code becomes much harder to follow, with many low-level tricks. My fear is that in a couple of months we won't be able to understand how it all works.

I wish for the performance improvements to happen without introducing that much of a code change.

RX14:
> I want to keep some flexibility, so I use `container_of` and do not assume that `T` has `next` and `prev` pointer. All things is just for flexibility.

This flexibility - when the type will never be used outside of the stdlib - is not worth the drop in readability and the complication of the user-facing list API.



> I personally like the performance improvements but I think the code becomes much harder to follow, with many low-level tricks. My fear is that in a couple of months we won't be able to understand how it all works.

I think really the only complicated tricks are in the list implementation. If that gets fixed, then i'm perfectly happy to merge this, pending @bcardiff's input.

bcardiff:
@firejox I made another round on this PR.

Applied a couple of refactors to use linked list specifically for this use and move the shared state to a class instead of a manual `Pointer.malloc`. These changes are applied in their own commits at https://github.com/bcardiff/crystal/tree/refactor/linked-list

The underlying behavior is the same.

Then I perform the benchmarks you sent at https://github.com/crystal-lang/crystal/pull/8322#issuecomment-541532202, with `-Dpreview_mt` and without (note that not all of them in your original posting use `-Dpreview_mt`) and I am not able to validate the improvements locally in the `channel-close-bench.cr`.

bench | base | firejox | bcardiff
---|---|---|---
MT channel-close-bench.cr | 00:10:42.358743895 | 00:13:56.305350595 | 00:14:03.869554707
MT channel-select-unwait-bench.cr | 00:00:00.686825065 | 00:00:00.635335868 | 00:00:00.684841458
MT channel-ping-pong-bench.cr | 00:00:07.174520691 | 00:00:05.834123856 | 00:00:05.903892141
ST channel-close-bench.cr | 00:03:55.545640555 | 00:03:26.212166600 | 00:03:20.613478228
ST channel-select-unwait-bench.cr | 00:00:00.216716945 | 00:00:00.048762337 | 00:00:00.051219858
ST channel-ping-pong-bench.cr | 00:00:02.652404693 | 00:00:02.439848184 | 00:00:02.458077994

* bcardiff - 3c8f69a946a3b49b4a68297e144c8288622334a0
* firejox - b540ce5481b24fcf5b6045e655d423083ee69924
* base - b01ce093ece9cbad39e081015511f8e5e4255b4d

Which platform/hardware are you using?
Mine is Darwin Kernel Version 18.6.0 MacBookPro14,3 Intel Core i7 2.9 GHz

Between the refactor and the fix of the potential invalid memory access, I don't mind the slowness in `channel-close-bench.cr`.

Do you want to add those commits here? Or we can open another PR based on this one and pick it up from there. As you prefer. Again, thanks for all the effort.

firejox:
@bcardiff Interesting. I don't know why it does not yield the improvement. It actually shows the difference on my computer.

> Which platform/hardware are you using?
> Mine is Darwin Kernel Version 18.6.0 MacBookPro14,3 Intel Core i7 2.9 GHz

My platform is Linux 5.1.7-arch1-1, x86_64, AMD Athlon X2 II 270 3.4 GHz

> Do you want to add those commits here? Or we can open another PR based on this one and pick it up from there. As you prefer. Again, thanks for all the effort.

The changes are great. It is more readable. I will merge them later. Thank you.

bcardiff:
Sure thing @firejox thanks for keep iterating.

src/channel.cr
```diff
@@ -102,82 +114,126 @@ class Channel(T)
...
 
-      @senders.clear
-      @receivers.clear
+    receiver_list.each do |receiver_ptr|
+      receiver_ptr.value.state = DeliveryState::Closed
+      select_context = receiver_ptr.value.select_context
+      if select_context.nil? || select_context.try_trigger
+        receiver_ptr.value.fiber.enqueue
+      end
     end
```
Sija: You could iterate both arrays in one go by combining them `(sender_list + receiver_list).each do |ptr|`

asterite: That allocates a new array and if this is a hot path it will slow down things.

src/crystal/pointer_linked_list.cr
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  # Returns and returns head from the list, `nil` if empty.
+  def shift?
+    shift { nil }
+  end
+
+  # Iterates the list.
+  def each : Nil
+    unless empty?
```
Sija: Prefer early returns - `return if empty?`.

src/channel.cr
Sija: You could rewrite it using a `Tuple` instead:

```cr
{sender_list, receiver_list}.each do |list|
  list.each do |ptr|
    # ...
  end
end
```

asterite: Yes, or a macro to unroll the loop

straight-shoota: Or just a private helper method for the iteration body?

firejox: @straight-shoota Like this?
```crystal
sender_list.each(&.value.close_helper)
receiver_list.each(&.value.close_helper)
```

src/channel.cr
```diff
@@ -102,82 +114,126 @@ class Channel(T)
...
+    @senders = Crystal::PointerLinkedList(Sender(T)).new
+    @receivers = Crystal::PointerLinkedList(Receiver(T)).new
+
     if capacity > 0
       @queue = Deque(T).new(capacity)
     end
   end
 
-  def close
+  private macro close_sender_receiver(ptr)
```
Sija: Why a `macro` instead of regular `def`?

firejox: I think there is no difference between `macro` and `def` in this case. Just prefer `macro`.

Sija: Macros should be used only when necessary. I'd suggest using `def` instead.

firejox: OK, I see. I will change that. BTW, I have a question about macro use. For example, [this](https://github.com/crystal-lang/crystal/blob/5a5704c4dd446be7570f8a39e9e552578c462ac8/src/debug/dwarf/line_numbers.cr#L264). Is it necessary or just a bad example?

Sija: @ysbaddaden will know the best, since he wrote this piece of code.

ysbaddaden: Macros should be avoided unless you need to generate code, yet macros can be a nice alternative to defs in some occasions.

In the DWARF::LineNumbers example, `registers` is a struct so we can avoid some HEAP allocations, but we can't pass the struct as a method argument because it would pass it by value (copy it) when we want `registers` to be modified in-place. I could pass a pointer, but pointers are meh in Crystal (and rightfully so). We could pass the struct and return it back, but it's boring, there is no guarantee that LLVM will optimize the copies away (or even inline the method).

I.e. the macro avoids a repetition, HEAP allocations and/or a pointer. I believe it's acceptable.

firejox: @ysbaddaden I understand now. Thank you.

src/channel.cr
```diff
@@ -69,8 +69,20 @@ class Channel(T)
...
+  private class SelectContextSharedState
+    @state : Atomic(SelectState)
+
+    def initialize(value : SelectState)
+      @state = Atomic(SelectState).new(value)
+    end
+
+    def compare_and_set(cmp : SelectState, new : SelectState) : {SelectState, Bool}
+      @state.compare_and_set(SelectState::Active, SelectState::Done)
+    end
```
Sija: hmm, seems like these arguments aren't used anywhere... 🤔

bcardiff: You're right. Fortunately, it is only used on those values 🙈.

PR #8324, opened 2019-10-14
Optimize Levenshtein.distance

r00ster91:
This avoids two expensive array allocations (`.chars`) because the strings can already be accessed directly.
This improves speed and vastly improves memory allocation.

```cr
Benchmark.ips do |bm|
  bm.report "old" do
    Levenshtein.distance("algorithm", "altruistic")
    Levenshtein.distance("hello", "hallo")
    Levenshtein.distance("こんにちは", "こんちは")
    Levenshtein.distance("hey", "hey")
    Levenshtein.distance("hippo", "zzzzzzzz")
    Levenshtein.distance("a" * 100000, "hello")
    Levenshtein.distance("hello", "a" * 100000)
  end

  bm.report "new" do
    Levenshtein.new_distance("algorithm", "altruistic")
    Levenshtein.new_distance("hello", "hallo")
    Levenshtein.new_distance("こんにちは", "こんちは")
    Levenshtein.new_distance("hey", "hey")
    Levenshtein.new_distance("hippo", "zzzzzzzz")
    Levenshtein.new_distance("a" * 100000, "hello")
    Levenshtein.new_distance("hello", "a" * 100000)
  end
end
```
```
old 295.78  (  3.38ms) (± 8.86%)  0.95MB/op   2.26× slower
new 669.21  (  1.49ms) (± 8.87%)   195kB/op        fastest
```

### DIFF
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -47,28 +47,29 @@ module Levenshtein
 
       last_cost
     else
-      reader1 = Char::Reader.new(string1)
-      reader2 = Char::Reader.new(string2)
+      reader = Char::Reader.new(string1)
+  
+      # Use an array instead of a reader to decode the string only once
+      chars = string2.chars 
 
       # This is to allocate less memory
       if t_size > s_size
-        reader2, reader1 = reader1, reader2
+        chars, reader = reader, chars
         t_size = s_size
       end
 
       v = Pointer(Int32).malloc(t_size + 1) { |i| i }
 
       last_cost = 1
-      reader1.each do |char1|
+      reader.each do |char1|
         j = 0
-        reader2.each do |char2|
+        chars.each do |char2|
           sub_cost = char1 == char2 ? 0 : 1
           cost = Math.min(Math.min(last_cost, v[j + 1]), v[j] + sub_cost)
           v[j] = last_cost
           last_cost = cost
           j += 1
         end
-        reader2.pos = 0
 
         last_cost += 1
 
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -13,11 +13,11 @@ module Levenshtein
   def self.distance(string1 : String, string2 : String) : Int32
     return 0 if string1 == string2
 
-    s = string1.chars
-    t = string2.chars
+    s = string1.to_unsafe
+    t = string2.to_unsafe
 
-    s_size = s.size
-    t_size = t.size
+    s_size = string1.size
+    t_size = string2.size
 
     return t_size if s_size == 0
     return s_size if t_size == 0
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -29,7 +29,7 @@ module Levenshtein
         t_size, s_size = s_size, t_size
       end
 
-      v = Pointer(Int32).malloc(t_size + 1) { |i| i }
+      costs = Pointer(Int32).malloc(t_size + 1) { |i| i }
 
       i = 0
       last_cost = 0
@@ -38,11 +38,11 @@ module Levenshtein
 
         t_size.times do |j|
           sub_cost = s[i] == t[j] ? 0 : 1
-          cost = Math.min(Math.min(last_cost + 1, v[j + 1] + 1), v[j] + sub_cost)
-          v[j] = last_cost
+          cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
+          costs[j] = last_cost
           last_cost = cost
         end
-        v[t_size] = last_cost
+        costs[t_size] = last_cost
       end
 
       last_cost
@@ -58,22 +58,20 @@ module Levenshtein
         t_size = s_size
       end
 
-      v = Pointer(Int32).malloc(t_size + 1) { |i| i }
+      costs = Pointer(Int32).malloc(t_size + 1) { |i| i }
 
       last_cost = 1
       reader.each do |char1|
-        j = 0
-        chars.each do |char2|
+        chars.each_with_index do |char2, index|
           sub_cost = char1 == char2 ? 0 : 1
-          cost = Math.min(Math.min(last_cost, v[j + 1]), v[j] + sub_cost)
-          v[j] = last_cost
+          cost = Math.min(Math.min(last_cost, costs[index + 1]), costs[index] + sub_cost)
+          costs[index] = last_cost
           last_cost = cost
-          j += 1
         end
 
         last_cost += 1
 
-        v[t_size] = last_cost
+        costs[t_size] = last_cost
       end
 
       last_cost
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -13,36 +13,70 @@ module Levenshtein
   def self.distance(string1 : String, string2 : String) : Int32
     return 0 if string1 == string2
 
-    s = string1.to_unsafe
-    t = string2.to_unsafe
-
     s_size = string1.size
     t_size = string2.size
 
     return t_size if s_size == 0
     return s_size if t_size == 0
 
-    # This is to allocate less memory
-    if t_size > s_size
-      t, s = s, t
-      t_size, s_size = s_size, t_size
-    end
+    if string1.ascii_only? && string2.ascii_only?
+      s = string1.to_unsafe
+      t = string2.to_unsafe
 
-    v = Pointer(Int32).malloc(t_size + 1) { |i| i }
-
-    s_size.times do |i|
-      last_cost = i + 1
-
-      t_size.times do |j|
-        sub_cost = s[i] == t[j] ? 0 : 1
-        cost = Math.min(Math.min(last_cost + 1, v[j + 1] + 1), v[j] + sub_cost)
-        v[j] = last_cost
-        last_cost = cost
+      # This is to allocate less memory
+      if t_size > s_size
+        t, s = s, t
+        t_size, s_size = s_size, t_size
       end
-      v[t_size] = last_cost
-    end
 
-    v[t_size]
+      v = Pointer(Int32).malloc(t_size + 1) { |i| i }
+
+      i = 0
+      last_cost = 0
+      s_size.times do |i|
+        last_cost = i + 1
+
+        t_size.times do |j|
+          sub_cost = s[i] == t[j] ? 0 : 1
+          cost = Math.min(Math.min(last_cost + 1, v[j + 1] + 1), v[j] + sub_cost)
+          v[j] = last_cost
+          last_cost = cost
+        end
+        v[t_size] = last_cost
+      end
+
+      last_cost
+    else
+      reader1 = Char::Reader.new(string1)
+      reader2 = Char::Reader.new(string2)
+
+      # This is to allocate less memory
+      if t_size > s_size
+        reader2, reader1 = reader1, reader2
+        t_size = s_size
+      end
+
+      v = Pointer(Int32).malloc(t_size + 1) { |i| i }
+
+      last_cost = 1
+      reader1.each do |char1|
+        j = 0
+        reader2.each do |char2|
+          sub_cost = char1 == char2 ? 0 : 1
+          cost = Math.min(Math.min(last_cost, v[j + 1]), v[j] + sub_cost)
+          v[j] = last_cost
+          last_cost = cost
+          j += 1
+        end
+        reader2.pos = 0
+
+        last_cost += 1
+
+        v[t_size] = last_cost
+      end
+
+      last_cost
+    end
   end
 
   # Finds the closest string to a given string amongst many strings.
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -48,9 +48,9 @@ module Levenshtein
       last_cost
     else
       reader = Char::Reader.new(string1)
-  
-      # Use an array instead of a reader to decode the string only once
-      chars = string2.chars 
+
+      # Use an array instead of a reader to decode the second string only once
+      chars = string2.chars
 
       # This is to allocate less memory
       if t_size > s_size
--- a/src/levenshtein.cr
+++ b/src/levenshtein.cr
@@ -31,7 +31,6 @@ module Levenshtein
 
       costs = Pointer(Int32).malloc(t_size + 1) { |i| i }
 
-      i = 0
       last_cost = 0
       s_size.times do |i|
         last_cost = i + 1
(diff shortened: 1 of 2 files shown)


src/levenshtein.cr
```diff
@@ -13,11 +13,11 @@ module Levenshtein
   def self.distance(string1 : String, string2 : String) : Int32
     return 0 if string1 == string2
 
-    s = string1.chars
-    t = string2.chars
+    s = string1.to_unsafe
+    t = string2.to_unsafe
```
yxhuvud: I think it should be possible to use `String#each_char_with_index` instead?

asterite:
This is wrong. to_unsafe will work at the byte level. chars work at the codepoint level. That specs pass is a coincidence.

straight-shoota:
Maybe this could be optimized using `Char::Reader`. But slice is not going to work. This just means specs need to be improved as well. There's only a single example with non-ASCII characters, and that seems to pass for some reason.

asterite:
**However**, it's true that if both strings are `ascii_only?`, a different, more efficient path could be written, and this using `to_unsafe` or `to_slice`.

src/levenshtein.cr
r00ster91: What slice do you mean? This is a pointer.

r00ster91:
Okay I think I figured it out now.

src/levenshtein.cr
r00ster91: Well doesn't matter too much. They are very similar.

src/levenshtein.cr
```diff
@@ -13,36 +13,70 @@ module Levenshtein
...
+      end
+
+      v = Pointer(Int32).malloc(t_size + 1) { |i| i }
+
+      last_cost = 1
+      reader1.each do |char1|
+        j = 0
+        reader2.each do |char2|
+          sub_cost = char1 == char2 ? 0 : 1
+          cost = Math.min(Math.min(last_cost, v[j + 1]), v[j] + sub_cost)
```
r00ster91: I got the opportunity to omit some `+ 1`s.

src/levenshtein.cr
```diff
@@ -13,36 +13,70 @@ module Levenshtein
...
+        v[t_size] = last_cost
       end
-      v[t_size] = last_cost
-    end
 
-    v[t_size]
+      last_cost
+    else
+      reader1 = Char::Reader.new(string1)
+      reader2 = Char::Reader.new(string2)
```
asterite: Could you provide a benchmark for large non-ascii strings with `chars` vs. `Char::Reader`? With `Char::Reader` it means we are decoding the second string many times, whereas before it was just decoded once, and turns into an array of chars.

r00ster91:
```cr
require "benchmark"

Benchmark.ips do |bm|
  bm.report "1" do
    Levenshtein.distance1("🐓" * 100000, "hello")
    Levenshtein.distance1("hello", "🐓" * 100000)
  end

  bm.report "2" do
    Levenshtein.distance2("🐓" * 100000, "hello")
    Levenshtein.distance2("hello", "🐓" * 100000)
  end
end
```
```
      .chars 253.65  (  3.94ms) (± 8.42%)  1.53MB/op        fastest
Char::Reader 184.69  (  5.41ms) (± 7.57%)   781kB/op   1.37× slower
```
Yes for large non-ASCII strings `.chars` is a bit faster.
Even for small non-ASCII strings (25 instead of 100000):
```
      .chars 682.39k (  1.47µs) (± 8.06%)  736B/op        fastest
Char::Reader 592.25k (  1.69µs) (± 9.69%)  320B/op   1.15× slower
```
But the important difference is also the memory usage. `.chars` uses a lot more memory than the `Char::Reader` way.

But actually, what if I *combine* `.chars` and `Char::Reader` and use `.chars` only for the second string?

```cr
require "benchmark"

Benchmark.ips do |bm|
  bm.report "2x .chars" do
    Levenshtein.distance1("🐓" * 100000, "hello")
    Levenshtein.distance1("hello", "🐓" * 100000)
  end

  bm.report "1x Char::Reader + 1x .chars" do
    Levenshtein.distance2("🐓" * 100000, "hello")
    Levenshtein.distance2("hello", "🐓" * 100000)
  end
end
```
```
                  2x .chars 522.71  (  1.91ms) (± 9.40%)  782kB/op   1.42× slower
1x Char::Reader + 1x .chars 740.84  (  1.35ms) (± 9.90%)  391kB/op        fastest
```
It seems like a good balance between memory usage and speed.
25 instead of 100000:
```
                  2x .chars 555.75k (  1.80µs) (±10.41%)  736B/op   1.50× slower
1x Char::Reader + 1x .chars 835.18k (  1.20µs) (±12.23%)  528B/op        fastest
```
What do you think?

asterite: Yes, mixing strategy would be great!

src/levenshtein.cr
```diff
@@ -13,36 +13,69 @@ module Levenshtein
...
-        v[j] = last_cost
-        last_cost = cost
+      i = 0
+      last_cost = 0
+      s_size.times do |i|
+        last_cost = i + 1
+
+        t_size.times do |j|
+          sub_cost = s[i] == t[j] ? 0 : 1
+          cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
```
wontruefree: Is this the part that does the `addition` and `deletion` cost analysis?  I dont understand this part.

src/levenshtein.cr
```diff
@@ -13,36 +13,68 @@ module Levenshtein
...
+      reader = Char::Reader.new(string1)
+
+      # Use an array instead of a reader to decode the second string only once
+      chars = string2.chars
+
+      # This is to allocate less memory
+      if t_size > s_size
+        chars, reader = reader, chars
+        t_size = s_size
+      end
```
jhass, line 58: Shouldn't we make sure to call `chars` on the shorter string only?

r00ster91: Are you suggesting to change line 52 to this?
```cr
chars = if t_size > s_size
  string1.chars
else
  string2.chars
end
```

straight-shoota: I think that makes sense. Need to switch the reader too, though. 
It would probably make sense to switch `string1` and `string2` right at the beginning, after determining the size.

src/levenshtein.cr
```diff
@@ -13,36 +13,68 @@ module Levenshtein
...
+
+      costs = Pointer(Int32).malloc(t_size + 1) { |i| i }
+
+      last_cost = 1
```
straight-shoota, line 62: Why is `last_cost = 1` in this branch, when it is `0` in the other?

src/levenshtein.cr
```diff
@@ -13,36 +13,68 @@ module Levenshtein
...
 
     return t_size if s_size == 0
     return s_size if t_size == 0
 
-    # This is to allocate less memory
-    if t_size > s_size
-      t, s = s, t
-      t_size, s_size = s_size, t_size
-    end
+    if string1.ascii_only? && string2.ascii_only?
```
straight-shoota, line 22: This should check `single_byte_optimizable?` because `ascii_only?` iterates and check that every codepoint is actually an ASCII codepoint.

src/levenshtein.cr
r00ster91: > I think that makes sense. Need to switch the reader too, though.

Then I suppose it would become this:
```cr
if t_size > s_size
  chars = string1.chars
  reader = Char::Reader.new(string2)
else
  chars = string2.chars
  reader = Char::Reader.new(string1)
end
```
> It would probably make sense to switch `string1` and `string2` right at the beginning, after determining the size. So that `string1` is always the shorter one.

So `string1` takes the bytes of `string2` and `string2` takes the bytes of `string1`. How would you imagine this to be implemented? Something using `Pointer#swap`?

src/levenshtein.cr
r00ster91: Good question. I'm not quite sure anymore so I will probably change this back.

src/levenshtein.cr
straight-shoota: just
```cr
if t_size > s_size
  string1, string2 = string2, string1
end
```

src/levenshtein.cr
```diff
@@ -13,36 +13,60 @@ module Levenshtein
...
-
-    s_size = s.size
-    t_size = t.size
+    s_size = string1.size
+    t_size = string2.size
 
     return t_size if s_size == 0
     return s_size if t_size == 0
 
-    # This is to allocate less memory
```
straight-shoota, line 25: This line should be kept (or maybe even more elaborate)

straight-shoota (approved):

oprypin (approved):

asterite:
Can you try using Slice instead of Pointer and seeing if there's any performance drop?

oprypin:
@asterite The pointer stuff is pre-existing code, by the way -- just repeated twice in the new branches. That actually was going to be my 1 comment -- that line `costs = Pointer(Int32).malloc(t_size + 1) { |i| i }` could be written only once at the top, instead of twice.

r00ster91:
```cr
require "benchmark"

Benchmark.ips do |x|
  x.report("1") { distance1("漢字", "かんじ"); distance1("hello", "HELLO") }
  x.report("2") { distance2("漢字", "かんじ"); distance2("hello", "HELLO") }
end

def distance1(string1 : String, string2 : String) : Int32
  return 0 if string1 == string2

  s_size = string1.size
  t_size = string2.size

  return t_size if s_size == 0
  return s_size if t_size == 0

  # This is to allocate less memory
  if t_size > s_size
    string1, string2 = string2, string1
    t_size, s_size = s_size, t_size
  end

  if string1.single_byte_optimizable? && string2.single_byte_optimizable?
    s = string1.to_unsafe
    t = string2.to_unsafe

    costs = Slice(Int32).new(t_size + 1) { |i| i }

    last_cost = 0
    s_size.times do |i|
      last_cost = i + 1

      t_size.times do |j|
        sub_cost = s[i] == t[j] ? 0 : 1
        cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
        costs[j] = last_cost
        last_cost = cost
      end
      costs[t_size] = last_cost
    end

    last_cost
  else
    reader = Char::Reader.new(string1)

    # Use an array instead of a reader to decode the second string only once
    chars = string2.chars

    costs = Slice(Int32).new(t_size + 1) { |i| i }

    last_cost = 0
    reader.each_with_index do |char1, i|
      last_cost = i + 1

      chars.each_with_index do |char2, j|
        sub_cost = char1 == char2 ? 0 : 1
        cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
        costs[j] = last_cost
        last_cost = cost
      end
      costs[t_size] = last_cost
    end

    last_cost
  end
end

def distance2(string1 : String, string2 : String) : Int32
  return 0 if string1 == string2

  s_size = string1.size
  t_size = string2.size

  return t_size if s_size == 0
  return s_size if t_size == 0

  # This is to allocate less memory
  if t_size > s_size
    string1, string2 = string2, string1
    t_size, s_size = s_size, t_size
  end

  if string1.single_byte_optimizable? && string2.single_byte_optimizable?
    s = string1.to_unsafe
    t = string2.to_unsafe

    costs = Pointer(Int32).malloc(t_size + 1) { |i| i }

    last_cost = 0
    s_size.times do |i|
      last_cost = i + 1

      t_size.times do |j|
        sub_cost = s[i] == t[j] ? 0 : 1
        cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
        costs[j] = last_cost
        last_cost = cost
      end
      costs[t_size] = last_cost
    end

    last_cost
  else
    reader = Char::Reader.new(string1)

    # Use an array instead of a reader to decode the second string only once
    chars = string2.chars

    costs = Pointer(Int32).malloc(t_size + 1) { |i| i }

    last_cost = 0
    reader.each_with_index do |char1, i|
      last_cost = i + 1

      chars.each_with_index do |char2, j|
        sub_cost = char1 == char2 ? 0 : 1
        cost = Math.min(Math.min(last_cost + 1, costs[j + 1] + 1), costs[j] + sub_cost)
        costs[j] = last_cost
        last_cost = cost
      end
      costs[t_size] = last_cost
    end

    last_cost
  end
end
```
```
1   5.82M (171.81ns) (± 9.03%)  96.0B/op        fastest
2   5.62M (178.05ns) (± 7.41%)  96.0B/op   1.04× slower
```
I've changed it.
`to_unsafe` can not be removed though.

caspiano:
I tried with `to_slice` and got the following, where old is `to_unsafe` and new is `to_slice`

```
old 459.71  (  2.18ms) (± 9.33%)  195kB/op        fastest
new 435.59  (  2.30ms) (± 9.91%)  195kB/op   1.06× slower
```

r00ster91:
It does not seem to be noise though but rather a consistent slowdown. `1` is always the slice version here.
```console
$ crystal run --release ../a.cr
1   5.24M (191.00ns) (± 8.74%)  96.0B/op   1.03× slower
2   5.37M (186.34ns) (± 9.72%)  96.0B/op        fastest
$ crystal run --release ../a.cr
1   5.27M (189.82ns) (± 7.37%)  96.0B/op   1.02× slower
2   5.38M (185.84ns) (± 9.13%)  96.0B/op        fastest
$ crystal run --release ../a.cr
1   5.19M (192.85ns) (± 7.41%)  96.0B/op   1.07× slower
2   5.55M (180.04ns) (± 8.09%)  96.0B/op        fastest
$ crystal run --release ../a.cr
2   5.53M (180.86ns) (± 7.69%)  96.0B/op        fastest
1   5.29M (188.96ns) (± 7.04%)  96.0B/op   1.04× slower
$ crystal run --release ../a.cr
2   5.34M (187.09ns) (±12.38%)  96.0B/op        fastest
1   4.73M (211.50ns) (±12.11%)  96.0B/op   1.13× slower
$ crystal run --release ../a.cr
2   5.55M (180.18ns) (± 7.90%)  96.0B/op        fastest
1   5.23M (191.14ns) (± 8.05%)  96.0B/op   1.06× slower
```
Is that worth it?

oprypin (approved):

asterite:
Thanks for the Slice stuff. I didn't realize it was using pointers before. I just wanted to make sure we don't end up with undefined behavior in case the code has a bug.

PR #8326, opened 2019-10-14
Added --nodefaultlibs option to compiler

dburnsii:
Attempting to cross-compile in an environment like Yocto is problematic with the current method of searching for libraries because the system libraries are hard-coded and cannot be excluded. With Yocto, all libraries are provided in an alternate folder, allowing the build to be agnostic of any build host configuration issues. 

In an effort to not break existing builds, I've made changes to the compiler so that the `--nodefaultlibs` ([Utilizing the -nodefaultlibs GCC option](https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html)) option prevents the compiler from using potentially incorrect or incompatible libraries.


straight-shoota:
Previous discussion in #8046

src/compiler/crystal/command.cr
```diff
@@ -346,6 +346,11 @@ class Crystal::Command
         @error_trace = true
       end
 
+      opts.on("--nodefaultlibs", "Disable standard library locations") do
+        compiler.flags << "-nodefaultlibs"
```
Sija: Is this flag necessary?

dburnsii: In most cases, if we're trying to prevent the compiler from using default library paths, I would say yes. GCC for example will try and use the standard system libraries if this is not added as a flag. This goes slightly beyond my initial intention of taking out the hard-coded path by following the change through down to the linker.

RX14: This is setting `-D` flags for crystal, not linker flags.

bcardiff:
Should we go with `CRYSTAL_CONFIG_LINK_FLAGS` at build time and `CRYSTAL_LINK_FLAGS` so default libs can be overridden when bundling the compiler and via de wrapper script?

dburnsii:
The main issue I'm trying to overcome with this approach is to remove the hard-coded library paths. Existing setups should be fine, as they won't be passing this new flag. Just don't want to have to go through the steps of recompiling the compiler in order to use Crystal in Yocto.

ysbaddaden:
I agree with RX14: this approach is wrong. We shouldnt have default paths. I believe we can use system environment variables that LD will recognize and maybe add a CLI argument to add paths to search for (i.e. `-L path`). But not a CLI argument to remove default paths.

dburnsii:
So then, should we opt not to add the default paths if provided with `LIBRARY_PATH`, or something along those lines?

bcardiff:
@RX14 @ysbaddaden So regarding this you would prefer to

1. Replace `library_path = ["/usr/lib", "/usr/local/lib"]` with the values of `LIBRARY_PATH` env variable.
2. Adapt wrapper scripts on distribution packages in case `/usr/lib`, `/usr/local/lib` is missing from the `LIBRARY_PATH` and actually needed.
3. Do not have a way to embed during the compiler compilation default paths to be added.

Did I get that right?

bcardiff:
Also, **if `LIBRARY_PATH` is not set** we can keep the current fixed paths

RX14:
If `LIBRARY_PATH` isn't set then the linker will know the paths.

There's more things wrong with the linking support in crystal than this, (like the `pkg-config` support in `@[Link]`), but this is a good start.

RX14:
Setting `library_path` to be `ENV["LIBRARY_PATH"]` is unnecesary because the linker looks at `LIBRARY_PATH` itself.

RX14:
I've been thinking about the linking code, and I propose removing the `static: true` option for `@[Link]` annotations. Traditional linkers, and `pkg-config`, support linking executable statically or dynamically, but they do not support linking some libraries statically and some dynamically. The problem which appears when trying to accomplish this is libraries with dependencies.

When trying to link dynamically to a library with dependencies, the `.so` file contains all the dependencies, and there's no problem. When trying to link statically, the `.a` file doesn't contain any dependency information, and `pkg-config` is essentially required. However using `pkg-config --libs --static` is designed for use when *everything* is being linked statically. For example `pkg-config --libs --static libssl` will return `-lssl -lcrypto -ldl -pthread`, where what you *probably* meant was to attempt to link `libssl.a` and `libcrypto.a`, but not libdl, or libpthread, which are considered part of libc.

If we're fine with removing `@[Link("foo", static: true)]` in favour of `--static`, then the link implementation becomes trivial: first try `pkg-config --libs` (with `--static` if we're linking statically), then fall back to `-lfoo`. The only place where `static: true` is used in crystal right now is for linking libgc, which on linux we distribute as a static library anyway, unless people are using a distro specific package.

bcardiff:
@RX14 There I think we can discuss evolution of the `Link` flow outside this particular issue. There are some ideas I can't find reference right now. 

Since the lookup of static libraries today uses `library_path = ["/usr/lib", "/usr/local/lib"]` I think we should set it's value to `ENV["LIBRARY_PATH"]`. Maybe that will go if we redo how linking work.

To be able to express better if a static is allowed or is mandatory would be a win. I think. But also to allow customizing the linking without touching the source code. That requires some further design iteration outside this issue.

I will try to make the change for reading `LIBRARY_PATH` and fallback if not set to the current values as a migration path. That should unlock this story, right @dburnsii? The wrapper scripts will allow to remove the default by setting an empty `LIBRARY_PATH`, or set the right values for the host os.

But yes, there are stories to improve on the linking, but I think they can be handled later.

RX14:
"Design" doesn't help when you're limited by the limitations of the platform. Why keep tweaking this and fixing edge cases when we can simplify it and remove all the edge cases?

dburnsii:
@bcardiff , Yes, the only thing we need to move forward and bring this into Yocto, is to have the option NOT to include `/usr/lib` and `/usr/local/lib` when running `crystal build`. I am all for fixing this the right way, but it sounds like that might require quite a bit of rework. If we can get those hard-coded path removed, we can get a recipe going that doesn't need source code modifications.

bcardiff:
I think there is value in changing this, but that is a bigger story and requires some further thinking. In particular, the link attribute is good for defaults, but some platforms named the libs different, or requires some dependencies, so to have a complete solution we need to take into consideration how to override that. 

I've found the topic in the forum I mean before: https://forum.crystal-lang.org/t/rfc-link-configuration/546 . We can iterate ideas there. 

Yet, I think there is value to unlock this scenario without revamping the whole link story.

oprypin:
The solution here should be to not apply `"/usr/lib", "/usr/local/lib"` at all (other than OSs that have no default themselves). `ld` itself already uses `LIBRARY_PATH` and has these defaults. There's no need for a flag, the changed behavior will always work on Linux.

PR #8327, opened 2019-10-14
[Compiler] Insert documentation for pseudo methods

### LINKED ISSUE #3041: Document pseudo methods in the API documentation
Namely, the following "methods" are missing in the API docs:
- `Object#is_a?`
- `Object#as`
- `Object#as?`
- `Object#responds_to`
- `Object#nil?`

They appear to be methods and not syntax, so people expect and should be able to find them documented in the API docs.

Just adding stub definitions to `src/object.cr` doesn't work, because redefinition is disallowed. So we either have to
- Add a special case in the documentation generator, detecting that it's generating for Crystal itself and injecting the definitions.
- Allow redefinitions of them in `Object` only. Subclasses would still get the error message, when somebody reopens `Object` to redefine them, just nothing happens.
- Allow redefinitions of them in the documentation generator only. That would allow to add a special file with stub definitions like we do for macro methods and types.

straight-shoota:
This adds documentation for pseudo-methods `!`, `as`, `as?`, `is_a?`, `nil?` and `responds_to?`, all on `Object`.
These methods are implemented by the compiler and cannot be redefined, so we can't simply insert them into code. The documentation is injected by the doc generator when it detects running in the crystal repo.

Fixes #3041


src/compiler/crystal/tools/doc/crystal_pseudo_methods.cr
```diff
@@ -0,0 +1,149 @@
...
+      doc: <<-DOC)
+        Returns `self`.
+
+        The type of this expression is restricted to *type* by the compiler.
+        *type* must be a constant, it cannot be evaluated at runtime.
+
+        If *type* is not a valid restriction for the expression type, it
+        is a compile-time error.
+        If *type*  is a valid restriction for the expression, but `self` can't
+        be restricted, it raises at runtime.
```
bew:
```suggestion
        be restricted to *type*, it raises at runtime.
```

src/compiler/crystal/tools/doc/crystal_pseudo_methods.cr
```diff
@@ -0,0 +1,149 @@
...
+        (`!foo`) but it can also be written as a regular method call (`foo.!`).
+        DOC
+
+    object.add_def declare_pseudo_method("is_a?",
+      args: [
+        Arg.new("type", restriction: Path.new(["Class"])),
+      ],
+      return_type: Path.new(["Bool"]),
+      doc: <<-DOC)
+        Returns `true` if  inherits or includes *type*.
```
asterite:
```suggestion
        Returns `true` if  `self` inherits or includes *type*.
```

bew: Note: there's 2 spaces after the `if`, `self` can go right in the middle

straight-shoota:
I don't know what's going on in CI. It keeps getting stuck. This seems like an unrelated issue though. Anyone noticed something similar?

straight-shoota:
Sure, now for the demonstration everything goes trough... 🙄

bcardiff:
Wouldn't be cleaner to have add this docs to regular .cr file that is included in `docs_main.cr` but not in `prelude`.

Since methods declaration can't have the name of a pseudo method we could use a prefix like `__crystal_pseudo_`, hence declaring `def __crystal_pseudo_as`, `def __crystal_pseudo_!` etc.

The doc tool can strip those prefix if run in the crystal repo.

The bonus point is that adding documentation seems easier and more comfortable if done in regular .cr files plus we can get the awesome example docs tool from @maiha to also check these docs straight away.

straight-shoota:
@bcardiff Yeah, that would work. And it's shouldn't be difficult to implement.

src/compiler/crystal/tools/doc/generator.cr
```diff
@@ -2,7 +2,7 @@ class Crystal::Doc::Generator
   getter program : Program
 
   @base_dir : String
-  @is_crystal_repo : Bool
+  property is_crystal_repo : Bool
```
bew:
```suggestion
  getter is_crystal_repo : Bool
```
Should be enough?

bew:
Did you forgot to add the file `docs_pseudo_methods.cr` ?

straight-shoota:
Ooops.

src/compiler/crystal/tools/doc/generator.cr
straight-shoota: No, the setter is used in `generator_spec`

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,134 @@
...
+  # !nil    # => true
+  # !1      # => false
+  # !"foo"  # => false
+  # ```
+  #
+  # This method is a unary operator and usually written in prefix notation
+  # (`!foo`) but it can also be written as a regular method call (`foo.!`).
+  #
+  # NOTE: This is a pseudo-method provided directly by the Crystal compiler.
+  # It cannot be redefined nor overridden.
```
bew: This could be added by the doc generator?

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,134 @@
...
+  # a.as(Int32)           # => 1
+  #
+  # typeof(a.as(Bool))    # Compile Error: can't cast (Int32 | String) to Bool
+  #
+  # typeof(a.as(String))  # => String
+  # a.as(String)          # Runtime Error: cast from Int32 to String failed
+  #
+  # typeof(a.as(Int32 | Bool)) # => Int32
+  # a.as(Int32 | Bool)         # => 1
+  # ```
```
bew: Might be worth to add that you can use `as(typeof(...))` too!

small example:
```cr
a : Int32? = 1
b = a.as(typeof(a.not_nil!))
typeof(b) # => Int32
```

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,134 @@
...
+  #
+  # typeof(a.as?(Bool))    # => Nil
+  # a.as?(Bool)            # => nil
+  #
+  # typeof(a.as?(String))  # => String | Nil
+  # a.as?(String)          # nil
+  #
+  # typeof(a.as?(Int32 | Bool)) # => Int32 | Nil
+  # a.as?(Int32 | Bool)         # => 1
+  # ```
```
bew: ditto

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,134 @@
...
+  #
+  # NOTE: This is a pseudo-method provided directly by the Crystal compiler.
+  # It cannot be redefined nor overridden.
+  def __crystal_pseudo_as?(type : Class)
+  end
+
+  # Returns `true` if `self` responds to *name*.
```
bew: It might help more to say something along the lines:
``Returns `true` if the method *name* can be called on `self` ``

asterite:
Really nice!

Some other pseudo-methods (in `top-level`):
- `typeof`
- `sizeof`
- `instance_sizeof`
- `pointerof`
- `offsetof`

src/docs_pseudo_methods.cr
straight-shoota: What's the purpose? You can just use `a.not_nil!` directly.

bew: Sure, it's just an example of a "Non-compile-time-looking" call, the example is silly I agree

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,197 @@
...
+  # Returns `self`.
+  #
+  # The type of this expression is restricted to *type* by the compiler.
+  # *type* must be a constant, it cannot be evaluated at runtime.
+  #
+  # If *type* is not a valid restriction for the expression type, it
+  # is a compile-time error.
+  # If *type*  is a valid restriction for the expression, but `self` can't
+  # be restricted to *type*, it raises at runtime.
+  # *type* may be a wider restriction than the expression type, the resulting
```
bcardiff: Not entirely true (or there is actually a bug :-) )

```
a : Int32
typeof(a.as(Int32 | String)) # => Int32 | String (yet the value is Int32)

b : Int32 | String
typeof(b.as(Int32 | Float64)) # => Int32
typeof(b.as(Int32 | Float64).as(Int32 | Float64)) # => Int32 | Float64
```

https://play.crystal-lang.org/#/r/7u51

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,197 @@
...
+  # typeof(a.as?(Int32)) # => Int32 | Nil
+  # a.as?(Int32)         # => 1
+  #
+  # typeof(a.as?(Bool)) # => Nil
+  # a.as?(Bool)         # => nil
+  #
+  # typeof(a.as?(String)) # => String | Nil
+  # a.as?(String)         # nil
+  #
+  # typeof(a.as?(Int32 | Bool)) # => Int32 | Nil
```
bcardiff: as per https://github.com/crystal-lang/crystal/pull/8327/files#r336605343 this might need to be 💭 

```
  # typeof(a.as?(Int32 | Bool)) # => Int32 | Bool | Nil
```

src/docs_pseudo_methods.cr
asterite: You can use `as` to extend the type of an expression from a single type to a union type. The `b` case is a bit strange but I guess it also makes sense (check that it's an `Int32` and then extends it to `Int32 | Float64`, raise otherwise).

bcardiff: What is weird is that 

* `expr.as(T) : T` does not hold
* `expr.as?(T) : T?` does not hold
* `expr.as(T) == expr.as(T).as(T)` does not hold

of course if `expr` was a value that is not compatible with `T` then raise / nil is expected. But I don't see why the above rules cannot hold


It is definitely something beyond the documentation. But I picked from the phrases and the examples 🕵

asterite: Oh, yes, `b.as(Int32 | Float64)` should give `Int32 | Float64`. It's a bug. Sorry, I missed that.

Actually, I don't think casting between union types like that should be allowed, it's strange.

src/docs_pseudo_methods.cr
straight-shoota: The current return value is `Int32 | Nil`. If we want to change that, we can change the documentation later.

straight-shoota:
I suppose `||` and `&&` are essentially also pseudo methods?

They're currently not even properly recognized:
```cr
def ! # '!' is a pseudo-method and can't be redefined
end

def && # expecting any of these tokens: IDENT, CONST, `, <<, <, <=, ==, ===, !=, =~, !~, >>, >, >=, +, -, *, /, //, !, ~, %, &, |, ^, **, [], []?, []=, <=>, &+, &-, &*, &** (not '&&')
end

def || # expecting any of these tokens: IDENT, CONST, `, <<, <, <=, ==, ===, !=, =~, !~, >>, >, >=, +, -, *, /, //, !, ~, %, &, |, ^, **, [], []?, []=, <=>, &+, &-, &*, &** (not '||')
end
```

I think they should be treated in the same way as the already documented pseudo methods. That means adding documentation and compiler support for parsing + appropriate error message.

bcardiff:
We can encode somehow the not allowed methods and decode after removing the `__crystal_pseudo_` prefix.

straight-shoota:
I actually didn't think that far that documenting these methods also needs compiler support. I was primarily concerned about the compiler message `'||' is a pseudo-method and can't be redefined` instead of `expecting any of these tokens:`.

Technically we could probably allow anything after `__crystal_pseudo_` as part of a method name, except for whitespace, semicolon and open parenthesis.

asterite:
and and or and not methods. That doesn't work like that in Ruby. But we could make it look like that if you think it's good.

straight-shoota:
Oh right, I thought you could call `foo.&&(bar)`. Yeah, I guess that doesn't need to be changed...

So `&&` and `||` are the only operators that are not implemented as methods (or pseudo methods).

asterite:
Right, though I wouldn't mind making it look like they do (same we do for `!` and any "operator")

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,197 @@
...
+# ```
+#
+# The expressions passed as arguments to `typeof` do not evaluate. The compiler
+# only analyzes their return type.
+def __crystal_pseudo_typeof(*expression) : Class
+end
+
+# Returns the size of the given type as number of bytes.
+#
+# *type* must be a constant, it cannot be evaluated at runtime.
```
RX14: Perhaps mention that `sizeof(typeof())` works to avoid confusion here?

src/docs_pseudo_methods.cr
```diff
@@ -0,0 +1,197 @@
...
+# For `Reference` types, the size is the same as the size of a pointer:
+#
+# ```
+# # On a 64 bits machine
+# sizeof(Pointer(Int32)) # => 8
+# sizeof(String)         # => 8
+# ```
+#
+# This is because a `Reference`'s memory is allocated on the heap and a pointer
+# to it is passed around. The effective size of a class can be determined using
```
RX14: The size of a class on the heap can be determined?

PR #8330, opened 2019-10-15
[compiler] Fix detect musl libc at runtime

### LINKED ISSUE #7926: http/client fails running in Alpine Linux 3.10 docker image
I posted about this on the [forum](https://forum.crystal-lang.org/t/utf8-issues-with-alpine-linux/888) but so far no response.

I've only seen the issue on Alpine Linux.  My [test case](https://github.com/horrendo/t-alpine)  runs fine on debian and OS/X.

straight-shoota:
The output from `ldd --version` is used to determine whether the default target libc is musl or gnu.

On alpine 3.10, the output from `ldd --version` is broken which leads to incorrect detection as gnu.
This regression bug in has been fixed in alpine edge (https://gitlab.alpinelinux.org/alpine/aports/commit/2f2f995d9d52484720dbe8b4b1649c132ec1feee) but it's still in 3.10. This PR fixes that by recognizing the alternative (error) output to successfully detect musl. It's not super important but a small improvement for a specific case.

The second commit adds specs for `Crystal::Config.default_target` which could help detect such issues earlier (assuming we get CI running on musl some time #6943).

Fixes #7926, #8330

### DIFF
--- a/spec/compiler/config_spec.cr
+++ b/spec/compiler/config_spec.cr
@@ -1,0 +1,14 @@
+require "../spec_helper"
+require "./spec_helper"
+
+describe Crystal::Config do
+  it ".default_target" do
+    Crystal::Config.default_target.should eq Crystal::Codegen::Target.new({{ `crystal --version`.lines[-1] }}.lstrip("Default target: "))
+  end
+
+  {% if flag?(:linux) %}
+    it ".linux_runtime_libc" do
+      Crystal::Config.linux_runtime_libc.should eq {{ flag?(:musl) ? "musl" : "gnu" }}
+    end
+  {% end %}
+end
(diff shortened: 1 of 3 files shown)


spec/compiler/config_spec.cr
```diff
@@ -0,0 +1,14 @@
+require "../spec_helper"
+require "./spec_helper"
+
+describe Crystal::Config do
+  it ".default_target" do
+    Crystal::Config.default_target.should eq {{ `crystal --version`.lines[-1] }}.lstrip("Default target: ")
```
Sija:
```suggestion
    Crystal::Config.default_target.to_s.should eq {{ `crystal --version`.lines[-1] }}.lstrip("Default target: ")
```

ysbaddaden: `Crystal::Config.default_target` is already a String.

Sija: nope, it's `Crystal::Codegen::Target`.

PR #8332, opened 2019-10-15
Add Enumerable#each_cons_pair and Iterator#cons_pair yielding a tuple

straight-shoota:
`Enumerable#each_cons` has a variable chunk size that can be selected dynamically.
This PR adds a special case for the most common use case of iterating over pairs. It simply yields tuples of adjacent items. This avoids heap allocations and makes it easier to use because you don't need to deconstruct an array.

```cr
# new:
[1, 2, 3, 4, 5].each_cons do |a, b|
  puts a * b
end
# old:
[1, 2, 3, 4, 5].each_cons(2, reuse: true) do |pair|
  a, b = pair
  puts a * b
end
```

Benchmark:
```
# iterating over (1..1000).to_a
new 228.27k (  4.38µs) (±17.91%)   0.0B/op        fastest
old  33.13k ( 30.19µs) (± 9.87%)  80.0B/op   6.89× slower
```

It's a convenience feature that also improves performance.

I've added the same behaviour to `Iterator#cons`

### DIFF
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -320,6 +320,14 @@ describe "Enumerable" do
     end
   end
 
+  describe "#each_cons_pair" do
+    it "returns running pairs" do
+      array = [] of Array(Int32)
+      [1, 2, 3, 4].each_cons_pair { |a, b| array << [a, b] }
+      array.should eq([[1, 2], [2, 3], [3, 4]])
+    end
+  end
+
   describe "each_slice" do
     it "returns partial slices" do
       array = [] of Array(Int32)
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -235,84 +235,88 @@ describe "Enumerable" do
     end
   end
 
-  describe "each_cons" do
-    it "returns running pairs" do
-      array = [] of Array(Int32)
-      [1, 2, 3, 4].each_cons(2) { |pair| array << pair }
-      array.should eq([[1, 2], [2, 3], [3, 4]])
-    end
-
-    it "returns running triples" do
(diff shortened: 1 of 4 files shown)


asterite:
I like this!

What if we name it `each_cons2`? Then we could eventually add `each_cons3`, `each_cons4`, etc. if we really wanted to.

Though it's true that one of the most common cases for `each_cons` is 2... 🤔

straight-shoota:
I don't see a reason for adding a number suffix. Iterating over pairs is by far the most common use case. If we wanted to add others, there's nothing hindering us from adding `each_cons3` later anyway.

RX14:
Want to rebase for a merge commit?

vlazar:
I love this optimization but I feel like the chosen size `2` mentioned explicitly in the method name like `#each_cons_pair` / `Iterator#cons_pair` would be even better.

bcardiff:
I like @vlazar's suggestion. Otherwise, the expected block arguments differ based on the presence or not of the argument. I like the explicitness of `each_cons_pair`

straight-shoota:
I guess that might be be good enhancement, but waiting for feedback on this suggestion.

asterite:
I think `each_cons_pair` is good. My suggestion `each_cons2`, `each_cons3`, etc. might also be taken into consideration, though I doubt we'll spend effort adding each of these other special variants, so 👍  `each_cons_pair` from me.

RX14:
+1 from me, might want to go ahead with this

bcardiff:
`each_cons_pair` or `each_cons2` should be used as a name. Other than that LGTM.

straight-shoota:
Renamed to `Enumerable#each_cons_pair` and `Iterator#cons_pair`.

Issue #8307, opened 2019-10-11
[RFC] C bindings helper


asterite:
I recently stumbled upon a tool called [hsc2hs](https://downloads.haskell.org/~ghc/7.6.3/docs/html/users_guide/hsc2hs.html): it's a template language for Haskell to easy writing C bindings.

## How it would work

The idea applied to Crystal would be something like this:

### You write a .crc file with some directives

```crystal
# pcre.crc

<%= include "pcre.h" %>

lib LibPCRE
  enum Options
    CASELESS = <%= const PCRE_CASELESS %>
  end
end
```

Here we just have two of the most simple directives:
- `include` will include a C header file
- `const` will output the value of a C `#define` or constant

### The crc generates a C file that, when executed, produces a Crystal file

The C file generated by the program above will look like this:

```c
// pcre.crc.c
// This line because of our "include" directive
#include "pcre.h"
#include <stdio.h>

int main(int argc, char** argv) {
  printf("lib LibPCR\n");
  printf("  enum Options\n");
  printf("    CASELESS =");
  printf(PCRE_CASELESS); // this line because of the "const" directive
  printf("\n");
  printf("  end\n");
  printf("end\n");
  return 0;
}
```

### We run the C file and pipe the output to the final Crystal file

```
$ clang pcre.crc.c -o some_temp_name
$ ./some_temp_name > pcre.cr
```

The generated file will look like this:

```crystal
# pcre.cr
lib LibPCRE
  enum Options
    CASELESS = 1
  end
end
```

We got the `1` directly from the pcre.h header file! :-)

## Thoughts

I think the idea of hsc2hs is brilliant: instead of trying to automatically generate bindings from a C header, like we tried to do in [crystal_lib](https://github.com/crystal-lang/crystal_lib), which is very complex, we just get what we really need from C headers.

The things we could get from C headers are:
- the value of #define and constants
- `sizeof` structs
- `alignment` of structs (maybe we don't need this)
- `offsetof` struct fields: with just this, we can write and read any value from a struct by casting the struct pointer to a `Pointer(UInt8)`, adding the offset, then casting the pointer to the type of the struct field and fetching the value from it. Of course it works too if we have a plain struct (not a pointer of a struct) and we use `pointerof`.

The above means we can bind to any C struct:
- we represent the struct as `UInt8[N]` where `N` is the `sizeof` the struct.
- to read or write a value we use `offsetof` and use pointers and casting

[...]

PR #8336, opened 2019-10-15
Proof of concept: C bindings helper

asterite:
**This PR is just a proof of concept: there are no tests, the code might be ugly or unfinished**

PoC for #8307

With this PR you can now have `.ccr` files (for C+Crystal file).

A simple example:

```crystal
# file.ccr
alias SizeT = <%= type size_t %>

puts SizeT
```

As explained in #8307, this will generate a C file that, when compiled and ran, will generate a Crystal file where `type size_t` will be replaced with, in the case of my 64 bits machine, `UInt64`.

Similarly we can do this:

```crystal
# file.ccr
puts <%= sizeof size_t %>
```

That `sizeof size_t` will get replaced by `8`.

There are other directives here:
- offsetof: to get the offset of a struct field
- const: to get the value of a constant

An example for `const`:

```crystal
# file.ccr
<%= include "pcre.h" %>

p <%= const PCRE_CASELESS %>
```

This will turn `const PCRE_CASELESS` to `1`. The `include` directive will `#include "pcre.h"` so we have that constant available.

For now the generated `cr` and `c` files are kept next to the original file. It's pretty fast so we could simply ignore those in `.gitignore`, and eventually do something smarter.

What do you think?


bcardiff:
I think one important use case is to check a version constant defined in the header and based on its value, use or not other constants.

Together with compiler flags this means having some sort of `{% if flags?` and `compare_versions` to match the `.cr` ones that should be translated to C and maybe evaluating them eagerly to avoid generating C code that will not compile.

src/compiler/crystal/crystal_path.cr
```diff
@@ -127,6 +140,8 @@ module Crystal
         else
           if filename.ends_with?(".cr")
             files << full_name
+          elsif filename.ends_with?(".ccr")
+            files << CCR.process(full_name)
```
Sija:
```cr
case ::Path[filename].extension
when ".cr"  then files << full_name
when ".ccr" then files << CCR.process(full_name)
end
```

asterite:
I just pushed some code that adds support for `if`, `ifdef`, `ifndef`, `else`, `elif` and `endif`.

So for example now you can do:

```crystal
<%= include "sys/resource.h" %>

<%= ifdef RUSAGE_SELF %>
  puts "Has getrusage"
<%= else %>
  puts "Doesn't have getursage"
<%= endif %>
```

This would be very helpful to support systems that don't have `getrusage` and use `times` in that case.

@bcardiff Is that what you meant?

One can of course do this:

```crystal
<%= ifdef RUSAGE_SELF %>
  HAS_RUSAGE = true
<%= else %>
  HAS_RUSAGE = true
<%= endif %>

# Now we can access HAS_RUSAGE in Crystal at compile-time
{% if HAS_RUSAGE %}
{% end %}
```

But `ifdef` is necessary to define the Crystal constant.

asterite:
I also forgot to say that the generated `.cr` files has `loc` pragmas to point back to the template file. This is similar to what ECR does. So when you get an error in that `.cr` file, the error will actually point to the template file. This makes it super transparent for the user.

asterite:
The next piece of the puzzle, and what might be most useful, is to be able to specify a Crystal struct's size and the offset of each field.

So for example right now we have https://github.com/crystal-lang/crystal/blob/master/src/ext/sigfault.c and we had to write it in C because, if I remember correctly, it was hard to map the `struct sigaction` because the fields, like `sa_sigaction `, were actually macros that expanded to other things, so it's not clear at which byte offset these things are located, and they might change between systems.

If we had a way to say:

```crystal
@[Sizeof(<%= sizeof struct sigaction %>)]
struct Sigaction
  @[Offsetof(<%= offsetof struct sigaction, ss_sp %>)]
  sa_sigaction : Void*
end
```

then Crystal will know:
- whether it needs to add padding at the end to fullfill the struct's bytesize
- how to arrange fields and where to put them in the struct

That way when you need to map a struct but you are only interested in some of the fields, using `Sizeof` and `Offsetof` is all you need to make it work in a portable way.

I'll try to work on those two annotations when I have time.

src/compiler/crystal/crystal_path.cr
asterite: Not really an optimization, getting the extension needs to allocate a string, whereas `ends_with?` is just a super fast memcmp.

Also, I wouldn't worry right now about how clean or optimal the code is.

asterite:
I just pushed a couple of commits that implement the `Sizeof` and `Offsetof` idea.

I believe with just this we have something really, really good. For example now you can bind any C struct and be sure you'll get its size right and whatever field you want (which is all that matters to C).

To try this out, I decided to rewrite our existing [sigfault.c](https://github.com/crystal-lang/crystal/blob/cb1ec832944752b69773b645600b70ee25cf28b6/src/ext/sigfault.c) in ccr:

```crystal
<%= include <signal.h> %>
<%= include <stdio.h> %>

<%= include <unistd.h> %>
<%= include <stdlib.h> %>

lib LibC
  # Here we map the siginfo_t type
  # We tell it the size of the struct
  @[Sizeof(<%= sizeof siginfo_t %>)]
  struct SiginfoT
    # We are also interested in this field so we add it and tell the compiler
    # in which byte offset it's located
    @[Offsetof(<%= offsetof siginfo_t, si_addr %>)]
    si_addr : Void*
  end

  # Same goes for stack_t
  @[Sizeof(<%= sizeof stack_t %>)]
  struct StackT
    @[Offsetof(<%= offsetof stack_t, ss_sp %>)]
    ss_sp : Void*
    @[Offsetof(<%= offsetof stack_t, ss_size %>)]
    ss_size : <%= type size_t %>
    @[Offsetof(<%= offsetof stack_t, ss_flags %>)]
    ss_flags : <%= type int %>
  end

  # And for struct sigaction
  @[Sizeof(<%= sizeof struct sigaction %>)]
  struct Sigaction
    @[Offsetof(<%= offsetof struct sigaction, sa_flags %>)]
    sa_flags : <%= type int %>
    @[Offsetof(<%= offsetof struct sigaction, sa_sigaction %>)]
    sa_sigaction : Void*
    @[Offsetof(<%= offsetof struct sigaction, sa_mask %>)]
    sa_mask : SigsetT
  end

  # Note that we don't have to map out the entire struct, which can be very
  # cumbersome: we can just map what we need, when we need it

  # Let's also fetch some `#define` values directly from c
  SA_ONSTACK = <%= const SA_ONSTACK %>
  SA_SIGINFO = <%= const SA_SIGINFO %>
  SIGSTKSZ = <%= const SIGSTKSZ %>

  # Mapping functions is done as usual, but we can use `type` to get the
  # Crystal equivalent of some integer/float primitives
  fun sigaltstack(ss : StackT*, oss : StackT*) : <%= type int %>
  fun sigemptyset(set : SigsetT*) : <%= type int %>
  fun sigaction(sig : Int, act : Sigaction*, oact : Sigaction*) : <%= type int %>
end

altstack = LibC::StackT.new
action = LibC::Sigaction.new

altstack.ss_sp = LibC.malloc(LibC::SIGSTKSZ)
altstack.ss_size = LibC::SIGSTKSZ
altstack.ss_flags = 0
LibC.sigaltstack(pointerof(altstack), nil)

LibC.sigemptyset(pointerof(action.@sa_mask))
action.sa_flags = LibC::SA_ONSTACK | LibC::SA_SIGINFO
action.sa_sigaction = ->(sig : LibC::Int, info : LibC::SiginfoT*, ctx : Void*) {
  addr = info.value.si_addr

  stack_top = Pointer(Void).new(Fiber.current.@stack.address - 4096)

  if stack_top <= addr < Fiber.current.@stack_bottom
    LibC.dprintf 2, "Stack overflow! (e.g., infinite or very deep recursion)\n"
  else
    LibC.dprintf 2, "Invalid memory access! (signal %d) at address 0x%lx\n", sig, addr
  end

  CallStack.print_backtrace
  LibC._exit(sig)
}.pointer

LibC.sigaction(LibC::SIGSEGV, pointerof(action), nil)
LibC.sigaction(LibC::SIGBUS, pointerof(action), nil)

def foo
  foo
end

foo
```

It works! We get a "Stack overflow! (e.g., infinite or very deep recursion)" error message. It works exactly like the C equivalent.

The best thing: it's portable. It will work on any machine, and if such C code doesn't compile on a machine we'll know right away (and then we can use `<%= ifdef %>` or another strategy).

It's also incredibly fast: compiling such a small C program is almost instantaneous. It would be great if the compiler could cache the `.c` file and the generated `.cr`, but even without that optimization it's still pretty fast.

The code is a bit messy and it's not DRY, but we could also introduce a directive to map a struct field. For example:

```crystal
@[Sizeof(<%= sizeof siginfo_t %>)]
struct SiginfoT
  <%= field siginfo_t, si_addr : Void* %>
end
```

would be equivalent to generating the field definition together with the `Offsetof` annotation.

And finally we could introduce a `struct` directive:

```crystal
<%= struct StackT, stack_t,
       ss_sp : Void*,
       ss_size : SizeT,
       ss_flags : Int %>
```

which would generate the entire `struct` definition, together with `Sizeof` and `Offsetof` fields.

You still have to map those fields, and map functions, but the tedious parts (making sure you get the right size, the right offsets, and sometimes the right primitive types) is solved.

I'll stop this PoC now until we reach a conclusion on whether we want to continue this route or not.

wontruefree:
This is really fascinating I think I can use this in some of the work I am doing.

src/compiler/crystal/syntax/ast.cr
```diff
@@ -254,7 +264,6 @@ module Crystal
       when UInt8   then :u8
       when UInt16  then :u16
       when UInt32  then :u32
-      when UInt64  then :u64
```
Sija: Why?

asterite: Oops! Mistake

src/compiler/crystal/semantic/type_declaration_visitor.cr
```diff
@@ -195,6 +216,30 @@ class Crystal::TypeDeclarationVisitor < Crystal::SemanticVisitor
...
+
+    value =
+      begin
+        arg.integer_value.to_i
+      rescue OverflowError
+        arg.raise "argument to Offsetof must fit in Int32"
+      end
+
+    if value < 0
+      arg.raise "argument to Offsetof must be positive"
```
Sija:
```suggestion
      arg.raise "argument to Offsetof must be positive or zero"
```

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -509,10 +515,35 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
...
+
+    value =
+      begin
+        arg.integer_value.to_i
+      rescue OverflowError
+        arg.raise "argument to Sizeof must fit in Int32"
+      end
+
+    if value < 0
+      arg.raise "argument to Sizeof must be positive"
```
Sija: I'm not sure about this:

```suggestion
      arg.raise "argument to Sizeof must be positive or zero"
```

j8r:
Shouldn't be pluggable to the compiler, like the Haskell implementation?
Being a POC, it would be better for experimentation.

oprypin:
I was thinking, isn't `<%=` too clunky of a syntax? How was it chosen?
As I understand, the `=` usually stands for "insert the expression" but then why is it also used for `include` and such?

asterite:
I chose `<%= ... %>` because it's similar to other template languages, even ECR.

There's only `<%= ... %>` and no `<% ... %>` because `<%= include %>` will append an include to the C file. So every `<%= ... %>` will append something to the C file, but not necessarily to the CR file.

However, I'm open to suggestions because I'm also not sure about that syntax.

hsc2hs uses `#...` but it's a bit strange because to put it in the middle of an expression you have to use parentheses, like `(#...)` and I think that complicates parsing a bit.

Otherwise, we can have `<% include ... %>` and other directives that output to the C file but not to the generated CR files.

Since this is just a POC I didn't put much thought about the syntax.

bcardiff:
Regarding https://github.com/crystal-lang/crystal/pull/8336#issuecomment-542476378,

The usecase I am thinking of is:

* libX-dev defines `VERSION` in `x.h`
* The binding nows that for `LibX` >= 2.0.0 there is a new symbol `my_new2_fun` (with a new struct `my_new2_result` as a return type) and constants defined `MY_NEW2_DEF`.

Currently we use some bash commands to find the current library version (eg: libssl) and define `X_VERSION` in Crystal's `LibX`. From there we use `compare_versions` to port the value of `MY_NEW2_DEF` and declare `my_new2_fun`. If the version is lower than 2.0.0 we just omit things.

The conditional declaration in crystal lib are used by macros: `compare_versions` and `flag?` are the most used ones though.


So I would like to be able to write the following to use the macros semantic in crystal mainly to use `compare_versions` and `flag?`

```crystal
<%= include "x.h" %>

lib LibX
  X_VERSION = <%= const VERSION %>

  {% if compare_versions(X_VERSION, "2.0.0") >= 0 %}
    MY_NEW2_DEF = <%= const MY_NEW2_DEF %>

    @[Sizeof(<%= sizeof struct my_new2_result %>)]
    struct my_new2_result
        # ...
    end

    fun my_new2_fun(ss : StackT*, oss : StackT*) : my_new2_result
  {% end %}
end
```

But evaluating that is a bit more difficult probably.

So we will probably need to write

```crystal
<%= include "x.h" %>

lib LibX
  X_VERSION = <%= const VERSION %>

  <%= if X_MACRO_TO_CHECK_VERSION_GTE("2.0.0") %>
    MY_NEW2_DEF = <%= const MY_NEW2_DEF %>

    @[Sizeof(<%= sizeof struct my_new2_result %>)]
    struct my_new2_result
        # ...
    end

    fun my_new2_fun(ss : StackT*, oss : StackT*) : my_new2_result
  <% endif %>
end
```

which is not terrible, but, there is still no way to use compiler flags. Maybe some `#define`s can be injected to port those. There we will cover the main usages of conditional libraries I think.

bcardiff:
On the other hand, the cross-compilation story becomes way more complex I think.
It is not clear how we can simulate the target context of the library in the origin host, or how we can split the compilation in phases to:

1. generate the .c files, 
2. build them in target, 
3. bring back the .cr files.
4. use .cr files instead of the .ccr files.

asterite:
> But evaluating that is a bit more difficult probably.

I see. We can't mix `#if` from C and `{% if ` from Crystal because some constants might not be defined in older versions of the library.

However, I think that one can conditionally check the version in C using `#if` or `#ifdef`.

In `hsc2hs` one can also do `#define` in the template file and that will be appended in the generated C file. So you could define a macro to check for versions, or we could even provide a macro in the generated C file by default.

The cross compilation is problematic. However, I believe cross compilation is mainly needed for the compiler, to port it to other platforms. That said, I like your 4 points approach. Maybe we just need a new flag for cross compilation that tells the compiler to search for headers for `ccr` in some directory. Then you can put include headers for that target in that directory.

If we want to move forward with this, I think that if we don't use `ccr` in the compiler, or at least not in libraries used by the compiler (we could for example use it for libxml2) then we'll be fine, until we sort out the cross-compilation issue.

ysbaddaden:
@asterite Cross compilation can be used to compile for a slow target from a much faster host (or one with enough memory). For example cross-compiling Crystal itself in release mode on a macOS host for an AArch64 target with 2GB of memory.

It can be achieved using the regular C/C++ method: install a full featured cross compilation toolchain and export `CC` to use it. See https://toolchains.bootlin.com/ for example. As a bonus it allows building an executable binary, not just an object file to eventually link on the target. BTW: maybe `--cross-compile` should be deprecated in favor of a generic `--emit=<executable|object|llvm>`?

Having a tool to generate the `.c` files, sending them to the target to compile/run and downloading the generated `.cr` files back into a local cache folder, could also do the trick.

**I believe this POC is a great approach**. Weirdly enough, it could simplify porting Crystal to unsupported targets, by avoiding the need to write/generate libc bindings beforehand, because CCR will just do it.

@bcardiff We should use the library's solution to compare versions, which is a good thing since this is C land, not Crystal. For example:
```c
#if OPENSSL_VERSION >= 0x0x10101000 %>
#endif

#if LIBXML_VERSION < 0x20900 %>
#endif

#if LLVM_VERSION_MAJOR < 8
#endif
```

RX14:
I don't think it makes sense to replace sigfault.c with ccr, since you're essentially replacing a C program with a more complex C program.

Using ccr as a AoT pregenerator for the libc bindings would also be great. But I see this as a tool which should exist outside this repository, and not be required to compile itself. Adding another bootstrapping problem for crystal is not a good idea.

Since openssl isn't needed to cross-compile the compiler, I'm perfectly fine with using ccr there.

asterite:
> since you're essentially replacing a C program with a more complex C program

No, we are replacing a C program with a Crystal program. The C program is just an intermediate step to get to the Crystal program.

RX14:
> No, we are replacing a C program with a Crystal program. The C program is just an intermediate step to get to the Crystal program.

Depends if you check the generated code into the repo or not. If you don't, there's no benefit.

asterite:
> Depends if you check the generated crystal code into the repo or not. If you don't, there's no benefit.

The generated code will never be committed. The idea is to generate it on the fly on your machine to get maximum portability.

vlazar:
Re-discovered this PR and I like it! Maybe I'm getting old, but simplicity attracts me more and more.

asterite:
I think having this open only introduces noise. I'll close this for now.

PR #8339, opened 2019-10-16
Use constant for websocket protocol GUID string

vlazar:
I think it would make sense to extract this magic WebSocket protocol string to a constant for clarification.

### DIFF
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -8,6 +8,8 @@ require "uri"
 
 # :nodoc:
 class HTTP::WebSocket::Protocol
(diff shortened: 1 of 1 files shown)


src/http/web_socket/protocol.cr
```diff
@@ -301,9 +303,9 @@ class HTTP::WebSocket::Protocol
 
   def self.key_challenge(key)
     {% if flag?(:without_openssl) %}
-      Digest::SHA1.base64digest("#{key}258EAFA5-E914-47DA-95CA-C5AB0DC85B11")
+      Digest::SHA1.base64digest("#{key}#{GUID}")
```
Sija: I think this could be done with `+` avoiding string allocation.

ditto below.

PR #8342, opened 2019-10-17
update mime type of .js files to text/javascript

mamantoha:
application/javascript is being deprecated per
https://datatracker.ietf.org/doc/draft-ietf-dispatch-javascript-mjs/

add .webp for builtin

### DIFF
--- a/src/mime.cr
+++ b/src/mime.cr
@@ -91,7 +91,7 @@ module MIME
     ".html" => "text/html; charset=utf-8",
     ".jpg"  => "image/jpeg",
     ".jpeg" => "image/jpeg",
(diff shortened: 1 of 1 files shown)


straight-shoota:
Changing `application/javascript` to `text/javascript` is good. This is already recommended by the HTML5 specification.

But I don't approve of `.mjs`. These default MIME types should just be a very basic set to get the simplest of HTTP deployments working. `.mjs` is a pretty new extension and not very common. I'm not sure there exists any standard specification yet. The one you linked is just a draft.

`.webp` is fine.

mamantoha:
@straight-shoota force-pushed

PR #8344, opened 2019-10-18
Add clickable anchor icon next to headings

### LINKED ISSUE #8127: Add ids to headings in generated docs

Blacksmoke16:
## Overview

* Adds clickable anchor icons next to headings within the generated docs
  * Added to auto generated headings
    * `Overview`
    * `Constant Summary`
    * `Class Methods Overview`
    * etc
  * Added to headings (`h1`-`h6`) within user doc comments
  * Sanitized of unsafe/reserved URL characters

Example:

image

resolves #8127


src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,18 @@
...
+  private ANCHOR_ICON = %(<svg class="octicon octicon-link" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path></svg>)
+
+  private module Anchorable
+    def anchor : String
+      @title.downcase.gsub(' ', '-')
+    end
+  end
+
+  def self.anchor_link(anchor : String)
+    %(<a id="#{anchor}" class="anchor" href="##{anchor}">#{ANCHOR_ICON}</a>)
```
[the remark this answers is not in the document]
Blacksmoke16: Oh nice, didn't know that was a thing.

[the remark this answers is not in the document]
Blacksmoke16: Oh nice, didn't know that was a thing.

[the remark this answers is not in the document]
Blacksmoke16: @straight-shoota What are your thoughts on the best way to implement this?  So far the options I thought are;
1. Include the SVG within the `_head.html` file so its available within every file
  1.1 but add a `display: none;` on it so its hidden
2. Add `img/svg-defs.svg` and do like `xlink:href="../img/svg-defs.svg#oction-link"`
  2.1 However this wouldn't work locally nor on IE browsers
3. Keep it as it is ATM.
4. Something else im not aware of?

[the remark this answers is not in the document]
straight-shoota: Yes, put the SVG symbol into header or footer with display: none. So 1.1 :+1:

src/compiler/crystal/tools/doc/html/css/style.css
```diff
@@ -630,3 +630,27 @@ span.flag.purple {
...
+}
+
+.main-content h1:hover .anchor .octicon-link,
+.main-content h2:hover .anchor .octicon-link,
+.main-content h3:hover .anchor .octicon-link,
+.main-content h4:hover .anchor .octicon-link,
+.main-content h5:hover .anchor .octicon-link,
+.main-content h6:hover .anchor .octicon-link {
+  visibility: visible
+}
```
straight-shoota: Please add a line break at the end of the file.

src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,32 @@
 require "ecr/macros"
 
 module Crystal::Doc
+  SVG_DEFS = <<-SVGS
+<div class="hidden">
```
straight-shoota: Minor: This `<div>` is unnecessary, `<svg class="hidden">` is better.

src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,32 @@
...
+  <svg>
+    <symbol id="octicon-link">
```
straight-shoota:
```suggestion
    <symbol id="octicon-link" viewbox="0 0 16 16">
```

src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,32 @@
...
+  private module Anchorable
+    def anchor : String
+      @title.downcase.gsub(' ', '-')
+    end
+  end
+
+  def self.anchor_link(anchor : String)
+    <<-ANCHOR
+    <a id="#{anchor}" class="anchor" href="##{anchor}">
+      <svg class="octicon-link" viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true">
```
straight-shoota: `viewBox` should be defined on `<symbol>`. `width` and `height` should be defined in CSS. `version` argument can be removed, it is deprecated.
```suggestion
      <svg class="octicon-link" aria-hidden="true">
```

src/compiler/crystal/tools/doc/templates.cr
Blacksmoke16: Reasoning was if we want to add other SVGs, they would all be hidden.

Blacksmoke16:
Test failure seems unrelated.

src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,30 @@
...
 module Crystal::Doc
+  SVG_DEFS = <<-SVGS
+<svg class="hidden">
+  <symbol id="octicon-link" viewBox="0 0 16 16">
+    <path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path>
+  </symbol>
+</svg>
+SVGS
+
+  private module Anchorable
```
straight-shoota: Actually, I don't think this module should exist. It's not very useful, you could just transform the anchor string in `.anchor_link` directly.

Blacksmoke16: I don't think I'm a fan of doing any transformations within `.anchor_link`.  It should just be responsible for outputting the markup for a given `anchor`.

straight-shoota: It's just a generic string transformation that is well within the scope of `anchor_link`.

Blacksmoke16: And you'd be okay with doing the extra gsub/downcase calls on every `anchor` even those that are already correct (from the `render_header` method)?

Currently this transformation is just scoped to the built in templates.

straight-shoota: Sure. Alternatively you could explicitly call a transformation method for those cases, where it't not already a slugified value. I don't care, but I'd want to avoid an additional type just for this when there are suitable alternative solutions.

src/compiler/crystal/tools/doc/templates.cr
```diff
@@ -1,6 +1,26 @@
 require "ecr/macros"
 
 module Crystal::Doc
+  SVG_DEFS = <<-SVGS
+<svg class="hidden">
+  <symbol id="octicon-link" viewBox="0 0 16 16">
+    <path fill-rule="evenodd" d="M4 9h1v1H4c-1.5 0-3-1.69-3-3.5S2.55 3 4 3h4c1.45 0 3 1.69 3 3.5 0 1.41-.91 2.72-2 3.25V8.59c.58-.45 1-1.27 1-2.09C10 5.22 8.98 4 8 4H4c-.98 0-2 1.22-2 2.5S3 9 4 9zm9-3h-1v1h1c1 0 2 1.22 2 2.5S13.98 12 13 12H9c-.98 0-2-1.22-2-2.5 0-.83.42-1.64 1-2.09V6.25c-1.09.53-2 1.84-2 3.25C6 11.31 7.55 13 9 13h4c1.45 0 3-1.69 3-3.5S14.5 6 13 6z"></path>
+  </symbol>
+</svg>
+SVGS
```
straight-shoota: Please indent these lines. Everything else looks great!

Blacksmoke16:
@straight-shoota I noticed a small issue with user defined headings.  I'll push up a fix soon.

`Formatting and Parsing Time` is being converted to `formattingandparsingtime` instead of `formatting-and-parsing-time`.

src/compiler/crystal/tools/doc/markdown/parser.cr
```diff
@@ -110,10 +110,10 @@ class Crystal::Doc::Markdown::Parser
 
   def render_header(level : Int32, line : String, increment : Int32)
     anchor = line
-      .underscore              # Underscore the string
-      .gsub(/[^\w\d\-.~]/, "") # Delete unsafe URL characters
-      .strip                   # Strip leading/trailing whitespace
-      .gsub(/[\ _]/, '-')      # Replace `_` and leftover whitespace with `-`
+      .underscore                # Underscore the string
+      .gsub(/[^\w\d\-.~\ ]/, "") # Delete unsafe URL characters
```
straight-shoota: You could probably use `\s` and also below.

src/compiler/crystal/tools/doc/markdown/parser.cr
```diff
@@ -108,8 +108,14 @@ class Crystal::Doc::Markdown::Parser
...
   end
 
-  def render_header(level, line, increment)
-    @renderer.begin_header level
+  def render_header(level : Int32, line : String, increment : Int32)
+    anchor = line
+      .underscore                # Underscore the string
+      .gsub(/[^\w\d\-.~\ ]/, "") # Delete unsafe URL characters
+      .strip                     # Strip leading/trailing whitespace
+      .gsub(/[\ _]/, '-')        # Replace `_` and leftover whitespace with `-`
```
straight-shoota:
```suggestion
      .gsub(/[\s_-]+/, '-')        # Replace `_` and leftover whitespace with `-`
```

This makes sure `Foo: Bar` turns into `foo-bar`, not `foo--bar`.

straight-shoota:
Thank you :heart:

Issue #7824, opened 2019-05-27
Steal Object#present?/blank? and #presence from ActiveSupport


Sija:
I'd like to suggest including implementation of `Object#present?`, `Object#blank?` and `Object#presence` [found](https://blog.arkency.com/2017/07/nil-empty-blank-ruby-rails-difference/) in Rails' ActiveSupport.

These are ones of the few extensions from AS I'm finding extremely handy in every-day coding - similarly to `Time::Span#ago`.

If agreed on, I'd happily open a PR.

j8r:
This remind me `be_truthy` and `be_falsy` of spec, but with a less explicit name.

straight-shoota:
I don't think we'll add this to the stdlib.

See https://github.com/crystal-lang/crystal/pull/2413#issuecomment-205446575 and #1299 for previous discussions.

kostya:
https://github.com/kostya/blank

PR #2413, opened 2016-04-04
Added present? and blank?

Seachaos:
add present? and blank? in string, float, integer, array, hash. and add length for them

PR #8345, opened 2019-10-18
Added String#presence

igor-alexandrov:
Added String#presence method that returns `self` if not blank, otherwise it returns `nil`.

Based on ActiveSupport#presence: https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/object/blank.rb#L45

This can be useful for params parsing:

```crystal
state   = params[:state]   if params[:state]
country = params[:country] if params[:country]
region  = state || country || "US"

# becomes

region = params[:state].presence || params[:country].presence || "US"
```
Or for example CSV parsing (to avoid blank strings in DB):

```crystal
flip_record.buyer_first_name = csv["buyerFirstName"].presence
flip_record.buyer_last_name = csv["buyerLastName"].presence
```

Thanks!

### DIFF
--- a/src/nil.cr
+++ b/src/nil.cr
@@ -106,7 +106,7 @@ struct Nil
     raise NilAssertionError.new
   end
 
-  # Returns `self`
+  # Returns `self`.
   # This method enables to call the `presence` method (see `String#presence`) on a union with `Nil`.
   # The idea is to return `nil` when the value is `nil` or empty.
   #
--- a/src/nil.cr
+++ b/src/nil.cr
@@ -112,8 +112,8 @@ struct Nil
   #
   # ```
   # config = {"empty" => ""}
-  # config["empty"]?.presence || "default"   # => "default"
-  # config["missing"]?.presence || "default" # => "default"
+  # config["empty"]?.presence   # => nil
+  # config["missing"]?.presence # => nil
   # ```
   def presence
     self
--- a/src/nil.cr
+++ b/src/nil.cr
@@ -106,6 +106,19 @@ struct Nil
     raise NilAssertionError.new
   end
 
+  # Returns `self`
+  # Added to make the code below compile.
+  #
+  # ```
+  # data = {} of Symbol => String
+  # data[:missing_key]?.presence || "default" # => "default"
+  # ```
+  #
+  # See also: `String#presence`.
+  def presence
+    self
+  end
+
   def clone
     self
   end
--- a/src/nil.cr
+++ b/src/nil.cr
@@ -107,14 +107,14 @@ struct Nil
   end
 
   # Returns `self`
-  # Added to make the code below compile.
+  # This method enables to call the `presence` method (see `String#presence`) on a union with `Nil`.
+  # The idea is to return `nil` when the value is `nil` or empty.
   #
   # ```
-  # data = {} of Symbol => String
-  # data[:missing_key]?.presence || "default" # => "default"
+  # config = {"empty" => ""}
+  # config["empty"]?.presence || "default"   # => "default"
+  # config["missing"]?.presence || "default" # => "default"
   # ```
-  #
-  # See also: `String#presence`.
   def presence
     self
   end
--- a/src/string.cr
+++ b/src/string.cr
@@ -2405,10 +2405,15 @@ class String
   # Returns `self` if not blank, otherwise returns `nil`.
   #
   # ```
-  # "".presence        # => nil
-  # "   ".presence     # => nil
-  # "   a   ".presence # => "   a   "
+  # "a".presence || "default" # => "a"
+  # "".presence || "default"  # => "default"
+  # nil.presence || "default" # => "default"
+  #
+  # data = {} of Symbol => String
+  # data[:missing_key]?.presence || "default" # => "default"
   # ```
+  #
+  # See also: `Nil#presence`.
   def presence
     self if !blank?
   end
--- a/src/string.cr
+++ b/src/string.cr
@@ -2405,9 +2405,9 @@ class String
   # Returns `self` if not blank, otherwise returns `nil`.
   #
   # ```
-  # "".presence         # => nil
-  # "   ".presence      # => nil
-  # "   a   ".presence  # => "   a   "
+  # "".presence        # => nil
+  # "   ".presence     # => nil
+  # "   a   ".presence # => "   a   "
   # ```
   def presence
     self if !blank?
--- a/src/string.cr
+++ b/src/string.cr
@@ -2405,12 +2405,15 @@ class String
   # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
   #
   # ```
-  # "a".presence || "default" # => "a"
-  # "".presence || "default"  # => "default"
-  # nil.presence || "default" # => "default"
+  # "a".presence || "default"         # => "a"
+  # "".presence || "default"          # => "default"
+  # "   ".presence || "default"       # => nil
+  # "    a    ".presence || "default" # => "a"
+  # nil.presence || "default"         # => "default"
   #
-  # data = {} of Symbol => String
-  # data[:missing_key]?.presence || "default" # => "default"
+  # config = {"empty" => ""}
+  # config["empty"]?.presence || "default"   # => "default"
+  # config["missing"]?.presence || "default" # => "default"
   # ```
   #
   # See also: `Nil#presence`.
--- a/src/string.cr
+++ b/src/string.cr
@@ -2402,7 +2402,7 @@ class String
     true
   end
 
-  # Returns `self` if not blank, otherwise returns `nil`.
+  # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
   #
   # ```
   # "a".presence || "default" # => "a"
--- a/src/string.cr
+++ b/src/string.cr
@@ -2402,6 +2402,17 @@ class String
     true
   end
 
+  # Returns `self` if not blank, otherwise returns `nil`.
+  #
+  # ```
+  # "".presence         # => nil
+  # "   ".presence      # => nil
+  # "   a   ".presence  # => "   a   "
+  # ```
+  def presence
+    self if !blank?
+  end
+
   def ==(other : self)
     return true if same?(other)
     return false unless bytesize == other.bytesize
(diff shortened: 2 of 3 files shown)


straight-shoota:
Similar discussions: #7824, #2413, #1299

But it seems there has never been an explicit discussion about only `String#presence`. And we already have `String#blank?`, so it might be a nice addition.

asterite:
I agree. Just having `presence` is good and would be idiomatic.

Blacksmoke16:
Is there any example use case that can't be done with `[]?`

```cry
region = params[:state]? || params[:country]? || "US"
```
or
```cry
flip_record.buyer_first_name = csv["buyerFirstName"]?
```

igor-alexandrov:
@Blacksmoke16 good point... to be honest, I forgot about `[]?`. Let me think a bit.

Blacksmoke16:
@igor-alexandrov Naw the `not blank` part makes it better since `""` isn't falsy, so wouldn't fallback on the other options if its blank.

src/string.cr
```diff
@@ -2402,6 +2402,17 @@ class String
     true
   end
 
+  # Returns `self` if not blank, otherwise returns `nil`.
+  #
+  # ```
+  # "".presence        # => nil
+  # "   ".presence     # => nil
+  # "   a   ".presence # => "   a   "
```
bcardiff: maybe we should add an example like 

```crystal
"a".try(&.presence) || "default" # => "a"
"".try(&.presence) || "default"  # => "default"
nil.try(&.presence) || "default" # => "default"
```

That would be a common use case unless `Nil#presence` is added.

jhass: Let's add `Nil#presence`

igor-alexandrov: > Let's add `Nil#presence`

You mean to add `String#presence` and `Nil.presence`?

jhass: No, `Nil#presence` so the following compiles:
```cr
data = {} of Symbol => String
value = data[:foo]?.presence || "bar"

Comment (igor-alexandrov) [reply src/string.cr]: I made a mistake, of course I meant `Nil#presence`. Now it makes sense, `Nil#presence` and `String#presence` will allow to compile the code above.

Comment (bcardiff) [reply src/string.cr]: shinny sample snippets:

```cr
"a".presence || "default" # => "a"
"".presence || "default"  # => "default"
nil.presence || "default" # => "default"

data = {} of Symbol => String
data[:missing_key]?.presence || "default" # => "default"
```

igor-alexandrov: Looks great! I will update PR a bit later.

Sija: Hmm, seems like we slowly drift back onto the #7824 turf... ;)

asterite: Indeed. I think these methods are good, but they are mostly useful for web frameworks where the strings come from query parameters that might or might not be there. Or from form parameters. So I still wonder where these should be included in the standard library. The std is not meant to be thought exclusively for web apps.

straight-shoota: @asterite I guess it's useful for general input parsing, not only web. Say configuration files or so. 

But then... I've never really needed `String#presence` in Crystal. Which might mean it's not too important to have it?

igor-alexandrov: @straight-shoota @asterite I found that I need this method during CSV file parsing, so this is not only a web-needed method.

ysbaddaden: @asterite `blank?` and `presence` come from Rails so we think of them as mostly useful for web development, but they're actually nice to have whenever we deal with user input.

A use case, for example, is dealing with legacy databases where many columns are `""` or `NULL` and should be treated the same, `#presence` is a life savior.

igor-alexandrov:
Updated usage examples. Added `Nil#presence`.

src/string.cr
bcardiff: I think that as long as we don't put presence in `Object` then we are slightly better than #7824.

src/string.cr
```diff
@@ -2402,6 +2402,22 @@ class String
     true
   end
 
+  # Returns `self` if not blank, otherwise returns `nil`.
+  #
+  # ```
+  # "a".presence || "default" # => "a"
```
bcardiff: I didn't mean for the other samples to go away.

Especially the one with 
```
"           "
"    a    "
```

they are good corner cases

src/nil.cr
```diff
@@ -106,6 +106,19 @@ struct Nil
     raise NilAssertionError.new
   end
 
+  # Returns `self`
+  # Added to make the code below compile.
```
straight-shoota: Please rephrase this so that it actually makes sense for someone who hasn't seen this PR.

For example: ``This method enables to call the `presence` method (see `String#presence`) on a union with `Nil`. The idea is to return `nil` when the value is `nil` or empty.``

src/string.cr
```diff
@@ -2402,6 +2402,22 @@ class String
...
+  # "".presence || "default"  # => "default"
+  # nil.presence || "default" # => "default"
+  #
+  # data = {} of Symbol => String
+  # data[:missing_key]?.presence || "default" # => "default"
```
straight-shoota: A hash with symbol keys is a Rubyism and often considered a code smell in Crystal. Let's not use that in an example.
```suggestion
  # config = {"empty" => ""}
  # config["empty"]?.presence || "default"   # => "default"
  # config["missing"]?.presence || "default" # => "default"
```

src/string.cr
```diff
@@ -2402,6 +2402,22 @@ class String
...
+  # ```
+  #
+  # See also: `Nil#presence`.
+  def presence
```
straight-shoota:
```suggestion
  def presence : self?
```

src/string.cr
```diff
@@ -2402,6 +2402,22 @@ class String
     true
   end
 
+  # Returns `self` if not blank, otherwise returns `nil`.
```
straight-shoota:
```suggestion
  # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
```

src/string.cr
wontruefree: Why is this a code smell? Seems like a valid use case.

asterite: I'd say using Symbol in Crystal is a code smell, no matter the context.

(I want to eventually remove Symbol from the language, if possible)

bcardiff: I considered valid use case too, but is true that is more likely to be a hash of strings.
When a hash is used with an _expected_ list of symbols, then a named tuple can be used.

straight-shoota: Because you either have dynamic keys in which case you'd use strings (because unlike Ruby, symbols can't be created at runtime) or a fixed set of keys which would better be implemented as a struct/class, named tuple, hash with enum keys. They all improve type safety.

igor-alexandrov:
Updated usage examples for both `String` and `Nil`.

Don't forget that it is Friday today! 🍻

src/string.cr
```diff
@@ -2402,6 +2402,25 @@ class String
...
   end
 
+  # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
+  #
+  # ```
+  # "a".presence || "default"         # => "a"
+  # "".presence || "default"          # => "default"
+  # "   ".presence || "default"       # => nil
+  # "    a    ".presence || "default" # => "a"
+  # nil.presence || "default"         # => "default"
```
straight-shoota: I don't think these examples should use `|| "default"`. It's just clutter and not necessary to understand the behaviour of the method.

Having one example for this application is fine, but in general the sample code should be fully focused on the described feature.

bcardiff: We can use other word than `"default"`, but I think this will become a used idiom. 

Having these cases here directly will be practical for the reader. In particular, I want them to not need to go to de definition of `blank?` to check and _deduce_ how many spaces will be handled.

straight-shoota: Yes, I want to keep the examples! Just strip off ` || "default"` from every line as it's not essential. These samples should primarily document what `#presence` returns, which is either self or nil.

igor-alexandrov: I agree with @straight-shoota here. Updated examples.

src/string.cr
```diff
@@ -2402,6 +2402,25 @@ class String
     true
   end
 
+  # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
+  #
+  # ```
+  # "a".presence || "default" # => "a"
+  # "".presence || "default"  # => "default"
+  # "   ".presence            # => nil
+  # "    a    ".presence      # => "a"
```
bcardiff:
```suggestion
  # "    a    ".presence      # => "    a    "
```

src/nil.cr
```diff
@@ -106,6 +106,19 @@ struct Nil
     raise NilAssertionError.new
   end
 
+  # Returns `self`
```
straight-shoota:
```suggestion
  # Returns `self`.
```

Maybe `s/self/nil`?

src/nil.cr
```diff
@@ -106,6 +106,19 @@ struct Nil
...
+  # This method enables to call the `presence` method (see `String#presence`) on a union with `Nil`.
+  # The idea is to return `nil` when the value is `nil` or empty.
+  #
+  # ```
+  # config = {"empty" => ""}
+  # config["empty"]?.presence || "default"   # => "default"
+  # config["missing"]?.presence || "default" # => "default"
```
straight-shoota: I'd prefer removing `|| "default"` here as well.

src/string.cr
```diff
@@ -2402,6 +2402,25 @@ class String
     true
   end
 
+  # Returns `self` unless `#blank?` is `true` in which case it returns `nil`.
+  #
+  # ```
+  # "a".presence || "default" # => "a"
```
straight-shoota: Please remove `|| "default"` here. The first examples should be as minimal as possible. Having it in the hash example is enough.

src/nil.cr
igor-alexandrov: Done.

src/string.cr
igor-alexandrov: Done.

bcardiff:
Thanks @igor-alexandrov ! Congrats on your first contribution to Crystal 💯 . I will definitely use this soon.

igor-alexandrov:
Awesome!  🍻

bew:
Note: there is no spec for `Nil#presence` is that intentional?

igor-alexandrov:
I will add it a bit later.

PR #8347, opened 2019-10-18
Array: fix range assignment index out of bounds

### LINKED ISSUE #8341: Assignment to unbounded (or empty) range of empty array should not fail
Consider the following examples

```crystal
# This works
a = [1]
a[..] = [1,2,3]

# This fails with "Index out of bounds"
a = [] of Int32
a[..] = [1,2,3]
```
The latter seems to be a valid idiom for replacing the content of an array (in contrast `a = [1,2,3]` creates a new array). Also note that reading does work: 

```crystal
a = [] of Int32
puts a[..] # this is ok
```

One could argue whether replacement of empty ranges should work (i.e. it means "insertion") if the resulting array would not contain gaps. In other words, one could argue if the following examples should work or not:

```crystal
#### The following examples should work
a = [1]
a[1..] = [2,3] # results in a = [1,2,3]

a = [1]
a[1...1] = [2,3] # as above because 1...1 is empty

# Note that insertion of empty ranges in the middle is does already work
a = [1,5]
a[1...1] = [2,3,4]  # results in a = [1,2,3,4,5]

#### The following examples should fail
a = [1]
a[2..] = [2, 3] # should fail because it would result in a = [1, XXX, 2, 3] where a[1] is undefined

a = [1]
a[1..1] = [2,3] # should fail because the non-empty range 1..1 is not valid for a.
```

asterite:
Fixes #8341

The main fix is that the `index` that we pass can be the `size` of the Array, because it starts a range from there and it should work if `count` is zero (or a greater value too). The usual `check_index_out_of_bounds` checks for `index < size`, in this case we need `index <= size`.

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -400,7 +400,9 @@ class Array(T)
   def []=(index : Int, count : Int, value : T)
     raise ArgumentError.new "Negative count: #{count}" if count < 0
 
-    index = check_index_out_of_bounds index
+    index += size if index < 0
+    raise IndexError.new unless 0 <= index <= size
+
     count = index + count <= size ? count : size - index
 
     case count
@@ -456,7 +458,9 @@ class Array(T)
   def []=(index : Int, count : Int, values : Array(T))
     raise ArgumentError.new "Negative count: #{count}" if count < 0
 
-    index = check_index_out_of_bounds index
+    index += size if index < 0
+    raise IndexError.new unless 0 <= index <= size
+
     count = index + count <= size ? count : size - index
     diff = values.size - count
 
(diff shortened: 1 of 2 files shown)


src/array.cr
```diff
@@ -400,7 +400,9 @@ class Array(T)
   def []=(index : Int, count : Int, value : T)
     raise ArgumentError.new "Negative count: #{count}" if count < 0
 
-    index = check_index_out_of_bounds index
+    index += size if index < 0
+    raise IndexError.new unless 0 <= index <= size
```
straight-shoota: Could you add a comment to explain why not using check_index_out_of_bounds index?

asterite: Good idea. Done!

PR #5990, opened 2018-04-22
 Add --canonical-base-url cli option to docs generator

Sija:
Closes #5952

PR #8348, opened 2019-10-18
[Compiler] Add sitemap to doc generator

straight-shoota:
The doc generator can produce a `sitemap.xml` which lists all HTML pages accessible for search engines. The goal is to use this sitemap to assign lower priorities to outdated doc pages on https://crystal-lang.org/api/ to make the most recent release (`latest`) show up first in search results. See crystal-lang/crystal-website#79 for details. Having such a sitemap is generally useful for shards as well, so it  makes sense to directly integrate it into the docs generator.

Sitemaps are only generated when `--sitemap-base-url=URL` option is set. Sitemaps require absolute paths, so it can't be produced without a base url. Additional options are `--sitemap-priority` and `--sitemap-changefreq` which correspond do the [respective sitemap features](https://www.sitemaps.org/protocol.html#xmlTagDefinitions).

This also reverts #5990 which tried an alternative approach to solving the search priority issue using canonical URLs. But this completely removes older versions from search results.

Ideally, `sitemap.xml` should be compressed, but the compiler is supposed to be built without libz, so this needs to be done separately.

### DIFF
--- a/src/compiler/crystal/tools/doc/html/sitemap.xml
+++ b/src/compiler/crystal/tools/doc/html/sitemap.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  <% types.each do |type| %><url>
+    <loc><%= ::Path.posix(base_url, type.path) %></loc>
+    <priority><%= priority %></priority>
+    <changefreq><%= changefreq %></changefreq>
+  </url><% end %>
+</urlset>
--- a/Makefile
+++ b/Makefile
@@ -104,7 +104,7 @@ compiler_spec: $(O)/compiler_spec ## Run compiler specs
 
 .PHONY: docs
 docs: ## Generate standard library documentation
-	$(BUILD_PATH) ./bin/crystal docs -b https://crystal-lang.org/api/latest src/docs_main.cr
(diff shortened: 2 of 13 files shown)


src/compiler/crystal/tools/doc/html/sitemap.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  <% types.each do |type| %><url>
+    <loc><%= ::Path.posix(base_url, type.path) %></loc>
```
bcardiff: Is there a reason to use `Path` used instead of `URI`?

straight-shoota: `Path#join` is more specific to this use case. `URI#resolve` would work as well but it's not necessary to parse and resolve these URIs because we know that `type.path` is relative and only needs to be appended to `base_url`. Thus it's also a bit more efficient.

straight-shoota: > It not a valid path value.

Of course, it is a perfectly valid posix path (although not normalized). That's why this works.

But sure, I can switch to `URI#resolve`.

bcardiff: My bad. I just would never go into Path for a URL. It's fine to leave it as-is. As you prefer.

PR #8350, opened 2019-10-19
Fix formatting issue in crystal init template

tcd:
The template for the `readme.md` file used by `crystal init` had redundant indentation for the first fenced codeblock.

Here's what's generated when that file is read by `crystal docs`:

before

And here's the output with the indentation removed:

after

### DIFF
--- a/src/compiler/crystal/tools/init/template/readme.md.ecr
+++ b/src/compiler/crystal/tools/init/template/readme.md.ecr
@@ -7,11 +7,11 @@ TODO: Write a description here
 <%- if config.skeleton_type == "lib" -%>
 1. Add the dependency to your `shard.yml`:
 
-   ```yaml
-   dependencies:
-     <%= config.name %>:
-       github: <%= config.github_name %>/<%= config.name %>
-   ```
+```yaml
+dependencies:
+  <%= config.name %>:
+    github: <%= config.github_name %>/<%= config.name %>
+```
 
 2. Run `shards install`
 <%- else -%>


bew:
This looks wrong to me, the idea was to have the code block as part of point `1.`. If you put it without indentation you break the numbered list (1. 2.)

I'm not sure what's happening with the generated html in the first example, I'd suggest resolving that instead.

Here is an example of the correct generated html on github:
Screenshot_20191019-090712__01
We should fix the doc generator here instead.

asterite:
This is broken because the Markdown implementation is broken.

I proposed using markd for the internal implementation, at least temporarily. If we would have merged that this issue, this discussion, and the time poured into it, wouldn't exit.

tcd:
@bew I see what you mean.

@asterite Is there any code used for markdown processing (aside from specs) outside of `src/compiler/crystal/tools/doc/markdown`?

tcd:
@bew I see what you mean.
@asterite Why was that idea opposed? Can't MIT code be used in an Apache project?

asterite:
Others don't want a temporary solution that's better than the current state but is not the final solution.

tcd:
So would the *final solution* be to implement a feature complete markdown parser that's not in the standard library that's only used by the `crystal` tool?

asterite:
@tcd The wanted solution (not by me, though) is to extract `crystal tool format` outside of the compiler, to an external tool, that would still ship with the compiler, and that could use external shards like markd.

tcd:
@asterite What's you're issue with that idea? Keeping the tool up to date with the main repo?

Also, just `format`? Not any of the other tool commands?

asterite:
My issue is that it lives in the uncertain future. Nobody knows when, how and if that will be done. What can be done **right now** is to use a complete implementation as a temporary solution. In fact I already sent a PR for that, which costs nothing to us maintainers and to users, but it was rejected. However, it costs us time and thought to discuss this issue. Like I said before, with my PR merged this entire issue and discussion wouldn't exist.

straight-shoota:
> Others don't want a temporary solution that's better than the current state but is not the final solution.

I don't think that assessment is accurate. But let's continue the discussion in #4613 to figure this out.

tcd:
@asterite @straight-shoota Going to close this pull (as it doesn't properly fix the problem it addresses) and will go check out the linked issue. Hopefully I can be of some help over there 🤞.

PR #8356, opened 2019-10-20
Add documentation for Channel#send and Channel#close (crystal-lang#8346)

### LINKED ISSUE #8346: Behaviour of unbuffered Channel#send on Channel#close in v0.31.x vs previous versions
The Crystal compiler v0.31.x seems to encode a different behaviour than the previous versions in the scenario where
1. Fiber `f1` is waiting on a `#send` call on an unbuffered channel `c`
2. Fiber `f2` invokes `#close` on `c`.

### Outcome in v0.30.x and previous versions
`f1` gets unblocked as if another Fiber had received the value

### Outcome in v0.31.x
`#send` raises a `Channel::ClosedError` exception for `f1` to handle

This is exemplified in the code below.
```crystal
c = Channel(Int32).new
spawn do
  puts "f1 sending"
  c.send 2
  puts "f1 sent"
rescue
  puts "f1 rescued"
end
 
spawn do
  puts "f2 closing"
  c.close
  puts "f2 closed"
end
 
sleep 1
```
This prints the following in v0.30.0
```
#output 1
f1 sending
f2 closing
f2 closed
f1 sent
```
Whereas it prints the following in v0.31.x
```
#output 1
f1 sending
f2 closing
f2 closed
f1 rescued
```

Just wanted to figure out if this is intended or a regression, as this can have a pretty dramatic impact on error handling in concurrent apps. Happy to add some docs if we're happy with the current behaviour.

lbarasti:
Closes  #8346

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -116,7 +116,7 @@ class Channel(T)
 
   # Closes the channel.
   # The method prevents any new value from being sent to / received from the channel.
-  # It wakes up any sender / receiver fibers waiting on the channel.
+  # All fibers blocked in `send` or `receive` will be awakened with `Channel::ClosedError`
   #
   # Both awaiting and subsequent calls to `#send` will consider the channel closed.
   # All items successfully sent to the channel can be received, before `#receive` considers the channel closed.
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -114,6 +114,10 @@ class Channel(T)
     end
   end
 
+  # Closes the channel.
+  # The method prevents any new value from being sent to / received from the channel.
+  # It wakes up any sender / receiver fibers waiting on the channel.
+  # It only has effect *after* any previously enqueued value in the channel has been received, up to the channel's capacity.
   def close
     @lock.sync do
       @closed = true
@@ -144,6 +148,11 @@ class Channel(T)
     @closed
   end
 
+  # Sends a value to the channel.
+  # If the channel has spare capacity, then the method returns immediately.
+  # Otherwise, this method blocks until the value is received from the channel or the channel's spare capacity becomes greater than 0.
+  #
+  # Raises `ClosedError` if the channel is closed or closes while waiting on a full channel.
   def send(value : T)
     @lock.sync do
       raise_if_closed
@@ -182,13 +191,15 @@ class Channel(T)
   end
 
   # Receives a value from the channel.
-  # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
+  # If there is a value waiting, then it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
   #
   # Raises `ClosedError` if the channel is closed or closes while waiting for receive.
   #
   # ```
   # channel = Channel(Int32).new
-  # channel.send(1)
+  # spawn do
+  #   channel.send(1)
+  # end
   # channel.receive # => 1
   # ```
   def receive
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -117,7 +117,8 @@ class Channel(T)
   # Closes the channel.
   # The method prevents any new value from being sent to / received from the channel.
   # It wakes up any sender / receiver fibers waiting on the channel.
-  # It only has effect *after* any previously enqueued value in the channel has been received, up to the channel's capacity.
+  # Both awaiting and subsequent calls to `#send` will consider the channel closed.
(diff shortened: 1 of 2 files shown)


src/channel.cr
```diff
@@ -182,13 +191,15 @@ class Channel(T)
...
+  # If there is a value waiting, then it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
   #
   # Raises `ClosedError` if the channel is closed or closes while waiting for receive.
   #
   # ```
   # channel = Channel(Int32).new
-  # channel.send(1)
+  # spawn do
+  #   channel.send(1)
+  # end
```
lbarasti: An alternative way to fix the code snippet would be to go buffered
```crystal
channel = Channel(Int32).new(1)
channel.send(42)
channel.receive # => 42
```

src/channel.cr
```diff
@@ -182,13 +191,15 @@ class Channel(T)
   end
 
   # Receives a value from the channel.
-  # If there is a value waiting, it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
+  # If there is a value waiting, then it is returned immediately. Otherwise, this method blocks until a value is sent to the channel.
```
lbarasti: Added `then` for consistency with the docs above - both are valid, I believe

src/channel.cr
```diff
@@ -114,6 +114,10 @@ class Channel(T)
     end
   end
 
+  # Closes the channel.
+  # The method prevents any new value from being sent to / received from the channel.
+  # It wakes up any sender / receiver fibers waiting on the channel.
+  # It only has effect *after* any previously enqueued value in the channel has been received, up to the channel's capacity.
```
lbarasti: I appreciate the wording can be improved, so happy to take suggestions here

lbarasti: I would add that calling close on a closed channel does not have any effect, but such behaviour hasn't been documented in the specs, so I'd better add a test too, if we're keen on being explicit on that

src/channel.cr
```diff
@@ -114,6 +114,10 @@ class Channel(T)
     end
   end
 
+  # Closes the channel.
+  # The method prevents any new value from being sent to / received from the channel.
+  # It wakes up any sender / receiver fibers waiting on the channel.
```
RX14: All fibers blocked in `send` or `receive` will recieve `Channel::ClosedError`?

src/channel.cr
RX14: I'd probably just say

> All items successfully sent to the channel can be received before `receive` considers the channel closed, but `send` will consider the channel closed before `close` returns.

RX14: I'd document and spec that. You can do that in this PR.

src/channel.cr
```diff
@@ -144,6 +148,11 @@ class Channel(T)
     @closed
   end
 
+  # Sends a value to the channel.
+  # If the channel has spare capacity, then the method returns immediately.
+  # Otherwise, this method blocks until the value is received from the channel or the channel's spare capacity becomes greater than 0.
```
RX14: > Otherwise, this method blocks the calling fiber until another fiber calls `receive`.

src/channel.cr
lbarasti: Exactly, see
* [spec 1](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr#L76) and [spec 2](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr#L449) for `#receive`
* [spec](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr#L143) for `#receive?`
* [spec 1](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr#L217) and [spec 2](https://github.com/crystal-lang/crystal/blob/master/spec/std/channel_spec.cr#L427) for `#send`

src/channel.cr
lbarasti: Done:

+ added to `Channel#close` docs
```
Calling `#close` on a closed channel does not have any effect.
```
+ Added spec
```
Channel "does not raise nor change its status when it is closed more than once"
```

src/channel.cr
lbarasti: @RX14 
> All fibers blocked in send or receive will receive Channel::ClosedError

is ambiguous in that one might expect to actually `receive` an object of type ClosedError. I'm opting for

> All fibers blocked in send or receive will be awakened with Channel::ClosedError 

Let me know what you think

PR #8358, opened 2019-10-21
Add missing LLVM::Attribute members

### LINKED ISSUE #8294: Port Crystal to LLVM9
LLVM9 been released a while ago. Arch Linux team is looking what need to be done to bring the latest version of LLVM.

I am checking how crystal works with LLVM9. The project compiles fine with LLVM9. It seems LLVM9 does not have any breaking API changes.
```
$ crystal --version
Crystal 0.31.1 (2019-10-04)

LLVM: 9.0.0
```

But when I try to compile a simple crystal app like this:
```
def main
  channel = Channel(Int32).new(9)

  (1..10).each do |i|
    spawn do
      channel.send(i)
    end
  end

  result = [] of Int32
  9.times do
    result << channel.receive
  end

  puts result
end

main
```

compilation fails:
```
$ crystal main.cr 
Missing hash key for value: 11 (KeyError)
Unhandled exception: Error opening file 'crystal' with mode 'r': No such file or directory (Errno)
[0x56139e0bd3c6] ???
[0x56139d744e4b] __crystal_raise +43
[0x56139d746645] ???
[0x56139d75246b] ???
[0x56139d74b75c] ???
[0x56139d74adf4] ???
[0x56139daa8133] ???
[0x56139d75dd07] ???
[0x56139d7489c7] main +55
[0x7f8e21c1d153] __libc_start_main +243
[0x56139d74355e] _start +46
[0x0] ???
```
What does this `Missing hash key for value: 11` mean?

RX14:
cc @anatol 

Applying this patch (or actually a subset of it, only `ImmArg` needs to be added) should allow building crystal for LLVM 9.

`ImmArg` is set on some function args by default, it can be ignored, just needs to be mapped correctly so as to not fail.

Fixes #8294.

### DIFF
--- a/src/llvm/enums.cr
+++ b/src/llvm/enums.cr
@@ -2,58 +2,62 @@ module LLVM
   {% if LibLLVM.has_constant?(:AttributeRef) %}
     @[Flags]
     enum Attribute : UInt64
-      Alignment                   = 1 << 0
-      AllocSize                   = 1 << 1
-      AlwaysInline                = 1 << 2
-      ArgMemOnly                  = 1 << 3
-      Builtin                     = 1 << 4
-      ByVal                       = 1 << 5
-      Cold                        = 1 << 6
-      Convergent                  = 1 << 7
-      Dereferenceable             = 1 << 8
-      DereferenceableOrNull       = 1 << 9
-      InAlloca                    = 1 << 10
-      InReg                       = 1 << 11
-      InaccessibleMemOnly         = 1 << 12
-      InaccessibleMemOrArgMemOnly = 1 << 13
-      InlineHint                  = 1 << 14
-      JumpTable                   = 1 << 15
-      MinSize                     = 1 << 16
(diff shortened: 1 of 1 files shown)


RX14:
LLVM completely uses the "kind ids" internally, which we query LLVM for them at runtime by string and store them. The raw enum values are only used for crystal in LLVN 3.9+. A look at the [llvm bindings](https://github.com/crystal-lang/crystal/blob/93e5fb9b409ebe42605f88d1dbffef3c3a0a6fcc/src/llvm/lib_llvm.cr#L330-L339) is probaby the best way to grok this.

anatol:
It works fine at Arch Linux. Thank you very much @RX14 for the fix.

RX14:
Note this still isn't the entire set of missing enum members, only the ones that changed between LLVM 8 and 9 which I thought were likely to be culprits. More work could be done on the bindings in the future.

Issue #8360, opened 2019-10-21
round fails with unsigned int parameter


hutou:
```
r2 : UInt8 = 3
11.308.round(r2)
```

**Error output**
```
Showing last frame. Use --error-trace for full trace.

In /usr/lib/crystal/number.cr:240:25

 240 | y = base.to_f ** (-digits)
                         ^
Error: wrong number of arguments for 'UInt8#-' (given 0, expected 1)

Overloads are:
 - UInt8#-(other : Int8)
 - UInt8#-(other : Int16)
 - UInt8#-(other : Int32)
 - UInt8#-(other : Int64)
 - UInt8#-(other : Int128)
 - UInt8#-(other : UInt8)
 - UInt8#-(other : UInt16)
 - UInt8#-(other : UInt32)
 - UInt8#-(other : UInt64)
 - UInt8#-(other : UInt128)
 - UInt8#-(other : Float32)
 - UInt8#-(other : Float64)
```

https://carc.in/runs/7uma.cr

PR #8361, opened 2019-10-22
Fixed crystal-lang/crystal/issues/8360

igor-alexandrov:
Changed `-digits` to `digits.abs`. Added specs.

### DIFF
--- a/src/number.cr
+++ b/src/number.cr
@@ -237,7 +237,7 @@ struct Number
   def round(digits = 0, base = 10)
     x = self.to_f
     if digits < 0
-      y = base.to_f ** (-digits)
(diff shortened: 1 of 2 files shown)


src/number.cr
```diff
@@ -237,7 +237,7 @@ struct Number
   def round(digits = 0, base = 10)
     x = self.to_f
     if digits < 0
-      y = base.to_f ** (-digits)
+      y = base.to_f ** (digits.abs)
```
Sija:
```suggestion
      y = base.to_f ** digits.abs
```

straight-shoota:
Fixed #8360

PR #6154, opened 2018-06-02
WIP - Add K/V type restrictions for Hash methods

bew:
This adds key/value type restrictions to most of the Hash methods.

It makes more compile error, but imo, safer/type-safe code:
```cr
h = {1 => "hello"}
h.has_key(:foo)   # previously: false      now: compile error: no overload...
h[:foo]  # previously: raise KeyError      now: compile error: no overload...
```

---
WIP: I'm fixing compiler compile error now, please wait...

Issue #1764, opened 2015-10-18
compiler not restricted types of hash keys


kostya:
``` ruby
x = {1 => 2}
p x["bla"]?
```

expected compile error.
in my code i trying to find where the bug, until i find that i pass wrong object to hash[]?

asterite:
This is related to #988 . We might want to do add type restrictions there.

asterite:
A relevant question is, does `[1, 2, 3] - [1, 'a', 3]` is valid code? When I added these type restrictions some Set specs stopped compiling, specifically some that do `Set{1, 2, 3} - Set{1, 'a', 3}`, because Set uses Hash internally and then it becomes invalid to delete an `Int32 | Char` from a Hash whose key is Int32.

This definitely needs a lot of thought.

jhass:
I'd say it's valid, since the operation uses element equality. While `[1, 2] - ['a', 'b']` seems odd on first glance, you'd want `[1, 2] - [2.0, 3.0]` or even just `[1u8, 2u8] - [2i32, 3i32]` to work.

kostya:
what if add this restriction by compile key? this restriction can help to find many bugs in programs.

jhass:
That's similar to warnings. If the compiler doesn't enforce something, somebody will depend on the standard behaviour and write a library that gets popular, everybody uses that library and can't use the flag, nobody uses the flag and it becomes redundant.

kostya:
i also spend some time to find bug, from ported ruby code.

minimized example:
```ruby
valid_wday = Set{1, 3}
t = Time.new 2015, 10, 1

result = [] of Int32
while t.month == 10
  result << t.day if valid_wday.includes?(t.day_of_week)
  t += 1.day
end

p result
```

result is empty, because `t.day_of_week` not a Int32
to fix it, just `t.day_of_week.to_i`

but i think compiler can help to find this bug, by key restriction

asterite:
Yes, definitely. And I'll admit I also had some bugs because of wrong types to delete. And other algorithms that work on any type (like `Set#-`) can be written in a different way once we put these restrictions.

/cc @waj

kostya:
i think in case of `[1, 2, 3] - [1, 'a', 3]` should be used special methods without checking types (which user can use when he understand what he doing), but by default it should check type of key for set, hash and array in `includes?`, `delete` and other methods

miketheman:
#triage 
This appears to be fixed by #1851 - but that has since become un-mergable.
[...]

Issue #988, opened 2015-07-16
No type error when deleting from an array an invalid type


kirbyfan64:
Example:

```crystal
ar = [] of Int32
ar.delete "abc"
```

This has caused bugs for me before.

kirbyfan64:
Real-life example:

Consider [these two functions](https://github.com/kirbyfan64/wit/blob/master/wit/Codegen.cr#L134):

```crystal
      # Mark a register as no longer used.
      def freereg(reg)
        @usedregs.delete reg
      end

      # Free a register if it's a register item.
      def ofree(maybereg)
        self.freereg maybereg.reg if maybereg.is_a? Parser::RegItem
      end
```

I had originally written the second-to-last line as:

```crystal
self.freereg maybereg if maybereg.is_a? Parser::RegItem
```

`@usedregs` is supposed to be of the type `Register`, which is completely independent from `Parser::RegItem`. Because I was passing the wrong type to `freereg`, the register was never freed. This caused that to lurk as a bug until my compiler crashed because it ran out of registers.

kirbyfan64:
Possible fix: change [line 418 of `array.cr`](https://github.com/manastech/crystal/blob/master/src/array.cr#L418) to:

```crystal
def delete(obj : T)
```

asterite:
This is interesting :-)

In Java for example [removing an element](http://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html#remove(java.lang.Object)) is not restricted to a type. This is because you delete elements by equality. So, this works in Ruby (and in Crystal):

```crystal
a = [1, 2, 3]
a.delete 1.0
a #=> [2, 3]
```

So I'm not sure what we should do here. This is an excellent opportunity to discuss it :-)

kirbyfan64:
@asterite Well, Ruby is duck-typed, and Java is pretty stupid. ;)

I'm just saying this because I don't think anyone has ever done that intentionally, as it doesn't really serve a purpose. What would the point of it be? `a` could never hold floats, anyway.

Just my three cents. :)

vendethiel:
I think I side with @kirbyfan64 here, for the sole purpose of catching type errors earlier

asterite:
I think it makes sense to put a type restriction there. The same happens with `Hash#[]`, which doesn't have a restriction but maybe it should.

rdp:
Another sighting:

[...]

PR #8362, opened 2019-10-22
Enforce type checking in Hash's methods

carlhoerberg:
Noticed that I had a bug similar to this that went undetected for a while in my code: 

```crystal
h = Hash(String, String).new
h.delete(1)
```

This PR enforces type checking on most public methods in `Hash` to catch above bugs at compile time.


bew:
Looks like #6154 ...

asterite:
It can't be done lightly.

See #6154, #1764, #988

We want to do this at some point but it's not trivial and it conflicts with how one works with union types, nil, etc.

PR #8363, opened 2019-10-22
Add --mcmodel option to compiler

ffwff:
This patch adds a compiler option to specify the target code model.

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -550,7 +550,7 @@ class Crystal::Command
                             Crystal::Warnings::None
                           else
                             error "--warnings should be all, or none"
-                            raise "unreachable"
+                            exit 1
                           end
     end
     opts.on("--error-on-warnings", "Treat warnings as errors.") do |w|
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -366,16 +366,11 @@ class Crystal::Command
         end
         opts.on("--mcmodel MODEL", "Target specific code model") do |mcmodel|
           compiler.mcmodel = case mcmodel
-                             when "default"
-                               LLVM::CodeModel::Default
-                             when "small"
-                               LLVM::CodeModel::Small
-                             when "kernel"
-                               LLVM::CodeModel::Kernel
-                             when "medium"
-                               LLVM::CodeModel::Medium
-                             when "large"
-                               LLVM::CodeModel::Large
+                             when "default" then LLVM::CodeModel::Default
+                             when "small"   then LLVM::CodeModel::Small
+                             when "kernel"  then LLVM::CodeModel::Kernel
+                             when "medium"  then LLVM::CodeModel::Medium
+                             when "large"   then LLVM::CodeModel::Large
                              else
                                error "--mcmodel should be one of: default, kernel, small, medium, large"
                                raise "unreachable"
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -364,7 +364,7 @@ class Crystal::Command
         opts.on("--mattr CPU", "Target specific features") do |features|
           compiler.mattr = features
         end
-        opts.on("--mcmodel CPU", "Target specific code model") do |mcmodel|
+        opts.on("--mcmodel MODEL", "Target specific code model") do |mcmodel|
           compiler.mcmodel = case mcmodel
                              when "default"
                                LLVM::CodeModel::Default
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -550,7 +550,7 @@ class Crystal::Command
                             Crystal::Warnings::None
                           else
                             error "--warnings should be all, or none"
-                            exit 1
+                            raise "unreachable"
                           end
     end
     opts.on("--error-on-warnings", "Treat warnings as errors.") do |w|
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -364,6 +364,23 @@ class Crystal::Command
         opts.on("--mattr CPU", "Target specific features") do |features|
           compiler.mattr = features
         end
+        opts.on("--mcmodel CPU", "Target specific code model") do |mcmodel|
+          compiler.mcmodel = case mcmodel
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/command.cr
```diff
@@ -364,6 +364,23 @@ class Crystal::Command
         opts.on("--mattr CPU", "Target specific features") do |features|
           compiler.mattr = features
         end
+        opts.on("--mcmodel CPU", "Target specific code model") do |mcmodel|
```
r00ster91:
```suggestion
        opts.on("--mcmodel MODEL", "Target specific code model") do |mcmodel|
```

src/compiler/crystal/command.cr
```diff
@@ -364,6 +364,23 @@ class Crystal::Command
...
+                               LLVM::CodeModel::Small
+                             when "kernel"
+                               LLVM::CodeModel::Kernel
+                             when "medium"
+                               LLVM::CodeModel::Medium
+                             when "large"
+                               LLVM::CodeModel::Large
+                             else
+                               error "--mcmodel should be one of: default, kernel, small, medium, large"
+                               raise "unreachable"
```
r00ster91:
```suggestion
```
Does it not compile without this? Isn't `error` a `NoReturn` method? I think so because then this shouldn't be required.

src/compiler/crystal/command.cr
straight-shoota: I think `CPU` is fine.

src/compiler/crystal/command.cr
ffwff: It's required or the compiler will spill out:

```
Error: instance variable '@mcmodel' of Crystal::Compiler must be LLVM::CodeModel, not (LLVM::CodeModel | Nil)
```

straight-shoota:
```suggestion
                               exit 1
```

straight-shoota: `#error` doesn't exit. You need to call `.exit` explicitly.

ffwff: It does for me; Do I really need to do it?

src/compiler/crystal/command.cr
```diff
@@ -364,6 +364,23 @@ class Crystal::Command
         opts.on("--mattr CPU", "Target specific features") do |features|
           compiler.mattr = features
         end
+        opts.on("--mcmodel MODEL", "Target specific code model") do |mcmodel|
+          compiler.mcmodel = case mcmodel
+                             when "default"
```
Sija: Please use `when ... then ...` format for brevity:

```cr
case mcmdel
when "default" then LLVM::CodeModel::Default
...
end

File: src/compiler/crystal/command.cr (hunk 2 above)
Comment (r00ster91) [reply src/compiler/crystal/command.cr]: Weird. This might be a compiler bug.

Comment (straight-shoota) [reply src/compiler/crystal/command.cr]: Yes. `raise "unreachable"` is incorrect because this location is very much reachable as soon as you use an incorrect flag value. At this point there should not be an exception raised. The command should just exit with an error.

Comment (r00ster91) [reply src/compiler/crystal/command.cr]: This is strange because this is the method called:
https://github.com/crystal-lang/crystal/blob/9f34c313d3cb3833565623a8f18a115bac1544f8/src/compiler/crystal/command.cr#L576
and in this method this method is called:
https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/util.cr#L14
and in this method the top-level method `exit` is called. Which is a `NoReturn` method so there shouldn't be a `raise` or another `exit 1` required after this.

Comment (straight-shoota) [reply src/compiler/crystal/command.cr]: @r00ster91 you're right. I don't understand why this wouldn't work.
This isolated example is just fine and as expected: https://carc.in/#/r/7upy

Comment (asterite) [reply src/compiler/crystal/command.cr]: The compiler might not know that something is unreachable, and eventually find it out, but then it might be too late and the nil will be there. So yes, probably a compiler bug.

Comment (bcardiff) [reply src/compiler/crystal/command.cr]: The problem here is that `error` will do a [conditional](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/util.cr#L17) exit. So the return type is `NoReturn | Nil`.

Since the behavior is to actually exit based on the runtime values that will always be there a `raise "unreachable"` is more accurate IMO.

I prefer to leave the `raise "unreachable"` here and also undo the following change in line 541/553 that replace the `raise "unreachable"` with an `exit 1` for the `--warnings`

Comment (straight-shoota) [reply src/compiler/crystal/command.cr]: An alternative solution would be to refactor `error` into separate methods which are dispatched depending on the type of `exit_code`.

```cr
  def self.error(msg, color, exit_code : Nil, stderr = STDERR, leading_error = true) : Nil
    stderr.print "Error: ".colorize.toggle(color).red.bold if leading_error
    stderr.puts msg.colorize.toggle(color).bright
  end

  def self.error(msg, color, exit_code : Int = 1, stderr = STDERR, leading_error = true)
    error(msg, color, nil, stderr, leading_error)
    exit(exit_code)
  end
```

bcardiff: Yes, or have something `error_and_exit`/`fatal_error`/`cli_option_error`. But that is for another PR. My comments apply for accepting this PR in the current state.

src/compiler/crystal/command.cr
```diff
@@ -538,7 +550,7 @@ class Crystal::Command
                             Crystal::Warnings::None
                           else
                             error "--warnings should be all, or none"
-                            raise "unreachable"
+                            exit 1
```
bcardiff:
```suggestion
                            raise "unreachable"
```

Let's rollback this change

bcardiff:
Thanks @ffwff! 

One more question, may I ask the use case that push you to add mcmodel to Crystal? What are you done with it or expecting to?

ffwff:
I needed it to generate code for my OS kernel and @RX14 suggested me to upstream this

rdp:
Out of curiosity what's an example of using this flag?  Thanks!

ffwff:
@rdp It's just a compiler flag, you can provide it when you're building something:

```
crystal build --mcmodel kernel main.cr
```

PR #8364, opened 2019-10-22
Remove Readline from std-lib

ftarulla:
As an exercise for writing shards (and C-bindings), with @bcardiff we extract the Readline module and create a shard at [crystal-lang/crystal-readline](https://www.github.com/crystal-lang/crystal-readline).

The v0.1.0 has the same code as in the Crystal std-lib.


bcardiff:
While extracting this I also notice that the `CRYSTAL_PATH` gives more precedence to the std-lib rather than the shards.
 
Maybe instead of:

`CRYSTAL_PATH="/usr/local/Cellar/crystal/0.31.1/src:lib"`

it should be

`CRYSTAL_PATH="lib:/usr/local/Cellar/crystal/0.31.1/src"`

So Crystal 0.31.1 users can today add `crystal-readline` as a dependency and use it, even if it also exists in the std-lib.

bcardiff:
It's not that I want to shrink the std-lib, but in the case of readline:

* it is not in the prelude, 
* there is no solid specs,
* is no longer used for the compiler, and
* the implementation has not evolved in a long time. I doubt it is widely used.

Removing it will allow contributions in a simpler way probably and it is not critical for the std-lib.

Fryguy:
Not sure how much history was in there, but it's a shame we lost the git history.  In the future, we could use the following tool (disclaimer, I helped write it) - https://github.com/juliancheal/code-extractor  (I also have it as a shell script in a gist here: https://gist.github.com/Fryguy/3490682c7761edeedba1ee28de5ed117)

Issue #8015, opened 2019-07-30
Invalid bitcast when casting null pointer to Nil


waj:
This code crashes during codegen:

```crystal
Pointer(Void).null.as(Nil)
```

with the following error:
```
Module validation failed: Invalid bitcast
  %69 = bitcast i8* %68 to %Nil
```

However, this code works just fine:

```crystal
nil.as(Void*).as(Nil)
```

PR #8368, opened 2019-10-23
Address FIXME

vlazar:
I've noticed a FIXME comment intended to be addressed once the issue #8015 is fixed. It is fixed now by #8019.

### DIFF
--- a/src/box.cr
+++ b/src/box.cr
@@ -27,7 +27,7 @@ class Box(T)
   # Unboxes a `Void*` into an object of type `T`. Note that for this you must
   # specify T: `Box(T).unbox(data)`.
   def self.unbox(pointer : Void*) : T
-    {% if T <= Reference %}
+    {% if T <= Reference || T == Nil %}
       pointer.as(T)
     {% else %}
       pointer.as(self).object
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
Wouldn't this break with `Nil` since there isn't any logic handling it now?  The FIXME said to merge it with the previous conditional, not to remove it entirely.

vlazar:
@Blacksmoke16 These are the cases I've checked locally:

```crystal
p! Pointer(Void).null.as(Nil) # => nil
p! nil.as(Void*).as(Nil) # => nil
a = 1_i64
p! pointerof(a).as(Nil).nil? ? 3 : 7 # => 3
```

First two from https://github.com/crystal-lang/crystal/issues/8015#issue-474832454
The third one from this spec https://github.com/crystal-lang/crystal/pull/8019/files#diff-c4efe0ebc6736f5e9dc5804721a547f6R68

Blacksmoke16:
I don't know enough about this to know if its an issue or not.  I was just pointing out that `Nil` gets handled by the else block now, versus the if block.

https://play.crystal-lang.org/#/r/7uyr

vlazar:
@Blacksmoke16 Thank you for pointing this out! I think there is a tiny chance (99.999...%) you are right and I should have did this instead:

```crystal
  def self.unbox(pointer : Void*) : T
     {% if T <= Reference || T == Nil %}
       pointer.as(T)
     {% else %}
       pointer.as(self).object
     {% end %}
  end
```

Issue #8377, opened 2019-10-25
Documentation is confusing for newcomer from "worlds" other than Ruby


jan-zajic:
I programmed in many languages, but not in Ruby. So i have to rely on Ruby documentation, because Crystal doc is currently little bit confusing for newcomers. 

During reading book at http://crystal-lang.org/docs again and again, when coding and searching for some help, I still can't help feeling like many parts of it expects some previous knowledge of programming in Ruby. Some things are just missing and i must read Ruby documentation or books about Ruby so that the moment comes when i say: I see! This is how it works / it is meant. 

Other place when i search help is api doc at https://crystal-lang.org/api, which looks unfinished / incomplete, so i also often have to search in Ruby API docs, to make sure how things are likely to work and then search in the sources of Crystal.

Please do not take this as criticism, i try to background introduce myself and my future effort that i want invest to help you with this. So i want to focus on adding missing things for others, until i forget how i started with Crystal.  It's not a problem for me, but I would like to help others, especially to real beginners to programming, because Crystal is an excellent language and deserves it.

I would one day be able to recommend, for example to students at university, Crystal as their first language without the need to learn Ruby first.

I also think the diversity that programmers from other worlds (for example I'm primarily Java Guy) will bring can bring a lot of good for Crystal language and libraries design, and for whole Crystal community.

straight-shoota:
Yes, we're still missing lot's of documentation. Yet for some reason, many Crystal developers still manage to get around. I don't know if this is just because of Ruby background. There are also many differences to Ruby...

Anyway, it's great that you're willing to help! Especially people who are new to the language have a fresh view on things and what documentation is needed the most. Thanks for that!
If you can't write documentation for something yourself but think that it's important to document, don't hesitate to open issue asking someone else to do it.

RX14:
> I still can't help feeling like many parts of it expects some previous knowledge of programming in Ruby.

This is a well known issue! However, since it's very difficult for us - who already know the language - to figure out what's missing we really really need the community's help to address this issue. If you have to look something up in ruby docs bcause you don't understand it, please make an issue, or a comment on this issue to let us know. We can then work together on improving the documentation!

jhass:
Two years later, how does everybody feel about status quo? Is it worth to keep this issue open, can find some concrete actionable items to work on?

beta-ziliani:
I think the status quo didn't change much, but I agree it's best to have actionable items instead. Let's close it.

Fryguy:
[...]

PR #8378, opened 2019-10-25
Only add some example to document Enumerable.reduce behavior

jan-zajic:
I added spec test to ensure that i can pass different type of value as initial which I wasn't sure either. Also add example to doc for other newcomers.

See issue #8377 for more info.

### DIFF
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -569,6 +569,8 @@ describe "Enumerable" do
   describe "reduce" do
     it { [1, 2, 3].reduce { |memo, i| memo + i }.should eq(6) }
     it { [1, 2, 3].reduce(10) { |memo, i| memo + i }.should eq(16) }
+    it { [1, 2, 3].reduce([] of Int32) { |memo, i| memo.unshift(i) }.should eq([3, 2, 1]) }
+    it { [[0, 1], [2, 3], [4, 5]].reduce([] of Int32) { |memo, i| memo.concat(i) }.should eq([0, 1, 2, 3, 4, 5]) }
 
     it "raises if empty" do
       expect_raises Enumerable::EmptyError do
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -628,7 +628,8 @@ module Enumerable(T)
   # Just like the other variant, but you can set the initial value of the accumulator.
   #
   # ```
-  # [1, 2, 3, 4, 5].reduce(10) { |acc, i| acc + i } # => 25
+  # [1, 2, 3, 4, 5].reduce(10) { |acc, i| acc + i }             # => 25
+  # [1, 2, 3].reduce([] of Int32) { |memo, i| memo.unshift(i) } # => [3, 2, 1]
   # ```
   def reduce(memo)
     each do |elem|


src/enumerable.cr
```diff
@@ -630,6 +630,12 @@ module Enumerable(T)
   # ```
   # [1, 2, 3, 4, 5].reduce(10) { |acc, i| acc + i } # => 25
   # ```
+  #
+  # The initial value can be of any type
```
straight-shoota:
```suggestion
  # The initial value can be of any type, for example a collection:
```

spec/std/enumerable_spec.cr
```diff
@@ -569,7 +569,9 @@ describe "Enumerable" do
   describe "reduce" do
     it { [1, 2, 3].reduce { |memo, i| memo + i }.should eq(6) }
     it { [1, 2, 3].reduce(10) { |memo, i| memo + i }.should eq(16) }
-
+	it { [1, 2, 3].reduce([4]) { |memo, i| [i] + memo }.should eq([4,3,2,1]) }
```
straight-shoota: These lines need to be indented one more step. Running `crystal tool format` should fix that.

The result is incorrect, it would be `[3, 2, 1, 4]`. The same can be achieved with `unshift` instead of `+` and I'd rather use that because it's more efficient:
```suggestion
	it { [1, 2, 3].reduce([4]) { |memo, i| memo.unshift(i) }.should eq([3, 2, 1, 4]) }
```

src/enumerable.cr
```diff
@@ -630,6 +630,12 @@ module Enumerable(T)
...
+  #
+  # ```
+  # [1, 1, 2, 3, 2, 3, 1].reduce({} of Int32 => Int32) { |acc, i| acc.merge!({i => 1}){|key, value| value + 1} } # => {1 : 3, 2: 2, 3: 2}
```
yxhuvud: While having an example is nice, perhaps an example that is a bit easier to understand (and efficient - creating a new hash for each round in the iteration is not how one want to implement that even if it isn't as horrible as before the improvements @asterite did some month ago) might be a good idea here.

src/enumerable.cr
jan-zajic: Ok

spec/std/enumerable_spec.cr
jan-zajic: Ok, i use unshift and simplified it to: `[1, 2, 3].reduce([] of Int32) { |memo, i| memo.unshift(i) }.should eq([3, 2, 1])` (which is the original intention, simple example - reverse the array).

src/enumerable.cr
straight-shoota: This example is IMO still too complex. We're not showcasing `merge` here. Why not simply use one of the examples from the spec?

jan-zajic: Ok, that will be enough

src/enumerable.cr
```diff
@@ -630,6 +630,12 @@ module Enumerable(T)
   # ```
   # [1, 2, 3, 4, 5].reduce(10) { |acc, i| acc + i } # => 25
   # ```
+  #
+  # The initial value can be of any type, for example a collection:
```
RX14: I'd honestly rather remove this text and just put both examples together:
> Just like the other variant, but you can set the initial value of the accumulator.
>
> ```
> [1, 2, 3, 4, 5].reduce(10) { |acc, i| acc + i } # => 25
> [1, 2, 3].reduce([] of Int32) { |memo, i| memo.unshift(i) } # => [3, 2, 1]
> ```

jan-zajic: ok

wontruefree:
This is a handy little script.
https://github.com/crystal-lang/crystal/blob/master/CONTRIBUTING.md#git-pre-commit-hook

jan-zajic:
@RX14 it's formatted. @wontruefree known about this hook, but cannot run it, because i commiting on windows ..

PR #8379, opened 2019-10-25
Add missing documentation for ArrayLiteral.reduce with memo

jan-zajic:
Hi, i found while writing macros that ArrayLiteral don't have reduce method that would allow me to pass initial value. Until i try it, and voilà, it worked. So i look into the spec and there's even a test for it. So it's only missing in generated docs here: https://crystal-lang.org/api/Crystal/Macros.html. So i add it to the definition. As bonus i add test similar to Enumerable, see my other PR  #8378.

Related to #8377.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -580,6 +580,7 @@ module Crystal
 
       it "executes reduce with initial value" do
         assert_macro "", %({{[1, 2, 3].reduce(4) { |acc, val| acc * val }}}), [] of ASTNode, "24"
+        assert_macro "", %({{[1, 2, 3].reduce([] of NumberLiteral) { |acc, val| acc = [val]+acc }}}), [] of ASTNode, "[3, 2, 1]"
       end
 
       it "executes map with constants" do
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -634,6 +634,10 @@ module Crystal::Macros
     def reduce(&block) : ASTNode
     end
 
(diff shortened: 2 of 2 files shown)


spec/compiler/macro/macro_methods_spec.cr
```diff
@@ -582,6 +582,10 @@ module Crystal
         assert_macro "", %({{[1, 2, 3].reduce(4) { |acc, val| acc * val }}}), [] of ASTNode, "24"
       end
 
+      it "executes reduce with initial value of other type" do
```
straight-shoota: I'd just merge this example with the one above. It can simply have to expectations.

src/compiler/crystal/macros.cr
```diff
@@ -634,6 +634,10 @@ module Crystal::Macros
     def reduce(&block) : ASTNode
     end
 
+    # Similar to `Enumerable#reduce`
+    def reduce(memo : ASTNode, &block) : ASTNode
```
j8r:
```suggestion
    def reduce(memo : ASTNode, &) : ASTNode
```

The new syntax for non-captured blocks can now be used.

straight-shoota: The thing is, this makes it now different from the existing `#reduce` signature. It's probably better to stay with `&block` for both of them.

We should have a global run to replace all uncaptured `&block` arguments with `&`. But it's safer to do this after 0.32.0 is released so it still compiles with pre 0.31.0 compiler. (Not super important, but there is no hurry either)

jan-zajic: ok guys, i changed it there and back again, no problem .. Please just make some consensus.

Issue #3803, opened 2016-12-30
Covariance and contravariance


asterite:
Given:

```cr
class Foo
end

class Bar < Foo
end
```

And an array of `Bar`:

```cr
array = Array(Bar).new
```

We can ask the question: is `array` an `Array(Foo)`?

Two answers are possible:
1. Yes: if we read from `array` we get a `Bar`, which is a `Foo`
2. No: if we want to put an element into `array`, we can't put a `Foo`, only `Bar` and its subclasses

Right now the compiler will answer "Yes" for type restrictions:

```cr
def foo(x : Array(Foo))
end

foo(Array(Bar).new) # OK
```

But i will answer "No" for `is_a?`:

```cr
Array(Bar).new.is_a?(Array(Foo)) # => false
```

This is not intuitive and should be fixed somehow.

This also affects union types, aliases and recursive aliases.

mverzilli:
`is_a?` being false sounds anti-intuitive, but maybe it's right and it's coherent with Crystal.

``` crystal
class Foo 
end

class Bar < Foo
end

class Baz
  def zzz_array(something : Array(Foo))
    puts something
  end

  def insert_sth(something : Array(Foo))
    something.push(Foo.new)
  end
end

puts Array(Bar).new.is_a?(Array(Foo)) # returns "false"

array = [Bar.new, Bar.new] of Bar

puts Baz.new.zzz_array(array) # works perfectly fine as expected
puts Baz.new.insert_sth(array) # including this line causes the program to fail compilation.
```

Crystal always intended to mimic the development experience of duck typing while ensuring safety, and the current behavior honors that philosophy. Maybe we already have the best tradeoff for Crystal and just need to be careful in the way we teach users why `is_a?` will be false.

asterite:
@mverzilli You are totally right with the above :-)

It's like Crystal says "OK, I let you pass that `Array(Bar)` in this `Array(Foo)` restriction, as long as you use it well then I'm fine with that". Then if you only read from it, that's OK. But if you try to do something wrong, something that would break the type system (like putting a `Bar` inside an `Array(Foo)`) then you get a compile error.

Another scenario where this "I'll only yell at you when yo make a real mistake" is this one:

```cr
def foo(x : Enumerable(Int32))
  x << 4
end

array = [1, 2, 3]
foo(array)
p array # => [1, 2, 3, 4]
```

[...]

Issue #1693, opened 2015-10-05
Problem merging Hash with recursive types


jreinert:
``` crystal
alias MyType = String | Int32 | Array(MyType) | Hash(Symbol, MyType)
hash = {foo: "bar"} as Hash(Symbol, MyType)
hash.merge({foobar: "foo"})
```

will raise

```
Error in line 3: instantiating 'Hash(Symbol, MyType)#merge(Hash(Symbol, String))'

in /usr/lib/crystal/hash.cr:437: no overload matches 'Hash(Symbol, String | Int32 | Array(MyType) | Hash(Symbol, MyType))#merge!' with types Hash(Symbol, MyType)
Overloads are:
 - Hash(Symbol, String | Int32 | Array(MyType) | Hash(Symbol, MyType))#merge!(other : Hash(K, V))
 - Hash(Symbol, String | Int32 | Array(MyType) | Hash(Symbol, MyType))#merge!(other : Hash(K, V), &block)

    hash.merge! self
         ^~~~~~
```

wmoxam:
Updated example for 0.17

    alias MyType = String | Int32 | Array(MyType) | Hash(Symbol, MyType)
    hash = {:foo => "bar"} as Hash(Symbol, MyType)
    hash.merge({:foobar => "foo"})

https://play.crystal-lang.org/#/r/zpf

PR #8380, opened 2019-10-25
Add missing documentation to Hash.merge!(other : Hash, &block)

jan-zajic:
Completely missing description in api docs for Hash.merge! method, version with block. Without description, one cannot known which arguments are passed to given block (Except Ruby programmers, of course, because it behave same).
Related to #8377.

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1343,7 +1343,17 @@ class Hash(K, V)
     self
   end
 
-  def merge!(other : Hash, &block)
+  # Adds the contents of *other* to this hash.
+  # If a key exists in both hashes, the given block is called to determine the value to be used.
+  # The block arguments are the key, the value in `self` and the value in *other*.
+  #
+  # ```
+  # hash = {"a" => 100, "b" => 200}
+  # other = {"b" => 254, "c" => 300}
+  # hash.merge!(other) { |key, v1, v2| v1 + v2 }
+  # hash # => {"a" => 100, "b" => 454, "c" => 300}
+  # ```
+  def merge!(other : Hash, &block) : self
     other.each do |k, v|
       if self.has_key?(k)
         self[k] = yield k, self[k], v


src/hash.cr
```diff
@@ -1343,6 +1343,15 @@ class Hash(K, V)
     self
   end
 
+  # Adds the contents of *other* to this hash. 
+  # The value of each duplicate key is determined by calling the block with the key, its value in this hash and its value in *other*.
+  #
+  #```
+  # hash = { "a" => 100, "b" => 200 }
+  # other = { "b" => 254, "c" => 300 }
+  # hash.merge!(other) { |key, v1, v2| v1 }
```
straight-shoota: Maybe return `v1 + v2` to show some different behaviour? Could also be an additional example.

src/hash.cr
```diff
@@ -1343,6 +1343,15 @@ class Hash(K, V)
     self
   end
 
+  # Adds the contents of *other* to this hash. 
+  # The value of each duplicate key is determined by calling the block with the key, its value in this hash and its value in *other*.
```
straight-shoota: Maybe a little bit more verbose description, but help understanding:
```suggestion
  # If a key exists in both hashes, the given block is called to determine the value to be used. The block arguments are the key, the value in `self` and the value in *other*.
```

src/hash.cr
```diff
@@ -1343,6 +1343,15 @@ class Hash(K, V)
...
+  #
+  #```
+  # hash = { "a" => 100, "b" => 200 }
+  # other = { "b" => 254, "c" => 300 }
+  # hash.merge!(other) { |key, v1, v2| v1 }
+  # hash # => {"a" => 100, "b" => 200, "c" => 300}
+  #```
   def merge!(other : Hash, &block)
```
straight-shoota: You can also add documentation for the block types:
```suggestion
   def merge!(other : Hash, & : (K, V, V) -> V) : self
```

jan-zajic: Are you sure? Types were specified, but removed in https://github.com/crystal-lang/crystal/commit/3c17d0708a49a6d2b931b0f6a7b808110bb2f440 (#1693). And it's sad for me, i read discussion here and say wow. It's seems like a very big bug for me - someone should fix this recursive types or it must be removed from lang, if they block clear type safety like here...

src/hash.cr
jan-zajic: You're right i chaged it.

src/hash.cr
jan-zajic: You're right i chaged it.

jan-zajic:
And now i found that i forgot non bang version with block, def merge(other : Hash(L, W), &block : K, V, W -> V | W) forall L, W. But when i try it, i found maybe a bug, at least it's very confusing. It's documented to accept other hash with any types of keys and values, but block is defined to receive type of key K as in self, but if you look at implementation, it iterates second one so it can't compile at all.
Try for example this:
```
h1 = {"a" => 1, "b" => 2, "c" => 3}
h2 = {:a => 1, :b => 2, :c => 3}
h3 = h1.merge(h2){|k,v1,v2| v1||v2}
```
which leads to:
`Error: argument #1 of yield expected to be String, not Symbol`
In my opinion it must be reimplemented or changed to: `def merge(other : Hash(K, W), &block : K, V, W -> V | W) forall W`
Than we should add documentation to it.

src/hash.cr
straight-shoota: Wow, amazing that you digged up that old issue.

But anyway, even without recursive aliases, the type restriction is incorrect, because the generic arguments of `other` don't need to equal those of `self`, they can also be smaller.
For example you can merge `Hash(String, Bool)` into `Hash(String | Int32, Bool | Nil)`. This can't be expressed currently, but is tracked in #3803. So, yeah let's leave it for now.

src/hash.cr
```diff
@@ -1343,7 +1343,17 @@ class Hash(K, V)
...
+  # If a key exists in both hashes, the given block is called to determine the value to be used.
+  # The block arguments are the key, the value in `self` and the value in *other*.
+  #
+  # ```
+  # hash = {"a" => 100, "b" => 200}
+  # other = {"b" => 254, "c" => 300}
+  # hash.merge!(other) { |key, v1, v2| v1 + v2 }
+  # hash # => {"a" => 100, "b" => 454, "c" => 300}
+  # ```
+  def merge!(other : Hash, &block) : self
```
```suggestion
  def merge!(other : Hash, &) : self
```
The new syntax for non-captured blocks can now be used.

src/hash.cr
j8r: At least the new syntax for non-captured blocks can be used.

RX14:
@jan-zajic how about 

```cr
def merge(other : Hash(K2, V2), &block : K | K2, V, V2 -> V | V2) forall K2, V2
```

?

src/hash.cr
```diff
@@ -1343,7 +1343,17 @@ class Hash(K, V)
     self
   end
 
-  def merge!(other : Hash, &block)
+  # Adds the contents of *other* to this hash.
```
Blacksmoke16:
```suggestion
  # Adds the contents of *other* to `self`.
```

straight-shoota:
@RX14
> ```crystal
> def merge(other : Hash(K2, V2), & : (K | K2, V, V2) -> V | V2) forall K2, V2
> ```

That would work, but I think it's to complicated. The type restriction is not necessary for technical reasons, but mainly for documentation. For improving the documentation, this looks too complex IMO. Only benefit would be more precise error messages when using this method with non-matching types.

Btw. the key in the block argument is not `K | K2` but only `K` because only keys from `self` are yielded.

RX14:
> They must be covariant with `K, V`.

Ah, I didn't look at the implementation....

```cr
class Hash(K, V)
  def merge2(other : Hash(K2, V2), & : (K2, V, V2) -> V | V2) : Hash(K | K2, V | V2) forall K2, V2
    hash = Hash(K | K2, V | V2).new
    hash.merge! self
    
    other.each do |k2, v2|
      if self.has_key?(k2)
        hash[k2] = yield k2, self[k2], v2
      else
        hash[k2] = v2
      end
    end
  
    hash
  end
end
```

This allows for the full type signature.

Another PR though...

straight-shoota:
@RX14 Your code looks good. But that's actually `#merge` not `#merge!`. And it matches its current implementation 😄

RX14:
@straight-shoota We were talking about the non-bang version the whole time! I was responding to @jan-zajic's comment about the non-bang version. Also, my implementation is different to the current implementation in that it uses `self[k2]` instead of `hash[k2]` which restricts the key type sent to the block from `K | K2` to `K2`.

jan-zajic:
@straight-shoota @RX14 yes that https://github.com/crystal-lang/crystal/pull/8380#issuecomment-546398715 comment ;). Well that this is already merged. We can continue that discussion in some issue or in the forum ...

RX14:
Yeah, I was planning to make this another PR.

straight-shoota:
Ooops, I think I got lost somewehere.

PR #8389, opened 2019-10-29
Allow constants to use ditto

### LINKED ISSUE #8388: Ditto doesn't work for constants
When documenting code, "# ditto" appears to work for methods, but not constants.

Reproducible case:

```
module Xyz

  # Description of method X
  def x; end

  # ditto
  def xdittoed; end

  # Description of constant Cons
  Cons = 1

  # ditto
  ConsDittoed = 1
end
```

After running `crystal docs` and opening `docs/index.html` in the browser, both methods and the first constant will be documented, while the second constant will just say "ditto".

Blacksmoke16:
Fixes #8388

### DIFF
--- a/spec/compiler/semantic/doc_spec.cr
+++ b/spec/compiler/semantic/doc_spec.cr
@@ -67,7 +67,7 @@ describe "Semantic: doc" do
       # A number
       ONE = 1
 
-      # ditto
+      # :ditto:
       TWO = 2
     ), wants_doc: true
     program = result.program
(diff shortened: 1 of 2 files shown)


spec/compiler/semantic/doc_spec.cr
```diff
@@ -62,6 +62,19 @@ describe "Semantic: doc" do
     bar.doc.should eq("Hello")
   end
 
+  it "stores doc for const when using ditto" do
+    result = semantic %(
+      # A number
+      ONE = 1
+
+      # ditto
```
j8r:
```suggestion
      # :ditto:
```
Both syntax are valid, but without colons will eventually go away with https://github.com/crystal-lang/crystal/pull/6362

docelic: In that case also page https://crystal-lang.org/reference/conventions/documenting_code.html should be updated to describe both ":ditto:" and "ditto" (currently it only mentions "ditto" without colons).

PR #8391, opened 2019-10-29
Add doc to Hash select methods - splat variants

jan-zajic:
n/c

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1418,7 +1418,9 @@ class Hash(K, V)
   # Returns a new `Hash` with the given keys.
   #
   # ```
-  # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select("a", "c") # => {"a" => 1, "c" => 3}
+  # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select({"a", "c"}) # => {"a" => 1, "c" => 3}
+  # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select("a", "c")   # => {"a" => 1, "c" => 3}
+  # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select(["a", "c"]) # => {"a" => 1, "c" => 3}
   # ```
   def select(keys : Array | Tuple)
     hash = {} of K => V
@@ -1426,6 +1428,7 @@ class Hash(K, V)
     hash
   end
 
+  # :ditto:
   def select(*keys)
     self.select(keys)
   end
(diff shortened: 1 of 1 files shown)


jhass:
Maybe we should just `ditto` them?

straight-shoota:
@jhass But then the code sample would not fit.

In fact, the example for `select(keys : Array | Tuple)` is actually demonstrating `select(*keys)` 🤦‍♂

jhass:
For cases like this I wouldn't mind to have code samples demonstrating all the overloads on the first one and the others just `ditto`ing them.

straight-shoota:
Yeah, that's probably fine.

straight-shoota:
We might even consider some kind of combined presentation for methods with the same (`ditto`) combination. But that's not particularly important.

straight-shoota:
@jan-zajic Would you mind adding a usage example for each `select(keys : Array | Tuple)` and `select!(*keys)` to the dittoed doc string?

jan-zajic:
@straight-shoota like this?

src/hash.cr
```diff
@@ -1418,29 +1418,36 @@ class Hash(K, V)
...
   # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select("a", "c") # => {"a" => 1, "c" => 3}
+  # {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select(["a", "c"]) # => {"a" => 1, "c" => 3}
   # ```
   def select(keys : Array | Tuple)
     hash = {} of K => V
     keys.each { |k| hash[k] = self[k] if has_key?(k) }
     hash
   end
 
+  # ditto
```
bcardiff:
```suggestion
  # :ditto:
```

src/hash.cr
```diff
@@ -1418,29 +1418,36 @@ class Hash(K, V)
...
+  # h3 = {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select!(["a", "c"])
+  # h1 == h2 == h3
+  # h1 # => {"a" => 1, "c" => 3}
   # ```
   def select!(keys : Array | Tuple)
     each { |k, v| delete(k) unless keys.includes?(k) }
     self
   end
 
+  # ditto
```
bcardiff:
```suggestion
  # :ditto:
```

src/hash.cr
```diff
@@ -1418,29 +1418,36 @@ class Hash(K, V)
...
 
   # Removes every element except the given ones.
   #
   # ```
-  # h = {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select!("a", "c")
-  # h # => {"a" => 1, "c" => 3}
+  # h1 = {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select!({"a", "c"})
+  # h2 = {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select!("a", "c")
+  # h3 = {"a" => 1, "b" => 2, "c" => 3, "d" => 4}.select!(["a", "c"])
+  # h1 == h2 == h3
```
Sija:
```suggestion
  # h1 == h2 == h3 # => true
```

straight-shoota:
Needs a format before being able to merge.

jan-zajic:
@straight-shoota it's done.

PR #8392, opened 2019-10-29
New `http::WebSocket#read` to avoid running an infinite loop with `#run`

KaneRoot:
Hello. This is a new function `#read` for the WebSocket class, allowing the user to avoid the `#run` function, which loops indefinitely.

It is useful when there are more than just websockets to listen to. For instance, I use it along with a C library that uses a `select` function internally, to which I provide the websocket file descriptors, so any network event (websocket or not) is handled the same way.

Thanks

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -148,6 +148,64 @@ class HTTP::WebSocket
       end
     end
   end
+
+  record Ping
+  record Pong
+  record Close
+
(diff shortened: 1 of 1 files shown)


bcardiff:
* It could be called `run_once` maybe since the new method is not only reading but honoring the protocol.
* I would add `:nodoc:`.
* But more important, there is duplication introduced with `run`, so that should be refactored.

z64:
I believe #8024 addresses this PR's subject (#5600)

PR #8394, opened 2019-10-29
Change `&block` to `&` for yielding method in docs

j8r:
Since https://github.com/crystal-lang/crystal/pull/8117 is merged, anonymous block arguments can be `&` instead of `&block`, including in docs.


Blacksmoke16:
Shouldn't this support both?  `&block` is still a valid thing when you want to capture the block; while `&` should just be used if its non-capturing.

straight-shoota:
@Blacksmoke16 This only affects methods that yield but don't have a block arg in their signature. Capturing methods always have a block arg, so they're not affected by this change.

PR #8395, opened 2019-10-29
Remove debug-leftovers

Sija:
Followup to #8308


straight-shoota:
Thanks @Sija!

wontruefree:
I noticed this in some debugging and was curious if there was a reason for it.

Issue #6417, opened 2018-07-20
Missing Crystal::Macros::TypeNode#class_vars


vypxl:
Crystal 0.25.1 (2018-06-29)

LLVM: 6.0.0
Default target: x86_64-pc-linux-gnu

I cannot find anyone pointing that out, nor a method on how to get the class variables of some Type in a macro. Should behave like Typenode#instance_vars.

PR #8405, opened 2019-10-30
Get the class variables of some Type in a macro

jan-zajic:
This PR fix issue #6417.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -1,5 +1,15 @@
 require "../../spec_helper"
 
+private def declare_class_var(container : ClassVarContainer, name, var_type : Type, annotations = nil)
+  var = MetaTypeVar.new(name)
+  var.owner = container
+  var.type = var_type
+  var.annotations = annotations
+  var.bind_to var
+  var.freeze_type = var_type
+  container.class_vars[name] = var
+end
+
 module Crystal
   describe "macro methods" do
     describe "node methods" do
@@ -1097,7 +1107,7 @@ module Crystal
       it "executes class vars" do
         assert_macro("x", "{{x.class_vars.map &.name}}", %([class_var])) do |program|
           klass = NonGenericClassType.new(program, program, "SomeType", program.reference)
-          klass.declare_class_var("@@class_var", program.string)
+          declare_class_var(klass, "@@class_var", program.string)
           [TypeNode.new(klass)] of ASTNode
         end
       end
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/macros/methods.cr
```diff
@@ -1704,6 +1706,15 @@ module Crystal
       end
     end
 
+    def self.class_vars(type)
+      ArrayLiteral.map(type.all_class_vars) do |name, ivar|
```
bcardiff: There is difference in the design that should be kept aligned: `instance_vars` performs de check of `InstanceVarContainer` and not in the `all_instance_vars`.

src/compiler/crystal/types.cr
```diff
@@ -557,6 +557,18 @@ module Crystal
       end
     end
 
+    def all_class_vars
+      all_class_vars = {} of String => MetaTypeVar
+      if (self.is_a?(ClassVarContainer))
```
straight-shoota:
```suggestion
      if container = self.as?(ClassVarContainer)
        container.class_vars?.try { |v| all_class_vars.merge!(v) }
```

src/compiler/crystal/types.cr
```diff
@@ -557,6 +557,18 @@ module Crystal
...
+        self.as(ClassVarContainer).class_vars?.try { |v| all_class_vars.merge!(v) }
+      end
+      ancestors.each do |ancestor|
+        next unless ancestor.is_a?(ClassVarContainer)
```
straight-shoota:
```suggestion
        if ancestor.is_a?(ClassVarContainer)
          ancestor.class_vars?.try { |v| all_class_vars.merge!(v) }
```

src/compiler/crystal/types.cr
```diff
@@ -974,6 +987,16 @@ module Crystal
 
       nil
     end
+
+    def declare_class_var(name, type : Type, annotations = nil)
```
bcardiff: I just noticed that `delcare_class_var` is only used in specs, so to avoid reviewing all the places where `ClassVarContainer` is included I would suggest moving this method to a private helper in the spec.

jan-zajic: Something like that?

spec/compiler/macro/macro_methods_spec.cr
```diff
@@ -1094,6 +1104,71 @@ module Crystal
...
+        assert_macro("x", "{{x.class.class_vars.map &.stringify}}", %([])) do |program|
+          klass = NonGenericClassType.new(program, program, "SomeType", program.reference)
+          declare_class_var(klass, "@@class_var", program.string)
+          [TypeNode.new(klass)] of ASTNode
+        end
+      end
+
+      it "executes instance_vars on symbol literal" do
+        assert_macro("x", "{{x.instance_vars.map &.stringify}}", %([])) do |program|
+          [TypeNode.new(SymbolLiteralType.new(program, "foo".symbol))] of ASTNode
```
asterite, line 1146: This type is used for autocasting, I think this test makes no sense (same as the next one) so I'm going to remove it.

PR #8406, opened 2019-10-30
JSON/YAML: add `use_discriminator`

asterite:
**Note:** This is a PR that also serves as an RFC. I decided to go the PR route because I wanted to provide code that proves that this is possible, and show how it could be done.

## What

The idea of this PR is to provide a way to deserialize a group of objects based on a field. This is **very** typical in JSON APIs. For example:

```json
[
  {
    "type": "point",
    "x": 1,
    "y": 2
  },
  {
    "type": "circle",
    "x": 3,
    "y": 4,
    "radius": 5
  }
]
```

Here the parser expects to decide what properties each object has based on the value of the "type" property.

There's no nice way to do this. [It's definitely possible](https://forum.crystal-lang.org/t/deserialize-from-yaml-based-on-a-propertys-value/1219) but since this use case is so common it would be great to provide built-in support for it.

With this PR it would work like this (it already works ga:

```crystal
require "json"

abstract class Shape
  include JSON::Serializable

  use_discriminator "type", {
    point: Point,
    circle: Circle,
  }

  property type : String
end

class Point < Shape
  property x : Int32
  property y : Int32
end

class Circle < Shape
  property x : Int32
  property y : Int32
  property radius : Int32
end

shapes = Array(Shape).from_json(%(
[
  {
    "type": "point",
    "x": 1,
    "y": 2
  },
  {
    "type": "circle",
    "x": 3,
    "y": 4,
    "radius": 5
  }
]
))
pp shapes
```

Output:

```
[#<Point:0x109b16dc0 @type="point", @x=1, @y=2>,
 #<Circle:0x109b20e70 @radius=5, @type="circle", @x=3, @y=4>]
```

### DIFF
--- a/src/json/serialization.cr
+++ b/src/json/serialization.cr
@@ -123,7 +123,7 @@ module JSON
   # [GeoJSON](https://tools.ietf.org/html/rfc7946) each object has a "type"
   # field, and the rest of the fields, and their meaning, depend on its value.
   #
-  # You can use `JSON::Serializable.use_discriminator` for this use case.
+  # You can use `JSON::Serializable.use_json_discriminator` for this use case.
   module Serializable
     annotation Options
     end
@@ -366,7 +366,7 @@ module JSON
     # abstract class Shape
     #   include JSON::Serializable
     #
-    #   use_discriminator "type", {point: Point, circle: Circle}
+    #   use_json_discriminator "type", {point: Point, circle: Circle}
     #
     #   property type : String
     # end
@@ -385,7 +385,7 @@ module JSON
     # Shape.from_json(%({"type": "point", "x": 1, "y": 2}))               # => #<Point:0x10373ae20 @type="point", @x=1, @y=2>
     # Shape.from_json(%({"type": "circle", "x": 1, "y": 2, "radius": 3})) # => #<Circle:0x106a4cea0 @type="circle", @x=1, @y=2, @radius=3>
     # ```
-    macro use_discriminator(field, mapping)
+    macro use_json_discriminator(field, mapping)
       {% unless mapping.is_a?(HashLiteral) || mapping.is_a?(NamedTupleLiteral) %}
         {% mapping.raise "mapping argument must be a HashLiteral or a NamedTupleLiteral, not #{mapping.class_name.id}" %}
       {% end %}
--- a/src/json/serialization.cr
+++ b/src/json/serialization.cr
@@ -2,9 +2,6 @@ module JSON
   annotation Field
   end
 
-  annotation Discriminator
-  end
-
   # The `JSON::Serializable` module automatically generates methods for JSON serialization when included.
   #
   # ### Example
@@ -118,6 +115,15 @@ module JSON
   #   @a : Int32?
   # end
   # ```
+  #
+  # ### Discriminator field
+  #
+  # A very common JSON serialization strategy for handling different objects
+  # under a same hierarchy is to use a discriminator field. For example in
+  # [GeoJSON](https://tools.ietf.org/html/rfc7946) each object has a "type"
+  # field, and the rest of the fields, and their meaning, depend on its value.
+  #
+  # You can use `JSON::Serializable.use_discriminator` for this use case.
   module Serializable
     annotation Options
     end
@@ -127,33 +133,7 @@ module JSON
       # so it overloads well with other possible initializes
 
       def self.new(pull : ::JSON::PullParser)
-        {% verbatim do %}
-          {% begin %}
-            {% discriminators = @type.instance_vars.select { |ivar| ivar.annotation(::JSON::Discriminator) } %}
-            {% if !discriminators.empty? %}
-              {% if discriminators.size > 1 %}
-                {% raise "Multiple JSON::Discriminator annotations found for type #{@type}, instance variables #{discriminators.map { |var| "@#{var}" }.join(", ").id}" %}
-              {% end %}
-              {% discriminator = discriminators[0] %}
-              {% mapping = discriminator.annotation(::JSON::Discriminator)[0] %}
-              {% unless mapping.is_a?(HashLiteral) || mapping.is_a?(NamedTupleLiteral) %}
-                {% raise "Argument to JSON::Discriminator must be a HashLiteral ot NamedTupleLiteral, not #{mapping}" %}
-              {% end %}
-              any = JSON::Any.new(pull)
-              discriminator_value = any[{{discriminator.id.stringify}}].to_s
-              case discriminator_value
-              {% for key, value in mapping %}
-                when {{key.id.stringify}}
-                  {{value.id}}.from_json(any.to_json)
-              {% end %}
-              else
-                raise "Unexpected {{discriminator.id}}: #{discriminator_value}"
-              end
-            {% else %}
-              new_from_json_pull_parser(pull)
-            {% end %}
-          {% end %}
-        {% end %}
+        new_from_json_pull_parser(pull)
       end
 
       private def self.new_from_json_pull_parser(pull : ::JSON::PullParser)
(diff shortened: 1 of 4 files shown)


straight-shoota:
To me this seems too advanced for the standard library. This PR only provides a very basic implementation for such a discrimination feature. There can be quite more complicated cases than just a 1:1 mapping from an identifier in a single property to a type. Identifiers can be more complex values, or split over multiple properties and you might need handling for undefined values, etc. It feels like if we start providing this, it will inevitably need to be enhanced more and more because new use cases ask for new features.

In my opinion the standard library should have a basic implementation of a JSON mapping. And that's good as it stands (obviously improvements can be made here and there). But it's better to keep it at some basic level without fancy features that may be desirable for some use cases but add lots of complexity to the base implementation.

lribeiro:
I like it. 
Is there a way to leverage @type.subclasses, 
And do convention over configuration here? 

So in the simplest case we could define the discriminator just with?

```crystal
@[JSON::Discriminator]
```

Nirvana would be going a step further and having

```
include JSON::Serializable 
```
Automatically select the field type if availability to be the discriminator.

Don't like the name, but nothing better comes to mind

wontruefree:
This is a cool idea I like it.  I have not used this feature before.

asterite:
I propose just this because the only time people wanted to do a custom deserialization was because of a discriminator. Maybe we can explore existing JSON APIs, or someone can provide the more complex cases that this simple discriminator doesn't cover.

src/json/serialization.cr
```diff
@@ -124,6 +127,36 @@ module JSON
...
+            {% if !discriminators.empty? %}
+              {% if discriminators.size > 1 %}
+                {% raise "Multiple JSON::Discriminator annotations found for type #{@type}, instance variables #{discriminators.map { |var| "@#{var}" }.join(", ").id}" %}
+              {% end %}
+              {% discriminator = discriminators[0] %}
+              {% mapping = discriminator.annotation(::JSON::Discriminator)[0] %}
+              {% unless mapping.is_a?(HashLiteral) || mapping.is_a?(NamedTupleLiteral) %}
+                {% raise "Argument to JSON::Discriminator must be a HashLiteral ot NamedTupleLiteral, not #{mapping}" %}
+              {% end %}
+              any = JSON::Any.new(pull)
```
RX14: The best way to do this is to do
```cr
data = IO::Memory.new
JSON.build(data) do |builder|
  builder.start_object
  pull.read_object do |key|
    if key == {{discriminator.id.stringify}}
      discriminator_value = pull.read_string
    end
    builder.field(key) { pull.read_raw(builder) }
  end
  builder.end_object    
end
data.rewind

...

```

Blacksmoke16:
Adding my thoughts from gitter.

I would rather see it be at the type level, it's feature of the type, not a specific property.  You could even implement it where it would default to `type` if one is not provided.  I don't think the argument of "you won't forget adding the property" is enough.  A compile time error could be raised if you forget it

vlazar:
Names like `serialize` or `stringify` are quire common, how about `JSON::Typeize` or better yet `JSON::Typify`?

watzon:
Doesn't seem to work with a JSON structure like this:

```json
{
  "type": "updateAuthorizationState",
  "authorization_state": {
    "type": "authorizationStateWaitTdlibParameters"
  }
}
```

Where the relevant classes look like this:

```crystal
# authorizationState
class AuthorizationState
end

# updateAuthorizationState
class Update::AuthorizationState < Update
  property authorization_state : AuthorizationState
end

# authorizationStateWaitTdlibParameters
class AuthorizationState::WaitTdlibParameters < AuthorizationState
end
```

Rather than being parsed as `Update::AuthorizationState(authorization_state: AuthorizationState::WaitTdlibParameters)` it ends up being `Update::AuthorizationState(authorization_state: AuthorizationState)`

src/json/serialization.cr
jhass: In theory we could optimize further by building a `IO::Union` (basically extract the `IO` concatenation feature from  `ARGF`). Then break the above loop as soon as the discriminator is found and pass a `JSON::PullParser.new(IO::Union.new(data, pull.io))`  to the child.

j8r:
Agree with @straight-shoota.
This feature will be implemented, then others too for JSON and perhaps SOAP/XML, etc.
The stdlib won't be a base set of utilities anymore.

What we definitely need is a Crystal serialization framework for this kind advanced stuff :pray:

asterite:
I think I found a way to do this without an annotation and without modifying the existing code base (well, just slightly to fix something else). We'll see.

src/json/serialization.cr
asterite: @RX14 I think that's the best way to do it because we retain the original JSON string. My way of using `JSON::Any` is a bit broken with, for example, large numbers.

asterite:
@watzon Could you provide working code that I can compile and run? I did some tests and all seem to work fine. I don't know what's the full hierarchy you are working with and what's the input JSON you are trying it with. It doesn't have to big an entire hierarchy, just enough to reproduce the problem.

watzon:
@asterite updated with a working example. Here's the runnable code https://carc.in/#/r/7xd0

asterite:
@watzon `AuthorizationState` is not abstract plus it doesn't have a `Discriminator` field, so the JSON serializer will deserialize it as usual. You'd have to say that this type should also use a discriminator to deserialize. In the final code that I have you would solve this by calling `use_discriminator` on those nested types too.

asterite:
It also seems authorization state is abstract but doesn't derive from any other type: https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1_authorization_state.html . So I think it would make sense to use a discriminator just there, and avoid having that `Types::Base` type. Or is there such a thing in the telegram API where `type` can be any object in the system?

watzon:
Ahh ok. So make it abstract and add the discriminator to it as well?

asterite:
Yes, though I'm not sure it will work with the code in this PR. But it does work with the code I have in my machine.

watzon:
Hmm yeah, just making `AuthorizationState` abstract breaks things

asterite:
I pushed a new commit with a totally different API. I also updated the original PR's text.

Now the PR introduces a `use_discriminator` macro where you pass the key and mapping. This in turn redefines the `new` method to deserialize based on the discriminator.

I believe this way is superior to the annotation-based one I proposed before because:
- No need for a new annotation
- It's unobtrusive: it will redefine a method to behave in a different way (the old code modified existing code)
- It serves as an example for people wanting to do custom deserialization: just look at what `use_discriminator` does (and there's no need to use a macro in all cases, but here's it's convenient)

I also added docs and examples.

As for adding this or not, I believe this pattern is very, very common in the wild. For example it's used in [GeoJSON](https://tools.ietf.org/html/rfc7946). @watzon already showed it's used in Telegram's tdlib API. It's so common that [OpenAPI 3.0 (formerly Swagger) supports it too](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/) . I'm sure there are other examples, if you know of them please let me know to support merging this PR 🙏

src/json/serialization.cr
jhass: Sure, don't feel strong about the naming.

straight-shoota: I suppose the discriminator field is likely the first one 80% of the time. In that case, optimization would be pretty simple, because we could essentially just delegate the pull parser to the subtype (there only needs to be some way to skip/repeat the start object event, but that's doable). This completely avoids the entire parse-build loop.

wontruefree:
What about using `VARY` or `VARY_ON` kind of like a `Vary` header so something like.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary

```crystal
require "json"

abstract class Shape
  include JSON::Serializable

  JSON::VARY_ON "type", {
    point: Point,
    circle: Circle,
  }

  property type : String
end

class Point < Shape
  property x : Int32
  property y : Int32
end

class Circle < Shape
  property x : Int32
  property y : Int32
  property radius : Int32
end
```

kostya:
`use_discriminator` in class text not so informative when you read code, may be `json_discriminator`. and also would it work when json and yaml together?

straight-shoota:
Maybe a better name would be `def_json_discriminator` similar to all the other `def_` macros for defining methods.

> would it work when json and yaml together?

Yes, they generate completely different method overloads.

asterite:
I agree about choosing a better name. But usually def_x defines an x method (def_hash, etc.) But I'm open to suggestions. use_json_discriminator is good too.

watzon:
@asterite with the recent change to using `use_discriminator` I'm getting an error:

```
Invalid memory access (signal 11) at address 0x7ffc4d469f18
[0x7fbb22891fc6] ???
[0x7fbb227e7f4e] ???
[0x7fbb2365f461] ???
```

with my massive hash. Any idea why that might be? [Here](https://github.com/watzon/proton/blob/master/src/proton/types/base.cr) is the class containing the hash, and [here](https://github.com/watzon/proton/blob/master/src/proton.cr#L55) is where I'm trying to parse the JSON.

asterite:
@watzon Two things:
- I tried compiling the project but got:
  ```
  In src/proton.cr:4:1

   4 | require "./core_extensions/*"
       ^
  Error: can't find file './core_extensions/*' relative to '/Users/asterite/Sandbox/proton/src'
  ```
  I guess you have `core_extensions` locally to override how JSON works. But this will not work because in `JSON::Serializable` there are inherited and included hooks and there's no way to replace this. Maybe this is why you are getting a strange crash?
- I removed that `require` line and used `bin/crystal` from the compiler at this branch, and that code compiles and runs fine.

> Now I'm getting another error:

For this you need to `use_discriminator` in intermediate types, like `Proton::Types::RichText`, otherwise I think in your case it's using the parent's discriminator (which it shouldn't, but it does because you are trying to reopen and change `JSON::Serializable` which won't work).

asterite:
@kostya 

> and also would it work when json and yaml together?

You can do:

```crystal
class SomeClass
  JSON::Serializable.use_discriminator
  YAML::Serializable.use_discriminator
end
```

I chose `use_discriminator` because in most cases you won't have both serialization strategies and `use_discriminator` is short and straight-forward. But I agree that a name that include `json` or `yaml` is better.

Maybe `def_new_using_json_discriminator`, etc.

asterite:
Cool!

I think I'll rename them to `use_json_discriminator` and `use_yaml_discriminator` to make it clearer they are related to JSON/YAML, and also to avoid collisions in case you want to use both (though you could always do `JSON::Serializable.use_discriminator`, etc.)

straight-shoota:
In chat, @yxhuvud mentioned that all the examples (and specs) use a common superclass for the different choices. If that was necessary it might be a limiting factor for being useful in practice.

But the implementation should really work without an inheritance relationship between the individual types You just need any type to define the discriminator method on, the mapping types don't need to be related.
Maybe there should be an example in the docs and/or a spec for that?

asterite:
It's a good point, but I can't see how it would work without at least some inheritance relationship.

Can you provide an example, together with code for serializing and deserializing?

For example if we take the `Point` and `Circle` example, let's say I have an object with a property of that type... how can I talk about these types without enumerating them? I need a supertype. So for example if I do `Array(Shape).from_json` I would expect that to return `Array(Shape)` not an `Array(Point | Circle)`... if it's the latter then it'll be incompatible when assigning it to `Array(Shape)` if those types are not `Shape` themselves.

Maybe you'd like the base type to be a module? That could work, but it currently doesn't. But PRs are welcome!

straight-shoota:
This kind of discrimination is completely independent of type relationships. It's common to implement this using inheritance, but that's actually not really required.

In fact, the discriminator implementation doesn't need to be tied to a single type's `from_json`/`from_yaml` deserializer. It can just be implemented at runtime and you can call it like `JSON::Serializable.discriminate(pull, "type", {point: Point, circle: Circle})`. This method would just return `Point | Circle` and these types don't need to have a shared ancestor.

It's trivial to change the implementation to runtime. There is actually no real need for a macro.
https://github.com/straight-shoota/crystal/commit/581d0ad3ddd094e3889d6ec03d5c5b9c426bb2eb?w=1

This is a little change that gives more options. You can use discrimination rules for different context without having to declare a common supertype for each discrimination type per context.

asterite:
Nice! Well, if it works equally fine you can send a PR. I'm still not exactly sure it will always work fine at runtime, but we'll see.

straight-shoota:
The heavy lifting is actually all happening at runtime. The only difference the macro makes is to unfold the mappings so the keys are matched in a `case` vs. a lookup using `#[]?` at runtime. When supplying a NamedTuple, it results in exactly the same kind of unfold.

PR #8412, opened 2019-10-31
Spec: restrict the type returned by `should_not be_nil` and others

### LINKED ISSUE #8369: Spec: cast values for be_nil and be_a expectations
A way to solve [this forum request](https://forum.crystal-lang.org/t/var-should-not-be-nil-does-not-eliminate-var-of-nil-from-subsequent-type-inference/1165)

Currently, when validating the type of a value in spec, we need to restrict the type afterwards in order to call restricted methods on the sample:

```crystal
index = "foo".index('o')
index.should_not be_nil
foo[index.not_nil!].should eq('o')
```

The `not_nil!` is needed because the compiler has no way to know that the `should_not be_nil` means that `index` will not be `nil`.

The idea is to let `be_nil` and `be_a` expectations return the value with appropriate type restriction:
```crystal
index = "foo".index('o')
index = index.should_not be_nil
foo[index].should eq('o')
```

Basically, the result of `obj.should_not be_nil` will return `obj` as a non-nil object.

The same applies for `obj.should be_a(T)` and `obj.should_not be_a(T)`.

An implementation was submitted in #8240, but it should be overhauled.

asterite:
Fixes #8369

Follow up to #8240

### DIFF
--- a/spec/std/spec/expectations_spec.cr
+++ b/spec/std/spec/expectations_spec.cr
@@ -23,6 +23,20 @@ describe "expectations" do
     it { 100_000.should_not be_a(String) }
     it { 100_000.should be_a(Int32) }
     it { "Hello".should_not be_a(Int32) }
+
+    it "restricts type on should" do
+      x = 1 || 'a'
+      y = x.should be_a(Int32)
+      typeof(x).should eq(Int32 | Char)
+      typeof(y).should eq(Int32)
+    end
+
+    it "restricts type on should_not" do
+      x = 1 || 'a'
+      y = x.should_not be_a(Char)
+      typeof(x).should eq(Int32 | Char)
+      typeof(y).should eq(Int32)
+    end
   end
 
   describe "be_close" do
@@ -34,6 +48,13 @@ describe "expectations" do
     it { nil.should be_nil }
     it { "".should_not be_nil }
     it { 10.should_not be_nil }
+
+    it "restricts type on should_not" do
+      x = 1 || nil
+      y = x.should_not be_nil
+      typeof(x).should eq(Int32?)
+      typeof(y).should eq(Int32)
+    end
   end
 
   describe "be_falsey" do
--- a/src/spec/expectations.cr
+++ b/src/spec/expectations.cr
@@ -419,7 +419,7 @@ module Spec
   module ObjectExtensions
     # Validates an expectation and fails the example if it does not match.
     #
-    # This overload returns a value whose type is restricted to the expected type.. For example:
+    # This overload returns a value whose type is restricted to the expected type. For example:
     #
     # ```
     # x = 1 || 'a'
--- a/src/spec/expectations.cr
+++ b/src/spec/expectations.cr
@@ -419,10 +419,70 @@ module Spec
   module ObjectExtensions
     # Validates an expectation and fails the example if it does not match.
     #
+    # This overload will be called when doing:
+    #
+    # ```
+    # result = subject.should be_a(Type)
+    # # result will be a Type
+    # ```
+    #
+    # and make sure, at compile-time, that the returned object is of type `Type`.
+    #
+    # See `Spec::Expecations` for available expectations.
+    def should(expectation : BeAExpectation(T), file = __FILE__, line = __LINE__) forall T
+      if expectation.match self
+        self.is_a?(T) ? self : (raise "Bug: expected #{self} to be a #{T}")
+      else
+        fail(expectation.failure_message(self), file, line)
+      end
+    end
+
+    # Validates an expectation and fails the example if it does not match.
+    #
     # See `Spec::Expecations` for available expectations.
     def should(expectation, file = __FILE__, line = __LINE__)
       unless expectation.match self
         fail(expectation.failure_message(self), file, line)
+      end
+    end
+
+    # Validates an expectation and fails the example if it matches.
+    #
+    # This overload will be called when doing:
+    #
+    # ```
+    # result = subject.should_not be_a(Type)
+    # # result will not be a Type
+    # ```
+    #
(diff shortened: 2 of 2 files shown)


spec/std/spec/expectations_spec.cr
```diff
@@ -23,6 +23,18 @@ describe "expectations" do
     it { 100_000.should_not be_a(String) }
     it { 100_000.should be_a(Int32) }
     it { "Hello".should_not be_a(Int32) }
+
+    it "restricts type on should" do
+      x = 1 || 'a'
+      y = x.should be_a(Int32)
+      typeof(y).should eq(Int32)
```
RX14: Perhaps verify `typeof(x).should eq(Int32 | Char)` in case the type checker gets cleverer and the spec isnt updated.

asterite: Done!

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
   end
 
   module ObjectExtensions
+    # Validates an expectation and fails the example if it does not match.
+    #
+    # This overload will be called when doing:
+    #
+    # ```
+    # result = subject.should be_a(Type)
+    # # result will be a Type
```
straight-shoota: Describing this in code is IMO clearer:
```suggestion
    # typeof(result) # => Type
```

Even better to use a working example (i.e. copied from the spec):
```cr
x = 1 || 'a'
typeof(x) # => Int32 | Char
x = x.should be_a(Int32)
typeof(x) # => Int32
```
This is certainly easier to understand for the `should_not` matcher.

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
   end
 
   module ObjectExtensions
+    # Validates an expectation and fails the example if it does not match.
+    #
+    # This overload will be called when doing:
```
straight-shoota: I'd avoid placing an example block inside a sentence.
```suggestion
    # This overload accepts a `be_a(Type)` expectation and makes sure, at compile-time, that the returned object is of type `Type`.
```

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
...
+    #
+    # ```
+    # result = subject.should be_a(Type)
+    # # result will be a Type
+    # ```
+    #
+    # and make sure, at compile-time, that the returned object is of type `Type`.
+    #
+    # See `Spec::Expecations` for available expectations.
+    def should(expectation : BeAExpectation(T), file = __FILE__, line = __LINE__) forall T
```
straight-shoota: Maybe
```suggestion
    def should(expectation : BeAExpectation(T), file = __FILE__, line = __LINE__) : T forall T
```

asterite:
> This is technically missing should be_nil. I assume that's on purpose because it's hard to think of a use case for this?

Right, the resulting type will always be `nil` and afterwards there's not much you can do with that.

src/spec/expectations.cr
asterite: Great idea! I did this for all the overloads. Let me know what you think.

src/spec/expectations.cr
asterite: Done!

straight-shoota: I think you added it to the wrong method, the type restriction only works on `should`.

src/spec/expectations.cr
straight-shoota: Any reason for not mentioning the `be_a` method here? It would make the documentation more accessible for readers familiar with the DSL.

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
   end
 
   module ObjectExtensions
+    # Validates an expectation and fails the example if it does not match.
+    #
+    # This overload will return a value whose type is restricted. For example:
```
straight-shoota: Present tense, please?
```suggestion
    # This overload returns a value whose type is restricted. For example:
```

src/spec/expectations.cr
asterite: Sorry, what do you mean? I changed the code example and `be_a` is used there.

asterite:
Fixed once more. Let me know if this is good now.

src/spec/expectations.cr
straight-shoota: I meant mentioning it in the text.

But yeah, having it in the sample is probably fine.

src/spec/expectations.cr
straight-shoota: Other thing: What do you think about adding `[... whose type is restricted] to the expected type.`?

asterite: Where?

straight-shoota: Damn, I commented on the wrong line. I'll make a new suggestion.

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
   end
 
   module ObjectExtensions
+    # Validates an expectation and fails the example if it does not match.
+    #
+    # This overload returns a value whose type is restricted. For example:
```
straight-shoota:
```suggestion
    # This overload returns a value whose type is restricted to the expected type.. For example:
```

src/spec/expectations.cr
```diff
@@ -417,6 +417,26 @@ module Spec
   end
 
   module ObjectExtensions
+    # Validates an expectation and fails the example if it does not match.
+    #
+    # This overload returns a value whose type is restricted to the expected type.. For example:
```
straight-shoota: A single dot... =)
```suggestion
    # This overload returns a value whose type is restricted to the expected type. For example:
```
And could you add the phrase to the other methods as well?

Then, it should really be fine 🙈

Issue #8414, opened 2019-11-01
crystal init app crash


tankf33der:
Linux, 0.29 or 0.31.1, x86

```
# crystal init app m
Error opening file 'm/.gitignore' with mode 'w': No such file or directory (Errno)
Unhandled exception: Error opening file 'crystal' with mode 'r': No such file or directory (Errno)
Failed to raise an exception: END_OF_STACK
[0x55ffff71d4f6] ???
[0x55fffedf16c9] __crystal_raise +153
[0x55fffedf2395] ???
[0x55fffedfdbc3] ???
[0x55fffedf6dac] ???
[0x55fffedf6455] ???
[0x55ffff15486a] ???
[0x55ffff164189] ???
[0x55fffedf49d7] main +55
[0x7f1c1c541229] ???
```

evvo:
It seems to be a problem with a single letter.

For example if I use "mm", it works:
`
$ crystal init app mm
    create  mm/.gitignore
    create  mm/.editorconfig
    create  mm/LICENSE
    create  mm/README.md
    create  mm/.travis.yml
    create  mm/shard.yml
    create  mm/src/mm.cr
    create  mm/spec/spec_helper.cr
    create  mm/spec/mm_spec.cr
Initialized empty Git ...
`

jan-zajic:
It's bug in dir.cr method #dirname. 
This code:
```
full_path = "m/.gitignore"
puts "Full path: #{full_path} dir name: #{File.dirname(full_path)}"
```
prints:
```
Full path: m/.gitignore dir name: .
```
I will look at it.

asterite:
What Crystal version? Because I remember a similar bug and it should be already fixed in master (and maybe in 0.31.1, not sure)

asterite:
Oh, nevermind, I just tried it and it's broken in 0.31.1

bcardiff:
@asterite I guess you are thinking of #8119

bcardiff:
Closed by #8415

PR #8415, opened 2019-11-01
Fix bug in Path.dirname

jan-zajic:
This PR fixes issue #8414


straight-shoota:
Thanks for fixing this!

However I think the implementation can be a bit simplified. Instead of tracking with the `trailing_separator` variable, we can simply exit early before stripping:

```cr
if reader.pos == 0 && !separators.includes?(reader.current_char)
  return "."
end
```

This also enables us to clean up the following `if reader.pos == 0` branch because it doesn't need to return `"."` here anymore.

The specs should test `Path#dirname` in `path_spec.cr`.

`Path#parents` is also broken in a similar way: `Path.posix("m/.gitignore").parents` returns `[".", "m"]`. This could be fixed in a separate PR. But if you're interested...

Let me know if you need help or rather don't want to proceed with this, then I'm happy to take over.

jan-zajic:
@straight-shoota you right, but it breaks all that windows stuff, if just returns "." early ...

jan-zajic:
@straight-shoota so what about like this?

src/path.cr
```diff
@@ -190,24 +189,22 @@ struct Path
       reader.previous_char
     end
 
+    if reader.pos == 0 && !separators.includes?(reader.current_char)
+      if windows? && reader.has_next? && reader.peek_next_char == ':'
+        reader.next_char
```
straight-shoota: At this point we can actually check for `self.anchor` and return its value. The conditional in lines 210-212 can probably be  removed then.

jan-zajic: So it can be like this. But lines 210-212 cannot be removed, it breaks tests for C:\foo etc.
And IMHO your example for Path#parents is correct, i don't see error.

straight-shoota:
`Path.posix("m/.gitignore").parents` should eq `["m"]`, no `"."` in the parents.

jan-zajic:
@straight-shoota why? According to specs, that you wrote (0c31189d7e53d53a96440d01af9666ec429a83de), that is right. 
It's same as any relative path, for example test case `Users/foo/bar.cr` or method doc example `foo/bar/file.cr`.  
And it's right, parents should behave similar to parent (which call dirname). So if parent of `"m/.gitignore"` is m and parent of `"m"` is `"."`, 
parents of `"m/.gitignore"` is `[".", "m"]`, or?

jan-zajic:
But i added new test cases to `Path#parents` as you wished and found broken behavior of parents merhod for these cases:
"m//" (posix)
"m//" (windows)
"m//a/b" (posix)
"m//a/b" (windows)
"C:" (windows)
"C:/" (posix)
"C:folder" (windows)
"C:\\." (windows)
So i rewrite `Path#parents`, see my next commit.

straight-shoota:
You're totally right, sry. I misinterpreted the order of parents.

spec/std/path_spec.cr
```diff
@@ -164,6 +164,20 @@ describe Path do
     assert_paths("foo/./bar/.", [".", "foo", "foo/.", "foo/./bar"], &.parents)
     assert_paths("foo/bar/.", [".", "foo", "foo/bar"], &.parents)
     assert_paths("foo/bar/./.", [".", "foo", "foo/bar", "foo/bar/."], &.parents)
+    assert_paths("m/.gitignore", [".", "m"], &.parents)
+    assert_paths("m", ["."], &.parents)
+    assert_paths("m/", ["."], &.parents)
+    assert_paths("m//", ["."], &.parents)
+    assert_paths("m//a/b", [".", "m", "m//a"], &.parents)
```
straight-shoota: For posix paths this should be `[".", "m\\a"]`.

src/path.cr
```diff
@@ -253,24 +253,30 @@ struct Path
   # ```
   def each_parent(&block : Path ->)
     return if @name.empty? || @name == "."
-
+    reader = Char::Reader.new(@name)
+    last_was_separator = false
+    last_sep_pos = 0
     first = true
-    each_part_separator_index do |pos|
```
straight-shoota: I think we should keep using `each_part_separator_index`. It's just horribly broken and the issues need to be fixed *there*.
`Path#part` uses the same helper and suffers from the same issues.

It seems there are also no specs for this method. Here are a few:
```cr

  assert_paths_raw("m/.gitignore", ["m", ".gitignore"], &.parts)
  assert_paths_raw("m", ["m"], &.parts)
  assert_paths_raw("m/", ["m"], &.parts)
  assert_paths_raw("m//", ["m\\"], ["m"], &.parts)
  assert_paths_raw("m//a/b", ["m\\a", "b"], ["m", "a", "b"], &.parts)
  assert_paths_raw("/m", ["/", "m"], &.parts)
  assert_paths_raw("/m/", ["/", "m"], &.parts)
  assert_paths_raw("C:", ["C:"], &.parts)
  assert_paths_raw("C:/", ["C:/"], &.parts)
  assert_paths_raw("C:\\", ["C:\\"], &.parts)
  assert_paths_raw("C:folder", ["C:folder"], ["C:", "folder"], &.parts)
  assert_paths_raw("C:\\folder", ["C:\\folder"], ["C:\\", "folder"], &.parts)
  assert_paths_raw("C:\\\\folder", ["C:\\\\folder"], ["C:\\", "folder"], &.parts)
  assert_paths_raw("C:\\.", ["C:\\."], ["C:\\"], &.parts)
```

Alternatively, instead of duplicating the expectations for `parents` and `parts`, we could just combine them.
The following should always be true:
```cr
parts == parents.map(&.basename) + basename

# The reverse is only valid for normalized paths because `#parts` loses information
parents == parts[1..].reduce([Path.new(parts.first)]) do |paths, part|
  paths << paths.last.join(part)
end
```

jan-zajic: I don't think `each_part_separator_index` is broken, i just see it impractical for implementation of `each_parent`. I like to yield on other char than separator. I try with `each_part_separator_index` and made 3 very messy and confused implementations.. Maybe you have some elegant variant in mind?

jan-zajic: I also chose a different approach to implement `parts` which seems very simple and clear for me. See last commit.

spec/std/path_spec.cr
jan-zajic: I don't think so. Only if slashes in path were backslashes which aren't.

src/path.cr
straight-shoota: The very purpose of `each_part_separator_index` is to provide the basic behaviour that is similar to `#parts` and `#parents` (and `#anchor` uses it as well). So if we couldn't use it in `#parts`, it wouldn't make much sense to have such a helper method.

straight-shoota:
I've refactored `each_part_separator_index` entirely and also the related methods which makes `each_part` and `each_parent` now really trivial. I also refactored the UNC share detection to not rely on `each_part_separator_index` so that can directly yield the entire anchor.

I can't push to this branch, so I created a new one, you can see the changes here:
https://github.com/crystal-lang/crystal/compare/pull/8415/head...straight-shoota:fix/path_parents?expand=1

src/path.cr
```diff
@@ -516,20 +488,19 @@ struct Path
 
   private def each_part_separator_index(offset = 0)
     reader = Char::Reader.new(@name)
+    offset = 0
     reader.pos = start_pos = offset
```
jan-zajic: Even that offset variable and put zero here?

straight-shoota: Yes, that's just a leftover from a step in between

jan-zajic:
I think this is ready to merge now.

PR #8421, opened 2019-11-01
Fix GZip Reader and Writer to handle large data sizes

### LINKED ISSUE #8418: GZip IO Fails on Very Large gzip Files
I have a very large gzip file (the .COM TLD Zone file, it is 3.2GB compressed and about 14GB uncompressed). I am reading it using crystal's GZip library in the STDLIB. The issue I run into is that there appears to be a problem with an integer counter in the library overflowing due to the size of the data that has been decompressed. Here's the error (Crystal 0.31.1 on both Ubuntu 16.04 LTS and macOS):

```
Unhandled exception: Arithmetic overflow (OverflowError)
  from /usr/share/crystal/src/gzip/reader.cr:0:16 in 'unbuffered_read'
  from /usr/share/crystal/src/io/buffered.cr:214:12 in 'fill_buffer'
  from /usr/share/crystal/src/io/buffered.cr:102:7 in 'peek'
  from /usr/share/crystal/src/io.cr:692:18 in 'gets_peek'
  from /usr/share/crystal/src/io.cr:636:9 in 'gets'
  from /usr/share/crystal/src/io.cr:605:5 in 'gets'
  from /usr/share/crystal/src/io.cr:575:5 in 'gets'
  from /usr/share/crystal/src/io.cr:574:3 in 'gets'
  from fast_scanner.cr:25:9 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:86:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```

Here would be some example code to trigger the error, however you'd need a sufficiently large gzip file to do so:

```
require "gzip"
io = File.open("large_gzip_file.gz")
gzip_reader = GZip::Reader.new(io)
while line = gzip_reader.gets
  # Do something
end
gzip_reader.close
```

I have tried changing `@isize` in gzip/reader.cr to be an unsigned 64 bit int but that introduced another problem with comparisons later down the line. I am no expert on how gzip works, so I don't know what I can fix it any further. Let me know if you need more clarification.

Thanks!

straight-shoota:
In the gzip format isize is 4 bytes wide and contains the size of the original (uncompressed) input
data modulo 2^32. So sizes can be larger than UInt32::MAX but the addition must wrap, not overflow. See https://tools.ietf.org/html/rfc1952#page-8

Fixes #8418

### DIFF
--- a/spec/manual/gzip_large_file_spec.cr
+++ b/spec/manual/gzip_large_file_spec.cr
@@ -1,0 +1,41 @@
+require "gzip"
+require "spec"
+
+# This spec tests piping a file with a size of more than
+# UInt32::MAX bytes through GZip::Writer and GZ::Reader.
+# Zipping and unzipping so many bytes takes some time,
+# so this spec is quite slow.
+it "Gzip file larger than UInt32::MAX" do
+  read, write = IO.pipe
+  bytes_written = 0_i64
+  bytes_read = 0_i64
+
+  spawn do
+    slice = Slice.new(1024, 0_u8, read_only: true)
+
+    Gzip::Writer.open(write) do |writer|
+      target_bytes = UInt32::MAX.to_i64 + 1
+      while bytes_written < target_bytes
+        writer.write(slice)
+        bytes_written += slice.bytesize
+      end
+    end
+
+    write.close
+  end
+
+  Gzip::Reader.open(read) do |reader|
+    slice = Slice.new(1024, 0_u8)
+
+    while true
+      read_bytes = reader.read(slice)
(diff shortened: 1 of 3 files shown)


rvictory:
I can confirm that this code (manually applied in my environment) fixes my issue (https://github.com/crystal-lang/crystal/issues/8418). Amazing work guys, this just confirms that this language is the one I want!

PR #8426, opened 2019-11-02
Cleanup Digest and OpenSSL::Digest

didactic-drunk:
* Use a Digest::Base as a common base class.
* Remove duplicate methods.
* Mark various methods private.
* Add specs to test #reset and internal state.

Digest::Base
* Add abstract methods solidifying the interface.

### DIFF
--- a/spec/std/digest/sha1_spec.cr
+++ b/spec/std/digest/sha1_spec.cr
@@ -10,28 +10,26 @@ describe Digest::SHA1 do
     {"a", "86f7e437faa5a7fce15d1ddcb9eaeaea377667b8", "hvfkN/qlp/zhXR3cuerq6jd2Z7g="},
     {"0123456701234567012345670123456701234567012345670123456701234567", "e0c094e867ef46c350ef54a7f59dd60bed92ae83", "4MCU6GfvRsNQ71Sn9Z3WC+2SroM="},
     {"fooø", "dcf4a1e3542b1a40a4ac2a3f7c92ffdb2d19812f", "3PSh41QrGkCkrCo/fJL/2y0ZgS8="},
-  ].each do |(string, hexdigest, base64digest)|
+  ].each do |(string, hexstring, base64digest)|
     it "does digest for #{string.inspect}" do
       bytes = Digest::SHA1.digest(string)
-      bytes.to_slice.hexstring.should eq(hexdigest)
+      bytes.hexstring.should eq(hexstring)
     end
 
     it "resets" do
       digest = Digest::SHA1.new
       digest.update string
-      digest.hexdigest.should eq(hexdigest)
-      # Check internal state isn't modified by first hexdigest call.
-      digest.hexdigest.should eq(hexdigest)
+      digest.final.hexstring.should eq(hexstring)
 
       digest.reset
       digest.update string
-      digest.hexdigest.should eq(hexdigest)
+      digest.final.hexstring.should eq(hexstring)
     end
 
-    it "resets" do
+    it "can't call #final more than once" do
       digest = Digest::SHA1.new
       digest.final
-      expect_raises(Digest::ApiMisuse) do
+      expect_raises(Digest::FinalizedError) do
         digest.final
       end
     end
@@ -43,11 +41,11 @@ describe Digest::SHA1 do
         end
       end
 
-      bytes.to_slice.hexstring.should eq(hexdigest)
+      bytes.hexstring.should eq(hexstring)
     end
 
-    it "does hexdigest for #{string.inspect}" do
-      Digest::SHA1.hexdigest(string).should eq(hexdigest)
+    it "does final.hexstring for #{string.inspect}" do
+      Digest::SHA1.hexdigest(string).should eq(hexstring)
     end
 
     it "does base64digest for #{string.inspect}" do
--- a/spec/std/digest/sha1_spec.cr
+++ b/spec/std/digest/sha1_spec.cr
@@ -28,6 +28,14 @@ describe Digest::SHA1 do
       digest.hexdigest.should eq(hexdigest)
     end
 
+    it "resets" do
+      digest = Digest::SHA1.new
+      digest.final
+      expect_raises(Digest::ApiMisuse) do
+        digest.final
+      end
+    end
+
     it "does digest for #{string.inspect} in a block" do
       bytes = Digest::SHA1.digest do |ctx|
         string.each_char do |chr|
--- a/spec/std/digest/sha1_spec.cr
+++ b/spec/std/digest/sha1_spec.cr
@@ -44,7 +44,7 @@ describe Digest::SHA1 do
       bytes.hexstring.should eq(hexstring)
     end
 
-    it "does final.hexstring for #{string.inspect}" do
+    it "does hexdigest for #{string.inspect}" do
       Digest::SHA1.hexdigest(string).should eq(hexstring)
     end
 
--- a/spec/std/digest/sha1_spec.cr
+++ b/spec/std/digest/sha1_spec.cr
@@ -16,6 +16,18 @@ describe Digest::SHA1 do
       bytes.to_slice.hexstring.should eq(hexdigest)
     end
 
+    it "resets" do
+      digest = Digest::SHA1.new
+      digest.update string
+      digest.hexdigest.should eq(hexdigest)
+      # Check internal state isn't modified by first hexdigest call.
+      digest.hexdigest.should eq(hexdigest)
+
+      digest.reset
+      digest.update string
+      digest.hexdigest.should eq(hexdigest)
+    end
+
     it "does digest for #{string.inspect} in a block" do
       bytes = Digest::SHA1.digest do |ctx|
         string.each_char do |chr|
@@ -33,5 +45,9 @@ describe Digest::SHA1 do
     it "does base64digest for #{string.inspect}" do
       Digest::SHA1.base64digest(string).should eq(base64digest)
     end
+  end
+
+  it "returns the digest_size" do
+    Digest::SHA1.new.digest_size.should eq(20)
   end
 end
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -24,7 +24,6 @@ abstract class Digest::Base
     context = new
     yield context
     context.final
-    context.result
   end
 
   # Returns the hexadecimal representation of the hash of *data*.
@@ -89,4 +88,28 @@ abstract class Digest::Base
 
     Base64.strict_encode(hashsum)
   end
+
+  def update(data) : self
+    update data.to_slice
+  end
+
+  def final : Bytes
+    dst = Bytes.new digest_size
+    final dst
+  end
+
+  # Dups and finishes the digest.
+  def digest : Bytes
+    dup.final
+  end
+
+  # Returns a hexadecimal-encoded digest.
+  def hexdigest : String
+    digest.hexstring
+  end
+
+  abstract def update(data : Bytes) : self
+  abstract def final(dst : Bytes) : Bytes
+  abstract def reset : self
+  abstract def digest_size : Int32
 end
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -102,7 +102,9 @@ abstract class Digest::Base
     update data.to_slice
   end
 
-  # Returns the final digest output.  Only call once.
+  # Returns the final digest output.
+  #
+  # This method can only be called once and raises `FinalizedError` on subsequent calls.
   #
   # NOTE: `.dup.final` call may be used to get an intermediate hash value.
   def final : Bytes
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -24,7 +24,7 @@ abstract class Digest::Base
     # end
     # digest.to_slice.hexstring # => "acbd18db4cc2f85cedef654fccc4a4d8"
     # ```
-    def self.digest : Bytes
+    def self.digest(& : Digest::Base -> _) : Bytes
       context = new
       yield context
       context.final
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -12,8 +12,8 @@ abstract class Digest::Base
       end
     end
 
-    # Yields an instance of `self` which can receive calls to `#update(data : String | Bytes)` and returns the finalized digest.
-    # method available. Returns the resulting digest afterwards.
+    # Yields an instance of `self` which can receive calls to `#update(data : String | Bytes)`
+    # and returns the finalized digest afterwards.
     #
     # ```
     # require "digest/md5"
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -1,7 +1,7 @@
 require "base64"
 
 module Digest
-  class DigestFinalizedError < Exception
+  class FinalizedError < Exception
   end
 end
 
@@ -121,7 +121,7 @@ abstract class Digest::Base
   end
 
   protected def check_finished : Nil
-    raise ApiMisuse.new("finish already called") if @finished
+    raise FinalizedError.new("finish already called") if @finished
   end
 
   protected def set_finished : Nil
@@ -135,9 +135,9 @@ abstract class Digest::Base
     self
   end
 
-  # When creating new digest class call #check_finished before mutating.
+  # When creating a new digest class call #check_finished before mutating.
   abstract def update(data : Bytes) : self
-  # When creating new digest class call #set_finished before mutating.
+  # When creating a new digest class call #set_finished before mutating.
   abstract def final(dst : Bytes) : Bytes
   abstract def digest_size : Int32
 end
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -1,7 +1,7 @@
 require "base64"
 
 module Digest
-  class ApiMisuse < Exception
+  class DigestFinalizedError < Exception
   end
 end
 
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -103,6 +103,7 @@ abstract class Digest::Base
   end
 
   # Returns the final digest output.  Only call once.
+  # `.dup.final` may be used to get an intermediate hash value.
   def final : Bytes
     dst = Bytes.new digest_size
     final dst
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -76,7 +76,9 @@ abstract class Digest::Base
       base64digest &.update(data)
     end
 
-    # Returns the base64-encoded hash of *data*.
+    # Yields a context object with an `#update(data : String | Bytes)`
+    # method available. Returns the resulting digest in base64 representation
+    # afterwards.
     #
     # ```
     # require "digest/sha1"
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -1,92 +1,94 @@
 require "base64"
 
 abstract class Digest::Base
-  # Returns the hash of *data*. *data* must respond to `#to_slice`.
-  def self.digest(data)
-    digest do |ctx|
-      ctx.update(data.to_slice)
-    end
-  end
-
-  # Yields a context object with an `#update(data : String | Bytes)`
-  # method available. Returns the resulting digest afterwards.
-  #
-  # ```
-  # require "digest/md5"
-  #
-  # digest = Digest::MD5.digest do |ctx|
-  #   ctx.update "f"
-  #   ctx.update "oo"
-  # end
-  # digest.to_slice.hexstring # => "acbd18db4cc2f85cedef654fccc4a4d8"
-  # ```
-  def self.digest
-    context = new
-    yield context
-    context.final
-  end
-
-  # Returns the hexadecimal representation of the hash of *data*.
-  #
-  # ```
-  # require "digest/md5"
-  #
-  # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
-  # ```
-  def self.hexdigest(data) : String
-    hexdigest &.update(data)
-  end
-
-  # Yields a context object with an `#update(data : String | Bytes)`
-  # method available. Returns the resulting digest in hexadecimal representation
-  # afterwards.
-  #
-  # ```
-  # require "digest/md5"
-  #
-  # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
-  # Digest::MD5.hexdigest do |ctx|
-  #   ctx.update "f"
-  #   ctx.update "oo"
-  # end
-  # # => "acbd18db4cc2f85cedef654fccc4a4d8"
-  # ```
-  def self.hexdigest : String
-    hashsum = digest do |ctx|
-      yield ctx
+  macro inherited
+    # Returns the hash of *data*. *data* must respond to `#to_slice`.
+    def self.digest(data)
+      digest do |ctx|
+        ctx.update(data.to_slice)
+      end
     end
 
-    hashsum.to_slice.hexstring
-  end
-
-  # Returns the base64-encoded hash of *data*.
-  #
-  # ```
-  # require "digest/sha1"
-  #
-  # Digest::SHA1.base64digest("foo") # => "C+7Hteo/D9vJXQ3UfzxbwnXaijM="
-  # ```
-  def self.base64digest(data) : String
-    base64digest &.update(data)
-  end
-
-  # Returns the base64-encoded hash of *data*.
-  #
-  # ```
-  # require "digest/sha1"
-  #
-  # Digest::SHA1.base64digest do |ctx|
-  #   ctx.update "f"
-  #   ctx.update "oo"
-  # end
-  # # => "C+7Hteo/D9vJXQ3UfzxbwnXaijM="
-  # ```
-  def self.base64digest : String
-    hashsum = digest do |ctx|
-      yield ctx
+    # Yields a context object with an `#update(data : String | Bytes)`
+    # method available. Returns the resulting digest afterwards.
+    #
+    # ```
+    # require "digest/md5"
+    #
+    # digest = Digest::MD5.digest do |ctx|
+    #   ctx.update "f"
+    #   ctx.update "oo"
+    # end
+    # digest.to_slice.hexstring # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # ```
+    def self.digest
+      context = new
+      yield context
+      context.final
     end
 
-    Base64.strict_encode(hashsum)
+    # Returns the hexadecimal representation of the hash of *data*.
+    #
+    # ```
+    # require "digest/md5"
+    #
+    # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # ```
+    def self.hexdigest(data) : String
+      hexdigest &.update(data)
+    end
+
+    # Yields a context object with an `#update(data : String | Bytes)`
+    # method available. Returns the resulting digest in hexadecimal representation
+    # afterwards.
+    #
+    # ```
+    # require "digest/md5"
+    #
+    # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # Digest::MD5.hexdigest do |ctx|
+    #   ctx.update "f"
+    #   ctx.update "oo"
+    # end
+    # # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # ```
+    def self.hexdigest : String
+      hashsum = digest do |ctx|
+        yield ctx
+      end
+
+      hashsum.to_slice.hexstring
+    end
+
+    # Returns the base64-encoded hash of *data*.
+    #
+    # ```
+    # require "digest/sha1"
+    #
+    # Digest::SHA1.base64digest("foo") # => "C+7Hteo/D9vJXQ3UfzxbwnXaijM="
+    # ```
+    def self.base64digest(data) : String
+      base64digest &.update(data)
+    end
+
+    # Returns the base64-encoded hash of *data*.
+    #
+    # ```
+    # require "digest/sha1"
+    #
+    # Digest::SHA1.base64digest do |ctx|
+    #   ctx.update "f"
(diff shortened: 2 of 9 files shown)


didactic-drunk:
A single abstract class allows [3rd party libraries](https://github.com/didactic-drunk/sodium.cr) to to be interchangably with the included crystal digest classes in type restrictions and instance variables.

didactic-drunk:
Duplicating the `Digest` object when calling `digest` or `hexdigest` has high overhead.  The default should be left alone so users don't shoot themselves in the foot.  A `final` method which removes the `dup` may be used to provide ~4x better performance depending on the `Digest` and data size.

Some use cases for a public `#final` method:
* Iterative hashing.
* Hashing small messages.

Benchmarks:
* `new` Creates a new object each iteration.
* `reset` Creates a single object and calls `#reset` each iteration.
* `final` Also resets and uses `#final` instead of `#digest`.
* `final(buf)` Same as above with with a preallocated buffer.
```
   md5 final(buf) 16   7.05M (141.86ns) (± 3.38%)   0.0B/op         fastest
        md5 final 16   6.28M (159.34ns) (± 3.29%)  32.0B/op    1.12× slower
        md5 reset 16   1.69M (591.69ns) (± 2.49%)   128B/op    4.17× slower
          md5 new 16 953.89k (  1.05µs) (± 1.76%)   224B/op    7.39× slower
```


`#final` (`#finish`) is already in use within `sodium.cr` for `Blake2b`.  `Blake2b` performance is comparable to `MD5` with better security properties.

watzon:
Awesome, this is sorely needed as is some additional documentation

didactic-drunk:
> Awesome, this is sorely needed as is some additional documentation

None of my previous PR's made it through without modification.  I'll consider documentation when I know this will be accepted.  Right now there's no feedback.  I assume everyone is busy or the domain expertise is lacking.

This PR is based on tuning and benchmarking [sodium.cr](https://github.com/didactic-drunk/sodium.cr) in a real application hashing small messages where speed is critical.

If this PR is accepted`Blake2b` wlll be drop in compatible with `Digest::Base` in sodium's next release.

src/digest/base.cr
```diff
@@ -89,4 +88,28 @@ abstract class Digest::Base
...
+  end
+
+  def final : Bytes
+    dst = Bytes.new digest_size
+    final dst
+  end
+
+  # Dups and finishes the digest.
+  def digest : Bytes
+    dup.final
```
straight-shoota: It might seem like a legitimate idea to decouple finalization from current state for re-usability. I'd like to see a realistic use case for that, though. And I don't think this should be the default behaviour.
The `#digest` looks like the regular means to (simply) get the digest result and shouldn't come with a heat allocation as side effect. IMO such a method should be more explicitly named, for example `intermediary_digest`.

src/digest/base.cr
```diff
@@ -89,4 +88,28 @@ abstract class Digest::Base
...
+
+  # Returns a hexadecimal-encoded digest.
+  def hexdigest : String
+    digest.hexstring
+  end
+
+  abstract def update(data : Bytes) : self
+  abstract def final(dst : Bytes) : Bytes
+  abstract def reset : self
+  abstract def digest_size : Int32
```
straight-shoota: Is this really useful as instance method or should it rather be a constant (or method) on the type.

src/digest/sha1.cr
```diff
@@ -10,23 +10,33 @@ class Digest::SHA1 < Digest::Base
...
+  @message_block_index = 0
+  @message_block = StaticArray(UInt8, 64).new(0_u8) # uninitialized UInt8[64]
+
   def initialize
-    @intermediate_hash = uninitialized UInt32[5]
+    reset
+  end
+
+  def reset : self
+    @message_block.each_index do |i|
```
straight-shoota: Assigning a new instance might be preferable?

src/digest/base.cr
didactic-drunk: Can't have it that way when considering OpenSSL::Digest.  The `digest_size` is variable based on the passed in args.  Blake2 allows for variable length digests based on the initialization parameters.  So do many modern digests like Keccak.

src/digest/sha1.cr
didactic-drunk: No.  See the benchmark results above.  Iterative hashing is one example where performance is critical.  Reducing the speed reduces the maximum security parameters.

src/digest/base.cr
straight-shoota: Okay. :+1:

src/digest/sha1.cr
straight-shoota: I mean `@message_block = StaticArray(UInt8, 64).new(0_u8)`, not initializing a new `Digest` instance. `reset` is fine.

src/digest/base.cr
didactic-drunk: Well, it's a departure from ruby.  In ruby `.to_s` calls `.digest`.  I assume the dupping behavior is so `.to_s` will work regardless of when it's called.  If `digest` mutates state either `to_s` prints the wrong result or worse `to_s` changes the resulting real hash value and breaks something.

In Crystal `OpenSSL::Digest` has a similar `.to_s` method.  So either showing the `.hexdigest` in `.to_s` has to go or the behavior needs to stay the same.

I could track state and disallow calling `final` more than once, leaving it to the programmer to call `dup.final` if they want to hash more than once, although I don't see why (other than inspection).  Printing the hash in `OpenSSL::Digest` obviously has to go with a change like this.

src/digest/sha1.cr
didactic-drunk: Oh.

Is there a `memset`?

didactic-drunk:
> I'm not sure about the change to `OpenSSL::Digest`. The entire `OpenSSL` namespace needs to be refactored anyway at some point, so it might be better to leave it as is for now.

That defeats one of the main purposes of this PR.  Instance vars need a type.  I'd like to use `Blake2b`, `Digest` and `OpenSSL::Digest` interchangably which means having a base type they all inherit from with a common interface.

src/digest/sha1.cr
didactic-drunk: I removed it.  It's not used.

src/digest/base.cr
didactic-drunk: Deprecate `#digest` and `#hexdigest` and replace with `#final`?

@watzon can comment if he likes regarding `#hexdigest`, `#hexfinal` or whatever it should be called.  `hexfinal(dst)` could avoid one `String(digest_length)` allocation.  I would use it but I'm not sure about other people.  

There was someone on gitter a few days ago looking to do as few memory allocations as possible when reading from `STDIN`.  Maybe it would be used if provided.

straight-shoota: Removing `#digest` in favour of `#final` seems fine. Theoretically, we could consider renaming the behaviour to `#digest`, but `#final` is more expressive.

`hexfinal(dst)` would be equivalent to `final(dst); dst.hexstring`. I don't think we need a short cut for that.

src/digest/base.cr
```diff
@@ -1,92 +1,143 @@
 require "base64"
 
+module Digest
+  class ApiMisuse < Exception
```
Sija:
```suggestion
  class DigestFinalizedError < Exception
```

didactic-drunk: Somewhere I read having lots of exception classes was bad for compiler performance.

This may be the first exception but it's intended as a catch all for further misuse cases.  Would you still like it changed?

src/digest/base.cr
didactic-drunk: I haven't run the benchmarks but could this be worth optimizing?

Creates two heap objects.
```
final.hexstring
```

Creates one heap object.
```
def hexfinal
  dst = StaticArray(...)
  String.new dst
end
```

Creates no heap objects.
```
def hexfinal(dst)
  tmp = StaticArray(...)
  final tmp
  some_hex_func tmp dst
end
```

src/digest/base.cr
Sija: > This may be the first exception but it's intended as a catch all for further misuse cases.

That's future-proofing, I'd better stick to the use-cases at hand, and solve the problems as they arise - there would be no problem with adding a common ancestor l8r on if deemed necessary - named probably more like `Digest::Error`.

> Would you still like it changed?

yes plz.

didactic-drunk:
Review?

src/digest/base.cr
```diff
@@ -1,92 +1,143 @@
...
-  # Returns the hexadecimal representation of the hash of *data*.
-  #
-  # ```
-  # require "digest/md5"
-  #
-  # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
-  # ```
-  def self.hexdigest(data) : String
-    hexdigest &.update(data)
+  # Returns the final digest output.  Only call once.
```
RX14: Please mention that you can use `.dup.final` to get an intermediate hash value.

src/digest/base.cr
```diff
@@ -1,92 +1,144 @@
...
+  # `.dup.final` may be used to get an intermediate hash value.
```
Sija:
```suggestion
  #
  # NOTE: `.dup.final` call may be used to get an intermediate hash value.
```

src/digest/base.cr
```diff
@@ -1,92 +1,155 @@
 require "base64"
 
+module Digest
+  class Digest::FinalizedError < Exception
+  end
+end
```
Sija: You should omit outer `module Digest ... end`.

src/digest/base.cr
```diff
@@ -1,92 +1,153 @@
...
+  @[Deprecated("Use `final.hexstring` instead.")]
+  def hexdigest : String
+    digest.hexstring
   end
+
+  private def check_finished : Nil
+    raise FinalizedError.new("finish already called") if @finished
+  end
+
+  abstract def update_impl(data : Bytes) : Nil
```
straight-shoota: I'd make the _impl methods private. They're not part of the public API and are only relevant for implementing classes. That would mean they don't show up in the API docs, but I'd rather add implementation instructions in the class documentation.

And `reset_impl` needs an abstract def.

didactic-drunk: I'm just copying what's in the crystal repo.

Other inherited classes don't document the abstract methods in the class documentation but keep it with the abstract method.

```
# Flushes the wrapped `IO`.
abstract def unbuffered_flush
```

They're also not `private`.  Maybe so they do show up in the base class documentation avoiding duplication of documenting the definitions, types, and return values?

Sija: > They're also not `private`. Maybe so they do show up in the base class documentation avoiding duplication of documenting the definitions, types, and return values?

I believe it's because you can't use visibility modifiers within the `abstract def` declaration - they have to be on the implementation side only.

didactic-drunk: `abstract def unbuffered_read` shows in `IO::Buffered` documentatation.

`src/http/server/response.cr:198:      private def unbuffered_read(slice : Bytes)` doesn't.

Is that not the correct way to do it?  It seems logical to me.  

If the `abstract def`'s are private I have to duplicate the definitions for every `_impl` method in the class documentation.  Which may get out of sync over time.

If that's not the correct way why aren't all the other method definitions changed?

straight-shoota: No, you can define `private abstract` methods.

The public methods are relevant for all users. But implementation methods are only relevant when implementing the interface, they're not supposed to be called on their own. So they should not be listed in the public interface. 

The purpose can differ depending on the API in question. But some implementations of `IO::Buffered` actually declare the methods implementations as `private`. They're simply not relevant for using it. Maybe we should change that for `IO::Buffered` abstract methods as well. But that's a separate issue.

straight-shoota:
> If the abstract def's are private I have to duplicate the definitions for every _impl method in the class documentation. Which may get out of sync over time.

Yes, that's some unfortunate duplication. But it's IMO still worth it, because it separates user documentation from implementer documentation. And the sync effort should be minimal because these interfaces should be relatively stable. Otherwise it would break implementations all the time.

src/digest/base.cr
asterite: I think that if an abstract method is supposed to be implemented by subclasses, and subclasses can be defined outside of the std, then they should be public, or maybe protected. But if they are protected or private they won't show up in docs, so maybe public is fine? In any case I don't think anyone will invoke those methods.

Or put another way: maybe it doesn't matter too much?

ysbaddaden:
I believe the current state is fine. The `abstract def` are documented in Digest::Base, where it matters, because develops must implementing them, then each subclass marks them as private because they're an implementation detail.

src/digest/base.cr
straight-shoota: I think it matters when you have a bunch of duplicate methods in the API docs where only halve of them are meant to be called from user code.

src/digest/md5.cr
```diff
@@ -7,23 +7,29 @@ require "./base"
...
     @buf[3] = 0x10325476_u32
-    @in = StaticArray(UInt8, 64).new(0_u8)
-    @digest = uninitialized UInt8[16]
   end
 
-  def update(data)
-    slice = data.to_slice
-    update(slice.to_unsafe, slice.bytesize.to_u32)
+  private def update_impl(data : Bytes) : Nil
+    check_finished
```
straight-shoota:
```suggestion
```

src/digest/sha1.cr
```diff
@@ -42,9 +48,20 @@ class Digest::SHA1 < Digest::Base
         process_message_block
       end
     end
+    self
```
Sija:
```suggestion
```

jwoertink:
@bcardiff Since this has 2 core-team approvals, maybe it can be tagged with 0.33 or merged in?

spec/std/digest/sha1_spec.cr
```diff
@@ -23,15 +41,19 @@ describe Digest::SHA1 do
         end
       end
 
-      bytes.to_slice.hexstring.should eq(hexdigest)
+      bytes.hexstring.should eq(hexstring)
     end
 
-    it "does hexdigest for #{string.inspect}" do
-      Digest::SHA1.hexdigest(string).should eq(hexdigest)
+    it "does final.hexstring for #{string.inspect}" do
```
straight-shoota: This spec test `Digest::SHA1.hexdigest` and `final.hexstring` only indirectly. IMO this should either keep the old description or change the implementation to call `final.hexstring` explicitly. I'd prefer the former.

src/digest/base.cr
```diff
@@ -1,92 +1,158 @@
...
+    # ```
+    # require "digest/md5"
+    #
+    # digest = Digest::MD5.digest do |ctx|
+    #   ctx.update "f"
+    #   ctx.update "oo"
+    # end
+    # digest.to_slice.hexstring # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # ```
+    def self.digest : Bytes
```
straight-shoota:
```suggestion
    def self.digest(& : Digest::Base -> _) : Bytes
```

src/digest/base.cr
```diff
@@ -1,92 +1,158 @@
...
+    # require "digest/md5"
+    #
+    # Digest::MD5.hexdigest("foo") # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # Digest::MD5.hexdigest do |ctx|
+    #   ctx.update "f"
+    #   ctx.update "oo"
+    # end
+    # # => "acbd18db4cc2f85cedef654fccc4a4d8"
+    # ```
+    def self.hexdigest : String
```
straight-shoota:
```suggestion
    def self.hexdigest(& : Digest::Base -> _) : String
```

src/digest/base.cr
```diff
@@ -1,92 +1,158 @@
...
+    # ```
+    # require "digest/sha1"
+    #
+    # Digest::SHA1.base64digest do |ctx|
+    #   ctx.update "f"
+    #   ctx.update "oo"
+    # end
+    # # => "C+7Hteo/D9vJXQ3UfzxbwnXaijM="
+    # ```
+    def self.base64digest : String
```
straight-shoota:
```suggestion
    def self.base64digest(& : Digest::Base -> _) : String
```

Sija:
@didactic-drunk ready for re-review/merge?

didactic-drunk:
> @didactic-drunk ready for re-review/merge?

Yes

didactic-drunk:
Based on [Approval and merging process](https://forum.crystal-lang.org/t/approval-and-merging-process/2020) this PR is stuck.  15 days waiting for a milestone so far (this time).  The entire month of Dec waiting for anything even with 2 approvals.

I haven't contacted core team members privately this year, so am I not pestering enough?

straight-shoota:
This is just waiting for reviews on the latest state. It's quite a big PR with some breaking changes that need to be properly reviewed again.
It's fine to bump this though to get some attention after two weeks.

didactic-drunk:
There's only ~100 new non-spec LOC.  The biggest change is due to the formatter shifting things around in `digest/base.cr` after adding `macro inherited` which you suggested.  Yes it's breaking, but it's not very big if you skip over the reformatted class methods.

src/digest/base.cr
```diff
@@ -1,92 +1,158 @@
...
+    # require "digest/sha1"
+    #
+    # Digest::SHA1.base64digest("foo") # => "C+7Hteo/D9vJXQ3UfzxbwnXaijM="
+    # ```
+    def self.base64digest(data) : String
+      base64digest &.update(data)
+    end
+
+    # Yields a context object with an `#update(data : String | Bytes)`
+    # method available. Returns the resulting digest in base64 representation
```
RX14, line 78: This is inconsistent with `self.digest` above: either mention a context object or `self`, but not both wordings.

didactic-drunk: I didn't write that comment.  The format tool reindented it.  It's already committed in crystal.

Is fixing other peoples comments on methods I haven't changed the function of related to this PR?  When I wrote the title I meant it to be a minimal cleanup, not taking on every task.  

Am I supposed to fix everything in `Digest` I didn't write before I get my PR through or is this PR already large/complicated and takes a long time to review? 

This seems like I'm being told conflicting information about the review process.    What will get my PR through faster?

straight-shoota: @didactic-drunk Please stay calm. There's no reason to assume that @RX14 was aware that this doc comment is not related to the change. It looks like it does if you don't ignore whitespace changes.
I understand you might be frustrated about not getting merged already, but your reaction is out of line. A simple note would've been enough.

Sija: > There's no reason to assume that @RX14 was aware that this doc comment is not related to the change. It looks like it does if you don't ignore whitespace changes.

@straight-shoota If someone is undertaking a task of doing a code-review, yes, I'd say that should be **the reason** to assume he understands the changes he's looking at.

> I understand you might be frustrated about not getting merged already, but your reaction is out of line. A simple note would've been enough.

IMO it's a perfectly understandable reaction (frustration as you've noted) to the situation going on with this - and other PRs - by @didactic-drunk. People have emotions, which are repeatedly put to the tests by the actions of the Core Team members - like lack of communication, stalling PRs and others. I don't think this reaction is out of line, more like insinuating that after ~half of year of waiting, someones irritation might be misplaced. Above all, all he gets from you, is the _understanding_ of the faulty, or lack of the processes on your side that leads to situation like this - perhaps it's a good time to implement some bot to remind you about PRs, or take other corrective actions instead of _sympathizing_...

src/digest/base.cr
```diff
@@ -1,92 +1,158 @@
...
+  macro inherited
+    # Returns the hash of *data*. *data* must respond to `#to_slice`.
+    def self.digest(data)
+      digest do |ctx|
+        ctx.update(data.to_slice)
+      end
+    end
+
+    # Yields an instance of `self` which can receive calls to `#update(data : String | Bytes)` and returns the finalized digest.
+    # method available. Returns the resulting digest afterwards.
```
Sija, line 16:
```suggestion
    # Yields an instance of `self` which can receive calls to `#update(data : String | Bytes)`
    # and returns the finalized digest. Returns the resulting digest afterwards.
```

src/digest/base.cr
straight-shoota: I'm sorry that everyone has limited resources. And it seems there are not many people interested in reviewing this. I don't know why. With that I don't just mean core team members. There are no approving reviews of non-core members either which would certainly help to push this.

But I also don't think the tone of @didactic-drunk's previous comment is appropriate. And sadly I'm sure that such an attutide doesn't help to encourage anyone to do reviews.

RX14: You changed the doc comment above:



I'm pointing out that if you change one doc comment, you need to change them all, or change none of them.

Sija: > I'm sorry that everyone has limited resources. And it seems there are not many people interested in reviewing this. I don't know why. With that I don't just mean core team members.

No need to be sorry, it ain't about limited resources, but the communication - or lack thereof - which is a major hurdle for everyone involved. It would be way easier if we could have any estimates in re: to when can we expect any reaction, who's responsible for doing so, what can we do to ease the process as much as possible (thinking of you), etc. - and all of that we must find out by ourselves, since there's no formal review process, no automation, anything that could help to clarify this matter.

> There are no approving reviews of non-core members either which would certainly help to push this.

I haven't seen neither any previous indication that approvals from non-core members count towards anything, nor a precedent that could cue me in on that, so either you just tried to manipulate this conversation, pointing at things that could have been done but weren't, or you've forgot to tell the community members it actually has any value...

> But I also don't think the tone of @didactic-drunk's previous comment is appropriate. And sadly I'm sure that such an attutide doesn't help to encourage anyone to do reviews.

What **is** and what **is not** appropriate is a very subjective matter. I for one, had such feeling many times when confronted by @RX14 simply rude behaviour - way off and below anything _appropriate_, especially for someone who IMO should steward this community, instead of treating it as its own playground. I wouldn't call stalling a PR an _appropriate_ either.

didactic-drunk: Regarding the documentation:
1. I didn't change the documentation.  @straight-shoota did as an addition I assumed was necessary to get the PR approved.  Proof provided below.

What I was trying to say:
1.  People are requesting (documentation or other cleanup) changes which creates more requests for changes leaving this PR in perpetual limbo.
2. I wanted a quick code cleanup of the parts I'm interested in.  At what point can I say "not it" and let someone else handle a documentation cleanup without my involvement?
3. Without sarcasm or frustration: What will get this PR through faster?

Regarding tone:
1. Tone is hard to get.  I admit being annoyed initially, but tried to edit it out the emotional cruft and get down to a single purpose.  What gets the PR through faster?
2. The purpose of the last message was to provide minimal context highlighting the problems I see.  Others have mentioned I both write too much or too little.  Apparently my attempt to hit the middle came off wrong.
3. If there's something out of line in my post I don't see it.  I was calm while editing and removed what I saw as emotional.  The only parts left are what I think are relevant context.  Maybe it's too terse?  IDK.

```
Author: didactic-drunk <1479616+didactic-drunk@users.noreply.github.com>
Date:   Mon Nov 25 15:29:48 2019 -0800

    Update src/digest/base.cr

    Co-Authored-By: Johannes Müller <johannes.mueller@smj-fulda.org>

diff --git a/src/digest/base.cr b/src/digest/base.cr
index b321d47bc..bf573a687 100644
--- a/src/digest/base.cr
+++ b/src/digest/base.cr
@@ -14,7 +14,7 @@ abstract class Digest::Base
       end
     end

-    # Yields a context object with an `#update(data : String | Bytes)`
+    # Yields an instance of `self` which can receive calls to `#update(data : String | Bytes)` and returns the finalized digest.
     # method available. Returns the resulting digest afterwards.
     #
```

didactic-drunk: If it's stlll not clear:
1. I made minimal code changes.
2. @straight-shoota wanted documentation/various cleanup changes (Mostly last year).
3. @RX14 is correcting @straight-shootas's (as I see it unrelated to my changes) changes.
4. Feedback loop noticed.  Why can't they fix the documentation in a separate PR?

^ Not bitchy.  Factual tone.

straight-shoota: > It would be way easier if we could have any estimates in re: to when can we expect any reaction

That would be really great, but I have no idea how it could be possible.

> who's responsible for doing so

We have no formalized code ownership or something similar, so it's just the whole core team in total responsible. I.e. nobody in particular. 

> what can we do to ease the process as much as possible (thinking of you)

Reviews are easy if they consist of only a few self-contained commits and don't change much during the PR discussion. That means the specification of the API should be finalized before the PR even starts and the PR itself should only be about implementation details.
So this PR is actually a very good example of a very hard review process. Over time it has collected a convoluted commit history and lots of comments. @didactic-drunk, please take no offense! It's just an observation from my point of view. Maybe it would have been better to put  the PR on hold right at the beginning and start a separate design discussion first.

> I haven't seen neither any previous indication that approvals from non-core members count towards anything

Honestly, I didn't expect there to be any doubt about that. It is true that PRs can only be merged with two core member approvals, but that doesn't mean non-core member's reviews are useless. I (and I can probably speak for the rest of the core team) am very happy about reviews from everyone in the community! And you're doing a great job at that. But the effective results are often only comments pointing out issues and suggesting improvements. When you do a review and think the changes are fine, please also leave an approval! It helps a lot to see such feedback from the community. I don't think this should be only a core team job.

straight-shoota: @didactic-drunk Sry, I didn't notice your comments prior to my previous one. As mentioned there the history of this PR is really a mess.

I guess misinterpreted the tone in your message. It sounded a bit bitchy to me. Sorry about that.

RX14: Sorry, I didn't realise the documentation changes were applied from @straight-shoota.

It's really hard to review PRs with complex, long histories such as this one, and everyone gets it wrong. Let's just accept everyone makes mistakes and try to interpret other people's actions in good faith, as I try my best to do.

RX14:
@straight-shoota if we merge this PR will you provide an update PR to sort out the inconsistent docs?

src/digest/base.cr
didactic-drunk: I'm not holding a grudge.  My tone is and has been flat or questioning this entire post.  If only github had some sort of [tone or musical markup](https://github.com/watzon/nacl/issues/1#issuecomment-504715065).  Please reference the link for my usual tone when questioning my posts.

didactic-drunk: My first 3 commits were split in an attempt to make the review process faster/easier.  Did that work or should I have a single jumbo commit when posting PR's?

waj:
The changes look great, but I did run a benchmark of `MD5.hexdigest` and `SHA1.hexdigest` and seems like it's slightly slower now, specially for `SHA1`. I run the test 20 times for each case and here are the best times for each series:

Crystal 0.34.0
```
MD5.hexdigest:    0.293540   0.063665   0.357205 (  0.297224)
SHA1.hexdigest:   0.282686   0.046906   0.329592 (  0.285463)
```

This PR:
```
MD5.hexdigest:    0.299037   0.056454   0.355491 (  0.302478)
SHA1.hexdigest:   0.314224   0.061860   0.376084 (  0.318149)
```

Is not a big difference, but I couldn't find the reason yet

RX14:
@waj please use `Benchmark.ips`. We have a microbenchmark framework because it produces more meaningful data!

waj:
With `Benchmark.ips`:

Crystal 0.34.0:
```
 MD5.hexdigest:   3.48M (287.76ns) (± 1.39%)  160B/op
SHA1.hexdigest:   3.37M (296.51ns) (± 1.18%)  176B/op
```

This PR:
```
 MD5.hexdigest:   3.36M (297.21ns) (± 1.02%)  192B/op
SHA1.hexdigest:   3.24M (308.19ns) (± 1.00%)  208B/op
```

I tried disabling the state checking but didn't get a big improvement. My best guess is something with the `final_impl` receiving the destination buffer as argument.

Not big deal anyway. I just did the tests to check that anything so broken made it to master. I don't think we have to block this PR because of this.

Something interesting is if the test is performed executing only one of the digest algorithms, the number decreases slightly, because there are some method dispatch in the code.

Allocating in the stack would be great. Actually I tried changing `class` for `struct` and the improvement is quite important. But I had to change the block arguments by pointers to avoid the struct to be copied.

straight-shoota:
@didactic-drunk Could you resolve the conflict? Then it's finally ready to be merged.

didactic-drunk:
Done.

RX14:
@didactic-drunk thank you so much for your heavenly patience, this wasn't the best PR experience, but we're trying our best to get better!

watzon:
True that lol. Good to see this has been merged though. Good work on all sides. The OpenSSL bindings have been needing some love.

HankAnderson:
Wow, did anyone notice that there's a `macro inherited` here? What's that for? This is totally wrong.

straight-shoota:
The reason for `macro inherited` is to only define the class methods on implementing classes, not on the abstract base class. I don't think there's any other way to do this.

HankAnderson:
They can be defined on the base type and it will work. If you try to use it on the abstract type you will get an error anyway. It's just simpler.

straight-shoota:
They'd still show up in the API docs of `Digest::Base`. And that should be avoided.

HankAnderson:
@straight-shoota No, that's a good thing. "This is what you can expect from all digests" in a single place, instead of having the docs in every single digest.

straight-shoota:
> This is what you can expect from all digests

But that's is not what it says. It says: there's a method `Digest::Base.digest` that you can call.
But you can't call that method. The API documentation must not suggest otherwise. Look at https://crystal-lang.org/api/0.34.0/Digest%2FBase.html there are a couple of class methods listed, but none of them actually works.

This is different for instance methods because they're obviously only defined on instances and as an abstract class, `Digest::Base` doesn't have direct instances. So it's clear that any instance method applies to implementing types.
But an abstract type *could* have class methods that you can call directly on that abstract type. There's no way to tell whether a defined method can actually be called on that type or if it only works on inheriting types.

HankAnderson:
You are right.

But `macro inherited` is not the way to do it. It should be a module, and then each implementation extends it. Using `macro inherited` is always a smell in my opinion.

straight-shoota:
Yeah having a better method directly in the type system would be nice. But I don't think module extension is the right way (see #8488 for issues with that).

PR #8429, opened 2019-11-03
Support named arguments in macro methods

### LINKED ISSUE #8291: StringLiteral.camelcase does not handle (lower: true) parameter
```cr
module A
  macro define_method(decl)
    def {{decl.var.camelcase(lower: true)}}
      puts "does not work!"
    end
  end
end

class Z
  include A
  define_method test_column : String
end

Z.new.testColumn
```
Result:
```
There was a problem expanding macro 'define_method'

Code in test.cr:11:3

 11 | define_method test_column : String
      ^
Called macro defined in test.cr:2:3

 2 | macro define_method(decl)

Which expanded to:

 > 1 | def TestColumn
                     ^
Error: unexpected token: NEWLINE
```
Expected: handles lowercase and defines testColumn method

```

LLVM: 8.0.0
```

Blacksmoke16:
Fixes #8291

### DIFF
--- a/src/compiler/crystal/macros/interpreter.cr
+++ b/src/compiler/crystal/macros/interpreter.cr
@@ -363,11 +363,7 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
-        named_args = nil
-
-        if nargs = node.named_args
-          named_args = nargs.to_h { |arg| {arg.name, accept arg.value} }
-        end
+        named_args = node.named_args.try &.to_h { |arg| {arg.name, accept arg.value} }
 
         begin
           @last = receiver.interpret(node.name, args, named_args, node.block, self)
--- a/src/compiler/crystal/macros/interpreter.cr
+++ b/src/compiler/crystal/macros/interpreter.cr
@@ -366,7 +366,7 @@ module Crystal
         named_args = nil
 
         if nargs = node.named_args
-          named_args = nargs.each_with_object(Hash(String, ASTNode).new) { |arg, named_arg_hash| named_arg_hash[arg.name] = accept arg.value }
+          named_args = nargs.to_h { |arg| {arg.name, accept arg.value} }
         end
 
         begin
--- a/src/compiler/crystal/macros/interpreter.cr
+++ b/src/compiler/crystal/macros/interpreter.cr
@@ -363,9 +363,14 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
+        named_args = Hash(String, ASTNode).new
+
+        if (nargs = node.named_args)
+          nargs.each_with_object(named_args) { |arg, named_arg_hash| named_arg_hash[arg.name] = accept arg.value }
+        end
 
         begin
-          @last = receiver.interpret(node.name, args, node.block, self)
+          @last = receiver.interpret(node.name, args, named_args, node.block, self)
         rescue ex : MacroRaiseException
           raise ex
         rescue ex : Crystal::Exception
@@ -511,13 +516,15 @@ module Crystal
 
     def visit(node : Splat)
       node.exp.accept self
-      @last = @last.interpret("splat", [] of ASTNode, nil, self)
+      named_args = Hash(String, ASTNode).new
+      @last = @last.interpret("splat", [] of ASTNode, named_args, nil, self)
       false
     end
 
     def visit(node : DoubleSplat)
       node.exp.accept self
-      @last = @last.interpret("double_splat", [] of ASTNode, nil, self)
+      named_args = Hash(String, ASTNode).new
+      @last = @last.interpret("double_splat", [] of ASTNode, named_args, nil, self)
       false
     end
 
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/macros/interpreter.cr
```diff
@@ -363,9 +363,14 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
+        named_args = Hash(String, ASTNode).new
```
asterite: could this be nilable and only created when named args is present?

asterite: doing that also means we can pass nil in a couple of places below

Blacksmoke16: I suppose we could, but would mean that the macro methods would need each handle nil.  I thought it was a better UX if it's just empty.  Similar to how the `args` is just an empty array every time vs nil.

asterite: Named args are very uncommon, unlike args. That's why they are nilable. If we don't make them nilable it means. hash allocation for every macro method being evaluated. I'd rather avoid allocations if possible.

src/compiler/crystal/macros/interpreter.cr
```diff
@@ -363,9 +363,14 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
+        named_args = nil
+
+        if (nargs = node.named_args)
```
straight-shoota: Why doesn't the formatter remove this?

```suggestion
        if nargs = node.named_args
```

src/compiler/crystal/macros/methods.cr
```diff
@@ -644,7 +644,13 @@ module Crystal
           StringLiteral.new(@value + piece)
         end
       when "camelcase"
-        interpret_argless_method(method, args) { StringLiteral.new(@value.camelcase) }
+        interpret_argless_method(method, args) do
+          lower = ((nargs = named_args) && (lower_arg = nargs["lower"]?)) ? lower_arg : BoolLiteral.new(false)
```
straight-shoota: Minor stylistic issue: I'd prefer proper `if` branches here instead of a convoluted conditional operator.

src/compiler/crystal/macros/methods.cr
```diff
@@ -644,7 +644,13 @@ module Crystal
...
+
+          raise "named argument 'lower' to StringLiteral#camelcase must be a bool, not #{lower.class_desc}" unless lower.is_a?(BoolLiteral)
```
straight-shoota: ?

```suggestion
          raise "named argument 'lower' to StringLiteral#camelcase must be a BoolLiteral, not #{lower.class_desc}" unless lower.is_a?(BoolLiteral)
```

Blacksmoke16: All the other exceptions like this use the non `*Literal` style.  `must be a number`, `must be a string`, etc.

straight-shoota: I see. Lets leave it then. It just feels a bit unconnected because when you provide say a string, the error message would say `... must be a bool, not StringLiteral`.

Blacksmoke16: True, possibly could be another issue to determine how we want to display the error and to update all the current usages.

Blacksmoke16: True, possibly could be another issue to determine how we want to display the error and to update all the current usages.

Blacksmoke16:
This should be good now @straight-shoota

src/compiler/crystal/macros/interpreter.cr
```diff
@@ -363,9 +363,14 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
+        named_args = nil
+
+        if nargs = node.named_args
+          named_args = nargs.each_with_object(Hash(String, ASTNode).new) { |arg, named_arg_hash| named_arg_hash[arg.name] = accept arg.value }
```
asterite:
```suggestion
          named_args = nargs.to_h { |arg| {arg.name, accept arg.value} }
```

Blacksmoke16: Whoa, what is this?

asterite: https://crystal-lang.org/api/0.31.1/Enumerable.html#to_h(&block:T-%3ETuple(K,V))forallK,V-instance-method

:-)

src/compiler/crystal/macros/interpreter.cr
```diff
@@ -363,9 +363,14 @@ module Crystal
         end
 
         args = node.args.map { |arg| accept arg }
+        named_args = nil
+
+        if nargs = node.named_args
+          named_args = nargs.to_h { |arg| {arg.name, accept arg.value} }
+        end
```
asterite: 🎨 Optional! I guess we could change this to be:

```
args = node.args.map { |arg| accept arg }
named_args = node.named_args.try &.to_h { |arg| {arg.name, accept arg.value} }
```

So the `args` and `named_args` assignments are next to each other and it's clear we are mapping `node.args` and `node.named_args`.

PR #8442, opened 2019-11-05
Update distribution-scripts

bcardiff:
Nightly docker images and 0.32.0 will be based on bionic. -build images will have llvm-8.

The Linux CI will use llvm-8 after 0.32. Meanwhile is still using llvm-4

Note that Linux packages were already using llvm-8


bcardiff:
Oops I lost track of this one. Thanks!

Issue #8383, opened 2019-10-27
Can add `not` instance method to `Bool` class in std library, and `all_index` to Enumerable?


excosy:
1. Bool instance's`itself` method inheriting from Object, allow we select true value conveniently, with `&.itself`, but it is invalid for false value. So why don't add `not` method to Bool in std library, use ```&.not``` to select false value?

2. Since we have `select` to do `a.map.with_index{|x, i| x if x.meet_some_conditions}.compact!`, Is it need `all_index` method to do `a.map.with_index{|x, i| i if x.meet_some_conditions}.compact!`?

straight-shoota:
1. It already exists as `Object#!`.
2. What's the usecase for this?

yxhuvud:
1: Also there is `reject` to select false values.

excosy:
> 1: Also there is `reject` to select false values.

Yes, it is. I mean, this opposite belongs to Bool itself, not Enumerator's method. This means, we must define opposite method with each Enumerator's method, such as {select: reject, take_while: skip_while, ...}. Is it necessary? And, There is no opposite method of `find` and `index`, even unnecessary .

excosy:
> 1. It already exists as `Object#!`.
> 2. What's the usecase for this?

1. Object#! can't write as `count(&.!)`, with ```Error: expecting any of these tokens```. We need a word rather than a symbol method. And it is same to `~`. Furthermore, is there method defined with word of Bool or class derived from Bool, to override——we can't write `define !`, such as `and`, `xor`, `nand` etc. ?

2. There are some cases that index is significative variable——such as used to calcalute offset of another array——and value is only used by `meet_some_conditions` filter. Using array rather than hash to store it is in view of performance, since the index is finite numeric.

straight-shoota:
1. I think that's actually just a bug. `&.!` should work. But yes, you can't reimplemt that method because its a pseudo method provided by the compiler. If you need different (overridable) behaviour, you need to define a custom method for this. But that doesn't require that method being in the stdlib.
2. So this would essentially be an extension to `Enumerable#index` which returns the index of the *first* matching element to return the indices of *all* matching elements. Sounds reasonable.

wontruefree:
can this be closed since https://github.com/crystal-lang/crystal/pull/8445#issuecomment-554316878 is merged?

asterite:
Closed by #8445

PR #8445, opened 2019-11-05
Implement .! in compiler

jan-zajic:
Implements the boolean negation to be written as a regular method call.

I first see mentioned it here in #8327 where @straight-shoota documented it, but it was never implemented. Second time he mentioned it in issue #8383. Also compiler have warning `'!' is a pseudo-method and can't be redefined` but if you try used it, current stack is:
```
expecting any of these tokens: IDENT, CONST, +, -, *, /, //, %, |, &, ^, ~, **, <<, <, <=, ==, !=, =~, !~, >>, >, >=, <=>, ===, [], []=, []?, [, &+, &-, &*, &** (not '!') (Crystal::SyntaxException)
	 from src/compiler/crystal/syntax/lexer.cr:3001:9 in 'raise'
	 from src/compiler/crystal/syntax/lexer.cr:3000:5 in 'raise'
	 from src/compiler/crystal/syntax/parser.cr:267:11 in 'check'
```
ie it not pass even `lexer` and in `parser` it's not implemented.

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -130,6 +130,7 @@ module Crystal
     it_parses "+ 1", Call.new(1.int32, "+")
     it_parses "~ 1", Call.new(1.int32, "~")
     it_parses "1.~", Call.new(1.int32, "~")
+    it_parses "1.!", Not.new(1.int32)
     it_parses "1 && 2", And.new(1.int32, 2.int32)
     it_parses "1 || 2", Or.new(1.int32, 2.int32)
     it_parses "&- 1", Call.new(1.int32, "&-")
@@ -375,6 +376,7 @@ module Crystal
     it_parses "foo &.[0]", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Call.new(Var.new("__arg0"), "[]", 0.int32)))
     it_parses "foo &.[0] = 1", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Call.new(Var.new("__arg0"), "[]=", 0.int32, 1.int32)))
     it_parses "foo(&.is_a?(T))", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], IsA.new(Var.new("__arg0"), "T".path)))
+    it_parses "foo(&.!)", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Not.new(Var.new("__arg0"))))
     it_parses "foo(&.responds_to?(:foo))", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], RespondsTo.new(Var.new("__arg0"), "foo")))
     it_parses "foo &.each {\n}", Call.new(nil, "foo", block: Block.new(["__arg0".var], Call.new("__arg0".var, "each", block: Block.new)))
     it_parses "foo &.each do\nend", Call.new(nil, "foo", block: Block.new(["__arg0".var], Call.new("__arg0".var, "each", block: Block.new)))
@@ -941,6 +943,12 @@ module Crystal
       )
     )
 
+    it_parses "foo.!", Not.new("foo".call)
+    it_parses "foo.!.!", Not.new(Not.new("foo".call))
+    it_parses "foo.!(  )", Not.new("foo".call)
+
+    it_parses "foo &.!", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Not.new(Var.new("__arg0"))))
+
     # multiline pseudo methods (#8318)
     it_parses "sizeof(\n  Int32\n)", SizeOf.new(Path.new("Int32"))
     it_parses "instance_sizeof(\n  Int32\n)", InstanceSizeOf.new(Path.new("Int32"))
@@ -952,6 +960,7 @@ module Crystal
     it_parses "1.is_a?(\n  Int32\n)", IsA.new(1.int32, Path.new("Int32"))
     it_parses "1.responds_to?(\n  :foo\n)", RespondsTo.new(1.int32, "foo")
     it_parses "1.nil?(\n)", IsA.new(1.int32, Path.global("Nil"), nil_check: true)
+    it_parses "1.!(\n)", Not.new(1.int32)
 
     it_parses "/foo/", regex("foo")
     it_parses "/foo/i", regex("foo", Regex::Options::IGNORE_CASE)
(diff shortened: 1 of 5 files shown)


asterite:
Hi!

Sorry, I don't understand what this does. Could you explain what you can do with this change? If it's about being able to define `! ` as a method, or redefine it, that's a no no

jan-zajic:
@ asterite No. It's a missing language feature as now documented in `src/docs_pseudo_methods.cr`. 
> This method is a unary operator and usually written in prefix notation # (`!foo`) but it can also be written as a regular method call (`foo.!`).

Introduced in #8327 which you reviewed without comments. So if it's true this PR fix bug in compiler.
With this change you can basically do `anyValue.!`, but for details see a newly introduced `negation_method_spec.cr`.

straight-shoota:
@asterite Currently you can't call the pseudo method `!` using regular call syntax: `foo.!` is a syntax error. You can only use prefix notation: `!foo`.

This PR seems to fix that.

In the mentioned issues I was under the impression that this already works, as it does with any other pseudo method. But it seems `!` wasn't taken care of. So currently, the documentation for `Object#!` is incorrect.

spec/compiler/codegen/is_nil_spec.cr
```diff
@@ -0,0 +1,27 @@
...
+describe "Codegen: nil?" do
+  it "codegens .nil? false for simple type" do
+    run("1.nil?").to_b.should be_false
+  end
+
+  it "codegens .nil? true for nil" do
+    run("nil.nil?").to_b.should be_true
+  end
+
+  it "codegens .nil?() false for simple type" do
```
straight-shoota: We don't need duplicate codegen specs for syntax alternatives. There should be specs that ensure both method calls (with and without parentheses) are semantically equal. By logical conclusion, codegen must also be identical so it doesn't need separate specs.
Ditto all below.

asterite:
Oooh... I see, I get it now. This is great, thank you!

However, this is just about parsing and expanding code. I think tests for parsing a d expanding code are needed, but semantic and codegen tests are not needed, like straight shoota mentions. That's why I was confused, I thought it changed more than parsing.

jan-zajic:
Ok, i can rework tests, i just don't known how they are structured and need them for implementing it.

bcardiff:
@jan-zajic the specs should be in `spec/compiler/normalize/not_spec.cr`. In that directory, there are some of the specs regarding literal_expander. So `expr.!` is normalized to `!expr`.

The parsing of `expr.!` needs to be like a regular call. So the changes in the parser should allow that syntax and return `Call` instead of `Not`.

That should eliminate the need to handle `when .!` at a parser level.

asterite:
> So the changes in the parser should allow that syntax and return Call instead of Not

I'm not sure about this. `Not` is a special call and it's handled specially by the compiler. Doing it with `Call` won't do type flow stuff.

jan-zajic:
@bcardiff ok i try if that is possible

jan-zajic:
I came out of it that other pseudo-methods `:is_a?, :as, :as?, :responds_to?, :nil?` are currently implemented same, ie without call.

bcardiff:
Ok, my bad. Sorry for the noise.

jan-zajic:
@asterite, @bcardiff I added aditional tests to parser spec and to case expanding spec. I also simplify codegen specs. No one question remains: should i remove that whole `negation_method_spec.cr` in codegen? I'd prefer to keep it, to document current behavior and test if it works as all, not only parsing. I do not share your opinion it must works if parser and expanding works.. 

Semantically `Not` now can be on places where it wasn't before. The "order of words in sentences" can now be different. Also that codegen folder with real code examples is good place to look, when documentation for something is confusing/incomplete/missing and you need to really know, if something works that way and is not broken. Imho we can remove file `spec/compiler/codegen/case_spec.cr` as well, because case is transformed to `if/elsif/else`? As such it has no effect to code generation. 

I think that codegen specs currently test if AST tree can produce **working** program and cover branches of semantic visitor that are not covered elsewhere. We can for example move that kind of tests to some different "integration" folder, but it's really that important? Is there some pressure on reducing the number of specs? I think coverage is more important.

But it's just my opinion, can just delete `negation_method_spec.cr` file and PR should be valid without it, if you make consensus.

asterite:
> Semantically Not now can be on places where it wasn't before

Yes, but `!1` produces the same `ASTNode` as `1.!` so if codegen works for the first one (and there are already specs for that) then it will definitely work for the second one.

The only place where a spec might be needed is in `when .!`, though that again is just replacing an `ImplicitObj` with a `Not` and then it eventually leads to the same codegen, because it will boil down to an `if`.

Also, these specs, if we really want to keep them, should go into the existing `codegen/not_spec.cr`.

> Imho we can remove file spec/compiler/codegen/case_spec.cr as well

I agree. It might be the case that `case` wasn't expanded in the past and there was specialized code for it, I don't know.

> Is there some pressure on reducing the number of specs?

Yes, CI already takes a lot of time and for 32 bits builds we had to split specs to avoid exhausting the memory. More specs will make it more likely that this will happen again.

jan-zajic:
@asterite OK, thank you for your time.

spec/compiler/parser/parser_spec.cr
```diff
@@ -1650,7 +1660,7 @@ module Crystal
 
     assert_syntax_error "def foo(var : Foo+); end"
 
-    %w(&& || !).each do |name|
```
straight-shoota: This change should only remove `!`, not `||`. This is a string array literal, not a logic expression ;)

We could however consider allowing regular call syntax for `&&` and `||` as well, and remove this section entirely.

jan-zajic: Oh, thank you, good eyes! You have a good point.

jan-zajic:
@asterite please review this, thank you.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -3909,6 +3928,11 @@ module Crystal
       end_location = token_end_location
       doc = @token.doc
 
+      if @token.type == :"!"
+        obj = Var.new("self").at(location)
+        return parse_negation_suffix(obj)
+      end
```
asterite: I don't think this is needed.

This would turn `!` into `!self`, but you never want to do that.

In other cases plain `is_a?` it turned into `self.is_a?`, but for `!` it doesn't make sense.

Maybe remove it and check that no parser spec is broken?

jan-zajic: It's currently takes effect at `parse_when_expression`, it breaks if removed.
On other places `parse_var_or_call` is not called if token is :!. So i can leave it, or make extra condition in `parse_when_expression` which can be a little bit messy.

asterite: Ah, I see. Maybe add a comment saying that this will only trigger from `parse_when_expression`?

jan-zajic: yes, added.

asterite:
Is there any reason darwin in CI takes aeons to finish? All other build steps finish relatively quickly.

bcardiff:
> Is there any reason darwin in CI takes aeons to finish? All other build steps finish relatively quickly.

'cause there is a single queue for osx. And they need to wait a lot sometimes.

jan-zajic:
> 
> 
> > Is there any reason darwin in CI takes aeons to finish? All other build steps finish relatively quickly.
> 
> 'cause there is a single queue for osx. And they need to wait a lot sometimes.

Maybe someone with permissions push that Rerurn button? if there is any chance at all it finish without timing out (same for test_preview_mt on almost all my PRs).

asterite:
`test_preview_mt` is expected to fail (I can't remember why)

jan-zajic:
So @asterite, can you approve this, or something else i can done?

PR #8446, opened 2019-11-06
Update named_tuple.cr (typo)

cserb:
typo

### DIFF
--- a/src/named_tuple.cr
+++ b/src/named_tuple.cr
@@ -30,9 +30,9 @@ struct NamedTuple
   # useful in macros and generic code.
   #
   # ```
-  # {}             # syntax error (tuple literal)
-  # NamedTuple.new # => {}
   # NamedTuple.new(name: "Crystal", year: 2011) #=> {name: "Crystal", year: 2011}
+  # NamedTuple.new # => {}
+  # {}             # syntax error
   # ```
   def self.new(**options : **T)
     options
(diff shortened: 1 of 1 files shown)


src/named_tuple.cr
```diff
@@ -26,7 +26,7 @@ struct NamedTuple
   # Creates a named tuple that will contain the given arguments.
   #
   # This method is useful in macros and generic code because with it you can
-  # creates empty named tuples, something that you can't do with a tuple literal.
+  # create empty named tuples, something that you can't do with a tuple literal.
```
jhass: Maybe a good opportunity to rephrase slight? :)

> With a named tuple literal you cannot create an empty named tuple. This method doesn't have this limitation, which makes it especially useful in macros and generic code.

cserb: I also thought about rephrasing, but couldn't come up with a concise solution. Yours is not better than the original one imho :grin:

cserb: Sorry, I should have mentioned, that just rephrasing doesn't add much value for me. I would have preferred adding more context

straight-shoota: > *options* can be empty, so this method can create an empty `NamedTuple`. A tuple literal can't express this.

cserb: I took @jhass rephrasing and changed the order of the examples to emphasize the method's use-case. Have look

cserb:
> Please leave the order of examples as it was before.

done

PR #8447, opened 2019-11-06
Add missing documentation to String methods

jan-zajic:
n/c

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -229,6 +229,18 @@ describe "String" do
     it "gets byte_slice with negative index" do
       "hello".byte_slice(-2, 3).should eq("lo")
     end
+
+    it "gets byte_slice(Int) with with start out of bounds" do
+      expect_raises(IndexError) do
+        "hello".byte_slice(10)
+      end
+    end
+
+    it "gets byte_slice(Int) with with start out of bounds" do
+      expect_raises(IndexError) do
+        "hello".byte_slice(-10)
+      end
+    end
   end
 
   describe "to_i" do
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -230,13 +230,10 @@ describe "String" do
       "hello".byte_slice(-2, 3).should eq("lo")
     end
 
-    it "gets byte_slice(Int) with with start out of bounds" do
+    it "gets byte_slice(Int) with start out of bounds" do
       expect_raises(IndexError) do
         "hello".byte_slice(10)
       end
-    end
-
-    it "gets byte_slice(Int) with with start out of bounds" do
       expect_raises(IndexError) do
         "hello".byte_slice(-10)
       end
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -932,6 +932,14 @@ describe "String" do
     it { "foo".byte_index('o'.ord).should eq(1) }
     it { "foo bar booz".byte_index('o'.ord, 3).should eq(9) }
     it { "foo".byte_index('a'.ord).should be_nil }
+    it { "foo".byte_index('a'.ord).should be_nil }
+    it { "foo".byte_index('o'.ord, 3).should be_nil }
+    it {
+      "Dizzy Miss Lizzy".byte_index('z'.ord).should eq(2)
+      "Dizzy Miss Lizzy".byte_index('z'.ord, 3).should eq(3)
+      "Dizzy Miss Lizzy".byte_index('z'.ord, -4).should eq(13)
+      "Dizzy Miss Lizzy".byte_index('z'.ord, -17).should be_nil
+    }
 
     it "gets byte index of string" do
       "hello world".byte_index("he").should eq(0)
--- a/src/string.cr
+++ b/src/string.cr
@@ -868,7 +868,7 @@ class String
     char_at(index) { raise IndexError.new }
   end
 
-  # Returns the `Char` at the given *index*, or yields if out of bounds.
+  # Returns the `Char` at the given *index*, or result of running the given block if out of bounds.
   #
   # Negative indices can be used to start counting from the end of the string.
   #
@@ -910,7 +910,7 @@ class String
   # This method should be avoided,
   # unless the string is proven to be ASCII-only (for example `#ascii_only?`),
   # or the byte positions are known to be at character boundaries.
-  # Otherwise the characters are splitted, which leads to invalid UTF-8 values.
+  # Otherwise, multi-byte characters may be split, leading to an invalid UTF-8 encoding.
   #
   # Raises `IndexError` if the *start* index is out of bounds.
   #
@@ -924,9 +924,9 @@ class String
   # "hello".byte_slice(-2, 5)  # => "he"
   # "¥hello".byte_slice(0, 2)  # => "¥"
   # "¥hello".byte_slice(2, 2)  # => "he"
-  # "¥hello".byte_slice(0, 1)  # => "�"
-  # "¥hello".byte_slice(1, 1)  # => "�"
-  # "¥hello".byte_slice(1, 2)  # => "�h"
+  # "¥hello".byte_slice(0, 1)  # => "�" (invalid UTF-8 character)
+  # "¥hello".byte_slice(1, 1)  # => "�" (invalid UTF-8 character)
+  # "¥hello".byte_slice(1, 2)  # => "�h" (invalid UTF-8 character)
   # "hello".byte_slice(6, 2)   # raises IndexError
   # "hello".byte_slice(-6, 2)  # raises IndexError
   # "hello".byte_slice(0, -2)  # raises ArgumentError
@@ -972,8 +972,10 @@ class String
   # *start* can can be negative to start counting
   # from the end of the string.
   #
-  # Be careful when working with multibyte characters - they can be splitted
-  # which may lead to unexpected result.
+  # This method should be avoided,
+  # unless the string is proven to be ASCII-only (for example `#ascii_only?`),
+  # or the byte positions are known to be at character boundaries.
+  # Otherwise, multi-byte characters may be split, leading to an invalid UTF-8 encoding.
   #
   # Raises `IndexError` if *start* index is out of bounds.
   #
@@ -982,7 +984,7 @@ class String
   # "hello".byte_slice(2)  # => "llo"
   # "hello".byte_slice(-2) # => "lo"
   # "¥hello".byte_slice(2) # => "hello"
-  # "¥hello".byte_slice(1) # => "�hello"
+  # "¥hello".byte_slice(1) # => "�hello" (invalid UTF-8 character)
   # "hello".byte_slice(6)  # raises IndexError
   # "hello".byte_slice(-6) # raises IndexError
   # ```
@@ -2551,6 +2553,8 @@ class String
   # Returns `true` if this string is equal to `*other*.
   # Comparison is done byte-per-byte: if a byte is different from the corresponding
   # byte, `false` is returned and so on.
+  #
+  # See `#compare` for more comparison options.
   def ==(other : self) : Bool
     return true if same?(other)
     return false unless bytesize == other.bytesize
--- a/src/string.cr
+++ b/src/string.cr
@@ -729,7 +729,7 @@ class String
   #
   # Negative indices can be used to start counting from the end of the string.
   #
-  # Raises `IndexError` if the *index* is out of range.
+  # Raises `IndexError` if the *index* is out of bounds.
   #
   # ```
   # "hello"[0]  # => 'h'
@@ -746,7 +746,7 @@ class String
   # as character indices. Indices can be negative to start
   # counting from the end of the string.
   #
-  # Raises `IndexError` if the range's start is out of range.
+  # Raises `IndexError` if the range's start is out of bounds.
   #
   # ```
   # "hello"[0..2]   # => "hel"
@@ -759,7 +759,7 @@ class String
     self[*Indexable.range_to_index_and_count(range, size)]
   end
 
-  # Like `#[Range]`, but returns `nil` if the range's start is out of range.
+  # Like `#[Range]`, but returns `nil` if the range's start is out of bounds.
   #
   # ```
   # "hello"[6..7]? # => nil
@@ -771,17 +771,17 @@ class String
 
   # Returns a substring starting from the *start* character of size *count*.
   #
-  # The *start* argument can be negative to start counting
+  # *start* can can be negative to start counting
   # from the end of the string.
   #
-  # Raises `IndexError` if the *start* index is out of range.
+  # Raises `IndexError` if the *start* index is out of bounds.
   #
   # Raises `ArgumentError` if *count* is negative.
   def [](start : Int, count : Int)
     self[start, count]? || raise IndexError.new
   end
 
-  # Like `#[Int, Int]` but returns `nil` if the *start* index is out of range.
+  # Like `#[Int, Int]` but returns `nil` if the *start* index is out of bounds.
   def []?(start : Int, count : Int)
     raise ArgumentError.new "Negative count: #{count}" if count < 0
     return byte_slice?(start, count) if ascii_only?
@@ -855,7 +855,7 @@ class String
   #
   # Negative indices can be used to start counting from the end of the string.
   #
-  # Raises `IndexError` if the *index* is out of range.
+  # Raises `IndexError` if the *index* is out of bounds.
   #
   # ```
   # "hello".char_at(0)  # => 'h'
@@ -873,8 +873,11 @@ class String
   # Negative indices can be used to start counting from the end of the string.
   #
   # ```
-  # "hello".char_at(4) { 'x' } # => 'o'
-  # "hello".char_at(5) { 'x' } # => 'x'
+  # "hello".char_at(4) { 'x' }  # => 'o'
+  # "hello".char_at(5) { 'x' }  # => 'x'
+  # "hello".char_at(-1) { 'x' } # => 'o'
+  # "hello".char_at(-5) { 'x' } # => 'h'
+  # "hello".char_at(-6) { 'x' } # => 'x'
   # ```
   def char_at(index : Int, &)
     if ascii_only?
@@ -897,18 +900,18 @@ class String
     end
   end
 
-  # Returns a new string consisted of *count* bytes starting at *start* byte.
+  # Returns a new string built from *count* bytes starting at *start* byte.
   #
-  # The *start* argument can be negative to start counting
+  # *start* can can be negative to start counting
   # from the end of the string.
-  # If `count` is bigger than number of bytes from *start* to `bytelen`,
+  # If *count* is bigger than the number of bytes from *start* to `#bytesize`,
   # only remaining bytes are returned.
   #
   # Be careful when working with multibyte characters - they can be splitted,
   # which may lead to invalid UTF-8 values. These,
   # when asked as chars, will use the unicode replacement �.
   #
-  # Raises `IndexError` if the *start* index is out of range.
+  # Raises `IndexError` if the *start* index is out of bounds.
   #
   # Raises `ArgumentError` if *count* is negative.
   #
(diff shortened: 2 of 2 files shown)


src/string.cr
```diff
@@ -4112,11 +4120,13 @@ class String
     io << '}' if char.ord > 0xFFFF
   end
 
+  # Returns true if this string starts with a given *str*.
```
```suggestion
  # Returns `true` if this string starts with a given *str*, otherwise `false`.
```

ditto all below

src/string.cr
```diff
@@ -4318,10 +4330,16 @@ class String
     pointerof(@c)
   end
 
+  # Returns *count* of underlying bytes of this String starting at given *byte_offset* in an **unsafe** way.
+  #
+  # The returned slice is read-only.
```
```suggestion
  # NOTE: The returned slice is read-only.
```

src/string.cr
```diff
@@ -4318,10 +4330,16 @@ class String
...
   def unsafe_byte_slice(byte_offset, count)
     Slice.new(to_unsafe + byte_offset, count, read_only: true)
   end
 
+  # Returns the underlying bytes of this String starting at given *byte_offset* in an **unsafe** way.
+  #
+  # The returned slice is read-only.
```
```suggestion
  # NOTE: The returned slice is read-only.
```

src/string.cr
```diff
@@ -853,11 +853,13 @@ class String
     char_at(index) { yield }
   end
 
+  # Returns the `Char` at the given *index*, or raises `IndexError` if out of bounds.
```
asterite: It would be good to add code examples to all of the methods, showing possible cases. In this particular method I'd show an example for `char_at` of an good index and a bad index.

src/string.cr
r00ster91:
```suggestion
  # Returns `true` if this string starts with the given *str*, otherwise `false`.
```
"the" instead of "a" like for the others.

src/string.cr
straight-shoota: Minor stylistic thingy: The raised exceptions are typically not mentioned in the first sentence. Consider if this method raised more than one exception, it would be pretty convoluted. And from a semantic point, the purpose and primary behaviour of this method is to return a character. So it should be described as such.
It raises only in case of an error. That should be mentioned, but in a separate paragraph.

src/string.cr
```diff
@@ -853,11 +853,13 @@ class String
...
   def char_at(index : Int)
     char_at(index) { raise IndexError.new }
   end
 
-  def char_at(index : Int)
+  # Returns the `Char` at the given *index*, or yield if out of bounds.
```
straight-shoota:
```suggestion
  # Returns the `Char` at the given *index*, or yields if out of bounds.
```

src/string.cr
```diff
@@ -907,19 +909,25 @@ class String
...
   end
 
+  # Returns the codepoint of `Char` at the given *index*, or raises `IndexError` if out of bounds.
+  #
+  # See also: `Char#ord`.
   def codepoint_at(index)
     char_at(index).ord
   end
 
+  # Returns the `UInt8` at the given *index*, or raises `IndexError` if out of bounds.
```
straight-shoota: This should describe the meaning of the return value, not its type. The type could be added as return type restriction.
```suggestion
  # Returns the byte at the given *index*.
  #
  # Raises `IndexError` if out of bounds. 
```
Ditto below.

src/string.cr
jan-zajic: done

jan-zajic:
Hi @Sija, @asterite, @straight-shoota, @r00ster91, second round. Also i fix bug that i found during creating code samples in `byte_slice(Int)`: currently it raises ` Negative count (ArgumentError)` for large index (for example `"hello".byte_slice(6)`), which makes no sense. It should raise `IndexError`.

src/string.cr
```diff
@@ -4112,29 +4244,65 @@ class String
...
+  # "hello".starts_with?("he") # => true
+  # "hello".starts_with?("hu") # => false
+  # ```
+  def starts_with?(str : String) : Bool
     return false if str.bytesize > bytesize
     to_unsafe.memcmp(str.to_unsafe, str.bytesize) == 0
   end
 
-  def starts_with?(char : Char)
+  # Returns true if this string starts with the given *char*, otherwise `false`.
```
```suggestion
  # Returns `true` if this string starts with the given *char*, otherwise `false`.
```

ditto below

Sija: @jan-zajic ditto all below plz :)

straight-shoota:
Could you please squash the changes into two commits: One fixing the exception, the other adding the docs?

jan-zajic:
@straight-shoota yes, of course.

src/string.cr
```diff
@@ -4112,29 +4242,65 @@ class String
     io << '}' if char.ord > 0xFFFF
   end
 
-  def starts_with?(str : String)
+  # Returns true if this string starts with the given *str*, otherwise `false`.
```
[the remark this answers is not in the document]
straight-shoota: There are a few similar below.

src/string.cr
```diff
@@ -4112,29 +4242,65 @@ class String
...
+  def starts_with?(char : Char) : Bool
     each_char do |c|
       return c == char
     end
 
     false
   end
 
-  def starts_with?(re : Regex)
+  # Returns true if this string starts with the given *re* regular expression, otherwise `false`.
```
straight-shoota: Probably better: `... if the regular expression *re* matches at the start of this string`

src/string.cr
```diff
@@ -4151,7 +4317,17 @@ class String
     true
   end
 
-  def ends_with?(re : Regex)
+  # Returns true if this string ends with the given *re* regular expression, otherwise `false`.
```
straight-shoota: Ditto: `... if the regular expression *re* matches at the end of this string`

src/string.cr
```diff
@@ -4290,18 +4466,22 @@ class String
     char_index
   end
 
-  def clone
+  # Returns `self`
```
straight-shoota:
```suggestion
  # Returns `self`.
```
Ditto below.

src/string.cr
```diff
@@ -4290,18 +4466,22 @@ class String
...
+  def dup : String
     self
   end
 
+  # Returns `self`
   def to_s : String
     self
   end
 
+  # Appends `self` characters to the given IO object.
```
straight-shoota: More concise: ``Appends `self` to *io*.``

src/string.cr
```diff
@@ -4318,10 +4498,16 @@ class String
     pointerof(@c)
   end
 
+  # Returns *count* of underlying bytes of this String starting at given *byte_offset* in an **unsafe** way.
+  #
+  # The returned slice is read-only.
   def unsafe_byte_slice(byte_offset, count)
```
straight-shoota:
```suggestion
  def unsafe_byte_slice(byte_offset, count) : Slice
```
Ditto below.

src/string.cr
```diff
@@ -4318,10 +4498,16 @@ class String
     pointerof(@c)
   end
 
+  # Returns *count* of underlying bytes of this String starting at given *byte_offset* in an **unsafe** way.
```
straight-shoota: "unsafe way" needs to be more specific otherwise it doesn't mean anything.

wontruefree:
@jan-zajic thank you for all your work on docs!

src/string.cr
jan-zajic: I copy this from current docs for to_slice method: https://crystal-lang.org/api/0.31.1/String.html#to_slice:Bytes-instance-method . But i think we can remove this comment, because slice involve bound checks?

jan-zajic:
@straight-shoota i made fixes according to yours review + according to your comment:
ad 2, it's almost 50:50 in current String doc. All replace range to bounds in whole file..
ad 3, I agree with you, done.
1, 4, good point but left for another PR, as we already have a lot here.

src/string.cr
```diff
@@ -2938,7 +3070,16 @@ class String
     {pre, mid, post}
   end
 
-  def byte_index(byte : Int, offset = 0)
+  # Returns the index of *byte* in the string, or `nil` if the byte is not present.
+  # If *offset* is present, it defines the position to start the search.
```
straight-shoota: Could you add the note about offset being negative here as well?

src/string.cr
```diff
@@ -4290,23 +4490,27 @@ class String
...
   end
 
-  def clone
+  # Returns `self`.
+  def clone : String
     self
   end
 
-  def dup
+  # ditto
```
straight-shoota:
```suggestion
  # :ditto:
```
Ditto below.

straight-shoota:
Thank you very much!

src/string.cr
jan-zajic: But this method unlike the others does not support negative offset. It use to_unsafe[index] so it may lead to undefined behavior.. I hadn't noticed it before, we must rewrite it. This PR or another?

straight-shoota: I think such an small enhancement fits into this PR.

jan-zajic: @straight-shoota done in last commit.

src/string.cr
```diff
@@ -1020,8 +1021,7 @@ class String
   # "¥hello".byte_at(0)  # => 194
   # "¥hello".byte_at(1)  # => 165
   # "¥hello".byte_at(2)  # => 104
-  # "¥hello".byte_at(-1) # => 111
-  # "¥hello".byte_at(6)  # => 111
+  # "¥hello".byte_at(-1) # => 111  # "¥hello".byte_at(6)  # => 111
```
Sija: oops

jan-zajic: @Sija merge fixed

src/string.cr
```diff
@@ -878,11 +901,42 @@ class String
...
 
-  def byte_slice(start : Int, count : Int)
+  # Returns a new string consisted of *count* bytes starting at *start* byte.
+  #
+  # The *start* argument can be negative to start counting
+  # from the end of the string.
+  # If `count` is bigger than number of bytes from *start* to `bytelen`,
+  # only remaining bytes are returned.
+  #
+  # Be careful when working with multibyte characters - they can be splitted,
```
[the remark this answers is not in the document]
straight-shoota: Any thoughts on that?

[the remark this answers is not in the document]
jan-zajic: @straight-shoota so maybe like this?

straight-shoota:
@jan-zajic Could you please rebase on master?

jan-zajic:
@straight-shoota yes, of course. It's done.

spec/std/string_spec.cr
```diff
@@ -229,6 +229,18 @@ describe "String" do
     it "gets byte_slice with negative index" do
       "hello".byte_slice(-2, 3).should eq("lo")
     end
+
+    it "gets byte_slice(Int) with with start out of bounds" do
```
straight-shoota: There's a duplicate `with`.
And these two specs could be combined into one, they test the same thing (identical description).

src/string.cr
```diff
@@ -879,11 +905,53 @@ class String
...
+  #
+  # *start* can can be negative to start counting
+  # from the end of the string.
+  # If *count* is bigger than the number of bytes from *start* to `#bytesize`,
+  # only remaining bytes are returned.
+  #
+  # This method should be avoided,
+  # unless the string is proven to be ASCII-only (for example `#ascii_only?`),
+  # or the byte positions are known to be at character boundaries.
+  # Otherwise the characters are splitted, which leads to invalid UTF-8 values.
```
straight-shoota:
```suggestion
  # Otherwise multibyte characters can be splitted, which leads to invalid UTF-8 characters.
```

src/string.cr
```diff
@@ -879,11 +905,53 @@ class String
...
+  #
+  # ```
+  # "hello".byte_slice(0, 2)   # => "he"
+  # "hello".byte_slice(0, 100) # => "hello"
+  # "hello".byte_slice(-2, 3)  # => "he"
+  # "hello".byte_slice(-2, 5)  # => "he"
+  # "hello".byte_slice(-2, 5)  # => "he"
+  # "¥hello".byte_slice(0, 2)  # => "¥"
+  # "¥hello".byte_slice(2, 2)  # => "he"
+  # "¥hello".byte_slice(0, 1)  # => "�"
```
straight-shoota: Maybe these cases should have an explanatory comment?
```suggestion
  # "¥hello".byte_slice(0, 1)  # => "�" (invalid UTF-8 character)
```

src/string.cr
```diff
@@ -879,11 +905,53 @@ class String
...
+  # "hello".byte_slice(6, 2)   # raises IndexError
+  # "hello".byte_slice(-6, 2)  # raises IndexError
+  # "hello".byte_slice(0, -2)  # raises ArgumentError
+  # ```
+  def byte_slice(start : Int, count : Int) : String
     byte_slice?(start, count) || raise IndexError.new
   end
 
-  def byte_slice?(start : Int, count : Int)
+  # Like `byte_slice(Int, Int)` but returns `Nil` if the *start* index is out of bounds.
```
straight-shoota: I'd rather duplicate the entire documentation from `byte_slice` instead of referring to it. But I'm not sure what others think about it?

straight-shoota: (Eventually, it would be great if we could group similar methods together for the API docs and document them combined.)

src/string.cr
```diff
@@ -2423,7 +2553,10 @@ class String
     self if !blank?
   end
 
-  def ==(other : self)
+  # Returns `true` if this string is equal to `*other*.
+  # Comparison is done byte-per-byte: if a byte is different from the corresponding
+  # byte, `false` is returned and so on.
+  def ==(other : self) : Bool
```
straight-shoota: You could add  ``See `#compare` for more comparison options``.

src/string.cr
```diff
@@ -904,24 +972,86 @@ class String
...
 
-  def byte_slice(start : Int)
-    byte_slice start, bytesize - start
+  # Returns a substring starting from the *start* byte.
+  #
+  # *start* can can be negative to start counting
+  # from the end of the string.
+  #
+  # Be careful when working with multibyte characters - they can be splitted
+  # which may lead to unexpected result.
```
RX14: This paragraph should be the same as in `byte_slice?`.

src/string.cr
```diff
@@ -2423,7 +2553,10 @@ class String
     self if !blank?
   end
 
-  def ==(other : self)
+  # Returns `true` if this string is equal to `*other*.
+  # Comparison is done byte-per-byte: if a byte is different from the corresponding
+  # byte, `false` is returned and so on.
```
RX14: I don't think this really needs to be specified since there's only one valid encoding for the same string - whether comparing codepoints or bytes.

src/string.cr
```diff
@@ -2940,7 +3073,25 @@ class String
     {pre, mid, post}
   end
 
-  def byte_index(byte : Int, offset = 0)
+  # Returns the index of *byte* in the string, or `nil` if the byte is not present.
```
RX14: "Returns the index of the first ocurrence"?

src/string.cr
```diff
@@ -2949,7 +3100,20 @@ class String
     nil
   end
 
-  def byte_index(search : String, offset = 0)
+  # Returns the byte index of *search* in the string, or `nil` if the string is not present.
```
RX14: ditto

src/string.cr
```diff
@@ -4344,11 +4558,17 @@ class String
     pointerof(@c)
   end
 
-  def unsafe_byte_slice(byte_offset, count)
+  # Returns *count* of underlying bytes of this String starting at given *byte_offset*.
+  #
+  # The returned slice is read-only.
+  def unsafe_byte_slice(byte_offset, count) : Slice
```
RX14: Perhaps these should be `:nodoc:` instead?

They're unsafe because they don't do any bounds checking.

src/string.cr
RX14: see below

straight-shoota: Even better.

src/string.cr
jan-zajic: Same formulation is already at <=> method. We need it here not to distinguish byte and char comparsion but to document how strings are compared at all - by real content and not for example by memory address or any other way.

RX14: If it's similar to `<=>` then this is fine

src/string.cr
```diff
@@ -869,11 +895,53 @@ class String
...
+  # Like `byte_slice(Int, Int)` but returns `Nil` if the *start* index is out of bounds.
+  #
+  # Raises `ArgumentError` if *count* is negative.
+  #
+  # ```
+  # "hello".byte_slice(0, 2)   # => "he"
+  # "hello".byte_slice(0, 100) # => "hello"
+  # "hello".byte_slice(6, 2)   # => nil
+  # "hello".byte_slice(-6, 2)  # => nil
+  # "hello".byte_slice(0, -2)  # raises ArgumentError
```
Sija: `byte_slice` -> `byte_slice?`

src/string.cr
```diff
@@ -894,24 +962,88 @@ class String
...
-  def byte_at?(index)
+  # Returns the byte at the given *index*, or `nil` if out of bounds.
+  #
+  # ```
+  # "¥hello".byte_at(0)  # => 194
+  # "¥hello".byte_at(1)  # => 165
+  # "¥hello".byte_at(2)  # => 104
+  # "¥hello".byte_at(-1) # => 111
+  # "¥hello".byte_at(6)  # => 111
+  # "¥hello".byte_at(7)  # => nil
```
Sija: `byte_at` -> `byte_at?`

src/string.cr
```diff
@@ -4305,23 +4519,27 @@ class String
...
     self
   end
 
-  def dup
+  # :ditto:
+  def dup : String
     self
   end
 
+  # :ditto:
```
Sija: That should be the other way around - starting with documented `to_s(io)` and having next `to_s` with `:ditto:`

src/string.cr
jan-zajic: you're right! fixed

src/string.cr
jan-zajic: you're right! fixed

src/string.cr
jan-zajic: No. It's mean ditto dup i.e ditto clone i.e `Returns self.`

src/string.cr
jan-zajic: So? It's similar to other docs in std lib, for example Array# []? or Array#pop? or String#[]? and others. I think it would be sufficient to improve doc generator to turn these `` marks into html links.

src/string.cr
jan-zajic: I don't like undocumented public methods. It does not prevent from use them, only make them mysteriously hidden.. Please remove them or make clear their risks in doc.

src/string.cr
jan-zajic: @straight-shoota what you think about this proposal? Is it not against previous considerations? I think my draft is in line with other docs in string file..

straight-shoota: I'm not sure, did we discuss this before? @RX14's suggestion sounds reasonable, it's more specific.

src/string.cr
straight-shoota: I'd leave this as is for now.

src/string.cr
jan-zajic: Maybe, but it's same with current existing documentation for [index](https://crystal-lang.org/api/0.32.1/String.html#index(search:Regex,offset=0)-instance-method) method .. So maybe i have to change it too? And please write always complete formulation, what about for example:

1, _Returns byte index of the first ocurrence of *byte* in the string, or `nil` if not present._

or 

2, _Returns the index of the first ocurrence of *byte* in the string, or `nil` if not present._

or

3, _Returns the index of the first ocurrence in the string, or `nil` if not present._

or 

4, _Returns byte index of the first ocurrence in the string, or `nil` if not present._

or?

straight-shoota: Number 2 please.
And I'd like to change `#index` accordingly, that is add "first occurrence" to the existing phrase. For the regex overload: "first match"

jan-zajic:
@straight-shoota i rebase this on current master (again). Please can we finally merge this, before another conflicts arise?

src/string.cr
RX14: It seems this change was lost.

jan-zajic:
@RX14 it's done

PR #8450, opened 2019-11-07
Fix Hash#rehash

asterite:
It wasn't working properly: entries were rehashed but an old hash value was used to compute the index where they should go.

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -572,9 +572,12 @@ class Hash(K, V)
     # while moving non-deleted entries to the beginning (compaction).
     new_entry_index = 0
     each_entry_with_index do |entry, entry_index|
+      entry_hash = entry.hash
+
       if rehash
         # When rehashing we always have to copy the entry
-        set_entry(new_entry_index, Entry(K, V).new(key_hash(entry.key), entry.key, entry.value))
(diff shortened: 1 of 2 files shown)


src/hash.cr
```diff
@@ -572,9 +572,12 @@ class Hash(K, V)
     # while moving non-deleted entries to the beginning (compaction).
     new_entry_index = 0
     each_entry_with_index do |entry, entry_index|
+      entry_hash = entry.hash
```
straight-shoota: Wouldn't this be better placed in the `else` branch?

asterite: Maybe, I'm not sure what's the difference.

Sija: In this way you'd avoid one unnecessary call.
I guess that was @straight-shoota's point.

asterite: Sounds good!

(just note that `entry.hash` is just accessing an instance variable from a struct and that's like super cheap, plus I'm almost sure that LLVM will figure out that `entry_hash` is only reassigned in the `if` branch and so the right thing)

straight-shoota: Performance shouldn't be an issue. But it's easier to understand the code. This assignment is only relevant in the else branch. So it should only be there.

Issue #8329, opened 2019-10-15
[RFC] Add `Hash#compare_by_identity`


asterite:
In Ruby there's [`Hash#compare_by_identity`](https://ruby-doc.org/core-2.7.0.preview1/Hash.html#method-i-compare_by_identity). What it basically does is to stop comparing things with `eql?` and use `equal?`.

In Crystal it would stop using `==` and compare `object_id` of Reference types (and I guess we could try to give a compile-time error when calling this method and the key is not a Reference type).

In the compiler many times I want to have a Hash with a key of `ASTNode` but I don't want to compare nodes by their content, I want to compare by their object_id. In that case I create a Hash of `UInt64` (the type of `object_id`) and then I have to remember to put and ask always for `obj.object_id` which is a bit annoying and also bug-prone.

I also recently fixed a bug in Ruby's irb where I needed this: https://github.com/ruby/ruby/pull/2555 . I didn't know `compare_by_identity` existed until nobu suggested it in the issue, and that's where I learned about it.

Now, another alternative is to let Hash hold a general comparer so that one could provide any kind of key comparison. But I think we could still eventually do that, and `compare_by_identity` would just set a comparer that compares by identity. But maybe we don't need custom comparers except this one (at least in the code I wrote so far, I never needed a custom comparer by I did needed to compare by `object_id`).

This is very easy to implement in `Hash`.

Thoughts?

asterite:
We'd also add `Set#compare_by_identity` (Ruby has it too).

Some places where we could use this:
- https://github.com/crystal-lang/crystal/blob/0004179c9576c89d14e617423828af6f31e6d7a1/src/compiler/crystal/semantic/fix_missing_types.cr#L9
- https://github.com/crystal-lang/crystal/blob/d36a654c6aaa0a4097c9b134cdaeaed2905a91eb/src/compiler/crystal/tools/context.cr#L76
- https://github.com/crystal-lang/crystal/blob/1cbafcbb50d475a49137068c4a0dd47116ed62b1/src/compiler/crystal/codegen/exception.cr#L4
- https://github.com/crystal-lang/crystal/blob/6128faa77a9f9eb22ea23883870ab1f671dce939/src/compiler/crystal/semantic/bindings.cr#L220
- https://github.com/crystal-lang/crystal/blob/f7c0e753663b4e851a2cd25adc8c3ba810d93532/src/compiler/crystal/semantic/call_error.cr#L641
[...]

PR #8451, opened 2019-11-07
Hash compare by identity

asterite:
Implements #8329

Has a commit from #8450 but I can later rebase on top of `master`, or just rewrite git history.

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -921,11 +921,19 @@ class Hash(K, V)
 
   private def entry_matches?(entry, key)
     entry_key = entry.key
-    if @compare_by_identity && entry_key.responds_to?(:object_id)
-      if key.responds_to?(:object_id)
-        entry_key.object_id == key.object_id
-      else
+
+    if @compare_by_identity
+      if entry_key.responds_to?(:object_id)
+        if key.responds_to?(:object_id)
+          entry_key.object_id == key.object_id
+        else
+          false
+        end
+      elsif key.responds_to?(:object_id)
+        # because entry_key doesn't respond to :object_id
         false
+      else
+        entry_key == key
       end
     else
       entry_key == key
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -187,6 +187,9 @@ class Hash(K, V)
   # Otherwise guaranteed to be at least 3.
   @indices_size_pow2 : UInt8
 
+  # Whether to compare objects using `object_id`.
+  @compare_by_identity : Bool = false
+
   # The optional block that triggers on non-existing keys.
   @block : (self, K -> V)?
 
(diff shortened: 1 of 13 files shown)


src/hash.cr
```diff
@@ -893,17 +896,73 @@ class Hash(K, V)
...
+            K < Enum %}
+      entry.key == key && entry.hash != 0_u32
+    {% else %}
+      entry.hash == hash && entry_matches?(entry, key)
+    {% end %}
+  end
+
+  private def entry_matches?(entry, key)
+    entry_key = entry.key
+    if @compare_by_identity && entry_key.responds_to?(:object_id)
```
straight-shoota: What if `@compare_by_identity` is true and `entry` responds to `object_id` but `entry_key` doesn't? Shouldn't that also return false even if `entry_key == key`?

asterite: I don't know. If `entry_key` is an int you still want ints to compare the same way.

I also thought about raising a compile-time error when using `compare_by_identity` on a hash where keys don't respond to `object_id`, but it involved a chunk of macro code. Would something like that be better?

In general, I wouldn't worry too much about this. If you use `compare_by_identity` you generally know what you are doing and you'll be already working with reference types, or reference-like types. But we can try adding the macro check.

straight-shoota: I'm not sure what's the best semantics here. But I'd try to be consistent.

Currently, when comparing by identity the behaviour for differing answers to `responds_to?(:object_id)` is inconsistent:
* when `entry_key.responds_to?(:object_id) && !key.responds_to?(:object_id)` the result is `false`
* when  `!entry_key.responds_to?(:object_id) && key.responds_to?(:object_id)` the result is `entry_key == key`

straight-shoota: I'm not sure what's the best semantics here. But I'd try to be consistent.

Currently, when comparing by identity the behaviour for differing answers to `responds_to?(:object_id)` is inconsistent:
* when `entry_key.responds_to?(:object_id) && !key.responds_to?(:object_id)` the result is `false`
* when  `!entry_key.responds_to?(:object_id) && key.responds_to?(:object_id)` the result is `entry_key == key`

asterite: Sounds good, I'll at least make it consistent

asterite: Done! Let me know what you think.

asterite:
> Shouldn't dup and clone preserve the compare_by_identity?

Great catch! Done

bew:
Is it intentional to not have a way to switch back to "compare by hash" ?

straight-shoota:
Switching back an forth on the same instance would be confusing. `compare_by_identity` should be configured only once and prior to any usage of the hash. Ideally, that's in the constructor and I think we should actually make it a constructor argument (see discussion in #8329).

PR #8452, opened 2019-11-07
Deprecate Enumerable#grep

j8r:
`Enumerable#grep` aliases `Enumerable#select`, which has already multiple overloads.

Reasons of the change:
- `select` is more widely used, `grep` is only present in this very method
- `select` is a more descriptive English word about the action performed than `grep`
- follows the least aliases principle

Forum thread: https://forum.crystal-lang.org/t/remove-enumerable-grep/1323

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -521,6 +521,7 @@ module Enumerable(T)
   # ```
   # ["Alice", "Bob"].grep(/^A/) # => ["Alice"]
   # ```
+  @[Deprecated("Use `#select` instead")]
   def grep(pattern)
     self.select { |elem| pattern === elem }
   end
@@ -1220,9 +1221,10 @@ module Enumerable(T)
   #
   # ```
   # [1, 3, 2, 5, 4, 6].select(3..5) # => [3, 5, 4]
+  # ["Alice", "Bob"].select(/^A/)   # => ["Alice"]
   # ```
   def select(pattern)
-    self.select { |e| pattern === e }
+    self.select { |elem| pattern === elem }
   end
 
   # Returns the number of elements in the collection.


asterite:
I would personally do it the other way around: use grep, remove select with pattern, remove reject with pattern, add grep_v. But that's just my opinion.

straight-shoota:
Keeping `select` sounds more consistent too me as well.

The inverse operation is `reject(pattern)`. We'd have to rename that overload as well, or accept having differently named counterparts for `reject`.

asterite:
I'm mainly against `select` because of [this](https://forum.crystal-lang.org/t/can-anyone-explain-range-as-array-select-arguments-to-me/1320):

```crystal
ary = [1, 6, 2, 4, 8]
ary.select!(2..4) # what do you expect?
```

Sija:
@asterite explanation you've given in https://forum.crystal-lang.org/t/can-anyone-explain-range-as-array-select-arguments-to-me/1320/7 makes a perfect sense to me, imo once you know that `#select` operates on values then there's not much room for confusion left anymoar.

asterite:
> once you know that

I think that's the main issue. People will face the issue, learn that it doesn't work as they expect it, then learn and move on. But I'd rather have something that is intuitive right from the go.

straight-shoota:
I understand that there is some potential for misinterpretation. But it only affects the rather uncommon case when value and index type are identical or at least similar.
And IMO that's less confusing than having methods with two different names for the same behaviour, and the already mentioned inconsistency with `reject`.

I guess another option would be to rename `#select` to `#filter`, which is another typical name for this behaviour (both are aliases in Ruby's `Enumerable` for example). `filter` better expresses that it operates on values. But that would probably also a pretty bold change at this stage...

j8r:
`grep` is also confusing, because it doesn't behave as UN*X's `grep`, which match for a string, with optionally wilcards, not a regex-like pattern.
The method would need to be `egrep`, or `grep_e`, to fit the logic perfectly.

It's a bad idea to reproduce UN*X commands, because the methods are behaving differently from them.

Issue #8409, opened 2019-10-31
Channel.non_blocking_select not working with when


j8r:
By looking how to fix CI warnings related to `select` being deprecated, I compiled the compiler against master.
Using `non_blocking_select` produces a compiler error: 

```cr
ch = Channel(Nil).new
non_blocking_select
# Or Channel.non_blocking_select
when m = ch.receive
end
```

```
 3 | when m = ch.receive
     ^
Error: expecting token 'EOF', not 'when'
```

j8r:
Related commit: https://github.com/crystal-lang/crystal/commit/1254fd2fc87c5ea65f68bf06d9e75d5a49ee95ce

asterite:
Check this comment:

https://github.com/crystal-lang/crystal/commit/1254fd2fc87c5ea65f68bf06d9e75d5a49ee95ce#diff-28534fbfedc125e25dfc58732fa05c52R303-R307

The compiler will eventually produce calls to `Channel.select` or `Channel.non_blocking_select`

It's not that we'll introduce a `non_blocking_select` keyword. I don't think there's anything to do here, @bcardiff or @waj will later tackle this.

j8r:
So it's still WIP, we can't do anything to remove this warnings?

asterite:
Wait for the core team to handle that TODO (well, it should have been a TODO).

bcardiff:
The warning is only produced by a 0.31 compiler against master.
Using a 0.32-dev compiler is already warning free related to this deprecation.

The only thing left is to remove the deprecation after 0.32. I don't see a need for a `# TODO`

asterite:
Oh, sorry, I think I misunderstood this.

However, there's still an existing call that produces the warning:

https://github.com/crystal-lang/crystal/blob/f9a4a14a7f92c15f19086e983b18645b2738727e/spec/std/channel_spec.cr#L365

Can that be fixed now or we need to wait for 0.32.0? Maybe that's the warning that appears in CI.

bcardiff:
Well, deprecated method should still work. So there should be specs. (Otherwise we can fall in https://github.com/crystal-lang/crystal/pull/8257#discussion_r333728144)

The behaviour of `non_blocking_select` is tested in  `describe ".non_blocking_select"` so that spec will go away.

This push the story of _allow warnings in within these lines_ :-)

PR #8454, opened 2019-11-07
Fix warning on using Channel.select

asterite:
https://github.com/crystal-lang/crystal/issues/8409#issuecomment-548513517

### DIFF
--- a/spec/std/channel_spec.cr
+++ b/spec/std/channel_spec.cr
@@ -362,7 +362,7 @@ describe "unbuffered" do
 
   it "works with select else" do
     ch1 = Channel(Int32).new
-    Channel.select({ch1.receive_select_action}, true).should eq({1, Channel::NotReady.new})
+    Channel.non_blocking_select({ch1.receive_select_action}).should eq({1, Channel::NotReady.new})
   end
 
   it "can send and receive nil" do


wontruefree:
should there be a test for this?

bcardiff:
If deprecated code is not tested it is not guaranteed to stay working. 

I prefer a CI with warnings rather than untested deprecated methods. But If I'm the only one who thinks that way, we can go ahead.

asterite:
> If deprecated code is not tested it is not guaranteed to stay working

Oh, I didn't know that was the reason it was kept.

straight-shoota:
Maybe we could consider adding comments for such specs that are explicitly meant to test deprecated behaviour. Otherwise I expect we will eventually get more "fix warning" PRs in the future.

straight-shoota:
I'll award bonus points for having a mechanism for designating a spec as testing a specific deprecated method and disabling deprecation warnings for this method inside the specs body.

bcardiff:
Definitely. But is either be 

* a new keyword to be applied in a given lexical scope (as `checked` in https://github.com/crystal-lang/crystal/pull/6223) or
* an annotation that could express only for the whole method level. (not useful in specs)
* magic comments (I don't like it)

So, something like this?

```cr
pragma(allow_deprecated: true) { expr }
```

asterite:
I don't think this is worth our efforts. After all it's just a warning.

straight-shoota:
I agree. It was just an idea that popped into my head but should be filed under "in case you're bored and have nothing else to do".

straight-shoota:
I don't know why I kept thinking about this. But it bugged me. And I think this could actually be very simple to implement. Because this really only affects specs and deprecated methods should only be called *directly* in the spec body, a naive static filter should be enough.

```cr
require "spec"

@[Deprecated]
def deprecated
end

it "test deprecated method" do
  deprecated
end
```
This is the warning:
```
In deprecation_spec.cr:8:3

 8 | deprecated
     ^---------
Warning: Deprecated top-level deprecated.
```

`crystal spec`, after compiling the program reports the warnings. These could simply be filtered based on file and line number. If it's in a spec file and maybe has some kind of annotation naming the deprecated method that should be suppressed.

Obviously this won't work if the deprecated method is called indirectly. But that should be avoided anyway because it would mean the deprecated method is used in some code other than specs.

asterite:
The warnings are reported during compilation, not at runtime. I don't know if you took that into account.

straight-shoota:
Yes. Runtime doesn't matter. The `spec` command knows about the warnings an has the source code. So it can easily lookup the warning location in the source and check whether there is for example a comment designating this spec as testing a deprecated method.

RX14:
The warnings code already has a check for which directories to ignore warnings in (`lib`). It would be trivial to expose that on the crystal commandline, and add `spec`.

straight-shoota:
This off-topic discussion should be continued in #8477

PR #8458, opened 2019-11-09
Type node#name without generic args

Blacksmoke16:
Requires #8429.  Will make this a draft until thats merged, then can rebase.

This PR allows getting the name of a `TypeNode` without the type's type vars.  The main use case for this is when iterating over the subclasses of a parent type, where the children have type vars, and you're wanting to include data from a macro or annotation as generic type arguments.  Currently since the type vars were always included you couldn't just do like `{{"#{type.name}(#{ann[:theType]})".id}}` since that would expand to say `SomeType(T)(Foo)`.  This allows skipping the `(T)` to make supplying the generic types easier.

See https://github.com/Blacksmoke16/assert/blob/master/src/assert.cr#L152 where I have to `gsub` the type arguments away before adding my own.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -416,6 +416,16 @@ module Crystal
         assert_macro "", %({{"foo_bar".camelcase}}), [] of ASTNode, %("FooBar")
       end
 
+      it "executes camelcase with lower" do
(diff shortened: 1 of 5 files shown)


straight-shoota:
Maybe a a spec example that ensures the default behaviour with a generic type when ,`generic_args` is not set?

PR #8464, opened 2019-11-12
Return nil (and don't raise) when "Content-Type" header is an empty string.

### LINKED ISSUE #8398: Unhandled exception on HTTP::Handler
Seems a missing media type is caused unhandled exception in `MIME`

```crystal
Unhandled exception on HTTP::Handler

Parsing error: Missing media type ("") (MIME::Error)

  from /usr/share/crystal/src/mime/media_type.cr:228:19 in 'parse'

  from /usr/share/crystal/src/http/client/response.cr:67:7 in '->'

  from /usr/share/crystal/src/http/server/request_processor.cr:255:3 in 'process'

  from /usr/share/crystal/src/fiber.cr:255:3 in 'run'

  from ???
```

Sija:
This PR changes the behavior of `HTTP::Client::Response#mime_type` to return `nil` (and not raise) when `Content-Type` header is an empty string.

As per https://github.com/crystal-lang/crystal/issues/8398#issuecomment-547595695
Fixes #8398

### DIFF
--- a/spec/std/http/client/response_spec.cr
+++ b/spec/std/http/client/response_spec.cr
@@ -284,6 +284,11 @@ class HTTP::Client
       response.charset.should eq("UTF-8")
     end
 
+    it "returns content type as nil when empty (#8398)" do
+      response = Response.new(:ok, "", headers: HTTP::Headers{"Content-Type" => ""})
+      response.content_type.should be_nil
+    end
+
     it "returns status_code" do
       response = Response.new(:created)
       response.status_code.should eq 201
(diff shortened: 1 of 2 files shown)


bcardiff:
Can we get a spec (and a PR description that describes what and not how?)

Sija:
@bcardiff done!

bararchy:
@bcardiff did he get the bonus points for using string presence? ;)

bcardiff:
Of course!

image

wontruefree:
should the be the default behavior or `Headers#[]?`. I am not aware of a header where an empty string is a valid value.

straight-shoota:
@wontruefree According to [RFC 7230](https://tools.ietf.org/html/rfc7230#page-23) empty header values are legal, so we should honor that and don't confuse it with an absent header.

PR #8465, opened 2019-11-12
Don't close pipes used for signal handlers in MT

### LINKED ISSUE #8375: Multiple Faulty Behaviors when using Process inside spawn in MT enabled program
Compiled with preview_mt and release mode.

This is the error I get: 
```crystal
Unhandled exception in spawn: BUG: Fiber was awaken without channel delivery state set (Exception)

  from /usr/share/crystal/src/deque.cr:463:14 in 'my_class'

  from ***/my_class:94:20 in '->'

  from /usr/share/crystal/src/fiber.cr:255:3 in 'run'

  from ???

```

The main logic in the code is something like 

```crystal
def get_from_channel
  data = @channel.receive
  # do stuff with it
end

10.times do
  spawn do
    loop {get_from_channel}
  end
end
```

I'm not sure this will be enough to reproduce, and the code is closed source.

waj:
This should fix #8375 

In MT mode, `fork` is disabled and the event loop is not reinitialized. In other words, `fork` only works to run a process (i.e: do an `exec`) right away, without any context switch or IO operations in between.

### DIFF
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -242,7 +242,9 @@ module Crystal::Signal
       LibC.signal(signal, LibC::SIG_DFL) if signal.set?
     end
   ensure
-    @@pipe.each(&.close)
+    {% unless flag?(:preview_mt) %}
+      @@pipe.each(&.close)
+    {% end %}
   end
 
   private def self.reader


ysbaddaden:
Whatever that fork is disabled in MT: we still fork/exec to run executables, and we must reset signals to SIG_DFL and close the IO pipe in the forked process before calling exec.

bcardiff:
@ysbaddaden you mean to extract these lines https://github.com/crystal-lang/crystal/commit/e5b401f72c815036e89ba90b0c7930013a135f47#diff-d7e78211632011c14cdae00909be40ddR239-R241 without the pipe close, right?

It would be good to have at least a manual spec to ensure the behavior of signals is not broken. Otherwise it is doing changes in the dark. Do you have some use case/example to stress that signals are handled as expected?

ysbaddaden:
Actually, not closing the reader part of the pipe fixes #8375 for me. I don't know why, but it's okayish: even if the child process receives a signal it won't write it to the pipe (the writer part is closed) and the reader file descriptor will be closed by `exec` since `IO.pipe` already set CLOEXEC on it.

```diff
diff --git a/src/signal.cr b/src/signal.cr
index f4c094f25..1737b85a0 100644
--- a/src/signal.cr
+++ b/src/signal.cr
@@ -242,7 +242,7 @@ module Crystal::Signal
       LibC.signal(signal, LibC::SIG_DFL) if signal.set?
     end
   ensure
-    @@pipe.each(&.close)
+    writer.close
   end
 
   private def self.reader
```

NOTE: we disable signals for the current thread (sigmask) for reasons [explained in comments](https://github.com/crystal-lang/crystal/blob/master/src/signal.cr#L228-L239) and in d195adf4dcee8f5e726e3f663697248bf992e09f, 979a5285f443120aed345ddc941560b585d84004 and e5b401f72c815036e89ba90b0c7930013a135f47. The sigmask will be inherited by the child process, which *must* reenable signals, otherwise the exec'd process would have disabled signals, which is unexpected.

waj:
@ysbaddaden The signals are still being reset and the pipes have `close_on_exec` set by default. I just checked that subprocesses doesn't have extra file descriptors opened when they run.

waj:
Correction... @ysbaddaden you're right, we need to reenable the signals. I'll move the fix into the `Signal` class.

ysbaddaden:
Yes.

The sigset is updated *before* changing the signal handler so I think we don't even need to ever close the pipe (even without MT). LLVM could reorder instructions and cause the signal handler to be set before the sigset, but the likelihood for a thread to fork while another adds a signal handler _and_ the forked child to receive a single before exec is very very window —and impossible with a single thread anyway.

PR #8473, opened 2019-11-14
Implement JSON,YAML discriminator in runtime

straight-shoota:
Followup to #8406 which completely implements the discriminators as a runtime method instead of a macro.

The change from the previous implementation is small. But having it as a runtime method provides much more flexibility. For example, you can defining different sets of discrimination rules without having to define a supertype for each of them. And most importantly, discriminated types don't need to inherit from the same supertype.

So far, this is only the minimal change as a proof of concept.

1. The implementation is not specific to `Serializable`, so I'd suggest to place the method directly in the `JSON`/`YAML` namespaces.
2. Alternatively we could even consider to place the method on the pull parser for JSON. That would make some kind of sense. But not so much for YAML, I guess.
3. The method name `discriminate` is up for debate. Any ideas?
4. The definition of an initializer method is trivial, I'd suggest to completely remove the macro methods:
```cr
use_json_discriminator "type", {point: Point, circle: Circle}
# vs.
  JSON.discriminate(pull, "type", {point: Point, circle: Circle}
end
```
5. Support for aliases in YAML still needs to be implemented. But I presume that should not be an issue.
6. Specs and documentation will be added when the previous points are settled.

### DIFF
--- a/src/json/serialization.cr
+++ b/src/json/serialization.cr
@@ -391,40 +391,41 @@ module JSON
       {% end %}
 
       def self.new(pull : ::JSON::PullParser)
-        location = pull.location
+        ::JSON::Serializable.discriminate(pull, {{field}}, {{mapping}})
+      end
+    end
 
-        discriminator_value = nil
+    def self.discriminate(pull, field, mapping)
+      location = pull.location
 
-        # Try to find the discriminator while also getting the raw
-        # string value of the parsed JSON, so then we can pass it
-        # to the final type.
-        json = String.build do |io|
-          JSON.build(io) do |builder|
-            builder.start_object
-            pull.read_object do |key|
-              if key == {{field.id.stringify}}
-                discriminator_value = pull.read_string
-                builder.field(key, discriminator_value)
-              else
-                builder.field(key) { pull.read_raw(builder) }
-              end
+      discriminator_value = nil
+
+      # Try to find the discriminator while also getting the raw
+      # string value of the parsed JSON, so then we can pass it
+      # to the final type.
+      json = String.build do |io|
+        JSON.build(io) do |builder|
+          builder.start_object
+          pull.read_object do |key|
+            if key == field
+              discriminator_value = pull.read_string
+              builder.field(key, discriminator_value)
+            else
+              builder.field(key) { pull.read_raw(builder) }
             end
-            builder.end_object
           end
+          builder.end_object
         end
+      end
 
-        unless discriminator_value
-          raise ::JSON::MappingError.new("Missing JSON discriminator field '{{field.id}}'", to_s, nil, *location, nil)
-        end
+      unless discriminator_value
+        raise ::JSON::MappingError.new("Missing JSON discriminator field '#{field}'", to_s, nil, *location, nil)
+      end
 
-        case discriminator_value
-        {% for key, value in mapping %}
-          when {{key.id.stringify}}
-            {{value.id}}.from_json(json)
-        {% end %}
-        else
-          raise ::JSON::MappingError.new("Unknown '{{field.id}}' discriminator value: #{discriminator_value.inspect}", to_s, nil, *location, nil)
-        end
+      if target = mapping[discriminator_value]?
+        target.from_json(json)
+      else
+        raise ::JSON::MappingError.new("Unknown '#{field}' discriminator value: #{discriminator_value.inspect}", to_s, nil, *location, nil)
       end
     end
   end
--- a/src/yaml/serialization.cr
+++ b/src/yaml/serialization.cr
@@ -381,32 +381,32 @@ module YAML
         {% mapping.raise "mapping argument must be a HashLiteral or a NamedTupleLiteral, not #{mapping.class_name.id}" %}
       {% end %}
 
-      private def self.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
+      def self.new(ctx : YAML::ParseContext, node : YAML::Nodes::Node)
         ctx.read_alias(node, \{{@type}}) do |obj|
           return obj
         end
-
-        unless node.is_a?(YAML::Nodes::Mapping)
-          node.raise "expected YAML mapping, not #{node.class}"
-        end
-
-        node.each do |key, value|
-          next unless key.is_a?(YAML::Nodes::Scalar) && value.is_a?(YAML::Nodes::Scalar)
-          next unless key.value == {{field.id.stringify}}
-
-          discriminator_value = value.value
-          case discriminator_value
-          {% for key, value in mapping %}
-            when {{key.id.stringify}}
-              return {{value.id}}.new(ctx, node)
-          {% end %}
-          else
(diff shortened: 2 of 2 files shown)


asterite:
> For example, you can defining different sets of discrimination rules without having to define a supertype for each of them. And most importantly, discriminated types don't need to inherit from the same supertype.

This is a valid point but I fail to see how you can nest such union of types inside another type without a common base type.

For example there's [GeoJSON](https://tools.ietf.org/html/rfc7946) with different shapes. Say you want to have an object that holds a shape, and make that object be serialized to JSON, something like:

```crystal
require "json"
require "geojson"

class MyObject
  include JSON::Serializable

  property shape : ...???
end
```

What type so you put in `shape`assuming you don't want a base type for all shapes? How does that work?

> The implementation is not specific to Serializable, so I'd suggest to place the method directly in the JSON/YAML namespaces.

👍 

> Alternatively we could even consider to place the method on the pull parser for JSON. That would make some kind of sense. But not so much for YAML, I guess.

Probably not necessary.

> The method name discriminate is up for debate. Any ideas?

Looks good to me.

> The definition of an initializer method is trivial, I'd suggest to completely remove the macro methods:

I don't think it's trivial. You need to mention `JSON::PullParser` and some boilerplate. My guess is that most users using JSON, either through `JSON.mapping` or `JSON::Serializable` don't know (not need to know!) about `JSON::PullParser` and how it all works underneath it. If you do know, then good, you have more power to customize and understand things. But if you don't, asking users to define `new` which receives a `JSON::PullParser` (no idea what that is) is inconvenient in my opinion.

We can either keep `use_json_disciminator` or rename it to `def_new_json_disciminate`, but I prefer the first name (so the current one).

---

At first I thought this wasn't going to work because the type of the Hash values would be a union and that could end up resolving to a base type where some of the types don't necessarily provide a JSON serialization. Oh, in fact, it does happen:

```crystal
require "json"

abstract class Shape
end

class Point < Shape
  include JSON::Serializable

  property type : String
  property x : Int32
  property y : Int32
end

class Circle < Shape
  include JSON::Serializable

  property type : String
  property x : Int32
  property y : Int32
  property radius : Int32
end

class Rectangle < Shape
end

pull = JSON::PullParser.new(%<
  {
    "type": "point",
    "x": 1,
    "y": 2
  }
>)

value = JSON::Serializable.discriminate(pull, "type", {point: Point, circle: Circle})
p value
```

You get:

```
In src/json/from_json.cr:13:3

 13 | new parser
      ^--
Error: wrong number of arguments for 'Rectangle.new' (given 1, expected 0)

Overloads are:
 - Reference.new()
```

That's because the hash values have `Circle` and `Point` and so the compiler infers that to be `Shape` and then it tries `new` on any shape and fails in Rectangle.

I know this is a bit of cheating because I did use a hierarchy, but you see that moving this to runtime has this problem. At compile-time only the specific types that you give are `new`ed and this problem doesn't exist (so you could have a subtype that is a valid type but not necessarily belongs in the disciminated values).

Just because of this point I hesitate moving the logic to runtime, even more because we don't have use cases for non-hierarchical types.

straight-shoota:
> I fail to see how you can nest such union of types inside another type without a common base type.

There are valid use cases where you don't even need to do that. For example, you could read the data (whatever type it is) and use it directly for something. No need for storing it in an ivar.

The main benefit is being able to define contexts without needing to declare a type for each of them.

I suppose this could also be implemented using a macro to circumvent the type merging problem.

yxhuvud:
> use cases for non-hierarchical types.

Well, I don't de-serialize often, but I *very* often serialize things with a type discriminator (in other words, I am a heavy user of json:api), and what I serialize generally don't share a common ancestor (in a relevant way. It is Ruby after all). I fail to see why people de-serializing would be forced to introduce structure where the serializing side wouldn't need the same structure. 

Regardless of how you do it, there will be a runtime error when someone sends you a rectangle and your deserializer lookup table only support Points and Circles (perhaps not having implemented any Rectangle in the first place). What can be done is compiletime verification that all entries of the right-hand-side have a disciminator defined. 

(somewhat unrelated, but as we use custom serializers for our entries we use a very similar lookup table for emitting our json)

straight-shoota:
> What can be done is compiletime verification that all entries of the right-hand-side have a disciminator defined.

How's that supposed to work. At compile time we can't know which JSON properties an instance would emit.

yxhuvud:
@straight-shoota I meant that the field is there. Not that the value actually exist.

straight-shoota:
@yxhuvud Still not following... What field is supposed to be where? This doesn't have anything to do with `JSON::Field`.

yxhuvud:
@straight-shoota I was referring to the existance of the descriminator property/method in the RHS of the lookup entries. Sorry about the confusion.

RX14:
Why did this change back to being a a macro?

asterite:
@RX14 See https://github.com/crystal-lang/crystal/pull/8473#issuecomment-554336232

RX14:
I thought Class types wouldn't be merged :(

asterite:
It's up to us. That's the current behavior.

src/json/serialization.cr
```diff
@@ -391,40 +391,44 @@ module JSON
...
-                builder.field(key, discriminator_value)
-              else
-                builder.field(key) { pull.read_raw(builder) }
-              end
+        ::JSON::Serializable.discriminate(pull, {{field}}, {{mapping}})
+      end
+    end
+
+    macro discriminate(pull, field, mapping)
+      location = {{ pull }}.location
```
RX14: Don't we want to use `%fresh` macro vars now that this is a macro?

RX14:
This looks good, overall.

RX14:
With macros you can now have some of the args evaluated twice, they should be assigned to temp variables instead.

erdnaxeli:
The title should be changed, it is incorrect now it is implemented back as a macro. If I understand the patch, what it does is to split the current `use_json_discriminator` macro in two macros? So you colud use `discriminate` without using `use_json_discriminator`?

I must be missing something because I don't really see the point :sweat_smile:

straight-shoota:
> So you colud use discriminate without using use_json_discriminator?

Exactly. That's the point. With this change `use_json_discriminator` is no longer necessary. I just left it for now to demonstrate that its function can entirely be delegated to the less-intrusive `discriminate` macro.

The big difference is that `use_json_discriminator` defines a constructor on a type. You have to use that type for deserialization and it only provides a single discrimination strategy.
The `discriminate` macro can be used any place individually. If you want the same behaviour as with `use_json_discriminator`, you place it in a `.new` method of some type. But you don't have to. There's no need for having a type to which the discrimination is bound.

dammer:
By the way, I'd like to suggest adding a default discriminator value if possible. This is good for some structures where the discriminator field is optional.
Right now I'm using a renamed and patched use_json_discriminator for this. 
https://gist.github.com/dammer/a0d9a5ea425e849ea8425898805e9f5f
Lines 79-83

Issue #8449, opened 2019-11-07
Remove String#codepoint_at 


straight-shoota:
`String#codepoint_at` simply delegates to `char_at(index).ord`. Instead of being a short cut, `codepoint_at(index)` is actually even longer (by one char) than calling `char_at(index).ord` directly.
Even if it technically was functioning as a short cut, such a thing only makes sense when it is at least somewhat useful. Currently, I could only find two real usages of this method in `markd`, while looking [all of GitHub's Crystal repositories](https://github.com/search?l=Crystal&q=codepoint_at&type=Code). All other occurrences seem to be mirrors or re-implementations of the stdlib and a few specs.

I'm suggesting to deprecate and remove this method. It's not good to maintain a short cut that defeats its purpose and isn't really used.

asterite:
Just my opinion, but I think `codepoint` is more descriptive than `string[index].ord`. Doesn't mean we couldn't remove it.

straight-shoota:
I'd rather rename `Char#ord` to `#codepoint` then.

asterite:
Yes! We copied that from Ruby. It's not bad because it's short, but it's also cryptic. Then we have `1.chr` which is also a bit cryptic, but I don't know.

vlazar:
The `#codepoint` looks much more meaningful. A small cons is with `#ord` gone one more method would be missing for those of us coming from Ruby.

straight-shoota:
But it would probably be for good... 🤷‍♂ 😄

vlazar:
> Then we have `1.chr` which is also a bit cryptic, but I don't know.

So `ord` / `chr` for symmetry can be something like `codepoint` / `character` (or `char`)?

RX14:
`1.to_char` and `'a'.codepoint` would be what i'd design in a cleanroom. But changing `chr` -> `to_char` is not overly worth it.

vlazar:
@RX14 If it's `1.to_char`, why it's not `'a'.to_codepoint` then? :)

asterite:
Because in `'a'.codepoint` you are asking the codepoint of a char, but in `1.char` you are not asking the char of 1 (that doesn't make sense), you are asking to convert it to a char. But that also doesn't make sense. `chr` is a bit cryptic but it's the char for a given codepoint, so it might be good. Maybe `1.considering_it_as_a_codepoint_to_char`, but it's super verbose :-P

vlazar:
[...]

PR #8475, opened 2019-11-15
Deprecate String#codepoint_at

vlazar:
Ref #8449

I'm not sure if the use of `String#codepoint_at` should be removed in this PR or in follow-up PR where `String#codepoint_at` is removed.

I see some other deprecation warnings when running crystal specs, so I assume it may be intentional and serves as a reminder to remove usage later when the method itself is removed.

Let me know if `String#codepoint_at` usage should be removed in this PR.

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -907,6 +907,7 @@ class String
     byte_slice start, bytesize - start
   end
 
+  @[Deprecated("Use .char_at(index).ord instead")]
   def codepoint_at(index)
     char_at(index).ord
   end


straight-shoota:
Deprecated methods should not be used, so please remove its uses in stdlib.

But they continue to be tested in the spec suite, that's where the warnings come from (see discussion in #8454).

vlazar:
@straight-shoota Sorry, I'm confused now. By stdlib do you mean only the `src/` excluding `spec/`?

The only usage is currently in specs, and those specs are not for `codepoint_at` itself, it's just that it's being used for testing.

From https://github.com/crystal-lang/crystal/pull/8454#issuecomment-551261909
> I prefer a CI with warnings rather than untested deprecated methods. But If I'm the only one who thinks that way, we can go ahead.

So should I remove the usage of `codepoint_at` from specs in this PR?

straight-shoota:
Oh sry, I was under the impression there was a single "real" use somewhere in stdlib.

> By stdlib do you mean only the `src/` excluding `spec/`?

Well, essentially all uses except those where specifically this method is tested should be removed. In this case, there are none. But "escapes with unicode" acts as such, so I guess we should leave it for now and only change that with #8476.

asterite:
I'm still not sure why we removed this...

vlazar:
No real world usage as described in the original issue #8449?
> I could only find two real usages of this method in markd, while looking all of GitHub's Crystal repositories.

asterite:
I think the number of actual usages don't matter. Crystal is not a hugely popular language. It's good to have the option to easly get a codepoint of a char. That's what makes Ruby great, the many shortcuts you have to things, to avoid boilerplate code or doing many hoops. But if others think removing this was fine then it's fine.

vlazar:
It's not late yet to revert before the release.

straight-shoota:
@asterite Useful shortcuts are great. But `codepoint_at(index)` doesn't provide a huge benefit over `char_at(index).ord`. If we follow the discussion from #8449 and rename `ord` to `codepoint`, it would at least be a bit shorter, but still not provide a simplification in terms of reducing boilerplate etc.
So you have two ways to do the exact same thing and they are largely identical from a user perspective. Per this argument we try to avoid aliases, and this is essentially the same thing. Removing this method doesn't remove any features because there's an equivalent alternative. But it's one less feature to learn.

PR #8476, opened 2019-11-15
Remove String#codepoint_at

### LINKED ISSUE #8449: Remove String#codepoint_at

vlazar:
Fixes #8449

Follow-up to #8475

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1751,10 +1751,10 @@ describe "String" do
   end
 
   it "escapes with unicode" do
-    "\u{12}".codepoint_at(0).should eq(1 * 16 + 2)
-    "\u{A}".codepoint_at(0).should eq(10)
-    "\u{AB}".codepoint_at(0).should eq(10 * 16 + 11)
-    "\u{AB}1".codepoint_at(1).should eq('1'.ord)
+    "\u{12}".char_at(0).ord.should eq(1 * 16 + 2)
+    "\u{A}".char_at(0).ord.should eq(10)
+    "\u{AB}".char_at(0).ord.should eq(10 * 16 + 11)
+    "\u{AB}1".char_at(1).ord.should eq('1'.ord)
   end
 
   it "does char_at" do
--- a/src/string.cr
+++ b/src/string.cr
@@ -901,11 +901,6 @@ class String
     byte_slice start, bytesize - start
   end
 
-  @[Deprecated("Use .char_at(index).ord instead")]
-  def codepoint_at(index)
-    char_at(index).ord
-  end
-
   def byte_at(index)
     byte_at(index) { raise IndexError.new }
   end


straight-shoota:
Thanks for already submitting the next step. But I'd actually advise against doing that in the future. This PR adds noise and it's not necessary for tracking because there is #8449 (as well as the deprecation annotation in the source). And it's not a timing relevant issue.
Unless we're beating this really fast, it might hang around for quite some time without being actionable. In the worst case, such a PR needs a rebase in the future because master has progressed, which means it would need to be touched twice.

vlazar:
@straight-shoota That makes sense, took a note for the future.

How about I close this PR now to reduce noise and reopen it later if/when #8475 is merged in?

Sija:
I'd merge changes from this PR into #8475.

jmbarnardgh:
Crystal newcomer here... looks like checks failed on Nov 21, 2019 potentially due to cloud infra issue. Can we re-run checks again to see if cloud really was the issue?

vlazar:
@jmbarnardgh Nothing to worry about. There are occasional failures on CI. It should be better with merged master.

I'm allowing this PR to sit quiet for a couple of releases, as deprecated methods are usually removed not in the next release. I'll check the usage in ecosystem and send some PRs first to remove the use of deprecated `codepoint_at` and then update this PR.

vlazar:
So the `String#codepoint_at` was deprecated for 2 releases already.

With recent move to prepare for Crystal 1.0 it seems like a good time to review this PR.

Most importantly the real usage of `String#codepoint_at` in popular shard `markd` has been already removed https://github.com/icyleaf/markd/commit/4105181a5003ab2b383b58e7b9f7c98836d17a74

I've taken care of:

https://github.com/crystallabs/term_app - I've sent a PR
https://github.com/oren/website-generator - I've created an issue to notify it needs to be updated to new markd release

Other usages from GitHub search (very specific and old projects, can be safely ignored I think):

https://github.com/tbpgr/crystal_samples - stale, last commit 3 years ago, no reason to fix a single deprecated method
https://github.com/Philpax/valhalla - stale, last commit 4 years ago
https://github.com/txe/crystal-win - stale, last commit 3 years ago
https://github.com/crystal-lang-tools/intellij-crystal-lang - stale, last commit 1.5 years ago
https://github.com/ozra/onyx-lang - stale, last commit 3.5 years ago
https://github.com/TheKernelCorp/NuummiteOS - stale, last commit 2.5 years ago
https://github.com/UteroOS/utero - stale, last commit 2 years ago

vlazar:
I've noticed this comment in original issue which lead to deprecation of `String#codepoint_at`
https://github.com/crystal-lang/crystal/issues/8449#issuecomment-560368685

I think it makes a good point. We also have `String#each_codepoint` and `String#codepoints`. With deprecated `String#codepoint_at` those seems to be lonely.

So let's maybe review `String#codepoint_at` in the this bigger context?

For consistency we can:
- bring back deprecated `String#codepoint_at`
- remove `String#each_codepoint` and `String#codepoints` too
- do something else?

vlazar:
And on related note I like the proposed change in https://github.com/crystal-lang/crystal/pull/8533

- `Char#ord` -> `Char#codepoint`
- `Int#chr` -> `Int#char`

Final names might be different, but already much better.

This change would require a lot of PRs to be sent. Time consuming, but totally doable.

straight-shoota:
Considering https://github.com/crystal-lang/crystal/issues/8449#issuecomment-560368685 I tend towards keeping `each_codepoint`, `codepoints` and by extension also `codepoint_at`, effectively removing the deprecation. I think it makes sense to treat this set of methods equally.
There could be valid arguments for removing them all or even removing only `codepoint_at`, but I don't think any of these alternatives provides a strong benefit. In the end I'd prefer the status quo (in this case: revert the deprecation message) because of the least resistance.

vlazar:
@straight-shoota Then more core team members should vote on this? Otherwise it will be another stale PR/issue. I think this does not require a lot of thinking from each voter, just plenty of votes.

asterite:
I think these methods should stay.

bcardiff:
I don't think it hurts to leave them. So reverting the deprecation is sound good to me.

vlazar:
With some thumbs up https://github.com/crystal-lang/crystal/pull/8476#issuecomment-597587441 I'm closing this in favor of #8902

PR #8481, opened 2019-11-16
Some misbehaving SSL servers hang at the end instead of responding wi…

### LINKED ISSUE #7136: Https connection not closing
I tried building a "connect with microsoft" oauth2 application, but while building it, I noticed that the requests took over a minute and finally crashed the program.

Crystal code (doesn`t work):
```crystal
require "http/client"

puts HTTP::Client.post("https://login.microsoftonline.com/common/oauth2/token", form: {"hi" => "a"}).body
```

The same thing in Ruby (works):
```ruby
require 'net/http'
uri = URI('https://login.microsoftonline.com/common/oauth2/token')
res = Net::HTTP.post_form(uri, 'hi' => 'a')
puts res.body
```

Crystal 0.27.0 (2018-11-02)

LLVM: 6.0.1

I couldn't try it with Crystal master because it doesn't compile on my machine.

OS: Manjaro Linux

rdp:
…th an SSL close_notify alert.  the docs for SSL_shutdown say 'If a unidirectional shutdown is enough (the underlying connection shall be closed anyway), this first successful call to SSL_shutdown() is sufficient.' which seems to fix the problem.  Also capture Errno's during close instead of bubbling them up, same as IO::Errors. Closes #7136

Sorry I couldn't format the commit message easily on the command line...

### DIFF
--- a/spec/manual/https_client_spec.cr
+++ b/spec/manual/https_client_spec.cr
@@ -12,8 +12,4 @@ describe "https requests" do
       break
     end
   end
-
-  it "can fetch from IIS servers that don't shutdown gracefully" do
-    HTTP::Client.post("https://login.microsoftonline.com/common/oauth2/token", form: {"hi" => "a"}).body
-  end
 end
--- a/spec/manual/https_client_spec.cr
+++ b/spec/manual/https_client_spec.cr
@@ -12,4 +12,8 @@ describe "https requests" do
       break
     end
   end
+
+  it "can fetch from IIS servers that don't shutdown gracefully" do
+    HTTP::Client.post("https://login.microsoftonline.com/common/oauth2/token", form: {"hi" => "a"}).body
+  end
 end
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -36,22 +36,21 @@ describe OpenSSL::SSL::Socket do
     end
   end
 
-  it "can close connection on servers that don't shut down right" do
+  it "closes connection to server that doesn't properly terminate SSL session" do
     tcp_server = TCPServer.new(0)
     server_context, client_context = ssl_context_pair
 
-    OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
-      client_done = Channel(Nil).new
-      spawn do
-        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com", sync_close: true) do |socket|
-        end
-        client_done.send(nil)
+    client_successfully_closed_socket = Channel(Nil).new
+    spawn do
+      OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
+        server_client = server.accept
+        # require client to close the socket from its side, without the server closing it, IIS behave this way.
+        client_successfully_closed_socket.receive
+        server_client.close
       end
-      client = server.accept
-      # allow client to close the socket from its side, without the server closing it, IIS behaves this way
-      # should not hang
-      client_done.receive
-      client.close
     end
+    OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com", sync_close: true) do |socket|
+    end
+    client_successfully_closed_socket.send(nil)
   end
 end
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -35,4 +35,23 @@ describe OpenSSL::SSL::Socket do
       client.close
     end
   end
+
+  it "can close connection on servers that don't shut down right" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
+      client_done = Channel(Nil).new
+      spawn do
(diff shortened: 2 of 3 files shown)


straight-shoota:
Can we implement a misbehaving server in Crystal for the spec?

src/openssl/ssl/socket.cr
```diff
@@ -154,7 +154,8 @@ abstract class OpenSSL::SSL::Socket < IO
       loop do
         begin
           ret = LibSSL.ssl_shutdown(@ssl)
-          break if ret == 1
+          break if ret == 1 # done bidirectional
+          break if ret == 0 && @sync_close # done unidirectional, "this first successful call to SSL_shutdown() is sufficient"
```
```suggestion
          break if ret == 0 && sync_close? # done unidirectional, "this first successful call to SSL_shutdown() is sufficient"
```

rdp: OK done

rdp:
OK added a spec that repro's it, thanks!

spec/manual/https_client_spec.cr
```diff
@@ -12,4 +12,8 @@ describe "https requests" do
       break
     end
   end
+
+  it "can fetch from IIS servers that don't shutdown gracefully" do
+    HTTP::Client.post("https://login.microsoftonline.com/common/oauth2/token", form: {"hi" => "a"}).body
```
Sija: This imo shouldn't be added - we don't want to rely on 3rd party services in tests.

rdp:
I believe that's the reason it was put in the "spec/manual" directory,
basically it's unstable specs "go there".  But feel free to pull it if
desired, now that we have a working local spec it isn't as necessary.

spec/manual/https_client_spec.cr
straight-shoota: This is not necessary since we can reproduce the behaviour in a regular spec.

spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -35,4 +35,23 @@ describe OpenSSL::SSL::Socket do
...
     end
   end
+
+  it "can close connection on servers that don't shut down right" do
+    tcp_server = TCPServer.new(0)
+    server_context, client_context = ssl_context_pair
+
+    OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
+      client_done = Channel(Nil).new
+      spawn do
```
straight-shoota: Could you please put the (entire) server into a separate fiber, since its only auxiliary. The code that's supposed to be tested here is the client connection and that should execute in the main fiber. This helps understanding what this spec does.

spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -35,4 +35,23 @@ describe OpenSSL::SSL::Socket do
       client.close
     end
   end
+
+  it "can close connection on servers that don't shut down right" do
```
straight-shoota:
```suggestion
  it "closes connection to server that doesn't properly terminate SSL session" do
```

spec/std/openssl/ssl/socket_spec.cr
rdp: OK check the latest.

spec/std/openssl/ssl/socket_spec.cr
rdp: Thanks, applied manually.

spec/std/openssl/ssl/socket_spec.cr
```diff
@@ -35,4 +35,22 @@ describe OpenSSL::SSL::Socket do
...
+    client_successfully_closed_socket = Channel(Nil).new
+    spawn do
+      OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
+        server_client = server.accept
+        # require client to close the socket from its side, without the server closing it, IIS behave this way.
+        client_successfully_closed_socket.receive
+        server_client.close
+      end
+    end
+    OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com", sync_close: true) do |socket|
```
RX14:
```suggestion
    socket = TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port
    socket = OpenSSL::SSL::Socket::Client.new(socket, client_context, hostname: "example.com", sync_close: true)
    socket.close
```

rdp:
I think this is ready to merge if somebody would like to help me out with it sometime.  Then I could move onto other OpenSSL stuff.  Thanks!

Sija:
This imho would perfectly fit into `0.32.0` :)

rdp:
Thanks!

bcardiff:
FYI it seems the CI is not always happy with this new spec. Any clues?

https://travis-ci.org/crystal-lang/crystal/jobs/623864676?utm_medium=notification&utm_source=github_status

```
Unhandled exception in spawn: Error writing to socket: Broken pipe (Errno)
  from src/socket.cr:82:13 in 'unbuffered_write'
  from src/io/buffered.cr:135:14 in 'write'
  from src/openssl/bio.cr:31:7 in '->'
  from ???
  from BIO_write
  from ???
  from BIO_ctrl
  from ???
  from ???
  from ???
  from SSL_do_handshake
  from src/openssl/ssl/socket.cr:57:15 in 'initialize'
  from src/openssl/ssl/socket.cr:54:5 in 'new:sync_close'
  from src/openssl/ssl/server.cr:63:5 in 'accept'
  from spec/std/openssl/ssl/socket_spec.cr:46:25 in '->'
  from src/fiber.cr:47:34 in '->'
  from ???
```

rdp:
Yes it's #8559 but I haven't tidied it all up yet...

PR #8486, opened 2019-11-18
when running a locally compiled crystal from a distant path, it can b…

rdp:
…e confusing which .build it is referring to, so show full path instead

### DIFF
--- a/bin/crystal
+++ b/bin/crystal
@@ -142,7 +142,7 @@ export CRYSTAL_PATH=$CRYSTAL_ROOT/src:lib
 export CRYSTAL_HAS_WRAPPER=true
 
 if [ -x "$CRYSTAL_DIR/crystal" ]; then
-  __warning_msg "Using compiled compiler at \`.build/crystal'"
+  __warning_msg "Using compiled compiler at $CRYSTAL_DIR/crystal"
   exec "$CRYSTAL_DIR/crystal" "$@"
 elif ! command -v crystal > /dev/null; then
   __error_msg 'You need to have a crystal executable in your path!'
--- a/bin/crystal
+++ b/bin/crystal
@@ -142,7 +142,7 @@ export CRYSTAL_PATH=$CRYSTAL_ROOT/src:lib
 export CRYSTAL_HAS_WRAPPER=true
 
 if [ -x "$CRYSTAL_DIR/crystal" ]; then
-  __warning_msg "Using compiled compiler at $CRYSTAL_DIR/crystal"
+  __warning_msg "Using compiled compiler at ${CRYSTAL_DIR#"$PWD/"}/crystal"
   exec "$CRYSTAL_DIR/crystal" "$@"
 elif ! command -v crystal > /dev/null; then
   __error_msg 'You need to have a crystal executable in your path!'


bin/crystal
```diff
@@ -142,7 +142,7 @@ export CRYSTAL_PATH=$CRYSTAL_ROOT/src:lib
 export CRYSTAL_HAS_WRAPPER=true
 
 if [ -x "$CRYSTAL_DIR/crystal" ]; then
-  __warning_msg "Using compiled compiler at \`.build/crystal'"
+  __warning_msg "Using compiled compiler at $CRYSTAL_DIR/crystal"
```
RX14: Could we print the full path only when `pwd` doesn't start with `$CRYSTAL_ROOT`?

rdp: That did work, thank you!  Now it says:

```
$ ./bin/crystal 

$ ./dev/crystal/bin/crystal 
Using compiled compiler at dev/crystal/.build/crystal
```
and the full path if you're not in a relative directory.

RX14: @Sija why the confused react?

@rdp thoughts on adding `./`before the path?

Sija: I was confused as to why not leave the full path regardless?

asterite:
I don't understand the purpose of this PR.

When you compile the compiler it ends up in `.build/crystal`. It's always relative to the compiler's source code. I have one copy of the compiler in my machine and then use git branches.

Why do we need more info for the path? Do you have multiple copies of the compiler? And if so, how do you run them? You need to use the `bin/crystal` wrapper. If it's from another location then you pass the full path. Unless you have multiple aliases to different `bin/crystal`?

I'm confused...

j8r:
@asterite if you call the wrapper from outside the repository source, say you are in `/home/$USER` and you execute `crystal/bin/crystal`, a message will say it uses the compiler at `.build/crystal` - but that's false, it uses the compiler at `crystal/.build/crystal`.

asterite:
I understand that some falsey thing is being said. But why does it matter? The only thing that matters is that it uses the compiled compiler, not the non-compiled compiler. The path doesn't matter. And `.build/crystal` is good enough.

asterite:
To make it clearer:

> /home/$USER and you execute crystal/bin/crystal

At this point you already knew how to compiler the compiler. You probably know it goes into `.build/crystal`. And `.build/crystal` is the only place where the compiled compiler is looked for. If you execute it from another path I don't expect anyone to think "oh, suddenly a .build/crystal" compiler appeared in this directory.

Maybe we could remove the path altogether? Or just say `.build/crystal` relative to the compiler's directory?

j8r:
Sometimes I even call `.build/crystal` directly, to be sure I am using the right binary. By the way tt's needed when using `gdb`.

asterite:
> Sometimes I even call .build/crystal directly

That shouldn't work well unless you have `CRYSTAL_PATH` set up. The `bin/crystal` wrapper does that.

Could you describe your workflow and why you want the full path to `.build/crystal`? Is it confusing or is it giving you some problem? I generally ignore that message, the only interesting part of it is

rdp:
For me yeah it's confusing.  Even though I just specified the full path to `bin/crystal`:

```
$ ./dev/crystal/bin/crystal 
```
Which `.build` directory is it referring to?  The relative `.build` directory?  One located in my current working directory?  One somewhere else?  Those familiar with the wrapper know what's going on but those of us that don't...aren't sure if we are supposed to trust it to be relative, even though it worked fine when in the crystal directory.

Even if it just said 

`Using compiled compiler at relative path ``.build/crystal'`

That would clarify it (as to where that .build directory is), but full path seems a better option?

Cheers!

straight-shoota:
@asterite I don't think it hurts to clarify this. I don't see any downside to this change, so if it helps some users, why not?

rdp:
Thank you!

Issue #8440, opened 2019-11-05
Inconsistent `Enumerable#minmax` behavior when an element is not comparable


kwwagner1693:
# Summary
The current (0.31.1) implementation of `Enumerable#minmax` does not define any behavior when the `Enumerable` contains an element which is not comparable with other elements. Consequently, some rather inconsistent results can occur depending on the ordering of elements in the `Enumerable`. Since the ordering of an `Enumerable` should not have any effect on the minimum or maximum element, this inconsistency appears erroneous, and I believe it should be fixed. While I have not verified, this inconsistency is likely an issue with all variants of `#min` and `#max` as well. 

# Example
I encountered this issue specifically in the context of trying to extract a minimum or maximum value from an `Enumerable(Float64)` which contained `Float64::NAN` values. It would seem that whenever the first element of an `Enumerable` is `NaN`, the result of `#minmax` will always be `{NaN, NaN}` regardless of any other non-`NaN` elements. In any other case, the result of `#minmax` will be as expected, with all `NaN` elements ignored.

```
nan_first = [Float64::NAN, 1.0, 2.0, Float64::NAN]
puts nan_first.minmax # => {NaN, NaN}

number_first = [1.0, Float64::NAN, 2.0, Float64::NAN]
puts number_first.minmax # => {1.0, 2.0}
``` 

# Possible Resolutions
In looking into this issue, I noticed that `Enumerable#minmax` only ever uses `<` and `>` between elements and never seems to have been updated to use the partial comparability of `<=>`. I think it would be best to follow the precedent of https://github.com/crystal-lang/crystal/pull/6611 and rework `Enumerable#min`, `Enumerable#max`, and `Enumerable#minmax` to use `<=>` and raise when an element is not comparable.

asterite:
Good catch! I think in this case a runtime exception should happen, similar to what happens when you sort an array with NaN.

TedTran2019:
Hey-- I'm a bit new to coding and new to open source as well. I have knowledge of C and Ruby, but I've never used the Crystal language or navigated large code bases before. Would it be okay for me to try at my hand at this, or would it be better to leave it to someone who's experienced and can solve the issue within a reasonable amount of time? (If it is a valid issue). I can probably spend about 5 - 10~ hours a week going at it and I'd really like to learn and contribute. Thanks for the feedback! (Do I just wait to be assigned or whatever, or do I just go for it then make a pull request?)

bararchy:
[...]

PR #8490, opened 2019-11-19
Issue 8440: Inconsistent `Enumerable#minmax`

TedTran2019:
Changed all forms of min, max, and minmax to raise an ArgumentError when NAN is one of the arguments. Also added testing for minmax. 

> asterite: Good catch! I think in this case a runtime exception should happen, similar to what happens when you sort an array with NaN.

I only included a test case for minmax, but I changed all the _internal functions which affects every variant of min/max/minmax. I could add more test cases, or solve the problem by using the <=> operator instead of raising an error. I also couldn't find "runtime exception" among exceptions, so I just used the plain old ArgumentError instead. 

Thank you for your patience.

[Issue being referenced](https://github.com/crystal-lang/crystal/issues/8440)

EDIT: Removed frivolous information irrelevant to issue

### DIFF
--- a/src/comparable.cr
+++ b/src/comparable.cr
@@ -64,6 +64,11 @@ module Comparable(T)
     cmp ? cmp >= 0 : false
   end
 
+  # Like <=>, but raises when nil is returned.
+  def compare_and_raise(other : T)
+    (self <=> other) || raise ArgumentError.new("Comparison of #{self} and #{other} failed")
+  end
+
   # The comparison operator. Returns `0` if the two objects are equal,
   # a negative number if this object is considered less than *other*,
   # a positive number if this object is considered greter than *other*,
--- a/src/comparable.cr
+++ b/src/comparable.cr
@@ -64,11 +64,6 @@ module Comparable(T)
     cmp ? cmp >= 0 : false
   end
 
-  # Like <=>, but raises when nil is returned.
-  def compare_or_raise(other : T)
-    (self <=> other) || raise ArgumentError.new("Comparison of #{self} and #{other} failed")
-  end
-
   # The comparison operator. Returns `0` if the two objects are equal,
   # a negative number if this object is considered less than *other*,
   # a positive number if this object is considered greter than *other*,
--- a/src/comparable.cr
+++ b/src/comparable.cr
@@ -65,7 +65,7 @@ module Comparable(T)
   end
 
   # Like <=>, but raises when nil is returned.
-  def compare_and_raise(other : T)
+  def compare_or_raise(other : T)
     (self <=> other) || raise ArgumentError.new("Comparison of #{self} and #{other} failed")
   end
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,7 +46,7 @@ jobs:
 
   test_darwin:
     macos:
-      xcode: "9.0"
+      xcode: "11.1.0"
     environment:
       <<: *env
       TRAVIS_OS_NAME: osx
@@ -57,6 +57,7 @@ jobs:
             - brew-cache-v1
       - checkout
       - run: bin/ci prepare_system
+      - run: echo 'export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/opt/openssl/lib/pkgconfig"' >> $BASH_ENV
       - run: echo 'export CURRENT_TAG="$CIRCLE_TAG"' >> $BASH_ENV
       - run: bin/ci prepare_build
       - run: bin/ci build
@@ -262,7 +263,7 @@ jobs:
 
   dist_darwin:
     macos:
-      xcode: "9.0"
+      xcode: "11.1.0"
     shell: /bin/bash --login -eo pipefail
     steps:
       - restore_cache:
--- a/bin/ci
+++ b/bin/ci
@@ -111,8 +111,7 @@ prepare_build() {
   on_osx curl -L https://github.com/crystal-lang/crystal/releases/download/0.31.1/crystal-0.31.1-1-darwin-x86_64.tar.gz -o ~/crystal.tar.gz
   on_osx 'pushd ~;gunzip -c ~/crystal.tar.gz | tar xopf -;mv crystal-0.31.1-1 crystal;popd'
 
-  on_osx brew install llvm@8 gmp libevent pcre pkg-config
-  on_osx brew upgrade libyaml
+  on_osx brew install llvm@8 gmp libevent pcre openssl pkg-config
   on_osx brew link --force llvm@8
   # Note: brew link --force might show:
   #   Warning: Refusing to link macOS-provided software: llvm
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -687,7 +687,7 @@ describe "Enumerable" do
 
     it "raises if not comparable" do
       expect_raises ArgumentError do
-        [-1.0, Float64::NAN, -3.0].max_of { |x| -x }
+        [-1, Float64::NAN, -3.0].max_of { |x| -x }
       end
     end
   end
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -730,6 +730,12 @@ describe "Enumerable" do
         ([] of Int32).minmax
       end
     end
+
+    it "raises if NAN is one of the elements" do
+      expect_raises ArgumentError do
+        [Float64::NAN, 1.0, 2.0].minmax
+      end
+    end
   end
 
   describe "minmax?" do
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -730,12 +730,6 @@ describe "Enumerable" do
         ([] of Int32).minmax
       end
     end
-
-    it "raises if NAN is one of the elements" do
-      expect_raises ArgumentError do
-        [Float64::NAN, 1.0, 2.0].minmax
-      end
-    end
   end
 
   describe "minmax?" do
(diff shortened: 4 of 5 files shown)


straight-shoota:
This should not be solved by hard-coding a special case into the algorithm. Instead, it should use `<=>` directly which has a generic case for values that are not comparable. That way the same behaviour would work not only for `Float::NAN` but also for other partially-comparable types.

straight-shoota:
About the semantics: An alternative to raising `ArgumentError` would be to disregard values that are not comparable. That could at least make some sense in the case of `Float::NAN`. But it can't work for generalized partially-comparable types, which might have more complex rules than just a single value that doesn't compare with any other.

So there is probably no alternative to enforcing full comparability on the values, i.e. raising `ArgumentError`.

TedTran2019:
> This should not be solved by hard-coding a special case into the algorithm. Instead, it should use `<=>` directly which has a generic case for values that are not comparable. That way the same behaviour would work not only for `Float::NAN` but also for other partially-comparable types.

Okay, thank you. Do I Just change it to use the <=> operator like sort then commit again? Next time around, do I also add more test cases or is just 1 for minmax enough? 

I was checking other values besides NAN, and I noticed it would just raise an overload error typically. e.g comparing char with number, putting in nil, etc.

straight-shoota:
> Do I Just change it to use the <=> operator like sort then commit again?  

Yes. You'll need to compare the return value from `<=>` with 0 (and raise if it's nil).

> Next time around, do I also add more test cases or is just 1 for minmax enough?

A few more tests would be great.

> I was checking other values besides NAN, and I noticed it would just raise an overload error typically.

Yes, in stdlib there aren't many partially comparables. Float's NAN might actually be the only one. But it's a general API and can be used for implementing any kind of partially comparable types.

TedTran2019:
Thanks you for your patience. 

I changed it to: 

`value = yield elem <=> max`
`raise ArgumentError.new("Comparison of #{elem} and #{max} failed") if value.nil?`
`if i == 0 || value == 1`

Would this be suitable? I'll be writing test cases now!

straight-shoota:
The return value of `<=>` can be 0 or any positive or negative integer, so you need to check for `value > 0` instead of `value == 1`.

And you'll probably need to add parenthesis: `(yield elem) <=> max`.

TedTran2019:
> The return value of `<=>` can be 0 or any positive or negative integer, so you need to check for `value > 0` instead of `value == 1`.
> 
> And you'll probably need to add parenthesis: `(yield elem) <=> max`.

Okay, thank you!  Ohh-- I was thinking it was like Ruby's <=> operator and only returned -1, 0, and 1, my bad. I started out with parenthesis then actually removed them haha-- but in hindsight that was a horrible idea. Since now I don't really know if `yield elem` or `elem <=> max` is happening first

TedTran2019:
I'm not particularly sure if you guys have a 80 line limit or something, but the line `raise ArgumentError.new("Comparison of #{ele_val} and #{max} failed") if value.nil?` goes a bit over.  I also think ele_val and value can be renamed for more clarity, but I kept it similar to the original code. 

Another heads up, I'm not sure if this behavior was present in the original max/min/minmax (I'll have to check), but using any of these functions and its derivatives on an array with Int32 and NANs results in a segmentation fault. 

`[-1, Float64::NAN, -3].max_of { |x| -x }`
`Invalid memory access (signal 11) at address 0x0...`

straight-shoota:
> Is it because I'm using the comparison operator <=> on the uninitialized U?

Exactly. This code must not run if `i == 0` which was previously ensured by the logical OR operator.

TedTran2019:
> > Is it because I'm using the comparison operator <=> on the uninitialized U?
> 
> Exactly. This code must not run if `i == 0` which was previously ensured by the logical OR operator.

Done! It's pretty messy, though. Hopefully that all works out.

straight-shoota:
Yeah, that's too convoluted. I'd suggest to move the common logic into a private helper method.
```cr
each_with_index do |elem, i|
  value = yield elem
  if i == 0 || compare_min_max(value, max) > 0
    max = value
    obj = elem
  end
end

# ...
private def compare_min_max(value, memo)
  value <=> max || raise ArgumentError.new("Comparison of #{value} and #{max} failed")
end
```

TedTran2019:
> Yeah, that's too convoluted. I'd suggest to move the common logic into a private helper method.
> 
> ```crystal
> each_with_index do |elem, i|
>   value = yield elem
>   if i == 0 || compare_or_raise(value, max) > 0
>     max = value
>     obj = elem
>   end
> end
> 
> # ...
> private def compare_or_raise(value, memo)
>   value <=> memo || raise ArgumentError.new("Comparison of #{value} and #{memo} failed")
> end
> ```

Alright, will do! That actually looks a billion times better, haha. I guess once it works, refactoring is really important too. I wonder if the circleci test is failing because I put [1, Float64::NAN, 2.0].max or if it's because of some other error

asterite:
I wonder if we can add a method to comparable that is like the spaceship operator but raises argument error on nil.

TedTran2019:
> I wonder if we can add a method to comparable that is like the spaceship operator but raises argument error on nil.

What should the method be called?

straight-shoota:
Maybe `<==>`? Or `<=>!`. It's going to by kind of cryptic. Or we chose an actual name such as `compare_or_raise`.

TedTran2019:
> Maybe `<==>`? Or `<=>!`. It's going to by kind of cryptic. Or we chose an actual name such as `compare_or_raise`.

I actually like <=>! a lot, since that's how it's generally done in Rails, I think. 

So the regular version would return nil upon failing and the ! version would raise an error, that sounds pretty reasonable and intuitive to me.

straight-shoota:
> Not to mention, to use it, somebody would have to do
`etc compare_and_raise etc2`

No, that can't work because it's not an operator. It would just be a normal call: `a.compare_or_raise(b)`.

TedTran2019:
> > Not to mention, to use it, somebody would have to do
> > `etc compare_and_raise etc2`
> 
> No, that can't work because it's not an operator. It would just be a normal call: `a.compare_or_raise(b)`.

Ohhhh. Normal calls are lame as hell compared to the syntactic sugar of operator methods, haha. It makes since that you can't do that style of call, though-- come to think of it. With the implicit parentheses rule, it looks like compare_or_raise b is being passed into a, and the parser probably can't read that at all.

src/comparable.cr
```diff
@@ -64,6 +64,11 @@ module Comparable(T)
     cmp ? cmp >= 0 : false
   end
 
+  # Like <=>, but raises when nil is returned.
+  def compare_and_raise(other : T)
```
Sija:
```suggestion
  def compare_or_raise(other : T)
```

Sija: btw, it looks really weird as an `Comparable` method...

straight-shoota:
> To make <==> readable as a method, wouldn't the lexer/parser have to be changed?

Exactly. I'm impressed that you digged all that up in the compiler source :+1:

But adding an additional method to `Comparable` is out of scope here. Let's focus on the fix for `Enumerable#minmax` using a local private method (4cec19515e81fba7221dceea65d70c622c68e7ba).

The idea for `Comparable` should be discussed in a new issue and we can later adapt the code for `Enumerable#minmax`.

asterite:
Yeah, sorry for bringing that idea. It was just a thought, not something we need to do now or something that we even will end up doing.

TedTran2019:
> Yeah, sorry for bringing that idea. It was just a thought, not something we need to do now or something that we even will end up doing.

It made a lot of sense and is a good idea, though. A private comparison method in enumerable.cr isn't nearly as clean as having the method in comparable.cr. But I guess that should be in a new issue-- which makes a whole lot of sense too. 

Man, that was really interesting ngl. I've only been studying alone, so I've never really had to face any merge conflicts. After a lot of failed attempts at reverting, I just ended up switching to the old commit, copying the enumerable file, switching back to the current head, then pasting in the old file. I'm pretty sure that's not how it's done and there's a specific command for grabbing specific commits (git cherry-pick?), but that was lots of fun regardless. 



> > To make <==> readable as a method, wouldn't the lexer/parser have to be changed?
> 
> Exactly. I'm impressed that you digged all that up in the compiler source 
> 
> But adding an additional method to `Comparable` is out of scope here. Let's focus on the fix for `Enumerable#minmax` using a local private method ([4cec195](https://github.com/crystal-lang/crystal/commit/4cec19515e81fba7221dceea65d70c622c68e7ba)).
> 
> The idea for `Comparable` should be discussed in a new issue and we can later adapt the code for `Enumerable#minmax`.

Tried to revert back to the old commit, but pretty sure I didn't do it properly. Hope that all works out! If you guys end up implementing a new operator in the future, I'd love to see the process or try my hand at it myself if possible. I think I can look through crystal-lang's old commits when you guys added new operators before, then see what was changed to make the operators work and why.

straight-shoota:
The commit history doesn't matter too much because it's a small change that's going to be squashed anyway. But for future reference, you might want to get acquainted with `git reset` ;)

asterite:
Also, don't worry, git is pretty complex to learn. 

> I just ended up switching to the old commit, copying the enumerable file, switching back to the current head, then pasting in the old file

Hehe, that's why I end up doing from time to time when I stop understanding what happened with git.

TedTran2019:
> The commit history doesn't matter too much because it's a small change that's going to be squashed anyway. But for future reference, you might want to get acquainted with `git reset` ;)

Will do! 



> Thank you!

Thank you both for your patience and time. I'm sure either of you two could have done it in like 5 minutes, but thanks for taking the time out of your days to guide me in the process. I really appreciate it! Ya'll are amazing

straight-shoota:
Getting the implementation in 5 minutes would certainly be possible, but only if the final intention had be known in advance. But reaching that is the real effort behind most PRs. It's a process requiring team effort and can only be reached through collaboratively sharing and discussing ideas.

PR #8491, opened 2019-11-19
Improve documentation for pseudo methods

straight-shoota:
This is a followup on #8327 adding a few minor text improvements as suggested by @RX14 in
https://github.com/crystal-lang/crystal/pull/8327#pullrequestreview-316950603.

### DIFF
--- a/src/docs_pseudo_methods.cr
+++ b/src/docs_pseudo_methods.cr
@@ -50,7 +50,7 @@ end
 
 # Returns the instance size of the given class as number of bytes.
 #
-# *type* must be a constant, it cannot be evaluated at runtime.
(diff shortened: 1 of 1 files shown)


src/docs_pseudo_methods.cr
```diff
@@ -25,11 +25,13 @@ end
 
 # Returns the size of the given type as number of bytes.
 #
-# *type* must be a constant, it cannot be evaluated at runtime.
+# *type* must be a constant or `typeof()` expression. It cannot be evaluated
+# at runtime.
```
RX14: I think this should be changed in a few other places on the page too.

PR #8494, opened 2019-11-20
Fix overflow exception in BitArray

jimbowhales:
(1 << count) - 1 underflows when count = 31 and type is Int32.

### DIFF
--- a/src/bit_array.cr
+++ b/src/bit_array.cr
@@ -140,7 +140,7 @@ struct BitArray
       bits = @bits[0]
 
       bits >>= start
-      bits &= (1 << count) - 1
+      bits &= (1 << count) &- 1
 
       BitArray.new(count).tap { |ba| ba.@bits[0] = bits }
     elsif size <= 64
@@ -148,10 +148,10 @@ struct BitArray
       bits = @bits.as(UInt64*)[0]
 
       bits >>= start
-      bits &= (1 << count) - 1
+      bits &= (1 << count) &- 1
 
       if count <= 32
-        BitArray.new(count).tap { |ba| ba.@bits[0] = bits.to_u32 }
+        BitArray.new(count).tap { |ba| ba.@bits[0] = bits.to_u32! }
       else
         BitArray.new(count).tap { |ba| ba.@bits.as(UInt64*)[0] = bits }
       end
@@ -169,7 +169,7 @@ struct BitArray
         bits = @bits[start_bit_index + i + 1]
 
         high_bits = bits
-        high_bits &= (1 << start_sub_index) - 1
+        high_bits &= (1 << start_sub_index) &- 1
         high_bits <<= 32 - start_sub_index
 
         ba.@bits[i] = low_bits | high_bits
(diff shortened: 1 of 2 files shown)


straight-shoota:
Could you please also add a test case that fails without the change?

straight-shoota:
I'd prefer specific test that not only ensure that the call does not fail but also returns the correct value.

jimbowhales:
I did come across another bug writing the test for the first. Fixed too.

src/bit_array.cr
```diff
@@ -140,18 +140,18 @@ struct BitArray
...
       # Original fits in int64, we can use bitshifts
       bits = @bits.as(UInt64*)[0]
 
       bits >>= start
-      bits &= (1 << count) - 1
+      bits &= (1 << count) &- 1
 
       if count <= 32
-        BitArray.new(count).tap { |ba| ba.@bits[0] = bits.to_u32 }
+        BitArray.new(count).tap { |ba| ba.@bits[0] = bits.to_u32! }
```
RX14: I don't see when this can overflow if `bits &= (1 << count) &- 1` is working correctly. However, it is not.

`(1 << count)` is simply zero when `count >= 32`, because `1` is an int32. This doesn't matter though, since we don't care about the MSB bits being zero. So I think the `bits &= (1 << count) &- 1` line can be removed and the `to_u32!` can stay. And I think the masking in the `size <=32` case can be removed too. I'd like a comment left about leaving the trailing upper though.

Usages of `(1 << count)` in the rest of the file must be audited though. I think they're all fine, but i'd like a second pair of eyes.

src/bit_array.cr
```diff
@@ -32,10 +32,17 @@ struct BitArray
 
   def ==(other : BitArray)
     return false if size != other.size
-    # NOTE: If BitArray implements resizing, there may be more than 1 binary
-    # representation and their hashes for equivalent BitArrays after a downsize as the
-    # discarded bits may not have been zeroed.
-    return LibC.memcmp(@bits, other.@bits, malloc_size) == 0
+    return true if size == 0
```
asterite: Could you explain why this entire change is needed? I don't understand why the last bit needs special handling. And if it does, there should be a comment in the code explaining why.

RX14: It needs special handling since unused bits at the end of the array are not guaranteed to be in a defined state, so they have to be masked out (i.e., set to zero) before comparing.

asterite: No, that's not correct. `Pointer.malloc` makes sure everything is zero. The comparison is broken because `LibC.memcmp` expects the number of bytes but it uses `malloc_size` which is in terms of UInt32. Changing it to use `malloc_size * sizeof(UInt32)` makes it work.

asterite: Oh, I see. The method `[]` will copy bits from an existing bit array and just copy them over without zeroing the non-interesting part, that's why `memcmp` doesn't work.

I think we should fix `[]` to guarantee the invariant that unused bits are always zero.

RX14: I don't really care which method is used, this PR is certainly fine.

HertzDevil:
Resolved by #10809. (This PR assumes `BitArray`s can have unused bits set which is no longer true.)

PR #8498, opened 2019-11-21
Improve error message for getaddrinfo failure

rdp:
…s may be connected to DNS #8262
See discussion in #8262 
I wasn't sure if the hint were appropriate or not but maybe useful to someone? :)

### DIFF
--- a/spec/std/socket/tcp_server_spec.cr
+++ b/spec/std/socket/tcp_server_spec.cr
@@ -83,13 +83,13 @@ describe TCPServer do
       end
 
       it "raises when host doesn't exist" do
-        expect_raises(Socket::Error, "No address found for doesnotexist.example.org.:12345 over TCP") do
+        expect_raises(Socket::Error, "Hostname lookup for doesnotexist.example.org. failed: No address found") do
           TCPServer.new("doesnotexist.example.org.", 12345)
         end
       end
 
       it "raises (rather than segfault on darwin) when host doesn't exist and port is 0" do
-        expect_raises(Socket::Error, /No address found for doesnotexist.example.org.:00? over TCP/) do
+        expect_raises(Socket::Error, /Hostname lookup for doesnotexist.example.org. failed: No address found/) do
           TCPServer.new("doesnotexist.example.org.", 0)
         end
       end
--- a/spec/std/socket/tcp_server_spec.cr
+++ b/spec/std/socket/tcp_server_spec.cr
@@ -89,7 +89,7 @@ describe TCPServer do
       end
 
       it "raises (rather than segfault on darwin) when host doesn't exist and port is 0" do
-        expect_raises(Socket::Error, /Hostname lookup for doesnotexist.example.org. failed: No address found/) do
+        expect_raises(Socket::Error, "Hostname lookup for doesnotexist.example.org. failed: No address found") do
           TCPServer.new("doesnotexist.example.org.", 0)
         end
       end
--- a/src/socket/addrinfo.cr
+++ b/src/socket/addrinfo.cr
@@ -82,12 +82,12 @@ class Socket
     class Error < Socket::Error
       getter error_code : Int32
 
-      def self.new(error_code, domain, service, protocol)
-        new error_code, String.new(LibC.gai_strerror(error_code)), domain, service, protocol
+      def self.new(error_code, domain)
+        new error_code, String.new(LibC.gai_strerror(error_code)), domain
       end
 
-      def initialize(@error_code, message, domain, service, protocol)
-        super("getaddrinfo: " + message + " for #{domain}:#{service} over #{protocol} when attempting to lookup internet address for host.  Hint: check hostname, check resolution system (which includes DNS)")
+      def initialize(@error_code, message, domain)
+        super("Hostname lookup for #{domain} failed: #{message}")
       end
     end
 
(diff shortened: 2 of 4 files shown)


straight-shoota:
Runtime error messages should be concise and describe the problem only as much as necessary. This seems to extensive.
How to deal with such an error is a different thing. There may be many reasons why address resolution fails. It shouldn't be Crystals job to provide general troubleshooting advice in an error message.

The only useful improvement IMO is to add context information to all errors, not just for `EAI_NONAME`. When address resolution fails, it's essential to know *which* address was tried.

straight-shoota:
BTW I tried this in several programming languages and most simply just pass the error message from getaddrinfo with no context information.

The only good message I found was in Golang:

> lookup badhostname on 192.168.1.1:53: no such host

It's great to show which DNS server was used.

rdp:
Whoa, golang doing their own DNS resolution, nice.
How about this instead: "getaddrinfo: <msg> for badhostname:80 over IP when attempting domain name resolution" ?

straight-shoota:
I'd go with a very concise message: `Hostname lookup for badhostname failed: #{errno_message}`.

* We can skip `getaddrinfo: ` at the beginning. This API call is only useful when the errno message is directly passed. Here, we have a customized message anyway.
* `protocol` and `socket` are usually not very helpful. I'd omit them unless the error is `EAI_SOCKTYPE` or `EAI_SERVICE`, respectively. Similarly, the address family could be shown for `EAI_ADDRFAMILY`. In all other cases, lookup failed generally and these details don't matter.

rdp:
OK I didn't see EAI_ADDRFAMILY defined so didn't include that but did some for the rest, please review, thank you!

src/socket/addrinfo.cr
```diff
@@ -121,9 +121,11 @@ class Socket
       when 0
         # success
       when LibC::EAI_NONAME
-        raise Error.new(LibC::EAI_NONAME, "No address found for #{domain}:#{service} over #{protocol}")
+        raise Error.new(LibC::EAI_NONAME, "No address found", domain)
```
Sija:
```suggestion
        raise Error.new(ret, "No address found", domain)
```

src/socket/addrinfo.cr
```diff
@@ -121,9 +121,11 @@ class Socket
       when 0
         # success
       when LibC::EAI_NONAME
-        raise Error.new(LibC::EAI_NONAME, "No address found for #{domain}:#{service} over #{protocol}")
+        raise Error.new(ret, "No address found", domain)
+      when LibC::EAI_SOCKTYPE, LibC::EAI_SERVICE
```
Sija: I'd split those into the different branches:

```cr
when LibC::EAI_SOCKTYPE
  raise Error.new(ret, "The requested socket type is not supported", domain)
when LibC::EAI_SERVICE
  raise Error.new(ret, "The requested service is not available for the requested socket type", domain)
```

rdp: OK I've split them (and added in the relevant information which I forgot before, heh).  Ready for review, thanks!

src/socket/addrinfo.cr
```diff
@@ -82,12 +82,12 @@ class Socket
     class Error < Socket::Error
       getter error_code : Int32
 
-      def self.new(error_code)
-        new error_code, "getaddrinfo: #{String.new(LibC.gai_strerror(error_code))}"
+      def self.new(error_code, domain)
```
straight-shoota:
```suggestion
      def self.new(error_code, hostname)
```
We're looking up hostnames, not domains.

rdp: Yeah the getaddrinfo man page seems to refer to it as a "node"  or something like that.  Referring to it as a "domain" is scattered throughout the entire file and comments (plus domain can be an IP address or a hostname), OK to leave as domain for now?  Cheers!

straight-shoota: Yeah, fine. They should all be changed to hostname or address.

rdp:
Looks like this breaks some other specs, I'll check those out.

rdp:
OK ready for review, thank you.

spec/std/socket/tcp_server_spec.cr
```diff
@@ -83,13 +83,13 @@ describe TCPServer do
...
       it "raises when host doesn't exist" do
-        expect_raises(Socket::Error, "No address found for doesnotexist.example.org.:12345 over TCP") do
+        expect_raises(Socket::Error, "Hostname lookup for doesnotexist.example.org. failed: No address found") do
           TCPServer.new("doesnotexist.example.org.", 12345)
         end
       end
 
       it "raises (rather than segfault on darwin) when host doesn't exist and port is 0" do
-        expect_raises(Socket::Error, /No address found for doesnotexist.example.org.:00? over TCP/) do
+        expect_raises(Socket::Error, /Hostname lookup for doesnotexist.example.org. failed: No address found/) do
```
straight-shoota: While you're at it, could yo replace the regular expression with a plain string? Ditto below.

rdp: OK I think I got it :)

straight-shoota:
Thank you @rdp ❤️

rdp:
Thanks!

PR #8500, opened 2019-11-21
have a better error message when a filename is misspelled

rdp:
It's a bit confusing to enter a filename and have crystal say "invalid command"

```
$ ./bin/crystal filename.cr
Error: unknown command: filename.cr
```
So attempt clarify.
```
$ ./bin/crystal filename.cr
Error: unknown command (and file not found): filename.cr
```

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -102,7 +102,11 @@ class Crystal::Command
     when File.file?(command)
       run_command(single_file: true)
     else
-      error "unknown command (and file not found): #{command}"
+      if command =~ /.+#{File::SEPARATOR}.+/ || command.ends_with?(".cr")
+        error "file not found: #{command}"
+      else
+        error "unknown command: #{command}"
+      end
     end
   rescue ex : Crystal::LocationlessException
     error ex.message
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -102,7 +102,7 @@ class Crystal::Command
     when File.file?(command)
       run_command(single_file: true)
     else
-      error "unknown command: #{command}"
+      error "unknown command (and file not found): #{command}"
     end
   rescue ex : Crystal::LocationlessException
     error ex.message
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -102,8 +102,8 @@ class Crystal::Command
     when File.file?(command)
       run_command(single_file: true)
     else
-      if command =~ /.+#{File::SEPARATOR}.+/ || command.ends_with?(".cr")
-        error "file not found: #{command}"
+      if command.ends_with?(".cr")
+        error "file not found: '#{command}'"
       else
         error "unknown command: #{command}"
       end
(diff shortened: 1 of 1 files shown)


yxhuvud:
I'm not certain the proposed change improves things, but it might make sense to at least tell people how to get the help.

rdp:
Yeah I can see that. Maybe this?
`Error: unknown command/filename: filename.cr`

rdp:
In my perfect world it would also "look past" a file if it looked like it was an already compiled executable, like if I ran (on accident):

$ crystal my_program
when the file "my_program" exists but is a binary executable and "my_program.cr" also exists, it would fail over to "my_program.cr" (durn pesky bash completion) but that's a subject for a different PR I suppose :)

asterite:
I think, if the "command" ends with ". cr" it's safe to say "file not found" without mentioning "command". That is, with a cr extension the user most likely wanted to compile and run a file but it doesn't exist.

straight-shoota:
Another indicator would be a path separator in the command. But checking for the extension should be enough.

rdp:
Good ideas, I've attempted to implement them, thank you.

src/compiler/crystal/command.cr
```diff
@@ -102,7 +102,11 @@ class Crystal::Command
     when File.file?(command)
       run_command(single_file: true)
     else
-      error "unknown command: #{command}"
+      if command =~ /.+#{File::SEPARATOR}.+/ || command.ends_with?(".cr")
```
straight-shoota: I know I suggested it, but it might be better to leave the separator check out. This implementation may behave unexpected. On windows for example, both `/` and `\` work as separators, so you need'd need to check for either of them.

In any case, the `ends_with?` call should come before the regex because it's faster and should already match in most cases.

asterite: Also, `command.includes?(File::SEPARATOR)` is simpler and faster. Why do we need a regex?

src/compiler/crystal/command.cr
```diff
@@ -102,7 +102,11 @@ class Crystal::Command
...
+        error "file not found: #{command}"
```
asterite: Please use the same error message you get when you do `crystal run nonexistent.cr`:

```crystal
error "File #{command} does not exist"
```

bew: I'd also suggest adding (single-)quotes around the filename

src/compiler/crystal/command.cr
RX14: Just `command.ends_with?(".cr")`, please.

src/compiler/crystal/command.cr
```diff
@@ -102,7 +102,11 @@ class Crystal::Command
     when File.file?(command)
       run_command(single_file: true)
     else
-      error "unknown command: #{command}"
+      if command.ends_with?(".cr")
+        error "file not found: '#{command}'"
```
asterite: Can we use the same error message as below? (File '...' does not exist). Also, I think all error messages here shouldn't be capitalized (so the other one must be changed).

PR #8503, opened 2019-11-21
HTTP::Client: don't turn body_io into nil if content-length is zero

### LINKED ISSUE #8461: error reading body_io of a zero length http client response
Here's the example:

```
$ crystal -v
Crystal 0.32.0-dev [d1c1f5a9e] (2019-11-11)

LLVM: 8.0.0

$ cat bad2.cr 
require "http/client"

HTTP::Client.get("http://bing.com") do |response|
  puts response.status_code  # => 301
  response.body_io
end

$ crystal bad2.cr 
301
Unhandled exception: HTTP::Client::Response#body_io cannot be nil (NilAssertionError)
  from dev/crystal/src/http/client/response.cr:10:3 in 'body_io'
  from bad2.cr:5:8 in '__crystal_main'
  from dev/crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from dev/crystal/src/crystal/main.cr:86:7 in 'main'
  from dev/crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```
would have expected the empty IO or something, so that the examples would still work?

asterite:
Fixes #8461

### DIFF
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -38,10 +38,7 @@ module HTTP
         if body_type.prohibited?
           body = nil
         elsif content_length = content_length(headers)
-          if content_length >= 0
-            # Don't create IO for Content-Length == 0
-            body = FixedLengthContent.new(io, content_length)
-          end
+          body = FixedLengthContent.new(io, content_length)
         elsif headers["Transfer-Encoding"]? == "chunked"
           body = ChunkedContent.new(io)
         elsif body_type.mandatory?
--- a/src/http/common.cr
+++ b/src/http/common.cr
@@ -38,7 +38,7 @@ module HTTP
         if body_type.prohibited?
           body = nil
         elsif content_length = content_length(headers)
-          if content_length != 0
+          if content_length >= 0
             # Don't create IO for Content-Length == 0
             body = FixedLengthContent.new(io, content_length)
           end
--- a/spec/std/http/client/response_spec.cr
+++ b/spec/std/http/client/response_spec.cr
@@ -123,6 +123,12 @@ class HTTP::Client
       response.body.should eq("")
     end
 
+    it "parses response without body but Content-Length == 0, block form (#8461)" do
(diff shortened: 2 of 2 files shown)


src/http/common.cr
```diff
@@ -38,7 +38,7 @@ module HTTP
         if body_type.prohibited?
           body = nil
         elsif content_length = content_length(headers)
-          if content_length != 0
+          if content_length >= 0
```
straight-shoota: The type is `UInt64`, so this condition will always be true.

yxhuvud: Regardless of the type, what would it even mean to have a negative length?

src/http/common.cr
```diff
@@ -38,7 +38,7 @@ module HTTP
...
             # Don't create IO for Content-Length == 0
```
yxhuvud: This comment needs touchup too.

src/http/common.cr
asterite: the `content_length` method seems to be calling `to_u64` on the string, so a negative length will raise overflow, I think.

I guess negative content length doesn't make sense, I was just checking because I don't know how to handle negative values. But since I now know it's handled (kind of badly) in the `content_length` method, it's probably good now.

src/http/common.cr
```diff
@@ -38,10 +38,7 @@ module HTTP
         if body_type.prohibited?
           body = nil
```
yxhuvud: Would it make sense to start the `when EndOfRequest` block as 
```
body = 
  if body_type.prohibited?
 [...]
```

RX14: No, it'd make the diff larger for no practical benefit.

It's style, but many people prefer to assign in each if branch instead of use if as the right hand side of an expression.

asterite:
Cool! I can't merge because of the requested changes, but if someone else can, please do.

PR #8505, opened 2019-11-21
Rewind Dir before iterating.

### LINKED ISSUE #8504: [RFC] Automatic Dir rewind
`Dir#methods` give unexpected results when called more than once.  The first call gives correct results.  Subsequent calls without using `#rewind` return empty `Array`'s.

The low level `opendir` API has this behavior. Should the high level Crystal API retain it?  Is there a use case for having an empty array returned when attempting to use `Dir#Iterable`?

```
dir = Dir.new("spec")

p dir.entries
p dir.entries
```

Output:
```
# First call.
[".", "..", "std", "compiler_spec.cr", "manual", "support", "spec_helper.cr", ...]
# Second call.
[]
```

didactic-drunk:
Resolves #8504.

### DIFF
--- a/src/dir.cr
+++ b/src/dir.cr
@@ -54,11 +54,13 @@ class Dir
   # Got config.h
   # ```
   def each : Nil
+    rewind
     while entry = read
       yield entry
     end
   end
 
+  # Calls #rewind before reading the first record.
   def each
     EntryIterator.new(self)
   end
@@ -90,11 +92,13 @@ class Dir
   # ```
   def each_child : Nil
     excluded = {".", ".."}
+    rewind
     while entry = read
       yield entry unless excluded.includes?(entry)
     end
   end
 
+  # Calls #rewind before reading the first record.
   def each_child
     ChildIterator.new(self)
   end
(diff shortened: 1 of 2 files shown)


src/dir.cr
```diff
@@ -282,6 +284,7 @@ class Dir
     include Iterator(String)
 
     def initialize(@dir : Dir)
+      @dir.rewind
```
asterite: I think we should rewind in the first call to `next`, it's a bit weird just that creating the iterator rewinds the dir. But for that we need more state in the iterator, and turn it into a class. It's probably fine.

jan-zajic:
@didactic-drunk i think you meant #8504 and not #8405.

asterite:
We should probably also document this behavior.

src/dir.cr
didactic-drunk: As you wish.

src/dir.cr
```diff
@@ -90,11 +92,13 @@ class Dir
   # ```
   def each_child : Nil
     excluded = {".", ".."}
+    rewind
     while entry = read
```
asterite: We can probably use `each` here

src/dir.cr
```diff
@@ -54,11 +54,13 @@ class Dir
   # Got config.h
   # ```
   def each : Nil
+    rewind
     while entry = read
       yield entry
     end
   end
 
+  # Calls #rewind before reading the first record.
```
asterite: This is okay as a notice, but the docs about returning an iterator are missing.

Issue #7177, opened 2018-12-11
symbol lookup error


Exilor:
After compiling successfully with 0.27 and LLVM 4.0.0 (Ubuntu on the WSL) running the executable gives this error:
```./game_server: symbol lookup error: ./game_server: undefined symbol: errno, version Object::to_s<String::Builder>:String::Builder```.
The specific method referenced can change between computers (computers that were able to run the executables before this error appeared).

Unfortunately I can't provide code to reproduce because the program has over 150k lines. I'm trying to comment out code in the hopes of narrowing down where this comes from.

The problem doesn't happen if I compile with ```--single-module``` or in MacOS.

It's not much to go on but does anyone have an inkling of where this could be coming from?

asterite:
> Ubuntu on the WSL

Do you know if it works well in a regular Ubuntu?

Exilor:
> 
> 
> > Ubuntu on the WSL
> 
> Do you know if it works well in a regular Ubuntu?

Good question. As soon as I get home I'll check.

asterite:
Sorry, closed accidentally.

I don't know what could be going on. You could check whether the missing symbol is indeed missing from the generated `.o` files (inside the cache). Unfortunately I won't have time to debug this, less without some code to reproduce it.

Exilor:
The symbol ```errno``` appears in a file called "E-rrno.o" and nowhere else.

RX14:
This only reproduces on one program, right? Compiling other programs works fine?

kostya:
seems i have similar problem once, fixed by removed `~/.cache/crystal/`

Exilor:
> 
> 
> This only reproduces on one program, right? Compiling other programs works fine?

Yes, only this one. But it happens in 3 different Linux installations. Not in MacOS.

> 
> 
> seems i have similar problem once, fixed by removed `~/.cache/crystal/`

That was among the first things I tried but nothing changed.


--

I've tried to comment out different code paths trying to pinpoint the source of the problem but I didn't get anywhere. Things that are unrelated to each other like logging or using class that wraps a Hash and a Mutex instead of a normal Hash look like triggers (the program runs if I remove either) but I have been using both extensively for a while without this happening. At one point even removing a single call to ```super``` (which other subclasses also call, for the same method, with no problems) has made it work.
Anyways this appears to be so random that if I don't figure it out in the next few days I'll close this and then open a proper issue if I can ever find out the cause and some code to reproduce.

[...]

PR #8506, opened 2019-11-21
Allow timeout in select statements

bcardiff:
* Based on #8006 by @carlhoerberg 
* Rebased regarding recent changes in the event loop and channel refactor
* Added specs
  * timeout is not activated if negative time amounts are used
  * timeout is not activated if select is non-blocking

Closes #8006

### DIFF
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -667,4 +667,65 @@ class Channel(T)
       raise ClosedError.new
     end
   end
+
+  class TimeoutAction
+    include SelectAction(Nil)
+
+    # Total amount of time to wait
+    @timeout : Time::Span
+
+    def initialize(@timeout : Time::Span)
+    end
+
+    def execute : DeliveryState
+      Fiber.current.timed_out = false
+      DeliveryState::None
+    end
+
+    def result : Nil
+      nil
+    end
+
+    def wait(context : SelectContext(Nil))
+      Fiber.timeout @timeout
+      context.try_trigger
+    end
+
+    def wait_result_impl(context : SelectContext(Nil))
+      nil
+    end
+
+    def unwait
+      Fiber.cancel_timeout
+    end
+
+    def lock_object_id
+      self.object_id
+    end
+
+    def lock
+    end
+
+    def unlock
+    end
+  end
+end
+
+# Timeout keyword for use in `select`.
+#
+# ```
+# select
+# when x = ch.recieve
+#   puts "got #{x}"
+# when timeout(1.seconds)
+#   puts "timeout"
+# end
+# ```
+#
+# NOTE: It won't trigger if the `select` has an `else` case (i.e.: a non-blocking select).
+#
+# NOTE: Using negative amounts will cause the timeout to not trigger.
+#
+def timeout_select_action(timeout : Time::Span)
+  Channel::TimeoutAction.new(timeout)
 end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -678,17 +678,18 @@ class Channel(T)
     end
   end
 
+  # :nodoc:
   class TimeoutAction
     include SelectAction(Nil)
 
     # Total amount of time to wait
     @timeout : Time::Span
+    @select_context : SelectContext(Nil)?
 
     def initialize(@timeout : Time::Span)
     end
 
     def execute : DeliveryState
-      Fiber.current.timed_out = false
       DeliveryState::None
     end
 
@@ -697,7 +698,8 @@ class Channel(T)
     end
 
     def wait(context : SelectContext(Nil))
-      Fiber.timeout @timeout
+      @select_context = context
+      Fiber.timeout(@timeout, self)
     end
 
     def wait_result_impl(context : SelectContext(Nil))
@@ -705,7 +707,6 @@ class Channel(T)
     end
 
     def unwait_impl(context : SelectContext(Nil))
-      context.try_trigger if Fiber.current.timed_out?
       Fiber.cancel_timeout
     end
 
@@ -717,6 +718,12 @@ class Channel(T)
     end
 
     def unlock
+    end
+
+    def time_expired(fiber : Fiber) : Nil
+      if @select_context.try &.try_trigger
+        Crystal::Scheduler.enqueue fiber
+      end
     end
   end
 end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -33,7 +33,7 @@ class Channel(T)
     abstract def execute : DeliveryState
     abstract def wait(context : SelectContext(S))
     abstract def wait_result_impl(context : SelectContext(S))
-    abstract def unwait
+    abstract def unwait_impl(context : SelectContext(S))
     abstract def result : S
     abstract def lock_object_id
     abstract def lock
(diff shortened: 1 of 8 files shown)


asterite:
**Warning: bike-shedding**

Given that we can do:

```crystal
puts "wait for it..."
sleep 1
puts "legendary"
```

I think it might make more sense to use `sleep` here too:

```crystal
select
when ch.receive
  # ...
when sleep 1
  # ...
end
```

so it looks like `ch.receive` and `sleep` are competing to see which one comes first.

If a user knows how to use `sleep`, using `sleep` in a select context might be what they would try to do, and if we use that name it would work right away.

bcardiff:
@asterite I actually like that `timeout` itself is not a top-level method.

Using `when sleep(t)` would break too much IMO how subexpressions can be understood. `sleep(t)` will behave very different within the `select` (due to a rewrite) and outside.

`timeout` is only available inside the `select` and I think reflects more directly the intention.

asterite:
> Using when sleep(t) would break too much IMO how subexpressions can be understood. sleep(t) will behave very different within the select (due to a rewrite) and outside

Isn't that true too for `ch.receive`, which gets rewritten to something like `ch.receive_select_action`?

So I don't see a problem with that.

bcardiff:
@asterite . You are right, it's more or less the same. The fact that sleep is a top-level methods is even more error-prone to my taste. But I can't justify why 😅.

I still like more timeout than sleep, but I can settle on any of them.

Sija:
_timeout_ IMO is way clearer: it's a distinct keyword conveying different meaning.

asterite:
Sounds good, sorry for the noise. I think there's also `timeout` in Erlang.

vlazar:
With `when timeout(t)` the meaning is immediately obvious to me when I first see this: when timeout of `t` seconds happens do something.

At `when sleep(t)` my brain kind of pauses and fails to come to conclusion right away.
The `sleep(t)` read more like an action, and so the whole `when sleep(t)` doesn't look like condition evaluating to true anymore.

src/channel.cr
```diff
@@ -653,4 +653,65 @@ class Channel(T)
       raise ClosedError.new
     end
   end
+
+  class TimeoutAction
```
j8r: Can be a struct, because immutable

j8r:
Can the new `timeout` keyword be used outside the `select` context?

src/channel.cr
bcardiff: Since others `SelectAction` are classes it is better to keep them all like that so the representation is a single pointer instead of a mixed union of value and reference types.

bcardiff:
@j8r there is no `timeout` keyword. It is parsed a function call. But due to the rewrite of `select`, there is no `timeout` top-level function. What would you expect by 

> used outside the `select` context?

?

src/channel.cr
```diff
@@ -653,4 +653,65 @@ class Channel(T)
...
+      DeliveryState::None
+    end
+
+    def result : Nil
+      nil
+    end
+
+    def wait(context : SelectContext(Nil))
+      Fiber.timeout @timeout
+      context.try_trigger
```
firejox: If I understand correctly, the below code will not receive the data during waiting.
```crystal
select
when ch.receive
when timeout(1.seconds)
end
```
Because the `context` is triggered before `Crystal::Scheduler.reschedule`, any `ch.send` will not awake the select fiber.

The timeline explanation:
```
    context.try_trigger( by timeout_action )
              |
              V
---->[ each action.wait ]---->[ reschedule ]------>[ ch.send ]---->[ timeout ]-->[ select fiber awake ]
                                    ^                   ^               ^
                                    |                   |               |
                              start waiting <-- 0.5s -->x<---- 0.5s --->x
                                                        |
                                            context.try_trigger ( fail because it has been triggered. )
```

I think the timeout worked as below is better.

```
                                                        context.try_trigger( failed then not enqueue fiber )
                                                                        |
                                                                        V
---->[ each action.wait ]---->[ reschedule ]------>[ ch.send ]---->[ timeout ]-->[ select fiber awake ]
                                    ^                   ^               ^
                                    |                   |               |
                              start waiting <-- 0.5s -->x<---- 0.5s --->x
                                                        |
                                            context.try_trigger ( succeed then enqueue fiber )

Or
                                                                      cleanup timeout
                                                                            |
                                                                            V
---->[ each action.wait ]---->[ reschedule ]------>[ ch.send ]---->[ select fiber awake ]---->x--->
                                    ^                   ^               ^                     ^
                                    |                   |               |                     |
                              start waiting <-- 0.5s -->x<--------------+--- 0.5s ----------->x
                                                        |
                                            context.try_trigger ( succeed then enqueue fiber )
```

j8r:
@bcardiff I meant, like `sleep()`, we can't use `timeout()` directly as a regular top-level method/keyword.

asterite:
Well, that's the point, `timeout` in a non-select context doesn't make sense.

j8r:
Yep exactly, that's why I asked, to be sure.

src/channel.cr
bcardiff: @firejox good catch. I'm surprised neither of the specs complains. I think somehow I (wrongly) expected to `Fiber.timeout` pause the execution of the fiber and continue from there.

I added more specs and move the `try_trigger` in the last commit.

The only specs that fails without the `try_trigger` change is "allows receiving while waiting (same fiber multiple times)"

asterite: @firejox You are on fire! 👏

firejox: @bcardiff I would suggest to put `context.try_trigger` into the event of `event loop`, just like [this](https://github.com/firejox/crystal/commit/31b1835a089aee6b12be134c1f3e34cf0770fcd3). Because it is hard to proof that `event loop` will not run after `ch.send`. If  `event loop` runs, the fiber will be enqueued again. Hence, I think that  `context.try_trigger` in event will be better.

bcardiff: @firejox I fail to see a real difference with the proposed refactor. I think there is still no guarantee that if the timeout event is triggered and at the same time a channel-action can be performed which one will succeed the `try_trigger`. And that is fine, is the same scenario as competing channel-actions.

And on the downside, the fiber and the channels end up more coupled after the refactor IMO.

firejox: @bcardiff Sorry for confusing refactor. The `try_trigger` determines which action can enqueue the fiber. However, the timeout event [here](https://github.com/crystal-lang/crystal/pull/8506/files#diff-4501d78495e747fe23a7fee036475ae2R35-R37) enqueue the fiber without `try_trigger` limitation. I worry that would lead the problem like #8300 . It enqueues the fiber more than one time.

bcardiff:
@firejox thanks for the feedback. I rebased this on master and apply a similar refactor to the one you suggested in https://github.com/crystal-lang/crystal/pull/8506#discussion_r357906714 

WDYT?

firejox:
@bcardiff Looks good. I have no more feedback now.

RX14:
This now triggers https://github.com/crystal-lang/crystal/issues/7177 in CI

i think, because more classes were added

bcardiff:
Rebased on master with hopefully a happier CI.

Some approval is still needed.

PR #8507, opened 2019-11-21
Add HTTP::Client#write_timeout

### LINKED ISSUE #8487: HTTP::Client is missing write_timeout setting
It seems that HTTP::Client has all the timeouts other then write_timeout.
I'm in a situation where I need this to be configurable.

Is there a reason for this not being there or just an oversight\didn't get to it situation? 

* Note: https://docs.ruby-lang.org/en/2.6.0/Net/HTTP.html#attribute-i-write_timeout (ruby has it :) )

Sija:
Fixes #8487 • (fwiw, Ruby [has it](https://ruby-doc.org/stdlib-2.6/libdoc/net/http/rdoc/Net/HTTP.html#write_timeout) since version `2.6.0` - see the [relevant ticket](https://bugs.ruby-lang.org/issues/13396))

### DIFF
--- a/src/http/client.cr
+++ b/src/http/client.cr
@@ -117,6 +117,7 @@ class HTTP::Client
   @dns_timeout : Float64?
   @connect_timeout : Float64?
   @read_timeout : Float64?
+  @write_timeout : Float64?
 
   # Creates a new HTTP client with the given *host*, *port* and *tls*
   # configurations. If no port is given, the default one will
@@ -293,6 +294,18 @@ class HTTP::Client
   # ```
   def read_timeout=(read_timeout : Time::Span)
     self.read_timeout = read_timeout.total_seconds
+  end
+
+  # Sets the write timeout - if any chunk of request is not written
+  # within the number of seconds provided, `IO::Timeout` exception is raised.
+  def write_timeout=(write_timeout : Number)
+    @write_timeout = write_timeout.to_f
+  end
+
+  # Sets the write timeout - if any chunk of request is not written
+  # within the provided `Time::Span`,  `IO::Timeout` exception is raised.
+  def write_timeout=(write_timeout : Time::Span)
+    self.write_timeout = write_timeout.total_seconds
   end
 
   # Sets the number of seconds to wait when connecting, before raising an `IO::Timeout`.
@@ -754,6 +767,7 @@ class HTTP::Client
     hostname = @host.starts_with?('[') && @host.ends_with?(']') ? @host[1..-2] : @host
     socket = TCPSocket.new hostname, @port, @dns_timeout, @connect_timeout
     socket.read_timeout = @read_timeout if @read_timeout
+    socket.write_timeout = @write_timeout if @write_timeout
     socket.sync = false
 
     {% if !flag?(:without_openssl) %}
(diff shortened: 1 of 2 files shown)


Sija:
@bcardiff @straight-shoota Could we have this merged into `0.32.0`? Pretty please :)

src/http/client.cr
```diff
@@ -295,6 +296,16 @@ class HTTP::Client
     self.read_timeout = read_timeout.total_seconds
   end
 
+  # Sets the number of seconds to wait for one block to be written (via one `write(2)` call) before raising an `IO::Timeout`.
```
bcardiff: I would prefer a description that does not involve syscall and more from the users perspective, like `read_timeout=` one.

Out of curiosity, other than the client suddenly not using the established connection, in which scenarios this is needed?

Sija: I copied the description from Ruby, though I'm open for improvement suggestions :)

bcardiff: It bugs me that write_timeout can be understood for any of the directions. Is the client not writing or the server not writing. The same applies to the read_timeout. But the sample code there shows no content written. Here that is not evident since the sample is only in the spec.

I'm pointing as a consumer of the documentation in mind, without knowledge of syscall.

My (optional) suggestion would be to clarify in which situation the timeout could raise and why. And to remove the mention to the syscall.

Sija: > Sets the write timeout in seconds - the maximum time you allow to write (send) the request data to the service before raising an `IO::Timeout`.

Would that be better?

Sija: @bcardiff @RX14 ping

RX14: It's the timeout while waiting to make *any progress* sending the request, not for sending the whole request.

Sija: >Sets the write timeout in seconds - if any chunk of response is not written within number of seconds provided to write_timeout, `IO::Timeout` exception is raised.

how about now?

bcardiff: (Sorry the delay, I’m on vacations) looks better, but either I am miss understanding something or I would expect to be: any chunk of the request, but not the response.

Isn’t the spec asserting that the client is not able to send the request body in the required amount of time?

Sija: I think that's correct, I've updated my previous comment.

Sija: @RX14 @bcardiff ping

RX14: I still prefer "if no progress is made sending the request data for the number of seconds provided to write_timeout"

Sija: I'm bit lost, could you please edit the docs in this PR to your liking and, since it's approved possibly merge it already? Would be a shame to keep monkey-patching that just because it was too hard to sync on the content of the docs - for over 2 week - which can always be improved later on...

RX14: Just put whats in your comment in, I'm not too bothered.

Sija: My intention is to have this method understandably documented, so if you have a full comment text I could put there for everyone to be happy with it I'd most happy to do that.

Sija: @RX14 @bcardiff ready for re-review.

PR #8508, opened 2019-11-22
Use String/Nil#presence in a few places throughout stdlib

Sija:
SSIA

### DIFF
--- a/src/compiler/crystal/compiler.cr
+++ b/src/compiler/crystal/compiler.cr
@@ -326,7 +326,7 @@ module Crystal
           object_name = %(%*)
         end
 
-        if (link_flags = @link_flags) && !link_flags.empty?
+        if link_flags = @link_flags.presence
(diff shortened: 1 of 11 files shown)


asterite:
Just note that `a && !a.empty?` is different than `a.presence` because the latter might be a bit slower because it needs to traverse all chars, while `empty?` just checks for bytesize.

I'm not saying that this PR is not good, but with it merged it's not exactly the same code, and it might actually be slightly slower (maybe not).

Sija:
> Just note that `a && !a.empty?` is different than `a.presence` because the latter might be a bit slower because it needs to traverse all chars, while `empty?` just checks for bytesize.

@asterite Good point! Also worth noting is the cause of that - i.e. using `String#blank?`, instead of `String#empty?` - therefore every case should be thought out in regards to whether whitespaces (and speed) are sth you care about.

Issue #8515, opened 2019-11-25
Optimize `Array#rotate!`


asterite:
Extracted from https://forum.crystal-lang.org/t/array-rotate-can-be-faster/1404

Right now `Array#rotate!` will use a temporary array for the elements to rotate, followed by a couple of memcopies.

There are many ways we could optimize `Array#rotate!`:

- Special case for `n = 1` and `n = -1`: avoid allocating a tmp array and copy to a tmp var
- Special case for `n <= N`: we could allocate a static array with a few elements to avoid a temporary allocation on the heap
- We could use the reversal algorithm, like Ruby does: https://www.geeksforgeeks.org/program-for-array-rotation-continued-reversal-algorithm/ (though in practice it seems to be slower than allocating a temporary array, at least for big arrays)

Maybe we could start with special-casing for `n = 1` and `n = -1`, or for `n` less than a small N like 10 using a static array.

straight-shoota:
I suppose most rotate use cases will apply a relatively small `n`, so the first two improvements would already be a good win.

PR #8516, opened 2019-11-25
Array rotate optimization

wontruefree:
fixes https://github.com/crystal-lang/crystal/issues/8515

Add special case for array rotate to optimize for small arrays.

Looks like the for size 3, 100 the new rotate is faster and 10, 32 the old rotate is faster.  I dont know why.

```crystal
require "benchmark"

x3 = Array(Int32).new(3){|i| i}
x10 = Array(Int32).new(10){|i| i}
x32 = Array(Int32).new(32){|i| i}
x100 = Array(Int32).new(100){|i| i}

GC.disable

Benchmark.ips do |x|
  x.report("3 rotate! 2") { x3.rotate!(2) }
  x.report("3 new_rotate! 2") { x3.new_rotate!(2) }
end
Benchmark.ips do |x|
  x.report("10 rotate! 2") { x10.rotate!(2) }
  x.report("10 new_rotate! 2") { x10.new_rotate!(2) }
end
Benchmark.ips do |x|
  x.report("32 rotate! 2") { x32.rotate!(2) }
  x.report("32 new_rotate! 2") { x32.new_rotate!(2) }
end
Benchmark.ips do |x|
  x.report("100 rotate! 2") { x100.rotate!(2) }
  x.report("100 new_rotate! 2") { x100.new_rotate!(2) }
end
```

```text
$ crystal tmp.cr --release
    3 rotate! 2  18.07M ( 55.33ns) (± 5.25%)  48.0B/op   1.00× slower
3 new_rotate! 2  18.12M ( 55.20ns) (± 5.39%)   0.0B/op        fastest
    10 rotate! 2  18.38M ( 54.40ns) (± 5.02%)  48.0B/op        fastest
10 new_rotate! 2  18.18M ( 55.01ns) (± 3.26%)   0.0B/op   1.01× slower
    32 rotate! 2  17.93M ( 55.77ns) (± 5.47%)  48.0B/op        fastest
32 new_rotate! 2  17.87M ( 55.98ns) (± 4.70%)   0.0B/op   1.00× slower
    100 rotate! 2  15.36M ( 65.12ns) (± 9.45%)  48.0B/op   1.10× slower
100 new_rotate! 2  16.87M ( 59.29ns) (± 7.10%)   0.0B/op        fastest
```


asterite:
Thank you! What happens if you remove `GC.disable` from the benchmark? I don't think that's something one should do in benchmarks because you do want to take into account the time it takes for the GC to claim memory.

src/array.cr
```diff
@@ -1545,22 +1544,55 @@ class Array(T)
...
+    elsif n == 1
+      tmp = self[0]
+      @buffer.move_from(@buffer + n, size - n)
+      self[-1] = tmp
+    elsif n == -1
+      tmp = self[-1]
+      @buffer.move_from(@buffer + n, size - n)
+      self[0] = tmp
+    elsif n <= SMALL_ARRAY_SIZE
+      tmp = StaticArray(T, SMALL_ARRAY_SIZE).new{|i| self[i % size] }
```
asterite: Maybe we should allocate `SMALL_ARRAY_SIZE` but not initialize it with every element in the range? Also `self[i % size]` will do a lot of bound checks, plus `%` is a bit slower than `remainder`. Not necessary to change it now, maybe an optimization for later.

wontruefree:
@asterite looks like the only one that is faster without the GC is an array of 3.
```text
    3 rotate! 2  17.85M ( 56.03ns) (± 3.01%)  48.0B/op   1.03× slower
3 new_rotate! 2  18.47M ( 54.14ns) (± 4.08%)   0.0B/op        fastest
    10 rotate! 2  19.00M ( 52.63ns) (± 1.90%)  48.0B/op        fastest
10 new_rotate! 2  18.53M ( 53.96ns) (± 4.80%)   0.0B/op   1.03× slower
    32 rotate! 2  18.32M ( 54.59ns) (± 1.78%)  48.0B/op        fastest
32 new_rotate! 2  18.09M ( 55.27ns) (± 2.54%)   0.0B/op   1.01× slower
    100 rotate! 2  18.09M ( 55.27ns) (± 1.13%)  48.0B/op        fastest
100 new_rotate! 2  16.72M ( 59.80ns) (± 5.26%)   0.0B/op   1.08× slower
```

src/array.cr
wontruefree: I did that because StaticArray size cannot be declared at runtime.
```crystal
StaticArray(Int32, n).new(0)
```
```text
expecting token ')', not 'n'
```

asterite: You can do:

```crystal
tmp = uninitialized Int32[SMALL_ARRAY_SIZE]
# now only copy n elements, not SMALL_ARRAY_SIZE, into it
```

wontruefree:
@asterite I filled the rest of the array with filler.  Looks like we are faster in all cases now.
```text
 crystal tmp.cr --release
    3 rotate! 2  18.27M ( 54.72ns) (± 1.80%)  48.0B/op   4.04× slower
3 new_rotate! 2  73.80M ( 13.55ns) (± 4.04%)   0.0B/op        fastest
    10 rotate! 2  19.47M ( 51.37ns) (± 3.52%)  48.0B/op   4.14× slower
10 new_rotate! 2  80.55M ( 12.41ns) (± 3.68%)   0.0B/op        fastest
    32 rotate! 2  18.42M ( 54.29ns) (± 2.21%)  48.0B/op   3.23× slower
32 new_rotate! 2  59.41M ( 16.83ns) (±10.28%)   0.0B/op        fastest
    100 rotate! 2  18.05M ( 55.41ns) (± 1.35%)  48.0B/op   2.99× slower
100 new_rotate! 2  54.05M ( 18.50ns) (± 4.67%)   0.0B/op        fastest
```

wontruefree:
With the uninitialized StaticArray size 3 is a little slower but faster then now.  And the rest are more performant.

```text
$ crystal tmp.cr --release                                                                          
    3 rotate! 2  17.06M ( 58.60ns) (± 4.18%)  48.0B/op   3.19× slower
3 new_rotate! 2  54.51M ( 18.35ns) (± 3.32%)   0.0B/op        fastest
    10 rotate! 2  18.17M ( 55.05ns) (± 2.08%)  48.0B/op   4.00× slower
10 new_rotate! 2  72.64M ( 13.77ns) (± 3.57%)   0.0B/op        fastest
    32 rotate! 2  17.10M ( 58.49ns) (± 2.52%)  48.0B/op   3.36× slower
32 new_rotate! 2  57.40M ( 17.42ns) (± 9.59%)   0.0B/op        fastest
    100 rotate! 2  16.75M ( 59.70ns) (± 1.49%)  48.0B/op   3.21× slower
100 new_rotate! 2  53.79M ( 18.59ns) (± 3.54%)   0.0B/op        fastest
```

wontruefree:
using `case` is even a little faster except for size 32 I have no idea why I thought llvm would basically optimize these to be the same.

```text
$ crystal tmp.cr --release
        3 rotate! 2  16.82M ( 59.44ns) (± 2.68%)  48.0B/op   3.69× slower
    3 new_rotate! 2  53.14M ( 18.82ns) (± 4.53%)   0.0B/op   1.17× slower
3 new_new_rotate! 2  62.16M ( 16.09ns) (± 4.54%)   0.0B/op        fastest
        10 rotate! 2  17.66M ( 56.64ns) (± 3.26%)  48.0B/op   4.01× slower
    10 new_rotate! 2  67.39M ( 14.84ns) (± 4.85%)   0.0B/op   1.05× slower
10 new_new_rotate! 2  70.79M ( 14.13ns) (± 4.88%)   0.0B/op        fastest
        32 rotate! 2  16.98M ( 58.89ns) (± 1.94%)  48.0B/op   3.35× slower
    32 new_rotate! 2  56.86M ( 17.59ns) (± 9.36%)   0.0B/op        fastest
32 new_new_rotate! 2  55.90M ( 17.89ns) (± 8.69%)   0.0B/op   1.02× slower
        100 rotate! 2  16.35M ( 61.18ns) (± 2.49%)  48.0B/op   3.28× slower
    100 new_rotate! 2  51.34M ( 19.48ns) (± 7.49%)   0.0B/op   1.04× slower
100 new_new_rotate! 2  53.57M ( 18.67ns) (± 4.99%)   0.0B/op        fastest
```

src/array.cr
```diff
@@ -1545,11 +1544,39 @@ class Array(T)
     self
   end
 
+  # Returns `self` with all the elements shifted `n` times.
+  #
+  # ```
+  # a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+  # a.rotate!    # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+  # a.rotate!(1) # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
```
straight-shoota: Since `#rotate!` modifies `a`, this is actually the result of a 0 rotation.
```suggestion
  # a.rotate!(0) # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
```
And the next one rotates by 2.

You might want to choose different results, and maybe also showcase a negative rotation.

wontruefree: This is 0 indexed so I think the return on `rotate(1)` would start with `1`

straight-shoota: Check the result to verify: https://carc.in/#/r/82k0

The first call to `#rotate!` modifies `a`, so the second one rotates the result from the first one.

wontruefree: I see your point.  I have updated the docs.

src/array.cr
```diff
@@ -1545,11 +1544,46 @@ class Array(T)
...
+  # a3 # => [3, 4, 5, 6, 7, 8, 9, 0, 1, 2]
+  # ```
   def rotate!(n = 1)
     return self if size == 0
+    # This is broken into cases for perfromance (#8515)
     n %= size
-    return self if n == 0
-    if n <= size // 2
+    case n
+    when 0
```
Sija:
```suggestion
```

src/array.cr
```diff
@@ -1545,11 +1544,46 @@ class Array(T)
...
+    when 1
```
Sija:
```suggestion
    when 0, 1
```

src/array.cr
Sija:
```suggestion
    when 0
      # ignore
```

src/array.cr
```diff
@@ -1545,11 +1544,46 @@ class Array(T)
     self
   end
 
+  # Returns `self` with all the elements shifted `n` times.
+  #
+  # ```
+  # a1 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+  # a2 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```
straight-shoota: These parallel arrays make the sample difficult to grasp. I think its better to just use a single array (or at least only one literal). That perfectly demonstrates that each call to `#rotate!` builds on the result of the previous.

src/array.cr
```diff
@@ -1545,11 +1544,47 @@ class Array(T)
...
+  # a2.rotate!(1)
+  # a3.rotate!(3)
+  #
+  # a1 # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+  # a2 # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+  # a3 # => [3, 4, 5, 6, 7, 8, 9, 0, 1, 2]
+  # ```
   def rotate!(n = 1)
     return self if size == 0
+    # This is broken into cases for perfromance (#8515)
```
asterite: I wouldn't mention a GitHub issue in code (in bug specs it's fine to understand why a spec for a specific case exists).

straight-shoota: Why not? What's the downside? Such a reference obviously provides context information. Worst case is when the repo switches away from GitHub, the reference might not lead anywhere. But apart from that...?

asterite: I guess this is fine.

straight-shoota: We're actually using quite a few references to GitHub issues in code comments. I think that's fine.

Sija: @asterite I misread your comment thinking that you're for it ;) we do it in specs and in this case would be pretty informative into _why_ exactly these optimizations were performed.

asterite: > why exactly these optimizations were performed

I think optimizations are performed because they are always good? I don't think we need to say "Here's an optimization because it's better to have fast code instead of slow code" or "Here we are doing an optimization because someone realized this could be optimized". Or what could another reason for optimizing code be?

straight-shoota: Sure, optimizations are good in general. But there always are different possibilities and you wouldn't want to just randomly shoot optimizations around. Some work better than others, but there are always trade-offs and complex decisions why to choose a specific optimization.

wontruefree:
please wait before merging I am almost done with implementing an in place transformation.

src/array.cr
```diff
@@ -1545,11 +1544,47 @@ class Array(T)
...
-    return self if n == 0
-    if n <= size // 2
+    case n
+    when 0
+      # ignore
+    when 1
+      tmp = self[0]
+      @buffer.move_from(@buffer + n, size - n)
+      self[-1] = tmp
+    when -1
```
firejox: This would never be triggered since `n % size` is always positive.

straight-shoota: Good catch. It should probably match `size - 1` because that's `-1 % size`

wontruefree:
@asterite and @sdogruyol this is ready to merge :)
I added a few more test but was unable to do an in place rotate and have it be more performant then the current solution.

src/array.cr
```diff
@@ -1545,22 +1544,67 @@ class Array(T)
...
+    if n == 0
+      # ignore
+    elsif n == 1
+      tmp = self[0]
+      @buffer.move_from(@buffer + n, size - n)
+      self[-1] = tmp
+    elsif n == -1
+      tmp = self[-1]
+      @buffer.move_from(@buffer + n, size - n)
+      self[0] = tmp
```
firejox: `n == (size  - 1)`

src/array.cr
```diff
@@ -1545,22 +1544,67 @@ class Array(T)
...
+    elsif n <= SMALL_ARRAY_SIZE // 2
+      tmp_buffer = uninitialized StaticArray(T, SMALL_ARRAY_SIZE)
+      tmp_buffer.to_unsafe.copy_from(@buffer, n)
+      @buffer.move_from(@buffer + n, size - n)
+      (@buffer + size - n).copy_from(tmp_buffer.to_unsafe, n)
```
firejox: `n <= SMALL_ARRAY_SIZE`

src/array.cr
wontruefree: I will add a test case for this as well.

src/array.cr
wontruefree: This should be checking for shift right vs shift left.
I added a comment to be more clear.

asterite:
There's actually another optimization we can do, if we really want to squeeze all we can from this method.

It seems that when the size of the array is 9 or less, the reversal algorithm works faster than doing a bunch of `memcpy`.

The logic is this:

```crystal
Slice.new(to_unsafe, size).reverse!
Slice.new(to_unsafe + size - n, n).reverse!
Slice.new(to_unsafe, size - n).reverse!
```

We should do that if `size <= 9`, so we could put that condition after `if n == 0`. Actually, my benchmarks seems to show that it's faster when `(n == 1 && size <= 7) || size <= 9` but I don't know if we'd like it to be that specific... so maybe `if size <= 8` is good enough. Or maybe just leave this PR like that, though the difference is significant in some cases (1.28x faster).

firejox:
I think we can apply reversal algorithm into this PR. In my [benchmark](https://gist.github.com/firejox/806a24c8c0a4dc3463b112cc234a6607), the reversal algorithm performs better when rotate distance is close to (N/2). If array is large, it would allocate large memory to do rotation. This will degrade performance. Hence, the rotate logic could be:

```crystal
if rotate_distance == 0
  # ignore
else rotate_distance <= SMALL_SIZE
  # stack-based copy algorithm
else
  # reversal algorithm
end
```

src/array.cr
firejox: So if I do `rotate! 9` on the array with size 100, it will not apply the stack-based algorithm, right? It doesn't make any sense. It allocates `SMALL_ARRAY_SIZE` static array, but it is only able to use the half of the array on left rotation.

asterite:
@firejox I benchmark Ed the existing code against the reversal algorithm and for some reason the existing code was faster for large arrays. You could maybe try to create a benchmark and see if you get the same results.

firejox:
@asterite  I think you're right. My benchmark result shows that the reversal algorithm is slower than existing code for large array.  But still, the reversal algorithm could be useful in some place.

```
Benchmark when rotate n/3 elements
    10 rotate! 3   7.57M (132.04ns) (± 8.60%)  65.0B/op   3.30× slower
10 new_rotate! 3  25.00M ( 40.01ns) (± 3.27%)   0.0B/op        fastest
    100 rotate! 33   1.44M (694.30ns) (± 1.45%)  177B/op   5.42× slower
100 new_rotate! 33   7.80M (128.21ns) (± 6.13%)  0.0B/op        fastest
    1000 rotate! 333 281.39k (  3.55µs) (± 2.32%)  2.03kB/op   4.03× slower
1000 new_rotate! 333   1.13M (881.35ns) (± 5.50%)    0.0B/op        fastest
    10000 rotate! 3333  48.39k ( 20.66µs) (± 1.84%)  13.1kB/op   2.31× slower
10000 new_rotate! 3333 111.98k (  8.93µs) (± 7.41%)    0.0B/op        fastest
    100000 rotate! 33333   5.74k (174.22µs) (± 1.01%)  130kB/op   1.75× slower
100000 new_rotate! 33333  10.02k ( 99.83µs) (± 6.07%)   0.0B/op        fastest
    1000000 rotate! 333333 496.68  (  2.01ms) (± 0.92%)  1.27MB/op        fastest
1000000 new_rotate! 333333 492.28  (  2.03ms) (± 2.31%)    0.0B/op   1.01× slower
    10000000 rotate! 3333333  55.13  ( 18.14ms) (± 0.90%)  12.7MB/op        fastest
10000000 new_rotate! 3333333  49.70  ( 20.12ms) (± 2.67%)    0.0B/op   1.11× slower
```
Besides, I found another in-place algorithm that has nice performance. It is complicate but perform well. The code is in [here](https://gist.github.com/firejox/e657a3c214eda56690d5d05aa9f90d9f). And the benchmark is the following.

```
Benchmark when rotate n/3 elements
        10 rotate! 3   7.80M (128.28ns) (± 8.15%)  65.0B/op   3.23× slower
    10 new_rotate! 3  25.17M ( 39.73ns) (± 3.33%)   0.0B/op        fastest
10 new_new_rotate! 3  24.89M ( 40.18ns) (± 3.75%)   0.0B/op   1.01× slower
        100 rotate! 33   1.36M (733.62ns) (± 2.83%)  177B/op   5.59× slower
    100 new_rotate! 33   7.62M (131.32ns) (± 2.79%)  0.0B/op        fastest
100 new_new_rotate! 33   7.09M (140.99ns) (± 2.96%)  0.0B/op   1.07× slower
        1000 rotate! 333 344.30k (  2.90µs) (± 1.80%)  1.35kB/op   3.16× slower
    1000 new_rotate! 333   1.08M (929.08ns) (± 2.13%)    0.0B/op   1.01× slower
1000 new_new_rotate! 333   1.09M (918.57ns) (± 0.63%)    0.0B/op        fastest
        10000 rotate! 3333  50.05k ( 19.98µs) (± 2.26%)  13.1kB/op   2.28× slower
    10000 new_rotate! 3333 109.90k (  9.10µs) (± 9.55%)    0.0B/op   1.04× slower
10000 new_new_rotate! 3333 114.09k (  8.77µs) (± 2.37%)    0.0B/op        fastest
        100000 rotate! 33333   5.67k (176.36µs) (± 2.06%)  130kB/op   1.80× slower
    100000 new_rotate! 33333  10.21k ( 97.90µs) (± 2.89%)   0.0B/op        fastest
100000 new_new_rotate! 33333   9.97k (100.31µs) (± 3.62%)   0.0B/op   1.02× slower
        1000000 rotate! 333333 501.28  (  1.99ms) (± 1.23%)  1.27MB/op   1.12× slower
    1000000 new_rotate! 333333 513.57  (  1.95ms) (± 1.35%)    0.0B/op   1.09× slower
1000000 new_new_rotate! 333333 559.62  (  1.79ms) (± 2.25%)    0.0B/op        fastest
        10000000 rotate! 3333333  55.62  ( 17.98ms) (± 1.29%)  12.7MB/op        fastest
    10000000 new_rotate! 3333333  50.82  ( 19.68ms) (± 1.51%)    0.0B/op   1.09× slower
10000000 new_new_rotate! 3333333  55.15  ( 18.13ms) (± 1.60%)    0.0B/op   1.01× slower
```

wontruefree:
@firejox I was using memcopy and could not get this to work.  I love that you figured it out.  I would be interested in seeing if we could speed this up by using the static array in some places and moving things in blocks.

src/array.cr
```diff
@@ -1545,22 +1544,67 @@ class Array(T)
...
+      # ignore
+    elsif n == 1
+      tmp = self[0]
+      @buffer.move_from(@buffer + n, size - n)
+      self[-1] = tmp
+    elsif n == (size - 1)
+      tmp = self[-1]
+      (@buffer + size - n).move_from(@buffer, n)
+      self[0] = tmp
+    elsif n <= SMALL_ARRAY_SIZE // 2 # shift right
```
RX14: Why `// 2` on this branch and not the left shift branch?

wontruefree: shift left and shift right is describing the direction of the algorithm.

RX14: Yes but why is it `<= SMALL_ARRAY_SIZE` on one branch and `<= SMALL_ARRAY_SIZE // 2` on the other?

jwoertink:
Is this something that could make it in to 0.33 coming up?

wontruefree:
@jwoertink there was some back and forth on this.
I don't know what is really blocking this.

straight-shoota:
For example, there was no response on the last review question: https://github.com/crystal-lang/crystal/pull/8516#discussion_r356364894

@jwoertink Since this is just an optimization, there's really no pressure to put this into 0.33.0

src/array.cr
wontruefree: so this comment might be out of date to the most current commit.  right now it is checking size.
```cr
elsif n <= size // 2
```

But the reason why one is checking against the `SMALL_ARRAY_SIZE` and the other `size` is because one is being optimized.  As for the `//2` and not I dont know.

RX14: I meant

```
elsif n <= SMALL_ARRAY_SIZE // 2
```

vs

```
elsif size - n <= SMALL_ARRAY_SIZE
```

wontruefree: that works it was just not how I was thinking about it.

RX14: ??

This is two excerpts from your PR

wontruefree:
Is there anything else needed to merge this?

src/array.cr
```diff
@@ -1596,22 +1595,69 @@ class Array(T)
...
+  # a3 # => [3, 4, 5, 6, 7, 8, 9, 0, 1, 2]
+  # ```
   def rotate!(n = 1)
     return self if size == 0
+    # This is broken into cases for perfromance (#8515)
     n %= size
-    return self if n == 0
-    if n <= size // 2
+
+    puts "n: #{n}, SMALL_ARRAY_SIZE: #{SMALL_ARRAY_SIZE}"
```
jhass, line 1618: Leftover debug puts.

src/array.cr
```diff
@@ -1596,22 +1595,69 @@ class Array(T)
...
+  # a1.rotate!
+  # a2.rotate!(1)
+  # a3.rotate!(3)
+  #
+  # a1 # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+  # a2 # => [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]
+  # a3 # => [3, 4, 5, 6, 7, 8, 9, 0, 1, 2]
+  # ```
   def rotate!(n = 1)
     return self if size == 0
```
jhass, line 1614: While we're here, should this raise on `n < 0`?

jhass:
The discussion above seems unresolved too.

src/array.cr
wontruefree: This conditional has changed.  Do you still have a concern about this?

src/array.cr
wontruefree: This previously worked for negative numbers and there are some tests for negative numbers so I don't think it should raise.

wontruefree:
@firejox Could you please review this PR?  If you have a proposal to this PR from this https://github.com/crystal-lang/crystal/pull/8516#issuecomment-561963769 I would love to see it.

src/array.cr
```diff
@@ -1601,11 +1600,45 @@ class Array(T)
...
     n %= size
-    return self if n == 0
-    if n <= size // 2
+
+    if n == 0
```
Sija, line 1622: Redundant branch.
Whole thing would imo look better with the `case` statement.

Issue #7339, opened 2019-01-27
Port signals to Windows


malte-v:
This issue is about porting `Signal` to Windows and therefore making it possible to implement other relevant features like `Process`.

Short introduction (moved from #5430):
--
[Microsft doc about signals](https://docs.microsoft.com/en-us/previous-versions/xdkz3x12(v%3Dvs.140))
I already looked at this page, and I think these signals are only there because they are part of the C standard library. From [wikipedia](https://en.wikipedia.org/wiki/C_signal_handling):

> The C standard defines only 6 signals. They are all defined in signal.h header (csignal header in C++):
> 
> SIGABRT - "abort", abnormal termination.
> SIGFPE - floating point exception.
> SIGILL - "illegal", invalid instruction.
> SIGINT - "interrupt", interactive attention request sent to the program.
> SIGSEGV - "segmentation violation", invalid memory access.
> SIGTERM - "terminate", termination request sent to the program.
> 
> Additional signals may be specified in the signal.h header by the implementation. For example, Unix and Unix-like operating systems (such as Linux) define more than 15 additional signals; see Unix signal.

And these 6 signals are the only ones "supported" by Windows.

There is already an [issue](https://github.com/nodejs/node/issues/12378) about this topic on the Node.js repo.
sam-github commented there:

> [...] Windows as an O/S has absolutely no signal support. [...]
> 
> If SIGTERM was supported on Windows, you could listen for it and receive it... but you cannot. You can listen..., but you'll never receive it.

And I guess this is exactly the case. You have a function for listening to signals (`signal()`), but there is nothing like `kill()` for sending signals on Windows.

---

So this is about finding replacement functions to "fake" signals on Windows. I already found some replacements for signals one might want to send:

- SIGTERM: `SendMessage(WM_CLOSE)` <- this is also what `taskkill.exe` uses; works for windowed processes
- SIGKILL: `TerminateProcess()` <- used by `taskkill.exe /f`; works for windowed and console processes
- SIGINT: `GenerateConsoleCtrlEvent()` <- works for console processes, simulates Ctrl + C

List of UNIX signals: https://en.wikipedia.org/wiki/Signal_(IPC)
Another microsoft doc that mentions replacements for UNIX signals: https://docs.microsoft.com/en-us/previous-versions/ms811896(v=msdn.10)#signals-and-signal-handling

maltevoos:
Possible replacements for SIGSTOP and SIGCONT:

[...]

Issue #6421, opened 2018-07-21
Process.fork has dangerous semantics


forksaber:
##  ```forkall```(Current)

```Process.fork``` currently has ```forkall``` semantics with respect to fibers. What this means is that all the fibers which were alive during the ```Process.fork``` call remain alive in the forked process.
### Advantages
- no deadlocks possible due to locked mutexes in contrast to when only a single fiber survives the fork
### Disadvantages
Since **fd/sockets** are shared across forks, fibers which were doing writes will do **double writes** which will cause:
- broken http responses
- duplicated data in files
- corrupted db queries

Similarly, **fd/sockets** reads may get divided across forks. So, both processes will see
- partial http requests
- partial db responses
- partial file data

An example of double writes:
```crystal
chan = Channel(Nil).new
spawn do
  10.times do |i|
    sleep rand
    puts "line #{i}"
  end
  chan.send(nil)
end
Process.fork do
  sleep 1000
end

chan.receive
```
Expected output:
```
line 0
line 1
line 2
line 3
line 4
```
Actual output:
```
line 0
line 0
line 1
line 2
line 1
line 3
line 4
line 2
line 3
line 4
```

IMO, these are really severe disadvantages so these semantics ought to change.

# Alternatives

## 1. ```fork``` like semantics (linux/unix)
In this scenario, only the fiber invoking ```Process.fork``` lives on in the forked process. Rest of the fibers are either collected(majority of them) and a few core runtime fibers like ```SignalHandler loop``` are recreated.

### Advantages
- no unintentional socket/fd corruption
- easy to reason about
- similar to how threads would behave once thread-support lands

### Disadvantages
- **Deadlocks**: Any attempt to lock a mutex which was locked in non forking Fiber will result in a deadlock. This is manageable when you own the mutexes but if mutexes belong to stdlib/shard then deadlocks are more or less imminent. ```pthread_atfork``` like api for fibers may help here though i am not sure how successful that will be.

## 2. Disable fork without exec (go/erlang)
This approach is in line with what **go** and **erlang** takes. They find ```fork``` too difficult to implement correctly in a multi-threaded/multi-fiber/multi-actor environment that they don't even bother implementing it.

### Disadvantages
- won't be able to daemonize the process using the double fork technique but can definitely use a C binary which ```exec()```s to a crystal binary.

## Conclusion
[...]

PR #8518, opened 2019-11-25
Windows Process API

jan-zajic:
Replace **Fork–exec** Unix technique with **CreateProcess** API on MS Win. As Windows processes don't use the *nix fork idiom, there is no analogue in Windows to forking. Therefore fork **without** exec on Win is **not** implemented (it's almost impossible and not so useful on Windows). 

Personally tested and working.

Little bit problematic because of currently win implementation of Crystal::EventLoop, but works, so will simplify further development of windows fork. (raise "No runnables in scheduler" must be removed, see changes, then wait on Process result using Channel works, otherwise it hang). 

At the same time Signal API reduced to minimal Win supported signals.  Also Signal api is not so useful on Win platform.

Part of #5430.
Existing discussion to signals is here: #7339.

There is lot of necessary magic on Win platform when calling CreateProcessW and related apis..


wontruefree:
this looks like a lot of work.  Thank you for pushing forward Windows support.

straight-shoota:
This is really awesome. Thank you very much!

It's also quite a lot of code and a huge task to review. The overall design looks good to me. But we'll certainly have to talk about the public API in detail.

We should try to split this feature into smaller bites of self-contained changes, to separate commits and preferably even separate PRs. I assume some parts can be extracted and merged on their own. This makes it easier to review and discuss each individual feature without getting lost.

Skimming over it, I could think of a few ideas:

* lib_c re-orderings
* Auxiliaries such as `to_unsafew` (which is not an ideal name).
* Extract POSIX implementation of Process and Signal to a portable API
* Implement Process and Signal for Win32 (maybe this can be even further broken down?)

ysbaddaden:
Thanks for initiating this!

I think the one thing I dont like are the Signal changes. The only reason we need them with Process is to know and report when a spawned process has terminated on POSIX (SIGCHLD), which is achieved differently on Windows. The sigset is only there because of how we receive/handle signals with the EventLoop and fork/exec semantics; it's not needed on Windows.

I would leave Signal alone for the time being, not abstracting it into System (yet) and obviously not requiring it on win32. This would help to focus the pull request on Process itself for Win32. It only needs Signal on POSIX.

Last but not least, having focused commits, instead of one big commit, as proposed by @straight-shoota would be very appreciated.

jan-zajic:
@ysbaddaden, @straight-shoota thank you for first feedback. I made these changes:
* **removed** completely **Signal changes** from this PR and exclude it on Win platform (@ysbaddaden as it's currently required in Process:Status as well, see commit c704eb4)
* **removed** to_unsafew for now, it will be another PR after merge this (it will unify duplicate code that convert Crystal String to windows wide character string in dir, file, process and so on)
* split remaining stuff to multiple commits

src/compiler/crystal/compiler.cr
```diff
@@ -577,7 +577,10 @@ module Crystal
       status = yield command, args
 
       unless status.success?
-        msg = status.normal_exit? ? "code: #{status.exit_code}" : "signal: #{status.exit_signal} (#{status.exit_signal.value})"
+        msg = "code: #{status.exit_code}"
+        {% unless flag?(:win32) %}
+          msg = status.normal_exit? ? "code: #{status.exit_code}" : "signal: #{status.exit_signal} (#{status.exit_signal.value})"
```
Sija:
```suggestion
          msg = "signal: #{status.exit_signal} (#{status.exit_signal.value})" unless status.normal_exit?
```

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -175,7 +175,10 @@ module Crystal::Playground
       spawn do
         status = process.wait
         @logger.info "Code execution ended (session=#{@session_key}, tag=#{tag}, filename=#{output_filename})."
-        exit_status = status.normal_exit? ? status.exit_code : status.exit_signal.value
+        exit_status = status.exit_code
+        {% unless flag?(:win32) %}
+          exit_status = status.normal_exit? ? status.exit_code : status.exit_signal.value
```
Sija:
```suggestion
          exit_status = status.exit_signal.value unless status.normal_exit?
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+require "c/unistd"
+
+class Process
+  private def initialize(@pid)
+    @waitpid = wait(pid)
+    @wait_count = 0
+  end
+
+  def wait(pid : LibC::PidT) : Channel(Int32)
+    return Crystal::SignalChildHandler.wait(pid)
```
Sija: Redundant `return`:

```suggestion
    Crystal::SignalChildHandler.wait(pid)
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+  # Returns `true` if the process identified by *pid* is valid for
+  # a currently registered process, `false` otherwise. Note that this
+  # returns `true` for a process in the zombie or similar state.
+  def self.exists_system(pid : Int)
+    ret = LibC.kill(pid, 0)
+    if ret == 0
+      true
+    else
+      return false if Errno.value == Errno::ESRCH
+      raise Errno.new("kill")
```
Sija:
```suggestion
      Errno.new("kill")
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+  end
+
+  # Returns `true` if the process identified by *pid* is valid for
+  # a currently registered process, `false` otherwise. Note that this
+  # returns `true` for a process in the zombie or similar state.
+  def self.exists_system(pid : Int)
+    ret = LibC.kill(pid, 0)
+    if ret == 0
+      true
+    else
```
Sija: Please use early return here:

```cr
return true if ret == 0
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+    kill Signal::KILL
+  end
+
+  # See also: `Process.kill`
+  def kill(sig = Signal::TERM)
+    Process.kill sig, @pid
+  end
+
+  # Sends a *signal* to the processes identified by the given *pids*.
+  def self.kill(signal : Signal, *pids : Int)
```
Sija:
```suggestion
  def self.kill(signal : Signal, *pids : Int) : Nil
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+    pids.each do |pid|
+      ret = LibC.kill(pid, signal.value)
+      raise Errno.new("kill") if ret < 0
+    end
+    nil
```
Sija:
```suggestion
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+  end
+
+  # Creates a process, executes it
+  def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?, reader_pipe, writer_pipe)
+    if pid = Process.fork_internal(will_exec: true)
+      pid
```
Sija: Please, use early return:

```suggestion
      return pid
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,259 @@
...
+      end
+    end
+  end
+
+  # Runs the given block inside a new process and
+  # returns a `Process` representing the new child process.
+  def self.fork : Process
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      new pid
```
Sija: ditto

```suggestion
      return new pid
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+    LibC.GetCurrentThread
+  end
+
+  # Returns `true` if the process identified by *pid* is valid for
+  # a currently registered process, `false` otherwise. Note that this
+  # returns `true` for a process in the zombie or similar state.
+  def self.exists_system(pid : Int)
+    ret = LibC.kill(pid, 0)
+    if ret == 0
+      true
```
Sija: ditto

```cr
return true if ret == 0
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+    else
+      return false if Errno.value == Errno::ESRCH
+      raise Errno.new("kill")
+    end
+  end
+
+  def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?, reader_pipe, writer_pipe)
+    # like in child process
+    begin
```
Sija: redundant `begin .. end`

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+      nil,                                                                   # module name
+      to_windows_path(command_line),                                         # command line args
+      nil,                                                                   # Process handle not inheritable
+      nil,                                                                   # Thread handle not inheritable
+      LibC::TRUE,                                                            # Set handle inheritance to TRUE
+      LibC::CREATE_UNICODE_ENVIRONMENT | LibC::EXTENDED_STARTUPINFO_PRESENT, # Use UTF-16 ENV Block
+      Process.create_env_block(env, clear_env),
+      chdir ? to_windows_path(chdir) : nil, # Use chdir or parent's starting directory
+      pointerof(startupinfoex),             # Pointer to STARTUPINFOEX structure
+      pointerof(@pi)    )                   # Pointer to PROCESS_INFORMATION structure
```
Sija:
```suggestion
      pointerof(@pi)                        # Pointer to PROCESS_INFORMATION structure
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+  protected def self.on_exited(context : Void*, isTimeOut : Bool)
+    process = Box(::Process).unbox(context)
+    process.on_exited
+  end
+
+  protected def on_exited
+    @child_descriptors.each do |fd|
+      LibC._close(fd.fd) # ignore fail when already closed by child process
+    end
+    if (LibC.GetExitCodeProcess(@pi.hProcess, out exit_code) == LibC::FALSE)
```
Sija:
```suggestion
    if LibC.GetExitCodeProcess(@pi.hProcess, out exit_code) == LibC::FALSE
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+        raise WinError.new("RegisterWaitForSingleObject")
+      end
+      @handle = wait_handle
+      @pi.dwProcessId.to_i32
+    else
+      raise WinError.new("CreateProcessW")
+    end
+  end
+
+  protected def self.on_exited(context : Void*, isTimeOut : Bool)
```
Sija:
```suggestion
  protected def self.on_exited(context : Void*, _did_time_out : Bool)
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+  # An environment block can contain Unicode characters because we includes CREATE_UNICODE_ENVIRONMENT flag in dwCreationFlags
+  # A Unicode environment block is terminated by four zero bytes: two for the last string, two more to terminate the block.
+  protected def self.create_env_block(env : Env?, clear_env : Bool)
+    final_env : Env = {} of String => String
+    if LibC.CreateEnvironmentBlock(out pointer, nil, LibC::FALSE) == LibC::FALSE
+      raise WinError.new("CreateEnvironmentBlock")
+    end
+    env_block = pointer.as(Pointer(UInt16))
+    begin
+      Crystal::System::Env.parse_env_block(env_block) { |key, val|
```
Sija:
```suggestion
      Crystal::System::Env.parse_env_block(env_block) do |key, val|
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+        final_env[key] = val
+      }
```
Sija:
```suggestion
      end
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+    env.try &.each do |key, val|
+      final_env[key] = val
+    end
+    builder = WinEnvBuilder.new
+    final_env.each do |key, val|
+      add_to_env_block(builder, key, val)
+    end
+    # terminate the block
+    builder.write(0_u16)
+    return builder.buffer
```
Sija:
```suggestion
    builder.buffer
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+    data.process_id = process_id
+    LibC.EnumWindows(->Process.enum_windows_callback(LibC::HWND, LibC::LPARAM), Box.box(data).address)
+    data.window_handle
+  end
+
+  protected def self.enum_windows_callback(handle : LibC::HWND, lparam : LibC::LPARAM) : LibC::BOOL
+    data = Box(FindMainWindowParam).unbox(Pointer(Void).new(lparam))
+    process_id = LibC::DWORD.new(0)
+    LibC.GetWindowThreadProcessId(handle, pointerof(process_id))
+    if (data.process_id != process_id || !is_main_window(handle))
```
Sija:
```suggestion
    if data.process_id != process_id || !is_main_window(handle)
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+      return 1
+    else
+      data.window_handle = handle
+      return 0
+    end
+  end
+
+  protected def self.is_main_window(handle : LibC::HWND) : Bool
+    return LibC.GetWindow(handle, LibC::GW_OWNER).address == 0 && LibC.IsWindowVisible(handle) == 1
```
Sija:
```suggestion
    LibC.GetWindow(handle, LibC::GW_OWNER).address == 0 && LibC.IsWindowVisible(handle) == 1
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+  end
+
+  def self.run_ps(powershell_script : String, env : Env = nil, clear_env : Bool = false, chdir : String? = nil) : String
+    Process.run_ps(powershell_script, env: env, clear_env: clear_env, chdir: chdir) do |proc|
+      proc.output.gets_to_end
+    end
+  end
+
+  private def to_windows_path(path : String) : LibC::LPWSTR
+    return path.check_no_null_byte.to_utf16.to_unsafe
```
Sija:
```suggestion
    path.check_no_null_byte.to_utf16.to_unsafe
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+      count = slice.size
+      new_size = @wchar_size + count
+      if new_size > @capacity
+        resize_to_capacity(Math.pw2ceil(new_size))
+      end
+
+      slice.copy_to(@buffer + @wchar_size, count)
+      @wchar_size += count
+
+      nil
```
Sija:
```suggestion
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+    def write(wchar : UInt16)
+      new_size = @wchar_size + 1
+      if new_size > @capacity
+        resize_to_capacity(Math.pw2ceil(new_size))
+      end
+
+      @buffer[@wchar_size] = wchar
+      @wchar_size += 1
+
+      nil
```
Sija:
```suggestion
```

src/crystal/system/win32/process.cr
```diff
@@ -0,0 +1,401 @@
...
+        resize_to_capacity(Math.pw2ceil(new_size))
+      end
+
+      slice.copy_to(@buffer + @wchar_size, count)
+      @wchar_size += count
+
+      nil
+    end
+
+    def write(wchar : UInt16)
```
Sija:
```suggestion
    def write(wchar : UInt16) : Nil
```

src/process.cr
```diff
@@ -1,60 +1,25 @@
...
-
-  # Sends a *signal* to the processes identified by the given *pids*.
-  def self.kill(signal : Signal, *pids : Int)
-    pids.each do |pid|
-      ret = LibC.kill(pid, signal.value)
-      raise Errno.new("kill") if ret < 0
-    end
-    nil
+  def self.pid
+    return self.pid_system
```
```suggestion
    pid_system
```

src/process.cr
```diff
@@ -1,60 +1,25 @@
...
   # returns `true` for a process in the zombie or similar state.
   def self.exists?(pid : Int)
-    ret = LibC.kill(pid, 0)
-    if ret == 0
-      true
-    else
-      return false if Errno.value == Errno::ESRCH
-      raise Errno.new("kill")
-    end
+    return self.exists_system(pid)
```
```suggestion
    exists_system(pid)
```

src/process.cr
```diff
@@ -186,18 +80,28 @@ class Process
...
     rescue ex
-      process.kill
+      process.terminate_gracefully
       raise ex
     end
   end
 
+  # Terminate process gracefully
+  def terminate_gracefully
+    return self.terminate_system
```
```suggestion
    terminate_system
```

src/process.cr
```diff
@@ -186,18 +80,28 @@ class Process
...
+  end
+
+  # Terminate process immediately (kill)
+  def terminate_immediately
+    return self.kill_system
```
```suggestion
    kill_system
```

asterite:
@jan-zajic Only to add to what others already said: this is excellent, thank you so much! With efforts like these Windows support is closer and closer. I'll try to review the commits during these days if I can, though I'll admit I'm a bit clueless regarding OS processes, fork, signals, and of course the Windows API.

But at this point I think it's better to get something working and being able to bootstrap the compiler to Windows. At that point development becomes much easier and we can worry more about the small details or performance matters.

Again, thank you! ❤️❤️❤️

straight-shoota:
It would also be nice if you could add some comments explaining the reasoning behind some changes. For example when some code is excluded on `win32` or entirely removed: Why is this necessary, and is it only a temporary solution that needs improvement later, or won't it ever work/is unnecessary, or are there alternatives?

@Sija Please don't suggest circumstantial style improvements while we still haven't discussed the overall API and implementation. This is just unnecessary noise at the moment.
@jan-zajic Maybe next time submit such a PR as a draft.

straight-shoota:
The most important feature is the public API design for `Process`. There need to be some changes to remove POSIXisms and make it portable.

All documentation for public API should be directly in `src/process.cr`. The documentation needs to be universal as does the behaviour. It should mention platform-specific details, though.
We could consider moving the platform-specific implementation entirely into `Crystal::System::Process` and delegate to that from `Process` (there are already examples for this in the file API). Maybe it would be easier to just reopen `Process` and either delegate to the platform-specific methods or utilize `previous_def` for attaching the documentation. I think the latter is what I'd prefer. It avoids duplicate internal methods.

* `Process.fork` should probably be removed entirely. But that's a different story (see #6421).
* Maybe `Process.chroot` as well.
* Return type for process ids should be Int32.
* I'm not sure about the proper API for sending signals/killing a process

Relevant precursor PR: #6744

jan-zajic:
@straight-shoota, @asterite and all other - i was thinking very much about that and here (#8524) is my proposal how I think it's the best to deal with public platform dependent apis. Please let's discuss it there, before we continue our work on this Process related PR ..

straight-shoota:
Sounds good. So do we agree the system implementation should provide the following interface:
```cr
self.exit_system(status : Int32) : NoReturn
self.pid_system : Int64
self.pgid_system : Int64
self.ppid_system : Int64
self.exist_system?(pid : Int64) : Bool
self.signal_system(pid : Int64, signal : Signal) : Nil
self.terminate_system(pid : Int64) : Nil
self.kill_system(pid : Int64) : Nil
system_close : Nil
create_and_exec(*args) : Int64
self.prepare_shell(command : String, args : Enumerable(String)) : {String, Enumerable(String)}
self.wait(pid : Int64)
```
All these methods should probably be private or protected. Those with `_system` suffix should directly relate to public methods without the suffix.

One issue exists for sending signals to or kill/terminate to arbitrary processes (by process id). On POSIX this works fine one process id but the windows implementations need process handle or thread id. When there are only class methods receiving pid, the windows implementation needs to lookup the handle/thread id and can't re-use the one from an instance. Alternatively, we could implement both class methods receiving pid and instance methods which can use available information. Another solution might be to re-purpose `Process` to also function as a handle for external processes and only have instance methods for signals etc. But that would mean most of the regular API wouldn't work on instances, so it's probably not a good idea.

jan-zajic:
I can implement ppid, but please be aware that pid will be reused by other process if parent terminates and we don't open handle to the parent process. It's different from Linux behavior. 

Regarding kill methods - i'm fine with kill/terminate if we rename existing kill method and you ok with breaking change. I try named them different to just to differentiate from that current Unix kill method which means send any signal and not send kill signal in accordance with Unix kill command.

jan-zajic:
Btw @ysbaddaden i forget to say you that i really don't known about some **GetProcessEntry** windows api, it's only name of the function in go where it is implemented..  NtQueryInformationProcess must be involved. Strictly speaking, the parent process (the process which created the child process) is not actually recorded. InheritedFromUniqueProcessId just gives you the process from which attributes were inherited. This is very rarely a problem, but is controllable from the CreateProcess call, and can be something other than parent.

straight-shoota:
We can surely defer the implementation of `ppid` for win32. It doesn't need to be in this PR.

jan-zajic:
* to api proposal - system_close i change to **close_system** and wait to **wait_system** to be in line with others
* **ppid** - Implemented
* OLD **kill** - renamed to signal, but keeped only in unix/process - if i put it in win process it make dependency to **Signal** which is currently excluded as  @ysbaddaden ask at begining. So keep it only in Unix/process and make another PR on signals later, or i can bring back signal refactoring to this PR.. 
* **pgid** i currently removed from Process api and don't want to give it back - it has same semantic as fork etc - Windows don't have any process groups. In Current Win there is similar concept of **job**, but it is only on demand, ie there is no job for Process unless implicitly created and jobs have string names, not any ID's ..

According to that non-portable apis, as it is last remaining design decision necessary to do:

i currently keep **pgid**, **signal**, **fork** and **chroot** as public methods in file unix/process. They are added to reopened class Process.

We can (please decide per-method):
* **keep** it as it is. (i.e. leave it for later decision on remove..?).
* Create these methods in file /process.cr, rename them to _system in impl and implement them as **raise** in Windows version.
* **remove** them completely.  (and _maybe_ implement them in separate shard).

straight-shoota:
It's okay to not implement the methods in question for win32.
But they should be kept available for now. Since they're public API, they should be in `src/process.cr`. I'd suggest to implement them directly there behind a flag to keep the friction low. That's just meant as an interim solution until we decide later whether and how to port them for win32 or remove them.

Regarding `#signal` we should consider to postpone the rename, or rather move it out of this PR. It's not relevant for win32 implementation.

Could you document the internal interface for `Process`'s system implementation to `src/crystal/system/process.cr`? Just copy https://github.com/crystal-lang/crystal/pull/8518#issuecomment-559215010 plus your modifications. This should be similar to other files in that directory (but without a `Crystal::System` namespace type).

asterite:
Splitting this PR into commits is good, but I think it's better to split this into smaller PRs. Reviewing 31 files changed and 1000 lines of code is a task I can't handle right now, maybe it's the same for others.

Also, because it's so big many discussions are all discussed in a single place and it's not easy to follow.

jan-zajic:
Ok let's make it step by step. Closed in favor to #8536 and #8537. After we stabilize API in #8536 i will make PR's for the rest.

RX14:
I know everyone wants to get stuck into the code as soon as possible, but to prevent this in the future, please open an issue before sending a PR.

We have that in the contribution guide, but it's worth repeating: always make an issue about API changes before sending a PR. Whether you've written the code or not.

oprypin:
I don't think splitting into multiple PRs is viable. Discussion on "later" changes could affect too much in how "earlier" changes would've been submitted. And pull requests cannot be stacked (not without write access to the repository, anyway).

oprypin:
The remainder of this PR merged with master is https://github.com/oprypin/crystal/compare/winapi

I am still planning to try to get that merged, but to be fair I don't quite understand it.

FYI @kubo. If you want to take that instead, let me know in advance.

PR #8526, opened 2019-11-27
Refactor String#to_utf16 optimizing for ascii-only

straight-shoota:


wontruefree:
Do you have any benchmarks for this?

src/string/utf16.cr
```diff
@@ -12,22 +12,30 @@ class String
   # "hi 𐂥".to_utf16 # => Slice[104_u16, 105_u16, 32_u16, 55296_u16, 56485_u16]
   # ```
   def to_utf16 : Slice(UInt16)
-    size = 0
-    each_char do |char|
-      size += char.ord < 0x10000 ? 1 : 2
+    if ascii_only?
+      # size == bytesize, so each char fits in one UInt16
+      u16_size = bytesize
```
asterite: If it's ascii only, could we maybe do `to_slice.map &.ord.to_u16` or something like that? I don't know if that's faster, maybe yes (no need to compare values in the loop) but then we can bail out earlier.

straight-shoota: Yep, that's even better 👍

straight-shoota:
Benchmark for 6a36aad

string length 2000:
```
ascii new 262.98k (  3.80µs) (± 7.44%)  4.5kB/op        fastest
ascii old 179.60k (  5.57µs) (±14.67%)  4.5kB/op   1.46× slower
```
```
non-ascii new  40.85k ( 24.48µs) (±10.17%)  4.5kB/op        fastest
non-ascii old  38.74k ( 25.81µs) (± 9.81%)  4.5kB/op   1.05× slower
```

string length 10:
```
ascii new  25.78M ( 38.78ns) (±15.99%)  16.0B/op        fastest
ascii old  21.17M ( 47.25ns) (±10.44%)  32.0B/op   1.22× slower
```
```
non-ascii new   6.24M (160.36ns) (±10.87%)  32.0B/op        fastest
non-ascii old   6.06M (165.09ns) (±13.33%)  32.0B/op   1.03× slower
```

src/string/utf16.cr
```diff
@@ -12,22 +12,30 @@ class String
   # "hi 𐂥".to_utf16 # => Slice[104_u16, 105_u16, 32_u16, 55296_u16, 56485_u16]
   # ```
   def to_utf16 : Slice(UInt16)
-    size = 0
+    if ascii_only?
+      # size == bytesize, so each char fits in one UInt16
+      return to_slice.map &.to_u16
```
asterite: Oh, actually, this is missing the trailing null. I totally forgot about that! Sorry, maybe the other way was fine, or we just need to copy the values but add a trailing null somehow.

asterite: Also, are there specs for these cases? They should have failed with this.

straight-shoota: Oh, right. There should be a spec for that.

src/string/utf16.cr
```diff
@@ -12,22 +12,30 @@ class String
...
+    end
+
+    # size < bytesize, so we need to count the number of characters that are
+    # two UInt16 wide.
+    u16_size = 0
     each_char do |char|
-      size += char.ord < 0x10000 ? 1 : 2
+      u16_size += char.ord < 0x1_0000 ? 1 : 2
```
Fryguy: While I like the usage of `_`, why did you choose 4 '0's for the location of it?

straight-shoota: That's the typical grouping for hexadecimal numbers.

In this specific case it's even particularly useful because we're talking about UInt16 values which happen to fit exactly 4 hex digits.

Fryguy: Thanks for that...I've seen the `_` used a ton with decimal digits before. but I've never seen hex digits broken up, even though I did know it was possible.

src/string/utf16.cr
```diff
@@ -12,22 +12,40 @@ class String
...
+      # This is essentially equivalent to `to_slice.map(&.to_u16)` but also makes
+      # sure to allocate a null byte after the string.
+      slice = Slice(UInt16).new(bytesize + 1) do |i|
+        if i == bytesize
+          0_u16
+        else
+          unsafe_byte_at(i).to_u16
+        end
+      end
+      return slice[0, bytesize]
```
asterite: Isn't this chopping off the null byte? Why can't `slice` be returned?

straight-shoota: That's the same as the  non-ascii behaviour. The method description explicitly states that the null byte is not part of the returned slice.

asterite: Oh, interesting. I guess it's fine. A bit strange, but fine :-)

PR #8528, opened 2019-11-27
Do not collapse whitespaces in playground sidebar.

hugopl:
`"   3"` should be show as `"   3"`, not `" 3"`.

### DIFF
--- a/src/compiler/crystal/tools/playground/public/application.css
+++ b/src/compiler/crystal/tools/playground/public/application.css
@@ -132,7 +132,7 @@ h4 {
 
 .sidebar div {
   position: absolute;
-  white-space: nowrap;
+  white-space: pre;
   left: 0;
   right: 0;
 }


asterite:
Hi! Could you provide a couple of before/after screenshots?

hugopl:
Before
before
After
after

hugopl:
No idea how to auto test this without the bloat used on normal web apps testing frameworks....

asterite:
Oh, come on, testing this should be easy. You just need a couple of npm packages: https://twitter.com/thomasfuchs/status/1103700462107164672?s=20 ;-)

asterite:
Just in case (because sometimes it's hard to tell when I'm saying something funny or serious): there's no need to write tests for this. Sorry if the above came harsh for some reason.

hugopl:
nah, all good. :-)

Issue #8527, opened 2019-11-27
[RFC] Spec: add a way to check that a method's value is truthy or falsey


asterite:
Right now we can do this:

```crystal
describe Array do
  it "is empty" do
    ary = [] of Int32
    ary.empty?.should be_true
  end
end
```

That's fine but we could improve this situation.

In RSpec one can do:

```ruby
ary.should be_empty
```

This works by letting `be_empty` use `method_missing` and creating a matcher on the fly. This can't work in Crystal, not because of `method_missing` but because the generated matcher can't be aware of what's the target object being sent to it.

Some ideas that were proposed in another PR, starting from https://github.com/crystal-lang/crystal/pull/7500#issuecomment-559102417 are to be able to do something like this:

```crystal
ary.should &.empty?
```

I pointed out that the readability of that can be improved a bit:

```crystal
ary.should_be &.empty?
```

But when I went ahead to implement that I noticed in specs we have, for example:

```crystal
File.exists?(...).should be_true
```

Replacing that with `should_be` looks awkward:

```crystal
File.should_be &.exists?(...)
```

So I had that idea that we could have `assert` and `refute`, similar to other languages (but it's not the same as those other languages).

Essentially, the above two examples would be:

```crystal
ary.assert &.empty?
File.assert &.exists?(...)
```

Then we could also do:

```crystal
[1, 2, 3].refute &.empty?
````

One downside of using those names is that they don't have "should" in them. When looking at some specs right now the assertions are easy to spot because they almost always have "should" in their names (the exception being `expect_raises`).

But maybe it's fine? `assert` and `refute` are very convenient methods to have, to avoid the `should be_true` boilerplate.

The implementation is pretty straight-forward:

```crystal
require "spec"

module Spec
  module ObjectExtensions
    def assert(file = __FILE__, line = __LINE__)
      fail "Assertion failed", file, line unless yield self
    end

    def refute(file = __FILE__, line = __LINE__)
      fail "Assertion failed", file, line if yield self
    end
  end
end

describe Array do
  it "is empty" do
    ([] of Int32).assert &.empty?
  end

  it "is not empty" do
    [1].refute &.empty?
  end
end
```

Thoughts? Any other names we could use?

Blacksmoke16:
FWIW you can already do `arr.should be_empty`.

asterite:
Now that I think about it, this could also work as a top-level method:

```crystal
assert ary.empty?
assert File.exists?(...)
```

[...]

Issue #3095, opened 2016-08-04
Replace expectations in spec with power asserts


RX14:
After using [groovy](http://www.groovy-lang.org/) for a while, and loving the [power assert built into the langauge](http://docs.groovy-lang.org/next/html/documentation/core-testing-guide.html#_power_assertions), and also using the wonderful [power_assert.cr](https://github.com/rosylilly/power_assert.cr) by @rosylilly, I wanted to propose adding power assert to crystal's stdlib. I believe that powerful assert should replace the expectations dsl in spec because it is cleaner, and can show you more information, which is helpful to debugging.

Powerful asserts consist of `assert` before an expression. When the expression returns false at runtime, an exception is generated containing a deconstruction of the statement with debug information on every part. For example:
```cr
foo = "foo"
assert foo.index('o') == 3
```
```txt
foo.index('o') == 3
|   |          |
|   |          false : Bool
|   1 : Int32
"foo" : String
```
Compare this too the default spec expectations:
```cr
foo = "foo"
foo.index('o').should eq(3)
```
```txt
expected: 3
     got: 1
```
In the power assert example you get much more context to the spec failure than you do in the expectations example, by slowing the line of code that failed and the values at various parts of the expression. This deconstruction grants you visibility into exactly what went wrong, which lets you debug faster.

The syntax is also clearer than the expectation syntax, because it's a normal crystal expression instead of having to learn a new DSL to express yourself with. There are less brackets, it's cleaner. For example, the code samples above. Finding an `==` is much easier visually than finding `.should eq(` in an expression consisting of method calls.

All of the expectation methods are easilly translatable to normal crystal expressions, except `#close`, and in this case I would argue that `Float#close?` should exist to enable fuzzy matching outside of specs.

@BlaXpirit has converted a whole file of stdlib specs to assertion syntax, in a gist [here](https://gist.github.com/BlaXpirit/f63ec5206e2036a25164d7a91336df67). There are a few warts in power_assert.cr, mainly to do with lacking proper resugaring or not showing types, visible [here](https://gist.github.com/BlaXpirit/72ef9c211749b00faa72784091998d64). These can be fixed before merging in.

It should be possible to implement a crystal formatter to convert expectation syntax into assert syntax quite easilly. Expectations could be deprecated, and power assertions and the formatter changes could be released in 0.19.0, followed by removing expectations in 0.20.0.

asterite:
Compare this:

```cr
require "spec"

describe "something" do
[...]

PR #8529, opened 2019-11-28
Some compiler optimizations

asterite:
Some optimizations from musings in https://github.com/crystal-lang/crystal/issues/8527#issuecomment-559300111

The first two commits are self-explanatory. The third commits has a big commit message explaining the description of the problem. After reading that you can understand how it applies to this particular example:

```crystal
require "spec"

macro assert(call, file = __FILE__, line = __LINE__)
  {% obj = call.receiver %}
  {% name = call.name %}
  %obj = {{obj}}
  %call = %obj.{{name}}
  unless %call
    fail "Expected #{%obj} to be {{name[0...-1]}}"
  end
end

describe "Foo" do
  {% for i in 0..12_000 %}
    it "bar {{i}}" do
      a = [1, 2, 3]
      assert a.empty?
    end
  {% end %}
end
```

In the above code the `assert` macro will define the local variables `%obj` and `%call` which will actually expand to variables named something like `__temp1` and so on, and they'll all be different. Because of the above compiler bug, all these local variables will be seen in each `assert` macro invocation. So the first `assert` expansion can only see the `a` local variable. The second one will see that plus the two local variables introduced by the previous macro invocation (which shouldn't be seen, but that's the bug the third commit is fixing). On each invocation a `Set` is created with these local vars. That has something like quadratic complexity which explains why the above code was taking 12 seconds. Now it takes 0.4s (well, at least the "main" semantic phase).

I always thought macro expansion was slow, but apparently it's not. Or at least not that slow.

Once we merge this PR we can see if #3095 is still slow for the [snippet](https://github.com/crystal-lang/crystal/issues/3095#issuecomment-237614293) I found it to be slow.

### DIFF
--- a/src/compiler/crystal/semantic/semantic_visitor.cr
+++ b/src/compiler/crystal/semantic/semantic_visitor.cr
@@ -323,7 +323,11 @@ abstract class Crystal::SemanticVisitor < Crystal::Visitor
                Parser::ParseMode::Normal
              end
 
-    generated_nodes = @program.parse_macro_source(expanded_macro, macro_expansion_pragmas, the_macro, node, Set.new(@vars.keys),
+    # We could do Set.new(@vars.keys) but that creates an intermediate array
+    local_vars = Set(String).new(initial_capacity: @vars.size)
+    @vars.each_key { |key| local_vars << key }
+
+    generated_nodes = @program.parse_macro_source(expanded_macro, macro_expansion_pragmas, the_macro, node, local_vars,
       current_def: @typed_def,
       inside_type: !current_type.is_a?(Program),
       inside_exp: @exp_nest > 0,
--- a/src/compiler/crystal/semantic/top_level_visitor.cr
+++ b/src/compiler/crystal/semantic/top_level_visitor.cr
@@ -938,7 +938,8 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
   end
 
   def visit(node : Block)
-    old_vars_keys = @vars.keys
+    # Remember how many local vars we had before the block
+    old_vars_size = @vars.size
 
     # When accepting a block, declare variables for block arguments.
     # These are needed for macro expansions to parser identifiers
@@ -949,9 +950,10 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
     node.body.accept self
 
-    # Now remove these vars, but only if they weren't vars before
-    node.args.each do |arg|
-      @vars.delete(arg.name) unless old_vars_keys.includes?(arg.name)
+    # After the block we should have the same number of local vars
+    # (blocks can't declare inject local vars to the outer scope)
+    while @vars.size > old_vars_size
+      @vars.delete(@vars.last_key)
     end
 
     false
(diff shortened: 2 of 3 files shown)


asterite:
Well, actually, it seems power_assert is still slow to compile but because it generates a ton of code. I think the `assert` could be made simpler, I'll see.

src/compiler/crystal/semantic/semantic_visitor.cr
```diff
@@ -323,7 +323,11 @@ abstract class Crystal::SemanticVisitor < Crystal::Visitor
                Parser::ParseMode::Normal
              end
 
-    generated_nodes = @program.parse_macro_source(expanded_macro, macro_expansion_pragmas, the_macro, node, Set.new(@vars.keys),
+    # We could do Set.new(@vars.keys) but that creates an intermediate array
```
wontruefree: did you intend to leave this?

wontruefree:
Do you have any benchmarks for this?

vlazar:
> Well, actually, it seems power_assert is still slow to compile but because it generates a ton of code. I think the `assert` could be made simpler, I'll see.

Maybe [this simpler generated assert code](https://github.com/crystal-lang/crystal/issues/3095#issuecomment-237660749) proposed by RX14 might help

straight-shoota:
I suppose there's a mistake in the example code from the third commit? Or else I just don't get it.

asterite:
Ah, yes, there's a mistake. It should be `puts id(a)`. I'll fix it later.

The benchmark is the snippet with that assert. Compile it with the existing compiler and with a compiler with this PR and you'll see a difference (of about 12 seconds less on my machine).

src/compiler/crystal/semantic/top_level_visitor.cr
```diff
@@ -949,9 +950,10 @@ class Crystal::TopLevelVisitor < Crystal::SemanticVisitor
 
     node.body.accept self
 
-    # Now remove these vars, but only if they weren't vars before
-    node.args.each do |arg|
-      @vars.delete(arg.name) unless old_vars_keys.includes?(arg.name)
+    # After the block we should have the same number of local vars
+    # (blocks can't declare inject local vars to the outer scope)
+    while @vars.size > old_vars_size
```
yxhuvud: Would there be any difference if this was `(old_vars_size - @vars.size).times do` ? I doubt it is noticeable but the method is called a lot of times, so who knows? 

I wish we had an ordered map type, then we could have just nuked the last n entries in a more efficient manner.

EDIT: Hmm. Isn't the insertion order kept track of? Perhaps it is possible to do bulk operations. :thinking: 

/probably not worth the trouble though

asterite: All excellent observations. Yeah, Hash is ordered and so is Set so we could nuke the last entries, though because it's a Hash we still need to find the key hash to see in which bucket it's located, clear it and so on, so probably the same performance in the end.

asterite:
I'd merge but there's no newer "milestone", so I guess we can wait until that exists... or we create one now? Can we merge before 0.32.0 is released?

bcardiff:
Merging it for 0.32 is fine on my end. As long as there are no breaking changes :-)

asterite:
Cool, let's do it!

Issue #651, opened 2015-05-10
method naming conventions


Vikaton:
I have seen a few methods in Crystal that are borrowed from Ruby, but some of the method names are unnecessarily short/unreadable

So far I have seen only 2, `String.tr()` and `Int32.chr`

For those 2, .tr can be .replace or something a bit more recognizable and .chr should be .char, so it can be much more readable.

This is just a suggestion, and have only found 2 methods so far in the course of writing my code.

jhass:
I think `tr` is actually a well known name used across many languages, most prominently it's present in the Gnu coreutils.

Vikaton:
method names can stay as they are if you see that they have good reason for their shotnames. if what you said is true than sure. Though I don't see why .char is .chr.

waterlink:
Both methods (`tr` and `chr`) are available in Ruby ;)

--
Best Regards,
Alexey Fedorov,
Sr Ruby Developer, Clojure Developer, Rust Developer,
+49 15757 486 476

2015-05-11 1:53 GMT+03:00 Unnamed <notifications@github.com>:

Vikaton:
Yes I understand that, but I'm doing this in Crystal's perspective.

These are just slight method naming conventions that can make Crystal better in my opinion for those who don't code in Ruby :)

In a situation like .chr, I would rather have it readable then making it as short as possible

asterite:
I think at this point I'd like to preverse as much compatibility with Ruby as possible. In fact, maybe using `includes?` instead of `include?` was a mistake (we still have time to change this). If someone knows Ruby and comes to Crystal, they already know most of the names. If not, they don't know any name so learning "tr" vs. "replace" or "chr" vs. "char", or "includes?" vs. "include?" is the same for them. So preserving Ruby names should be something to think about.

On the other hand, I never really liked `attr_accessor`, `attr_reader` and `attr_writer`. In every other language they are called property, getter and setter.

@Vikaton Would you also change `puts` and `gets` to more readable, but probably longer, names?

asterite:
There's also the thing that `if "foo".includes?("o")` reads like English, but `if "foo".include?("o")` reads like Tarzan. On the other hand `array.includes?` reads well too, but the variable is probably named `elements` or `children` so `elements.includes?` reads wrong. And having aliases for these is not a good thing...

[...]

PR #8533, opened 2019-11-29
Deprecate char#ord and int#chr

wdhwg001:
As mentioned in #8449 , they are now `Char#codepoint` and `Int#char` in this pull request.

The background of these two cryptic methods is that they actually come from an old programming language, Pascal.

http://xoomer.virgilio.it/gciabu/engl/pascal/pas036.htm

The reason of having int#char instead of `Int#to_char` or others is because when we say `1.char`, it looks more like "accessing a `char` property of a number" or "getting a number as char form".

By contrast, `1.to_char` and other `to_x` ideas look like "convert the number to a char", which may cause confusion with `Int#to_s`.

`Char#codepoint` has the similar idea: to avoid the use of `to_x`, to prevent the confusion with `Char#to_i`, and to keep it slightly shorter compared to `to_codepoint`.

For the merging, I would suggest that we either merge this before the release to deprecate `String#codepoint_at`, `Char#ord` and `Int#chr` together, or we deprecate `String#codepoint_at` first, and then deprecate the other two in future releases. I personally like to deprecate it together, because it will avoid one refactor.

Sorry for creating this pull request before any further discussions. The original issue is quite inactive, so we may continue discussions here.

It's my first commit to Crystal, any ideas or suggestions will be much appreciated. Considering that `Char#ord` and `Int#chr` are widely used in the source code, I might miss some points.

### DIFF
--- a/samples/brainfuck.cr
+++ b/samples/brainfuck.cr
@@ -42,7 +42,7 @@ class Program
       when '<'; tape.devance
       when '+'; tape.inc
       when '-'; tape.dec
-      when '.'; print tape.get.chr
+      when '.'; print tape.get.char
       when '['; pc = @bracket_map[pc] if tape.get == 0
       when ']'; pc = @bracket_map[pc] if tape.get != 0
       end
--- a/samples/compiler/transformer_example.cr
+++ b/samples/compiler/transformer_example.cr
@@ -2,14 +2,14 @@
 # to transform source code.
 #
 # Here we transform all number literals with their char
-# equivalent using `chr`.
+# equivalent using `char`.
 
 # Use `require "compiler/crystal/syntax"` in your programs
 require "../../src/compiler/crystal/syntax"
 
 class Charify < Crystal::Transformer
   def transform(node : Crystal::NumberLiteral)
-    Crystal::CharLiteral.new(node.value.to_i.chr)
+    Crystal::CharLiteral.new(node.value.to_i.char)
   end
 end
 
--- a/samples/meteor.cr
+++ b/samples/meteor.cr
@@ -119,7 +119,7 @@ def to_utf8(raw_sol)
     raw_sol.each do |m|
       id = get_id(m)
       50.times do |i|
-        buf[i] = '0'.ord.to_u8 + id if bm(m, i) != 0
+        buf[i] = '0'.codepoint.to_u8 + id if bm(m, i) != 0
       end
     end
     {50, 50}
--- a/samples/meteor.cr
+++ b/samples/meteor.cr
@@ -132,7 +132,7 @@ def print_sol(str)
     puts if i % 5 == 0
     print ' ' if (i + 5) % 10 == 0
     print ' '
-    print c.chr
+    print c.char
     i += 1
   end
   puts
--- a/samples/sudoku.cr
+++ b/samples/sudoku.cr
@@ -70,7 +70,7 @@ def sd_solve(mr, mc, s)
   ret = [] of Array(Int32)
   sr, sc, hints = Array.new(729, 0), Array.new(324, 9), 0
   (0...81).each do |i|
-    a = ('1' <= s[i] <= '9') ? s[i].ord - 49 : -1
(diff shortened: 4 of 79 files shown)


wdhwg001:
test_preview_mt failed with:
```
crystal spec spec/std/concurrent/select_spec.cr:135 # select select fiber has one chance to be enqueued into scheduler (2)
```

I'm not sure what this issue is, it seems unrelated?

oprypin:
I did not see agreement on the change. You could argue that writing it like this from the start would have been good. But changing it is not worth it.
https://github.com/crystal-lang/crystal/issues/8449#issuecomment-557120526

We can't even clearly discount `chr`/`ord` -- of course they have familiarity not just from Pascal but also Python and Ruby, at least

And I want to clarify, here you're *adding* Char#codepoint and Int#char, not only deprecating as per title.

ysbaddaden:
I don't have an opinion on the renames. `ord` and `chr` are well-known, `codepoint` and `char` are more explicit. It's a matter of habit vs taste.

Yet, please, make distinct commits: one for adding `Char#codepoint` + deprecating `#ord` (+ unsafe methods), then another one for `Int#chr`, and eventually two commits for changing `#ord` and `#chr` usages in the repository. It will be easier to review and help keep a clean history.

straight-shoota:
I agree with @ysbaddaden Changing this sounds nice, but is it worth the friction? Maybe in the long run...

Anyway, I would recommend to not apply the suggested changes (although good) right away. We should figure out what we really want first. It would've been better to have a dedicated discussion, not 
just piggybacked on another one.

wdhwg001:
Sorry for having a bad commit history, I'll try to improve it next week as I'm having an exam on Monday.

The initial discussion actually starts in #651 . At that time, compatibility with Ruby is still important, but for now, things have changed.

A short history of `chr` and `ord` is that they came from Pascal, and were borrowed to Perl, and then were taken to Python, PHP and Ruby. They were designed before the birth of Unicode, so the `ord` actually means `ASCII Order` as its original meaning.

Furthermore, at that time, parsing syntax is still a huge work, the input latency is high, the monitor is small, and the storage for the code is highly limited. So at that time, people prefer to omit as many letters as possible to keep the code shorter. It's totally different now.

And actually, Perl, PHP, Python and Ruby are nearly the only successors of these functions. Modern languages, even including Java, do not have these ancient and cryptic function names.

IMO, having such conventions is less important compared to the improvement of readability. Crystal is far beyond a "faster ruby" now, and is always considered as a totally redesigned modern language. There will be more newcomers in the future, so I would suggest that we rename those cryptic names before reaching 1.0. We are not "too late" yet.

oprypin:
You're putting too much effort into discrediting this naming. But it still all can be summarized as "old is bad" and "short is bad". And that's just not enough to inflict breakage on everyone. It's not about 1.0, it's just not worth it regardless.

I think this naming has outgrown the familiarity just within these languages. People just learn it like that. They can be terms of their own. I bet even in pop culture there are some references.

If `ord` can stand for "ordinal of a codepoint", that's fine by me.

At least it's not specific to something like OS... Because we have `grep` as a function name, anyway. I can't even...

straight-shoota:
If we keep `String#codepoint_at` (#8449), I think we should definitely rename the Char method to match that name (probably `#as_codepoint` as sugggested by @RX14).

wdhwg001:
The thing I worried about is that `as_x` is only used in XML, YAML and JSON to cast `Any` to specific types. It's more like a terminology that means to cast a unknown/generic type to a specific one (compared to `as(to)`, which is a terminology that means to cast a Union type to a specific one).

And if we rename it to `#as_codepoint` instead of `#codepoint`, should we rename `#bytes` to `#as_bytes` for consistency? I therefore think that using `as_x` when `x` is not a type but a conventional name might not be a great idea.

straight-shoota:
Closing stale PR. This proposed change has gone completely out of sync and would need a reboot.
But there doesn't seem to be a convincing drive to even change this because the benefit-cost ratio is questionable.

PR #8535, opened 2019-11-29
Add VaList#next

ffwff:
This patch adds `VaList#next(type)` for getting the next element in a variadic argument list (ala va_arg), it also fixes a linking bug when `Intrinsics.va_start`, `Intrinsics.va_end` is used.

Quick and dirty example usage:

```crystal
{% `
cat <<C >foo.c
int foo_f(int,...);
void foo() {
  foo_f(3,1,2,3);
}
C
gcc -c -o foo.o foo.c
` %}
@[Link(ldflags: "#{__DIR__}/foo.o")]
lib Foo
  fun foo()
end

fun foo_f(count : Int32, ...)
  VaList.open do |list|
    count.times do |i|
      puts list.next(Int32)
    end
  end
end

`rm -rf foo.o foo.c`

Foo.foo # prints out 1, 2, 3
```


spec/compiler/codegen/primitives_spec.cr
```diff
@@ -237,4 +237,19 @@ describe "Code gen: primitives" do
...
+        @[Primitive(:va_arg)]
+        def next(type)
+        end
+      end
+
+      list = VaList.new
+      list.next(Int32)
+      ))
+    str = mod.to_s
+    str.should contain("va_arg %VaList* %list")
```
RX14: Is there perhaps an executable test which you can do to make sure the whole `VaList` aparatus is working?

for example, pass multiple args into a function and make sure they're accessed correctly

ffwff: Should I do it like the example code above? It'd have to generate C code, link the crystal program with it and run it.

RX14: Why would you have to generate C code? You can call var-arged functions from crystal.

ffwff: Nope, crystal can't call vararg `fun` functions, it just spouts out "wrong number of arguments" when doing so

RX14: oh, well this is fine for now then.

Would be nice if crystal would support that in the future.

RX14:
ping

spec/compiler/codegen/primitives_spec.cr
Sija: I'm pretty sure it was working...

RX14: @Sija you can call vararg `fun`s in `lib`s, but not `fun`'s with their bodies written in crystal

src/va_list.cr
```diff
@@ -17,4 +17,8 @@ struct VaList
       Intrinsics.va_end pointerof(ap)
     end
   end
+
+  @[Primitive(:va_arg)]
```
bcardiff: This needs to be defined in `{% if compare_versions(Crystal::VERSION, "0.33.0-0") > 0 %}`

Otherwise, the std-lib will not build in 0.32.1 since the new primitive is not available yet.

src/intrinsics.cr
```diff
@@ -156,12 +156,12 @@ module Intrinsics
     LibIntrinsics.counttrailing128({{src}}, {{zero_is_undef}})
   end
 
-  def self.va_start(ap)
-    LibIntrinsics.va_start(ap)
+  macro va_start(ap)
```
bcardiff: Why is this change needed? When LibIntrinsic and module Intrinsic got split did that broke `VaList`?

bcardiff:
Please, rebase on master.

src/intrinsics.cr
ffwff: Using a function for va_start generates an error:

```
Invalid memory access (signal 11) at address 0x18
[0x558b0e4b6b36] *CallStack::print_backtrace:Int32 +118                                       
[0x558b0e49e831] __crystal_sigfault_handler +209                                              
[0x558b0f4c0cea] sigfault_handler +40                                                         
[0x7fd82d64b930] ???                                                                          
[0x7fd82e8e5733] _ZNK4llvm14MachineOperand5printERNS_11raw_ostreamERNS_17ModuleSlotTrackerENS_3LLTEbbbjPKNS_18TargetRegisterInfoEPKNS_19TargetIntrinsicInfoE +739
[0x7fd82e8c488c] _ZNK4llvm12MachineInstr5printERNS_11raw_ostreamERNS_17ModuleSlotTrackerEbbbbPKNS_15TargetInstrInfoE +1164
[0x7fd82e8c69e1] _ZNK4llvm12MachineInstr5printERNS_11raw_ostreamEbbbbPKNS_15TargetInstrInfoE +177
[0x7fd830c4dd08] ???                                                                          
[0x7fd830c4e2ae] ???                                                                          
[0x7fd82ed5cb31] _ZN4llvm10AsmPrinter16EmitFunctionBodyEv +561                                
[0x7fd830a66f8d] ???                                                                          
[0x7fd82e8bb535] _ZN4llvm19MachineFunctionPass13runOnFunctionERNS_8FunctionE +181
[0x7fd82e691530] _ZN4llvm13FPPassManager13runOnFunctionERNS_8FunctionE +1248
[0x7fd82e6915b2] _ZN4llvm13FPPassManager11runOnModuleERNS_6ModuleE +98                        
[0x7fd82e690836] _ZN4llvm6legacy15PassManagerImpl3runERNS_6ModuleE +870
[0x7fd82fd9c76f] ???                                                                          
[0x7fd82fd9c894] LLVMTargetMachineEmitToFile +244
[0x558b0f1d4ee9] *LLVM::TargetMachine#emit_to_file<LLVM::Module, String, LLVM::CodeGenFileType>:Bool +89
[0x558b0f1d4f2b] *LLVM::TargetMachine#emit_obj_to_file<LLVM::Module, String>:Bool +11
[0x558b0f385f37] *Crystal::Compiler::CompilationUnit#compile_to_object:(LLVM::Module | Nil) +887
[0x558b0f385a63] *Crystal::Compiler::CompilationUnit#compile:(LLVM::Module | Nil) +51
[0x558b0e49ad3d] ~procProc(Nil) +637
[0x558b0e5a00a1] *Fiber#run:(IO::FileDescriptor | Nil) +113
[0x558b0e4986b6] ~procProc(Fiber, (IO::FileDescriptor | Nil)) +6
[0x0] ???                                      
```

RX14: @bcardiff these intrinsics implicitly relate to the surrounding function. Putting them in a different function won't work.

bcardiff: I suspect that. I wasn't aware there was no test coverage for this module.

bcardiff:
I would still want a spec to be added in spec/manual with a code similar to the OC. Other than that LGTG.

spec/manual/valist_spec.cr
```diff
@@ -0,0 +1,29 @@
+require "spec"
+require "../compiler/spec_helper"
+
+it "VaList works with C code" do
```
RX14: Does this need to be a manual spec at this point @bcardiff ?

bcardiff:
Thanks @ffwff !

RX14:
@bcardiff did you mean to merge that with the spec still in `spec/manual`?

bcardiff:
Good catch. No, I missed it was in manual since the code looked like a normal spec. It can be moved since there are other specs using `test_c`.

RX14:
I'll move it to `primatives_spec`

PR #8536, opened 2019-11-29
Extract POSIX implementation of Process and Signal to a portable API

jan-zajic:
Prepare Process API for Windows implementation. After discussion in #8518 i extract the system implementation which provide 
the following private interface (all methods are protected):
```
# for current running program
self.exit_system(status : Int32) : NoReturn
self.pid_system : Int64
self.ppid_system : Int64
self.fork_system(&) : Process
self.fork_system : Process?
self.chroot_system(path : String) : Nil
self.times_system : Tms

# for Process instance
terminate_system : Nil
kill_system : Nil
close_system: Nil
exist_system? : Bool
signal_system(signal : Signal) : Ni
wait_system : Nil
pgid_system : Int64
create_and_exec(command, *args, ...) : Int64

# for given pid
self.exists_system?(pid : Int64)
self.pgid_system(pid : Int64) : Int64

# util
self.prepare_shell_system(command : String, args : Enumerable(String)) : {String, Enumerable(String)}
```

Public API design for Process
There need to be some changes to remove POSIXisms and make it portable.
* ~~kill method is renamed to signal~~ move to #8642
* new methods terminate (graceful), interrupt (immediately)

Precursor PR: #8518

### DIFF
--- a/src/crystal/system/unix/process.cr
+++ b/src/crystal/system/unix/process.cr
@@ -1,0 +1,254 @@
+require "c/signal"
+require "c/stdlib"
+require "c/sys/resource"
+require "c/unistd"
+
+class Process
+  protected def wait_system
+    Crystal::SignalChildHandler.wait(@waitpid, @pid)
+  end
+
+  protected def self.exit_system(status = 0) : NoReturn
+    LibC.exit(status)
+  end
+
+  protected def self.pid_system : Int64
+    LibC.getpid.to_i64
+  end
+
+  protected def self.ppid_system : Int64
+    LibC.getppid.to_i64
+  end
+
+  protected def exists_system?
+    Process.exists_system?(@pid)
+  end
+
+  protected def self.exists_system?(pid : Int64)
+    ret = LibC.kill(pid, 0)
+    return true if ret == 0
+    return false if Errno.value == Errno::ESRCH
+    raise Errno.new("kill")
+  end
+
+  protected def pgid_system : Int64
+    Process.pgid_system(@pid)
+  end
+
+  protected def self.pgid_system(pid : Int64) : Int64
+    ret = LibC.getpgid(pid)
+    raise Errno.new("getpgid") if ret < 0
+    ret.to_i64
+  end
+
+  protected def signal_system(signal : Signal)
+    ret = LibC.kill(@pid, signal.value)
+    raise Errno.new("kill") if ret < 0
+  end
+
+  protected def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?)
+    reader_pipe, writer_pipe = IO.pipe
+
+    pid = create_and_exec_impl(command, args, env, clear_env, fork_input, fork_output, fork_error, chdir, reader_pipe, writer_pipe)
+    writer_pipe.close
+    bytes = uninitialized UInt8[4]
+    if reader_pipe.read(bytes.to_slice) == 4
+      errno = IO::ByteFormat::SystemEndian.decode(Int32, bytes.to_slice)
+      message_size = reader_pipe.read_bytes(Int32)
+      if message_size > 0
+        message = String.build(message_size) { |io| IO.copy(reader_pipe, io, message_size) }
+      end
+      reader_pipe.close
+      raise Errno.new(message, errno)
+    end
+    reader_pipe.close
+    pid
+  end
+
+  protected def create_and_exec_impl(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?, reader_pipe : IO::FileDescriptor, writer_pipe : IO::FileDescriptor)
+    if pid = Process.fork_internal(will_exec: true)
+      return pid
+    end
+    begin
+      reader_pipe.close
+      writer_pipe.close_on_exec = true
+      Process.exec_internal(command, args, env, clear_env, fork_input, fork_output, fork_error, chdir)
+    rescue ex : Errno
+      writer_pipe.write_bytes(ex.errno)
+      writer_pipe.write_bytes(ex.message.try(&.bytesize) || 0)
+      writer_pipe << ex.message
+      writer_pipe.close
+    rescue ex
+      ex.inspect_with_backtrace STDERR
+      STDERR.flush
+    ensure
+      LibC._exit 127
+    end
+  end
+
+  protected def self.fork_system : Process
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      return new pid
+    end
+    begin
+      yield
+      LibC._exit 0
+    rescue ex
+      ex.inspect_with_backtrace STDERR
+      STDERR.flush
+      LibC._exit 1
+    ensure
+      LibC._exit 254 # not reached
+    end
+  end
+
+  protected def self.fork_system : Process?
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      new pid
+    end
+  end
+
+  ORIGINAL_STDIN  = IO::FileDescriptor.new(0, blocking: true)
+  ORIGINAL_STDOUT = IO::FileDescriptor.new(1, blocking: true)
+  ORIGINAL_STDERR = IO::FileDescriptor.new(2, blocking: true)
+
+  protected def self.reopen_io(src_io : IO::FileDescriptor, dst_io : IO::FileDescriptor)
+    src_io = to_real_fd(src_io)
+    dst_io.reopen(src_io)
+    dst_io.blocking = true
+    dst_io.close_on_exec = false
+  end
+
+  protected def self.to_real_fd(fd : IO::FileDescriptor)
+    case fd
+    when STDIN  then ORIGINAL_STDIN
+    when STDOUT then ORIGINAL_STDOUT
+    when STDERR then ORIGINAL_STDERR
+    else             fd
+    end
+  end
+
+  protected def self.exec_internal(command, args : (Array | Tuple)?, env, clear_env, input, output, error, chdir) : NoReturn
+    reopen_io(input, ORIGINAL_STDIN)
+    reopen_io(output, ORIGINAL_STDOUT)
+    reopen_io(error, ORIGINAL_STDERR)
+
+    ENV.clear if clear_env
+    env.try &.each do |key, val|
+      if val
+        ENV[key] = val
+      else
+        ENV.delete key
+      end
+    end
+
+    Dir.cd(chdir) if chdir
+
+    argv = [command.check_no_null_byte.to_unsafe]
+    args.try &.each do |arg|
+      argv << arg.check_no_null_byte.to_unsafe
+    end
+    argv << Pointer(UInt8).null
+
+    LibC.execvp(command, argv)
+
+    error_message = String.build do |io|
+      io << "execvp ("
+      command.inspect_unquoted(io)
+      args.try &.each do |arg|
+        io << ' '
+        arg.inspect(io)
+      end
+      io << ")"
+    end
+    raise Errno.new(error_message)
+  end
+
+  protected def self.prepare_shell_system(command, args)
+    command = %(#{command} "${@}") unless command.includes?(' ')
+    shell_args = ["-c", command, "--"]
+
+    if args
+      unless command.includes?(%("${@}"))
+        raise ArgumentError.new(%(can't specify arguments in both, command and args without including "${@}" into your command))
+      end
+
+      {% if flag?(:freebsd) %}
+        shell_args << ""
+      {% end %}
+
+      shell_args.concat(args)
+    end
+
+    command = "/bin/sh"
+    {command, shell_args}
+  end
+
+  # :nodoc:
+  protected def self.fork_internal(*, will_exec : Bool) : Int64?
+    newmask = uninitialized LibC::SigsetT
+    oldmask = uninitialized LibC::SigsetT
+
+    LibC.sigfillset(pointerof(newmask))
+    ret = LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(newmask), pointerof(oldmask))
+    raise Errno.new("Failed to disable signals") unless ret == 0
+
+    case pid = LibC.fork
+    when 0
+      # child:
+      pid = nil
+      if will_exec
+        # reset signal handlers, then sigmask (inherited on exec):
+        Crystal::Signal.after_fork_before_exec
+        LibC.sigemptyset(pointerof(newmask))
+        LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(newmask), nil)
+      else
+        {% unless flag?(:preview_mt) %}
+          ::Process.after_fork_child_callbacks.each(&.call)
+        {% end %}
+        LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
+      end
+    when -1
+      # error:
+      errno = Errno.value
+      LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
+      raise Errno.new("fork", errno)
+    else
+      # parent:
+      LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
+    end
+
+    pid.try &.to_i64
+  end
+
+  protected def close_system
+  end
+
+  protected def self.chroot_system(path : String) : Nil
+    path.check_no_null_byte
+    if LibC.chroot(path) != 0
+      raise Errno.new("Failed to chroot")
+    end
+
+    if LibC.chdir("/") != 0
+      errno = Errno.new("chdir after chroot failed")
+      errno.callstack = CallStack.new
+      errno.inspect_with_backtrace(STDERR)
+      abort("Unresolvable state, exiting...")
+    end
+  end
+
+  protected def self.times_system : Tms
+    LibC.getrusage(LibC::RUSAGE_SELF, out usage)
+    LibC.getrusage(LibC::RUSAGE_CHILDREN, out child)
+
+    Tms.new(
+      usage.ru_utime.tv_sec.to_f64 + usage.ru_utime.tv_usec.to_f64 / 1e6,
+      usage.ru_stime.tv_sec.to_f64 + usage.ru_stime.tv_usec.to_f64 / 1e6,
+      child.ru_utime.tv_sec.to_f64 + child.ru_utime.tv_usec.to_f64 / 1e6,
+      child.ru_stime.tv_sec.to_f64 + child.ru_stime.tv_usec.to_f64 / 1e6,
+    )
+  end
+end
--- a/src/crystal/system/unix/process.cr
+++ b/src/crystal/system/unix/process.cr
@@ -24,6 +24,14 @@ class Process
     Process.exists_system?(@pid)
   end
 
+  protected def terminate_system
+    signal_system Signal::TERM
+  end
+
+  protected def interrupt_system
+    signal_system Signal::KILL
+  end
+
   protected def self.exists_system?(pid : Int64)
     ret = LibC.kill(pid, 0)
     return true if ret == 0
--- a/src/process.cr
+++ b/src/process.cr
@@ -1,7 +1,4 @@
-require "c/signal"
-require "c/stdlib"
-require "c/sys/resource"
-require "c/unistd"
+require "crystal/system/process"
 
 class Process
   # Terminate the current process immediately. All open files, pipes and sockets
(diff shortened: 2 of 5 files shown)


src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,243 @@
...
+  protected def self.pid_system : Int64
+    LibC.getpid.to_i64
+  end
+
+  protected def self.ppid_system : Int64
+    LibC.getppid.to_i64
+  end
+
+  protected def exists_system?
+    ret = LibC.kill(@pid.to_i32, 0)
```
ysbaddaden: It should be a class method `self.exists_system(pid)` that could be called directly from `Process.exists?(pid)` for example, avoiding the instantiation of a Process object.

Also don't manually cast values to LibC calls: the compiler does it automatically.

src/process.cr
```diff
@@ -1,60 +1,51 @@
...
+
+  protected def initialize(@pid)
+    open_system
+    wait_system
+    @wait_count = 0
+  end
+
+  # Open process by identifier
+  #
+  # check that process with given pid exists and is acessible
```
ysbaddaden:
```suggestion
  # Checks that process with given *pid* exists and is accessible.
```

src/process.cr
```diff
@@ -1,60 +1,51 @@
...
+
   # Terminate the current process immediately. All open files, pipes and sockets
   # are flushed and closed, all child processes are inherited by PID 1. This does
   # not run any handlers registered with `at_exit`, use `::exit` for that.
   #
   # *status* is the exit status of the current process.
-  def self.exit(status = 0)
-    LibC.exit(status)
+  def self.exit(status = 0) : NoReturn
+    ::Process.exit_system(status)
```
ysbaddaden:
```suggestion
    exit_system(status)
```

src/process.cr
```diff
@@ -1,60 +1,51 @@
...
-    pgid(0)
-  end
-
-  # Returns the process group identifier of the process identified by *pid*.
-  def self.pgid(pid : Int32) : LibC::PidT
-    ret = LibC.getpgid(pid)
-    raise Errno.new("getpgid") if ret < 0
-    ret
+  def self.pid : Int64
+    self.pid_system
```
ysbaddaden:
```suggestion
    pid_system
```

src/process.cr
```diff
@@ -1,60 +1,51 @@
...
-  def self.exists?(pid : Int)
-    ret = LibC.kill(pid, 0)
-    if ret == 0
-      true
-    else
-      return false if Errno.value == Errno::ESRCH
-      raise Errno.new("kill")
-    end
+  def self.ppid : Int64
+    self.ppid_system
```
ysbaddaden:
```suggestion
    ppid_system
```

src/process.cr
```diff
@@ -69,86 +60,7 @@ class Process
...
-      errno = Errno.value
-      LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
-      raise Errno.new("fork", errno)
-    else
-      # parent:
-      LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
-    end
-
-    pid
+    Process.times_system
```
ysbaddaden:
```suggestion
    times_system
```

src/process.cr
```diff
@@ -186,18 +98,28 @@ class Process
...
     rescue ex
-      process.kill
+      process.terminate
       raise ex
     end
   end
 
+  # Terminate process gracefully
+  def terminate
+    self.terminate_system
```
ysbaddaden:
```suggestion
    terminate_system
```

src/process.cr
```diff
@@ -186,18 +98,28 @@ class Process
...
+  end
+
+  # Terminate process immediately (kill)
+  def kill
+    self.kill_system
```
ysbaddaden:
```suggestion
    kill_system
```

src/process.cr
```diff
@@ -377,39 +253,29 @@ class Process
...
-        {% if flag?(:freebsd) %}
-          shell_args << ""
-        {% end %}
-
-        shell_args.concat(args)
-      end
-
-      command = "/bin/sh"
-      args = shell_args
+      ::Process.prepare_shell_system(command, args)
```
ysbaddaden:
```suggestion
      prepare_shell_system(command, args)
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,243 @@
...
+    return true if ret == 0
+    return false if Errno.value == Errno::ESRCH
+    raise Errno.new("kill")
+  end
+
+  protected def pgid_system : Int64
```
ysbaddaden: Same here, it should be a `self.pgid_system(pid)` class method, so we can avoid instantiating a Process object in `Process.pgid` methods.

jan-zajic:
@ysbaddaden it's done, please look at it.

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+      ensure
+        LibC._exit 127
+      end
+    end
+  end
+
+  protected def self.fork_system : Process
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      new pid
```
Sija: Prefer early returns:

```suggestion
      return new pid
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+      end
+    end
+  end
+
+  protected def self.fork_system : Process?
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      new pid
+    else
+      nil
```
Sija: `else` branch is redundant (that's already the implicit behavior)

src/process.cr
```diff
@@ -451,94 +317,104 @@ class Process
...
-      else
-        ENV.delete key
-      end
+  # See also: `Process.signal`
+  def signal(sig = Signal::TERM)
+    signal_system sig
+  end
+
+  # Sends a *signal* to the processes identified by the given *pids*.
+  def self.signal(signal : Signal, *pids : Int64)
```
```suggestion
  def self.signal(signal : Signal, *pids : Int64) : Nil
```

src/process.cr
```diff
@@ -451,94 +317,104 @@ class Process
...
+    pids.each do |pid|
+      Process.new(pid).signal(signal)
     end
+    nil
```
Sija: redundant

```suggestion
```

src/process.cr
```diff
@@ -451,94 +317,104 @@ class Process
...
+  end
 
-    Dir.cd(chdir) if chdir
+  # Returns `true` if the process identified by *pid* is valid for
+  # a currently registered process, `false` otherwise. Note that this
+  # returns `true` for a process in the zombie or similar state.
+  def self.exists?(pid : Int)
```
```suggestion
  def self.exists?(pid : Int) : Bool
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+
+      {% if flag?(:freebsd) %}
+        shell_args << ""
+      {% end %}
+
+      shell_args.concat(args)
+    end
+
+    command = "/bin/sh"
+    args = shell_args
```
Sija: redundant

```suggestion
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+    {command, args}
```
Sija:
```suggestion
    {command, shell_args}
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+
+  protected def self.pid_system : Int64
+    LibC.getpid.to_i64
+  end
+
+  protected def self.ppid_system : Int64
+    LibC.getppid.to_i64
+  end
+
+  protected def exists_system?
```
ysbaddaden: AFAIK we don't need the `#exists_system?` and `#pgid_system` instance methods, even on Windows. The class methods are sufficient and `Process.exists?(pid)` and `Process.pgid(pid)` can directly call both system class methods. It will avoid a few indirections.

src/crystal/system/unix/process.cr
ysbaddaden: Why a `return` here (not necessary) but not in `create_and_exec`?

src/crystal/system/unix/process.cr
ysbaddaden: It's just moving the existing, but yeah, it would clean up :)

ysbaddaden:
Oh, and *thank you* for spending so much time on making this super clean!

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,252 @@
...
+  end
+
+  protected def signal_system(signal : Signal)
+    ret = LibC.kill(@pid, signal.value)
+    raise Errno.new("kill") if ret < 0
+  end
+
+  protected def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?, reader_pipe : IO::FileDescriptor, writer_pipe : IO::FileDescriptor)
+    if pid = Process.fork_internal(will_exec: true)
+      pid
```
Sija: Prefer early returns:

```suggestion
      return new pid
```

ysbaddaden:
@RX14 Let's avoid holding things up and keep them hanging.

IMO: the abstraction here is *good*. It doesn't remove any functionality and will be enough to implement `Process` on Windows. We can make it perfect afterwards, deciding in their own dedicated issues, whether we keep fork, chroot or include (e)(uid|gid) or not.

src/crystal/system/unix/process.cr
jan-zajic: I prefer to keep them to be able to reuse process handle in Windows implementation when working with process instance (on Win, you need open handle to the process for most of the methods).

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,249 @@
...
+
+  protected def signal_system(signal : Signal)
+    ret = LibC.kill(@pid, signal.value)
+    raise Errno.new("kill") if ret < 0
+  end
+
+  protected def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?, reader_pipe : IO::FileDescriptor, writer_pipe : IO::FileDescriptor)
+    if pid = Process.fork_internal(will_exec: true)
+      return pid
+    else
```
Sija:
```suggestion
    end
```

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,249 @@
...
+        LibC._exit 127
+      end
+    end
+  end
+
+  protected def self.fork_system : Process
+    {% raise("Process fork is unsupported with multithread mode") if flag?(:preview_mt) %}
+    if pid = fork_internal(will_exec: false)
+      return new pid
+    else
```
Sija:
```suggestion
    end
```

src/crystal/system/unix/process.cr
Sija: As I mentioned before, `else` branch is redundant now.

src/process.cr
```diff
@@ -1,60 +1,51 @@
...
+
+  protected def initialize(@pid)
+    wait_system
+    @wait_count = 0
+  end
+
+  # Open process by identifier
+  #
+  # Checks that process with given *pid* exists and is accessible.
+  def self.open(pid : Int64)
```
ysbaddaden: There is no such method right now.

src/crystal/system/unix/process.cr
jan-zajic: Oh yeah, I can see the point now, thanks.

src/process.cr
jan-zajic: Ok i removed it so we keep changes minimal and can move on.

rdp:
What does `exists_system` mean?  The system part of the method name is a bit confusing to myself... :)

src/process.cr
```diff
@@ -258,26 +158,8 @@ class Process
 
     reader_pipe, writer_pipe = IO.pipe
```
RX14: I think the `reader_pipe`/`writer_pipe` stuff should be moved into `Crystal::System`, it's platform-specific. WIndows has no need for it: there's no fork.

ysbaddaden:
@RX14 Process.kill wasn't removed but renamed to Process.signal which is a better name. No need to open repetitive PR.

RX14:
@ysbaddaden I agree - but it still needs to be in seperate PRs.

src/process.cr
jan-zajic: Yes i can do it.

jan-zajic:
Ok  @RX14, @straight-shoota i move kill method renaming to new PR #8642. But because of this a our intention with @ysbaddaden to have portable API to stopping processes i rename proposed kill method that *immediately* terminate process to *interrupt*. Otherwise it would be super unclear until we do remove original kill methods. The more so that someone already suggested to make it only deprecated in #8642 (which I expected).

RX14:
@jan-zajic I'd rather just have no `kill` or `interrupt` or `terminate`, just `signal(:kill)` and `signal(:interrupt)`. On windows, `signal` would only be able to send sigkill, an attempt to send sigint would raise. This is similar to how go approaches it.

**Please** no user-visible API changes in this PR.

jan-zajic:
@RX14 i rearranged it as you wish. But i insist on to have some methods to stop Process (no matter how they are named, it can be discussed). We try to create portable API (at least @ysbaddaden and i), that can be implemented on every platform. If you can *create Process*, you must be *able to stop* it. Do not confuse it with that signal stuff. It's only implementation detail on unix. Stopping or killing most of the processes on Windows, for example, have nothing to do with signal API. When you mention go for example, it also has Kill() method on Process that causes the Process to exit immediately in platform specific way. But Go lang is not a good example of pretty apis at all, please let's not be influenced by previous compromises of other teams.

jan-zajic:
@bcardiff builds now starts to fail on:
> 1) HTTP::Cookie::Parser expired? not expired
     Failure/Error: parse_set_cookie("bla=1; expires=Thu, 01 Jan 2020 00:00:00 -0000").expired?.should eq false
       Expected: false
            got: true

P.S. happy New Year to everyone involved in Crystal Language!

RX14:
> Stopping or killing most of the processes on Windows, for example, have nothing to do with signal API

That is true, but that does not mean that the API for killing a process on windows cannot be exposed as `signal(:kill)`, which is then translated to `TerminateProcess`. This is how go handles it, and is a subjective API design issue. I'm fine with either way.

src/process.cr
```diff
@@ -229,7 +154,7 @@ class Process
     end
   end
 
-  getter pid : Int32
+  getter pid : Int64 = 0
```
RX14: Is this default zero neccesary?

Sija:
@jan-zajic You need to rebase on `master` in order to get green CI builds.

src/process.cr
```diff
@@ -1,60 +1,47 @@
...
+    ppid_system
   end
 
   # Sends a *signal* to the processes identified by the given *pids*.
-  def self.kill(signal : Signal, *pids : Int)
+  def self.kill(signal : Signal, *pids : Int64) : Nil
     pids.each do |pid|
-      ret = LibC.kill(pid, signal.value)
-      raise Errno.new("kill") if ret < 0
+      Process.new(pid).kill(signal)
```
RX14: This isn't good on unix since you need to allocate a `Process` and call `waitpid` just to send a signal. I think there needs to be both` .signal_system(signal, pid)` and `#signal_system(signal)`.

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,254 @@
...
+    raise Errno.new("getpgid") if ret < 0
+    ret.to_i64
+  end
+
+  protected def signal_system(signal : Signal)
+    ret = LibC.kill(@pid, signal.value)
+    raise Errno.new("kill") if ret < 0
+  end
+
+  protected def create_and_exec(command : String, args : (Array | Tuple)?, env : Env?, clear_env : Bool, fork_input : IO::FileDescriptor, fork_output : IO::FileDescriptor, fork_error : IO::FileDescriptor, chdir : String?)
```
RX14: Perhaps call this just `spawn`?

src/process.cr
```diff
@@ -15,15 +15,17 @@ class Process
     pid_system
   end
 
-  # Returns the process group identifier of the current process.
-  def self.pgid : Int64
-    pgid(0)
-  end
+  {% if flag?(:unix) || flag?(:docs) %}
```
RX14: These methods must exist on windows, and raise at runtime. i.e. delete this entire commit and raise inside `pgid_system` in `src/crystal/system/win32/process.cr`.

src/process.cr
```diff
@@ -273,6 +273,16 @@ class Process
     @waitpid.closed? || !exists_system?
   end
 
+  # Terminate process gracefully
```
RX14: Does this exist on windows?

ysbaddaden: Yes.

src/crystal/system/unix/process.cr
ysbaddaden: I prefer the explicit name than a confusing name with spawning a fiber.

src/process.cr
j8r: But if one use this methods on Windows, this error will be raised only at runtime, despite being catchable a compile-time.

At least, maybe having a warning on this ones for Windows.

src/process.cr
```diff
@@ -186,14 +111,14 @@ class Process
   #
   # Returns the block's value.
   def self.run(command : String, args = nil, env : Env = nil, clear_env : Bool = false, shell : Bool = false,
-               input : Stdio = Redirect::Pipe, output : Stdio = Redirect::Pipe, error : Stdio = Redirect::Pipe, chdir : String? = nil)
+               input : Stdio = Redirect::Pipe, output : Stdio = Redirect::Pipe, error : Stdio = Redirect::Pipe, chdir : String? = nil, &block : Process ->)
```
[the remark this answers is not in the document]
j8r: This big one line (and the above one) can be split to one argument per line.

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,262 @@
...
+    LibC.execvp(command, argv)
+
+    error_message = String.build do |io|
+      io << "execvp ("
+      command.inspect_unquoted(io)
+      args.try &.each do |arg|
+        io << ' '
+        arg.inspect(io)
+      end
+      io << ")"
```
j8r:
```suggestion
      io << ')'
```

src/process.cr
straight-shoota: `signal_system` is internal API, having a class method would be enough. It should be used for the (system specific) implementations of both class and instance methods `kill`, `terminate` and `interrupt`, respectively.
Adding an instance method `#signal_system(signal)` wouldn't be super useful when you can just call `Process.signal_system(signal, @pid)` instead. It would only be internal API anyway.

src/crystal/system/unix/process.cr
```diff
@@ -0,0 +1,262 @@
...
+
+  protected def self.times_system : Tms
+    LibC.getrusage(LibC::RUSAGE_SELF, out usage)
+    LibC.getrusage(LibC::RUSAGE_CHILDREN, out child)
+
+    Tms.new(
+      usage.ru_utime.tv_sec.to_f64 + usage.ru_utime.tv_usec.to_f64 / 1e6,
+      usage.ru_stime.tv_sec.to_f64 + usage.ru_stime.tv_usec.to_f64 / 1e6,
+      child.ru_utime.tv_sec.to_f64 + child.ru_utime.tv_usec.to_f64 / 1e6,
+      child.ru_stime.tv_sec.to_f64 + child.ru_stime.tv_usec.to_f64 / 1e6,
```
j8r: The `#to_f64` are not needed anymore since https://github.com/crystal-lang/crystal/pull/8120

src/process.cr
straight-shoota: @j8r This PR is mostly an internal refactor to extract the POSIX implementation. Not a place to discuss API changes.
I think we agreed to eventually remove fork and chroot, so they don't need to exist for win32. I'm not sure about the current discussion on pgid. @ysbaddaden and @jan-zajic voted to treat it like fork and chroot because it's a very POSIX-specific thing.

src/crystal/system/unix/process.cr
straight-shoota: Good catch. But I'd prefer to put such changes in a separate PR.

src/process.cr
RX14: @straight-shoota that's fine on unix, but on windows you call `TerminateProcess` on the handle.

Converting from the process handle to the PID just to convert to the process handle is inefficient, and that optimization is why there is both a class and an instance method: the `Process` already has the handle open.

src/process.cr
RX14: `fork` is going, and will not be in 1.0 in any form. That's why it's special.

Every other API has to be identical at compile time on every platform. This was agreed in an issue, and is the status quo across the whole rest of the stdlib: see `NotImplementedError`.

src/process.cr
RX14: How? I can't see a way to terminate an arbitrary process gracefully. Go doesn't support it.

src/process.cr
straight-shoota: Oh yes, you're right.

src/process.cr
straight-shoota: Yes, the question is: will process group id be in 1.0 or not? If I understood correctly, this is supposed to go as well.

RX14: @straight-shoota where's the discussion on that?

straight-shoota: To my knowledge there has not been a dedicated discussion on that specific feature. But it has been mentioned here and in #8515. There were no objections voiced, so it seems as if everyone is on board with this.

src/process.cr
jan-zajic: Most of time you deal with GUI process, because Windows is about, surprise, windows ;). So you must find main window of target process and send WM_CLOSE message to it. If window not found, you can send WM_QUIT message to main thread. For console process, you should send crtl+c event to it using GenerateConsoleCtrlEvent.

RX14: Go doesn't seem to want to deal with that complexity for graceful process termination. I guess we'll see how much code there is :)

PR #8537, opened 2019-11-29
Get rid off dependency on Signal in Win32 Process::Status

jan-zajic:
Remove dependency on Signal on win32 platform.

### DIFF
--- a/src/compiler/crystal/compiler.cr
+++ b/src/compiler/crystal/compiler.cr
@@ -577,7 +577,10 @@ module Crystal
       status = yield command, args
 
       unless status.success?
-        msg = status.normal_exit? ? "code: #{status.exit_code}" : "signal: #{status.exit_signal} (#{status.exit_signal.value})"
+        msg = "code: #{status.exit_code}"
+        {% unless flag?(:win32) %}
+          msg = "signal: #{status.exit_signal} (#{status.exit_signal.value})" unless status.normal_exit?
+        {% end %}
         code = status.normal_exit? ? status.exit_code : 1
         error "execution of command failed with #{msg}: `#{command}`", exit_code: code
       end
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/compiler.cr
```diff
@@ -577,7 +577,10 @@ module Crystal
       status = yield command, args
 
       unless status.success?
-        msg = status.normal_exit? ? "code: #{status.exit_code}" : "signal: #{status.exit_signal} (#{status.exit_signal.value})"
+        msg = "code: #{status.exit_code}"
+        {% unless flag?(:win32) %}
+          msg = status.normal_exit? ? "code: #{status.exit_code}" : "signal: #{status.exit_signal} (#{status.exit_signal.value})"
```
Sija: Please, keep it dry:

```suggestion
          msg = "signal: #{status.exit_signal} (#{status.exit_signal.value})" unless status.normal_exit?
```

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -175,7 +175,10 @@ module Crystal::Playground
       spawn do
         status = process.wait
         @logger.info "Code execution ended (session=#{@session_key}, tag=#{tag}, filename=#{output_filename})."
-        exit_status = status.normal_exit? ? status.exit_code : status.exit_signal.value
+        exit_status = status.exit_code
+        {% unless flag?(:win32) %}
+          exit_status = status.normal_exit? ? status.exit_code : status.exit_signal.value
```
Sija: ditto

```suggestion
          exit_status = status.exit_signal.value unless status.normal_exit?
```

straight-shoota:
I'd like to see a comment for each location that is removed which explains why this doesn't work for win32, whether that's a problem and if this needs attention later.

RX14:
Yeah, I'd much rather add something to the windows stubs file than ifdef out code. Ifdefs should be kept to a minimum.

straight-shoota:
The result of #8647 is also relevant for this.

PR #8549, opened 2019-12-03
Adds support for __dir__, but parses it as __DIR__ (and emits __DIR__ in the code formatter)

### LINKED ISSUE #8546: Would it be possible to add an alias for `__dir__`?
Hello, sorry if this has already been asked! I'm working on the [advent of code puzzles](https://adventofcode.com/), and I figured out how to read a file in the same directory as the script:

```
File.read(File.expand_path("../input", __FILE__)) #...
```

I noticed that `__dir__` isn't defined. But I actually just realized that Crystal supports `__DIR__`, and that does work:

```
File.read(File.join(__DIR__, "input")) #...
```

I was just wondering if you might consider adding an alias for `__dir__`? `__DIR__` does make more sense and it has the same case as `__FILE__`, but it's just one less surprise for people coming from Ruby.

ndbroadbent:
Fixes #8546.

Would this be a good workaround? I've been doing the advent of code puzzles, and I thought it was cool that my programs were almost exactly the same for both Ruby and Crystal:

```
puts File.read(File.join(__DIR__, "input")).lines.sum { |n| (n.to_i / 3).floor.to_i - 2 }
```

```
def fuel_for(mass)
  fuel = (mass.to_i / 3).floor.to_i - 2
  return 0 if fuel <= 0
  fuel + fuel_for(fuel)
end

puts File.read(
  File.join(__DIR__, "input")
).lines.sum { |mass| fuel_for(mass) }
```

So `__DIR__` is the only difference (and it took me a little while to figure out.) I think it would just be nice to support `__dir__` in this way. I know compatibility isn't a goal, but this might reduce a little bit of friction for Rubyists.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -400,8 +400,10 @@ describe Crystal::Formatter do
   assert_format %("foo \#{ 1  +  2 }"), %("foo \#{1 + 2}")
   assert_format %("foo \#{ 1 } \#{ __DIR__ }"), %("foo \#{1} \#{__DIR__}")
   assert_format %("foo \#{ __DIR__ }"), %("foo \#{__DIR__}")
+  assert_format %("foo \#{ __dir__ }"), %("foo \#{__DIR__}")
   assert_format "__FILE__", "__FILE__"
   assert_format "__DIR__", "__DIR__"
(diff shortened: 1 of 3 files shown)


asterite:
Thank you! It probably looks good but there was no agreement about doing this.

PR #3924, opened 2017-01-21
Add String#center

ilja:
This adds String#center: `"hello".center(9, '-')` => `--hello--`

PR #8557, opened 2019-12-05
Add String#center method

hutou:
`String#center(width, char) `returns a new string of length `width` with original string centered, and padded with `char` (default = space)

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -3616,7 +3616,7 @@ class String
   # "Aubergine".ljust(8)   # => "Aubergine"
   # ```
   def ljust(len, char : Char = ' ')
-    just len, char, true
+    just len, char, -1
   end
 
   # Adds instances of *char* to left of the string until it is at least size of *len*.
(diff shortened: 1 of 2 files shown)


src/string.cr
```diff
@@ -3627,37 +3627,61 @@ class String
...
+          leftpadding.times do
+            Intrinsics.memcpy(buffer.as(Void*), bytes.to_unsafe.as(Void*), count, false)
+            buffer += count
+          end
         end
       end
-
-      unless left
-        buffer.copy_from(to_unsafe, bytesize)
+      Intrinsics.memcpy(buffer.as(Void*), to_unsafe.as(Void*), bytesize, false)
```
wontruefree: I think you might be able to use `Pointer.copy_from`

```suggestion
      buffer.copy_from(to_unsafe, bytesize)
```

wontruefree: I see now this is diff noise so this might be outside the scope of this PR.

straight-shoota:
This continues the previous PR #3924 (stalled).

src/string.cr
hutou: Is this "better" than memcpy ?

src/string.cr
```diff
@@ -3627,37 +3627,61 @@ class String
...
-    new_bytesize = bytesize + difference * count
+    padding = (len - size)
+    new_bytesize = bytesize + padding * count
+    case justify
+    when .< 0
+      leftpadding, rightpadding = 0, padding
+    when .> 0
+      leftpadding, rightpadding = padding, 0
+    else
+      leftpadding = padding >> 1
```
RX14: `// 2`, please

there's no need for manual pinhole optimizations the compiler is perfectly capable of.

src/string.cr
RX14: It's the same as `memcpy`, but shorter.

It's fine to leave this as-is for this PR.

src/string.cr
hutou: Ok, done !

RX14:
Thank you!

PR #8559, opened 2019-12-06
fix test for linux, don't hang after Error writing to socket: Broken …

rdp:
…pipe (Errno) ...SSL_do_handshake

Test worked great in OS X, but in Linux somehow even after the client establishes the connection it's not yet "fully" established or something.  Either that or packets are just plain dropped when #close is called on a localhost sockets?  Not quite sure what was going on there.

Regardless, in linux boxes running the spec suite would freeze/hang at this point.  Now it continues and passes.

### DIFF
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -41,9 +41,11 @@ describe OpenSSL::SSL::Socket do
     server_context, client_context = ssl_context_pair
 
     client_successfully_closed_socket = Channel(Nil).new
+    server_received_socket = Channel(Nil).new
     spawn do
       OpenSSL::SSL::Server.open(tcp_server, server_context, sync_close: true) do |server|
         server_client = server.accept
+        server_received_socket.send(nil)
         # require client to close the socket from its side, without the server closing it, IIS behave this way.
         client_successfully_closed_socket.receive
         server_client.close
@@ -51,6 +53,7 @@ describe OpenSSL::SSL::Socket do
     end
     socket = TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port)
     socket = OpenSSL::SSL::Socket::Client.new(socket, client_context, hostname: "example.com", sync_close: true)
+    server_received_socket.receive # make sure connection established
     socket.close
     client_successfully_closed_socket.send(nil)
   end


rdp:
OK seems the problem is a bit deeper than I expected, still investigating...

rdp:
Here's a "righter" fix, working on updating here:

```
diff --git a/spec/std/openssl/ssl/socket_spec.cr b/spec/std/openssl/ssl/socket_spec.cr
index 1b6675327..9ba587f1f 100644
--- a/spec/std/openssl/ssl/socket_spec.cr
+++ b/spec/std/openssl/ssl/socket_spec.cr
@@ -20,6 +20,27 @@ describe OpenSSL::SSL::Socket do
     end
   end
 
+  it "accepts clients that don't read anything and close the connection" do
+    # in tls 1.3, if clients don't read anything and close the connection
+    # the server might still try and send it a ticket, resulting in a pipe failure
+    server_context.disable_session_resume_tickets
+
+    OpenSSL::SSL::Server.open(tcp_server, server_context) do |server|
+      spawn do
+        # the :sync_close portion, as implemented, effects a socket close from the client
+        OpenSSL::SSL::Socket::Client.open(TCPSocket.new(tcp_server.local_address.address, tcp_server.local_address.port), client_context, hostname: "example.com", sync_close: true) do |socket|
+          # doesn't read anything, just write and close connection immediately
+          socket.puts "hello"
+        end
+      end
+
+      client = server.accept # shouldn't raise
+      client.close
+    end
+  end
+
   it "returns the TLS version" do
     tcp_server = TCPServer.new(0)
     server_context, client_context = ssl_context_pair
@@ -39,6 +60,7 @@ describe OpenSSL::SSL::Socket do
   it "closes connection to server that doesn't properly terminate SSL session" do
     tcp_server = TCPServer.new(0)
     server_context, client_context = ssl_context_pair
+    server_context.disable_session_resume_tickets
 
     client_successfully_closed_socket = Channel(Nil).new
     spawn do
diff --git a/src/openssl/lib_ssl.cr b/src/openssl/lib_ssl.cr
index f663c3ea4..455e83c41 100644
--- a/src/openssl/lib_ssl.cr
+++ b/src/openssl/lib_ssl.cr
@@ -206,6 +206,12 @@ lib LibSSL
   # Hostname validation for OpenSSL <= 1.0.1
   fun ssl_ctx_set_cert_verify_callback = SSL_CTX_set_cert_verify_callback(ctx : SSLContext, callback : CertVerifyCallback, arg : Void*)
 
+  # control TLS 1.3 session ticket generation
+  {% if compare_versions(OPENSSL_VERSION, "1.1.1") >= 0 %}
+    fun ssl_ctx_set_num_tickets = SSL_CTX_set_num_tickets(ctx : SSLContext, larg : LibC::SizeT) : Int
+    fun ssl_set_num_tickets = SSL_set_num_tickets(ctx : SSL, larg : LibC::SizeT) : Int
+  {% end %}
+
   {% if compare_versions(OPENSSL_VERSION, "1.1.0") >= 0 %}
     fun tls_method = TLS_method : SSLMethod
   {% else %}
diff --git a/src/openssl/ssl/context.cr b/src/openssl/ssl/context.cr
index cad1934b6..e0eea9e6f 100644
--- a/src/openssl/ssl/context.cr

```

rdp:
superceded by #8582

PR #8560, opened 2019-12-06
Make mutex reentrant by default

bcardiff:
Revert the breaking-change of #8295

### DIFF
--- a/src/mutex.cr
+++ b/src/mutex.cr
@@ -1,6 +1,14 @@
 require "crystal/spin_lock"
 
 # A fiber-safe mutex.
+#
+# By default the mutex is reentrant and can only be unlocked by the
+# same fiber that locked it. This behaviour is enforced by some additional
+# runtime checks.
+#
+# A non-reentrant mutex can be aquired only when unlocked and can lead
+# to deadlocks if used recursively. A non-reentrant mutex is neededin more
+# complex programs where the lock and unlock can happen in different fibers.
 class Mutex
   @state = Atomic(Int32).new(0)
   @mutex_fiber : Fiber?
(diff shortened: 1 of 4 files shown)


src/mutex.cr
```diff
@@ -1,6 +1,14 @@
 require "crystal/spin_lock"
 
 # A fiber-safe mutex.
+#
+# By default the mutex is reentrant and can only be unlocked by the
+# same fiber that locked it. This behaviour is enforced by some additional
+# runtime checks.
+#
+# A non-reentrant mutex can be aquired only when unlocked and can lead
+# to deadlocks if used recursively. A non-reentrant mutex is neededin more
```
bew:
```suggestion
# to deadlocks if used recursively. A non-reentrant mutex is needed in more
```

asterite:
I have some comments regarding the code that's not part of this diff.

In lock there's:

```
if @reentrant && @mutex_fiber == Fiber.current
  @lock_count += 1
  return
end
```

What happens if `@mutex_fiber` is a fiber but it's not `Fiber.current`? Should that raise?

Likewise, in `unlock` we have:

```crystal
unless @mutex_fiber == Fiber.current
  raise "Attempt to unlock a mutex which is not locked"
end
```

That means that if `@mutex_fiber` is not `nil` but it's a Fiber that's different from `Fiber.current`, you will get that error message which is pretty misleading. In the case I mention the error message should probably say "Attempt to unlock a mutex from a Fiber different than the one that locked it".

What do you think?

bcardiff:
@asterite the `if @reentrant && @mutex_fiber == Fiber.current` is fine. If the condition is not met the `lock_slow` will be called to reschedule the current fiber to compete later for the mutex.

Regarding the second thing, if the unlock is done out of order then yes, the exception message is misleading. Maybe that can be handled if/when the approach in https://github.com/crystal-lang/crystal/pull/8295#issuecomment-558063219 is implemented. Or the line could be 

```crystal
raise "Attempt to unlock a mutex which is not locked in the current fiber"
``` 
(which is always true)

src/mutex.cr
bcardiff: thanks

bcardiff:
Closing in favor of #8563

PR #8562, opened 2019-12-06
Compiler: handle virtual types for visibility modifiers

### LINKED ISSUE #8561: Inheritance breaks protected visibility
```crystal
module Namespace
  class Foo 
    protected def foo 
      puts "Foo"
    end 
  end 

  class Bar 
    def bar 
      Foo.new.foo
    end 
  end 

  class Baz 
    def initialize
      @bar = Bar.new
    end 

    def bar 
      @bar.bar
    end 
  end 
end

Namespace::Baz.new.bar
```
This runs as expected, printing "Foo".  However, I need to make `Baz` inherit from `Bar` (for a decorator pattern setup):
```crystal
module Namespace
  class Foo 
    protected def foo 
      puts "Foo"
    end 
  end 

  class Bar 
    def bar 
      Foo.new.foo
    end 
  end 

  class Baz < Bar
    def initialize
      @bar = Bar.new
    end 

    def bar 
      @bar.bar
    end 
  end 
end

Namespace::Baz.new.bar
```
This gives the error
```
 10 | Foo.new.foo
              ^--
Error: protected method 'foo' called for Namespace::Foo
```
Having this inheritance doesn't seem like it should affect the visibility this way.

OS: Ubuntu 18.04
`crystal -v`:
```

LLVM: 8.0.0
```
This also happens on a crystal compiler built from master.
Compiled-from-master `crystal -v`:
```
Crystal 0.32.0-dev [7fd3863c4] (2019-12-06)

LLVM: 8.0.0
```

asterite:
Fixes #8561

### DIFF
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -311,16 +311,17 @@ module Crystal
     # the types are in the same namespace. Otherwise, it means they are just
     # in the same type hierarchy.
     def has_protected_acces_to?(type, allow_same_namespace = true)
-      owner = self
+      owner = self.devirtualize
+      type = type.devirtualize
 
       # Allow two different generic instantiations
       # of the same type to have protected access
       type = type.generic_type.as(Type) if type.is_a?(GenericInstanceType)
       owner = owner.generic_type.as(Type) if owner.is_a?(GenericInstanceType)
 
-      self.implements?(type) ||
-        type.implements?(self) ||
-        (allow_same_namespace && same_namespace?(type))
+      owner.implements?(type) ||
+        type.implements?(owner) ||
+        (allow_same_namespace && owner.same_namespace?(type))
     end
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/types.cr
```diff
@@ -311,16 +311,17 @@ module Crystal
     # the types are in the same namespace. Otherwise, it means they are just
     # in the same type hierarchy.
     def has_protected_acces_to?(type, allow_same_namespace = true)
```
bew:
```suggestion
    def has_protected_access_to?(type, allow_same_namespace = true)
```
found a  typo :eyes:

asterite: Oh, yes! I saw it too but while fixing the bug I completely forgot about it. I'll add another commit. Thanks!

PR #8563, opened 2019-12-06
Add error check to Mutex by default

ysbaddaden:
Alternative to #8560 that implements deadlock protection and prevents bogus unlocks by default (raises on deadlock), while still allowing reentrant (optional) and unsafe (no error check).

See https://github.com/crystal-lang/crystal/pull/8295#issuecomment-558063219

### DIFF
--- a/src/mutex.cr
+++ b/src/mutex.cr
@@ -1,6 +1,20 @@
 require "crystal/spin_lock"
 
 # A fiber-safe mutex.
+#
+# Provides deadlock protection by default. Attempting to re-lock the mutex from
+# the same fiber will raise an exception. Trying to unlock an unlocked mutex, or
+# a mutex locked by another fiber will raise an exception.
+#
+# The reentrant protection maintains a lock count. Attempting to re-lock the
+# mutex from the same fiber will increment the lock count. Attempting to unlock
+# the counter from the same fiber will decrement the lock count, eventually
+# releasing the lock when the lock count returns to 0. Attempting to unlock an
+# unlocked mutex, or a mutex locked by another fiber will raise an exception.
+#
+# You also disable all protections with `unchecked`. Attempting to re-lock the
+# mutex from the same fiber will deadlock. Any fiber can unlock the mutex, even
+# if it wasn't previously locked.
 class Mutex
   @state = Atomic(Int32).new(0)
   @mutex_fiber : Fiber?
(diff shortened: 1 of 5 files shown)


bcardiff:
I would like to avoid the usage of `type` and `unsafe` terms.

Instead of `type` can we use `protection` or `policy` with maybe deadlock_detection, reentrant, non_reentrant/unprotected or similar as values?

RX14:
Why?

To me, the name must have unsafe in it to let people know that this is unsafe, and they have to be extra careful to avoid deadlocks.

bcardiff:
@RX14 the proposed "unsafe" behavior is totally valid and has its use cases. I think that calling it `unsafe` is unprecise on what are the guarantees offered.

---

Another alternative to `type` is `kind` it's still generic but I try to avoid type named properties. I found it ambiguous when describing it.

ysbaddaden:
Well, manipulating pointers is valid, unsafe but valid.

Using unchecked mutexes properly is fine, but it has dangerous/tricky semantics that developers must take care to avoid, without runtime protections. I mean: any fiber being capable to unlock the mutex *is* unsafe behavior to me.

Type is because pthreads' spec uses that term. It can be Kind. I'm not sure about Protection.

Default is because that's the default behavior and you should never have to specify it.

Reentrant was already used; I think I would have used Recursive otherwise.

oprypin:
I don't like `default` because there's no information behind it, specifically if someone were to read just code.
`checked` would be nice for the default.

As for the parameter name, `policy` seems fine, if `type` is somehow undesirable.

asterite:
Does Golang have these two kind of mutexes? If not, could we maybe have a mutex internal to the runtime that's unsafe but fast, but always expose a safe mutex to users? From a user's perspective, it's not clear when to use reentrant or not. If we do know at runtime, let's keep that logic to ourselves.

bcardiff:
`kind / policy / behavior` and options `checked, reentrant, unchecked`. (@oprypin 💡)

`unsafe` as a term is used a lot in crystal for low level stuff and I don't see this as one of them, and it could be read as nop/skip.

The main thing I want is to avoid the usage of `type` and `unsafe` terms.

ysbaddaden:
@asterite Go mutexes don't associate the lock to any goroutine. I.e. they're unchecked, period. No reentrancy, no checks.

We only use reentrant for crystal once (for lazy initializers?) to avoid a potential deadlock, if I understand correctly. Thought I think a recursive initializer would probably recurse infinitly (stack overflow, so I'm not sure rentrant is that useful.

bcardiff:
@ysbaddaden I think reentrant can be used for example in #8140 without ending in an infinite recursion. Lazy initializers are also a use case and the compiler checks there is no cyclic dependency IIRC.

ysbaddaden:
Updated. I finally went for `Protection` (which is the actual term used in documentation, so it ain't that weird, and won't be used much publicly anyway) with `:checked` (default), `:reentrant` and `:unchecked`.

src/mutex.cr
```diff
@@ -1,6 +1,20 @@
 require "crystal/spin_lock"
 
 # A fiber-safe mutex.
+#
+# Provides deadlock protection by default. Attempting to re-lock the mutex from
+# the same fiber will raise an exception. Trying to unlock an unlocked mutex, or
+# a mutex locked by another fiber will raise an exception.
+#
+# The reentrant protection maintains a lock count. Attempting to re-lock the
+# mutex# from the same fiber will increment the lock count. Attempting to unlock
```
bcardiff:
```suggestion
# mutex from the same fiber will increment the lock count. Attempting to unlock
```

RX14:
Wonderful, thank you @ysbaddaden!

ysbaddaden:
Fixed typo.

Issue #8539, opened 2019-11-30
Can't use an endless range to slice a string with String#[]?


sudo-nice:
This will work:

```crystal
string = "foobar"
string[3..] #=> bar
```

But this won't:
```crystal
string = "foobar"
string[3..]? #=> Error: no overload matches 'String#[]?' with type Range(Int32, Nil)
```
---
Crystal 0.31.1 [0e2e1d067] (2019-09-30)
LLVM: 8.0.0
Default target: x86_64-unknown-linux-gnu

KarthikMAM:
@straight-shoota Mind if I take a stab at this? Would love to start contributing to crystal.

PR #8567, opened 2019-12-07
add endless range fix for strings and arrays

KarthikMAM:
### Description
This basically fixes crystal-lang/crystal#8539. This behavior is observed in `array.cr` as well. So, this PR takes care of that as well.

### DIFF
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -281,6 +281,15 @@ describe "Array" do
       [1, 2, 3][1, 3]?.should eq([2, 3])
       [1, 2, 3][4, 0]?.should be_nil
     end
+
+    it "gets with range without end" do
+      [1, 2, 3][1..nil]?.should eq([2, 3])
+      [1, 2, 3][4..nil]?.should be_nil
+    end
+
+    it "gets with range without beginning" do
+      [1, 2, 3][nil..1]?.should eq([1, 2])
+    end
   end
 
   describe "[]=" do
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -188,6 +188,15 @@ describe "String" do
       "hello"[1, 3]?.should eq("ell")
       "hello"[6, 3]?.should be_nil
     end
+
+    it "gets with range without end" do
+      "hello"[1..nil]?.should eq("ello")
+      "hello"[6..nil]?.should be_nil
+    end
+
+    it "gets with range without beginning" do
+      "hello"[nil..2]?.should eq("hel")
+    end
   end
 
   describe "byte_slice" do
--- a/src/array.cr
+++ b/src/array.cr
@@ -537,13 +537,14 @@ class Array(T)
     self[*Indexable.range_to_index_and_count(range, size)]
   end
 
-  # Like `#[Range(Int, Int)]`, but returns `nil` if the range's start is out of range.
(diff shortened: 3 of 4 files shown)


spec/std/string_spec.cr
```diff
@@ -188,6 +188,16 @@ describe "String" do
...
     end
+
+    it "gets with range without end" do
+      "há日本語"[1..nil]?.should eq("á日本語")
+      "hello"[6..nil]?.should be_nil
+    end
+
+    it "gets with range without beginning" do
+      "há日本語"[nil..2]?.should eq("há日")
+      "hello"[nil..6]?.should be_nil
```
KarthikMAM: @asterite Shouldn't this line return a nil and throw exception without `?` option? Another bug? Thoughts?

```cr
"hello"[nil..7] # this doesn't raise error
"hello"[6..7] # this raises an error
```

https://play.crystal-lang.org/#/r/85o1

spec/std/array_spec.cr
```diff
@@ -279,7 +279,15 @@ describe "Array" do
 
     it "gets with start and count" do
       [1, 2, 3][1, 3]?.should eq([2, 3])
-      [1, 2, 3][4, 0]?.should be_nil
+    end
```
straight-shoota: You shouldn't remove this expectation.

spec/std/array_spec.cr
```diff
@@ -279,7 +279,15 @@ describe "Array" do
...
+
+    it "gets with range without beginning" do
+      [1, 2, 3][1..nil]?.should eq([2, 3])
+      [1, 2, 3][4..nil]?.should be_nil
+    end
+
+    it "gets with range without end" do
+      [1, 2, 3][nil..1]?.should eq([1, 2])
```
straight-shoota: Could you also add a spec for `[nil...0]`?

spec/std/array_spec.cr
```diff
@@ -279,7 +279,15 @@ describe "Array" do
 
     it "gets with start and count" do
       [1, 2, 3][1, 3]?.should eq([2, 3])
-      [1, 2, 3][4, 0]?.should be_nil
+    end
+
+    it "gets with range without beginning" do
```
straight-shoota: The spec descriptions need to be swapped.

straight-shoota: Dito for string.

spec/std/array_spec.cr
straight-shoota: Ditto for string.

KarthikMAM: If I am not wrong `[nil..0]` seems to be a valid case just like `[nil..1]` right? Anything I am missing?

spec/std/array_spec.cr
KarthikMAM: Done.

KarthikMAM: Think I removed that accidentally. Sorry about that.

spec/std/array_spec.cr
KarthikMAM: Done

spec/std/array_spec.cr
KarthikMAM: Is there any significance to this additional spec... Seems to be returning just any empty array [].

straight-shoota: Yes, it's an edge case and it should be specified whether it returns empty array or `nil`.

PR #8569, opened 2019-12-08
Fix Hash's initial_capacity documentation

r00ster91:
#8017 changed the default `initial_capacity` to 8.

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -220,8 +220,8 @@ class Hash(K, V)
   # The *initial_capacity* is useful to avoid unnecessary reallocations
   # of the internal buffer in case of growth. If the number of elements
   # a hash will hold is known, the hash should be initialized with that
-  # capacity for improved performance. Otherwise, the default is 11 and inputs
-  # less than 11 are ignored.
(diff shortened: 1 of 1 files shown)


bew:
Just an idea, since this is a rather arbitrary value afaik, we could have it in a constant `INITIAL_CAPACITY`, and in the documentation refer to it using a macro `{{ INITIAL_CAPACITY }}` to avoid these kind of desynchronization between the code and the docs?

r00ster91:
That doesn't sound too bad but I'm pretty sure that would require wrapping the entire constructors with `{% begin %}` and `{% end %}` and then that could become pretty ugly when there's also additional indentation in the `{% begin %}` and `{% end %}`. I'm not sure it's worth the complexity and I doubt the initial capacity is changing that often.

straight-shoota:
The default capacity actually isn't really important for users. It's an implementation detail. It's worth mentioning it, but not exposing it as a constant.

bew:
We could use a private constant to not expose this, but you're right, it's probably not worth the burden

straight-shoota:
Thank you @r00ster91 ❤️

PR #8580, opened 2019-12-12
Update CI to use 0.32.0

bcardiff:
* Update version to 0.33.0-dev
* Use llvm@9 on osx builds


src/VERSION
```diff
@@ -1 +1 @@
-0.32.0
+0.33.0-dev
```
straight-shoota:
```suggestion
0.32.1-dev
```

bcardiff: Since this is based on master, it should be 0.33.0-dev.
The release/0.32 branch will have `0.32.1-dev` until it is released. And will be using 0.31.1 for building the compiler, instead of 0.32.0.

PR #8582, opened 2019-12-13
fix sporadic failure in specs with OpenSSL 1.1+.

rdp:
https://github.com/openssl/openssl/issues/6904

### DIFF
--- a/src/openssl/ssl/context.cr
+++ b/src/openssl/ssl/context.cr
@@ -191,6 +191,20 @@ abstract class OpenSSL::SSL::Context
     def self.from_hash(params) : self
       super(params)
     end
+
+    # Disables all session ticket generation for this context.
+    # Tickets are used to resume earlier sessions more quickly,
+    # but in TLS 1.3 if the client connects, sends data, and closes the connection
+    # unidirectionally, the server connects, then sends a ticket
+    # after the connect handshake, the ticket send can fail with Broken Pipe.
+    # So if you have that kind of behavior (clients that never read) call this method.
+    def disable_session_resume_tickets
+      add_options(OpenSSL::SSL::Options::NO_TICKET) # TLS v1.2 and below
+      {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.1.1") >= 0 %}
+        out = LibSSL.ssl_ctx_set_num_tickets(self, 0) # TLS v1.3
+        raise OpenSSL::Error.new("SSL_CTX_set_num_tickets") if out != 1
+      {% end %}
+    end
   end
 
   protected def initialize(method : LibSSL::SSLMethod)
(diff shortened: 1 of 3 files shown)


ysbaddaden:
Reading the linked OpenSSL issue, there is another solution that would work all the time, and transparently: insert a dummy `SSL_read` after the first `SSL_shutdown`. That would force a read and handle any pending session tickets, before a second `SSL_shutdown`.

src/openssl/ssl/context.cr
```diff
@@ -191,6 +191,20 @@ abstract class OpenSSL::SSL::Context
...
+    # Disables all session ticket generation for this context.
+    # Tickets are used to resume earlier sessions more quickly,
+    # but in TLS 1.3 if the client connects, sends data, and closes the connection
+    # unidirectionally, the server connects, then sends a ticket
+    # after the connect handshake, the ticket send can fail with Broken Pipe.
+    # So if you have that kind of behavior (clients that never read) call this method.
+    def disable_session_resume_tickets
+      add_options(OpenSSL::SSL::Options::NO_TICKET) # TLS v1.2 and below
+      {% if compare_versions(LibSSL::OPENSSL_VERSION, "1.1.1") >= 0 %}
+        out = LibSSL.ssl_ctx_set_num_tickets(self, 0) # TLS v1.3
```
Sija: `out` is a keyword, please use some other name (`ret`?)

bcardiff:
@ysbaddaden I can't find exactly where in the linked issue it stands about the dummy `SSL_read`.
Is that a read with zero bytes?

Sija:
@bcardiff See https://github.com/openssl/openssl/issues/6904#issuecomment-411829769

ysbaddaden:
Actually (3) at https://github.com/openssl/openssl/issues/6904#issuecomment-412033275 

Or we try to avoid *raising* inside the BIO and try to properly report the Errno to OpenSSL, if that's possible?

bcardiff:
I would definitely prefer a transparent solution.
Since it requires a bit more of investigation, and due to the update in the CI to use bionic (and that uses openssl 1.1.1) I would say to

1. Apply the suggestions by @Sija
2. Merge this PR to be able to have a green master
3. Iterate for 0.33 how to allow a better solution

That is unless, someone can make things work in the next couple of days.

I checked this PR on top of the CI using 0.32.0 and it works https://circleci.com/workflow-run/96ac4542-1977-40ca-b631-fa1c19f02033 .

Sija:
https://github.com/openssl/openssl/issues/7948#issuecomment-562931724 looks like a possible (albeit pretty hackish) solution.

rdp:
calling `SSL_read` "fixes it and works" if the client does it.  However what I was hoping for in this PR is a fix for the server so that it "can work" with any generic client, including ones that don't call that (that just unidirectionally close, which some do)...if that makes sense... :)

src/openssl/ssl/context.cr
rdp: OK (maybe crystal could warn in this case? But that's a topic for another day...)

ysbaddaden:
Ah yeah, I didnt properly read.

I guess that OpenSSL is behaving wrongly and shall be fixed. In the meantime, we can merge this, and/or disable TLS 1.3 for the failing spec, or add a dummy read in the spec to avoid the write then close behavior that triggers the issue.

bcardiff:
Thanks @rdp for iterating here. When I checked if using bionic would break the new specs were not in master yet.

PR #8587, opened 2019-12-16
Compiler: fix bug related to closured var not considered as such

### LINKED ISSUE #5609: Incorrect behaviour when capturing var and changing it in a loop
Crystal 0.24.1

```crystal
def capture(&block)
  block
end

x = 1
3.times do
  p x
  capture do
    x = "hello"
  end.call
end
```

Output:

```
1
194613232
194613232
```

The problem is that we only bind the variable to the metavariable if it's a closured var on a first pass, but the first read of `x` happens before we determine it's a closure.

I found this while thinking about how to fix https://github.com/crystal-lang/crystal/issues/4359

asterite:
Hopefully fixes #5609

This one got me again in Advent of Code so I decided it's to fix it for me and for everyone.

I say "hopefully" because I think the fix is correct and I didn't have time to run the entire spec suit locally, but CI will have the final word.

### DIFF
--- a/spec/compiler/semantic/closure_spec.cr
+++ b/spec/compiler/semantic/closure_spec.cr
@@ -524,4 +524,26 @@ describe "Semantic: closure" do
       ),
       "can't send closure to C function (closured vars: x)"
   end
+
+  it "correctly detects previous var as closured (#5609)" do
+    assert_error %(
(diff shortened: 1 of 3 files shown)


asterite:
Nah, it's bad.

PR #8593, opened 2019-12-17
ARGF: Refactor to remove `count` variable

arcage:
At the line 14,`count` is initialized as a size of given `slice` for reading.

It will be never modified while the reading process after that.
And, in current version, `count` has been used only to truncate the `slice`.

Because `slize[0, count]` is as same as just `slice` when `count` is the size of `slice`, I think this code can work without `count`.


arcage:
I will close this once and reopen to test again.

PR #8596, opened 2019-12-18
Remove PartialComparable

Blacksmoke16:
Module was deprecated 4 minor versions ago as part of #7664, probably is safe to remove now.


RX14:
:+1: for the cleanup branch

PR #6004, opened 2018-04-25
Remove dependency on libxml from spec runner

straight-shoota:
This PR reverts the implementation of #4090 but modifies the plain string formatter to produce the same, indented output as `XML::Builder`. This avoids having all specs depend on `libxml` while still enabling JUnit formatter.

Closes #5662

PR #8599, opened 2019-12-19
Spec::JUnitFormatter enhancements

Sija:
- Outputs elapsed time for testsuite and single testcases
- Outputs timestamp for testsuite
- Outputs hostname for testsuite
- Outputs exception class name as an `<error type="...">` attribute
- Escapes control characters within `<testcase name="...">` attribute
- Marks `pending` testcases as `skipped`
- Returns count for `skipped` testcases (from `pending`)
- `<testcase classname="...">` is now built from relative path (to `Dir.current`)
- Changes `—junit_output` flag to take full path to the junit xml file, instead of just a directory

### DIFF
--- a/spec/std/spec/junit_formatter_spec.cr
+++ b/spec/std/spec/junit_formatter_spec.cr
@@ -16,7 +16,7 @@ describe "JUnit Formatter" do
                  </testsuite>
                  XML
 
-    output.chomp.should eq(expected)
+    output.should eq(expected)
   end
 
   it "reports failures" do
@@ -33,7 +33,7 @@ describe "JUnit Formatter" do
                  </testsuite>
                  XML
 
-    output.chomp.should eq(expected)
+    output.should eq(expected)
   end
 
   it "reports errors" do
@@ -50,7 +50,7 @@ describe "JUnit Formatter" do
                  </testsuite>
                  XML
 
-    output.chomp.should eq(expected)
+    output.should eq(expected)
   end
 
   it "reports mixed results" do
@@ -77,7 +77,7 @@ describe "JUnit Formatter" do
                  </testsuite>
                  XML
 
-    output.chomp.should eq(expected)
+    output.should eq(expected)
   end
 
   it "escapes spec names" do
@@ -125,7 +125,7 @@ private def build_report
   formatter = Spec::JUnitFormatter.new(output)
   yield formatter
   formatter.finish(Time::Span.zero, false)
-  output.to_s
+  output.to_s.chomp
 end
 
 private def exception_with_backtrace(msg)
--- a/spec/std/spec/junit_formatter_spec.cr
+++ b/spec/std/spec/junit_formatter_spec.cr
@@ -81,16 +81,19 @@ describe "JUnit Formatter" do
   it "reports mixed results" do
     now = Time.utc
 
+    this_filename = __FILE__
+
     output = build_report_with_mocked_timestamp(now) do |f|
       f.report Spec::Result.new(:success, "should do something1", "spec/some_spec.cr", 33, 2.seconds, nil)
       f.report Spec::Result.new(:fail, "should do something2", "spec/some_spec.cr", 50, 0.5.seconds, nil)
       f.report Spec::Result.new(:error, "should do something3", "spec/some_spec.cr", 65, nil, nil)
       f.report Spec::Result.new(:error, "should do something4", "spec/some_spec.cr", 80, nil, nil)
+      f.report Spec::Result.new(:pending, "should do something5", this_filename, 90, nil, nil)
     end
 
     expected = <<-XML
                  <?xml version="1.0"?>
-                 <testsuite tests="4" disabled="0" errors="2" failures="1" time="0.0" timestamp="#{now.to_rfc3339}" hostname="#{System.hostname}">
+                 <testsuite tests="5" disabled="1" errors="2" failures="1" time="0.0" timestamp="#{now.to_rfc3339}" hostname="#{System.hostname}">
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something1" time="2.0"/>
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something2" time="0.5">
                      <failure/>
@@ -100,6 +103,9 @@ describe "JUnit Formatter" do
                    </testcase>
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something4">
                      <error/>
+                   </testcase>
+                   <testcase file="#{this_filename}" classname="spec.std.spec.junit_formatter_spec" name="should do something5">
+                     <skipped/>
                    </testcase>
                  </testsuite>
                  XML
--- a/spec/std/spec/junit_formatter_spec.cr
+++ b/spec/std/spec/junit_formatter_spec.cr
@@ -109,10 +109,14 @@ describe "JUnit Formatter" do
   it "escapes spec names" do
     output = build_report do |f|
       f.report Spec::Result.new(:success, %(complicated " <n>'&ame), __FILE__, __LINE__, nil, nil)
+      f.report Spec::Result.new(:success, %(ctrl characters follow - \r\n), __FILE__, __LINE__, nil, nil)
     end
 
     name = XML.parse(output).xpath_string("string(//testsuite/testcase[1]/@name)")
     name.should eq(%(complicated \" <n>'&ame))
+
+    name = XML.parse(output).xpath_string("string(//testsuite/testcase[2]/@name)")
+    name.should eq(%(ctrl characters follow - \\r\\n))
   end
 
   it "report failure stacktrace if present" do
--- a/spec/std/spec/junit_formatter_spec.cr
+++ b/spec/std/spec/junit_formatter_spec.cr
@@ -10,7 +10,7 @@ describe "JUnit Formatter" do
 
     expected = <<-XML
                  <?xml version="1.0"?>
-                 <testsuite tests="2" disabled="0" errors="0" failures="0" time="0.0">
+                 <testsuite tests="2" disabled="0" errors="0" failures="0" time="0.0" hostname="#{System.hostname}">
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something"/>
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something else"/>
                  </testsuite>
@@ -26,7 +26,7 @@ describe "JUnit Formatter" do
 
     expected = <<-XML
                  <?xml version="1.0"?>
-                 <testsuite tests="1" disabled="0" errors="0" failures="1" time="0.0">
+                 <testsuite tests="1" disabled="0" errors="0" failures="1" time="0.0" hostname="#{System.hostname}">
                    <testcase file="spec/some_spec.cr" classname="spec.some_spec" name="should do something">
                      <failure/>
                    </testcase>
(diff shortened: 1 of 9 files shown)


Sija:
Sorry for the ongoing changes, fwiw this is ready for review.

spec/std/spec/junit_formatter_spec.cr
```diff
@@ -3,14 +3,16 @@ require "xml"
...
+
+    output = build_report_with_mocked_timestamp(now) do |f|
       f.report Spec::Result.new(:success, "should do something", "spec/some_spec.cr", 33, nil, nil)
       f.report Spec::Result.new(:success, "should do something else", "spec/some_spec.cr", 50, nil, nil)
     end
 
     expected = <<-XML
                  <?xml version="1.0"?>
-                 <testsuite tests="2" disabled="0" errors="0" failures="0" time="0.0" hostname="#{System.hostname}">
+                 <testsuite tests="2" disabled="0" errors="0" failures="0" time="0.0" timestamp="#{now.to_rfc3339}" hostname="#{System.hostname}">
```
RX14: These tests don't really test anything except `build_report_with_mocked_timestamp` - we need a time mocking capability inside the stdlib.

I'd even rather that `build_report_with_mocked_timestamp` was `build_report_with_no_timestamp`, and it just took the timestamp out of the report entirely. At least that way it's obvious it's not being tested.

Sija: Done.

src/spec/junit_formatter.cr
```diff
@@ -18,7 +18,7 @@ module Spec
       io = @io
       io.puts %(<?xml version="1.0"?>)
       io << %(<testsuite tests=") << @results.size
-      io << %(" disabled=") << (@summary[:pending]? || 0)
+      io << %(" skipped=") << (@summary[:pending]? || 0)
```
RX14: Where's the specification for this format? Are we sure the rest of the formatter is up to spec>

Sija: I was using sources below as the reference:
- https://help.catchsoftware.com/display/ET/JUnit+Format
- https://stackoverflow.com/questions/4922867/what-is-the-junit-xml-format-specification-that-hudson-supports

Sija: Note that, before this PR `pending` tests were simply ignored. I chose to go with `skipped` because that's the only option for `<testcase>` besides `<error/>` and `<failure/>`.

RX14:
Thank you!

Sija:
Last commit should fix https://github.com/crystal-lang/crystal/pull/8617#issuecomment-568722915

Sija:
It's GTG from my side, dogfooding experiment in #8617 proved the capabilities to produce well-formed resulting xml.

Sija:
Additional commit is to help fixing https://github.com/crystal-lang/crystal/pull/8617#discussion_r361207627 - without it, we'd need to aggregate xml files from all of the runs into a single directory, whereas with `--junit_output_path` it's possible to do the same but directly.

RX14:
As far as I can tell, `--junit_output_path` tries to open the directory as a file...

Sija:
@RX14 no, why?

RX14:
oh, right, it's the old code which hardcoded `output.xml`, my bad

I'd rather change the semantics of the existing option. It's stupid as it is.

Sija:
> I'd rather change the semantics of the existing option. It's stupid as it is.

I was thinking the same, addressed in the last commit.

src/spec/junit_formatter.cr
```diff
@@ -100,7 +100,11 @@ module Spec
     end
 
     private def classname(result)
-      result.file.sub(%r{\.[^/.]+\Z}, "").gsub("/", ".").gsub(/\A\.+|\.+\Z/, "")
+      path = result.file
+      if relative_path = path.lchop?(Dir.current).try(&.lchop('/'))
```
straight-shoota: This path handling is very crude and will fail in a few edge cases.

Both paths need at least to be normalized before comparison. 
Unfortunately, we're still missing an efficient method to check if one path is a parent of another. `Path.new(path).parents.includes?(Path.new(Dir.current).normalize)` should to for now.

src/spec/junit_formatter.cr
```diff
@@ -41,14 +41,29 @@ module Spec
       JUnitFormatter.new(file)
     end
 
+    private def escape_xml_attr(value)
+      String.build do |io|
```
straight-shoota: Why not simply return `value.inspect_unquoted`?

spec/std/spec/junit_formatter_spec.cr
```diff
@@ -124,14 +154,19 @@ private def build_report
...
-private def exception_with_backtrace(msg)
-  begin
-    raise Exception.new(msg)
-  rescue e
-    e
+private def build_report_with_no_timestamp
+  output = build_report do |formatter|
+    yield formatter
   end
+  output.gsub(/\s*timestamp="(.+?)"/, "")
```
straight-shoota: I'd rather add a way to inject `@started_at` instead of tampering with the output.

The current workaround completely avoids testing the correctness of the `timestamp` attribute.

src/spec/junit_formatter.cr
Sija: Because we just wanna quote control characters.

spec/std/spec/junit_formatter_spec.cr
Sija: I concur, yet I'd do it as a part of a rewrite of the `Spec::JUnitFormatter` and its specs, to use `XML` library and more robust testing (not based on exact source code like it is atm)

src/spec/junit_formatter.cr
Sija: > This path handling is very crude and will fail in a few edge cases.

Could you specify those edge cases?

> Both paths need at least to be normalized before comparison.

There's no need to, since both are absolute normalized paths.

> Unfortunately, we're still missing an efficient method to check if one path is a parent of another. `Path.new(path).parents.includes?(Path.new(Dir.current).normalize)` should to for now.

True, but imho that's different issue.

straight-shoota: > Could you specify those edge cases?

Mostly windows paths, such as `\` as path separator, UNC shares etc.

> There's no need to, since both are absolute normalized paths.

How do you know? `Result#file` can be an almost arbitrary path. And I'm not sure `Dir#current` makes any guarantees for being normalized.

Adding a proper method is a different issue, but here we should use at least a correct workaround.

src/spec/junit_formatter.cr
straight-shoota: Yeah, right. We could consider making `String#dump_or_inspect_unquoted` public for such use cases. But it's fine for now.
How about `HTML.escape`, though?

spec/std/spec/junit_formatter_spec.cr
straight-shoota: Okay, so you're just adding a few features in this PR and plan to refactor the entire code later?
Why not refactor first?

Also please be aware that the JUnit formatter does not use `XML` *on purpose* to avoid requireing `libxml` as a dependency for every spec run (see #6004).

src/spec/junit_formatter.cr
Sija: > Yeah, right. We could consider making `String#dump_or_inspect_unquoted` public for such use cases. But it's fine for now.

👍

>  How about `HTML.escape`, though?

It is being escaped l8r on, these two things have nothin' in common.

spec/std/spec/junit_formatter_spec.cr
Sija: > Why not refactor first?
> 
> Also please be aware that the JUnit formatter does not use `XML` _on purpose_ to avoid requireing `libxml` as a dependency for every spec run (see #6004).

Exactly because of that. I'd like to re-open discussion about it. Crafting XML "by hand" comes hardly as a proper solution here.

src/spec/junit_formatter.cr
RX14: Why would `HTML.escape` not escape control characters??

spec/std/spec/junit_formatter_spec.cr
RX14: Well, re-adding libxml is not an acceptable solution either. This is not a proper solution, but it does work. Crafting a basic format by hand is fine.

src/spec/junit_formatter.cr
Sija: Because they have nuffin' to do with HTML...?

RX14: Well `HTML.escape` really applies XML escaping, does it not?

Sija: We need to escape control characters into their literals (`\n` => `\\n`), still nuffin' to do with XML/HTML escaping.

RX14: Oh, my bad, I assumed you were escaping these as XML entities.

spec/std/spec/junit_formatter_spec.cr
Sija: @straight-shoota How about adding specific test just for `timestamp` attribute? This way we could have more-or-less readable specs and tested attribute in the same time...

src/spec/junit_formatter.cr
Sija: Yes, this implementation won't handle such edge-cases indeed - as it was the case with the old one, btw.

straight-shoota: The previous implementation didn't even handle relative path resolution. Nevertheless a new solution should be correct anyway.

spec/std/spec/junit_formatter_spec.cr
straight-shoota: How's that better/easier to implement than mocking the timestamp property?

Sija: Mocking timestamp throughout the tests would necessitate adding this to every tests using exact matchers. Also, see my newest commit.

src/spec/junit_formatter.cr
Sija: I've force-pushed sth which should work as u mentioned.

bcardiff:
I'm having second thoughts on the following change

> * Changes —junit_output flag to take full path to the junit xml file, instead of just a directory

I checked other tools at the usual configuration is specifying a directory. As it was before.
Using a directory will allow future splitting in different files.

It seems that this change was motivated to grab the result of the splitted output of the std_specs, if so I would prefer to revert this part of the change and use different directories for the parts.

Am I missing something @Sija ?

Sija:
> I checked other tools at the usual configuration is specifying a directory. As it was before.

What other tools you have in mind?
https://github.com/sj26/rspec_junit_formatter as an example takes a full path, not a directory.

Also, it as @RX14 already mentioned, it doesn't make much sense to take only directory as an argument? Why won't users be allowed to name they files as they wish?

bcardiff:
* gradle uses a main configuration testReportDir
* junit has junit.outdir

I would prefer to have another cli option to indicate file (or file pattern) to be used within the output directory. The file pattern could be used to split results based , but scope creep.

This idea does not necesesarily requires a new option, but then at least what I want is that if the path option does not end up with `.xml` keep the old behavior and use it as a folder. with a fixed `output.xml` or `results.xml`.

RX14:
If we're going to revert behaviur i'd at least want an explanation for why these other tools do this.

Are we ever going to emit multiple files? What's the benefit.

bcardiff:
Splitting files is useful for reporting purposes. If the spec is used for acceptance test and they are grouped doing different files per group my help. Or it can be splitted by test result status directly. Or by namespace.

I can't say why other tools work like that. But I found it odd the change. And i discover becure the CircleCI test result integration broke on nightly on some projects.

If the path does not ends with .xml it should behave as before. The previous behaviour wasn't broken.

Issue #8594, opened 2019-12-17
Invalid memory access in Array#push


vlazar:
[Array#push fails](https://play.crystal-lang.org/#/r/87od)

```crystal
ary=[] of UInt32
p! ary.push(3)
```

Output
```
Invalid memory access (signal 11) at address 0x0
[0x55d1aecf1b06] *CallStack::print_backtrace:Int32 +118
[0x55d1aece4981] __crystal_sigfault_handler +209
[0x7f42fe0ad930] ???
[0x0] ???
```

[Array#<< works](https://play.crystal-lang.org/#/r/87of)

```crystal
ary=[] of UInt32
p! ary << 3 # => [3]
```

Note however that locally I have this error instead for Array#push:
```crystal
ary=[] of UInt32
p! ary.push(3) # => Program received and didn't handle signal ILL (4)
```

<details>
  <summary>Crystal 0.32.0 on MacOS</summary>

```
$ crystal --version
Crystal 0.32.0 (2019-12-12)

LLVM: 9.0.0
Default target: x86_64-apple-macosx
```
</details>

The behavior changed many times in previous versions of Crystal. Although I'm not sure if I can trust playground.

In Crystal 0.24.2 both [Array#push](https://play.crystal-lang.org/#/r/87og) and [Array#<<](https://play.crystal-lang.org/#/r/87oh) gives `no overload matches` error.

Since Crystal 0.25.0 [Array#push](https://play.crystal-lang.org/#/r/87oo) is broken and [Array#<<](https://play.crystal-lang.org/#/r/87on) works.

Since Crystal 0.27.0 [Array#push](https://play.crystal-lang.org/#/r/87oo) gives `ambiguous call, implicit cast of 3 matches all of UInt32` and [Array#<<](https://play.crystal-lang.org/#/r/87or) works.

SInce Crystal 0.29.0 [Array#push](https://play.crystal-lang.org/#/r/87p6) gives `Program received and didn't handle signal SYS (31)` and [Array#<<](https://play.crystal-lang.org/#/r/87p1) works.

SInce Crystal 0.31.0 [Array#push](https://play.crystal-lang.org/#/r/87p9) gives `Invalid memory access` and [Array#<<](https://play.crystal-lang.org/#/r/87pa) works.

vlazar:
Seem to be related to #6074 #6618 and #7747

vlazar:
Only `Int32` is working with `Array#push`

```crystal
# ary=[] of UInt64
[...]

PR #8600, opened 2019-12-19
Compiler: fix multidispatch with autocasting

### LINKED ISSUE #8217: Bug dispatching enum shorthand to splat and non-splat overloads
```cr
@[Flags]
enum Mode
  A
  B
  C
end

def foo(mode : Mode)
  puts "a"
end

def foo(*modes : Mode)
  puts "b"
end

foo(Mode::A)
```

correctly prints "a", but replacing the end with

```cr
foo(:a)
```

causes the program to segfault at runtime.

The codegen is invalid:
```llvm
  br i1 false, label %current_def, label %next_def, !dbg !79

current_def:                                      ; preds = %entry
  call void @"*foo<Mode>:Nil"(i32 1), !dbg !82
  br label %exit, !dbg !82

next_def:                                         ; preds = %entry
  br i1 false, label %current_def1, label %next_def2, !dbg !82

exit:                                             ; preds = %current_def1, %current_def
  ret void, !dbg !82

current_def1:                                     ; preds = %next_def
  call void @"*foo<Mode>:Nil"(i32 1), !dbg !82
  br label %exit, !dbg !82

next_def2:                                        ; preds = %next_def
  unreachable, !dbg !82
```

Crystal 0.31.0 and 0.30.1 tested.

Discovered in #8011.

asterite:
Fixes #8217 (and consequently #8594)

### DIFF
--- a/src/compiler/crystal/codegen/call.cr
+++ b/src/compiler/crystal/codegen/call.cr
@@ -359,7 +359,11 @@ class Crystal::CodeGenVisitor
           end
           node.args.each_with_index do |node_arg, i|
             a_def_arg = a_def.args[i]
-            result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            if autocast_literal?(node_arg)
+              result = and(result, int1(1))
+            else
+              result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            end
           end
 
           current_def_label, next_def_label = new_blocks "current_def", "next_def"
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/codegen/call.cr
```diff
@@ -359,7 +359,11 @@ class Crystal::CodeGenVisitor
           end
           node.args.each_with_index do |node_arg, i|
             a_def_arg = a_def.args[i]
-            result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            if autocast_literal?(node_arg)
+              result = and(result, int1(1))
```
bcardiff: I guess you preferred the explicitness of `result && true` and have as many `and` as args, rather than left result untouched `if autocast_literal?(node_arg)`. Right?

RX14:
I don't think this entirely fixes https://github.com/crystal-lang/crystal/issues/8217, since there's also the semantic (wrong type) issue there.

asterite:
Ah, right, it does a multidispatch when it shouldn't....

Issue #8595, opened 2019-12-17
NamedTuple Compilation Failure on 0.32.0


toddsundsted:
The following code compiles in versions before 0.32.0 but triggers a compiler failure/bug in version 0.32.0:

```
def invoke(**kwargs)
  kwargs.to_h.has_key?(:out)
end

invoke
```

Working example on 0.31.1:
https://play.crystal-lang.org/#/r/87r1

Failing example on 0.32.0:
https://play.crystal-lang.org/#/r/87r2

My local crystal:

```
$ crystal -v
Crystal 0.32.0 (2019-12-12)

LLVM: 9.0.0
Default target: x86_64-apple-macosx
```

But this fails in my Travis CI build, as well:

https://travis-ci.org/toddsundsted/mxnet.cr/builds/625304487

Stack trace from a locally built crystal compiler:

```
$ ./bin/crystal run /tmp/kwargs.cr
Using compiled compiler at .build/crystal
Module validation failed: Function return type does not match operand type of return inst!
  ret i8 %1, !dbg !73
 i1 (Exception)
  from src/llvm/module.cr:80:9 in 'verify'
  from src/compiler/crystal/codegen/codegen.cr:369:9 in 'finish'
  from src/compiler/crystal/codegen/codegen.cr:69:7 in 'codegen'
  from src/compiler/crystal/codegen/codegen.cr:65:5 in 'codegen:debug:single_module'
  from src/compiler/crystal/compiler.cr:22:7 in 'codegen'
  from src/compiler/crystal/compiler.cr:170:16 in 'compile'
  from src/compiler/crystal/command.cr:275:3 in 'compile'
  from src/compiler/crystal/command.cr:191:14 in 'run_command'
  from src/compiler/crystal/command.cr:92:7 in 'run'
  from src/compiler/crystal/command.cr:49:5 in 'run'
  from src/compiler/crystal/command.cr:48:3 in 'run'
  from src/compiler/crystal.cr:8:1 in '__crystal_main'
  from src/crystal/main.cr:97:5 in 'main_user_code'
  from src/crystal/main.cr:86:7 in 'main'
  from src/crystal/main.cr:106:3 in 'main'
Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
```

[...]

PR #8601, opened 2019-12-19
Fix for "NamedTuple Compilation Failure on 0.32.0"

toddsundsted:
Possible fix for https://github.com/crystal-lang/crystal/issues/8595.

Includes a test demonstrating the failure/bug as well as a fix. I'm not super-satisfied because it seems like the broken code **should** work, which indicates a possible root-cause elsewhere.

Even something as seemingly innocuous as adding an alternative that returns a literal `false` is enough to break it again:
```
if @compare_by_identity && entry_key.responds_to?(:object_id) && key.responds_to?(:object_id)
  entry_key.object_id == key.object_id
elsif @compare_by_identity
  false
else
  ...
  ...
```

Interestingly, return a `nil` succeeds:
```
if @compare_by_identity && entry_key.responds_to?(:object_id) && key.responds_to?(:object_id)
  entry_key.object_id == key.object_id
elsif @compare_by_identity
  nil
else
  ...
  ...
```

### DIFF
--- a/spec/std/hash_spec.cr
+++ b/spec/std/hash_spec.cr
@@ -1224,5 +1224,10 @@ describe "Hash" do
       h = ({} of String => Int32).compare_by_identity
       h.clone.compare_by_identity?.should be_true
     end
+
+    it "compiles" do
+      h = Hash(NoReturn, NoReturn).new
+      h.has_key?(nil).should be_false
+    end
   end
 end
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -935,11 +935,11 @@ class Hash(K, V)
         if key.responds_to?(:object_id)
           entry_key.object_id == key.object_id
         else
-          false
+          return false
         end
       elsif key.responds_to?(:object_id)
         # because entry_key doesn't respond to :object_id
-        false
+        return false
       else
         entry_key == key
       end
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -908,7 +908,7 @@ class Hash(K, V)
     else
       hash = key.hash.to_u32!
     end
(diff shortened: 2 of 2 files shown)


RX14:
If it's a compiler bug, the correct fix is in the compiler, not working around it in the stdlib.

I'm not sure the new `entry_matches?` logic is equivalent either way.

toddsundsted:
If it’s a compiler bug it would certainly be more satisfying to have it fixed in the compiler.

RX14:
> It looks like the code wants the comparison to fail if `compare_by_identity` is enabled and either argument doesn't support `object_id`

This was the intent. And I don't think we can drop this behaviour.

Merging a workaround id fine, but I don't think this is the right one.

toddsundsted:
@RX14 thanks! I think I have a fix that avoids the bugs and preserves existing behavior

RX14:
If others (mostly @asterite) are fine to change the behaviour to explicitly disallowing this then thats fine too

toddsundsted:
i changed the implementation to match the original logic. the only change, in fact, is explicit returns, which highlight the fact that the real flaw is likely elsewhere. interestingly, if you rewind all the way back to the commit where the bug was introduced, it manifests just by the introduction of the code, even if nothing at all is actually using the the `compare_by_identity` code path.

RX14:
Are you *sure* the explicit `return false` is required, not just the change in types?

asterite:
Thank you! But I'd like to fix the compiler bug instead.

toddsundsted:
@asterite awesome! sounds good

@RX14 thank you for your feedback!

Issue #8496, opened 2019-11-20
Implement cumulative folding


ghost:
Analog to `reduce`, but returns an `Enumerable` or `Iterator` containing the successive reduced values (e.g. running totals, maxima). The last element has the same value as `reduce`. Cumulative folding is also available in many other languages, see the following section:

### naming

Cumulative folding goes by many different names in various languages and tools out there:

- c++: [`partial_sum`, `{in,ex}clusive_scan`](https://en.cppreference.com/w/cpp/algorithm/partial_sum)
- clojure: [`reductions`](https://clojuredocs.org/clojure.core/reductions)
- d: [`cumulativeFold`](https://dlang.org/library/std/algorithm/iteration/cumulative_fold.html)
- haskell: [`scanl`](http://zvon.org/other/haskell/Outputprelude/scanl_f.html)
- matlab: [`cum{sum,min,max,prod}`](https://www.mathworks.com/help/matlab/ref/cumsum.html)
- python itertools: [`accumulate`](https://docs.python.org/3/library/itertools.html#itertools.accumulate)
- rust: [`scan`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan)
- compsci: [`prefix sums (running totals)`](https://en.wikipedia.org/wiki/Prefix_sum)

Using the plurals of functions included in the stdlib is merely a suggestion. Futhermore, `scan` is already associated with string scanning.

### usage examples

```cr
# running max:
[1,4,2,5,3].reduce     { |a, b| {a, b}.max } # => 5
[1,4,2,5,3].reductions { |a, b| {a, b}.max } # => [1, 4, 4, 5, 5]

[4,2,4].reduce     64 { |a, b| a // b } # => 2
[4,2,4].reductions 64 { |a, b| a // b } # => [16, 8, 2]

# also lazy iterations:
(1..).reductions { |a, b| a + b }.first(5).to_a # => [1, 3, 6, 10, 15]

# running sums; convenience variant of reductions, analog to sum vs reduce:
[1,2,3,4,5].sum  # => 15
[1,2,3,4,5].sums # => [1, 3, 6, 10, 15]

[1,2,3,4,5].sum  100 # => 115
[1,2,3,4,5].sums 100 # => [101, 103, 106, 110, 115]

# running products:
[1,2,3,4,5].product  # => 120
[1,2,3,4,5].products # => [1, 2, 6, 24, 120]
```

[...]

PR #8603, opened 2019-12-20
Adds cumulative to Enumerable and Array

wontruefree:
resolves: https://github.com/crystal-lang/crystal/issues/8496

This adds cumulative to Enumerable and Array.
This is overloaded on Array to use the existing array for an optimization.

### DIFF
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -507,6 +507,26 @@ describe "Array" do
     end
   end
 
+  describe "cumulative" do
+    it "returns a new array with cumulative reductions" do
+      a = [1, 4, 2, 5, 3]
+      (a.cumulative { |e| e.reduce { |a, b| {a, b}.max } }).should eq [1, 4, 4, 5, 5]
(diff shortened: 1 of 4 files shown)


asterite:
Thank you, but it seems this makes all operations be O(n^2). I don't think this should be merged. The idea of cumulative is to reuse previous results in the calculations.

straight-shoota:
I agree with @asterite. This implements a recent suggestion in #8496 which didn't get much feedback anyways. I'd rather prefer `#reductions` and `#each_reduction` as suggested in https://github.com/crystal-lang/crystal/issues/8496#issuecomment-559440350.

wontruefree:
There was an optimization added to array to make it more performant.  The big problem is with enumerable and only having each as the interface.

RX14:
It's about the fundamental algorithmic complexity of calling `yield` `O(n^2)`  times. There's no possible way to make the API you propose work, because people expect their (possibly side-effect having) block to be called once per item in the list.

PR #8605, opened 2019-12-21
Simplify LLVM_CONFIG shell macro

j8r:
It simplifies the LLVM_CONFIG macro in order make it easier to debug, figure out commands and versions priority and for adding support to new LLVM version.

Alpine Linux support is also added, which has a `llvm8-config` command.


bcardiff:
The Makefile should be also updated if we go with this change: https://github.com/crystal-lang/crystal/blob/255bfc5fa925b95b72e34b26ad997fb2b3f83059/Makefile#L39-L57

j8r:
Crystal compilation depending on Crystal itself, `crystal eval 'require "llvm"; printf LibLLVM::LLVM_CONFIG'` can be used.

j8r:
I hate Makefile. I tried to put the script in-line, doesn't work. I created a script file.

j8r:
Unexpected error on Bash 3 for Darwin: can't use `case` in `$()`.
The parser isn't smart enough to parse correctly `$(case "" in "") ;; esac)`

straight-shoota:
I'd prefer if we could avoid duplicating the code used by the Makefile and libllvm.

bin/find-llvm-config
```diff
@@ -0,0 +1,15 @@
...
+
+if ! LLVM_CONFIG=$(command -v "$LLVM_CONFIG"); then
+  llvm_config_version=$(llvm-config --version 2>/dev/null)
+  for version in 8.0 7.1 6.0 5.0 4.0 3.9 3.8; do
+    LLVM_CONFIG=$(
+    command -v llvm-config-${version%.*} || \
+    command -v llvm-config-$version || \
+    command -v llvm-config${version%.*}${version#*.} || \
+    command -v llvm${version%.*}-config || \
+    [ "${llvm_config_version#$version}" = "$llvm_config_version" ] || command -v llvm-config \
```
j8r: It's like `lchop`: if the string isn't changed, which means the version isn't included in `llvm_config_version`, stop.
`case` isn't used because can't be parsed by Bash 3 when inside `$()`, used in macOS.

j8r:
Sorry for all the push-forces, have to change CircleCI config for darwin - the modified `PATH` put in `BASH_ENV` wasn't taken into account by Make's `shell`.

Agree @straight-shoota, but I didn't find an other way.

src/llvm/lib_llvm.cr
```diff
@@ -1,25 +1,19 @@
...
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 4.0*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.9 || command -v llvm-config39 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.9*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.8 || command -v llvm-config38 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.8*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config
-                  `.chomp.stringify
+                  `if ! LLVM_CONFIG=$(command -v "$LLVM_CONFIG"); then
+                   llvm_config_version=$(llvm-config --version 2>/dev/null)
+                     for version in 8.0 7.1 6.0 5.0 4.0 3.9 3.8; do
```
Sija: I guess `9.0` could be added here too?

```suggestion
                     for version in 9.0 8.0 7.1 6.0 5.0 4.0 3.9 3.8; do
```

j8r: Don't know if it is supported, haven't tested.

RX14:
the code can't be duplicated between the `bin` script and the llvm library.

src/llvm/lib_llvm.cr
Sija: Isn't it? See #8358

src/llvm/lib_llvm.cr
```diff
@@ -1,25 +1,20 @@
...
+                       command -v llvm-config-${version%.*} || \
+                       command -v llvm-config-$version || \
+                       command -v llvm-config${version%.*}${version#*.} || \
+                       command -v llvm${version%.*}-config || \
+                       [ "${llvm_config_version#$version}" = "$llvm_config_version" ] || command -v llvm-config)
+                       [ "$LLVM_CONFIG" ] && break
+                     done
+                   fi
+                   printf "$LLVM_CONFIG"
+                  `.stringify
```
Sija: I'd embed the support file directly, addressing https://github.com/crystal-lang/crystal/pull/8605#issuecomment-568296496

```cr
read_file("#{__DIR__}../../bin/find-llvm-config")
```

src/llvm/lib_llvm.cr
```diff
@@ -1,26 +1,6 @@
...
-                   command -v llvm-config-4.0 || command -v llvm-config40 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 4.0*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.9 || command -v llvm-config39 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.9*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.8 || command -v llvm-config38 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.8*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config
-                  `.chomp.stringify
-                }}
+  LLVM_CONFIG = {{ system(read_file("#{__DIR__}/../../bin/find-llvm-config")).stringify }}
```
Sija:
```suggestion
  LLVM_CONFIG = {{ system(read_file("#{__DIR__}/../../bin/find-llvm-config")).chomp.stringify }}
```

j8r: Not needed, `printf` doesn't add a new line.

Sija: but `system` call does, doesn't it?

j8r: No, tested with `p {{ system(%(printf foo)).stringify`, and it prints `"foo"`, with no `\n` at the end.

j8r: No, tested with `p {{ system(%(printf foo)).stringify }}`, and it prints `"foo"`, with no `\n` at the end.

straight-shoota: `system` just returns what's written to stdout.

straight-shoota: Unfortunately, `bin/find-llvm-config` wouldn't be included in most distribution packages of crystal. Not even in our own ones. This would need an update in https://github.com/crystal-lang/distribution-scripts

So currently this code only works in the compiler repository. Using the libllvm bindings outside the compiler would fail. Since `LLVM` is public API, it needs to work anywhere.

One way to solve this is updating distribution-scripts and making sure all other package distributors make this change, too. Maybe there should be a simple rule that all files in `bin` should be included in the package?
The other option would be to move this helper script to `src/llvm/find-llvm-config`. This is obviously easier to achieve, and probably even a better solution, because it clearly belongs to the LLVM module.

j8r: Good catch, this will be one less thing to do for the packagers.

I will move the script to a `src/llvm/helpers` directory, in order to not have shell scripts along with Crystal code, a bit like the existing `ext`.

RX14: I'd be happy to put it in `src/llvm/ext`.

src/llvm/lib_llvm.cr
```diff
@@ -1,26 +1,6 @@
...
-                   command -v llvm-config-4.0 || command -v llvm-config40 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 4.0*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.9 || command -v llvm-config39 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.9*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.8 || command -v llvm-config38 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.8*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config
-                  `.chomp.stringify
-                }}
+  LLVM_CONFIG = {{ system(read_file("#{__DIR__}/ext/find-llvm-config")).stringify }}
```
RX14: why not just ``{{ `#{__DIR__}/ext/find-llvm-config`.stringify }}``??

j8r: Good idea!

src/llvm/lib_llvm.cr
```diff
@@ -1,26 +1,6 @@
...
-                   command -v llvm-config-4.0 || command -v llvm-config40 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 4.0*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.9 || command -v llvm-config39 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.9*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config-3.8 || command -v llvm-config38 || \
-                   (command -v llvm-config > /dev/null && (case "$(llvm-config --version)" in 3.8*) command -v llvm-config;; *) false;; esac)) || \
-                   command -v llvm-config
-                  `.chomp.stringify
-                }}
+  LLVM_CONFIG = {{ `#{__DIR__}/ext/find-llvm-config`.stringify }}
```
straight-shoota: Just a final note: Could we skip executing the helper script if `LLVM_CONFIG` is already defined as environment variable? This would make it more flexible. Shell scripts won't work on windows for example.
```suggestion
  LLVM_CONFIG = {{ ENV["LLVM_CONFIG"]? || `#{__DIR__}/ext/find-llvm-config`.stringify }}
```

j8r: Oops this syntax doesn't work in the macro-land

straight-shoota: Maybe we should change that to be less surprising ... 🤔

PR #8606, opened 2019-12-21
Edits in kernel.cr

fxn:
Minor details I saw while reading source code in `kernel.cr`.

* "consists from" should be "consists of".
* "use" → "uses" for consistency with the rest of the list items.
* "modifies" is not correct in "The flags modifies". Leveraged the edit to reword a bit.
* Removes periods in "Precision for..." sentences for consistency with similar surrounding ones.

The other two are clear.

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -180,9 +180,9 @@ end
 #       | The precision specifies the number of significant digits.
 #   G   | Equivalent to g, but use an uppercase E in exponent form.
 #   a   | Formats floating point argument as [-]0xh.hhhhp[+-]dd,
-#       | which is consisted from optional sign, "0x", fraction part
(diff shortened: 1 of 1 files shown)


src/kernel.cr
```diff
@@ -320,24 +321,24 @@ end
 # sprintf "%20.8b", -11   # => "               -1011"
 # ```
 #
-# Precision for `e` is number of digits after the decimal point.
+# Precision for `e` is number of digits after the decimal point
```
straight-shoota: I think I'd rather add full stops where they're missing, instead of removing them here.

fxn: @straight-shoota great. I ended up adding colons, which are used commonly to introduce code examples, and are also present in the code base.

PR #8610, opened 2019-12-22
[BUG] CSV cell String containing number only must be escaped with quotes

anykeyh:
## Current Behavior

Writing cell with a string number then importing in excel of libre office will let the tool treat it as a number. 

For example:

```crystal
require "csv"

str = CSV.build do |csv|
  csv.row "015551234" #Assuming first column is a phone number
end

str.to_s # "015551234"
```

This would be imported as `15551234` in Excel of Libre Office.

## Expected behavior

The output should be quoted: `"015551234"`.

Currently there's zero way to quote this number, as using `"` character will lead to quote escaping:

```crystal
require "csv"

str = CSV.build do |csv|
  csv.row "\"015551234\"" 
end

str.to_s # ""015551234""
```

@ Crystal team: I'm just waiting for direction (actually I'm not sure about RFC for CSV on theses cases) before creating specs and real PR

Thanks :)

### DIFF
--- a/src/csv/builder.cr
+++ b/src/csv/builder.cr
@@ -158,6 +158,7 @@ class CSV::Builder
     private def needs_quotes?(value : String)
       case @quoting
       when .rfc?
+        return true if value =~ /\d+/ 
         value.each_byte do |byte|
           case byte.unsafe_chr
           when @separator, @quote_char, '\n'


RX14:
How about just introducing an `:all_strings` quoting mode, which quotes all string values.

src/csv/builder.cr
```diff
@@ -158,6 +158,7 @@ class CSV::Builder
     private def needs_quotes?(value : String)
       case @quoting
       when .rfc?
+        return true if value =~ /\d+/
```
asterite: Why not `if value.starts_with?('0')`? A regex is overkill here.

anykeyh:
Actually I'm closing this PR as the change requested DO NOT change anything on the import in Excel nor Libre office. 

Both continue to treat the quoted string AS number unless specific import rules, and there's no way to force the cells as string without this manual operation.

Sorry, I should have checked before; I was so sure quoting the number should force it as string, but office softwares live in another world !

bcardiff:
Usually prefixing numbers with single quotes helps excel: `"'42"`.

Due to the untyped nature of csv is hard to propose a default that will work for everyone.

Having converters based on column name, column index and data type can be helpful. But is not much very different from doing an explicit call.

PR #8616, opened 2019-12-24
Custom yaml tags

### LINKED ISSUE #8598: Custom YAML tag processors
Currently the YAML parser handles a static set of [tags](https://github.com/crystal-lang/crystal/blob/master/src/yaml/schema/core.cr#L275-L292).  I propose to allow custom handlers to be registered to allow for handling custom tags.

This would also allow the stdlib to register some utility tags, like `env: !crystal/env APP_ENV` for example.

Blacksmoke16:
Resolves #8598

Implementation TBD, but this is at least a start.  Assumes tags are registered globally and affect all `YAML.parse` and `from_yaml` calls.

Also implements custom `!crystal/env` and `!crystal/env?` tags within the stdlib itself.

TODO:
- [ ] I suppose I should allow custom handlers in the failsafe schema, just cast the result to strings.

### DIFF
--- a/src/yaml.cr
+++ b/src/yaml.cr
@@ -75,6 +75,14 @@ require "base64"
 # File.open("foo.yml", "w") { |f| {hello: "world"}.to_yaml(f) } # writes it to the file
 # ```
 module YAML
+  private alias HandlerResponse = Bool | Float64 | Int64 | String | Time | Nil
+
+  class_getter tag_handlers : Hash(String, Proc(String, HandlerResponse)) { Hash(String, Proc(String, HandlerResponse)).new }
+
+  def self.add_tag_handler(tag : String, &block : String -> HandlerResponse) : Nil
+    tag_handlers[tag] = block
+  end
+
   class Error < Exception
   end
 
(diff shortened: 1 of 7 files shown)


lbguilherme:
Having the ability to leak any ENV variable by simply using `YAML.parse` on untrusted input *by default* is a huge security issue. Please make sure anything like it is opt-in.

Security aside, I don't think the stdlib should provide custom predefined tags like those.

src/yaml.cr
```diff
@@ -75,6 +75,14 @@ require "base64"
 # File.open("foo.yml", "w") { |f| {hello: "world"}.to_yaml(f) } # writes it to the file
 # ```
 module YAML
+  private alias HandlerResponse = Bool | Float64 | Int64 | String | Time | Nil
+
+  class_getter tag_handlers : Hash(String, Proc(String, HandlerResponse)) { Hash(String, Proc(String, HandlerResponse)).new }
```
Sija: Lazy instantiation here is pointless anyway due to a `YAML.tag_handlers[tag]?` call.

```suggestion
  class_getter tag_handlers = Hash(String, Proc(String, HandlerResponse)).new
```

RX14:
I think this still needs a lot more discussion in #8598 .

Blacksmoke16:
Closing until an implementation can be decided upon within #8598.

Issue #8459, opened 2019-11-09
Incorrect String#gsub behavior


igor-alexandrov:
I need to escape all whitespace chars with '\'.

```
"two words".gsub(/\s+/, "\ ")
```

Expected result: ```"two\ words"```.
Actual result: ```"two words"```.

Crystal 0.31.1 (2019-10-02)

LLVM: 8.0.1
Default target: x86_64-apple-macosx

straight-shoota:
`gsub` behaves totally correct.

Did you look at the string replacement? It is actually `two words`.
In string literals the backslash character denotes an escape sequence and by default escapes the next character as itself (unless it is one of the special escape sequences). See https://crystal-lang.org/reference/syntax_and_semantics/literals/string.html for details.

In order to write a literal backslash, you can use the backslash escape `\\`: `"two\\ words`.
Alternatively, there is also a non-interpolation string literal: `%q[two\ words]`.

igor-alexandrov:
Got it!

RX14:
Should we just make invalid escapes be syntax errors, so they're reserved for additional escapes possibly in the future. Otherwise it's always a breaking change to add an escape, and these kinds of errors can pop up.

straight-shoota:
I suppose that behaviour is simply a Ruby legacy. It's a valid option, though. I don't think there are many escape sequences that could possibly be added, so the breaking changes should not be a huge issue.

The possible alternatives for unrecognized escape sequences are:
* Ignore the backslash (current behaviour)
* Treat the backslash as literal backslash (Python)
* Syntax error (C)

asterite:
I think syntax error is good.

igor-alexandrov:
Should we reopen it?

straight-shoota:
No, better make it a new issue.

asterite:
I think it should be a different issue, unrelated to gsub.

PR #8619, opened 2019-12-24
Formatter: remove unnecessary escape sequences 

RX14:
After https://github.com/crystal-lang/crystal/issues/8459, I'd like to make invalid escape sequences a syntax error. The first step towards doing that is to introduce a warning, and to make the formatter automatically correct this mistake.

However, I found plumbing the warning apparatus (in `Program`, which is a hell class) through to the `Parser` painful, so I left it out for now.

I'd like feedback on whether a warning should necessarily be required before making this change, given that the formatter can fix it automatically. And if not, how is the best way to plumb this (duplicate the code, or pass the whole `Program` into the `Parser`?)

### DIFF
--- a/spec/compiler/crystal/tools/doc/generator_spec.cr
+++ b/spec/compiler/crystal/tools/doc/generator_spec.cr
@@ -122,7 +122,7 @@ describe Doc::Generator do
         a_def.doc = "Some Method"
         a_def.add_annotation(program.deprecated_annotation, Annotation.new(Crystal::Path.new("Deprecated"), ["don't use me".string] of ASTNode))
         doc_method = Doc::Method.new generator, doc_type, a_def, false
-        doc_method.formatted_summary.should eq %(<p>Some Method</p>\n\n<p><span class=\"flag red\">DEPRECATED</span>  don't use me</p>\n\n)
+        doc_method.formatted_summary.should eq %(<p>Some Method</p>\n\n<p><span class="flag red">DEPRECATED</span>  don't use me</p>\n\n)
       end
     end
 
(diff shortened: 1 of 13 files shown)


RX14:
Anyone want to discuss whether a warning is required for this?

bcardiff:
There is no context shared between parsing and compiling, where the deprecation.
Something like that might help to remove some duplicates between the cli and program.

Given the nature of this change, I say that emitting always the warnings to STDERR is enough.
And I can tolerate the breaking change without warning. The fix is syntactic, there is no documentation or semantic change. The fix should be able to be inferred locally:

`\" invalid escape sequence, did you mean "?`

seems helpful enough, plus applying the change today works on latest. So people can have a codebase working on latest+nightly.

RX14:
Okay, so should we merge this formatter fix for next release, then make this a compile error a release after?

straight-shoota:
@RX14 Needs a rebase.

bcardiff:
I think that making an invalid char escape sequence a compiler error is an improvement. 
I'm not sure the formatter should automatically fix the code in this case though.

As of 0.32.1 `'\s'` is a parser error but `"\s"` is not (same with other invalid char escape sequence)

Wouldn't it be better to make `"\s"` a parser error? 
And keep the formatter away from changing the code?

RX14:
@bcardiff you can see it's an existing error in the codebase, I really wanted a formatter release which allows people to easily fix all the occurences of this.

bcardiff:
So on a future release, the invalid escape sequence within a string will cause a parser error and the formatter will stop correcting this. Right?

If so, this pr is fine as is, no need for a warning.

RX14:
@bcardiff yeah that was my intention

This has been rebased and reformatted, so a quick merge would be nice

PR #8620, opened 2019-12-25
Tweak CI scripts a bit

Sija:
This PR removes an additional (?) `make std_spec clean` call in `bin/ci build` task.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -684,4 +684,3 @@ workflows:
             - dist_darwin
             - dist_snap
             - dist_docs
-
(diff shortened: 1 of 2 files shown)


ysbaddaden:
AFAIK it was added to have a "quick" fail when std spec fails (with current crystal), then run 'em again with crystal master, along with compiler specs.

straight-shoota:
This is necessary to ensure std_spec passes with the latest compiler release.

j8r:
I think a comment in the script explaining the rationale wouldn't hurt then :)

PR #8621, opened 2019-12-25
Fix Enumerable#zip to work with union types

### LINKED ISSUE #8608: Indexable itself cannot be used for Enumerable#zip
With the following definition, it says that `Enumerable#zip` accepts `Indexable` (and objects include it), but `Indexable` itself cannot be used as an argument of `Enumerable#zip`, and getting error on the internal macro expansion.
https://github.com/crystal-lang/crystal/blob/8dd327103dc7447ca966e7118853762f60cbfdf7/src/enumerable.cr#L1507-L1546

I wanted to create a generic class which accept two any objects which can be used for `#zip` (with other data).
```crystal
class Foo(T)
  def initialize(@a : Indexable(T) | Iterable(T) | Iterator(T), @b : Indexable(T) | Iterable(T) | Iterator(T))
  end
  
  def some_work
    @a.zip(@b) do |a, b|
      puts a, b
    end
  end
end
 
f = Foo.new([1.0, 2.0, 3.0], [4.0, 5.0, 6.0])
f.some_work
```
https://play.crystal-lang.org/#/r/88yn

Getting following error:
```

There was a problem expanding macro 'macro_140238057131296'

Code in /usr/lib/crystal/enumerable.cr:1648:5

 1648 | {% begin %}
        ^
Called macro defined in /usr/lib/crystal/enumerable.cr:1648:5

 1648 | {% begin %}

Which expanded to:

 > 10 |           
 > 11 |             # Otherwise advance the iterator
 > 12 |             other_elem0 = iter0.next
                                  ^----
Error: undefined local variable or method 'iter0' for Enumerable(T):Module
```

I have not checked for `Iterable` nor `Iterator`.

asterite:
Fixes #8608

Using runtime `is_a?` instead of compile-time checks makes it work for union types.

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1646,30 +1646,35 @@ module Enumerable(T)
   # :nodoc:
   def self.zip(main, others : U, &block) forall U
     {% begin %}
+      {% for type, type_index in U %}
+        other{{type_index}} = others[{{type_index}}]
+      {% end %}
+
       # Try to see if we need to create iterators (or treat as iterators)
       # for every element in `others`.
       {% for type, type_index in U %}
-        {% if type < Indexable %}
+        case other{{type_index}}
+        when Indexable
           # Nothing to do, but needed because many Indexables are Iterable/Iterator
-        {% elsif type < Iterable %}
-          iter{{type_index}} = others[{{type_index}}].each
-        {% elsif type < Iterator %}
-          iter{{type_index}} = others[{{type_index}}]
-        {% end %}
+        when Iterable
+          iter{{type_index}} = other{{type_index}}.each
+        else
+          iter{{type_index}} = other{{type_index}}
+        end
       {% end %}
 
       main.each_with_index do |elem, i|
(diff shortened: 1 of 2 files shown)


src/enumerable.cr
```diff
@@ -1646,30 +1646,34 @@ module Enumerable(T)
...
     {% begin %}
+      {% for type, type_index in U %}
+        other{{type_index}} = others[{{type_index}}]
+      {% end %}
+
       # Try to see if we need to create iterators (or treat as iterators)
       # for every element in `others`.
       {% for type, type_index in U %}
-        {% if type < Indexable %}
+        if other{{type_index}}.is_a?(Indexable)
```
Sija: `case` would be imo tad moar cleaner here:

```cr
case other{{type_index}}
when Indexable
  # ...
when Iterable
  # ...
end
```

ditto below

asterite: Thanks, fixed. And also did it a bit differently...

PR #4114, opened 2017-03-04
Parser: parse var as Call evan if var is declared in call args

MakeNowJust:
Fixed #4106

Issue #4106, opened 2017-03-04
Improve error message for "undefined local variable or method" when variable is previosuly defined because of "var : Type"


sh01ch1:
```crystal
module Mod
  class Klass
    getter name : String

    def initialize(@name)
    end
  end

  puts name
end
```

raises:

```
Error in test.cr:9: read before assignment to local variable 'name'

  puts name
       ^~~~
```

Crystal 0.21.0, mac os x yosemite

MakeNowJust:
What's a problem? You just did **read before assignment to local variable 'name'**, so you got the error of course.

Crystal needs a assignment against the local variable before using this, not like Ruby.

sh01ch1:
Hi @MakeNowJust, Module has the `.name` method. The following code works:

```crystal
module Mod
  class Klass
    getter name : String

    def initialize(@name)
    end
  end

  puts self.name # prints "Mod"
end
```

MakeNowJust:
Small version:

```crystal
module A
  module B
    name : String
  end
  p name
end
```

MakeNowJust:
Another version:

```crystal
def foo
  42
end

module Foo
  foo : Int32
end

puts foo
```

The problem is caused when same name variable is defined in class/module because `foo` is parsed as `Var` instaed of `Call` in this case.

asterite:
Maybe the error message should be improved, but it will still be an error. I think this is minor.

sh01ch1:
Hi @asterite, I'm a little confusing.

```crystal
def foo
  "foo"
end

class Klass
  @foo : String

  def foo : String
    @foo
  end

  def initialize(@foo)
  end
end

puts foo
```

is OK. 

```crystal
def foo
  "foo"
end

class Klass
  getter foo : String

  def initialize(@foo)
  end
end

puts foo
```
Why does this code raise "read before assignment to local variable 'foo'"?

MakeNowJust:
@sh01ch1 Crystal's scope rule looks like LISP-2, in short it decides whether a name is a variable or a method on parsing (It is same as Ruby).

Then, how does it decide this? -- It is simple: the parser memorizes declared or assigned variable names on parsing, and it decides a name is variable if this name is memorized.

On your example, `foo : String` is a variable declaration of `foo`, so the parser memorizes `foo` is variable. The compiler treats `foo` of `p foo` as variable, but `foo` assignment is not found, so you got such an error.

[...]

PR #8622, opened 2019-12-25
Parser: push var for type declaration

### LINKED ISSUE #8611: TypeDeclaration variable in macro argument is not viewed as defined variable
The code:

```Crystal
macro let(*vars)
  {% for v in vars %}
    {% if v.is_a?(Assign) %}
      {{ v }}
    {% elsif v.is_a?(TypeDeclaration) %}
      {{ v }}
    {% elsif v.is_a?(UninitializedVar) %}
      {{ v }}
    {% end %}
  {% end %}
end

macro embed(&block)
end

z : Int32 = 1
let x = 1, y : Int32 = 1

embed do
  x += 2
  z += 1
  y += 1
end
```

The error:
```
Using compiled compiler at /home/developer/workspace/.build/crystal                                                                                                           
In /home/developer/workspace/tmp/test2.cr:22:7                                                                                                                                
                                                                                                                                                                              
 22 | y += 1                                                                                                                                                                  
          ^                                                                                                                                                                   
Error: '+=' before definition of 'y'
```

asterite:
Fixes #8611

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -1835,6 +1835,8 @@ module Crystal
 
       it_parses %(annotation Foo\nend\nrequire "bar"), [AnnotationDef.new("Foo".path), Require.new("bar")]
 
+      it_parses "id(x : Int32 = 1); x", Expressions.new([Call.new(nil, "id", TypeDeclaration.new("x".var, "Int32".path, 1.int32)), "x".var])
+
       it "gets corrects of ~" do
         node = Parser.parse("\n  ~1")
         loc = node.location.not_nil!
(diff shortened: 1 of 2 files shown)


asterite:
It seems this conflicts with #4114 and #4106

I don't know what to do. Maybe reverting #4114 is good because #8611 will start working and #4106 will still give an error message, though not the ideal error message (but probably we can change the error message to say "undefined local variable blah blah").

RX14:
I was trying to search for that issue, to bring it up against this change. I think being able to declare variables this way isn't necessary, and the previous behaviour was correct. In general, I don't think macros should be allowed to declare local variables.

In the absence of finding any past discussion though, I did approve this PR.

asterite:
It does work if you assign to a variable, though. That's why I think it should work in all cases, or in none.

RX14:
> That's why I think it should work in all cases, or in none.

I agree.

asterite:
Closing because this didn't work.

Issue #8624, opened 2019-12-26
File.size and File.info do not work with Path parameters


snluu:
Compilation fails because the error code path tries to call `inspect_unquoted` on the parameter, which is only available for strings.

Sija:
Can be closed.

PR #8625, opened 2019-12-26
Fix File.size and File.info to work with Path parameters

snluu:
Currently, `File.size` and `File.info` do not work with `Path` parameter types because the error path tries to call `inspect_unquoted`, which is only available for `String`s. 

This path simply call `to_s` on the `Path` parameters to convert them back to strings.


snluu:
This fixes #8624 8624

PR #8634, opened 2019-12-28
Allow to create an enum from a macro symbol

### LINKED ISSUE #8573: Allow to create an enum from a macro symbol
Example usecase:
```cr
enum SomeEnum
  A
  B

  # add this!
  def self.from_self(value : self)
    value
  end
end

# to make this work:
{% begin %}
  {% value = :b %} # get the symbol from somewhere..
  case SomeEnum.from_self({{ value }})
  when .a?
    puts "It's an A !"
  when .b?
    puts "It's a B !"
  end
{% end %}
```
https://carc.in/#/r/86ip

Crystal 0.31.1

bew:
Closes #8573

### DIFF
--- a/spec/std/enum_spec.cr
+++ b/spec/std/enum_spec.cr
@@ -37,14 +37,8 @@ describe Enum do
     end
   end
 
-  it "creates an enum instance from a (macro) symbol (#8573)" do
-    enum_value = SpecEnum.new(:one)
-    enum_value.should eq SpecEnum::One
-
-    enum_value =
-      {% begin %}
-        SpecEnum.new({{ :two }})
-      {% end %}
+  it "creates an enum instance from an auto-casted symbol (#8573)" do
+    enum_value = SpecEnum.new(:two)
     enum_value.should eq SpecEnum::Two
   end
 
--- a/spec/std/enum_spec.cr
+++ b/spec/std/enum_spec.cr
@@ -37,6 +37,14 @@ describe Enum do
     end
   end
 
+  it "creates an enum instance from a macro symbol" do
+    enum_value =
+      {% begin %}
+        SpecEnum.from_self({{ :two }})
+      {% end %}
+    enum_value.should eq SpecEnum::Two
+  end
+
   it "gets value" do
     SpecEnum::Two.value.should eq(1)
     SpecEnum::Two.value.should be_a(Int8)
--- a/spec/std/enum_spec.cr
+++ b/spec/std/enum_spec.cr
@@ -37,10 +37,13 @@ describe Enum do
     end
   end
 
-  it "creates an enum instance from a macro symbol" do
+  it "creates an enum instance from a (macro) symbol (#8573)" do
+    enum_value = SpecEnum.new(:one)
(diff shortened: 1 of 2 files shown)


spec/std/enum_spec.cr
```diff
@@ -37,6 +37,14 @@ describe Enum do
     end
   end
 
+  it "creates an enum instance from a macro symbol" do
+    enum_value =
+      {% begin %}
+        SpecEnum.from_self({{ :two }})
+      {% end %}
```
bew: Note: I'm not sure why I need to use macro begin/end here, otherwise it fails with:
`no overload matches 'SpecEnum.from_self' with type Symbol`
I guess it's because auto-casted symbols are resolved earlier and not when the method is actually called?

src/enum.cr
```diff
@@ -89,6 +89,26 @@
...
+  #   {% macro_sym = :red %} # get the macro symbol from somewhere..
+  #   case Color.from_self({{ macro_sym }})
+  #   when .red?
+  #     puts "It's red !"
+  #   when .blue?
+  #     puts "It's blue !"
+  #   end
+  # {% end %}
+  # ```
+  def self.from_self(value : self)
```
asterite: This should be called `new` and I wouldn't mention any specific use case in mind. I think the method is generally useful.

bew:
Pretty weird CI error for `test_linux`:
```
.build/std_spec: relocation error: .build/std_spec: symbol __gmon_start__ version LIBXML2_2.4.30 not defined in file libxml2.so.2 with link time reference
```

src/enum.cr
bew: Okay, how should I document the method then?

asterite: Returns `value`. :-)

bew: Done thank you!

spec/std/enum_spec.cr
```diff
@@ -37,6 +37,17 @@ describe Enum do
     end
   end
 
+  it "creates an enum instance from a (macro) symbol (#8573)" do
```
Sija:
```suggestion
  it "creates an enum instance from a symbol (#8573)" do
```

bew: Well I'd like to keep the wording and test the original usecase I had in mind (creating an enum instance from a macro symbol)

Sija: IMO real usecase here is initializing enums from symbols, doing this from within the macro-generated code is just a technical detail.

thus, you could also remove the `{% begin ... end %}` macro from the specs.

asterite:
To be honest I don't understand the use case. Why would you get a symbol from somewhere? And if you do get a symbol, why don't you compare symbols instead of enums?

spec/std/enum_spec.cr
bew: That's fair, I'll make the edits when I can

bew:
@asterite in my case I had an annotation with symbols that should map to real enum values.
And then generated some code which should do various things based on the given enum values.

bew:
Thanks!

PR #8641, opened 2020-01-02
CI: use lld if present (Linux only)

### LINKED ISSUE #7177: symbol lookup error

bcardiff:
EXPORT_CC value can be overridden to disable lld lookup.

* Fixes #7177 by using lld if present.
* `crystallang/crystal:0.32.1-build` and `crystallang/crystal:0.32.1-i386-build` docker images have been updated to include lld-8 according to https://github.com/crystal-lang/crystal-dist/pull/8
* [All-required-libraries wiki](https://github.com/crystal-lang/crystal/wiki/All-required-libraries) has been updated to _suggest_ the installation of lld.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,10 @@ DEPS = $(LLVM_EXT_OBJ) $(LIB_CRYSTAL_TARGET)
 CFLAGS += -fPIC $(if $(debug),-g -O0)
 CXXFLAGS += $(if $(debug),-g -O0)
 
+ifeq ($(shell command -v ld.lld >/dev/null && uname -s),Linux)
+  EXPORT_CC ?= CC="cc -fuse-ld=lld"
+endif
+
 ifeq (${LLVM_CONFIG},)
   $(error Could not locate llvm-config, make sure it is installed and in your PATH, or set LLVM_CONFIG)
 else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -65,6 +65,10 @@ DEPS = $(LLVM_EXT_OBJ) $(LIB_CRYSTAL_TARGET)
 CFLAGS += -fPIC $(if $(debug),-g -O0)
 CXXFLAGS += $(if $(debug),-g -O0)
 
+ifeq ($(shell uname -s),Linux)
+  EXPORT_CC ?= $(if $(shell command -v ld.lld),CC="cc -fuse-ld=lld")
+endif
+
```
RX14: How about just

```makefile
ifeq ($(shell command -v ld.lld >/dev/null && uname -s),Linux)
    CC ?= cc -fuse-ld=lld
endif
```

then leaving off all the stuff down the page.

bcardiff: I updated the condition but the CC and other [vars are not exported automatically](https://www.linuxquestions.org/questions/programming-9/exporting-makefile-variables-to-%24-shell-environment-807422/)

That is also a reason why an `EXPORT` variable is used somewhere else in the Makefile.

straight-shoota: Sounds reasonable. But when overriding `CC`, you'd need to remember to add `-fuse-ld=lld` explicitly...

RX14: Writing `export CC` works, and is much cleaner.

RX14:
Please don't merge things before I have had a chance to respond to review comments!

straight-shoota:
I agree. I've approved the PR because it seemed fine to me, but I also expected we'd wait for your approval before merging. Maybe there's a communication error and @bcardiff assumed I was sending my approval to merge this immediately.
This has been broken for quite some time so it shouldn't matter whether we merge it a few hours sooner or later.

In general, I'd prefer to wait at least 24 hours before merging a PR to giver others a chance to review and comment. I'm not sure this should go into a policy or something, but it could help to think about that. While it's not a huge issue to apply changes after the PR has been merged via a second PR, it can still cause quite some friction which can be avoided by simply postponing a merge for a few hours. 😉

bcardiff:
I couldn't make it work with 

```
  CC ?= cc -fuse-ld=lld
  export CC
```

Adding a `@env | grep CC || true` to check which the value supposed to be used shows that the default `CC=cc`  is used.

It seems that overriding `CC` can't be done with `?=`.

It also leads to less explicit output. In the current PR the `CC="cc -fuse-ld=lld"` is displayed explicitly.

Since this is somehow a workaround that might affect some, and using lld is optional I think is better to have an explicit output to allow diagnostic whether lld is used or not.

This trial and error + the approval encourage me to merge it.

Probably I am missing something in my makefile-fu-ness. But I couldn't improve the current solution.

RX14:
Looks like `make` has a "prelude" with `CC ?= cc` in it. So I think the best way to solve this is just to use `CC += -fuse-ld=lld`.

PR #8642, opened 2020-01-02
Rename Process kill methods to signal

jan-zajic:
As suggested by @ysbaddaden in #8518 and further discussed in draft #8536 rename Process#kill(Signal) to Process#signal(Signal) that we can eventually implement on Windows (if needed) with a deprecation notice. The name kill is confusing as it can means two things - send signal to process, or send signal kill (9) to process when argument ommited. But currently default behavior of kill method without providing signal argument is to send term signal to process (simulates posix behavior of kill command, but we need to remove POSIXisms of apis to become realy portable imho).

### DIFF
--- a/spec/std/process_spec.cr
+++ b/spec/std/process_spec.cr
@@ -159,24 +159,24 @@ describe Process do
     end
   end
 
-  describe "kill" do
+  describe "signal" do
     it "kills a process" do
       process = Process.new("yes")
-      process.kill(Signal::KILL).should be_nil
+      process.signal(Signal::KILL).should be_nil
     end
 
     it "kills many process" do
       process1 = Process.new("yes")
       process2 = Process.new("yes")
-      process1.kill(Signal::KILL).should be_nil
-      process2.kill(Signal::KILL).should be_nil
+      process1.signal(Signal::KILL).should be_nil
+      process2.signal(Signal::KILL).should be_nil
     end
   end
 
   it "gets the pgid of a process id" do
     process = Process.new("yes")
     Process.pgid(process.pid).should be_a(Int32)
-    process.kill(Signal::KILL)
+    process.signal(Signal::KILL)
     Process.pgid.should eq(Process.pgid(Process.pid))
   end
 
--- a/spec/std/process_spec.cr
+++ b/spec/std/process_spec.cr
@@ -228,6 +228,20 @@ describe Process do
     process.terminated?.should be_true
   end
 
+  it "checks for existence after terminate" do
+    process = Process.new("yes")
+    process.exists?.should be_true
+    process.terminated?.should be_false
+
+    process.terminate
+    process.exists?.should be_true
+    process.terminated?.should be_false
+
+    process.wait
+    process.exists?.should be_false
+    process.terminated?.should be_true
+  end
+
   describe "executable_path" do
     it "searches executable" do
       Process.executable_path.should be_a(String | Nil)
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -161,7 +161,7 @@ module Crystal::Playground
         @logger.info "Code execution killed (session=#{@session_key}, filename=#{@running_process_filename})."
         @process = nil
         File.delete @running_process_filename rescue nil
-        process.kill rescue nil
+        process.signal rescue nil
       end
     end
 
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -161,7 +161,7 @@ module Crystal::Playground
         @logger.info "Code execution killed (session=#{@session_key}, filename=#{@running_process_filename})."
         @process = nil
         File.delete @running_process_filename rescue nil
-        process.signal rescue nil
+        process.signal(::Signal::TERM) rescue nil
       end
     end
 
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -161,7 +161,7 @@ module Crystal::Playground
         @logger.info "Code execution killed (session=#{@session_key}, filename=#{@running_process_filename})."
         @process = nil
         File.delete @running_process_filename rescue nil
-        process.signal(::Signal::TERM) rescue nil
+        process.terminate rescue nil
       end
     end
 
--- a/src/process.cr
+++ b/src/process.cr
@@ -35,8 +35,17 @@ class Process
     LibC.getppid
   end
 
-  # Sends a *signal* to the processes identified by the given *pids*.
+  @[Deprecated("Use #signal instead")]
   def self.kill(signal : Signal, *pids : Int)
+    pids.each do |pid|
+      ret = LibC.kill(pid, signal.value)
+      raise Errno.new("kill") if ret < 0
+    end
+    nil
+  end
+
+  # Sends a *signal* to the processes identified by the given *pids*.
+  def self.signal(signal : Signal, *pids : Int)
     pids.each do |pid|
       ret = LibC.kill(pid, signal.value)
       raise Errno.new("kill") if ret < 0
@@ -193,7 +202,7 @@ class Process
       $? = process.wait
       value
     rescue ex
-      process.kill
+      process.signal
       raise ex
     end
   end
@@ -349,9 +358,14 @@ class Process
     @wait_count = 0
   end
 
-  # See also: `Process.kill`
+  @[Deprecated("Use #signal instead")]
   def kill(sig = Signal::TERM)
-    Process.kill sig, @pid
+    signal sig
+  end
+
+  # See also: `Process.signal`
+  def signal(sig = Signal::TERM)
+    Process.signal sig, @pid
   end
 
   # Waits for this process to complete and closes any pipes.
--- a/src/process.cr
+++ b/src/process.cr
@@ -199,7 +199,7 @@ class Process
       $? = process.wait
       value
     rescue ex
-      process.signal Signal::TERM
+      process.terminate
       raise ex
     end
   end
@@ -397,6 +397,11 @@ class Process
     close_io @input
     close_io @output
     close_io @error
+  end
+
+  # Ask process for terminate gracefully
+  def terminate
+    signal Signal::TERM
   end
 
   # :nodoc:
--- a/src/process.cr
+++ b/src/process.cr
@@ -35,21 +35,19 @@ class Process
     LibC.getppid
   end
 
+  # Sends a *signal* to the processes identified by the given *pids*.
   @[Deprecated("Use #signal instead")]
   def self.kill(signal : Signal, *pids : Int)
     pids.each do |pid|
-      ret = LibC.kill(pid, signal.value)
-      raise Errno.new("kill") if ret < 0
+      signal signal, pid
     end
     nil
   end
 
-  # Sends a *signal* to the processes identified by the given *pids*.
-  def self.signal(signal : Signal, *pids : Int)
-    pids.each do |pid|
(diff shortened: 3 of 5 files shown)


bcardiff:
Please keep `Process#kill` with a `@[Deprecated("Use `#signal` instead")]` annotation.
And keep a minimal spec that uses `#kill`

Blacksmoke16:
I'd also advocate for including something like, `and will be removed in Crystal 0.34.0` so that there is an actual expectation when it will/should be removed.

spec/std/process_spec.cr
```diff
@@ -218,7 +218,7 @@ describe Process do
     process.terminated?.should be_false
 
     # Kill, zombie now
-    process.kill
+    process.signal
```
ysbaddaden: This is a perfect example of why it wasn't a good idea to split this patch from the POSIX abstraction PR, which also introduces .kill and .terminate. Here `.signal` is bad. This PR could also introduce them, but why bother when the other PR was perfect?

jan-zajic: I know, and yes i expected this :). See my comment in #8536 for the proposed way out.. But yet another split to some N other PRs and i probably lost my remaining time and energy to work on that topic..

jan-zajic:
Either way i think it's better that renamed signal method should not have default value, so i removed it. Because of code readability - with process.signal without argument it's very unclear what it does.

jan-zajic:
btw @bcardiff build now fail on some crazy error (in more PRs, not only this): 
> .build/std_spec: relocation error: .build/std_spec: symbol __gmon_start__ version GLIBC_2.2.5 not defined in file libpthread.so.0 with link time reference

bcardiff:
@jan-zajic Yeah, it should be fixed by https://github.com/crystal-lang/crystal/pull/8641

src/process.cr
```diff
@@ -35,7 +35,7 @@ class Process
     LibC.getppid
   end
 
-  # Sends a *signal* to the processes identified by the given *pids*.
+  @[Deprecated("Use #signal instead")]
   def self.kill(signal : Signal, *pids : Int)
```
RX14: The implementation of this should be in terms of `Process.signal`

src/process.cr
```diff
@@ -44,6 +44,15 @@ class Process
     nil
   end
 
+  # Sends a *signal* to the processes identified by the given *pids*.
+  def self.signal(signal : Signal, *pids : Int)
```
RX14: Since we're changing the name anyway: I really don't like that this method takes multiple pids. This behaviour isn't even tested. Should this method just be `def self.signal(signal : Signal, pid : Int) : Nil`?

src/process.cr
straight-shoota: And please keep the documentation comment until the method is entirely removed.

src/process.cr
straight-shoota: Absolutely. There seems to be no reason why receiving multiple pids would be better than iterating externally.

src/process.cr
```diff
@@ -349,9 +358,14 @@ class Process
...
   end
 
-  # See also: `Process.kill`
+  @[Deprecated("Use #signal instead")]
   def kill(sig = Signal::TERM)
-    Process.kill sig, @pid
+    signal sig
+  end
+
+  # See also: `Process.signal`
```
straight-shoota:
```suggestion
  # Sends *signal* to the process.
  #
  # See also: `Process.signal`
```

src/process.cr
```diff
@@ -349,9 +358,14 @@ class Process
...
+  def signal(sig)
```
straight-shoota: API should be compatible to `Process.signal`:

```suggestion
  def signal(signal : Signal)
```

src/process.cr
```diff
@@ -36,14 +36,21 @@ class Process
   end
 
   # Sends a *signal* to the processes identified by the given *pids*.
+  @[Deprecated("Use #signal instead")]
   def self.kill(signal : Signal, *pids : Int)
     pids.each do |pid|
-      ret = LibC.kill(pid, signal.value)
-      raise Errno.new("kill") if ret < 0
+      signal signal, pid
```
bew:
```suggestion
      signal(signal, pid)
```
To avoid confusion

jan-zajic:
@RX14 , @straight-shoota updated according to reviews.

src/process.cr
```diff
@@ -36,14 +36,21 @@ class Process
...
+      signal(signal, pid)
     end
     nil
   end
 
+  # Sends a *signal* to the process identified by the given *pid*.
+  def self.signal(signal : Signal, pid : Int)
+    ret = LibC.kill(pid, signal.value)
+    raise Errno.new("kill") if ret < 0
+    nil
```
RX14: This is fine, but it's *slightly* cleaner to just add an explicit `: Nil` return type to the function. The compiler will automatically ignore the return value of the function in that case, and return `nil`.

jan-zajic: ok, done

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -161,7 +161,7 @@ module Crystal::Playground
         @logger.info "Code execution killed (session=#{@session_key}, filename=#{@running_process_filename})."
         @process = nil
         File.delete @running_process_filename rescue nil
-        process.kill rescue nil
+        process.signal(::Signal::TERM) rescue nil
```
straight-shoota: This one

straight-shoota: This one.

src/process.cr
```diff
@@ -193,7 +199,7 @@ class Process
       $? = process.wait
       value
     rescue ex
-      process.kill
+      process.signal Signal::TERM
```
straight-shoota: And this one.

jan-zajic:
@straight-shoota I'm all for it. But that assumes we will agree on names of that methods. I still don't know any opinion of others. I'am currently for `#terminate` and `#destroy` as shown in new commit. 

The options considered so far are (in order gracefully, force shutdown):
* `#terminate_gracefully`, `#terminate_immediately`
* `#terminate`, `#kill` (impossible to do in this PR where old kill is only deprecated)
* `#terminate`, `#interrupt`
* `#terminate`, `#destroy`

So please vote! @straight-shoota, @RX14, @ysbaddaden, @ j8r, @Sija, @bew, @bcardiff ping. If I forgot someone please let them know.

straight-shoota:
The method to terminate a process is commonly named `kill` in most APIs, which is named after the homonymous unix program.
I think it's best to follow that line and keep the name. The existing overload of `#kill` without arguments can stay, only the behaviour should be changed to send SIGKILL on Posix. This is a breaking change, but not very relevant.

Let's not bother with an alternative method for graceful termination in this PR. Having separate methods for sending signals and terminating a process is sufficient. If you need graceful termination, you can easily send SIGTERM manually. Whether we should have a dedicated method for this should be a separate discussion to not prolong this PR even more.

Sija:
`kill` is a commonly used term as @straight-shoota mentioned, so my vote goes for:

> `#terminate`, `#kill`

j8r:
The `kill` man is: `kill - send a signal to a process`.

I prefer `Process#send_signal`.

RX14:
Please let's not use `kill`.

`signal` should be used for sending signals

`terminate` should terminate the process

`interrupt` will interrupt the process and ask it to shut down

i like the current naming, and i'd prefer for it to stay

Sija:
> Please let's not use `kill`, it's confusing with the existing API.

How come? It sends `KILL` signal, as you would expect from the naming...

straight-shoota:
> it's confusing with the existing API.

I wouldn't worry about that. We need to focus on the best API for the future.
It seems literally *everyone* names this "kill", so I doubt we should settle for something else.

RX14:
No, `kill` is confusingly for sending signals, with the default being `SIGTERM`.

So would you rather have `kill` send `SIGTERM` and `terminate` send `SIGKILL`?

RX14:
Actually this is all too confusing the only interface should be `#signal` as I originally said.

Sija:
@RX14 What's confusing about calling `Process#kill`, which in turn would send `Signal::KILL` signal...? For my taste it can't be clearer than that.

straight-shoota:
> No, kill is confusingly for sending signals, with the default being SIGTERM.

The default should be SIGKILL. Even if it was SIGTERM, that's merely an implementation detail.
`Process#kill` just provides a generic, high level interface to terminate a process. The exact implementation and behaviour is obviously platform specific, but we should try to make it as similar as possible across platforms.

Simply having everyone call `signal(:kill)` is not sufficient. It just mimics the POSIX API and is neither generic nor idiomatic. There is an explicit method to spawn a process (which is already at a very high level of abstraction) and there should also be an explicit method to terminate a process.
(We can certainly make `signal(:kill)` work on windows as well)

RX14:
We can't use the `#kill` name *right now* because `#kill` with no args means `SIGTERM`.

There must be at least one release where `#kill` is *removed*, before we reuse it's name with a different semantics.

RX14:
How about introducing just `#terminate` in this PR which sends `SIGTERM`, then after `#kill` is gone, we can add it back sending `SIGKILL`. So this PR would just have `#signal` and `#terminate`?

jan-zajic:
That sounds reasonable to me.

čt 9. 1. 2020 v 12:34 odesílatel Stephanie Hobbs <notifications@github.com>
napsal:

straight-shoota:
Sure, we can do that. But IMO we don't need a full deprecation cycle just to change the implementation of `Process#kill` from `SIGTERM` to `SIGKILL`. That seems to be an *over*kill. The method still works the same way terminating a process. It only makes a difference when the process actually traps SIGTERM.

However I'm still not sure we actually *want* a `#terminate` method for signalling graceful termination.

spec/std/process_spec.cr
```diff
@@ -228,6 +228,20 @@ describe Process do
...
   end
 
+  it "checks for existence after terminate" do
+    process = Process.new("yes")
+    process.exists?.should be_true
+    process.terminated?.should be_false
+
+    process.terminate
+    process.exists?.should be_true
+    process.terminated?.should be_false
```
RX14: This seems like a race condition to me. If the program gets rescheduled between the `terminate` call and the `exists?` check then this spec will fail. I suspect that running this test in a loop on a loaded machine will already fail.

I'm not sure there's an appropriate way to test this.

RX14:
This is OK for me to merge except for the race condition in the spec suite

straight-shoota:
@jan-zajic Are you still up for this addressing the above comment?

jan-zajic:
@straight-shoota ok i look at it in few days

spec/std/process_spec.cr
oprypin: Was write access given to the branch? Could someone simply remove this spec and merge the PR?

oprypin:
Was write access given to the branch? Could someone simply remove this spec and merge the PR?

oprypin:
Mhm, the merge is non-trivial.

I have merged it and I'm ready to send a new PR.
https://github.com/crystal-lang/crystal/compare/master...oprypin:winkill

I think such a PR, if merged, would show me as the primary author, which is not ideal.

oprypin:
@jan-zajic, could you give some input?
Would you be able to keep pushing this forward?
Do you want to hand over the ownership of this somehow?
Are you OK with someone else re-creating a pull request with your commits?

jan-zajic:
@oprypin unfortunately i don't have time right now to be able to keep pushing this forward..  So i'm ok with your help, as long as you keep my original work (to some point that make sense), so my original effort will not be wasted.. Last time i put a lot of effort into this, but but we got lost in discussions that led nowhere.

jan-zajic:
Closed in favor of #9006

PR #8643, opened 2020-01-02
Fix mutex specs race condition on mt

bcardiff:
In https://circleci.com/gh/crystal-lang/crystal/36001 I experienced a race condition introduced in https://github.com/crystal-lang/crystal/pull/8563 .

After this changes the loop will not only wait for the mutex to be assigned but for the lock to occur.


RX14:
nice catch, weird ci didnt properly trigger here though...

PR #8644, opened 2020-01-02
Fix #8637: Null added if str.sub(Hash) match last char

### LINKED ISSUE #8637: `String#sub(Hash(Char, _))` with str length 1 adds a NUL
```cr
"C".sub({'C' => 'G'}) # => "G\u0000"
"C".sub({'C' => "G"}) # => "G\u0000"
```
https://carc.in/#/r/8b3o

mimame:
Example:
  "helloC".sub({'C' => 'G'}) # => "helloG\u0000"

This is my first pull request to Crystal
I hope to be useful

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1195,6 +1195,7 @@ describe "String" do
     it "subs with char hash" do
       str = "hello"
       str.sub({'e' => 'a', 'l' => 'd'}).should eq("hallo")
+      str.sub({'a' => 'b', 'o' => 'a'}).should eq("hella")
 
       empty = ""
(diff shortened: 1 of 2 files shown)


straight-shoota:
Thank you for this contribution, it's very welcome!

There should also be specs for all the other `sub` overloads not only for char hash. Maybe it would be best to add a new sample to `string_spec.cr` for testing last-character behaviour.

mimame:
Thanks so much @straight-shoota  for the feedback
I've done it in another commit so it's ready to be reviewed and for squash them
Don't hesitate to continue to demand more changes

straight-shoota:
These are actually quite a lot of test cases. But I guess we just have that many overloads for `sub`.

PR #8645, opened 2020-01-03
Fix deprecated warnings in some specs

mimame:
I hope to be useful.

Regards

### DIFF
--- a/spec/std/channel_spec.cr
+++ b/spec/std/channel_spec.cr
@@ -418,7 +418,7 @@ describe "unbuffered" do
 
   it "works with select else" do
     ch1 = Channel(Int32).new
-    Channel.select({ch1.receive_select_action}, true).should eq({1, Channel::NotReady.new})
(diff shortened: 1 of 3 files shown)


asterite:
There was a recent discussion about this. The idea is to test the deprecated methods to see that they still work because people might still be using them. We should only remove them once we remove the associated methods. So I'm not sure we'll merge this PR or similar PRs.

asterite:
All of this to say, deprecation warnings in specs are not a big deal.

mimame:
I see, very interesting discussion.
@asterite feel free to close the pull request

bcardiff:
The motivation is to not have untested code in the std-lib since it might lead to code that does not actually compile.

Sometimes specs needs to be updated to get rid of the deprecation warning. That is a good change. This is the case when the spec uses the method but is not the tested method.

In other cases, the spec should be already present for the replacement alternative and probably the spec can be removed once the deprecation method is effectively removed.

PR #8648, opened 2020-01-04
Close underlying WebSocket connection when connection is closed

### LINKED ISSUE #7199: WebSocket protocol don't close the underlying IO
When called with a network address, `HTTP::WebSocket.new` delegates to `Protocol.new` which creates a `TCPSocket` and connects it to the given address. This socket is stored as an ivar in the `Protocol` instance and is never accessible from outside.

`WebSocket#close` sets a `closed` flag and sends a `close` message, but the underlying IO (`TCPSocket`) is never closed.

This can leak open TCP connections.

On the server-side implementation, `WebSocketHandler` explicitly calls `io.close` after the `WebSocket` session has finished. This call is not executed however, when the session handling raises.

See also #7197

willhbr:
This fixes #7199 by making `WebSocket::Protocol#close` call `TCPSocket#close`. `WebSocket#close` is called when an exception is raised, or when a CLOSE opcode is received.

Also wrapped `ws_session.run` in a begin...ensure to make sure that the `io` is closed on exceptions.

### DIFF
--- a/src/http/server/handlers/websocket_handler.cr
+++ b/src/http/server/handlers/websocket_handler.cr
@@ -38,7 +38,7 @@ class HTTP::WebSocketHandler
       response.headers["Connection"] = "Upgrade"
       response.headers["Sec-WebSocket-Accept"] = accept_code
       response.upgrade do |io|
-        ws_session = WebSocket.new(io)
+        ws_session = WebSocket.new(io, sync_close: false)
         @proc.call(ws_session, context)
         ws_session.run
       ensure
--- a/src/http/server/handlers/websocket_handler.cr
+++ b/src/http/server/handlers/websocket_handler.cr
@@ -41,6 +41,7 @@ class HTTP::WebSocketHandler
         ws_session = WebSocket.new(io)
         @proc.call(ws_session, context)
         ws_session.run
+      ensure
         io.close
       end
     else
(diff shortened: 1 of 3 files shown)


src/http/server/handlers/websocket_handler.cr
```diff
@@ -40,8 +40,11 @@ class HTTP::WebSocketHandler
       response.upgrade do |io|
         ws_session = WebSocket.new(io)
         @proc.call(ws_session, context)
-        ws_session.run
-        io.close
+        begin
```
Sija: `begin ... end` is redundant here, `ensure` block would be just enough.

src/http/web_socket/protocol.cr
```diff
@@ -232,10 +232,13 @@ class HTTP::WebSocket::Protocol
   end
 
   def close(message = nil)
-    if message
-      send(message.to_slice, Opcode::CLOSE)
-    else
-      send(Bytes.empty, Opcode::CLOSE)
+    unless @io.closed?
```
Sija: Prefer early-returns:

```suggestion
    return if @io.closed?
```

RX14:
I think there should be a `sync_close` parameter (see `IO::Stapled` for an example) to make this behaviour configurable.

willhbr:
Thanks for the review, I’m still working on this - just slowly

willhbr:
Sadly I don't think I've got the time to work on this, so I'm going to close this for now. Will re-open if I have time in the future.

PR #8650, opened 2020-01-05
Fix #8649: Future(Nil).get when the future raises

### LINKED ISSUE #8649: Future doesn't raise exceptions if its return type is Nil
```cr
x = future do
  raise "err" if rand < 10
end

p x.get # nil, should raise
```

This happens because this is a `Concurrent::Future(Nil)` and the current check to raise an exception on `get` is `!@value.is_a?(T)`, which for `Nil` will always be true.

I know that a future that returns nothing isn't really the usecase of futures, but I was doing a quick simple job queue (spawning all jobs and waiting for them) and didn't use Channel explicitly. Either way this is a surprising behavior and is easy to fix.

Proposed solution: change the check to validate that `@error` isn't nil and raising, only then checking the value for the correct type. If this is ok, I'll open a PR.

lbguilherme:
See #8649

### DIFF
--- a/spec/std/concurrent/future_spec.cr
+++ b/spec/std/concurrent/future_spec.cr
@@ -46,6 +46,19 @@ describe Concurrent::Future do
       f.completed?.should be_true
     end
 
+    it "computes a value even if the value is nil" do
+      chan = Channel(Nil).new
+
+      f = future { chan.receive }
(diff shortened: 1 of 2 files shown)


lbguilherme:
CI failure (on preview_mt) seems unrelated

PR #8651, opened 2020-01-05
make wait_readable and wait_writable public

stakach:
restoring the changes made in #7366
Change was originally made in this release: https://github.com/crystal-lang/crystal/blob/3aa1a04ca666dcf12a625b58deccbf0f209e1e4d/CHANGELOG.md#0280-2019-04-17

### DIFF
--- a/src/io/evented.cr
+++ b/src/io/evented.cr
@@ -115,11 +115,13 @@ module IO::Evented
     end
   end
 
-  protected def wait_readable(timeout = @read_timeout)
+  # :nodoc:
+  def wait_readable(timeout = @read_timeout)
     wait_readable(timeout: timeout) { |err| raise err }
   end
 
-  protected def wait_readable(timeout = @read_timeout) : Nil
+  # :nodoc:
+  def wait_readable(timeout = @read_timeout) : Nil
     readers = @readers.get { Deque(Fiber).new }
     readers << Fiber.current
     add_read_event(timeout)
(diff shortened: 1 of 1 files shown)


stakach:
Thanks @RX14 
I feel that access to functions like this are what make a language powerful.

Agree that they shouldn't be used day to day but useful when doing something a little more complex that lies outside of the typical.

I'd put this in a similar category to the `uninitialized` keyword

straight-shoota:
@ysbaddaden Is there any particular reason this was reverted in #7505?

ysbaddaden:
Because it's internal API detail, and there aren't any need nor incentive for them to be publicly available, and nobody complained in over a year. Be aware that using this *will* break, most likely when adding support for Windows, or moving from libevent to a custom solution.

I'd love to see a real use-case. If it's just a "feel that it should be accessible", then let's close this PR.

RX14:
@ysbaddaden the real usecase is binding C libraries: https://github.com/spider-gazelle/ssh2.cr/blob/a93eb957420b9e9f31bb8cd46e50ff862c5bc54b/src/session.cr#L39-L49

libgit2 returns EAGAIN from function calls when it needs to read more data, just like `read` and `write`, then expects the caller to handle them like `read` and `write`. So it needs this API.

ysbaddaden:
@RX14 ok.

bcardiff:
The refactors in #7505 is where the changes in #7366 got dropped. (this is maybe a note for future self to write the changelog 🙈 )

PR #8657, opened 2020-01-07
Drop support for comma separators in enums and other cleanups

bcardiff:
* Cleanup code
* Cleanup unneeded primitives
* Remove support for commas inside enums (Follow up from #7618)


straight-shoota:
Shouldn't removal of compiler support for `cast` and usage in stdlib be in separate releases? Otherwise 0.33.0 won't be able to compile 0.32.1.

bcardiff:
0.32.1 has both primitives, this is the required step to keep things moving forward.

The source of 0.33.0 after this PR uses convert and no longer uses :cast. Hence there is no future need to support :cast

RX14:
We really should make PRs/issues for removing things we deprecate and attach them to future milestones ahead of time so we don't forget

bcardiff:
They will always be subject to be outdated when the time comes.

I would use a `next`/`pinned` milestone or even a future version milestone created ahead of time to flag those PRs. Yet they would be an intention, but not a hard commitment to be included in that release.

straight-shoota:
I don't think scheduling this is really important. Whether we remove a deprecated feature in the scheduled release or a few later doesn't make a huge difference. If all, delayed removal gives developers more time to update their code.

Having such collective cleanups from time to time seems perfectly fine to me.

straight-shoota:
Thanks @bcardiff ❤️

PR #8658, opened 2020-01-07
Fix WebSocket IPv6 host parsing

j8r:
Related to https://github.com/crystal-lang/crystal/pull/8066

### DIFF
--- a/spec/std/socket/spec_helper.cr
+++ b/spec/std/socket/spec_helper.cr
@@ -13,6 +13,6 @@ def each_ip_family(&block : Socket::Family, String, String ->)
   end
 
   describe "using IPv6" do
-    block.call Socket::Family::INET6, "::1", "::"
+    block.call Socket::Family::INET6, "[::1]", "::"
   end
 end
--- a/src/http/client.cr
+++ b/src/http/client.cr
@@ -764,7 +764,7 @@ class HTTP::Client
     socket = @socket
     return socket if socket
 
-    hostname = @host.starts_with?('[') && @host.ends_with?(']') ? @host[1..-2] : @host
+    hostname = host.lchop?('[').try &.rchop?(']') || host
     socket = TCPSocket.new hostname, @port, @dns_timeout, @connect_timeout
     socket.read_timeout = @read_timeout if @read_timeout
     socket.write_timeout = @write_timeout if @write_timeout
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -247,6 +247,7 @@ class HTTP::WebSocket::Protocol
     {% end %}
 
     port = port || (tls ? 443 : 80)
+    host = host.lchop?('[').try &.rchop?(']') || host
 
     socket = TCPSocket.new(host, port)
     begin


src/http/client.cr
```diff
@@ -764,7 +764,7 @@ class HTTP::Client
     socket = @socket
     return socket if socket
 
-    hostname = @host.starts_with?('[') && @host.ends_with?(']') ? @host[1..-2] : @host
+    hostname = host.lchop?('[').try &.rchop?(']') || host
```
straight-shoota: This change doesn't optimize. On the contrary, it allocates an additional intermediary string for bracketed hostname.

It also accepts an hostname only starting with a bracket but not ending with it. I suppose this is considered invalid.

src/http/web_socket/protocol.cr
```diff
@@ -247,6 +247,7 @@ class HTTP::WebSocket::Protocol
     {% end %}
 
     port = port || (tls ? 443 : 80)
+    host = host.lchop?('[').try &.rchop?(']') || host
```
straight-shoota: Ditto.

spec/std/socket/spec_helper.cr
```diff
@@ -13,6 +13,6 @@ def each_ip_family(&block : Socket::Family, String, String ->)
   end
 
   describe "using IPv6" do
-    block.call Socket::Family::INET6, "::1", "::"
+    block.call Socket::Family::INET6, "[::1]", "::"
```
straight-shoota: Why change this spec? It's at most unrelated.

src/http/client.cr
j8r: I doesn't accept a hostname only starting with a bracket, because it's `lchop?`.
If not starting with `[`, it will return nil.

Not sure it's less optimized, because there is no `Char::Reader` involved.
I have to benchmark it

straight-shoota: Please don't invest to much time in such tiny optimizations which might have an effect or might not. And yes, allocating a string is likely less efficient than a char reader. The reader could be avoided using `byte_slice`, though.
In any case, it's not related to this PR.

j8r: Now worry, hat's because I already used this logic already in my code to fix this issue

j8r: Now worry, that's because I already used this logic already in my code to fix this issue

spec/std/socket/spec_helper.cr
j8r: By the way, should we chop the brackets or should every socket type constructor should support the IPv6 bracket syntax out of the box (not only just WebSocket)?

src/http/client.cr
j8r: No worry, that's because I already used this logic already in my code to fix this issue

bcardiff:
Should we avoid duplication regarding `URI#hostname`? https://github.com/crystal-lang/crystal/blob/d84b708b583e4dc27b6aa27d6f4c8e4bb997f2ba/src/uri.cr#L185-L187

j8r:
I'm afraid it will add on overhead if it's done for each http/websocket connection.

Or, another alternative is to not support brackets at all, everywhere there is `host`, because that's the correct and official way of representing IPv6 addresses.

Maybe `URI` could have a class method to chop the brackets.

j8r:
Closing, according to the above comment.
I was mistaken to use `[::1]`, I should simply use the standard `::1`

PR #8659, opened 2020-01-07
Add HTTP::Params#to_h

straight-shoota:
`#to_h` provides direct access to the underlying raw_params property. This allows to easily modify parameter data without going through (de-)serialization. `HTTP::Params` doesn't and shouldn't be mutable, but this feature can be useful, for example for proxy applications.

The implementation is identical to #7444 but renamed from `#raw_params` to `#to_h`.

Closes #7444

/cc @Sija

### DIFF
--- a/spec/std/http/params_spec.cr
+++ b/spec/std/http/params_spec.cr
@@ -247,5 +247,19 @@ module HTTP
         Params.new.empty?.should be_true
       end
     end
+
+    describe "#to_h" do
+      it "duplicate values" do
+        Params.parse("foo=bar&foo=baz&baz=qux").to_h.should eq({
+            "foo" => ["bar", "baz"],
+            "baz" => ["qux"],
+        })
+      end
+
+      it "passes as input for .new" do
+        params = Params.parse("foo=bar&foo=baz&baz=qux")
+        Params.new(params.to_h).should eq params
+      end
+    end
   end
 end
--- a/spec/std/http/params_spec.cr
+++ b/spec/std/http/params_spec.cr
@@ -251,8 +251,8 @@ module HTTP
     describe "#to_h" do
       it "duplicate values" do
         Params.parse("foo=bar&foo=baz&baz=qux").to_h.should eq({
-            "foo" => ["bar", "baz"],
-            "baz" => ["qux"],
+          "foo" => ["bar", "baz"],
+          "baz" => ["qux"],
         })
       end
 
--- a/src/http/params.cr
+++ b/src/http/params.cr
@@ -152,7 +152,7 @@ module HTTP
     # params2.to_s # => "foo=bar&foo=baz&qux=zoo"
     # ```
     def to_h : Hash(String, Array(String))
-      @raw_params
+      @raw_params.dup
     end
 
     def ==(other : self)
--- a/src/http/params.cr
+++ b/src/http/params.cr
@@ -152,7 +152,7 @@ module HTTP
     # params2.to_s # => "foo=bar&foo=baz&qux=zoo"
     # ```
     def to_h : Hash(String, Array(String))
-      @raw_params.dup
+      @raw_params.clone
     end
 
     def ==(other : self)
--- a/src/http/params.cr
+++ b/src/http/params.cr
@@ -132,8 +132,6 @@ module HTTP
       form_builder.to_s
     end
 
-    protected getter raw_params
-
     # Returns an empty `HTTP::Params`.
     def initialize
       @raw_params = {} of String => Array(String)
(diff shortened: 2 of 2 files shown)


bew:
The name `to_h` looks to me like it'll create a new hash with the params instead of give me the "internal" field, maybe use another name like `as_h` to convey the intention of having access to the "internal" but as a hash?
Do you see what I mean?

straight-shoota:
I see what you mean, but the fact that `HTTP::Params` uses the same value internally to store the raw params is an implementation detail. We have lot's of `#to_x` methods in stdlib and the meaning is "return the equivalent x for self". Whether x needs to be constructed on demand or is already available is not important. Some methods like `Array#to_a` or `String#to_s` even just return `self`.
`#as_x` methods on the other hand mean "return self interpreted as x". This could potentially also apply here because `HTTP::Params` is essentially just a hash wrapper. But again, that's based on an implementation detail.

However, it might be worth consideration, if the returned hash should be duped to protect the immutability of `HTTP::Params`.

Blacksmoke16:
Why not just make `raw_params` not protected and call it a day, or at at the least keep it internal and define a `#raw` method.  There isn't any conversion going on so I don't see the point in doing `#to_*` or `#as_*`.

```cr
def raw
  @raw_params.dup
end
```

straight-shoota:
Conceptually, `raw_params` isn't a property of `Params` but an internal detail. Instead, `Params` can be considered having a dynamic set of properties, with 1 or more values each. That's provided by its API. So we're not adding a getter. We're adding a conversion method which just happens to return an instance variable. If for whatever reason the internal representation was changed (for example to an Array of KV-pairs) `#to_h` would still be expected to return `Hash(String, Array(String))`.

Surely we could call the method `#raw` or whatever else comes to mind. But it's not very intuitive (what does "raw" mean in this context?) and there is actually a perfectly fitting and commonly known name for such a method: `#to_h`.

There is also `HTTP::Cookies#to_h` which works exactly the same (including the `dup`

Sija:
`crystal tool format && 🚀`

bcardiff:
I would prefer to keep the immutability of http params or at least controlled access to its modification.

What prevents you from clearing all the values for a given key, but leaving that key?

I would expect that:

```
params2 = HTTP::Params.new(params)
```

to work, without going through a `.to_h`. 

But more important the current state of the PR is creating two http params baked by the same hash. And that doesn't seem right to me. Or am I reading something wrong?

straight-shoota:
@bcardiff I'm going to add `dup` which should address your concerns.

bcardiff:
The current state still allows you to create a `HTTP::Params` with an invalid hash and unwanted aliasing can be created.

Some of this issues were already present but this are my proposals.

* `initialize(raw_params : Hash(String, Array(String)))` should either
  * be private
  * be `:nodoc:`
  * validate along the hay the hash is valid & do not alias the params (nor the inner array)
* if duplicating `HTTP::Params` is wanted then add `initialize(other : self)` that will **clone** the underlying hash
  * doing dup will still leave an alias to the Array.
* `def to_h : Hash(String, Array(String))` should **clone**. Again alias to the Array is not wanted
* `set_all(name, values)` should fail if values are empty (Note that `delete(name)` enforce this)
* add `def dup` and `def clone` to the structure that will both **clone** the underlying hash

straight-shoota:
I agree that `.new(raw_params)` could be improved by adding some kind of validation. But I'm not sure whether that's strictly necessary.

Why not have `#clone` instead of `initialize(other : self)`?

bcardiff:
> Why not have #clone instead of initialize(other : self)?

`.clone` is enough. I was trying to offer something similar to the initialize from hash

bcardiff:
It seems that `def initialize(@raw_params : Hash(String, Array(String)))` is used only for `HTTP::Params.parse` specs. It seems we can remove it and use `.to_h` for those specs.

I'm sorry of the whole overhaul.

straight-shoota:
Using `#to_h` *in combination with* `.new(raw_params)` is one of the reasons for this PR, to be able to  indirectly modify.

bcardiff:
The bottom line is that exposing somehow the underlying hash (or array) is wrong. You can't preserve the invariants if you do so.

* Modification to params should be done in controlled way
* Initialization should enforce the invariants

Another example that seems wrong is that `def fetch_all(name)` still gives access to the underlying array.

If the type allows you to change breacking invariants, then is equivalente to use `alias Params = Hash(String,Array(String))`  and having helper methods to use that structure _likely_ as params.

straight-shoota:
> Modification to params should be done in controlled way

How would you suggest to do that? Make `HTTP::Params` mutable?

bcardiff:
Modification to params should be done in controlled way with the existing methods like `[]=`, `set_all`, `delete`, etc.

`to_h` could serve for exposing the values with a hash/array to do filters and manipulations that are not supported by `HTTP::Params` API.

The building a `HTTP::Params` from a hash if wanted need to check the Array is not empty for example (either by raising Argument error or dropping the key with no values)

RX14:
> Make `HTTP::Params` mutable

I'm confused, `#[]=` and `#add` and `#delete` are already there?

bcardiff:
@RX14 yes they are. The class is mutable already.

Sija:
@straight-shoota Why closing?

straight-shoota:
It doesn't seem to be merged.

PR #8662, opened 2020-01-07
Numbers: Use binary primitive for Float#fdiv

bcardiff:
This is a follow up of #8120 

fdiv is a binary primitive since 0.31 so the code in float.cr can go away in favor of the one in primitive.cr

The changes in bin/number.cr look a bit convoluted but is because that file has all the code required to make big and primitive numbers work together.

### DIFF
--- a/src/big/number.cr
+++ b/src/big/number.cr
@@ -4,6 +4,10 @@ struct BigInt
 end
 
 struct BigFloat
+  def fdiv(other : Number::Primitive) : self
+    self.class.new(self / other)
+  end
+
   def /(other : UInt8 | UInt16 | UInt32 | UInt64)
     # Division by 0 in BigFloat is not allowed, there is no BigFloat::Infinitiy
     raise DivisionByZeroError.new if other == 0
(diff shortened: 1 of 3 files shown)


src/primitives.cr
```diff
@@ -400,6 +400,11 @@ end
           def {{op.id}}(other : {{num.id}}) : self
           end
         {% end %}
+
+        # Returns the float division of division `self` and *other*.
```
bew:
```suggestion
        # Returns the float division of `self` and *other*.
```

src/big/number.cr
```diff
@@ -4,6 +4,10 @@ struct BigInt
 end
 
 struct BigFloat
+  def fdiv(other : Number::Primitive) : self
+    self.class.new(self / other)
```
RX14: Why the `self.class.new`?

bcardiff: Because copy-past of abstract code, but I actually like it that way. Same as why annotate with `: self` in the return type.

I would leave it that way.

straight-shoota:
Thanks @bcardiff

PR #8670, opened 2020-01-10
Fix specs on win32

straight-shoota:
This either fixes or disables specs that don't succeed on win32.

With this and #8668 `win32_std_spec.cr` should be all green :+1:

```
Finished in 6.84 seconds
4920 examples, 1 failures, 5 errors, 13 pending
```

### DIFF
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -305,7 +305,7 @@ describe "Dir" do
     it "pattern ending with .." do
       Dir["#{datapath}/dir/.."].sort.should eq [
         datapath("dir", ".."),
-      ]
+      ].sort
     end
 
     it "pattern ending with */.." do
@@ -313,13 +313,13 @@ describe "Dir" do
         datapath("dir", "dots", ".."),
         datapath("dir", "subdir", ".."),
         datapath("dir", "subdir2", ".."),
-      ]
+      ].sort
     end
 
     it "pattern ending with ." do
       Dir["#{datapath}/dir/."].sort.should eq [
         datapath("dir", "."),
-      ]
+      ].sort
     end
 
     it "pattern ending with */." do
@@ -327,7 +327,7 @@ describe "Dir" do
         datapath("dir", "dots", "."),
         datapath("dir", "subdir", "."),
         datapath("dir", "subdir2", "."),
-      ]
+      ].sort
     end
 
     context "match_hidden: true" do
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -266,8 +266,7 @@ describe "Dir" do
       end
     end
 
-    # TODO: This spec is broken on win32 because of `raise` weirdness on windows
-    pending_win32 "tests with relative path starting recursive" do
+    it "tests with relative path starting recursive" do
       Dir["**/dir/*/"].sort.should eq [
         datapath("dir", "dots", ""),
         datapath("dir", "subdir", ""),
@@ -298,8 +297,12 @@ describe "Dir" do
       Dir[""].should eq [] of String
     end
 
-    pending_win32 "root pattern" do
-      Dir["/"].should eq ["/"]
+    it "root pattern" do
+      {% if flag?(:windows) %}
+        Dir["C:/"].should eq ["C:\\"]
+      {% else %}
+        Dir["/"].should eq ["/"]
+      {% end %}
     end
 
     it "pattern ending with .." do
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -274,7 +274,7 @@ describe "Dir" do
       ].sort
     end
 
-    it "matches symlinks" do
+    pending_win32 "matches symlinks" do
       link = datapath("f1_link.txt")
       non_link = datapath("non_link.txt")
 
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -257,11 +257,11 @@ describe "Dir" do
 
     it "tests with relative path (starts with ..)" do
       Dir.cd(datapath) do
-        base_path = "../data/dir"
+        base_path = Path["..", "data", "dir"]
         Dir["#{base_path}/*/"].sort.should eq [
-          File.join(base_path, "dots", ""),
-          File.join(base_path, "subdir", ""),
-          File.join(base_path, "subdir2", ""),
+          base_path.join("dots", "").to_s,
+          base_path.join("subdir", "").to_s,
+          base_path.join("subdir2", "").to_s,
         ].sort
       end
     end
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -1,6 +1,30 @@
 require "./spec_helper"
 require "../support/errno"
 
+private def unset_tempdir
+  {% if flag?(:windows) %}
+    old_tempdirs = {ENV["TMP"]?, ENV["TEMP"]?, ENV["USERPROFILE"]?}
+    begin
+      ENV.delete("TMP")
+      ENV.delete("TEMP")
+      ENV.delete("USERPROFILE")
+
+      yield
+    ensure
+      ENV["TMP"], ENV["TEMP"], ENV["USERPROFILE"] = old_tempdirs
+    end
+  {% else %}
+    begin
+      old_tempdir = ENV["TMPDIR"]?
+      ENV.delete("TMPDIR")
+
+      yield
+    ensure
+      ENV["TMPDIR"] = old_tempdir
+    end
+  {% end %}
+end
+
 private def it_raises_on_null_byte(operation, &block)
   it "errors on #{operation}" do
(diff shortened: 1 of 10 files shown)


spec/std/dir_spec.cr
```diff
@@ -298,8 +297,12 @@ describe "Dir" do
       Dir[""].should eq [] of String
     end
 
-    pending_win32 "root pattern" do
-      Dir["/"].should eq ["/"]
+    it "root pattern" do
+      {% if flag?(:windows) %}
+        Dir["C:/"].should eq ["C:\\"]
+      {% else %}
+        Dir["/"].should eq ["/"]
```
RX14: Why is it equal to a string array here??

spec/std/dir_spec.cr
```diff
@@ -274,7 +274,7 @@ describe "Dir" do
       ].sort
     end
 
-    it "matches symlinks" do
+    pending_win32 "matches symlinks" do
```
RX14: Why/how does this fail and can it be fixed?

spec/std/dir_spec.cr
```diff
@@ -364,7 +364,7 @@ describe "Dir" do
     end
 
     it "raises" do
-      expect_raises_errno(Errno::ENOENT, "Error while changing directory to '/nope'") do
+      expect_raises_errno(Errno::ENOENT, {{ flag?(:win32) ? /SetCurrentDirectory: .* No such file or directory/ : "Error while changing directory to '/nope'" }}) do
```
RX14: Do we have to spec the error messages? I'd rather just not

src/dir/glob.cr
```diff
@@ -161,7 +161,8 @@ class Dir
           path_stack << {next_pos, root, nil}
         when DirectoriesOnly
           raise "unreachable" unless path
-          fullpath = path == File::SEPARATOR_STRING ? path : path + File::SEPARATOR
+          # FIXME: [win32] File::SEPARATOR_STRING comparison is not sufficient for Windows paths.
+          fullpath = path == File::SEPARATOR_STRING ? path : Path[path].join("").to_s
```
RX14: Can we expand this ternary into an `if`? I think it's a little ugly.

spec/std/float_printer_spec.cr
```diff
@@ -106,7 +106,7 @@ describe "#print Float64" do
     test_pair 5.5626846462680035e-309, "5.562684646268003e-309"
   end
 
-  it "falure case" do
+  pending_win32 "failure case" do
```
RX14: Why is this broken?

spec/std/raise_spec.cr
```diff
@@ -1,16 +1,16 @@
-require "spec"
+require "./spec_helper"
 
 describe "raise" do
   callstack_on_rescue = nil
 
-  it "should set exception's callstack" do
+  pending_win32 "should set exception's callstack" do
```
RX14: We don't have callstacks on windows right?

spec/std/dir_spec.cr
straight-shoota: What else would you expect?  Path array?

spec/std/dir_spec.cr
straight-shoota: IMO it's better in order to avoid false positives where the reason for failure might be something entirely different.

spec/std/float_printer_spec.cr
straight-shoota: I have no idea. It returns a different result. Maybe that's expected? IDK. Needs investigation into the win32 libs.

spec/std/raise_spec.cr
straight-shoota: Not yet, but we should eventually, I suppose?

spec/std/dir_spec.cr
RX14: Oh, sorry, I was thinking of `Path[]`. I'm clearly tired hehe

spec/std/dir_spec.cr
straight-shoota: The first `File.symlink` fails with `WinError 1314 A required privilege is not held by the client`.

It seems [`CreateSymbolicLinkW`](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw) requires admin rights.

RX14: Can we design a spec helper which detects admin rights and marks the spec pending depending?

straight-shoota: Sure. But this can be postponed. It's marked as pending, so it naturally means we need to come back to this.

RX14: Well judging by the number of pending specs which have been hanging around for years...

spec/std/dir_spec.cr
```diff
@@ -355,19 +382,19 @@ describe "Dir" do
...
     it "returns default directory for tempfiles" do
-      old_tmpdir = ENV["TMPDIR"]?
-      ENV.delete("TMPDIR")
-      Dir.tempdir.should eq("/tmp")
-    ensure
-      ENV["TMPDIR"] = old_tmpdir
+      unset_tempdir do
+        # FIXME: [win32] GetTempPathW returns the Windows directory, which is not
+        # necessarily `C:\\WINDOWS`.
+        Dir.tempdir.should eq {{ flag?(:windows) ? "C:\\WINDOWS" : "/tmp" }}
```
oprypin: But what if it's `C:\windows`? :grimacing: 

I'm eager to get this fixed because for me it's [the last failing spec](https://github.com/oprypin/crystal/runs/384714527#step:15:45).

If there's no easy solution, we could just put `pending_win32` on this one.

straight-shoota: We could simply call `upcase`, windows paths are usually case-insensitive :laughing:

oprypin: Oh btw, this spec is actually failing on Mac in this state (see circleci)

straight-shoota: Technically, we could verify the value using `GetWindowsDirectoryW`. But I don't want to go that deep into the implementation details of the OS API.

It's probably best to simply verify that the directory exists which should be the case for Windows directory.

src/dir/glob.cr
```diff
@@ -168,7 +168,7 @@ class Dir
             fullpath = Path[path].join("").to_s
           end
 
-          if (dir_entry && dir_entry.dir?) || dir?(fullpath)
+          if (dir_entry && dir_entry.dir?) || (!dir_entry && dir?(fullpath))
```
RX14: Why not just revert to the old code? It was actually simpler.

spec/generate_windows_spec.sh
```diff
@@ -65,7 +65,7 @@ echo
 for spec in $(find "spec/$SPEC_SUITE" -type f -iname "*_spec.cr" | sort); do
   require="require \"./${spec##spec/}\""
 
-  if ! linker_command=$($CRYSTAL_BIN build --cross-compile --target x86_64--windows-msvc "$spec" 2>/dev/null); then
+  if ! linker_command=$($CRYSTAL_BIN build --cross-compile --target x86_64--windows-msvc -Dwithout_zlib -Dwithout_openssl "$spec" 2>/dev/null); then
```
oprypin: I don't think this change should be in this PR?
And in general, why should the list of specs be generated in an incomplete state?  These flags aren't needed for the current specs, anyway.

straight-shoota: Oops, that commit belongs to a different branch.

spec/std/path_spec.cr
```diff
@@ -631,8 +631,7 @@ describe Path do
     end
 
     it "doesn't expand ~" do
-      ["~", "~/foo"].each do |path|
-        path = Path[path]
+      [Path["~"], Path["~", "foo"]].each do |path|
```
Sija: I'd use `Tuple` here.

```suggestion
      {Path["~"], Path["~", "foo"]}.each do |path|
```

straight-shoota:
Rebased to squash fixups.

spec/std/float_printer_spec.cr
straight-shoota: It seems there is something fundamentally broken with floating point numbers on win32:
```cr
"%f" % 1.0 # => "0.000000"
```

src/dir/glob.cr
```diff
@@ -161,11 +161,15 @@ class Dir
...
+          # FIXME: [win32] File::SEPARATOR_STRING comparison is not sufficient for Windows paths.
+          if path == File::SEPARATOR_STRING
+            fullpath = path
           else
-            yield fullpath if dir?(fullpath)
+            fullpath = Path[path].join("").to_s
+          end
+
+          if (dir_entry && dir_entry.dir?) || dir?(fullpath)
+            yield fullpath
```
RX14: You managed to drop your fixup changes.

Never attempt to fixup another fixup commit, it breaks. Always re-fixup the original one

straight-shoota: Na, I just messed up the rebase. Did only pay atttention to the first change.

PR #8673, opened 2020-01-11
Fix colorize specs on dumb terminal

oprypin:
The wrapper that unconditionally applies colorize was not correctly used in all places.

### DIFF
--- a/spec/std/colorize_spec.cr
+++ b/spec/std/colorize_spec.cr
@@ -12,7 +12,7 @@ end
 private class ColorizeToS
   def to_s(io)
     io << "hello"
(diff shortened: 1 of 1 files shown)


oprypin:
See failures before this: https://github.com/oprypin/crystal/runs/384687812#step:15:45

spec/std/colorize_spec.cr
```diff
@@ -12,7 +12,7 @@ end
 private class ColorizeToS
   def to_s(io)
     io << "hello"
-    io << "world".colorize.blue
+    io << ::colorize("world").toggle(true).blue
```
RX14: this `.toggle(true)` isn't required.

PR #8674, opened 2020-01-11
Fix multi_writer specs to close file before reading/deleting it

oprypin:


oprypin:
See failures before this: https://github.com/oprypin/crystal/runs/384687812#step:15:66

PR #8675, opened 2020-01-11
Union::OrderedConverter

KimBurgess:
Provides a generalised converter for reading JSON values into union types based on an arbitrary type order.

### DIFF
--- a/src/json/from_json.cr
+++ b/src/json/from_json.cr
@@ -324,3 +324,43 @@ module String::RawConverter
     value.read_raw
   end
 end
+
+module Union::OrderedConverter(*T)
+  def self.from_json(pull : JSON::PullParser) : Union(*T)
+    location = pull.location
+    string = nil
+
+    # Optimised path
+    {% if T.all? do |type|
+            type == Nil ||
+              type == Bool ||
+              type == Int8 || type == Int16 || type == Int32 || type == Int64 ||
+              type == UInt8 || type == UInt16 || type == UInt32 || type == UInt64 ||
+              type == Float32 || type == Float64 ||
+              type == String
+          end %}
+      {% for type in T %}
+        {% if type == Nil %}
+          return pull.read_null if pull.kind.null?
+        {% else %}
+          value = pull.read?({{type}})
+          return value unless value.nil?
+        {% end %}
+      {% end %}
+      string = pull.read_raw
+
+    # Fallback
+    {% else %}
+      string = pull.read_raw
+      {% for type in T %}
+        begin
+          return {{type}}.from_json(string)
+        rescue JSON::ParseException
+          # Ignore
+        end
+      {% end %}
+    {% end %}
+
+    raise JSON::ParseException.new("Couldn't parse #{Union(*T)} from #{string}", *location)
+  end
+end
--- a/src/json/to_json.cr
+++ b/src/json/to_json.cr
@@ -286,3 +286,31 @@ module String::RawConverter
     json.raw(value)
   end
 end
+
+# Converter to be used with `JSON.mapping` and `YAML.mapping` to read JSON
+# values into a `Union` type with a non-standard type order.
+#
+# Default conversion prioritises component types based on an alphabetical order
+# due to type system behaviour. This may be used to to specify an alternative
+# order as required.
+#
+# ```
+# require "json"
+#
+# class Example
+#   JSON.mapping({
+#     value:           {type: Int64 | Float64},
+#     converted_value: {type: Int64 | Float64, converter: Union::OrderedConverter(Int64, Float64)},
+#   })
+# end
+#
+# example = Example.from_json(%({"value": 1, "converted_value": 1}))
+# example.value           # => "1.0"
+# example.converted_value # => "1"
+# example.to_json         # => %({"value": 1.0, "converted_value": 1})
+# ```
+module Union::OrderedConverter(*T)
+  def self.to_json(value, json : JSON::Builder)
+    value.to_json(json)
+  end
+end
--- a/spec/std/json/mapping_spec.cr
+++ b/spec/std/json/mapping_spec.cr
@@ -127,6 +127,18 @@ private class JSONWithTimeEpochMillis
   })
 end
 
+private class JSONWithOrderedUnion
+  JSON.mapping({
+    value: {type: Int64 | Float64, converter: Union::OrderedConverter(Int64, Float64)},
+  })
+end
+
+private class JSONWithComplexOrderedUnion
+  JSON.mapping({
+    value: {type: Int64 | Float64 | Array(Int32), converter: Union::OrderedConverter(Int64, Float64, Array(Int32))},
+  })
+end
+
 private class JSONWithRaw
   JSON.mapping({
     value: {type: String, converter: String::RawConverter},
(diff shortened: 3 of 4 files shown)


src/json/to_json.cr
```diff
@@ -286,3 +286,31 @@ module String::RawConverter
...
+# end
+#
+# example = Example.from_json(%({"value": 1, "converted_value": 1}))
+# example.value           # => "1.0"
+# example.converted_value # => "1"
+# example.to_json         # => %({"value": 1.0, "converted_value": 1})
+# ```
+module Union::OrderedConverter(*T)
+  def self.to_json(value : Union, json : JSON::Builder)
+    value.to_json(json4)
```
bew: json4 ? :smiley:

src/json/from_json.cr
```diff
@@ -324,3 +324,31 @@ module String::RawConverter
     value.read_raw
   end
 end
+
+module Union::OrderedConverter(*T)
+  def self.from_json(pull : JSON::PullParser) : Union(*T)
+    location = pull.location
+    string = pull.raw_value
```
RX14: `pull.read_raw` is correct here, `pull.raw_value` won't work for arrays etc.

src/json/to_json.cr
```diff
@@ -286,3 +286,31 @@ module String::RawConverter
...
+# due to type system behaviour. This may be used to to specify an alternative
+# order as required.
+#
+# ```
+# require "json"
+#
+# class Example
+#   JSON.mapping({
+#     value:           {type: Int64 | FLoat64},
+#     converted_value: {type: Int64 | FLoat64, converter: Union::OrderedConvert(Int64, Float64)},
```
bew: `FLoat64` > `Float64` on the above 2 lines

src/json/to_json.cr
KimBurgess: As above. Evidentially this needs specs :)

src/json/from_json.cr
```diff
@@ -324,3 +324,31 @@ module String::RawConverter
...
+      {% elsif type == Bool ||
+                 type == Int8 || type == Int16 || type == Int32 || type == Int64 ||
+                 type == UInt8 || type == UInt16 || type == UInt32 || type == UInt64 ||
+                 type == Float32 || type == Float64 ||
+                 type == String %}
+        value = pull.read?({{type}})
+        return value unless value.nil?
+      {% else %}
+        begin
+          value = {{type}}.from_json(string)
```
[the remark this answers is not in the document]
KimBurgess: `#read_raw` will progress the parser to the next token, but it should be possible to restructure this a little so that it takes the optimised route until the first non-primitive is found then falls back the `.from_json` attempts and exception black hole.

KimBurgess:
That should be a little neater now.

@RX14 I ran into some difficulties getting it to take the optimised route for unions containing a mix of primitive and non-primitive values without introducing a lot of additional runtime checks. To bypass that and also make the implementation a little clearer for maintenance it will now expand to either extraction via `#read?` for unions made entirely of primitives, or calling `#.to_json` for each type otherwise.

straight-shoota:
I'm not sure if this should be in stdlib. Depends on how much value this provides for generic use cases.

Could you please show some examples where this would be useful?

KimBurgess:
The original motivation for this was conversion of numeric union types. Although the JSON spec does not differentiate between integer and floating point values it is extremely common to see this encoding in the wild (i.e. `1.to_json #=> "1"`, `1.0.to_json #=> "1.0"`), including from crystal's std lib. Due to the default lexical order a union that accepts both `Float` and `Int` types will always decode both of these to the floating point representation.
```crystal
require "json"

a : JSON::Any::Type = 1i64
a.is_a? Int
#=> true

b = JSON::Any::Type.from_json a.to_json
b.is_a? Int
#=> false
```
It's worth noting that this only applies when using `Union.from_json`. `JSON::Parser` / `JSON::Lexer` already provides [discrete kinds for int and float tokens and preferences int](https://github.com/crystal-lang/crystal/blob/41bd18fbea4aec50aad33aa3beb7a0bf30544186/src/json/parser.cr#L20-L23).
```crystal
require "json"

a : JSON::Any::Type = 1i64
a.is_a? Int
#=> true

c = JSON.parse a.to_json
c.raw.is_a? Int
#=> true
```

For an example outside of primitives, take a service that returns string based response codes. These may fall into a set of subtypes such that parsing order is meaningful.
```crystal
require "json"

module DeviceResponse
  enum Success
    SuccessA
    SuccessB
    SuccessC
    # ...
  end

  enum Error
    ErrorA
    ErrorB
    ErrorC
    # ...
  end

  # Unhandled response types
  alias Other = String

  alias Type = Success | Error | Other
  Converter = Union::OrderedConverter(Success, Error, Other)
end

struct Response
  include JSON::Serializable

  @[JSON::Field(converter: DeviceResponse::Converter)]
  getter code : DeviceResponse::Type
end

struct ResponseWithoutConverter
  include JSON::Serializable

  @[JSON::Field]
  getter code : DeviceResponse::Type
end

response = Response.from_json %({"code": "SuccessA"})
response.code.is_a? DeviceResponse::Success
#=> true

response = ResponseWithoutConverter.from_json %({"code": "SuccessA"})
response.code.is_a? DeviceResponse::Success
#=> false
```

straight-shoota:
The issue I see with this is that - apart from primitives - the generic ordered converter performs a naive trial and error approach which is relatively inefficient.
A custom converter which can already tell from the value which result type will fit would be much more performant than raising an exception on every failed conversion.
This is a much better converter than what the default on can provide:
```cr
module ResponseCodeConverter
  def self.from_json(pull : JSON::PullParser)
    value = pull.read?(String)
    DeviceResponse::Success.parse?(value) || DeviceResponse::Error.parse?(value) || value
  end
end
```

We could maybe also change the behaviour of the union parser to prefer parsing more specific types, so that when parsing a `Union(Float64, Int32)`, it returns `Int32` when it parses as int.

KimBurgess:
Good point regarding performance in the above example as opposed to a custom converter.

The trial and error method is [used by `Union.new`](https://github.com/crystal-lang/crystal/blob/41bd18fbea4aec50aad33aa3beb7a0bf30544186/src/json/from_json.cr#L256-L263). There would be no impact as opposed to not specifying a converter when used with a union composed entirely of non-primitives. Similarly one composed entirely of primitives would use the optimal path. The mixed case would take a hit. Regardless, the existence of this in std lib could arguably steer others away from writing a more optimised converter for their use case.

With that in mind, perhaps the best path here is to:
1. Split this out into a standalone shard. This would enable limitations to be specified in docs, and force some cognizance around its use.
2. Change the behaviour of `Union.new` to preference numerics in the same order as `JSON::Lexer`. This will ensure that the behaviour of both parsers in stb lib is consistent and remove the need for specifying a converter in the first use case presented above.

@straight-shoota / @RX14 let me know if this works for you and I'll close this off, and resubmit a new PR in the next few days. This will introduce a breaking change, but discussion on how to best manage that could be discussed on the new PR.

KimBurgess:
Shard for standalone use here: https://github.com/aca-labs/ordered-union

straight-shoota:
As a general consideration, I don't think such micro shards are a good idea. It only provides a single small feature in a couple of LOCs. The little value such a shard provides does IMO not justify the overhead of maintaining an additional dependency. I'd personally never use such a shard as a dependency.

In my opinion, vendoring small code snippets directly into a repository is a much better option for this. Especially considering that it probably won't receive much changes anyway in the future.

Another option is obviously to provide such a feature in a larger shard together with other extensions to the serialization framework.

But standalone shards providing only such isolated features don't seem to be worth the trouble, for neither maintainers nor potential users. I also believe this is not healthy for the shards ecosystem when it gets cluttered with mirco dependencies. That's one of the main headaches I get when I have to install an npm-based project: Even smallest applications have easily dozens of dependencies which each providing a small feature. That's just many points of failure.

PR #8679, opened 2020-01-11
Print built sample command in CI.

j8r:
Makes also xargs compatible with Busybox, used in Alpine.

Purpose: little PR to know failed samples in CI.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -97,7 +97,7 @@ build() {
       ;;
   esac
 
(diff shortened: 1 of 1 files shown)


bcardiff:
Can you show the output when something goes wrong?

j8r:
If a sample fails to compile, the error will be printed but not necessarily the sample file (example: https://circleci.com/gh/crystal-lang/crystal/36685)

straight-shoota:
Thanks @j8r

PR #8686, opened 2020-01-13
JSON: deserialize int before float in union types

### LINKED ISSUE #7333: Array.from_json fails to parse int and float in the same object
If the JSON string includes an Array with both Int and Float, the parser converts all the numbers to Float.
```Crystal
require "json"
str = "[1, 2.0, \"a\"]"
p Array(Int32 | Float64 | String).from_json(str)

>> Output: [1.0, 2.0, "a"]
```
##
Side note: The from_yaml methods works fine
```Crystal
require "yaml"
str = "[1, 2.0, \"a\"]"
p Array(Int32 | Float64 | String).from_yaml(str)

>> Output: [1, 2.0, "a"]

### PR DESCRIPTION
**Pull request by asterite**
Fixes #7333

Removed the need of #8675

### DIFF
--- a/src/json/from_json.cr
+++ b/src/json/from_json.cr
@@ -231,18 +231,29 @@ def Union.new(pull : JSON::PullParser)
     # Here we store types that are not primitive types
     {% non_primitives = [] of Nil %}
 
+    # Deserialize Nil, Bool, Int and String. Leave Floats last because
+    # an integer literal should be deserialized as an int, not as a float.
     {% for type, index in T %}
       {% if type == Nil %}
         return pull.read_null if pull.kind.null?
       {% elsif type == Bool ||
                  type == Int8 || type == Int16 || type == Int32 || type == Int64 ||
                  type == UInt8 || type == UInt16 || type == UInt32 || type == UInt64 ||
-                 type == Float32 || type == Float64 ||
                  type == String %}
         value = pull.read?({{type}})
         return value unless value.nil?
+      {% elsif type == Float32 || type == Float64 %}
+        # Skip for now
       {% else %}
         {% non_primitives << type %}
+      {% end %}
+    {% end %}
+
+    # Now see if we have any Float
+    {% for type, index in T %}
+      {% if type == Float32 || type == Float64 %}
+        value = pull.read?({{type}})
+        return value unless value.nil?
       {% end %}
     {% end %}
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/json/from_json.cr [hunk 1]
```diff
@@ -231,21 +231,32 @@ def Union.new(pull : JSON::PullParser)
...
       {% if type == Nil %}
         return pull.read_null if pull.kind.null?
       {% elsif type == Bool ||
                  type == Int8 || type == Int16 || type == Int32 || type == Int64 ||
                  type == UInt8 || type == UInt16 || type == UInt32 || type == UInt64 ||
-                 type == Float32 || type == Float64 ||
                  type == String %}
         value = pull.read?({{type}})
         return value unless value.nil?
+      {% elsif type == Float32 || type == Float64 %}
```
wontruefree: What about just moving this conditional down where the else statement is?
There would be one less branching case but it might be more confusing to understand.

```suggestion
      {% elsif type != Float32 || type != Float64 %}
        {% non_primitives << type %}
      {% end %}
```

asterite:
@bcardiff I think that's unrelated to this. That seems to be a bug:

```crystal
require "json"
str = "#{Int32::MAX.to_i64 + 2}"
p str
p Int32.from_json(str)
```

I think that should raise an exception.

And once we fix that, parsing a union of Int32 and Int64 will work correctly.

That said, I have no idea what's the use case for all of this. Maybe we should advise more to keep things simple. In which program do you need a union of int and float, and which API is producing such abomination?

bcardiff:
Ok, maybe the `Int32.parse?` should complain and that should avoid the issue. But if the number is in range of Int32, what prevents it to be returned as an Int64? Is there something enforcing that the smallest type is been used? I thought it was more about a type priority.

I don't expect unions between ints will appear in json mapping, but I do expect int | float. So yeah, we can scope to that.

asterite:
> But if the number is in range of Int32, what prevents it to be returned as an Int64? Is there something enforcing that the smallest type is been used?

I don't think there's a reasonable use case for that because `1` in JSON is just an integer, and a user wanting to parse JSON as Int32 or Int64, whichever, makes no sense. One would choose just one integer type.

For that reason, a JSON Api that returns `[1, 2.3]` makes very little sense in my mind. Maybe they are all floats and should be parsed as such, or maybe the array always has two elements, one int and one float, in which case one would use a tuple. But since JSON doesn't distinguish between int and float (everything is just a number), it just doesn't make sense to parse something as Int32 or Float64.

I just sent this PR because I was bored and this was something small I could tackle...

asterite:
And by that I mean, maybe we should close this PR and close the original issue as invalid.

PR #8689, opened 2020-01-14
JSON: improved deserialization into union types

### LINKED ISSUE #7333: Array.from_json fails to parse int and float in the same object

KimBurgess:
Modified behaviour when reading into union types to prioritize:
1. native types used by `JSON::PullParser`
2. casts to numeric types of the same kind in order of precision
3. casts to numeric types of a different kind
4. non-primitive types

Resolves #7333 
Supersedes #8675
Extends  #8686

### DIFF
--- a/src/json/from_json.cr
+++ b/src/json/from_json.cr
@@ -226,42 +226,55 @@ end
 def Union.new(pull : JSON::PullParser)
   location = pull.location
 
-  # Optimization: use fast path for primitive types
   {% begin %}
-    # Here we store types that are not primitive types
-    {% non_primitives = [] of Nil %}
-
-    {% for type, index in T %}
-      {% if type == Nil %}
-        return pull.read_null if pull.kind.null?
-      {% elsif type == Bool ||
-                 type == Int8 || type == Int16 || type == Int32 || type == Int64 ||
-                 type == UInt8 || type == UInt16 || type == UInt32 || type == UInt64 ||
-                 type == Float32 || type == Float64 ||
-                 type == String %}
-        value = pull.read?({{type}})
-        return value unless value.nil?
-      {% else %}
-        {% non_primitives << type %}
-      {% end %}
+    case pull.kind
+    {% if T.includes? Nil %}
+    when .null?
+      return pull.read_null
     {% end %}
+    {% if T.includes? Bool %}
+    when .bool?
+      return pull.read_bool
+    {% end %}
+    {% if T.includes? String %}
+    when .string?
+      return pull.read_string
+    {% end %}
+    when .int?
+    {% type_order = [Int64, UInt64, Int32, UInt32, Int16, UInt16, Int8, UInt8, Float64, Float32] %}
+    {% for type in type_order.select { |t| T.includes? t } %}
+      value = pull.read?({{type}})
+      return value unless value.nil?
+    {% end %}
+    when .float?
+    {% type_order = [Float64, Float32] %}
+    {% for type in type_order.select { |t| T.includes? t } %}
+      value = pull.read?({{type}})
+      return value unless value.nil?
+    {% end %}
+    end
+  {% end %}
+
+  {% begin %}
+    {% primitive_types = [Nil, Bool, String] + Number::Primitive.union_types %}
+    {% non_primitives = T.reject { |t| primitive_types.includes? t } %}
 
     # If after traversing all the types we are left with just one
     # non-primitive type, we can parse it directly (no need to use `read_raw`)
     {% if non_primitives.size == 1 %}
       return {{non_primitives[0]}}.new(pull)
+    {% else %}
+      string = pull.read_raw
+      {% for type in non_primitives %}
+        begin
+          return {{type}}.from_json(string)
+        rescue JSON::ParseException
+          # Ignore
+        end
+      {% end %}
+      raise JSON::ParseException.new("Couldn't parse #{self} from #{string}", *location)
     {% end %}
   {% end %}
-
-  string = pull.read_raw
-  {% for type in T %}
-    begin
-      return {{type}}.from_json(string)
-    rescue JSON::ParseException
-      # Ignore
-    end
-  {% end %}
-  raise JSON::ParseException.new("Couldn't parse #{self} from #{string}", *location)
 end
 
 # Reads a string from JSON parser as a time formated according to [RFC 3339](https://tools.ietf.org/html/rfc3339)
--- a/spec/std/json/serialization_spec.cr
+++ b/spec/std/json/serialization_spec.cr
@@ -206,6 +206,22 @@ describe "JSON serialization" do
       Union(Float64, Array(Int32)).from_json(%(1.23)).should eq(1.23)
     end
 
+    it "deserializes union of Int32 and Float64 (#7333)" do
+      value = Union(Int32, Float64).from_json("1")
+      value.should be_a(Int32)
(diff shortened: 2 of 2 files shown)


src/json/from_json.cr
```diff
@@ -226,31 +226,33 @@ end
...
+    {% if T.includes? Bool %}
+    when .bool?
+      return pull.read_bool
+    {% end %}
+    {% if T.includes? String %}
+    when .string?
+      return pull.read_string
+    {% end %}
+    when .int?
+    {% type_order = [Int64, Int32, Int16, Int8, UInt64, UInt32, UInt16, UInt8, Float64, Float32] %}
```
RX14: I'm interested in other's thoughts on ordering `Int`s and `UInt`s here.

src/json/from_json.cr
```diff
@@ -226,31 +226,33 @@ end
...
-        {% non_primitives << type %}
-      {% end %}
+    case pull.kind
+    {% if T.includes? Nil %}
+    when .null?
+      return pull.read_null
     {% end %}
-
-    # If after traversing all the types we are left with just one
-    # non-primitive type, we can parse it directly (no need to use `read_raw`)
```
RX14: We've lost this optimization :(

src/json/from_json.cr
KimBurgess: Same.

src/json/from_json.cr
KimBurgess: Restored.

src/json/from_json.cr
asterite: I don't think it matters. Maybe it should be from smaller to bigger. For example if you have `Union(Int8, Int16).from_json(1)`, what would you expect? `1` fits in both types.

That's why I'm not sure whether we should do anything about it other than putting floats to the end.

asterite: That is, my PR. But whenever I send a PR there are always thoughts, comments, etc. It's so slow to develop in Crystal. Meanwhile in Ruby they push commits without much thoughts, and it's a success.

KimBurgess: > I don't think it matters.

`JSON::PullParser#read?` silently wraps in the case of an overflow so order is important. Combine that with the alphabetic type order and your end up with `Int16` > `Int32` > `Int64` > `Int8` which would likely lead to some interesting behaviour _if_ someone was to parse into a union of indiscriminate types. That was the main motivation for this following your PR.

```crystal
alias T = Int16 | Int32

a : T = Int16::MAX.to_i + 2
# => 32769

b : T = T.from_json(a.to_json)
# => -32767
```

An initial implementation prioritised smaller types, converted with `{{type}}.new` and caught Overflow errors along the way, but this seemed like a lot of runtime overhead when the type could be satisfied without it.

There is likely a separate discussion needed on if that silent cast is the correct behaviour from `#read?`, but in the interests of keeping this focused on `Union` I've not touched that here.

KimBurgess: > But whenever I send a PR there are always thoughts, comments, etc

FWIW this is one of the things I enjoy most about crystal. The discussion, reasoning and associated rigour this brings is what is helping to shape it into such a beautiful language. There definitely a balance to be had (no wants a design-by-commitee-everything-has-an-rfc-and-18-month-turn-around style project), but you all seem to nail that pretty well at the moment :)

RX14: `#read` needs to be fixed to now overflow. I think ordering by descending `::MAX` is my preferred semantics.

KimBurgess: Opened (link in timeline below).

Re: ordering, `PullParser` [reads directly into a Int64](https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/json/pull_parser.cr#L23) so it seems returning this initially makes the most sense. How about `[Int64, UInt64, Int32, UInt32, Int16, UInt16, Int8, UInt8, Float64, Float32]`?

RX14: @KimBurgess might as well, i'm not too bothered, though i agree that `Int64` should be fir first.

bcardiff: The `PullParser` stores an Int64 to avoid having a union there, for performance reasons.

I think the order should be such that if it fits in an Int8 (and Int8 is an option for the mapping), then Int8 is used. Even if Int32 was also in the mapping.

Although I think this is a corner case and we might not support it.

Same with floats | ints, it should be used only if the other types are not enough to hold the value or if a decimal separator is explicitly used.

RX14: @bcardiff i'd prefer to always use the largest type in the union, so that if you deserialize `Array(Int32 | Int8)`, you get all int32's.

Again, this really doesn't matter. If you're doing that, you're doing it wrong.

bcardiff: Ok. for `Int32 | Int8` it might be more stable to return always `Int32` in that case.

I haven't checked how it currently works but whay is your opinion regarding `Int32 | Float32` ? Should it behave as it was `Float32` alone? For the UInt64::MAX < Float32::MAX but that is no true for UInt128. So I would prefer to honor `Int32 | Float32` depending if the input is a float or int based on what I said before.

straight-shoota: > So I would prefer to honor Int32 | Float32 depending if the input is a float or int based on what I said before.

This sounds acceptable, but I'm not sure if the JSON parser should read some context into number literals when there is *absolutely* no difference between `1.0` and `1` literals in JSON.

I'd probably prefer the largest float data type over any int.

KimBurgess: As a user of the std lib, a primary argument _for_ reading into the that decimal point is to keep encode/decode behavior symmetric.

KimBurgess: Looks like we still have some split opinions. In the interests of keeping things moving / forcing a decision I've amended the order to match what was mentioned to @RX14 above as well as adding spec to demonstrate stability when pulling into a (highly questionable) union of different integer types as well as symmetry of encode/decode for the `Int | Float` case.

A re-review would be lovely as people have time.

src/json/from_json.cr
```diff
@@ -226,42 +226,55 @@ end
...
+    {% for type in type_order.select { |t| T.includes? t } %}
+      value = pull.read?({{type}})
+      return value unless value.nil?
     {% end %}
+    when .float?
+    {% type_order = [Float64, Float32] %}
+    {% for type in type_order.select { |t| T.includes? t } %}
+      value = pull.read?({{type}})
+      return value unless value.nil?
+    {% end %}
```
RX14: Do we need this *and* `Float64, Float32` in the above branch??

KimBurgess: Yes. The `int?` case handles extracting a value if the parser picks up the original value as an integer, but the only numeric in the Union is a float. This then captures float → float.

RX14: Oh right, I forgot about the `when .float?` because of the indentation.

PR #8692, opened 2020-01-16
Spec::JUnitFormatter: Allow output directory as --junit_output

bcardiff:
Follow up of https://github.com/crystal-lang/crystal/pull/8599#issuecomment-575214414

### DIFF
--- a/src/spec/junit_formatter.cr
+++ b/src/spec/junit_formatter.cr
@@ -35,6 +35,10 @@ module Spec
     end
 
     def self.file(output_path : Path)
+      if output_path.extension != ".xml"
+        output_path = output_path.join("output.xml")
+      end
+
       Dir.mkdir_p(output_path.dirname)
       file = File.new(output_path, "w")
       JUnitFormatter.new(file)


RX14:
Why not just append `output.xml` to your configuration? I really don't get why hardcoding `output.xml` has any benefit at all here.

bcardiff:
Because it keeps compatible with the previous behaviour without ambiguity.
And leave space for future improvements.
Allowing `--junit_output ./test_results/foo.xml` should have not invalidate `--junit_output ./test_results`

RX14:
Nobody has explained why splitting the results would be an improvement, and we've already made the breaking change in a release. I see no benefit to previous behaviour at this point.

bcardiff:
This is preventing the breaking change on master. There was no tagged release with that behavior.

I mentioned here some scenarios for splitting https://github.com/crystal-lang/crystal/pull/8599#issuecomment-575214414 .

RX14:
Right, sorry, I thought that'd made it into a release somehow.

src/spec/junit_formatter.cr
```diff
@@ -35,6 +35,10 @@ module Spec
     end
 
     def self.file(output_path : Path)
+      if output_path.extension != ".xml"
+        output_path = output_path.join("output.xml")
+      end
+
```
Sija: That belongs to `src/spec.cr` not here.

Sija: That belongs to `src/spec.cr`, not here.

https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/spec.cr#L132-L135

Sija: IMHO `.file` is expecting a path to... the file - not some arbitrary path to mutate afterwards.

bcardiff: For me is a path configuration of the output formatter. If there is splitting how to split and which file to create would belong in the formatter and not spec itself since it is not something shared with other formatters like DotFormatter.

Sija: > [...] since it is not something shared with other formatters like DotFormatter.

That's exactly my point, formatter itself should be responsible for doing any magic with the given path. If I'm calling `Spec::JUnitFormatter.file(path)` i expect to pass a **file path** to it...

straight-shoota:
@RX14 I agree, it's not ideal solution. At some point we need to figure out if writing to different files is actually going to happen and we can reconsider and if necessary break things then.

RX14:
I'd prefer to rule out ever doing that, as a form of YAGNI simplification.

PR #8693, opened 2020-01-16
Add a `empty?` method for `IO`.

KaneRoot:
This PR adds an **empty?** method to the `IO` class allowing to know when there are still data in the input buffer. This is relevant for any class including the `IO::Buffered` module, such as the `Socket` class for instance.

This problem has been discussed here: https://forum.crystal-lang.org/t/check-for-new-received-message-in-http-websocket/1599/3

Thanks to the reviewers!

EDIT: of course, it's always when you pushed your code that you see your typos everywhere... but now it's fixed.

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -128,6 +128,13 @@ abstract class IO
     raise IO::Error.new "Closed stream" if closed?
   end
 
+  # Returns `true` when all buffered data as been read.
+  #
+  # `IO` defines this as a `true` value, but including types may override.
+  def empty? : Bool
+    true
+  end
+
   # Flushes buffered data, if any.
   #
   # `IO` defines this is a no-op method, but including types may override.
--- a/src/io.cr
+++ b/src/io.cr
@@ -128,7 +128,7 @@ abstract class IO
     raise IO::Error.new "Closed stream" if closed?
   end
 
-  # Returns `true` when all buffered data as been read.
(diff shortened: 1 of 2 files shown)


straight-shoota:
I'm sorry, but there are multiple issues with this proposal:

Why is this method on `IO` when it's supposed behaviour only works for `IO::Buffered`? There's no way to tell whether an unbuffered IO's buffer is empty.
Also there are two separate buffers in `IO::Buffered`, in and out buffers. Such a method should either return the status of both buffers, or the method should be specifically named according to the respective buffer.
The name doesn't make sense either, because you'd probably expect `IO#empty?` to tell whether the IO has any content *at all*. Not just if there is some content in the buffer. So this should rather be something like `buffer_empty?`. 

And finally, I'm not entirely sure why you would need such a method. The behaviour of `IO::Buffered` should be completely encapsulated and transparent to the outside. I don't think there is a valid reason for any introspection into its internal state.

The use case you described on the forum is a bit hard to follow. I'm sure there are different solutions to the problems you're facing, which seem to be related to your custom modifications to the stdlib websocket implementation. It's probably best to discuss the use case and incentive in the forum.

j8r:
Looks like it can be closed, I hope you have understood @KaneRoot how to better tackle the problem (feel free to ask in the forum :slightly_smiling_face: )

PR #8698, opened 2020-01-18
`JSON::PullParser` overflow handling

### LINKED ISSUE #8694: JSON::PullParser numeric overflow behaviour
The current implementation of `JSON::PullParser` includes a [set of `#read?` methods](https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/json/pull_parser.cr#L267-L315). In the case of an overflow these will silently wrap the value to fit the target type. Usage of these within std lib is limited to [within `Union.from_json`](https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/json/from_json.cr#L242).

Similarly, [methods for instantiating numeric primitives from a `PullParser`](https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/json/from_json.cr#L69-L81) all silently wrap.

This can lead to some unexpected behaviour as shown [here](https://github.com/crystal-lang/crystal/pull/8689#discussion_r367175575) and [here](https://github.com/crystal-lang/crystal/pull/8686#issuecomment-573830506).

It appears that it may be more desirable to have both of this instances use conversions that raise an `OverflowError` in cases where the parsed value cannot be represented by the target type.

Happy to submit a PR for this, but opening up for discussion / approval first.

KimBurgess:
Changes behaviour of:

- `JSON::PullParser#read?` to either fully read into the specified type, or return nil
- Numeric primitive instantiation from `JSON::PullParser` to raise a `JSON::ParseException` wrapping an `OverflowError` in the case of overflow.

Fixes #8694

### DIFF
--- a/spec/std/json/pull_parser_spec.cr
+++ b/spec/std/json/pull_parser_spec.cr
@@ -329,4 +329,49 @@ describe JSON::PullParser do
     assert_raw %({"foo":[1,2,{"bar":[1,"hello",true,false,1.5]}]})
     assert_raw %({"foo":"bar"})
   end
+
+  describe "read?" do
+    {% for pair in [[Int8, 1_i8],
+                    [Int16, 1_i16],
+                    [Int32, 1_i32],
+                    [Int64, 1_i64],
+                    [UInt8, 1_u8],
+                    [UInt16, 1_u16],
+                    [UInt32, 1_u32],
+                    [UInt64, 1_u64],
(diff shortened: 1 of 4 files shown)


straight-shoota:
Could you please add specs for `read?` as well?

PR #8699, opened 2020-01-18
Fix expected error message in libxml2 error spec

straight-shoota:
Starting from libxml2 2.9.10 the error message for the case we're testing has changed from `Premature end of data in tag people line 1` to `EndTag: '</' not found`.
The change was introduced by this commit in libxml2: https://gitlab.gnome.org/GNOME/libxml2/commit/62150ed2ab19a4dd76c15acc62c7d923d9f3b2cc

### DIFF
--- a/spec/std/xml/xml_spec.cr
+++ b/spec/std/xml/xml_spec.cr
@@ -158,23 +158,14 @@ describe XML do
   end
 
   it "handles errors" do
-    xml = XML.parse(%(<people>))
+    xml = XML.parse(%(<people></foo>))
     xml.root.not_nil!.name.should eq("people")
     errors = xml.errors.not_nil!
     errors.size.should eq(1)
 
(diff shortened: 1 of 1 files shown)


straight-shoota:
We could probably also just try to trigger a different error to make the spec code less complex. The specific error doesn't seem to matter, the spec just needs to ensure error handling works in general.

straight-shoota:
@bcardiff Yes, but the purpose of this spec is testing error handling in general. It doesn't matter which specific error is used as an example. It just needs to make sure the error messages fits to the error being tested.

yxhuvud:
@bcardiff: Unterminated XML is not fundamentally different to wrongly structured XML anyhow. Compare with the following error message from the compiler: `Error: expecting identifier 'end', not 'EOF'`. Is that badly structured or is it an unterminated something? The answer is that it can be both (often at the same time), and it is typically not super easy to see where the error is without invoking the formatter.

Issue #8704, opened 2020-01-21
Typo in Channel causes break


bararchy:
code to trigger is using select

```crystal
      select
      when data = chan.recieve
        data.to_http
      when timeout_chan.receive
        raise IO::Timeout.new
      end
```

```crystal
 334 | when data = chan.recieve
                        ^------
Error: undefined method 'recieve_select_action' for Channel(Nil)

Did you mean 'receive_select_action'?
```

bcardiff:
Fixed by #8705

Sija:
@bcardiff It's not really fixed, the issue here is cryptic error message related to the potential typo.

bcardiff:
I agree that the errors introduced by syntax sugar transformation are more cryptic than they should be.

Another example of those are:

```cr
class Foo
end
 
Foo{1,2,3}
# error in line 4
# Error: undefined method '<<' for Foo
```

We can open another issue to frame the issue differently and cover other cases.

PR #8705, opened 2020-01-21
Fix typo in example

vlazar:
Related #8704

Since #8506 is in 0.33.0 already it would make sense to add this fix in 0.33.0 too.


bcardiff:
Thanks @vlazar , @bararchy

PR #8707, opened 2020-01-21
IO: Detect closed IO when resuming readable/writable event

bcardiff:
This should fix some race conditions detected in the preview_mt build.

I think there are still some race conditions in `IO::Evented` since there is no guarantee that, for example, a thread closes the IO while another thread has already initiated a read operation, succeed the `check_open` in `IO::Bufferend#read`, but haven't reach the unbuffered_read -> evented_read -> wait_readable.

There is currently no mechanism to ensure that the check for open IO holds long enough until the fibers reschedule.

Even without fixing that scenario this PR should improove the current state for MT:

* If the IO was closed by another thread while waiting we need to check if it's open or not.
* If it's closed an IO::Error is raised as a result of the read/write operation as if the IO was has been closed from the beginning.
* The event_close is now performed before actually closing the underlying file descriptor to prevent libevent accessing it in a closed state.
* The @closed = true in IO::FileDescriptor needs to happen before the system_close so it is available when the awaiting fibers are woken up.

### DIFF
--- a/src/crystal/system/file.cr
+++ b/src/crystal/system/file.cr
@@ -40,6 +40,13 @@ module Crystal::System::File
 
     oflag = m | o
   end
+
+  # Closes the internal file descriptor without notifying libevent.
+  # This is directly used after the fork of a process to close the
+  # parent's Crystal::Signal.@@pipe reference before re initializing
+  # the event loop. In the case of a fork that will exec there is even
+  # no need to initialize the event loop at all.
+  # def file_description_close
 end
 
 {% if flag?(:unix) %}
--- a/src/crystal/system/file.cr
+++ b/src/crystal/system/file.cr
@@ -46,7 +46,7 @@ module Crystal::System::File
   # parent's Crystal::Signal.@@pipe reference before re initializing
   # the event loop. In the case of a fork that will exec there is even
   # no need to initialize the event loop at all.
-  # def file_description_close
+  # def file_descriptor_close
 end
 
 {% if flag?(:unix) %}
--- a/src/crystal/system/unix/file_descriptor.cr
+++ b/src/crystal/system/unix/file_descriptor.cr
@@ -115,10 +115,10 @@ module Crystal::System::FileDescriptor
     # always lead to undefined results. This is not specific to libevent.
     evented_close
 
-    file_description_close
+    file_descriptor_close
   end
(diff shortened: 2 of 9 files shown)


wontruefree:
Looks like your mt_preview build timed out.  Is that because of this change?

bcardiff:
No it's something Process.

Running `./bin/crystal spec spec/std/callstack_spec.cr:16` hangs.

And if the changes in `private def system_close` are reverted it works. But we are back to leaving closed fd in libevent...

bcardiff:
And the CI is happy 🎉

src/crystal/system/unix/file_descriptor.cr
```diff
@@ -115,6 +115,10 @@ module Crystal::System::FileDescriptor
     # always lead to undefined results. This is not specific to libevent.
     evented_close
 
+    file_description_close
+  end
+
+  def file_description_close
```
RX14: `descriptor`

src/crystal/system/file.cr
```diff
@@ -40,6 +40,13 @@ module Crystal::System::File
 
     oflag = m | o
   end
+
+  # Closes the internal file descriptor without notifying libevent.
+  # This is directly used after the fork of a process to close the
+  # parent's Crystal::Signal.@@pipe reference before re initializing
+  # the event loop. In the case of a fork that will exec there is even
+  # no need to initialize the event loop at all.
+  # def file_description_close
```
RX14: This seems to be misplaced.....

bcardiff: Why? is the intended behavior of the method that should implement all the systems.

RX14: This comment is inside a method....

RX14: Oh, I'm blind, never mind!!

src/socket.cr
```diff
@@ -250,7 +250,7 @@ class Socket < IO
         if closed?
           return
         elsif Errno.value == Errno::EAGAIN
-          wait_readable
+          wait_readable rescue nil
```
grioja, line 253: @bcardiff shouldn't be this `rescue return nil` ? The loop doesn't break otherwise. We're seeing hanging on closing a socket.

PR #8713, opened 2020-01-23
Added documentation for HTTP::WebSocket#run

### LINKED ISSUE #5641: HTTP::WebSocket#run
Hello!

I did not find documentation about HTTP::WebSocket#run

I am not certain, am I doing it correct, but I was able to get it working as following:

```crystal
external_connection = HTTP::WebSocket.new(uri)

spawn do
  external_connection.run
end

external_connection.send("subscribe channel")

external_connection.on_message do |message|
  puts "Received message: #{message}"
end
```

1. Why run blocks all flow?
2. It seems a bit counter intuitive that "subscribe channel" works before connection is initiated in fiber (thread)? Some clever buffering?
3. In Ruby there is #on_open, that is quite clear DSL imho.

Could you please teach, am I doing it correctly? (I am willing to PR documentation update, but I am having trouble understanding why it works like that.)

RubenBrocke:
resolves #5641
Took some simple info from the issue

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -100,6 +100,20 @@ class HTTP::WebSocket
     @ws.close(message)
   end
 
+  # Will run the websocket in an infinite loop where it receives messages and calls previously set callbacks.
+  # The websocket will autmatically respond with pong on pings and exit the loop when the websocket is closed.
+  #  
+  # ```
+  # ws = WebSocket.new(uri)
+  # 
+  # # Create callback
+  # ws.on_message do |msg|
+  #   ws.send "response"
+  # end
+  #
+  # # Start infinite loop
+  # ws.run
+  # ```
   def run
     loop do
       begin


src/http/web_socket.cr
```diff
@@ -100,6 +100,20 @@ class HTTP::WebSocket
     @ws.close(message)
   end
 
+  # Will run the websocket in an infinite loop where it receives messages and calls previously set callbacks.
```
jhass:
```suggestion
  # Continuously receives messages and calls the previously set callbacks until the Websockt is closed.
```

"Running a socket" sounds a bit weird to me, how about this? :)

src/http/web_socket.cr
```diff
@@ -100,6 +100,20 @@ class HTTP::WebSocket
...
+  # The websocket will autmatically respond with pong on pings and exit the loop when the websocket is closed.
```
jhass:
```suggestion
  # Ping and pong messages are automatically handled.
```

It's not the socket that answers the request, but the loop inside this method :)

RubenBrocke: Seems good to me!
(It's my first contribution so sorry for the small problems)

jhass: Wanna apply the suggestions then? :) Since you didn't give maintainer push rights to your fork I can't do it myself in this PR :)

PR #8718, opened 2020-01-26
Skip explicit creation of docs/ dir

### LINKED ISSUE #8717: crystal docs -f json still makes docs folder
Hi, I've been using the crystal docs json generator tool for my own uses. 

One behavior that I found to be unexpected was that even when the format is specified to json the docs folder is still generated.

The problem for me is that I'm using this in a series of bash scripts and I have to make sure to cleanup the unnecessary docs folder. This to me seems like unintended behavior.

Sija:
Fixes #8717

### DIFF
--- a/src/compiler/crystal/tools/doc/generator.cr
+++ b/src/compiler/crystal/tools/doc/generator.cr
@@ -44,8 +44,6 @@ class Crystal::Doc::Generator
   end
 
   def run
-    Dir.mkdir_p @output_dir
-
     types = collect_subtypes(@program)
 
     program_type = type(@program)


jhass:
Well, we still want to do it for HTML output, no?

Sija:
@jhass And it's done - see my commit comment.

jhass:
Ah good catch. Sorry, the mobile client didn't make it obvious there's one :)

straight-shoota:
I'd still keep this line and add is as first expression in `generate_docs_html`. That's preferable to an implicit creation which is purely based on order of instructions.

PR #8720, opened 2020-01-28
Add Object#in?

### LINKED ISSUE #8671: Object#in?
Maybe this is redundant, but I have always thought that a Python-ish `obj.in?(enum)` would be more intuitive than the Ruby-ish `enum.includes?(obj)`:

```crystal
class Object
  def in?(e : Enumerable) : Bool
    e.includes?(self)
  end
end

1.in? 1..10  #=> true
[1, 2].in? [[1, 2], [3, 4]]  #=> true
```

Sija:
Closes #8671

### DIFF
--- a/src/object.cr
+++ b/src/object.cr
@@ -201,6 +201,23 @@ class Object
     yield self
   end
 
+  # Returns `true` if `self` is included in the *collection* argument.
+  #
+  # ```
+  # 10.in?(0..100)     # => true
+  # 10.in?({0, 1, 10}) # => true
+  # 10.in?(0, 1, 10)   # => true
+  # 10.in?(:foo, :bar) # => false
+  # ```
+  def in?(collection) : Bool
+    collection.includes?(self)
+  end
+
+  # :ditto:
+  def in?(*values : Object) : Bool
+    in?(values)
+  end
+
   # Returns `self`.
   # `Nil` overrides this method and raises `NilAssertionError`, see `Nil#not_nil!`.
   def not_nil!
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -444,6 +444,26 @@ describe Object do
     end
   end
 
+  describe "#in?" do
+    it "works with Enumerable-s" do
+      :foo.in?([:foo, :bar]).should be_true
+      :bar.in?({:foo, :baz}).should be_false
+      42.in?(0..100).should be_true
+      4242.in?(0..100).should be_false
+    end
+
+    it "works with splatted arguments" do
+      :baz.in?(:foo, :bar).should be_false
+      1.in?(1, 10, 100).should be_true
+    end
+
+    it "works with other objects implementing #includes?" do
+      "o".in?("foo").should be_true
+      'o'.in?("foo").should be_true
+      'x'.in?("foo").should be_false
+    end
+  end
+
   it "#unsafe_as" do
     0x12345678.unsafe_as(Tuple(UInt8, UInt8, UInt8, UInt8)).should eq({0x78, 0x56, 0x34, 0x12})
   end


src/object.cr
```diff
@@ -201,6 +201,23 @@ class Object
...
 
+  # Returns `true` if `self` is included in the *collection* argument.
+  #
+  # ```
+  # 10.in?(0..100)     # => true
+  # 10.in?({0, 1, 10}) # => true
+  # 10.in?(0, 1, 10)   # => true
+  # 10.in?(:foo, :bar) # => false
+  # ```
+  def in?(collection : Enumerable) : Bool
```
j8r: This means `Object` depends on `Enumerable`, and `vice-versa`.
Furthermore, any object that implements `includes?` is valid, not only enumerables.
Better to drop this type annotation.

src/object.cr
```diff
@@ -201,6 +201,23 @@ class Object
...
+    collection.includes?(self)
+  end
+
+  # :ditto:
+  def in?(*values : Object) : Bool
```
RX14: I'd rather drop this type restriction too

Sija: That's on purpose - otherwise it will be possible to call it with no arguments.

straight-shoota: Hm, that's weird. Why does having a type restriction enforce size > 0...

Alternatively, this should work: `def in?(value, *values)`. It's not as nice, though.

Sija: > Hm, that's weird. Why does having a type restriction enforce size > 0...

I dunno, it's probably better to ask @asterite about that.

>Alternatively, this should work: `def in?(value, *values)`. It's not as nice, though.

It's not. I'd rather leave it like it is right now, regardless the perceived _weirdness_.

asterite: Because if you specify a type, any empty tuple would match it, and if you have multiple overloads with different type restrictions it's not clear which should match. So the solution is: none match. The empty tuple doesn't match a splat with a type restriction. And if you want to match that you provide another overload without args.

And specifying `Object` is a convenient way of saying "match one or more args".

asterite: I think this overload is confusing and I'd remove it.

```crysta
1.in?([1, 2, 3], 4) # => false
1.in?([1, 2, 3]) # => true
```

That is, it's not clear that the one-argument version is a specialization of the multi-args but with one arg, or a completely different behavior.

You can always do `elem.in?({a, b, c})`

RX14: @asterite wow! Thanks I never knew this trick.

straight-shoota:
Do we really want the splat overload? I agree with @asterite that it's confusing (https://github.com/crystal-lang/crystal/pull/8720#discussion_r371929818).

asterite:
Nevermind, maybe it's good and convenient.

src/object.cr
Sija: @asterite IMO this method will be most be used in place of `{:foo, :bar, :baz}.includes?(val)` notation (used often with `Errno#errno` or `HTTP::Client::Response#status_code`), mixing the splats and arguments is sth more of an edge-case.

src/object.cr
```diff
@@ -201,6 +201,23 @@ class Object
...
+  # 10.in?({0, 1, 10}) # => true
+  # 10.in?(0, 1, 10)   # => true
+  # 10.in?(:foo, :bar) # => false
+  # ```
+  def in?(collection) : Bool
+    collection.includes?(self)
+  end
+
+  # :ditto:
+  def in?(*values : Object) : Bool
```
wontruefree: I am curious. Wouldn't everything inherit from `Object`?
Would this ever constrain the types that this could call?

Sija: You seems to have missed https://github.com/crystal-lang/crystal/pull/8720#discussion_r371924639

PR #8723, opened 2020-01-30
Use Object#in? throughout the stdlib

Sija:
Followup to #8720

### DIFF
--- a/src/string.cr
+++ b/src/string.cr
@@ -2260,7 +2260,7 @@ class String
   # returns the number of times the block returned a truthy value.
   #
   # ```
-  # "aabbcc".count { |c| ['a', 'b'].includes?(c) } # => 4
+  # "aabbcc".count &.in?('a', 'b') # => 4
   # ```
   def count
     count = 0
@@ -2291,7 +2291,7 @@ class String
   # block returned a truthy value removed.
   #
   # ```
-  # "aabbcc".delete { |c| ['a', 'b'].includes?(c) } # => "cc"
+  # "aabbcc".delete &.in?('a', 'b') # => "cc"
   # ```
   def delete
     String.build(bytesize) do |buffer|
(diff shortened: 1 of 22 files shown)


spec/std/array_spec.cr
```diff
@@ -1002,7 +1002,7 @@ describe "Array" do
       [1].sample.should eq(1)
 
       x = [1, 2, 3].sample
-      [1, 2, 3].includes?(x).should be_true
+      x.in?(1, 2, 3).should be_true
```
ysbaddaden: We are testing arrays here, I'd keep the `includes?` call. Same below.

src/uri/encoding.cr
```diff
@@ -172,8 +172,8 @@ class URI
   # Unreserved characters are ASCII letters, ASCII digits, `_`, `.`, `-` and `~`.
   def self.unreserved?(byte) : Bool
     char = byte.unsafe_chr
-    char.ascii_alphanumeric? ||
-      {'_', '.', '-', '~'}.includes?(char)
+    char.ascii_alphanumeric? |
+      char.in?('_', '.', '-', '~')
```
ysbaddaden: bug: you changed `||` into `|`.

src/string.cr
```diff
@@ -2260,7 +2260,7 @@ class String
   # returns the number of times the block returned a truthy value.
   #
   # ```
-  # "aabbcc".count { |c| ['a', 'b'].includes?(c) } # => 4
+  # "aabbcc".count(&.in?('a', 'b')) # => 4
```
bcardiff:
```suggestion
  # "aabbcc".count &.in?('a', 'b') # => 4
```

Usually the last block is used without parenthesis in other doc examples.
There are other ocurrences also of this in the PR

Sija: I'm fine with either version.

Should I remove the parenthesis from all of the examples?
- 👍 — YES
- 👎 — NO

Sija: Done.

Sija:
GTG?

PR #8724, opened 2020-01-30
Ensure sending fiber finished in select specs

bcardiff:
These specs are failing every now and then on preview_mt.

The changes ensure the sending fiber finished executing before performing the assertion in the main fiber.

Sample of the failure: https://circleci.com/gh/crystal-lang/crystal/37483?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/spec/std/concurrent/select_spec.cr
+++ b/spec/std/concurrent/select_spec.cr
@@ -128,7 +128,7 @@ describe "select" do
     end
 
     ch3.receive.should eq(3)
-    while !f.dead?
+    until f.dead?
       Fiber.yield
     end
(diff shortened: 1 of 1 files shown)


spec/std/concurrent/select_spec.cr
```diff
@@ -128,7 +128,9 @@ describe "select" do
     end
 
     ch3.receive.should eq(3)
-    Fiber.yield
+    while !f.dead?
```
Sija: Tiny suggestion:

```suggestion
    until f.dead?
```

ditto elsewhere

PR #8733, opened 2020-02-03
IO: Polish IO closing for MT

bcardiff:
For now IO#close is not thread-safe. You should not close the IO from concurrently.

Even though if that's the case, before this PR the LibEvent events could be freed twice and the awaiting readers/writers fibers could be enqueued again. This is now fixed by the changes in `io/evented.cr`.

I also align the exception raised when the awaiting reader is awakened with a EBADF / closed fd.

The `IO::Error` added is to match the description when a writer is awakened with a EBADF. These does not match 100% the error if the fd was closed from the beginning. This bugs me a little.

Also the `socket.cr` does not have this special handling of Errno in `unbuffered_read`/`unbuffered_write`.

The exact exception type and message might come back with Errno exceptions are revisited.

After this PR if `IO#close` is called concurrently from multiple threads one might fail with `Errno.new("Error closing file")` in `Crystal::System::FileDescriptor#file_descriptor_close` since the fd might be already closed. Without adding a lock for `IO#close` it is not possible to prevent this situation. The implementation of `evented_close` it is thread-safe now and that protects the runtime enough.

### DIFF
--- a/src/crystal/system/win32/file_descriptor.cr
+++ b/src/crystal/system/win32/file_descriptor.cr
@@ -6,7 +6,11 @@ module Crystal::System::FileDescriptor
   private def unbuffered_read(slice : Bytes)
     bytes_read = LibC._read(@fd, slice, slice.size)
     if bytes_read == -1
-      raise Errno.new("Error reading file")
+      if Errno.value == Errno::EBADF
+        raise IO::Error.new "File not open for reading"
+      else
+        raise Errno.new("Error reading file")
+      end
     end
     bytes_read
   end
--- a/spec/std/io/io_spec.cr
+++ b/spec/std/io/io_spec.cr
@@ -386,6 +386,14 @@ describe IO do
 
       str.read_fully?(slice).should be_nil
     end
+
+    it "raises if trying to read to an IO not opened for reading" do
+      IO.pipe do |r, w|
+        expect_raises(IO::Error, "File not open for reading") do
+          w.gets
+        end
+      end
+    end
   end
 
   describe "write operations" do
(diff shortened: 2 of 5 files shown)


RX14:
I'd at least like a better name...

bcardiff:
The lock in thread local value did not need to be exposed until now at least.

I agree the name is not the best, but it's private api and is only used in the lines that were changed in this PR. I see it more like a helper for the implementation.

Another name could be `consume_each`  or something like that. But altough each_and_clear seems ugly, it's pretty clear what it does.

RX14:
I'd prefer `consume_each` actually, it seems more obvious to me what it does. But only a style issue, I'll ack this if you don't wish to change the name.

bcardiff:
Rebased on master and method renamed. Approval pending.

src/crystal/system/win32/file_descriptor.cr
```diff
@@ -6,7 +6,11 @@ module Crystal::System::FileDescriptor
   private def unbuffered_read(slice : Bytes)
     bytes_read = LibC._read(@fd, slice, slice.size)
     if bytes_read == -1
-      raise Errno.new("Error reading file")
+      if Errno.value == Errno::EBADF
```
RX14: File descriptors can be reused - if this `if` branch can be hit, then doesn't that imply that a write may happen to the wrong file descriptor because of a race condition?

bcardiff: I'm not sure how probably that reutilization.
The same check was added to `#unbuffered_write` in #6497.
Also with waj we mention that these checks might need to move to evented directly.

RX14: > I'm not sure how probably that reutilization

Well they're sequential, so very probably.

If you `open`, `close`, `open` again, the second open will have exactly the same FD as the first.

rdp: Yeah looks like a possible race condition, tricky one... Appears that java sets the internal "fd" to -1 before doing a close: https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/solaris/native/java/io/io_util_md.c#L93 they also synchronize the calls to close FWIW: https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/io/FileInputStream.java#L289

Issue #8387, opened 2019-10-28
Bug: Failed to raise an exception: END_OF_STACK


asterite:
Code to reproduce:

```crystal
class Foo
  def initialize
    @buffer = Pointer(UInt8).malloc(17)
    header = @buffer.as({Int32, Int32, Int32}*)
    header.value = {String::TYPE_ID, 4, 0}
  end

  def to_s
    @buffer.as(String)
  end
end

string = Foo.new.to_s
data = string.to_slice
if data[0, 4].to_a != [] of UInt8
  raise "OH NO!"
end
```

You have to compile it with `--release`, then when you run it you get:

```
Failed to raise an exception: END_OF_STACK
[0x1040f3f27] *CallStack::print_backtrace:Int32 +39
[0x1040ccc90] __crystal_raise +48
[0x1040d2468] *raise<NilAssertionError>:NoReturn +136
[0x1040d2361] *Nil#not_nil!:NoReturn +17
[0x1040f3b97] __crystal_sigfault_handler +55
[0x7fff79632b5d] _sigtramp +29
```

This was extracted from a larger code that uses `String.build`, which essentially does the same as `Foo`.

I don't know what's going on, maybe stack corruption of some sort.

ysbaddaden:
It seems to break when trying to print the backtrace. What happens if you disable it?

kostya:
similar error on linux and mac, with `Crystal 0.31.0 (2019-09-24) LLVM: 8.0.1 Default target: x86_64-apple-macosx` with pipes.
```
crystal --version | echo

Error writing file: Broken pipe (Errno)
  from IO::FileDescriptor@Crystal::System::FileDescriptor#unbuffered_write<Slice(UInt8)>:Nil
  from IO::FileDescriptor@IO::Buffered#flush:IO::FileDescriptor
  from exit<Int32>:NoReturn
  from exit:NoReturn
  from Crystal::Command#run:(Bool | Crystal::Compiler::Result | Nil)
  from __crystal_main
  from main
Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
Unhandled exception: Error writing file: Broken pipe (Errno)
  from IO::FileDescriptor@Crystal::System::FileDescriptor#unbuffered_write<Slice(UInt8)>:Nil
  from IO::FileDescriptor@IO::Buffered#flush:IO::FileDescriptor
[...]

Issue #6249, opened 2018-06-23
you've found a bug in the Crystal compiler.


set:
I'm created a new Dokku app in my Digitalocean droplet. Show me this error when I push Amber app with Buildpack. What's the solution?

```
Counting objects: 58, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (47/47), done.
Writing objects: 100% (58/58), 35.59 KiB | 3.56 MiB/s, done.
Total 58 (delta 1), reused 0 (delta 0)
-----> Cleaning up...
-----> Building amber from herokuish...
-----> Adding BUILD_ENV to build environment...
-----> Fetching custom buildpack
-----> Crystal app detected
-----> Installing Crystal (0.25.0 due to latest release at https://github.com/crystal-lang/crystal) from https://github.com/crystal-lang/crystal/releases/download/0.25.0/crystal-0.25.0-1-linux-x86_64.tar.gz... done
-----> Installing Dependencies... Fetching https://github.com/amberframework/amber.git
       Fetching https://github.com/amberframework/amber-router.git
       Fetching https://github.com/mosop/cli.git
       Fetching https://github.com/mosop/optarg.git
       Fetching https://github.com/mosop/callback.git
       Fetching https://github.com/mosop/string_inflection.git
       Fetching https://github.com/elorest/compiled_license.git
       Fetching https://github.com/jeromegn/kilt.git
       Fetching https://github.com/TechMagister/liquid.cr.git
       Fetching https://github.com/phoffer/inflector.cr.git
       Fetching https://github.com/amberframework/micrate.git
       Fetching https://github.com/crystal-lang/crystal-db.git
       Fetching https://github.com/crystal-lang/crystal-mysql.git
       Fetching https://github.com/will/crystal-pg.git
       Fetching https://github.com/stefanwille/crystal-redis.git
       Fetching https://github.com/ysbaddaden/pool.git
       Fetching https://github.com/jwaldrip/shell-table.cr.git
       Fetching https://github.com/jeromegn/slang.git
       Fetching https://github.com/crystal-lang/crystal-sqlite3.git
       Fetching https://github.com/mosop/teeplate.git
       Fetching https://github.com/amberframework/granite.git
       Fetching https://github.com/amberframework/quartz-mailer.git
       Fetching https://github.com/arcage/crystal-email.git
       Fetching https://github.com/amberframework/jasper-helpers.git
       Fetching https://github.com/amberframework/citrine-i18n.git
[...]

PR #8735, opened 2020-02-03
Top level exception handler

waj:
There are many issues where the reported error contains a `Failed to raise an exception: END_OF_STACK`. Many times this is because an error happens outside the `begin/rescue` around `__crystal_main` and it ends hiding the real source of the issue. (#8387, #6249, #7810, among others)

I added this exception handler a the top level of `Crystal.main` to dump the unhandled exception. I tried to be as conservative as possible, executing the least amount of Crystal's std code.

### DIFF
--- a/src/crystal/main.cr
+++ b/src/crystal/main.cr
@@ -82,8 +82,10 @@ module Crystal
   # is not setup yet, so nothing that allocates memory
   # in Crystal (like `new` for classes) can be used.
   def self.main(argc : Int32, argv : UInt8**)
-    main do
-      main_user_code(argc, argv)
+    handle_exceptions do
+      main do
+        main_user_code(argc, argv)
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


RX14:
Can we merge this for `0.33.0`?

RX14:
Can we merge this for `0.33.0`?

bcardiff:
There is a need to a dprintf replacement for Windows. We were thinking of adding a method in the Crystal namespace that will mimic a dprintf in Windows. That's the thing that is missing to avoid breaking the build.

src/crystal/main.cr
```diff
@@ -96,6 +98,26 @@ module Crystal
   def self.main_user_code(argc : Int32, argv : UInt8**)
     LibCrystalMain.__crystal_main(argc, argv)
   end
+
+  protected def self.handle_exceptions
+    begin
```
Sija: `begin ... end` pair is redundant here.

Issue #8682, opened 2020-01-12
update `crystal init` to handle `.`


jethrodaniel:
Currently, `crystal init <type> .` doesn't expand the `.` to the current directory, which I found odd, compared to say, `rails new .`

I've got a PR with the feature [here](https://github.com/crystal-lang/crystal/pull/8681).

straight-shoota:
The entire logic is flawed and fails for any value other than a plain directory name without any hierarchical component. The `NAME` argument is used literally as value for the shard name, so `crystal init lib foo/bar` results in `name: foo/bar`.

There are several aspects to this problem:

* `config.name` should be validated to match the [specification for a shard name](https://github.com/crystal-lang/shards/blob/master/SPEC.md#name). Otherwise, `crystal init` should fail.
* The first argument should always be expanded using `File.expand` (with `home: true`) and the name retrieved using `File.basename`, unless obviously, both `NAME` and `DIR` arguments are present.
* Following that, the arguments usage should probably rather be specified as `crystal init TYPE (DIR | NAME DIR)`.

jethrodaniel:
Awesome points, didn't even know we had explict shard naming rules.

Will update the pr when i get time

bcardiff:
Regarding the shard convention, I think it would be handy to remove `crystal-` prefix and `.cr`/`-crystal` suffix from dir name to determine the shard's name.

asterite:
Fixed by #8681

Please open other issues/PRs for the [remaining issues](https://github.com/crystal-lang/crystal/issues/8682#issuecomment-573420302).

PR #8681, opened 2020-01-12
update `crystal init` to handle `.`

jethrodaniel:

PR #8737, opened 2020-02-03
Refactor Crystal::Init adding name validation

straight-shoota:
Follows up on #8681 with improvements from https://github.com/crystal-lang/crystal/issues/8682#issuecomment-573420302:

* If only two argument are given to `crystal init` (the first is always the type), the second one is `DIR`. `NAME` is derived from its basename. When three arguments are given, the second is `NAME` and the third one `DIR` (as before). The effective change is now you can pass in only a directory path and the name is automatically extracted from the path instead of (unexpectedly) using the entire path as name.
* `DIR` is expanded using `Path#expand(home: true)`
* `NAME` is validated to match the [specification for a shard name](https://github.com/crystal-lang/shards/blob/master/SPEC.md#name).

This PR also adds specs for `.parse_args` and `.validate_name`. A few minor refactors were necessary to make sure error messages present only the original argument path, not the expanded path. This also leads to simplified code.

I'm not sure about stripping `crystal-`, `-crystal` and `.cr` as suggested in https://github.com/crystal-lang/crystal/issues/8682#issuecomment-579574452 This could be implemented, but I'd rather have the user specify the name explicitly. Also the shard spec does not completely prohibit `crystal` in the name. It's only a recommendation. So I'm not even sure if `crystal init foo-crystal` should fail at all. Maybe print a warning, though?

`foo.cr` is technically an invalid name, because the spec prohibits dots in the name.
But I would actually suggest to change that and allow dots. I'm not sure whether there is an explicit reason for that. /cc @ysbaddaden

### DIFF
--- a/src/compiler/crystal/tools/init.cr
+++ b/src/compiler/crystal/tools/init.cr
@@ -44,16 +44,17 @@ module Crystal
 
       OptionParser.parse(args) do |opts|
         opts.banner = <<-USAGE
-          Usage: crystal init TYPE NAME [DIR]
+          Usage: crystal init TYPE (DIR | NAME DIR)
+
+          Initializes a project folder as a git repository and default folder
+          structure for Crystal projects.
 
           TYPE is one of:
-              lib                      creates library skeleton
-              app                      creates application skeleton
+              lib                      Creates a library skeleton
+              app                      Creates an application skeleton
 
-          NAME - name of project to be generated,
-                 eg: example
-          DIR  - directory where project will be generated,
-                 default: NAME, eg: ./custom/path/example
+          DIR  - directory where project will be generated
+          NAME - name of project to be generated, default: basename of DIR
 
           USAGE
 
@@ -72,8 +73,16 @@ module Crystal
 
         opts.unknown_args do |args, after_dash|
           config.skeleton_type = fetch_skeleton_type(opts, args)
-          config.name = fetch_name(opts, args)
-          config.dir = fetch_directory(args, config.name)
+          dir = fetch_required_parameter(opts, args, "DIR")
+          if args.empty?
+            # crystal init TYPE DIR
+            config.dir = dir
+            config.name = config.expanded_dir.basename
+          else
+            # crystal init TYPE NAME DIR
+            config.name = dir
+            config.dir = args.shift
+          end
         end
       end
 
@@ -81,10 +90,7 @@ module Crystal
         raise Error.new "Cannot use --force and --skip-existing together"
       end
 
-      if config.name == "."
-        config.dir = Dir.current
-        config.name = File.basename config.dir
-      end
+      validate_name(config.name)
 
       config.author = fetch_author
       config.email = fetch_email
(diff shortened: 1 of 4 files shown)


src/compiler/crystal/tools/init.cr
```diff
@@ -145,6 +137,20 @@ module Crystal
       args.shift
     end
 
+    def self.validate_name(name)
+      case
```
Blacksmoke16: How does this method work without `case name`?

bcardiff:
If `crystal init foo-crystal` fails without suggesting the user to add an explicit name of the project, then it will not happen. The user will change the command to `crystal init foo`

Between failing and doing a best attempt to provide a valid name, I would do the later.

The use case to support `crystal init foo-crystal` is to have a repo org/foo-crystal in github but the named of the shard been foo. The repo url **should** be suggested by the directory and not by the shard name IMO. The means that `src/compiler/crystal/tools/init/template/readme.md.ecr` should be updated also in this PR.

src/compiler/crystal/tools/init.cr
straight-shoota: What do you mean?

Blacksmoke16: I didn't realize you could omit the case's value.  I.e. you have `case` versus `case name`.

straight-shoota:
> Between failing and doing a best attempt to provide a valid name, I would do the later.

Technically, a name containing `crystal` is still valid. It's not recommended but valid.
So `crystal init foo-crystal` shouldn't fail. The question is, should it take the name verbatim or transform it to `foo`? The latter adds more complexity, but it's probably fine.

src/compiler/crystal/tools/init.cr
```diff
@@ -168,10 +174,15 @@ module Crystal
         @skip_existing = false
       )
       end
+
+      getter expanded_dir : ::Path { ::Path.new(dir).expand(home: true) }
+
+      getter github_repo : String { "#{github_name}/#{expanded_dir.basename}" }
```
Sija: Since the computation here depends on the other properties I'd keep those as a regular methods with no memoization.

src/compiler/crystal/tools/init.cr
```diff
@@ -145,6 +137,20 @@ module Crystal
       args.shift
     end
 
+    def self.validate_name(name)
+      case
+      when name.blank?                       then raise Error.new("NAME must not be empty")
+      when name.size > 50                    then raise Error.new("NAME must not be longer than 50 characters")
+      when name.each_char.any?(&.uppercase?) then raise Error.new("NAME should be all lower cased")
+      when name.index("crystal")             then raise Error.new("NAME should not contain 'crystal'")
```
RX14: This must go, there's no reason to error on this case.

asterite: I agree: https://github.com/mverzilli/crystalla

(not active, but still...)

straight-shoota: Totally agree. I was just waiting on responses to my question in https://github.com/crystal-lang/crystal/pull/8737#issuecomment-583173458

src/compiler/crystal/tools/init.cr
straight-shoota: Technically you're completely right. But `Config` properties are populated once and then never changed. This is not expressed in an explicit contract, but in the limited scope of its usage, I believe this is fine.
IMO it doesn't really matter one way or the other.

straight-shoota:
`crystal init` should fail if name contains a period (see https://github.com/crystal-lang/shards/pull/321).

This is ready merge, but I'd like to get another approval.

PR #8739, opened 2020-02-04
Add documentation to Fiber

straight-shoota:


asterite:
This is great!

That said, in Go you can only spawn fibers but you can't reference them. I think that's a better design because it limits what you can do. There's really no need to call name on fiber, resume, or anything else. The more abstracted and hidden this is, the easier it is to change things without breaking user code.

Ideally, I imagine Fiber being inside the Crystal namespace.

Did we discuss all of this?

straight-shoota:
We could consider hiding `Fiber` as internal API. This doesn't affect this PR, though. Documentation is good nevertheless. And this documentation would need to go some place else (for example on `spawn`).

Blacksmoke16:
> you can't reference them

This would be problematic for some Athena stuff of mine.  I'm currently storing the DI container on the current fiber for each request.  The main reasoning is the Fiber is the only construct that I can think of that meets the requirements:
1. Be a unique instance on a per request basis
2. Accessible anywhere in the application, versus being tied to a `HTTP::Server::Context` for example

straight-shoota:
Please let's focus on documentation here . `Fiber`s  status can be discussed in a dedicated issue. Thanks.

PR #8743, opened 2020-02-04
Handle exceptions raised at `__crystal_sigfault_handler`

waj:
Similar to #8735 but for exceptions being raised within the function that capture fault signals.

Sometimes the memory is corrupted before the SEGV signal and this function fails while checking for a stack overflow. An exception is raised and the tipical END_OF_STACK message is displayed instead.


RX14:
What fails? `Fiber.current`?

I'm surprised you get an exception instead of a segfault if there's corruption?

PR #8745, opened 2020-02-04
Guard against non-structure values in {JSON,YAML}::Any#dig?

### LINKED ISSUE #8740: Any#dig raises on nil value
When `JSON::Any#dig?` or `YAML::Any#dig` try to dig into a raw value that's not a `Hash`, they raise an exception:

```cr
require "json"
JSON::Any.new(nil).dig?("foo") # Error: Expected Hash for #[]?(key : String), not Nil
```
```cr
require "yaml"
YAML::Any.new(nil).dig?("foo") # Error: Expected Array or Hash, not Nil
```

That's unexpected behaviour, because the documentation states it return `nil` if the digging key is not found.
It seems the implementation was intended to handle this case properly with checking `value.responds_to?(:dig?)`, but since `typeof(value)` is `Any`, `#dig?` is obviously always defined. It should probably be something like `value.raw.responds_to?(:[]?)` instead.

Sija:
Fixes #8740

### DIFF
--- a/src/json/any.cr
+++ b/src/json/any.cr
@@ -120,14 +120,17 @@ struct JSON::Any
   # Traverses the depth of a structure and returns the value.
   # Returns `nil` if not found.
   def dig?(key : String | Int, *subkeys)
-    if (value = self[key]?) && value.responds_to?(:dig?)
+    if value = self[key]?
       value.dig?(*subkeys)
     end
   end
 
   # :nodoc:
   def dig?(key : String | Int)
-    self[key]?
+    case @raw
+    when Hash, Array
+      self[key]?
+    end
   end
 
   # Traverses the depth of a structure and returns the value, otherwise raises.
--- a/spec/std/json/any_spec.cr
+++ b/spec/std/json/any_spec.cr
@@ -108,6 +108,11 @@ describe JSON::Any do
       obj.dig?("bar", "baz").should be_nil
       obj.dig?("").should be_nil
     end
+
+    it "returns nil for non-Hash/Array intermediary values" do
+      JSON::Any.new(nil).dig?("foo").should be_nil
+      JSON::Any.new(0.0).dig?("foo").should be_nil
+    end
   end
 
   describe "dig" do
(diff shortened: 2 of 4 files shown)


straight-shoota:
I suppose the `value.responds_to?(:dig?)` mentioned in the issue could be removed?

Sija:
@straight-shoota Sorry, I don't get what's with the `value.responds_to?(:dig?)` thing...

Sija:
> I suppose the `value.responds_to?(:dig?)` mentioned in the issue could be removed?

Indeed! I've force-pushed the suggested changes.

src/json/any.cr
```diff
@@ -120,14 +120,15 @@ struct JSON::Any
   # Traverses the depth of a structure and returns the value.
   # Returns `nil` if not found.
   def dig?(key : String | Int, *subkeys)
-    if (value = self[key]?) && value.responds_to?(:dig?)
-      value.dig?(*subkeys)
-    end
+    self[key]?.try(&.dig?(*subkeys))
```
straight-shoota: I'd prefer to keep `if value = self[key]? ...`  even without the second condition. Thats more readable than the proposed change.

Sija:
GTG?

src/json/any.cr
```diff
@@ -120,14 +120,17 @@ struct JSON::Any
   # Traverses the depth of a structure and returns the value.
   # Returns `nil` if not found.
   def dig?(key : String | Int, *subkeys)
-    if (value = self[key]?) && value.responds_to?(:dig?)
+    if value = self[key]?
       value.dig?(*subkeys)
     end
```
asterite: Just a nitpick, totally optional, but I think this can be written like:

```crystal
self[key]?.try &.dig?(*subkeys)
```

Same thing in `YAML::Any`.

straight-shoota: https://github.com/crystal-lang/crystal/pull/8745#discussion_r374990371

manveru:
Thanks a lot!

PR #8746, opened 2020-02-05
Codegen: Use llvm memset and memcpy matching arch

bcardiff:
On 64bits llvm.memset.p0i8.i64 will be used
On 32bits llvm.memset.p0i8.i32 will be used
Same for memcpy

There is no spec in llvm that guarantees the 32bits version works in 64 bits. We should always use the right one.

Some initial regression tests using FileCheck are added that are not automated in the CI but helps as a starting point.

---

Fixes the memory corruption in --release mode for #8387


RX14:
Can/should we automate these `FileCheck` tests?

bcardiff:
Eventually yes. They are usually run with `lit` tool. But I haven’t invest time to learn about it and setup the whole environment. I think FileCheck is not even present in build images.

RX14:
I think if possible it would be nice to build our own DSL for specifying ir output, but it might not be worth it to loose the dependency.

wontruefree:
It might be worth an RFC.  I would be interesting in hearing what you think the scope, benifits and costs are.

PR #8747, opened 2020-02-05
Use a high-res logo in README.md

waj:
Use this new logo for the README file:

crystal

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ---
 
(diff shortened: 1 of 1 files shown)


Sija:
Would it be possible to use really high-res version, i.e. SVG?

waj:
> Would it be possible to use really high-res version, i.e. SVG?

That would be infinite-res 🙂 
I tried but it doesn't play nice with the text with that font

straight-shoota:
In SVG you'd need to convert the text to a path. Then this should work like a charm in infinite resolution ;)

bcardiff:
This got reverted for now because it didn't play nice with the doc generator tool.

straight-shoota:
Embedded SVG should work...

PR #8750, opened 2020-02-06
Close underlying WebSocket connection when connection is closed

### LINKED ISSUE #7199: WebSocket protocol don't close the underlying IO

Sija:
Reload of #8648
Fixes #7199

### DIFF
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -106,12 +106,11 @@ class HTTP::WebSocket
         info = @ws.receive(@buffer)
       rescue
         @on_close.try &.call("")
-        @ws.close
         break
       end
 
       case info.opcode
-      when Protocol::Opcode::PING
+      when .ping?
         @current_message.write @buffer[0, info.size]
         if info.final
           message = @current_message.to_s
@@ -119,25 +118,25 @@ class HTTP::WebSocket
           pong(message) unless closed?
           @current_message.clear
         end
-      when Protocol::Opcode::PONG
+      when .pong?
         @current_message.write @buffer[0, info.size]
         if info.final
           @on_pong.try &.call(@current_message.to_s)
           @current_message.clear
         end
-      when Protocol::Opcode::TEXT
+      when .text?
         @current_message.write @buffer[0, info.size]
         if info.final
           @on_message.try &.call(@current_message.to_s)
           @current_message.clear
         end
-      when Protocol::Opcode::BINARY
+      when .binary?
         @current_message.write @buffer[0, info.size]
         if info.final
           @on_binary.try &.call(@current_message.to_slice)
           @current_message.clear
         end
-      when Protocol::Opcode::CLOSE
+      when .close?
         @current_message.write @buffer[0, info.size]
         if info.final
           message = @current_message.to_s
--- a/src/http/web_socket.cr
+++ b/src/http/web_socket.cr
@@ -2,8 +2,8 @@ class HTTP::WebSocket
   getter? closed = false
 
   # :nodoc:
-  def initialize(io : IO)
-    initialize(Protocol.new(io))
+  def initialize(io : IO, sync_close = true)
+    initialize(Protocol.new(io, sync_close: sync_close))
   end
 
   # :nodoc:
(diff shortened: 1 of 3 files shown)


src/http/web_socket/protocol.cr
```diff
@@ -58,15 +58,8 @@ class HTTP::WebSocket::Protocol
...
-      if @pos == @buffer.size
-        flush(final: false)
-      end
-
-      if count < slice.size
-        write(slice + count)
-      end
-
-      nil
+      flush(final: false) if @pos == @buffer.size
```
straight-shoota: I'd rather not have these changes. They're purely stylistic and there's not real benefit in either variant. Some prefer one, some prefer the other. Changing back and force for the sake of it, is just noise and avoidable.

src/http/web_socket/protocol.cr
```diff
@@ -121,9 +114,10 @@ class HTTP::WebSocket::Protocol
     @io.write_byte(flags.value | opcode.value)
 
     mask = @masked ? MASK_BIT : 0_u8
-    if size <= 125
+    case size
+    when .<= 125
```
straight-shoota: This doesn't seem like an improvement to me either.

src/http/web_socket/protocol.cr
```diff
@@ -164,14 +158,16 @@ class HTTP::WebSocket::Protocol
...
     raw_opcode = @header[0] & 0x0f_u8
+
     parsed_opcode = Opcode.from_value?(raw_opcode)
     unless parsed_opcode
       raise "Invalid packet opcode: #{raw_opcode}"
     end
 
-    if parsed_opcode == Opcode::CONTINUATION
+    case
+    when parsed_opcode == Opcode::CONTINUATION
```
straight-shoota: Ditto. I don't see any sense in changing this.

src/http/web_socket.cr
```diff
@@ -97,6 +97,8 @@ class HTTP::WebSocket
   def close(message = nil)
     return if closed?
     @closed = true
+
+    @on_close.try &.call(message || "")
```
RX14: This is a change of behaviour, previously `@on_close` would only fire when a websocket close frame was sent from the remote side. I'm not sure this is good change.

RX14:
@Sija please get out of the habit of tacking on unrelated changes to your PRs. They make your PRs harder to review and less likely to be excepted.

src/http/web_socket.cr
Sija: That's correct. I'm not sure why wouldn't it be a good change, since you'd expect to have it called either way, IMHO it should be like that from the beginning.

RX14: > since you'd expect to have it called either way

Personally I wouldn't expect *any* callbacks to be called after calling `websocket.close`, let alone `on_close`. Not sure what others think.

straight-shoota: `close` is an explicit call from user code. There's no need for a callback to notify it was called.

Sija: I've changed it back.

src/http/web_socket.cr
```diff
@@ -110,38 +110,38 @@ class HTTP::WebSocket
...
           @current_message.clear
         end
-      when Protocol::Opcode::CLOSE
+      when .close?
         @current_message.write @buffer[0, info.size]
         if info.final
           message = @current_message.to_s
           @on_close.try &.call(message)
-          close(message) unless closed?
+          close(message.presence)
```
RX14: Why change behaviour here?

Sija: Why would we want to pass an empty string here?

RX14: Why would you convert empty strings into `nil`?

straight-shoota:
Thank you @Sija

PR #8752, opened 2020-02-06
Change CRYSTAL_PATH to allow shards to hide std-lib

bcardiff:
Ref: https://github.com/crystal-lang/crystal/pull/8364#issuecomment-545092803

The distribution-scripts counter part is at https://github.com/crystal-lang/distribution-scripts/pull/56

### DIFF
--- a/bin/crystal
+++ b/bin/crystal
@@ -138,7 +138,7 @@ SCRIPT_ROOT="$(dirname "$SCRIPT_PATH")"
 CRYSTAL_ROOT="$(dirname "$SCRIPT_ROOT")"
 CRYSTAL_DIR="$CRYSTAL_ROOT/.build"
 
-export CRYSTAL_PATH=$CRYSTAL_ROOT/src:lib
+export CRYSTAL_PATH=lib:$CRYSTAL_ROOT/src
 export CRYSTAL_HAS_WRAPPER=true
 
 if [ -x "$CRYSTAL_DIR/crystal" ]; then


jhass:
/cc @anatol for updating this in the Archlinux package.

Please CC any other package maintainers you know. Maybe we should set this in the `Makefile` and expose something like `CRYSTAL_CONFIG_STDLIB_PATH` to the package maintainers.

anatol:
Thank you for the headsup. Interesting enough Arch build does not use `lib:` path at all:

```
make release=1 \
       FLAGS="--release --no-debug" \
       CRYSTAL_PATH="$srcdir/$pkgname-$pkgver/src" \
       CRYSTAL_CONFIG_VERSION="$pkgver" \
       CRYSTAL_CONFIG_PATH="lib:/usr/lib/crystal" \
       CRYSTAL_CACHE_DIR="/tmp/crystal"
  make docs CRYSTAL_CACHE_DIR="/tmp/crystal"
```

Could you please help me to understand what needs to be done with this build script?

bcardiff:
@anatol you do use it :-) `CRYSTAL_CONFIG_PATH` is the default value the compiler will use for `CRYSTAL_PATH` when it is not defined.
And you are doing the right thing: `lib` comes before `/usr/lib/crystal` where the std-lib lives.

In osx binary packages we were using that since the begining of times. But on linux packages and homebrew formula we are not.

anatol:
Tanks for the clarification @bcardiff. One more question - is `CRYSTAL_PATH` option needed in the build script above?

bcardiff:
probably, because you want to use the crystal binary with the updated std-lib to build the new version of the compiler. But it depends on the whole context of the script. 

if you use the `bin/crystal` wrapper of this repo that should already happen. that script uses the crystal binary in path when no `.build/crystal` exists but it overrides the std-lib to use the working copy one.

jhass:
Yes it's needed since the archlinux package ships no wrapper (and I don't see any reason why it should) and installs the stdlib to /usr/lib/crystal.

RX14:
I think `bin/crystal` sets `CRYSTAL_PATH` to the same as it's manually set in the `PKGBUILD`...

PR #8756, opened 2020-02-07
Keep expanding home in CRYSTAL_CACHE_DIR

jhass:
Fixes a small regression from #7903

Somebody with more time than me to spend on this might want to go through and evaluate all `File.expand_path` uses in the compiler  for whether it's safe and beneficial to expand home.

### DIFF
--- a/src/compiler/crystal/codegen/cache_dir.cr
+++ b/src/compiler/crystal/codegen/cache_dir.cr
@@ -73,7 +73,7 @@ module Crystal
       ]
       candidates = candidates
         .compact
-        .map { |file| File.expand_path(file) }
+        .map { |file| File.expand_path(file, home: true) }
         .uniq
 
       # Return the first one for which we could create a directory


oprypin:
Why do you think this should be expanded? I think it would be unusual to do.

jhass:
Well I noticed it because I've set mine to `CRYSTAL_CACHE_DIR=~/.crystal`

bcardiff:
I would expect that `Dir.mkdir_p`, `Dir.exists?` and others would resolve `~` and not requiring the user to expand it. Am I wrong? Currently, a directory named `~` is created in the current location.

asterite:
> I would expect that Dir.mkdir_p, Dir.exists? and others would resolve ~ and not requiring the user to expand it. Am I wrong?

At least Ruby doesn't expand `~` in those calls.

I believe Ruby and other language never expand `~` unless explicitly asked (in Ruby you have to do `File.expand_path`). I think we should do the same in Crystal.

The problem is that we recently changed `File.expand_path` to not expand `~` by default because it's considered dangerous.

jhass:
And that was a good change! Expansion of  `~`  should  always  be  explicit.  We just have to revisit in which places inside the compiler it makes sense to do it. I  believe this is one of them. (`CRYSTAL_PATH`  might be another obvious candidate).

straight-shoota:
@bcardiff If we would expand home in all these methods, there would have been no point in disabling it by default in `Path#expand`. Also they couldn't handle paths with a literal `$` which is a perfectly valid directory name in most file systems.

We just need to explicitly expand paths where necessary, for example when they come from environment variables or similarly trusted user input. For command line applications, probably every user input should be considered.

bcardiff:
Ok. I thought the intention was to manipulate path while preserving ~, but once we actually use the string to access a path the ~ should be solved always.

oprypin:
I'm pretty sure that very few tools expand `*_PATH` variables like that. To me this is just messing with the user's intention and has the potential of great confusion. 

Namely, there's no way to specify a directory that starts with `~`. Not only that, paths that happen to be that (in some dynamic context) will break. Sure, this will never *reasonably* happen. But security issues don't have to be reasonable.

@jhass's justification is that he relies on this behavior. But worth pointing out that
`CRYSTAL_CACHE_DIR=~/.crystal` will actually be expanded by the shell and be passed as a full path to Crystal.
The other part of the justification is that it *used to be* expanded. But that's circular reasoning; it simply used to be wrong.

oprypin:
> I would expect that Dir.mkdir_p, Dir.exists? and others would resolve ~ and not requiring the user to expand it. Am I wrong?

Yes, sorry.

> What's the harm in doing it?

All the same harm that #7903 is trying to prevent in the first place.

straight-shoota:
>  But worth pointing out that `CRYSTAL_CACHE_DIR=~/.crystal` will actually be expanded by the shell and be passed as a full path to Crystal.

That's a fair point. It's sometimes hard to separate what kind of behaviour come from the shell and what from the actual program. 

Maybe it's best to leave this as is and expect home paths to be fully expanded by the shell. I'm not sure. Might be worth looking closer at how other applications behave.

@jhass What's your specific use case? How do you configure the environment variable without the path being expanded by the shell?

oprypin:
As an example, `CRYSTAL_CACHE_DIR="~/.crystal"` would not be expanded.
Then you could argue that it's nice for Crystal to guess the intent, but it's not a safe guess.

Users deal with this mistake for all other tools (subsequently learning something about the shell) and I'm sure they can deal with it for Crystal.

jhass:
Well looks like the majority disagrees here, fair.

PR #8758, opened 2020-02-07
Compiler: fix type of vars after begin/rescue if all rescue unreachable

### LINKED ISSUE #8751: Weird behaviour when referencing a var within rescue block
```cr
loop do
  begin
    foo = "foo"
    raise "foo"
  rescue
    foo # Removing this line makes it compile without a problem
    break
  end
 
  foo.size
end
```

Results in:

```
error in line 10
Error: undefined method 'size' for Nil (compile-time type is (String | Nil))
```

https://carc.in/#/r/8ipq

asterite:
Fixes #8751

The type of the vars after an exception handler where all rescue/else clauses end up being unreachable (break or raise, essentially NoReturn) was taken from an incorrect computation. This fixes it.

The diff is essentially replacing:

```crystal
@vars = exception_handler_vars
```

with 

```crystal
@vars = after_exception_handler_vars
```

and then moving the logic that begins with the "However, ..." comment to inside an `if` branch because it doesn't need to be applied if all rescue/else are unreachable.

### DIFF
--- a/spec/compiler/semantic/exception_spec.cr
+++ b/spec/compiler/semantic/exception_spec.cr
@@ -670,4 +670,20 @@ describe "Semantic: exception" do
       foo
     )) { nilable int32 }
   end
+
+  it "gets a non-nilable type if all rescue are unreachable (#8751)" do
+    semantic(%(
+      while true
+        begin
+          foo = 1
+          break
+        rescue
+          foo
+          break
+        end
+
+        foo &+ 2
+      end
+      ))
+  end
 end
--- a/spec/compiler/semantic/ssa_spec.cr
+++ b/spec/compiler/semantic/ssa_spec.cr
@@ -267,7 +267,7 @@ describe "Semantic: ssa" do
         LibC.exit
       end
       a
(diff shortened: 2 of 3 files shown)


Sija:
There's one spec left to fix:

```sh
Failures:

  1) Semantic: ssa types a var after begin rescue with no-return in rescue
     Failure/Error: result.type.should eq(expected_type)
       Expected: (Char | Int32 | String)
            got: String
     # spec/spec_helper.cr:43

Finished in 28:26 minutes
9122 examples, 1 failures, 0 errors, 9 pending

Failed examples:

crystal spec spec/compiler/semantic/ssa_spec.cr:256 # Semantic: ssa types a var after begin rescue with no-return in rescue
```

asterite:
@Sija Thanks for pointing out the error!

Looking at the spec, I would expect `a` to be a String after the begin/rescue because only two things can happen:
- an exception is not raised: so `a` will definitely have "hello" in it and be a string
- an exception is raised: the `rescue` is triggered and the program exists, so the type of `a` doesn't matter anymore

Which makes me believe the changes in this PR are good 🎉

PR #8759, opened 2020-02-07
HTTP::Cookie: parse max-age as Int64, not Int32

### LINKED ISSUE #8744: HTTP::Client::Response.cookies Invalid Int32 (ArgumentError)
# Summary
* Problems caused by bad cookie parsing implementation.
  * Maybe it should use UInt64 (`to_u64`)? Or something.
## Sample

```crystal
require "http"

payload = %(HTTP/1.1 200 OK
Date: Tue, 04 Feb 2020 19:56:19 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Server: openresty
X-Powered-By: HHVM/3.7.3-dev
Vary: Accept-Encoding
Set-Cookie: h5_user_id=H5UserId_1584511; expires=Thu, 11-Jan-2120 19:56:19 GMT; Max-Age=3153600000; path=/; domain=k.e
X-Location: /

)

response = HTTP::Client::Response.from_io IO::Memory.new payload
response.cookies
```

## Verbose

```
Unhandled exception: Invalid Int32: 3153600000 (ArgumentError)
  from unwind
  from initialize
  from new
  from raise
  from to_i32
  from to_i
  from parse_set_cookie
  from fill_from_headers
  from from_headers
  from cookies
  from __crystal_main
  from main_user_code
  from main
  from main
```

asterite:
Fixes #8744

I considered parsing it as UInt64 too, but if max-age is `Int64::MAX` it already overflows when trying to create a time instance, so Int64 is probably fine.

(or put another way, in the rare case it breaks with a much bigger number let's tackle that later)

### DIFF
--- a/spec/std/http/cookie_spec.cr
+++ b/spec/std/http/cookie_spec.cr
@@ -203,6 +203,13 @@ module HTTP
         delta.should be > 0.seconds
         delta.should be < 1.seconds
       end
+
+      it "parses large max-age (#8744)" do
+        cookie = parse_set_cookie("a=1; max-age=3153600000")
+        delta = cookie.expires.not_nil! - Time.utc
+        delta.should be > (3153600000 - 1).seconds
+        delta.should be < (3153600000 + 1).seconds
+      end
     end
 
     describe "expired?" do
--- a/spec/std/http/cookie_spec.cr
+++ b/spec/std/http/cookie_spec.cr
@@ -195,20 +195,17 @@ module HTTP
       it "parse max-age as seconds from current time" do
         cookie = parse_set_cookie("a=1; max-age=10")
         delta = cookie.expires.not_nil! - Time.utc
-        delta.should be > 9.seconds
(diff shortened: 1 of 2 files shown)


spec/std/http/cookie_spec.cr
```diff
@@ -203,6 +203,13 @@ module HTTP
         delta.should be > 0.seconds
         delta.should be < 1.seconds
       end
+
+      it "parses large max-age (#8744)" do
+        cookie = parse_set_cookie("a=1; max-age=3153600000")
+        delta = cookie.expires.not_nil! - Time.utc
+        delta.should be > (3153600000 - 1).seconds
```
straight-shoota: Why not use `be_close` expectation?

asterite: I copied the code from the previous spec without too much thought.

asterite: I changed it to use `be_close`

PR #8761, opened 2020-02-07
Rewrite block buffered channel spec

bcardiff:
Previous spec was relying on on the receiving fiber only running after the main fiber is blocked. This might not hold in multi-thread, and yields to sporadic failures.


straight-shoota:
Thanks @bcardiff

PR #8762, opened 2020-02-07
Compiler: propagate visibility to macro expansions in all cases

### LINKED ISSUE #8715: `private record` is allowed but doesn't make the struct private
I defined a `private record` in one of my classes (reasoning that the `private` would apply to the generated `struct`) and was surprised to see it showing up in the Crystal-generated docs. Looking into it, the "private" doesn't actually do anything, despite being allowed:

```crystal
# Crystal 0.32.1

class Outer
  private record Inner, foo : String
end

Outer::Inner.new("bar") # this works
```

I would have expected this to either

* not compile, or
* generate a `private struct`

asterite:
Fixes #8715

### DIFF
--- a/spec/compiler/semantic/private_spec.cr
+++ b/spec/compiler/semantic/private_spec.cr
@@ -447,4 +447,76 @@ describe "Semantic: private" do
       compiler.compile sources, "output"
     end
   end
+
+  it "doesn't find private class defined through macro (#8715)" do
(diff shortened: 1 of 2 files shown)


asterite:
Feel free to merge this and any of my PRs. I won't because I don't know if these should go into 0.33.0 or where.

PR #8772, opened 2020-02-09
Remove obsolete TODOs

Sija:
SSIA


Sija:
> I would prefer to keep the same style in the case statements. Full expression over the `when .< bound`

@bcardiff Addressed.

straight-shoota:
Please leave the last commit off. It's not much of an improvement and unrelated to this change.

Sija:
Seems like ready to merge?

straight-shoota:
Thanks @Sija

PR #8708, opened 2020-01-21
[CI] Build Alpine Docker images

straight-shoota:
Incorporates the changes from crystal-lang/distribution-scripts#29 into the CI workflow

PR #8779, opened 2020-02-10
Update distribution-scripts

bcardiff:
Bump distribution-scripts to include https://github.com/crystal-lang/distribution-scripts/pull/57 and https://github.com/crystal-lang/distribution-scripts/pull/56 .

I also noticed that #8708 was merged pointing to a commit that is not in master. So this PR also fixes that glitch.


straight-shoota:
Oh, I'm sure I've had the fix for the bad commit hash from #8707 in #7420 but it seems it didn't make it in the end. Perhaps because we changed to the regular docker image.

PR #8781, opened 2020-02-11
Remove redundant case-when statement

rhysd:
`peek_next_char` at the removed `case` statement always returns `'_'` so it always reaches `else` clause and other `when` clauses are unreachable.


src/compiler/crystal/syntax/lexer.cr
```diff
@@ -1662,26 +1662,7 @@ module Crystal
         consume_uint_suffix
         set_token_raw_from_start(start)
       when '_'
```
rhysd: By this `when` clause, `peek_next_char` is always `'_'`

PR #8782, opened 2020-02-11
Drop uppercase F32 and F64 float number suffixes

rhysd:
With Crystal 0.32.1,

```crystal
# OK
0f32

# OK
1f32

# ERROR
0F32

# OK
1F32
```

~This PR fixes `0F32` not to cause an error.~ EDIT: We decided to drop uppercase suffixes. Please see following discussion.

### DIFF
--- a/spec/compiler/lexer/lexer_spec.cr
+++ b/spec/compiler/lexer/lexer_spec.cr
@@ -313,6 +313,14 @@ describe "Lexer" do
   assert_syntax_error "00", "octal constants should be prefixed with 0o"
   assert_syntax_error "01_i64", "octal constants should be prefixed with 0o"
 
+  assert_syntax_error "4f33", "invalid float suffix"
+  assert_syntax_error "4f65", "invalid float suffix"
+  assert_syntax_error "4f22", "invalid float suffix"
+  # Tests for #8782
+  assert_syntax_error "4F32", "unexpected token: F32"
+  assert_syntax_error "4F64", "unexpected token: F64"
+  assert_syntax_error "0F32", "unexpected token: F32"
+
   it "lexes not instance var" do
     lexer = Lexer.new "!@foo"
     token = lexer.next_token
--- a/src/compiler/crystal/syntax/lexer.cr
+++ b/src/compiler/crystal/syntax/lexer.cr
@@ -1434,7 +1434,7 @@ module Crystal
             end
           end
 
-          if current_char == 'f' || current_char == 'F'
+          if current_char == 'f'
             suffix_size = consume_float_suffix
           else
             @token.number_kind = :f64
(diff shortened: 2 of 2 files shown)


rhysd:
Hmm, I'm not able to locate which test case fails on CircleCI.

https://circleci.com/gh/crystal-lang/crystal/38220?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
Error while trying to determine if a stack overflow has occurred. Probable memory corrpution

Invalid memory access (signal 11) at address 0x40

[0x5621f5b60026] *CallStack::print_backtrace:Int32 +118

[0x5621f55889cc] __crystal_sigfault_handler +348

[0x5621f72a9fe4] sigfault_handler +40

[0x7f349bd970d4] ???
```

straight-shoota:
As far as I know, this is intended. Type suffixes use only lower case characters. That also affects all similar identifiers. What makes you think this needs to support upper case?

test_alpine failing is not related to this PR, you can ignore it.

asterite:
> As far as I know, this is intended. Type suffixes use only lower case characters.

It seems `2F32` and `4F64` parse well, but using capital `I` for integer does not. So I think we should actually drop the capital `F` in those cases.

rhysd:
@straight-shoota @asterite 

Actually `4.0F32` is accepted. I'm not understanding the intention, but lexer implementation explicitly accepts uppercase.

https://github.com/crystal-lang/crystal/blob/5519fbdb4533bdadedde02e87e621924815d4e1d/src/compiler/crystal/syntax/lexer.cr#L1470

So I thought 'F' should be accepted in the case of `0F32`.

I agree that uppercase suffix looks weird. There is no test for `F32` and `F64` suffixes. So I agree with dropping uppercase suffix.

But it is a breaking change. Should we do it? If we should do, I'll update this PR to drop the uppercase `F32` and `F64` suffixes. Modification would be easy.

asterite:
Yes, I think we should drop `F`, even if it's a breaking change. Probably not many use it or know about it, and fixing it is easy.

rhysd:
OK, I'll update this branch tonight.

rhysd:
@straight-shoota @asterite 

I modified this PR for dropping the uppercase suffixes. And I also added small tests for invalid float number suffixes. Would you review this?

straight-shoota:
I'm almost sure upper case characters are unlikely to be in use somewhere. But we should at least consider a deprecation path, either updating the formatter first or providing a deprecation message when an upper case character is encountered.
IMO it's not necessary in this case because it's completely undocumented and probably not used.

bcardiff:
If some shard is using `F`, it can be changed to `f` and publish a patch release if needed. That will also work in previous crystal versions.

I'm fine with merging this right away.

rhysd:
CI fails due to https://github.com/crystal-lang/crystal/pull/8782#issuecomment-584482885 but it seems not related to changes in this PR since in other PR it also fails with the same error. It's weird that #8781 does not cause the error, tho.

Issue #2713, opened 2016-06-01
Wrong output when piping to less


asterite:
Extracted from [Google Groups](https://groups.google.com/forum/#!msg/crystal-lang/2y-JTQ6a_7Q/9KuIPctxCwAJ)

A consistent way to reproduce it:

``` crystal
# foo.cr
string = <<-EOF
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
01234567890123456789012345678901234567890123456789012345678901234567890123456789
EOF
puts string
```

Then:

```
$ crystal build foo.c
$ ./foo | less
```

The issue seems to be related to using `fcntl` with stdin/stdout/stderr in a shell. I found [this](http://stackoverflow.com/questions/19485751/linux-c-why-fcntl-act-on-stdin-will-also-affect-on-stdout-and-stderr) but I can't seem to understand what's going on in this case.

miketheman:
#triage 
This is an interesting one. I am definitely able to repro off current master 9cf6e06 .

Redirecting stderr to stdout produces the expected result:

```
$ ./foo 2>&1 | less
```
(also `./foo |& less` - |& was added to Bash 4 as an abbreviation for 2>&1 )

[...]

PR #8787, opened 2020-02-11
Try to open stdio in non-blocking mode

waj:
Currently if the stdio descriptors are not a TTY they are re-opened with `blocking: true`. That makes sense for files, for example if the output is being redirected (i.e: `./foo > somefile`). But it doesn't if the input is a pipe (i.e: `cat | ./foo`).

The problem is that programs like this will block all the fibers:
```crystal
spawn do
  STDIN.gets
end
```

We used to have problems with stdio (#2713) but it was fixed by not changing the blocking mode from the original file descriptor (#6518)

This PR adds a third state to the `blocking` argument of `FileDescriptor.new`. When a `nil` is passed (the default) it detects the type of descriptor and sets the blocking mode accordingly.

Unfortunately this is quite difficult to test, so there is no specs to find regressions. I added a spec to check the STDIN opens in different modes depending on the fd type. But please, help me find if I didn't miss anything and I'm actually breaking everything again 😅

### DIFF
--- a/spec/std/io/file_descriptor_spec.cr
+++ b/spec/std/io/file_descriptor_spec.cr
@@ -2,7 +2,7 @@ require "../../spec_helper"
 
 describe IO::FileDescriptor do
   it "reopen STDIN with the right mode" do
-    code = "puts \"\#{STDIN.blocking} \#{STDIN.info.type}\""
+    code = %q(puts "#{STDIN.blocking} #{STDIN.info.type}")
     build(code) do |binpath|
       `#{binpath} < #{binpath}`.chomp.should eq("true File")
       `echo "" | #{binpath}`.chomp.should eq("false Pipe")
--- a/spec/std/io/file_descriptor_spec.cr
+++ b/spec/std/io/file_descriptor_spec.cr
@@ -1,0 +1,11 @@
+require "../../spec_helper"
+
+describe IO::FileDescriptor do
+  it "reopen STDIN with the right mode" do
+    code = "puts \"\#{STDIN.blocking} \#{STDIN.info.type}\""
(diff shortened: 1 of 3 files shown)


src/io/file_descriptor.cr
```diff
@@ -9,9 +9,19 @@ class IO::FileDescriptor < IO
   # platform-specific.
   getter fd
 
-  def initialize(@fd, blocking = false)
+  def initialize(@fd, blocking = nil)
     @closed = system_closed?
 
+    if blocking == nil
```
Sija:
```suggestion
    if blocking.nil?
```

src/io/file_descriptor.cr
```diff
@@ -9,9 +9,19 @@ class IO::FileDescriptor < IO
...
+      blocking =
+        case system_info.type
+        when File::Type::Pipe, File::Type::Socket, File::Type::CharacterDevice
```
Sija:
```suggestion
        when .pipe?, .socket?, .character_device?
```

spec/std/io/file_descriptor_spec.cr
```diff
@@ -0,0 +1,11 @@
+require "../../spec_helper"
+
+describe IO::FileDescriptor do
+  it "reopen STDIN with the right mode" do
+    code = "puts \"\#{STDIN.blocking} \#{STDIN.info.type}\""
```
Sija:
```suggestion
    code = %(puts "\#{STDIN.blocking} \#{STDIN.info.type}")
```

waj:
> Couple of pedantic suggestions

Love all the suggestions. You know the language better than me :)

PR #4876, opened 2017-08-23
Implement BigDecimal

vegai:
This PR adds a BigDecimal implementation. 

The motivation for having BigDecimals is to have an arbitrary precision exact decimal type. This is required at least in financial calculations, where the precision of Floats and is not enough and BigRationals may have other problems due to denominator potentially being something else than 10**x. At least Ruby, Python and Java contain something like this in their standard libs.

The design is mostly adapted from bigdecimal-rs written for Rust. It holds its actual value in a BigInt, and a scale (decimal point place) in UInt64. It currently contains basic + - * / arithmetics. Possible future needs: implementation of ** (power) and supporting scientific "123E45" notation.

Please critique liberally. This would be my first larger Crystal contribution.

PR #8790, opened 2020-02-12
Fix `BigDecimal#to_big_i` regression

### LINKED ISSUE #8789: Regression in `BigDecimal#to_big_i`
Regression introduced in Crystal 0.31.0 (#8120)

```cr
require "big"
 
"-100.000".to_big_d.to_big_i # => -100000
```

Before: https://carc.in/#/r/8jrq
After: https://carc.in/#/r/8jro

/cc @bcardiff

Sija:
Fixes #8789

### DIFF
--- a/spec/std/big/big_decimal_spec.cr
+++ b/spec/std/big/big_decimal_spec.cr
@@ -337,36 +337,50 @@ describe BigDecimal do
     bd2 = BigDecimal.new(-123, 5)
     bd3 = BigDecimal.new(123, 0)
     bd4 = BigDecimal.new(-123, 0)
+    bd5 = "-123.000".to_big_d
+    bd6 = "-1.1".to_big_d
 
     bd1.to_i.should eq 0
     bd2.to_i.should eq 0
     bd3.to_i.should eq 123
     bd4.to_i.should eq -123
+    bd5.to_i.should eq -123
+    bd6.to_i.should eq -1
 
     bd1.to_u.should eq 0
     bd2.to_u.should eq 0
     bd3.to_u.should eq 123
     bd4.to_u.should eq 123
+    bd5.to_u.should eq 123
+    bd6.to_u.should eq 1
 
     bd1.to_f.should eq 0.00123
     bd2.to_f.should eq -0.00123
     bd3.to_f.should eq 123.0
     bd4.to_f.should eq -123.0
+    bd5.to_f.should eq -123.0
+    bd6.to_f.should eq -1.1
 
     bd1.to_i!.should eq 0
     bd2.to_i!.should eq 0
     bd3.to_i!.should eq 123
     bd4.to_i!.should eq -123
+    bd5.to_i!.should eq -123
+    bd6.to_i!.should eq -1
 
     bd1.to_u!.should eq 0
     bd2.to_u!.should eq 0
     bd3.to_u!.should eq 123
     bd4.to_u!.should eq 123
+    bd5.to_u!.should eq 123
+    bd6.to_u!.should eq 1
 
     bd1.to_f!.should eq 0.00123
     bd2.to_f!.should eq -0.00123
     bd3.to_f!.should eq 123.0
     bd4.to_f!.should eq -123.0
+    bd5.to_f!.should eq -123.0
+    bd6.to_f!.should eq -1.1
   end
 
   it "hashes" do
@@ -435,6 +449,9 @@ describe BigDecimal do
 
     it { -2.01.to_big_d.ceil.should eq(-2) }
     it { -2.91.to_big_d.ceil.should eq(-2) }
+
+    it { "-123.000".to_big_d.ceil.value.should eq(-123) }
+    it { "-1.1".to_big_d.ceil.value.should eq(-1) }
   end
 
   describe "#floor" do
@@ -445,6 +462,9 @@ describe BigDecimal do
     it { 2.11.to_big_d.floor.should eq(2) }
     it { 2.91.to_big_d.floor.should eq(2) }
     it { -2.91.to_big_d.floor.should eq(-3) }
+
+    it { "-123.000".to_big_d.floor.value.should eq(-123) }
+    it { "-1.1".to_big_d.floor.value.should eq(-2) }
   end
 
   describe "#trunc" do
--- a/src/big/big_decimal.cr
+++ b/src/big/big_decimal.cr
@@ -260,7 +260,8 @@ struct BigDecimal < Number
     in_scale(new_scale.scale)
   end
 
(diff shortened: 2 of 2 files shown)


Sija:
Since it's a pretty major regression it would be gr8 having this merged into `0.33.0`.

spec/std/big/big_decimal_spec.cr
```diff
@@ -337,36 +337,43 @@ describe BigDecimal do
...
     bd2.to_i.should eq 0
     bd3.to_i.should eq 123
     bd4.to_i.should eq -123
+    bd5.to_i.should eq -123
 
     bd1.to_u.should eq 0
     bd2.to_u.should eq 0
     bd3.to_u.should eq 123
     bd4.to_u.should eq 123
+    bd5.to_u.should eq 123
```
Sija: @bcardiff btw, why do `Big*#to_u` methods for values < 0 are returning them as absolute instead of raising?

src/big/big_decimal.cr
```diff
@@ -331,11 +333,7 @@ struct BigDecimal < Number
 
   # Converts to `BigInt`. Truncates anything on the right side of the decimal point.
   def to_big_i
-    if self >= 0
-      self.floor.value
-    else
-      self.ceil.value
-    end
+    (self >= 0 ? floor : ceil).value
```
straight-shoota: Why this change? It seems unrelated.

Sija:
@straight-shoota Force-pushed amended commit with the fix.

src/big/big_decimal.cr
```diff
@@ -260,7 +260,7 @@ struct BigDecimal < Number
     in_scale(new_scale.scale)
   end
 
-  private def in_scale(new_scale : UInt64) : BigDecimal
+  def in_scale(new_scale : UInt64) : BigDecimal
```
straight-shoota: This method should probably not be public API but `protected` (or `:nodoc:`).

straight-shoota:
Is it possible to (also) have a spec for `#ceil` which tests the changed behaviour?

src/big/big_decimal.cr
Sija: I don't see any reason to make it `protected`, why not bring it into the public API?

Sija:
> Is it possible to (also) have a spec for `#ceil` which tests the changed behaviour?

Done.

src/big/big_decimal.cr
straight-shoota: I can't tell whether it would be functionally relevant for the public API. But there's already `#scale_to` which seems to be intended as the public facing front. 

If it goes public, it should be documented.

Sija: I've made it `:nodoc:`. `#scale_to` has a bit odd (imo) interface forcing you to pass another `BigDecimal` - which is not always wanted if you want to just scale a number at hand.

spec/std/big/big_decimal_spec.cr
```diff
@@ -435,6 +449,9 @@ describe BigDecimal do
 
     it { -2.01.to_big_d.ceil.should eq(-2) }
     it { -2.91.to_big_d.ceil.should eq(-2) }
+
+    it { "-123.000".to_big_d.ceil.should eq(-123) }
```
straight-shoota: These specs pass on `0.32.1` because comparison applies the correct scale. That's why I asked very tentatively whether there's a way to spec this because it needs to look at internal state.

The specs are probably good, still but they don't test what they're supposed to.

src/big/big_decimal.cr
straight-shoota: Feel free to open a separate issue about changing the API! But this PR is better focused on fixing the bug.

spec/std/big/big_decimal_spec.cr
Sija: You're right. I've changed it to compare the `#value` instead.

spec/std/big/big_decimal_spec.cr
straight-shoota: That seems unexpected...

bcardiff: I think is the effect of time, when `Int#to_u` used to be more like a cast, and `BigDecimal#to_big_u` simple uses an `.abs` as an approximation. Nobody submit any opinions regarding that in the original PR #4876 .

It should raise `OverflowError` for negative inputs.

Sija: Was thinking the same, although I'd leave it for another PR.

Sija:
Is this GTG?

spec/std/big/big_decimal_spec.cr
straight-shoota: For further consideration: All these specs don't really test what you would expect. Equality between different number types already works without explicit conversation. So these specs should also validate the return type.

src/big/big_decimal.cr
straight-shoota: @Sija Could you revert this change?

spec/std/big/big_decimal_spec.cr
Sija: @straight-shoota Could we leave it for another PR? I don't think we should do much more here.

src/big/big_decimal.cr
Sija: I changed to use `trunc` since IMO it should be used in the first place instead of duplicating the logic with more surface for bugs like that one.

spec/std/big/big_decimal_spec.cr
straight-shoota: Absolutely

Sija: btw, `BigInt#to_u` has the same issue:

```cr
require "big"

-123.to_big_i.to_u # => 123
```

Sija:
Merge time? :)

Sija:
Yo! Anything else to do here? 🕦

Your PR-review process guys is a real mystery to me. You let tested, approved and tagged PRs lay idly with no explanation about the next steps or anything else that's suppose to come next... Could you be please more transparent in your actions, so it's less of a guessing-game and more of a understandable process - at least to the outsiders, who do not share your private communication channels or whatever you use to discuss things like that - if you do...

straight-shoota:
Sadly, there's really no process in place.
I suppose merging this one was simply delayed to not interfere with the release of 0.33.0. Otherwise @RX14 or @bcardiff probably would've directly merged it.
Sometimes merging approved PRs just gets lost and forgotten. It can be really hard to keep on top of all the things going on. If there's a PR that looks like it could be merged, please just leave a message to bump it.

straight-shoota:
Thanks @Sija and sorry for the delay!

Issue #4754, opened 2017-07-28
Add metadata to docs output


straight-shoota:
Currently, the API docs generated by `crystal doc` contain only minimal information about *what* it actually documents: apart from the homepage with the obvious `README`, all other pages only show the repository name in the page title - but this is unreliable because the docs generator takes the first github or gitlab remote URL which might point to an unrelated mirror or not be present at all. And such a HTML page title might not always be visible to the reader. Other important information like the API version needs to be inferred from context, e.g. an URL like `crystal-lang.org/api/0.23.0/`. Such context cannot necessarily be assumed to be available and/or easily accessible to the reader.

I suggest that every docs page should have at least a reference to the name and version of the documented API and link to an authoritative website. Preferably in a spot with high visibility like the left top corner. Additional information like author, license etc. would also be welcome, but does not need to be displayed as prominently.

A great source for this data would be the `shard.yml` file available in almost all crystal packages. The compiler `deps` tool already directly integrates the `shards` binary which is distributed as part of the crystal package. To read metadata from the shard file however would require to include the [shard spec parser](https://github.com/crystal-lang/shards/blob/master/src/spec.cr) into the standard library.

It should also be possible to provide the metadata as command line arguments to the docs generator, if there is no `shards.yml` file or to explicitly overwrite specific values. The stdlib (which does not have a `shards.yml` file) could be handled as a special case with hardcoded metadata.

bew:
is this an error @straight-shoota ?

straight-shoota:
I'd like to proceed with this. Here is a more detailed plan on how this can be implemented.

For a minimum, I think the following properties should be available to the docs generator. This meta data can be published as a header and also be used in other places (for example meta tags, page title).

* `name`: shard name (or equivalent)
* `version`: release version. If not on a release (e.g. `master`), we should consider whether it should show the actual commit data or just a label (=`master`).

Besides that bare minimum, the following options might also be useful:

* `homepage_url`: Link to the homepage of the shard. This might not strictly be necessary because it can (and should) be included in the `README`.
* `source_url_pattern`: The patterns for source links are currently automatically populated from the output of `git remote` and only supports public hosting on github.com and gitlab.com. It would be great to provide command line arguments to enable this feature regardless of the current workdir situation and supporting custom hosting providers. Potentially, this should also allow to specify mappings for different paths (see https://github.com/crystal-lang/crystal/issues/8550#issuecomment-583638640).

[...]

PR #8792, opened 2020-02-12
Add project name and version to API docs

straight-shoota:
MVP implementation of #4754

This PR adds a new section in the API docs sidebar which contains the name and version of the project.

image

The values can be set using the new `--project-name` and `--project-version` CLI arguments.

**name**
When `--project-name` is missing the doc generator currently errors. This should be changed to provide a good default, so that `crystal docs` continues to just work and not require any configuration for most shards.
The optimal solution would be to use the `name` property from `shard.yml`. Properly reading that file depends on `libyaml` which is not an option. Alternatively we could use a naive text-based extraction (such as grep for `/^name:\s+(.*)/`) or call the `shards` binary to get the information. The latter seems like a good solution to me, but needs to be implemented in `shards` first (`shards info --name`, cf crystal-lang/shards#86).

Another alternative could be using the repository name (which has so far been used as the title) or directory name. Both are not ideal though, because shard names sometimes differ from repo names and we really want to show the shard name.

**version**
`version` currently defaults to `master` when `--project-version` is missing. IMO we could leave it like that because that's what you usually have. When building docs for a specific commit, it should be easy to notice and fix (by supplying `--project-version` explicitly).
An alternative would be reading the tag from git. Using the last tag from the commit history (i.e. `git describe --abbrev=0 --tags`) doesn't take into account that the repo might have advanced since then. Only when directly at a tagged commit (i.e. `git tag --points-to HEAD`) it would work reliably (ideally workdir and index need to be clean, too).
This is essentially the reason why we have `src/VERSION` in the crystal repository and it gets updated on every tag for the plain tag and afterwards `-dev` is added.
We could try to do this automatically, first checking for `git tag --points-to HEAD` and if there's no result use `git describe --abbrev=0 --tags` with `-dev` appended.
I'm not sure if there would be any issues with that? We didn't choose this approach for the compiler because it's supposed to work when distributed as a tarball without git.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -73,10 +73,6 @@ prepare_system() {
 }
 
 build() {
-  # Set git config for doc generator specs
-  with_build_env 'git config --global user.email "you@example.com"'
-  with_build_env 'git config --global user.name "Your Name"'
-
   with_build_env 'make std_spec clean threads=1 junit_output=.junit/std_spec.xml'
 
   case $ARCH in
--- a/bin/ci
+++ b/bin/ci
@@ -73,6 +73,10 @@ prepare_system() {
 }
 
 build() {
+  # Set git config for doc generator specs
+  with_build_env 'git config --global user.email "you@example.com"'
+  with_build_env 'git config --global user.name "Your Name"'
+
   with_build_env 'make std_spec clean threads=1 junit_output=.junit/std_spec.xml'
 
   case $ARCH in
--- a/spec/compiler/crystal/tools/doc/project_info_spec.cr
+++ b/spec/compiler/crystal/tools/doc/project_info_spec.cr
@@ -51,6 +51,27 @@ describe Crystal::Doc::ProjectInfo do
         ProjectInfo.new_with_defaults(nil, "1.1") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "1.1")
         ProjectInfo.new_with_defaults("bar", "2.0") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("bar", "2.0")
       end
+
+      it "git non-tagged commit" do
+        `git init`
+        `git add shard.yml`
+        `git commit -m 'Initial commit' --no-gpg-sign`
+
+        ProjectInfo.new_with_defaults(nil, nil) { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "master")
+        ProjectInfo.new_with_defaults(nil, "1.1") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "1.1")
+        ProjectInfo.new_with_defaults("bar", "2.0") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("bar", "2.0")
+      end
+
+      it "git non-tagged commit dirty" do
+        `git init`
+        `git add shard.yml`
+        `git commit -m 'Initial commit' --no-gpg-sign`
+        File.write("foo.txt", "bar")
+
+        ProjectInfo.new_with_defaults(nil, nil) { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "master-dev")
+        ProjectInfo.new_with_defaults(nil, "1.1") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "1.1")
+        ProjectInfo.new_with_defaults("bar", "2.0") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("bar", "2.0")
+      end
     end
 
     it "no shard.yml, but git tagged version" do
@@ -70,28 +91,44 @@ describe Crystal::Doc::ProjectInfo do
     # Non-git directory
     ProjectInfo.find_git_version.should be_nil
 
+    # Empty git directory
     `git init`
     ProjectInfo.find_git_version.should be_nil
 
+    # Non-tagged commit
     File.write("file.txt", "foo")
     `git add file.txt`
     `git commit -m 'Initial commit' --no-gpg-sign`
-    ProjectInfo.find_git_version.should be_nil
+    ProjectInfo.find_git_version.should eq "master"
 
+    # Non-tagged commit, dirty workdir
+    File.write("file.txt", "bar")
+    ProjectInfo.find_git_version.should eq "master-dev"
+
+    `git checkout -- .`
+
+    # Tagged commit
     `git tag v0.1.0`
     ProjectInfo.find_git_version.should eq "0.1.0"
 
+    # Tagged commit, dirty workdir
     File.write("file.txt", "bar")
-    ProjectInfo.find_git_version.should be_nil
+    ProjectInfo.find_git_version.should eq "0.1.0-dev"
 
+    # Tagged commit, dirty index
     `git add file.txt`
-    ProjectInfo.find_git_version.should be_nil
+    ProjectInfo.find_git_version.should eq "0.1.0-dev"
 
     `git reset --hard v0.1.0`
     ProjectInfo.find_git_version.should eq "0.1.0"
 
+    # Multiple tags
     `git tag v0.2.0`
-    ProjectInfo.find_git_version.should be_nil
+    ProjectInfo.find_git_version.should eq "master"
+
+    # Other branch
+    `git checkout -b foo`
+    ProjectInfo.find_git_version.should eq "foo"
   end
 
   describe ".read_shard_properties" do
--- a/spec/compiler/crystal/tools/doc/project_info_spec.cr
+++ b/spec/compiler/crystal/tools/doc/project_info_spec.cr
@@ -3,6 +3,10 @@ require "../../../../support/tempfile"
 
 private alias ProjectInfo = Crystal::Doc::ProjectInfo
 
+private def run_git(command)
+  `git  -c user.email="" -c user.name="spec" #{command}`
+end
+
 describe Crystal::Doc::ProjectInfo do
   around_each do |example|
     with_tempfile("docs-project") do |tempdir|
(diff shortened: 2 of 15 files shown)


straight-shoota:
@crystal-lang/crystallers Any feedback on this? I'd like to move on with this 💪 

Regarding the version problem, it might be best to use a generic version identifier like `dev` by default. If in a git repository, the working directory and index are clean and HEAD is a tagged version, that version tag should be used. That should work great without any configuration when generating docs for releases.
When building docs in any other state, it often doesn't really matter what the version says (for development build). Otherwise it requires some explicit configuration using `--project-version` option, but that should be fine. It would often be difficult to determine automatically, anyway.
If we discover other use cases that can be easily modeled to work without configuration, we can add improve on it later.

asterite:
I think this is great

src/compiler/crystal/command/docs.cr
```diff
@@ -103,9 +115,27 @@ class Crystal::Command
...
-    Doc::Generator.new(result.program, included_dirs, output_directory, output_format, sitemap_base_url, sitemap_priority, sitemap_changefreq).run
+    Doc::Generator.new(result.program, included_dirs, output_directory, output_format, sitemap_base_url, sitemap_priority, sitemap_changefreq, project_info).run
 
     report_warnings result
     exit 1 if warnings_fail_on_exit?(result)
   end
+
+  private def create_project_info(name, version)
+    name ||= Doc::ProjectInfo.find_default_name
+    unless name
```
RX14, line 126: What about just... Not including the little "project summary" box if there's no name.

straight-shoota:
I've considered simply ignoring missing data. But my train of thought was that the doc generator would often be used in automatic build scripts which generate and deploy the documentation. If anything goes wrong - for whatever reason - but the generator just proceeds with a default value, the error will go unnoticed. But I'd like a CI script to fail if it can't determine the values it needs. And not publish docs for `unknown` shard.
Obviously we could add another flag like `--no-project-defaults` for this mode but that makes everything even more complex. I don't think erroring is such a big deal.
Ideally, errors should only happen in irregular situations, so I'd definitely work on improving the version detection. For example, we could fall back to using the branch name and/or commit ref if not tagged, and append `-dev` if the index/workdir are dirty. That should probably eliminate most causes for the error.

RX14:
Erroring on not finding the project name is acceptable. Erroring if it's not a git repo is not. I think a fallback to `shard.yml` version would be fine?

bcardiff:
Also, I would avoid making git a hard requirement in the compiler. Surely it will most likely be there, but either way.

Other that than and what @RX14 said, LGTM.

spec/compiler/crystal/tools/doc/project_info_spec.cr
```diff
@@ -0,0 +1,118 @@
...
+        `git init`
+        `git add shard.yml`
+        `git commit -m 'Initial commit' --no-gpg-sign`
+        `git tag v3.0`
+        ProjectInfo.new_with_defaults(nil, nil) { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "3.0")
+        ProjectInfo.new_with_defaults("bar", "2.0") { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("bar", "2.0")
+
+        # git dirty dir
+        File.write("foo.txt", "bar")
+        ProjectInfo.new_with_defaults(nil, nil) { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "1.0")
```
straight-shoota, line 31: This case needs some refinement. It feels wrong to simply fall back to the shard.yml version if the git repo is dirty. This is likely incorrect and a source of confusion. To make a dirty repo more explicit, I'd suggest to append `-dev` to the version (or some other indicator if there are better suggestions). In that specific case, I would still use the tag from HEAD (and append `-dev`) because git tags should have priority over shard.yml property.
If we're not in a git repo, using shard.yml version is fine.

spec/compiler/crystal/tools/doc/project_info_spec.cr
```diff
@@ -0,0 +1,128 @@
...
+
+    it "shard.yml and git tagged version dirty" do
+      File.write("shard.yml", "name: foo\nversion: 1.0")
+      File.write("foo.txt", "bar")
+      `git init`
+      `git add shard.yml`
+      `git commit -m 'Initial commit' --no-gpg-sign`
+      `git tag v3.0`
+
+      ProjectInfo.new_with_defaults(nil, nil) { |name, version| "missing:#{name}:#{version}" }.should eq ProjectInfo.new("foo", "1.0")
```
straight-shoota, line 49: This case needs some refinement. It feels wrong to simply fall back to the shard.yml version if the git repo is dirty. This is likely incorrect and a source of confusion. To make a dirty repo more explicit, I'd suggest to append -dev to the version (or some other indicator if there are better suggestions). In that specific case, I would still use the tag from HEAD (and append -dev) because git tags should have priority over shard.yml property.
Shard.yml version directly should only be used when we're not in a git repo or HEAD is not tagged and workdir is clean.

straight-shoota: > when we're not in a git repo or HEAD is not tagged and workdir is clean.

On second though, that shouldn't use the version from shard.yml either. It usually points to the latest release and doesn't represent the current state. Probably best to use the branch name then.

bin/ci
```diff
@@ -73,6 +73,10 @@ prepare_system() {
 }
 
 build() {
+  # Set git config for doc generator specs
+  with_build_env 'git config --global user.email "you@example.com"'
+  with_build_env 'git config --global user.name "Your Name"'
```
RX14, line 78: If we shell out to `git` in the specs, we should support all git setups without environment tweaks.

`git -c user.email="" -c user.name="" commit ...` is a more robust way to solve this.

straight-shoota: I just copied what shards does. But your solution looks better, yes.

straight-shoota:
I've added some improvements. Ready for final review.

src/compiler/crystal/tools/doc/project_info.cr
```diff
@@ -1,19 +1,29 @@
...
-        version ||= shard_version
+    def initialize(@name : String? = nil, @version : String? = nil )
+    end
 
-        unless name && version
-          return yield name, version
+    def_equals_and_hash @name, @version
+
+    def fill_with_defaults
+      unless self.version
```
Sija: Shouldn't this use @ivar? Calling bang getter will raise in case of `nil`.

ditto below

straight-shoota:
This is now live on nightly: https://crystal-lang.org/api/master/

bcardiff:
FYI for contributors of std-lib this PR force them to compile the compiler before been able to run `make docs`, otherwise they'll get.
 
```
./bin/crystal docs src/docs_main.cr  --project-name=Crystal --project-version=0.35.0-dev
Error: Invalid option: --project-name=Crystal
```

It will only be for one release. Until then, is changing a bit the workflow.

fabianhjr:
Just as a reference, encountered the issue mentioned by bcardiff while trying to build 0.35.1

```
building '/nix/store/7166z26xf45mfn6xz4vymm4g5q98y8cj-crystal-0.35.1.drv'...
unpacking sources
unpacking source archive /nix/store/bg8ib2w8m8y45dgj8f6gjx3rdl724pi4-source
source root is source
patching sources
substituteStream(): WARNING: pattern '{% if flag?(:gnu) %}"listen: "{% else %}"bind: "{% end %}' doesn't match anything in file 'spec/std/socket/tcp_server_spec.cr'
substituteStream(): WARNING: pattern '01 Jan 2020' doesn't match anything in file 'spec/std/http/cookie_spec.cr'
configuring
no configure script, doing nothing
building
build flags: -j16 -l16 SHELL=/nix/store/2jysm3dfsgby5sw5jgj43qjrb5v79ms9-bash-4.4-p23/bin/bash CRYSTAL_CONFIG_VERSION=0.35.1 all docs
Using /nix/store/gijq02j6irrp5f6q0f5nandzv5cdzx8b-llvm-9.0.1/bin/llvm-config [version=9.0.1]
g++ -c  -o src/llvm/ext/llvm_ext.o src/llvm/ext/llvm_ext.cc -I/nix/store/gijq02j6irrp5f6q0f5nandzv5cdzx8b-llvm-9.0.1/include -std=c++11   -fno-exceptions -D_GNU_SOURCE -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS
gcc -fPIC    -c -o src/ext/sigfault.o src/ext/sigfault.c
./bin/crystal docs src/docs_main.cr  --project-name=Crystal --project-version=0.35.1 --source-refname=
Error: Invalid option: --project-name=Crystal
make: *** [Makefile:96: docs] Error 1
make: *** Waiting for unfinished jobs....
builder for '/nix/store/7166z26xf45mfn6xz4vymm4g5q98y8cj-crystal-0.35.1.drv' failed with exit code 2
error: build of '/nix/store/7166z26xf45mfn6xz4vymm4g5q98y8cj-crystal-0.35.1.drv' failed
```

Will investigate

straight-shoota:
Judging from the error message it seems bin/crystal points to the previous release 0.34.0 which doesn't know the CLI options added in 0.35.0. Docs should be build with the same (i.e. newly built) compiler version.

fabianhjr:
Yeah, it seems to be the case though trough reading the makefile I am wondering if `docs` should depend on `crystal`.

fabianhjr:
For the time being I think I will add a patch for `docs:` -> `docs: crystal` on the Makefile that seems to fix the build issue I encountered.

straight-shoota:
> I am wondering if docs should depend on crystal.

That's probably a bad idea. It would force everyone to build the compiler when you just want to build the docs. But that's usually completely unnecessary. It's only relevant when there are significant differences between the version of the compiler binary and the source code which is rarely the case.

Most package builder workflows build the compiler first and then use that fresh compiler to build the docs. Does this not work for your use case? (I'm not familiar with nix build process, except that everythings pretty much isolated; but you should be able to somehow use the matching compiler version for dependend tasks?)

fabianhjr:
It should be safe to apply that patch to the nix build, this only came up because it is attempting to do a  `make all docs` simultaneously. Since this is for the build system it should not be a problem. That patch would only be added downstream.

PR #8795, opened 2020-02-12
Fix MIME spec examples to not collide with actual registry

### LINKED ISSUE #8793: std_spec fails when 'bar' is a valid MIME type extension
## Environment

* Linux

## Replicate

```shell
# Set MIME type associations to mailcap
curl -o /etc/mime.types https://pagure.io/mailcap/raw/master/f/mime.types

# Clone the crystal repo
clone https://github.com/crystal-lang/crystal

# Run unit tests
cd crystal
make std_spec
```

## Error

Failures:

```
1) MIME .load_mime_database
    Failure/Error: MIME.from_extension?(".bar").should be_nil

    Expected: "application/vnd.qualcomm.brew-app-res" to be nil

    # spec/std/mime_spec.cr:132
```

## Note

* Within ***mailcap***, 'bar' has an associated MIME.
* ***mailcap*** is a [default dependency](https://www.archlinux.org/packages/extra/x86_64/firefox/), on arch linux, for firefox.

straight-shoota:
Fixes #8793 

Prefixing the file extensions should make sure they don't match any existing values in the system's MIME registry.

### DIFF
--- a/spec/std/mime_spec.cr
+++ b/spec/std/mime_spec.cr
@@ -171,14 +171,12 @@ describe MIME do
     end
 
     it "skips loading defaults" do
-      MIME.reset!
       MIME.init(load_defaults: false)
       MIME.initialized.should be_true
       MIME.from_extension?(".html").should be_nil
     end
 
     it "loads file" do
-      MIME.reset!
       MIME.initialized.should be_false
       MIME.init(datapath("mime.types"))
       MIME.from_extension?(".foo").should eq "foo/bar"
--- a/spec/std/mime_spec.cr
+++ b/spec/std/mime_spec.cr
@@ -6,7 +6,7 @@ module MIME
     @@initialized
   end
 
-  def self.reset
+  def self.reset!
     @@initialized = false
     @@types = {} of String => String
     @@types_lower = {} of String => String
@@ -45,6 +45,8 @@ describe MIME do
 
   describe ".register" do
     it "registers new type" do
+      MIME.reset!
+      MIME.init(load_defaults: false)
       MIME.register(".Custom-Type", "text/custom-type")
 
       MIME.from_extension(".Custom-Type").should eq "text/custom-type"
@@ -57,6 +59,8 @@ describe MIME do
       MIME.register(".custom-type", "text/custom-type-lower")
       MIME.from_extension(".custom-type").should eq "text/custom-type-lower"
       MIME.from_extension(".Custom-Type").should eq "text/custom-type"
+    ensure
+      MIME.reset!
     end
 
     it "fails for invalid extension" do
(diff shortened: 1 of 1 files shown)


DanielDavis5:
Granted that name will almost probably never conflict, wouldn't it just be better to reset and reinit with the loaddefaults set to false?

straight-shoota:
I'm sure there was a reason why I didn't do that in the first place. But I don't recall. It looks you might be right.

straight-shoota:
Updated the specs to call `MIME.reset` and `MIME.init(load_defaults: false)`. Also added that to all other specs.

Blacksmoke16:
Might be a good usecase for `before_each` hook now that those exist?

RX14:
`around_each` would abstract the whole thing with no changes to specs, I think.

RX14:
@straight-shoota want to work on this or should this be merged without any changes to fix the spec suite?

straight-shoota:
Rebased on master and changed to use spec hooks.

spec/std/mime_spec.cr
```diff
@@ -147,30 +165,23 @@ describe MIME do
...
       MIME.init
       MIME.initialized.should be_true
       MIME.from_extension(".html").partition(';')[0].should eq "text/html"
-    ensure
-      MIME.reset
     end
 
     it "skips loading defaults" do
-      MIME.reset
+      MIME.reset!
```
Sija: Shouldn't it be removed as well?

spec/std/mime_spec.cr
```diff
@@ -147,30 +165,23 @@ describe MIME do
...
       MIME.init(load_defaults: false)
       MIME.initialized.should be_true
       MIME.from_extension?(".html").should be_nil
-    ensure
-      MIME.reset
     end
 
     it "loads file" do
-      MIME.reset
+      MIME.reset!
```
Sija: ditto

spec/std/mime_spec.cr
```diff
@@ -15,7 +15,16 @@ module MIME
 end
 
 describe MIME do
+  before_each do
+    MIME.reset!
+  end
+
+  after_each do
+    MIME.reset!
+  end
```
Sija: Why not use `around_each` instead?

spec/std/mime_spec.cr
straight-shoota: Oh, missed that.

spec/std/mime_spec.cr
straight-shoota: Does it make a difference? `around_each` is for when you need to share state between setup and teardown steps. There's no locally shared state here, so can just use before and after hooks.

PR #8796, opened 2020-02-13
Compiler: fix inherited visibility of macro hooks

### LINKED ISSUE #8794: Regression bug regarding visibility of initializers
The following code compiles in 0.32.1 and does not in `0.33.0-dev [380570325]`.
Reverting #8762 fixes this, but the #8715 is actually something that should not work.

```cr
module M1
  macro included
    include M2
  end
end

module M2
  macro setup_initializer_hook
    macro finished
      generate_needy_initializer
    end

    macro included
      setup_initializer_hook
    end

    macro inherited
      setup_initializer_hook
    end
  end

  macro included
    setup_initializer_hook
  end

  macro generate_needy_initializer
    {% if !@type.abstract? %}
      def initialize(a)
      end
    {% end %}
  end
end

abstract class Base
  include M1
end

private class Foo < Base
end

Foo.new(1)
```

This code was reduced from some luckyframework/lucky specs that are not building in nightly.

asterite:
Fixes #8794

When expanding a macro we always pass a related ASTNode and were taking the visibility from it. The problem is, on macro hooks the node was the class definition, and if it was `private` it was taken from that which was wrong.

To make the code simpler, now the macro expansion code requires an explicit visibility to be passed. Then in all places I made sure to pass the correct visibility. For macro hooks the initial visibility should always be `public`.

### DIFF
--- a/spec/compiler/semantic/private_spec.cr
+++ b/spec/compiler/semantic/private_spec.cr
@@ -519,4 +519,50 @@ describe "Semantic: private" do
       "private constant Foo::Bar referenced",
       inject_primitives: false
   end
+
+  it "doesn't inherit visibility from class node in macro hook (#8794)" do
+    semantic(%(
+      module M1
+        macro included
+          include M2
+        end
+      end
+
+      module M2
+        macro setup_initializer_hook
+          macro finished
+            generate_needy_initializer
+          end
+
+          macro included
+            setup_initializer_hook
+          end
+
+          macro inherited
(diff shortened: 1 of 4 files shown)


bcardiff:
Nice spec :-). Great way to start the day. Thanks!

asterite:
> Nice spec :-)

Haha 😅 

I wanted to reduce it but just in case I left it as is to be sure I'm fixing the actual issue.

PR #8803, opened 2020-02-14
refactoring URI resolve method

VanyaNeytrino:
Hi) This is mini refactoring for to improve the readability of the code

### DIFF
--- a/src/uri.cr
+++ b/src/uri.cr
@@ -328,13 +328,13 @@ class URI
       target.port = port
       target.user = user
       target.password = password
-      target.path = target_path
+      target.path = fill_target_path
     end
 
     target
   end
 
-  private def target_path(target : URI)
+  private def fill_target_path(target : URI)
     if target.path.empty?
       target.path = remove_dot_segments(path)
       target.query ||= query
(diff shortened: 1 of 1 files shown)


src/uri.cr
```diff
@@ -317,37 +317,33 @@ class URI
...
-      target = uri.dup
-    else
-      target = URI.parse(uri)
-    end
-
-    if target.absolute? || opaque?
-      return target
-    end
-
+    uri.is_a?(URI) ? target = uri.dup : target = URI.parse(uri)
```
bew:
```suggestion
    target = uri.is_a?(URI) ? uri.dup : URI.parse(uri)
```

src/uri.cr
```diff
@@ -317,37 +317,33 @@ class URI
...
-        base = path
-        if base.empty? && target.absolute?
-          base = "/"
-        end
-        target.path = resolve_path(target.path, base: base)
-      end
+      target.path = target_path
     end
-
+
```
yxhuvud: I won't touch the discussion of if the changes improve stuff or not, but you should probably get into the habit of doing `crystal tool format` before you commit.

straight-shoota:
Hey @VanyaNeytrino, you don't need to close this PR right away. If you feel it improves the code for better readability, I'm open to hear why 👍 Also I'm just a single voice, maybe others would approve it. I don't know.

If you agree it's not necessarily an enhancement or don't want to invest time into a discussion that's fine, too. Thanks, for posting this PR anyways then.

PR #8804, opened 2020-02-14
Update CI to use 0.33.0

bcardiff:


src/va_list.cr
```diff
@@ -18,9 +18,7 @@ struct VaList
     end
   end
 
-  {% if compare_versions(Crystal::VERSION, "0.33.0-0") > 0 %}
```
straight-shoota: This would mean a 0.32.1 compiler can't build 0.34.0-dev compiler. Can we leave this guard in for at least another release for easier reproduction? Also this change and the similar one in int_spec are not related to the purely administrative CI/repo update.

bcardiff: We use the previous compiler to build the next one. In this case it can be omitted, but usually there is no guarantee for older compilers.

If wanted we can change it to

```
{% if compare_versions(Crystal::VERSION, "0.33.0") > 0 %}
```

at least, but we usually should not have many of them.

bcardiff: I reverted this change. It doesn't hurt.

straight-shoota:
Thanks @bcardiff =)

Issue #8738, opened 2020-02-04
Memory corruption with Thread on musl


straight-shoota:
The test_alpine job fails in some workflows on master (Example: https://circleci.com/gh/crystal-lang/crystal/37573) but not every time (Example: https://circleci.com/gh/crystal-lang/crystal/37643). 
The error happens while running stdlib specs with a newly built compiler. The error message is `Failed to raise an exception: END_OF_STACK`, so it's not clear what's the issue. I'd assume that would be improved by #8735 but on that branch the error message is identical: https://circleci.com/gh/crystal-lang/crystal/37656

Blacksmoke16:
Maybe #8743?

bcardiff:
It fails often. Either we find a fix soon or we should turn off the CI for alpine. It's getting noisy.

straight-shoota:
Running against master with #8743 the error message is now `Error while trying to determine if a stack overflow has occurred. Probable memory corrpution`.

The error seems to occur in different thread specs, I've found these locations:
* `spec/std/thread/condition_variable_spec.cr:17`
* `spec/std/thread/mutex_spec.cr:5`
* `spec/std/thread/thread_spec.cr:30`
* `spec/std/thread/thread_spec.cr:4`
* `spec/std/thread/thread_spec.cr:11`

It doesn't seem to be consistent. And it's not consistent whether the error occurs at all. When running only thread specs (`--example Thread`) it's about 50% failure rate. I didn't tally when running the entire spec suit, but it feels like it fails more often.

When running with strace (and `--example Thread`), the error occurs every single time. But it's always in one of the first specs of `thread_spec.cr` which never failed when running the entire spec suite without strace.

Those are the traces:
```
write(4, "allows passing an argumentless f"..., 45allows passing an argumentless fun to execute) = 45
mmap(NULL, 143360, PROT_NONE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f91579cf000
mprotect(0x7f91579d1000, 135168, PROT_READ|PROT_WRITE) = 0
rt_sigprocmask(SIG_BLOCK, ~[RTMIN RT_1 RT_2], [], 8) = 0
[...]

PR #8807, opened 2020-02-15
Fix exceptions not being inspectable when running binary from PATH

willhbr:
If the binary is added to $PATH, PROGRAM_NAME is the name of the file,
but not the path to read debug information from. This causes a file open
error, which is incredibly hard to debug. Using `Process.executable_path` will
instead always be the current executable.

A simple demonstration of this issue is:

```crystal
def bar
  raise "Oh no it didn't work"
rescue e
  puts e.inspect_with_backtrace
end

bar
```

Build and move this binary somewhere in your `$PATH`:

```shell
$ crystal build exceptions.cr
$ mv exceptions ~/.local/bin
$ exceptions
Unhandled exception: Error opening file 'exceptions' with mode 'r': No such file or directory (Errno)             
Error while trying to dump the backtrace: Error opening file 'exceptions' with mode 'r': No such file or directory (Errno)
```

With this patch, the exception is printed as expected:

```
Oh no it didn't work (Exception)
  from exceptions.cr:2:3 in 'bar'
  from exceptions.cr:11:1 in '__crystal_main'
  from /home/will/projects/gh/crystal-lang/crystal/src/crystal/main.cr:106:5 in 'main_user_code'
  from /home/will/projects/gh/crystal-lang/crystal/src/crystal/main.cr:92:7 in 'main'
  from /home/will/projects/gh/crystal-lang/crystal/src/crystal/main.cr:115:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```

### DIFF
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -367,7 +367,9 @@ struct CallStack
       @@base_address : UInt64|UInt32|Nil
 
       protected def self.read_dwarf_sections
-        Debug::ELF.open(PROGRAM_NAME) do |elf|
+        program = Process.executable_path
+        return unless program && File.readable? program
+        Debug::ELF.open(program) do |elf|
           elf.read_section?(".text") do |sh, _|
             @@base_address = sh.addr - sh.offset
           end


willhbr:
Ok I thought alpine was failing because it didn't support `/proc/self/exe`, but I've changed it to use `Process.executable_path` and it's still failing. Any advice would be appreciated here!

rdp:
Looks like alpine's failing on something unrelated...
LGTM, almost wish we'd fix PROGRAM_NAME to always be absolute instead but... :)

bcardiff:
It’s #8738 do not worry about alpine

straight-shoota:
I'm not sure this should use `PROGRAM_NAME` at all. It's a compile time value and not necessarily represents the binary name at runtime.

I'd also like to make this behaviour more resilient in case the file can't be read at all. For example because the executable doesn't exist anymore. We should make sure to still print a meaningful account of the original error.

src/callstack.cr
```diff
@@ -367,7 +367,7 @@ struct CallStack
       @@base_address : UInt64|UInt32|Nil
 
       protected def self.read_dwarf_sections
-        Debug::ELF.open(PROGRAM_NAME) do |elf|
+        Debug::ELF.open(Process.executable_path || PROGRAM_NAME) do |elf|
```
ysbaddaden:
```suggestion
        return unless path = Process.executable_path
        Debug::ELF.open(path) do |elf|
```

willhbr: Done!

straight-shoota: What if `exception_path` is present but `File.open` fails?

src/callstack.cr
```diff
@@ -367,7 +367,8 @@ struct CallStack
       @@base_address : UInt64|UInt32|Nil
 
       protected def self.read_dwarf_sections
-        Debug::ELF.open(PROGRAM_NAME) do |elf|
+        return unless path = Process.executable_path
```
Sija: I'd suggest using exactly the same code as above, i.e.:

```cr
        program = Process.executable_path
        return unless program
```

willhbr:
I've added a check for if the file exists (like the MacOS version), unsure of what to do to indicate that there was a problem reading the backtrace, but I think that can be addressed in another PR.

src/callstack.cr
```diff
@@ -367,7 +367,10 @@ struct CallStack
       @@base_address : UInt64|UInt32|Nil
 
       protected def self.read_dwarf_sections
-        Debug::ELF.open(PROGRAM_NAME) do |elf|
+        program = Process.executable_path
+        return if program.nil?
+        return unless File.exists? program
```
straight-shoota:
```suggestion
        return unless File.readable? program
```

src/callstack.cr
willhbr: Then we're back in the same position, but I think making that more robust should be done in another PR - it'll have to be done for the other executable types as well.

src/callstack.cr
```diff
@@ -367,7 +367,10 @@ struct CallStack
       @@base_address : UInt64|UInt32|Nil
 
       protected def self.read_dwarf_sections
-        Debug::ELF.open(PROGRAM_NAME) do |elf|
+        program = Process.executable_path
+        return if program.nil?
+        return unless File.readable? program
```
Sija: this can be shorten to just:

```cr
return unless program && File.readable? program
```

PR #8676, opened 2020-01-11
Add continuous testing on Windows using GitHub Actions

oprypin:
This is a pipeline using a Linux job that cross-compiles for Windows, followed by an actual Windows job linking and running that.
    
Based on https://github.com/crystal-lang/crystal/wiki/Porting-to-Windows/30956f45a4e739204881694ee6c92f4fb0bc5395

Note that all the actual commits from this PR are prefixed with "Win CI:", the rest is merges of pull requests that this relies on.
This PR creates and touches *only* the file ".github/workflows/win.yml".

**[Example in action](https://github.com/oprypin/crystal/actions?query=branch%3Agh-ci+is%3Asuccess)**

Fixes #8666.

PR #8811, opened 2020-02-16
Win CI: Use latest versions of libraries

oprypin:
* bdwgc has switched to CMake, which is nicer to use, but that change is not in any release yet (there hasn't been one in a while). Use an unreleased commit and switch to CMake.
* use the workflow file itself as the cache key for external libs, otherwise forgetting to update it will cause confusion.
* improve wording of final step names.

### DIFF
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -40,7 +40,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: libs
-          key: win-libs-v1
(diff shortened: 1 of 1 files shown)


j8r:
It is not a good idea to use a different bdwgc version than other CIs. Is CMake the only advantage?

oprypin:
Yes. And the fact that the lib is updated.

https://github.com/crystal-lang/crystal/pull/8676#issuecomment-573416370

Issue #7588, opened 2019-03-25
Fix example codes more often


straight-shoota:
I suppose running these tests like #7569 is pretty much automated and it shouldn't require much manual effort to get a list of broken examples.

Maybe we could integrate that into the pre-release cycle? Or even into CI with nightly-builds and maybe PR tests? The latter could be limited to files/examples changed in the pull request.

This would allow us to fix errors in the examples sooner and avoid distributing errors in the documentation.

/cc @maiha @bcardiff

bcardiff:
Weekly seems enough. I am not sure how automated is the tool.

But I am reluctant to include dependencies outside crystal-lang org since it's outside of our control. It could act as an optional check before the release. But in that case, it could be automated outside the organization and still be valuable to know if everything is ok or not.

In the mid/long term, it sounds awesome, of course!

straight-shoota:
It would probably even suffice to run it manually for now. Maybe at the same time as preparing release notes.

maiha:
Because there would be various examples such as **pseudo code** and **random output**, I think it is dangerous to put in CI now.
```crystal
Greetings, <%= @name %>!
```

```crystal
Random.rand(5) # => 2
```

For the time being, as you say, I agree with the idea of ​​checking before release. Please contact me once the code is fixed. :pray:

straight-shoota:
It's been a while since last checking the example codes (most recent PR: https://github.com/crystal-lang/crystal/pull/12870).
I suppose one of the reasons is that it requires async manual labor to run the checks and apply fixes, which is hard to manage. Striving for automation is the key in my opinion. Ideally every change to example code should be tested in CI directly so we don't inadvertedly merge broken examples.

I think this really shouldn't be too hard with a couple of improvements.

It is my understanding that the major obstacle for automation right now is that the tool https://github.com/maiha/crystal-examples stores information about individual example codes in an external database. This is a screenshot from the web interface:

![grafik](https://github.com/user-attachments/assets/5ba30690-4f58-43ae-b722-d709903f4443)

It offers multiple reasons for why a code example would not be expected to validate.

I think this information should not be stored externally. It should be attached directly to the example. The markdown syntax offers two options for that:
* We can change the language tag. Currently any code block without a tag is implicitly tagged as `crystal`. We could introduce other language tags such as `crystal-invalid` to indicate that the block does not describe a valid example. These examples should of course still _format_ with Crystal highlighting, but the example checker would know that it's not supposed to work.
[...]

PR #8816, opened 2020-02-18
Fix pretty_json outdated sample

### LINKED ISSUE #8815: 0.33 breaks samples/pretty_json.cr
##### `Crystal 0.33.0 [612825a53] (2020-02-14)`
Write nothing.
```console
$ echo '[1, {"two": "three"}, false]' | crystal samples/pretty_json.cr
```

##### `Crystal 0.32.1 [41bd18fbe] (2019-12-18)` or older versions
Works fine.
```console
$ echo '[1, {"two": "three"}, false]' | crystal samples/pretty_json.cr
[
  1,
  {
    "two": "three"
  },
  false
]
```

Best regards,

asterite:
Fixes #8815

JSON pull parser kind is no longer a symbol, it's an enum.

### DIFF
--- a/samples/pretty_json.cr
+++ b/samples/pretty_json.cr
@@ -20,32 +20,34 @@ class PrettyPrinter
 
   def read_any
     case @pull.kind
-    when :null
+    when .null?
       with_color.bold.surround(@output) do
         @pull.read_null.to_json(@output)
       end
-    when :bool
+    when .bool?
       with_color.light_blue.surround(@output) do
         @pull.read_bool.to_json(@output)
       end
-    when :int
+    when .int?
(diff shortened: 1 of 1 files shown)


oprypin:
Should building the samples be part of CI? 🤔

oprypin:
Nevermind, looking at the issue, building wouldn't catch it.

asterite:
@oprypin It is a thing! The problem is that the program compiled but it was wrong. If we had exhaustive case I would sure have put an `else raise` there.

straight-shoota:
I suppose https://github.com/maiha/crystal-examples can recognize such failures.
See #7588

PR #8817, opened 2020-02-18
Fix UDP specs in the case of a local firewall

RX14:


spec/std/socket/udp_socket_spec.cr
```diff
@@ -98,7 +98,13 @@ describe UDPSocket do
         udp.multicast_loopback?.should eq(true)
 
         udp.send("testing", addr)
-        udp.receive[0].should eq("testing")
+        udp.read_timeout = 1.second
+        begin
+          udp.receive[0].should eq("testing")
+        rescue timeout : IO::Timeout
```
bew:
```suggestion
        rescue IO::Timeout
```
?

spec/std/socket/udp_socket_spec.cr
```diff
@@ -98,7 +98,13 @@ describe UDPSocket do
         udp.multicast_loopback?.should eq(true)
 
         udp.send("testing", addr)
-        udp.receive[0].should eq("testing")
+        udp.read_timeout = 1.second
+        begin
+          udp.receive[0].should eq("testing")
+        rescue IO::Timeout
+          # Since this test doesn't run over the loopback interface, this test
+          # fails when there is a firewall in use. Don't fail in that case.
```
ysbaddaden: maybe warn about the spec being skipped and why?

RX14: It's just extra noise, I really don't think it's necessary.

ysbaddaden: Not failing the test because of local configuration is great, but _skipping_ a test is still different than _passing_ a test.

With minitest it would be a mere `skip "local firewall causes the test to hang forever"`. No noise, but a yellow `S` in the progress instead of a green dot and a final warning (some tests were skipped), and the reason printed when running the tests with `--verbose`.

RX14: It would be nice if we could skip tests from inside specs. But this isn't completely skipping the test: there's still a *lot* of syscalls run even without sending any packets.

straight-shoota: Since there's no way to report that a spec was not 100% run, I think this is fine. We're also skipping parts of other specs based on architecture.

j8r: It may be possible to test before hand if the test pass or not, then execute a `it` if it does, else `pending`.

Something like:
```cr
def test_udp : String?
  #stuff
  true
rescue
  nil
end

if result = test_udp
  it "UDP works" do
    result.should eq "testing"
  end
else
  pending "UDP not supported" { }
end
```

straight-shoota: Yeah, but that really seems too much effort for this. It would be nice if there was a simple solution to get some feedback about partial spec skipping, but there is none. And it's not a big deal to leave it as is.

ysbaddaden: @j8r that's too complex. `spec` is just missing a `skip(message)` method —which exists in both RSpec and Minitest— to properly skip a running test.

PR #8829, opened 2020-02-21
Raise when calling #size on an open range

### LINKED ISSUE #8151: Calling Range(_, Nil)#size hangs in runtime
Code below is enough to kill the compiler:

```cr
(0..).size
```

Infinite `Iterator`s are having the same issue btw:

```cr
[1, 2, 3].cycle.size
```

Sija:
Closes #8151

### DIFF
--- a/spec/std/range_spec.cr
+++ b/spec/std/range_spec.cr
@@ -481,6 +481,18 @@ describe "Range" do
     it "works for other types" do
       ('a'..'c').size.should eq(3)
     end
+
+    it "raises on beginless range" do
+      expect_raises(ArgumentError, "Can't calculate size of an open range") do
+        ((true ? nil : 1)..3).size
+      end
+    end
+
+    it "raises on endless range" do
+      expect_raises(ArgumentError, "Can't calculate size of an open range") do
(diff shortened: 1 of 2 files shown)


spec/std/range_spec.cr
```diff
@@ -481,6 +481,18 @@ describe "Range" do
     it "works for other types" do
       ('a'..'c').size.should eq(3)
     end
+
+    it "raises on beginless range" do
+      expect_raises(ArgumentError, "Can't calculate size of an open range") do
+        ((true ? nil : 1)..3).size
```
bcardiff: Why `nil..3` is not enough?

Sija: Because #size raises at the compile-time, you need to have an union to invoke the runtime behaviour, other way of writing that would be `nil.as(Int32?)`.

bcardiff: Ok. I missed that detail

Sija:
That's a breaking change, mind you.

RX14:
I don't think that any program which worked before will fail to work now.

PR #8839, opened 2020-02-23
Add the binary name of the project/lib to the .gitignore.

postmodern:
A lazy developer might run `crystal build src/project.cr` and accidentally add the `./project` binary. Safer to ignore any built binaries.

### DIFF
--- a/spec/compiler/crystal/tools/init_spec.cr
+++ b/spec/compiler/crystal/tools/init_spec.cr
@@ -62,6 +62,7 @@ module Crystal
           gitignore.should contain("/.shards/")
           gitignore.should contain("/shard.lock")
           gitignore.should contain("/lib/")
+          gitignore.should contain("/example")
         end
 
         with_file "example_app/.gitignore" do |gitignore|
@@ -69,6 +70,7 @@ module Crystal
           gitignore.should contain("/.shards/")
           gitignore.should_not contain("/shard.lock")
           gitignore.should contain("/lib/")
+          gitignore.should contain("/example")
         end
 
         {"example", "example_app", "example-lib", "camel_example-camel_lib"}.each do |name|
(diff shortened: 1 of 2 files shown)


Sija:
Binary name could differ from the project name, especially when configuring it/them via `targets` section in `shard.yml`. I'd leave it up to the user to decided whether he'd like to ignore it.

postmodern:
@Sija if the binary name differed, then would it be safe to assume the developer manually changed `shards.yml` after generating a new shard project? I do not see an option in `crystal init` to specify a custom `targets`. If the developer manually changed `shards.yml`, couldn't they simply edit `.gitignore` as well, or would that be introducing yet another step?

Also, is there a scenario where the user would want to `git add` the compiled binary?

straight-shoota:
The idiomatic way to build a project is using `shards build` which puts the binary in `bin` folder. We could consider adding `bin` to `.gitignore`. But arbitrary file names just because they match the shard name seems to restrictive default to me. Also it's weird when there's different behaviour depending on the output path of `shards build`.

bcardiff:
I don't think it hurts this addition.

* `crystal build src/app.cr` is common
* yet more common is `crystal [run] src/app.cr` 
* but crystal build is still suggested in lots of guides so is an expected flow to me

j8r:
`shards build` has lots of limitations, like missing options and flags. It can't always be used.

Sija:
@bcardiff hmm, as far as I remember, there's no open tickets regarding issue at hand, so we can say that's imaginary problem. Thus I'd say it **does** hurt to add noise to the codebase - why am I being forced to remove it every time from the `.gitignore` files from now on?

postmodern:
Closing since `/bin/` is already added to `.gitignore` for generated "apps".

PR #8840, opened 2020-02-23
All files shall have a descriptive file extension name.

postmodern:
Generally a good practice and helps with Windows support.


Sija:
If any, I'd use `.md`, since that's sth more usable these days than plain text.

straight-shoota:
On second thought: While I still agree with this change, I'd suggest not to merge it before reviewing other aspects of `crystal init` as well. I think it needs some love and further improvements.
I don't like to stall this, but it's also neither a huge change nor very important.

PR #8843, opened 2020-02-23
Expand home within the path in couple of places moar

Sija:
Some places seem unjustly forgotten, followup to #7903.

### DIFF
--- a/src/http/server/handlers/static_file_handler.cr
+++ b/src/http/server/handlers/static_file_handler.cr
@@ -19,7 +19,7 @@ class HTTP::StaticFileHandler
   # If *directory_listing* is `false`, directory listing is disabled. This means that
   # paths matching directories are ignored and next handler is called.
   def initialize(public_dir : String, fallthrough = true, directory_listing = true)
-    @public_dir = Path.new(public_dir).expand
+    @public_dir = Path.new(public_dir).expand(home: true)
     @fallthrough = !!fallthrough
     @directory_listing = !!directory_listing
   end


straight-shoota:
Why do we need home expansion in these places? There should be a clear reason for each of them. If we would just apply `home: true` randomly across stlib and compiler sources, we could've left it as default as well.

See also discussion in #8756. This is similar.

src/http/server/handlers/static_file_handler.cr
```diff
@@ -19,7 +19,7 @@ class HTTP::StaticFileHandler
   # If *directory_listing* is `false`, directory listing is disabled. This means that
   # paths matching directories are ignored and next handler is called.
   def initialize(public_dir : String, fallthrough = true, directory_listing = true)
-    @public_dir = Path.new(public_dir).expand
+    @public_dir = Path.new(public_dir).expand(home: true)
```
Sija: I'd expect that passing a `public_dir` will be expanded, incl. home.

src/compiler/crystal/compiler.cr
```diff
@@ -381,13 +381,13 @@ module Crystal
         end
       end
 
+      output_filename = File.expand_path(output_filename, home: true)
```
Sija: I'd expect that calling `-o ~/foo.obj` would work.

src/compiler/crystal/command.cr
```diff
@@ -501,10 +501,10 @@ class Crystal::Command
 
   private def gather_sources(filenames)
     filenames.map do |filename|
+      filename = File.expand_path(filename, home: true)
```
Sija: ditto, I'd expect expanding filepaths passed to the _compile_ command.

src/compiler/crystal/compiler.cr
asterite: And it works: the shell expands it for you.

Sija: yeah, unless it doesn't ;)

straight-shoota:
Expanding CLI arguments is a task for the shell. If your shell doesn't support that, you can use some other means (for example `$HOME`). That's not a feature that needs to be supplemented by the compiler.

straight-shoota:
Expanding home is intended for paths that are *not* passed from the shell, for example in configuration files, external requests or any kind of direct user input.

Sija:
@straight-shoota I agree, thus I've removed my changes except the one re: static file handler.

straight-shoota:
I don't think StaticFileHandler qualifies either.

asterite:
I'd say, if nobody creates a bug report saying "I was expecting this to expand home" based on a real use case, we shouldn't be changing things.

straight-shoota:
Also, it's always really easy to *enable* home expansion outside of stdlib code when you need it. The other way around doesn't work.

Sija:
Fair 'nuff, closing.

Issue #8810, opened 2020-02-16
Logger formatter does not have access to the location of the log call


willhbr:
I find it very useful to have the file name and line number of where a log happened in the header of the log message, for example:

```
[D 2020-02-15 06:04:32 UTC job.cr:114] Something exciting happened
```

To do this I currently have to implement a wrapper around the stdlib `Logger` class so that `.log()` and friends capture the `__FILE__` and `__LINE__` parameters. It would be incredibly useful to have these available to the formatter function, so that this can be done without messy hacks.

It could be worth taking this opportunity to refactor the formatter proc to take a single struct argument rather than multiple arguments, so this kind of change can be made without breaking everyone's loggers.

I can implement this, if this approach is OKed by the core team.

bcardiff:
The logger will have a revamp soon. No need for a PR until we share a bit of the future design. Stay tuned :-)

Sija:
See #5874 for reference

Blacksmoke16:
https://github.com/crystal-lang/crystal/pull/8847#discussion_r384551035 is probably the way I would handle this.  File/line numbers are specific to a given entry so they should just exist on the Entry itself.

Adding it into context could work, but it would have to be added to that one single entry via `Log.with_context`, otherwise it would be added to the main fiber context which doesn't make sense.

willhbr:
I'm still happy to implement this in the new logger, I've currently got a personal shard that is doing a poor job at overriding the `Log.*` methods to add the call location.

If there are concerns with including the locations of log calls in the compiled binary, there could be a compiler option to omit them.

PR #8847, opened 2020-02-24
Log module

### LINKED ISSUE #5874: Proposal: improve stdlib logger
This is piggybacking off of some chat discussion by @RX14 and @straight-shoota earlier today.

Currently there is no great way for library authors to log without friction. A library can't just write to STDERR arbitrarily because that's a potentially unwanted side effect. It instead has to take an optional `Logger?` that is nil by default, which requires boilerplate and is unlikely to actually be used.

If we instead establish a class-based logging hierarchy into the stdlib, library code can log unconditionally, with downstream users selectively enabling logs from the libraries they care about. I propose improving the situation by making the following changes:

### Structure loggers hierarchically, with granular level controls
A Logger should be able to take a parent, to whom the actual log writing is delegated. By default, a child Logger's level should be `nil`, which means it inherits the level of its parent.

```crystal
main_logger = Logger.new(STDOUT)
child = Logger.new(main_logger)
noisy_child = Logger.new(main_logger)

main_logger.level = Logger::WARN
noisy_child.level = Logger::DEBUG

child.info{ "This logger uses its parent's log level and this message does not get printed" }
noisy_child.info{ "This message will get printed because noisy_child has its own log level set" }
```

### Make logging class-aware by default
It should be trivial to define a logger that is associated with a particular class or module. That logger will inherit from the logger of its parent in the module namespace hierarchy. Parents can be created automagically.

```crystal
module MyProgram
  class App
    include Loggable # creates a Logger and adds helper methods e.g. `warn`, `info`
    def initialize
      info{ "creating an app" }
    end
  end
end

# Sets the effective log level for every Loggable under the MyProgram namespace.
# Note that the Logger for MyProgram got created behind the scenes at the same time as the Logger for App
Loggable.get_logger(MyProgram).level = Logger::DEBUG

# Logs the following (note the class name is included): 
# I, [2018-03-26 14:43:17 -07:00 #25575] INFO -- MyProgram::App: creating an app
app = MyProgram::App.new
```

Loggable will have a root level logger that by default does not log anything, so library code will stay quiet. To up the verbosity, just do:
```crystal
Loggable.get_root_logger.level = Logger::DEBUG  # Enables logging for every class and module in your program, including libraries.
```

Unrelated to the above points, Loggers currently only operate on IO objects. Modern applications frequently eschew traditional log files in favor of structured logging, so this behavior should be more customizeable. Thus, I also propose to
### Abstract Logger away from IO
Instead of storing an IO directly, Loggers should store a LogWriter, which is a module with `abstract def write` on it. The standard library should include an IOLogWriter class that takes on the formatting behavior of the current Logger, but is interchangeable with custom-defined classes like so:
```crystal
class GelfLogWriter
  include LogWriter

  def initialize(@transport, @host, @port: 12201)
  end

  def write(severity, datetime, progname, message)
    # build the document and send it over the network here...
  end
end

Loggable.get_root_logger.handler = (GelfLogWriter.new :udp, "graylog.example.com")
```

I'll be happy to work on implementing these changes if the core team thinks they'd be useful.

bcardiff:
Implements Log module as described in https://github.com/crystal-lang/crystal/issues/5874#issuecomment-587090645. With most of the documentation needed to use it.
Replaces the logger usage in the compiler (the playground used it)
Mark logger module as deprecated.
I added an internal macro method to expand to types similar to JSON::Any and YAML::Any since I needed two of them.

Fixes #5874 
Supersedes #6901


spec/std/log/broadcast_backend_spec.cr
```diff
@@ -0,0 +1,63 @@
+require "spec"
+require "log/config"
+require "../../support/log"
+
+private def s(value : Log::Severity)
+  value
+end
```
asterite: Why is this needed? Passing `:info` or `:error` directly without this indirection works just fine.

Same for all other specs.

bcardiff: In `exp.should eq(x)` the `x` is not autocasted.
And in the config specs since there are hash of arrays of config, the autocasting does not apply that either.

I agree there are many of this invocations that are not needed, but when working in bulk it was easier for me this way.

straight-shoota: Why not `private alias S = Log::Severity`?

bcardiff: That would mean `S.new(value)` rather than `s(value)` doesn't it?

asterite: Makes sense. It's a good workaround :-)

src/log/context.cr
```diff
@@ -0,0 +1,35 @@
+# Immutable structured context information for logging.
+#
+# See `Log.context`, `Log.context=`, `Log::Context#clear`, `Log::Context#set`, `Log::Context#using`.
+class Log::Context
+  Crystal.datum types: {bool: Bool, i: Int32, i64: Int64, f: Float32, f64: Float64, s: String, time: Time}, hash_key_type: Symbol, immutable: true
+
+  # Creates an empty `Log::Context`
+  def initialize
+    @raw = Hash(Symbol, Context).new
```
asterite: Why Symbol?

spec/std/log/broadcast_backend_spec.cr
bcardiff: The alias or how it is right now?

src/log/context.cr
bcardiff: Since the context are changed by `Log.context.set key: value` and thowse kwargs have symbols as key it was simple to preserve that type.

spec/std/log/broadcast_backend_spec.cr
asterite: How it is right now

src/log/context.cr
asterite: I guess that means that the context keys can't be dynamic. Is that intended?

src/log/builder.cr
```diff
@@ -0,0 +1,115 @@
...
+      # do not match foobar with foo.*
+      return true if source.starts_with?(prefix) && source[prefix.size] == '.'
+    end
+    false
+  end
+
+  private def update_matches(matches, level : Severity, backend : String)
+    if matches.nil?
+      {level: level, backend: backend}
+    elsif matches.is_a?(NamedTuple)
```
Sija:
```cr
case matches
when Nil
  {level: level, backend: backend}
when NamedTuple
  # ...
```

src/log/main.cr
```diff
@@ -0,0 +1,104 @@
+class Log
+  # Creates a `Log` for the given source.
+  # If *level* is given, it will override the configuration.
+  def self.for(source : String, level : Severity? = nil) : Log
+    log = builder.for(source)
+    log.level = level if level
+    log
+  end
+
```
Sija: Some sugar

```cr
def self.for(klass : Class, level : Severity? = nil) : Log
  for(klass.name.underscore, level)
end
```

```cr
class Foo::Bar
  Log = ::Log.for(self) # => #<Log: ... @source="foo.bar">
end
```

would be nice :)

src/log.cr
```diff
@@ -0,0 +1,121 @@
...
+# end
+# ```
+#
+# That way, any `Log.info` call within the `DB` module will use the `db` source. And not the top-level `::Log.info`.
+#
+# Sources can be nested. Continuing the last example, to declare a `Log` constant `db.pool` source you can do as follows:
+#
+# ```
+# class DB::Pool
+#   Log = DB::Log.get("pool") # => Log for db.pool source
```
vlazar: Should it be
```suggestion
#   Log = DB::Log.for("pool") # => Log for db.pool source
```

src/log/memory_backend.cr
```diff
@@ -0,0 +1,10 @@
+# A `Log::Backend` suitable for testing. It is registered as `memory`.
+#
+@[Log::RegisterBackend("memory")]
+class Log::MemoryBackend < Log::Backend
```
vlazar: I'm sure the annotation name is carefully chosen but why not just use a simpler one:

```crystal
@[Log::Backend("memory")]
class Log::MemoryBackend < Log::Backend
```

src/log/entry.cr
```diff
@@ -0,0 +1,23 @@
...
+  None
+end
+
+struct Log::Entry
+  getter source : String
+  getter severity : Severity
+  getter message : String
+  getter timestamp = Time.local
+  getter context = Log.context
+  getter exception : Exception?
```
willhbr: Any chance of including file/line information here? (from #8810) ☺️

src/log/memory_backend.cr
asterite: Well, there's already a class named like that...

ysbaddaden:
My main concern is the YAML file, for which I have some questions:

1. Is the YAML configuration an optional nicety, or is it required? It looks like _all_ crystal applications that `require "log"` will now include this YAML parser and depend on `libyaml` which may not be needed (e.g. small CLI tools).

2. Can the YAML file to load be configured _programmatically_? A framework will likely prefer to have a `./config/log.yml` for example, instead of having an out of place `./crystal_log.yml` —and not have to meddle with environment variables.

3. Can `Log` be configured programmatically instead? Seems like it, but Log::Config is only about YAML so that's weird.

straight-shoota:
@ysbaddaden 

Re 1. the idea was that `require "log"` doesn't include the config parser by default and you need `require "log/config"` explicitly. It seems that the implementation behaves like that. `log/config.cr` is not required in `log.cr`.

Re 2. I guess you could do `ENV["CRYSTAL_LOG_CONFIG"] = "./config/log.yml"` in user code, but having a direct API to specify the config location would be preferable. It should also an option to pass in the configuration directly without the setup code trying to read any files.

I agree with 3. There should be an usable Crystal API for configuring `Log`. Not every application needs flexible configuration from an external source.

asterite:
Where is `Log::Context` used?

I see it can be set, read, etc., but I don't see it used. Also, when I do `Log.info { ... }` I can't see it in the output.

Is it there so that formatters can optionally use it? If so, maybe provide an example of such formatter?

src/log/entry.cr
bcardiff: You can add it as context information if needed.

Explicitly or with a method that will do this.

```crystal
Log.context.using do
  Log.context.set file: __FILE__, line: __LINE__
  Log.info { ... }
end # remove file/line context keys
```

I'm not sure every entry should have information. But I am open to discuss the idea.

bcardiff:
@ysbaddaden The `Log::Builder` allows programmatic configuration or by redefining the `Log.setup(config : Configuration)` method you can setup the whole configuration you want. No yaml and no env vars is required. But it needs to be in that method if you want to be able to hook into the initialization of constants I am propossing to use. Using the `Log::Builder` allows use create loggers that are not reachable from the `Log.for` method.

@asterite I will add the output of the context to the stdio backend. Yes, there will be a `Symbol#to_s` I guess that is what you are thinking about it.

bcardiff:
@straight-shoota / @ysbaddaden  . regarding 2. There are some method that are defined in `log/config.cr` that are `:nodoc:`, by overriding the `Log.setup` you could reuse them to specify custom sources for the yaml configuration. 

In particular overriding `Log.yaml_config_path` you choose another path. And by overriding `Log.setup` you can call `setup_from_yaml` with whatever you want. If these two scenarios work I can make them public and add some documentation to them.

src/log/context.cr
asterite: What I mean is that you can't do something like:

```crystal
h = {:foo => 1, :bar => 2}
Log.context.set h
```

Is that intentional?

straight-shoota:
I don't think customizing the setup process should require reopening stdlib classes to redefine methods.
I'd prefer if `Log.yaml_config_path` was a property and allows to assign a custom path or disable YAML file loading.
Can we make `Log.setup` use a proc? And probably rename it, so that the public API method `Log.setup` receives a proc, like this:
```cr
require "log"
Log.setup do |builder|
  builder.bind "*", :info, SomeBackend.new
end
```

bcardiff:
@straight-shoota 

> Can we make Log.setup use a proc? 

The challenge is how the logs initialized in shards or std-lib will grab that configuration. If it is a method with proc the evaluation order is not error prone. We could have some macro to hide the redefinition of the method. But I don't think a plain proc will be enough.

```cr
require "myshard"
# shard
module MyShard
  Log = ::Log.for("myshard") # use the current configuration
end

# my app
Log.setup do |builder|
 # ... but the MyShard::Log could be already initialized at this point
end
```

straight-shoota:
I see. Making `Log.setup` a macro that redefines the method might be a good idea :+1:

bcardiff:
@straight-shoota the last commit changes `Log.setup` to macro. Is a bit uglier than I expected.

* there is a need to tell if the configuration was done (truthy/falsey block value)
* since it is a macro, the block argument needs to used as `{{config}}`. (plain `config` work but only because the underlying declared method uses that as a method argument and the block would become superfluous)

asterite:
As I understand it, right now the setup needs to happen before you call `Log.for(...)`, right?

Why is that so? Can't `Log.for("...")` just give you some logger registered with that name, and setup configures them at any point in time?

bcardiff:
How you ensure the logs are configured properly for the very first entry?

If we support reconfiguration of existing loggers we can mitigate a bit the issue, but still some logs might be emitted to invalid sources. So I think we need to setup loggers before any `Log.for` is returned.

asterite:
> How you ensure the logs are configured properly for the very first entry?

If we allow changing the log configuration at runtime, then everything is properly configured for the very first entry: with default values (I guess stdout).

I don't think it matters much if a few log lines are written to an incorrect place. Usually one would configure logging before the main code.

That's why I also think the Log configuration should be explicit in code, with something like:

```crystal
Log.configure do |config|
  # ... 
end
```

If it's hidden behind a `require "log/config"` it's not clear.

(the above snippet can also be `Log.configure_from_yaml_file("...")`.

asterite:
In general, explicit > implicit.

paulcsmith:
I agree the explicit would be clearer and more customizable. For example you could remove the need to set an ENV var or have a particular yaml file. You could do `Log.configure_from_yaml(ENV["MY_CUSTOM_NAME"])` or an explicit string like @asterite. Nice and clean, less implicit code to learn, easier to document

src/log/entry.cr
asterite: But why not add file and line information to every Entry?

It's as simple as letting `info`, `debug` and other methods accept `file = __FILE__, line = __LINE__`.

On the other hand, you can usually find where a log line happens by searching that text in code, so I'm not sure how useful it is to include it in every log line.

paulcsmith: > On the other hand, you can usually find where a log line happens by searching that text in code, so I'm not sure how useful it is to include it in every log line.

@asterite Yeah that's my main reason for not wanting it in every log line. For example, if a shard outputs an error, I don't want/need to know which line it is from, or if it is just general info I probably don't need it either. 

Another option might be to always include file/line like you suggested, but make whether that info is actual set in the context an option:

```crystal
Log.log_file_and_line = true
```

And then the block version if you just want it temporarily as suggested above. That might be pretty cool and very useful if you really do need to figure out where a message is coming from and can't find it

RX14:
The datum macro seems very premature. Could it be introduced as a seperate PR, after this one and see what code it cleans up?

src/log/entry.cr
willhbr: The default formatter doesn't have to output this information, but being able access this information in a custom formatter is super useful - especially if the log message is somewhat generic.

The alternative is having to maintain my own wrapper around the standard logger, which is less than ideal.

src/log/main.cr
```diff
@@ -0,0 +1,130 @@
...
+  #   {{config}}.add_backend("stdio", "stdio", Log::Backend::Config.new)
+  #   {{config}}.bind("*", :debug, "stdio")
+  #   true
+  # end
+  # ```
+  #
+  macro setup
+    class ::Log
+      def self.setup_impl(config : Configuration)
+        done = {{yield :config.id}}
```
RX14: You can introspect the variable name of the first block arg, declare that var, and then paste the block body. Then no `{{config}}` required.

src/log/entry.cr
paulcsmith: @willhbr That is a very nice idea! May as well always have it available and let the formatter decide. The default formatters could have a configurable option for outputting it (or not, someone can add it later)

src/log/log.cr
```diff
@@ -0,0 +1,37 @@
...
+                               debug:   Severity::Debug,
+                               verbose: Severity::Verbose,
+                               info:    Severity::Info,
+                               warn:    Severity::Warning,
+                               error:   Severity::Error,
+                               fatal:   Severity::Fatal,
+                             } %}
+
+    # Logs a message if the logger's current severity is lower or equal to `{{severity}}`.
+    def {{method.id}}(*, exception : Exception? = nil)
```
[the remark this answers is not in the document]
j8r: It will prevent using a macro :+1:

src/log.cr
```diff
@@ -0,0 +1,133 @@
...
+#
+# How to configure in general the logging can vary but the std-lib comes with three alternatives.
+#
+# ### Configure logging explicitly in the code
+#
+# Use `Log.setup` macro to configure the logging manually.
+#
+# ```
+# Log.setup do |config|
+#   {{config}}.add_backend("stdio", "stdio", Log::Backend::Config.new)
```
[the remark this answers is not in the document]
bcardiff: Log.setup removed. Yet I agree. It's ugly.

src/log/main.cr
bcardiff: 👻 but is a pattern it might work for another scenario.

src/log/log.cr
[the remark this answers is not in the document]
bcardiff: Why would you need a parametric severity when emitting a log?

That will also prevent doing some optimizations like removing in compile time all `.debug` methods for example.

bcardiff:
With the last push I removed `Log::Config`, all the yaml related support and the `RegisterBackend` annotation.

Now a `Log::Builder` is exposed in `Log.builder` that is the one used by the top `Log.for`. It allows reconfiguration of existing logs via `Builder#bind` and `Builder#clear` methods.

The Logs are tracked via weak refs in case to avoid leaking them.

The builder is thread-safe.

The top level docs are updated

---

* @rx14 the datum is extracted from JSON and YAML any. It can replace both actually. But is a private macro, to avoid making it public. It can evolve there I think.

* regarding file/line, it could be added to entry, the string should be already in the binary so there is not much runtime penalty to have them. But I am still hesitant to add it to every entry. Wouldn't that leak to much information of the machine used to build the source?

src/log/builder.cr
```diff
@@ -0,0 +1,113 @@
...
+    @logs.each_value do |log_ref|
+      log = log_ref.value
+      yield log if log
+      # TODO should remove entry if log.nil?
+    end
+  end
+
+  # :nodoc:
+  private def append_backend(log : Log, level : Severity, backend : Backend)
+    current_backed = log.backend
```
Sija: typo:

```suggestion
    current_backend = log.backend
```

src/log/log.cr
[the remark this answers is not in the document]
RX14: > That will also prevent doing some optimizations like removing in compile time all `.debug` methods for example.

Those optimizations don't need to be present when using parametric severity.

We can wait for someone to complain, if you want, though.

Sija:
> * regarding file/line, it could be added to entry, the string should be already in the binary so there is not much runtime penalty to have them. But I am still hesitant to add it to every entry. Wouldn't that leak to much information of the machine used to build the source?

IMO the optimum solution would be to add 'em to the `Log::Entry` and leave the decision whether it's to be shown, up to the formatter.

src/log/log.cr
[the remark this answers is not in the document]
asterite: Every log library or framework that I used, it's `log.info`, `log.debug`. I see no reason why we need to write `Log.log(:info) { "Hello" }` instead of `Log.info { "hello" }`.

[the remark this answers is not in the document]
asterite: And then the log level you use is never parameterized in each invocation.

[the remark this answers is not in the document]
paulcsmith: I think it'd be nice to have a generic `log` , but having `debug/etc` is a must (IMO). The log that accepts severity is nice when you have a dynamic value. For example you have an ENV, or a value stored in Redis/Memory that is used to determine the level. Another example is determining the log level based on something else that can't be determined at compile time

```crystal
def run_query
  log_level = if query.was_slow?
    :info
  else
     :debug
   end
   Log.write(log_level, query.to_sql)
end
```

Definitely helpful to have if needed, but the concrete methods should be preferred in most cases IMO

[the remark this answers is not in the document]
Sija: Hmm, I thought @RX14 was talking about `abstract def log` in addition to the macro magic, **not** meant to replace it altogether.

straight-shoota:
> Wouldn't that leak to much information of the machine used to build the source?

If the log contains an exception, it's going to be there anyways.

bcardiff:
Today's additions are

* Add `Log.for(type)` overload (and fix some corner cases on `for`)
* Allow to extend the context via a hash and named tuples argument
* Make the context keys String to allow scenarios where the keys are dynamic. 

Regarding the file/line information with waj we settle on not include it. I couldn't find a logging framework that does that. There are workaround. And is not core to the feature.

Redarging a generic write method in `Log` it is not needed initially. There are workarounds and is not something common.

Are the API and documenation good enough for approval?

src/log/context.cr
bcardiff: fixed

src/log/main.cr
bcardiff: done

src/log/context.cr
asterite: Cool, now one can pass a hash.

My question was more about why Symbol and not String.

I think we should avoid using Symbol, and eventually remove it from the language. Symbol is a good thing in Ruby, but in Crystal I can't find a really good use case...

bcardiff: I still like them. Is a separate domain of simple values for me.

But allowing the backend to filter directly keys based on string operations, and allowing dynamic keys was the selling point here.

bcardiff:
I squashed the history and rebased on master.

paulcsmith:
Looks awesome @bcardiff!

src/log/log.cr
```diff
@@ -0,0 +1,54 @@
...
+      return unless backend = @backend
+      severity = Severity.new({{severity}})
+      return unless level <= severity
+      entry = Entry.new @source, severity, yield.to_s, exception
```
Sija:
```suggestion
      message = yield.to_s
      entry = Entry.new @source, severity, message, exception
```

bcardiff: done

z64:
Just asking for posterity: this should supersede the previous draft #6901?

Really glad to see the logger being updated! Thank you for working on this.

bcardiff:
Yes @z64, thanks for pointing that out.

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -1,23 +1,29 @@
...
 require "ecr/macros"
 require "compiler/crystal/tools/formatter"
 require "compiler/crystal/tools/doc/markdown"
 
 module Crystal::Playground
+  Log = begin
+    builder = ::Log::Builder.new
+    builder.bind("crystal.playground", :warning, ::Log::StdioBackend.new)
+    builder.for("crystal.playground")
+  end
```
waj: Why isn't this just `Log = ::Log.for("crystal.playground", :warning)`?
This doesn't look like the pattern we want the users to embrace

src/log/log.cr
```diff
@@ -0,0 +1,55 @@
...
+                               info:    Severity::Info,
+                               warn:    Severity::Warning,
+                               error:   Severity::Error,
+                               fatal:   Severity::Fatal,
+                             } %}
+
+    # Logs a message if the logger's current severity is lower or equal to `{{severity}}`.
+    def {{method.id}}(*, exception : Exception? = nil)
+      return unless backend = @backend
+      severity = Severity.new({{severity}})
```
waj: Ins't `{{severity}}` a `Severity` already?

asterite: I thought the same, but because they are enum constants the macro execution outputs their value, like 1, 2, etc. It's super confusing but I don't know how to "fix" it...

src/compiler/crystal/tools/playground/server.cr
bcardiff: The playground is the only compiler module that logs something. `crystal play` logs warnings only and `crystal play -v` changes the log level to debug.

The there is a need for a custom logger initialization to keep that.

This is the change to keep the usage of the logger isolated in the playground.

The alternative would be to Use `Log = ::Log.for("crystal.playground")`. Define that the compiler configs the logger with some policy. Allow the playground cli to change the log level of `crystal.playground` logger.

We can do that if prefered.

bcardiff: The commit https://github.com/crystal-lang/crystal/pull/8847/commits/56d0e6aa83045adc54499dc9cb79c19b382db359 enables the logger on the whole compiler only for `crystal.*` sources and warning level by default. The `crystal play -v` will change the playground log level only. The CRYSTAL_LOG_LEVEL/CRYSTAL_LOG_SOURCES can be used to tweak the logger configuration further.

src/log/main.cr
```diff
@@ -0,0 +1,142 @@
...
+  def context : Log::Context
+    Log.context
+  end
+
+  # Sets the current fiber logging context.
+  def context=(value : Log::Context)
+    Log.context = value
+  end
+
+  class Context
```
RX14: Why is the `Context` class split between two files? It's a little confusing...

src/log/main.cr
```diff
@@ -0,0 +1,142 @@
...
+      context = fiber.logging_context
+      fiber.logging_context = context.merge(values)
+    end
+
+    # Method to save and restore the current logging context.
+    #
+    # ```
+    # Log.context.set a: 1
+    # Log.info { "message with {"a" => 1} context" }
+    # Log.using do
```
RX14: `Log.context.using`.

src/log/main.cr
bcardiff: A matter of taste. The `log/main.cr` describes the api the user will use. I found that more clear to have a better perspective and express the intention of the methods.

bcardiff:
Rebased on master. @RX14 I addressed your feedback on the last 4 commits.

spec/std/log/context_spec.cr
```diff
@@ -0,0 +1,119 @@
+require "spec"
+require "log"
+
+def c(value)
```
straight-shoota: These helper methods should be private to the spec file to avoid global pollution.

spec/std/log/context_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+  Log::Context.new(value)
+end
+
+describe Log::Context do
+  describe "initialize" do
```
straight-shoota: This block should be `it`. Ditto to the following ones.

spec/std/log/context_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+    context = c({a: 1})
+    other = context.as_h
+    other["a"] = c(2)
+
+    other.should eq({"a" => c(2)})
+    context.should eq(c({a: 1}))
+  end
+
+  describe "merge" do
+    c({a: 1}).merge(c({b: 2})).should eq(c({a: 1, b: 2}))
```
straight-shoota: Can we get an example of override behaviour? 
```cr
c({a: 1, b: 3}).merge(c({b: 2})).should eq(c({a: 1, b: 2}))
```

spec/std/log/context_spec.cr
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  describe "implicit context" do
+    it "can be set and cleared" do
+      Log.context.clear
```
straight-shoota: Maybe add `before_each` and `after_each` hooks for clearing the context?

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+#
+# ```
+# Log.warn(exception: e) { "Oh no!" }
+# ```
+#
+# The block is only evaluated if the current message is to be emitted to some `Log::Backend`.
+#
+# To add structured information to the message you can use the `Log::Context`.
+#
+# When creating logs messages they belong to a _source_. If the top-level `Log` is used
```
straight-shoota:
```suggestion
# When creating log messages they belong to a _source_. If the top-level `Log` is used
```

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+#
+# ```
+# module DB
+#   Log = ::Log.for(self) # => Log for db source
+# end
+# ```
+#
+# ### Configure logging explicitly in the code
+#
+# Use `Log.builder` to indicate the which sources should go to which backends.
```
straight-shoota:
```suggestion
# Use `Log.builder` to indicate which sources should go to which backends.
```

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+#
+# You can indicate actual sources or patterns.
+#
+# * the empty string will show only the top-level source
```
straight-shoota:
```suggestion
# * the empty string matches only the top-level source
```
Ditto below.

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+# * `*` will show all the sources
+# * `foo.bar.*` will show `foo.bar` and every nested source
+# * `foo.bar` will show `foo.bar`, but not its nested sources
+#
+#
+# The following configuration will setup for all sources to emit
+# warnings (or higher) to `STDOUT`, allow any of the `db.*` and
+# nested source to emit debug (or higher), and to also emit for all
+# sources errors (or higher) to an elasticsearch search.
```
straight-shoota:
```suggestion
# sources errors (or higher) to an elasticsearch backend.
```

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+#
+# ### Configure logging from environment variables
+#
+# Include the following line to allow configuration from environment variables.
+#
+# ```
+# Log.setup_from_env
+# ```
+#
+# The `CRYSTAL_LOG_LEVEL`, `CRYSTAL_LOG_SOURCES` environment variables are used to indicate
```
straight-shoota:
```suggestion
# The environment variables `CRYSTAL_LOG_LEVEL` and `CRYSTAL_LOG_SOURCES` are used to indicate
```

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+# which severity level emit (defaults to `INFO`, use `NONE` to skip all messages) and to restrict
```
straight-shoota:
```suggestion
# which severity level to emit (defaults to `INFO`; use `NONE` to skip all messages) and to restrict
```

src/log.cr
```diff
@@ -0,0 +1,128 @@
...
+# which sources you are interested in.
+#
+# The valid values for `CRYSTAL_LOG_SOURCES` are:
+#
+# * the empty string will show only the top-level source (default)
```
straight-shoota: Wouldn't it be better to refer to the above list and only clarify the default and that multiple values can be concatenated by comma?

src/log/backend.cr
```diff
@@ -0,0 +1,13 @@
+require "crystal/datum"
+
+# Base class for all backends.
+abstract class Log::Backend
+  # Writes the *entry* to this backend
```
straight-shoota:
```suggestion
  # Writes the *entry* to this backend.
```

src/log/backend.cr
```diff
@@ -0,0 +1,13 @@
...
+  abstract def write(entry : Entry)
+
+  # Closes underlying resources used by this backend
```
straight-shoota:
```suggestion
  # Closes underlying resources used by this backend.
```

src/log/broadcast_backend.cr
```diff
@@ -0,0 +1,33 @@
+# A backend that broadcast to others backends.
+# Each of the referenced backends may have a different severity level filter.
+#
+# When this backend level is set that level setting takes precedence
+# over the severity filter of each referenced backend.
+#
+# This backend is not to be used explictly. It used by `Log::Builder` configuration
+# to allow a given source to emit to multiple backends.
+class Log::BroadcastBackend < Log::Backend
```
straight-shoota: The IO implementation with similar (broadcast) behaviour is called `MultiWriter`. Should we consider naming this one `MultiBackend` for consistency? `BroadcastBackend` seems more specific, though. So it might be a better choice after all.

src/log/broadcast_backend.cr
```diff
@@ -0,0 +1,33 @@
+# A backend that broadcast to others backends.
+# Each of the referenced backends may have a different severity level filter.
+#
+# When this backend level is set that level setting takes precedence
+# over the severity filter of each referenced backend.
```
straight-shoota: I don't quite follow what this sentence means.

The implementation looks like if `BroadcastBackend#level` is not nil, it overrides any levels configured for the child backends. Why is that so? It feels a bit surprising.

src/log/broadcast_backend.cr
```diff
@@ -0,0 +1,33 @@
...
+#
+# This backend is not to be used explictly. It used by `Log::Builder` configuration
```
straight-shoota:
```suggestion
# This backend is not to be used explictly. It is used by `Log::Builder` configuration
```

src/log.cr
bcardiff: Can we have relative links in docs. Although the fact that the environment variable and the builder supports the same syntax is convenient, but they are different API.

src/log/broadcast_backend.cr
bcardiff: I think broadcast is a better description. :-) Although it is unlikely to be used publicly in the current state of the builder.

src/log/broadcast_backend.cr
bcardiff: It is to allow programatic override of the level from a Log instance. See `Log#level=`.

src/log/builder.cr
```diff
@@ -0,0 +1,114 @@
...
+        @logs[source] = WeakRef.new(log)
+      end
+
+      log
+    end
+  end
+
+  # :nodoc:
+  private def each_log
+    @logs.reject! { |_, log_ref| log_ref.value.nil? }
```
Sija: This is unexpected side-effect.

src/log/broadcast_backend.cr
```diff
@@ -0,0 +1,31 @@
+# A backend that broadcast to others backends.
+# Each of the referenced backends may have a different severity level filter.
+#
+# When this backend level is set that level setting takes precedence
+# over the severity filter of each referenced backend.
+#
+# This backend is not to be used explictly. It is used by `Log::Builder` configuration
+# to allow a given source to emit to multiple backends.
+class Log::BroadcastBackend < Log::Backend
+  property level : Severity? = nil
```
Sija:
```suggestion
  property level : Severity?
```

straight-shoota: Why?

src/log/builder.cr
bcardiff: I don’t follow

Sija: Why calling  `each_log` mutates `@logs`?

src/log/broadcast_backend.cr
Sija: Because it's redundant?

src/log/builder.cr
asterite: I'm almost sure it's because they are weak refs. So this is returning only the valid ones, while making sure the invalid ones are deleted. I think this fine.

Sija: @asterite I'm not questioning the logic, but the side-effect in itself - `each` (and friends) is suppose to enumerate the values, not modify 'em... it could be easily done without modifying `@logs` and if that's intended there should be a better place to do it, not in here...

asterite: You mean like moving this line to a helper function?

The idea of this function is to iterate the log entries. Right before doing that it makes sure to remove invalid entries. Why is that wrong?

asterite: Would you move that line or a helper function to each call site?

Sija: > Would you move that line or a helper function to each call site?

Moving that into a helper method would IMO be a cleaner design, yes. _each call site_ counts at exactly 2 occurrences, so it's not like that would be adding a ton of overhead anyway.

asterite: The idea is for call sites to not have to know all of this.

I guess it's double checked because weak refs can become nil at any point.

straight-shoota: @Sija I don't think it matters. Whether `@logs` contains deleted references or not is only internal information. To the outside both states count as non-existing.

src/log/broadcast_backend.cr
straight-shoota: That may be, but more importantly it's explicit about the default value.

src/log/builder.cr
Sija: Fair 'nuff, thanks for bearing with me.

src/log/broadcast_backend.cr
Sija: `Explicit > Implicit` argument, let it be.

src/log/builder.cr
Sija: One last question: is it needed at all, given the check below?

src/compiler/crystal/tools/playground/server.cr
RX14: Thanks, I like this solution!

src/log/builder.cr
```diff
@@ -61,11 +61,12 @@ class Log::Builder
   end
 
   # :nodoc:
-  def each_log
+  private def each_log
+    @logs.reject! { |_, log_ref| log_ref.value.nil? }
```
RX14: Can't we do this deletion while iterating? I'd much rather do it in one loop if possible.

src/log.cr
```diff
@@ -57,8 +57,9 @@
 # A `Log` will emit the messages to the `Log::Backends` attached to it as long as
 # the configured severity filter `level` permits it.
 #
-# Logs can also be created from a type directly. The full name without generic arguments
-# will be used as a source. So for the type `Foo::Bar(Baz)` the source `foo.bar` will be used.
+# Logs can also be created from a type directly. For the type `DB::Pool` the source `db.pool` will be used
+# The full name without generic arguments will be used as a source. So for the type `Foo::Bar(Baz)`
+# the source `foo.bar` will be used.
```
RX14: nit: The order of these sentences are weird: example, explanation, example.

src/log.cr
```diff
@@ -112,6 +113,11 @@
 #
 # The logs are emitted to `STDOUT` using a `Log::StdioBackend`.
 #
+# If `./bin/app` uses `Log.setup_from_env` you can tweak the logging as
```
RX14: I'd rather say "the executable" than `./bin/app`.

src/log/builder.cr
bcardiff: If it would be an Array I would say that iterating backwards should work (or playing with indices). I would like mutating iterators :-P

Since it is a Hash, iterating it and mutating it is not a nice option. So either as is, or have a copy of all the keys to iterate.

I ended up with the simpler code.

straight-shoota: I think this is fine. If it shows to be a performance issue, we can improve it later.

src/log.cr
bcardiff: done

src/log.cr
bcardiff: done

bcardiff:
```cr
Log.info "approval pending" unless feedback
```

src/log/log.cr
straight-shoota: Alternative would be to use strings and macro id.

src/log/main.cr
```diff
@@ -0,0 +1,142 @@
...
+    #   Log.context.set b: 2
+    #   Log.info { "message with {"a" => 1, "b" => 2} context" }
+    # end
+    # Log.info { "message with {"a" => 1} context" }
+    # ```
+    def using
```
straight-shoota: "using" is very unspecific. I'd call it `with_context`.

src/log/stdio_backend.cr
```diff
@@ -0,0 +1,40 @@
...
+        format(entry, stdout)
+      end
+      stdout.puts
+      stdout.flush
+    end
+  end
+
+  # Emits the *entry* to the given *io*.
+  def format(entry : Entry, io : IO)
+    label = entry.severity.to_s.upcase
```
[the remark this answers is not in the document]
asterite: Also, doing `upcase` in every log instruction will allocate a separate string for this which is really inefficient. These strings should be stored somewhere and reused.

src/log/stdio_backend.cr
```diff
@@ -0,0 +1,40 @@
...
+    io << label[0] << ", [" << entry.timestamp << " #" << Process.pid << "] "
```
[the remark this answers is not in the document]
RX14: RFC3339 isn't immediately readable to me, because of the `T` there is no visual separation between the components.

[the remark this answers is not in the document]
straight-shoota: RFC 3339 allows the date and time components to be separated by a space character.

src/log/stdio_backend.cr
```diff
@@ -0,0 +1,40 @@
+# A `Log::Backend` that emits to STDOUT.
+class Log::StdioBackend < Log::Backend
```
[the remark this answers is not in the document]
bcardiff: My idea for `StdioBackend` was to allow usage of colors and maybe to split the output in stdout/stderr depending on the severity. This two ideas don't fit in a plain `IOBackend`.

Maybe a `IOBackend` could be created and use it in the StdioBackend implementation.

Also, I would not use `IOBackend` for logging to files. In a `FileBackend` we will likely want options to rotate files.

So since both scenarios include their own options I would rather avoid creating the `IOBackend` until we can iterate with more examples to get it right.

The implementation of `StdioBackend` can evolve and keep the API simple and aimed to be used only in STDOUT/STDERR.

src/log/main.cr
bcardiff: But it is used as `Log.context.using do`
I am used that using handles resources or disposable since .NET.

`Log.context.with` could be an alternative, but it reads more weird to me. And is using a keyword as a method name. I prefer to avoid if possible.

Meanwhile I notice the sample had some errors due to symbol to string refactor... fixing those.

bcardiff:
@straight-shoota I addressed most of the feedback and argued about the ones I don't agree with :-) . Thanks for the review.

src/log/stdio_backend.cr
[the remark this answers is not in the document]
straight-shoota: > usage of colors

That seems like the responsibility of a formatter, not the backend.

> split the output in stdout/stderr depending on the severity

That should be implemented with separate backend instances for stderr and stdout and severity filters, not some specific internal behaviour. With such a nice framework for compositing logger backends, why not use it?

> In a FileBackend we will likely want options to rotate files.

Certainly not for a basic implementation. More advanced features could be built upon `IOBackend`.

Currently, there's nothing special about `StdioBackend` that wouldn't fit in a generic `IOBackend`.
For the sake of simplicity, I'd prefer to rename it to `IOBackend`. Should we later need additional features we can add specialized `StdioBackend` and `FileBackend` implementations. But for now there's really no reason for that.

src/log/main.cr
straight-shoota: What about `Log.with_context do`?

src/log/entry.cr
```diff
@@ -8,6 +8,20 @@ enum Log::Severity
...
+    case self
+    when Debug  ; "DEBUG"
+    when Verbose; "VERBOSE"
+    when Info   ; "INFO"
+    when Warning; "WARNING"
+    when Error  ; "ERROR"
+    when Fatal  ; "FATAL"
+    when None   ; "NONE"
+    else
+      raise NotImplementedError.new("Log::Severity#label")
```
straight-shoota: It's not really relevant, but I don't think should use `NotImplementedError`. This should just be `raise "unreachable"`.

src/log/main.cr
bcardiff: I can live with that :-)

src/log/stdio_backend.cr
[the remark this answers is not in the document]
bcardiff: done

src/log/main.cr
bcardiff: done

src/log/entry.cr
asterite: Better yet: this shouldn't be needed at all, cough, cough.

src/log/io_backend.cr
```diff
@@ -0,0 +1,43 @@
+# A `Log::Backend` that emits to an `IO` (defaults to STDOUT).
+class Log::IOBackend < Log::Backend
+  property io : IO = STDOUT
+  property progname : String
+  property formatter : Formatter? = nil
+
+  def initialize
```
Sija:
```suggestion
  def initialize(@io = STDOUT)
```

Sija:
```suggestion
  property io : IO
  property progname : String
  property formatter : Formatter? = nil

  def initialize(@io = STDOUT)
```

wontruefree:
Has there been any benchmarking on these changes?
With using blocks as the main api for logging I could see higher memory usage and possible performance hit.

Sija:
> With using blocks as the main api for logging I could see higher memory usage and possible performance hit.

@wontruefree Blocks are being inlined, so where would you see it coming from?

wontruefree:
@Sija I am sorry just to be clear I have not done benchmarking I was asking if any benchmarking was done.  I thought that might be a cause of some performance hit but I have not verified it.

bcardiff:
@wontruefree , no hard benchmark was done. Yet:
* the blocks that no impose a penalty as @Sija mentioned
* once the `Log` instances are built there is little runtime code to emit the entry, all the objects have direct references; there are no lookups
* the IOBackend default format (now) avoids all string allocations :-P

Maybe the builder could do some more intelligent work to save logs in a tree like structure rather than iterating them all to check if the bind call should reconfigure a specific logger. But after the bind are done there is no more penalty and that should happen while starting the app in most of the cases.

Sija:
@wontruefree fwiw, my dead-simple 2 min tests are showing that the new logger is about ~10% faster.

```cr
require "benchmark"
require "logger"
require "log"

logger = Logger.new(STDOUT)
logger.level = :info

Log.setup_from_env(
  backend: Log::IOBackend.new(STDOUT),
  level: "INFO"
)

Benchmark.ips do |x|
  x.report("Logger.info") do
    logger.info "foo"
  end

  x.report("Log.info") do
    Log.info { "foo" }
  end
end
```

```
Logger.info  26.57k ( 37.63µs) (±11.50%)  32.0B/op   1.07× slower
   Log.info  28.43k ( 35.18µs) (±14.70%)  32.0B/op        fastest
```

wontruefree:
@Sija that is amazing and now what I expected

src/log/env_config.cr
```diff
@@ -0,0 +1,17 @@
+class Log
+  # Setups *builder* based on `CRYSTAL_LOG_LEVEL` and `CRYSTAL_LOG_SOURCES`
+  # environment variables.
+  def self.setup_from_env(*, builder : Log::Builder = Log.builder,
```
Sija: Shouldn't that be called `setup` with a `NOTE` in docs about using `ENV` as default argument values?
I mean there's nothing in the method itself related to the `ENV`, aside of the before-mentioned default values...

wontruefree: Is there another way to setup the log configuration?  It seems weird to have to pass strings into the config API.

bcardiff: I think calling it setup_from_env is more clear regarding what it does. But no strong feelings, more a evolution of the code regarding env vs yaml. 

@straight-shoota any preferences?

src/log/env_config.cr
```diff
@@ -0,0 +1,17 @@
...
+                          backend = Log::IOBackend.new)
+    builder.clear
+
+    level = Log::Severity.parse(level)
+    sources.split(',', remove_empty: false) do |source|
+      source = source.strip
+
+      builder.bind(source, level, backend)
+    end
+  end
```
Sija: Perhaps sth like:

```suggestion
  # ...
  def self.setup(*, builder : Log::Builder = Log.builder,
                    level : Log::Severity = :info,
                    sources : Array(String) = %w[],
                    backend : Log::Backend = Log::IOBackend.new)
    builder.clear

    sources = [""] if sources.empty?
    sources.each do |source|
      builder.bind(source, level, backend)
    end
  end

  # Setups *builder* based on `CRYSTAL_LOG_LEVEL` and `CRYSTAL_LOG_SOURCES`
  # environment variables.
  def self.setup_from_env(*, builder : Log::Builder = Log.builder,
                          level_key : String = "CRYSTAL_LOG_LEVEL",
                          sources_key : String = "CRYSTAL_LOG_SOURCES",
                          backend : Log::Backend = Log::IOBackend.new)
    level = ENV.fetch(level_key, "INFO")
    level = Log::Severity.parse(level)

    sources = %w[]
    env_sources = ENV.fetch(sources_key, "")
    env_sources.split(',', remove_empty: true) do |source|
      sources << source.strip
    end

    setup(builder: builder, level: level, sources: sources, backend: backend)
  end
```

src/log/env_config.cr
straight-shoota: This is not a *generic* setup method, but a specific implementation based on confguration values from env. So I agree that the current name fits the behaviour.

src/log/env_config.cr
straight-shoota: I see hardly any benefit in separating the behaviour into two separate methods. I can only imagine the case when you want to achieve something similar as `setup_from_env` but with type safe arguments? That doesn't seem particularly more useful to the alternative. Chances are that you might need to tweak the implementation sooner or later. Anyway, this is really trivial to implement, so not a big deal to have partial duplication in application code.

src/log/env_config.cr
Sija: @straight-shoota I disagree. The implementation is totally generic, only relation to the `ENV` is via argument defaults. Also, @wontruefree raised a valid issue - how would you setup logging in a generic manner?

The way it is right now:

```cr
Log.setup_from_env(
  backend: Log::IOBackend.new(STDOUT),
  level: "INFO"
)
```

Makes lil' sense to me, logic-wise (no `ENV` relation at all here despite using `setup_from_env` method).

straight-shoota: I don't follow. Why would you want to use this a method in the first place? It's very specific for setting up logging from environment variables. That's it's sole purpose. If you want something else (for example hard coding the logger setup), there's no point in using this method.

This is a much simpler way to set things up as in your example. No need for calling `#setup_from_env`.
```cr
Log.builder.bind("", :info, Log::IO::Backend.new(STDOUT))
```

src/log/stdio_backend.cr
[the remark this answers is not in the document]
RX14: Oh, neat! Do we have that option in the stdlib?

src/log/entry.cr
```diff
@@ -0,0 +1,37 @@
...
+  Info
+  Warning
+  Error
+  Fatal
+  # Used only for severity level filter.
+  None
+
+  def label
+    case self
+    when Debug  ; "DEBUG"
```
RX14: Final nitpick: `when Debug then "DEBUG"` syntax exists :)

src/log/io_backend.cr
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  # Emits the *entry* to the given *io*.
+  def default_format(entry : Entry)
+    label = entry.severity.label
+    io << label[0] << ", ["
+    entry.timestamp.to_rfc3339(io)
+    io << " #" << Process.pid << "] "
+    io << label.rjust(7) << " -- " << @progname << ":" << entry.source << ": " << entry.message
```
asterite: This `label.rjust(7)` is the only bit that allocates memory when logging info.

Suggestion: just like we have a `label` method in `Severity`, introduce a new method `label_rjust7` or similar with the labels with necessary spaces. Maybe not make it public (nodoc), but it can be useful for other formatters too.

Then the performance is a bit better, but mainly there are no allocations which is really good for a logging system.

asterite: Actually, nevermind. It would be nice to have `rjust`, `ljust` and `center` write to an IO too. I'll try to implement that.

asterite: Done! #8923 is merged so you can rebase against master and use that io overload.

bcardiff: done 👍

src/log/entry.cr
bcardiff: done

bcardiff:
Thanks everybody for the feedback and the discussion on this topic. I'm happy with the outcome.

spec/std/log/io_backend_spec.cr
```diff
@@ -0,0 +1,91 @@
+require "spec"
+require "log"
+
+private def s(value : Log::Severity)
+  value
+end
+
+private def stdio_logger(*, stdout : IO, config = nil, source : String = "", progname : String? = nil)
```
bew: Should be named `io_logger`

src/log.cr
```diff
@@ -0,0 +1,133 @@
...
+#
+# Sources can be nested. Continuing the last example, to declare a `Log` constant `db.pool` source you can do as follows:
+#
+# ```
+# class DB::Pool
+#   Log = DB::Log.for("pool") # => Log for db.pool source
+# end
+# ```
+#
+# A `Log` will emit the messages to the `Log::Backends` attached to it as long as
```
bew: `` `Log::Backends` `` -> `` `Log::Backend`s ``

src/log/main.cr
```diff
@@ -0,0 +1,142 @@
...
+      extend_fiber_context(Fiber.current, Log::Context.new(values))
+    end
+
+    # :ditto:
+    def set(values : Hash(Symbol, V)) forall V
+      extend_fiber_context(Fiber.current, Log::Context.new(values))
+    end
+
+    # :ditto:
+    def set(values : NamedTuple) forall V
```
bew: forall is useless here (ditto below)

bcardiff:
Since the final state of the PR changed from the initial proposal, I wanted to submit an updated version of it. Although the docs are available as well.

---

## Generating logs

The are a couple of severity levels it can be used to emit logs: `debug`, `verbose`,`info`, `warn`, `error`, `fatal`.

Usually you will type `Log.[level] { "a message" }`

```crystal
Log.info { "Hi there!" }
```

**Note:** Using the result of the block for the message will allow the runtime to ignore possible conversions, string interpolations to build the message if it is not going be emitted.

The `Log` is a new top-level constant in the std-lib (`Logger` will be deprecated and removed).

When generating logs you will use a `Log` instance. Each `Log` instance is bound to a `source` or path. `Log.for(source : String) : Log` will allow you get a new instance. The top-level `Log` acts as an instance of an empty string source `Log`.

When defining a new shard or module you will define a `Log` constant for that namespace.

```crystal
module DB
  Log = ::Log.for("db")
end
```

Thanks to the relative resolution of constants, using `Log.info { ... }` inside `DB` types will resolve to the one previously defined and not the top-level one.

If you move code around during development, you will not need to deal with `logger`/`@logger`/`@@logger` substitutions.

If you need nested sources, you can use the parent namespace `Log` and avoid hard-coding the source name.

```crystal
class DB::Pool
  Log = DB::Log.get("pool")
end
```

Or, if you want to hard-code it.

```crystal
class DB::Pool
  Log = ::Log.get("db.pool")
end
```

Or, you can pass directly a type and the source will be inferred from it.

```crystal
class DB::Pool
  Log = ::Log.get(self)
end
```

The result of `Log.get(source)` will return the same object instance for the same source. This will allow overriding the severity level of the log during specific moments programmatically.

```crystal
Log.get("db").level = :verbose # or Log::Severity::Verbose
do_something_with_db
Log.get("db").level = :none # Log::Severity::None disables all levels

# DB::Log can be used instead of Log.get("db")
```

**Note:** Setting the level is specific to a source and not it's nested ones. The previous example doest not affect the `db.pool` source.

**Note:** Note that setting the level of a `Log` instance will affect all other Fibers that are concurrently using it.

When building a program it will be possible to ignore completely as part of the compilation the `debug` calls. So there is no penalty even to check if those logs should be emitted. The `debug` methods can be replaced by noop during compile-time. This will not be implemented initially.

You can also change programatically the logger level upon creation if you are hacking around.

```crystal
class DB::Pool
  Log = ::Log.get(self, :debug)
end
```

## Writing backends

The std-lib will come with a basic backends:

* `Log::IOBackend` to write to an IO
* `Log::MemoryBackend` to store logs in memory (it's suitable for testing)

But you can define your own backend by inheriting `Log::Backend`.

* `#write(Log::Entry)` will be called to write the given log entry.
* `#close` to close resources used by the backend

```crystal
class MyCustomBackend < Log::Backend
  def initialize
  end

  def write(e : Log::Entry)
  end

  def close
  end
end
```

Whether the write operation in sync or async it will be up to the implementation. There might be some base modules to implement easily an async backends (similar to `IO::Buffered`).

## Configuration

### Configure logging explicitly in the code

Use `Log.builder` to indicate which sources should go to which backends.

You can indicate actual sources or patterns.

* the empty string matches only the top-level source
* `*` matches all the sources
* `foo.bar.*` matches `foo.bar` and every nested source
* `foo.bar` matches `foo.bar`, but not its nested sources

The following configuration will setup for all sources to emit
warnings (or higher) to `STDOUT`, allow any of the `db.*` and
nested source to emit debug (or higher), and to also emit for all
sources errors (or higher) to an elasticsearch backend.

```
backend = Log::IOBackend.new
Log.builder.bind "*", :warning, backend
Log.builder.bind "db.*", :debug, backend
Log.builder.bind "*", :error, ElasticSearchBackend.new("http://localhost:9200")
```

### Configure logging from environment variables

Include the following line to allow configuration from environment variables.

```
Log.setup_from_env
```

The environment variables `CRYSTAL_LOG_LEVEL` and `CRYSTAL_LOG_SOURCES` are used to indicate
which severity level to emit (defaults to `INFO`; use `NONE` to skip all messages) and to restrict
which sources you are interested in.

The valid values for `CRYSTAL_LOG_SOURCES` are:

* the empty string matches only the top-level source (default)
* `*` matches all the sources
* `foo.bar.*` matches `foo.bar` and every nested source
* `foo.bar` matches `foo.bar`, but not its nested sources
* Any comma separated combination of the above

The logs are emitted to `STDOUT` using a `Log::IOBackend`.

If `Log.setup_from_env` is called on startup you can tweak the logging as:

```
$ CRYSTAL_LOG_LEVEL=DEBUG CRYSTAL_LOG_SOURCES=* ./bin/app
```

## Implicit Context

The context information will live in the fiber. A new fiber will start with an empty context. Context are immutable but can be extended with convenient methods. The context are immutable so the backend can keep a reference to the context that was used when generating the `Log::Entry`.

The current context can be read from `Log.context`.

You can change the current context with `clear` and `set` methods.

```crystal
Log.context.clear # => clears current context
Log.context.set request_id: 34, user_id: "jdoe@example.org" # => extends the current context
```

When spawning `Fibers` you can inherit the context explicitly:

```crystal
c = Log.context # current context
spawn do
  Log.context = c
  # ...
end
```

`Log.with_context` can be used with a block to restore to the previous context.

```crystal
Log.context["request_id"] # => 23

Log.with_context do 

  Log.context.set request_id: 57689
  Log.context["request_id"] # => 57689

end # previous context will be restored at this point

Log.context["request_id"] # => 23
```

## Other types

Disclaimer: this is more the structure description rather than the actual definition.

```crystal
alias ContextValue = Bool | Int32 | Int64 | Float32 | Float64 | String | Time | Hash(String, ContextValue)
alias Context = Hash(String, ContextValue)

record Entry, source : String, severity : Severity, message : String, timestamp : Time, context : Context, exception : Exception?
enum Severity = Debug ; Verbose ; Info ; Warning ; Error ; Fatal ; None
```

PR #8850, opened 2020-02-25
Deque#from_json/to_json

carlhoerberg:


src/json/from_json.cr
```diff
@@ -58,6 +58,14 @@ def Array.from_json(string_or_io) : Nil
   nil
 end
 
+def Deque.from_json(string_or_io) : Nil
+  parser = JSON::PullParser.new(string_or_io)
+  new(parser) do |element|
+    yield element
+  end
+  nil
```
Sija: Redundant

```suggestion
```

carlhoerberg: just copied the Array.from_json implementation

Issue #7272, opened 2019-01-05
Format the value after # => 


r00ster91:
I think the value after `# =>` should be formatted by the Crystal formatter.
Currently there are a lot of cases in for example the [`Array`](https://crystal-lang.org/api/0.27.0/Array.html) documentation where the value after `# => ` doesn't match Crystal's formatting rules at all.
Example: [`Array#permutations`](https://crystal-lang.org/api/0.27.0/Array.html#permutations%28size%3AInt%3Dself.size%29-instance-method):
> Returns an [`Array`](https://crystal-lang.org/api/0.27.0/Array.html) with all possible permutations of *size*.
> ```cr
> a = [1, 2, 3]
> a.permutations    # => [[1,2,3],[1,3,2],[2,1,3],[2,3,1],[3,1,2],[3,2,1]]
> a.permutations(1) # => [[1],[2],[3]]
> a.permutations(2) # => [[1,2],[1,3],[2,1],[2,3],[3,1],[3,2]]
> a.permutations(3) # => [[1,2,3],[1,3,2],[2,1,3],[2,3,1],[3,1,2],[3,2,1]]
> a.permutations(0) # => [[]]
> a.permutations(4) # => []
> ```
Crystal would format e.g. `[[1],[2],[3]]` to `[[1], [2], [3]]` if it were code.

r00ster91:
No, I don't think there should be errors raised for comments. I think in the comment land you should feel safe and free of errors.
The comments that don't parse should be ignored in my opinion.

straight-shoota:
Now that an implementation has been proposed, I reckon we should continue discussing the strictness. 

I prefer it to be strongly formalized and error when `# =>` is not followed by a valid expression. The main benefit of this is that it helps catch errors. If there's a typo that results in valid code but not properly formatted, the formatter fixes the formatting. If the typo leads to invalid code however, it should even be more important to point that out.

While being very similar in many cases regarding primitive values, there is however a discrepancy between "valid code" and "output of p! (i.e. #inspect)". As evidenced in the example in the OP, `p! [] of Int23` returns `[]` which itself is not a valid expression.
We need to clarify what `=>` is supposed to represent with respect to the expression immediately before it: Does it literally mean the result of `#inspect`? Or is it supposed to be an unambiguous and re-usable representation of the value? Or as an alternative route, should the result of `#inspect` fit the latter specification?
[...]

PR #8851, opened 2020-02-25
Format code in comments following # =>

fschrfl:
Implements the feature discussed in  #7272.

Using the same example as in the issue:
```crystal
a = [1, 2, 3]
a.permutations    # => [[1,2,3],[1,3,2],[2,1,3],[2,3,1],[3,1,2],[3,2,1]]
a.permutations(1) # => [[1],[2],[3]]
a.permutations(2) # => [[1,2],[1,3],[2,1],[2,3],[3,1],[3,2]]
a.permutations(3) # => [[1,2,3],[1,3,2],[2,1,3],[2,3,1],[3,1,2],[3,2,1]]
a.permutations(0) # => [[]]
a.permutations(4) # => []
```
The above is now formatted like this:
```crystal
a = [1, 2, 3]
a.permutations    # => [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
a.permutations(1) # => [[1], [2], [3]]
a.permutations(2) # => [[1, 2], [1, 3], [2, 1], [2, 3], [3, 1], [3, 2]]
a.permutations(3) # => [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
a.permutations(0) # => [[]]
a.permutations(4) # => []
```

Note that the last two lines are no valid Crystal expressions in which case the formatter doesn't change them.

This is my first attempt at a contribution to the project so if I did anything wrong please let me know and I will try to improve.

### DIFF
--- a/src/compiler/crystal/tools/formatter.cr
+++ b/src/compiler/crystal/tools/formatter.cr
@@ -4516,7 +4516,15 @@ module Crystal
         raw_after_comment_value = value[1..-1]
         after_comment_value = raw_after_comment_value.strip
         if after_comment_value.starts_with?("=>")
-          value = "\# => #{after_comment_value[2..-1].strip}"
+          # If a comment starts with `=>`, it indicates the result of an expression which
+          # is a valid expression itself and should be properly formatted.
+          formatted_result_expression = after_comment_value[2..-1]
+          begin
+            formatted_result_expression = Formatter.format(formatted_result_expression)
+          rescue Crystal::SyntaxException
+            # A failure to parse the comment is ignored. It is only a comment after all.
+          end
+          value = "\# => #{formatted_result_expression.strip}"
         elsif after_comment_value.each_char.all? { |c| c.ascii_whitespace? || c == '#' }
           # Nothing, leave sequences of whitespaces and '#' as is
         else


src/compiler/crystal/tools/formatter.cr
```diff
@@ -4516,7 +4516,15 @@ module Crystal
...
-          value = "\# => #{after_comment_value[2..-1].strip}"
+          # If a comment starts with `=>`, it indicates the result of an expression which
+          # is a valid expression itself and should be properly formatted.
+          formatted_result_expression = after_comment_value[2..-1]
+          begin
+            formatted_result_expression = Formatter.format(formatted_result_expression)
+          rescue Crystal::SyntaxException
+            # A failure to parse the comment is ignored. It is only a comment after all.
+          end
+          value = "\# => #{formatted_result_expression.strip}"
```
j8r: `#` doesn't have to be escaped.

src/compiler/crystal/tools/formatter.cr
```diff
@@ -4516,7 +4516,15 @@ module Crystal
         raw_after_comment_value = value[1..-1]
         after_comment_value = raw_after_comment_value.strip
         if after_comment_value.starts_with?("=>")
-          value = "\# => #{after_comment_value[2..-1].strip}"
+          # If a comment starts with `=>`, it indicates the result of an expression which
+          # is a valid expression itself and should be properly formatted.
+          formatted_result_expression = after_comment_value[2..-1]
```
j8r: Can be replaced by:
`if formatted_result_expression = after_comment_value.lchop? "=>"`

j8r: To clarify, both the initial if condition and the line with `[2..-1]` can be replaced by the above expression.

fschrfl:
I added some specs and adjusted the code from the first commit based on your suggestions. Thanks for the feedback. Hope, the specs cover everything. I am not very imaginative when it comes to test.

spec/compiler/formatter/formatter_spec.cr
```diff
@@ -1610,4 +1610,8 @@ describe Crystal::Formatter do
       # second comment
     }
     CODE
+
+  # 7272
```
straight-shoota: There's no need to reference the issue, a textual description is better:
```suggestion
  # format expressions in `# =>` comments
```

spec/compiler/formatter/formatter_spec.cr
```diff
@@ -1610,4 +1610,8 @@ describe Crystal::Formatter do
...
+  assert_format "1 # => =>1+2"
```
straight-shoota: Better use an example that's explicitly about invalid code and doesn't look like there might be some special behaviour with `=> =>` :
```suggestion
  assert_format "1 # => ?invalid  code"
```

src/compiler/crystal/tools/formatter.cr
```diff
@@ -4515,8 +4515,15 @@ module Crystal
         value = @token.value.to_s.strip
         raw_after_comment_value = value[1..-1]
         after_comment_value = raw_after_comment_value.strip
-        if after_comment_value.starts_with?("=>")
-          value = "\# => #{after_comment_value[2..-1].strip}"
+        if formatted_result_expression = after_comment_value.lchop? "=>"
```
straight-shoota: Minor detail, but it's not formatted at the beginning, so I'd just call this variable `result_expression` (everywhere).
```suggestion
        if result_expression = after_comment_value.lchop? "=>"
```

spec/compiler/formatter/formatter_spec.cr
fschrfl: I did it this way because a comment just like that exists in several other places in the file and I wanted to be consistent. However, I do agree that a textual description is superior and I implemented that along with your other suggested changes.

spec/compiler/formatter/formatter_spec.cr
```diff
@@ -1610,4 +1610,8 @@ describe Crystal::Formatter do
       # second comment
     }
     CODE
+
+  # format expressions in `# =>` comments
+  assert_format "1 # => ?invalid code"
+  assert_format "1 # => #=>1+2", "1 # => # => 1 + 2"
```
oprypin: These specs cover known edge cases but it's as important to test a few basic cases first.

spec/compiler/formatter/formatter_spec.cr
oprypin: FYI commenting in that style is typical only for specs that get added specifically in response to a newly found bug.

spec/compiler/formatter/formatter_spec.cr
fschrfl: What basic cases would you suggest? The two cases that need to be covered are invalid and valid code. Invalid code is covered by the first case, for valid code I could add another basic case like `1 # => 1 + 2` but that is also already covered by the recursion in case two.

asterite: What. happens if the comment expression isa method definition with, say, 1 + 2 as the body?

fschrfl: That is indeed a problem I did not consider. As it is now, the formatter will put the body and `end` in separate lines which are not commented out:
```crystal
1 # => def add12() 1 + 2 end
```
turns into
```crystal
1 # => def add12
  1 + 2
end
```
It would be a good idea, anyway, to only apply formatting to certain kinds of expressions, or better still, to only apply formatting if the formatted string is equal to the result of `.inspect`ing the expression before the comment (see discussion in #7272). If that was implemented, the above problem wouldn't even come up. So before fixing this, it might be a good idea to overthink the whole approach here...

asterite: I think it's simpler: try to format it. If you find newlines, then don't use the result.

asterite: Of course that means that we can't always require what comes after `# => ` to be properly formatted. /cc @straight-shoota

straight-shoota: If the formatter inserts a line break, that looks like an indicator that there is something wrong. I'm convinced it should fail in that case. I can't think of any example where the proper result of an expression described with ` =>` would be split in multiple lines.

asterite: I don't think the code formatter should ever fail. It would be pretty annoying...

asterite: Specially considering that it's a tool that's combined with the editor. I would find it pretty annoying if the editor starts spitting errors at me because of the way I organize my code.

straight-shoota: Yeah maybe the formatter is not the right floor this. But it would be great to have some formal validation for `# =>` comments and similar.
Maybe that's something for ameba?

asterite: If anything, a tool like doctest would be great.

I think having the formatter try to format the value if possible is good enough. Then doctest would actually run it and verify that the results are what you'd expect them to be.

fschrfl: Incidentally, is there a reason why `# def add12 : Int32 1 + 2 end` and `# def add12() 1 + 2 end` are valid syntax while `# def add12 1 + 2 end` raises the syntax error "unexpected token: 1"?

asterite: Just a lack of thought from us on those details. We should probably copy Ruby or decide what to do, but it's very low priority.

PR #8852, opened 2020-02-25
feat(slice.cr): add `unsafe_bytes` method

stakach:
for obtaining the `Slice(UInt8)` representation of the underlying bytes of any slice.

### DIFF
--- a/spec/std/slice_spec.cr
+++ b/spec/std/slice_spec.cr
@@ -143,6 +143,11 @@ describe "Slice" do
 
     expect_raises(IndexError) { slice.copy_to(pointer, 5) }
   end
+  
+  it "does convert to_bytes" do
+    slice = Slice.new(3) { |i| i + 1 }
+    slice.to_bytes.should eq(Bytes[1, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0])
+  end
 
   describe ".copy_to(Slice)" do
     it "copies bytes" do
--- a/spec/std/slice_spec.cr
+++ b/spec/std/slice_spec.cr
@@ -144,9 +144,9 @@ describe "Slice" do
     expect_raises(IndexError) { slice.copy_to(pointer, 5) }
   end
 
-  it "does convert to_bytes" do
+  it "does provide access to unsafe_bytes" do
     slice = Slice.new(3) { |i| i + 1 }
-    slice.to_bytes.should eq(Bytes[1, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0])
+    slice.unsafe_bytes.should eq(Bytes[1, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0])
   end
 
   describe ".copy_to(Slice)" do
--- a/spec/std/slice_spec.cr
+++ b/spec/std/slice_spec.cr
@@ -143,7 +143,7 @@ describe "Slice" do
 
     expect_raises(IndexError) { slice.copy_to(pointer, 5) }
   end
-  
+
   it "does convert to_bytes" do
     slice = Slice.new(3) { |i| i + 1 }
     slice.to_bytes.should eq(Bytes[1, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0])
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -415,6 +415,11 @@ struct Slice(T)
     @pointer.move_to(target.to_unsafe, size)
   end
 
+  # Returns the `Slice(UInt8)` representation of the underlying bytes
+  def to_bytes
+    Slice(UInt8).new(self.to_unsafe.as(Pointer(UInt8)), self.bytesize)
+  end
+
   # Moves the contents of *source* into this slice. *source* and `self` may
   # overlap; the copy is always done in a non-destructive manner.
   #
--- a/src/slice.cr
+++ b/src/slice.cr
@@ -416,7 +416,7 @@ struct Slice(T)
   end
 
   # Returns the `Slice(UInt8)` representation of the underlying bytes
-  def to_bytes
+  def unsafe_bytes
     Slice(UInt8).new(self.to_unsafe.as(Pointer(UInt8)), self.bytesize)
   end
 


stakach:
Discussion here: https://forum.crystal-lang.org/t/convert-slice-uint16-to-slice-uint8/1747/4

asterite:
I'm not sure this is a generally useful method to have.

stakach:
Would have been handy for me.
My code currently looks like this:

```
    password = @password.to_utf16
    # Obtain the bytes that make up the UTF16 slice
    password = Bytes.new(password.to_unsafe.as(Pointer(UInt8)), password.bytesize)

    key = IO::Memory.new(Bytes.new(32))
    key.write password
    cipher.key = key.to_slice
```

Anyone that uses the `.to_utf16` will probably want to write those bytes to an IO at some point (otherwise what's the point?) so it seems odd there is no simple method to get the bytes

asterite:
> Anyone that uses the .to_utf16

This is what I mean: in my many several years of programming, I never had to deal with UTF-16. Not even when coding in Crystal, except for binding against Windows.

I don't think UTF-16 is common enough. And converting a Slice(T) to Slice(UInt8) is just a one liner:

```crystal
Bytes.new(slice.to_unsafe.as(UInt8*), slice.bytesize)
```

I feel that `to_bytes` doesn't convey a good meaning of what's going on inside the method. But with that one-liner it's very clear.

straight-shoota:
> Anyone that uses the .to_utf16 will probably want to write those bytes to an IO at some point (otherwise what's the point?) so it seems odd there is no simple method to get the bytes

UTF-16 encoding is only used for interfacing with the win32 API. Crystal strings internally always use UTF-8, and so do most libraries. For that reason the conversion to/from UTF-16 usually happens directly where the respective C functions are called.

Your use case that you need to pass a `Slice(UInt16)` as plain `Bytes` is actually quite rare.
The typical way to handle strings in different encoding is through `IO` with `set_encoding`. For `UInt16` there's obviously a faster method using `from_utf16`/`to_utf16`  and casting from/to `Bytes`.

I tend to agree with @asterite that it's probably better to use a more verbose expression for this. But given some use cases, it might be worth adding a method.
It should probably be named `#unsafe_bytes`, though.

lribeiro:
Recalling to_unsafe and cast it as pointer to UInt8 doesn't immediately come to mind.
Hiding this call behind a method makes it easily accessible and lowers the entry barrier.

straight-shoota:
@lribeiro In this case we're talking about a very specific low-level feature that involves unsafe behaviour. We should expect that everyone using such a method needs to know what exactly is going on behind the scenes anyway. Lowering the entry barrier might even be counterproductive in this regard and lead to unexpected results.

Having a dedicated method for this would help readers understand the code. That's a benefit.

asterite:
Let's wait. If someone else needs this method we can consider adding it.

My point is, when I'm coding something I also find some things useful. It's tempting to go ahead and add it to the standard library. But if we do that we'll end up with thousands of methods, each with a very specific use case.

I really don't think we should add it. It's too specific.

What's wrong with adding that method in the shard you are using? Is it that important that it's in the standard library?

stakach:
I basically assumed that it should be simple to write any slice to IO.
It isn't and it's seems quite counter intuitive how one has to go about achieving a simple task like that.

asterite:
Solution: `password.encode("UTF-16")`. You get a slice of bytes with the password encoded in UTF-16.

No need to ask for `utf_16`, use an `IO::Memory`, etc.

That's why I'm more convinced that this method is not needed.

asterite:
Also, `to_utf16` is mainly there for Windows compatibility. The general way to encode a string in another encoding and get the bytes is `encode`. `to_utf16` shouldn't be public API in my opinion (I think I mentioned it in the past, this is one result of it being public API).

stakach:
Ignoring `.utf_16` it seems weird to me that it's super simple to write `Array(UInt32)` to an IO but if you have a `Slice(UInt32)` which quacks like an array, it requires C style pointer manipulation to write to an IO

stakach:
it could also be used to make functions like `#hexdump` work for more variations of Slices.

asterite:
How do you write Array of Int32 to an IO?

asterite:
The point is, Slice is there for UInt8. That it's generic is a bit of a mistake. I wanted it to be nongeneric but it ended up being generic.

stakach:
> How do you write Array of Int32 to an IO?
My mistake, you loop through the items and write them one at a time

> Solution: password.encode("UTF-16").
This actually returns a different set of bytes, prefixes the string with `Bytes[254, 255, 0]`

> The point is, Slice is there for UInt8
Ok, I take your point :)

asterite:
I guess, if others think this is useful, we can add it. `to_bytes` was probably a better name. There's nothing unsafe in converting between slice types.

stakach:
I guess it's unsafe to assume the endianness of the underlying bytes?
Just another nail in the coffin that is this function ;)

stakach:
I might even move away from my current
`Bytes.new(self.to_unsafe.as(Pointer(UInt8)), self.bytesize)` and instead write the bytes to an IO with the correct endianness so the decrypt will work on arm platforms.

stakach:
Although I guess `password.encode("UTF-16")` probably already puts the bytes in the correct order.
Thanks for the debate @asterite - I think on the whole you are correct to reject this method

straight-shoota:
> `to_utf16` shouldn't be public API in my opinion (I think I mentioned it in the past, this is one result of it being public API).

It's not only useful to stdlib's win32 API calls but also for third party bindings. So it makes sense to have these methods publicly exposed.

> This actually returns a different set of bytes, prefixes the string with `Bytes[254, 255, 0]`

That's actually just iconv prepending a byte order mark. You can avoid that by specifying the encoding with explicit byte order:
```cr
"password".encode("UTF-16BE") # => Bytes[0, 112, 0, 97, 0, 115, 0, 115, 0, 119, 0, 111, 0, 114, 0, 100]
"password".encode("UTF-16LE") # => Bytes[112, 0, 97, 0, 115, 0, 115, 0, 119, 0, 111, 0, 114, 0, 100, 0]
```

stakach:
Thanks @straight-shoota that is super useful to know!

PR #8854, opened 2020-02-25
Remove return type restriction on Tuple.from and Class#cast

### LINKED ISSUE #8842: Tuple.from doesn't support parent types
https://play.crystal-lang.org/#/r/8lyw

```cr
module Foo
end
 
class Bar
  include Foo
end
 
pp Tuple(Int32, Bar?).from [1, Bar.new.as(Foo)]
```

It seems this is a result of `Class#cast` and `Tupel.from` having a return type of `self`, which does not take parent types into consideration.

Blacksmoke16:
Fixes #8842

### DIFF
--- a/spec/std/tuple_spec.cr
+++ b/spec/std/tuple_spec.cr
@@ -11,6 +11,13 @@ private class TupleSpecObj
   end
 end
 
+module TestModule
+end
+
+class TestClass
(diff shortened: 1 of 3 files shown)


asterite:
Invalid, #8842 is not a bug.

PR #8857, opened 2020-02-26
Allow Time::Span.new with positional + nanoseconds

### LINKED ISSUE #8825: Missing (deprecated) definition for Time::Span
In [crystal/crystal-mysql](https://github.com/crystal-lang/crystal-mysql/blob/4c4737c22f69df7ab68e14a442b282f70430c281/src/mysql/types.cr#L225) the following construct is used

```cr
::Time::Span.new(0, time.hour, time.minute, time.second, nanoseconds: time.nanosecond)
```

Although it will be removed due to #8257, in master is totally gone already.

I think we should add that overload in case others uses them. I will update crystal-mysql for 0.34, but with the current situation I can't publish a patch release that will work with 0.33 and 0.34 (unless `{% if compare_version ...`). But the whole point of deprecated message is to allow the current shard version to compile on the new std-lib with warnings.

bcardiff:
Fixes #8825

### DIFF
--- a/spec/std/time/span_spec.cr
+++ b/spec/std/time/span_spec.cr
@@ -46,6 +46,11 @@ describe Time::Span do
     t.total_seconds.should eq(1231231231231)
   end
 
+  it "initialize deprecated constructors" do
+    t = Time::Span.new(0, 0, 0, 0, nanoseconds: 1)
+    t.total_nanoseconds.should eq(1)
+  end
+
   it "days overflows" do
     expect_overflow do
       days = 106751991167301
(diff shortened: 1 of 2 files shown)


straight-shoota:
I don't follow why we need this to fix #8825. What's wrong with my suggestion from https://github.com/crystal-lang/crystal/issues/8825#issuecomment-589006913 ?

I suppose we could still follow with this because `nanoseconds` is more likely to be passed as named parameter than other arguments. I'm not sure, but it might have even been a named-only argument at one point.

bcardiff:
The idea was to use named argument for subseconds units since the nano/micro was not as obvious as the rest of the parts.

Some db drivers use the named argument so without this we are breaking that code without a deprecation cycle. People are forced to update in order to use nightly rather that noticing a deprecation warning on nightly.

PR #8863, opened 2020-02-28
Use flag?(:i386) instead of i686

### LINKED ISSUE #8824: nightly on dist_linux32 are failing regarding i386-linux-musl c/sys/types
https://circleci.com/gh/crystal-lang/crystal/38954

```
In /crystal/src/lib_c/i386-linux-musl/c/sys/types.cr:6:20

 6 | alias BlksizeT = Long
                      ^---
Error: undefined constant Long
```

bcardiff:
The normalized architecture i386 is the one present in the flags

Fixes #8824.
The cause was a conflict between

* https://github.com/crystal-lang/crystal/pull/7282/files#diff-782340590099fc3f5015755a3b74b586R31
* https://github.com/crystal-lang/crystal/pull/7282/files#diff-0817fdb7bea983f15d454e56956e8725R21
* https://github.com/crystal-lang/crystal/pull/8772/files#diff-782340590099fc3f5015755a3b74b586L31

And this PR should be the missing piece.

The test ecosystem is not checking progress on the 32-bits platform, obviously. Because there is no cross-compiling there. Progress is guaranteed on 64-bits on nightlies thanks to test_dist_linux_on_docker job though.

### DIFF
--- a/spec/compiler/codegen/asm_spec.cr
+++ b/spec/compiler/codegen/asm_spec.cr
@@ -2,7 +2,7 @@ require "../../spec_helper"
 
 describe "Code gen: asm" do
   # TODO: arm asm tests
-  {% if flag?(:i686) || flag?(:x86_64) %}
+  {% if flag?(:i386) || flag?(:x86_64) %}
     it "codegens without inputs" do
       run(%(
         dst = uninitialized Int32
--- a/src/callstack.cr
+++ b/src/callstack.cr
@@ -52,7 +52,7 @@ struct CallStack
     @backtrace ||= decode_backtrace
   end
 
-  {% if flag?(:gnu) && flag?(:i686) %}
+  {% if flag?(:gnu) && flag?(:i386) %}
     # This is only used for the workaround described in `Exception.unwind`
     @@makecontext_range : Range(Void*, Void*)?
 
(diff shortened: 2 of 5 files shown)


j8r:
I doubt Crystal really support true i386 systems. Most modern OS doesn't support it anymore, even Debian.
i686 is the (one of?) most modern 32bit x86 architecture.

straight-shoota:
@j8r this is just following the nomenclature of arch in LLVM target triple. We don't make the decision whether it should be i386 or i686.

j8r:
LLVM nomenclature supports both architectures, and will return i686 on modern 32bit computers.
Nowadays, nearly every x86-32 are i686.

j8r:
It may have changed though https://reviews.llvm.org/D26764

j8r:
On Alpine, I got:
```sh
llvm-config --host-target
i586-alpine-linux-musl
```
On Debian:
```sh
llvm-config --host-target
i686-pc-linux-gnu
```
So here no reason to choose i386 over i486, i586 or i686, unless copying Debian architecture which is named i386 for historical purposes - but doesn't mean much now.

RX14:
If we really wanted we could rename it `IA_32`. `i386` is a very well known name for this, though. I don't think we should swim against the stream here.

Issue #8853, opened 2020-02-25
Multiple inheritance: type mismatch when no child defined


j8r:
At least one child object has to be defined to make the program compiling.

The code below compiles, but when the two lines are commented, the error is `Error: no overload matches 'met' with type AbstractMain+.class`.

I'd expect to always compile, since `AbstractSubMain.class` is a `AbstractMain.class`.

```cr
abstract struct AbstractMain
end

abstract struct AbstractSubMain < AbstractMain
end

# Comment this two lines to have the error
struct SubSubMain < AbstractSubMain
end

def met(s : AbstractSubMain.class)
end

struct Container
  property main : AbstractMain.class | Nil = nil
  def yield(& : AbstractMain.class ->)
    if main = @main
      yield main
    end
  end
end

Container.new.yield do |m|
  met m
end
```

asterite:
The snippet, either with those two commented lines or not, should not compile. An overload for all class types in the hierarchy must be provided. One for `AbstractMain` is missing.

asterite:
Similar, or reduced:

```crystal
abstract class Foo
end

abstract class Bar < Foo
end

def foo(x : Bar.class)
  p x
end

f = Foo || Bar
p f

foo(f)
```

Wrong output:

```
Foo
Bar
```

Expected compilation to fail.

PR #8870, opened 2020-03-01
Compiler: don't say "undefined method" on abstract leaf types

### LINKED ISSUE #6996: Undefined method with empty array of abstract class T
https://play.crystal-lang.org/#/r/5cxo

```Crystal
abstract class Assertion
  abstract def valid? : Bool
end

class MyAssertion(T) < Assertion 
  def valid? : Bool
    false
  end
end

class Validator
  getter assertions : Array(Assertion) = [] of Assertion
  
  def validate
    @assertions.reject(&.valid?)
  end
end

pp Validator.new.validate
```

The example above gives an error saying `undefined method 'Assertion#valid?()'`

However if you replace the `getter assertions` with `getter assertions : Array(Assertion) = [MyAssertion(String).new] of Assertion` It'll work fine.  It'll also work fine when replacing the `abstract def valid? : Bool` with an actual method that returns a bool, even with the empty array.

It looks like when the array is empty the type of each value is inferred to be `Assertion`. Which doesn't implement the `valid?` method itself.

asterite:
Fixes #6996

Consider this code:

```crystal
abstract class Foo
end

foos = [] of Foo
foos.all?(&.valid?) # => true
```

The code above compiles fine in Crystal. The reason is that, because `Foo` is abstract, only its subclasses can actually be instantiated. But there are no subclasses! So there's no way to actually instantiate a `Foo`.

The compiler will then consider any call on such abstract types to be valid, but produce `NoReturn`... which makes sense: there's no way to reach that (but see the "Reaching that NoReturn" at the end).

The reason this also compiles is because Crystal tries to mimic Ruby. Let's consider this Ruby code:

```ruby
class Foo
end

foos = [] of Foo
foos.all?(&:valid?) # => true
```

It works! Because we didn't pass any Foo, it's trivially true. Ruby doesn't even need to check that `valid?` method, **and Crystal does the same thing**.

Crystal is always "if you don't call it, we don't know". Which makes sense according to its prototypey and duck-typey nature.

## But it didn't always work

If `Foo` had an abstract subclasses, the compiler used to complain:

```crystal
abstract class Foo
end

abstract class Bar < Foo
end

foos = [] of Foo
foos.all?(&.valid?) # Error: undefined method 'valid?' for Foo+
```

The bug is that the compiler just checked whether `Foo` was abstract without subclasses. But the condition should also apply if `Foo` is abstract and all of its subclasses are in the same condition. The compiler calls this "abstract leaf".

It can also happen with generic types, even if they are concrete:

```crystal
abstract class Foo
end

class Bar(T) < Foo
end

foos = [] of Foo
foos.all?(&.valid?) # => true
```

The above snippet didn't work, but with this PR it works. It works because `Bar(T)` has no generic instances. `Foo` was never actually instantiated.

And if you do instantiate `Bar` you'll get an error saying "undefined method 'valid?' for Bar(Int32) (compile-time type is Foo+)".

## Why is this useful?

This is useful because one could define a hierarchy of abstract types, or generic types, and write code to work with that. Maybe a test creates instances. Maybe other tests don't because they expect third party shards to defined subclasses. All of these cases should work.

## Reaching that NoReturn

Actually, there is:

```crystal
abstract class Foo
end

ptr = Pointer(Foo).malloc(1)
ptr.value.valid?
```

The above compiles fine and raises an error on runtime:

```
can't execute `ptr.value.valid?` at /Users/asterite/Projects/crystal/foo.cr:5:3: `ptr.value.valid?` has no type (Exception)
```

But that's unsafe and it usually never happens in real code: the pointer is inside an array, there are bound checks, etc.

### DIFF
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -41,6 +41,25 @@ module Crystal
       false
     end
 
+    # Returns `true` if this type is abstract and and it has no
+    # concrete subclasses. This can happen if this type has abstract
+    # subclasses, or non-abstract generic subclasses without instantiations.
+    def abstract_leaf?
+      type = self.devirtualize
+
+      case type
+      when GenericType
+        (type.abstract? && type.generic_types.empty?) ||
+          (type.generic_types.all? { |name, type| type.abstract_leaf? })
+      when GenericClassInstanceType
+        type.abstract? && type.subclasses.all?(&.abstract_leaf?)
+      when ClassType
+        type.abstract? && type.subclasses.all?(&.abstract_leaf?)
+      else
+        false
+      end
+    end
+
     # Returns `true` if this type is a struct.
     def struct?
       false
--- a/src/compiler/crystal/types.cr
+++ b/src/compiler/crystal/types.cr
@@ -41,7 +41,7 @@ module Crystal
       false
     end
 
-    # Returns `true` if this type is abstract and and it has no
+    # Returns `true` if this type is abstract and it has no
     # concrete subclasses. This can happen if this type has abstract
     # subclasses, or non-abstract generic subclasses without instantiations.
     def abstract_leaf?
--- a/spec/compiler/semantic/class_spec.cr
+++ b/spec/compiler/semantic/class_spec.cr
@@ -1091,4 +1091,66 @@ describe "Semantic: class" do
       ),
       "can't read instance variables of union types (@x of (Bar | Foo))"
   end
+
+  it "types as no return if calling method on abstract class with all abstract subclasses (#6996)" do
+    assert_type(%(
+      require "prelude"
+
+      abstract class Foo
+        abstract def foo?
+      end
+
+      abstract class Bar < Foo
+      end
+
+      Pointer(Foo).malloc(1_u64).value.foo?
+      )) { no_return }
+  end
+
+  it "types as no return if calling method on abstract class with generic subclasses but no instances (#6996)" do
+    assert_type(%(
+      require "prelude"
+
+      abstract class Foo
+        abstract def foo?
+      end
+
+      class Bar(T) < Foo
+        def foo?
+          true
+        end
+      end
+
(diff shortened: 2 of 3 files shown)


yxhuvud:
What I don't understand about the explanation is why `malloc` is allowed to allocate an abstract class. What does that even mean?

asterite:
It's not allocating an abstract class. It's allocating space to put an abstract class.

asterite:
Better said: space to put anything that inherits from that abstract class.

yxhuvud:
Ah ok, and it only works because it references have a fix size. Got it.

src/compiler/crystal/types.cr
```diff
@@ -41,6 +41,25 @@ module Crystal
       false
     end
 
+    # Returns `true` if this type is abstract and and it has no
```
bcardiff:
```suggestion
    # Returns `true` if this type is abstract and it has no
```

j8r:
I hoped it fixes https://github.com/crystal-lang/crystal/issues/8853 too, but actually no.

bew:
In my opinion, the first code should NOT work, unless you declare an abstract def for the method, so:
```cr
abstract class Foo
  # Without this is should fail to compile
  abstract def valid? : Bool
end

foos = [] of Foo
foos.all?(&.valid?) # => true
```

no?

asterite:
@bew Ideally, maybe yes. Abstract methods are not required at all, they are just there to better document the code. The compiler doesn't use abstract methods at all except for checking that subclasses implement them. But it's a different topic.

bew:
> Abstract methods are not required at all

But then, how `foos.all?(&.valid?)` is supposed to work type-wise if `valid?` isn't declared anywhere?

asterite:
The compiler says: there are no concrete instances. It doesn't matter what this method is or how is this called, I'm going to type it as NoReturn.

bew:
@asterite I see thanks!

PR #8873, opened 2020-03-02
IO: Set fd to -1 on closed File/Socket

bcardiff:
Based on https://github.com/crystal-lang/crystal/pull/8733#discussion_r379699783

When closing a File or Socket the fd is set to -1 to force an invalid file descriptor on other threads that passed the `check_open` assertion before closing it.

Atomic is used to force a volatile Java-like semantics so the reads are sequentially consistent with the writes of other threads.

### DIFF
--- a/src/file/preader.cr
+++ b/src/file/preader.cr
@@ -4,7 +4,7 @@ class File::PReader < IO
 
   getter? closed = false
 
-  def initialize(@fd : Int32, @offset : Int32, @bytesize : Int32)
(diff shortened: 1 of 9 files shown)


src/file/preader.cr
```diff
@@ -4,7 +4,7 @@ class File::PReader < IO
 
   getter? closed = false
 
-  def initialize(@fd : Int32, @offset : Int32, @bytesize : Int32)
```
bew: why?

Issue #8712, opened 2020-01-23
Read-Only functionality for HTTP::Headers


repomaa:
As soon as the response body receives a write call, the response headers are written to the socket. It says as much in the docs.

This is a suggestion that, in my opinion, would save a lot of devs from headache:

As soon as the response body is written the headers are set into a read-only mode, after which setting headers will raise exceptions.

I had a hard time debugging why it looks like i've set the correct response headers but none of them is sent to the client. This is increasingly hard when using web frameworks that handle the request life cycle for you.

rdp:
Seems well received, PR encouraged! :)

PR #8876, opened 2020-03-04
Add flush to many builders

### LINKED ISSUE #8875: Incomplete CSV file if writing a file with CSV module
Hello together,

while generating some test data in a CSV file I recognized that the produced file doesn't contain all the lines it should have. I reduced the code as much as possible. As a workaround I'm able to produce a correct file by using the normal "File" module and joining my fields with ";".

```crystal
require "csv"

class Testcase
  property status
  property tool
  property comment

  def initialize(
    status = "",
    tool = "",
    comment = ""
  )
    @status = status
    @tool = tool
    @comment = comment
  end
end

testcases = Hash(String, Testcase).new

4000.times do |x|
  testcases["Testcase_#{x}"] = Testcase.new(status: "passed", tool: "selenium", comment: "some unnecessary comment #{x}")
end

puts "hash has correct size of 4000 = #{testcases.size == 4000}"

CSV.build(File.new(filename: "./generated_tests.csv", mode: "w"), separator: ';') do |csv|
  csv.row "TESTNAME", "STATUS", "TOOL", "COMMENT"
  counter = 0
  testcases.each do |tc_name, tc|
    counter += 1
    csv.row tc_name, tc.status, tc.tool, tc.comment
  end
  puts "number of hash iterations is also #{counter}, so no issue here"
end

lines = File.read_lines("./generated_tests.csv")
puts "Lines of created csv file: #{lines.size}" # => "Lines of created csv file: 3995"
```

Tested it on:
Current Ubuntu stable & Crystal 0.33.0
Windows Ubuntu WSL & Crystal 0.33.0
play.crystal-lang.org & Crystal 0.33.0: https://play.crystal-lang.org/#/r/8o8z

asterite:
Fixes #8875

Sorry that there are no specs but it's a bit cumbersome to test. Plus it's really easy to see what's going on.

### DIFF
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -349,7 +349,7 @@ module XML
   # Writes XML document into the given `IO`. An `XML::Builder` is yielded to the block.
   #
   # Builds an XML document (see `#document`) including XML declaration (`<?xml?>`).
-  def self.build(io : IO, version : String? = nil, encoding : String? = nil, indent = nil, quote_char = nil)
+  def self.build(io : IO, version : String? = nil, encoding : String? = nil, indent = nil, quote_char = nil) : Nil
     build_fragment(io, indent: indent, quote_char: quote_char) do |xml|
       xml.start_document version, encoding
       yield xml
@@ -360,7 +360,7 @@ module XML
   # Writes XML fragment into the given `IO`. An `XML::Builder` is yielded to the block.
   #
   # Builds an XML fragment without XML declaration (`<?xml?>`).
-  def self.build_fragment(io : IO, *, indent = nil, quote_char = nil)
+  def self.build_fragment(io : IO, *, indent = nil, quote_char = nil) : Nil
     xml = XML::Builder.new(io)
     xml.indent = indent if indent
     xml.quote_char = quote_char if quote_char
@@ -370,6 +370,7 @@ module XML
     # when StartDocument is omitted.
     xml.end_document
     xml.flush
+    io.flush
     v
   end
 end
--- a/src/yaml/builder.cr
+++ b/src/yaml/builder.cr
@@ -49,9 +49,10 @@ class YAML::Builder
 
   # Creates a `YAML::Builder` that will write to the given `IO`,
   # invokes the block and closes the builder.
-  def self.new(io : IO)
+  def self.new(io : IO) : Nil
     emitter = new(io)
     yield emitter ensure emitter.close
+    io.flush
   end
 
   # Starts a YAML stream.
@@ -244,7 +245,7 @@ module YAML
   end
 
   # Writes YAML into the given `IO`. A `YAML::Builder` is yielded to the block.
-  def self.build(io : IO)
+  def self.build(io : IO) : Nil
     YAML::Builder.new(io) do |yaml|
       yaml.stream do
         yaml.document do
--- a/src/csv.cr
+++ b/src/csv.cr
@@ -155,9 +155,10 @@ class CSV
   # end
   # io.to_s # => "HEADER\none,two\nthree\n"
   # ```
-  def self.build(io : IO, separator : Char = DEFAULT_SEPARATOR, quote_char : Char = DEFAULT_QUOTE_CHAR, quoting : Builder::Quoting = Builder::Quoting::RFC)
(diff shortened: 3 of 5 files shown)


src/yaml/builder.cr
```diff
@@ -244,7 +245,7 @@ module YAML
   end
 
   # Writes YAML into the given `IO`. A `YAML::Builder` is yielded to the block.
-  def self.build(io : IO)
+  def self.build(io : IO) : Nil
```
Sija: No `io.flush` here?

asterite: No, because `YAML::Builder` already does that.

src/xml/builder.cr
```diff
@@ -360,7 +360,7 @@ module XML
   # Writes XML fragment into the given `IO`. An `XML::Builder` is yielded to the block.
   #
   # Builds an XML fragment without XML declaration (`<?xml?>`).
-  def self.build_fragment(io : IO, *, indent = nil, quote_char = nil)
+  def self.build_fragment(io : IO, *, indent = nil, quote_char = nil) : Nil
```
wontruefree: If you are enforcing `nil` do you still need to return `v` at the end?

https://github.com/crystal-lang/crystal/pull/8876/files#diff-c34fb2ed7bd91eabe52d9954efba8a88R374

asterite: Oh, I guess here we don't want `nil`, I'm not sure... I'll remove the `Nil` annotation

wontruefree: I am surprised this compiled with the type restriction.

asterite: The `Nil` type restriction forces the method to return `nil`, whatever it returns.

straight-shoota: I was a bit surprised about returning the yielded value when I visited this code in #8813. I suppose it's kind of nice, but I really doubt there are particular use cases for this. Returning `nil` would be fine, too. It's probably a question of continuity: should we expect all yielding methods that only have side effects and don't return anything inherently to return the block value? I guess this could get complicated when block execution is optional. Also, while it helps to simplify code that would otherwise rely on closured variables, I doubt there are many actually use cases where this comes in handy.

src/yaml/builder.cr
```diff
@@ -49,9 +49,10 @@ class YAML::Builder
 
   # Creates a `YAML::Builder` that will write to the given `IO`,
   # invokes the block and closes the builder.
-  def self.new(io : IO)
+  def self.new(io : IO) : Nil
```
straight-shoota: General question: Should this method be named `build`? It's not a constructor and doesn't return an instance of `YAML::Builder` which I'd usually expect from a `.new` method.

src/xml/builder.cr
asterite: Yeah, I think we should leave it as `Nil`

src/yaml/builder.cr
asterite: Yeah, I was surprised by this name too. We could probably rename it, but I'd like to avoid doing that in this PR.

src/yaml/builder.cr
straight-shoota: Why does `YAML::Builder` flush but `XML::Builder` doesn't? Shouldn't we make this behave identical?
`JSON::Builder#flush` delegates to `io.flush` which is another different behaviour.

asterite: I noticed that too. I didn't want to start making changes all over the place. That's why I added `flush` when needed.

Also, maybe this PR is not needed...

Sija:
I love the branch name, btw 🤣

Sija:
Hmm, I believe I've found out one possible regression: calling `#to_json/yaml` on `HTTP::Server::Response` will flush the underlying `io`, making it impossible to modify the response afterwards (thinking headers) before closing or passing it further on.

Blacksmoke16:
Related https://github.com/crystal-lang/crystal/issues/8712.

Darimac1:
yaml, as in Yahoo mail server?

Blacksmoke16:
@Darimac1 https://en.wikipedia.org/wiki/YAML.

Darimac1:
*Fast

Sija:
Could any1 delete these offtopic/spam comments, please? /cc @waj @bcardiff @asterite

asterite:
I don't have rights, otherwise I would have already banned him on the second comment.

PR #8877, opened 2020-03-04
Set HTTP::Headers in read-only mode after HTTP::Headers were written

### LINKED ISSUE #8712: Read-Only functionality for HTTP::Headers

asterite:
Fixes #8712

### DIFF
--- a/spec/std/http/headers_spec.cr
+++ b/spec/std/http/headers_spec.cr
@@ -209,7 +209,7 @@ describe HTTP::Headers do
 
   it "raises when read_only (#8712)" do
     headers = HTTP::Headers.new
-    headers.read_only = true
+    headers.read_only!
 
     expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
     expect_raises(ReadOnlyError) { headers["Foo"] = ["Bar", "Baz"] }
--- a/spec/std/http/headers_spec.cr
+++ b/spec/std/http/headers_spec.cr
@@ -206,4 +206,18 @@ describe HTTP::Headers do
     headers.add?("foobar", invalid_value).should be_false
     headers.add?("foobar", [invalid_value]).should be_false
   end
+
+  it "raises when read_only (#8712)" do
+    headers = HTTP::Headers.new
+    headers.read_only = true
+
+    expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
+    expect_raises(ReadOnlyError) { headers["Foo"] = ["Bar", "Baz"] }
+    expect_raises(ReadOnlyError) { headers.add "Foo", "Bar" }
+    expect_raises(ReadOnlyError) { headers.add "Foo", ["Bar", "Baz"] }
+    expect_raises(ReadOnlyError) { headers.add? "Foo", "Bar" }
+    expect_raises(ReadOnlyError) { headers.add? "Foo", ["Bar", "Baz"] }
+    expect_raises(ReadOnlyError) { headers.delete "Foo" }
+    expect_raises(ReadOnlyError) { headers.merge!({"Foo" => "Bar"}) }
+  end
 end
--- a/spec/std/http/headers_spec.cr
+++ b/spec/std/http/headers_spec.cr
@@ -207,9 +207,9 @@ describe HTTP::Headers do
     headers.add?("foobar", [invalid_value]).should be_false
   end
 
-  it "raises when read_only (#8712)" do
+  it "raises when already sent (#8712)" do
     headers = HTTP::Headers.new
-    headers.read_only!
+    headers.mark_as_sent!
 
     expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
     expect_raises(ReadOnlyError) { headers["Foo"] = ["Bar", "Baz"] }
--- a/spec/std/http/headers_spec.cr
+++ b/spec/std/http/headers_spec.cr
@@ -207,17 +207,41 @@ describe HTTP::Headers do
     headers.add?("foobar", [invalid_value]).should be_false
   end
 
-  it "raises when already sent (#8712)" do
-    headers = HTTP::Headers.new
-    headers.mark_as_sent!
+  describe "sent" do
+    it "raises when already sent (#8712)" do
+      headers = HTTP::Headers.new
+      headers.sent = true
 
-    expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
-    expect_raises(ReadOnlyError) { headers["Foo"] = ["Bar", "Baz"] }
-    expect_raises(ReadOnlyError) { headers.add "Foo", "Bar" }
-    expect_raises(ReadOnlyError) { headers.add "Foo", ["Bar", "Baz"] }
-    expect_raises(ReadOnlyError) { headers.add? "Foo", "Bar" }
-    expect_raises(ReadOnlyError) { headers.add? "Foo", ["Bar", "Baz"] }
-    expect_raises(ReadOnlyError) { headers.delete "Foo" }
-    expect_raises(ReadOnlyError) { headers.merge!({"Foo" => "Bar"}) }
+      expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
+      expect_raises(ReadOnlyError) { headers["Foo"] = ["Bar", "Baz"] }
+      expect_raises(ReadOnlyError) { headers.add "Foo", "Bar" }
+      expect_raises(ReadOnlyError) { headers.add "Foo", ["Bar", "Baz"] }
+      expect_raises(ReadOnlyError) { headers.add? "Foo", "Bar" }
+      expect_raises(ReadOnlyError) { headers.add? "Foo", ["Bar", "Baz"] }
+      expect_raises(ReadOnlyError) { headers.delete "Foo" }
+      expect_raises(ReadOnlyError) { headers.merge!({"Foo" => "Bar"}) }
+    end
+
+    it "can still read them" do
+      headers = HTTP::Headers{"Foo" => "Bar"}
+      headers.sent = true
+
+      headers["Foo"].should eq("Bar")
+    end
+
+    it "can still read them" do
+      headers = HTTP::Headers{"Foo" => "Bar"}
+      headers.sent = true
+
+      headers["Foo"].should eq("Bar")
+    end
+
+    it "can write them if sent is false" do
+      headers = HTTP::Headers{"Foo" => "Bar"}
+      headers.sent = true
+      headers.sent = false
+
+      headers["Foo"] = ["Bar", "Baz"]
+    end
   end
 end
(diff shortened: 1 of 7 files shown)


src/exception.cr
```diff
@@ -163,3 +163,15 @@ class NilAssertionError < Exception
...
+# Raised when trying to modify an object that is read-only.
+#
+# ```
+# slice = Slice.new(10, read_only: true) { |i| i }
+# slice[0] = 1
+# ```
+class ReadOnlyError < Exception
+  def initialize(message)
+    super(message)
+  end
```
j8r: Why defining an `initialize` here?

src/http/headers.cr
```diff
@@ -48,16 +48,24 @@ struct HTTP::Headers
...
+  # When in read-only mode, trying to modify these headers
+  # will raise `ReadOnlyError`.
+  def read_only!
+    @read_only = true
+  end
+
+  # Returns `true` if these headers are in read-only mode.
+  def read_only?
+    @read_only
   end
```
Sija:
```suggestion
  getter? read_only : Bool
```

src/exception.cr
asterite: Because the message is optional in `Exception`. I want to force users of this exception to provide why the type was set to read-only.

Sija: If so, then it needs to be defined with `: String` argument restriction, otherwise you could pass a `nil` as well.

j8r: This error is supposed to be re-used elsewhere? It may be a better idea to have a dedicated error for reading a read-only header.

j8r: Having both will be the best, to rescue only headers errors, or only read-only errors - or both.

asterite: Why? I can't imagine a scenario where both can happen. And this exception should not be rescued, it means a bug in your code.

j8r: Sure here, why using a global `ReadOnly` error then?
I don't think all `ReadOnly` exceptions (if any) would be on this case.

Just create a dedicated error where it is used; in the `HTTP::Headers` module - more importantly if not planned to be reused elsewhere.

asterite: Because the standard library shouldn't be raising `Exception`, it should be raising specific errors. Slice was already doing that and I thought about using the same exception type because the reason is the same: an object was set to read-only mode.

j8r: Sure, just wondering about where to place this very error, to provide more context, which will prevent to specify `HTTP::Headers` in the message (and just say `headers`)...
That's not a big deal then.

j8r: I was thinking of something like `Cannot modify headers (HTTP::Headers::ReadOnlyError)` instead of `HTTP::Headers are in read-only (ReadOnlyError)`.

asterite: Just my opinion, but I don't think discussing these things matter.

bcardiff:
I think the top-level error is fine, and eventually a freezable module could become handy.

Do you recall why you found freeze confusing? Although since we are not implementing it in Object, I am good with going on read_only route.

asterite:
Maybe `freeze` is not confusing but I only found that term in Ruby. Also, Slice already uses `read_only` so I thought reusing that name for consistency was good.

vlazar:
Not requesting a change, just want to point out that `freeze` name is more common nowadays :)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

bcardiff:
@vlazar I would reserve freeze only for a more general way to freeze objects, enforced by the compiler/runtime implicitly. I am not saying that that would/could happen. But since that is the usage in other languages, I would prevent the overlap and probably generate some confusion.

z64:
> I'm open to suggestions about other ways to implement this.

To be honest, the original issue to me sounds like something the library should have solved this by providing an immutable view, and not expose `HTTP::Headers` directly in the first place.. If we want to provide this in the standard library though, I'd like to explore generating a compile-time error with the current Crystal language:

- Provide `Headers.build` that yields a builder pbject, returning a finalized `HTTP::Headers`
- Provide `Headers.new(hash)` or other ctors for compatible header layouts
- Make `Headers` immutable

We would still need some way for HTTP clients to supplement header "ergonomics", for example, supplying the correct `Content-Length` header or other default header values that the user didn't supply. 

In which case, I'd probably suggest a "lower level" `HTTP::Headers::Writer` for serializing `Headers` or header values to sockets / IO that `HTTP::Client` would use after serializing the user's headers. This should hopefully mitigate the cost of otherwise providing an awkward `#merge` method that returns an update copy, but maybe that's not a big deal performance-wise.. I don't think header collections get terribly large in most apps? I don't know, honestly.

Then, #8712 should be a compile-time error for frameworks that directly pass the user `HTTP::Headers` instances.

This has other analogs in our std, for concepts that construct `String`s, but of course these don't have the same consequences as `Headers`:
- `HTTP::Params.build` 
- `JSON.build` 
- `String.build`

Downsides would be that the deprecation period for the mutable methods would affect a lot of code, I imagine, so I don't think the above is compelling enough. But, WDYT?

src/http/headers.cr
```diff
@@ -356,4 +377,10 @@ struct HTTP::Headers
       end
     end
   end
+
+  private def check_writeable
+    if read_only?
+      raise ReadOnlyError.new("HTTP::Headers are in read-only mode (it probably means the HTTP::Server headers were already sent)")
```
straight-shoota: This has nothing to do with `HTTP::Server`. You can build and send headers without a server.
I would also suggest to let the error message explicitly say that headers were already sent. Maybe even make `read_only!` nodoc and rename it to `headers_written`. I don't think there's a need to call this method directly, read-only mode is not a standalone concept. We should assume read-only mode is only an implicity result of writing the headers to an IO.

```suggestion
      raise ReadOnlyError.new("HTTP::Headers were already sent)")
```

straight-shoota:
@z64 The problem with a builder approach is that HTTP headers are often not constructed at a single place, but dispersed over several handlers, middleware, framework code etc. I don't see how a builder could improve that without removing the ability to set header values *somewhere* in the HTTP handling stack (as long as it's before sending the first byte).

z64:
Right. Conceptually `Headers::Writer` would be the mutable counterpart, which I leave open-ended in this proposal.. it should be whatever it needs to be to allow what you described. Perhaps it's just exactly what `HTTP::Headers` is right now (but perhaps `Headers::Writable` would be a better name then, I don't know).

The intention was to divide `Headers` into mutable / immutable counterparts, so that it should be obvious to whoever is handling the type what the intention is, backed up by the compiler.

`Headers.build` would be for simple use cases (passing to one-shot `HTTP::Client.get`, for example), and streamed across a `Writer` internally.

straight-shoota:
@z64 Yes, `Headers` currently works like a builder but it skips a dedicated immutable representation and directly serializes to an IO. It also doubles as a regular reader, which may make it a bit more complex. But I agree with @asterite that the current API probably offers the best usability.
Having a dedicated builder wouldn't help with issues that arise from the order in which the headers are modified and serialised.

src/http/headers.cr
```diff
@@ -356,4 +377,10 @@ struct HTTP::Headers
       end
     end
   end
+
+  private def check_writeable
```
RX14: It's `writable` not `writeable`.

straight-shoota: @asterite

src/http/headers.cr
asterite: Good idea

asterite:
Ready for another review

spec/std/http/headers_spec.cr
```diff
@@ -206,4 +206,18 @@ describe HTTP::Headers do
     headers.add?("foobar", invalid_value).should be_false
     headers.add?("foobar", [invalid_value]).should be_false
   end
+
+  it "raises when already sent (#8712)" do
+    headers = HTTP::Headers.new
+    headers.mark_as_sent!
+
+    expect_raises(ReadOnlyError) { headers["Foo"] = "Bar" }
```
straight-shoota, line 214: Could you add a spec that accessing a header still works after `mark_as_sent!`?

asterite: Done!

asterite:
Ready for yet another review. I had to add a way to unmark it as sent, because the request processor reuses http headers. So, I made it a property.

spec/std/http/headers_spec.cr
```diff
@@ -206,4 +206,42 @@ describe HTTP::Headers do
...
+    end
+
+    it "can still read them" do
+      headers = HTTP::Headers{"Foo" => "Bar"}
+      headers.sent = true
+
+      headers["Foo"].should eq("Bar")
+    end
+
+    it "can still read them" do
```
straight-shoota, line 232: This spec seems duplicate.

src/http/server/response.cr
```diff
@@ -50,6 +50,7 @@ class HTTP::Server
     def reset
       # This method is called by RequestProcessor to avoid allocating a new instance for each iteration.
       @headers.clear
+      @headers.sent = false
```
straight-shoota, line 53: Shouldn't this happen in `Headers#clear`?

asterite: That was my first thought, but I'm not sure. It's not obvious that any time you want to clear the headers you also want to mark them as not sent.

I think that `clear` method delegates to Hash, so effectively jumping over the `sent?` guard. I think `HTTP::Headers` should stop forwarding stuff to Hash. Then we need to move `sent = false` above this line.

This looks like it will take a long time to settle, so for now I'm abandoning this PR. Anyone feel free to continue this.

straight-shoota: Clear already changes the headers, so it can't work any other way than with `sent == false`. It either needs to call `check_not_sent` or implicitly set `sent = false`. When doing the latter, we can also revert from `sent=` to `mark_as_send!` and the only way to remove the sent flag is calling `clear`. That should actually cover all use cases.
Maybe `clear` shoud be renamed to `reset` to better communicate its purpose.

asterite: The thing is, imagine the headers were already sent. Someone calls `clear`, appends a few headers, everything works fine. Except that it has no effect on the output because the headers were sent and you didn't get a runtime error.

This is what I'm trying to avoid.

straight-shoota: Then `clear` shouldn't be part of the public API or make it `clear!`/`reset!` to emphasize the possible dangers related to it.

asterite:
Closing. Let's discuss what API we want in the issue and then (someone) send a single PR without discussions.

PR #8881, opened 2020-03-05
Move Adler32 and CRC32 to Digest

bcardiff:
Adler32 and CRC32 interface are kept unmodified and hence they don't implement `Digest::Base`, but their results are 32-bit fixes intead of a slice/string; plus they support the combine alternative. I think the current API while simple is enough.

The top-level versions are deprecated.

### DIFF
--- a/spec/std/digest/adler32_spec.cr
+++ b/spec/std/digest/adler32_spec.cr
@@ -1,0 +1,16 @@
+require "spec"
+require "digest/adler32"
+
+describe Digest::Adler32 do
+  it "should be able to calculate adler32" do
+    adler = Digest::Adler32.checksum("foo").to_s(16)
+    adler.should eq("2820145")
+  end
+
+  it "should be able to calculate adler32 combined" do
+    adler1 = Digest::Adler32.checksum("hello")
+    adler2 = Digest::Adler32.checksum(" world!")
+    combined = Digest::Adler32.combine(adler1, adler2, " world!".size)
+    Digest::Adler32.checksum("hello world!").should eq(combined)
+  end
+end
--- a/spec/std/digest/crc32_spec.cr
+++ b/spec/std/digest/crc32_spec.cr
@@ -1,0 +1,16 @@
+require "spec"
+require "digest/crc32"
+
+describe Digest::CRC32 do
+  it "should be able to calculate crc32" do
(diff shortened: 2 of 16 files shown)


asterite:
`docs_main.cr` should be updated as well

bcardiff:
The files are still there. And digest is included with `require "./digest/**"`.

When the deprecated are effectively dropped, yes, the docs_main.cr should be updated.

straight-shoota:
I agree on moving Adler32 and CRC32 under a combined namespace. But I'm not sure they fit under the `Digest` namespace. As far as I am familiar with the matter, a (message) digest is usually associated with a cryptographic hash function (such as MD5 or SHA-1 which are currently available in `Digest` namespace; `OpenSSL::Digest` has even more).
Adler32 and CRC32 are not cryptographic hash functions but much weaker checksum/CRC.

We could still continue with this and accept there's some mismatch. `Digest` is not very specific and already established namespace in Crystal's stdlib.
Alternatively, we could consider a more broader term as top level namespace for hash functions. `Hash` is already taken, though ;)

bcardiff:
I've found that many std-libs put CRC32 and Adler32 under Digest and none that uses another namespace ...

asterite:
golang uses the `hash` namespace:
- https://golang.org/pkg/hash/adler32/
- https://golang.org/pkg/hash/crc32/

That said, I'm fine with digest

ysbaddaden:
Some digest algorithms are cryptographically strong but that's hardly a requirement, otherwise both MD5 (broken) and SHA-1 (weak) wouldn't be digest algorithms anymore, which would be nonsense.

Don't use CRC32, Adler32, MD5 or SHA-1 for secure usages like verifying that a file wasn't tampered with. Use SHA-2, SHA-3 or BLAKE. But not all usages need to be secure (e.g. gzip, deflate).

This is similar to random number generators: _some_ are them are cryptographically strong.

ysbaddaden:
Sure we can say hash instead of digest, but the term would be confusing in Crystal (`Hash`).

bcardiff:
I forgot to change the usages of them in the std-lib.
Rebased on master.

whidbey:
i dont notice this change.but  i dont' think is good idea to move A to B and move C to D 's breaking change like this.
thoght I admire your commits..but It's a wasting time i think... what I want is better "C" ..

PR #8882, opened 2020-03-05
Remove DL module

bcardiff:
The module is no longer used, nor included in prelude, and does not offer much functionality.


src/dl.cr
```diff
@@ -1,7 +0,0 @@
-require "c/dlfcn"
-
-module DL
-  def self.dlopen(path, mode = LibC::RTLD_LAZY | LibC::RTLD_GLOBAL) : Void*
-    LibC.dlopen(path, mode)
```
asterite: Should we remove the `fun` definitions for `dlopen` too?

asterite:
Should be removed from `docs_main` too.

Deprecating it wouldn't hurt either, people might be using it because it's documented.

asterite:
https://crystal-lang.org/api/0.33.0/DL.html

bcardiff:
In this case I think there is no need for a deprecation cycle. The module is most likely not used in the wild. But we can go that route if that is preferred. I'll wait for reviews.

straight-shoota:
I couldn't find any single usage in Crystal repositories available on Github. So I guess it's fine to remove without deprecation (there's certainly Crystal code outside of Github, but it's a very strong indicator that there is really no real world use).

PR #8885, opened 2020-03-06
Getting rid of Errno and WinError

### LINKED ISSUE #8827: [RFC] Getting rid of Errno
# Replacing `Errno`

Currently there are many places in the standard library where an `Errno` exception is raised when a syscall or a call to a C function returns an error condition. The value comes from the `errno` value and we use the system provided error message.

This approach has some issues, most of them already discussed (#8305, #5003):
  * Portability: even though `errno` should be defined by POSIX, different platforms have different error codes. Also Windows has a separate set of error codes.
  * Abstraction leak: the `Errno` exception handled at the user code it's sometimes leaking the syscall being executed underneath.
  * Hard to handle: there is a single `Errno` class (with no subclasses). Doing a `rescue Errno` catches all kind of errors. Also some `IO` functions raises either a `IO::Error` or `Errno` and callers must rescue from both classes.

## Proposed replacements

Instead of raising a generic `Errno`, some specific exception classes will be used depending on the are where the error occurs. All these errors will contain, at least, the error message provided by the system for the `errno` code.

**Note**: These are the subclasses that I found so far necessary to add to the hierarchy but it's not by any means finished and I'm open to suggestions to add more.

### IO::Error

`IO` module is where most of the errors coming from the system are raised. It makes sense because it's where most of the side effects happen.

`IO::Error` will replace `Errno` for all the general errors. However some subclasses of it will be created.

### IO::Timeout

This one already exists, but it's not a subclass of IO::Error. This will be fixed.

### IO::FileSystemError

This subclass of `IO::Error` will be raised for operations with the filesystem (mostly from `File` and `FileUtils` classes)

Some subclasses might also be necessary. So far I found the need to create a `IO::NotFoundError` so it can be handed specifically for some situations.

Other subclasses we could add: `AccessDeniedError`, `FileExistsError`, `NotDirectoryError`, etc.

### IO::SocketError

Similar to the previous exception, but for errors directly related with a network socket.
There is currently a `Socket::Error`, so we can actually rename it and make it a subclass of `IO::Error` (Socket functions currently raises a mix of `IO::Error` and `Socket::Error`).

### RuntimeError

For other situations where a system error occurs, a `RuntimeError` will be raised. For example, issues with threads, processes

## Exception Hierarchy

```
Exception
+- RuntimeError
+- IO::Error
   +- IO::FileSystemError
   |  +- IO::NotFoundError
   |  +- ...
   +- IO::SocketError
   +- IO::Timeout
```

waj:
This PR removes the usage of `Errno` and `WinError` as exceptions and instead provides some new classes to distinguish different kind of errors.

This is the hierarchy of the new or updated exceptions:
```
Exception
+- RuntimeError
+- IO::Error
   +- IO::TimeoutError
   +- File::Error
   |  +- File::NotFoundError
   |  +- File::AccessDeniedError
   |  +- File::AlreadyExistsError
   +- Socket::Error
      +- Socket::ConnectError
      +- Socket::BindError
```
More subclasses might be added in the future but I think this is a good starting point.

  * `Exception` now has an optional `os_error` containing the original error code that anyone can use in case the exception class hierarchy is not enough for proper error handling
  * `IO::Timeout` is renamed to `IO::TimeoutError`
  * `Errno` and `Winerror` are now enums
  * `Exception.from_errno` and `Exception.from_winerror` eases creation of exceptions from OS errors
  * `RuntimeError` is used, often when there is a system call that shouldn't fail in normal conditions

(closes #8827)


straight-shoota:
I don't like having `os_error` on `Exception`, spilling into the generic exception namespace when it's use is actually limited to a very small amount of exception types from stdlib and maybe a few exceptions in shards with C bindings. This propert doesn't make sense for the vast majority of exceptions in a typical Crystal program.
Can we only add it selectively on the exception types that use it?

src/exception.cr
```diff
@@ -67,6 +73,56 @@ class Exception
 
     io.flush
   end
+
+  protected def os_error=(@os_error)
```
Fryguy: `:nodoc:` ?  (Not sure we nodoc protected methods, but I saw some elsewhere that had it)

waj:
@straight-shoota actually I like it because it ensures uniformity. Many different new exceptions from shards or std might be created from system errors, and so many errors come from the system. I think we should properly document the `from_errno` and `from_winerror` methods and encourage their use. And to be honest, I don't feel that adding a field with a value type is spilling the API. It's convenient and unnoticed most of the time.

jhass:
As a compromise we could also have `module SystemError` (or `SyscallError` or so) for it.

waj:
Ok, agreed. I moved the implementation to the `SystemError` module and included in `RuntimeError` and `IO::Error`. Also added some docs to that module 🙂

src/exception.cr
waj: Done, and moved to `SystemError` module

src/crystal/system/win32/file.cr
```diff
@@ -44,12 +44,12 @@ module Crystal::System::File
...
 
-  private def self.check_not_found_error(func_name)
-    error = LibC.GetLastError
+  private def self.check_not_found_error(message, path)
+    error = WinError.value
     if NOT_FOUND_ERRORS.includes? error
       return nil
     else
-      raise WinError.new(func_name, error)
+      raise ::File::Error.from_winerror(message, file: path)
```
bcardiff: The `error` variable should be used in this call

src/socket.cr
```diff
@@ -179,15 +195,15 @@ class Socket < IO
   # sock = Socket.udp(Socket::Family::INET)
   # sock.bind Socket::IPAddress.new("192.168.1.25", 80)
   # ```
-  def bind(addr)
+  private def bind(addr, addrstr)
```
bcardiff: It seems we lost `sock.bind Socket::IPAddress.new("192.168.1.25", 80)` since the method is now private. A miss edition probably/

oprypin:
At a glance, this seems strictly better than https://github.com/crystal-lang/crystal/pull/5003 
👍

waj:
Thanks @oprypin. I considered some of the ideas and the feedback received on that PR.

src/file/error.cr
```diff
@@ -0,0 +1,41 @@
+class File < IO::FileDescriptor
+end
+
+class File::Error < IO::Error
+  getter file : String
```
RX14: Would be nice to have some documentation on these.

src/process.cr
```diff
@@ -142,7 +142,7 @@ class Process
       # error:
       errno = Errno.value
       LibC.pthread_sigmask(LibC::SIG_SETMASK, pointerof(oldmask), nil)
-      raise Errno.new("fork", errno)
+      raise RuntimeError.from_errno("fork", errno)
```
RX14: I think it would be kind of nice if `Process.new` on a non-existing file raised `File::NotFoundError`, i.e. all file-related errors were handled like this, even in mixed-function syscalls.

src/errno.cr
```diff
@@ -16,232 +16,52 @@ end
...
+                    EACCES EFAULT ENOTBLK EBUSY EEXIST EXDEV ENODEV ENOTDIR EISDIR EINVAL ENFILE
+                    EMFILE ENOTTY ETXTBSY EFBIG ENOSPC ESPIPE EROFS EMLINK EPIPE EDOM ERANGE EAGAIN
+                    EWOULDBLOCK EINPROGRESS EALREADY ENOTSOCK EDESTADDRREQ EMSGSIZE EPROTOTYPE ENOPROTOOPT
+                    EPROTONOSUPPORT ESOCKTNOSUPPORT EPFNOSUPPORT EAFNOSUPPORT EADDRINUSE EADDRNOTAVAIL
+                    ENETDOWN ENETUNREACH ENETRESET ECONNABORTED ECONNRESET ENOBUFS EISCONN ENOTCONN
+                    ESHUTDOWN ETOOMANYREFS ETIMEDOUT ECONNREFUSED ELOOP ENAMETOOLONG EHOSTDOWN
+                    EHOSTUNREACH ENOTEMPTY EUSERS EDQUOT ESTALE EREMOTE ENOLCK ENOSYS EOVERFLOW
+                    ECANCELED EIDRM ENOMSG EILSEQ EBADMSG EMULTIHOP ENODATA ENOLINK ENOSR ENOSTR
+                    EPROTO ETIME EOPNOTSUPP ENOTRECOVERABLE EOWNERDEAD) %}
+    {% if LibC.has_constant?(value) %}
```
RX14: We should either move this enum into `Crystal::` so it is not public or have it be always the same list of errno despite the platform: this enum must have the same *members* regardless of platform.

src/process.cr
```diff
@@ -479,17 +474,7 @@ class Process
...
-      io << "execvp ("
-      command.inspect_unquoted(io)
-      args.try &.each do |arg|
-        io << ' '
-        arg.inspect(io)
-      end
-      io << ")"
-    end
-    raise Errno.new(error_message)
+    raise RuntimeError.from_errno
```
straight-shoota: Why is this not `from_error("execvp")`?

src/errno.cr
waj: But not all errno values exist in all platforms. Maybe we can map those to EINVAL. But there should be at least a warning if you try to use any of those values in a platform where the value doesn't exist.

src/file/error.cr
waj: Of course. Documentation is still missing in many classes and methods.

src/process.cr
waj: I think eventually we should make an effort to improve many error messages. I'm not sure it's a good idea to leak the syscall name that failed. Maybe the same error message we have above: "Error executing process"

straight-shoota: Yes, that could work. "execvp" has always just been a stand-in for a proper message.

Using an explicit message here would make `from_errno` without arguments obsolete. I think it should be required to provide at least some message for context information.

src/errno.cr
RX14: We decided a long time ago that the compile-time API must be identical between platforms so that everything compiles on every platform.

If the errno value doesn't exist on every platform, it just never gets used, never gets raised, so it can be any value.

I don't think we need warnings for this.

aravindavk: The existence of an Error number is one part of the problem. There is platform-specific value for some error codes. For example, `grep -R "EDEADLK" src/lib_c/*`

```
src/lib_c/aarch64-linux-gnu/c/errno.cr:        EDEADLK = 35
src/lib_c/aarch64-linux-musl/c/errno.cr:       EDEADLK = 35
src/lib_c/amd64-unknown-openbsd/c/errno.cr:    EDEADLK = 11
src/lib_c/arm-linux-gnueabihf/c/errno.cr:      EDEADLK = 35
src/lib_c/i386-linux-gnu/c/errno.cr:           EDEADLK = 35
src/lib_c/i386-linux-musl/c/errno.cr:          EDEADLK = 35
src/lib_c/i686-linux-gnu/c/errno.cr:           EDEADLK = 35
src/lib_c/i686-linux-musl/c/errno.cr:          EDEADLK = 35
src/lib_c/x86_64-darwin/c/errno.cr:            EDEADLK = 11
src/lib_c/x86_64-dragonfly/c/errno.cr:         EDEADLK = 11
src/lib_c/x86_64-freebsd/c/errno.cr:           EDEADLK = 11
src/lib_c/x86_64-linux-gnu/c/errno.cr:         EDEADLK = 35
src/lib_c/x86_64-linux-musl/c/errno.cr:        EDEADLK = 35
src/lib_c/x86_64-macosx-darwin/c/errno.cr:     EDEADLK = 11
src/lib_c/x86_64-openbsd/c/errno.cr:           EDEADLK = 11
src/lib_c/x86_64-portbld-freebsd/c/errno.cr:   EDEADLK = 11
src/lib_c/x86_64-unknown-freebsd/c/errno.cr:   EDEADLK = 11
src/lib_c/x86_64-windows-msvc/c/errno.cr:      EDEADLK = 36
```

src/process.cr
didactic-drunk: Is it leaking the syscall name or telling you exactly what failed, which manpage to read and what to search for?

straight-shoota: I actually don't think leaking the name is an issue. It's really pretty concise about defining the context of the error. (Btw. it's a libc call, not a syscall)

But in other cases, I think we can perhaps be more expressive about the error context. There's a huge problem space with localization, though. So IDK, but using just the call looks like a pretty reasonable solution, at least for the time being.

PR #8887, opened 2020-03-07
Update hash.cr

rdp:
Tweak Hash docs

### DIFF
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1239,7 +1239,7 @@ class Hash(K, V)
     @size == 0
   end
 
-  # Calls the given block for each key-value pair and passes in the key and the value as a `Tuple`.
+  # Calls the given block for each key-value pair and passes in the key and the value as `each(& : Tuple(K, V) ->) : Nil`
   #
   # ```
   # h = {"foo" => "bar"}
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1239,7 +1239,7 @@ class Hash(K, V)
     @size == 0
   end
 
-  # Calls the given block for each key-value pair and passes in the key and the value.
+  # Calls the given block for each key-value pair and passes in the key and the value as a `Tuple`.
   #
   # ```
   # h = {"foo" => "bar"}
@@ -1855,6 +1855,7 @@ class Hash(K, V)
   # h = {1 => 'a', 2 => 'b', 3 => 'c'}
   # h.to_a # => [{1, 'a'}, {2, 'b'}, {3, 'c'}]
   # ```
+  # The order of the array follows the order the keys were inserted in the Hash.
   def to_a : Array({K, V})
     to_a_impl do |entry|
       {entry.key, entry.value}
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -1239,7 +1239,7 @@ class Hash(K, V)
     @size == 0
   end
 
-  # Calls the given block for each key-value pair and passes in the key and the value as `each(& : Tuple(K, V) ->) : Nil`
+  # Calls the given block for each key-value pair and passes in the key and the value.
   #
   # ```
   # h = {"foo" => "bar"}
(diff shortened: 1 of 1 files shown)


j8r:
To me it this information was quite obvious: Hash is ordered, as stated already in the object description, and the return type already state it will return an Array of `{K, V}` tuples.

src/hash.cr
```diff
@@ -1239,7 +1239,7 @@ class Hash(K, V)
     @size == 0
   end
 
-  # Calls the given block for each key-value pair and passes in the key and the value.
+  # Calls the given block for each key-value pair and passes in the key and the value as a `Tuple`.
```
j8r: Add a type annotation for the block instead

j8r: It will be `each(& : Tuple(K, V) ->) : Nil`.

src/hash.cr
```diff
@@ -1855,6 +1855,7 @@ class Hash(K, V)
   # h = {1 => 'a', 2 => 'b', 3 => 'c'}
   # h.to_a # => [{1, 'a'}, {2, 'b'}, {3, 'c'}]
   # ```
+  # The order of the array follows the order the keys were inserted in the Hash.
```
j8r: The fact that the ordered is preserved can be put on the main method description above.

src/hash.cr
rdp: Good idea.

src/hash.cr
rdp: Yeah it does seem weird at the bottom.  For whatever reason all the other methods did it this way so I just followed suit...

rdp:
Yeah it is already fairly obvious, but the other methods called it out so I figured why not :)

src/hash.cr
rdp: OK check it now, thanks.

src/hash.cr
alex-lairan: I think this is a coincidence due to the implementation, but this isn't true every time.

asterite: It's not a coincidence.

alex-lairan: Ok :+1:

rdp:
I think this is ready for review again, thanks!

src/hash.cr
straight-shoota: The idea was to add that directly to the method signature, not in the documentation text.

src/hash.cr
```diff
@@ -1239,7 +1239,7 @@ class Hash(K, V)
     @size == 0
   end
 
-  # Calls the given block for each key-value pair and passes in the key and the value.
+  # Calls the given block for each key-value pair and passes in the key and the value as `each(& : Tuple(K, V) ->) : Nil`
```
RX14: This sounds strange to me, what do you want to convey here?

src/hash.cr
rdp: OK added it to the signature now, thanks!

src/hash.cr
rdp: Hopefully avoided it by moving signature to the method itself now, thanks!

rdp:
I think this is ready for review again, thanks!

straight-shoota:
@asterite What do you mean regarding the type annotation?

asterite:
I mean, you usually do `hash.each { |k, v|`, you don't specify a tuple there because the compiler will expand it. But if there's a type annotation then maybe un experienced users will specify a tuple as a block argument.

Just that feeling.

straight-shoota:
Yeah, but we need that signature. Having no signature at all is obviously worse than a signature that could be misunderstood (but really don't lead to any errors).

`& : K, V  -> _` would techincally work, there shouldn't be a relevant difference (then remove the tuple from `yield`, obviously), but the tuple signature is required for implementing `Enumerable#each(& T : ->)`.

rdp:
I'll see if I can clarify it...

asterite:
There's no need to do anything else here. I'm just waiting on CI.

rdp:
I think CI is stuck FWIW...

straight-shoota:
Thanks @rdp

PR #2996, opened 2016-07-14
Add restrictions to collection methods. Fixes #1764. Fixes #988

asterite:
Instead of rebasing #1851 I decided to do it again after discussing this better with @mverzilli and @juanedi 

With this, all of these don't compile anymore:

``` cr
[1, 2, 3].includes?(nil)
[1, 2, 3].includes?(1_i64)
[1, 2, 3].delete "foo"
[1, 2, 3] - [1, 'a']

{1 => 2}["foo"]
{1 => 2}["foo"]?
{1 => 2}.has_key?('foo")
```

In this way you get a compile error when trying to pass an invalid type to an operation. The reason behind this is that, for example, `{1 => 2}["foo"]` will always raise, so why would you want that in a program. Similarly, `[1, 2].includes?("foo")` is always going to be false, so that check is redundant.

This is a bit against generic programming, but the few cases I had to change/fix in the compiler, plus some other few cases I noticed needed to change in some shards makes me think this is a good change.

This is, of course, not backwards compatible.

PR #8893, opened 2020-03-10
Enforce hash key types at compile time

jgaskins:
I noticed in one of my apps, I changed the type of a hash key, assuming the compiler would point out all the places I missed. But then all my `Hash#[]?(key)` calls started to fail silently and my `Hash#fetch(key, &)` calls invoked their blocks when I didn't expect them to.

I looked into how much code it would take to get the compiler to enforce the key types. It was more than I expected at first because I forgot the compiler is _also_ written in Crystal and uses these same hashes. :joy:

I don't think the semantics of this PR are correct yet, especially `Hash#==(other : Hash)`. I hardcoded `false` just to get that to compile, but before I spend too much time on trying to polish it all up, I'd like to get some conversation going about the pros and cons of this patch in case the current live functionality of this was intentional with full knowledge of the tradeoffs. 🙂

### DIFF
--- a/spec/std/set_spec.cr
+++ b/spec/std/set_spec.cr
@@ -179,13 +179,6 @@ describe "Set" do
   end
 
   it "does -" do
-    set1 = Set{1, 2, 3, 4, 5}
-    set2 = [2, 4, 'a']
-    set3 = set1 - set2
-    set3.should eq(Set{1, 3, 5})
-  end
-
-  it "does -" do
     set1 = Set{1, 2, 3, 4, 'b'}
     set2 = [2, 4, 5]
     set3 = set1 - set2
@@ -242,13 +235,6 @@ describe "Set" do
   end
 
   it "does subtract" do
-    set1 = Set{1, 2, 3, 4, 5}
-    set2 = Set{2, 4, 'a'}
-    set1.subtract set2
-    set1.should eq(Set{1, 3, 5})
-  end
-
-  it "does subtract" do
     set1 = Set{1, 2, 3, 4, 'b'}
     set2 = Set{2, 4, 5}
     set1.subtract set2
@@ -258,13 +244,6 @@ describe "Set" do
   it "does subtract" do
     set1 = Set{1, 2, 3, 4, 5}
     set2 = [2, 4, 6]
-    set1.subtract set2
-    set1.should eq(Set{1, 3, 5})
-  end
-
-  it "does subtract" do
-    set1 = Set{1, 2, 3, 4, 5}
-    set2 = [2, 4, 'a']
     set1.subtract set2
     set1.should eq(Set{1, 3, 5})
   end
--- a/src/compiler/crystal/semantic/type_declaration_processor.cr
+++ b/src/compiler/crystal/semantic/type_declaration_processor.cr
@@ -463,7 +463,7 @@ struct Crystal::TypeDeclarationProcessor
     ancestor = owner.ancestors.first?
     ancestor = uninstantiate(ancestor)
     if ancestor
-      ancestor_non_nilable = @non_nilable_instance_vars[ancestor]?
+      ancestor_non_nilable = @non_nilable_instance_vars[ancestor]? if ancestor.is_a?(Type)
     end
 
     # If the ancestor has non-nilable instance vars, check that all initialize either call
@@ -538,7 +538,7 @@ struct Crystal::TypeDeclarationProcessor
 
     owner.ancestors.any? do |ancestor|
       ancestor = uninstantiate(ancestor)
-      @instance_vars_outside[ancestor]?.try &.includes?(name)
+      @instance_vars_outside[ancestor]?.try &.includes?(name) if ancestor.is_a? Type
     end
   end
 
@@ -553,7 +553,7 @@ struct Crystal::TypeDeclarationProcessor
   end
 
   private def compute_non_nilable_outside_single(owner, non_nilable_outisde)
-    if vars = @instance_vars_outside[owner]?
+    if owner.is_a?(Type) && (vars = @instance_vars_outside[owner]?)
       non_nilable_outisde ||= [] of String
       vars.each do |name|
         non_nilable_outisde << name unless non_nilable_outisde.includes?(name)
@@ -570,7 +570,7 @@ struct Crystal::TypeDeclarationProcessor
 
     owner.ancestors.each do |ancestor|
       ancestor = uninstantiate(ancestor)
-      infos = @initialize_infos[ancestor]?
+      infos = @initialize_infos[ancestor]? if ancestor.is_a?(Type)
       return infos if infos && !infos.empty?
     end
 
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -470,7 +470,7 @@ module Crystal::Playground
 
       agent_ws = PathWebSocketHandler.new "/agent" do |ws, context|
         match_data = context.request.path.not_nil!.match(/\/(\d+)\/(\d+)$/).not_nil!
-        session_key = match_data[1]?.try(&.to_i)
+        session_key = match_data[1]?.try(&.to_i) || 0
         tag = match_data[2]?.try(&.to_i)
         @logger.info "#{context.request.path} WebSocket connected (session=#{session_key}, tag=#{tag})"
 
--- a/src/hash.cr
+++ b/src/hash.cr
@@ -329,7 +329,7 @@ class Hash(K, V)
 
   # Inserts or updates a key-value pair.
   # Returns an `Entry` if it was updated, otherwise `nil`.
-  private def upsert(key, value) : Entry(K, V)?
+  private def upsert(key : K, value : V) : Entry(K, V)?
     # Empty hash table so only initialize entries for now
     if @entries.null?
       @indices_size_pow2 = 3
@@ -416,7 +416,7 @@ class Hash(K, V)
 
   # Implementation of deleting a key.
   # Returns the deleted Entry, if it existed, `nil` otherwise.
-  private def delete_impl(key) : Entry(K, V)?
+  private def delete_impl(key : K) : Entry(K, V)?
     # Empty hash table, nothing to do
     if @indices_size_pow2 == 0
       return nil
@@ -465,7 +465,7 @@ class Hash(K, V)
   end
 
   # Finds an entry with the given key.
-  protected def find_entry(key) : Entry(K, V)?
+  protected def find_entry(key : K) : Entry(K, V)?
     # Empty hash table so there's no way it's there
     if @indices_size_pow2 == 0
       return nil
@@ -501,7 +501,7 @@ class Hash(K, V)
   end
 
   # Finds an Entry with the given key by doing a linear scan.
-  private def find_entry_linear_scan(key) : Entry(K, V)?
+  private def find_entry_linear_scan(key : K) : Entry(K, V)?
     # If we have less than 8 elements we avoid computing the hash
     # code and directly compare the keys (might be cheaper than
     # computing a hash code of a complex structure).
@@ -902,7 +902,7 @@ class Hash(K, V)
   end
 
   # Computes the hash of a key.
-  private def key_hash(key)
+  private def key_hash(key : K)
     if @compare_by_identity && key.responds_to?(:object_id)
       hash = key.object_id.hash.to_u32!
     else
@@ -911,7 +911,7 @@ class Hash(K, V)
     hash == 0 ? UInt32::MAX : hash
   end
 
-  private def entry_matches?(entry, hash, key)
+  private def entry_matches?(entry, hash, key : K)
     # Tiny optimization: for these primitive types it's faster to just
     # compare the key instead of comparing the hash and the key.
     # We still have to skip hashes with value 0 (means deleted).
@@ -927,7 +927,7 @@ class Hash(K, V)
     {% end %}
   end
 
-  private def entry_matches?(entry, key)
+  private def entry_matches?(entry : Entry(K, V), key : K)
     entry_key = entry.key
 
     if @compare_by_identity
@@ -1022,7 +1022,7 @@ class Hash(K, V)
   # h = Hash(String, String).new
   # h["foo"] # raises KeyError
   # ```
-  def [](key)
+  def [](key : K)
     fetch(key) do
       if (block = @block) && key.is_a?(K)
         block.call(self, key.as(K))
@@ -1043,7 +1043,7 @@ class Hash(K, V)
   # h = Hash(String, String).new("bar")
   # h["foo"]? # => nil
   # ```
-  def []?(key)
+  def []?(key : K)
     fetch(key, nil)
   end
 
@@ -1093,7 +1093,7 @@ class Hash(K, V)
   # h.has_key?("foo") # => true
   # h.has_key?("bar") # => false
   # ```
-  def has_key?(key)
+  def has_key?(key : K)
     !!find_entry(key)
   end
 
@@ -1104,7 +1104,7 @@ class Hash(K, V)
   # h.has_value?("foo") # => false
   # h.has_value?("bar") # => true
   # ```
-  def has_value?(val)
+  def has_value?(val : V)
     each_value do |value|
       return true if value == val
     end
@@ -1119,7 +1119,7 @@ class Hash(K, V)
   # h.fetch("foo", "foo") # => "bar"
   # h.fetch("bar", "foo") # => "foo"
   # ```
-  def fetch(key, default)
+  def fetch(key : K, default)
     fetch(key) { default }
   end
 
@@ -1131,7 +1131,7 @@ class Hash(K, V)
   # h.fetch("bar") { "default value" }  # => "default value"
   # h.fetch("bar") { |key| key.upcase } # => "BAR"
   # ```
-  def fetch(key)
+  def fetch(key : K)
     entry = find_entry(key)
     entry ? entry.value : yield key
   end
@@ -1752,8 +1752,12 @@ class Hash(K, V)
     self
   end
 
-  # Compares with *other*. Returns `true` if all key-value pairs are the same.
   def ==(other : Hash)
+    false
+  end
+
+  # Compares with *other*. Returns `true` if all key-value pairs are the same.
+  def ==(other : self)
     return false unless size == other.size
     each do |key, value|
       entry = other.find_entry(key)
--- a/src/named_tuple.cr
+++ b/src/named_tuple.cr
@@ -72,7 +72,7 @@ struct NamedTuple
   #
   # speak_about(**{thing: String, n: Int64}.from(hash)) # => "I see 2 worlds"
   # ```
-  def from(hash : Hash)
+  def from(hash : Hash(K, V)) forall K, V
     if size != hash.size
       raise ArgumentError.new("Expected a hash with #{size} keys but one with #{hash.size} keys was given.")
     end
@@ -80,7 +80,13 @@ struct NamedTuple
     {% begin %}
       NamedTuple.new(
       {% for key, value in T %}
-        {{key.stringify}}: self[{{key.symbolize}}].cast(hash.fetch({{key.symbolize}}) { hash["{{key}}"] }),
+        {% if K == String %}
+          {{key.stringify}}: self[{{key.symbolize}}].cast(hash[{{key.stringify}}]),
+        {% elsif K == Symbol %}
+          {{key.stringify}}: self[{{key.symbolize}}].cast(hash[{{key.symbolize}}]),
+        {% else %}
+          {% raise "Hash keys must be String or Symbol to convert into a NamedTuple" %}
+        {% end %}
       {% end %}
       )
     {% end %}
--- a/src/yaml/any.cr
+++ b/src/yaml/any.cr
@@ -61,7 +61,11 @@ struct YAML::Any
 
       new hash
     when YAML::Nodes::Alias
-      anchors[node.anchor]
+      if anchor = node.anchor
+        anchors[anchor]
+      else
+        raise "Unknown YAML alias: #{node}"
+      end
     else
       raise "Unknown node: #{node.class}"
     end
@@ -101,7 +105,14 @@ struct YAML::Any
         raise "Expected int key for Array#[], not #{object.class}"
       end
     when Hash
-      object[index_or_key]
+      case index_or_key
+      when String
+        object[Any.new(index_or_key)]
+      when Any
+        object[index_or_key]
+      else
+        raise "Expected YAML::Any | String, not #{object.class}"
+      end
     else
       raise "Expected Array or Hash, not #{object.class}"
     end
@@ -120,7 +131,14 @@ struct YAML::Any
         nil
       end
     when Hash
-      object[index_or_key]?
+      case index_or_key
+      when String
+        object[Any.new(index_or_key)]
+      when Any
+        object[index_or_key]
+      else
+        raise "Expected YAML::Any | String, not #{object.class}"
+      end
     else
       raise "Expected Array or Hash, not #{object.class}"
     end
@@ -296,14 +314,14 @@ struct YAML::Any
     @raw.pretty_print(pp)
   end
 
-  # Returns `true` if both `self` and *other*'s raw object are equal.
-  def ==(other : YAML::Any)
-    raw == other.raw
-  end
-
   # Returns `true` if the raw object is equal to *other*.
   def ==(other)
     raw == other
+  end
+
+  # Returns `true` if both `self` and *other*'s raw object are equal.
+  def ==(other : YAML::Any)
+    raw == other.raw
   end
 
   # See `Object#hash(hasher)`
--- a/src/compiler/crystal/program.cr
+++ b/src/compiler/crystal/program.cr
@@ -576,11 +576,11 @@ module Crystal
       file_module?(filename).try &.lookup_matches(signature)
     end
 
-    def file_module?(filename)
+    def file_module?(filename : String)
       file_modules[filename]?
     end
 
-    def file_module(filename)
+    def file_module(filename : String)
       file_modules[filename] ||= FileModule.new(self, self, filename)
     end
 
--- a/src/compiler/crystal/tools/typed_def_processor.cr
+++ b/src/compiler/crystal/tools/typed_def_processor.cr
@@ -11,7 +11,7 @@ module Crystal::TypedDefProcessor
 
   private def process_result(result : Compiler::Result)
     process_type result.program
-    if file_module = result.program.file_module? target_location.original_filename
+    if (filename = target_location.original_filename) && (file_module = result.program.file_module?(filename))
       process_type file_module
     end
   end
--- a/src/mime.cr
+++ b/src/mime.cr
@@ -220,7 +220,7 @@ module MIME
     # If the same extension had a different type registered before, it needs to
     # be removed from the extensions list.
     if previous_type = @@types[extension]?
-      if extensions = @@extensions[parse_media_type(previous_type)]?
+      if (parsed_type = parse_media_type(previous_type)) && (extensions = @@extensions[parsed_type]?)
         extensions.delete(extension)
       end
     end
--- a/src/regex/match_data.cr
+++ b/src/regex/match_data.cr
@@ -254,7 +254,7 @@ class Regex
 
       caps = [] of String?
       (1...size).each do |i|
-        caps << self[i]? unless name_table.has_key? i
+        caps << self[i]? unless name_table.has_key? i.to_u16
       end
 
       caps
@@ -276,7 +276,7 @@ class Regex
 
       caps = {} of String => String?
       (1...size).each do |i|
-        if (name = name_table[i]?) && !caps.has_key?(name)
+        if (name = name_table[i.to_u16]?) && !caps.has_key?(name)
           caps[name] = self[name]?
         end
       end
@@ -315,7 +315,7 @@ class Regex
 
       hash = {} of (String | Int32) => String?
       (0...size).each do |i|
-        if name = name_table[i]?
+        if name = name_table[i.to_u16]?
           hash[name] = self[name]? unless hash.has_key?(name)
         else
           hash[i] = self[i]?
@@ -334,7 +334,7 @@ class Regex
 
       io << "Regex::MatchData("
       size.times do |i|
-        io << ' ' << name_table.fetch(i, i) << ':' if i > 0
+        io << ' ' << name_table.fetch(i.to_u16, i) << ':' if i > 0
         self[i]?.inspect(io)
       end
       io << ')'
@@ -350,7 +350,7 @@ class Regex
             if i == 0
               self[i].pretty_print pp
             else
-              pp.text "#{name_table.fetch(i, i)}:"
+              pp.text "#{name_table.fetch(i.to_u16, i)}:"
               pp.nest do
                 pp.breakable ""
                 self[i]?.pretty_print pp
--- a/src/set.cr
+++ b/src/set.cr
@@ -132,7 +132,7 @@ struct Set(T)
   # s.includes? 5 # => true
   # s.includes? 9 # => false
   # ```
-  def includes?(object)
+  def includes?(object : T)
     @hash.has_key?(object)
   end
 
@@ -144,7 +144,7 @@ struct Set(T)
   # s.delete 5
   # s.includes? 5 # => false
   # ```
-  def delete(object)
+  def delete(object : T)
     @hash.delete(object)
     self
   end
--- a/src/yaml/parse_context.cr
+++ b/src/yaml/parse_context.cr
@@ -46,7 +46,9 @@ class YAML::ParseContext
 
   private def read_alias_impl(node, expected_crystal_type_id, raise_on_alias)
     if node.is_a?(YAML::Nodes::Alias)
-      value = @anchors[node.anchor]?
+      if anchor = node.anchor
+        value = @anchors[anchor]?
+      end
 
       if value
         object_id, crystal_type_id = value


asterite:
See https://github.com/crystal-lang/crystal/pull/2996 and the many related issues and PRs.

This is very unlikely to change, it makes the language too rigid.

bcardiff:
I was too slow to find #6154 🙃

jgaskins:
Well, I didn’t want it to go _too_ long before proposing it again. Gotta keep the tradition alive! 😂

I had a look through those PRs and I don’t think I agree that it makes it too rigid/pedantic. The extra type safety seems worth the very few extra type casts and nil checks you’d have to add for a very few hashes in your app — most hashes would not be affected. I use Crystal specifically for the type-safety guarantees, even if I’m using duck typing in most places and the type check is 10 stack frames away from where I’ve passed in the wrong object type.

I’m all about the safer code. I _like_ that Crystal forces me to handle `nil` instead of letting it propagate through my app — it’s the thing that frustrates me every time I have to fix a `NoMethodError` in one of my Ruby apps. It seems to make more sense to apply those same guarantees to hash keys. It would’ve saved me hours of investigation yesterday if it had simply told me that my hash key was the wrong type at compile time.

It might also help compile times since each `Hash` type would be able to focus on key type `K` instead of compiling `[]`, `[]?`, `fetch(key)`, `fetch(key, &)`, `key?`, and `delete` for every type passed in to every hash. This is at best an educated guess, though. I haven’t benchmarked it.

asterite:
I guess you can continue making changes to this PR until CI is green and then we can take a look at the diff and see if it's worth it.

bew:
I remember during #6154 that there was a complex hierarchy in the compiler that involves modules in modules, abstract and probably some inheritance in the mix that made me give up because of a compiler bug or something like that..

But we can definitely try again, find it and eventually fix it (if not already fixed in the recent compiler versions!)

jgaskins:
@bew This PR compiles the compiler and stdlib as-is. The only thing that seems to be broken based on the failing specs is hash equivalence.

I would love to see some realistic hash operations where the strong types don’t work, if anyone’s got one handy, so I can check it against this implementation and see what the application/shard code would look like in a before/after.

src/compiler/crystal/tools/playground/server.cr
```diff
@@ -470,7 +470,7 @@ module Crystal::Playground
 
       agent_ws = PathWebSocketHandler.new "/agent" do |ws, context|
         match_data = context.request.path.not_nil!.match(/\/(\d+)\/(\d+)$/).not_nil!
-        session_key = match_data[1]?.try(&.to_i)
+        session_key = match_data[1]?.try(&.to_i) || 0
```
asterite: Why is this needed?

src/hash.cr
```diff
@@ -1752,8 +1752,12 @@ class Hash(K, V)
     self
   end
 
-  # Compares with *other*. Returns `true` if all key-value pairs are the same.
   def ==(other : Hash)
+    false
```
asterite: This doesn't seem correct.

src/yaml/any.cr
```diff
@@ -296,16 +314,16 @@ struct YAML::Any
...
-
   # Returns `true` if the raw object is equal to *other*.
   def ==(other)
     raw == other
   end
 
+  # Returns `true` if both `self` and *other*'s raw object are equal.
+  def ==(other : YAML::Any)
+    raw == other.raw
+  end
```
asterite: Why was the order filpped?

src/compiler/crystal/tools/playground/server.cr
jgaskins: `session_key` can't be `nil` because of [this line](https://github.com/crystal-lang/crystal/blob/2d5b06cf99172520ff733e05e8499f5ddafd94cc/src/compiler/crystal/tools/playground/server.cr#L477). Because it can't store `nil` and it's using `[]` instead of `[]?` below, this line seems like it _can't_ return nil even without this PR or it would raise a `KeyError` below.

src/hash.cr
jgaskins: Indeed, I called it out in the original PR description.

src/yaml/any.cr
jgaskins: I was experimenting with the error messages. Before, when I've done a catch-all and a more specific overload of a method, I've had to put the catch-all version first or the catch-all would override the more specific overload. I don't know if this was me misunderstanding the type system at the time (it was a few years ago) and misattributing the error, but either way that's apparently not how it works anymore, if it ever did.

<details>

```crystal
def foo(bar : String)
  "strings only"
end

def foo(bar)
  "catch all"
end

def foo(bar : Int)
  123
end

foo 123 # => 123

foo "abc" # => "strings only"

foo %w[abc] # => "catch all"
```

</details>

jgaskins:
Alrighty, I think this PR is ready for some conversation about how it affects code. A few of my own observations:

- The majority of the changes in this PR are due to the `YAML::Any` wrapper trying to present as a hash
  - It’s not that that was a bad idea, it was actually a _fantastic_ idea from a code ergonomics perspective, but I feel like this PR does point out a few places where trying to hide that abstraction
  - Surprisingly, I don’t think this would’ve been a problem if we’d stuck with the recursive `YAML::Type` alias from back in the day
  - Equally surprisingly, `JSON` didn’t have any trouble with this, but I think it’s because its keys are guaranteed strings, not `JSON::Any` instances
- `Set` required an explicit typecast in order to handle subtraction of an enumerable whose types were a superset of its own
  - Admittedly, this feels wonky
  - I don’t think it’s going to be a big deal in practice since most collection types are homogeneous
- The compiler only needed 2 explicit `is_a?` checks
  - This is way less than I expected, if I’m being completely honest
  - All other checks were specifically against `nil`, which are extremely common in Crystal code — and even they weren’t required in too many places
- The key type restriction didn’t seem to be that much of a problem in specs outside of `YAML`
  - One thing to be aware of would be, for example, instead of `Hash(Int32, X)`, you may want to do `Hash(Int::Signed, X)` if you want to be more permissive with key types
  - Subclasses also work
    <details>
      <summary>Screenshot of subclasses and union types working through the playground running in this PR’s version of Crystal</summary>

      Screenshot of subclasses and union types working

    </details>
- Ary’s suggestion of `{1 => “a”}[1.0]` surprised me that it worked in the current version
  - This doesn’t work with arrays, though. You can’t do `array[1.0]` — it must be an `Int`
  - I can’t bind to port `8080.0` or `8080_u64`, so a simple `==` check seems like an exception to the rule
  - This does require thinking about hash key dereference as a method call, not just in implementation, but also conceptually

I feel like this is overall closer to the rest of the Crystal type system than relying on simple equality. I’ve admittedly been looking at it through my own lens, though, and applying my own mental model of the type system. So I admit it’s probably subjective.

I’m really interested to hear what kinds of surprises this would create in real-world apps. If I have some time, I’ll try to run it against a couple of mine tomorrow and see what I come up with.

spec/std/set_spec.cr
```diff
@@ -179,7 +179,7 @@ describe "Set" do
   end
 
   it "does -" do
-    set1 = Set{1, 2, 3, 4, 5}
+    set1 = Set(Int32 | Char){1, 2, 3, 4, 5}
```
bcardiff: It bugs me how strict you need to be in this examples. I imagine that `set1 - set2` no longer compiles when `set1 : Set(Int32)` and `set2 : Enumerable(Int32 | Char)`. To me it falls into the same behaviour we want on equality: although we can infer it will always be false due to typing missmatch, we want to make it a valid that will return false instead of not compile.

It could happen that this set for example is generated from templates, macros and the explicit typing will be hard to inject. Hence, the type will be the infered from the elements, leading to `Set(Int32)`.

jgaskins: I agree, this is a a pretty irritating tradeoff on paper, but let’s think about the overall impact of this consequence. How often is it that you have a homogeneous set that you subtract a heterogeneous set from?

Honestly, I can’t think of a single time I’ve ever used a heterogeneous set in application code, let alone in this particular scenario where I’m subtracting one from a homogenous set. To be clear, you don’t need to specify types for _all_ sets. Just ones where the type inference doesn’t get you what you need.

I do think that if you’re going to compare values, it’s probably a good idea to know that the types you’re comparing against can be different and the type specifier helps ensure that not only do you recognize that now, but you’ll recognize that when you change the code later. My perspective is that it’s not the same as adding two sets of different types where the result is a union of all the types in both sets. If there’s a bug there it’ll likely be caught elsewhere in the app because of the new type of the resulting set. For example:

```crystal
numbers = Set{1, 2, 3}
another_set = Set{“1”, “2”}

result = numbers - another_set # => Set(Int32){1, 2, 3}
result = numbers + another_set # => Set(Int32 | String){1, 2, 3, “1”, “2”}
```

I don’t think the `-` version makes any sense at all and if I’m doing that, it’s probably a bug. When adding them, it changes the type and is more likely to catch that bug. With this PR, the subtraction would be a bug.

IMO, the type checker is there to help you. It’s why a lot of people still use static languages. Something that becomes a compile-time error that could be a runtime error is often a good thing. Being able to tell the compiler you know what you’re doing by simply adding a type annotation is also good.

All that said, I found a way to let me remove the explicit type specifier. I don’t think I agree with it, but it keeps the same behavior in this scenario.

asterite: So I guess we should disallow `-` between sets of different types? That is, you can only do A - B if A's element types are a superset of B.

That would need to apply to all collection types.

The problem is always equality and inclusion, specially regarding `==`. Because `1` == `1.0`, and because `==` can be generally defined between any two types, when you have `Set{1, 2, 3}` then subtracting `Set{1.0}` could be valid...

Interestingly, in Ruby:

```
Set[1, 2, 3] - Set[1.0, 2.0, 3.0] # => #<Set: {1, 2, 3}>
```

Also:

```
irb(main):004:0> h = {1 => 2}
=> {1=>2}
irb(main):005:0> h[1.0]
=> nil
```

That's because Hash uses `eql?` and not `==` for comparing key values.

But then also in Ruby:

```
{a: 1} == {a: 1.0} # => true
```

but that's because it uses `==` for comparing elements, and `1 == 1.0 # => true`.

So maybe we can disallow `h[1.0]` for `Hash(Int32, V)`. However, if you have `Hash(Int32 | Float64)` and you have `1` as a key and you ask for `1.0` it will work and return the value even though with `Hash(Int32, V)` it wouldn't work... because we don't make the distinction of `==` and `eql?` like Ruby.

This is a hard problem... :-)

We'll need to decide if we want to introduce `eql?` to types, which might make sense in the context of union types and the way Ruby works.

Maybe we can also disallow `==` between any two types. For example comparing `Int32` vs `String` wouldn't compile, which might make sense. But we can allow comparison between any two reference types, that being `false` by default.

Sorry for the wall of text... my point is that this PR can't be thought in isolation will all of these other things. And changing that is a huge breaking change which I'm not sure we can/should do right before 1.0.

jgaskins: > So I guess we should disallow - between sets of different types? That is, you can only do A - B if A's element types are a superset of B.

In an ideal world, I think that makes sense.

> That would need to apply to all collection types.

Yeah, that's the catch. 🙂 It's why, even though I disagree with @bcardiff I still made it work without the explicit type annotation.

Your focus on consistency is important and I'm glad you called it out. Consistency can often be more important than anyone's perspective of what's "correct" because it allows you to guess at how something works. That's a super power for a programming language. Having to consult documentation all the time gets old quick.

I definitely want to have the conversation about whether `Array(Int32)#-(Array(String))` should compile, but that particular change is too far outside the scope of this PR and I apologize for hammering on it. All I wanted was for the compiler to tell me when I changed my hash key type that I'd missed some of my key checks. :joy:

>  if you have Hash(Int32 | Float64) and you have 1 as a key and you ask for 1.0 it will work and return the value even though with Hash(Int32, V) it wouldn't work... because we don't make the distinction of == and eql? like Ruby.

I've been using Ruby for 16 years and, honestly, I never understood why Ruby used `eql?` for that. But this _definitely_ sounds like it might be why.

jgaskins:
I ran one of my apps through this build and I only had two compile-time errors. One was the bug I was trying to squash with this PR itself and the other was in the `nats_io/nats.cr` shard, which _also_ turned out to be a bug.

- `nats_io/nats.cr`: Symbol key was being used with a JSON-parsed hash — it was impossible for this conditional to execute its block

  <details><summary>Error output</summary>

  ```
  $ ~/Code/crystal/bin/crystal src/scatternote.cr
  Using compiled compiler at /Users/jamie/Code/crystal/.build/crystal

  In lib/nats/src/nats/connection.cr:511:28

   511 | unless @server_info[:max_payload]?.nil?
                            ^
  Error: no overload matches 'Hash(String, JSON::Any)#[]?' with type Symbol

  Overloads are:
   - Hash(K, V)#[]?(key : K)
  ```

  </details>

- My own app code: I changed the hash to use an `AppChannel` (a concept within my app) as a key instead of a `String` property of that `AppChannel`, but accidentally left one place still using the `String` key. Once I fixed this compile-time error, my app was working. Before, it was silently failing.

  <details><summary>Error output</summary>

  ```
  $ ~/Code/crystal/bin/crystal src/scatternote.cr
  Using compiled compiler at /Users/jamie/Code/crystal/.build/crystal

  In src/scatternote.cr:215:41

   215 | if sockets = @channels[app_channel.subscription_key]?
                               ^
  Error: no overload matches 'Hash(Scatternote::AppChannel, Set(HTTP::WebSocket))#[]?' with type String
  Overloads are:
   - Hash(K, V)#[](key : K)
  ```

  </details>

No other compilation errors, despite using hashes all over the place (it's a Pusher-like app, so it has to map channel subscriptions to sets of websockets, connection ids to websockets, `AppChannel`s to `NATS::Subscription`s, etc) and you can see from my error output that I'm also using `Set`s, which was one of one the point of contention in this PR.

straight-shoota:
I've also tried this branch with a some shards. It helped identify a few bugs. In some cases explicitly handling `nil` values was needed, but that's not a huge annoyance and helps to clearly specify what happens in case of a `nil` value.

We should continue with these tests on more shards, but for now it really seems like this would be a great improvement!

asterite:
Definitely! I'd like to try this out at least for a release or two to see how much of a breakage it causes, but also how many bugs are detected.

We should also restrict the type of `Array#delete` and others where it makes sense.

spec/std/set_spec.cr
paulcsmith: I'm loving the discussion here. I also use Crystal in large part due to the type-safety and would prefer the A must be a superset of B. I've been bitten by things like this when developing the internals of Lucky where I'm working with data from outside the system (cookies, params, sql).

I personally would love a more strict equality with an "escape hatch" of some kind. Maybe that's with `eql?` or a different method. The key is that you have that escape hatch. So strict by default to catch bugs, with an escape hatch for if you want to get into the kinky stuff.

You could even do something similar to what we do in Lucky where there is a fallback method that generates a nice little error telling you why it happened and what you can do isntead:

```
Can't compare/set ekys/etc Objects of different types.

Make sure the objects are the type you expect, or use `some_other_method` to allow setting/getting/comparing objects of different types.

This way you get the safety, but you have the escape hatch along with a nice error. I think this will really make the lagnuage super powerful and easy for beginners.
```

paulcsmith:
Loving this change! I had a suggestion for adding helpful error messages to some methods if people have trouble with it, but just an idea. Would really love to have this 🎉

spec/compiler/crystal/tools/init_spec.cr
```diff
@@ -39,6 +39,10 @@ private def run_init_project(skeleton_type, name, author, email, github_name, di
   ).run
 end
 
+private def yaml(value)
+  YAML::Any.new(value)
```
straight-shoota, line 43: I'm not sure this extraction is worth it.

src/hash.cr
```diff
@@ -1022,7 +1022,7 @@ class Hash(K, V)
   # h = Hash(String, String).new
   # h["foo"] # raises KeyError
   # ```
-  def [](key)
+  def [](key : K)
     fetch(key) do
       if (block = @block) && key.is_a?(K)
```
straight-shoota, line 1027: Don't need `key.is_a?(K)` here anymore.

src/named_tuple.cr
```diff
@@ -72,15 +72,21 @@ struct NamedTuple
...
     {% begin %}
       NamedTuple.new(
       {% for key, value in T %}
-        {{key.stringify}}: self[{{key.symbolize}}].cast(hash.fetch({{key.symbolize}}) { hash["{{key}}"] }),
+        {% if K == String %}
+          {{key.stringify}}: self[{{key.symbolize}}].cast(hash[{{key.stringify}}]),
+        {% elsif K == Symbol %}
+          {{key.stringify}}: self[{{key.symbolize}}].cast(hash[{{key.symbolize}}]),
+        {% else %}
+          {% raise "Hash keys must be String or Symbol to convert into a NamedTuple" %}
```
straight-shoota, line 88: I think that's more accurate?
```suggestion
          {% raise "Hash keys must be all String or all Symbol to convert into a NamedTuple" %}
```

spec/compiler/crystal/tools/init_spec.cr
jgaskins: Maybe not. I mainly did it to make the code that uses it less cluttered, not for DRY purposes. To be fair, I'm not sure how much better it is.

```crystal
parsed["targets"].should eq({
  yaml("example_app") => {
    yaml("main") => "src/example_app.cr",
  },
})

# vs

parsed["targets"].should eq({
  YAML::Any.new("example_app") => {
    YAML::Any.new("main") => "src/example_app.cr",
  },
})
```

src/hash.cr
jgaskins: Nice catch!

src/compiler/crystal/semantic/type_declaration_processor.cr
```diff
@@ -463,7 +463,7 @@ struct Crystal::TypeDeclarationProcessor
     ancestor = owner.ancestors.first?
     ancestor = uninstantiate(ancestor)
     if ancestor
-      ancestor_non_nilable = @non_nilable_instance_vars[ancestor]?
+      ancestor_non_nilable = @non_nilable_instance_vars[ancestor]? if ancestor.is_a?(Type)
```
RX14, line 466: You can just put `if ancestor.is_a? Type` on the line above.

RX14:
I'm really worried about all these new PRs introducing big changes like this and enforcing `==` types. These are huge, fundamental changes, coming where we as a community are supposed to be slowing down - annealing towards 1.0.

I'm not opposed to this change per se, but I wish it'd been merged 2/3 years ago. It was proposed back then, and we rejected it. And I don't think this is the time to be revisiting these kinds of decisions or we may never see 1.0.

jgaskins:
@RX14 Your concern is totally reasonable and it's the main reason I've been trying to see what the impact is on real-world code. It seems like there's been a lot of worry about breaking existing code but the only actual breakages found in the wild with this PR (which a handful of people tested) have been bugs. Finding bugs like that was precisely the goal of this PR.

After looking at the other PRs for this feature, I didn't get the impression that anyone involved had seen actual bugs without this type check, so it seemed to still be in the hypothetical/theoretical range. But folks testing this PR have actually uncovered bugs with almost none of the pedantic side effects (other than `nil` checks) that people seemed to be the most worried about.

Discussion of enforcing `==` types probably came out of this (I didn't catch that other discussion, but I'll look for it) because Ary and I had some interesting back and forth about it here.

straight-shoota:
So, do we want to continue with this? I'd say a big YES because it has shown to help avoid bugs. Other than the recently merged exhaustive case feature this doesn't seem to cause much code to break, and if it does, it's likely an indicator for a bug.

waj:
Since we're trying to stabilize things and this is a big change in the type system, I'd like to see approvals from @bcardiff and @asterite before merging.

straight-shoota:
Just to be clear: This change doesn't affect the type system. It's all in stdlib and works without changes to the language.

asterite:
@straight-shoota Hash is ubiquitous so changing how it works affects every program out there. That's why I think this is a huge change and we should be careful about merging it right before 1.0.

That said, I think this is a good change but it feels like this is just half the way. For example one can still do `[1, 2, 3].includes?("foo")`, and if we really want type-safety we should disallow that too. I don't know if we have to go as far as disallowing `==` for different types, but `Enumerable#includes?`, `Array#delete`, `Set#delete` and `Deque#delete` are a few that come to mind.

spec/compiler/crystal/tools/init_spec.cr
```diff
@@ -146,7 +150,11 @@ module Crystal
 
         with_file "example_app/shard.yml" do |shard_yml|
           parsed = YAML.parse(shard_yml)
-          parsed["targets"].should eq({"example_app" => {"main" => "src/example_app.cr"}})
+          parsed["targets"].should eq({
+            yaml("example_app") => {
+              yaml("main") => "src/example_app.cr",
+            },
```
asterite, line 156: I think this is not very intuitive. What prevents us from not modifying equality at all in this PR?

asterite: That is, I'd like this change to be gone but still have it compiling and working.

paulcsmith:
Agree that this would be hugely helpful. IMO it's before 1.0 or it will never happen. Since this will catch a lot of bugs I think it should be in Crystal, but that's just my opinion :)

I also agree with the type restriction in other methods like `#delete` and friends! That would be ah-mazing!

bcardiff:
I've been checking how this impact via test-ecosystem. Being a breaking change it is expected to break things. I don't fear that.
Some of this are already mentioned in https://github.com/crystal-lang/crystal/pull/8893#issuecomment-597464869 , but I want to rephrase to share my thoughts.

So far, I've seen two main breaking things that impact too much IMO:

1. This PR makes it invalid to call `Hash(K, V)#has_key?(K | O)`. The same applies for `Hash#[]`.
2. `Hash` stops playing nice with key types that overrides `#hash` and `#==`.

Regarding (1) if the argument type, `Hash(K, V)#has_key?(K2)` would be such that there is no intersection with `K`, then I could consider it. 
But since, being `K2 = K | O` can hold values in `K` I don't think this PR is in the right direction.

And then, even if we detect empty value intersection between `K` and `K2`, or allow any `K2` where `K < K2`, there is still the (2) issue. 

The (2) issue appears in wrappers like `YAML::Any`. They keep the `#hash` and `#==` of the raw values. For example `String`. 
A `Hash(YAML::Any, V)` is no longer usable with `String` keys. That is uncomfortable. If we would consider going back to recursive alias instead of wrappers for `YAML::Any`
then this issue could be dismissed. But it's not the case.

So I've found that `Hash#hash_key?`, `Hash#[]`, `Hash#[]?` seems to be used mostly with `K?` as argument type or with a wrapper type. 

* If `nil` is passed as an argument the behavior without this PR are safe.
* If the wrapper implements correctly `#hash` and `#==` the behavior without this PR are safe.
* With this PR I still think things end up been more uncomfortable for the user.

As a reference some of the issues I detected so far via test-ecosystem are:

* In std-lib, `[1,"a"] - ["a"]` breaks since `Hash#has_key?` typing is strict in the key type. `Hash(String, String)#has_key?(Int32 | String)` is called. Of course this is solvable with another implementation, but the why Hash should be more terse and not Array?. This was mentioned in https://github.com/crystal-lang/crystal/pull/8893#discussion_r390991973
* In std-lib, before this PR `Hash(YAML::Any, YAML::Any)` could be indexed with plain `String`. This will no longer be the case even if `String` `#hash` and `#==` are in place. This is used in ameba and cli shards.
* In crystal-db, since `URI#scheme : String?`, the line `driver_class(uri.scheme)` can fail. `driver_class` already handle missing keys via a custom exception. With this PR the logic needs to be duplicated to have nice error messages, or change the Hash to be indexable by `String?`.
* In crystal-molinillo, since `State#name` is `String?` and that is used to access a `Hash(String, ...)`.
* In amber, `Hash(String, Amber::WebSockets::Channel)#[]?' is used with `JSON::Any`

Despite the temptation of more annotations and stricter interface in `Hash` I think this PR goes against some flexibility.
I acknowledge that some have found bugs, yet that was not my case so far.
I also agree with https://github.com/crystal-lang/crystal/pull/8893#issuecomment-620614610. I still feel that hash indexing is strongly related to equality https://github.com/crystal-lang/crystal/pull/8893#discussion_r390940922 . So, using different types is allowed on both or none. And I feel that that might be way to strict.

I prefer closing or at least not consider it before 1.0

paulcsmith:
Thanks for the thorough writeup and investigation. That all makes a lot of sense and helps us see why it may not be useful.

> I prefer closing or at least not consider it before 1.0

I'm a bit nervous about this because if it doesn't land pre-1.0 I don't think it ever will for a very long time since it will certainly break programs in the future. 

My thought would be that we have the "safer/stricter" version as the default, and have an overload for an "unsafe/less strict" version

Here's an example of what I mean:

```crystal
hash = {"name" => "James"}
# By default it should not compile since it will not work
# unless you define a custom `==` or `hash` as mentioned in bcardiff's post
hash.has_key?(:name) 

# Instead let people wrap types in an `Hash::AnyKey` (or some other name)
hash.has_key?(Hash::AnyKey.new(:name))
hash[Hash::AnyKey.new(:name)]

struct Hash::AnyKey(T)
  getter obj

  def initialize(@obj : T)
  end
end
```

So basically the methods have an overload for `Hash::AnyKey` so you can use any type you want. Yes, it is uglier, but it is much better for the common case where 90%+ of the time you want to use `K` and not something else. It has caught a few bugs in people's testing, and it would have caught some in Lucky that have since been fixed.

I think this is a great safe way to handle it, with an escape hatch (that is ugly), but works when you _really_ need it. 

Thoughts?

straight-shoota:
@bcardiff I don't agree to the assessment of the breaking examples.

* `[1,"a"] - ["a"]` should obviously work and it's only an implementation detail that this is based on `Hash`.
* `Hash(YAML::Any, YAML::Any)` is not `Hash(String, YAML::Any)` for a reason: keys can be anything. There's however a useful API for dealing with that: `YAML::Any#[]`. This PR already makes sure that it continues to work with string keys by wrapping them in an `Any`.
* In the crystal-db example, I'm sure the argument to `driver_class` should already be restricted to `String`. A driver for `nil` scheme is impossible to find. But the caller should handle the case of a schemaless URI because it's a part of URI parsing, not driver lookup. So I'd actually count that as identification of a flaw in the API.
* In molinillo, one of the methods where name is used as a hash lookup is `DependencyGraph::vertex_named!`. That method's documentation even states that `name` is expected to be a `String` (it's the original Ruby annotation, not a Crystal type restriction). When the intention is that this method should only accept `String`, calling it with a nilable argument is wrong. Again, to me that's a flaw in the API (or its usage).
* The amber example seems like it should take more care about converting `JSON::Any` to Crystal data types. What if that `Any` is a bool or int, or even an object? I'm sure that should result in a proper error message instead of just failing to look up a channel subscription under that key.

@paulcsmith I'm sure there could be alternatives for retaining the current behaviour. It could be as simple as having a separate type that inherits from `Hash`:
```cr
class AnyHash < Hash
  def [](any_key)
  end
  def []?(any_key)
  end
  def has_key?(any_key)
  end
end
```

But I'd very much appreciate if the default `Hash` implementation would have type safe keys because I'm convinced that this is more useful for catching bugs than it causes inconveniences.

oprypin:
I don't support the push to get breaking changes in right before 1.0. If you find yourself thinking like that, it's already too late. It's **the worst** time to do it, as 1) newcomers excited about 1.0 come to the ecosystem of libraries in total disrepair and 2) you have no opportunity to course-correct in case of a mistake.

(Now, I also don't support the push for 1.0 itself at this time, but that's another topic)

oprypin:
As for the change itself, I personally have never ever felt the need for this. I also think that the suggested behavior is less Crystal-like than the current one. The improvement is arguable at best and is a big enough breakage (thanks @bcardiff for the investigation) to definitely bring it below the bar of acceptance for me. Appreciate the effort, though.

jgaskins:
@bcardiff Thank you for taking the time to look into this more. I can appreciate the balance you’re trying to strike and want to address some of the points you made.

> This PR makes it invalid to call `Hash(K, V)#has_key?(K | O)`. The same applies for `Hash#[]`.

This is a really great point I thought I’d covered, but that was for `Hash(K, V)#==(Hash(K | O, V))`. Is there a way to make it work similarly?

> `Hash` stops playing nice with key types that overrides `#hash` and `#==`.

I feel like this actually adds confusion. Data structures in Crystal are inherently about types but this point seems to pretend we’re duck-typing hashes.

I love that duck-typing is indeed a thing in Crystal, but object state is not duck-typable. Types are a hard requirement there — in a `Hash(K, V)`, the key _must_ be a `K`, not something that quacks like a `K`. Pretending otherwise is more confusing than helpful, IMO.

For example, I can’t think of a reason `{1.0 => “a”}[1] == “a”` evaluating to `true` would ever make sense and I didn’t know it did until @asterite pointed it out in this PR.

---

Regarding your bullet points, I sympathize with you on them because you’re trying to balance convenience, existing code, and type safety, but @straight-shoota made some great counterarguments that I not only agree with but would like to extend a couple of:

> In std-lib, before this PR `Hash(YAML::Any, YAML::Any)` could be indexed with plain `String`. This will no longer be the case even if `String` `#hash` and `#==` are in place. This is used in ameba and cli shards.

This goes along with what I mentioned above about false duck typing. It seems convenient until you have to hunt down a bug.

That said, I haven’t had to work directly with YAML in Crystal. I’ve rarely had to do it in Ruby, which I’ve used professionally for 16 years and it uses YAML everywhere. I can’t imagine a significant number of people will be affected by this. And for those few who are, wrapping string keys in a `YAML::Any` (as @straight-shoota mentioned) is all that’s needed.

> In crystal-db, since `URI#scheme : String?`, the line `driver_class(uri.scheme)` can fail. `driver_class` already handle missing keys via a custom exception. With this PR the logic needs to be duplicated to have nice error messages, or change the Hash to be indexable by `String?`.

I apologize if this comes across indelicately, but how this is different from everywhere else I have to handle nils in every Crystal program I write? I feel like this one of the biggest wins of `Nil` being its own type in Crystal, that you have to consider the case where a value can be `nil`. Letting a `nil` propagate here feels like it goes against that.

> In amber, `Hash(String, Amber::WebSockets::Channel)#[]?` is used with `JSON::Any`

It seems Amber can simply call `.as_s` here, no? It’s something I’d do in my own apps because I didn’t realize `JSON::Any` overrode `#hash` and `#==` to allow this shorthand. And I’d honestly be confused if I saw a `Hash(String, V)` accessing keys with a `JSON::Any` in one of my own apps.

---

Again, I appreciate your thoughts and opinions here just like I appreciated discussing it with @asterite in other places in the thread, even (and maybe especially) when we were disagreeing. These are _excellent_ things to discuss because the more perspectives we apply to this problem the better we’ll understand it.

However, I don’t think the problems you mentioned are solved by being permissive with Hash key types. Instead, I feel like that permissiveness allows solutions to be applied that are mildly convenient in the moment and problematic later on. And for cases when you just want the convenience, the permissive hash implementation described by @straight-shoota and @paulcsmith is easy enough to bolt on.

paulcsmith:
I tend to agree with @straight-shoota and @jgaskins. I think the types are important. I know some people haven't run into this, but many more have. And I think the Amber example (as mentioned) should probably have a more strict type anyway and this would encourage that.

I think I'm having trouble seeing the long term downside. Short term downside is of course breakage of existing libs, but long term downside I think is not really a problem. The long-term upside is it can catch bugs that many have run into.

Plus, there are ways to fallback like @straight-shoota showed if people really need the escape hatch.

> I don't support the push to get breaking changes in right before 1.0.

I think waiting will be even worse. If we don't do it now, it likely will never be moved in. It is to too fundamental a change. Adding it later will cause even bigger breakage for more programs (more people using Crystal) and you'll introduce fragmentation in the ecosystem. Pre-1.0 is likely the only time it can be added.

> right before 1.0

This part I do agree with. It may mean pushing back 1.0 a bit to get the ecosystem up to date. I think making Crystal more type-safe (which is the goal of most of the Crystal APIs)

My thought here is that we *absolute have to have overridable dependencies* in shards before merging this in. So people can fork, update, and use shards without waiting for maintainers.

bcardiff:
I like the intent of this PR. And I know there are ways to code things differently, in a less quacky fashion.

Between how numbers equality and std-lib JSON/YAML wrappers works, I think the current status is that _a value_ can belong / have their equivalent in multiple types. Even beyond union types.

It is subtle, but it is something that has been around for a long time and I hardly believe it's the time for breaking that.

One scenario is that maybe wrappers would need to go back to recursive aliases to be more comfortable to use and allow the changes in this PR. That has its own challenges.

Regarding @paulcsmith proposal, I think something along refinements; where you can define in the lexical scope, how you intend to use a value and operations; would be a nicer approach. Otherwise, you will be stuck with wrapping/unwrapping or converting.

I think that **everybody involved in the discussion agree with the facts exposed**. What will break, what will not. **Where we disagree is whether is a right choice to do it now**. I think it breaks a lot of behavior.

The idea of 1.0 is to acknowledge more or less the current state of Crystal that has been used to built a lot of things. This PR is really a hard stopper for that idea.

---

Now, thinking out loud. Maybe another course of action is to enable this behind a preview flag: `strict-hash`. Code written for the `strict-hash` will also work without it. The opposite is not true clearly.

That could allow some iteration and discover more edge cases, which I believe they are. But this idea is post 1.0 IMO.

paulcsmith:
@bcardiff I think you are right we agree on the facts 👍 

Could you share how this would work post 1.0? I'm not sure I understand a path forward, but maybe I'm thinking with too narrow a vision! I'm all for doing it post 1.0 I just don't understand how it'd work without causing ecosystem fracturing

HankAnderson:
@paulcsmith Very simple: the change is introduced in 2.0, where breaking changes will be allowed. Or that's my guess.

paulcsmith:
@HankAnderson I think programming languages have a *much* harder time introducing backwards incompatible changes. Python made changes in 2 -> 3 but was not simple and to this day causes issues and a fractured ecosystem. So I think we need to be really careful since this change would change Hash and Array, 2 very frequently uses classes. 

For example, Ruby 2.0 was almost 100% compatible with Ruby 1.0. https://www.ruby-lang.org/en/news/2013/02/24/ruby-2-0-0-p0-is-released/

> We have also taken care with the 2.0.0 design to make it compatible with 1.9. It will be easier to migrate from 1.9 to 2.0 than it was from 1.8 to 1.9. (The notable incompatibilities are described later.)

The tiny changes that were breaking were fairly esoteric and quite easy to fix.

So I was hoping to learn more about how this could be a phased roll-out, or opt-in, or some other way to make it so it doesn't fracture the ecosystem. It sounds like some kind of opt-in compile time flag could be a way to slowly roll it out and require it in a future release

bcardiff:
@paulcsmith exactly as you said, a compile time flag could help rolling out changes like this one IMO. The process can be:

* on version A, a --preview-X is allowed. Ecosystem can choose to support it or no. What ever compiles with X, should compile and work without it.
* on version A+n, we can choose to make X the default. But and introduce a --without-X for those that didn't jump into the X party
* on version A+n+m, drop --without-X / --preview-X. And leave X as default.
* somewhere between A and A+n we could decide that X is not worth it.

This process was done in with overflow actually. And it's not a guarantee that issues will not be discovered after X is made stable.

For whatever feature X that is introducing a stricter behaviour, I think the above process work. The language and std-lib can still evolve. For example, there was the idea to avoid mixing integer types, or at least signed and unsigned. The same path can be taken for that feature. The recent exhaustive case could have been done with this approach I think. Yet, it would have iterated more slowly maybe.

Of course having flags make things more difficult to diagnose when things are not working as expected. The alternative of breaking things in a harder way is not good IMO.

Either as a shard or tweaking the crystal_path, it is possible to check these feature your own applications and gather feedback. I explained before the issues I found in test-ecosystem, those are solvable but were not an exhaustive list and I timeout after descending deep in the dependency & forking shards that hasn't been updated in a while, but are used as dependencies here and there.

Finally, although we might go from 0.35 to 1.0, that does not mean that we will need to wait for 1.40 before jumping to 2.0. There is no timeline settle.

paulcsmith:
@bcardiff thanks for the explanation! That makes a lot of sense 👍

HertzDevil:
Can we do this with normal deprecated overloads instead?

src/hash.cr
```diff
@@ -1753,13 +1753,24 @@ class Hash(K, V)
...
+      return false unless size == other.size
+      each do |key, value|
+        entry = other.find_entry(key.as(K2))
+        return false unless entry && entry.value == value
+      end
+      true
+    {% else %}
+      false
+    {% end %}
   end
```
HertzDevil, line 1774: `Hash#==` would then fail if the argument is a union of `Hash` instances, so one level of indirection is needed here:

```crystal
class Hash(K, V)
  def ==(other : Hash)
    other.equals_impl(self)
  end

  protected def equals_impl(other : Hash(K2, V2)) forall K2, V2
    # ...
  end
end
```

spec/std/set_spec.cr
HertzDevil: This would require something like https://github.com/crystal-lang/crystal/issues/6997#issuecomment-450353013:

```crystal
struct Set(T)
  def -(other : Set(U)) forall U <= T
    # ...
  end

  @[Deprecated("Ensure *other*'s element type is `T` or a subtype of it")]
  def -(other : Set)
    dup
  end
end
```

The story is more complex if we want to allow unions of `Set` instances as arguments (mainly due to #8973).

wolfgang371:
I ran in this common trap these days as well. I wanted to implement a simple cache and was dealing with both `Set`s and `Hash`es. When playing around I switched one variable from the former to the latter, but left the method unchanged, effectively having something like this:
```crystal
hash_cache.includes?(key)
```
It took me quite a couple of hours figuring out what was happening here; my "cache" was slowing things down instead of speeding up. Of course, my call from above always returned `false`, because the "key" type didn't even match. I was completely blind on this spot, fully relying on the compiler...

Just today I found there are already several discussions on this topic, this seemingly to be the newest (and open) still.

I wanted to be extra smart and then at least have a strict version for myself...
```crystal
# default Hash signature / implementation
# class Hash(K, V)
#     def has_key?(key) : Bool # no type, no type checks; dito #includes? and some others
#         # ...
#     end
# end

# trying to do monkey-patching
class Hash(K, V)
    private def check_key(key : T) forall T
        {{ puts T, K, T.is_a?(K.class) }}
        # raise("Hash argument type mismatch") if !T.is_a?(K.class) # works (checks), but only on runtime
        # {{ raise("Hash argument type mismatch") if !T.is_a?(K.class) }} # compile time, but doesn't work (compiler always stops)
    end
    def has_key?(key) : Bool
        check_key(key)
        previous_def
    end
end

h = {"1"=>1}
p h.has_key?("1") # target: should compile
p h.has_key?(1) # target: should not compile
```
... but I was not even able to do a proper monkey patch.
Seems I'm once more lost with Crystal's macro language (see items (1) and (2) below)...
Ah yeah, and (3): `check_key` probably should be a macro as well...

```crystal
{{p String.is_a?(String.class) }} # false (compile-time, gets emitted first in console) - shouldn't this be the same as (1) below??
p String.is_a?(String) # false
p String.is_a?(String.class) # true (1)
p (Int32).is_a?((Int32|Float64).class) # false, but we need sth. that returns true as well... (2)
p (Int32|Float64).is_a?((Int32|Float64).class) # true
```

Any ideas? Should I file an issue for the deviation of (1) and (2), ...?
Thank you!

straight-shoota:
The type comparisons are both flawed:
* Runtime `T.is_a?(K.class)` technically works, but it's equivalent to `T == K` and requires the allowed type to match `K` _exactly_. Maybe this is intended for your use case, but a generic implementation should probably allow any subtype of `K` to match as well. This could be expressed as `T <= K`, or rather `key.is_a?(K)`.
* Macro `T.is_a?(K.class)` is entirely different: The type of `T` is `Crystal::Macro::AST::TypeNode`, not a metaclass. So this could never be true. A correct expression would again be `T == K` or better `T <= K`.

However, you don't need any of this and can leave the type matching to the compiler. All you need is a proper type restriction for the `key` parameter. No method body necessary:
```cr
private def check_key(key : K)
end
```

This is basically what this PR does as well (just directly on the public methods).

wolfgang371:
Works for me like a charm - just for `#[]?` I decided it's safer/better to leave it the unrestricted way.

```crystal
# make Hash more restrictive
class Hash(K, V)
    # def []?(key) # we are less restrictive for #[]?
    #     check_key(key)
    #     previous_def # directly from Hash
    # end
    def [](key)
        check_key(key)
        previous_def # directly from Hash
    end
    def has_key?(key) : Bool
        check_key(key)
        previous_def # directly from Hash
    end
    def includes?(obj) : Bool
        check_keyvalue(obj)
        super # indirectly from Enumerable
    end
    private def check_key(key : K)
    end
    private def check_keyvalue(keyvalue : {K,V})
    end
end
```

PR #8894, opened 2020-03-10
Fix document of Int#gcd

nard-tech:
Tiny fix to document of `Int#gcd`

```crystal
# NG
5.gcd(10) # => 2

# OK
5.gcd(10) # => 5
```


yxhuvud:
Ooopsie.

PR #8899, opened 2020-03-11
Avoid using the default --exclude-warnings value if some is specified

bcardiff:
Currently there is no way to exclude the default value of lib from the `--exclude-warnings` path.

This PR solves that. If the used mentions some `--exclude-warnings` the default value it is discarded.

This will allow some more exhauistive deprecation checks for dependencies when wanted.

Note: It would be great if the `OptionParser` supported some better constructs to detect if an option was not explictly used, or some way to inject default values. But with the current  features, this is the best I could do.

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -549,7 +549,7 @@ class Crystal::Command
 
   private def setup_compiler_warning_options(opts, compiler)
     compiler.warnings_exclude << Crystal.normalize_path "lib"
-    compiler.warnings_exclude_default = true
+    warnings_exclude_default = true
     opts.on("--warnings all|none", "Which warnings detect. (default: all)") do |w|
       compiler.warnings = case w
                           when "all"
@@ -565,10 +565,10 @@ class Crystal::Command
       compiler.error_on_warnings = true
     end
     opts.on("--exclude-warnings <path>", "Exclude warnings from path (default: lib)") do |f|
-      if compiler.warnings_exclude_default
+      if warnings_exclude_default
         # if an --exclude-warnings is used explicitly, then we remove the default value
         compiler.warnings_exclude.clear
-        compiler.warnings_exclude_default = false
+        warnings_exclude_default = false
       end
       compiler.warnings_exclude << Crystal.normalize_path f
     end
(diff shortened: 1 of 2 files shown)


straight-shoota:
Would it not be possible to add `Crystal.normalize_path "lib"` only if `compiler.warnings_exclude` is empty?

Apart from the implementation details: This would mean when any `--exclude-warnings` option is given, warnings from `lib` suddenly start showing up, despite being hidden by default? That seems very much unexpected. I'd figure you usually want to exclude the specific path given as an option *plus* `lib`. Not instead of it. For this use case this seems like really bad UX.
Or am I missing something?

bcardiff:
> Would it not be possible to add Crystal.normalize_path "lib" only if compiler.warnings_exclude is empty?

Only if that is done in the `Compiler`. If the `OptionParser` would have an after hook that could be done there.

> This would mean when any --exclude-warnings option is given, warnings from lib suddenly start showing up.

Yes. It's a default. If there is another value given, the default is gone. The alternative is adding a `--include-warnings` but is seems over-complicated since it will be used only for the `lib`. (Or maybe the whole CRYSTAL_PATH since that would be another possible default 🤔)

straight-shoota:
Why would we need a hook for that? It could just be applied after `OptionParser#parse` is called.
Since that's distributed over multiple location it might not be straightforward to keep it dry. But it should work and be contained to `Compiler::Command`.

Maybe a better alternative would be to only remove the default `lib` folder if a path starting with `lib` is explicitly given as an option. That way you could use for example `--exclude-warnings lib/notexist` to include evetything in `lib` (except for `notexist`).

bcardiff:
> Why would we need a hook for that? ... Since that's distributed over multiple location

That's the reason. Since the `opts.on` calls are scattered by concerns, the `.parse` calls are not in the same scope. Preserving the state to check if the `.on` was called or not needs to be tracked somewhere.

I would have preferred to code something like

```crystal
private def setup_compiler_warning_options(opts, compiler)
  exclude_warnings_called = false
  opts.on("--exclude-warnings <path>", "Exclude warnings from path (default: lib)") do |f|
    exclude_warnings_called = true
    compiler.warnings_exclude << Crystal.normalize_path f
  end
  opts.after_parse do 
    compiler.warnings_exclude << Crystal.normalize_path "lib" unless exclude_warnings_called
  end
end
```

So the state `exclude_warnings_called` is closured and the default logic is together with the parsing.

I think that what you are proposing: explicitly calling a method to set the default after parsing is more error prone and tear appart something that is now kept together and make sense.

straight-shoota:
Couldn't `warnings_exclude_default` even be a closured local variable?

bcardiff:
@straight-shoota done, why I didn't do it that way in the first place?! 😆

PR #8900, opened 2020-03-11
Use CRYSTAL_OPTS env variable as default compiler options

bcardiff:
This PR make the compiler read the `CRYSTAL_OPTS` environment variable and use is as additional compiler flags in some commands: build, eval, run, spec and <file>.

This is handy to configure the compiler to be more pedantic regarding deprecations for example. The compiler is sometimes used in postinstall tasks where injecting flags is not possible.

With this and #8899 we should be able to checks deprecation along all the source code used to build and app.

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -68,6 +68,7 @@ class Crystal::Command
       init
     when "build".starts_with?(command)
       options.shift
+      use_crystal_opts
       result = build
       report_warnings result
       exit 1 if warnings_fail_on_exit?(result)
@@ -86,12 +87,15 @@ class Crystal::Command
       env
     when command == "eval"
       options.shift
+      use_crystal_opts
       eval
     when "run".starts_with?(command)
       options.shift
+      use_crystal_opts
       run_command(single_file: false)
     when "spec/".starts_with?(command)
       options.shift
+      use_crystal_opts
       spec
     when "tool".starts_with?(command)
       options.shift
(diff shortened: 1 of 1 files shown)


asterite:
Just bikesheddding, but why `CRYSTAL_OPTS` and not `CRYSTAL_OPTIONS`?

bcardiff:
No special reasons. I'm happy to change it.

asterite:
I'm fine with that name too, I was just curious.

src/compiler/crystal/command.cr
```diff
@@ -585,4 +590,8 @@ class Crystal::Command
     @color = false if ARGV.includes?("--no-color") || ENV["TERM"]? == "dumb"
     Crystal.error msg, @color, exit_code: exit_code
   end
+
+  private def use_crystal_opts
+    @options = ENV.fetch("CRYSTAL_OPTS", "").split.concat(options)
```
oprypin: Is this a declaration that nobody will ever need to use spaces within options to Crystal compiler?

Would be nice to support shell quoting.

asterite: Good catch. Definitely an improvement.

bcardiff: Is there a safe already way to split that respecting shell quoting?

asterite: No. We'd need to port this from Ruby: https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html (or do something similar)

PR #8903, opened 2020-03-12
Optimized Implementation of `Array#fill` for Zero Values

toddsundsted:
See discussion here:

https://forum.crystal-lang.org/t/fastest-and-iconic-way-to-zero-an-array/1799/16

This PR provides a high performance (~10x) means to initialize arrays of integers to zero by optimizing that case in `Array#fill`.

```
          fill(0)   2.50M (400.01ns) (± 1.82%)  0.0B/op   1.43× slower
       fill(0, 0)   2.52M (397.50ns) (± 2.39%)  0.0B/op   1.42× slower
fill(0, 0, 1_000)   3.57M (279.82ns) (± 2.95%)  0.0B/op        fastest
      fill(0, ..)   2.48M (403.21ns) (± 3.81%)  0.0B/op   1.44× slower
```
```
          fill(0)  23.72M ( 42.16ns) (± 4.14%)  0.0B/op   1.15× slower
       fill(0, 0)  25.42M ( 39.34ns) (± 2.81%)  0.0B/op   1.08× slower
fill(0, 0, 1_000)  27.29M ( 36.65ns) (± 5.10%)  0.0B/op   1.00× slower
      fill(0, ..)  27.37M ( 36.54ns) (± 3.41%)  0.0B/op        fastest
```

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -861,7 +861,17 @@ class Array(T)
   # a.fill(9) # => [9, 9, 9]
   # ```
   def fill(value : T)
-    fill { value }
+    {% if Int::Primitive.union_types.includes?(T) %}
+      if value == 0
+        to_unsafe.clear(size)
+
+        self
+      else
+        fill { value }
+      end
+    {% else %}
+      fill { value }
+    {% end %}
   end
 
   # Replaces every element in `self`, starting at *from*, with the given *value*. Returns `self`.
@@ -873,7 +883,21 @@ class Array(T)
   # a.fill(9, 2) # => [1, 2, 9, 9, 9]
   # ```
   def fill(value : T, from : Int)
-    fill(from) { value }
+    {% if Int::Primitive.union_types.includes?(T) %}
+      if value == 0
+        from += size if from < 0
+
+        raise IndexError.new unless 0 <= from < size
+
+        (to_unsafe + from).clear(size - from)
+
+        self
+      else
+        fill(from) { value }
+      end
+    {% else %}
+      fill(from) { value }
+    {% end %}
   end
 
   # Replaces every element in `self`, starting at *from* and only *count* times,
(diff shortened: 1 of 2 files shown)


asterite:
Thank you for this! ❤️ 

> Using compiled compiler at .build/crystal

Just a note: when making changes to the standard library (not the compiler) there's no need to compiler the compiler to try things out. Just using `bin/crystal` will pick the new standard library (while `crystal` will use the installed standard library).

toddsundsted:
6769781 adds support for arrays for floating point numbers, as well. this should cover arrays of all non-union numeric types, and should be suitable for most computational purposes.

src/array.cr
```diff
@@ -861,7 +861,7 @@ class Array(T)
   # a.fill(9) # => [9, 9, 9]
   # ```
   def fill(value : T)
-    {% if Int::Primitive.union_types.includes?(T) %}
+    {% if Int::Primitive.union_types.includes?(T) || Float::Primitive.union_types.includes?(T) %}
```
Sija: Perhaps:

```suggestion
    {% if Number.all_subclasses.includes?(T) %}
```

toddsundsted: much better. i will fix.

asterite: NOOOOOOOOOOOOO! Please don't.

This will include BigInt, BigDecimal, etc., and the code will work WRONG.

asterite: @toddsundsted Please wait until I confirm before making any further changes.

toddsundsted: i tested before changing, but now i suspect these become subclasses only when required:

https://play.crystal-lang.org/#/r/8ptx

I'll confirm and fix.

asterite: Yes, subclasses depends on the types you require. Also anyone can subclass `Int` or `Float` so it's pretty bad if we blindly apply this optimization to all numbers.

toddsundsted: yep, boom.

reverting.

toddsundsted: thank you!

RX14:
Didn't even have time to merge it myself :)

toddsundsted:
woo! my first legit contribution!

PR #8908, opened 2020-03-16
IO::Hexdump: forward missing methods to underlaying IO

carlhoerberg:
Allows you do things like `IO::Hexdump#pos` when the underlaying IO supports it.

### DIFF
--- a/src/io/hexdump.cr
+++ b/src/io/hexdump.cr
@@ -40,27 +40,5 @@ class IO::Hexdump < IO
     end
   end
 
-  def peek
-    @io.peek
-  end
-
-  def closed?
-    @io.closed?
-  end
-
-  def close
-    @io.close
-  end
-
-  def flush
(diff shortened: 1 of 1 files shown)


asterite:
I'm a bit hesitant to use `forward_missing_to`. If you look at the stdlib, we only use it in `HTTP::Headers` and even there I'm not sure we should be using it.

Why not call `pos` on the underlying IO?

carlhoerberg:
Ok, I'll update the PR later today

Sija:
Or maybe use a `delegate :peek, :close, :closed?, :flush, :tty?, :pos, to: @io`?

Sija:
@carlhoerberg https://github.com/crystal-lang/crystal/pull/8908#issuecomment-600734492

straight-shoota:
Thank you!

PR #8911, opened 2020-03-16
Fix File.dirname with unicode chars

### LINKED ISSUE #8910: File.dirname is broken for directories ending with a non-latin letter
This code:
```crystal
%w[
  /τελεία/τελεία
  /Månedå/Månedå
].each { |dir| pp File.dirname dir }
```
produces
```
"/τελεί\xCE"
"/Måned\xC3"
```
https://carc.in/#/r/8q5k

---
Crystal 0.33.0 [612825a53] (2020-02-14)

LLVM: 8.0.0

asterite:
Fixes #8910

There are a lot of `+ 1` in the code that might be worth reviewing. If someone wants they can try adding unicode tests to all of the `File` or `Path` methods to make sure they work well.

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -279,6 +279,7 @@ describe "File" do
     File.dirname("/Users/foo/bar.cr").should eq("/Users/foo")
     File.dirname("foo").should eq(".")
     File.dirname("").should eq(".")
+    File.dirname("/τελεία/τελεία").should eq("/τελεία")
   end
 
   it "gets basename" do
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -279,7 +279,7 @@ describe "File" do
     File.dirname("/Users/foo/bar.cr").should eq("/Users/foo")
     File.dirname("foo").should eq(".")
     File.dirname("").should eq(".")
-    File.dirname("/τελεία/τελεία").should eq("/τελεία")
+    File.dirname("/τελεία/łódź").should eq("/τελεία")
   end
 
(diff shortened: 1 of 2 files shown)


spec/std/file_spec.cr
```diff
@@ -279,6 +279,7 @@ describe "File" do
     File.dirname("/Users/foo/bar.cr").should eq("/Users/foo")
     File.dirname("foo").should eq(".")
     File.dirname("").should eq(".")
+    File.dirname("/τελεία/τελεία").should eq("/τελεία")
```
Sija: IMO it would be better to have a different path parts, in order to make sure that a proper piece is returned it - as now it could any of those.

asterite: Good idea. Thanks!

straight-shoota:
I'll take a look at all implementations to make sure they work with unicode strings.

straight-shoota:
The rest of `Path` seems to be good. We could consider adding lots of specs with unicode paths to make sure there won't be any issues in the future.

asterite:
@straight-shoota Thank you for checking! If everything is good then I think we don't need to add those tests.

Sija:
> @straight-shoota Thank you for checking! If everything is good then I think we don't need to add those tests.

Well, it's good _for now_, we just need to subtly break the implementation to find out we need these tests anyway, so why for it if we can be prepared? 👑

straight-shoota:
Yes, having such test would be great. But it requires also quite some effort to implement and review. IMO that effort is currently better put to use elsewhere.

PR #8912, opened 2020-03-18
Add CSV#rewind

asterite:
For https://forum.crystal-lang.org/t/how-to-iterate-csv-objects-multiple-times/1831

I think this makes sense to have and it's convenient.


asterite:
\o/ 🎉

PR #8916, opened 2020-03-19
Add Number#positive? and #negative?

nard-tech:
For example, when an array `[-2, -1, 0, 1, 2]` is given, I'd like to write
```crystal
[-2, -1, 0, 1, 2].select(&.positive?)
[-2, -1, 0, 1, 2].select(&.negative?)
```
instead of
```crystal
[-2, -1, 0, 1, 2].select { |i| i > 0 }
[-2, -1, 0, 1, 2].select { |i| i < 0 }
```

So I added `Number#positive?` and `#negative?` like Ruby.

### DIFF
--- a/spec/std/number_spec.cr
+++ b/spec/std/number_spec.cr
@@ -174,6 +174,30 @@ describe "Number" do
     1f32.zero?.should eq false
   end
 
+  it "test positive?" do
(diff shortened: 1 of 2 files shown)


Sija:
Duplicate of #7917

PR #8919, opened 2020-03-20
copy_file_range is a Linux-only optimization for FD to FD copying

carlhoerberg:
WIP, request for comments


carlhoerberg:
```
copy_file_range 4.0kiB 271.66k (  3.68µs) (± 1.35%)  0.0B/op        fastest
           copy 4.0kiB 261.68k (  3.82µs) (± 1.29%)  0.0B/op   1.04× slower
copy_file_range 1.0MiB   3.95k (253.02µs) (± 1.87%)  0.0B/op        fastest
           copy 1.0MiB   1.53k (653.40µs) (± 1.11%)  0.0B/op   2.58× slower
copy_file_range 128MiB  21.18  ( 47.22ms) (± 7.38%)  0.0B/op        fastest
           copy 128MiB   9.94  (100.62ms) (± 7.93%)  0.0B/op   2.13× slower
```

carlhoerberg:
XFS
```
copy_file_range 4.0kiB 470.95k (  2.12µs) (± 0.31%)  0.0B/op        fastest
           copy 4.0kiB 402.23k (  2.49µs) (± 0.13%)  0.0B/op   1.17× slower
copy_file_range 1.0MiB   7.15k (139.95µs) (± 0.14%)  0.0B/op        fastest
           copy 1.0MiB   2.45k (408.85µs) (± 0.36%)  0.0B/op   2.92× slower
copy_file_range 128MiB  25.69  ( 38.93ms) (± 6.52%)  0.0B/op        fastest
           copy 128MiB  12.72  ( 78.59ms) (± 0.27%)  0.0B/op   2.02× slower
```

asterite:
Nice!

Some things:
- What's the source code of the benchmarks?
- I think we should do the change innline inside the original `copy` method. Having an overload works too but it's a lot of code duplication.

carlhoerberg:
```crystal
require "./src/stdlib/libc"
require "./src/stdlib/io"
require "benchmark"

tmp = ARGV.shift? || "/tmp"
{ 4096, 1024**2, 128 * 1024**2 }.each do |size|
  File.open(File.join(tmp, "a"), "w+") do |a|
    File.open(File.join(tmp, "b"), "w") do |b|
      a.delete
      b.delete
      a.print "a" * size

      Benchmark.ips do |x|
        x.report("copy_file_range #{size.humanize_bytes}") do
          a.rewind
          b.rewind
          IO.copy_file_range a, b, size
        end
        x.report("copy #{size.humanize_bytes}") do
          a.rewind
          b.rewind
          IO.copy a, b, size
        end
      end
    end
  end
end
```

asterite:
Another thing: `IO::FileDescriptor` is buffered. So what happens if you read a bit from it, then use this optimization? Since it goes directly to the file descriptor, it would not copy the parts that are already buffered. So we should be really careful about that. Maybe only use the optimization when the buffer is empty? Or maybe just do this in a specialized `File.copy(filename1, filename2)` thing? I'm not sure...

asterite:
Or if there's some part in the buffer, copy that first, then call `copy_file_range`. I see Ruby uses `copy_file_range` is available, but it's still tricky.

ysbaddaden:
Maybe flush somehow?

asterite:
Right now we have `IO#peek` that will return a slice to buffered data if available. Combined with `skip` we can avoid introducing a new method (buffered_read).

However, I see `IO::Buffered#peek` will perform a read if the buffer is empty, probably to peek some data and be able to process it in memory (we use that in a few places).

So maybe we can provide direct access to the underlying slice, documenting it as unsafe, maybe calling it `buffer`. Then you would use that, then call `skip`, which might be better than introducing a method specifically to read from the buffer. It should also be faster because you avoid copying the internal buffer to a different slice.

But let's first discuss the implementation here or in a separate issue. It's a bit hard to discuss when code is changing all the time.

asterite:
I also see that you added `sendfile`, which Ruby does too. I think it's good, but I suggest to open an issue to discuss this and possible implementations.

src/io.cr
```diff
@@ -1148,6 +1148,32 @@ abstract class IO
...
+
+      dst.write(src.read_buffer)
+      remaining -= src.read_buffer.size
+      src.skip(src.read_buffer.size)
+      dst.flush
+
+      while remaining > 0
+        len = LibC.copy_file_range(src.fd, nil, dst.fd, nil, remaining, 0)
+        if len == -1
+          raise Errno.new "copy_file_range"
```
Sija: `Errno` is no moar :)

carlhoerberg:
How about EAGAIN and hooking into the event loop?

src/io.cr
carlhoerberg: oh yeah? what should be used instead?

Sija: yup, see #8885

RX14:
EAGAIN can't happen on `File` IO, file copies are always blocking on linux.

I think a simple heuristic of flushing both IOs and then using `copy_file_range` when the copy request is above a certain size would be enough.

RX14:
Also is it worth applying this optimization to `IO.copy` without a given size? You can pass `SizeT::MAX` to `len`.

carlhoerberg:
I now use Int32::MAX in IO.copy(src, dst), because with `SizeT`:
```
Error: undefined constant SizeT::MAX
```

asterite:
It's `LibC::SizeT::MAX`

carlhoerberg:
Of course, thanks. 

However, doing that (which is a UInt64 on my machine) results in
```
copy_file_range: Value too large for defined data type 
```

But by using `LibC::SSizeT` (Int64), it works. But the man page states that parameter should be a `size_t`, documentation error? http://man7.org/linux/man-pages/man2/copy_file_range.2.html

carlhoerberg:
`copy_file_range` was first added to musl in 1.1.24 (https://git.musl-libc.org/cgit/musl/tree/WHATSNEW), released October 13, 2019, and in glibc 2.27, released on 2018-02-01. Not sure how to act if the current version is older?

RX14:
I think those are a bit too new to add yet :(

Do all currently-supported versions of debian have a libc which is up-to-date enough?

asterite:
I _think_ with https://github.com/crystal-lang/crystal/pull/8336 it should be possible to know whether there's a `copy_file_range` function or not, I'm not sure...

yxhuvud:
> I think those are a bit too new to add yet :(

I wonder how a `splice` based implementation would compare. It would take a little more to set up, as it require a pipe as a buffer in the middle and two calls to splice, but it would still be a more general solution to get kernel side copying. `splice` was added in 2.6 somewhere so it should easily be old enough.

carlhoerberg:
> I wonder how a `splice` based implementation would compare.

`splice` doesn't take advantage of reflinking/copy-on-write. 

All supported versions of Debian and Ubuntu supports `copy_file_range`. But only FreeBSD 13 and later (and 12 is still supported).

src/io.cr
```diff
@@ -1155,8 +1155,20 @@ abstract class IO
   # io2.to_s # => "hello"
   # ```
   def self.copy(src, dst) : Int64
-    buffer = uninitialized UInt8[DEFAULT_BUFFER_SIZE]
     count = 0_i64
+    {% if LibC.has_method?(:copy_file_range) %}
+      if src.is_a?(File) && dst.is_a?(File)
```
yxhuvud, line 1160: Googling to figure out why this is limited to `File` - the man pages don't say anything about it not working for arbitrary file descriptors, I ran into https://lwn.net/Articles/846403/ describing some of the issues with that. In particular there seems to be (or was, back in 21 when that article was written) issues with virtual file systems and reading from `/proc/*`. So I agree that `File` is the reasonable option. Perhaps there need to be some comment reflecting the motivations?

yxhuvud (commented):
Nice. I initially started to write my comment on the first update that had the `File`s in the argument list instead. This is definitely better than that.

src/io.cr
carlhoerberg: thank you, good comments, i added so that if copy_file_range returns 0 we fallback, like go and rust does.

carlhoerberg: https://github.com/rust-lang/rust/commit/4ddedd521418d67e845ecb43dc02c09b0af53022#diff-4a74019cd4088485a9e0706be6ccf93ff091cb19d9aad474e4e9d03e2bb8522d

PR #8923, opened 2020-03-21
Add String overloads for ljust, rjust and center that take an IO

asterite:
Useful to avoid memory allocations, as always.

In particular we could use it in the new Log module to avoid memory allocations at all when logging stuff: https://github.com/crystal-lang/crystal/pull/8847/files#r395989420

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -1997,6 +1997,13 @@ describe "String" do
     it { "123".ljust(5).should eq("123  ") }
     it { "12".ljust(7, '-').should eq("12-----") }
     it { "12".ljust(7, 'あ').should eq("12あああああ") }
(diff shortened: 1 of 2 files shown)


spec/std/string_spec.cr
```diff
@@ -2578,3 +2599,9 @@ describe "String" do
     end
   end
 end
+
+private def with_io_memory
```
straight-shoota: Why did you not just use `String.build` instead of this helper?

asterite: Oooh... it's a long time since I programmed in Crystal, I forgot this already 😅

RX14: It's my bad, I didn't check the specs well enough!

PR #8924, opened 2020-03-21
Add Int#bit_length

asterite:
Will be useful to implement https://forum.crystal-lang.org/t/integer-sqrt-in-stdlib . It's probably also a generally useful method. Also available in Ruby.

### DIFF
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -785,5 +785,7 @@ describe "Int" do
     0b1001001_i64.bit_length.should eq(7)
     -1.bit_length.should eq(0)
     -10.bit_length.should eq(4)
+
+    BigInt.new("1234567890123456789012345678901234567890").bit_length.should eq(130)
   end
 end
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -783,6 +783,8 @@ describe "Int" do
     0b1.bit_length.should eq(1)
     0b1001.bit_length.should eq(4)
     0b1001001_i64.bit_length.should eq(7)
+    0b1111111111.bit_length.should eq(10)
+    0b1000000000.bit_length.should eq(10)
     -1.bit_length.should eq(0)
     -10.bit_length.should eq(4)
 
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -1,4 +1,4 @@
-require "spec"
+require "./spec_helper"
 {% unless flag?(:win32) %}
   require "big"
 {% end %}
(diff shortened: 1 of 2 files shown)


src/int.cr
```diff
@@ -395,6 +395,36 @@ struct Int
...
   end
 
+  # Returns the number of bits of this int value.
+  #
+  # “The number of bits” means that the bit position of the highest bit
+  # which is different to the sign bit.
+  # (The bit position of the bit 2**n is n+1.)
+  # If there is no such bit (zero or minus one), zero is returned.
+  #
+  # I.e. This method returns `ceil(log2(self < 0 ? -self : self+1))`.
```
vlazar: Example formatted with `crystal tool format`
```suggestion
  # I.e. This method returns `ceil(log2(self < 0 ? -self : self + 1))`.
```

asterite: I copied this verbatim from Ruby's documentation. But I guess this is fine. Thanks!

spec/std/int_spec.cr
```diff
@@ -777,4 +777,11 @@ describe "Int" do
     65.unsafe_chr.should eq('A')
     (0x10ffff + 1).unsafe_chr.ord.should eq(0x10ffff + 1)
   end
+
+  it "#bit_length" do
```
RX14: Test negative numbers? perhaps `BigInt` too but that's not neccesary if it doesn't work (but you should move it to only primitive ints if it doesn't)

spec/std/int_spec.cr
```diff
@@ -777,4 +777,11 @@ describe "Int" do
...
+    0.bit_length.should eq(0)
+    (1..100).each do |i|
+      i.bit_length.should eq(Math.log2(i).to_i + 1)
```
oprypin: This is testing that the result of `bit_length` matches some mathematical formula. But it has about as much to do with the expected answers as the actual implementation of the method. Should you then have to write tests for this mathematical formula?
Tests shouldn't have logic, I'd suggest checking actual inputs.

oprypin: Things like `0b1000` and `0b111.bit_length` might look nice

spec/std/int_spec.cr
asterite: I added one for negative numbers. I also added a fallback for non Int::Primitive types.
Thanks!

spec/std/int_spec.cr
asterite: Good idea. Done!

oprypin: Nice, thanks. But I think it still needs some checks specifically with "1+all-zeros" (like `0b1000`) and "all-ones" (`0b111`) to catch off-by-one errors.

asterite: How about now?

spec/std/int_spec.cr
```diff
@@ -1,4 +1,4 @@
-require "spec"
+require "./spec_helper"
```
straight-shoota: Why do we need `spec_helper`?

asterite: Good question. The code in master doesn't compile if you run just that spec file, it's missing `pending_win32` or something.

Sija: This change shouldn't be needed anymore.

asterite: Why not?

PR #8926, opened 2020-03-22
Socket#sendfile

carlhoerberg:
Request for comments

Benchmark result:
```
    copy 4.0kiB 209.50k (  4.77µs) (± 3.76%)  0.0B/op   1.23× slower
sendfile 4.0kiB 257.89k (  3.88µs) (± 3.72%)  0.0B/op        fastest
    copy 1.0MiB   1.09k (914.69µs) (± 5.40%)  0.0B/op   2.30× slower
sendfile 1.0MiB   2.51k (397.83µs) (± 2.92%)  0.0B/op        fastest
    copy 128MiB   7.94  (125.88ms) (± 5.99%)  0.0B/op   2.16× slower
sendfile 128MiB  17.17  ( 58.24ms) (± 3.59%)  0.0B/op        fastest
```

Benchmark code:
```crystal
require "benchmark"
require "socket"

nc = Process.new "nc", ["-l", "-k", "localhost", "1234"]

{ 4096, 1024**2, 128 * 1024**2 }.each do |size|
  TCPSocket.open("localhost", 1234) do |s|
    File.open("/tmp/a", "w+") do |a|
      a.delete
      a.print "a" * size

      Benchmark.ips do |x|
        x.report("copy #{size.humanize_bytes}") do
          a.rewind
          IO.copy a, s, a.size
        end
        x.report("sendfile #{size.humanize_bytes}") do
          a.rewind
          s.sendfile a, a.size
        end
      end
    end
  end
end
nc.kill
```

### DIFF
--- a/src/io/evented.cr
+++ b/src/io/evented.cr
@@ -88,6 +88,23 @@ module IO::Evented
     end
   end
 
+  def evented_sendfile(remaining, errno_msg : String) : Nil
+    while remaining > 0
+      bytes_written = (yield remaining).to_i32
+      if bytes_written == -1
+        if Errno.value == Errno::EAGAIN
+          wait_writable
+        else
+          raise IO::Error.from_errno(errno_msg)
+        end
+      else
+        remaining -= bytes_written
+      end
+    end
+  ensure
+    resume_pending_writers
+  end
+
   def evented_send(slice : Bytes, errno_msg : String) : Int32
     bytes_written = yield slice
     raise Socket::Error.from_errno(errno_msg) if bytes_written == -1
--- a/src/io/evented.cr
+++ b/src/io/evented.cr
@@ -88,19 +88,26 @@ module IO::Evented
     end
   end
 
-  def evented_sendfile(remaining, errno_msg : String) : Nil
-    while remaining > 0
+  def evented_sendfile(limit : Int, errno_msg : String) : UInt64
+    limit = limit.to_u64
+    remaining = limit
+    loop do
       bytes_written = (yield remaining).to_i32
-      if bytes_written == -1
+      case bytes_written
+      when -1
         if Errno.value == Errno::EAGAIN
           wait_writable
         else
-          raise IO::Error.from_errno(errno_msg)
+          raise Socket::Error.from_errno(errno_msg)
         end
+      when 0
+        break
       else
         remaining -= bytes_written
+        break if remaining.zero?
       end
     end
+    limit - remaining
   ensure
     resume_pending_writers
   end
--- a/src/lib_c/x86_64-linux-gnu/c/unistd.cr
+++ b/src/lib_c/x86_64-linux-gnu/c/unistd.cr
@@ -43,4 +43,5 @@ lib LibC
   fun sysconf(name : Int) : Long
   fun unlink(name : Char*) : Int
   fun write(fd : Int, buf : Void*, n : SizeT) : SSizeT
+  fun sendfile(fd_out : Int, fd_in : Int, offset : OffT*, count : SizeT) : SSizeT
 end
--- a/src/io.cr
+++ b/src/io.cr
@@ -1147,8 +1147,7 @@ abstract class IO
 
     if socket = dst.as?(Socket)
       if file = src.as?(IO::FileDescriptor)
-        socket.sendfile file, limit
-        return limit
+        return socket.sendfile file, limit
       end
     end
 
--- a/src/io.cr
+++ b/src/io.cr
@@ -1115,6 +1115,12 @@ abstract class IO
   # io2.to_s # => "hello"
   # ```
   def self.copy(src, dst) : UInt64
+    if socket = dst.as?(Socket)
+      if file = src.as?(File)
+        return socket.sendfile file, file.size - file.pos
+      end
+    end
+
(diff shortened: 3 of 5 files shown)


carlhoerberg:
Have to "invalidate" the read buffer of the source, just as in #8919

src/io/evented.cr
```diff
@@ -88,6 +88,23 @@ module IO::Evented
     end
   end
 
+  def evented_sendfile(remaining, errno_msg : String) : Nil
```
RX14: I'm not sure about this method, for example `recvfrom` uses `evented_read` too. I think this is just telling us that `evented_write` is too specific and needs to be abstracted more to handle the `sendfile` case (i.e. move the slice handling into the callers).

I'm happy to do this refactor myself, though. For example, merge this then merge a refactor after.

src/lib_c/x86_64-linux-gnu/c/unistd.cr
```diff
@@ -43,4 +43,5 @@ lib LibC
   fun sysconf(name : Int) : Long
   fun unlink(name : Char*) : Int
   fun write(fd : Int, buf : Void*, n : SizeT) : SSizeT
+  fun sendfile(fd_out : Int, fd_in : Int, offset : OffT*, count : SizeT) : SSizeT
```
RX14: Obviously more bindings need to be added and platforms which dont support it handled.

RX14 (commented):
Please allow this discussion to continue *before* making any changes to the code.  Otherwise you'll burn out making changes before a consensus is reached.

src/io/evented.cr
```diff
@@ -88,6 +88,23 @@ module IO::Evented
...
+    while remaining > 0
+      bytes_written = (yield remaining).to_i32
+      if bytes_written == -1
+        if Errno.value == Errno::EAGAIN
+          wait_writable
+        else
+          raise IO::Error.from_errno(errno_msg)
```
bew: Should be a `Socket::Error` here I think

src/lib_c/x86_64-linux-gnu/c/unistd.cr
carlhoerberg: of course, was just a start. BSD/Darwin support sendfile too but with two extra arguments (prepend/append data). not sure about windows but i assume there should something similar..

src/io/evented.cr
carlhoerberg: ok! sounds good

src/io/evented.cr
carlhoerberg: Ok, but would be nice to see what errno it is..

RX14: `Socket::Error` contains `errno` too....

RX14:
> Please allow this discussion to continue _before_ making any changes to the code

src/lib_c/x86_64-linux-gnu/c/unistd.cr
bcardiff: Based on https://tinyclouds.org/iocp-links.html
* [Windows](https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nf-mswsock-transmitfile?redirectedfrom=MSDN)
* [Linux](http://man7.org/linux/man-pages/man2/sendfile.2.html)
* [FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=sendfile&sektion=2)
* [Darwin](http://www.manpagez.com/man/2/sendfile/)

How about adding a `Crystal::System.sendfile?(src : IO, dst : IO, limit : Int) : UInt64?` that returns `nil` if there was not a way to send the data from one `IO` to another in a specialized way; otherwise the number of bytes sent. 

The type checks would be done inside that definition and using it could be used like

```cr
len = Crystal::System.sendfile?(src, dst, limit)
return len if len
```

carlhoerberg:
libevent implements sendfile in its event buffer thingy. But I guess there must been some reason to why Crystal has it's own buffering layer instead of using libevent's? 

http://www.wangafu.net/~nickm/libevent-2.1/doxygen/html/buffer_8h.html#a601996b1fc1f5c165dc62b89acbd069e

asterite:
I think those functions were unknown to the implementors.

ysbaddaden:
I think IO::Buffered was implemented _before_ we started using libevent.

RX14:
I'd prefer not to bind crystal's IO design closer to libevent than it already it, personally.

src/lib_c/x86_64-linux-gnu/c/unistd.cr
niacat: *FreeBSD* supports sendfile. Not BSD in general.

waj:
How is the impact of using `sendfile` with a large file? Does the system gets into a long syscall? I guess when the fd is set as nonblocking it should return as soon as the output buffer is full, but is worth to try. Just because the test benchmark gives better numbers doesn't necessarily mean is good for a system running with many concurrent users. We need to check how it affects latency in those cases.

In any case, I think I'd prefer to have the `Socket#sendfile` but not make the automatic switch in `IO#copy`. If `sendfile` only works with sockets (and files?) then it makes sense to be a method in  the `Socket` class. And let the one writing the code to decide if the syscall is preferred or not. At least as a non-breaking change starting point.

Issue #8921, opened 2020-03-20
option parser quit on help


smalls12:
Hello,

The OptionParser in the documents is as follows

```
require "option_parser"

upcase = false
destination = "World"

OptionParser.parse do |parser|
  parser.banner = "Usage: salute [arguments]"
  parser.on("-u", "--upcase", "Upcases the salute") { upcase = true }
  parser.on("-t NAME", "--to=NAME", "Specifies the name to salute") { |name| destination = name }
  parser.on("-h", "--help", "Show this help") { puts parser }
  parser.invalid_option do |flag|
    STDERR.puts "ERROR: #{flag} is not a valid option."
    STDERR.puts parser
    exit(1)
  end
end

destination = destination.upcase if upcase
puts "Hello #{destination}!"
```

I expected that if you actually used the "-h" or "--help" that the program would display the help and then would end.

However in this example above, the help menu would be printed and the program will continue.

If the program writes a lot of data to the screen, the help menu could be lost until the program ends and then the user scrolls up to see the help.

The solution might be to follow what `parser.invalid_option` does; something like `parser.help`.

Could this behavior be added?

Sija:
It's as easy as doing `puts parser; exit`...

smalls12:
Oh fair enough, I guess that works.

RX14:
I think that printing the help message being just `OptionParser#to_s` is a bit confusing.

straight-shoota:
The example should have an `exit` call to demonstrate proper usage.

vlazar:
@RX14 Are you talking about renaming `OptionParser#to_s` to something like `OptionParser#help` or `OptionParser#options` instead? I can't think of better method name.

vlazar:
On the second though maybe `parser.print_help` replacing `puts parser.to_s` would do?

smalls12:
Maybe related, does the user have to specify the help option? Could it not just be added in the background automatically?

PR #8927, opened 2020-03-22
Make OptionParser example to exit on --help

vlazar:
Ref #8921


src/option_parser.cr
```diff
@@ -18,7 +18,10 @@
 #   parser.banner = "Usage: salute [arguments]"
 #   parser.on("-u", "--upcase", "Upcases the salute") { upcase = true }
 #   parser.on("-t NAME", "--to=NAME", "Specifies the name to salute") { |name| destination = name }
-#   parser.on("-h", "--help", "Show this help") { puts parser }
+#   parser.on("-h", "--help", "Show this help") do
+#     puts parser
+#     exit(0)
```
jhass: `0` is the default value, no need to give it explicitly.

vlazar: Thanks! Updated.

PR #8931, opened 2020-03-23
Correctly implement bit_length for BigIng

asterite:
Follow up to #8924

Some bugs were detected, so this fixes that.

Also:
- Made `bit_length` always return `Int32` which makes more sense
- Optimized for `BigInt`: I found the specific function to this in libgmp

Here's a benchmark:

```
require "big"
require "benchmark"

i = 10.to_big_i ** 309

Benchmark.ips do |x|
  x.report("bit_length") do
    i.bit_length
  end
  x.report("to_s(2).size") do
    i.to_s(2).size
  end
end
```

Output:

```
  bit_length 113.16M (  8.84ns) (± 3.95%)    0.0B/op         fastest
to_s(2).size 333.97k (  2.99µs) (± 5.92%)  2.63kB/op  338.85× slower
```

Also a benchmark for Ruby:

```ruby
require "benchmark/ips"

i = 10 ** 309

Benchmark.ips do |x|
  x.report("bit_length") do
    i.bit_length
  end
end
```

Output:

```
Warming up --------------------------------------
          bit_length   424.827k i/100ms
Calculating -------------------------------------
          bit_length     17.227M (± 3.5%) i/s -     86.240M in   5.012266s
```

About 10x faster than Ruby :-)

### DIFF
--- a/spec/std/int_spec.cr
+++ b/spec/std/int_spec.cr
@@ -791,7 +791,8 @@ describe "Int" do
     end
 
     pending_win32 "for BigInt" do
-      BigInt.new("1234567890123456789012345678901234567890").bit_length.should eq(130)
+      (10.to_big_i ** 20).bit_length.should eq(67)
+      (10.to_big_i ** 309).bit_length.should eq(1027)
     end
   end
 end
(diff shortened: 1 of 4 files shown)


vlazar:
Should this also be milestoned to 0.34.0 right away?

PR #8935, opened 2020-03-24
Add `IO::Buffered#flush_on_newline` back and set it to true for non-tty

### LINKED ISSUE #7431: Should STDOUT always set flush_on_newline by default?

asterite:
Reverts #7470
Fixes #7431 (somehow)

Before #7470 if you output to the console and redirected that output to somewhere else, for example a file, then you would see output right after each line was emitted.

Once we merged #7470 you would only see output after the IO buffer was full and it was flushed.

That meant that if you did `tail -f` on that file you wouldn't see new lines appended to it immediately. I believe this is not very intuitive. If you don't output enough data you might think something is broken and lose a lot of time debugging this.

This PR reverts the behavior back to how it was before.

This is **slower**, but you still have the option to make is faster (if you don't care about not seeing "correct" output using `tail -f`) by doing:

```crystal
STDOUT.flush_on_newline = false
```

If we don't add this method back, the only options you have are:
- fully buffer the output: fast, but not very intuitive
- no buffer: **very** slow, but intuitive

With `flush_on_newline` it's not fast but it's also not **very** slow. So I think this is a good default.

💭 We could also consider having an environment variable, for example `CRYSTAL_STDOUT_FLUSH_ON_NEWLINE` so you can control, at runtime, how the program behaves (on startup). That means that you can compile your program once and control how it behaves later on. But I didn't include this here. We can discuss it in a different issue if needed, and do it in a separate PR.

Please 👍 if you agree with this change, 👎 if not.

### DIFF
--- a/src/io/file_descriptor.cr
+++ b/src/io/file_descriptor.cr
@@ -32,22 +32,27 @@ class IO::FileDescriptor < IO
 
   # :nodoc:
   def self.from_stdio(fd)
-    # If we have a TTY for stdin/out/err, it is possibly a shared terminal.
-    # We need to reopen it to use O_NONBLOCK without causing other programs to break
+    {% if flag?(:win32) %}
+      new(fd)
+    {% else %}
+      # If we have a TTY for stdin/out/err, it is possibly a shared terminal.
+      # We need to reopen it to use O_NONBLOCK without causing other programs to break
 
-    # Figure out the terminal TTY name. If ttyname fails we have a non-tty, or something strange.
-    path = uninitialized UInt8[256]
-    ret = LibC.ttyname_r(fd, path, 256)
-    return new(fd).tap(&.flush_on_newline=(true)) unless ret == 0
+      # Figure out the terminal TTY name. If ttyname fails we have a non-tty, or something strange.
+      # For non-tty we set flush_on_newline to true for reasons described in STDOUT and STDERR docs.
+      path = uninitialized UInt8[256]
+      ret = LibC.ttyname_r(fd, path, 256)
+      return new(fd).tap(&.flush_on_newline=(true)) unless ret == 0
 
-    clone_fd = LibC.open(path, LibC::O_RDWR)
-    return new(fd).tap(&.flush_on_newline=(true)) if clone_fd == -1
+      clone_fd = LibC.open(path, LibC::O_RDWR)
+      return new(fd).tap(&.flush_on_newline=(true)) if clone_fd == -1
 
-    # We don't buffer output for TTY devices to see their output right away
-    io = new(clone_fd)
-    io.close_on_exec = true
-    io.sync = true
-    io
+      # We don't buffer output for TTY devices to see their output right away
+      io = new(clone_fd)
+      io.close_on_exec = true
+      io.sync = true
+      io
+    {% end %}
   end
 
   def blocking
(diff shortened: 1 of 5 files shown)


src/io/file_descriptor.cr
```diff
@@ -38,10 +38,10 @@ class IO::FileDescriptor < IO
     # Figure out the terminal TTY name. If ttyname fails we have a non-tty, or something strange.
     path = uninitialized UInt8[256]
     ret = LibC.ttyname_r(fd, path, 256)
-    return new(fd) unless ret == 0
+    return new(fd).tap(&.flush_on_newline=(true)) unless ret == 0
 
     clone_fd = LibC.open(path, LibC::O_RDWR)
-    return new(fd) if clone_fd == -1
+    return new(fd).tap(&.flush_on_newline=(true)) if clone_fd == -1
```
yxhuvud: Do we want a comment explaining why the flush is happening? It was obviously not obvious enough the first time :)

asterite: Good idea! I added a couple more commits explaining this. Thank you!

src/io/file_descriptor.cr
```diff
@@ -32,22 +32,27 @@ class IO::FileDescriptor < IO
...
+      new(fd)
+    {% else %}
+      # If we have a TTY for stdin/out/err, it is possibly a shared terminal.
+      # We need to reopen it to use O_NONBLOCK without causing other programs to break
+
+      # Figure out the terminal TTY name. If ttyname fails we have a non-tty, or something strange.
+      # For non-tty we set flush_on_newline to true for reasons described in STDOUT and STDERR docs.
+      path = uninitialized UInt8[256]
+      ret = LibC.ttyname_r(fd, path, 256)
+      return new(fd).tap(&.flush_on_newline=(true)) unless ret == 0
```
Sija: Wouldn't that set `flush_on_newline = true` for `STDIN` as well?

asterite: Yes, but it doesn't matter because you don't write to STDIN.

Sija: Just wanted to be sure since it's a bit inconsistent.

asterite: Yeah, makes sense. I thought about that too. But in the end, this is just setting a boolean for a scenario that will never be used, and I guess it's fine.

asterite:
Let's go forward with this. It's been like this for years, and we only had one complaint and then I changed it. And there's always a way to go back to the old behavior (by changing your code).

PR #8940, opened 2020-03-25
LLVM 10 support

RX14:
Nice and easy upgrade for a change :)


bcardiff:
Is that it? 

I was just checking and brew does not have llvm yet. Someday we could have CI for different llvm versions...

RX14:
That's it! I compiled LLVM 10 and tested it working.

wontruefree:
I am not sure if these are used but you might want to add it here too.
https://github.com/crystal-lang/crystal/blob/master/src/llvm/lib_llvm.cr#L21

RX14:
It's not used so there's no need to add it.

PR #8943, opened 2020-03-26
Allow include and extend inside enum declarations

### LINKED ISSUE #8946: Add include and extend to Enum
I would like the posibility to use include and extend on Enum.

Now they are added to Enum.class but not to the actual Enum.

Say we do:
```
enum Status
  Opened
  Closed
end
```
And extend or include in an enum it will be only in Enum. And using include and extend won't work inside `enum Status`

Why I want this?
Most ORM adds module or methods to parse, get data from database and to database to a type, so-called type extensions. Enum is handling include and extend differently making this not work. And can therefore not use Enums as a type for columns without some type of wrapper class around the enum to cast it correctly and so on.

asterite:
Fixes #8946

For some reason this wasn't allowed, but it's very easy to implement and it makes sense.

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -1396,6 +1396,9 @@ module Crystal
 
     it_parses "enum Foo; @[Bar]; end", EnumDef.new("Foo".path, [Annotation.new("Bar".path)] of ASTNode)
 
+    it_parses "enum Foo; include Bar; end", EnumDef.new("Foo".path, [Include.new("Bar".path)] of ASTNode)
+    it_parses "enum Foo; extend Bar; end", EnumDef.new("Foo".path, [Extend.new("Bar".path)] of ASTNode)
+
     assert_syntax_error "enum Foo; A B; end", "expecting ';', 'end' or newline after enum member"
     assert_syntax_error "enum Foo\n  A,   B,   C\nend\n", "expecting ';', 'end' or newline after enum member"
 
--- a/spec/compiler/semantic/enum_spec.cr
+++ b/spec/compiler/semantic/enum_spec.cr
@@ -492,4 +492,40 @@ describe "Semantic: enum" do
       end
     ))
   end
+
+  it "can include module" do
+    assert_type(%(
+      module Moo
+        def moo
+          1
+        end
+      end
+
+      enum Color
+        Red
+
+        include Moo
+      end
+
+      Color::Red.moo
+      )) { int32 }
+  end
+
+  it "can extend module" do
+    assert_type(%(
+      module Moo
+        def moo
+          1
+        end
+      end
+
+      enum Color
+        Red
+
+        extend Moo
+      end
+
+      Color.moo
+      )) { int32 }
+  end
 end
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -5715,6 +5715,10 @@ module Crystal
             member = parse_macro.at(def_location)
             member = VisibilityModifier.new(visibility, member) if visibility
             members << member
+          when :include
+            members << parse_include
+          when :extend
+            members << parse_extend
           else
             unexpected_token
           end


confact:
Thanks for the fast solving. Feel free to close my issue #8946 when this is merged.

confact:
Will this accept including/extending a module in the module you including/extending? 

As an example, you can't have a module inside an enum directly. But you can add modules to other types.

asterite:
>Will this accept including/extending a module in the module you including/extending?

Sorry, I don't understand what this means? Can you provide an example?

> As an example, you can't have a module inside an enum directly. But you can add modules to other types.

No, enums can't have types inside them and they will never have. The reason is that the only types inside enums are the enum members. If we allow other types then you can't rely on the constants inside an enum to be all members. It would also be pretty confusing:

```crystal
enum Color
  Red
  Green
  Blue

  module Foo
  end
end

Color::Red # okay...
Color::Foo # weird, looks like another color, but it isn't
```

confact:
Okay, you answered the first one on the second part without knowing it.
As an example, Lucky want to call methods in their Avram on Type like this: Int32::Lucky.parse and in this way this won't work on enum. My idea was to have a module with module Lucky inside it, like:

```
module ExtraEnumThings
  module Lucky
  end
end
```
and add that to enum.

I understand the idea of not having Type inside enum, like a module, as constants should be there. But I also would like to add a module to expand the Enum to handle more stuff. 

It does exist a way to find the constants and know what is a module and not. But I am okay with include & extend. We just need to remove the module on all types :)

asterite:
Oh, yeah, that will work. I just tried this:

```crystal
module Moo
  module Foo
  end
end

enum Color
  Red
  Green
  Blue

  include Moo
end

p Color::Foo # => Moo::Foo
```

That said, I have no idea how will `Moo::Foo` know it's dealing with Color.

What does `Int32::Lucky.parse` do? Do every type in lucky gets a `Lucky` added inside it?

asterite:
Could you show an example usage or code of what you want to achieve?

You shared this in gitter: https://gist.github.com/confact/27d62ee6ca27a47182bd1a665f9b6a3f

But there's no usage of that code so I don't know how that's supposed to work, or why adding a module inside another type is needed at all.

asterite:
Or do you need this functionality for all enums? If that's the case, you can `include` and `extend` inside `struct Enum`, so that would solve your use case.

My guess is that Lucky has a convention for types that go inside DB columns, to be able to store them and retrieve them. It seems to revolve around an `adapter` method, ` Lucky` module and so on. But I have no idea.

Can you provide the code that Lucky ends up generating for a type, and how would you make that manually work for one specific enum? From there, we can start generalizing the code using macros (or not).

confact:
@asterite I am trying to add extensions to Enum to make it possible to cast it from the database.
The code in the gist is similar to the other type extension in Avram, like int32: https://github.com/luckyframework/avram/blob/master/src/avram/charms/int32_extensions.cr
It is then triggered from `SaveOperation` to get type and then call parse on the class here: https://github.com/luckyframework/avram/blob/master/src/avram/save_operation.cr#L131

asterite:
Yeah... that won't work for enums because you can't define `Lucky` inside each enum type. It seems Lucky considers there's `Int32::Lucky`, `String::Lucky`, etc. But here, assuming you have `enum Color`, `enum Kind`, etc., you want `Color::Lucky` and `Kind::Lucky` to be different modules, and that's not possible.

I think Lucky will need special code for this to support enums.

Another way would be to flip things. Instead of adding `Lucky` to every type, why not define types inside a `Lucky` module? Then you would have:

```crystal
module Lucky
  module Extensions
    module Int32 # extensions for Int32
    end
 
    module String # extensions for String
    end

    module Color # extensions for Color
    end
  end
end
```

and so on. Just an idea. But for now you probably need to modify Lucky to handle enums in a different way.

confact:
@asterite okay, thanks for checking and answering. Include and Extend will be nice in Enum whatever the outcome from Lucky is :)

Keep up the good job!

paulcsmith:
Thanks for the great discussion here! I'm working on a new way to handle parsing for different types.  @asterite you are correct that these methods are used when parsing params before saving, and for converting values when inserting/reading. I may try the approach of `Lucky::Types::Int32`, but am not sure. I wrote this quite a long time ago and I remember there being lots of edge cases that made things impossible, but with more experience I think we can come up with something better.

Either way, thanks for adding this ability to enums!

RX14:
> Oh, yeah, that will work. I just tried this:

That makes me dubious about accepting this feature. Seems this limitation exists for a reason.

asterite:
I think it's fine because calling `constants` on an enum type will still return only the constants defined explicitly in the enum, not in parent types.

PR #8948, opened 2020-03-26
Compiler: Honor LIBRARY_PATH as default library path

bcardiff:
* If not set, default to /usr/lib, /usr/local/lib.
* If set to empty it allows you to skip adding /usr/lib and /usr/local/lib to the linker.
* Setting to a `list:of:directories`, those directories will be used for static libs lookup and appended as `-L` in the linker.

Closes #8326

We will fail to lookup some libraries if we don't keep the default, probably because:

* OSX does not set LIBRARY_PATH by default
* FreeBSD won't search /usr/local/lib by default

I check that GCC seems to add directory with some heuristics https://github.com/gcc-mirror/gcc/blob/41d6b10e96a1de98e90a7c0378437c3255814b16/config/lib-link.m4#L389-L396

Note: I accidentally pushed it to master as e1b4e8fd5e5b6ddc1e5426803b559ad4992dd133, but I rolled back ef5f10f6ead9004ae65db5f2fb637bb718659abd. (I will learn to use git worktrees better next time ¬¬, sorry)

### DIFF
--- a/src/compiler/crystal/codegen/link.cr
+++ b/src/compiler/crystal/codegen/link.cr
@@ -105,7 +105,8 @@ module Crystal
     end
 
     private def lib_flags_posix
-      library_path = ["/usr/lib", "/usr/local/lib"]
+      library_path = ENV["LIBRARY_PATH"]?.try(&.split(':', remove_empty: true)) ||
+                     ["/usr/lib", "/usr/local/lib"]
       has_pkg_config = nil
 
       String.build do |flags|


oprypin:
Have you found any situation where this makes a difference? `LIBRARY_PATH` [works just fine already](https://github.com/oprypin/crsfml/runs/526083111?check_suite_focus=true#step:11:1), as `lld` inherits and picks up the variable itself. So this code basically duplicates the addition of paths.

I think all the code involving `library_path` is unnecessary on Linux, even the defaults of `["/usr/lib", "/usr/local/lib"]`. But guessing from the comments that BSD is a different story.

Blacksmoke16:
> Note: I accidentally pushed it to master as [e1b4e8f](https://github.com/crystal-lang/crystal/commit/e1b4e8fd5e5b6ddc1e5426803b559ad4992dd133), but I rolled back [ef5f10f](https://github.com/crystal-lang/crystal/commit/ef5f10f6ead9004ae65db5f2fb637bb718659abd). (I will learn to use git worktrees better next time ¬¬, sorry)

You could setup master as a protected branch to make it so it can't be pushed to directly.

References:
* https://github.community/t5/Support-Protips/Best-practices-for-protected-branches/ba-p/10224
* https://help.github.com/en/github/administering-a-repository/about-protected-branches

bcardiff:
> You could setup master as a protected branch

Thanks @Blacksmoke16, but even with that: Organization administrators, repository administrators, and users with the Maintain role. These members can always push. 😅

oprypin:
https://github.com/crystal-lang/crystal/pull/8326#issuecomment-604113511 - same assessment as mine.

bcardiff:
@oprypin skipping the default value for the library_path changes the lookup for static libraries and will require changes in the setup. I'm trying to keep the current behaviour as much as possible. That is until we can rethink the linking stories with some other features.

My BSD comment is based on a comment on the code only.

Blacksmoke16:
@bcardiff There is a setting to prevent that.

image

bcardiff:
Thank @Blacksmoke16 (and @spalladino) for pointing me that at the same time on different channels :-)

src/compiler/crystal/codegen/link.cr
```diff
@@ -105,7 +105,8 @@ module Crystal
     end
 
     private def lib_flags_posix
-      library_path = ["/usr/lib", "/usr/local/lib"]
+      library_path = ENV["LIBRARY_PATH"]?.try(&.split(':', remove_empty: true)) ||
+                     ["/usr/lib", "/usr/local/lib"]
```
oprypin: Equivalent change, to make the absurdity explicit.

```suggestion
      library_path = ENV.has_key?("LIBRARY_PATH") ? [] of String :
                     ["/usr/lib", "/usr/local/lib"]
```

oprypin: To be fair, there's also nothing wrong with what this code accomplished, it just has a potential to be misunderstood / cargo-culted.

bcardiff: But I do want to look up the static libraries in the LIBRARY_PATH directories.

oprypin: That is already the case

asterite: I think what oprypin means is that we don't have to search them ourselves because the linker will do that. So maybe all of this is dead code? I'm not sure.

asterite: That said, LIBRARY_CRYSTAL is empty for me and without this it will stop working.

oprypin:
```suggestion
      # Note: most linkers (Linux, macOS) already look at `LIBRARY_PATH`
      # themselves (and also have these defaults), so this code would only
      # duplicate that work, but the duplication is harmless.
      library_path = ENV["LIBRARY_PATH"]?.try(&.split(':', remove_empty: true)) ||
                     ["/usr/lib", "/usr/local/lib"]
```

bcardiff: I still hold that the behavior changes. Without library_path being set `-lfoo` will be passed to the linker, instead of `/path/to/libfoo.a`. This comes from the `find_static_lib` invocation.

bcardiff:
Thanks for the approval. I now this can be better in the future, and I'm looking forward to it.

PR #8949, opened 2020-03-26
Allow config of IO buffer size

bcardiff:
follow up from https://forum.crystal-lang.org/t/io-copy-buffer-too-small/1873

I try to make it from env vars instead of macros, but the math interpreter does not allow env vars to be read. And going all the way with macro run seems too much for this.

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -59,7 +59,22 @@ require "c/errno"
 # avoided, as string operations might need to read extra bytes in order to get characters
 # in the given encoding.
 abstract class IO
-  BUFFER_SIZE = 8192
+  # Changes the default buffer size that is used in `IO.copy`, `IO::Buffered`
+  # and othe operations. The first invocation defines the value.
+  # Defaults to 8192 (8kb)
+  #
+  # ```
+  # IO.set_buffer_size 16384
+  # ```
+  macro set_buffer_size(value)
+    {% if !IO.has_constant?(:BUFFER_SIZE) %}
+      ::IO::BUFFER_SIZE = {{value}}
+    {% end %}
+  end
+
+  macro finished
+    set_buffer_size 8192
+  end
 
   # Argument to a `seek` operation.
   enum Seek
--- a/src/io.cr
+++ b/src/io.cr
@@ -59,6 +59,8 @@ require "c/errno"
 # avoided, as string operations might need to read extra bytes in order to get characters
 # in the given encoding.
 abstract class IO
+  BUFFER_SIZE = 8192
+
   # Argument to a `seek` operation.
   enum Seek
     # Seeks to an absolute location
@@ -541,7 +543,7 @@ abstract class IO
           decoder.write(str)
         end
       else
-        buffer = uninitialized UInt8[4096]
+        buffer = uninitialized UInt8[BUFFER_SIZE]
         while (read_bytes = read(buffer.to_slice)) > 0
           str.write buffer.to_slice[0, read_bytes]
         end
@@ -815,9 +817,9 @@ abstract class IO
   # io.skip(1) # raises IO::EOFError
   # ```
   def skip(bytes_count : Int) : Nil
-    buffer = uninitialized UInt8[4096]
+    buffer = uninitialized UInt8[BUFFER_SIZE]
     while bytes_count > 0
-      read_count = read(buffer.to_slice[0, Math.min(bytes_count, 4096)])
+      read_count = read(buffer.to_slice[0, Math.min(bytes_count, BUFFER_SIZE)])
       raise IO::EOFError.new if read_count == 0
 
       bytes_count -= read_count
@@ -827,7 +829,7 @@ abstract class IO
   # Reads and discards bytes from `self` until there
   # are no more bytes.
   def skip_to_end : Nil
-    buffer = uninitialized UInt8[4096]
+    buffer = uninitialized UInt8[BUFFER_SIZE]
     while read(buffer.to_slice) > 0
     end
   end
@@ -1115,7 +1117,7 @@ abstract class IO
   # io2.to_s # => "hello"
   # ```
   def self.copy(src, dst) : UInt64
-    buffer = uninitialized UInt8[4096]
+    buffer = uninitialized UInt8[BUFFER_SIZE]
     count = 0_u64
     while (len = src.read(buffer.to_slice).to_i32) > 0
       dst.write buffer.to_slice[0, len]
@@ -1139,7 +1141,7 @@ abstract class IO
 
     limit = limit.to_u64
 
-    buffer = uninitialized UInt8[4096]
+    buffer = uninitialized UInt8[BUFFER_SIZE]
     remaining = limit
     while (len = src.read(buffer.to_slice[0, Math.min(buffer.size, Math.max(remaining, 0))])) > 0
       dst.write buffer.to_slice[0, len]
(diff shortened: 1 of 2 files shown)


src/io.cr
```diff
@@ -59,6 +59,23 @@ require "c/errno"
...
+  # IO.set_buffer_size 16384
+  # ```
+  macro set_buffer_size(value)
+    {% if !IO.has_constant?(:BUFFER_SIZE) %}
+      ::IO::BUFFER_SIZE = {{value}}
+    {% end %}
+  end
+
+  macro finished
+    set_buffer_size 8192
```
wontruefree: Should this be set to a constant?

bcardiff: It is the default constant definition

Sija:
> I try to make it from env vars instead of macros, but the math interpreter does not allow env vars to be read. And going all the way with macro run seems too much for this.

@bcardiff `BUFFER_SIZE = {{ env("FOO") || 4096 }}` should work, no?

bcardiff:
@Sija You will get an "invalid constant value"

Sija:
oh, right, you'd have to do sth like `{{ env("FOO").id }} || 4096` but that would be brittle if work at all...

src/io.cr
```diff
@@ -59,6 +59,23 @@ require "c/errno"
 # avoided, as string operations might need to read extra bytes in order to get characters
 # in the given encoding.
 abstract class IO
+  # Changes the default buffer size that is used in `IO.copy`, `IO::Buffered`
+  # and othe operations. The first invocation defines the value.
+  # Defaults to 8192 (8kb)
```
Sija:
```suggestion
  # Changes the default buffer size that is used in `IO.copy`, `IO::Buffered`
  # and other operations. The first invocation defines the value.
  # Defaults to `8192` (8kb)
```

asterite:
I don't think all buffer sizes should be changed. The original discussion was just about IO.copy

bcardiff:
@asterite changing the two `IO.copy` and `IO::Buffered` seems good enough? A `IO::Buffered` instance change its buffer size, but I think the same default should be used for `IO.copy` as well at least.

RX14:
Unless we have a real usecase for making these stack-sized buffers configurable size, then I don't want to introduce an ugly macro hack to make that so.

Setting the buffer size for `IO::Buffered` is a different story, since it doesn't need a macro hack.

One page or two for `IO.copy` buffers is moot, I don't mind the change.

asterite:
Yeah, I think for now. I wouldn't do anything. At most I would increase the bufer size ofor copy to 128K (or a smaller value) , which only affects stack size.

j8r:
Why it is a constant instead of a class var? It could be inherited with a default value, and overwritten.

carlhoerberg:
Configuring it via env vars at compile time just seems brittle/messy. Increasing it to 8KB (to align it with the default IO::Buffer size) seems like a no-brainer though. Increasing it to 128KB seems optimal but then I don't know what the downsides are.

One thing that really improved responsiveness in our application with multiple active high-throughput clients was a `Fiber.yield` in the while loop when using a 128KB buffer. 

Imagine this simple http download server:
```crystal
require "http/server"

class IO
  def self.copy_fast(src, dst, limit : Int) : UInt64
    raise ArgumentError.new("Negative limit") if limit < 0

    limit = limit.to_u64

    buffer = uninitialized UInt8[131_072]
    remaining = limit
    while (len = src.read(buffer.to_slice[0, Math.min(buffer.size, Math.max(remaining, 0))])) > 0
      dst.write buffer.to_slice[0, len]
      remaining -= len
      break if remaining.zero?
      Fiber.yield
    end
    limit - remaining
  end
end

zero = File.open("/dev/zero")
  size = 10_u64*1024**3 # 10 GB
  context.response.content_length = size
  case context.request.path
  when "/fast"
    IO.copy_fast zero, context.response, size
  else
    IO.copy zero, context.response, size
  end
end

address = server.bind_tcp "127.0.0.1", 8080
server.listen
```

Download from it with:
```
curl -o /dev/null localhost:8080
```
Speed ~1,4GBit/s

Start two downloads at the same time and notice how the other download won't start until the first has finished. 

Now try 
```
curl -o /dev/null localhost:8080/fast
```
2.4Gbit/s

And start multiple simultaneous downloads and see how they all can download at the same time, and the bandwidth is shared between all the clients.

asterite:
My suggestion: increase it to 8KB fixed. The code to copy one IO to another is just 5 lines of code. If one needs a bigger buffer, one can implement it in app code or a shard.

There's nothing to do here. And this is not necessary for 1.0

carlhoerberg:
Or could making `IO.copy` a macro with a fourth argument *buffer_size* be a solution? 

```crystal
class IO
  macro copy(src, dst, limit, buffer_size = 8192)
    begin
      raise ArgumentError.new("Negative limit") if {{limit}} < 0

      limit = {{limit}}.to_u64

      buffer = uninitialized UInt8[{{buffer_size}}]
      remaining = limit
      while (len = {{src}}.read(buffer.to_slice[0, Math.min(buffer.size, Math.max(remaining, 0))])) > 0
        {{dst}}.write buffer.to_slice[0, len]
        remaining -= len
        break if remaining.zero?
      end
      limit - remaining
    end
  end
end
```

Or an additonal `IO.large_copy` method with a 128kb buffer. 3x difference (between 8KB and 128KB) is quite significant and it would be sad for ppl to loose out on such simple boost. 

Or an if statement in the `IO.copy` method were we use a small buffer for small copies (less than 128kb?) and larger buffer for larger copies?

If "doing nothing" then it should probably be documented that if high throughput is desired then the method should be overridden. Again, not sure what the downsizes are..

RX14:
> Increasing it to 128KB seems optimal but then I don't know what the downsides are.

On small copies, you'd be faulting in a bunch of pages to be used only once.

By default, each fiber has 8KiB cost, for 4GiB of ram that means 524,288 fibers. If each fiber was doing copies like this, it'd cost 128+8KiB per fiber, or only ~31,000 fibers.

I agree with ary, I think this is fine as-is, and we should increase the `IO.copy` size to 2 pages, not one. I'd be fine merging making `IO::Buffered`'s size configurable, but none of this requires macro magic.

bcardiff:
I wonder @carlhoerberg, if #8926 evolves, doesn't that supersedes this story of changing buffers?

carlhoerberg:
not really, `sendfile` requires a flush of buffers, so it's only beneficial for rather large writes.

carlhoerberg:
All ruby's buffers are 16KB btw, and so are the default buffers for nginx and haproxy too

waj:
16KB sounds good to me, and if Ruby and other well known apps also uses that value maybe someone did a research and found that as optimal value (or they just copy each other 😆). Also I just did a quick test on my machine and seems to be an inflection point at 16KB.

@bcardiff the declaration of the constant could be replaced to this, so it can be changed with an environment variable:
```crystal
{% begin %}
BUFFER_SIZE = {{ (env("CRYSTAL_IO_BUFFER_SIZE") || "16384").to_i }}
{% end %}
```

I wouldn't mind not adding that anyway, but just changing the default. It's a very simple change that apparently has a big performance impact.

asterite:
I would prefer just changing the default value. I don't think there's a need to globally change the buffer size, specially if 16KB seems to be an optimal value. Also, the buffer size can be changed on individual IOs if needed.

carlhoerberg:
I think this can be closed now that most/all default io buffers are 16KB.

PR #8952, opened 2020-03-27
Rename Flate to Deflate

bcardiff:
Closes #8886.

From #8886 the renaming of Flate to Deflate is something we all agree on. Since there is no `Compress` namespace, there is no need to create it only for `Deflate`.


spec/std/deflate/deflate_spec.cr
```diff
@@ -0,0 +1,106 @@
...
+          writer.write "Hello Crystal!".to_slice
+        end
+
+        io.rewind
+        io.to_slice.should eq(Bytes[243, 72, 205, 201, 201, 87, 112, 46, 170, 44, 46, 73,
+          204, 81, 4, 0])
+      end
+    end
+  end
+end
```
j8r: This are new specs? It is on the `Rename Flate to Deflate` commit.

bcardiff: No, they are the old ones copied. The old ones are still there to stress the deprecated definitions. But on this case the file could be moved also since the implementation is a plain alias. Since the original specs are unchanged git recognice a new file.

straight-shoota:
> Since there is no Compress namespace, there is no need to create it only for Deflate.

I don't follow that argument. #8886 suggests to move `Gzip`, `Zlib`, `Zip` to the `Compress` namespace as well. So it wouldn't be "only for `Deflate`".

Moving them into a combined namespace would be consistent with `Random`, `Crypto` and `Digest` namespaces. It seems oddy that `Gzip`, `Zlib`, `Zip` are in the top level, while comparable modules like `MD5` and `SHA1` are grouped under `Digest`. I can't see any justification for doing differently.

bcardiff:
There were some concerns about the explicit `Compress::` that will be required.

The other PRs didn't got that much positive reactions, so I extract the part that was safe in this PR.

The main conflicting aspectes are:

* Grouping the modules for discoverability can be done in other ways than namespace.
* The names do not include library names, but formats. They are safe to be top-level.
* They are are not as common as serialization formats, so they could be grouped in a namespace.

RX14:
It's hard to gauge sentiment changing as the discussion progresses, most reactions on an issue are left in the first few hours, so I posted a vote.

bcardiff:
Rebased no master

PR #8958, opened 2020-03-27
Respect --no-color for the error indicator in stack traces

oprypin:
Regressed in #8892


oprypin:
Fun fact: this is how I discovered it :joy: 

screenshot from Windows terminal

PR #8538, opened 2019-11-30
Improve debugging support

skuznetsov:
Crystal debug support is here. It allows LLDB to show Crystal variables properly in C lang style.

PR #8959, opened 2020-03-28
Add Enumerable#last/last? methods

Sija:
- Adds `Enumerable#last?`, `Enumerable#last` and `Enumerable#last(&)` methods
- Makes `Indexable#last?` and `Indexable#last` inherit from `Enumerable`
- Due to the above `Indexable#last` throws now `Enumerable::EmptyError` instead of `IndexError`

Aligns with #8999

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -493,8 +493,13 @@ module Enumerable(T)
     nil
   end
 
-  # Returns the last element in the collection. Raises `Enumerable::EmptyError`
-  # if the collection is empty.
+  # Returns the last element of `self` if it's not empty,
+  # or the given block's value.
+  #
+  # ```
+  # ([1, 2, 3]).last { 4 }   # => 3
+  # ([] of Int32).last { 4 } # => 4
+  # ```
   def last
     found = false
     obj = uninitialized T
@@ -507,13 +512,23 @@ module Enumerable(T)
     obj
   end
 
-  # :ditto:
+  # Returns the last element of `self` if it's not empty,
+  # or raises `Enumerable::EmptyError`.
+  #
+  # ```
+  # ([1, 2, 3]).last   # => 3
+  # ([] of Int32).last # raises Enumerable::EmptyError
+  # ```
   def last
     last { raise Enumerable::EmptyError.new }
   end
 
-  # Returns the last element in the collection.
-  # When the collection is empty, returns `nil`.
+  # Returns the last element of `self` if it's not empty, or `nil`.
+  #
+  # ```
+  # ([1, 2, 3]).last?   # => 3
+  # ([] of Int32).last? # => nil
+  # ```
   def last?
     last { nil }
   end
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -493,6 +493,31 @@ module Enumerable(T)
     nil
   end
 
+  # Returns the last element in the collection. Raises `Enumerable::EmptyError`
+  # if the collection is empty.
+  def last
+    found = false
+    obj = uninitialized T
+
+    each do |elem|
+      found = true
+      obj = elem
+    end
+    return yield unless found
+    obj
+  end
+
+  # :ditto:
+  def last
+    last { raise Enumerable::EmptyError.new }
+  end
+
+  # Returns the last element in the collection.
+  # When the collection is empty, returns `nil`.
+  def last?
+    last { nil }
+  end
+
   # Returns a new array with the concatenated results of running the block
   # (which is expected to return arrays) once for every element in the collection.
   #
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -887,7 +887,7 @@ describe "Array" do
     end
 
     it "raises when empty" do
-      expect_raises IndexError do
+      expect_raises Enumerable::EmptyError do
         ([] of Int32).last
       end
     end
--- a/spec/std/deque_spec.cr
+++ b/spec/std/deque_spec.cr
@@ -380,7 +380,7 @@ describe "Deque" do
     end
 
     it "raises when empty" do
-      expect_raises IndexError do
+      expect_raises Enumerable::EmptyError do
         Deque(Int32).new.last
       end
     end
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -466,6 +466,38 @@ describe "Enumerable" do
     end
   end
 
+  describe "last" do
+    it "gets last" do
+      (1..3).last.should eq(3)
+    end
+
+    it "calls block if empty" do
+      (1...1).last { 10 }.should eq(10)
+    end
+
+    it "raises if enumerable is empty" do
+      expect_raises Enumerable::EmptyError do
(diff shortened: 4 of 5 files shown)


asterite:
These aren't in Ruby either.

The reason is that it's very inefficient to get the last element of an Enumerable by traversing it completely. That's why `last` should be defined separately for each container in an efficient way.

I'm not sure we should merge this.

Sija:
@asterite Of course it is not, but that's the most generalised version, which nevertheless comes handy when you need it - i.e. you only include `Enumerable` and that's it. More specialised versions are already in place for `Indexable` and friends.

src/enumerable.cr
```diff
@@ -493,6 +493,21 @@ module Enumerable(T)
     nil
   end
 
+  # Returns the last element in the collection. Raises `Enumerable::EmptyError`
+  # if the collection is empty.
+  def last
+    raise Enumerable::EmptyError.new unless any? { true }
```
asterite: Thus traverses the collection twice. You should use uninitialized, a flag, and traverse once. There are other similar methods that do this, like max.

asterite:
I think this is good to have if:
- we implement it by traversing it just once
- we mention in the docs that it traverses the entire collection by default, but usually subclasses override this with a more performant implementation (like the case of Indexable, Array, Set, etc.).

src/enumerable.cr
Sija: Since this whole approach is not especially effective than gaining **one** iteration doesn't seem worth to me going all fancy with optimizations here...

asterite: It's not an optimizaton. Enumerable methods implicitly guarantee that the collection is traversed just once.

If you want, I can send a separate PR to implement it.

Sija: Done, have a look is it what you had in mind.

asterite: It looks good!

> Didn't know about that. Where's it stated btw?

It's not stated, but for stdlib tries it's best to avoid repeating logic.

I think there's no need for `last_internal`. For `last` we need to use `uninitialized`. But for `last?` we can do it like this:

```crystal
def last?
  obj = nil
  each { |e| obj = e }
  obj
end
```

Sija: Done and done.

src/enumerable.cr
```diff
@@ -493,6 +493,28 @@ module Enumerable(T)
     nil
   end
 
+  # Returns the last element in the collection. Raises `Enumerable::EmptyError`
+  # if the collection is empty.
+  def last
+    found = false
+    obj = uninitialized T
```
oprypin: I don't understand, are we admitting that it's OK to use `uninitialized` in high-level code?

asterite: It could be seen as high-level code. It can also be seen as a fundamental thing in the stdlib. I know Ruby does all kind of things in C in their stdlib code, so maybe that's fine?

That said... do you think there's a way to implement this without `uninitialized`?

Sija: Well, I don't know about that. It is being used there anyway, so if you think that's wrong please open a separate issue about that, thanks.

Sija: > That said... do you think there's a way to implement this without `uninitialized`?

@asterite I can think of one at least, but given there's plenty of other places within `Enumerable` where this pattern is being used, I don't think we should go into some obscure mambo-jumbo just for an abstract reasoning - whether or not it is, or it is not _OK_ - implementation is safe & simple, what's wrong with that?

Blacksmoke16: Could you not just do like:
```cr
def last
  obj = nil
  each do |elem|
    obj = elem
  end
  raise Enumerable::EmptyError.new if obj.nil?
  obj
end
```

oprypin: This doesn't understand containers which can actually have `nil`

asterite: That's exactly the problem. I guess we can have a marker type which isn't expected to be in any collection. For example:

```crystal
module Enumerable(T)
  record Dummy

  def last
    obj = Dummy
    each { |e| obj = e }
    raise Enumerable::EmptyError.new if obj.is_a?(Dummy.class)
    obj
  end
end
```

What do you think?

Not that we want to do it here, but maybe for a later iteration. Maybe we could even make it a public type and explain how it can't be used. I think we could use it in `Channel` too, and many other places where `uninitialized` is used. Then we can almost remove `uninitialized` from the language, or just keep it for `StaticArray`. Then nobody will get crashes because of misuse of `uninitialized`.

Sija: @asterite Sounds good, we already have a `Iterator::Stop` which kinda "paves the way" serving similar purpose.

straight-shoota: Alternative implementation would be a separate flag variable.

Sija:
> Especially when the reason to add it would only be "seems oddly missing".

@straight-shoota I've recently stumbled upon the use case in https://github.com/crystal-lang/crystal/pull/8538/files#diff-3fbcce607939d327f6e0f379e711052e

asterite:
@straight-shoota I used to think the same about adding `last` to `Enumerable`: what happens if the Enumerable is infinite? Well, the method simply crashes. There's also `map`, `select`, etc., all of which will crash on an infinite enumerable. Yet they are still there. So I think having `last` there for convenience is good.

oprypin:
Wait, why would this method crash on an infinite Enumerable?

Sija:
@oprypin lol, I guess @asterite meant _hang_ (infinitely) ;)

Sija:
Same goes for `count` btw.

straight-shoota:
`map` and `select` at least *can* be useful even with infinite enumerations when the block variant is used because you can escape from the iteration.

Sija:
@straight-shoota All in all I don't it matters whether the code actually crashes or hangs forever, as both are unwanted behaviours anyway. IMO the only thing we're missing is a proper addition to the docs explicitly stating that such behaviour might occur with infinite enumerables and some of these methods might be unsafe to use with such.

Sija:
> A drawback of the two separate implementations is that inheritors of the module will need to override both to have an efficient implementation.
> 
> If we were to add this I think we should have a single method to override. This will imply changing Indexable probably to use that.

Makes sense, but what about other methods like this - i.e. `first?` & `first` for example?

> And the docs part, suggesting that it should be overriden in case there is an efficient way to implement it should be done in the PR.

Agreed.

bcardiff:
> Makes sense, but what about other methods like this - i.e. first? & first for example?

Since each iterates starting from the beginning there is no need for further optimizations.

Sija:
> If we were to add this I think we should have a single method to override. This will imply changing Indexable probably to use that.

Changed. I've also synced this PR with #8999 in re: to adding block version of `#last`, throwing `Enumerable::EmptyError` instead of `IndexError` and inheriting additional methods + docs.

asterite:
Yes, I'm not sure about this either. I actually prefer the unrolled loop in the code that interacts with the C code to make it clear it loops over all of the items.

bcardiff:
Closing, there seems to not be enough 👍 for merging this. If containers can provide an efficient `#last`, then that should be enough.

PR #8960, opened 2020-03-28
Add Enumerable#empty? method

Sija:
For some reason there's only `Indexable#empty?`, whereas IMHO `Enumerable` deserves it's own one too.

### DIFF
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -10,6 +10,13 @@ private class SpecEnumerable
   end
 end
 
+private class SpecEmptyEnumerable
+  include Enumerable(Int32)
+
+  def each(&block : T -> _)
+  end
+end
+
 describe "Enumerable" do
   describe "all? with block" do
     it "returns true" do
@@ -414,6 +421,11 @@ describe "Enumerable" do
       iter.next.should eq({2, "a"})
       iter.next.should be_a(Iterator::Stop)
     end
+  end
+
+  describe "empty?" do
+    it { SpecEnumerable.new.empty?.should be_false }
+    it { SpecEmptyEnumerable.new.empty?.should be_true }
   end
 
   describe "find" do
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -1241,6 +1241,17 @@ module Enumerable(T)
     count { true }
   end
 
+  # Returns `true` if `self` is empty, `false` otherwise.
+  #
+  # ```
+  # ([] of Int32).empty? # => true
+  # ([1]).empty?         # => false
+  # ```
+  def empty?
+    each { return false }
+    true
+  end
+
   # Returns an `Array` with the first *count* elements removed
   # from the original collection.
   #


asterite:
Well, there's the fact that Ruby doesn't have `empty?` in Enumerable. I don't know exactly why.

src/enumerable.cr
```diff
@@ -1241,6 +1241,16 @@ module Enumerable(T)
...
   end
 
+  # Returns `true` if `self` is empty, `false` otherwise.
+  #
+  # ```
+  # ([] of Int32).empty? # => true
+  # ([1]).empty?         # => false
+  # ```
+  def empty?
+    !any? { true }
```
asterite:
```suggestion
    each { return true }
    false
```

spec/std/enumerable_spec.cr
```diff
@@ -416,6 +416,11 @@ describe "Enumerable" do
     end
   end
 
+  describe "empty?" do
+    it { %w[foo bar].empty?.should be_false }
+    it { %w[].empty?.should be_true }
```
jhass: This is testing `Indexable#empty?` since `Array` includes that.

Could just check `.each.empty?` for now since `Iterator` is `Enumerable` but not `Indexable`, but I think I would prefer a custom type including `Enumerable` for the test to future proof it from testing the wrong thing in case we add  `Iterator#empty?` for some reason.

src/enumerable.cr
```diff
@@ -1241,6 +1241,17 @@ module Enumerable(T)
...
+  # Returns `true` if `self` is empty, `false` otherwise.
+  #
+  # ```
+  # ([] of Int32).empty? # => true
+  # ([1]).empty?         # => false
+  # ```
+  def empty?
+    each { return true }
+    false
+  end
```
jhass: This is backwards:

https://carc.in/#/r/8smy

Sija: lol 🤦‍♂, good catch!

spec/std/enumerable_spec.cr
Sija: > This is testing `Indexable#empty?` since `Array` includes that.

That's correct, and it's a current situation for many other methods tested in the spec btw...

asterite: You should use `SpecEnumerable` defined in that spec file.

asterite: And possibly introduce another one to prove that `empty?` returns `true`. Maybe change `SpecEnumerable` to wrap an array and call `each` on it. That way testing is easier. And we could eventually use that throughout the whole spec file.

Sija: That's what I ended up doing, see the force-pushed code.

straight-shoota:
Dziekuje, @Sija

Sija:
Is it gonna land in `0.34.0`? There's no tag, that's why I'm asking.

straight-shoota:
Sorry, missed that. Everything's gonna be in the next release until its tagged =)

PR #8969, opened 2020-03-28
Compiler: let Proc(T) pass a restriction of Proc(Nil)

### LINKED ISSUE #8964: `Proc(Nil)` in a method type restriction does not work like it does elsewhere
> Added $170 bounty to this https://www.bountysource.com/issues/90719678-proc-nil-in-a-method-type-restriction-does-not-work-like-it-does-elsewhere

#7527 the *very* handy ability to use Proc with a Nil return type restriction (e.g. `Proc(Nil)`)  and have Procs that return non-nil values works

This allows you to do something like:

```crystal
class ProcNil
  getter proc : Proc(Nil)
 
  def initialize
    @proc = -> { 123 }
  end
end
 
pp! ProcNil.new.proc.call # Which will return nil
```

### The problem

Using a `Proc(Nil)` when restricting in a method argument does *not* work the same. It requires an explicit nil return value in the Proc

```crystal
class ProcNil
  def initialize(proc : Proc(Nil))
  end
end

# Error: no overload matches 'ProcNil.new', proc: Proc(Int32)
#
# Overloads are:
#  - ProcNil.new(proc : Proc(Nil))
pp! ProcNil.new(proc: ->{123})
```

To get this to work you'd need to explicitly return `nil`: 

```crystal
ProcNil.new(proc: ->{123;nil}
```

### Where the problem seems to be

I think the reason the version with `getter` works is that there is an odd behavior when using `Proc(Nil)` with a method

```crystal
def my_proc : Proc(Nil)
  ->{ 123 }
end

# You'd expect this to be `nil`
pp! my_proc.call # 123
# And this to be `Proc(Nil)
pp! typeof(my_proc) # Proc(Int32)
````

So it seems like this may be 2 issues. The type restriction in a method incorrectly return a `Proc(Int32)` even though the restriction is `Proc(Nil)`. This by happenstance allows the first example to work

### Fixing this

This is something I would really love to use in my Lucky projects to do stuff like this for generic HTML components where you can add Proc "slots" for passing in HTML https://gist.github.com/paulcsmith/ec853d33d1a0ef9b1eedd070e58136b9 

So I've added a $170 bounty to this since I know I'm asking for a potentially complicated fix! :D https://www.bountysource.com/issues/90719678-proc-nil-in-a-method-type-restriction-does-not-work-like-it-does-elsewhere

My hope is this also helps someone during the crazy economy/COVID-19

### PR DESCRIPTION
**Pull request by asterite**
Fixes #8964

### DIFF
--- a/src/compiler/crystal/semantic/restrictions.cr
+++ b/src/compiler/crystal/semantic/restrictions.cr
@@ -1175,12 +1175,25 @@ module Crystal
       end
 
       other.type_vars.each_with_index do |other_type_var, i|
+        # If checking the return type, any type matches Nil
+        if i == other.type_vars.size - 1 && nil_type?(other_type_var, context)
+          # Also, all other types matched, so the matching type is this proc type
+          # except that it has a Nil return type
+          new_proc_arg_types = arg_types.dup
+          new_proc_arg_types << program.nil_type
+          return program.proc_of(new_proc_arg_types)
+        end
+
         proc_type = arg_types[i]? || return_type
         restricted = proc_type.restrict other_type_var, context
         return nil unless restricted == proc_type
       end
 
       self
+    end
+
+    def nil_type?(node, context)
+      node.is_a?(Path) && context.defining_type.lookup_path(node).is_a?(NilType)
     end
 
     def compatible_with?(other : ProcInstanceType)
--- a/spec/compiler/codegen/proc_spec.cr
+++ b/spec/compiler/codegen/proc_spec.cr
@@ -761,4 +761,17 @@ describe "Code gen: proc" do
       f.call(20)
       )).to_i.should eq(21)
   end
+
+  it "can pass Proc(T) to Proc(Nil) in type restriction (#8964)" do
+    run(%(
+      def foo(x : Proc(Nil))
+        x
+      end
+
+      a = 1
+      proc = foo(->{ a = 2 })
+      proc.call
+      a
+      )).to_i.should eq(2)
+  end
 end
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/compiler/crystal/semantic/restrictions.cr [hunk 1]
```diff
@@ -1175,6 +1175,15 @@ module Crystal
       end
 
       other.type_vars.each_with_index do |other_type_var, i|
+        # If checking the return type, any type matches Nil
+        if i == other.type_vars.size - 1 && nil_type?(other_type_var, context)
+          # Also, all other types matched, so the matching type is this proc type
+          # except that it has a Nil return type
```
bcardiff: I thought the proc were honoring subtyping of arrow types.

It is safe for the following example to compile.

```cr
class Foo
end
 
class Bar < Foo
end
 
def foo(x : Proc(Bar, Nil))
  x
end
 
a = 1
 
proc = foo(->(b : Foo) { a = 2 })
r = proc.call(Bar.new)
```

It is subject for a later change, but is the subtying of arrow type in procs that we are missing it seems.

asterite: Yeah, we need to make subtyping work, but it won't work all of the time: only for class references. For structs it's not possible because the representation is different. Also for unions, I think. So for now I didn't implement it until we think it well. But matching against `Nil` is more common and easier to check.

asterite:
I won't merge any more PRs because I don't know whether they should be in 0.34.0 or later.

bcardiff:
I think this is safe to fit into 0.34 since it's handling more cases without loosing or changing others. On my end, feel free to merge it if you want.

paulcsmith:
Thanks @asterite!

PR #8970, opened 2020-03-28
Compiler: allow assigning Proc(R) to Proc(Nil)

### LINKED ISSUE #7698: Proc with Nil return type causes other procs to always return nil
Started happening after #7527.

Code:
```cr
abstract struct Action; end

record RouteAction(A) < Action, action : A

class Handler
  getter actions : Array(Action) = [] of Action

  def initialize
    action1_proc = ->(val : Int32) do
      nil
    end
    action1 = RouteAction(Proc(Int32, Nil)).new action1_proc

    actions << action1

    action2_proc = ->(val : Int32) do
      val
    end
    action2 = RouteAction(Proc(Int32, Int32)).new action2_proc

    actions << action2
  end
end

handler = Handler.new

pp handler.actions[0].action.call 123
pp handler.actions[1].action.call 123
```
https://play.crystal-lang.org/#/r/6ram

Result before (0.27.2):
```
nil
123
```

Result after (0.28.0):
```
nil
nil
```
Currently this has totally broken Athena for anyone who defined a route action with a return type of `Nil`.

asterite:
This PR:
- reverts #7527
- fixes #7698

Then it implements what #7527 was trying to accomplish in a different way.

What #7527 did was merging any two proc types with same argument types but different return types, where one return type was nil, to that return type with nil. Example:

```crystal
a = ->{ 1 }
b = ->{ nil }  # Proc(Nil)
typeof(a)      # => Proc(Int32)
typeof(b)      # => Proc(Int32)
typeof(a || b) # => Proc(Nil)
```

The problem is, it's not good to discard the information returned by `b`.

What this PR does is just allowing the assignment of such procs to instance and class variables, and local variables with a type annotation. For example:

```crystal
class Foo
  @block : -> Nil

  def initialize(block)
    @block = block
  end

  def block
    @block
  end
end

a = ->{ 1 }
b = ->{ nil }

Foo.new(a) # okay
Foo.new(b) # okay

c = a || b
typeof(c) # => Proc(Int32) | Proc(Nil)
Foo.new(c) # also okay!
```

Note that `c` preserves the real type. But when we assign it to the instance variable `@block`, which is already typed to be `Proc(Nil)`, just then the information is lost. And this is fine because, since the return type is `Nil` there's no other thing to do than losing that information.

The compiler had no such logic for this before. The only thing the compiler did, all the time was:
- if you assign to a variable, you make the union of the variable's existing type with the new type, then you assign that to the variable

So, this introduces new logic to apply just when assigning a type to a variable whose type is already determined by the user. This is the new `assign_type` method. You don't need to know about this. I'm just mentioning it here because this might feel like a cheap implementation. However, this is the **only** case need so far. If we need more special cases like this, we can refactor the code to make it a bit more generic.

Also, because the compiler will now try to do this when you specify a type, this now behaves like this:

```crystal
def foo : Proc(Nil)
  ->{ 1 }
end

typeof(foo) # => Proc(Nil)
```

Which I think is good. If you explicitly declare the return type of the Proc to be Nil, the only reason you might want to do that (other than verify the type matches) is to return such type, not a type with more information in the return type. So this would be analogous to explicitly using the `Nil` return type.

Finally, I think this is important to have in the language because the case of wanting to store callbacks is very common. In many cases callbacks don't return anything, and forcing the users to type that extra `nil` inside their procs is not nice. Also, the compiler error of that is probably not very intuitive.

### DIFF
--- a/spec/compiler/codegen/proc_spec.cr
+++ b/spec/compiler/codegen/proc_spec.cr
@@ -774,4 +774,49 @@ describe "Code gen: proc" do
       a
       )).to_i.should eq(2)
   end
+
+  it "can assign proc that returns anything to proc that returns nil (#3655)" do
+    run(%(
+      class Foo
+        @block : -> Nil
+
+        def initialize(@block)
+        end
+
+        def call
+          @block.call
+        end
+      end
+
+      a = 1
+      block = ->{ a = 2 }
+
+      Foo.new(block).call
+
+      a
+      )).to_i.should eq(2)
+  end
+
+  it "can assign proc that returns anything to proc that returns nil, using union type (#3655)" do
+    run(%(
+      class Foo
+        @block : -> Nil
+
+        def initialize(@block)
+        end
+
+        def call
+          @block.call
+        end
+      end
+
+      a = 1
+      block1 = ->{ a = 2 }
+      block2 = ->{ a = 3; nil }
+
+      Foo.new(block2 || block1).call
+
+      a
+      )).to_i.should eq(3)
+  end
 end
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -1051,4 +1051,14 @@ describe "Semantic: proc" do
       foo(->(x : String) { 1 })
       )) { proc_of string, nil_type }
   end
+
+  it "casts to Proc(Nil) when specified in return type" do
+    assert_type(%(
+      def foo : Proc(Nil)
+        ->{ 1 }
+      end
+
+      foo
+      )) { proc_of nil_type }
+  end
 end
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -889,7 +889,7 @@ describe "Semantic: proc" do
       )) { union_of proc_of(int32, int32), proc_of(int32, nil_type) }
   end
 
-  it "can assign proc that returns anything to proc that returns nil (#3655)" do
+  it "can assign proc that returns anything to proc that returns nil, with instance var (#3655)" do
     assert_type(%(
       class Foo
         @block : -> Nil
@@ -904,6 +904,36 @@ describe "Semantic: proc" do
       end
 
       Foo.new.block
+      )) { proc_of(nil_type) }
+  end
+
+  it "can assign proc that returns anything to proc that returns nil, with class var (#3655)" do
+    assert_type(%(
+      module Moo
+        @@block : -> Nil = ->{ nil }
+
+        def self.block=(@@block)
+        end
+
+        def self.block
+          @@block
+        end
+      end
+
+      Moo.block = ->{ 1 }
+      Moo.block
+      )) { proc_of(nil_type) }
+  end
+
+  it "can assign proc that returns anything to proc that returns nil, with local var (#3655)" do
+    assert_type(%(
+      proc : -> Nil
+
+      a = ->{ 1 }
+      b = ->{ nil }
+      proc = a || b
+
+      proc
       )) { proc_of(nil_type) }
   end
 
--- a/spec/compiler/semantic/proc_spec.cr
+++ b/spec/compiler/semantic/proc_spec.cr
@@ -889,6 +889,14 @@ describe "Semantic: proc" do
       )) { union_of proc_of(int32, int32), proc_of(int32, nil_type) }
   end
 
+  it "merges return type" do
+    assert_type(%(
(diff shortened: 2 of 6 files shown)


src/compiler/crystal/semantic/bindings.cr
```diff
@@ -215,6 +224,22 @@ module Crystal
       type
     end
 
+    # Computes the type resulting from assigning type to freeze_type,
+    # in the case where freeze_type is not nil.
+    #
+    # Special cases are listed inside the method body.
+    def assign_type(freeze_type, type)
```
Sija: I'd suggest renaming this method due to the fact it doesn't actually _do_ anything, it just returns the desired type:

```suggestion
    def assignment_type(freeze_type, type)
```

asterite: Thanks! Yeah, I'll actually rename it to something more descriptive.

asterite: On second thought, I prefer to leave it like that. We have an `Crystal::Assign` ASTNode that represent an assignment. It's called Assign... I don't know why. But I prefer to leave that name too for the method so it's clear.

Also, we have `map_type` a few lines above. It maps the type to another type. This just assigns the type to `freeze_type` and returns the resulting type. I think it's clear (from a compiler perspective).

asterite: Well, I ended up renaming it 🤷‍♂

spec/compiler/codegen/proc_spec.cr
```diff
@@ -761,4 +761,49 @@ describe "Code gen: proc" do
       f.call(20)
       )).to_i.should eq(21)
   end
+
+  it "can assign proc that returns anything to proc that returns nil (#3655)" do
+    run(%(
+      class Foo
+        @block : -> Nil
```
yxhuvud: Hmm. What are the corresponding rules around `@block : -> Void`?

asterite: Same. `Void` in Proc return type is the same as `Nil`.

RX14:
Does this work for container types too, like `Array(-> Nil)` being able to take a `-> Int32`? In effect, does this casting still work in type restrictions too?

asterite:
> Does this work for container types too, like Array(-> Nil) being able to take a -> Int32? In effect, does this casting still work in type restrictions too?

Apparently yes. There's also #8969 which should be merged too, but without that it still works for Arrays for some reason... I think it's because the restriction is `T`, then a different code path is taken and that codepath is already covered.

spec/compiler/codegen/proc_spec.cr
```diff
@@ -761,4 +761,49 @@ describe "Code gen: proc" do
...
+        def call
+          @block.call
+        end
+      end
+
+      a = 1
+      block1 = ->{ a = 2 }
+      block2 = ->{ a = 3; nil }
+
+      Foo.new(block2 || block1).call
```
bcardiff: Does `(block2 || block1).call` compile and yields a `Int32?` as a result?

asterite: Yes :-)

bcardiff: Is there a spec guaranteeing that? I could not find it.

spec/compiler/semantic/proc_spec.cr
```diff
@@ -881,15 +881,15 @@ describe "Semantic: proc" do
...
   end
 
-  it "merges Proc that returns Nil with another one that returns something else (#3655)" do
+  it "*doesn't* merge Proc that returns Nil with another one that returns something else (#3655) (this was reverted)" do
     assert_type(%(
       a = ->(x : Int32) { 1 }
       b = ->(x : Int32) { nil }
       a || b
-      )) { proc_of(int32, nil_type) }
+      )) { union_of proc_of(int32, int32), proc_of(int32, nil_type) }
```
asterite: @bcardiff This is similar to the spec that you want. In a previous PR I made this always be `Proc(Int32, Nil)`, and this PR reverts that. But I will add a test that `call` returns the union.

spec/compiler/codegen/proc_spec.cr
asterite: Done!

spec/compiler/semantic/proc_spec.cr
bcardiff: Cool, thanks. Now we need a rebase on master due to conflicts

asterite: Done!

PR #8972, opened 2020-03-29
Simplify Link annotation handling

RX14:
This pull request deprecates the `@[Link("lib", static: true)]` request to prefer static linking of libraries. The only use of this is a `@[Link("gc", static: true)]` in the crystal repo. I have been unable to find any exmaples of `static: true` being used outside of this repo. The reason for this, I suspect, is because it doesn't work. People who want the existing behaviour can simply pass `ldflags: "-l :libgc.a"`

Also deprecated is passing positional arguments to `@[Link]` other than "lib" (nobody uses that either).

Because of these simplifications, we can implement a new, simpler, linking algorithm with less code.

First, a `pkg_config` link argument is added, since almost all libraries have a different pkg-config name to the `lib<foo>.so` name. For example, `libz.so` is `zlib.pc`, `libgc.so` is `bdw-gc.pc`. Only *one* of 10 different libs used by the crystal standard library was correctly detected using the original mechanism. Allowing pkg-config module names to be specified correctly allows the pkg-config mechanism to work properly, which makes linking far more reliable. The fallback remains `-lfoo` as before.

When neither `static: true` or `--static` is specified, the current algorithm is identical to before, but with the `pkg_config` argument added. `static: true` is now ignored, and passing `--static` simply passes `-static` to the linker and `--static` to `pkg-config`.

### DIFF
--- a/src/annotations.cr
+++ b/src/annotations.cr
@@ -27,7 +27,7 @@ end
 annotation Flags
 end
 
-# A `lib` can be marked with `@[Link(lib : String, ldflags : String, static : Bool, framework : String)]`
+# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, static : Bool, framework : String)]`
 # to declare the library that should be linked when compiling the program.
 #
 # At least one of the *lib*, *ldflags*, *framework* arguments needs to be specified.
--- a/src/annotations.cr
+++ b/src/annotations.cr
@@ -38,11 +38,6 @@ end
 # 1. will lookup `pcre` using `pkg-config`, if not found
 # 2. will pass `-lpcre` to the linker.
 #
-# `@[Link("pcre", static: true)]` will favor static libraries over shared libraries.
-# 1. will lookup `libpcre.a` in `CRYSTAL_LIBRARY_PATH`, if not found
-# 2. will lookup `pcre` using `pkg-config --static`, if not found,
-# 3. will lookup `libpcre.a` in `/usr/lib`, `/usr/local/lib`
-#
 # `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker.
 #
 # When an `-l` option is passed to the linker, it will lookup the libraries in
--- a/src/annotations.cr
+++ b/src/annotations.cr
@@ -27,7 +27,7 @@ end
 annotation Flags
 end
 
-# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, static : Bool, framework : String)]`
+# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, static : Bool, framework : String, pkg_config : String)]`
 # to declare the library that should be linked when compiling the program.
 #
 # At least one of the *lib*, *ldflags*, *framework* arguments needs to be specified.
@@ -36,6 +36,11 @@ end
 #
 # `@[Link("pcre")]` will lookup for a shared library.
 # 1. will lookup `pcre` using `pkg-config`, if not found
+# 2. will pass `-lpcre` to the linker.
+#
+# `@[Link("pcre", pkg_config: "libpcre")]` will lookup for a shared library.
+# 1. will lookup `libpcre` using `pkg-config`, if not found
+# 2. will lookup `pcre` using `pkg-config`, if not found
 # 2. will pass `-lpcre` to the linker.
 #
 # `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker.
--- a/src/annotations.cr
+++ b/src/annotations.cr
@@ -27,7 +27,7 @@ end
 annotation Flags
 end
 
-# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, static : Bool, framework : String, pkg_config : String)]`
+# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, framework : String, pkg_config : String)]`
 # to declare the library that should be linked when compiling the program.
 #
 # At least one of the *lib*, *ldflags*, *framework* arguments needs to be specified.
@@ -46,9 +46,9 @@ end
 # `@[Link(framework: "Cocoa")]` will pass `-framework Cocoa` to the linker.
 #
 # When an `-l` option is passed to the linker, it will lookup the libraries in
-# paths passed with the `-L` option. `CRYSTAL_LIBRARY_PATH`, `/usr/lib`,
-# and `/usr/local/lib` are added by default. Custom paths can be passed
-# using `ldflags`: `@[Link(ldflags: "-Lvendor/bin")]`.
+# paths passed with the `-L` option. Any paths in `CRYSTAL_LIBRARY_PATH` are
+# added by default. Custom paths can be passed using `ldflags`:
+# `@[Link(ldflags: "-Lvendor/bin")]`.
 annotation Link
 end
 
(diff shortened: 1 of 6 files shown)


src/annotations.cr
```diff
@@ -27,7 +27,7 @@ end
 annotation Flags
 end
 
-# A `lib` can be marked with `@[Link(lib : String, ldflags : String, static : Bool, framework : String)]`
+# A `lib` can be marked with `@[Link(lib : String, *, ldflags : String, static : Bool, framework : String, pkg_config : String)]`
```
straight-shoota: You should add that `static` is deprecated. Not in this line but further down. And maybe not mention it here at all.

straight-shoota: Should also note that `CRYSTAL_LIBRARY_PATH` is added to the lookup paths.

src/compiler/crystal/codegen/link.cr
```diff
@@ -105,83 +110,68 @@ module Crystal
...
-          flags.join ' '
-        else
-          `pkg-config #{libname} --libs`.chomp
-        end
+    PKG_CONFIG_PATH = Process.find_executable("pkg-config")
+
+    # Returns the result of running `pkg-config mod` but returns nil if
+    # pkg-config is not installed, or the module does not exist.
+    private def pkg_config(mod, static = false) : String?
+      return unless pkg_config_path = PKG_CONFIG_PATH
```
straight-shoota: Stylistic: It is really hard to follow the control flow when the actual effect (i.e. assignment to local var) happens in a trailing conditional.
```suggestion
      pkg_config_path = PKG_CONFIG_PATH || return
```
`return unless pkg_config_path` could also be on a separate line.

RX14: thanks i hate it

src/compiler/crystal/codegen/link.cr
```diff
@@ -105,83 +110,68 @@ module Crystal
...
+        if ldflags = ann.ldflags
+          flags << ldflags
         end
 
-        # Append the CRYSTAL_LIBRARY_PATH values as -L flags.
-        CrystalLibraryPath.paths.each do |path|
-          flags << " -L#{path}"
+        # First, check pkg-config for the pkg-config module name if provided, then
+        # check pkg-config with the lib name, then fall back to -lname
+        if (pkg_config_name = ann.pkg_config) && (flag = pkg_config(pkg_config_name, static_build))
```
bcardiff, line 132: This is giving more precedence to the pkg-config settings for the lib, instead of giving priority to the CRYSTAL_LIBRARY_PATH as it should. Otherwise having a local package for bdw-gc will override the patched embedded version we ship.

RX14: yeah, it will, I don't think `CRYSTAL_LIBRARY_PATH` was a good idea because it binds you to the "path" model of library lookup, which is outdated and replaced with pkgconfig (and `PKG_CONFIG_PATH`)

The better solution is `CRYSTAL_LIBRARY_OVERRIDES="gc=/path/to/whatever otherlib=-lwhatever"` which is far more robust and flexible.

waj:
@RX14 I like the overall idea of this PR. The `Link` annotation needed some love for long time.

However, I'm not so sure yet about deprecating the `static` option. This was introduced long time ago (7ec3a82d800aec4c171c77f11011a3f9e4aea630) to solve a specific issue. I'm not proud at all of that code and I'm open to suggestions for alternatives, but this workaround was used to force static linking when both static and dynamic libraries are available, specially when trying to use a patched version of the GC for example, while the system wide library is still available. Also, the `-l:libgc.a` option doesn't work everywhere (it doesn't on macOS). I'll be really happy actually if that library lookup is removed from the source, as I understand that should be the task of the linker only, but let's make sure all scenarios are still supported.

I love the idea of `CRYSTAL_LIBRARY_OVERRIDES` and I think we should implement it soon (probably after 1.0 though). Now, to make it work it needs a name to use as reference. The `lib` argument is now optional, so I'm thinking we should make it mandatory (and maybe rename it to just `name`?). What do you think?

asterite:
`CRYSTAL_LIBRARY_OVERRIDES` can work by specifying the name of a `lib`. For example `lib LibPCRE`. Then you would pass `CRYSTAL_LIBRARY_OVERRIDES="LibPCRE=..."`. There's really no need for the "name" to be what you specify to override. Specially, with the approach proposed here we have to check whether "name" was already specified somewhere else, but there's no way to define two "LibPCRE", there's always just one or none.

By the way, I mentioned doing it this way more than a year ago: https://forum.crystal-lang.org/t/rfc-link-configuration/546/2?u=asterite . And there's no need for an environment variable to do it, it can just be a compiler flag.

waj:
@asterite There are some `lib` that has more than one `Link` annotation. Maybe overriding all of them at the same time it's what we want? I don't know.

jhass:
Yeah, the lib/link relationship is many to many, additionally to one lib definition having many link annotations, the same link annotation could also be on many lib definitions. Using the lib name the behavior seems unclear in either scenario.

RX14:
Yeah, I was thinking of making `Link`'s name mandatory.

I appreciate that `static: true` can be useful sometimes, but I've only found a single usage, so I'd like to *try* removing it and see if we get any complaints.

waj:
Oh, but I thought I made the issue clear. For example, if there is a system wide installed `libgc.so`, are you sure the `libgc.a` shipped with Crystal will still be picked? IIRC some linkers might prefer the shared object instead of the static library unless `--static` is specified.

RX14:
Presumably in the distribution, where we would ship a `libgc.a`, we'd use `CRYSTAL_LIBRARY_OVERRIDES`, because we want to override the linker's "common sense" that the system libraries are good enough.

bcardiff:
@RX14 can you rebase this on master due to conflicts?

RX14:
Yep.

RX14:
I'd like to implement `CRYSTAL_LIBRARY_OVERRIDES` sooner rather than later, and definitely start requiring a name for every link annotation, but yeah we should see how it goes.

I'll address those comments.

bcardiff:
I would avoid designing CRYSTAL_LIBRARY_OVERRIDES before we discuss the story of a compiler configuration files in the .crystal folder. That will probably cover many stories together.

How to override the libs link annotation is definitely one of the main stories to cover there.

src/annotations.cr
RX14: I tweaked the link annotation docs. `CRYSTAL_LIBRARY_PATH` was already mentioned further down, but I removed some old references to hardcoded paths.

I removed mention of `static` entirely, I presume that'll be gone by 1.0 if all goes well.

RX14:
> I would avoid designing CRYSTAL_LIBRARY_OVERRIDES before we discuss the story of a compiler configuration files in the .crystal folder. That will probably cover many stories together.

I think there's definitely need for a system/user-level override, and an env var is a good way to do that. Perhaps you want to wait on that but I'm not convinced it's neccesary.

src/annotations.cr
```diff
@@ -38,17 +38,17 @@ end
...
 # 2. will pass `-lpcre` to the linker.
 #
-# `@[Link("pcre", static: true)]` will favor static libraries over shared libraries.
-# 1. will lookup `libpcre.a` in `CRYSTAL_LIBRARY_PATH`, if not found
-# 2. will lookup `pcre` using `pkg-config --static`, if not found,
-# 3. will lookup `libpcre.a` in `/usr/lib`, `/usr/local/lib`
+# `@[Link("pcre", pkg_config: "libpcre")]` will lookup for a shared library.
+# 1. will lookup `libpcre` using `pkg-config`, if not found
+# 2. will lookup `pcre` using `pkg-config`, if not found
+# 2. will pass `-lpcre` to the linker.
```
bcardiff, line 44:
```suggestion
# 3. will pass `-lpcre` to the linker.
```

PR #8975, opened 2020-03-30
Correctly support websocket close codes

### LINKED ISSUE #6469: WebSocket close frame messages aren't being parsed

RX14:
The first two bytes of the message in a WebSocket Close frame is anumeric code. Make sure to always encode this correctly, to avoid being in violation of the websocket spec. Also automatically decode this for clients.

This is a breaking change since on_close now takes two arguments: close code and message, but this change is unable to be made in a backwards compatible way.

Closes #6469

### DIFF
--- a/src/compiler/crystal/tools/playground/server.cr
+++ b/src/compiler/crystal/tools/playground/server.cr
@@ -490,7 +490,7 @@ module Crystal::Playground
         origin = context.request.headers["Origin"]
         if !accept_request?(origin)
           Log.warn { "Invalid Request Origin: #{origin}" }
-          ws.close "Invalid Request Origin"
+          ws.close HTTP::WebSocket::CloseCodes::PolicyViolation, "Invalid Request Origin"
         else
           @sessions_key += 1
           @sessions[@sessions_key] = session = Session.new(ws, @sessions_key, @port)
--- a/src/http/web_socket/close_codes.cr
+++ b/src/http/web_socket/close_codes.cr
@@ -1,0 +1,17 @@
+module HTTP::WebSocket::CloseCodes
+  NormalClosure           = 1000
+  GoingAway               = 1001
+  ProtocolError           = 1002
+  UnsupportedData         = 1003
+  NoStatusReceived        = 1005
+  AbnormalClosure         = 1006
+  InvalidFramePayloadData = 1007
+  PolicyViolation         = 1008
+  MessageTooBig           = 1009
+  MandatoryExtension      = 1010
+  InternalServerError     = 1011
+  ServiceRestart          = 1012
+  TryAgainLater           = 1013
+  BadGateway              = 1014
+  TLSHandshake            = 1015
+end
--- a/src/http/web_socket/protocol.cr
+++ b/src/http/web_socket/protocol.cr
@@ -233,13 +233,27 @@ class HTTP::WebSocket::Protocol
     end
   end
 
-  def close(message = nil)
+  def close(code : Int? = nil, message = nil)
     return if @io.closed?
+
     if message
-      send(message.to_slice, Opcode::CLOSE)
+      message = message.to_slice
+      code ||= CloseCodes::NormalClosure
+
+      payload = Bytes.new(2 + message.size)
+      IO::ByteFormat::NetworkEndian.encode(code.to_u16, payload)
+      message.copy_to(payload + 2)
     else
-      send(Bytes.empty, Opcode::CLOSE)
+      if code
+        payload = Bytes.new(2)
+        IO::ByteFormat::NetworkEndian.encode(code.to_u16, payload)
+      else
+        payload = Bytes.empty
+      end
     end
+
+    send(payload, Opcode::CLOSE)
+
     @io.close if @sync_close
   end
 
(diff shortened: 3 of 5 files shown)


src/compiler/crystal/tools/playground/server.cr
```diff
@@ -490,7 +490,7 @@ module Crystal::Playground
         origin = context.request.headers["Origin"]
         if !accept_request?(origin)
           Log.warn { "Invalid Request Origin: #{origin}" }
-          ws.close "Invalid Request Origin"
+          ws.close 1008, "Invalid Request Origin"
```
Blacksmoke16: Should these numeric codes be turned into an enum?  This would also allow the type restrictions to be a bit more clear on what the possible values are.

It could also allow for some method on the enum member that would return its byte representation.

src/http/web_socket/protocol.cr
```diff
@@ -233,13 +233,26 @@ class HTTP::WebSocket::Protocol
...
+
+    case {code, message}
+    when {nil, nil}
+      payload = Bytes.empty
+    when {Int, nil}
+      payload = Bytes.new(2)
+      IO::ByteFormat::NetworkEndian.encode(code.to_u16, payload)
     else
-      send(Bytes.empty, Opcode::CLOSE)
+      message = message.not_nil!.to_slice
```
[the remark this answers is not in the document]
RX14: There's a million ways to write this code, and all of them are a little ugly. I think I'd rather stick with this and not bikeshed.

src/compiler/crystal/tools/playground/server.cr
RX14: No, because there's a set of predefined ones *and* a range of possible undefined values (4000-4999). It'd turn into `WebSocket::CloseCode | Int` and it's probably not worth it.

There could be constants for the ~10 predefined codes, but otherwise I don't think there's much benefit.

src/http/web_socket/protocol.cr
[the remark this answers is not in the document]
asterite: I agree with avoiding bikeshedding. However, when I saw that `not_nil!` I thought "Hmm... but what happens if this is actually `nil`?". Then I had to look at the code above to realize that it's probably not possible for that to not be `nil`.

If `not_nil!` was never used here this would be much easier to understand and follow. But we don't need to do it now, I can send a following PR to change it.

src/compiler/crystal/tools/playground/server.cr
Blacksmoke16: Fair enough.  FWIW it could be done similar to `HTTP::Status`.  Where the known ones are defined, but you can still do `HTTP::Status.new 800`.

In which case it could validate that the custom value is within the required range.  This would keep the possible types to the enum itself.  It would also provide validation in what gets sent is valid at a cost of more verbosity.  `WebSocket::CloseCode.new(4500)` versus `4500`.

asterite: In the future I would like to disallow Enum with non-valid values

straight-shoota: If not an enum, maybe at least extract them to constants?

RX14:
Added `HTTP::WebSocket::CloseCodes` and refactored  the ugly case statement

src/http/web_socket/close_codes.cr
```diff
@@ -0,0 +1,14 @@
...
+  ProtocolError           = 1002
+  UnsupportedData         = 1003
+  NoStatusReceived        = 1005
+  AbnormalClosure         = 1006
+  InvalidFramePayloadData = 1007
+  PolicyViolation         = 1008
+  MessageTooBig           = 1009
+  MandatoryExtension      = 1010
+  InternalServerError     = 1011
+  TLSHandshake            = 1015
```
Sija: These look more like an enum members, constants are usually defined as `SNAKE_CASED_WITH_ALL_CAPS`.

Sija:
Just for the reference, here it is a list of error codes from IANA -> https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number

src/http/web_socket/close_codes.cr
RX14: This is intentional.

Sija: And what's the intention behind it?

RX14: to look like an enum

Sija: So why not make one (it's extensible, so there would be no need for an Union) instead of this?

RX14: code is art, everything is subjective, nothing matters, let it go

RX14:
This is essentially a bug in websockets right now, so I'd really like if this could make it into 0.34.0 because it's affecting a number of people I know right now.

RX14:
Squashed and added the extra close codes from the registry. I'll merge this once CI passes.

PR #8977, opened 2020-03-31
Fix the evaluation order of p! to match the expectation of a normal method

oprypin:
Evaluate the argument before starting to print anything.

Code example:
```crystal
p! "a" + p! "b"'
```
Expected output, also output with this change:
```
"b" # => "b"
"a" + (p!("b")) # => "ab"
```
Output before this change:
```
"a" + (p!("b")) # => "b" # => "b"
"ab"
```
Because it was being turned into an equivalent of
```crystal
::print %["a" + (p!("b")) # => ]
::p "a" + begin
    ::print %["b" # => ]
    ::p("b")
end
```
But now we turn it into an equivalent of
```crystal
result1 = begin
  result2 = "b"
  ::print %["b" # => ]
  ::p(result2)
end
::print %["a" + (p!("b")) # => ]
::p result1
```


straight-shoota:
A positive side effect of this behaviour was, that the expression was always printed even if executing it took forever. That's helpful to identify inifite loops and similar bugs. Also you still see the expression when the execution raises or terminates.
I'm not sure if that's valuable enough to consider keeping it as it is, but it is a benefit we're going to lose with this PR. I would miss it.

Obviously, the current behaviour is odd when `p!`'s are nested. But I'd probably accept that. Once you understand the behaviour, it's easy to avoid. And nesting `p!`'s is neither common nor really useful.

oprypin:
Well it was especially fun to figure out why output streams are interleaved *while porting stuff to Windows*. Is it printing to both stdout and stderr? Is flushing broken? Am I crashing in the middle?
The answer is neither.
I had many of those nested prints. Not sure why it's non typical.
Of course the code won't be in production, but when debugging it's a free-for-all.

Sija:
> Well it was especially fun to figure out why output streams are interleaved _while porting stuff to Windows_. Is it printing to both stdout and stderr? Is flushing broken? Am I crashing in the middle?
> The answer is neither.

Sounds like a debugging nightmare xD

> I had many of those nested prints. Not sure why it's non typical.

Neither do I, but it seems you're the first person here to report it...

oprypin:
I think this can get really confusing if one is deep into debugging. The nested `p!` don't even need to be in the same vicinity. There could just be long-running methods (possibly involving `spawn`), and one would `p!` the call to that, and also `p!` something inside (possibly indirectly inside) that method.

The lucky side effect is not something to be proud of, really. It is just an unexpected side effect regardless.
People used to this kind of debugging will not think to rely on this behavior, and will surely add another print statement before it for the same effect.

oprypin:
Aaaa please I need this.
Just found myself merging this commit into my hack branch to get some sanity back.

It's not like it's possible to just benefit from the abuse per https://github.com/crystal-lang/crystal/pull/8977#issuecomment-606868377 in a controlled manner, you **will** sooner or later run into a situation with two completely separate functions that just happen to call each other and also use `p!`.

```crystal
def foo
  p! bar + 1
end
 
def bar
  baz + 1
end
 
def baz
  p! 5
end
 
foo
```

```
bar + 1 # => 5 # => 5
7
```

WHY is this seen as OK

asterite:
Oh, yeah, I ran into that.

jhass:
I think @straight-shoota has a very valid point, it seems like  trading one gotcha for another, and thus making people having learned  the first one not make learn another one, seems momentarily better to me.

I wish we  could somehow deal with both gotchas, but I don't see any obvious solution.

jhass:
Albeit I have to agree that this behavior is maybe better. How about we preserve the old behavior under a different syntax,  like passing a block?

oprypin:
The advantage from https://github.com/crystal-lang/crystal/pull/8977#issuecomment-606868377 can be easily replicated with 
`print "working..."; p! whatever`

asterite:
If you have p! and the execution never ends, you know it never ended because nothing got printed. If you are not sure about that, add a puts before the call. I think this change has a greater benefit than the current one. For Instance, I never benefited from the existing behavior but I always complain when I have nested p calls.

jhass:
> `print "working..."; p! whatever`

The fixed  behavior can be replicated likewise. I think avoiding to come up with an explicit message to print is the entire reason d'etre here. Any downsides from `p! { server.accept }`  preserving current behavior  while changing `p!(p!(1))` like proposed here?

asterite:
How can it be replicated likewise?

jhass:
```
result = expr
p("expr => #{result.inspect}")

asterite:
I think expected evaluation order is more intuitive...

jhass:
I'm  not arguing against that, I just see some value in the current behavior too. Hence I'm arguing  for promoting the  current behavior to the "uglier"/less discoverable syntax.

oprypin:
Well who is ever going to be like "I expect an endless loop somewhere so I want the prompt and the arrow to print earlier so I'm intentionally going to add curly brackets"

jhass:
Not expect, suspect!

PR #8978, opened 2020-03-31
Fix Windows codegen (LLVM globals) in non-singlemodule mode

oprypin:
(the suggestions for root causes and fixes all come from @RX14)

Windows compilation so far was always being done in --single-module mode because --cross-compile implies it. But if one adds enough workarounds to try out the compiler *on* Windows, they would run into these problems, unless --single-module is specified.

1. `__CxxFrameHandler3` is being added only in the main module but can be used in any other module, and the attempt would cause an error:

       Undefined llvm function: __CxxFrameHandler3

   Instead, define this function once per module (still on-demand).

   https://gitter.im/crystal-lang/crystal?at=5e7ce5b3bf65703264de4711

2. `void_ptr_type_descriptor` is being defined as an LLVM global once per module. But because it's the exact same symbol in all the modules, linking them together leads to this error (repeated per module):

       I-O-5858F-ileD-escriptor43.o : error LNK2005: "void *`RTTI Type Descriptor'" (??_R0PEAX@8) already defined in _main.o.

   Instead, define it only in the main module (still on-demand) and then let other modules refer to it as an external constant.

   https://gitter.im/crystal-lang/crystal?at=5e80a76fd71a6e7e8d7efa55

### DIFF
--- a/src/compiler/crystal/codegen/codegen.cr
+++ b/src/compiler/crystal/codegen/codegen.cr
@@ -183,9 +183,7 @@ module Crystal
 
       if @program.has_flag? "windows"
         @personality_name = "__CxxFrameHandler3"
-
-        personality_function = @llvm_mod.functions.add(@personality_name, [] of LLVM::Type, llvm_context.int32, true)
-        @main.personality_function = personality_function
+        @main.personality_function = windows_personality_fun
       else
         @personality_name = "__crystal_personality"
       end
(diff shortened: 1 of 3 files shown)


RX14:
> The other alternative would be to declare types in the crystal land and access their representation

It's actually impossible to represent these types as crystal constants, because we don't support emitting constant math expressions to LLVM (see `sub_image_base`, Crystal would codegen this as a method call).

PR #8980, opened 2020-04-01
Syntax: allow '?!' as a name suffix

### LINKED ISSUE #8979: [RFC] Add `?!` as a name suffix
# Context

Crystal, like Ruby, allows names to end with `?` and `!`.
Why? See, right there, I just used an `?`. We humans use them all the time.
Ruby using them makes Ruby more human. After all, Ruby is a programmer's best friend.
And we'd like Crystal to be as human and friendly as Ruby.

Be we can go a step forward.

# Enter `?!`

The `?!` symbol combination is used all the time too!

We can call it the "WTF" name suffix, but I'm open for suggestions. For example
it could be called [Interrobang](https://en.wikipedia.org/wiki/Interrobang
), and we could even allow using `‽` instead of the more verbose `?!`.

However, compared to `?` and `!`, its meaning is a bit more vague.
So here are some situations in which we can use it:

## Not understanding a piece of code

Sometimes when code reviewing we stumble upon a method we don't fully understand.
We could leave a `TODO: I don't know what this is doing` but that's too verbose.
Instead, simply add `?!` to the method name and everywhere where it's called.

```crystal
def sum?!(n,t)
  n*(1..t/n).to_a.sum
end

sum?!(...)
```

This informs everyone reading the code that it's not clear why we are doing what
we are doing.

## Being angry about code

Other times we stumble upon code we dislike. We think to ourselves "Why the heck
are we doing this?!" or "Why the heck do we need to do this?!". The `?!` suffix can be used in that situation:

```crystal
Model.fetch_from_soap_api?!
```

## Not being sure why we are doing something.

This is similar to "Not understanding a piece of code", but not quite. The former
means we are reading someone else's code and we don't understand it, or maybe our
past self wrote it and we don't understand it anymore.

But "Not being sure why we are doing something" refers to us not understanding why
we are writing the code we write, right now, at the present moment.

One such example is using `not_nil!` because a variable can be `nil` and we don't
understand why. Using `not_nil!` we are making a strong assertion, we say
"Hey, compiler, I know this is not nil. I know better than you". But sometimes
we use it without understanding why. In that case, we can use `not_nil?!`
(we will have to introduce it the version following the introduction of `?!`, or
we can use `compare_versions`, I have no preference). The implementation is really
simple:

```crystal
class Object
  # Similar to `not_nil!` but to be used in cases where one doesn't
  # quite know why they are calling `not_nil!`.
  def not_nil?!
    not_nil!
  end
end
```

## Other use cases?

I'm happy to hear other use cases you have in mind!

I think this will be a radical change in the language.
That's why I think this should be seriously considered for 1.0.

Thoughts?!

asterite:
Fixes #8979

I'll add support for [Interrobang](https://en.wikipedia.org/wiki/Interrobang) later today.

### DIFF
--- a/spec/compiler/lexer/lexer_spec.cr
+++ b/spec/compiler/lexer/lexer_spec.cr
@@ -150,12 +150,14 @@ describe "Lexer" do
                      :pointerof, :sizeof, :instance_sizeof, :offsetof, :as, :as?, :typeof, :for, :in,
                      :with, :self, :super, :private, :protected, :asm, :uninitialized, :nil?,
                      :annotation, :verbatim]
-  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "fooBar",
+  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "foo?!", "fooBar",
                    "❨╯°□°❩╯︵┻━┻"]
   it_lexes_idents ["def?", "if?", "else?", "elsif?", "end?", "true?", "false?", "class?", "while?",
                    "do?", "yield?", "return?", "unless?", "next?", "break?", "begin?"]
   it_lexes_idents ["def!", "if!", "else!", "elsif!", "end!", "true!", "false!", "class!", "while!",
                    "nil!", "do!", "yield!", "return!", "unless!", "next!", "break!", "begin!"]
+  it_lexes_idents ["def?!", "if?!", "else?!", "elsif?!", "end?!", "true?!", "false?!", "class?!", "while?!",
+                   "nil?!", "do?!", "yield?!", "return?!", "unless?!", "next?!", "break?!", "begin?!"]
   it_lexes_i32 ["1", ["0i32", "0"], ["1hello", "1"], "+1", "-1", "1234", "+1234", "-1234",
                 ["1.foo", "1"], ["1_000", "1000"], ["100_000", "100000"]]
   it_lexes_i64 [["1i64", "1"], ["1_i64", "1"], ["1i64hello", "1"], ["+1_i64", "+1"], ["-1_i64", "-1"]]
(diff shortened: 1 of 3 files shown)


wontruefree:
should there also be a test for `!?`

asterite:
I'm not sure about `!?`. Is that common? But we can vote, I'm not against it.

spec/compiler/lexer/lexer_spec.cr
```diff
@@ -150,12 +150,14 @@ describe "Lexer" do
                      :pointerof, :sizeof, :instance_sizeof, :offsetof, :as, :as?, :typeof, :for, :in,
                      :with, :self, :super, :private, :protected, :asm, :uninitialized, :nil?,
                      :annotation, :verbatim]
-  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "fooBar",
+  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "foo?!", "fooBar",
```
straight-shoota, line 153:
```suggestion
  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "april_fools?!", "fooBar",
```

spec/compiler/parser/parser_spec.cr
```diff
@@ -179,6 +180,7 @@ module Crystal
     # #4815
     assert_syntax_error "def foo!=; end", "unexpected token: !="
     assert_syntax_error "def foo?=(x); end", "unexpected token: ?"
+    assert_syntax_error "def foo?!=(x); end", "unexpected token: !="
```
straight-shoota, line 183: I'm might actually think this shouldn't be a syntax error. It's entirely possible you're unsure of something, but still want to assign a value to it. To avoid mistakes we could enforce adding an additional `!` at the end:
```cr
def foo?!=!(x); end
```

And you could call it like this:
```cr
foo?! = "foo" !

File: spec/compiler/lexer/lexer_spec.cr (hunk 1 above)
Comment (asterite) [reply spec/compiler/lexer/lexer_spec.cr:153]:
```suggestion
  it_lexes_idents ["ident", "something", "with_underscores", "with_1", "foo?", "bar!", "april_foos?!", "fooBar",
```

PR #8982, opened 2020-04-01
Install LLVM 10 for OSX CI

bcardiff:
LLVM 10 has been releases on homebrew. 

The osx ci is failing since `LLVM_CONFIG` environment variable in `test_darwin` points to a non-aliased llvm-config path. Either that path or the formula version needs to change when a new llvm formula version is released in homebrew for now.

### DIFF
--- a/bin/ci
+++ b/bin/ci
@@ -111,8 +111,8 @@ prepare_build() {
   on_osx curl -L https://github.com/crystal-lang/crystal/releases/download/0.33.0/crystal-0.33.0-1-darwin-x86_64.tar.gz -o ~/crystal.tar.gz
   on_osx 'pushd ~;gunzip -c ~/crystal.tar.gz | tar xopf -;mv crystal-0.33.0-1 crystal;popd'
 
-  on_osx brew install llvm@9 gmp libevent pcre openssl pkg-config
-  on_osx brew link --force llvm@9
+  on_osx brew install llvm@10 gmp libevent pcre openssl pkg-config
+  on_osx brew link --force llvm@10
   # Note: brew link --force might show:
   #   Warning: Refusing to link macOS-provided software: llvm
   #


bin/ci
```diff
@@ -111,8 +111,8 @@ prepare_build() {
   on_osx curl -L https://github.com/crystal-lang/crystal/releases/download/0.33.0/crystal-0.33.0-1-darwin-x86_64.tar.gz -o ~/crystal.tar.gz
   on_osx 'pushd ~;gunzip -c ~/crystal.tar.gz | tar xopf -;mv crystal-0.33.0-1 crystal;popd'
 
-  on_osx brew install llvm@9 gmp libevent pcre openssl pkg-config
-  on_osx brew link --force llvm@9
+  on_osx brew install llvm@10 gmp libevent pcre openssl pkg-config
+  on_osx brew link --force llvm@10
```
Sija, line 115: Shouldn't this be on par with the homebrew formula?

```suggestion
  on_osx brew install llvm gmp libevent pcre openssl pkg-config
  on_osx brew link --force llvm
```

bcardiff: I prefer to use the alias to a fixed version and bump it manually when we are ready. Otherwise until we have support for the next llvm version we would need to add the `@V` as  a fix.

Ideally the LLVM_CONFIG path should work on any alias picked in this line. But since we already support llvm@10, this was the shortest change.

yxhuvud: One thing I note is that the definition of LLVM_CONFIG_FINDER in `/Makefile` hasn't been updated for a while. That seems .. related. How do these things interact?

RX14:
Would be nice if we fixed the CI scripts to set `LLVM_CONFIG` correctly regardless of if it's an old formula or no.

PR #8989, opened 2020-04-01
Add `Regex#matches?` and `String#matches?`

### LINKED ISSUE #8950: Support match? : Bool for String and Regex
### Ruby
In Ruby, `String` and `Regexp` both have a `match?` method which returns `true` or `false` depending on whether there is a match or not, rather then MatchData.

This is quite useful as often all you what to know is _if_ there is a match or not, rather than the details.

So for example, if I want to know if all the letters in a content string are in caps (e.g. "yelling"), but not the details,
I'd just like to use `match?` and return `true` or `false`

e.g.

```
[6] pry(main)> 'THE dog IS NOT IN CAPS'.match?(/^[^a-z]*[A-Z][^a-z]*$/)
=> false
[7] pry(main)> 'THE CAT IS IN CAPS'.match?(/^[^a-z]*[A-Z][^a-z]*$/)
=> true
```

### Crystal
At the moment `String` and `Regex` only support `match : MatchData?`, not `match?`.

To achieve the same with  Crystal ("spare me the details - just tell me if it matches") *is* possible with a double - bang...

```
$ cat issue.cr 
puts !!"THE dog IS NOT IN CAPS".match(/^[^a-z]*[A-Z][^a-z]*$/)
puts !!"THE CAT IS IN CAPS".match(/^[^a-z]*[A-Z][^a-z]*$/)
$ crystal issue.cr 
false
true
```

...but I think it would be more intuitive (and readable) if there was a `match? : Bool`

MakeNowJust:
Fixed #8950

It is another version of `match` method, which does not allocate `MatchData` when matched, instead it returns `true` or `false`. Avoiding allocation, it is efficient in fact.

https://github.com/crystal-lang/crystal/issues/8950#issuecomment-607516911

### DIFF
--- a/src/regex.cr
+++ b/src/regex.cr
@@ -491,13 +491,13 @@ class Regex
   end
 
   # Match at character index. It behaves like `#match`, however it returns `Bool` value.
-  # It does not create `MatchData` and does not update `$~` also.
+  # It neither returns `MatchData` nor assigns it to the `$~` variable.
   #
   # ```
   # /foo/.matches?("bar") # => false
   # /foo/.matches?("foo") # => true
   #
-  # # `$~` is not set even if last matching is succeeded.
+  # # `$~` is not set even if last match succeeds.
   # $~ # raises Exception
   # ```
   def matches?(str, pos = 0, options = Regex::Options::None) : Bool
--- a/src/regex.cr
+++ b/src/regex.cr
@@ -509,7 +509,7 @@ class Regex
   end
 
   # Match at byte index. It behaves like `#match_at_byte_inbdex`, however it returns `Bool` value.
-  # It does not create `MatchData` and does not update `$~` also.
+  # It neither returns `MatchData` nor assigns it to the `$~` variable.
   def matches_at_byte_index?(str, byte_index = 0, options = Regex::Options::None) : Bool
     return false if byte_index > str.bytesize
 
--- a/src/regex.cr
+++ b/src/regex.cr
@@ -481,14 +481,36 @@ class Regex
 
     ovector_size = (@captures + 1) * 3
     ovector = Pointer(Int32).malloc(ovector_size)
-    ret = LibPCRE.exec(@re, @extra, str, str.bytesize, byte_index, (options | Options::NO_UTF8_CHECK), ovector, ovector_size)
-    if ret > 0
+    if internal_matches?(str, byte_index, options, ovector, ovector_size)
       match = MatchData.new(self, @re, str, byte_index, ovector, @captures)
     else
       match = nil
     end
 
     $~ = match
+  end
+
+  # Match at character index. It behaves like `#match`, however it does not create `MatchData`.
+  def matches?(str, pos = 0, options = Regex::Options::None) : Bool
+    if byte_index = str.char_index_to_byte_index(pos)
+      matches_at_byte_index?(str, byte_index, options)
+    else
+      false
+    end
+  end
+
+  # Match at byte index. It behaves like `#match`, however it does not create `MatchData`.
+  def matches_at_byte_index?(str, byte_index = 0, options = Regex::Options::None) : Bool
+    return false if byte_index > str.bytesize
+
+    internal_matches?(str, byte_index, options, nil, 0)
+  end
+
+  # Calls `pcre_exec` C function, and handles returning value.
+  private def internal_matches?(str, byte_index, options, ovector, ovector_size)
+    ret = LibPCRE.exec(@re, @extra, str, str.bytesize, byte_index, (options | Options::NO_UTF8_CHECK), ovector, ovector_size)
+    # TODO: when `ret < -1`, it means PCRE error. It should handle correctly.
+    ret >= 0
   end
 
   # Returns a `Hash` where the values are the names of capture groups and the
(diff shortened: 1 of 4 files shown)


src/regex.cr
```diff
@@ -491,6 +490,29 @@ class Regex
...
+  def matches_at_byte_index?(str, byte_index = 0, options = Regex::Options::None) : Bool
+    return false if byte_index > str.bytesize
+
+    internal_matches?(str, byte_index, options, nil, 0)
+  end
+
+  # Calls `pcre_exec` C function, and handles returning value.
+  private def internal_matches?(str, byte_index, options, ovector, ovector_size)
+    ret = LibPCRE.exec(@re, @extra, str, str.bytesize, byte_index, (options | Options::NO_UTF8_CHECK), ovector, ovector_size)
+    # TODO: when `ret < -1`, it means PCRE error. It should handle correctly.
```
MakeNowJust, line 512: Please keep TODO in this PR. Because, fixing this needs bigger diff than adding new feature IMO.

j8r:
It is not ideal to have both `#match` and `#matches?` - they can be easily confused with non-existing `#matches` and `match?`.

yxhuvud:
There should probably be a discussion about naming though. Ruby defines a similar `match?` method returning a boolean. That clashes with the pretty strong naming convention in Crystal to add `?` to the end of methods if they may return `nil`.

Sija:
> That clashes with the pretty strong naming convention in Crystal to add `?` to the end of methods if they may return `nil`.

@yxhuvud ... but only when you have a non-nilable method version, like `first` & `first?` vs `includes?`

MakeNowJust:
`#matches` is very unconventional in Crystal naming, so no one is confused with this naming.

`#match?` is... sure, it may exist. However we can notice that this naming looks odd. First, we know searching methods like `Enumerable#find`, `#bserarch` or `String#index` returns `nil` when needle is not found in haystack, even though method name is not ended with `?`. And, we expect `#match` should be such a method, then we cannot guess `#match?` behavior from its naming.

On the other hand, `#matches?` is clear in Crystal naming convention that it returns `Bool` value at least. Indeed, it is hard to notice this method does not create `MatchData` and it is efficient. However I don't know better name for this, and more descriptive name (like `#match_without_matchdata_creation`) seems not good to me.

As the above reason, I consider `#matches?` is reasonable naming.

j8r:
One thing, this method can be used in `Regex#==`.

MakeNowJust:
Documents of `String` and `Regex`'s `#match` and `#matches?` are updated. They are now included a simple example showing `$~` is not set after `#matches?`.

I think it is ready to merge. Please.

src/regex.cr
```diff
@@ -491,6 +490,39 @@ class Regex
     $~ = match
   end
 
+  # Match at character index. It behaves like `#match`, however it returns `Bool` value.
+  # It does not create `MatchData` and does not update `$~` also.
```
Sija, line 494:
```suggestion
  # It neither returns `MatchData` nor assigns it to the `$~` variable.
```

src/regex.cr
```diff
@@ -491,6 +490,39 @@ class Regex
...
+  #
+  # ```
+  # /foo/.matches?("bar") # => false
+  # /foo/.matches?("foo") # => true
+  #
+  # # `$~` is not set even if last matching is succeeded.
```
Sija, line 500:
```suggestion
  # # `$~` is not set even if last match succeeds.
```

src/regex.cr
```diff
@@ -491,6 +490,39 @@ class Regex
...
+  def matches?(str, pos = 0, options = Regex::Options::None) : Bool
+    if byte_index = str.char_index_to_byte_index(pos)
+      matches_at_byte_index?(str, byte_index, options)
+    else
+      false
+    end
+  end
+
+  # Match at byte index. It behaves like `#match_at_byte_inbdex`, however it returns `Bool` value.
+  # It does not create `MatchData` and does not update `$~` also.
```
Sija, line 512:
```suggestion
  # It neither returns `MatchData` nor assigns it to the `$~` variable.
```

src/string.cr
```diff
@@ -3834,12 +3834,35 @@ class String
   end
 
   # Finds match of *regex*, starting at *pos*.
+  # It updaate `$~` by its result.
```
Sija, line 3837:
```suggestion
  # It also updates `$~` with the result.
```

src/string.cr
```diff
@@ -3834,12 +3834,35 @@ class String
...
+  # $~                 # raises Exception
+  # ```
   def match(regex : Regex, pos = 0) : Regex::MatchData?
     match = regex.match self, pos
     $~ = match
     match
   end
 
+  # Finds match of *regex* like `#match`, but it returns `Bool` value.
+  # It does not create `MatchData` and does not update `$~` also.
```
Sija, line 3853:
```suggestion
  # It neither returns `MatchData` nor assigns it to the `$~` variable.
```

src/string.cr
```diff
@@ -3834,12 +3834,35 @@ class String
...
+  #
+  # ```
+  # "foo".matches?(/bar/) # => false
+  # "foo".matches?(/foo/) # => true
+  #
+  # # `$~` is not set even if last matching is succeeded.
```
Sija, line 3859:
```suggestion
  # # `$~` is not set even if last match succeeds.
```

MakeNowJust:
Thanks @Sija ❤️

PR #8990, opened 2020-04-02
Base class for exceptions, not handled by default

waj:
This was a feature that was requested many times and initially I was against it, but lately we found some use cases where it might come handy.

Until now any exception raised in the program could be handled with a `begin/rescue` without types. In other words, an `rescue` statement handles, by default, any subclass of `Exception` class.

With this PR, there is now a `Raisable` where `Exception` inherits from. The `rescue` statements without types still handle only `Exception` instances. So in order to handle other exceptions, an explicit type must be specified.

Some already existing exceptions extends now from `Raisable`: `TypeCastError`, `NotImplementedError` and `NilAssertionError`. There are exceptions that normally shouldn't be handled in any program and they necessarily indicate a bug in the source code. There is a new exception type `UnexpectedStateError`, also extending from `Raisable`, useful for unreachable or inconsistent sates. These exceptions are normally errors that cannot be just ignored and retried. They should either crash the program or be logged with a big red flag.

Although this is a breaking change, most programs should continue running normally without changes.

### DIFF
--- a/spec/compiler/codegen/cast_spec.cr
+++ b/spec/compiler/codegen/cast_spec.cr
@@ -47,8 +47,8 @@ describe "Code gen: cast" do
       begin
         a.as(Char)
         false
-      rescue ex
-        ex.message.not_nil!.includes?("cast from Int32 to Char failed") && (ex.class == TypeCastError)
+      rescue ex : TypeCastError
+        ex.message.not_nil!.includes?("cast from Int32 to Char failed")
       end
       )).to_b.should be_true
   end
@@ -71,8 +71,8 @@ describe "Code gen: cast" do
       begin
         a.as(Float64 | Char)
         false
-      rescue ex
-        ex.message.not_nil!.includes?("cast from Int32 to (Char | Float64) failed") && (ex.class == TypeCastError)
+      rescue ex : TypeCastError
+        ex.message.not_nil!.includes?("cast from Int32 to (Char | Float64) failed")
       end
       )).to_b.should be_true
   end
@@ -119,8 +119,8 @@ describe "Code gen: cast" do
       begin
         a.as(CastSpecBaz)
         false
-      rescue ex
-        ex.message.not_nil!.includes?("cast from CastSpecBar to CastSpecBaz failed") && (ex.class == TypeCastError)
+      rescue ex : TypeCastError
+        ex.message.not_nil!.includes?("cast from CastSpecBar to CastSpecBaz failed")
       end
       )).to_b.should be_true
   end
@@ -186,8 +186,8 @@ describe "Code gen: cast" do
       begin
         a.as(Nil)
         false
-      rescue ex
-        ex.message.not_nil!.includes?("cast from Reference to Nil failed") && (ex.class == TypeCastError)
+      rescue ex : TypeCastError
+        ex.message.not_nil!.includes?("cast from Reference to Nil failed")
       end
       )).to_b.should be_true
   end
--- a/spec/compiler/codegen/exception_spec.cr
+++ b/spec/compiler/codegen/exception_spec.cr
@@ -1319,4 +1319,24 @@ describe "Code gen: exception" do
       end
     ))
   end
+
+  it "default rescue doesn't handle Raisable" do
+    run(%(
+      require "prelude"
+
+      def foo
+        begin
+          begin
+            raise Raisable.new("foo")
+          rescue
+            1
+          end
+        rescue Raisable
+          2
+        end
+      end
+
+      foo
+      )).to_i.should eq(2)
+  end
 end
--- a/spec/compiler/codegen/special_vars_spec.cr
+++ b/spec/compiler/codegen/special_vars_spec.cr
@@ -29,7 +29,7 @@ describe "Codegen: special vars" do
 
         begin
           #{name}
-        rescue ex
+        rescue ex : NilAssertionError
           "ouch"
         end
         )).to_string.should eq("ouch")
@@ -49,7 +49,7 @@ describe "Codegen: special vars" do
 
         begin
           #{name}
-        rescue ex
+        rescue ex : NilAssertionError
           "ouch"
         end
         )).to_string.should eq("foo")
--- a/spec/compiler/normalize/select_spec.cr
+++ b/spec/compiler/normalize/select_spec.cr
@@ -10,7 +10,7 @@ describe "Normalize: case" do
       when 1
         baz
       else
-        ::raise("BUG: invalid select index")
+        ::raise(::UnexpectedStateError.new("BUG: invalid select index"))
       end
       CODE
   end
@@ -23,7 +23,7 @@ describe "Normalize: case" do
         x = __temp_2.as(typeof(foo))
         x + 1
       else
-        ::raise("BUG: invalid select index")
+        ::raise(::UnexpectedStateError.new("BUG: invalid select index"))
       end
       CODE
   end
@@ -48,7 +48,7 @@ describe "Normalize: case" do
         x = __temp_2.as(typeof(foo?))
         x + 1
       else
-        ::raise("BUG: invalid select index")
+        ::raise(::UnexpectedStateError.new("BUG: invalid select index"))
       end
       CODE
   end
@@ -61,7 +61,7 @@ describe "Normalize: case" do
         x = __temp_2.as(typeof(foo!))
         x + 1
       else
-        ::raise("BUG: invalid select index")
+        ::raise(::UnexpectedStateError.new("BUG: invalid select index"))
       end
       CODE
   end
--- a/spec/compiler/semantic/exception_spec.cr
+++ b/spec/compiler/semantic/exception_spec.cr
@@ -144,12 +144,23 @@ describe "Semantic: exception" do
       ") { nilable int32 }
   end
 
-  it "errors if catched exception is not a subclass of Exception" do
-    assert_error "begin; rescue ex : Int32; end", "Int32 is not a subclass of Exception"
+  it "errors if catched exception is not a subclass of Raisable" do
+    assert_error "begin; rescue ex : Int32; end", "Int32 is not a subclass of Exception (or Raisable)"
   end
 
-  it "errors if catched exception is not a subclass of Exception without var" do
-    assert_error "begin; rescue Int32; end", "Int32 is not a subclass of Exception"
+  it "errors if catched exception is not a subclass of Raisable without var" do
+    assert_error "begin; rescue Int32; end", "Int32 is not a subclass of Exception (or Raisable)"
+  end
+
+  it "allows rescue from Raisable exception" do
+    assert_type("
+      a = nil
+      begin
+      rescue ex : Raisable
+        a = ex
+      end
+      a
+    ") { union_of(nil_type, raisable.virtual_type) }
   end
 
   assert_syntax_error "begin; rescue ex; rescue ex : Foo; end; ex",
--- a/spec/std/string_scanner_spec.cr
+++ b/spec/std/string_scanner_spec.cr
@@ -146,7 +146,7 @@ describe StringScanner, "#[]" do
     s = StringScanner.new("Fri Dec 12 1975 14:39")
     s.scan(/this is not there/)
 
-    expect_raises(Exception, "Nil assertion failed") { s[0] }
+    expect_raises(IndexError) { s[0] }
   end
 
   it "raises when there is no subgroup" do
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -147,11 +147,11 @@ describe "String" do
 
     it "gets with a string" do
       "FooBar"["Bar"].should eq "Bar"
-      expect_raises(Exception, "Nil assertion failed") { "FooBar"["Baz"] }
+      expect_raises(ArgumentError, "Value not found in the string") { "FooBar"["Baz"] }
     end
 
     it "gets with a char" do
-      expect_raises(Exception, "Nil assertion failed") { "foo/bar"['-'] }
+      expect_raises(ArgumentError, "Value not found in the string") { "foo/bar"['-'] }
     end
   end
 
@@ -163,7 +163,7 @@ describe "String" do
 
     it "gets with a char" do
       "foo/bar"['/'].should eq '/'
-      expect_raises(Exception, "Nil assertion failed") { "foo/bar"['-'] }
+      expect_raises(ArgumentError, "Value not found in the string") { "foo/bar"['-'] }
       "foo/bar"['/']?.should eq '/'
       "foo/bar"['-']?.should be_nil
     end
--- a/src/channel.cr
+++ b/src/channel.cr
@@ -49,7 +49,7 @@ class Channel(T)
     # wait_result_impl with the right type and Channel.select_impl
     # to allow dispatching over unions that will not happen
     def wait_result(context : SelectContext)
-      raise "BUG: Unexpected call to #{typeof(self)}#wait_result(context : #{typeof(context)})"
+      raise UnexpectedStateError.new "BUG: Unexpected call to #{typeof(self)}#wait_result(context : #{typeof(context)})"
     end
 
     def wait_result(context : SelectContext(S))
@@ -58,7 +58,7 @@ class Channel(T)
 
     # idem wait_result/wait_result_impl
     def unwait(context : SelectContext)
-      raise "BUG: Unexpected call to #{typeof(self)}#unwait(context : #{typeof(context)})"
+      raise UnexpectedStateError.new "BUG: Unexpected call to #{typeof(self)}#unwait(context : #{typeof(context)})"
     end
 
     def unwait(context : SelectContext(S))
@@ -232,7 +232,7 @@ class Channel(T)
       when DeliveryState::Closed
         raise ClosedError.new
       else
-        raise "BUG: Fiber was awaken without channel delivery state set"
+        raise UnexpectedStateError.new "BUG: Fiber was awaken without channel delivery state set"
       end
     end
 
@@ -291,7 +291,7 @@ class Channel(T)
     case state
     when DeliveryState::Delivered
       @lock.unlock
-      raise "BUG: Unexpected UseDefault value for delivered receive" if value.is_a?(UseDefault)
+      raise UnexpectedStateError.new "BUG: Unexpected UseDefault value for delivered receive" if value.is_a?(UseDefault)
       value
     when DeliveryState::Closed
       @lock.unlock
@@ -309,7 +309,7 @@ class Channel(T)
       when DeliveryState::Closed
         yield
       else
-        raise "BUG: Fiber was awaken without channel delivery state set"
+        raise UnexpectedStateError.new "BUG: Fiber was awaken without channel delivery state set"
       end
     end
   end
@@ -397,7 +397,7 @@ class Channel(T)
 
   def self.select(ops : Indexable(SelectAction))
     i, m = select_impl(ops, false)
-    raise "BUG: blocking select returned not ready status" if m.is_a?(NotReady)
+    raise UnexpectedStateError.new "BUG: blocking select returned not ready status" if m.is_a?(NotReady)
     return i, m
   end
 
@@ -470,7 +470,7 @@ class Channel(T)
       end
     end
 
-    raise "BUG: Fiber was awaken from select but no action was activated"
+    raise UnexpectedStateError.new "BUG: Fiber was awaken from select but no action was activated"
   end
 
   # :nodoc:
@@ -524,9 +524,9 @@ class Channel(T)
       when DeliveryState::Closed
         raise ClosedError.new
       when DeliveryState::None
-        raise "BUG: StrictReceiveAction.wait_result_impl called with DeliveryState::None"
+        raise UnexpectedStateError.new "BUG: StrictReceiveAction.wait_result_impl called with DeliveryState::None"
       else
-        raise "unreachable"
+        raise UnexpectedStateError.new "unreachable"
       end
     end
 
@@ -589,9 +589,9 @@ class Channel(T)
       when DeliveryState::Closed
         nil
       when DeliveryState::None
-        raise "BUG: LooseReceiveAction.wait_result_impl called with DeliveryState::None"
+        raise UnexpectedStateError.new "BUG: LooseReceiveAction.wait_result_impl called with DeliveryState::None"
       else
-        raise "unreachable"
+        raise UnexpectedStateError.new "unreachable"
       end
     end
 
@@ -649,9 +649,9 @@ class Channel(T)
       when DeliveryState::Closed
         raise ClosedError.new
       when DeliveryState::None
-        raise "BUG: SendAction.wait_result_impl called with DeliveryState::None"
+        raise UnexpectedStateError.new "BUG: SendAction.wait_result_impl called with DeliveryState::None"
       else
-        raise "unreachable"
+        raise UnexpectedStateError.new "unreachable"
       end
     end
 
--- a/src/compiler/crystal/codegen/exception.cr
+++ b/src/compiler/crystal/codegen/exception.cr
@@ -168,10 +168,25 @@ class Crystal::CodeGenVisitor
           # `this_rescue_block` if the exception type is covered by the type restriction, or
           # branches to `next_rescue_block` if the restriction doesn't match.
           this_rescue_block, next_rescue_block = new_blocks "this_rescue", "next_rescue"
-          if a_rescue_types = a_rescue.types
+
+          # If no rescue types are specified we catch any subclass of Exception.
+          # Otherwise, if only Raisable was specified, the check can be skipped
+          # because every exception inherits from it.
+          rescue_types =
+            if a_rescue_types = a_rescue.types
+              if a_rescue_types.size == 1 && a_rescue_types[0] == @program.raisable
+                nil
+              else
+                a_rescue_types.map &.type
+              end
+            else
+              [@program.exception]
+            end
+
+          if rescue_types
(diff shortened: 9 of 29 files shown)


RX14:
I really don't like this pattern. I don't see it in new languages. 

We've come a long way as a language without this feature and the only time I've seen it requested is "cancelling" a fiber. Can you expand on the usecases for this and why all these existing errors should suddenly change behaviour?

ysbaddaden:
Same here. I don't understand why these errors should suddenly not be rescueable like any other errors and crash the program, or require a special treatment for them to be handled. They're normal _runtime_ errors and don't put the program into an irrecoverable state.

Sija:
The only class of errors that IMO fits into this case are the `BUG: xxx` errors.

straight-shoota:
I'm not against `Raisable` per se but only when we've agreed on a valid use case for. I'm with @ysbaddaden that `TypeCastError`, `NotImplementedError` and `NilAssertionError` are just ordinary runtime errors and should be rescued by default. They are not different from any other exception type that currently exists. Every exception indicates a bug in the source code if it's not rescued in the right place.

The concept of `Raisable` is similar to `Throwable` in Java or `BaseException` in Python. They too are not caught by default. But looking at their exception hierarchy, the types that do not descend from the default `Exception` type are not like any runtime errors. In Java they're summarized in the [`Error`](https://docs.oracle.com/javase/10/docs/api/java/lang/Error.html) class, which is described as such:

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch. Most such errors are abnormal conditions. The ThreadDeath error, though a "normal" condition, is also a subclass of Error because most applications should not try to catch it.

Most of the errors originate form invalid state in the virtual machine/runtime environment. And there's `ThreadDeath` which is similar to a fiber cancellation exception.

In Python, the only non-`Exception` error types are `SystemExit` and `KeyboardInterrupt` which represent `exit()` and `SIGINT` and `GeneratorExit` which represents the end of a generator (similar to `Iterator::Stop`).

bcardiff:
As when doing `raise "BUG: .."` or `raise NotImplementedError.new`, the fact that a `v.not_nil!`, or `v.as(T)` fails is due to some unexpected path taken by the code.

If the user wants to deal with nil or optional casts there are other idiomatic ways to deal with them: `if v` / `if v.nil?` / `v.as?(T)`.

Usually if the `v.not_nil!`, or `v.as(T)` are used is because the user don't want to consider the alternative path and is dealing or doing workarounds on the type system. Even if they are runtime exceptions.

That is the motivation why I believe that `NilAssertionError` and `TypeCastError` are a good fit to inherit directly from `Raisable`.

RX14:
These are still typically transient errors, and not ones worth taking down a whole application vs just one request for. Most people aren't so careful in practice.

asterite:
I agree with this change, but in the original text of the issue there's:

> but lately we found some use cases where it might come handy

Could you describe those use cases?

I can imagine in some code there was:

```crystal
begin
  # ...
rescue
  # Something went wrong, handle this and move on
end
```

and what happened inside the `begin` block was actually a bug, like `not_nil!` triggering or a failing cast, and so the error was silently hidden.

But... if that's the case, `rescue` would probably log this, or report it somewhere (like Rollbar) so you will find out. If we make `rescue` not trap these exceptions they will bubble to the top and exit the program. Is it better to exit the program than to let it continue but still know the problem? (assuming you log errors somewhere).

bcardiff:
> These are still typically transient errors

I don't agree. The issues originate from invalid assumptions of the user of how the code behaves. It will push the users to use `as?`, `.nil?` leading to more idiomatic and faster code insted of `v.as?(Int32) || 0` vs `v.as(Int32) rescue 0` for example.

> whole application vs just one request for.

The request is handled by a separate fiber as before. The changes in fiber catches Raisable and log it as an unhandled exception.

---

> If we make rescue not trap these exceptions they will bubble to the top and exit the program. Is it better to exit the program than to let it continue but still know the problem? (assuming you log errors somewhere).

I think a generic exception logger should log all exception, so it should rescue from Raisable. Why would you not raise an NotImplementedError or UnexpectedStateError?

asterite:
> Why would you not raise an NotImplementedError or UnexpectedStateError?

It's not about raising these or not. It's about what do you do when they happen. Do you want to abort the program, or continue the program but log the error?

Though I guess the question depends on how the code is structured. If `Raisable` won't be rescued by default then most users will probably `rescue Raisable` in very specific places to log these abnormal exceptions and move on. So it's probably okay to introduce `Raisable`.

waj:
Woah! I'm having the feeling that you guys don't like this feature 😄. I knew also that if I posted this PR one day before it would be just filled with smiles and disregarded. 

I understand most of your concerns, and believe me, I was there when we designed the exceptions for Crystal and decided that a `rescue` would catch ALL errors without exceptions (pun intended). We did it with the "let it fail" mantra that works really well in Erlang for example. Instead of constantly trying to rescue exceptions to "solve" the problem, just handle them to gracefully exit to some point where either the task is retried or just logged. However that's not a general rule. Sometimes we want to handle as much as possible within a smaller scope, and activate some alternate logic. But that would work if the state is still consistent and within an expected set of possible values. This is when an error raised from a type cast (both `as` and `not_nil!` are) represents a conflict between what the programmer had in mind and the reality. The current state is not anymore within the limits that were expected to be, and the task shouldn't be retried or continued with that state. Is not that we don't want to allow these errors to be handled (like in Swift). Just that maybe it's a good idea not handling them by default.

Let me clarify or debunk some of your statements:

@RX14:

> I don't see it in new languages

Please, let me know which languages are you evaluating. In Swift for example, type casting with `as!` or dereferencing a nilable value with `!`, both raises runtime errors that cannot be handled by a `catch`. I think that's too extreme. Adding a `rescue Raisable` would let you catch them without crashing the program.

> We've come a long way as a language without this feature

Well... we've come a long way with too many things :)
That not necessarily means it's right or the best way. Let's stay open minded for changes all the time.

> These are still typically transient errors, and not ones worth taking down a whole application vs just one request for. Most people aren't so careful in practice.

I'm not saying we must take the whole application down! I'm not transforming the exceptions into fatal errors like other languages do. Just making sure these errors are handled in the right place. Most people actually won't notice any change at all if they use `HTTP::Server` for example.

@ysbaddaden:

>  I don't understand why these errors should suddenly not be rescueable like any other errors and crash the program

They don't. I don't want programs crashing! In most cases, let's say a web server, these errors will still be handled in the same place and logged. That's all.

> They're normal runtime errors and don't put the program into an irrecoverable state.

It depends, like I explained above.

@straight-shoota:

> TypeCastError, NotImplementedError and NilAssertionError are just ordinary runtime errors and should be rescued by default. They are not different from any other exception type that currently exists.

I don't see it the same way. Those errors could have been handled within the code without understanding of any side effects or anything outside the world of the program. They are not the same as a missing file for example. In an ideal world, these errors could have been handled by a compiler with infinite intelligence and an impossible type checking algorithm.

straight-shoota:
> It will push the users to use as?, .nil? leading to more idiomatic and faster code

Maybe. But it could also push users to `rescue Raisable` everywhere, completely counteracting the original intention. Anyways, I don't think this is an appropriate means to educate users on efficient language use.

> > TypeCastError, NotImplementedError and NilAssertionError are just ordinary runtime errors and should be rescued by default. They are not different from any other exception type that currently exists.
>  I don't see it the same way. Those errors could have been handled within the code without understanding of any side effects or anything outside the world of the program.

Could they, really? A `NotImplementedError` for example is only predictable through reconstructing the exact conditions under which a feature is implemented. We don't want to enforce users to introduce such logic at call site, thus we made it a runtime error instead of a compile time error. When you say that could be handled within the program, it implies installing guards for non-implemented methods (or parts of methods) to avoid calling them.

Again using the example of Java's, this is the definition of the `Exception` class:

> The class Exception and its subclasses are a form of Throwable that indicates conditions that a reasonable application might want to catch. 

IMO "conditions that a reasonable application might want to catch" fits for any of the mentioned Crystal exceptions.

RX14:
I think it comes down to your belief in user behaviour. The argument seems to be centred around this:

> It will push the users to use `as?`, `.nil?` leading to more idiomatic and faster code insted of `v.as?(Int32) || 0` vs `v.as(Int32) rescue 0` for example.

I think this has far too much faith in people not to be lazy. I think that instead people will do the same thing they already have done, and sometimes use these methods for input validation. Because it's easy and simple and they already do it. So in effect this change will end up changing very few people's behaviour and just annoying the users.

To guide users into changing behaviour you have to remind them when they're writing or designing code or at compile time. Edge-case error behaviour does not happen often enough to cause an effect on user behaviour.

waj:
People, I think the conversation diverges too much into beliefs, behaviour of developers and programs crashing.

To be clear, I don't want programs to crash, where the program is the OS process. With this change the program behaviour will not change for 99.99% cases. I just want to avoid unnoticed handling of errors where they shouldn't. The "let it crash" lemma of Erlang applies only to agents or sub-components of the program as a whole. And 

For example, it's quite common to write something like `foo() rescue nil` to have a default value in case something (that it's likely to fail) fails. But it's very unlikely that when I write that program I really want to fallback to a default value when actually there was a programming mistake.

Maybe it's impossible to draw a precise line on which exceptions should be handled by default and which ones not. But the rationale so far was to consider those that have an intimate relationship with the type system or those that stands for a condition that was (initially) unexpected to happen.

This is also a big dilemma for me, and for really long time I was on the same side of the opinion as most of you. Please help me understand the pros and cons with constructive challenges against the arguments and positive feedback where possible. Considering both the cases where the new behaviour would be undesirable or unexpected but also scenarios where it actually makes sense and makes the language more helpful with the goal of writing resilient applications.

asterite:
My gut feeling tells me this is the right decision. It's done like that in Ruby, Java, C# and Python. And I think it makes sense not to hide those errors by default when doing `rescue`.

Could we maybe try it for 0.34.0 and see if it's disruptive? My guess is that everything will be fine and, at most, some bugs will be uncovered and handled in a better way.

straight-shoota:
@asterite 
> My gut feeling tells me this is the right decision. It's done like that in Ruby, Java, C# and Python. And I think it makes sense not to hide those errors by default when doing rescue.

I don't follow. You're argument refers other languages but they explicitly do NOT do it like this propsal. I'm not familiar with C# but Java, Python and Ruby rescue by default all regular exceptions that would come from user code (including null pointer and cast exceptions). They don't rescue exceptions by default that are caused by the runtime environment, for example to unwind the stack, handle signals and VM errors.

@waj 
> But it's very unlikely that when I write that program I really want to fallback to a default value when actually there was a programming mistake.

I don't think a hard distinction makes sense. Some exceptions are more indicative of a programming mistake than others, but in fact *any* exception can be caused by a programming mistake (and many of them are usually not expected to happen). When it's not caught in an appropriate place, it is definitely a programming mistake when it bubbles out of the originating context.
I fully agree that "it's impossible to draw a precise line". So I'd question why try to draw a line in the first place? IMO it's much simpler to treat all exceptions equal (except for those that would be out of scope like a potential `FiberCancelledException`).

In the end, it comes down to the question what's `rescue` (without arguments) supposed to mean?

asterite:
Should we maybe remove `exp rescue ...` from the language? That way at least one can't add trailing rescue to code that easily.

asterite:
@straight-shoota You are right. It seems in Ruby things like SyntaxError, LoadError and NotImplementedError are things that aren't rescued by default. Probably things that will cause a program to always fail (there's no sane way to recover from a syntax error or a missing file to require).

I wonder if distinguishing between Exception and Raisable only makes sense for languages with a VM then? What are statically typed, compiled languages, without a VM, with an exception hierarchy?

...

I remembered D. They also make this distinctions. Some examples of non-catchable-by-default errors:
- AssertError (when an assertion fails)
- FinalizeError (something goes wrong in a finalizer, I guess)
- OutOfMemoryError
- InvalidMemoryOpreationError
- SwitchError (no switch branch matches)

So I guess in our case we'd still like the Raisable distinction but we need to decide what to include in it? I don't know why AssertError would be fine to include, but not stuff like `not_nil!`, `as` failing, etc., since these are assertions after all...

RX14:
My gut feeling says this is wrong and even if it wasn't it's not worth the subtle breaking change for.

Cancelling fibers? Maybe that's worth it, but I wouldn't like to approve any of these exceptions as uncatchable.

yxhuvud:
Segfault might be another candidate, FWIW.

waj:
I found this post a friend of mine wrote long time ago about design of exceptions: https://forum.dlang.org/post/sehvjdrvyzfvmxhrgoyn@forum.dlang.org

I don't agree 100% but I find interesting the classification into "recoverable" and "fatal".

To be honest, since early days of Crystal I considered the option to not having things like `not_nil!` and `as`. Hence one would be forced to check the type upfront and raise a "fatal" error as last resort. So, that exception is not meant to be handled. It means there was a wrong assumption. Again, that doesn't necessarily mean the program crashes. It might be just that request, job or whatever unit of work.

But for convenience we have these exceptions in the language. Still the semantics for the raised exception is the same.

@asterite @RX14 it's not about gut feelings... let's try to keep a sane conversation with real arguments

waj:
Now that I think it more deeply I guess these exceptions should also be Raisable: ArgumentError, IndexError, KeyError, DivisionByZeroError, OverflowError. In fact these errors are unchecked exceptions in Java. We shouldn't write code relying on specific handling of these errors and there are ways to avoid all of them just by changing code. Also they are fatal errors because there was a precondition violation.

asterite:
> In fact these errors are unchecked exceptions in Java

Where is that?

RX14:
lets just make everything raisable lol

waj:
@asterite:
  * ArgumentError -> [IllegalArgumentException](https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html)
  * IndexError -> [IndexOutOfBoundsException](https://docs.oracle.com/javase/7/docs/api/java/lang/IndexOutOfBoundsException.html)
  * KeyError -> none? Dictionary returns null if the key was not found
  * DivisionByZeroError -> [ArithmeticException](https://docs.oracle.com/javase/7/docs/api/java/lang/ArithmeticException.html)
  * OverflowError -> there is no overflow checking in Java?

@RX14: who said uncatchable? 🤔 `Raisable` is not catched by default but they still can be handled by doing `rescue ex : Raisable` or with a more specific type.

asterite:
@waj In Java those inherit from Runtime Exception so they are caught by default

waj:
@asterite well, there is no such thing as "by default" because Java doesn't allow a `catch` without types. I said they were "unchecked". It's not 1:1 relationship.

oprypin:
[Reference on Python](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)

<pre>
BaseException
 ├── SystemExit
 ├── KeyboardInterrupt
 ├── GeneratorExit
 └── Exception
      └── <i>everything else</i>
</pre>

Quoting @straight-shoota from before

> In Python, the only non-Exception error types are `SystemExit` and `KeyboardInterrupt` which represent `exit()` and `SIGINT` and `GeneratorExit` which represents the end of a generator (similar to `Iterator::Stop`).

The suggested exclusions *here* would not be similarly excluded in Python (`NotImplementedError` certainly isn't). These aren't merely "you shouldn't catch it", they are non-*error* events. Actually that points towards the idea there being completely opposite.

Moreover, `except:` in Python actually catches **everything**, unlike previously claimed in this discussion. But that's why it is widely recommended to always write `except Exception:` instead.

Also note that <code><b>*</b>able</code> is always a module in standard library, other than this change. I don't like the name `Raisable` overall.

Sija:
@waj Similarly, Ruby doesn't do that either ([reference](https://ruby-doc.org/core-2.5.0/Exception.html)):

```
Exception
├── NoMemoryError
├── ScriptError
    └── LoadError
    └── NotImplementedError
    └── SyntaxError
├── SecurityError
├── SignalException
    └── Interrupt
├── StandardError (default for `rescue`)
    └── ArgumentError
        └── UncaughtThrowError
    └── EncodingError
    └── FiberError
    └── IOError
        └── EOFError
    └── IndexError
        └── KeyError
        └── StopIteration
    └── LocalJumpError
    └── NameError
        └── NoMethodError
    └── RangeError
        └── FloatDomainError
    └── RegexpError
    └── RuntimeError (default for `raise`)
    └── SystemCallError
        └── Errno::*
    └── ThreadError
    └── TypeError
    └── ZeroDivisionError
├── SystemExit
├── SystemStackError
└── fatal (impossible to rescue)
```

asterite:
@waj what does unchecked means, regarding Java exceptions?

RX14:
unchecked means you don't need to explicitly annotate them with `int foo() throws ExceptionType {...}` in the method signature and you're not forced to check for them.

No other language has this so it's a bit hard to compare

asterite:
oh, yeah, now I remember. Yeah, I don't know if it's related.

j8r:
Several languages have a notion of critical errors, or panic.
Here, 2 kind of exceptions could exist:
- Standard ones - rescued as usual.
- Critical ones - requires an explicit `rescue ex : CriticalError`. This way, the user acknowledge (s)he rescuing critical compiler/language errors, which can affect the safety of the program as a whole, and are not supposed to be recoverable.
For instance, a race condition related to fibers, and thread.

If the safety of the whole program can't be guaranteed anymore, it is fine to make let it crash by default.

waj:
@asterite I see the relationship because you’re not supposed to handle those exceptions. They are normally exceptions that are not reasonable to recover. If you make your code dependant on those exceptions there is probably a better way to do it without them (like checking array bounds, or argument ranges).

waj:
Can someone against this feature mention, with a link to real code if possible, where this change would affect the behavior negatively?

straight-shoota:
Well, for me it would have the result that I would have to write `rescue Raisable` everywhere I was using plain `rescue` before, because I usually don't want to miss any relevant exceptions in such a case and can't be bothered to recall which one's would be included in `rescue` without type restriction and which not.

This is a current code example from shardbox-core:
```diff
   def perform
     elapsed = Time.measure do
       import_catalog
     end

     @db.log_activity "import_catalog:finished", metadata: stats(elapsed)
-  rescue exc
+  rescue exc : Raisable
     begin
       # Log failure in a separate connection because the main transaction
       # has already failed and won't be committed.
       ShardsDB.transaction do |db|
         db.log_activity "import_catalog:failed", metadata: {"exception" => exc.class.to_s}
       end
-    rescue
+ rescue Raisable
     end

     raise exc
   end
```

ysbaddaden:
Don't get me wrong. I like the idea, and others probably do:

- Critical exceptions shouldn't be rescued (e.g. out of memory);
- We can raise an uncatchable exception yet still run ensure blocks (e.g. cancel fiber while still cleaning things).

But this proposal is awful: why put all those non critical exceptions is a call for writing blunt `rescue Raisable` statements, to avoid crashing the whole program when a single actor or request crashed because of a NilAssertionError, which defeats the whole purpose and benefit of this proposal.

waj:
@straight-shoota you say:
> I would have to write rescue Raisable everywhere I was using plain rescue before

but the idea of my question is "give examples of affected behaviour if you dont change the rescue". Changing the `rescue` statements everywhere rollbacks the effect.

Anyway, in your example, the first `rescue` looks like a top level error handler and yes, that statement would have to be changed to avoid the error keep bubbling up. Maybe the design could be adjusted to not have too many "top level" handlers. The second `rescue` is more controversial IMO because it could be hiding critical programming errors and they are just ignored. I know you do it because you don't want anything else to be affected by a bug during logging. But that would open an entirely new branch of this discussion. Again, centralizing that code in a single place could reduce the burden of rescuing `Raisable` everywhere.

Within the same codebase, at `ShardsDB#log_activity` there is this statement:
```crystal
repo_ref = get_repo(repo_id).ref rescue nil
```
That statement will return `nil` not only in response of transient errors, but also in presence of programming errors, like a `not_nil!` or a bad type cast.

@ysbaddaden I have huge respect for all of you and the fact that I'm the only one seeing something positive from this is telling me that I'm probably the one that it's wrong. But I'm still quite confident of my position so it might not be so easy yet to admit my mistake, if that's the case.

Again, you talk about "crashing the whole program when a single actor or request crashed" and of course I expect actors implementations or web servers to handle ALL errors at the right place. I just don't want an actor or web request to continue working as if nothing happened when there was an invalid assumption in the state of some object. 

Maybe I'm more advocate to fast iterations of "fail, fix, improve" to create more robust programs. I'm not saying we should abandon resiliency. But I've seen whole complex systems fall down in Erlang, even with all the "let it crash and restart" machinery, because someone decided to hide programming errors within "catch all" statements. With a fully strict functional language (read Haskell) 99% of these errors doesn't even exist at runtime. But we're not so lucky and I see those kind of errors quite "fatal" and I don't want to be hidden for long time in my programs. That's the whole motivation of these changes. Maybe in that quest I'm affecting the language in ways it shouldn't. Let's see.

An alternative I'm thinking after going through real code is to define the default of a `rescue` depending on wether the exception is received on a variable or not. The rationale is, if I declare a variable I'm very likely doing something useful with it (logging, re-raising within another more contextual exception, etc.). Otherwise, if I'm just trying to provide a sane default state when something fails, avoid hiding those programming errors that has nothing to do with transient errors. In other words, if I write `rescue ex` the type would be `Raisable`; otherwise if I write just `rescue` the type is `Exception`, This would fix for example all the `rescue nil` cases.

RX14:
For me the thought process goes like this:

If the `not_nil!` or `.as` failure happened unconditionally, or on an expected case, then it would get caught by test test suite or testing, so the most likely scenario is that you're encountering an unexpected, rare case in production.

All of the `rescue`s without a specific exception specified that I write are like the ones in `HTTP::Server`: they're there so that the application can recover from a transient error and hopefully keep running. The exception is logged and handled later. These are the cases where you want to catch everything.

Perhaps postfix rescue without logging anything is more common than I think, but there only seems to be 7 cases of `rescue` being used without capturing `ex` outside the crystal repo in all my cloned shards.

RX14:
Actually, I did it once :P 

```
@connection << "zEND\0" rescue nil # This can fail if the connection was closed by clamd
```

I think the only major uses of postfix rescue is.... crystal's stdlib itself.

asterite:
> Now that I think it more deeply I guess these exceptions should also be Raisable: ArgumentError, IndexError, KeyError, DivisionByZeroError, OverflowError

Let's say we code something like this:

```crystal
index = some_string.index(some_char)
index.not_nil!
```

We use `not_nil!` because we expect `some_char` to be there because we assume `some_char` came from a same place, or some other assumptions hold.

But then a user gives bad input, or the file we were reading from changed. Then this `not_nil!` won't be rescued, for example in some parsing logic.

Yes, there's a bug in the code and if there's a bug we must fix the code to handle that. But aren't most bugs fixed by changing code? In the case above, maybe there's nothing we can do on a `not_nil!` and we'll have to change it to:

```crystal
index || raise RescueableException.new("...")
```

because `not_nil!` is not rescueable by default.

My point is, all the other languages use very specific exception types for non-rescueable-by-default things: out of memory, stack overflow, syntax error. Basically, things you probably can't recover from. But one instance of `not_nil!` because of bad user input, or same scenario for an invalid cast, should be rescueable exceptions. (in my opinion)

straight-shoota:
> That statement will return nil not only in response of transient errors, but also in presence of programming errors, like a not_nil! or a bad type cast.

I actually like some aspects of checked exceptions in Java: checked exceptions are part of the API and signal that an intended action could not be completed.
You're expected to deal with such exceptions at call site, either by catching them or delegating them upwards.
In Crystal the use case of many exception types are similar, but there's nothing to enforce rescue like in Java. We just document them in the API docs.

Unchecked errors on the other hand are exceptions that can be raised anywhere and are not explicitly documented (or enforced to catch in Java).
I suppose most unchecked exceptions are indicative of programming errors, either internal to the implementation or because a feature is used wrongly. They could technically be avoided, but it's often hard and inefficient to cover really all bases.

If I understand you correctly, @waj, that also sketches what you mean by transient and programming errors?

I think we all agree that `NilAssertionError` and `TypeCastError` are programming errors. They can definitely be avoided in all cases because you can add checks at the respective places.
But what would happen when such a check triggers? You would probably just raise a different exception, because these errors usually inidicate some kind of invalid state and the algorithm can't continue as expected.
So instead of `NilAssertionError`, a different, probably some generic exception would be raised. Yay, we've eliminated a `NilAssertionError`, but it really doesn't change anything. Except maybe the error is communicated more explicitly with a descriptive error message and maybe exception type. But since it's a different exception type, it would probably inherit from `Exception` and be rescued by `rescue` when the original `NilAssertionError` wouldn't. To me that doesn't make sense to see a different behaviour, only because a specific exception was raised.
For complete compliance with the suggested approach, maybe the specific exception should be required to inherit from `Raisable` instead of `Exception` to keep the semantics. But honestly, I don't hink that would see much adaption.

waj:
@straight-shoota you are very close to my point. But "because these errors usually indicate some kind of invalid state and the algorithm can't continue as expected" these errors shouldn't be handled (besides some top level place to log and avoid crashing).

For the rules to create new exception types, I think I like the proposal from Juan Manuel: https://forum.dlang.org/post/sehvjdrvyzfvmxhrgoyn@forum.dlang.org

Anyway... I'm not planning to push this so much forward. At least for now. Thank you for all the feedback. I'll keep thinking how to make Crystal a stronger language without taking the fun out of it.

RX14:
I don't think it's reasonable to move forward on a PR which is basically a 50/50 among the core team.

ysbaddaden:
@waj Please open a RFC that exposes the problem along with examples, explain why it's a problem, compare to how Haskell does this, and how we could do better. Reading the PR description, I failed to grasp any of it. The examples and discussion eventually helped.

j8r:
Yes please a proper RFC issue instead of in-PR discussion.

PR #8992, opened 2020-04-02
add Location to HTTP::Request common header names

mamantoha:


mamantoha:
Mostly it used as "Location". But for example in GitHub response headers are all downcased.
I think it would be nice to have it.

PR #8995, opened 2020-04-03
Add to_safe and to_unsafe methods for a flexible JSON mapping serialization

grkek:
### Changes

- Added a custom `to_safe` method which simply skips out on the keys which have `hidden: true` defined in the extended type annotation.
- Added a custom `to_unsafe` method which redirects to the to_json method of the mapping, simply to avoid confusion and maintain better readability.

### Use cases

- When defining a mapping for an object most of the time we don't want some fields to go through and get serialized with the rest, for example: 

    1. Define a `password` field for an `User` object.
    2. Create an `User` instance and name it `user`.
    3. Serialize the `user` instance using the `user.to_json` function.
    4. As you can see the `password` field went through and got serialized with the rest.

- With this PR the mapping will be able to "safe serialize" the `User` object, for example:

    1. Define a `password` field for an `User` object like this: 
        ```crystal
        password: {type: String, hidden: true}
        ```
    2. Create an `User` instance and name it `user`.
    3. Serialize the `user` instance using the `user.to_safe` function.
    4. As you can see the `password` field didn't go through and got dropped from the rest of the fields.

### Additional thoughts

- This feature certainly can be considered as a non-crucial and dropped, but if you look at it from another perspective it is a really useful one, for NoSQL databases, model serialization, deserialization, etc.

### DIFF
--- a/src/json/mapping.cr
+++ b/src/json/mapping.cr
@@ -20,6 +20,13 @@ module JSON
   #   )
   # end
   #
+  # class SecretHouse
+  #   JSON.mapping(
+  #     address: String,
+  #     location: {type: Location, nilable: true, hidden: true},
+  #   )
+  # end
+  #
   # house = House.from_json(%({"address": "Crystal Road 1234", "location": {"lat": 12.3, "lng": 34.5}}))
   # house.address  # => "Crystal Road 1234"
   # house.location # => #<Location:0x10cd93d80 @lat=12.3, @lng=34.5>
@@ -28,6 +35,11 @@ module JSON
   # houses = Array(House).from_json(%([{"address": "Crystal Road 1234", "location": {"lat": 12.3, "lng": 34.5}}]))
   # houses.size    # => 1
   # houses.to_json # => %([{"address":"Crystal Road 1234","location":{"lat":12.3,"lng":34.5}}])
+  #
+  # secret_house = SecretHouse.from_json(%({"address": "Crystal Road 1234", "location": {"lat": 12.3, "lng": 34.5}}))
+  # secret_house.to_json   # => %({"address":"Crystal Road 1234","location":{"lat":12.3,"lng":34.5}})
+  # secret_house.to_safe   # => %({"address":"Crystal Road 1234"})
+  # secret_house.to_unsafe # => %({"address":"Crystal Road 1234","location":{"lat":12.3,"lng":34.5}})
   # ```
   #
   # ### Usage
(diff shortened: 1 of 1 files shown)


oprypin:
That's not what `to_unsafe` does. The name is basically reserved, to return a pointer.

Also I don't like the functionality itself and I bet there's a good way to achieve it without this addition.

grkek:
> That's not what `to_unsafe` does. The name is basically reserved, to return a pointer.

Shouldn't it throw an exception when I redefine a reserved name?

> Also I don't like the functionality itself and I bet there's a good way to achieve it without this addition.

I would like to see the better way to achieve it, since I am in need of this feature, I would be glad to see a better way.

jhass:
Yes this feels like a bit too much for stdlib, a shard such as https://github.com/Blacksmoke16/CrSerializer seems like a better place

straight-shoota:
> Shouldn't it throw an exception when I redefine a reserved name?

No. Defining `to_unsafe` methods is fine, but they should return a value's representation that can be used in C APIs.

RX14:
> I redefine a reserved name

It's not a reserved name, just a convention. Like defining a `to_s` method which doesn't return a string: possible, but will break many things.

Sija:
> I would like to see the better way to achieve it, since I am in need of this feature, I would be glad to see a better way.

1. Define a custom serializer within your project which outputs only wanted fields.
2. Expand `Hash/NamedTuple#to_json` to accept `:only` and `:except` keys allowing to white/blacklist certain properties - that's more tricky since you'd probably want to make it work on different levels, like `user.to_json(except: {"password", "address.street"})`

bew:
> > Shouldn't it throw an exception when I redefine a reserved name?
> 
> No. Defining `to_unsafe` methods is fine, but they should return a value's representation that can be used in C APIs.

Maybe the compiler could enforce that ?

asterite:
Hi! This is useful but there are other ways to do it. For example (assuming you want the names `to_safe` and `to_unsafe`):

```crystal
class SecretHouse
  property address : String
  proprety location : Location

  def to_unsafe
    {address: address, location: location}.to_json
  end

  def to_safe
    {address: address}.to_json
  end
end
```

The majority of the use cases don't need this "safe" vs. "unsafe" distinction, so we shouldn't make things more complex for a very specific use case.

There are other ways to model this: you could have `House` and `SecretHouse` with different proprties, and so on.

Blacksmoke16:
FWIW since I was mentioned, `CrSerializer` has support for this via the [Exclude](https://blacksmoke16.github.io/CrSerializer/CrSerializer/Annotations/Exclude.html) annotation.  It also supports deserializing the value but not serializing it via the [ReadOnly](https://blacksmoke16.github.io/CrSerializer/CrSerializer/Annotations/ReadOnly.html) annotation.

grkek:
> Hi! This is useful but there are other ways to do it. For example (assuming you want the names `to_safe` and `to_unsafe`):
> 
> ```crystal
> class SecretHouse
>   property address : String
>   proprety location : Location
> 
>   def to_unsafe
>     {address: address, location: location}.to_json
>   end
> 
>   def to_safe
>     {address: address}.to_json
>   end
> end
> ```
> 
> The majority of the use cases don't need this "safe" vs. "unsafe" distinction, so we shouldn't make things more complex for a very specific use case.
> 
> There are other ways to model this: you could have `House` and `SecretHouse` with different proprties, and so on.

Yes I know I can do it that way but the main reason why I did it this way was that the macro would automatically generate it for me and I wouldn't have to write it, just trying to keep the DRY principle. Thank you all for reviewing my PR.

PR #8997, opened 2020-04-03
Add Indexable#reverse_map

caspiano:
Create `Indexable#reverse_map` which avoids an intermediate allocation.

Here's a [gist with benchmarks](https://gist.github.com/caspiano/9d2b20a6b49586592cd9f826ddb77722)

### DIFF
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -924,6 +924,12 @@ describe "Array" do
     a.should eq([2, 4, 6])
   end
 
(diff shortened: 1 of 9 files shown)


straight-shoota:
Resolves #8997

RX14:
If it's the ~same performance I don't see why we should add this

straight-shoota:
Anyways: Thank you for your  engagement @caspiano

PR #8999, opened 2020-04-03
Add `Enumerable#first` with fallback block

MakeNowJust:
Such a method exists in `Indexable`, but not in `Enumerable`.
It simplify `first`/`first?` implementation a bit.

### DIFF
--- a/src/enumerable.cr
+++ b/src/enumerable.cr
@@ -481,6 +481,11 @@ module Enumerable(T)
 
   # Returns the first element in the collection,
   # or the given block's value.
+  #
+  # ```
+  # ([1, 2, 3]).first { 4 }   # => 1
+  # ([] of Int32).first { 4 } # => 4
+  # ```
   def first
     each { |e| return e }
     yield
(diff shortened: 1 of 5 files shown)


MakeNowJust:
`Indexable#first` and `#first?` are removed. Then, `#first` raises `Enumerable#EmptyError` now. Is this breaking-change?

RX14:
Unifying enumerable and indexable exceptions sounds something worth considering

MakeNowJust:
Ah, I'll fix `Deque` spec to pass CI.

src/indexable.cr
```diff
@@ -454,16 +444,6 @@ module Indexable(T)
     size == 0 ? yield : unsafe_fetch(0)
```
bcardiff, line 444: Might be worth moving it the examples to enumerable. Even if the use arrays (and show one that raises `Enumerable::EmptyError`)

Then the `:inherited:` can be used. Or, maybe this `def first` can also go. Is not that the default implementation will be more inneficient.

MakeNowJust:
Docs are updated and specs are fixed.

src/enumerable.cr
```diff
@@ -479,18 +479,38 @@ module Enumerable(T)
     if_none
   end
 
+  # Returns the first element in the collection,
+  # or the given block's value.
```
straight-shoota, line 483:
```suggestion
  # If the collection is empty, calls the block and returns its value.
```

straight-shoota:
@bcardiff Alternatively move `IndexError` to `Indexable::IndexError`? Both errors are strongly related to the respective collection types.

MakeNowJust:
I think `Array#pop` and `#shift` can raise `EmptyError` instead `IndexError`.

But I'm not sure this discussion is scope of this PR.

PR #9000, opened 2020-04-03
Fix {Array, Deque, Hash}#== and #hash for recursive data

MakeNowJust:
`Reference#recursive_exec_pair` is added for `#==`, and `Reference#recursive_exec_outer` is added for `#hash`.

Now, `==` works for recursive collections and they can store to `Hash` as key, then this example works:

```crystal
alias Rec = Array(Rec)

a = [] of Rec
b = [] of Rec
c = [] of Rec

a << b
b << a
c << c << c

p! a == b # => true
p! a == c # => false

hash = {} of Rec => Int32
hash[a] = 1
hash[b] = 2
hash[c] = 3

p! hash # => {[[[...]]] => 2, [[...], [...]] => 3}
```

Note that such example works fine in Ruby also, so I think Crystal should support recursive collection equality and hashing.

Thank you.

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -163,12 +163,24 @@ class Array(T)
   # ary == [2, 3]    # => false
   # ```
   def ==(other : Array)
-    equals?(other) { |x, y| x == y }
+    exec_recursive_pair(:==, other) do
+      return false unless equals?(other) { |x, y| x == y }
+    end
+
+    true
   end
 
   # :nodoc:
   def ==(other)
     false
+  end
+
+  # See `Object#hash(hasher)`
+  def hash(hasher)
+    # TODO: it returns same hash value when an array is recursive.
+    # While `Hash` works fine, but it is inefficient.
+    exec_recursive_outer(:hash) { hasher = super(hasher) }
+    hasher
   end
 
   # Combined comparison operator.
--- a/src/reference.cr
+++ b/src/reference.cr
@@ -136,6 +136,76 @@ class Reference
   # :nodoc:
   module ExecRecursive
     alias Registry = Hash({UInt64, Symbol}, Bool)
+    alias Outers = Hash(Symbol, Bool)
+
+    class OuterSignal < Exception
+    end
+
+    {% if flag?(:preview_mt) %}
+      @@exec_recursive = Crystal::ThreadLocalValue(Registry).new
+      @@outers = Crystal::ThreadLocalValue(Outers).new
+    {% else %}
+      @@exec_recursive = Registry.new
+      @@outers = Outers.new
+    {% end %}
+
+    def self.hash
+      {% if flag?(:preview_mt) %}
+        @@exec_recursive.get { Registry.new }
+      {% else %}
+        @@exec_recursive
+      {% end %}
+    end
+
+    def self.outers
+      {% if flag?(:preview_mt) %}
+        @@outers.get { Outers.new }
+      {% else %}
+        @@outers
+      {% end %}
+    end
+  end
+
+  private def exec_recursive(method)
+    hash = ExecRecursive.hash
+    key = {object_id, method}
+    if hash[key]?
+      false
+    else
+      hash[key] = true
+      begin
+        yield
+      ensure
+        hash.delete(key)
+      end
+      true
+    end
+  end
+
+  private def exec_recursive_outer(method)
+    outers = ExecRecursive.outers
+    outmost = !outers[method]?
+
+    unless outmost
+      unless exec_recursive(method) { yield }
+        raise ExecRecursive::OuterSignal.new
+      end
+      return true
+    end
+
+    outers[method] = true
+    begin
+      exec_recursive(method) { yield }
+    rescue ExecRecursive::OuterSignal
+      return false
+    ensure
+      outers.delete method
+    end
+  end
+
+  # :nodoc:
+  module ExecRecursivePair
+    alias Registry = Hash({UInt64, UInt64, Symbol}, Bool)
 
     {% if flag?(:preview_mt) %}
       @@exec_recursive = Crystal::ThreadLocalValue(Registry).new
@@ -152,15 +222,19 @@ class Reference
     end
   end
 
-  private def exec_recursive(method)
-    hash = ExecRecursive.hash
-    key = {object_id, method}
+  private def exec_recursive_pair(method, pair)
+    hash = ExecRecursivePair.hash
+    pair_id = pair.object_id
+    key = {object_id, pair_id, method}
     if hash[key]?
       false
     else
       hash[key] = true
-      value = yield
-      hash.delete(key)
+      begin
+        yield
+      ensure
+        hash.delete key
+      end
       true
     end
   end
--- a/spec/std/array_spec.cr
+++ b/spec/std/array_spec.cr
@@ -1,6 +1,6 @@
 require "spec"
 
-private alias RecursiveArray = Array(RecursiveArray)
+private alias RecursiveArray = Array(RecursiveArray) | Int32
 
 private class BadSortingClass
   include Comparable(self)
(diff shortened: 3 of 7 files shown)


src/array.cr
```diff
@@ -163,14 +163,26 @@ class Array(T)
...
 
   # :nodoc:
   def ==(other)
     false
   end
 
+  # See `Object#hash(hasher)`
+  def hash(hasher)
+    # TODO: it returns same hash value when an array is recursive.
+    # While `Hash` works fine, but it is inefficient.
```
bcardiff, line 181: I don't follow what you would want to do different.

MakeNowJust: For example,

```crystal
alias Rec = Array(Rec)

a = [] of Rec
b = [] of Rec
c = [] of Rec

a << b
b << a
c << c << c
```

then, `a.hash == b.hash` and `a.hash == c.hash` even though `a == b` and `a != c`.

Avoiding hash collision, it is better that `a.hash != c.hash` if possible. However it is hard to me.

src/reference.cr
```diff
@@ -159,8 +173,68 @@ class Reference
...
+      begin
+        yield
+      ensure
+        hash.delete(key)
+      end
+      true
+    end
+  end
+
+  private def exec_recursive_outer(method)
```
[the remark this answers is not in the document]
MakeNowJust: It is variant of `exec_recursive`, which jumps to outmost call when recursion is detected.

One of usage of this is `hash` calculation. For example, let `a = [b]` and `b = [b]`, it should be `a == b` because they are same structure, and so `a.hash == b.hash`. However recursion is detected at `a == a[0][0]` on `a`, but detected at `b == b[0]` on `b`, different recursion point produces different `hash` value and it causes a problem on storing to `Hash`. `exec_recursive_outer` is needed to prevent this.

For now, `hash` calculation is only use-case. But this method is called in some places.

Name of `exec_recurse_outer` is derived from Ruby's one, and I don't know why this is named. ([here](https://github.com/ruby/ruby/blob/eaf06f457af92505fb718a8b711f4d825f399b69/thread.c#L5105-L5112))

I think `exec_recursive_shortcircuit` is better (?)

MakeNowJust:
I updated PR description to describe what is problem.

RX14:
I'm worried about performance impacts... It seems unfortunate that this introduces memory writes to all equality and hashing operations even when they're non-recursive...

waj:
Idea: it might be possible to detect at compile time if a type is recursive (maybe a macro function?). So this algorithm could be enabled only for recursive types.

MakeNowJust:
@waj It is hard or impossible because recursion can be indirect. For example:

```crystal
class Foo
  def initialize(@arr : Array(Foo))
  end

  property arr

  def_equals arr
end

a = Foo.new([] of Foo)
b = Foo.new([] of Foo)

a.arr << a
b.arr << b

p a == b
```

Although, I think possible simple case should be handled (for example only containing primitive type).

waj:
There is actually a recursive type check being used to avoid recursive structs: https://github.com/crystal-lang/crystal/blob/0.33.0/src/compiler/crystal/semantic/recursive_struct_checker.cr

Maybe that algorithm could be made generic to flag a recursive flag on each type.

asterite:
> It is hard or impossible because recursion can be indirect

Yes, the idea is to check for indirect recursion through pointers and other types.

The recursive struct check doesn't work in this case because it allows recursion through pointers. Here we want to consider a type recursive if you can reach it even through pointers.

waj:
@asterite Oh, that's true. Anyway, I wonder if this is actually possible because the check must be done while the macros are being expanded. But after some other macros are expanded there could be more and more types.

asterite:
Oh, right. It probably needs to be a primitive method. Then at runtime you would check if it's recursive or not, but that's a constant at that point.

MakeNowJust:
Important note: there is a case in which a class has recursive pointer but it is not used on equality check. However detecting it is really impossible.

MakeNowJust:
@asterite @waj We can write recursive type detection without any compiler extension. Crystal macro is powerful enough.

```crystal
macro recursive?(type)
  {% type = type.resolve %}

  {% stack = {} of TypeNode => Bool %}
  {% stack[type] = true %}
  {% rec = false %}
  {% for t, ok in stack %}
    {% us = [] of TypeNode %}
    {% for v in t.instance_vars %}
      {% us << v.type %}
    {% end %}
    {% for u in t.subclasses %}
      {% us << u %}
    {% end %}
    {% for u in t.union_types %}
      {% us << u %}
    {% end %}
    {% if t <= Tuple || t <= Pointer %}
      {% for u in t.type_vars %}
        {% us << u %}
      {% end %}
    {% end %}
    {% if t <= NamedTuple %}
      {% for k in t.keys %}
        {% us << t[k] %}
      {% end %}
    {% end %}
    {% for u in us %}
      {% if type < u %}
        {% rec = true %}
      {% end %}
      {% if !rec && !stack[u] %}
        {% stack[u] = true %}
      {% end %}
    {% end %}
  {% end %}

  {{ rec }}
end

alias Rec = Array(Rec)

def rec
  recursive?(Rec)
end

p rec # => true
```

I am not sure it should be used for this, but it works as expected.

MakeNowJust:
I'm curious this change makes compiling time slow, so I measure `make clean_crystal crystal stats=1` 10 times against `master` branch and this branch. These logs and result are [here](https://gist.github.com/MakeNowJust/bf733d70f752de7e9d17ca803a399c09).

Result:

  - Average compiling time of compiler on **`master`**: 45.3933s
  - Average compiling time of compiler on **`fix/recursive-eq-hash`**: 45.71s

To see these logs, this change increases `Semantic (main)` time, however it decreases `Codegen (crystal)` time. I think, the reason is this change reduces codegen size of `#to_s` and `#inspect` of non-recursive time.

Then, I conclude an effect to compiling performance by this change is little.

(To be honest, I thought this change makes compiling slow more, and I'm surprising this result.)

RX14:
I'd rather add a `TypeNode#recursive?` to the compiler personally...

MakeNowJust:
Adding a new method is no problem, and it improves compiling speed.
But I'm not sure such a method has general use-case. 

By the way, `TypeNode#recursive?` seems not good to me. Because this name is ambiguous. **I** feel this method only detects recursive alias, not detect recursive pointer reference. `#has_recursive_reference?` is better.

src/reference.cr
```diff
@@ -133,14 +133,50 @@ class Reference
     io << '>'
   end
 
+  # Returns whether this can be recursive or not.
+  # It calculates on compile-time, and it determines `exec_recursive` handling is needed really.
+  private def possibly_recursive?
```
asterite, line 138: Structs can also be recursive...

asterite:
This can't be done in Crystal code. It must be a primitive. Consider this:

```crystal
class Foo
end

class Bar(T) < Foo
  def initialize(@x : T)
  end
end

p Foo.new.possibly_recursive? # COMMENT THIS LINE

Bar(Foo).new(Foo.new)

p Foo.new.possibly_recursive?
```

If you comment the indicated line you'll get different results.

The reason is that `possibly_recursive?` is instantiated before or after `Bar(Foo)` comes into existence.

Maybe recursive checks like these shouldn't be in the standard library. It makes sense for `p` because you generally want to inspect the entire object for debugging, and it can (eventually) contain references to itself. But `hash` and `==` in Reference type usually are based on the object_id, so there's no possible recursion. And if there is, well, your program will hang or crash. But in my experience that never happened, because such comparisons or checks are already expensive or don't make sense that I try to avoid them.

Let's wait until this actually happens to someone before trying to fix it.

src/reference.cr
MakeNowJust: @asterite When structs are recursive, inner struct must be wrapped by reference type. So this is no problem.

asterite: Ah, you are right. I guess the check will be done for any referenced types, but it doesn't make sense to do it for a struct.

PR #9003, opened 2020-04-04
Regenerate implementation tool sample

nulty:
# Description

The comments in the samples directory for implementation cli tool require updating.

There is one difference that I don't understand which is the last output for the lorem property.
The old output refers to the object class but the new output is more cryptic.

### DIFF
--- a/samples/impl.cr
+++ b/samples/impl.cr
@@ -28,30 +28,29 @@ puts baz(Foo.new)
 
 Foo.new.lorem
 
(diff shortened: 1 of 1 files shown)


straight-shoota:
Thanks @nulty

bcardiff:
FWIW the last example is broken because the expansion of escaped macro seems to not be handling well the tracking of the `expanded_location`.

PR #9004, opened 2020-04-04
Compiler: fix autocasting in multidispatch

### LINKED ISSUE #8217: Bug dispatching enum shorthand to splat and non-splat overloads

asterite:
Fixes #8217

Let's see if this one works...

@RX14 Could you try your code with this PR? Also, any other case we should take a look? There were lots of conversations and snippets about this, I'm not sure which were the problematic cases...

### DIFF
--- a/spec/compiler/codegen/automatic_cast.cr
+++ b/spec/compiler/codegen/automatic_cast.cr
@@ -216,7 +216,7 @@ describe "Code gen: automatic cast" do
       )).to_i.should eq(10)
   end
 
-  it "does multidispatch with automatic casting (#8217)" do
+  it "does multidispatch with automatic casting (1) (#8217)" do
     run(%(
       def foo(mode : Int64, x : Int32)
         10
@@ -226,5 +226,17 @@ describe "Code gen: automatic cast" do
       end
       foo(1, 1 || "a")
       )).to_i.should eq(10)
+  end
+
+  it "does multidispatch with automatic casting (2) (#8217)" do
+    run(%(
+      def foo(mode : Int64, x : Int32)
+        10
+      end
+      def foo(mode : Int64, x : String)
+        20
+      end
+      foo(1, "a" || 1)
+      )).to_i.should eq(20)
   end
 end
--- a/spec/compiler/codegen/automatic_cast.cr
+++ b/spec/compiler/codegen/automatic_cast.cr
@@ -215,4 +215,16 @@ describe "Code gen: automatic cast" do
       Foo(Int64).new.x
       )).to_i.should eq(10)
   end
+
+  it "does multidispatch with automatic casting (#8217)" do
+    run(%(
+      def foo(mode : Int64, x : Int32)
+        10
+      end
+      def foo(mode : Int64, x : String)
+        20
+      end
+      foo(1, 1 || "a")
+      )).to_i.should eq(10)
+  end
 end
--- a/spec/compiler/semantic/automatic_cast.cr
+++ b/spec/compiler/semantic/automatic_cast.cr
@@ -468,4 +468,28 @@ describe "Semantic: automatic cast" do
       end
       )) { int64 }
   end
+
+  it "doesn't do multidispatch if an overload matches exactly (#8217)" do
+    assert_type(%(
+      def foo(x : Int64)
+        x
+      end
+
+      def foo(*xs : Int64)
+        xs
+      end
+
+      foo(1)
+      )) { int64 }
+  end
+
+  it "autocasts first argument and second matches without autocast" do
+    assert_type(%(
+      def fill(x : Float64, y : Int)
+        x
+      end
+
+      fill(0, 0)
+      )) { float64 }
+  end
 end
--- a/src/compiler/crystal/codegen/call.cr
+++ b/src/compiler/crystal/codegen/call.cr
@@ -360,7 +360,11 @@ class Crystal::CodeGenVisitor
           end
           node.args.each_with_index do |node_arg, i|
             a_def_arg = a_def.args[i]
-            result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            if autocast_literal?(node_arg)
+              # Matches, so nothing to do
+            else
+              result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            end
           end
 
           current_def_label, next_def_label = new_blocks "current_def", "next_def"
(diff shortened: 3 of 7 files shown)


src/compiler/crystal/codegen/call.cr
```diff
@@ -360,7 +360,11 @@ class Crystal::CodeGenVisitor
           end
           node.args.each_with_index do |node_arg, i|
             a_def_arg = a_def.args[i]
-            result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            if autocast_literal?(node_arg)
+              # Matches, so nothing to do
+            else
+              result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
+            end
```
Sija, line 367: Why not simplify it?

```suggestion
            unless autocast_literal?(node_arg)
              result = and(result, match_type_id(node_arg.type, a_def_arg.type, arg_type_ids[i]))
            end
```

asterite: Because I wanted to keep the comment?

Sija: I thought you could add it above `unless` but ok.

spec/compiler/codegen/automatic_cast.cr
```diff
@@ -215,4 +215,16 @@ describe "Code gen: automatic cast" do
...
+
+  it "does multidispatch with automatic casting (#8217)" do
+    run(%(
+      def foo(mode : Int64, x : Int32)
+        10
+      end
+      def foo(mode : Int64, x : String)
+        20
+      end
+      foo(1, 1 || "a")
```
RX14, line 227: Can we have a spec that `"a" || 1` gives 20?

asterite: Done!

src/compiler/crystal/semantic/call.cr
```diff
@@ -361,7 +361,19 @@ class Crystal::Call
...
+      # Now that we have all our matches, check if any of them matches exactly
+      # all types, assuming autocasted values will always match (because they
+      # matches and they were not ambiguous). If so, only keep matches up to
+      # that exact match. We need to do this here because with autocasting
+      # we consider all overloads to detect ambiguous usage.
+      stop_index = matches.index do |match|
+        signature.matches_exactly?(match, with_literals: true)
+      end
+      matches = matches[..stop_index] if stop_index
+    end
```
asterite, line 375: I think this logic shouldn't go here. I'll try to put it somewhere else where it makes more sense.

asterite: Or, maybe I'll leave a TODO to refactor it later...

asterite: Actually, this is fine. Moving it to some other place will make it much complex. The `with_literals` logic is mostly kept in this file, so this might be a good place to do this check.

In general, the entire "with_literals" logic might benefit from a rewrite, but this PR is not the place to do it.

spec/compiler/semantic/automatic_cast.cr
```diff
@@ -468,4 +468,28 @@ describe "Semantic: automatic cast" do
...
+    assert_type(%(
+      def foo(x : Int64)
+        x
+      end
+
+      def foo(*xs : Int64)
+        xs
+      end
+
+      foo(1)
```
bcardiff, line 482: how is this checking no multidispatch is done? It comparing positional vs named to me.

src/compiler/crystal/semantic/call.cr
```diff
@@ -361,7 +361,19 @@ class Crystal::Call
...
+    if with_literals
+      # Now that we have all our matches, check if any of them matches exactly
+      # all types, assuming autocasted values will always match (because they
+      # matches and they were not ambiguous). If so, only keep matches up to
+      # that exact match. We need to do this here because with autocasting
+      # we consider all overloads to detect ambiguous usage.
+      stop_index = matches.index do |match|
+        signature.matches_exactly?(match, with_literals: true)
+      end
+      matches = matches[..stop_index] if stop_index
```
bcardiff, line 374: If there is an exact match why the matches array does not contain a single element?

spec/compiler/semantic/automatic_cast.cr
asterite: Because on multidispatch the type would be `Int64 | Tuple(Int64)`.

src/compiler/crystal/semantic/call.cr
asterite: When autocasting is not considered, because there can't be an ambiguity, as soon as a match is an exact match there's no need to see other methods.

But with autocasting we need to check if there's ambiguity. For example:

```crystal
def fill(x : Int64, y : Int)
end

def fill(x : Int8, y : Int)
end

fill(0, 0)
```

That will match the first method, but also the second, and that's when you get the "ambiguous call" error.

If there are no ambiguities, it means all methods matched... but maybe one of them was an exact match? We need to do this after we considered all methods to check for ambiguity.

Ideally, `lookup_matches` would do this, but then we need to pass `with_literals` to those methods, and they are many. It's simpler to do it here.

Did this answer the question?

I think autocasting is a nice feature but it makes the method lookup very complex. Maybe we can remove it from 1.0?

asterite:
This seems to be working fine, so let's try it.

asterite:
Oh, we changed one setting so if there are more commits after an approval a new approval is needed.

PR #9006, opened 2020-04-05
Rename Process kill methods to signal [by @jan-zajic]

oprypin:
> Rename Process#kill(Signal) to Process#signal(Signal) that we can eventually implement on Windows (if needed) with a deprecation notice. The name kill is confusing as it can mean two things - send signal to process, or send signal kill (9) to process when argument omitted. But currently default behavior of kill method without providing signal argument is to send term signal to process (simulates posix behavior of kill command, but we need to remove POSIXisms of apis to become really portable)

Closes #8642. I'm re-creating the PR to be able to resolve outstanding comments.

This probably should be squash-merged, with the above message.

### DIFF
--- a/src/process.cr
+++ b/src/process.cr
@@ -44,7 +44,7 @@ class Process
     nil
   end
 
-  # Sends a *signal* to the process identified by the given *pid*.
+  # Sends *signal* to the process identified by *pid*.
   def self.signal(signal : Signal, pid : Int) : Nil
     ret = LibC.kill(pid, signal.value)
     raise RuntimeError.from_errno("kill") if ret < 0
--- a/src/process.cr
+++ b/src/process.cr
@@ -394,7 +394,7 @@ class Process
     close_io @error
   end
 
-  # Ask process for terminate gracefully
+  # Asks the process to terminate gracefully
   def terminate
     signal Signal::TERM
   end
--- a/src/process.cr
+++ b/src/process.cr
@@ -356,7 +356,7 @@ class Process
     signal sig
   end
 
-  # Sends *signal* to the process.
+  # Sends *signal* to this process.
   def signal(signal : Signal)
     Process.signal signal, @pid
   end
@@ -394,7 +394,7 @@ class Process
     close_io @error
   end
 
-  # Asks the process to terminate gracefully
+  # Asks this process to terminate gracefully
   def terminate
     signal Signal::TERM
   end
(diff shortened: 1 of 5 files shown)


src/process.cr
```diff
@@ -36,14 +36,20 @@ class Process
...
   def self.kill(signal : Signal, *pids : Int)
     pids.each do |pid|
-      ret = LibC.kill(pid, signal.value)
-      raise RuntimeError.from_errno("kill") if ret < 0
+      signal(signal, pid)
     end
     nil
   end
 
+  # Sends a *signal* to the process identified by the given *pid*.
```
straight-shoota, line 47:
```suggestion
  # Sends *signal* to the process identified by *pid*.
```

src/process.cr
```diff
@@ -345,8 +351,14 @@ class Process
   end
 
   # See also: `Process.kill`
+  @[Deprecated("Use #signal instead")]
   def kill(sig = Signal::TERM)
-    Process.kill sig, @pid
+    signal sig
+  end
+
+  # Sends *signal* to the process.
```
straight-shoota, line 359:
```suggestion
  # Sends *signal* to this process.
```

src/process.cr
```diff
@@ -382,6 +394,11 @@ class Process
     close_io @error
   end
 
+  # Asks the process to terminate gracefully
```
straight-shoota, line 397:
```suggestion
  # Asks this process to terminate gracefully
```

src/process.cr
oprypin: Isn't this just copying the documentation right above it?

straight-shoota: Yes, but that one's deprecated and will be removed =)
I consider the suggested wording strictly better. The method doesn't send "a" signal but the value of the `signal` argument (which is by convention expressed as italics). That also makes "the given" unnecessary.

straight-shoota:
@oprypin Why did you remove the example which included specs for `terminate`?

oprypin:
@straight-shoota , https://github.com/crystal-lang/crystal/pull/8642#discussion_r365191248

Funnily enough, that spec is exactly the same as the one preceding it, as kill and terminate are the same thing for now

straight-shoota:
I see. We could still run `#terminate` in a spec, though. Even if it's not possible / very difficult to test it behaves correctly. But we should make sure it compiles and runs.

oprypin:
(make sure to squash when merging)

PR #9007, opened 2020-04-05
Use "shards version" command in generated VERSION constant

Sija:
This removes 2nd place where the shard version constant is being defined - 1st being `version` value defined in `shard.yml` - reducing the surface for mistakes in this regard.


straight-shoota:
Duplicate of #5428

Sija:
I had a feeling there was a previous PR about that... fair 'nuff

Issue #3357, opened 2016-09-26
Handle subcommands (without dashes) in OptionParser


bew:
I'm trying to make a CLI app with subcommands like crystal init/build/run subcommands (without dashes), but not like crystal way ([crystal/command.cr#L65](https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/command.cr#L65)), but using OptionParser only (with nested parsers)
So I need to use (at least) 2 nested OptionParser, and options without dashes and options.

``` crystal
require "option_parser"

args = ["one", "subcommand", "two", "three"]

OptionParser.parse args do |parser|
  parser.on "subcommand", "message" do

    # args == ["two", "three"]
    OptionParser.parse args do |parser|
      parser.on "two ARG", "description" do |arg|
        # never go here
      end

      parser.on "verbose", "description" do
        # go here
      end
    end
  end
end
```

I'm on Crystal v0.18.7 on Archlinux

It seems that `parser.on "option VALUE"` never match, but `parser.on "--option VALUE"` will..
It it intentionnal ?

asterite:
Yes, OptionParser doesn't support commands and subcommands. The way to do it is to switch over the first element in ARGV, like what the compiler does. Alternatively there are some shards that provide this behaviour, for example [commander](https://github.com/mrrooijen/commander) (but I personally think switching over a string is really simple and works quite well).

Bew78LesellB:
Thanks for the link, but it seems too compllicated and not as readable as OptionParser..
As stated here : [OptionParse#on][OP#on] : 
> Flags can (but don't have to) start with a dash. They can also have an optional argument, which will get passed to the block.

Flags without dashes should be able to have an optionnal argument, but as told earlier, `parser.on("flag OPTION", "description", &block)` doesn't work..
I don't think it's the wanted behavior, is it ?

[OP#on]: https://crystal-lang.org/api/0.18.7/OptionParser.html#on%28flag%2Cdescription%2C%26block%3AString-%3E%29-instance-method

asterite:
I think it's pretty standard that command line options must start with a dash. Ruby's OptionParser also doesn't support what you are trying to do. We can probably raise an exception if the given string doesn't start with a dash...

chaniks:
I actually love the current implementation.  It is fully customisable.

However, as he showed, the document might be misleading.
> Flags can (but don't have to) start with a dash. ***"They"*** can also have an optional argument, which will get passed to the block.

[...]

PR #9009, opened 2020-04-05
OptionParser rewrite with subcommands

### LINKED ISSUE #8937: OptionParser skips `--` regardless of context
```crystal
require "option_parser"
 
OptionParser.parse(["-m", "--"]) do |parser|
  parser.on("-m MESSAGE", "Foo") { |message| p! message }
end
```
Expected:
```
message # => "--"
```
[Actual](https://carc.in/#/r/8rj8):
```
Unhandled exception: Missing option: -m (OptionParser::MissingOption)
```

---
This happens because the double-dash special option is checked for before doing any parsing
https://github.com/crystal-lang/crystal/blob/612825a53c831ce7d17368c8211342b199ca02ff/src/option_parser.cr#L205

But to do it correctly, the behavior must depend on context. If we're currently expecting an option value, the next token must be recognized as that regardless of its content.

E.g. Git recognizes this as an option correctly:
`git commit -m --` (create a commit with the message "--") 
and also accepts flags following it:
`git commit -m -- -a --` (add tracked files and create a commit with the message "--")

---
[For context](https://unix.stackexchange.com/a/452793),
In general the double dash can be used to indicate that the following arguments must not be parsed as flags.
```shell
search_text="--this"
grep -E -- "$search_text"
```

RX14:
I got angry with `OptionParser` for being a pile of bad code so I rewrote the implementation. It's simpler now, and easier to understand the implementation, and not `O(n^2)` so probably miles faster. And it has a completely compatible interface, no changes were required to any existing usages across the stdlib.

While I was rewriting it I had a universe brain moment and realised that I could make it handle subcommands very simply, so I threw that in as a bonus.

Example subcommand usage:
```cr
OptionParser.parse(args) do |opts|
  opts.on("subcommand", "Description") do
    opts.on("--foo arg", "Foo") { |v| foo = v }
    opts.on("--bar", "Bar") { bar = true }
    opts.on("-z", "--baz", "Baz") { z = true }
  end
  opts.on("--verbose", "") { verbose = true }
end
```

There's some tweaks still required to get the `to_s` output to not suck when using subcommands but I wanted to submit this for comment before I forgot about it tomorrow morning.

Fixes #8937.

### DIFF
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -101,8 +101,8 @@ describe "OptionParser" do
     expect_capture_option ["--flag=123"], "--flag=FLAG", "123"
   end
 
-  it "has required option with = (3) raises" do
-    expect_missing_option ["--flag="], "--flag=FLAG", "--flag"
+  it "has required option with = (3) handles empty" do
+    expect_capture_option ["--flag="], "--flag=FLAG", ""
   end
 
   it "raises if missing required argument separated from long flag" do
@@ -115,6 +115,14 @@ describe "OptionParser" do
 
   it "has required option with long flag space" do
     expect_capture_option ["--flag", "123"], "--flag ", "123"
+  end
+
+  it "gets short option with value -- (#8937)" do
+    expect_capture_option ["-f", "--"], "-f ARG", "--"
+  end
+
+  it "gets long option with value -- (#8937)" do
+    expect_capture_option ["--flag", "--"], "--flag [ARG]", "--"
   end
 
   it "doesn't raise if required option is not specified" do
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -542,4 +542,27 @@ describe "OptionParser" do
     bar.should be_false
     args.should eq(%w(--bar))
   end
+
+  it "can run a callback on every argument" do
+    args = %w(--foo file --bar)
+    foo = false
+    bar = false
+    OptionParser.parse(args) do |opts|
+      opts.on("--foo", "") { foo = true }
+      opts.on("--bar", "") { bar = true }
+      opts.before_each do |arg|
+        if arg == "file"
+          opts.stop
+        end
+      end
+      opts.unknown_args do |before, after|
+        before.should eq(%w(file))
+        after.should eq(%w(--bar))
+      end
+    end
+
+    foo.should be_true
+    bar.should be_false
+    args.should eq(%w(file --bar))
+  end
 end
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -470,6 +470,53 @@ describe "OptionParser" do
     z.should be_true
   end
 
+  it "parses with subcommands twice" do
+    args = %w(--verbose subcommand --foo 1 --bar sub2 -z)
+    verbose = false
+    subcommand = false
+    foo = nil
+    bar = false
+    sub2 = false
+    z = false
+
+    parser = OptionParser.new do |opts|
+      opts.on("subcommand", "") do
+        subcommand = true
+        opts.on("--foo arg", "") { |v| foo = v }
+        opts.on("--bar", "") { bar = true }
+        opts.on("sub2", "") { sub2 = true }
+      end
+      opts.on("--verbose", "") { verbose = true }
+      opts.on("-z", "--baz", "") { z = true }
+    end
+
+    parser.parse args
+
+    verbose.should be_true
+    subcommand.should be_true
+    foo.should be("1")
+    bar.should be_true
+    sub2.should be_true
+    z.should be_true
+
+    args = %w(--verbose subcommand --foo 1 --bar sub2 -z)
+    verbose = false
+    subcommand = false
+    foo = nil
+    bar = false
+    sub2 = false
+    z = false
+
+    parser.parse args
+
+    verbose.should be_true
+    subcommand.should be_true
+    foo.should be("1")
+    bar.should be_true
+    sub2.should be_true
+    z.should be_true
+  end
+
   it "unregisters subcommands on call" do
     foo = false
     bar = false
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -524,4 +524,22 @@ describe "OptionParser" do
           -f, --foo                        Foo
       USAGE
   end
+
+  it "stops when asked" do
+    args = %w(--foo --stop --bar)
+    foo = false
+    bar = false
+    OptionParser.parse(args) do |opts|
+      opts.on("--foo", "") { foo = true }
+      opts.on("--bar", "") { bar = true }
+      opts.on("--stop", "") { opts.stop }
+      opts.unknown_args do |before, after|
+        before.should eq(%w())
+        after.should eq(%w(--bar))
+      end
+    end
+    foo.should be_true
+    bar.should be_false
+    args.should eq(%w(--bar))
+  end
 end
--- a/spec/std/option_parser_spec.cr
+++ b/spec/std/option_parser_spec.cr
@@ -429,12 +429,8 @@ describe "OptionParser" do
     end
   end
 
-  it "raises if flag doesn't start with dash (#4001)" do
+  it "raises if flag pair doesn't start with dash (#4001)" do
     OptionParser.parse([] of String) do |opts|
-      expect_raises ArgumentError, %(Argument 'flag' ("foo") must start with a dash) do
-        opts.on("foo", "") { }
-      end
-
       expect_raises ArgumentError, %(Argument 'short_flag' ("foo") must start with a dash) do
         opts.on("foo", "bar", "baz") { }
       end
@@ -445,5 +441,87 @@ describe "OptionParser" do
 
       opts.on("", "-bar", "baz") { }
     end
+  end
+
+  it "handles subcommands" do
+    args = %w(--verbose subcommand --foo 1 --bar sub2 -z)
+    verbose = false
+    subcommand = false
+    foo = nil
+    bar = false
+    sub2 = false
+    z = false
+    OptionParser.parse(args) do |opts|
+      opts.on("subcommand", "") do
+        subcommand = true
+        opts.on("--foo arg", "") { |v| foo = v }
+        opts.on("--bar", "") { bar = true }
+        opts.on("sub2", "") { sub2 = true }
+      end
+      opts.on("--verbose", "") { verbose = true }
+      opts.on("-z", "--baz", "") { z = true }
+    end
+
+    verbose.should be_true
+    subcommand.should be_true
+    foo.should be("1")
+    bar.should be_true
+    sub2.should be_true
+    z.should be_true
+  end
+
+  it "unregisters subcommands on call" do
+    foo = false
+    bar = false
+    baz = false
+    OptionParser.parse(%w(foo baz)) do |opts|
+      opts.on("foo", "") do
+        foo = true
+        opts.on("bar", "") { bar = true }
+      end
+      opts.on("baz", "") { baz = true }
+    end
+    foo.should be_true
+    bar.should be_false
+    baz.should be_false
+  end
+
+  it "handles subcommand --help well (top level)" do
+    help = nil
+    OptionParser.parse(%w(--help)) do |opts|
+      opts.banner = "Usage: foo"
+      opts.on("subcommand", "Subcommand Description") do
+        opts.on("-f", "--foo", "Foo") { }
+      end
+      opts.on("--verbose", "Verbose mode") { }
+      opts.on("--help", "Help") { help = opts.to_s }
+    end
+
+    help.should eq <<-USAGE
+      Usage: foo
+          subcommand                       Subcommand Description
+          --verbose                        Verbose mode
+          --help                           Help
+      USAGE
+  end
+
+  it "handles subcommand --help well (subcommand)" do
+    help = nil
+    OptionParser.parse(%w(subcommand --help)) do |opts|
+      opts.banner = "Usage: foo"
+      opts.on("subcommand", "Subcommand Description") do
+        opts.banner = "Usage: foo subcommand"
+        opts.on("-f", "--foo", "Foo") { }
+      end
+      opts.on("--verbose", "Verbose mode") { }
+      opts.on("--help", "Help") { help = opts.to_s }
+    end
+
+    help.should eq <<-USAGE
+      Usage: foo subcommand
+          --verbose                        Verbose mode
+          --help                           Help
+          -f, --foo                        Foo
+      USAGE
   end
 end
--- a/src/option_parser.cr
+++ b/src/option_parser.cr
@@ -220,8 +220,9 @@ class OptionParser
     end
   end
 
-  # Adds a separator, with an optional header message,
-  # that will be used to print the help.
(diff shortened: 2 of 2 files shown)


spec/std/option_parser_spec.cr
```diff
@@ -101,8 +101,8 @@ describe "OptionParser" do
     expect_capture_option ["--flag=123"], "--flag=FLAG", "123"
   end
 
-  it "has required option with = (3) raises" do
-    expect_missing_option ["--flag="], "--flag=FLAG", "--flag"
+  it "has required option with = (3) handles empty" do
+    expect_capture_option ["--flag="], "--flag=FLAG", ""
```
RX14, line 105: I changed the behaviour of `OptionParser` here because I thought this behaviour was stupid.

oprypin:
I think this fixes https://github.com/crystal-lang/crystal/issues/8937

RX14:
Yeah, me too

I'll add a spec.

straight-shoota:
Awesome!

This might also enable to implement #5338 (could be a follow-up though).

RX14:
Ah, that seems rather difficult with the current code because there's no way to "peek" at the next argument and see if it's valid or not

I might just get  rid of the "iterator" business and use indexing in the array, then.

straight-shoota:
I suppose it might be best to support only `--optional-flag=VALUE` or `--optional-flag` (no VALUE), an not `--optional-flag VALUE` because the latter is always kind of ambiguous, even with a fixed set of values and it can't handle arbitrary values at all.

RX14:
It's definitely doable to have a "next argument doesn't start with -" heuristic.

I switched the code to use array indexing and it's neater anyway.

Blacksmoke16:
Thoughts on allowing multiple shortflags to be grouped?

```cr
require "option_parser"

OptionParser.parse do |opts|
  opts.on("-f", "Prints foo") { pp "foo" }
  opts.on("-b", "Prints bar") { pp "bar" }
end
```

Works:
`./test -f -b`

Does not work:
`./test -fb`

RX14:
@Blacksmoke16 I thought about adding that feature, but I'd rather leave it for another PR.

Blacksmoke16:
Would also close #3357.

Another thought I had, probably better for another PR; but what about parsing options with more than one value?  An example of this is the `--arg name value` within https://stedolan.github.io/jq/manual/.

oprypin:
@Blacksmoke16  It is indeed unrelated and is also not part of any spec or common use.

oprypin:
I wrote this reply before realizing that something else was being asked, but

if one wants to support an arbitrary number of values, it should definitely not be done as
`--stuff foo bar baz` (non-compliant and very ambiguous in terms of what can follow it)
but rather `--stuff foo --stuff=bar --stuff=baz`,
or, as a shell would let you write, `--stuff={foo,bar,baz}`.

And this is already supported, just push it to an array:
<details>
<summary>Code</summary>

```crystal
require "option_parser"

argv = ["--stuff", "foo", "--stuff=bar"]

stuff = [] of String

OptionParser.parse(argv) do |parser|
  parser.on("--stuff=NAME", "") { |name| stuff << name }
end

p! stuff # => ["foo", "bar"]
```

</details>

Blacksmoke16:
@oprypin I wasn't imagining supporting an arbitrary number of values, just a way to handle a fixed number that is more than one.

Like `parser.on("--arg NAME VALUE", "Add an arg") { |name, value| # Do whatever }`

oprypin:
>>>>I wrote this reply before realizing that something else was being asked, but

straight-shoota:
@Blacksmoke16 IMO that looks like a very uncommon feature. I'd rather not have such specialized things in OptionParser.

Btw. why not just use `--arg NAME=VALUE` for example? Would be much more convenient IMO.

Blacksmoke16:
Oh I don't really have a use case.  Was just something I noticed in `jq`.

sudo-nice:
Thank you for the great improvement of the OptionParser!
Will those sticky arguments (like `command -ufoo -tbar`) still be working after the update?
<details>
  <summary>Currently working example</summary>

  ```crystal
  require "option_parser"

  ary = [] of String
  options = %w[-ufoo -tbar]

  OptionParser.parse(options) do |p|
    p.on("-u FOO", "Description") { |o| ary << o }
    p.on("-t FOO", "Description") { |o| ary << o }
  end

  pp ary # => ["foo", "bar"]
  ```
</details>

oprypin:
You can just copy paste [the file](https://github.com/RX14/crystal/blob/feature/option-parser-rewrite/src/option_parser.cr) before your example to try.

Yes, it works before and after.

RX14:
I've just added commits to unregister subcommands when they're called, this means that when you call a subcommand `foo`, *all* subcommands are unregistered, so `foo foo` doesn't handle `foo` twice.

I've also added two new features: `OptionParser#stop` and `OptionParser#every(&)`. The latter takes a callback to run on every single argument before it's parsed. This is neccesary, because performing a `map` or `each` on `ARGV` wouldn't be able to tell the difference between flags and their arguments, whereas `every` is not passed flag arguments. Then `OptionParser#stop` allows you to bail out of parsing at any point.

spec/std/option_parser_spec.cr
```diff
@@ -438,4 +442,127 @@ describe "OptionParser" do
...
+  end
+
+  it "can run a callback on every argument" do
+    args = %w(--foo file --bar)
+    foo = false
+    bar = false
+    OptionParser.parse(args) do |opts|
+      opts.on("--foo", "") { foo = true }
+      opts.on("--bar", "") { bar = true }
+      opts.every do |arg|
```
straight-shoota, line 553: I find it a bit confusing that calling `opts.stop` in `every` block behaves differently than in `on` block (/re whether the current arg is consumed). Maybe renaming it to `before_every` would help. Or `before_each`?

src/option_parser.cr
```diff
@@ -201,141 +228,134 @@ class OptionParser
...
-          @parser.missing_option.call(flag) if raise_if_missing && value.empty?
-          block.call value
+        # If we require a value and we don't have one, call missing option
+        @missing_option.call(flag) if handler.value_type.required? && value.nil?
+
+        # If this is a subcommand (flag not starting with -), delete all
+        # subcommands since they are no longer valid.
+        unless flag.starts_with?('-')
+          @handlers.select! { |k, v| k.starts_with?('-') }
+          @flags.select! { |flag| flag.starts_with?("    -") }
```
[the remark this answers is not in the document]
RX14: Check out https://github.com/crystal-lang/crystal/blob/7e2e84042a393223660a216fbc5d0d5cd6bb80a5/src/option_parser.cr#L178

spec/std/option_parser_spec.cr
RX14: yeah, before_each sounds good for me.

RX14:
`every` has been renamed to `before_each` and documentation is complete. This is ready for final review.

straight-shoota:
I'm not entirely sure about using the same method for defining a subcommands and flags when the only difference is whether the string starts with a dash.
Maybe a different method name for sucommand would be better to more clearly communicate the different concepts. For example `opts.subcommand("subcommand", "") do`.

RX14:
I mean, we could have different methods for short and long flags because they have differet behaviours too.

I think this is fine.

didactic-drunk:
Is reuse of an `OptionParser` allowed with new data?  I have a repl loop that reuses `OptionParser` objects for it's command interface.

Does this handle nested subcommands? I have a program that has approximately 80 commands total up to 4? subcommands deep.  Each command has it's options parsed at the subcommand level.
```
-v foo --log=log.txt bar --drinks=2 baz 
-v foo --log=log.txt --drinks=2 bar baz  # Invalid.  --drinks is bar scope only.
-v foo bar --log=log.txt --drinks=2 baz  # Invalid.  --log is foo scope only.
foo -v --log=log.txt bar --drinks=2 baz  # Invalid.  -v is global scope only.
```

Will this accomodate?

RX14:
@didactic-drunk you can't reuse an optionparser after it's parsed subcommands because it's mutated the handlers list.

Nested subcommands work fine. `--flags` are supported *after* the subcommand which registers them. So your last two examples would be valid, and this is fine.

This is not intended to do everything. This is intended to solve common problems. Whatever you're doing doesn't seem to be typical...

RX14:
Please can I have another review, I have a PR waiting on this...

asterite:
> @didactic-drunk you can't reuse an optionparser after it's parsed subcommands because it's mutated the handlers list

I think this should be modified. The current optionparser (not the one in this PR) creates a `ParseTask` and runs that so that it doesn't modify the original option parser.

@RX14 I think doing the same would be easy: just clone all the handlers and everything and pass it to another class that does the parsing. That way it'll also be easier to understand which class is in charge of the configuration and which is in charge of the parsing.

straight-shoota:
> You should be able to reuse an OptionParser

Why? What's the use case for that? OptionParser is mostly used in CLI applications that just need to parse the program arguments once.

If you want to implement a REPL or similar that needs repeated parsing, you can just configure the option parser and clone it before doing the actual parse. It should be fine to require this simple extra step for such advanced use cases. It keeps the overall complexity low.

asterite:
> Why?

Because in my mind it doesn't make sense that if I use OptionParser twice I get two different results. Plus it's a breaking change. And I'm sure it works like that in Ruby too.

RX14:
I don't think this is possible, because of the way `OptionParser` implements subcommands:

```cr
OptionParser.parse(args) do |opts|
  opts.on("subcommand", "") do
    opts.on("--foo arg", "") { |v| foo = v }
    opts.on("--bar", "") { bar = true }
    opts.on("sub2", "") { sub2 = true }
  end
  opts.on("--verbose", "") { verbose = true }
  opts.on("-z", "--baz", "") { z = true }
end
```

The `OptionParser` is mutated while parsing using `opts.on` to register new flags.

It'd take a redesign of the interface to support this, and I'm personally not sure it's worth it. It's not a breaking change either, because the only time `OptionParser` is mutated is when doing subcommands, which we didn't support before anyway.

asterite:
I think this can be implemented without mutation. Until that's done, sorry, I'm not approving this.

didactic-drunk:
Couldn't mutation be solved by passing a new `OptionParser` object to the subcommand block?

```
OptionParser.parse(args) do |opts|
  opts.on("subcommand", "") |subopts| do
  end
end
```

straight-shoota:
@asterite Do you think there's anything wrong with explicitly using `clone` if you want to reuse an instance?

asterite:
> Do you think there's anything wrong with explicitly using clone if you want to reuse an instance?

Yes, it's unexpected. OptionParser defines how options should be parsed. I find it pretty strange that if I use it twice, it behaves differently.

asterite:
Crystal is not a functional programming language, but we tend to use and like purity. That is, methods shouldn't have unexpected side effects.

RX14:
I made `#parse` restore state on exit.

I tried going with the route of cloning the `OptionParser` and creating a DSL like @didactic-drunk suggested, but it's not neat to synchronise `@stop`, and the `@before_each` handlers etc. between the *original* instance which is passed to the `OptionParser.new do |parser|` block, and the instance which is passed to `parser.subcommand("foo", "") do |parser|`. Basically you need to unify all the instance variables between the two *except* `@flags` and `@handlers`, which is much uglier than what I just comitted.

didactic-drunk:
Can you have an additional `OptionParser` per subcommand and start parsing ARGV from the leftovers?  Doesn't the parent OptionParser stop when a subcommand is reached (eliminating the @stop problelm)?  Perhaps @asterite's state suggestion would solve the issue?

`def parse(args = ARGV, state = State.new)`

asterite:
I guess with that proof we can go ahead without the last commit, and add it later if needed.

RX14:
@didactic-drunk in short, no.

RX14:
We might as well leave the last commit. Why not?

src/option_parser.cr
```diff
@@ -417,6 +425,15 @@ class OptionParser
...
     end
+  ensure
+    @flags = old_flags.not_nil!
+    @handlers = old_handlers.not_nil!
+    @stop = false
+    @banner = old_banner
+    @unknown_args = old_unknown_args
+    @missing_option = old_missiong_option.not_nil!
+    @invalid_option = old_invalid_option.not_nil!
+    @before_each = old_before_each
```
Sija, line 436: Wouldn't be better to use `begin ... end` block instead of using `.not_nil!`?

RX14: I don't think it's worth the extra indent. They're semantically equivalent anyway.

oprypin: OK then how about clumping it all into one variable (tuple? record? maybe store the whole state in a sub-record in the first place?)

straight-shoota: It should be either as is or without the reset.

RX14:
I am really ambivalent about leaving the last commit in or not, since it's less work to do so, I'll leave it in.

I'll merge this today if there's no more API-level objections.

src/option_parser.cr
Sija: @RX14 Using `.not_nil!` should be treated as a last resort, so if there's a way to avoid it, even at the "cost" of an extra indent, then I'd say it's worth it.

asterite: Just put the begin after you save all those variables. I agree with Sija.

src/option_parser.cr
```diff
@@ -201,141 +293,151 @@ class OptionParser
...
 
-  private struct ParseTask
-    @double_dash_index : Int32?
+  # Parses the passed *args* (defaults to `ARGV`), running the handlers associated to each option.
+  def parse(args = ARGV)
+    old_flags = @flags.clone
+    old_handlers = @handlers.clone
+    old_banner = @banner
+    old_unknown_args = @unknown_args
+    old_missiong_option = @missing_option
```
Sija, line 302:
```suggestion
    old_missing_option = @missing_option
```

waj:
@RX14 thanks for working on this. I agree that the current `OptionParser` is bad designed and buggy. This implementation is better than the one we currently have and from that perspective I have no objections to merge this PR.

Now, I wouldn't consider this implementation "good code". I don't need to check usage stats to understand that the approach used for subcommands is not well designed. Mutating the parser is unnecessary and TBH relying on string comparison with a specific number of spaces from a formatted text to do so, makes me cringe.

I think it could actually just make it store the options more structured, format those options just when it's needed (on `--help`) and filter based on those structures, not strings.

Also, I wish the subcommands implementation could work somehow with separate `OptionParser` instances. That way subcommands could be defined in separate files, or make them join/split to different processes more easily. For example, take a look at how it's currently used for Crystal compiler subcommands, and think how this new implementation would fit there.

I could understand that you don't feel like doing it right now, but I wouldn't say this PR is the ultimate design and good code that we could be looking for.

didactic-drunk:
> Also, I wish the subcommands implementation could work somehow with separate `OptionParser` instances. That way subcommands could be defined in separate files, or make them join/split to different processes more easily.

Should subcommands be marked experimental?  Should `opts.on` be `opts.subcommand` or some other method to make the meaning clear?

RX14:
> I could understand that you don't feel like doing it right now, but I wouldn't say this PR is the ultimate design and good code that we could be looking for.

On that, I totally agree. The `--help` text generation needs to be refactored. The rest isn't so bad. But this is still better than before.

PR #5858, opened 2018-03-23
Unify #inspect output formatting in several places

Sija:
_Reference_: `#<Type …>`
_Struct_: `Type(…)`

PR #9010, opened 2020-04-06
Replace old-style `Regex::MatchData` inspection in docs examples

MakeNowJust:
#5858 fixes `Regex::MatchData#inspect` output, however docs examples is forgot to update.
Thank you.


straight-shoota:
Thanks, @MakeNowJust

PR #9015, opened 2020-04-08
Fix File#pos, File#seek and File#truncate over 2G on Windows

kubo:
[_lseek][_lseeki64] and [_chsize][] cannot handle files bigger than 2G because they use Long, whose maximum size is 2G on Windows. This commit uses [_lseeki64][] and [_chsize_s][] instead.

[_chsize]: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/chsize
[_chsize_s]: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/chsize-s
[_lseeki64]: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/lseek-lseeki64

### DIFF
--- a/src/crystal/system/win32/file.cr
+++ b/src/crystal/system/win32/file.cr
@@ -213,7 +213,7 @@ module Crystal::System::File
   end
 
   private def system_truncate(size : Int) : Nil
-    if LibC._chsize(fd, size) != 0
+    if LibC._chsize_s(fd, size) != 0
       raise ::File::Error.from_errno("Error truncating file", file: path)
     end
   end
--- a/src/crystal/system/win32/file_descriptor.cr
+++ b/src/crystal/system/win32/file_descriptor.cr
@@ -78,7 +78,7 @@ module Crystal::System::FileDescriptor
   end
 
   private def system_seek(offset, whence : IO::Seek) : Nil
-    seek_value = LibC._lseek(fd, offset, whence)
+    seek_value = LibC._lseeki64(fd, offset, whence)
 
     if seek_value == -1
       raise IO::Error.from_errno "Unable to seek"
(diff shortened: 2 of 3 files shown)


src/crystal/system/win32/file.cr
```diff
@@ -212,8 +212,8 @@ module Crystal::System::File
     end
   end
 
-  private def system_truncate(size : Int) : Nil
-    if LibC._chsize(fd, size) != 0
+  private def system_truncate(size) : Nil
```
asterite, line 215: Why is the `Int` restriction gone?

kubo: That's because I'm a newbie about crystal and I didn't know that `Int` was the base type of all integer types.

I thought that `Int` was 32-bit by analogy with C and changed the restriction from `Int` to `Int64` at first because the `size` argument is passed to the `Int64` argument of `LibC._chsize_s()`. However I got an error by the following code.

```crystal
File.open("test-file.txt", "w") do |file|
  file << "123456789012345678901234567890"
  file.truncate(20)
end
```
```shell
$ ./crystal/.build/crystal build --cross-compile --target x86_64-pc-windows-msvc "$@" trunc-test.cr 

In crystal/src/file.cr:763:5

 763 | system_truncate(size)
       ^--------------
Error: no overload matches 'File#system_truncate' with type Int32

Overloads are:
 - Crystal::System::File#system_truncate(size : Int64)
```
I could not imagine the error reason. I checked the definition of `system_truncate` in `src/crystal/system/unix/file.cr` and found that it didn't have such restriction. So I removed the restriction and it worked. I made a pull request after that.

I just now checked that it worked with the `Int` restriction. Should I revert the change at the line and push a commit?

asterite: Yes, please!

`Int` is not well named, it should be named `Integer`. I hope we can rename it in the future.

kubo: I pushed 5c20b89. Is it Okay?

asterite: Perfect, thank you!

PR #9019, opened 2020-04-08
Define PID type as Int64

oprypin:
`Int64` is enough to fit the actual underlying values of different platforms (`Int32` on Linux, `UInt32` on Windows).

This will allow to make Process cross-platform without returning a different type to users on different platforms.

I am keeping the internal variable as `LibC::PidT`, because it is part of the UNIX-specific implementation details which process.cr is still full of. That detail will be moved out later.

### DIFF
--- a/src/process.cr
+++ b/src/process.cr
@@ -4,10 +4,6 @@ require "c/sys/resource"
 require "c/unistd"
 
 class Process
-  # A platform-specific identifier of a running process.
-  # Int64 is enough to fit underlying values of different platforms (Int32 or UInt32).
-  alias PID = Int64
-
   # Terminate the current process immediately. All open files, pipes and sockets
   # are flushed and closed, all child processes are inherited by PID 1. This does
   # not run any handlers registered with `at_exit`, use `::exit` for that.
@@ -18,25 +14,25 @@ class Process
   end
 
   # Returns the process identifier of the current process.
-  def self.pid : PID
-    PID.new(LibC.getpid)
+  def self.pid : Int64
+    LibC.getpid.to_i64
   end
 
   # Returns the process group identifier of the current process.
-  def self.pgid : PID
-    PID.new(pgid(0))
+  def self.pgid : Int64
+    pgid(0).to_i64
   end
 
   # Returns the process group identifier of the process identified by *pid*.
-  def self.pgid(pid : Int) : PID
+  def self.pgid(pid : Int) : Int64
     ret = LibC.getpgid(pid)
     raise RuntimeError.from_errno("getpgid") if ret < 0
-    PID.new(ret)
+    ret.to_i64
   end
 
   # Returns the process identifier of the parent process of the current process.
-  def self.ppid : PID
-    PID.new(LibC.getppid)
+  def self.ppid : Int64
+    LibC.getppid.to_i64
   end
 
   # Sends a *signal* to the processes identified by the given *pids*.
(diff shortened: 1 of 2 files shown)


src/process.cr
```diff
@@ -4,6 +4,10 @@ require "c/sys/resource"
 require "c/unistd"
 
 class Process
+  # A platform-specific identifier of a running process.
+  # Int64 is enough to fit underlying values of different platforms (Int32 or UInt32).
+  alias PID = Int64
```
straight-shoota, line 9: I don't think this alias is really helpful. It suggests a distinct type safe type when it is really just an alias of `Int64`.
I'd rather make it explicit as `Int64` or use an actual value type `record PID, value : Int64`.

oprypin: My goal here is to make it easy to find usages of PID in source code, rather than digging through Int64s. I don't see why not.

straight-shoota: I don't like the false type-safety it suggests. You would assume these methods return and receive only instances of `PID`. But that doesn't exist, it's just `Int64`.

oprypin: OK. How do we answer the inevitable question "why does it return Int64 if pids are Int32"?

straight-shoota: Documentation shouldn't be an issue. It can just go into the class' doc comment.

oprypin: I'll wait for a tie-breaker comment :upside_down_face:

RX14: I don't see a problem with this.

oprypin: Could you clarify which you prefer?

asterite: I think using a `record PID` is best, if the API returns such things and accepts such things. Then it can have `to_i` and `to_i64` if needed, and we use it internally, and people can use that too if needed. And then if we need to change the underlying type it should be safe.

But maybe it's overkill, I don't know.

RX14: I don't see a problem with it as it is now - `alias PID = Int64`. After all, we do the same with `Bytes`.

straight-shoota: But `Bytes` is conceptually just a slice of UInt8, so this really just a different name fore xactly the same thing. But for a PID type I would expect some specific properties, that are not shared by Int64. For example a narrower value range. And that I can't accidentally assign arbitrary int values, but would be required to explicitly cast a number to PID.

RX14: Bikeshedding. There's no need to care deeply about this minor detail. This is good enough, let it be.

straight-shoota: I disagree on being a minor detail. To me it's essential for approving this PR. When PID is just Int64, the only benefit is improved discoverability in the API docs. But at the cost of suggesting a false type safety. I would probably prefer calling it Int64 directly then.

RX14:
This needs a rebase, and I'll wait for other core team members to chip in, but I think i've made my opinion on the issue above clear.

bcardiff:
Something I don't get is that if the PIDs are Int32 or UInt32; isn't it better to use Int32 and cast them to UInt32 on Windows with `to_u32!`?

oprypin:
Do we want Windows users to operate on possibly-negative PIDs which would not represent the truth?

bcardiff:
I guess not. :-) Int64 it is then.

pebauer68:
> Do we want Windows users to operate on possibly-negative PIDs which would not represent the truth?

How could they get a negative PID ? 32 Bit is both enough for Linux and Windows users.

oprypin:
Max PID on Windows is `UInt32::MAX`, which doesn't fit in `Int32`
https://stackoverflow.com/questions/17868218/what-is-the-maximum-process-id-on-windows#comment26090248_17868515

PR #9021, opened 2020-04-08
Make Process::Status Windows-compatible

oprypin:
Skip the bitmasks/signal handling.

Closes #8537

### DIFF
--- a/src/process/status.cr
+++ b/src/process/status.cr
@@ -3,7 +3,7 @@ class Process::Status
   # Platform-specific exit status code, which usually contains either the exit code or a termination signal.
   # The other `Process::Status` methods extract the values from `exit_status`.
   def exit_status : Int32
-    @exit_status.to_i32
+    @exit_status.to_i32!
   end
 
   {% if flag?(:win32) %}
--- a/src/process/status.cr
+++ b/src/process/status.cr
@@ -1,38 +1,65 @@
-# The status of a terminated process.
+# The status of a terminated process. Returned by `Process#wait`.
 class Process::Status
   # Platform-specific exit status code, which usually contains either the exit code or a termination signal.
   # The other `Process::Status` methods extract the values from `exit_status`.
-  getter exit_status : Int32
-
-  def initialize(@exit_status : Int32)
+  def exit_status : Int32
+    @exit_status.to_i32
   end
 
+  {% if flag?(:win32) %}
+    # :nodoc:
+    def initialize(@exit_status : UInt32)
+    end
+  {% else %}
+    # :nodoc:
+    def initialize(@exit_status : Int32)
+    end
+  {% end %}
+
   # Returns `true` if the process was terminated by a signal.
-  def signal_exit?
-    # define __WIFSIGNALED(status) (((signed char) (((status) & 0x7f) + 1) >> 1) > 0)
-    ((LibC::SChar.new(@exit_status & 0x7f) + 1) >> 1) > 0
+  def signal_exit? : Bool
+    {% if flag?(:unix) %}
+      # define __WIFSIGNALED(status) (((signed char) (((status) & 0x7f) + 1) >> 1) > 0)
+      ((LibC::SChar.new(@exit_status & 0x7f) + 1) >> 1) > 0
+    {% else %}
+      false
+    {% end %}
   end
 
   # Returns `true` if the process terminated normally.
-  def normal_exit?
-    # define __WIFEXITED(status) (__WTERMSIG(status) == 0)
-    signal_code == 0
+  def normal_exit? : Bool
+    {% if flag?(:unix) %}
+      # define __WIFEXITED(status) (__WTERMSIG(status) == 0)
+      signal_code == 0
+    {% else %}
+      true
+    {% end %}
   end
 
   # If `signal_exit?` is `true`, returns the *Signal* the process
   # received and didn't handle. Will raise if `signal_exit?` is `false`.
-  def exit_signal
-    Signal.from_value(signal_code)
+  #
+  # Available only on Unix-like operating systems.
+  def exit_signal : Signal
+    {% if flag?(:unix) %}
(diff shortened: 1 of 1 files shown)


src/process/status.cr
```diff
@@ -1,38 +1,65 @@
-# The status of a terminated process.
+# The status of a terminated process. Returned by `Process#wait`.
 class Process::Status
   # Platform-specific exit status code, which usually contains either the exit code or a termination signal.
   # The other `Process::Status` methods extract the values from `exit_status`.
-  getter exit_status : Int32
-
-  def initialize(@exit_status : Int32)
+  def exit_status : Int32
+    return @exit_status.to_i32
```
Sija, line 6:
```suggestion
    @exit_status.to_i32
```

straight-shoota:
This is essentially an alternative to #8537, right? At least the `Status` part of that.

As already mentioned there, we should also consider #8647, which essentially makes part of the special handling unneccessary.

oprypin:
It's not an alternative, more like a subset.

---
Deprecating that method is just that, an added annotation. It does not remove any need for that special handling.

straight-shoota:
Yeah, you're right. Let me rephrase: It would have been better to deprecate and remove `exit_status` first 😆

oprypin:
Hm no, it really makes no difference. The two PRs support each other well (i.e. merging either one makes the other one make even more sense) but don't depend on each other in any way at all (other than needing a manual merge)

oprypin:
BTW there's no backwards-incompatible change here

RX14:
Adding more `if flag?` outside of `src/crystal/system` is not the way I want to go about this.

oprypin:
I can agree with that only on the matter of principle. But I definitely think for this particular case keeping it in one file is better for maintenance.

straight-shoota:
@RX14 As per https://github.com/crystal-lang/crystal/pull/8647#issuecomment-571679213 I'd remove those platform-specifics from `Process::Status` altogether.

RX14:
Me too. 

Just, `record Process::Status` with `exit_signal` and `exit_code` and re-add the query helpers is fine.

oprypin:
I can grant this wish by copying the implementations of reader methods to the place where the struct is constructed, without having any idea how this exit status is processed.
I was thinking that perhaps this could be a dumb struct, and the implementations could just be put into crystal/system/process, but I don't see how to nicely solve the requirement that calling `exit_signal` needs to `raise` on Windows.

RX14:
> I don't see how to nicely solve the requirement that calling `exit_signal` needs to `raise` on Windows.

The problem is that the method returns `Signal`, even when the program didn't signal exit on posix either. It'd be much saner to have it return `nil` if the program didn't exit with a signal, be it either on windows or unix.

straight-shoota:
Or just don't have `exit_signal` at all...

oprypin:
Yes I'm very tempted to implement it as such...
Also curious to explore the possibility of making it just a number like many other programming languages do.

---

The discussion is sadly spread so much,
not much on the issue #8381
but quite some discussion on PR #8647

---

I'll add by comparing Crystal to what Python does. For my own understanding, too.

The outcome is that Python just returns a negative number if it's a signal.

```shell
$ crystal eval 'require "process"; r = Process.run("cat", input: :inherit); p! r.exit_status, r.exit_code, r.exit_signal' & while ! killall cat; do sleep 0.5; done; fg 
r.exit_status # => 15
r.exit_code   # => 0
r.exit_signal # => TERM
```
```shell
$ python3 -c 'import subprocess; print(subprocess.call("cat"))' & while ! killall cat; do sleep 0.5; done; fg
-15
```
---
```shell
$ crystal eval 'require "process"; r=Process.run("false", input: :inherit); p! r.exit_status, r.exit_code, r.exit_signal'                                              
r.exit_status # => 256
r.exit_code   # => 1
r.exit_signal # => Unhandled exception: Unknown enum Signal value: 0 (Exception)
```
```shell
$ python3 -c 'import subprocess; print(subprocess.call("false"))'                                              
1
```
---
It also seems quite strange to me that Crystal enforces that you can't find out the signal if it's one that Crystal is not aware of.

Also returning `exit_code` as 0 when the process is terminated.....

oprypin:
And coming back to the "dumb struct", it could turn out *quite* inefficient in terms of storage, perhaps as far as this:
`record exit_code : Int64?, exit_signal : Int64?`
\- what is it now - 24 bytes? 32?
when really 4 bytes is all that's needed

oprypin:
Honestly I'm not so interested in solving the design issue right here.

I think that the current state of the pull request is the only way to do it without hitting any of the disadvantages that other approaches here are prone to:
* backwards incompatible
* inefficient
* blocked on other changes

If one wants to follow this up with a more idealistic approach, this doesn't make it harder to do.

oprypin:
Please read this, I summarized all possible status codes:
https://github.com/crystal-lang/crystal/issues/8381#issuecomment-612487955

Having looked into how diverse the exit status can be, I don't see it as reasonable to pre-process and store this data in a plain struct that is the same on all platforms.

oprypin:
I am asking to merge this as is. There are no changes in behavior on supported systems. The only complaints are about where the code should *ideally* be. But it's clear that this will need more thought, and we will be in a better position for maneuvers after Process support on Windows is landed (as it will actually make the file crystal/system/process exist, which is where you want me to put this). But *that* change is *blocked* on *this one*.

cc @asterite

straight-shoota:
Thanks @oprypin

oprypin:
https://github.com/crystal-lang/crystal/pull/9064 will be my only followup to this. It's the only way to meet the above requirements.

The suggestions for refactoring here insisted on storing/exposing only processed data and throwing out `exit_status`. In that case we have to ensure that the processed data provides the full richness of information, otherwise some things become impossible to do (not *conveniently* do, but at all), and that's just a downgrade from the current state.

PR #9026, opened 2020-04-09
Update SSL server secure defaults

### LINKED ISSUE #9022: SECURITY: context_spec.cr specifies "only TLS 1.0" by calling tlsv1_method.  OpenSSL says avoid tlsv1_method.  TLS 1.0 and 1.1 are deprecated.
A. OpenSSL docs tell us to avoid calling `tlsv1_method` but crystal is calling it.

B. Calling `tlsv1_method` means a TLS/SSL connection will only understand TLS 1.0.

C. Current industry recommendation is to use least TLS 1.2:

* __June 30, 2018 deadline to drop TLS 1.0__ - The PCI Council (Payment Card Industry) suggested that organizations migrate from TLS 1.0 to TLS 1.1 or higher before June 30, 2018.

* __March 2020 deadline to deprecate TLS 1.1__ - In October 2018, Apple, Google, Microsoft, and Mozilla jointly announced they would deprecate TLS 1.0 and 1.1 in March 2020.

D. See Background (provided below) for timeline and some attacks on TLS for more context.

<hr>

### OpenSSL docs

__OpenSSL 1.1.0 says use TLS_method__ and avoid TLSv1_method, TLSv1_1_method, etc.
__OpenSSL 1.0.2 says use SSLv23_method__ and avoid TLSv1_method, TLSv1_1_method, etc.

<details>
  <summary>(click to expand) Quote from docs and links</summary><p>

>TLS_method(), TLS_server_method(), TLS_client_method()
>
>These are the general-purpose version-flexible SSL/TLS methods. The actual protocol version used will be negotiated to the highest version mutually supported by the client and the server. The supported protocols are SSLv3, TLSv1, TLSv1.1 and TLSv1.2. Applications should use these methods, and avoid the version-specific methods described below.
> ...
> TLSv1_2_method(), TLSv1_2_server_method(), TLSv1_2_client_method()
> ...
> TLSv1_1_method(), TLSv1_1_server_method(), TLSv1_1_client_method()
> ...
> TLSv1_method(), TLSv1_server_method(), TLSv1_client_method()
>  A TLS/SSL connection established with these methods will only understand the TLSv1 protocol.

* https://www.openssl.org/docs/man1.1.0/man3/TLSv1_method.html
* https://www.openssl.org/docs/man1.0.2/man3/TLSv1_method.html

</details>

### Relevant code in crystal

https://github.com/crystal-lang/crystal/blob/fd0780c3a4ef972a6090e09abc9b09a0e39345ff/spec/std/openssl/ssl/context_spec.cr#L31

https://github.com/crystal-lang/crystal/blob/fd0780c3a4ef972a6090e09abc9b09a0e39345ff/spec/std/openssl/ssl/context_spec.cr#L48

Additionally, TLS 1.0 and TLS 1.1 should be disabled from being chosen during auto-negotiation.  See Background (provided below).

## Background

__1999__. TLS 1.0 was first defined in RFC 2246 in January 1999. 

__2006__. TLS 1.1 was defined in RFC 4346 in April 2006.

__2008__. TLS 1.2 was defined in RFC 5246 in August 2008.

__2014__. TLS 1.0 allows downgrading the connection to SSL 3.0, thus weakening security (POODLE SSL Variant).   Source:   
https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.0  

__2014__. TLS 1.0, TLS 1.1, and TLS 1.2 (if not implemented properly) are vulnerable to POODLE TLS Variant even if SSLv3 is disabled.  Source:  
https://en.wikipedia.org/wiki/POODLE

__2017__. Google Chrome set TLS 1.3 as the default version for a short time in 2017. It then removed it as the default, due to incompatible middleboxes such as Blue Coat web proxies. Source:  
https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.3

__2018__.  TLS 1.3 was defined in RFC 8446 in August 2018. It mandates use of AEAD ciphers, key exchanges that offer perfect forward secrecy, integrates session hash, and drops support for many insecure or obsolete features.  Source:  
https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.3

#### TLS Interception Appliances
POODLE from 2014 is not the only security issue to consider, see this paper:
> To analyze  TLS-encrypted  data, network appliances implement a Man-in-the-Middle TLS proxy, by acting as the intended web server to a requesting client (e.g., a browser), and acting as the client to the actual/outside web server.  Source:  
[The Sorry State of TLS Security in Enterprise - Interception Appliances (PDF, arxiv.org)]( https://arxiv.org/pdf/1809.08729.pdf)

straight-shoota:
This PR updates the cipher list used for SSL::Context to only include ciphers from the intermediate compatibility level of the Mozilla recommandedation (https://wiki.mozilla.org/Security/Server_Side_TLS).
Besides less ciphers, `NO_TLS_V1` and `NO_TLS_V1_1` are added to the default options.

It also integrates a script to automatically update these cipher definitions and a script to test it with [testssl.sh.](https://testssl.sh/). Maybe those could also be integrated into the CI setup.

Fixes #9022


straight-shoota:
The testssl reports are available at https://gist.github.com/straight-shoota/d767dfa330e126519c4b60e27b656b58

With this change it looks pretty good 👍 

During the ssl test the server however raises lots of SSL_accept errors that are not handled anywhere. They should probably be rescued and ignores / turned into `Log.debug` messages.

straight-shoota:
For OpenSSL 1.1.0 the recommended means to select supported versions is through `set_min_proto_version` instead of setting individual options.

x448:
More detailed info than provided by Mozilla's recommendation:

https://github.com/ssllabs/research/wiki/SSL-and-TLS-Deployment-Best-Practices

scripts/ssl_server_defaults.cr
```diff
@@ -0,0 +1,46 @@
+require "http"
+require "json"
+
+level = ARGV.shift? || "intermediate"
+url = ARGV.shift? || "https://ssl-config.mozilla.org/guidelines/latest.json"
+
+json= JSON.parse(HTTP::Client.get(url).body)
```
Sija, line 7:
```suggestion
json = JSON.parse(HTTP::Client.get(url).body)
```

waj:
This is awesome!
How often do you think we should update the guidelines? Before each release?
Should we expose the guideline version in a constant? We could use it to check for updates later.

scripts/generate_ssl_server_defaults.cr
```diff
@@ -0,0 +1,46 @@
...
+
+File.open(DEFAULTS_FILE, "w") do |file|
+  file.print <<-CR
+  # THIS FILE WAS AUTOMATICALLY GENERATED BY script/ssl_server_defaults.cr
+  # on #{Time.utc}.
+
+  abstract class OpenSSL::SSL::Context
+  CR
+
+  configuration = json["configurations"].as_h.each do |level, configuration|
```
asterite, line 23: Could we use JSON::Serializable? I think it would make the code a bit easier to read.

straight-shoota:
We could technically check for a new version of the recommendations on every nightly build 🤷 

I suppose watching https://github.com/mozilla/server-side-tls for releases should be sufficient.

scripts/generate_ssl_server_defaults.cr
straight-shoota: I don't think that improves much. Would just remove a few square brackets and as_a.

asterite: Yeah, maybe it's not worth it, I don't know.

Just in case, here's the mapping if we decide to use it:

```crystal
record Guidelines,
  href : String,
  version : Float64,
  configurations : Hash(String, Configuration) do
  include JSON::Serializable
end

record Configuration,
  oldest_clients : Array(String),
  ciphersuites : Array(String),
  ciphers : Ciphers do
  include JSON::Serializable
end

record Ciphers, openssl : Array(String) do
  include JSON::Serializable
end
```

x448:
It's too bad Mozilla doesn't have a "semi-modern" recommendation requiring AEAD+PFS with TLS 1.2-TLS1.3.  All major browsers, etc. have supported AEAD+PFS for years already.

straight-shoota:
@x448 We could still consider adding such a configuration, even if Mozilla doesn't have it.

x448:
> @x448 We could still consider adding such a configuration, even if Mozilla doesn't have it.

@straight-shoota that would be great!

Suggestions for "semi-modern":
 * Allow both TLS 1.2 and TLS 1.3
 * Ciphers must be AEAD+FS but avoid DHE (see list below.)
 * Client preferred order for ciphers (phones, etc. will move CHACHA20 to the front)
 * Server preferred order for named groups:   
    x25519, secp256r1, x448, secp384r1, secp521r1

```
CIPHERS_SEMIMODERN = 
"TLS_AES_128_GCM_SHA256
 TLS_CHACHA20_POLY1305_SHA256 
 TLS_AES_256_GCM_SHA384 
 ECDHE-ECDSA-AES128-GCM-SHA256 
 ECDHE-RSA-AES128-GCM-SHA256 
 ECDHE-ECDSA-CHACHA20-POLY1305 
 ECDHE-RSA-CHACHA20-POLY1305 
 ECDHE-ECDSA-AES256-GCM-SHA384 
 ECDHE-RSA-AES256-GCM-SHA384 
 !RC4 !aNULL !eNULL !LOW !3DES !MD5 !EXP !PSK !SRP !DSS"
```

I moved CHACHA20 ahead of AES256 and removed DHE.  Looks like Firefox did the same!

### Firefox ESR on linux-amd64

image

straight-shoota:
I'd prefer to discuss that in a separate issue. Let's focus here.

x448:
OK, deleted suggestion for semi-modern settings.  I misunderstood "@x448 We could still consider adding such a configuration, even if Mozilla doesn't have it." as invitation in this PR.

I think the inclusion of DHE cipher suites (instead of limiting to ECDHE) will lead to security issues.  It was one of the issues addressed by semi-modern settings.

ysbaddaden:
Please remember this is only defaults, meant to support most clients, trying to keep breakages to a minimum. You can tweak the context to add/remove support for clients in your applications (be careful).

Using rules from an external authority —that knows much better than we do— allows us to provide acceptable defaults, but also meant to avoid arguments on security topics. What could be discussed (in a dedicated issue) is introducing the `modern` and `legacy` levels, in addition to the `intermediate` level.

Introducing a script to create the context automatically from these recommendations is very welcome :+1:

x448:
I opened #9060 which affects all TLS settings that allow DHE.

PR #9028, opened 2020-04-09
Add support to jquery version 3

deiv:
Change some envents handlers, to use the 'on' method

### DIFF
--- a/src/compiler/crystal/tools/playground/public/application.js
+++ b/src/compiler/crystal/tools/playground/public/application.js
@@ -43,7 +43,7 @@ $(function(){
   session.onChange = function() {
     saveAsLastCode();
   };
-  $(window).unload(function(){
+  $(window).on("unload", function(){
     saveAsLastCode();
   });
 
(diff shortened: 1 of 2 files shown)


RX14:
But why? We use jquery 2.2.1?

asterite:
Thank you! 

Is there something broken that this fixes?

deiv:
Not, it not fixes something, it onlys allows to run the play server, with jQuery >= 3.

Use case:
For package in Debian, we drop the javascript libraries distributed inside the crystal source code (jquery, codemirror, etc). The current jQuery version in Debian is 3, so I use it in the play server.

asterite:
Did anyone try this? If it works, I see no reason not to merge it.

Sija:
Would it make sense to update vendored jQuery to v3 then?

straight-shoota:
Thanks @deiv 

@Sija jQuery could be updated, but I don't think it's a pressing matter. The playground is only intended for local use anyways, and as longs as it works...
But if someone feels like doing that, please provide a PR.

PR #9031, opened 2020-04-09
Allow building compiler without 'playground', to avoid dependency on sockets

oprypin:
This would, for example, be useful to bootstrap the compiler on Windows before having a "socket" implementation.

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -75,7 +75,12 @@ class Crystal::Command
       result
     when "play".starts_with?(command)
       options.shift
-      playground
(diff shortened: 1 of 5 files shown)


oprypin:
Yes it would be nice to not always pass `-Dwithout_playground` on Windows, but this way it's less controversial.

I was thinking about some explicit logic like
`if has_flag?(:win32); add_flag(:without_playground)`
but [flags.cr](https://github.com/crystal-lang/crystal/blob/f3c8f31aab2c4385f15647ea15671e635bb7c218/src/compiler/crystal/semantic/flags.cr) doesn't look like it should be concerned with that.

Ready to merge, from my side

PR #9033, opened 2020-04-09
Parser: correctly parse `foo:"bar"` inside call or named tuple

### LINKED ISSUE #9027: Parsing errors with no spaces after colon
The following expressions fail to parse when no space is added right after the colon:

```crystal
bar(foo:"a")
bar(foo:a)
{foo:"a"}
{foo:'a'}
{foo:a}
```

asterite:
Fixes #9027

Diff better seen with "Hide whitespace changes" checked.

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -1508,6 +1508,12 @@ module Crystal
       it_parses "{% begin %}%q#{open} %s #{close}{% end %}", MacroIf.new(true.bool, MacroLiteral.new("%q#{open} %s #{close}"))
     end
 
+    it_parses %(foo(bar:"a", baz:"b")), Call.new(nil, "foo", named_args: [NamedArgument.new("bar", "a".string), NamedArgument.new("baz", "b".string)])
+    it_parses %(foo(bar:a, baz:b)), Call.new(nil, "foo", named_args: [NamedArgument.new("bar", "a".call), NamedArgument.new("baz", "b".call)])
+    it_parses %({foo:"a", bar:"b"}), NamedTupleLiteral.new([NamedTupleLiteral::Entry.new("foo", "a".string), NamedTupleLiteral::Entry.new("bar", "b".string)])
+    it_parses %({foo:'a', bar:'b'}), NamedTupleLiteral.new([NamedTupleLiteral::Entry.new("foo", CharLiteral.new('a')), NamedTupleLiteral::Entry.new("bar", CharLiteral.new('b'))])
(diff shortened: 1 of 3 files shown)


asterite:
Haha, apparently. I think it consists of all the possible operators as symbols.

spec/compiler/parser/parser_spec.cr
```diff
@@ -1508,6 +1508,12 @@ module Crystal
       it_parses "{% begin %}%q#{open} %s #{close}{% end %}", MacroIf.new(true.bool, MacroLiteral.new("%q#{open} %s #{close}"))
     end
 
+    it_parses %(foo(bar:"a", baz:"b")), Call.new(nil, "foo", named_args: [NamedArgument.new("bar", "a".string), NamedArgument.new("baz", "b".string)])
+    it_parses %(foo(bar:a, baz:b)), Call.new(nil, "foo", named_args: [NamedArgument.new("bar", "a".call), NamedArgument.new("baz", "b".call)])
+    it_parses %({foo:"a", bar:"b"}), NamedTupleLiteral.new([NamedTupleLiteral::Entry.new("foo", "a".string), NamedTupleLiteral::Entry.new("bar", "b".string)])
+    it_parses %({foo:'a', bar:'b'}), NamedTupleLiteral.new([NamedTupleLiteral::Entry.new("foo", CharLiteral.new('a')), NamedTupleLiteral::Entry.new("bar", CharLiteral.new('b'))])
+    it_parses %({foo:a, bar:b}), NamedTupleLiteral.new([NamedTupleLiteral::Entry.new("foo", "a".call), NamedTupleLiteral::Entry.new("bar", "b".call)])
```
bcardiff, line 1515: What happens with `{foo::bar, baz::qux}` ?

asterite: Good question. It gives a syntax error. I think it's fine, I wouldn't parse that as `{foo: :bar, baz: :qux}`.

asterite: mainly because `::` has a different meaning

bcardiff: `foo ::bar` is `foo(::bar)`
But `::` as a namespace separator is only that between `Constants`

I'm fine either way :-)

asterite: Yes, in Crystal. In Ruby you can do that with any expression, or at least with var. I wouldn't discard this in the future.

waj:
> Is that really a 170 line long if branch? 😲

Maybe parsers it's the only place where we allow this. This should be autogenerated code from a grammar, but @asterite writes better parsers than any generator out there! 😄

Issue #8108, opened 2019-08-22
openssl ssl_accept sometimes does not return, causes server to hang permanently


rdp:
Requisite code sample:

```cr
require "socket"
require "openssl"

tcp_server = TCPServer.new(8081)

ssl_context = OpenSSL::SSL::Context::Server.new
ssl_context.certificate_chain = "cert.pem"
ssl_context.private_key = "_key.pem"
ssl_server = OpenSSL::SSL::Server.new(tcp_server, ssl_context)

puts "SSL Server listening on #{tcp_server.local_address}"
while true
    begin
      connection = ssl_server.accept
      connection.close rescue nil
    rescue ex
       print "unable to accept", ex
    end
end
```

I discovered that my server "in production", with similar code underneath, will sometimes hang and stop accepting connections seemingly randomly (incoming connections enter SYN_RECV [TCP backlog]) but no new connections are accepted by the crystal app.

My hypothesis from adding some `puts` calls is that it enters the call to `SSL_Accept` and never returns.  Turns out there there are several reads and writes within `SSL_Accept` so if you have a "rogue connection" (or malicious?) that does the TLS handshake "half way" (or even, just connects and does nothing), SSL_Accept won't exit (it's waiting on reads that never come).  since the servers are single threaded through the "wrapped accept" portion, this effectively hangs the  server.  This affects the HTTP server if it uses SSL, etc.  HTTP works fine.

Seemingly related discussion here: https://openssl-users.openssl.narkive.com/5QNG9OWX/ssl-accept-hang

AFAICT you can repro it by running the snippet above and connecting and sending nothing, ex: this:

```cr
require "socket"
TCPSocket.open("localhost", 8081) { |socket|
  puts "connected #{socket}, sleeping"
  sleep
}
```
After running both then all future connections to the server "aren't accepted", ex: `openssl s_client -connect localhost:8081` works before you run that, and afterward hangs on connect.

crystal 0.30.1

Ubuntu 19,04

rdp:
current workaround, for followers:

```
diff --git a/src/openssl/ssl/server.cr b/src/openssl/ssl/server.cr
index b31e53090..60b9eefb7 100644
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -68,6 +68,9 @@ class OpenSSL::SSL::Server
[...]

PR #9034, opened 2020-04-09
Add Log handler for HTTP::Server

straight-shoota:
Previously, errors while establishing a connection where just dumped out to STDERR.
Such errors are usually not very relevant but sometimes you might want to delve into details. The new Log system offers a great solution to bury these errors as debug messages, still making them available when logger level is low enough.

This creates a new dependency from `HTTP::Server` to `Log` that has not been there before. But I think that's okay. `HTTP::Server` is quite a high level API in stdlib terms and should often be used with a logging system anyways. And we would probably want to use the new Log system more often now in stdlib libraries.

In case you don't want that behaviour, overwriting `handle_exception` still allows inserting specific behaviour. But the default implementation should do a sane thing and not spill every failed SSL connection attempt to STDERR but still have a way to get access to these errors.

### DIFF
--- a/src/http/server.cr
+++ b/src/http/server.cr
@@ -5,6 +5,7 @@ require "./server/handler"
 require "./server/response"
 require "./server/request_processor"
 require "./common"
+require "log"
 {% unless flag?(:without_openssl) %}
   require "openssl"
 {% end %}
@@ -128,6 +129,8 @@ require "./common"
 # Reusing the connection also requires that the request body (if present) is
 # entirely consumed in the handler chain. Otherwise the connection will be closed.
 class HTTP::Server
+  Log = ::Log.for("http.server")
+
   @sockets = [] of Socket::Server
 
   # Returns `true` if this server is closed.
@@ -498,9 +501,9 @@ class HTTP::Server
     @processor.process(io, io)
   end
 
+  # This method handles exceptions raised at `Socket#accept?`.
   private def handle_exception(e : Exception)
-    e.inspect_with_backtrace STDERR
-    STDERR.flush
+    Log.debug(exception: e) { "Error while connecting a new socket" }
   end
 
   # Builds all handlers as the middleware for `HTTP::Server`.
--- a/src/http/server.cr
+++ b/src/http/server.cr
@@ -503,7 +503,10 @@ class HTTP::Server
 
   # This method handles exceptions raised at `Socket#accept?`.
   private def handle_exception(e : Exception)
-    Log.debug(exception: e) { "Error while connecting a new socket" }
+    # TODO: This needs more refinement. Not every exception is an actual server
+    # error and should be logged as such. Client malfunction should only be informational.
+    # See https://github.com/crystal-lang/crystal/pull/9034#discussion_r407038999
+    Log.error(exception: e) { "Error while connecting a new socket" }
   end
 
   # Builds all handlers as the middleware for `HTTP::Server`.


src/http/server.cr
```diff
@@ -498,9 +501,9 @@ class HTTP::Server
     @processor.process(io, io)
   end
 
+  # This method handles exceptions raised at `Socket#accept?`.
   private def handle_exception(e : Exception)
-    e.inspect_with_backtrace STDERR
-    STDERR.flush
+    Log.debug(exception: e) { "Error while connecting a new socket" }
```
RX14, line 506: I'd say this was more of an info-level or verbose warning personally.

I actually think the logger's default severity should be info or verbose, meaning we can add `debug` logging to the stdlib without it being on by default.

waj: Actually I think this has to be an error. That's exactly what it is.
BTW, why is handled in a method? it's only used once 🙂

straight-shoota: The purpose of the method is that sublasses can override it and alter the behaviour.

straight-shoota: Yes, it's an error. But it has nowhere to go from here. It just bubbles up to the current fiber's root and is printed to stderr. That's not useful at all. Do you have any better idea than sending it to `Log`? IMO this is a perfect use case.

I don't mind changing the severity to info, @RX14

bcardiff: Logging it as error it does not cause a bubbling. Since it’s a condition outside of what this method and class can control neither debug nor verbose are good fit I think.

If i were checking a server and something prevents connection to be accepted I think I would rather see it as an error than a debug.

RX14: It's not an error, or a warning, just because it's an exception. Socket errors happen all the time in the real world.

Neither is it debug info.

Verbose or info fit the best.

waj: No, I'm not saying the exception has to bubble up. Just saying it has to be logged as error. Please, use `Log.error` instead.
Socket errors happen all the time but in certain places we should log them. Here this error is actually quite unusual, so if it happens something more serious could be happening.

straight-shoota: Sorry, I misunderstood what you mean with being an error.

straight-shoota: I checked what others are doing: Apache and nginx both seem to emit info level log messages when TLS handshake fails.

waj: I don’t know the reasons for logging that as just info. Anyway here’s any exception raised at the accept call. It could be TLS handshake or anything else. Let’s log as error and if some day someone complains about too much logs we could evaluate how to discern the errors.

straight-shoota: Here, I'm already complaining. 🖐️ That's why I did this PR because the exceptions were overwhelming when running testssl.sh. Obviously that's a specific use case destined to cause such errors. And with logging it can at least be disabled. But generally, I want to see actual errors in my log, not a mess of connection failures.
For reference, that's the current default output when logging as error while running testssl.sh: 
https://gist.github.com/straight-shoota/59b7405edfe762a2fc6688e0e7c05f4c

waj: Woah! those are clients trying to connect to the server and failing to do so. I better want to see the reason in the log. Those ARE actual errors. If we want to separate (ignore) errors happening because client configuration we should do something to distinguish them first (maybe raise a different exception class from SSL classes?).

But I wont accept hide all errors by default. The default must be log any error. Period.

RX14: I don't want these exceptions hidden, they should be logged, by default. But I don't think they are errors, they are not fatal to the application, just a single connection. These random IO exceptions happen all the time, usually cause someone just navigated away on their browser while the page is loading.

waj: That's the problem, you say "I don't think they are errors". In fact, we don't know and the best guess is that they are. There is a `rescue` catching ALL exceptions around the `accept` call. So unless there is a demonstration that some exceptions can be ignored, we MUST log them all. Now if you tell me there is a really common exception raised under load that we shouldn't consider as error, then let's open a separate discussion and analyze that particular case. However, I have the impression that is not the case, since I have servers running with Crystal and I didn't see logs flooded with those kind of messages.

I'll give you one example that I just tested with nginx. When the system is low on file descriptors, new connections fail to connect and nginx logs this to the `error.log`:
```
2020/04/10 23:55:55 [crit] 82077#0: accept() failed (24: Too many open files)
```

Similar error can be seen with a simple HTTP server in Go:
```
2020/04/10 23:47:20 http: Accept error: accept tcp [::]:8080: accept: too many open files; retrying in 5ms
```

Those are errors. It's something a system administrator wants to get notified ASAP.

straight-shoota: You're right about different errors that could be raised from that method. What you mentioned should be logged as error. But what's typically caused by an individual misbehaving should not get the same attention level.
If my server fails to accept new connections because it's file descriptors are exhausted, I want to know. But I really don't care if a client fails to connects because it can't do TLS right. That would add extra noise and could cause the really noteworthy errors disappear in the mass.

ysbaddaden: Yes, it all depends on the error:

- Too many open files => that's a critical error happening on the server, we _can_ do something to fix it;
- Connection error => a client failed, maybe it tried HTTP on TLS port, closed the connection immediately, failed TLS handshake, or whatever, but there is _nothing_ the server can do. It's informational, at best.

straight-shoota: Considering how to implement this, the difficult part is discerning which category an error belongs to. That's not trivial because everything related to TLS is just `OpenSSL::SSL::Error`. We plan on changing some things here anyway, such as a generic `TLS` API and change the integration with `HTTP::Server` to do the handshake inside the client fiber.
So it's probably best to postpone solving the log severity assignment. I'll go for everything being logged as error for now and we can refine it with/after the TLS/server refactor.

waj: Thanks @straight-shoota. I agree 100% now.
Probably we need to subclass `OpenSSL::SSL::Error` or maybe just make the `OpenSSL::SSL::Socket::Server#accept` method to not return until a connection was made or a fatal error occur, and just log client errors from there as `info` or `debug` but with a different source.

Maybe Apache sources can give us a hint about what errors to consider here: https://github.com/apache/httpd/blob/add4e2892b4c6b35be790843aac5713302dc1d2f/modules/ssl/ssl_engine_io.c#L1410-L1453.

straight-shoota: I think such errors should not be handled by SSL implementation but the calling code. For an HTTPS server, it's usually okay to essentially ignore client errors. In other use cases that might need to be handled differently.
Alsoas per https://github.com/crystal-lang/crystal/issues/8108#issuecomment-531767712  `SSL::Socket::Server` needs to be removed from HTTP::Server implementation. Probably  also entirely.

waj: Well... the call to `SSL_accept` still has to happen somewhere, right? Aren't we just moving the problem somewhere else? But yes, having a `SSL::ClientError` subclass should be enough to be handled by servers, close everything and log as debug.

straight-shoota: Yes, we'd still have to sort out what's what. But it would prob be a bit easier when TCP accept and TSL accept don't happen in the same call.
(I was mainly commenting on your idea to make `SSL::Server#accept` block until it is successful and log every failure internally.)

PR #9037, opened 2020-04-10
Fix file_spec and related exception types on Windows

oprypin:


spec/std/file_spec.cr
```diff
@@ -313,7 +314,7 @@ describe "File" do
     File.extname("").should eq("")
   end
 
-  it "constructs a path from parts" do
+  pending_win32 "constructs a path from parts" do
```
straight-shoota, line 317: I would actually remove these specs because this is already tested extensively in `path_spec`.
One example can be kept to ensure `File.join` is working, but that would be a simple one that doesn't have any platform-specific differences.
```cr
it ".join" do
  File.join(["/foo/", "/bar"]).should eq("/foo/bar")
end
```
Ditto below for `.expand_path`. 
I can see this being a subsequent PR, if you don't want to tackle this directly. Inserting a TODO comment wouldn't hurth, though.

src/crystal/system/win32/file.cr
```diff
@@ -173,22 +177,22 @@ module Crystal::System::File
     end
 
     unless exists? real_path
-      raise ::File::Error.from_errno("Error resolving real path", Errno::ENOTDIR, file: path)
+      raise ::File::Error.from_errno("Error resolving real path", file: path)
```
straight-shoota, line 180: Shouldn't this use `from_winerror`?

oprypin: Umm yes so this is definitely incorrect, I see now. But the prior value was also incorrect, specs failed with it.

The thing is, the function could cause both errno and winerror, so either one would be wrong.
Maybe I just need to hardcode a different value?

PR #9038, opened 2020-04-10
Make File.rename overwrite the destination file on Windows, like elsewhere

oprypin:
The compiler, for example, relies on it here: https://github.com/crystal-lang/crystal/blob/0bb3fe98d0b2698e646ab6eb24097eb6e4e8bafa/src/compiler/crystal/compiler.cr#L702 But right now it throws an error that the file already exists.

You could argue whether it's desirable. For example, Python, to preserve backwards compatibility where the behavior differed like this for ages, decided to keep `rename` as replacing on Unix but erroring on Windows, but also added [`replace`](https://docs.python.org/3/library/os.html#os.replace) to replace on both. So on Unix they don't have any `rename` which would error out if the destination exists.

Another important caveat is that replacing a file like that is *atomic* on Unix, but on Windows that is impossible to achieve.

Let me know what you think, I could do the same as Python instead.

### DIFF
--- a/spec/std/file_spec.cr
+++ b/spec/std/file_spec.cr
@@ -516,6 +516,17 @@ describe "File" do
       end
     end
 
+    it "replaces a file" do
+      with_tempfile("rename-source.txt", "rename-target.txt") do |source_path, target_path|
+        File.write(source_path, "foo")
+        File.write(target_path, "bar")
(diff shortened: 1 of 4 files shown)


RX14:
it's windows-only, you only need one approval

PR #9043, opened 2020-04-11
Add Process.quote and fix shell usages in the compiler

oprypin:
Add a method to produce a command line string from a list of args.
Examples:
```crystal
files = ["my file.txt", "another.txt"]
`grep -E 'fo+' -- #{Process.quote(files)}`

Process.quote_posix(files) # => "'my file.txt' another.txt"

Process.quote_windows({ %q(C:\"foo" project.txt) }) # => %q("C:\\\"foo\" project.txt")
```
Yes, all of *these* examples could've been written better without `shell: true` but the reality is that sometimes one needs to intermix shell handling and non-shell handling, the best example of which is `link_flags` in the compiler, which I am also touching here (search for it in the diff!).

---

The source code of `Process.quote_windows` is **required** for the upcoming `Process.new` port on Windows.
Sure, it doesn't have to be *public* for *that* case, but I found another use case for it in the compiler (search for it in the diff!), so why not publish it.

I'm intentionally not hiding `quote_posix` in something like `Crystal::System::Process` because it can have more uses (in fact, I used it immediately in the compiler here). E.g. one might want to tell a user to run a Bash command specifically and make sure it's correct and copy-pasteable, and there's no reason that the program should not be able to do that on Windows.

Both `quote_posix` and `quote_windows` are available on all systems, then `quote` just selects for the current OS.

---

~~Removing the usage of `Process.new(, args, shell: true)` from the compiler is also a **requirement** for Windows porting, as this combination won't be supported.~~

Additionally, I did a quick search throughout the codebase for shell misuses which would surely fail eventually (with likely security vulnerabilities!) if any atypical path was involved, and applied the new method there. This just tends to happen more often on Windows, which made me notice.


src/process/shell.cr
```diff
@@ -0,0 +1,81 @@
...
+  # ```
+  def self.shell_quote_posix(args : Enumerable(String)) : String
+    args.join(' ') do |arg|
+      if arg.empty?
+        "''"
+      elsif arg =~ %r([^a-zA-Z0-9%+,\-./:=@_]) # not all characters are safe, needs quoting
+        "'" + arg.gsub("'", %('"'"')) + "'"    # %(foo'ba#r) becomes %('foo'"'"'ba#r')
+      else
+        arg
+      end
```
Sija, line 39:
```suggestion
      case arg
      when .empty?
        "''"
      when .matches? %r([^a-zA-Z0-9%+,\-./:=@_]) # not all characters are safe, needs quoting
        "'" + arg.gsub("'", %('"'"')) + "'"      # %(foo'ba#r) becomes %('foo'"'"'ba#r')
      else
        arg
      end
```

oprypin: why

Sija: - same indent for all branches
- `String#matches?`

oprypin: Applied `matches?`

oprypin:
Mmm no idea what's up with CI today. Surely some of these failures are not real

RX14:
I don't like the amount of `Process.shell_quote` used everywhere. It feels sort of like manual SQL escaping: super easy to miss one, and overly verbose. That was replaced with prepared queries, which is basically what `Process.new("command", args, shell: true)` does.

RX14:
github's CI is broken today, the others are fine.

oprypin:
> don't like the amount of Process.shell_quote used everywhere

Feel free to rework all of these cases. I'm not adding a new problem with this.

src/compiler/crystal/compiler.cr
```diff
@@ -344,17 +340,14 @@ module Crystal
           cc = CC
         end
 
-        if object_name
-          object_name = %('#{object_name}')
-        else
-          object_name = %("${@}")
```
oprypin, line 350: I suppose *here* I'm dropping the "advantage" of using `args`. That's the one case.

At the same time I'm removing an arbitrary code execution with a singular `object_name`

straight-shoota:
Looks good :+1:

@RX14 This might be a bit controversial, but we could consider calling `shell_quote` implicitly for string interpolation in `` ` `` literals.
That's probably the intended usage in most cases. Constructing shell commands with unquoted components would no longer work this way. For that you'd have to manually construct the command string first.
It would also only apply to `` ` `` literals, leaving similar methdods like `Process.run`, `system` etc. behave differently. This could cause unnoticed security issues when switching from "safe" `` ` `` literals to other methods because the same code would implicitly turn unsafe.

oprypin:
So the CI problem is real. Reproduced it locally with
```shell
make
CC="cc -fuse-ld=lld"  ./bin/crystal build --threads 1  --exclude-warnings spec/std --exclude-warnings spec/compiler -o .build/std_spec spec/std_spec.cr
```

Now the final linker command is passed all as one arg to the shell, and, being *162174* bytes long, it surpasses `MAX_ARG_STRLEN` (131072). This limitation is separate from `ARG_MAX` (2092454) for the sum of all lengths.

While this is cause to reconsider this approach, I'd like to point out that the limit on the commands is much much smaller on Windows so we probably want to limit this command's size anyway, for example by not linking so many separate files (almost 4000 in this case) at once.

oprypin:
I backed out of the change that caused the CI failure, and what I'm guessing is the main source of worry from @RX14 's comment

oprypin:
I realized that this escaping works only for CreateProcess, it's not good enough for the shell.
There actually doesn't seem to be any escaping that's always good enough for CMD, e.g. there's no way to insert a newline.

I renamed the methods accordingly and :nodoc:'d the Windows-specific one.

So I'll just keep this method reserved for use with `shell: true`, which on Windows I intend to not actually put through any shell. We can delay this PR.

oprypin:
So this change is no longer blocking Windows support, but it'd still be nice to have.

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+      quote_windows(args)
+    {% else %}
+      quote_posix(args)
+    {% end %}
+  end
+
+  # Shell-quotes one item, same as `quote({arg})`.
+  def self.quote(arg : String) : String
+    quote({arg})
+  end
```
Sija, line 24:
```suggestion
  # Shell-quotes given items, same as `quote(*args)`.
  def self.quote(*args : String) : String
    quote(args)
  end
```

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+      else
+        arg
+      end
+    end
+  end
+
+  # Shell-quotes one item, same as `quote_posix({arg})`.
+  def self.quote_posix(arg : String) : String
+    quote_posix({arg})
+  end
```
Sija, line 49:
```suggestion
  # Shell-quotes given items, same as `quote_posix(*args)`.
  def self.quote_posix(*args : String) : String
    quote_posix(args)
  end
```

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+  # :nodoc:
+  #
+  # Shell-quotes one item, same as `quote_windows({arg})`.
+  #
+  # ```
+  # Process.quote_windows(%q(C:\"foo" project.txt)) # => %q("C:\\\"foo\" project.txt")
+  # ```
+  def self.quote_windows(arg : String) : String
+    quote_windows({arg})
+  end
```
Sija, line 104:
```suggestion
  # :nodoc:
  #
  # Shell-quotes given items, same as `quote_windows(*args)`.
  #
  # ```
  # Process.quote_windows(%q(C:\"foo" project.txt)) # => %q("C:\\\"foo\" project.txt")
  # ```
  def self.quote_windows(*args : String) : String
    quote_windows(args)
  end
```

src/process/shell.cr
oprypin: You can be sure that I considered this option, it's not like I didn't know about it.

But I'm wary about the effect of proliferating these splats on the compilation times.

And I also want to stress the case of quoting *one* item as something that one may need specifically. If there are multiple items, it's nice to make the user think, is this a List situation or is it a Tuple situation.

I don't know. I don't feel that strongly about this though.

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+  # ```
+  # files = ["my file.txt", "another.txt"]
+  # Process.quote_posix(files) # => "'my file.txt' another.txt"
+  # ```
+  def self.quote_posix(args : Enumerable(String)) : String
+    args.join(' ') do |arg|
+      if arg.empty?
+        "''"
+      elsif arg.matches? %r([^a-zA-Z0-9%+,\-./:=@_]) # not all characters are safe, needs quoting
+        "'" + arg.gsub("'", %('"'"')) + "'"          # %(foo'ba#r) becomes %('foo'"'"'ba#r')
```
j8r, line 39:
```suggestion
        '\'' + arg.gsub('\'', %('"'"')) + '\''      # %(foo'ba#r) becomes %('foo'"'"'ba#r')
```

oprypin: Have you benchmarked this?

asterite: Could we please stop with these nonessential comments? These changes can be done after this PR is merged just fine, if really needed.

src/process/shell.cr
```diff
@@ -0,0 +1,81 @@
...
+  #
+  # ```
+  # files = ["my file.txt", "another.txt"]
+  # Process.shell_quote_posix(files) # => "'my file.txt' another.txt"
+  # ```
+  def self.shell_quote_posix(args : Enumerable(String)) : String
+    args.join(' ') do |arg|
+      if arg.empty?
+        "''"
+      elsif arg.matches? %r([^a-zA-Z0-9%+,\-./:=@_]) # not all characters are safe, needs quoting
```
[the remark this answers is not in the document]
oprypin: Not as far as I'm aware.

There's the command "`:`" but that's not affecting anything.
And oh huh, guess if one wanted to run an executable called "`%`" (which is in PATH) in Bash, they wouldn't be able to. That's Bash though, not POSIX.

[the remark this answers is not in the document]
straight-shoota: > the following may need to be quoted under certain circumstances. That is, these characters may be special depending on conditions described elsewhere in this volume of IEEE Std 1003.1-2001:
>
>   `*   ?   [   #   ˜   =   %`

https://pubs.opengroup.org/onlinepubs/007904875/utilities/xcu_chap02.html

It doesn't give more detail about the circumstances. So maybe it's okay.

oprypin:
Just found in the wild
https://github.com/crystal-lang/shards/search?q=escape

spec/std/process_spec.cr
```diff
@@ -321,4 +321,42 @@ describe Process do
...
+      it { Process.quote_posix(["my file.txt", "another.txt"]).should eq "'my file.txt' another.txt" }
+      it { Process.quote_posix(["foo ", "", " ", " bar"]).should eq "'foo ' '' ' ' ' bar'" }
+      it { Process.quote_posix(["foo'", "\"bar"]).should eq %('foo'"'"'' '"bar') }
+    end
+  end
+
+  describe "quote_windows" do
+    it { Process.quote_windows("").should eq %("") }
+    it { Process.quote_windows(" ").should eq %(" ") }
+    it { Process.quote_windows(orig = "%hi%").should eq orig }
```
straight-shoota, line 347: Is this really correct? Wouldn't `%hi%` expand an environment variable?

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+      end
+    end
+  end
+
+  # Shell-quotes one item, same as `quote_posix({arg})`.
+  def self.quote_posix(arg : String) : String
+    quote_posix({arg})
+  end
+
+  # :nodoc:
```
straight-shoota, line 51: Why is this method nodoc? Even if it might have less use cases, I think it should still be considered documented.

spec/std/process_spec.cr
oprypin:
```
  # NOTE: This is **not** safe to pass to the CMD shell.
```
https://github.com/crystal-lang/crystal/pull/9043#issuecomment-612642548

straight-shoota: Please add this to the method documentation!

straight-shoota: Hm, maybe I just didn't understand it correctly what's already there. So you can leave it as is unless someone else finds this should be more detailled.

oprypin:
Just found some independent "research" that matches my finding that it's impossible to safely quote for CMD shell.
https://github.com/actions/toolkit/blob/6a744be7ed9bab4b95ca7600529b7ce5ced1e54a/packages/exec/src/toolrunner.ts#L237

So it's good that I'm not attempting to solve that here. The quote_windows function here is still really useful to be available. Realized another example. Tools can sometimes accept the command line from a file instead. `cl.exe` is one. So if we were to refactor the code forming linker flags to keep them as an array until the end to pass them to the subprocess, we'd still have to explicitly use this function to convert them to a string here
https://github.com/crystal-lang/crystal/blob/2cbe65b0a13c8a2ce91bd4799f5b96b8e1953097/src/compiler/crystal/compiler.cr#L342
and this is not specific just to Crystal's implementation

bcardiff:
I like the idea of making the compiler escape interpolation between `` `#{expr}` ``, yet that requires first this PR and then the compiler logic in a separate release

If we eventually want to do that we would need in this PR a `ShellString` type to mark the safe strings. Making them safe to concat directly in `` `#{expr}` ``, otherwise they need to be escaped. This is similar to html template engines where the user is not forced to explicitly escape interpolated values. I haven't ported this kind of design to Crystal yet, not sure if there would be any surprised along the road.

I guess this will come handle here also https://github.com/crystal-lang/crystal/pull/8900#discussion_r391834158 

I find escape more appropiate than quote. Since it will quote if needed.

straight-shoota:
FYI: I've been using a [`SafeString`](https://github.com/straight-shoota/crinja/blob/ec715a4ece0d7e42a9970274e19e303d1e06e5b0/src/runtime/safe_string.cr) type with crinja. Works pretty well. Obviously it can be a bit confusing when you need to take care of strings that are not `String`. Maybe a `Stringlike` module could be an improvement to that at some point. But for now it's not that bad and in the context of shell escaping it would probably be restricted to a relative tight scope.

oprypin:
Well, I don't think this PR is blocked on any such future consideration.

I didn't want to name it "escape" because I don't feel it's appropriate for the multi-argument usage. Important to insert quotes in between them, not escaping the spaces.

I don't feel that strongly about it but maybe at least one more vote that it's good to change to `quote`.

oprypin:
I do appreciate the perspective that the action here is *escaping* but it just happens to be implemented through *quoting*.

I may also be biased by Python using the name "quote".

Still, though, for Windows this *must* be done by quoting. And the name "escape" may give a false sense of security that it's safe to use in the shell.

oprypin:
Please consider this for 0.35.0, as I think some follow-up ideas voiced here can't be done before there's a released compiler with this.
This can also be used in Shards nicely (and it needs Windows work too).
And it's a nice security fix in general.

straight-shoota:
... another conflict =)

oprypin:
The CI breakage isn't real, or at least that command line works locally.
Is it possible to retry it?

bcardiff:
I'm a bit lost with something. Why is `Process.quote` not used in `Crystal::System::Process.prepare_args` ? Wouldn't that be desired?

oprypin:
But it is used! in [src/crystal/system/win32/process.cr](https://github.com/crystal-lang/crystal/pull/9043/files#diff-bddb491649f0e834d93b55e9206da218)

Not for posix though, the concerns there are indeed separate.

oprypin:
For a fuller explanation:

A command can be passed either as a list of args or as one string (this is what the `Process` API exposes).
On POSIX the native format is a list of args. On Windows the native format is one string.

Because the recommended way to pass args is a list of args, we already have the implementation to convert them to a single string on Windows; this PR only moves it to a public location.

For POSIX there's no conversion required, it's already a list.

However, since we _do_ also have a way to specify a command as one string (for which on Windows the conversion is no-op), we need special support for that on POSIX. We defer that to the shell, and that implementation has been there since the beginning of Crystal.

The reason that the "quote" functions should be public, even on POSIX, is for the use case where one wants "the best of both worlds": conveniently specify an arbitrary string-based command, but also safely intersperse a list of arguments into it. That needs to be used by the caller beforehand, and is never needed by any internal implementation in `Process` on POSIX.

bcardiff:
Thanks for the clarification @oprypin

src/compiler/crystal/codegen/cache_dir.cr
```diff
@@ -119,7 +119,7 @@ module Crystal
         .sort_by! { |dir| File.info?(dir).try(&.modification_time) || Time.unix(0) }
         .reverse!
         .skip(10)
-        .each { |name| `rm -rf "#{name}"` rescue nil }
+        .each { |name| `rm -rf -- #{Process.quote(name)}` rescue nil }
```
bcardiff, line 122: Why is the `--` needed/introduced here?

oprypin: Oh. Well it's not directly related to the overall theme of quoting, but does fix a security-related issue. If there was a file named just `"-i"`, `rm` would see it as a flag, and the quoting wouldn't help either. I don't think it can actually happen here because cache names are controlled by Crystal, but still good practice to always add the double dash to indicate the end of flags.

waj: Why not just do `Dir.delete` instead?

oprypin: Unrelated change

waj: I know, so is it adding the `--` then, but I'm not saying you have to do it in this PR. Just asking if there is any reason this was made this way. You were doing the analysis of parts of the code that require `Process.quote` and it could be fixed in a different way instead.

oprypin: I could've split it into a separate commit but they're ignored under the current merging process.
I could've split it into a separate PR, but the process overhead (mainly, but not exclusively, due to GitHub) is too much.
So I just ignore it.

straight-shoota: So but we could/should have a PR that changes this to `Dir.delete`?

oprypin: Yes, it'd be good. But I am not interested in sending that out. Maybe the whole codebase could be swept for such things.

src/process/shell.cr
```diff
@@ -0,0 +1,105 @@
...
+  # ```
+  # files = ["my file.txt", "another.txt"]
+  # Process.quote_windows(files) # => %("my file.txt" another.txt)
+  # ```
+  def self.quote_windows(args : Enumerable(String)) : String
+    String.build { |io| quote_windows(io, args) }
+  end
+
+  private def self.quote_windows(io : IO, args)
+    args.join(' ', io) do |arg|
```
oprypin, line 68: Aaaa I missed the argument swap

oprypin: Should we fail CI on warnings? Or maybe we already do on non-Windows

asterite: No because there are tests that check that the deprecated methods still work.

oprypin: But we `--exclude-warnings spec/std`. Maybe it's not 100% reliable though.

straight-shoota: Yes, this should be improved. Not sure if it's not reliable or just not configured properly...

bcardiff: I think that the only reliable way is to fail on warnings on std_spec, and add deprecated_spec target were deprecations can happen. That way we can error on std_spec.

bcardiff: Handled in #9369

PR #9045, opened 2020-04-11
Compiler: turn exhaustive case warnings into errors

asterite:
This PR consists of two things:
1. if a case is not exhaustive it will now be an error instead of a warning
2. if a case is exhaustive, the `else` branch is unreachable (with type `NoReturn`) instead of `nil`

The second point means this:

```crystal
x = 1 || "a"
z = case x
when Int32
  true
when String
  false
end

typeof(z) # => Bool
```

Before this PR `typeof(z)` was `Bool | Nil`.

Now, when comparing against types there's no way to reach the `else` because the compiler proved that all cases are covered.

However, when comparing against enum values, it's possible that the `else` branch is reached if we use `SomeEnum.new(some_int_value)` and it falls outside of the valid enum members. If that's the case, the `else` branch will raise an `Exception` saying "Unhandled case: enum value outside of defined enum members".

I thought about making the message tell you which enum value was that, but it would make the code a lot more complex. And I don't want to make it more complex if the code might not be needed later on.

In the future, I'd like the language to prevent creating enum values outside of the declared members, unless the enum is marked in a special way, for example with an `@[OpenEnum]` annotation. In that case the compiler would also always require an `else` in such case. Then the `else` will be really unreachable.

But this can be done in a later PR after we decide we want that.

### DIFF
--- a/src/compiler/crystal/semantic/ast.cr
+++ b/src/compiler/crystal/semantic/ast.cr
@@ -807,4 +807,12 @@ module Crystal
       end
     end
   end
+
+  # Ficticious node to mean a location in code shouldn't be reached.
+  # This is used in the implicit `else` branch of a case.
+  class Unreachable < ASTNode
+    def clone_without_location
+      Unreachable.new
+    end
+  end
 end
--- a/src/compiler/crystal/semantic/cleanup_transformer.cr
+++ b/src/compiler/crystal/semantic/cleanup_transformer.cr
@@ -175,9 +175,67 @@ module Crystal
       @exhaustiveness_checker.check(node)
 
       if expanded = node.expanded
+        unless node.else
+          replace_unreacahble_if_needed(node, expanded)
+        end
+
         return expanded.transform(self)
       end
+
       node
+    end
+
+    # If any of the types checked in `case` is an enum, it can happen that
+    # the unreachable can be reached by doing `SomeEnum.new(some_value)`.
+    # In that case we replace the Unreachable node with `raise "..."`.
+    # In the future we should disallow creating such values unless the
+    # enum is marked as "open".
+    def replace_unreacahble_if_needed(node, expanded)
+      cond = node.cond
+      return unless cond
+
+      if cond.is_a?(TupleLiteral)
+        return unless cond.elements.all?(&.type?) &&
+                      cond.elements.any? { |element| has_enum_type?(element.type) }
+      else
+        cond_type = cond.type?
+        return unless cond_type && has_enum_type?(cond_type)
+      end
+
+      an_if = find_unreachable_parent(expanded)
+      unless an_if
+        node.raise "BUG: expected to find Unreachable node"
+      end
+
+      an_if.else = build_raise("Unhandled case: enum value outside of defined enum members", node)
+    end
+
+    def has_enum_type?(type)
+      if type.is_a?(UnionType)
+        type.union_types.any? &.is_a?(EnumType)
+      else
+        type.is_a?(EnumType)
+      end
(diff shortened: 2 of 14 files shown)


RX14:
I think a small breaking change to deprecate the `Enum.from_value*` methods and replace them with `Enum.new` (validates or raises), `Enum.new!` (same as current `.new`) and `Enum.new?` (same as current `from_value?`) would be good. Then add `OpenEnum` which affected these behaviours.

This is a good PR though, and I'd be fine merging it before any change to enums is made.

asterite:
The issue is that when returning enums from C, the value is not checked. This isn't as simple as tweaking the enum API, there needs to be some compiler changes.

Thats why I'd like to leave all of that as a separate discussion and PR.

straight-shoota:
I think we should merge the fix for all warnings in stdlib and compiler first before turning the warning into an error.

asterite:
@straight-shoota I believe I already did that in the PR that introduced the warnings?

asterite:
@straight-shoota Otherwise, CI would have failed because these are now errors.

straight-shoota:
Yeah right. I was still seeing tons of them, so probably just need a rebase.

asterite:
> so probably just need a rebase

This is in one of your branches, right? This PR seems to have no warnings related to `case`.

straight-shoota:
Yeah, I'm seeing warnings on a local branch. But it was indeed not correctly rebased on master, so all good :+1:

src/compiler/crystal/semantic/cleanup_transformer.cr
```diff
@@ -175,11 +175,69 @@ module Crystal
...
+
       node
     end
 
+    # If any of the types checked in `case` is an enum, it can happen that
+    # the unreachable can be reached by doing `SomeEnum.new(some_value)`.
+    # In that case we replace the Unreachable node with `raise "..."`.
+    # In the future we should disallow creating such values unless the
+    # enum is marked as "open".
+    def replace_unreacahble_if_needed(node, expanded)
```
n-rodriguez, line 193: Isn't it a typo?

src/compiler/crystal/semantic/ast.cr
```diff
@@ -807,4 +807,12 @@ module Crystal
       end
     end
   end
+
+  # Ficticious node to mean a location in code shouldn't be reached.
+  # This is used in the implicit `else` branch of a case.
+  class Unreachable < ASTNode
```
bcardiff, line 813: 😍

src/compiler/crystal/semantic/to_s.cr
```diff
@@ -79,5 +79,10 @@ module Crystal
       @str << node.value
       false
     end
+
+    def visit(node : Unreachable)
```
bcardiff, line 83: Shouldn't the to_s here remove the `else unreachable` entirerly to preserve better the semantics of the input code?

asterite: Yes, maybe. The thing is that to_s is never called in the compiler for unreachable, just for debug purposes.

asterite: What I mean is, the Unreachable node is generated when expanding case into if else. There's no else we can remove from the case because it's gone. If we remove the else from the if then we get something nilable. raise is the most similar translation.

bcardiff: Ok, i see. My worries is that some code instrumentation might loose exhaustive case check properties. But probably i am over worrying here.

PR #9047, opened 2020-04-11
Minimal Windows support for Process

oprypin:
Based on https://github.com/RX14/crystal/commit/ed14ad5ee4b0422200a18f63350ab83d90153f61


spec/std/process_spec.cr
```diff
@@ -3,54 +3,106 @@ require "process"
 require "./spec_helper"
 require "../spec_helper"
 
+private def exit_code_command(code)
+  if {{flag?(:win32)}}
```
Sija, line 7: Why not use macro all the way down?

```suggestion
  {% if flag?(:win32) %}
```

ditto elsewhere

RX14: Because it makes no difference and is nicer to work with.

Nitpicking style again, please stop.

asterite: I don't think this is nitpicking. The current way is generating code for both cases where it's really not needed.

Sija: @RX14 What you're doing has a name and it's called _bullying_ m8, I'd like @crystal-lang members to ensure that this kind of behaviour doesn't continue, thank you.

spec/std/process_spec.cr
```diff
@@ -3,54 +3,110 @@ require "process"
...
   it "receives arguments in tuple" do
-    Process.run("/bin/sh", {"-c", "exit 123"}).exit_code.should eq(123)
+    command, args = exit_code_command(123)
+    Process.run(command, args.as(Tuple)).exit_code.should eq(123)
   end
 
   it "redirects output to /dev/null" do
     # This doesn't test anything but no output should be seen while running tests
-    Process.run("/bin/ls", output: Process::Redirect::Close).exit_code.should eq(0)
+    if {{flag?(:win32)}}
```
Sija, line 91: ditto

spec/std/process_spec.cr
oprypin: @Sija, regardless of what we call this, I do think dubious nitpicks provide mostly negative value early on in a PR's lifecycle.

Sija: @oprypin First, I don't think it's irrelevant how you'll call it because it adds up to the erosion of OSS culture that we practice here, you're rude at the times as well but there are limits to this which IMO need to clearly stated.

Second, I don't see any _negative value_ as you called it - as I see it, this PR is not at its _early_ stage and if that's desired, please be clear about when you'd expect reviews. There are tools designed by GH for that like PR drafts for instance that are supposed to make this intention more clear, yet you don't use them. In such situation I can just say it's your ill intention to behave in such manner.

Third, these as you called _dubious nitpicks_ increase the amount of generated code for no reason other than your personal preference, so what's _dubious_ here is your attitude, not the nitpicks.

asterite: I think the point here is that Windows support is inexistent. We don't have a compiler for Windows. So pushing code in that direction is great. When there are small comments like this which are mostly about personal style but they don't affect what happens, it slows down the process.

I think right now, regarding Windows support, we should make all our effort to avoid getting into these small discussions. Instead, we should try to push more code.

Once Windows is finally supported everything is much simpler: those with Windows can try it, and we can slowly improve things from that point forward. But now it's not the time for that (as always, just my opinion).

Sija: @asterite I agree, what I don't understand and agree with is the aggressive attitude which helps no one. We're all here doing things from different perspectives, having different set of skills and approaches, yet IMO we all should strive to respect each other in the process, especially having a common goal in mind.

RX14: I think there's a consensus that small style nitpicks are largely noise unless there's an objective argument for one over the other. This is why the core team agreed to implement a style guide: to make style objective so that we don't get stuck with non-objective feedback.

So, from my perspective many of these comments are unnecessary and I'd like to discourage them. From previous conversations, I gathered this was the perspective of the core team too. I've talked about this with others in the past too. If I misunderstood, then I'm happy to stop.

Perhaps in this case it's a push to call it a style issue, but I think we can agree that both approaches are roughly equivalent.

oprypin: Honestly in this particular case I'd even concede that it was a *good* PR comment.

My prior statement was a general one. I worry about newcomers, who can't take these comments with a grain of salt. Can't imagine a worse start than fixing numerous nitpicks only to be told 2 days later that the whole approach is no-go.

Sija: > I think there's a consensus that small style nitpicks are largely noise unless there's an objective argument for one over the other. This is why the core team agreed to implement a style guide: to make style objective so that we don't get stuck with non-objective feedback.

@RX14 There's nothing about the style in here. Generating code for no reason is a pretty objective argument to me. And yes, specs matter too - why should we add a waiting time to already hefty spec suite?

RX14: I don't want to focus on this instance in particular, I have objective reasons for why I chose to write it this way as well, I just think it makes no measurable difference either way. I mentioned my subjective reasons because you asked, and *because* it is subjective I called it a style point and resolved it. Clearly others disagree! And that's my bad.

straight-shoota: > Can't imagine a worse start than fixing numerous nitpicks only to be told 2 days later that the whole approach is no-go.

Exactly. It doesn't have to be a total no-go, even a necessary refactoring makes the previous discussion completely futile.

@Sija I fail to see any offensive behaviour that could be considered bullying. @RX14 just asked you to stop a practice that has been previously experienced. I can understand that you might feel repeatedly and undeservingly critizised. But you seem to have a habit on posting mostly style-related suggestions to new PRs. Most other contributers and reviewers consider that inadequate (and I'm sure you are aware of that).
I'm annoyed so much that I tend to pay no attention to them anymore. At that point it doesn't really matter if a concrete case is an objectively justified suggestion because we already classify it as "another one of those".

So to get along better in the future, I'd ask you @Sija to avoid suggesting non-essential code improvements when it's not the time for such details. If you need guidance on what's considered appropriate, let me know. I'm sure that would improve the culture of discussion.

Just to let you know: I often encounter code that just begs for stylistic improvement but isn't at the point of discussing that. In case it's something that might be relevant later, there's a tool that might help: I use to add a short review comment but don't commit the review and instead write a regular comment on the PR. Github will save the review comment so I can come back if the issue prevails and might need to be brought to attention later in the review process.

spec/std/process_spec.cr
oprypin: See https://github.com/crystal-lang/crystal/pull/9047/commits/dffe05a0fee9035ece309238820034b63729cb3d and its CI run.
It doesn't compile like that.
So I did it another way, not sure if it's as nice

Sija: Looks fine, thank you ❤️

spec/std/process_spec.cr
```diff
@@ -155,7 +153,11 @@ describe Process do
 
   it "sets working directory" do
     parent = File.dirname(Dir.current)
-    command = {{flag?(:win32)}} ? "cmd /c echo %cd%" : "pwd"
+    command = {% if flag?(:win32) %}
+                "cmd /c echo %cd%"
```
straight-shoota, line 157:
```suggestion
                "cmd.exe /c echo %cd%"
```
?

oprypin: Sure, why not

oprypin:
Sigh, CI infra failure

PR #9049, opened 2020-04-12
Implement host_flag? macro method, not affected by cross-compilation

oprypin:
The `flag?` macro tells you information about the target architecture, but one may need to run a command at compile time (so *on the host*), and it would be a different command depending on which OS it is.

Example:
```diff
diff --git a/src/compiler/crystal/config.cr b/src/compiler/crystal/config.cr
index 8b756583e..38215b7c8 100644
--- a/src/compiler/crystal/config.cr
+++ b/src/compiler/crystal/config.cr
@@ -32,7 +32,9 @@ module Crystal
     end
 
     def self.date
-      time = {{ (env("SOURCE_DATE_EPOCH") || `date +%s`).to_i }}
+      time = {{ (env("SOURCE_DATE_EPOCH") || (
+                  host_flag?(:win32) ? `powershell Get-Date -Millisecond 0 -UFormat %s` : `date +%s`
+                )).to_i }}
       Time.unix(time).to_s("%Y-%m-%d")
     end
```


asterite:
This makes me wonder if we should be computing all of those things with macros.

That said, this needs to also be added tui the macro docs (search for def flag? and you'll find it).

oprypin:
The bad thing here isn't really the macros but the shelling out :)
Each instance of that is technical debt, and we'll be scratching our heads now that Windows is relevant.
Definitely the most fun instance of this is ``ldflags: "`stuff`"``
https://github.com/crystal-lang/crystal/blob/f4b01c807276804fd3f52a5e2316f4b6676ed004/src/llvm/lib_llvm.cr#L10 See, no macros there :)

oprypin:
> needs to also be added to the macro docs

I realized it later, but that also made me think, maybe we should keep it undocumented for now. I don't know if we'll want to keep it, in this shape, or at all.

First off, I'm thinking to resolve that particular example above with `macro run` instead.

Then I also noticed that some important use cases are missing with this data representation. E.g. what if I want to check whether host OS differs from target OS? Should I go through every possible pair of flags and check if they're not equal?

So yeah, I'm mostly hedging my bets at this point, because this will need a Crystal release before it can be used.

waj:
@oprypin could you please fix the conflict so we can merge this? Thanks!

src/compiler/crystal/macros/methods.cr
```diff
@@ -145,10 +145,18 @@ module Crystal
     def interpret_flag?(node)
       if node.args.size == 1
         node.args[0].accept self
-        flag = @last.to_macro_id
-        @last = BoolLiteral.new(@program.has_flag?(flag))
+        flag_name = @last.to_macro_id
+        flags = case node.name
```
straight-shoota, line 149: Minor improvement: `flags` could be passed as an argument to this method, so we don't need to compare the node name again here. Instead, make it two separate branches in the already existing `case`.

src/compiler/crystal/semantic/flags.cr
```diff
@@ -11,6 +12,10 @@ class Crystal::Program
     @flags ||= flags_for_target(codegen_target)
   end
 
+  def host_flags
+    @host_flags ||= flags_for_target(Config.default_target)
```
straight-shoota, line 16: I'd also suggest to rename `Config.default_target` to `Config.host_target` because that's what it is.

src/compiler/crystal/macros/methods.cr
oprypin: No, just look at the code around it, it's a very strong pattern

src/compiler/crystal/semantic/flags.cr
oprypin: OK, didn't turn out *too* big.

src/llvm.cr
```diff
@@ -73,7 +73,7 @@ module LLVM
     LibLLVM.is_multithreaded != 0
   end
 
-  def self.default_target_triple : String
+  def self.host_target_triple : String
```
straight-shoota, line 76: I only meant `Config.host_target`.
IMO this should stay `LLVM.default_target_triple` because it' just wraps `LibLLVM.get_default_target_triple`.

oprypin:
Woops, pushed a bad commit previously, should be good now

PR #9052, opened 2020-04-12
Compiler: don't merge union types if there's a type annotation

### LINKED ISSUE #9050: [RFC] Don't merge unions into parent type when explicitly said so
Right now given a class hierarchy like this:

```crystal
class Base; end
class One < Base; end
class Two < Base; end
class Three < Base; end
```

If you do:

```crystal
One.new || Two.new
```

the type you get is `Base`, meaning `Base` or any of its subtypes.

The reason this is done like this is to prevent creating many union combinations: One | Two, One | Three, Two | Three, etc. Imagine with a type hierarchy as big as the AST nodes of the compiler.

However, it would be nice if you could use a subset of types in instance and class var declarations for a better type safety.

For example, in the compiler there's the `TypeDeclaration` AST node which represents `x : Int32`, `@x : Int32`, `@@x : Int32`, `$x : Int32` (for C library globals). The only possible types of the left-hand side are: `Var`, `InstanceVar`, `ClassVar` or `Global`. However, there's currently no way to represent that because it will unify to `ASTNode`.

# Proposal

If an instance variable is declared to be a union type, don't merge the types. So for example:

```crystal
class Foo
  @x : One | Two

  def initialize(@x)
  end

  def x
    @x
  end
end

foo = Foo.new(One.new)
typeof(foo.x) # => One | Two

# No warnings/errors: all types are covered
# (currently it will error because foo.x will be of type Base)
case foo.x
when One
when Two
end
```

However, the type union is lost when merged with other types... even with the same union type:

```crystal
typeof(foo.x || foo.x) # => Base
```

But I believe that at least having guarantees around the type of an instance var is good because then you can `case` over it without having to write an `else raise "Can't happen"`.

Also, if you have a `Base` and you want to assign it to something of type `One | Two` you'd have to do `foo.x = base.as(One | Two)`. It's a bit annoying, but I believe it's better to do it on the write side than always check it on the read side.

Thoughts?

asterite:
Fixes #9050

It would be nice to test the ecosystem if this goes through because it's a big change.

### DIFF
--- a/src/compiler/crystal/semantic/type_merge.cr
+++ b/src/compiler/crystal/semantic/type_merge.cr
@@ -166,7 +166,7 @@ module Crystal
       # Sort by depth so that types deep in the hierarchy will combine with
       # types we find first.
       if merge_to_parent == false
-        types = types.sort_by!(&.depth)
+        types.sort_by!(&.depth)
       end
 
       types.each do |t2|
--- a/src/compiler/crystal/semantic/type_merge.cr
+++ b/src/compiler/crystal/semantic/type_merge.cr
@@ -2,7 +2,7 @@ require "../program"
 
 module Crystal
   class Program
-    def type_merge(types : Array(Type?))
+    def type_merge(types : Array(Type?)) : Type?
       case types.size
       when 0
         return nil
@@ -20,7 +20,7 @@ module Crystal
       combined_union_of compact_types(types)
     end
 
-    def type_merge(nodes : Array(ASTNode))
+    def type_merge(nodes : Array(ASTNode)) : Type?
       case nodes.size
       when 0
         return nil
@@ -67,8 +67,26 @@ module Crystal
       end
     end
 
-    def type_merge_union_of(types : Array(Type))
-      union_of compact_types(types)
+    def type_merge_union_of(types : Array(Type)) : Type?
+      combined_union_of compact_types(types), merge_to_parent: false
+    end
+
+    def type_merge_union_of(nodes : Array(ASTNode)) : Type?
+      case nodes.size
+      when 0
+        return nil
+      when 1
+        return nodes.first.type?
+      when 2
+        # Merging two types is the most common case, so we optimize it
+        first, second = nodes
+        did_merge, merged_type = type_merge_two(first.type?, second.type?)
+        return merged_type if did_merge
+      else
+        # union_of
+      end
+
+      combined_union_of compact_types(nodes, &.type?), merge_to_parent: false
     end
 
     def compact_types(types)
@@ -111,25 +129,54 @@ module Crystal
       # Nothing to do
     end
 
-    def combined_union_of(types : Array)
+    def combined_union_of(types : Array, merge_to_parent = true)
       case types.size
       when 0
         nil
       when 1
         types.first
       else
-        combined_types = type_combine types
+        combined_types = type_combine types, merge_to_parent: merge_to_parent
         union_of combined_types
       end
     end
 
-    def type_combine(types)
+    # If merge to parent is true, types will be merged to common ancestors.
+    # If false, types will only be merged to a common ancestor if any of those
+    # types is the common ancestor. For example:
+    #
+    # ```
+    # class Base; end
+    #
+    # class One < Base; end
+    #
+    # class Two < Base; end
+    # ```
+    #
+    # Merging One | Two:
+    # - merge_to_parent = true: Base+
+    # - merge_to_parent = false: One | Two
+    #
+    # Merging Base | One:
+    # - merge_to_parent = true: Base
+    # - merge_to_parent = false: Base+
+    def type_combine(types, merge_to_parent = true)
       all_types = [types.shift] of Type
+
+      # Sort by depth so that types deep in the hierarchy will combine with
+      # types we find first.
+      if merge_to_parent == false
+        types = types.sort_by!(&.depth)
+      end
 
       types.each do |t2|
         not_found = all_types.all? do |t1|
           ancestor = t1.common_ancestor(t2)
-          if ancestor
+          if ancestor &&
+             (
+               merge_to_parent ||
+               (ancestor.virtual_type == t1.virtual_type || ancestor.virtual_type == t2.virtual_type)
+             )
             all_types.delete t1
             all_types << ancestor.virtual_type
             false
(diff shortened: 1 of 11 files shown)


src/compiler/crystal/semantic/type_merge.cr
```diff
@@ -111,25 +129,54 @@ module Crystal
...
+    # Merging Base | One:
+    # - merge_to_parent = true: Base
+    # - merge_to_parent = false: Base+
+    def type_combine(types, merge_to_parent = true)
       all_types = [types.shift] of Type
 
+      # Sort by depth so that types deep in the hierarchy will combine with
+      # types we find first.
+      if merge_to_parent == false
+        types = types.sort_by!(&.depth)
```
Sija, line 169:
```suggestion
        types = types.sort_by(&.depth)
```

asterite: Oh, it should just sort in place. Thanks!

Sija: I don't think so, since you shouldn't modify the method argument.

RX14:
There was a change last time and we reverted it because of one project's really bad compile-time regression

could we test that project again?

src/compiler/crystal/semantic/type_merge.cr
asterite: I do think so, because I know it's safe to mutate the argument and I don't want to allocate more memory.

asterite:
> There was a change last time and we reverted it because of one project's really bad compile-time regression
> could we test that project again?

We could, but it's different. The old change made all unions not to unify into their parent. This time it's controlled: only when you specify an instance/class var as a union. Give that the feature doesn't exist yet, nobody is doing that, so the only way to try this is to modify existing projects to use the feature and try it with a compiled compiler. It's not straight-forward.

I tried it on the compiler, by making `TypeDeclaration#@var` to be a union. Here are the times:
```
Before:
63.98s user 4.39s system 204% cpu 33.353 total
After + no changes to code:
64.01s user 4.32s system 191% cpu 35.649 total
After + changes:
65.57s user 4.22s system 203% cpu 34.226 total
```

The times are similar. There are variances but I don't think it's because of the changes but rather what the computer was doing (I'm listening to music, doing other stuff).

I think this shouldn't affect performance.

bcardiff:
I found it confusing that the semantics will change depending on how things are written and not what values are involved. This can lead to hard bugs in case the expressions are not preserved along the way.

If I would need to tag some times as in this examples I would use modules.

```crystal
module Foo; end

class Base; end

class One < Base
  include Foo
end

class Two < Base
  include Foo
end

class Three < Base; end

m = Two.new.as(Foo)

x = case m
    when One
      1
    when Two
      2
    end
```

But the new exhaustive case does not cover this, forcing a `when Foo` or `else`. If this is handled, what would be the downsides of this approach?

asterite:
> But the new exhaustive case does not cover this, forcing a when Foo or else. If this is handled, what would be the downsides of this approach?

The downside is that if this is library code, and someone uses your library and adds a new type `Foo`, the code will stop compiling. And given that this is in a library code, there's no way to fix it.

Exhaustiveness check can only be done if the types are guaranteed to be a determined set. With type hierarchy or included modules, this is not the case.

In Ruby's sorbet there's the concept of [sealed](https://sorbet.org/docs/sealed), which means only types defined directly in that module right there are the only possible types, and these can't be extended. In Crystal we could do something similar:

```crystal
@[Sealed]
module Foo
  # I think Base is not needed at all for this or your example?
  class Base; end

  class One < Base
    include Foo
  end

  class Two < Base
    include Foo
  end
end

foo = Foo::Two.new.as(Foo)

# compiles just fine
case foo
when Foo::One
when Foo::Two
end

module Foo
  class Three
    include Foo # error: Foo is sealed
  end
end
```

asterite:
> This can lead to hard bugs in case the expressions are not preserved along the way.

This is one of the problems with this PR. However, I prefer the compiler telling me "can't assign ASTNode to Var | InstanceVar | ClassVar", which is kind of easy to understand (ASTNode can be anything), and adding `as(Var | InstanceVar | ClassVar)` at write time, than having to add an `else raise "unreachable"` every time the value is read. This will be specially useful for shards, where values will be constructed by a shard but consumed by others.

bcardiff:
I think that a module flagged as a `:nodoc:` can be used as a sealed definition. It is not intended to be used, hence removing the scenario of what-if-user-does-X with the module.

With that + handling the exhaustive case for modules I think is fine.

There is always some room for unintended usage of some language construct. Weighting that and the change in semantics depending on how things are written and not what values are involved; I think I prefer the former.

---

NB: In general like the idea of sealing some definitions. Maybe because some familiarity with it from C#. But with the monkey patching it kind of looses the applicability maybe.

straight-shoota:
I would not derive any semantic meaning from comments. That includes documentation and `:nodoc:`. 
If we want some way to express a sealed type, it should be done with a proper annotation.

asterite:
> I think that a module flagged as a :nodoc: can be used as a sealed definition

I'm not sure. In the recent example of `Process::Status`, sealed would have come handy to say: I want a type hierarchy, but these are the only possible types (not that we'll merge that, but still an example).

Another example: ASTNode in the compiler. We know all the AST nodes. We use a hierarchy to share common functionality. But the hierarchy is sealed. And ASTNode is far from `:nodoc:`.

I think this is pretty easy to implement: if you use `@[Sealed]` on top of a type, the only definitions can happen inside that same file. That's a bit nicer than in Sorbet because in Sorbet you are forced to put children inside the parent namespace.

Then, if something is `@[Sealed]` we know all the types and can do exhaustive check.

It's a very, very simple change.

But I don't think it's related to this PR. This PR is about nor merging union types and not losing information about an instance var by going to the parent type. The sealed discussion is related to exhaustive case.

asterite:
@bcardiff Just to be clear, you can't check exhaustiveness if the type is, say, `ASTNode` because anyone can inherit it. This PR makes it so you can have a variable of type `Var | InstanceVar`, something that you couldn't do before. In that case the compiler can check exhaustiveness: it must be either `Var` or `InstanceVar`. Of course these types can be extended, but if `Var` and `InstanceVar` are covered, all subtypes are covered too. This is what the compiler will check.

beta-ziliani:
I'm reviving this one for fun. I think we can give it another shot.

HertzDevil:
Here is the discussion that prompted this: https://forum.crystal-lang.org/t/reducible-unions-from-type-filters/4525

There it is mentioned that those unmerged union types may still appear as the result of type filtering on `#is_a?` or `#responds_to?`:

```crystal
class A; end
class B1 < A; end
class B2 < A; end
class B3 < A; end

x = B1.new || B2.new

if x.is_a?(B1) || x.is_a?(B2)
  typeof(x) # => (B1 | B2)
end

if x.is_a?(B1 | B2)
  # since 1.2.0 (#11176)
  typeof(x) # => (B1 | B2)
end
```

This PR, I assume, extends this to arbitrary type expressions (such as for instance variables), and ensures that no variant types inside a union are subtypes of other variant types, whereas `||` still merges unions because no type declarations are possible there. Am I right?

Does the `Union` generic merge?

asterite:
> Do instances of the Union generic merge?

I can't remember how it behaves in this PR, but if you have `@x : Union(A, B)` then I think it should remain as `A | B` and not go to the parent.

Was that the question?

asterite:
This wasn't liked, so closing.

PR #4349, opened 2017-04-29
Add a simple method to run a block every X time

meew0:
I haven't received a response to #4255 yet, so I'm not sure if this method actually belongs in the standard library, but this is how I'd implement it, and I'd certainly use this method a lot in my code. Closes #4255.

PR #9053, opened 2020-04-12
Add #every top-level method

dscottboggs:
Just a suggestion, I just wrote this method in my code and thought it might be useful if it were in the stdlib.

Example run in the playground:

https://carc.in/#/r/8voj

### DIFF
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -104,6 +104,20 @@ def loop
   end
 end
 
+# Repeatedly executes the block, but with the given delay between executions.
+#
+# ```
+# every 1.second do
+#   puts Time.local
+# end
+# ```
+def every(delay : Time::Span) : Nil
+  loop do
+    yield
+    sleep delay
+  end
+end
+
 # Reads a line from `STDIN`.
 #
 # See also: `IO#gets`.


jhass:
Should this measure run time of the block and substract it from the sleep time?

z64:
For reference, this was previously submitted here: https://github.com/crystal-lang/crystal/pull/4349

dscottboggs:
> Should this measure run time of the block and substract it from the sleep time?

I thought of that, but there's the concern that the executed block would take longer than the given delay. I thought it better to keep things simple and make the delay constant. This way, you don't run the risk of turning a call to some method which also retries for the given delay into an infinite fast loop and stressing some other resource down the line.

I'm open to debate on the topic though, perhaps it should just fall back on being a fast loop if the block takes longer than the delay

dscottboggs:
> For reference, this was previously submitted here: #4349

Thanks for pointing that out. To quote @asterite in that thread

> I don't think this belongs in the std.

I respectfully disagree -- it seems too common of a use case to leave to a 3rd party dependency. I don't really see the downside to including this four-liner alongside others like `loop`, `future` and `printf`

z64:
Personally, I don't feel strongly about it. I think the argument over adding a small method to std goes both ways. That said, I believe the implementation from #4349 is preferable.

straight-shoota:
Comparing both #4349 it should be evident that this is actually not a simple feature. There are even more variants to that pattern that just those two implementations. For example, what happens when the execution time of the block exceeds the time span? What happens if it raises? Is the scheduler supposed to be strictly based on elapsed time (`Time.monotonic`) or should it take wall clock into account (`Time.local`)?

I'd excpect many use cases would have different needs than others and each one requires some customizable or specialized solutions. This all seems to much for stdlib and is better solved in dedicated shards (such as [tasker](https://shardbox.org/shards/tasker) for example).

@dscottboggs Since the implementation proposed here is really trivial, it should be easy to just keep that method local to your app.

PR #9054, opened 2020-04-13
Miscellaneous Windows-related fixes in the compiler

oprypin:
* Temporary file needs .exe extension to be executable
* Replacing '/' is not enough to sanitize a filename, use Path splitting and replace the drive ':' manually
* Split PATH by the correct delimiter
* Don't try to fork (n_threads=1)
* Don't try to set a directory's mtime on Windows, it errors
* ~~Don't rely on an external `date` command to get current date, use macro run instead~~
* Remove custom abs-path code that is already handled by expand_path
* Correct string interpolation in ECR
* Allow overriding `llvm-config --targets-built` because the host LLVM doesn't necessarily match target LLVM
* Skip signal code

Unrelated: handle non-exit non-signal exit statuses correctly on POSIX.

### DIFF
--- a/src/compiler/crystal/codegen/link.cr
+++ b/src/compiler/crystal/codegen/link.cr
@@ -81,7 +81,7 @@ module Crystal
     end
 
     class_getter paths : Array(String) do
-      default_path.split(':', remove_empty: true)
+      default_path.split(Process::PATH_DELIMITER, remove_empty: true)
     end
   end
 
--- a/src/compiler/crystal/crystal_path.cr
+++ b/src/compiler/crystal/crystal_path.cr
@@ -13,7 +13,7 @@ module Crystal
     @crystal_path : Array(String)
 
     def initialize(path = CrystalPath.default_path, codegen_target = Config.default_target)
-      @crystal_path = path.split(':').reject &.empty?
+      @crystal_path = path.split(Process::PATH_DELIMITER).reject &.empty?
       add_target_path(codegen_target)
     end
 
@@ -144,7 +144,6 @@ module Crystal
     end
 
     private def make_relative_unless_absolute(filename)
-      filename = "#{Dir.current}/#{filename}" unless filename.starts_with?('/')
       File.expand_path(filename)
     end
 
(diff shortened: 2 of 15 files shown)


src/compiler/crystal/codegen/link.cr
```diff
@@ -81,7 +81,7 @@ module Crystal
     end
 
     class_getter paths : Array(String) do
-      default_path.split(':', remove_empty: true)
+      default_path.split(Process::PATH_DELIMITER, remove_empty: true)
```
RX14, line 84: Is there a reason to make `CRYSTAL_PATH`'s format platform-dependent?

src/compiler/crystal/crystal_path.cr
```diff
@@ -13,7 +13,7 @@ module Crystal
     @crystal_path : Array(String)
 
     def initialize(path = CrystalPath.default_path, codegen_target = Config.default_target)
-      @crystal_path = path.split(':').reject &.empty?
+      @crystal_path = path.split(Process::PATH_DELIMITER).reject &.empty?
```
RX14, line 16: ditto

src/compiler/crystal/codegen/link.cr
oprypin: Well yeah, otherwise you literally can't specify any absolute path (C:)

RX14: Oh, right...

:( platform-consistency

oprypin:
This CL is ready, it's not waiting on any macro additions.
I can undo the change to `macro run`, it's not absolutely necessary.

straight-shoota:
Thank you @oprypin

src/compiler/crystal/crystal_path.cr
```diff
@@ -144,7 +144,6 @@ module Crystal
     end
 
     private def make_relative_unless_absolute(filename)
-      filename = "#{Dir.current}/#{filename}" unless filename.starts_with?('/')
       File.expand_path(filename)
     end
```
Sija, line 148: This could easily be removed altogether.

straight-shoota: Care to send a PR?

Sija: Sure, will do.

Sija: #9095

PR #9062, opened 2020-04-13
Pass the command line to cl.exe through a file

oprypin:
This is necessary because large projects (like the Crystal compiler) surpass the maximal size for the command line when linking the numerous object files.

This PR also makes Crystal actually be able to run the linker subprocess on Windows, as it replaces the unsupported combination of `args` and `shell: true`.

This PR is technically independent from #9043, but clearly some shell-escaping would be very helpful here.

### DIFF
--- a/src/compiler/crystal/compiler.cr
+++ b/src/compiler/crystal/compiler.cr
@@ -314,7 +314,11 @@ module Crystal
 
       target_machine.emit_obj_to_file llvm_mod, object_name
 
-      stdout.puts linker_command(program, [object_name], output_filename, nil)
+      print_command(*linker_command(program, [object_name], output_filename, nil))
+    end
+
+    private def print_command(command, args)
+      stdout.puts command.sub(%("${@}"), args && args.join(" "))
     end
 
     private def linker_command(program : Program, object_names, output_filename, output_dir)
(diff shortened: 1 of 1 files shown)


src/compiler/crystal/compiler.cr
```diff
@@ -314,22 +314,29 @@ module Crystal
...
 
-        %(#{CL} #{object_name} "/Fe#{output_filename}" #{program.lib_flags} #{link_flags})
+        args = %(#{object_names.join(" ")} "/Fe#{output_filename}" #{program.lib_flags} #{link_flags})
+        cmd = "#{CL} #{args}"
+
+        if cmd.to_utf16.size > 32000
+          # The command line would be too big, pass the args through a UTF-16-encoded file instead.
+          # TODO: Do this the proper way when iconv is supported.
+          cmd_filename = "#{output_dir}/link_command.cmd"
+          cmd_16 = "\ufeff#{args}".to_utf16
```
RX14, line 333: What is the `\ufeff`?

oprypin: [Byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-16)

jhass: [BOM](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-16)  :)

RX14: Can we get a comment explaining this?

src/compiler/crystal/compiler.cr
```diff
@@ -314,22 +314,29 @@ module Crystal
...
           link_flags = "/link #{link_flags}"
         end
 
-        %(#{CL} #{object_name} "/Fe#{output_filename}" #{program.lib_flags} #{link_flags})
+        args = %(#{object_names.join(" ")} "/Fe#{output_filename}" #{program.lib_flags} #{link_flags})
+        cmd = "#{CL} #{args}"
+
+        if cmd.to_utf16.size > 32000
+          # The command line would be too big, pass the args through a UTF-16-encoded file instead.
+          # TODO: Do this the proper way when iconv is supported.
```
straight-shoota, line 331: Maybe a minor thing, but this comment insinuates that we can actually get `iconv` to work on win32 and be confident to use it in the compiler 🤷 Even then, this implementation might actually be "the proper way" we should be doing this because it doesn't need a library.
So I would suggest to put this code already in a dedicated method on `String`.

oprypin: Removed mention of "iconv".
Probably too early to think of such a method. Likely to be the only use case, and the proper API for it already exists separately, so I probably disagree.

PR #9063, opened 2020-04-14
Set default values directly on record ivars

### LINKED ISSUE #9008: Allow `record` to support default values when using *::Serializable
https://play.crystal-lang.org/#/r/8u3w

```cr
require "json"
require "yaml"
 
yaml = <<-YAML
---
id: 1
YAML
 
json = <<-JSON
{"id":1}
JSON
 
record Test, id : Int32, active : Bool? = false do
  include JSON::Serializable
  include YAML::Serializable
end
 
 
pp Test.from_yaml yaml # => Test(@active=nil, @id=1)
pp Test.from_json json # => Test(@active=nil, @id=1)
```

I would expect this to set `active` to false if the property is missing, however it gets set to `nil`.  Similarly, if the type of the property is not nillable, an exception saying the property is missing is raised versus just using the default value.

Blacksmoke16:
Resolves #9008

### DIFF
--- a/spec/std/record_spec.cr
+++ b/spec/std/record_spec.cr
@@ -19,6 +19,12 @@ end
 
 private record Sub < Base, x : Int32
 
+private record CustomInitializer, id : Int32, active : Bool = false do
+  def initialize(*, __id id : Int32)
+    @id = id
+  end
+end
+
 describe "record" do
   it "defines record with type declarations" do
     ary = [2, 3]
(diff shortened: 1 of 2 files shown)


spec/std/record_spec.cr
```diff
@@ -19,6 +19,12 @@ end
 
 private record Sub < Base, x : Int32
 
+private record CustomInitializer, id : Int32, active : Bool = false do
```
RX14, line 22: What about just `record Foo, active = false`.

Blacksmoke16: Wanted to replicate what was happening within the original issue; which would be using a custom override, setting one value, but leaving the other to be `false`.

RX14: Maybe I should have been more clear, but what I was asking for was a spec that

```cr
private record CustomInitializer, id : Int32, active = false
```

works, as well as the record you have above.

PR #9064, opened 2020-04-14
Super-fancy refactor of Process::Status

oprypin:
This uses a hierarchy of structs; it's a type-safe approach to the variety of POSIX exit statuses.

Based on https://github.com/crystal-lang/crystal/issues/8381#issuecomment-612487955, https://github.com/crystal-lang/crystal/issues/8381#issuecomment-612691909

[Usage examples](https://github.com/crystal-lang/crystal/pull/9064#issuecomment-613693961)

It is also backwards-compatible

I don't know how this will be received, but it doesn't matter, I had to put this out there :joy:

### DIFF
--- a/src/crystal/system/process.cr
+++ b/src/crystal/system/process.cr
@@ -16,7 +16,7 @@ struct Crystal::System::Process
   # def pid : Int
 
   # Waits until the process finishes and returns its status code
-  # def wait : Int
+  # def wait : ::Process::Status
 
   # Whether the process is still registered in the system.
   # def exists? : Bool
--- a/src/crystal/system/unix/process.cr
+++ b/src/crystal/system/unix/process.cr
@@ -14,7 +14,23 @@ struct Crystal::System::Process
   end
 
   def wait
-    @channel.receive
+    status = @channel.receive
+    case status & 0xff
+    when 0
+      ::Process::Status::Exited.new((status & 0xff00) >> 8)
+    when 0x01..0x7e
+      ::Process::Status::Signaled.new(status & 0x7f, core_dumped: false)
+    when 0x7f
+      ::Process::Status::Stopped.new((status & 0xff00) >> 8)
+    when 0x81..0xfe
+      ::Process::Status::Signaled.new(status & 0x7f, core_dumped: true)
+    else
+      if status == 0xffff
+        ::Process::Status::Continued.new
+      else
+        raise ::Process::Status::UnexpectedStatusError.new(status)
+      end
+    end
   end
 
   def exists?
--- a/src/crystal/system/win32/process.cr
+++ b/src/crystal/system/win32/process.cr
@@ -30,9 +30,9 @@ struct Crystal::System::Process
       raise RuntimeError.from_winerror("GetExitCodeProcess")
     end
     if exit_code == LibC::STILL_ACTIVE
-      raise "BUG: process still active"
+      raise ::Process::Status::UnexpectedStatusError.new(exit_code.to_i32!, "The process might still be active")
     end
-    exit_code
+    ::Process::Status::Exited.new(exit_code.to_i32!)
   end
 
   def exists?
--- a/src/process.cr
+++ b/src/process.cr
@@ -9,6 +9,7 @@ class Process
   record Tms, utime : Float64, stime : Float64, cutime : Float64, cstime : Float64
 end
 
+require "process/status"
 require "crystal/system/process"
 
 class Process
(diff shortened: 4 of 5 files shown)


oprypin:
The POSIXisms are very well established as is, though.
This is quite the opposite, *isolating* them.

If you pretend that those other states don't exist, that's fine too.
But you can't also pretend that an external shard can fix that, because essential information will be thrown away irrecoverably before it can get to it. Unless you suggest re-implementing Process entirely for such a shard.

straight-shoota:
I'm sure we can expose the platform-specific exit value and use that for such specific implementations.

asterite:
Could we have a single `Process::Status` where info that's not available for one possible status is just simply nilable?

oprypin:
I'm not particularly excited about a struct with 6 members to store 1 number. Which would also do nothing to prevent invalid states.

This *is* also the most straightforward way to implement this, even if you :nodoc: the underlying types. You can try to write out the code, there will be more of it than here.

Although my vision is that the types would be used more directly, with benefits of exhaustive `case` etc. Then the methods `normal_exit?`, `signal_exit?`, `exit_signal` can be deprecated.
People who care about these details will look at the subclasses, and people who don't will have `exit_code`+`success?` cover 99% of cases, and that will be *the entirety* of the top-level API.

asterite:
> with benefits of exhaustive case

Those benefits don't apply here because this is a type hierarchy and anyone could extend `Process::Status` and add their own type. We would need `@[Sealed]` for that. Or just change `Process::Status` to be a union of a fixed set of types.

oprypin:
> Those benefits don't apply here because this is a type hierarchy and anyone could extend `Process::Status` and add their own type.

Should this really be an argument? There are easier ways than that to break stdlib if one does such obviously unintended things.

---

Also maybe it's nice for when Linux adds more possible meanings to the range of exit statuses. :joy: 

> `WIFCONTINUED` is available since Linux 2.6.10

asterite:
> Should this really be an argument?

I'm just saying that we can't prove exhaustiveness if the type is open for everyone to open. We can't just crash the program when we reach the `else`.

What are the problems with Ruby: https://ruby-doc.org/core-2.5.1/Process/Status.html ?

oprypin:
> What are the problems with Ruby: https://ruby-doc.org/core-2.5.1/Process/Status.html ?

Well, the API is cluttered and every method's return value is nilable.
Specifically what that means for Crystal is:
`if s.signal_exit?; p s.exit_signal.not_nil!`

It's not the worst solution, though, and we're just minor tweaks (though quite a few added methods) away from that. It just means that the logic can't move into `Crystal::System::Process` (which I was against anyway).

Just that people have been suggesting refactors where we don't store the exit_status.

And I found a very nice way to achieve that, while also potentially decreasing the number of methods that programmers are exposed to by default.

asterite:
> if s.signal_exit?; p s.exit_signal.not_nil!

Can we make each method return a nilable something that has all the properties if it's there, but `nil` if it doesn't apply?

Otherwise, it might be fine to return a union type... as long as they all quack similarly for the common cases (`success?`). I'm just not sure if `$?` handles unions well, or even abstract structs, maybe yes.

oprypin:
Mm I guess it's about time for me to share the intended usage.

Basic:
```crystal
if !status.success?
  puts "Oops, code #{status.exit_code}"  # can be a negative number for signals, which is not great
end
```

Advanced:
```crystal
case status
when Process::Status::Exited
  puts "Exited with code #{status.exit_code}"
when Process::Status::Signaled
  puts "Terminated with signal #{status.signal}"
  puts "Core dumped" if #{status.core_dumped?}"
else # mandatory!
  puts "Something wrong, don't care!"
end
```
Alternate (per previous comment's suggestion, and already works, though ideally I'd remove it):
```crystal
if (signaled = status.signal_exit?)
  puts "Terminated with signal #{signaled.signal}"
else
  puts "Uh probably exited normally? No idea"
end
```
Regarding the last part, this is what can happen frequently in the current state - people want to check all outcomes but don't realize that normal exit and signal exit are not the only possible states.
In the compiler, I'm guessing this code is broken for STOPPED and CONTINUED, it tries to obtain some signal value where there is none:
https://github.com/crystal-lang/crystal/blob/f85bdb2095616848e277a2bbff20ffdb9337360a/src/compiler/crystal/compiler.cr#L579

oprypin:
Mm I just realized that `Process` never causes any state where `STOPPED` or `CONTINUED` would be possible. (does not set `WUNTRACED`)

Yeah now this is overkill.
My bad.

PR #9066, opened 2020-04-14
Reference global Spec in `be_a` macro

asterite:
Because it was doing `Spec::BeAExpectation`, if you use `be_a` inside a module where theree's a `Spec` type, it will find that instead of the global one.


asterite:
@RX14 Thank you!

Issue #7904, opened 2019-06-19
[RFC] Optional documentation generation for crystal version, stdlib and shards.


didactic-drunk:
Can the following options be added to the documentation generator to further my proposal in #5215 and for general use?

`crystal docs`
* --shards Also generate documentation for all shards.
* --stdlib Also generate documentation for required stdlib files.
* --stdlib-all Also generate documentation for required for all stdlib files.
* --core Also generate documentation for `Int` | `String` | Etc.
* --all All of the above.

Documentation for each class or method would ideally be marked/colored with the shard it came from, stdlib, core or with the most prominent indicator for the current application or shard.

RX14:
Having a `crystal docs --all` which provides a single combined documentation for developers working on a project is something I've thought about and wanted before. Adding indicators of where each type came from is an improvement which can come later.

didactic-drunk:
I've tested generating `--all` documentation locally.  Just like the crystal documentation it's useful but the list is too long when looking for information.  Especially so when trying to reason about what an application does or where the boundaries are between shards, the app, and crystal.

A filter with the following may help:

| All | App | Crystal | Shards |
| --- | --- | --- | --- |
| Shard1 | Shard2 | Shard3 | Etc |

Why each shard?  Sometimes your hunting for basic documentation.  Other times It's nice to know a shard adds methods to String or another base class.

jhass:
Rather than having to regenerate documentation I would like  to see project grouping support in documentation browser. So adding another hierarchy level on top when there's more than one project.  Similar to what devdocs.io does for languages, so you can search across everything in the top level but also move into the project and limit the search there.

straight-shoota:
As nice as this all sounds, I seriously doubt whether that fits into the scope of the docs generator. The current concept is quite clear: Provide docs for all features in `src` folder for the given project (shard) and the current version. That's a very clearly defined mode of operation, and yet has many aspects to cover (and lots of improvements waiting). Adding another layer on top of that, adds a lot of complexity that doesn't cover the main purpose.
As an alternative approach I'd suggest to try to improve connecting between individual API doc instances for separate shards. The main feature necessary for this would be to have way to map namespaces not included in the current API docs build to external locations where these namespaces are documented. That way individual code packages (=shards) would still have their encapsulated API docs but you can easily navigate to related API docs.
[...]

PR #9070, opened 2020-04-14
Refactor is_crystal_repo based on project name

straight-shoota:
This changes how the docs generator knows it's working on stdlib. Previously it looked for a git remote pointing to `github.com/crystal-lang/crystal`. Now with #8792 it can simply tell by the project name. That allows generating stdlib docs outside a git environment (for example from a tarball) which was impossible before.

For refernce: #7904 might add a separate option for enabling ouput of stdlib features which could make this change obsolete. But that's not fleshed out yet, and this is a simple improvement we can do right now.


RX14:
Needs CI fixed.

PR #9071, opened 2020-04-14
Support for pwrite

carlhoerberg:


asterite:
If anything, this should be an instance method of `IO::FileDescriptor`. Maybe call it `write_at` instead of the cryptic name `pwrite`. Also add a spec. Then, I don't know what happens with the buffer in `IO::Buffered` for file descriptors... so I think this is a bit more complex.

carlhoerberg:
thank you for the feedback!

wontruefree:
Instead of `pwrite` can we come up with a more descriptive api for file descriptor.

src/crystal/system/unix/file_descriptor.cr
```diff
@@ -181,4 +181,14 @@ module Crystal::System::FileDescriptor
...
+
+  def self.write_at(fd, buffer, offset)
+    bytes_written = LibC.pwrite(fd, buffer, buffer.size, offset)
+
+    if bytes_written == -1
+      raise IO::Error.from_errno("Could not write to file")
+    end
+
+    bytes_written
+  end
```
lbguilherme, line 193: This should be an instance method and should probably be named `unbuffered_write_at`, more in line with `unbuffered_write`. Also, it should use `evented_write` to handle the fact that this write might be async.

For reference:

```cr
  private def unbuffered_write(slice : Bytes)
    evented_write(slice, "Error writing file") do |slice|
      LibC.write(fd, slice, slice.size).tap do |return_code|
        if return_code == -1 &amp;&amp; Errno.value == Errno::EBADF
          raise IO::Error.new "File not open for writing"
        end
      end
    end
  end
```

carlhoerberg:
Fixed the comments

src/io/file_descriptor.cr
```diff
@@ -149,6 +149,14 @@ class IO::FileDescriptor < IO
     system_fsync(flush_metadata)
   end
 
+  # Write to a specific position in a file
+  # It will flush the current write buffer and not buffer this write.
+  def write_at(bytes : Bytes, offset)
+    check_open
+    flush
+    unbuffered_write_at(bytes, offset)
```
straight-shoota, line 157: Since there's no buffering involved, I'd suggest not to follow the naming scheme for buffered operations. Maybe we could call the method `system_write_at`?

straight-shoota (approved):

src/crystal/system/unix/file_descriptor.cr
```diff
@@ -27,6 +27,17 @@ module Crystal::System::FileDescriptor
     end
   end
 
+  private def system_write_at(bytes : Bytes, offset)
+    evented_write(bytes, "Error writing file") do |slice|
+      at = offset - (bytes.bytesize - slice.bytesize)
```
carlhoerberg, line 32:
```suggestion
      at = offset + (bytes.bytesize - slice.bytesize)
```

PR #9073, opened 2020-04-14
Replace File.new with File.open

RX14:
This pull request replaces `File.new(filename, "w")` with `File.open(filename, :write)` etc.

Follows on from and closes #8011.


RX14:
Further to #8011 I've added `File::Mode::Overwrite` - an alias for `:write, :create, :truncate` (the same as the old `"w"`). This shortens the refactors a fair bit.

RX14:
`O_SYNC` doesn't exist in windows. How should we handle `File::Mode::Sync` existing then?

wontruefree:
If the File API received an enum would that help provide some compile time guarantees? Something like `File.open(filename, File::Write)`

RX14:
It does receive an enum. I'm just writing the shorthand.

asterite:
The title is a bit confusing. I think this was just renaming `new` to `open`. Could we split this PR into two separate PRs? 🙏

RX14:
It's replacing the `File.new(..., "string")` API with a different `File.open` API.

I agree it's a little confusing, I probably should have written more detail in the PR OP.

spec/std/file_spec.cr
```diff
@@ -711,7 +742,7 @@ describe "File" do
       with_tempfile("append-existing.txt") do |path|
         File.write(path, "hello")
         File.read(path).should eq("hello")
-        File.write(path, " world", mode: "a")
+        File.write(path, " world", mode: File::Mode.flags(Create, Append))
```
bcardiff, line 745: Can we have a simpler way to use `File.write`? Maybe this imply moving arguments from it to named arguments to match the `File.open` ones.

spec/std/file_spec.cr
```diff
@@ -754,7 +785,7 @@ describe "File" do
   it "opens with perm (int)" do
     with_tempfile("write_with_perm-int.txt") do |path|
       perm = 0o600
-      File.open(path, "w", perm) do |file|
+      File.open(path, :write, :create, permissions: perm) do |file|
```
bcardiff, line 788: I think we are forcing named arguments in order to overcome a limitation of autocasting of enums flags.

If the mode enum covers the more common combination like `:overwrite` then I would prefer to remove the named argument in favor of the type-safe arguments only. 

From the `File::Mode.flags(Create, Append)` and this maybe we need another enum value that can be used plainly as `:write_create` or `:w`

src/crystal/system/file.cr
```diff
@@ -1,8 +1,51 @@
...
+      else
+        raise "Unknown mode #{m}"
+      end
+    end
+    flags
+  end
+
+  # Helper method for calculating file open modes on systems with posix-y `open`
+  # calls.
+  private def self.open_flag(mode : String)
```
bcardiff, line 48: Your idea is to remove this eventually together with the deprecated methods, right?

src/file/mode.cr
```diff
@@ -0,0 +1,16 @@
+class ::File < IO::FileDescriptor
+  @[Flags]
+  enum Mode
+    Read
+    Write
+    Overwrite
```
bcardiff, line 6: It's odd that this is not defined as `Overwrite = Write | Create | Truncate` and others.
I get that the meaning is given in `Crystal::System` but it due to the Flags, as long as the main values are difined everything should be covered.

src/file_utils.cr
```diff
@@ -41,8 +41,8 @@ module FileUtils
   def cmp(filename1 : String, filename2 : String)
     return false unless File.size(filename1) == File.size(filename2)
 
-    File.open(filename1, "rb") do |file1|
-      File.open(filename2, "rb") do |file2|
+    File.open(filename1) do |file1|
```
bcardiff, line 44: What happened to the binary mode?

spec/std/file_spec.cr
RX14: Using `File.write` with a custom mode is very rare. I think optimizing this would be a little premature.

src/crystal/system/file.cr
RX14: Yep.

src/file/mode.cr
RX14: Right, I could totally do that!

I just wasn't thinking well when I wrote this commit last night :P

src/file_utils.cr
RX14: The binary mode is a no-op, see https://github.com/crystal-lang/crystal/blob/master/src/crystal/system/file.cr#L33

spec/std/file_spec.cr
RX14: I think the `*args` interface is the neatest. Passing custom flags like `:write, :create` seems more common than passing custom create permissions.

That being said, there might be a better interface I can't think of right now.

src/file_utils.cr
bcardiff: Yes, but even if it's no-op today, wouldn't that be needed eventually or for other platforms? Almost every language still allows you to specify binary mode. Probably I'm missing some history lesson here.

spec/std/file_spec.cr
bcardiff: I think I prefer to

* cover the `write|create`  and `creat|append` with additional enum values
* say with a single positional mode argument
* force longer construct as in File.write spec if those are not covered in the enum

This will leave similar api on File.write and File.open since due to the default arguments in the former there can't be a varyadic mode.

src/file_utils.cr
RX14: Binary mode applies only to windows, and it's not implemented there either. I think adding binary/text mode support is a new feature and out of scope for this PR, so I'd rather discuss this elsewhere.

spec/std/file_spec.cr
RX14: I still think this is optimizing for the wrong thing. I only added `Overwrite` since it was a *lot* shorter and it was easy to name. It's obviously not feasible to add one for every combination, and I'd like to keep the API nice for those.

Empirical data would be the use rate of `permissions`. Since we havent supported these more specific combinations before, there's no empirical data on their use rate unfortunately.

RX14:
I've tried defining `Overwrite = Write | Create | Truncate` but the definition of `Enum#each` is kinda broken and I really don't want to fix that in this PR.

bcardiff:
I would expect that the `mode.each` in `system/file.cr` to be gone if you are explicitly iterating the main values only. But ok.

I think we need more opinions before merging this since the api changes are not minor.

RX14:
Yeah I was planning to let it sit for a while longer.

RX14:
@crystal-lang/crystallers any more opinions?

straight-shoota:
Oh there's one aspect I'm not super excited about: Previously, there was a destinction between `File.new` as the "unsafe" method which needs manual closing and `File.open` which automatically closes at the end of the block. That's nice because you can always recommend to use `File.open` because it's safer and the non-yielding `File.new` is only useful in special cases where you need to store the file handle somewhere.
So with this change you'd need to be more explicit about `File.open with a block` for the recommended version.

I realize that `File.open` is technically a more fitting name than `File.new` because it just opens a file handle, doesn't necessarily create a *new* file. But then this only matters for the non-yielding method which is more of an advanced API. 

This also breaks with the established convention of having `.new` and `.open(&)` methods for accessing a resource without or with automatic release. AFAIK all resource-related stdlib APIs work this way. With this change, `File` would be the only outlier.

We could consider changing this convention. Replacing `.new` with `.open!` would probably be even better to express the unsafe nature. But this should be discussed for all APIs.

So I'd prefer the apply the implementation changes but don't rename `File.new` to `File.open`.

waj:
I love the idea of this refactor. The current api, specially the flags defined with a string only existed because of the Ruby an C apis.

However, I wonder if we're losing the opportunity to design even a better interface. `File.open` documentation read: `*modes* should contain at least one of`. It's almost crying out for a separate enum. Some of the other flags are just, well, flags that could be simply named arguments or a separate enum. If I write something like `:write, :create, :sync, symlink_no_follow` it feels weird that I had to put all those flags in the same set. Also, there are some invalid combinations that we could avoid just by making wise use of the type system. The only reason all those flags exist within the same argument is probably POSIX history. They keep adding features to the same API by adding more values.

It's quite late already and I'm a little burned out, so don't take this as an exact proposal, just an idea. We could have a `FileMode` (not flags) enum with values `Read | Write | ReadWrite | Append` values. And another flags enum with the `FileOption` values. Also, I'm thinking if should use the `new` vs `open` methods to indicate the "create" flag. This could be understand as I "open" existing files and I create "new" files. Or if using the `new` for this proposal is controversial, we could have a separate `create` method as well.

RX14:
How would you envision the "Overwrite" API if we had two different enums. I agree it's conceptually nicer, but i'm not sure the practicalities are quite as good.

j8r:
@RX14 This can work:
```cr
class File
  enum Mode
    Append
    Read
    ReadWrite
    Overwrite
  end

  @[Flags]
  enum Options
    Create
    CreateNew
    Truncate
    Sync
    SymlinkNoFollow 
  end
end
```

j8r:
@RX14 This can work:
```cr
class File
  enum Mode
    Append
    Read
    ReadWrite
    Overwrite
    Write
  end

  @[Flags]
  enum Options
    Create
    CreateNew
    Truncate
    Sync
    SymlinkNoFollow 
  end
end
```

oprypin:
Hasn't this been discussed over and over on the issue? Even the possibility of separate enums.

j8r:
@oprypin you probably initially mean also https://github.com/crystal-lang/crystal/issues/7857 - apparently, not enough.

didactic-drunk:
`.new` vs `.open` is completely unrelated to the flags used with `File`.  Can we please not tie the 2 PR's together.  I've been waiting 1.5+ years for open flags.

didactic-drunk:
Bump 2.5 years.

PR #9074, opened 2020-04-14
Add version selector to API docs

straight-shoota:
This adds a version select box to the sidebar on the API docs. The idea is described in https://github.com/crystal-lang/crystal/pull/7011#issuecomment-583656393

This is an opt-in feature of the docs generator. It doesn't work out of the box. You need to publish a JavaScript config file which describes the available releases that should be shown in the version select.
The content of such a file looks like this (example for stdlib, reduced to versions >= 0.30.0):
```js
var projectVersions = [
  {name: "0.35.0-dev", url: "/api/master/", released: false},
  {name: "0.34.0", url: "/api/0.34.0/", latest: true},
  {name: "0.33.0", url: "/api/0.33.0/"},
  {name: "0.32.1", url: "/api/0.32.1/"},
  {name: "0.32.0", url: "/api/0.32.0/"},
  {name: "0.31.1", url: "/api/0.31.1/"},
  {name: "0.31.0", url: "/api/0.31.0/"},
  {name: "0.30.1", url: "/api/0.30.1/"},
  {name: "0.30.0", url: "/api/0.30.0/"}
];
```

In order to properly resolve #7011 we should probably add some additional notification when browsing an outdated or unreleased version. This is signalled by `latest: true` and `released: false` which currently both don't have any effect. But I'd like to land this PR first and then continue with the next steps.

### DIFF
--- a/scripts/docs-versions.sh
+++ b/scripts/docs-versions.sh
@@ -1,0 +1,23 @@
+#! /usr/bin/env sh
+
+git tag --list | \
+grep -v -E '0\.1?[0-9]\.' | \
+grep '^[0-9]' | \
+tac | \
+awk '
+  BEGIN {
+    print "{"
+    print "  \"versions\": ["
+    getline current_version < "src/VERSION"
+    printf "    {\"name\": \"" current_version "\", \"url\": \"/api/master/\", \"released\": false}"
+  }
+
+  {
+    printf ",\n    {\"name\": \"" $1 "\", \"url\": \"/api/" $1 "/\"}"
+  }
+
+  END {
+    print "\n  ]"
+    print "}"
+  }
+'
--- a/man/crystal.1
+++ b/man/crystal.1
@@ -180,6 +180,8 @@ In case no default can be found, this option is mandatory.
 .It Fl -project-version Ar VERSION
 Set the project version. The default value is extracted from current git commit or shard.yml if available.
 In case no default can be found, this option is mandatory.
+.It Fl -json_config_url Ar URL
+Set the URL pointing to a config file (used for discovering versions).
 .It Fl o Ar DIR, Fl -output Ar DIR
 Set the output directory (default: ./docs).
 .It Fl b Ar URL, Fl -sitemap-base-url Ar URL
--- a/src/compiler/crystal/command/docs.cr
+++ b/src/compiler/crystal/command/docs.cr
@@ -44,6 +44,10 @@ class Crystal::Command
         output_format = value
       end
 
+      opts.on("--json_config_url=URL", "Set the URL pointing to a config file (used for discovering versions)") do |value|
+        project_info.json_config_url = value
+      end
+
       opts.on("--canonical-base-url=URL", "Deprecated option. Use --sitemap-base-url instead.") do |value|
         abort "Option --canonical-base-url is no longer supported.  Use --sitemap-base-url instead."
       end
--- a/src/compiler/crystal/tools/doc/html/_head.html
+++ b/src/compiler/crystal/tools/doc/html/_head.html
@@ -1,6 +1,11 @@
 <meta charset="utf-8" />
 <meta http-equiv="X-UA-Compatible" content="IE=edge">
 <meta name="generator" content="Crystal Docs <%= Crystal::VERSION %>">
+<meta name="crystal_docs.project_version" content="<%= project_info.version %>">
+<meta name="crystal_docs.project_name" content="<%= project_info.name %>">
+<% if json_config_url = project_info.json_config_url %>
+  <meta name="crystal_docs.json_config_url" content="<%= json_config_url %>">
+<% end %>
 
 <link href="<%= base_path %>css/style.css" rel="stylesheet" type="text/css" />
 <script type="text/javascript" src="<%= base_path %>js/doc.js"></script>
--- a/src/compiler/crystal/tools/doc/html/css/style.css
+++ b/src/compiler/crystal/tools/doc/html/css/style.css
@@ -56,6 +56,7 @@ body {
   padding: 0 0 30px;
   box-shadow: inset -3px 0 4px rgba(0,0,0,.35);
   line-height: 1.2;
+  z-index: 0;
 }
 
 .sidebar .search-box {
@@ -107,9 +108,7 @@ body {
 }
 
 .project-summary {
-  display: inline-block;
   padding: 9px 15px 30px 30px;
-  text-align: right;
 }
 
 .project-name {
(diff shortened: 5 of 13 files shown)


RX14:
Could it be possible to make the config file JSON instead?

bcardiff:
Making the configuration a json file will require to use ajax to load it https://stackoverflow.com/questions/13515141/html-javascript-how-to-access-json-data-loaded-in-a-script-tag-with-src-set

That will not work on local filesystem.

But at least I would suggest to use a scoped property instead of a global variable.

```js
CrystalDocs.projectVersions = [
   // ...
]
```

straight-shoota:
This is more difficult to implement, specially for local use over `file://` protocol. I'm not sure if there's a huge use case for that, though. So we might just be able to ignore that. It's still easier to just plop in a `<script>` tag.

But a plain `.js` file is also more versatile because you can put arbitrary code in it. This could be useful for example to roll out enhancements for older generated docs, or insert some kind of temporary announcements... Lot's of possibilities that might be worth considering in the future.

I don't see an immediate benefit to using JSON for this.

straight-shoota:
@bcardiff I avoided to scope the variable into an existing namespace because it is affected by order of execution.
But we could consider using a dedicated namespace:
```js
var CrystalDocsConfig = {
  projectVersions: [
   // ...
  ]
}
```

jhass:
Well one benefit I see is an easy job for third party consumers. Thinking of scrapers for things like Dash, devdocs.io etc, in theory they could automatically discover all versions with something like this.

But then actually generating this file is out of scope of this anyways, so it would only be relevant here towards pushing the possible generator(s) to generate well parsable data.

straight-shoota:
I suppose we can go with a JSON file. There's a clear use case for better reusability with other tools. I don't think we need to care about `file://` protocol. It's fully supported for browsing docs of a single version, but I don't think the version switch feature is really necessary. When you already have the different version folders available locally, it's easy to jump to a different version on the file system. Or just run a simple static file server if you want the select tool.

Going for reusability would also ask to make the file easily discoverable, preferredly by a standardized name. I suppose the way we're going to use this for stdlib andcrystal-lang shards will most likely be adapted by other shards too.

So it's just a question should we call it `versions.json` or a more generic name like `config.json` to be less specific and open for additional use cases? I'd probably go with `versions.json` because it's easier to understand what it does (that's also what I have been working with so far). We should still put the versions list in a property, so we can add additional properties later, if necessary:

`versions.json`:
```json
{
  "versions": [
    {"name": "0.35.0-dev", "url": "/api/master/", "released": false},
    {"name": "0.34.0", "url": "/api/0.34.0/", "latest": true},
    {"name": "0.33.0", "url": "/api/0.33.0/"},
    {"name": "0.32.1", "url": "/api/0.32.1/"},
    {"name": "0.32.0", "url": "/api/0.32.0/"},
    {"name": "0.31.1", "url": "/api/0.31.1/"},
    {"name": "0.31.0", "url": "/api/0.31.0/"},
    {"name": "0.30.1", "url": "/api/0.30.1/"},
    {"name": "0.30.0", "url": "/api/0.30.0/"}
  ]
}
```

bcardiff:
Yeah, maybe we can not care about `file:///`. Using `versions.json` seems fine. The only thing that I would change is for the `latest` to be a separate field. Easier to get and avoid allowing multiple "lastests".

```js
{
  "latest": "0.34.0", // or {"name": "0.34.0", "url": "/api/0.34.0/"},
  "versions": [
    {"name": "0.35.0-dev", "url": "/api/master/", "released": false},
    {"name": "0.34.0", "url": "/api/0.34.0/"},
    {"name": "0.33.0", "url": "/api/0.33.0/"},
    {"name": "0.32.1", "url": "/api/0.32.1/"},
    {"name": "0.32.0", "url": "/api/0.32.0/"},
    {"name": "0.31.1", "url": "/api/0.31.1/"},
    {"name": "0.31.0", "url": "/api/0.31.0/"},
    {"name": "0.30.1", "url": "/api/0.30.1/"},
    {"name": "0.30.0", "url": "/api/0.30.0/"}
  ]
}
```

straight-shoota:
Well this PR doesn't add any meaning to `latest` so we should discuss these details when it's time for that.

straight-shoota:
Btw. it seems hex docs chose `.js` because of `file://` use.

Makefile
```diff
@@ -91,7 +91,7 @@ compiler_spec: $(O)/compiler_spec ## Run compiler specs
 
 .PHONY: docs
 docs: ## Generate standard library documentation
-	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION)
+	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION) --json_config_url=/api/versions.json
```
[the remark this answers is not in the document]
straight-shoota: Yeah, this should be configured in distribution-scripts.

scripts/docs-versions.sh
```diff
@@ -1,10 +1,20 @@
...
+    { print "    {\"name\": \"" $1 "\", \"url\": \"/api/" $1 "/\"}," }
+    END {
+        getline < "src/VERSION"
+        print "    {\"name\": \"" $0 "\", \"url\": \"/api/master/\", \"released\": false},"
+        print "  \"versions\": ["
+        print "{"
+    }
+    ' | \
+sed -e "3 s/.\$//" | \
+tac
```
RX14, line 20: This is, quite literally, backwards

```awk
BEGIN {
  print "{"
  print "  \"versions\": ["
  getline current_version < "src/VERSION"
  printf "    {\"name\": \"" current_version "\", \"url\": \"/api/master/\", \"released\": false}"
}

{
  printf ",\n    {\"name\": \"" $1 "\", \"url\": \"/api/" $1 "/\"}"
}

END {
  print "\n  ]"
  print "}"
}
```

this is cleaner

straight-shoota: Both are just not ideal solutions. Buy yeah, between backwards code and leading commans the latter is probably better.

straight-shoota: Oh I missed the line break, so this is really pretty good 👍

straight-shoota:
Rebased and squashed.

bcardiff:
It looks a bit weird that the current version appears twice in the dropdown

Screen Shot 2020-04-27 at 10 44 52

And adding the dropdown generates a couple of pixel displacements.

straight-shoota:
The current version should only been added when it's not in the list. But at some time during refactoring I renamed the property from `version` to `name` and it seems the check doesn't reflect that. It's an easy fix.

Can you be more specific about the pixel displacements?

bcardiff:
@straight-shoota I mean the following displacement that occurs after the versions.js file is loaded.

VkZnhA9uNN

straight-shoota:
Ah yeah, I got it.

Blacksmoke16:
Could we get something added to the book regarding this feature?  Like that it exists, what the format structure of the `json` file should be, and how it works?

straight-shoota:
The version selector was primarily pushed forward for the stdlib API. Obviously, it's also useful for shards. But at least I haven't checked how that works out and is great to use yet.

PR #9076, opened 2020-04-15
Move CallStack to Exception::CallStack

bcardiff:
And remove explicit filter of function prefix names since they are handled in `Exception::CallStack.skip(__FILE__)`.

Since CallStack was nodoc, there is no deprecation needed.


asterite:
But `caller` returns the callstack and it's not related to any exception.

jhass:
Alternatives to keeping it on the top level I see are `Crystal` and `Debug`. None of these seems like a perfect fit though. I don't think it really matters where it lives as it's a nodoc type mostly transparent to the user. `Exception` makes sense since that's its main consumer in stdlib. I've seen languages categorizing their `caller` into their exception framework and providing it there as something like `Exception.current_callstack` or so. So I think this is fine really.

bcardiff:
`caller` returns an `Array(String)`, so is more an implementation dependency than a return value.
I added some documentation to it and clarify the return type.

To fully hide the CallStack as it was intendend the `Exception#callstack` property needs to also be `:nodoc:`.

asterite:
> caller returns an Array(String), so is more an implementation dependency than a return value.

Oh, yeah, makes sense 👍

src/exception.cr
```diff
@@ -17,6 +17,8 @@ class Exception
   # This is useful for wrapping exceptions and retaining the original
   # exception information.
   getter cause : Exception?
+
+  # :nodoc:
   property callstack : CallStack?
```
straight-shoota, line 22: Why make this nodoc? This seems like an important component of the API.
It should be the other way around: Document `Exception::CallStack`.

bcardiff: I think is rather a runtime internal. The main justification to move CallStack inside Exception for me is that it was `:nodoc:` at a first place.

asterite: I agree. This was an undocumented type. The idea seems to be to hide it more, not to suddenly expose it. That can always happen later, though.

straight-shoota: Oh, never mind. I confused it with `backtrace`. Similar story as above...

Sija:
Heads up: this PR broke BC with Crystal `0.34.0`.

straight-shoota:
In what way did id break something? As far as I can see it's merely an internal refactor.

Sija:
If someone relied on `CallStack` being defined in the top-level namespace then this PR broke it - I know it's `:nodoc:`-ed, just saying.

PR #9079, opened 2020-04-15
Parse `\ ` (backslash + space) inside regex literal to ` ` (space)

### LINKED ISSUE #9016: `RegexLiteral#to_s` returns invalid regex literal when literal value starts with whitespace character.
For example this code fails on compiling with `Error: Unterminated regular expression`.

```crystal
macro foo(x)
  p {{x}}
end

foo %r( )
```

Because `foo %r( )` expands to `p / /` and this first slash is recognized as divide operator and second is recognized as begin delimiter of regex.

MakeNowJust:
Fixed #9016

After #9017, `RegexLiteral#to_s` appends backslash to its result if its value starts with whitespace. However this changes the regex value. This PR fixes it.

You may concern this behavior is too implicit, and it seem day-to-day fix. But `\/` inside regex literal treats like this. For example:

```crystal
/\w/.source # => "\\w"
/\//.source # => "/"
/\ /.source # => " " (after this PR is merged)
```

And, in fact, this fix is needed because space is used to determine which `/` is regex start delimiter or divide operator.


src/compiler/crystal/syntax/lexer.cr
```diff
@@ -1990,8 +1990,8 @@ module Crystal
             char = next_char
             next_char
             @token.type = :STRING
-            if string_end == '/' && char == '/'
-              @token.value = "/"
+            if string_end == '/' && (char == '/' || char.ascii_whitespace?)
```
oprypin, line 1993: Why should it depend on string_end? Will `%r(\ )` differ?? Needs spec for that too anyway

MakeNowJust: Because this this fix is needed to other delimiters. And, I respects existing process for `\/`.

asterite: I think we should make it the same regardless of delimiter. Otherwise it might be a bit confusing.

MakeNowJust: @asterite You added this at 19fdf45... Did you change your mind?

asterite: I don't know what happened 🤷

MakeNowJust: Ah, it is problematic. For example, this output should be `true`, but `true` is shown:

```cryystal
macro identity(x)
  {{x}}
end

p %r(\/) == identity(%r(\/))
```

I'll fix. Ok?

MakeNowJust:
It is ready to ship. Could any review it?

bcardiff:
Woudn't be good add a couple of specs in `compiler/parser/parser_spec.cr`? (around line 950 aprox)

MakeNowJust:
@bcardiff Good. I added such a spec.

MakeNowJust:
Good. I added such a spec.

PR #9083, opened 2020-04-15
Remove special handling for version tags in docs generator

straight-shoota:
This was trying to be too smart but it wouldn't work reliably without
proper version tag recognition. This added complexity is also unnecessary.
Just displaying the plain git tag (plus suffix if dirty) should be sufficient.

### DIFF
--- a/spec/compiler/crystal/tools/doc/project_info_spec.cr
+++ b/spec/compiler/crystal/tools/doc/project_info_spec.cr
@@ -22,10 +22,10 @@ describe Crystal::Doc::ProjectInfo do
     end
   end
 
-  describe ".new_with_defaults" do
+  describe "#fill_with_defaults" do
     it "empty folder" do
       assert_with_defaults(ProjectInfo.new(nil, nil), ProjectInfo.new(nil, nil))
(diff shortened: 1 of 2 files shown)


RX14:
Stripping the `v` was nice...

bcardiff:
What's the problem with removing the `v` from a git tag? I think that is nice also.

If we wan't more reliable removal of that, we can check if the remaining tag is a valid semantic version. But I would prefer to avoid adding that `v`.

straight-shoota:
Okay. So the important change here is that it doesn't blindly chop off a `v` at the beginning of a tag.
The doc generator shouldn't care about how tags are applied. If a commit is tagged, it should prefer that tag over anything else.

Displaying shards-style version tags without the `v` prefix is nice. Still, it's not strictly necessary. Essentially it's just a stylistic feature. Some might prefer it that way, other not. Using just the plain git tag would be perfectly fine. And if you don't agree with that, there is always the option to pass in `--project-version`.

But we can re-add removing the `v` prefix on version tags. It just needs a more elaborate implementation.

PR #9087, opened 2020-04-15
Correct duplicated articles/prepositions in comments

MakeNowJust:
Vanish 'the the' / 'with with' / 'to to'.

> War.. War never changes.
>
> From "Fallout 4"

### DIFF
--- a/src/compiler/crystal/command.cr
+++ b/src/compiler/crystal/command.cr
@@ -5,7 +5,7 @@
 # some in `tools`, some here, and some create a Compiler and
 # manipulate it.
 #
-# Other commands create a `Compiler` and use it to to build
+# Other commands create a `Compiler` and use it to build
 # an executable.
 
 require "json"
--- a/src/compiler/crystal/semantic/method_lookup.cr
+++ b/src/compiler/crystal/semantic/method_lookup.cr
@@ -79,7 +79,7 @@ module Crystal
             # If the argument types are compatible with the match's argument types,
             # we are done. We don't just compare types with ==, there is a special case:
             # a function type with return T can be transpass a restriction of a function
-            # with with the same arguments but which returns Void.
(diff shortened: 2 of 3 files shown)


MakeNowJust:
And bonus! Super-cool Ruby script for you.

```ruby
ARTICLES = %w[a an the my your her his our their]
PREPOSITIONS = %w[
  about aboard above across after against along alongside amid among anti around as at
  bar before behind below beneath beside besides between beyond but by
  considering
  despite down during
  except
  for from
  in inside into
  less like
  minus
  near notwithstanding
  of off on onto opposite out outside over
  pace past pending per plus
  re regarding round
  save saving since
  than through throughout till to touching toward
  under underneath unless unlike until up
  versus via vice
  with within without
]

RE_ARTICLES = Regexp.union(*ARTICLES)
RE_PREPOSITIONS = Regexp.union(*PREPOSITIONS)

RES = [
  /^(\s*#.*)\b(#{RE_ARTICLES})([\s#]+)(#{RE_ARTICLES})\b(.*)$/,
  /^(\s*#.*)\b(#{RE_PREPOSITIONS})([\s#]+)(\2)\b(.*)$/,
]

Dir["**/*.cr"].each do |path|
  content = File.read(path)
  RES.each do |re|
    content.scan(re) do
      m = $~
      lineno = content[0...m.begin(0)].scan("\n").size + 1
      highlighted = "> #{m[1]}\e[1;4m#{m[2]}\e[0m#{m[3].gsub("\n", "\n> ")}\e[1;4m#{m[4]}\e[0m#{m[5]}"
      puts "\e[1;34m#{path}\e[0m:\e[33m#{lineno}\e[0m: found '\e[4;1m#{m[2]} #{m[4]}\e[0m' here"
      puts
      puts highlighted
      puts
    end
  end
end
```

PR #9088, opened 2020-04-15
Use SOURCE_DATE_EPOCH only to determine compiler date

straight-shoota:
This removes a platform-specific detail from the compiler (see #9054). With this change, the compiler does not automatically use the current time as build date. It must be set by env var `SOURCE_DATE_EPOCH`. Otherwise it's just empty.

This env var is by default populated in the Makefile in the same way we're already handling `CRYSTAL_CONFIG_BUILD_COMMIT` there.

Resolves #9057

This approach was suggested by @asterite in https://github.com/crystal-lang/crystal/pull/9057#issuecomment-613660736 and I agree that this is probably the best solution (per https://github.com/crystal-lang/crystal/pull/9057#issuecomment-613712633).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,10 +31,12 @@ override FLAGS += $(if $(release),--release )$(if $(stats),--stats )$(if $(progr
 SPEC_WARNINGS_OFF := --exclude-warnings spec/std --exclude-warnings spec/compiler
 SPEC_FLAGS := $(if $(verbose),-v )$(if $(junit_output),--junit_output $(junit_output) )
 CRYSTAL_CONFIG_BUILD_COMMIT := $(shell git rev-parse --short HEAD 2> /dev/null)
+SOURCE_DATE_EPOCH := $(shell git show -s --format=%ct HEAD 2> /dev/null)
 EXPORTS := \
   $(if $(release),,CRYSTAL_CONFIG_PATH="$(PWD)/src") \
   CRYSTAL_CONFIG_LIBRARY_PATH="$(shell crystal env CRYSTAL_LIBRARY_PATH)" \
-  CRYSTAL_CONFIG_BUILD_COMMIT="$(CRYSTAL_CONFIG_BUILD_COMMIT)"
+  CRYSTAL_CONFIG_BUILD_COMMIT="$(CRYSTAL_CONFIG_BUILD_COMMIT)" \
+	SOURCE_DATE_EPOCH="$(SOURCE_DATE_EPOCH)"
 SHELL = sh
 LLVM_CONFIG := $(shell src/llvm/ext/find-llvm-config)
 LLVM_EXT_DIR = src/llvm/ext
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ override FLAGS += $(if $(release),--release )$(if $(stats),--stats )$(if $(progr
 SPEC_WARNINGS_OFF := --exclude-warnings spec/std --exclude-warnings spec/compiler
 SPEC_FLAGS := $(if $(verbose),-v )$(if $(junit_output),--junit_output $(junit_output) )
 CRYSTAL_CONFIG_BUILD_COMMIT := $(shell git rev-parse --short HEAD 2> /dev/null)
-SOURCE_DATE_EPOCH := $(shell git show -s --format=%ct HEAD 2> /dev/null)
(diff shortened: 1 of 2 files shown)


bcardiff:
This will require some updates in https://github.com/crystal-lang/distribution-scripts/blob/master/omnibus/config/software/crystal.rb#L52, the linux packages seems ok since they relay on `make crystal`.

The downside of using the commit date is that the nightly builds dates will not refer to the build date, but to the las update of the source.

oprypin:
Yes, I'm not sure *why* this PR also changes the behavior of what's being put in there. Why not just stick to `date`?

oprypin:
Good idea overall though

straight-shoota:
Oh we can certainly change it to display the current date for development/nightly builds. I just didn't want to push to deeply into this without gettin proper feedback if that's where we want to go.

straight-shoota:
In general I'd say that the git commit timestamp is more significant than the actual build time (assuming building the exact content of a git commit, i.e. git workdir and index are clean). This value closely relates to `CRYSTAL_CONFIG_BUILD_COMMIT` so it makes sense to pull from the same source.

Makefile
```diff
@@ -31,10 +31,12 @@ override FLAGS += $(if $(release),--release )$(if $(stats),--stats )$(if $(progr
 SPEC_WARNINGS_OFF := --exclude-warnings spec/std --exclude-warnings spec/compiler
 SPEC_FLAGS := $(if $(verbose),-v )$(if $(junit_output),--junit_output $(junit_output) )
 CRYSTAL_CONFIG_BUILD_COMMIT := $(shell git rev-parse --short HEAD 2> /dev/null)
+SOURCE_DATE_EPOCH := $(shell git show -s --format=%ct HEAD 2> /dev/null)
```
oprypin, line 34: No, I mean, why is this changing the behavior where the default previously was the current date but now it's something from a commit
```suggestion
SOURCE_DATE_EPOCH := $(shell date +%s)
```

straight-shoota: Because that's what SOURCE_DATE_EPOCH is supposed to be. It should represent the date of the source code, not the date of the build.

oprypin: OK but this is 2 separate changes in one pull request, just makes it harder to agree

straight-shoota: Why would you prefer it the other way around?

straight-shoota:
@bcardiff Maybe nightly builds should actually have the build date as version metatada, like `0.35.0-dev.nightly.2020-04-15`

Is it correct that the omnibus build doesn't use `CRYSTAL_CONFIG_BUILD_COMMIT`? Or am I just missing that?

Makefile
straight-shoota: Hm, maybe.

straight-shoota: But actually I don't think so. Without changing to source based date by default, even I would probably prefer #9057 over just externalizing `date +%s`.
With this change it is clear that determining the value should be handeld outside the compiler.

bcardiff:
> Maybe nightly builds should actually have the build date as version metatada, like 0.35.0-dev.nightly.2020-04-15 or just 2020.04.15-nightly.

👍 for something that includes the build date lile 0.35.0-dev-20200415 or whatever variation. I would not use just 2020.04.15-nightly because I think eventually we will have more than one development track.

Watch out how the docker images are labeled. They should stay as `:nighty` so it's we don't pollute the images every day.

> Is it correct that the omnibus build doesn't use CRYSTAL_CONFIG_BUILD_COMMIT? Or am I just missing that? And why does it not simply use the make recipe?

I think you are right. Make is not used becuase history I think. Maybe when https://github.com/crystal-lang/omnibus-crystal/commit/c1a2091c06c932579e973767301a2e04691723c6 was done, the makefile could be used since the `.build/crystal` is located manually.

Makefile
jhass: Can we try falling back to `stat -f "%c" Makefile` for tarball based builds?

straight-shoota: For GNU stat it's `stat -c "%Y" Makefile`, so it's not portable. Neither is `date -r Makefile +%s`.

jhass: MacOS seems to error out on `stat -c "%Y" Makefile`, so we can just chain the approaches?

```
(git show -s --format=%ct HEAD  || stat -c "%Y" Makefile ||  stat -f "%m" Makefile) 2> /dev/null

PR #9090, opened 2020-04-15
Refactor spec hooks

### LINKED ISSUE #8930: Spec: Improve top-level hook API
When spec hooks were added in #8302, they only work inside example groups.
```cr
require "spec"

before_all { } # Error: Can't call `before_all` outside of a describe/context

describe "foo"
  before_all { } # Ok
end
```

Since before those group-level hooks were added, we already had some top-level hooks:

```cr
Spec.before_each { }
Spec.before_suite { }
```
They have not changed with the arrival of the group-level hooks.

But I think it would be time for some adaptions in order to provide a clean and consistent API.

* Top-level hooks should not require to be called as methods on `Spec`.  `before_each` is already in the top-level namespace for the DSL. We should just make it work outside of example groups instead of raising an error. This means to remove `Spec.before_each` and insert the behaviour where `::before_each` currently raises.
* `before_suite` is a descriptive name, but it would be better called `before_all` for consistency with the group-level hook. It's confusing to have different names for the same concept on different levels. `before_all` is not a specific name to the group-level and would also match to the top-level, so that's what people would expect it to be. Regarding the namespace, the same reasoning applies as for `before_each`. This means to remove `Spec.before_suite` and insert the behaviour where `::before_all` currently raises.
* The same arguments apply to the respective `after_*` and `around*` variants.
* Internally, we could consider refactoring the implementation of the top-level hooks to integrate with the group-level implementation so we don't have two separate hook systems.

straight-shoota:
* <del>Refactor all spec hooks to work in the root context</del>
* <del>Deprecate hook methods in Spec namespace</del>
* <ins>Refactor and simplify implementation of spec hooks</ins>
* Improve documentation of spec hooks
* Add spec for spec hooks

Note: `around_all` does not work on the root context. Because the block is supposed to be called with a `ExampleGroup`. We could perhaps make it work with some effort, but I don't think that's necessary. It would not be have many use cases anyways. This wasn't possible before, so it won't be in this PR.

### DIFF
--- a/spec/std/spec/hooks_spec.cr
+++ b/spec/std/spec/hooks_spec.cr
@@ -7,14 +7,45 @@ describe Spec do
         require "prelude"
         require "spec"
 
-        before_all { puts "top:before_all" }
-        before_each { puts "top:before_each" }
-        after_all { puts "top:after_all" }
-        after_each { puts "top:after_each" }
-        around_each do |example|
-          puts "top:around_each:before"
+        begin
+          before_all {}
+        rescue exc
+          puts exc.message
+        end
+        begin
+          before_each {}
+        rescue exc
+          puts exc.message
+        end
+        begin
+          after_all {}
+        rescue exc
+          puts exc.message
+        end
+        begin
+          after_each {}
+        rescue exc
+          puts exc.message
+        end
+        begin
+          around_all {}
+        rescue exc
+          puts exc.message
+        end
+        begin
+          around_each {}
+        rescue exc
+          puts exc.message
+        end
+
+        Spec.before_suite { puts "Spec:before_suite" }
+        Spec.before_each { puts "Spec:before_each" }
+        Spec.after_suite { puts "Spec:after_all" }
+        Spec.after_each { puts "Spec:after_each" }
+        Spec.around_each do |example|
+          puts "Spec:around_each:before"
           example.run
-          puts "top:around_each:after"
+          puts "Spec:around_each:after"
         end
 
         describe "foo" do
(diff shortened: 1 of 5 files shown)


src/spec/dsl.cr
```diff
@@ -213,6 +215,7 @@ module Spec
   # ```
   #
   # will print, just before the spec suite starts, 1 and then 2.
+  @[Deprecated("Use `Spec::Methods.before_all  ` instead")]
```
Sija, line 218:
```suggestion
  @[Deprecated("Use `Spec::Methods.before_all` instead")]
```

spec/std/spec/hooks_spec.cr
```diff
@@ -0,0 +1,108 @@
+require "../../spec_helper"
+
+describe Spec do
+  describe "hooks" do
+    it "runs in correct order" do
```
asterite, line 5: 😍

spec/std/spec/hooks_spec.cr
```diff
@@ -0,0 +1,108 @@
...
+      run(<<-CR).to_string.lines[..-5].should eq <<-OUT.lines
+        require "prelude"
+        require "spec"
+
+        before_all { puts "top:before_all" }
+        before_each { puts "top:before_each" }
```
bcardiff, line 11: These blocks will be executed on all the specs of all the files, right?
If so, it is slightly missleading. I would expect them to run the specs of these file only.

With `Spec.before_each` it was more clear to me that the block will be executed beyond the scope of the file.

I'm not sure there is a clear way to solve this. Thoughts?

src/spec/dsl.cr
```diff
@@ -175,6 +175,7 @@ module Spec
   # ```
   #
   # will print, just before each spec, 1 and then 2.
+  @[Deprecated("Use `Spec::Methods.before_each` instead")]
```
bcardiff, line 178: The message should suggest to use the top-level before_each. Users of spec are not neccesarily familiar with it's design and the fact there is a `include Spec::Methods` somewhere in the root scope.

straight-shoota: I know this is not ideal, but there is no other way to address the replacement method. In ``Use `::before_each` instead``, the method name wouldn't be linked. This is an issue of `Spec::Methods`, not of the deprecation message.

spec/std/spec/hooks_spec.cr
straight-shoota: Hm, I never considered file-local hooks. Not sure how useful they would be, especially considering that hooks that span multiple contexts would often be defind in helper files.
But I wouldn't be opposed to scope `before_each` in the root context to the current file and keep `Spec.before_each` as global.

But that would require to even introduce a file scope first. Maybe we don't need a separate instance for that but could use the location property of individual specs. But that feels like a separate story.

bcardiff: Maybe a `__FILE__` argument can be used to filter in runtime if the block applies to the file or not.

And we can have a `before|after|around_suite` top level method for the things that in this PR are top

But from the syntax I would expect them to be per file.

The edge case would be if the `before_each` is used in a "support" file.

straight-shoota: Yeah, filtering by file shouldn't be an issue. It's just the question of *how* to design such a file context. Should it be a separate `Context` type? Or an implicit hidden `ExampleGroup`? Or do we just filter dynamically in `run_before_each`?

straight-shoota: `before_suite` could be an equivalent to `before_all` on the suite scope. But we would still need to differentiate `before_each` on suite and file scope.

straight-shoota:
I've removed the deprecation of `Spec.` methods. Instead top-level methods continue to raise on the top level context.

We can iterate on this behaviour later. It leaves the option to implement top-level methods as file-scoped hooks.

straight-shoota:
I don't think this is a breaking change anymore.

Also it doesn't resolve #8930 because it essentially only addresses the last point.

Issue #8994, opened 2020-04-02
New Log module documentation, IOBackend Context  issues


z64:
1. `@type` in `Crystal.datum` macros emits `Crystal`. This causes some wrongly generated documentation. Some examples:

"Creates a Crystal that wraps the given Raw."
https://crystal-lang.org/api/master/Log/Context.html#new(raw:Type)-class-method

"All possible Crystal types."
https://crystal-lang.org/api/master/Log/Context/Type.html

---

2. Logging with `IOBackend` crashes if the "top level" `Context::Type` is not `Hash` or `Array`, because of [this line](https://github.com/crystal-lang/crystal/blob/ab8b04449162052a4e6375ba3c9b86766a59c6b9/src/log/io_backend.cr#L38). 

This needs to be documented for `IOBackend`, or the default formatter for `IOBackend` needs to be resilient to other `Context` types.

```cr
require "log"

backend = Log::IOBackend.new
Log.builder.bind("*", :info, backend)
Log.context = Log::Context.new(1337)
Log.for("").info { "test" }
```
```
I, [2020-04-02T22:50:05.555750000Z #1290]    INFO -- crystal-run-crash.tmp:: testUnhandled exception: Expected Array or Hash for #size, not Int32 (Exception)
```

Perhaps `datum` should add `#size?`, and then the formatter will fall back to `inspect`?

(I didn't follow the design discussions of `datum`, etc. closely, nor do I have any stake/opinion in which solution is picked.)

bcardiff:
I would expect `Log.context=` to fail if the Context is not a hash like. Since that is also assumed in `extend_fiber_context`.

Regarding the documentation issues, I extracted that to #9099

z64:
It looks like these items have been resolved now. Thank you! :heart:

PR #9092, opened 2020-04-16
Additional Log::Context methods

Blacksmoke16:
* Add #empty? and #to_json(builder) to Log::Context
* Fix some doc display bugs

Implements the two method suggestions from https://forum.crystal-lang.org/t/log-module-suggestions/1962.  Will wait for some feedback on the others, this one was useful/easy enough to just go ahead and do.

### DIFF
--- a/spec/std/log/context_spec.cr
+++ b/spec/std/log/context_spec.cr
@@ -54,8 +54,11 @@ describe Log::Context do
     end
 
     describe String do
+      it "empty" do
+        c("").empty?.should be_true
+      end
+
       it "not empty" do
-        c("").empty?.should be_false
         c("foo").empty?.should be_false
       end
     end
--- a/spec/std/log/context_spec.cr
+++ b/spec/std/log/context_spec.cr
@@ -43,14 +43,11 @@ describe Log::Context do
       c(17).empty?.should be_false
     end
 
-    describe Array do
-      it "empty" do
-        c([] of Int32).empty?.should be_true
-      end
-
-      it "not empty" do
-        c([1, 2, 3]).empty?.should be_false
-      end
+    it Bool do
+      c(false).empty?.should be_false
+    end
+    it Nil do
+      c(nil).empty?.should be_true
     end
 
     describe String do
@@ -60,6 +57,16 @@ describe Log::Context do
 
       it "not empty" do
         c("foo").empty?.should be_false
+      end
+    end
+
+    describe Array do
+      it "empty" do
+        c([] of Int32).empty?.should be_true
+      end
+
+      it "not empty" do
+        c([1, 2, 3]).empty?.should be_false
       end
     end
 
--- a/spec/std/log/context_spec.cr
+++ b/spec/std/log/context_spec.cr
@@ -1,5 +1,6 @@
 require "spec"
 require "log"
+require "json"
 
 private def c(value)
   Log::Context.new(value)
@@ -35,6 +36,53 @@ describe Log::Context do
   it "merge" do
     c({a: 1}).merge(c({b: 2})).should eq(c({a: 1, b: 2}))
     c({a: 1, b: 3}).merge(c({b: 2})).should eq(c({a: 1, b: 2}))
+  end
+
+  describe "#empty?" do
+    it Number do
+      c(17).empty?.should be_false
+    end
+
+    describe Array do
+      it "empty" do
+        c([] of Int32).empty?.should be_true
+      end
+
+      it "not empty" do
+        c([1, 2, 3]).empty?.should be_false
+      end
+    end
+
+    describe String do
+      it "not empty" do
+        c("").empty?.should be_false
+        c("foo").empty?.should be_false
+      end
+    end
+
+    describe Hash do
+      it "empty" do
+        c(Hash(String, String).new).empty?.should be_true
+      end
+
+      it "not empty" do
+        c({"key" => "value"}).empty?.should be_false
+      end
+    end
+  end
+
+  describe "#to_json" do
+    it Number do
+      c(17).to_json.should eq "17"
+    end
+
+    it Array do
+      c([1, 2, 3]).to_json.should eq "[1,2,3]"
+    end
+
+    it Hash do
+      c({"key" => "value"}).to_json.should eq %({"key":"value"})
+    end
   end
 
   describe "implicit context" do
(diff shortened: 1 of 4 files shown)


spec/std/log/context_spec.cr
```diff
@@ -37,6 +38,53 @@ describe Log::Context do
...
+      end
+
+      it "not empty" do
+        c([1, 2, 3]).empty?.should be_false
+      end
+    end
+
+    describe String do
+      it "not empty" do
+        c("").empty?.should be_false
```
jhass, line 58: Should this be true? 🤔

src/crystal/datum.cr
```diff
@@ -144,13 +155,18 @@ module Crystal
       @raw.pretty_print(pp)
     end
 
+    # :nodoc:
+    def to_json(builder : JSON::Builder) : Nil
+      @raw.to_json builder
+    end
```
RX14, line 161: `Log` shouldn't depend on `JSON`. I'm pretty sure we don't already. Does `require "log"` in an empty file work after this?

straight-shoota: It doesn't depend on JSON as long as this method is not called. This is the same as `Object#to_json`.

spec/std/log/context_spec.cr
Blacksmoke16: In this case no because `#empty?` here is checking if there is _ANY_ value stored within the context.  Not if the underlying value is empty.

jhass: So empty array and empty hash are not values...?

Blacksmoke16: Hmm, yea I guess it could be `true` since an empty string would essentially be `nil`, or the same as if a value was not provided.  Also it's probably not a common use case so, I'll update that.

src/crystal/datum.cr
RX14: I thought that the type restriction would cause a parse error but I guess not

asterite: It's done like that so you don't have to use forward definitions. Then these types are only checked when called.

That said, I wouldn't discard that in the future these will be checked after all types have been defined to catch errors earlier. And that will mean this will break.

So I suggest putting this into `require "log/json"`.

spec/std/log/context_spec.cr
asterite: Nevermind `blank?`, I misread

Blacksmoke16: I updated it to consider `""` as empty.  I also ran into https://play.crystal-lang.org/#/r/8wgb 🤔.

src/crystal/datum.cr
Blacksmoke16: In that case, should I move these methods into `log/context.cr` versus `datum`?  Really these methods are useable in anything datum would be used in.

spec/std/log/context_spec.cr
asterite: It's already reported as a bug where you can't more than two types in `case`. I would put each type in a different `when`, even if the code is the same.

src/crystal/datum.cr
straight-shoota: At the current state of things this feels unjustified. You don't need to `require "object/json"` to call `to_json` on any object. Maybe in the future it will be required. But we don't actually know how this could play out. Maybe we'll end up with an entirely different solution.
So I'd not force an additional require just to use a simple delegation method.

asterite: You need to `require "json"`. `object.to_json` is otherwise not defined

straight-shoota: Yes, but but it only works if the actual type implements `#to_json(json : JSON::Builder)`.

RX14: Let's just leave this as-is now.

Either this changes before 1.0, or it changed in 2.0, because it'd be a massive breaking change anyway. Neither of these possibilities affect this PR.

bcardiff: It is not, but the files are parsed. 

Also it is not true that datum values needs to be JSON compatible. 

I think the`Log::Context#to_json` should be in a `log/json.cr` file

Blacksmoke16: Maybe something in the middle and just move it to `Log::Context`, so it's specific just to that?  I'm not so sure needing to require another file is really necessary.

src/crystal/datum.cr
```diff
@@ -53,15 +53,30 @@ module Crystal
...
     end
 
+    # Returns `true` if `self` does not currently have a value, or the value is `nil`.
+    #
+    # If the underlying value is an `Array` or `Hash` then `Enumerable#empty?` is invoked.
+    #
+    # If the underlying value is a `String`, then `String#empty?` is invoked.
+    #
+    # Otherwise, `false` is returned.
+    def empty? : Bool
```
bcardiff, line 70: Something that bothers me of `JSON::Any` and `YAML::Any` is that there is no easy way to know if the value is nested. I think that that is needed.

I found the `empty?` a bit ambiguous. Because it delegates the empty to the raw value. It is not separating fully where `size` method will work or not.

The goal was to know if the `size` will raise or not, right?

I would prefer to add a `nested?` that might return the raw value as `Array | Hash | Nil` directly. So it's like equivalent to `as_nested? = as_h? || as_a?`.

That way we can use it either as a truthy check or with `case` in a comfortable manner.

bcardiff:
* Changes in the documentation were included in #9117
* JSON support has been included in #9101
* The need for `empty?` which I think was motivated by #8994 second concert,  will be no longer needed after #9118

So I think that the only thing left here is either iterate on the idea proposed on https://github.com/crystal-lang/crystal/pull/9092#discussion_r410413902 or close it.

So much attention on the new logs module :-)

Blacksmoke16:
> * The need for `empty?` which I think was motivated by #8994 second concert,  will be no longer needed after #9118

More so for the case of me having a custom formatter doing like

```cr
...

entry.context.to_json io

...
```

However in the case the entry does not have any context, an empty object `{}` gets printed.  I was, using `#empty?` to avoid printing it at all if the context has no content.  I'm open to other options as well.

Blacksmoke16:
Actually after looking at the default formatter, I could probably just do what https://github.com/crystal-lang/crystal/blob/master/src/log/io_backend.cr#L38 is doing.  Or if we would want `#empty?` could just be that logic?

Blacksmoke16:
Handled by #9101

PR #9094, opened 2020-04-16
Standardize HTTP method parsing

grkek:
### Description

In case if the method parser is unable to find a proper match the received request method is evaluated which leads to unauthorized access of certain routes protected by request method checks. This will give the request parser a more standard way of facing sent requests to the server instance.

### Changes

Updated the method parsing, it returns an error code to the request in case of a non-capital/non-standard method name.

### Reference

Most of the other major web-server technologies respond with an error code to these kind of requests.

### Impact

Fixes all of the issues with frameworks not being able to categorize and case match lowercase method names.

### DIFF
--- a/src/http/request.cr
+++ b/src/http/request.cr
@@ -173,7 +173,7 @@ class HTTP::Request
     # Optimization: see if it's one of the common methods
     # (avoids a string allocation for these methods)
     method = METHODS.find { |method| method.to_slice == subslice } ||
-             String.new(subslice)
+             return HTTP::Status::BAD_REQUEST
 
     # Skip spaces.
     # The RFC just mentions a single space but most servers allow multiple.


grkek:
Props to this guy for discovering the issue, [issues/kemal/564](https://github.com/kemalcr/kemal/pull/564).

jhass:
I think we need to try coming up with a spec for this :)

grkek:
> I think we need to try coming up with a spec for this :)

Basically anything that is not listed in the method list will return a 400

jhass:
Something like/extending https://github.com/crystal-lang/crystal/blob/master/spec/std/http/request_spec.cr#L206-L211 should work.

straight-shoota:
This behaviour is incorrect. HTTP methods are not a closed list. You can technically define your own if you like. The parser must not prevent that. The server implementation is responsible for deciding which methods it supports and which it does not.

grkek:
> This behaviour is incorrect. HTTP methods are not a closed list. You can technically define your own if you like. The parser must not prevent that. The server implementation is responsible for deciding which methods it supports and which it does not.

If the std is so open to the method variety lets disregard the RFC and upcase all incoming request methods then, does that sound fine?

straight-shoota:
No. That also violates the RFC. Method names are case sensitive. I don't think there is anything to do here. The behaviour is expected and according to the specifications.

This needs to be handled by the frameworks. They must decide which methods to accept and how to treat them.

grkek:
Let the frameworks handle part sounds fine to me but when the method is passed in as get and it is not picked-up by the parser as the "GET" method for example in kemal it would result a bypass for the before_get "/" filter function.

straight-shoota:
Then kemal needs to make sure to handle methods appropriately.

jhass:
It does seem like a common implementation pitfall that stdlib ought to prevent. I see three options how we could improve: 

1. Upcase standard methods but pass everything else as is, that is turn `get` into `GET`, `Post` into `POST` etc but pass `FooBar` as is. 
2. Return 400 if the method matches one of the standard method but has the wrong case. 
3. Make the list of acceptable methods a configuration option that has the standards options as a default. Then return a 400 if a method is not in that list. This would probably need an option to accept arbitrary methods.

grkek:
> Then kemal needs to make sure to handle methods appropriately. It seems to apply different to handlers and filters.

You don't get the main picture here If I send a `get` request or a `GET` request they should be handled equally, @jhass the 3rd option sounds fine to me

RX14:
`get` should be the same as `GET`. Whatever the RFC says, in the real world, the common methods are not case sensitive.

For non-standard methods, leaving the case alone is fine.

Sija:
Just to note, if any error should be returned, according to the [RFC](https://tools.ietf.org/html/rfc2616#section-5.1.1), that should be `501 Not Implemented`.

grkek:
> Just to note, if any error should be returned, according to the [RFC](https://tools.ietf.org/html/rfc2616#section-5.1.1), that should be `501 Not Implemented`.

NGINX returns 400, so I thought it would be wise to use 400.

straight-shoota:
`Request.from_io` is responsible for parsing the HTTP request into a data structure. It must not alter the request but pass through what it reads. The only instance responsible for interpreting the meaning of HTTP requests (and thus methods) is `HTTP::Handler`.

In stdlib we have only one relevant implementation: `StaticFileHandler`. If we want to make that support `get` requests, that's fine. We can change that there. If it is configured with `fallthrough: false` it returns *501 Method not allowed*. Otherwise the handler chain terminates with *404 Not Found* (which might be up for debate).

For Kemal this means adjusting the method handling in `FilterHandler` and `RouteHandler`. So kemalcr/kemal#564 is actually the correct way to fix this (at least part of it).

SlayerShadow:
@RX14 
> Whatever the RFC says, in the real world, the common methods are not case sensitive.

Why do you think so?

* Nginx supports only camel-case, case-sensitive format (code: [1 - detect](https://github.com/nginx/nginx/blob/3ba88365b5acef17f01671cd969c909dee5e2cde/src/http/ngx_http_parse.c#L152); [2 - respond](https://github.com/nginx/nginx/blob/3ba88365b5acef17f01671cd969c909dee5e2cde/src/http/ngx_http_request.c#L1167) (if you follow the call chain, `rc == 10` in that case, so it goes here); [StackOverflow answer](https://stackoverflow.com/a/9645072)) . This is the core of the engine and it is case sensitive.
* Puma returns 400 on all lower-case methods ([issue](https://github.com/puma/puma/issues/1681), and [here](https://github.com/puma/puma/blob/df9cdc703b8fc7dcd79858aefa37f4128a7986f2/ext/puma_http11/http11_parser_common.rl#L38) where does it do this (did no go deeper)). This is the core of the server and it is case sensitive.
* Sinatra returns 404 ([here](https://github.com/sinatra/sinatra/blob/91477e6dad7c9d8eb1c358e47b9d83b73d626f84/lib/sinatra/base.rb#L1007) is checking and [here](https://github.com/sinatra/sinatra/blob/91477e6dad7c9d8eb1c358e47b9d83b73d626f84/lib/sinatra/base.rb#L1027) is handler). This is not a core of Rack... Sinatra works over Rack and both does not support RFC completely but in any case it is case sensitive.

Please note that Nginx is the HTTP server itself; Puma (and other popular Ruby servers) is the wrapper of raw `TCPServer`. I suppose that's the main reason why Ruby's HTTP engines handle HTTP methods manually.

---

Crystal frameworks mostly wrapped around `HTTP::Server` that, I guess, if it placed in the stdlib then at least it should warn user about unsafe methods from the outside somehow. Or like poposed in this PR - decline this behaviour.

Of course if user writes his own socket server and handles HTTP over it with a couple of vulnerabilites, there can be possible anything, including not the following for specification.

@jhass, in (1) there is restriction too in RFC: if it says that only upper case is allowed, `FooBar` should be `FOOBAR` instead. I don't see another restrictions about user-specified naming.

(3), if I understand correctly, there is a HTTP header `Allow` that does this.

I completely agree with (2). _(possible with replacing 400 with 501)_

straight-shoota:
> warn user about unsafe methods from the outside somehow. Or like poposed in this PR - decline this behaviour.

Lower case method names are not standard, but they are also not unsafe and perfectly legal with the HTTP specification.

SlayerShadow:
@straight-shoota, agreed (updated spec: [RFC7231, sec. 4](https://tools.ietf.org/html/rfc7231#section-4)). In any case, HTTP _default_ methods are all upper case:

> This specification defines a number of standardized methods that are commonly used in HTTP, as outlined by the following table. By convention, standardized methods are defined in all-uppercase US-ASCII letters.

RX14:
@straight-shoota I don't want to make it easy for crystal web frameworks to shoot themselves in the foot. This *security* bug was there in kemal because of a situation which most reasonable people wouldn't think of. To me, this indicates that it's a good idea to solve it once in the stdlib.

I'm fine with either normalizing the standard methods, or denying lowercase methods. Since NGINX only allows uppercase HTTP methods, I think that's a **really** good indication that the usecase for lower-case doesn't exist in the real world.

straight-shoota:
The security issue described in kemalcr/kemal#564 is caused by kemal treating HTTP methods differently for route and filter handlers. This is clearly a bug in Kemal and doesn't need to be fixed outside.

I'm fine with changing the behaviour for non-standard request methods, but this needs to be handled in `HTTP::Server`, not in `HTTP::Request.from_io`. A request with lower case method is still a valid request and the server can decide what to do about it. This could be implemented as a `Server::Handler` that gets implicitly inserted as first item of the handler chain, or hardwiring this inside `RequestProcessor`.

grkek:
The request parser is responsible for parsing the request and spitting out a proper Request object which wont contain any arbitrary data that might cause undefined behavior. There is no need to stress the framework developers for parsing the actual request body, if someone would want to do that they would certainly use own hand written HTTP servers and not the default one.

straight-shoota:
> The request parser is responsible for parsing the request and spitting out a proper Request object which wont contain any arbitrary data that might cause undefined behavior.

I don't follow your train of though here. Reading HTTP methods as they are written in the request is not arbitrary at all. Quite on the contrary, implicitly adjusting the casing of some HTTP methods could in fact be considered arbitrary.
Also non-standard HTTP methods must not be considered "undefined behaviour" in gneral. The behaviour is defined by the HTTP application.

> if someone would want to do that they would certainly use own hand written HTTP servers and not the default one.

I agree, except that you don't need a hand written HTTP server from scratch but can re-use a customized `HTTP::Server`. That's exactly what I proposed in https://github.com/crystal-lang/crystal/pull/9094#issuecomment-615316759

SlayerShadow:
Will it be merged before version 1.0?

jhass:
I feel like so far there's no real consensus reached on what the solution should be, but there's quite some opinions that the one presented here is not it.

asterite:
The RFC doesn't say methods outside GET/POST, etc. should be an error.

I don't understand why this is needed. It seems the bug in Kemal is a bug in Kemal. I don't see any urgency to change this before 1.0.

asterite:
I just tried writing a simple server in Go. I hit it with `curl -XFOO` and it worked fine.

I think this issue should be closed.

PR #9095, opened 2020-04-16
Remove obsolete CrystalPath#make_relative_unless_absolute method

Sija:
Followup to #9054


Sija:
@straight-shoota Could be, but since there are probably more places worth changing I'd leave it for a dedicated PR.

Issue #7204, opened 2018-12-19
`parallel` macro leaks fibers


straight-shoota:
The `parallel` macro is used to run pieces of code concurrently. Ideally, it waits for all fibers to finish and returns the combined return values of all jobs.

However, when one fiber raises, the exception is raised in the enclosing context which means it will immediately resume execution. This breaks the waiting connection to the other concurrent fibers and `send` calls to the collector channel block indeterminately. This makes them zombie fibers.

There is no optimal strategy for all use cases, but IMO it would be a better solution to remember the exception but wait for the other fibers to complete and only raise it afterwards. This makes sure that all tasks are properly finished and only raises when execution is to continue in the parent context.

Related to #6468

asterite:
We introduced parallel just for fun, to see if we could make it work. In my opinion we should remove all these quick features (parallel, future, etc.), rethink them, then reintroduce them.

Sija:
Can be closed.

PR #9097, opened 2020-04-16
Drop top-level parallel macro

bcardiff:
Since it was used in channel_spec.cr I moved a slightly different version as a private file macro. The only change is that the custom exception is gone.


straight-shoota:
Resolves #7204

PR #9100, opened 2020-04-16
Compiler: fix inconsistency of @type resolution inside macros

### LINKED ISSUE #9099: Inconsistent {{@type}} expansion in macros
If the macro is inside a module, the `@type` is referring to the container module instead of the type where the macro is expanded.

Running the docs tool with the following code exhibits the differences. But the issue affects also the generated code.

I was expecting to always be the type were the macro is executed (since the otherone can be explicitly mentioned) 

```crystal
module XXX
  macro m
    # This will be XXX (not ideal) `{{@type}}`
    def foo
      {{@type}}
    end
  end
end

macro m
  # This will be Foo (ideal) `{{@type}}`
  def bar
    {{@type}}
  end
end

class Foo
  XXX.m
  m
end

pp! Foo.new.foo # => XXX
pp! Foo.new.bar # => Foo
```

Extracted from #8994

asterite:
Fixes #9099

### DIFF
--- a/spec/compiler/codegen/macro_spec.cr
+++ b/spec/compiler/codegen/macro_spec.cr
@@ -1189,7 +1189,7 @@ describe "Code gen: macro" do
       end
 
       Foo.bar
-      )).to_string.should eq("Foo")
+      )).to_string.should eq("<Program>")
   end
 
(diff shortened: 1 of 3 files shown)


asterite:
Doesn't work.

PR #9102, opened 2020-04-16
Parser: introduce endless method syntax

### LINKED ISSUE #9080: [RFC] endless method definition
Recently, Ruby introduces a new method definition syntax called endless method definition ([ticket](https://bugs.ruby-lang.org/issues/16746) and [pull request](https://github.com/ruby/ruby/pull/2996)):

```ruby
def inc(x) = x + 1
def square(x) = x * x

p square(inc(1)) # => 4
```

It seems good to me. So, I'd propose to add this feature to Crystal. 

Thank you.

asterite:
Fixes #9080

Just for fun. Doesn't mean we are going to go with this.

Let's keep the discussions on #9080

But given that Crystal already tries to be succinct... for example we have `def initialize(@name)`, or we have `record Point, x : Int32, y : Int32`... maybe not having to write a full method body, `end` included, is nice. Specially for really short methods that are just literals, forwarding, small stuff like that.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -226,6 +226,62 @@ describe Crystal::Formatter do
   assert_format "def foo\n  1 #\nrescue\nend"
   assert_format "def foo\n  1\n  #\n\n\nrescue\nend", "def foo\n  1\n  #\n\nrescue\nend"
 
+  assert_format <<-BEFORE, <<-AFTER
+    def   foo   =   1
+    BEFORE
+    def foo = 1
+    AFTER
+
+  assert_format <<-BEFORE, <<-AFTER
+    def   foo   =
+         1
+    BEFORE
+    def foo =
+      1
+    AFTER
+
+  assert_format <<-BEFORE, <<-AFTER
+    def   foo   =   if 1
+      2
+    else
+      3
+    end
+    BEFORE
+    def foo = if 1
+                2
+              else
+                3
+              end
+    AFTER
+
+  assert_format <<-BEFORE, <<-AFTER
+    def   foo   =
+      if 1
+        2
+      else
+        3
+      end
+    BEFORE
+    def foo =
+      if 1
+        2
+      else
+        3
+      end
+    AFTER
+
+  assert_format <<-BEFORE, <<-AFTER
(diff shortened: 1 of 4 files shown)


MakeNowJust:
Ugh, I have an implementation for this RFC when I opened issue... It is said that early implementation with less discussion makes confusing sometimes, so I keep secret this patch.
Like exhaustive `case`, I'm sick that my name is missing from my own proposal!

MakeNowJust:
@asterite  I know my wish is selfish, and it goes smoothly if I hand over the baton to you, but... Could you entrust the job to me?

asterite:
You can send a different PR with the same code or similar, it's fine

asterite:
> It is said that early implementation with less discussion makes confusing sometimes, so I keep this patch secret

You are right. I think it's fine if you send an issue, then a PR saying "this is in case the issue gets accepted".

In my case I was just curious to see how hard this would be to implement, and how it felt using it. Then I sent the PR so it doesn't get lost in case we need it.

But feel free to send a separate PR

j8r:
Meh, beginners used to C style are able to not fully embrace Crystal style:
```cr
def test = (
  a = 0
  a += 1
  a += 2
)
```
That a "danger" of this proposal, which will defeat its initial purpose of declaring short methods.

Another point, on this implementation, return type annotations are still not valid:
Invalid code:
```cr
def test1 : Int32 = 0
def test2(x : Int32) = x
```
Result for both methods: `Error: unexpected token: =`

MakeNowJust:
@j8r **My implementation allows return type annotation of course**. And, I think we should not discuss about the specification in Pull Request. Please go to #9080.

asterite:
@j8r anyone can write crappy code, even before this proposal

wontruefree:
@asterite truer words have never been said

watzon:
I kinda like this. Will I use it much? Probably not. But I can see why it was added to Ruby.

MakeNowJust:
@watzon Don't discuss about syntax here. Go to #9080.

watzon:
@MakeNowJust wasn't a syntax discussion. I was voicing my support for the PR. I am fully aware of #9080.

MakeNowJust:
@watzon Sorry for harsh words.

PR #9103, opened 2020-04-16
Use correct command for running a particular spec

paulcsmith:


bcardiff:
Both work actually. But the crystal spec/file_spec.cr alternative does not allow filter by :line at the end.

paulcsmith:
@bcardiff I just ran it and it did not work 🤔

paulcsmith:
Oh I see, it works but only if you run the *exact* file.

IMO this should still be the default since it lets you do stuff like `bin/crystal spec spec/std/log/` and use line numbers

Either that or maybe a `bin/spec` that does it. But IMO this is probably a simpler/clearner way to do it. Up to you though!

PR #9104, opened 2020-04-16
Allow Nil values in Log::Context

paulcsmith:
It is often helpful to log when something is *not* set.

Right now that is not possible without weird workarounds like settings
nil to `""`. Even that doesn't always work because you may have a
complex Hash/NamedTuple where doing this is complex.

This allows setting values to `nil` and adds a spec to make sure it
works along with an example use case

> Note: this does not allow `Log::Context.set nil`. So we still protect against accidental logging of *just* nil


bcardiff:
Closed by #9121

PR #9105, opened 2020-04-16
Add Log::IOBackend#new with formatter

paulcsmith:
Makes it so you can do stuff like this:

```crystal
Log.builder.bind("*", :info, Log::IOBackend.new(formatter: MyFormatter))
```

Instead of

```crystal
backend = Log::IOBackend.new
backend.formatter = MyFormatter
Log.builder.bind("*", :info, backend)
```

### DIFF
--- a/src/log/io_backend.cr
+++ b/src/log/io_backend.cr
@@ -2,7 +2,7 @@
 class Log::IOBackend < Log::Backend
   property io : IO
   property progname : String
-  property formatter : Formatter? = nil
+  property formatter : Formatter?
 
   def initialize(@io = STDOUT, @formatter = nil)
     @mutex = Mutex.new(:unchecked)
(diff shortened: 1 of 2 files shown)


src/log/io_backend.cr
```diff
@@ -4,7 +4,7 @@ class Log::IOBackend < Log::Backend
   property progname : String
   property formatter : Formatter? = nil
```
bew, line 5:
```suggestion
  property formatter : Formatter?
```
since it's set to nil in the initialize ?

paulcsmith: Yep!

bcardiff:
@paulcsmith I've just noticed that your intended usage was with named argument and the code allows also positional.

I think is better to enforce named argument in this case since there might be other (optional) options in the future.

paulcsmith:
@bcardiff Yeah that makes sense! Good idea

Issue #9089, opened 2020-04-15
cl.exe `c:crystalllvm-10.0.0.srcReleaseinllvm-config.exe --libs --system-libs --ldflags 2> /dev/null`


oprypin:
With PR #9054 in (currently declared as the last step for [bootstrapping the compiler on Windows](https://github.com/crystal-lang/crystal/issues/9032)), one can *almost almost* [compile the compiler for Windows and then *on* Windows](https://github.com/oprypin/crystal/wiki/Porting-to-Windows#building-crystal-for-windows), but at the end, the following is passed to the linker as part of the command line:

`cl.exe`[…]``C:\crystal\src\llvm/ext/llvm_ext.o `c:crystalllvm-10.0.0.srcReleaseinllvm-config.exe --libs --system-libs --ldflags 2> /dev/null` stdc++.lib``[…]

(sure, one can edit that and run the linker manually, which *works*, but that kinda sours the experience)

Let's break this one down.

## llvm-config
The main culprit, of course, is this line:
https://github.com/crystal-lang/crystal/blob/1b62767c266ca2c80ab12ebbb57a3ad5e86af996/src/llvm/lib_llvm.cr#L12 Which declares that, whenever this `lib` is involved, we should insert a POSIX shell command expansion into the linker command:
1. starting with a string (LLVM path) pasted into source code, causing funny characters `\c`&rarr;c, `\b`&rarr;backspace, etc. but that's easy to fix with correct interpolation, whatever.
2. wanting to run llvm-config, which actually would've worked fine with these flags.
3. trying to redirect stderr into a POSIX-specific device.
4. expects the result of that command to be expanded dynamically and inserted back.

Sooo yes, we can't do it like this on Windows, at all. I'm kinda blanking on what the alternative solution would be. That's why I created this issue rather than just fixing it.

We need to not rely on shell expansion and in general `ldflags` is not cross-platform.
We *could* devise a custom way to do intentional delayed expansion of a subcommand within the linker command, but it's messy.

## llvm_ext
Then there's also the other part to this but it's really minor in comparison:
https://github.com/crystal-lang/crystal/blob/1b62767c266ca2c80ab12ebbb57a3ad5e86af996/src/llvm/lib_llvm_ext.cr#L2 which
1. should be <i>llvm_ext.o<ins><b>bj</b></ins></i>
2. we don't have a Makefile to rely on to pre-build this file from *llvm_ext.cc*

## stdc++.lib
[...]

PR #9106, opened 2020-04-16
Also expand `subcommands` in Windows linker flags

oprypin:
This is a hack to get the compiler to compile itself on Windows without interventions.

See #9089

### DIFF
--- a/src/compiler/crystal/compiler.cr
+++ b/src/compiler/crystal/compiler.cr
@@ -321,13 +321,16 @@ module Crystal
       stdout.puts command.sub(%("${@}"), args && args.join(" "))
     end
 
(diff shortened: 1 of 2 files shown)


asterite:
In fact, this is how we did it in the beginning. The problem was that for cross compile you don't want to expand that in the host system. But maybe this code isn't triggered on cross compile anymore?

oprypin:
Yes I skip it on cross-compile

oprypin:
Agh, needed like 6 attempts before I finally pushed the commits in the right way. Sorry, CI :(

straight-shoota:
@oprypin If we need to buy additional containers for circleci, we'll send you the bill 😛

oprypin:
Can we merge, then? :upside_down_face:

PR #9107, opened 2020-04-16
Move verbose to a lower level than debug

paulcsmith:
Typically "verbose" is the most verbose output you can get and implies
verbose will give you "more"

Right now "debug" is lower than verbose which does not seem like it
matches how most people think of or are using these levels.

My proposal is to make "verbose" the most "verbose" log level possible.

For example and ORM might use "debug" to show what queries run, and what
the args are, but might have "verbose" mode that tells you when the
transactions start, when it was committed, whether there was a cache
hit, etc. Stuff that normally is not helpful, but if you *really* need
it you can go "verbose" and get all the nitty gritty details

### DIFF
--- a/spec/std/log/broadcast_backend_spec.cr
+++ b/spec/std/log/broadcast_backend_spec.cr
@@ -34,9 +34,9 @@ describe Log::BroadcastBackend do
     main.append(backend_b, s(:error))
 
     log = Log.new("", main, s(:info))
-    log.level = s(:verbose)
+    log.level = s(:info)
 
-    log.verbose { "lorem" }
+    log.info { "lorem" }
 
     backend_a.entries.should_not be_empty
     backend_b.entries.should_not be_empty
--- a/spec/std/log/log_spec.cr
+++ b/spec/std/log/log_spec.cr
@@ -16,8 +16,8 @@ describe Log do
 
   describe Log::Severity do
     it "values are ordered" do
-      s(:debug).should be < s(:verbose)
-      s(:verbose).should be < s(:info)
+      s(:verbose).should be < s(:debug)
+      s(:debug).should be < s(:info)
       s(:info).should be < s(:warning)
       s(:warning).should be < s(:error)
       s(:error).should be < s(:fatal)
--- a/spec/std/log/log_spec.cr
+++ b/spec/std/log/log_spec.cr
@@ -16,26 +16,29 @@ describe Log do
 
   describe Log::Severity do
     it "values are ordered" do
-      s(:verbose).should be < s(:debug)
+      s(:trace).should be < s(:debug)
       s(:debug).should be < s(:info)
-      s(:info).should be < s(:warning)
+      s(:info).should be < s(:notice)
+      s(:notice).should be < s(:warning)
       s(:warning).should be < s(:error)
       s(:error).should be < s(:fatal)
       s(:fatal).should be < s(:none)
     end
 
     it "parses" do
+      Log::Severity.parse("trace").should eq s(:trace)
       Log::Severity.parse("debug").should eq s(:debug)
-      Log::Severity.parse("verbose").should eq s(:verbose)
       Log::Severity.parse("info").should eq s(:info)
+      Log::Severity.parse("notice").should eq s(:notice)
       Log::Severity.parse("warning").should eq s(:warning)
       Log::Severity.parse("error").should eq s(:error)
       Log::Severity.parse("fatal").should eq s(:fatal)
       Log::Severity.parse("none").should eq s(:none)
 
+      Log::Severity.parse("TRACE").should eq s(:trace)
       Log::Severity.parse("DEBUG").should eq s(:debug)
-      Log::Severity.parse("VERBOSE").should eq s(:verbose)
       Log::Severity.parse("INFO").should eq s(:info)
+      Log::Severity.parse("NOTICE").should eq s(:notice)
       Log::Severity.parse("WARNING").should eq s(:warning)
       Log::Severity.parse("ERROR").should eq s(:error)
       Log::Severity.parse("FATAL").should eq s(:fatal)
@@ -47,9 +50,10 @@ describe Log do
     backend = Log::MemoryBackend.new
     log = Log.new("a", backend, :warning)
 
+    log.trace { "trace message" }
     log.debug { "debug message" }
-    log.verbose { "verbose message" }
     log.info { "info message" }
+    log.notice { "notice message" }
     log.warn { "warning message" }
     log.error { "error message" }
     log.fatal { "fatal message" }
@@ -67,9 +71,10 @@ describe Log do
 
     log.level = :error
 
+    log.trace { "trace message" }
     log.debug { "debug message" }
-    log.verbose { "verbose message" }
     log.info { "info message" }
+    log.notice { "notice message" }
     log.warn { "warning message" }
     log.error { "error message" }
     log.fatal { "fatal message" }
@@ -86,9 +91,10 @@ describe Log do
     backend = Log::MemoryBackend.new
     log = Log.new("a", backend, :debug)
 
+    log.trace(exception: ex) { "trace message" }
     log.debug(exception: ex) { "debug message" }
-    log.verbose(exception: ex) { "verbose message" }
     log.info(exception: ex) { "info message" }
+    log.notice(exception: ex) { "notice message" }
     log.warn(exception: ex) { "warning message" }
     log.error(exception: ex) { "error message" }
     log.fatal(exception: ex) { "fatal message" }
--- a/src/log.cr
+++ b/src/log.cr
@@ -3,7 +3,7 @@
 # The messages, or `Log::Entry` have associated levels, such as `Info` or `Error`
 # that indicate their importance. See `Log::Severity`.
 #
-# To log a message `debug`, `verbose`, `info`, `warn`, `error`, and `fatal` methods
+# To log a message `trace`, `debug`, `info`, `notice`, `warn`, `error`, and `fatal` methods
 # can be used. They expect a block that will evaluate to the message of the entry.
 #
 # ```
--- a/src/log/entry.cr
+++ b/src/log/entry.cr
@@ -1,19 +1,28 @@
 # A logging severity level.
 enum Log::Severity
-  Verbose
+  # Used for tracing the code and trying to find one part of a function specifically.
+  Trace
+  # Used for information that is diagnostically helpful to people more than just developers (IT, sysadmins, etc.).
   Debug
+  # Used for generally useful information to log.
   Info
+  # Used for normal but significant conditions.
+  Notice
+  # Used for conditions that can potentially cause application oddities, but that can be automatically recovered.
   Warning
+  # Used for any error that is fatal to the operation, but not to the service or application.
   Error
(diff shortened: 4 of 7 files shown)


waj:
Actually we put debug as the lowest level because we thought we could actually get rid of debug calls when the program is compiled in non debug (release) mode. I'm not saying that we're actually doing this, but it was an idea.

In other words, I think "debug" stands for something so detailed that only a developer might be interested. And "verbose" is an increased level of detail that still a user might want to enable.

In your case, maybe queries should be logged as "verbose" and maybe use "debug" to log actual details about query results.

waj:
IIRC we checked syslog levels also to check that "debug" was the lowest level: https://en.wikipedia.org/wiki/Syslog

I can see other logging frameworks have a "trace" level under "debug". Maybe we could add that to have two debug levels.

waj:
Actually I like the idea of having a "trace" level. That would leave two levels for normal operation (info and verbose) and two levels for debugging (debug and trace). What do you think?

Blacksmoke16:
Given how `verbose` is a bit ambiguous, I think I'd rather go with `trace`/`debug`, and `info`/`notice` to make the hierarchy a bit more clear/defined.

straight-shoota:
TRACE doesn't sound like a bad idea. Maybe that could be disabled by default, and only compiled when a debug flag is set? Just an idea.

But also, VERBOSE is not a very common severity level. I can understand that it's difficult to place.
Maybe we should better remove it entirely. TRACE would be a replacement (and also in the spirit of this PR). We could also consider adding a NOTICE level which many logging systems have between WARN and INFO.

paulcsmith:
I love the idea of disambiguating it by making a `trace` that is the lowest level and where you get *everything*. I'm not sure about what to call the others, but I'd be ok keeping it debug and getting rid of verbose and making it trace as the lowest level. 

And the idea to remove at compile time could be cool. I'm not sure I'd use it, but might be helpful!

straight-shoota:
I like this classification from https://stackoverflow.com/a/2031209/7401689

> * Trace - Only when I would be "tracing" the code and trying to find one part of a function specifically.
> * Debug - Information that is diagnostically helpful to people more than just developers (IT, sysadmins, etc.).
> * Info - Generally useful information to log (service start/stop, configuration assumptions, etc). Info I want to always have available but usually don't care about under normal circumstances. This is my out-of-the-box config level.
> * Warn - Anything that can potentially cause application oddities, but for which I am automatically recovering. (Such as switching from a primary to backup server, retrying an operation, missing secondary data, etc.)
> * Error - Any error which is fatal to the operation, but not the service or application (can't open a required file, missing data, etc.). These errors will force user (administrator, or direct user) intervention. These are usually reserved (in my apps) for incorrect connection strings, missing services, etc.
> * Fatal - Any error that is forcing a shutdown of the service or application to prevent data loss (or further data loss). I reserve these only for the most heinous errors and situations where there is guaranteed to have been data corruption or loss.

waj:
But NOTICE is used for something more important than INFO that is still not a WARNING. Wikipedia entry for syslog says: "Conditions that are not error conditions, but that may require special handling"
It's quite common that apps has a "--verbose" flag and I liked how that could match with that level.

waj:
> I like this classification from https://stackoverflow.com/a/2031209/7401689

The definition of DEBUG is the one that I might disagree. For Wikipedia for syslog says: "Messages that contain information normally of use only when debugging a program". And Crystal (and every language) uses "debug" for things that are just for developers.

straight-shoota:
I thought so at first, too. But the crucial question is, who debugs a program. I wouldn't say that's strictly limited to developers. As a sysadmin I would definitely try debug logging with a program that doesn't work as intended, even if I would never consider working on it as a developer.
The exact classification depends on the use case and can be very different depending on the type of program. In the end it doesn't really matter *who* is the target of the severity level, as long as we agree that it is detailed diagnostic information that you wouldn't want to see in a normal production scenario.

It might also be worth noting that log messages at DEBUG level typically have no long-term value, so they shouldn't be collected

paulcsmith:
I also like the classification but would consider adding one level between INFO and WARN -> NOTICE. Which as @waj was saying it is more than just info, but not quite a warning.

This is a preference though and I'd definitely use it for things like "This form failed to save" kind of thing, but warning would work if we don't do notice

waj:
So I think if we add `Notice` and `Trace`, but we keep `Verbose` we please everybody, right? 

```crystal
Fatal > Error > Warning > Notice > Info > Verbose > Debug > Trace
```

And now there are two levels for each kind (problems, caution, information and debugging).

But... are these too many levels? 😛 
Maybe not. I wouldn't mind adding all these. Nobody is obligated to use all of them and you can use a subset as you wish.

paulcsmith:
I am kind of with you on adding them. Yes there are a lot, but you don't have to use them, and for those that do want more granularity, this is super helpful

Usually I like less, but this does seem like a good balance to me with little downside. It's all backward compatible and has the common ones people will try without looking at the docs 👍

straight-shoota:
I'd definitely remove VERBOSE. What other log system uses this?
With NOTICE and INFO there shouldn't be any need to have another severity in that area, especially it if is not implicitly clear where to place it.

waj:
But I like `Verbose` 😢. As I said I like how it maps with the standard "verbose" mode that programs usually have. If we had only `Notice` and `Info`, I'd have to use `Notice` for standard logging and `Info` for verbose entries. That doesn't sound right to me.

straight-shoota:
> That doesn't sound right to me.

Why not? INFO is what I'd usually want to see with `--verbose`. It's generally useful information, but you usually don't care about it. If you have logging configured to go to syslog/file/logserver etc. you would log this level by default, but a CLI wouldn't show it unless explicitly asked to do so (because you usually don't care).

waj:
Because for me `Info` is the default level I'd like the logging framework to be. Maybe for a CLI you want to change that to `Warning`, I don't know. For Shards for example, we're already using `Log.info` for all the information normally given to the user. If you want more info, then enable verbose mode.

For a HTTP server for example, every request should be logged as `Info`. Again, it's the default level and every request is printed to the output.

straight-shoota:
So what would be the difference between `--verbose` and DEBUG level? I think this could also line up pretty well.

RX14:
I'm fine with info > debug > trace or info > verbose > debug

both make sense to me, but i don't think we need 3 levels under "info"

oprypin:
FTR, **verbosity** is often used as a concept orthogonal to **severity**.

In some cases it's a user-facing feature (ever seen `-v -v -v` in Linux tools), in other cases it's a more advanced debugging feature.

For example Google does this:

https://gitcdn.link/repo/google/glog/master/doc/glog.html#verbose
https://github.com/google/glog/blob/3ba8976592274bc1f907c402ce22558011d6fc5e/src/glog/logging.h.in#L219

In case you ever want to possibly re-introduce it as such, it's better to stay away from the name `Verbose` for now.

paulcsmith:
> FTR, verbosity is often used as a concept orthogonal to severity

This is a great point and I think that is why I was so confused. To me it means "what you have, give me more".

> Fatal > Error > Warning > *Notice* > Info > ~Verbose~ > Debug > Trace

Mostly because I do feel verbose is a bit confusing as a term

**However** I know @waj feels strongly and if others do too I am not that against it. The main things I'd like is *something* below debug (whether that is trace or moving verbose) and having "Notice" between warning and info

waj:
@paulcsmith no, I don't feel strongly. I just had a usage in mind already and it was hard to, you know, change my mind. But after sleeping on it I think I now understand the concerns and I'm ready to let it go 🙂 

So, these levels sound better then?

```crystal
Fatal > Error > Warning > Notice > Info > Debug > Trace
```

I still don't know when I'm going to use `Notice` but if others find it useful we can add it.

I'd like descriptions like the ones @straight-shoota shared, added to each severity level as documentation.

BTW, I found this announcement (https://github.com/aspnet/Announcements/issues/124) where they had exactly the same levels as we have right now, and they decided to make a similar change. The underlying discussion have some interesting research done: https://github.com/aspnet/Logging/issues/299

paulcsmith:
@waj thank you for being so willing to change your mind! I like your proposed log levels. 

I would be ok without Notice, but it would a nice to have. Here's an example:

Right now Avram (Lucky's ORM) logs a warning when an record fails to save because of validation errors. This is more important than info, but not quite a warning or error. So NOTICE would fit in well there. But it is a nice to have so if not added that is ok

Great links and I agree on adding those descriptions to the docs!

didactic-drunk:
> ```crystal
> Fatal > Error > Warning > Notice > Info > Debug > Trace
> ```

For those who want syslog compatibiity I suggest (as someone mentioned above) keeping the default log level at `Notice`.  User/Sysadmin level debugging should use `Info` since syslog is often configured to ignore `Debug`.

Should the suggested usage be documented?

waj:
@didactic-drunk no... I just mentioned syslog to show where "debug" was positioned within the severities. I still think the default level should be "info". Of course there might be different requirements on some applications, so it will be just the default default :)

@paulcsmith do you want to update the PR with the described levels?

didactic-drunk:
>  I still think the default level should be "info".

@waj That creates a problem.  Services using syslog wouldn't have an easy way to increase verbosity without multiple steps and often leading to confusion as increasing the programs verbosity apparently does nothing.  How long does it take sysadmins to figure out they need to reconfigure syslog the first time they do this?  20mins?  More?

didactic-drunk:
> so it will be just the default default :)

Logging at `Info` by default creates bigger problems.  Shards and crystal logging at `Info` are incompatible with any program or shard doing it differently.  So no, it's not "just a default default".  It's the way every program and shard will need to operate to get consistent logging.  All of them will be incapable of increasing logging using syslog without additional sysadmin intervention to change the syslog configuration.

straight-shoota:
@didactic-drunk Default levels are very different in different logging frameworks and reach from DEBUG to WARN. Either is probably a fine choice for a generic default. So going for INFO which is roughly in the middle sounds okay. Maybe NOTICE would be better to reduce log flooding. I'm not strongly on either side.
But the key aspect is that applications have different use cases and need to configure custom defaults that fit for them.

didactic-drunk:
@straight-shoota If `HTTP` (or other base classes or shards) use `Log.debug` by default and "verbose" is enabled the messages will show when logging to a file but won't show in syslog.  Is that intended behavior?  What is the intended behavior?

straight-shoota:
I don't follow what you're talking about. All I see is that it doesn't seem to be significant for this topic. Please open a separate issue with a comprehensible description of the problem.

bcardiff:
@paulcsmith re: https://github.com/crystal-lang/crystal/pull/9107#issuecomment-615563313

We would need to add a Notice level above Info, and probably the nice descriptions pointed out in https://github.com/crystal-lang/crystal/pull/9107#issuecomment-614930697 and the following couple of comments.

If you want to hand off these changes let us know.

paulcsmith:
@bcardiff I have become a bit busy this week 🤪

If someone else would like to make the changes and take this over I'd appreciate it! If not I can try to tackle this next week

bcardiff:
Rebased on master.

Applied the changes add trace and notice, drop verbose.

I removed verbose directly, without deprecation warning. But I could do that if wanted. I'm not sure it's worth it.

PR #9110, opened 2020-04-17
Some `Log` improvements

Blacksmoke16:
Implements the first suggestion from https://forum.crystal-lang.org/t/log-module-suggestions/1962.

* Simplify `initializer` and `set` overloads for `Log::Context`
  * `#new` and `#set` now allows any type of hash and calls `.to_s` on each key
* Add overloads to `Log` to support passing entry specific context.  Including:
  * Hash
  * NamedTuple
  * named arguments
* Add `Log#supports?(severity)` that returns `true` if a given logger is able to handle the provided severity
  * Usecase would be to check if a given logger would log before logging so that context arguments do not have to be executed if it would not log anyway.  Mainly in the case where the context would require some none trivial calculations.

### DIFF
--- a/spec/std/log/io_backend_spec.cr
+++ b/spec/std/log/io_backend_spec.cr
@@ -14,6 +14,16 @@ private def io_logger(*, stdout : IO, config = nil, source : String = "", progna
   builder.for(source)
 end
 
+private def assert_logged(expected : Regex, & : Log -> Nil)
+  IO.pipe do |r, w|
+    logger = io_logger(stdout: w)
+
+    yield logger
+
+    r.gets.should match expected
+  end
+end
+
 describe Log::IOBackend do
   it "logs messages" do
     IO.pipe do |r, w|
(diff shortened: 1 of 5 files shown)


paulcsmith:
@Blacksmoke16 Cool changes! I have a thought on how you could avoid using `supports?` for performance critical logging.

You could flip this method around so that the first arg is just a String, and the context goes in a block:

```crystal
# And disallow string return in block if you give the message as first arg to 'info'
Log.info("My event") { {my_data: expensive_calculation} } 
```

Though you could still leave `supports?` since it may just be a handy method and could almost certainly be used internally

## With the proposal above^

* We get better perf without the need for `supports?`
* Also means we could have `Log.info { {my_data:expensive_calculation} }` that just sets message to `nil`

I would really love this. After our discussion yesterday I think I could use this and fulfill all the needs of the Lucky formatter! I can just get the `message` and merge it with the context if it is set 👍 I'd also need the data stored in *some* key other than the top-level. See `Store in context under 'local' key?` below. 

As you mentioned in Gitter this uses the existing implementation, breaks nothing, **but** it would allow for greater formatter flexibility and clarifies global from entry specific context
## Simplify initializer and types

> Simplify initializer and set overloads for Log::Context
> #new and #set now allows any type of hash and calls .to_s on each key

Could you explain this a bit more? I'm not for or against it, just trying to understand the "why" behind it :)

## Store in context under 'local' key?

We kind of talked about this in Gitter, but what do you think about putting the data in a `local` key so it is clear this is not normal "global" context. This would still work with existing formatters and provide more info to log consumers about where it comes from. It *also* makes it so formatters can do some cool stuff!

In fact if you did that change and moved the local context to the block (mainly for perf) that'd totally solve all my needs we discussed as well. Seems like a total win to me!

Blacksmoke16:
@paulcsmith I think this is a better proposal than what you were suggesting before.  However, given that it would be a breaking change I would suggest maybe opening a new issue for it in order to gather feedback.

paulcsmith:
@Blacksmoke16 Thank you! How is this a breaking change though? Maybe I didn't explain it clearly because I'm only suggesting adding a new method overload

```crystal
# still works!
Log.info { "my message" } 
# This is new and is basically your proposal but with context in the block for perf reasons
Log.info("my message") { {my_data: "foo"} } 
# This is also new. Same as above, but sets message to ""
Log.info { {my_data: "foo" } }  
```

Log::Entry remains the same. Only adds the one new method with String message overload and modifying `info` to check if the block result is a NamedTuple or Hash, and if it is, set it in the context under the local key. I already did this in Dexter and pretty simple to do

Blacksmoke16:
Ah ok, yea just adding this is an additional overload would be fine I think.  I'm going to let this stand for a while to see what others think first.

straight-shoota:
It is really difficult to track unrelated discussions about three individual changes in a single thread. Could you please open separate issues to discuss them?

Also, the previous talk has been mostly brain storming, so it would be preferred to discuss the proposals in issues, not in a PR which already includes an implementation.

paulcsmith:
@straight-shoota I'm fine with that: https://forum.crystal-lang.org/t/proposal-to-have-all-data-be-key-value-data-in-log/1966/11?u=paulcsmith

didactic-drunk:
Would it be simpler to pass a context to the block?
```
log.info { "foo" }
log.info do |ctx|
  ctx.set :a, 1
  ctx.set {  b: 2, c: 3}
  "foo"
end
```

`ctx.set` should probably return an ignorable value (maybe `Context` which also allows method chaining) for cases when messages aren't used.

bcardiff:
> Add Log#supports?(severity) that returns true if a given logger is able to handle the provided severity

binding of log can change along the way. So it is not sound to have that method.

> #new and #set now allows any type of hash and calls .to_s on each key

👎. The only override I would allow is `Hash(Symbol, V)` but not for any key.

---

Why you appended a `: Nil` to the `def initialize` methods?

Blacksmoke16:
> binding of log can change along the way. So it is not sound to have that method.

Not even for this use case?  I suppose it wouldn't catch the case where the ENV var or somethings changes just after it completes the check.  However, I'd still think it would be useful as I don't see that happening that often.

```cr
if Log.handles? :info
  Log.info(id: some_method) { "Some message" }
end
```

The idea was to avoid having the context execute if the log isn't going to actually be logged.

> The only override I would allow is `Hash(Symbol, V)` but not for any key.

That and `String` keys would be the most common, but is there a reason to just not support `{1 => "str"}`?  Given the idea of context is just dynamic user supplied data, this would allow them to just pass any hash and it just work.

> Why you appended a `: Nil` to the `def initialize` methods?

Habit. But I guess in this case it's kinda redundant.

bcardiff:
@Blacksmoke16 

> Not even for this use case?  ...

With #9146 you will be able to do the following, without affecting the current Fiber.context after the log.

```crystal
if Log.handles? :info
  Log.info { Log.context.set(id: some_method); "Some message" }
end
```

Since the context is extended on demand, the usecase should be covered. And it even allows you big dynamic context to be created if wanted.

> ... this would allow them to just pass any hash and it just work.

It is never _any_ hash. There has to be some constraint. For example, the story of creating a JSON of the context requires the keys to be translated to a JSON key. And the values are also constrained.

I understand you want to call `K#to_s` so everything will be `String` in the `Context`. But is odd that will be set as `K` and read as `String` IMO.

To me, either you use named arguments for an explicit key, or you build a hash-like with String keys.

Blacksmoke16:
@bcardiff I'll close this PR for now and open a new one with just the additional context overloads once #9146 is merged to avoid conflicts.

PR #9111, opened 2020-04-17
refactor process_extra_assigns

wontruefree:
@asterite : I thought this might be a little more concise.

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -3520,7 +3520,11 @@ module Crystal
 
     def process_extra_assigns(body, extra_assigns)
       return body if extra_assigns.empty?
-      body.is_a?(Expressions) ? extra_assigns.concat body.expressions : extra_assigns.push body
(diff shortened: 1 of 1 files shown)


src/compiler/crystal/syntax/parser.cr
```diff
@@ -3519,18 +3519,9 @@ module Crystal
...
-          exps.concat body.expressions
-        else
-          exps.push body
-        end
-        body = Expressions.from(exps)
-      else
-        body
-      end
+      return body if extra_assigns.empty?
+      body.is_a?(Expressions) ? extra_assigns.concat body.expressions : extra_assigns.push body
```
straight-shoota, line 3523: The ternary operator is more confusing than helping. It should stay a regular if.

yxhuvud: That said, the code when rewritten to a regular if statement is still quite a bit simpler than the original.

asterite:
Thank you, but this seems to mutate the method argument, something which wasn't happening in the original. I think the original code is fine.

PR #9113, opened 2020-04-17
Parser: allow combination of anonymous splat and block arg

### LINKED ISSUE #9108: Can't use anonymous block arg with only named arguments
```cr
def run(*, id : Int32, &)
  yield id
end
 
run id: 17 # => Error: duplicated argument name:
```

https://play.crystal-lang.org/#/r/8wjn

MakeNowJust:
Fixed #9108

### DIFF
--- a/src/compiler/crystal/syntax/parser.cr
+++ b/src/compiler/crystal/syntax/parser.cr
@@ -3570,8 +3570,16 @@ module Crystal
         next_token_skip_space_or_newline
         block_arg = parse_block_arg(extra_assigns)
         skip_space_or_newline
-        if (!block_arg.name.empty? && args.any?(&.name.==(block_arg.name))) || (found_double_splat && found_double_splat.name == block_arg.name)
-          raise "duplicated argument name: #{block_arg.name}", block_arg.location.not_nil!
+        # When block_arg.name is empty, this is an anonymous argument.
+        # An anonymous argument should not conflict other arguments names.
+        # (In fact `args` may contain anonymous splat argument. See #9108).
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/syntax/parser.cr
```diff
@@ -3570,7 +3570,7 @@ module Crystal
         next_token_skip_space_or_newline
         block_arg = parse_block_arg(extra_assigns)
         skip_space_or_newline
-        if args.any?(&.name.==(block_arg.name)) || (found_double_splat && found_double_splat.name == block_arg.name)
+        if (!block_arg.name.empty? && args.any?(&.name.==(block_arg.name))) || (found_double_splat && found_double_splat.name == block_arg.name)
```
MakeNowJust, line 3573: Which is better condition, it or the following:

```crystal
        if !block_arg.name.empty? && (args.any?(&.name.==(block_arg.name)) || (found_double_splat && found_double_splat.name == block_arg.name))
```

It is same meaning because `found_double_splat` never be empty named.

jhass: I prefer the current but maybe to extract both sides into two more locals?

```cr
duplicate_block_arg_name = !block_arg.name.empty? && args.any?(&.name.==(block_arg.name)
block_name_conflicts = found_double_splat && found_double_splat.name == block_arg.name
if duplicate_block_arg_name || block_name_conflicts

MakeNowJust:
@jhass I refactored it and added comment to describe the reason. What do you think?

PR #9115, opened 2020-04-17
Improve error handling and logging in HTTP::Server

### LINKED ISSUE #9067: HTTP::Server errors are silently ignored
These lines of code:

https://github.com/crystal-lang/crystal/blob/9f9b0a86b19f78e2bc83121c5c3c99d9df59e8fa/src/http/server/request_processor.cr#L50-L53

are not doing what one would expect. As soon as `response.respond_with_error` is called, the connection is closed and the next lines are not executed (I'm not sure if another exception is raised or what). Then you never get to see the error in the output.

I also think we should probably use `Log` here too.

waj:
Currently `HTTP::Server` logs (or at least it tries: #9067) every error, including those produced by a sudden client disconnection. These changes aim to:

 * Log everything through `Log` api
 * Ignore errors raised when a client connection is lost
 * Don't write back headers (with status 500) or error messages when there was some content sent already

Changes are made to `HTTP::Server::RequestProcessor`, `HTTP::ErrorHandler` and `HTTP::LogHandler`. The last one used to receive an optional `IO` to customize where to write the logs, so this could be seen as a breaking change. Although not so big, I think.

fixes #9067


spec/std/http/server/response_spec.cr
```diff
@@ -214,5 +214,16 @@ describe HTTP::Server::Response do
       response.respond_with_status(HTTP::Status::URI_TOO_LONG, "Request Error")
       io.to_s.should eq("HTTP/1.1 414 Request Error\r\nContent-Type: text/plain\r\nContent-Length: 18\r\n\r\n414 Request Error\n")
     end
+
+    it "closes when if fails to write" do
```
bcardiff, line 218:
```suggestion
    it "closes when it fails to write" do
```

src/http/server/handlers/log_handler.cr
```diff
@@ -1,20 +1,18 @@
 # A handler that logs the request method, resource, status code, and
-# the time used to execute the next handler, to the given `IO`.
+# the time used to execute the next handler, to the log with source `http.server`.
 class HTTP::LogHandler
   include HTTP::Handler
-
-  # Initializes this handler to log to the given `IO`.
-  def initialize(@io : IO = STDOUT)
```
bcardiff, line 7: This is a breaking-change.

Ideally we could add a `@Deprecated` to a `def initialize(io : IO)` constructor even if we are no longer using that `io`.

As you prefer.

waj: Done

waj: Not done... I broke everything 🙄

spec/std/http/server/request_processor_spec.cr
```diff
@@ -324,4 +324,18 @@ describe HTTP::Server::RequestProcessor do
...
+      context.response.content_type = "text/plain"
+      context.response.print "Hello "
+      context.response.flush
+      context.response.print "world"
+      raise "OH NO"
+    end
+    input = IO::Memory.new("GET / HTTP/1.1\r\n\r\n")
+    output = IO::Memory.new
+    processor.process(input, output)
+    output.rewind.gets_to_end.should match(/world/)
```
straight-shoota, line 306: Maybe not as much necessary as above, but matching the entire expected response would be great.

spec/std/http/server/handlers/error_handler_spec.cr
```diff
@@ -33,4 +53,31 @@ describe HTTP::ErrorHandler do
...
+  it "doesn't write errors when the output is closed" do
+    io = RaiseIOError.new
+    request = HTTP::Request.new("GET", "/")
+    response = HTTP::Server::Response.new(io)
+    context = HTTP::Server::Context.new(request, response)
+
+    handler = HTTP::ErrorHandler.new
+    handler.next = ->(ctx : HTTP::Server::Context) { ctx.response.flush }
+    handler.call(context)
+  end
```
straight-shoota, line 47: This example misses an expectation.

spec/std/http/server/handlers/error_handler_spec.cr
```diff
@@ -33,4 +53,31 @@ describe HTTP::ErrorHandler do
...
+    context = HTTP::Server::Context.new(request, response)
+
+    handler = HTTP::ErrorHandler.new
+    handler.next = ->(ctx : HTTP::Server::Context) do
+      ctx.response.print "Hi"
+      ctx.response.flush
+      raise "OH NO!"
+    end
+    handler.call(context)
+    io.to_s.should_not match(/500 Internal Server Error/)
```
straight-shoota, line 62: Ditto as above, this should specify the expected io content.

src/http/server/handlers/log_handler.cr
```diff
@@ -10,7 +10,9 @@ class HTTP::LogHandler
 
     Log.info { "#{context.request.method} #{context.request.resource} - #{context.response.status_code} (#{elapsed_text})" }
   rescue e
-    Log.error(exception: e) { "#{context.request.method} #{context.request.resource} - Unhandled exception:" }
+    unless context.response.closed?
```
straight-shoota, line 22: What's the reason for this? I'd expect an error to show up in the logs, even if the io is closed. The response could even have been sent!

src/http/server/request_processor.cr
```diff
@@ -47,18 +50,22 @@ class HTTP::Server::RequestProcessor
         begin
           @handler.call(context)
         rescue ex
-          response.respond_with_status(:internal_server_error)
-          error.puts "Unhandled exception on HTTP::Handler"
-          ex.inspect_with_backtrace(error)
+          unless response.closed?
+            Log.error(exception: ex) { "Unhandled exception on HTTP::Handler" }
```
straight-shoota, line 54: How does this play with `HTTP::ErrorHandler`? In case the error handler triggers, there won't be a log entry, right?

src/http/server/handlers/log_handler.cr
```diff
@@ -1,20 +1,27 @@
...
     elapsed = Time.measure { call_next(context) }
     elapsed_text = elapsed_text(elapsed)
 
-    @io.puts "#{context.request.method} #{context.request.resource} - #{context.response.status_code} (#{elapsed_text})"
+    Log.info { "#{context.request.method} #{context.request.resource} - #{context.response.status_code} (#{elapsed_text})" }
   rescue e
-    @io.puts "#{context.request.method} #{context.request.resource} - Unhandled exception:"
-    e.inspect_with_backtrace(@io)
+    unless context.response.closed?
+      Log.error(exception: e) { "#{context.request.method} #{context.request.resource} - Unhandled exception:" }
```
straight-shoota, line 23: Slightly different story but also worth fixing: In case `Log.info` or `elapsed_text` fails, this `rescue` block would be triggered. I don't think that's expected because it's not an error related to the HTTP application, but the internal logging setup. The `rescue` should just cover `call_next(context)`, not even `Time.measure` (which is also an internal detail and failing that is not relevant to the HTTP protocol).

src/http/server/handlers/log_handler.cr
straight-shoota: There should be an initializer accepting a `Log` instance so users can use a custom configured one. The legacy initializer accepting `IO` could be implemented that way, creating a custom `Log` instance that logs to the given io.

src/http/server/handlers/log_handler.cr
waj: I was just thinking the same because even though the client might have closed the connection there was still some cpu being burned, so something must be logged. We could actually log just as a warning (or notice if we accept #9107) because there is nothing doable to fix the problem.
Would you still log the exception in that case?

src/http/server/request_processor.cr
waj: Right, `HTTP::ErrorHandler` swallows all the exceptions already. This is used when `HTTP::Server` is used without handlers or when something more severe happens.

waj:
> There should be an initializer accepting a Log instance so users can use a custom configured one
Why? That's the point of the new logging framework, that provides flexibility to change the backend for each source. Otherwise, would you provide flexibility to change any logger in the std?

src/http/server/handlers/log_handler.cr
straight-shoota: As long as it is an app error, it should definitely be logged. The IO might have actually been closed after the exception was raised.

src/http/server/request_processor.cr
straight-shoota: So if you only use `ErrorHandler` you get less than without it (with respect to error logging) and you need to include `LogHandler` to go back to the behaviour that was already provided by plain `RequestProcessor`. This seems a bit awkward.

straight-shoota:
You can easily have different HTTP servers running side by side. And I would like to treat the log messages from a server that serves say an internal JSON API differently than from a server that serves public HTML requests. I don't want all log entries mangled at `http.server` but separate log sources at `myapp.public_http` and `myapp.json_api`.
This could technically be implemented as a context-based filter. But using a different log source is more approriate. Context-based filtering could handle different vhosts for example.

Such a flexibility is certainly not necessary for typical loggers. But `LogHandler` is an explicit tool to configure log handling for an HTTP application and it should offer some options for customizability.

src/http/server/handlers/log_handler.cr
waj: Under normal circumstances the output shouldn't be closed at that point.

The intention is to avoid logging as error the `IO::Error` caused by a client disconnection. Thus at `Response` when there is an error writing or flushing to the output it just mark the `Response::Output` as closed and re-raise the exception.

Maybe I should take a different approach, because the original `IO::Error` might have been handled and the exception rescued here is a different one. I'm thinking about wrapping the `IO::Error` within a more distinguishable exception here, like `ClientError`. So, unless the handler code doesn't rescue that exception (if only there was a way to make some exceptions not so easy to catch 😛), it will bubble up but we know for sure it doesn't have to be logged as error. It would still mark the output as closed so we don't keep trying to write and get the same error again. The only drawback I see is that raising a new exception makes it unwind the stack again.

I'll try to work around that idea and improve this.

src/http/server/request_processor.cr
waj: Well, that's currently explained in the `ErrorHandler` doc. Maybe we could remove the logging from `RequestProcessor` so it doesn't get less 😆 

The way I see it, the handler at `RequestProcessor` is there as last resort when no middleware are used. Would you combine `ErrorHandler` and `LogHandler` in the same class? Or remove it altogether and have always the same behavior at `RequestProcessor`?

Anyway, I didn't change that in this PR and these classes are getting as old as Crystal I think. Let's open a RFC or a separate PR to modify this stack.

src/http/server/handlers/log_handler.cr
waj: Agreed. I also noticed that but I was trying to focus just on the logging and not handling client disconnections for this PR.
I'll make the change anyway.

src/http/server/handlers/log_handler.cr
RX14: I don't really mind so much whether it accepts a `Log`, but it must at least accept the source string.

src/http/server/handlers/log_handler.cr
RX14: I think `IO::Error` should contain a reference to the `IO` which errored, then you can compare against the request and response IOs, and you know whether the `IO::Error` was related to the client or to another unrelated error.

Alternatively, we can just try not to be too smart and stop hiding client errors.

src/http/server/request_processor.cr
RX14: Can we remove `ErrorHandler`?

src/http/server/handlers/log_handler.cr
waj: Using the reference to the `IO` would break when the `IO` is actually wrapped. We would be forced to handle all errors and re-raise them updating the value.

Regarding hiding client errors, some days ago you agreed it was right to not show client disconnections as errors 🙂: https://github.com/crystal-lang/crystal/pull/9034#discussion_r406977081

I think this is (in part) what is making people complain about "broken pipe" errors in the logs all the time.

waj: Now the concerns related with closed IO and client errors are separated

src/http/server/request_processor.cr
waj: What I did was sneak peek how rack solves this issue and do something similar. Effectively, `LogHandler` must be the first handler to work properly and it doesn't handle or logs any error. It just logs requests. The `ErrorHandler`, if used, will send appropriate responses to the client if possible, and also log the errors.

So now, if you only use `LogHandler` the requests are logged, and the errors are still handled by the `RequestProcessor` (but not sent to the client).

If you only use `ErrorHandler` the errors are sent to the client and logged, just like the `RequestProcessor` do, although we might want to make it more detailed in the future. Another reason I found to log errors from here is that I think it's more clear that the exception is logged _before_ the request as it happens when `LogHander` is used alone.

src/http/server/handlers/log_handler.cr
waj: Ok, solved on this PR

waj:
I think I covered all the comments and changes requested. I tested it a lot with a "real" app in different scenarios and the results feel reasonable to me.
I also added the remote address and HTTP version to the log entries to make it more similar to other frameworks.

spec/std/http/server/handlers/log_handler_spec.cr
```diff
@@ -1,35 +1,57 @@
...
-    expect_raises(Exception, "foo") do
-      handler.call(context)
+    logs = capture_logs("http.server") do
+      expect_raises(Exception, "foo") do
+        handler.call(context)
+      end
     end
-    (log_io.to_s =~ %r(GET / - Unhandled exception:)).should be_truthy
+    match_logs(logs,
+      {:info, %r(^- - GET / HTTP/1.1 - 200 \(\d+(\.\d+)?[mµn]s\)$)}
```
straight-shoota, line 54: The logged status code is plainly wrong here. The server actually responds with 500, not 200 and the log must reflect that.
I'm not sure what's the best way to make this accurate. The `LogHandler` probably needs to differentiate whether the handler chain was successfully completed or an exception was raised. The exception part would then need to duplicate some of the logic from `RequestProcessor` to tell the expected response status.

Another option would be to integrate `LogHandler` directly into `RequestProcessor`. This would avoid duplication of the error handling and centralize logging to a single place. It would obviously need to be configurable.

The purpose of `ErrorHandler` should also be considered in this regard. It seems quite useless when its only effect is to change this:
```
I, [2020-04-22T15:21:41.548787000Z #5279]   INFO -- crystal-run-server.tmp:http.server: 127.0.0.1:38650 - GET / HTTP/1.1 - 200 (74.30µs)
E, [2020-04-22T15:21:41.548973000Z #5279]  ERROR -- crystal-run-server.tmp:http.server: Unhandled exception on HTTP::Handler -- Exception: foo
```
To this:
```
E, [2020-04-22T15:31:56.224456000Z #5412]  ERROR -- crystal-run-server.tmp:http.server: Unhandled exception -- Exception: foo
I, [2020-04-22T15:31:56.224721000Z #5412]   INFO -- crystal-run-server.tmp:http.server: 127.0.0.1:50044 - GET / HTTP/1.1 - 500 (376.90µs)
```

waj: I also realized the problem with the status codes. That happens when only `LogHandler` is used. In Rack the log handler writes logs without handling errors so when this happens the request is not even logged. I think it's better (not ideal though) to log the wrong status code than not logging anything at all.

The purpose of `ErrorHandler` is handling errors, showing them to the client and writing the logs. I think the issue is the placement of the `LogHandler` and the fact that both handlers must be used at the same time to work properly. Rack also has the same "problem" but it setups both handlers automatically by default. Maybe we could do that but probably this could be better if the design is changed a little bit. Maybe `LogHandler` cannot be a regular `HTTP::Handler` and it should be a strategy instead, plugged in the `RequestProcessor`.

src/http/server/handlers/log_handler.cr
```diff
@@ -1,21 +1,31 @@
...
-# the time used to execute the next handler, to the given `IO`.
+# the time used to execute the next handler
 class HTTP::LogHandler
   include HTTP::Handler
 
-  # Initializes this handler to log to the given `IO`.
-  def initialize(@io : IO = STDOUT)
+  @[Deprecated("Use `new([Log])` instead")]
+  def initialize(io : IO)
+    @log = Log.for("http.server")
```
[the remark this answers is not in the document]
waj: Yup, good observation. I'll change that.

waj:
> We can't have three (!) different locations where status information about a request gets logged. That's a total mess.

Well, I agree just 50%. It's not actually status information that it's logged in three places. Request status is logged just in one place: `LogHandler`. Is the error logging that it's duplicated between `LogHandler` and `ErrorHandler`

> Maybe we need to consider the whole approach to these handlers, see my comment on log_handler_spec.cr

I agree and I've been discussing this for more than 1h with @bcardiff at the phone and I've been moving around the code several times without settling for any option. I was trying to keep the Rack design but it's not possible to "make it right" this way. As I said above, the alternative is make the `LogHandler` not a regular `Handler` but something that's plugged to the `RequestProcessor`. 

Anyway, I think this is an improvement, and actually a fix, over the design that we currently have. That again, it's not "our" design, but a translation of the Rack design. I'll open a RFC with a proposal to make this different but for now we can merge this that solves issues we currently have without changing the design too much.

RX14:
I'd prefer to have the logging built into the server, and not a handler. We already have existing ways to configure logging levels, and I think a handler ends up duplicating that.

waj:
Yes, maybe not embedded because I’d like to keep it configurable in case anyone want to log differently, but definitely not as a handler. I’ll return to this once I close more priority things.

PR #9118, opened 2020-04-17
Log: validate Fiber.logging_context is a hash

bcardiff:
Ref: #8994

### DIFF
--- a/src/log/context.cr
+++ b/src/log/context.cr
@@ -51,5 +51,11 @@ end
 
 class Fiber
   # :nodoc:
-  property logging_context : Log::Context = Log::Context.new
+  getter logging_context : Log::Context = Log::Context.new
+
+  # :nodoc:
+  def logging_context=(value : Log::Context)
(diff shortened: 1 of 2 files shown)


src/log/context.cr
```diff
@@ -51,5 +51,11 @@ end
 
 class Fiber
   # :nodoc:
-  property logging_context : Log::Context = Log::Context.new
+  getter logging_context : Log::Context = Log::Context.new
+
+  # :nodoc:
+  def logging_context=(value : Log::Context)
+    raise ArgumentError.new "Expected hash context, not #{value}" unless value.as_h?
```
Sija, line 58:
```suggestion
    raise ArgumentError.new "Expected hash context, not #{value.class}" unless value.as_h?
```

bcardiff: Is there anything wrong with showing the value in the exception?

straight-shoota: Showing only the type is more concise and I don't think we need the value here.
It would also be confusing when `value.to_s` looks like a stringified hash.

PR #9119, opened 2020-04-18
Refactor Docs::Generator source link generation

straight-shoota:
This refactors how the docs generator generates source links to hosted git providers. All configuration is done via `ProjectInfo` which is already responsible for name and version properties.

The default behaviour is still to extract the remote tracking branch from the current git checkout and derive the URL pattern from there if it is a known git provider. But there is now also an option `--source_url_pattern` to configure the URL pattern explicitly, which can be used to point source links to providers not covered by the automatic recognition and also works outside a git repository. `--source-refname` accompanies that use case and is used to configure the refname in source URL links.

Changes to the generated HTML output are minimal and shouldn't have a perceivable effect. The only visual difference is that the locations in a type's locations list are now ordered. Internal data-id values have changed because they are no longer based on the repository but the configured name.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ compiler_spec: $(O)/compiler_spec ## Run compiler specs
 
 .PHONY: docs
 docs: ## Generate standard library documentation
-	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION)
+	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION) --source-refname=$(CRYSTAL_CONFIG_BUILD_COMMIT)
 
 .PHONY: crystal
 crystal: $(O)/crystal ## Build the compiler
--- a/man/crystal.1
+++ b/man/crystal.1
@@ -176,12 +176,36 @@ Options:
 .Pp
 .It Fl -project-name Ar NAME
 Set the project name. The default value is extracted from shard.yml if available.
+
 In case no default can be found, this option is mandatory.
 .It Fl -project-version Ar VERSION
 Set the project version. The default value is extracted from current git commit or shard.yml if available.
+
 In case no default can be found, this option is mandatory.
 .It Fl -json_config_url Ar URL
 Set the URL pointing to a config file (used for discovering versions).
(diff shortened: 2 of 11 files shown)


bcardiff:
Can't the project_name and the repository_name be different?

I would expect the former to don't have `crystal-` `-crystal` or `.cr` while the second one does. Does this PR play nice with that?

straight-shoota:
My idea was that `project_name` replaces `repository_name`. I don't hink it makes much sense to have the latter. It was my understanding that it only existed because no alternative was available. Now we have `project_name` to identify the project, which is much better than the repository name. Shards are identified by name, not by repo.

We could add properties `repository_url` and `homepage_url`, though. They could be used to provide links to those relevant locations. I've planned this as one of the following steps, but don't think it's super important because most shards already include these links in the readme.

Makefile
```diff
@@ -91,7 +91,7 @@ compiler_spec: $(O)/compiler_spec ## Run compiler specs
 
 .PHONY: docs
 docs: ## Generate standard library documentation
-	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION)
+	$(BUILD_PATH) ./bin/crystal docs src/docs_main.cr $(DOCS_OPTIONS) --project-name=Crystal --project-version=$(CRYSTAL_VERSION) --source-refname=$(CRYSTAL_CONFIG_BUILD_COMMIT)
```
bcardiff, line 94: if there is a fallback to the local working copy git information, why this change?

straight-shoota: Because it should also work if the source code is not a git working directory (for example from a tarball). Then `CRYSTAL_CONFIG_BUILD_COMMIT` can be configured externally.

bcardiff:
I have some concerns regarding adding git.sr.ht as a known git provider. Is stated to be in alpha stage in their own site. I would rather leave that out of the PR. That's my only concern.

@RX14 I'm not sure if you aim to review this as per @straight-shoota request.

RX14:
Would be nice to know what changed since my last review :/

straight-shoota:
Nothing has changed. I just fixed merge conflicts in project_info.cr and man/crystal.1 due to #9074. It was apply both changes every time.

straight-shoota:
@bcardiff I don't think sr.ht is a problem. IMO having as much source code providers available as possible is a good thing. I addded all that I have at least some knowledge of how their URLs work.

RX14:
me too, sr.ht isn't going anywhere

PR #9120, opened 2020-04-18
Add .each and .each_with_index to various macro types

Blacksmoke16:
* Extracted from 131fcba77536997be90f7ca9b214a05ee77faab7 and 63e37f03dfa0be78f0b78799473598cd24b57c94

Since #9091 is in a holding pattern; I figured it would good to at least add some of macro methods that it was adding.  I was actually able to reimplement my DI refactor (minus the compiler passes) with just this PR, and a lot of duplication.  This will satisfy my needs until #9091 can go through a proper design phase.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -556,7 +556,7 @@ module Crystal
       end
     end
 
-    describe "array methods" do
+    describe ArrayLiteral do
       it "executes index 0" do
         assert_macro "", %({{[1, 2, 3][0]}}), [] of ASTNode, "1"
       end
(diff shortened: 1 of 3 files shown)


RX14:
Isn't this redundant with macro `{% for`?

Though I'm happy to add this, I'd rather have one or the other eventually, not both

Blacksmoke16:
> Isn't this redundant with macro `{% for`?

No because:
```cr
{% begin %}
  {% hash = {"one" => {data: [[2, 2, 2], 2, 3]}, "two" => {data: [4, 5, 6]}} %}

  {% for key, value in hash %}
    {% arguments = value[:data].map do |v|
         if v.is_a? ArrayLiteral
           c = 0
           # Can't iterate over the inner array since you cant nest macro expressions
           c
         else
           v + 1
         end
       end %}
    {{pp arguments}}
  {% end %}
{% end %}
```

This enables that.  I'm sure you could rewrite this to use `{% for ..` but It would be quite an unwieldy beast.

Blacksmoke16:
Can I get another review on this?

Blacksmoke16:
Anything left to do?  Would be nice to get this into the nightly build.

straight-shoota:
Thanks @Blacksmoke16

PR #9122, opened 2020-04-18
Implement Process.executable_path by using Win32 API

kubo:
This patch uses Win32 API to get the executable path of the current process.

### DIFF
--- a/src/process/executable_path.cr
+++ b/src/process/executable_path.cr
@@ -90,6 +90,24 @@ end
       "/proc/self/exe"
     end
   end
+{% elsif flag?(:win32) %}
+  require "crystal/system/windows"
+  require "c/libloaderapi"
+
+  class Process
+    private def self.executable_path_impl
+      Crystal::System.retry_wstr_buffer do |buffer, small_buf|
+        len = LibC.GetModuleFileNameW(nil, buffer, buffer.size)
+        if 0 < len < buffer.size
+          break String.from_utf16(buffer[0, len])
+        elsif small_buf && len == buffer.size
+          next 32767 # big enough. 32767 is the maximum total path length of UNC path.
+        else
+          break nil
+        end
+      end
+    end
+  end
 {% else %}
   # openbsd, ...
   class Process
--- a/src/lib_c/x86_64-windows-msvc/c/libloaderapi.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/libloaderapi.cr
@@ -1,0 +1,5 @@
+require "c/winnt"
+
+lib LibC
+  fun GetModuleFileNameW(hModule : HMODULE, lpFilename : LPWSTR, nSize : DWORD) : DWORD
+end
--- a/src/lib_c/x86_64-windows-msvc/c/winnt.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/winnt.cr
@@ -9,6 +9,7 @@ lib LibC
   alias LPWCH = WCHAR*
 
   alias HANDLE = Void*
+  alias HMODULE = Void*
 
   INVALID_FILE_ATTRIBUTES      = DWORD.new!(-1)
   FILE_ATTRIBUTE_DIRECTORY     =  0x10


src/process/executable_path.cr
```diff
@@ -90,6 +90,25 @@ end
       "/proc/self/exe"
     end
   end
+{% elsif flag?(:win32) %}
+  require "crystal/system/windows"
```
straight-shoota, line 94: Why is this necessary?

asterite:
Thank you!

Could you change these `pending_win32` to `it` in this same PR?

https://github.com/crystal-lang/crystal/blob/2d41d70f13a55d0fe2a171edfa2c1f076a129f74/spec/std/process_spec.cr#L305-L322

kubo:
> Could you change these `pending_win32` to `it` in this same PR?
> 
> https://github.com/crystal-lang/crystal/blob/2d41d70f13a55d0fe2a171edfa2c1f076a129f74/spec/std/process_spec.cr#L305-L322

This PR is for `Process.executable_path`. The specs are for `Process.find_executable`. It isn't reasonable to include unrelated spec in a same PR. In addition only one of the three pending tests passed.

asterite:
Oops, sorry. Nevermind then.

straight-shoota:
Having a proper spec for `executable_path` would be nice, but I don't think that's in scope for this PR.

src/process/executable_path.cr
```diff
@@ -90,6 +90,25 @@ end
...
+  require "c/libloaderapi"
+
+  class Process
+    private def self.executable_path_impl
+      hMod = LibC.GetModuleHandleW(nil)
+      Crystal::System.retry_wstr_buffer do |buffer, small_buf|
+        len = LibC.GetModuleFileNameW(hMod, buffer, buffer.size)
```
asterite, line 101: From the [docs](https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew):

> hModule: A handle to the loaded module whose path is being requested. If this parameter is NULL, GetModuleFileName retrieves the path of the executable file of the current process.

So maybe we don't need to call `GetModuleHandleW`?

kubo: Thanks for pointing. I'll remove `GetModuleHandleW`.

oprypin:
Umm many of these Win32 things are already implemented, maybe you should hold off on separate implementations

src/process/executable_path.cr
kubo: I tested it on my local machine and pushed a new commit (568e1ad).

oprypin:
Yes, I think the implementation in https://github.com/crystal-lang/crystal/pull/8518/files#diff-5f66ba2dbcb4e101da5364aa2808471b is better

oprypin:
Ah no, this one is likely better, I confused them :sweat_smile:

src/process/executable_path.cr
```diff
@@ -90,6 +90,24 @@ end
...
+  class Process
+    private def self.executable_path_impl
+      Crystal::System.retry_wstr_buffer do |buffer, small_buf|
+        len = LibC.GetModuleFileNameW(nil, buffer, buffer.size)
+        if 0 < len < buffer.size
+          break String.from_utf16(buffer[0, len])
+        elsif small_buf && len == buffer.size
+          next 32767 # big enough. 32767 is the maximum total path length of UNC path.
+        else
+          raise ::File::Error.from_winerror("Error getting executable path", file: "?")
```
oprypin, line 106: Wait, what does `file: "?"` mean?

kubo: I should have heard opinions about it first.

At first I wrote the line without `file: "?"` and got the following error.
```
In src/file/error.cr:11:27

 11 | File::NotFoundError.new(message, **opts)
                          ^--
Error: missing argument: file
```
I came up with two ideas. One is using another error class. The other is supplying a dummy file name such as `"?"`. I thought that the error was related to file so the `File::Error` was proper. So I chose the latter. However I have not been satisfied with it yet.

Any idea in order not to use `file: "?"`?

oprypin: Hm well why is is `::File::Error` in the first place? That is what requires a `file` arg.

BTW I don't see any other `executable_path_impl` raising any errors, they just return `nil`. Maybe it should be that?

If not, maybe go back to `RuntimeError` and possibly specialize the exception in a different direction?

straight-shoota: It should just return `nil` if the file can't be found, as per https://crystal-lang.org/api/0.34.0/Process.html#executable_path-class-method

kubo: @straight-shoota 
Thanks. I fixed it just to return `nil` on error.

PR #9123, opened 2020-04-18
Win CI: Bootstrap Crystal, build things *on* Windows, publish the binary

oprypin:
Example:
First run: https://github.com/crystal-lang/crystal/runs/598199987
With cache: https://github.com/crystal-lang/crystal/runs/598303791

Sadly, LLVM here, if not cached, takes 1 hour to build. Although future builds will not need to build it, this might be too much reliance on the cache. Ideally Crystal org would provide a pre-built LLVM download. There is no official download that has the necessary "*.lib" files. Zig lang [provides LLVM downloads](https://github.com/ziglang/zig/wiki/Building-Zig-on-Windows) but surely we can't just use it.

In the future, if there is also a permanent place to grab a "crystal.exe" from, we can just drop the Linux step from this.

### DIFF
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -9,9 +9,11 @@ jobs:
     steps:
       - name: Download Crystal source
         uses: actions/checkout@v2
+
       - name: Install LLVM
         run: |
           apt-get -q update; apt-get -qy install llvm-8-dev
+
       - name: Build Crystal
         run: |
           make
@@ -31,10 +33,12 @@ jobs:
       - name: Disable CRLF line ending substitution
         run: |
           git config --global core.autocrlf false
-      - name: Download Crystal source
-        uses: actions/checkout@v2
       - name: Enable Developer Command Prompt
         uses: ilammy/msvc-dev-cmd@1eed9c12159dd2d2f8804980fbba1abe36ff77d2
+
+      - name: Download Crystal source
+        uses: actions/checkout@v2
+
       - name: Cache libraries
         id: cache-libs
         uses: actions/cache@v1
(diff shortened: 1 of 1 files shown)


bcardiff:
We can move de GitHub Action to build LLVM for windows to diatributions-script repo. That is already done for the omnibus osx packages using CircleCI.

Once built we can drop it in the same S3 bucket.

waj:
Did you try with binaries released at https://releases.llvm.org/download.html#10.0.0 or https://github.com/llvm/llvm-project/releases?
Otherwise, we could fork the llvm repository within crystal-lang org and include the build process there, releasing the binary within the GitHub releases.

oprypin:
@waj Yes. From my post:

> There is no official download that has the necessary "*.lib" files.

@bcardiff's suggestion seems good.

This will work as is anyway and can be merged. The building and caching is very stable.
I think we should keep things like this until Crystal 0.35 release, then both llvm and crystal.exe can be tackled.

oprypin:
I was just thinking, maybe this approach to LLVM isn't even bad. Everything is declared explicitly, which is nice. But even better, with the cache this doesn't have any outbound traffic as far as GitHub is concerned :)

oprypin:
Well, the Mac CI failure is obviously unrelated.
I just [double-checked](https://github.com/oprypin/crystal/actions/runs/82969651) that this action still passes with latest master (which wasn't a certainty...)
Please review, this is ready to go :>

asterite:
@oprypin rebasing on top of `master` will make CI pass

oprypin:
Yes but that's a heavy-handed move, as reviewers would need to double check that the commits have the same content.
I'm really against this practice of rebasing over nothing. Usually I'd **merge** master, but as I want the commits to look nice for a non-squash merge, that's also not a good option.
There's no point in re-running CI anyway, as this addition has no way to affect anything existing.

oprypin:
Ah well, not a big deal here :grimacing:

oprypin:
Oh it is a big deal apparently. Does GitHub now dismiss reviews on force push? Nice.

asterite:
> Does GitHub now dismiss reviews on force push? Nice.

It's a setting we flipped on.

oprypin:
please :s

straight-shoota:
As you wish =) Thank you!

PR #9126, opened 2020-04-18
Fix Enum#to_s for private enum

straight-shoota:
`Enum#to_s` no longer relies on constant lookup which would fail if the type is declared with private visibility.

Resolves #9044

### DIFF
--- a/src/compiler/crystal/tools/doc/html/type.html
+++ b/src/compiler/crystal/tools/doc/html/type.html
@@ -54,12 +54,12 @@
     Defined in:
   </h2>
   <% locations.each do |location| %>
-    <% if url = location.url %>
-      <a href="<%= url %>#L<%= location.line_number %>" target="_blank">
-        <%= location.filename %><% if location.show_line_number %>:<%= location.line_number %><% end %>
+    <% if url = project_info.source_url(location) %>
+      <a href="<%= url %>" target="_blank">
+        <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
       </a>
     <% else %>
-      <%= location.filename %><% if location.show_line_number %>:<%= location.line_number %><% end %>
+      <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
     <% end %>
     <br/>
   <% end %>
--- a/src/compiler/crystal/tools/doc/html/type.html
+++ b/src/compiler/crystal/tools/doc/html/type.html
@@ -54,12 +54,12 @@
     Defined in:
   </h2>
   <% locations.each do |location| %>
-    <% if url = project_info.source_url(location) %>
-      <a href="<%= url %>" target="_blank">
-        <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
+    <% if url = location.url %>
+      <a href="<%= url %>#L<%= location.line_number %>" target="_blank">
+        <%= location.filename %><% if location.show_line_number %>:<%= location.line_number %><% end %>
       </a>
     <% else %>
-      <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
+      <%= location.filename %><% if location.show_line_number %>:<%= location.line_number %><% end %>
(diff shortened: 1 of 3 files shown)


src/enum.cr
```diff
@@ -139,10 +139,11 @@ struct Enum
...
       String.build { |io| to_s(io) }
     {% else %}
-      case value
-      {% for member in @type.constants %}
-      when {{@type}}::{{member}}.value
-        {{member.stringify}}
+      {% for member, i in @type.constants %}
+        # Can't use `case` here because case with duplicate values do
+        # not compile, but enums can have duplicates (such as `enum Foo; FOO = 1; BAR = 1; end`).
+        {{ (i > 0 ? "elsif" : "if").id }} value == {{@type.constant(member)}}
```
asterite, line 145: It's easier if we always use `if` and then `return` inside it.

bcardiff:
There are other places in enum.cr that uses `{{@type}}::{{member}}`. I guess those should be changed also, but is still a workaround of not been able to pass private types to macros.

src/enum.cr
```diff
@@ -139,10 +139,11 @@ struct Enum
     {% if @type.has_attribute?("Flags") %}
       String.build { |io| to_s(io) }
     {% else %}
-      case value
-      {% for member in @type.constants %}
-      when {{@type}}::{{member}}.value
```
waj, line 144: Why can't we just use `when {{member}}.value` here?
Also, is there a good reason why `Foo::FOO` isn't accessible from within the type? Isn't that the real bug?

src/enum.cr
```diff
@@ -139,10 +139,11 @@ struct Enum
...
-        {{member.stringify}}
+      {% for member, i in @type.constants %}
+        # Can't use `case` here because case with duplicate values do
+        # not compile, but enums can have duplicates (such as `enum Foo; FOO = 1; BAR = 1; end`).
```
waj, line 144: Can we add a spec for enums with duplicates?

src/enum.cr
straight-shoota: `undefined constant FOO`... but that should probably work.

asterite: The problem is that type lookup doesn't take into account where the type lookup starts from. So trying to access `Foo::FOO` from anywhere will fail. Similar to `private` methods, you can only reference them without qualifying them.

That said, we can try to fix it...

The problem is, what happens if you have `Foo::Bar::Baz::Qux` and `Bar` is a private type inside `Foo`. Trying to find `Bar` inside `Foo` will fail. It'll need a big hack for it to work.

asterite: `{{member}}` doesn't work because the enum member will be looked up relative to `Enum`, not the actual enum type. This is how type lookup works (constants can't be found in subtypes).

It's all a big mess... probably because we never spec'd it. I think the current workaround is fine. The issues with `private` seems to only affect this enum macro code.

waj: Oh, I get it. So this method actually should be generated within the subclass (the enum)?
Maybe using the no so well accepted `macro inherited`

waj: Anyway, I shouldn't diverge from the issue here. Just I wasn't aware of this issue. Resolving this conversation...

src/compiler/crystal/tools/doc/html/type.html
```diff
@@ -54,12 +54,12 @@ <h2>
...
+    <% if url = project_info.source_url(location) %>
+      <a href="<%= url %>" target="_blank">
+        <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
       </a>
     <% else %>
-      <%= location.filename %><% if location.show_line_number %>:<%= location.line_number %><% end %>
+      <%= location.filename_in_project %><% if location.show_line_number %>:<%= location.line_number %><% end %>
     <% end %>
     <br/>
   <% end %>
```
jhass, line 65: Ups :)

straight-shoota: Somewhere in between I messed up with `git stash`

asterite:
CI is red

Issue #8984, opened 2020-04-01
[RFC] Enum members style guide


asterite:
I think it's time we decide on what we **prefer** when naming enum members. As usual, the only requirement is that they start with an uppercase letter. But in the standard library we should be consistent.

Two options:
- PascalCase
- SCREAM_CASE

I prefer PascalCase. The reason is that, even though enum members can be seen as constants, general constants can hold arbitrary types, but enum members only integers. So if you see `Foo::SCREEM_CASE` you will probably think it's some cached value, or a singleton instance. But if you see `Foo::Bar` being assigned to something, or compared against, you will probably think it's an enum member. (It could be a type too, but it's less common to move and compare types around.)

Also, I think this looks nicer (all of this is subjective, of course):

```
Color::Red
Color::Green
Color::SkyBlue

# vs.
Color::RED
Color::GREEN
Color::SKY_BLUE
```

But I'd like to know your opinion.

If we agree on a style, we should apply it throughout the standard library, and write about it in the style guide.

asterite:
Let's also vote to see trends.
- PascalCase 😄 
- SCREEM_CASE 🚀

Blacksmoke16:
What about `Pascal_Snake_Case`?  This way you would be able to tell the difference between a type, a constant, and enum member visually.

Sija:
_PascalCase_ IMO is nicer for the eyes but in the case of abbreviations _SCREAM_CASE_ is more readable: `HTMLAttr` vs `HTML_ATTR`

girng:
For me, it depends on the context. If I'm creating an enum for game commands for a gameserver, I like it to all be uppercase.  If i'm creating a list of Colors or maybe Settings, I prefer PascalCase

straight-shoota:
Duplicate of #7270. I would like to continue the discussion there.

asterite:
Oh, I searched "enum style guide" or similar but couldn't find that other issue. Thanks! I'll close this.

But noting that PascalCase was voted unanimously, so I think we should go with that.

Issue #7270, opened 2019-01-05
Consistent capitalisation of enum values/constants


dwightwatson:
Rolling on from some discussion in #7247 it's been pointed out that as it stands there is no convention for the naming of values on an enum. Through the Crystal code base there are both examples of `SNAKE_UPPERCASE` and `PascalCase` implementations.

```crystal
module Colorize
  enum ColorANSI
    Default = 39
    Black = 30
    # ...
  end
end
```

```crystal
class CSV::Builder
  enum Quoting
    NONE
    RFC
    # ...
  end
end
```

I think it would be ideal if a convention could be established here for the consistency of the codebase.

One option I see as preferable is that constants will be `SNAKE_UPPERCASE` and that enum values will be `PascalCase`. One additional benefit from this would be possible (generally) to infer if a value is a constant or an enum from its capitalisation. 

For example, `Crypto::Blowfish::DEFAULT_ROUNDS` would be a constant, but `Colorize::ColorANSI::Default` would be an enum.

Related, and not sure if it's possible at all, but if a convention was decided on it would be great if the Crystal formatter could automatically handle the change.

straight-shoota:
Enum members are semantically like constants, so I don't follow why they should have different naming conventions.

Choosing one style or the other is mostly a question of preference and heritage/sentimentality (others might call it religion). Languages in the Java family typically use UPPERCASE for enum members, whereas C# and related languages use PascalCase. Of the more modern languages, Rust and Go advocate PascalCase, Kotlin accepts both. (These mentions are only referring to the stdlib naming conventions not what the respective language supports)

Crystal, too has previously accepted both, but PascalCase is more dominant. UPPERCASE is mostly but not exclusively used when creating an enum to represent grouped C constants.

While being mostly equivalent, there are a few practical differences:

* PascalCase is actually less expressive because not all characters have upper and lower case variants. These characters are unable to denote word boundaries which can lead to ambiguities. In pure PascalCase, an enum member like `X86_64` would be `X8664` which means loosing detail. Or it would need to insert an underscore after all (could also be a different separator).
* UPPER_SNAKE_CASE aligns perfectly with lower_snake_case which is used in the auto generated predicate methods on enum types. Following up on the previous point this avoids any confusion whether the respective predicate method for `Foo1` is `#foo1?` or `#foo_1?`.
[...]

PR #9127, opened 2020-04-18
Prefer HTTP basic authentication in OAuth2 client

crush-157:
As discussed in forum post [Can't get OAuth2 access token with grant type client credentials](https://forum.crystal-lang.org/t/cant-get-oauth2-access-token-with-grant-type-client-credentials/1958), the preferred behaviour should be to use HTTP Basic authentication to pass client  credentials.

This PR provides the option to specify HTTP Basic or passing the credentials in the request body when creating the client, with HTTP Basic as the default.

### DIFF
--- a/spec/std/oauth2/client_spec.cr
+++ b/spec/std/oauth2/client_spec.cr
@@ -48,7 +48,6 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("redirect_uri=&grant_type=authorization_code&code=SDFhw39fwfg23flSfpawbef")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
@@ -67,14 +66,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("grant_type=password&username=user123&password=monkey&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http"
 
           token = client.get_access_token_using_resource_owner_credentials(username: "user123", password: "monkey", scope: "read_posts")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("grant_type=password&username=user123&password=monkey&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
@@ -86,14 +84,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("grant_type=client_credentials&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http"
 
           token = client.get_access_token_using_client_credentials(scope: "read_posts")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("grant_type=client_credentials&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
@@ -105,14 +102,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("grant_type=refresh_token&refresh_token=some_refresh_token&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http"
 
           token = client.get_access_token_using_refresh_token(scope: "read_posts", refresh_token: "some_refresh_token")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("grant_type=refresh_token&refresh_token=some_refresh_token&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
@@ -125,14 +121,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("client_id=client_id&client_secret=client_secret&redirect_uri=&grant_type=authorization_code&code=SDFhw39fwfg23flSfpawbef")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http", auth_scheme: OAuth2::AuthScheme::RequestBody
 
           token = client.get_access_token_using_authorization_code(authorization_code: "SDFhw39fwfg23flSfpawbef")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("client_id=client_id&client_secret=client_secret&redirect_uri=&grant_type=authorization_code&code=SDFhw39fwfg23flSfpawbef")
           token.access_token.should eq "access_token"
         end
       end
@@ -144,14 +139,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("client_id=client_id&client_secret=client_secret&grant_type=password&username=user123&password=monkey&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http", auth_scheme: OAuth2::AuthScheme::RequestBody
 
           token = client.get_access_token_using_resource_owner_credentials(username: "user123", password: "monkey", scope: "read_posts")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("client_id=client_id&client_secret=client_secret&grant_type=password&username=user123&password=monkey&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
@@ -163,14 +157,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("client_id=client_id&client_secret=client_secret&grant_type=client_credentials&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http", auth_scheme: OAuth2::AuthScheme::RequestBody
 
           token = client.get_access_token_using_client_credentials(scope: "read_posts")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("client_id=client_id&client_secret=client_secret&grant_type=client_credentials&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
@@ -182,14 +175,13 @@ describe OAuth2::Client do
           context.response.print response.to_json
         end
 
-        expected = %("client_id=client_id&client_secret=client_secret&grant_type=refresh_token&refresh_token=some_refresh_token&scope=read_posts")
         address = server.bind_unused_port "::1"
 
         run_server(server) do
           client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http", auth_scheme: OAuth2::AuthScheme::RequestBody
 
           token = client.get_access_token_using_refresh_token(scope: "read_posts", refresh_token: "some_refresh_token")
-          token.extra.not_nil!["body"].should eq expected
+          token.extra.not_nil!["body"].should eq %("client_id=client_id&client_secret=client_secret&grant_type=refresh_token&refresh_token=some_refresh_token&scope=read_posts")
           token.access_token.should eq "access_token"
         end
       end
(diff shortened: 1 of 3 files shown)


src/oauth2/auth_scheme.cr
```diff
@@ -0,0 +1,16 @@
...
+#  two parameters is NOT RECOMMENDED and SHOULD be limited to
+#  clients unable to directly utilize the HTTP Basic authentication
+#   scheme (or other password-based HTTP authentication schemes)."
+#
+# Therefore, HTTP Basic is preferred, and Request Body should only
+# be used if the server does not support HTTP Basic.
+#  
+enum OAuth2::AuthScheme
+  HTTP_Basic
+  Request_Body
```
jhass, line 15: Let's remove the underscores from the names here :)

See #8984

src/oauth2/client.cr
```diff
@@ -145,17 +152,25 @@ class OAuth2::Client
...
       "Accept"       => "application/json",
       "Content-Type" => "application/x-www-form-urlencoded",
     }
 
+    body = HTTP::Params.build do |form|
+      case @auth_scheme
+      when .http_basic?
+        headers.add(
+          "Authorization",
+          "Basic #{Base64.strict_encode "#{@client_id}:#{@client_secret}"}"
```
jhass, line 165: What do you think of making use of`Http::Client#basic_auth` here? I could see that it makes the rest of the code too verbose, especially the then more complex handling of `token_uri`

src/oauth2/client.cr
```diff
@@ -60,12 +60,19 @@ class OAuth2::Client
...
+  #  Request Body should only be used if the server does not support
+  #  HTTP Basic.
   def initialize(@host : String, @client_id : String, @client_secret : String,
                  @port : Int32? = nil,
                  @scheme = "https",
                  @authorize_uri = "/oauth2/authorize",
                  @token_uri = "/oauth2/token",
-                 @redirect_uri : String? = nil)
+                 @redirect_uri : String? = nil,
+                 @auth_scheme = AuthScheme::HTTPBasic)
```
asterite, line 75: Can we put a type restriction here?

src/oauth2/auth_scheme.cr
crush-157: OK - the style guide https://crystal-lang.org/reference/conventions/coding_style.html might need to be updated to reflect this.

src/oauth2/client.cr
crush-157: Done

src/oauth2/client.cr
crush-157: Done

src/oauth2/client.cr
```diff
@@ -54,18 +54,35 @@
...
+  private getter http_client : HTTP::Client do
+    HTTP::Client.new(URI.new(@scheme, @host, @port)).tap do |client|
+      client.basic_auth(client_id, client_secret) if auth_scheme.http_basic?
+    end
+  end
+
   # Creates an OAuth client.
   #
   # Any or all of the customizable URIs *authorize_uri* and
   # *token_uri* can be relative or absolute.
```
jhass, line 70: So this was what I was talking about in https://github.com/crystal-lang/crystal/pull/9127#discussion_r410750791 with "complex handling".

This implementation will break if you pass an absolute URL as `token_uri` now. I don't even know what will happen if you pass an absolute URL, possibly for a different host, to an instance of `HTTP::Client`, but I can't imagine it to just work.

Now the question is, do we want to keep supporting this or do we say the token endpoint being on another host is just too uncommon? I would be fine with either.

In case we keep it like this, the docs here need to be updated and this needs to be documented in the changelog as a breaking change, even more so. In case we go back I'd still be okay with the reimplementation of `HTTP::Client#basic_auth` in here.

What do others think?

crush-157: Actually I tested that it worked OK with an absolute URL as token URI:

```
require "oauth2"
require "base64"

host = ENV["HOST"]
client_id = ENV["CLIENT_ID"]
client_secret = ENV["CLIENT_SECRET"]
token_uri_suffix = ENV["TOKEN_URI_SUFFIX"]

token_uri = "https://#{host}#{token_uri_suffix}"

oauth2_client = OAuth2::Client.new(
  host: host,
  client_id: client_id,
  client_secret: client_secret,
  token_uri: token_uri
)

begin
  access_token = oauth2_client.get_access_token_using_client_credentials
  puts access_token.access_token
  puts access_token.extra
rescue ex : OAuth2::Error
  puts ex.class
  puts ex.message
end
```

```
eslater:oauth-test/ $ /Users/eslater/code/crystal/lang/crystal/bin/crystal oauth2-test.cr
J10i_m06f_wOiMS5QoTupA
```

crush-157: I'm happy to revert to the prior approach of just setting an auth header and using a one shot `HTTP::Client` if you prefer.

jhass: Mmh, did you test what happens if the host in the token URI is different to the one given to `OAuth2::Client`?

crush-157: No I didn't, because I don't have an environment like that available.

jhass: Maybe it's possible to write a spec for this somehow 🤔

src/oauth2/auth_scheme.cr
Sija: @crush-157 Why's that?

spec/std/oauth2/client_spec.cr
```diff
@@ -40,79 +40,158 @@ describe OAuth2::Client do
...
+        end
+
+        expected = %("redirect_uri=&grant_type=authorization_code&code=SDFhw39fwfg23flSfpawbef")
+        address = server.bind_unused_port "::1"
+
+        run_server(server) do
+          client = OAuth2::Client.new "[::1]", "client_id", "client_secret", port: address.port, scheme: "http"
+
+          token = client.get_access_token_using_authorization_code(authorization_code: "SDFhw39fwfg23flSfpawbef")
+          token.extra.not_nil!["body"].should eq expected
```
Sija, line 58: IMO it would be better to inline it in case it fails (will show the whole expectation and not the just the variable name).

```suggestion
          token.extra.not_nil!["body"].should eq %("redirect_uri=&grant_type=authorization_code&code=SDFhw39fwfg23flSfpawbef")
```

ditto elsewhere

src/oauth2/auth_scheme.cr
```diff
@@ -0,0 +1,16 @@
+# Enum of supported mechanisms used to pass credentials to the server.
+#
+# According to https://tools.ietf.org/html/rfc6749#section-2.3.1:
+#
+# "Including the client credentials in the request-body using the
+#  two parameters is NOT RECOMMENDED and SHOULD be limited to
+#  clients unable to directly utilize the HTTP Basic authentication
+#   scheme (or other password-based HTTP authentication schemes)."
```
Sija, line 8:
```suggestion
# > "Including the client credentials in the request-body using the
# > two parameters is NOT RECOMMENDED and SHOULD be limited to
# > clients unable to directly utilize the HTTP Basic authentication
# > scheme (or other password-based HTTP authentication schemes)."
```

src/oauth2/auth_scheme.cr
```diff
@@ -0,0 +1,16 @@
...
+#
+# Therefore, HTTP Basic is preferred, and Request Body should only
+# be used if the server does not support HTTP Basic.
+#
```
Sija, line 12:
```suggestion
```

src/oauth2/client.cr
```diff
@@ -54,18 +54,28 @@
...
   # If they are relative, the given *host*, *port* and *scheme* will be used.
   # If they are absolute, the absolute URL will be used.
+  #
+  # As per https://tools.ietf.org/html/rfc6749#section-2.3.1,
+  # HTTP Basic is the default `auth_scheme` (the mechanism used to
+  #  transmit the client credentials to the server).
+  #  Request Body should only be used if the server does not support
+  #  HTTP Basic.
```
Sija, line 71:
```suggestion
  # `AuthScheme::HTTPBasic` is the default *auth_scheme* (the mechanism used to
  # transmit the client credentials to the server). `AuthScheme::RequestBody` should
  # only be used if the server does not support HTTP Basic.
```

src/oauth2/client.cr
```diff
@@ -54,18 +54,28 @@
...
   def initialize(@host : String, @client_id : String, @client_secret : String,
                  @port : Int32? = nil,
                  @scheme = "https",
                  @authorize_uri = "/oauth2/authorize",
                  @token_uri = "/oauth2/token",
-                 @redirect_uri : String? = nil)
+                 @redirect_uri : String? = nil,
+                 @auth_scheme : AuthScheme = AuthScheme::HTTPBasic)
```
Sija, line 78:
```suggestion
                 @auth_scheme : AuthScheme = :http_basic)
```

src/oauth2/client.cr
```diff
@@ -145,18 +155,26 @@ class OAuth2::Client
...
-    response = HTTP::Client.post(token_uri, form: body, headers: headers)
+    body = HTTP::Params.build do |form|
+      case auth_scheme
+      when .request_body?
+        form.add("client_id", client_id)
+        form.add("client_secret", client_secret)
+      when .http_basic?
+        headers.add(
+          "Authorization",
+          "Basic #{Base64.strict_encode "#{@client_id}:#{@client_secret}"}"
```
Sija, line 171:
```suggestion
          "Basic #{Base64.strict_encode("#{client_id}:#{client_secret}")}"
```

src/oauth2/auth_scheme.cr
crush-157: Because I didn't read it right!  Sorry, my bad.

src/oauth2/client.cr
```diff
@@ -54,18 +54,27 @@
 # You can also use an `OAuth2::Session` to automatically refresh expired
 # tokens before each request.
 class OAuth2::Client
+  private getter host, client_id, client_secret, port, scheme, authorize_uri,
+    redirect_uri, auth_scheme, token_uri
+
```
Sija, line 59: What's the point of this `private getter` instead of using instance variables?
Smells like YAGNI :)

src/oauth2/client.cr
```diff
@@ -165,15 +182,12 @@ class OAuth2::Client
     end
   end
 
-  private def token_uri
-    uri = URI.parse(@token_uri)
-
+  private def get_token_uri : URI
```
Sija, line 185: What's the point of this rename?

src/oauth2/client.cr
crush-157: I always prefer to use getters/setters instead of accessing instance variables directly.

Sija: Well, refactoring like this might be considered out of scope of this PR. Secondly it's more costly, so if there's no good reason I'd advice against that.

crush-157: My reason for preferring getters/setters over instance variables even privately is that if the feature needs to change from being a variable to a method then you restrict your changes to the getter/setter.

Context: I once "won" an argument over this on a project where my view was that we *should* access private instance variables directly, and then lived to regret it.  Since then, I've always had a horror of accessing instance variables directly (outside of constructors/getters/setters).

crush-157: OTOH if you have a horror of private accessors I can change them back.  Your call.

src/oauth2/client.cr
crush-157: Good question ;-)

Will switch it back.

src/oauth2/auth_scheme.cr
crush-157: @Sija I slept on this and remembered why...

I think the style guide should be updated, because it doesn't state what the case should be for the *values* of an enum, just the enum name itself.

> enum Time::DayOfWeek
> end

And if you look at the examples in the GitBook section on enums, ( https://crystal-lang.org/reference/syntax_and_semantics/enum.html ) the values are all single words.

So I think the style guide should explicitly state the conclusions of [#7270](https://github.com/crystal-lang/crystal/issues/7270).

src/oauth2/client.cr
Sija: I'd suggest doing so, yes :)

jhass: @Sija fwiw https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/codegen/call.cr#L430-L437

Sija: @jhass I'd argue that in this case it's simply out of scope of this PR, but thanks for sharing this anyway ❤️

crush-157: @jhass Good to know

crush-157: @Sija private accessors dropped for using instance variables directly.  I won't tell Bertrand Meyer if you don't.

spec/std/oauth2/client_spec.cr
```diff
@@ -40,79 +40,150 @@ describe OAuth2::Client do
...
 
-        token = client.get_access_token_using_resource_owner_credentials(username: "user123", password: "monkey", scope: "read_posts")
-        token.extra.not_nil!["body"].should eq expected
-        token.access_token.should eq "access_token"
+          token = client.get_access_token_using_refresh_token(scope: "read_posts", refresh_token: "some_refresh_token")
+          token.extra.not_nil!["body"].should eq %("grant_type=refresh_token&refresh_token=some_refresh_token&scope=read_posts")
+          token.access_token.should eq "access_token"
+        end
       end
     end
```
asterite, line 115: Is a newline missing after this line?

PR #9130, opened 2020-04-19
Update docs for and slightly improve readability of HTTP::WebSocketHandler

dscottboggs:


src/http/server/handlers/websocket_handler.cr
```diff
@@ -9,45 +9,54 @@ require "http/web_socket"
   require "openssl/sha1"
 {% end %}
 
+# A handler which adds websocket functionality to an `HTTP::Server`.
+#
+# When a request can be upgraded, the associated `HTTP::Websocket` and
+# `HTTP::Server::Context` will be yielded to the block. For example:
+#
+# ```crystal
```
bcardiff, line 17: there is no need to indicate the language for crystal code

src/http/server/handlers/websocket_handler.cr
```diff
@@ -9,45 +9,54 @@ require "http/web_socket"
...
-      response.upgrade do |io|
-        ws_session = WebSocket.new(io, sync_close: false)
-        @proc.call(ws_session, context)
-        ws_session.run
-      ensure
-        io.close
-      end
-    else
-      call_next(context)
+    return call_next context unless websocket_upgrade_request? context.request
```
asterite, line 30: Can we put the condition before the return? It's done like that everywhere else later on. I find the line a bit long and it is less understandable than before.

dscottboggs: Sure, I just found the original pretty hard to read because the else block was so far away from the condition.

src/http/server/handlers/websocket_handler.cr
```diff
@@ -9,45 +9,57 @@ require "http/web_socket"
...
-        @proc.call(ws_session, context)
-        ws_session.run
-      ensure
-        io.close
-      end
-    else
-      call_next(context)
+    unless websocket_upgrade_request? context.request
+      call_next context
+      return
```
Sija, line 32:
```suggestion
      return call_next(context)
```

dscottboggs: I was following the pattern set out below by putting the return on a separate line

asterite: I think the return in the same line is fine. I was just taking about not inlining the unless.

Sija: This also keeps the previous behaviour.

dscottboggs: fixed, and changed the other to be consistent with this in that case

Sija: @dscottboggs actually, the other case should be left as-is.

PR #9133, opened 2020-04-19
Refactor preserving state out of OptionParser#parse

Sija:
Followup to #9009, as per https://github.com/crystal-lang/crystal/pull/9009#pullrequestreview-394978973


straight-shoota:
This method already has incredibly high complexity. I'd rather separate the implementation to a private method, so that the reset logic is closer together.

asterite:
Probably worth extracting the "save into old, restore from old" into a separate method that yields.

Sija:
Refactored as suggested.

PR #9134, opened 2020-04-19
Refactor to standardize on first argument for methods receiving `IO`

### LINKED ISSUE #5916: Standardize overloads for methods with IO/returning String
The idiomatic Crystal way of converting a value to a `String` is to append it to an `IO`. The most notable it `Object#to_s(io : IO)` which writes a string representation to the `IO`. When an actual `String` is required, there is an overload `Object#to_s()` which calls the IO variant wrapped in `String.build`.

Other examples are `Object#inspect`, `Base64#encode`, `Base64#decode` or `Enumerable#join` which all adapt the same technique.

Since they provide the same functionality, both methods usually have the same signature except that one returns `String` and the other accepts an additional argument `io : IO`. That's the distinctive difference between them. It is obvious that the positioning of the other arguments should be the same to avoid any friction.

If the methods also accept other arguments, there is a question as to where the `io` argument should be placed. There are different styles throughout the standard lib:

* `Time#to_s(format : String, io : IO)`
* `Int#to_s(base : Int, io : IO, upcase : Bool = false)`
* `BigRational#to_s(io : IO, base = 10)`

It doesn't make any sense to add `io` in between as it screws up argument order. At the end is better but it will cause issues when optional arguments are added to the signature (that's probably what happened with `Int#to_s`). I think it should be preferred to put `io` at the front of the other arguments. This clearly signals that this is the IO variant of the method. And additional arguments can simply be appended at the end in both variants. IMHO this is the most logical solution.

I propose to adapt "`io` in front" style throughout the standard library and encourage others to do the same. This way it would be simple to switch from a String method to the IO method by inserting `io` as first argument and you don't need to think about where to put it.

straight-shoota:
This introduces some breaking changes but rewards with great developer happyness because you don't need to recall the different placements of `io` arguments. It is now always the first argument.

There is a somewhat unrelated change in cf613ea which makes `upcase` argument to `Int#to_s` a named argument. But it makes sense to introduce that change at the same time as the argument order. It was already only used as a named argument in stdlib so this didn't even need any changes here and it underlines the fact that you would not pass it as positional argument.

Resolves #5916

### DIFF
--- a/spec/std/enumerable_spec.cr
+++ b/spec/std/enumerable_spec.cr
@@ -620,25 +620,35 @@ describe "Enumerable" do
     end
   end
 
-  describe "join" do
-    it "joins with separator and block" do
-      str = [1, 2, 3].join(", ") { |x| x + 1 }
-      str.should eq("2, 3, 4")
+  describe "#join" do
+    it "()" do
+      [1, 2, 3].join.should eq("123")
     end
 
-    it "joins without separator and block" do
+    it "(separator)" do
+      ["Ruby", "Crystal", "Python"].join(", ").should eq "Ruby, Crystal, Python"
+    end
+
+    it "(&)" do
       str = [1, 2, 3].join { |x| x + 1 }
       str.should eq("234")
     end
 
-    it "joins with io and block" do
+    it "(separator, &)" do
+      str = [1, 2, 3].join(", ") { |x| x + 1 }
+      str.should eq("2, 3, 4")
+    end
+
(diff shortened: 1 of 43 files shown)


straight-shoota:
Rebased on master to fix conflicts

oprypin:
Missed https://github.com/crystal-lang/crystal/blob/d49375b181b9803a5a405b9b04846dce6aec2870/src/crystal/system/win32/process.cr#L159

PR #2375, opened 2016-03-27
Added Process.daemonize, spec, and samples

DougEverly:
Detaches a process from the terminal to run in background.

STDIN, STDOUT, STDERR default to /dev/null. Can override with file paths.

`Process.daemonize(stdin: "/dev/null", stdout: "/dev/null", stderr: "/dev/null”)`

New PR to clean up commits from #2363.

PR #9136, opened 2020-04-19
Deprecate top-level fork

bcardiff:
`Process.fork` is still available, yet it will only work in single-thread.

There is no point in having a top-level API that will work in only one mode.

### DIFF
--- a/src/compiler/crystal/compiler.cr
+++ b/src/compiler/crystal/compiler.cr
@@ -457,7 +457,7 @@ module Crystal
               end
             end
 
-            codegen_process = fork do
+            codegen_process = Process.fork do
               pipe_w = pw
               slice.each do |unit|
                 unit.compile
--- a/src/crystal/system/unix/event_loop_libevent.cr
+++ b/src/crystal/system/unix/event_loop_libevent.cr
@@ -5,6 +5,7 @@ class Thread
   getter(event_base) { Crystal::Event::Base.new }
 end
 
+# :nodoc:
 module Crystal::EventLoop
   {% unless flag?(:preview_mt) %}
     # Reinitializes the event loop after a fork.
--- a/src/kernel.cr
+++ b/src/kernel.cr
@@ -491,6 +491,8 @@ end
 
 {% unless flag?(:preview_mt) %}
   class Process
+    # :nodoc:
+    #
     # Hooks are defined here due to load order problems.
     def self.after_fork_child_callbacks
       @@after_fork_child_callbacks ||= [
--- a/src/process.cr
+++ b/src/process.cr
@@ -69,6 +69,8 @@ class Process
     Crystal::System::Process.times
   end
 
+  # :nodoc:
+  #
   # Runs the given block inside a new process and
   # returns a `Process` representing the new child process.
   #
@@ -90,6 +92,8 @@ class Process
     end
   end
 
+  # :nodoc:
+  #
   # Duplicates the current process.
   # Returns a `Process` representing the new child process in the current process
   # and `nil` inside the new child process.
@@ -439,22 +443,6 @@ def `(command) : String
   status = process.wait
   $? = status
   output
-end
-
-# See also: `Process.fork`
-#
-# Available only on Unix-like operating systems.
-@[Deprecated("Use `Process.fork` instead")]
-def fork
-  Process.fork { yield }
-end
-
-# See also: `Process.fork`
-#
-# Available only on Unix-like operating systems.
-@[Deprecated("Use `Process.fork` instead")]
-def fork
-  Process.fork
 end
 
 require "./process/*"
--- a/src/process.cr
+++ b/src/process.cr
@@ -444,6 +444,7 @@ end
 # See also: `Process.fork`
 #
 # Available only on Unix-like operating systems.
+@[Deprecated("Use `Process.fork` instead")]
 def fork
   Process.fork { yield }
 end
@@ -451,6 +452,7 @@ end
 # See also: `Process.fork`
 #
 # Available only on Unix-like operating systems.
+@[Deprecated("Use `Process.fork` instead")]
 def fork
   Process.fork
 end


straight-shoota:
Do we not want to remove `Process.fork` entirely per #6421?

It wouldn't make sense to suggest using `Process.fork` if that get's removed next.

bcardiff:
I'm fine with leaving `Process.fork` as `:nodoc:`, and directly removing top-level `fork`.

oprypin:
Umm I think it's a bit of a problem to deprecate a method that the compiler fully relies on, in the default code path

https://github.com/crystal-lang/crystal/blob/d06603220d3d1a55c49a203c25c46ae817fd1e49/src/compiler/crystal/compiler.cr#L460

oprypin:
Oh I see you're replacing it with the "alias". But what's the point really?

straight-shoota:
The point is that we shouldn't provide a public API for `fork` as per #6421.
It's just an implementation detail for executing a process on posix.

bcardiff:
@straight-shoota are we ok with this as is, or making `Process.fork` :nodoc: & dropping instead of deprecating the top-level method is better on your opinion?

asterite:
Can you still call `Process.fork` in MT?

Maybe we should just remove those definitions in MT?

bcardiff:
If you use Process.fork it fails on compile-time with the message "Process fork is unsupported with multithread mode"

straight-shoota:
I don't think `Process.fork` should be part of the public API. Whether it's still available as `:nodoc:` or completely removed (and only used internally in `Crystal` namespace) I don't really care. But removing might be better.

refi64:
Could there be e.g. an unsafe fork still that raises if >1 thread is
active? Golang's fork being "hidden" of sorts means that if you *really*
need it (fork is the only way to get data from another namespace on Linux
without spawning a helper process), you need to go through an ugly and
elaborate dance, even if you can guarantee there are no other thread a
running or that you would interact with them at any point in the short
forked process's lifetime.

RX14:
I'd rather have a `Crystal::System::Process.fork` available for the few specialized usecases (@refi64 mentions a few) that it still has on linux. Those programs which use it are expected to have done the carefulwork to avoid segfaults and be aware that it's *very* unsafe and subject to change.

This PR is a good step towards removing `Process.fork`. We can do the rest later.

RX14:
Actually, we should probably provide a `@[Deprecated]` notice for `Process.fork` so that applications are aware of this.

oprypin:
Um that would mean that compiling the compiler would show a deprecation warning.

jhass:
Maybe we need a `@[SuppressWarnings("deprecatation")]` annotation then? 😆

straight-shoota:
@oprypin The method should just be named `Crystal::System::Process.fork` and the compiler use that.

asterite:
> we should probably provide a @[Deprecated] notice for Process.fork so that applications are aware of this

did you mean to `fork`?

oprypin:
No, fork is slated to be directly deleted *now* according to discussions here

straight-shoota:
I think we should deprecate both `::fork` and `Process.fork` now without replacement. Implementation is moved to `Crystal::System::Process.fork` and that should be used by the compiler.

RX14:
Why don't we just compile the compiler with preview_mt and use threads?

straight-shoota:
We can probably do that, too. But it's a separate change from this.

bcardiff:
There are a couple of approvals and later discussions. I get that this fine to be merged.

The `Process.fork` is `:nodoc:`, so if later we add `Crystal::System::Process.fork` we can still deprecate or remove that since it's not public api.

straight-shoota:
I didn't realise it was nodoc, so this is totally fine.

will:
Will there still be a way to create crystal programs that can self-daemonize in the future? The technique I've been using uses fork. I just noticed that the `preview_mt` flag throws an exception, and I'm worried that if fork is removed entirely, there will no longer be a way to self-daemonize. For my particular case, I don’t need the mt enhancements so it hasn't been a problem yet.

( I just recently [asked this on an issue](https://github.com/crystal-lang/crystal/pull/2375#issuecomment-686660794) that had daemonize in the title, since that was what I was searching for, but this is a better place to ask I think )

jhass:
I'm with 6 years younger Mike here https://www.mikeperham.com/2014/09/22/dont-daemonize-your-daemons/

will:
That post addresses "long-lived daemon process", and I agree with it in that situation. My use case however is _short-lived_ programs that I trigger on a remote server over ssh, where I don't want to hold the ssh connection open while the program runs.

jhass:
`man nohup` :)

yxhuvud:
Another reason to remove fork altogether is that it prevents incremental garbage collection in boehm GC. 

I'd much rather have the ability to set up custom pthread based threadpools, perhaps similar to how I do it in https://github.com/yxhuvud/nested_scheduler#threads

zw963:
Thank god, the fork wasn't really removed (just nodoc) only because it not work on windows or MT, instead, have to use LibC to do some, that can be a very bad experience.

the users use fork will care about the opened files, sockets etc, and used in single-threaded scenarios.

as far as I know,  until now, Nov 7, 2022, most of Crystal user still use POSIX compatible systems, and only few shards MT enabled.

I consider we should remove the `:nodoc:` for `Process.fork`,  and explain not supported on MT clearly and raise a error on non-posix is enough.

straight-shoota:
@zw963 `Process.fork` is `:nodoc:` for a reason. It's not advisable to use that method as part of the public stdlib API because it's not. It is an implementation detail and may disappear without notice.

If you want to use `fork` in a Crystal program despite its issues (cf. #6421) I recommend to use a separate implementation instead of depending on `Process.fork`. You can of course use the stdlib implementation as a source for that. But it's better to duplicate and internalize the implementation than depending on an internal API.

Ideally, such non-portable APIs could be collected in a platform-specific shard for easier reusability.

zw963:
platform-specific feature available in a general purpose language is not rare.

Users use Crystal in Linux want to start a background daemon, have to seeking a shards for this, this is not consider a good experiences.

do we have precedent for other common used general purpose languages? target linux as a tie 1 support platfrom?

if you say golang, it maintain a minimal kernel on purpose, it totally not the same style of Crystal.

If I'm not mistaken, people who like Crystal don't like golang, at this point, Crystal does not have to do the same as golang.

asterite:
> as far as I know, until now, Nov 7, 2022, most of Crystal user still use POSIX compatible systems

Well of course! Windows support isn't there yet, so of course the only system that people use is POSIX.

> do we have precedent for other common used general purpose languages?

- Go doesn't have fork
- Rust doesn't have fork
- Java doesn't have no fork
- C# doesn't have fork
- Dart doesn't have fork
- Kotlin doesn't have fork

> If I'm not mistaken, people who like Crystal don't like golang

I created Crystal, I like Go, and I hope people using Crystal like other languages too. There's no reason not to like a language. They are all efforts to create something useful for others.

yxhuvud:
> platform-specific feature available in a general purpose language is not rare.

I am among the users that think that Crystal should provide bindings to operating system specific features and make use of that to provide the best possible experience possible for each and every platform, even if it is means some programs using only the standard distribution would be platform specific. In some cases I'd like to have a OS specific namespace that provide certain features (*) and in some cases things should just silently work better. I also think asyncronicity of methods should be on a best effort basis - things that can be made asynchronous and nonblocking on one OS should be made async but doesn't need to be async everywhere.  

I'm probably more open to things like that than most of the core members, from what I've seen in various discussions.

But `fork` is still a big no-no from my side. Why? Because having support for it directly make developing the language runtime harder. The event-loop, fibers, locking, scheduling, GC. The impact is all over the place and always make things either impossible to implement in a nicer way or at least much harder. `fork` may need to be there now as everything isn't in place, but I certainly hope we don't ever promise it will stick around. 

> Users use Crystal in Linux want to start a background daemon

This does not in any way require fork. There may be other use cases where fork currently is the least bad solution, but background daemons are not a case of that.

(*) currently existing candidates I know of: `wait_readable` and `wait_writable` are kinda essential to be a good linux citizen for interacting with other libraries and quite nice to have on mac as well. But they are more or less impossible to implement on windows.

zw963:
> This does not in any way require fork.

Maybe i missing something? if i understand correct, what you mean is, fork is not necessary for start a background daemon, can i know how to do `Process.daemon` in ruby for crystal? we know `exec` is a way to transform the current process into another one, how to do the `fork + exec` in Crystal?

zw963:
> > do we have precedent for other common used general purpose languages?
> 
>     * Go doesn't have fork
> 
>     * Rust doesn't have fork
> 
>     * Java doesn't have no fork
> 
>     * C# doesn't have fork
> 
>     * Dart doesn't have fork
> 
>     * Kotlin doesn't have fork
> 

Wow, never realized that there are so many languages that don't have forks, as replied in previous comment,  i don't know how to start a daemon if no fork without any external tools support.

> I created Crystal, I like Go, and I hope people using Crystal like other languages too. 

I'm so surprised for this.

straight-shoota:
Let's move the discussion about daemonizing a program to the forum, please.

zw963:
Created new thread in the forum.

https://forum.crystal-lang.org/t/deprecate-top-level-fork/5130

Let us discuss there.

PR #9137, opened 2020-04-20
Add Path#empty?

straight-shoota:
This adds a new query method to `Path` to tell whether it's empty.
The existing definition of empty is expanded to also include `"."` which hasn't been covered before but is actually the normalized form of `""`.


Sija:
I'm not sure I understand the definition of `Path#empty?` - what does it mean? And why `.` is considered as empty?

straight-shoota:
Empty means  `""` or `"."` which usually represents the current directory.

Sija:
> And why `.` is considered as empty?

@straight-shoota ☝️

straight-shoota:
Because it's semantically equivalent to `""` and actually its normalized form.

Sija:
> Because it's semantically equivalent to `""` and actually its normalized form.

I'd rather say the opposite, `""` is semantically equivalent to `"."`. Still, I can't find one reason to call either of those _empty_ - apart from being an empty string, which in this case is an implementation detail.

Also, note the Ruby's `Pathname#empty?`:

> Tests the file is empty.

straight-shoota:
Do you have a suggestion for a better name? IMO empty is pretty much self explanatory and the typical term for this.

Sija:
> Do you have a suggestion for a better name?

I'm not sure we need this method at all, since you just came up with that with no good usecase, there was no one requesting, or even mentioning this before, so...

> IMO empty is pretty much self explanatory and the typical term for this.

Well, for sure it ain't for me, hence my question. I don't think I've seen such method anywhere else.

straight-shoota:
The description has always been there in the doc comment for `Path` (it just lacked the normalized form), this just adds a method to programmatically query this property.
Immediate use cases are already provided in the diff. Apart from internal reusability I figure this method could also be useful for custom algorithms working with `Path` which can often implement special behaviour when an path is empty.

asterite:
Agreed, "." doesn't look like empty to me.

straight-shoota:
> I think that this is picking Path#empty? becuase String#empty? and @name : String.

No, it's not. I actually copied that name from boost.

But fine, I'll make the method private.

PR #9138, opened 2020-04-20
Add Path#each_part iterator

straight-shoota:
This PR adds a new overload of `Path#each_part` that returns an iterator.


oprypin:
Nice demo for why the compiler should support generating such iterators automatically, but also why it's so difficult.

straight-shoota:
@crystal-lang/crystallers Would be great to get reviews on this because I have a follow-up waiting =)

PR #9140, opened 2020-04-20
Object: fix lazy property not forwarding annotations

### LINKED ISSUE #9139: Annotation Not Set on Property Definition
Code like the following does not add the `Perishable` annotation to the `@test` instance variable:
```crystal
class Test
  annotation Perisable
  end

  @[Perisable]
  property test : String? { "test" }

  def initialize
    {% puts @type.instance_vars.map(&.name) %}
    {% puts @type.instance_vars.select(&.annotation(Perisable)).map(&.name) %}
  end
end

Test.new.inspect
```

Running this example produces:
```
$ crystal run /tmp/test.cr
[test]
[]
```

Expanding the `property` macro shows that when `property` is used with a block, the `setter` macro is used to generate the setter, which results in a secondary definition of `@test`. The secondary definition does not get the annotation:
```
$ crystal tool expand /tmp/test.cr -c /tmp/test.cr:6:3
[test]
[]
1 expansion found
expansion 1:
   property(test : String | ::Nil) do
     "test"
   end

# expand macro 'property' (/usr/local/Cellar/crystal/0.34.0/src/object.cr:277:3)
~> setter(test : String | ::Nil)
   @test : String | ::Nil | ::Nil
   def test : String | ::Nil | ::Nil
     if (  value = @test).nil?
       @test = ("test")
     else
       value
     end
   end

# expand macro 'setter' (/usr/local/Cellar/crystal/0.34.0/src/object.cr:277:3)
~> @test : String | ::Nil
   def test=(test : String | ::Nil)
     @test = test
   end
   @test : String | ::Nil | ::Nil
   def test : String | ::Nil | ::Nil
     if value = @test.nil?
       @test = "test"
     else
       value
     end
   end
```

You can see the code [here](https://github.com/crystal-lang/crystal/blob/master/src/object.cr#L911).

The `property?` macro [handles this correctly](https://github.com/crystal-lang/crystal/blob/master/src/object.cr#L1153), and generates its own setter. Likewise, simple assignment does not have this problem, nor does attaching the annotation to the instance variable directly.

```
$ crystal -v
Crystal 0.34.0 (2020-04-07)

LLVM: 10.0.0
```

asterite:
Fixes #9139

### DIFF
--- a/spec/std/object_spec.cr
+++ b/spec/std/object_spec.cr
@@ -112,6 +112,16 @@ private class TestObject
   def []=(key, value)
     {key, value}
   end
+
+  annotation TestAnnotation
+  end
+
+  @[TestAnnotation]
+  property(x : Int32) { 1 }
+
+  def self.test_annotation_count
+    {{ @type.instance_vars.select(&.annotation(TestObject::TestAnnotation)).size }}
+  end
 end
 
 private class DelegatedTestObject
@@ -481,5 +491,9 @@ describe Object do
     it "shouldn't return same hash for different property values" do
       HashedTestObject.new(1, 2).hash.should_not eq HashedTestObject.new(3, 4).hash
     end
+  end
+
+  it "applies annotation to lazy property (#9139)" do
+    TestObject.test_annotation_count.should eq(1)
   end
 end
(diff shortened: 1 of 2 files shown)


toddsundsted:
thank you!

bcardiff:
Sounds good. For a more complete approach for the future I like @jhass proposal at https://github.com/crystal-lang/crystal/issues/9139#issuecomment-616517082 .

PR #9144, opened 2020-04-20
Win CI: Enable more specs, add zlib and yaml

oprypin:


src/crystal/system/win32/file.cr
```diff
@@ -26,8 +26,8 @@ module Crystal::System::File
     fd
   end
 
-  def self.mktemp(prefix : String, suffix : String?, dir : String) : {LibC::Int, String}
-    path = "#{tempdir}\\#{prefix}.#{::Random::Secure.hex}#{suffix}"
+  def self.mktemp(prefix : String?, suffix : String?, dir : String) : {LibC::Int, String}
+    path = "#{Crystal::System::Dir.tempdir}\\#{prefix}.#{::Random::Secure.hex}#{suffix}"
```
Sija, line 30: Shouldn't this be similar to the unix impl? 👇 https://github.com/crystal-lang/crystal/blob/625884b1acbbbf4db321ac76738ebef35a0aba93/src/crystal/system/unix/file.cr#L16-L18

```suggestion
  def self.mktemp(prefix, suffix, dir) : {LibC::Int, String}
    path = "#{dir}\\#{prefix}.#{::Random::Secure.hex}#{suffix}"
```

straight-shoota: The method signature doesn't matter much, but yes, the `dir` argument should be honoured.

spec/win32_std_spec.cr
```diff
@@ -1,64 +1,75 @@
...
 # require "./std/big/big_float_spec.cr" (failed linking)
 # require "./std/big/big_int_spec.cr" (failed linking)
 # require "./std/big/big_rational_spec.cr" (failed linking)
-# require "./std/big/number_spec.cr" (failed codegen)
+# require "./std/big/number_spec.cr" (failed linking)
 require "./std/bit_array_spec.cr"
 require "./std/bool_spec.cr"
 require "./std/box_spec.cr"
-# require "./std/callstack_spec.cr" (failed codegen)
+# require "./std/callstack_spec.cr"
```
RX14, line 17: What happened here?

oprypin: Yes, it is neither of those. It no longer fails codegen and it doesn't crash, but it just has a lot of test failures that I don't want to deal with as part of this PR. So I commented it out.

oprypin:
Why does it dismiss reviews now? There was no force-push 😩

straight-shoota:
By adding the merge commit the PR head changed.

oprypin:
I can get that, but it should be allowed still

bcardiff:
We changed that repo setting to avoid having superfluous/old approvals @oprypin .

spec/win32_std_spec.cr
```diff
@@ -1,64 +1,75 @@
...
 require "./std/comparable_spec.cr"
 require "./std/complex_spec.cr"
+require "./std/compress/deflate/deflate_spec.cr"
+require "./std/compress/gzip/gzip_spec.cr"
+require "./std/compress/zip/zip_file_spec.cr"
+require "./std/compress/zip/zip_spec.cr"
+require "./std/compress/zlib/reader_spec.cr"
+require "./std/compress/zlib/stress_spec.cr"
+require "./std/compress/zlib/writer_spec.cr"
+# require "./std/concurrent/future_spec.cr" (failed codegen)
```
bcardiff, line 32: FYI this file no longer exists on master.

spec/win32_std_spec.cr
bcardiff: This file has moved to call_stack_spec on master, which is already included later.

src/crystal/system/win32/file.cr
```diff
@@ -26,8 +26,9 @@ module Crystal::System::File
     fd
   end
 
-  def self.mktemp(prefix : String, suffix : String?, dir : String) : {LibC::Int, String}
-    path = "#{tempdir}\\#{prefix}.#{::Random::Secure.hex}#{suffix}"
+  def self.mktemp(prefix : String?, suffix : String?, dir : String) : {LibC::Int, String}
```
bcardiff, line 29: It's odd that these type restrictions are not present in the unix counter part

spec/win32_std_spec.cr
oprypin: Thanks. I double-checked by re-generating the list, and your two comments cover everything.

src/crystal/system/win32/file.cr
oprypin: I suppose, yes. I'm not changing the status quo, though.

oprypin:
It dismisses reviews on *any* commit??? That's just unacceptable. Please undo that setting and let the person who merges judge what's stale or not. It's hard enough to get 2 people to approve as is.
In this particular case, if @bcardiff approved and I made only a tiny change exactly in accordance to his comment, obviously that's not stale.

straight-shoota:
@oprypin Let's focus on the PR here. Meta discussion about GitHub settings should be separate.

oprypin:
Rest assured that the thought crossed my mind, but I don't know what the place would be.
And feel free to collapse the comments.

src/crystal/system/win32/file.cr
```diff
@@ -26,8 +26,9 @@ module Crystal::System::File
...
+    dir = dir + ::File::SEPARATOR
+    path = "#{dir}#{prefix}.#{::Random::Secure.hex}#{suffix}"
```
straight-shoota, line 31: This avoids an extra allocation:
```suggestion
    path = "#{dir}#{::File::SEPARATOR}#{prefix}.#{::Random::Secure.hex}#{suffix}"
```
Or use `::File.join` I think we can depend on that from here.

oprypin: This is not by accident, it's just copying the implementation from unix. I'd prefer to not refactor that here (pretend that I'm "not changing the code"); maybe someone can do both at the same time.

```
$ grep -rF1 'dir = dir + ::File::SEPARATOR'
src/crystal/system/unix/file.cr-  def self.mktemp(prefix, suffix, dir) : {LibC::Int, String}
src/crystal/system/unix/file.cr:    dir = dir + ::File::SEPARATOR
src/crystal/system/unix/file.cr-    path = "#{dir}#{prefix}.XXXXXX#{suffix}"
--
src/crystal/system/win32/file.cr-  def self.mktemp(prefix : String?, suffix : String?, dir : String) : {LibC::Int, String}
src/crystal/system/win32/file.cr:    dir = dir + ::File::SEPARATOR
src/crystal/system/win32/file.cr-    path = "#{dir}#{prefix}.#{::Random::Secure.hex}#{suffix}"
```

oprypin: Ah whatever, applied it. Thanks. Leaving it wouldn't make as much sense as I make it seem.

PR #9145, opened 2020-04-20
Log: rework setup_from_env and defaults

### LINKED ISSUE #9041: Class/module local Log should emit logs by default
According to the docs I should be able to do something like this with the new Log module:

```crystal
require "log"

Log.setup_from_env

class Foo
  Log = ::Log.for(self)

  def bar
    Log.info { "Starting" }
    puts "bar"
    Log.info { "Finished" }
  end
end

foo = Foo.new
foo.bar
```

However if you run that code you'll find that the only thing that gets printed is "bar". The same thing occurs even if you do `Log = ::Log.for(self, ::Log::Severity::Info)`.

Am I doing something wrong?

bcardiff:
Fixes #9041

* The `Log.setup_from_env` call is now included by default
* The default source level is `*`

These two things will make all logs of info and above to show in stdout.

* The `setup_from_env` named arguments were changed to allow simpler overrides of defaults (for when env variables are not set).

Since the builders can be reconfigured the effect of calling `Log.setup_from_env` by default is that some logs might appear in the STDOUT if they are emitted before the framework / application reconfigure it as pleased. Either by calling `setup_from_env` again or manually defining the bindings.


Blacksmoke16:
Do we not want to leave calling `Log.setup_from_env` to the end application?  The second bullet point and added documentation should address #9041.

bcardiff:
Enforcing a default logging setup allows the following small programs to work out of the box.

```crystal
require "log"

Log.warn { "hello world" }
```

Since the `Log.builder` can be reconfigured entirely, as long as there are no entries emitted before the last configuration there is no real difference.

So, I think there are no actual downsides on this approach.

src/log.cr
```diff
@@ -85,38 +85,41 @@
 #
 # ```
 # backend = Log::IOBackend.new
+# Log.builder.clear
```
wontruefree, line 88: Do you have to run `clear` before `bind`?

src/log.cr
```diff
@@ -85,38 +85,41 @@
...
 #
 # ### Configure logging from environment variables
 #
-# Include the following line to allow configuration from environment variables.
-#
-# ```
-# Log.setup_from_env
-# ```
-#
-# The environment variables `CRYSTAL_LOG_LEVEL` and `CRYSTAL_LOG_SOURCES` are used to indicate
```
paulcsmith, line 101: I'm not sure that this is an improvement. I think manually calling `Log.setup_from_env` makes sense because people may not want/expect it to be pulling something from the env. 

Is there a way to change the default log level without forcing people into `Log.setup_from_env`?

Or maybe I misunderstood and you still need to call `Log.setup_from_env`?

src/log.cr
bcardiff: After this PR, if you want to remove the default binds added by setup_from_env, yes you would need that.

src/log.cr
bcardiff: If you want to change the default log level, you need to call again `setup_from_env`.

Using `Log.setup_from_env` enables out of the box minimal configuration via ENV var.

The app or framework is able to redefine the whole bindings and default level despite a built-in configuration. Even leaving out the flexibility of env vars.

Blacksmoke16: I guess my main concern is what percentage of people are going to be okay with these defaults.  It's probably fine for a development environment, but I have a feeling most people are going to just end up overriding the defaults anyway to setup things how they want; at which point were the defaults even worth it?  Given that `Log.setup_from_env` and `Log.builder.clear` are both one line, I'm still in favor of using the new defaults/config of `Log.setup_from_env` but requiring the user to call it if they want to use those defaults.

src/log.cr
paulcsmith: This seems like a big change and may introduce some weird regressions. If I understand this correctly, if I did this:

```crystal
Log.builder.bind("*", :info, MyCoolbackend.new)
```

The logs would now have 2 backends because it has the default backend + the MyCoolBackend? If so I think this is going to be very confusing for people. I think that setup_from_env should be called explicitly and instead the default should not be "set" but should be something like `Log#severity || :info`. So it does not really on bind at all for it to work and doesn ot require "clearing" the defaults

carlhoerberg: i think it's very important that things gets logged by default. But what about if the "default" configuration gets cleared on the first "manual" bind?

paulcsmith: I agree a default is great. I think there could be a default that doesn’t have to be “set” from  along seup_from_env though. It could work like most defaults where if nothing is explicitly set it uses a default value

src/log.cr
asterite: I don't know, I find it pretty strange that if I write this:

```crystal
require "log"

Log.error { "hello world!" }
```

and run it, I get nothing as output. I would expect logging to go to STDOUT by default, without having to setup the logging. That should be the default.

Unrelated, but this example from the docs looks pretty strange:

```crystal
backend = Log::IOBackend.new
Log.builder.bind "*", :warning, backend
```

IOBackend? Which IO? Oh, I guess it uses STDOUT as a default? Not clear at all from that code. I think we should make the IOBackend constructor parameter required. Then you need to do `Log::IOBackend.new(STDOUT)`, which is longer but clearer in my opinion.

waj: We could actually have a subclass named `TTYBackend` or similar, that logs to `STDOUT` by default and uses coloring when the output is a tty

straight-shoota: @asterite Logging by default should be fine, I think everyone agrees on that. But `setup_from_env` does not only enable logging by default but also exposes configuration via environment variables and initializes the log builder so that you need to take that into account when configuring custom log setup.

I think the point is that for a simple program like your example logging should just work. But when you want to set up a custom logging configuration, you should not have to call `Log.builder.clear`. If you haven't setup anything explicitly that shouldn't be required.
This could be implemented using a flag which is set after initializing the logger and when a new log rule gets added and the flag is set, it clear the default rules. Afterwards the flag is disabled.

straight-shoota: @waj Coloring is fine but I don't think we need a another class for this.

asterite: > but also exposes configuration via environment variables

In most Ruby libraries I've used, they always try to get a default config value from environment first. That's so nice because then you don't have to manually configure it. You know that whenever you need to use env vars, you can, and they have a precedence over everything else.

> This could be implemented using a flag which is set after initializing the logger and when a new log rule gets added and the flag is set, it clear the default rules.

I agree. Configuring the logger after the default has been set should reset it first.

bcardiff:
Proposal iterated. 

* Defaults that that not use environment variables.
* Configurations are now overwritten entirely without needing to call `builder.clear`.
* Actually, `Log::Builder` is no longer exposed.

The `Log.setup` overloads should allow simple configurations straight away, and the yielding method can be used for more complex binding, still without exposing the builder.

src/log.cr
straight-shoota: I don't think the configuration over environment variables is a problem by itself. But we need to figure out how to treat them when custom configuration comes into play. I don't think they can be easily integrated with a custom setup. And it their effect is reset like the default configuration that might be unexpected as well.

src/log/builder.cr
```diff
@@ -1,9 +1,17 @@
 require "weak_ref"
 
+# Used in `Log.setup` methods to configure the binding to be used.
+module Log::Configuration
```
straight-shoota, line 4: What's the purpose of having a separate module?

waj: Not exposing the builder for the setup configuration in case we want to change the implementation details in the future

src/log.cr
waj: > I don't think we need a another class for this

Sure, we don't. But it wouldn't hurt either and we could separate concerns, like checking if the fd is a tty or not is unnecessary for files. Log4j for example has a `ConsoleAppender`. I think it's much clear than creating a backend with `STDOUT`. Also it could switch to STDERR automatically or with configuration.

src/log/builder.cr
straight-shoota: But we are actually exposing the builder... casting to `Log::Configuration` doesn't prevent using the non-exposed methods on `Log::Builder`. I'd rather avoid the complexity if it's not immediately necessary. Also the `Log::Configuration` sounds very arbitrary. `Log::Builder` describes the purpose much better.

waj: Honestly I don't see how using an interface, err... module, it's adding complexity. But having the builder right now is an implementation detail. It's better for the documentation.

straight-shoota: How is it an implementation detail, when I can create an instance and pass it to `Log.setup`?

waj: The custom builder can be used for advanced usage or for testing. What I'm saying it's an implementation detail is that the object you receive within `setup` is the builder itself. We could change that in the future more easily if we just document the interface it's supposed to be used within the `setup`. Also, for beginners they could just read the docs about `Log.setup` yielding a configuration object and ignoring the builder completely.

If it bothers you too much we can remove it, but we thought it could be a small detail to simplify the api.

src/log.cr
```diff
@@ -66,9 +66,23 @@
...
+# By default entries from all sources with `Info` and above severity will
+# be logged to `STDOUT` using the `Log::IOBackend`.
+#
+# If you need to change the default level, backend or sources call `Log.setup` upon startup.
+#
+# ```
+# Log.setup(:debug)                     # Log debug and above for all sources to STDOUT
+# Log.setup("myapp.*, http.*", :notice) # Log notice and above for myapp.* and http.* sources only
+# backend_with_formatter = Log::IOBackend.new(formatter: custom_formatter)
+# Log.setup(:debug, backend_with_formatter) # Log notice and above for myapp.* and http.* sources only
```
RX14, line 80: The comment for this is wrong

src/log.cr
```diff
@@ -66,9 +66,23 @@
...
+# ### Default logging configuration
+#
+# By default entries from all sources with `Info` and above severity will
+# be logged to `STDOUT` using the `Log::IOBackend`.
+#
+# If you need to change the default level, backend or sources call `Log.setup` upon startup.
+#
+# ```
+# Log.setup(:debug)                     # Log debug and above for all sources to STDOUT
+# Log.setup("myapp.*, http.*", :notice) # Log notice and above for myapp.* and http.* sources only
```
RX14, line 78: Perhaps "Log notice and above for myapp.* and http.* sources, and log nothing for any other source" to be extra clear.

src/log.cr
```diff
@@ -98,25 +115,30 @@
...
+# which severity level to emit. By default entries from all sources with `Info` and above severity will
+# be logged to `STDOUT` using the `Log::IOBackend`.
 #
-# The valid values for `CRYSTAL_LOG_SOURCES` are:
+# To change the level and sources change the environment variables value:
 #
-# * the empty string matches only the top-level source (default)
-# * `*` matches all the sources
+# ```console
+# $ LOG_LEVEL=DEBUG LOG_SOURCES=myapp.* ./bin/app
```
RX14, line 125: I think the more common usecase is "I want info logs for everything, *and* debug logs for `myapp.*`". `LOG_SOURCES` concerns me.

I think `CRYSTAL_LOG_LEVEL=DEBUG` would log everything `> DEBUG`, then `CRYSTAL_LOG_LEVEL="myapp.*=DEBUG"` would leave the default of `*=INFO` in place and override just `myapp`.

waj: I like this idea. It's more flexible without too much complications.
I could also approve this PR assuming sources is always `*` now, and add the parsing later. We could support both `LOG_LEVEL=info` and `LOG_LEVEL="*=info, bar=warning". If the source is missing it assumes `*`.

bcardiff: My idea was to have something simple and later a file base configuration could enable more fine grained control.

Would it make sense to make in this PR `LOG_LEVEL=<level>` use always `*` as source? And add in a later PR the functionality to parse `<source1>=<level1> <source2>=<level2>` syntax

That would be dropping the `LOG_SOURCES` variable in this PR.

---

Note: With @waj we thought about dropping the `CRYSTAL_` prefix of the variable. It doesn't feel right to force an app to mention the language used under the hood.

waj: haha @bcardiff we're in sync

src/log/builder.cr
straight-shoota: It bothers me and I'd not merge it like this. But when others agree, it's fine by me.

src/log.cr
RX14: We should really try and avoid env var conflicts, `LOG_LEVEL` sounds like it'd easily cause havoc when the crystal program starts spawning subprograms. `CRYSTAL_LOG_LEVEL` is uses crystal-specific syntax for the crystal-specific logging library. I don't think it's a good idea in practice.

I'm 100% ok with dropping `LOG_SOURCES` and adding nicer `CRYSTAL_LOG_LEVEL` parsing in a later PR though.

waj: That's not really a valid argument. If the spawned process is built with Crystal you still have a conflict. And if `LOG_LEVEL` causes so much "conflicts" it's because is a broadly accepted name for such purpose. You can still customize the variable name for your apps if you want.

bcardiff:
Implemented the proposal in https://github.com/crystal-lang/crystal/pull/9145#discussion_r413963144 

I left the default_sources argument since that is still handy to narrow the source you want with the `LOG_LEVEL=<severity>` without path.

The `LOG_LEVEL="<source1>=<level1> <source2>=<level2>"` syntax support will come in a later PR.

src/log/setup.cr
```diff
@@ -0,0 +1,39 @@
...
+
+  # Setups logging for all sources using the specified *level*, *backend*.
+  def self.setup(level : Log::Severity = Log::Severity::Info,
+                 backend : Log::Backend = IOBackend.new,
+                 *, builder : Log::Builder = Log.builder)
+    Log.setup("*", level, backend, builder: builder)
+  end
+
+  # Setups logging for *sources* using the specified *level*, *backend*.
+  def self.setup(sources : String = "*",
```
RX14, line 16: I'd rather get rid of all the sources-based `Log.setup` overloads in this PR, and leave it until the infra for parsing `k=v` is merged. This is because we'll then have a canonical way of easily specifying log levels, and the `Log.setup` interface here behaves quite differently (for example not keeping the *=INFO default, instead disabling logging for all but the mentioned sources), creating confusion.

I'll accept it if you want to leave these in, but to me, I think people will be fine with the `Log.setup(&block)` interface *for now* if they want to setup per-source levels.

waj: I think those overloads are useful for simple scenarios, where even configuring through environment variables is not needed or undesired. For example `Log.setup(:warning)` or `Log.setup("http.server", :info)` to just show request logs. Do you say you would use the same syntax as environment variables instead, even for programmatic setup?

It has to be understood that calling `Log.setup` is never additive. It always replace the current configuration.

src/log/builder.cr
bcardiff: It's either a module like this or a dsl struct to be returned that will hide the builder.

The fact that `Log.setup` receives a builder it's a side effect of testing. Although it can be used for creating isolated pool of loggers. But I would rather discourage that scenario.

src/log/setup.cr
RX14: > It has to be understood that calling `Log.setup` is never additive. It always replace the current configuration.

I'm just not sure that this is the semantic that users want in practice. For all means, I'm not going to delay this PR over that discussion though.

src/log.cr
```diff
@@ -66,28 +66,45 @@
...
-# backend = Log::IOBackend.new
-# Log.builder.bind "*", :warning, backend
-# Log.builder.bind "db.*", :debug, backend
-# Log.builder.bind "*", :error, ElasticSearchBackend.new("http://localhost:9200")
+# Log.setup |c|
+#   backend = Log::IOBackend.new
+#
+#   c.bind "*", :warning, backend
+#   c.bind "db.*", :debug, backend
+#   c.bind "*", :error, ElasticSearchBackend.new("http://localhost:9200")
```
paulcsmith, line 106: I am not sure when to use `Log.setup` v `bind`. Maybe that could be explain here? It seems to me that `Log.setup` does the same thing but I'm not sure

src/log/setup.cr
```diff
@@ -0,0 +1,39 @@
+class Log
+  # Setups logging bindings discarding all previous configurations.
+  def self.setup(*, builder : Log::Builder = Log.builder)
+    builder.clear
+    yield builder.as(Configuration)
+  end
+
+  # Setups logging for all sources using the specified *level*, *backend*.
+  def self.setup(level : Log::Severity = Log::Severity::Info,
+                 backend : Log::Backend = IOBackend.new,
```
paulcsmith, line 10: Is it possible to have the backend instead be `nil` and use whatever the *existing* backend is (if set) and if none is set *then* set the IOBackend? I'm imagining something like this:

```crystal
if prod?
  Log.setup(:info, some_fancy_backend)
else
  Log.setup(:info, IOBackend.new)
end

Log.setup("db.*", :debug) # I want to use whatever is specified in the first `Log.setup`
```

src/log.cr
bcardiff: A `Log.setup` will configure entirely the logging module.
`c.bind` will be used to define the binding.

If you only need 1 bind, there is an overload of `Log.setup` that does that.

Since calling `Log.setup` clears the existing binding, the last call of `Log.setup` is the one that counts.

src/log/setup.cr
bcardiff: No, because the reasons pointed before. The last setup is the one that counts, they are not additive. Otherwise we are back to the `clear` method.

paulcsmith: Oh good. I thought this was like before where it would append backends!

src/log.cr
paulcsmith: Oh I see. I misunderstood what setup was doing

src/log/setup.cr
paulcsmith: Oh wait so if you Log.setup and then later bind, will it append the backend like it does now, or overwrite the backend? I think it might be confusing if binding appended the backend instead of overwriting. 

But this is just a thought. Maybe I’m just overthinking it

bcardiff: With the `Log.setup` methods we are actually able to hide `Log::Builder`. it will still work if you want to do `Log.builder.bind(...)` as long as there is no later `Log.setup` or `Log.builder.clear`

PR #9146, opened 2020-04-20
Log: Allow override context within logging calls

bcardiff:
Extracted from https://forum.crystal-lang.org/t/proposal-to-have-all-data-be-key-value-data-in-log/1966/14

Let's add a spec to ensure that will keep working + allow the restoring of the previous context.

cc: @paulcsmith

### DIFF
--- a/spec/std/log/log_spec.cr
+++ b/spec/std/log/log_spec.cr
@@ -106,4 +106,16 @@ describe Log do
 
     backend.entries.first.context.should eq(Log::Context.new({a: 1}))
   end
+
+  it "context can be changed within the block and is restored" do
+    Log.context.set a: 1
+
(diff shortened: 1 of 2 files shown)


straight-shoota:
Not sure if this was discussed somewhere, but could we yield the context? That would make it more explicit at callsite that mutating methods affect the local context only.

bcardiff:
@straight-shoota the context is immutable, so we would need to yield some other structure.

Yielding the whole Log::Entry could be an option, but then we need to allow entries to be mutable also, and initialize them without message.

All the options I can think of that depart from these facts yeilds to more API to learn that will have their own quirks IMO.

straight-shoota:
Ah, okay. And `Log.conetxt.set()` works because it's not assigned to a local variable.

PR #9149, opened 2020-04-21
Implement Process.times on Windows.

### LINKED ISSUE #9131: [RFC] Process.times on Windows
Process.times is not implemented on Windows now. This is a proposal how to implement it.

CPU usage consumed by a process can be got by [GetProcessTimes][]. It is easy to fill [Process::Tms][]#utime and #stime. Just calling `GetProcessTimes(GetCurrentProcess(), ....)`. I want to hear opinions about #cutime and #cstime.

The #cutime and #cstime values come from `getrusage()` with `RUSAGE_CHILDREN` on Unix. They include the resources used by not only direct children but also further descendants on [Linux][]. [POSIX][] just says "children of the current process" and doesn't notice whether they include the resources used by further descendants or not. The manpage of [FreeBSD][] and [macOS][] is same with that of POSIX. Does the word "children" in English include descendants in this context?

By the way I came up with two options.

1. Doesn't support #cutime and #cstime. They are zeros always.
2. Support resources about direct children, not including further descendants.
   Call `GetProcessTimes` just before [this line] and add up to a class variable of Crystal::System::Process.

I prefer the former because most users, I guess, won't care about #cutime and #cstime.
In addition the latter is similar to but not same with that on Linux.

[GetProcessTimes]: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getprocesstimes
[Process::Tms]: https://crystal-lang.org/api/0.34.0/Process/Tms.html
[Linux]: http://man7.org/linux/man-pages/man2/getrusage.2.html
[POSIX]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrusage.html
[macOS]: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/getrusage.2.html
[FreeBSD]: https://www.freebsd.org/cgi/man.cgi?query=getrusage&apropos=0&sektion=2&format=html
[this line]: https://github.com/crystal-lang/crystal/blob/456b4ebdbed36380c30f858d27b638f098020b5f/src/crystal/system/win32/process.cr#L15

kubo:
Process.times using Win32 API GetProcessTimes (#9131)

I tested this PR by the following code:
```crystal
require "benchmark"

Benchmark.bm do |x|
  x.report("consuming user time:") do
    start_time = Time.monotonic
    n = 0_u128
    while Time.monotonic - start_time < 5.seconds
      n += 1
    end
  end
  x.report("consuming system time as possible:") do
    start_time = Time.monotonic
    while Time.monotonic - start_time < 5.seconds
      File.open(File::NULL).close
    end
  end
end
```

On Linux to check the test code:
```
consuming user time:                 4.994000   0.000000   4.994000 (  5.000003)
consuming system time as possible:   2.305000   2.821000   5.126000 (  5.000813)
```

On Windows after applying this patch:
```
consuming user time:                 5.000000   0.000000   5.000000 (  5.000005)
consuming system time as possible:   1.812500   3.421875   5.234375 (  5.000005)
```

Closes #9131

### DIFF
--- a/src/crystal/system/win32/process.cr
+++ b/src/crystal/system/win32/process.cr
@@ -83,7 +83,14 @@ struct Crystal::System::Process
   end
 
   def self.times
-    raise NotImplementedError.new("Process.times")
+    if LibC.GetProcessTimes(LibC.GetCurrentProcess, out create, out exit, out kernel, out user) == 0
+      raise RuntimeError.from_winerror("GetProcessTimes")
+    end
+    ::Process::Tms.new(
+      Crystal::System::Time.filetime_to_f64secs(user),
+      Crystal::System::Time.filetime_to_f64secs(kernel),
+      0,
+      0)
   end
 
   def self.fork
--- a/src/crystal/system/win32/time.cr
+++ b/src/crystal/system/win32/time.cr
@@ -35,6 +35,10 @@ module Crystal::System::Time
     ::Time.utc(seconds: seconds, nanoseconds: nanoseconds)
   end
 
+  def self.filetime_to_f64secs(filetime) : Float64
+    ((filetime.dwHighDateTime.to_u64 << 32) | filetime.dwLowDateTime.to_u64).to_f64 / FILETIME_TICKS_PER_SECOND.to_f64
+  end
+
   @@performance_frequency : Int64 = begin
     ret = LibC.QueryPerformanceFrequency(out frequency)
     if ret == 0
--- a/src/lib_c/x86_64-windows-msvc/c/processthreadsapi.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/processthreadsapi.cr
@@ -39,6 +39,8 @@ lib LibC
                      bInheritHandles : BOOL, dwCreationFlags : DWORD,
                      lpEnvironment : Void*, lpCurrentDirectory : LPWSTR,
                      lpStartupInfo : STARTUPINFOW*, lpProcessInformation : PROCESS_INFORMATION*) : BOOL
+  fun GetProcessTimes(hProcess : HANDLE, lpCreationTime : FILETIME*, lpExitTime : FILETIME*,
+                      lpKernelTime : FILETIME*, lpUserTime : FILETIME*) : BOOL
 
   PROCESS_QUERY_INFORMATION = 0x0400
 end
--- a/src/process.cr
+++ b/src/process.cr
@@ -62,9 +62,7 @@ class Process
   end
 
   # Returns a `Tms` for the current process. For the children times, only those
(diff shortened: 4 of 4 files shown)


src/crystal/system/win32/process.cr
```diff
@@ -190,3 +194,7 @@ private def close_handle(handle)
     raise RuntimeError.from_winerror("CloseHandle")
   end
 end
+
+private def filetime_to_f64secs(filetime)
+  ((filetime.dwHighDateTime.to_u64 << 32) | filetime.dwLowDateTime.to_u64).to_f64 / 1e7
```
RX14, line 199: `Time::FILETIME_TICKS_PER_SECOND` would be better than `1e7` here.

src/crystal/system/win32/process.cr
```diff
@@ -190,3 +194,7 @@ private def close_handle(handle)
     raise RuntimeError.from_winerror("CloseHandle")
   end
 end
+
+private def filetime_to_f64secs(filetime)
```
straight-shoota, line 198: We could move this method to `Crystal::System::Time` because there's already the similar `filetime_to_seconds_and_nanoseconds`.

kubo: `Crystal::System::Time.filetime_to_seconds_and_nanoseconds` changes the epoch from year 1601 to 1970. On the other hand this `filetime_to_f64secs` just changes integer in 100-nanoseconds to f64 in seconds. When this method is moved to `Crystal::System::Time`, I think it is better to rename the method name or add a comment in order not to misuse it. However I don't come up with its name nor comment. Do you have any idea? Otherwise, am I thinking too match?

kubo: I fixed it as requested anyway.

straight-shoota: Na, that's fine. These methods are only internal API anyway. I just want them to be at the same place to see what's already there. Maybe the next API port needs something similar and would write another local helper method. Better to have related methods grouped.

src/crystal/system/win32/process.cr
```diff
@@ -83,7 +83,14 @@ struct Crystal::System::Process
   end
 
   def self.times
-    raise NotImplementedError.new("Process.times")
+    if LibC.GetProcessTimes(LibC.GetCurrentProcess, out create, out exit, out kernel, out user) == 0
```
Sija, line 86: Since `LibC.GetProcessTimes` returns `Bool` this will never enter into this branch.

straight-shoota: `alias BOOL = Int32`

PR #9150, opened 2020-04-21
Log: empty singleton context

bcardiff:
This should reduce stressing the GC a bit due to new context being created for each fiber and merging context against an empty one.

### DIFF
--- a/src/log/context.cr
+++ b/src/log/context.cr
@@ -6,6 +6,15 @@
 class Log::Context
   Crystal.datum types: {nil: Nil, bool: Bool, i: Int32, i64: Int64, f: Float32, f64: Float64, s: String, time: Time}, hash_key_type: String, immutable: true, target_type: Log::Context
 
+  @@empty = Log::Context.new
+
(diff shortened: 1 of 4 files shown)


src/log/context.cr
```diff
@@ -6,6 +6,15 @@
...
   Crystal.datum types: {nil: Nil, bool: Bool, i: Int32, i64: Int64, f: Float32, f64: Float64, s: String, time: Time}, hash_key_type: String, immutable: true, target_type: Log::Context
 
+  @@empty = Log::Context.new
+
+  # Returns an empty `Log::Context`.
+  #
+  # NOTE: Since `Log::Context` is immutable, it's safe to share this instance.
+  def self.empty
+    @@empty
+  end
```
Sija, line 16:
```suggestion
  # Returns an empty `Log::Context`.
  #
  # NOTE: Since `Log::Context` is immutable, it's safe to share this instance.
  class_getter empty : Log::Context = Log::Context.new
```

PR #9153, opened 2020-04-21
Add support for Path arguments to more methods

### LINKED ISSUE #7688: Undefined method 'split' for Path

straight-shoota:
Closes #7688

### DIFF
--- a/spec/std/dir_spec.cr
+++ b/spec/std/dir_spec.cr
@@ -362,9 +362,17 @@ describe "Dir" do
   end
 
   describe "cd" do
-    it "should work" do
+    it "accepts string" do
       cwd = Dir.current
       Dir.cd("..")
+      Dir.current.should_not eq(cwd)
+      Dir.cd(cwd)
+      Dir.current.should eq(cwd)
+    end
+
+    it "accepts path" do
+      cwd = Dir.current
+      Dir.cd(Path.new(".."))
       Dir.current.should_not eq(cwd)
       Dir.cd(cwd)
       Dir.current.should eq(cwd)
@@ -376,7 +384,17 @@ describe "Dir" do
       end
     end
 
-    it "accepts a block" do
+    it "accepts a block with path" do
+      cwd = Dir.current
+
+      Dir.cd(Path.new("..")) do
+        Dir.current.should_not eq(cwd)
+      end
+
+      Dir.current.should eq(cwd)
+    end
+
+    it "accepts a block with string" do
       cwd = Dir.current
 
       Dir.cd("..") do
@@ -385,6 +403,10 @@ describe "Dir" do
 
       Dir.current.should eq(cwd)
     end
+  end
+
+  it ".current" do
+    Dir.current.should eq(`#{{{ flag?(:win32) ? "cmd /c cd" : "pwd" }}}`.chomp)
   end
 
   describe ".tempdir" do
@@ -435,6 +457,14 @@ describe "Dir" do
     end
 
     filenames.includes?("f1.txt").should be_true
+  end
+
+  describe "#path" do
+    it "returns init value" do
+      path = datapath("dir")
+      dir = Dir.new(path)
+      dir.path.should eq path
+    end
   end
 
   it "lists entries" do
--- a/src/process/executable_path.cr
+++ b/src/process/executable_path.cr
@@ -31,20 +31,28 @@ class Process
   # Searches an executable, checking for an absolute path, a path relative to
   # *pwd* or absolute path, then eventually searching in directories declared
   # in *path*.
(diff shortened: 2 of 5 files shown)


src/process/executable_path.cr
```diff
@@ -31,20 +31,28 @@ class Process
   # Searches an executable, checking for an absolute path, a path relative to
   # *pwd* or absolute path, then eventually searching in directories declared
   # in *path*.
-  def self.find_executable(name, path = ENV["PATH"]?, pwd = Dir.current)
-    if name.starts_with?(File::SEPARATOR)
-      return name
+  def self.find_executable(name : Path | String, path : String = ENV["PATH"]?, pwd : Path | String = Dir.current) : String?
```
Sija, line 34:
```suggestion
  def self.find_executable(name : Path | String, path : Path | String? = ENV["PATH"]?, pwd : Path | String = Dir.current) : String?
```

straight-shoota: Nilable is correct. But I don't think it should accept `Path` because the value is a path list, which can't be expressed as that data type. So I'd rather not give that option.
Maybe we could redesign this interface even more and make it `Enumerable(Path)` or so but I'd leave that for another day.

spec/std/dir_spec.cr
```diff
@@ -387,6 +405,10 @@ describe "Dir" do
     end
   end
 
+  pending_win32 ".current" do
+    Dir.current.should eq(`#{{{ flag?(:win32) ? "cd" : "pwd" }}}`.chomp)
```
Sija, line 409: Why the interpolation?

oprypin: This is all correct, there's no other way to put a string into ``.

On the other hand, there's no plan to make this construct work like this on Windows, so it's not good to add it pending like this. You can use `cmd /c cd` maybe...

Sija: Not true, see https://carc.in/#/r/8xpe

oprypin: Fine. The only safe way.

https://carc.in/#/r/8xpl
https://carc.in/#/r/8xpp

I **just** fixed a few of these. It'd be a change for worse.

https://github.com/crystal-lang/crystal/pull/9054/files#diff-42858a99f8c45d6bcda97f09679eea8dL284
https://github.com/crystal-lang/crystal/pull/9106/files#diff-a49a030cea42f05be9fa6ee72cb04404L10

oprypin: And the issue where I describe the latter fix: https://github.com/crystal-lang/crystal/issues/9089

Sija:
How about `Dir.current` returning `Path`?

straight-shoota:
@Sija Yeah that should be the goal. But it's also a breaking change. I want to ship the non-breaking changes to prepare the API first.
There are also a lot of other methods on `Dir` and `File` that should return `Path`, not just `Dir.current`.

straight-shoota:
Rebased on master and squashed. Needs an approval again.

RX14:
Can we get rid of the force push dismiss? It's *really* useful to know whether past reviews were :+1: or :-1: *before* the force push, and this setting completely destroys that info.

waj:
> Can we get rid of the force push dismiss? It's _really_ useful to know whether past reviews were 👍 or 👎 _before_ the force push, and this setting completely destroys that info.

@RX14 the setting says "Dismiss stale pull request approvals when new commits are pushed". So isn't just 👍 that are dismissed?

RX14:
It seems to do both, and everyone seems to dislike the setting in practice.

With better UX and better semantics from github, it could be a wonderful feature, but in practice, right now, it seems to do more harm than good.

waj:
I don’t get it. It’s just approvals that get dismissed but comments are still there.

waj:
Oh you mean if someone request changes you don’t see the red flag anymore after the push?

asterite:
I think he means it's annoying having to reapprove things when minor fixes are. made.

straight-shoota:
We should better discuss this somewhere else. Maybe on the forum.

RX14:
@asterite she means both what you and waj said :)

PR #9158, opened 2020-04-23
Add Union#nilable? macro method

Blacksmoke16:
Returns `true` if any type in the union is `Nil`.

### DIFF
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -1947,7 +1947,7 @@ module Crystal
       end
     end
 
-    describe "union methods" do
+    describe Union do
       it "executes types" do
         assert_macro "x", %({{x.types}}), [Crystal::Union.new(["Int32".path, "String".path] of ASTNode)] of ASTNode, "[Int32, String]"
       end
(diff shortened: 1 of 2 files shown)


spec/compiler/macro/macro_methods_spec.cr
```diff
@@ -1960,6 +1960,35 @@ module Crystal
         assert_macro "x", %({{x.resolve?}}), [Crystal::Union.new(["Int32".path, "String".path] of ASTNode)] of ASTNode, "(Int32 | String)"
         assert_macro "x", %({{x.resolve?}}), [Crystal::Union.new(["Int32".path, "Unknown".path] of ASTNode)] of ASTNode, "nil"
       end
+
+      describe "#nilable?" do
```
Sija, line 1964:
```suggestion
      context "#nilable?" do
```

spec/compiler/macro/macro_methods_spec.cr
```diff
@@ -1960,6 +1960,35 @@ module Crystal
...
+        it "without nil" do
```
Sija, line 1965: Please, use descriptive names, ditto elsewhere.

Blacksmoke16: How could it get more descriptive?  The method is called `nilable?` and it either can or cannot have nil.

Sija: it's suppose to be read like a sentence - see other examples.

jhass:
Mmh, I would have expected `Union#types` to return an array of each type in the union, but it seems to return just an array with the union as its only element? If we fix this we could just do `types.includes?(Nil)` and have achieved a more general solution at the same time, no?

Blacksmoke16:
Hold on.  I have some questions: https://play.crystal-lang.org/#/r/8xvp

The specs pass fine when the underlying type is a `TypeNode`; which also seems to be the default type of a Union.  

An actual `Union` type seems to only exist within method arguments and maybe return type.  I can use `#types` on the union and see that it has `Int32` and `Nil`, but `includes? Nil` is false.  This makes sense as to by the current code work, by first resolving the `Union` to a `TypeNode`; which means the specs are really testing `TypeNode#nilable?`.

It also seems that a `Union` within an annotation is a `Generic`?

Blacksmoke16:
@asterite Any thoughts on what's going on there? ^

asterite:
I didn't look into this, but the compiler will sometimes use `Union` for `A | B` and sometimes `Genreic`.

PR #9169, opened 2020-04-24
Add Path#relative_to

straight-shoota:
This PR adds a method `Path#relative_to` which is essentially the inverse of `#expand` and calculates the path that describes the relationship between two other paths.

Also includes a commit which applies this method in stdlib.
The compiler also has a number of use cases, but I haven't included them here because it's a bigger refactoring and will be a subsequent PR.

### DIFF
--- a/src/path.cr
+++ b/src/path.cr
@@ -946,6 +946,95 @@ struct Path
     end
   end
 
+  # Returns a relative path that is lexically equivalent to `self` when joined
+  # to *base* with an intervening separator.
+  #
+  # The returned path is in normalized form.
+  #
+  # That means with normalized paths `base.join(target.relative_to(base))` is
+  # equivalent to `target`.
+  #
+  # Returns `nil` if `self` cannot be expressed as relative to *base* or if
+  # knowing the current working directory would be necessary to resolve it. The
+  # latter can be avoided by expanding the paths first.
+  def relative_to?(base : Path) : Path?
+    base_anchor = base.anchor
+    target_anchor = self.anchor
+
+    # if paths have a different anchors, there can't be a relative path between
+    # them.
+    if base_anchor != target_anchor
+      return nil
+    end
+
+    # work on normalized paths otherwise we would need to backtrack on `..` parts
+    base = base.normalize
+    target = self.normalize
+
+    # check for trivial case of equal paths
+    if base == target
+      return new_instance(".")
+    end
+
+    base_iterator = base.each_part
+    target_iterator = target.each_part
+
+    if target_anchor
+      # process anchors, we have already established they're equal
+      base_iterator.next
+      target_iterator.next
+    end
+
+    # consume both paths simultaneously as long as they have identical components
+    base_part = base_iterator.next
+    target_part = target_iterator.next
+    while base_part.is_a?(String) && target_part.is_a?(String)
+      if base_part.compare(target_part, case_insensitive: windows?) != 0
+        break
+      end
+
+      base_part = base_iterator.next
+      target_part = target_iterator.next
+    end
+
+    path = new_instance("")
+
+    # base_path is not consumed, so we go up before descending into target_path
+    if base_part.is_a?(String)
+      # Can't relativize upwards from current working directory without knowing
+      # its path
+      if base_part == ".."
+        return nil
+      end
+
+      path /= ".." unless base_part == "."
+      base_iterator.each do
+        path /= ".."
+      end
+    end
+
+    # target_path is not consumed, so we append what's left to the relative path
+    if target_part.is_a?(String)
+      path /= target_part
+      target_iterator.each do |part|
+        path /= part
+      end
+    end
+
+    return path
+  end
+
+  # :ditto:
+  def relative_to?(base : String) : Path?
+    relative_to?(new_instance(base))
+  end
+
+  # Same as `#relative_to` but returns `self` if `self` can't be expressed as
+  # relative path to *base*.
+  def relative_to(base : Path | String) : Path
+    relative_to?(base) || self
+  end
+
   # Compares this path to *other*.
   #
   # The comparison is performed strictly lexically: `foo` and `./foo` are *not*
--- a/src/spec/junit_formatter.cr
+++ b/src/spec/junit_formatter.cr
@@ -119,13 +119,8 @@ module Spec
     end
 
     private def classname(result)
-      path = Path[result.file].expand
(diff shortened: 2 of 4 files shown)


straight-shoota:
Thanks @Sija. But I'm not very convinced it would be an improvement, both styles work and I probably like the current one a tiny bit better.

oprypin:
Can you slay this one too?
https://github.com/crystal-lang/crystal/blob/5609eef6a323c76763df7b4acef45a0733f99747/src/compiler/crystal/util.cr#L4

oprypin:
Is there any foolproof way to check whether one path is inside another? Wouldn't want to do something like `!relative_to.starts_with("..")` (which I already know is wrong)

straight-shoota:
@oprypin `relative_filename` is on my radar. I have a follow-up PR in waiting which refactors path useage in the compiler. It depens on this one and #9153.

For common ancestor checking I was thinking about a `#common_parent` method. But that's also a different feature.

src/path.cr
```diff
@@ -946,6 +946,95 @@ struct Path
...
+  # :ditto:
+  def relative_to?(base : String) : Path?
+    relative_to?(new_instance(base))
+  end
+
+  # Same as `#relative_to` but returns `self` if `self` can't be expressed as
+  # relative path to *base*.
+  def relative_to(base : Path | String) : Path
+    relative_to?(base) || self
+  end
```
oprypin, line 1036: I don't think this function with the fallback should exist. The choice on what to do for that case should be explicit. Otherwise it could be a security hazard. Regardless of that, I think users should almost always use the variant with the question mark, and the fact that it's easier to type the one without the question mark (also easy to make a mistake) makes me worry.

oprypin:
Well no, it's not a different feature. Python has the same function for both purposes.
Perhaps you misunderstood me, I'm not talking about common ancestor.

"/foo/bar" is under "/foo" but "/foobar" is not under "/foo".

And this `relative_to` impl knows that, returning `"foo"` for the former and `"../foobar"` for the latter.

Now, forming these paths is nice and all, but we're *so close* to also being able to answer the question "is path1 under path2" directly. And the method really *has* that information. However, with just its return value, we'd have to resort to a construct like this:

```crystal
def is_path_under_base(path = "/foo/bar", base = "/foo")
  Path[path].relative_to?(base).try &.parts.first != ".."
end
```

In my view, this is a very important feature, often in the contexts of security. It's important to be able to confine allowed paths to a subdirectory.

Now back to Python. See, [pathlib.Path.relative_to](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to) has a different behavior. It will refuse to return any path above the base (anything with "../"). So you can implement the above as such:

```python
def is_path_under_base(path="/foo/bar", base="/foo"):
  try:
    pathlib.Path[path].relative_to(base)
    return True
  except ValueError:
    return False
```

(yea, not a nice showcase of Python, but whatever, the point is there's no room for mistake).

Now, Python *also* has a legacy function [os.path.relpath](https://docs.python.org/3/library/os.path.html#os.path.relpath) that behaves (AFAIK exactly) like this one's `a.relative_to(b.expand)`. But that's meh.

---

To be fair, one *could* also implement what I'm requesting as `path.common_ancestor(base) == base`, but I, for one, wouldn't trust off-hand that such an implementation has no corner cases, so it really needs to be a dedicated function (and if based on `relative_to` it would probably be safer than if based on `common_parent`).

---

How do we get there? Well, perhaps you wouldn't want to just drop the use case of returning paths *with* `".."`, so maybe as an option like `allow_paths_outside_base: true` to `relative_to?`. And return `nil` in more situations then.

src/path.cr
straight-shoota: I originally had only the variant the returns `nil` if a relative path can't be resolved. But every single use case in stdlib and compiler would fall back to the original path, so I'm quite confident this is actually the most common use case.

This is not a security feature. If you rely on `relative_to` for sandboxing paths your security is already broken.

oprypin:
May I suggest

```crystal
class Path
  def relative_to?(base, confined = true) : Path?
    # same but returns nil instead of "../*" if *confined*.
  end

  def relative_to(base) : Path
    relative_to?(base.expand, confined: false).not_nil!
  end
end

def is_path_under_base(path = "/foo/bar", base = "/foo")
  !!Path[path].relative_to?(base)
end
```

src/path.cr
straight-shoota: Most use cases also have `Dir.current` as argument, so it would even make sense as a default argument. But `path.relative_to()` looks strange.
Maybe `Path#relativize` would be a better name. That's what I called it initially but I felt `#relative_to` better expresses that the argument is the base.

src/spec/junit_formatter.cr
```diff
@@ -119,13 +119,9 @@ module Spec
...
     private def classname(result)
-      path = Path[result.file].expand
-      path = path.to_s
-        .lchop(Dir.current)
-        .rchop(path.extension)
-        .lchop(Path::SEPARATORS[0])
+      path_without_extension = result.file.rchop(File.extname(result.file))
 
-      Path[path].parts.join(".")
+      Path.new(path_without_extension).expand.relative_to(Dir.current).parts.join(".")
```
j8r, line 124: Can be simplified
```cr
path = Path.new result
path.expand.relative_to(Dir.current).parts.join('.').rchop path.extname
```

straight-shoota:
The purpose of this method is to strip unnecessary base paths for displaying short paths in user interfaces.

Security applications related to path sandboxing are certainly related, but require different features. I haven't looked at that in detail, so I don't have a full understanding of it. But for example a strong implementation of `is_path_under_base` would need to take symlinks into account. There should probably be an option for that because it depends on the use case whether links outside the base path should be considered valid or not. But this can't be determined by the paths alone, you need to look at the actual file system.

I'd rather not introduce a half-baked implementation until the security aspects are thorougly explored.

src/path.cr
straight-shoota: In case you want to take a look at some use cases: https://github.com/straight-shoota/crystal/compare/feature/path-relative...straight-shoota:refactor/compiler-path

straight-shoota:
Finally... Thanks for your patience @RX14

waj:
@RX14 @straight-shoota all right, all right... I just reverted the config 😆 
But please, dismiss existing reviews manually and/or give more time when considerable changes are pushed after the last approval.

Issue #9170, opened 2020-04-24
FileUtils.cp_r will fail if dest_path exists


xmonader:
# Submitting bugs

Make sure to review these points before submitting issues - thank you!

FileUtils.cp_r will fail if the dest path exists
https://github.com/crystal-lang/crystal/blob/ccf46c095a7e10cb120d4d04c96333858be4aaa2/src/file_utils.cr#L110

# version
```
Crystal 0.34.0 [4401e90f0] (2020-04-06)

LLVM: 8.0.0
Default target: x86_64-unknown-linux-gnu
```

Issue #5231, opened 2017-11-01
FileUtils: refactoring


j8r:
What are you thoughts about this module?

Ideas are to merge the methods to the `File` and `Dir` Objects, and replace the `_r` by a boolean `recurse` argument.

Sija:
in any case please use `recursive` over `recurse`.

j8r:
To sum up, the propositions for `FileUtils`:
- Remove the aliases `cd`, `mkdir`, `pwd`, `rm`, `rmdir` and `touch`
- Move `FileUtils.cmp` to `File.compare`
- Move `FileUtils.cp` to `File.copy`
- Move `FileUtils.cp_r` to `Dir.copy`
- Move `FileUtils.rm_r` to `Dir.delete`
- Create `Dir.chown` and `Dir.chmod` for recursive action through the directory
- Change `Dir.mkdir` and `Dir.rmdir` to `Dir.create` and `Dir.delete`
- Add maybe an overload `force : Bool = false` to replace `rm_rf`, and `recursive : Bool = false` for `mkdir_p`

For recursiveness, a others options are to create overloads `recursive : Bool = false` for each method, or create a method that yields all sub files/directories.

RX14:
`FileUtils` must go.

Ideally a `File` instance would represent an opened file, and a `Dir` instance would represent an opened dir, both would include the `Path` module which represented operations on paths. We can then work out what static methods we want based on that completely refactored core where everything can be done in terms of `Path` instances.

I've used java's `Path` class and it truly was refreshing.

j8r:
@RX14 Ok, I've nothing against `FileUtils`, but there are inconsistences. e.g. there is a `File.delete` but no `Dir.delete`, but there is no `File.unlink` but a `Dir.rmdir`.

So anyone has an idea to replace this not-so-nice `_r` ? I have opened a PR to implement a `chown_r` and a `chmod_r`, but after discussion maybe we can consider an alternative for recursive action.

Anyway, why keeping aliases?

RX14:
@j8r What aliases?

With my idea we'd just obviate the `recursive` by making it as easy as `path.each_child(recursive: true, &.delete)`

j8r:
[...]

PR #9175, opened 2020-04-24
Cleanup File & FileUtils

bcardiff:
* Add missing methods to File of features available in FileUtils: same_content?/cmp copy/cp
* Only `FileUtils.cp` treats the dest as a potential directory. `File.copy` works with file paths.
* I left a TODO related to #8926, #8919 for `File.copy` optimizations via syscall.
* Deprecate `Dir.rmdir` in favor of `Dir.delete` (`File.delete` is the file counter-part)
* Another breacking change is the straight removal of `FileUtils.cmp(IO, IO)` (which seems totally misplaced)
* Allow `File.same?` and `File.touch` with `Path` arguments.

Related to #5231.

The APIs that are not strict aliases in FileUtils are

* `FileUtils.touch, .mv, .mkdir, .mkdir_p, .rm`: because they allow multiple files
* `FileUtils.cp`: because it allows destination to be a directory (and allows multiple files)
* `FileUtils.cp_r`: there is no Dir counter-part and it handle the semantic of File vs Dir source / destination.
* `FileUtils.ln[_s|_sf]`: because it allows destination to be a directory and allows multiple files
* `FileUtils.rm_r[f]`: there is no Dir counter-part and allows multiple directories.

So, the following are the criteria that limited the scope of the PR are:

* I don't think we need to allow multiple input in File/Dir API.
* I'm not sure about doing the special treatment in the destination path that `cp_r`, `ln_s` will fit in File API.
* There is also some treatment of `File.symlink?` in `rm_r` that I hesistated to move into `Dir`, and so, I didn't implement `Dir.delete(recursive: true)` proposal of https://github.com/crystal-lang/crystal/issues/5231#issuecomment-343418239

### DIFF
--- a/src/dir.cr
+++ b/src/dir.cr
@@ -249,8 +249,14 @@ class Dir
   end
 
   # Removes the directory at the given path.
-  def self.rmdir(path : Path | String) : Nil
-    Crystal::System::Dir.delete(path)
+  @[Deprecated("Use `Dir.delete` instead")]
+  def self.rmdir(path : Path | String)
+    delete(path)
+  end
+
+  # Removes the directory at the given path.
+  def self.delete(path : Path | String)
+    Crystal::System::Dir.delete(path.to_s)
   end
 
   def to_s(io : IO) : Nil
--- a/src/file.cr
+++ b/src/file.cr
@@ -177,10 +177,10 @@ class File < IO::FileDescriptor
   # File.same_content?("file.cr", "bar.cr") # => true
   # ```
   def self.same_content?(path1 : Path | String, path2 : Path | String) : Bool
-    return false unless size(path1) == size(path2)
-
     open(path1, "rb") do |file1|
       open(path2, "rb") do |file2|
+        return false unless file1.size == file2.size
+
         same_content?(file1, file2)
       end
     end
--- a/src/file.cr
+++ b/src/file.cr
@@ -729,10 +729,13 @@ class File < IO::FileDescriptor
   # ```
   def self.copy(src : String | Path, dst : String | Path)
     open(src) do |s|
-      open(dst, "wb", s.info.permissions) do |d|
+      open(dst, "wb") do |d|
         # TODO use sendfile or copy_file_range syscall. See #8926, #8919
         IO.copy(s, d)
       end
+
+      # Set the permissions after the content is written in case src permissions is read-only
+      chmod(dst, s.info.permissions)
     end
   end
 
--- a/src/file.cr
+++ b/src/file.cr
@@ -719,6 +719,23 @@ class File < IO::FileDescriptor
     end
   end
 
+  # Copies the file *src* to the file *dst*.
+  # Permission bits are copied too.
+  #
+  # ```
+  # File.chmod("afile", 0o600)
+  # File.copy("afile", "afile_copy")
+  # File.info("afile_copy").permissions.value # => 0o600
+  # ```
+  def self.copy(src : String | Path, dst : String | Path)
+    open(src) do |s|
+      open(dst, "wb", s.info.permissions) do |d|
+        # TODO use sendfile or copy_file_range syscall. See #8926, #8919
+        IO.copy(s, d)
+      end
+    end
+  end
+
   # Returns a new string formed by joining the strings using `File::SEPARATOR`.
   #
   # ```
--- a/src/file.cr
+++ b/src/file.cr
@@ -168,6 +168,24 @@ class File < IO::FileDescriptor
     info(path1.to_s, follow_symlinks).same_file? info(path2.to_s, follow_symlinks)
   end
 
+  # Compares two files *filename1* to *filename2* to determine if they are identical.
+  # Returns `true` if content are the same, `false` otherwise.
+  #
+  # ```
+  # File.write("file.cr", "1")
+  # File.write("bar.cr", "1")
+  # File.same_content?("file.cr", "bar.cr") # => true
+  # ```
+  def self.same_content?(path1 : Path | String, path2 : Path | String) : Bool
+    return false unless size(path1) == size(path2)
+
+    open(path1, "rb") do |file1|
+      open(path2, "rb") do |file2|
+        same_content?(file1, file2)
+      end
+    end
+  end
+
   # Returns the size of the file at *filename* in bytes.
   # Raises `File::NotFoundError` if the file at *filename* does not exist.
   #
--- a/src/file_utils.cr
+++ b/src/file_utils.cr
@@ -381,7 +381,7 @@ module FileUtils
         src = File.join(path, entry)
         rm_r(src)
       end
-      Dir.rmdir(path)
+      Dir.delete(path)
     else
       File.delete(path)
     end
@@ -447,7 +447,7 @@ module FileUtils
   #
   # NOTE: Alias of `Dir.rmdir`
   def rmdir(path : String) : Nil
-    Dir.rmdir(path)
+    Dir.delete(path)
   end
 
   # Removes all directories at the given *paths*.
@@ -459,7 +459,7 @@ module FileUtils
   # ```
   def rmdir(paths : Enumerable(String)) : Nil
     paths.each do |path|
-      Dir.rmdir(path)
+      Dir.delete(path)
     end
   end
 end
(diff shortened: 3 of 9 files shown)


bcardiff:
Finally I forgot to say that I would rather leave FileUtils, specially if the semantics are not 100% translatable to Dir/File since they will offer some terminal semantics I found very handy.

Side notes:
* In C# File.copy does not accept a directory as target [ref](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.copy?view=netcore-3.1)
* In Ruby File.copy was deprecated in favor of FileUtils [ref](https://apidock.com/ruby/v1_8_7_330/File/copy/class) 🙃

src/file_utils.cr
```diff
@@ -480,7 +447,7 @@ module FileUtils
   #
   # NOTE: Alias of `Dir.rmdir`
   def rmdir(path : String) : Nil
```
j8r, line 449: This method can be deprecated too. I don't know how the core members sees the future of this module, but we can start now removing unnecessary aliases at least.

src/file_utils.cr
```diff
@@ -38,39 +38,10 @@ module FileUtils
...
-    while true
-      read1 = stream1.read(buf1.to_slice)
-      read2 = stream2.read_fully?(buf2.to_slice[0, read1])
-      return false unless read2
-
-      return false if buf1.to_unsafe.memcmp(buf2.to_unsafe, read1) != 0
-      return true if read1 == 0
-    end
+  # NOTE: Alias of `File.same_content?`
+  def cmp(filename1 : String, filename2 : String)
```
j8r, line 43: I didn't know `cmp` was a command. Same here, I also think it can be deprecated.

src/dir.cr
```diff
@@ -248,7 +248,13 @@ class Dir
...
 
   # Removes the directory at the given path.
+  @[Deprecated("Use `Dir.delete` instead")]
   def self.rmdir(path)
+    delete(path)
+  end
+
+  # Removes the directory at the given path.
+  def self.delete(path)
     Crystal::System::Dir.delete(path)
```
j8r, line 258:
```suggestion
  def self.delete(path : Path | String)
    Crystal::System::Dir.delete(path.to_s)
```

src/file.cr
```diff
@@ -746,7 +781,7 @@ class File < IO::FileDescriptor
   # in the *filename* parameter to the value given in *time*.
   #
   # If the file does not exist, it will be created.
-  def self.touch(filename : String, time : Time = Time.utc)
+  def self.touch(filename : String | Path, time : Time = Time.utc)
```
j8r, line 784: Unrelated, the good default would be `Time.local` and not `Time.utc`, to be on par with the OS' time.
Otherwise, the file will be dated in the past/future.

straight-shoota: I can't see how that would be the case. `Time.utc` and `Time.local` return the same time instance, just with different location properties.

src/file_utils.cr
bcardiff: I get that you want to deprecate the whole FileUtils module. But why you said this method can be deprecated specifically?

src/file_utils.cr
bcardiff: I think is valid to have this unix-isms in a module. And FileUtils offers some operations that I think does not fit neatly in Dir or File.

If you say this can be deprecated because is a plain alias, then I don't agree. My intention is to keep FileUtils with similar methods and semantics as you would find in a unix shell.

I think @maiha has some opinions regarding why unix-isms are nice.

src/file_utils.cr
j8r: `cmp` is not very common, and quite cryptic name. Not sure this alias really helps.

src/file.cr
j8r: My bad, just tried it. Then, how can I touch a file with the UTC time and not my local time?

src/file_utils.cr
j8r: I was only pointing duplicated methods. There are lots of terminal methods missing, which are very useful and very commonly used like `ls`, `head`, `tail`, `wc`, `find`, etc...

I don't think they will be accepted (will they?), because all aliases at the end can be better written in more idiomatic Crystal.

src/file.cr
straight-shoota: There's no difference because file timestamps are always in UTC.

src/file_utils.cr
j8r: Anyway, sorry it starts to be out of scope :sweat_smile: .
Let's keep it then.

src/file.cr
j8r: Oh right, yep. The files dates are in a UNIX epoch time - like every times in UNIX.

j8r: The conversion is done by our OS.

j8r: The conversion is done by the commands/tools.

src/file_utils.cr
maiha: This is not a "plain alias" as @bcardiff says, nor is it a "duplicate method".

Since Crystal is an "object-oriented" language, the core classes such as `File` and `Dir`, their APIs are provided in an object-oriented paradigm.

On the other hand, since UNIX is a "procedural" concept, it has different APIs.

Here, the two paradigm API entry points are each defined and their **implementations** are only shared.

So, if you want to clean up, I think it's a good idea to remove the confusing annotation part of "NOTE: Alias of".

j8r: Both are called in a functional way – they are class methods, not instance methods. Only the name change.

I think `FileUtils` is just there because of Ruby, a reason why I believe this is nobody asked to add `ls`, which is IMO of the most common and useful command.

`Utils` is also quite bad, if the main purpose of the module is to provide a command-like interface, it would better be called `FileCommands`or `Shell`.

j8r: Both are called in a functional way – they are class methods, not instance methods. Only the name change.

I think `FileUtils` is just there because of Ruby, a reason why I believe this is nobody asked to add `ls`, which is one of the most common and useful command.

`Utils` is also quite bad, if the main purpose of the module is to provide a command-like interface, it would better be called `FileCommands`or `Shell`.

j8r: I don't say everything on FileUtils has to go to trash. I'm only saying we don't have to copy blindly 1-1 the Ruby API, and review the actual module to see what is often used (`cp`, `dir` methods) and those not or confusing like `cmp` or `pwd` (I searched both on GitHub before saying this)

maiha: >  nobody asked to add ls

Well, why not try to think of `FileUtils` as `File (mutable) Utils`?

asterite: Members of the core team like FileUtils so it's going to stay. I know opinions are diverse here but the decision is already taken, so please let's not discuss removing anything.

bcardiff:
Rebased on master. And addressed @RX14 feedback.

I changed what happens on File.copy when destination exists to match the Ruby and `cp` default semantics. https://github.com/crystal-lang/crystal/pull/9175#discussion_r415261607

bcardiff:
@oprypin the permissions support in windows is not there yet, right? The new specs that asserts the permissions are set on the destination file in File.copy are failing.

oprypin:
I don't know, wasn't working on that

bcardiff:
Ok, thanks. 

Ignoring those specs for now. Otherwise the CI complains with:

```
Failures:

  1) File .copy copies permissions
     Failure/Error: File.info(out_path).permissions.should eq(File::Permissions.new(0o700))

       Expected: rwx------ (0o700)
            got: rw-rw-rw- (0o666)

     # \spec\std\file_spec.cr:1260

  2) File .copy overwrites existing destination and permissions
     Failure/Error: File.info(out_path).permissions.should eq(File::Permissions.new(0o700))

       Expected: rwx------ (0o700)
            got: rw-rw-rw- (0o666)

     # \spec\std\file_spec.cr:1275
```

oprypin:
Oh. fileutil is not tested on Windows, so makes sense that stuff is broken if you're moving it out
https://github.com/crystal-lang/crystal/blob/2cbe65b0a13c8a2ce91bd4799f5b96b8e1953097/spec/win32_std_spec.cr#L62

Sija:
@bcardiff Would be sensible to close #9170 perhaps, along with rest of the fixes while you're at it :)

bcardiff:
Done @Sija although I would say it is unrelated to the goal of this PR.

I checked Ruby code and it covers some scenarios we don't. For example, it deals with symlinks in the destination path by removing them.

PR #9176, opened 2020-04-24
Move Debug to Crystal namespace

bcardiff:
Although DWARF, Elf and MachO could be used independently of the runtime, we want to avoid the commitment of having them in the std-lib as public API.

I moved the `:nodoc:` to the beginning of some comments since that's how they work 🤷‍♂️


bcardiff:
Also, this was the last PR on my plate regarding the cleanup of the top-level namespace 🎉

straight-shoota:
Gracias @bcardiff

PR #9177, opened 2020-04-25
HTTP server do SSL handshake in client fiber

### LINKED ISSUE #8108: openssl ssl_accept sometimes does not return, causes server to hang permanently

waj:
This fixes #8108 by forcing the call to `SSL_accept` within the fiber that handles the request.

To implement this fix, I added the `OpenSSL::SSL::Server#start_immediately` property just like Ruby does: https://ruby-doc.org/stdlib-2.5.1/libdoc/openssl/rdoc/OpenSSL/SSL/SSLServer.html

### DIFF
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -63,7 +63,7 @@ class OpenSSL::SSL::Server
   #
   # This method calls `@wrapped.accept` and wraps the resulting IO in a SSL socket (`OpenSSL::SSL::Socket::Server`) with `context` configuration.
   def accept : OpenSSL::SSL::Socket::Server
-    new_server(@wrapped.accept)
+    new_ssl_socket(@wrapped.accept)
   end
 
   # Implements `::Socket::Server#accept?`.
@@ -71,11 +71,11 @@ class OpenSSL::SSL::Server
   # This method calls `@wrapped.accept?` and wraps the resulting IO in a SSL socket (`OpenSSL::SSL::Socket::Server`) with `context` configuration.
   def accept? : OpenSSL::SSL::Socket::Server?
     if socket = @wrapped.accept?
-      new_server(socket)
+      new_ssl_socket(socket)
     end
   end
 
-  private def new_server(io)
+  private def new_ssl_socket(io)
     server = OpenSSL::SSL::Socket::Server.new(io, @context, sync_close: @sync_close)
     server.accept if @start_immediately
     server
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -76,9 +76,7 @@ class OpenSSL::SSL::Server
   end
 
   private def new_ssl_socket(io)
-    server = OpenSSL::SSL::Socket::Server.new(io, @context, sync_close: @sync_close)
-    server.accept if @start_immediately
-    server
+    OpenSSL::SSL::Socket::Server.new(io, @context, sync_close: @sync_close, accept: @start_immediately)
   end
 
   # Closes this SSL server.
--- a/src/openssl/ssl/server.cr
+++ b/src/openssl/ssl/server.cr
@@ -35,6 +35,9 @@ class OpenSSL::SSL::Server
   # Returns `true` if this SSL server has been closed.
   getter? closed : Bool = false
 
+  # When `true` a call to `#accept` will also initiate the SSL handshake.
+  property start_immediately : Bool = true
+
   # Creates a new SSL server wrapping *wrapped*.
   #
   # *context* configures the SSL options, see `OpenSSL::SSL::Context::Server` for details
@@ -60,7 +63,7 @@ class OpenSSL::SSL::Server
   #
   # This method calls `@wrapped.accept` and wraps the resulting IO in a SSL socket (`OpenSSL::SSL::Socket::Server`) with `context` configuration.
   def accept : OpenSSL::SSL::Socket::Server
-    OpenSSL::SSL::Socket::Server.new(@wrapped.accept, @context, sync_close: @sync_close)
+    new_server(@wrapped.accept)
   end
 
   # Implements `::Socket::Server#accept?`.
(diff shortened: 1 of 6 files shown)


src/openssl/ssl/server.cr
```diff
@@ -60,18 +63,24 @@ class OpenSSL::SSL::Server
...
   #
   # This method calls `@wrapped.accept?` and wraps the resulting IO in a SSL socket (`OpenSSL::SSL::Socket::Server`) with `context` configuration.
   def accept? : OpenSSL::SSL::Socket::Server?
     if socket = @wrapped.accept?
-      OpenSSL::SSL::Socket::Server.new(socket, @context, sync_close: @sync_close)
+      new_server(socket)
     end
   end
 
+  private def new_server(io)
```
straight-shoota, line 78: I'd rather call this `new_server_socket` or `new_ssl_socket`. It doesn't create a server (`self` is already the server).

src/openssl/ssl/socket.cr
```diff
@@ -53,10 +53,13 @@ abstract class OpenSSL::SSL::Socket < IO
...
     def initialize(io, context : Context::Server = Context::Server.new, sync_close : Bool = false)
       super(io, context, sync_close)
+    end
+
+    def accept
       begin
         ret = LibSSL.ssl_accept(@ssl)
         unless ret == 1
-          io.close if sync_close
+          @bio.io.close if @sync_close
```
Sija, line 62:
```suggestion
          @bio.io.close if sync_close?
```

src/openssl/ssl/server.cr
```diff
@@ -35,6 +35,9 @@ class OpenSSL::SSL::Server
   # Returns `true` if this SSL server has been closed.
   getter? closed : Bool = false
 
+  # When `true` a call to `#accept` will also initiate the SSL handshake.
+  property start_immediately : Bool = true
```
Sija, line 39:
```suggestion
  property? start_immediately : Bool = true
```

src/openssl/ssl/server.cr
```diff
@@ -60,18 +63,24 @@ class OpenSSL::SSL::Server
...
   def accept? : OpenSSL::SSL::Socket::Server?
     if socket = @wrapped.accept?
-      OpenSSL::SSL::Socket::Server.new(socket, @context, sync_close: @sync_close)
+      new_server(socket)
     end
   end
 
+  private def new_server(io)
+    server = OpenSSL::SSL::Socket::Server.new(io, @context, sync_close: @sync_close)
+    server.accept if @start_immediately
```
Sija, line 80:
```suggestion
    server.accept if start_immediately?
```

src/openssl/ssl/socket.cr
Sija, line 62:
```suggestion
          @bio.io.close if sync_close?
```

src/openssl/ssl/server.cr
asterite: This is not needed. A call is a bit more expensive for the compiler in terms of compiling.

asterite: Put another way: both do the same thing and we don't prefer one over the other.

Sija: I'd argue that in the vein of OOP we should prefer using getters over ivars (ability to overload 'em by subclasses)

asterite: I then call YAGNI

src/openssl/ssl/server.cr
waj: Done

src/openssl/ssl/server.cr
waj: I don't think every boolean property must have a question mark. In particular I like to use it when is a read only property or with methods. Also, I copied this name from Ruby's OpenSSL and I prefer to keep the same name.

waj:
Just added a spec and fixed the segfaults due to a double SSL_free call. Because now the `accept` is not called within the constructor, the finalizer will be executed correctly.

src/openssl/ssl/socket.cr
```diff
@@ -53,15 +53,13 @@ abstract class OpenSSL::SSL::Socket < IO
...
-        unless ret == 1
-          io.close if sync_close
-          raise OpenSSL::SSL::Error.new(@ssl, ret, "SSL_accept")
-        end
-      rescue ex
-        LibSSL.ssl_free(@ssl) # GC never calls finalize, avoid mem leak
-        raise ex
+    end
+
+    def accept
```
RX14, line 58: Isn't this a breaking change?

waj: That's true. I just refactored this fix so it doesn't break compatibility. Thanks!

waj:
@RX14 I agree with you, like a do with @straight-shoota. I just did the smallest change to fix this bug with the current design, but I'm also with you in that these wrappers around OpenSSL api is not ideal.

636f7374:
Great!

PR #9178, opened 2020-04-25
Support DragonFly(BSD)

mneumann:
* This work is based on lib_c/x86_64-freebsd, but carefully
  reviewed and modified.

* Note that the name "DragonFly" and not "DragonFlyBSD" is used in code
  (#ifdef __DragonFLy__) and in LLVM traget triple
  (x86_64-unknown-dragonfly).

* DragonFly's libc does not expose ntons() and htons() functions. As it
  is x86_64 only, use bswap16 intrisic directly. This is also what
  DragonFly's libc uses internally as a C macro.  I did not find a way
  to add ntons() and htons() directly as functions to "lib LibC" so that
  I had to pollute some other code with conditional "if flag?(:dragonfly)"
  code.

* DragonFly does not support fdatasync(). Use fsync() instead.

* This does not (yet) run on an unpatched DragonFly system, as
  DragonFly currently limits the main thread's stack to 4MB which is not
  enough for the crystal compiler to compile itself. I have a patch for
  DragonFly's libpthread in the works that uppers the stack size to
  RLIMIT_STACK. With that patch, crystal can successfully boostrap
  itself.

* A large amount of the test suite passes. Though there are some
  hangs which needs to be fixed in a follow up commit.  For instance,
  the spec for IO.pipe() "raises if trying to read to an IO not opened
  for reading" fails, as pipe(2) on DragonFly creates two bidirectional
  endpoints and neither one is special cased for read- or write-only
  operation, meaning, reading from the "write"-end is perfectly legal on
  DragonFly and blocks until data is available via the "read"-end.

* Cross-compiled from a FreeBSD system. Thanks for making
  cross-compilation that easy!


mneumann:
I run "crystal format" on it :)

RX14:
I was helping @niacat with a NetBSD port recently and I ended up introducing a `bsd` flag to help avoid the `flag?(:freebsd) || flag?(:openbsd) || flag?(:netbsd)` mess. Commits are [here](https://github.com/niacat/crystal/commit/e4f14a10e55316c46268abb0a2e2c84a7c2011af) and [here](https://github.com/niacat/crystal/commit/ccbf9b7667beffe701492dd39c47e78562201331), with some fairly thorough research about what is portable across the BSDs and which parts are bsd-specific.

I haven't reviewed this PR yet, but if you'd like (and you don't have to), we could include the change to add the BSD flag in this PR?

mneumann:
@RX14: having a flag?(:bsd) makes sense! 

I also got a stack overflow when I let crystal compile itself. But I figured out that DragonFly has a fixed 4MB stack limit for the main thread (under pthreads), whereas FreeBSD uses the process resource limit. Are you sure that this is really a stack overflow, or is crystal just using recursion quite a lot? When I upper the stack limit on DragonFly for the main thread, the crystal compiler can compile itself fine. Did you run into this on NetBSD?

RX14:
> Did you run into this on NetBSD?

We're not sure, but it's definitely something to look into, thank you.

src/compiler/crystal/codegen/target.cr
```diff
@@ -82,8 +88,12 @@ class Crystal::Codegen::Target
     @environment.starts_with?("linux")
   end
 
+  def bsd?
+    freebsd? || dragonfly? || openbsd?
```
Sija, line 92:
```suggestion
    freebsd? || openbsd? || dragonfly?
```

src/compiler/crystal/semantic/flags.cr
```diff
@@ -39,6 +39,7 @@ class Crystal::Program
       flags.add "freebsd"
       flags.add "freebsd#{codegen_target.freebsd_version}"
     end
+    flags.add "dragonfly" if codegen_target.dragonfly?
     flags.add "openbsd" if codegen_target.openbsd?
```
Sija, line 43:
```suggestion
    flags.add "openbsd" if codegen_target.openbsd?
    flags.add "dragonfly" if codegen_target.dragonfly?

    flags.add "bsd" if codegen_target.bsd?
```

src/compiler/crystal/codegen/target.cr
straight-shoota: ???

src/compiler/crystal/semantic/flags.cr
RX14: The actual code to add the `bsd` flag here seems to have been lost, but there's no need at all to suggest reordering terms.

src/compiler/crystal/codegen/target.cr
mneumann: I always group freebsd? and dragonfly? closer together as they are more similar. but I am fine with changing the order towards FCFS :)

src/compiler/crystal/semantic/flags.cr
mneumann: We can't make the changes to use ```flag?(:bsd)``` yet. That does not work, it breaks at least the FreeBSD build! We first have to add the ```flags.add "bsd" ...``` code, then bootstrap the compiler with it, and only the bootstrapped compiler is able to build code that contains ```flag?(:bsd)``` code. If we add it now, it will break all BSD platforms :(. I suggest adding the bsd flag in semantic/flags.cr, but revert the parts of commit 1181ab4 that use flag bsd.

mneumann: This wasn't caught by CI because there is no FreeBSD or OpenBSD CI.

mneumann:
> > Did you run into this on NetBSD?
> 
> We're not sure, but it's definitely something to look into, thank you.

FYI: I can reproduce a stack overflow on FreeBSD when I set the stack limit to about 4650 kbytes (ulimit -s 4650). A limit of 4700 is just fine, crystal compiles. Interestingly, on FreeBSD I get a rather short stack trace, vs. hundreds of lines on DragonFly.

straight-shoota:
Dankeschön @mneumann

PR #9180, opened 2020-04-26
HTTP::Status docs

wontruefree:


src/http/status.cr
```diff
@@ -70,41 +70,102 @@ enum HTTP::Status
...
+  # HTTP::Status.new(123)  # => 123
+  # HTTP::Status.new(1000) # raises ArgumentError
+  # ```
   def self.new(status_code : Int32)
     raise ArgumentError.new("Invalid HTTP status code: #{status_code}") unless 100 <= status_code <= 999
     previous_def(status_code)
   end
 
-  def code
+  # Returns the number that repreents the HTTP status code.
```
jhass, line 87:
```suggestion
  # Returns the number that represents the HTTP status code.
```

Or maybe going beyond fixing the typo:

```suggestion
  # Returns the code that represents the HTTP status.
```
Or even just
```suggestion
  # Returns the HTTP status code.
```

wontruefree:
Thanks @jhass for the suggestion. Could you take another look.

wontruefree:
@jhass I agree there could be better word choices, but I wanted to distinguish that API has an Int32 return type. But the Types will be in the docs anyways.

What is written I think gets across the most information, but if you have a strong opinion, I could change it.

PR #9185, opened 2020-04-27
IO#skip returns the number of bytes it skipped

carlhoerberg:
Can make some code much better looking, eg:

```crystal
len = Int32.from_io @file
@file.skip(len)
skipped += len
``` 
can become
```crystal
skipped += @file.skip(Int32.from_io(@file))
```

### DIFF
--- a/src/io.cr
+++ b/src/io.cr
@@ -814,22 +814,26 @@ abstract class IO
   # io.gets    # => "world"
   # io.skip(1) # raises IO::EOFError
   # ```
-  def skip(bytes_count : Int) : Nil
+  def skip(bytes_count : Int) : Int
+    remaining = bytes_count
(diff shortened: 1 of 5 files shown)


carlhoerberg:
Would also suggest increasing the buffer to 8kb as done in `IO.copy`. Can add it here or create another PR.

straight-shoota:
Looks good, but you'll need to add specs for this.
Buffer size should be a separate issue.

RX14:
For `#skip` I don't see the point because it's guaranteed to be the amount specified. For `#skip_to_end` it's definitely useful.

carlhoerberg:
Correct, but it allows for shorter user code, see  example above. Do have a reason why not to return it?

PR #9191, opened 2020-04-27
Parser: fix parsing of named args after call named the same as a var

### LINKED ISSUE #9189: Incorrect parsing of variable or call
Considering this snipped of code:
``` crystal
def m(a = 0)
  args
end

m = 10

puts m a: 20
```
What is printed on the console is `10` and not `20`. The code can be found [here](https://play.crystal-lang.org/#/r/8yt3/edit).

The problem seems to be in `Parser#parse_var_or_call` in the condition `maybe_var && args.size == 0` that does not take into account named arguments array, causing it to be dropped even though correctly parsed.

I don't know if in this snippet `m` is wanted to be a call or a variable, but in this last case I guess a syntax error should be raised.

asterite:
Fixes #9189

### DIFF
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -415,6 +415,20 @@ module Crystal
       Call.new(nil, "a", args: [-1.int32] of ASTNode, named_args: [NamedArgument.new("b", 2.int32)]),
     ]
 
+    it_parses "a = 1; c a b: 2", [
+      Assign.new("a".var, 1.int32),
+      Call.new(nil, "c", args: [
+        Call.new(nil, "a", named_args: [NamedArgument.new("b", 2.int32)]),
+      ] of ASTNode),
+    ]
+
+    it_parses "a = 1; c a -1, b: 2", [
+      Assign.new("a".var, 1.int32),
+      Call.new(nil, "c", args: [
+        Call.new(nil, "a", args: [-1.int32] of ASTNode, named_args: [NamedArgument.new("b", 2.int32)]),
+      ] of ASTNode),
+    ]
+
     it_parses %(foo("foo bar": 1, "baz": 2)), Call.new(nil, "foo", named_args: [NamedArgument.new("foo bar", 1.int32), NamedArgument.new("baz", 2.int32)])
     it_parses %(foo "foo bar": 1, "baz": 2), Call.new(nil, "foo", named_args: [NamedArgument.new("foo bar", 1.int32), NamedArgument.new("baz", 2.int32)])
 
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -405,6 +405,16 @@ module Crystal
     it_parses "foo(a: 1\n)", Call.new(nil, "foo", named_args: [NamedArgument.new("a", 1.int32)])
     it_parses "foo(\na: 1,\n)", Call.new(nil, "foo", named_args: [NamedArgument.new("a", 1.int32)])
 
+    it_parses "a = 1; a b: 2", [
+      Assign.new("a".var, 1.int32),
+      Call.new(nil, "a", named_args: [NamedArgument.new("b", 2.int32)]),
+    ]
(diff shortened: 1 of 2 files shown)


spec/compiler/parser/parser_spec.cr
```diff
@@ -405,6 +405,16 @@ module Crystal
...
 
+    it_parses "a = 1; a b: 2", [
+      Assign.new("a".var, 1.int32),
+      Call.new(nil, "a", named_args: [NamedArgument.new("b", 2.int32)]),
+    ]
+
+    it_parses "a = 1; a -1, b: 2", [
+      Assign.new("a".var, 1.int32),
+      Call.new(nil, "a", args: [-1.int32] of ASTNode, named_args: [NamedArgument.new("b", 2.int32)]),
+    ]
```
jhass, line 416: Can we have one for `c a b: 2` and one for `c a -1, b: 2`, with `c` being a method, too? :)

asterite: Done! The specs pass without any change.

straight-shoota:
This might be a bit cleaner solution. `CallArgs#args` doesn't need to be nilable. The array of `ASTNode` is going to be initialized anyway, so it could just be done in the single branch where `nil` was used:
```diff
--- i/src/compiler/crystal/syntax/parser.cr
+++ w/src/compiler/crystal/syntax/parser.cr
@@ -4057,9 +4057,9 @@ module Crystal
         else
           if args
             maybe_var = !force_call && is_var && !has_parentheses
-            if maybe_var && args.size == 0
+            if maybe_var && args.size == 0 && !named_args
               Var.new(name)
-            elsif maybe_var && args.size == 1 && (num = args[0]) && (num.is_a?(NumberLiteral) && num.has_sign?)
+            elsif maybe_var && !named_args && args.size == 1 && (num = args[0]) && (num.is_a?(NumberLiteral) && num.has_sign?)
               sign = num.value[0].to_s
               num.value = num.value.byte_slice(1)
               Call.new(Var.new(name), sign, args)
@@ -4262,7 +4262,7 @@ module Crystal
     end

     record CallArgs,
-      args : Array(ASTNode)?,
+      args : Array(ASTNode),
       block : Block?,
       block_arg : ASTNode?,
       named_args : Array(NamedArgument)?,
@@ -4328,7 +4328,7 @@ module Crystal
         next_token

         if stop_on_do_after_space && @token.keyword?(:do)
-          return CallArgs.new nil, nil, nil, nil, true, end_location, has_parentheses: false
+          return CallArgs.new [] of ASTNode, nil, nil, nil, true, end_location, has_parentheses: false
         end

         if control && @token.keyword?(:do)
```

asterite:
@straight-shoota the args are not needed in case a `Var` is produced. That's why it's nilable.

Also, we can talk about refactors on a separate PR, this is about fixing the existing issue.

straight-shoota:
The only case where it was assigned to be `nil` was the branch with `@token.keyword?(:do)` which can't turn out to be a `Var`. This instance has `stopped_on_do_after_space: true` and thus is already handled in the block branch.
I see this just as an alternate way to fix the issue by removing the ambiguity between `args.nil?` and `args.size == 0`. IMO that's better than introducing a workaround on that ambiguity. But I'm fine with cleaning up on a follow-up.

max-codeware:
I think there is still an issue: this line of code
``` crystal
elsif maybe_var && !named_args && (args && args.size == 1) && (num = args[0]) && (num.is_a?(NumberLiteral) && num.has_sign?)
```
Takes into account only this case
``` crystal
def m(x)
  x
end

m = 10
puts( m -2)  #=> 8
```
and not this case that is the same IMO
``` crystal
m = 10
b = 2
puts(m -b) #=> -2
```
In Ruby both the syntaxes produce the same result, no matter the ambiguity coming from the call `puts`, the code is iterpreted as `m.-(b)`

asterite:
@max-codeware maybe, but it's a separate issue. Could you report it? If it turns out to be a bug, I'd like to fix it in a separate PR.

PR #9194, opened 2020-04-28
Parser: fix parsing of ambiguous '+' and '-'

### LINKED ISSUE #9193: Not unified parsing of ambiguous suffixes `+` and `-`
As asked by @asterite I report here what I wrote in PR #9191.
Supposing to have this code:
``` crystal
def m(x)
  x
end

m = 10

puts(m -2) #=> 8
``` 
Crystal interprets the ambiguous suffix as `m.-(2)` giving the precedence to `m` as variable.

But if we wrote the same code in this way
``` crystal
def m(x)
  x
end

m = 10
b = 2

puts(m -b) #=> -2
```
what saw previously is no longer valid, and `m` is interpreted as a call.
IMO both the syntaxes should produce the same result like in Ruby, as I don't see why `-b` should be treated differently than `-2`. 

The same thing is valid for ambiguous `+`

max-codeware:
This should fix #9193


straight-shoota:
Could you also add specs with your examples from #9193 to make sure it works as expected when there's also a method of that name.

max-codeware:
Of course. Is that a problem if the check_format fails?

straight-shoota:
Yeah. Just run `crystal tool format`.

src/compiler/crystal/syntax/parser.cr
```diff
@@ -4063,6 +4063,9 @@ module Crystal
               sign = num.value[0].to_s
               num.value = num.value.byte_slice(1)
               Call.new(Var.new(name), sign, args)
+            elsif maybe_var && args.size == 1 && (arg = args[0]) && arg.is_a?(Call) && !arg.obj.nil? &&
+                  ({"+", "-"}.includes? arg.name) && (!arg.args || arg.args.size == 0)
```
Sija, line 4067: `Object#in?` will read more naturally here:

```suggestion
                  (arg.name.in?("+", "-") && (!arg.args || arg.args.size == 0)
```

straight-shoota:
Grazie tante, @max-codeware

Issue #8809, opened 2020-02-16
Docs won't scroll to methods that have `"`


Blacksmoke16:
1. https://crystal-lang.org/api/master/File.html
2. Click on either `.new`/`.open` constructors
3. Notice you don't get taken to the detail section of those methods

Seems to be caused by escaping the `"` for the `id` of the element. `mode=&quot;r&quot;,` doesn't match the anchor value of `mode="r"`.

Blacksmoke16:
Related to this; is https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/tools/doc/method.cr#L221-L269 really necessary?

I know its intention is to be able to link to specific overloads of the same type.  However, would it not be easier to just use the method object's `#hash` value as the anchor?  This would be a bit more URI safe character wise, shorter, and a simpler implementation.  But on the downside, you would lose some "metadata" from the URI.  Given the purpose of it is just to link to documentation its probably not a big deal to lose that.

Blacksmoke16:
Actually yes we need that.  Wouldn't want the links to break every time a new version is released as they might have different hash values.

Blacksmoke16:
Fixed by #9200

PR #3623, opened 2016-12-02
Select anchor from location.hash on page load

Sija:
This PR fixes scrolling to section given in URI hash (like in https://crystal-lang.org/api/0.20.0/Array.html#clear-instance-method). This works around bug which resets container scroll on page refresh.

PR #9200, opened 2020-04-29
Docs: fix links to methods with String default values

bcardiff:
The double-quotes value in the id need to be escaped with &quot; in order to work.

For example https://crystal-lang.org/api/0.34.0/Path.html#new(name:String=%22%22):Path-class-method is not working without this PR.

### DIFF
--- a/src/compiler/crystal/tools/doc/html/js/doc.js
+++ b/src/compiler/crystal/tools/doc/html/js/doc.js
@@ -186,14 +186,12 @@ document.addEventListener('DOMContentLoaded', function() {
   var scrollToEntryFromLocationHash = function() {
     var hash = window.location.hash;
     if (hash) {
-      var targetAnchor = unescape(hash.substr(1));
-      var targetEl = document.querySelectorAll('.entry-detail[id="' + targetAnchor.replace(/"/g, '&quot;') + '"]')
-
-      if (targetEl && targetEl.length > 0) {
-        targetEl[0].offsetParent.scrollTop = targetEl[0].offsetTop;
+      var targetAnchor = decodeURI(hash.substr(1));
+      var targetEl = document.getElementById(targetAnchor)
+      if (targetEl) {
+        targetEl.offsetParent.scrollTop = targetEl.offsetTop;
       }
     }
   };
-  window.addEventListener("hashchange", scrollToEntryFromLocationHash, false);
   scrollToEntryFromLocationHash();
 });
--- a/src/compiler/crystal/tools/doc/html/js/doc.js
+++ b/src/compiler/crystal/tools/doc/html/js/doc.js
@@ -193,5 +193,6 @@ document.addEventListener('DOMContentLoaded', function() {
       }
     }
   };
+  window.addEventListener("hashchange", scrollToEntryFromLocationHash, false);
   scrollToEntryFromLocationHash();
 });
--- a/src/compiler/crystal/tools/doc/html/js/doc.js
+++ b/src/compiler/crystal/tools/doc/html/js/doc.js
@@ -187,7 +187,7 @@ document.addEventListener('DOMContentLoaded', function() {
     var hash = window.location.hash;
     if (hash) {
       var targetAnchor = unescape(hash.substr(1));
-      var targetEl = document.querySelectorAll('.entry-detail[id="' + targetAnchor + '"]');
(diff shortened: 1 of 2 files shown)


Blacksmoke16:
Fixes #8809.

jhass:
Wait, why is there any JS  for this? Browsers should support scrolling to an element with the right `id` natively, as long as we keep consistency in generating links and `id=` tags, it should just work, no?

bcardiff:
I would expect that also @jhass . I'm not sure how what other JS are required in the documentation html to expand/hightlight methods.

Maybe @straight-shoota is more aware.

straight-shoota:
Seems like this was introduced in #3623 to fix an issue with scroll position being reset when reloading the page. Affecting every fragment link action seems quite intrusive, though.
/cc @Sija

straight-shoota:
I actually can't notice any difference in the user experience when this method is removed. #3623 lacks a detailled explanation what exactly goes wrong, so I might be missing something.

Sija:
@straight-shoota That was ~3,5yrs ago, I lot could have changed since then (and did), so I'd suggest trying out whether this workaround is still needed and applicable these days.

straight-shoota:
Yeah, but I'm not even sure what *exactly* this workaround fixed. Can you recall any specific behaviour we can test?

Sija:
@straight-shoota It was related to scrolling - as mentioned in the original PR - this should be enough to  let you test it - both ff and chrome woud be needed as their behavior differs in such matters.

straight-shoota:
Okay so I found the faulty behaviour. When you reload the page in Chrome it always scrolls the main content pane to the top, even if the URL points to an anchor. This seems to be related to the script responsible for the sidebar to scroll the currently selected entry into view. That should just affect the sidebar. But Chrome seems to disagree.

But the fix to this problem should be simpler than what we currently do.
```js
var elem = document.getElementById(window.location.hash.substr(1))
if (elem) elem.scrollIntoView();
```
This only needs to be executed once, not on every hashchange event. Klicking anchor links on the page should not be affected. It's only relevant when the page is loaded.

bcardiff:
It seems unrelated to this change then.

For ids with quotes  the following is still needed:

```js
document.getElementById(unescape(window.location.hash.substr(1)).replace(/"/g, '&quot;'))
```

Another approach would be to use safe chars for ids.

Either way, a story for another PR it seems.

straight-shoota:
No, this is the wrong change. It just fixes a symptom in our JavaScript code. Instead we should fix the cause, that is the anchor names and link targets must work even when special characters are involved.

straight-shoota:
It's not even enough because there are other characters that cause issues (for example `|`).
This is the fragment for `File.open` that Firefox produces when clicking on the link in the constructors section:
```
#open(filename:Path%7CString,mode="r",perm=DEFAULT_CREATE_PERMISSIONS,encoding=nil,invalid=nil):self-class-method
```
This fragment actually works
```
#open(filename:Path|String,mode=&quot;r&quot;,perm=DEFAULT_CREATE_PERMISSIONS,encoding=nil,invalid=nil):self-class-method
```

bcardiff:
But is wrong because we are mixing how `"` need to be escaped in uri and html.

The ideal is to use safe chars for ids.

straight-shoota:
> The ideal is to use safe chars for ids.

Maybe. But the characters arenot really the problem. They actually work fine:

```html
<a href="#&quot;|">goto</a>
<a id="&quot;|">target</a>
```
So the error is on our side how we process the anchor values. For example, id values must not be URI escaped.

bcardiff:
Ok, drilled down a little bit, the issue was that the arguments are escaped twice.
I removed the query selector in favor of getElementById
I removed the hashchange event
But I didn't add the `scrollIntoView` because the note at https://caniuse.com/#search=scrollIntoView

Sija:
> I removed the hashchange event

Did you tested did the current version works with the _Back_ browser button - i.e. correctly scrolls to the given _id_?

bcardiff:
Now I did @Sija . Thanks 🙈

straight-shoota:
We need to make add a comment to explain why this code is there and how it works.

bcardiff:
> We need to make add a comment to explain why this code is there and how it works.

Definitely, but I only know that is need. I haven't tracked why/how it works.

Issue #8914, opened 2020-03-18
Add top-level has_constant? macro method (or expose Program type)


bcardiff:
In #1585 `has_constant?` was added and allows you to check if a constant is defined within a type. But I think there is no hay to check if a top-level constant exists.

Adding a top-level macro method to check if the constant is defined in the internal `Program` type that holds the top level could look like:

```crystal
{{ has_constant?(:MyShard) }}
```

Or if we expose the top level as an alias we could directly use.

```crystal
{{ Program.has_constant?(:MyShard) }}
```

It can be useful to check if a shard is required or not. And to use feature checks instead of comparing versions.

Blacksmoke16:
I think the latter would be better, as this is also an issue with like "getting methods on the top level", etc.

asterite:
Actually, you can do this right now :-D

```crystal
{% puts @type.has_constant?(:Int32) %}
{% puts @type.has_constant?(:Int256) %}
```

https://play.crystal-lang.org/#/r/8qfs

`@type` at the top-level refers to the program.

That said, it's true that there's no way to explicitly refer to the top level...

But maybe we can do `@type.top_level` to access the top level from any type? That way we don't need to replicate all of the methods on `TypeNode` on the top level.

Blacksmoke16:
Ahhh yes, the issue was more so having access to top level constants/methods when the macro itself is _NOT_ in the top level as well.

asterite:
Then I still think it makes sense to expose the top level as `@type.top_level`. `Program` _could_ work but it can be quite confusing if there's an actual type called `Program`.

jhass:
How about another "macro instance var" like `@toplevel`? having it as a method on `@type` somehow irks me since it's not really a property of a type node. And indeed `Program` isn't very explicit about what it is either.

asterite:
Yeah, another macro instance var sounds good too.

straight-shoota:
`@program`?

RX14:
Why `Program`? Seems strange to tie the name to a badly named god object in the compiler. Top-level is what we call it everywhere else.

bcardiff:
Program is probably a leak from the compiler internals. `@top_level` sounds about right.

bcardiff:
The main benefit of `@top_level` vs `::has_constant?` is that this will allow other methods defined in `TypeNode` to be applied to the top level.

RX14:
[...]

PR #9201, opened 2020-04-29
Log: Add Log.capture spec helper

bcardiff:
Adds `Log.capture` method (when `require "log/spec"`) that plays nice with spec to assert log entries are emitted.

This is somewhat influenced in FileCheck that is able to express valid traces. You will be able to validate the order of the entries emitted while been more strict or loose whether there can be more entries that the one that are stated in the test. Forcing the order semantics allows some nicer error messages:

* X was expected, but the next message was Y
* X was expected, but I couldn't find it, yet N entries were skipped
* etc

I rewrite the recent http.server specs that introduced a more limited helper function and served as a motivation for this change.

There are no others `*/spec` modules in the std-lib. But is something I also did in crystal-db to add some reusable helper methods for drivers development. I like that convention.

Regarding the implementation I needed to extend a bit the `Log::Builder` capabilities but left them as `:nodoc:` for reasons explained in the commit.

With this PR libraries can use `Log.capture` and there is no need to use `IO::MemoryBackend` explicitly for testing as long as the `Log` instances are created with the `Log.for` method.


spec/support/log.cr
```diff
@@ -1,33 +1,9 @@
...
-  begin
-    yield mem_backend.entries
-    mem_backend.entries
-  ensure
-    log.backend = old_backend
-    log.level = old_level
-  end
+# Turn off logging in specs
+def log_setup_on_specs
+  Log.setup(:none)
```
waj, line 6: Should this file be removed completely? Maybe when `"log/spec"` is required it might disable all the logs by default.

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+  #
+  # `EntriesChecker#entry` returns the last matched `Entry`. Useful to check additional entry properties other than the message.
+  #
+  # `EntriesChecker#empty` validates there are no pending entries to match.
+  #
+  # Using the yielded `EntriesChecker` allows clearing the entries between statements.
+  #
+  # Invocations can be nested in order to capture each source in their own `EntriesChecker`.
+  #
+  def self.capture(source : String = "*", level : Severity = :info, *, builder = Log.builder)
```
waj, line 55: Can we capture all the logs by default? (`:trace`)

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+
+    # Clears the emitted entries so far
+    abstract def clear
+
+    # Validates that there are no outstanding entries to be validated
+    abstract def empty(file = __FILE__, line = __LINE__)
+  end
+
+  # :nodoc:
+  class ArrayEntriesChecker
```
waj, line 114: This is `:nodoc:` but also explicitly used in the tests, so it's useful. Should this be called `EntriesChecker` instead and drop the module? Or do you have other implementations in mind?

spec/support/log.cr
bcardiff: There is one more call of `log_setup_on_specs` in the `Log.setup` specs that is needed to restore the "tuning off" of the logs.

I wasn't sure if doing a `Log.setup(:none)` in `"log/spec"` was right. If some module emit logs and `"log/spec"` is not required, the logs will be shown in stdout.

If we want to turn off logging in specs by default I would suggest doing that in `"spec"` module directly. Checking if `Log` is defined to avoid requiring `"log"` in the first place.

src/log/spec.cr
straight-shoota: Yeah, I don't see a reason to separate interface and implementation here either.

src/log/spec.cr
bcardiff: I wanted to preserve the same default as in `Log.setup`, but I was doubting whether to use `:info` or `:trace` as a default.

spec/support/log.cr
waj: I see, maybe `spec` should set some flag or something so logs can be disabled by default when testing?

src/log/spec.cr
bcardiff: The only need for it was in specs that used `MemoryBackend` explicitly.

We can go that way and make it public. I just wanted to minimize the exposed interface.

The alternative implementations I can think of `ArrayEntriesChecker` is for other testing frameworks. Yet the extensibility is not ready yet.

bcardiff:
Feedback addressed.

I needed some macro to workaround #8914 and turn off logging by default on specs without forcing a `require "log"`.

```terminal-session
$ ./bin/crystal spec spec/std/http/server/request_processor_spec.cr  # no logs
$ LOG_LEVEL=trace ./bin/crystal spec spec/std/http/server/request_processor_spec.cr  # show logs in stdout also
```

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+  #
+  # By default logs of all sources and with severity `Severity::Info` or above will be captured.
+  #
+  # Use *level* to only capture of the given severity or above.
+  #
+  # Use *source* to narrow which source are captured. Values that represent single pattern like `http.*` are allowed.
+  #
+  # The `EntriesChecker` will hold a list of emitted entries.
+  #
+  # `EntriesChecker#check`  will look for the first entry that matches the level and message.
```
RX14, line 41: Perhaps "will find the next entry which matches"?

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+  # `EntriesChecker#next` will validate if the following entry in the list matches the given level and message.
```
RX14, line 42: "will validate that the following entry"

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+  # `EntriesChecker#clear` will clear the emitted and captured entries.
+  #
+  # With these methods it is possible to express expected traces in a strict and loosely way, yet always ordered.
```
RX14, line 45: "either a strict or loose way, while checking ordering"

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+  class ArrayEntriesChecker
+    include Log::EntriesChecker
+
+    def initialize(@entries : Array(Log::Entry))
+    end
+
+    getter! entry : Entry
+
+    def check(description, file = __FILE__, line = __LINE__, & : Entry -> Bool) : self
+      fail("No entries found expected #{description}", file, line) if @entries.empty?
```
RX14, line 85: "No entries found, expected"

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+
+      while entry = @entries.shift?
+        matches = yield entry
+        if matches
+          @entry = entry
+          return self
+        end
+      end
+
+      fail("No matching entries found expected #{description}, skipped (#{original_size})", file, line)
```
RX14, line 96: ditto

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+    end
+
+    def next(description, file = __FILE__, line = __LINE__, & : Entry -> Bool) : self
+      if entry = @entries.shift?
+        matches = yield entry
+        if matches
+          @entry = entry
+          return self
+        else
+          fail("No matching entries found expected #{description}, but got #{entry.severity} with #{entry.message.inspect}", file, line)
```
RX14, line 117: ditto

src/log/spec.cr
```diff
@@ -0,0 +1,166 @@
...
+        end
+      else
+        fail("No entries found expected #{description}", file, line)
```
RX14, line 120: ditto

spec/std/log/spec_spec.cr
```diff
@@ -0,0 +1,154 @@
+require "spec"
+require "log"
+require "log/spec"
+
+Log.setup(:none)
+
+describe "log/spec" do
```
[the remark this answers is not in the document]
bcardiff: There is, but I will split in it's own.

bcardiff:
Feedback addressed and I notice some loggers were still emitting to stdout because the level is manually overridden in some spec and the same Log was used in other places. Too much `foo`.

paulcsmith:
This is insanely cool and useful. I added a spec helper to Dexter for something similar, but this is way better. I'm excited to remove something from Dexter and use this instead ❤️ 🎉

PR #9202, opened 2020-04-29
Fix for ambiguous call args parsing with '+' and '-' ambiguous prefix

max-codeware:
This should fix at least one case of the bug family #9189 and is meant to work with PR #9191.
The addressed issue is mainly this:
``` crystal
def m(x, y)
  ...
end

m = 10

puts m -1, 3
```
that, from what I understood, should be interpreted as `puts(m - 1, 3)` and not as `puts m(-1, 3)`.
This also partly fixes `puts m -1, y: 3` in which named args were silently swallowed during parsing.
All the other cases should remain unaffected.


spec/compiler/parser/parser_spec.cr
```diff
@@ -1849,6 +1849,14 @@ module Crystal
...
+      it_parses "def foo(x,y);end;foo = 10;puts(foo +a, b)", [Def.new("foo", ["x".arg, "y".arg] of Arg), Assign.new("foo".var, 10.int32), Call.new(nil, "puts", [Call.new("foo".var, "+", "b".var), "b".var] of ASTNode)]
+      it_parses "def foo(x,y);end;foo = 10; puts foo a, b", [Def.new("foo", ["x".arg, "y".arg] of Arg), Assign.new("foo".var, 10.int32), Call.new(nil, "puts", Call.new(nil, "foo", ["a".var, "b".var] of ASTNode))]
+      it_parses "def foo(x,y);end;foo = 10; puts(foo a, b)", [Def.new("foo", ["x".arg, "y".arg] of Arg), Assign.new("foo".var, 10.int32), Call.new(nil, "puts", Call.new(nil, "foo", ["a".var, "b".var] of ASTNode))]
+      it_parses "def foo(x,y);end;foo = 10; puts foo -a, b: 2", [Def.new("foo", ["x".arg, "y".arg] of Arg), Assign.new("foo".var, 10.int32), Call.new(nil, "puts", [Call.new("foo".var, "-", "a".var), NamedArgument.new("b", 2.int32)] of ASTNode)]
+      it_parses "def foo(x,y);end; puts foo -a, b", [Def.new("foo", ["x".arg, "y".arg] of Arg), Call.new(nil, "puts", Call.new(nil, "foo", ["a".var, "b".var] of ASTNode))]
```
asterite, line 1858: There's no need to define the `def`s here, they don't affect the parsing.

max-codeware: I know, I just kept following the suggestion I got in #9194. But I can remove it if you want

max-codeware:
By the way, the check_format seems to fail, but running it on my pc seems to produce no change...

spec/compiler/parser/parser_spec.cr
asterite: Oh, but that suggestion doesn't make sense because the parser doesn't care about methods defined. It would be pretty catastrophic if parsing depended on this. So there's no need at all do it.

straight-shoota (approved):

src/compiler/crystal/syntax/parser.cr
```diff
@@ -70,6 +70,20 @@ module Crystal
       #
       # then this flag is set to `true` when parsing `foo`'s arguments.
       @stop_on_do = false
+
+      # This flag tells the parser to stop on comma while parsing call
+      # when there is an ambiguous suffix `+` or `-`
+      # ```plain
```
Sija, line 76:
```suggestion
      #
      # ```
```

src/compiler/crystal/syntax/parser.cr
```diff
@@ -70,6 +70,20 @@ module Crystal
...
+      # def m(x, y)
+      #   ...
+      # end
+      #
+      # m = 10
+      # puts m -2, 3
+      # ```
```
Sija, line 83:
```suggestion
      # ```
      #
```

src/compiler/crystal/syntax/parser.cr
straight-shoota: This is intentional. Without language tag the formatter would apply and render the example useless.

PR #9204, opened 2020-04-29
Add titleize method to String.

hugopl:
If you guys like the idea I can add tests to this PR.

This is similar `titleize` method Rails[1] add to Ruby strings, however with a difference:

Rails replaces the `-` by spaces from words, but this doesn't work on all languages, e.g. in portuguese `"fala-se muito"` should be `"Fala-se Muito"`, not `"Fala Se Muito"` as rails does.

So this titleize just upcase letters after whitespaces, similar to already existing `capitalize` method.

[1] https://apidock.com/rails/ActiveSupport/Inflector/titleize

### DIFF
--- a/spec/std/string_spec.cr
+++ b/spec/std/string_spec.cr
@@ -563,6 +563,16 @@ describe "String" do
     it { "iO".capitalize(Unicode::CaseOptions::Turkic).should eq("İo") }
   end
 
+  describe "titleize" do
+    it { "hEllO tAb\tworld".titleize.should eq("Hello Tab\tWorld") }
+    it { "  spaces before".titleize.should eq("  Spaces Before") }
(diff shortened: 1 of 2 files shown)


j8r:
Missing specs (you can take the examples in the method description).

hugopl:
> Besides adding some specs, move it nearer the capitalize, upcase, downcase methods.

:+1:

straight-shoota:
Maybe this should preserve upper case characters? This would be useful for abbreviations. The current behaviour can be achieved with `downcase.titelize`.

hugopl:
> Maybe this should preserve upper case characters? This would be useful for abbreviations. The current behaviour can be achieved with `downcase.titelize`.

Preserve case makes sense to me, but I'm really not sure about what would be better. `capitalize` don't do this and downcase everything but first letter. I would like to listen what other reviewers think about it.

PR #9206, opened 2020-04-29
Full inspect for ASTNode

### LINKED ISSUE #9205: Unclear spec error message as output
I am trying to fix the parser spec for a modification I made, but I get this output as error
```
Parser parses "b = 0;puts foo -1, b"
     Failure/Error: node.should eq(Expressions.from(local_expected_node))

       Expected: b = 0
       puts(foo(-1, b))
        : Crystal::Expressions
            got: b = 0
       puts(foo(-1, b))
        : Crystal::Expressions

     # spec/compiler/parser/parser_spec.cr:21
```

I don't know if there is any other issue open like this, as I don't know how to look for it, and I also guess it's just a matter of how the ast node is constructed, but what can I do to know what's wrong?

max-codeware:
Should fix #9205


src/compiler/crystal/syntax/to_s.cr
```diff
@@ -3,9 +3,9 @@ require "./visitor"
 
 module Crystal
   class ASTNode
-    def inspect(io : IO) : Nil
-      to_s(io)
-    end
+    # def inspect(io : IO) : Nil
+    #   to_s(io)
+    # end
```
asterite, line 8: Can we simply remove these lines? 🙏

max-codeware: What would happen in this case?
```
macro node(n)
  {{puts n}}
end

node a + b
```

It is not a so important feature, but sometimes I use it to debug

asterite: I don't think `inspect` is used in that case?

max-codeware: All right, makes sense.

The spec seem to fail, but I have no idea about the reason

straight-shoota (approved):

straight-shoota:
Actually this isn't a good idea. `inspect` now prints the entire `program` because that's a property of every ASTNode. We'll need customized output for this.

asterite:
Well, not every ast nodes, just some

PR #9207, opened 2020-04-30
Set `sync` or `flush_on_newline` for standard I/O on Windows.

kubo:
This PR sets `sync` or `flush_on_newline` for standard I/O on Windows as on Unix.
This uses [GetConsoleMode][], which succeeds only when the handle is to the console.

I tested it by the code in https://github.com/crystal-lang/crystal/pull/9149#issue-406684705. Without this PR, the benchmark results are displayed on process exit. With this PR, they are displayed as on Unix.

I want to hear opinions. Is it better to move win32 and unix code in `IO::FileDescriptor.from_stdio` to `crystal/system/win32/file_descriptor.cr` and `crystal/system/unix/file_descriptor.cr` respectively?

[GetConsoleMode]: https://docs.microsoft.com/en-us/windows/console/getconsolemode

### DIFF
--- a/src/crystal/system/win32/file_descriptor.cr
+++ b/src/crystal/system/win32/file_descriptor.cr
@@ -162,4 +162,8 @@ module Crystal::System::FileDescriptor
 
     bytes_read
   end
+
+  def self.from_stdio(fd)
+    IO::FileDescriptor.new(fd)
+  end
 end
--- a/src/crystal/system/win32/file_descriptor.cr
+++ b/src/crystal/system/win32/file_descriptor.cr
@@ -1,4 +1,5 @@
 require "c/io"
+require "c/consoleapi"
 
 module Crystal::System::FileDescriptor
   @volatile_fd : Atomic(LibC::Int)
@@ -164,6 +165,22 @@ module Crystal::System::FileDescriptor
   end
 
   def self.from_stdio(fd)
-    IO::FileDescriptor.new(fd)
+    console_handle = false
+    handle = LibC._get_osfhandle(fd)
+    if handle != -1
+      if LibC.GetConsoleMode(LibC::HANDLE.new(handle), out _) != 0
+        console_handle = true
+      end
+    end
+
+    io = IO::FileDescriptor.new(fd)
+    # Set sync or flush_on_newline as described in STDOUT and STDERR docs.
+    # See https://crystal-lang.org/api/toplevel.html#STDERR
+    if console_handle
+      io.sync = true
+    else
+      io.flush_on_newline = true
+    end
+    io
   end
 end
--- a/src/crystal/system/unix/file_descriptor.cr
+++ b/src/crystal/system/unix/file_descriptor.cr
@@ -161,4 +161,24 @@ module Crystal::System::FileDescriptor
 
     bytes_read
   end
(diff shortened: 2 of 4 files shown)


oprypin:
Yes, it is better to move the code as you described.

src/crystal/system/win32/file_descriptor.cr
```diff
@@ -162,4 +163,22 @@ module Crystal::System::FileDescriptor
...
+    handle = LibC._get_osfhandle(fd)
+    if handle != -1
+      if LibC.GetConsoleMode(LibC::HANDLE.new(handle), out _) != 0
+        console_handle = true
+      end
+    end
+
+    io = IO::FileDescriptor.new(fd)
+    if console_handle
+      io.sync = true
```
oprypin, line 178: Wait, isn't this inverted? console should be flush_on_newline, otherwise sync

kubo: https://github.com/crystal-lang/crystal/blob/0.34.0/src/kernel.cr#L15-L20
```
# - if it's a TTY device (like the console) then `sync` is `true`,
#   meaning that output will be outputted as soon as it is written
#   to STDOUT. This is so users can see real time output data.
# - if it's not a TTY device (like a file, or because the output
#   was piped to a file) then `sync` is `false` but `flush_on_newline`
#   is `true`. This is so that if you pipe the output to a file, and,
```

straight-shoota: Maybe you could add a reference to this as a doc comment like in the unix implementation?

kubo:
Add a comment as requested by @straight-shoota [here](https://github.com/crystal-lang/crystal/pull/9207#discussion_r418496768).
Others were not changed.

oprypin:
This should be merged as separate commits btw.

kubo:
> This should be merged as separate commits btw.

Could you explain more specific? I cannot understand your request.

oprypin:
Sorry for confusion :(
This is not a request for you.
I'm saying that, because you kept this PR as 2 separate and tidy commits, the person who will be merging this should not automatically squash it into one commit.

RX14:
Sorry, we can't do that any more, @waj decided everything had to be squashed.

waj:
Yeah, how lovely...
I just re-enabled merges. Enjoy!

asterite:
Please let's not say X decided something. Many of us agree that squashing is fine because you can always go to the PR to see the commit history.

@oprypin do you think it would be fine to squash this anyway? In my experience whenever I ended up using blame and in a single commit, I navigated to the PR anyway to understand the overall context.

oprypin:
I would be fine with **always** squashing, and announcing that that is the case, then people won't have any reason to force-push into the PR branch. Otherwise we're getting the worst of both worlds.

RX14:
I've had an idea for a solution which would make everyone happy for a few years, but unfortunately it requires writing code and starting a new project is hard to justify...

I apologise for being ratty this morning, I think I might need to take a break from Crystal...

asterite:
I think that's another benefit of squashing: PR authors don't need to care about keeping a clean commit history. That should speed up contributions.

PR #9211, opened 2020-04-30
Refactor Log::Formatter

waj:
This replaces the `Log::Formatter` with a module with two basic implementations:

  * `ProcFormatter`: used to allow formatting with a proc, just like it's right now
  * `StaticFormatter` inspired in the date/time formatters with the addition of a macro to generate subclasses

`Log::Formatter.new { |entry, io| }` returns a `ProcFormatter` and it can be used when the `StaticFormatter` is not enough.

The macro `Log.formatter` can be used to generate `StaticFormatter` subclasses from string patterns, like this:

```crystal
Log.formatter MyFormatter, "[#{timestamp}] #{severity}: #{message}"
```

That will create a `MyFormatter` that will format entries like this:
```
[2020-04-30T23:33:09.363434000Z]    INFO: Message goes here!
```

Also, I replaced the default formatter of the `IOBackend` with one that generates less output by default. We could add other built-in formatters with common patterns.


straight-shoota:
I'm not sure about this. I didn't have time yet to actually work with the new Log framework to get some experience on whether such an extension is even necessary. It also doesn't seem super important to do and it's not even a substantial change. As far as I can see, it's not even breaking (at least it doesn't need to be).

In any case I would appreciate to discuss the general use cases and API design in a dedicated issue first, before returning to this PR.

waj:
@straight-shoota it’s ok not to have experience with every aspect of the language. I’ve been working on this project since the very first lines of code and I also don’t know every corner.

I don’t think I have to open a RFC for every small detail we want to change. This change is actually not big but I do believe it needed some rework to make it easier to use. Formatting the output of a logger is quite an essential part of it.

Please, give it a chance if you want to participate in the changes of this small component. And take this PR as a proposal for the design.

straight-shoota:
I'm not opposing this, I'm just saying it's hard to see whether this is necessary. I wasn't suggesting that I personally need to know about everythings that's going on, but that the Log framework is very new and so far has only been adapted by a few useres as far as I am aware. So it might be a good idea to see what experiences are made with the current API and whether we actually need to change the way the formatter works.
For example, I'm not sure why you would need so much configurability. IMO the defaults should be good enough for at least 80% so that typical applications don't need to bother about defining custom formats. There might be some use cases where you want specific behaviour, but I'd expect that to be handled in a few libraries that provide logging extensions.

waj:
I don’t like the current default output format of the logger so I wanted it to change. But if I just change the format the (infinite) discussion would start ardound what should or shouldn’t be on the output. So at least I wanted to provide an easy way to set a custom output that at the same time lowers the risk of less efficient implementations that affect the performance while logging.
I don’t know any log framework that doesn’t provide a way to customize the output format. And the current one (just a proc) was something Brian and I decided to put just to get it done on the first iteration. 

The proposed api is not even complete in terms of flexibility. More methods should be added to the static formatter to avoid having to fall back so easily to the proc. But this is actually what I’d like to be exercised and get feedback so we can fine tune the api.

bcardiff:
I like this.

I do have some concerns whether using the struct metaclass instead of a struct value would be limiting. But since the `Log::Formatter` is a module I guess we are fine for this limitation to apply only in the built-in formatters.

I know @carlhoerberg cares about overriding the progname in some scenarios within the same app. Since this PR is moving progname from the backend to the formatter there would be a need for new formatters for him. Or maybe with the split of context and local data, that can be used instead of progname.

So, 👍 on my end.

waj:
> Since this PR is moving progname from the backend to the formatter

Actually is being moved to a class variable, so it's the same for all formatters. Changing the progname sounds weird to me, and probably a context value fits better for that case.

z64:
It's not highlighted in the discussion here, but this appears to add backtraces to exception logging. Regardless of this PRs movement, this is :+1: from me; currently I've been having to add `Log.error { ex.inspect_with_backtrace }` in a lot of places to retain our quality of logging.

straight-shoota:
> I don’t like the current default output format of the logger so I wanted it to change. 

Changing the current default is great. We must improve that. And I'm sure we can design a default that just fits for most typical applications.

> But if I just change the format the (infinite) discussion would start ardound what should or shouldn’t be on the output.

I don't agree on that. A) I expect we can agree on a reasonable default formatter. B) Giving too much options on a feature that (usually) doesn't need such options is a wrong development.

IMO this PR is a premature answer to a problem that doesn't exist yet.

Let's work on the default to be great and see how it works for actual users.
I doubt there's gonna be much need for customized formatters. And for the rare cases there is, `alias Log::Formatter = Entry, IO ->` is good enough to roll your own.

If it turns out that custom formatters are needed all over the place and people want a better tools for that, I'm all in for adding this. But as long as there's no indication this is actually useful, I'm strictly against adding mor complexity without reason.

waj:
@straight-shoota the problem exists. I can speak for myself, but many times I wanted to customize the output of a logging format. Of course it's not something I do all the time, but when I wanted the logging frameworks had some mechanism to do so. And I don't know any logging framework not supporting something for that. Ruby's Logger is extremely simple, and in my opinion too much. On the other extreme of the spectrum, log4j has many different kind of [Layout](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Layout.html). But I don't pretend go that far. Only provide the adequate and extensible interface.

I think you're overreacting about the complexity of my PR. In any case, we could remove the macro. I think it's neat but @asterite also expressed privately that he wasn't so sure about it.

All I did was replace the formatter requirement from a proc to a module, so it's now possible use static functions, that favors inlining and dispatches faster. Also it's more convenient to create stateful or configurable formatters than passing a closured proc.

I don't think we're moving away too much from this proposal. It matches what in general we think Crystal should provide as the basis for the standard library.

straight-shoota:
Maybe I think a bit to strong about this. But the fact is not a single real use case has been presented so far. And I'm sure we can't be conrete about this until we have improved the default formatter to a reasonable standard. Only then it's possible to see what use cases it fits and what need extra customization. And for that I'd like to hear about experiences from users.

So I'm not saying we should cut the feature completely. But IMO a necessary step *before* adding enhanced customization options is to provide a sensible default. We're lacking that currently. I understand that this PR adds that, but I'd prefer to work on that first and separate because it doesn't need a new API.

src/log/format.cr
```diff
@@ -0,0 +1,92 @@
...
+    end
+
+    def self.format(entry, io)
+      new(entry, io).run
+    end
+
+    abstract def run
+  end
+
+  macro format(name, pattern)
```
asterite, line 76: For macros that define things, we usually use the `def_` prefix. Or maybe we can name this `define_format`, because `Log.format` seems like it's formatting something, when in reality it's defining a format.

asterite: Maybe even `define_formatter` because the resulting type is a formatter, not a format.

asterite:
I'm fine with this PR as long as:
- `Log::Formatter`, all its subtypes and methods are properly documented.
- An example in `Log::Formatter` is shown defining a subclass of it and using the available methods
- At the very end, include a section mentioning the macro way, and what it expands to.

That is, the macro way should be demystified, and it shouldn't be presented as the main option.

didactic-drunk:
Thank you @waj, this is sorely needed.   The default logger is way too noisy.  4 of my projects already have custom formatters and each is different depending on need.

> Maybe I think a bit to strong about this. But the fact is not a single real use case has been presented so far.

Here's *one* use case (or several depending on how you see it):

```
  INFO [ 0.0015 0.0001 #1074926c0]  -- b_pool: ObjectB ↓ checkout 0.00007 -- {"oid" => "107491a80", "qsize" => 30}
  INFO [ 0.0016 0.0001 #1074926c0]  -- example.a: doing x -- {"x" => 0, "size" => 0 }
  INFO [ 0.0017 0.0001 #1074926c0]  -- example.b -- {"z" => 0, "name" => "..." }
  INFO [ 0.0020 0.0030 #107492cc0]  -- a_pool: ObjectB ↑ checkin  0.00000 -- {"oid" => "10748a300", "qsize" => 74}
          ^       ^                                     ^ Different color shades depending on in or out
          |       | Relative time between log calls or between calls per fiber depending on formatter.
          | Relative time since program start.
```

Colors are missing making the example harder to see.
* Both times have grey/white/bright/yellow/red colors depending on the threshold.
* When qsize drops below a threshold it's colored yellow or red.
* Relative time is more useful than absolute for short program runs like specs and benchmarks.
* The output above represents 2 different formatters.  One tracks relative time between log calls and the other between log calls per fiber.  

[Almost](https://github.com/Blacksmoke16/crylog/tree/master/src/formatters) [every](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger/Formatter.html) [log](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Layout.html) [framework](https://log4cplus.sourceforge.io/docs/html/classlog4cplus_1_1PatternLayout.html) has a formatter/layout.  There's an obvious need if everyone is doing it.

Please don't let your strong opinion of not using hinder those of us who rely on it daily.

straight-shoota:
Thanks @didactic-drunk! It helps to look at this use case.
I'm still not convinced it demonstrates a clear argument for a highly-customizable formatter as proposed in this PR. Maybe a adding another default formatter suited for this use case would be a better idea.
IMO it's always better to try to use standardized log formats as much as possible, because that greatly improves usability. I see little point in having a different log format for every program, just because it a slightly different use case than the others. That's can easily cause sysadmin's nightmares.

Also note that using a custom formatter is already possible and will stay that way. It's only about the interface we're exposing for customizing a formatter. Note that the libraries you linked all expose a very thight interface with only a singl `format` method (or similarly named). Nothing like `StaticFormatter` with essentially a custom DSL for formatting log messages.
Logs4cplus is a bit different as it uses a formatter string, similar to sprintf or strftime. That would actually be more useful because it allows configuration at runtime, so the annoyed sysadmin can configure the logging output to be in a standardized format, even if the author intended some fancy special logging format.

didactic-drunk:
For specs and benchmarks I sometimes replace `timestamp` without touching the remaining log format.

@waj provided exactly what I'm looking for.  The ability to override each log value without recreating an entire log format.

Or to change the format and n values without recreating everything else.  I planned my own PR with exactly these features which I discussed with @Blacksmoke16 privately last week.

didactic-drunk:
(Incomplete) Reusable component case:
```
module ColorSeverity
  def severity(...)
    case ...
    end
  end
end

module ColorTimestamp
  def timestamp
    ...
  end
end

class MyFormat
  include ColorSeverity
  include ColorTimeStamp

  timestamp_colors [{0.001, :gray}, {0.005, :white}, {0.01, ;yellow}, {0.05, :red}]

  def format(...)
    ...
  end
end

class SpecFormat < MyFormat
  include RelativeTime
end

class BenchFormat < MyFormat
  include FiberRelativeTime
end
```

asterite:
@straight-shoota I don't think it's fair to block this PR given that everyone else think this is a good change.

Maybe we can merge this and iterate on the default format later on, even after 1.0? Why does the default format needs to be perfect before 1.0?

straight-shoota:
I have no intention of blocking this. Just wanted to make my objections clear. If you want to merge it, go ahead.

> Maybe we can merge this and iterate on the default format later on, even after 1.0? Why does the default format needs to be perfect before 1.0?

If we find a good default format (or a couple ones to choose from) we likely won't need  `StaticFormatter` at all. So I'd rather ask why we need *that at all* before 1.0. We can perfectly enhance the formatter API after 1.0 without breaking changes by keeping the `Log#formatter=((Entry, IO) ->)` overload.

Good defaults should **always** come before more customization features.
This seems like making two step at once without doing the basic one properly.

asterite:
> I have no intention of blocking thi

Because you requested changed, this can't be merged. We have to dismiss your review, but that's a bit rude.

waj:
> I have no intention of blocking this. Just wanted to make my objections clear. If you want to merge it, go ahead.

But that's what you just did. I heard every objection as always, but you think this is overdesigned and I don't. You think extensibility is not required and I do. But it's just opinions vs opinions.

The code in this PR is the simplest thing to introduce some flexibility and even reducing overhead.

I know I can dismiss the review, but that's not how I'd like things to work here.

bcardiff:
@straight-shoota I think that before and after this PR `Log#formatter : Log::Formatter` holds.
The same goes for `Log::Formatter.new proc` (where `proc : (Entry, IO) ->`)

I think there is a leak of abstraction if you expect `Log#formatter=((Entry, IO) ->)`.

The StaticFormatter is a implementation details that allows simpler definition of formatters, but that is not changing the above interface.

straight-shoota:
Sorry, I wasn't considering that the review type would block merging. I've revoked it.

PR #9216, opened 2020-05-01
Compiler: add implicit block arguments (_1, _2, etc.)

asterite:
**Note**: this PR exists because I wanted to see how hard was it to implement it. It's not decided yet whether this is something that we want... but I enjoy coding these things :-) . Also with this people can try it out themselves.

This is exactly like in Ruby. I chose `_1`, `_2`, etc. instead of `&1`, `&2`, etc. because `&` means the start of a block or a block arguments:

```crystal
foo &.bar
foo &block_argument
```

so it would be a bit confusing to reuse `&` to also mean "block argument":

```crystal
foo { bar &1 } # is &1 a block given to bar?
```

I also think `_1` is visually less noisy than `&1`. And maybe `_1` is a bit easier to type (no need to move the right hand to the middle of the keyboard).

Finally, it's the same as in Ruby so Rubyists will feel familiar with this syntax.

## Examples

```crystal
[1, 2, 3].each { puts _1 }
# Output:
# 1
# 2
# 3

files = ["foo.cr", "bar.cr"]
files.map { File.exists?(_1) }

# same as:
files.map { |file| File.exists?(file) }

%w(World Computer).each { puts "Hello #{_1}!" }
# Output:
# Hello World
# Hello Computer

Hash(String, Array(Int32)).new { _1[_2] = [0] }

# same as:
Hash(String, Array(Int32)).new { |h, k| h[k] = [0] }
```

## Do I like it?

I think for cases like the `File.exists?` above where it's obvious that each block argument will be a file, and the block is pretty small, that it actually improves readability.

Also with `try` when you need to do something other than invoke a method on the yielded object:

```crystal
file = # a nilable string
file.try { File.exists?(_1) }

# instead of:
file.try { |file| File.exists?(file) } # So many files! So confusing!

# or:
file.try { |the_file| File.exists?(the_file) }
```

Note how the two latter ways of writing it introduce a lot of noise (the block syntax, the variable name). The first way is pretty "zen" in my mind.

I also think it finishes generalizing the block syntax. Right now we have the explicit block syntax:

```crystal
[1, 2, 3].map { |x| x.to_s }
```

We have the `&.` shorthand for when you want to invoke a method on the first and only block argument:

```crystal
[1, 2, 3].map &.to_s
```

But no syntax exists for the general case when you want to do something with the block arguments without needing to explicitly mention them. Other languages have `it` (Kotlin, just for the first argument), Ruby now has `_1` and `_2`, and Elixir has the general `&(exp(&1, &2))` expansion syntax.

## Breaking change

This is a tiny breaking change because if you had a variable named `_1`, well, now it's no longer a variable. But I think those cases shouldn't exist... I hope! And fixing those occurrences is easy.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -1650,4 +1650,7 @@ describe Crystal::Formatter do
     1 # foo
     / #{1} /
     CODE
+
+  assert_format "_1"
+  assert_format "_42"
 end
--- a/spec/compiler/lexer/lexer_spec.cr
+++ b/spec/compiler/lexer/lexer_spec.cr
@@ -136,6 +136,15 @@ private def it_lexes_global_match_data_index(globals)
   end
 end
 
+private def it_lexes_implicit_block_argument(code, number)
+  it "lexes #{code}" do
+    lexer = Lexer.new code
+    token = lexer.next_token
+    token.type.should eq(:IMPLICIT_BLOCK_ARGUMENT)
+    token.value.should eq(number)
+  end
+end
+
 describe "Lexer" do
   it_lexes "", :EOF
   it_lexes " ", :SPACE
@@ -274,6 +283,10 @@ describe "Lexer" do
 
   it_lexes "$~", :"$~"
   it_lexes "$?", :"$?"
+
+  it_lexes_implicit_block_argument "_1", 1
+  it_lexes_implicit_block_argument "_2", 2
+  it_lexes_implicit_block_argument "_42", 42
 
   assert_syntax_error "128_i8", "128 doesn't fit in an Int8"
   assert_syntax_error "-129_i8", "-129 doesn't fit in an Int8"
--- a/spec/compiler/parser/parser_spec.cr
+++ b/spec/compiler/parser/parser_spec.cr
@@ -398,6 +398,11 @@ module Crystal
     it_parses "foo &.as?(T).bar", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Call.new(NilableCast.new(Var.new("__arg0"), "T".path), "bar")))
     it_parses "foo(\n  &.block\n)", Call.new(nil, "foo", block: Block.new([Var.new("__arg0")], Call.new(Var.new("__arg0"), "block")))
 
+    it_parses "_1", ImplicitBlockArgument.new(1)
+    it_parses "_42", ImplicitBlockArgument.new(42)
+
+    it_parses "foo _1", Call.new(nil, "foo", [ImplicitBlockArgument.new(1)] of ASTNode)
+
     it_parses "foo.[0]", Call.new("foo".call, "[]", 0.int32)
     it_parses "foo.[0] = 1", Call.new("foo".call, "[]=", [0.int32, 1.int32] of ASTNode)
 
--- a/spec/compiler/semantic/implicit_block_argument_spec.cr
+++ b/spec/compiler/semantic/implicit_block_argument_spec.cr
@@ -1,0 +1,71 @@
+require "../../spec_helper"
+
+describe "Semantic: implicit block argument" do
+  it "errors if implicit block argument outside of block" do
+    assert_error %(
+      _1
+      ),
+      "implcit block argument can only be used inside a block"
+  end
+
+  it "uses implicit block argument" do
+    assert_type(%(
+      def foo
+        yield 1, 'a', true
+      end
+
+      foo do
+        {_1, _2, _3}
+      end
+    )) { tuple_of [int32, char, bool] }
+  end
+
+  it "errors if uses implicit argument where positional argument already exists" do
+    assert_error %(
+        def foo
+          yield 1
+        end
+
+        foo do |x|
+          _1
+        end
+      ),
+      "an explicit block argument at position 1 already exists"
+  end
+
+  it "uses implicit block argument with macro" do
+    assert_type(%(
+      macro moo(x)
+        {{x}}
+      end
+
+      def foo
+        yield 1
+      end
+
+      def bar
+        foo do
+          moo(_1)
+        end
+      end
+
+      bar
+    )) { int32 }
+  end
+
+  it "uses implicit block argument with macro (top level)" do
+    assert_type(%(
+      macro moo(x)
+        {{x}}
+      end
+
+      def foo
+        yield 1
+      end
+
+      foo do
+        moo(_1)
+      end
+    )) { int32 }
+  end
+end
--- a/src/compiler/crystal/semantic/implicit_block_argument.cr
+++ b/src/compiler/crystal/semantic/implicit_block_argument.cr
@@ -1,0 +1,76 @@
+module Crystal
+  class ImplicitBlockArgumentDetector < Visitor
+    def self.has_implicit_block_arguments?(block : Block)
+      detector = new
+      block.body.accept(detector)
+      detector.has_implicit_block_arguments?
+    end
+
+    getter? has_implicit_block_arguments = false
+
+    def visit(node : ImplicitBlockArgument)
+      @has_implicit_block_arguments = true
+    end
+
+    def visit(node : ExpandableNode | Call)
+      if expanded = node.expanded
+        expanded.accept(self)
+        false
+      else
+        true
+      end
+    end
+
+    def visit(node : Block)
+      false
+    end
+
+    def visit(node : ASTNode)
+      !@has_implicit_block_arguments
+    end
+  end
+
+  class ImplicitBlockArgumentTransformer < Transformer
+    def self.transform(program : Program, block : Block)
+      transformer = new(program, block)
+      block.body = block.body.transform(transformer)
+    end
+
+    @initial_block_args_size : Int32
+
+    def initialize(@program : Program, @block : Block)
+      @initial_block_args_size = @block.args.try(&.size) || 0
+    end
+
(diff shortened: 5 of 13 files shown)


paulcsmith:
I love it. And agree on the usefulness for shorter code and for try is really nice. That’s primarily what I’d use it for. Easier to write *and* read in these cases

waj:
This is similar in Elixir also: https://elixir-lang.org/getting-started/modules-and-functions.html

I don’t find it confusing using `&1` because the concepts are related. 

I love this feature. You know it because we talked several times about it 😉. But in any case I would leave the feature and discussion post 1.0

jzakiya:
I could/would definitely use this right now, as it would make allot of my code shorter and easier to read.

oprypin:
*Note: "you" here doesn't mean the author, just anyone*

---

Ah yes. "Pass a block that accepts no arguments and gets the 2nd from the 1st."

That's how one reads this code. Easy discount on [WTFs/sec](https://www.osnews.com/story/19266/wtfsm/). And don't try to tell me you'll get used to it. Well, even if *you* will, the vast majority won't. But I don't believe you will either. This will be a mental interrupt every single time.

Because you couldn't be bothered to declare the args, every time someone reads that code will have to stumble on it. Do we deserve that? Is it OK to keep going in the direction where code can be read only by its author even if it's not logically convoluted?

I personally wouldn't want that, and I definitely don't trust people to not abuse it (i.e. put it into any expression > 20 chars, where the distance between "cool, it accepts no args" and "wait, what's *first*?" is already too much) -- sadly Crystal would be particularly exposed to this compared to other languages due to the association with Ruby where there's just no common sense of restraint on one-liners.

bew:
What's the semantic for nested blocks and block short syntax?

```cr
files.each { Log.debug { _1 } }
```
That won't work, and it'll log nothing, or the arg to Log.debug, or fail at compile time if nothing is given to Log.debug' block (I don't remember). 

But what about:
```cr
files.each { Log.debug &.entry(_1) }
```
Using the dsl syntax demoed in another log pr iirc, will that log each file in `files` ?

asterite:
@bew it aways applies to the closest surrounding block

I do agree with @oprypin that this is not needed and that it might be negative to introduce the feature.

jhass:
Yes, I think we should rather think about how to improve the pass a method as block syntax. Maybe we can allow something like this just in a block shorthand? I think this can solve @oprypin's arguments somewhat, first of all enforcing a short distance and and second it doesn't make you mentally enter into block parse mode by introducing a new "short hand block" parse mode.

Not a syntax proposal, just trying to show that the examples that motivate this can be equally solved with that:
```cr
[1, 2, 3].each { puts _1 }
[1, 2, 3].each &:puts
[1, 2, 3].each &:puts(_1)

files = ["foo.cr", "bar.cr"]
files.map { File.exists?(_1) }
files.map &:File.exists?
files.map &:File.exists?(_1)

%w(World Computer).each { puts "Hello #{_1}!" }
%w(World Computer).each &puts("Hello #{_1}!")

Hash(String, Array(Int32)).new { _1[_2] = [0] }
Hash(String, Array(Int32)).new  &.[_2]=([0]) 
```

Some more from the forum post. Not arguing that the proposal makes them strictly better than the current expanded form, just to demonstrate the problem space that can be attacked by it.
```cr
[1, 2, 3].map { _1 + 3 }
[1, 2, 3].map &.+(3)

(1..9).each_slice(3).map { _1 + _2 + _3 }
(1..9).each_slice(3).map &.+(_2 + _3) # of course that should just be &.sum

[{name: 'foo'}, {name: 'bar'}].map { @1[:name] }
[{name: 'foo'}, {name: 'bar'}].map &.[:name]

user = User.first?
user.try { ChargeCard.call(prince_in-cents: 500, user: _1 }
user.try &:ChargeCard.call(prince_in-cents: 500, user: _1)

password.value.try do
  copy_and_encrypt_password _1, to: encrypted_password
end
password.value.try &:copy_and_encrypt_password(to: encrypted_password)
# above is going to far probably and should be 
password.value.try &:copy_and_encrypt_password(_1, to: encrypted_password)
```

So to formalize:

* Extend `&.` expressions to allow to refer the arguments as `_1, _2, _3, ...`.
* If the call in the `&.` expression  has no explicit argument list, pass it the remaining arguments (second, third, fourth...).
* Introduce `&:` Where `&.` expands to a call on the first block argument, `&:` simply skips that part of the expansion. 
* `&:` expressions support to refer to the arguments as `_1, _2, _3, ...` as well.
* For `&:`, if the call has no explicit argument list, pass the arguments (first, second, third...)

So much for the base. Possible extensions

* Don't require the number of block arguments to match the number of positional arguments exactly. That is if the the method takes fewer positional arguments than there's block arguments, ignore the extra block arguments.
* If the call only has explicit named arguments, pass  the block arguments (for `&.` starting with the second, for `&:` starting with the first) to the method as positional arguments as above.

*Please let's focus on the semantics of this proposal first. Any signs chosen here are explicitly not part of the proposal and just used for demonstration purposes!* I would prefer to keep the discussion focused on semantics and in case those semantics would be a feature beneficial to the language, we can then discuss the best syntax for this.

Sija:
I feel this feature takes us back to the cryptic Perl-land with all those magic `$1` variables and such.

What I'd be **reeeeally** excited about though, is Elixir-like syntax allowing me to do thing like `{"foo.txt" => "bar"}.each(&->File.write(&1, &2))` - which I was missing from the day I've discovered Elixir.

oprypin:
> `{"foo.txt" => "bar"}.each(&->File.write(&1, &2))`

// note that you can do this already:
`{"foo.txt" => "bar"}.each(&->File.write(String, String))`

But yes, that, or https://github.com/crystal-lang/crystal/pull/9197, seems fine to me, as these don't exhibit the biggest problem I see with *this* proposal, which is that the syntax for a block with 0 args is reused for a block with any (not immediately visible) number of args.

Sija:
@oprypin true indeed, what I've missed was the `&.foo($1, $2)` syntax.

oprypin:
Well, that's exactly what this current proposal gives you, then :smile: :question:

asterite:
@Sija it seems this PR is what you want but with a different syntax?

paulcsmith:
> Ah yes. "Pass a block that accepts no arguments and gets the 2nd from the 1st."

> That's how one reads this code. Easy discount on WTFs/minute. And don't try to tell me you'll get used to it. Well, even if you will, the vast majority won't. But I don't believe you will either. This will be a mental interrupt every single time.

@oprypin I get what you are saying here. BUT, I personally think this is easier to read in the examples @asterite posted. Elixir has some similar syntax and I don't see it abused much and is generally only used for short method lines. Let's be honest too, people use short meaningless variable names with blocks *already*. 

For example:

```crystal
Benchmark.ips do |x|
  x.report("short sleep") { sleep 0.01 }
  x.report("shorter sleep") { sleep 0.001 }
end
```

What is `x`? Does anyone know or care? Probably not. I think it is *very* common when yielding blocks that the name is not particularly important.

With all that said, if this is not added it is fine. It is just syntactc sugar that I think has proven to be valuable in places like Elixir and is generally *not* abused, but I get the argument against it 👍

asterite:
Also know that I first implemented this with &1, & 2, etc., so that syntax is definitely possible.

What @jhass suggests, or doing it almost like how elixir does it, by surrounding the block with &(...) might also be a better idea so that you know the replacement is a bit more scoped.

paulcsmith:
I like the idea of wrapping it so it is scoped! And I personally think `&1` is easier to understand since it seems like a block related thing. `_1` looks like an unused variable to me. But that could just be me :)

oprypin:
> Elixir has some similar syntax and I don't see it abused much 

OK but see the last part of my message.

> people use short meaningless variable names with blocks already.

Yes but at least you see upfront how many of the variables there are (as opposed to seeming like there are none)

oprypin:
Yea maybe this proposal could go somewhere if it doesn't reuse the syntax for block with 0 args to mean block with any number of args.

paulcsmith:
@oprypin You mean this?

> sadly Crystal would be particularly exposed to this compared to other languages due to the association with Ruby where there's just no common sense of restraint on one-liners.

Most elixir programmers came from Ruby so if anything this shows that it can work. *especially* if scoped with `&()`

rafaelfesi:
> 
> 
> I like the idea of wrapping it so it is scoped! And I personally think `&1` is easier to understand since it seems like a block related thing. `_1` looks like an unused variable to me. But that could just be me :)

For me, `_1` also seems like an unused variable.

RX14:
Most code is read many (10x) more times than it's written. Often by you. At 2am. Six months later.

For this reason, I'm a big fan of naming things even if it makes typing it a big longer.

paulcsmith:
> Most code is read many (10x) more times than it's written. Often by you. At 2am. Six months later.

> For this reason, I'm a big fan of naming things even if it makes typing it a bit longer

I generally agree with well named methods, argument, etc. But some cases there is no better name. Like in the `try` example or the `benchmark` example. IMO the &1 is just as easy or easier to read in those cases. Just wondering if you felt the same about those examples to or if those looked readable?

e.g.

```crystal
title.value.try { slugify(&1) }
```

pannous:
For one argument blocks add the keyword ‘it’

[1, 2, 3].each { print it }

 (maybe in addition to _1 &1 $1 or #1)

Sija:
Any chances on getting this merged after all? That would be a candy... @asterite?

PR #9217, opened 2020-05-02
Drop `skip_abstract_def_check` flag support

MakeNowJust:
Nine months have passed since 0.30.0 was released. It's about time to drop it.

### DIFF
--- a/src/compiler/crystal/semantic.cr
+++ b/src/compiler/crystal/semantic.cr
@@ -70,12 +70,8 @@ class Crystal::Program
       TypeDeclarationProcessor.new(self).process(node)
     end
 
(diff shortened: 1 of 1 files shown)


asterite:
Thank you!

But what actually needs to be done is turn the warnings into errors. But I think there are still some. bugs with the abstract de checker, so I don't know...

straight-shoota:
Abstract def checks are solid and the warnings have become errors in #9810. I think we can drop the flag now. It would be a breaking change, though. So we might have to delay that for 2.0. But I don't expect the flag to be used and it was only a dev tool that was never really documented. I think it should be fine to remove.

straight-shoota (approved):

PR #9220, opened 2020-05-03
Print original exception after failing to raise it

jhass:


jhass:
Well, unfortunately that was too easy. It seems to break CI

jhass:
Well, looks like I fixed it!

waj:
@jhass, this looks good, but is that `rescue nil` really necessary?

jhass:
Probably not, it's admittedly defensive. I just figured given we're already in a weird state, we don't need to confuse users with any potential more weird error messages they cannot do anything about.

waj:
Yes, I understand. `print_exception` is actually quite defensive already 🙂 https://github.com/crystal-lang/crystal/blob/master/src/crystal/system/print_error.cr#L26

If you don't mind, I would remove that. Actually there are still many `rescue nil` in the std that I want to get rid of (and after that I would like to remove that feature completely 😝)

PR #9224, opened 2020-05-04
Support `TypeNode.name(generic_args: false)` for generic instances

### LINKED ISSUE #9223: TypeNode with generic_args as false doesn't omit populated types
Related to #8483 

Not sure if this is intentional or not, but when using `@type` in a macro method, the resolved type parameters still get included when `generic_args` is set to false.

```crystal
class Foo(T)
  getter foo : T

  def initialize(@foo : T)
  end

  def do_something
    puts {{@type.name(generic_args: false).stringify}}
  end
end

Foo.new(42).do_something

# Produces:
# Foo(Int32)
# instead of:
# Foo
```

Crystal v0.34

Blacksmoke16:
Fixes #9223


straight-shoota:
Thank you, @Blacksmoke16

PR #9227, opened 2020-05-04
Add Log Metadata

### LINKED ISSUE #9188: [RFC] Log Metadata
This RFC raises some possible changes over the current API for log metadata (currently context), in particular with the concern of possible performance issues. I discused these points with @bcardiff over the phone and we decided to share some thoughts.

## 1. Split context and entry metadata

Currently the only way to write metadata values for each log entry is using the log context. This was intentional as the context was, in the initial design, understood as the "contextual" information for the current activity, and not data applicable only to the current log entry. So, for example, the current context could contain things like a "request ID", or the current user logged in. However, specific data like the duration of an operation, or the error code returned by a remote server, are not contextual values. So if someone wants to log that information as structured data and not just a text message, the current context must be replaced just for that log entry. Some work was done in #9146 to make that task easier, but with an extra performance impact.

Instead, for these cases, we propose that each entry has not only the current context metadata, but also some specific data we want solely that log entry to contain. If this data is provided as a named tuple, then this operation is very lightweight, avoiding having to modify the current fiber's context and merging with the existing values in that structure.

In other words `Log::Entry` will now contain not only `context` but also an optional `data` property.

This has one caveat that is there could be now some overlapping between the keys in the context and the entry data. I think the impact is minimal because the nature of the data on each case makes it rare to occur. We could either defer the compaction of both structures at the time the entries are processed by the backend, or it could be stored as separate entities.

## 2. DSL to set entry metadata

The design of the log APIs circle around the idea of minimizing the performance impact they have. So, it's really important to avoid executing operations that are later discarded because of the severity level currently configured for the process (or the log source).

Because of this, passing the metadata as arguments to the `Log` calls is not always a good option. Currently the text message is returned from a block due the same reason: skip executing interpolations if the result is then discarded.

The proposal is to extend this idea and let the block return either a `String` or a `Log::Entry` and provide some basic DSL to ease this task.

So, while this continue to be fully valid:

```crystal
Log.info { "Some message" }
```

in case extra metadata is desired to be attached to the entry, a DSL object passed as argument could be used:

```crystal
Log.info { |dsl| dsl.entry("Some message", foo: 123, bar: "More data with #{interpolation}") }
```

And actually, a more compact way is possible:

```crystal
Log.info &.entry("Some message", foo: 123, bar: "More data with #{interpolation}")
```

Other options for this DSL could be added for other scenarios yet to discover.

## 3. Use symbols as metadata keys

During the first iterations of the log context design, the keys of the entries were symbols. This was later changed to be strings, in order to allow dynamic values unknown at compile time. I strongly think this is a mistake and we want to revert it back to just symbols. The reason is again, simple: performance. Even though we could optimize comparison between literal strings, if the API allows any string to be used as a key, there is always the possibility to affect the performance unnecessarily. The comparison of symbols will be always much better, and there is simply no reason to use unknown values for the keys. For example, if the request headers are desired to be stored in the current context, they header names themselves don't need be the keys of the context values. Instead, we could store the headers under a `:http_headers` key or similar.

bcardiff:
Fixes #9188 

* The `Log::Context` datum is renamed to `Log::Metadata` and it is used as the Fiber's logging (that is accessed as before via `Log.context`)
* I needed to keep a `Log::Context` type to allow `Log.context.set a: 1` and `Log.context.clear`
* I changed the IOBackend to show the `Entry#data` in at least some way if it's present.
* If the block returns a string, the `Log.context` is able to be changed within the block (related to #9146), but it no longer reverts it.

The following constructs are all possible with this PR:

```crystal
Log.info { "Program started" }
Log.info &.emit("Program started") # same as previous
Log.info &.emit("User logged in", user_id: 42) # local data
Log.info &.emit("User logged in", expr_that_computes_hash_named_tuple_or_metadata)
Log.warn exception: e, &.emit("Oh no!", user_id: 42) # with exception
Log.warn exception: e, &.emit("Oh no!") # with exception, no local data
Log.warn(exception: e) { "Oh no!" } # same as previous
Log.info &.emit(action: "log_in", user_id: 42) # empty message
```

cc: @waj, @paulcsmith


straight-shoota:
> JSONLogFormatter

I'm pretty sure we should eventually have that in stdlib as well. Maybe it could be upstreamed from Dexter.

paulcsmith:
I'd be happy to make a PR for it once this is merged. Should make the implementation a bit simpler actually :)

bcardiff:
* I fixed some docs that was missing changes due to the rename
* And added a spec/enforce immutability via clone instead of dup since it was breaking the immutability guarantee. #shame

bcardiff:
I will keep iterating with some ideas we talk with @waj to improve the performance of contexts. That might affect a bit in some APIs of Log::Metadata

But since this PR is more focused on adding local data to Entries and reuse Log::Metadata in fiber context and entries, it's better to move this forward and also iterate on #9211.

Stay tuned 🤞

will:
@bcardiff All this logging stuff is cool :) 

I was hoping to log both symbols and class names, but it looks like right now I have to call `to_s` on them myself. I poked around a bit at making this automatic  in `Value`, but hit a bit of a dead end with getting it not to also to_s arrays. I also noticed `UInt64` wasn't in the datum macro, but that entire macro is new to me.

PR #9229, opened 2020-05-05
Array Docs and Types

wontruefree:
This adds some docs and examples for methods that did not have some and adds some types.  I left some types empty based on my preserved usage.

### DIFF
--- a/src/array.cr
+++ b/src/array.cr
@@ -1196,9 +1196,11 @@ class Array(T)
   # sums                                        # => [3, 4, 3, 5, 4, 5]
   # ```
   #
-  # By default, a new `Array` is created and returned for each combination. If reuse is given, the `Array`
-  # can be reused: if reuse is an `Array`, this `Array` will be reused; if reuse if truthy, the method will
-  # create a new `Array` and reuse it. This can be used to prevent many memory allocations when each slice of
+  # By default, a new array is created and yielded for each permutation.
+  # If *reuse* is given, the array can be reused: if *reuse* is
+  # an `Array`, this array will be reused; if *reuse* if truthy,
+  # the method will create a new array and reuse it. This can be
+  # used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def each_permutation(size : Int = self.size, reuse = false) : Nil
     n = self.size
@@ -1247,9 +1249,11 @@ class Array(T)
   # iter.next # => #<Iterator::Stop>
   # ```
   #
-  # By default, a new `Array` is created and returned for each combination. If reuse is given, the `Array`
-  # can be reused: if reuse is an `Array`, this `Array` will be reused; if reuse if truthy, the method will
-  # create a new `Array` and reuse it. This can be used to prevent many memory allocations when each slice of
+  # By default, a new array is created and returned for each permutation.
+  # If *reuse* is given, the array can be reused: if *reuse* is
+  # an `Array`, this array will be reused; if *reuse* if truthy,
+  # the method will create a new array and reuse it. This can be
+  # used to prevent many memory allocations when each slice of
   # interest is to be used in a read-only fashion.
   def each_permutation(size : Int = self.size, reuse = false) : Iterator
     raise ArgumentError.new("Size must be positive") if size < 0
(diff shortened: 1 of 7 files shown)


src/array.cr
```diff
@@ -2058,7 +2286,14 @@ class Array(T)
     self
   end
 
-  def update(index : Int)
+  # Modifies *self* at *index* to the value *yieled* block and returns the result of the block.
+  # Raises IndexError if the index is out or range.
```
jhass, line 2290:
```suggestion
  # Sets the value returned from the given block at the given index.
 #
  # Raises `IndexError` if the index is out or range.
```

src/array.cr
```diff
@@ -2104,6 +2339,7 @@ class Array(T)
     super
   end
 
+  # :nodoc:
   private class PermutationIterator(T)
```
jhass, line 2343: It's private, it already doesn't show up in docs.

src/array.cr
```diff
@@ -2163,6 +2399,7 @@ class Array(T)
     end
   end
 
+  # :nodoc:
   private class CombinationIterator(T)
```
jhass, line 2403: Ditto

src/array.cr
```diff
@@ -2224,6 +2461,7 @@ class Array(T)
     end
   end
 
+  # :nodoc:
   private class RepeatedCombinationIterator(T)
```
jhass, line 2465: Ditto.

src/array.cr
```diff
@@ -2282,14 +2520,15 @@ class Array(T)
     end
   end
 
+  # :nodoc:
   private struct FlattenHelper(T)
```
jhass, line 2524: Ditto.

src/array.cr
```diff
@@ -2282,14 +2520,15 @@ class Array(T)
...
-    def self.flatten(ary)
+    def self.flatten(ary) : Array(T)
```
jhass, line 2525: Given this is an internal and undocumented method on an internal and undocumented type, maybe the return type restriction is of little value here?

src/array.cr
```diff
@@ -2282,14 +2520,15 @@ class Array(T)
...
       result = [] of T
       flatten ary, result
       result
     end
 
-    def self.flatten(ary : Array, result)
+    def self.flatten(ary : Array, result) : Array(T)?
```
jhass, line 2531: Ditto.

src/array.cr
```diff
@@ -1393,10 +1393,10 @@ class Array(T)
   # arr = [1, 2, 3]
   # arr.repeated_combinations(0) # => [[]]
   # arr.repeated_combinations(2) # => [[1, 1], [1, 2], [1, 3], [2, 2], [2, 3], [3, 3]]
-  # arr.repeated_combinations    # => [[1, 1, 1], [1, 1, 2], [1, 1, 3], [1, 2, 2], [1, 2, 3],
-  #                              #     [1, 3, 3], [2, 2, 2], [2, 2, 3], [2, 3, 3], [3, 3, 3]
+  # arr.repeated_combinations    # => [[1, 1, 1], [1, 1, 2], [1, 1, 3], [1, 2, 2], [1, 2, 3], [1, 3, 3], [2, 2, 2] ...]
```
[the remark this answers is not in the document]
wontruefree: The full result was very wide but I can add it back.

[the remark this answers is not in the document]
jhass: Maybe we can add some newlines?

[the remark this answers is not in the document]
wontruefree: this new lines looked a little weird after the formatter ran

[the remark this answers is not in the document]
jhass: Might be something to optimize in the formatter then. But I think removing information here is not helpful to understanding this method.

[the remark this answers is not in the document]
straight-shoota: The code should work with the example checker. So abbreviating is not a solution. Can we use a shorter example?

src/array.cr
```diff
@@ -1393,10 +1393,10 @@ class Array(T)
...
   # ```
-  def repeated_combinations(size : Int = self.size) forall Array(T)
+  def repeated_combinations(size : Int = self.size) forall Array
+    (T)
```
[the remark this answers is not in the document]
wontruefree: The formatter is actually doing this???

straight-shoota:
@wontruefree Is there a particular reason to close this PR? Or do you just don't intend to put any more effort in it?

wontruefree:
@straight-shoota I would like to work on these Docs PRs.  There were a lot of comments and some of them I was confused on where they stood.  I figure I would just submit smaller PRs that could be reviewed more easily.

PR #9230, opened 2020-05-05
Escape CDATA end sequences

### LINKED ISSUE #9228: Should `XML::Builder#cdata` escape nested `]]>`
Currently `XML::Builder#cdata` happily encodes `]]>` within the provided string.

https://play.crystal-lang.org/#/r/90ev

```cr
require "xml"
 
string = XML.build(indent: "  ") do |xml|
  xml.element("data") { xml.cdata "FOO]]>BAR" }
end
 
string # => <data><![CDATA[FOO]]>BAR]]></data>
```

This ultimately causes it to generate invalid XML.

The question is, should it internally handle escaping the end sequence, or leave that up to the user to handle?

Blacksmoke16:
Fixes #9228

### DIFF
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -191,7 +191,7 @@ struct XML::Builder
     yield.tap { end_cdata }
   end
 
-  # Emits a `CDATA` section.  Escapes nested CDATA end sequences.
+  # Emits a `CDATA` section.  Escapes nested `CDATA` end sequences.
   def cdata(text : String) : Nil
     call WriteCDATA, string_to_unsafe(text.gsub(CDATA_END, CDATA_ESCAPE))
   end
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -184,7 +184,7 @@ struct XML::Builder
   # Emits the start of a `CDATA` section, invokes the block
   # and then emits the end of the `CDATA` section.
   #
-  # NOTE: CDATA end sequences written within the block
+  # NOTE: `CDATA` end sequences written within the block
   # need to be escaped manually.
   def cdata(&)
     start_cdata
--- a/src/xml/builder.cr
+++ b/src/xml/builder.cr
@@ -5,6 +5,9 @@
 # without a matching `start_element`, or trying to use
 # a non-string value as an object's field name)
 struct XML::Builder
+  private CDATA_END    = "]]>"
+  private CDATA_ESCAPE = "]]]]><![CDATA[>"
+
   @box : Void*
 
   # Creates a builder that writes to the given *io*.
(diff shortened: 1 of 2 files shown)


src/xml/builder.cr
```diff
@@ -180,14 +183,17 @@ struct XML::Builder
...
+  def cdata(&)
     start_cdata
     yield.tap { end_cdata }
   end
 
-  # Emits a `CDATA` section.
+  # Emits a `CDATA` section.  Escapes nested CDATA end sequences.
   def cdata(text : String) : Nil
-    call WriteCDATA, string_to_unsafe(text)
+    call WriteCDATA, string_to_unsafe(text.gsub(CDATA_END, CDATA_ESCAPE))
```
Blacksmoke16, line 196: I think ideally `gsub!` would make more sense here to prevent needing to allocate another string.  But unfortunately that method doesn't exist :/

jhass: Yes, strings are immutable. You wouldn't want to modify the caller's value anyways.

jhass: I guess potentially faster and maybe we allocation free could be splitting on the end marker and writing each part separately? Nothing I would require here though.

src/xml/builder.cr
```diff
@@ -180,14 +183,17 @@ struct XML::Builder
 
   # Emits the start of a `CDATA` section, invokes the block
   # and then emits the end of the `CDATA` section.
-  def cdata
+  #
+  # NOTE: CDATA end sequences written within the block
```
Sija, line 187:
```suggestion
  # NOTE: `CDATA` end sequences written within the block
```

asterite:
Out of curiosity: did you check what nokogiri does?

src/xml/builder.cr
Sija, line 187: Why the behaviour of `cdata(&)` differs from the non-yielding version?

Blacksmoke16:
@asterite I checked Go and Nokogiri

* Go - https://golang.org/src/encoding/xml/xml.go#L2014
* Nokogiri:

```ruby
require "nokogiri"
builder = Nokogiri::XML::Builder.new do |xml|
  xml.data {
  	xml.cdata "FOO]]>BAR"
  }
end

builder.to_xml # => <data><![CDATA[FOO]]]]><![CDATA[>BAR]]></data>
```

@Sija Because in the block version the method never gets a reference to the string.

I.e. you could do https://play.crystal-lang.org/#/r/90i5.  In order to handle that we would need to escape `#text` too which might not be wanted as it only need to be escaped in the context of a cdata node.

asterite:
@Blacksmoke16 What I mean is, how they do it. I would be pretty surprised if libxml2 didn't provide a way to escape CDATA.

asterite:
Oh, well, I guess go does it manually. Nokogiri seems to use `xmlNewCDataBlock` but I'm not sure we can use this. I guess this implementation is fine then.

src/xml/builder.cr
straight-shoota: Strings are immutable. There are no mutating methods on `String` at all. And it would be quite unexpected that a string is changed in the call scope after it is passed to this method.

straight-shoota: Oops, somehow the comment thread was not updated while the review view was open.

straight-shoota:
> In order to handle that we would need to escape #text too which might not be wanted as it only need to be escaped in the context of a cdata node.

And it wouldn't be safe because the CDATA end sequence could be written in separate calls to `text`.

Blacksmoke16:
Bump.  Can I get another review on this?

PR #9231, opened 2020-05-05
refactor: remove heredoc related code from StringLiteral formatter

MakeNowJust:
Because all heredoc are `StringInterpolation` even if it has no interpolation.


MakeNowJust:
ping. It is simple dead code removing.

PR #9232, opened 2020-05-05
Parser: fix parsing of capitalized named argument

### LINKED ISSUE #9221: Key name restriction inconsistency between keyword arguments and named tuple keys
We intend lossless conversion between keyword arguments and named tuples, otherwise we would not have the double splat (`**args`) syntax.

Yet while `{Foo: 1}` is a valid named tuple, `NamedTuple.new(Foo: 1)` is not a valid call.

We should resolve this either by disallowing the first case or allowing the second case. I don't have a strong preference either way.

asterite:
Fixes #9221


straight-shoota:
Muchas gracias, @asterite

PR #9234, opened 2020-05-05
Fix `crystal spec` file paths on Windows

oprypin:
The entry point file is being generated with the paths being pasted directly into source code (`require "string"`) but `require` always needs forward slashes, while backslashes actually end up being interpreted as escape sequences.

### DIFF
--- a/src/compiler/crystal/command/spec.cr
+++ b/src/compiler/crystal/command/spec.cr
@@ -66,7 +66,9 @@ class Crystal::Command
 
     source_filename = File.expand_path("spec")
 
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
Probably doesn't really matter in this case since it's just a spec.  But you could probably use the block version of `#join` to avoid the intermediary array.

https://crystal-lang.org/api/master/Enumerable.html#join(separator=%22%22,&)-instance-method

oprypin:
Absolute paths don't work anyway with crystal spec

oprypin:
@Blacksmoke16 That's a good suggestion but I tend to not refactor pre-existing issues unless it's the main point of my PR or it depends on it

PR #9236, opened 2020-05-05
Add IO overloads to various String case methods

### LINKED ISSUE #9215: IO overloads for `String#*case` methods
Related to #8923 but for the various `*case` methods including:
* underscore
* downcase
* upcase
* capitalize
* camelcase
* titleize #9204?

Blacksmoke16:
resolves #9215

I'd suggest reviewing with whitespace changes hidden.


spec/std/string_spec.cr
```diff
@@ -541,9 +541,22 @@ describe "String" do
     it { "ﬀ".downcase(Unicode::CaseOptions::Fold).should eq("ff") }
     it { "tschüß".downcase(Unicode::CaseOptions::Fold).should eq("tschüss") }
     it { "ΣίσυφοςﬁÆ".downcase(Unicode::CaseOptions::Fold).should eq("σίσυφοσfiæ") }
+
+    describe "with IO" do
+      it { with_io_memory { |io| "HELLO!".downcase io }.should eq "hello!" }
```
asterite, line 546: I introduced `with_io_memory` in a recent commit and was suggested to use `String.build`. Can we use that instead? Totally optional though, I can send a separate PR to fix that later (there's no need for that helper function).

Blacksmoke16: Oo good call.

src/string.cr
```diff
@@ -1056,59 +1056,70 @@ class String
...
+  # Returns a new `String` with each uppercase letter replaced with its lowercase counterpart.
   #
   # ```
   # "hEllO".downcase # => "hello"
   # ```
-  def downcase(options = Unicode::CaseOptions::None)
+  def downcase(options : Unicode::CaseOptions = :none) : String
     return self if empty?
 
+    String.build(bytesize) { |io| downcase io, options }
```
asterite, line 1072: This is losing the optimization that for ascii we know exactly the size. `String.build` is a bit slower than `String.new`.

src/string.cr
```diff
@@ -1056,59 +1056,70 @@ class String
...
+  # Returns a new `String` with each lowercase letter replaced with its uppercase counterpart.
   #
   # ```
   # "hEllO".upcase # => "HELLO"
   # ```
-  def upcase(options = Unicode::CaseOptions::None)
+  def upcase(options : Unicode::CaseOptions = :none) : String
     return self if empty?
 
+    String.build(bytesize) { |io| upcase io, options }
```
asterite, line 1104: Same comment as before. I think the definitions need to be duplicated or a different reuse must be done to avoid losing that optimization.

spec/std/string_spec.cr
asterite: Thank you! 🙏

src/string.cr
Blacksmoke16: Is there a benefit in having the optimization in the IO overload as well?  I.e. if the main benefit is _how_ the string is created could we leave the optimization only in the non IO overload.  Then the IO overload could just use `eac_char`?

asterite: I think only a benchmark can tell that. But given that we didn't have the IO overload before, using each_char is better than nothing.

Blacksmoke16: Looks like `#each_char` as an optimization for `ascii_only` strings, so probably the best we're going to get since it has to be written to an IO.

src/string.cr
```diff
@@ -3672,74 +3712,83 @@ class String
...
+  # "DoesWhatItSaysOnTheTin".underscore io
+  # io.to_s # => "does_what_it_says_on_the_tin"
+  # ```
+  def underscore(io : IO, options : Unicode::CaseOptions = :none) : Nil
     first = true
     last_is_downcase = false
     last_is_upcase = false
     last_is_digit = false
-    mem = nil
+    mem : Char? = nil
```
straight-shoota, line 3731: Why the type restriction?

Blacksmoke16: Make it more clear to the next person what it is.  I suppose I could have renamed it to `char` or something, but this was the least invasive solution.

Blacksmoke16:
Bump. Can I get another review on this?

I'll resolve the conflicts this afternoon.

Issue #9172, opened 2020-04-24
Support half-precision floating point (Float16)


albertorestifo:
While working on an implementation of a [CBOR encoder/decoder][1] I hit a roadblock, as the protocol allows for the transmission of [half-precision floating point numbers][2], but crystal is lacking a `Float16` type.

It would be nice to have a `Float16` type in Crystal.

[1]: https://tools.ietf.org/html/rfc7049
[2]: https://en.wikipedia.org/wiki/Half-precision_floating-point_format

jhass:
Meanwhile it's probably possible to encode a small enough `Float32` as  `Float16` into a `UInt16` with casting it into a  `UInt32` first and  then  doing the right  dance of masks and shifts.

albertorestifo:
I tried, but it was very difficult to handle all cases properly (like NaN and Infinity). Most likely I was also doing something wrong as I'm not very familiar with bit operations, I based it off [this gist](https://gist.github.com/martin-kallman/5049614).

RX14:
This would need compiler support too.

RX14:
Actually, taking for example rust, there's no `Float16` type, but there's a crate which can convert back and forth from Float32. I think a shard to do that would be the best option, and all the arithmetic is performed on `Float32`. I'd like to try that approach first before adding anything to the compiler or stdlib.

albertorestifo:
That's a very good idea!

Look at the rust crate, I have absolutely not idea [what is going on here][1], however [this function][2] might be a very good starting point for a shard.
I'll have a look and see if I can pull something good off it.

[1]: https://github.com/starkat99/half-rs/blob/master/src/binary16/convert.rs#L420-L425
[2]: https://github.com/starkat99/half-rs/blob/master/src/binary16/convert.rs#L269

RX14:
@albertorestifo the first function is a function which uses the `VCVTPH2PS` X86_64 instruction to speed up the conversion from f16 to f32. That would be done in crystal with some custom assembly, but getting something working first is good, before you make it fast :)

albertorestifo:
[Here it is][1] a first basic implementation as part of the CBOR library I'm working on.

All the CBOR Float16 tests provided in the RFC are now passing, giving me a reasonable certainty that the conversion from the Rust code was correct.

I'll now focus on finishing the CBOR library before properly extracting this into a Float16 library.

[...]

PR #9237, opened 2020-05-06
Add Float16

asterite:
Implements #9172


src/float16.cr
```diff
@@ -0,0 +1,107 @@
+lib LibIntrinsics
+  fun f16tof32 = "llvm.convert.from.fp16.f32"(Int16) : Float32
```
Blacksmoke16, line 2:
```suggestion
  fun f16tof32 = "llvm.convert.from.fp16.f32"(Float16) : Float32
```

And ditto below?

asterite: I thought about this, but I wanted to keep the original LLVM API. All of this is then abstracted by `Float16`. Using `Float16` here feels a bit circular dependency to me (which is not, but still...).

asterite:
Well, thank you LLVM and your great "portability" which means implementing the same functions over and over again in every front-end out there... \*sigh\*...

asterite:
Can't be merged, only works on mac for some reason...

asterite:
This seems to require some functions to be defined (maybe there are defined in mac for some reason):
- `__gnu_h2f_ieee`
- `__truncdfhf2`

Similar to how we define `mulodi4` in `crystal/compiler_rt.cr`.

asterite:
This is LLVM:
- Yeah, we support i128 out of the box! -> Truth: as long as you implement the functions that we call, which basically mean you have to implement i128
- Yeah, we support float 16 functions! -> Truth: as long as you implement the functions that we call, which basically mean you have to implement float16

LLVM ❤️

jhass:
Actually forgive my ignorance, I didn't fully research how all of this fits together, but why are we not just including https://compiler-rt.llvm.org/?

waj:
@bcardiff might know better, but IIRC some functions are defined in headers. Some wrapper would be necessary to link with Crystal and complicate the inlining. It would be great if we find a way to do it.

RX14:
I'm extremely unsure how compiler-rt fits in with `libgcc`, which is also linked in with all gcc-compiled programs. Do they conflict?

We should definitely be linking `compiler-rt` and not trying to implement all these in crystal, but how does that work in practice?

jhass:
FWIW  Archlinux: Package [compiler-rt](https://www.archlinux.org/packages/extra/x86_64/compiler-rt/)

```
$ nm /usr/lib/clang/10.0.0/lib/linux/libclang_rt.builtins-x86_64.a  2>/dev/null | grep mulodi4
mulodi4.c.o:
0000000000000000 T __mulodi4
$ nm /usr/lib/clang/10.0.0/lib/linux/libclang_rt.builtins-x86_64.a  2>/dev/null | grep gnu_h2f_ieee
00000000000000a0 T __gnu_h2f_ieee
$ m /usr/lib/clang/10.0.0/lib/linux/libclang_rt.builtins-x86_64.a  2>/dev/null | grep truncdfhf2
truncdfhf2.c.o:
0000000000000000 T __truncdfhf2
```

I didn't check extensively but I think all the ones from #8313 are there too.

macos:

```
$ nm /usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a 2>/dev/null | grep mulodi4
/usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a(mulodi4.c.o):
0000000000000000 T ___mulodi4
$ nm /usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a 2>/dev/null | grep gnu_h2f_ieee
0000000000000090 T ___gnu_h2f_ieee
$ nm /usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a 2>/dev/null | grep truncdfhf2
/usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin/libclang_rt.osx.a(truncdfhf2.c.o):
0000000000000000 T ___truncdfhf2
```

Similarly the int128 ones seem  to be there too.

The naming differences on the library files baffles me a bit but I'd hope  there's some command that gives us the right linker flags.

asterite:
Note that if we find a way to include compiler-rt, we'll also have full support for Int128

jhass:
Also interesting tidbit:

> The default runtime library is target-specific. For targets where GCC is the dominant compiler, Clang currently defaults to using libgcc_s. On most other targets, compiler-rt is used by default.

https://clang.llvm.org/docs/Toolchain.html#compiler-runtime

bcardiff:
The history was that 

1. compiler-rt is not as minimal as some would like. binary size will increase.
2. linking always against compiler-rt will make it harder on some non official distros probably. arm has many flavors due to soft/hard float AFAIK. https://embeddedartistry.com/blog/2017/03/17/compiler-rt/
3. If we embedded it I'm not sure if there will be incompatibilities between compiler-rt versions and llvm version.

In #8313 a significant effort to port compiler-rt was done. The goal was Int128 support, it does not seem to cover the Float 16 functions.

Since Int128 was treated as a second class citizen it didn't got enough efforts to make that PR through. The last thought was to put it in a separate repo as a shard and vendor the code in the compiler. The specs are slow and the code won't change that often.

For the record, I'm fine with linking to compiler-rt. Yet I find it appealing to use Crystal code when possible.

jhass:
> Yet I find it appealing to use Crystal code when possible.

I agree, we could look at giving a flag to prevent linking it, as well as a flag to override it's location, then promote a shard that reimplements it in Crystal for experienced/experimental users. If that shard matures we can upstream it into the compiler.

Meanwhile:
```
arch $ clang -print-libgcc-file-name -rtlib=compiler-rt
/usr/lib/clang/10.0.0/lib/linux/libclang_rt.builtins-x86_64.a
arch $ clang -print-libgcc-file-name -rtlib=libgcc
/usr/bin/../lib64/gcc/x86_64-pc-linux-gnu/9.3.0/libgcc.a
macos $ clang -print-libgcc-file-name -rtlib=compiler-rt
/usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin19.4.0/libclang_rt.builtins-x86_64.a
macos $ clang -print-libgcc-file-name -rtlib=libgcc
clang-10: error: unsupported runtime library 'libgcc' for platform 'darwin'
/usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin19.4.0/libclang_rt.builtins-x86_64.a
macos $ ls /usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin19.4.0/libclang_rt.builtins-x86_64.a
ls: /usr/local/Cellar/llvm/10.0.0_3/lib/clang/10.0.0/lib/darwin19.4.0/libclang_rt.builtins-x86_64.a: No such file or directory
```

Sigh. Something  is wrong with the homebrew build I guess.

For cross compiling the user would need to figure out the right path when linking on the target, yeah.

RX14:
Linkers ignore when there are two shared libraries with symbols with the same name right? So we can include these symbols into crystal as long as we're sure they're identical and there are no problems?

jhass:
Mmh, not sure how `a`s work, aren't they always fully integrated into the binary. But when it comes to symbol resolution it's definitely first come first serve, the main binary  can shadow any other symbols, that's how github.com/jhass/crystal-malloc_pthread_shim works. 

I never tried but  I would assume that also goes for the order of `-l` flags. Say with both defining symbol `x`, `-lfoo -lbar` takes the one from `foo` while `-lbar -foo` takes the one from `bar`. At least that's how it would make sense for me, but somebody should verify. Point is I know the main binary can shadow library symbols for sure.

RX14:
> I never tried but I would assume that also goes for the order of `-l` flags. Say with both defining symbol `x`, `-lfoo -lbar` takes the one from `foo` while `-lbar -foo` takes the one from `bar`. At least that's how it would make sense for me, but somebody should verify. Point is I know the main binary can shadow library symbols for sure.

Yeah, I thought it was based on library ordering too - first one wins. Looks like compiler-rt and libgcc_s are abi compatible where they do share symbols so it should be fine??

Issue #4276, opened 2017-04-11
Static linking + unwinding are flat-out broken under Alpine Linux


refi64:
I've been working on a project called [Rootbox](https://project-rootbox.github.io) that runs Alpine Linux inside a chroot, and I thought a great, easy way to show it off would be to statically-compile a Crystal project inside of it.

Key word: **thought**

Of course, I quickly realized that libgc and libevent aren't available statically under Alpine Linux. I built them manually, [running into a Boehm GC bug in the process](https://github.com/ivmai/bdwgc/issues/154). And then I made it work!! And then I realized that IT DOESN'T ACTUALLY WORK!!

Basically, here's how I built Boehm GC:

```
sudo apk add libatomic_ops-dev

git clone https://github.com/ivmai/bdwgc.git
cd bdwgc
./autogen.sh

export CFLAGS="-D_GNU_SOURCE -DNO_GETCONTEXT -DUSE_MMAP -DIGNORE_DYNAMIC_LOADING"
./configure --prefix=/usr --datadir=/usr/share/doc/gc --enable-cplusplus \
            --disable-parallel-mark --disable-shared --enable-munmap
make
sudo cp .libs/libgc.a /usr/lib
```

and then libevent:

```
curl -Lfo libevent.tgz $LIBEVENT_URL
tar xvf libevent.tgz
cd $LIBEVENT_STEM
./configure --prefix=/usr --sysconfdir=/etc
make
sudo cp .libs/libevent.a /usr/lib
```

and now the broken program itself:

```
echo 'raise "hello!"' > x.cr
crystal build -o x x.cr --link-flags -static
```

Then, when I run it, I get this:

```
DevPC-LX:~/w$ ./x
Invalid memory access (signal 11) at address 0x0
Segmentation fault (core dumped)
DevPC-LX:~/w$ 
```

*The unwind handler in libgcc_eh is crashing.* Great. Then I tried using [LLVM's libunwind](https://github.com/llvm-mirror/libunwind) instead. Results:

```
DevPC-LX:~/w$ crystal build -o x x.cr --link-flags "-static -L/home/user/w"
DevPC-LX:~/w$ ./x
Failed to raise an exception: END_OF_STACK
[0] ???
DevPC-LX:~/w$ 
```

If you want to reproduce in the Rootbox, here's how (I swear, this isn't just a plug ;):

- Follow [these *Getting Started* directions](https://project-rootbox.github.io/guide.html#getting-started).
- Run:

    ```
[...]

PR #9238, opened 2020-05-06
Fix segfaults when static linking with musl

waj:
This is a workaround for segfaults that are showing up when raising exceptions in musl and the binary was statically linked.

The issue is with `pthread_once`. In musl is defined as a weak symbol:
```
$ nm /usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/libgcc.a
...
                 w pthread_once
...
```

The unwind API implementation uses this function, but through a wrapper `__gthread_once`: https://github.com/gcc-mirror/gcc/blob/8d9254fc8aa32619f640efb01cfe87cc6cdc9ce1/libgcc/unwind-dw2.c#L1597-L1600

But the implementation of `__gthread_once` refers to the weak version of the function:
https://github.com/gcc-mirror/gcc/blob/8d9254fc8aa32619f640efb01cfe87cc6cdc9ce1/libgcc/gthr-posix.h#L696-L703

For some reason I don't fully understand yet, that makes the linker to not include that function in the final binary.

Note also that `__gthread_once` checks `__gthread_active_p` to see if the current code is multithreaded or not. It's a hack that looks for other existing symbols within the same binary. That's why when linking with the GC it failed but it doesn't when compiling with `-Dgc_none`: the GC uses multithreaded code and thus links to `pthread_create` and other symbols.

Thank you very much to @ggiraldez for help with this debugging!!!

*NOTE*: the backtraces still cannot be decoded with static linking, but that's a different issue

Fixes: #4276, #6934

### DIFF
--- a/src/crystal/system/unix/pthread.cr
+++ b/src/crystal/system/unix/pthread.cr
@@ -178,3 +178,19 @@ class Thread
     @th
   end
 end
+
+# In musl (alpine) the calls to unwind API segfaults
+# when the binary is statically linked. This is because
+# some symbols like `pthread_once` are defined as "weak"
+# and, for some reason, not linked into the final binary.
+# Adding an explicit reference to the symbol ensures it's
+# included in the statically linked binary.
+{% if flag?(:musl) && flag?(:static) %}
+  lib LibC
+    fun pthread_once(Void*, Void*)
+  end
+
+  fun __crystal_static_musl_workaround
+    LibC.pthread_once(nil, nil)
+  end
+{% end %}


bcardiff:
I smoke tested the build process and shards binary without the hack on 64/32 and everything seems to work fine.

j8r:
Should we now create a new issue to track the empty strack traces problem for static linking against musl?

waj:
@j8r sure, I'll do it. It's a totally unrelated issue. I've been doing some research and I have a possible solution already.

j8r:
That's a really great work @waj , thanks a lot.

I didn't see Crystal reaching 1.0 without this essential issues solved, static linking is a very good selling point for the language: easy application distribution.
Go has it too, but it is not as easy to link statically. Last time, I had to set `LDFLAGS='-extldflags "-static -fuse-ld=bfd"'` (?!).

rdp:
Should this also be reported upstream somewhere?  Just wondering...

waj:
@j8r well... I didn't open the issue, but just sent a PR instead: #9267 😉

waj:
@rdp I'm not sure where the bug actually is. Anyway, I have a really small example in plain C to reproduce the issue, so I might send it to Alpine.

PR #9244, opened 2020-05-07
Add Experimental annotation and doc label

bcardiff:
The annotation can be used without args or with a single string argument (similar to Deprecated).

There is no semantics given to the annotation.

Eventually we can have a compiler flag that disables experimental (and maybe deprecated) features to stay in a strict version of the language/std-lib/shard.

For now is a way to document unstable features.

### DIFF
--- a/spec/compiler/codegen/experimental_spec.cr
+++ b/spec/compiler/codegen/experimental_spec.cr
@@ -1,0 +1,50 @@
+require "../spec_helper"
+
+describe "Code gen: experimental" do
+  it "compiles with no argument" do
+    run(%(
+      @[Experimental]
+      def foo
+      end
+
+      2
+      )).to_i.should eq(2)
+  end
+
+  it "compiles with single string argument" do
+    run(%(
+      @[Experimental("lorem ipsum")]
+      def foo
+      end
+
+      2
+      )).to_i.should eq(2)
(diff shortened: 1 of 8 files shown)


RX14:
What do you have this in mind for?

bcardiff:
When new modules and features in the std-lib is introduced, it will be better to signal them as experimental/subject-to-change. That way we can iterate. We could also use them for 1.0 to signal some other parts of the std-lib we might feel there are not as polished, if any.

Ruby, Nim, Kotlin advertise things as experimental. Then we can decide, without breaking the semver contract, whether the feature needs to be iterated or dropped.

I also envision in the future a way to stay on a non-experimental non-deprecated subset of the language/std-lib/shards if desired.

Although this PR will not serve for marking experimental language it serves as a starting point to the idea of experimental features as a whole.

straight-shoota:
Effectively, we have both in the past and present treated features as experimental. It seems like a good idea to formalize that internal knowledge.
We should have a separate discussion about what should be designated as experimantal, but there are definitely some possible candidates in current stdlib.

RX14:
I like the idea, I was just wondering if you had anything specific in mind which bought this feature to the top of your todo list. Especially interested in the pre-1.0 actions.

PR #9248, opened 2020-05-08
Skip no closure check for non-Crystal procs

### LINKED ISSUE #7279: Module validation failed: Call parameter type does not match function signature!
Crystal 0.27.0 (2018-11-04)
LLVM: 6.0.1

Attempting to compile this code produces the following error:

```cr
lib LibFoo
  struct S
    callback : ->
  end
end

s = LibFoo::S.new
s.callback = nil
```

```
Module validation failed: Call parameter type does not match function signature!
%Nil ()* null
 %"->" = type { i8*, i8* }  %71 = call i8* @"~check_proc_is_not_closure"(%Nil ()* null), !dbg !74
 (Exception)
  from __crystal_main
  from main
```

jhass:
They can't have a closure anyways and more importantly are represented as a
single pointer. So the check could not even deconstruct them into two
pointers. This previously generated invalid code trying to pass the
function pointer type to the check function expecting the Crystal proc type

fixes #7279

### DIFF
--- a/spec/compiler/codegen/closure_spec.cr
+++ b/spec/compiler/codegen/closure_spec.cr
@@ -697,4 +697,35 @@ describe "Code gen: closure" do
       ))
     end
   end
+
+  it "allows passing an external function along" do
+    codegen(%(
+      lib LibC
+        fun exit(c : Int32) : NoReturn
+      end
+
+      def raise(a) : NoReturn
+        LibC.exit(1)
+      end
+
+      lib LibA
+        fun a(a : Void* -> Void*)
+      end
+
+      fun b(a : Void* -> Void*)
+        LibA.a(a)
+      end
+    ))
+
+    codegen(%(
+      lib LibFoo
+        struct S
+          callback : ->
+        end
+      end
+
+      s = LibFoo::S.new
+      s.callback = nil
+    ))
+  end
 end
--- a/spec/compiler/codegen/closure_spec.cr
+++ b/spec/compiler/codegen/closure_spec.cr
@@ -681,4 +681,20 @@ describe "Code gen: closure" do
       f1.call &+ f2.call
     ))
   end
+
+  it "ensures it can raise from the closure check" do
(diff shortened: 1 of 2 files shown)


src/compiler/crystal/codegen/codegen.cr
```diff
@@ -1530,10 +1531,13 @@ module Crystal
...
     def check_proc_is_not_closure(value, type)
-      check_fun_name = "~check_proc_is_not_closure"
-      func = @main_mod.functions[check_fun_name]? || create_check_proc_is_not_closure_fun(check_fun_name)
-      func = check_main_fun check_fun_name, func
-      value = call func, [value] of LLVM::Value
+      if value.type == llvm_typer.proc_type
+        check_fun_name = "~check_proc_is_not_closure"
+        func = @main_mod.functions[check_fun_name]? || create_check_proc_is_not_closure_fun(check_fun_name)
+        func = check_main_fun check_fun_name, func
+        proc_value = bit_cast value, llvm_typer.proc_type
```
RX14, line 1538: Is this extra bit case needed if we already checked `value.type`?

spec/compiler/codegen/closure_spec.cr
```diff
@@ -681,4 +681,51 @@ describe "Code gen: closure" do
       f1.call &+ f2.call
     ))
   end
+
+  it "ensures it can raise from the closure check" do
+    expect_raises(Exception, /NoReturn/) do
```
straight-shoota, line 686: Can you compare with the entire expected error message here? It helps so much understanding the purpose of the spec.
Also Regex is unnecessary, String is fine.

src/compiler/crystal/codegen/codegen.cr
jhass: Yup, sorry. Totally missed that out from an earlier experiment, good catch.

jhass:
Thanks! Would love to see (solutions for) #9253 and #9255 in 0.35 too <3

PR #9250, opened 2020-05-08
Fix version selector in API docs shows current version twice

bcardiff:
#9187 didn't completely fix the issue.

We can name this: The return of the fix 😅


straight-shoota:
@jhass This isn't even one of the really weird idiosyncrasies of JavaScript. Just got mixed up in writing JavaScript between Crystal/Ruby syntax 😆

Issue #9226, opened 2020-05-04
GC related crash using C libraries integration with GTK


hugopl:
Hi,

I'm experiencing a cash that may be a problem in Crystal or in GTK integration with Crystal GC. The original problem was reported at https://github.com/jhass/crystal-gobject/issues/57, and I tried to reduce the issue to something that could be reproduced with a crystal build command without any dependencies.

## The issue

Code bellow just call GTK3 C functions directly, GTK3 doesn't call any Crystal code, but the process crash inside GTK when we call `LibGtkSource.gtk_source_buffer_set_language(buffer, lang)`. A C version of this program works and can be checked out in the original bug report.

Now the interesting part, if we run the program with GC disabled the crash doesn't happen.

There's an stack overflow page about something that seems related to this https://stackoverflow.com/q/43141659/2199687. So is this an issue on Crystal GC? GSourceView4? GLib?

## Crash environment

Crash tested on ArchLinux x86_64 with packages:
 * crystal 0.34.0-2
 * gc 8.0.4-3
 * llvm-libs-10.0.0-1
 * gtk3 1:3.24.20-1
 * gtksourceview4 4.6.0-1

## Crashing code
```Crystal
@[Link("gtk-3")]
lib LibGtk
  fun gtk_init(argc : Int32*, argv : UInt8***) : Void
  fun gtk_main : Void
  fun gtk_builder_new_from_string(string : UInt8*, length : Int64) : Void*
  fun gtk_builder_connect_signals(this : Void*, user_data : Void*) : Void
  fun gtk_builder_get_object(this : Void*, name : UInt8*) : Void*
  fun gtk_widget_show_all(this : Void*) : Void
  fun gtk_text_view_get_buffer(this : Void*) : Void*
end

@[Link("gtksourceview-4")]
lib LibGtkSource
  fun init = gtk_source_init : Void
  fun gtk_source_language_manager_get_default : Void*
  fun gtk_source_language_manager_guess_language(this : Void*, filename : UInt8*, content_type : UInt8*) : Void*
  fun gtk_source_buffer_set_language(this : Void*, language : Void*) : Void
end

xml = <<-EOF
<?xml version="1.0" encoding="UTF-8"?>
<!-- Generated with glade 3.22.2 -->
<interface>
  <requires lib="gtk+" version="3.20"/>
  <requires lib="gtksourceview" version="4.0"/>
  <object class="GtkApplicationWindow" id="main_window">
    <property name="can_focus">False</property>
[...]

PR #9253, opened 2020-05-08
Fix passing a function pointer to a proc argument

jhass:
Couldn't find an existing issue for this :)

Not sure this is the best place to do this conversion in the code, it does feel a bit out of place.

Maybe it's just a mistake that function pointers are typed as Proc and they should have their own Crystal type. Right now it's a bit weird that Proc's are sometimes generated as function pointers in LLVM and sometimes as `{i8*, i8*}`. I feel like this basically the origin of these bugs.

### DIFF
--- a/src/compiler/crystal/codegen/cast.cr
+++ b/src/compiler/crystal/codegen/cast.cr
@@ -549,6 +549,9 @@ class Crystal::CodeGenVisitor
 
     if to_type != from_type
       value = upcast_distinct(value, to_type, from_type)
+    elsif to_type.is_a?(ProcInstanceType) && value.type != llvm_typer.proc_type
+      value = bit_cast(value, llvm_context.void_pointer)
+      value = make_fun(to_type, value, llvm_context.void_pointer.null)
     end
     value
   end
--- a/spec/compiler/codegen/cast_spec.cr
+++ b/spec/compiler/codegen/cast_spec.cr
@@ -411,4 +411,12 @@ describe "Code gen: cast" do
       x.foo
       )).to_i.should eq(1)
   end
+
+  it "casts from function pointer to proc" do
+    codegen(%(
+      fun a(a : Void* -> Void*)
+        Pointer(Proc((Void* -> Void*), Void*)).new(0_u64).value.call(a)
+      end
+    ))
+  end
 end


src/compiler/crystal/codegen/cast.cr
```diff
@@ -549,6 +549,9 @@ class Crystal::CodeGenVisitor
 
     if to_type != from_type
       value = upcast_distinct(value, to_type, from_type)
+    elsif to_type.is_a?(ProcInstanceType) && value.type != llvm_typer.proc_type
+      value = bit_cast(value, llvm_context.void_pointer)
+      value = make_fun(to_type, value, llvm_context.void_pointer.null)
```
asterite, line 554: You should instead define an overload of `upcast_distinct` to handle this case. 

Also, what's the type of `value` here? Why is it not a `ProcInstanceType`? `a` looks like a `ProcInstanceType` to me in the spec. What am I missing?

jhass: 1. The overload would not be hit because in this case `to_type == from_type`.
2. Ignoring one, `def upcast_distinct(value, to_type : ProcInstanceType, from_type : ProcInstanceType)` would be a bit weird, no?
3. The spec defines a `fun`, so it's `i8* (i8*)*` rather than `%"->"`.
4. #9248 has essentially the same check on `value.type`.

asterite: I think then the type of the argument should be `%"->"`. So this is not the right fix.

And doing so, would it also fix the other PR?

asterite: What I mean, the outside argument should be a pointer, but inside the fun it should be a Proc, because for Crystal that's a proc.

jhass: No. `fun` definitions should get exported and follow C ABI. Generating this as `%"->"` will make it impossible to provide certain interfaces that third party libraries may expect.

jhass: The effect of this is exactly this. What's the better place to do this then?

jhass: Actually I still believe we should have presentation of a function pointer that is not `Proc`. 

See this usecase: https://gist.github.com/jhass/53b56308b1d4013c32cf2325d800e130#file-malloc_pthread_shim-cr-L185-L197

asterite: No. We will never have two things to say "a function pointer". `Proc` can represent both: if it doesn't have closure data then it's a C function pointer, and also a Crystal `Proc`.

> The effect of this is exactly that. What's the better place to do this then?

When we codegen the body of the `fun`, we copy the argument to a local variable that has a `Proc` type. Every other reference to the argument will point to that local variable. The "function pointer" no longer exists and the codegen only has to deal with `Proc`.

The way it's done in this PR, everytime you access that argument it needs to be cast to a `Proc`. My proposal is doing it once at the header of the function.

We already do some special things with arguments, because they can't be modified, so we already copy them to a local variable. Here we just need to say "is it a C function with a Proc arg type? Then the local copy is a Proc and we create it".

jhass: > We will never have two things to say "a function pointer". Proc can represent both: if it doesn't have closure data then it's a C function pointer, 

Did you study the usecase in detail? That the proc has no closure data is only known at runtime. How can `Proc#call` generate the right `call` in this case? Note it's the argument to `Proc#call`, another `Proc`  that would need  to switch types depending on whether it has a closure or not.

jhass: Mmmh, isn't #9255 doing what you suggest actually? I just tried, it doesn't fix this issue.

asterite: Whenever you pass a Proc to a `fun`, the compiler checks at runtime, or if possible at compile-time, that it's not a closure. Is that what you mean with "How can ..."?

Yeah, #9255 seems to be what I was suggesting. It's strange that it doesn't fix this particular case.

jhass: I also checked whether it would cover #9248, which it doesn't. I can't say I fully understand why it doesn't, but I take this as strong datapoints to my general argument: A distinct type is required to properly handle and cover all these special cases where you need to convert between the two... see I naturally want to say types.

jhass: > Whenever you pass a Proc to a fun, the compiler checks at runtime, or if possible at compile-time, that it's not a closure. Is that what you mean with "How can ..."?

I feel like you still didn't get the usecase I linked above. At times I want to statically declare that a function pointer can only take a function pointer, and not a Proc, as one of its arguments. Currently it's impossible to express this in Crystal.

Discussing this I realized a better workaround is to just type the proc argument as `Void*` for now and hack my way to a `Void*` of the function pointer (which still requires #9255), so I updated gist to do this and updated the link above to the old revision. I still would strongly argue that Crystal should allow to do this in a type safe way.

jhass: So, where does that leave us?

asterite: Out of curiosity: why do you need a `fun` doing this?

I think we added `fun` in Crystal so we can have `__crystal_malloc` and so on. We never though writing a lot of crystal code, not even using Proc inside `fun`. I agree that the current language allows that, but it was never a use case we had in mind.

So I see all these function pointer <-> proc issues really minor things, because %99.99 of the users won't have to deal with them. So I'd like to know why you need to do this.

asterite: Put another way: I don't see a reason to worry about this before 1.0.

jhass: @asterite because of https://github.com/crystal-lang/crystal/issues/9226#issuecomment-626710784

The three patches I submitted are certainly not a fix to the design issues around this stuff, but they're the minimum to get this workaround working.

In other words they're the minimum to workaround the GC messing with memory allocated in threads spawned  in shared libraries bound to Crystal programs. We've just been extremely lucky so far that nobody built some more complex application with a library doing that.

In simple terms, if you want to see GTK apps build with Crystal 1.0, we need this.

asterite: I have a fix.

asterite: Here: https://github.com/crystal-lang/crystal/pull/9287 :-)

asterite: I think all the issues with function pointers inside C funs were that we weren't immediately casting them to Procs. Once I did that, everything works fine.

The other issue assigning `nil` to a struct proc/pointer type is different: it's a special case. It was already handled by the compiler, but in a wrong way.

After this, do you think there are still issues with this approach?

jhass: It does fix my immediate problems, thank you! So I will close this one in favor.

But I do maintain there's a design issue in doing that and that function pointers should get their own type. I made this point above extensively above I think to reiterate:

Without given the user a way to explicitly declare a function pointer, there's no type safe way to invoke a function pointer to an external (non-Crystal) function, that takes another function pointer (a single pointer, not one pointer for the function and one for the closure). It's impossible to correctly generate, the compiler simply lacks the information to do so because the user has no way of expressing that information in the language.

jhass: A smaller point for a distinct type for function pointers could probably by made by performance: The only reason we copy function pointers to procs now is that we have no other way of expressing them in the current type system. If we had it, we could entirely avoid allocating a `Proc` on the stack for this purpose, resulting in smaller stack frames.

jhass:
Superseded by #9287

src/compiler/crystal/codegen/cast.cr
asterite: > It's impossible to correctly generate, the compiler simply lacks the information to do so because the user has no way of expressing that information in the language.

This is almost true. You can express a function pointer by representing it as `Void*`. Then there's `Proc#pointer` which gives you a `Void*` to the function, without the closure. The problem is, this is not type-safe.

I think having two types of `Proc` might be confusing for users. We also need some kind of syntax for this, semantic, compiler code. All for dealing with C functions, which we can already do, just in a slightly less type-safe way. But since the boundary between Crystal and C is so thin, maybe the current approach is good enough.

jhass: Yes, I realized that by now, that's why I always included "no type safe way". 

I don't think we need a new syntax, just a new primitive type.  Writing `FunctionPointer(Void*, Void*)` is fine for the cases  where you need it and  we could optionally let the existing proc syntax return that in  unambiguous contexts, such as `fun` argument definitions and perhaps even lib-structs.

PR #9254, opened 2020-05-08
Rename CLI argument to json-config-url

straight-shoota:
All other CLI arguments use kebab-case, this keeps it consistent.

That argument has not been released yet, so we don't need to keep the old spelling as a fallback.


bcardiff:
BTW https://crystal-lang.org/api/versions.json is uploaded already, we can change the doc generation in the CI to start using that.

Would the docs work still for local usages if they were built with --json-config-url, right?

straight-shoota:
> Would the docs work still for local usages if they were built with --json-config-url, right?

Of course.

PR #9255, opened 2020-05-08
Fix taking a pointer of a function pointer

jhass:
Same as the previous one, not sure if this the right way and place, but it doesn't seem to break anything :shrug:


jhass:
Superseded by #9287

PR #9257, opened 2020-05-09
Implement File#fsync on Windows

kubo:
This PR uses [`_commit`][] to implement `File#fsync` on Windows.

Metadata is also flushed as [`fsync`][].
* `_commit` calls [`FlushFileBuffers`][] internally. (I checked it by looking the source code of crt  installed with MSVC.)
* In [`CreateFileW`][] document:
  > To ensure that the metadata is flushed to disk, use the FlushFileBuffers function.

[`_commit`]: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/commit
[`fsync`]: https://linux.die.net/man/2/fsync
[`FlushFileBuffers`]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-flushfilebuffers
[`CreateFileW`]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew

### DIFF
--- a/src/crystal/system/win32/file.cr
+++ b/src/crystal/system/win32/file.cr
@@ -241,6 +241,8 @@ module Crystal::System::File
   end
 
   private def system_fsync(flush_metadata = true) : Nil
-    raise NotImplementedError.new("File#fsync")
+    if LibC._commit(fd) != 0
+      raise IO::Error.from_errno("Error syncing file")
+    end
   end
 end
--- a/src/file.cr
+++ b/src/file.cr
@@ -771,6 +771,9 @@ class File < IO::FileDescriptor
   # then the syscall *fdatasync* will be used and only data required for
   # subsequent data retrieval is flushed. Metadata such as modified time and
   # access time is not written.
+  #
+  # NOTE: Metadata is flushed even when *flush_metadata* is false on Windows
+  # and DragonFly BSD.
   def fsync(flush_metadata = true) : Nil
     flush
     system_fsync(flush_metadata)
--- a/src/lib_c/x86_64-windows-msvc/c/io.cr
+++ b/src/lib_c/x86_64-windows-msvc/c/io.cr
@@ -15,4 +15,5 @@ lib LibC
   fun _get_osfhandle(fd : Int) : IntPtrT
   fun _pipe(pfds : Int*, psize : UInt, textmode : Int) : Int
   fun _dup2(fd1 : Int, fd2 : Int) : Int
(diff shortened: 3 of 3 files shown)


kubo:
This issue was closed by mistake. I removed kubo:file-fsync-on-windows branch on my repository and it made this PR closed.

bcardiff:
Why is it better to use _commit instead of the public API FlushFileBuffers ?

kubo:
> Why is it better to use _commit instead of the public API FlushFileBuffers ?

Does this sentence imply that [`_commit`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/commit) is not a public API?

bcardiff:
I will rephrase. Why is _commit preferred instead of FlushFileBuffers? Since CreateFileW is used to create the handle and FlushFileBuffers is the suggested function to use, it seems strange to me to use _commit.

Does it do something else that is needed or they are aliases?

kubo:
I uses `_commit` because I think that crystal uses [C-runtime low-level I/O](https://docs.microsoft.com/en-us/cpp/c-runtime-library/low-level-i-o) mainly and uses [Win32 API](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/) only when low-level I/O lacks ability. For example, `File.open` and `File.delete` use low-level I/O (`_wopen` and `_wunlink`), not Win32 API (`CreateFileW` and `DeleteFileW`). `File.rename` doesn't use low-level I/O because `_rename` lacks the ability to replace existing files. Well, this is just a guess. I may think too much.

I have no other reasons to use `_commit`. I feel no difference with the following code except error messages (based on errno or based on win32 error code) when the method fails.
```crystal
  private def system_fsync(flush_metadata = true) : Nil
    if LibC.FlushFileBuffers(windows_handle) == 0
      raise IO::Error.from_winerror("Error syncing file")
    end
  end
```

> Does it do something else that is needed or they are aliases?

`_commit(fd)` is same with `FlushFileBuffers((HANDLE*)_get_osfhandle(fd))` when error checking is omitted. As for the above crystal code, `_get_osfhandle(fd)` is called inside of `windows_handle`.

bcardiff:
Ok, thanks for the explanation 🙇

RX14:
It's intended that we migrate from the "half-posix" file API to the full windows API. I used the unix-like wrappers only because of the ease of storing a fd instead of a `HANDLE*`, but this all needs refactoring.

PR #9258, opened 2020-05-09
Let `case when` be non-exhaustive, introduce `case in` as exhaustive

asterite:
## Overview

This PR makes `case ... when` behave like it was before:
- exhaustiveness is not checked
- `else` can be omitted, by default it's `nil`

This make Crystal behave exactly like Ruby in that sense.

Then, `case ... in` is introduced as the exhaustive variant.

For example:

```crystal
# Compiles! Totally fine
case 1 || "a"
when Int32
end

# Error: missing case String
case 1 || "a"
in Int32
end
```

`case .. in`, from now on called "exhaustive case" is a bit different than `case .. when`:
- the expression that comes after `case` is mandatory (in the regular case you can omit it and it behaves like an if-elsif-else)
- the expressions that you can put after `in` are limited to:
  - nil literal
  - bool literals
  - constants that refer to types (Foo::Bar)
  - constants that refer to enum members (Color::Red)
  - generics
  - question methods like `.red?` (for enum members)
- an `else` can't be used: because it's exhaustive, you should list all the cases you expect to handle. If you need an `else` you can use `case ... when`.

## Experimental

The exhaustive case will be released in version 1.0 as experimental. That means that we could, in the future, either remove it, or replace its syntax, or change its behavior. It's very unlikely (I don't see why we would do that), but that's what we'll do.

## Wait... Ruby has `case ... in` too!

We know, but this `case ... in` is much more limited than the Ruby version.

In Ruby you can do:

```crystal
case array
in [1, 2]
in [1, 2, a]
  puts a
end
```

That will pattern match `array` against an array of values in the first case, and an array of two values and a third value that's captured into a variable.

None of that is possible right now.

Will it be possible in the future? We didn't even think about it. Maybe it would be nice. But note that if we really wanted to do that, we need to allow an `else` clause because there's no way we can know, at compile time, all the shapes of all arrays.

Also, this feature is experimental in Ruby, so who knows what will happen to it.

## Breaking change?

Yes, this is a small breaking change because `in` now needs to be a keyword that signals the end of an expression. That way the compiler knows that when it finds `in`, it knows the previous `in` ends. Same reason why `when` is a keyword and you can't use it as a variable name.

I first thought "Wait, did Ruby made a breaking change with this?". The answer is no: Ruby has the `for x in y` syntax so `in` was already a keyword. In Crystal I think it's used in macros but it was never a proper keyword. Now it is. This PR includes a fix for three such usages.

I believe making this breaking change is fine and it should be easy to upgrade. Also, `in` is a keyword in a lot of languages so having it as a keyword might enable us to use it in more places later on.

## Thoughts

I personally think having these two syntaxes will be great! If you are casing over a big enum, or just casing over some chars, strings, numbers, whatever, you can use the good old `case ... when`. But when you want to make sure you cover all types, or all enum members, you can use `case ... in` and get a bit more type safety, and also avoid having to write `else raise "can't happen"`.

### DIFF
--- a/spec/compiler/formatter/formatter_spec.cr
+++ b/spec/compiler/formatter/formatter_spec.cr
@@ -542,6 +542,8 @@ describe Crystal::Formatter do
   assert_format "case\nend"
   assert_format "case\nelse\n  1\nend"
 
+  assert_format "case  1 \n in Int32 \n 3 \n end", "case 1\nin Int32\n  3\nend"
+
   assert_format <<-CODE
     case 0
     when 0 then 1; 2
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -1990,7 +1990,7 @@ module Crystal
     end
 
     describe "case methods" do
-      case_node = Case.new(1.int32, [When.new([2.int32, 3.int32] of ASTNode, 4.int32)], 5.int32)
+      case_node = Case.new(1.int32, [When.new([2.int32, 3.int32] of ASTNode, 4.int32)], 5.int32, exhaustive: false)
 
       it "executes cond" do
         assert_macro "x", %({{x.cond}}), [case_node] of ASTNode, "1"
--- a/spec/compiler/macro/macro_methods_spec.cr
+++ b/spec/compiler/macro/macro_methods_spec.cr
@@ -1990,26 +1990,36 @@ module Crystal
     end
 
     describe "case methods" do
-      case_node = Case.new(1.int32, [When.new([2.int32, 3.int32] of ASTNode, 4.int32)], 5.int32, exhaustive: false)
+      describe "when" do
+        case_node = Case.new(1.int32, [When.new([2.int32, 3.int32] of ASTNode, 4.int32)], 5.int32, exhaustive: false)
 
-      it "executes cond" do
-        assert_macro "x", %({{x.cond}}), [case_node] of ASTNode, "1"
+        it "executes cond" do
+          assert_macro "x", %({{x.cond}}), [case_node] of ASTNode, "1"
+        end
+
+        it "executes whens" do
+          assert_macro "x", %({{x.whens}}), [case_node] of ASTNode, "[when 2, 3\n  4\n]"
+        end
+
+        it "executes when conds" do
+          assert_macro "x", %({{x.whens[0].conds}}), [case_node] of ASTNode, "[2, 3]"
+        end
+
+        it "executes when body" do
+          assert_macro "x", %({{x.whens[0].body}}), [case_node] of ASTNode, "4"
+        end
+
+        it "executes else" do
+          assert_macro "x", %({{x.else}}), [case_node] of ASTNode, "5"
+        end
       end
 
-      it "executes whens" do
-        assert_macro "x", %({{x.whens}}), [case_node] of ASTNode, "[when 2, 3\n  4\n]"
-      end
+      describe "in" do
+        case_node = Case.new(1.int32, [When.new([2.int32, 3.int32] of ASTNode, 4.int32)], 5.int32, exhaustive: true)
 
-      it "executes when conds" do
-        assert_macro "x", %({{x.whens[0].conds}}), [case_node] of ASTNode, "[2, 3]"
-      end
-
-      it "executes when body" do
-        assert_macro "x", %({{x.whens[0].body}}), [case_node] of ASTNode, "4"
-      end
-
-      it "executes else" do
-        assert_macro "x", %({{x.else}}), [case_node] of ASTNode, "5"
+        it "executes whens" do
+          assert_macro "x", %({{x.whens}}), [case_node] of ASTNode, "[in 2, 3\n  4\n]"
+        end
       end
     end
 
--- a/spec/compiler/normalize/case_spec.cr
+++ b/spec/compiler/normalize/case_spec.cr
@@ -6,103 +6,103 @@ describe "Normalize: case" do
   end
 
   it "normalizes case with var in cond" do
-    assert_expand_second "x = 1; case x; when 1; 'b'; else; end", "if 1 === x\n  'b'\nend"
+    assert_expand_second "x = 1; case x; when 1; 'b'; end", "if 1 === x\n  'b'\nend"
   end
 
   it "normalizes case with Path to is_a?" do
-    assert_expand_second "x = 1; case x; when Foo; 'b'; else; end", "if x.is_a?(Foo)\n  'b'\nend"
+    assert_expand_second "x = 1; case x; when Foo; 'b'; end", "if x.is_a?(Foo)\n  'b'\nend"
   end
 
   it "normalizes case with generic to is_a?" do
-    assert_expand_second "x = 1; case x; when Foo(T); 'b'; else; end", "if x.is_a?(Foo(T))\n  'b'\nend"
+    assert_expand_second "x = 1; case x; when Foo(T); 'b'; end", "if x.is_a?(Foo(T))\n  'b'\nend"
   end
 
   it "normalizes case with Path.class to is_a?" do
-    assert_expand_second "x = 1; case x; when Foo.class; 'b'; else; end", "if x.is_a?(Foo.class)\n  'b'\nend"
+    assert_expand_second "x = 1; case x; when Foo.class; 'b'; end", "if x.is_a?(Foo.class)\n  'b'\nend"
   end
 
   it "normalizes case with Generic.class to is_a?" do
-    assert_expand_second "x = 1; case x; when Foo(T).class; 'b'; else; end", "if x.is_a?(Foo(T).class)\n  'b'\nend"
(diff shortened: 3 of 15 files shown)


j8r:
Oh, I thought initially it was like shell's `case var in ...`.
I suppose `when` and `in` can't be mixed inside a same `case` expression.

Another thing, can class types be put (`FooBar.class`)?

asterite:
> I suppose when and in can't be mixed inside a same case expression

Correct, there's a spec for that.

> Another thing, can class types be put (FooBar.class)?

Yes, there are specs for that.

asterite:
> Another thing, can class types be put (FooBar.class)?

And in fact, that wasn't handled before this PR, I just noticed it when translating `when` to `in`. So great question! 😊

RX14:
I think `case in` should have the same semantics as the current `case when`, not be restricted to only things it can prove.

I want to use `case in` always and I want the compiler to *force me* to use `else`. I don't want to have to switch back and forth and think about when I want to be exhaustive or not: I always want to be exhaustive.

asterite:
> I want to use case in always and I want the compiler to force me to use else. I don't want to have to switch back and forth and think about when I want to be exhaustive or not: I always want to be exhaustive.

I understand that, but it just doesn't make sense to check exhaustiveness all the time, even when there's no way to prove exhaustiveness.

For example in cases like this:

```crystal
def foo
  case value
  when 1
    return "early return 1"
  when 2
    if some_condition
      return "early return 2"
    end
  end

  common_logic
end
```

it's just annoying to have to put an `else # go on` for no reason. It's pretty clear that the `case` there is meant to group a series of early returns. This happens all over the place in the compiler, and I'm sure the complaints about this change were around these kind of things. Or just casing over a value in special cases, but then going on normally with the code.

Then casing over keyboard events in SDL: there's really no need to have the compiler check that you covered all keys, you are usually interested in a few keys.

Or casing over http statuses: do you really want to cover all of them? Of course you might need an `else` in that case, but maybe not all of the time.

Like that, there are many cases where you just want to case over a value to avoid having to write it over and over in an `if/elsif`.

The cases where you want exhaustiveness are very limited:
- you either want to cover all types in a union (or eventually in a sealed type hierarchy, I hope)
- you either want to cover all enum members

In the compiler's source code, and the standard library, my guess is that maybe 10% will be changed to use `case ... in` and the rest will remain `case ... when`. So we shouldn't have a single feature that covers both cases, where the default is a benefit for the minority of the cases.

That said, ameba could always be enhanced with a lint rule to force `else` on a `case ... when`. Then you can make sure, either in CI or somewhere else, that you always add an `else`.

Finally, when the compiler asks you to add an `else` for a case that can't be proved to be exhaustive no matter what, it reduces a bit the joy of programming with Crystal. People think (and I think it too) "Why do I need to add this useless thing with just an empty content or a dummy comment?". The idea of Crystal is to have fun too while programming, and this is not fun. But you still get type safety in the cases you need it, with `case ... in`. So we get both: fun, and strictness when needed (which is not always needed).

RX14:
> it's just annoying to have to put an `else # go on` for no reason.

If that's your opinion, use `when`. I don't share it, so I'd like to use `in` all the time.

It seems like an arbitrary restriction to me, is all.

asterite:
@RX14 You **can** use `in` all the time. Then one of two things will happen:
- it will compile just fine. You get type safety, and if you add a new type or enum member the compiler will complain.
- the compiler will complain saying that a pattern can't be proved exhaustively. For example if you do `in 1` or `in "foo"`. In that case, add an `else` and change all then `in` to `when`. You don't get type safety, but **you didn't get it in the first place** because there was nothing there to prove or check.

straight-shoota:
There might be a good argument for `else` on exhaustive case: When you case over a value that can technically be exhaustive (union, enum) but doesn't need to consider all individual cases at the moment (but maybe later). IMO that might be benefitial to use exhaustive case with `else`. So it is still conceptually an exhaustive case and keeps the same restrictions for data types but you can avoid having to handle each individual value and instead have `else`.

asterite:
@RX14 Again, I understand that you don't like this direction, but then @ysbaddaden doesn't like the current situation. Both of you are core team members. What do we do? Maybe we should just revert to the old `case` behavior when exhaustiveness wasn't a thing, and both of you were happy.

Then let's go back in time to that happy moment. Forget about all the paths that we've taken so far. A new PR arrives where now you can have a case be exhaustively checked, but you can also have the regular non-exhaustive way. You keep using the usual `case ... when`, you were both happy with that, right? Except that now people that want to exhaustively check things can also be happy. Everyone can be happy!

I'm sorry, but there's no way we can please everyone. Actually, there is: we provide two options here, if you want exhaustiveness check you can do it (where it makes sense!), and when you don't need it, you can do it too. I still can't see how that is not an ideal scenario.

> There might be a good argument for else on exhaustive case

Can you provide an example? If later on you want to check all cases, just change `when` to `in`.

RX14:
I quite like the current exhaustive `case` semantics. I like that crystal gives you choice: you can take ruby semantics with `when` or exhaustive semantics with `in`. This makes everyone happy! But this PR implements only a *subset* of the current `case` semantics with `in`. There's no need for that. If you implement both the old and the new semantics fully, then everyone is happy.

straight-shoota:
@RX14 WDYT about going with this PR as a MVP for now. The new exhaustive case is explicitly meant to be experimental and we can iterate on it. As far as I can see, the semantics you enviosion can be added later without breaking anything.
So it seems like a good idea to defer the discussion about wheter exhaustive case needs to be strictly limited to types where exhaustiveness can be proven or it should be more open (with `else`). That discussion shouldn't hold up this PR.

asterite:
@RX14 If I get it right, what you want is this code:

```crystal
x = 1
case x
when 1
when 2
end
```

to give a compile-time error saying to add an `else` clause?

If that's the case, would it be too much to ask to make sure you just add an `else` every time you use case? What's the difficulty of doing that?

If you can do that, everyone else is not forced to do it when it just doesn't make sense that the compiler forces you to add it.

@RX14 @straight-shoota I still don't understand why `in` would allow an expression that can't be proved exhaustively, or why adding an `else` would be good where it just means "disable exhaustiveness check". Then you have an exhaustive case that doesn't check exhaustiveness, whic is nonsense in my mind.

RX14:
@asterite I wouldn't want that to give a compile time error at all, since it's not the "old" `when`  semantics to do so.

What I'd like is for

```cr
x = 1
case x
in 1
in 2
end
```

to give a compile error to add an `else`. I'm not sure there's a simpler way to describe it than "`in` gets the current `case` semantics, and `when` gets the old case semantics".

asterite:
@RX14 There's a good reason not to allow **any** expression in `case ... in`.

Imagine in the future we want to allow this:

```crystal
case exp
in [1, a]
  puts a
else
  puts "No match"
end
```

That is, to make it work like in Ruby where that `a` was actually assigned the value inside the array, in case it was an array.

If we make the `case ...in` in **this PR** allow any expression, then expanding it to new use cases is impossible without a breaking change, because it's perfectly fine to do `when [1, a]` to mean "match against an array with 1 as the first element, and en existing `a` as the second element.

I think restricting `case ... in` to just the minimum current use case is best. That minimum case also excludes an `else` because it's not needed.

asterite:
More arguments: if you look at my original proposal for exhaustive case, I wanted to retain the existing case when behavior, and introduce a new case. The main use case for this new case was making sure all types or all enum members were covered, and that a superfluous raise wasn't needed. And when a new type or enum member is added, to remind the user that a new case must be considered.

The goal was never to make sure that you always add an else clause. Why? Because an else clause is not always needed. In fact, in the compiler all we did was add these superfluous else for no reason. No bug was caught. It was just an annoyance. And it proved to be an annoyance for others too.

RX14:
Oh well, that's a pretty good reason. I just don't like that [this bug](https://github.com/RX14/predict.cr/commit/ff0ce728c469642375dd840f0ce94e3fbaab8362) and it's entire category wouldn't be caught by the compiler after this PR. I would just have kept using `when` and allowing "unknown" errors to pass by undetected if the C library was changed. I like the idea of the compiler forcing you to think about every control path. I thought that it would be possible without tradeoffs if we changed `in` to be equivalent to the old `when` semantics, but seems not.

straight-shoota:
> In fact, in the compiler all we did was add these superfluous else for no reason. 

I wouldn't be so pessimistic about it. In *some* cases it was probably a good idea to have think about the else and add an explicit comment.
Doesn't change the argument here, just wanted to set this straight ;)

didactic-drunk:
While developing 0.35-dev I was forced to add `else` to all my case statements.  0 bugs were found.  Every condition was intended as `else nothing`.  Exhaustive case sounds good but I don't have evidence it improves most code other than enums and classes.   Perhaps those should be the only default exhaustive case requirements?

An exhaustive case with raise/abort seems more generally useful.  I often have `else raise "unknown value"` as @RX14 mentioned.

Also, I  liked `case!`.  It stood out more visually compared to `case in`.

RX14:
I'm :+1: to this PR's semantics now, just to be clear. Seems the best compromise.

asterite:
Yeah, an `else` might be useful. We can start without an else. Later on adding it is not a breaking change so we can do that before or after 1.0 just fine.

PR #9262, opened 2020-05-09
Add HTTP::Cookie::SameSite::None

j8r:
RFC: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00#section-4.2
MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies

Setting `SameSite` to `None` ensures that the browser will send cookies with both cross-site requests and same-site requests, because `Lax` may be set by default on newer browsers.

MDN states that setting `SameSite` to `None` requires the `Secure` attribute. I did find a [Chromium Blog](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html) saying it is for security reasons.

### DIFF
--- a/src/http/cookie.cr
+++ b/src/http/cookie.cr
@@ -5,9 +5,12 @@ module HTTP
   class Cookie
     # Possible values for the `SameSite` cookie as described in the [Same-site Cookies Draft](https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-4.1.1).
     enum SameSite
+      # The browser will send cookies with both cross-site requests and same-site requests.
+      #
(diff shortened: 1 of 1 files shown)


Blacksmoke16:
Bump

j8r:
Recent mozilla blog to better understand the topic: https://hacks.mozilla.org/2020/08/changes-to-samesite-cookie-behavior/.
After Chromium, Firefox will change the default `SameSite` cookie value from `None` to `Lax`.

j8r:
Recent mozilla blog to better understand the topic: https://hacks.mozilla.org/2020/08/changes-to-samesite-cookie-behavior/.
After Chromium, Firefox will change the default `SameSite` cookie attribute from `None` to `Lax`.

PR #9267, opened 2020-05-10
Rework DWARF loading

waj:
In this PR I first separated the code for call stack decoding in different files because it was already too long and confusing.

Then, I changed the way it calculate offsets at runtime for ELF platforms. Instead of using `dladdr` (which turned to be unreliable on musl: https://github.com/bpowers/musl/blob/master/src/ldso/dladdr.c) it now uses `dl_iterate_phdr`. It also has the advantage that the offset is calculated only once, when the metadata is loaded.

This fixes the issue with the empty backtraces in musl when the binary is statically linked.

Related: #4276


src/exception/call_stack/elf.cr
```diff
@@ -0,0 +1,73 @@
+require "crystal/elf"
+
+struct Exception::CallStack
+  protected def self.load_dwarf
+    phdr_callback = LibC::DlPhdrCallback.new do |info, size, data|
+      # HACK: this assumes the first entry is the header for the current program
```
RX14, line 6: Seems like `dl_iterate_phdr` is guaranteed to iterate in the order which ELF objects are loaded, so unless crystal is being used in a shared library, this would be correct.

src/exception/call_stack/elf.cr
```diff
@@ -0,0 +1,73 @@
...
+      end
+    end
+  end
+
+  protected def self.decode_address(ip)
+    ip.address
+  end
+end
+
+lib LibC
```
RX14, line 52: This needs to be moved into the `lib_c` directory for each target. At the least, `DlPhdr` will be architecture dependent between 32/64bits.

src/exception/call_stack/elf.cr
waj: Yes, that’s true, at least it’s in glibc and musl. We should check if that assumption works on BSD too. At first I was using the name, which in glibc is empty for the main program. But musl returns /proc/self/exe.

src/exception/call_stack/elf.cr
waj: Absolutely! I was just too lazy at 2AM 😅

waj: Done. Not only types were different. The order of the struct members change according to the architecture.

PR #9269, opened 2020-05-10
Add `StringLiteral#titleize` macro method

MakeNowJust:
This also adds `SymbolLiteral#titleize` and `MacroId#titleize`.

Other case conversion methods (`upcase`, `camelcase`, etc)  are avialble in macro. So, `titleize` should not be exception.

One use-case is generating error message from type name.


straight-shoota:
There are no specs for MacroId and SymbolLiteral and it seems the implementation is missing for those, too.

MakeNowJust:
`SymbolLiteral` and `MacroId` implementations delegate to `StringLiteral` implementation (see [here](https://github.com/crystal-lang/crystal/blob/8705717eb1bb280744138f9462c47c230e06d782/src/compiler/crystal/macros/methods.cr#L1512-L1514) and [here](https://github.com/crystal-lang/crystal/blob/8705717eb1bb280744138f9462c47c230e06d782/src/compiler/crystal/macros/methods.cr#L1542-L1544)). So, no problem.

And, other methods of `SymbolLiteral` and `MacroId` lacks theirs specs too.

watzon:
Do we have this method on standard strings? If so I must've missed the PR, but if not it would be nice to have it there too.

straight-shoota:
@watzon This PR uses the exact method as implementation:

https://github.com/crystal-lang/crystal/pull/9269/files#diff-a4a8d7f01e581cce938af0cf27f263d9R756

It was added in #9204

watzon:
Yeah I saw that when I was going through the code, I guess I missed the PR. Awesome though, I've wished it was there a number of times.

PR #9270, opened 2020-05-10
`TypeNode` methods to check what "type" the node is

Blacksmoke16:
* Add methods to `TypeNode` to check if its a module, class, or struct
* Adds specs and updates docs for `#nilable?`, `#abstract?` and `#union?`

### DIFF
--- a/src/compiler/crystal/macros.cr
+++ b/src/compiler/crystal/macros.cr
@@ -1810,9 +1810,9 @@ module Crystal::Macros
     #
     # struct Three; end
     #
-    # {{One.stuct?}}   # => false
-    # {{Two.stuct?}}   # => false
-    # {{Three.stuct?}} # => true
+    # {{One.struct?}}   # => false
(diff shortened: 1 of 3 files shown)


src/compiler/crystal/macros.cr
```diff
@@ -1728,24 +1728,99 @@ module Crystal::Macros
...
+    # ```
+    # module One; end
+    #
+    # class Two; end
+    #
+    # struct Three; end
+    #
+    # {{One.stuct?}}   # => false
+    # {{Two.stuct?}}   # => false
+    # {{Three.stuct?}} # => true
```
jhass, line 1815:
```suggestion
    # {{One.struct?}}   # => false
    # {{Two.struct?}}   # => false
    # {{Three.struct?}} # => true
```

Blacksmoke16:
Bump.  Can I get another review on this?

Issue #6663, opened 2018-09-04
Types in annotations are resolved in the namespace of the macro instead of the annotation instance


RX14:
```cr
annotation Ann
end

module UsingAnn
  def method
    {% for ivar in @type.instance_vars %}
      {{ ivar.annotation(::Ann)[0].resolve }}
    {% end %}
  end
end

class Foo
  include UsingAnn

  @[Ann(Bar)]
  getter foo : String?

  module Bar
  end
end

Foo.new.method
```

This code should find the type `Bar` as it's in the namespace where the annotation is used, but the `resolve` call fails because `Bar` is looked up in the `UsingAnn` namespace instead.

Found using this snippet, which fails to compile:

```cr
require "json"

class Foo
  include JSON::Serializable
  
  @[JSON::Field(converter: Converter)]
  getter foo : String
  
  module Converter
    def self.from_json(pull)
      "foo"
    end
  end
end
  
Foo.from_json("")

```

vladfaust:
Can it be related to https://github.com/crystal-lang/crystal/issues/6655?

asterite:
I knew about this problem but didn't say anything. I don't know how it can be fixed. The problem is not just `resolve`, is pasting the type in the macro, which ends up in a different context (`resolve` isn't being used in the macro of `JSON::Serialiazable`).

The other problem is a converter can also be `Time::Format.new("...")`, and `resolve` won't be useful there.

I think annotations were not well thought (my fault, again). Or maybe all expressions in annotations should try to be resolved (but then I have no idea how will they "paste" correctly in macro code).

Maybe `annotation` could define what keys are available and whether they need to be typed or just passed as AST... just throwing ideas here.

The workaround is of course fully qualifying the type name, but it's not nice.

asterite:
@bew Yeah, but that's just adding more on top of the pile of complexity that the language is already. That's not a simple solution.

RX14:
@asterite why not just resolve the full path of types as soon as you use them in annotations so it's always unambiguous. So `@[Ann(Bar)]` will always be recorded as `@[::Ann(::Foo::Bar)]`. That does mean either all types have to be defined before they're used in annotations, or that the compiler needs another pass.

vladfaust:
Is this related to this particular issue?

```crystal
require "json"

struct Enum
  class StringConverter(T)
    def self.from_json(value : JSON::PullParser) : T
      T.parse(value.read_string)
    end

[...]

Issue #6194, opened 2018-06-14
Incorrect scope after including **::Serializable


z64:
Given:
```cr
require "json"
require "json/next/serialization" # (required as I'm using docker nightly; "0.25.0+9")

module Options
end

class Foo
  include JSON::Serializable
  include Options
end

Foo.new
```
Yields:
```
Error in app.cr:9: JSON::Serializable::Options is not a module, it's a annotation

  include Options
          ^~~~~~~
```

Compiles if `Serializable` is included last (or if `Options` is renamed to something not taken):
```cr
require "json"
require "json/next/serialization"

module Options
end

class Foo
  include Options
  include JSON::Serializable
end

Foo.new
```
```
Crystal 0.25.0+9 [bb8968476] (2018-06-14)

LLVM: 4.0.0
Default target: x86_64-unknown-linux-gnu
```
Not fully sure of the scope of the issue (is it with annotations? `macro included`?), this is just the case I ran into this with.

kostya:
this is because exist annotation `JSON::Serializable::Options` it conflict with your module, try:
```crystal
class Foo
  include JSON::Serializable
  include ::Options
end
```

z64:
Yeah I knew that would fix it too, forgot to mention. 

I guess I didn't realize that annotations come along with `include` like that and that the compiler wouldn't skip annotations because I was trying to `include`.

RX14:
Feels to me like we shouldn't be including stuff into people's classes when they include `JSON::Serializable`.

Perhaps a rename to `JSON::SerializableOption`

straight-shoota:
@RX14 I'd question if annotations should be included into other namespaces at all...? I can't think of a use case for this.

straight-shoota:
But yes, removing `Options` from the includeable `JSON::Serializable` namespace would be a quick fix that doesn't need to change the compiler.

`JSON::Serializable::Strict` and `JSON::Serializable::Unmapped` similarly shouldn't be duplicated into namespaces where `JSON::Serializable` is included.

asterite:
Well, if annotations would have been named `OptionAnnotation` we woudn't be having this discussion... 🙄 

(however, it of course can happen with any other type, not just annotations)

Just learn to live with these deficiencies of the language, they are not a big deal.

kostya:
yes, this is just any other modules conflicts, it happens from time to time.

straight-shoota:
[...]

PR #9272, opened 2020-05-11
Deprecate JSON.mapping and YAML.mapping

### LINKED ISSUE #8263: [RFC] Remove JSON.mapping and YAML.mapping
Now that we have `JSON::Serializable` and `YAML::Serializable` I think we should remove `JSON.mapping` and `YAML.mapping` as usual: first deprecate them (warning), then remove them.

I believe the `Serializable` approach is superior:
- it just generates a constructor (`mapping` will generate getters and setters too, from which you can opt-out, but it feels a bit messy)
- it's DRY: with `mapping` you need to specify the type of instance vars, and you have to do this if you have multiple `mapping`s (like `JSON` and `YAML` in a single class)
- it supports inheritance but `mapping` doesn't

I'd like to deprecate `mapping` in 0.32.0 and remove it in 0.33.0.

straight-shoota:
Resolves #8263

### DIFF
--- a/src/json/mapping.cr
+++ b/src/json/mapping.cr
@@ -1,5 +1,4 @@
 module JSON
-  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.")]
   # The `JSON.mapping` macro defines how an object is mapped to JSON.
   #
   # ### Example
@@ -66,10 +65,8 @@ module JSON
   # If *strict* is `true`, unknown properties in the JSON
   # document will raise a parse exception. The default is `false`, so unknown properties
   # are silently ignored.
-  #
-  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.
+  @[Deprecated("use JSON::Serializable instead (the legacy behaviour is also available in a shard at github:crystal-lang/json_mapping.cr)")]
   macro mapping(_properties_, strict = false)
-    {% puts "JSON.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`." %}
     {% for key, value in _properties_ %}
       {% _properties_[key] = {type: value} unless value.is_a?(HashLiteral) || value.is_a?(NamedTupleLiteral) %}
     {% end %}
@@ -229,11 +226,9 @@ module JSON
     end
   end
 
-  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.")]
   # This is a convenience method to allow invoking `JSON.mapping`
   # with named arguments instead of with a hash/named-tuple literal.
-  #
-  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as shard at `github:crystal-lang/json_mapping.cr`.
+  @[Deprecated("use JSON::Serializable instead (the legacy behaviour is also available in a shard at github:crystal-lang/json_mapping.cr)")]
   macro mapping(**_properties_)
     ::JSON.mapping({{_properties_}})
   end
--- a/src/json/mapping.cr
+++ b/src/json/mapping.cr
@@ -1,4 +1,5 @@
 module JSON
+  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.")]
   # The `JSON.mapping` macro defines how an object is mapped to JSON.
   #
   # ### Example
@@ -65,8 +66,10 @@ module JSON
   # If *strict* is `true`, unknown properties in the JSON
   # document will raise a parse exception. The default is `false`, so unknown properties
   # are silently ignored.
-  @[Deprecated("use JSON::Serializable instead (the legacy behaviour is also available in a shard at github:crystal-lang/json_mapping)")]
+  #
+  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.
   macro mapping(_properties_, strict = false)
+    {% puts "JSON.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`." %}
     {% for key, value in _properties_ %}
       {% _properties_[key] = {type: value} unless value.is_a?(HashLiteral) || value.is_a?(NamedTupleLiteral) %}
     {% end %}
@@ -226,9 +229,11 @@ module JSON
     end
   end
 
+  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.")]
   # This is a convenience method to allow invoking `JSON.mapping`
   # with named arguments instead of with a hash/named-tuple literal.
-  @[Deprecated("use JSON::Serializable instead")]
+  #
+  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as shard at `github:crystal-lang/json_mapping`.
   macro mapping(**_properties_)
     ::JSON.mapping({{_properties_}})
   end
--- a/src/json/mapping.cr
+++ b/src/json/mapping.cr
@@ -1,5 +1,5 @@
 module JSON
-  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.")]
+  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.")]
   # The `JSON.mapping` macro defines how an object is mapped to JSON.
   #
   # ### Example
@@ -67,9 +67,9 @@ module JSON
   # document will raise a parse exception. The default is `false`, so unknown properties
   # are silently ignored.
   #
-  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.
+  # DEPRECATED: Use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.
   macro mapping(_properties_, strict = false)
-    {% puts "JSON.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`." %}
+    {% puts "JSON.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`." %}
     {% for key, value in _properties_ %}
       {% _properties_[key] = {type: value} unless value.is_a?(HashLiteral) || value.is_a?(NamedTupleLiteral) %}
     {% end %}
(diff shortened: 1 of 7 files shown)


asterite:
Does a shard for this already exist? We need a safe migration path for those using `mapping`.

straight-shoota:
Not that I'm aware of. But this PR only deprecates the methods, they're still usable.
The deprecation needs to be added before 1.0 or we'll have to carry them along until 2.0.

z64:
Discussion in #8263 noted, but I'll add here, because it's important it's said - I'd strongly advocate for a shard to be published sometime *before* this deprecation is released. Even better if before this PR is merged for nightly consumers.

This API has been around forever, and is used *everywhere* in the web sphere. I am still cleaning up `Time.now` in lots of dependencies, all the time.. and migrating between `JSON.mapping` and `JSON::Serializable` is much more work than s/`Time.now`/`Time.utc/local`. I expect this fallout to be much more prevalent in this space.

Is there any automated tool of doing so in the community, or that could be built? That would also be fantastic.

straight-shoota:
@z64 I doubt anyone would publish a shard before it's deprecated in stdlib. It doesn't hurt to have the deprecation message show up in nightly builds.
And I'm sure we won't turn the deprecations into hard breaking changes soon unless there's an easy migration path.

Since you brought up the comparison, `Serializable` has been around for two years and we've supported both approaches in parallel for that long. The `Time` changes arguably didn't provide an immediate benefit if you were already using the deprecated API, but `Serializable` is strictly better than `mapping` and much more versatile.

z64:
Yes, I agree these changes should be made. But the real problem is, **all the time** I run into shards that are not maintained, or lag behind several versions. I don't have time to fork them all, patch them, or to send PRs everywhere. `Time` I could deal with, but I am very unenthusiastic about doing this for `JSON.mapping`. I still see new code written with `JSON.mapping`.

I can just make my own shard for sure and just deal with it, but I'm just trying to advocate for other downstream consumers of out-of-date shards. If users can immediately start using an "official" shard broken off of std, that would greatly smooth out this process in the ecosystem.

asterite:
Are warnings shown for shards? They shouldn't.

Blacksmoke16:
> If users can immediately start using an "official" shard broken off of std, that would greatly smooth out this process in the ecosystem.

I'd fear that people would just install this new shard with `*.mapping` in it, and just use continue to use it.

asterite:
I remember someone not being happy with `JSON::Serializable` because the default value for a field was copied from the default value in Crystal, but they wanted a different default value when mapping via JSON/YAML. That doesn't exist in `JSON::Serializable` so there's at least one case where people will not upgrade because they can't.

z64:
@asterite You're right, sorry. I just got through replacing a bunch of `Time.now`, I didn't pay attention they were errors and not the local warning anymore.

@Blacksmoke16 Of course. But who is going to do all of this updating? That is the problem I'm running into.

Blacksmoke16:
@asterite I have been working on a more flexible serialization shard.  It's not a drop in replacement, but pretty close for most basic setups.  However, it's not really ready yet.

I think it could be quite useful for when users need more than what the stblib offers.  Especially even having something like `JSON::Serializable` in the stdlib is already quite helpful as I imagine most use cases would be fine using it.

@z64 The owner of the shard ideally.  Otherwise it would fall to the users of the shard, either by PR or forking it.  Unfortunately is the price for using abandoned shards :/

straight-shoota:
@z64 Sorry, but I don't see any value in holding up progressing the stdlib because unmaintained shards are not catching up. There's nothing to to about that, really. If a shard author doesn't continue maintaining it, either someone else picks it up or it's dead.
Shards who can't keep up with such rather trivial API changes are likely lacking in other aspects as well, so depending on them should be discouraged.

z64:
> Shards who can't keep up with such rather trivial API changes are likely lacking in other aspects as well, so depending on them should be discouraged.

Sorry, I don't believe this is fair genralizarion at all. There are plenty of great shards that I've used that at one point or another have fallen in/out of being maintained for one reason or another. I will sooner use a copy of `JSON.mapping`, than to irresponsibly fork something I can't actually maintain, only to fix deprecated code.

> Sorry, but I don't see any value in holding up progressing the stdlib.. 

I am also still not advocating for this not to happen in std. I just want to help make these boring problems in the ecosystem have simple solutions for when things inevitably do break for real, and no one is around to fix them, and you have a job get done..

sam0x17:
@straight-shoota to put things in perspective, I know for Amber the size of the warnings on a default install just for the logger changes goes beyond the default scrollback history size on vscode. So it depends what you mean by usability, but deprecations are not without reverberations of pain throughout the shard community even if it's just a warning message for now.

What @z64 is asking is pretty reasonable -- making sure a shard capturing this functionality (literally as easy as CTRL+Xing the relevant code and extracting it into a shard) is published before releasing the deprecation. This would mean that on day 1 of the next version release when 1000s of shards suddenly have a wall of deprecation warnings, they will have a 3-line recourse by way of adding a shard dependency that fixes it. I think the 15 minutes it takes to do this is well worth the 20*n minutes it is going to take hundreds of individual shards authors to figure out a quick fix for the deprecation warnings. The hedonic calculus here suggests it is clearly worth taking the few minutes of extra time. While you could portray this as "holding up development on the stdlib" think of the amount of time developers like @z64 could spend contributing to the stdlib instead of running around cleaning up deprecation warnings on abandoned packages to get their app(s) to work. We're all in the same community, we all have our own shards and responsibilities to individual projects, so this impacts all of us.

But is this our job? Well, since we have put the burden of having an outdated (but useful) codepath in the stdlib used by 1000s of shards on the shards community, it is our responsibility to provide a reasonable exit strategy. While in practice this may not always be possible, we should at least feel guilty and try to make things as painless as possible. If this was a hard ask, I'd totally agree with you @straight-shoota.

sam0x17:
all of that said I would definitely :+1: muting shard warnings

straight-shoota:
> the warnings on a default install just for the logger changes goes beyond the default scrollback history size on vscode. So it depends what you mean by usability

The way deprecation warnings are currently displayed is out of control. This was never intended like that and needs to be sorted out. I understand that this doesn't help in this context because it still means deprecation messages will go by unnoticed (that's definitely unintendend!), but it's a differnt issue.

> making sure a shard capturing this functionality (literally as easy as CTRL+Xing the relevant code and extracting it into a shard) is published before releasing the deprecation.

I mean this in no way offensive, but literally anyone can do that. If someone publishes a shard and intends to maintain it, I'm sure we could even mention that alternative migration path in the deprecation message.
But I don't think this needs to be the responsibility of the core team. Actually, I'm pretty sure it shouldn't be because it would essentially mean we would still have to maintain both approaches  (one in stdlib, the other in a crystal-lang shard). This PR is explicitly supposed to avoid that.

> on day 1 of the next version release when 1000s of shards suddenly have a wall of deprecation warnings, they will have a 3-line recourse by way of adding a shard dependency that fixes it.

Deprecation messages don't need to be responded to immediately. Such a long-standing feature would pretty certainly not be removed from one version to the next. A deprecation means it's still supported, but will eventually be removed. Maybe it would help to give a time frame on this (and this will definitely be easier post 1.0).

In my opinion, the better way for shard authors to deal with this is to hold off it to take some time to migrate to `Serializable`. It's clearly more future-proof, unless someone is willing to maintain `mapping` indefinitely. But that would also be counterproductive when there's a better alternative in stdlib.

And the migration path to `Serializable` might actually be easier. You don't have to go to `shard.yml` to add another dependency, run `shards update` and add the necessary requires. Instead, `mapping` calls can be directly translated to property macros with annotations. For most use cases, this can even be automated pretty easily.

sam0x17:
@straight-shoota that makes sense

One complicating factor here is for a long time the `JSON.mapping` stuff and the `Serializable` stuff were both in the stdlib, without any deprecation warning on either, and both were given equal treatment in the docs implying that both are here to stay long-term. Retrospectively, maybe it would have been a good idea to introduce a deprecation warning back when `Serializable` was added or shortly after to signal the future direction of things, or at least to have added a note in the docs that `JSON.mapping` will eventually be deprecated and that `Serializable` is the preferred solution going forward. This is all in retrospect though

asterite:
JSON::Serializable had limitations for a long time and maybe still has. That's why I never removed JSON.mapping from the std. Maybe all the issues are fixed now.

straight-shoota:
@asterite I can't see any huge limitations. At least nothing that `mapping` doesn't have. The only open issues I could find are about namespacing (#6194, #6663) but they're not hard bugs and can easily worked around using fully qualified paths.
Unless there remain unreported issues, there doesn't seem to be a reason to use mapping over `Serializable`. As per #8263, `Serializable` is a better solution in many (if not all) regards and we should focus both development and users on that.

j8r:
People will always have cases that the stdlib does not satisfy @asterite.
The purpose of the stdlib is to fit most (say 95%) of the use cases.

bcardiff:
@veelenga / @benoist is it ok to extract `JSON.mapping` and `YAML.mapping` (maybe later `DB.mapping`) to a `mapping.cr` shard under crystal-community organization?

I still think there is value in simpler API for these mapping, so if there are to be deprecated and removed in the std-lib they should somewhere where they can be maintained.

If, so. @straight-shoota can you create & transfer such shard?

straight-shoota:
Sorry, but I'm not going to publish a shard I'm not willing to maintain.
It would probably be best if someone who likes to continue using `mapping` steps in.

bcardiff:
@straight-shoota I was not asking for you to maintain it. But to create the initial extraction so a maintainer can later be picked in the crystal-community organization.

bcardiff:
I will extract the both mappings as separate shards for now and then we can move forward with this... I will drop them in crystal-lang organization for now. If needed, we can transfer them to community later. Thanks @veelenga

bcardiff:
The repos are created:
* https://github.com/crystal-lang/json_mapping.cr
* https://github.com/crystal-lang/yaml_mapping.cr

I will wait a little bit before tagging 0.1.0 on both in case I missed something. When those shards are required, the definitons in them will override ones in the std-lib.

@straight-shoota I think the deprecation message could point to those packages.

I'm not sure if you are aware there is no warning message printed upon deprecated macros invocations.

straight-shoota:
Added the references to those shards.

> I'm not sure if you are aware there is no warning message printed upon deprecated macros invocations.

Do you think I should add an explicit macro `puts` to print a deprecation message on compilation?

straight-shoota:
Argh, those annotations don't even show up in the API docs and instead hide the actual documentation. 🤦

straight-shoota:
I guess the way to move on is to place the annotations above the doc comment (even if they currently have no effect), add a `DEPRECATED` admonition to the doc comment with the same message. 
And probably a macro `puts` to print the deprecation message on compilation.

bcardiff:
Yeah, it's either docs comments (and optionally the puts inside the macro). Or implement it in the compiler and in the docs tool.

I thought the annotation were working between the docs and the method. It's probably an issue with the macros rather than the annotations itself.

src/json/mapping.cr
```diff
@@ -1,4 +1,5 @@
 module JSON
+  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`.")]
```
bcardiff, line 2:
```suggestion
  @[Deprecated("use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping.cr`.")]
```

ditto others

src/yaml/mapping.cr
```diff
@@ -67,7 +68,10 @@ module YAML
   # it and initializes this type's instance variables.
   #
   # This macro also declares instance variables of the types given in the mapping.
+  #
+  # DEPRECATED: use YAML::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/yaml_mapping`.
   macro mapping(_properties_, strict = false)
+    {% puts "YAML.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/json_mapping`." %}
```
bcardiff, line 74:
```suggestion
    {% puts "YAML.mapping is deprecated, use JSON::Serializable instead. The legacy behaviour is also available as a shard at `github:crystal-lang/yaml_mapping.cr`." %}
```

ditto others

bcardiff:
The compiler and docs feature to play nice with deprecated macros have been merged on master and this PR can be updated now. Nothing like dogfooding 🐶 , right?

straight-shoota:
So you mean revert 795efde ?

bcardiff:
Yes @straight-shoota. Expect the new compiler to handle those cases.

PR #9273, opened 2020-05-11
Standardize YAML error location

straight-shoota:
Changes format for error location to `line X, column Y` everywhere:
* `CSV::MalformedCSVError`
* `INI::ParseException`
* `JSON::ParseException`

-------

Original description: <del>Change location reporting in YAML::ParseException to line:column

This shorter notation is already used in JSON::ParseException and INI::ParseException and the compiler.</del>


asterite:
We should use the longer version everywhere. The short version is too cryptic.

straight-shoota:
I don't think it's cryptic, just concise. Which is good because it makes better error messages.
The meaning is pretty self-explanatory in the context. This notation is ubiquitous.

Sija:
> The meaning is pretty self-explanatory in the context. This notation is ubiquitous.

@straight-shoota I disagree, `at 3:7` is far is from being self-explanatory, I'd understand if the context would be a `filename:line:col` format, which indeed is pretty clear, but that's not the case.

straight-shoota:
Has anyone ever complained about not understanding what the location reference in a JSON::ParseError message means?

straight-shoota:
> I'd understand if the context would be a filename:line:col format

I agree that this would really be a great enhancement. Similar to compiler exceptions, we could add a `filename` property to location-based exceptions which can be set when the exception bubbles up to a scope where such information exists:
```cr
File.open("file.json", "r") do |file|
  JSON.parse(file)
rescue exc : JSON::ParseException
  exc.filename = "file.json"
end
```

straight-shoota:
I've changed to the longer `line x, column y` format everywhere. Seems to be most favoured and also more common.

Sija:
@straight-shoota Sth went wrong with the rebase?

straight-shoota:
Ah no, I forgot to actually change the pre-existing occurances of `line:column` format. 🤦

bcardiff (approved):

PR #9274, opened 2020-05-11
Add YAML::Mapping error

straight-shoota:
Based on #9273

### DIFF
--- a/spec/std/yaml/schema/fail_safe_spec.cr
+++ b/spec/std/yaml/schema/fail_safe_spec.cr
@@ -46,8 +46,8 @@ describe YAML::Schema::FailSafe do
   it_raises_on_parse %(
     this: "gives"
       an: "error"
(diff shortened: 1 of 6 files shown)


spec/std/yaml/schema/fail_safe_spec.cr
```diff
@@ -46,8 +46,8 @@ describe YAML::Schema::FailSafe do
   it_raises_on_parse %(
     this: "gives"
       an: "error"
-  ), "did not find expected key at line 3, column 7, while parsing a block mapping at line 2, column 5"
-  it_raises_on_parse ":", "did not find expected key at line 1, column 1, while parsing a block mapping at line 1, column 1"
+  ), "did not find expected key at 3:7, while parsing a block mapping at 2:5"
```
asterite, line 49: "3:7" looks much more cryptic than "line 3, column 7" to me

straight-shoota: This is actually #9273, please discuss it there.

PR #9278, opened 2020-05-11
Link MSVC runtime dynamically

oprypin:
This is the best practice and also a better default, because the alternative contradicts the defaults when building any other library to link with (as evidenced by the simplification of the CI config).

But this will mean the built executables will have this DLL dependency.


src/lib_c.cr
```diff
@@ -1,5 +1,5 @@
 {% if flag?(:win32) %}
-  @[Link("libcmt")]
+  @[Link("msvcrt")]
```
RX14, line 2: I think there should be a `-Dwin_MT` to link `libcmt` for those that want to use `/MT` builds.

oprypin: Probably. But not sure I like that exact name and approach to choosing this option.

But it *is* true that this is different from `--static`.
(Wait, does that one even apply to Windows? It's *.lib either way? And I think we've been linking statically without `--static`)

oprypin:
Egh you know what, it's quite a nice "side effect" that currently any executable built just from Crystal stdlib is fully self-contained. So I don't want to change the default anymore, maybe let's just make this overridable for people who need it. (still looking for ideas about a nice flag name or such).

PR #9280, opened 2020-05-11
Update CRYSTAL_PATH to allow .crystal/shards

bcardiff:
We need to keep :lib: for a couple of releases as a migration path.

Depends on https://github.com/crystal-lang/shards/pull/367

### DIFF
--- a/bin/crystal
+++ b/bin/crystal
@@ -138,7 +138,7 @@ SCRIPT_ROOT="$(dirname "$SCRIPT_PATH")"
 CRYSTAL_ROOT="$(dirname "$SCRIPT_ROOT")"
 CRYSTAL_DIR="$CRYSTAL_ROOT/.build"
 
-export CRYSTAL_PATH=lib:$CRYSTAL_ROOT/src
+export CRYSTAL_PATH=.crystal/shards:lib:$CRYSTAL_ROOT/src
 export CRYSTAL_HAS_WRAPPER=true
 
 if [ -x "$CRYSTAL_DIR/crystal" ]; then


jhass:
So, this triggers something that's in the back of my mind for a while now.

Currently we require all distributors (people building distributions of Crystal, so package maintainers effectively), to maintain the shards lookup path. I think this wrong and there's an easy solution to this: Can we make `CRYSTAL_CONFIG_PATH` (the compile time option, not the runtime option) always prepend `lib:.crystal/shards:`? So distributors only need to set the location to where they install the standard library, and site packages if they prefer to support so. The user would still have full control by setting `CRYSTAL_PATH` in case they need to workaround something.

bcardiff:
I think it sounds good @jhass. Some where concerned in the past about the separation of concerns, since it's shards who is picking the installation path of the library, putting it in the compiler might seem weird. 

This change in the wrapper script is in order to allow a non released compiler or std-lib to be used in other projects.

The only things that will be left out is how to completely override the crystal_path since the compiler will always prepend `.crystal/shards` in your proposal, right? Or you ment for the wrapper script to prepend things to whatever `$(crystal env CRYSTAL_PATH)` returns and setting a `CRYSTAL_PATH` will set completely the path?

jhass:
Crystal does not need a wrapper script and it absolutely never should. Installing the binary into `/usr/bin` is perfectly fine if it's properly configured. Here's an example, in a distribution used by tens of thousands of users, on how to do that: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/crystal

I've been setting up the Archlinux build like this since day one. It always worked and we shouldn't break it. This does not break it. I'm just proposing to make it even simpler! If you look at the build file above, you can see it does `CRYSTAL_CONFIG_PATH="lib:/usr/lib/crystal"`. In other words it needs to maintain the shards inclusion path. 

I'm proposing that we improve here and make the situation simpler for maintainers of build scripts such as the one linked above. This can be very simply done by changing 

https://github.com/crystal-lang/crystal/blob/35c1f41f05144d9645c7bba83cd1b1deaea68fb0/src/compiler/crystal/config.cr#L6

to 
```cr
    def self.path
      ".crystal/shards:lib:#{{{env("CRYSTAL_CONFIG_PATH") || ""}}}"
    end
```

With this change build script such as the above can then instead just set `CRYSTAL_CONFIG_PATH="/usr/lib/crystal"` and don't need to maintain the shards include path.

This has absolutely nothing to do with any wrapper script. It's not mandatory to have a wrapper script to run Crystal and that's awesome and should stay that way.

oprypin:
Something should be said about Windows. You also absolutely cannot rely on a wrapper script, that just subtly breaks many kinds of usages. But you also cannot rely on specifying the path at build time, as building and installing are separate. Maybe there also needs to be a way to choose a path relative to crystal.exe. Overall, honestly, I don't see why there is no general desire to get rid of that wrapper.

jhass:
@oprypin Maybe we could just resolve relative values in `CRYSTAL_CONFIG_PATH` at runtime, relative too the binary location.

oprypin:
That *is* a good idea, but we need to resolve the conflict between the two different usages of relative path: relative to working directory, as here, and relative to the executable.
That's somewhat easier if the current always-present relative paths were to be moved out of the variable, as @jhass suggests

jhass:
Yes it would be a bit weird that `CRYSTAL_CONFIG_PATH`would be resolved relative to the binary while `CRYSTAL_PATH` would be resolved to the runtime working directory.

oprypin:
I have several concerns with this, as per https://github.com/crystal-lang/shards/issues/258#issuecomment-627589814 onwards.

1. This gives the Shards package manager preferential treatment and brings confusion and potential dilemmas for the future when someone develops a different package manager. Namely, they'll have to choose between populating the _.crystal/shards_ directory even though their package manager is not *shards* anymore, *or* asking their users to always override `CRYSTAL_PATH`.
You can also say "oh but we just call Crystal libraries shards, that's what it stands for", but then honestly, naming the package manager after that is just a separate issue of itself.
By the way, the issue of "other package managers" is not a theoretical one. For example, Git is a pretty good package manager as is: I can just clone all the repos I need into lib/ and I'm done.

2. This directory contains code that ultimately goes directly into your program. Source code should not be in a hidden directory. It is not kind towards users (esp. new ones) to have *hidden code* (that has no obvious way to get to it) that somehow implicitly gets picked up and compiled.
Additionally, I am not familiar with any other package manager that downloads libraries into a hidden subdirectory of the current directory. The ones that work best tend to put them into a visible subdirectory.
This complaint is minor, though, as misuse of dot-directories has coincidentally become more popular elsewhere so people always end up showing them anyway.

3. This decision that is ultimately about [Crystal](https://github.com/crystal-lang/crystal) was never discussed in the general Crystal community, only [in Shards repo](https://github.com/crystal-lang/shards/issues/258#issuecomment-627589814), although core team members [were implicitly made aware of it](https://github.com/crystal-lang/shards/pull/367#issuecomment-621155860).

oprypin:
What is *good* even with this change is that I can still make my own choice regarding these directories. And also that *Crystal* doesn't complain about the "lib" directory existing, only Shards does.

What is *not so good* is that I can't *conveniently* override `CRYSTAL_PATH` because I need to re-populate the default absolute path to the standard library. Changing this is already [suggested above](https://github.com/crystal-lang/crystal/pull/9280#issuecomment-626965657), and I think it would be a must-have with this change in

jhass:
>  Changing this is already suggested above, and I think it would be a must-have with this change in

No, my suggestion above does *specifically not* that. It *exclusively* talks about the compile setting for configuring the *compiler default*, `CRYSTAL_CONFIG_PATH`.  Any  mention  of the *runtime override* `CRYSTAL_PATH` was  only meant as a clarification.

jhass:
> I have several concerns with this

You raise  some valid points, some others reiterate your opinion that you already raised elsewhere. But neither of  them enlighten me into an obvious and better path forward than the one laid out so far. Well, it doesn't enlighten me of what your solution to any of them would be  really, to be  honest. I don't see how this PR  (or my proposals mentioned here) would make  status quo any worse,  so I see no point in blocking this on any of that :)

jhass:
> What is not so good is that I can't conveniently override CRYSTAL_PATH because I need to re-populate the default absolute path to the standard library.

This is already status quo, improving upon it seems like a separate discussion from here.

asterite:
I think at this point we should simply do nothing. lib is working fine and it has worked fine for years. Then someone complained that they have a directory with multiple projects in it. I think the solution is for that person to change their coding habits or rules, it's just not good to have one directory with everything mixed in it. That, instead of making an entire language and package manager make a huge change.

Can we please just drop this issue and close it?

Sija:
@asterite Whatever the habits of this person might be, that IMO was a valid point since `lib` directory has other uses and shards should have its own, exclusive one for its purposes.

oprypin:
> neither of them enlighten me into an obvious and better path forward than the one laid out so far.

Well what I imply is... something that doesn't start with a dot and definitely doesn't contain "shards". The idea space is large, I didn't want to make any random suggestion.

.crystal/lib would work as a compromise, I suppose.

oprypin:
> I don't see how this PR (or my proposals mentioned here) would make status quo any worse

My view is that it *does* make the status quo worse, because in the end state (though technically not in *this* transitional state) there will be no directory in the default `CRYSTAL_PATH` that doesn't mention `shards`.

jhass:
I'm sorry I cannot accept that argument. It's purely ideological and opinionated. There's zero technical difference.

oprypin:
A lot of decisions are not technical but affect users nonetheless.

As I was saying, a package manager that is not named "*shards*" putting libraries into a directory "*shards*" (which happens to be a name of a different package manager) is a very likely outcome of this. That's confusing.

jhass:
> That's confusing.

Even this is just your opinion and not an objective fact :)

straight-shoota:
Can we just agree that `shard` is the name for a Crystal library and `shards` is a depdency manager for installing such libraries? I don't see anything wrong with that naming.

> I can just clone all the repos I need into lib/ and I'm done.

And you can still do that with this change. In fact it even works better because there's no chance for shards accidentally interfering with those vendored-in libraries becuase it uses a separate directory.

asterite:
@Sija what other uses are there for lib? lib is the directory where libraries go. It's managed by shards if you use that.

If the person that wants lib renamed wants to share that folder with other projects, they can fork shards, change the directory shards uses, and change CRYSTAL_PATH. Sounds complex? It's much simpler a d easier than continuing this discussion for months, and making a big change like this when nobody else needs this change.

@Sija @jhass @straight-shoota do you need this change? Or is it that it might just be a good idea, but for your personal projects and use cases using lib works well?

asterite:
Specially when we are about to release 1.0, we shouldnt make suhh a big change. We can always say "yes, this will be possible after 1.0". We can also have crystal and shards read a configuration file that tells them which is the lib directory to use. Then we can use lib by default, but people that don't like that can configure it. But this can all be done after 1.0.

jhass:
@asterite I never argued for lib to be changed and have been purposefully  silent on the entire discussion because I agree  it's  pretty pointless. I don't care what the directory is named at all.  But I also  see no issue with renaming if it makes some people happy.  It's a directory, shards uses it. Whatever name it has is functionally equivalent for everything. 

The issue I brought up  here is in my mind for years, pretty much since I started maintaining a Crystal package for archlinux. Which is basically since the beginning of using Crystal. It was just not really relevant because we only ever renamed the directory once so far, from `libs` to `lib`. That we're doing it again now, I thought it's a good opportunity to bring it up. But it's completely orthogonal to what any directory is named. It's not a requirement to making this change and improving the situation I explained above is also completely independent from renaming. I just felt it's the right opportunity to think about this too, so package maintainer have to touch their scripts only once.

asterite:
> But I also see no issue with renaming if it makes some people happy

So far I've only seen people complain about this change. People that like some other name (like ".crystal/shards") are equally fine with keeping the existing name. So it just makes one people happy at the cost of making a lot of people unhappy.

> The issue I brought up here is in my mind for years

I agree that we shouldn't need a wrapper. I just can't see how you would know where the standard library is when you compile the crystal executable before knowing where the executable and the std will go when installing them... unless it's always relative to the crystal executable.

RX14:
I agree that any future crystal package manager will still manage "shards", just like bundler manages gems despite not being called "gem" (the situation is a bit different but you can see where I'm coming from). The way `crystal` looks up in `CRYSTAL_PATH` is not `shards`-specific. I don't agree that this is significant preferential treatment.

I'm not 100% happy about the hidden crystal directory, but it's *not a big deal* either. Nobody is going to look at this (unlike `_crystal`) and say "this is terrible". It shows up in most editors, and this situation is much better than most other languages. The status quo for most language-specific package managers is to store stuff hidden in `$HOME` even! That's a lot worse than this.

jhass:
> I agree that we shouldn't need a wrapper. I just can't see how you would know where the standard library is when you compile the crystal executable before knowing where the executable and the std will go when installing them... unless it's always relative to the crystal executable.

I linked a live and real world example of how you know that above. One that's that way for years and never caused a single issue or complaint. A lot of software does this, this is what any `--prefix=` argument  to any `configure` script or any `PREFIX` variable to any`make` invocation ever was.

I agree we should think about a way to specify the standard library path relative to the crystal binary. I think an easy solution is to just do placeholder and relative path expansion in `_PATH` and initially have the directory in the crystal binary is in as a placeholder variable. Like I indicated in my comments above. But that's a problem and solution space that's again wholly independent from the issue I raised, it only touchpoint is modifying the semantics of `CRYSTAL_CONFIG_PATH`, but in entirely orthogonal and compatible ways.

patleb:
@asterite

> Then someone complained that they have a directory with multiple projects in it.

No, it was one project with multiple languages and, I wasn't complaining, but pointing out the problem which I thought was better solved at the source, in `shards`, for everyone: I'm still confused that you consider `lib` as a good choice for a name for a private/reserved directory.

> I think the solution is for that person to change their coding habits or rules, it's just not good to have one directory with everything mixed in it.

Again, this wasn't the case and it's quite judgmental on your part, what makes you the authority on what I should use for my projects and how I organize them?

> It's much simpler and easier than continuing this discussion for months, and making a big change like this when nobody else needs this change.

The goal was to address the issue before 1.0 and it's short sighted to make the affirmation that nobody else needs this change. As far as I know, this was the second issue opened for this exact problem (which other people in the issues said they had also) and Crystal doesn't have a huge userbase. So, it's not very far fetched that given a growing userbase (which I'm starting to suspect that it might not be a goal) and a sufficient amount of time, the occurence of this issue would happen again.

> Specially when we are about to release 1.0, we shouldnt make suhh a big change.

I might not see the implication of such a change, so maybe it's not worth it: you're better placed than me to evaluate the matter. I assumed that making this kind of change was better done before 1.0 and that it wasn't a breaking change.

> So far I've only seen people complain about this change.

Well, you won't hear a lot about people happy as much, these are not the one that needs to argue/discuss about it, but there are some thumbs up as far as I know.

> So it just makes one people happy at the cost of making a lot of people unhappy.

Seriously, I don't care that much, I can manage without the change, it was an effort on my part to point out a problem that I thought could be solved to improve `shards` when used with tools from other languages.

Also, for what I saw about unhappy comments, it's mostly because `crystal` is dependent by default on what path `shards` uses. May I suggest to read the new constant `Shards::INSTALL_DIR` in [shards/src/config.cr](https://github.com/crystal-lang/shards/blob/master/src/config.cr#L6) instead of hardcoding the value?

patleb:
Or, just a thought, the constant could be moved to the `crystal` repo and a new command could be added to the wrapper script or the binary to output the default shards path (something like `$ crystal lib_path => .crystal/shards`). So, this command could be used in `shards` to get the constant. This way, crystal would become the authority to what is the default value and other tools could just use the same approach without the need of using `shards` or knowing beforehand where the shards are installed. I think that it would help with what @oprypin brought up.

oprypin:
@patleb That last suggestion could actually be nice in general, but unfortunately it doesn't solve the concerns. Because it *is* already *Crystal* that decides the path regardless if it's configurable. Shards just needs to follow it.

BTW another bad thing is that Shards [next] won't be usable with any older version of Crystal. Which sucks especially because the prior two releases both have stability issues.

oprypin:
BTW another bad thing is that Shards [next] won't be usable with any older version of Crystal. Which sucks especially because the prior two releases both have stability issues.

Actually not just "not usable", it will silently "succeed" without any effect on Crystal and users will just be confused what is going on.

straight-shoota:
> BTW another bad thing is that Shards [next] won't be usable with any older version of Crystal. 

That's not true. It may not work out of the box but it's as simple as setting an environment variable.

patleb:
@oprypin, both crystal and shards need to use the same path and both can be configured like @straight-shoota suggested (also, the current pull request still look for `lib` folder). My suggestion was about making crystal expose its default configuration (to make it easier for non-crystal tools) and being the one responsible of the constant `INSTALL_DIR` (and it could be better also for `shards` to re-use the same environment variable as crystal instead of `SHARDS_INSTALL_PATH`). This way, `shards` can require the file defining the constant from crystal and if it's not defined, then it's as simple as using the old path `lib`. This way, `shards` [next] could use old crystal without any problem.

waj:
I've been talking with @bcardiff and @asterite and we decided to postpone this change. I'll revert the change in shards soon, all this in favor of a, hopefully, better thought and more comprehensive solution.

The idea still requires more elaboration and probably won't arrive before 1.0, but it goes around a pair of configuration files (one in the project root dir, meant to be persisted in the repository, and one more ephemeral with local overrides). These configuration files would allow defining not only source paths, but also default compile parameters, override native library settings, tooling configuration, etc.

Please, refrain from comment about such configuration mechanism in this thread. I'll open a proper RFC soon where we can discuss the idea. I just wanted to share the reason behind this turn back. If the proposal is later dismissed we can still bring back any change on the installation directory. But at least we'll know another options were already evaluated.

In the meantime, and to solve the issue with overrides of the CRYSTAL_PATH, I'll open another PR to always add `lib` to the lookup, regardless of the build or runtime environment.

patleb:
@waj, could you explain how configuring the `shards` path has anything to do with the default value? I think that you're conflating different issues: what @oprypin brought up is another issue entirely which wasn't addressed by the pull request in `shards` (meaning the implicit dependency on `crystal` default or vice versa).

PR #9282, opened 2020-05-12
Add #each_bit method to IO

dscottboggs:
This adds a method to IO which will iterate over each bit. This came up in the discord chat and I thought it would be simple and fun to implement.

Let me know if there's anything I can do to improve the implementation of this feature.

### DIFF
--- a/spec/std/io/io_spec.cr
+++ b/spec/std/io/io_spec.cr
@@ -352,6 +352,15 @@ describe IO do
       bytes.should eq ['a'.ord.to_u8, 'b'.ord.to_u8, 'c'.ord.to_u8]
     end
 
+    it "does each_bit" do
+      bits = Array(Bool).new
+      io = SimpleIOMemory.new "\x0F"
+      io.each_bit do |b|
+        bits << b
+      end
+      bits.should eq Array(Bool).new 8, &.< 4
+    end
+
     it "raises on EOF with read_line" do
       str = SimpleIOMemory.new("hello")
       str.read_line.should eq("hello")
--- a/src/io.cr
+++ b/src/io.cr
@@ -995,6 +995,15 @@ abstract class IO
     ByteIterator.new(self)
   end
 
+  # Invokes the given block with each bit (`Bool`) in this IO
+  def each_bit
+    each_byte do |byte|
+      8.times do |i|
+        yield byte.bits_set? 1 << i
+      end
+    end
+  end
+
   # Rewinds this `IO`. By default this method raises, but including types
   # may implement it.
   def rewind


jhass:
Hey, as you say it's quite a simple thing to do. I'm not sure if its usecases are all that frequent though.

In my opinion providing a good standard library is not about providing every little utility method you can come up with. Doing so may actually clutter it and make it hard to explore. For me this is one of the major points on why we discourage aliases, if there's too much API you start to wonder too much what to use now. Also at some point it becomes impossible to remember what's there or get an intuition onto whether the thing I need right now is something I'd find in the standard library, a shard or is simple enough to just quickly roll on my own.

Finally for a standard library method I'd also like to have things that nudge me into either an expressive or performant implementation of certain problems, ideally both.

I'm not sure I see enough of these boxes checked here, so I would actually vote against including this one, unless you have some compelling usecases. Ideally showing refactors in the compiler, standard library or existing shards, where this method would make their code better :)

waj:
This has the problem when the block exited prematurely. Some bits might be lost. So, the "streaming" interface of the `IO` is not guaranteed.

I mean, if I do:
```crystal
io.each_byte do |b|
  break if ...
end
```
the next time `each_byte` is called again, the stream is resumed where it was left. That's not the case with `each_bit`.

dscottboggs:
> unless you have some compelling usecases

I only created this suggestion because someone in the discord was asking if it was a thing. It's no skin off my back if it's not included, I just thought it was a useful feature.

> the next time each_byte is called again, the stream is resumed where it was left

I hadn't realized the side-effects of `#each_byte`. Maintaining that behavior with this suggestion would require adding an additional 8 bytes to every IO class. Another valid reason to not accept this suggestion.

dscottboggs:
Given the above feedback, I think I'll make this feature an opt-in library. If anyone disagrees, feel free to reopen the PR.

watzon:
Personally I don't feel like adding a method like this is a huge deal in terms of standard library "bloat". While it may not be useful for the majority of cases, I have needed something like it several times myself when doing lower level things.

If the whole goal of Crystal, like Ruby, is to be expressive and easy to use, why not add methods like this?

jhass:
Ruby does not have such a method, right? After 25 years of existence it does not. 

Crystal already has more such narrow purpose methods than Ruby, I don't think it helps its expressiveness in all cases and on the contrary, can often invade increase cognitive load. 

Ruby historically strikes a good balance here, just look into the bug tracker on how careful they discuss any such tiny addition. Well until recently I guess, they started to add more and more less thought out stuff.

PR #9287, opened 2020-05-13
Compiler: cast fun function pointer to Proc

asterite:
Like #9255 but correctly done (the check must be done even if the argument doesn't `need_copy`). In #9255 it was almost good, but it didn't work in all cases.

Also does #9248 in a different way.

Fixes #9253
Fixes #9255

### DIFF
--- a/spec/compiler/codegen/closure_spec.cr
+++ b/spec/compiler/codegen/closure_spec.cr
@@ -700,6 +700,8 @@ describe "Code gen: closure" do
 
   it "allows passing an external function along" do
     codegen(%(
+      require "prelude"
+
       lib LibC
         fun exit(c : Int32) : NoReturn
       end
@@ -716,7 +718,9 @@ describe "Code gen: closure" do
         LibA.a(a)
       end
     ))
+  end
 
+  it "allows passing an external function along (2)" do
     codegen(%(
       lib LibFoo
         struct S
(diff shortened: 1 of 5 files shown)


spec/compiler/codegen/closure_spec.cr
```diff
@@ -700,6 +700,8 @@ describe "Code gen: closure" do
 
   it "allows passing an external function along" do
     codegen(%(
+      require "prelude"
+
```
jhass, line 704: I think if you load the prelude you don't need to `raise` (and subsequently `LibC`)  definition. I just pasted my minimal bug trigger as spec that I used to study  the generated LLVM IR.

asterite: Oooh... I see. Makes sense. I'll remove the `raise`. I think `raise` is needed now because there will be a closure check in `LibA.a(a)`.

jhass: It was before, that's why I added it :D

jhass: Btw there's a funny bug if you leave it and the prelude of, the closure check generates an empty branch (label?) causing the LLVM IR verification to fail.

spec/compiler/codegen/closure_spec.cr
```diff
@@ -700,14 +700,12 @@ describe "Code gen: closure" do
...
       lib LibC
         fun exit(c : Int32) : NoReturn
       end
```
jhass, line 707:
```suggestion
```

PR #9288, opened 2020-05-13
XML: improve xpath regarding namespaces

### LINKED ISSUE #9279: XML::Node.xpath_node dose not accept XML::Namespace
Hi
XML::Node.xpath_node dose not accept XML::Namespace as its second argument. it only accepts Hash.

Crystal 0.34.0 [4401e90f0] (2020-04-06)

LLVM: 8.0.0

```
require "xml"
str_xml=<<-XML
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope">
  <soap:Body>
    <NumberToWordsResponse xmlns="http://www.1w1g.com/">
      <NumberToWordsResult>string</NumberToWordsResult>
    </NumberToWordsResponse>
  </soap:Body>
</soap:Envelope>
XML
x = XML.parse str_xml
puts x.xpath_node("/soap:Envelope/soap:Body",x.root.not_nil!.namespace).not_nil!.name
```

Error output:
In /usr/lib/crystal/xml/xpath_context.cr:39:16

 39 | namespaces.each do |prefix, uri|
                 ^---
Error: undefined method 'each' for XML::Namespace

asterite:
Fixes #9279

I checked Nokogiri and if you do this:

```ruby
require "nokogiri"

str_xml = <<-XML
<?xml version="1.0" encoding="utf-8"?>
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope">
  <soap:Body>
    <NumberToWordsResponse xmlns="http://www.dataaccess.com/webservicesserver/">
      <NumberToWordsResult>string</NumberToWordsResult>
    </NumberToWordsResponse>
  </soap:Body>
</soap:Envelope>
XML

xml = Nokogiri::XML(str_xml)
xml.xpath("/soap:Envelope/soap:Body")
```

It works out of the box. My guess is that it passes the root namespaces of the document so they can be found with the registered prefixes. So in this PR I do the same.

I also `lchop` the "xmlns:" prefix of namespaces when registering them. Again, checking what Nokogiri does, the namespaces are returned with the "xmlns:" prefix, but that needs to be removed when registering namespaces for the xpath lookup.

### DIFF
--- a/src/xml/xpath_context.cr
+++ b/src/xml/xpath_context.cr
@@ -42,6 +42,7 @@ struct XML::XPathContext
   end
 
   def register_namespace(prefix, uri)
+    prefix = prefix.lchop("xmlns:")
     LibXML.xmlXPathRegisterNs(self, prefix.to_s, uri.to_s)
   end
 
--- a/src/xml/xpath_context.cr
+++ b/src/xml/xpath_context.cr
@@ -41,9 +41,9 @@ struct XML::XPathContext
     end
   end
 
-  def register_namespace(prefix, uri)
+  def register_namespace(prefix : String, uri : String?)
     prefix = prefix.lchop("xmlns:")
-    LibXML.xmlXPathRegisterNs(self, prefix.to_s, uri.to_s)
+    LibXML.xmlXPathRegisterNs(self, prefix, uri.to_s)
   end
 
   def register_variables(variables)
--- a/spec/std/xml/xpath_spec.cr
+++ b/spec/std/xml/xpath_spec.cr
@@ -75,7 +75,7 @@ module XML
       doc.xpath_node("//invalid").should be_nil
     end
 
-    it "finds with namespace" do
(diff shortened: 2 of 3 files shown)


src/xml/xpath_context.cr
```diff
@@ -42,6 +42,7 @@ struct XML::XPathContext
   end
 
   def register_namespace(prefix, uri)
+    prefix = prefix.lchop("xmlns:")
     LibXML.xmlXPathRegisterNs(self, prefix.to_s, uri.to_s)
```
Sija, line 46:
```suggestion
    prefix = prefix.to_s.lchop("xmlns:")
    LibXML.xmlXPathRegisterNs(self, prefix, uri.to_s)
```

asterite: I actually think the `to_s` is not needed there at all. Let me check...

PR #9292, opened 2020-05-13
Add faster `Digest#hexfinal`

didactic-drunk:
```
md5 final.hexstring 16   4.39M (227.78ns) (± 1.29%)  80.0B/op   1.13× slower
       md5 hexfinal 16   4.77M (209.85ns) (± 1.32%)  48.0B/op   1.04× slower
  md5 hexfinal(buf) 16   4.96M (201.77ns) (± 1.31%)   0.0B/op        fastest
```


waj:
Why are these methods called `hexfinal` instead of using the (deprecated) `hexdigest`?

asterite:
Also, what's the benchmark source code?

didactic-drunk:
The name `hexfinal` matches it's use and other methods with the same semantics (`Digest#final`, `Cipher#final`).

`#hexdigest` is safe to call multiple times but copies the context incurring significant (4x) overhead.  One of the reasons for it's deprecation.

didactic-drunk:
> Also, what's the benchmark source code?

[Here](https://gist.github.com/didactic-drunk/81c74230c577c29c5fd2a5b2d3d5b18c).  It's a bit of a mess but the unused portions may be of interest when comparing `#digest` and `#final`.

src/openssl/digest/digest.cr
```diff
@@ -62,7 +62,7 @@ module OpenSSL
     end
 
     private def final_impl(data : Bytes) : Nil
-      raise ArgumentError.new("data size incorrect") unless data.bytesize == digest_size
+      raise ArgumentError.new("data size incorrect #{data.bytesize}, expected #{digest_size}") unless data.bytesize == digest_size
```
Sija, line 65: That's a long line...

```suggestion
      unless data.bytesize == digest_size
        raise ArgumentError.new("Incorrect data size: #{data.bytesize}, expected: #{digest_size}")
      end
```

src/digest/base.cr
```diff
@@ -118,13 +118,52 @@ abstract class Digest::Base
...
+    raise ArgumentError.new("dst size incorrect #{dst.bytesize}, expected #{dsize}") unless dst.bytesize == dsize * 2
+
+    sary = uninitialized StaticArray(UInt8, 64)
+    tmp = sary.to_slice[0, dsize]
+    final tmp
+    tmp.hexstring dst
+  end
+
+  # Writes a hexadecimal-encoded digest to `IO`.
+  def hexfinal(io : ::IO) : Nil
```
Sija, line 160: Why the absolute scoping (`::`)?

```suggestion
  def hexfinal(io : IO) : Nil
```

src/digest/base.cr
```diff
@@ -118,13 +118,52 @@ abstract class Digest::Base
...
+    tmp.hexstring
+  end
+
+  # Writes a hexadecimal-encoded digest to `dst`.
+  #
+  # Faster than the `String` allocating version.
+  # Use when hashing in loops.
+  def hexfinal(dst : Bytes) : Nil
+    dsize = digest_size
+    raise ArgumentError.new("dst size incorrect #{dst.bytesize}, expected #{dsize}") unless dst.bytesize == dsize * 2
```
Sija, line 151:
```suggestion
    unless dst.bytesize == dsize * 2
      raise ArgumentError.new("Incorrect dst size: #{dst.bytesize}, expected: #{dsize * 2}")
    end
```

didactic-drunk:
It would be nice if these could make it in to the coming release as those getting a `deprecated` warning will see `use hexfinal` instead of use `final.hexstring` pointing them to slightly faster methods.  If it doesn't make in to the next release those who upgrade to `final.hexstring` won't be warned of the new `hexfinal` methods unless they read the documentation.

didactic-drunk:
Done.

didactic-drunk:
Soooo.....  This didn't get merged and what I [predicted](https://github.com/crystal-lang/crystal/pull/9292#issuecomment-630602151) would happen is (based on gitter chat).

>  i noticed they deprecated hexdigest in favor of Bytes#hexstring. While I can kind of understand that change, it does make porting ruby code to crystal just slightly more work, than if crystal tried to maintain compatibility. Also not sure why they wouldn't leave hexstring in, but mark it @[AlwaysInline] or such.

I suppose `hexfinal` could be renamed back to `hexdigest`.  If they use it twice they get an `Exception`.

This should probably be resolved sooner rather than later.

didactic-drunk:
Bump.

didactic-drunk:
Stalled?

didactic-drunk:
Should I rename `hexfinal` back to `hexdigest` and let programmers handle rare double calls?  A 2nd call may be embedded in Log statements that aren't in a ci or normal call path.

beta-ziliani:
We discussed it with the team, and decided to push for it. We discussed the name, being `hexdigest!` the preferred one in the team. But we are also ok with `hexdigest`, `final_hexdigest`, `hexdigest_final`. Happy to hear anyone's opinion on this matter (in particular, yours @didactic-drunk !).

didactic-drunk:
Maybe I should read my own comments (it's been so long)

`hexfinal` was chosen to match `final` and provide familiarity to ruby porting efforts

src/digest/digest.cr
```diff
@@ -133,13 +133,68 @@ abstract class Digest
     final dst
   end
 
+  # Writes the final digest output to `dst`.
```
straight-shoota, line 136: "writing to s.th." is typically associated with a stream-based operation, so I'm wondering if we can find better wording here.

Maybe this:
```suggestion
  # Puts the final digest output into `dst`.
```

src/digest/digest.cr
```diff
@@ -133,13 +133,68 @@ abstract class Digest
...
   end
 
+  # Returns a hexadecimal-encoded digest in a new `String`.
+  #
+  # This method can only be called once and raises `FinalizedError` on subsequent calls.
+  #
+  # NOTE: `.dup.final` call may be used to get an intermediate hash value.
+  def hexfinal : String
+    sary = uninitialized StaticArray(UInt8, 64)
+    tmp = sary.to_slice[0, digest_size]
```
straight-shoota, line 158: Just wondering is there some effective limitation on `digest_size` that would ensure it's less than 64 bytes? Implementations could technically use whatever size they like. Practical implications might make digests bigger than 64 bytes unlikely, but I presume it would be possible?

If we can't excuse the case for `digest_size > 64` with good certainty, there should be an alternative here.

src/digest/digest.cr
```diff
@@ -133,13 +133,68 @@ abstract class Digest
...
     @finished = true
     final_impl dst
     dst
   end
 
+  # Returns a hexadecimal-encoded digest in a new `String`.
+  #
+  # This method can only be called once and raises `FinalizedError` on subsequent calls.
+  #
+  # NOTE: `.dup.final` call may be used to get an intermediate hash value.
```
straight-shoota, line 155:
```suggestion
  # NOTE: `.dup.hexfinal` call may be used to get an intermediate hash value.
```

src/digest/digest.cr
```diff
@@ -133,13 +133,68 @@ abstract class Digest
...
   def final(dst : Bytes) : Bytes
     check_finished
     @finished = true
     final_impl dst
     dst
   end
 
+  # Returns a hexadecimal-encoded digest in a new `String`.
+  #
+  # This method can only be called once and raises `FinalizedError` on subsequent calls.
```
straight-shoota, line 153: The "can only be called once" is not limited to this method, but includes all finalizing methods. You can only call one of them, not each method once.

We could document that with references to `#finalize` in all methods' docs, or references to all methods from all methods.

src/digest/digest.cr
didactic-drunk: Some newer hashes allow variable, almost infinite output sizes requiring each implementation to validate the size.

Either this:
```crystal
protected abstract def final_output_size_check(size) : Bool
````
or an overridable
```crystal
protected def final_output_size_check(size) : Bool
  true
end
```

didactic-drunk: Sorry, I responded when busy.  

 Is there an equivalent to `ALLOCA` to dynamically stack alloc a `Bytes` compatible type?

straight-shoota: I don't think so.

We'll have to allocate on the heap when `digest_size > sary.size`.

straight-shoota: Or maybe a crazier alternative? In order to avoid an additional heap allocation we could re-purpose the the string buffer which already needs to be allocated for building the string (in `Slice#hexstring`).

```cr
string_size = digest_size * 2
string = String.new(string_size) do |buffer|
  tmp = Slice.new(buffer + digest_size, digest_size)
  final tmp
  tmp.hexstring(buffer)
  {string_size, string_size}
end

Comment (didactic-drunk) [reply src/digest/digest.cr:158]: If not I'd say leave it to those implementations to override `#hexfinal` or fix the issue in a future PR.

File: src/digest/digest.cr (hunk 7 above)
Comment (didactic-drunk) [reply src/digest/digest.cr]: See latest "documentation" commit

File: src/digest/digest.cr (hunk 5 above)
Comment (straight-shoota) [reply src/digest/digest.cr:158]: That makes a lot of sense 👍 
We should add a note in the docs for `digest_size`.

Comment (didactic-drunk) [reply src/digest/digest.cr]: Output size limitations removed except for `hexfinal(io)`

File: src/digest/digest.cr [hunk 8]
```diff
@@ -125,21 +125,79 @@ abstract class Digest
...
+  # Writes a hexadecimal-encoded digest to `IO`.
+  #
+  # Faster than the `String` allocating version.
+  #
+  # `final` or `hexfinal` can only be called once and raises `FinalizedError` on subsequent calls.
+  #
+  # NOTE: `.dup.final` call may be used to get an intermediate hash value.
+  def hexfinal(io : IO) : Nil
+    sary = uninitialized StaticArray(UInt8, 128)
+    tmp = sary.to_slice[0, digest_size * 2]
```
straight-shoota, line 195: We should address the size restriction in this method. It's not possible to use the same optimization as in the `String` overload.
I think it's probably best to raise a meaningful error if `digest_size * 2 > 128` and mention this in the documentations (that implementing types can override this method for a larger buffer size).

straight-shoota:
@didactic-drunk 
> `hexfinal` was chosen to match `final` and provide familiarity to ruby porting efforts

What kind of familiarity are you addressing? I can't find a `hexfinal` method in Ruby. Is it something else you mean here?

In fact, Ruby has [`Digest#hexdigest!`](https://ruby-doc.org/stdlib-3.1.2/libdoc/digest/rdoc/Digest/Instance.html#method-i-hexdigest-21) which is the name proposed in https://github.com/crystal-lang/crystal/pull/9292#issuecomment-1118980172
Its semantics also include an implicit call to `#reset`. After finalizing, you can't make any use of the `Digest` instance until its reset. So I believe this makes a lot of sense and is pracitcally useful.
It also simplifies the "can only get called once" logic. It implicitly resets, afterwards you can start building a new digest.

didactic-drunk:
> @didactic-drunk
> 
> > `hexfinal` was chosen to match `final` and provide familiarity to ruby porting efforts
> 
> What kind of familiarity are you addressing? I can't find a `hexfinal` method in Ruby. Is it something else you mean here?

`hex(action)` matches Ruby.
action = `final` matches Crystal.

> In fact, Ruby has [`Digest#hexdigest!`](https://ruby-doc.org/stdlib-3.1.2/libdoc/digest/rdoc/Digest/Instance.html#method-i-hexdigest-21) which is the name proposed in [#9292 (comment)](https://github.com/crystal-lang/crystal/pull/9292#issuecomment-1118980172) Its semantics also include an implicit call to `#reset`. After finalizing, you can't make any use of the `Digest` instance until its reset. So I believe this makes a lot of sense and is pracitcally useful. It also simplifies the "can only get called once" logic. It implicitly resets, afterwards you can start building a new digest.


Auto reset may cause subtle bugs with misuse. Consider:
```crystal
# Incorrect usage may be buried in a Log statement and hard to see
Log.info { "... #{digest.hexdigest!} ..." }
# Incorrect result / data corruption
digest.hexdigest! 
```

```crystal
Log.info { "... #{digest.hexfinal} ..." }
# Raises
digest.hexfinal
```

Here's another bug that's easy to create but hard to replicate with auto reset:
```crystal
@@digests = ThreadLocalValue<Digest>... # Or Pool or other class that reuses digests

def fiber_handler
  digest = @@digests.get...
  digest.update header
  digest.update data.decrypt
  digest.hexdigest!
end
```

So where's the bug?
`data.decrypt` may `raise` leaving `digest` partially updated.  The next call to `fiber_handler` running on the same `Thread` will compute an incorrect hash value.  Extremely hard to debug and something I've already encountered by putting `reset` after `*final` instead of before.  Also, your specs may pass (because there aren't network or transient errors) then fail infrequently in production.  And because it's not the current request but some future request (not necessarily the next) that fails, figuring it out is **very** time consuming.

Ok, so what's the harm in auto reset for code that can't `raise`.  Reuse requires `reset` **before** using `update`.  Auto `reset` means double reset and unnecessary overhead.  Also, can't `raise` today doesn't mean can't `raise` tomorrow.


`hexdigest!`with auto reset:
* May lead to subtle bugs
* Doesn't indicate to users `final` is any part of the method (but it does match ruby).
* Incurs `#reset` overhead for single use digests.  Example: Hashing in `HTTP` handlers or other `Fiber` spawned handlers
* Incurs `#reset` overhead for reused objects alternatively creates subtle bugs. 

Fundamentally auto reset assumes the object will be reused.  Except it **can't** be reused reliably with auto reset.

Personally I think the Ruby's `OpenSSL::Digest` class has a horrible design with additional overhead for rare uses.  I would rather see the rare uses have an explicit `reset` **in the correct place**.

If you want to solve stale updates by putting `reset` in correct place, I guess this would work.
```crystal
# Reset at start of block

digest.open do |d|
# Or
digest.reset do |d|
  digest.update ...
  digest.final
end
# Or
digest.hexdigest do |d|
  d.update ...
end
```

There may be other options but please not auto reset.

beta-ziliani (approved):
Thanks a lot for the long, very convincing explanation, and for the patience <3

straight-shoota (approved):

